PR #43, opened 2015-07-09
Bug 1178874: Only update changed translations for pofiles.

Osmose:
This is the first part of my work on [bug 1178874](https://bugzilla.mozilla.org/show_bug.cgi?id=1178874). After checking the code out, I decided that the best place to handle only updating changed translations would be within the `dump_*` functions, as they all function differently from one another.

I figured a decent way of tracking when translations are new is by storing the date of the last commit for each project and comparing it to the `approved_date` on the translations. If a `last_committed` date doesn't exist, the functionality is ignored and all translations are written.

In case of a failure during the commit process, `last_committed` won't be updated, and the new translations will be applied again during the next run. There's an edge case around the `last_committed` update failing, but I think that's fine; it'll reapply the translations in an hour and hopefully succeed next time.

Most of this PR is actually testing code; the actual changes necessary were pretty small.

PR #44, opened 2015-07-10
Remove redundant timestamp from mgmt command logs.

Osmose:
From https://github.com/mozilla/pontoon/pull/43#discussion_r34347338.

@mathjazz r?


mathjazz:
Do we need the output method at all (with encoding and newline stuff) or shall we simply call self.stdout.write? Otherwise r+.

Osmose:
> Do we need the output method at all (with encoding and newline stuff) or shall we simply call self.stdout.write? Otherwise r+.

I think it's worth keeping around, it also makes the code a little easier to read.

PR #45, opened 2015-07-13
Fix field validator

mathjazz:
This is a fix for overriden CommaSeparatedIntegerField validator:
https://github.com/mathjazz/pontoon/commit/2680745895d86270b4c8699502cba807b674c157#commitcomment-12129310

@Osmose, r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -29,7 +29,10 @@ class UserProfile(models.Model):
 
 def validate_cldr(value):
     for item in value.split(','):
-        number = int(item.strip())
+        try:
+            number = int(item.strip())
+        except ValueError:
+            return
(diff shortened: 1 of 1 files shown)


pontoon/base/models.py
```diff
@@ -29,7 +29,10 @@ class UserProfile(models.Model):
 
 def validate_cldr(value):
     for item in value.split(','):
-        number = int(item.strip())
+        try:
+            number = int(item.strip())
+        except ValueError:
+            return
```
Osmose: Since this overrides the integer check you should raise a ValidationError here complaining that it's not an integer.

Osmose:
r+wc

mathjazz:
It actually doesn't override the default validator, it only executes before it and that’s why it broke without the try-catch block.

PR #47, opened 2015-07-13
Properly select default resources and subpages

mathjazz:
This PR does two things:
1. If no resource or subpage is requested in URL, it selects the first one available **for locale**.
2. In translate view, if selected locale doesn't support selected resource or subpage, fallback to the first option available.

Also removes "unreachable code after return statement" JS warning (should be a separate branch).

@Osmose, r?

### DIFF
--- a/pontoon/base/static/js/pontoon.js
+++ b/pontoon/base/static/js/pontoon.js
@@ -99,25 +99,6 @@
               hideToolbar(target);
               return false;
             }
-
-            if (key === 9) { // Tab: status quo + move around entities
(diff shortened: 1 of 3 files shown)


Osmose:
r+. I'm still kinda fuzzy on all the frontend logic (I still don't have a good grasp on what the HTML there actually is and what it means or what a "part" is and so on) but if it works it doesn't look actively harmful. :D

mathjazz:
Part is a short name for "resource or subpage". ;)

PR #49, opened 2015-07-14
Placeables fixes

mathjazz:
* Render escape placeables with single backslash instead of double.
* Do not turn actual newline into "\n", but render it.
* Detect newline escape ("\n") as placeable.


mathjazz:
@Osmose, r?

pontoon/base/utils.py
```diff
@@ -89,9 +95,11 @@ class SpacesPlaceable(base.Ph):
         parse = classmethod(general.regex_parse)
 
     PARSERS = [
+        NewlineEscapePlaceable.parse,
         TabEscapePlaceable.parse,
         EscapePlaceable.parse,
-        general.NewlinePlaceable.parse,
+        # DISABLED: turns newline character into \n, see Firefox: netError.dtd
```
Osmose: If it's disabled we might as well just remove it rather than commenting it out. It's not like we'll add it back in later and _not_ notice that it just doesn't work right.

Osmose:
The changes look fine! I created https://github.com/Osmose/pontoon-nonweb-test as a simple repo for testing the issues dealt with here and everything seems to be fixed.

However, I'd like to try adding some automated tests for these changes. They seem small enough that testing them shouldn't be too bad. How does that sound?

mathjazz:
@Osmose, r?

pontoon/base/tests/test_placeables.py
```diff
@@ -0,0 +1,16 @@
...
+from pontoon.base.tests import TestCase
+from pontoon.base.utils import NewlineEscapePlaceable
+
+
+class PlaceablesTests(TestCase):
+    def test_newline_escape_placeable(self):
+        """Test detecting newline escape sequences"""
+        placeable = NewlineEscapePlaceable
+
+        assert_equal(placeable.parse(u'A string\\n')[1], placeable([u'\\n']))
```
Osmose: This is basically testing if your regex is correct, which I'm not terribly concerned about testing. To me, the thing to test here is `mark_placeables`, specifically if it is given input with newlines does it output the correct HTML tags.

Osmose:
If you think this is vital to merge, than I'm r+ on removing the test case and merging, _provided_ that you immediately start work on a PR adding tests for newlines being passed to `mark_placeables`. Otherwise, I think holding off is better.

PR #50, opened 2015-07-15
Fix sorting

mathjazz:
* If two projects/locales have equal share of approved strings, we sort them by comparing unapproved.
* In locale selector, sort locales by alphabetic order.
* Remove project.details from project selector when data not available.

### DIFF
--- a/pontoon/base/static/js/main.js
+++ b/pontoon/base/static/js/main.js
@@ -479,25 +479,19 @@ $(function() {
 
   // Menu sort
   $('.menu .sort span').click(function (e) {
-    function val(index, el, type) {
-      // Sort by alphabetical order
-      if (index !== 2) {
-        return $(el).find('span:eq(' + index + ')').html();
+    function getString(el) {
+      return $(el).find('span:eq(' + index + ')').html();
+    }
 
-      // Sort by approved, then by unapproved percentage
-      } else {
-        if (!$(el).find('.chart').length) {
-          return 0;
-        }
-        var chart = $(el).find('.chart').data('chart'),
-            data = JSON.parse(chart.replace(/'/g, "\""));
-
-        if (type !== "unapproved") {
-          return data.approved/data.total;
-        } else {
-          return data.translated/data.total;
-        }
+    function get(el, type) {
+      if (!$(el).find('.chart').length) {
+        return 0;
       }
+
+      var chart = $(el).find('.chart').data('chart'),
+          data = JSON.parse(chart.replace(/'/g, "\""));
+
+      return data[type]/data.total;
     }
 
     var index = $(this).index(),
@@ -511,10 +505,15 @@ $(function() {
     $(this).addClass(cls);
 
     listitems.sort(function(a, b) {
-      return (val(index, a) < val(index, b)) ? -dir :
-        (val(index, a) > val(index, b)) ? dir :
-        (val(index, a, "unapproved") < val(index, b, "unapproved")) ? -dir :
-        (val(index, a, "unapproved") > val(index, b, "unapproved")) ? dir : 0;
+      if (index !== 2) {
+        // Sort by alphabetical order
+        return getString(a).localeCompare(getString(b)) * dir;
+
+      } else {
+        // Sort by approved, then by unapproved percentage
+        return (get(a, "approved") - get(b, "approved")) * dir ||
+          (get(a, "translated") - get(b, "translated")) * dir;
+      }
     });
 
     ul.append(listitems);
--- a/pontoon/base/static/js/main.js
+++ b/pontoon/base/static/js/main.js
@@ -479,17 +479,24 @@ $(function() {
 
   // Menu sort
   $('.menu .sort span').click(function (e) {
-    function val(index, el) {
+    function val(index, el, type) {
+      // Sort by alphabetical order
       if (index !== 2) {
         return $(el).find('span:eq(' + index + ')').html();
 
+      // Sort by approved, then by unapproved percentage
       } else {
         if (!$(el).find('.chart').length) {
           return 0;
         }
         var chart = $(el).find('.chart').data('chart'),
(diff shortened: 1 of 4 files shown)


mathjazz:
@Osmose, r?

pontoon/base/static/js/main.js
```diff
@@ -479,17 +479,24 @@ $(function() {
 
   // Menu sort
   $('.menu .sort span').click(function (e) {
-    function val(index, el) {
+    function val(index, el, type) {
```
Osmose: You don't need to pass index here I think, it's hoisted to the click handler function so it's accessible within this function.

pontoon/base/static/js/main.js
```diff
@@ -505,7 +512,9 @@ $(function() {
 
     listitems.sort(function(a, b) {
       return (val(index, a) < val(index, b)) ? -dir :
-        (val(index, a) > val(index, b)) ? dir : 0;
+        (val(index, a) > val(index, b)) ? dir :
```
Osmose: `val(index, a) - val(index, b)` is equivalent to this and only calls `val` twice.

pontoon/base/static/js/main.js
```diff
@@ -505,7 +512,9 @@ $(function() {
...
+        (val(index, a, "unapproved") < val(index, b, "unapproved")) ? -dir :
```
Osmose: It feels weird to call `val` again with a single parameter different. Perhaps consider either splitting `val` up into multiple functions to avoid re-doing unnecessary work, or moving the code from `val` here instead of calling it.

pontoon/base/templates/project_selector.html
```diff
@@ -26,7 +26,7 @@
         {% endif %}
           <span class="name"
             data-slug="{{ p.slug }}"
-            data-details="{{ p.details }}">{{ p.name }}</span>
+            {% if p.details %}data-details="{{ p.details }}"{% endif %}>{{ p.name }}</span>
```
Osmose: From what I can tell from the JS if no project details are found, `Pontoon.getProjectDetails` will return null and code in `translate_old.js` will fail as it's not set up to handle null being returned.

Why is this being removed, and how does the JS not fail when it is?

Osmose:
I'm having trouble testing the unapproved sorting thing, do you have any tips on how to test that easily?

pontoon/base/templates/project_selector.html
mathjazz: Besides translate.html, we also include project_selector.html in locale.html, projects.html and admin.html templates that don't use translate_old.js. In translate.html, p.details are always available.

pontoon/base/static/js/main.js
mathjazz: True, but that will not work with string comparison.

pontoon/base/static/js/main.js
mathjazz: Yeah, I'll refactor.

pontoon/base/static/js/main.js
mathjazz: Nice catch, thanks.

mathjazz:
Updated code, please have a look again.

You can test unapproved sorting by adding at least two projects to your local install. Both of them could be the same "real" project https://pontoon.mozilla.org/projects/google-play/, you can only add e.g. Slovenian locale. Then make some translations unapproved in the DB and compare how the sort works.

pontoon/base/static/js/main.js
```diff
@@ -479,18 +479,19 @@ $(function() {
...
 
-      } else {
-        if (!$(el).find('.chart').length) {
-          return 0;
-        }
-        var chart = $(el).find('.chart').data('chart'),
-            data = JSON.parse(chart.replace(/'/g, "\""));
-        return data.approved/data.total;
+    function get(el, type) {
+      if (!$(el).find('.chart').length) {
```
Osmose: You should save this to a variable and avoid calling `$(el).find()` twice. But there's an ever better way: `data()` returned `undefined` if the jQuery object it's called on is empty, so you could even just do:

```js
var chartData = $(el).find('.chart').data('chart');
if (chartData) {
    // Return
} else {
    // Fail return
}
```

pontoon/base/static/js/main.js
```diff
@@ -479,18 +479,19 @@ $(function() {
...
+        return 0;
       }
+
+      var chart = $(el).find('.chart').data('chart'),
+          data = JSON.parse(chart.replace(/'/g, "\""));
```
Osmose: This is not really a good idea; you're relying on the happy accident that Python dicts are almost valid JSON. Instead, in the templates where you output the `data-chart` attribute, you should convert the dict to JSON. The upside is that `data()` automatically parses JSON so you won't have to call it here once that happens.

pontoon/base/static/js/main.js
```diff
@@ -479,18 +479,19 @@ $(function() {
...
+
+      return data[type]/data.total;
```
Osmose: Instead of just returning one index from the data, why not return both the `approved` and `translated` percentages? Then you only call `get` once per list item and save on some math and JSON parsing.

pontoon/base/static/js/main.js
```diff
@@ -504,8 +505,15 @@ $(function() {
     $(this).addClass(cls);
 
     listitems.sort(function(a, b) {
-      return (val(index, a) < val(index, b)) ? -dir :
-        (val(index, a) > val(index, b)) ? dir : 0;
+      if (index !== 2) {
+        // Sort by alphabetical order
+        return getString(a).localeCompare(getString(b)) * dir;
```
Osmose: Nice!

Osmose:
There's a definite improvement here, but still some changes I'd like to review. Sorry for dragging this out.

mathjazz:
@Osmose your turn. ;)

Osmose:
r+, this is top shelf now

PR #51, opened 2015-07-16
Enable Mozilla product localization

mathjazz:
Mozilla product localization involves Firefox, Firefox for Android, Thunderbird, Lightning and Mozilla Suite. Instead of having 5 different repositories with l10n resources like all other projects do, we are dealing with just 2 repositories, one for Firefox and Firefox for Android, and one for Thunderbird, Lightning and Mozilla Suite. Both of these epositories are huge, because they contain the entire codebase of these products.

That means we have to move folders around to create fake per-project repositories, because this is what Pontoon expects. Depending on how long will it take to clone 2,5 GB of repositories on each sync cycle on staging, we might need to adapt sync to do it only every 2 hours for these projects.

@Osmose, r? It was convenient to simply add 3 blocks of code to existing methods. But it's also ugly. We should probably use separate methods to do these project-specific tasks. Any recommendations on what approach to use here?

**Note**: This PR is *very* important, because once it lands, we will be able to use one tool (Pontoon) to localize everything we localize at Mozila for the very first time.

### DIFF
--- a/bin/mozilla-en-US.py
+++ b/bin/mozilla-en-US.py
@@ -1,0 +1,110 @@
+#!/usr/bin/env python
+"""This script creates or updates en-US repository from Mozilla source
+   code to use for Mozilla product localization."""
+
+import os
+import shutil
+import subprocess
+
+
+CONFIG = {
+    'sources': ['comm-aurora', 'mozilla-aurora'],
+    'target': 'mozilla-aurora',
+    'folders': [
+        'browser', 'browser/branding/official', 'dom', 'netwerk',
+        'security/manager', 'services/sync', 'toolkit', 'webapprt',
+        'mobile', 'mobile/android', 'mobile/android/base',
+        'chat', 'editor/ui', 'mail', 'other-licenses/branding/thunderbird',
+        'calendar',
+        'suite',
+    ],
+}
+
+
+def execute(command, cwd=None):
+
+    try:
+        st = subprocess.PIPE
+        proc = subprocess.Popen(
+            args=command, stdout=st, stderr=st, stdin=st, cwd=cwd)
+
+        (output, error) = proc.communicate()
+        code = proc.returncode
+        return code, output, error
+
+    except OSError as error:
+        return -1, '', error
+
+
+def pull(url, target):
+
+    # Undo local changes
+    execute(['hg', 'revert', '--all', '--no-backup'], target)
+
+    code, output, error = execute(['hg', 'pull', '-u'], target)
+    if code == 0:
+        print('Repository at ' + url + ' updated.')
+
+    else:
+        print(unicode(error))
+        print('Clone instead.')
+
+        code, output, error = execute(['hg', 'clone', url, target])
+        if code == 0:
+            print('Repository at ' + url + ' cloned.')
+        else:
+            print(unicode(error))
+
+
+def push(path):
+
+    # Add
+    execute(['hg', 'add'], path)
+
+    # Commit
+    code, output, error = execute(['hg', 'commit', '-m', 'Update'], path)
+    if code != 0 and len(error):
+        print(unicode(error))
+
+    # Push
+    code, output, error = execute(['hg', 'push'], path)
+    if code == 0:
+        print('Repository at ' + path + ' pushed.')
+    elif len(error):
+        print(unicode(error))
+
+
+# Change working directory to where script is located
+abspath = os.path.abspath(__file__)
+dname = os.path.dirname(abspath)
+os.chdir(dname)
+
+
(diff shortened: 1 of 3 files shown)


mathjazz:
@Osmose, r?

I added a script to create en-US repository with l10n files only, available at:
http://hg.mozilla.org/users/m_owca.info/mozilla-aurora/

It's much (500 times!) smaller than the full Mozilla source repository (mozilla-central + comm-central), making update cycle much faster.

A cron job is currently running the script on the old Pontoon dev server 10 min before the hour. We will have to move it to Heroku together with static pages.

This was the final step needed to make Firefox & Co. localizable using Pontoon.

pontoon/administration/files.py
```diff
@@ -770,9 +768,14 @@ def update_from_repository(project, locales=None):
         repository_url_master = repository_url.rsplit(ending, 1)[0]
         repository_path = os.path.join(repository_path_master, ending)
 
+    # If Mozilla repo, set paths
```
Osmose: Set paths to what? If the repo URL points to your special checkouts then we set the URL we're pulling updates from to be... the official mozilla HG repo? I thought we were supposed to be updating from your special version instead?

pontoon/administration/files.py
```diff
@@ -129,21 +131,18 @@ def modify_entity_mine(self, id, value):
 """ End monkeypatching """
 
 
-def get_locale_paths(project, locale):
+def get_locale_paths(project, locale, source_paths, source_directory):
```
Osmose: So instead of walking the FS you're just doing a string replacement and then checking for existence. Why the change? I wouldn't expect `os.walk` to be a source of much slowness.

bin/mozilla-en-US.py
```diff
@@ -0,0 +1,131 @@
...
+        'repository': 'comm-aurora',
+    },
+}
+
+
+SOURCE_REPOS = set(v["repository"] for v in TARGET_REPOS.values())
+
+
+def execute(command, cwd=None):
+
```
Osmose: I've been curious, why the spaces at the beginning of functions? In general your style has so many more spaces and I find it odd (purely as a matter of personal taste)

bin/mozilla-en-US.py
```diff
@@ -0,0 +1,131 @@
...
+def push(path):
+
+    # Add
+    execute(['hg', 'add'], path)
+
+    # Commit
+    code, output, error = execute(['hg', 'commit', '-m', 'Update'], path)
+    if code != 0 and len(error):
+        print(unicode(error))
```
Osmose: Can't use print as a function on Python 2 without importing it from the future.

Osmose:
The changes to remove the `locales` argument from `update_from_repository` and to get locale paths differently seems unrelated to getting Firefox localizable on Pontoon. Am I missing something? If they are unrelated, I'd rather review them as separate PRs so we don't get into a habit of batching tons of changes at once simply because that's what we were working on at the time.

pontoon/administration/files.py
Osmose: Reading this more carefully, we pull the en-US version from your URL but then we pull the locale repositories one at a time. Gotcha.

Osmose:
It was confusing at first, but I get what you're doing now and it looks pretty good. I know I've been told this before, but what is the current mechanism that updates l10n repos with strings? In theory we may be able to leverage that at a later date to have an en-US l10n repo, but I'm making that assumption while not remembering anything about how it works so we can just call me a liar.

I'm r+wc on this. Great work!

pontoon/administration/files.py
mathjazz: That way we only extract locale paths that have source paths counterparts available. This is necessary for this PR, since we use per-project en-US repos (to only display e.g. Firefox strings in Firefox project) and full locale repos, which contain files for all (up to 5, depending on locale) Mozilla projects.

bin/mozilla-en-US.py
mathjazz: SublimeLinter reminds me of doing this:
https://www.python.org/dev/peps/pep-0008/#blank-lines

Osmose: Wait, what? SublimeLinter wants you to put spaces after the `def` line? I understood pep8 to be saying that you should surround the entire function definition (including the body of it) rather than the line with `def`.

mathjazz: Oh, now I see what you mean. Blank lines at the beginning of functions removed.

pontoon/administration/files.py
Osmose: Ohhhhhhhhhhhh

PR #52, opened 2015-07-21
Bug 1186090

mathjazz:
Handle the case where the subpages don't have any resources.

### DIFF
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -267,7 +267,10 @@ def translate(request, locale, slug, part=None, template='translate.html'):
             page = pages.get(name=part)
         except Subpage.DoesNotExist:
             # If page not specified or doesn't exist
-            page = pages.filter(resource__stats__locale=l)[0]
+            page = pages[0]
+            locale_pages = pages.filter(resource__stats__locale=l)
+            if locale_pages.exists():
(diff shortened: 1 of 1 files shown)


mathjazz:
@Osmose, r?

pontoon/base/views.py
```diff
@@ -267,7 +267,10 @@ def translate(request, locale, slug, part=None, template='translate.html'):
             page = pages.get(name=part)
         except Subpage.DoesNotExist:
             # If page not specified or doesn't exist
-            page = pages.filter(resource__stats__locale=l)[0]
+            page = pages[0]
+            locale_pages = pages.filter(resource__stats__locale=l)
+            if locale_pages.exists():
```
Osmose: This does an extra query to the database outside of the one to fetch the locale pages, because you do one query for checking existence, then another to actually fetch the items.

`if locale_pages` will be True if at least one result exists, and it'll evaluate the queryset so that when you do `locale_pages[0]` below you use the cached results rather than doing another query.

Osmose:
r+wc. The whole page fetching thing could use some optimization but this addition looks fine to me.

I'd love a test to ensure this doesn't happen again but this will be so hard to isolate as to not be worth it until the translate view itself gets some love. Nice work!

PR #53, opened 2015-07-22
Fix project admin links for non-imported projects

mathjazz:
If project was created but not imported, it was impossible to open it from the admin project list, because it wasn't linked to. Because we relied on stats to check if we should create the link or not.

@Osmose, r?


Osmose:
r+, nice work :D

PR #54, opened 2015-07-27
Bug 1178874: Add new sync_projects command.

Osmose:
Adds a new management command called `sync_projects` that simultaneously pulls updates from VCS to update the database and commits changes from the database to VCS.

More details to follow, and this is still slightly incomplete, but I figured putting it up now for some pre-reviewing would be helpful. Don't merge for now.

@mathjazz Comments, criticism, and questions are welcome. :D

### DIFF
--- a/pontoon/administration/management/commands/sync_projects.py
+++ b/pontoon/administration/management/commands/sync_projects.py
@@ -141,7 +141,7 @@ class Command(BaseCommand):
         Generate a key for the given entity that is unique within the
         project.
         """
-        return ':'.join([entity.resource.path, entity.string])
+        return ':'.join([entity.resource.path, entity.key])
 
     def commit_changes(self, db_project):
         """Commit the changes we've made back to the VCS."""
@@ -299,7 +299,7 @@ class ChangeSet(object):
             'resource': self.resources[vcs_entity.resource.path],
             'string': vcs_entity.string,
             'string_plural': '',  # TODO: Support plural source.
-            'key': '',  # TODO: Support keys separate from source.
+            'key': vcs_entity.key,
             'comment': '\n'.join(vcs_entity.comments),
             'order': vcs_entity.order,
             'source': ''  # TODO: Support source
--- a/pontoon/administration/management/commands/sync_projects.py
+++ b/pontoon/administration/management/commands/sync_projects.py
@@ -5,7 +5,7 @@ from django.utils import timezone
 from bulk_update.helper import bulk_update
 
 from pontoon.administration.files import update_from_repository
-from pontoon.administration.vcs import commit_to_vcs
+from pontoon.administration.vcs import commit_to_vcs, CommitToRepositoryException
 from pontoon.base.models import Entity, Locale, Project, Resource, Translation
 from pontoon.base.utils import match_attr
 from pontoon.base.vcs_models import VCSProject
@@ -23,6 +23,7 @@ class Command(BaseCommand):
             default=False,
             help='Do not commit changes to VCS'
         )
+
     def log(self, msg, *args, **kwargs):
         """Log a message to the console."""
         self.stdout.write(msg.format(*args, **kwargs))
@@ -107,10 +108,15 @@ class Command(BaseCommand):
             changeset.create_db_entity(vcs_entity)
         else:
             for locale in db_project.locales.all():
+                if not vcs_entity.has_translation_for(locale.code):
+                    # VCS lacks an entity for this locale, so we can't
+                    # pull updates nor edit it. Skip it!
+                    continue
+
                 if db_entity.has_changed(locale.code):
                     # Pontoon changes overwrite whatever VCS has.
                     changeset.update_vcs_entity(locale.code, db_entity, vcs_entity)
-                elif vcs_entity.has_translation_for(locale.code):
+                else:
                     # If Pontoon has nothing or has not changed, and the VCS
                     # still has the entity, update Pontoon with whatever may
                     # have changed.
@@ -165,7 +171,7 @@ class Command(BaseCommand):
                     user,
                     db_project.repository_url
                 )
-            except Exception as err:
+            except CommitToRepositoryException as err:
                 result = {'message': unicode(err)}
 
             if result is not None:
--- a/pontoon/administration/management/commands/sync_projects.py
+++ b/pontoon/administration/management/commands/sync_projects.py
@@ -1,0 +1,363 @@
+from django.contrib.auth.models import User
+from django.core.management.base import BaseCommand, CommandError
+from django.utils import timezone
+
+from bulk_update.helper import bulk_update
+
+from pontoon.administration.files import update_from_repository
+from pontoon.administration.vcs import commit_to_vcs
+from pontoon.base.models import Entity, Locale, Project, Resource, Translation
+from pontoon.base.utils import match_attr
+from pontoon.base.vcs_models import VCSProject
+
+
+class Command(BaseCommand):
+    args = '<project_slug project_slug ...>'
+    help = 'Synchronize database and remote repositories.'
+
+    def log(self, msg, *args, **kwargs):
+        """Log a message to the console."""
+        self.stdout.write(msg.format(*args, **kwargs))
+
+    def info(self, msg, *args, **kwargs):
+        """Log a message to the console if --verbosity=1 or more."""
+        if self.verbosity >= 1:
+            self.log(msg, *args, **kwargs)
+
+    def debug(self, msg, *args, **kwargs):
+        """Log a message to the console if --verbosity=2."""
+        if self.verbosity == 2:
+            self.log(msg, *args, **kwargs)
+
+    def handle(self, *args, **options):
+        self.verbosity = options['verbosity']
+
+        self.log('SYNC PROJECTS: start')
+        projects = Project.objects.filter(disabled=False)
+        if args:
+            projects = projects.filter(slug__in=args)
+
+        if len(projects) < 1:
+            raise CommandError('No matching projects found.')
+
+        for project in projects:
+            self.handle_project(project)
+        self.log('SYNC PROJECTS: done')
+
+        # Once we've synced, we can delete all translations scheduled
+        # for deletion.
+        Translation.deleted_objects.all().delete()
+
+    def handle_project(self, db_project):
+        if not db_project.can_commit:
+            self.log('Skipping project {0}, cannot commit to repository.'
+                     .format(db_project.name))
+            return
+
+        # Pull changes from VCS and update what we know about the files.
+        update_from_repository(db_project)
+        vcs_project = VCSProject(db_project)
+        self.update_resources(db_project, vcs_project)
+
+        # Collect all entities across VCS and the database and get their
+        # keys so we can match up matching entities.
+        vcs_entities = self.get_vcs_entities(vcs_project)
+        db_entities = self.get_db_entities(db_project)
+        entity_keys = set().union(db_entities.keys(), vcs_entities.keys())
+
+        changeset = ChangeSet(db_project, vcs_project)
+        for key in entity_keys:
+            db_entity = db_entities.get(key, None)
+            vcs_entity = vcs_entities.get(key, None)
+            self.handle_entity(changeset, db_project, key, db_entity, vcs_entity)
+
+        # Apply the changeset to the files and then commit them.
+        changeset.execute()
+        self.commit_changes(db_project)
+
+        # Once we've successfully committed, update the last_synced date.
+        db_project.last_synced = timezone.now()
+        db_project.save()
+
+        self.log('Synced project {0}', db_project.slug)
+
+    def handle_entity(self, changeset, db_project, key, db_entity, vcs_entity):
+        """
+        Determine what needs to be synced between the database and VCS versions
+        of a single entity and log what needs to be changed in the changeset.
+        """
+        if vcs_entity is None:
+            if db_entity is None:
+                # This should never happen. What? Hard abort.
+                raise CommandError('No entities found for key {0}'.format(key))
+            else:
+                # VCS no longer has the entity, remove it from Pontoon.
+                changeset.obsolete_db_entity(db_entity)
+        elif db_entity is None:
+            # New VCS entities are added to Pontoon.
+            changeset.create_db_entity(vcs_entity)
+        else:
+            for locale in db_project.locales.all():
+                if db_entity.has_changed(locale.code):
+                    # Pontoon changes overwrite whatever VCS has.
+                    changeset.update_vcs_entity(locale.code, db_entity, vcs_entity)
+                elif vcs_entity.has_translation_for(locale.code):
+                    # If Pontoon has nothing or has not changed, and the VCS
+                    # still has the entity, update Pontoon with whatever may
+                    # have changed.
+                    changeset.update_db_entity(locale.code, db_entity, vcs_entity)
+
+    def update_resources(self, db_project, vcs_project):
+        """Update the database on what resource files exist in VCS."""
+        relative_paths = vcs_project.resources.keys()
+        db_project.resource_set.exclude(path__in=relative_paths).delete()
+
+        for relative_path, vcs_resource in vcs_project.resources.items():
+            resource, created = db_project.resource_set.get_or_create(path=relative_path)
+            resource.format = Resource.get_path_format(relative_path)
+            resource.entity_count = len(vcs_resource.entities)
+            resource.save()
+
+    def get_vcs_entities(self, vcs_project):
+        return {self.entity_key(entity): entity for entity in vcs_project.entities}
+
+    def get_db_entities(self, db_project):
+        entities = (Entity.objects
+                    .select_related('resource')
+                    .filter(resource__project=db_project, obsolete=False))
+        return {self.entity_key(entity): entity for entity in entities}
+
+    def entity_key(self, entity):
+        """
+        Generate a key for the given entity that is unique within the
+        project.
+        """
+        return ':'.join([entity.resource.path, entity.string])
+
+    def commit_changes(self, db_project):
+        """Commit the changes we've made back to the VCS."""
+        # TODO: Set translation author properly.
+        user = User.objects.filter(is_superuser=True).first()
+
+        for locale in db_project.locales.all():
+            commit_message = (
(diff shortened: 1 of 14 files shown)


pontoon/administration/management/commands/sync_projects.py
```diff
@@ -0,0 +1,327 @@
...
+
+from pontoon.administration.files import update_from_repository
+from pontoon.administration.vcs import commit_to_vcs
+from pontoon.base.models import Entity, Locale, Project, Resource, Translation
+from pontoon.base.utils import search
+from pontoon.base.vcs_models import VCSProject
+
+
+class Command(BaseCommand):
+    args = '<project_id project_id ...>'
```
mathjazz: Shall we switch to using slug instead of id?

pontoon/administration/management/commands/sync_projects.py
```diff
@@ -0,0 +1,327 @@
...
+    help = 'Synchronize database and remote repositories.'
+
+    def log(self, msg, *args, **kwargs):
+        """Log a message to the console."""
+        self.stdout.write(msg.format(*args, **kwargs) + u'\n')
+
```
mathjazz: Is that newline character at the end even necessary? https://docs.djangoproject.com/en/1.8/howto/custom-management-commands/

pontoon/administration/management/commands/sync_projects.py
```diff
@@ -0,0 +1,327 @@
...
+    def info(self, msg, *args, **kwargs):
+        """Log a message to the console if --verbosity=1 or more."""
+        if self.verbosity >= 1:
+            self.log(msg, *args, **kwargs)
+
+    def debug(self, msg, *args, **kwargs):
+        """Log a message to the console if --verbosty=2."""
```
mathjazz: Typo: verbosty -> verbosity

pontoon/administration/management/commands/sync_projects.py
```diff
@@ -0,0 +1,327 @@
...
+
+        if len(projects) < 1:
+            raise CommandError('No matching projects found.')
+
+        for project in projects:
+            self.handle_project(project)
+
+        # Once we've synced, we can deleted all translations scheduled
+        # for deletion.
+        Translation.deleted_objects.all().delete()
```
mathjazz: Typo: we can deleted -> we can delete

pontoon/base/models.py
```diff
@@ -201,6 +272,10 @@ class Entity(models.Model):
     obsolete = models.BooleanField(default=False)
 
     @property
+    def unique_key(self):
+        return ':'.join([self.resource.path, self.string])
+
```
mathjazz: Some file formats (xliff, properties, dtd, inc, ini, l20n) use "key" as a unique string identifier in a resource file.

pontoon/base/vcs_models.py
```diff
@@ -0,0 +1,107 @@
...
+        self.resource = resource
+        self.string = string
+        self.translations = {}
+        self.comments = comments
+        self.order = order
+
+    @property
+    def unique_key(self):
+        return ':'.join([self.resource.relative_path, self.string])
+
```
mathjazz: Some file formats (xliff, properties, dtd, inc, ini, l20n) use "key" as a unique string identifier in a resource file.

pontoon/base/models.py
```diff
@@ -232,6 +307,24 @@ def serialize(self):
             'obsolete': self.obsolete,
         }
 
+    def has_changed(self, locale_code):
```
mathjazz: Would these querysets be evaluated only once instead of two times if you moved the .exists() piece right after .filter()? https://docs.djangoproject.com/en/1.8/ref/models/querysets/#when-querysets-are-evaluated

mathjazz: locale_code is never used

pontoon/base/utils.py
```diff
@@ -320,3 +320,16 @@ def req(method, project, resource, locale,
...
+    return next((c for c in collection if test(c)), default)
+
+
+def search(collection, **attributes):
+    return first(
+        collection,
+        lambda i: all(getattr(i, attrib) == value
+                      for attrib, value in attributes.items()),
+        default=None
+    )
```
mathjazz: Adding short comments to both of these methods would be helpful.

pontoon/base/models.py
```diff
@@ -1,15 +1,20 @@
...
 import json
 import logging
+import os.path
 
+from django.conf import settings
 from django.contrib.auth.models import User
 from django.core.exceptions import ValidationError
 from django.db import models
-from django.db.models import Sum
+from django.db.models import Q, Sum
```
mathjazz: Q is never used.

pontoon/base/formats/lang.py
```diff
@@ -0,0 +1,133 @@
...
+
+TAG_REGEX = re.compile(r'\{(ok|l10n-extra)\}')
+
+
+class LangEntity(object):
+    def __init__(self, string, translation, comments, tags):
+        self.string = string
+        self.translations = {None: translation}
+        self.comments = comments
+        self.fuzzy = False
```
mathjazz: There is no concept of fuzzy in .lang file format.

pontoon/base/vcs_models.py
```diff
@@ -0,0 +1,107 @@
...
+                continue  # File doesn't exist, let's move on
+
+            self.files[locale.code] = resource_file
+            for index, format_entity in enumerate(resource_file.entities):
+                if format_entity.key in self.entities:
+                    vcs_entity = self.entities[format_entity.key]
+                else:
+                    vcs_entity = VCSEntity(
+                        resource=self,
+                        string=format_entity.string,
```
mathjazz: Later on, we will also need plurals.

pontoon/base/vcs_models.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def save(self):
+        for locale_code, resource_file in self.files.items():
+            resource_file.save()
+
+
+class VCSEntity(object):
+    def __init__(self, resource, string, comments, order):
+        self.resource = resource
+        self.string = string
```
mathjazz: Later on, we will also need plurals.

pontoon/administration/management/commands/sync_projects.py
```diff
@@ -0,0 +1,327 @@
...
+
+    def update_resources(self, db_project, vcs_project):
+        relative_paths = vcs_project.resources.keys()
+        db_project.resource_set.exclude(path__in=relative_paths).delete()
+
+        for relative_path, vcs_resource in vcs_project.resources.items():
+            resource, created = db_project.resource_set.get_or_create(path=relative_path)
+            resource.format = Resource.get_path_format(relative_path)
+            resource.entity_count = len(vcs_resource.entities)
+            resource.save()
```
mathjazz: Why aren't resource.project and resource.path also set?

pontoon/administration/management/commands/sync_projects.py
```diff
@@ -0,0 +1,327 @@
...
+        self.debug(u'Handling entity: `{0}`'.format(key))
+
+        # If the VCS no longer has the entity, we remove it from
+        # Pontoon.
+        if vcs_entity is None:
+            if db_entity is None:
+                # This should never happen. What? Hard abort.
+                raise CommandError('No entities found for key {0}'.format(key))
+            else:
+                # VCS no longer has the entity, remove it from Pontoon.
```
mathjazz: We should set obsolete=True instead. Removing entity will also remove translations and thus affect translation memory and contribution stats.

pontoon/administration/management/commands/sync_projects.py
```diff
@@ -0,0 +1,327 @@
...
+            resource.save()
+
+    def get_vcs_entities(self, vcs_project):
+        return {entity.unique_key: entity for entity in vcs_project.entities}
+
+    def get_db_entities(self, db_project):
+        entities = (Entity.objects
+                    .select_related('resource')
+                    .filter(resource__project=db_project))
+        return {entity.unique_key: entity for entity in entities}
```
mathjazz: We should exclude obsolete=True entities.

pontoon/administration/management/commands/sync_projects.py
```diff
@@ -0,0 +1,327 @@
...
+
+    def commit_changes(self, db_project):
+        # TODO: Set translation author properly.
+        user = User.objects.filter(is_superuser=True).first()
+
+        for locale in db_project.locales.all():
+            commit_message = (
+                'Pontoon: Update {locale.name}({locale.code}) localization of '
```
mathjazz: Add space before openning bracket.

pontoon/administration/management/commands/sync_projects.py
```diff
@@ -0,0 +1,327 @@
...
+                    user,
+                    db_project.repository_url
+                )
+            except Exception as err:
+                result = {'message': unicode(err)}
+
+            if result is not None:
+                self.log(
+                    'Committing project {project.name} for {locale.name}'
+                    '({locale.code}) failed: {reason}',
```
mathjazz: Add space before openning bracket.

mathjazz:
Very well done! I'd love to see more code comments (for methods) and more detailed description how this PR actually works. Some comments added inline, minor improvements only.

I've evaluated the code in my head against the following cases that have caused us problems recently:
- Strings that get unfuzzied get fuzzy again after update from VCS.
- If translation added to VCS, that is already stored as unapproved in Pontoon, it doesn't prevail.
- https://bugzilla.mozilla.org/show_bug.cgi?id=1178874#c2.

And it works just fine! Bravo!

A question on the new .lang file parser: since it's the only format parser that was part of Pontoon (stolen from bedrock), I suspect we won't be needing any such rewrites for other formats (which are handled by external libraries), except for the dump_* and extract_* replacemenets.

Also, shall we keep the line limit to max. 79 characters? I was trying to do that until now, but it's also causing headaches.

pontoon/base/models.py
Osmose: > Would these querysets be evaluated only once instead of two times if you moved the .exists() piece right after .filter()?

They're not evaluated when `filter` is called because it's lazy, so they only get evaluated once when `exists` is called. Putting the call on the same line vs the line below is equivalent because assigning the queryset to a variable doesn't evaluate it.

pontoon/base/formats/lang.py
Osmose: Correct, but if we don't do this then the sync code that handles setting the fuzzy flag or reading it in the database would throw AttributeErrors when we try to access the property. So we fake it out on the entity and never use it when reading or writing to the lang files.

pontoon/base/vcs_models.py
Osmose: Yeah, the plural support here is half-assed for now and will be fleshed out when we add a second format that supports proper plurals.

pontoon/administration/management/commands/sync_projects.py
Osmose: Because we're using the manager from `db_project.resource_set`, the project is implicitly set to `db_project`, and the path is set by `get_or_create`.

pontoon/administration/management/commands/sync_projects.py
mathjazz: Also, we should probably say "Updated" instead of "Update".

pontoon/administration/management/commands/sync_projects.py
```diff
@@ -0,0 +1,327 @@
...
+        return {entity.unique_key: entity for entity in entities}
+
+    def commit_changes(self, db_project):
+        # TODO: Set translation author properly.
+        user = User.objects.filter(is_superuser=True).first()
+
+        for locale in db_project.locales.all():
+            commit_message = (
+                'Pontoon: Update {locale.name}({locale.code}) localization of '
+                '{project.name}.'
```
mathjazz: And no period at the end.

Osmose:
> A question on the new .lang file parser: since it's the only format parser that was part of Pontoon (stolen from bedrock), I suspect we won't be needing any such rewrites for other formats (which are handled by external libraries), except for the dump_* and extract_* replacemenets.

Correct, sort've. For the other formats we'll need to make sure they conform to the same API that we use to alter the lang files (where we have "format entities" that have attributes we can change, and a resource class that we can call `save()` on), but we should be able to have light wrappers around the existing parsers for those. Before I merge this I'd like to have that API specifically written out as some classes for the langfile classes to inherit from.

Osmose:
Updated. I've amended the previous commit with a proper message and made changes in response to your feedback.

Notably I've added a bunch of comments explaining some things, and added some classes to define the interface between `sync_projects`, `pontoon.base.formats`, and `pontoon.base.vcs_models`.

@mathjazz Can you take a look at this again?
@mythmon If you could give some feedback on this too that'd be great. :D

As a side note, I'm willing to land and test on some projects without automated tests to see if it solves the issues we're currently having, but my next PR will be a followup with automated tests.

Osmose:
Oh snap I totally forgot to add the logging you asked for

Osmose:
There, updated based on the etherpad.

pontoon/administration/management/commands/sync_projects.py
```diff
@@ -0,0 +1,363 @@
...
+                    .select_related('resource')
+                    .filter(resource__project=db_project, obsolete=False))
+        return {self.entity_key(entity): entity for entity in entities}
+
+    def entity_key(self, entity):
+        """
+        Generate a key for the given entity that is unique within the
+        project.
+        """
+        return ':'.join([entity.resource.path, entity.string])
```
mathjazz: This will still not work for formats that use "key" as a unique string identifier in a resource file (as opposed to "string").

pontoon/administration/management/commands/sync_projects.py
```diff
@@ -0,0 +1,363 @@
...
+
+            try:
+                result = commit_to_vcs(
+                    db_project.repository_type,
+                    db_project.locale_directory_path(locale.code),
+                    commit_message,
+                    user,
+                    db_project.repository_url
+                )
+            except Exception as err:
```
mythmon: This gives me flashbacks to overly zealous exception handlers that disguised typos in plugins and made things a lot harder to debug. Is it possible to narrow the caught types of exceptions?

pontoon/administration/management/commands/sync_projects.py
```diff
@@ -0,0 +1,363 @@
...
+    def log(self, msg, *args, **kwargs):
+        """Log a message to the console."""
+        self.stdout.write(msg.format(*args, **kwargs))
+
+    def info(self, msg, *args, **kwargs):
+        """Log a message to the console if --verbosity=1 or more."""
+        if self.verbosity >= 1:
+            self.log(msg, *args, **kwargs)
+
+    def debug(self, msg, *args, **kwargs):
```
mythmon: These are logging methods probably interesting enough to move into something like a `PontoonBaseCommand` class at some point.

pontoon/administration/management/commands/sync_projects.py
```diff
@@ -0,0 +1,363 @@
...
+        """
+        Generate a key for the given entity that is unique within the
+        project.
+        """
+        return ':'.join([entity.resource.path, entity.string])
+
+    def commit_changes(self, db_project):
+        """Commit the changes we've made back to the VCS."""
+        # TODO: Set translation author properly.
+        user = User.objects.filter(is_superuser=True).first()
```
mathjazz: I'd clal this a blocker, because it's very important to give credit to localizers also in VCS commits.

Osmose: What makes me pause about this is that the old logic is to pick the most recent translator. We're not actually giving credit to every translator via VCS commits, just the one lucky enough to get approved last before they get synced.

Perhaps we can list all the authors in the commit message?

pontoon/base/models.py
```diff
@@ -255,6 +368,23 @@ def add_translation(self, string, locale, user):
...
+
+class DeletedTranslationManager(models.Manager):
+    def get_queryset(self):
+        return (super(DeletedTranslationManager, self).get_queryset()
+                .filter(deleted__isnull=False))
+
+
+def extra_default():
+    """Default value for the Translation.extra field."""
+    return {}
```
mythmon: This seems a bit silly. Why not just `extra = JSONFiled(default=lambda: {})`? Putting it out in a separate function makes me wonder what is so fancy about this function while I'm reading the model definition.

pontoon/administration/management/commands/sync_projects.py
mathjazz: Giving credit to the recent most translator is good enough, because there aren't many occasions when two localizers submit translations for the same project since last sync.

If there's more than one translator, listing the rest of them in the commit message would be better.

And the best we can do is probably this:
https://bugzilla.mozilla.org/show_bug.cgi?id=1179213

pontoon/base/vcs_models.py
```diff
@@ -0,0 +1,127 @@
+"""
+Models for working with remote translation data stored in a VCS.
+"""
+import os.path
+from itertools import chain
+
+
+class VCSProject(object):
```
mythmon: Is it normal in Pontoon to have models that aren't Django models?

pontoon/base/models.py
Osmose: See https://docs.djangoproject.com/en/1.8/ref/models/fields/#default. Lambdas can't be serialized for migrations, nor can methods that aren't defined in the top level of the module. Thus, I use a top-level function that can be serialized.

mythmon: Ah. That makes sense. I didn't know that.

pontoon/base/vcs_models.py
Osmose: Not until now!

I mean, they are undoubtly modeling the data that we work with, and they are a nice counterpoint to the DB-based models. But if your point is that calling them models is potentially misleading, I see that. I'm not sure if there's a better name for them than models, though.

Osmose:
Updated based on feedback.

@mathjazz How's it look now?

Osmose:
Updated with commit author fix.

I also found a possible blocker: The l10n langfile parser doesn't preserve linebreaks properly; it assumes everything has two line breaks between it. I'm working on a fix.

Osmose:
Updated, whitespace is now preserved in lang files.

Osmose:
@mathjazz r?

mathjazz:
r+

As I already said, this is great and can't wait for it!

PR #56, opened 2015-07-29
Fix bug 1188929: Build assets in post_compile.

Osmose:
Since collectstatic has trouble building the assets during the normal build, we build them during post_compile, when the path issues that cause collectstatic to fail don't exist anymore.

This commit also reverts the previous removal of the translate app as that was a temporary fix.

This has been tested on staging and seems to work fine. Before deploying to production we'll want to set the `DISABLE_COLLECTSTATIC` environment variable on prod to get rid of annoying non-error messages during the build.

@mathjazz r?

### DIFF
--- a/bin/post_compile
+++ b/bin/post_compile
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+export PATH=/app/.heroku/node/bin:$PATH
+./manage.py migrate --noinput
+./manage.py collectstatic --noinput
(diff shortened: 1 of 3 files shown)


mathjazz:
Nice job! r+.

So what happened? They changed the buildpack? Can we specify the version we want to use?

Osmose:
> So what happened? They changed the buildpack? Can we specify the version we want to use?

Looks like we can, I'll put up a followup pinning our buildpacks.

PR #58, opened 2015-07-31
Pin buildpacks

Osmose:
Pins our buildpacks to specific commits that we know work for deploying Pontoon. Also documents the process of pinning the multi-buildpack that combines them together.

This has been tested and works on staging, including pinning the multi-buildpack.

@mathjazz r?


mathjazz:
r+

I see server errors on staging, but they don't seem related to this.

PR #59, opened 2015-07-31
Bug 1179963: Re-enable raygun and use fork.

Osmose:
Re-enables Raygun and temporarily uses my fork of raygun4py with a fix
for our unicode issues. Once the fix gets merged we should switch back
to the official library.

@mathjazz r?


mathjazz:
r+

PR #63, opened 2015-08-03
Filter by locale when checking if translation changed.

Osmose:
This fixes an issue with deletion where approved translations for other locales would cause Entity.has_changed to return False when no approved translations exist for the given locale.

@mathjazz r?


mathjazz:
r+

PR #67, opened 2015-08-04
Bug 1190754: Do not allow newlines in .lang files translations

mathjazz:
Do not allow newlines in .lang files translations:
https://bugzilla.mozilla.org/show_bug.cgi?id=1190754

@Osmose, r?

### DIFF
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -659,6 +659,9 @@ def update_translation(request, template=None):
     translations = Translation.objects.filter(
         entity=e, locale=l, plural_form=plural_form)
 
+    if e.resource.format == 'lang' and '\n' in string:
+        return HttpResponse("Newline characters are not allowed.")
+
     # Translations exist
     if len(translations) > 0:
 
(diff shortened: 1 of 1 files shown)


pontoon/base/views.py
```diff
@@ -659,6 +659,9 @@ def update_translation(request, template=None):
     translations = Translation.objects.filter(
         entity=e, locale=l, plural_form=plural_form)
 
+    if e.resource.format == 'lang' and '\n' in string:
+        return HttpResponse("Newline characters are not allowed.")
```
Osmose: You should probably note that newlines are only not allowed for langfiles. Also, single quotes please.

Osmose:
Has there ever been a case where a newline was absolutely required in a string? I'm assuming not considering you're comfortable with this fix.

The fix works and I'm r+wc on it, but this isn't the best way to handle validation. The proper way to do this would be (model validation)[https://docs.djangoproject.com/en/1.8/ref/models/instances/#validating-objects], but I'm fine with merging this as-is and filing a bug to refactor pre-translation checks to use Django's built-in validation framework instead of putting all the validation logic in the views.

Nice work!

mathjazz:
No, newlines are not possible in .lang files by definition. Newline marks the end of the string just like quotes do in e.g. .po files.

I agree about model validation, but I added the code here because it should be moved under model together with most of the update_translation() method.

https://bugzilla.mozilla.org/show_bug.cgi?id=1191029

PR #68, opened 2015-08-04
Bug 1183013: Make Sign in warning more visible

mathjazz:
Show Sign in warning to non-authenticated users right after projects get selected. Make it persistent.
https://bugzilla.mozilla.org/show_bug.cgi?id=1183013

@Osmose, r?

### DIFF
--- a/pontoon/base/static/js/translate_old.js
+++ b/pontoon/base/static/js/translate_old.js
@@ -1235,6 +1235,10 @@ var Pontoon = (function (my) {
             name = project.html(),
             slug = project.data('slug');
 
+        if (slug !== 'pontoon-intro' && !self.user.email) {
+          return self.endLoader('Sign in to access more projects.', 'error', true);
+        }
+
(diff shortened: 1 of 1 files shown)


pontoon/base/static/js/translate_old.js
```diff
@@ -1235,6 +1235,10 @@ var Pontoon = (function (my) {
             name = project.html(),
             slug = project.data('slug');
 
+        if (slug !== 'pontoon-intro' && !self.user.email) {
```
Osmose: We don't have a better solution for this right now so don't worry about it, but someday I'd love to stop hardcoding pontoon-intro stuff everywhere (or id=1 stuff and so on).

Osmose:
r+, nice work :+1:

pontoon/base/static/js/translate_old.js
mathjazz: Yup, this is shaky!

PR #69, opened 2015-08-04
Remove unused locales

mathjazz:
We have way more locales enabled than there are localization teams at
Mozilla, so things like he-IL vs he confuse people. This migration
removes unused locales and we'll and new ones if requested.

@Osmose, r?

### DIFF
--- a/pontoon/base/migrations/0012_auto_20150804_0859.py
+++ b/pontoon/base/migrations/0012_auto_20150804_0859.py
@@ -6,10 +6,7 @@ from django.db import models, migrations
 
 def remove_unused_locales(apps, schema_editor):
     Locale = apps.get_model('base', 'Locale')
-
-    for unused_locale in UNUSED_LOCALES:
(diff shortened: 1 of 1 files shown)


pontoon/base/migrations/0012_auto_20150804_0859.py
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+def remove_unused_locales(apps, schema_editor):
+    Locale = apps.get_model('base', 'Locale')
+
+    for unused_locale in UNUSED_LOCALES:
```
Osmose: Wouldn't `Locale.objects.filter(code__in=UNUSED_LOCALES).delete()` work as well? Not that it matters really, I'm more curious if I'm missing something.

Osmose:
r+wc, good work!

PR #71, opened 2015-08-04
Do not remove en-US, which is a source locale

mathjazz:
@Osmose, r?


Osmose:
r-, once a migration is merged we can't change it. I have a PR to fix this coming shortly.

PR #72, opened 2015-08-04
Add en and en-US locales back to locale list.

Osmose:
Both are source locales and shouldn't have been removed.

@mathjazz r?


mathjazz:
r+

PR #74, opened 2015-08-05
Bug 1178874: Handle paths properly in Project path methods.

Osmose:
A subtle bug in the path methods on the Project model caused paths to be
incorrect for projects that do not have locale directories at the root of their
repo. We fix this bug and also simplify the path methods based on how they're
used.

@mathjazz r? This is the bug that was causing issues with the Firefox 3.6 Start project on staging, since it has all it's locales underneath a top-level directory named `locale`.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -172,49 +172,43 @@ class Project(models.Model):
         """Path that this project's VCS checkout is located."""
         return os.path.join(settings.MEDIA_ROOT, self.repository_type, self.slug)
 
-    def source_directory_name(self):
-        """Name of the directory that source strings are stored in."""
+    def source_directory_path(self):
+        """Path to the directory where source strings are stored."""
         for root, dirnames, filenames in os.walk(self.checkout_path):
             for dirname in dirnames:
                 if dirname in ('templates', 'en-US', 'en-GB', 'en'):
-                    return dirname
+                    return os.path.join(root, dirname)
 
-        return None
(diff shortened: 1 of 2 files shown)


mathjazz:
r+

I'm probably missing something, but why is the path handling code being rewritten as part of bug 1178874? It works just fine in the old codebase and while it's not in the right place and should be refactored, we should do things like that in separate bugs in the future and instead focus on actually fixing stuff.

mathjazz:
Also, I'm not sure this code doesn't break repository_type="file" projects:
https://github.com/mozilla/pontoon/blob/master/pontoon/administration/files.py#L241

Since we don't use them ATM, I'm fine with it, but we should definitely add that feature back.

Osmose:
> I'm probably missing something, but why is the path handling code being rewritten as part of bug 1178874? It works just fine in the old codebase and while it's not in the right place and should be refactored, we should do things like that in separate bugs in the future and instead focus on actually fixing stuff.

I originally created the duplicate path methods because I wanted to modify them to work with sync_projects without affecting update_projects. If something breaks with sync_projects, I want to be able to fall back to update_projects immediately on prod to avoid a sudden fire.

I put them on the Project model because that seemed like a good place for them to eventually live once update_projects gets deprecated and removed.

> Also, I'm not sure this code doesn't break repository_type="file" projects

It does break them, I plan to add support for them back in once we start adding support for the rest of the file types.

mathjazz:
Ok, that makes sense!

PR #75, opened 2015-08-05
Bug 1178874: Fall back to string when entity key is empty.

Osmose:
When constructing a unique key for entities during sync, if Entity.key is blank,
we should use Entity.string as the key value instead.

@mathjazz r? Found this while testing Mozilla.org sync locally. sync_projects used to do this but it got lost during some refactoring, so we need it again.

Without this fix, the langfile parser will create VCSEntities that have a key set to their string value, while database Entities will have a blank key. The code that matches up VCSEntities and database Entities for comparison will not match them up correctly since they will have different keys.


mathjazz:
r+

PR #77, opened 2015-08-05
Sync: Fallback commit author should be pontoon, not the first admin

mathjazz:
This resolves issues like this:
http://viewvc.svn.mozilla.org/vc?view=revision&sortby=date&revision=144394

@Osmose, r?


Osmose:
Nice trick! r+

PR #78, opened 2015-08-05
Remove en-GB as possible source locale directory

mathjazz:
This is a relic from the past, which we no longer use at Mozilla. We treat en-GB as a separate locale we localize into.

@Osmose, r?


Osmose:
r+

PR #79, opened 2015-08-05
Fix sorting on project landing page

mathjazz:
We returned 0 instead of dict with values set to 0 if data wasn't
available. It didn't break on other landing pages but project page,
because data is always available.

@Osmose, r?


Osmose:
r+

PR #80, opened 2015-08-05
Dir & lang attributes

mathjazz:
This fixes bugs 1190566, 1190796 and 1190953 and also adds dir and lang attributes on Terminology Search and Contributor pages.

@amire80, r?
@Osmose, r?

screen shot 2015-08-05 at 18 02 46

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -76,6 +76,11 @@ class Locale(models.Model):
         else:
             return map(int, self.cldr_plurals.split(','))
 
+    @property
+    def dir(self):
+        RTL_LOCALES = ('ar', 'fa', 'he', 'ur')
+        return 'rtl' if self.code in RTL_LOCALES else 'ltr'
+
     def __unicode__(self):
         return self.name
 
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -78,8 +78,7 @@ class Locale(models.Model):
 
     @property
     def dir(self):
-        RTL_LOCALES = ('ar', 'fa', 'he', 'ur')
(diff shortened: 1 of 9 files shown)


amire80:
I don't have a local instance to test it myself, but the code and the screenshot look sensible. Thanks!

pontoon/base/models.py
```diff
@@ -76,6 +76,10 @@ def cldr_plurals_list(self):
         else:
             return map(int, self.cldr_plurals.split(','))
 
+    @property
+    def dir(self):
```
Osmose: This is a fine way to do this. I like it!

It's interesting though, because if we wanted to be consistent with how we store data for other locales, we'd store this as a new field on the model and add a migration that sets it to true for the 4 locales we care about. But there's no real benefit to that besides being consistent with how we store the rest of the data about a locale.

I honestly can't decided between either choice. What do you think?

Osmose:
r+, nice work!

pontoon/base/models.py
mathjazz: I agree and I thought about that. But with Locales being static, I wonder if we should completely remove them from the DB? Shouldn't that also have (positive) impact on performance?

Osmose: It might, but there's also the argument that having locales in the DB allows a particular Pontoon instance to only bother supporting locales it cares about, and lets us add new locales for new teams without a code change, but that's kind've a weak argument.

It's certainly worth trying out depending on how hard of a change it'd be.

mathjazz: That's also true. And - we might not even need the dir property, because we can just use dir=auto?
https://bugzilla.mozilla.org/show_bug.cgi?id=1190566#c4

PR #81, opened 2015-08-05
Verbatim migration

mathjazz:
I'm using these to scripts to dump data from Verbatim that's not stored in VCS, like attributions and translations submission dates.

@Osmose, please review the import_verbatim.py part.

### DIFF
--- a/pontoon/administration/management/commands/import_verbatim.py
+++ b/pontoon/administration/management/commands/import_verbatim.py
@@ -1,6 +1,10 @@
 from datetime import datetime
 from django.contrib.auth.models import Permission, User
-from django.core.management.base import BaseCommand, CommandError
+from django.core.management.base import (
+    BaseCommand,
+    CommandError,
+    NoArgsCommand,
+)
 from optparse import make_option
 from pontoon.base.models import Translation
 
--- a/pontoon/administration/management/commands/import_verbatim.py
+++ b/pontoon/administration/management/commands/import_verbatim.py
@@ -50,7 +50,10 @@ class Command(BaseCommand):
                     entity__resource__project__slug=options['project'],
                     entity__resource__path=path,
                     entity__string=source,
-                    locale__code=options['locale'])
+                    locale__code=options['locale'],
+                    user=None,
+                    string=verbatim[path][source]["translation"]
+                )
 
                 for t in translations:
                     try:
--- a/pontoon/administration/management/commands/import_verbatim.py
+++ b/pontoon/administration/management/commands/import_verbatim.py
@@ -5,45 +5,52 @@ from django.core.management.base import (
     CommandError,
     NoArgsCommand,
 )
-from optparse import make_option
 from pontoon.base.models import Translation
 
-import urllib2
-import json
+import requests
 
 
 class Command(BaseCommand):
     help = 'Import translation authors and dates from Verbatim.'
 
-    option_list = NoArgsCommand.option_list + (
-        make_option('--project', action='store', dest='project',
-                    help='Project slug to import Verbatim data to'),
-        make_option('--locale', action='store', dest='locale',
-                    help='Locale code to import Verbatim data to'),
-        make_option('--filename', action='store', dest='filename',
-                    help='File name to import Verbatim data from'),
+    def add_arguments(self, parser):
+
+        parser.add_argument(
+            '--project',
+            action='store',
+            dest='project',
+            required=True,
+            help='Project slug to import Verbatim data to'
+        )
+
+        parser.add_argument(
+            '--locale',
+            action='store',
+            dest='locale',
+            required=True,
+            help='Locale code to import Verbatim data to'
+        )
+
+        parser.add_argument(
+            '--filename',
+            action='store',
+            dest='filename',
+            required=True,
+            help='File name to import Verbatim data from'
         )
 
     def handle(self, *args, **options):
-        project = options.get('project', False)
-        locale = options.get('locale', False)
-        filename = options.get('filename', False)
-
-        if not project or not locale or not filename:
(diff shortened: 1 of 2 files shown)


mathjazz:
A short description of how this works:

1. I run dump-verbatim as a management command on Verbatim server first.
2. It commits ```LOCALE_PROJECT.json``` file with the following structure to SVN:
```
FILE_PATH: {
  ORIGINAL_STRING: {
    "translation": TRANSLATION,
    "date": DATE
    "author": AUTHOR
  }  
}
```
3. Finally I run import_verbatim on heroku to set author and date on imported project.

pontoon/administration/management/commands/import_verbatim.py
```diff
@@ -0,0 +1,74 @@
...
+        make_option('--filename', action='store', dest='filename',
+                    help='File name to import Verbatim data from'),
+        )
+
+    def handle(self, *args, **options):
+        project = options.get('project', False)
+        locale = options.get('locale', False)
+        filename = options.get('filename', False)
+
+        if not project or not locale or not filename:
```
Osmose: I'm not sure we care, but Django 1.8 has a new way of adding arguments that might make this particular line unnecessary: https://docs.djangoproject.com/en/1.8/howto/custom-management-commands/#accepting-optional-arguments

pontoon/administration/management/commands/import_verbatim.py
```diff
@@ -0,0 +1,74 @@
...
+            raise CommandError('You must provide project, locale and file name.')
+
+        url = 'http://svn.mozilla.org/projects/l10n-misc/trunk/pontoon/verbatim/' + filename + '.json'
+        response = urllib2.urlopen(url)
```
Osmose: Alternatively, `requests.get(url).json()`.

pontoon/administration/management/commands/import_verbatim.py
```diff
@@ -0,0 +1,74 @@
...
+
+                        # Save permissions
+                        if not user.has_perm('base.can_localize'):
+                            can_localize = Permission.objects.get(codename="can_localize")
+                            user.user_permissions.add(can_localize)
+                            user.save()
+
+                    except User.DoesNotExist:
+                        # TODO: create missing user?
+                        self.stdout.write("Pontoon user not found: " + email + '\n')
```
Osmose: `\n` is unnecessary.

pontoon/administration/management/commands/import_verbatim.py
```diff
@@ -0,0 +1,74 @@
...
+                    except Exception as e:
```
Osmose: Not a huge fan of just catching any exception. We shouldn't expect any unknown errors, right, and thus a hard abort if we run into one is reasonable, yeah?

pontoon/administration/management/commands/import_verbatim.py
```diff
@@ -0,0 +1,74 @@
...
+                for t in translations:
+                    try:
+                        # Save user
+                        email = verbatim[path][source]["author"]
+                        user = User.objects.get(email=email)
+                        t.user = user
+                        t.approved_user = user
+
+                        # Save date
+                        date = verbatim[path][source]["date"].split("+")[0]
```
Osmose: `%z` is available for UTC offsets, why remove them?

Osmose:
Will this be run once? Or manually? Or on a regular basis?

There's a major part missing from `import_verbatim`: It doesn't actually check to see if the translations match, it just blindly assigns all existing translations to whoever the author is in Verbatim. You should only assign the metadata from Verbatim if the translation object found has the same translation as Verbatim.

Fix that and let me know and I'll give this another pass. Nice work thusfar!

pontoon/administration/management/commands/import_verbatim.py
mathjazz: They are all set to 00:00 in Verbatim and:
ValueError: 'z' is a bad directive in format '%Y-%m-%d %H:%M:%S+%z'

mathjazz:
This is now updated, thanks for great ideas!

The script is run manually right after I import data from repositories, which is why checking if translations match wasn't necessary. Checking if user==None should be enough, but I'm now doing both, just in case. We might also automate this in the future.

pontoon/administration/management/commands/import_verbatim.py
Osmose: Huh? It's totally in the docs: https://docs.python.org/2/library/datetime.html#strftime-strptime-behavior

Weird.

Osmose:
> We might also automate this in the future.

Does this mean on a regular basis? Basically I'm asking if we should try and be performant with this or not. If this is a one-off thing that only gets run when we want to occasionally import a project from Verbatim, it's fine, but if this is something that will be repeatedly used several times then we should optimize it a bit more.

mathjazz:
Not on a regular basis.

Hypothetically, one day we'll decide we want to move an entire project over from Verbatim and than we'll run this script for all locales. Which means max. 1 time per locale-project combination, because as soon as we run it, we delete data from Verbatim.

pontoon/administration/management/commands/import_verbatim.py
mathjazz: Indeed.

pontoon/administration/management/commands/import_verbatim.py
```diff
@@ -0,0 +1,82 @@
...
+from pontoon.base.models import Translation
+
+import requests
+
+
+class Command(BaseCommand):
+    help = 'Import translation authors and dates from Verbatim.'
+
+    def add_arguments(self, parser):
+
```
Osmose: dat pep8 misconception doe :P

Osmose:
r+, nice work!

PR #83, opened 2015-08-05
Use dir="auto" on all elements holding translations

mathjazz:
That means we no longer need to keep information which locales are RTL and which LTR.

@Osmose, r?


Osmose:
r+, nice work! Much simpler than the previous version.

PR #84, opened 2015-08-05
Use jinja to render commit messages.

Osmose:
This lets us put some slightly more complex logic in the commit message
template, like hiding the list of authors if none are available.

@mathjazz r?


mathjazz:
Lepa finta! r+

PR #85, opened 2015-08-06
Bug 1191133: Support using Memcached as a cache backend.

Osmose:
This adds the ability to use a shared cache between all the dynos in the app. I've provisioned a 250mb cache for staging and deployed this and since then I haven't been able to replicate the login failure issues, which I suspect are related to the cache given that session-csrf stores anonymous CSRF tokens in the cache.

Plus it's just a good idea for us to use a shared cache anyway.

@mathjazz r?


mathjazz:
Lgtm, r+

PR #86, opened 2015-08-06
Teams landing page

mathjazz:
This adds the Teams AKA Locales landing page with a familiar dashboard listing all available locales and their stats. It also removes "Developers" link from the main menu because it got too crowded.

It still uses the old way of assigning chart data to Locale objects, which should be refactored as model instance method one day.


mathjazz:
@Osmose, r?

pontoon/base/models.py
```diff
@@ -561,16 +561,33 @@ def get_chart_data(stats):
...
+    """Add chart data to locales."""
+
+    locales = Locale.objects.filter(stats__isnull=False).distinct()
+
+    for locale in locales:
+        stats = Stats.objects.filter(
+            resource__entity__obsolete=False,
+            locale=locale).distinct()
+
+        locale.chart = get_chart_data(stats)
```
Osmose: `get_chart_data` looks like a method that should be on the `Locale` class, that takes an optional `project` argument for filtering by project. Then you could add a manager method to the Locale class that prefetches the necessary stats that the function uses to reduce the number of queries, like what you did with the entity translations.

pontoon/base/templates/locale_selector.html
```diff
@@ -27,7 +27,7 @@
       {% for l in locales %}
       <li class="clearfix{% if locale and locale == l %} current{% endif %}{% if l and l.chart %} filter{% endif %}">
         {% if l and l.chart %}
-        <a href="/{{ l.code|lower }}/{{ project.slug }}/" class="clearfix">
```
Osmose: You should use the `url` helper (from `pontoon.base.templatetags.helpers`) to generate these URLs by their name instead of building them manually. It helps protect us in case we want or need to change up the URL structure.

pontoon/base/urls.py
```diff
@@ -30,6 +30,15 @@
     url(r'^contributor/(?P<email>[\w.%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,4})/$',
         RedirectView.as_view(url="/contributors/%(email)s/", permanent=True)),
 
+    # List all active localization teams
+    url(r'^teams/$',
```
Osmose: I wonder if maybe `/locale/sl/` is a better URL structure, as the _thing_ we're dealing with is really a locale rather than a locale team.

pontoon/base/urls.py
```diff
@@ -30,6 +30,15 @@
...
+        views.locales,
+        name='pontoon.teams'),
+
+    # Redirect to a team page
+    url(r'^teams/(?P<locale>[A-Za-z0-9\-\@\.]+)/$',
```
Osmose: Why this other URL? I don't see it being used anywhere.

If there's some JS that auto-generates links that end up looking like this I think we should change the JS instead of accommodating it.

Osmose: Also, I think perhaps `/locale/sl` should be the canonical URL rather than just `/sl`. As you already noticed, having a weird catch-all URL makes routing more complicated (hence why it's the last route in the urlconf). But we also probably don't want to break the old links if a bunch of external things are linking to them; do you have a sense of if anything is linking to pages like https://pontoon.mozilla.org/sl currently?

pontoon/base/views.py
```diff
@@ -104,6 +105,17 @@ def locale(request, locale, template='locale.html'):
     return render(request, template, data)
 
 
+def locales(request, template='locales.html'):
```
Osmose: Should probably be consistent with similar view names and call this the singular "locale" instead.

Osmose: Why put the template in the view arguments? I can't imagine ever wanting a different template for this page.

pontoon/base/views.py
```diff
@@ -104,6 +105,17 @@ def locale(request, locale, template='locale.html'):
...
+    """Localization teams."""
+    log.debug("Localization teams.")
```
Osmose: Why do you log at the beginning of every view? If it's about determining which view function responded to a request, that seems like a job for a middleware.

pontoon/base/views.py
```diff
@@ -104,6 +105,17 @@ def locale(request, locale, template='locale.html'):
...
+
+    data = {
```
Osmose: Nitpicking: Might as well pass this directly to render instead of bothering with the variable.

Osmose:
Nice work! The page is simple and useful. I like it!

Lemme know when you're ready for a second look. :D

pontoon/base/urls.py
mathjazz: Teams are something localizers can feel part of and they are already used to the expression:
https://wiki.mozilla.org/L10n:Teams
https://l10n.mozilla.org/teams/

pontoon/base/urls.py
mathjazz: The `/teams/sl` URL is only there if localizers type it in the URL by chance. They might be expecting it because it follows the same pattern as `/projects/x` and `/contributors/y`. Also, it doesn't hurt.

The reasons I went with the `/sl` URLs instead of only `/teams/sl` are:
* It allows us to removing chunks from the translate URLs `/sl/project/path` and never 404.
* Users of Pootle and Verbatim are familiar with it.
* It's shorter. :-)

We didn't have any internal links to `/sl`, so localizers don't even know these pages exists and are asking for them (which is also why we need the `/teams` page). But there are blog posts linking to them.

pontoon/base/views.py
mathjazz: Actually no - we already use `locale` for single team (`/sl`), so we should use `locales` (or `teams`) for teams overview. The same pattern as we use with `project` and `projects`.

And I agree about the template, this is a relic from the past.

pontoon/base/views.py
mathjazz: True, another relic.

pontoon/base/templates/locale_selector.html
mathjazz: Fixed.

pontoon/base/models.py
mathjazz: I agree, and the same goes for get_projects_with_stats(). We should make these changes at the same time, because they follow the same logic. They affects 7 differents views (locale, locales, project, projects, admin, translate, search), so they should be done separately.

mathjazz:
@Osmose, I think I'm ready for the second look.

pontoon/base/models.py
Osmose: Sounds good to me!

Osmose:
r+, looks great! Nice work.

PR #89, opened 2015-08-06
Track Entity.has_changed status in the database.

Osmose:
Instead of querying for translations every time we need to see if an entity has
changed for a particular locale, we instead store whether the entity has changed
in the database whenever translations are saved. This significantly speeds up
sync_projects by removing a ton of unnecessary database queries.

@mathjazz r?


mathjazz:
r+

PR #90, opened 2015-08-06
Bug 1191718: Only update DB during sync when necessary.

Osmose:
Uses django-dirtyfields to determine if an Entity or Translation has actually
changed, so that we only update the database in cases where changes are
necessary rather than blindly updating all the time.

My local sync test case went from 34 seconds to 20 seconds with this change. Progress!

@mathjazz r?


mathjazz:
Looks simple and straightforward. Thumbs up on performance gains! r+

PR #92, opened 2015-08-06
Update stats after sync.

Osmose:
Stats weren't actually being updated due to the bulk actions in `sync_projects`, now they're updated once all the action has happened for each project. Yay!

@mathjazz r?


mathjazz:
r+, although we could only run them for changed locales since we have that information.

PR #94, opened 2015-08-07
Update stats on sync when resource exists or is asymmetric.

Osmose:
@mathjazz Does this look any better?

### DIFF
--- a/pontoon/administration/management/commands/sync_projects.py
+++ b/pontoon/administration/management/commands/sync_projects.py
@@ -15,7 +15,8 @@ from pontoon.base.models import (
     Locale,
     Project,
     Resource,
-    Translation
+    Translation,
+    update_stats
 )
 from pontoon.base.utils import match_attr
 from pontoon.base.vcs_models import VCSProject
@@ -91,9 +92,11 @@ class Command(BaseCommand):
             vcs_entity = vcs_entities.get(key, None)
             self.handle_entity(changeset, db_project, key, db_entity, vcs_entity)
 
-        # Apply the changeset to the files and then commit them.
+        # Apply the changeset to the files, commit them, and update stats
+        # entries in the DB.
         changeset.execute()
         self.commit_changes(db_project, changeset)
+        self.update_stats(db_project, vcs_project, changeset)
 
         # Clear out the list of changed locales for entity in this
         # project now that we've finished syncing.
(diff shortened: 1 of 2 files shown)


pontoon/base/models.py
```diff
@@ -263,6 +263,11 @@ class Resource(models.Model):
 
     ALLOWED_EXTENSIONS = [f[0] for f in FORMAT_CHOICES] + ['pot']
 
+    @property
+    def is_asymmetric(self):
+        """Return True if this resource is in an asymmetric format."""
+        return self.format in ('dtd', 'properties', 'ini', 'inc')
```
mathjazz: Bah, I forgot to mention l20n - please add it to the list.

mathjazz:
r+wc: Please update the asymmetric. Nice work!

On a related note (and feel free to file a separate bug), we now call update_stats for each project and than Translation.delete() at the end of sync_projects, which also calls update_stats for each deletion:
https://github.com/Osmose/pontoon/blob/sync-stats-redux/pontoon/administration/management/commands/sync_projects.py#L70

Actually, we might just add a comment to https://bugzilla.mozilla.org/show_bug.cgi?id=1179954 and use it to refactor stats in general.

pontoon/administration/management/commands/sync_projects.py
```diff
@@ -383,6 +401,7 @@ def execute_update_db(self):
 
                     if db_translation.is_dirty():
                         self.translations_to_update.append(db_translation)
+                        self.updated_locales.add(self.locales[locale_code])
```
mathjazz: Why only here? Why not also in e.g. execute_create_db() and in the "# Any existing translations that were not approved get unapproved block"?

mathjazz:
Another minor thing: feel free to replace "asymmetric" with a more appropriate word if you can find it. We could also say "monolingual" (or "not bilingual").

Osmose:
> On a related note (and feel free to file a separate bug), we now call update_stats for each project and than Translation.delete() at the end of sync_projects, which also calls update_stats for each deletion:
> https://github.com/Osmose/pontoon/blob/sync-stats-redux/pontoon/administration/management/commands/sync_projects.py#L70

Queryset.delete() does a bulk delete and does not call delete on each instance, so update_stats is not called in that case.

Osmose:
> Actually, we might just add a comment to https://bugzilla.mozilla.org/show_bug.cgi?id=1179954 and use it to refactor stats in general.

Yeah, the stats model is weird because not only is it data that we could just generate and cache instead of calculating and saving, but the model is used for things I think it shouldn't be, like controlling what resources show up in the menu.

PR #95, opened 2015-08-07
Bug 1190843: Fix escaping HTML

mathjazz:
This fix unifies the way we display translations, original strings and
comments across the entire app. E.g. `&hellip;` and `<br />` now always
show up as is instead of being rendered sometimes.

It also unifies the way we insert these escaped strings into textarea when
e.g. copying them from suggestions or pasting placeables. We simply use
jQuery's .text() method.

### DIFF
--- a/pontoon/base/static/js/main.js
+++ b/pontoon/base/static/js/main.js
@@ -116,20 +116,10 @@ var Pontoon = (function (my) {
     /*
      * Do not render HTML tags
      *
-     * string HTML snippet that has to be displayed as code instead of rendered
+     * string String that has to be displayed as is instead of rendered
      */
     doNotRender: function (string) {
-      return string.replace(/</g, '&lt;').replace(/>/g, '&gt;');
-    },
-
-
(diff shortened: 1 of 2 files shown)


mathjazz:
@Osmose, r?

pontoon/base/static/js/main.js
```diff
@@ -116,20 +116,10 @@ var Pontoon = (function (my) {
...
-
-
-    /*
-     * Reverse function: do render HTML tags
-     *
-     * string HTML snippet that has to be rendered instead of displayed as code
-     */
-    doRender: function (string) {
-      return string.replace(/&lt;/g, '<').replace(/&gt;/g, '>');
+      return $('<div/>').text(string).html()
```
Osmose: Note that this isn't guaranteed to preserve whitespace due to differences between browsers. As long as you're aware and know that that is okay then this is fine.

Osmose:
r+, works locally and makes sense.

PR #98, opened 2015-08-10
Properly detect asymmetric projects

mathjazz:
Project is asymmetric, if any resource asymmetric (not all).

Fixes a regression:
https://github.com/mozilla/pontoon/commit/a85ee5c48299466393aa8c8c94b69e77e6a2c11e#diff-c5dc7820ff79f0945fc21de86ffc85aaL1040

### DIFF
--- a/pontoon/administration/files.py
+++ b/pontoon/administration/files.py
@@ -1229,7 +1229,7 @@ def dump_from_database(project, locale):
     # Asymmetric formats: Remove l10n files from locale repository
     asymmetric = ['dtd', 'properties', 'ini', 'inc']
 
-    if all(x in formats for x in asymmetric):
+    if all(x in asymmetric for x in formats):
         for root, dirnames, filenames in os.walk(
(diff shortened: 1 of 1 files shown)


mathjazz:
@Osmose, r?

pontoon/administration/files.py
```diff
@@ -1229,7 +1229,7 @@ def dump_from_database(project, locale):
     # Asymmetric formats: Remove l10n files from locale repository
     asymmetric = ['dtd', 'properties', 'ini', 'inc']
 
-    if all(x in formats for x in asymmetric):
+    if all(x in asymmetric for x in formats):
```
Osmose: We've got `Resource.is_asymmetric` now, how about we just loop through the resources and use that?

Also, the PR comment mentions it should be considered asymmetric if _any_ resource is asymmetric, but you're using `all` here. Should that be `any`?

mathjazz:
The PR is wrong on so many levels.

PR #100, opened 2015-08-10
Proper asymmetric projects detection

mathjazz:
Instead of checking if project contains at least one resource of any asymmetric format, we now check if all project formats are asymmetric.

Use Resource object's is_asymmetric property.

@Osmose, r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -263,10 +263,12 @@ class Resource(models.Model):
 
     ALLOWED_EXTENSIONS = [f[0] for f in FORMAT_CHOICES] + ['pot']
 
+    ASYMMETRIC = ('dtd', 'properties', 'ini', 'inc', 'l20n')
+
     @property
     def is_asymmetric(self):
         """Return True if this resource is in an asymmetric format."""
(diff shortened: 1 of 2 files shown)


pontoon/base/models.py
```diff
@@ -263,10 +263,12 @@ class Resource(models.Model):
 
     ALLOWED_EXTENSIONS = [f[0] for f in FORMAT_CHOICES] + ['pot']
 
+    ASYMMETRIC = ('dtd', 'properties', 'ini', 'inc', 'l20n')
```
Osmose: `ASYMMETRIC_FORMATS` is nicer. :D

Osmose:
You should rebase those two commits into a single commit.

Is this attached to a bug? If not, why the change?

r+wc on the code though, nice work!

mathjazz:
Renamed and rebased.

This is not attached to a bug. The change was needed because Firefox for Android Aurora only uses 3 ASYMMETRIC_FORMATS (not all), so dumps weren't handled properly. Instead of removing all locale files first and than recreating them from source files for which translations existed, we just updated existing files. That means outdated files were not staged for deletion.

PR #102, opened 2015-08-11
Fix bug 1193359: Replace unsafe glyph with FA icon

mathjazz:
Also replaces "Unchanged" filter icon (thumb up) with bars (equivalency),
because the new "No results" icons was too similar and might apply relation.

@Osmose, r?

### DIFF
--- a/pontoon/base/static/css/translate.css
+++ b/pontoon/base/static/css/translate.css
@@ -240,7 +240,7 @@ body > header .locale.select {
 
 .unchanged .status:before {
   color: #AAAAAA;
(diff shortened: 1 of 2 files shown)


Osmose:
Code's fine but I don't think the three bars is a good idea, it conflicts with the menu icon. What about an [open circle](http://fontawesome.io/icon/circle-o/)?

mathjazz:
Replaced with copy icon, leaving open circle for https://bugzilla.mozilla.org/show_bug.cgi?id=1190572.

PR #103, opened 2015-08-12
Significant entities loading optimization

mathjazz:
Loads translate view up to 80% faster. Uses dicts and hashes to minimize DB queries.

@Osmose, r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -320,27 +320,6 @@ class Entity(DirtyFieldsMixin, models.Model):
     def __unicode__(self):
         return self.string
 
-    def serialize(self):
-        try:
-            source = eval(self.source)
-        except SyntaxError:
-            source = self.source
-
-        return {
-            'pk': self.pk,
-            'original': self.string,
-            'marked': self.marked,
-            'original_plural': self.string_plural,
-            'marked_plural': self.marked_plural,
-            'key': self.key,
-            'path': self.resource.path,
-            'format': self.resource.format,
-            'comment': self.comment,
-            'order': self.order,
-            'source': source,
-            'obsolete': self.obsolete,
-        }
-
     def has_changed(self, locale):
         """
         Check if translations in the given locale have changed since the
@@ -531,6 +510,23 @@ def create_user_profile(sender, instance, created, **kwargs):
 post_save.connect(create_user_profile, sender=User)
 
 
+def get_translation_dict(entity, translations_dict, i=None):
+    translation = {}
+
+    if entity in translations_dict:
+        translations = translations_dict[entity]
+        if i:
+            translations = [x for x in translations if x['plural_form'] == i]
+        translation = sorted(translations, key=lambda k: (-k['approved'], k['date']))[0]
+
+    return {
+        'fuzzy': translation.get('fuzzy', False),
+        'string': translation.get('string', u''),
+        'approved': translation.get('approved', False),
+        'pk': translation.get('id', None)
+    }
+
+
 def get_entities(project, locale, paths=None):
     """Load project entities with locale translations."""
 
@@ -540,27 +536,53 @@ def get_entities(project, locale, paths=None):
         resources = resource_with_paths or resources
 
     entities = Entity.objects.filter(resource__in=resources, obsolete=False)
-    entities_array = []
 
-    for e in entities:
+    entities_array = entities.values(
+        'pk', 'string', 'string_plural', 'key', 'resource__path',
+        'resource__format', 'comment', 'order', 'source', 'obsolete'
+    )
+
+    translations_array = Translation.objects.filter(locale=locale, entity__in=entities).values()
+    translations_dict = {}
+
+    for t in translations_array:
+        translations_dict.setdefault(t['entity_id'], []).append(t)
+
+    for e in entities_array:
+        # Rename fields
+        e['original'] = e.pop('string')
+        e['original_plural'] = e.pop('string_plural')
+        e['format'] = e.pop('resource__format')
+        e['path'] = e.pop('resource__path')
+
+        # Mark placeables
+        e['marked'] = utils.mark_placeables(e['original'])
+        e['marked_plural'] = utils.mark_placeables(e['original_plural'])
+
+        # Set source
+        try:
+            e['source'] = eval(e['source'])
+        except SyntaxError:
+            pass
+
+        # Set translations
         translation_array = []
 
-        # Entities without plurals
-        if e.string_plural == "":
-            translation = get_translation(entity=e, locale=locale)
-            translation_array.append(translation.serialize())
+        # Singular
+        if e['original_plural'] == u'':
+            translation_array.append(
+                get_translation_dict(e['pk'], translations_dict)
+            )
 
-        # Pluralized entities
+        # Plural
         else:
             for i in range(0, locale.nplurals or 1):
-                translation = get_translation(
-                    entity=e, locale=locale, plural_form=i)
-                translation_array.append(translation.serialize())
+                translation_array.append(
+                    get_translation_dict(e['pk'], translations_dict, i)
+                )
 
-        obj = e.serialize()
-        obj["translation"] = translation_array
+        e['translation'] = translation_array
 
-        entities_array.append(obj)
     return sorted(entities_array, key=lambda k: k['order'])
 
 
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -1,3 +1,4 @@
+import collections
 import datetime
 import json
 import logging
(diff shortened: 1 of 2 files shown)


pontoon/base/models.py
```diff
@@ -531,6 +510,23 @@ def create_user_profile(sender, instance, created, **kwargs):
 post_save.connect(create_user_profile, sender=User)
 
 
+def get_translation_dict(entity, translations_dict, i=None):
```
Osmose: This should really be a method like `serialize` was. General rule of thumb is that if a function always takes a model instance as it's first argument, you should consider if it should be a method on the class instead.

pontoon/base/models.py
```diff
@@ -531,6 +510,23 @@ def create_user_profile(sender, instance, created, **kwargs):
...
+    translation = {}
+
+    if entity in translations_dict:
+        translations = translations_dict[entity]
+        if i:
+            translations = [x for x in translations if x['plural_form'] == i]
+        translation = sorted(translations, key=lambda k: (-k['approved'], k['date']))[0]
+
+    return {
+        'fuzzy': translation.get('fuzzy', False),
```
Osmose: This `get` stuff is kind of unclear; your intent is to return a default value if the requested entity isn't in `translations_dict`, so you might as well just check for it directly instead of using `get` to handle that case:

```python
if entity not in translations_dict:
    # return a dict with all the default values being passed to get
else:
    #return the real dict
```

pontoon/base/models.py
Osmose: What is `i`? `plural_form` would be a better argument name.

pontoon/base/models.py
```diff
@@ -540,27 +536,53 @@ def get_entities(project, locale, paths=None):
...
+    entities_array = entities.values(
+        'pk', 'string', 'string_plural', 'key', 'resource__path',
+        'resource__format', 'comment', 'order', 'source', 'obsolete'
+    )
+
+    translations_array = Translation.objects.filter(locale=locale, entity__in=entities).values()
+    translations_dict = {}
+
+    for t in translations_array:
+        translations_dict.setdefault(t['entity_id'], []).append(t)
```
Osmose: There's a class in the stdlib called [`collections.defaultdict`](https://docs.python.org/2/library/collections.html#collections.defaultdict) that provides this functionality for you.

Osmose:
So the actual optimization here is fetching all translations at once instead of doing a single DB query for each translation you need, which is a good and necessary optimization, and kudos for noticing it! But this isn't the way to go about it.

First, you should keep `serialize` a method on the entity class - the name and functionality is a little weird (it sort of sounds like something you'd use to save an entity to a file) but it's fine for now. You should move the logic to fetching translations within `serialize`, as it makes sense that the entity is responsible for getting its own translations and lowers the complexity of the code calling `serialize`.

But then how do you make fetching the translations a single query? There's a lovely queryset function called [`prefetch_related`](https://docs.djangoproject.com/en/1.8/ref/models/querysets/#prefetch-related) which lets us instruct Django to fetch related objects in a single query. In this case, we want to tell Django to fetch all translations for the entities that we're interested in:

```python
entities = Entity.objects.filter(resource__in=resources, obsolete=False).prefetch_related('translation_set')
```

Then, when we call `entity.translation_set.all()`, it won't do another query, it will just return all the related translations.

In fact, Django 1.7 even lets us specify a custom queryset for the prefetch query, meaning we can do even more filtering to save us work:

```python
entities = Entity.objects.filter(resource__in=resources, obsolete=False).prefetch_related(
    Prefetch('translation_set', queryset=Translation.objects.filter(locale=locale, to_attr='fetched_translations')
)
```

After that you can just make `serialize` or some other method/property on the Entity class search `entity.fetched_translations` and filter out translations that don't match the requested plural (if necessary). It fetches the translations in one query, while keeping the complexity of the view code that calls `entity.serialize` low.

So, r- on the current code, but this is absolutely the right direction to be moving in. Also note that there are pitfalls with `prefetch_related` around shared model instances and duplication. I don't think you'll run into any of the annoying ones with this particular problem, but in any case it's a confusing tool to be used with care.

I highly recommend setting up [django-debug-toolbar](https://github.com/django-debug-toolbar/django-debug-toolbar) locally before you try this so you can measure the number of queries and ensure that they're actually being reduced. In the future we might want to always have it enabled if `DEBUG` is True but for now just using it to measure your progress locally.

Lemme know if you have any questions or need help!

pontoon/base/models.py
mathjazz: This is how I implemented it initially. Then this looked like a good idea to save some lines of code and keep dict structure in one place. But it's also less clear, I agree.

mathjazz:
Updated. Thank you, this was very educative!

That being said, I haven't yet figured out why is it ~30% slower than 7ae1168 (where all your comments are addressed except for the last one (prefetch), but my old approach with dicts is still used). Django debug toolbar (thanks!) shows they're making the same number of DB queries.

To put this in perspective, getting entities for messages.po file of AMO takes
- 22 seconds on master
- 4,5 seconds in a0d14d9
- 3,2 seconds in 7ae1168

I ran this a few times and tried other projects and the ration is about the same.

pontoon/base/models.py
```diff
@@ -377,6 +357,87 @@ def add_translation(self, string, locale, user):
...
 
+    def get_translation(self, plural_form=None):
+        """Get fetched translation of a given entity."""
+        translations = self.fetched_translations
+
+        if translations:
+            if plural_form:
+                translations = [t for t in translations if t.plural_form == plural_form]
+
+            translation = sorted(translations, key=lambda k: (-k.approved, k.date))[0]
```
Osmose: Approved is a boolean, putting `-` in front is weird. Maybe `not approved` or `-1 if k.approved else 0`?

Osmose: If there are no translations matching the requested plural form, this will throw an `IndexError`. You should probably do the plural filtering before the `if translations` check.

pontoon/base/models.py
```diff
@@ -377,6 +357,87 @@ def add_translation(self, string, locale, user):
...
+        else:
+            return {
+                'fuzzy': False,
+                'string': u'',
+                'approved': False,
+                'pk': None
+            }
+
+    @classmethod
+    def serialize(self, project, locale, paths=None):
```
Osmose: This needs a better name, this method now has little to do with serialization now.

pontoon/base/models.py
```diff
@@ -377,6 +357,87 @@ def add_translation(self, string, locale, user):
...
+
+        entities = (entities
+                    .prefetch_related(
+                        Prefetch(
+                            'translation_set',
+                            queryset=Translation.objects.filter(locale=locale),
+                            to_attr='fetched_translations'
+                        )
+                    )
+                    .prefetch_related('resource')
```
Osmose: You should be able to pass this as an argument to the first `prefetch_related` call.

pontoon/base/models.py
```diff
@@ -377,6 +357,87 @@ def add_translation(self, string, locale, user):
...
+                    )
```
Osmose: Adding this on another line but not un-indented is weird. The hanging indent is useful for querysets where you're chaining multiple calls, but since we're just doing the one call to `prefetch_related` here, I'd recommend just indenting once:

```python
entities = entities.prefetch_related(
    'resource',
    Prefetch(
        # blah blah
    )
)
```

pontoon/base/models.py
```diff
@@ -377,6 +357,87 @@ def add_translation(self, string, locale, user):
...
+
+        else:
+            return {
+                'fuzzy': False,
+                'string': u'',
+                'approved': False,
+                'pk': None
+            }
+
+    @classmethod
```
Osmose: Generally methods like this (that fetch a set of model instances) go on the Manager class, meaning you'll need to make a [custom manager](https://docs.djangoproject.com/en/1.8/topics/db/managers/#custom-managers).

pontoon/base/models.py
```diff
@@ -377,6 +357,87 @@ def add_translation(self, string, locale, user):
...
+                'path': entity.resource.path,
+                'format': entity.resource.format,
+                'comment': entity.comment,
+                'order': entity.order,
+                'source': source,
+                'obsolete': entity.obsolete,
+                'translation': translation_array,
+            })
+
+        return sorted(entities_array, key=lambda k: k['order'])
```
Osmose: Alternatively, you can just sort the entities when you query them with `.order('order')` and then use `yield` to turn this method into a generator. Would save on building the list then appending.

Osmose:
Nice work on the improvement! I wonder if the difference between the prefetched version and the dict version is due to the dict version using `values` to avoid building full model instances? I wouldn't worry about that difference for now.

r+wc

mathjazz:
Fixed IndexError and a few other things. Merging and deploying, we can add other improvements later.

PR #105, opened 2015-08-13
Bug 1178874: Add tests for sync_projects.py.

Osmose:
Phew! This took a while but will be really worthwhile in the long run. This is a set of tests for the stuff in `sync_projects.py`. This doesn't cover `vcs_models.py` or the langfile parser, but it's about time to get back to working on features, and these tests will still be useful for the coming sync work.

@mathjazz r?

PR #106, opened 2015-08-17
POFile support for sync_projects

Osmose:
This PR adds support for the po/pot formats to sync_projects. It also updates sync_projects to support plural strings and sources for formats that require them.

Because I had to modify sync_projects a bit, I based this PR off the existing #105 PR. That should be merged before this, and I can rebase once it lands. In the meantime, you can view individual commits to see the three extra commits this adds.

This also adds tests for the pofile parser. They're designed to be generic enough that any format can use them as long as they provide the proper input and output files for each test, but I haven't yet actually done the work yet to make them generic. I figure the PR for xliff can support that so we can have a common set of "requirements" that each format supports, with opt-ins to optional features (for example, testing for plurals on po and xliff formats, but not for langfiles).

I've also done local testing and things seem to work, using the Affiliates project as a testcase, but this could definitely use some more attention on a manual testing side. :D
@mathjazz r?

### DIFF
--- a/pontoon/base/tests/formats/test_po.py
+++ b/pontoon/base/tests/formats/test_po.py
@@ -289,7 +289,10 @@ class POTests(TestCase):
         """)))
 
     def test_save_plural_remove(self):
-        """Any missing plurals should be removed from the pofile."""
+        """
+        Any missing plurals should be set to an empty string in the
+        pofile.
+        """
         test_input = self.generate_pofile(dedent("""
             msgid "Plural %(count)s string"
             msgid_plural "Plural %(count)s strings"
@@ -307,6 +310,7 @@ class POTests(TestCase):
             msgid "Plural %(count)s string"
             msgid_plural "Plural %(count)s strings"
             msgstr[0] "New Plural"
+            msgstr[1] ""
         """)))
 
     def test_save_remove_fuzzy(self):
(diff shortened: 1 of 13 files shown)


mathjazz:
Among other things, we should use this as a real-world test:
https://bugzilla.mozilla.org/show_bug.cgi?id=1193953

mathjazz:
r+, the code looks good. But I'd also like to test it on staging first.

You should probably unify classnames of POResource and LangFile (I see why you didn't use POFile).

Osmose:
Rebased now that the tests have been merged.

pontoon/base/tests/formats/test_po.py
```diff
@@ -0,0 +1,385 @@
...
+
+        self.assert_pofile_equal(path, self.generate_pofile(dedent("""
+            msgid "Plural %(count)s string"
+            msgid_plural "Plural %(count)s strings"
+            msgstr[0] "New Plural"
+            msgstr[1] "New Plurals"
+        """)))
+
+    def test_save_plural_remove(self):
+        """Any missing plurals should be removed from the pofile."""
```
mathjazz: Not like this. You need to keep the length of msgstr to be nplurals, otherwise PO file will not be valid. When deleting strings, you basically need to set them to empty string: "".

Osmose:
I've addressed the feedback and this is ready to merge, but you mentioned wanting to test on staging, so I deployed this on staging. Lemme know when you're happy with this and I'll merge!

PR #107, opened 2015-08-18
Fix bug 1195912 - Django 1.8.4 security update

mathjazz:
@Osmose, r?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -80,8 +80,6 @@ dj-database-url==0.3.0
 # sha256: EgxGIdHk9a2r4KaDRj0b56WmuZLt1HZNMjxifSKSUeA
 django-dotenv==1.3.0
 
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -82,7 +82,8 @@ django-dotenv==1.3.0
 
 # sha256: BH0PTJMmKzOAEEmi3N2u8Jwp50HAOpR6NVbqR0ju0uI
 # sha256: K7ZU_MBf1TAXyIyvK8OLXF6iPJH4rH8KKLKQ2vIwW7o
```
Osmose: I wonder if we should remove these old SHAs?

Osmose:
r+

requirements.txt
mathjazz: We probably should.

PR #108, opened 2015-08-19
Exclude unchanged translations from stats

mathjazz:
In both, contributor profile and top contributors page we no longer
count unchanged translations.

This is a quick fix for, but more improvements are needed to the top contributors page:
https://bugzilla.mozilla.org/show_bug.cgi?id=1196140

@Osmose, r?

### DIFF
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -19,7 +19,7 @@ from django.contrib.auth.decorators import login_required
 from django.contrib.auth.models import User
 from django.core.mail import send_mail
 from django.core.urlresolvers import reverse
-from django.db.models import Count
+from django.db.models import Count, F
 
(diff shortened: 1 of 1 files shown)


pontoon/base/views.py
```diff
@@ -313,7 +313,12 @@ def contributor(request, email, template='user.html'):
     except User.DoesNotExist:
         raise Http404
 
-    translations = Translation.objects.filter(user=user)
+    translations = (
+        Translation.objects.filter(user=user)
+        .exclude(string=F('entity__string'))
```
Osmose: Woah, I didn't know F objects could span relations. Crazy!

Osmose:
r+, this is a good temporary fix. Nice work!

PR #110, opened 2015-08-19
Bug 1178874: Add xliff support to sync_projects.

Osmose:
Also refactors the pontoon.base.formats tests to be reusable between formats.

@mathjazz r?

### DIFF
--- a/pontoon/base/formats/__init__.py
+++ b/pontoon/base/formats/__init__.py
@@ -5,7 +5,7 @@ See base.py for the ParsedResource base class.
 """
 import os.path
 
-from pontoon.base.formats import lang, po
+from pontoon.base.formats import lang, po, xliff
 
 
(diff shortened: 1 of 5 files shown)


mathjazz:
1. You need to enable xliff in formats/__init__.py.

2. It's not possible to delete translations.

3. The file is completely restructured, even though we use the same library. But that's a minor problem and probably a bug in translate toolkit.
https://viewvc.svn.mozilla.org/vc/projects/l10n-misc/trunk/firefox-ios/sl/firefox-ios.xliff?r1=144802&r2=145339&sortby=date

Osmose:
Updated, fixed the embarrassing miss of adding xliff to formats/__init__.py and fixed an issue with how we deleted the target element when deleting the translation.

mathjazz:
Works! r+

PR #115, opened 2015-08-20
Fix Django SHA in requirements.txt.

Osmose:
When I tried installing the requirements via peep I got an error with the Django hash not match. I've updated it to the has that works for me.

@mathjazz Can you confirm if this hash works for you? If not, then maybe there's something wrong on my end?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -80,6 +80,7 @@ dj-database-url==0.3.0
 # sha256: EgxGIdHk9a2r4KaDRj0b56WmuZLt1HZNMjxifSKSUeA
 django-dotenv==1.3.0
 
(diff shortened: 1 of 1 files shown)


mathjazz:
Sorry...

    django: expected I3bD2MX0lbMCsRLXIyyEdhEwxDDhhAwFoqArKPF91ZY
                 got gmmWyB4cx3NQASTVwZIS5KdoGlXuFp-rkIXyswFacNg

Osmose:
Updated to include both SHAs, one is for the wheel and the other is for the tar.gz.

mathjazz:
r+

PR #117, opened 2015-08-21
Top contributors improvements

mathjazz:
Top contributors order is broken, this fixes it and also adds notice on excluding unchanged translations.

@Osmose, r?


pontoon/base/templates/users.html
```diff
@@ -13,7 +13,7 @@
       <tr>
         <th>Rank</th>
         <th>Contributor</th>
-        <th>Translations</th>
+        <th>Translations <span>* without unchanged</span></th>
```
Osmose: `sup` would probably be a better tag to use here. 

Also I think adding something like a footer with a complete sentence like "Translation counts do not include unchanged translations." would be better, maybe with a `sup` tag next to the header that links to the footnote.

pontoon/base/static/css/users.css
```diff
@@ -14,6 +14,11 @@ th {
   text-transform: uppercase;
 }
 
+th span {
```
Osmose: Font size should probably be smaller too.

Osmose:
Could use some tests to make sure your filtering works as expected.

r+wc, nice work!

mathjazz:
@Osmose r?

I think I forgot to ask for it last week.

pontoon/base/templates/users.html
```diff
@@ -47,6 +47,11 @@
       </tr>
     {% endfor %}
     </tbody>
+    <tfoot>
+      <tr>
+        <td colspan="3"><sup id="fn-translations">Translation counts do not include unchanged translations.<a href="#ref-translations" title="Jump back on top.">↩</a></sup></td>
```
Osmose: "to top" is more appropriate than "on top" here.

Osmose:
r+, looks great! Even better than I imagined. :D

PR #118, opened 2015-08-21
Move gravatar_url() to User moder

mathjazz:
@Osmose, r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -34,17 +34,15 @@ User.add_to_class('display_name', user_display_name)
 
 
 def user_gravatar_url(self, size):
-    gravatar_url = "//www.gravatar.com/avatar/" + \
-        hashlib.md5(self.email.lower()).hexdigest() + "?"
+    email = hashlib.md5(self.email.lower()).hexdigest()
     data = {'s': str(size)}
 
     if not settings.DEBUG:
         append = '_big' if size > 44 else ''
-        default = settings.SITE_URL + static('img/anonymous' + append + '.jpg')
-        data['d'] = default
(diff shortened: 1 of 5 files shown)


pontoon/base/models.py
```diff
@@ -30,6 +33,21 @@ def user_display_name(self):
...
 
 
+def user_gravatar_url(self, size):
+    gravatar_url = "//www.gravatar.com/avatar/" + \
+        hashlib.md5(self.email.lower()).hexdigest() + "?"
+    data = {'s': str(size)}
+
+    if not settings.DEBUG:
+        append = '_big' if size > 44 else ''
+        default = settings.SITE_URL + static('img/anonymous' + append + '.jpg')
```
Osmose: No need for the default variable, just assign it directly.

pontoon/base/models.py
```diff
@@ -30,6 +33,21 @@ def user_display_name(self):
 User.add_to_class('display_name', user_display_name)
 
 
+def user_gravatar_url(self, size):
+    gravatar_url = "//www.gravatar.com/avatar/" + \
```
Osmose: DOUBLE QUOTED STRINGS AAAAAAAAAHHHHHHH (we prefer single quotes unless the string contains single quotes that you don't want to escape)

Also, I'm generally not a fan of line continuations using `\`. I prefer using parenthesis to group statements, but in this case you'd do better to assign the hash to a variable and then using string formatting to insert it in the URL. You could even just do a format at the end to add the query args too instead of using `+=`. These are all nits though, the code itself seems to work fine.

Osmose:
Tests for the function are optional but appreciated.

r+wc, looks great!

PR #119, opened 2015-08-21
Search2terminology

mathjazz:
Rename both, the title and the URL from Search to Terminology. Search can mean many things and Terminology is much more specific. Also keep the legacy /search URL and redirect it to /terminology. Also update links on static 500 page.

@Osmose r?

### DIFF
--- a/pontoon/base/urls.py
+++ b/pontoon/base/urls.py
@@ -75,9 +75,13 @@ urlpatterns = patterns(
         name='pontoon.profile'),
 
     # Terminology Search
-    url(r'^search/$',
+    url(r'^terminology/$',
         views.search,
         name='pontoon.search'),
+
+    # Legacy: Redirect to /terminology
+    url(r'^search/$',
+        RedirectView.as_view(url="/terminology/", permanent=True)),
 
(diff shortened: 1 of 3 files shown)


pontoon/base/urls.py
```diff
@@ -75,10 +75,14 @@
...
 
     # Terminology Search
-    url(r'^search/$',
+    url(r'^terminology/$',
         views.search,
         name='pontoon.search'),
 
+    # Legacy: Redirect to /terminology
+    url(r'^search/$',
+        RedirectView.as_view(url="/terminology/", permanent=True)),
```
Osmose: Use `pattern_name='pontoon.search'` instead of `url` to make this not rely on the specific URL. :D

Osmose:
It'd be neat if you changed the helper tab in the translate view to say "Terminology" instead of search as well.

r+wc otherwise, nice work!

mathjazz:
Search tab is different from Machinery (another lame name) in that they both work with Terminology, Machinery using the original string and search by actually searching with provided keyword. So I'd rather keep it as it is.

PR #120, opened 2015-08-21
Update OpenSans font to display correct Romanian diacritics

mathjazz:
Also removes references to local OpenSans fonts, which probably also
use incorrect Romanian diacritics.

@Osmose r?


Osmose:
r+

Jobava:
++ excellent!

PR #122, opened 2015-08-24
Fix IndexError when resource has no stats in translate view.

Osmose:
Specifically, if a resource has no stats for a given locale, but it has
stats for other locales, viewing strings for that locale in translate
causes an IndexError due to the view code only filtering out unimportant
locales _after_ checking the existence of stats for the resource.

@mathjazz r?

### DIFF
--- a/pontoon/base/tests/__init__.py
+++ b/pontoon/base/tests/__init__.py
@@ -6,6 +6,7 @@ from django.contrib.auth.models import User
 from django.template.defaultfilters import slugify
 from django.test import TestCase as BaseTestCase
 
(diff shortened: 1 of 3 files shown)


mathjazz:
r+

Osmose:
> Did you run into the IndexError somehow or you figured this one out just by reading the code?

When I was tracking down the issue with the missing stats object, I added en-GB to my local instance directly to the database (bypassing update_stats) and then ran sync_projects, and saw the errors that this fixes since it didn't bother creating stats objects in en-GB.

PR #126, opened 2015-08-26
Update raygun4py fork commit.

Osmose:
The new commit contains fixes for raygun4py where it was printing errors to
stderr instead of logging them properly, causing extra output during test runs
and other situations.

https://github.com/Osmose/raygun4py/commit/52489dfe17e27119940f0a640e80a6a7577b0606 is the commit being added on top of our existing changes in the fork. Still waiting on Raygun to release a new version with our old fixes as well.

I also added a logging filter for raygun4py logging so we don't see "No API Key aaaaahhh" errors every time a test triggers the middleware.

@mathjazz r?

### DIFF
--- a/pontoon/settings/base.py
+++ b/pontoon/settings/base.py
@@ -395,7 +395,7 @@ LOGGING = {
 
 ## Tests
 TEST_RUNNER = 'django_nose.NoseTestSuiteRunner'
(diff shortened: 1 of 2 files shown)


mathjazz:
r+

PR #127, opened 2015-08-26
When building Go and Sign out URLs, make them absolute

mathjazz:
This is a regression caused by removal of the <base> tag.

@Osmose r?


Osmose:
r+

PR #128, opened 2015-08-26
Only show resource menu if project uses > 1 resource

mathjazz:
This is a regression caused by #122.

@Osmose r?


Osmose:
Does this break the unit test for this? I'd suspect so, and if not, maybe it should. Either way this is something that should be in the tests somehow. Maybe add a unit test afterwards?

r+ otherwise, sorry for the miss!

PR #132, opened 2015-08-27
Only count approved translations when calculating progress

mathjazz:
To avoid leaving the false impression work is complete.

@Osmose r?


Osmose:
r+, looks good!

PR #134, opened 2015-08-27
Add test for leading UTF-8 BOM in langfiles.

Osmose:
@mathjazz r?

The langfile has the BOM, but you'll need a hex editor to see it.


mathjazz:
LGTM, r+

PR #135, opened 2015-08-28
Fix tooltip flickering

mathjazz:
When hovering progress charts on project or team or admin page, tooltip
flickers if you hover over the handle (arrow). This solution clips the
arrow, so that hovering over it is not possible, which means flickering
is gone. To achieve this, I use CSS clip property (deprecated, but
needed for Firefox), -webkit-clip-path (needed for WebKit) and clip-path
(standard).

@Osmose r?


Osmose:
r+, nice work!

PR #136, opened 2015-08-28
On dashboards, also show progress as percentage

mathjazz:
Sometimes it's hard to distinguish if progress chart is at 99.99% or 100%.
Percentage allows localizers to quickly see if there's work to do or not.
For that reason, we also use floor as rounding function, because showing
100% when 9999/10000 strings are translated is a red herring.

@Osmose r?


Osmose:
r+, looks good!

PR #138, opened 2015-08-31
Only show entities for which locale has stats for

mathjazz:
This is needed for project-locale pairs which only use one resource,
whereas source locale uses multiple resources. In such cases, we have
been showing all entities to the locale, which is clearly wrong.

@Osmose I have to merge this immediately.


Osmose:
So stats is being used here to mark that a locale actually has the specific resource, right? So we're making this function only return entities for the resources that actually exist for that locale.

A test would be lovely. r+ other than that, looks good!

PR #139, opened 2015-08-31
Hotifx hide single parts

mathjazz:
@Osmose I'm merging this because Affiliates, Firefox Updater and Firefox Tiles are confusing people.

### DIFF
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -261,12 +261,13 @@ def translate(request, locale, slug, part=None, template='translate.html'):
         except Subpage.DoesNotExist:
             # If page not specified or doesn't exist
             page = pages[0]
-            locale_pages = pages.filter(resources__stats__locale=l)
-            if locale_pages:
-                page = locale_pages[0]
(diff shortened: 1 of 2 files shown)


pontoon/base/views.py
```diff
@@ -261,12 +261,13 @@ def translate(request, locale, slug, part=None, template='translate.html'):
         except Subpage.DoesNotExist:
             # If page not specified or doesn't exist
             page = pages[0]
-            locale_pages = pages.filter(resources__stats__locale=l)
```
Osmose: `locale_pages` is a nicer variable name. Instead of checking the length of `pages` down below, just set `page` to `None` and check `if page is not None`. Or, even better, just set `data['part']` directly in each of these places.

Osmose: Oh wait the `>` check means something. Let me think...

Osmose: Yeah, ignore me, there's no better way than what you've done without a more serious refactoring.

Osmose:
A test would be nice! r+wc :D

PR #142, opened 2015-08-31
Also detect L10nJs using data-l10n-ids

mathjazz:
@Osmose


Osmose:
r+, with the caveat that I have no idea how that attribute works.

mathjazz:
L20n and webl10n (l10n.js) both require data-l10n-id attributes to be added to all localizable nodes. Their value is set to what ends up as key in the .l20n or .properties file.

The attribute is also very convenient for Pontoon, because it makes it really easy to detect in-page localizable nodes.

PR #143, opened 2015-08-31
Match source strings with all occurences in the DOM

mathjazz:
If we had multiple occurences of the same string in the DOM, we only
matched the last one.

Merging this one too to ship Donations form. Today is the day to break the rules I guess.

@Osmose


Osmose:
r+, this makes sense and helped me understand how pontoon.js works a little more. :D

PR #145, opened 2015-09-01
Bug 1179950: Enable timezone support.

Osmose:
This will require a config change to the Heroku apps before deploying them, namely setting the `TZ` environment variable to `UTC`.

I believe this will also fix bug 1199763.

@mathjazz r?


mathjazz:
I think you hardcoded TIME_ZONE in settings, otherwise it looks fine. r+wc

PR #146, opened 2015-09-02
Make a better icon for "No results"

yfdyh000:
We currently use the `fa-thumbs-down` for "No results", this gesture is not a friendly gesture.
I think the `fa-exclamation-circle` is a good choice.

Do I need to write a bug for the tiny patch?

### DIFF
--- a/pontoon/base/templates/translate.html
+++ b/pontoon/base/templates/translate.html
@@ -227,7 +227,7 @@
       <h2 id="not-on-page">Not on the current page</h2>
       {% endif %}
       <ul class="uneditables"></ul>
-      <h3 class="no-match"><div class="fa fa-thumbs-down"></div>No results</h3>
+      <h3 class="no-match"><div class="fa fa-exclamation-circle"></div>No results</h3>
     </div>
   </div>
 


mathjazz:
Thanks for the patch!

I agree `fa-thumbs-down` is not the best choice, however `fa-exclamation-circle` is very similar to `fa-exclamation-triangle` we use for strings that Need work.

What about `meh-o` or `frown-o`?

yfdyh000:
I hope users don't get the negative emotions because of this situation, so I feel these not a good choice.

I think the `fa-exclamation-circle` is enough to distinguish, through the size and shape.

mathjazz:
Don't you get negative emotions if you get No results regardless of the image? ;)

But I think I know what you mean, ideally we should use something that not only communicates "sad", but also "cute". Something like this: http://i.imgur.com/HtCOIDp.jpg

yfdyh000:
No, I just know I'm wrong, if the image is a neutral or prompt nature, and no regrets.

mathjazz:
OK, I'm giving this r+ and merging, will probably get into production tomorrow.

mathjazz:
Thank you for your contribution! ;)

PR #147, opened 2015-09-02
Remove chrome:// URLs from tiles CSS

mathjazz:
They prevent Pontoon from building in production by raising
whitenoise.django.MissingFileError.

@jotes r?


jotes:
Overall looks good for me, however, I'm not sure if browser don't require them.

mathjazz:
I tested the page and it looks exactly the same, so I guess it doesn't. Thanks!

PR #149, opened 2015-09-03
Bug 1178874: Add support for silme formats to sync_projects.

Osmose:
Adds support for ini, dtd, and properties files, including more format tests for them. This also switches us to using the silme fork at https://github.com/Osmose/silme/ instead of the HG-hosted version.

This also changes how we build entities in vcs_models; in order to support both asymmetric and symmetric formats, we first build entities based on the source resource files, and then load translations from each translated resource. 

I did manual testing for DTD files, but haven't yet manually tested ini and properties files. Suggestions on good projects to use for testing those are welcome, otherwise we can just test on staging before merging.

@mathjazz r?

### DIFF
--- a/pontoon/base/formats/silme.py
+++ b/pontoon/base/formats/silme.py
@@ -1,0 +1,161 @@
+from __future__ import absolute_import  # Same name as silme library.
+"""
+Parser for silme-compatible translation formats.
+"""
+import codecs
+from collections import OrderedDict
+
+import silme
+from silme.format.dtd import FormatParser as DTDParser
+from silme.format.ini import FormatParser as IniParser
+from silme.format.properties import FormatParser as PropertiesParser
+
+from pontoon.base import SyncError
+from pontoon.base.formats.base import ParsedResource
+from pontoon.base.vcs_models import VCSTranslation
+
+
+class SilmeEntity(VCSTranslation):
+    def __init__(self, silme_object, comments=None, order=0):
+        self.silme_object = silme_object
+        self.strings = {None: self.silme_object.value} if self.silme_object.value else {}
+        self.comments = comments or []
+        self.order = order
+        self.last_translator = None
+        self.last_update = None
+
+    @property
+    def key(self):
+        return self.silme_object.id
+
+    @property
+    def source_string(self):
+        return self.silme_object.value or ''
+
+    @property
+    def source_string_plural(self):
+        return ''
+
+    @property
+    def fuzzy(self):
+        return False
+
+    @fuzzy.setter
+    def fuzzy(self, fuzzy):
+        pass  # We don't use fuzzy in silme
+
+    @property
+    def source(self):
+        return []
+
+    def __eq__(self, other):
+        return self.key == other.key and self.strings.get(None) == other.strings.get(None)
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+
+class SilmeResource(ParsedResource):
+    def __init__(self, parser, path, source_resource=None):
+        self.parser = parser
+        self.path = path
+        self.source_resource = source_resource
+
+        # Preserve entity order via an OrderedDict.
+        if source_resource:
+            self.entities = source_resource.entities
+        else:
+            self.entities = OrderedDict()
+
+        with codecs.open(path, 'r', 'utf-8') as f:
+            self.structure = parser.get_structure(f.read())
+
+        comments = []
+        current_order = 0
+        for obj in self.structure:
+            if isinstance(obj, silme.core.entity.Entity):
+                entity = SilmeEntity(obj, comments, current_order)
+                self.entities[entity.key] = entity
+                current_order += 1
+                comments = []
+            elif isinstance(obj, silme.core.structure.Comment):
+                for comment in obj:
+                    # Silme groups comments together, so we strip
+                    # whitespace and split them up.
+                    lines = unicode(comment).strip().split('\n')
+                    comments += [line.strip() for line in lines]
+
+    @property
+    def translations(self):
+        return self.entities.values()
+
+    def save(self, locale):
+        """
+        Load the source resource, modify it with changes made to this
+        Resource instance, and save it over the locale-specific
+        resource.
+        """
+        if self.source_resource is None:
+            raise SyncError('Cannot save silme resource {0}: No source resource given.'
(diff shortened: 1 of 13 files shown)


pontoon/base/formats/silme.py
```diff
@@ -0,0 +1,161 @@
...
+                self.entities[entity.key] = entity
+                current_order += 1
+                comments = []
+            elif isinstance(obj, silme.core.structure.Comment):
+                for comment in obj:
+                    # Silme groups comments together, so we strip
+                    # whitespace and split them up.
+                    lines = unicode(comment).strip().split('\n')
+                    comments += [line.strip() for line in lines]
+
```
mathjazz: We should probably move this to Silme. One day.

pontoon/base/formats/silme.py
```diff
@@ -0,0 +1,161 @@
...
+                translated_entity
+                and None in translated_entity.strings
+                and translated_entity != silme_entity
+            )
+            if have_translation:
+                new_structure.modify_entity(key, translated_entity.strings[None])
+            else:
+                # Remove untranslated entity and following newline
+                pos = new_structure.entity_pos(key)
+                new_structure.remove_entity(key)
```
mathjazz: This line should be the only thing required, the rest should be moved to Silme. One day.

mathjazz:
Nice, seems like you also took care of bug https://bugzilla.mozilla.org/show_bug.cgi?id=1199489:
https://github.com/Osmose/pontoon/blob/silme/pontoon/base/formats/silme.py#L105

You will also need to ignore some files for some projects:
https://github.com/mozilla/pontoon/blob/master/pontoon/administration/files.py#L728
https://github.com/mozilla/pontoon/blob/master/pontoon/administration/files.py#L1235-L1238

And, obsolete files (removed from source directory) also need to get removed from locale directory:
https://github.com/mozilla/pontoon/blob/master/pontoon/administration/files.py#L1227-L1246

But I think that's currently not working anyways, so it can wait:
https://bugzilla.mozilla.org/show_bug.cgi?id=1201431

I tested the code locally with FHR, which is a PO & .properties project. I changed one string only (Jan -> Janvar) and this happened on sync:
http://viewvc.svn.mozilla.org/vc/projects/l10n-misc/trunk/firefoxhealthreport/locale/sl/dateFormat.properties?r1=115824&r2=145923&sortby=date

Osmose:
> I tested the code locally with FHR, which is a PO & .properties project. I changed one string only (Jan -> Janvar) and this happened on sync:
> http://viewvc.svn.mozilla.org/vc/projects/l10n-misc/trunk/firefoxhealthreport/locale/sl/dateFormat.properties?r1=115824&r2=145923&sortby=date

Just a note, you can pass the `--no-commit` flag to `sync_projects` to avoid actually committing the changes. That lets you check out the sync results locally without modifying the remote repo.

Osmose:
Updated with a fix for the removed strings mentioned above, including some extra tests for the cases that caused the issues.

Osmose:
> You will also need to ignore some files for some projects:
> https://github.com/mozilla/pontoon/blob/master/pontoon/administration/files.py#L728
> https://github.com/mozilla/pontoon/blob/master/pontoon/administration/files.py#L1235-L1238

That should be handled by https://github.com/mozilla/pontoon/blob/5c45a6b25acc236cc4162024c5a76ef22899f427/pontoon/base/models.py#L246, I think. Did you run into an issue during testing? How could I replicate?

There's still the issue of connecting to your special repo vs the normal repo though: https://github.com/mozilla/pontoon/blob/5c45a6b25acc236cc4162024c5a76ef22899f427/pontoon/administration/files.py#L770

I need a refresher on how that works to figure out how I'd like to handle it with the new sync process.

Osmose:
> There's still the issue of connecting to your special repo vs the normal repo though: https://github.com/mozilla/pontoon/blob/5c45a6b25acc236cc4162024c5a76ef22899f427/pontoon/administration/files.py#L770
> 
> I need a refresher on how that works to figure out how I'd like to handle it with the new sync process.

I'll file a separate bug for this issue to keep this PR slim.

mathjazz:
I continued testing FHR and your fix did the trick.

But I found a weird bug when I updated a translation in SVN and then synced. A new file showed up among resources;
http://viewvc.svn.mozilla.org/vc/projects/l10n-misc/trunk/firefoxhealthreport/locale/en-US/date.lang?view=markup&sortby=file

I guess the old script we still use when adding project didn't detect it because it looks a bit weird (probably not valid .lang file). And it was actually OK that it wasn't detected, beause it looks like converted dateFormat.properties file.

Also, only 34/47 strings were translated for sl:
https://etherpad.mozilla.org/pontoon-silme

Although sl file looks fully translated:
http://viewvc.svn.mozilla.org/vc/projects/l10n-misc/trunk/firefoxhealthreport/locale/sl/date.lang?view=markup&sortby=file

I think we should change our .lang parser to ignore such files. But it would be interesting to figure out why not all translations were imported.

I'll keep testing with Firefox OS.

mathjazz:
Firefox OS, Pontoon Intro and the Donations Form (or should I rename it to Donation form?) all work fine! Great job! So there's only FHR edge case remaining.

Osmose:
> But I found a weird bug when I updated a translation in SVN and then synced. A new file showed up among resources;
> http://viewvc.svn.mozilla.org/vc/projects/l10n-misc/trunk/firefoxhealthreport/locale/en-US/date.lang?view=markup&sortby=file
> 
> I guess the old script we still use when adding project didn't detect it because it looks a bit weird (probably not valid .lang file). And it was actually OK that it wasn't detected, beause it looks like converted dateFormat.properties file.

Filed as https://bugzilla.mozilla.org/show_bug.cgi?id=1201936, it's separate from silme.

mathjazz:
In this case: r+!

Osmose:
> Also, only 34/47 strings were translated for sl:
> https://etherpad.mozilla.org/pontoon-silme
> 
> Although sl file looks fully translated:
> http://viewvc.svn.mozilla.org/vc/projects/l10n-misc/trunk/firefoxhealthreport/locale/sl/date.lang?view=markup&sortby=file

The translated strings are missing `{ok}` tags to mark them as unchanged but translated. Hence, they're marked as untranslated. Saving the translations and syncing again should add the tags in.

PR #153, opened 2015-09-03
Fix pluralized string length calculation

mathjazz:
@Osmose r?


pontoon/base/static/js/translate_old.js
```diff
@@ -270,7 +270,6 @@ var Pontoon = (function (my) {
       }
 
       // Length
-      console.log(this.isPluralized());
```
Osmose: Thaaaaaaaanks

Osmose:
r+

PR #156, opened 2015-09-04
Properly detect in-place localization mode

mathjazz:
We might have projects with both, subpages with page_urls and resource
files without in-place localization.

@Osmose r?


Osmose:
r+ :D

PR #157, opened 2015-09-04
Subpage names don't need to match resource paths anymore

mathjazz:
If we wanted to match subpage against specific resource(s), its name
had to be exactly the same as the path to the resource file. Which is
lame. This is no loner required. As a consequence, we can now have
several different subpages matched agains the same resource.

@Osmose r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -429,6 +429,11 @@ class Entity(DirtyFieldsMixin, models.Model):
         )
 
         if paths:
+            try:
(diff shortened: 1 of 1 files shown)


Osmose:
TESSSSSTSSS

r+wc :+1:

mathjazz:
Filing bug to write tests and deploying this because it's important:
https://bugzilla.mozilla.org/show_bug.cgi?id=1202000

PR #159, opened 2015-09-04
Improve plurals

mathjazz:
This changes a few things related to plurals:

1. Plural forms are displayed as CLDR values now (one, two, many...) with supscripted example number, calculated from plural rule.

2. For that to work, I had to add missing plural data for all locales. I also removed two unused (duplicate) locales and nplurals field, which can be calculated from the CLDR plurals list length.

@Osmose r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -72,7 +72,6 @@ def validate_cldr(value):
 class Locale(models.Model):
     code = models.CharField(max_length=20, unique=True)
     name = models.CharField(max_length=128)
-    nplurals = models.SmallIntegerField(null=True, blank=True)
     plural_rule = models.CharField(max_length=128, blank=True)
 
     # CLDR Plurals
@@ -94,6 +93,16 @@ class Locale(models.Model):
         else:
             return map(int, self.cldr_plurals.split(','))
 
+    @classmethod
+    def cldr_plural_to_id(self, cldr_plural):
+        for i in self.CLDR_PLURALS:
+            if i[1] == cldr_plural:
+                return i[0]
+
+    @property
+    def nplurals(self):
+        return len(self.cldr_plurals_list())
+
     def __unicode__(self):
         return self.name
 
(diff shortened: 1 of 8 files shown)


pontoon/base/models.py
```diff
@@ -94,6 +93,16 @@ def cldr_plurals_list(self):
...
             return map(int, self.cldr_plurals.split(','))
 
+    @classmethod
+    def cldr_plural_to_id(self, cldr_plural):
+        for i in self.CLDR_PLURALS:
+            if i[1] == cldr_plural:
+                return i[0]
+
+    @property
+    def nplurals(self):
```
Osmose: This is great!

pontoon/base/models.py
```diff
@@ -94,6 +93,16 @@ def cldr_plurals_list(self):
         else:
             return map(int, self.cldr_plurals.split(','))
 
+    @classmethod
+    def cldr_plural_to_id(self, cldr_plural):
+        for i in self.CLDR_PLURALS:
```
Osmose: The only other Locale.CLDR_PLURALS used is in `validate_cldr`, and the way it's used there and here could be simplified if you made it a dict instead of a list. Then you could just use `.get()` to fetch the string and `in` to check if the integer is valid.

pontoon/base/static/js/translate_old.js
```diff
@@ -236,16 +236,16 @@ var Pontoon = (function (my) {
...
 
             if (nplurals === 2) {
-              examples[0] = 1;
-              examples[1] = 2;
+              examples = {0: 1, 1: 2};
+
             } else {
               while (Object.keys(examples).length < nplurals) {
-                rule = eval(this.locale.plural_rule);
+                var rule = eval(this.locale.plural_rule);
```
Osmose: Someday (not today) we should figure out a way to do this without `eval`.

pontoon/base/static/js/translate_old.js
```diff
@@ -253,9 +253,10 @@ var Pontoon = (function (my) {
               }
             }
 
-            $('#plural-tabs li a span').each(function(i) {
-              $(this).html(examples[i]);
-              return i < nplurals-1;
+            $.each(this.locale.cldr_plurals, function(i) {
+              $('#plural-tabs li:eq(' + i + ') a')
+                .find('span').html(CLDR_PLURALS[this]).end()
```
Osmose: Woah, I've never seen `end()` before. Cool!

Osmose:
If you could add some tests for the methods on the locale class it'd be nice, but I wouldn't call it required here.

r+, looks great! Really nice improvement.

PR #161, opened 2015-09-07
Bug 1196140. Added possibility to see top contributors aggregated by last 12|6|3 months.

jotes:
@mathjazz Hi, could you review?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -8,7 +8,7 @@ from django.conf import settings
 from django.contrib.auth.models import User
 from django.core.exceptions import ValidationError
 from django.db import models
-from django.db.models import Sum, Prefetch
+from django.db.models import Sum, Prefetch, F, Count, Q, Case, When
 from django.db.models.signals import post_save
 from django.forms import ModelForm
 from django.templatetags.static import static
@@ -23,12 +23,70 @@ from pontoon.base import utils
 log = logging.getLogger('pontoon')
 
 
+class UserTranslationsManager(models.Manager):
+    """
+    Provides various method to interact with larger sets of translations and their stats for user.
+    """
+
+    def _changed_translations_count(self, *args):
+        """
+        Helper method, returns expression object which allows us to annotate querysets
+        with counts of translations.
+        """
+        translation_query = (
+            ~Q(translation__string=F('translation__entity__string'))
+            & ~Q(translation__string=F('translation__entity__string_plural'))
+            & Q(translation__user__isnull=False)
+        )
+        for arg in args:
+            translation_query &= arg
+
+        # For each translation that matches the filter, return 1. Aggregate
+        # the sum of all those results to count the number of matching
+        # translations.
+        return Sum(
+            Case(
+                When(translation_query, then=1), output_field=models.IntegerField(), default=0))
+
+    def with_translation_counts(self, start_date=None, limit=100):
+        """
+        Returns contributors list, sorted by count of their translations.
+        Every user instance has added following properties:
+        * translations_count
+        * translations_approved_count
+        * translations_unapproved_count
+        * translations_needs_work_count
+        Method has been created mainly to improve performance and to optimize
+        count of sql queries during generation of metrics.
+        All counts will be returned from start_date to now().
+        :param date start_date: start date for translations.
+        """
+        def translations_count(query=None):
+            """Short helper to avoid duplication of passing dates."""
+            query = query or Q()
+            if start_date:
+                query &= Q(translation__date__gte=start_date)
+            return self._changed_translations_count(query)
+        return (
+            self
+            .exclude(email__in=settings.EXCLUDE)
+            .annotate(translations_count=translations_count(),
+                      translations_approved_count=translations_count(Q(translation__approved=True)),
+                      translations_unapproved_count=translations_count(Q(translation__approved=False, translation__fuzzy=False)),
+                      translations_needs_work_count=translations_count(Q(translation__fuzzy=True)))
+            .exclude(translations_count=0)
+            .distinct().order_by('-translations_count')[:limit]
+        )
+
+
 # User class extensions
 @property
 def user_display_name(self):
     name = self.first_name or self.email.split('@')[0]
     return u'{name} <{email}>'.format(name=name, email=self.email)
+
 User.add_to_class('display_name', user_display_name)
+User.add_to_class('translators', UserTranslationsManager())
 
 
 def user_gravatar_url(self, size):
--- a/pontoon/base/static/css/style.css
+++ b/pontoon/base/static/css/style.css
@@ -915,6 +915,19 @@ body > nav,
   color: #7BC876;
 }
 
+.contributors .links {
+  margin-top: 50px;
+}
+
+.contributors .links a {
+  padding: 10px;
+  border-radius: 3px;
+}
+
+.contributors .links li.active a {
+  background-color: #3F4752;
+}
+
 noscript p {
   text-transform: uppercase;
 }
--- a/pontoon/base/templates/users.html
+++ b/pontoon/base/templates/users.html
@@ -7,6 +7,25 @@
 {% block subtitle %}Top Contributors{% endblock %}
 
 {% block middle %}
+<div class="contributors">
+  <ul class="links">
+    <li {% if not period %}class="active"{% endif %}>
+      <a href="{{ url('pontoon.contributors') }}">All time</a>
+    </li>
+    <li {% if period == 12 %}class="active"{% endif %}>
+      <a href="{{ url('pontoon.contributors') }}?period=12">12 months</a>
+    </li>
+    <li {% if period == 6 %}class="active"{% endif %}>
+      <a href="{{ url('pontoon.contributors') }}?period=6">6 months</a>
+    </li>
+    <li {% if period == 3 %}class="active"{% endif %}>
+      <a href="{{ url('pontoon.contributors') }}?period=3">3 months</a>
+    </li>
+    <li {% if period == 1 %}class="active"{% endif %}>
+      <a href="{{ url('pontoon.contributors') }}?period=1">Last month</a>
+    </li>
+  </ul>
+</div>
 <form>
   <table>
     <thead>
@@ -31,16 +50,16 @@
           <div class="details">
             <div class="total">
               <span>Total</span>
-              <p>{{ contributor.translations.count() }}</p>
+              <p>{{ contributor.translations_count }}</p>
             </div><div class="approved">
               <span>Approved</span>
-              <p>{{ contributor.translations.filter(approved=True).count() }}</p>
+              <p>{{ contributor.translations_approved_count }}</p>
             </div><div class="translated">
               <span>Unapproved</span>
-              <p>{{ contributor.translations.exclude(approved=True).exclude(fuzzy=True).count() }}</p>
+              <p>{{ contributor.translations_unapproved_count }}</p>
             </div><div class="fuzzy">
               <span>Needs work</span>
-              <p>{{ contributor.translations.filter(fuzzy=True).count() }}</p>
+              <p>{{ contributor.translations_needs_work_count }}</p>
             </div>
           </div>
         </td>
--- a/pontoon/base/tests/__init__.py
+++ b/pontoon/base/tests/__init__.py
@@ -1,3 +1,6 @@
+from datetime import datetime, timedelta
+from itertools import cycle
+
 import factory
 from factory import LazyAttribute, Sequence, SubFactory
 from factory.django import DjangoModelFactory
@@ -5,6 +8,7 @@ from factory.django import DjangoModelFactory
 from django.contrib.auth.models import User
 from django.template.defaultfilters import slugify
 from django.test import TestCase as BaseTestCase
+from django.utils import timezone
 
 from django_browserid.tests import mock_browserid
 from django_nose.tools import assert_equal
@@ -100,9 +104,14 @@ class TranslationFactory(DjangoModelFactory):
     locale = SubFactory(LocaleFactory)
     string = Sequence(lambda n: 'translation {0}'.format(n))
     user = SubFactory(UserFactory)
-
     class Meta:
         model = Translation
+
+
+class IdenticalTranslationFactory(TranslationFactory):
+    @factory.sequence
+    def entity(n):
+        return EntityFactory.create(string='translation {}'.format(n))
 
 
 class StatsFactory(DjangoModelFactory):
--- a/pontoon/base/tests/__init__.py
+++ b/pontoon/base/tests/__init__.py
@@ -150,6 +150,8 @@ def assert_attributes_equal(original, **expected_attrs):
     Assert that the given object has attributes matching the given
     values.
     """
+    if not expected_attrs:
+        raise ValueError('Expected some attributes to check.')
     for key, value in expected_attrs.items():
         original_value = getattr(original, key)
         assert_equal(
--- a/pontoon/base/tests/__init__.py
+++ b/pontoon/base/tests/__init__.py
@@ -2,7 +2,7 @@ from datetime import datetime, timedelta
 from itertools import cycle
 
 import factory
-from factory import LazyAttribute, Sequence, SubFactory
+from factory import LazyAttribute, Sequence, SubFactory, SelfAttribute
 from factory.django import DjangoModelFactory
 
 from django.contrib.auth.models import User
@@ -109,9 +109,7 @@ class TranslationFactory(DjangoModelFactory):
 
 
 class IdenticalTranslationFactory(TranslationFactory):
-    @factory.sequence
-    def entity(n):
-        return EntityFactory.create(string='translation {}'.format(n))
+    entity = SubFactory(EntityFactory, string=SelfAttribute('..string'))
 
 
 class StatsFactory(DjangoModelFactory):
--- a/pontoon/base/tests/test_models.py
+++ b/pontoon/base/tests/test_models.py
@@ -1,8 +1,11 @@
-from django_nose.tools import assert_equal
+from datetime import datetime
+
+from django_nose.tools import assert_equal, assert_true
 from django.test.utils import override_settings
 
-from pontoon.base.models import Translation
+from pontoon.base.models import Translation, User
 from pontoon.base.tests import (
+    assert_attributes_equal,
     IdenticalTranslationFactory,
     TranslationFactory,
     UserFactory,
@@ -68,7 +71,7 @@ class UserTranslationManagerTests(TestCase):
 
     def test_users_without_translations(self):
         """
-        Checks if user contributors without translations are returned.
+        Checks if user contributors without translations aren't returned.
         """
         active_contributor = TranslationFactory.create(user__email='active@example.com').user
         inactive_contributor = UserFactory.create(email='inactive@example.com')
@@ -80,7 +83,7 @@ class UserTranslationManagerTests(TestCase):
     def test_unique_translations(self):
         """
         Checks if contributors with identical translations are returned.
-            """
+        """
 
         unique_translator = TranslationFactory.create().user
         identical_translator = IdenticalTranslationFactory.create().user
@@ -94,14 +97,20 @@ class UserTranslationManagerTests(TestCase):
         """
         Checks if users are ordered by count of contributions.
         """
-        users = UserFactory.create_batch(5)
+        contributors = [
+            self.create_contributor_with_translation_counts(2),
+            self.create_contributor_with_translation_counts(4),
+            self.create_contributor_with_translation_counts(9),
+            self.create_contributor_with_translation_counts(1),
+            self.create_contributor_with_translation_counts(6),
+        ]
 
-        for i, user in enumerate(users):
-            TranslationFactory.create_batch(5-i, user=user)
-
-        top_translations_totals = [user.translations_count for user in User.translators.with_translation_counts()]
-        assert_equal(top_translations_totals, [5, 4, 3, 2, 1])
-
+        assert_equal(list(User.translators.with_translation_counts()), [
+            contributors[2],
+            contributors[4],
+            contributors[1],
+            contributors[0],
+            contributors[3]])
 
     def test_contributors_limit(self):
         """
@@ -111,9 +120,9 @@ class UserTranslationManagerTests(TestCase):
 
         top_contributors = User.translators.with_translation_counts()
 
-        assert_true(len(top_contributors) == 100)
+        assert_equal(top_contributors.count(), 100)
 
-    def create_contributor_with_translation_counts(self, approved, unapproved, needs_work, **kwargs):
+    def create_contributor_with_translation_counts(self, approved=0, unapproved=0, needs_work=0, **kwargs):
         """
         Helper method, creates contributor with given translations counts.
         """
@@ -135,24 +144,27 @@ class UserTranslationManagerTests(TestCase):
 
         top_contributors = User.translators.with_translation_counts()
 
-        assert_equal(top_contributors.count(), 3)
-        assert_equal(top_contributors[0], second_contributor)
-        assert_equal(top_contributors[0].translations_count, 16)
-        assert_equal(top_contributors[0].translations_approved_count, 5)
-        assert_equal(top_contributors[0].translations_unapproved_count, 9)
-        assert_equal(top_contributors[0].translations_needs_work_count, 2)
+        assert_attributes_equal({
+            top_contributors.count(): 3,
 
-        assert_equal(top_contributors[1], first_contributor)
-        assert_equal(top_contributors[1].translations_count, 12)
-        assert_equal(top_contributors[1].translations_approved_count, 7)
-        assert_equal(top_contributors[1].translations_unapproved_count, 3)
-        assert_equal(top_contributors[1].translations_needs_work_count, 2)
+            top_contributors[0]: second_contributor,
+            top_contributors[0].translations_count: 16,
+            top_contributors[0].translations_approved_count: 5,
+            top_contributors[0].translations_unapproved_count: 9,
+            top_contributors[0].translations_needs_work_count: 2,
 
-        assert_equal(top_contributors[2], third_contributor)
-        assert_equal(top_contributors[2].translations_count, 8)
-        assert_equal(top_contributors[2].translations_approved_count, 1)
-        assert_equal(top_contributors[2].translations_unapproved_count, 2)
-        assert_equal(top_contributors[2].translations_needs_work_count, 5)
+            top_contributors[1]: first_contributor,
+            top_contributors[1].translations_count: 12,
+            top_contributors[1].translations_approved_count: 7,
+            top_contributors[1].translations_unapproved_count: 3,
+            top_contributors[1].translations_needs_work_count: 2,
+
+            top_contributors[2]: third_contributor,
+            top_contributors[2].translations_count: 8,
+            top_contributors[2].translations_approved_count: 1,
+            top_contributors[2].translations_unapproved_count: 2,
+            top_contributors[2].translations_needs_work_count: 5,
+        })
 
 
     def test_period_filters(self):
@@ -170,39 +182,48 @@ class UserTranslationManagerTests(TestCase):
 
         top_contributors = User.translators.with_translation_counts(datetime(2015, 6, 10))
 
-        assert_equal(top_contributors.count(), 1)
-        assert_equal(top_contributors[0], first_contributor)
-        assert_equal(top_contributors[0].translations_count, 5)
-        assert_equal(top_contributors[0].translations_approved_count, 5)
-        assert_equal(top_contributors[0].translations_unapproved_count, 0)
-        assert_equal(top_contributors[0].translations_needs_work_count, 0)
+        assert_attributes_equal({
+            top_contributors.count(): 1,
+
+            top_contributors[0]: first_contributor,
+            top_contributors[0].translations_count: 5,
+            top_contributors[0].translations_approved_count: 5,
+            top_contributors[0].translations_unapproved_count: 0,
+            top_contributors[0].translations_needs_work_count: 0,
+        })
 
         top_contributors = User.translators.with_translation_counts(datetime(2015, 5, 10))
 
-        assert_equal(top_contributors.count(), 2)
-        assert_equal(top_contributors[0], second_contributor)
-        assert_equal(top_contributors[0].translations_count, 15)
-        assert_equal(top_contributors[0].translations_approved_count, 2)
-        assert_equal(top_contributors[0].translations_unapproved_count, 11)
-        assert_equal(top_contributors[0].translations_needs_work_count, 2)
+        assert_attributes_equal({
+            top_contributors.count(): 2,
 
-        assert_equal(top_contributors[1], first_contributor)
-        assert_equal(top_contributors[1].translations_count, 5)
-        assert_equal(top_contributors[1].translations_approved_count, 5)
-        assert_equal(top_contributors[1].translations_unapproved_count, 0)
-        assert_equal(top_contributors[1].translations_needs_work_count, 0)
+            top_contributors[0]: second_contributor,
+            top_contributors[0].translations_count: 15,
+            top_contributors[0].translations_approved_count: 2,
+            top_contributors[0].translations_unapproved_count: 11,
+            top_contributors[0].translations_needs_work_count: 2,
+
+            top_contributors[1]: first_contributor,
+            top_contributors[1].translations_count: 5,
+            top_contributors[1].translations_approved_count: 5,
+            top_contributors[1].translations_unapproved_count: 0,
+            top_contributors[1].translations_needs_work_count: 0,
+        })
 
         top_contributors = User.translators.with_translation_counts(datetime(2015, 1, 10))
 
-        assert_equal(top_contributors.count(), 2)
-        assert_equal(top_contributors[0], first_contributor)
-        assert_equal(top_contributors[0].translations_count, 20)
-        assert_equal(top_contributors[0].translations_approved_count, 17)
-        assert_equal(top_contributors[0].translations_unapproved_count, 1)
-        assert_equal(top_contributors[0].translations_needs_work_count, 2)
+        assert_attributes_equal({
+            top_contributors.count(): 2,
 
-        assert_equal(top_contributors[1], second_contributor)
-        assert_equal(top_contributors[1].translations_count, 15)
-        assert_equal(top_contributors[1].translations_approved_count, 2)
-        assert_equal(top_contributors[1].translations_unapproved_count, 11)
-        assert_equal(top_contributors[1].translations_needs_work_count, 2)
+            top_contributors[0]: first_contributor,
+            top_contributors[0].translations_count: 20,
+            top_contributors[0].translations_approved_count: 17,
+            top_contributors[0].translations_unapproved_count: 1,
+            top_contributors[0].translations_needs_work_count: 2,
+
+            top_contributors[1]: second_contributor,
+            top_contributors[1].translations_count: 15,
+            top_contributors[1].translations_approved_count: 2,
+            top_contributors[1].translations_unapproved_count: 11,
+            top_contributors[1].translations_needs_work_count: 2,
+        })
--- a/pontoon/base/tests/test_models.py
+++ b/pontoon/base/tests/test_models.py
@@ -1,5 +1,3 @@
-from datetime import datetime
-
 from django_nose.tools import assert_equal, assert_true
 from django.test.utils import override_settings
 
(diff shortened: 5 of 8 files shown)


pontoon/base/static/css/style.css
```diff
@@ -884,6 +884,20 @@ body > nav,
   color: #7BC876;
 }
 
+.contributors .links {
+  margin-top: 20px;
```
mathjazz: Let's make it 50px, so it can breath.

pontoon/base/static/css/style.css
```diff
@@ -884,6 +884,20 @@ body > nav,
...
+}
+
+.contributors .links a {
+  padding:10px;
```
mathjazz: Nit - missing space after colon.

pontoon/base/static/css/style.css
```diff
@@ -884,6 +884,20 @@ body > nav,
...
+  border-radius: 25px;
```
mathjazz: Let's make it 3px radius to match other "buttons".

pontoon/base/urls.py
```diff
@@ -60,7 +60,7 @@
         name='pontoon.translate'),
 
     # List contributors
-    url(r'^contributors/$',
+    url(r'^contributors/(?P<period>overall|12months|6months|3months|lastmonth)$',
```
mathjazz: Let's use alltime instead of overall to match the menu entry.

Also, I think we should keep the old link (/contributors) to avoid breaking any (external) links. It should be redirected to /contributors/alltime

mathjazz:
Nice work!

I'd also like to ask @Osmose to have a look at the view (and the PR in general). IIRC we talked about optimizing DB queries. Also, I noticed a RuntimeWarning in a console (DateTimeField Translation.date received a naive datetime), which requires us to use time zone aware dates.

pontoon/base/templates/users.html
```diff
@@ -31,16 +50,16 @@
           <div class="details">
             <div class="total">
               <span>Total</span>
-              <p>{{ contributor.translations.count() }}</p>
+              <p>{{ contributor.translations_total }}</p>
```
Osmose: Moving queryset logic out of the template is _always_ a good thing. :D

pontoon/base/urls.py
Osmose: I think we should not include these in the base URL. The _thing_ that users are accessing is the same (a list of contributors), it's just limited to a certain period of time. 

It'd be better, I think, to use a querystring parameter, and keep this URL as it is. The querystring parameter can default to the `overall` view so that we don't have to issue a redirect.

pontoon/base/views.py
```diff
@@ -364,20 +367,80 @@ def contributors(request, template='users.html'):
...
+    if period == '12months':
+        start_date = end_date - relativedelta(month=+12)
+    elif period == '6months':
+        start_date = end_date - relativedelta(month=+6)
+    elif period == '3months':
+        start_date = end_date - relativedelta(month=+3)
+    elif period == 'lastmonth':
+        start_date = end_date - relativedelta(month=+1)
+
+    if start_date and end_date:
```
Osmose: If you filter the translation queryset before building the separate querysets and users list you can drop this block down to like 4 lines and avoid a bunch of extra filtering.

pontoon/base/views.py
```diff
@@ -364,20 +367,80 @@ def contributors(request, template='users.html'):
...
+    translations_approved_map = translation_counts(translations_approved)
+    translations_unapproved_map = translation_counts(translations_unapproved)
+    translations_needs_work_map = translation_counts(translations_needs_work)
+    contributors = list(users.order_by('-translation_count')[:100])
+
+    for user in contributors:
+        user.translations_total = translations_total_map.get(user.pk, 0)
+        user.translations_approved = translations_approved_map.get(user.pk, 0)
+        user.translations_unapproved = translations_unapproved_map.get(user.pk, 0)
+        user.translations_needs_work = translations_needs_work_map.get(user.pk, 0)
```
Osmose: A good rule of thumb is that if you're ever manually assigning values to an instance of a class, you should consider if it instead could be a property on that class. In this case, it makes a whole lot of sense to just make `User.translation_count`, `User.approved_translation_count`, etc. as properties on the `User` class.

(There's some other options that actually might be better, like a custom translation manager, but those are kinda complex and have some nasty edge cases that we probably just don't want to bother with.)

This is probably a good time for us to set up a [custom User model](https://docs.djangoproject.com/en/1.8/topics/auth/customizing/#specifying-a-custom-user-model) for this and other purposes.

pontoon/base/views.py
jotes: okay!

Osmose:
So the problem we're facing here is that Django doesn't provide an easy way of saying "annotate my results with these aggregates (IE the count of translations), but when calculating that count, filter the translations you're counting like so.

Turns out, there is a solution to this, called [conditional expressions](https://docs.djangoproject.com/en/1.8/ref/models/conditional-expressions/). I just learned about them today!

I have some code that fetches the correct data, but it's kind of complex:

```python
def filtered_translation_count(**kwargs):
    """
    Return an aggregate object that counts the number of related 
    translations that match the given queryset filters.
    """
    # Filter out translations that are identical to their source.
    translation_filter = (
        ~Q(translation__string=F('translation__entity__string'))
        & ~Q(translation__string=F('translation__entity__string_plural'))
    )
    if kwargs:
        translation_filter &= Q(**kwargs)
    # For each translation that matches the filter, return 1. Aggregate
    # the sum of all those results to count the number of matching
    # translations.
    return Sum(
        Case(When(translation_filter, then=1), 
             output_field=IntegerField(), default=0)
    )

users = (
    User.objects
    .annotate(
        translation_count=filtered_translation_count(),
        approved_translation_count=filtered_translation_count(
            translation__approved=True
        ),
        unapproved_translation_count=filtered_translation_count(
            translation__approved=False,
            translation__fuzzy=False
        ),
        needs_work_translation_count=filtered_translation_count(
            translation__fuzzy=True
        ),
    )
    .exclude(translation_count=0)
    .order_by('-translation_count')[:100]
)
```

I haven't measured the number of queries yet but my expectation is that it will be fairly low. I ran it in a shell on prod to see how it performs and it's practically instantaneous. While `filtered_translation_count` is a bit hard to understand, it makes the view itself incredibly easy to understand and simplifies a lot of the logic.

Sorry for just dumping my own code on your PR, by the way. Your method works fine, too, I just wanted to see if there was a better way.

Thoughts on trying to integrate this?

Osmose:
Oh, also, the example above doesn't take into account filtering by date range, but that can probably be a parameter to `filtered_translation_count`.

Also also, some unit tests for the new behavior would be lovely. :D

All in all, this is nice work and a great start. Sorry for getting carried away.

jotes:
@osmose looks like we both learned new things today :D conditonal
expressions are looking pretty promising, i'll look into them.

I've refactored code already, could you review diff? I'll change methods in
manager to reflect your suggestions about conditonal expressions in next
batch of changes. I'll definetely will add some tests.

Thanks for review :)

On Wed, Sep 9, 2015 at 12:35 AM, Michael Kelly <notifications@github.com>
wrote:

> Oh, also, the example above doesn't take into account filtering by date
> range, but that can probably be a parameter to filtered_translation_count.
>
> Also also, some unit tests for the new behavior would be lovely. :D
>
> All in all, this is nice work and a great start. Sorry for getting carried
> away.
>
> —
> <https://github.com/mozilla/pontoon/pull/161#issuecomment-138722851>.
>



-- 
jarek@reijutsu:~$ fortune
jarek@reijutsu:~$ fortune
You now have Asian Flu.

pontoon/base/models.py
```diff
@@ -25,12 +25,77 @@
 log = logging.getLogger('pontoon')
 
 
+class UserTranslationsManager(models.Manager):
```
Osmose: Oh, I like the idea of having a separate manager for this! I've never actually tried using specialized managers like this before so I'm interested in checking it out.

pontoon/base/models.py
```diff
@@ -25,12 +25,77 @@
...
+        * translations_total
+        * translations_approved
+        * translations_unapproved
+        * translations_needs_work
+        Method has been created mainly to improve performance and to optimize
+        count of sql queries during generation of metrics.
+        """
+
+        users = (
+            User.objects
```
Osmose: Since this is a manager, you should probably use `self` here.

Osmose:
> I've refactored code already, could you review diff?

I like most of the changes you've made. I'm really excited to see how you work the conditional expressions in. :D

jotes:
hi @mathjazz @Osmose, could you review again? I've implemented your notes. If everything is okay then only thing that is left is tests for this view -> i'll try to implement them today and pass for your review.

pontoon/base/models.py
```diff
@@ -25,12 +25,103 @@
...
+        e.g.
+        _changed_translations_counts(exclude__approved=True)
+        Will exclude "approved" field with True value.
+        """
+        translation_query = (
+            ~Q(translation__string=F('translation__entity__string'))
+            & ~Q(translation__string=F('translation__entity__string'))
+        )
+
+
```
mathjazz: Remove extra line.

pontoon/base/models.py
```diff
@@ -25,12 +25,103 @@
...
+        if filter:
+            translation_query &= Q(**filter)
+
+        if exclude:
+            translation_query &= ~Q(**exclude)
+
+        if start_date and end_date:
+            translation_query &= Q(translation__date__range=(start_date, end_date))
+
+
```
mathjazz: Remove extra line.

pontoon/base/static/css/style.css
```diff
@@ -884,6 +884,20 @@ body > nav,
   color: #7BC876;
 }
 
+.contributors .links {
+  margin-top: 50px;
+}
+
+.contributors .links a {
+  padding:10px;
```
mathjazz: Add space after colon.

pontoon/base/static/css/style.css
```diff
@@ -884,6 +884,20 @@ body > nav,
...
+  border-radius: 3px;
+}
+
+.contributors .links li.active a {
+  background-color: #3F4752;
+}
+
+
```
mathjazz: Remove extra line.

pontoon/base/urls.py
```diff
@@ -63,7 +63,7 @@
     url(r'^contributors/$',
         views.contributors,
         name='pontoon.contributors'),
-
+
```
mathjazz: Remove extra tab.

mathjazz:
Nice work!

I've commented on a few nits, I hope that doesn't make you hate me. ;)

I have a question: isn't end_date always supposed to be now()? Because we're looking 1/3/6/12/all months back from *now*, right? So you could use __gte instead of __range and even get rid of py-dateutil.

Also, seems like logic doesn't work OK: locally I get results for the last 1/3/6/all months, but I get no results for the last 12 months.

jotes:
@MathJazz -> yeah, i've spot this bug after i've posted comment. Will fix
it today.
Don't worry about nits ;-) it's important to keep code clean for everyone.

On Thu, Sep 10, 2015 at 10:31 AM, Matjaž Horvat <notifications@github.com>
wrote:

> Nice work!
>
> I've commented on a few nits, I hope that doesn't make you hate me. ;)
>
> I have a question: isn't end_date always supposed to be now()? Because
> we're looking 1/3/6/12/all months back from *now*, right? So you could
> use __gte instead of __range and even get rid of py-dateutil.
>
> Also, seems like logic doesn't work OK: locally I get results for the last
> 1/3/6/all months, but I get no results for the last 12 months.
>
> —
> <https://github.com/mozilla/pontoon/pull/161#issuecomment-139166522>.
>



-- 
jarek@reijutsu:~$ fortune
jarek@reijutsu:~$ fortune
You now have Asian Flu.

pontoon/base/models.py
```diff
@@ -25,12 +25,101 @@
...
+        Method has been created mainly to improve performance and to optimize
+        count of sql queries during generation of metrics.
+        """
+        def translations_counts(filter=None, exclude=None):
+            """Short helper to avoid duplication of passing dates."""
+            return self._changed_translations_counts(start_date, end_date, filter=filter, exclude=exclude)
+
+        contributors = (
+            self
+            .filter(translation__in=self.get_changed_translations(start_date, end_date)).distinct()
```
Osmose: This is unnecessary because the filters in `_changed_translations_counts` already filter out unchanged translations. You can remove this line and the entire `get_changed_translations` method.

pontoon/base/models.py
```diff
@@ -25,12 +25,101 @@
...
+            translations = translations.filter(date__range=(start_date, end_date))
+        return translations
+
+    def get_approved(self, start_date=None, end_date=None):
+        return self.get_changed_translations(start_date, end_date).filter(approved=True)
+
+    def get_unapproved(self, start_date=None, end_date=None):
+        return self.get_changed_translations(start_date, end_date).exclude(approved=True, fuzzy=True)
+
+    def get_needs_work(self, start_date=None, end_date=None):
```
Osmose: These methods can be removed too, they're not used anywhere.

pontoon/base/models.py
```diff
@@ -25,12 +25,101 @@
...
+            translation_query &= ~Q(**exclude)
+
+        if start_date and end_date:
+            translation_query &= Q(translation__date__range=(start_date, end_date))
+
+        # For each translation that matches the filter, return 1. Aggregate
+        # the sum of all those results to count the number of matching
+        # translations.
+        return Sum(
+                Case(
```
Osmose: This is nitpicking, but the indentation in this block is really inconsistent. Probably stick to 4-space indents unless a hanging indent lined up with something would work better.

pontoon/base/models.py
```diff
@@ -25,12 +25,101 @@
...
+class UserTranslationsManager(models.Manager):
+    """
+    Provides various method to interact with larger sets of translations and their stats for user.
+    """
+
+    def _changed_translations_counts(self, start_date=None, end_date=None, filter=None, exclude=None):
+        """
+        Helper method, returns expression object which allows us to annotate querysets
+        with counts of translations.
+        Method supports exclusion of fields, you can exclude field if param contains
```
Osmose: This is incorrect. Also, instead of having `exclude` and `include` parameters, maybe just accept `Q` objects as `*args` and allow the caller to determine whether they include or exclude things? It'd simplify this function a bit.

pontoon/base/models.py
```diff
@@ -25,12 +25,101 @@
...
+
+
 # User class extensions
 @property
 def user_display_name(self):
     name = self.first_name or self.email.split('@')[0]
     return u'{name} <{email}>'.format(name=name, email=self.email)
+
 User.add_to_class('display_name', user_display_name)
+User.add_to_class('translations', UserTranslationsManager())
```
Osmose: `translations` isn't a good name because it implies that `User.translations` has to deal with querysets containing Translations instead of Users.

Maybe `translators`?

pontoon/base/models.py
```diff
@@ -25,12 +25,101 @@
...
+    def get_approved(self, start_date=None, end_date=None):
+        return self.get_changed_translations(start_date, end_date).filter(approved=True)
+
+    def get_unapproved(self, start_date=None, end_date=None):
+        return self.get_changed_translations(start_date, end_date).exclude(approved=True, fuzzy=True)
+
+    def get_needs_work(self, start_date=None, end_date=None):
+        return self.get_changed_translations(start_date, end_date).filter(fuzzy=True)
+
+    def get_contributors_with_stats(self, start_date=None, end_date=None, limit=100):
```
Osmose: `with_translation_counts` is a more descriptive name; `get_contributors` is redundant given that this is called with `User.translations`, and we should be specific about what stats we're adding in case we want to get other stats at a later date.

pontoon/base/models.py
```diff
@@ -25,12 +25,101 @@
 log = logging.getLogger('pontoon')
 
 
+class UserTranslationsManager(models.Manager):
+    """
+    Provides various method to interact with larger sets of translations and their stats for user.
+    """
+
+    def _changed_translations_counts(self, start_date=None, end_date=None, filter=None, exclude=None):
```
Osmose: `counts` should be singular, we're returning a single number.

pontoon/base/models.py
```diff
@@ -25,12 +25,101 @@
...
+        Returns contributors list, sorted by count of their translations.
+        Every user instance has added following properties:
+        * translations_count
+        * translations_approved_count
+        * translations_unapproved_count
+        * translations_needs_work_count
+        Method has been created mainly to improve performance and to optimize
+        count of sql queries during generation of metrics.
+        """
+        def translations_counts(filter=None, exclude=None):
```
Osmose: I like this helper!

pontoon/base/models.py
```diff
@@ -25,12 +25,101 @@
...
+            """Short helper to avoid duplication of passing dates."""
+            return self._changed_translations_counts(start_date, end_date, filter=filter, exclude=exclude)
+
+        contributors = (
```
Osmose: Might as well just return this and tack the order and slice on the end, the variable isn't used for anything.

pontoon/base/views.py
```diff
@@ -353,31 +352,25 @@ def contributor(request, email, template='user.html'):
     return render(request, template, data)
 
 
-def contributors(request, template='users.html'):
+def contributors(request, period='alltime', template='users.html'):
```
Osmose: No need for `period` anymore, nothing is being passed in.

(Someday, but not today, that `template` kwarg is going to die soon.)

pontoon/base/views.py
```diff
@@ -353,31 +352,25 @@ def contributor(request, email, template='user.html'):
...
-        .annotate(translation_count=Count('translation'))
-        .exclude(translation_count=0)
-        .order_by('-translation_count')[:100]
-    )
-
-    for user in users:
-        user.translations = translations.filter(user=user)
+    now = timezone.now()
+    start_date = None
+    end_date = utils.aware_datetime(now.year, now.month, 1)
```
Osmose: Why pin this to the first of the month? I think the general expectation is that looking at a "last month" view would show you from today to a month ago rather than from the beginning of this month to the beginning of last month.

pontoon/base/views.py
```diff
@@ -353,31 +352,25 @@ def contributor(request, email, template='user.html'):
...
+
+    try:
+        period = int(request.GET['period'])
+    except (KeyError, ValueError):
+        period = None
+
+    if period:
```
Osmose: You can put this if statement in the try block and then just have the except block be a `pass` to make it slightly simpler to read.

Osmose: Or even better, set `start_date` to `None` in the except block and you won't have to declare it above.

Osmose:
My comments are all tweaks, but this looks to be pretty much what we want. Nice work on the changes!

pontoon/base/models.py
```diff
@@ -25,12 +25,69 @@
...
+        * translations_needs_work_count
+        Method has been created mainly to improve performance and to optimize
+        count of sql queries during generation of metrics.
+        All counts will be returned from start_date to now().
+        :param date start_date: start date for translations.
+        """
+        def translations_count(query=None):
+            """Short helper to avoid duplication of passing dates."""
+            query = query or Q()
+            query &= Q(translation__date__gte=start_date)
```
Osmose: Moving the start date here is clever, I like it!

pontoon/base/models.py
```diff
@@ -25,12 +25,69 @@
...
+        Helper method, returns expression object which allows us to annotate querysets
+        with counts of translations.
+        """
+        translation_query = (
+            ~Q(translation__string=F('translation__entity__string'))
+            & ~Q(translation__string=F('translation__entity__string'))
+        )
+
+        if args:
+            translation_query &= Q(*args)
```
Osmose: I couldn't find where passing `Q` objects into another `Q` object is documented, and it's kinda weird anyway. Might as well just loop over `args`, which should always be a list anyway, even if it's empty.

pontoon/base/models.py
Osmose: `start_date` is an optional argument, you should check that it's not `None` before using it here.

pontoon/base/views.py
```diff
@@ -356,30 +355,18 @@ def contributor(request, email, template='user.html'):
...
-
-    for user in users:
-        user.translations = translations.filter(user=user)
+    try:
+        period = int(request.GET['period'])
+        start_date = (timezone.now() + relativedelta(months=-period))
+    except (KeyError, ValueError):
+        period = None
+        # When no given data, then we asume receive all data from 1970
+        start_date = timezone.make_aware(datetime.datetime.fromtimestamp(0))
```
Osmose: Since `start_date` is optional, might as well just set it to `None` here as it'll have the same effect.

Osmose:
Looks great to me! Only major thing missing now is tests. :D

jotes:
@Osmose I've updated code again, could you review? Thanks for your patience :)

pontoon/base/models.py
```diff
@@ -25,12 +25,72 @@
...
+        with counts of translations.
+        """
+        translation_query = (
+            ~Q(translation__string=F('translation__entity__string'))
+            & ~Q(translation__string=F('translation__entity__string_plural'))
+            & Q(translation__user__isnull=False)
+        )
+
+        if args:
+            translation_query &= args
```
Osmose: I don't think this is valid (does this work when you test locally? I'd be really surprised if it did).

Sorry if I was misleading before. What I was trying to convey was that if the arguments to this function are `self, *args` (which I think they should be), then you can change this block to be:

```python
for q_filter in args:
    translation_query &= q_filter
```

Because `*args` as a parameter means the value of `args` within the function is always a list, you can loop over it. And it makes calling this function cleaner and more convenient because you don't have to wrap the arguments in a list manually.

pontoon/base/views.py
```diff
@@ -314,11 +317,7 @@ def contributor(request, email, template='user.html'):
         raise Http404
 
     # Exclude unchanged translations
-    translations = (
-        Translation.objects.filter(user=user)
-        .exclude(string=F('entity__string'))
-        .exclude(string=F('entity__string_plural'))
-    )
+    translations = User.translations.get_changed_translations().filter(user=user)
```
Osmose: `User.translations` doesn't exist, and it returns a list of users anyway, whereas `translations` here needs to be a list of translations.

I think you should remove your changes to the `contributor` view and focus on the leaderboard for this PR. It's true that we eventually want to remove these excessive DB queries for the contributor view as well, but we should do that in a PR after this one to keep this one simple.

pontoon/base/models.py
```diff
@@ -25,12 +25,72 @@
...
+
+    def _changed_translations_count(self, args):
+        """
+        Helper method, returns expression object which allows us to annotate querysets
+        with counts of translations.
+        """
+        translation_query = (
+            ~Q(translation__string=F('translation__entity__string'))
+            & ~Q(translation__string=F('translation__entity__string_plural'))
+            & Q(translation__user__isnull=False)
```
Osmose: Why did you add this in? It shouldn't be necessary because we're filtering translations that are related to the users we're querying for, so it's implicit that we only operate on translations that have a user assigned.

Did you see a problem with your local testing which led you to add this in?

pontoon/base/tests/__init__.py
```diff
@@ -95,12 +99,18 @@ class Meta:
         model = ChangedEntityLocale
 
 
+# This helps to test translation from different time periods.
+PERIOD_SIZE = 14 # months
```
Osmose: The custom logic for creating Translations at specific dates should not be added to the `TranslationFactory` class, because it will change how translation dates are created for the entire test suite but is only necessary for a few tests. Instead, you should use a method in your test class that creates the translation instances and passes in the dates you want to `TranslationFactory`.

pontoon/base/models.py
jotes: Yes, if there was an issue with translations_counts, query returned always 1 even if there wasn't any record.

pontoon/base/tests/test_views.py
```diff
@@ -140,3 +143,39 @@ def test_no_subpage_no_stats_in_current_locale(self):
         with patch('pontoon.base.views.render', wraps=render) as mock_render:
             self.client.get(url)
             assert_true('part' not in mock_render.call_args[0][2])
+
+
+class ContributorsTests(TestCase):
```
Osmose: The two tests you've added only test that the queryset within the view has the same length as the number of users created by your test, which doesn't really test the functionality we want.

These are the tests I'd expect for a pull request like this:

- View tests
  - Test that when `period` is not specified, that `start_date` (when passed to `with_translation_counts`) is `None` and `period` in the template context is `None`. You'll want to mock out `with_translation_counts` so you can inspect the argument passed to it.
  - Similar test to the one above, but for the case when `period` is not a valid `int`.
  - Test that when period is a valid int, `start_date` is calculated correctly based on the current date. You'll want to use `patch` to mock out the `timezone.now` function in the view so that the test doesn't rely on the current date when it's running. See [here](https://github.com/mozilla/pontoon/blob/master/pontoon/administration/tests/test_sync_projects.py#L41) for an example of a test that does this.
- `UserTranslationsManagerTests`
  - Test that users who have an email in `settings.EXCLUDE` aren't included in the results from `with_translation_counts`.
  - Test that users who have no translations whatsoever aren't included in the results from `with_translation_counts`.
  - Test that the number of users returned by `with_translation_counts` is limited to the `limit` parameter.
  - Test that users returned by `with_translation_counts` are ordered by their translation count.
  - Test that the 4 attributes added by `with_translation_counts` are calculated correctly. For this one you'll want to create 1 or 2 users and some translations with various combinations of `fuzzy` and `approved`, and then assert that the stats that are added by `with_translation_counts` matches the test data you created.
  - Test that the stats added by `with_translation_counts` do not count translations that are identical to their source value. This could possibly be combined with the previous test if you also create some identical translations before checking the stats.

It's a long list but this code does a lot of stuff. :P

pontoon/base/models.py
Osmose: wat

That's really odd. I guess leave it in for now, and we can take a closer look later. If this works as expected and the tests pass, it probably doesn't have any side effects. So either I'm wrong (very very possible) or something else is causing the issue and this just happens to fix it. Either possibility can be resolved later.

jotes: yeah, it was really weird. However, i'll start by writing more tests and
then we'll if they help to pinpoint more possible suspect.

On Fri, Sep 11, 2015 at 5:12 PM, Michael Kelly <notifications@github.com>
wrote:

> In pontoon/base/models.py
> <https://github.com/mozilla/pontoon/pull/161#discussion_r39281036>:
>
> > @@ -25,12 +25,72 @@
> >  log = logging.getLogger('pontoon')
> >
> >
> > +class UserTranslationsManager(models.Manager):
> > +    """
> > +    Provides various method to interact with larger sets of translations and their stats for user.
> > +    """
> > +
> > +    def _changed_translations_count(self, args):
> > +        """
> > +        Helper method, returns expression object which allows us to annotate querysets
> > +        with counts of translations.
> > +        """
> > +        translation_query = (
> > +            ~Q(translation__string=F('translation__entity__string'))
> > +            & ~Q(translation__string=F('translation__entity__string_plural'))
> > +            & Q(translation__user__isnull=False)
>
> wat
>
> That's really odd. I guess leave it in for now, and we can take a closer
> look later. If this works as expected and the tests pass, it probably
> doesn't have any side effects. So either I'm wrong (very very possible) or
> something else is causing the issue and this just happens to fix it. Either
> possibility can be resolved later.
>
> —
> <https://github.com/mozilla/pontoon/pull/161/files#r39281036>.
>



-- 
jarek@reijutsu:~$ fortune
jarek@reijutsu:~$ fortune
You now have Asian Flu.

jotes:
@mathjazz @Osmose hi, could you review again? i've added some tests.

pontoon/base/tests/__init__.py
```diff
@@ -100,11 +104,16 @@ class TranslationFactory(DjangoModelFactory):
     locale = SubFactory(LocaleFactory)
     string = Sequence(lambda n: 'translation {0}'.format(n))
     user = SubFactory(UserFactory)
-
     class Meta:
         model = Translation
 
 
+class IdenticalTranslationFactory(TranslationFactory):
```
Osmose: Oh, neat idea! It's never occurred to me to make useful subfactories for stuff like this.

pontoon/base/tests/__init__.py
```diff
@@ -100,11 +104,16 @@ class TranslationFactory(DjangoModelFactory):
...
+    @factory.sequence
+    def entity(n):
```
Osmose: Does this rely on the sequence counter being on the same number as the translation string sequence counter? What happens if they get out of sync?

Also, does this work when I want to define properties about the entity? Like:

```python
IdenticalTranslationFactory.create(
     entity__resource__project__slug='foobar'
)
```

pontoon/base/tests/test_models.py
```diff
@@ -0,0 +1,167 @@
...
+        included_contributor = TranslationFactory.create(user__email='included@example.com').user
+        excluded_contributor = TranslationFactory.create(user__email='excluded@example.com').user
+
+        top_contributors = User.translators.with_translation_counts()
+        assert_true(included_contributor in top_contributors)
+        assert_true(excluded_contributor not in top_contributors)
+
+    def test_users_without_translations(self):
+        """
+        Checks if user contributors without translations are returned.
```
Osmose: I think you mean they _aren't_ included.

pontoon/base/tests/test_models.py
```diff
@@ -0,0 +1,167 @@
...
+        inactive_contributor = UserFactory.create(email='inactive@example.com')
+
+        top_contributors = User.translators.with_translation_counts()
+        assert_true(active_contributor in top_contributors)
+        assert_true(inactive_contributor not in top_contributors)
+
+    def test_unique_translations(self):
+        """
+        Checks if contributors with identical translations are returned.
+            """
```
Osmose: Remove this extra indent.

pontoon/base/tests/test_models.py
jotes: Yup, you're right.

On Thu, Sep 17, 2015 at 11:42 PM, Michael Kelly <notifications@github.com>
wrote:

> In pontoon/base/tests/test_models.py
> <https://github.com/mozilla/pontoon/pull/161#discussion_r39802822>:
>
> > +    @override_settings(EXCLUDE=('excluded@example.com',))
> > +    def test_excluded_contributors(self):
> > +        """
> > +        Checks if contributors with mails in settings.EXCLUDE are excluded
> > +        from top contributors list.
> > +        """
> > +        included_contributor = TranslationFactory.create(user__email='included@example.com').user
> > +        excluded_contributor = TranslationFactory.create(user__email='excluded@example.com').user
> > +
> > +        top_contributors = User.translators.with_translation_counts()
> > +        assert_true(included_contributor in top_contributors)
> > +        assert_true(excluded_contributor not in top_contributors)
> > +
> > +    def test_users_without_translations(self):
> > +        """
> > +        Checks if user contributors without translations are returned.
>
> I think you mean they *aren't* included.
>
> —
> <https://github.com/mozilla/pontoon/pull/161/files#r39802822>.
>



-- 
jarek@reijutsu:~$ fortune
jarek@reijutsu:~$ fortune
You now have Asian Flu.

pontoon/base/tests/test_models.py
```diff
@@ -0,0 +1,167 @@
...
+        """
+        Checks if users are ordered by count of contributions.
+        """
+        users = UserFactory.create_batch(5)
+
+        for i, user in enumerate(users):
+            TranslationFactory.create_batch(5-i, user=user)
+
+        top_translations_totals = [user.translations_count for user in User.translators.with_translation_counts()]
+        assert_equal(top_translations_totals, [5, 4, 3, 2, 1])
```
Osmose: You'll want to check the actual user objects rather than their counts to ensure that the exact users we expect are ordered.

pontoon/base/tests/test_models.py
```diff
@@ -0,0 +1,167 @@
...
+            """
+
+        unique_translator = TranslationFactory.create().user
+        identical_translator = IdenticalTranslationFactory.create().user
+        top_contributors = User.translators.with_translation_counts()
+
+        assert_true(unique_translator in top_contributors)
+        assert_true(identical_translator not in top_contributors)
+
+
```
Osmose: One space between method declarations in a class.

pontoon/base/tests/test_models.py
```diff
@@ -0,0 +1,167 @@
...
+    def test_contributors_limit(self):
+        """
+        Checks if proper count of user is returned.
+        """
+        TranslationFactory.create_batch(110)
+
+        top_contributors = User.translators.with_translation_counts()
+
+        assert_true(len(top_contributors) == 100)
```
Osmose: Use `assert_equal` here.

pontoon/base/tests/test_models.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def create_contributor_with_translation_counts(self, approved, unapproved, needs_work, **kwargs):
```
Osmose: Hot fire. This is excellent.

pontoon/base/tests/test_models.py
```diff
@@ -0,0 +1,167 @@
...
+
+        first_contributor = self.create_contributor_with_translation_counts(approved=7, unapproved=3, needs_work=2)
+        second_contributor = self.create_contributor_with_translation_counts(approved=5, unapproved=9, needs_work=2)
+        third_contributor = self.create_contributor_with_translation_counts(approved=1, unapproved=2, needs_work=5)
+
+        top_contributors = User.translators.with_translation_counts()
+
+        assert_equal(top_contributors.count(), 3)
+        assert_equal(top_contributors[0], second_contributor)
+        assert_equal(top_contributors[0].translations_count, 16)
```
Osmose: We have a helper that eliminates a lot of the boilerplate for asserting attributes: https://github.com/mozilla/pontoon/blob/master/pontoon/base/tests/__init__.py#L141

pontoon/base/tests/test_views.py
```diff
@@ -140,3 +146,42 @@ def test_no_subpage_no_stats_in_current_locale(self):
...
+    @patch('pontoon.base.models.UserTranslationsManager.with_translation_counts')
+    @patch('pontoon.base.views.render', wraps=render)
+    def test_default_period(self, mock_render, mock_translations_manager):
+        """
+        Checks how view handles default period.
```
Osmose: Docstrings on tests are useful for elaborating what should be happening. This one doesn't really explain what you want to happen.

Also, if the docstring is sufficiently short (I use 72 characters myself), you can keep it all on one line to save space.

pontoon/base/tests/test_views.py
```diff
@@ -140,3 +146,42 @@ def test_no_subpage_no_stats_in_current_locale(self):
...
+        """
+        response = self.client.get('/contributors/')
+        assert_true(mock_render.call_args[0][2]['period'] == None)
+        assert_true(mock_translations_manager.call_args[0][0] == None)
+
+    @patch('pontoon.base.models.UserTranslationsManager.with_translation_counts')
```
Osmose: I find that using the decorator syntax for patching makes it harder to read through the test cases vs using the `with patch()` syntax within the function.

In either case, you're patching these methods for pretty much every test in this class, so you might as well patch them in `setUp` for all the methods. See https://github.com/mozilla/pontoon/blob/master/pontoon/administration/tests/test_sync_projects.py#L40 for some examples.

pontoon/base/tests/test_views.py
```diff
@@ -140,3 +146,42 @@ def test_no_subpage_no_stats_in_current_locale(self):
...
+    @patch('pontoon.base.views.render', wraps=render)
+    def test_invalid_period(self, mock_render, mock_translations_manager):
+        """
+        Checks how view handles invalid period, it result in period being None - displays all data.
+        """
+        # If period parameter is invalid value
+        response = self.client.get('/contributors/?period=invalidperiod')
+        assert_true(mock_render.call_args[0][2]['period'] is None)
+        assert_true(mock_translations_manager.call_args[0][0] == None)
```
Osmose: `assert_equal` here and the other spots you use `assert_true` for equality. Also, you're testing if the value is `None`, which means you should use `is` to test for identity.

pontoon/base/tests/test_views.py
```diff
@@ -140,3 +146,42 @@ def test_no_subpage_no_stats_in_current_locale(self):
...
+
+    @patch('pontoon.base.models.UserTranslationsManager.with_translation_counts')
+    @patch('django.utils.timezone.now', wraps=now, return_value=datetime(2015, 7, 5))
+    @patch('pontoon.base.views.render', wraps=render)
+    def test_given_period(self, mock_render, mock_now, mock_translations_manager):
+        """
+        Checks if view sets and returns data for right period.
+        """
+        response = self.client.get('/contributors/?period=6')
+        assert_true(mock_render.call_args[0][2]['period'] == 6)
```
Osmose: `assert_equal` for equality. Happens in a few other spots too.

Osmose:
This looks great! I like all the tests you've added, all my comments are just refactoring improvements on them. Nice work!

jotes:
@Osmose hi, could you review again? :)

Osmose:
Fix that use of `assert_attributes_equal` and I'm r+ on this, great work! This is a huge improvement.

jotes:
@Osmose I've fixed asserts, could you look again? Thx for your support.

PR #162, opened 2015-09-08
Fix sync issues

Osmose:
This fixes the two issues causing [bug 1202259](https://bugzilla.mozilla.org/show_bug.cgi?id=1202259):

- #149 changed sync to build VCSEntity instances using the data from the source locale. POFiles use `pot` files instead of `po` files for their source resources, so sync was attempting to read from the wrong files as it assumed source resources shared the same relative filename as their translated counterparts.

While I was fixing this I decided to go ahead and move the path functions from `Project` to `VCSProject`, since they're only used there and it makes sense for the VCS stuff to be concerned with file paths and the DB stuff to be concerned with database stuff. I left `checkout_path` on `Project` because it relies on data in the database.

- Previously, the langfile parser assumed comments associated with entities started with one `#` and unassociated ones started with two `#`s. A new langfile showed up that had a standalone comment with one `#`, so I fixed the parser to recognize all comments as starting with an arbitrary number of `#`s. Instead, it associates comments by running through the parsed file and finding comments that immediately precede an entity.

- The langfile parser also failed on files that did not end with a newline. That's no longer a requirement.

@mathjazz r?


mathjazz:
r+

The amount of changes required to fix this scared me a little, but after reading description, I see you also made other changes. It would be easier to digest and have smaller potential to break things if those changes were in s separate PR.

I think we should be using the same .lang parser as bedrock to avoid similar issues in the future. It can be either yours or theirs (which we used before). In any case, you should get in touch with them.

PR #163, opened 2015-09-09
Fix bug 1201212: Preserve fuzzy status for po entities.

Osmose:
Our pofile sync code had an issue where it would remove the fuzzy flag for an entity if it was fuzzy in VCS, because we weren't checking that the entity was marked as not fuzzy before removing the flag.

To help debug this issue I also added a `--no-pull` argument to `sync_projects` that skips updating from VCS during sync. This let me test syncing on a specific SVN commit, which is really useful for debugging.

@mathjazz r?

### DIFF
--- a/pontoon/base/formats/po.py
+++ b/pontoon/base/formats/po.py
@@ -46,7 +46,7 @@ class POEntity(VCSTranslation):
 
         if self.fuzzy and 'fuzzy' not in self.po_entry.flags:
             self.po_entry.flags.append('fuzzy')
-        elif 'fuzzy' in self.po_entry.flags:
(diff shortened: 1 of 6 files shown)


pontoon/base/formats/po.py
```diff
@@ -36,6 +36,8 @@ def __init__(self, po_entry, order):
 
     def update_entry(self, locale):
         """Update the POEntry associated with this translation."""
+        if self.source_string == 'Ask a Question':
+            import ipdb; ipdb.set_trace()
```
mathjazz: Whooops?

mathjazz:
LGTM, r+

PR #164, opened 2015-09-09
Fix bug 1203057: Allow blank translations in silme.

Osmose:
The problem is that blank translations would be exported as missing translations by the silme format plugin (meaning Entity.strings would be `{}` instead of `{None: ''}`). Thus, sync thought that VCS had no translation and unapproved all translations in the database, instead of identifying the empty translation as the correct one.

@mathjazz r?


mathjazz:
r+

PR #165, opened 2015-09-10
Latest activity

mathjazz:
This removes the **URL** column and adds the **Latest activity** column to the dashboards. It tells us when was the latest translation submitted/approved for a specific project or locale and who submitted/approved it (if we have that information). Later we could also include string deletions (if we decide to keep dates for this) and source string updates (we'll might need to store this information as part of [bug 1191718](https://bugzilla.mozilla.org/show_bug.cgi?id=1191718) anyways).

This PR is mostly about html/css/js, but the tiny piece of python I added into the model is what I'm particularly unhappy about. There must be a more elegant way of getting the latest translation/approval.

I wanted to get to this before [bug 1178880](https://bugzilla.mozilla.org/show_bug.cgi?id=1178880), which will also include a similar dashboard. And we'll need to [redesign dashboards](https://bugzilla.mozilla.org/show_bug.cgi?id=1203543), too.

@Osmose r?


pontoon/base/templates/project_selector.html
```diff
@@ -13,33 +13,38 @@
...
     {% endif %}
     <ul>
       {% for p in projects %}
       <li class="clearfix{% if project and project == p %} current{% endif %}">
-        {% if url_prefix %}
-        <a href="/{{ url_prefix }}/{{ p.slug }}/" class="clearfix">
+        <span class="name"
+          data-slug="{{ p.slug }}"
+          {% if p.details %}data-details="{{ p.details }}"{% endif %}>
+          {% if url_prefix %}<a href="/{{ url_prefix }}/{{ p.slug }}/" class="clearfix">{% endif %}
```
Osmose: Someday, `url_prefix`, we will destroy you. SOEMDAAY

pontoon/base/templates/project_selector.html
```diff
@@ -13,33 +13,38 @@
...
+            {{ p.name }}
+          {% if url_prefix %}</a>{% endif %}
+        </span>
+        {% if not project %}
+        <span class="latest">
+          {% set latest_activity = p.latest_activity(locale) if locale else p.latest_activity() %}
+          {% include 'latest_activity.html' %}
```
Osmose: This pattern for reusing template code isn't a great idea because the included template and this template share a template context, leading to possible variable name collisions, as well as making it very difficult to figure out where a variable is being set (is it set in the template? In the view?).

You should replace this with a [macro](http://jinja.pocoo.org/docs/dev/templates/#macros).

pontoon/base/templates/project_selector.html
```diff
@@ -13,33 +13,38 @@
...
+        <span class="name"
+          data-slug="{{ p.slug }}"
+          {% if p.details %}data-details="{{ p.details }}"{% endif %}>
+          {% if url_prefix %}<a href="/{{ url_prefix }}/{{ p.slug }}/" class="clearfix">{% endif %}
+            {{ p.name }}
+          {% if url_prefix %}</a>{% endif %}
+        </span>
+        {% if not project %}
+        <span class="latest">
+          {% set latest_activity = p.latest_activity(locale) if locale else p.latest_activity() %}
```
Osmose: It's a bad idea to put logic like this in the template itself because it makes it harder to track what code is getting executed during a request when debugging an issue, and it's really difficult to test. The solution is usually to put this logic in the view and pass in the result as part of the template context.

pontoon/base/models.py
```diff
@@ -67,6 +67,28 @@ def validate_cldr(value):
                 '%s must be a list of integers between 0 and 5' % value)
 
 
+def get_latest_activity(translations):
```
Osmose: The logic here is mostly fine, but I'm not a huge fan of functions living in `models.py` when there's more interesting places to hook this stuff in.

Because the only parameter is a translation queryset, I think making a [custom Translation queryset](https://docs.djangoproject.com/en/1.8/topics/db/managers/#calling-custom-queryset-methods-from-the-manager) is a good idea. Then we can take _any_ translation queryset and just call `translations.latest_activity()` and get back the object we want. Thoughts?

pontoon/base/models.py
```diff
@@ -67,6 +67,28 @@ def validate_cldr(value):
...
+    """Get latest activity data from provided translations."""
+    if not translations.exists():
+        return
```
Osmose: Better to be explicit about the fact that we're returning `None` here.

pontoon/base/models.py
```diff
@@ -67,6 +67,28 @@ def validate_cldr(value):
...
+
+    latest_translation = translations.order_by('-date')[0]
+    latest_approval = translations.order_by('-approved_date')[0]
+
+    if latest_approval.approved_date and latest_translation.date < latest_approval.approved_date:
+        latest = latest_approval
```
Osmose: Might as well just return the object here, makes it easier to see the difference between the approval vs the translation, even though it's slightly more lines of code.

pontoon/base/models.py
```diff
@@ -116,6 +138,14 @@ def stringify(self):
     class Meta:
         ordering = ['name', 'code']
 
+    def latest_activity(self, project=None):
```
Osmose: If you move the template logic mentioned below into the view, these model methods become less useful. I think it'd be better to go without them and move this logic into the view as well.

pontoon/base/templates/project_selector.html
Osmose: Now that I notice it's within a for loop it makes things harder... hmmm...

Osmose: Yeah, if the locale is an optional argument anyway, could you just always pass it in? That'd remove the if statement from the template and avoid the issue as well.

pontoon/base/static/js/main.js
```diff
@@ -485,28 +481,41 @@ $(function() {
...
+      var date = $(el).find('time').attr('datetime') || 0;
+      return new Date(date).getTime().toString();
+    }
+
+    function getString(el) {
+      return $(el).find('span:eq(' + index + ')').html();
+    }
+
+    var node = $(this),
+        index = $(this).index(),
```
Osmose: You can replace `$(this)` with `node` and save yourself a ton of re-jQuerying things throughout this function.

pontoon/base/static/js/main.js
```diff
@@ -485,28 +481,41 @@ $(function() {
       };
     }
 
-    var index = $(this).index(),
-        current = $(this).attr('class'),
+    function getDate(el) {
+      var date = $(el).find('time').attr('datetime') || 0;
+      return new Date(date).getTime().toString();
```
Osmose: Why convert this to a string and then compare? Why not just compare dates directly?

Osmose:
This looks pretty good! I deleted some ill-thought-out comments I had before fully understanding this PR. You're pretty close, just the queryset and macro stuff to handle and this will be good!

mathjazz:
@Osmose, r?

Osmose:
A thing of beauty. 

Some tests would really round this out nicely, though. But still, this is great!

mathjazz:
Tests added.

@Osmose I need some help. I noticed TranslationFactory doesn't set the date attribute to the specified value (but it does set approved_date). Is that because auto_now_add is set to True?

I should probably also squash all these commits.

pontoon/base/models.py
```diff
@@ -116,6 +116,14 @@ def stringify(self):
...
+        """Get latest activity for locale and project if provided."""
+        translations = Translation.objects.filter(locale=self)
+        if project:
```
Osmose: You could add tests for this if statement but it's iffy whether it's necessary or not. Up to you!

pontoon/base/tests/test_models.py
```diff
@@ -0,0 +1,75 @@
...
+    ProjectFactory,
+    ResourceFactory,
+    TranslationFactory,
+    UserFactory,
+    TestCase
+)
+from pontoon.base.utils import aware_datetime
+
+
+class LocaleTests(TestCase):
```
Osmose: This class should be called `TranslationQuerySetTests` since that's the class being tested.

pontoon/base/tests/test_models.py
```diff
@@ -0,0 +1,75 @@
...
+    def test_latest_activity_translated(self):
+        """
+        If latest activity in Translation QuerySet is translation submission,
+        return submission date and user.
+        """
+        user0 = UserFactory.create()
```
Osmose: These users are shared and could go in a `setUp` method. You could also use `UserFactory.create_batch(2)` to create them both at the same time.

Osmose:
> @Osmose I need some help. I noticed TranslationFactory doesn't set the date attribute to the specified value (but it does set approved_date). Is that because auto_now_add is set to True?

Yep, that's probably it. I found https://github.com/rbarrois/factory_boy/issues/102 explaining the issue and a possible solution. I like making `timezone.now` the default instead of relying on `auto_now_add`.

pontoon/base/tests/test_models.py
```diff
@@ -0,0 +1,75 @@
...
+        user1 = UserFactory.create()
+
+        translation0 = TranslationFactory.create(
```
Osmose: Descriptive variable names make tests a lot easier to understand (I used to name test variables like this but I ended up getting confused a lot). In this case, `latest_submission` and `latest_approval` would be good names for these translations.

Osmose: Also, you call `TranslationFactory.create` often enough for these tests that it might be worth it to create a helper to save on lines. Something with descriptive kwargs like this might work:

```python
latest_submission = self._translation(user0, submitted=(1970, 1, 1), approved=(1970, 1, 1))
```

pontoon/base/tests/test_models.py
```diff
@@ -0,0 +1,75 @@
...
+            approved_user=user0
+        )
+        translation1 = TranslationFactory.create(
+            date=aware_datetime(1970, 1, 2),
+            user=user1,
+            approved_date=aware_datetime(1970, 1, 2),
+            approved_user=user1
+        )
+
+        translations = Translation.objects.filter(id__in=[translation0.id, translation1.id])
```
Osmose: No need to filter here. The tables are cleared between tests, so the only translations that exist are the ones you create. `Translations.objects.all().latest_activity()` is equivalent!

Osmose:
r+wc, great work! :D

pontoon/base/tests/test_models.py
mathjazz: Oh! That's good to know!

mathjazz:
Squashed commits in a separate branch and created #171.

PR #166, opened 2015-09-11
Fix bug 1201707: Add support for multiple repos.

Osmose:
Firefox and friends have multiple repos where strings are stored. The
old sync process used some magic code during sync to detect this and
download and commit strings to multiple repositories. This commit adds
support directly to Pontoon for multiple repos so that the new sync
process can support these special repos.

The old sync process still works by assuming that the first-found repo
is the singular repo that it expects.

While this adds UI to the admin interface, it's unpolished due to our
future plans of replacing the admin interface with the Django admin. In
addition, some code for supporting Transifex and File repos has been
removed due to those features also being slated for removal. Rather
than properly support these unused and about-to-die features, we do the
lazy thing and ignore them.

@mathjazz I'll be updating this with tests, but it's about ready for some preliminary feedback and testing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ tmp/*
 .env
 docs/_build
 venv
+/media/
--- a/pontoon/administration/files.py
+++ b/pontoon/administration/files.py
@@ -715,10 +715,6 @@ def update_from_repository(project):
                     os.path.join(repository_url_master, l.code),
                     os.path.join(repository_path_master, l.code))
 
-    # Store project repository_path
-    project.repository_path = repository_path
-    project.save()
-
 
 def dump_po(project, locale, relative_path):
     """Dump .po (gettext) file with relative path from database."""
--- a/pontoon/administration/forms.py
+++ b/pontoon/administration/forms.py
@@ -1,0 +1,26 @@
+from django.forms import ModelForm
+from django.forms.models import inlineformset_factory
+
+from pontoon.base.models import Project, Repository, Subpage
+
+
+class ProjectForm(ModelForm):
+    class Meta:
+        model = Project
+        fields = ('name', 'slug', 'locales', 'transifex_project',
+                  'transifex_resource', 'info_brief', 'url', 'width',
+                  'links', 'disabled')
+
+
+SubpageInlineFormSet = inlineformset_factory(
+    Project, Subpage,
+    extra=1,
+    fields=('project', 'name', 'url')
+)
+
+
+RepositoryInlineFormSet = inlineformset_factory(
+    Project, Repository,
+    extra=1,
+    fields=('type', 'url', 'multi_locale', 'source_repo'),
+)
--- a/pontoon/administration/management/commands/sync_projects.py
+++ b/pontoon/administration/management/commands/sync_projects.py
@@ -8,8 +8,7 @@ from django.utils import timezone
 
 from bulk_update.helper import bulk_update
 
-from pontoon.administration.files import update_from_repository
-from pontoon.administration.vcs import commit_to_vcs, CommitToRepositoryException
+from pontoon.administration.vcs import CommitToRepositoryException
 from pontoon.base.models import (
     ChangedEntityLocale,
     Entity,
@@ -86,7 +85,7 @@ class Command(BaseCommand):
     def handle_project(self, db_project):
         # Pull changes from VCS and update what we know about the files.
         if not self.no_pull:
-            update_from_repository(db_project)
+            self.pull_changes(db_project)
         vcs_project = VCSProject(db_project)
         self.update_resources(db_project, vcs_project)
 
@@ -189,6 +188,11 @@ class Command(BaseCommand):
         key = entity.key or entity.string
         return ':'.join([entity.resource.path, key])
 
+    def pull_changes(self, db_project):
+        """Update the local files with changes from the VCS."""
+        for repo in db_project.repositories.all():
+            repo.pull()
+
     def commit_changes(self, db_project, vcs_project, changeset):
         """Commit the changes we've made back to the VCS."""
         for locale in db_project.locales.all():
@@ -207,15 +211,11 @@ class Command(BaseCommand):
                 'authors': authors
             })
 
+            locale_path = vcs_project.locale_directory_path(locale.code)
             try:
-                result = commit_to_vcs(
-                    db_project.repository_type,
-                    vcs_project.locale_directory_path(locale.code),
-                    commit_message,
-                    commit_author,
-                    db_project.repository_url
-                )
-            except CommitToRepositoryException as err:
+                repo = db_project.repository_for_path(locale_path)
+                result = repo.commit(commit_message, commit_author, locale_path)
+            except (CommitToRepositoryException, ValueError) as err:
                 result = {'message': unicode(err)}
 
             if result is not None:
--- a/pontoon/administration/static/css/admin_project.css
+++ b/pontoon/administration/static/css/admin_project.css
@@ -131,7 +131,8 @@ form a:link, form a:visited {
   text-transform: uppercase;
 }
 
-form a.add-subpage {
+form a.add-subpage,
+form a.add-repo {
   font-size: 14px;
   letter-spacing: 0;
   margin-top: 17px;
@@ -185,6 +186,16 @@ form a.add-subpage {
 .checkbox label {
   float: left;
   padding: 6px 0 0 13px;
+}
+
+.repository .locale-wrapper {
+  padding-top: 20px;
+}
+
+.repository .locale-wrapper label,
+.repository .source-wrapper label,
+.repository .delete-wrapper label {
+  display: inline;
 }
 
 .repository.authenticate .bottom label {
@@ -328,7 +339,8 @@ form a.add-subpage {
   text-transform: uppercase;
 }
 
-.subpages[data-count] {
+.subpages[data-count],
+.repository-empty {
   display: none;
 }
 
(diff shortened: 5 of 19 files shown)


Osmose:
@mathjazz Tests are added. This is ready for a full review!

I skipped adding tests for a few things, mainly because either they were going to be removed eventually (like `can_commit`, which will be unnecessary once Transifex and File repo types are gone) or because they were so trivial as to not need them (like the `checkout_path` properties).

mathjazz:
Seems to be working with projects that don't use multiple repositories. :-) For others, I get the following error when running sync_projects locally:
```Exception: No source directory found for project firefox-os```

Happenned with Firefox OS, Firefox and Firefox for Android. I tried with existing projects and also added the new ones. Same error.

Also, ```Please wait while strings are imported from the repository.``` message is still displayed after new project is added or locales are added/removed, even though updating projects no longer works through the admin. This is a minor issue, of course, that can be resolved as part of Admin redesign.

--

I'd like to take a step back here and think about the approach we're taking. Why do we have to change the way multiple repositories are handled in first place? If the answer is to make the code easier to maintain/look at, then I think it's insufficient. For the following reasons:

1. We are simplifying the code on the account of user experience. We're making the UX for adding new locales even more complicated and much more error prone, because one can easily forget to add the locale repository. Or add the repository for locale ab-CD instead of ab_CD. Currently, all of this is automated.

2. We are changing something that works fine and that change has the potential to break any project. I'd be surprised (and happy!) if we don't break anything, even with code review, unit tests and local testing. We did something similar with .lang files parser/serializer and problems keep hitting us for weeks.

@Osmose, thoughts?

Osmose:
Huh, I tested locally with a multi-locale repo and it worked, I'll test those specifically and see what I got wrong.

Osmose:
> I'd like to take a step back here and think about the approach we're taking. Why do we have to change the way multiple repositories are handled in first place? If the answer is to make the code easier to maintain/look at, then I think it's insufficient. For the following reasons:
>
> 1. We are simplifying the code on the account of user experience. We're making the UX for adding new locales even more complicated and much more error prone, because one can easily forget to add the locale repository. Or add the repository for locale ab-CD instead of ab_CD. Currently, all of this is automated.

I disagree. Currently, if you have a bunch of repos (one per locale), you have to _know_ that Pontoon expects you to have all your locale directories on one level, and have your en-US/templates/whatever on that same level too. Then, you give Pontoon the URL to your en-US locale and it just pulls that off the end of the URL and replaces the locales and attempts to check each of them out.

The problem with all of this is that it requires the user to be very aware of exactly how Pontoon is automatically figuring things out. It limits you to only a certain set of possibilities (one repo per locale) that we've already encountered, and any deviations will require us to update the code. It took me over half an hour just to figure out enough to write the paragraph above, _and I'm a developer on this code_.

Also, we can't support multiple repos on Github because our automagic finds other locale repos by doing path operations on the repo URL, whereas Github gives you https://github.com/Osmose/mozorg-lang.git URLs.

With this PR, the process is more manual, which is a downside. But the cognitive load is greatly reduced. You tell Pontoon where all the strings are, and you tell it which locales it should care about. There's also possibilities around making client-side widgets for entering the URLs quickly in the future.

There's also future benefits, like storing the last checked-out revision on the Repository model and using that to check if we even need to bother updating resources within that repo.

Basically, I think cognitive load is a larger problem than making the user type and click a bunch, and I don't expect us to make the user click and type a bunch in the long run anyway. 

> 2. We are changing something that works fine and that change has the potential to break any project. I'd be surprised (and happy!) if we don't break anything, even with code review, unit tests and local testing. We did something similar with .lang files parser/serializer and problems keep hitting us for weeks.

The new behavior only affects projects using sync_projects; are there multi-repo projects using that? (I didn't think there were given this PR was intended to add that ability in the first place, or maybe I'm mixing up multi-repo projects with projects that need to use your en-US moz-central repo?)

In any case, as long as we test to make sure switching is possible, we can slowly test multi-repo projects one at a time while switching them over to sync_projects. If they break, we move them back to update_projects and fix the error, until we're satisfied with their behavior.

That was my intention with keeping both sync processes in place, but we didn't test it well enough and I didn't push for us to start doing that. We should always be able to fall back to a working system.

mathjazz:
I agree the new code is more powerful, but we might never need the extra features it brings (e.g. multi-locale GitHub repositories). We should focus on numerous features we need today instead of things we might never need. I made that mistake with Transifex support, so it's great that this PR is eliminating it. ;)

One of the features we need today is to simplify adding new locales. However, this PR is doing the exact opposite.

It's true that the way Pontoon handles multiple repositories is not documented, but that's only because I'm the only admin and didn't feel the need for that. A simple note in the admin interface would resolve this problem. Once you read it, it makes sense.

Also, what are the odds we'll need to support a multi-repo project that will keep locale repositories in URLs that couldn't be abstracted with a template? Currently we have 8 multi-repo projects (4 Firefox OS and 4 Auroras).

Osmose:
I'm still convinced this is a better data model for repos, but I do understand your objection about making adding new locales hard. Would a good compromise be adding templated repo URLs like you suggested? Something like allowing repo URLs like `https://hg.mozilla.org/gaia-l10n/{locale_code}/` and adding a checkbox/boolean for repos that marks them as "templated". Then, when we pull repos that are templated, we loop through the locales for the related project and pull.

Since `Repo.commit` currently takes a path to commit, we can detect which locale repo needs to be committed from that.

Thoughts?

mathjazz:
That sounds like a win-win - go for it!

BTW, I noticed projects I tested this PR on got checked out to /media/projects/slug, maybe that's why I got that "No source directory found" Exception?

Osmose:
> BTW, I noticed projects I tested this PR on got checked out to /media/projects/slug, maybe that's why I got that "No source directory found" Exception?

Sort've, yeah. The problem is that we want to group repo checkouts together so the auto-detection for locale and source repos works when we start them from the root of the multiple checkouts, but repos currently name their checkouts after their PK because they have no other useful info. Which means that locale repos that have the locale in the URL but not in a folder within the checkout won't be found, including the source repo.

I'm working on a fix that will incorporate the repo URL into the directory names to make the auto-detection work again.

Osmose:
I've updated the branch with support for "Multi-locale repos". When you create a repo, if you mark it as multi-locale, you can add a format variable `https://example.com/{locale_code}/` to the URL that will get replaced with the locale codes for each active locale on the project.

In order to identify the source locale without having to hardcode special URLs, I also added a "source repo" checkbox. When it is checked, the repo in question will append "en-US" to the checkout path so that it's detected as a source repo. Long-term I'd like to have a better solution than this, but it works for now and avoids hardcoding the URLs.

I tested this on some mozorg-lang locale-specific repos and was able to sync and commit without any issues: https://github.com/Osmose/mozorg-lang-ca/commit/d2207eaa87b8dbcf0a645ce5f9cbc92ceda0deac and https://github.com/Osmose/mozorg-lang-af/commit/55a41aa1cb8c730a8bd7932f2a97bf3b67f64843.

@mathjazz I think this is ready for an r?

mathjazz:
I confirm this works fine locally for a bunch of single-repo projects and two multi-repo projects, set up as follows:
* Firefox OS
 * ssh://hg.mozilla.org/gaia-l10n/en-US/
 * ssh://hg.mozilla.org/gaia-l10n/{locale_code}/ (multi locale)
* Firefox Aurora
 * ssh://hg.mozilla.org/users/m_owca.info/firefox-aurora/ (source repo)
 * ssh://hg.mozilla.org/releases/l10n/mozilla-aurora/{locale_code}/ (multi locale)

Nice work!

It would be nice if ```Multi locale``` option would be set automatically if ```{locale_code}``` is detected inside repository URL. In which case ```multi_locale``` no longer needs to be stored in the DB and can become a property.

For Firefox OS, the second URL should be enough, the ```en-US``` one seems redundant, but it's required.

I made Firefox Aurora work by faking .inc support, but I also noticed a bug in ```vcs_models.py```:
https://github.com/Osmose/pontoon/blob/multiple-project-repos/pontoon/base/vcs_models.py#L128

The line should be replaced with:
```filenames = [f for f in filenames if not f.endswith('region.properties')]```

Another can-be-fixed-later: After clicking on Add more (Repositories), only Delete option is available for the newly added entry. You need to save before Source repo and Multi locale options show up.

Osmose:
> Might be a good idea to "squash" migrations.

Django docs say not to bother with this unless you have thousands of migrations as it gets complicated with pure-python migrations and doesn't really help speed or anything, and I tend to agree. But I did squash the commits!

Osmose:
> It would be nice if Multi locale option would be set automatically if {locale_code} is detected inside repository URL. In which case multi_locale no longer needs to be stored in the DB and can become a property.

Filed: https://bugzilla.mozilla.org/show_bug.cgi?id=1207826

> For Firefox OS, the second URL should be enough, the en-US one seems redundant, but it's required.

I agree that this could be improved but I'm not yet sure of how. Perhaps we should file and discuss?

> I made Firefox Aurora work by faking .inc support, but I also noticed a bug in vcs_models.py:
> https://github.com/Osmose/pontoon/blob/multiple-project-repos/pontoon/base/vcs_models.py#L128
> 
> The line should be replaced with:
> filenames = [f for f in filenames if not f.endswith('region.properties')]

Filed: https://bugzilla.mozilla.org/show_bug.cgi?id=1207829

> Another can-be-fixed-later: After clicking on Add more (Repositories), only Delete option is available for the newly added entry. You need to save before Source repo and Multi locale options show up.

Weird, I didn't notice that! On one hand it's fixable, on the other hand it will be fixed by [bug 1203788](https://bugzilla.mozilla.org/show_bug.cgi?id=1203788) so I'm not yet filing a bug for it.

PR #167, opened 2015-09-14
Fix bug 1204234: process empty strings on INI file

mathjazz:
By adding silme version number to egg, which installs our patched
version of silme library.

@Osmose r?


Osmose:
r+

PR #172, opened 2015-09-20
Bug 1206176. From now on en-US is default locale for pontoon-intro

jotes:
@Osmose @mathjazz i wasn't sure how do you work  with database on existing production instances, so i decided to create separate migration instead of modyfing initial_data -> it will update all dev environments. It should resolve issue with 'tl' locale.


mathjazz:
This looks like an empty migration?

En-US is not a good default, we only ever use it as a source locale. We should even remove it from the list of supported locales -> which would break code -> which needs to be modified accordingly.

jotes:
@mathjazz sorry, it looks like i've messed something with my branches, i'll push fix in the next hour.

jotes:
@mathjazz could you  review?

mathjazz:
Looks good.  Tested this locally and doesn't break anything. r+

PR #173, opened 2015-09-21
Bug 1200191. Added field for team description

jotes:
hi @mathjazz could you review if that's enough for now? thx.

### DIFF
--- a/pontoon/base/migrations/0024_locale_team_description.py
+++ b/pontoon/base/migrations/0024_locale_team_description.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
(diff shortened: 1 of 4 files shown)


mathjazz:
Nice work, @jotes!

In template, you should use the same structure as we use in the project page: "form > .about > p":
https://github.com/mozilla/pontoon/blob/master/pontoon/base/templates/project.html#L23

Also, you should use the same CSS, by moving the following rules to style.css (and probably using more strict selector):
https://github.com/mozilla/pontoon/blob/master/pontoon/base/static/css/project.css#L1-L15

jotes:
@mathjazz i've updated pr, could you look again?

mathjazz:
Good job! r+

PR #177, opened 2015-09-22
Add support for Mozilla Beta repositories

mathjazz:
This extends the script to also support Mozilla Beta repositories, which we can use for adding Firefox Beta etc. to Pontoon.

Appears to be working fine:
http://hg.mozilla.org/users/m_owca.info

@Osmose r?

### DIFF
--- a/bin/mozilla-en-US.py
+++ b/bin/mozilla-en-US.py
@@ -9,35 +9,35 @@ import shutil
 import subprocess
 
 TARGET_REPOS = {
-    'firefox-aurora': {
+    'firefox': {
         'folders': [
             'browser', 'browser/branding/official', 'dom', 'netwerk',
             'security/manager', 'services/sync', 'toolkit', 'webapprt',
         ],
-        'repository': 'mozilla-aurora',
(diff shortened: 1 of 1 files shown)


mathjazz:
This is **LOW** priority.

bin/mozilla-en-US.py
```diff
@@ -96,33 +96,35 @@ def push(path):
...
-# Clone or update source repositories
-for repo in SOURCE_REPOS:
-    base = 'ssh://hg.mozilla.org/releases/'
-    url = os.path.join(base, repo)
-    target = os.path.join('source', repo)
-    pull(url, target)
+for channel in ['aurora', 'beta']:
+    for repo in SOURCE_REPOS:
+        ending = repo + '-' + channel
+        url = os.path.join('ssh://hg.mozilla.org/releases/', ending)
```
Osmose: Really should be using `'/'.join` for this, URLs ain't paths.

Osmose:
r+, lgtm, and if it works, yay!

PR #178, opened 2015-09-23
Fix bug 1202000: Add tests for Entity.for_project_locale()

mathjazz:
@Osmose r? Another low priority thing.

### DIFF
--- a/pontoon/base/tests/test_models.py
+++ b/pontoon/base/tests/test_models.py
@@ -1,7 +1,12 @@
 from django_nose.tools import assert_equal
 
-from pontoon.base.models import Translation
+from pontoon.base.models import Entity, Translation
 from pontoon.base.tests import (
+    EntityFactory,
+    LocaleFactory,
+    ProjectFactory,
+    ResourceFactory,
+    SubpageFactory,
     TranslationFactory,
     UserFactory,
     TestCase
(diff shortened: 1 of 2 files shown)


pontoon/base/tests/test_models.py
```diff
@@ -48,3 +53,85 @@ def test_latest_activity_approved(self):
...
+            locale=self.locale,
+            string='Other Translated String'
+        )
+        self.subpage = SubpageFactory.create(
+            project=self.project,
+            name='Subpage',
+            resources=[self.main_resource]
+        )
+
+    def assert_entities_equal(self, entity, path, original, translation):
```
Osmose: You're not really comparing two entities... maybe `assert_serialized_entity`? I dunno, it's minor.

Osmose:
Looks pretty good, and could be merged as-is, but there's a few more interesting test cases missing (in rough order of how much they matter):

- Serializing a plural translation.
- Ensuring that the entities are sorted by their order (create entities in a different order than they're assigned and assert they're sorted correctly).
- Making sure the filters at the beginning work right (normally these aren't worth testing if all the method does is filter the queryset, but in this case the method is complex enough to warrant ensuring that the locale, project, and obsolete filtering all work as expected).
- Ensuring at least once that _all_ of the attributes in the serialized entity are assigned correctly.

Osmose:
r+ on what's here thusfar, though. Nice work!

mathjazz:
So the bug was filed after I made the changes to the logic that is being tested in my initial tests. But it's true that the name of the bug implies they should be testing the entire Entity.for_project_locale() method. Added more tests as you suggested, which should cover the method in full.

@Osmose r?

Osmose:
r+, great work! Although I wonder why the Travis isn't running the tests?

mathjazz:
Merged in 3b85e6ac12df2b023ac8404dcb06fa3c25ecbc23.

PR #179, opened 2015-09-23
Add stats to subpages in the parts menu

mathjazz:
I'll add tests.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -284,6 +284,56 @@ class Project(models.Model):
 
         return translations.latest_activity()
 
+    @property
+    def locales_parts_stats(self):
+        """Get project locales with their pages/paths and stats."""
+        pages = self.subpage_set.all()
+        resources = self.resource_set.filter(entity__obsolete=False).distinct()
+        details = {}
+
+        for locale in self.locales.all():
+            stats = Stats.objects.filter(resource__in=resources, locale=locale)
+            locale_details = []
+
+            # Get subpage name and stats if subpages defined
+            if pages.exists():
+
+                # List only subpages, whose resources are available for locale
+                pages = pages.filter(resources__stats=stats) or pages
+
+                for page in pages:
+
+                    # Stats for subpage resources only if defined
+                    if page.resources.exists():
+                        page_stats = stats.filter(resource__in=page.resources.all())
+
+                    # Or for the entire project
+                    else:
+                        page_stats = stats
+
+                    dictionary = {
+                        'resource__path': page.name
+                    }
+                    dictionary.update(page_stats.aggregate(
+                        resource__entity_count=Sum('resource__entity_count'),
+                        translated_count=Sum('translated_count'),
+                        approved_count=Sum('approved_count')
+                    ))
+                    locale_details.append(dictionary)
+
+            # Get resource paths and stats if more than one resource present
+            elif resources.count() > 1:
+                locale_details = stats.order_by('resource__path').values(
+                    'resource__path',
+                    'resource__entity_count',
+                    'translated_count',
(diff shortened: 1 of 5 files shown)


Osmose:
> I'll add tests.

YOU'RE DAMN RIGHT YOU WILL

mathjazz:
I've moved this to model, where it belongs.

@Osmose r?

pontoon/base/models.py
```diff
@@ -284,6 +284,52 @@ def latest_activity(self, locale=None):
 
         return translations.latest_activity()
 
+    @property
+    def locales_parts_stats(self):
+        """Get project locales with their pages/paths and stats."""
+        pages = Subpage.objects.filter(project=self)
```
Osmose: `self.subpage_set.all()` works too, and allows us to prefetch them if we want.

pontoon/base/tests/test_models.py
```diff
@@ -80,6 +84,102 @@ def test_repository_for_path(self):
...
+            path='/main/path.po'
+        )
+        EntityFactory.create(resource=resource)
+        StatsFactory.create(resource=resource, locale=locale)
+
+        return locale, locale_other, project, resource
+
+    def test_locales_parts_stats_no_page_one_resource(self):
+        """
+        Return empty list.
```
Osmose: It's useful to also include what conditions should cause `locales_parts_stats` to return an empty list.

pontoon/base/tests/test_models.py
```diff
@@ -80,6 +84,102 @@ def test_repository_for_path(self):
         path = os.path.join(repo2.checkout_path, 'foo', 'bar')
         assert_equal(project.repository_for_path(path), repo2)
 
+    def locales_parts_stats_helper(self):
+        locale = LocaleFactory.create()
```
Osmose:
```python
locale, locale_other = LocaleFactory.create_batch(2)
```

pontoon/base/tests/test_models.py
```diff
@@ -80,6 +84,102 @@ def test_repository_for_path(self):
         path = os.path.join(repo2.checkout_path, 'foo', 'bar')
         assert_equal(project.repository_for_path(path), repo2)
 
+    def locales_parts_stats_helper(self):
```
Osmose: This is a nice reduction in repeated code; perhaps putting it in a `setUp` method would be even better? Rather than making tests call it.

If you decide to do that I'd suggest making a separate test class for it so the `setUp` doesn't interfere with tests that don't need it. Although I never really thought about it until now, I don't see why having multiple test classes for a single class to be tested is a bad thing if you use it for stuff like that.

Osmose:
I'm having trouble wrapping my head around how this function works; it's really confusing why the method returns sort've different things based on whether subpages exist or not, or why it only checks the first subpage for resources, and stuff like that. I suspect that fixing these issues would involve a lot more work and probably fall under whatever redesign we eventually want to consider.

Having said that, I don't see any major sins here, and if it's got passing tests that check the things you think it should be doing, that's good enough for me, at least for now. r+wc if you're comfortable with that.

I wasn't able to locally test this because I couldn't actually find any projects on prod using subpages anymore. What projects will this affect?

mathjazz:
Thank you for your comments. I'll update this next week.

This method allows us to have the same UX in the parts menu, no mater if it contains subpages or resources. It needs to be different, because subpages usually don't have 1-1 relation with resources. 

On prod, Fundraising, Firefox Tiles and Firefox Updater are good candidates to test. We'll need more of this once we start adding in-page support to more projects.

mathjazz:
Code and tests updated as suggested. Checking the first subpage for resources eliminated.

PR #180, opened 2015-09-23
Fix bug 1204047: Avoid race conditions from translations during sync.

Osmose:
When updating from VCS, only unapprove translations that existed before sync
started. This might cause multiple approved translations, so remove duplicate
approvals before committing the sync transaction.

@mathjazz r?

### DIFF
--- a/pontoon/administration/management/commands/sync_projects.py
+++ b/pontoon/administration/management/commands/sync_projects.py
@@ -3,6 +3,7 @@ from datetime import datetime
 
 from django.contrib.auth.models import User
 from django.core.management.base import BaseCommand, CommandError
+from django.db import connection, transaction
 from django.template.loader import render_to_string
 from django.utils import timezone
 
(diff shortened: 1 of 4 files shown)


pontoon/administration/management/commands/sync_projects.py
```diff
@@ -64,6 +65,11 @@ def handle(self, *args, **options):
         self.no_pull = options['no_pull']
 
         self.log('SYNC PROJECTS: start')
+
+        # Mark "now" at the start of sync to avoid messing with
+        # translations submitted during sync.
+        self.now = timezone.now()
```
mathjazz: Shouldn't we move this into handle_project()? Storing ```self.now``` here for all projects significantly increases the time window for possible duplicate approvals we have to clean up at the end. It also makes more translations commit later than necessary.

mathjazz:
This is a brilliant trick. Excellent job! r+

pontoon/administration/management/commands/sync_projects.py
Osmose: Good idea!

PR #181, opened 2015-09-23
Bug 1205445. Moved sphinx dependencies to separate file to fix builds

jotes:
This should fix docs @ readthedocs.org . You only need to set requirements file to 'requirements_docs.txt' in advanced settings.
Example build is available here: http://jotes-pontoon.readthedocs.org/en/latest/

@mathjazz @Osmose  could you review?

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,0 +1,7 @@
+# sha256: _5spJogkCW_JETkEYF2D9qqBcQLA8XARW0s58K4GUfM
+# sha256: dPYz7Tph2h0dWcMYVIPIGp1zRr8Oe18prQdkpvFZtoo
+sphinx_rtd_theme==0.1.8
+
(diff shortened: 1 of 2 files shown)


Osmose:
I don't have time to test right now but can later, but in the meantime, I'd suggest moving the requirements file to `docs/requirements.txt` to avoid cluttering the root of the repo, and also I'd recommend requiring the doc requirements in `requirements.txt` to avoid having to document how to install the separate documentation requirements. See [here](https://github.com/mozilla/funfactory/blob/master/funfactory/requirements/dev.txt#L5) for an example of what the syntax looks like.

jotes:
@Osmose yup, -r was added in requirements.txt . I've rebased PR - should reflect your suggestions. Thanks for your time :bow:

mathjazz:
This is another great fix @jotes, thank you so much!

I'm also giving @Osmose the opportunity for "more proper praise". ;)

Osmose:
r+, lgtm :D

PR #183, opened 2015-09-24
Fix bug 1208064: Use correct path for updating.

Osmose:
@mathjazz r?

It gets past the error mentioned in the bug but it fails for me locally because I don't have permission to commit to hg. Help with testing is greatly appreciated. :D


Osmose:
Oh yes, with this change you'll want to uncheck the source repo box for projects still using `commit_projects`. Source Repos is intended for the Firefox & Co repos once they're using `sync_projects`.

mathjazz:
r+

PR #186, opened 2015-09-25
Change placeables style

mathjazz:
Marking placeables with red color is not the best idea, because we use
this color to mark untranslated strings. They also stand out in the
entity list.

Instead, we put them in rounded boxes now, to make them look more like
HTML tags. Boxes also allows us to display empty space placeables. And
we use neutral UI colors.

@Osmose r?


Osmose:
Ohhhh, pretty. I love it! r+

PR #188, opened 2015-09-25
Fix bug 1201936: Handle invalid langfiles with blank translations.

Osmose:
While I was messing with the langfile tests I also added the standard format tests that we added for all the other supported formats.

@mathjazz r?

### DIFF
--- a/pontoon/base/formats/lang.py
+++ b/pontoon/base/formats/lang.py
@@ -4,9 +4,8 @@ Parser for the .lang translation format.
 import codecs
 import re
 import sys
-from collections import namedtuple
 
-from parsimonious.exceptions import ParseError as ParsimoniousParseError
+from parsimonious.exceptions import ParseError as ParsimoniousParseError, VisitationError
 from parsimonious.grammar import Grammar
 from parsimonious.nodes import NodeVisitor
 
@@ -150,6 +149,10 @@ class LangVisitor(NodeVisitor):
             tags = [m.group(1) for m in tag_matches]
             translation = translation[:tag_matches[0].start()].strip()
 
+        if translation == '':
+            raise Exception('Blank translation for key {key} is not allowed in '
+                            'langfiles.'.format(key=string))
+
         return LangEntity(string, translation, tags)
 
     def visit_string(self, node, (marker, content, end)):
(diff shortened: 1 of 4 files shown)


pontoon/base/formats/lang.py
```diff
@@ -118,9 +117,10 @@ class LangVisitor(NodeVisitor):
     def visit_lang_file(self, node, children):
         """
         Find comments that are associated with an entity and add them
-        to the entity's comments list.
+        to the entity's comments list. Also assign order to entities.
```
mathjazz: We didn't do that earlier for .lang? Wow!

pontoon/base/formats/lang.py
```diff
@@ -147,6 +149,10 @@ def visit_entity(self, node, (string, translation)):
             tags = [m.group(1) for m in tag_matches]
             translation = translation[:tag_matches[0].start()].strip()
 
+        if translation == '':
+            raise Exception('Blank translation for key {key} is not allowed in '
+                            'langfiles.'.format(key=string))
+
```
mathjazz: Nit: FHR fails with en-US already, so with original string, not translation. You might generalize this and use string instead of translation.

pontoon/base/formats/lang.py
```diff
@@ -183,8 +189,8 @@ def parse(path, source_path=None):
 
     try:
         children = LangVisitor().parse(content)
-    except ParsimoniousParseError as err:
+    except (ParsimoniousParseError, VisitationError) as err:
         wrapped = ParseError(u'Failed to parse {path}: {err}'.format(path=path, err=err))
-        raise wrapped, None, sys.exc_info()[2]
+        raise wrapped, None, sys.exc_info()[2]  # NOQA
```
mathjazz: NOQA?

mathjazz:
r+

This not only looks good but also works great with Firefox Health Report which is the project we filed this bug for in first place.

pontoon/base/formats/lang.py
Osmose: True, but this gets wrapped by a `ParseException` that includes the filename, so we'll know which locale failed and whether it's a source file or not. Also, technically, the strings in the source files are still translations, they just happen to be to the same language.

But honestly it's just that I don't want to have to add some logic for a conditional message and it won't help us debug any better, so I want to be lazy. :P

pontoon/base/formats/lang.py
Osmose: [flake8](https://flake8.readthedocs.org/en/latest/) was complaining about this syntax for raising an exception being deprecated in Python 3.2. `# NOQA` is a marker you can put on a line to tell flake8 to stop linting that specific line, and sometimes I sprinkle them around when I'm knowingly breaking a rule that I care about.

Interestingly, adding this comment didn't stop flake8 from complaining about this particular line, but I think that's a bug with my version of flake8, so I left it in.

PR #189, opened 2015-09-28
Bug 1207826 - Changed multi_locale db field into object property.

jotes:


jotes:
@mathjazz @Osmose Hi, could you review?

pontoon/base/models.py
```diff
@@ -330,12 +330,14 @@ class Repository(models.Model):
...
-        If true, this repo corresponds to multiple locale-specific repos. The
-        URL should have the string "{locale_code}" in it, which will be replaced
-        by the locale codes of all enabled locales for the project during pulls
-        and and commits.
-    """)
+
+    @property
+    def multi_locale(self):
+        """
+        Checks if url contains locale code variable. System will replace this variable by the locale codes
```
Osmose: I like to wrap comments at 72 characters per line as per [PEP8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) for readability purposes. I suggest you do that here too.

Also, you should be explicit about what the locale code variable is. Just straight up putting `{locale_code}` in the docstring is fine.

pontoon/base/models.py
```diff
@@ -373,7 +375,8 @@ def locale_checkout_path(self, locale):
         if not self.multi_locale:
             raise ValueError('Cannot get locale_checkout_path for non-multi-locale repos.')
 
-        return os.path.join(self.checkout_path, locale.code)
+        path_components = [self.project.checkout_path] + urlparse(self.url).path.split('/')
+        return os.path.join(*path_components).format(locale_code=locale.code)
```
Osmose: Why this change? `checkout_path` already includes path components from the URL and we append the URL to the end of the path instead of formatting intentionally (so that the locale code directory is at the top level of the checkout).

Osmose:
r+wc, nice work! This looks good to me except for that change to `locale_checkout_path`. Fix that up and we can probably merge this. :D

jotes:
I'll fix these things today.

pontoon/base/models.py
jotes: I've changed it because i thought about urls which don't have locale in last part of url (e.g. '/{locale_code}/smth'. Anyway, i can restore this - no problem.

jotes:
@Osmose Hi, could you look @ this again? thx!

pontoon/base/models.py
Osmose: Ah, that makes sense. We might not actually support those correctly currently, although I'm not entirely sure. The issue is that all of our code about resource paths makes an assumption that everything below the source directory (which is named en-US, en, or templates) and everything below the locale directories is identical in structure. I _think_ that assumption holds currently, but I'm not 100%.

In any case, restoring it is the right move for this PR, so we're good here.

Osmose:
This looks good to merge but Travis-CI is sad right now and won't tell me why. I'll wait until tomorrow to see if it's better and merge this. Locally the tests work fine for me, though. Nice job!

jotes:
My bad - i need to fix conflict in migrations.

On Fri, Oct 9, 2015 at 12:44 AM, Michael Kelly <notifications@github.com>
wrote:

> This looks good to merge but Travis-CI is sad right now and won't tell me
> why. I'll wait until tomorrow to see if it's better and merge this. Locally
> the tests work fine for me, though. Nice job!
>
> —
> <https://github.com/mozilla/pontoon/pull/189#issuecomment-146709028>.
>



-- 
jarek@reijutsu:~$ fortune
jarek@reijutsu:~$ fortune
You now have Asian Flu.

jotes:
@Osmose travis is green again.

Osmose:
Merged in 6458409. Nice work!

PR #190, opened 2015-09-29
Fix bug 1207829: Filter region.properties properly.

Osmose:
@mathjazz r?

Note that this isn't entirely correct, as it still uses the `repository_url` property, which defaults to the first repo defined in the database for the project. But it should function correctly for the repos as we currently have them and I've filed [bug 1209375](https://bugzilla.mozilla.org/show_bug.cgi?id=1209375) with what I think the proper long-term fix is.

### DIFF
--- a/pontoon/base/tests/test_vcs_models.py
+++ b/pontoon/base/tests/test_vcs_models.py
@@ -1,10 +1,11 @@
+import os
 import os.path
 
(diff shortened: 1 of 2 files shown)


mathjazz:
r+

This should mean you've resolved the final blocker to move the remainder of the projects to new sync.

\o/

PR #191, opened 2015-09-29
Optimize DB queries

mathjazz:
This is a quick optimization for Project.locales_parts_stats(), which affects pages using project selector (translate view and some dashboards).

It reduces the number of SQL queries by 70% and the translate view page load by 60% (compared to 274ac35abc54ceb362fb456451b76a5dbd75aade). With 100+ SQL queries, we still have room for improvement left.

@Osmose r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -287,50 +287,77 @@ class Project(models.Model):
     @property
     def locales_parts_stats(self):
         """Get project locales with their pages/paths and stats."""
-        pages = self.subpage_set.all()
-        resources = self.resource_set.filter(entity__obsolete=False).distinct()
         details = {}
 
-        for locale in self.locales.all():
-            stats = Stats.objects.filter(resource__in=resources, locale=locale)
+        def add_detail(array, path, entity, translated, approved):
+            return array.append({
+                'resource__path': path,
+                'resource__entity_count': entity,
+                'translated_count': translated,
+                'approved_count': approved
+            })
+
+        # Project Locales with prefetched Stats and Resources
+        qs = (
+            Stats.objects
+            .select_related('resource')
+            .filter(resource__project=self, resource__entity__obsolete=False)
+            .distinct()
+        )
+        locales = self.locales.prefetch_related(
+            Prefetch(
+                'stats_set',
+                queryset=qs
+            )
+        )
+
+        # Project Pages with prefetched Resources
+        pages = self.subpage_set.prefetch_related(
+            'resources'
+        )
+
+        has_pages = pages.exists()
+
+        for locale in locales:
             locale_details = []
 
-            # Get subpage name and stats if subpages defined
-            if pages.exists():
+            # If project has defined subpages, return their names with
+            # corresponding project stats. If subpages have defined resources,
+            # only include stats for page resources.
+            if has_pages:
 
                 # List only subpages, whose resources are available for locale
-                pages = pages.filter(resources__stats=stats) or pages
+                pages = pages.filter(resources__stats__in=locale.stats_set.all()) or pages
 
                 for page in pages:
+                    entity_count, translated_count, approved_count = 0, 0, 0
+                    paths = page.resources.values_list('path', flat=True)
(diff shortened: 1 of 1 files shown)


pontoon/base/models.py
```diff
@@ -284,53 +284,82 @@ def latest_activity(self, locale=None):
...
+            # only include stats for page resources.
+            if has_pages:
 
                 # List only subpages, whose resources are available for locale
-                pages = pages.filter(resources__stats=stats) or pages
+                pages = pages.filter(resources__stats__in=locale.fetched_stats) or pages
 
                 for page in pages:
+                    entity_count = translated_count = approved_count = 0
+                    paths = page.resources.values_list('path')
```
Osmose: Pass `flat=True` to get a list of the paths directly, instead of a list of 1-tuples.

pontoon/base/models.py
```diff
@@ -284,53 +284,82 @@ def latest_activity(self, locale=None):
...
-                    'resource__path',
-                    'resource__entity_count',
-                    'translated_count',
-                    'approved_count',
-                )
-
-            details[locale.code.lower()] = list(locale_details)
+                        for stats in locale.fetched_stats:
+                            if stats.resource.path not in [x[0] for x in paths]:
+                                page_stats.remove(stats)
```
Osmose: You clone the list and remove the stats that don't match, but then below you check again if the path is in the list anyway. I think that the cloning and removal can just be removed, and you can keep the `stats.resource.path in paths` check in the for loop below to keep it simple.

pontoon/base/models.py
```diff
@@ -284,53 +284,82 @@ def latest_activity(self, locale=None):
...
+            Stats.objects
+            .select_related('resource')
+            .filter(resource__project=self, resource__entity__obsolete=False)
+            .distinct()
+        )
+        locales = self.locales.prefetch_related(
+            Prefetch(
+                'stats_set',
+                queryset=qs,
+                to_attr='fetched_stats'
```
Osmose: Is there a reason to use the custom attribute instead of just using `stats_set`?

pontoon/base/models.py
Osmose: You clone the list and remove the stats that don't match, but then below you check again if the path is in the list anyway. I think that the cloning and removal can just be removed, and you can keep the `stats.resource.path in paths` check in the for loop below to keep it simple.

I suspect, based on the check for `if page.resources.exists()`, that you only want to do this filtering sometimes, but the way it's written now I think it will always do the filtering. Maybe that's what needs to change?

pontoon/base/models.py
```diff
@@ -284,53 +284,82 @@ def latest_activity(self, locale=None):
...
+            # corresponding project stats. If subpages have defined resources,
+            # only include stats for page resources.
+            if has_pages:
 
                 # List only subpages, whose resources are available for locale
-                pages = pages.filter(resources__stats=stats) or pages
+                pages = pages.filter(resources__stats__in=locale.fetched_stats) or pages
 
                 for page in pages:
+                    entity_count = translated_count = approved_count = 0
```
Osmose: If you want to avoid having three lines here, don't rely on assignment returning a value. Use destructuring instead:

```python
entity_count, translated_count, approved_count = 0, 0, 0
```

pontoon/base/models.py
```diff
@@ -284,53 +284,82 @@ def latest_activity(self, locale=None):
 
         return translations.latest_activity()
 
+    def add_detail(self, array, path, entity, translated, approved):
```
Osmose: `add_detail` is an odd method to add to the `Project` class because it doesn't actually rely on the project instance directly. I think this'd be better directly in the function, even if it's repeated twice (or maybe define an anonymous function within the other method?)

Osmose:
r+wc, nice work! Glad to see an improvement in the number of DB queries. :D

mathjazz:
Merged in bdd10869a1c5bc62cdd49f6ec59e7b718c2905ff.

PR #192, opened 2015-09-29
Fix project parts tests

mathjazz:
1) If only one resource defined, don't show it
2) Subpage names are now sorted alphabetically

@Osmose r?


Osmose:
r+, closer to fixed tests!

PR #194, opened 2015-09-30
Fix bug 1209423: Correctly handle MOZ_LANGPACK_CONTRIBUTORS.

Osmose:
When importing strings from VCS, only uncomment the entity when parsing
source resources. When exporting strings, only uncomment the entity if
Pontoon has a translation for it.

@mathjazz r?


mathjazz:
Looks good, works locally! r+

PR #195, opened 2015-09-30
Optimize locales_parts_stats()

mathjazz:
To actually make it usable on production. This cuts DB requests in half.
I tested it with prod DB and the translate view loads even faster than
current production instance. By 20-30%.

@Osmose r?

--

I wanted to fix this, not so much because subpage stats are an important feature, but because it also partially blocks bug 1178880.

In any case it will be interesting to observe how this scales. We might need to consider loading project and locale data only and using AJAX to get resources, pages and stats. We could probably also load all these data for the requested and localizers' locale.

### DIFF
--- a/pontoon/base/tests/test_models.py
+++ b/pontoon/base/tests/test_models.py
@@ -98,11 +98,20 @@ class ProjectPartsTests(TestCase):
         EntityFactory.create(resource=self.resource)
         StatsFactory.create(resource=self.resource, locale=self.locale)
 
+    def _fetch_locales_parts_stats(self):
+        # Fake Prefetch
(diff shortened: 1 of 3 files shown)


pontoon/base/tests/test_models.py
```diff
@@ -98,11 +98,20 @@ def setUp(self):
         EntityFactory.create(resource=self.resource)
         StatsFactory.create(resource=self.resource, locale=self.locale)
 
+    def _call_test_method(self):
```
Osmose: `fetch_locales_parts_stats` is a better name for this, it explains what it's actually doing.

Osmose:
r+, looks fine to me, let's see how it does on prod! :D

PR #196, opened 2015-09-30
Fix bug 1191718: Don't sync when there are no changes.

Osmose:
Now we track the revision number/hash from the last time we synced, and
use that to determine if VCS has changed since the last sync. If not,
and there are no changes in the DB, we can skip syncing completely.

If things fail, sync should default to assuming changes exist, degrading gracefully to the old behavior.

@mathjazz r?

### DIFF
--- a/pontoon/administration/management/commands/sync_projects.py
+++ b/pontoon/administration/management/commands/sync_projects.py
@@ -92,7 +92,19 @@ class Command(BaseCommand):
 
         # Pull changes from VCS and update what we know about the files.
         if not self.no_pull:
-            self.pull_changes(db_project)
+            repos_changed = self.pull_changes(db_project)
+        else:
+            repos_changed = True  # Assume changed.
+
+        # If the repos haven't changed since the last sync and there are
+        # no Pontoon-side changes for this project, quit early.
+        db_changed = (ChangedEntityLocale.objects
+                      .filter(entity__resource__project=db_project)
+                      .exists())
+        if not repos_changed and not db_changed:
+            self.log('Skipping project {0}, no changes detected.'.format(db_project.slug))
+            return
+
         vcs_project = VCSProject(db_project)
         self.update_resources(db_project, vcs_project)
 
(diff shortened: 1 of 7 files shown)


pontoon/administration/management/commands/sync_projects.py
```diff
@@ -212,9 +224,25 @@ def entity_key(self, entity):
...
+            repo_revisions = repo.pull()
+
+            # If any revision is None, we can't be sure if a change
+            # happened or not, so we default to assuming it did.
+            unsure_change = None in repo_revisions.values()
+            if unsure_change or repo_revisions != repo.last_synced_revisions:
+                changed = True
+
+            repo.last_synced_revisions = repo_revisions
+            repo.save()
```
mathjazz: You want to hit the DB only if values aren't equal.

mathjazz:
Nice work, r+ on code. This should make sync much faster, because log commits are full of "Nothing to commit" messages.

We could probably optimize this even further for per-locale repos and only sync locales with new revisions.

pontoon/administration/vcs.py
```diff
@@ -275,3 +275,69 @@ def get_svn_env():
...
+        code, output, error = execute(cmd, cwd=cwd, env=env)
+        if log_errors and code != 0:
+            log.error('Error while executing command `{cmd}` in `{cwd}`: {stderr}'.format(
+                cmd=unicode(cmd), cwd=cwd, stderr=error
+            ))
+        return code, output, error
+
+
+class SvnRepository(VCSRepository):
+    def execute(self, cmd, cwd=None, env=None):
```
mathjazz: Add missing argument: log_errors=False.

mathjazz:
Nice work, r+wc for the missing argument. Tested locally with different projects, repository types and VCS types and works fine.

This should make sync much faster, because log commits are full of "Nothing to commit" messages.

One day, we could probably optimize this even further for per-locale repos and only sync locales with new revisions.

PR #198, opened 2015-10-01
Bug 1178880 project locale dashboard with search

mathjazz:
This brings search across entire project, one of the most requested features by localizers. Full description is available as part of 3 commit messages. (I don't know how 9289678 ended up there, please ignore it before I remove it from this PR.)

@Osmose, I'd like to hear your feedback.

I decided to stick with my original plan of having joint Search and Resources/Subpages menu. Having separate menus is problematic, because it doesn't allow selecting resources when you're on the results page. I will however change the design to put resource selection and search in tabs.

For now it only searches original strings. I'd like to add options for searching translations, suggestions, comments, keys, and search case sensitive.

I'll add and update tests before asking for full code review.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -194,6 +194,20 @@ class ProjectQuerySet(models.QuerySet):
         """
         return self.filter(disabled=False, resource__isnull=False)
 
+    def prefetch_parts(self):
+        """
+        Prefetch subpages and resources for locales_parts_stats().
+        """
+        return self.prefetch_related(
+            'subpage_set',
+            Prefetch(
+                'resource_set',
+                queryset=Resource.objects.filter(
+                    pk__in=Entity.objects.filter(obsolete=False).values('resource')
+                ),
+                to_attr='active_resources'
+            )
+        )
 
 class Project(models.Model):
     name = models.CharField(max_length=128, unique=True)
@@ -288,36 +302,39 @@ class Project(models.Model):
         else:
             return repo
 
-    def latest_activity(self, locale=None):
+    def latest_activity(self, locale=None, path=None):
         """Get latest activity for project and locale if provided."""
         translations = Translation.objects.filter(entity__resource__project=self)
         if locale:
             translations = translations.filter(locale=locale)
+            if path:
+                translations = translations.filter(entity__resource__path=path)
 
         return translations.latest_activity()
 
-    @property
-    def locales_parts_stats(self):
+    def locales_parts_stats(self, loc=None):
         """Get project locales with their pages/paths and stats."""
         def get_details(stats):
             return stats.order_by('resource__path').values(
                 'resource__path',
                 'resource__entity_count',
+                'fuzzy_count',
                 'translated_count',
                 'approved_count',
             )
 
+        details = {}
         pages = self.subpage_set.all()
         resources = self.active_resources
-        details = {}
+        locales = [loc] if loc else self.locales.all()
 
-        for locale in self.locales.all():
+        for locale in locales:
             stats = Stats.objects.filter(resource__in=resources, locale=locale)
             locale_details = []
 
             # Is subpages aren't defined and project uses more than one
             # resource, return resource paths with corresponding resource stats
-            if len(pages) == 0 and len(resources) > 1:
+            if len(pages) == 0:
                 locale_details = get_details(stats)
 
             # If project has defined subpages, return their names with
@@ -331,6 +348,7 @@ class Project(models.Model):
                         pages.filter(resources__stats__locale=locale).annotate(
                             resource__path=F('name'),
                             resource__entity_count=F('resources__entity_count'),
+                            fuzzy_count=F('resources__stats__fuzzy_count'),
                             translated_count=F('resources__stats__translated_count'),
                             approved_count=F('resources__stats__approved_count')
                         )
@@ -341,12 +359,16 @@ class Project(models.Model):
                         pages.annotate(
                             resource__path=F('name'),
                             resource__entity_count=F('project__resource__entity_count'),
+                            fuzzy_count=F('project__resource__stats__fuzzy_count'),
                             translated_count=F('project__resource__stats__translated_count'),
                             approved_count=F('project__resource__stats__approved_count')
                         )
                     )
 
             details[locale.code.lower()] = list(locale_details)
+
+        if loc:
+            details = list(locale_details)
 
         return details
 
@@ -557,7 +579,7 @@ class Entity(DirtyFieldsMixin, models.Model):
     resource = models.ForeignKey(Resource)
     string = models.TextField()
     string_plural = models.TextField(blank=True)
-    key = models.TextField(blank=True)  # Needed for webL10n
+    key = models.TextField(blank=True)
     comment = models.TextField(blank=True)
     order = models.PositiveIntegerField(default=0)
     source = JSONField(blank=True, default=list)  # List of paths to source code files
@@ -620,7 +642,7 @@ class Entity(DirtyFieldsMixin, models.Model):
             }
 
     @classmethod
-    def for_project_locale(self, project, locale, paths=None):
+    def for_project_locale(self, project, locale, paths=None, search=None):
         """Get project entities with locale translations."""
         entities = self.objects.filter(
             resource__project=project,
@@ -628,7 +650,28 @@ class Entity(DirtyFieldsMixin, models.Model):
             obsolete=False
         )
 
-        if paths:
+        # Filter by search parameters
+        if search:
+            keyword = search.get('keyword', None)
+            i = '' if search.get('casesensitive', None) else 'i'
+
+            entity_query = Q()  # Empty object
+            if search.get('sources', None):
+                entity_query |= Q(**{'string__%scontains' % i: keyword}) | Q(**{'string_plural__%scontains' % i: keyword})
+
+            if search.get('translations', None):
+                entity_query |= Q(**{'translation__string__%scontains' % i: keyword})
+
+            if search.get('comments', None):
+                entity_query |= Q(**{'comment__%scontains' % i: keyword})
+
+            if search.get('keys', None):
+                entity_query |= Q(**{'key__%scontains' % i: keyword})
+
+            entities = entities.filter(entity_query).distinct()
+
+        # Filter by path
+        elif paths:
             try:
                 subpage = Subpage.objects.get(project=project, name__in=paths)
                 paths = subpage.resources.values_list("path")
--- a/pontoon/base/static/css/style.css
+++ b/pontoon/base/static/css/style.css
@@ -388,7 +388,7 @@ label {
   background: #3f4752;
   font-size: 16px;
   height: 20px;
-  margin: 10px 10px 10px 0px;
+  margin: 10px 10px 10px 0;
   overflow: hidden;
   padding: 10px 20px 10px 40px;
   text-overflow: ellipsis;
@@ -455,12 +455,16 @@ label {
   color: #272A2F;
 }
 
-.project.select {
+.project.select,
+.locale.select,
+.part.select {
   padding-left: 0;
   width: 100%;
 }
 
-.project.select .menu {
+.project.select .menu,
+.locale.select .menu,
+.part.select .menu {
   bottom: auto;
   display: inline-block;
   position: relative;
@@ -470,18 +474,23 @@ label {
   box-sizing: border-box;
 }
 
-.project.select input[type="search"] {
+.project.select input[type="search"],
+.locale.select input[type="search"],
+.part.select input[type="search"] {
   width: 100%;
 
   -moz-box-sizing: border-box;
   box-sizing: border-box;
 }
 
-.project.select .menu ul {
+.project.select .menu ul,
+.locale.select .menu ul,
+.part.select .menu ul {
   max-height: none;
 }
 
-.project.select .menu ul li .name {
+.project.select .menu ul li .name,
+.part.select .menu ul li .name {
   overflow: hidden;
   padding-left: 1px; /* Insted of 0; because of overflow */
   text-overflow: ellipsis;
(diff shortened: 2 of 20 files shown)


Osmose:
I think the use of icons in the navigation is slick to signify a search vs a resource, although I think the icon should be on the left instead of the right.

The only problem I see is that it's unclear that "search the entire project" is a button that opens up a different search bar. It looks like you enter in the search query, then click that link or hit enter. It took me a few tries to figure things out.

I'm still of the opinion that combining the two is confusing (and it did confuse me), but if you've made the decision to do things that way then this looks nice. Great work!

mathjazz:
Yes, this button is totally confusing, which is why I'll redesign it tomorrow and use tabs.

I'm also considering dropping the locale-project dashboard. At least for now. The only added value is last activity column and more detailed progress. And it probably makes search entire project more obvious. I don't know, need to slep on it.

mathjazz:
Using tabs now in the resources menu and also in the locale-project dashboard. We should use something similar for dashboard subpages (e.g. #201). Tabs looks quite boring, we should iterate and make them sexier.

I also added search options (sources, translations, case-sensitive, etc.).

Tests are fixed, but I'll add a few more before merging. I want to hear if the UX is any better now and the actual code doesn't suck too much.

@Osmose r?

Osmose:
UX wise this is way better, I like it a lot more. Nice work!

pontoon/base/static/js/locale_project.js
```diff
@@ -0,0 +1,65 @@
...
+    $('.part .menu input[type=search]:visible').trigger("keyup").focus();
+  });
+
+  // If showing search results, show search pane by default
+  var search = $('#server').data("search"),
+      tab = 'resources';
+  if (search) {
+    tab = 'search-project';
+  }
+  $('.part .menu nav a[href=#' + tab + ']').click();
```
Osmose: It's fine for this PR because I still think the frontend needs a major refactoring rather than incremental change, but one of the anti-patterns I've seen a lot in Pontoon is simulating events like this to trigger other functions. What you should really be doing is putting the logic that you want to reuse in a function and having both handlers call that function instead. Simulating events like this makes it very hard to figure out what piece of code is being executed when this line is called.

Osmose:
I think I'm going to take this opportunity to point out some of the issues I've seen in Pontoon's frontend code, but I don't think it's worth changing any of the frontend code for this PR, so don't worry about fixing any of the frontend issues.

pontoon/base/static/css/style.css
```diff
@@ -489,34 +498,49 @@ label {
...
-  display: inline-block;
-  position: relative;
-  width: 100%;
+.part .menu nav ul li.hover {
+  background: transparent;
+}
 
-  -moz-box-sizing: border-box;
-  box-sizing: border-box;
+.part .menu nav ul li.active {
```
Osmose: Selectors like this make it really hard to understand what this is actually styling. The class names are very brief, and it uses tag selectors, which generally are a bad idea because they make it hard to make changes to the HTML and are hard for a developer to remember VS a descriptive class name.

Granted, I've been searching for years for a good guide to CSS class naming because I've been disappointed with how I do it and haven't found it yet, so maybe it's impossible. :P

pontoon/base/static/js/locale_project.js
```diff
@@ -0,0 +1,65 @@
+$(function() {
+
+  // Search entire project with Enter (and remove default search handler)
+  $('.part .menu .search-project input[type=search]').off('keyup.search').keydown(function (e) {
```
Osmose: Instead of using `off` to un-bind the older event handler, you should instead fix the selector for the other event handler such that this element doesn't get bound in the first place. Otherwise the programmer has to keep in their head every single event handler in the site in order to figure out what events are bound to which elements.

Osmose: Also, semantically it's better to wrap the search field in a `<form>` element so that everything related to submitting the search is grouped together. Then, you can hook on the submit event of the form, which avoids running an event handler on every keypress and allows for submission of the form in ways besides hitting enter.

pontoon/base/static/js/locale_project.js
```diff
@@ -0,0 +1,65 @@
...
+    if (e.which === 13) {
+      e.stopPropagation();
+
+      var q = $(this).val(),
+          action = '';
+
+      if (q.length > 0) {
+        if ($('.locale .selector .language').length) {
+          var locale = $('.locale .selector .language').attr('class').split(' ')[1],
```
Osmose: In general it's a bad idea to store actual data in classnames or other non-data attributes, because the form of your data should not have the side-effect on how the HTML is displayed or functions.

pontoon/base/static/js/locale_project.js
```diff
@@ -0,0 +1,65 @@
...
+              project = $('.project .selector .title').data('slug'),
+              action = '/' + locale + '/' + project + '/';
+        }
+
+        var options = '';
+        $('.search-project .options li').each(function() {
```
Osmose: The `<form>` comment above would also help here, as the search options would just be checkboxes included in the form's data.

pontoon/base/static/js/locale_project.js
```diff
@@ -0,0 +1,65 @@
...
+            options += '&' + $(this).data('name') + '=' + $(this).data('name');
+          }
+        });
+
+        // At least one option must be selected
+        if (!options) {
+          return false;
+        }
+
+        var post = $('<form>', {
```
Osmose: And, well, continuing on the theme of using a `<form>` directly in the HTML, you'd be able to avoid all this. In any situation creating a `<form>` only to immediately append and submit it is an anti-pattern.

pontoon/base/static/js/translate_old.js
```diff
@@ -1839,24 +1821,40 @@ var Pontoon = (function (my) {
...
      * Initialize Pontoon: load entities, store data, prepare UI
      */
     initialize: function(advanced, project) {
-      var self = this;
+      var self = this,
+          search = $('#server').data('search'),
+          params =
+            'project=' + $('#server').data('id') +
+            '&locale=' + $('#server').data('locale').code +
+            '&paths=' + JSON.stringify(/* Pontoon.paths || */[$('header .part .selector').attr('title')]) +
```
Osmose: Remove the comment.

pontoon/base/tests/test_views.py
```diff
@@ -97,14 +98,13 @@ def test_no_subpage_multiple_stats_in_current_locale(self):
...
 
-        # Need at least two resources and stats to trigger setting the part value.
         resource1 = ResourceFactory.create(project=project, path='foo1.lang', entity_count=1)
         resource2 = ResourceFactory.create(project=project, path='foo2.lang', entity_count=1)
         StatsFactory.create(resource=resource1, locale=locale)
         StatsFactory.create(resource=resource2, locale=locale)
 
         self.client_login()
-        url = '/{locale.code}/{project.slug}/'.format(locale=locale, project=project)
+        url = '/{locale.code}/{project.slug}/{resource.path}/'.format(locale=locale, project=project, resource=resource1)
```
Osmose: Maybe `/`.join instead?

pontoon/base/views.py
```diff
@@ -244,37 +287,33 @@ def translate(request, locale, slug, part=None, template='translate.html'):
...
-    # Set part if subpages not defined and entities in more than one file
+    # Set path if subpages not defined
     else:
         paths = (Resource.objects
                     .filter(project=p, entity_count__gt=0, stats__locale=l)
                     .order_by('path')
                     .values_list('path', flat=True))
 
-        if len(paths) > 1:
+        if len(paths) > 0:
```
Osmose: Maybe `if paths:`?

pontoon/base/views.py
```diff
@@ -158,9 +159,51 @@ def projects(request, template='projects.html'):
     return render(request, template, data)
 
 
-def translate(request, locale, slug, part=None, template='translate.html'):
+def locale_project(request, locale, slug):
+    """Locale-project overview."""
+
+    # Redirect to translate view
+    if request.GET.get('q'):
```
Osmose: Am I correct in thinking that this is to handle the search submission? If so, just make the form for the search submit directly to the translate view, this shouldn't be necessary.

pontoon/base/views.py
```diff
@@ -158,9 +159,51 @@ def projects(request, template='projects.html'):
...
+        return translate(request, locale, slug, '')
+
+    try:
+        l = Locale.objects.get(code__iexact=locale)
```
Osmose: Check out [one of my favorite shortcut functions](https://docs.djangoproject.com/en/1.8/topics/http/shortcuts/#get-object-or-404).

pontoon/base/views.py
```diff
@@ -158,9 +159,51 @@ def projects(request, template='projects.html'):
...
+        raise Http404
+
+    # Prefetch subpages and resources for locales_parts_stats()
+    projects = (
+        Project.objects.filter(
+            disabled=False,
+            slug=slug,
+            pk__in=Resource.objects.values('project')
+        )
+        .prefetch_related(
```
Osmose: You've officially written this prefetch twice, time to refactor it to a queryset method!

pontoon/base/views.py
```diff
@@ -158,9 +159,51 @@ def projects(request, template='projects.html'):
...
+            )
+        )
+    )
+
+    if not projects.exists():
+        raise Http404
+
+    data = {
+        'locale': l,
+        'project': projects[0],
```
Osmose: Don't bother with the `filter` and indexing if you're just going to take a single project. Use `get_object_or_404` and just provide a queryset set up with the prefetching to it.

pontoon/base/views.py
```diff
@@ -244,37 +287,33 @@ def translate(request, locale, slug, part=None, template='translate.html'):
...
                     .filter(project=p, entity_count__gt=0, stats__locale=l)
                     .order_by('path')
                     .values_list('path', flat=True))
 
-        if len(paths) > 1:
+        if len(paths) > 0:
             data['part'] = part if part in paths else paths[0]
 
+    # Set search keyword
+    data['search'] = request.GET.urlencode()
```
Osmose: I'm having a bit of trouble finding where you use this outside of checking if a search was submitted, which should be something you should be able to determine from the URL without having the server pass it back. Am I missing something?

pontoon/base/views.py
```diff
@@ -408,7 +447,11 @@ def entities(request, template=None):
         log.error(str(e))
         return HttpResponse("error")
 
-    entities = Entity.for_project_locale(project, locale, paths)
+    search = None
+    if request.GET.get('q', None):
```
Osmose: This needs a better parameter name than `q`.

pontoon/base/models.py
```diff
@@ -276,36 +276,39 @@ def repository_for_path(self, path):
...
         if locale:
             translations = translations.filter(locale=locale)
+            if path:
+                translations = translations.filter(entity__resource__path=path)
 
         return translations.latest_activity()
 
-    @property
-    def locales_parts_stats(self):
+    def locales_parts_stats(self, loc=None):
```
Osmose: It's awkward to have this method living on the project model, especially with a parameter that can limit it to a single locale. It's starting to look more like a method for the locale queryset that takes in a project to filter the stats by. That way you can better control which locales you want to get stats from.

pontoon/base/models.py
```diff
@@ -276,36 +276,39 @@ def repository_for_path(self, path):
         else:
             return repo
 
-    def latest_activity(self, locale=None):
+    def latest_activity(self, locale=None, path=None):
```
Osmose: This is fine for now, but as we find a need for more options and filters on the set of translations I wonder about how appropriate it is to have these helpers on the Project and Locale models, especially now that they're diverging in the arguments they support.

pontoon/base/models.py
```diff
@@ -627,7 +635,30 @@ def for_project_locale(self, project, locale, paths=None):
...
+            if search.get('comments', None):
+                entity_query |= Q(**{'comment__%scontains' % i: q})
+
+            if search.get('keys', None):
+                entity_query |= Q(**{'key__%scontains' % i: q})
+
+            entities = entities.filter(entity_query).distinct()
+
+        # Filter by path
+        elif paths:
```
Osmose: I don't have a good solution so we can keep it for now, but having two optional parameters that are mutually exclusive is a bad code smell. I think sometime in the future we'll want to refactor this into multiple methods or something.

pontoon/base/models.py
```diff
@@ -627,7 +635,30 @@ def for_project_locale(self, project, locale, paths=None):
...
+        if search:
+            q = search.get('q', None)
+            i = '' if search.get('casesensitive', None) else 'i'
+
+            entity_query = Q()  # Empty object
+            if search.get('sources', None):
+                entity_query |= Q(**{'string__%scontains' % i: q}) | Q(**{'string_plural__%scontains' % i: q})
+
+            if search.get('translations', None):
+                entity_query |= Q(translation__in=translations.filter(**{'string__%scontains' % i: q}))
```
Osmose: No need for the translations subquery here, just check directly:

```python
entity_query |= Q(translation__string__contains=q)
```

pontoon/base/models.py
```diff
@@ -627,7 +635,30 @@ def for_project_locale(self, project, locale, paths=None):
             obsolete=False
         )
 
-        if paths:
+        translations = Translation.objects.filter(locale=locale)
+
+        # Filter by search parameters
+        if search:
+            q = search.get('q', None)
+            i = '' if search.get('casesensitive', None) else 'i'
```
Osmose: This is a bit awkward but again I have no better solution. Is case sensitivity useful anyway? Could we maybe just make all the searches case-insensitive by default?

pontoon/base/static/css/style.css
Osmose: I asked craigcook and he recommended, among other things https://smacss.com/. I'll be checking it out to see how I feel about it. :D

pontoon/base/models.py
mathjazz: They are case-insensitive by default. Still, it's useful sometimes to have an option to make case sensitive searches.

pontoon/base/models.py
mathjazz: This method needs to be refactored for performance reasons soon, so I'd rather not touch it at this stage. That being said, it might be that we hit performance issues again (will test on stage), so I'll have to refactor it before merging this PR anyways.

pontoon/base/views.py
mathjazz: Yeah, I thought setting a keyword was more transparent, but on the other hand it's redundant.

pontoon/base/views.py
mathjazz: Yes, and we were using the same URL for search as locale_project uses. I'll add a '/search/' piece to it now.

pontoon/base/views.py
mathjazz: I'm officially embarrassed.

mathjazz:
@Osmose this is ready for another look.

mathjazz:
Hm, do we even need the locale-project dashboard, now that the search is also available in the translate view?

Sure, it allows for sorting resources by A) Latest activity and B) Progress, which is currently the best way of finding files with A) recent changes (e.g. translations to review) and B) missing translations.

But on the other hand, we could have all these information available in the translate view (we only need to add latest activity column and enable sorting). Plus, having so many dashboards and steps to get to the translate view is tedious (1. Teams - 2. Team - 3. Project - 4. Locale-project - 5. Translate OR 1. Projects - 2. Project - 3. team - 4. Locale-project - 5. Translate).

What do you think? Maybe we can leave it as it is for now and than iterate.

Osmose:
> What do you think? Maybe we can leave it as it is for now and than iterate.

I think we should pick whichever choice takes the least amount of effort (IE since the dashboard is already coded, just roll with it) and save our effort for the UX refresh we're gunning for. Any new ideas about how to structure the site we have now will be affected by that, and it's better to move ourselves closer to starting that than it is to keep iterating.

Osmose:
r+ contingent on why that form is serialized and stuck into the `action` attribute. Nice work!

PR #200, opened 2015-10-01
Fix 1204874: Be smarter about detecting source directories.

Osmose:
Source directory names now have ranks based on how probable it is
that they're the directory we want, and we also score higher for
directories with source-only files (like po templates).

I figured implementing this via scoring instead of directly wouldn't be that much more complicated and sets us up well in case we ever need to handle more special cases. It also matches how we think about the problem better.

@mathjazz r?


pontoon/base/models.py
```diff
@@ -408,7 +408,7 @@ def checkout_path(self):
             path_components = [c for c in path_components if c != '{locale_code}']
 
         if self.source_repo:
-            path_components.append('en-US')
+            path_components.append('templates')
```
mathjazz: Good move!

mathjazz:
r+

Bonus score (pun intended) for the scoring system!

PR #201, opened 2015-10-04
Bug 1206779 - added contributors subpages to project and locale views.

jotes:
@Osmose @mathjazz Hi, could you review?

I've removed validation if locale contains projects -> i'm not sure if that's really required.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -49,7 +49,7 @@ class UserTranslationsManager(models.Manager):
             Case(
                 When(translation_query, then=1), output_field=models.IntegerField(), default=0))
 
-    def with_translation_counts(self, start_date=None, limit=100):
+    def with_translation_counts(self, start_date=None, query_filters=None, limit=100):
         """
         Returns contributors list, sorted by count of their translations.
         Every user instance has added following properties:
@@ -61,12 +61,18 @@ class UserTranslationsManager(models.Manager):
         count of sql queries during generation of metrics.
         All counts will be returned from start_date to now().
         :param date start_date: start date for translations.
+        :param django.db.models.Q query_filters: filters contributors by given query_filters.
+        :param int limit: limit results to this number.
         """
         def translations_count(query=None):
             """Short helper to avoid duplication of passing dates."""
             query = query or Q()
             if start_date:
                 query &= Q(translation__date__gte=start_date)
+        
+            if query_filters:
+                query &= query_filters
+
             return self._changed_translations_count(query)
         return (
             self
--- a/pontoon/base/templates/locale.html
+++ b/pontoon/base/templates/locale.html
@@ -1,5 +1,7 @@
 {% extends "landing.html" %}
 
+{% import "widgets/menu.html" as Menu %}
+
 {% block title %}Pontoon &middot; {{ locale.name }} ({{ locale.code }}){% endblock %}
 
 {% block class %}locale{% endblock %}
@@ -13,7 +15,16 @@
 			<p>{{ locale.team_description|safe }}</p>
 		</div>
 	{% endif %}
+</form>
 
-	{% include 'project_selector.html' %}
+{{ Menu.submenu((
+	(url('pontoon.locale', locale.code), 'Projects', ''),
+	(url('pontoon.locale.contributors', locale.code), 'Contributors', 'contributors')
+), request.path.split('/')[2]|default('')) }}
+
+<form>
+	{% block locale_subpage %}
+		{% include 'project_selector.html' %}
+	{% endblock %}
 </form>
 {% endblock %}
--- a/pontoon/base/templates/locale_contributors.html
+++ b/pontoon/base/templates/locale_contributors.html
@@ -1,0 +1,18 @@
+{% extends 'locale.html' %}
+{% import 'widgets/contributors.html' as Contributors %}
+
+{% block title %}{{ super() }} &middot; Contributors {% endblock %}
+
+{% block class %}locale{% endblock %}
+
+{% block locale_subpage %}
+	{{ Contributors.select_period(period, url('pontoon.locale.contributors', locale.code)) }}
+
+	<form>
+	  {{ Contributors.list_contributors(contributors) }}
+	</form>
+{% endblock %}
+
+{% block extend_css %}
+  {% stylesheet 'users' %}
+{% endblock %}
--- a/pontoon/base/templates/project.html
+++ b/pontoon/base/templates/project.html
@@ -1,4 +1,5 @@
 {% extends "landing.html" %}
+{% import "widgets/menu.html" as Menu %}
 
 {% block title %}Pontoon &middot; {{ project.name }}{% endblock %}
 
@@ -24,7 +25,17 @@
 	{% if project.info_brief %}<p>{{ project.info_brief|safe }}</p>{% endif %}
 	{% if project.url %}<p><span>Website: </span><a href="{{ project.url }}">{{ project.url }}</a></p>{% endif %}
   </div>
-  {% include 'locale_selector.html' %}
+</form>
+
+{{ Menu.submenu((
+  (url('pontoon.project', project.slug), 'Teams', ''),
+  (url('pontoon.project.contributors', project.slug), 'Contributors', 'contributors')
+), request.path.split('/')[3]|default('')) }}
+
+<form>
+  {% block project_subpage %}
+    {% include 'locale_selector.html' %}
+  {% endblock %}
 </form>
 {% endblock %}
 
--- a/pontoon/base/templates/project_contributors.html
+++ b/pontoon/base/templates/project_contributors.html
@@ -1,0 +1,16 @@
+{% extends "project.html" %}
+{% import 'widgets/contributors.html' as Contributors %}
+
+{% block title %}{{ super() }} &middot; Contributors {% endblock %}
+
+{% block project_subpage %}
+{{ Contributors.select_period(period, url('pontoon.project.contributors', project.slug)) }}
+
+<form>
+  {{ Contributors.list_contributors(contributors) }}
+</form>
+{% endblock %}
+
+{% block extend_css %}
+  {% stylesheet 'users' %}
+{% endblock %}
--- a/pontoon/base/templates/widgets/contributors.html
+++ b/pontoon/base/templates/widgets/contributors.html
@@ -1,0 +1,63 @@
+{% import "widgets/menu.html" as Menu %}
+
+{# Widget to select period for filtering contributors #}
+{% macro select_period(period, viewurl) %}
+{{
+	Menu.submenu((
+		(viewurl, 'All time', None),
+		(viewurl + '?period=12', '12 months', 12),
+		(viewurl + '?period=6', '6 months', 6),
+		(viewurl + '?period=3', '3 months', 3),
+		(viewurl + '?period=1', 'Last month', 1),
+	), period)
+}}
+{% endmacro %}
+
+{# Widget to display list of contributors. #}
+{% macro list_contributors(contributors) %}
+  <table>
+    <thead>
+      <tr>
+        <th>Rank</th>
+        <th>Contributor</th>
+        <th>Translations<sup><a href="#fn-translations" id="ref-translations">*</a></sup></th>
+      </tr>
+    </thead>
+    <tbody>
+    {% for contributor in contributors %}
+      <tr>
+        <td class="rank">{{ loop.index }}</td>
+        <td class="contributor">
+          <a href="{{ url('pontoon.contributor', contributor.email) }}">
+            <img class="rounded" src="{{ contributor.gravatar_url(44) }}">
+            <p class="name">{{ contributor.first_name or contributor.email.split('@')[0] }}</p>
+          </a>
+          <p class="email">{{ contributor.email }}</p>
+        </td>
+        <td class="stats">
+          <div class="details">
+            <div class="total">
+              <span>Total</span>
+              <p>{{ contributor.translations_count }}</p>
+            </div><div class="approved">
+              <span>Approved</span>
+              <p>{{ contributor.translations_approved_count }}</p>
+            </div><div class="translated">
+              <span>Unapproved</span>
+              <p>{{ contributor.translations_unapproved_count }}</p>
+            </div><div class="fuzzy">
+              <span>Needs work</span>
+              <p>{{ contributor.translations_needs_work_count }}</p>
+            </div>
+          </div>
+        </td>
+      </tr>
+    {% endfor %}
+    </tbody>
+    <tfoot>
+      <tr>
+        <td colspan="3"><sup id="fn-translations">Translation counts do not include unchanged translations.<a href="#ref-translations" title="Jump back to top.">↩</a></sup></td>
+      </tr>
+    </tfoot>
+  </table>
+{% endmacro %}
--- a/pontoon/base/templates/widgets/menu.html
+++ b/pontoon/base/templates/widgets/menu.html
@@ -1,0 +1,12 @@
+{# defines menu widget #}
+{% macro submenu(choices=(), value=None, menuclass='submenu') %}
+<div class="{{ menuclass }}">
+  <ul class="links">
+    {% for choiceurl, label, val in choices %}
+    <li {% if val == value %}class="active"{% endif %}>
+      <a href="{{ choiceurl }}">{{ label }}</a>
+    </li>
+    {% endfor %}
+  </ul>
+</div>
+{% endmacro %}
--- a/pontoon/base/tests/__init__.py
+++ b/pontoon/base/tests/__init__.py
@@ -108,6 +108,7 @@ class ResourceFactory(DjangoModelFactory):
     project = SubFactory(ProjectFactory)
     path = '/fake/path.po'
     format = 'po'
+    entity_count = 1
 
     class Meta:
         model = Resource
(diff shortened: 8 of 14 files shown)


pontoon/base/templates/user.html
```diff
@@ -1,4 +1,4 @@
-{% extends "landing.html" %}
+\{% extends "landing.html" %}
```
mathjazz: Whoops?

jotes: Whoops :D added to fix list.

On Tue, Oct 6, 2015 at 5:27 PM, Matjaž Horvat <notifications@github.com>
wrote:

> In pontoon/base/templates/user.html
> <https://github.com/mozilla/pontoon/pull/201#discussion_r41279225>:
>
> > @@ -1,4 +1,4 @@
> > -{% extends "landing.html" %}
> > +\{% extends "landing.html" %}
>
> Whoops?
>
> —
> <https://github.com/mozilla/pontoon/pull/201/files#r41279225>.
>



-- 
jarek@reijutsu:~$ fortune
jarek@reijutsu:~$ fortune
You now have Asian Flu.

pontoon/base/templates/locale_contributors.html
```diff
@@ -0,0 +1,18 @@
+{% extends 'locale.html' %}
+{% import 'widgets/contributors.html' as Contributors %}
+
+{% block title %}{{ super() }} - contributors {% endblock %}
```
mathjazz: Instead of dash and lowercase (- contributors), let's keep the pattern and use: &middot; Contributors.

pontoon/base/templates/project.html
```diff
@@ -24,7 +25,14 @@
 	{% if project.info_brief %}<p>{{ project.info_brief|safe }}</p>{% endif %}
 	{% if project.url %}<p><span>Website: </span><a href="{{ project.url }}">{{ project.url }}</a></p>{% endif %}
   </div>
-  {% include 'locale_selector.html' %}
+  {{ Menu.submenu((
+    (url('pontoon.project', project.slug), 'Locale', ''),
```
mathjazz: Locale -> Teams

pontoon/base/templates/project_contributors.html
```diff
@@ -0,0 +1,16 @@
+{% extends "project.html" %}
+{% import 'widgets/contributors.html' as Contributors %}
+
+{% block title %}{{ super() }} - contributors {% endblock %}
```
mathjazz: Instead of dash and lowercase (- contributors), let's keep the pattern and use: &middot; Contributors.

pontoon/base/templates/locale.html
```diff
@@ -13,7 +15,13 @@
 			<p>{{ locale.team_description|safe }}</p>
 		</div>
 	{% endif %}
+	{{ Menu.submenu((
+		(url('pontoon.locale', locale.code), 'Projects', ''),
+		(url('pontoon.locale.contributors', locale.code), 'Contributors', 'contributors')
+	), request.path.split('/')[2]|default('')) }}
```
mathjazz: Menu (div class="submenu") shall be moved above the form tag, to keep it consistent with contributors page. This will also fix some spacing issues.

pontoon/base/templates/project.html
```diff
@@ -24,7 +25,14 @@
...
+    (url('pontoon.project.contributors', project.slug), 'Contributors', 'contributors')
+  ), request.path.split('/')[3]|default('')) }}
```
mathjazz: Menu (div class="submenu") shall be moved above the form tag, to keep it consistent with contributors page. This will also fix some spacing issues.

pontoon/base/templates/widgets/contributors.html
```diff
@@ -0,0 +1,63 @@
+{% import "widgets/menu.html" as Menu %}
+
+{# Widget to select period for filtering of contributors #}
+{% macro select_period(period, viewurl) %}
```
mathjazz: Nit: no "of", filtering contributors should ne enough.

pontoon/base/templates/widgets/contributors.html
```diff
@@ -0,0 +1,63 @@
...
+      </tr>
+    {% endfor %}
+    </tbody>
+    <tfoot>
+      <tr>
+        <td colspan="3"><sup id="fn-translations">Translation counts do not include unchanged translations.<a href="#ref-translations" title="Jump back to top.">↩</a></sup></td>
+      </tr>
+    </tfoot>
+  </table>
+{% endmacro %}
```
mathjazz: Nit: Add newline.

pontoon/base/templates/widgets/menu.html
```diff
@@ -0,0 +1,12 @@
...
+<div class="{{ class_ }}">
+  <ul class="links">
+    {% for choiceurl, label, val in choices %}
+    <li {% if val == value %}class="active"{% endif %}>
+      <a href="{{ choiceurl }}">{{ label }}</a>
+    </li>
+    {% endfor %}
+  </ul>
+</div>
+{% endmacro %}
```
mathjazz: Nit: Add newline.

pontoon/base/templates/widgets/menu.html
```diff
@@ -0,0 +1,12 @@
+{# defines menu widget #}
+{% macro submenu(choices=(), value=None, class_='submenu') %}
```
mathjazz: menuclass would be more appropriate variable name than than class_

pontoon/base/views.py
```diff
@@ -20,6 +21,8 @@
 from django.contrib.auth import authenticate, login
 from django.contrib.auth.decorators import login_required
 from django.contrib.auth.models import User
+from django.db.models import Q
```
mathjazz: We already import from django.db.models, see 4 lines below.

pontoon/base/views.py
```diff
@@ -76,34 +81,81 @@ def home(request):
     return translate(request, locale, project.slug)
 
 
-def locale(request, locale, template='locale.html'):
-    """Locale view."""
-    log.debug("Locale view.")
+class TemplateModelView(TemplateView):
```
Osmose: This is essentially a simplified reimplementation of the built-in [DetailView](https://ccbv.co.uk/projects/Django/1.8/django.views.generic.detail/DetailView/). Was that intentional?

In any case, I think catching the exception this far up in the call stack isn't wise, as an `ObjectDoesNotExist` error is not always because the main object we're trying to fetch couldn't be found. It could be another type of error, maybe fetching a related object, and this would swallow that error up without us knowing. That's why `DetailView` only catchs the `ObjectDoesNotExist` error within `get_object` and not everywhere.

pontoon/base/models.py
```diff
@@ -49,7 +49,7 @@ def _changed_translations_count(self, *args):
             Case(
                 When(translation_query, then=1), output_field=models.IntegerField(), default=0))
 
-    def with_translation_counts(self, start_date=None, limit=100):
+    def with_translation_counts(self, start_date=None, query_args=None, limit=100):
```
Osmose: `query_filters` might be a better name, `args` doesn't tell me what the argument is for. Also, it could use some documentation in the docstring.

pontoon/base/views.py
```diff
@@ -76,34 +81,81 @@ def home(request):
...
+        try:
+            period = int(self.request.GET['period'])
+            if period <= 0:
+                raise ValueError
+            start_date = (timezone.now() + relativedelta(months=-period))
+        except (KeyError, ValueError):
+            period = None
+            start_date = None
+
+        context['contributors'] = User.translators.with_translation_counts(start_date, kwargs.get('query_args'))
```
Osmose: The preferred way of passing something like the `query_args` down to subclasses is to make an attribute or method on the class that parent classes can overload. So instead of having classes that inherit from `ContributorsView` override `get_context_data` and pass in the extra filters they want to apply, you'd have either a `translation_query_filters` attribute or `get_translation_query_filters` method for them to override. Then just call/reference that here and they can avoid overriding the unrelated `get_context_data` method.

pontoon/base/views.py
Osmose: I see now that you want send a message on 404. In that case, you'd want to sublcass `DetailView`, call it `PontoonDetailView` or something, and override `get_object` to catch 404s and send a message that subclasses can override.

pontoon/base/views.py
```diff
@@ -76,34 +81,81 @@ def home(request):
...
+        context['period'] = period
+        return context
+
+
+class LocaleView(TemplateModelView):
```
Osmose: I don't think replacing the existing locale view with a class-based one is necessary here, nor is it within the scope of this bug. I'd rather we leave the old views alone and just add new views.

pontoon/base/views.py
```diff
@@ -116,31 +168,51 @@ def locales(request):
     return render(request, 'locales.html', data)
 
 
-def project(request, slug, template='project.html'):
-    """Project view."""
-    log.debug("Project view.")
+class ProjectView(TemplateModelView):
+    """
+    Renders a page for the project.
```
Osmose: This docstring doesn't tell us anything more than the classname did, no need to include it.

Also my comments about not changing the locale view apply here as well.

pontoon/base/views.py
```diff
@@ -76,34 +81,81 @@ def home(request):
...
 
-    projects = Project.objects.filter(
-        disabled=False, pk__in=Resource.objects.values('project'), locales=l) \
-        .order_by("name")
 
-    if not projects:
-        messages.error(
-            request, "Oops, no projects available for this locale.")
-        request.session['translate_error'] = {
+class ContributorsView(TemplateView):
```
Osmose: I like this class, it's clever that it's both a standalone page and a base class for other pages.

pontoon/base/tests/__init__.py
```diff
@@ -40,6 +40,20 @@ def client_login(self, user=None):
 
         return user
 
+    def assert_redirect_with_error(self, url, message, after_redirect_url='http://testserver/'):
+        """
+        Some of the views are handling errors by returning redirect to the main page
+        and displaying message to the user
+        """
+        response = self.client.get(url)
```
Osmose: I think it's better to pass the response into the assertion instead of passing a URL and getting the response here, so that we can test all kinds of client requests instead of just plain GETs with no arguments.

pontoon/base/tests/__init__.py
```diff
@@ -40,6 +40,20 @@ def client_login(self, user=None):
...
+
+        assert_code(response, 302)
+        assert_equal(response['Location'], after_redirect_url)
+
+        response = self.client.get('/')
```
Osmose: Asserting things about Django messages is a nightmare, but making a request to the index each time we want to isn't a great solution.

And if we're passing in the response instead, it becomes much harder to mock messages out for testing. Hmm.

I almost think maybe we should just use the built-in `TestCase.assertRedirects` for checking the redirect, and have the test classes mock out the messages stuff and test that instead of trying to make a new assertion method for this.

Or, if we moved this to a more specific subclass for the view tests, it might make sense to pass the URL. I'm not sure, but leaning towards the former. Thoughts?

jotes:
@mathjazz @Osmose could you review again? I've committed some fixes.

mathjazz:
r+wc from me. HTML has to be changed before we deploy this, which I'll address in a separate PR after this is merged, as we already agreed.

jotes:
@mathjazz done :)

mathjazz:
Thanks! I'm fine with merging this once @Osmose has another look at the things he reviewed.

pontoon/base/views.py
```diff
@@ -120,16 +114,8 @@ def project(request, slug, template='project.html'):
...
-    try:
-        p = Project.objects.get(slug=slug, disabled=False,
-                                pk__in=Resource.objects.values('project'))
-    except Project.DoesNotExist:
-        messages.error(request, "Oops, project could not be found.")
-        request.session['translate_error'] = {
-            'none': None,
-        }
-        return HttpResponseRedirect(reverse('pontoon.home'))
+    p = get_object_or_404(Project, slug=slug, disabled=False,
```
Osmose: PRAISE BE, YOU DO A CHARITY UNTO US

pontoon/base/views.py
```diff
@@ -341,24 +327,68 @@ def contributor(request, email, template='user.html'):
     return render(request, template, data)
 
 
-def contributors(request, template='users.html'):
-    """Top contributors view."""
-    log.debug("Top contributors view.")
+class ContributorsMixin(object):
+    def contributors_filter(self, **kwargs):
+        """
+        Method returrns filters for given contributors page. Method is dedicated to use
```
Osmose: Mispelled "returns". No need to mention about overriding, either. "Return Q() filters for fetching contributors. Fetches all by default." would be enough.

pontoon/base/tests/test_views.py
```diff
@@ -186,5 +189,96 @@ def test_given_period(self):
...
+        """
+        ResourceFactory.create(project=Project.objects.get(pk=1))
+
+
+class ProjectTests(ViewTestCase):
+    def test_project_doesnt_exist(self):
+        """
+        Checks if view is returning error when project slug is invalid.
+        """
+        assert_code(self.client.get('/projects/project_doesnt_exist/'), 404)
```
Osmose: I am diggin' this `assert_code` thing.

pontoon/base/tests/test_views.py
```diff
@@ -186,5 +189,96 @@ def test_given_period(self):
...
+class ProjectContributorsTests(ViewTestCase):
+    def test_project_doesnt_exist(self):
+        """
+        Checks if view handles invalid project.
+        """
+        assert_code(self.client.get('/projects/project_doesnt_exist/contributors'), 404)
+
+    def test_project_contributors(self):
+        """
+        Tests if view returns contributors spe-cific for given project.
```
Osmose: "specific", not "spe-cific".

pontoon/base/tests/test_views.py
```diff
@@ -186,5 +189,96 @@ def test_given_period(self):
...
+        """
+        project = Project.objects.get(pk=1)
+
+        with nested(patch.object(views.ProjectContributorsView, 'render_to_response', return_value=HttpResponse('')),
```
Osmose: We run on Python 2.7, so `nested` is unnecessary. You can just use a line continuation to line things up:

```python
with patch('blah.blah'), \
     patch('foo.bar'):
    pass
```

pontoon/base/tests/test_views.py
```diff
@@ -186,5 +189,96 @@ def test_given_period(self):
...
+            patch('pontoon.base.models.UserTranslationsManager.with_translation_counts')) as (mock_render, mock_contributors_counts):
+
+            response = self.client.get('/projects/pontoon-intro/contributors')
+
+            assert_equal(mock_render.call_args[0][0]['project'], project)
+            assert_equal(dict(mock_contributors_counts.call_args[0][1].children),
```
Osmose: If I'm reading this right, are you asserting against an internal property of the `Q()` object passed to `with_translation_counts`? That's not a good idea. That makes this test fragile; we could reasonably change the code to perform the same thing as it does now but slightly differently and break this test.

If you wanted to do a really full end-to-end-ish test, you could set up some test data and compare against the models that are returned, but I'd also be fine with comparing against a `Q()` object here. So like `assert_equal(mock.call_args[0][1], Q(expected_filter=True))`.

pontoon/base/tests/test_views.py
```diff
@@ -186,5 +189,96 @@ def test_given_period(self):
...
+        """
+        Checks if view handles invalid project.
+        """
+        assert_code(self.client.get('/projects/project_doesnt_exist/contributors'), 404)
+
+    def test_project_contributors(self):
+        """
+        Tests if view returns contributors spe-cific for given project.
+        """
+        project = Project.objects.get(pk=1)
```
Osmose: If possible, I prefer to use newly-created models when at all possible instead of relying on the locales and projects already in the database. It makes the tests easier to reason about since we don't have to remember what data is already in the database, and frees us up to remove or modify the pre-existing data later on.

Osmose:
r+wc, this looks great. Nice work, thank you! :D

pontoon/base/tests/test_views.py
jotes: I totally agree, however i had problems with equality of Q() instances, that's why i decided to compare their children property. I would chooose to create some data and just filter it if you agree.

Osmose: Go for it!

I'm actually surprised Django doesn't have an `__eq__` implementation for `Q()` objects. That's what I get for making an assumption.

jotes:
@Osmose hi, could you look again? I've rebased and fixed tests.

pontoon/base/tests/test_views.py
```diff
@@ -186,5 +188,113 @@ def test_given_period(self):
...
+
+        with patch.object(views.ProjectContributorsView, 'render_to_response', return_value=HttpResponse('')) as mock_render:
+
+            self.client.get('/projects/{}/contributors'.format(first_project.slug))
+            assert_equal(mock_render.call_args[0][0]['project'], first_project)
+            assert_equal(list(mock_render.call_args[0][0]['contributors']), [first_project_contributor])
+
+            self.client.get('/projects/{}/contributors'.format(second_project.slug))
+            assert_equal(mock_render.call_args[0][0]['project'], second_project)
+            assert_equal(list(mock_render.call_args[0][0]['contributors']), [second_project_contributor])\
```
Osmose: Remove the `\` and the end, here.

Osmose:
This looks ready to merge except for that `\`. Excellent work!

jotes:
@Osmose done.

jotes:
@Osmose @mathjazz i've rebase this pr against latest cleanups on master and stashed everything to single commit (it's always easier to revert feature if somethings isn't okay).

PR #202, opened 2015-10-04
pep8 and pyflakes cleanups

jotes:
Hi, i've ran pyflakes and pep8 to make some cleanups. Most of them are fixed, however there are 3 relatively important errors:
```
$ pylama
pontoon/administration/files.py:806:1: E0602 undefined name 'original' [pyflakes]
pontoon/administration/views.py:299:1: E0602 undefined name '_save_entity' [pyflakes]
pontoon/administration/views.py:303:1: E0602 undefined name '_save_translation' [pyflakes]
```
Also, i've added syntax checking to travis-ci to validate every PR against pep8/pyflakes compatibility. I think it will improve readability and reliability of code.
@mathjazz @Osmose could you tell me if that's good enough to merge or if you agree with general idea? Thx in advance.

### DIFF
--- a/pontoon/base/tests/formats/test_silme.py
+++ b/pontoon/base/tests/formats/test_silme.py
@@ -214,7 +214,6 @@ class PropertiesTests(FormatTestsMixin, TestCase):
         self.run_parse_basic(BASE_PROPERTIES_FILE, 0)
 
     def test_parse_multiple_comments(self):
-        #import ipdb; ipdb.set_trace()
         self.run_parse_multiple_comments(BASE_PROPERTIES_FILE, 1)
 
     def test_parse_no_comments_no_sources(self):
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ install:
 before_script:
   - psql -c 'create database pontoon;' -U postgres
 script:
+  - pylama
   - python manage.py test
 addons:
   postgresql: "9.4"
--- a/bin/dump-verbatim.py
+++ b/bin/dump-verbatim.py
@@ -7,7 +7,6 @@ from optparse import make_option
 from pootle_store.models import Store, Unit
 
 import codecs
-import io
 import json
 import subprocess
 
--- a/pontoon/administration/files.py
+++ b/pontoon/administration/files.py
@@ -1,5 +1,4 @@
 import codecs
-import configparser
 import datetime
 import fnmatch
 import logging
@@ -91,9 +90,6 @@ def get_locale_directory(project, locale):
                 'path': os.path.join(root, dirname),
             }
 
-    # Projects not using locale directories (format=file)
-    formats = Resource.objects.filter(project=project).values_list(
-        'format', flat=True).distinct()
     if project.repository_type == 'file':
         return {
             'name': '',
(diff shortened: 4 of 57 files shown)


pontoon/base/tests/formats/test_silme.py
```diff
@@ -214,7 +214,6 @@ def test_parse_basic(self):
         self.run_parse_basic(BASE_PROPERTIES_FILE, 0)
 
     def test_parse_multiple_comments(self):
-        #import ipdb; ipdb.set_trace()
```
Osmose: Whoops

Osmose:
Thanks for the PR! I'll actually take a look later, but two comments:

- There's a lot of pep8/flake8 rules that I think we should ignore, like line length or pretty much every indentation-related rule. I suspect that we'll have to figure out what we want to ignore. Lately I've even been of the opinion that style checking isn't useful at all, and that having differing styles in a codebase is fine. The error-checking bits and stuff like useless imports and variables are still highly useful.
- I'm fine with having rules and encouraging people to use them, but I'm against making Travis fail on them. Code style as a blocker to merging is more tedious than useful, I think.

jotes:
@Osmose I agree in some points, however:
* in my experience, if you don't lint project during PR phase, there's no place for this later. I've worked in projects with multiple different styles of coding and comfort of working on this code decreased exponentially because of conflicts between people about their own guidelines and policies. However, i'm only sharing my experiences - i just want to make codebase a little bit better.
* some of the nits can be easily discovered by linters so reviewers don't have to skim to through the code for untrimmed lines etc.
* I would suggest to enable checks in travis only for critical things like: unresolved variable/class/function name, unused variables. I think it's important to make distinction  between visual style (which can be optional, i fully agree) and things critical to code execution.

Osmose:
Along with not being so hot on style checking, I generally try to shy away from comments in PRs about code style unless they're egregious (like docstrings that are super-long lacking readability). I'm not expecting PR reviewers to do style checking instead of tools.

What if we make Travis only run pyflakes? That'd catch actual errors vs style problems. If you do that and remove the pep8-based fixes I'd be quite alright with this change.

@mathjazz What are your thoughts?

jotes:
Sure, there's no problem to use pyflakes only. I don't want to push
anything what would make our work less comfortable.

On Wed, Oct 7, 2015 at 12:23 AM, Michael Kelly <notifications@github.com>
wrote:

> Along with not being so hot on style checking, I generally try to shy away
> from comments in PRs about code style unless they're egregious (like
> docstrings that are super-long lacking readability). I'm not expecting PR
> reviewers to do style checking instead of tools.
>
> What if we make Travis only run pyflakes? That'd catch actual errors vs
> style problems. If you do that and remove the pep8-based fixes I'd be quite
> alright with this change.
>
> @mathjazz <https://github.com/mathjazz> What are your thoughts?
>
> —
> <https://github.com/mozilla/pontoon/pull/202#issuecomment-146019281>.
>



-- 
jarek@reijutsu:~$ fortune
jarek@reijutsu:~$ fortune
You now have Asian Flu.

mathjazz:
I only see benefits in using a shared code style.

We should agree on what exceptions/modifications of PEP8 shall we use and have checks set up in our editors. Before we do this, making Travis cry for each of the inconsistencies will get annoying.

But I agree having pyflakes as part of Travis is a good start.

Nice work @jotes!

PR #203, opened 2015-10-05
Fix server error after locale added, but not imported

mathjazz:
After locale was added to the project and before it got imported, dashboards
with locale selector (namely project dashboard) failed with server error,
becuse we had no chart data an attempted to divide with None.

@Osmose r?


Osmose:
r+ :+1:

PR #204, opened 2015-10-05
Parallelized Sync via Celery

Osmose:
This is two big changes:

- Move all sync-related code to a new `sync` app (except `vcs.py`, at least until we kill the old sync for good).
- Set up celery and run per-project syncing in parallel. This involved a major refactoring of the `sync_projects` management command into a few different files.

A few other things are included as well:

- Removed the `deleted` field on Translations and switched back to deleting translations directly. We previously added this so we could detect when an entity was changed by a translation being deleted, but we don't need this functionality now that we track when entities are changed by creating `ChangedEntityLocale` objects.

  I removed it now because keeping it in meant that we had to coordinate the sync tasks so that deleted translations were removed once all the syncing was done. Doing _that_ required us to set up a backend for storing task results, and also resulted in a bunch of weirdness around the execution of tasks. It was easier to just remove it completely.

@mathjazz I've tested this on staging and as far as I can tell it works fine, but this could use some careful attention while testing since it's such a major change to syncing.

r?


Osmose:
Oh yeah, this is currently deployed on staging. Logs of sync runs show up in Papertrail, and I've documented how to inspect the celery queues via Flower. You can also check out the new RabbitMQ add-on to get access to a RabbitMQ web admin for inspecting the state of the queues directly.

Another thing to note is that timing how long sync takes as a whole is a bit harder now due to there being no task run once all the other sync tasks finish. In the future I hope to address this with the structured logging we keep talking about for sync, but for now Flower can tell us how long an individual project's sync took to execute. Also, finding the time via logs is still possible, just tedious.

pontoon/sync/management/commands/sync_projects.py
```diff
@@ -0,0 +1,49 @@
...
+    def handle(self, *args, **options):
+        """
+        Collect the projects we want to sync and trigger worker jobs to
+        sync each one.
+        """
+        projects = Project.objects.filter(disabled=False)
+        if args:
+            projects = projects.filter(slug__in=args)
+
+        if len(projects) < 1:
```
jotes: projects.exists() should be a little faster.

pontoon/sync/management/commands/sync_projects.py
```diff
@@ -0,0 +1,49 @@
...
+            raise CommandError('No matching projects found.')
+
+        for project in projects:
+            if not project.can_commit:
+                self.stdout.write(u'Skipping project {0}, cannot commit to repository.'
+                                  .format(project.name))
+            else:
```
jotes: I would add here stdout with something like "Syncing project {}".format(project.slug) -> it will be valuable for the later investigation what happened.

pontoon/sync/management/commands/sync_projects.py
jotes: Second thing: maybe it rather should be projects.count() == len(args)? Because if someone calls this command with mispell, there's no other way to notice it.

pontoon/settings/base.py
```diff
@@ -58,6 +58,13 @@ def path(*args):
 # tests.
 SSLIFY_DISABLE = DEBUG or os.environ.get('CI', False)
 
+# URL to the RabbitMQ server
+BROKER_URL = os.environ.get('RABBITMQ_URL', None)
```
jotes: os.environ.get('RABBITMQ_URL') is the same.

pontoon/settings/base.py
```diff
@@ -58,6 +58,13 @@ def path(*args):
...
+
+# Execute celery tasks locally instead of in a worker unless the
+# environment is configured.
+CELERY_ALWAYS_EAGER = os.environ.get('CELERY_ALWAYS_EAGER', 'False') != 'False'
```
jotes: bool(os.environ.get('CELERY_ALWAYS_EAGER'))

pontoon/sync/core.py
```diff
@@ -0,0 +1,227 @@
...
+
+    # Collect all entities across VCS and the database and get their
+    # keys so we can match up matching entities.
+    vcs_entities = get_vcs_entities(vcs_project)
+    db_entities = get_db_entities(db_project)
+    entity_keys = set().union(db_entities.keys(), vcs_entities.keys())
+
+    changeset = ChangeSet(db_project, vcs_project, now)
+    for key in entity_keys:
+        db_entity = db_entities.get(key, None)
```
jotes: None is default value for get()

pontoon/sync/management/commands/sync_projects.py
Osmose: > projects.exists() should be a little faster.

It would be, except immediately after this we're iterating over the queryset. Using exists would result in two queries: one for checking existence, and another for actually fetching the projects. It's a micro-optimization in either case, but this one reduces it to a single query in both cases.



> Second thing: maybe it rather should be projects.count() == len(args)? Because if someone calls this command with mispell, there's no other way to notice it.

Maybe not that in particular, but the greater problem of not warning when a slug is invalid is a great point, and I've filed [bug 1211789](https://bugzilla.mozilla.org/show_bug.cgi?id=1211789) for it.

pontoon/settings/base.py
Osmose: Yeah, but I like being explicit about it.

Except all those other places I wasn't explicit about it. Let's pretend those don't exist because I'm always consistent.

pontoon/settings/base.py
Osmose: `bool('False') == True`, and environment variables are always strings, so we can't use numbers or primitives.

Long term this is still weird, though. A reorganization of the settings using django-configurations has been floating around my head as a someday task for a while.

pontoon/sync/management/commands/sync_projects.py
jotes: @Osmose 
You're right about len(). It was too early in morning for me to see this more clearly :)
Secondthing:
Yeah, meant mispells in parameters.

pontoon/base/apps.py
```diff
@@ -16,6 +16,9 @@ def ready(self):
         super(BaseConfig, self).ready()
         self.monkeypatch()
 
+        # Load celery app so shared_task uses it for executing tasks.
```
mathjazz: Who is Mr. shared_task?

mathjazz:
The code makes sense and looks great and is beautifully documented and thank you a ton for moving sync into a separate app!

I'm not sure removing a delete field is actually necessary, because we should never delete translations (or entities) anyways:
https://bugzilla.mozilla.org/show_bug.cgi?id=1192872

But we might get to that later.

Testing it on staging now.

mathjazz:
r+. I was unable to break it. :-)

pontoon/base/apps.py
Osmose: http://celery.readthedocs.org/en/latest/django/first-steps-with-django.html#using-the-shared-task-decorator

Osmose:
Updated with the change from @jotes's review.

pontoon/base/apps.py
mathjazz: Yeah, celery.shared_task would help a lot.

I was afraid shared_task might confuse a reader, especially if she's not in the process of reviewing this particuar PR.

PR #205, opened 2015-10-07
Fix bug 1211271: Mark projects as changed on locale add.

Osmose:
By marking projects as changed when we update locales, we force sync to run even if there haven't been any new translations or changes in VCS.

This depends on #204 because I didn't want to handle the rebasing process after it gets merged. Feel free to comment just on the last commit, or wait until #204 lands. Probably the latter.

@mathjazz r?


mathjazz:
Very nice! r+

(Merge after #204).

PR #207, opened 2015-10-08
Bug 1203788: Replace administration app with Django admin.

Osmose:
This is a WIP PR that tears out the old administration app and replaces it with a new app that implements a slightly customized Django admin interface. It replaces the `/a/` with `/admin/` and updates the Project listing and editing page to support all the options in the old admin interface, albeit with a slightly worse look (for now). There's also a button for triggering sync for a specific project again, although we don't yet have a way to view the sync logs outside of papertrail yet (I think we should add that in [bug 1208171](https://bugzilla.mozilla.org/show_bug.cgi?id=1208171) after this lands).

Also, this relies on #204 for the same reason everything I'm working on relies on it. :P

@mathjazz I still need to write tests for this but I want feedback on how it feels before I start. Is this good enough to ship?


pontoon/admin/urls.py
```diff
@@ -0,0 +1,16 @@
+from django.conf.urls import include, patterns, url
+
+from pontoon.admin import views
+from pontoon.admin.site import admin_site
+
+
+urlpatterns = patterns(
+    '',
```
jotes: we could use 'pontoon.admin.views' here to avoid repetition in urls.

Osmose: Wouldn't that affect the admin URLs too, though?

I actually wonder if I shouldn't include the custom views in the `admin_site` object as well, so that they're consistent with the `admin:` namespace and naming style. Hmmm....

mathjazz:
This is definitely a step in the right direction.

* I think some model classes are missing (e.g. Repository).
* Looks like I cannot disable a project (from showing in the UI and syncing).
* I'm not sure we need the latest activity column (old admin was sharing template with /projects).
* I wonder how hard would it be to make adding Resources to Subpages possible by searching (the popup doesn't have a search box).

mathjazz:
When deleting a project, we should also clean up project files in the filesystem.

adngdb:
What's the status here? I see there are many conflicts, and as far as I know Osmose isn't working on Pontoon anymore. Is anyone taking over? Do we want to drop this for now?

mathjazz:
We made a good progress towards closing bug 1203788, however separate of this PR. There's still some work left to do and there might be pieces of code in here we can use.

I'll close this PR and reference it in the bug with a short description of what needs to be done.

PR #209, opened 2015-10-13
Enabled support for pyflakes in pontoon.

jotes:
@Osmose @mathjazz I've enabled pyflakes and cleaned up code a little. Pylama allows us to use "# noqa" and has a lot more configuration possibilities than raw pyflakes.


Osmose:
This is too cool, I'm r+ :D

mathjazz:
\o/

PR #211, opened 2015-10-14
Add reverse FK names and optimize locale list.

Osmose:
Only fetch data for locales active in the project, and avoid
selecting all the project locales for each locale in the project.

Testing with django-debug-toolbar on a project with 3 locales, the project page with the locale listing went from 235 database queries to 18 and loads significantly faster.

This also has some query fixes I found while I was looking that don't particularly improve our performance, but improve our comprehensibility.

@mathjazz r?


mathjazz:
Good move!

It breaks adding/requesting new locales though.

Osmose:
Updated, adding locales works now.

pontoon/base/templates/locale_selector.html
```diff
@@ -31,7 +31,7 @@
       <li class="clearfix{% if locale and locale == l %} current{% endif %}{% if l and l.chart %} filter{% endif %}">
         <span class="language {{ l.code|lower }}">{% if l and l.chart %}<a href="{% if project %}{{ url('pontoon.translate', l.code, project.slug) }}{% else %}{{ url('pontoon.locale', l.code) }}{% endif %}" class="clearfix">{% endif %}{{ l.name }}{% if l and l.chart %}</a>{% endif %}</span>
         <span class="code">{% if l and l.chart %}<a class="code" href="{% if project %}{{ url('pontoon.translate', l.code, project.slug) }}{% else %}{{ url('pontoon.locale', l.code) }}{% endif %}" class="clearfix">{% endif %}{{ l.code }}{% if l and l.chart %}</a>{% endif %}</span>
-        {% if not locale %}
+        {% if not locale and l.chart %}
```
Osmose: I've become the thing I hated the most. We should not be using the existence of `l.chart` as a flag that the locale is not active on the project.

But these pages are in my sights as the first ones to be rewritten, so WHO CARES

mathjazz:
r+

PR #212, opened 2015-10-15
Bug 1206779: Fix locale/project contributors design

mathjazz:
This is branched off #201, so please only look at this commit:
https://github.com/mozilla/pontoon/commit/1d28fa6c8bf23523fbc9d4683256ee41df5e2e3f

@Osmose @jotes r?

I'll reuse more tabs code when implementing tabs for locale requests. And redesigning it again. It could look much better, but it's good enough for release.

### DIFF
--- a/pontoon/base/static/css/project.css
+++ b/pontoon/base/static/css/project.css
@@ -1,21 +1,5 @@
-.about {
-  padding: 0 12px;
-  text-align: left;
-}
-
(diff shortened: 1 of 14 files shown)


jotes:
As we talked: r+wc -> there's only one small thing to fix (bug created by me) https://gist.github.com/jotes/8585835852a223290277

PR #214, opened 2015-10-19
No need to run migrations when depoying (anymore)

mathjazz:
Not sure about the first deploy though.

@Osmose r?


Osmose:
r+, although I think you're right in that they don't have to be run manually ever.

PR #216, opened 2015-10-19
Speed up locale list

Osmose:
ae324e04bb27038bf6c5b4e841dec782791f53aa got pulled because it made the translate view slow. I've added an extra commit that removed a suspicious query from the prefetching for `locales_parts_stats` that was doing in `IN` query on a set of hundreds of ids. Locally, my translate view now loads roughly as fast as it does on the current master with this tweak, whereas without the extra commit it runs twice as slow.

The original "Add reverse FK names" commit should be identical to the one previously merged (different hash since it's rebased but it's the same changes otherwise).

@mathjazz r?


mathjazz:
Thank you for a separate commit! r+

PR #217, opened 2015-10-19
Bug 1214392: Denormalize latest activity.

Osmose:
This relies on #216, will rebase after that gets merged. Pay attention to the last commit only.

This denormalizes the latest activity such that Project, Resource, Locale, and ProjectLocale (which was previously an implicit through model for Project.locales) all have a `latest_translation` field that gets updated when translations are saved. The listings that show the latest activity now efficiently pull the `latest_translation` field when trying to figure out which translation qualifies as the latest activity, as opposed to doing a bunch of queries to find the latest activity.

Locally this led to a major drop in load time for the project, locale, and resource listings for me. It also reduced the number of database queries on each page by quite a bit.

@mathjazz r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -19,6 +19,7 @@ from jsonfield import JSONField
 
 from pontoon.administration.vcs import commit_to_vcs, get_revision, update_from_vcs
 from pontoon.base import utils
+from pontoon.base.utils import get_object_or_none
 
 
 log = logging.getLogger('pontoon')
@@ -152,6 +153,15 @@ class Locale(models.Model):
 
     team_description = models.TextField(blank=True)
 
+    #: Most recent translation approved or created for this locale.
+    latest_translation = models.ForeignKey(
+        'Translation',
+        blank=True,
+        null=True,
+        related_name='+',
+        on_delete=models.SET_NULL
+    )
+
     def cldr_plurals_list(self):
         if self.cldr_plurals == '':
             return [1]
@@ -183,13 +193,15 @@ class Locale(models.Model):
     class Meta:
         ordering = ['name', 'code']
 
-    def latest_activity(self, project=None):
-        """Get latest activity for locale and project if provided."""
-        translations = Translation.objects.filter(locale=self)
-        if project:
-            translations = translations.filter(entity__resource__project=project)
-
-        return translations.latest_activity()
+    def get_latest_activity(self, project=None):
+        """Get latest activity for project and locale if provided."""
+        if project is None:
+            if self.latest_translation is not None:
+                return self.latest_translation.latest_activity
+            else:
+                return None
+        else:
+            return ProjectLocale.get_latest_activity(project, self)
 
 
 class ProjectQuerySet(models.QuerySet):
@@ -204,7 +216,7 @@ class ProjectQuerySet(models.QuerySet):
 class Project(models.Model):
     name = models.CharField(max_length=128, unique=True)
     slug = models.SlugField(unique=True)
-    locales = models.ManyToManyField(Locale)
+    locales = models.ManyToManyField(Locale, through='ProjectLocale')
 
     transifex_project = models.CharField(
         "Project", max_length=128, blank=True)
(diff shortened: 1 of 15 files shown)


mathjazz:
Thank you for a separate commit! r+

Osmose:
Rebased, easier to review now. :P

pontoon/base/views.py
```diff
@@ -138,19 +142,25 @@ def locale_project(request, locale, slug):
...
+    projects = Project.objects.prefetch_related('subpage_set').distinct()
+    project = get_object_or_404(projects, disabled=False, slug=slug, resources__isnull=False)
+
+    # Amend the parts dict with latest activity info.
+    stats_qs = (
+        Stats.objects
+        .filter(resource__project=project, locale=l)
+        .select_related('resource', 'latest_translation')
     )
+    stats = {s.resource.path: s for s in stats_qs}
```
mathjazz: Nice reduction of DB queries!

pontoon/base/views.py
```diff
@@ -138,19 +142,25 @@ def locale_project(request, locale, slug):
...
     l = get_object_or_404(Locale, code__iexact=locale)
 
-    project = (
-        get_object_or_404(
-            Project.objects.prefetch_related('subpage_set'),
-            disabled=False,
-            slug=slug,
-            pk__in=Resource.objects.values('project')
-        )
+    projects = Project.objects.prefetch_related('subpage_set').distinct()
```
mathjazz: Is that distinct() really necessary?

pontoon/base/models.py
```diff
@@ -152,6 +153,15 @@ class Locale(models.Model):
 
     team_description = models.TextField(blank=True)
 
+    #: Most recent translation approved or created for this locale.
+    latest_translation = models.ForeignKey(
+        'Translation',
+        blank=True,
+        null=True,
+        related_name='+',
+        on_delete=models.SET_NULL
```
mathjazz: Woohooo!

mathjazz: That thing up there means I've never seen on_delete before. But I'm also wondering, what would actually change if latest_translation got deleted in a cascade with the foreign key? Seems like you have that case covered in get_latest_activity().

pontoon/base/models.py
```diff
@@ -228,6 +240,15 @@ class Project(models.Model):
     # Whether this project has changed since the last sync.
     has_changed = models.BooleanField(default=False)
 
+    # Most recent translation approved or created for this project.
+    latest_translation = models.ForeignKey(
+        'Translation',
+        blank=True,
+        null=True,
+        related_name='+',
```
mathjazz: Why do we need no backwards relation? I understand we'll never use it, I'm just wondering why do we want to explicitly prevent it? Optimization?

mathjazz:
An error occurs when adding a new locale to a project, but it's easy to fix:
`Cannot set values on a ManyToManyField which specifies an intermediary model.  Use base.ProjectLocale's Manager instead.`

Excellent work, this is a huge improvement, can't wait to deploy this! r+wc

Tested with a few days old production database, following are the average of the 3 page load times before vs after applying the PR:
/projects: 4.5s vs 3.5s
/teams: 4.7s vs 3.5s
/sl: 3.4s vs 1.9s
/projects/pontoon-intro: 4.2s vs 3.3s
/sl/firefox-aurora 6.5s vs 4.1s

pontoon/base/views.py
Osmose: Yep, although because I split this into two lines (and I'm not entirely sure if I had a good reason to do so now, but didn't remove it) it's not obvious. The `get_object_or_404` call filters by `resources__isnull`, and anytime you filter by a ManyToMany, you'll get a result for each of the related objects. In this case, we get as many projects as there are resources in the project. Hence, the `distinct` to avoid the duplication.

pontoon/base/models.py
Osmose: It's the other way around: Deleting a translation deletes the locale with the foreign key. This is how I mysteriously deleted projects on my local instance while running a sync, which deleted some translations. 

Suffice it to say that we do _not_ want locales being deleted when translations are deleted. :P

pontoon/base/models.py
Osmose: Nah, it's a naming conflict. By default the `related_name` on the Translation for the Locale that it is the `latest_translation` of would be `translation.locale`, but we already have an attribute with that name on the Translation.

And then I copy-pasted it around since, as you said, we'll never use it, but it's only truly necessary on the Locale class.

Osmose:
Rebased and fixed the locale issue. I'm not worried about tests since the admin is going to be killed soon anyway.

We may want to hold off on merging this if we still intend to revert the other two commits this relies on, though.

mathjazz:
r+

Not merging until we figure out what to do with reverted commits this relies on.

PR #218, opened 2015-10-20
Bug 1215562: robots.txt file

EmiXLabs:
Tested locally, using sslserver (https://github.com/teddziuba/django-sslserver) module in local.py, works.

### DIFF
--- a/pontoon/urls.py
+++ b/pontoon/urls.py
@@ -31,6 +31,8 @@ urlpatterns = patterns('',
     url(r'^403/$', TemplateView.as_view(template_name='403.html')),
     url(r'^404/$', TemplateView.as_view(template_name='404.html')),
     url(r'^500/$', TemplateView.as_view(template_name='500.html')),
+    # Robots.txt
+    url(r'^robots.txt$', TemplateView.as_view(template_name='robots.txt', content_type='text/plain')),
 
     # Favicon
     url(r'^favicon\.ico$',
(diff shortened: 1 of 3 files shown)


Osmose:
Thanks for the PR!

pontoon/urls.py
```diff
@@ -40,6 +41,9 @@
     url(r'^pontoon\.js$',
         RedirectView.as_view(url='/static/js/pontoon.js', permanent=True)),
 
+    # Robots.txt
+    (r'^robots.txt$', lambda r: HttpResponse("User-agent: *\nDisallow: /a/\nDisallow: /admin/", content_type="text/plain")),
```
Osmose: This totally works, but I'd rather have this in a template file for maintainability's sake. A simple view that just calls render should be all that's needed.

EmiXLabs: @Osmose Added in one line as it will not be used a lot, but yes, I can change it to template if needed.

EmiXLabs: BTW, @Osmose, which one will be better, rendering in views or using direct_to_template in urls? The only extra line in views will be logging I guess.

mathjazz:
Looks good @EmiXLabs! Mind rebasing into a single commit?

pontoon/urls.py
Osmose: You seem to have already found a fine answer to that question. :P
For the record, the in-view logging isn't something we want to keep doing, we just haven't cleaned it up yet. So no worries.

mathjazz:
@Osmose says r+ on IRC!

Congrats on your first PR merged, @EmiXLabs!

PR #219, opened 2015-10-20
Fix bug 952488 - Added locale managers

jotes:
@Osmose @mathjazz Hi,

I would say it's an alpha version of migrations. There's a lot to discuss because i wasn't sure about some things. Hopefully every change should be relatively easy.
My questions and doubts:
- i'm assigning groups to locale because it's a lot easier to assign users. I think it will benefit in future, however i don't want to over-engineer anything.
- i'm not sure if goods idea to have signals which create these permission groups associated with locale. I mean, i don't now how often locale are added, maybe it's a rare case so this isn't so important feat. I could add separate bootstrap function for locale groups. It would speedup tests and object creation.
- I've added some tests for models, i'm not sure if views require them as they use has_perm functions.

### DIFF
--- a/pontoon/base/errors.py
+++ b/pontoon/base/errors.py
@@ -1,0 +1,9 @@
+from raygun4py import raygunprovider
+
+from django.conf import settings
+
+def send_exception(exc):
+    """
+    Function sends exception to selected provider.
+    """
+    raygunprovider.RaygunSender(settings.RAYGUN4PY_API_KEY).send_exception(exc)
--- a/pontoon/base/migrations/0041_create_locale_permissions.py
+++ b/pontoon/base/migrations/0041_create_locale_permissions.py
@@ -1,0 +1,41 @@
+# -*- coding: utf-8 -*-
+"""
+Django's (version 1.8.4) executes post_migration hooked in django.contrib.auth after every migration.
+This particular signal creates content_types and adds object's permissions to database.
+However, next migrations will require ContentTypes and some of permissions that's
+why we have to create them before post_migration is called. Important fact is there's currently no way
+to remove ContentType and Permissions created in this migration because they will be recreated by Django.
+That's why downgrade is a noop function.
+"""
+
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def create_locale_permissions(apps, schema_editor):
+    Permission = apps.get_model('auth', 'Permission')
+    ContentType = apps.get_model('contenttypes', 'ContentType')
+
+    locale_content_type, _ = ContentType.objects.get_or_create(app_label='base', model='locale')
+
+    Permission.objects.get_or_create(codename='can_translate_locale', content_type=locale_content_type,
+        name="Can add translations")
+    Permission.objects.get_or_create(codename='can_manage_locale', content_type=locale_content_type,
+        name="Can manage locale")
+
+
+def noop(apps, schema_editor):
+    pass
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('base', '0040_auto_20151017_0005'),
+        ('auth', '0006_require_contenttypes_0002'),
+    ]
+
+    operations = [
+        migrations.RunPython(create_locale_permissions, noop)
+    ]
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -6,11 +6,12 @@ import urllib
 from urlparse import urlparse
 
 from django.conf import settings
-from django.contrib.auth.models import User
+from django.contrib.auth.models import User, Group
 from django.core.exceptions import ValidationError
 from django.db import models
 from django.db.models import Sum, Prefetch, F, Q, Case, When
 from django.db.models.signals import post_save
+from django.dispatch import receiver
 from django.templatetags.static import static
 from django.utils import timezone
 
@@ -94,6 +95,7 @@ def user_display_name(self):
     name = self.first_name or self.email.split('@')[0]
     return u'{name} <{email}>'.format(name=name, email=self.email)
 
+
 User.add_to_class('display_name', user_display_name)
 User.add_to_class('translators', UserTranslationsManager())
 
@@ -138,6 +140,13 @@ class Locale(models.Model):
     code = models.CharField(max_length=20, unique=True)
     name = models.CharField(max_length=128)
     plural_rule = models.CharField(max_length=128, blank=True)
+
+    # Locale contains references to user groups who translate or manage them.
+    # Groups also store respective permissions for users.
+    translators_group = models.ForeignKey(Group, related_name='translated_locales', null=True,
+        on_delete=models.SET_NULL)
+    managers_group = models.ForeignKey(Group, related_name='managed_locales', null=True,
+        on_delete=models.SET_NULL)
 
     # CLDR Plurals
     CLDR_PLURALS = (
@@ -193,6 +202,10 @@ class Locale(models.Model):
 
     class Meta:
         ordering = ['name', 'code']
+        permissions = (
+            ('can_translate_locale', 'Can add translations'),
+            ('can_manage_locale', 'Can manage locale')
+        )
 
     def get_latest_activity(self, project=None):
         """Get latest activity for project and locale if provided."""
@@ -212,6 +225,7 @@ class ProjectQuerySet(models.QuerySet):
         resource defined.
         """
         return self.filter(disabled=False, resource__isnull=False)
+
 
 
 class Project(models.Model):
@@ -255,6 +269,9 @@ class Project(models.Model):
     class Meta:
         permissions = (
             ("can_manage", "Can manage projects"),
+
+            # @TODO: This permission should be removed after we'll migrate all users to new
+            # object-based permission system.
             ("can_localize", "Can localize projects"),
         )
 
@@ -918,12 +935,10 @@ class Stats(models.Model):
         return str(int(round(percent)))
 
 
+@receiver(post_save, sender=User)
 def create_user_profile(sender, instance, created, **kwargs):
     if created:
         UserProfile.objects.create(user=instance)
-
-# For every newly created user
-post_save.connect(create_user_profile, sender=User)
 
 
 def get_chart_data(stats):
--- a/pontoon/base/signals.py
+++ b/pontoon/base/signals.py
@@ -1,7 +1,13 @@
-from django.db.models.signals import post_save
+from guardian.models import GroupObjectPermission
+
+from django.core.exceptions import ObjectDoesNotExist
+from django.contrib.auth.models import Group, Permission
+from django.contrib.contenttypes.models import ContentType
+from django.db.models.signals import post_save, pre_save
 from django.dispatch import receiver
 
-from pontoon.base.models import ProjectLocale
+from pontoon.base import errors
+from pontoon.base.models import ProjectLocale, Locale
 
 
 @receiver(post_save, sender=ProjectLocale)
@@ -17,3 +23,78 @@ def project_locale_added(sender, **kwargs):
         project = project_locale.project
         project.has_changed = True
         project.save(update_fields=['has_changed'])
+
+
+@receiver(pre_save, sender=Locale)
+def create_locale_translators_group(sender, **kwargs):
+    """
+    Creates respective translators group in django's permission system.
+    Managers should be able to translate and manage locale.
+    """
+    instance = kwargs['instance']
+
+    if kwargs['raw'] or instance.translators_group is not None:
+        return
+    try:
+        locale_ct = ContentType.objects.get(app_label='base', model='locale')
+        locale_group, _ = Group.objects.get_or_create(name='{} translators'.format(instance.code))
+        can_translate = Permission.objects.get(content_type=locale_ct, codename='can_translate_locale')
+        locale_group.permissions.add(can_translate)
+        instance.translators_group = locale_group
+    except ObjectDoesNotExist as e:
+        errors.send_exception(e)
+
+
+@receiver(pre_save, sender=Locale)
+def create_locale_managers_group(sender, **kwargs):
+    """
+    Creates respective managers group in django's permission system.
+    Managers should be able to translate and manage locale.
+    """
+    instance = kwargs['instance']
+
+    if kwargs['raw'] or instance.managers_group is not None:
+        return
+
+    try:
+        locale_ct = ContentType.objects.get(app_label='base', model='locale')
+        locale_group, _ = Group.objects.get_or_create(name='{} translators'.format(instance.code))
+        can_translate = Permission.objects.get(content_type=locale_ct, codename='can_translate_locale')
+        can_manage = Permission.objects.get(content_type=locale_ct, codename='can_manage_locale')
+        locale_group.permissions.add(can_translate)
+        locale_group.permissions.add(can_manage)
+        instance.managers_group = locale_group
+    except ObjectDoesNotExist as e:
+        errors.send_exception(e)
+
+@receiver(post_save, sender=Locale)
+def assign_group_permissions(sender, **kwargs):
+    """"
+    After creation of locale, we have to assign translation and management permissions
+    to groups of translators and managers assigned to locale.
+    """
+    if kwargs['raw'] or not kwargs['created']:
+        return
+
+    instance = kwargs['instance']
+    try:
+        locale_ct = ContentType.objects.get(app_label='base', model='locale')
+        can_translate = Permission.objects.get(content_type=locale_ct, codename='can_translate_locale')
+        can_manage = Permission.objects.get(content_type=locale_ct, codename='can_manage_locale')
+
+        GroupObjectPermission.objects.get_or_create(object_pk=instance.pk,
+            content_type=locale_ct,
+            group=instance.translators_group,
+            permission=can_translate)
+
+        GroupObjectPermission.objects.get_or_create(object_pk=instance.pk,
+            content_type=locale_ct,
+            group=instance.managers_group,
+            permission=can_translate)
+
+        GroupObjectPermission.objects.get_or_create(object_pk=instance.pk,
+            content_type=locale_ct,
+            group=instance.managers_group,
+            permission=can_manage)
+    except ObjectDoesNotExist as e:
+        errors.send_exception(e)
--- a/pontoon/base/templates/user.html
+++ b/pontoon/base/templates/user.html
@@ -37,7 +37,7 @@
 {% block middle %}
 <ul class="info">
   <li><span class="fa fa-fw fa-envelope"></span><a href="mailto:{{ contributor.email }}">{{ contributor.email }}</a></li>
-  <li><span class="fa fa-fw fa-lock"></span>Can {% if contributor.has_perm('base.can_localize') %}submit and approve translations{% else %}make suggestions{% endif %}</li>
+  <li><span class="fa fa-fw fa-lock"></span>Can {% if 'base.can_translate_locale' in contributor.get_all_permissions() %}submit and approve translations{% else %}make suggestions{% endif %}</li>
   {% if user.is_authenticated() and user.email == contributor.email %}
   <li class="check-box quality-checks{% if user.profile.quality_checks %} enabled{% endif %}"><span class="fa fa-fw"></span>Quality checks</li>
   {% endif %}
--- a/pontoon/base/tests/test_models.py
+++ b/pontoon/base/tests/test_models.py
@@ -12,7 +12,12 @@ from django.test.utils import override_settings
 
 from mock import call, Mock, patch
 
-from pontoon.base.models import Entity, ProjectLocale, Repository, User
+from pontoon.base.models import (
+    Entity,
+    ProjectLocale,
+    Repository,
+    User
+)
 from pontoon.base.tests import (
     assert_attributes_equal,
     ChangedEntityLocaleFactory,
@@ -825,6 +830,61 @@ class LocaleTests(TestCase):
             mock_get_latest_activity.return_value = 'latest'
             assert_equal(locale.get_latest_activity(project=project), 'latest')
             mock_get_latest_activity.assert_called_with(project, locale)
+
+    def test_translators_group(self):
+        """
+        Tests if user has permission to translate locales after assigment.
+        """
+        user = UserFactory.create()
+        [first_locale, second_locale] = LocaleFactory.create_batch(2)
+
+        assert_equal(user.has_perm('base.can_translate_locale'), False)
+        assert_equal(user.has_perm('base.can_translate_locale', first_locale), False)
+        assert_equal(user.has_perm('base.can_translate_locale', second_locale), False)
+
+        user.groups.add(second_locale.translators_group)
+
+        assert_equal(user.has_perm('base.can_translate_locale'), False)
+        assert_equal(user.has_perm('base.can_translate_locale', first_locale), False)
+        assert_equal(user.has_perm('base.can_translate_locale', second_locale), True)
+
+        user.groups.add(first_locale.translators_group)
+
+        assert_equal(user.has_perm('base.can_translate_locale'), False)
+        assert_equal(user.has_perm('base.can_translate_locale', first_locale), True)
+        assert_equal(user.has_perm('base.can_translate_locale', second_locale), True)
+
+    def test_managers_group(self):
+        """
+        Tests if user has permission to manage and translate locales after assigment.
+        """
+        user = UserFactory.create()
+        [first_locale, second_locale] = LocaleFactory.create_batch(2)
+
+        assert_equal(user.has_perm('base.can_translate_locale'), False)
+        assert_equal(user.has_perm('base.can_translate_locale', first_locale), False)
+        assert_equal(user.has_perm('base.can_translate_locale', second_locale), False)
+        assert_equal(user.has_perm('base.can_manage_locale'), False)
+        assert_equal(user.has_perm('base.can_manage_locale', first_locale), False)
+        assert_equal(user.has_perm('base.can_manage_locale', second_locale), False)
+
+        user.groups.add(second_locale.managers_group)
+
+        assert_equal(user.has_perm('base.can_translate_locale'), False)
+        assert_equal(user.has_perm('base.can_translate_locale', first_locale), False)
+        assert_equal(user.has_perm('base.can_translate_locale', second_locale), True)
+        assert_equal(user.has_perm('base.can_manage_locale'), False)
+        assert_equal(user.has_perm('base.can_manage_locale', first_locale), False)
+        assert_equal(user.has_perm('base.can_manage_locale', second_locale), True)
+
+        user.groups.add(first_locale.managers_group)
+
+        assert_equal(user.has_perm('base.can_translate_locale'), False)
+        assert_equal(user.has_perm('base.can_translate_locale', first_locale), True)
+        assert_equal(user.has_perm('base.can_translate_locale', second_locale), True)
+        assert_equal(user.has_perm('base.can_manage_locale'), False)
+        assert_equal(user.has_perm('base.can_manage_locale', first_locale), True)
+        assert_equal(user.has_perm('base.can_manage_locale', second_locale), True)
 
 
 class ProjectLocaleTests(TestCase):
--- a/pontoon/base/tests/test_views.py
+++ b/pontoon/base/tests/test_views.py
@@ -158,7 +158,7 @@ class ContributorsTests(TestCase):
 
     def test_default_period(self):
         """
-        Calling the top_contributors should result in period being None.
+        Calling the top contributors should result in period being None.
         """
         self.client.get('/contributors/')
         assert_true(self.mock_render.call_args[0][0]['period'] is None)
@@ -224,9 +224,9 @@ class ProjectContributorsTests(ViewTestCase):
         """
         assert_code(self.client.get('/projects/project_doesnt_exist/contributors'), 404)
 
-    def test_project_contributors(self):
+    def test_project_top_contributors(self):
         """
-        Tests if view returns contributors specific for given project.
+        Tests if view returns top contributors specific for given project.
         """
         first_project = ProjectFactory.create()
         ResourceFactory.create(project=first_project)
@@ -275,9 +275,9 @@ class LocaleContributorsTests(ViewTestCase):
         """
         assert_code(self.client.get('/missing-locale/contributors'), 404)
 
-    def test_locale_contributors(self):
+    def test_locale_top_contributors(self):
         """
-        Tests if view returns contributors specific for given locale.
+        Tests if view returns top contributors specific for given locale.
         """
         first_locale = LocaleFactory.create()
         first_locale_contributor = TranslationFactory.create(locale=first_locale,
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -30,7 +30,6 @@ from django.utils import timezone
 from django.utils.datastructures import MultiValueDictKeyError
 from django.views.generic import TemplateView
 from django.views.generic.detail import DetailView
-from django_browserid.views import Verify as BrowserIDVerifyBase
 from operator import itemgetter
 from pontoon.administration import files
 from pontoon.base import utils
@@ -607,7 +606,7 @@ def delete_translation(request, template=None):
         return HttpResponse("error")
 
     # Non-privileged users can only delete own non-approved translations
-    if not request.user.has_perm('base.can_localize'):
+    if not request.user.has_perm('base.can_translate_locale', translation.locale):
         if translation.user == request.user:
             if translation.approved is True:
                 log.error(
@@ -627,7 +626,7 @@ def delete_translation(request, template=None):
     next = get_translation(
         entity=entity, locale=locale, plural_form=plural_form)
 
-    if next.pk is not None and request.user.has_perm('base.can_localize'):
+    if next.pk is not None and request.user.has_perm('base.can_translate_locale', next.locale):
         next.approved = True
         next.approved_user = request.user
         next.approved_date = timezone.now()
@@ -700,7 +699,7 @@ def update_translation(request, template=None):
         ignore = True
 
     now = timezone.now()
-    can_localize = request.user.has_perm('base.can_localize')
+    can_translate = request.user.has_perm('base.can_translate_locale', l)
     translations = Translation.objects.filter(
         entity=e, locale=l, plural_form=plural_form)
 
@@ -716,7 +715,7 @@ def update_translation(request, template=None):
             t = translations.get(string=string)
 
             # If added by privileged user, approve and unfuzzy it
-            if can_localize:
+            if can_translate:
 
                 # Unless there's nothing to be changed
                 if t.user is not None and t.approved and t.approved_user \
@@ -780,7 +779,7 @@ def update_translation(request, template=None):
             if warnings:
                 return warnings
 
-            if can_localize:
+            if can_translate:
                 unapprove(translations)
 
             unfuzzy(translations)
@@ -788,9 +787,9 @@ def update_translation(request, template=None):
             t = Translation(
                 entity=e, locale=l, user=user, string=string,
                 plural_form=plural_form, date=now,
-                approved=can_localize)
+                approved=can_translate)
 
-            if can_localize:
+            if can_translate:
                 t.approved_user = user
                 t.approved_date = now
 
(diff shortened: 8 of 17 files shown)


pontoon/settings/base.py
```diff
@@ -179,8 +180,11 @@ def path(*args):
 AUTHENTICATION_BACKENDS = [
     'django_browserid.auth.BrowserIDBackend',
     'django.contrib.auth.backends.ModelBackend',
+    'guardian.backends.ObjectPermissionBackend',
 ]
 
+ANONYMOUS_USER_ID = -1
```
mathjazz: You never use this.

pontoon/base/tests/test_views.py
```diff
@@ -221,9 +221,9 @@ def test_project_doesnt_exist(self):
         """
         assert_code(self.client.get('/projects/project_doesnt_exist/contributors'), 404)
 
-    def test_project_contributors(self):
+    def test_project_top_contributors(self):
         """
-        Tests if view returns contributors specific for given project.
+        Tests if view returns top_contributors specific for given project.
```
mathjazz: Nit: replace underscore with space - "top contributors".

pontoon/base/tests/test_models.py
```diff
@@ -783,3 +788,60 @@ def test_for_project_locale_order(self):
...
+
+        user.groups.add(first_locale.translators_group)
+
+        assert_equal(user.has_perm('base.can_translate'), False)
+        assert_equal(user.has_perm('base.can_translate', first_locale), True)
+        assert_equal(user.has_perm('base.can_translate', second_locale), True)
+
+    def test_mamagers_group(self):
+        """
+        Tests if user has permission to managea and translate locales after assigment.
```
mathjazz: Nit: managea - manage.

pontoon/base/models.py
```diff
@@ -841,13 +908,11 @@ def __unicode__(self):
         return str(int(round(percent)))
 
 
+@receiver(post_save, sender=User)
```
mathjazz: Oh, look! Nice!

pontoon/base/migrations/0039_locale_managers.py
```diff
@@ -0,0 +1,63 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+TRANSLATION_THRESHOLD = 10
```
mathjazz: You don't use this anywhere and I also don't think we should.

pontoon/settings/base.py
jotes: It was recommended in django-guardian's docs. However i can remove this :)

pontoon/base/views.py
```diff
@@ -1313,10 +1313,6 @@ def request_locale(request):
 
 class BrowserIDVerify(BrowserIDVerifyBase):
     def login_success(self):
-        # Check for permission to localize if not granted on every login
-        if not self.user.has_perm('base.can_localize'):
-            utils.add_can_localize(self.user)
-
```
mathjazz: Godspeed! Please also remove add_can_localize() from utils.py.

pontoon/base/views.py
```diff
@@ -674,7 +674,7 @@ def update_translation(request, template=None):
         ignore = True
 
     now = timezone.now()
-    can_localize = request.user.has_perm('base.can_localize')
+    can_localize = request.user.has_perm('base.can_translate', l)
```
mathjazz: Please rename can_localize variable name to can_translate for consistency.

pontoon/base/models.py
```diff
@@ -192,6 +204,60 @@ def latest_activity(self, project=None):
...
+
+    GroupObjectPermission.objects.get_or_create(object_pk=instance.pk,
+        content_type=locale_ct,
+        group=instance.translators_group,
+        permission=can_translate)
+
+    GroupObjectPermission.objects.get_or_create(object_pk=instance.pk,
+            content_type=locale_ct,
+            group=instance.managers_group,
+            permission=can_translate)
```
mathjazz: Less indent.

pontoon/base/models.py
```diff
@@ -192,6 +204,60 @@ def latest_activity(self, project=None):
...
+
+    GroupObjectPermission.objects.get_or_create(object_pk=instance.pk,
+            content_type=locale_ct,
+            group=instance.managers_group,
+            permission=can_manage)
```
mathjazz: Less indent.

pontoon/base/models.py
```diff
@@ -192,6 +204,60 @@ def latest_activity(self, project=None):
...
+
+    locale_ct = ContentType.objects.get(app_label='base', model='locale')
+    locale_group, _ = Group.objects.get_or_create(name='{} translators'.format(instance.code.upper()))
+    can_translate = Permission.objects.get(content_type=locale_ct, codename='can_translate')
+    locale_group.permissions.add(can_translate)
+    instance.translators_group = locale_group
+
+
+@receiver(pre_save, sender=Locale)
+def create_locale_managers_group(sender, instance, raw, **kwargs):
```
mathjazz: It would be nice to have a comment here explaining what the function does, similarly as you did with create_locale_translators_group().

pontoon/base/models.py
```diff
@@ -192,6 +204,60 @@ def latest_activity(self, project=None):
...
+    locale_group, _ = Group.objects.get_or_create(name='{} translators'.format(instance.code.upper()))
+    can_translate = Permission.objects.get(content_type=locale_ct, codename='can_translate')
+    can_manage = Permission.objects.get(content_type=locale_ct, codename='can_manage')
+    locale_group.permissions.add(can_translate)
+    locale_group.permissions.add(can_manage)
+    instance.managers_group = locale_group
+
+
+@receiver(post_save, sender=Locale)
+def assign_group_permissions(sender, instance, created, raw, **kwargs):
```
mathjazz: Comment would be nice here too.

pontoon/base/views.py
mathjazz: And MOZILLIANS_API_KEY from settings.py.

pontoon/base/migrations/0040_locale_translators.py
```diff
@@ -0,0 +1,51 @@
...
+    Bug 952488 - We're assigning translation permissions to active contributors.
+    """   
+    User = apps.get_model('auth', 'User')
+
+    # We're using through migration because it's easier to map
+    # user/group primary keys without requiring to query full objects.
+    UserGroup = User.groups.through
+
+    contributors_locale = (
+        User.objects.exclude(email=settings.EXCLUDE)
```
mathjazz: We use settings.EXCLUDE for excluding us from contributor lists, no need for it here.

pontoon/base/migrations/0040_locale_translators.py
```diff
@@ -0,0 +1,51 @@
...
+            .filter(translation__approved=True)
+            .values('pk', 'translation__locale')
+            .annotate(approved_translations=models.Count('translation__locale', distinct=True))
+            .filter(approved_translations__gte=ACTIVITY_THRESHOLD)
```
mathjazz: No need for the trashold.

pontoon/base/migrations/0040_locale_translators.py
```diff
@@ -0,0 +1,51 @@
...
+    """   
+    User = apps.get_model('auth', 'User')
+
+    # We're using through migration because it's easier to map
+    # user/group primary keys without requiring to query full objects.
+    UserGroup = User.groups.through
+
+    contributors_locale = (
+        User.objects.exclude(email=settings.EXCLUDE)
+            .filter(translation__approved=True)
```
mathjazz: Add .distinct()

pontoon/settings/base.py
mathjazz: Oh, in this case I'd add a comment instead of removing it.

pontoon/base/migrations/0040_locale_translators.py
mathjazz: Instead, you should filter users, that have only submitted translations to one locale. We'll take care of the others manually.

mathjazz:
I like where this is going. The comments I've made are mostly trivial. Nice work!

That being said, I was unable to run migrations locally, due to the following error:
```
django.db.utils.IntegrityError: duplicate key value violates unique constraint 
"django_content_type_app_label_45f3b1d93ec8c61c_uniq"
DETAIL:  Key (app_label, model)=(base, locale) already exists.
```

I'll look at this again once you address the issues I pointed out.

jotes:
@mathjazz It looks like this issue wasn't ocurring when database was empty.
However, i've fixed this.
I've applied changes you've requested.

On Tue, Oct 20, 2015 at 6:20 PM, Matjaž Horvat <notifications@github.com>
wrote:

> I like where this is going. The comments I've made are mostly trivial.
> Nice work!
>
> That being said, I was unable to run migrations locally, due to the
> following error:
>
> django.db.utils.IntegrityError: duplicate key value violates unique constraint
> "django_content_type_app_label_45f3b1d93ec8c61c_uniq"
> DETAIL:  Key (app_label, model)=(base, locale) already exists.
>
> I'll look at this again once you address the issues I pointed out.
>
> —
> <https://github.com/mozilla/pontoon/pull/219#issuecomment-149620997>.
>



-- 
jarek@reijutsu:~$ fortune
jarek@reijutsu:~$ fortune
You now have Asian Flu.

mathjazz:
Looks great! In 0040_locale_translators.py, you should fix that contributors_locale query (no exclude or transhold, filter users that have only submitted translations to one locale).

pontoon/base/views.py
Osmose: If you're removing this you might as well get rid of the whole class and use the default verifier.

pontoon/base/models.py
```diff
@@ -192,6 +204,69 @@ def latest_activity(self, project=None):
         return translations.latest_activity()
 
 
+@receiver(pre_save, sender=Locale)
+def create_locale_translators_group(sender, instance,  raw, **kwargs):
```
Osmose: `signals.py` is a better place to put model signals.

Osmose: [According to the docs](https://docs.djangoproject.com/en/1.8/topics/signals/#receiver-functions), signals should only accept `sender` and `kwargs` in case signal arguments change over time.

pontoon/base/models.py
```diff
@@ -192,6 +204,69 @@ def latest_activity(self, project=None):
...
+    """
+    Creates respective translators group in django's permission system.
+    Managers should be able to translate and manage locale.
+    """
+    if raw:
```
Osmose: Why not abort if the locale isn't being created as well? I don't think we'd want to do this creation check on _every_ save.

pontoon/base/migrations/0038_auto_20151017_0005.py
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+
+from django.db import migrations
+
+
+def create_locale_permissions(apps, schema_editor):
```
Osmose: Wow, Permissions aren't automatically migrated? That's lame.

pontoon/base/migrations/0040_locale_translators.py
```diff
@@ -0,0 +1,47 @@
...
+    contributors_locale = (
+        User.objects.filter(translation__approved=True)
+            .values('pk', 'translation__locale')
+            .annotate(approved_translations=models.Count('translation__locale', distinct=True))
+            .filter(approved_translations=1)
+    )
+
+    for contributor in contributors_locale:
+        UserGroup.objects.create(user_id=contributor['pk'],
+            group_id=contributor['translation__locale'].group.pk)
```
Osmose: Does this really work? The `values` call should have `['translation__locale']` as an integer with the pk on the locale rather than the locale instance itself.

pontoon/base/migrations/0039_locale_managers.py
```diff
@@ -0,0 +1,59 @@
...
+
+    dependencies = [
+        ('base', '0038_auto_20151017_0005'),
+        ('guardian', '__first__'),
+    ]
+
+    operations = [
+        migrations.AddField("Locale", "translators_group", models.ForeignKey(to="auth.Group", null=True)),
+        migrations.AddField("Locale", "managers_group", models.ForeignKey(to="auth.Group", null=True)),
+        migrations.RunPython(create_locale_groups),
```
Osmose: Could you please make this reversible by just deleting all the groups?

pontoon/base/tests/test_models.py
```diff
@@ -783,3 +788,60 @@ def test_for_project_locale_order(self):
...
+        assert_equal(user.has_perm('base.can_translate', first_locale), False)
+        assert_equal(user.has_perm('base.can_translate', second_locale), True)
+
+        user.groups.add(first_locale.translators_group)
+
+        assert_equal(user.has_perm('base.can_translate'), False)
+        assert_equal(user.has_perm('base.can_translate', first_locale), True)
+        assert_equal(user.has_perm('base.can_translate', second_locale), True)
+
+    def test_mamagers_group(self):
```
Osmose: managers

pontoon/base/models.py
```diff
@@ -137,6 +140,11 @@ class Locale(models.Model):
     name = models.CharField(max_length=128)
     plural_rule = models.CharField(max_length=128, blank=True)
 
+    # Locale contains referennces to user group who translate or manage them.
```
Osmose: References. User groups. Also replace the `/` at the end with a period.

Osmose:
This look pretty good, and smaller than I expected! Get pylama happy and fix the mentioned issues and I'll take one last look before approving and merging. 

RE: The tests, I think you're fine not adding tests for those views for now.

Osmose:
Also, since #217 got merged, you'll need to rebase and rename those migrations. Sorry!

pontoon/base/migrations/0040_locale_translators.py
jotes: Yeah, sorry - i didn't had time to fix this yesterday but i realized how wrong this query is. Definitely fix this.

jotes:
@Osmose don't worry about rebases, i'm used to it ;-). Anyway i would humbly ask you to look at my changes again.

Osmose:
I get the following while attempting to rollback 0041:

```
  File "./manage.py", line 20, in <module>
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/core/management/__init__.py", line 338, in execute_from_command_line
    utility.execute()
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/core/management/__init__.py", line 330, in execute
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/core/management/base.py", line 393, in run_from_argv
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/core/management/base.py", line 444, in execute
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/core/management/commands/migrate.py", line 222, in handle
    executor.migrate(targets, plan, fake=fake, fake_initial=fake_initial)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/migrations/executor.py", line 112, in migrate
    self.unapply_migration(states[migration], migration, fake=fake)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/migrations/executor.py", line 168, in unapply_migration
    state = migration.unapply(state, schema_editor)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/migrations/migration.py", line 162, in unapply
    operation.database_backwards(self.app_label, schema_editor, from_state, to_state)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/migrations/operations/fields.py", line 70, in database_backwards
    schema_editor.remove_field(from_model, from_model._meta.get_field(self.name))
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/backends/base/schema.py", line 434, in remove_field
    self.execute(self._delete_constraint_sql(self.sql_delete_fk, model, fk_name))
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/backends/base/schema.py", line 111, in execute
    cursor.execute(sql, params)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/backends/utils.py", line 79, in execute
    return super(CursorDebugWrapper, self).execute(sql, params)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/utils.py", line 97, in __exit__
    six.reraise(dj_exc_type, dj_exc_value, traceback)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
django.db.utils.OperationalError: cannot ALTER TABLE "base_locale" because it has pending trigger events
```

Are you able to rollback successfully?

Osmose:
I also get this when migrating forwards again through 0040 after commenting out the python part of 0041 and leaving the `AddField` and `AlterField` parts:

```
  File "./manage.py", line 20, in <module>
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/core/management/__init__.py", line 338, in execute_from_command_line
    utility.execute()
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/core/management/__init__.py", line 330, in execute
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/core/management/base.py", line 393, in run_from_argv
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/core/management/base.py", line 444, in execute
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/core/management/commands/migrate.py", line 222, in handle
    executor.migrate(targets, plan, fake=fake, fake_initial=fake_initial)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/migrations/executor.py", line 110, in migrate
    self.apply_migration(states[migration], migration, fake=fake, fake_initial=fake_initial)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/migrations/executor.py", line 148, in apply_migration
    state = migration.apply(state, schema_editor)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/migrations/migration.py", line 115, in apply
    operation.database_forwards(self.app_label, schema_editor, old_state, project_state)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/migrations/operations/special.py", line 183, in database_forwards
    self.code(from_state.apps, schema_editor)
  File "/Users/mkelly/Projects/pontoon/pontoon/base/migrations/0040_auto_20151017_0005.py", line 15, in create_locale_permissions
    Permission.objects.create(codename='can_translate', content_type=locale_content_type)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/models/manager.py", line 127, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/models/query.py", line 348, in create
    obj.save(force_insert=True, using=self.db)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/models/base.py", line 734, in save
    force_update=force_update, update_fields=update_fields)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/models/base.py", line 762, in save_base
    updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/models/base.py", line 846, in _save_table
    result = self._do_insert(cls._base_manager, using, fields, update_pk, raw)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/models/base.py", line 885, in _do_insert
    using=using, raw=raw)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/models/manager.py", line 127, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/models/query.py", line 920, in _insert
    return query.get_compiler(using=using).execute_sql(return_id)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/models/sql/compiler.py", line 974, in execute_sql
    cursor.execute(sql, params)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/backends/utils.py", line 79, in execute
    return super(CursorDebugWrapper, self).execute(sql, params)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/utils.py", line 97, in __exit__
    six.reraise(dj_exc_type, dj_exc_value, traceback)
  File "/Users/mkelly/.virtualenvs/pontoon/lib/python2.7/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
django.db.utils.IntegrityError: duplicate key value violates unique constraint "auth_permission_content_type_id_codename_key"
DETAIL:  Key (content_type_id, codename)=(24, can_translate) already exists.
```

I was able to get around this by faking down to 0039, deleting the translate and manage permissions, and deleting all the user groups and retrying.

Osmose:
Is it intentional that people with the translator permission can get their own translations approved but not others? My expectation was that they should be able to approve any translation but that may be wrong.

Osmose:
The can_manage permission as it's being used here doesn't do anything with locales, it simply gives permission to edit all the projects, even if it's associated with a locale. The intent was that locale managers (or owners, or whatever) are granted the ability to grant the can_translate permission to anyone for their specific locale, and nothing more.

Perhaps instead of reusing the existing can_manage permission, we need a new permission completely? Since the only way I know of to add permissions is via the admin, this may get a bit tricky with admin customization. I get the feeling this needs more thought.

Sorry for not catching these on my first pass, I was only checking for code issues when I should've given the functionality a closer look too. :(

jotes:
@Osmose Hi, thx for thorough review. I've restructured migrations, from now on downgrades are possible (with a little twist in 0040 but a docstring explains why). I've added _locale suffix to "can migrate" and "can translate" permissions.
I'm not sure if @mathjazz should look at this again because there are some changes in code of views.

> Is it intentional that people with the translator permission can get their own translations approved but not others? My expectation was that they should be able to approve any translation but that may be wrong.

My thoughts: translator is specialized in one locale, so there's no point to give him possiblity to approve strings in locale which he doesn't know. Anyway - if you think it should be done the other way, i'll gladly reimplement this.

Osmose:
> My thoughts: translator is specialized in one locale, so there's no point to give him possiblity to approve strings in locale which he doesn't know. Anyway - if you think it should be done the other way, i'll gladly reimplement this.

What I meant is that, as a translator, I cannot approve translations submitted by other users _even in my own locale_.

So if you are not a pt-BR translator and I am, and you submit a translation for a string in pt-BR, I cannot approve it because I don't see the approval buttons in the UI. All I can do is submit my own translation that will automatically get approved. 

My expectation is that, as a translator for pt-BR, I should see approve/delete buttons for all translations in pt-BR.

mathjazz:
I agree with @Osmose - users with *can_translate* permission for the locale should be able to approve all translations for that locale, not just their own translations. It's a simple fix, you'll have to replace the old *can_localize* permission check in:
```
pontoon/base/templates/translate.html
```

Please also update the user page and the verbatim import script to new permissions:
```
pontoon/base/templates/user.html
pontoon/administration/management/commands/import_verbatim.py
```

I'll work on the UI for users with can_manage_locale permission after this gets r+. It will be a simple page for adding/removing can_translate permission to users and also editing locale description. It will not be a part of the admin page.

jotes:
@mathjazz @Osmose ok guys, no problem!

jotes:
@mathjazz Done. Should i remove can_localize completely?

mathjazz:
Good job @jotes! Please check the translation history piece.

I think we'll be safe to remove can_localize completely once we migrate all users, including the ones we need to take care of manually. You can add a TODO comment in models.py, which holds the only remaining reference of can_localize.

jotes:
@mathjazz i've added docstring and removed permission check from history view.

pontoon/base/migrations/0041_create_locale_permissions.py
```diff
@@ -0,0 +1,39 @@
+# -*- coding: utf-8 -*-
+"""
+Django's (version 1.8.4) executes post_migration hooked in django.contrib.auth after every migration.
+This particular signal creates content_types and adds object's permissions to database.
+However, next migrations will require ContentTypes and some of permissions that's
+why we have to create them before post_migration is called. Important fact is there's currently no way
+to remove ContentType and Permissions created in this migration because they will be recreated by Django.
+That's why downgrade is a noop function.
```
Osmose: This is excellent sleuthing and an excellent docstring.

pontoon/base/models.py
```diff
@@ -94,7 +95,14 @@ def user_display_name(self):
     name = self.first_name or self.email.split('@')[0]
     return u'{name} <{email}>'.format(name=name, email=self.email)
 
+
+@property
+def can_approve_translations(self):
+    return self.groups.filter(permissions__codename='can_translate_locale').exists()
```
Osmose: Is it possible to use `has_perm` here? I wonder if it returns true if you have the given permission for a single project?

pontoon/base/models.py
```diff
@@ -139,6 +147,11 @@ class Locale(models.Model):
     name = models.CharField(max_length=128)
     plural_rule = models.CharField(max_length=128, blank=True)
 
+    # Locale contains references to user groups who translate or manage them.
+    # Groups also store respective permissions for users.
+    translators_group = models.ForeignKey(Group, related_name='translated_locales')
```
Osmose: You might want to set `on_delete=models.SET_NULL` for these two attributes. Otherwise, if those groups are ever deleted, the entire locale will be deleted as well.

pontoon/base/models.py
Osmose: I believe this is called "leaving the place cleaner than how you found it"

pontoon/base/signals.py
```diff
@@ -17,3 +21,71 @@ def project_locale_added(sender, **kwargs):
...
+
+@receiver(pre_save, sender=Locale)
+def create_locale_translators_group(sender, **kwargs):
+    """
+    Creates respective translators group in django's permission system.
+    Managers should be able to translate and manage locale.
+    """
+    instance = kwargs['instance']
+
+    if kwargs['raw'] and getattr(instance, 'translators_group', False):
```
Osmose: Why `getattr`? The moment this code starts existing so will the model attribute. You can just check if it's `None`, right?

pontoon/base/signals.py
```diff
@@ -17,3 +21,71 @@ def project_locale_added(sender, **kwargs):
...
+        return
+
+    locale_ct = ContentType.objects.get(app_label='base', model='locale')
```
Osmose: You do a lot of `get`ing in these signals. If any of the groups accidentally get deleted from the database, we will lose our ability to save anything to database across the entire site.

It'd be nicer if we caught these errors and sent them to RayGun while continuing on. That way we'd be alerted to the error but the site functionality degrades in case of an issue instead of breaking completely.

pontoon/base/signals.py
Osmose: Also, don't you mean `or` instead of `and`?

pontoon/base/templates/user.html
```diff
@@ -37,7 +37,7 @@
 {% block middle %}
 <ul class="info">
   <li><span class="fa fa-fw fa-envelope"></span><a href="mailto:{{ contributor.email }}">{{ contributor.email }}</a></li>
-  <li><span class="fa fa-fw fa-lock"></span>Can {% if contributor.has_perm('base.can_localize') %}submit and approve translations{% else %}make suggestions{% endif %}</li>
+  <li><span class="fa fa-fw fa-lock"></span>Can {% if contributor.can_approve_translations %}submit and approve translations{% else %}make suggestions{% endif %}</li>
```
Osmose: In a future PR we should consider listing the locales that the user has permission for. But this is great for now.

pontoon/base/signals.py
jotes: yup, you're right.

pontoon/base/models.py
Osmose: Just saw the tests below, looks like we can't, huh? The assumption that we assign permissions via groups here is weird, though.

Maybe something like checking if the set returned by `get_objects_for_perm` is empty?

Osmose:
This is really nice! I tested locally and it totally works as expected now. Great job!

I'll leave it to mathjazz to decide when this is ready for merging. :D

pontoon/base/models.py
jotes: Sure, however this fields will have to be nullable.

jotes:
@Osmose i've rebased and implemented your notes. Could you look at this again? Thx :)

pontoon/base/errors.py
```diff
@@ -0,0 +1,10 @@
+from raygun4py import raygunprovider
+
+from django.conf import settings
+
+def send_exception():
+    """
+    Function sends exception to selected provider. This will allow us to change
+    error-service in future. Currently we support Raygun.io.
+    """
+    raygunprovider.RaygunSender(settings.RAYGUN4PY_API_KEY).send_exception()
```
Osmose: You should be passing the exception in as an argument to this function and passing it on to RayGun. While RayGun is smart enough to find the last exception, it's not good to rely on that, especially for complicated error cases.

pontoon/base/errors.py
```diff
@@ -0,0 +1,10 @@
+from raygun4py import raygunprovider
+
+from django.conf import settings
+
+def send_exception():
+    """
+    Function sends exception to selected provider. This will allow us to change
```
Osmose: Explaining that this allows us to abstract away which specific error service we're using isn't necessary, leave it out. Same with the mention of Raygun; the developer can already see that we're importing Raygun libs, they don't need to be told twice.

pontoon/base/signals.py
```diff
@@ -17,3 +22,78 @@ def project_locale_added(sender, **kwargs):
...
+
+    if kwargs['raw'] or instance.translators_group:
+        return
+    try:
+        locale_ct = ContentType.objects.get(app_label='base', model='locale')
+        locale_group, _ = Group.objects.get_or_create(name='{} translators'.format(instance.code.upper()))
+        can_translate = Permission.objects.get(content_type=locale_ct, codename='can_translate_locale')
+        locale_group.permissions.add(can_translate)
+        instance.translators_group = locale_group
+    except:
```
Osmose: Never use a try block without specifying the type of error you're catching. Otherwise, any and all errors will be caught regardless of if they're expected or not.

In this case you want to be catching [ObjectDoesNotExist](https://docs.djangoproject.com/en/1.8/ref/exceptions/#objectdoesnotexist) errors.

pontoon/base/signals.py
```diff
@@ -17,3 +22,78 @@ def project_locale_added(sender, **kwargs):
...
+
+@receiver(pre_save, sender=Locale)
+def create_locale_translators_group(sender, **kwargs):
+    """
+    Creates respective translators group in django's permission system.
+    Managers should be able to translate and manage locale.
+    """
+    instance = kwargs['instance']
+
+    if kwargs['raw'] or instance.translators_group:
```
Osmose: This is a nit, but checking for `is None` instead of relying on the fact that `None` is falsy is a good general practice in case you ever run into an object that is falsy for weird reasons.

Osmose:
Looks fiiiiiiiine. :D

pontoon/base/signals.py
jotes: Yeah, i know it's called "pokemon-style" catching of exceptions and in 99% of cases i wouldn't use it. I had version with ObjectDoesNotExist, but i had second thoughts. What if code inside changes in a way that cause different exception and site still will be broken? Everything will be logged into raygun so it won't be silenced in any way. Maybe i'm overthinking it :thought_balloon:

jotes:
@Osmose I've fixed nits :)

pontoon/base/signals.py
Osmose: The difference is that a database entry not existing is a recoverable error that could happen and that we know how to fix, and is also something that the user can go on without. In that situation, it's appropriate for us to get a notification and fix it manually.

But the entire space of errors that could happen include errors that could corrupt data or break some part of the site if execution continues. That's why, in the general case, we want most exceptions to propagate up all the way to the error middleware, get sent to us, and show the user an error page.

pontoon/base/signals.py
```diff
@@ -17,3 +25,78 @@ def project_locale_added(sender, **kwargs):
...
+    if kwargs['raw'] or instance.translators_group is not None:
+        return
+    try:
+        locale_ct = ContentType.objects.get(app_label='base', model='locale')
+        locale_group, _ = Group.objects.get_or_create(name='{} translators'.format(instance.code.upper()))
+        can_translate = Permission.objects.get(content_type=locale_ct, codename='can_translate_locale')
+        locale_group.permissions.add(can_translate)
+        instance.translators_group = locale_group
+    except ObjectDoesNotExist:
+        errors.send_exception(*sys.exc_info())
```
Osmose: Why not just pass the exception itself in?

```python
except ObjectDoesNotExist as err:
    errors.send_exception(err)
```

jotes: Exception don't contain information about traceback. It's an example from raygun4py, however we don't hook this function with global sys.excepthoook yet.

Osmose: Huh, I misread https://github.com/MindscapeHQ/raygun4py/blob/master/python2/raygun4py/raygunprovider.py#L74 and missed that it takes the traceback from exc_info when an exception is passed in. Either way, since they already call exc_info to handle it for you you don't need to call it yourself, but it's equivalent either way, so whatevs. :D

jotes: okay :D then i'll call it inside send_exception.

jotes:
@osmose send_exception takes only exception as you suggested.

mathjazz:
As per @Osmose on IRC, this is now r+. Excellent work @jotes!

I'll merge when my UI tweaks are ready (working on it).

pontoon/administration/management/commands/import_verbatim.py
```diff
@@ -0,0 +1,76 @@
...
+                        dateObj = timezone.make_aware(datetime.strptime(date, "%Y-%m-%d %H:%M:%S"))
+                        t.date = dateObj
+                        t.approved_date = dateObj
+
+                        t.save()
+
+                        user.groups.add(locale_translators_map[options['locale']])
+                    except User.DoesNotExist:
+                        # TODO: create missing user?
+                        self.stdout.write("Pontoon user not found: " + email)
```
mathjazz: This file was replaced with pontoon/sync/management/commands/import_pootle.py, feel free to remove it.

pontoon/sync/management/commands/import_pootle.py
```diff
@@ -113,13 +113,12 @@ def handle_project_locale(self, project, locale):
         users_dict = {}
         users = User.objects.filter(email__in=emails)
         missing_users_list = []
-
+        locale_translators_map = dict(Locale.objects.values_list('code', 'translators_group'))
         for u in users:
             users_dict[u.email] = u
             if not u.has_perm('base.can_localize'):
```
mathjazz: You should use the new API here:
u.has_perm('base.can_translate_locale', locale)

mathjazz:
@jotes I see that you're using uppercase locale codes in group names, is there a reason for this?

jotes:
@mathjazz not really, do you have any better convention in mind?

mathjazz:
@jotes I was thinking about simply keeping the default case, because that's what we use everywhere else.

jotes:
@mathjazz done.

mathjazz:
Thanks. I think there are 2 additional ".upper()" occurences you want to remove in pontoon/base/signals.py.

jotes:
@mathjazz mea culpa - fixed.

mathjazz:
@jotes Let's rebase into a single commit and we're ready to merge!

jotes:
@mathjazz i hope that i didn't broke anything during rebase.

jotes:
@mathjazz i've added names to permissions in migrations as you've suggested on irc.

mathjazz:
Epic stuff, @jotes!

PR #221, opened 2015-10-20
Simplify Celery settings

mathjazz:
@Osmose r?


Osmose:
`'False'` is truthy since it's a non-empty string. We need the `!= 'False'` to deal with the fact that environment variables are always strings. 

Or am I missing something?

mathjazz:
Updated.

Osmose:
r+

PR #224, opened 2015-10-20
Use new anonymous icon everywhere

mathjazz:
The old disputable anonymous icon is still cached with Gravatar.

New name (i.e. URL) should fix this.

@Osmose r?


Osmose:
r+

PR #225, opened 2015-10-21
Use AJAX to fetch main menu data

mathjazz:
This changes the way we load project, locale, resource and stats data for the main menu. Instead of loading data as part of the main server request, we make separate AJAX requests for each project. Except for the currently translated project, so its resources/subpages/search are available immediately.

AJAX requests are triggered all at once, without any smart logic, letting the browser handle them within its limits. Requests are made after loading entities, which performs better than the other way around. Projects in the project menu are hidden until their AJAX request completes.

This is deployed on [staging](https://mozilla-pontoon-staging.herokuapp.com/) and is showing significant performance improvements:
- Main request: 12s vs 2s (> 80%)
- All requests: 18s vs 9s (> 80%)
- Full page load including the new main menu AJAX calls: 18s vs 14s (> 20%)

PROS
- Decreases homepage main request (/) time by > 80%.
- Decreases homepage total request (all) time by > 50%.
- Decreases homepage fully loaded time (incl. new AJAX calls) time by > 20%.
- Decreases homepage HTML size by > 80%.
- No more homepage timeouts!

CONS
- Main menu is fully ready a few seconds after page load.
- Possible: memory consumption.

I'm impressed how big of an improvement this is given the size of a changeset. What also surprises me is the full page load time being shorter than without this PR. It must be related to the fact that I'm splitting one huge server request into serveral smaller ones. Something to note for optimizing get-entities/ request.

Next, I'm thinking about changing the Go button into AJAX call as well. That would allow users to switch between resources/subpages (and projects/locales) without triggering project-details/ AJAX requests again.

@Osmose @jotes r?

### DIFF
--- a/pontoon/base/urls.py
+++ b/pontoon/base/urls.py
@@ -54,6 +54,10 @@ urlpatterns = patterns(
         views.ProjectContributorsView.as_view(),
         name='pontoon.project.contributors'),
 
+    # AJAX: Get project details
+    url(r'^projects/(?P<slug>[\w-]+)/details/$', views.get_project_details,
+        name='pontoon.project.details'),
+
     # Translate project
     url(r'^(?P<locale>[A-Za-z0-9\-\@\.]+)/(?P<slug>[\w-]+)/(?P<part>.+)/$',
         views.translate,
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -21,6 +21,7 @@ from django.db.models import Count, F, Q
 from django.http import (
     Http404,
     HttpResponse,
+    HttpResponseBadRequest,
     HttpResponseRedirect,
 )
 
@@ -397,6 +398,21 @@ def search(request, template='search.html'):
     }
 
     return render(request, template, data)
+
+
+def get_project_details(request, slug):
+    """Get project locales with their pages/paths and stats."""
+    if not request.is_ajax():
+        raise HttpResponseBadRequest
+
+    project = get_object_or_404(Project, slug=slug)
+
+    data = {
+        "slug": slug,
+        "details": project.locales_parts_stats(),
+    }
+
+    return HttpResponse(json.dumps(data), content_type='application/json')
 
 
 def entities(request, template=None):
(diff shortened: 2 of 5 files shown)


pontoon/base/views.py
```diff
@@ -399,6 +399,32 @@ def search(request, template='search.html'):
     return render(request, template, data)
 
 
+def get_project_details(request):
+    """Get project locales with their pages/paths and stats."""
+    if not request.is_ajax():
+        log.error("Non-AJAX request")
```
Osmose: I don't think logging these kinds of errors is necessary, we can search request logs for this URL if we want stats on it.

pontoon/base/views.py
```diff
@@ -399,6 +399,32 @@ def search(request, template='search.html'):
...
+        raise Http404
```
Osmose: 400 is a more appropriate response code for this than 404.

pontoon/base/views.py
```diff
@@ -399,6 +399,32 @@ def search(request, template='search.html'):
...
+
+    try:
+        slug = request.GET['slug']
+    except MultiValueDictKeyError as e:
+        log.error(str(e))
+        return HttpResponse("error")
```
Osmose: 404 is more appropriate for this and the DoesNotExist case below.

pontoon/base/urls.py
```diff
@@ -125,6 +125,8 @@
         name='pontoon.transifex'),
     url(r'^csrf/$', views.get_csrf,
         name='pontoon.csrf'),
+    url(r'^project-details/$', views.get_project_details,
```
Osmose: I think `projects/slug/details/` is a more sensible URL for this, and lets you avoid having a try block in the view just to catch the missing parameter.

Osmose:
r+wc, this is impressively smaller than I expected. I think that speaks to good choices in the architecture of this menu, nice work!

jotes:
@mathjazz r+(+) i was thinking about something pretty similar.  We could e.g. pre-cache whole html structure and then load just counters on demand - it would require a little more changes. Anyway your PR and effects look great :dancer: :+1:

PR #226, opened 2015-10-22
Fix bug 1216452: Add support for msgctx in PO files

mathjazz:
Some 70+ strings aren't even displayed in SUMO without this.

@Osmose r?

### DIFF
--- a/pontoon/sync/formats/po.py
+++ b/pontoon/sync/formats/po.py
@@ -8,6 +8,7 @@ from django.utils import timezone
 import polib
 import sys
 
+from pontoon.sync import KEY_SEPARATOR
 from pontoon.sync.formats.base import ParseError, ParsedResource
 from pontoon.sync.vcs_models import VCSTranslation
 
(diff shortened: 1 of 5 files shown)


pontoon/sync/formats/po.py
```diff
@@ -24,8 +24,14 @@ def __init__(self, po_entry, order):
         # Remove empty strings from the string dict.
         strings = {key: value for key, value in strings.items() if value}
 
+        # Pofiles use the source as the key prepended with context if available.
+        # \04 character is used as a separator as proposed by Translate Toolkit.
+        key = po_entry.msgid
+        if po_entry.msgctxt:
+            key = po_entry.msgctxt + '\04' + key
```
Osmose: The separator should be in a constant somewhere rather than hard-coded.

pontoon/sync/tests/formats/__init__.py
```diff
@@ -246,6 +246,26 @@ def run_parse_empty_translation(self, input_string, translation_index):
                 source_string='Empty Translation',
             )
 
+    def run_parse_context(self, input_string, translation_index):
```
Osmose: If this test is pofile-specific, then it can just go in `test_po.py`. Do any other formats support context in this way?

Osmose:
This could use a test ensuring that identical strings with different contexts in the same file don't break anything.

r+wc, although I can take another look after the changes too, if you want. Nice work, really simple fix!

pontoon/sync/tests/formats/test_xliff.py
```diff
@@ -53,7 +54,7 @@ class XLIFFTests(FormatTestsMixin, TestCase):
 
     def key(self, source_string):
         """XLIFF keys are prefixed with the file name."""
-        return u'filename\x04' + super(XLIFFTests, self).key(source_string)
+        return u'filename' + KEY_SEPARATOR + super(XLIFFTests, self).key(source_string)
```
Osmose: Good catch on my mistake :D

Osmose:
r+ if the tests pass :D

PR #227, opened 2015-10-22
Fix bug 1201212: Do not approve translations with fuzzy VCS counterparts

mathjazz:
Before this PR, we approved all DB translations with matching VCS counterparts, including the ones with fuzzy VCS counterparts.

Such translations were marked as both, approved and fuzzy in Pontoon, which means:
A. They were treated as approved in the translate view.
B. They were treated as fuzzy in sync (didn't get committed).
C. They were treated as both in stats, so stats were broken.

This PR fixes A. C will be fixed the next time sync will run for these projects. We can also do it manually.

@Osmose r?

### DIFF
--- a/pontoon/sync/changeset.py
+++ b/pontoon/sync/changeset.py
@@ -196,7 +196,7 @@ class ChangeSet(object):
                                             plural_form=plural_form,
                                             string=string)
                 if db_translation:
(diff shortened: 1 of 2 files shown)


Osmose:
r+, looks great!

PR #228, opened 2015-10-22
Show all metadata by default, below original string

mathjazz:
Which means we no longer need More details button.

@jotes @Osmose r?


jotes:
r+

PR #229, opened 2015-10-22
Search by key

mathjazz:
@jotes @Osmose r?

Note before merging: This is branced from #228.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -750,7 +750,9 @@ class Entity(DirtyFieldsMixin, models.Model):
                 'marked': entity.marked,
                 'original_plural': entity.string_plural,
                 'marked_plural': entity.marked_plural,
-                'key': entity.key,
+                # Remove source string from the key and the delimiter
+                # as required by Translate Toolkit
+                'key': entity.key.split('\x04')[0],
                 'path': entity.resource.path,
                 'format': entity.resource.format,
                 'comment': entity.comment,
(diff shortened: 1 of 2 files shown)


pontoon/base/models.py
```diff
@@ -750,7 +750,8 @@ def for_project_locale(self, project, locale, paths=None, search=None):
                 'marked': entity.marked,
                 'original_plural': entity.string_plural,
                 'marked_plural': entity.marked_plural,
-                'key': entity.key,
+                # Remove source string fom the key
+                'key': entity.key.split('\x04')[0],
```
jotes: We could switch this magic hex into EOT from https://docs.python.org/2/library/curses.ascii.html. It would be good to add little info about why we're doing it and what components requires this particular format. It will be valuable for new contributors.

jotes:
r+wc

pontoon/base/models.py
mathjazz: Translate Toolkit (external lib) uses it for joining source file name and source string to form unique string identifier. We're also following the convention for PO files now:
https://github.com/mozilla/pontoon/pull/226/files

Updated comment.

pontoon/base/models.py
```diff
@@ -750,7 +750,9 @@ def for_project_locale(self, project, locale, paths=None, search=None):
                 'marked': entity.marked,
                 'original_plural': entity.string_plural,
                 'marked_plural': entity.marked_plural,
-                'key': entity.key,
+                # Remove source string from the key and the delimiter
+                # as required by Translate Toolkit
+                'key': entity.key.split('\x04')[0],
```
Osmose: Clever. Almost more clever than I'd like (in that it works even if `\x04` isn't present). Almost. :P

Osmose:
r+, nice work

mathjazz:
We could also use a new Translation attribute ("unique") instead of key, and only store "key.split('\x04')[0]" in "key" as it was initialy meant for. Or use JSON.

PR #230, opened 2015-10-22
Use timezone aware dates on Verbatim import

mathjazz:
@jotes @Osmose another quick r?


Osmose:
As long as Verbatim is using UTC, this is r+. If Verbatim is not using UTC, you'll need to pass the proper timezone to `make_aware`.

mathjazz:
Well, that's the problem - Verbatim doesn't give as any timezone, so we pretend it's UTC.

PR #231, opened 2015-10-22
Fix bug 1208393: Rename Unapproved to Suggested

mathjazz:
Also replaces "Needs work" with "Need work".

@Osmose @jotes?


Osmose:
I think "Needs work" is actually more fitting than "Need work". The rest of the terms are ambiguous as to whether they're describing a plural or a singular, but "Needs work" seems more natural to me.

r+ to whatever you choose.

mathjazz:
I always trust the native speaker! Reverted to "Needs work".

PR #233, opened 2015-10-22
Notify New Relic on deploy.

Osmose:
Was trying to figure out performance stuff this morning and got tired of not knowing when deploys are happening while on New Relic.

The necessary environment variables are already on staging and prod. I tested on staging earlier and it seems to work fine: https://rpm.newrelic.com/accounts/263620/applications/8734596/deployments

@mathjazz r?


mathjazz:
r+

PR #235, opened 2015-10-23
Fix git user name and email

mathjazz:
@Osmose r?

I wonder if this also fixes the funny fact that @stenox is marked as committer of pretty much all Pontoon commits on GitHub?

See:
https://github.com/mozilla-services/FindMyDevice-l10n/commits/master
https://github.com/mozilla/loop-client-l10n/commits/master


Osmose:
r+

PR #237, opened 2015-10-23
Reduce static resource requests with more bundling.

Osmose:
This is half of a change to both the main site and pontoon-intro that reduces the number of static resource requests by doing better bundling on the intro and translate views.

The other half can be found at https://github.com/mozilla/pontoon-intro/pull/2.

Once this gets approved, I'll merge the change into pontoon-intro and rebase this change, including an update to the submodule, and then merge. 

@mathjazz r?


mathjazz:
r+

PR #238, opened 2015-10-25
Fix bug 1215109. Added unchanged filter to project search.

jotes:
Hi @mathjazz @Osmose  could you review? Whole fix will need more tweaks in UI, i guess. However, i've added some tests for searching through entities. I hope i didn't break anything.

### DIFF
--- a/pontoon/base/templates/part_selector.html
+++ b/pontoon/base/templates/part_selector.html
@@ -1,4 +1,6 @@
 {% import 'latest_activity.html' as latest_activity %}
+{% import 'widgets/search.html' as Search %}
+
 {% set search = request.GET.urlencode() %}
 
 <!-- Part selector -->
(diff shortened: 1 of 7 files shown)


Osmose:
This one is all @mathjazz :D

pontoon/base/templates/part_selector.html
```diff
@@ -66,31 +68,16 @@
...
-          </li>
-        </ul>
+        {{ Search.options((
+          ('sources', 'Source strings', true),
+          ('translations', 'Translations', true),
+          ('comments', 'Comments', false),
+          ('keys', 'Keys', false),
+          ('casesensitive', 'Case sensitive', false),
+          ('unchanged', 'Unchanged', false),
+        ), request)}}
```
mathjazz: Beautiful!

mathjazz:
Nice work, @jotes! The code is simple and straightforward.

But I find the UI confusing. Previously available options in the list are joined with OR, whereas "Unchanged" is joined with AND. We should also add all other filters available in the sidebar search, which will make things even more confusing.

We've been talking about Search redesign for a while, so this might be the right time to re-think it.

@Osmose I wonder how far have you come with the entity list optimization (bug 1216546), because these two bugs are related. I agree with your proposal we should only have one search field. What if we remove the one in the resource menu completely, add "All resource" item to the resource menu and then only keep the search in the sidebar?

Osmose:
> @Osmose I wonder how far have you come with the entity list optimization (bug 1216546), because these two bugs are related. I agree with your proposal we should only have one search field. What if we remove the one in the resource menu completely, add "All resource" item to the resource menu and then only keep the search in the sidebar?

I haven't really done anything touching search yet. My current focus was simply refactoring the view and improving the backend performance of get-entities in the average case.

I think I like your idea, overall.

jotes:
Personally i thought about adding checkboxes on the right side on the input than below or simple label "filter by" above case-sensitive and unchanged.

PR #240, opened 2015-10-28
Fix bug 1214567: improve Pootle import/export scripts

mathjazz:
* Obsolete filename parameter
* Make locale and project parameters optional
* Use bulk save
* Move the script under the sync app

@jotes @Osmose r?


jotes:
Looks like there's one small bug in pylama.

jotes:
It would be also good to have some kind of integration tests for this script - we could add new bug for this.

jotes:
Anyway r+wc :+1:

Osmose:
> It would be also good to have some kind of integration tests for this script - we could add new bug for this.

The reason I've been ignoring the fact that this script has no tests is because it's very specific to our Pootle instance (and thus isn't usable as a generic import of Pootle data) and won't be useful after we kill Verbatim. It just really needs to work once.

mathjazz:
Agree with @Osmose about test.

Updated.

pontoon/sync/management/commands/import_pootle.py
```diff
@@ -0,0 +1,153 @@
...
+        )
+
+        parser.add_argument(
+            '--locale',
+            action='store',
+            dest='locale',
+            help='Locale code to import Verbatim data to'
+        )
+
+    def handle(self, *args, **options):
```
Osmose: You should wrap this whole thing in `transaction.atomic` so that if it fails the changes are rolled back.

pontoon/sync/management/commands/import_pootle.py
```diff
@@ -0,0 +1,153 @@
...
+        if users:
+            bulk_update(users)
+
+        for t in translations:
+            try:
+                pootle_translation = pootle[t.entity.resource.path][t.entity.string]
+
+                # Save user
+                email = pootle_translation["author"]
+                user = users.get(email=email)
```
Osmose: This does an extra query per-user I think, you may want to turn users into a dictionary keyed by email.

pontoon/sync/management/commands/import_pootle.py
```diff
@@ -0,0 +1,153 @@
...
+                date = pootle_translation["date"].split("+")[0]
+                dateObj = timezone.make_aware(datetime.strptime(date, "%Y-%m-%d %H:%M:%S"))
+                t.date = dateObj
+                t.approved_date = dateObj
+
+            except User.DoesNotExist:
+                missing_users_list.append(email)
+
+            # Path - entity mismatch, skip to next translation
+            except KeyError:
```
Osmose: You'll want to move this up to only surround the `pootle_translation` line so you don't catch any unexpected errors, and also saves you some indenting. :D

pontoon/sync/management/commands/import_pootle.py
```diff
@@ -0,0 +1,153 @@
...
+                continue
+
+        if translations:
+            bulk_update(translations)
+
+        missing_users = Counter(missing_users_list)
```
Osmose: THE MOST CLEVER THING

Osmose:
r+wc

PR #241, opened 2015-10-28
Add STATIC_HOST setting to support CDN use.

Osmose:
I've already had jp help set us up with a cloudfront CDN for prod (e.g. https://d183y7hobzemob.cloudfront.net/static/css/base.min.09ddf404e0aa.css). Once this hits prod we can set the `STATIC_HOST` environment variable and all of our static assets should be served through a CDN instead of directly through our site.

@mathjazz r?


mathjazz:
r+

PR #242, opened 2015-10-28
Fix bug 1216988: Redirect pt to pt-PT and ga to ga-IE

mathjazz:
@Osmose @jotes r?


jotes:
r+ :)

Osmose:
r+

Although this will become a concern in the future if we ever aim to make Pontoon generic for non-Mozilla installations. Just to keep that in mind.

PR #243, opened 2015-10-29
Bug 1201431: remove missing files & folders from HG

mathjazz:
We need this pretty quickly, because the merge day just happenned today instead of Monday. I'm only implementing it for HG for now, which is also the only place we currently need it.

It works on the old Pontoon server (within the mozilla-en-US script):
http://hg.mozilla.org/users/m_owca.info/firefox-aurora/rev/72d219e997ef

I'll add tests once I extend it to other VCS types. Although I'm not exactly sure how to test these things.

@Osmose @jotes r?

### DIFF
--- a/bin/mozilla-en-US.py
+++ b/bin/mozilla-en-US.py
@@ -76,8 +76,8 @@ def pull(url, target):
 
 
 def push(path):
-    # Add
-    execute(['hg', 'add'], path)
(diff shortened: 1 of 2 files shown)


Osmose:
r+

Testing this is hard because I don't want to get into specifics of how each VCS works in our tests.

PR #244, opened 2015-10-29
Bug 1214670: Explicitly push master when dealing with git.

Osmose:
I don't get the warning locally but commit totally still works and I believe this will get rid of the warning.

@mathjazz r?


mathjazz:
r+

PR #245, opened 2015-10-30
Css fixes

mathjazz:
@jotes r?

Tabs in the resource/search menu aren't align properly in production, because "calc(50% + 2px)" is eaten by minifer. This solution uses less-hacky approach for tab positioning.


jotes:
r+

PR #246, opened 2015-10-30
Make request locale UX more pleasant

mathjazz:
I'm trying to make messaging around requesting new locale clearer. Because I'm getting many duplicate submissions. Which means UX sucks. I suppose for several reasons:

1) On project pages with longer locale list, confirmation is never visible.
-> Scrolling to top after request is made.
2) "Add" implies users can add locales themselves to projects.
-> Changing to "Request".
3) We never told anyone they will be notified when locale is enabled via email.
-> Adding that to the confirmation message.

@Osmose @jotes r?

### DIFF
--- a/pontoon/base/static/css/style.css
+++ b/pontoon/base/static/css/style.css
@@ -751,7 +751,7 @@ label {
 }
 
 .search-wrapper > a.new:after {
-  content: "Add";
+  content: "Request";
 }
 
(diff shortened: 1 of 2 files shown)


jotes:
r+  :)

Osmose:
> r+ :)

Ditto!

PR #247, opened 2015-11-02
Bug 952488 - Add locale managers UI

mathjazz:
Adding team admin page, available to locale managers, where they can select privileged users. This PR is branched from #219, so only check the last commit. I will rebase once #219 gets merged.

I'm especially interested in more elegant solution for exposing the "Admin" in Locale dashboard to locale managers only.

@jotes @Osmose r?

### DIFF
--- a/pontoon/base/signals.py
+++ b/pontoon/base/signals.py
@@ -1,8 +1,8 @@
 from guardian.models import GroupObjectPermission
 
-from django.core.exceptions import ObjectDoesNotExist
 from django.contrib.auth.models import Group, Permission
 from django.contrib.contenttypes.models import ContentType
+from django.core.exceptions import ObjectDoesNotExist
 from django.db.models.signals import post_save, pre_save
 from django.dispatch import receiver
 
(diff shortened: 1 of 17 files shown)


mathjazz:
@Osmose @jotes This is now ready for another look. Not sure why the test is failing, it works locally. But I'll rebase against master after we merge #219, so it's irrelevant.

This is a full diff against #219:
https://github.com/mathjazz/pontoon/compare/70f6240...mathjazz:b7b68e5#files_bucket

I decided not to use forms, because that would cause changes to HTML which would require refactoring CSS (multi select widget design) and JS (search). And the only benefit of using forms in this case is validation, which is IMHO not really a big deal for us (because we're dealing with a closed collection of users anyways).

Osmose:
I disagree with the logic to not use forms (I could, for example, flood the site with 500 errors by submitting non-integer ids), but it's your choice to avoid it for now or not.

Other than that, I don't see anything major that I haven't commented on. r+, good work!

PR #248, opened 2015-11-04
Fix bug 1214387. Added decorator to run only one task at time.

jotes:
hi @Osmose, could you look at this?

### DIFF
--- a/pontoon/settings/test.py
+++ b/pontoon/settings/test.py
@@ -1,4 +1,4 @@
 """Django settings for the test suite."""
 
 PIPELINE_ENABLED = False
-CELERY_ALWAYS_EAGER = 1
+CELERY_ALWAYS_EAGER = True
--- a/pontoon/settings/test.py
+++ b/pontoon/settings/test.py
@@ -1,3 +1,4 @@
 """Django settings for the test suite."""
 
 PIPELINE_ENABLED = False
+CELERY_ALWAYS_EAGER = 1
--- a/pontoon/sync/core.py
+++ b/pontoon/sync/core.py
@@ -6,6 +6,7 @@ from collections import Counter
 from celery import shared_task
 from django.contrib.auth.models import User
 
+from django.core.cache import cache
 from django.db import transaction
 from django.template.loader import render_to_string
 
@@ -38,18 +39,14 @@ def serial_task(timeout, **celery_args):
     :param timeout: time after which lock is released.
     """
     def wrapper(func):
-
         @shared_task(bind=True, **celery_args)
         @wraps(func)
         def wrapped_func(self, *args, **kwargs):
-            # Moved import here to mock cache during tests.
-            from django.core.cache import cache
-
             lock_name = "serial_task.{}".format(self.name)
             # Acquire the lock
             if not cache.add(lock_name, True, timeout=timeout):
-                raise RuntimeError("Can't execute task: {}  because the previous"
-                    " task is already running.".format(self.name))
+                raise RuntimeError("Can't execute task '{}' because the previously called"
+                    " task is still running.".format(self.name))
             try:
                 return func(self, *args, **kwargs)
             finally:
--- a/pontoon/sync/core.py
+++ b/pontoon/sync/core.py
@@ -1,7 +1,11 @@
+from functools import wraps
 import logging
+
 from collections import Counter
 
+from celery import shared_task
 from django.contrib.auth.models import User
+
 from django.db import transaction
 from django.template.loader import render_to_string
 
@@ -14,7 +18,6 @@ from pontoon.sync.changeset import ChangeSet
 from pontoon.sync.vcs_models import VCSProject
 from pontoon.sync.utils import locale_directory_path
 
-
 log = logging.getLogger(__name__)
 
 
@@ -26,6 +29,35 @@ def sync_project(db_project, now):
         changeset = ChangeSet(db_project, vcs_project, now)
         update_entities(db_project, vcs_project, changeset)
         changeset.execute()
+
+
+def serial_task(timeout, **celery_args):
+    """
+    Decorator ensures that there's only one running task with given task_name.
+    Decorated tasks are bound tasks, meaning their first argument is always their Task instance
+    :param timeout: time after which lock is released.
+    """
+    def wrapper(func):
+
+        @shared_task(bind=True, **celery_args)
+        @wraps(func)
+        def wrapped_func(self, *args, **kwargs):
+            # Moved import here to mock cache during tests.
+            from django.core.cache import cache
+
+            lock_name = "serial_task.{}".format(self.name)
+            print "lock_name", lock_name
+            # Acquire the lock
+            if not cache.add(lock_name, True, timeout=timeout):
+                raise RuntimeError("Can't execute task: {}  because the previous"
+                    " task is already running.".format(self.name))
+            try:
+                return func(self, *args, **kwargs)
+            finally:
+                # release the lock
+                cache.delete(lock_name)
+        return wrapped_func
+    return wrapper
 
 
 def collect_entities(db_project, vcs_project):
--- a/pontoon/sync/core.py
+++ b/pontoon/sync/core.py
@@ -46,7 +46,6 @@ def serial_task(timeout, **celery_args):
             from django.core.cache import cache
 
             lock_name = "serial_task.{}".format(self.name)
-            print "lock_name", lock_name
             # Acquire the lock
             if not cache.add(lock_name, True, timeout=timeout):
                 raise RuntimeError("Can't execute task: {}  because the previous"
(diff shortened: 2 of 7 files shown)


Osmose:
Will get to this after my work on [bug 1214670](https://bugzilla.mozilla.org/show_bug.cgi?id=1214670) is done, which might take a bit. Sorry for the delay.

jotes:
No worries!

pontoon/sync/core.py
```diff
@@ -20,6 +21,27 @@
 
 log = logging.getLogger(__name__)
 
+def run_only_one(func):
```
Osmose: The naming for this is actually really hard. I think it should have the word "task" in it to signify it's only for celery tasks.

After asking around I think `serial_task` is a decent name for this.

pontoon/sync/core.py
```diff
@@ -20,6 +21,27 @@
...
+    """
+    Decorator ensures that there's only one running task with given task_name.
+    """
+
+    task_name = "celery.{}".format(func.__name__)
```
Osmose: This bars us from having two separate tasks with the same name. Normally celery tasks are named after their containing module and the function name to avoid naming conflicts.

Since this is meant to wrap tasks anyway, you may want to just decorate the wrapped function with Celery's task decorator (and pass arguments given to your decorator to it so that we can still pass arguments to celery), and set `bind` to True so that you can access `self.name`, which is a guaranteed unique task name generated by celery.

pontoon/sync/core.py
```diff
@@ -20,6 +21,27 @@
...
+    def wrapped_func(*args, **kwargs):
+        # Moved import here to mock cache during tests.
+        from django.core.cache import cache
+
+        # Acquire the lock
+        if not cache.add(task_name, True):
```
Osmose: You'll want to set an expiry on this set to something big, like 3 hours. Otherwise, a stray lock could block sync from _ever_ happening.

pontoon/sync/core.py
```diff
@@ -20,6 +21,27 @@
...
+            return
```
Osmose: At a minimum we want to log here, and ideally we want to get a more noisy notification when this happens, as it's a bad thing if it does. #251 adds an error handler to tasks that takes exceptions raised in tasks and sends them to RayGun, so I think raising a `RuntimeError` with a descriptive message would be a good thing to do here.

pontoon/sync/core.py
jotes: Yeah, but uncontrolled expire could spawn more processes that could stray lock... i could add send exception via RaygunApi to be notified about such things.

pontoon/sync/tests/test_tasks.py
```diff
@@ -15,3 +15,47 @@ def test_missing_project(self):
...
+
+class SyncExecutionTests(TestCase):
+    def setUp(self):
+        self.project = ProjectFactory.create()
+
+    def test_primary_process(self):
+        """
+        Test if sync will create lock in cache and release this after task is done.
+        """
+        with patch('django.core.cache.cache') as mock_cache, \
```
Osmose: Mocking the cache and ensuring it sets the right cache keys is a decent way to unit test this, but IMO it's brittle and we don't want to unit test this. The thing we care about is that we can't run two instances of this task at the same time. So what we really want to do is create a function that we pause during execution, and then attempt to run another function at the same time and observe the resulting error.

A reasonable approach would be to write a function decorated with the`@run_once_only` decorator that takes in a callback and executes it. Then, we call it and pass in a callback that attempts to call the decorated function again, and asserts that it raises an exception, since it's already in the middle of running once.

This also avoids testing if `sync_project` is decorated with the decorator, and only tests if the decorator does what we think it does.

pontoon/sync/tests/test_tasks.py
```diff
@@ -15,3 +15,47 @@ def test_missing_project(self):
...
+            patch('pontoon.sync.tasks.perform_sync') as mock_perform_sync:
+            mock_cache.add.return_value = False
+
+            sync_project(self.project.pk)
+
+            mock_cache.assert_has_calls([call.add('celery.sync_project', True)])
+            assert_false(mock_perform_sync.called)
+
+    def test_exception_during_sync(self):
```
Osmose: Good thing to test for!

pontoon/sync/core.py
Osmose: What do you mean? If the timeout is three hours and a sync job exits without releasing the lock, then 3 hours later the lock will expire and the next sync job will run again, and either succeed or fail normally. Why would this cause an issue?

Osmose:
This is actually a really nice PR and is a great first pass. Nice work thusfar!

jotes:
@Osmose could you look again?

Osmose:
Nice progress, this is a lot closer. Thanks for the work!

jotes:
@Osmose I'm sorry but it was a lot easier to rebase things into single commit and then rebase with current master. Could you look again?

pontoon/settings/test.py
```diff
@@ -1,3 +1,4 @@
 """Django settings for the test suite."""
 
 PIPELINE_ENABLED = False
+CELERY_ALWAYS_EAGER = 1
```
Osmose: `True` works fine as well. :P

jotes: Ah, sorry - i'll fix it.

On Wed, Nov 11, 2015 at 7:42 PM, Michael Kelly <notifications@github.com>
wrote:

> In pontoon/settings/test.py
> <https://github.com/mozilla/pontoon/pull/248#discussion_r44568595>:
>
> > @@ -1,3 +1,4 @@
> >  """Django settings for the test suite."""
> >
> >  PIPELINE_ENABLED = False
> > +CELERY_ALWAYS_EAGER = 1
>
> True works fine as well. :P
>
> —
> <https://github.com/mozilla/pontoon/pull/248/files#r44568595>.
>



-- 
jarek@reijutsu:~$ fortune
jarek@reijutsu:~$ fortune
You now have Asian Flu.

pontoon/sync/core.py
```diff
@@ -28,6 +31,34 @@ def sync_project(db_project, now):
...
+    Decorator ensures that there's only one running task with given task_name.
+    Decorated tasks are bound tasks, meaning their first argument is always their Task instance
+    :param timeout: time after which lock is released.
+    """
+    def wrapper(func):
+
+        @shared_task(bind=True, **celery_args)
+        @wraps(func)
+        def wrapped_func(self, *args, **kwargs):
+            # Moved import here to mock cache during tests.
```
Osmose: I KNEW I forgot to mention something last pass over this. This comment is inaccurate. Also, we can move this import with the rest of the imports now.

pontoon/sync/tests/test_tasks.py
```diff
@@ -150,3 +154,65 @@ def test_remove_duplicate_approvals(self):
...
+        assert_true(second_call.failed())
+        assert_raises(RuntimeError, second_call.get)
+
+    def test_release_lock_after_timeout(self):
+        """
+        Tests if lock is released after specified timeout.
+        """
+        @serial_task(3)
+        def timeout_task(self):
+            return wait(lambda: not cache.get('serial_task.pontoon.sync.tests.test_tasks.timeout_task'),
```
Osmose: I'm not a fan of adding a new library just to wait for this, and I'm still uncomfortable with having tests that are reliant on the timing between two separate areas of the code. It seems very flimsy.

I think at this point we should just mock the cache for this one test and confirm the timeout was passed properly and get on with our lives.

(You shouldn't need to move that cache import to mock it so still move it to the top of the file)

Osmose:
r+wc, this looks great! Nice work!

jotes:
@Osmose okay, test of timeout is now checking only mock calls.

pontoon/sync/tests/test_tasks.py
```diff
@@ -161,3 +164,60 @@ def test_remove_duplicate_approvals(self):
...
+        with patch('pontoon.sync.core.cache') as mock_cache:
+            @serial_task(3)
+            def timeout_task(self):
+                return 42
+            first_call = timeout_task.delay()
+
+            assert_false(cache.get('serial_task.pontoon.sync.tests.test_tasks.timeout_task'))
+            assert_true(first_call.successful())
+            assert_equal(first_call.get(), 42)
+            mock_cache.assert_has_calls([call.add('serial_task.pontoon.sync.tests.test_tasks.timeout_task', True, timeout=3),
```
Osmose: You're asserting too much, here. All we want to know is if the proper timeout was passed. `mock_cache.add.assert_called_with(ANY, ANY, timeout=3)` is plenty.

Osmose:
Merged in https://github.com/mozilla/pontoon/commit/8e1a35c7d88a7dd882e4d33a47ff4326903e9c0a, last nit was easily fixable while I rebased. Nice work! :D

PR #253, opened 2015-11-09
Only grant new permissions to users with old can_localize permission

mathjazz:
Two changes:

1. Permission shouldn't be granted to users who didn't have old (can_localize) permissions. They could have approved translations, because someone else approved them for them.

2. Also, approved_translations__gte=1 is implicit with translation__approved=True.

@jotes r?


jotes:
r+

PR #254, opened 2015-11-10
Use per-view transactions when modifying the database.

Osmose:
This should help with the IntegrityError in https://app.raygun.io/crashreporting/cqwxbq/errors/935498833, although it won't help with [bug 1220286](https://bugzilla.mozilla.org/show_bug.cgi?id=1220286).

@mathjazz r?

### DIFF
--- a/pontoon/administration/views.py
+++ b/pontoon/administration/views.py
@@ -71,6 +71,7 @@ def get_slug(request):
     return HttpResponse(slug)
 
 
+@transaction.atomic
 def manage_project(request, slug=None, template='admin_project.html'):
     """Admin project."""
     log.debug("Admin project.")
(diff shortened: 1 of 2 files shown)


mathjazz:
Nice catch!

I'm curious why do we need to make e.g. request_locale() atomic? It doesn't even change the database.Or quality_checks_switch(), which only changes the user profile.

Osmose:
> I'm curious why do we need to make e.g. request_locale() atomic? It doesn't even change the database. 

Because I forgot that all it does is email you. I'll remove it.

> Or quality_checks_switch(), which only changes the user profile.

Basically any view that touches the DB at all should have a transaction. In this case it protects against changes to the profile that may occur while the view is executing.

mathjazz:
Thanks for explaining! r+

PR #255, opened 2015-11-10
Allow Managers to see and edit other Managers of their Team

mathjazz:
Also: Rename Team Admin(s) to Team Manage(rs) to distinguish against
(Pontoon) Admins. Will be useful when we add Project Managers.

@Osmose @jotes r?


Osmose:
I actually really like the UI for this, it's neato!

pontoon/base/templates/locale_manage.html
```diff
@@ -0,0 +1,71 @@
...
+{% block title %}{{ super() }} &middot; Manage {% endblock %}
+
+{% block locale_subpage %}
+<form method="POST" action="{{ url('pontoon.locale.manage', locale.code) }}">
+  <input type="hidden" value="{{ csrf_token }}" name="csrfmiddlewaretoken">
+  <input type="hidden" name="translators">
+  <input type="hidden" name="managers">
+
+  <div class="clearfix">
+      <div class="user select available">
```
Osmose: This looks like a potential macro, since the only difference between these three is the classname, input id, label, and description. That might be a bad idea, though. It's hard to tell with HTML when a macro is a good idea.

mathjazz: I'm already playing with this locally, let's see.

pontoon/base/views.py
```diff
@@ -118,21 +117,27 @@ def locale_admin(request, locale):
...
+
+    if request.method == 'POST':
+        update_group(l.translators_group, 'translators')
+        update_group(l.managers_group, 'managers')
 
-    users_translators = l.translators_group.user_set.all()
-    users_all = User.objects.exclude(pk__in=users_translators).exclude(email="")
-    users_contributors = User.translators.filter(translation__locale=l).distinct()
+    managers = l.managers_group.user_set.all()
+    translators = l.translators_group.user_set.exclude(pk__in=managers).all()
```
Osmose: Is there a way you could filter by the group itself instead of the managers subquery? Like `group_set=l.managers_group` or something. I'd expect that to be faster than the subquery, which may end up mattering at high user counts.

pontoon/base/static/js/locale_manage.js
```diff
@@ -0,0 +1,65 @@
+$(function() {
+
+  // Before submitting the form, update translators and managers
+  $('form').submit(function (e) {
+    $.each(['translators', 'managers'], function(i, value) {
```
Osmose: What's our browser support? `Array.forEach` has been around for a long time and is supported in modern browsers.

Osmose: Or if you enable Babel for this file you can use a `for..of` loop.

pontoon/base/static/js/locale_manage.js
```diff
@@ -0,0 +1,65 @@
...
+
+  // While in contributors tab, search contributors only
+  $('.menu input[type=search]').on('keyup.search', function(e) {
+    if ($('.user.available label a.contributors').is('.active')) {
+      $('.user.available li:not(".contributor")').hide();
+    }
+  });
+
+  // Set translators arrow direction
+  function setArrow(element, event) {
```
Osmose: This works and is neato, but I would've just put two spans inside and had them each take up half of the list item. Which might be worse for long lists of users, maybe.

Osmose:
Use Django forms already jeeeeeeeezzzz

r+, nothing major here. Nice work!

pontoon/base/static/js/locale_manage.js
mathjazz: I was actually hoping to find a better UI and avoid any of this. But I didn't. I think having to select a user first and then clicking on a button (left/right) is even worse.

pontoon/base/views.py
mathjazz: I didn't find anything like this.

PR #257, opened 2015-11-10
Support for coveralls.io

jotes:
This service monitors tests coverage in project. Example report is here: https://coveralls.io/builds/4111747
I think it would be nice addition to pontoon. The setup is very easy thing - all is documented @ coveralls.io .
@Osmose @mathjazz could you review and share your thoughts?

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,4 @@
+[run]
+branch = true
+include = pontoon/*
+omit = *migrations*
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,14 @@ before_install:
   - git submodule update --init --recursive
 install:
   - pip install -r requirements.txt
+  - pip install coveralls coverage
 before_script:
   - psql -c 'create database pontoon;' -U postgres
 script:
   - pylama
-  - python manage.py test
+  - python manage.py test --with-coverage
+after_success:
+  - coveralls
 addons:
   postgresql: "9.4"
 env:


Osmose:
I have rather strong negative feelings about code coverage. The idea is that higher coverage indicates that your tests are running more code while they execute and thus can make guarantees about the behavior of a larger set of code, but I don't think that holds. Coverage says nothing about the quality of your tests, nor does it actually say whether the "covered" code is actually being tested as opposed to being incidentally executed during a test.

In practice, you end up gaming your tests to increase coverage instead of actually trying to make things better. Or, if you don't, you end up with a sub-100% number that drags on your motivation as it never rises. Or, if you hit 100%, you end up adding a barrier to new code that also drags you down.

Of course a lot of this is based on personal experience, so I'm interested in hearing alternate viewpoints. :D

jotes:
Yeah, i know that "100% of coverage" is a some kind of mantra for a lot of people. However, i don't see it that way. 
You can easily specify what these 100% means by excluding/including critical modules for your code.
Coverage helps a lot with refactoring/removing unused parts of code. Also, coverage can point out  places that require more attention because there aren't tested at all (it's helpful when you add new code).

I would think about coverage more like a navigator than a really bad chosen requirement.

Osmose:
> Coverage helps a lot with refactoring/removing unused parts of code. 

I'm curious, how does it help in that situation?

mathjazz:
Hmmmm, before reading comments by @Osmose I was sold for this change, but now I'm not exactly sure what to say.

I think nothing's wrong with having coverage data available, but you always need to have what @Osmose wrote at the back of your head. It's somewhat similar to translation stats - when you're at 100% you think you're done, but you could still have a bunch of typos. Or to a number of bugs.

It probably also depends on what kind of personality you are. I'm deeply affected by numbers in general. ;) Meaning I would probably quickly go and make code with 90+% coverage reach 100%. Even if it's totally irrelevant. It's so tempting!

jotes:
I think coverage it's a lot like testing -> there still will be people
whose won't test because they're afraid of amount of work that will be
generated by invalid tests.
There's no such thing like a 100% of certainty that your/our code will work
correctly and tests with coverage won't guarantee 100% success. However,
(imho) that's not a reason to abandon them, because coverage still tells
you a lot about your tests.
As Osmose said, it's a very easy tempting to abuse idea of coverage by
bending whole universe to finally display these 100%.

I agree with all your comments, however i'm still standing on the more
positive side of things. I believe we'll be able to figure it out in good
way for project and us. Also, this isn't big change so, we can disable it
at any point in time or just won't merge at all.

On Wed, Nov 11, 2015 at 9:19 AM, Matjaž Horvat <notifications@github.com>
wrote:

> Hmmmm, before reading comments by @Osmose <https://github.com/Osmose> I
> was sold for this change, but now I'm not exactly sure what to say.
>
> I think nothing's wrong with having coverage data available, but you
> always need to have what @Osmose <https://github.com/Osmose> wrote at the
> back of your head. It's somewhat similar to translation stats - when you're
> at 100% you think you're done, but you could still have a bunch of typos.
> Or to a number of bugs.
>
> It probably also depends on what kind of personality you are. I'm deeply
> affected by numbers in general. ;) Meaning I would probably quickly go and
> make code with 90+% coverage reach 100%. Even if it's totally irrelevant.
> It's so tempting!
>
> —
> <https://github.com/mozilla/pontoon/pull/257#issuecomment-155701269>.
>



-- 
jarek@reijutsu:~$ fortune
jarek@reijutsu:~$ fortune
You now have Asian Flu.

jotes:
Hi! I can open this PR again, if we ever change our minds.

PR #258, opened 2015-11-11
Fixed assertion in test: pontoon.sync.tests.test_core:CommitChangesTe…

jotes:
@Osmose could you look at this? Currently tests can't pass on master :/


Osmose:
I'm freaking out, man. Swear to god, I had `first_author` in that assert and it was working locally but failing on Travis and I couldn't figure out why. I even had someone else look it over right next to me and agree that it was correct. WHY IS IT SECOND_AUTHOR

r+

PR #259, opened 2015-11-11
Cycle celery workers and add --force option to sync_projects.

Osmose:
There's two commits here:

- Adds a `--force` parameter to `sync_projects` that bypasses the logic that skips repos with no changes. I needed this while attempting to profile the sync process so that I didn't have to keep making changes every time I ran sync.
- Adds the `CELERYD_MAX_TASKS_PER_CHILD` setting, which sets the number of tasks a celery worker can handle before being replaced by a new worker. The hope is that this will help rein in the amount of memory celery workers use (see the explanation and workaround sections from [this blog post](http://chase-seibert.github.io/blog/2013/08/03/diagnosing-memory-leaks-python.html) for my inspiration).

I'm still going to see if I can reduce our memory consumption but this may help get rid of the warnings, and we can configure the setting to be lower or higher if it affects our sync performance.

@mathjazz r?

### DIFF
--- a/docs/admin/deployment.rst
+++ b/docs/admin/deployment.rst
@@ -35,6 +35,10 @@ you create:
    Celery or executed immediately and synchronously. Set this to ``False`` on
    production.
 
+``CELERYD_MAX_TASKS_PER_CHILD``
+   Maximum number of tasks a Celery worker process can execute before it’s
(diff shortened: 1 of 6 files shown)


mathjazz:
Oh I like both of these, nice work! r+

PR #260, opened 2015-11-11
Make translate view header always visible

mathjazz:
Project loader now overlaps everything *but* the header (where all the
menus are). It also uses lighter background colors, so menus are visible
if openned while projects are loading.

@Osmose @jotes r?


jotes:
r+

Osmose:
r+

PR #262, opened 2015-11-12
Display what locales does the translator have permissions for

mathjazz:
Also check if user has translator permission on each project load (potential locale change).

And also remove the old can_localize permission.

@jotes @Osmose r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -16,6 +16,7 @@ from django.utils import timezone
 from django.utils.functional import cached_property
 
 from dirtyfields import DirtyFieldsMixin
+from guardian.shortcuts import get_objects_for_user
 from jsonfield import JSONField
 
 from pontoon.administration.vcs import commit_to_vcs, get_revision, update_from_vcs
@@ -27,6 +28,7 @@ from pontoon.sync import KEY_SEPARATOR
 log = logging.getLogger('pontoon')
 
 
+# User class extensions
 class UserTranslationsManager(models.Manager):
     """
     Provides various method to interact with larger sets of translations and their stats for user.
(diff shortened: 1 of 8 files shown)


pontoon/base/models.py
```diff
@@ -27,6 +27,7 @@
 log = logging.getLogger('pontoon')
 
 
+# User class extensions
```
Osmose: Someday we'll just have a custom user class. SOMEDAY

pontoon/base/migrations/0045_auto_20151112_0857.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
```
Osmose: I've tried to make a habit of renaming auto-named migrations with a more descriptive name, it helps make things clearer.

pontoon/base/models.py
```diff
@@ -89,17 +90,40 @@ def translations_count(query=None):
         )
 
 
-# User class extensions
+@property
+def user_translated_locales(self):
```
Osmose: Could you use `get_objects_for_user` along with the permissions to build this instead? https://github.com/django-guardian/django-guardian/blob/devel/guardian/shortcuts.py#L287

pontoon/base/models.py
```diff
@@ -89,17 +90,40 @@ def translations_count(query=None):
...
+        if hasattr(group, 'managed_locales'):
+            locale = group.managed_locales.first()
+            if locale:
+                locales.append(locale)
+
+    return [l.code for l in set(locales)]
+
+
+@property
+def user_display_permissions(self):
```
Osmose: This is display logic and I think would do better as a template helper function or filter.

Osmose:
The model methods could use some tests with setting up permissions and checking if they return the right values.

r+wc otherwise, looks nice and simple!

pontoon/base/models.py
mathjazz: Ohhhhh! accept_global_perms=False!

jotes: :+1:

pontoon/base/models.py
jotes: :+1: i think we should create new macro or something in the T(emplates)-layer.

Osmose: A macro would be okay but I prefer helpers for things like this because it's mostly logic rather than markup. If it was more markup-heavy than a macro would make more sense.

That's my preference though, it doesn't make a huge difference either way.

PR #265, opened 2015-11-12
Fix bug 1200184: Allow translators to make suggestions

mathjazz:
Uses different title (Suggest) and color (blue) for Save button if
user make suggestions (either because they don't have translator
privileges or because they explicitly force to make suggestions in
preferences).

Other minor fixes and improvements I *had* to make:
* Remove obsolete user profile fields
* Fix notification loader position on translate page
* Make entire plural tab area clickable
* Fix unsaved changes warning when adding suggestions
* Check for unsaved changed when clicking Go button
* Fix Safari error with onpopstate event, triggered on first page load

@Osmose @jotes r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -127,11 +127,8 @@ class UserProfile(models.Model):
     user = models.OneToOneField(User, related_name='profile')
 
     # Other fields here
-    transifex_username = models.CharField(max_length=40, blank=True)
-    transifex_password = models.CharField(max_length=128, blank=True)
-    svn_username = models.CharField(max_length=40, blank=True)
-    svn_password = models.CharField(max_length=128, blank=True)
     quality_checks = models.BooleanField(default=True)
+    force_suggestions = models.BooleanField(default=False)
 
 
 def validate_cldr(value):
--- a/pontoon/base/static/js/main.js
+++ b/pontoon/base/static/js/main.js
@@ -525,23 +525,37 @@ $(function() {
         .find('section.' + section).show();
   });
 
-  // Toggle quality checks
-  $('.check-box.quality-checks').click(function() {
+  // Toggle user profile attribute
+  $('.check-box').click(function() {
     var self = $(this);
     Pontoon.startLoader();
 
     $.ajax({
-      url: '/quality-checks-switch/',
+      url: '/api/v1/user/' + Pontoon.user.email + '/',
       type: 'POST',
       data: {
-        csrfmiddlewaretoken: $('#server').data('csrf')
+        csrfmiddlewaretoken: $('#server').data('csrf'),
+        attribute: self.data('attribute'),
+        value: !self.is('.enabled')
       },
       success: function(data) {
         if (data === 'ok') {
           self.toggleClass('enabled');
-          var status = self.is('.enabled') ? 'enabled' : 'disabled';
-          Pontoon.endLoader('Quality checks ' + status + '.');
+          var is_enabled = self.is('.enabled'),
+              status = is_enabled ? 'enabled' : 'disabled';
+
+          Pontoon.endLoader(self.text() + ' ' + status + '.');
           $('.notification').addClass('menu-open');
+
+          if (self.is('.force-suggestions')) {
+            Pontoon.user.forceSuggestions = is_enabled;
+            Pontoon.postMessage("UPDATE-ATTRIBUTE", {
+              object: 'user',
+              attribute: 'forceSuggestions',
+              value: is_enabled
+            });
+            $('[id^="save"]').toggleClass('suggest', is_enabled);
+          }
         }
       },
       error: function() {
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -148,7 +148,7 @@ var Pontoon = (function (my) {
                 (this.approved ? ' class="approved"' : '') +
                 'title="Click to copy">' +
                   '<header class="clearfix' +
-                    ((self.user.localizer) ? ' localizer' :
+                    ((self.user.isTranslator) ? ' translator' :
                       ((self.user.email === this.email && !this.approved) ?
                         ' own' : '')) +
                     '">' +
@@ -194,6 +194,22 @@ var Pontoon = (function (my) {
      */
     appendMetaData: function (title, text) {
       $('#metadata').append('<p><span class="title">' + title + '</span>' + text + '</p>');
+    },
+
+
+    /*
+     * Update current translation length
+     */
+    updateCurrentTranslationLength: function () {
+      $('#translation-length .current-length').html($('#translation').val().length);
+    },
+
+
+    /*
+     * Update cached translation, needed for unsaved changes check
+     */
+    updateCachedTranslation: function () {
+      this.cachedTranslation = $('#translation').val();
     },
 
 
@@ -284,10 +300,10 @@ var Pontoon = (function (my) {
           translationString = entity.translation[0].string,
           translation = translationString ? translationString.length : 0;
 
-      $('#translation-length')
-        .show() // Needed if sidebar opened by default
-        .find('.original-length').html(original).end()
-        .find('.current-length').html(translation);
+      // Need to show if sidebar opened by default
+      $('#translation-length').show().find('.original-length').html(original).end();
+      self.updateCurrentTranslationLength();
+      self.updateCachedTranslation();
 
       // Update entity list
       $("#entitylist .hovered").removeClass('hovered');
@@ -350,11 +366,10 @@ var Pontoon = (function (my) {
         return callback();
       }
 
-      var pluralForm = this.getPluralForm(true),
-          translation = entity.translation[pluralForm].string,
-          source = $('#translation').val();
+      var before = this.cachedTranslation,
+          after = $('#translation').val();
 
-      if ((translation !== null) && (translation !== source)) {
+      if ((before !== null) && (before !== after)) {
         $('#unsaved').show();
         $("#translation").focus();
         this.checkUnsavedChangesCallback = callback;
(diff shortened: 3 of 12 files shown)


pontoon/base/views.py
```diff
@@ -1224,20 +1228,19 @@ def save_to_transifex(request, template=None):
...
-def quality_checks_switch(request):
-    """Turn quality checks on/off for the current user."""
-    log.debug("Turn quality checks on/off for the current user.")
+def toggle_user_profile_attribute(request, attribute):
+    profile = request.user.profile
 
-    if request.method != 'POST':
-        log.error("Non-POST request")
-        raise Http404
+    if not hasattr(profile, attribute):
```
jotes: I like the idea of generalization of such things, but in this case i have this fear that at some point this view could be easily abused if user could "toggle" any field in model. I mean, if there was a property similar to "is_superuser".

mathjazz: Good point. We might add many things in the user profile later on and we'll totally forget about this. Updated.

Osmose:
Have you considered adding an extra button for submitting suggestions instead of using a setting? I think it'd require less changes and allow translators to submit both approved translations and suggestions without having to switch their profile back and forth, and it also makes it more obvious to them that they have the option in the first place.

pontoon/base/static/js/translate.js
```diff
@@ -1916,7 +1938,8 @@ var Pontoon = (function (my) {
       this.user = {
         email: $('#server').data('email') || '',
         name: $('#server').data('name') || '',
-        localizer: isLocalizer($('#server').data('user-translated-locales')),
+        forceSuggestions: $('.force-suggestions').is('.enabled'),
```
Osmose: It seems weird to read this from the state of the checkbox vs the server data tag. The box being checked isn't actually strongly linked to what the real value on the server is (because maybe autocomplete screws up keeping it checked or the user's JS loads slowly and they click the checkbox before we reach this point).

pontoon/base/urls.py
```diff
@@ -55,9 +55,15 @@
         name='pontoon.project.contributors'),
 
     # AJAX: Get project details
-    url(r'^projects/(?P<slug>[\w-]+)/details/$', views.get_project_details,
+    url(r'^projects/(?P<slug>[\w-]+)/details/$',
+        views.get_project_details,
         name='pontoon.project.details'),
 
+    # AJAX: Toogle user profile attribute
+    url(r'^toggle-user-profile-attribute/(?P<attribute>[\w-]+)/$',
```
Osmose: This URL is too specific (the old one was too). If we follow this pattern we'd have to create a new URL for every individual action we want to do even if many of them alter the same resource. 

Ideally we try to be RESTful and have the URL point to the user's profile (not necessarily the same URL as their profile page, can be an API-specific endpoint like `/api/v1/user/{id}/`), and the POST arguments are the attribute names that we compare to the whitelist.

Even more ideally we use something like Django REST Framework to implement this API but that's way too out of scope for this PR to consider. But perhaps we can inch closer incrementally in the meantime?

pontoon/base/views.py
```diff
@@ -1224,20 +1228,19 @@ def save_to_transifex(request, template=None):
...
 
-    if request.method != 'POST':
-        log.error("Non-POST request")
-        raise Http404
+    if attribute not in ['quality_checks', 'force_suggestions']:
+        return HttpResponse('error')
 
-    profile = request.user.profile
-    profile.quality_checks = not profile.quality_checks
+    value = not getattr(profile, attribute)
```
Osmose: I know you're whitelisting these above but it's sooooooo scary.

pontoon/base/views.py
```diff
@@ -1224,20 +1228,19 @@ def save_to_transifex(request, template=None):
...
-    """Turn quality checks on/off for the current user."""
-    log.debug("Turn quality checks on/off for the current user.")
+def toggle_user_profile_attribute(request, attribute):
+    profile = request.user.profile
 
-    if request.method != 'POST':
-        log.error("Non-POST request")
-        raise Http404
+    if attribute not in ['quality_checks', 'force_suggestions']:
+        return HttpResponse('error')
```
Osmose: Probably want to return a `403` here instead of a `200`.

pontoon/base/static/js/translate.js
```diff
@@ -198,6 +198,21 @@ var Pontoon = (function (my) {
...
     /*
+     * Update current translation length and cache
+     *
+     * lengthOnly Do not update cache
+     */
+    updateCurrentLengthAndCache: function (lengthOnly) {
+      var translation = $('#translation').val();
+      $('#translation-length .current-length').html(translation.length);
+
+      if (!lengthOnly) {
```
Osmose: Better to have two methods than one method with a weird boolean.

pontoon/base/static/js/translate.js
```diff
@@ -198,6 +198,21 @@ var Pontoon = (function (my) {
 
 
     /*
+     * Update current translation length and cache
+     *
+     * lengthOnly Do not update cache
+     */
+    updateCurrentLengthAndCache: function (lengthOnly) {
```
Osmose: Might wanna include the word "translation" in the name so we know what cache and what length we're updating.

pontoon/base/static/js/translate.js
```diff
@@ -1916,7 +1938,8 @@ var Pontoon = (function (my) {
       this.user = {
         email: $('#server').data('email') || '',
         name: $('#server').data('name') || '',
-        localizer: isLocalizer($('#server').data('user-translated-locales')),
+        forceSuggestions: $('.force-suggestions').is('.enabled'),
+        translator: isTranslator($('#server').data('user-translated-locales')),
```
Osmose: Maybe rename this to `isTranslator` to be clear it's a flag?

pontoon/base/models.py
```diff
@@ -127,11 +127,8 @@ class UserProfile(models.Model):
     user = models.OneToOneField(User, related_name='profile')
 
     # Other fields here
-    transifex_username = models.CharField(max_length=40, blank=True)
```
Osmose: Yesssssssssssssss

pontoon/base/static/js/translate.js
mathjazz: That was my idea initially. But it didn't work, because we recreate Pontoon object data every time a new project is loaded. So if forceSuggestions setting has changed, we'd override it with whatever the value on page load was.

Now I refactored a bit to only recreate Pontoon object data that requires that.

mathjazz:
I find always showing both buttons overkill and confusing, because this is a feature used significantly less often than Save (or Copy and Clear).

I did consider moving quality check and force suggestions toggles somewhere to the editor, because that's where users need these features. But I'm not sure about the UI yet, so I left it as it is for now.

Anyhow, this is ready for another look.

pontoon/base/views.py
```diff
@@ -1224,20 +1229,21 @@ def save_to_transifex(request, template=None):
 
 
 @login_required(redirect_field_name='', login_url='/403')
+@require_POST
 @transaction.atomic
-def quality_checks_switch(request):
-    """Turn quality checks on/off for the current user."""
-    log.debug("Turn quality checks on/off for the current user.")
+def toggle_user_profile_attribute(request, email):
+    user = get_object_or_404(User, email=email)
```
Osmose: I'm pretty sure this means any user can change the `quality_checks` and `force_suggestions` preferences of any other user. You'll want to ensure that the current user has permission to edit this user (for now just checking that they're equal is fine, no need for full-blown permissions yet).

pontoon/base/urls.py
```diff
@@ -30,6 +30,11 @@
     url(r'^contributor/(?P<email>[\w.%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,4})/$',
         RedirectView.as_view(url="/contributors/%(email)s/", permanent=True)),
 
+    # API: Toogle user profile attribute
```
jotes: Do you have any specific design of api in mind? We could use e.g. Piston or something like that.

pontoon/base/views.py
jotes: I didn't catch this. You're right, we should use permission_required_or_403 and definetely test this view to avoid similar problems in future.

pontoon/base/static/js/main.js
```diff
@@ -525,23 +525,36 @@ $(function() {
...
     Pontoon.startLoader();
 
     $.ajax({
-      url: '/quality-checks-switch/',
+      url: '/api/v1/user/' + Pontoon.user.email + '/',
       type: 'POST',
       data: {
-        csrfmiddlewaretoken: $('#server').data('csrf')
+        csrfmiddlewaretoken: $('#server').data('csrf'),
+        attribute: $(this).data('attribute')
```
Osmose: I don't think it's a blocker, but typically RESTful APIs have you write the value you want persisted rather than toggling the value, so that you don't have to read and then write the preference's value in order to set it correctly. Other possibilities include editing the preference in a different tab de-syncing the translate tab's UI from what's stored in the backend.

I'm totally down with shipping this and dealing with that later, though.

pontoon/base/urls.py
Osmose: Django REST Framework is the main choices for Django APIs these days, and we should use it eventually. But for this PR it's not worth it.

Osmose:
Tested locally and it seems to work. Fix that security bug and I'm r+. Nice work!

pontoon/base/views.py
mathjazz: Ugh! That's a big one. Thanks for spotting!

pontoon/base/static/js/main.js
mathjazz: Fixed.

PR #266, opened 2015-11-12
From now on serial_task support locks for tasks with different parame…

jotes:
@mathjazz @Osmose could you look at this again?

### DIFF
--- a/pontoon/sync/core.py
+++ b/pontoon/sync/core.py
@@ -15,6 +15,7 @@ from pontoon.base.models import (
     Resource,
     update_stats
 )
+from pontoon.base.tasks import PontoonTask
 from pontoon.sync.changeset import ChangeSet
 from pontoon.sync.vcs_models import VCSProject
 from pontoon.sync.utils import locale_directory_path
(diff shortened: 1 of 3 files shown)


mathjazz:
I didn't test, but the code LGTM.

pontoon/sync/core.py
```diff
@@ -32,17 +33,19 @@ def sync_project(db_project, now):
...
 
-def serial_task(timeout, **celery_args):
+def serial_task(timeout, lock_key="", base=None, **celery_args):
     """
     Decorator ensures that there's only one running task with given task_name.
     Decorated tasks are bound tasks, meaning their first argument is always their Task instance
     :param timeout: time after which lock is released.
+    :param lock_key: allows to define different lock for respective parameters of task.
     """
+    task_class = base or PontoonTask
```
Osmose: Why add the base argument? You could leave it in `celery_args` and it would function identically.

pontoon/sync/core.py
```diff
@@ -32,17 +33,19 @@ def sync_project(db_project, now):
...
     def wrapper(func):
-        @shared_task(bind=True, **celery_args)
+        @shared_task(base=task_class, bind=True, **celery_args)
         @wraps(func)
         def wrapped_func(self, *args, **kwargs):
-            lock_name = "serial_task.{}".format(self.name)
+            lock_name = "serial_task.{}[{}]".format(self.name, lock_key.format(*args, self=self, **kwargs))
```
Osmose: Why pass self into format? I guess we _could_ use the task instance if we wanted but we don't.

pontoon/sync/core.py
```diff
@@ -32,17 +33,19 @@ def sync_project(db_project, now):
         changeset.execute()
 
 
-def serial_task(timeout, **celery_args):
+def serial_task(timeout, lock_key="", base=None, **celery_args):
```
Osmose: I really like the `lock_key` parameter as a solution. Nice work!

Osmose:
Code looks good, made a few nitpicky comments.

Briefly tested it locally and it didn't work, but I'm probably using an in-memory cache not shared between two projects. I'm going to try it with memcached or something and report what I find.

jotes:
@Osmose Pontoon is using LocMemcace by default. Switching default cache backend to something more persistent between processes should help.

jotes:
@Osmose thx for review, could you look at this again?

Osmose:
Enabled a shared cache and successfully tested this locally. Nice work!

Osmose:
r+, great fix!

PR #267, opened 2015-11-13
More accurate locale_directory_path detection

mathjazz:
Instead of returning the first directory with the name that matches
the locale code, we collect all directories with matching names and
return the first one that contains resources. If none of them contains
resources, we return the first one available (needed for asymmetric
formats).

@jotes r? @Osmose r?


mathjazz:
Aaaaaand all the comments are gone because I force pushed.

So yeah, we should totally add tests for this (we don't seem to have tests for this method at all). But I feel quite confident merging this (famous last words) after testing in on stage with all projects.

PR #268, opened 2015-11-14
Fix Bug 1224419: Show plural string to locales with a single plural form

mathjazz:
It includes variable identifier, which matters to some locales. Showing
both variants, singular and plural might confuse some users to translate
both variants.

@jotes @Osmose r?


jotes:
r+ :+1:

Osmose:
r+

I find it a little weird that the strings in the list of entities still have `1` instead of `%s`.

mathjazz:
Thanks for pointing it out! Updated, so now it uses the correct form everywhere, e.g. in the entity list, when copying source to translation or when requesting suggestions from machinery.

PR #269, opened 2015-11-14
Make single Transvision query

mathjazz:
Also link to Transvision search results instead of homepage

@Osmose @jotes r?


pontoon/base/views.py
```diff
@@ -1087,7 +1085,6 @@ def transvision(request, repo, title):
 
             return HttpResponse(json.dumps({
```
jotes: We could just pass contents of `r` here.

pontoon/base/views.py
```diff
@@ -1064,9 +1064,8 @@ def amagama(request):
         return HttpResponse("error")
 
 
-def transvision(request, repo, title):
+def transvision(request):
     """Get Mozilla translations from Transvision service."""
-    log.debug("Get Mozilla translations from Transvision service.")
```
jotes: Why do we need this view? Maybe there's a way to move this thing from backend to client-side (i'm not sure what kind of apis transvision provides).

pontoon/base/views.py
mathjazz: I considered doing it now that "title" is removed, but this format is consistent with other similar (Machinery) views.

pontoon/base/views.py
mathjazz: Some of the machinery services require HTTP instead of HTTPS and also API keys (none of which is true for Transvision). And to make things consistent, I've moved it to view as well.

jotes: r+

On Sat, Nov 14, 2015 at 2:07 PM, Matjaž Horvat <notifications@github.com>
wrote:

> In pontoon/base/views.py
> <https://github.com/mozilla/pontoon/pull/269#discussion_r44858316>:
>
> >      """Get Mozilla translations from Transvision service."""
> > -    log.debug("Get Mozilla translations from Transvision service.")
>
> Some of the machinery services require HTTP instead of HTTPS and also API
> keys (none of which is true for Transvision). And to make things
> consistent, I've moved it to view as well.
>
> —
> <https://github.com/mozilla/pontoon/pull/269/files#r44858316>.
>



-- 
jarek@reijutsu:~$ fortune
jarek@reijutsu:~$ fortune
You now have Asian Flu.

pontoon/base/views.py
```diff
@@ -1075,9 +1074,8 @@ def transvision(request, repo, title):
         log.error(str(e))
         return HttpResponse("error")
 
-    src = "en-US"
-    url = "https://transvision.mozfr.org/api/v1/tm/%s/%s/" \
-          "%s/%s/?max_results=%s&min_quality=70" % (repo, src, locale, text, 5)
+    url = "https://transvision.mozfr.org/api/v1/tm/global/en-US/" \
```
Osmose: `str.format` is preferred over `%` formatting. Now's a good time to update this.

Also, wrapping strings in parenthesis and relying on adjacent strings being concatenated is preffered over using `\` to escape the line break, as it's less fragile to changes.

pontoon/base/views.py
```diff
@@ -1075,9 +1074,8 @@ def transvision(request, repo, title):
...
+          "%s/%s/?max_results=%s&min_quality=70" % (locale, text, 5)
```
Osmose: Requests can handle the query string with the `params` argument.

pontoon/base/static/js/main.js
```diff
@@ -325,35 +325,35 @@ var Pontoon = (function (my) {
       }).error(error).complete(complete);
 
       // Transvision
-      $(['aurora', 'gaia', 'mozilla-org']).each(function(i, v) {
-        requests++;
+      requests++;
```
Osmose: SUCH DELICIOUS SIMPLIFICATION

pontoon/base/static/js/main.js
```diff
@@ -325,35 +325,35 @@ var Pontoon = (function (my) {
...
-                original: this.source,
-                quality: Math.round(this.quality) + '%',
-                url: 'http://transvision.mozfr.org/',
-                title: 'Visit Transvision',
-                source: data.title,
-                translation: this.target
-              });
+      }).success(function(data) {
+        if (data.translations && !data.translations.error) {
+          $.each(data.translations, function() {
```
Osmose: Consider native JS `forEach`?

pontoon/base/views.py
Osmose: I'm with @jotes on this one, this is just visual consistency, which isn't really all that helpful.

pontoon/base/views.py
Osmose: This totally works. A future improvement would be to abstract machinery services on the client side to some API that each services implements, and then have the Transvision one contact Transvision directly.

Osmose:
r+wc, with the caveat that I'm not familiar enough with Transvision to effectively test this. Nice work!

pontoon/base/views.py
mathjazz: Updated.

pontoon/base/views.py
mathjazz: Updated.

pontoon/base/views.py
mathjazz: Done.

PR #270, opened 2015-11-16
Warn anonymous users their translations aren't saved

mathjazz:
We don't save translations from anonymous users. So the notification shouldn't ready "Translation saved/updated".

@Osmose @jotes r?


Osmose:
r+

jotes:
r+

PR #271, opened 2015-11-16
Fix bug 1225021: Add Has suggestions filter

mathjazz:
Adding probably the most wanted missing filter, listing all entities that contain suggestions. Suggested filter only lists strings which don't have approved translations.

@Osmose @jotes r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -797,13 +797,18 @@ class Entity(DirtyFieldsMixin, models.Model):
 
         for entity in entities:
             translation_array = []
+            has_suggestions = False
 
             if entity.string_plural == "":
                 translation_array.append(entity.get_translation())
+                if len(entity.fetched_translations) > 1:
+                    has_suggestions = True
 
             else:
                 for plural_form in range(0, locale.nplurals or 1):
                     translation_array.append(entity.get_translation(plural_form))
+                    if len([t for t in entity.fetched_translations if t.plural_form == plural_form]) > 1:
+                        has_suggestions = True
 
             entities_array.append({
                 'pk': entity.pk,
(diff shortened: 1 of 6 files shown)


pontoon/base/static/js/translate.js
```diff
@@ -1084,11 +1091,27 @@ var Pontoon = (function (my) {
...
       var status = this.getEntityStatus(entity),
           translation = entity.translation[0].string;
+
       entity.ui.addClass(status);
       entity.ui.find('.translation-string')
         .html(this.doNotRender(translation || ''));
 
       this.updateProgress();
+
+      // Update has-suggestions status
```
mathjazz: This is epic, but also the only reliable way to get this information. We should probably bring all translations to the client with get-entities and then we don't need to make requests just to get this information.

It's a good enough solution UX wise, because we don't need this information instantly anyways.

pontoon/base/views.py
```diff
@@ -496,13 +496,10 @@ def get_translations_from_other_locales(request, template=None):
             json.dumps(payload, indent=4), content_type='application/json')
 
 
-def get_translation_history(request, template=None):
```
Osmose: djtfbhg

pontoon/base/models.py
```diff
@@ -797,13 +797,18 @@ def for_project_locale(self, project, locale, paths=None, search=None):
 
         for entity in entities:
             translation_array = []
+            has_suggestions = False
 
             if entity.string_plural == "":
                 translation_array.append(entity.get_translation())
+                if len(entity.fetched_translations) > 1:
```
Osmose: What if it has translations that aren't approved?

Osmose:
r+wc, works locally. Nice work!

PR #272, opened 2015-11-17
Bug 1208171: Log sync timing to the database and display it.

Osmose:
This is sort've in-progress but I think is sufficiently useful to merge now (plus I need to focus on different things for a while and thus won't really be able to improve this soon).

This adds some models for storing timing data about sync jobs in the database, and some views to view the data. The pages are public but not linked anywhere on the site, so we can use them and play around with them without bugging users. The data itself isn't sensitive so having the page public isn't an issue.

There are some known bugs that I haven't handled that I don't think are blockers:

- The big one, projects that are skipped due to a lack of changes have no timing data, which I think causes the entire run to not have an end time. This should be fixed with a followup PR soon, but I may not get to it for a while.
- If a job fails with an error, it is listed as "In-progress" forever.
- The views aren't smart at all about in-progress jobs, don't auto-update, etc.
- The changes aren't 100% well tested as I expect them to evolve a bunch before we settle into the logging that we want.
- Source repos are shown as having taken 0 seconds to sync.

Even with all this I think it will significantly improve our current method of verbose logging and human work to get sync times.

@mathjazz @jotes r?

### DIFF
--- a/pontoon/sync/urls.py
+++ b/pontoon/sync/urls.py
@@ -1,0 +1,13 @@
+from django.conf.urls import patterns, url
+
+from pontoon.sync import views
+
+
+urlpatterns = patterns(
+    '',
+
+    url(r'^sync/log/$', views.sync_log_list, name='pontoon.sync.logs.list'),
+
+    url(r'^sync/log/(?P<sync_log_pk>\d+)/', views.sync_log_details,
+        name='pontoon.sync.logs.details'),
+)
--- a/docs/admin/deployment.rst
+++ b/docs/admin/deployment.rst
@@ -214,6 +214,15 @@ It's recommended to run this job once an hour. It commits any string changes in
 the database to the remote VCS servers associated with each project, and pulls
 down the latest changes to keep the database in sync.
 
+Sync Log Retention
+~~~~~~~~~~~~~~~~~~
+You may also optionally run the ``clear_old_sync_logs`` management command on a
+schedule to remove sync logs from the database that are over 90 days old:
+
+.. code-block:: bash
+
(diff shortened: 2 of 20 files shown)


pontoon/sync/urls.py
```diff
@@ -0,0 +1,13 @@
...
+
+from pontoon.sync import views
+
+
+urlpatterns = patterns(
+    '',
+
+    url(r'^sync/log/$', views.sync_log_list, name='pontoon.sync.logs.list'),
+
+    url(r'^sync/log/(?P<sync_log_pk>\d+)/', views.sync_log_details,
```
mathjazz: Start timestamp would be more useful than id here.

pontoon/sync/templatetags/helpers.py
```diff
@@ -0,0 +1,17 @@
...
+        return 'Unknown'
+    elif len(projects) == 1:
+        return projects[0].name
+    elif len(projects) < 4:
+        return ', '.join(p.name for p in projects)
+    else:
+        return '{names} and {total} more'.format(
+            names=', '.join(p.name for p in projects[:3]),
+            total=len(projects) - 3
+        )
```
mathjazz: We'll be syncing all projects most of the time, so "All projects" should be added here.

mathjazz:
This is a very exciting, r+!

Looking forward to new iterations when we find time for them.

pontoon/sync/urls.py
Osmose: Theoretically timestamps aren't unique, though. I doubt anyone is going to construct a URL like that anyway.

PR #278, opened 2015-11-17
Base view cleanup

mathjazz:
* Removed template as attribute
* Moved data dict into render() params
* Removed log.debug(), which was used nowhere
* Use HttpResponseBadRequest if request not AJAX
* Use @require_POST decorator to detect non-POST requests
* Removed traces of Transifex

@Osmose @jotes r?

### DIFF
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -1,5 +1,3 @@
-import base64
-import hashlib
 import json
 import Levenshtein
 import logging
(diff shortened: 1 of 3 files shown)


pontoon/base/views.py
```diff
@@ -392,21 +349,16 @@ def get_project_details(request, slug):
...
 
-def entities(request, template=None):
+def entities(request):
     """Get entities for the specified project, locale and paths."""
-    log.debug("Get entities for the specified project, locale and paths.")
-
     if not request.is_ajax():
-        log.error("Non-AJAX request")
-        raise Http404
+        return HttpResponseBadRequest('Bad Request: Request must be AJAX')
```
Osmose: I SMELL A DECORATOR

Osmose:
r+wc, excellent work!

pontoon/base/views.py
mathjazz: Of course!

PR #280, opened 2015-11-18
Bug 1219212: Add ability to download translated resources

mathjazz:
This is first part of Bug 1219212: re-enabling resource download. When translating a resource, you can now download it from the profile menu, by clicking on the "Download FORMAT" link, where FORMAT is PO or DTD etc.

Since we don't have storage, this is how it works:
1) Pull source repository.
2) Get source (en-US) file - works for all formats.
3) Update file from DB.
4) Pass it to the user for download.

I left the code in the view, because I found myself in the import loop when moving it to utils.py. When we agree where should we put it and if the basic direction of this PR is correct, I should write some tests for get_download_content().

I tried to reuse as much sync code as possible. I also had to modify it a bit, but nothing serious. Given that all tests pass, I think we should be safe. I noticed VCSTranslation.update_from_db() is pretty slow and takes close to 20 seconds combined for the biggest file, but that's 2719 strings.

@Osmose @jotes r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -21,7 +21,6 @@ from jsonfield import JSONField
 
 from pontoon.administration.vcs import commit_to_vcs, get_revision, update_from_vcs
 from pontoon.base import utils
-from pontoon.base.utils import get_object_or_none
 from pontoon.sync import KEY_SEPARATOR
 
 
@@ -444,7 +443,7 @@ class ProjectLocale(models.Model):
     @classmethod
     def get_latest_activity(cls, project, locale):
         """Get the latest activity within this project and locale."""
-        project_locale = get_object_or_none(ProjectLocale, project=project, locale=locale)
+        project_locale = utils.get_object_or_none(ProjectLocale, project=project, locale=locale)
         if project_locale is not None and project_locale.latest_translation is not None:
             return project_locale.latest_translation.latest_activity
         else:
@@ -476,6 +475,12 @@ class Repository(models.Model):
         choices=TYPE_CHOICES
     )
     url = models.CharField("URL", max_length=2000, blank=True)
+
+    """
+    Prefix of the resource URL, used for direct downloads. To form a full
+    URL, relative path must be appended.
+    """
+    permalink_prefix = models.CharField("Permalink prefix", max_length=2000, blank=True)
 
     """
     Mapping of locale codes to VCS revisions of each repo at the last
@@ -907,7 +912,7 @@ class Translation(DirtyFieldsMixin, models.Model):
             self.check_latest_translation(self.locale)
             self.check_latest_translation(stats)
 
-            project_locale = get_object_or_none(
+            project_locale = utils.get_object_or_none(
                 ProjectLocale,
                 project=self.entity.resource.project,
                 locale=self.locale
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -3,63 +3,30 @@ var Pontoon = (function (my) {
   return $.extend(true, my, {
 
     /*
-     * Save data in different formats
-     *
-     * type Data format
-     * value Data
+     * Download resource file. Not possible with AJAX.
      */
-    save: function (type, value) {
+    download: function () {
+      var params = {
+          slug: this.project.slug,
+          code: this.locale.code,
+          path: this.part,
+          csrfmiddlewaretoken: $('#server').data('csrf')
+      };
 
-      var self = this,
-          params = {
-            type: type,
-            locale: this.locale.code,
-            project: this.project.slug
-          };
+      var post = $('<form>', {
+        method: 'POST',
+        action: '/download/'
+      });
 
-      function strip(entities) {
-        // Deep copy: http://api.jquery.com/jQuery.extend
-        var e = $.extend(true, [], entities);
-        $(e).each(function () {
-          delete this.ui;
-          delete this.hover;
-          delete this.unhover;
-          delete this.id;
-          delete this.pk;
-          delete this.body;
-        });
-        return e;
+      for (var key in params) {
+        $('<input>', {
+          type: 'hidden',
+          name: key,
+          value: params[key]
+        }).appendTo(post);
       }
 
-      // It is impossible to download files with AJAX
-      function download(params) {
-        params.csrfmiddlewaretoken = $('#server').data('csrf');
-        var post = $('<form>', {
-          method: 'post',
-          action: '/download/'
-        });
-        for(var key in params) {
-          $('<input>', {
-            type: 'hidden',
-            name: key,
-            value: params[key]
-          }).appendTo(post);
-        }
-        post.appendTo('body').submit().remove();
-      }
-
-      if (type === "html") {
-        params.content = value;
-        download(params);
-
-      } else if (type === "json") {
-        var entities = strip(this.entities);
-        params.content = JSON.stringify(entities, null, "\t");
-        download(params);
-
-      } else if (type === "zip") {
-        download(params);
-      }
+      post.appendTo('body').submit().remove();
     },
 
 
@@ -1569,6 +1536,9 @@ var Pontoon = (function (my) {
               self.endLoader('Oops, something went wrong.', 'error');
             });
 
+          } else if ($(this).is('.download')) {
+            self.download();
+
           } else if ($(this).is(".hotkeys")) {
             $('#hotkeys').show();
 
@@ -1576,17 +1546,6 @@ var Pontoon = (function (my) {
             e.stopPropagation();
           }
         });
-      });
-
-      // Download menu
-      $('#profile .menu .file-format').click(function (e) {
-        e.preventDefault();
-
-        if ($(this).is(".html")) {
-          self.postMessage("HTML");
-        } else {
-          self.save($(this).attr('class').split(" ")[1]);
-        }
       });
 
       // Close notification on click
@@ -1651,6 +1610,7 @@ var Pontoon = (function (my) {
      */
     updateProfileMenu: function () {
       $('#profile .admin-current-project a').attr('href', '/admin/projects/' + this.project.slug + '/');
+      $('#profile .download .file-format').html(this.entities[0].format.toUpperCase());
     },
 
 
@@ -1878,10 +1838,6 @@ var Pontoon = (function (my) {
         case "DELETE":
           var entity = Pontoon.entities[message.value];
           Pontoon.updateEntityUI(entity);
-          break;
-
-        case "HTML":
-          Pontoon.save("html", message.value);
           break;
 
         }
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -2,6 +2,7 @@ import json
 import Levenshtein
 import logging
 import math
+import os
 import requests
 import xml.etree.ElementTree as ET
 import urllib
@@ -32,7 +33,7 @@ from django.views.generic import TemplateView
 from django.views.generic.detail import DetailView
 from guardian.decorators import permission_required_or_403
 from operator import itemgetter
-from pontoon.administration import files
+
 from pontoon.base import utils
 from pontoon.base.utils import require_AJAX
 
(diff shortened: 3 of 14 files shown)


pontoon/base/views.py
```diff
@@ -988,50 +993,75 @@ def transvision(request):
         return HttpResponse("error")
 
 
+def get_download_content(slug, code, relative_path):
```
mathjazz: Translation table is updated often - every time a translation is added, deleted, approved or unfuzzied. Either by a user or by the sync job.

pontoon/base/static/js/translate.js
```diff
@@ -1649,6 +1650,10 @@ var Pontoon = (function (my) {
      */
     updateProfileMenu: function () {
       $('#profile .admin-current-project a').attr('href', '/admin/projects/' + this.project.slug + '/');
+
+      $('#profile .download .resource')
+        .html(this.entities[0].format.toUpperCase())
+        .toggle(!this.project.win);
```
Osmose: What's `project.win`?

pontoon/base/views.py
```diff
@@ -988,50 +993,75 @@ def transvision(request):
...
+    """
+    Get content for the file to be downloaded.
+
+    :param str slug: Project slug.
+    :param str code: Locale code.
+    :param str relative_path: Resource path.
+    """
+    # Clone or update project repository
+    project = get_object_or_404(Project, slug=slug)
+    project.repositories.first().pull()
```
Osmose: This doesn't work if the source files and the files from the locale we want aren't in the first repo. Technically we have to pull from all the repos for the project in order to be sure we have both the source strings and the locale strings.

pontoon/base/views.py
```diff
@@ -988,50 +993,75 @@ def transvision(request):
...
+
+    # Get source file path
+    locale = get_object_or_404(Locale, code__iexact=code)
```
Osmose: This should be above the point where we pull files, as we don't want to trigger a pull only to abort with a 404.

pontoon/base/static/js/translate.js
mathjazz: A reference to the iframe window where in-place localization takes place. We use this to check if the website supports in-place localization. Because if it does, it might be linked to several resources, not just one, so instead of downloading the resource, we offer you to download HTML. Which is probably not very useful. In the future, we should probably allow you to download a ZIP of all resources (but it will take time to generate for e.g Firefox).

pontoon/base/views.py
mathjazz: We only need the source file (en-US), which we then update from the DB.

pontoon/base/views.py
mathjazz: True that.

pontoon/base/views.py
```diff
@@ -988,50 +993,75 @@ def transvision(request):
...
+    vcs_project = VCSProject(project, locales=[locale])
+
+    source_path = relative_path
+    if relative_path.endswith('.po'):
```
Osmose: I wish we had a single common place to put this so that pot files aren't a special case everywhere.

pontoon/base/views.py
```diff
@@ -988,50 +993,75 @@ def transvision(request):
...
+        source_path = relative_path + 't'
+    full_path = os.path.join(vcs_project.source_directory_path(), source_path)
```
Osmose: This is very scary because you're taking a path submitted by the user and loading it up, opening us up to things like users submitting paths like `../../../../../../../../../../../../etc/passwd`. Granted, we're attempting to parse the given file as some sort've resource file, so that particular example would fail, but whenever taking user input as a path you always have to ensure that the path is within some "safe" directory that you're limiting them to. In this case, we want to limit them to the checkout of the project in question.

pontoon/base/views.py
```diff
@@ -988,50 +993,75 @@ def transvision(request):
...
+        entity__resource__project=project,
+        entity__resource__path=relative_path,
+        entity__obsolete=False,
+        approved=True,
+        locale=locale
+    )
+    for vcs_translation in resource_file.translations:
+        db_translations = translations.filter(entity__key=vcs_translation.key)
+        vcs_translation.update_from_db(db_translations)
+    resource_file.save(locale)
```
Osmose: There's got to be a way we can serve this up without saving it back to the filesystem. We've got all the data in memory, and we're writing and then reading it again.

pontoon/base/views.py
Osmose: We also cannot guarantee that the source files are in the first repo. It just so happens that our prod database and all the projects we're currently handling have that, but it's not guaranteed by the code or how Repositories work.

Osmose:
I did some light load testing on staging as well as on my local machine. My local dev server went nuts and staging started throwing exceptions and 503 Service Unavailable errors once response times went up to 30 seconds.

I tested using siege set to simulate 50 users hitting the endpoint with an up-to-10-second delay between each repeated attempt by a user. The staging errors hit Raygun: https://app.raygun.io/crashreporting/cqxgfm/errors/957793967

Basically, two things are happening here:

1. The media directory is shared between requests since this is all hitting a pool of 1 or 2 web dynos, so concurrent processes are trying to HG pull and read/write the same files at the same time, causing conflicts. This is a direct symptom of us pulling from VCS.

2. The VCS and filesystem actions are taking long enough to cause the request queue to get severely backed up. This is more symptom of us generating these files in response to all user requests vs generating the file and saving it somewhere so static hosting can serve it up.

Granted, Pontoon _probably_ won't get that kind of traffic. But it makes me uncomfortable with this feature as it's currently designed.

I won't say r- to the approach yet, and I'll leave the final say of it to you, but I'd rather admit that we didn't anticipate some use cases of Verbatim and fail at killing Verbatim before Mozlando than try and push this through.

mathjazz:
Ad 1.
Instead of pulling, we could generate files, but with limited success only, because we don't keep all the necessary information in the DB. The only right solution to the constant pull problem is persistent storage. Cloning all repos once a new dyno is created would also help a bit. We should be able to avoid the problem with multiple processes reading/writing the same file by using temporary file for each request.

Ad 2.
We can only serve static files with their current repository state, which is suboptimal. What we should do is be prepared for requests that take more than 30 seconds.

Osmose:
> Instead of pulling, we could generate files, but with limited success only, because we don't keep all the necessary information in the DB.

What information are we missing? The two things I can think of are comments not tied to an entity, and resource metadata.

> The only right solution to the constant pull problem is persistent storage.

Persistent storage might help in some areas, but I disagree that it's the only right solution. If it's shared, you get all the problems related to NFS (IE reading a single file can time out and reading files is even slower).

> Cloning all repos once a new dyno is created would also help a bit.

Cloning all repos on dyno boot is an interesting idea. I'd worry about increasing dyno startup time significantly but that depends on when we actually do the pulls and how we do them at dyno startup.

> We should be able to avoid the problem with multiple processes reading/writing the same file by using temporary file for each request.

A bunch of questions come to mind when considering this:

- So we'd start out the request by copying the resource file to a temporary file? Would we then not do a pull at the start of the request so HG doesn't conflict? 
- Do the workers share the storage with the webheads? 
- If so, what happens when there's a sync going on at the same time?

> We can only serve static files with their current repository state, which is suboptimal. What we should do is be prepared for requests that take more than 30 seconds.

This is tricky; Heroku has a hard [30 second window](https://devcenter.heroku.com/articles/request-timeout#long-polling-and-streaming-responses) for returning at least one byte back to the client, and a rolling 55-second window for every byte afterwards. We'd have to run our VCS commands asynchronously and loop to check them, while still sending some sort of data back to the client telling them to hold on until we can get back to them.

Alternatively, we schedule generation of these files and return a placeholder to the user, and make the UI poll the server until the returned file is ready. This is what services that have long-running tasks all end up doing, because it is more or less universally true that web requests that last longer than 30 seconds (barring streaming responses) are unmaintainable.

mathjazz:
I've made two rather important changes:

1) Instead of the original (source) file, we use the locale file as the basis. Then, we use ChangedEntityLocale class to only update relevant entities. This reduces the number of DB queries significantly and is the fastest possible way of updating the file. In vast majority of cases no modifications of the file will be needed.

2) That means pulling repositories is the slowest operation now. Since we don't have persistent storage, we should clone all repositories after the dyno is started (currently takes 9 minutes in production, with HG and SVN being much slower than git(hub)). It's a simple command, but I don't know yet where to put it. ;) In the meantime, I'm also playing with the idea of downloading resources instead of pulling the whole repos.

mathjazz:
I've added a field to each repository to be used as a template for links to downloadable files. Using this field, we should be able to identify direct links to all original and translated resources.

The template looks like this:
https://raw.githubusercontent.com/mozilla/zamboni/master/locale/{locale_code}/

We replace the optional "{locale_code}" string with (surprise!) a locale code or SOURCE_DIR_NAME and append relative path to get the full link, e.g.:
https://raw.githubusercontent.com/mozilla/zamboni/master/locale/templates/LC_MESSAGES/javascript.pot

That way we can simply download a translated resource (and the original one for asymmetric formats) and then update it from DB.

Osmose:
> That way we can simply download a translated resource (and the original one for asymmetric formats) and then update it from DB.

This is a nice improvement over the original, nice work!

pontoon/base/models.py
```diff
@@ -579,7 +585,7 @@ def url_for_path(self, path):
 
         raise ValueError('No repo found for path: {0}'.format(path))
 
-    def pull(self):
+    def pull(self, locale=None):
```
Osmose: Might as well just make this `locales` and pass a list, eh?

pontoon/base/views.py
```diff
@@ -988,50 +994,119 @@ def transvision(request):
         return HttpResponse("error")
 
 
+def download_file(prefixes, dirnames, relative_path):
+    for prefix in prefixes:
+        for dirname in dirnames:
+            url = os.path.join(prefix.replace('{locale_code}', dirname), relative_path)
```
Osmose: - Don't use `os.path.join` for URLs. `'/'.join` is more appropriate.
- Why replace instead of `string.format`?

pontoon/base/views.py
```diff
@@ -988,50 +994,119 @@ def transvision(request):
...
+            r = requests.get(url, stream=True)
+            if not r.ok:
```
Osmose: Today I learned about `response.ok`.

pontoon/base/views.py
```diff
@@ -988,50 +994,119 @@ def transvision(request):
...
+                continue
+
+            filename = str(uuid.uuid4()) + os.path.splitext(url)[1]
```
Osmose: For temporary files, you want to use the [tempfile](https://docs.python.org/2/library/tempfile.html) module.

Osmose: Although I still think StringIO is the way to go rather than writing to the disk.

pontoon/base/views.py
```diff
@@ -988,50 +994,119 @@ def transvision(request):
...
+    # Get locale file
+    locale_prefixes = (
+        project.repositories.filter(permalink_prefix__contains='{locale_code}')
+        .values_list('permalink_prefix', flat=True)
+        .distinct()
+    )
+    dirnames = set([locale.code, locale.code.replace('-', '_')])
+    locale_path = download_file(locale_prefixes, dirnames, relative_path)
+    if not locale_path:
+        return False
```
Osmose: `None` is semantically more correct than `False` as a return value for this function.

pontoon/base/views.py
```diff
@@ -988,50 +994,119 @@ def transvision(request):
...
+        dirnames = VCSProject.SOURCE_DIR_NAMES
+        source_path = download_file(source_prefixes, dirnames, relative_path)
+        if not source_path:
+            return False
+
+    # Update file from database
+    resource_file = formats.parse(locale_path, source_path)
+    entities = {}
+
+    for e in Entity.objects.filter(
```
Osmose: I like to store complex querysets that I want to iterate over in a variable so that the for loop is easier to read.

pontoon/base/views.py
```diff
@@ -988,50 +994,119 @@ def transvision(request):
...
+        if key in entities:
+            vcs_translation.update_from_db(entities[key])
+
+    resource_file.save(locale)
+
+    # Return download content and remove temporary files
+    with codecs.open(locale_path, 'r', 'utf-8') as f:
+        content = f.read()
+
+        os.remove(locale_path)
```
Osmose: Dedent this, I have no idea what happens when you delete a file with an open file handle.

pontoon/sync/utils.py
```diff
@@ -60,3 +60,13 @@ def locale_directory_path(checkout_path, locale_code):
 
     raise IOError('Directory for locale `{0}` not found'.format(
                   locale_code or 'source'))
+
+
+def relative_source_path(path):
```
Osmose: Neato!

Osmose:
r+wc, haven't really tested it. Ping me if you want more attention on this. Nice work!

pontoon/base/views.py
mathjazz: The reason I didn't use tempfile is because it's gone after you close it (at the end of with clause).

I'll see how risky modifying the parsers to accept content would be.

pontoon/base/views.py
mathjazz: The problem with '/'.join is it doesn't take care of duplicate slashes, e.g.:
`'/'.join(['http://example.com/', 'file/path.txt']) => 'http://example.com//file/path.txt'`

mathjazz:
Thanks for the initial "almost r-", I feel much more confident about this now.

pontoon/base/views.py
Osmose: True! Interestingly, browsers seem to not care about duplicate slashes. I wonder if requests does?

I do wish Python had a urljoin method built in. _sigh_

pontoon/base/views.py
Osmose: tempfile.mkstemp will give you a file that doesn't disappear that you can delete yourself at the end.

jotes: System removes all files from /tmp/ during the boot.

mathjazz: If I have to manually remove them, is there really a difference between using a tempfile and manually creating a file with a unique name?

Osmose: Yes, because

1. They're guaranteed to be made in a temporary directory that won't affect other programs incidentally.
2. As jotes mentioned, if your code fails to delete the file due to an error, it will be removed later by the OS (at least on Linux and OSX).
3. tempfile makes some guarantees about no race conditions during creation that are desireable.

But really to me, it's that, when given two roughly equivalent options with the same outcome, choose the semantically correct one. It's more correct to use the stdlib's functions for making temporary files rather than using other functions that accomplish the same thing but aren't for making temporary files.

mathjazz: Makes sense. Updated #281.

PR #281, opened 2015-11-19
Bug 1219212: Add ability to UPLOAD translated resources

mathjazz:
This is based on #280, so look at 6853f0c only.

It adds the ability to upload translations for the currently translated resource (disabled on search and for non-translators).

To update translations in the DB, I reused the sync code. There shouldn't be any problems it upload is performed during sync, because it's same as if translations were manually submitted.

TODO (can wait until next release):
- I can't wait to get my hands dirty with drag & drop upload and fancy progress bars.
- Enable upload to anyone and add suggestions if user not translator or forces suggestions.

@Osmose @jotes r?

### DIFF
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -8,8 +8,8 @@ import xml.etree.ElementTree as ET
 import urllib
 
 from dateutil.relativedelta import relativedelta
-
 from django.conf import settings
+from django.contrib import messages
 from django.contrib.auth.decorators import login_required
 from django.contrib.auth.models import User
 from django.core.exceptions import ValidationError, ImproperlyConfigured
@@ -34,6 +34,7 @@ from django.views.generic.detail import DetailView
 from guardian.decorators import permission_required_or_403
 from operator import itemgetter
 
+from pontoon.base import forms
 from pontoon.base import utils
 from pontoon.base.utils import require_AJAX
 
@@ -193,8 +194,8 @@ def locale_project(request, locale, slug):
 
 def translate(request, locale, slug, part):
     """Translate view."""
-    l = get_object_or_404(Locale, code__iexact=locale)
-    p = get_object_or_404(
+    locale = get_object_or_404(Locale, code__iexact=locale)
+    project = get_object_or_404(
         Project.objects.distinct(),
         slug=slug,
         disabled=False,
@@ -202,7 +203,7 @@ def translate(request, locale, slug, part):
     )
 
     # Check if user authenticated
-    if not request.user.is_authenticated() and not p.pk == 1:
+    if not request.user.is_authenticated() and not project.pk == 1:
         return render(request, '403.html', status=403)
 
     projects = (
@@ -216,10 +217,12 @@ def translate(request, locale, slug, part):
 
     return render(request, 'translate.html', {
         'accept_language': utils.get_project_locale_from_request(request, Locale.objects),
-        'locale': l,
+        'download_form': forms.DownloadFileForm(),
+        'upload_form': forms.UploadFileForm(),
+        'locale': locale,
         'locales': Locale.objects.all(),
         'part': part,
-        'project': p,
+        'project': project,
         'projects': projects,
     })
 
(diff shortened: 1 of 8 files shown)


pontoon/base/views.py
```diff
@@ -201,9 +255,19 @@ def translate(request, locale, slug, part):
     )
 
     # Check if user authenticated
-    if not request.user.is_authenticated() and not p.pk == 1:
+    if not request.user.is_authenticated() and not project.pk == 1:
```
jotes: I'm not sure if it's the part of this PR but i think we add new field to model which would be used instead of hardcoding id == 1.

pontoon/base/views.py
```diff
@@ -190,10 +200,54 @@ def locale_project(request, locale, slug):
...
+
+    changeset = ChangeSet(
+        project,
+        VCSProject(project, locales=[locale]),
+        timezone.now()
+    )
+    changeset.locales = {locale.code: locale}
+    for vcs_translation in resource_file.translations:
+        key = vcs_translation.key
+        if key in entities:
```
jotes: entities.iteritems()?

pontoon/base/views.py
```diff
@@ -190,10 +200,54 @@ def locale_project(request, locale, slug):
...
+        obsolete=False
+    )}
+
+    changeset = ChangeSet(
+        project,
+        VCSProject(project, locales=[locale]),
+        timezone.now()
+    )
+    changeset.locales = {locale.code: locale}
+    for vcs_translation in resource_file.translations:
```
jotes: Maybe i'm not seeing it but how do you plan too resolve conflicts? Every user can override every translation in database?

pontoon/sync/changeset.py
```diff
@@ -176,6 +158,51 @@ def execute_create_db(self):
...
+                    approved=not vcs_translation.fuzzy,
+                    approved_date=self.now if not vcs_translation.fuzzy else None,
+                    fuzzy=vcs_translation.fuzzy,
+                    extra=vcs_translation.extra
+                ))
+
+        # Any existing translations that were not approved get unapproved.
+        for translation in db_translations.filter(approved_date__lte=self.now):
+            if translation not in approved_translations:
+                translation.approved = False
```
jotes: maybe it's good material for method called "unapprove()".

pontoon/base/views.py
mathjazz: Yeah, this is ugly. One day. ;)

pontoon/sync/changeset.py
mathjazz: I'd rather not change this method as part of this PR.

pontoon/base/views.py
mathjazz: So this feature is only enabled for translators. And we handle conflicts in pontoon/sync/changeset.py: update_entity_translations_from_vcs.

mathjazz:
Made a few updates:
* Using django form also for downloads, not just for uploads.
* Using separate view for upload, no longer abusing translate view.
* Adding file size and type validation.
* Allowing download to everyone, upload only to translators.
* Using temp file for storing uploaded contents.
* Modifying changeset API.

This time I'm only asking @jotes for r, while @Osmose is busy with his turkey.

mathjazz:
Three more changes:
* Using tempfile finally!
* Giving attributions to the uplaoder.
* And most importantly - reducing the number of DB queries. Uploading a file with almost 3000 strings only takes a few seconds locally. I think we'll have to do something similar with sync soon and then our memory issues will be gone hopefully. Which is also why I didn't bother changing the sync API for now.

pontoon/base/views.py
```diff
@@ -1017,6 +1024,35 @@ def download(request):
...
+        return HttpResponseForbidden("Forbidden: You don't have permission to upload files")
+
+    form = forms.UploadFileForm(request.POST, request.FILES)
+
+    if form.is_valid():
+        f = request.FILES['uploadfile']
+        utils.handle_upload_content(slug, code, part, f, request.user)
+        messages.success(request, 'Translations updated from uploaded file.')
+    else:
+        text = form.errors.as_data().itervalues().next()[0][0]
```
mathjazz: What is the non-stupid way of reading these error messages?

jotes: I would change it into something like: ```form.errors.as_data()['fieldname']``` or ```text = '<br>'.join([ "{}: {}".format(k,v)" for k, v in form.errors.as_data().iteritems())```

mathjazz: So `form.errors.as_data()` gives me:
`{'__all__': [ValidationError([u'Upload failed. Use .po.'])]}`

And if I do `for k, v in form.errors.as_data().iteritems()`, I'm only interested in `v[0][0][0]`. Which also sucks.

Osmose: What about:

```python
for field, errors in form.errors:
    for error in errors:
        messages.error(request, error)
```

Osmose: Also, as a side note, `iteritems` was removed in Python 3 since `items` in Python 3 does the same or better, and in most cases we don't need the savings from using an iterable. I tend to use `items` to make eventual Python 3 porting easier.

mathjazz:
Merging this as per IRC.

PR #282, opened 2015-11-20
Fix bug 1225087: Add user email address into Reply-To instead of From

mathjazz:
@Osmose @jotes r?

No more email addres spoofing.

### DIFF
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -11,8 +11,8 @@ from dateutil.relativedelta import relativedelta
 from django.conf import settings
 from django.contrib.auth.decorators import login_required
 from django.contrib.auth.models import User
-from django.core.exceptions import ValidationError
-from django.core.mail import send_mail
+from django.core.exceptions import ValidationError, ImproperlyConfigured
+from django.core.mail import EmailMessage
 from django.core.validators import validate_comma_separated_integer_list
 from django.db import transaction
 from django.db.models import Count, F, Q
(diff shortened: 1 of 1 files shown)


pontoon/base/views.py
```diff
@@ -1105,11 +1105,17 @@ def request_locale(request):
     subject = '[Pontoon] Locale Request'
     message = 'Add locale %s (%s) to project %s (%s).' % (
         locale.name, locale.code, project.name, project.slug)
-    sender = request.user.email
 
     if settings.ADMINS[0][1] != '':
-        recipients = [settings.ADMINS[0][1]]
-        send_mail(subject, message, sender, recipients)
+        sender = 'pontoon@mozilla.com'
```
Osmose: I like the use of variables for clarity, but I bet we could just pass these as kwargs too. Nice, though!

pontoon/base/views.py
```diff
@@ -1105,11 +1105,17 @@ def request_locale(request):
...
+        recipient = settings.ADMINS[0][1]
```
Osmose: While we're here, maybe have a default or raise an appropriate error due to lack of ADMINs? Not a blocker, just a thought.

Osmose:
r+, nice fix!

jotes:
:+1: r+

pontoon/base/views.py
```diff
@@ -1090,29 +1090,18 @@ def request_locale(request):
...
-        return HttpResponse("error")
-
-    subject = '[Pontoon] Locale Request'
-    message = 'Add locale %s (%s) to project %s (%s).' % (
-        locale.name, locale.code, project.name, project.slug)
-    sender = request.user.email
+    project = get_object_or_404(Project, slug=project)
+    locale = get_object_or_404(Locale, code__iexact=locale)
 
     if settings.ADMINS[0][1] != '':
```
Osmose: `ADMINS` defaults to an empty tuple, you'll want to defend against that here.

Osmose:
r+ again :D

pontoon/base/views.py
mathjazz: It defaults to (('', ''),), so we should be fine?

Osmose: Documentation says an empty tuple: https://docs.djangoproject.com/en/1.8/ref/settings/#admins

But we do override it in `base.py`. Huh. Do whatever you think is best, then. I figure being defensive about it is safer but it's not a big deal either way.

PR #285, opened 2015-11-23
Only mark approved strings as changed

mathjazz:
When saving or deleting translations, only mark entity as changed if
approved translation was added or deleted.

@Osmose @jotes r?


jotes:
r+

Osmose:
r+!

PR #287, opened 2015-11-24
Django security update

mathjazz:
Upgrade to the just released security release:
https://www.djangoproject.com/weblog/2015/nov/24/security-releases-issued/

Didn't notice anything unusual.

@Osmose @jotes r?


jotes:
After fast grep, it looks like we aren't affected :dancers: r+

Osmose:
r+

PR #288, opened 2015-11-28
Optimize suggestions from other locales

mathjazz:
@Osmose @jotes r?

### DIFF
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -396,44 +396,21 @@ def get_translations_from_other_locales(request):
         entity = request.GET['entity']
         locale = request.GET['locale']
     except MultiValueDictKeyError as e:
-        log.error(str(e))
-        return HttpResponse("error")
+        return HttpResponseBadRequest('Bad Request: {error}'.format(error=e))
 
-    try:
-        entity = Entity.objects.get(pk=entity)
-    except Entity.DoesNotExist as e:
-        log.error(str(e))
-        return HttpResponse("error")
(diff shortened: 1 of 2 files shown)


pontoon/base/views.py
```diff
@@ -396,44 +396,28 @@ def get_translations_from_other_locales(request):
...
-    else:
-        return HttpResponse(
-            json.dumps(payload, indent=4), content_type='application/json')
+    payload = [{
+        'locale': {
+            'code': t.locale.code,
+            'name': t.locale.name
+        },
+        'translation': t.string
+    } for t in translations]
```
jotes: we could change this into values() at some pont.

mathjazz: Yeah, I tried that, but it's not possible to serialize values() as json directly.

mathjazz: Thanks for the list() tip on IRC! This is much cleaner now.

pontoon/base/views.py
```diff
@@ -396,44 +396,21 @@ def get_translations_from_other_locales(request):
...
-                "translation": translation.string
-            })
-
-    if len(payload) == 0:
-        return HttpResponse("error")
-    else:
-        return HttpResponse(
-            json.dumps(payload, indent=4), content_type='application/json')
+    payload = list(translations.values('locale__code', 'locale__name', 'string'))
+    return HttpResponse(json.dumps(payload, indent=4), content_type='application/json')
```
Osmose: The indent seems unnecessary here.

Osmose:
Nice fix!

PR #289, opened 2015-11-28
Refactored some views and creates generic generator.

jotes:
@mathjazz As we talked on irc, i've refactored some of patterns we use in our code into decorator. However, some things like a handling of errors would require discussion and review.

### DIFF
--- a/pontoon/base/tests/test_utils.py
+++ b/pontoon/base/tests/test_utils.py
@@ -1,8 +1,23 @@
-from django_nose.tools import assert_equal, assert_false, assert_is_none, assert_true
+import json
+
+from django_nose.tools import (assert_equal,
+                               assert_false,
+                               assert_is_none,
+                               assert_true)
+
+from nose.tools import assert_raises
+
+from django.core.exceptions import ObjectDoesNotExist
+from django.db import transaction
+from django.http.response import HttpResponse, HttpResponseBadRequest, HttpResponseGone
+from django.http import QueryDict
+from django.views.decorators.http import require_POST
+
+from mock import MagicMock
 
 from pontoon.base.tests import ProjectFactory, TestCase
 from pontoon.base.models import Project
-from pontoon.base.utils import extension_in, get_object_or_none
+from pontoon.base.utils import extension_in, get_object_or_none, ajaxview
 
 
 class UtilsTests(TestCase):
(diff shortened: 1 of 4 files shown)


mathjazz:
I love this. It makes the viewsmuch clearer and reduces views.py alone by 135 lines! ;)

Two questions:
* Maybe we should add 'GET' / 'POST' as first (and required) argument to the @ajaxview, to make things more obvious for the reader?
* In the next iteration, @ajaxview should raise Http404 instead of return HttpResponse("error"), but that will require changing JS. It can wait.

But I'd rather defer to someone who knows how to code in Django, and the first one that comes to mind is @Osmose. ;)

pontoon/intro
```diff
@@ -1 +1 @@
-Subproject commit 30273cd602cb4ad2dbe669417769a9c17452f597
+Subproject commit d12907dd681c1d16c623b9dcceed9ff5e85c2ac6
```
mathjazz: Watch out!

jotes: whooops, sorry!

Osmose:
I really like the direction here, and agree that we should find a way to not repeat these parameters as often, but the current implementation isn't quite there:

- The whole "return a dict that will automatically be wrapped by an HttpResponse and JSONified" thing is a bit too magic for my tastes. A `JsonResponse` class that handles converting to JSON and such would be a more explicit yet concise alternative.
- As mathjazz mentioned, it's not obvious what the parameters to `ajaxview` mean. I can't really think of a concise way to express as arguments to a decorator. Maybe just always be explicit with the kwarg names?
- Adding arguments to the view function gets confusing when you already have the request argument and the arguments coming from the URL. It's annoying to have to keep in mind that ajax arguments come after the URL arguments.

It wouldn't remove as many lines, but maybe something like `get_object_or_404` but for parameters? It'd get rid of the try but still be very explicit about where the values are coming from.

Those are just my thoughts. If you have some alternative ideas too I'd love to hear them, this is a really cool refactoring! :D

jotes:
Hi,

Yeah, i agree about the parameters - i was not sure how to make this part explicit enough, without unecessary code duplication. I don't think that returning dictionary isn't explicit enough, however explicitness/implicitness is a really subjective thing ;-) Anyway, i'll rebase this PR and change into something more verbosive.

jotes:
Closing this for the time being.

PR #290, opened 2015-11-28
Fix bug 1218980: Unfuzzy existing fuzzy translations

mathjazz:
@Osmose r?

This should resolve issues we have with stats sometimes not counting fuzzies properly.

### DIFF
--- a/pontoon/sync/changeset.py
+++ b/pontoon/sync/changeset.py
@@ -147,6 +147,12 @@ class ChangeSet(object):
             )
         approved_translations = []
 
+        # Unfuzzy existing fuzzy translations to prevent duplicate fuzzies.
+        for translation in db_translations:
+            if translation.fuzzy:
+                translation.fuzzy = False
+                self.translations_to_update.append(translation)
+
         for plural_form, string in vcs_translation.strings.items():
             # Check if we need to modify an existing translation or
             # create a new one.
--- a/pontoon/sync/changeset.py
+++ b/pontoon/sync/changeset.py
@@ -146,12 +146,7 @@ class ChangeSet(object):
                 locale__code=locale_code,
             )
         approved_translations = []
-
-        # Unfuzzy existing fuzzy translations to prevent duplicate fuzzies.
-        for translation in db_translations:
-            if translation.fuzzy:
-                translation.fuzzy = False
-                self.translations_to_update.append(translation)
+        fuzzy_translations = []
 
         for plural_form, string in vcs_translation.strings.items():
             # Check if we need to modify an existing translation or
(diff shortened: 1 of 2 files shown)


pontoon/sync/changeset.py
```diff
@@ -147,6 +147,12 @@ def update_entity_translations_from_vcs(
             )
         approved_translations = []
 
+        # Unfuzzy existing fuzzy translations to prevent duplicate fuzzies.
+        for translation in db_translations:
```
Osmose: Wouldn't this possibly result in the same translation being in `translations_to_update` multiple times? Perhaps you should do this in the `old_translations` block below.

mathjazz: Is that an optimization issue or could it cause more serious problems?

In `old_translations` we only keep approved ones.

Osmose: I think it might cause more serious problems. A quick skim of django-bulk-update reveals that [they append all the pks of objects to be updated in a single list](https://github.com/aykut/django-bulk-update/blob/master/bulk_update/helper.py#L94), and eventually use that list in one big [generated update statement](https://github.com/aykut/django-bulk-update/blob/master/bulk_update/helper.py#L146). My expectation is that that would fail, but we'd need to test it to be sure.

> In old_translations we only keep approved ones.

You don't necessarily have to add them to `old_translations`, but you'll need to use that value when figuring out what translations have already been added to the list of translations to update.

mathjazz: So I tested this, and it seems to be working without a problem:
`bulk_update([Translation.objects.get(pk=13813), Translation.objects.get(pk=13813)])`

Still, at the bottom of the function I could do something like:
`self.translations_to_update = list(set(self.translations_to_update))`

Osmose: Interesting! However, that only checks that it doesn't throw an error.

I tested myself:

```
In [1]: from bulk_update.helper import bulk_update

In [2]: from pontoon.base.models import *

In [4]: trans = Translation.objects.all()[0]

In [8]: trans.approved
Out[8]: True

In [9]: trans2 = Translation.objects.all()[0]

In [10]: trans.approved = False

In [11]: trans2.fuzzy
Out[11]: False

In [12]: trans2.fuzzy = True

In [13]: bulk_update([trans, trans2])

In [14]: trans3 = Translation.objects.all()[0]

In [15]: trans3.approved
Out[15]: True

In [16]: trans3.fuzzy
Out[16]: False
```

The changes from `trans` overrode those from `trans2`, such that `approved` was changed to True, but `fuzzy` was still False.

mathjazz: Darn, that's interesting indeed. ;)

mathjazz:
@Osmose 2nd look?

Osmose:
r+, I think this is good. Nice work!

PR #293, opened 2015-12-01
Fix bug 1229067: Implement suggested celery settings for CloudAMQP.

Osmose:
@mathjazz r?

The big thing here, I think, is `CELERY_SEND_EVENTS` and the process arguments, as those queues appear to be generating 1.2 messages per second according to the CloudAMQP board vs 0.2 messages per second for our actual task queues.


mathjazz:
r+

PR #294, opened 2015-12-01
Fix bug 1190271: Automatically goes to next string to translate

sudheesh001:


mathjazz:
Thanks for the fix!

It works, but it's a bit shaky in terms of UI, because it opens the entity list briefly and then closes it to show the next entity. I suggest you do this:

1. Get rid of `gotoEntityListOrNextEntity()`. Simply remove the entire function, because we don't need it anymore. We always want to go to next entity.

2. Replace both `gotoEntityListOrNextEntity()` function calls with the line you added: `$('#next').click();`. And that's it.

We generally try to get rid of my old habit of trigering events like this and use function calls defined in event handler instead, so the code is easier to understand. But in this example I find `$('#next').click();` obvious enough, so it can wait until next time.

Good work!

sudheesh001:
@mathjazz I've made changes as you suggested. Could you please review it now :smile:

mathjazz:
r+!

Deploying in a few hours...

sudheesh001:
@mathjazz Could you assign me to the next issue that I could take up ?

mathjazz:
A few ideas:
https://bugzilla.mozilla.org/show_bug.cgi?id=1223770
https://bugzilla.mozilla.org/show_bug.cgi?id=1208569
https://bugzilla.mozilla.org/show_bug.cgi?id=1229079
https://bugzilla.mozilla.org/show_bug.cgi?id=1226759
https://bugzilla.mozilla.org/show_bug.cgi?id=1225827
https://bugzilla.mozilla.org/show_bug.cgi?id=1228770

Also, do you wanna join us on IRC?
https://cbe001.chat.mibbit.com/?url=irc%3A%2F%2Firc.mozilla.org%2Fpontoon

PR #295, opened 2015-12-02
Use local jQuery, Google CDN is slow in China

mathjazz:
@Osmose r?

### DIFF
--- a/pontoon/base/templates/base.html
+++ b/pontoon/base/templates/base.html
@@ -40,15 +40,10 @@
 
     <!-- Javascript at the bottom for fast page loading -->
     {% block site_js %}
-      <!-- Grab Google CDN's jQuery. fall back to local if necessary -->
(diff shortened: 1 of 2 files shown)


jimaek:
You could also use jsDelivr. We have a huge network in China mainland, plus we have a valid ICP license to operate there.
More info https://hacks.mozilla.org/2014/03/jsdelivr-the-advanced-open-source-public-cdn/
Scroll down for performance map http://www.jsdelivr.com/statistics

Osmose:
> You could also use jsDelivr. We have a huge network in China mainland, plus we have a valid ICP license to operate there.

Just for context, we're aware of jsDelivr, but part of this change is intentionally not relying on a generic CDN for libraries since we're going to enable our own CDN soon for Pontoon-specific bundles and wanted to include jQuery in those bundles.

Osmose:
r+!

PR #297, opened 2015-12-03
Skip updating translations for missing entities.

Osmose:
If we don't have both a VCS entity and a DB entity, we can't do anything useful to update the translations for the entity, so we just skip them.

This fixes an issue we were running into when testing the new MasterFirefoxOS translation files. Since there are source files that may not exist in certain locales, we get `None` back for `vcs_translation` and hit an error, when really we should be skipping those.

@mathjazz r?

### DIFF
--- a/pontoon/sync/core.py
+++ b/pontoon/sync/core.py
@@ -102,11 +102,9 @@ def update_resources(db_project, vcs_project):
 
 def update_translations(db_project, vcs_project, locale, changeset):
     for key, db_entity, vcs_entity in collect_entities(db_project, vcs_project):
-        # We shouldn't hit this situation in a real sync, but we might
-        # hit it during a test, so log it and continue just in case.
+        # If we don't have both the db_entity and vcs_entity we can't
+        # do anything with the translations.
         if db_entity is None or vcs_entity is None:
-            log.warning(u'Could not find VCS/DB entity for key `{key}` while '
-                        'updating translations, skipping.'.format(key=key))
             continue
 
         if not vcs_entity.has_translation_for(locale.code):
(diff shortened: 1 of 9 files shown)


mathjazz:
1) I'm trying to understand why is this hitting us in Master Firefox OS, but not in e.g. Mozilla.org, which also has "source files that may not exist in certain locales".

2) I still get "Skipping resource" [error messages](https://pastebin.mozilla.org/8853721). But only for af, bn_BD and el. Again, we don't get these for Mozilla.org, but I guess that's because we handle error differently for .lang?

3) And most importantly, only 138 strings get imported (for all locales), whereas in repositories there are actually 267 - 403 (which is correct and same as in Verbatim).

Osmose:
> And most importantly, only 138 strings get imported (for all locales), whereas in repositories there are actually 267 - 403 (which is correct and same as in Verbatim).

Wait, what? My local test of this code and MasterFirefoxOS (the new files) pulls in 271 strings for af, ja, etc. How are you seeing only 138 strings imported?

Osmose:
> 1) I'm trying to understand why is this hitting us in Master Firefox OS, but not in e.g. Mozilla.org, which also has "source files that may not exist in certain locales".

Because, as I just found out, the `po` parser is currently catching `IOError`s raised from the pofile parser and wrapping them in `ParseError`, but the langfile parser is still raising `IOError`s. The code that loads translated resources currently only skips over missing files by catching `IOError`, meaning that the `ParseError`s propagated upward and cancelled resource loading for _all_ the locales for that resource.

> 2) I still get "Skipping resource" error messages. But only for af, bn_BD and el. Again, we don't get these for Mozilla.org, but I guess that's because we handle error differently for .lang?

Yuuuuuup. See above.

> 3) And most importantly, only 138 strings get imported (for all locales), whereas in repositories there are actually 267 - 403 (which is correct and same as in Verbatim).

Because the mixup with `ParseError`s above propagated upward, it ended up cancelling parsing for the resource in every locale if a single locale was missing the resource. In effect, we were only syncing the resources that every active locale had, which is why you had less strings for having more locales active.

I've added a commit that fixes this. The string counts now closely match Verbatim prod I think (I believe they're off by 2 due to some lingering test data and differences in my local database vs prod).

mathjazz:
This explains everything. Tested all locales locally and strings match the ones in the repository. Excellent work! r+

PR #298, opened 2015-12-04
Fix bug 1229067: Implement suggested celery settings for CloudAMQP.

Osmose:
After some testing, it seems the `BROKER_HEARTBEAT` setting is what caused issues last time. Since it doesn't really reduce the amount of messages we send anyway (which was kind've the point) I just removed it.

@mathjazz r?


Osmose:
Oh yeah, I tested this on staging and got no errors while running a sync task. Not sure if there was some other failure that occurred that I also need to test for besides that.

mathjazz:
Nice work! r+

PR #300, opened 2015-12-04
Fix Bug 1223770: Do not allow edits/suggests to non-authenticated users

sudheesh001:
Makes the translation contents purely readonly for non authenticated users.

### DIFF
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -107,7 +107,7 @@ var Pontoon = (function (my) {
                       (this.approved ? this.approved_user ?
                         'Approved by ' + this.approved_user : '' : 'Approve') +
                       '"></button>' +
-                      '<button class="delete fa" title="Delete"></button>' +
+                      ((self.user.email && (self.user.email === this.email)) ? '<button class="delete fa" title="Delete"></button>' : '') +
                     '</menu>' +
                   '</header>' +
                   '<p class="translation" dir="auto" lang="' + self.locale.code + '">' +
@@ -1338,10 +1338,6 @@ var Pontoon = (function (my) {
             name = project.html(),
             slug = project.data('slug');
 
-        if (slug !== 'pontoon-intro' && !self.user.email) {
-          return self.endLoader('Sign in to access more projects.', 'error', true);
-        }
-
         $('.project .selector .title')
           .html(name)
           .data('slug', slug);
@@ -1501,6 +1497,16 @@ var Pontoon = (function (my) {
       // Profile menu
       // Register handlers to prep django-browserid before binding menu.
       django_browserid.registerWatchHandlers(function() {
+
+        function signIn() {
+          self.startLoader();
+          django_browserid.login().then(function(verifyResult) {
+            window.location.reload();
+          }, function(jqXHR) {
+            self.endLoader('Oops, something went wrong.', 'error');
+          });
+        }
+
         $('#profile .menu li').click(function (e) {
           if ($(this).has('a').length) {
             return;
@@ -1508,13 +1514,7 @@ var Pontoon = (function (my) {
           e.preventDefault();
 
           if ($(this).is(".sign-in")) {
-            self.startLoader();
-            django_browserid.login().then(function(verifyResult) {
-              window.location.reload();
-            }, function(jqXHR) {
-              self.endLoader('Oops, something went wrong.', 'error');
-            });
-
+            signIn();
           } else if ($(this).is('.download')) {
             self.updateFormFields($('form#download-file'));
             $('form#download-file').submit();
@@ -1528,6 +1528,11 @@ var Pontoon = (function (my) {
           } else if ($(this).is('.check-box')) {
             e.stopPropagation();
           }
+        });
+
+        $('p#sign-in-required > a#sidebar-signin').click(function (e) {
+          e.preventDefault();
+          signIn();
         });
       });
 
--- a/pontoon/base/templates/translate.html
+++ b/pontoon/base/templates/translate.html
@@ -239,27 +239,32 @@
         <li><a href="#plural6"><span></span><sup></sup></a></li>
       </ul>
     </nav>
-    <textarea dir="auto" lang="{{ locale.code }}" id="translation" placeholder="Type translation and press Enter to save"></textarea>
+    
+    <textarea {% if not user.is_authenticated() %} readonly {% endif %} dir="auto" lang="{{ locale.code }}" id="translation" placeholder="Type translation and press Enter to save"></textarea>
 
     <menu>
-      <div id="unsaved" class="warning-overlay">
-        <a href="#" class="cancel">&times;</a>
-        <h2>You have unsaved changes</h2>
-        <p>Sure you want to leave?</p>
-        <button id="leave-anyway">Leave anyway</button>
-      </div>
-      <div id="quality" class="warning-overlay">
-        <a href="#" class="cancel">&times;</a>
-        <h2>The following checks have failed</h2>
-        <ul></ul>
-        <button id="save-anyway"></button>
-      </div>
-      <div id="translation-length">
-        <span class="current-length"></span>|<span class="original-length"></span>
-      </div>
-      <button id="copy" title="Copy From Source (Alt + C)">Copy</button>
-      <button id="clear" title="Clear Translation (Alt + Backspace)">Clear</button>
-      <button id="save" title="Save Translation (Enter)"></button>
+      {% if user.is_authenticated() %}
+        <div id="unsaved" class="warning-overlay">
+          <a href="#" class="cancel">&times;</a>
+          <h2>You have unsaved changes</h2>
+          <p>Sure you want to leave?</p>
+          <button id="leave-anyway">Leave anyway</button>
+        </div>
+        <div id="quality" class="warning-overlay">
+          <a href="#" class="cancel">&times;</a>
+          <h2>The following checks have failed</h2>
+          <ul></ul>
+          <button id="save-anyway"></button>
+        </div>
+        <div id="translation-length">
+          <span class="current-length"></span>|<span class="original-length"></span>
+        </div>
+        <button id="copy" title="Copy From Source (Alt + C)">Copy</button>
+        <button id="clear" title="Clear Translation (Alt + Backspace)">Clear</button>
+        <button id="save" title="Save Translation (Enter)"></button>
+      {% else %}
+        <p id="sign-in-required"><a id="sidebar-signin" href="">Sign in</a> to translate.</p>
+      {% endif %}
     </menu>
 
     <div id="helpers" class="tabs">
--- a/pontoon/base/static/css/translate.css
+++ b/pontoon/base/static/css/translate.css
@@ -1092,6 +1092,14 @@ body > header aside p {
   content: "Suggest anyway";
 }
 
+p#sign-in-required {
+  font-style: italic;
+}
+
+p#sign-in-required > a#sidebar-signin {
+  color: #7BC876;
+}
+
 #editor > menu button#leave-anyway {
   background: #FF0A43;
   color: #EBEBEB;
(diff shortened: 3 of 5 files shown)


pontoon/base/templates/translate.html
```diff
@@ -239,27 +239,34 @@ <h3 class="no-match"><div class="fa fa-exclamation-circle"></div>No results</h3>
         <li><a href="#plural6"><span></span><sup></sup></a></li>
       </ul>
     </nav>
-    <textarea dir="auto" lang="{{ locale.code }}" id="translation" placeholder="Type translation and press Enter to save"></textarea>
-
+    {% if user.is_authenticated() %}
+      <textarea dir="auto" lang="{{ locale.code }}" id="translation" placeholder="Type translation and press Enter to save"></textarea>
+    {% else %}
+      <textarea readonly dir="auto" lang="{{ locale.code }}" id="translation" placeholder="Type translation and press Enter to save"></textarea>
+    {% endif %}
```
mathjazz: Let's only wrap `readonly` attribute in the if clause, so we don't duplicate code if it's not necessary.

pontoon/base/templates/translate.html
```diff
@@ -239,27 +239,34 @@ <h3 class="no-match"><div class="fa fa-exclamation-circle"></div>No results</h3>
...
+          <button id="save-anyway"></button>
+        </div>
+        <div id="translation-length">
+          <span class="current-length"></span>|<span class="original-length"></span>
+        </div>
+        <button id="copy" title="Copy From Source (Alt + C)">Copy</button>
+        <button id="clear" title="Clear Translation (Alt + Backspace)">Clear</button>
+        <button id="save" title="Save Translation (Enter)"></button>
+      {% else %}
+        <p><i>You need to be logged in to suggest or make changes.</i></p>
```
mathjazz: You should use CSS for styling instead of markup.

Another nit: please you something like "Sign in to suggest or make changes.", because we use sign in instead of log in elsewhere and because it's shorter and as such less likely to break into two lines.

mathjazz:
Nice work, thanks!

In addition to that, we'd like to make all projects accessible to non-authenticated users. You might only have the default (Pontoon Intro) project installed, so you should add some test projects, e.g. https://github.com/mozilla-l10n/www.mozilla.org. That way you will see that Pontoon doesn't let you open them, unless you're signed in.

You will need to make two changes:
1. In translate.js, remove the JS warning triggerring 'Sign in to access more projects'.
2. In views.py, remove the Check if user authenticated in the translate() view.

pontoon/base/templates/translate.html
sudheesh001: Is there a way to do this ? If I wrap the `readonly` in the `if` clause, for a logged in user, he/she would see 2 text areas, one which is readonly and one which is editable

pontoon/base/templates/translate.html
sudheesh001: > You should use CSS for styling instead of markup.

Are you referencing to the `<i>`

mathjazz: Yes.

pontoon/base/templates/translate.html
mathjazz: You only need to use one `<textarea>` tag and put `readonly` into if-block.

sudheesh001:
@mathjazz How about now ? I've had quite some problems setting up a new project. Some documentation added to the sphinx-docs might be really helpful.

sudheesh001:
I think we'd have to also remove or change the test case
```python
    def test_not_authed_nonpublic_project(self):
        """
        If the user is not authenticated and we're not translating
        project ID 1, redirect home.
        """
        # Clear out existing project with ID=1 if necessary.
        Project.objects.filter(id=2).delete()
        locale = LocaleFactory.create(code='fakelocale')
        project = ProjectFactory.create(id=2, slug='valid-project', locales=[locale])
        ResourceFactory.create(project=project)

        response = self.client.get('/fakelocale/valid-project/path/')
        assert_equal(response.status_code, 403)
```
to
```python
    def test_not_authed_nonpublic_project(self):
        """
        If the user is not authenticated and we're not translating
        project ID 1, redirect home.
        """
        # Clear out existing project with ID=1 if necessary.
        Project.objects.filter(id=2).delete()
        locale = LocaleFactory.create(code='fakelocale')
        project = ProjectFactory.create(id=2, slug='valid-project', locales=[locale])
        ResourceFactory.create(project=project)

        response = self.client.get('/fakelocale/valid-project/path/')
        assert_equal(response.status_code, 200)
```
I am making this change and rebasing

mathjazz:
Thanks, you've made great progress! I agree about the docs.

I think we are safe to remove the test, because not being authenticated is no longer a special case for the translate view.

Earlier I forgot two things I just spotted, sorry about that:

1. You can still copy suggestions to #translate textarea by clicking on them. Search for 'Copy helpers result to translation' click handler in translate.js and return if user not authenticated (email not set).

2. The Delete button is still visible for suggestions, which you can test by having at least two translations submitted for the same entity for the same language. Check out the getHistory() in translate.js and search for 'own' class. We make Delete visible if emails are the same, which is a false assumption, because it's true if emails are empty. You'll need to add another condition that user's email is actually set.

pontoon/base/templates/translate.html
```diff
@@ -239,27 +239,31 @@ <h3 class="no-match"><div class="fa fa-exclamation-circle"></div>No results</h3>
         <li><a href="#plural6"><span></span><sup></sup></a></li>
       </ul>
     </nav>
-    <textarea dir="auto" lang="{{ locale.code }}" id="translation" placeholder="Type translation and press Enter to save"></textarea>
-
+    
+    <textarea {% if not user.is_authenticated() %} readonly {% endif %} dir="auto" lang="{{ locale.code }}" id="translation" placeholder="Type translation and press Enter to save"></textarea>
```
mathjazz: Nit: You could add another blank line after <textarea>.

pontoon/base/templates/translate.html
```diff
@@ -239,27 +239,31 @@ <h3 class="no-match"><div class="fa fa-exclamation-circle"></div>No results</h3>
...
+          <button id="save-anyway"></button>
+        </div>
+        <div id="translation-length">
+          <span class="current-length"></span>|<span class="original-length"></span>
+        </div>
+        <button id="copy" title="Copy From Source (Alt + C)">Copy</button>
+        <button id="clear" title="Clear Translation (Alt + Backspace)">Clear</button>
+        <button id="save" title="Save Translation (Enter)"></button>
+      {% else %}
+        <p id="sign-in-required">Sign in to suggest or make changes.</p>
```
mathjazz: Let's simplify this to "Sign in to translate". Sorry for my split personality. ;)

sudheesh001:
> Thanks, you've made great progress! I agree about the docs.

> I think we are safe to remove the test, because not being authenticated is no longer a special case for the translate view.

Removed the tests, rebased and pushed the commit.

>1. You can still copy suggestions to #translate textarea by clicking on them. Search for 'Copy helpers result to translation' click handler in translate.js and return if user not authenticated (email not set).

So removing the `<button id="copy" title="Copy From Source (Alt + C)">Copy</button>` from the `is.authenticated()` and placing it outside the if-block should fix this ? So both `Copy` as well as the message `Sign in to translate` should be available ?

>2. The Delete button is still visible for suggestions, which you can test by having at least two translations submitted for the same entity for the same language. Check out the getHistory() in translate.js and search for 'own' class. We make Delete visible if emails are the same, which is a false assumption, because it's true if emails are empty. You'll need to add another condition that user's email is actually set.

```javascript
success: function(data) {
          if (data !== "error") {
            $.each(data, function() {
              list.append(
                '<li data-id="' + this.id + '" ' +
                (this.approved ? ' class="approved"' : '') +
                'title="Copy Into Translation (Tab)">' +
                  '<header class="clearfix' +
                    ((self.user.isTranslator) ? ' translator' :
                      ((self.user.email === this.email && !this.approved) ?
                        ' own' : '')) +
                    '">' +
                    '<div class="info">' +
                      ((!this.email) ? this.user :
                        '<a href="/contributors/' + this.email + '">' + this.user + '</a>') +
                      '<time class="stress" datetime="' + this.date_iso + '">' + this.date + '</time>' +
                    '</div>' +
                    '<menu class="toolbar">' +
                      '<button class="approve fa" title="' +
                      (this.approved ? this.approved_user ?
                        'Approved by ' + this.approved_user : '' : 'Approve') +
                      '"></button>' +
                      '<button class="delete fa" title="Delete"></button>' +
                    '</menu>' +
                  '</header>' +
                  '<p class="translation" dir="auto" lang="' + self.locale.code + '">' +
                    self.doNotRender(this.translation) +
                  '</p>' +
                '</li>');
            }
```
So I understand that this button `      '<button class="delete fa" title="Delete"></button>' +` needs a conditional check before this it's rendered right ?

mathjazz:
> Removed the tests, rebased and pushed the commit.

Excellent!

> So removing the <button id="copy" title="Copy From Source (Alt + C)">Copy</button> from the is.authenticated() and placing it outside the if-block should fix this ? So both Copy as well as the message Sign in to translate should be available ?

No. That's a copy button, which copies original strings into translation field. It's hidden and that's how it should be. I'm talking about copying suggestions from the History, Machinery, Locales or Search tabs, which you can do by clicking on them or pressing Tab. Ideally, we should prevent this, but on a second thought we could also keep it for users to explore. This will allow them to experience the functionality, but they won't be able to save, so we're safe. So let's forget about this.

> So I understand that this button '<button class="delete fa" title="Delete"></button>' + needs a conditional check before this it's rendered right ?

Yes. Basically `self.user.email === this.email` needs to be upgraded to something like `self.user.email && (self.user.email === this.email)`.

sudheesh001:
@mathjazz Made the required changes, the delete button is only visible in the history if and only if the user is logged in and it's the same user who has made the suggestion. Or else the button isn't rendered.

mathjazz:
Fantastic work again!

I got another idea for improvement - what if you turn Sign in to translate into a link to actually sign in? Maybe only use "Sign in" as the link and paint it "Pontoon" green (#7BC876).

sudheesh001:
> link to actually sign in

I couldn't retrieve it using `{{ url('signin') }}` Is there a specific URL route for `signin` I was unable to find it, so I left the `href=` empty.

mathjazz:
There's no URL, so that's ok. Now you need to add a click handler in the same fashion as we do it in the `#profile` menu.

sudheesh001:
@mathjazz Fixed ! :smile:

pontoon/base/static/js/translate.js
```diff
@@ -1529,6 +1525,19 @@ var Pontoon = (function (my) {
...
+
+          if($(this).is("#sidebar-signin")) {
+            self.startLoader();
+            django_browserid.login().then(function(verifyResult) {
+              window.location.reload();
+            }, function(jqXHR) {
+              self.endLoader('Oops, something went wrong.', 'error');
+            });
+          }
+        })
```
mathjazz: You should move this inside [django_browserid.registerWatchHandlers()](http://django-browserid.readthedocs.org/en/latest/api/javascript.html#registerWatchHandlers). You should also put lines 1533-1538 in a private function and reuse it in both Sign in triggers.

pontoon/base/static/js/translate.js
```diff
@@ -1529,6 +1525,19 @@ var Pontoon = (function (my) {
             e.stopPropagation();
           }
         });
+
+        $('p#sign-in-required > a#sidebar-signin').click(function (e) {
+          e.preventDefault();
+
+          if($(this).is("#sidebar-signin")) {
```
mathjazz: No need for this if-block.

mathjazz:
Great work, turn that sign in handler into a reusable function and I'll merge.

sudheesh001:
Changed it and put the sign in into a private `signIn()` function

mathjazz:
r+!

PR #301, opened 2015-12-04
Fix Bug 1208569: Autolinkify texts, introduces linkify string prototype

sudheesh001:


pontoon/base/static/js/translate.js
```diff
@@ -168,10 +168,31 @@ var Pontoon = (function (my) {
...
       $("#editor")[0].entity = entity;
 
+      /*
+       * Linkifies any traces of URLs present in a given string.
+       * Usage : string.linkify();
+       *
+       * Matches the URL Regex and parses the required matches.
+       * Can find more than one URL in the given string.
+       */
+      if(!String.linkify) {
```
mathjazz: Nit: space between if and (.

pontoon/base/static/js/translate.js
```diff
@@ -168,10 +168,31 @@ var Pontoon = (function (my) {
...
+          var urlPattern = /\b(?:https?|ftp):\/\/[a-z0-9-+&@#\/%?=~_|!:,.;]*[a-z0-9-+&@#\/%=~_|]/gim;
+          // www. sans http:// or https://
+          var pseudoUrlPattern = /(^|[^\/])(www\.[\S]+(\b|$))/gim;
+
+          return this
+            .replace(urlPattern, '<mark class="placeable"><a href="$&">$&</a></mark>')
+            .replace(pseudoUrlPattern, '<mark class="placeable">$1<a href="http://$2">$2</a></mark>');
+        };
+      }
+
```
mathjazz: Could you make this a public function, that takes string as an argument, to match with the rest of Pontoon's code style?

pontoon/base/static/js/translate.js
```diff
@@ -168,10 +168,31 @@ var Pontoon = (function (my) {
...
       // Metadata: comments, sources, keys
       $('#metadata').empty();
       if (entity.comment) {
-        self.appendMetaData('Comment', this.doNotRender(entity.comment));
+        entity.comment = entity.comment.linkify();
```
mathjazz: You should keep entity.comment in its original form, we might use it somewhere. Let's use a new variable, e.g. comment.

pontoon/base/static/js/translate.js
```diff
@@ -168,10 +168,31 @@ var Pontoon = (function (my) {
...
+        self.appendMetaData('Comment', entity.comment);
```
mathjazz: We have a lot of comments that contain markup and we don't want to render them:
https://public.etherpad-mozilla.org/p/pontoon.comments

We only want to render the ones that contain links. To detect them, I suggest you compare entity.comment to the output of the linkify(). If they are the same, we don't have a link, so we should set comment to this.doNotRender(entity.comment). Else, we should set it to the linkified value.

sudheesh001:
@mathjazz Made the changes as suggested. :smile:

pontoon/base/static/js/translate.js
```diff
@@ -171,7 +171,13 @@ var Pontoon = (function (my) {
...
       $('#metadata').empty();
       if (entity.comment) {
-        self.appendMetaData('Comment', this.doNotRender(entity.comment));
+        comment = this.linkify(entity.comment);
+        if (comment !== entity.comment) {
+          self.appendMetaData('Comment', comment);
+        }
+        else {
+          self.appendMetaData('Comment', this.doNotRender(entity.comment));
+        }
```
mathjazz: You could reduce the number of lines by doing:
```
var comment = this.linkify(entity.comment);
if (comment === entity.comment) {
  comment = this.doNotRender(entity.comment);
}
self.appendMetaData('Comment', comment);
```

pontoon/base/static/js/translate.js
```diff
@@ -171,7 +171,13 @@ var Pontoon = (function (my) {
       // Metadata: comments, sources, keys
       $('#metadata').empty();
       if (entity.comment) {
-        self.appendMetaData('Comment', this.doNotRender(entity.comment));
+        comment = this.linkify(entity.comment);
```
mathjazz: `var comment`

mathjazz:
Great work! r+

pontoon/base/static/js/main.js
```diff
@@ -107,6 +107,24 @@ var Pontoon = (function (my) {
...
+     */
+    linkify: function (string) {
+      // http://, https://, ftp://
+      var urlPattern = /\b(?:https?|ftp):\/\/[a-z0-9-+&@#\/%?=~_|!:,.;]*[a-z0-9-+&@#\/%=~_|]/gim;
+      // www. sans http:// or https://
+      var pseudoUrlPattern = /(^|[^\/])(www\.[\S]+(\b|$))/gim;
+
+      return string
+        .replace(urlPattern, '<mark class="placeable"><a href="$&">$&</a></mark>')
+        .replace(pseudoUrlPattern, '<mark class="placeable">$1<a href="http://$2">$2</a></mark>');
```
mathjazz: One last thing: we should probably add `target="_blank"` attribute.

sudheesh001:
@mathjazz Thank you ! Made the changes :smile:

mathjazz:
Beautiful!

jotes:
There's a django template function: https://docs.djangoproject.com/en/1.9/ref/templates/builtins/#urlize . Is there any particular reason to do this in javascript?

sudheesh001:
@jotes I believe it's because `urlize` converts the strings URL into `<a href="url">url</a>` where as for the UI we needed that anchor tags to be wrapped within `<mark class="placeable"><a>.......</a></mark>`

mathjazz:
Thanks for pointing this out, @jotes.

@sudheesh001 I actually removed that `<mark>` tag, I don't think it's necessary:
https://github.com/mozilla/pontoon/commit/185939c0b2c75dcac64fd063564def64b78cfe96

I guess it would be actually simpler to do it in Django, but I'm not sure it makes sense to change it t this stage.

PR #302, opened 2015-12-04
Ignore non-failure exceptions like Http404 and Permission Denied.

Osmose:
If we still want notifications for excessive 404s we can configure a search in Papertrail.

@mathjazz r?


mathjazz:
r+

PR #303, opened 2015-12-05
Properly log project syncs that skip sync completely.

Osmose:
Also logs end time for repo syncs that fail due to no locales being found within a repo.

This should fix most of the data issues with the sync logs. I'll hopefully come up with a separate PR implementing pagination.

@mathjazz r?

### DIFF
--- a/pontoon/base/utils.py
+++ b/pontoon/base/utils.py
@@ -496,3 +496,17 @@ def handle_upload_content(slug, code, part, f, user):
             changed_entities[key] = ChangedEntityLocale(entity=t.entity, locale=t.locale)
 
     ChangedEntityLocale.objects.bulk_create(changed_entities.values())
+
+
+def latest_datetime(datetimes):
+    """
+    Return the latest datetime in the given list of datetimes,
+    gracefully handling `None` values in the list. Returns `None` if all
+    values in the list are `None`.
+    """
+    if all(map(lambda d: d is None, datetimes)):
+        return None
+
+    min_datetime = timezone.make_aware(datetime.min)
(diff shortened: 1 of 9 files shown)


pontoon/base/utils.py
```diff
@@ -496,3 +496,17 @@ def handle_upload_content(slug, code, part, f, user):
...
+    Return the latest datetime in the given list of datetimes,
+    gracefully handling `None` values in the list. Returns `None` if all
+    values in the list are `None`.
+    """
+    if all(map(lambda d: d is None, datetimes)):
+        return None
+
+    min_datetime = timezone.make_aware(datetime.min)
+    datetimes = map(lambda d: d or min_datetime, datetimes)
+    return max(datetimes)
```
mathjazz: Cool! We can probably also reuse this in https://github.com/mozilla/pontoon/blob/master/pontoon/base/models.py#L875.

pontoon/sync/models.py
```diff
@@ -43,18 +50,41 @@ class ProjectSyncLog(BaseLog):
...
+            return self.skipped_end_time
+        elif self.finished:
             aggregate = (self.repository_sync_logs
                             .all()
                             .aggregate(Max('end_time')))
             return aggregate['end_time__max']
+        else:
+            return None
+
+    # Possible sync status'. May eventually become a model field.
```
mathjazz: Remove (or add) '.

mathjazz:
Nice addition, r+.

This should make sync log actually useful. :-) Later on, once we add things like what exactly got synced, it will be crazy useful.

PR #306, opened 2015-12-14
Fix bug 1232378. Show projects only if search is empty or name matches.

jotes:
Hello! @mathjazz could you look at this? thx!


mathjazz:
r+

But on long term, we should get rid of this project loading stuff. By either making DB queries faster and/or splitting them further. To e.g. only get project-locale data first and when one of them is selected, then have another request bring resource-stats data.

PR #307, opened 2015-12-16
Fix bug 1231516. Changed keyboard shortcuts to avoid collisions with …

jotes:
Hi, @mathjazz, could you review?

### DIFF
--- a/pontoon/base/static/css/translate.css
+++ b/pontoon/base/static/css/translate.css
@@ -493,7 +493,7 @@ body > header .notification li {
 
 #profile.select .popup {
   right: 5px;
-  width: 300px;
+  width: 350px;
 }
 
 #info {
--- a/pontoon/base/static/js/main.js
+++ b/pontoon/base/static/js/main.js
@@ -657,9 +657,9 @@ $(function() {
       return false;
     }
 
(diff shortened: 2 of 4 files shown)


mathjazz:
Looks great. We should also use Alt + Ctrl combination with Up/Down to move to (Next/Previous) string.

r+wc

jotes:
Actually, I tried but for some reason, keyup() wasn't called if i pressed
these three keys (ctrl+alt+arrow). Maybe it was some kind of glitch, i'll
look into this.

On Thu, Dec 17, 2015 at 3:36 PM, Matjaž Horvat <notifications@github.com>
wrote:

> Looks great. We should also use Alt + Ctrl combination with Up/Down to
> move to (Next/Previous) string.
>
> r+wc
>
> —
> <https://github.com/mozilla/pontoon/pull/307#issuecomment-165467991>.
>



-- 
Kind regards

​Jarek "jotes" Śmiejczak,
​P​

jotes:
@mathjazz Unfortunately, it looks like that ctrl+alt+arrow collide with default shortcuts for the Unity in ubuntu (they are responsible for switching workspaces). I'm not sure if that's the case for the rest of the systems (?).

PR #309, opened 2015-12-20
Fix bug 1229854: Allow users to login from anywhere

sudheesh001:


pontoon/base/static/js/translate.js
```diff
@@ -1538,6 +1538,13 @@ var Pontoon = (function (my) {
           e.preventDefault();
           signIn();
         });
+
+        $('ul.links > li#sign-in').click(function (e) {
+          console.log('Reached here 1 !');
```
jotes: I think that console.log isn't required here.

pontoon/base/static/js/translate.js
```diff
@@ -1538,6 +1538,13 @@ var Pontoon = (function (my) {
           e.preventDefault();
           signIn();
         });
+
+        $('ul.links > li#sign-in').click(function (e) {
```
jotes: Is there a specific reason to not use just '#sign-in'?

pontoon/base/static/js/translate.js
```diff
@@ -1538,6 +1538,13 @@ var Pontoon = (function (my) {
...
+          console.log('Reached here 1 !');
+          e.preventDefault();
+          signIn();
```
jotes: Unfortunately, you have to move signIn (body of function and this event handler) to main.js because we have to support login on pages like https://pontoon.mozilla.org/terminology/ etc.

pontoon/base/templates/landing.html
```diff
@@ -14,6 +14,7 @@
         <li><a href="{{ url('pontoon.search') }}">Terminology</a></li>
         <li><a href="https://developer.mozilla.org/en-US/docs/Localizing_with_Pontoon" target="_blank">Help</a></li>
         <li id="admin"{% if not perms.base.can_manage %} class="hidden"{% endif %}><a href="{{ url('pontoon.admin') }}">Admin</a></li>
+        <li id="sign-in" {% if user.is_authenticated() %} class="hidden" {% endif %}><a href="">Sign in</a></li>
```
jotes: I would change this into something like this:
```
{% if user.is_authenticated() %}
<li id="sign-in"
{% else %}
<li id="signout"....
{% endif %}
```
It's better to not leave hidden things in markup if they aren't needed.

pontoon/base/templates/landing.html
```diff
@@ -14,6 +14,7 @@
...
         <li id="signout"{% if not user.is_authenticated() %} class="hidden"{% endif %}><a href="{{ url('signout') }}" title="{{ user.email }}">Sign out</a></li>
```
jotes: I would change signout into 'sign-out' to keep clasnames consistent.

jotes:
hi @sudheesh001, i've added  my suggestions, i hope you won't mind. The final decisions are belong to @mathjazz and @Osmose.

sudheesh001:
Hi @jotes Thanks for your suggestions, this PR is incomplete, I added the information to translate.js knowing that it needs to be in main.js but that'd mean rewriting the django-browserid event handler to login ? Why can't we just go for a `{{ url_for('signin') }}` just like how we do for signout so that its easier, rather than doing it via a JS event ? Is this possible @mathjazz and @Osmose ?

pontoon/base/static/js/translate.js
sudheesh001: I used these to debug, In the final rebased commit, i'll remove all of these :smile:

sudheesh001:
Reading the documentation i included the `browserid/api.js` and `browserid/browserid.js` but that threw a jQuery error, so I included jQuery too into the required files, I am not sure if this is the way for this to be done ... Right now I have console errors `TypeError: Cannot read property 'watch' of undefined`

I was wondering why we haven't used http://django-browserid.readthedocs.org/en/latest/api/python.html#django_browserid.helpers.browserid_login instead ?

pontoon/base/static/js/main.js
```diff
@@ -448,6 +448,55 @@ $(function() {
       .toUpperCase().indexOf(m[3].toUpperCase()) >= 0;
   };
 
+    // Profile menu
```
mathjazz: Nit: Mind the indent.

pontoon/base/static/js/main.js
```diff
@@ -448,6 +448,55 @@ $(function() {
...
+  // Register handlers to prep django-browserid before binding menu.
+  django_browserid.registerWatchHandlers(function() {
+    function signIn() {
+      self.startLoader();
```
mathjazz: Pontoon.startLoader();

pontoon/base/static/js/main.js
```diff
@@ -448,6 +448,55 @@ $(function() {
...
+      }
+    });
+
+    $('p#sign-in-required > a#sidebar-signin').click(function (e) {
+      e.preventDefault();
+      signIn();
+    });
+
+    $('ul.links > li#sign-in').click(function (e) {
+      console.log('Reached here 1 !');
```
mathjazz: Remove console.log().

pontoon/base/static/js/main.js
```diff
@@ -448,6 +448,55 @@ $(function() {
...
+      e.preventDefault();
+      signIn();
+      console.log('Reached here 2 !');
```
mathjazz: Remove console.log().

pontoon/base/static/js/main.js
```diff
@@ -448,6 +448,55 @@ $(function() {
...
+      if ($(this).has('a').length) {
+        return;
+      }
+      e.preventDefault();
+
+      if ($(this).is(".sign-in")) {
+        signIn();
+
+      } else if ($(this).is('.download')) {
+        self.updateFormFields($('form#download-file'));
```
mathjazz: *Pontoon*.updateFormFields

pontoon/base/static/js/main.js
```diff
@@ -448,6 +448,55 @@ $(function() {
...
 
+    // Profile menu
+  // Register handlers to prep django-browserid before binding menu.
+  django_browserid.registerWatchHandlers(function() {
+    function signIn() {
+      self.startLoader();
+      django_browserid.login().then(function(verifyResult) {
+        window.location.reload();
+      }, function(jqXHR) {
+        self.endLoader('Oops, something went wrong.', 'error');
```
mathjazz: **Pontoon**.endLoader();

pontoon/base/templates/landing.html
```diff
@@ -1,7 +1,8 @@
 {% extends "base.html" %}
-
 {% block content %}
-
+<script src="{{ static('js/jquery-1.11.1.min.js') }}"></script>
+<script src="{{ static('browserid/api.js') }}"></script>
+<script src="{{ static('browserid/browserid.js') }}"></script>
```
mathjazz: No need for that here, please also restore removed blank lines.

mathjazz:
Nice work!

You need to make a few changes:
* Move `<script src="https://login.persona.org/include.js"></script>` from `translate.html` to `base.html` (just before the `</body>`).
* In `pontoon/settings/base.py`, move `browserid/api.js` from `translate` to `main` (after jQuery).

sudheesh001:
@mathjazz I've made the required changes and rebased and squashed it into one commit but I still face a problem with the `api.js` file complaining on the console.
```py
Uncaught TypeError: Cannot read property 'extend' of undefinedgetAssertion @ api.js:54
login @ api.js:24
signIn @ main.js:456
(anonymous function) @ main.js:495
m.event.dispatch @ jquery-1.11.1.min.js:3
r.handle @ jquery-1.11.1.min.js:3
```
Can you help with this ? Do I have to include the jQuery ? but it looks like jquery-1.11.1.min.js is already there.

pontoon/settings/base.py
```diff
@@ -280,6 +280,7 @@ def path(*args):
     },
     'main': {
         'source_filenames': (
+            'browserid/api.js',
```
mathjazz: Move this *after* jQuery. :)

pontoon/base/static/js/main.js
```diff
@@ -448,6 +448,54 @@ $(function() {
       .toUpperCase().indexOf(m[3].toUpperCase()) >= 0;
   };
 
+  // Profile menu
+  // Register handlers to prep django-browserid before binding menu.
+  django_browserid.registerWatchHandlers(function() {
+    function signIn() {
+      Pontoon.startLoader();
+      django_browserid.login().then(function(verifyResult) {
+        console.log(verifyResult);
```
mathjazz: And remove console.log and you should be fine.

sudheesh001: Ah ! I had used this to figure out what was going wrong. I guess I understood it now. Fixing it right away. :smile:

sudheesh001:
@mathjazz Hopefully complete now ! :smile: Thanks, Learnt a lot by reading documentation of new pip libraries which I could use in my personal projects.

Also what do you think about [django-toolbar](https://github.com/django-debug-toolbar/django-debug-toolbar) ? Would it be helpful integrating it into pontoon to check the SQL performance and optimize on the queries being made ?

mathjazz:
Bravo! r+

As for the django-toolbar, I think it's a great tool and use it. You should install and enable it locally.

sudheesh001:
Could we integrate it into pontoon itself ? so that during development it'll be a nice addon ?

mathjazz:
I don't think it make sense - you install it locally and then add it to INSTALLED_APPS when you need it, without polluting the codebase.

sudheesh001:
Okay great. Would this https://bugzilla.mozilla.org/show_bug.cgi?id=1219431 be a nice next bug to fix ? Or is there some other bug you'd suggest ?

mathjazz:
That's one possibility and there might be others even better. I sent you a list of good bugs to fix a few weeks ago, but I can't find it ATM. :/ Do you happen to have it?

sudheesh001:
@mathjazz Found it.
https://github.com/mozilla/pontoon/pull/294#issuecomment-160949399

PR #310, opened 2015-12-22
Fix bug 1219431: Adding translated word count to profile view

sudheesh001:
I think this is a very naïve approach to show the word count, is there a better way of querying the QuerySet without iterating through all of them ? Also, should something similar be made for the `/contributors` view ? The top contributors view is passed a context of the `user` and `period`, How do I retrieve the other information from here ? @mathjazz

### DIFF
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -982,17 +982,36 @@ var Pontoon = (function (my) {
       });
     },
 
-
     /*
      * Update progress indicator and value
      */
     updateProgress: function () {
+
+      /*
+       * Count the number of words translated
+       */
+      function wordProgress () {
+        var entities = Pontoon.entities;
+        var entitiesLength = entities.length;
+        var wordCount = 0;
+        for (var entityIndex = 0; entityIndex < entitiesLength; entityIndex++) {
+          var entityString = entities[entityIndex].translation[0].string;
+          if (entityString !== null) {
+            var tempEntityString = entityString.split(' ');
+            wordCount += tempEntityString.length;
+          }
+        }
+        return wordCount;
+      }
+
       var all = $("#entitylist .entity").length,
           approved = $("#entitylist .entity.approved").length,
           translated = $("#entitylist .entity.translated").length,
           fuzzy = $("#entitylist .entity.fuzzy").length,
           untranslated = all - translated - approved - fuzzy,
+          words = wordProgress();
           fraction = {
+            words: words ? words / 1 : 0,
             approved: all ? approved / all : 0,
             translated: all ? translated / all : 0,
             fuzzy: all ? fuzzy / all : 0,
@@ -1031,6 +1050,7 @@ var Pontoon = (function (my) {
       // Update details in the menu
       $('#progress .menu').find('header span').html(all).end()
         .find('.details')
+          .find('.words p').html(words).end()
           .find('.approved p').html(approved).end()
           .find('.translated p').html(translated).end()
           .find('.fuzzy p').html(fuzzy).end()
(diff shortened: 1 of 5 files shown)


sudheesh001:
Here's a screenshot.
screen shot 2015-12-22 at 12 41 12 pm

pontoon/base/views.py
```diff
@@ -264,11 +264,17 @@ def contributor(request, email):
         })
 
     timeline.reverse()
+    words = []
```
jotes: Is there a need for keeping this list of words? we display only count of them.

sudheesh001: So do I just use a counter and increment `+=` or do I do what is being done here and return only the  `len(words)` to the view instead of doing `{{ words | length }}` ?

I used it this way because I felt it might come in handy since translations from the QuerySet is also being sent to the view where the `.count()` is calculating the values.

Should I be changing this @jotes :smile:

mathjazz:
This is actually not what the bug is about. ;) Added another comment.

sudheesh001:
So @mathjazz Is this about the list that pops up when we hover on the progress bar ? Should I add the word count over there ?

mathjazz:
That and in the translate view, when you click on the rounded progress chart.

sudheesh001:
screen shot 2015-12-26 at 6 21 12 pm

sudheesh001:
I have a small query regarding implementing the word count on the other pages like `/project` etc.., I've noticed the model which is as follows
```py
def get_chart_data(stats):
    """Get chart data for stats."""

    return stats.aggregate(
        total=Sum('resource__entity_count'),
        approved=Sum('approved_count'),
        translated=Sum('translated_count'),
        fuzzy=Sum('fuzzy_count')
    )
```
I have to add a new `words` count, to this stats aggregation but I am unable to figure out how to proceed with this. Could you guide me through @mathjazz @jotes

pontoon/base/static/js/translate.js
```diff
@@ -982,17 +982,36 @@ var Pontoon = (function (my) {
...
     /*
      * Update progress indicator and value
      */
     updateProgress: function () {
+
+      /*
+       * Count the number of words translated
+       */
+      function wordProgress () {
+        var entities = $("#entitylist .entity .string-wrapper .translation-string");
```
jotes: There's a global array of entities, look for Pontoon.entities.

sudheesh001: In the objects in `Pontoon.entitites` how would I know which of the strings are translated, we only have to count the words from the strings which have already been translated right ?

sudheesh001: Okay on closer inspection, I guess I figured it out. `Pontoon.entities[x].translation[0].string` being `null` means no translations made, and a string means the translated words. :smile:

jotes:
Hi @sudheesh001! As @mathjazz already mentioned here: https://bugzilla.mozilla.org/show_bug.cgi?id=1219431#c3 you should add new field to the resource class.
I would start by doing following things: https://gist.github.com/jotes/9eb8b86752a6643d4df3 . Next steps would be to add next aggregate to the get_chart_data, something like a 'words_count=Sum('resource__words_count'),' and display this aggregated field  in the UI. If you have more questions - feel free to ask, i'll try to respond tomorrow.

sudheesh001:
Thanks for this @jotes I've made the required changes to the `models` and the `core.py` and migrated the required changes, The database has the default value of 0 at this moment, How do I execute the changes in core.py to check all the entities and populate the database with the word count accordingly. :smile: ? 

Can you point me to some learning resource where I can learn more about this from ?

jotes:
@sudheesh001 Some of the concepts are explained here: http://mozilla-pontoon.readthedocs.org/en/latest/dev/sync.html#translation-sync .
Also, i think it would be good idea (if not necessity) to create unit/integration tests, it will help you to understand code more thoroughly.

mathjazz:
Closing due to inactivity.

PR #311, opened 2015-12-22
Fix bug 1234501: Make profile toggles work

mathjazz:
@jotes @sudheesh001 r?

### DIFF
--- a/pontoon/base/static/js/main.js
+++ b/pontoon/base/static/js/main.js
@@ -597,7 +597,7 @@ $(function() {
     Pontoon.startLoader();
 
     $.ajax({
-      url: '/api/v1/user/' + Pontoon.user.email + '/',
+      url: '/api/v1/user/' + $('#server').data('email') + '/',
       type: 'POST',
       data: {
         csrfmiddlewaretoken: $('#server').data('csrf'),
@@ -613,7 +613,7 @@ $(function() {
           Pontoon.endLoader(self.text() + ' ' + status + '.');
           $('.notification').addClass('menu-open');
 
-          if (self.is('.force-suggestions')) {
+          if (self.is('.force-suggestions') && Pontoon.user) {
             Pontoon.user.forceSuggestions = is_enabled;
             Pontoon.postMessage("UPDATE-ATTRIBUTE", {
               object: 'user',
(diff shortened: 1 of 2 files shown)


pontoon/base/static/js/main.js
```diff
@@ -597,7 +597,7 @@ $(function() {
     Pontoon.startLoader();
 
     $.ajax({
-      url: '/api/v1/user/' + Pontoon.user.email + '/',
+      url: '/api/v1/user/' + $('#server').data('email') + '/',
```
jotes: Maybe we should move initialization of user data before this line to keep one source of data for the user profile in javascript?

mathjazz: This method is also used by user.js, where we don't have Pontoon.user.

sudheesh001:
@mathjazz r- ; The Ajax query results in `OOPS SOMETHING WENT WRONG` with the Network information throwing a 404 in request for the `http://localhost:8000/api/v1/user/undefined/`

pontoon/base/static/js/main.js
jotes: Sure, no problem.

jotes:
@sudheesh001 yup it probably won't work for anonymous users.

sudheesh001:
@jotes I am logged in :+1: In one browser as admin and in another as a regular user.

mathjazz:
That's weird, what does `$('#server').data()` return for you? Did you merge OK?

sudheesh001:
@mathjazz It returns the csrf `Object {csrf: "cOoPkHfO0JRKU1D63bTUjrr1tZca4xku"}`

mathjazz:
It should also return email, you either didn't merge correctly or you're not authenticated;
https://github.com/mozilla/pontoon/pull/311/files#diff-5993865859aacd25a269caf6f0218b3cR13

sudheesh001:
@mathjazz I guess so, seems to be working fine now, I was able to see the
```html
<div id=​"server" class=​"hidden" data-csrf=​"vNiPMATEweL0iuGZlQvao0l4mIfCwlZm" data-email=​"root@root.com">​
​</div>​
```
but the `$('#server').data()` ended up only with `csrf`, It was probably the cache, seems to be working fine now. r+ :smile: Sorry for the confusion

jotes:
post r+

PR #315, opened 2015-12-31
Bug 1212415: removed old version of sync and added support for new sy…

jotes:
Hi,
I know that this bug has P4 priority, but it was confusing for me to read through unsupported legacy code.
I've done three things here:
* updated docs
* removed old code from administration module
* (optional and need to be discussed) added two small buttons for pull/commit of projects in project synchronization module, they would need some css magic (unfortunately, css isn't my strong suite).

Could you review this @mathjazz? Thanks in advance. :bow:


mathjazz:
Thanks for taking care of this, much appreciated!

Two considerations:
1. I'm not sure why we need pull/commit buttons. We need to check logs to see the progress anyways. It's used rarely and ony by one person ATM.
2. Removing files.py removes support for l20n file format, so we should either file a bug (P2) or implement it using the new sync API as part of this bug.

jotes:
@mathjazz 
ad 1. I removed buttons.
ad 2. https://bugzilla.mozilla.org/show_bug.cgi?id=1236109

pontoon/administration/templates/admin_project.html
```diff
@@ -7,7 +7,7 @@
 {% block header %}{% endblock %}
 
 {% block middle %}
-<form method="POST" action="{{ request.path }}" class="{% if pk %}edit{% else %}add{% endif %}">
+<form method="POST" action="{{ request.path }}" class="project {% if pk %}edit{% else %}add{% endif %}">
```
mathjazz: Why do we need to add the `project` class here?

mathjazz:
Thanks, r+.

We should also move `vcs.py` to the `sync` app, because that's where it's being used and because we'll remove `administration` app anyways (#207). But that doesn't need to be part of this PR.

jotes:
@mathjazz i created new module for vcs-related code, moved administration.vcs there.

mathjazz:
Great job!

PR #316, opened 2015-12-31
Bug 1211789: synchronization of all projects shouldn't display warning.

jotes:
Hi, i'm sorry that i've discovered this so late. Could you review @mathjazz?


mathjazz:
r+

PR #317, opened 2016-01-05
Download projects with few resources as ZIP

mathjazz:
If project has mutiple resources, but less than 10, download all
files at once in a ZIP. Else, download separate files.

This wasn't tracked in bugzilla, it's a quick response to:
https://groups.google.com/forum/#!msg/mozilla.dev.l10n/aiJm4EH4s8w/V14iu-IUBwAJ

@jotes r?

### DIFF
--- a/pontoon/base/utils.py
+++ b/pontoon/base/utils.py
@@ -4,7 +4,9 @@ import logging
 import os
 import re
 import requests
+import StringIO
 import tempfile
+import zipfile
 
 from datetime import datetime
 
@@ -321,7 +323,7 @@ def _download_file(prefixes, dirnames, relative_path):
 
 
 def _get_relative_path_from_part(slug, part):
-    """ Check if part is a Resource path or Subpage name. """
+    """Check if part is a Resource path or Subpage name."""
     # Avoid circular import; someday we should refactor to avoid.
     from pontoon.base.models import Subpage
     try:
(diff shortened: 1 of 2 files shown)


pontoon/base/utils.py
```diff
@@ -398,17 +386,75 @@ def get_download_content(slug, code, part):
...
+        Locale,
+        Project,
+        Resource,
+    )
+
+    project = get_object_or_404(Project, slug=slug)
+    locale = get_object_or_404(Locale, code__iexact=code)
+
+    # Download a ZIP of all files if project has less than 10 resources
+    if 1 < project.resources.count() < 10:
```
jotes: 0 < project.resources.count() < 10 ?

jotes: also we could create some django config for this, it will pay-off in future.

pontoon/base/utils.py
```diff
@@ -398,17 +386,75 @@ def get_download_content(slug, code, part):
...
+            os.remove(locale_path)
+            if source_path:
+                os.remove(source_path)
+
+        zf.close()
+
+        content = s.getvalue()
+        filename = project.slug + '.zip'
+
+    # Download a single file if project has more resources
```
jotes: Code suggest that it will work if project has only one resource (get_object_or_404 should throw exception).

pontoon/base/utils.py
```diff
@@ -398,17 +386,75 @@ def get_download_content(slug, code, part):
...
+
+            if not locale_path:
+                return None, None
+
+            files[resource.path] = (locale_path, source_path)
+
+        s = StringIO.StringIO()
+        zf = zipfile.ZipFile(s, "w")
+
+        for fpath, value in files.iteritems():
```
jotes: (Ideas for future) What would you say to move this into celery task which would generate this file after changes in translations? Zipping few big files can take some time and it will still be more faster if we generate them once for some time than per every request. We could also use some fancy CDN like cloudflare which could be benefit for end-users.

pontoon/base/utils.py
```diff
@@ -398,17 +386,75 @@ def get_download_content(slug, code, part):
...
+            zf.write(locale_path, fname)
+
+            os.remove(locale_path)
+            if source_path:
+                os.remove(source_path)
+
+        zf.close()
+
+        content = s.getvalue()
+        filename = project.slug + '.zip'
```
jotes: I'm not sure if we shouldn't add locale to the name of file?

jotes:
r+ :)

pontoon/base/utils.py
mathjazz: Usually localizers download these files for personal use and they usually only localize into one locale. We also don't put locale names when a single file is downloaded.

pontoon/base/utils.py
mathjazz: On longer term, we'll need to be able to download a ZIP for bigger projects as well. Firefox and Firefox OS contain 100s of files and obviously the current approach won't scale.

I like your proposal with using celery. I guess generating a new ZIP after submitting each translation would be an overkill. So we could combine this with cron and only run the task every N minutes in case there were any changes affecting the ZIP.

Persistent storage would also help. (Or fake persistent storage - cloning all repositories on dyno startup). That way we'll have all files locally, we'll just have to update the changed ones and ZIP them.

PR #323, opened 2016-01-12
Centralize user name and email presentation

mathjazz:
We present users differently in different places. I thought it would be nice to have a centralized place to generate all these display names.

@jotes r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -100,8 +100,18 @@ def user_translated_locales(self):
 
 
 @property
+def user_name_or_email(self):
+    return self.first_name or self.email
+
+
+@property
 def user_display_name(self):
-    name = self.first_name or self.email.split('@')[0]
+    return self.first_name or self.email.split('@')[0]
+
+
+@property
+def user_display_name_and_email(self):
+    name = self.display_name
     return u'{name} <{email}>'.format(name=name, email=self.email)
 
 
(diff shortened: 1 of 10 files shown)


pontoon/base/models.py
```diff
@@ -100,8 +100,18 @@ def user_translated_locales(self):
...
+
+
+@property
 def user_display_name(self):
-    name = self.first_name or self.email.split('@')[0]
+    return self.first_name or self.email.split('@')[0]
+
+
+@property
+def user_display_name_and_email(self):
```
jotes: i would change name to the vcs_display_name to give more context.

pontoon/base/models.py
```diff
@@ -100,8 +100,18 @@ def user_translated_locales(self):
 
 
 @property
+def user_name_or_email(self):
```
jotes: Somehow, i feel we should use more generic for the methods, because they'll be easier to change in future.

jotes:
r+ but we still need add some kind of profile management for the translators.

pontoon/base/models.py
mathjazz: We also use the name on other occasions, e.g. in PO file headers.

mathjazz:
> r+ but we still need add some kind of profile management for the translators.

You mean custom user model?

pontoon/base/models.py
mathjazz: You mean more generic method names?

jotes:
@mathjazz i'm not sure if user can change first_name on the our side. However, it's probably connected with our transition from the persona to self-hosted accounts.

mathjazz:
They can change their first_name on the /profile page.

PR #326, opened 2016-01-13
Fix bug 1193860: escape quotes in some DTD files

mathjazz:
Some DTD files require quotes to be escaped. This is my attempt to do so.

@jotes or @Osmose r?

The curly quotes (‘ and “) don't get escaped for some reason. Any idea why? I'll add tests.

### DIFF
--- a/pontoon/sync/formats/silme.py
+++ b/pontoon/sync/formats/silme.py
@@ -4,10 +4,11 @@ Parser for silme-compatible translation formats.
 """
 import codecs
 import os
+import silme
+
 from collections import OrderedDict
 from copy import copy
 
-import silme
 from silme.format.dtd import FormatParser as DTDParser
 from silme.format.ini import FormatParser as IniParser
 from silme.format.inc import FormatParser as IncParser
@@ -78,6 +79,9 @@ class SilmeResource(ParsedResource):
         self.source_resource = source_resource
         self.entities = OrderedDict()  # Preserve entity order.
 
+        # Bug 1193860: unescape quotes in some files
+        self.escape_quotes_on = 'mobile/android/base' in path and parser is DTDParser
+
         # Copy entities from the source_resource if it's available.
         if source_resource:
             for key, entity in source_resource.entities.items():
@@ -103,6 +107,10 @@ class SilmeResource(ParsedResource):
         current_order = 0
         for obj in self.structure:
             if isinstance(obj, silme.core.entity.Entity):
+
+                if self.escape_quotes_on:
+                    obj.value = self.unescape_quotes(obj.value)
+
                 entity = SilmeEntity(obj, comments, current_order)
                 self.entities[entity.key] = entity
                 current_order += 1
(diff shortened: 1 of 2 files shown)


pontoon/sync/formats/silme.py
```diff
@@ -78,6 +85,9 @@ def __init__(self, parser, path, source_resource=None):
         self.source_resource = source_resource
         self.entities = OrderedDict()  # Preserve entity order.
 
+        # Bug 1193860: unescape quotes in some files
+        self.escape_quotes_on = 'mobile/android/base' in path and parser is DTDParser
```
Osmose: Is there any better heuristic for this? Is it bad if we escape all quotes in all DTD files?

pontoon/sync/formats/silme.py
```diff
@@ -118,6 +132,15 @@ def __init__(self, parser, path, source_resource=None):
     def translations(self):
         return self.entities.values()
 
+    def escape_quotes(self, value):
+        def escape(match):
+            return '\\' + match.group(0)
+
+        return re.sub(r'(?<!\\)[' + QUOTES + r']', escape, value)
```
Osmose: Could you, instead of regex, simply do a replace each quote variant? I'm not sure if that'd be faster, slower, or easier to maintain. This might be fine.

Osmose:
r+, nice work! I think you should really avoid the keying on the translation file path, and I suspect you could do simpler than a regex, but I trust your judgement on whether to actually do those things.

pontoon/sync/formats/silme.py
mathjazz: Yeah, that was my original idea (do it for all .dtd files), but it's sadly not possible:
https://dxr.mozilla.org/mozilla-central/source/python/compare-locales/compare_locales/checks.py#389

We have to do it the ugly way. One day we should get rid of the hardcoded path and move this to project config.

pontoon/sync/formats/silme.py
mathjazz: Yeah, it looks much simpler with str.replace().

mathjazz:
Thanks. Turns out we don't need to escape curly quotes. I'll add tests tomorrow and then merge.

gion-andri:
Maybe I misunderstand your patch or how DTDs work, in this case just ignore this comment. 

But if I understand correct your patch, you always escape the quotes in the DTD files with **\'**. I think, in DTD files, witch use a single quote to delimit the String (like search_strings.dtd), this does not work. I that case, the quote has to be replaced with something like **\&apos;**. 

Background: 
This commit: 
http://hg.mozilla.org/releases/l10n/mozilla-aurora/rm/diff/6ac304b05f25/mobile/android/base/search_strings.dtd
caused these compare-locales errors:
https://l10n.mozilla.org/dashboard/compare?run=563338
This commit fixed the issue:
http://hg.mozilla.org/releases/l10n/mozilla-aurora/rm/diff/d7a76071741e/mobile/android/base/search_strings.dtd
Or maybe this is just a compare-locales issue.

mathjazz:
Good point @gion-andri!

Could you please check if it looks better now?

gion-andri:
This looks way better, thanks!

jotes:
r+

On Thu, Jan 14, 2016 at 9:04 AM, Gion-Andri Cantieni <
notifications@github.com> wrote:

> This looks way better, thanks!
>
> —
> <https://github.com/mozilla/pontoon/pull/326#issuecomment-171567174>.
>



-- 
Kind regards

​Jarek "jotes" Śmiejczak,
​P​

PR #327, opened 2016-01-13
Fix bug 1235893: Don't approve suggestions on delete

mathjazz:
(Diff looks big because block alignment changed several times)

Do not automatically approve next translation when deleting the active
one. Not just when in suggest mode like the bug description suggests,
but in general.

Automatic approval is dangerous and from now on we require localizers
to explicitly approve one of the remaining suggestions. To avoid any
surprises.

@jotes r?

### DIFF
--- a/pontoon/base/static/css/translate.css
+++ b/pontoon/base/static/css/translate.css
@@ -1161,7 +1161,7 @@ p#sign-in-required > a#sidebar-signin {
   cursor: pointer;
   padding-left: 10px;
   padding-right: 10px;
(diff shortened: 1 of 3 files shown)


jotes:
​r+​

On Wed, Jan 13, 2016 at 11:43 PM, Matjaž Horvat <notifications@github.com>
wrote:

> —
> <https://github.com/mozilla/pontoon/pull/327>.
>




-- 
Kind regards

​Jarek "jotes" Śmiejczak,
​P​

PR #329, opened 2016-01-15
Bug 1228038: Switch positions of locale and project main menus

mathjazz:
Switching locale and project menu order in the main menu of the translate
view has severl consequences.

1. All locales are immeditally accessible through the locales menu.
Previously only locales for the currently selected projects were
displayed, making it hard to find your locale if it wasn't on the list.

2. We no longer need to make AJAX requests for all projects to get the
list of available locales, parts and stats. It made the translate page
block other AJAX requests for around 10 seconds after page was loaded.
We only load data for the currently translated locale and then load
data for other locales in and when they are selected. It's highly likely
that localizers will changed locales less often than they change projects.

3. Requests locale (for a project) becomes request project (for locale).
That means the request form has been moved to the project menu and to
the locale page. It allows localizers to see a full list of projects
that's not available for their locale yet.

4. Main menu order now matches the order of the URL structure.

@jotes r?

That looks heavier than it is. It's mostly replacing `locale` with `project` and vice versa. :)


pontoon/base/models.py
```diff
@@ -235,6 +235,77 @@ def get_latest_activity(self, project=None):
...
+                'resource__path',
+                'resource__entity_count',
+                'fuzzy_count',
+                'translated_count',
+                'approved_count',
+            )
+
+        details = {}
+
+        if p:
```
jotes: personally i would change p in to `project`, verbosity won't hurt (i guess)

jotes:
r+wc -> tests are broken :(

mathjazz:
Thanks for the reminder. Forgot to fix them before pushing.

PR #331, opened 2016-01-21
Fix bug 1216516: Load all helper requests at once

mathjazz:
Loads all helper requests (history + machinery + locales) at once and displays a number of results in a tab.

@jotes r?


jotes:
It look neat when all tabs load immediately :) r+

PR #332, opened 2016-01-26
Fix bug 1228038: Request multiple projects

mathjazz:
This allows requesting multiple projects at once. Also adds some code reuse.

Other improvements are possible, but I find them premature before we decide what to do with the frontend code.

@jotes r?


pontoon/base/views.py
```diff
@@ -997,21 +996,25 @@ def save_user_name(request):
...
+    """Request projects to be added to locale."""
+    slug_list = request.POST.getlist('projects[]')
     locale = get_object_or_404(Locale, code__iexact=locale)
 
+    # Validate projects
+    project_list = Project.objects.available().filter(slug__in=slug_list)
+    if not project_list:
+        return HttpResponseBadRequest('Bad Request: Non-existent projects specified')
+
+    projects = ''
```
jotes: Suggestion: projects = ''.join('{} ({})\n'.format(p.name, p.slug) for p in project_list) . I heard somewhere that .join is faster than a loop with concatenation. However, readability is discussable here.

pontoon/base/views.py
```diff
@@ -997,21 +996,25 @@ def save_user_name(request):
...
+    for project in project_list:
+        projects += '{project} ({slug})\n'.format(project=project.name, slug=project.slug)
+
     if settings.ADMINS[0][1] != '':
         EmailMessage(
```
jotes: A little bit general question: what will happen if e.g. smtp fails? Can we recover a list of requested locale?

pontoon/base/views.py
```diff
@@ -997,21 +996,25 @@ def save_user_name(request):
...
-
-    project = get_object_or_404(Project, slug=slug)
+def request_projects(request, locale):
+    """Request projects to be added to locale."""
+    slug_list = request.POST.getlist('projects[]')
     locale = get_object_or_404(Locale, code__iexact=locale)
 
+    # Validate projects
+    project_list = Project.objects.available().filter(slug__in=slug_list)
+    if not project_list:
```
jotes: At some point we'll have to add a form here ;-)

jotes:
I tested it locally. I think we could add some pagination in the future, but besides that it works fine. r+

pontoon/base/views.py
mathjazz: Yeah, we could catch the exception.

PR #333, opened 2016-01-27
Optimize latest activity DB requests

mathjazz:
I realized `Project.objects.select_related('latest_translation')` is never used, because `p.get_latest_activity(locale)` actually ends up accessing `latest_translation` property of the `ProjectLocale` object (not `Project`).

So this PR collects the appropriate data and sends it to the template. It reduces DB requests by 40% on the locale page.

@Osmose @jotes Before I do something similar on the project page, I wonder if there's a more elegant way of doing this.

For more context, I noticed this while preparing for https://bugzilla.mozilla.org/show_bug.cgi?id=1214392.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -242,7 +242,7 @@ class Locale(models.Model):
         )
 
     def available_projects_list(self):
-        """Get latest activity for project and locale if provided."""
+        """Get a list of available project slugs."""
         return list(
             self.project_set.available().values_list('slug', flat=True)
         )
@@ -260,14 +260,7 @@ class Locale(models.Model):
         return Locale.cldr_id_to_plural(self.cldr_plurals_list()[plural_id])
 
     def get_latest_activity(self, project=None):
-        """Get latest activity for project and locale if provided."""
-        if project is None:
-            if self.latest_translation is not None:
-                return self.latest_translation.latest_activity
-            else:
-                return None
-        else:
-            return ProjectLocale.get_latest_activity(project, self)
+        return ProjectLocale.get_latest_activity(self, project)
 
     def parts_stats(self, project):
         """Get locale-project pages/paths with stats."""
@@ -436,14 +429,7 @@ class Project(models.Model):
             return repo
 
     def get_latest_activity(self, locale=None):
-        """Get latest activity for project and locale if provided."""
-        if locale is None:
-            if self.latest_translation is not None:
-                return self.latest_translation.latest_activity
-            else:
-                return None
-        else:
-            return ProjectLocale.get_latest_activity(self, locale)
+        return ProjectLocale.get_latest_activity(self, locale)
 
     def serialize(self):
         return {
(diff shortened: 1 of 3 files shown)


pontoon/base/views.py
```diff
@@ -73,19 +74,25 @@ def home(request):
 def locale(request, locale):
     """Locale view."""
     l = get_object_or_404(Locale, code__iexact=locale)
-
-    projects = (
-        Project.objects.available()
-        .select_related('latest_translation')
```
Osmose: Would `prefetch_related` to fetch the ProjectLocale objects, fed with a `ProjectLocale` queryset with the `select_related` on it work instead?

pontoon/base/views.py
```diff
@@ -73,19 +74,25 @@ def home(request):
...
+    projects = Project.objects.available().order_by("name")
 
     if not projects:
         raise Http404
 
+    project_locales = (
+        ProjectLocale.objects.filter(locale=l, project__in=projects)
+        .select_related('project')
+        .select_related('latest_translation__user')
+    )
```
jotes: Isn't this equivalent to:
```
latest_translations = dict(
        ProjectLocale.objects.filter(locale=l, project__in=projects)
        .select_related('project')
        .select_related('latest_translation__user')
       .values_list('project__slug', 'latest_translation')
    )
```
?

pontoon/base/views.py
jotes: We could try this.

pontoon/base/views.py
mathjazz: No, dict value needs to be the Translation instance, not the pk.

mathjazz:
This is now ready for a review. It prefetches all Translation and User data needed in latest activity columns on 4 pages:
- /projects
- /teams
- locale page (e.g. /de)
- project page (e.g. /projects/pontoon-intro)

@Osmose or @jotes r?

pontoon/base/models.py
```diff
@@ -267,7 +267,13 @@ def get_latest_activity(self, project=None):
             else:
                 return None
         else:
-            return ProjectLocale.get_latest_activity(project, self)
+            if hasattr(self, 'fetched_project_locales'):
```
jotes: somehow i Feel we should document/marke this "remotely" initialized properties, because they're making this code magical. However, i'm not sure where's a good place for such docs. Any advices would be appreciated :)

pontoon/base/models.py
```diff
@@ -443,7 +449,13 @@ def get_latest_activity(self, locale=None):
             else:
                 return None
         else:
-            return ProjectLocale.get_latest_activity(self, locale)
+            if hasattr(self, 'fetched_project_locales'):
```
jotes: I would refactor get_latest_activity into some kind of mixin, to avoid unecesssary code duplication.

pontoon/base/models.py
```diff
@@ -1097,7 +1112,20 @@ def get_locales_with_stats():
 
 def get_locales_with_project_stats(project):
     """Add chart data to locales for specified project."""
-    project_locales = project.locales.all()
+    project_locales = (
```
jotes: I would move this into separate mixin and just call it here, and below.

jotes:
@mathjazz r+ awesome :+1: I only added my humble and "nitsy" comments about code structure.

pontoon/base/models.py
mathjazz: I agree. And I also don't know what's the best practice around this. :) I'm trying to make custom attribute names start with `fetched_` so they are at least easier to spot.

pontoon/base/models.py
mathjazz: This function will go away soon anyways.

mathjazz:
Updated.

PR #334, opened 2016-01-27
Bug 12820001: Login page and change password form.

jotes:
@mathjazz Hi, i'm sorry but i had to take some break from coding progressive loading of entities.
Could you look at this? I would gladly request your help, because i totally lack of css-skills.
I think it would be a good start for our migration.

### DIFF
--- a/pontoon/accounts/backends.py
+++ b/pontoon/accounts/backends.py
@@ -1,0 +1,23 @@
+"""
+Contains authorization backends for accounts in the pontoon.
+"""
+
+from django.contrib.auth.backends import ModelBackend
+from django.contrib.auth.models import User
+
+
+class EmailModelBackend(ModelBackend):
+    """
+    Default Model backend supports new structure of User models. Unfortunately,
+    migration would require a few manual and potentially dangerous moves to migrate
+    database schema.
+    However, all what we currently need is an authentication via Email field.
+    """
+    def authenticate(self, username=None, password=None, **kwargs):
+        try:
+            user = User.objects.get(email=username)
+            if user.check_password(password):
+                return user
+        except User.DoesNotExist:
+            User().set_password(password)
+        return None
--- a/pontoon/base/static/js/main.js
+++ b/pontoon/base/static/js/main.js
@@ -460,10 +460,14 @@ $(function() {
   // Profile menu
   // Register handlers to prep django-browserid before binding menu.
   django_browserid.registerWatchHandlers(function() {
-    function signIn() {
+    function signIn(url) {
       Pontoon.startLoader();
       django_browserid.login().then(function(verifyResult) {
-        window.location.reload();
+        if (url) {
+          window.location = url;
+        } else {
+          window.location.reload();
+        }
       }, function(jqXHR) {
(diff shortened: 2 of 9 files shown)


mathjazz:
I'd also like to hear what @Osmose thinks of this approach.

So this is actually a bugfix to replace Persona:
https://bugzilla.mozilla.org/show_bug.cgi?id=1218794

We thought about using Firefox Accounts first, but it shares some drawbacks with Persona like:
* It loads external JS, which is slow.
* It uses popups, which are slow.
* It uses popups, which are... popups.

I'm leaning towards the custom authentication solution with a classic registration/sign in form with email & password. Later on we can still add things like FxA, GitHub login etc.

I think this PR is a step in the right direction. The following pieces are needed:
* Removal of Persona/BrowserID.
* Registration form.
* Migration UX for existing users (e.g. show migration page saying "hey, we're replacing persona with custom authentication solution, please add your password" and block all pages until password is set).

Osmose:
Going back to usernames and passwords is... well, it's okay, I guess. Storing passwords is a big deal, and before deploying any code that does so you'll want to sit down with the security team and chat about what concerns they have with us storing them and how we handle them.

Worrying about popups and remote JS are weak reasons to shy away from Firefox Accounts IMO. Do they not have a redirect flow instead of a popup flow?

And if you do still choose to go with passwords, I'd recommend looking into external libraries that handle this stuff rather than rolling your own, which is generally a recipe for disaster. Although in this case a lot of the heavy handling is being done by Django, which is not quite "rolling your own".

jotes:
Personally, I'm not so sure if that's good direction to store passwords/manage in the pontoon.

Beside obvious security concerns, as a developer/contributor I would love to see authentication via github in pontoon. I don't have any specific research about which auth-vendor is more popular, but i guess that github account is a something that every kind of contributor will need at some point.

I totally agree with @osmose, I want to use reliable libraries where it's possible to avoid any kind of possibility of data leak.

Also, maybe we could ask @ l10n mailing lists which vendor would be more preferred by the contributors?

mathjazz:
I understand concerns regarding storing passwords.

Delegating that functionality to a third-party solution has some drawbacks, too:

1. If it goes away, we're in the same position we're right now.
2. If it's down, Pontoon authentication is down.

If we choose to go with the third-party solution, Firefox Accounts is a better choice than GitHub:

* Being developed by Mozilla makes it more trustworthy.
* Being open means we can participate in the development.
* Many Mozilla services are adopting/using it, so users might already have one.
* Many localizers aren't developers and don't use GitHub.

I can imagine using GitHub as an alternative way of authentication one day, but not as the primary/only one. Requiring Pontoon users to get a GitHub account first doesn't make much sense.

jotes:
@mathjazz I can start looking into integration with FxA if that's our final decision. Django supports multiple authentication backends so there shouldn't be a big problem with existing FxA and our own.

mathjazz:
This was planned for Q2, because we wanted to wait for other Persona 2 FxA early adopters and learn from their mistakes. I'd still prefer to wait a bit longer before things get clearer.

mathjazz:
This is now tracked under #436.

PR #336, opened 2016-02-02
Bug 1214392 denormalize stats

mathjazz:
@Osmose Since you worked on #217, I wanted to check with you if this isn't fundamentally flawed. No need for an in-depth review yet.

What does it do?
* Adds denormalized stats fields to Project, Locale and ProjectLocale objects (in addition to Stats).
* Populates these fields with data migration and on every translation save/delete/upload and sync.
* Refactors charts in templates a little bit (reuses code via macro).
* Removes unused code: Transifex fields and update_entity_count().

What remains to be done (some things could be deferred to a later PR)?
- This might be a good opportunity to rename Stats to TranslatedResource.
- Stats object should use same stats field names as other objects.
- I guess pontoon.sync.core:update_resources() will be slow.
- Chart calculations could be moved from template to view.
- We could update latest translation and stats in the same place.
- We should update latest translation and stats also on project delete/disable/changelocales (pontoon.base.signals).


pontoon/base/models.py
```diff
@@ -354,6 +358,12 @@ class Project(models.Model):
     # Whether this project has changed since the last sync.
     has_changed = models.BooleanField(default=False)
 
+    # Aggregated stats for this project.
+    total_strings = models.PositiveIntegerField(default=0)
```
Osmose: I've never been able to decide if aggregated fields should be prefixed with `aggregate_` or something. It's super-verbose, but it's also explicit.

pontoon/base/models.py
```diff
@@ -496,6 +515,42 @@ def get_latest_activity(cls, self, extra=None):
...
+            locale = self if isinstance(self, Locale) else extra
+            project_locale = utils.get_object_or_none(ProjectLocale, project=project, locale=locale)
+
+            if project_locale is not None:
+                chart = cls.get_chart_dict(project_locale)
+
+        return chart
+
+    @classmethod
+    def get_chart_dict(cls, obj):
```
Osmose: Does this need to be a dict? If not, maybe this method is unnecessary?

pontoon/base/models.py
```diff
@@ -1308,9 +1269,42 @@ def update_stats(resource, locale):
...
-    stats.translated_count = max(translated_entities.count() - approved - fuzzy, 0)
+    translated = max(translated_entities.count() - approved - fuzzy, 0)
+
+    # Calculate diffs to reduce DB queries
+    approved_strings_diff = approved - stats.approved_count
+    fuzzy_strings_diff = fuzzy - stats.fuzzy_count
+    translated_strings_diff = translated - stats.translated_count
+
+    # Stats
+    stats.approved_count += approved_strings_diff
```
Osmose: You'll want to use `F` objects to make this an atomic update.

pontoon/base/views.py
```diff
@@ -143,18 +140,38 @@ def update_group(group, name):
...
 
 
 def project(request, slug):
     """Project view."""
-    p = get_object_or_404(Project.objects.available(), slug=slug)
+    project = get_object_or_404(Project.objects.available(), slug=slug)
+
+    locales = (
+        project.locales.all()
+        .prefetch_related(
```
Osmose: Smells like a queryset method!

Osmose:
A few minor comments, but the overall shape of this looks good. I think you're on the right track!

pontoon/base/models.py
mathjazz: The data is being JSON serialized in `*_selector.html` templates (needed for the tooltip), which is why I used the `dict`.

mathjazz:
I've addressed the comments, updated tests, and made some additional cleanups.

Stats are now also updated if locale is removed from project, project or locale is deleted, or project disabled status changes.

@Osmose r?

pontoon/base/models.py
```diff
@@ -451,6 +527,12 @@ class ProjectLocale(models.Model):
     project = models.ForeignKey(Project, related_name='project_locale')
     locale = models.ForeignKey(Locale, related_name='project_locale')
 
+    # Aggregated stats for this project in this locale.
+    total_strings = models.PositiveIntegerField(default=0)
```
Osmose: These fields and the operations around them smell like an abstract model base class you can reuse for all the models in question.

pontoon/base/models.py
```diff
@@ -1073,255 +1215,73 @@ class Stats(models.Model):
         on_delete=models.SET_NULL
     )
 
+    objects = TranslatedResourceQuerySet.as_manager()
```
Osmose: `as_manager` is the bestest thing ever

Osmose:
I only had time to skim this for now; if you want a more detailed review, I can give one, it just will take a while to happen. But it looks good! Abstracting the denormalization out to a base class is the only major change I can see that would be worthwhile. Nice work!

PR #337, opened 2016-02-03
Bug 1244423: updated requirements.txt to support pip 8

jotes:
Unfortunately we have to say goodbye to old good peep.


jotes:
Okay, i think i can close this PR. Full migration is incoming ;-)

PR #345, opened 2016-02-20
Removed support for django-sha2

jotes:
Hi, @mathjazz? Could you review this patch? This change definetely would require additional testing on your local environment before deployment on production.


mathjazz:
I think we're safe to remove this - it's a Playdoh legacy. Especially given that we don't store user passwords.

@Osmose double-check?

Osmose:
Go for it!

mathjazz:
Thank you!

PR #346, opened 2016-02-21
Temporarily removed support for the secure requirements during the bu…

jotes:
@mathjazz I'll remove this hack when the issue with pip8 on the readthedocs.org will be resolved https://github.com/rtfd/readthedocs.org/issues/1843

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,8 @@
--r docs/requirements.txt
+# Temporarily we had to remove support for the secure dependencies, because
+# readthedocs.org don't support pip8.
+# More info at https://github.com/rtfd/readthedocs.org/issues/1843
+
+-r docs/safe_requirements.txt
 
 argparse==1.3.0 \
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -1,4 +1,8 @@
--r docs/requirements.txt
+# Temporarily we had to remove support for the secure dependencies, because
+# readthedocs.org don't support pip8.
+# More info at https://github.com/rtfd/readthedocs.org/issues/1843
+
+-r docs/safe_requirements.txt
```
mathjazz: You mean `docs/requirements.txt`, right?

mathjazz:
Nice work, r+wc.

requirements.txt
jotes: I wanted to leave safe requirements for developers. I can remove safe_requirements file and just leave docs/requirements.txt if that's what you suggest.

mathjazz: Whoops, that's the main requirements file, my bad!

PR #347, opened 2016-02-21
Added an context menu with suggestions for the placeables during tran…

jotes:
hi @mathjazz could you review this? This is an initial version.

### DIFF
--- a/pontoon/base/static/css/style.css
+++ b/pontoon/base/static/css/style.css
@@ -1237,3 +1237,32 @@ body > .container {
   font-style: italic;
   margin-top: 20px;
 }
+
+#placeableMenu {
+  display: none;
+}
+
+#placeableMenu.opened {
+  display: block;
+  position: absolute;
+  top: 100px;
+  background-color: #4D5967;
+  width: 200px;
+}
+
+#placeableMenu.opened ul {
+  list-style: none;
+}
+
+#placeableMenu.opened li {
+  list-style: none;
+  padding: 10px;
+}
+
+#placeableMenu.opened li.active {
+  background: #3F4752;
+}
+
+#placeableMenu.opened li a {
+  outline: none;
+}
--- a/pontoon/base/templates/translate.html
+++ b/pontoon/base/templates/translate.html
@@ -166,6 +166,18 @@
                 <td><span>Ctrl</span>&nbsp;+&nbsp;<span>Alt</span>&nbsp;+&nbsp;<span>F</span></td>
               </tr>
               <tr>
+                <td>Suggest a placeable</td>
+                <td>
+                    <span>Ctrl</span>&nbsp;+&nbsp;<span>Alt</span>&nbsp;+&nbsp;<span>P</span><br>
+                </td>
+              </tr>
+              <tr>
+                <td>Select the next placeable</td>
+                <td>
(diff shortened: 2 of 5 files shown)


pontoon/base/static/css/style.css
```diff
@@ -1237,3 +1237,32 @@ body > .container {
   font-style: italic;
   margin-top: 20px;
 }
+
+#placeableMenu {
```
mathjazz: `placeable-menu`

mathjazz: `placeable-menu`

pontoon/base/templates/translate.html
```diff
@@ -166,6 +166,18 @@
...
               <tr>
+                <td>Suggest a placeable</td>
+                <td>
+                    <span>Ctrl</span>&nbsp;+&nbsp;<span>Alt</span>&nbsp;+&nbsp;<span>P</span><br>
+                </td>
+              </tr>
+              <tr>
+                <td>Select the next placeable</td>
+                <td>
+                  <span>Tab</span>
```
mathjazz: Tab keyboard shortcut is already taken for navigation between helpers' suggestions.

We could we use Up + Down instead, like we do in menus.

pontoon/base/templates/translate.html
```diff
@@ -241,7 +253,8 @@ <h3 class="no-match"><div class="fa fa-exclamation-circle"></div>No results</h3>
     </nav>
 
     <textarea {% if not user.is_authenticated() %} readonly {% endif %} dir="auto" lang="{{ locale.code }}" id="translation" placeholder="Type translation and press Enter to save"></textarea>
-
+    <ul id="placeableMenu">
+    </ul>
```
mathjazz: Should be in one line. And have a blank line before and after, like other siblings.

mathjazz:
First of all, sorry it took so long to review this. There were other PRs of higher importance that needed to be fixed.

I've found a few bugs:
1. You can't close the menu (with Esc or click outside) if you don't want to insert anything.
2. If caret is close to the right border of textarea, the menu breaks the UI.
3. If there are multiple placeables, sometimes the top placeable is cut off.
4. If there are 2 placeables, it's unclear which one is selected.
5. If placeable is an empty string, you can't insert it from the menu.

There are two possible solutions:
1. Turn placeables menu into a regular menu by reusing code for look and feel. It should resolve all the bugs (apart from the last one).
2. Add a number next to each placeable and insert it with Ctrl + Alt + N. That way you don't need menus or external libraries. Plus, it's a better (faster) UX, because you only need to use one keyboard shortcut. Besides, most of the time there is only one placeable, so the menu doesn't make sense.

jotes:
No problem :) Thank you for your input and time.

> most of the time there is only one placeable
Yeah, i admit that this patch was mostly driven by my curiosity (as I love when a webapp supports mouseless interface for the users) than a solid UX research.

I can fix the issues you mentioned earlier or we can just decline this patch.  I guess we shouldn't keep a feature that nobody will use.

mathjazz:
I think if we choose the 2nd solution, it's best to start over.

jotes:
I totally agree. I'm closing this for now. And thank you again :)

PR #348, opened 2016-02-21
Fix bug 1224273. Added support for python format strings in the mark_…

jotes:
Hi @mathjazz could you review?


mathjazz:
Excellent! r+

PR #350, opened 2016-02-22
Block IP addresses via settings variable BLOCKED_IPS

mathjazz:
@jotes r?

### DIFF
--- a/pontoon/settings/base.py
+++ b/pontoon/settings/base.py
@@ -110,9 +110,12 @@ INSTALLED_APPS = (
     'guardian',
 )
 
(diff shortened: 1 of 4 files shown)


pontoon/settings/base.py
```diff
@@ -114,9 +114,12 @@ def path(*args):
     'guardian',
 )
 
+BLOCKED_IPS = ('188.72.116.59',)
```
jotes: os.env.get('BLOCKED_IPS', '').split(',') would be more configurable.

jotes:
r+wc

PR #351, opened 2016-02-22
Fix bug 1250303: Properly aggregate total strings

mathjazz:
We have some issues with denormalized stats not being always in sync, see bug 1250303.

This fixes that and:
* Adds a little optimization to `update_resources()`.
* Displays projects in locale dashboard even if not imported yet.

@jotes r?


jotes:
r+

PR #353, opened 2016-03-01
Fix machinery errors

mathjazz:
@jotes r?


jotes:
r+wc because tests are broken.

mathjazz:
Thanks!

PR #357, opened 2016-03-05
Fix bug 1225827. Set the lang and dir attributes on editable elements.

jotes:
hi @mathjazz, could you look at this patch?


mathjazz:
We should set lang attribute to English if we don't have a translation or when translation is deleted. And back to locale when translation is added.

But I guess leaving it as it is shouldn't do much damage, so let's observe.

r+

PR #358, opened 2016-03-06
Asynchronously load Persona to avoid blocking JS execution

mathjazz:
@jotes r?

I had to switch to Persona API, because django_browserid API was hit by popup blocker (and I didn't investigate further).

### DIFF
--- a/pontoon/base/static/js/main.js
+++ b/pontoon/base/static/js/main.js
@@ -435,50 +435,25 @@ $(function() {
   });
 
   // Profile menu
-  // Register handlers to prep django-browserid before binding menu.
-  django_browserid.registerWatchHandlers(function() {
-    function signIn() {
-      Pontoon.startLoader();
-      django_browserid.login().then(function(verifyResult) {
-        window.location.reload();
-      }, function(jqXHR) {
-        Pontoon.endLoader('Oops, something went wrong.', 'error');
-      });
+  $('#profile .menu li').click(function (e) {
+    if ($(this).has('a').length) {
+      return;
     }
+    e.preventDefault();
 
-    $('#profile .menu li').click(function (e) {
-      if ($(this).has('a').length) {
-        return;
-      }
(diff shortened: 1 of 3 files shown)


pontoon/base/static/js/main.js
```diff
@@ -435,50 +435,61 @@ $(function() {
   });
 
   // Profile menu
-  // Register handlers to prep django-browserid before binding menu.
-  django_browserid.registerWatchHandlers(function() {
```
jotes: I guess we could remove browserid/api.js from the settings/base.py

pontoon/base/static/js/main.js
```diff
@@ -435,50 +435,61 @@ $(function() {
...
+            url: info.loginUrl,
+            type: 'POST',
+            data: {
+              csrfmiddlewaretoken: csrfToken,
+              assertion: verifyResult
+            },
+            success: function(data) {
+              window.location.reload();
+            },
+            error: function(data) {
```
jotes: I think we could report any errors that are happening here. Look at: https://raygun.io/raygun-providers/javascript

pontoon/base/static/js/main.js
```diff
@@ -435,50 +435,61 @@ $(function() {
...
-    });
-
-    $('ul.links > li#sign-in').click(function (e) {
-      e.preventDefault();
-      signIn();
-    });
+    try {
+      navigator.id.request(info.requestArgs);
+    }
+    catch(ex) { }
```
jotes: The same here. Also, i think it's a case of Diaper pattern -> c2.com/cgi/wiki?PokemonExceptionHandling .

jotes:
r+ :) This works perfectly in Firefox and Chrome.

pontoon/base/static/js/main.js
mathjazz: Good idea!

PR #360, opened 2016-03-09
Allow download if asymmetric file doesn't exist

mathjazz:
@jotes r?

### DIFF
--- a/pontoon/base/utils.py
+++ b/pontoon/base/utils.py
@@ -380,7 +380,7 @@ def get_download_content(slug, code, part):
         )
         dirnames = set([locale.code, locale.code.replace('-', '_')])
         locale_path = _download_file(locale_prefixes, dirnames, resource.path)
-        if not locale_path:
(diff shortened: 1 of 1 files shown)


pontoon/base/utils.py
```diff
@@ -381,7 +381,14 @@ def get_download_content(slug, code, part):
         dirnames = set([locale.code, locale.code.replace('-', '_')])
         locale_path = _download_file(locale_prefixes, dirnames, resource.path)
         if not locale_path:
```
jotes: I would add `and not resource.is_asymmetric` here and an whole code related to is_assymetric  into this block: https://github.com/mozilla/pontoon/pull/360/files#diff-cfad39fede62fdc24a97edc6622b538dL388 I think this change would make reading of this code a little simpler.

jotes:
r+wc :sunny:

mathjazz:
Thanks, I like your proposal very much!

PR #361, opened 2016-03-09
Do not mark translation approved prematurely

mathjazz:
When approved button is clicked, we should only update the approved
icon once potential warning has been discarded.

@jotes r?


jotes:
:+1:

PR #362, opened 2016-03-10
Fix updating parts menu for subpages

mathjazz:
By renaming `resource__path` attribute (holding resource paths and subpage
names) to `title` and adding a new `resource__path` attribute holding
resource paths only.

@jotes r?


jotes:
r+

PR #365, opened 2016-03-16
Fix bug 1199489: delete obsolete translations

mathjazz:
If translation is removed from the source file, we mark it as obsolete
in Pontoon. Now we also remove them from translated files in asymmetric
formats (they are removed on merge in symmetric ones), even if no other
changes have been made in those files (like translation added).

We do so by bringing a list of obsolete entities from the `sync_project`
task to `sync_project_repo` tasks and then marking `changed_resources` and
`locales_to_commit` properly so these files get updated and commited.

### DIFF
--- a/pontoon/sync/changeset.py
+++ b/pontoon/sync/changeset.py
@@ -5,6 +5,7 @@ from bulk_update.helper import bulk_update
 from pontoon.base.models import (
     Entity,
     Locale,
+    Resource,
     Translation,
     TranslationMemoryEntry
 )
@@ -17,7 +18,7 @@ class ChangeSet(object):
     translations stored in VCS. Once all the necessary changes have been
     stored, execute all the changes at once efficiently.
     """
-    def __init__(self, db_project, vcs_project, now):
+    def __init__(self, db_project, vcs_project, now, obsolete_vcs=None):
         """
         :param now:
             Datetime to use for marking when approvals happened.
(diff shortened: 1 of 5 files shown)


pontoon/sync/changeset.py
```diff
@@ -104,6 +105,12 @@ def execute_update_vcs(self):
             # Track which translators were involved.
             self.commit_authors_per_locale[locale_code].extend([t.user for t in db_translations if t.user])
 
+        # Remove obsolete entities from VCS
+        for r in Resource.objects.filter(entities__pk__in=self.changes['obsolete_db']).distinct():
```
jotes: I would use here values_list() or defer() to save some memory. Also, we could create an method in queryset that could easily return assymetric resources, instead of calling property each time.

mathjazz:
@jotes r?

pontoon/sync/tests/test_changeset.py
```diff
@@ -374,6 +376,34 @@ def test_unchanged_resources_sync(self):
...
+
+        resource_file = MagicMock()
+        self.changeset.vcs_project.resources = {
+            self.main_db_entity.resource.path: resource_file
+        }
+
+        # Entity must be made obsolete
+        self.changeset.changes['obsolete_vcs'] = [self.main_db_entity.pk]
+
+        # Resource file format must be assymetric
```
jotes: asymmetric

jotes:
r+ :+1:

PR #335, opened 2016-02-01
Bug 1216546 progressive load

jotes:
TODO list:
Backend:
- [x] translations_count is not doing what it should.
- [x] Search has to search by more than just the source string.

Frontend:
- [x] Progress bar only shows the status of the first 40 loaded entities instead of all.
- [x] 'No results' never appears when searching or filtering entities.
- [x] When you scroll to the bottom, 'Loading entities' starts jumping (showing and hiding) and several requests are made. You also need to wait for that to complete before you can scroll back up.
- [x] Translated filter is still not doing what it should. It's returning entities containing both, suggestions and approved translations. It should only display entities with blue checkmark icons. Also, it behaves differently if you select filter multiple times in a row. What I don't understand is why after page load and simple filter selection (Suggested) get-entities request puts the first 40 displayed entities on the excludeEntities list.
- [x]  Tests are failing.
- [x]  Progress bar is still off. I haven't checked the code, but numbers don't compute. It might be related to the translated filter issue descirbed above, because translated entities aren't calculated properly. It doesn't update on add/delete/approve translation. I'm also concerned if moving the logic to backend is a good idea, because it makes at least 4 big DB queries (and additional XmlHttpRequest) every time a translation is saved/deleted/approved.
- [x] On homepage (or any project with in-place l10n support), the normal (out of context) view is loaded first, which then jumps to in-place view with the iframe.
- [x] Progress doesn't get updated when you delete a translation.
- [x] On scroll to bottom, sometimes "Loading entities" title is always visible - it doesn't disappear. (Maybe when resource file contains just above 40 entities?)
- [x] On scroll to bottom, sometimes not all entities are loaded. They only load after you select Filter - All.
- [x] I go to http://localhost:8000/sl/firefox-os/apps/bluetooth/bluetooth.properties/ with 28 translated and 1 suggested string. I select Suggested filter to only show the one suggested translation. I switch to http://localhost:8000/sl/firefox-os/apps/bluetooth/manifest.properties/ and I get all 4 suggested translations available inside the entire firefox-os project (even if none of them belongs to the manifest.properties file). If I select All from the filte menu, I still see 4 suggested strings and 4 strings that actually belong to the manifest.properties file.

PR #366, opened 2016-03-17
Bug 1216546: Progressive loading of entities

mathjazz:
This brings the effort stated in #335 as mostly backend code to frontend and adds some fixes.

**TODO**
- [x] Fix stats when All resources or subpages with multiple resources selected.
- [x] Fix homepage: flickering on load.
- [x] Fix homepage: unable to switch to other project.
- [x] Fix history (back/forward).


mathjazz:
Phew, that took me way longer than I thought! Lots of things were broken and many things should now work better than before. Here's a short summary of tasks:

* Prettier entity loading indicator.
* Removed Search tab from the Parts menu and locale-project page.
* Added All Resources menu entry instead.
* Fixed history: navigating back/forward between selected locale/project/parts.
* Fixed flickering on inplace page load, e.g. homepage.
* Fixed switching between in-place and non-in-place projects.
* Fixed has suggestions filter.
* Removed client-side filters and search.
* Do not trigger search with modifier keys etc. - only if value changes (old problem, but more visible with backend search).
* Don't show out of date editor content on search and filter (select frist entity or hide editor if no match).
* Focus & unfocus iframe properly so history navigation (back/forward) doesn't trigger in the iframe.

TODO:
- [ ] https://pastebin.mozilla.org/8864713: Search inside missing returns translated & search inside translated doesnt return anything (homepage - translated - github).

mathjazz:
@jotes r?

pontoon/base/models.py
```diff
@@ -1313,11 +1313,11 @@ def aggregate_stats(self, instance):
 
         instance.save()
 
-    def stats(self, paths, locale):
+    def stats(self, project, paths, locale):
         """
         Returns statistics for the given paths and locale.
         """
-        return self.get(locale=locale, resource__path__in=paths).stats
+        return self.get(resource__project=project, resource__path__in=paths, locale=locale).stats
```
jotes: good catch.

jotes:
:+1: :100: :dancers: :fireworks:

jotes:
r+

PR #369, opened 2016-03-25
Bug 983096: Batch operations

mathjazz:
This is not ready for a review yet.

### DIFF
--- a/pontoon/base/forms.py
+++ b/pontoon/base/forms.py
@@ -6,14 +6,22 @@ from pontoon.base.models import User
 from pontoon.sync.formats import SUPPORTED_FORMAT_PARSERS
 
 
+class NoTabStopCharField(forms.CharField):
+    widget = forms.TextInput(attrs={'tabindex': '-1'})
+
+
+class NoTabStopFileField(forms.FileField):
+    widget = forms.FileInput(attrs={'tabindex': '-1'})
+
+
 class DownloadFileForm(forms.Form):
-    slug = forms.CharField()
-    code = forms.CharField()
-    part = forms.CharField()
+    slug = NoTabStopCharField()
+    code = NoTabStopCharField()
+    part = NoTabStopCharField()
 
 
 class UploadFileForm(DownloadFileForm):
-    uploadfile = forms.FileField()
+    uploadfile = NoTabStopFileField()
 
     def clean(self):
         cleaned_data = super(UploadFileForm, self).clean()
--- a/pontoon/base/static/css/translate.css
+++ b/pontoon/base/static/css/translate.css
@@ -310,23 +310,19 @@ body > header .locale.select {
 
 .status:before {
   color: #FD5F60;
-  content: "";
+  content: "";
 }
 
 .translated .status:before {
   color: #4FC4F6;
-  content: "";
 }
 
 .fuzzy .status:before {
   color: #FED271;
-  content: "";
-  margin-left: -1px;
 }
 
 .approved .status:before {
   color: #7BC876;
-  content: "";
 }
 
 .not-translated .status:before {
@@ -458,9 +454,8 @@ body > header .notification {
   position: absolute;
   right: 15px;
   top: 70px;
-  visibility: visible;
   width: auto;
-  /* No z-index to prevent from covering the Next button */
+  /* No z-index or visibility to prevent covering the Next button */
 }
 
 body > header .notification.hide {
@@ -654,7 +649,7 @@ body > header aside p {
 #sidebar .search-wrapper .icon {
   font-size: 19px;
   left: auto;
-  right: 5px;
+  right: 10px;
   top: 12px;
 }
 
@@ -663,7 +658,7 @@ body > header aside p {
   color: #FFFFFF;
   font-size: 16px;
   font-weight: 300;
-  padding: 10px 30px 10px 43px;
+  padding: 10px 35px 10px 43px;
   width: 100%;
 
   -moz-box-sizing: border-box;
@@ -764,8 +759,9 @@ body > header aside p {
 
 #entitylist > .wrapper > ul > li:hover,
 #entitylist > .wrapper > ul > li.hovered,
-#entitylist > .wrapper > ul > li.active {
-  background-color: #4d5967;
+#entitylist > .wrapper > ul > li.active,
+#entitylist > .wrapper > ul > li.selected {
+  background-color: #4D5967;
 }
 
 #entitylist > .wrapper > ul > li .arrow {
@@ -777,6 +773,21 @@ body > header aside p {
 
 #entitylist > .wrapper > ul > li:hover .arrow {
   display: block;
+}
+
+/* Make selection area bigger and fit the entire row for easier use */
+#entitylist > .wrapper > ul > li > .status {
+  margin: -13px -13px -13px -16px;
+  padding: 13px 13px 13px 16px;
+  font-size: 16px;
+}
+
+#entitylist > .wrapper > ul > li > .status:not(.unselectable):hover:before {
+  content: "";
+}
+
+#entitylist > .wrapper > ul > li.selected > .status:before {
+  content: "";
 }
 
 #entitylist > .wrapper > ul > li p {
@@ -827,6 +838,24 @@ body > header aside p {
   padding-left: 16px;
 }
 
+#sidebar #entitylist .batch-bar {
+  display: none;
+  background: #272A2F;
+  border-top: 1px solid #5E6475;
+  border-bottom: none;
+  bottom: 0;
+  position: absolute;
+  width: 100%;
+}
+
+#sidebar.batch:not(.advanced) #entitylist .batch-bar {
+  display: block;
+}
+
+#sidebar.batch:not(.advanced) #entitylist > .wrapper {
+  bottom: 44px;
+}
+
 
 
 
@@ -854,6 +883,22 @@ body > header aside p {
   box-sizing: border-box;
 }
 
+#editor > section {
+  height: 100%;
+}
+
+#editor #batch {
+  display: none;
+}
+
+.batch #editor #single {
+  display: none;
+}
+
+.batch #editor #batch {
+  display: block;
+}
+
 #editor p {
   text-align: left;
 }
@@ -864,7 +909,7 @@ body > header aside p {
   white-space: pre-wrap;
 }
 
-#topbar {
+#sidebar .topbar {
   background: #4D5967;
   border-bottom: 1px solid #5E6475;
   line-height: 23px;
@@ -875,28 +920,30 @@ body > header aside p {
   box-sizing: border-box;
 }
 
-#topbar a {
+#sidebar .topbar a {
   color: #CCCCCC;
   float: right;
 }
 
-#topbar a#back {
+#sidebar .topbar a#back,
+#sidebar .topbar a.quit-batch-editing,
+#entitylist .topbar a.select-all {
   float: left;
 }
 
-.advanced #topbar a#back {
+#sidebar.advanced .topbar a#back {
   display: none;
 }
 
-#topbar a#previous {
+#sidebar .topbar a#previous {
   margin-right: 30px;
 }
 
-#topbar a:hover {
+#sidebar .topbar a:hover {
   color: #7BC876;
 }
 
-#topbar a span {
+#sidebar .topbar a .fa {
   padding-right: 5px;
 }
 
@@ -1072,7 +1119,7 @@ body > header aside p {
   padding-left: 1px;
 }
 
-#editor > menu {
+#editor #single > menu {
   background: #272A2F;
   border-bottom: 1px solid #5E6475;
   padding: 10px;
@@ -1131,7 +1178,7 @@ body > header aside p {
   padding: 9px 0;
 }
 
-#editor > menu button {
+#editor #single > menu button {
   background: transparent;
   border: none;
   color: #EBEBEB;
@@ -1139,9 +1186,9 @@ body > header aside p {
   text-transform: uppercase;
 }
 
-#editor > menu button[id^="save"],
-#editor > menu button#leave-anyway {
-  background: #7bc876;
+#editor #single > menu button[id^="save"],
+#editor #single > menu button#leave-anyway {
+  background: #7BC876;
   border-radius: 3px;
   color: #272A2F;
   font-weight: 600;
@@ -1149,7 +1196,7 @@ body > header aside p {
   padding: 10px 15px;
 }
 
-#editor > menu button[id^="save"].suggest {
+#editor #single > menu button[id^="save"].suggest {
   background: #4FC4F6;
 }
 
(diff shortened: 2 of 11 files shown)


mathjazz:
Batch editing brings:
* Ability for to select entities in entity list and perform actions on them in a batch editor (approve all, delete all, find & replace). It's only exposed to privileged users (Translators).
* Colored checkboxes as status icons in entitylist (multi select)
* Colored radio buttons as status icons in history tab (single select)
* Ability to select multiple entities by holding Shift
* Ability to select all entities (even if not in entity list due to progressive loading)

This PR also contains commits with some other bugfixes and improvements:
* 9db6938 Remove unused old parts menu CSS
* ef5eedb No tab stop on hidden download/upload forms
* 0e0f7bb Unify alignment of search field and editor topbar in translate view
* 992e028 & a166edb Prevent notifications from covering the Next button
* 03608fa No need for entity key in entity list 
* a830b0d Expose /manage to managers, not translators
* 0317fec Join updateTranslation() and updateEntityUI()

**Note**
While this is ready for a review (and branch out for other frontend facing bugs), there's one open issue left I'm still working on and I'll fix in a separate commit. If lots (1000s) of entities are batch-edited, request will most probably timeout on production. It takes over 20 seconds locally for all entities selected in the currently biggest project (Firefox Aurora).

There are 2 slow parts, both related to the DB:
1. Getting "active" (approved or latest) translations for each entity (on which batch operations are ran): https://github.com/mozilla/pontoon/pull/369/files#diff-d964cbd7ee26bf4860da2c98511ccb58R456
2. (Related to 1) Updating stats for all changed TranslatedResources: https://github.com/mozilla/pontoon/pull/369/files#diff-d964cbd7ee26bf4860da2c98511ccb58R527

We'll probably have to change model to make it easier to get currently active translations, because we need that operation to be quick on multiple occasions: when getting entities for the entity list, when calculating stats and here - when doing batch operations. It should probably be a separate bug, because it might get quite heavy, and in the meantime we could split big requests in multiple requests (e.g. 1 per 1000 entities).

@jotes r?

jotes:
A first round of questions/thoughts:
* delete of selected translation causes reload of  -> it would be a good thing to notify user what happened
and how many items were removed, because user could not catch that status was displayed on the button right before reload.
* progressbar display invalid counts for entities/doesn't update after mass action.
* if you click on the label of entity by the mistake, previously selected entities are unselected again and panel with mass-action vanishes.
* i guess that "reversible" batch operations will be our next feature ;-)
* we could also add a prompt("Are you sure?") for the mass delete.
* alternatively we could add a button that would sync selected translations with their latest states from the repository.
* also, if user for some reason selects all entities, we could do this the way gmail handles selecting all e-mails (that wouldn't require us to load all entities to client)

# nits/questions
* select all could be below a list of entities and the status of selected items and buttons for the actions? It's just an idea, i'm not against the current state of this feature.

pontoon/base/forms.py
```diff
@@ -7,13 +7,13 @@
 
 
 class DownloadFileForm(forms.Form):
-    slug = forms.CharField()
-    code = forms.CharField()
-    part = forms.CharField()
+    slug = forms.CharField(widget=forms.TextInput(attrs={'tabindex': '-1'}))
```
jotes: We could create our custom fields to avoid duplication of the code.

pontoon/base/forms.py
```diff
@@ -7,13 +7,13 @@
...
+    code = forms.CharField(widget=forms.TextInput(attrs={'tabindex': '-1'}))
+    part = forms.CharField(widget=forms.TextInput(attrs={'tabindex': '-1'}))
 
 
 class UploadFileForm(DownloadFileForm):
-    uploadfile = forms.FileField()
+    uploadfile = forms.FileField(widget=forms.FileInput(attrs={'tabindex': '-1'}))
```
jotes: same here.

pontoon/base/static/css/translate.css
```diff
@@ -1299,3 +1354,116 @@ p#sign-in-required > a#sidebar-signin {
 #helpers > section ul li > header button.delete:hover:before {
   color: #FD5F60;
 }
+
+
+
```
jotes: too many newlines?

pontoon/base/static/js/translate.js
```diff
@@ -271,6 +282,24 @@ var Pontoon = (function (my) {
...
+
+      $('#sidebar').addClass('batch');
+
+      $('#sidebar .batch-bar')
+        .find('.variant').toggleClass('plural', selectedCount > 1).end()
+        .find('.selected-count').html(selectedCount);
+
+      $('#batch')
+        .find('button').removeClass('loading loaded')
+          .find('.message').html('');
```
jotes: I would match the indentation with the previous statement.

pontoon/base/static/js/translate.js
```diff
@@ -271,6 +282,24 @@ var Pontoon = (function (my) {
 
 
     /*
+     * Open batch editor
+     */
+    openBatchEditor: function () {
```
jotes: Idea for the next refactorings: I guess we could create an separate object called an Editor and add proper methods like setSelectedCount. At some point it's a really hard to figure what selectors are really meaning for the application.
However, that's not something we should do in this pull request.

pontoon/base/static/js/translate.js
```diff
@@ -1076,6 +1166,133 @@ var Pontoon = (function (my) {
...
+            // Empty translations produced by replace might not be always allowed
+            } else if (data.error) {
+              message = data.error;
+              $('#batch .replace').focus();
+            }
+          },
+          error: function(error) {
+            message = 'Oops, something went wrong';
+          },
+          complete: function() {
```
jotes: It's a philosophical thing but maybe it's better to handle success/error in the complete() ? it would be easier to track (i mean read) what happens with the message.

pontoon/base/views.py
```diff
@@ -427,6 +433,106 @@ def entities(request):
     }, safe=False)
 
 
+@login_required(redirect_field_name='', login_url='/403')
```
jotes: We could use functools partial to create our version of login required.

pontoon/base/views.py
```diff
@@ -427,6 +433,106 @@ def entities(request):
...
+            translations.filter(string=find)
+            .exclude(entity__resource__format__in=['properties', 'ini', 'dtd'])
+        )
+
+        if not replace and forbidden.exists():
+            return JsonResponse({
+                'error': 'Empty translations not allowed',
+            })
+
+        # Find translations
```
jotes: I would move this whole part into TranslationManager, i guess it will be an easier to test later.

pontoon/base/views.py
```diff
@@ -427,6 +433,106 @@ def entities(request):
...
+        translations = translations.filter(string__contains=find)
+        translatedresources = translations.translated_resources()
+        translations_to_create = []
+
+        # Create translations' clones and replace strings
+        for translation in translations:
+            translation.pk = None  # Create new translation
+            translation.string = translation.string.replace(find, replace)
+            translation.user = translation.approved_user = request.user
+            translation.date = translation.approved_date = timezone.now()
```
jotes: Maybe it's a little premature optimalization but we could call this method once per batch.

pontoon/base/views.py
```diff
@@ -427,6 +433,106 @@ def entities(request):
...
+        translations.delete()
+
+    elif action == 'replace':
+        find = request.POST.get('find')
+        replace = request.POST.get('replace')
+
+        # Empty translations produced by replace might not be always allowed
+        forbidden = (
+            translations.filter(string=find)
+            .exclude(entity__resource__format__in=['properties', 'ini', 'dtd'])
```
jotes: We could move is_asymmetric property into database, it would remove redundancies like this.

pontoon/base/views.py
```diff
@@ -427,6 +433,106 @@ def entities(request):
...
+        )
+
+        if not replace and forbidden.exists():
+            return JsonResponse({
+                'error': 'Empty translations not allowed',
+            })
+
+        # Find translations
+        translations = translations.filter(string__contains=find)
```
jotes: I guess we could use defer() here, because initialization of instances of models cost a lot of cpu/memory. Alternatively  we could use values() and pack these values into Translation() objects. We could test which way will be faster.

pontoon/base/views.py
```diff
@@ -427,6 +433,106 @@ def entities(request):
...
+        translatedresources = translations.translated_resources()
```
jotes: Am i right that we could check if translations.count() == 0 and return JsonResponse({'count': 0}) here?

pontoon/base/views.py
```diff
@@ -427,6 +433,106 @@ def entities(request):
     }, safe=False)
 
 
+@login_required(redirect_field_name='', login_url='/403')
+@require_POST
+@require_AJAX
+@transaction.atomic
+def batch_edit_translations(request):
```
jotes: I'm not sure if it wouldn't be such bad idea to consider moving some parts of this view into celery tasks because there's a chance that parts like calculation of stats will take a while.

pontoon/base/templates/translate.html
```diff
@@ -218,101 +238,155 @@ <h3 class="loading"><div class="fa fa-refresh fa-spin"></div></h3>
...
-        <ul></ul>
-      </section>
 
-      <section class="other-locales">
-        <ul></ul>
-      </section>
-    </div>
+        <div class="review">
+          <h2>Review translations</h2>
+          <button id="approve-all">
```
jotes: I would create {% macro button(id, title) %}

pontoon/base/templates/translate.html
```diff
@@ -218,101 +238,155 @@ <h3 class="loading"><div class="fa fa-refresh fa-spin"></div></h3>
...
+          </div>
+          <button id="copy" title="Copy From Source (Ctrl + Alt + C)">Copy</button>
+          <button id="clear" title="Clear Translation (Ctrl + Alt + Backspace)">Clear</button>
+          <button id="save" title="Save Translation (Enter)"></button>
+        {% else %}
+          <p id="sign-in-required"><a id="sidebar-signin" href="">Sign in</a> to translate.</p>
+        {% endif %}
+      </menu>
+
+      <div id="helpers" class="tabs">
```
jotes: i guess we could refactor these tabs into jinja macros. However, maybe its a thing for the separate task.

jotes:
This is awesome feature :) i've pointed some small nits and written some ideas for the next things/improvements in future. Anyway, after i tested this locally - a whole feature works fine.
About the issues with the performance, i have only two ideas for now:
- we could created an another cache object with precalculated status of translation
- heavy calculations could be done in separate celery task.

I'll ping you if i'll figure out something better. Anyway, you have my r+wc for now.

mathjazz:
Thanks for the feedback!

>     delete of selected translation causes reload of -> it would be a good thing to notify user what happened and how many items were removed, because user could not catch that status was displayed on the button right before reload.

Yeah, we actually reload after any batch operation, because that was the easiest way not to break anything. I'll try to be smarter about it.

>     progressbar display invalid counts for entities/doesn't update after mass action.

Thanks, fixed locally, will push once you're done with the review.

>     if you click on the label of entity by the mistake, previously selected entities are unselected again and panel with mass-action vanishes.

True, but that's also the only way to clear selection in the single-column layout (not sure there's a good spot for the clear selection button). I've extended the entity checkbox selection area to window border in the coming commit, so you can move your mouse to the left of your screen and just move it vertically to select entities. With filters and keyboard shortcuts it should also be relatively easy to select entities.

>     i guess that "reversible" batch operations will be our next feature ;-)

Yeah. Should be fun. ;)

>     we could also add a prompt("Are you sure?") for the mass delete.

True, but then again we have this warning introductory text on top. If we simply show the prompt every time, we might warn users before the change and then it turns out no changes actually happen. So we'd have to be smarter, make the request, see if there are any changes pending, show the prompt and then make the request again. Maybe we could add this in the next iteration.

>     alternatively we could add a button that would sync selected translations with their latest states from the repository.

Good idea. It's just that we'd need to make big changes to sync to make this work.

>     also, if user for some reason selects all entities, we could do this the way gmail handles selecting all e-mails (that wouldn't require us to load all entities to client)

Actually we already do this. We select everything, without loading entities on the client side.

>     select all could be below a list of entities and the status of selected items and buttons for the actions? It's just an idea, i'm not against the current state of this feature.

Yeah I'm not entirely happy with the UI. We have something like this in the single-column layout (check out the homepage). Later on we'll probably add more features (buttons), so we'll need a full panel anyways.

pontoon/base/static/css/translate.css
mathjazz: :-) Probably, but I'll keep this consistent with the rest of the file.

pontoon/base/static/js/translate.js
mathjazz: Actually, if you look at the first block, there's `.end()` at the end, which means we jump a level higher. So `.variant` and `.selected-count` are siblings.

In the 2nd block, `.message` is a child of the `button`, hence the indent.

pontoon/base/static/js/translate.js
mathjazz: Agreed. Also, there are many global variable we're assigning to Pontoon object and it's hard to keep track of them. We should spread them among such objects and maybe keep them in ObjectName.data instead of the top level, so they're easier to manage.

pontoon/base/static/js/translate.js
mathjazz: Probably true, but I'm leaving it for later. In case we decide to do it, we should do it everywhere.

pontoon/base/views.py
mathjazz: I'll see what I can do with stats, celery is also an option, but probably not the most user friendly one.

pontoon/base/views.py
mathjazz: Thanks for the suggestion, I'll investigate!

mathjazz:
Thanks for the very valuable input, @jotes! I've added 3 commits addressing your comments.

**TODO**
* [ ] Do not reload the UI after batch changes are made.
* [ ] Prevent timing out on huge requests.

mathjazz:
@jotes In the last 5 commits I've addressed your comments and fixed a bug that prevented batch updates to sync (ed126eb). This is ready for a 2nd look and then I'll rebase.

Sadly I wasn't able to fix the problem with big requests. I'm thinking about filing a bug to move bigger requests to celery, but I'm not sure how the UX would look in this case. Do you have any experience with that? Until then, we could set a hard limit of e.g. max. 3000 strings to be edited at once.

jotes:
I had some experiences with celery in my previous projects.
For the starters: Javascript sends an request to server. Django view creates an async task and returns an id of the task result to the client.
Client starts periodically (alternatively via websockets) polling data from a separate resource (e.g. /task/status/?id=<id task). This resource should return a current status (or the result) of task.
If task succeeded, the /task/status resource should return the results. Javascript on the client-side can display this result in UI via some kind of a popup message/notification.

mathjazz:
Thanks again, I've implemented your comments and added an optimization for stats calculation.

I'll experiment a bit on stage and see where should I set the hard limit. Then we can rebase, merge, deploy and file a bug to move big requests to celery.

I'm also thinking about adding a model class for faster access to currently active translations, which is also needed when getting & filtering entity translations for entity list and when calculating stats. Something like this:

```
class ActiveTranslation(models.Model):
    entity = models.ForeignKey(Entity)
    locale = models.ForeignKey(Locale)
    translations = models.ManyToManyField(Translation)
```

PR #370, opened 2016-03-27
Fix bug 1201431. Remove the obsolete resources.

jotes:
Hi @mathjazz, could you review?

### DIFF
--- a/pontoon/settings/base.py
+++ b/pontoon/settings/base.py
@@ -396,6 +396,11 @@ LOGGING = {
             'class': 'logging.StreamHandler',
         },
     },
+    'formatters': {
+        'verbose': {
+            'format': '[%(levelname)s:%(name)s] %(asctime)s %(message)s'
+        },
+    },
     'loggers': {
         'django': {
             'handlers': ['console'],
@@ -409,6 +414,9 @@ LOGGING = {
         },
     }
 }
+
+if DEBUG:
+    LOGGING['handlers']['console']['formatter'] = 'verbose'
 
 ## Tests
 TEST_RUNNER = 'django_nose.NoseTestSuiteRunner'
--- a/pontoon/sync/changeset.py
+++ b/pontoon/sync/changeset.py
@@ -1,4 +1,6 @@
 from collections import defaultdict
+import logging
+import os
 
 from bulk_update.helper import bulk_update
 
@@ -10,7 +12,9 @@ from pontoon.base.models import (
     TranslationMemoryEntry
 )
 from pontoon.base.utils import match_attr
+from pontoon.sync.utils import locale_directory_path
 
+log = logging.getLogger(__name__)
 
 class ChangeSet(object):
     """
@@ -18,7 +22,7 @@ class ChangeSet(object):
     translations stored in VCS. Once all the necessary changes have been
     stored, execute all the changes at once efficiently.
     """
-    def __init__(self, db_project, vcs_project, now, obsolete_vcs=None):
+    def __init__(self, db_project, vcs_project, now, obsolete_vcs_entities=None, obsolete_vcs_resources=None):
         """
         :param now:
             Datetime to use for marking when approvals happened.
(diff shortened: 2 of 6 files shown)


pontoon/sync/changeset.py
```diff
@@ -112,10 +118,12 @@ def execute_update_vcs(self):
         ).asymmetric().values_list('path', flat=True).distinct()
 
         for path in obsolete_entities_paths:
-            changed_resources.add(resources[path])
+            # Path can't be available if e.g. resource has been removed from the repository.
```
mathjazz: This doesn't do any harm, but I'm also not sure it's neccessary given that we `update_resources()` before `update_entities()` in `core.py:sync_project()`.

pontoon/sync/changeset.py
```diff
@@ -112,10 +118,12 @@ def execute_update_vcs(self):
...
+            if path in resources:
+                changed_resources.add(resources[path])
 
         if len(obsolete_entities_paths) > 0:
-            self.locales_to_commit = self.locales.values()
+            self.locales_to_commit = set(self.locales.values())
```
mathjazz: Nice catch! I guess `self.locales.values()` doesn't contain any duplicates, but it's still nice to keep the original type.

pontoon/sync/tasks.py
```diff
@@ -13,7 +13,7 @@
     sync_project as perform_sync_project,
     serial_task,
     update_translated_resources,
-    update_translations,
+    update_translations
```
mathjazz: Why do we need that?

pontoon/sync/core.py
```diff
@@ -1,6 +1,7 @@
 from functools import wraps
 import logging
 
+
```
mathjazz: Why do we need that?

pontoon/settings/base.py
```diff
@@ -410,6 +415,9 @@ def _allowed_hosts():
     }
 }
 
+if DEBUG:
+    LOGGING['handlers']['console']['formatter'] = 'verbose'
+
```
mathjazz: Thank. You. So. Much!

pontoon/sync/changeset.py
```diff
@@ -234,6 +242,18 @@ def execute_obsolete_db(self):
...
 
+    def execute_obsolete_resources(self):
+        for path in self.changes['obsolete_resources']:
+            for locale in self.db_project.locales.all():
+                file_path = os.path.join(
+                    locale_directory_path(self.vcs_project.checkout_path, locale.code),
+                    path
+                )
+                if os.path.exists(file_path):
+                    log.debug('Removing obsolete file {} for {}.'.format(path, locale.code))
```
mathjazz: We might want to shift this to log.info.

I'd also remove all other log messages in this PR.

pontoon/sync/changeset.py
```diff
@@ -18,7 +22,7 @@ class ChangeSet(object):
     translations stored in VCS. Once all the necessary changes have been
     stored, execute all the changes at once efficiently.
     """
-    def __init__(self, db_project, vcs_project, now, obsolete_vcs=None):
+    def __init__(self, db_project, vcs_project, now, obsolete_vcs=None, obsolete_resources=None):
```
mathjazz: Maybe 'obsolete_vcs_entities' and 'obsolete_vcs_resources'?

mathjazz:
Great job, @jotes! I've only added a few comments for more or less cosmetic improvements. I tested it and it works fine!

The only thing that needs to be added is stats update. During sync we update stats in `update_translated_resources`, but we skip this step for removed Resources of course.

You could probably do that in `sync_project_repo`, where aggregated stats will have to be update for each `Project` and then for each `Locale` and `ProjectLocale`. We could make use of the `aggregate_stats` methods available in all these models.

pontoon/sync/tasks.py
```diff
@@ -69,7 +69,7 @@ def sync_project(self, project_pk, sync_log_pk, no_pull=False, no_commit=False,
 
         return
 
-    obsolete_vcs = perform_sync_project(db_project, now)
+    obsolete_vcs, obsolete_resources = perform_sync_project(db_project, now)
```
mathjazz: `obsolete_vcs_entities, obsolete_vcs_resources` (and also in the rest of the file).

mathjazz:
Rename those variables in tasks.py and we're done! Stellar job! r+wc

mathjazz:
r+

jotes:
@mathjazz i rebased everything into one commit.

PR #372, opened 2016-03-31
Fix untranslated filter

mathjazz:
Entity is untranslated if it doesn't have a single (approved, fuzzy or
suggested) translation. Previously, entity with expected_count of 1 and
2 suggestions was also treated as untranslated.

@jotes r?


jotes:
I would extend tests for this filter, just to be sure that right conditions are met. Besides that r+ :)

mathjazz:
Good point!

PR #375, opened 2016-04-08
Bug 1257812 xliff output

mathjazz:
@jotes r?

### DIFF
--- a/pontoon/sync/formats/xliff.py
+++ b/pontoon/sync/formats/xliff.py
@@ -89,10 +89,26 @@ class XLIFFResource(ParsedResource):
         for entity in self.entities:
             entity.sync_changes()
 
+        locale_code = locale.code
+
+        # TODO: should be made iOS specific
+        # Map locale codes for iOS: http://www.ibabbleon.com/iOS-Language-Codes-ISO-639.html
+        locale_mapping = {
+            'bn-IN': 'bn',
+            'ga-IE': 'ga',
+            'nb-NO': 'nb',
+            'nn-NO': 'nn',
+            'sv-SE': 'sv'
+        }
+
+        if locale_code in locale_mapping:
+            locale_code = locale_mapping[locale_code]
+
         # Update target-language where necessary.
         file_node = self.xliff_file.namespaced('file')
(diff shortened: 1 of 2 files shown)


pontoon/sync/formats/xliff.py
```diff
@@ -89,10 +89,26 @@ def save(self, locale):
         for entity in self.entities:
             entity.sync_changes()
 
+        locale_code = locale.code
+
+        # TODO: should be made iOS specific
```
mathjazz: At some point in the future we'll need to externalize these kinds of hacks. We have another one in silme.py, used in Firefox for Android. Something in the style of https://bugzilla.mozilla.org/show_bug.cgi?id=1209375.

requirements.txt
```diff
@@ -47,8 +47,8 @@ suds==0.4 \
     --hash=sha256:d5524ae523bb9e09c57bcdbd1efae2c287d20603688ea31f6020ed180a489af0
 https://github.com/jbalogh/test-utils/archive/e42e031.zip#egg=test-utils \
     --hash=sha256:8a1dbd1c7ade85206ae770c9ce5f43081a098c95f994af113f6434e900559a67
-translate-toolkit==1.13.0 \
-    --hash=sha256:dcbbf49058e4196a06e988d9dc1e762321ab0d057c4be035b84e3c11353fc2f8
+https://github.com/mathjazz/translate/archive/1.13.1.zip#egg=translate==1.13.1 \
+    --hash=sha256:0130db76e886f26ba0d6dc69c1ef2659315ea972af4aa95cda69c1e53f322707
```
mathjazz: We need to use a custom fork before the fix proposed upstream gets released:
https://github.com/translate/translate/issues/3424

mathjazz:
Perhaps @flodolo wants to have a look, too?

flodolo:
As far as I can tell everything makes sense. I also suddenly realizes that we have both bn-IN and bn-BD in the repo at this point, and it's going to be an issue.

pontoon/sync/formats/xliff.py
jotes: Yup, i guess we could move this into separate field of the Locale model.

jotes:
r+

PR #376, opened 2016-04-10
Updated kombu version to fix error with _uuid.

jotes:
Fixes `ImportError: cannot import name _uuid_generate_random`
@mathjazz could you review?


jotes:
Oh, it looks like xliff tests are broken because of latest changes on pontoon's master.

mathjazz:
r+

Thanks for the fix, I'll update the xliff serializer and tests.

PR #377, opened 2016-04-11
Bug 1217758 - Add a dropmarker to filters button

MekliCZ:


mathjazz:
Thanks for the PR. The code looks ok.

I'm not sure about the appearance though. The idea sounds good, but it doesn't look very slick. ;) And it's hard to find the position of the caret that would work with al the icons we use.

What do you think? Maybe center align the caret at the bottom of the icon (`bottom: 2px;` and `left: 19px;`)?

pontoon/base/static/css/translate.css
```diff
@@ -696,7 +707,7 @@ body > header aside p {
   color: #AAAAAA;
   display: none;
   font-size: 16px;
-  left: 43px;
+  left: 53px;
```
mathjazz: That should remain `43px`. `42px` looks even better.

MekliCZ: @mathjazz It looks better, but when there is caret, it looks really bad. But i can hide caret when menu is opened, what do you think?

MekliCZ:
@mathjazz I changed caret position to bottom of the icons, but I have to use decimals in px values for the best look on my Retina display. `18px` is not enough, `19px` is too big. I'm not sure that decimals in px are the best way, what do you think?

mathjazz:
What if you change the icon font-size?

MekliCZ:
I can't find the right combination.

mathjazz:
Could you paste a screenshot demonstrating the issue?

MekliCZ:
Caret is slightly shifted to the left.
snimek obrazovky 2016-04-11 v 22 41 11

mathjazz:
Doesn't look too bad. Or does it in real app? Could you commit?

MekliCZ:
I see it in app. Maybe it's my personal view, but I see that few pixels are wrong here.

mathjazz:
If you think it's worse than what's currently in production, we can also wontfix the bug. :)

MekliCZ:
Great idea, I can find another bug to fix as my first.

mathjazz:
Thanks for giving it a try!

PR #378, opened 2016-04-12
Better info about what translation actions do

MikkCZ:
First time I opened this function, I was a little confused about to what strings the actions would apply (if the square works as a button for opening the action menu (because it's "pop" effect steals your focus), or a checkbox.


mathjazz:
That's definitely helpful. Thanks! r+

MikkCZ:
Maybe even more can be done, but I failed to install Pontoon locally so far because of pip collisions with Kitsune.

PR #374, opened 2016-04-07
Fix bug 1243086. Added direct links to the search, filtering and the translations.

jotes:
Unfortunately, I still need some work before we'll be able to review this

* [x] rebase with latest master
* [x] load application state from the url in an one query to the backend
* [x] modify backend view to load currently edited entity

PR #383, opened 2016-04-18
Bug 1243086 direct links

mathjazz:
This is a next iteration over #374, nicely crafted by @jotes.

**Fixed Bugs**
* [x] Inplace: impossible to open entity in the sidebar (JS error on click on the entity list)
* [x] Inplace: if search parameter in the URL, it's not used (all entities visible)
* [x] Inplace: if filter parameter in the URL without any match, `Not on the current page` visible
* [x] Inplace: if any URL parameter used, the sidebar opens without resizing the iframe
* [x] Out-of-context: If no matches found, sidebar not displayed and resized properly
* [x] Main menu is broken: when you click Go, `self.updateCurrentUrl is not a function`
* [x] Out-of-context -> inplace -> Entity selected inplace and in sidebar, but sidebar closed
* [x] `Can't load selected entity` if no match is not neccessary
* [x] Investigate use of `.visible` vs `.limited`
* [x] If specified entity not on the first page it's not properly appended
* [x] If requested entity on the page, it sometimes gets duplicated in the list, due to Paginator sorting issue

**New Features**
* [x] If Go clicked: reset filters (can't load selected entity if switching project anyways)
* [x] Filter and search should not trigger getEntities()` with changes in the main menu taken into account
* [x] Filters should use same names we use in the filter list (e.g. suggested instead of translated)
* [x] Make history (back/forward) work again, by only pushing state if it changes

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -935,7 +935,7 @@ class EntityQuerySet(models.QuerySet):
             )
         )
 
-    def untranslated(self, locale):
+    def missing(self, locale):
         return self.with_status_counts(locale).filter(
             Q(approved_count=0) & Q(fuzzy_count=0) & Q(suggested_count=0)
         )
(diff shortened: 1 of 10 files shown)


mathjazz:
@jotes r?

I'm almost done, only need to fix history.

And, I just discovered an issue with sorting in Paginator that is making `entities_to_map.values_list('pk', flat=True)` return something completely different than what's expcted. And since `entities_to_map.values_list('pk', flat=True)[0]` and `entities_to_map[0].pk` are different, we sometimes append an extra (duplicate) entity to the list.

jotes:
I was so sure that i already given r+ here... I'm sorry! everything looks alright.

mathjazz:
yessssssssss :-)

PR #384, opened 2016-04-27
Fixed urls for readthedocs.

jotes:
Hi @mathjazz,
I'm not sure if you had a time to the latest mail from readthedocs but it looks like we have all links into .io just because of the security concerns.


mathjazz:
Thanks, nice catch! r+

PR #385, opened 2016-04-28
Fix bug 1233979: Make keyboard shortcuts more visible

mathjazz:
@jotes r?

### DIFF
--- a/pontoon/base/templates/translate.html
+++ b/pontoon/base/templates/translate.html
@@ -2,6 +2,7 @@
 
 {% import "widgets/checkbox.html" as Checkbox %}
 {% import "widgets/button.html" as Button %}
+{% import "widgets/keyboard_shortcut.html" as Shortcut %}
 
 {% block class %}translate{% endblock %}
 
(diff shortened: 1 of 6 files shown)


pontoon/base/templates/translate.html
```diff
@@ -283,6 +233,59 @@ <h3 class="loading"><div class="fa fa-refresh fa-spin"></div></h3>
...
+                  <tr>
+                    <td>Save Translation</td>
+                    <td><span>Enter</span></td>
+                  </tr>
+                  <tr>
+                    <td>Cancel Translation</td>
+                    <td><span>Esc</span></td>
+                  </tr>
+                  <tr>
+                    <td>Insert A New Line</td>
```
jotes: Small thing but we could create a jinja2 macro.

jotes:
r+ :-)

pontoon/base/templates/translate.html
mathjazz: Good point!

PR #387, opened 2016-05-04
Bug 1243118 - [translate] Highlight matches when searching entities

MekliCZ:


MekliCZ:
I forget to translated string.

PR #389, opened 2016-05-04
Bug 1243118 - [translate] Highlight matches when searching entities

MekliCZ:
Fixed regex problem.

### DIFF
--- a/pontoon/base/static/css/translate.css
+++ b/pontoon/base/static/css/translate.css
@@ -786,6 +786,13 @@ body > header aside p {
   padding-left: 2%;
 }
 
+#entitylist > .wrapper > ul > li p > span .searchHighlight {
+  color: #7BC876;
+  display: inline-block;
+  margin-bottom: -6px;
+  width: auto;
+}
+
 #entitylist > .wrapper > ul > li.translated p span.source-string,
 #entitylist > .wrapper > ul > li.suggested p span.source-string,
 #entitylist > .wrapper > ul > li.fuzzy p span.source-string,
--- a/pontoon/base/static/css/translate.css
+++ b/pontoon/base/static/css/translate.css
@@ -791,6 +791,7 @@ body > header aside p {
   display: inline-block;
   margin-bottom: -6px;
   width: auto;
+  white-space: pre;
 }
 
 #entitylist > .wrapper > ul > li.translated p span.source-string,
--- a/pontoon/base/static/css/translate.css
+++ b/pontoon/base/static/css/translate.css
@@ -787,7 +787,7 @@ body > header aside p {
 }
 
 #entitylist > .wrapper > ul > li p > span .searchHighlight {
-  color: #7BC876;
+  color: #FED271;
   display: inline-block;
   margin-bottom: -6px;
   width: auto;
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -41,6 +41,15 @@ var Pontoon = (function (my) {
       $('#search').val(value);
     },
 
+    highlightSearch: function(string, query) {
+      if (string.indexOf('<') == -1) {
+        string = string.replace(
+            new RegExp("(" + query.replace(/(\s+)/,"(<[^>]+>)*$1(<[^>]+>)*") + ")", "gi"), "<span class=\"searchHighlight\">$1</span>")
+            .replace(/(<span class="searchHighlight">[^<>]*)((<[^>]+>)+)([^<>]*<\/span>)/,"$1</span>$2<span class=\"searchHighlight\">$4");
+      }
+
+      return string;
+    },
 
     renderEntity: function (index, entity) {
       var self = this,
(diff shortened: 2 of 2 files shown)


pontoon/base/static/js/translate.js
```diff
@@ -54,16 +54,15 @@ var Pontoon = (function (my) {
         '" data-entry-pk="' + entity.pk + '">' +
         '<span class="status fa' + (self.user.isTranslator ? '' : ' unselectable') + '"></span>' +
         '<p class="string-wrapper">' +
-          '<span class="source-string">' + source_string + '</span>' +
+          '<span class="source-string">' + ((source_string.indexOf('<') == -1) ? source_string.replace(new RegExp("("+ Pontoon.getSearch().replace(/(\s+)/,"(<[^>]+>)*$1(<[^>]+>)*") + ")", "gi"), "<span class=\"searchHighlight\">$1</span>").replace(/(<span class="searchHighlight">[^<>]*)((<[^>]+>)+)([^<>]*<\/span>)/,"$1</span>$2<span class=\"searchHighlight\">$4") : source_string) + '</span>' +
           '<span class="translation-string" dir="auto" lang="' + self.locale.code + '">' +
-            self.doNotRender(entity.translation[0].string || '') +
+              self.doNotRender(entity.translation[0].string || '').replace(new RegExp("("+ Pontoon.getSearch().replace(/(\s+)/,"(<[^>]+>)*$1(<[^>]+>)*") +")", "gi"), "<span class=\"searchHighlight\">$1</span>").replace(/(<span class="searchHighlight">[^<>]*)((<[^>]+>)+)([^<>]*<\/span>)/,"$1</span>$2<span class=\"searchHighlight\">$4") +
```
jotes: I guess you could move this line into a separate method to improve readability.

MekliCZ:
@jotes Is it better now?

pontoon/base/static/css/translate.css
```diff
@@ -786,6 +786,13 @@ body > header aside p {
   padding-left: 2%;
 }
 
+#entitylist > .wrapper > ul > li p > span .searchHighlight {
+  color: #7BC876;
```
mathjazz: Maybe `#FED271` (Pontoon yellow) instead? Usually yellow color is used for hihglighting. Also, we use `#7BC876` in many other places and it has a special meaning ("translated").

pontoon/base/static/js/translate.js
```diff
@@ -54,16 +63,15 @@ var Pontoon = (function (my) {
         '" data-entry-pk="' + entity.pk + '">' +
         '<span class="status fa' + (self.user.isTranslator ? '' : ' unselectable') + '"></span>' +
         '<p class="string-wrapper">' +
-          '<span class="source-string">' + source_string + '</span>' +
+          '<span class="source-string">' + Pontoon.highlightSearch(source_string, Pontoon.getSearch()) + '</span>' +
           '<span class="translation-string" dir="auto" lang="' + self.locale.code + '">' +
-            self.doNotRender(entity.translation[0].string || '') +
+              Pontoon.highlightSearch(self.doNotRender(entity.translation[0].string || ''), Pontoon.getSearch()) +
```
mathjazz: You should replace `Pontoon` with `self`. And - the second parameter (`self.getSearch()`) is always the same, so it does not need to be passed - you should call it in the `highlightSearch()` function.

pontoon/base/static/js/translate.js
```diff
@@ -54,16 +63,15 @@ var Pontoon = (function (my) {
...
           '</span>' +
         '</p>' +
         '<span class="arrow fa fa-chevron-right fa-lg"></span>' +
         '</li>', self.app.win);
       li[0].entity = entity;
       entity.ui = li; /* HTML Element representing string in the main UI */
-
```
mathjazz: Could you restore this line? :)

pontoon/base/static/js/translate.js
```diff
@@ -41,6 +41,15 @@ var Pontoon = (function (my) {
       $('#search').val(value);
     },
 
+    highlightSearch: function(string) {
+      if (string.indexOf('<') == -1) {
```
mathjazz: We should highlight matches in placeables, too.

mathjazz:
Nice work @MekliCZ!

I commented on some nits. There's really just one thing that could be improved functionality-wise: we should also highlight matches inside placeables. E.g. string `WHAT`.

jotes:
@MekliCZ yup, it's better now.

MekliCZ:
@mathjazz I tried to highlight matches inside placeables, but I didn't find any way how to do it. If there is some HTML in the string, highlight can break it, e.g. searching for "pa"
`<mark class="passed">some pastring</mark>`
and after highlight
`<mark class="<span class="highlight">pa</span>ssed">some <span class="highlight">pa</span>string</mark>`

jotes:
@mathjazz @MekliCZ maybe we could move this into the server side and leave javascript as it was before?

MekliCZ:
@jotes Maybe, but then it will be work for somebody else, because my Python knowledge is poor.

mathjazz:
@jotes What would be the benefit of moving the code to backend?

It seems like we only need a stronger regex to match placeables.

MekliCZ:
@mathjazz I spent few hours over thinking about it and Googling it, but I didn't find anything. Do you have any idea? :/

jotes:
@mathjazz i guess it would be easier to apply wrapping html inside the code that is responsible for pleaceable than trying to extract these phrases in javascript.

(separate but a connected thing)
We don't have any tests for the javascript code (as we do for python) -> I'm a little concerned about the security of using complex regular expressions that we can't check automatically.
But, I guess we can create a new bug for this issue

MekliCZ:
snimek obrazovky 2016-05-04 v 22 28 14
@jotes I tried to add " ss" after variable with placeable in Python. If highlighting will be moved to backend, highlighted will be not only string in search, but also strings in translate window. So JS is the only way how to highlight text only in search.

jotes:
> highlighted will be not only string in search, but also strings in translate window
I guess it could be an acceptable feature, e.g when string is longer is longer than 50 characters.
However, It's up to @mathjazz  to decide :) I just tried to help.

MekliCZ:
Maybe you're right. As you said, it's up to @mathjazz .

mathjazz:
I'm undecided which approach (frontend or backend) is better for this particular case. I had a quick look at the placeables code and it seems the JS solution will be more elegant. And given that we already started, let's try to finish it.

We need to find a RegEx, that will exclude `mark.placeable` within strings when highlighting matches.

A question for @MekliCZ: why do these two replace calls do?
* `.replace(/(<span class="searchHighlight">[^<>]*)((<[^>]+>)+)([^<>]*<\/span>)/,"$1</span>$2<span class=\"searchHighlight\">$4");`
* `.replace(/(\s+)/, '(<[^>]+>)*$1(<[^>]+>)*')`

MekliCZ:
@mathjazz We don't need two `.replace()` calls, it's just rest of experimenting with regex.

mathjazz:
Hey @MekliCZ, we'd like to close this one. Let us know if you need any help. Alternativelly, we can also take it from where you started and finish it.

MekliCZ:
@mathjazz I'm sorry, I'll take a look at in a few days. I hope I know how to solve it, it should be only few lines of code.

MekliCZ:
Because I'm going to Sweden with no internet connection, I can't fix this in a few days as I said. It would be awesome if someone else would take this bug.

mathjazz:
Continuing in #462.

PR #391, opened 2016-05-05
Fix bug 1267554: Prevent simultaneous getEntities() requests

mathjazz:
@jotes r?


pontoon/base/static/js/translate.js
```diff
@@ -2781,6 +2790,10 @@ var Pontoon = (function (my) {
         }
         self.pushState();
       }).always(function() {
+        // Skip if other request in progres
```
jotes: is in progress (?)

jotes:
r+ :)

PR #393, opened 2016-05-06
Always close menu on click outside the menu

mathjazz:
We close menus on clicks outside. In theory. But in practice, we use
stopPropagation in many event handlers, preventing the event from
bubbling to the top of the DOM, where we catch clicks outside.

Yes, stopping event propagation can be dangerous:
https://css-tricks.com/dangers-stopping-event-propagation/

This patch removes all event.stopPropagation() calls, which prevented
menus from closing. In most cases simply removing the call was enough.
I only had to slightly refactor event handlers for helpers.

In addtion to that, I also replaced " with ' in modified event handler
selectors.

@jotes r?

### DIFF
--- a/pontoon/base/static/js/main.js
+++ b/pontoon/base/static/js/main.js
@@ -580,9 +580,8 @@ $(function() {
   });
 
   // Tabs
-  $(".tabs nav a").click(function (e) {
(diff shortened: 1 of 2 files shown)


jotes:
r+ ;-)

PR #394, opened 2016-05-09
Fix Bug 1271331 - Progress on Pontoon doesn't work properly

bychek-ru:
Now approved_share equals approved_percent.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -622,7 +622,7 @@ class ProjectLocale(AggregatedStats):
                 'approved_strings': obj.approved_strings,
                 'translated_strings': obj.translated_strings,
                 'fuzzy_strings': obj.fuzzy_strings,
-                'approved_share': round(obj.approved_strings / obj.total_strings * 100),
+                'approved_share': int(math.floor(obj.approved_strings / obj.total_strings * 100)),
                 'translated_share': round(obj.translated_strings / obj.total_strings * 100),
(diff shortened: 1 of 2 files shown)


pontoon/base/models.py
```diff
@@ -622,7 +622,7 @@ def get_chart_dict(cls, obj):
                 'approved_strings': obj.approved_strings,
                 'translated_strings': obj.translated_strings,
                 'fuzzy_strings': obj.fuzzy_strings,
-                'approved_share': round(obj.approved_strings / obj.total_strings * 100),
+                'approved_share': int(math.floor(obj.approved_strings / obj.total_strings * 100)),
```
jotes: Hi Victor, it's nice to see your patch. After a fast look: I'm pretty sure that math.floor isn't equivalent to the round in this context. I would suggest you to just cast the results of round() to integer.

mathjazz:
Thanks for the patch, @bychek-ru!

That's an interesting challenge here. The example in the bug has 64 strings total, 40 approved and 24 translated. Which gives us:
* `approved_share`: 63% (`round(0.625)`)
* `translated_share`: 38% (`round(0.375)`)

That forces the translated (blue) bar to jump out of the space available.

mathjazz:
This looks like a possible solution:
http://stackoverflow.com/questions/13483430/how-to-make-rounded-percentages-add-up-to-100

I've also opened #395 with a CSS solution, which should protect us from all progress bar overflow issues.

PR #396, opened 2016-05-12
Add support for the L20n's FTL format

mathjazz:
@zbraniecki r?

I'm leaving the original L20n file format support in the codebase for now. We should remove it eventually.

### DIFF
--- a/pontoon/sync/formats/__init__.py
+++ b/pontoon/sync/formats/__init__.py
@@ -5,7 +5,7 @@ See base.py for the ParsedResource base class.
 """
 import os.path
 
-from pontoon.sync.formats import lang, po, silme, xliff, l20n
+from pontoon.sync.formats import lang, po, silme, xliff, l20n, ftl
 
 # To add support for a new resource format, add an entry to this dict
 # where the key is the extension you're parsing and the value is a
@@ -20,6 +20,7 @@ SUPPORTED_FORMAT_PARSERS = {
     '.ini': silme.parse_ini,
     '.inc': silme.parse_inc,
     '.l20n': l20n.parse,
+    '.ftl': ftl.parse,
 }
 
 
--- a/pontoon/sync/formats/ftl.py
+++ b/pontoon/sync/formats/ftl.py
@@ -5,7 +5,7 @@ import copy
 import logging
 import os
 
-from ftl.format.parser import FTLParser
+from ftl.format.parser import FTLParser, ParseContext
 from ftl.format.serializer import FTLSerializer
 
 from pontoon.sync import SyncError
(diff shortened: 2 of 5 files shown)


pontoon/sync/formats/ftl.py
```diff
@@ -0,0 +1,148 @@
...
+                )
+
+        with codecs.open(path, 'r', 'utf-8') as resource:
+            self.structure = FTLParser().parseResource(resource.read())
+
+        def get_comment(obj):
+            return [obj['comment']['content']] if obj['comment'] else []
+
+        def parse_entity(obj, section_comment=[]):
+            if obj['value'] and obj['value']['type'] == 'Pattern':
```
zbraniecki: Why not `if obj['type'] == 'Entity'` ?

zbraniecki: Ah, I see. You're just checking if the entity has a value. A value is always a Pattern

pontoon/sync/formats/ftl.py
```diff
@@ -0,0 +1,148 @@
...
+        with codecs.open(self.source_resource.path, 'r', 'utf-8') as resource:
+            structure = FTLParser().parseResource(resource.read())
+
+        def serialize_entity(obj, entities):
+            entity_id = obj['id']['name']
+            translations = self.entities[entity_id].strings
+
+            if obj['value'] and obj['value']['type'] == 'Pattern':
+                if translations:
+                    obj['value']['elements'][0]['value'] = translations[None]
```
zbraniecki: I don't think I understand that. If you're recreating a Pattern, you need to also recreate it's source (`obj['value']['source']`), and you probably should regenerate the whole `elements` list.

zbraniecki:
r+
You seem to have some failing tests for l20n format?

pontoon/sync/formats/__init__.py
```diff
@@ -20,6 +20,7 @@
     '.ini': silme.parse_ini,
     '.inc': silme.parse_inc,
     '.l20n': l20n.parse,
+    '.ftl': ftl.parse,
```
jotes: I guess it would be good idea to add a small comment about the future of support l20n format.

mathjazz:
@zbraniecki This is now ready for a second look.

pontoon/sync/formats/__init__.py
mathjazz: Yeah. Or maybe we can just remove the old l20n format support? @zbraniecki?

zbraniecki: you can definitely remove the old format.

pontoon/sync/formats/ftl.py
```diff
@@ -103,7 +103,8 @@ def serialize_entity(obj, entities):
             translations = self.entities[entity_id].strings
 
             if translations:
-                obj['value']['elements'][0]['value'] = translations[None]
+                source = translations[None]
+                obj['value'] = ParseContext(source).getPattern().toJSON()
```
zbraniecki: wow! that's a neat approach, I like it :)

zbraniecki:
r+

PR #382, opened 2016-04-16
Bug 1214411: Skip unchanged resources on sync

jotes:
**TODO**:
* [x] Support for the git repositories
* [x] Support for the svn repositories
* [x] Support for the hg repositories
* [x]  Add support to sync all files and ignore changed files.
* [x] Fix nits in inline comments.
* [x] Change "Could not sync repo X, no locales found within." message to something like "Skipping repo X, no locales to sync found within.”
* [x] Do not run sync_project_repo on Project repository if `source_repo=True`.
* [x] Store the pushed revision on push, otherwise we don't skip the following sync (1st bug).
* [x] Update stats if only source strings added/removed (2nd and 3rd bug). See c913528.
* [x] When adding new project, sync doesn't sync anything. One has to use `--force`.
* [x] When adding a new locale to existing project, sync doesn't sync anything. One has to use `--force`. We should be able to prevent resources from skipping if TranslatedResource is not created yet: https://github.com/mozilla/pontoon/pull/382/files#diff-c3cfdee64b8608d7c2751b1df6d78acdR173.
* [x]  Write tests for the repositories interfaces
* [ ]  Write tests for the chagned files for the multi_locale=True and multi_locale=False

**Functional tests**, ran with Fennec (`source_repo=True`) and Pontoon Intro:
* [x] Add new translation in Pontoon in one file and for one locale: sync only edited locale and file.
* [x] Store commit in `last_synced_revisions` to be able to skip the following sync (previously used to read back commited revision).
* [x] Make sync work (no skip or error) on first project sync.
* [x] Make sync work after locale added to project (no skip).
* [x] Set stats when adding/removing project locales.
* [x] If source string changes in repo without changing the key, update it in Pontoon.
* [x] If source string removed in the repo, do not break (`KeyError`).

PR #397, opened 2016-05-16
Fix bug 1272958: Go to next entity if same translation exists

mathjazz:
@jotes r? This can totally wait before we merge #382.


jotes:
r+

PR #399, opened 2016-05-19
Bug 1214411: Properly join dicts of VCS and DB changes

mathjazz:
@jotes r?

This fixes the following scenario:
We merge new PO project strings, which means all files for all locales (`sl`, `fr`, `it`, `de`) are updated in the repo. We also have one new translation pending for file `file/path.po` in locale `fr` in Pontoon.

With the old code, `vcs_project.changed_files[file/path.po]` becomes `set([fr])` instead of `[sl, fr, it, de]`.


jotes:
r+, you're totally right -> .update() overrides all keys instead of merging them (which is a correct solution).

PR #402, opened 2016-05-23
Add gzip compression to django views.

jotes:
hi @mathjazz, I noticed that we don't use gzip for the django views.
More info at:
https://developers.google.com/speed/pagespeed/insights/?url=pontoon.mozilla.org


mathjazz:
r+

PR #403, opened 2016-05-23
Small tools that are helpful during the development.

jotes:
Hi @mathjazz,
Could you look at these tools?
Django debug toolbar will be helpful during tracking possible performance issues and django extensions have a pretty neat set of commands like shell_plus (loads all models automatically).

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,0 +1,11 @@
+# requirements.txt contains all dependencies required to run python in the production environment.
+-r requirements.txt
+
+django-debug-toolbar==1.4 \
+    --hash=sha256:0cbae8760f4851d480a70b72ace5b075f8191ecf899bc97427715e50fb0e90b9 \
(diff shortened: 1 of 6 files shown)


mathjazz:
@jotes I like those, thanks. Could you make sure they are only installed and enabled in dev environment?

jotes:
@mathjazz Could you look again?

requirements-dev.txt
```diff
@@ -0,0 +1,8 @@
+django-debug-toolbar==1.4 \
```
mathjazz: What if we include the `requirements.txt`file on top of this file? That way you could simply update the dev docs to pip install `requirements-dev.txt` instead of `requirements.txt`.

Usually one would put different requirements in the requirements folder, but in case of Heroku we still need to keep the `requirements.txt`, so I think that's not necessary.

jotes: You're totally right!

mathjazz:
This is very helpful, thanks! r+

PR #408, opened 2016-05-31
Add locales to repository if not yet available

mathjazz:
@jotes r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -549,6 +549,14 @@ class Project(AggregatedStats):
         else:
             return repo
 
+    @property
+    def has_multi_locale_repositories(self):
+        for repo in self.repositories.all():
+            if repo.multi_locale:
+                return True
+
+        return False
+
     @cached_property
     def source_repository(self):
         """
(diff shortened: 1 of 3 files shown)


pontoon/base/models.py
```diff
@@ -549,6 +549,14 @@ def repository_for_path(self, path):
         else:
             return repo
 
+    @property
+    def has_multi_locale_repositories(self):
+        for repo in self.repositories.all():
+            if repo.multi_locale:
+                return True
+
+        return False
```
jotes: nit: multi_locale() could be a good method for repository's queryset.

pontoon/sync/vcs/models.py
```diff
@@ -293,6 +295,33 @@ def resources_for_path(self, path):
...
+                        path = os.path.join(root, filename)
+                        if is_resource(filename):
+                            os.rename(path, source_to_locale_path(path))
+                        else:
+                            os.remove(path)
+
+            return locale_directory
+
+        else:
+            raise IOError('Directory for locale `{0}` not found'.format(locale.code))
```
jotes: Maybe It's my thing but i would add separate class for exception, something like a ```

class MissingLocaleDirectory(IOError)
    pass```
Just because it's easier to filter out (e.g. during testing) this specific case from the all other IOErrors that can happen here.

pontoon/sync/vcs/models.py
```diff
@@ -293,6 +295,33 @@ def resources_for_path(self, path):
...
+
+            # For asymmetric formats, create empty folder
+            if is_asymmetric_resource(path):
+                os.makedirs(locale_directory)
+
+            # For other formats, copy resources from source directory
+            else:
+                shutil.copytree(source_directory, locale_directory)
+
+                for root, dirnames, filenames in os.walk(locale_directory):
```
jotes: A corner case but we could skip processing if directory is empty.

jotes:
r+ :+1: Imho there's nothing blocking.

pontoon/sync/vcs/models.py
mathjazz: We already have that case covered in https://github.com/mozilla/pontoon/pull/408/files/2c4f2d574d6510ebf76d4a76f2c1790384977eb0#diff-c3cfdee64b8608d7c2751b1df6d78acdR265.

PR #412, opened 2016-06-03
FTL: Parse entire entities, including traits

mathjazz:
@stasm r?

### DIFF
--- a/pontoon/sync/formats/ftl.py
+++ b/pontoon/sync/formats/ftl.py
@@ -61,7 +61,7 @@ class L20NResource(ParsedResource):
             return [obj['comment']['content']] if obj['comment'] else []
 
         def parse_entity(obj, section_comment=[]):
-            translation = FTLSerializer().dumpEntity(obj).split(' = ', 1)[1]
+            translation = FTLSerializer().dumpEntity(obj).split('=', 1)[1].lstrip(' ')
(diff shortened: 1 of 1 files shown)


pontoon/sync/formats/ftl.py
```diff
@@ -61,11 +61,12 @@ def get_comment(obj):
             return [obj['comment']['content']] if obj['comment'] else []
 
         def parse_entity(obj, section_comment=[]):
+            translation = FTLSerializer().dumpEntity(obj).split(' = ', 1)[1]
```
stasm: The string is coming from the serializer so we control it right now, but it might be prudent to consider FTL's whitespace semantics here in case the serializer chages in the future.  There might not be whitespace around `=` or there might be more than one after it.  The following examples result in the exact same value of "Foo":

    foo=Foo
    foo = Foo
    foo =          Foo

For leading whitespace to be significant and considered part of the translation it needs to be surrounded by quotes:

    foo = "         Foo"

Perhaps something like the following will be more future-proof?

    translation = FTLSerializer().dumpEntity(obj).split('=', 1)[1].lstrip()

mathjazz: Good point!

mathjazz:
@stasm second r?

stasm:
r+, nice work!

PR #413, opened 2016-06-07
Log RuntimeErrors in Sync log

mathjazz:
If the previous project or repository sync hasn't completed yet, we
prevent next project or repository sync by raising RuntimeError.
From now on we also mark project as skipped in the sync log and
repository as finished (synced). We should probably add a special
status for this use case.

Until now, project and repository sync job that were prevented due to
RuntimeError always remained in-progress in the sync log.

Also, end_repo_sync() has been moved to RepositorySyncLog.end().

@jotes r?

### DIFF
--- a/pontoon/sync/core.py
+++ b/pontoon/sync/core.py
@@ -35,12 +35,13 @@ def sync_project(db_project, now, full_scan=False):
     return changeset.changes['obsolete_db'], removed_paths
 
 
-def serial_task(timeout, lock_key="", **celery_args):
+def serial_task(timeout, lock_key="", on_error=None, **celery_args):
     """
     Decorator ensures that there's only one running task with given task_name.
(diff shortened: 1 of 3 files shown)


jotes:
r+, I'm not sure if we can do this better without bigger refactoring.

pontoon/sync/core.py
```diff
@@ -35,7 +35,7 @@ def sync_project(db_project, now, full_scan=False):
     return changeset.changes['obsolete_db'], removed_paths
 
 
-def serial_task(timeout, lock_key="", **celery_args):
+def serial_task(timeout, lock_key="", on_error=None, **celery_args):
```
jotes: I would add docstring about this parameter.

mathjazz:
Deployed.

PR #415, opened 2016-06-08
Only raise Http404 when content is None

stasm:
There are valid scenarios when `content` can be an empty string.  For instance,
this may happen when there hasn't been any translations made to the file that's
supposed to be downloaded.

@mathjazz r?


mathjazz:
Thanks, r+.

PR #416, opened 2016-06-08
Bug 1278826 - Text overflow in two places

bychek-ru:
I see first part of this bug only in Webkit browsers. Commit fixes it.
Possible fix of second one:
1)consider such strings as untranslated, left span with class "translation-string" emply
2)add class for parent "half_translated", and assign "source-string" width 49% if parent has class "half_translated"


bychek-ru:
First one 
Second one

mathjazz:
Thank you for your patch, @bychek-ru! r+

PR #417, opened 2016-06-09
Fix bug 1225302: Add filter by translation author

mathjazz:
@jotes r?

Authors are listed in the filter menu. Max height is set to fit 3 authors, then you need to scroll. If there aren't any translations submitted by Pontoon users, we don't show the filter. We update the authors list on translate page load, translation submission/deletion, batch operations and file upload.

I haven't found a much better UX solution for the authors selector. Filter menu is getting huge. Also, we'll be adding date/time (bug 1225305) and errors/checks (bug 1237667) filters, which will make it even bigger.

We should probably refactor filter UI as part of 1243115. We might also be able to merge it with the progress bar menu.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -582,6 +582,13 @@ class Project(AggregatedStats):
             resource__entities__obsolete=False
         ).distinct().aggregate_stats(self)
 
+    def parts_to_paths(self, paths):
+        try:
+            subpage = Subpage.objects.get(project=self, name__in=paths)
+            return subpage.resources.values_list("path")
+        except Subpage.DoesNotExist:
+            return paths
+
 
 class ProjectLocale(AggregatedStats):
     """Link between a project and a locale that is active for it."""
@@ -1020,6 +1027,9 @@ class EntityQuerySet(models.QuerySet):
             approved_count=F('expected_count')
         )
 
+    def authored_by(self, locale, email):
+        return self.filter(translation__locale=locale, translation__user__email=email)
+
     def untranslated(self, locale):
         return self.with_status_counts(locale).exclude(Q(approved_count=F('expected_count')))
 
@@ -1148,6 +1158,9 @@ class Entity(DirtyFieldsMixin, models.Model):
             elif filter_type == 'unchanged':
                 entities = self.objects.unchanged(locale)
 
+            elif filter_type in Translation.authors(locale, project, paths).values_list('email', flat=True):
+                entities = self.objects.authored_by(locale, filter_type)
+
             else:
                 raise ValueError(filter_type)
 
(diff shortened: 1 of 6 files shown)


pontoon/base/models.py
```diff
@@ -1148,6 +1158,9 @@ def for_project_locale(self, project, locale, paths=None, filter_type=None,
             elif filter_type == 'unchanged':
                 entities = self.objects.unchanged(locale)
 
+            elif filter_type in Translation.authors(locale, project, paths).values_list('email', flat=True):
```
jotes: What would you say to change this DB-call for the regexp email matching? Could be a little bit faster.

pontoon/base/models.py
```diff
@@ -1309,6 +1317,28 @@ class Translation(DirtyFieldsMixin, models.Model):
...
+
+        if paths:
+            paths = project.parts_to_paths(paths)
+            translations = translations.filter(entity__resource__path__in=paths)
+
+        authors = User.objects.filter(translation__in=translations).distinct()
+
+        if serialize:
+            authors_list = []
+            for a in Translation.authors(locale, project, paths):
```
jotes: Isn't `Translation.authors(locale, project, paths)` the same as `authors`?

pontoon/base/models.py
```diff
@@ -1309,6 +1317,28 @@ class Translation(DirtyFieldsMixin, models.Model):
...
+    def authors(self, locale, project, paths, serialize=False):
+        translations = Translation.objects.filter(entity__resource__project=project, locale=locale)
+
+        if paths:
+            paths = project.parts_to_paths(paths)
+            translations = translations.filter(entity__resource__path__in=paths)
+
+        authors = User.objects.filter(translation__in=translations).distinct()
+
+        if serialize:
```
jotes: I feel that we shouldn't mix up data sources with serialization methods

pontoon/base/static/js/translate.js
```diff
@@ -1953,6 +1964,27 @@ var Pontoon = (function (my) {
...
+      $('#filter').find('.authors, .for-authors').toggle(this.authors.length > 0);
+      var authors = $('#filter .authors ul').empty();
+
+      $.each(this.authors, function() {
+        authors.append('<li data-type="' + this.email + '">' +
+          '<figure class="author">' +
+            '<img class="rounded" src="' + this.gravatar_url + '">' +
+            '<figcaption>' +
+              '<p class="name">' + this.display_name + '</p>' +
+              '<p class="email">' + this.email + '</p>' +
```
jotes: I know It's a bigger topic but maybe we should change .email into username to protect contributors privacy? At some point of course.

jotes:
r+ :-)

pontoon/base/models.py
mathjazz: Yes, that's a possible optimization. It's also slightly less powerful than this check. For now I'll leave it, but it's good to know we can be faster here.

pontoon/base/models.py
mathjazz: Ouch!

pontoon/base/static/js/translate.js
mathjazz: Yeah. I like the idea users don't need anything besides their email address to authenticate and be presented, but I also understand the privacy concern.

mathjazz:
Updated. Also enabled author filter in URLs.

jotes:
r+ x2

PR #421, opened 2016-06-17
Prevent double translation submissions which break stats

mathjazz:
@jotes r?

### DIFF
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -823,6 +823,28 @@ var Pontoon = (function (my) {
     /*
      * Attach event handlers to editor elements
      */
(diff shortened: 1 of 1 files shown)


pontoon/base/static/js/translate.js
```diff
@@ -823,6 +823,27 @@ var Pontoon = (function (my) {
     /*
      * Attach event handlers to editor elements
      */
+    saveTranslation: function (e) {
+      e.preventDefault();
+      var self = Pontoon,
+          entity = self.getEditorEntity(),
+          source = $('#translation').val();
+
+      $(this).off('click.save');
```
jotes: A small comment would be valuable.

jotes:
r+

PR #422, opened 2016-06-18
Fixed Manager that broke Django migrations.

jotes:
hi @mathjazz,

I discovered this issue after I tried to create new migrations. Could you apply this patch and check if you can create migration?h


mathjazz:
Thanks, r+.

One day we should probably use a custom User model.

PR #424, opened 2016-06-20
Switched pontoon-intro into official mozilla repo.

jotes:
Hi @mathjazz, could you look at this? I remember that we talked @irc about this change quite a while ago.

I decided to modify a base migration because this won't change anything on your installation or production, but will affect only new contributors/fresh databases.


mathjazz:
Thank you! r+

PR #425, opened 2016-06-22
Move move cursor to the beginning of textarea

mathjazz:
When modifying translation textarea programatically, move mouse cursor
to the beginning, because this is where users start reading from and
making changes.

@jotes r?

### DIFF
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -306,6 +306,14 @@ var Pontoon = (function (my) {
     /*
      * Update current translation length
      */
+    moveCursorToBeginning: function () {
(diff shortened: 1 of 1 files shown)


pontoon/base/static/js/translate.js
```diff
@@ -306,6 +306,14 @@ var Pontoon = (function (my) {
     /*
      * Update current translation length
      */
+    moveCursorToBeginning: function () {
```
jotes: I'd update/remove this docstring above.

jotes:
:+1:

pontoon/base/static/js/translate.js
mathjazz: Ouch. :) Thanks!

PR #426, opened 2016-06-24
Fix bug 1278826 - incorrect stats, filters and UI

bychek-ru:
A half-translated string is a string which has plular forms but not all forms translated.

- [ ] Fix stats when half-translated string is shown as suggested but not missing (maybe we need to add filter in condition nevertheless it works in this way)

- [ ] Fix missing filter that can't catch half-translated strings

- [ ] Fix breaking UI by half-translated strings

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -1541,6 +1541,7 @@ class TranslatedResource(AggregatedStats):
         fuzzy = translations.filter(fuzzy=True).count()
 
         # Plural
+        half_translated = 0
         nplurals = locale.nplurals or 1
         for e in translated_entities.exclude(string_plural=''):
             translations = Translation.objects.filter(entity=e, locale=locale)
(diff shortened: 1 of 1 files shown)


pontoon/base/models.py
```diff
@@ -1541,15 +1541,18 @@ def calculate_stats(self, save=True):
         fuzzy = translations.filter(fuzzy=True).count()
 
         # Plural
+        half_translated = 0
```
mathjazz: Maybe `partially_translated` to be more accurate?

pontoon/base/models.py
```diff
@@ -1541,15 +1541,18 @@ def calculate_stats(self, save=True):
...
         nplurals = locale.nplurals or 1
         for e in translated_entities.exclude(string_plural=''):
             translations = Translation.objects.filter(entity=e, locale=locale)
             if translations.filter(approved=True).count() == nplurals:
                 approved += 1
             elif translations.filter(fuzzy=True).count() == nplurals:
                 fuzzy += 1
+            else:
+                half_translated += 1
```
mathjazz: This `else` block will also catch strings that have suggestions for all plural forms, in which case we need to count them as suggested instead of missing. So they should not be counted as `partially_translated`.

To properly count `partially_translated` strings, you need to make sure that at least one plural form doesn't have any translations.

mathjazz:
Hey @bychek-ru, I'm curious - why have you closed this PR?

bychek-ru:
@mathjazz Because I tried to add 'suggested' to https://github.com/mozilla/pontoon/blob/master/pontoon/base/models.py#L1133 and it doesn't work. It's not enough for working. Also tried to add 'suggest' to class Translated and it says that there's no this field in BD. I'll reopen:D Maybe you know the best way to do it? I

PR #427, opened 2016-06-26
Fix bug 1235895. Allow to discard translations.

jotes:
I need to update the frontend side and add some tests before this PR will be fully reviewable.


jotes:
@mathjazz could you look at this?

pontoon/base/models.py
```diff
@@ -1418,6 +1418,16 @@ def check_latest_translation(self, instance):
             instance.latest_translation = self
             instance.save(update_fields=['latest_translation'])
 
+    def discard(self, stats=True):
```
mozilla-pontoon: Let's rename `discard` to `unapprove` everywhere in the code.

We'll replace delete with discard sometime in the future:
https://bugzilla.mozilla.org/show_bug.cgi?id=1192872

pontoon/base/models.py
```diff
@@ -1418,6 +1418,16 @@ def check_latest_translation(self, instance):
...
+        """
+        Discards translation.
+        """
+        self.approved = False
+        self.save()
+
+        TranslatedResource.objects.get(resource=self.entity.resource, locale=self.locale).calculate_stats()
+        self.entity.mark_changed(self.locale)
+
```
mozilla-pontoon: We should also remove TranslationMemory entry if it doesn't exist anywhere else.

pontoon/base/models.py
mathjazz: Let's rename `discard` to `unapprove` everywhere in the code.

We'll replace delete with discard sometime in the future:
https://bugzilla.mozilla.org/show_bug.cgi?id=1192872

pontoon/base/models.py
mathjazz: We should also remove TranslationMemory entry if it doesn't exist anywhere else.

pontoon/base/static/css/translate.css
```diff
@@ -1432,11 +1432,18 @@ p#sign-in-required > a#sidebar-signin {
...
+  color: #7BC876;
+}
+
+#helpers > section ul li > header button.unapprove::before {
+  color: #AAA;
+}
+
 #helpers > section ul li > header button.delete:before {
   content: "";
   margin-left: 5px;
```
mathjazz: Approved icon should stay green, because we use this color everywhere in the stats. It should become blue on hover to mark the state after click.

This is the diff to the original file that makes this possible:
https://gist.github.com/mathjazz/bb08e90c4faca5bb8053366f1aad0b7b

pontoon/base/static/js/translate.js
```diff
@@ -236,7 +236,7 @@ var Pontoon = (function (my) {
                       '<time class="stress" datetime="' + this.date_iso + '">' + this.date + '</time>' +
                     '</div>' +
                     '<menu class="toolbar">' +
-                      '<button class="approve fa" title="' +
+                      '<button class="' + (this.approved ? 'unapprove' : 'approve') + ' fa" title="' +
                       (this.approved ? this.approved_user ?
                         'Approved by ' + this.approved_user : '' : 'Approve') +
```
mathjazz: The title tag should be changed to `approve`/`unapprove`. `Approved by approver` can be moved to the `title` tag of the translation author (line 235).

pontoon/base/static/js/pontoon.js
```diff
@@ -531,16 +531,21 @@
...
             }
             break;
+          case "DISCARD":
+            /**
+             * Messsage is sent when the locale maintainer discards translation of the entity.
+             */
 
+            var entity = message.value.id ? Pontoon.entities[message.value.id] : $('.pontoon-editable-toolbar')[0].target.entity,
+                translationValue = message.value.translation,
+                translation = translationValue !== undefined ? translationValue : message.value;
```
mathjazz: This block is never called... And also doesn't do anything except for burning some memory and time. ;)

pontoon/base/static/js/pontoon.js
```diff
@@ -531,16 +531,21 @@
...
-            var entity = null,
+            var entity = message.value.id ? Pontoon.entities[message.value.id] : $('.pontoon-editable-toolbar')[0].target.entity,
                 translationValue = message.value.translation,
                 translation = translationValue !== undefined ? translationValue : message.value;
-            if (message.value.id) {
-              entity = Pontoon.entities[message.value.id];
-            } else {
-              entity = $('.pontoon-editable-toolbar')[0].target.entity;
-            }
+
```
mathjazz: Nice simplification!

pontoon/base/static/js/translate.js
```diff
@@ -1109,25 +1109,57 @@ var Pontoon = (function (my) {
...
+
+           self.stats = data.stats;
+           self.authors = data.authors;
+           self.updateProgress();
+           self.updateAuthors();
+           self.updateTranslation(entity, translation, data.translation);
+
+           $('#translation').val('').focus();
+           self.cachedTranslation = '';
+           self.postMessage("SAVE", data.translation);
```
mathjazz: This logic needs to change a bit, and act more like deletinga active translations (line 1241).

We need to find the newest translation for the given entity, locale and plural form and return it as `data.translation`. And then set that value into `#translation`, `cachedTranslation` and in-place translation (via `postMessage`).

mathjazz:
Thanks for the patch, nice job! There are a few things to fix, but nothing major.

jotes:
@mathjazz could you look again?

pontoon/base/static/css/translate.css
```diff
@@ -1445,19 +1445,26 @@ p#sign-in-required > a#sidebar-signin {
...
 }
 
 #helpers > section ul li > header button.delete:hover:before {
   color: #FD5F60;
-}
+
```
mathjazz: Restore `}` or CSS breaks (e.g. in batch editing view).

pontoon/base/static/js/translate.js
```diff
@@ -256,6 +256,21 @@ var Pontoon = (function (my) {
         success: function(data) {
           if (data.length) {
             $.each(data, function() {
+              var approveButtonTitle = "";
```
mathjazz: Please use single quotes.

pontoon/base/static/js/translate.js
```diff
@@ -256,6 +256,21 @@ var Pontoon = (function (my) {
...
+                } else {
+                  approveButtonTitle = 'Unapprove';
+                }
+              } else {
+                if (this.unapproved_user) {
+                  approveButtonTitle = 'Unapproved by ' + this.unapproved_user;
+                } else {
+                  approveButtonTitle = 'Approve';
+                }
+              }
```
mathjazz: This is a nice simplification! I'd simplify it even further by always showing `Approve`/`Unapprove` in the button title.

`Approved/Unapproved by` can instead by moved to the `title` tag of the translation author `<a>` element (line 285).

Also, please make sure that `<button>` and `<a> `title` tags are updated after (un)approval.

pontoon/base/views.py
```diff
@@ -514,7 +514,7 @@ def batch_edit_translations(request):
             for plural_form in range(0, locale.nplurals or 1):
                 translation_pks.add(entity.get_translation(plural_form)['pk'])
 
-    translation_pks.discard(None)
+    translation_pks.unapprove(None)
```
mathjazz: Keep using `discard`, this is a `set()` function.

pontoon/base/views.py
```diff
@@ -634,26 +634,66 @@ def get_translation_history(request):
     payload = []
     offset = timezone.now().strftime('%z')
 
+    def get_user_display(user):
+        """Shorcut function that displays user login/credentials or nothing is user is None."""
+        return (user.name_or_email if user else "")
+
```
mathjazz: Shall we move this to the `User` class?

pontoon/base/views.py
```diff
@@ -634,26 +634,66 @@ def get_translation_history(request):
...
+    translation.unapprove(request.user)
+    try:
+        previous_translation = translation.entity.translation_set.filter(
+            locale=translation.locale,
+            plural_form=translation.plural_form,
+            pk__lte=translation.pk,
+        )[0].serialize()
+    except IndexError:
+        previous_translation = translation.serialize()
```
mathjazz: The idea is to find **the latest translation for the current entity**. It can be the same translation we unapproved, or one of the newer translations.

Something like this:
```
translation = translation.entity.translation_set.filter(
    locale=translation.locale,
    plural_form=translation.plural_form,
).latest('date').serialize()
```

mathjazz:
Nice progress, there's still some things to be fixed.

And I've noticed a regression: when you have multiple suggestions and approve one, the "Are you sure you want to leave" box opens and "Same translation already exists" notification is displayed.

pontoon/base/static/css/translate.css
jotes: I'm sorry, it looks like a merge gone wrong between my laptop and my desktop. I'll fix this.

jotes:
Hi @mathjazz I've added next batch of fixes -> could you look at this again? I can't reproduce this regresssion you mentioned earlier - could you verify if it still exists?
Thanks!

pontoon/base/views.py
```diff
@@ -636,24 +636,60 @@ def get_translation_history(request):
...
+        return HttpResponseForbidden("Forbidden: You can't unapprove this translation.")
+
+    translation.unapprove(request.user)
+    try:
+        latest_translation = translation.entity.translation_set.filter(
+            locale=translation.locale,
+            plural_form=translation.plural_form,
+        ).latest('date').serialize()
+    except Translation.DoesNotExist:
+        latest_translation = translation.serialize()
```
mathjazz: Do we really need to catch the exception here? At least one translation (the one we're unapproving) must exist.

jotes: Yeah, I was actually wondered about the same thing. I'll remove it.

pontoon/base/static/js/translate.js
```diff
@@ -231,6 +231,23 @@ var Pontoon = (function (my) {
...
+      var approveButtonTitle = '';
+
+      /* Moved the logic from ternary operators to make this a little bit easier to read. */
+      if (translation.approved && translation.approved_user) {
+        return 'Approved by ' + translation.approved_user;
+      } else {
+        if (translation.unapproved_user) {
+          return 'Unapproved by ' + translation.unapproved_user;
+        } else {
+          return 'This translation requires verification & approval.';
```
mathjazz: We should also add this text if we don't have user email and display `Imported` instead - see line **286**. But only if translation is not approved. I'd change the text to: "This translation has not been reviewed yet.".

Actually, we can skip this case and leave it for the next related bug, that will introduce the **unreviewed** state, for which we'll have to change the UI anyways:
https://bugzilla.mozilla.org/show_bug.cgi?id=1229335#c5

mathjazz:
Perfect! r+

Please rebase into a single commit with a message like:
`Fix bug 1235895: Add ability to unapprove translations`

PR #428, opened 2016-07-06
Fix bug 1225305: Filter by translation time

mathjazz:
@jotes r?

This is ready for a review functionality-wide, but needs a lot of love in terms of UI and style before it hits production.


pontoon/base/models.py
```diff
@@ -1079,6 +1079,12 @@ def has_suggestions(self, locale):
     def unchanged(self, locale):
         return self.with_status_counts(locale).filter(unchanged_count=F('expected_count'))
 
+    def authored_by(self, locale, email):
+        return self.filter(translation__locale=locale, translation__user__email=email)
+
+    def between_time_interval(self, locale, start, end):
```
jotes: I think we could use range here: https://docs.djangoproject.com/en/dev/ref/models/querysets/#range

jotes:
So, I noticed following things (some of them are pretty obvious):
* We don't update the state of the datepickers after loading the app.
* I noticed something strange after applying this filter ui glitch
* I'm not the best UX expert but I  personally would choose something simpler for the timepicker e.g. a form with `<select>` widgets for the hours and minutes.
* How do we plan to support the different timezones? There will be issues with filtering if we won't e.g. convert time to the timezone available at pontoon servers.
* There's a lot of filters here :D We could create some tabs and keyboard shortcuts for them. Its only an idea which I'm not sure how we should execute.

Beside possible issues with handling timezones (e.g. When user chooses "now" or "last hour") & UI obviously, I don't see any critical bugs at this point.

jotes:
@mathjazz I hope this video will give some hints: https://vid.me/snqi

mathjazz:
OK, this is now ready for a new review. I've addressed your comments and added a few more changes, most notably:

**1. Filter menu has been slightly redesigned**
- It has become a modal window with fixed width and simplified selector (no more funky background and title replacing the search field).
- Authors are listed at the bottom, so that the rest of the filters are always on their fixed position. The entire modal can be scrolled if needed.
- Counts have been added to Translation Status filters. Which made me think that we might be able to join the progress chart and filter in the future.

**2. Date selector has been changed**
- It is prepopulated with the default filter (last hour - not sure that's the best pick).
- It is used just like any other filter, you just click on it to apply it.
- A few shortcuts are available (last hour, day, week, month) to modify the range.
- The range can be edited by manually typing into the input fields or with the help of the date picker (time is always entered manually).
- I haven't found any decent time range picker library. The current one is OKish, has some validation, but I had to patch it anyways. I'm desperately waiting for browser to finally implement this functionality with native widgets.

**3. Time zones**
- Local timezone is used in the Date & Time range selector on frontend.
- It gets converted to UTC in the URL and backend.
- So that URLs can be shared and always return the same result.

@jotes r?

jotes:
I'm on it.

jotes:
Ok, I have some notes in regard to current UI:
	* if you click on the one of the predefined time range buttons, "custom range" stays unchanged
	* click on one of predefined time ranges could automatically submit the filter state, currently you have to click twice.
	* I can't modify the state of the inputs even when i'm trying to input an custom range and because of that I can't modify the time.
	* I'm unsure if the current time ranges are intuitive, because e.g. last day you actually mean the last 24 hours instead of the whole day from 00:00 to 23:59/now(). The same thing goes for the week and the month.
	* Absolute urls aren't supported -> the state of time range filters isn't populated from passed url.

It's probably a first batch of optional suggestions :D

pontoon/base/models.py
```diff
@@ -1225,6 +1231,10 @@ def for_project_locale(self, project, locale, paths=None, filter_type=None,
             elif filter_type in Translation.authors(locale, project, paths).values_list('email', flat=True):
                 entities = self.objects.authored_by(locale, filter_type)
 
+            elif re.match('^[0-9]{12}-[0-9]{12}$', filter_type):
+                start, end = utils.parse_time_interval(filter_type)
```
jotes: I think we could return an error/ignore this filter if value is malformed somehow e.g. start > end

pontoon/base/templates/translate.html
```diff
@@ -144,22 +144,68 @@
...
         <div class="menu">
           <ul>
             <li class="horizontal-separator">By Translation Status</li>
-            <li class="all" data-type="all"><span class="status fa"></span>All</li>
-            <li class="missing" data-type="missing"><span class="status fa"></span>Missing</li>
-            <li class="fuzzy" data-type="fuzzy"><span class="status fa"></span>Fuzzy</li>
-            <li class="suggested" data-type="suggested"><span class="status fa"></span>Suggested</li>
-            <li class="translated" data-type="translated"><span class="status fa"></span>Translated</li>
+
+            <li class="all" data-type="all">
```
jotes: We could use some jinja2 macros to avoid repetition here.

pontoon/base/templates/translate.html
```diff
@@ -144,22 +144,68 @@
...
+            <li class="time-range clearfix" data-type="time-range">
+              <span class="status fa"></span>
+              <label for="from">From<input type="datetime" id="from" disabled></label>
+              <label for="to">To<input type="datetime" id="to" disabled></label>
+            </li>
+            <li class="time-range-toolbar">
+              <button class="shortcut" data-hours="1">Last Hour</button> &middot;
+              <button class="shortcut" data-hours="24">Day</button> &middot;
+              <button class="shortcut" data-hours="168">Week</button> &middot;
+              <button class="shortcut" data-hours="720">Month</button>
```
jotes: Yeah, I know this pain behind this type of widgets -> "month" sounds less confusing for the user but requires more work. Then "30 days" is less "human" but explains everything perfectly.

jotes:
Ok, I think I'm done for now. Imho, the only visibly critical thing is missing support for absolute urls. Besides that, everything looks good!

mathjazz:
>     * if you click on the one of the predefined time range buttons, "custom range" stays unchanged
>     * click on one of predefined time ranges could automatically submit the filter state, currently you have to click twice.

That's by design. ;) You are proposing two separate solutions and the one in the middle is what we currenty use:

1. On click, update value in inputs and make them editable.
 * Requires at 3 clicks before the filter is applied.
 * User can see how does each shortcut reflect in the input (understands that e.g. *Month* means 30 days).

2. On click, update value in inputs.
 * Requires at 2 clicks before the filter is applied.
 * User can see how does each shortcut reflect in the input (understands that e.g. *Month* means 30 days).

3. On click, update value in inputs and apply the filter.
 * Requires at 1 click before the filter is applied.
 * User cannot see how does each shortcut reflect in the input before waiting for the request to complete and reopenning the filter menu.
 * User does not know what will happen, because shortcuts look different than other filters.

I think the current solution is somewhere in the middle and brings best of both worlds.

>     * I can't modify the state of the inputs even when i'm trying to input an custom range and because of that I can't modify the time.

Are you saying that you cannot type into the input after you click on "Custom range"? What browser do you use?

>     * I'm unsure if the current time ranges are intuitive, because e.g. last day you actually mean the last 24 hours instead of the whole day from 00:00 to 23:59/now(). The same thing goes for the week and the month.

See the explanation under the three points above. I think clicking on the link and updating the inputs will help user understand what the shortcuts mean. The other option is to use something like Last 60 min, 24 hours, 7 days, 30 days.

>     * Absolute urls aren't supported -> the state of time range filters isn't populated from passed url.

Thanks, totally overlloked this! It is now fixed.

@jotes r?

pontoon/base/templates/translate.html
```diff
@@ -149,46 +150,17 @@
...
-              <span class="status fa"></span>
-              <span class="title">Unchanged</span>
-            </li>
-            <li class="has-suggestions" data-type="has-suggestions">
-              <span class="status fa"></span>
-              <span class="title">Has Suggestions</span>
-            </li>
+            {{ Filter.item('untranslated', 'Untranslated') }}
+            {{ Filter.item('unchanged', 'Unchanged') }}
+            {{ Filter.item('has-suggestions', 'Has Suggestions') }}
```
jotes: <3

jotes:
Thanks for an explanation, unfortunately I can't look at UI (currently I'm a little busy @ work) but the code looks okay.

jotes:
Small UI notes:
* `.error` class stays on the input even when user clicks on the one of predefined time ranges.
* the time range filter is being reset to default value after user returns from the list on which time range filter was applied. I think it should keep the same state.

pontoon/base/static/js/translate.js
```diff
@@ -15,25 +15,74 @@ var Pontoon = (function (my) {
...
 
       // Validate filter
       if (node.length === 0) {
-        type = 'all';
-        node = $('#filter .menu li[data-type="' + type + '"]');
+        // Special case: filter by time range
+        if (/^[0-9]{12}-[0-9]{12}$/.test(type)) {
+          // Populate from/to inputs if filter specified in URL
+          if (!$('#from').val() && !$('#to').val()) {
+            var from = new Date(
```
jotes: I know it may be too late but my friend suggested that we may use the momentjs library for the date parsing.

jotes:
> Are you saying that you cannot type into the input after you click on "Custom range"? What browser do you use?

Haha, you're right -> I couldn't type any letters, that's why assumed that input is disabled. My bad.

> The other option is to use something like Last 60 min, 24 hours, 7 days, 30 days.

Yeah, that was my point. But I'm not convinced if it's the best way. I trust your experience in that matter.

> I think the current solution is somewhere in the middle and brings best of both worlds.

Thank you for an such thoroughful explanation. 

> Thanks, totally overlloked this! It is now fixed.

Confirmed :D

jotes:
r+ :dancers: :fireworks:

PR #429, opened 2016-07-09
Added missing libmecached-dev package.

jotes:
Hi @mathjazz could you review? @davidascher had problems with the pylibmc because this package was missing.

### DIFF
--- a/Aptfile
+++ b/Aptfile
@@ -1,2 +1,3 @@
 libsqlite3-dev
 subversion
+libmemcached-dev


davidascher:
As I understand the setup on Heroku, this won't help.  Heroku doesn't let
one configure the system level packages that get installed, only the
language-specific ones (requirements.txt for python, package.json for node,
etc.).  In the case of memcached, there is a hack that Heroku seems to use
which is to scan requirements.txt for pylibmc and if it finds it, then it
magically installs libmemcached. Or at least that's what
http://stackoverflow.com/questions/11507639/memcached-on-heroku-w-django-cant-install-pylibmc-memcacheify#comment37537854_11587142
seems
to say.  I'm going to poke at the source of the buildpack to find out
what's actually going on, but in any case it doesn't seem to work for me.
I'd be curious to know if someone else can do a deploy to heroku and can
get it to work with the current tree.

davidascher:
Figured it out.  Latest pip broke pip-grep, which broke the standard Heroku
buildpack: https://github.com/kennethreitz/pip-pop/pull/12

Workaround is to fork the Heroku python buildpack and fix it there.

@edmorley seems to have gotten there before me (for a different package).

jotes:
@davidascher In my understanding, this is a separate issue. Aptfile is processed by our buildpack which is available here: https://github.com/mozilla/heroku-buildpack-apt
I didn't know about these issues with the pip @ heroku. Do you know if there's a buildpack that fixes this problem? We could temporarily switch buildpacks in our docs.

mathjazz:
@davidascher We're currently using the previous version of the python build pack:
https://github.com/heroku/heroku-buildpack-python.git#v79

Does that help?

Not sure if it's related to this PR, though.

davidascher:
Doubt it, as the problem is PIP was upgraded breaking an (unpublished) API
that Heroku's `pip-grep` tool uses.

mathjazz:
> I'd be curious to know if someone else can do a deploy to heroku and can get it to work with the current tree.

@davidascher, we use the following buildpacks to deploy the current master:

```
https://github.com/mozilla/heroku-buildpack-apt.git#v0.1
https://github.com/Osmose/heroku-buildpack-ssh.git#v0.1
heroku/nodejs
https://github.com/heroku/heroku-buildpack-python.git#v79
```

@jotes Since the deployment to prod and stage servers work fine (pylibmc gets installed properly), I don’t think we need to change anything in the Aptfile.

mathjazz:
Closing.

PR #430, opened 2016-07-13
Use default Git branch instead of master

mathjazz:
We shouldn't be requiring Git repository branch name to be called master. We should use the default branch. We just got a request to enable a project that uses "develop" branch as default and wants it to be used by Pontoon.

Later on we could add a new Repository model property to specify a branch name in addition to current type and URL.

@jotes r?


jotes:
r+

PR #431, opened 2016-07-16
Do not break sync if file added and not all locales in repo

mathjazz:
This is an improvement over #408, which introduced adding locales to
projects, even if they their respective folders weren't present in the
repository.

While #408 made this possible, it broke sync if locale that wasn't
enabled in the repository was added to the project and a new source
file was added to the project.

@jotes r?


jotes:
r+

PR #432, opened 2016-07-18
Vector version of Pontoon's logo

StoyanDimitrov:
This vector version is based on the logo.png file in the same folder. The file is derived from hand-crafted SVG source available pre request (because of its incompatibility with image editing programs).


mathjazz:
Thank you very much for your first patch, @StoyanDimitrov! r+

It's great to finally have a vector version of our logo.

PR #434, opened 2016-07-19
Disable in-place editing

mathjazz:
Translating on the page itself is insufficient for various reasons:
* Original string is not always fully seen (e.g. if containing markup)
* Additional string details like comments and context are not visible
* Suggestions from history, machinery and other locales are not available
* Only the first plural form can be translated
* Features like Copy from original or Move to next string are missing

In addition to that, when translating in the page itself using the
contentEditable attribute, we were sometimes unable to fully control the
[markup that was submitted](https://github.com/mozilla/donate.mozilla.org/issues/1596).

From now on, sidebar is always open and strings can only be translated
in there, while it's still possible to select them on the page. The
sidebar switch icon (hamburger) is thus obsolete and has been removed.

Fix bug 1183819.

@jotes r?

### DIFF
--- a/pontoon/base/static/css/pontoon.css
+++ b/pontoon/base/static/css/pontoon.css
@@ -1,12 +1,10 @@
 .pontoon-hovered {
   outline: 1px dashed !important;
-}
-
-/* Prevent converting white-spaces at the beginning or
- * at the end of the node to &nbsp; and <br>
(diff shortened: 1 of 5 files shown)


jotes:
Whoah, this is a pretty huge change! r+

PR #435, opened 2016-07-24
Add more details to project request email

mathjazz:
@jotes r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -11,6 +11,7 @@ from django.conf import settings
 from django.contrib.auth.models import User, Group, UserManager
 from django.contrib.postgres.fields import ArrayField
 from django.core.exceptions import ValidationError
+from django.core.urlresolvers import reverse
 from django.db import models
 from django.db.models import Sum, Prefetch, F, Q, Case, When
 from django.templatetags.static import static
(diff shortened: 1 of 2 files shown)


pontoon/base/views.py
```diff
@@ -1257,16 +1258,29 @@ def request_projects(request, locale):
...
+        reverse('pontoon.contributor', kwargs={'email': user.email})
+    )
 
     if settings.PROJECT_MANAGERS[0] != '':
         EmailMessage(
-            'Project request for {locale} ({code})'.format(locale=locale.name, code=locale.code),
-            'Please add the following projects to {locale} ({code}):\n{projects}'.format(
-                locale=locale.name, code=locale.code, projects=projects
+            subject='Project request for {locale} ({code})'.format(locale=locale.name, code=locale.code),
+            body=u'''
```
jotes: [optional] We could move the contents of this email into separate template.

pontoon/base/views.py
```diff
@@ -1257,16 +1258,29 @@ def request_projects(request, locale):
         return HttpResponseBadRequest('Bad Request: Non-existent projects specified')
 
     projects = ''.join('- {} ({})\n'.format(p.name, p.slug) for p in project_list)
+    user = request.user
+    user_url = request.build_absolute_uri(
```
jotes: I would move this into The User class.

pontoon/base/models.py
```diff
@@ -118,12 +118,25 @@ def serialize(self):
...
 
-    return [locale.code for locale in locales]
+
+def user_gravatar_url(self, size):
+    email = hashlib.md5(self.email.lower()).hexdigest()
+    data = {'s': str(size)}
+
+    if not settings.DEBUG:
+        append = '_big' if size > 44 else ''
+        data['d'] = settings.SITE_URL + static('img/anon' + append + '.jpg')
```
jotes: [nit] we could use urljoin.

jotes:
r+

pontoon/base/views.py
mathjazz: Good idea!

pontoon/base/models.py
mathjazz: Keeping it as it is, since this was copied from the gravatar example.

pontoon/base/views.py
mathjazz: Keeping it in the view for now for easier access. But if we keep adding stuff, I agree it should be moved into a template.

PR #437, opened 2016-07-26
Added sane way for handling events for the webdevelopers.

jotes:
Hi @mathjazz, could you look at this PoC?
Basically I wanted to create an interface for the webdevelopers to handle events that are sent by pontoon. I think this approach could be better than current state that forces user to create separate receiver for POST messages sent from a parent frame. This can be unsafe on the long run, because user can meet the unexpected issue with timing or unexpected changes in internal structure of pontoon.js that cause unexpected bugs.

This my initial idea how we could tackle this problem.  With a set of predefined events we create an safe interface which webdevelopers could use to integrate their sites with pontoon.

### DIFF
--- a/docs/dev/integration.rst
+++ b/docs/dev/integration.rst
@@ -1,0 +1,24 @@
+Integration
+===========
+
+The following is a list of mechanism that can help you to integrate your website/app
+with pontoon.
+
+Event-driven integration
+------------------------
+Pontoon exposes various events that can be handled jQuery callbacks.
+Following code will call an alert window when your page is called inside Pontoon Editor:
+
+.. code-block:: javascript
+
+   $(function() {
+     $(document).on('pontoonInit', function() {
+       alert('Hello world!');
+     });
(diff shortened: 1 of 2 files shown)


jotes:
Ok, this can be done a little simpler, but the whole idea stays the same. Stay tuned -> I'll push the fix later.

jotes:
Okay, this was actually a lot simpler than I initially thought. jQuery exposes beautiful interface for events and their callbacks. I think we can use this, because we utilize jQuery a lot in our codebase.

mathjazz:
Thanks for looking into pontoon.js! There's a lot to be desired in the in-context translation field. I was actually thinking about dropping jQuery in pontoon.js, because we need to load the entire library in case the hosting website doesn't use it.

As per discussion on IRC, we're not sure there's currently a use case for these kind of events.

jotes:
'kay, I think we can close this fella for the time being.

PR #438, opened 2016-07-27
No need to update authors on unapprove

mathjazz:
If we unapprove a translation, that does not change the set of translation authors, so no need to update it.

@jotes r?


jotes:
r+

PR #439, opened 2016-07-28
Chart range

mathjazz:
@jotes r?

This is a request for the UX testing only.

Code is not ready yet, I still need to remove the old date picker.


pontoon/base/models.py
```diff
@@ -1371,6 +1371,32 @@ def find_and_replace(self, find, replace, user):
...
+            .extra({'minute': "date_trunc('minute', date)"})
+            .order_by('minute')
+            .values('minute')
+            .annotate(count=Count('id'))
+        )
+
+        data = []
+        for period in translations:
+            data.append([
+                utils.convert_to_unix_time(period['minute']),
```
jotes: I think we could replace this loop with time conversion inside the sql query and  .values_list() call at the end.

mathjazz: We need a list of lists, but the `.values_list()` only gives us a list of tuples I think.

jotes:
My initial impressions: whoa, this looks awesome. I can't wait until buttons will be fully working. This looks awesome. I'll try to put some more data and stress this widget a little.

mathjazz:
@jotes r?

jotes:
r+ :-)

PR #440, opened 2016-07-29
Optimize detecting relevant changed files

mathjazz:
Detecting which on the list of changed files since the last sync is
an actual resource file and which locale does it belong to could take
minutes, now it should take seconds.

@jotes r?

This is a quite significant improvement. Due to this bottleneck, sync sometimes runs significantly faster with the `--force` flag (which syncs all resources)!

### DIFF
--- a/pontoon/sync/vcs/models.py
+++ b/pontoon/sync/vcs/models.py
@@ -151,35 +151,21 @@ class VCSProject(object):
                     for path in find_changed_files(repo, locale):
                         files.setdefault(path, []).append(locale)
             else:
-                for changed_file in find_changed_files(repo):
-                    path_info = self.get_path_info(changed_file, repo)
+                locale_directories = self.locale_directories(repo)
+                locale_directories_paths = locale_directories.keys()
 
-                    if path_info:
-                        path, locale_path, locale = path_info
-                        path = path[len(locale_path):].lstrip(os.sep)
-                        files.setdefault(path, []).append(locale)
+                for changed_file in find_changed_files(repo):
+                    if is_hidden(changed_file):
+                        continue
+
+                    for ldp in locale_directories_paths:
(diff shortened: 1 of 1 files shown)


pontoon/sync/vcs/models.py
```diff
@@ -151,36 +151,22 @@ def find_changed_files(repo, locale=None):
...
                 for changed_file in find_changed_files(repo):
-                    path_info = self.get_path_info(changed_file, repo)
+                    if is_hidden(changed_file):
+                        continue
 
-                    if path_info:
-                        path, locale_path, locale = path_info
-                        path = path[len(locale_path):].lstrip(os.sep)
-                        files.setdefault(path, []).append(locale)
+                    for ldp in locale_directories_paths:
```
jotes: Erm, right... Good idea.  `get_path_info` is a legacy from the one of the more "functional" versions of this code.9

pontoon/sync/vcs/models.py
```diff
@@ -151,36 +151,22 @@ def find_changed_files(repo, locale=None):
                     for path in find_changed_files(repo, locale):
                         files.setdefault(path, []).append(locale)
             else:
+                locale_directories = self.locale_directories(repo)
+                locale_directories_paths = locale_directories.keys()
+
                 for changed_file in find_changed_files(repo):
-                    path_info = self.get_path_info(changed_file, repo)
+                    if is_hidden(changed_file):
```
jotes: Hm, do you have any exaples of repos that have a large number of hidden files in their structure ?

jotes:
A small step for code, a giant leap for pontoon. r+ ;D

pontoon/sync/vcs/models.py
mathjazz: I don't think so. Are you implying we should remove this check?

PR #442, opened 2016-08-01
Live update in context

mathjazz:
@jotes r? I know this is tedious review, so if you don't feel comfortable doing it, just let me know.

It's a few bugfixes, code cleanups and most of all - live updating in-context translations as you change them in the editor. Sadly there's now good way of detecting changes to translation `textarea`, so we have to make a number of function calls.

### DIFF
--- a/pontoon/base/static/js/pontoon.js
+++ b/pontoon/base/static/js/pontoon.js
@@ -494,44 +494,29 @@
             break;
 
           case "SAVE":
-            var entity = message.value.id ? Pontoon.entities[message.value.id] : $('.pontoon-editable-toolbar')[0].target.entity,
-                translationValue = message.value.translation,
-                translation = translationValue !== undefined ? translationValue : message.value;
+            var entity = Pontoon.entities[message.value.id],
+                translation = message.value.translation;
 
             entity.translation[0].string = translation;
 
             $(entity.node).each(function() {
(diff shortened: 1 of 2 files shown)


pontoon/base/static/js/pontoon.js
```diff
@@ -507,31 +506,19 @@
...
-
-            entity.translation[0].pk = null;
-            entity.translation[0].string = null;
-            entity.translation[0].approved = false;
-            entity.translation[0].fuzzy = false;
-
-            sendData();
-            postMessage("DELETE", entity.id);
+              entity.translation[0].pk = null;
+              entity.translation[0].string = null;
```
jotes: a nice simplification!

pontoon/base/static/js/translate.js
```diff
@@ -1369,7 +1369,13 @@ var Pontoon = (function (my) {
            $('#translation').val(data.translation.string).focus();
            self.updateCachedTranslation();
            self.updateCurrentTranslationLength();
-           self.postMessage("SAVE", data.translation);
+
+           if (entity.body && pf === 0) {
+             self.postMessage("SAVE", {
```
jotes: 'k, right.

jotes:
r+, i see a lot of good fixes.

PR #447, opened 2016-08-08
Use appropriate locale separator when creating folders

mathjazz:
@jotes r?


jotes:
This looks a little bit hacky, but should work in general. r+

PR #448, opened 2016-08-08
Fix bug 1292890: Sync stats if only resources changed in monolingual files

mathjazz:
If only source files change in repo (e.g. when new strings are merged), we skip syncing translations. Which is the right thing to do, but as part of this we also skip updating stats. Which we shouldn't.

@jotes r?


jotes:
r+

PR #449, opened 2016-08-12
Add tooltip to the TMs' matches count in Machinery tab.

StoyanDimitrov:
Even though my English skills are clearly outstanding I am open to other suggestions.

### DIFF
--- a/pontoon/base/static/js/main.js
+++ b/pontoon/base/static/js/main.js
@@ -130,13 +130,14 @@ var Pontoon = (function (my) {
 
       function append(data) {
         var title = loader !== 'search' ? ' title="Copy Into Translation (Tab)"' : ' title="Copy to clipboard"',
-            sources = sourcesMap[data.original + data.translation];
+            sources = sourcesMap[data.original + data.translation],
+            occurrencesTitle = 'Number of translation occurrences';
 
         if (sources) {
           sources.append(
             '<li><a class="translation-source" href="' + data.url + '" target="_blank" title="' + data.title + '">' +
               '<span>' + data.source + '</span>' +
-              (data.count ? '<sup>' + data.count  + '</sup>' : '') +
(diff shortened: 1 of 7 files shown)


mathjazz:
Thanks for the patch!

What if we add titles to all three tabs? Instead of counts, I think we should explain the meaning of the tabs, because they are very cryptic (due to limited space available), so the `title` tag should be added to the `li` element.

Something like this:
* History: Translation history
* Machinery: Translation memory, machine translation & terminology
* Locales: Translations to other languages

Let me know what do you think.

StoyanDimitrov:
The PR's title is misleading, sorry about that. It is adding a title attribute to the count of perfect matches in a single TM provider (i.e. transvision) that is forming a single item in the list of "machinery" suggestions and not the count of all items in this tab.

Even though I've used variety of other similar tools it was not immediately obvious what this number represents and how should I use it. That's the reason behind this PR.

OTOH, what you suggest is even more helpfull and visible to the users and I'll add it with this PR.

mathjazz:
Ouch, sorry. I spoke too fast. You're right, that's not clear at all. I suggest you store the title tag in a variable and then reuse it, so that we keep the text in one place. Maybe the text should be "Number of translation occurrences"?

If you still want to submit a patch I mentioned in my original comment, you can do so in a separate PR.

StoyanDimitrov:
This must be it.

mathjazz:
I suggest we only keep the text in the variable (occurrencesTitle = 'Number of translation occurrences'). Please update that and rebase into a single commit.

Nice work! r+wc

mathjazz:
Thanks!

Could you please [squash](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) the commits?

mathjazz:
Right, I can squash those myself. ;)

Thanks for the patch!

PR #451, opened 2016-08-13
Update translated resources when skipping locales

mathjazz:
This is even simpler than it looks: if we skip locale sync, but the resources have been changed, we need to update translated resources anyways to update the total_strings count.

Also remove unused argument from the `update_translated_resources()` function and update tests.

@jotes r?


jotes:
Looks okay -> r+

PR #453, opened 2016-08-15
Add helper tabs tooltips.

StoyanDimitrov:
As described in https://github.com/mozilla/pontoon/pull/449#issuecomment-239624613


mathjazz:
Thanks for the patch!

PR #454, opened 2016-08-16
Fix bug 1278826: filter partially translated pluralized strings as missing

mathjazz:
This fixes the missing filter to also include partially translated pluralized strings.

@jotes r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -1087,7 +1087,7 @@ class EntityQuerySet(models.QuerySet):
 
     def missing(self, locale):
         return self.with_status_counts(locale).filter(
-            Q(approved_count=0) & Q(fuzzy_count=0) & Q(suggested_count=0)
+            Q(approved_count__lt=F('expected_count')) & Q(fuzzy_count__lt=F('expected_count')) & Q(suggested_count__lt=F('expected_count'))
         )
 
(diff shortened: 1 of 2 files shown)


pontoon/base/models.py
```diff
@@ -1087,7 +1087,7 @@ def with_status_counts(self, locale):
 
     def missing(self, locale):
         return self.with_status_counts(locale).filter(
-            Q(approved_count=0) & Q(fuzzy_count=0) & Q(suggested_count=0)
+            Q(approved_count__lt=F('expected_count')) & Q(fuzzy_count__lt=F('expected_count')) & Q(suggested_count__lt=F('expected_count'))
```
jotes: A relative nit: I think we don't test case for this behaviour....

jotes:
r+wc

pontoon/base/models.py
mathjazz: I'm afraid we don't test filters at all. :/

jotes: Are you sure?
https://github.com/mozilla/pontoon/blob/master/pontoon/base/tests/test_models.py#L1079

mathjazz: Thanks for introducing me to Pontoon! :-)

PR #456, opened 2016-08-17
Prevent simultaneous save translation requests

mathjazz:
While we disable the save button while request is in progress, users
can still submit multiple translation update requests with their
keyboards. That can lead to stats inconsistency due to DB latency
in production. With this patch, it is now actively prevented.

@jotes r?

### DIFF
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -2151,7 +2151,11 @@ var Pontoon = (function (my) {
         submittedPluralForm = 0;
       }
 
-      return $.ajax({
+      if (self.XHRupdateOnServer) {
+        self.XHRupdateOnServer.abort();
+      }
+
+      self.XHRupdateOnServer = $.ajax({
         url: '/update/',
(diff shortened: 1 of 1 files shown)


pontoon/base/static/js/translate.js
```diff
@@ -2151,7 +2151,11 @@ var Pontoon = (function (my) {
         submittedPluralForm = 0;
       }
 
-      return $.ajax({
+      if (self.XHRupdateOnServer) {
+        self.XHRupdateOnServer.abort();
```
jotes: I think that abort() won't guarantee that server didn't already processed your call. Maybe it should be better if you we would (at the very top of updateOnServer()) check if there's any ongoing request to the backend and then return.

mathjazz: True. But in this case (if the server already processed the call) we no longer have the problem we are solving - which is being able to submit two approved translations for the same string for the same locale.

jotes:
r+ :godmode:

PR #457, opened 2016-08-23
Use username instead of email as user identifier in URL

mathjazz:
We use email addresses as user identifiers in contributor URLs, which is a bad idea, at least due to poor privacy respect and spam protection, making e.g the [top contributors](https://pontoon.mozilla.org/contributors/) page very spam-bot friendly.

This patch replaces email addresses with usernames, while keeping the old email-based URLs alive (to avoid breaking links and to help with manual URL entering).

@jotes r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -134,7 +134,9 @@ def user_role(self, locale):
 
 @property
 def user_profile_url(self):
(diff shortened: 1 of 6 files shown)


jotes:
r+ I just overlooked this PR. I think we can land it now, before the migration to fxa.

PR #459, opened 2016-08-30
Fix bug 1299084 - Use username instead of email

bychek-ru:
Now filters by authours use usernames.
Deleted email from entries' history.
Use names instead of emails on frontend.
Deleted emails from contributors' lists,


bychek-ru:
@mathjazz Do you know where I can add username in this block to use it on frontend?

mathjazz:
@bychek-ru Please see https://bugzilla.mozilla.org/show_bug.cgi?id=1299084#c2.

PR #460, opened 2016-08-30
Fix bug 1299084 - Use email only if user didn't specify their name

bychek-ru:
Changes:
Use user id in JS.
Use roles instead of emails contributors' lists.

Additionaly: fixed links on MDN

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -115,9 +115,11 @@ class UserQuerySet(models.QuerySet):
         for user in self:
             users.append({
                 'email': user.email,
-                'display_name': user.display_name,
+                'display_name': user.name_or_email,
+                'id': user.id,
                 'gravatar_url': user.gravatar_url(44),
                 'translation_count': user.translation_count,
+                'rights': user.rights
             })
 
         return users
@@ -180,6 +182,27 @@ def user_translated_locales(self):
 
     return [locale.code for locale in locales]
 
+
+@property
+def user_managed_locales(self):
+    locales = get_objects_for_user(
+        self, 'base.can_manage_locale', accept_global_perms=False)
+
+    return [locale.code for locale in locales]
+
+
+@property
+def user_rights(self):
+    if self.is_superuser:
+        return 'Admin'
+    if self.managed_locales:
+        return 'Manager for ' + ', '.join(self.managed_locales)
+    if self.translated_locales:
+        return 'Translator for ' + ', '.join(self.translated_locales)
+    else:
+        return 'Contributor'
+
+
 User.add_to_class('role', user_role)
 User.add_to_class('profile_url', user_profile_url)
 User.add_to_class('gravatar_url', user_gravatar_url)
@@ -188,6 +211,8 @@ User.add_to_class('display_name', user_display_name)
 User.add_to_class('display_name_and_email', user_display_name_and_email)
 User.add_to_class('display_name_or_blank', user_display_name_or_blank)
 User.add_to_class('translated_locales', user_translated_locales)
+User.add_to_class('managed_locales', user_managed_locales)
+User.add_to_class('rights', user_rights)
 User.add_to_class('translators', UserTranslationsManager())
 User.add_to_class('objects', UserCustomManager.from_queryset(UserQuerySet)())
 
--- a/pontoon/base/static/css/translate.css
+++ b/pontoon/base/static/css/translate.css
@@ -899,7 +899,7 @@ body > header aside p {
   padding-bottom: 4px;
 }
 
-#filter .menu li.author .email {
+#filter .menu li.author .rights {
   color: #AAAAAA;
 }
 
--- a/pontoon/base/static/css/users.css
+++ b/pontoon/base/static/css/users.css
@@ -88,7 +88,7 @@ sup a:visited {
   height: 1.5em;
 }
 
-.contributor p.email {
+.contributor p.rights {
   left: 67px;
   position: absolute;
   top: 47px;
--- a/pontoon/base/static/js/main.js
+++ b/pontoon/base/static/js/main.js
@@ -606,7 +606,7 @@ $(function() {
     Pontoon.startLoader();
 
     $.ajax({
-      url: '/api/v1/user/' + $('#server').data('email') + '/',
+      url: '/api/v1/user/' + $('#server').data('username') + '/',
       type: 'POST',
       data: {
         csrfmiddlewaretoken: $('#server').data('csrf'),
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -69,8 +69,8 @@ var Pontoon = (function (my) {
       var title = node.find('.title').text();
 
       // Special case: filter by author
-      if (node.find('.email').length) {
-        title = node.find('.email').text();
+      if (node.find('.name').length) {
+        title = node.find('.name').text() + '\'s translations';
         selectorType = 'all';
       }
 
@@ -320,18 +320,18 @@ var Pontoon = (function (my) {
                 '" title="Copy Into Translation (Tab)">' +
                   '<header class="clearfix' +
                     ((self.user.isTranslator) ? ' translator' :
-                      ((self.user.email === this.email && !this.approved) ?
+                      ((self.user.id === this.uid && !this.approved) ?
                         ' own' : '')) +
                     '">' +
                     '<div class="info">' +
-                      ((!this.email) ? '<span title="' + self.getApproveButtonTitle(this) + '">' + this.user + '</span>' :
+                      ((!this.uid) ? '<span>' + this.user + '</span>' :
                         '<a href="/contributors/' + this.username + '" title="' + self.getApproveButtonTitle(this) + '">' + this.user + '</a>') +
                       '<time class="stress" datetime="' + this.date_iso + '">' + this.date + ' UTC</time>' +
                     '</div>' +
                     '<menu class="toolbar">' +
                       '<button class="' + (this.approved ? 'unapprove' : 'approve') + ' fa" title="' +
                        (this.approved ? 'Unapprove' : 'Approve')  + '"></button>' +
-                      ((self.user.email && (self.user.email === this.email) || self.user.isTranslator) ? '<button class="delete fa" title="Delete"></button>' : '') +
+                      ((self.user.id && (self.user.id === this.uid) || self.user.isTranslator) ? '<button class="delete fa" title="Delete"></button>' : '') +
                     '</menu>' +
                   '</header>' +
                   '<p class="translation" dir="auto" lang="' + self.locale.code + '">' +
@@ -602,8 +602,7 @@ var Pontoon = (function (my) {
     checkUnsavedChanges: function (callback) {
       var entity = this.getEditorEntity();
 
-      // Ignore for anonymous users, for which we don't save translations
-      if (!this.user.email || !entity) {
+      if (!entity) {
         this.restoreInPlaceTranslation();
         return callback();
       }
@@ -1318,7 +1317,7 @@ var Pontoon = (function (my) {
         e.preventDefault();
 
         // Ignore for anonymous users
-        if (!self.user.email) {
+        if (!self.user.id) {
           return;
         }
 
@@ -1538,7 +1537,7 @@ var Pontoon = (function (my) {
         }
 
         // Ignore for anonymous users
-        if (!self.user.email) {
+        if (!self.user.id) {
           return;
         }
 
@@ -2497,7 +2496,7 @@ var Pontoon = (function (my) {
             '<img class="rounded" src="' + this.gravatar_url + '">' +
             '<figcaption>' +
               '<p class="name">' + this.display_name + '</p>' +
-              '<p class="email">' + this.email + '</p>' +
+              '<p class="rights">' + this.rights + '</p>' +
             '</figcaption>' +
             '<span class="count">' + this.translation_count + '</span>' +
           '</figure>' +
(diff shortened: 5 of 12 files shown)


pontoon/base/urls.py
```diff
@@ -49,7 +49,7 @@
         name='pontoon.profile'),
 
     # API: Toogle user profile attribute
-    url(r'^api/v1/user/(?P<email>[\w.%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,4})/$',
+    url(r'^api/v1/user/(?P<username>[\w-]+)/$',
```
jotes: I think this won't match usernames of already existing usernames, because they are base64 encoded strings.

pontoon/base/static/js/translate.js
```diff
@@ -320,18 +320,18 @@ var Pontoon = (function (my) {
...
                   '<header class="clearfix' +
                     ((self.user.isTranslator) ? ' translator' :
-                      ((self.user.email === this.email && !this.approved) ?
+                      ((self.user.username === this.username && !this.approved) ?
                         ' own' : '')) +
                     '">' +
                     '<div class="info">' +
-                      ((!this.email) ? '<span title="' + self.getApproveButtonTitle(this) + '">' + this.user + '</span>' :
+                      ((!this.username) ? '<span>' + this.user + '</span>' :
                         '<a href="/contributors/' + this.username + '" title="' + self.getApproveButtonTitle(this) + '">' + this.user + '</a>') +
```
jotes: Pontoon had some issues with XSS attacks, I think we should use something that will escape username as we do in the jinja2 templates.

jotes:
Hi @bychek-ru,
This looks awesome! I have a question: where the user can set an username? I think we should add some strict input sanitization just to avoid any security issues.

mathjazz:
Thanks for the patch, @bychek-ru! Nice work!

Before I dig deeper into it, wouldn't it be easier to just use the user ID instead of the username in JavaScript, API calls, etc.?

bychek-ru:
@jotes As I understand Pontoon gerenates username only one time during registration and user can't change it.

@mathjazz It's easier but for example, every history entry has its own id in JSON response. Should we use "id" (and call "uid" in frontend) instead of "username"?

mathjazz:
> @mathjazz It's easier but for example, every history entry has its own id in JSON response. Should we use "id" (and call "uid" in frontend) instead of "username"?

Sure, you can change the key to `uid` or `user_id`.

mathjazz:
No, profile URLs are a special case:

1. We don't want to use IDs,  because they are incremental and somebody could scrape user data by simply looping through contributor URLs.

2. We still allow for old email-based URLs, which are easier to remember if you type them manually: https://pontoon.mozilla.org/contributors/a@bychek.ru/.

bychek-ru:
@mathjazz Let's dig deeper into it

pontoon/base/static/css/translate.css
```diff
@@ -891,6 +891,7 @@ body > header aside p {
 #filter .menu li.author figcaption {
   display: inline-block;
   padding: 2px 0 2px 8px;
+  line-height:40px;
```
mathjazz: Nit: Please add space between colon and the number.

pontoon/base/static/css/users.css
```diff
@@ -85,13 +85,7 @@ sup a:visited {
...
   display: inline-block;
   font-size: 1.5em;
-  height: 1.5em;
-}
-
-.contributor p.email {
-  left: 67px;
-  position: absolute;
-  top: 47px;
+  line-height:44px;
```
mathjazz: Nit: Please add space between colon and the number.

pontoon/base/static/js/main.js
```diff
@@ -606,7 +606,7 @@ $(function() {
     Pontoon.startLoader();
 
     $.ajax({
-      url: '/api/v1/user/' + $('#server').data('email') + '/',
+      url: '/api/v1/user/' + $('#server').data('uid') + '/',
```
mathjazz: I know I suggested otherwise, but I wonder if we should use username here, so that we always use username in URLs. Should prevent crawlers to increment uid and make calls. @jotes?

Anyhow, please make sure that you also update the server data- attribute in the `user_settings.html` template.

pontoon/base/models.py
```diff
@@ -1116,8 +1116,8 @@ def has_suggestions(self, locale):
     def unchanged(self, locale):
         return self.with_status_counts(locale).filter(unchanged_count=F('expected_count'))
 
-    def authored_by(self, locale, email):
-        return self.filter(translation__locale=locale, translation__user__email=email)
+    def authored_by(self, locale, id):
+        return self.filter(translation__locale=locale, translation__user__id=id)
```
mathjazz: I actually think we should keep the email address in the URL until we fix 1243115, which will allow for filters like `?status=translated&user=usernameORemail`. Using ID (or username) makes it impossible to filter by manually typing the URL.

pontoon/base/static/js/translate.js
```diff
@@ -320,18 +320,18 @@ var Pontoon = (function (my) {
                 '" title="Copy Into Translation (Tab)">' +
                   '<header class="clearfix' +
                     ((self.user.isTranslator) ? ' translator' :
-                      ((self.user.email === this.email && !this.approved) ?
+                      ((self.user.uid === this.uid && !this.approved) ?
                         ' own' : '')) +
                     '">' +
                     '<div class="info">' +
-                      ((!this.email) ? '<span title="' + self.getApproveButtonTitle(this) + '">' + this.user + '</span>' :
+                      ((!this.uid) ? '<span>' + this.user + '</span>' :
```
mathjazz: I wonder why remove the `title` attribute here?

pontoon/base/static/js/translate.js
```diff
@@ -600,10 +600,11 @@ var Pontoon = (function (my) {
...
-      var entity = this.getEditorEntity();
+      var entity = this.getEditorEntity(),
+	      self = this;
 
       // Ignore for anonymous users, for which we don't save translations
-      if (!this.user.email || !entity) {
+      if (!self.user.uid || !entity) {
         this.restoreInPlaceTranslation();
         return callback();
       }
```
mathjazz: We can remove this entire block, because we don't allow submitting anonymous translation anymore.

pontoon/base/static/js/translate.js
```diff
@@ -3543,8 +3543,7 @@ window.onpopstate = function(e) {
 };
 
 Pontoon.user = {
-  email: $('#server').data('email') || '',
-  name: $('#server').data('name') || '',
+  uid: $('#server').data('uid') || '',
```
mathjazz: Nit: We should probably call it `Pontoon.user.id`, so `id` instead of `uid`.

pontoon/base/templates/500.html
```diff
@@ -376,7 +376,7 @@
             <li><a href="/contributors">Contributors</a></li>
             <li><a href="/terminology">Terminology</a></li>
             <li><a href="/terms">Terms</a></li>
-            <li><a href="https://developer.mozilla.org/en-US/docs/Localizing_with_Pontoon" target="_blank">Help</a></li>
+            <li><a href="https://developer.mozilla.org/docs/Localizing_with_Pontoon" target="_blank">Help</a></li>
```
mathjazz: Nice catch!

pontoon/base/static/css/users.css
mathjazz: "Names" are now longer, because we often present email addresses instead. We need smaller font size. Maybe 16 px like in the sidebar in the translate view. Also, the actual height should be 48px (also in the translate view), like the image. Currently text is not vertially center-aligned to the image.

It would be nice to unify the HTML code we use in all these user presentation, but we can also do this in a separate PR.

mathjazz:
Good job, @bychek-ru! I added a few notes, mostly nits.

Two things that I'd like to point our are the user identifiers in the filter and API URLs.

pontoon/base/static/js/translate.js
bychek-ru: If we delete this entire block we'll see:

pontoon/base/static/js/translate.js
bychek-ru: It hasn't any useful info because this condition will work only if translation is "imported". In this state we don't know who appoved or did anything else with this string.

bychek-ru:
@mathjazz r?

pontoon/base/models.py
```diff
@@ -1116,8 +1116,8 @@ def has_suggestions(self, locale):
     def unchanged(self, locale):
         return self.with_status_counts(locale).filter(unchanged_count=F('expected_count'))
 
-    def authored_by(self, locale, email):
-        return self.filter(translation__locale=locale, translation__user__email=email)
+    def authored_by(self, locale, name_or_email):
+        return self.filter(translation__locale=locale, translation__user__first_name=name_or_email) or self.filter(translation__locale=locale, translation__user__email=name_or_email)
```
mathjazz: We should only take email as an argument, because firstname is not unique.

pontoon/base/static/js/translate.js
mathjazz: Correct, not the entire block - we should only remove the `!self.user.uid || ` part.

mathjazz:
Nice job, @bychek-ru!

The only two things I'd like to see changed is:
* using only email as an argument for filtering by user (since firstname is not unique)
* removing the `self.user.uid` check in `checkUnsavedChanges`.

bychek-ru:
@mathjazz Updated.

mathjazz:
That looks great, thank you! r+

I do however have one more feature request and I'm sorry it's coming so late. The /contributors list look pretty plain now with only name or email printed in relatively small font (used because the bigger wouldn't fit). And the same goes for the contributor list in the filter menu.

So I suggest we do this:
- use the same CSS (big font) and HTML as before for presenting display_name above and email below.
- underneath, where we used to display email, display user role, which could be any of the following:
 * Contributor (registered user without any permissions)
 * Translator for dsb, hsb
 * Manager for sl

We can do it by adding a new user property, similar to https://github.com/mozilla/pontoon/blob/master/pontoon/base/templatetags/helpers.py#L90. It also convers a special case for superusers so that we don't list all locales.

It would allow us to use the old and familiar layout, but also add additional information on a user. 

What do you think?

Since this is coming so late, I can also merge your pull request and provide the patch for this myself. Please let me know if you want to do it instead.

bychek-ru:
@mathjazz Yes, I've this idea too. Done but don't know how to determine if an user is a manager or not. Give me a hint or let's merge and fix it later.

jotes:
@bychek-ru you can look here: https://github.com/mozilla/pontoon/blob/master/pontoon/base/models.py#L127 I hope this will help.

bychek-ru:
@jotes Thank you, I decided choose this function and I did it
@mathjazz Please review

pontoon/base/models.py
```diff
@@ -180,6 +183,28 @@ def user_translated_locales(self):
...
     return [locale.code for locale in locales]
 
+
+@property
+def user_rights(self):
+    output = 'Contributor'
+
+    if self.translated_locales:
+        if self.is_superuser:
+            locales = 'all locales'
```
mathjazz: Actually, can we just call the user 'Admin' if it's a super user (without listing any locales)?

pontoon/base/models.py
```diff
@@ -180,6 +183,28 @@ def user_translated_locales(self):
...
+        else:
+            locales = ', '.join(self.translated_locales)
+
+        locale = get_object_or_404(Locale, code__iexact=self.translated_locales[0])
```
mathjazz: We actually need to create `managed_locales` and check `if self.managed_locales:` and then list them. Currently you only check if the user is a manager of the first locale he is a translator for and then list all translated_locales as locales he is a manager for.

mathjazz:
Great job! Only added two notes.

bychek-ru:
@mathjazz Fixed

pontoon/base/models.py
```diff
@@ -180,6 +183,35 @@ def user_translated_locales(self):
...
     return [locale.code for locale in locales]
 
+
+@property
+def user_rights(self):
+    output = 'Contributor'
+
+    if self.translated_locales:
+        if self.is_superuser:
+            return 'Admin'
```
mathjazz: You can move the `if self.is_superuser` block out - to the top of the function.

pontoon/base/models.py
```diff
@@ -180,6 +183,35 @@ def user_translated_locales(self):
...
+
+        output = ''
+        managed = []
+        translated = []
```
mathjazz: `translated` is basically `self.translated_locales`, so we don't need a separate variable.

Also, we should create `self.managed_locales` (check [this](https://github.com/mozilla/pontoon/blob/master/pontoon/base/models.py#L178) out ), so we don't need `managed`.

Once you have that, all you need to do is:
```
if self.is_superuser:
  return 'Admin'
if self.managed_locales:
  return 'Manager for ' + ', '.join(self.managed_locales)
if self.translated_locales:
  return 'Translator for ' + ', '.join(self.managed_locales)
else:
  return 'Contributor'
```

mathjazz:
Works great, added a not on how to simplify the code.

bychek-ru:
@mathjazz Simplified

pontoon/base/models.py
```diff
@@ -180,6 +182,30 @@ def user_translated_locales(self):
...
+@property
+def user_rights(self):
+    if self.is_superuser:
+        return 'Admin'
+    if self.translated_locales or self.managed_locales:
+        output = ''
+        if self.managed_locales:
+            output += 'Manager for ' + ', '.join(self.managed_locales)
+        if self.translated_locales:
+            output += (', ' if self.managed_locales else '') + 'Translator for ' + ', '.join(self.translated_locales)
```
mathjazz: We should only say if it's a Manager or Translator, not both (since all Managers are also Translators).

mathjazz:
Thanks, one last thing!

bychek-ru:
@mathjazz Done, tests ran successfully. A manager can be only only a translator in other locale, but let's not consider this exotic situation.

mathjazz:
A thing of beauty! r+

Bravo!

PR #461, opened 2016-09-01
Bug 1298610: Optimize syncing new locales

mathjazz:
This optimization affects all project repository types, fixes some bugs and hopefully makes code more readable. It's particularly noticeable on projects that use multiple repositories and when adding new locales or entire projects.

**Optimizations**
* When syncing resources, only check out source repositories (currently we check out all repositories twice: during resources and during translation sync). 3f939b2
* If no changes detected in source repository, skip syncing resources (currently we sync resources even if only translations need to be synced). 3f939b2
* Acknowledge changed files when syncing resources (currently if resources need to be synced, we sync all resources). 871cb88
* Optimize changed files detection by 90% using @cached_property. 3f83b32
* Do not detect all following files detected as changed when one detected as chanegd. Over 90% optimization for small syncs on big projects. cd6fdaf
* 80 % faster execute_update_db(), commonly the slowest function on syncing translations, mostly noticable when adding new locales or projects. 8c8fd31
* Skip syncing translations as early as possible. Avoids syncing first project locale even if not needed. 9407d86

**Code improvements**
* Return Project.source_repository and Repository.locales even if repository not checked out (needed above). b5125c1 163d99c
* Use `translation_repositories()` instead of `.exclude(source_repo=True)`, which doesn't exclude source repositories not marked with the `source_repo` flag. 3f83b32
* `VCSProject.changed_files` now also includes DB changes. 73406bd
* Changed some function names for better readability. bf01222 56af5fc
* Ignore trailing + in HG revision number (caused by local changes) to avoid hg: parse error, preventing changes to be detected. 8ecfcd8

@jotes r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -7,6 +7,7 @@ import os.path
 import re
 import urllib
 
+from collections import defaultdict
 from dirtyfields import DirtyFieldsMixin
 
 from django.conf import settings
@@ -551,6 +552,24 @@ class Project(AggregatedStats):
                 locale.aggregate_stats()
 
     @property
+    def changed_resources(self):
+        """
+        Returns a map of resource paths and their locales
+        that where changed from the last sync.
+        """
+        resources = defaultdict(set)
+        changes = (
+            ChangedEntityLocale.objects
+                .filter(entity__resource__project=self)
+                .prefetch_related('locale', 'entity__resource')
+        )
+
+        for change in changes:
+            resources[change.entity.resource.path].add(change.locale)
+
+        return resources
+
+    @cached_property
     def unsynced_locales(self):
         """
         Project Locales that haven't been synchronized yet.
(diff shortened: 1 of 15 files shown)


pontoon/base/models.py
```diff
@@ -627,18 +646,35 @@ def has_multi_locale_repositories(self):
 
         return False
 
+    @property
+    def has_single_repo(self):
+        if self.repositories.count() == 1:
```
jotes: `return self.repositories.count() == 1` would be enough.

pontoon/sync/vcs/repositories.py
```diff
@@ -357,10 +357,14 @@ def revision(self):
...
 
+    def _strip(self, rev):
+        "Ignore trailing + in revision number. It marks local changes."
+        return rev.rstrip('+')
+
     def get_changed_files(self, path, from_revision, statuses=None):
         statuses = statuses or ('A', 'M')
         code, output, error = self.execute(
-            ['hg', 'status', '-a', '-m', '-r', '--rev={}'.format(from_revision), '--rev=tip'],
+            ['hg', 'status', '-a', '-m', '-r', '--rev={}'.format(self._strip(from_revision)), '--rev=tip'],
```
jotes: [nit] Probably we could strip this string before it's saved to database.

pontoon/sync/core.py
```diff
@@ -216,28 +187,23 @@ def entity_key(entity):
...
-def pull_changes(db_project):
+def pull_changes(db_project, source_only=False):
     """
-    Update the local files with changes from the VCS. Returns
-    whether any of the updated repos have changed since the last
-    sync, based on the revision numbers.
-    Returns a tuple containing:
-    - if repository has changed.
-    - a map of revisions for each repository.
+    Update the local files with changes from the VCS. Returns if
```
jotes: `Returns True` I suppose.

jotes:
```
<joke>
10 lines of code = 10 issues. 500 lines of code = "looks fine."
</joke>
```
I reviewed whole pull request, I don't see anything obvious that should be fixed immediately. r+ :-)

pontoon/sync/vcs/repositories.py
mathjazz: I agree, but I'll leave it for now in case we need that `+` bit later for some reason.

PR #463, opened 2016-09-10
Bug 1243115 - Add ability to combine different filters

bychek-ru:


pontoon/base/models.py
```diff
@@ -1272,45 +1272,56 @@ def get_translation(self, plural_form=None):
...
-            elif filter_type in Translation.for_locale_project_paths(locale, project, paths).authors().values_list('email', flat=True):
-                entities = self.objects.authored_by(locale, filter_type)
-
-            elif re.match('^[0-9]{12}-[0-9]{12}$', filter_type):
-                try:
-                    start, end = utils.parse_time_interval(filter_type)
-                except ValueError:
+        entities = None
+        
+        if by_type or by_author or by_type:
```
jotes: I think that you meant `by_type or by_author or by_time`.

bychek-ru: Oops, yes, I was trying to understand why filter by time doesn't work. Thank you!

pontoon/base/models.py
```diff
@@ -1272,45 +1272,56 @@ def get_translation(self, plural_form=None):
...
-                entities = self.objects.unchanged(locale)
-
-            elif filter_type in Translation.for_locale_project_paths(locale, project, paths).authors().values_list('email', flat=True):
-                entities = self.objects.authored_by(locale, filter_type)
-
-            elif re.match('^[0-9]{12}-[0-9]{12}$', filter_type):
-                try:
-                    start, end = utils.parse_time_interval(filter_type)
-                except ValueError:
+        entities = None
```
jotes: I think that If you would do the `entities = Entity.objects.all()` here then you would could remove `if entities:` below.

pontoon/base/models.py
jotes: No problem. I hope that my comments help. You're doing a great job :+1:

pontoon/base/models.py
```diff
@@ -1272,46 +1272,49 @@ def get_translation(self, plural_form=None):
...
-        search=None, exclude=None):
+    def for_project_locale(self, project, locale, paths=None, by_type=None,
+        search=None, exclude=None, by_author=None, by_time=None):
         """Get project entities with locale translations."""
-        if filter_type and filter_type != 'all':
-            if filter_type == 'missing':
-                entities = self.objects.missing(locale)
+        entities = Entity.objects.all()
+        
+        if by_type or by_author or by_time:
```
jotes: After re-reading this code second time, I may be wrong but could we remove this condition? This makes this part of code more than it's necessary, also you check these values twice.

pontoon/base/static/js/translate.js
```diff
@@ -5,45 +5,54 @@ var Pontoon = (function (my) {
...
 
 
     /*
      * Set filter widget
      */
-    setFilter: function (type) {
-      type = type || 'all';
-      var node = $('#filter .menu [data-type="' + type + '"]'),
-          selectorType = type;
+    setFilter: function (filter) {
```
jotes: I think we can rename this function into `setFilters` to be more accurate. Also, could you add a docstring about required structure of passed parameter?

pontoon/base/static/js/translate.js
```diff
@@ -5,45 +5,54 @@ var Pontoon = (function (my) {
     /*
      * UI helper methods
      */
-    getFilter: function() {
-      return $('#filter').data('current-filter') || 'all';
+    getFilter: function(type) {
+      return type ? $('#filter').data('current-filter')[type] : $('#filter').data('current-filter') || {
```
jotes: Could you add a comment about parameters and return values from this method? I see that getFilter returns an object default state for all the filters - this could be misleading (correct me if I'm wrong).

pontoon/base/models.py
```diff
@@ -1272,46 +1272,49 @@ def get_translation(self, plural_form=None):
...
+                if by_author in Translation.for_locale_project_paths(locale, project, paths).authors().values_list('email', flat=True):
+                    entities = EntityQuerySet.authored_by(entities, locale, by_author)
 
-            elif re.match('^[0-9]{12}-[0-9]{12}$', filter_type):
-                try:
-                    start, end = utils.parse_time_interval(filter_type)
-                except ValueError:
-                    raise ValueError(filter_type)
-                entities = self.objects.between_time_interval(locale, start, end)
+            if by_time:
```
jotes: Could you add tests that to check if all filters are applied properly?

bychek-ru:
@jotes I fixed all your notes expect the last one. Could you tell me where we fill out test DB and which dates and emails I can use [here](https://github.com/bychek-ru/pontoon/blob/49c2123123635443655851bb0f04b47b613d57a2/pontoon/base/tests/test_views.py#L507)?

jotes:
@bychek-ru I think you should create new new objects for an User/Entities the setUp method of this testcase. You can use (User|Entity)Factory for this task.

bychek-ru:
@jotes @mathjazz I'm trying to improve test. Maybe will we fix it later? Or am I close to do it? Also, we don't have time in EntityFactory and TranslationFactory.

mathjazz:
Thanks for digging into this complex issue, @bychek-ru! Good job!

Before we start looking into tests, you might want to finish the feature implementation completely, so you don't end up writing tests for something that will get changed (or god forbid removed!) during the process. I have two usability improvement suggestions:

**User interface**
UI for selecting multiple filters should work the same way as the UI for selecting multiple strings:

- If you hover a filter icon, it turns into a checkbox.
- If you click the checkbox, the filter gets applied, but the filter menu remains open.
- If you click another checkbox, that filter gets added to the currently applied filters.
- If you click a filter item outside the checkbox, only that filter is applied and the filter menu gets closed.

The `All filters` filter should not have the ability to be "checked", because it obviously doesn't make sense to select it along with any other filter.

We could also remove the `Untranslated` filter, as it is basically a combination of `Missing`, `Fuzzy` and `Suggested`, or select these three filters when `Untranslated` is clicked.

**URL parameters**
The URL parameters should be simply called `status`, `smart`, `time` and `author`. Or maybe `extra` instead of `smart`? In which case we should also change the title to `Extra filters`.

Also, it seems impossible to select multiple filters from the `Translation Status` section at the same time.

How does that sound?

bychek-ru:
@mathjazz It sounds nice. I fixed URL params and added checkboxes.
I'm gonna let users to select multiple filters by status and authors and also fix behaviour of selecting "Untranslated" filter. Should I add a button to apply filters or may be left the old behaviour?

bychek-ru:
@mathjazz Thanks for your review! Does this work as expected now?

mathjazz:
Sorry it took me so long. We had to ship Firefox Accounts support because it's kind of urgent.

1. I love the Apply selected filters button! It makes it much faster to select multiple filters compared to making a new request after selecting each one of them. It should only be visible when at least one filter is checked.

2. I'm not sure about the Apply selected filters positon. It looks like it's part of Translation status filters. How about you append it to a toolbar at the bottom of the filter menu, similarly as we display a toolbar when multiple strings are selected in 1-column view (e.g. on Pontoon homepage).

3. The 'All' filter should not have a checkbox on hover and should not be selectable as part of the multiple filter selection. When you click on it, it shows all strings and that's all.

4. Clicking anywhere within the filter item (except for the checkbox) should apply the filter instantly. Currently you need to click the filter text, e.g. Suggested.

5. If you click Untranslated, 3 additional filters get selected, but the title says 2 combined filters.

6. Please remove the checkbox in front of translation authors and put it inside the profile picture, center aligned to it. Check out http://localhost:8000/profile/ for inspiration on how to reduce the contrast of the profile image.

mathjazz:
**1. + 2. Apply selected filters:**
* I like the button position on top!
* It sometimes overflows the search placeholder, so I propose a different design, very similar to how we present the action button in the 1-column layout when you select multiple strings: https://cl.ly/3S190D0y1f3D. So full-width and full-height, dark background, and the number of selected filters dynamic and coloured. Mind the plurals (filter/filters) - check out `Pontoon.openBatchEditor()`!
* I suggest you move the HTML element into the menu, so it will appear/disappear with the menu.
* The button still appears when it shouldn't. E.g., if you unselect all filters, it's still there.

**3. + 4. + 5.**
Everything works great now - good job!

**6.**
Works fine now, thanks! The only thing I would replace is the checkbox color `#FD5F60` -> `#AAAAAA`. The current one (red) is used for missing translations, errors and such, so let's use something more neutral. Also, as much as I love transitions, I'd rather remove it to be consistent with other checkboxes.

mathjazz:
**Apply N filters**
- It always says 'Apply 1 filter' (the number doesn't reflect the number of selected filters).
- Should be positioned absolutely above the menu, so it doesn't push the filter menu down. Something like this: https://gist.github.com/mathjazz/2a5bac44963f20f896a73af4e0f34582. (still need to change JS to fix openning/closing)
- When selecting status filters, search placeholder specifies filters that are selected. When other filters are selected, it says 'N combined filters'. We should unify this and always list selected filters.

**New things I didn't notice before (sorry!)**
- When all filters are unchecked, the filter menu should remain open.
- Why are smart filters mutually exclusive (you can only select one)?
- Checkboxes need bigger click area, something like this: https://github.com/mozilla/pontoon/blob/master/pontoon/base/static/css/translate.css#L987.

bychek-ru:
@mathjazz Take a look again please

bychek-ru:
@mathjazz Done with it.

My answer on TODO:
1)I think we should do it like I did in the code:D Multiple filters must be AND filters. But we had to use OR filters for status filters, because for example a string can't be missing and fuzzy at the time (and don't think that we need look for... partially translated plural strings) and when we select several authors (we don't find string on which working 2 or more users). I added more logic for the Untranslated filter. I think it's very useful for translators/contibutors to translate new and fix fuzzy strings. And for managers/translators which can approve new translations and also fix fuzzy and translate new strings (if they exists). One click is always better than two or three. But, of course, it makes the code more complicated.
2)In this case we must use scrollbars. Alternatively I can suggest to move this div to the top (lower than placeholder, but higher than translation status filter) and use `.disabled` class for it.
3)I've deleted repeating things from `setFilter` function.

And also about one thing from changelog:
"Do not use checkbox if only 1 filter clicked (not checked)"
We must mark selection of this filter too to let the users select more filters later.

pontoon/base/static/css/user.css
```diff
@@ -30,6 +30,11 @@
   opacity: 0.3;
 }
 
+#title a img {
+  transition: opacity 0.4s;
+  -webkit-transition: opacity 0.4s;
+}
+
```
bychek-ru: I use it for the better hover effect for the avatar in user's profille.

pontoon/base/static/js/translate.js
```diff
@@ -1063,13 +1037,16 @@ var Pontoon = (function (my) {
             title = node.find('.name').text() + "'s translations";
           }
           placeholder.push(title);
+
+          if (filter[type][i] == "untranslated") {
```
mathjazz: `=== 'untranslated'`: triple `=` and single quotes for consistency.

pontoon/base/static/js/translate.js
```diff
@@ -1063,13 +1037,16 @@ var Pontoon = (function (my) {
...
+            $('#filter .menu [data-type="missing"], #filter .menu [data-type="fuzzy"], #filter .menu [data-type="suggested"]').addClass("selected");
```
mathjazz: Nit: `addClass('selected')`: single quotes.

pontoon/base/static/js/translate.js
```diff
@@ -1160,16 +1135,37 @@ var Pontoon = (function (my) {
...
 
         } else if (el.hasClass('untranslated') || el.hasClass('unchanged') || el.hasClass('has-suggestions')) {
           updateFilterValue('extra');
+          if (value == 'untranslated') filter.status = [];
 
         } else {
-          updateFilterValue('status');
+          // Covering all cases for untranslated filter
+          var num = $.inArray('untranslated', filter.extra);
+          if (num != -1) {
```
mathjazz: `!==`

pontoon/base/static/js/translate.js
```diff
@@ -1160,16 +1135,37 @@ var Pontoon = (function (my) {
...
+            filter.extra.splice(num, 1);
+
+            if (value == 'translated') {
```
mathjazz: `===`

pontoon/base/static/js/translate.js
```diff
@@ -1160,16 +1135,37 @@ var Pontoon = (function (my) {
...
           updateFilterValue('author');
 
         } else if (el.hasClass('untranslated') || el.hasClass('unchanged') || el.hasClass('has-suggestions')) {
           updateFilterValue('extra');
+          if (value == 'untranslated') filter.status = [];
 
         } else {
-          updateFilterValue('status');
+          // Covering all cases for untranslated filter
+          var num = $.inArray('untranslated', filter.extra);
```
mathjazz: Ommit `var`: the variable is already defined above.

pontoon/base/static/js/translate.js
```diff
@@ -1160,16 +1135,37 @@ var Pontoon = (function (my) {
...
+              var a = ['missing', 'fuzzy', 'suggested'];
+              a.splice($.inArray(value, a), 1);
+              filter.status.push(a[0], a[1]);
+            }
+
+          } else {
+            updateFilterValue('status');
+          }
+
+          if ($.inArray('missing', filter.status)!=-1 && $.inArray('fuzzy', filter.status)!=-1 && $.inArray('suggested', filter.status)!=-1) {
```
mathjazz: Replace all three instances of `!=-1` with ` !== -1` (mind the spaces).

mathjazz:
Thanks for the patch!

> 1)I think we should do it like I did in the code:D Multiple filters must be AND filters. But we had to use OR filters for status filters, because for example a string can't be missing and fuzzy at the time (and don't think that we need look for... partially translated plural strings) and when we select several authors (we don't find string on which working 2 or more users). I added more logic for the Untranslated filter. I think it's very useful for translators/contibutors to translate new and fix fuzzy strings. And for managers/translators which can approve new translations and also fix fuzzy and translate new strings (if they exists). One click is always better than two or three. But, of course, it makes the code more complicated.

Good points! Do you think we should replace commas in the placeholder text with AND and OR, to give users a better idea what's going on? Or would that be even more confusing?

Also agree about the untranslated filter.

> 2)In this case we must use scrollbars. Alternatively I can suggest to move this div to the top (lower than placeholder, but higher than translation status filter) and use .disabled class for it.

I have an idea - will try to land a patch tomorrow.

> 3)I've deleted repeating things from setFilter function.

Thanks! I added another improvement over this, cold you review the commit? And the other two - one with a `selectedCount` bugfix and the one with a tiny `preventDefault` improvement?

> And also about one thing from changelog:
> "Do not use checkbox if only 1 filter clicked (not checked)"
> We must mark selection of this filter too to let the users select more filters later.

I'm not entirely sure this is the right behaviour, because it's inconsistent with the string list. But I see your point, so I'm keeping it for now.

bychek-ru:
@mathjazz Looks good to me. I think placeholder now looks by the easiest way and we don't need mark OR/AND there. Have you any thoughts about the login issue which I've shared with you earlier?

bychek-ru:
@mathjazz @jotes As everything works fine at my side now I'm open to your any new ideas about this.
Also I wanna write why I changed the way how do we count selectedCount because of when we select "untranslated" filter it marks that it's equivalent of combination filters missing+fuzzy+suggested. And the filter view shows that we select 4 filters but we apply the only one. But it looks more logical if we count amount of all checkboxes.

mathjazz:
@bychek-ru One final thing. Selecting Untranslated filter and Translated at the same time is not possible. I suggest you do this:
* Remove the Untranslated filter from the python code
* If Untranslated is selected, store it as `filter.status = ['missing', 'fuzzy', 'suggested']` in JS and call these three filters in the URL as part of the `status` param.

Out of curiosity - why do you use `$.inArray(value, array)`? Isn't that the same as native JS `array.indexOf(value)`?

bychek-ru:
@mathjazz "Selecting Untranslated filter and Translated at the same time is not possible."
But the combination of Untranslated and Translated strings equals selection of All filter. Do you really think that we need it?

I don't know what browsers we support, just checked we really don't support IE8, changed to native method.

So do you have ideas to fix tests? I have the simpliest idea is to return functions missing, fuzzy and etc. to EntityQuerySet class (without adding author and time range to tests) because now I've 15 tests broken by new code. After this we can change logic of tests and make it the different PR.

mathjazz:
> @mathjazz "Selecting Untranslated filter and Translated at the same time is not possible."
> But the combination of Untranslated and Translated strings equals selection of All filter. Do you really think that we need it?

Yes. I know selecting all the status options equals all strings, but this way all status checkboxes behave the same, which is what the users should expect, given that look & feel of all of them is the same.

> I don't know what browsers we support, just checked we really don't support IE8, changed to native method.

Thanks!

> So do you have ideas to fix tests? I have the simpliest idea is to return functions missing, fuzzy and etc. to EntityQuerySet class (without adding author and time range to tests) because now I've 15 tests broken by new code. After this we can change logic of tests and make it the different PR.

I'll look into that, but first I think we should update the Untranslated behaviour.

bychek-ru:
@mathjazz Fixed

pontoon/base/static/js/translate.js
```diff
@@ -1005,6 +1006,10 @@ var Pontoon = (function (my) {
             $('#filter .menu [data-type="missing"], #filter .menu [data-type="fuzzy"], #filter .menu [data-type="suggested"]').addClass('selected');
           }
         }
+
+        if (type == 'status' && self.arrayInArray(filter[type], ['missing', 'fuzzy', 'suggested'])) {
+          $('#filter .menu [data-type="untranslated"]').addClass('selected');
+        }
```
mathjazz: This block can be moved outside `markSelectedFilters()`, so it's only exdcuted when neccessary.

pontoon/base/static/js/translate.js
```diff
@@ -1116,6 +1121,30 @@ var Pontoon = (function (my) {
...
+     */
+    arrayInArray: function(array, values) {
+      var value = true;
+      for (var i=0, l=values.length; i<l; i++) {
+        value &= (array.indexOf(values[i]) !== -1) ? true : false;
+      }
+      return value;
+    },
+
+
```
mathjazz: These two functions can be made more specific, e.g. `applyUntranslatedFilter` and `untranslatedFilterApplied`. Both should be passed only one argument - `filter.status`. And `['missing', 'fuzzy', 'suggested']` should be externalized as e.g. `getUntranslatedFilters` and reused. It will help save some lines of code.

mathjazz:
Regarding tests, I think it would be easier if you keep the existing `EntityQuerySet.missing()` and friends (or at least their Q()) as separate methods.

As @jotes has put it, `for_project_locale` will be harder to test & maintain because there will be a lot of nested blocks. We could even use things like this: https://gist.github.com/jotes/a9a17a5b3cd84746b4d9f529a797260b.

bychek-ru:
@mathjazz Jarek's code broken we need the second argument for `map`.
Also the code in the test doesn't work as we want. Do you have any ideas to fix it? I think it's because now we don't call functions from `EntityQuerySet` class.

jotes:
@bychek-ru Yup, I forgot about passing "statuses" as a second parameter, but besides that -> is there anything wrong with the whole idea?

mathjazz:
@bychek-ru, there's 3 things I see you need to do:
1. Remove untranslated from code and tests.
2. Turn `EntityQuerySet.missing()` and friends into functions returning Qs.
3. Fix tests to suit step 2.

Check this out for what steps 2 and 3 could look like:
https://gist.github.com/mathjazz/15066167c611720ed89fc945300c94fd

@jotes Thoughts?

jotes:
I agree with things you pointed out above... I can even help after the serverless conf.

-----Wiadomość oryginalna-----
Od: "Matjaž Horvat" <notifications@github.com>
Wysłano: ‎2016-‎10-‎26 19:06
Do: "mozilla/pontoon" <pontoon@noreply.github.com>
DW: "Jarek" <jot@smiejczak.net>; "Mention" <mention@noreply.github.com>
Temat: Re: [mozilla/pontoon] Bug 1243115 - Add ability to combine differentfilters (#463)

@bychek-ru, there's 3 things I see you need to do:
1. Remove untranslated from code and tests.
2. Turn EntityQuerySet.missing() and friends into functions returning Qs.
3. Fix tests to suit step 2.
Check this out for what steps 2 and 3 could look like:
https://gist.github.com/mathjazz/15066167c611720ed89fc945300c94fd
@jotes Thoughts?

bychek-ru:
@mathjazz Done. But do you have an idea how to correct `test_entity_filters`?

@jotes I think it's a good suggestion. I've thought that you mean `statuses` too. But I got `'instancemethod' object is not iterable` error.

mathjazz:
Thanks, good job!

I added two frontend fixes: 1. The filter toolbar now finally behaves the way I like it - it's always visible. 2. The time range button is less obvious now (it's not as important as it looked previously). Let me know your thoughts on this.

I have a meeting now and then I'll look at `test_entity_filters` and then we can deploy!

pontoon/base/tests/test_views.py
```diff
@@ -557,10 +557,9 @@ def test_entity_filters(self):
             else:
                 params['status'] = filter_
 
-            with patch('pontoon.base.models.Entity.objects.{}'.format(filter_name), return_value=Entity.objects.all()) as filter_mock:
+            with patch('pontoon.base.models.Entity.objects.{}'.format(filter_name), return_value=getattr(Entity.objects, filter_name)()) as filter_mock:
```
mathjazz: The trick is here. Previously we used `Entity.objects.all()` to fake `Entity.objects.missing()` and friends, but this no longer works since we don't return QuerySets, but Q objects.

mathjazz:
From my perspective we're ready to go!

@bychek-ru @jotes Do you have any additional thoughts?

bychek-ru:
Squashed. r+

mathjazz:
@bychek-ru Fan-tastic job! Thanks for keeping up with all the nits and ever-changing course. ;)

mathjazz:
This is an interesting bug:

1. _Suggested_ selects 0 strings (which is correct).
2. The default _Time Range_ selects all 49 strings (which is also correct).
3. When used together, we should get 0 string, but instead we get 14.

https://mozilla-pontoon-staging.herokuapp.com/sl/pontoon-intro/all-resources/?status=suggested&time=201409291607-201606061222&string=33066

The problem is that in example No. 3 all strings which contain suggestions are listed. So it's like calling _Time Range_ and _Has Suggestions_.

We need to fix this before deploying to production.

@bychek-ru, do you want to have a look?

PR #464, opened 2016-09-14
Speed up sync log listing

mathjazz:
Loading /sync page is slow. This should speed it by ~80%.

The main trick is in not calling `SyncLog.finished` property.

@jotes r?


jotes:
I think we could even cache this whole view... but this is still okay :+1:

PR #466, opened 2016-09-16
Bug 1299084: fix performance regression in contributor roles

mathjazz:
@jotes, @bychek-ru r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -89,7 +89,27 @@ class UserTranslationsManager(models.Manager):
                 query &= query_filters
 
             return self._changed_translations_count(query)
-        return (
+
+        def with_roles(self):
+            """Add user roles."""
+            managers = {}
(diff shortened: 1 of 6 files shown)


pontoon/base/models.py
```diff
@@ -89,7 +89,27 @@ def translations_count(query=None):
                 query &= query_filters
 
             return self._changed_translations_count(query)
-        return (
+
+        def with_roles(contributors):
```
jotes: I would change contributors into `self` just for the sake of readablity.

pontoon/base/models.py
```diff
@@ -192,19 +203,36 @@ def user_managed_locales(self):
     return [locale.code for locale in locales]
 
 
-@property
-def user_rights(self):
+def user_role(self, managers=None, translators=None):
```
jotes: Could you add a comment that explains performance issues and why cases user_roles() helps to minimize number of queries?

PR #469, opened 2016-10-04
Fixed merging of accounts.

jotes:
@mathjazz Hi, could you look at this patch? This should solve an issue mentioned by flod but it needs testing the rest of scenarios.

### DIFF
--- a/pontoon/base/adapter.py
+++ b/pontoon/base/adapter.py
@@ -1,6 +1,7 @@
 import base64
 import hashlib
 
+from allauth.account.adapter import get_adapter
 from allauth.socialaccount.adapter import DefaultSocialAccountAdapter
 
 from django.contrib.auth.models import User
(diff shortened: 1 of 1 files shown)


pontoon/base/adapter.py
```diff
@@ -29,30 +32,42 @@ def save_user(self, request, sociallogin, form=None):
...
+        migration_message = 'Your Persona account and Firefox Account have been connected.'
         if not email:
             return
 
         try:
-            user = User.objects.get(email=email)
+            user = User.objects.get(email=email, is_active=True)
         except User.DoesNotExist:
+            if request.user.is_authenticated():
+                user = None
```
mathjazz: We don't need to set the `user` because we `return` in the following line.

mathjazz:
r+ :)

PR #471, opened 2016-10-04
List resources not bound to subpages

mathjazz:
@jotes r?

What is this?

Currently, if a project has subpages defined, we only list those pages in locale-project pages and in the parts menu. If new resources are added to repository, we don't list them, but we show their strings when "All Resources" is selected. So much for the consistency.

With this patch, we also list all other resources available for locale, that are not bound to any subpage. This helps us when new resources are added to project without a notice to update the project setup.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -482,6 +482,7 @@ class Locale(AggregatedStats):
             locale=self
         ).distinct()
         details = []
+        unbound_details = []
 
         # If subpages aren't defined,
         # return resource paths with corresponding stats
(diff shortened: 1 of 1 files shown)


pontoon/base/models.py
```diff
@@ -521,14 +528,22 @@ def get_details(parts):
...
+
         all_resources = ProjectLocale.objects.get(project=project, locale=self)
         all_paths = (
             TranslatedResource.objects
             .filter(resource__project=project, locale=self)
             .values_list("resource__path", flat=True)
         )
 
-        details_list = list(details)
+        details_list = list(details) + list(unbound_details)
```
jotes: [nit] I think we could move list() into get_details function.

PR #472, opened 2016-10-05
Fix connecting FxA and Persona accounts with different email addresses

mathjazz:
Alternative proposal for #469 - without migrating User data.

- [x] Instead of logout, login as a merged (Persona) user
- [ ] Show connected message when connecting Persona account to FxA with different email

@jotes r?


mathjazz:
Dupe of #469.

PR #480, opened 2016-10-13
Fix csp

mathjazz:
This PR does 3 things in 3 separate commits:

1. Restores CSP (yay!). It's a simple revert of commit revert.
2. Make Pontoon work locally by adding http:// protocol to some CSP rules.
3. And finally, makes Persona sign in work by moving JS from the `href` attribute to `main.js`.

@marumari r?


marumari:
Looks good to me!

PR #481, opened 2016-10-16
Bug 1310510 - add virtualenv and fix CSP

bychek-ru:
1)We must modify django-debug-toolbar or add ajax.googleapis.com in CSP, because requirements-dev.txt contains django-debug-toolbar which needs access to ajax.googleapis.com
2)we need "data:" for base64 images. For example:

3)About Virtualenv

### DIFF
--- a/docs/dev/install.rst
+++ b/docs/dev/install.rst
@@ -19,7 +19,7 @@ all the prerequisites using the following command:
 
    .. code-block:: bash
 
-      sudo apt install git python-pip nodejs-legacy npm postgresql postgresql-server-dev-9.5 postgresql-contrib-9.5 libxml2-dev libxslt1-dev python-dev libmemcached-dev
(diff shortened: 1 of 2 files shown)


mathjazz:
As @jotes pointed out in the bug, let's override the CSP rules in the `pontoon/settings/dev.py` file, so that we keep the production and stage CSP unchanhged: 

```
CSP_SCRIPT_SRC = base.CSP_SCRIPT_SRC + ('http://ajax.googleapis.com',)
CSP_IMG_SRC = base.CSP_IMG_SRC + ('data:',)
```

We actually have virtualenv listed as a requirement, but it's not present in that `apt-get install` snippet, so good move in adding it there, thanks!

bychek-ru:
@mathjazz @jotes r?

mathjazz:
Good job, r+!

PR #483, opened 2016-10-22
Fix bug 1310510 - fix running server command

bychek-ru:


docs/dev/install.rst
```diff
@@ -129,7 +129,7 @@ running:
 
 .. code-block:: bash
 
-   python manage.py runserver
+   ./manage.py runserver localhost:8000
```
jotes: Our current FXA setup rely on the fact that developers run their instances at http://localhost:8000 (this is also a default value for the django runserver command).

mathjazz: Good point! The FxA redirect URL is `localhost` right?

We should change the `SITE_URL` section in step 4 to something like this:

```
   - ``SITE_URL`` should be set to the URL you will use to connect to your
     local development site. Some people prefer to use
     ``http://127.0.0.1:8000`` instead of ``localhost``. However, should you
     decide to change the ``SITE_URL``, you also need to
     [request](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/Firefox_Accounts/Introduction)
     the new ``FXA_CLIENT_ID``and ``FXA_SECRET_KEY``.
```

PR #484, opened 2016-10-23
docs/conf.py: remove unused boilerplate

stepshal:


mathjazz:
Hi @stepshal! Thanks for the patch!

Commenting out code in combination with using a VCS system like Git is an anti-pattern, so I understand your intention here. However, since this is a configuration file, it makes sense to keep unused settings in the file, so it'll be easier to turn them on if we'll need them in the future.

What do you think? What was your idea bihnd removing unused settings from `docs/conf.py`?

stepshal:
@mathjazz to get rid of the pycodestyle warning just  by removing the commented-out code.

mathjazz:
In this case I suggest we keep the config as it is.

PR #486, opened 2016-10-24
Add help text to plural-related Locale fields

mathjazz:
@flodolo r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -355,7 +355,16 @@ class LocaleQuerySet(models.QuerySet):
 class Locale(AggregatedStats):
     code = models.CharField(max_length=20, unique=True)
     name = models.CharField(max_length=128)
-    plural_rule = models.CharField(max_length=128, blank=True)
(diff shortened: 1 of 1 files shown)


pontoon/base/models.py
```diff
@@ -355,7 +355,16 @@ def prefetch_latest_translation(self, project):
...
     name = models.CharField(max_length=128)
-    plural_rule = models.CharField(max_length=128, blank=True)
+    plural_rule = models.CharField(
+        max_length=128,
+        blank=True,
+        help_text="""
+        Plural rule is part of the plurals header in
+        <a href="https://www.gnu.org/software/gettext/manual/gettext.html#Plural-forms">Gettext PO files</a>,
+        following the <i>plural="</i> string, without the trailing semicolon.
+        E.g. (n != 1).
```
flodolo: The final period here is confusing

mathjazz:
Thanks!

If the `cldr_plurals` field is empty, the plural tabs are not displayed, so the user can only provide one form in pluralized strings.

PR #487, opened 2016-10-29
Fix bug 1243115 - fixed issue with suggested & time-range filter

bychek-ru:
@mathjazz I fixed [this issue](https://github.com/mozilla/pontoon/pull/463#issuecomment-257048719). Also make `authored_by` and `between_time_interval` to return Qs.
I don't know why but time range filter requires to apply first, in this way your combination works fine.
r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -1254,15 +1254,21 @@ class EntityQuerySet(models.QuerySet):
     def unchanged(self):
         return Q(unchanged_count=F('expected_count'))
 
-    def authored_by(self, locale, email):
-        if ',' in email:
-            email = email.split(',')
-            return self.filter(translation__locale=locale).filter(reduce(lambda x, y: x | y, [Q(translation__user__email=item) for item in email]))
+    def authored_by(self, email):
+        email = email.split(',')
+        query = None
 
-        return self.filter(translation__locale=locale, translation__user__email=email)
+        for e in email:
+            q = Q(translation__user__email=e)
+            if query:
+                query |= q
+            else:
+                query = q
 
-    def between_time_interval(self, locale, start, end):
-        return self.filter(translation__locale=locale, translation__date__range=(start, end))
+        return query
+
+    def between_time_interval(self, start, end):
+        return Q(translation__date__range=(start, end))
 
     def prefetch_resources_translations(self, locale):
         """
@@ -1362,9 +1368,9 @@ class Entity(DirtyFieldsMixin, models.Model):
         search=None, exclude=None, extra=None, time=None, author=None):
         """Get project entities with locale translations."""
         entities = Entity.objects.all()
+        query = None
 
         if status:
-            query = None
             status = status.split(',')
 
             for s in status:
@@ -1385,11 +1391,7 @@ class Entity(DirtyFieldsMixin, models.Model):
                 else:
                     query = q
 
-            if query:
-                entities = entities.with_status_counts(locale).filter(query)
-
         if extra:
-            query = None
             extra = extra.split(',')
 
             for s in extra:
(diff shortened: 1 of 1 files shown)


pontoon/base/models.py
```diff
@@ -1255,14 +1255,20 @@ def unchanged(self):
         return Q(unchanged_count=F('expected_count'))
 
     def authored_by(self, locale, email):
```
mathjazz: You can remove the `locale` attribute.

pontoon/base/models.py
```diff
@@ -1255,14 +1255,20 @@ def unchanged(self):
...
 
-        return self.filter(translation__locale=locale, translation__user__email=email)
+        for e in email:
+            q = Q(translation__user__email=e)
+            if query:
+                query |= q
+            else:
+                query = q
+
+        return Q(translation__locale=locale) & (query);
```
mathjazz: `return query` is enough.

Also, no `;` at the end of python statement. :-)

pontoon/base/models.py
```diff
@@ -1255,14 +1255,20 @@ def unchanged(self):
...
 
     def between_time_interval(self, locale, start, end):
-        return self.filter(translation__locale=locale, translation__date__range=(start, end))
+        return Q(translation__locale=locale, translation__date__range=(start, end))
```
mathjazz: `return Q(translation__date__range=(start, end))` is all you need.

pontoon/base/models.py
```diff
@@ -1255,14 +1255,20 @@ def unchanged(self):
...
+        for e in email:
+            q = Q(translation__user__email=e)
+            if query:
+                query |= q
+            else:
+                query = q
+
+        return Q(translation__locale=locale) & (query);
 
     def between_time_interval(self, locale, start, end):
```
mathjazz: You can remove the `locale` attribute.

pontoon/base/models.py
```diff
@@ -1385,11 +1391,9 @@ def for_project_locale(self, project, locale, paths=None, status=None,
                 else:
                     query = q
 
-            if query:
-                entities = entities.with_status_counts(locale).filter(query)
+            query = (query)
```
mathjazz: You can remove this statement.

pontoon/base/models.py
```diff
@@ -1404,23 +1408,28 @@ def for_project_locale(self, project, locale, paths=None, status=None,
...
 
         if author:
-            entities = entities.authored_by(locale, author)
+            q = Entity.objects.authored_by(locale, author)
+
+            if query:
+                query &= q
+            else:
+                query = q
+
```
mathjazz: Replace this entire block with `entities = entities.filter(Entity.objects.authored_by(author))`. It will fix the same issue as with suggested & time range, but for authors.

bychek-ru:
@mathjazz Updated. I mean I don't know why we need to apply time range filter before status filters.
P.S. I didn't catch the same bug for the suggested and authors, and before, and after my changes:D But, of course, you suggestion `entities = entities.filter(Entity.objects.authored_by(author))` makes the code shorter.

mathjazz:
Thanks, great job!

mathjazz:
Still doesn't work:
https://mozilla-pontoon-staging.herokuapp.com/sl/pontoon-intro/all-resources/?status=suggested&author=miha.kosi%40amis.net&string=33050

Tried Suggested with one of the other two users and also with time range and it worked.
But not the combination linked above.

bychek-ru:
I don't have any ideas about this bug...
If I even comment all code for filters in `for_project_locale` and write just
`entities = entities.with_status_counts(locale).filter(Q(suggested_count__gt=0) & ~Q(fuzzy_count=F('expected_count')) & ~Q(approved_count=F('expected_count')) & Q(translation__user__email='a@bychek.ru'))`
it doesnt't work correctly. This problem is `suggested` filter
`Q(suggested_count__gt=0) & ~Q(fuzzy_count=F('expected_count')) & ~Q(approved_count=F('expected_count'))` I think.

mathjazz:
OK, I've reverted the combined filters commit for now, so I could deploy the History count fix.

Adding @jotes who's also looking into the issue.

Let me know if I can help you work this out.

PR #492, opened 2016-11-03
Switch back to latest upstream django-allauth

mathjazz:
The bug our fork fixes is now fixed upstream, let's move back to using the official branch.

@jotes r? Is there a reason why we were using a pretty old version - 0.25.2? Maybe because we started working on FxA support before 0.26.0 was release?

BTW, shall we also remove all social accounts with provider Persona?


jotes:
I think that was the latest version at the time :D I think we can remove sociall accounts related to the persona and obviously remove persona from the settings.py just to be double sure.

PR #495, opened 2016-11-10
Make download prefix field mandatory

mathjazz:
@jotes r?

This makes the `permalink_prefix` and `url` fields mandatory and adds some helper text.


pontoon/base/models.py
```diff
@@ -890,18 +891,17 @@ class Repository(models.Model):
...
         choices=TYPE_CHOICES
     )
-    url = models.CharField("URL", max_length=2000, blank=True)
-
-    """
-    Prefix of the resource URL, used for direct downloads. To form a full
-    URL, relative path must be appended.
-    """
-    permalink_prefix = models.CharField("Permalink prefix", max_length=2000, blank=True)
+    url = models.CharField("URL", max_length=2000)
```
jotes: I think we can use [URLField](https://docs.djangoproject.com/en/1.10/ref/models/fields/#urlfield)

mathjazz: Sadly we can't. I remember trying that, but it didn't recognize the GitHub SSH-based URLs as valid.

PR #496, opened 2016-11-10
Add support for .xlf

mathjazz:
An Alternative .xliff file extension

@jotes r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -1115,6 +1115,7 @@ class Resource(models.Model):
     FORMAT_CHOICES = (
         ('po', 'po'),
         ('xliff', 'xliff'),
+        ('xlf', 'xliff'),
         ('properties', 'properties'),
         ('dtd', 'dtd'),
         ('inc', 'inc'),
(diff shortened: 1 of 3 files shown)


pontoon/sync/vcs/models.py
```diff
@@ -343,10 +343,7 @@ def relative_resource_paths(self):
         """
         path = self.source_directory_path
         for absolute_path in self.resources_for_path(path):
-            # .pot files in the source directory need to be renamed to
-            # .po files for the locale directories.
-            if absolute_path.endswith('.pot'):
-                absolute_path = absolute_path[:-1]
+            absolute_path = source_to_locale_path(absolute_path)
```
mathjazz: Not part of the bug this is fixing, but I notice it while fixing it.

pontoon/base/models.py
```diff
@@ -1158,7 +1159,12 @@ def get_path_format(self, path):
         path_format = extension[1:].lower()
 
         # Special case: pot files are considered the po format
-        return 'po' if path_format == 'pot' else path_format
+        if path_format == 'pot':
```
jotes: I'd use a dict instead of a "switch" statement.

PR #499, opened 2016-11-11
Load jquery.mark.js lib before the script that uses it

mathjazz:
It's causing JS error in local development randomly:
`TypeError: item.unmark is not a function  translate.js:3306:7`

@jotes r?


jotes:
R+ :)

PR #501, opened 2016-11-16
Fix bug 1316171: JS_TRUSTED_ORIGINS can be loaded from ENV.

jotes:
@mathjazz I think this can be helpful on the production or even on the staging.


mathjazz:
Agreed!

And with that, we can also remove `localHostnames` from `pontoon.js`, right?

jotes:
Yup :)

16 lis 2016 09:24 "Matjaž Horvat" <notifications@github.com> napisał(a):

mathjazz:
r+

PR #506, opened 2016-11-24
Add support for downloading Firefox language packs

mathjazz:
@jotes r?

Adds the ability to download Firefox language packs from:
https://github.com/mathjazz/pontoon-langpacks

The repository is being updated every 20 minutes using the `make-langpacks.py` script.


bin/make-langpacks.py
```diff
@@ -0,0 +1,152 @@
...
+    'locales_url': 'https://api.bitbucket.org/2.0/repositories/mozilla-l10n'
+}]
+
+
+def write(text):
+    timestamp = datetime.datetime.now().strftime('[%Y-%m-%d %H:%M:%S] ')
+    print(timestamp + text)
+
+
+def execute(command, cwd=None):
```
jotes: Did you want to have something similar to subprocess.check_call?

mathjazz: Thanks for bringing it up! It doesn't seem to include the error message, which we also need.

PR #511, opened 2016-12-02
Strip whitespace to catch all IPs

mathjazz:
@jotes r?

I noticed we weren't catching all IPs during a ddos attempt.

### DIFF
--- a/pontoon/base/middleware.py
+++ b/pontoon/base/middleware.py
@@ -40,6 +40,8 @@ class BlockedIpMiddleware(object):
         except KeyError:
             ip = request.META['REMOTE_ADDR']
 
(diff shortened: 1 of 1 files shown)


pontoon/base/middleware.py
```diff
@@ -40,6 +40,8 @@ def process_request(self, request):
         except KeyError:
             ip = request.META['REMOTE_ADDR']
 
+        ip = ip.strip()
```
jotes: Do you have any examples of those values?

mathjazz: Yeah, I saw IP addresses starting with a single space not being caught, e.g. " abc.def.ghi.jkl".

PR #515, opened 2016-12-11
Bug 1225109: added test for contributors profile page.

jotes:
@mathjazz couple of tests related to the implementation of new timeline.


mathjazz:
r+

PR #519, opened 2016-12-19
bug_1223945_locale_project_privileges

jotes:
It's the alpha version and I still need to finish some things before I can pass it to final review.

**TODO**
- [x] Simplify team permissions page: only show general permissions by default and allow team managers to add a special entry for each project and override permissions for that project.
- [x] Start actually using project-locale permissions.
- [x] Add a /LOCALE/manage redirect to /LOCALE/permission to avoid breaking legacy links.
- [x] Style remove project button (maybe move to the title bar).
- [x] Fix performance on http://localhost:8000/sl/permissions/.
- [x] When you add custom project permissions, copy current Translators from General. [@mathjazz]
- [x] If user listed under General/Translators, but not under Project X/Translators, it shouldn't have permissions for Project X. That allows us to grant Translator permissions to a user for all projects, except for Project X. It's also not possible to save permissions for Project X without any Translator selected.


pontoon/base/templates/locale_permissions.html
```diff
@@ -0,0 +1,48 @@
+{% extends 'locale.html' %}
+{% import 'widgets/permissions_groups.html' as PermissionsGroups with context %}
+{% block title %}{{ super() }} &middot; Manage {% endblock %}
```
mathjazz: Manage -> Permissions

jotes:
@mathjazz I think it is ready to a next round of review.

mathjazz:
Before I start the review, it seems like something is broken:
* on my local instance http://localhost:8000/sl/permissions/ takes almost a minute to load (probably because of 300+ DB queries)
* the CSS and JS are also broken: https://cl.ly/1b0K101C3u3v

Maybe you forgot to push latest updates?

mathjazz:
Thanks for the updates, @jotes. I haven't looked at the code yet, but it seems like we're getting closer!

**TODO**
- [ ] When you add custom project permissions, copy current Translators from General. [@mathjazz]
- [ ] Style remove project button (maybe move to the title bar). [@mathjazz]
- [ ] If user listed under General/Translators, but not under Project/Translators, it shouldn't have permissions for that project. That allows us to grant Translator permissions to a user for all projects, except for project X.
- [ ] Fix performance on http://localhost:8000/sl/permissions/.

pontoon/base/templates/translate.html
```diff
@@ -143,7 +144,6 @@
       </div>
     </div>
 
-  </div>
```
mathjazz: Restore.

pontoon/base/templates/locale_permissions.html
```diff
@@ -0,0 +1,98 @@
...
+                {{ manager_selector() }}
+            </div>
+        </div>
+
+        {{ project_locale_form.management_form }}
+        {% for pk, slug, name, all_users, translators, contributors, has_custom_translators in locale_projects %}
+            <div class="permissions-groups project-locale clearfix"
+                 data-slug="{{ slug }}"
+                 data-index="{{ loop.index0 }}"
+                 {% if not has_custom_translators  %} style="display:none;" {% endif %}>
```
mathjazz: We should use classes instead of inline CSS.

pontoon/base/static/js/manage_permissions.js
```diff
@@ -0,0 +1,136 @@
...
+    // Scroll to the right project locale
+    $('html, body').animate({
+      scrollTop: $permsForm.offset().top
+    }, 500);
+
+    $(this).hide();
+
+    if ($('#project-selector .menu li:visible').length === 0) {
+      $('#project-selector').hide();
+    }
```
mathjazz: Move above Copying Translators to avoid hidding selector after selection is made.

pontoon/base/static/js/translate.js
```diff
@@ -3214,7 +3207,18 @@ var Pontoon = (function (my) {
 
       this.part = this.getSelectedPart();
       this.locale = self.getLocaleData();
-      this.user.isTranslator = isTranslator($('#server').data('user-translated-locales'));
+
+      this.user.canTranslate = function() {
+        var translatedLocales = $('#server').data('user-translated-locales') || [],
+            translatedProjects = $('#server').data('user-translated-projects') || {},
+            localeProject = self.locale.code + '-' + self.project.slug;
+        debugger;
```
mathjazz: Remove.

pontoon/base/models.py
```diff
@@ -472,6 +507,93 @@ def cldr_id_to_plural(self, cldr_id):
...
+                    translation__entity__resource__project__pk__in=[project_locale['project__pk'] for project_locale in project_locales]
+                )
+                .exclude(email='')
+                .prefetch_related('translation__entity__resource__project')
+                .values('id', 'first_name', 'email', 'translation__entity__resource__project__pk')
+                .distinct(),
+            'translation__entity__resource__project__pk'
+        )
+
+        projects_all_users = defaultdict(set)
```
mathjazz: This is actually all users without `projects_translators`. Shall we simply use exclude?
https://github.com/mozilla/pontoon/blob/master/pontoon/base/views.py#L129

pontoon/base/signals.py
```diff
@@ -54,53 +54,73 @@ def project_deleted(sender, **kwargs):
...
+    group, _ = Group.objects.get_or_create(name='{} {}'.format(name_prefix, group_name))
+
+    for perm_name in perms:
+        perm = Permission.objects.get(content_type=ct, codename=perm_name)
+        group.permissions.add(perm)
+
+        setattr(instance, '{}_group'.format(group_name), group)
+
+
+def assign_group_permissions(instance, group_name, perms):
```
mathjazz: Can't we join this function with `create_group`? And then we could also get rid of `assign_locale_group_permissions` and `assign_project_locale_group_permissions`.

mathjazz:
Great job, @jotes! r+

I identified some nits and fixed them.

I also left 3 comments, could you please have a look?

pontoon/base/models.py
jotes: Yup, but we still need to create a map of those users for every projectlocale (?).

pontoon/base/signals.py
jotes: goodpoint, will do!

pontoon/base/migrations/0075_create_project_locale_translators_groups.py
```diff
@@ -0,0 +1,51 @@
...
+    ContentType = apps.get_model('contenttypes', 'ContentType')
+    GroupObjectPermission = apps.get_model('guardian', 'GroupObjectPermission')
+
+    project_locale_ct = ContentType.objects.get(app_label='base', model='projectlocale')
+
+    can_translate = Permission.objects.get(content_type=project_locale_ct, codename='can_translate_project_locale')
+
+    for project_locale in ProjectLocale.objects.all():
+        translators_group = Group.objects.create(name='{}/{} translators'.format(
+            project_locale.project.name, project_locale.locale.code,
```
jotes: Yup!

pontoon/base/models.py
mathjazz: Correct, let's ignore this comment.

pontoon/base/signals.py
jotes: Okay... I looked at this code again and apparently we cannot those functions because:
create_group is called in pre-save signal (because translators_group is not-null field in locale/project locale)
assign_group_permissions is done in post-save signal (because of django-guardian).

splewako:
Is it normal that I get "Oops Something is wrong. We're going to fix it soon." message after trying to "Add custom permissions for project" and saving?

mathjazz:
No. I see server errors.

Let me have a look and I'll get back once it's fixed.

Thanks for reporting!

mathjazz:
PR ready:
https://github.com/mozilla/pontoon/pull/525

Will let you know once it's fixed. Thanks for spotting a bug!

splewako:
Thank you!

mathjazz:
Fixed. Could you please try again?

splewako:
Works without errors now, thanks.

PR #520, opened 2016-12-23
Fix bug 1305281: Support search across all projects

mathjazz:
@jotes r?

### DIFF
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -131,12 +131,8 @@ def locale_project_parts(request, locale, slug):
 @utils.require_AJAX
 def authors_and_time_range(request, locale, slug, part):
     locale = get_object_or_404(Locale, code=locale)
+    project = get_object_or_404(Project.objects.available(), slug=slug)
     paths = [part] if part != 'all-resources' else None
-
-    if slug == 'all-projects':
-        project = Project(slug=slug)
-    else:
-        project = get_object_or_404(Project.objects.available(), slug=slug)
 
     translations = Translation.for_locale_project_paths(locale, project, paths)
 
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -116,7 +116,11 @@ def locale_projects(request, locale):
 def locale_project_parts(request, locale, slug):
     """Get locale-project pages/paths with stats."""
     locale = get_object_or_404(Locale, code=locale)
-    project = get_object_or_404(Project, slug=slug)
+
+    if slug != 'all-projects':
+        project = None
+    else:
+        project = get_object_or_404(Project, slug=slug)
 
     try:
         return JsonResponse(locale.parts_stats(project), safe=False)
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -117,7 +117,7 @@ def locale_project_parts(request, locale, slug):
     """Get locale-project pages/paths with stats."""
     locale = get_object_or_404(Locale, code=locale)
 
-    if slug != 'all-projects':
+    if slug == 'all-projects':
         project = None
     else:
         project = get_object_or_404(Project, slug=slug)
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -77,16 +77,20 @@ def home(request):
 def translate(request, locale, slug, part):
     """Translate view."""
     locale = get_object_or_404(Locale, code=locale)
-    project = get_object_or_404(Project.objects.available(), slug=slug)
-
-    if locale not in project.locales.all():
-        raise Http404
 
     projects = (
         Project.objects.available()
         .prefetch_related('subpage_set')
         .order_by('name')
     )
+
+    if slug.lower() == 'all-projects':
+        project = Project(name='All Projects', slug=slug.lower())
+
+    else:
+        project = get_object_or_404(Project.objects.available(), slug=slug)
+        if locale not in project.locales.all():
+            raise Http404
 
     return render(request, 'translate.html', {
         'download_form': forms.DownloadFileForm(),
@@ -225,8 +229,13 @@ def entities(request):
     if not form.is_valid():
         return HttpResponseBadRequest(form.errors.as_json())
 
-    project = get_object_or_404(Project, slug=form.cleaned_data['project'])
     locale = get_object_or_404(Locale, code=form.cleaned_data['locale'])
+
+    project_slug = form.cleaned_data['project']
+    if project_slug == 'all-projects':
+        project = Project(slug=project_slug)
+    else:
+        project = get_object_or_404(Project, slug=project_slug)
 
     # Only return entities with provided IDs (batch editing)
     if form.cleaned_data['entity_ids']:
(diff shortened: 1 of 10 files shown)


pontoon/base/views.py
```diff
@@ -440,9 +444,13 @@ def entities(request):
     except (MultiValueDictKeyError, ValueError) as err:
         return HttpResponseBadRequest('Bad Request: {error}'.format(error=err))
 
-    project = get_object_or_404(Project, slug=project)
     locale = get_object_or_404(Locale, code__iexact=locale)
 
+    if project == 'all-projects':
+        project = Project(slug=project)
```
jotes: I'm kinda worried that at some point some part of the code will mutate (e.g. .save) this object.Maybe we should create a separate instance of project that would be loaded in case when slug is all-projects? That would decrease the risk of spawning multiple new projects in future and would be less additional conditions on backend side.

jotes: Also, all new locales could be automatically added to this project.

mathjazz: Yeah, I was thinking about creating a fake `all-projects` instance of the `Project` model, too. 

I'd also like to get rid of the `if project == all-projects` checks if possible.

And, make sure this actually performs in production.

mathjazz:
As you hinted, this doesn't perform well with prod DB. The problem is that `Entity.for_project_locale()` is expensive.
```
entities = Entity.for_project_locale(
    project, locale, paths, status, search, exclude_entities, extra, time, author
)
```
It needs to evaluate on top of all strings (even if we actually only need 50 thanks to pagination). It works good enough for a single project, even a big one, but not for a locale with many projects enabled.

I wonder what options do we have here. Refactor query? Refactor data model? Elasticsearch?

jotes:
Ok, I'm currently busy at my day work (I'll be around 8PM at irc) so I'll put my findings here:
Status filters are probably slowest in our app. From what I saw in SQL Explain, it looks like Postgresql does Sequential SCAN on Entities/Translations tables and that's a really slow option.
I tried to create an index on fields containing strings/translations but it looks like there's too much data to fit in a single index page and Postgresql throws an Error with a Hint that suggest to create smaller fields that will contain md5 hash of field create index on them.
For easier tracking what I'm doing I'll keep my todo list in this PR:

[ ] create hash fields for strings with indexes
[ ] use full text indexes
[ ] check partitioning in postgresql
[ ] denormalize status fields by additional model.
[ ] research Elasticsearch

And I'll ping you if any of those things gives us significant speedups (I'll try to tackle this after today's workday and probably during the weekend).

mathjazz:
Thanks very much for looking into it!

adngdb:
Is this still blocked on performance issues? What is needed for it to move forward?

mathjazz:
Yeah, bug 1305281 is still blocked by bug 1377099.

adngdb:
@mathjazz I've been playing with this locally and it seems to be quite fast. Wanna give it another try? 

Also, I have force-pushed my rebase to your branch, so you don't have to do it again. You'll have to force-pull on your local branch to avoid conflicts and stuff.

jotes:
@adngdb 🎉  I believe that's because of the recent optimizations. 🕺 r+

mathjazz:
> @mathjazz I've been playing with this locally and it seems to be quite fast. Wanna give it another try?

That's weird, it does work surprisingly well. I'll give it another look tomorrow, because I'm not sure why are the filters so fast (I thought they'd fallback to the *old* variant of the code, which is supposed to be slow).

I see a few other things to do:
1. When you open the filter menu, the time range and authors filter don't show up, because the `/authors-and-time-range/ ` 404s.
2. We should link the progress chart legend items in the heading of the [team dasboard](http://localhost:8000/fr/) to the translate view, similarly to how these items are linked in the [localization dashboard](http://localhost:8000/fr/firefox/).
3. We should link items in the Progress column of the [Teams dashboard](http://localhost:8000/teams/) to the translate view as well.

And I wonder what should we do with the Resource menu in the translate view when "All Projects" are selected in the project menu. Hide it? Listing all Resources is not really usable, because there's too many of them.

mathjazz:
> That's weird, it does work surprisingly well. I'll give it another look tomorrow, because I'm not sure why are the filters so fast (I thought they'd fallback to the old variant of the code, which is supposed to be slow).

It turns out `no_plurals` in https://github.com/mozilla/pontoon/blob/master/pontoon/base/models.py#L1977 is always `True` when `All Projects` are selected. That explains why things work fast - the fast filters code is always used, even when it shouldn't be, because strings might be pluralized.

If I manually set `no_plurals` to `False`, the Suggested filter for all projects in the Slovenian locale takes over 2 minutes on my local machine.

We should rewrite filters, so that the same code will work fast for all projects, regardless if they contain pluralized strings or not.

adngdb:
@mathjazz I've rebased this branch, resolved the conflict and pushed to your branch.

mathjazz:
Thanks!

adngdb:
I'm noticing a few problems with this as I'm testing it with #728: 

 * sometimes during navigation, the JS code bugs and tries to send the entire list of paths, which is way too big for the server to handle and results in a 400
 * when clicking on the path of a string, the link goes to ``/locale/all-projects/path`` and that is not valid, as ``all-projects`` doesn't handle paths. We should instead link to the actual project that path belongs to

mathjazz:
> * sometimes during navigation, the JS code bugs and tries to send the entire list of paths, which is way too big for the server to handle and results in a 400
> * when clicking on the path of a string, the link goes to /locale/all-projects/path and that is not valid, as all-projects doesn't handle paths. We should instead link to the actual project that path belongs to

I'm unable to reproduce those. Could you provide exact steps? Note that I pushed some changes which might have had impact on that.

I have 3 relatively simple remaining items on the TODO list:
- [ ] Loading of Resources for the Resources menu is slow, making the translate view (for `/all-projects/` URLs) and selecting All Projects in the projects menu slow. We should fix this by removing the Resource menu, because it's useless with so many paths (incl. duplicates) anyways.
- [ ] Loading of translation authors in the filter menu is slow. We should speed it up or drop it when All Projects are selected.
- [ ] Entity Resource paths in the source string metadata are not linked when All Projects are selected.

adngdb:
@mathjazz I do not know why, but simply loading http://localhost:8000/fr/all-projects/all-resources/ does it. It shows an error, and in the console I can see that the JS code sends the entire list of all paths to the server, which errors out. I really don't get it, I want it to be a cache error but cleaning the cache doesn't solve it. The only way I can fix that issue is by changing to a different locale. Then everything works smoothly. 

You've fixed the second problem by allowing to set a path when "All Projects" is selected. I'm not sure if that's a good thing, but I'll trust you on that. :)

mathjazz:
Reproduced, thanks! The `TooManyFieldsSent` exception is raised for `fr`. Looks like all other locales are good, which is not surprising - `fr` is the only locale for which all projects are enabled.

adngdb:
Closing for now, please reopen when it's ready for review again!

PR #524, opened 2017-01-10
Fix uploading files, 1c24429 regression

mathjazz:
Thanks @ibrahimasarr for reporting!

### DIFF
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -1251,7 +1251,7 @@ def upload(request):
         raise Http404
 
     locale = get_object_or_404(Locale, code=code)
-    project = get_object_or_404(Project, slug=code)
+    project = get_object_or_404(Project, slug=slug)
 
     if not request.user.can_translate(project=project, locale=locale):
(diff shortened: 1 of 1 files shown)


ibrahimasarr:
Thanks for fixing!
Looks like you cannot updload .po files, only .properties??

mathjazz:
Shouldn't be the case. Just tried Firefox Accounts (.po) and it work.

ibrahimasarr:
Works now. But I did get the message .po Files not supported. Use .properties (in red capitals)

mathjazz:
I guess you tried to upload .po file to a project that uses .properties.

ibrahimasarr:
Actually it was the same .po on the same project (Firefox Accounts) file
that worked later: client.po
Thanks

Le 10 janv. 2017 2:14 PM, "Matjaž Horvat" <notifications@github.com> a
écrit :

I guess you tried to upload .po file to a project that uses .properties.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/mozilla/pontoon/pull/524#issuecomment-271572428>, or mute
the thread

PR #526, opened 2017-01-12
Fix bug 1215874: Markup &nbsp; and other special spaces

mathjazz:
@jotes r?

I implemented this with JS, because I think one day we should rewrite original string placeables (to avoid false positives) and move the code to frontend. Should be faster.

Having this code in JS should also help us with implementing diff view for suggestions:
https://bugzilla.mozilla.org/show_bug.cgi?id=1329003

### DIFF
--- a/pontoon/base/static/js/main.js
+++ b/pontoon/base/static/js/main.js
@@ -79,7 +79,25 @@ var Pontoon = (function (my) {
      * string String that has to be displayed as is instead of rendered
      */
     doNotRender: function (string) {
-      return $('<div/>').text(string).html()
+      return $('<div/>').text(string).html();
+    },
+
+    /*
+     * Markup placeables
+     */
+    markPlaceables: function (string) {
+      function markup(string, regex, title) {
+        return string.replace(regex, '<mark class="placeable" title="' + title + '">$&</mark>');
+      }
+
+      string = this.doNotRender(string);
+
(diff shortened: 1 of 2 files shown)


pontoon/base/static/js/main.js
```diff
@@ -79,7 +79,23 @@ var Pontoon = (function (my) {
...
-      return $('<div/>').text(string).html()
+      return $('<div/>').text(string).html();
+    },
+
+    /*
+     * Markup placeables
+     */
+    markPlaceables: function (string) {
+      function markup(string, regex, title) {
+        return string.replace(regex, '<mark class="placeable" title="' + title + '">$&</mark>');
```
jotes: I think we could reuse "mark.js" here.

pontoon/base/static/js/main.js
```diff
@@ -156,6 +172,9 @@ var Pontoon = (function (my) {
             '</header>' +
             '<p class="original">' + self.doNotRender(data.original || '') + '</p>' +
             '<p class="translation" dir="auto" lang="' + self.locale.code + '">' +
+              self.markPlaceables(self.doNotRender(data.translation)) +
```
jotes: I'd move doNotRender as the last step of transformation, I mean:`self.doNotRender(self.markPlaceables(data.translation))`

jotes: Or even better -> use self.doNotRender inside MarkPlaceables to avoid code duplication.

PR #528, opened 2017-01-16
Small template fixes for DJANGO_LOGIN

mastizada:
- Added check for `DJANGO_LOGIN` in translate template for standalone sidebar signin.
- Username label name fixed for standalone login template.

### DIFF
--- a/pontoon/base/static/js/main.js
+++ b/pontoon/base/static/js/main.js
@@ -479,10 +479,11 @@ $(function() {
   }
 
   // Sign in button action
-  $('#fxa-sign-in, #sidebar-signin').on('click', function(ev) {
+  $('#fxa-sign-in, #standalone-signin a, #sidebar-signin').on('click', function(ev) {
     var $this = $(this);
-
-    $this.prop('href', $this.prop('href') + '&next=' + getRedirectUrl());
+    var loginUrl = $this.prop('href'),
+        startSign = loginUrl.match(/\?/) ? '&': '?';
+    $this.prop('href', loginUrl + startSign + 'next=' + getRedirectUrl());
   });
 
   // Sign out button action
--- a/pontoon/base/static/js/main.js
+++ b/pontoon/base/static/js/main.js
@@ -105,8 +105,11 @@ var Pontoon = (function (my) {
      */
     diff: function (reference, string) {
       var self = this,
-          diff = diff_main(reference, string),
+          diff_obj = new diff_match_patch(),
+          diff = diff_obj.diff_main(reference, string),
           output = '';
+      diff_obj.diff_cleanupSemantic(diff);
+      diff_obj.diff_cleanupEfficiency(diff);
 
       $.each(diff, function() {
         var type = this[0],
--- a/pontoon/base/templates/django/base.html
+++ b/pontoon/base/templates/django/base.html
@@ -44,7 +44,7 @@
         <li><a href="{% url 'pontoon.contributors' %}">Contributors</a></li>
         <li><a href="{% url 'pontoon.search' %}">Terminology</a></li>
         <li><a href="{% url 'pontoon.terms' %}">Terms</a></li>
-        <li><a href="https://developer.mozilla.org/en-US/docs/Localizing_with_Pontoon" target="_blank">Help</a></li>
+        <li><a href="https://developer.mozilla.org/docs/Mozilla/Localization/Localizing_with_Pontoon" target="_blank">Help</a></li>
         <li id="admin"{% if not perms.base.can_manage %} class="hidden"{% endif %}><a href="{% url "pontoon.admin" %}">Admin</a></li>
         {% if settings.DJANGO_LOGIN %}
             {% if not user.is_authenticated %}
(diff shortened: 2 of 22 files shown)


pontoon/base/templates/registration/login.html
```diff
@@ -14,7 +14,7 @@ <h2 id="subtitle">Sign in</h2>
 
 <form id="standalone-sign-in" method="POST" action="{{ request.path }}">
     <div class="clearfix">
-        <label for="username">Email</label> <br/>
```
jotes: Good point.

pontoon/base/templates/translate.html
```diff
@@ -326,7 +326,13 @@ <h3 class="loading"><div class="fa fa-refresh fa-spin"></div></h3>
           <button id="clear" title="Clear Translation (Ctrl + Shift + Backspace)">Clear</button>
           <button id="save" title="Save Translation (Enter)"></button>
         {% else %}
-          <p id="sign-in-required"><a id="sidebar-signin" href="{{ provider_login_url(request) }}">Sign in</a> to translate.</p>
+          <p id="sign-in-required">
```
jotes: I started to wonder if we can make it more reusable via a macro or something similar. Anyway that looks good for me. Thx!

mastizada: Its not working currently, as main.js adds next parameter to `#sidebar-signin`. Already asked @mathjazz in irc about it. Possible solutions:
1. Adding custom id for standalone url (will require adding css rule (only color)
2. Adding support for next parameter in standalone login

jotes: ad2. Could you look [here](https://github.com/mozilla/pontoon/blob/master/pontoon/base/static/js/main.js#L485)? There's only a selector missing (or you should change your html these selectors).

mastizada: @jotes standalone login url do not accept next parameter, that causes error. I was asking, should I add support for `next` parameter in `standalone_login`? If not, I will change id for link and it will work same as in navigation.

jotes: @mastizada what kind of error?

mastizada: @jotes main.js add this: `$this.prop('href', $this.prop('href') + '&next=' + getRedirectUrl());`
When using alla

jotes:
@mastizada I think We can call @mathjazz If you think that's ready for final review.

mastizada:
Its ready. Tested on translate.siteledger.com, works as expected. Any suggestions?

pontoon/base/static/js/main.js
```diff
@@ -479,10 +479,11 @@ $(function() {
   }
 
   // Sign in button action
-  $('#fxa-sign-in, #sidebar-signin').on('click', function(ev) {
+  $('#fxa-sign-in, a#standalone-sign-in, #sidebar-signin').on('click', function(ev) {
     var $this = $(this);
-
-    $this.prop('href', $this.prop('href') + '&next=' + getRedirectUrl());
+    var loginUrl=$this.prop('href'),
+        startSign=loginUrl.match(/\?/) ? '&': '?';
```
mathjazz: Nit: Please add space around `=` in both of the lines above.

pontoon/base/templates/django/base.html
```diff
@@ -48,7 +48,7 @@
         <li id="admin"{% if not perms.base.can_manage %} class="hidden"{% endif %}><a href="{% url "pontoon.admin" %}">Admin</a></li>
         {% if settings.DJANGO_LOGIN %}
             {% if not user.is_authenticated %}
-                <li id="standalone-sign-in"><a href="{% url 'standalone_login' %}">Sign in</a></li>
+                <li id="standalone-sign-in"><a id="standalone-sign-in" href="{% url 'standalone_login' %}">Sign in</a></li>
```
mathjazz: Now two elements have the same ID. You can probably remove the one you added.

pontoon/base/templates/translate.html
```diff
@@ -326,7 +326,13 @@ <h3 class="loading"><div class="fa fa-refresh fa-spin"></div></h3>
...
+            {% if settings.DJANGO_LOGIN %}
+              <a id="sidebar-signin" href="{{ url('standalone_login') }}">Sign in</a> to translate.
+            {% else %}
+              <a id="sidebar-signin" href="{{ provider_login_url(request) }}">Sign in</a> to translate.
+            {% endif %}
```
mathjazz: To avoid duplicating code, I suggest you do something like this:

```
        {% if settings.DJANGO_LOGIN %}
          {% set login_url = url('standalone_login') %}
        {% else %}
          {% set login_url = provider_login_url(request) %}
        {% endif %}
        <a id="sidebar-signin" href="{{ login_url }}">Sign in</a> to translate.
```

mastizada:
@mathjazz added unique id for all links, and changed selector in js file. Also changed to set login_url as you suggested.

mathjazz:
@jotes Do you have any objections to `registration/login.html` extending `base.html` instead of `django/base.html`?

We just discovered this with @mastizada while learning that `django/base.html` loads `main.js` but doesn't load `jQuery`.

mastizada:
@mathjazz so, `django.contrib.auth.views` is used for standalone login and logout. As @jotes wrote, `base.html` is written in jinja, but django doesn't support that. That is why `django/base.html` is used. Also changed id for li tags due to form-id (it is used in styles), also label ids in `login.html` (checked using w3 validator).

PR #530, opened 2017-01-17
Fix bug 1331710: Only privileged can flip suggestions

mathjazz:
If the user doesn't have translator permission for the current project, she should not be able to turn Make suggestions toggle on/off from the translate view.

@jotes r?

### DIFF
--- a/pontoon/base/templates/translate.html
+++ b/pontoon/base/templates/translate.html
@@ -287,7 +287,7 @@
             <div class="menu">
               <ul>
                 {{ Checkbox.checkbox('Quality Checks', class='quality-checks', attribute='quality_checks', is_enabled=user.profile.quality_checks) }}
-                {% if user.translated_locales %}
+                {% if user.can_translate(project=project, locale=locale) %}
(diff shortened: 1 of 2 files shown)


BychekRU:
I've already pointed out on this bug. If I register a new user, his rights will be shown as:

But he won't has rights for quality checks:)

mathjazz:
> But he won't has rights for quality checks:)

Hey @BychekRU! Why not? Any registered user can flip quality checks on and off.

mathjazz:
Added another commit to fix permissions: Managers should have permissions to all locale projects, regardless of custom project permissions.

pontoon/base/templates/translate.html
```diff
@@ -287,7 +287,7 @@ <h3 class="loading"><div class="fa fa-refresh fa-spin"></div></h3>
             <div class="menu">
               <ul>
                 {{ Checkbox.checkbox('Quality Checks', class='quality-checks', attribute='quality_checks', is_enabled=user.profile.quality_checks) }}
-                {% if user.translated_locales %}
```
jotes: Ok, my regexes didn't catch this one. Thx!

PR #531, opened 2017-01-17
Permissions frontent improvements

mathjazz:
4 smaller improvements to the permissions page. Details in commit messages.

@jotes r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -569,19 +569,6 @@ class Locale(AggregatedStats):
             'groups__pk'
         )
 
-        projects_contributors = create_users_map(
-            User.objects
-                .filter(
-                    translation__locale=self,
-                    translation__entity__resource__project__pk__in=[project_locale['project__pk'] for project_locale in project_locales]
-                )
-                .exclude(email='')
-                .prefetch_related('translation__entity__resource__project')
-                .values('id', 'first_name', 'email', 'translation__entity__resource__project__pk')
-                .distinct(),
-            'translation__entity__resource__project__pk'
-        )
-
         projects_all_users = defaultdict(set)
 
         for project_locale in project_locales:
@@ -601,7 +588,6 @@ class Locale(AggregatedStats):
                 project_locale['project__name'],
                 projects_all_users[group_pk],
                 projects_translators[group_pk],
-                projects_contributors[project_locale['project__pk']],
                 project_locale['has_custom_translators']
             ))
 
--- a/pontoon/base/static/js/manage_permissions.js
+++ b/pontoon/base/static/js/manage_permissions.js
@@ -91,6 +91,11 @@ $(function() {
     $('#project-selector .menu li').removeClass('hover');
   });
 
+  // Focus project selector search field
+  $('#project-selector .selector').click(function() {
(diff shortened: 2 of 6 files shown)


pontoon/base/models.py
```diff
@@ -601,7 +588,6 @@ def create_users_map(qs, keyfield):
                 project_locale['project__name'],
                 projects_all_users[group_pk],
                 projects_translators[group_pk],
-                projects_contributors[project_locale['project__pk']],
```
mathjazz: Not needed. We always want to use team contributors.

pontoon/base/static/js/manage_permissions.js
```diff
@@ -102,7 +107,7 @@ $(function() {
 
     // Update menu (must be above Copying Translators)
     $(this).addClass('hidden');
-    if ($('#project-selector .menu li:visible').length === 0) {
+    if ($('#project-selector .menu li:not(".hidden")').length === 0) {
```
mathjazz: The `:visible` selector wasn't reliable - if you searched for a project, items could get hidden, so if only one project matched the criteria and you added it, the selector disappeared.

pontoon/base/templates/widgets/multiple_choice.html
```diff
@@ -8,7 +8,7 @@
     </div>
     <ul>
       {% for user in choices %}
-      <li class="clearfix{% if user in extra %} contributor{% endif %}" data-id="{{ user.id }}" title="{{ user.first_name }}"><span class="arrow fa"></span>{{ user.email }}</li>
+      <li class="clearfix{% if user.email in extra %} contributor{% endif %}" data-id="{{ user.id }}" title="{{ user.first_name }}"><span class="arrow fa"></span>{{ user.email }}</li>
```
mathjazz: That was actually a bug. For project permissions, we compared a user dict to user objects, so we never had any matches.

pontoon/base/views.py
```diff
@@ -123,7 +123,7 @@ def locale_permissions(request, locale):
     translators = l.translators_group.user_set.exclude(pk__in=managers).all()
     all_users = User.objects.exclude(pk__in=managers).exclude(pk__in=translators).exclude(email='')
 
-    contributors = User.translators.filter(translation__locale=l).distinct()
+    contributors = User.translators.filter(translation__locale=l).values_list('email', flat=True).distinct()
```
mathjazz: Same as above: That was actually a bug. For project permissions, we compared a user dict to user objects, so we never had any matches.

pontoon/base/models.py
jotes: totally makes sense - and that decreases queries count!

pontoon/base/static/js/manage_permissions.js
jotes: I think we're near this point when some kind if UI testing viaselenium or something similar would be helpful.

pontoon/base/templates/widgets/multiple_choice.html
jotes: There's a huge possibility that I broke this during doing my performance refactorings :(

PR #532, opened 2017-01-18
Bug 1329003: improve diff experience

BychekRU:
@mathjazz Don't you think that comparing previous translations with next will look more logically?
Before:

After:

### DIFF
--- a/pontoon/base/static/css/style.css
+++ b/pontoon/base/static/css/style.css
@@ -1184,7 +1184,7 @@ body > .container {
   text-align: left;
 }
 
-#helpers > section ul li > p.translation-raw {
+#helpers > section ul li > p.translation {
   display: none;
 }
 
@@ -1195,12 +1195,12 @@ body > .container {
 
 #helpers > section ul li > p ins {
   background: #4B6259;
-  color: #7BC876;
+  color: #9CD699;
 }
 
 #helpers > section ul li > p del {
   background: #674B54;
-  color: #FE7777;
+  color: #FE8F8F;
 }
 
 #helpers > section ul li > p ins mark,
--- a/pontoon/base/static/css/translate.css
+++ b/pontoon/base/static/css/translate.css
@@ -1708,6 +1708,14 @@ p#sign-in-required > a#sidebar-signin {
   float: left;
 }
 
+#helpers > section ul li > header .toggle-diff {
+  margin-right: 6px;
+}
+
+#helpers > section ul li > header .toggle-diff.active {
+  color: #7BC876;
+}
+
 #helpers > section ul li > header .toolbar {
   float: right;
 }
--- a/pontoon/base/static/js/lib/diff.js
+++ b/pontoon/base/static/js/lib/diff.js
@@ -1,10 +1,4 @@
 /**
- * This library modifies the diff-patch-match library by Neil Fraser
- * by removing the patch and match functionality and certain advanced
- * options in the diff function. The original license is as follows:
- *
- * ===
- *
  * Diff Match and Patch
  *
  * Copyright 2006 Google Inc.
@@ -23,6 +17,46 @@
  * limitations under the License.
  */
 
+/**
+ * @fileoverview Computes the difference between two texts to create a patch.
+ * Applies the patch onto another text, allowing for errors.
+ * @author fraser@google.com (Neil Fraser)
+ */
+
+/**
+ * Class containing the diff, match and patch methods.
+ * @constructor
+ */
+function diff_match_patch() {
+
+  // Defaults.
+  // Redefine these in your program to override the defaults.
+
+  // Number of seconds to map a diff before giving up (0 for infinity).
+  this.Diff_Timeout = 1.0;
+  // Cost of an empty edit operation in terms of edit characters.
+  this.Diff_EditCost = 4;
+  // At what point is no match declared (0.0 = perfection, 1.0 = very loose).
+  this.Match_Threshold = 0.5;
+  // How far to search for a match (0 = exact location, 1000+ = broad match).
+  // A match this many characters away from the expected location will add
+  // 1.0 to the score (0.0 is a perfect match).
+  this.Match_Distance = 1000;
+  // When deleting a large block of text (over ~64 characters), how close do
+  // the contents have to be to match the expected contents. (0.0 = perfection,
+  // 1.0 = very loose).  Note that Match_Threshold controls how closely the
+  // end points of a delete need to match.
+  this.Patch_DeleteThreshold = 0.5;
+  // Chunk size for context length.
+  this.Patch_Margin = 4;
+
+  // The number of bits in an int.
+  this.Match_MaxBits = 32;
+}
+
+
+//  DIFF FUNCTIONS
+
 
 /**
  * The data structure representing a diff is an array of tuples:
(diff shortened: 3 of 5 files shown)


mathjazz:
Thanks for the patch! I see your point and since you're not the first one with this idea, I'd like to open a discussion. ;)

Some context - we'd like to figure out how to implement suggestion diffs. There are (at least) two possible approaches.

1. Make suggestion diffs against active translation.
* It's the first translation listed in the History tab, also displayed in the sidebar and textarea.
* It's usually the approved translation and if no such translation exist, it's the newest suggestion.
* This is the approach currently deployed to production.
* See the "Before" screenshot above.
* Only @mathjazz supports this approach. :) Maybe also @tomer, who filed bug 1329003?

2. Make suggestion diff against the first translation submitted.
* It's the last translation listed in the History tab, unless it's approved - then it's first.
* It's always the oldest suggestion.
* This is the approach proposed in this PR.
* See the "After" screenshot above.
* @jotes @TheoChevalier @BychekRU are in favour of this approach.

Also adding @flodolo @MikkCZ @mastizada @StoyanDimitrov @stasm @zbraniecki @Pike @gueroJeff.

I'm also thinking that we should make the diff display optional, because it's sometimes more useful to see the actual translation (especially if the change is relativelly big). Maybe display diffs by default and the actual translation on hover (or vice versa). Or have a toggle in settings (I'd like to avoid cluttering the UI with non-essential features, so a button for each suggestion would be an overkill).

Thoughts?

tomer:
I'd also suggest that when a translation is completely different than the other translation, there is no point in providing differences. It can be done by counting the number of changed characters or words, and if it is above a defined threshold - a diff should not be provided. 

default

flodolo:
As Tomer wrote, option 1 makes the suggestion almost unreadable if there are a lot of differences compared to the current translation.

I'm not sure I understand option 2: if there's a different translation stored, why would it be useful to diff against the oldest one, which is already irrelevant?

Personally I would like option 1 (diff against the current stored translation, none if no translation is submitted yet), with a switch to show the string or the diff. 

I don't think it would clutter the UI too much, and I would prefer not to hide that in Settings: I normally touch them once per session, e.g. to enable suggestions, I would like to enable/disable the diff string by string if I can't spot immediately what is changing.

BTW, 👍 for the diff, I've always wanted that.

stasm:
Comparing against the oldest, original translation sounds like an approach which won't scale well over time.  For this reason I'd support a change to always diff against the most recent approved translation.

I'm also not sure if I understand the screenshots well: is the diff displayed inside of the older translation? ( the diff is between a _from_ and a _to_ revision, by _older_ I mean _from_.) In the _Before_ screenshot in https://github.com/mozilla/pontoon/pull/532#issue-201501440, the green text seems to denote obsolete parts of the translation which were _removed_ in the new iteration.  If that's the case, I'd advise changing this too. Could we show the diff inside of the _to_ revision? It should probably be optional so that it doesn't clutter the most recent revision by default.

tomer:
> I don't think it would clutter the UI too much, and I would prefer not to hide that in Settings

What if diff view would be visible only on mouse hover or something alike?

tomer:
I think that both options could be useful, and I'll add a third option - diff against the source string, which would be useful for non en-us English locales, as well as everyone else when source string changes without key change, and when translating long strings with some untranslatable keywords. 

I'd suggest that the base point for the diff would be left for the user to choose, like how it is dealt on Wikipedia and the diff on each locale sign-off page.

flodolo:
> What if diff view would be visible only on mouse hover or something alike?

I would prefer something that doesn't change without an explicit action on my part.

StoyanDimitrov:
This is a must-have! The readability will improve greatly if diff-by-character is replaced with diff-by-word, given the language is not German (longer words, jk). I've been there (not Germany).

flodolo:
Example of how diff can be detrimental, I basically can't read the original sentence

schermata 2017-01-18 alle 13 48 08

Pike:
Yeah, there are two problems in that screenshot, one is that the diff is too fragmented. diff-match-patch has options there, IIRC. (Options I don't use on elmo yet, either)

And the styling can be made less harsh.

Also I slowly start to understand what the patch actually did. On elmo, we show both strings with highlights on what's new and what's changed. It might be an option to not show the removed text at all.

MikkCZ:
I agree with stasm, that the diff should be against the last approved revision. Or maybe against what is currently was loaded in the input (but not changing dynamically when writing there).

Tomer has also point, that when the diff is too big (Levenshtein distance lev(rev1, rev2) > X ?) the diff highlighting makes it only worse to read.

MikkCZ:
@Pike or maybe just highlight somehow the place where some change happened? I like Pontoon kind of minimal UI, but the colors break it a bit and a lot of attention to the history. I have asked my friend, what does she think this is, and she thought that the colors mean some feedback from the translation manager (green what is 👍 and red what is wrong).

mastizada:
I like current implementation: showing changes on old translation. But using diff on words instead of letters will be a lot better as it is not readable.

pontoon-diff

mathjazz:
All, thank you very much for the extensive feedback!

There are two groups of issues pointed out in your comments:

**1. What should we diff against?** There's a pretty strong consensus we should keep making diffs against the currently active translations. No action needed here.

**2. Readability.** It sucks, especially if diff is relatively big. Action items:
* Make use of [diff-match-patch](https://neil.fraser.name/software/diff_match_patch/svn/trunk/demos/demo_diff.html) options to increase human readability. 
* Investigate options for more gentle styling.
* Add ability to turn diffs on and off for each suggestion.

I'll get back to you as soon as I address the action items.

stasm:
Am I the only one who finds it weird that the diff is displayed in the _from_ revision and that the green text actually means "this was removed in the subsequent revision"?

mathjazz:
> Am I the only one who finds it weird that the diff is displayed in the from revision and that the green text actually means "this was removed in the subsequent revision"?

@stasm The diff is displayed within each revision. Obviously that means there's no diff to be displayed for the base revision, which is the active translation - listed on top of the History tab.

For each revision, we run `diff(base revision, this revision)`. So the green text in each revision represented the text that needs to be added to the base revision in order to get this revision.

Does that make more sense? If not, what would?

(BTW, we should probably keep the chronological order in the History tab instead of moving the active revision on top.)

mathjazz:
So @BychekRU and I have made 4 changes:
* The diffs are now less fragmented.
* Styling of changed text uses some more contrast.
* There's a Show/Hide diff togle located above each suggestion. Sadly there's no familiar diff icon we could use.
* Not sure about this one, needs more real-life testing: diffs are hidden by default.

You can test the changed on stage:
https://mozilla-pontoon-staging.herokuapp.com/sl/appstores/whatsnew/whatsnew_android_45.lang/?string=149465

Feel free to add or remove translations in any of the projects and locales on the staging server - it doesn't push to repositories.

Thoughts?

mastizada:
Is it possible to show complete words when there are more than 2-3 changes in the same word? It will improve readability but if it is too complex then no problem.

PS. Show diff icon: can be placed near to approve, delete icons with similar to git icon, which will mean versions.
pontoon-diff
diff

BychekRU:
Output with using diff_cleanupSemantic()

Output with diff_cleanupEfficiency()

Now I use this two functions together:

stasm:
> For each revision, we run diff(base_revision, this_revision). So the green text in each revision represents the text that needs to be added to the base_revision in order to get to this_revision.

> Does that make more sense? If not, what would?

@mathjazz Commenting here a bit late, sorry. Thanks for the explanation.  I was a bit confused at first by the fact that the green color was used to show text removed in the active translation.  I'm used to diffs showing that in red.

However, after browsing Pontoon for a while I see that it makes sense for the use-case.  For instance https://pontoon.mozilla.org/fr/firefox-for-ios/firefox-ios.xliff/?string=66899 is very clear to me, so everything seems fine!

PR #535, opened 2017-01-24
[WIP] Fix bug 1243140. Added download of terminology files.

jotes:
Questions:
* [x] handle terms with several possible nouns/a few different translations
* [x] handle missing translations for terms in a locale
* [x] [discuss] Microsoft terminology contains a lot of terms strongly connected to their products and those aren't very useful for projects at Mozilla.
* [x] [discuss] if this really should be an separate model instead of our current model/translation entity (will be easier to add them to l10n flow)

Todo:
* [x] - Add tests for import of .tbx files.
* [x] - Cleanup import of terms.
* [x] - optimize term fetching queries in get-entities
* [x]  - add config='english' to the DBTermIndex and create an expression index for faster lookup queries.
* [x] - remove all debug prints
* [x] - add tests for the views.


mathjazz:
1. Good point. This should be taken care of by the tooltip or whatever we end up using on frontend instead of the title attribute.

2. If translation is not available, we only show part of speech and description in the tooltip ATM. That's probably sufficient for now.

3. Correct. The plan is to only use MS Terminology temporarily. We already started the effort of building our own terminology which will replace it at some point in the future.

4. I think this should be a separate model indeed. It required fields that aren't part of the Translation model and vice versa. We also can't match terminology records against existing entities, resources, projects. What would be the benefit of using the same model as we use for translations?

jotes:
ad4. I think you have a point, My confusion was caused by the fact that some of terms from Microsoft terminology files don't have translations and I just wanted to confirm If I should worry about this problem right now.

pontoon/base/migrations/0090_auto_20170413_2101.py
```diff
@@ -0,0 +1,220 @@
+# -*- coding: utf-8 -*-
```
jotes: @mathjazz could you look at this migration?

pontoon/base/migrations/0090_auto_20170413_2101.py
```diff
@@ -0,0 +1,220 @@
...
+    u'quc-Latn-GT': u'',  # MISSING?
+    u'quz-pe': u'',  # MISSING?
+    u'rw-rw': u'',  # MISSING?
+    u'sd-Arab-PK': u'',  # MISSING?
+    u'sr-Cyrl-ba': u'sr',  # CONFLICT?
+    u'sr-Cyrl-rs': u'sr',  # CONFLICT?
+    u'sr-Latn-rs': u'sr',  # CONFLICT?
+    u'sw-ke': u'sw',  # CONFLICT?
+    u'ta-in': u'ta',  # CONFLICT?
+    u'tg-Cyrl-tj': u'',  # MISSING?
```
mathjazz: tg

pontoon/base/migrations/0090_auto_20170413_2101.py
```diff
@@ -0,0 +1,220 @@
...
+    u'th': u'th',
+    u'tlh': u'',  # MISSING?
+    u'tlh-Qaak': u'',  # MISSING?
+    u'tr': u'tr',
+    u'uk': u'uk',
+    u'ur': u'ur',
+    u'vi': u'vi',
+    u'yua': u'',  # MISSING?
+    u'zh-CHS': u'zh-CN',  # CONFLICT?
+    u'zh-CHT': u'zh-CN'  # CONFLICT
```
mathjazz: We use Simplified Chinese script:
https://pontoon.mozilla.org/zh-CN/

So use `zh-CHS`.

pontoon/base/migrations/0090_auto_20170413_2101.py
```diff
@@ -0,0 +1,220 @@
...
+    u'no': u'',  # MISSING?
+    u'otq': u'', # MISSING?
+    u'pl': u'pl',
+    u'pt': u'pt-PT',
+    u'ro': u'ro',
+    u'ru': u'ru',
+    u'sk': u'sk',
+    u'sl': u'sl',
+    u'sr-Cyrl': u'sr',  # CONFLICT?
+    u'sr-Latn': u'sr',  # CONFLICT?
```
mathjazz: We use Cyrillic script:
https://pontoon.mozilla.org/sr/

So use `sr-Cyrl`.

pontoon/base/migrations/0090_auto_20170413_2101.py
```diff
@@ -0,0 +1,220 @@
...
+# Generated by Django 1.10.5 on 2017-04-13 21:01
+from __future__ import unicode_literals
+
+from bulk_update.helper import bulk_update
+from django.db import migrations, models
+
+TRANSLATOR_LOCALES = {
+    u'ar': u'ar',
+    u'bg': u'bg',
+    u'bs-Latn': u'bs',  # CONFLICT?
```
mathjazz: What does that conflict with?

pontoon/base/migrations/0090_auto_20170413_2101.py
```diff
@@ -0,0 +1,220 @@
...
+    u'vi-vn': u'vi',
+    u'wo-sn': u'wo',
+    u'yo-ng': u'yo',
+
+    # Locales that couldn't be directly mapped to the pontoon.
+    u'az-Latn-az': u'az',
+    u'bn-bd': u'bn-BD',
+    u'bn-in': u'bn-IN',
+    u'bs-Cyrl-ba': u'bs',  #CONFLICT?
+    u'bs-Latn-ba': u'bs',  #CONFLICT?
```
mathjazz: We use Latin script:
https://pontoon.mozilla.org/bs/

So use `bs-Latn-ba`.

pontoon/base/migrations/0090_auto_20170413_2101.py
```diff
@@ -0,0 +1,220 @@
...
+    u'ps-af': u'',  # MISSING?
+    u'pt-br': u'pt-BR',
+    u'pt-pt': u'pt-PT',
+    u'quc-Latn-GT': u'',  # MISSING?
+    u'quz-pe': u'',  # MISSING?
+    u'rw-rw': u'',  # MISSING?
+    u'sd-Arab-PK': u'',  # MISSING?
+    u'sr-Cyrl-ba': u'sr',  # CONFLICT?
+    u'sr-Cyrl-rs': u'sr',  # CONFLICT?
+    u'sr-Latn-rs': u'sr',  # CONFLICT?
```
mathjazz: We use Cyrillic script:
https://pontoon.mozilla.org/sr/

So use `sr-Cyrl-rs`.

pontoon/base/migrations/0090_auto_20170413_2101.py
```diff
@@ -0,0 +1,220 @@
...
+    u'lb-lu': u'',  # MISSING?
+    u'mi-nz': u'',  # MISSING?
+    u'mk-mk': u'mk',
+    u'mn-mn': u'mn',
+    u'ms-bn': u'ms', # CONFLICT?
+    u'ms-my': u'ms', # CONFLICT?
+    u'mt-mt': u'', # MISSING?
+    u'nb-no': u'nb-NO',
+    u'nn-no': u'nn-NO',
+    u'nso-za': u'nso', # CONFLICT?
```
mathjazz: What does that conflict with?

pontoon/base/migrations/0090_auto_20170413_2101.py
```diff
@@ -0,0 +1,220 @@
...
+    u'or-in': u'or',  # CONFLICT?
```
mathjazz: What does that conflict with?

pontoon/base/migrations/0090_auto_20170413_2101.py
```diff
@@ -0,0 +1,220 @@
...
+    u'ti-et': u'',  # MISSING?
+    u'tk-tm': u'',  # MISSING?
+    u'tl-PH': u'tl',
+    u'tn-za': u'',  # MISSING?
+    u'tt-ru': u'',  # MISSING?
+    u'ug-cn': u'',  # MISSING?
+    u'uz-Latn-uz': u'uz',
+    u'xh-za': u'xh',
+    u'zh-cn': u'zh-CN',
+    u'zh-hk': u'',  # CONFLICT?
```
mathjazz: What does that conflict with? Use `zh-HK`.

pontoon/base/migrations/0090_auto_20170413_2101.py
```diff
@@ -0,0 +1,220 @@
...
+    u'zh-tw': u'zh-TW',
+    u'zu-za': u'zu'  # CONFLICT?
```
mathjazz: What does that conflict with?

pontoon/base/migrations/0090_auto_20170413_2101.py
```diff
@@ -0,0 +1,220 @@
...
+    u'pt-br': u'pt-BR',
+    u'pt-pt': u'pt-PT',
+    u'quc-Latn-GT': u'',  # MISSING?
+    u'quz-pe': u'',  # MISSING?
+    u'rw-rw': u'',  # MISSING?
+    u'sd-Arab-PK': u'',  # MISSING?
+    u'sr-Cyrl-ba': u'sr',  # CONFLICT?
+    u'sr-Cyrl-rs': u'sr',  # CONFLICT?
+    u'sr-Latn-rs': u'sr',  # CONFLICT?
+    u'sw-ke': u'sw',  # CONFLICT?
```
mathjazz: What does that conflict with? Use `sw`.

pontoon/base/migrations/0090_auto_20170413_2101.py
```diff
@@ -0,0 +1,220 @@
...
+    u'ta-in': u'ta',  # CONFLICT?
```
mathjazz: What does that conflict with? Use `ta`.

pontoon/base/migrations/0090_auto_20170413_2101.py
```diff
@@ -0,0 +1,220 @@
...
+    u'iu-Latn-ca': u'',  # MISSING?
+    u'kok-in': u'',  # MISSING?
+    u'ku-Arab-IQ': u'ku',
+    u'ky-kg': u'',  # MISSING?
+    u'lb-lu': u'',  # MISSING?
+    u'mi-nz': u'',  # MISSING?
+    u'mk-mk': u'mk',
+    u'mn-mn': u'mn',
+    u'ms-bn': u'ms', # CONFLICT?
+    u'ms-my': u'ms', # CONFLICT?
```
mathjazz: Take `ms-my`.

jotes:
​@mathjazz
AWESOME! Thank you so much.​ Some of conflicts were probably my
mis-vimisms/overlooks.


-- 
Kind regards

​Jarek "jotes" Śmiejczak,
​P​
Homepage: https://evilb.it, Github: http://github.com/jotes

pontoon/base/migrations/0090_auto_20170413_2101.py
jotes: I probably wasn't sure if nso is equal to nso-za. Thanks for clarification.

jotes:
@mathjazz 
Hi,
So, there's probably a lot of things to clean up (missing docstrings, tests etc.) but I think it's usable enough to add some ui.
Usage:
```
./manage.py sync_projects
./manage.py download_microsoft_terminology --locale sl
./manage.py import_tbx sl.tbx
```
Hopefully, if none of those commands fail, you shoudl be able to see array of terms in a list of `entities` from `/get-entities/`.
screenshot

mathjazz:
Great job! Works very well!

Sorry for the bad push! The plans was to only add this:
https://github.com/mozilla/pontoon/pull/535/commits/df9606c090557e1cc1772d818310d64eae77b28d

mathjazz:
Thanks. Trying to be more careful this time:
https://github.com/jotes/pontoon/pull/1

Could you also rebase against master after you take the PR?

mathjazz:
I only found one issue.

Say you have the following terms:
* `home`
* `page`
* `home page`

And your string is: `This is my home page`.

We match `home` and `page` in the string, but we should actually match `home page`.

jotes:
I squashed commits and fixed issue with matching terms to entities.

jotes:
Backend should be able to match terms with multiple meanings (e.g. 'stream') and return them to frontend.

mathjazz:
@jotes I have two proposals:

1. Can we join `download_microsoft_terminology` and `import_tbx` into a single management command. e.g. `import_microsoft_terminology`? I think this is necessary for Heroku, because by the time the second command is ran, the file downloaded by the first one is no longer available. If it's not, we should keep them separated.

2. I think `terms` should be grouped by `source_term` (`term` in the screenshot below). In the example below, we should only have two items in the `terms` array - `extension` and `about`. And then each of those items should be an array - `extension` of 2 items and `about` of 1.
screen shot 2017-05-25 at 20 59 30

mathjazz:
Two more things: 

1. It seems like the original (en-US) term is included in the `translations` Array. No need for that.

2. We can file a bug for this and implement it after the current patch lands: we should stem/lemmatize source strings and then match words in their default form against terms. That means the term `extension` should match the string `Browse all extensions`. See: https://stackoverflow.com/questions/771918/how-do-i-do-word-stemming-or-lemmatization.

jotes:
@mathjazz I rewrote the algorithm to use porter2 stemmer and I got following results:
```python

In [8]: e.string
Out[8]: u'Browse all extensions'

In [9]: e.terminology_terms.all().values_list('source_term', 'description')
Out[9]: <QuerySet [(u'browse', u'To scan a database, a list of files, or the Internet, either for a particular item or for anything that seems to be of interest.'), (u'extension', u'A set of characters added to the end of a file name that identifies the file type or format.'), (u'extension', u'A program or program module that adds functionality to or extends the effectiveness of a program.'), (u'extensibility', u"The capacity to extend or stretch the functionality of the development environment - to add something to it that didn't exist there before."), (u'extension', u'An installable unit that contributes new capabilities into Visual Studio Online and results in increased value for users.'), (u'extension', u'A grouping of properties of an entity in the Office Graph, such as a document or an email.')]>
```
Is that good enough? I think we could improve that in the next bug.

mathjazz:
Looks good, except for `extensibility` - that shouldn't be included.

Not sure how this ended up on the list. Maybe you also stem terminology? That's not necessary.

jotes:
@mathjazz ok, I decided to change porter2 into simple singularization:

```
In [2]: e.terminology_terms.all().values_list('source_term', 'description')
Out[2]: <QuerySet [(u'extension', u'A program or program module that adds functionality to or extends the effectiveness of a program.'), (u'browse', u'To scan a database, a list of files, or the Internet, either for a particular item or for anything that seems to be of interest.'), (u'extension', u'A set of characters added to the end of a file name that identifies the file type or format.'), (u'extension', u'An installable unit that contributes new capabilities into Visual Studio Online and results in increased value for users.'), (u'extension', u'A grouping of properties of an entity in the Office Graph, such as a document or an email.')]>
```

mathjazz:
@jotes Thanks for the update! `import_microsoft_terminology` and singularization work fine.

1. It seems like the original (en-US) term is still included in the translations Array. Could you remove it?

2. How hard would it be to tweak the JSON a bit further: the keys should be the actual terms in the source string and their values should be lists of matching term objects. See example for `Browse all extensions`:

```
{
  "Browse": [
    {
      "description": "To scan a database...",
      "note": "Verb",
      "term": "browse",
      "translation": "brskati",
    }
  ],
  "extensions": [
    {
      "description": "A grouping of properties...",
      "note": "Noun",
      "term": "extension",
      "translation": "razširitev",
    },
    {
      "description": "A set of characters...",
      "note": "Noun",
      "term": "extension",
      "translation": "pripona",
    }
  ]
}
```

mathjazz:
This looks good.

I think you can remove the `import_tbx` management command at this point.

Seems like `assign_terms_to_entities` is pretty slow, I wonder if we can optimize that. It takes 6 minutes for a single locale for me, vs. 6:25 for the entire management command.

jotes:
@mathjazz what should backend do when there's no translation of a term in the current locale?

mathjazz:
> @mathjazz what should backend do when there's no translation of a term in the current locale?

The `translation` attribute should be `null`.

mathjazz:
This PR is getting big. I wonder if it makes sense to split the `TRANSLATOR_LOCALES` and `TERMINOLOGY_LOCALES` and the related code changes into a separate PR. @jotes? If it's more than 15 minutes of work, we shouldn't bother.

We should also prune the MS TBX files on import. @unclenachoduh created a list of tokens to blacklist - if there's a match in the term itself or the term description, we shouldn't import the term.

* Product names: Windows, Xbox, Microsoft, Skype, Bing, ActiveX, BitLocker, OneDrive, OneNote
* `^\s*[A-Z]+\s*[A-Z]*\s*$` (token for all of the acronyms)
* `^[^a-zA-z]` (tokens that start with special characters)
* `.*@.*` (tokens that contain @ because I hate this symbol. Also because emails and social media handles)

@unclenachoduh You mentioned you already have the code that prunes the tbx. Do you mind sharing it? Again, if this turns out to be more than a trivial task, we should leave it for a separate PR, so that we can ship one earlier.

unclenachoduh:
I used the following code to test the blacklist filters on the Microsoft .tbx. Some of the testing features have been removed for readability, so this code may not run as is.

Regardless, the filter is in there, but some adjustments will need to be made before this can be patched.

```
# This file reads a doctored .tbx file that is one line per tbx entry (entry being a term concept)
import re
readfile = open("MS.tbx", encoding="utf8")
ms = readfile.read()
readfile.close()
msterms = ms.split('\n')

# The following lists are written to .txt files at the end of script

# Entire tbx entry of a term with no branding in definition or term
writeme2 = []
# Entire tbx entry of a term with branding in defintion
dntwriteme2 = []

# Only the term in a tbx entry with no branding in definition or term
writeterms2 = []
# Only the term in a tbx entry with branding in definition
dntwriteterms2 = []

count = 0

for line in msterms:
    #print(str(count) +'\t'+ line)
    count += 1
    parts = line.split(">")

    print(parts[9])

    stupid = re.sub("</term", "",  parts[9])
    stupid2 = re.sub("</descrip", "", parts[4])

    if re.match(".*\d.*", stupid) \
            or re.match(".*Windows.*", stupid) \
            or re.match(".*Xbox.*", stupid) \
            or re.match(".*Microsoft.*", stupid) \
            or re.match(".*Skype.*", stupid) \
            or re.match(".*Bing.*", stupid) \
            or re.match(".*ActiveX.*", stupid) \
            or re.match("^\s*[A-Z]+\s*[A-Z]*\s*$", stupid) \
            or re.match(".*BitLocker.*", stupid) \
            or re.match(".*OneDrive.*", stupid) \
            or re.match(".*OneNote.*", stupid) \
            or re.match("^[^a-zA-z].*", stupid) \
            or re.match(".*@.*", stupid):
        dntwriteme2.append(line)
        dntwriteterms2.append(stupid)
    elif re.match(".*\d.*", stupid2) \
            or re.match(".*Windows.*", stupid2) \
            or re.match(".*Xbox.*", stupid2) \
            or re.match(".*Microsoft.*", stupid2) \
            or re.match(".*Skype.*", stupid2) \
            or re.match(".*Bing.*", stupid2) \
            or re.match(".*ActiveX.*", stupid2) \
            or re.match("^\s*[A-Z]+\s*[A-Z]*\s*$", stupid2) \
            or re.match(".*BitLocker.*", stupid2) \
            or re.match(".*OneDrive.*", stupid2) \
            or re.match(".*OneNote.*", stupid2) \
            or re.match(".*@.*", stupid2):
        dntwriteme2.append(line)
        dntwriteterms2.append(stupid + "\t" + stupid2)
    else:
        writeme2.append(line)
        writeterms2.append(stupid)

# -----------------------------
# Below here is where we write to files for checking definitions too
# -----------------------------

writeme2 = sorted(writeme2)
dntwriteme2 = sorted(dntwriteme2)
writeterms2 = sorted(writeterms2)
dntwriteterms2 = sorted(dntwriteterms2)

scribe = ""

for line in writeme2:
    scribe += line + "\n"

scribe = re.sub("\n$", "", scribe)

writefile = open("output3/No_brand2.txt", 'w+', encoding="utf8")
writefile.write(scribe)

writefile.close()

scribe = ""

for line in dntwriteme2:
    scribe += line + "\n"

scribe = re.sub("\n$", "", scribe)

writefile = open("output3/branded2.txt", 'w+', encoding="utf8")
writefile.write(scribe)

writefile.close()

scribe = ""

for line in dntwriteterms2:
    scribe += line + "\n"

scribe = re.sub("\n$", "", scribe)

writefile = open("terms/brandedterms2.txt", 'w+', encoding="utf8")
writefile.write(scribe)

writefile.close()

scribe = ""

for line in writeterms2:
    scribe += line + "\n"

scribe = re.sub("\n$", "", scribe)

writefile = open("terms/notbrandedterms2.txt", 'w+', encoding="utf8")
writefile.write(scribe)

writefile.close()`
```

pontoon/terminology/models.py
```diff
@@ -0,0 +1,250 @@
...
+        log.info('Updated {} translations of terms.'.format(len(update_translations)))
+        self.assign_terms_to_entities(
+            Entity.objects.values_list(
+                'pk',
+                'string',
+                'string_singulars',
+                'string_plural',
+                'string_plural_singulars'
+            )
+        )
```
mathjazz: Everytime we import TBX, `assign_terms_to_entities` is called for all Entities in the DB, removing all existing terms relations and assigning all entities to all terms?

pontoon/terminology/models.py
```diff
@@ -0,0 +1,250 @@
...
+    term_index = DBTermIndex
+
+    def find_terms(self, original_string, entity_singulars):
+        """Return terms ids and associated phrases in string."""
+        terms_phrases = defaultdict(set)
+        original_string_words = get_words(original_string, lower=False)
+        entity_singular_words = entity_singulars.split()
+
+        # Search for terms that could be good candidates to match them with the entity string.
+        candidates = self.term_index.get_terms_candidates(entity_singulars)
```
mathjazz: Could you explain a bit more how do you find term candidates?

pontoon/terminology/management/commands/import_microsoft_terminology.py
```diff
@@ -0,0 +1,185 @@
...
+STOPWORDS = map(re.compile, STOPWORDS)
+
+
+def retrieve_microsoft_terms((locale_code, ms_locale_code)):
+    """
+    Retrieves a terminology file from Microsoft Terminology Portal and parse them.
+    """
+    @retry(stop_max_attempt_number=10)
+    def get_tbx_contents():
+        # Because it's an ASP.NET page we have to extract viewstate and eventvalidation from the page first.
```
mathjazz:
```
# Because it's an ASP.NET page, we have to first extract viewstate and eventvalidation
# from the page.
```

pontoon/terminology/management/commands/import_microsoft_terminology.py
```diff
@@ -0,0 +1,185 @@
...
+    Import terminology files for a set of locales or all of them in the concurrent fashion.
+    """
+    help = 'Import all or selected terminology files from the Microsoft Language Portal.'
+
+
+    def add_arguments(self, parser):
+        # A number of processes that will be spawned to download and parse terms from tbx files.
+        parser.add_argument('--workers', type=int, default=10, help='A number of download workers.')
+
+        # A list of locales that user want to import. Fetches all files if the user doesn't pass any parameter.
```
mathjazz: A list of locales to import. Fetches all files if no parameter is passed.

pontoon/terminology/management/commands/import_microsoft_terminology.py
```diff
@@ -0,0 +1,185 @@
...
+            log.error("Couldn't load data from server. Please, try again later.")
+            sys.exit(-1)
+
+        # Merge all terms imported from different locale files
+        for locale_terms in terms_per_locale:
+            for term in locale_terms:
+                if term.term_id in terms:
+                    terms[term.term_id].translations.update(
+                       term.translations
+                    )
```
mathjazz: Nit: This fits into one line.

pontoon/terminology/management/commands/import_microsoft_terminology.py
```diff
@@ -0,0 +1,185 @@
...
+                else:
+                    terms[term.term_id] = term
+
+        # filter terms with MS specific phrases
```
mathjazz: Filter.

Move belo the `log.info` statements, right above `cleaned_terms` definition.

pontoon/base/models.py
```diff
@@ -1843,6 +1849,19 @@ def get_translation(self, plural_form=None):
                 'pk': None,
             }
 
+    def get_terms(self, locale):
+        """
+        Terminology terms for a given locale, ready to be serialized in JSON.
+        """
+        terms = getattr(self, 'prefetch_terms', self.terms.all())
```
jotes: :+1:

pontoon/base/signals.py
```diff
@@ -8,13 +8,48 @@
...
+    instance = kwargs.get('instance', None)
+
+    if instance.string:
+        instance.string_singulars = ' '.join(get_singulars(instance.string))
+    else:
+        instance.string_singulars = ''
+    if instance.string_plural:
+        instance.string_plural_singulars = ' '.join(get_singulars(instance.string_plural))
+    else:
+        instance.string_plural_singulars = ''
```
jotes: I'll double check what was the reason, because you're probably right -> that's not needed anymore.

pontoon/base/signals.py
```diff
@@ -8,13 +8,48 @@
...
+
+
+@receiver(post_save, sender=Entity)
+def entity_assign_terms(sender, **kwargs):
```
jotes: I guess that would be helpful if someone would ever try to manipulate Entity objects in django console (e.g. for unit tests, manual changes etc.)

pontoon/base/utils.py
```diff
@@ -654,3 +657,41 @@ def build_translation_memory_file(creation_date, locale_code, entries):
...
+    if len(chars) != len(replace_chars):
+        raise ValueError('Chars and replace_chars have to have the same length.')
+    return {
+        ord(c): ord(replace_chars[i])
+        for i, c in enumerate(chars)
+    }
+
+
+whitespace_trans_table = unicode_make_trans(unicode(string.whitespace), u' ' * len(string.whitespace))
+punctuation_trans_table = unicode_make_trans(unicode(string.punctuation), u' ' * len(string.punctuation))
```
jotes: yes, we can :-)

pontoon/base/migrations/0101_add_fields_for_singular_forms.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    Entity = apps.get_model('base', 'entity')
+    entities_to_update = []
+
+    for entity in Entity.objects.all():
+        if entity.string:
+            entity.string_singulars = ' '.join(get_singulars(entity.string))
+
+        if entity.string_plural:
+            entity.string_plural_singulars = ' '.join(get_singulars(entity.string_plural))
```
jotes: added to my to do list.

pontoon/terminology/models.py
jotes: The whole logic is in `term_indexes.py`. Basically there are two classes:
* MemoryIndex - imports all terms from the database into memory and builds the index from words inside the terms and returns terms which words intersect with an entity.
* DBIndex - every call causes sql query on the database. Here I use postgresql vector fields to do queries.

pontoon/terminology/models.py
```diff
@@ -0,0 +1,250 @@
...
+    term_id = models.CharField(max_length=30, unique=True)
+
+    """Source string (in en-US) that contains value to search in strings."""
+    source_term = models.TextField(db_index=True)
+
+    """
+    Stems from source_text, we store them to avoid expensive stemming operations
+    during metching terminology terms with entities.
+    """
+    source_term_singulars = models.TextField(db_index=True)
```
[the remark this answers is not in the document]
jotes: I noticed a few which weren't in neutral form: "background graphics", "Relationships", "Movies", etc. I wasn't sure if that situation is specific to MS Terminology.

adngdb:
Woopsy, sorry I kinda forgot about this. It needs to be rebased though, care to do that @jotes?

jotes:
@adngdb I'll do this today.

jotes:
@adngdb  Hey, there're some pep8 bugs to fix - I'll do this tomorrow. However, feel free to start the review.

pontoon/terminology/models.py
```diff
@@ -0,0 +1,251 @@
...
+from pontoon.base.utils import (
+    get_words,
+    get_singulars
+)
+from pontoon.terminology.term_indexes import DBTermIndex
+
+log = logging.getLogger(__name__)
+
+
+class TermManager(models.Manager):
```
adngdb: I think this is a bad use of the ``Manager`` class. As far as I understand it, the ``Manager`` is made for constructing queries on a database model. Basically, all base methods of the ``models.Manager`` class return a new instance of itself with a different context (with a few exceptions that do not matter, like ``.count()``). So, in my opinion, if your method doesn't return an instance of its class, then it shouldn't be in a Manager. 

However, if I'm wrong regarding this, please correct me!

pontoon/terminology/models.py
```diff
@@ -0,0 +1,251 @@
...
+    term_index = DBTermIndex
+
+    def find_terms(self, original_string, entity_singulars):
```
adngdb: I think this should be a simple function in this module. Maybe something else, I'm not entirely sure. But I strongly believe it shouldn't be a method of this Manager class.

pontoon/sync/changeset.py
```diff
@@ -191,6 +197,13 @@ def execute_create_db(self):
                         fuzzy=vcs_translation.fuzzy
                     ))
 
+        Term.objects.assign_terms_to_entities([(
```
adngdb: This sounds like it's making changes to the ``Entity`` data. If so, why is that function in the ``Term`` model namespace?

pontoon/sync/changeset.py
```diff
@@ -313,12 +326,20 @@ def execute_update_db(self):
...
                 setattr(db_entity, field, value)
 
             if db_entity.is_dirty(check_relationship=True):
                 self.entities_to_update.append(db_entity)
+                updated_entities.append((
+                    db_entity.pk,
+                    db_entity.string,
+                    ' '.join(get_singulars(db_entity.string)),
+                    db_entity.string_plural,
+                    ' '.join(get_singulars(db_entity.string_plural))
```
adngdb: Commas at the end of all lines make diffs easier to read, next time someone adds a new line here. :)

pontoon/terminology/apps.py
```diff
@@ -0,0 +1,7 @@
+from __future__ import unicode_literals
+
+from django.apps import AppConfig
+
+
+class TerminologyConfig(AppConfig):
+    name = 'terminology'
```
adngdb: What's the use of this file? Can you add a docstring to explain?

pontoon/terminology/formats/tbx.py
```diff
@@ -0,0 +1,192 @@
...
+            if child.nodeType == child.TEXT_NODE:
+                continue
+
+            if child.tagName == 'descrip':
+                try:
+                    return child.childNodes[0].data
+                except IndexError:
+                    return ''
+
+            elif child.tagName == 'descripGrp':
```
adngdb: No need to use ``elif`` since you're returning or continuing in both previous if blocks.

pontoon/terminology/formats/tbx.py
```diff
@@ -0,0 +1,192 @@
...
+                return child.childNodes[0].childNodes[0].data
```
adngdb: Why not have the same ``try`` block here than in the previous if block? Pretty sure if you can have a bad structure in the previous case, you can have one here too. :)

pontoon/terminology/formats/tbx.py
```diff
@@ -0,0 +1,192 @@
...
+            return self.xml_node.getElementsByTagName('term')[0].childNodes[0].data
+        except IndexError:
+            return ''
+
+    @property
+    def notes(self):
+        notes_ = {}
+        for note in self.xml_node.getElementsByTagName('termNote'):
+            note_type = str(note.getAttribute('type'))
+            note_text = str(note.childNodes[0].data)
```
adngdb: I think I'd rather be safe pretty much everywhere you're using the XML structure. It's hard to rely on those files to be all perfectly formatted, and it would be better to have failsafe cases everywhere.

pontoon/terminology/management/commands/import_microsoft_terminology.py
```diff
@@ -0,0 +1,181 @@
...
+        # Filter terms with MS specific phrases
+        cleaned_terms = {
+            k: v for k, v in terms.items()
+            if not self.ms_term(v.source_text, v.description)
+        }
+
+        log.info('Removed %s MS Terms.', len(terms) - len(cleaned_terms))
+        Term.objects.import_tbx_terms(cleaned_terms)
+
+    def ms_term(self, source_text, description):
```
adngdb: I'd propose to call this method `contains_ms_term` instead. I think it's clearer.

adngdb: Also, it can be just a function, it doesn't have to be part of this class.

pontoon/terminology/models.py
adngdb: These would benefit from a better docstring. I'm not sure what `entity_singulars` is.

pontoon/terminology/models.py
```diff
@@ -0,0 +1,251 @@
...
+                db_term.description = term.description
+                update_terms.append(db_term)
+                updated_records += 1
+            except KeyError:
+                new_terms.append(
+                    Term(
+                        term_id=term.term_id,
+                        source_term=term.source_text,
+                        source_term_singulars=' '.join(get_singulars(term.source_text)),
+                        note=term.part_of_speech or '',
```
adngdb: I don't understand, why is this called `note` in one case and `part_of_speech` in another?

pontoon/terminology/models.py
```diff
@@ -0,0 +1,251 @@
...
+    def import_tbx_terms(self, tbx_terms):
+        """
+        Creates/updates term objects passed from terminology files.
+        All terms will be automatically matched with their entities.
+
+        :param VCSTerm tbx_terms: a list of terms imported from the filesystem/vcs.
+        """
+        db_terms = {}
+
+        for term in Term.objects.all():
```
adngdb: This is kinda worrisome. How big can this table become? Could this lead to out-of-memory problems? Are we sure it cannot?

pontoon/terminology/models.py
```diff
@@ -0,0 +1,251 @@
...
+        log.info('Updated {} translations of terms.'.format(len(update_translations)))
+        self.assign_terms_to_entities(
+            Entity.objects.values_list(
+                'pk',
+                'string',
+                'string_singulars',
+                'string_plural',
+                'string_plural_singulars'
+            )
+        )
```
adngdb: I am concerned about the performance of all this. It seems to be very very expansive on the database. If I understand correctly, every time we import a TBX file, we'll recreate all the terms, then add all the translations again (it seems, without updating the existing ones, or deleting them), then update all existing entities? That sounds like a lot of work, and could be a real burden to our database. 

Can this be smarter? Is it possible to detect changes, and only update the data that changed? Or is it possible to be smarter about our database schema to avoid updating that much data every time?

adngdb:
I am wondering about your approach of storing the matching terms in the database. It seems quite expansive and not flexible. Every time a single thing changes in the Terminology, we have to rebuild the entire dataset. So, my question is, could we do it differently? Could we detect the words from our Terminology set at run time? Maybe on the server side, maybe on the client side, I don't know. 

Maybe we could build a microservice that takes a string and returns all matching terms? Then we could use that from the client side to enhance the display of entities? Or we could call that service from the django side before rendering the template, even. 

I'm very worried about how this adds a huge layer of complexity over our already complex database schema, and about the maintenance burden this is going to be. I'm afraid over time this code will also cause a lot of performance troubles. 

I know this is questioning the very bases of your work here, and might require a lot more work, but those are honest concerns, and we should aim for the best possible code as soon as possible. So, what are your opinions, @jotes, @mathjazz ?

pontoon/base/static/js/translate.js
```diff
@@ -415,6 +415,24 @@ var Pontoon = (function (my) {
...
+     * Mark Terminology terms in source string
+     *
+     * text Source string
+     */
+    markTerms: function (text, terms) {
+      for (var term in terms) {
+        var re = new RegExp(term, 'gi'),
+            details = encodeURI(JSON.stringify(terms[term]));
+
+        // TODO: Make sure only actual text is marked (avoid placeables and previously marked terms)
```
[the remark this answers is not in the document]
jotes: Javascript code was created by @mathjazz, so I assume it's not ready.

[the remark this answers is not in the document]
mathjazz: It's not done. :)

pontoon/base/static/js/translate.js
```diff
@@ -1741,8 +1761,44 @@ var Pontoon = (function (my) {
...
+      // TODO: Improve positioning to avoid closing the tooltip by mistake; maybe add an arrow
+      $('#original').on('mouseenter', '.term', function () {
+        var terms = JSON.parse(decodeURI($(this).data().details));
+
+        $(this).append('<div id="terminology-tooltip">' +
+          '<ul class="terms"></ul>' +
+          '<footer>' +
+            '<p><a href="https://www.microsoft.com/Language/" target="_blank">Provided by Microsoft Terminology</a></p>' +
+          '</footer>' +
+        '</div>');
```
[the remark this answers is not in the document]
jotes: I agree. Maybe we should create a separate bug to move all those concatenations into templates or something else?

jotes:
Hey,
> I am wondering about your approach of storing the matching terms in the database. It seems quite expansive and not flexible. Every time a single thing changes in the Terminology, we have to rebuild the entire dataset. So, my question is, could we do it differently? Could we detect the words from our Terminology set at run time? Maybe on the server side, maybe on the client side, I don't know

From what I remember, @mathjazz's first idea was to import terminology from Microsoft Portal (once) and then start to edit it inside Pontoon.

Over the time the whole concept changed and one of the last ideas was to add synchronization of TBX files which didn't land in this patch.

I assumed that a lot of this code will be removed (the whole import of tbx files from MS) after an initial phase.
A single .tbx file from MS Terminology Portal contains almost ~30k of entries, but Mozilla will probably use only a small subset of those terms.
I decided to go with the database approach because of the speed  -> a microservice could introduce latency and lower performance because it would have to process entities ad-hoc. 

> Maybe we could build a microservice that takes a string and returns all matching terms? Then we could use that from the client side to enhance the display of entities? Or we could call that service from the django side before rendering the template, even.

Sounds good to me. If you have any links/examples how to add a microservice that would help me to think how to approach this idea.

> I'm very worried about how this adds a huge layer of complexity over our already complex database schema, and about the maintenance burden this is going to be.I'm afraid over time this code will also cause a lot of performance troubles.

Yeah, I think it's somehow introduced by the current architecture of Pontoon. I'm totally in favor in doing more extensible design.

> we should aim for the best possible code as soon as possible.

From my perspective:
A lot of my work was an effect of the previous reviews/methodologies introduced by them.
From my perspective, every tech-leading person judges the quality of the code in a different way (e.g. https://github.com/mozilla/pontoon/pull/257).
I think it would be useful if you could list all of those ideas, because "best possible code" is a vague term . I just want to adjust my code to your standards.

So, I'm in favor of even closing this patch and start thinking how to do this right (I hope that @adngdb will help me with that).

requirements.txt
```diff
@@ -263,6 +272,16 @@ jsonpickle==0.7.0 \
     --hash=sha256:177b0e8311c6edbfc7d184157a7e80f68b6b874c5645f7300d41f475f05f231e
 
 # ----------------------------------------------------------------------------------
+stemming==1.0.1 \
```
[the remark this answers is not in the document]
Pike: Do we even use that? AFAICT, we're using Pattern directly in this patch.

Which has similar problems, it's last release has been ages ago, but the code itself is maintained.

jotes:
So, We discussed issues of this PR during our water-cooling meetings. I had some time to read through all of this code during the weekend.
As I mentioned, I really value all comments sent by @adngdb and I feel they're the good base to start over working on this PR.
I haven't a lot of time recently, but it's going to change during this month and I'm planning to work on this again in cooperation with the rest of Pontoon Team.

PR #538, opened 2017-01-31
WIP: Adding optional branch name to github sync.

ScottDowne:


pontoon/administration/templates/admin_project.html
```diff
@@ -80,6 +80,12 @@ <h1 id="title">
           {{ repo_form.DELETE }}
         </div>
       </section>
+      <section class="branch-wrapper">
+        <div class="branch">
+          {{ repo_form.branch.label_tag() }}
+          {{ repo_form.branch }}
+        </div>
+      </section>
```
mathjazz: We need to add a similar block also to the `.repository-empty` block below, which is used as a template for adding new repositories.

pontoon/base/models.py
```diff
@@ -1082,6 +1082,7 @@ class Repository(models.Model):
         choices=TYPE_CHOICES
     )
     url = models.CharField("URL", max_length=2000)
+    branch = models.CharField("Branch", default='master', max_length=2000)
```
mathjazz: This field should be made optional (`blank=True`) and the default parameter should be removed. We currently don't specify the branch, which means the default branch is used, which is not always called master (e.g. https://github.com/mozilla/activate.mozilla.community/).

This will probably require some changes to pontoon/sync/vcs/repositories.py.

ScottDowne: Agreed, good catch on master isn't always default.

jotes: Also, maybe we could use a TextField with blank=True?

jotes: Also, we could retrieve a name of default branch in the first pull and save it to the database, that would safe us from the unecessary "ifs" in the rest of code.

ScottDowne: I tried a textField just now, but it created a multi line input box. Probably not what we want. I think 2000 characters is probably enough for a branch name ;) Is there another advantage to TextField?

ScottDowne:
Updates. I noticed I need to fix up and add some unit tests. I'll dig into that later.

pontoon/sync/vcs/repositories.py
```diff
@@ -40,8 +42,16 @@ def pull(self, source=None, target=None):
         code, output, error = execute(command, target)
 
         if code == 0:
+
```
mathjazz: Nit: Remove blank line.

pontoon/sync/vcs/repositories.py
```diff
@@ -40,8 +42,16 @@ def pull(self, source=None, target=None):
...
             log.debug("Git: Repository at " + source + " updated.")
 
+            if branch:
+                command = ["git", "checkout", branch]
+                code, output, error = execute(command)
+
+                if code == 0:
+                    log.debug("Git: Branch " + branch + " checked out.")
```
mathjazz: We should raise `PullFromRepositoryException` if code != 0, otherwise Pontoon will silently ignore the branch and fallback to the default.

pontoon/sync/vcs/repositories.py
```diff
@@ -40,8 +42,16 @@ def pull(self, source=None, target=None):
...
+
         else:
             log.info("Git: " + unicode(error))
             log.debug("Git: Clone instead.")
```
mathjazz: If the repo doesn't exist yet, we need to clone it. And again, switch to the right branch. Otherwise, on first sync Pontoon will synchronize with the default branch.

pontoon/sync/vcs/repositories.py
```diff
@@ -40,8 +42,16 @@ def pull(self, source=None, target=None):
         code, output, error = execute(command, target)
 
         if code == 0:
+
             log.debug("Git: Repository at " + source + " updated.")
 
+            if branch:
+                command = ["git", "checkout", branch]
+                code, output, error = execute(command)
```
mathjazz: This command should be ran inside the checkout path, so add `target` as the second argument to `execute()`.

pontoon/sync/vcs/repositories.py
ScottDowne: You didn't notice this before because it's new ;)

I'll fix up.

ScottDowne:
@mathjazz Fixed up the comments and tests.

One thing that doesn't quite add up. Locally when I sync, after making a few changes, it starts to reject the push, and the commit locally is one behind the remote. It's also curious that pull doesn't actually do a pull.

pontoon/sync/vcs/repositories.py
```diff
@@ -39,21 +41,26 @@ def pull(self, source=None, target=None):
...
-            if code == 0:
-                log.debug("Git: Repository at " + source + " cloned.")
+            if code != 0:
+                raise PullFromRepositoryException(unicode(error))
 
-            else:
+        log.debug("Git: Repository at " + source + " cloned.")
+
+        if branch:
+            command = ["git", "checkout", "master"]
```
mathjazz: Hardcoded master. :)

ScottDowne:
Whoops, forgot to remove that. That was me testing the issue. Gone now.

I'm going to do more debugging today. I have vague STRs, and I want to see if I can make it happen on master or not. If I cannot, it's likely a bug in my patch, if I can, it's possibly a bug on master. Seems to be fine on a fresh clone. But if you sync again after more changes on the same repo, it doesn't do a pull, but also doesn't do a fresh clone, and thus looks to be behind by 1 commit.

ScottDowne:
I left a --force in here from debugging as well. I'll remove once I finish up something else.

ScottDowne:
Things still work on master as expected. I have a bug.

ScottDowne:
@mathjazz fixed my issues.

The issues was this line: `command = ["git", "reset", "--hard", remote]`

Would reset my branch to be like the default branch. I added this to fix it:

```
        remote = "origin"
        if branch:
            remote += "/" + branch

        # Undo local changes
        command = ["git", "reset", "--hard", remote]
```

This way, the reset is always based on the custom branch of default.

I think this is ready for some more review :)

pontoon/sync/vcs/repositories.py
```diff
@@ -16,44 +16,55 @@ class PullFromRepositoryException(Exception):
...
         source = source or self.source
         target = target or self.target
+        branch = branch or self.branch
 
         command = ["git", "fetch", "--all"]
         execute(command, target)
 
+        remote = "origin"
+        if branch:
+            remote += "/" + branch
```
mathjazz: Let's move this block under the `# Undo local changes` comment.

pontoon/sync/vcs/repositories.py
```diff
@@ -145,7 +158,11 @@ def commit(self, path=None, message=None, user=None):
             raise CommitToRepositoryException(unicode(error))
 
         # Push
-        push = ["git", "push", self.url, "HEAD"]
+        pushTarget = 'HEAD'
```
mathjazz: Nit: `pushTarget` -> `push_target` for consistency with PEP8.

pontoon/sync/vcs/repositories.py
```diff
@@ -16,44 +16,55 @@ class PullFromRepositoryException(Exception):
...
         # Undo local changes
-        command = ["git", "reset", "--hard", "origin"]
+        command = ["git", "reset", "--hard", remote]
         code, output, error = execute(command, target)
 
-        if code == 0:
-            log.debug("Git: Repository at " + source + " updated.")
-
-        else:
+        if code != 0:
```
mathjazz: I'm not sure I understand this change. You will always print "Repo cloned" debug message, even if the repo is actually just updated.

pontoon/sync/vcs/repositories.py
ScottDowne: I was on the fence about this one, actually. So glad to have a decision made on this ;)

pontoon/sync/vcs/repositories.py
ScottDowne: Ah, yes, not my intent. I'll fix.

ScottDowne:
@mathjazz updated. I left it as a separate commit so you don't need to go through previous changes, however, we should probably squash the commits when we merge.

mathjazz:
Looks good!

I'd still love to have that `log.debug("Git: Repository at " + source + " updated.")` debug message restored and then we're good to squash and merge.

ScottDowne:
@mathjazz Good catch, I also didn't intend to remove that. Fixed. I think it's now how it was originally intended.

mathjazz:
A thing of beauty! Welcome to Pontoon, @ScottDowne! 🎉

PR #540, opened 2017-02-09
Fix bug 1338276 upgrade django to 1.10.5

jotes:


pontoon/base/models.py
```diff
@@ -453,16 +453,16 @@ class Locale(AggregatedStats):
         (5, 'other'),
     )
 
-    cldr_plurals = models.CommaSeparatedIntegerField(
+    cldr_plurals = models.CharField(
         "CLDR Plurals",
         blank=True,
         max_length=11,
-        validators=[validate_cldr],
+        validators=[validate_comma_separated_integer_list, validate_cldr],
```
jotes: I'll remove `validate_comma_separated_integer_list`.

requirements.txt
```diff
@@ -147,3 +146,8 @@ requests-oauthlib==0.6.1 \
     --hash=sha256:7c708e8e2a4aa6a905cf91f28420d75db37270e0ec8fc951915c098dd8bde53e
 django-csp==3.1 \
     --hash=sha256:9208219c341ddbe371b5fd217ced0c916a5e2f7184bc603415074afcace6b51c
+html5lib==1.0b10 \
+    --hash=sha256:08a3efc117a4fc8c82c3c6d10d6f58ae266428d57ed50258a1466d2cd88de745 \
+    --hash=sha256:0d5fd54d5b2b79b876007a70c033a4023577768d18022c15681c00561432a0f9
+webencodings==0.5 \
+    --hash=sha256:a5c55ee93b24e740fe951c37b5c228dccc1f171450e188555a775261cce1b904
```
jotes: I experimented with my auto-bumping script, it looks like it added some optional dependecies. I'll remove them.

jotes:
ad1. Actually, you can see those warnings when you run e.g. `python -W once ./manage.py runserver`
ad2. New version of django introduced two migrations in auth module:
* https://github.com/django/django/blob/1.10.5/django/contrib/auth/migrations/0007_alter_validators_add_error_messages.py
* https://github.com/django/django/blob/1.10.5/django/contrib/auth/migrations/0008_alter_user_username_max_length.py

I think it's a pretty rare situation, previous migrations where added 2 years ago.

jotes:
@mathjazz Hey, could you look again?

jotes:
I updated hash for chardet package because the maintainers released a `.whl` distribution of this package and that broke the build.

More info is available here:
https://github.com/chardet/chardet/commit/b33477d7a802493e12bb176c825e36ee2e7c398c

mathjazz:
Thanks!

Tested on production and (of course) all migrations get applied, including the one for the `auth` module, because `post_compile` hook takes care of it.

PR #541, opened 2017-02-10
Fix bug 1305233: Properly match changes in repository with enabled locales

mathjazz:
The old code checked if changed paths in repository start with paths to locale directories without the trailing slash. If the changed path was `kab/mozorg/internet-health/privacy-security.lang`, it would match it with the Georgian locale directory `ka` and break.

The new code adds trailing slashes to locale directory paths, so it checks if the path starts with `ka/` (no match) and then continues all the way to `kab/` (match).

@jotes r?

@jotes I didn't add any tests yet, mostly because #405 is still open. Are you planning on continuing with that or is it too old to be useful now?

### DIFF
--- a/pontoon/sync/vcs/models.py
+++ b/pontoon/sync/vcs/models.py
@@ -153,8 +153,10 @@ class VCSProject(object):
                     repo.get_last_synced_revisions()
                 )[0]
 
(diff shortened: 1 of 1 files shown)


jotes:
Those tests require a rebase, I'll look into this during this weekend.

PR #542, opened 2017-02-15
Fix bug 1336984: Update to latest LTS node.js

mathjazz:
Also update babel to latest version.

@jotes r?


jotes:
I think we're fine if deployment on heroku passes without errors.

mathjazz:
Yeah I tested on stage and it worked without a problem.

PR #543, opened 2017-02-16
Fix bug 1340283. Fixed ProjectForm locales validation.

jotes:


jotes:
@mathjazz Hey, could you review?

pontoon/sync/tests/test_changeset.py
```diff
@@ -479,8 +479,8 @@ def test_plural_translations(self):
 
         self.changeset.execute_update_vcs()
 
-        assert_equal(self.changeset.commit_authors_per_locale[self.translated_locale.code],
-            [first_author, third_author])
+        assert_equal(set(self.changeset.commit_authors_per_locale[self.translated_locale.code]),
```
jotes: This one fixed a test that was failing randomly from time to time.

mathjazz:
Thanks for the patch. This fixes the problem, stated in the bug.

However, there seems to be another regression, related to the Django upgrade. The repository fileds (type, url, download prefix) are now.

And, the migration introduced in https://github.com/mozilla/pontoon/commit/207773fe6d0d0e6eb67697572320f2c8246b245c doesn't resemble the change in the model. 

Would you like to fix these two issues, too? If not, I can file separate bugs.

jotes:
@mathjazz Ok, I updated migrations and added constraints on repo forms. Could you look again?

pontoon/administration/forms.py
```diff
@@ -11,6 +11,10 @@ class Meta:
                   'info_brief', 'url', 'width',
                   'links', 'disabled')
 
+    def __init__(self, *args, **kwargs):
+        super(ProjectForm, self).__init__(*args, **kwargs)
+        self.fields['locales'].required = False
+
```
mathjazz: This makes locales optional, which is not what we want.

The problem started after we upgraded Django to 1.10, which apparently adds required attributes to required fields, which forces validation by the browser. Validation naturally happens before the javascript executes, which populates the locales field. So we need to do something like this:

```
// Submit form with button
$('.save').click(function(e) {
  e.preventDefault();
  $('#admin-form').submit();
});
```

Also, we should add `{{ repo_form.url.errors }}` under `{{ repo_form.url }}` to show errors for the repository URL fields.

jotes: I was pretty convinced that locales where optional some time ago that's why I fixed it this way. Ok, I know what to do.

mathjazz:
@jotes r?

jotes:
:+1:

PR #544, opened 2017-02-17
Dashboard

mathjazz:
@jotes r? This is huge. The following text will try to provide more context and help you focus on the imporant changes. If it still feels too heavy for you to review, please let me know. This pull request blocks [notifications](https://bugzilla.mozilla.org/show_bug.cgi?id=983092), so I'd like to land it in the next week, max two.

The initial idea was to simply add the ability to track deadlines and prioritize projects. It turned out we needed to redesign dashboards to fit new columns, so I revived my old work on [bug 1203543](https://bugzilla.mozilla.org/show_bug.cgi?id=1203543), which took care of that.

Here's a short summary of the changes:

**1. The `pontoon.base` app has been split into several (5) new apps:**
* `pontoon.teams`
* `pontoon.projects`
* `pontoon.localizations` (i.e. project-locales, accessible via /locale-code/project-slug/)
* `pontoon.machinery` (previously referred to as Terminology in the UI)
* `pontoon.contributors`

I moved URL patterns, views, templates and static files to new apps. I also refactored some of the macros, widgets and includes to better suit changes described in the following two items.

The `pontoon.base` app is still around and contains reusable resources, models, translate view (should be our next target to split into a spearate app) and the terms page. Hopefully, the code will now be easier to read and maintain.

**2. Richer dashboards.**
Additional information can now be assigned to projects:
* _deadline_ (styled differently if overdue or approaching)
* _priority level_ (1 to 5)
* _public repository website_ (e.g. https://github.com/mozilla/pontoon-intro/)

Team (locale) dashboards also include new information:
* _writing direction_ (left-to-right or right-to-left)
* _script_ (e.g. latin)
* _number of literate speakers_ (taken from CLDR)
* _bugs_ (bugzilla query stolen from elmo)

Project, Team, and Localization dashboards have received the biggest changes:
* Navigation between their subpages (e.g. between Slovenian projects and contributors) is now AJAX-based, which should be faster. Links and history should still work as expected. Pages loaded via AJAX live in the `includes` folder withing their corresponding app.
* Together with Teams, Projects and Top Contributors pages, they introduce a richer heading with more information about teams, projects, top contributor in the last month, project, team, localization and the current stats.
* They contain direct links to translate page for each item in their listing of teams, projects or resources. Hover the item and under the horizontal progress bar you will notice links for each of the status filters.

**3. Visual refresh.**
* Header on top of each page has been simplified with less frequent actions moved to the menu on the right hand side.
* Header contains a placeholder for notifications, which now slide down from the top of the page. 
* There's also a slim YouTube-like horizontal progress bar on top of the header, which shows up on AJAX requests, e.g. as you navigate between project dasboard subpages.
* Since header is included almost everywhere, the majority of pages have been slightly redesigned to match the new design (color scheme, title style, form widgets).
* The default page width has been increased to 980px (from 735px).

**Review instructions**
To help with testing and the review, I'm listing all the pages the redesign affects, which is pretty much all of our pages. There are 3 commits, only because I'd like to keep the middle one isolated for now, so don't worry about that too much. (it removes the progress over time chart, which we'll add in the future, but we don't have all the required data just yet).

I'm particularly interested in your thoughts about the changes related to:
* Splitting `pontoon.base` app
* Template refactoring
* AJAX-based views
* Visual design changes
```
http://localhost:8000/teams/
http://localhost:8000/en-GB/
http://localhost:8000/projects/
http://localhost:8000/projects/pontoon-intro/
http://localhost:8000/en-GB/pontoon-intro/
http://localhost:8000/en-GB/pontoon-intro/all-resources/
http://localhost:8000/contributors/
http://localhost:8000/profile/
http://localhost:8000/settings/
http://localhost:8000/machinery/
http://localhost:8000/admin/
http://localhost:8000/admin/projects/
http://localhost:8000/sync/log/
http://localhost:8000/sync/log/1/
http://localhost:8000/500/
http://localhost:8000/404/
http://localhost:8000/403/
http://localhost:8000/terms/
http://localhost:8000/accounts/standalone-login/
```

**TODO**
- [x] Fix tests (views tests need to be moved to their respective apps and refactored for AJAX)
- [x] We'll probably need some DB optimization, particularly for the latest activity column.


TheoChevalier:
I wonder if the links could/should take more space, maybe the full height?

giphy

(And the new design & dashboards rock 🤘)

pontoon/administration/views.py
```diff
@@ -1,31 +1,38 @@
...
+
 
 log = logging.getLogger('pontoon')
 
 
 def admin(request, template='admin.html'):
     """Admin interface."""
     if not request.user.has_perm('base.can_manage'):
-        return render(request, '403.html', status=403)
+        return views.error403(request)
```
jotes: I would use `from django.core.exceptions import PermissionDenied`.

pontoon/base/migrations/0082_populate_repository_website.py
```diff
@@ -0,0 +1,33 @@
...
+    for repository in Repository.objects.all():
+        repository.website = (
+            re.sub(re.escape('.git') + '$', '', repository.url)
+            .replace('git@github.com:', 'https://github.com/')
+            .replace('ssh://', 'https://')
+        )
+        repository.save()
+
+
+def noop(apps, schema_editor):
```
jotes: I recently discovered `migrations.RunPython.noop` :D

pontoon/base/models.py
```diff
@@ -593,6 +615,10 @@ def create_users_map(qs, keyfield):
 
         return locale_projects
 
+    @property
+    def contributors(self, start_date=None):
```
jotes: I think you won't be able to pass start_date. You could do somethin like this: http://stackoverflow.com/questions/5715620/python-how-to-pass-more-than-one-argument-to-the-property-getter and create a separate method called `get_contributors_since` and create a property that uses specific call of this method.

pontoon/base/models.py
```diff
@@ -1335,6 +1378,13 @@ def is_asymmetric(self):
     def __unicode__(self):
         return '%s: %s' % (self.project.name, self.path)
 
+    def save(self, *args, **kwargs):
+        super(Resource, self).save(*args, **kwargs)
+
+        if self.deadline < self.project.deadline:
```
jotes: What will happen if there will be a couple of resouces with different deadlines:

* resource1 - 2017-02-01
* resource2 - 2017-01-10
I think that current implementation will choose 2017-01-10 as a deadline for the whole project.

mathjazz: Correct. And that's expected.

mathjazz:
Good point @TheoChevalier, let me know if it's better now.

pontoon/base/models.py
mathjazz: Thanks for the advice! Turns out, we don't even use (or need) that parameter.

TheoChevalier:
@mathjazz Looks better, thanks 👍

pontoon/base/models.py
```diff
@@ -421,6 +454,18 @@ def prefetch_latest_translation(self, project):
             )
         )
 
+    def get_stats_sum(self):
```
jotes: I think we can move it into a mixin class.

pontoon/base/templates/widgets/latest_activity.html
```diff
@@ -1,12 +1,16 @@
...
 <span class="latest">
   {% if latest_activity.date %}
-  <time datetime="{{ latest_activity.date.isoformat() }}">{{ latest_activity.date|naturaltime }}</time>
     {% if latest_activity.user %}
-    &bull; <a href="{{ url('pontoon.contributor.username', latest_activity.user.username) }}">{{ latest_activity.user.name_or_email|nospam }}</a>
+      {% set action = latest_activity.type + ' by ' + latest_activity.user.name_or_email %}
+    {% else %}
+      {% set action = 'imported' %}
     {% endif %}
+    <a href="{{ url('pontoon.translate', latest_activity.translation.locale.code, latest_activity.translation.entity.resource.project.slug, resource) }}?string={{ latest_activity.translation.entity.id }}" title="Translation '{{ latest_activity.translation.string }}' {{action}}">
```
jotes: I noticed that `/teams/` view started to take 30 seconds to load on my local machine. After I commented out `latest_activity.html`, loading time dropped to 10 seconds. I wonder if we should prefetch these relations or even denormalize last activity.

pontoon/base/models.py
```diff
@@ -494,15 +556,20 @@ def serialize(self):
...
         }
 
-    def cldr_plurals_list(self):
+    def cldr_id_list(self):
         if self.cldr_plurals == '':
             return [1]
         else:
             return map(int, self.cldr_plurals.split(','))
 
+    def cldr_plurals_list(self):
```
jotes: I think we could change all param-less methods into properties.

pontoon/base/templates/widgets/latest_activity.html
jotes: Ok, After I changed my machine to `/teams/` take 10 seconds to load. Which is better but We still execute 600 sql queries in this view, but I think We can optimize it to load faster.

pontoon/base/static/js/tabs.js
```diff
@@ -0,0 +1,168 @@
...
+            var tbody = $('<tbody>'),
+                formatter = new Intl.DateTimeFormat('en-GB', {
+                  day: 'numeric',
+                  month: 'long',
+                  year: 'numeric'
+                });
+
+            $.each(data.bugs, function(i, bug) {
+              tr = $('<tr>', {
+                title: bug.summary
```
jotes: I wonder What if somebody will create a bug with a malformed summary with something like a `<script>alert()</script>`?

pontoon/base/static/js/table.js
```diff
@@ -0,0 +1,125 @@
...
+              return (chartA.translated - chartB.translated) * dir ||
+                (chartA.suggested - chartB.suggested) * dir;
+
+            // Sort by date
+            } else if (node.is('.deadline') || node.is('.latest-activity')) {
+              return (getDate(b) - getDate(a)) * dir;
+
+            // Sort by priority
+            } else if (node.is('.priority')) {
+              return (getPriority(b) - getPriority(a)) * dir;
```
jotes: I wonder if we can generalize all of those comparators?

pontoon/base/templates/widgets/latest_activity.html
mathjazz: Yup. This is already on the TODO list above. 30 seconds (and even 10) sounds a lot though, it works a little bit faster for me on stage (https://mozilla-pontoon-staging.herokuapp.com/teams/).

pontoon/base/static/js/table.js
mathjazz: E.g. by using a `data-sort` attribute with integer value representing different value types?

pontoon/base/models.py
mathjazz: Yeah. I'd keep this for the next PR to avoid making this one even bigger.

Will mark all such suggestions with :hooray.

pontoon/base/static/js/tabs.js
mathjazz: Good point. I forgot to mention this is a code from Elmo, where it's been in use for a while. Maybe Bugzilla has some protection, too. But still, shall we use `Pontoon.doNotRender()`?

jotes: I simulated a situation when bugzilla returns summary which contains script tags and I successfully displayed an js alert. From what I checked, bugzilla returns raw contents of the summary in the api (e.g. https://bugzilla.mozilla.org/rest/bug/344890). So, If I'm not wrong, then doNotRender() would be advised here.

pontoon/base/static/js/table.js
jotes: :+1:

pontoon/base/static/js/tabs.js
mathjazz: https://github.com/mozilla/pontoon/pull/544/commits/b15fe039b81c81801f9876fc54b864a789bb8abc

mathjazz:
@jotes I've landed optimizations in the following commits: dee5b16 ebce1b1 c66763e 53e3996. All pages should perform at least as good as in production. The code is deployed to stage. Please see commit messages for more details.

Things to note:
* Entries in the latest activity column no longer link to the translate view.
  * We'll need to denormalize a la AggreatedStats for this feature to perform well.
* All variants of the /contributors page take a few seconds to load (same as in production). 
  * Seems like another candidate for denormalization.
* Contributors count in the Team, Project, Localization dashboard tab is disabled.
  * Denormalization should solve this problem too.

mathjazz:
Tests and optimizations finished, this is now ready for the final review.

pontoon/base/views.py
```diff
@@ -71,172 +55,49 @@ def home(request):
...
-        .order_by('name')
-    )
-
-    if not projects:
-        raise Http404
-
-    return render(request, 'locale.html', {
-        'projects': projects,
-        'locale': l,
+def error403(request):
```
jotes: Why do we need a separate view for 403? Django automatically loads 403.html when you throw authorization error?

pontoon/base/views.py
```diff
@@ -71,172 +55,49 @@ def home(request):
...
-        'locale': l,
-        'all_users': all_users,
-        'contributors': contributors,
-        'translators': translators,
-        'managers': managers,
-        'locale_projects': locale_projects,
-        'project_locale_form': project_locale_form,
-        'all_projects_in_translation': all([x[5] for x in locale_projects])
+def error404(request):
+    return render(request, '404.html', {
```
jotes: the same as above.

jotes:
@mathjazz 
I totally admire visual changes you've made -> dashboard is looking awesome for me (as a developer) :D

A few notes from my field of expertise:

* a huuuuge :+1: for your split of pontoon.base into smaller apps. That's makes this PR really hard to read ;D but I feel that's a step into the right direction.
* I think we could do this in a series of smaller pull requests -> that would it make easier to rebase, review and merge to the master.
* After skimming through current version I think we can land this but... It's hard to give a binary answer to a such big PR :D
* I can't wait until this see master :)

pontoon/base/views.py
mathjazz: Nice catch, this is no longer used.

pontoon/base/views.py
mathjazz: the same as above. :D

pontoon/base/migrations/0044_locale_translators.py
```diff
@@ -9,8 +9,9 @@ def create_translators(apps, schema_editor):
     Bug 952488 - We're assigning translation permissions to active contributors.
     """
     User = apps.get_model('auth', 'User')
+    users = User.translators if hasattr(User, 'translators') else User.objects
```
jotes: users = getattr(User, 'translators', User.objects)

mathjazz:
@jotes Thank you so much for the review! I understand it was impossible to review due to size, but you did a great job nevertheless.

Chunks of this code were written months ago, but for various reasons didn't made it into a PR. And now they became so interconnected, it was hard to split them into smaller PRs. Next time I'll definitely make sure to do so and having `pontoon.base` split into multiple apps should help with that.

I'm writing a blog post now and then I'll merge this and deploy to production (it's already on stage).

PR #548, opened 2017-03-01
Fix translate links on localization dashboard

mathjazz:
@jotes r?


jotes:
hey, I checked this locally and looks like it's okay. The only think I'm wondering: maybe we there's no point to display link if counter displays '0'? because that point's to empty localization page.

mathjazz:
I just played locally with not linking if the counter is 0. I don't like it too much and I'd prefer to be consisten. Maybe you still want to quickly share a link to show there are 0 missing strings.

PR #552, opened 2017-03-02
No more red in latest activity column

mathjazz:
Red color implies errors, so we're replacing with "Pontoon green" in the latest activity column and removing the underline. "No activity yet" text has been replaced with &middot; to make it easier to distinguish from other entries. Also in the deadline column.

@MikkCZ @mastizada r?

Check it out on stage:
https://mozilla-pontoon-staging.herokuapp.com/projects/


MikkCZ:
I cannot tell much about the code, but it looks better in staging. :) r+

mastizada:
+1

PR #558, opened 2017-03-06
More fields

mathjazz:
Added a few more fields to the Locale and Project model and fixed bug 1309287 along the way.

@jotes r?


pontoon/base/templatetags/helpers.py
```diff
@@ -231,3 +231,9 @@ def pretty_url(url):
         url = url[4:]
 
     return url
+
+
+@library.filter
+def local_url(url, code='en-US'):
```
jotes: I'm not sure if that's a misspell - did you mean `locale_url`?

mathjazz: This is intentional. I'm trying to be consistent in using adjectives, we have `pretty_url` and `local_url`.

PR #559, opened 2017-03-07
Fix management command

mathjazz:
@jotes r?


mathjazz:
Fixed tests.

pontoon/base/migrations/0044_locale_translators.py
```diff
@@ -9,7 +9,7 @@ def create_translators(apps, schema_editor):
     Bug 952488 - We're assigning translation permissions to active contributors.
     """
     User = apps.get_model('auth', 'User')
-    users = getattr(User, 'translators', User.objects)
+    users = getattr(User, 'translators', getattr(User, 'objects', None))
```
mathjazz: This should now finally work both, locally and on GitHub.

PR #561, opened 2017-03-08
Faster HG pull()

mathjazz:
Instead of removing the entire repository folder and running `hg clone` on each sync, we try to do something similar as `git reset --hard`:
* discard uncommitted changes,
* delete untracked files,
* discard local commits.

We need two HG extensions to make this work. They are enabled inside each repo's `.hg/hgrc file` on each sync. The code to do that would be more elegant if we could enable the extensions in the system/user `hgrc` file, but I'm not sure where to put it. I tried [several paths](https://www.selenic.com/mercurial/hgrc.5.html), but none them actually enabled the extensions. Mercurial is installed via pip.

We currently clone ~200 HG repositories during each sync cycle. Quick test showed that this patch reduces the average repo `pull()` time from 5 to 2 seconds (except afer dyno restart of course). It's hard to predict how much time can this save us per sync cycle due to parallelization, but I'm hoping for a couple of minutes per sync cycle.

(Also including two other minor fixes I found along the way.)

@jotes r?

### DIFF
--- a/pontoon/sync/vcs/repositories.py
+++ b/pontoon/sync/vcs/repositories.py
@@ -71,15 +71,33 @@ class PullFromGit(PullFromRepository):
 class PullFromHg(PullFromRepository):
 
     def pull(self, source=None, target=None):
-        log.debug("Mercurial: Update repository.")
-
         source = source or self.source
         target = target or self.target
 
-        # Undo local changes: Mercurial doesn't offer anything more elegant
-        command = ["rm", "-rf", target]
-        code, output, error = execute(command)
+        if os.path.exists(target):
+            log.debug("Mercurial: Update repository.")
 
+            # Undo local changes: discard uncommitted changes
+            command = ["hg", "update", "--clean"]
+            execute(command, target)
+
+            # Undo local changes: delete untracked files
+            command = ["hg", "--config", "extensions.strip=", "purge"]
+            execute(command, target)
+
+            # Undo local changes: discard local commits
+            command = ["hg", "--config", "extensions.strip=", "strip", "'roots(draft())'"]
+            execute(command, target)
+
+            command = ["hg", "pull", "-u"]
+            code, output, error = execute(command, target)
+
+            if code == 0:
+                return log.debug("Mercurial: Repository at " + source + " updated.")
+            else:
+                log.info("Mercurial: " + unicode(error))
+
+        log.debug("Mercurial: Clone repository.")
         command = ["hg", "clone", source, target]
         code, output, error = execute(command)
 
--- a/pontoon/administration/static/css/admin_project.css
+++ b/pontoon/administration/static/css/admin_project.css
@@ -124,7 +124,8 @@ form .button {
   width: 130px;
 }
 
-.controls .button.sync {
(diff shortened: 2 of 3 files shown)


pontoon/sync/vcs/repositories.py
```diff
@@ -71,15 +71,41 @@ def pull(self, source=None, target=None, branch=None):
...
 
-        # Undo local changes: Mercurial doesn't offer anything more elegant
-        command = ["rm", "-rf", target]
-        code, output, error = execute(command)
+        if os.path.exists(target):
+            log.debug("Mercurial: Update repository.")
+
+            # Make sure hg extensions needed to undo local changes are enabled
+            with open(os.path.join(target, '.hg/hgrc'), 'a+') as hgrc:
+                pontoon_hook = '# pontoon-extensions\n'
```
jotes: `'\n# pontoon-extensions\n'`

pontoon/sync/vcs/repositories.py
```diff
@@ -71,15 +71,41 @@ def pull(self, source=None, target=None, branch=None):
...
+                hgrc.seek(0)
+                if not pontoon_hook in hgrc.read():
+                    hgrc.write('\n[extensions]\n')
+                    hgrc.write('strip =\n')
+                    hgrc.write('purge =\n')
+                    hgrc.write(pontoon_hook)
+
+            # Undo local changes: discard uncommitted changes
+            command = ["hg", "update", "--clean"]
```
jotes: nit: I'd create a smaller wrapper of those two calls.

pontoon/sync/vcs/repositories.py
```diff
@@ -71,15 +71,41 @@ def pull(self, source=None, target=None, branch=None):
...
+            execute(command, target)
+
+            # Undo local changes: discard local commits
+            command = ["hg", "strip", "'roots(outgoing())'"]
+            execute(command, target)
+
+            command = ["hg", "pull", "-u"]
+            code, output, error = execute(command, target)
+
+            if code == 0:
```
jotes: Should we raise an error if code != 0? I think it would be good to log what happened

pontoon/sync/vcs/repositories.py
```diff
@@ -71,15 +71,41 @@ def pull(self, source=None, target=None, branch=None):
...
         target = target or self.target
 
-        # Undo local changes: Mercurial doesn't offer anything more elegant
-        command = ["rm", "-rf", target]
-        code, output, error = execute(command)
+        if os.path.exists(target):
+            log.debug("Mercurial: Update repository.")
+
+            # Make sure hg extensions needed to undo local changes are enabled
+            with open(os.path.join(target, '.hg/hgrc'), 'a+') as hgrc:
```
jotes: I'd suggest to add a log info about what happens here because it may be later confusing to users if they'll notice unwanted changes in their hgrc file.

mathjazz:
Thanks for the review! I added a 4th commit, which I'll squash into the 3rd after the review, to address your comments and make extension configuration much more elegant. Sorry for not doing that as part of the original patch already.

PR #562, opened 2017-03-08
Copy locales from another projects

mathjazz:
Adding ability to copy enabled locales from one project to another.

Split into two commits for easier review.

@jotes r?

### DIFF
--- a/pontoon/administration/templates/admin_project.html
+++ b/pontoon/administration/templates/admin_project.html
@@ -37,15 +37,29 @@
       {{ form.slug }}
       {{ form.slug.errors }}
   </div>
+
+  <h2>Locales</h2>
+
   <div class="locales clearfix">
       {{ multiple_locale_selector.render(locales_available, locales_selected) }}
       {{ form.locales }}
       {{ form.locales.errors }}
   </div>
-  <div class="checkbox clearfix">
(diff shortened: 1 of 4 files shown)


pontoon/administration/templates/admin_project.html
```diff
@@ -37,15 +37,29 @@
...
+  <div class="locales-toolbar clearfix">
+    <section class="can-be-requested checkbox clearfix">
       <label for="id_can_be_requested">
-          {{ form.can_be_requested }}{{ form.can_be_requested.help_text }}
+          {{ form.can_be_requested }}Locales can opt-in
       </label>
+    </section>
+    <section class="copy-locales clearfix">
+      <select id="copy-locales">
+        <option value="{{ []|to_json() }}">Copy locales from another project</option>
```
jotes: A total nit but `<option value="[]">` could be an alternative solution.

jotes: After looking on it I think that this particular item overlaps with "remove all" button and imho it could be disabled.

jotes: Also, if user changes adds/removes items to chosen locales, this select should return to its initial state.

PR #564, opened 2017-03-10
Reflect PM wishes in the new Project & Locale fields

mathjazz:
Replace Locale style_guide field and Project preview_url & project_url fields with more general ExternalResource model, consisting of name and URL. Users can now add multiple resources to both, Teams and Projects and we show them in Team, Project and Localization dashboard headings.

Remove `project_contact` field and add `admin_notes`. Rename l10n_contact to `contact`.

Create a Project Manager group and only lists members of the group in the Project Admin as potential Contacts. Grant can_manage_project permission to group members (renamed from can_manage). Enable Groups in Django Admin.

@jotes r?


mathjazz:
Force pushed an update to Admin Form.

PR #566, opened 2017-03-22
Add ability to automatically log in a user

mathjazz:
Automatically logs in the user with username `ADMIN_EMAIL` and password
`ADMIN_PASSWORD` if `AUTO_LOGIN` enabled. Useful in demo environment.

@jotes r?

### DIFF
--- a/pontoon/base/middleware.py
+++ b/pontoon/base/middleware.py
@@ -1,4 +1,5 @@
 from django.conf import settings
+from django.contrib import auth
 from django.contrib.sites.models import Site
 from django.shortcuts import redirect
 from django.core.exceptions import PermissionDenied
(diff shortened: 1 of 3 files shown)


pontoon/base/middleware.py
```diff
@@ -47,3 +50,20 @@ def process_request(self, request):
...
+class AutomaticLoginUserMiddleware(object):
+    """
+    This middleware automatically logs in the user with username
+    'ADMIN_EMAIL' and password 'ADMIN_PASSWORD' if AUTO_LOGIN enabled.
+    """
+    def process_request(self, request):
+        if settings.AUTO_LOGIN and not request.user.is_authenticated():
+            user = auth.authenticate(
+                username=os.environ.get('ADMIN_EMAIL'),
+                password=os.environ.get('ADMIN_PASSWORD')
```
jotes: A kinda low-priority note: Auth backends could help you to avoid keeping plain-text password in the environment. http://stackoverflow.com/questions/2787650/manually-logging-in-a-user-without-password

pontoon/settings/base.py
```diff
@@ -168,6 +172,7 @@ def _site_url():
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
     'django.middleware.security.SecurityMiddleware',
     'csp.middleware.CSPMiddleware',
+    'pontoon.base.middleware.AutomaticLoginUserMiddleware',
```
jotes: nit: f you would override this particular setting on the demo instance the AUTO_LOGIN flag wouldn't be required.

PR #568, opened 2017-04-01
Fix bug 1350695 Suggest button posts suggestions between different browser tabs.

jotes:
Hi @mathjazz 
I take a small break from Terminology, I'll do a few small bugs and then resume working on Terminology (this weekend)
Could you look at this?

### DIFF
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -2399,7 +2399,8 @@ var Pontoon = (function (my) {
           original: entity['original' + self.isPluralized()],
           ignore_check: $('#quality').is(':visible') || !syncLocalStorage,
           approve: self.approvedNotSubmitted || false,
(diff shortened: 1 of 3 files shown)


pontoon/settings/base.py
```diff
@@ -114,7 +114,7 @@ def _site_url():
 
 # Log emails to console if the SendGrid credentials are missing.
 if EMAIL_HOST_USER and EMAIL_HOST_PASSWORD:
-    EMAIL_BACKEND = 'django.core.mail.backends.smtp.EmailBackend'
+    EMAIL_BACKEforce_suggestionsND = 'django.core.mail.backends.smtp.EmailBackend'
```
mathjazz: Whooops. :-)

jotes: I had to push it before work and I didn't notice. Could you look again?

PR #571, opened 2017-04-05
Fix bug 1353135: Use different font for Serbian and Macedonian cursive

kostich:
Use Ubuntu Regular font for Serbian and Macedonian Cyrillic cursive in translation-string
<span> elements because Open Sans doesn't support Serbian and Macedonian variant forms
for lower case cursive letters b, g, d, p and t.


pontoon/base/static/css/translate.css
```diff
@@ -1087,6 +1087,11 @@ body > header aside p {
   padding-left: 2%;
 }
 
+#entitylist > .wrapper > ul > li p span.translation-string:lang(mk),
+#entitylist > .wrapper > ul > li p span.translation-string:lang(sr) {
```
mathjazz: Please use the standard [attribute selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors):
`#entitylist > .wrapper > ul > li p span.translation-string[lang="sr"]`

Surprisingly to me, your _colon_ syntax also works fine (at least in Firefox), but let's stick to the standard.

kostich:
Updated the syntax to the standard one.

mathjazz:
Excellent job!

PR #574, opened 2017-04-10
Add custom error pages. With Tetris!

mathjazz:
I thought it would be nice to create some fun error pages. It should make landing #554 more fun.

@jotes r?

What's the best way of keeping the source code and hosting these pages? Shall I create a separate repository? Or use a `gh-pages` branch?


jotes:
I'd keep them in the main repository at master branch because Heroku won't serve them and we have to upload them CDN(S3+cloudfront?)

mathjazz:
As agreen on IRC, once this is merged, I'll put files in a separate repository and make use of GitHub Pages to serve files.

PR #575, opened 2017-04-13
Fix bug 983092: send notifications manually

mathjazz:
This PR fixes [bug 983092](https://bugzilla.mozilla.org/show_bug.cgi?id=983092) by adding the ability for PMs to send manual notifications:
* _Notifications_ tab has been added to the Project dashboard.
* PMs can compose new or view sent notifications.
* Notification is sent to all or a subset of project teams.
* Two shortcuts are available for selecting teams: complete (100%) & incomplete.
* Notification display has been adapted to support manually sent notifications.
* Code behind the view with left sidebar acting as menu has been made more reusable.

My plan is to land this feature and then file bugs for a few related issues:
* Send notifications if new symmetric file added. And take into account that some new strings might be pretranslated in the repository (e.g. Firefox Beta or when files are moved in the l10n repo on merge day). Only send notifications to contributors to locales which got new project strings (e.g. Engagement).
* Create a `pontoon.notifications` app to replace `django-notifications` (underused). This should allow for a much nicer code structure. Also think about using `bulk_create` for sending multiple notifications in one DB hit.
* Add similar Send notifications functionality in the Team dashboard. Talk to PMs if we need it. And also about other features (delete?).

@jotes r?


pontoon/projects/views.py
```diff
@@ -62,6 +71,90 @@ def ajax_info(request, slug):
...
+@require_AJAX
+def ajax_notifications(request, slug):
+    """Notifications tab."""
+    project = get_object_or_404(Project.objects.available(), slug=slug)
+    available_locales = project.locales.order_by('name')
+
+    # Send notifications
+    if request.method == 'POST':
+        message = request.POST.get('message', None)
+        selected_ids = request.POST.get('selected_locales')
```
jotes: I think we could use a django form here to give validation the better structure.

pontoon/projects/views.py
```diff
@@ -62,6 +71,90 @@ def ajax_info(request, slug):
...
+        if available_ids != selected_ids:
+            contributors = User.objects.filter(
+                translation__entity__resource__project=project,
+                translation__locale__in=available_locales.filter(id__in=selected_ids)
+            )
+
+        identifier = uuid.uuid4().hex
+        for contributor in contributors.distinct():
+            notify.send(
+                request.user,
```
jotes: I wonder if we could keep this here, because every send will cause an insert to database:
https://github.com/django-notifications/django-notifications/blob/master/notifications/models.py#L270
If I'm reading that code correctly.
If there's a lot of contributors, it may be reasonable to defer this task to celery and build UI on top of that.

However, I don't have any benchmarks, maybe that's not the problem.

pontoon/projects/views.py
mathjazz: Good point!

pontoon/projects/views.py
mathjazz: I have benchmarks. :) We've been using the same code for automated notifications and it performs good enough to be used via `XMLHttpRequest`, too.

The slow part is actually retrieving the list of contributors, which is also hitting us on /contributors pages. We should figure out the way to optimize it.

But in general I agree - we should defer this to celery if the requests start timing out.

mathjazz:
Updated with Django Forms. @jotes r?

pontoon/projects/views.py
```diff
@@ -102,7 +94,7 @@ def ajax_notifications(request, slug):
         # For performance reasons, only filter contributors for selected
         # locales if different from all project locales
         available_ids = sorted(list(available_locales.values_list('id', flat=True)))
-        selected_ids = sorted(map(int, selected_ids.split(',')))
+        selected_ids = sorted(map(int, form.cleaned_data.get('selected_locales').split(',')))
```
jotes: Can we use a `models.CommaSeparatedIntegerField`?

mathjazz: That's deprecated, so I added `validate_comma_separated_integer_list`.

I'm not sure we can simplify getting data from the field, though.

mathjazz:
@jotes Another r?

pontoon/projects/views.py
jotes: I remember there's a function called 'split_ints' -> maybe we could add something similar.

mathjazz: Good catch! Updated.

jotes:
:+1: :dancer:

PR #576, opened 2017-04-14
Update en-US repo script to include central

mathjazz:
r=me

### DIFF
--- a/bin/mozilla-en-US.py
+++ b/bin/mozilla-en-US.py
@@ -115,10 +115,18 @@ abspath = os.path.abspath(__file__)
 dname = os.path.dirname(abspath)
 os.chdir(dname)
 
-for channel in ['aurora', 'beta']:
+for channel in ['beta', 'central']:
     for repo in SOURCE_REPOS:
         ending = repo + '-' + channel
-        url = 'ssh://hg.mozilla.org/releases/' + ending
+
+        url_folder = ''
+        if channel == 'central':
+            if repo == 'mozilla':
+                url_folder = 'l10n/'
+        else:
+            url_folder = 'releases/'
+
+        url = 'ssh://hg.mozilla.org/' + url_folder + ending
         target = os.path.join('source', ending)
 
         # Clone or update source repositories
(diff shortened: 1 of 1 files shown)


bin/mozilla-en-US.py
```diff
@@ -115,10 +115,18 @@ def push(path):
...
 os.chdir(dname)
 
-for channel in ['aurora', 'beta']:
+for channel in ['beta', 'central']:
     for repo in SOURCE_REPOS:
         ending = repo + '-' + channel
-        url = 'ssh://hg.mozilla.org/releases/' + ending
+
+        url_folder = ''
+        if channel == 'central':
```
flodolo: Is it just me or this would take stuff from https://hg.mozilla.org/mozilla-central? That's definitely not now what we want

mathjazz: It's OK actually (but the code is ugly, I agree).

```
[mhorvat@pontoon1.dev.webapp.scl3 comm-central]$ hg paths
default = ssh://hg.mozilla.org/comm-central

[mhorvat@pontoon1.dev.webapp.scl3 mozilla-central]$ hg paths
default = ssh://hg.mozilla.org/l10n/mozilla-central
```

flodolo: Ah, I completely missed the following lines of code where you add l10n/ to the url_folder for central

Since there are only two repos, I wonder if you need this much code instead of a simple mapping? Unless I'm missing other usages of this code

```python
urls = {
    'central': 'ssh://hg.mozilla.org/l10n/mozilla-central',
}

repository = urls.get(channel, 'ssh://hg.mozilla.org/releases/mozilla-{0}'.format(channel))

```

flodolo: Something like this should work though

```PYTHON
urls = {
    'mozilla-central': 'ssh://hg.mozilla.org/l10n/mozilla-central',
    'comm-central': 'ssh://hg.mozilla.org/comm-central',
}

repo_name = '{0}-{1}'.format(repo, channel)
url = urls.get(repo_name, 'ssh://hg.mozilla.org/releases/{0}'.format(repo_name))
target = os.path.join('source', '{0}'.format(repo_name))
```

mathjazz: I agree. This script will go away as soon as we land x-channel, so I'd rather avoid touching it at this stage.

flodolo: Good point.

PR #577, opened 2017-04-15
A few frontend bugfixes

mathjazz:
A few bug fixes I discovered while working on #575.

@jotes r?


pontoon/base/static/js/main.js
```diff
@@ -671,6 +672,11 @@ $(function() {
     } else {
       ul.find('.no-match').hide();
     }
+  }).on('keydown.search', '.menu input[type=search]', function(e) {
+    // Prevent form submission on Enter
+    if (e.which === 13) {
```
jotes: nit: I wonder if we could change those codes into something more readable for humans.

mathjazz: Yeah, a map or something would be nice.

PR #581, opened 2017-04-19
Make URLs case-sensitive

mathjazz:
We currently allow case-insensitive locale codes in URLs, but only on server side (and without any redirects). Which means some of our pages are broken:
https://pontoon.mozilla.org/bn-bd/              

And some are more broken:
https://pontoon.mozilla.org/bn-bd/pontoon-intro/

OTOH, we require project slugs to be case-sensitive.

Let's make all URL parameters case-sensitive, which menas we'll simply 404 for the examples above.

@jotes r?

### DIFF
--- a/pontoon/base/utils.py
+++ b/pontoon/base/utils.py
@@ -42,7 +42,7 @@ def get_project_locale_from_request(request, locales):
 
     for a in accept:
         try:
-            return locales.get(code__iexact=a[0]).code
+            return locales.get(code=a[0]).code
         except:
             continue
 
@@ -377,7 +377,7 @@ def get_download_content(slug, code, part):
     from pontoon.base.models import Entity, Locale, Project, Resource
 
     project = get_object_or_404(Project, slug=slug)
-    locale = get_object_or_404(Locale, code__iexact=code)
+    locale = get_object_or_404(Locale, code=code)
 
     # Download a ZIP of all files if project has > 1 and < 10 resources
     resources = Resource.objects.filter(project=project, translatedresources__locale=locale)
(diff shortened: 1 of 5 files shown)


jotes:
@mathjazz :+1:  I wonder if we could redirect users from the lower-cased locales into proper ones (?).

mathjazz:
With something like this in middleware?

```python
path = urlparse(url).path
chunks = path.split('/')

if (len(chunks)) > 1:
    candidate = chunks[1]
    locale = Locale.objects.filter(code__iexact=candidate)

    if locale.exists():
        chunks[1] = locale.code
        return redirect('/'.join(chunks))
```

jotes:
Maybe We could avoid a db query with this:
```
candidate = chunks[1].split('-')
if len(candidate) > 1 and not candidate[1].isupper():
   candidate[1] = candidate[1].upper()
   chunks[1] = '-'.join(candidate)
   return redirect('/'.join(chunks))
```

mathjazz:
Wouldn't that break '/heroku-setup/' and such?

jotes:
Okay, you're right, I suggested previous example because I misread an indexes.... The other solution could be to create a map and cache it (a hit to memory should be faster than any db lookup). Anyway if we don't worry about the performance that's a total nit and you have my :+1;

mathjazz:
We worry about performance. :)

Since lowercase locales break ATM and this patch displays a nice 404 page instead, I'm fine with landing the patch as is. And then we can monitor 404 pages (like we did for Aurora projects) and decide if adding a middleware is even neccessary.

PR #584, opened 2017-04-21
Fix TypeError when DOM storage is unavailable

StoyanDimitrov:
Fixes the `TypeError: localStorage is null` error when DOM storage is disabled (i.e. dom.storage.enabled is false).

### DIFF
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -2236,22 +2236,24 @@ var Pontoon = (function (my) {
      * Update all translations in localStorage on server
      */
     syncLocalStorageOnServer: function() {
-      if (localStorage.length !== 0) {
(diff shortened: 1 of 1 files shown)


pontoon/base/static/js/translate.js
```diff
@@ -2236,22 +2236,24 @@ var Pontoon = (function (my) {
...
 
-        // Clear all other translations
-        localStorage.clear();
+      var len = this.entities.length;
+      for (var i = 0; i < len; i++) {
+        var entity = this.entities[i],
+            key = this.getLocalStorageKey(entity),
+            value = localStorage[key];
+        if (value) {
+          value = JSON.parse(localStorage[key]);
```
jotes: :+1: - Me gusta -> another nested block bites the dust.

PR #588, opened 2017-04-24
Fix bug 1346180. Set database collation for search queries.

jotes:
Hi,
@mathjazz this should be able to fix search issues for turkish locale.
I don't set db_collation for turkish locale inside migration because some of Pontoon installations may not include turkish locale in their system/postgresql instance. That's why you'll have to set it on staging/production by hand (I suggest to use admin panel).


pontoon/base/admin.py
```diff
@@ -41,13 +43,39 @@ class ExternalLocaleResourceInline(ExternalResourceInline):
...
+class LocaleAdminForm(ModelForm):
+    """
+    Dynamically loads a list of available collations in your database.
+    """
+    db_collation = ChoiceField()
+
+    @property
+    def db_collations_choices(self):
+        """
+        Return all locales available in the current postgresql database.
```
mathjazz: You mean return all locale *collations*?

pontoon/base/models.py
```diff
@@ -469,6 +469,16 @@ def get_top_instances(self):
 
 class Locale(AggregatedStats):
     code = models.CharField(max_length=20, unique=True)
+    db_collation = models.CharField(
+        max_length=20,
+        blank=True,
+        help_text="""
+        Some of locales require to use different database collation than default ('en_US').
+
+        <strong>Use with caution, because it may brake the search for this locale.</strong>
```
mathjazz:
```Some locales require different database collation than default ('en_US').

<strong>Use with caution, because it may break search for this locale.</strong>
```

pontoon/db/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+)
+
+
+class IContainsCollate(IContains):
+    """
+    Searching through translations may produce invalid results if you don't specify a correct
+    database collation.
+    This filter require to pass a tuple with search phrase and collation to apply.
+    e.g.
+    translation__string__icontains_collate=('search_phrase', 'tr_tr')
```
mathjazz:
```
Searching for translations may produce invalid results if you don't specify a correct
database collation. This filter requires a tuple with search phrase and collation to apply.

E.g. translation__string__icontains_collate=('search_phrase', 'tr_tr')
```

pontoon/base/admin.py
jotes: Yup

jotes:
Hi,

I deployed a new instance of pontoon and I'm able to search dotless/dotted i:
https://aaaaaaaasaaaaaa.herokuapp.com/?search=nasil (but select 'TR' locale first and search for 'nasil')

`db_collation` is set to "tr_TR": 
Imgur

The list of collations can vary between instances of postgres, I'm not sure why you can see more of them at staging.
You can try to create them by hand (https://www.postgresql.org/docs/9.2/static/sql-createcollation.html):
```
CREATE COLLATION tr_TR (LOCALE = 'tr_TR.utf8');
```

mathjazz:
Thanks, confirming that it works:
https://aaaaaaaasaaaaaa.herokuapp.com/tr/pontoon-intro/all-resources/?search=nasil&string=8

I wonder if we can also highlight the match on frontend.

I tried searching for "i" and I found this among the results (`Tanıtımı`):
https://aaaaaaaasaaaaaa.herokuapp.com/tr/pontoon-intro/all-resources/?search=i&string=23

Is this expected?

jotes:
> I tried searching for "i" and I found this among the results (Tanıtımı):
I think it's caused by the "Pontoon intro" in the entity source.

> I wonder if we can also highlight the match on frontend.
I'm not sure if mark.js handles any kind of collation system... BUT we may be able to highlight query results in PostgreSQL -> https://www.postgresql.org/docs/9.1/static/textsearch-controls.html

pontoon/base/models.py
jotes: DAMN :D I thought I'm too old for the dyslexia. Yup, I meant "break".

mathjazz:
> I think it's caused by the "Pontoon intro" in the entity source.

Ah, you're right - I tried searching for `itm` and it yields no results.

> I'm not sure if mark.js handles any kind of collation system... BUT we may be able to highlight query results in PostgreSQL -> https://www.postgresql.org/docs/9.1/static/textsearch-controls.html

Unless this is trivial to implement, we should defer this to a later date.

mathjazz:
> I think it's caused by the "Pontoon intro" in the entity source.

Ah, you're right - I tried searching for `itm` and it yields no results.

> I'm not sure if mark.js handles any kind of collation system... BUT we may be able to highlight query results in PostgreSQL -> https://www.postgresql.org/docs/9.1/static/textsearch-controls.html

Unless this is trivial to implement, we should defer this to a later date.

mathjazz:
And we just got an r+ from Selim:
https://bugzilla.mozilla.org/show_bug.cgi?id=1346180#c6

mathjazz:
Once you fix the nits, I'll deploy this to stage again and test there.

jotes:
@mathjazz I fixed nits and added tests.

jotes:
@mathjazz ok, highlighting of results in PG requires fulltext search. If we can postpone it if that's not a blocker.

PR #589, opened 2017-04-25
Use script name in CSS selector instead of locale codes

mathjazz:
Just to make it more robust, we don't hardcode Arabic locale codes anymore.

@jotes r?

### DIFF
--- a/pontoon/contributors/templates/contributors/includes/timeline.html
+++ b/pontoon/contributors/templates/contributors/includes/timeline.html
@@ -23,7 +23,7 @@
         {% if event.translation %}
         {% set direction = event.translation.locale.direction %}
         <div class="fa fa-2x quote {% if direction == 'ltr' %}fa-quote-left{% else %}fa-quote-right{% endif %}"></div>
-        <p dir="{{ direction }}" lang="{{ event.translation.locale.code }}">{{ event.translation.string }}</p>
+        <p dir="{{ direction }}" lang="{{ event.translation.locale.code }}" data-script="{{ event.translation.locale.script }}">{{ event.translation.string }}</p>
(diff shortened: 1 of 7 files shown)


pontoon/contributors/templates/contributors/includes/timeline.html
```diff
@@ -23,7 +23,7 @@
         {% if event.translation %}
         {% set direction = event.translation.locale.direction %}
         <div class="fa fa-2x quote {% if direction == 'ltr' %}fa-quote-left{% else %}fa-quote-right{% endif %}"></div>
-        <p dir="{{ direction }}" lang="{{ event.translation.locale.code }}">{{ event.translation.string }}</p>
+        <p dir="{{ direction }}" lang="{{ event.translation.locale.code }}" data-script="{{ event.translation.locale.script }}">{{ event.translation.string }}</p>
```
jotes: I wonder if backend could prefetch this script - that could reduce a number of queries.

mathjazz: Good point! I guess we should prefetch locale for each translation - we already reference other locale fields in the code. I tried something, but don't see any impact on performance, so I'll skip this for now:

```
diff --git a/pontoon/base/models.py b/pontoon/base/models.py
index d2bbd31..35a430b 100644
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -300,6 +300,7 @@ def contributed_translations(self):
         Translation.objects.filter(user=self)
             .exclude(string=F('entity__string'))
             .exclude(string=F('entity__string_plural'))
+            .prefetch_related('locale')
     )
     return translations
```

PR #594, opened 2017-05-02
Fix bug 1346150 Fix for the regression introduced by the previous cha…

jotes:
@mathjazz could you...:bowing_man:


mathjazz:
That works for the deadline column, but still doesn't fix the issue for the latest activity if it contains empty values (`-`).

a-fullpage

PR #596, opened 2017-05-03
Fix bug 1354812: Do not copy from helpers on Ctrl + Tab

mathjazz:
@jotes r?

I checked other keyboard shortcuts in Firefox with the `Tab` key and they are all combinations with `Ctrl`.


jotes:
R+

PR #601, opened 2017-05-05
Latest activity

mathjazz:
We don't set latest activity on batch editing actions and changes made via file upload.

We also don't show latest activity data for projects with subpages.

This patch fixes all of that.

@jotes r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -2025,29 +2025,30 @@ class Translation(DirtyFieldsMixin, models.Model):
             if self.approved:
                 self.entity.mark_changed(self.locale)
 
-            # Check and update the latest translation where necessary.
-            self.check_latest_translation(self.entity.resource.project)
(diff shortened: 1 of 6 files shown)


pontoon/base/models.py
```diff
@@ -1887,8 +1887,8 @@ def find_and_replace(self, find, replace, user):
         translations.update(approved=False, approved_user=None, approved_date=None)
 
         # Create new translations
-        Translation.objects.bulk_create(translations_to_create)
-        return translations
+        new_translations = Translation.objects.bulk_create(translations_to_create)
+        return translations, new_translations[-1].pk
```
jotes: I think we could add a docstring to explain what's returned here.

mathjazz: Good point, will add in #603 (which changes this line) to avoid conflict.

PR #604, opened 2017-05-09
Fix bug 1363320: Properly compose tab URLs

mathjazz:
Fixes Turkish contributors tab.

And another bug I found along the way (JS error if bug list empty for a locale).

@jotes r?


jotes:
```
path = window.location.pathname + window.location.search
"/tr/contributors/"
'/' + path.split('/' + 'tr' + '/')[1]
"/contributors/"
```
Works like a charm!

mathjazz:
Thanks for the test case. :P

PR #605, opened 2017-05-10
Fix bug 1362929: Fix contributor timeline loading

mathjazz:
If the last child is positioned absolutely, the bottom of it can get cut off the viewport. That prevents the next batch of timeline items to load.

@jotes r?


jotes:
:+1: :tada: :taco:

PR #607, opened 2017-05-11
Fix bug 1349330: Prevent too long translation submissions

mathjazz:
* Use string comments in .lang files to determine translation length limit
* If length limited, show character countdown
* If limit exceeded, prevent translation from being submitted
* Strip HTML tags when calculating string length
* Some code cleanups

@jotes r?

This fix is entirely frontend based, as we already mangle with comments in JS to present links and screenshots. We should file a bug to store this data in a more structured fashion on import.

### DIFF
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -496,7 +496,7 @@ var Pontoon = (function (my) {
             splitComment = entity.comment;
         if (split[0].startsWith('MAX_LENGTH')) {
           try {
-            self.translationLengthLimit = parseInt(split[0].split('MAX_LENGTH: ')[1].split(' ')[0]);
+            self.translationLengthLimit = parseInt(split[0].split('MAX_LENGTH: ')[1].split(' ')[0], 10);
             splitComment = split.length > 1 ? entity.comment.substring(entity.comment.indexOf('\n') + 1) : '';
           } catch (e) {} // Catch unexpected comment structure
         }
@@ -1574,9 +1574,6 @@ var Pontoon = (function (my) {
           entity = self.getEditorEntity(),
           translation = $('#translation').val();
 
-      // Prevent double translation submissions
-      $(this).off('click.save');
-
       // Prevent empty translation submissions if not supported
       if (translation === '' &&
         ['properties', 'ini', 'dtd', 'ftl'].indexOf(entity.format) === -1) {
@@ -1589,6 +1586,9 @@ var Pontoon = (function (my) {
         self.endLoader('Translation too long.', 'error');
         return;
       }
+
+      // Prevent double translation submissions
+      $(this).off('click.save');
 
       self.updateOnServer(entity, translation, true);
     },
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -429,7 +429,21 @@ var Pontoon = (function (my) {
      * Update current translation length
      */
     updateCurrentTranslationLength: function () {
-      $('#translation-length .current-length').html($('#translation').val().length);
+      var limit = this.translationLengthLimit,
+          translation = $('#translation').val();
+
+      if (limit) {
+        var length = this.stripHTML(translation).length,
+            charactersLeft = limit - length;
+
+        $('#translation-length .characters-left')
+          .toggleClass('overflow', charactersLeft < 0)
+          .html(charactersLeft);
+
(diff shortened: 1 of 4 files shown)


pontoon/base/static/js/translate.js
```diff
@@ -469,18 +483,31 @@ var Pontoon = (function (my) {
...
-        var comment = this.linkify(entity.comment);
-        if (comment === entity.comment) {
-          comment = this.doNotRender(entity.comment);
+
+        // Translation length limit
+        var split = entity.comment.split('\n'),
+            splitComment = entity.comment;
+        if (split[0].startsWith('MAX_LENGTH')) {
+          try {
+            self.translationLengthLimit = parseInt(split[0].split('MAX_LENGTH: ')[1].split(' ')[0]);
```
jotes: `parseInt(number, 10)`

pontoon/base/static/js/translate.js
```diff
@@ -469,18 +483,31 @@ var Pontoon = (function (my) {
...
       $('#metadata').empty();
       if (entity.comment) {
-        var comment = this.linkify(entity.comment);
-        if (comment === entity.comment) {
-          comment = this.doNotRender(entity.comment);
+
+        // Translation length limit
+        var split = entity.comment.split('\n'),
+            splitComment = entity.comment;
+        if (split[0].startsWith('MAX_LENGTH')) {
```
jotes: What if `MAX_LENGTH` is in the second line?

jotes:
@mathjazz 
There's a possible bug during validation:
1. I typed a string with to cause "overflow".
2. I saw validation message: "Translation too long."
3. I fixed the length of the string
4. "suggest" button doesn't work, validation error doesn't display even if I type a too long translation again.

pontoon/base/static/js/translate.js
```diff
@@ -429,7 +429,21 @@ var Pontoon = (function (my) {
      * Update current translation length
      */
     updateCurrentTranslationLength: function () {
-      $('#translation-length .current-length').html($('#translation').val().length);
+      var limit = this.translationLengthLimit,
+          translation = $('#translation').val();
+
+      if (limit) {
+        var length = this.stripHTML(translation).length,
```
jotes: I started to wonder how should We handle various placeables like e.g. `%s` or `{0}` because they can change the length of result string.

jotes:
Besides that bug, everything looks to work okay, r+wc

pontoon/base/static/js/translate.js
mathjazz: We always put tags above comments, I checked all examples.

But you're right, tags are allowed to be positioned after the comment. We should catch this case when moving the logic to backend.

pontoon/base/static/js/translate.js
mathjazz: Good point.

pontoon/base/static/js/translate.js
```diff
@@ -469,18 +483,31 @@ var Pontoon = (function (my) {
...
-          comment = this.doNotRender(entity.comment);
+
+        // Translation length limit
+        var split = entity.comment.split('\n'),
+            splitComment = entity.comment;
+        if (split[0].startsWith('MAX_LENGTH')) {
+          try {
+            self.translationLengthLimit = parseInt(split[0].split('MAX_LENGTH: ')[1].split(' ')[0]);
+            splitComment = split.length > 1 ? entity.comment.substring(entity.comment.indexOf('\n') + 1) : '';
+          } catch (e) {} // Catch unexpected comment structure
```
mathjazz: Same as https://github.com/mozilla/pontoon/pull/607#discussion_r116114483.

pontoon/base/static/js/translate.js
mathjazz: [The spec](https://github.com/mozilla-l10n/langchecker/wiki/.lang-files-format#string-meta-data) doesn't specify anything.

BTW, this is Mozilla's format, probably nobody else uses it, so it's a little easier to make some asumptions. :P

mathjazz:
Updated.

pontoon/base/static/js/translate.js
jotes: Haha, a totally good point :-)
I just thought that MAX_LENGTH is a protection mechanism against strings that won't fit in the viewport or something like that.

jotes:
Ok, looks good r+

PR #610, opened 2017-05-19
Fix bug 1366164: Prevent too long translations on approval

mathjazz:
@jotes r?

### DIFF
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -1566,6 +1566,14 @@ var Pontoon = (function (my) {
 
 
     /*
+     * Check if translation length limit exceeded
+     */
+    translationLengthLimitExceeded: function (translation) {
+      return this.translationLengthLimit && this.stripHTML(translation).length > this.translationLengthLimit;
+    },
+
+
+    /*
      * Submit translation to DB
      */
     saveTranslation: function (e) {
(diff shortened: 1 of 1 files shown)


pontoon/base/static/js/translate.js
```diff
@@ -1566,6 +1566,19 @@ var Pontoon = (function (my) {
 
 
     /*
+     * Prevent too long translation submissions
+     */
+    translationLengthLimitExceeded: function (translation) {
+      if (this.translationLengthLimit && this.stripHTML(translation).length > this.translationLengthLimit) {
+        this.endLoader('Translation too long.', 'error');
```
jotes: `is too long?` - correct me if I'm wrong.

jotes: Also, I'd move a side-effect of this function outside and leave a simple:
`return this.translationLengthLimit && this.stripHTML(translation).length > this.translationLengthLimit`

jotes: But I see what you did here -> You want to avoid duplication, but still leaving side-effect in a method that should return a simple boolean, can be a little misleading.  Anyway, as there's no clear solution - I trust your judgment.

mathjazz: We use "Translation deleted", "Translation saved", "Translation approved", so I think it's OK to use "Translation too long" and follow this pattern.

I agree about the second part, let's remove the notifications. Besides, there are other pieces that are already duplicated the calling the function (if clause and return).

PR #613, opened 2017-05-29
Fix bug 1367005: Highlight more special characters in translations

mathjazz:
@jotes r?

TODO: At some point we should also merge `pontoon.base.utils:mark_placeables()` logic (handling source string placeables) into `Pontoon.markPlaceables()` (handling translation placeables).

### DIFF
--- a/pontoon/base/static/js/main.js
+++ b/pontoon/base/static/js/main.js
@@ -139,16 +139,44 @@ var Pontoon = (function (my) {
      * Markup placeables
      */
     markPlaceables: function (string) {
-      function markup(string, regex, title) {
-        return string.replace(regex, '<mark class="placeable" title="' + title + '">$&</mark>');
+      function getReplacement(title, replacement) {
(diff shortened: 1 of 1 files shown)


pontoon/base/static/js/main.js
```diff
@@ -139,17 +139,45 @@ var Pontoon = (function (my) {
...
     markPlaceables: function (string) {
-      function markup(string, regex, title) {
-        return string.replace(regex, '<mark class="placeable" title="' + title + '">$&</mark>');
+      function getReplacement(title, replacement) {
+        return '<mark class="placeable" title="' + title + '">' + replacement + '</mark>';
+      }
+
+      function markup(string, regex, title, replacement) {
+        replacement = replacement || '$&';
+        return string.replace(regex, getReplacement(title, replacement));
```
jotes: I wonder if We could just use mark.js here?

mathjazz: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1368602.

PR #614, opened 2017-05-31
Add command for installation all the prerequisites if user is on Ubuntu 17.04.

Facyber:
As Ubuntu 17.04 use Postgresql 9.6, the previous command won't work so I added new and updated command.


jotes:
:+1: :tada: Welcome @Facyber! :beers:

PR #616, opened 2017-06-02
[WiP] Move document field to Entity model

mathjazz:
This fixes the missing string search bug without negative impact on performance. Confirmed on Heroku.

TODO:
- [x] Create `document` field in the `Entity` model.
- [x] Run data migration to populate `document` field for existing `Entity` instances.
- [x] Set `document` field upon `Entity` creation.
- [x] Remove `entity_document_update_trigger`.
- [ ] `MultiFieldTRGMIndex` replacement?

@jotes r?

2 questions:

1. Should I simply add something like [this](https://github.com/mozilla/pontoon/commit/63de630c4780e6b9b8162e4f49034c96a65a7af7#diff-6a8132d5898e35931ed3541f5e51474aR39) to the migration to remove the `entity_document_update_trigger`?

2. I'm a bit lost on how to change `MultiFieldTRGMIndex` now that we search across two tables.


jotes:
hey,
ad 1. Yup, triggers should be dropped if they're not needed anymore.
ad 2. Actually, you could use MultiFieldTRGMIndex twice, for base_entities and then the next entry for base_translation.

mathjazz:
@jotes Thanks.

I copied `entity_document_update_trigger_drop_sql` and `entity_document_update_trigger_create_sql` from migration 0091. Is that OK?

Could you check if I defined `MultiFieldTRGMIndex` properly? It's going to take some time to apply this.

I don't need to `CreateExtension('btree_gin')` and `TrigramExtension()` again (given that they are already present in 0091), right?

pontoon/base/migrations/0094_entity_document.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def set_entity_document(apps, schema_editor):
+    Entity = apps.get_model('base', 'Entity')
+    Entity.objects.update(
+        document=Concat('key', Value(' '), 'string', Value(' '), 'string_plural', Value(' '), 'comment')
+    )
+
+entity_document_update_trigger_create_sql = """
+    CREATE FUNCTION base_translation_entity_document_update() RETURNS TRIGGER AS $$
```
jotes: I'd change the name of this function, let's call it `base-entity_document_update`.

jotes:
ad 1. Looks okay
ad 2. Definitions of indexes look okay, but I'm not sure if query planner will pick them up -> Did you look at the performance of search after those changes?
ad 3. You don't need to create btree_gin & trgm extensions again.

mathjazz:
I ran the [ActiveEntityTranslation experiment](https://gist.github.com/mathjazz/825ecd1544a34fca7576b747eb96bb35) and it speeds up loading, searching and filtering of strings dramatically. Locally, all .operations completed in under 2 seconds, mostly in less than 1 second. For some operations that means over 90% speedup.

This approach also opens the door for landing #520.

And, makes this PR obsolete, so I'm closing it.

PR #617, opened 2017-06-05
Bug 1344658. Add indexes to speed-up calculation of stats.

jotes:
@mathjazz hey... I haven't a much time today :( r?


Pike:
Can those be in the models.py, and generate the migration instead?

https://docs.djangoproject.com/en/1.10/ref/models/options/#django.db.models.Options.index_together seems like it?

jotes:
Hey, you're right. Will do in a moment.

6 cze 2017 9:00 AM "Axel Hecht" <notifications@github.com> napisał(a):

Can those be in the models.py, and generate the migration instead?

https://docs.djangoproject.com/en/1.10/ref/models/options/#django.db.models.
Options.index_together seems like it?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/mozilla/pontoon/pull/617#issuecomment-306399631>, or mute
the thread

jotes:
@flod could you look again?

Pike:
> @flod could you look again?

Let's pretend that I'm flod.

On top of the models change, I expect a migration part in this patch via ./manage.py makemigrations

jotes:
@pike I'm so sorry, I just read about an other bug and I mistyped nicknames. Yup, for some reason Pycharm didn't add that migration and I thought that Github needs a moment to sync the data. Anyway, fixed.

Pike:
Yes, this looks good, I'd just recommend to use a dedicated name for the migration. That is, instead of 0094_auto_20170606_1931.py use something like 0094_bug_1344658_save_delete_perf.py or so

jotes:
@Pike I changed the name of migration, could you look if that's okay?

Pike:
lgtm, @mathjazz ?

mathjazz:
A thing of beauty!

Thanks to both of you!

PR #618, opened 2017-06-06
Fix bug 1366259: In latest activity tooltip, beautify FTL translations

mathjazz:
@jotes r?

### DIFF
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -348,7 +348,7 @@ var Pontoon = (function (my) {
 $(function() {
 
   // Ignore editing for anonymous users
-  if (!Pontoon.user.id) {
+  if (Pontoon && Pontoon.user && !Pontoon.user.id) {
     return;
   }
 
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -348,7 +348,7 @@ var Pontoon = (function (my) {
 $(function() {
 
   // Ignore editing for anonymous users
-  if (Pontoon && Pontoon.user && !Pontoon.user.id) {
+  if (!(Pontoon && Pontoon.user && Pontoon.user.id)) {
     return;
   }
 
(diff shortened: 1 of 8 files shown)


pontoon/base/static/js/fluent_interface.js
```diff
@@ -348,7 +348,7 @@ var Pontoon = (function (my) {
 $(function() {
 
   // Ignore editing for anonymous users
-  if (!Pontoon.user.id) {
+  if (Pontoon && Pontoon.user && !Pontoon.user.id) {
```
mathjazz: The `Pontoon` object is only available in translate view. Now that we run this code elsewhere, we need an additional check.

pontoon/base/static/js/table.js
```diff
@@ -23,8 +23,12 @@ var date_formatter = new Intl.DateTimeFormat('en-GB', {
 $('body').on('mouseenter', '.latest-activity .latest time', function() {
   var $element = $(this);
 
+  // Translation Preview (needed for FTL)
+  var translationData = $element.data('translation'),
+      translationPreview = Pontoon.fluent.getSimplePreview({string: translationData}, translationData, {format: 'ftl'});
+
```
mathjazz: Retrieving translation format is expensive ATM, so we assume all translations are in the FTL format. That way `getSimplePreview` will return simplified previews of FTL translations and fallback to the default presentation of all other translations.

pontoon/teams/templates/teams/team.html
```diff
@@ -128,5 +128,7 @@
 {% endblock %}
 
 {% block extend_js %}
+  <!-- TODO: integrate into django-pipeline -->
```
jotes: Did you consider Webpack?

pontoon/base/static/js/fluent_interface.js
adngdb: This should be ``if (!(Pontoon && Pontoon.user && Pontoon.user.id)) {``.

adngdb:
@mathjazz Friendly reminder that this is r+'ed. :) Do you want to merge it? Is it blocked on anything?

mathjazz:
I know, sorry... :) This is really ugly. In particular, I'd like to avoid enabling the FTL scripts on so many pages just to run this simple function.

I'm working on bug 1362072 now, as part of which I'll try to come up with a backend solution for rendering FTL strings, which will be useful in the translate view as well.

Closing.

PR #619, opened 2017-06-07
Fix bug 1344658: optimize authors and time range

mathjazz:
Instead of updating a list of authors and the time range in the filter menu on any event that can change them, we do it when the filter menu is opened.

This has the biggest impact on saving and deleting translations. Together with 242de20c42663d8a0476cb1081deaae4df3b4312, those two tasks should be at least 80% faster.

It also speeds up loading of the translate page and entities.

@jotes r?

### DIFF
--- a/pontoon/base/static/css/translate.css
+++ b/pontoon/base/static/css/translate.css
@@ -685,7 +685,13 @@ body > header aside p {
   padding-right: 5px;
 }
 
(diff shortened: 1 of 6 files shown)


jotes:
@mathjazz I had similar idea: I wanted to move that counts to a separate view and rewrite authors filter for progressive loading.

mathjazz:
Great! Leaving the bug open.

PR #620, opened 2017-06-07
Use prefetched ProjectLocales on team/project dashboards

mathjazz:
It should speed up Project and Team dashboard default tabs by some 20 %.

@jotes r?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -440,9 +440,9 @@ class LocaleQuerySet(models.QuerySet):
         """
         return self.filter(translatedresources__isnull=False).distinct()
 
-    def prefetch_latest_translation(self, project):
+    def prefetch_project_locale(self, project):
         """
-        Prefetch latest translation data for given project.
+        Prefetch ProjectLocale and latest translation data for given project.
         """
         return self.prefetch_related(
             Prefetch(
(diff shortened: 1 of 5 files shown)


pontoon/base/models.py
```diff
@@ -1169,11 +1169,9 @@ def get_latest_activity(cls, self, extra=None):
         """
         latest_translation = None
 
-        if hasattr(self, 'fetched_latest_translation'):
-            if self.fetched_latest_translation:
-                latest_translation = self.fetched_latest_translation[0].latest_translation
-            else:
-                latest_translation = None
+        if hasattr(self, 'fetched_project_locale'):
```
jotes: `if getattr(self, 'fetched_project_locale'):` and remove the second if?

pontoon/teams/templates/teams/includes/projects.html
```diff
@@ -16,12 +16,14 @@
 
   {{ ProjectList.header(request=True) }}
 
+  {% set locale_projects = locale.available_projects_list() %}
+
```
mathjazz: We already make this request in the main template, so we could also move the logic for showing and hiding enabled project on first load to JS. But, that wouldn't have any noticable perf impact. And would be a much bigger codebase change. So I'm keeping it as is for now.

pontoon/base/models.py
mathjazz: Thanks!

I ended up using this: `if getattr(self, 'fetched_project_locale', None):`

mathjazz:
@jotes r?

I found another (bigger) optimization for the Team and Project dashboard. Check out the second commit. Together, both commits should bring around 70% faster loading [Team](localhost:8000/pl/) and [Project](localhost:8000/projects/pontoon-intro/) dashboards.

jotes:
R+, looks good

PR #621, opened 2017-06-08
Bug 1310668: Recalculate stats when sync completes

mathjazz:
This is a temporary fix for bug 1310668. We recalculate any broken stats when the sync task completes.

@jotes r?

### DIFF
--- a/pontoon/sync/models.py
+++ b/pontoon/sync/models.py
@@ -1,11 +1,18 @@
+import logging
+
+from bulk_update.helper import bulk_update
+
 from django.core.urlresolvers import reverse
 from django.db import models
-from django.db.models import Max
+from django.db.models import F, Max, Sum
 from django.utils import timezone
 from django.utils.functional import cached_property
 
-from pontoon.base.models import Project, Repository
+from pontoon.base.models import Project, ProjectLocale, Repository, TranslatedResource
 from pontoon.base.utils import latest_datetime
+
+
+log = logging.getLogger(__name__)
 
 
 class BaseLog(models.Model):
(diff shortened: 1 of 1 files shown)


pontoon/sync/models.py
```diff
@@ -45,6 +50,37 @@ def finished(self):
...
+        temporary fix for bug 1310668.
+        """
+
+        if not self.finished:
+            return
+
+        # total_strings missmatch between TranslatedResource & Resource
+        for t in TranslatedResource.objects.exclude(total_strings=F('resource__total_strings')):
+            t.total_strings = t.resource.total_strings
+            t.save()
```
Pike: Can we put the F into an update expression?

If not, we should at least select_related('resource')?

mathjazz: `F()` is allowed in `update()`, but [not with joined fields](https://code.djangoproject.com/ticket/14104). :/

I've ammended the commit.

pontoon/sync/models.py
```diff
@@ -45,6 +50,36 @@ def finished(self):
...
+        Recalculate any broken stats when sync task is finished. This is a
+        temporary fix for bug 1310668.
+        """
+        if not self.finished:
+            return
+
+        # total_strings missmatch between TranslatedResource & Resource
+        for t in TranslatedResource.objects.exclude(total_strings=F('resource__total_strings')).select_related('resource'):
+            t.total_strings = t.resource.total_strings
+            t.save()
```
jotes: I wonder if bulk_update wouldn't be faster.

pontoon/sync/models.py
```diff
@@ -45,6 +50,36 @@ def finished(self):
...
+
+        # total_strings missmatch in ProjectLocales within the same project
+        for p in Project.objects.available():
+          if ProjectLocale.objects.filter(project=p).values("total_strings").distinct().count() > 1:
```
jotes: I see there are broken indentations.

pontoon/sync/models.py
```diff
@@ -45,6 +50,36 @@ def finished(self):
...
+            for pl in ProjectLocale.objects.filter(project=p):
```
jotes: I think It's possible to change those queries to a single aggregation.

pontoon/sync/models.py
mathjazz: OK, now that you both pointed it out, I give up. :)

pontoon/sync/models.py
mathjazz: On a quick look I'm not sure how. :)

Since this is only temporary (sic!) and doesn't perform slowly, I'm keeping it unchanged.

PR #623, opened 2017-06-28
Fixes bug 1376813 - THIS... IS... DOCKER!

adngdb:
This is not fully ready yet, as I think it is missing some important features. Those are: 

- [x] add documentation
- [x] add commands to run the tests
- [x] add commands to easily import a database dump
- [x] update documentation about building docs (as the requirements now need to be installed separately)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,25 @@
+FROM local/pontoon_base
+
+WORKDIR /app
+
+# Install OS-level things
+COPY docker/set_up_ubuntu.sh /app/
+RUN DEBIAN_FRONTEND=noninteractive /app/set_up_ubuntu.sh
+
+COPY package.json /app/
+RUN cd /app && npm install -g
+
+# Install Socorro Python requirements
+COPY requirements.txt /app/requirements.txt
+COPY requirements-dev.txt /app/requirements-dev.txt
+RUN pip install -U 'pip>=8'
+RUN pip install --no-cache-dir -r requirements-dev.txt
+
+COPY . /app/
+COPY ./docker/config/webapp.env /app/.env
+
+ENV PYTHONUNBUFFERED 1
+ENV PYTHONDONTWRITEBYTECODE 1
+ENV PYTHONPATH /app
+
+CMD ["/app/docker/run_webapp.sh"]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -6,14 +6,11 @@ WORKDIR /app
 COPY docker/set_up_ubuntu.sh /app/
 RUN DEBIAN_FRONTEND=noninteractive /app/set_up_ubuntu.sh
 
-COPY package.json /app/
-RUN cd /app && npm install -g
-
 # Install Socorro Python requirements
 COPY requirements.txt /app/requirements.txt
 COPY requirements-dev.txt /app/requirements-dev.txt
 RUN pip install -U 'pip>=8'
-RUN pip install --no-cache-dir -r requirements-dev.txt
+RUN pip install --no-cache-dir --require-hashes -r requirements-dev.txt
 
 COPY . /app/
 COPY ./docker/config/webapp.env /app/.env
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -4,7 +4,7 @@
 # Either use the LOCAL_USER_ID if passed in at runtime or
 # fallback
 
-USER_ID=${LOCAL_USER_ID:-9001}
+USER_ID=${LOCAL_USER_ID:-10001}
 
 echo "Starting with UID : $USER_ID"
 useradd --shell /bin/bash -u $USER_ID -o -c "" -m app
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+# Add local user
+# Either use the LOCAL_USER_ID if passed in at runtime or
+# fallback
+
+USER_ID=${LOCAL_USER_ID:-9001}
+
+echo "Starting with UID : $USER_ID"
+useradd --shell /bin/bash -u $USER_ID -o -c "" -m app
+export HOME=/home/app
+
+exec /usr/local/bin/gosu app "$@"
--- a/docker/run_tests_in_docker.sh
+++ b/docker/run_tests_in_docker.sh
@@ -1,0 +1,72 @@
+#!/bin/bash
+
+# Script that sets up the docker environment to run the tests in and runs the
+# tests.
+
+# Pass --shell to run a shell in the test container.
+
+# Failures should cause setup to fail.
+set -v -e -x
+
+DC="$(which docker-compose)"
+
+# Use the same image we use for building docker images because it'll be cached
+# already.
+BASEIMAGENAME="local/pontoon_base"
+
+# Start services in background (this is idempotent).
+echo "Starting services in the background..."
+${DC} up -d postgresql
+
+# If we're running a shell, then we start up a test container with . mounted
+# to /app.
+if [ "$1" == "--shell" ]; then
+    echo "Running shell..."
+
+    docker run \
+           --rm \
+           --volume "$(pwd)":/app \
+           --workdir /app \
+           --network pontoon_default \
+           --link pontoon_postgresql_1 \
+           --env-file ./docker/config/webapp.env \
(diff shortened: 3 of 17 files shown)


adngdb:
cc @Pike, if you want to test this and tell me how it goes for you.

docker/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+RUN DEBIAN_FRONTEND=noninteractive /app/set_up_ubuntu.sh
+
+COPY package.json /app/
+RUN cd /app && npm install -g
+
+# Install Socorro Python requirements
+COPY requirements.txt /app/requirements.txt
+COPY requirements-dev.txt /app/requirements-dev.txt
+RUN pip install -U 'pip>=8'
+RUN pip install --no-cache-dir -r requirements-dev.txt
```
jotes: Can We pass `--require-hashes` here?

docker/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM local/pontoon_base
+
+WORKDIR /app
+
+# Install OS-level things
+COPY docker/set_up_ubuntu.sh /app/
+RUN DEBIAN_FRONTEND=noninteractive /app/set_up_ubuntu.sh
+
+COPY package.json /app/
+RUN cd /app && npm install -g
```
Pike: Probably doesn't want -g here

docker/entrypoint.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+# Add local user
+# Either use the LOCAL_USER_ID if passed in at runtime or
+# fallback
+
+USER_ID=${LOCAL_USER_ID:-9001}
+
+echo "Starting with UID : $USER_ID"
+useradd --shell /bin/bash -u $USER_ID -o -c "" -m app
```
Pike: The warning is surprising to folks, I guess.

It'd be nice if we could only do this if required.

Pike:
Importing a db dump has been hard and kinda unsuccessful so far for me. It'd be nice if we had an entrypoint in the makefile to make that smooth.

requirements.txt
```diff
@@ -2,8 +2,6 @@
 # readthedocs.org don't support pip8.
 # More info at https://github.com/rtfd/readthedocs.org/issues/1843
 
--r docs/requirements.txt
-
```
adngdb: The prod ``requirements.txt`` file should not be needed for rtd to build the docs. According to the configuration in rtd, it's only been using ``docs/requirements.txt`` anyway, and I guess it's been working fine so far. :)

adngdb:
I believe this is now fully working, at least on the docker side (I still need to update the documentation about building the docs). 

@Pike and co, do you care to retry installing Pontoon with docker from scratch and see if that works for you? Also, do you think anything is missing from the current documentation?

adngdb:
This is ready now. r?

docker/run_webapp.sh
```diff
@@ -0,0 +1,21 @@
...
+
+echo "Install latest npm modules"
+cd /app
+npm install
+
+echo "Setting up the db for Django"
+python manage.py migrate
+
+echo "Collect static assets"
+python manage.py collectstatic --noinput
```
Pike: Do we need this for the DEBUG container?

docker/run_webapp.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# Note: this should be a one-off, but if that project is already synced,
+# it will simply not do anything and won't complain about it.
+echo "Running migration of the pontoon-intro module"
+python manage.py sync_projects --projects=pontoon-intro --no-commit
```
Pike: Should this move into set_up_webapp.sh, given the comment?

docs/dev/install-docker.rst
```diff
@@ -0,0 +1,91 @@
...
+Prerequisites
+-------------
+
+1. Install `Docker <https://docs.docker.com/engine/installation/>`_.
+
+2. Install `docker-compose <https://docs.docker.com/compose/install/>`_. You need
+   1.10 or higher.
+
+3. Install `make <https://www.gnu.org/software/make/>`_ using either your
+   system's package manager (Linux) or homebrew (OSX) or FIXME (Windows).
```
Pike: Should we recommend https://wiki.mozilla.org/MozillaBuild for windows?

Or, alternatively, we could not use make by maybe something written in python? Other projects at mozilla have a d0cker script with a numeric zero, to encapsulate stuff.

docs/dev/install-docker.rst
```diff
@@ -0,0 +1,91 @@
...
+   webapp.
+
+2. Then run the webapp::
+
+      $ make dockerrun
+
+   .. Note::
+
+        The first time you run this, the PostgreSQL container needs to do
+        some work before it becomes available to the webapp container. Hence,
```
Pike: I'm not really happy with healthchecks in docker, but if we'd add one for the postgresql container, the webapp would only start once that healthcheck succeeds.

The caveat is that the healthchecks only run at one single frequency, and thus you're either pounding your container with checks, or you wait for a while. As the first healthcheck is done after one cycle period.

docs/dev/install-docker.rst
```diff
@@ -0,0 +1,91 @@
...
+
+At that point, you're good to go. Access the webapp via this URL: http://localhost:8000/
+
+Database
+--------
+
+By default, you will have default data loaded for only the Pontoon Intro project.
+If you have a database dump, you can load it into your PostgreSQL database by running::
+
+    $ make dockerloaddb DB_DUMP_FILE=path/to/my/dump
```
Pike: I learned that folks with access to heroku can get a DB dump from there. Is that OK to add to the docs?

docker/run_tests_in_docker.sh
```diff
@@ -0,0 +1,72 @@
...
+           --tty \
+           --interactive \
+           local/pontoon /bin/bash
+
+else
+    # Create a data container to hold the repo directory contents and copy the
+    # contents into it.
+    if [ "$(docker container ls --all | grep pontoon-tests)" != "" ]; then
+        echo "Removing previously existing container"
+        docker rm pontoon-tests
```
adngdb: This is the easiest way to ensure that the tests run without old content. For example, in cases when we remove files from the repo, if we don't make sure we delete the entire content and re-copy it, then the tests will still run the removed files. That hit us on socorro, and remove the container is the fastest and simplest way to circumvent this.

docker/run_webapp.sh
adngdb: :+1:

docs/dev/install-docker.rst
adngdb: That sounds like Mozilla-specific documentation, which should probably go into a mana page for example. If you think it's important we add it here, I'm fine with it, but I wanted to express that concern first. :)

docs/dev/install-docker.rst
adngdb: This will happen only the first time you create the postgres container, and hopefully that shouldn't happen too often. I think this is a bit too much work for the benefits. Maybe we keep that for later if it turns out to be a pain to people?

adngdb:
I've rebased the branch, solved the conflict, and hopefully addressed all your concerns. :)

adngdb:
@Pike can I go ahead and merge this? I believe it is in a good-enough state, and we can fix issues as they appear in follow-up PRs.

docs/dev/install-docker.rst
```diff
@@ -0,0 +1,92 @@
...
+
+To run the entire test suite, simply run::
+
+    $ make dockertest
+
+If you want to run only some unit tests, or want to avoid rebuilding the
+docker container every time you run tests, you can start a shell that will
+allow to run your own commands. To do that, run::
+
+    $ make dockertestshell
```
Pike: we renamed this ;-)

mathjazz:
Welcome to Pontoon, @adngdb.

jotes:
:tada:

docker/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM local/pontoon_base
+
+WORKDIR /app
+
+# Install OS-level things
+COPY docker/set_up_ubuntu.sh /app/
+RUN DEBIAN_FRONTEND=noninteractive /app/set_up_ubuntu.sh
+
+# Install Socorro Python requirements
```
flodolo: Leftover from another script?

PR #625, opened 2017-07-03
Adjust @require_AJAX decorator for New Relic use

mathjazz:
[Without this change](https://docs.newrelic.com/docs/agents/python-agent/supported-features/python-tips-tricks#decorators_and_introspection), New Relic groups all functions using @require_AJAX decorator into a single transaction, which makes it impossible to see how each of them performs. Most if not all AJAX views use this decorator (17 at the moment).

### DIFF
--- a/pontoon/base/utils.py
+++ b/pontoon/base/utils.py
@@ -1,4 +1,5 @@
 import codecs
+import functools
 import json
 import logging
 import os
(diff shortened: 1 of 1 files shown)


mathjazz:
@jotes r?

pontoon/base/utils.py
```diff
@@ -332,6 +333,7 @@ def require_AJAX(f):
     """
     AJAX request required decorator
     """
+    @functools.wraps(f)
```
jotes: Awww... I wonder if there's a some kind of pre-caution we should take to avoid similar bugs in the future?

mathjazz: I doubt it. Well, we'll see in New Relic if any of the important transactions is missing.

PR #628, opened 2017-07-10
Bug 1377830: Remove duplicates from array without using ES6

mathjazz:
The current implementation makes `django-pipeline` throw an error on compression.

@jotes r?


jotes:
:( :+1:

PR #629, opened 2017-07-13
On translation delete, remove corresponding TM entry

mathjazz:
@jotes r?

Fixing bug reported by @jordis.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -2082,7 +2082,10 @@ class Translation(DirtyFieldsMixin, models.Model):
         self.entity.mark_changed(self.locale)
 
     def delete(self, stats=True, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


jotes:
https://github.com/mozilla/pontoon/blob/master/pontoon/base/tests/test_models.py#L1093

I'm not sure if it's a feature or a bug.

mathjazz:
It's a bug. If the translation is deleted by the user, it should be removed from TM. Otherwise you submit a typo, update it, but it will still show in TM results.

The other case is if the resource or entity gets obsolete. In this case we should keep the TM entry. And we do. And we will after this patch.

I'll update the test.

jotes:
Ok, i forgot that delete method has a differrent context.

mathjazz:
Test updated. Also no need to test for deleting entities (because we don't do that).

PR #630, opened 2017-07-14
Add GraphQL

stasm:
<del>WIP branch for adding a GraphQL endpoint to Pontoon.</del> This PR adds the `/graphql` API endpoint to Pontoon.

 The main driver is the use case from [bug 1302053](https://bugzilla.mozilla.org/show_bug.cgi?id=1302053):

   - Stats for a locale: supported projects, status of each project.
   - Stats for a project: supported locales, incomplete locales, complete locales.

Research: https://wiki.mozilla.org/L10n:Pontoon/API
Discussion: https://groups.google.com/forum/#!topic/mozilla.tools.l10n/R1S7Pk-c6uU.


stasm:
@jotes: For some reason I can't flag you for review. Please take a look and let me know what you think. 

@flodolo: Same thing, please. Can you take make a functional review and see if this answers the questions you had?

mathjazz:
Thanks for the patch, stas!

I'll wait for @flodolo's functional review. I'll also try to fix the test, which is failing unrelatedly to this PR.

flodolo:
>  Stats for a project: supported locales, incomplete locales, complete locales. 

Incomplete/complete locales: right now it would be based on the missing_strings, i.e. you parse the JSON and extract the locales with 0 missing strings. 

I've realized that we need something more, though, and that would be errors: if a locale has errors, it shouldn't be considered complete. That's likely a follow up fix with a lot more work (I don't think Pontoon knows about those errors yet).

I'd really like the JSON to be simpler (less nesting), but since it's going to be parsed by scripts, that shouldn't be an issue.

As far as I'm concerned, this is a step in the right direction 👍

stasm:
> I've realized that we need something more, though, and that would be errors: if a locale has errors, it shouldn't be considered complete. That's likely a follow up fix with a lot more work (I don't think Pontoon knows about those errors yet).

That's a good point. Errors and warnings are #622. I'm in favor of adding them to the API as a follow-up. How soon that happens in another matter: if you think having this data is a blocker, we should escalate the priority of #622.

flodolo:
> How soon that happens in another matter: if you think having this data is a blocker, we should escalate the priority of #622.

It's a blocker to start publicizing the API to the outside, but I don't think it should block this PR, since it already has a lot going on.

jotes:
@stasm hey, that's probably because i'm not assigned to mozilla organization. I'll try to look into today/tomorrow.

pontoon/api/schema.py
```diff
@@ -0,0 +1,120 @@
...
+    locale = graphene.Field(Locale, code=graphene.String())
+
+    def resolve_projects(self, args, context, info):
+        qs = ProjectModel.objects.all()
+        fields = get_fields(info)
+
+        if 'projects.localizations' in fields:
+            qs = qs.prefetch_related('project_locale__locale')
+
+        if 'projects.localizations.locale.localizations' in fields:
```
adngdb: I wonder if you could make this more generic. Maybe see if some defined strings appear more than once in those fields?

pontoon/api/schema.py
```diff
@@ -0,0 +1,120 @@
...
+        qs = LocaleModel.objects
+        fields = get_fields(info)
+
+        if 'locale.localizations' in fields:
+            qs = qs.prefetch_related('project_locale__project')
+
+        if 'locale.localizations.project.localizations' in fields:
+            raise Exception('Cyclic queries are forbidden')
+
+        return qs.get(code=args['code'])
```
adngdb: There's a pattern to those 4 ``resolve`` methods. Maybe you can refactor them into a more generic function that would be called with different arguments by each method?

requirements.txt
```diff
@@ -171,3 +171,21 @@ setuptools==36.2.4 \
...
+    --hash=sha256:36134e026ec987514b240b57e3b6a995d6aed0fc5dece0567e77d5aed0ad0ece
+iso8601==0.1.11 \
+    --hash=sha256:c68dbd1b6ecc0c13c1d94116aec79d5d5c3bc7444f99159b968f12d83cbc7fa6 \
+    --hash=sha256:e8fb52f78880ae063336c94eb5b87b181e6a0cc33a6c008511bac9a6e980ef30
+singledispatch==3.4.0.3 \
+    --hash=sha256:833b46966687b3de7f438c761ac475213e53b306740f1abfaa86e1d1aae56aa8 \
+    --hash=sha256:5b06af87df13818d14f08a028e42f566640aef80805c3b50c5056b086e3c2b9c
+graphql-core==1.1 \
+    --hash=sha256:63bb8593aeeadb0a53e14207b910027fe51158d017927fad87326dac806185ee
+graphql-relay==0.4.5 \
```
adngdb: Also, I don't think we use ``relay`` just yet, so maybe remove it from the list?

pontoon/urls.py
```diff
@@ -76,4 +81,11 @@
 
     # Team page: Must be at the end
     url(r'^(?P<locale>[A-Za-z0-9\-\@\.]+)/$', team, name='pontoon.teams.team'),
+
+    # GraphQL endpoint
+    # In DEV mode it serves the GraphiQL editor if accessed with
+    # Accept: text/html
+    url(r'^graphql', csp_exempt(
+        csrf_exempt(
+            GraphQLView.as_view(schema=schema, graphiql=DEV)))),
```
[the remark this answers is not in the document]
stasm: No worries, I'll change it.

requirements.txt
```diff
@@ -171,3 +171,21 @@ setuptools==36.2.4 \
...
+    --hash=sha256:2716b7245d97091af21abf096fabafac576905096d21ba7118fba722596f65db
+promise==2.0.2 \
+    --hash=sha256:f6ac7ec189bfb14ab8987f211db85fa8147fe13ca5d2a8c15926307d8948c082
+typing==3.6.1 \
+    --hash=sha256:c36dec260238e7464213dcd50d4b5ef63a507972f5780652e835d0228d0edace
```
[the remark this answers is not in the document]
stasm: Sure, I like this format!

pontoon/api/schema.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class StringStats(graphene.AbstractType):
+    missing_strings = graphene.Int()
+
+    @graphene.resolve_only_args
+    def resolve_missing_strings(self):
+        return (
+            self.total_strings - self.translated_strings -
+            self.approved_strings - self.fuzzy_strings)
```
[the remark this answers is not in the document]
stasm: Funny: I usually write my JS with closing `)` on a new line. Here I thought it was more PEP8-compliant to put the `)` on the same line. I now see that PEP8 uses both conventions throughout its examples. I'm fine with whichever :)

pontoon/api/schema.py
stasm: I'd like to move this to a follow-up.  The whole thing might change when we upgrade to Graphene 2.0, too.

pontoon/api/schema.py
stasm: I didn't want to refactor too early. I'm leaning towards proposing that we keep these methods as they are for now and try to abstract them when doing research into upgrading to Graphene 2.0.

For instance, we might want to move the cyclic-dependency check to a decorator. `get_fields` could also be a decorator.

requirements.txt
stasm: It looks like graphene-django requires it regardless: https://github.com/graphql-python/graphene-django/blob/9cac83b1688f3f608fa22fe2e95e2dd8145ccce1/graphene_django/fields.py#L9

pontoon/api/schema.py
adngdb: Sounds good to me! Those "I wonder" comments are there to trigger some thoughts in your mind, in case you missed them. I'm fine with whatever you decide regarding them.

pontoon/api/schema.py
adngdb: Same here, sounds good to me!

pontoon/api/schema.py
stasm: Cool, thanks for a great review. I'll file follow-up bugs to keep track of these refactoring tangents.

pontoon/api/schema.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+class StringStats(graphene.AbstractType):
+    missing_strings = graphene.Int()
+
+    @graphene.resolve_only_args
+    def resolve_missing_strings(self):
+        return (
+            self.total_strings - self.translated_strings -
+            self.approved_strings - self.fuzzy_strings
```
adngdb: I think you missed this question in the previous review (it's my fault, it was mixed in with another comment). Is ``rejected`` not missing here? What is ``translated``? How is it different from ``approved``? cc @mathjazz

mathjazz: translated is suggested here. sorry. :)

stasm: I used the definition I found here: https://github.com/mozilla/pontoon/blob/dd266843219e14eb41a7d871a63d7750f0386857/pontoon/base/templates/widgets/progress_chart.html#L35

Does this look right?

mathjazz: It does look right.

Added a comment in https://bugzilla.mozilla.org/show_bug.cgi?id=1377969#c1.

adngdb: @mathjazz why do we not need to account for rejected strings here?

mathjazz: Stats only keep track of active translations, which can be either:
* Translated (`approved_strings`)
* Suggested (`translated_strings`) - either unreviewed or rejected
* Fuzzy (`fuzzy_strings`)
* Missing (`resolve_missing_strings()`)

So translated + suggested + fuzzy + missing = total.

In [bug 1377969](https://bugzilla.mozilla.org/show_bug.cgi?id=1377969) we'll get rid of the Suggested and treat them as missing. So translated + fuzzy + missing = total will be true.

We'll add unreviewed suggestions as a separate count. We could also add rejected and missing counts and store them in the model.

adngdb: Thanks for that explanation!

pontoon/urls.py
```diff
@@ -2,6 +2,11 @@
 from django.contrib import admin
 from django.contrib.auth.views import logout
 from django.views.generic import RedirectView, TemplateView
+from django.views.decorators.csrf import csrf_exempt
+from csp.decorators import csp_exempt
+from graphene_django.views import GraphQLView
```
mathjazz: Nit: Please add blank lines in 5.5 and 7.5 to split Django, "other requirements" and pontoon imports into different sections. And move "other requirements" section above Django.

@adngdb Didn't we put the instructions for structuring import statements somewhere in our docs? Can't find.

stasm:
> Since this is listed as a popular use case in the bug, I wonder if we also need to add other abstractions like complete/incomplete locales for a project.

That's a great idea! I added it in a recent commit.

pontoon/api/schema.py
```diff
@@ -0,0 +1,124 @@
+import graphene
+from graphene_django import DjangoObjectType
+from graphene_django.debug import DjangoDebug
+
+from .util import get_fields
+
+from ..base.models import (
```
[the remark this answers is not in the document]
stasm: Leaving the relative import here sounds good to me.

pontoon/api/util.py
```diff
@@ -0,0 +1,54 @@
+from graphql.language.ast import FragmentSpread
+
+
+def get_fields(info):
```
jotes: Do you plan to add tests in this PR? I can try to help

pontoon/api/schema.py
```diff
@@ -0,0 +1,118 @@
...
+    @graphene.resolve_only_args
+    def resolve_localizations(obj):
+        return obj.project_locale.all()
+
+
+class Query(graphene.ObjectType):
+    debug = graphene.Field(DjangoDebug, name='__debug')
+
+    projects = graphene.List(Project)
+    project = graphene.Field(Project, slug=graphene.String())
```
jotes: nit and curiosity: I wonder what's the convention -> I saw that a few random internet pages capitalize all fields in their GraphQL examples.

jotes:
Hey...
So, I'm a fan of your work @stasm :D Totally can't wait until this will see the production.
There are some relatively things that you probably have in mind but I'm not sure how look the current roadmap;
* Access to objects like e.g. user profile (that would require research on the autorization), I think that pontoon-tools could benefit from such api.
* Documentation for endpoints: is there anything that could generate documentation from the code?
* I wonder if on the production that will be a public api or will you restrict it to the selected applications/will require authorization from api user?

stasm:
Thanks, @jotes!

> Access to objects like e.g. user profile (that would require research on the autorization), I think that pontoon-tools could benefit from such api.

Yeah, absolutely. Let's build a list of use-cases and then map them to features. Getting the user profile data (after auth) might be one of them. I started to keep track of those ideas in https://wiki.mozilla.org/L10n:Pontoon/API#Roadmap.

> Documentation for endpoints: is there anything that could generate documentation from the code?

I don't think that's needed. The GraphiQL IDE has a Docs dock which should be sufficient for now. I do wonder if there's a way to provide one-sentence summaries of the fields (maybe in `models.py`) so that they show up there.

> I wonder if on the production that will be a public api or will you restrict it to the selected applications/will require authorization from api user?

I'd like the read-only part to be public for now. In the future we might introduce API keys or JWT tokens to have better control over who accesses the data (and possibly throttle them) and to allow write queries.

pontoon/api/schema.py
stasm: Looking at the example on http://graphql.org/, the convention is to capitalize shape types and use all-lowercase for field names.

pontoon/api/util.py
stasm: Yes, I'm working on it!

pontoon/api/schema.py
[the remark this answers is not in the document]
adngdb: Actually, I think we should avoid relative imports altogether.

pontoon/urls.py
adngdb: It's a part of PEP8: https://www.python.org/dev/peps/pep-0008/#imports

The only difference that we do is that we separate the framework imports from the rest of third party imports. I guess we could add a note about that in ``CONTRIBUTING.rst``?

pontoon/api/schema.py
adngdb: That would be coherent with PEP8's naming conventions, so I like it. Let's not be weird. :)

stasm:
Update: I need a couple of days to finish working on something else but I should be able to finish writing tests here by the end of the week.

pontoon/api/tests/test_urls.py
```diff
@@ -0,0 +1,72 @@
...
+class TestGraphiQL(TestCase):
+
+    body = {
+        'query': '{projects{name}}'
+    }
+
+    def test_graphql_dev_get(self):
+        with self.settings(DEV=True):
+            response = self.client.get('/graphql', self.body,
+                                       HTTP_ACCEPT="application/json")
```
adngdb: I don't like this styling. It is harder to maintain, as if you change the variable name for example, you have to realign the entire block. Instead, can you write these like that: 

```python
response = self.client.get(
    '/graphql', 
    self.body,
    HTTP_ACCEPT="application/json"
)
```

Also, remember that we allow lines to be 100 characters long nowadays. Maybe that one can fit on a single line?

pontoon/api/urls.py
```diff
@@ -4,8 +4,8 @@
 from django.conf.urls import url
 from django.views.decorators.csrf import csrf_exempt
 
+from pontoon.settings import DEV
```
adngdb: ``api`` comes before ``settings`` in my alphabet. ;-)

pontoon/api/tests/test_schema.py
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        response = self.client.get('/graphql', body,
+                                   HTTP_ACCEPT="application/json")
+
+        self.assertEqual(response.status_code, 200)
+        self.assertJSONEqual(
+            force_text(response.content),
+            {"data": {"projects": [
+                {"name": "Pontoon Intro"}]}})
```
adngdb: Please put closing brackets and parentheses on the same level as the opening one. Like this:

```python
self.assertJSONEqual(
    force_text(response.content),
    {
        "data": {
            "projects": [
                {"name": "Pontoon Intro"}
            ]
        }
    }
)
```

It takes a lot more lines but I think it is more readable and easier to maintain. One thing you can ask yourself when writing code is: if someone changes that, will the diff be readable? That's why we add commas at the end of lines in lists even if it's the last element, for example. 

This:

```diff
  [
      'abc',
+     'def',
  ]
```

... is better than this:

```diff
  [
-     'abc'
+     'abc',
+     'def',
  ]
```

pontoon/api/tests/test_schema.py
```diff
@@ -0,0 +1,149 @@
...
+                }
+            }'''
+        }
+
+        response = self.client.get('/graphql', body,
+                                   HTTP_ACCEPT="application/json")
+
+        self.assertEqual(response.status_code, 200)
+        self.assertJSONEqual(
+            force_text(response.content),
```
adngdb: Why do you use ``force_text``?

pontoon/api/tests/test_schema.py
```diff
@@ -0,0 +1,149 @@
...
+            {'data': {'projects': None},
+             'errors': [
+                 {'locations': [{'column': 17, 'line': 2}],
```
adngdb: This might be too specific. If you update the library and they change the format of this response, you'll have to update the test. I think here you can be fine with only verifying if ``Cyclic queries are forbidden`` is in the response content.

pontoon/api/schema.py
```diff
@@ -0,0 +1,117 @@
...
+
+    @graphene.resolve_only_args
+    def resolve_localizations(obj):
+        return obj.project_locale.all()
+
+
+class Query(graphene.ObjectType):
+    debug = graphene.Field(DjangoDebug, name='__debug')
+
+    projects = graphene.List(Project, with_disabled=graphene.Boolean(False))
```
adngdb: Can you add a comment that explains that this with_disabled thing is a default value for queries? I assume that's what it does because of what you told me earlier, but it isn't obvious looking at the code.

adngdb: And here's the proof that a comment is needed: I didn't understand what that was actually doing. I think I would call that argument ``include_disabled`` instead, to make very clear what it does. And please do add a comment too. :)

pontoon/api/urls.py
stasm: Ah, sorry, I thought more general module paths should go first.

pontoon/api/tests/test_schema.py
stasm: I ran into some encoding problem when I was writing these tests, but now I just removed `force_text` and everything works fine.

pontoon/api/tests/test_util.py
```diff
@@ -5,12 +5,18 @@
 
 
 class MockResolveInfo:
+    """Mock of ResolveInfo with data required by get_fields.
+
+    GraphQL passes ResolveInfo objects to resolvers. They include the name of the field being
+    resolved, the query AST, fragments ASTs, the schema, the execution context and others.
+    get_fields uses ResolveInfo to introspect the AST and return a flat list of requested fields.
+    """
```
adngdb: :+1:

PR #632, opened 2017-07-19
Prefix element IDs in Fluent UI

mathjazz:
@jotes r?

Instead of using attribute names as element IDs in Fluent UI, prefix them with `ftl-id-` to prevent Pontoon CSS and JS code having impact.

Without this fix, Pontoon can easily look like this (`title` attribute sets the input element ID to `title`, which has a rule in CSS):

schermata 2017-07-19 alle 20 37 50

### DIFF
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -48,10 +48,10 @@ var Pontoon = (function (my) {
             $('#ftl-area .main-value ul')
               .append(
                 '<li class="clearfix">' +
-                  '<label class="id built-in" for="' + this + '">' +
+                  '<label class="id built-in" for="ftl-id-' + this + '">' +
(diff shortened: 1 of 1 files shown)


pontoon/base/static/js/fluent_interface.js
```diff
@@ -73,10 +73,10 @@ var Pontoon = (function (my) {
             maxlength = '1';
             input = '<div class="accesskeys"></div>';
           }
-          input += '<input class="value" id="' + id + '" type="text" value="' + value + '" maxlength="' + maxlength + '">';
+          input += '<input class="value" id=ftl-id-"' + id + '" type="text" value="' + value + '" maxlength="' + maxlength + '">';
```
jotes: `"ftl-id-`?

mathjazz: Brilliant catch!

PR #634, opened 2017-07-20
Fixes bug 1288956 - Reject translations instead of deleting them.

adngdb:
@mathjazz this is a work in progress. It is missing a few things that you listed in the bug. Also, some things I just copy/pasted without really understanding what they were doing... Anyway, the basic UI is there, so if you want to try it out and give me some feedback, that'd be cool!


adngdb:
@mathjazz I believe I have addressed all the missing points!

mathjazz:
Great progress!

One nit: after rejecting all suggestions as a mass action, we should update the button title with "N suggestions rejected" (instead of "N strings rejected").

Two more things I apparently forgot from my initial review (sorry for that!):
* If string only has rejected and unreviewed suggestions (so no translation is approved), we should display the latest unreviewed suggestion as an active translation, i.e. in the string list and on top of the history tab. Check out `Entity.get_translation()` and the `get_translation_history()` view.
* When translation is rejected or unrejected, we don't immediately update the state on the client side (string list icon, progress bar). It does update on refresh (because the data gets properly updated in the DB).

Let me know if you want me to do the code review before you address these issues.

adngdb:
@mathjazz I'm confident issue 1 is solved, but I'm not so sure about issue 2. Here's [the change I made](https://github.com/mozilla/pontoon/pull/634/commits/08cdf156e04140a2c2b03b80b6c17f640b6833ae), that cover everything? If not, can you be more specific about what exact behavior you expect?

mathjazz:
Thanks for the update, sorting works great.

> @mathjazz I'm confident issue 1 is solved, but I'm not so sure about issue 2. Here's the change I made, does that cover everything? If not, can you be more specific about what exact behavior you expect?

You can update translation state by calling `Pontoon.updateTranslation(entity, plural_form, translation_data)`. The function also takes care of updating the UI (in string list and the progress bar). You'll also need to return the translation data from the `reject_translation`view.

You can follow the code after the `/unapprove-translation/` call. And while you're at it, you can remove the redundant call to `self.updateProgress(entity)` in there. It's not neccessary, because `Pontoon.updateTranslation()`already calls it.

adngdb:
@mathjazz how's it looking now?

pontoon/base/static/js/translate.js
```diff
@@ -1995,98 +1996,77 @@ var Pontoon = (function (my) {
...
+            self.updateTranslation(entity, pf, data.translation);
 
-                // Update number of history entities
-                var count = $('#helpers .history .suggestion').length;
-                $('#helpers a[href="#history"] .count')
-                  .toggle(count > 0)
-                  .html(count);
-              });
+            var item = button.parents('li');
+            item.addClass('rejected').removeClass('translated');
```
mathjazz: You should also remove the `suggested` and `fuzzy` class.

pontoon/base/static/js/translate.js
```diff
@@ -1995,98 +1996,77 @@ var Pontoon = (function (my) {
...
+            paths: self.getPartPaths(self.currentPart)
+         }).then(function(data) {
+           var entity = self.getEditorEntity();
+           var pf = self.getPluralForm(true);
+
+           self.stats = data.stats;
+
+           self.updateTranslation(entity, pf, data.translation);
+
+           $('#translation').val(data.translation.string).focus();
```
mathjazz: In case #651 gets merged earlier, we should replace this with a function. :-)
https://github.com/mozilla/pontoon/pull/651/files#diff-5ec2d12043f59ff839712736b821ec71L458

pontoon/base/templates/translate.html
```diff
@@ -173,7 +173,7 @@
 
             {{ Filter.item('untranslated', 'Untranslated') }}
             {{ Filter.item('unchanged', 'Unchanged') }}
-            {{ Filter.item('has-suggestions', 'Has Suggestions') }}
+            {{ Filter.item('has-suggestions', 'Has Unreviewed Suggestions') }}
```
adngdb: Changing the slug will break compatibility for bookmarks and links. We should keep this is mind if we do change it. Maybe keep an alias of the old value?

pontoon/base/views.py
```diff
@@ -314,13 +314,28 @@ def get_translations_info(translations):
...
+        )
+        count, translated_resources, changed_entities = get_translations_info(
+            suggestions
+        )
+        TranslationMemoryEntry.objects.filter(translation__in=suggestions).delete()
+        suggestions.update(
+            rejected=True,
+            rejected_user=request.user,
+            rejected_date=timezone.now(),
+        )
```
adngdb: Mind if I do that in a follow-up bug? This is getting pretty big, and I plans to refactor this batch actions logic. I could do that there.

adngdb: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1386267

adngdb:
@mathjazz all things should be addressed now! :)

pontoon/base/models.py
```diff
@@ -2020,7 +2042,8 @@ def save(self, imported=False, *args, **kwargs):
             (Translation.objects
                 .filter(entity=self.entity, locale=self.locale, plural_form=self.plural_form)
                 .exclude(pk=self.pk)
-                .update(approved=False, approved_user=None, approved_date=None))
+                .update(approved=False, approved_user=None, approved_date=None,
+                        rejected=True, rejected_user=None, rejected_date=None))
```
mathjazz: Looks like you forgot to fix this one.

pontoon/base/static/js/translate.js
```diff
@@ -2240,7 +2220,14 @@ var Pontoon = (function (my) {
             success: function(data) {
               if ('count' in data) {
                 var strings = data.count === 1 ? 'string' : 'strings';
-                message = data.count + ' ' + strings + ' ' + action + 'd';
+                var actionText = action + 'd';
+
+                if (action === 'reject') {
+                  itemsText = 'suggestion' + (data.count === 1 ? '' : 's');
```
mathjazz: You should initialize `itemsText`, otherwise it will raise a `ReferenceError` on any action that is not *reject*.

pontoon/base/views.py
```diff
@@ -306,21 +306,46 @@ def get_translations_info(translations):
...
-        TranslationMemoryEntry.objects.filter(translation__in=translations).delete()
-        translations.delete()
+        # Reject all other non-rejected translations.
+        suggestions = translations.filter(approved=False, rejected=False)
+        suggestions.update(
+            rejected=True,
+            rejected_user=request.user,
+            rejected_date=timezone.now(),
+            fuzzy=False,
+        )
```
mathjazz: You need to reject non-rejected suggestions to strings for which you just approved active translations for. (This sounds funny, but it's true).

Instead you're currently rejecting active non-rejected suggestions (which don't even exist, because you just approved a superset of it - `approved`).

adngdb: I think I get it, and I think it's fixed with my last commit.

adngdb:
@mathjazz ^ this is quite a big change, and I'm not exactly sure how to best test it. I think it could have impact on many things that I do not even suspect. Can I count on your review? :)

mathjazz:
@adngdb Seems like none of the changed files in those commits is changed in this PR, so we should be fine.

Have you seen https://github.com/mozilla/pontoon/pull/634#pullrequestreview-54110118?

adngdb:
@mathjazz The last commit actually changes those files, but the way the diff works makes it seem like it didn't change anything. See here: https://github.com/mozilla/pontoon/pull/634/commits/11b4d5b1040ac5dd1ed4de266daf566dfa99b77a

mathjazz:
Oops, I overlooked that! Testing...

adngdb:
@mathjazz Here's the migration. I've tested it locally and it seems to work fine.

pontoon/base/migrations/0097_reject_suggestions_older_than_approved_translation.py
```diff
@@ -0,0 +1,44 @@
...
+    SET rejected = TRUE
+    WHERE approved = FALSE
+    AND date < (
+        SELECT GREATEST(bt.date, bt.approved_date)
+        FROM base_translation bt
+        WHERE bt.approved = TRUE
+        AND bt.entity_id = base_translation.entity_id
+        AND bt.locale_id = base_translation.locale_id
+        ORDER BY bt.date, bt.approved_date
+        LIMIT 1
```
adngdb: Guess what? It seems we sometimes have several approved translations for an entity. At least, I've had an error with this sub query returning several results when running this on my local database. Anyway, this is a safe way to make sure we only take one.

mathjazz: I believe that's because we can have multiple approved translations for pluralized entities.

You need to add something like this:
`AND bt.plural_form = base_translation.plural_form`

pontoon/base/migrations/0097_reject_suggestions_older_than_approved_translation.py
```diff
@@ -0,0 +1,44 @@
...
+
+A translation will be rejected if:
+  * it is not approved, and
+  * it is older than either the submission date or the approval date
+    of the approved translation for its entity.
+"""
+SQL = """
+    UPDATE base_translation
+    SET rejected = TRUE
+    WHERE approved = FALSE
```
mathjazz: We also need to avoid rejecting fuzzy translations.

pontoon/base/migrations/0097_reject_suggestions_older_than_approved_translation.py
```diff
@@ -0,0 +1,44 @@
...
+    AND date < (
+        SELECT GREATEST(bt.date, bt.approved_date)
```
mathjazz: If the `bt.approved_date` is not set, this returns the value of `bt.date`, right?

adngdb: Yep!

adngdb:
@mathjazz those commits look good to me!

mathjazz:
Woohoo, let's merge this thing!

PR #636, opened 2017-07-20
Fix bug 1382551: Extend keyboard navigation to entire translate view

mathjazz:
Turn keyboard shortcuts for moving between strings from a translation
editor to the entire translate page.

@jotes r?

### DIFF
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -1811,18 +1811,6 @@ var Pontoon = (function (my) {
           return false;
         }
 
-        // Alt + Down: Go to next string
-        if (e.altKey && key === 40) {
-          self.navigateToEntity('next');
-          return false;
-        }
-
-        // Alt + Up: Go to previous string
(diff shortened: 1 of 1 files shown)


pontoon/base/static/js/translate.js
```diff
@@ -2723,6 +2711,23 @@ var Pontoon = (function (my) {
     attachMainHandlers: function () {
       var self = this;
 
+      // Main keyboard shortcuts
+      $('html').on('keydown', function (e) {
+        var key = e.which;
```
jotes: I wonder if something like this:
```
var direction = {
 40: 'next',
 38: 'previous'
}[e.which]; //yeah, that can be moved into something nicer
if (e.altKey && direction) {
   self.navigateToEntity(direction);
   return false;
}
```
Would reduce complexity/number of nested blocks? what do you think about that approach?

mathjazz: I actually find the original version easier to read. 😂

PR #637, opened 2017-07-23
Fix bug 1383482: Replace tip with default

mathjazz:
Unify the way we reference the latest revision in Mercurial repos:
1. When determining the repository's latest revision
2. When comparing with the last synced revision to determine what
changed since the last sync.

@Pike r?


mathjazz:
I wonder if we even need to use the `--rev`parameter. Even if we don't, I'd rather be explicit.

mathjazz:
Thanks! Let's talk about this at the water cooler. :P

PR #638, opened 2017-07-24
Faster locale and source directories detection on sync

mathjazz:
While reviewing #634, I noticed sync is slow at finding directories for each locale. It turned out the fix is pretty straightforward. I tested it with a few projects locally and on stage. We can expect the patch to shave off up to a couple of minutes of sync time from bigger projects.

Currently, we search for locale directories by walking the entire project checkout path for each locale. It's not surprising that it performs badly for projects with 10,000s of files. With this patch, we only do this once, and then try to find paths for other locales on the [same nesting level](https://developer.mozilla.org/en-US/docs/Mozilla/Implementing_Pontoon_in_a_Mozilla_website).

I've also added a smaller (a few seconds worth) optimization of detecting source directories.

@jotes r?


pontoon/sync/core.py
```diff
@@ -228,6 +227,6 @@ def commit_changes(db_project, vcs_project, changeset, locale):
         'authors': set(authors)
     })
 
-    locale_path = locale_directory_path(vcs_project.checkout_path, locale.code)
+    locale_path = vcs_project.locale_directory_paths[locale.code]
```
mathjazz: No need to search for locale directory path on each commit - we already have this information.

pontoon/sync/vcs/models.py
```diff
@@ -206,17 +207,19 @@ def locale_directory_paths(self):
         Create locale directory, if not in repository yet.
         """
         locale_directory_paths = {}
+        parent_directories = set()
```
mathjazz: Sometimes locales can actually be on different paths - if we use multiple repositories (e.g. in Firefox we have locales on `hg.mozilla.org` and `BitBucket`). So we store `parent_direcotry` for each of them.

pontoon/sync/utils.py
```diff
@@ -63,24 +70,34 @@ def directory_contains_resources(directory_path, source_only=False):
...
     possible_paths = []
-    for root, dirnames, filenames in os.walk(checkout_path):
-        if locale_code in dirnames:
-            possible_paths.append(os.path.join(root, locale_code))
 
-        locale_variant = locale_code.replace('-', '_')
-        if locale_variant in dirnames:
-            possible_paths.append(os.path.join(root, locale_variant))
+    # Check paths that use underscore as locale/country code separator
+    locale_variant = locale_code.replace('-', '_')
```
jotes: `locale_code_variants = [locale_code, locale_code.replace('-', '_')]`?

pontoon/sync/utils.py
```diff
@@ -63,24 +70,34 @@ def directory_contains_resources(directory_path, source_only=False):
...
+
+    # Optimization for directories with a lot of paths
+    for parent_directory in parent_directories:
+        for locale in [locale_code, locale_variant]:
+            candidate = os.path.join(parent_directory, locale)
+            if os.path.exists(candidate):
+                possible_paths.append(candidate)
+
+    if not possible_paths:
+        for root, dirnames, filenames in os.walk(checkout_path):
```
jotes: I think we could try https://github.com/benhoyt/scandir

pontoon/sync/utils.py
```diff
@@ -63,24 +70,34 @@ def directory_contains_resources(directory_path, source_only=False):
...
-        if locale_code in dirnames:
-            possible_paths.append(os.path.join(root, locale_code))
 
-        locale_variant = locale_code.replace('-', '_')
-        if locale_variant in dirnames:
-            possible_paths.append(os.path.join(root, locale_variant))
+    # Check paths that use underscore as locale/country code separator
+    locale_variant = locale_code.replace('-', '_')
+
+    # Optimization for directories with a lot of paths
```
jotes: I'd only add examples of directories that will be matched -> it will be a little easier to understand.

pontoon/sync/utils.py
mathjazz: Good idea! It's around 50% faster!

PR #639, opened 2017-07-25
use requests and jinja2 template instead of external suds library (deprecated)

mastizada:
Suds library is deprecated and will not support python 3

### DIFF
--- a/pontoon/machinery/templates/machinery/microsoft_terminology.jinja
+++ b/pontoon/machinery/templates/machinery/microsoft_terminology.jinja
@@ -1,0 +1,20 @@
+<?xml version='1.0' encoding='utf-8'?>
+<soap-env:Envelope xmlns:soap-env="http://schemas.xmlsoap.org/soap/envelope/">
+    <soap-env:Header>
+        <wsa:Action xmlns:wsa="http://www.w3.org/2005/08/addressing">http://api.terminology.microsoft.com/terminology/Terminology/GetTranslations</wsa:Action>
+        <wsa:MessageID xmlns:wsa="http://www.w3.org/2005/08/addressing">urn:uuid:{{ uuid }}</wsa:MessageID>
+        <wsa:To xmlns:wsa="http://www.w3.org/2005/08/addressing">http://api.terminology.microsoft.com/Terminology.svc</wsa:To>
+    </soap-env:Header>
+    <soap-env:Body>
+        <ns0:GetTranslations xmlns:ns0="http://api.terminology.microsoft.com/terminology">
+            <ns0:text>{{ text }}</ns0:text>
+            <ns0:from>en-US</ns0:from>
+            <ns0:to>{{ to }}</ns0:to>
+            <ns0:sources>
+                <ns0:TranslationSource>Terms</ns0:TranslationSource>
+                <ns0:TranslationSource>UiStrings</ns0:TranslationSource>
+            </ns0:sources>
+            <ns0:maxTranslations>{{ max_result }}</ns0:maxTranslations>
+        </ns0:GetTranslations>
+    </soap-env:Body>
+</soap-env:Envelope>
--- a/pontoon/machinery/views.py
+++ b/pontoon/machinery/views.py
@@ -2,14 +2,15 @@ import logging
 import requests
 import urllib
 import xml.etree.ElementTree as ET
+from uuid import uuid4
 
 from collections import defaultdict
 from django.conf import settings
 from django.db import DataError
 from django.http import HttpResponse, HttpResponseBadRequest, JsonResponse
 from django.shortcuts import get_object_or_404, render
+from django.template.loader import get_template
 from django.utils.datastructures import MultiValueDictKeyError
-from suds.client import Client, WebFault
 
 from pontoon.base import utils
 from pontoon.base.models import Locale, TranslationMemoryEntry
@@ -131,8 +132,11 @@ def microsoft_terminology(request):
 
     obj = {}
     locale = locale.lower()
-    url = 'http://api.terminology.microsoft.com/Terminology.svc?singleWsdl'
-    client = Client(url)
+    url = 'http://api.terminology.microsoft.com/Terminology.svc'
+    headers = {
+        'SOAPAction': '"http://api.terminology.microsoft.com/terminology/Terminology/GetTranslations"',
+        'Content-Type': 'text/xml; charset=utf-8'
+    }
 
     # On first run, check if target language supported
     if check == "true":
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -46,8 +46,18 @@ rsa==3.1.4 \
...
+cached-property==1.3.0 \
+    --hash=sha256:8aacd79aa85a793419c76de9662d9a72ad4751eda2a8b8eb94223adc4d8d85f1
+isodate==0.5.4 \
+    --hash=sha256:42105c41d037246dc1987e36d96f3752ffd5c0c24834dd12e4fdbe1e79544e31
+requests-toolbelt==0.8.0 \
+    --hash=sha256:42c9c170abc2cacb78b8ab23ac957945c7716249206f90874651971a4acff237
+defusedxml==0.5.0 \
+    --hash=sha256:702a91ade2968a82beb0db1e0766a6a273f33d4616a6ce8cde475d8e09853b20
+zeep==2.2.0 \
+    --hash=sha256:21749a2d89f4aa6e83811dd01c611afccc81af8cb52bb2c574e852802e404f9d
```
mastizada: Tested in both Python 2 and 3 environments, works for both. All these were required by zeep (using >= version number which throws error when using hashes).

pontoon/machinery/views.py
```diff
@@ -9,7 +9,8 @@
 from django.http import HttpResponse, HttpResponseBadRequest, JsonResponse
 from django.shortcuts import get_object_or_404, render
 from django.utils.datastructures import MultiValueDictKeyError
-from suds.client import Client, WebFault
+from zeep import Client
+from zeep.exceptions import Fault as WebFault
```
mastizada: Decided to leave as WebFault as Fault is casual word and less meaningful. Should I leave it as Fault?

Pike:
This is a lot of dependencies we take for a single API call.

Can we instead just mock the SOAP call with something built-in?

mastizada:
@Pike, makes sense, will try to make it using requests and jinja libraries which pontoon already uses.

pontoon/machinery/views.py
```diff
@@ -160,34 +164,34 @@ def microsoft_terminology(request):
...
-    sources = client.factory.create('ns0:TranslationSources')
-    sources["TranslationSource"] = ['Terms', 'UiStrings']
-
     payload = {
-        'text': text,
-        'from': 'en-US',
-        'to': locale,
-        'sources': sources,
-        'maxTranslations': 5
+        "uuid": uuid4(),
```
jotes: `'uuid'`, because we use `'` where it's possible.

pontoon/machinery/views.py
```diff
@@ -160,34 +164,34 @@ def microsoft_terminology(request):
...
+                    'target': translation.find(xpath + 'TranslatedText').text,
+                    'quality': int(translation.find(xpath + 'ConfidenceLevel').text),
                 })
 
             obj['translations'] = translations
-
         return JsonResponse(obj)
 
-    except WebFault as e:
+    except Exception as e:
```
jotes: Do you have any specific exception that could happen here? e.g. like xpath errors? because we probably don't want to overshadow any errors introduced by us during the maintenance of this code.

pontoon/machinery/views.py
```diff
@@ -160,34 +164,34 @@ def microsoft_terminology(request):
...
     try:
-        r = client.service.GetTranslations(**payload)
+        r = requests.post(url, data=payload, headers=headers)
         translations = []
-
-        if len(r) != 0:
-            for translation in r.Match:
+        xpath = ".//{http://api.terminology.microsoft.com/terminology}"
+        if r.status_code != 200:
+            return HttpResponseBadRequest('Bad Request: Got {code} status code from service'.format(code=r.status_code))
```
jotes: I wonder If that information would be useful to the end-user of the pontoon  -> I'd log them to our logs with `.error` level and just do http://docs.python-requests.org/en/master/api/#requests.Response.raise_for_status here. That's my personal taste -> the final decision is always on @mathjazz/rest of the folks.

pontoon/machinery/views.py
mathjazz: You can use [requests.exceptions.RequestException](https://stackoverflow.com/questions/16511337/correct-way-to-try-except-using-python-requests-module).

pontoon/machinery/templates/machinery/microsoft_terminology.jinja
```diff
@@ -0,0 +1,20 @@
...
+    <ns0:from>en-US</ns0:from>
+    <ns0:to>{{ to }}</ns0:to>
+    <ns0:sources>
+      <ns0:TranslationSource>Terms</ns0:TranslationSource>
+      <ns0:TranslationSource>UiStrings</ns0:TranslationSource>
+    </ns0:sources>
+    <ns0:maxTranslations>{{ max_result }}</ns0:maxTranslations>
+  </ns0:GetTranslations>
+</soap-env:Body>
+</soap-env:Envelope>
```
mathjazz: Can you send this through a beautifier please? Indents are off.

pontoon/machinery/views.py
mastizada: Request time-out error, other request related error, xpath error (will be `text` call on NoneType). I don't expect any error here but if Microsoft's translation service will crash, that will make it to crash too. Just wanted to prevent that.

pontoon/machinery/views.py
mastizada: Thanks, changing.

pontoon/machinery/views.py
mastizada: Not sure if we even need that. Should I remove it?

mastizada:
@jotes @mathjazz update code.

pontoon/settings/base.py
```diff
@@ -191,6 +191,7 @@ def _site_url():
 TEMPLATES = [
     {
         'BACKEND': 'django_jinja.backend.Jinja2',
+        'NAME': 'jinja2',
```
mathjazz: Just curious, what do we need that for? https://github.com/mozilla/pontoon/pull/639/files#diff-f94a39e4ed5b96101a481a61122332dcR194

mathjazz:
LGTM!

pontoon/settings/base.py
mastizada: Saw that in django-jinja's documentation, its own test project also. Used for django.template.engines (engines['jinja2']) which returns `env` (environment). Don't know if it will be useful in pontoon in future, I solved problem without using it. Should I remove it?

mathjazz: I see, I thought it was somehow related. Let's keep it.

jotes:
:tada: :dancer:

PR #641, opened 2017-07-25
Fix bug 1384248: render strings with value and attributes

mathjazz:


pontoon/base/static/js/fluent_interface.js
```diff
@@ -57,6 +57,10 @@ var Pontoon = (function (my) {
           });
 
           $('#entity-value').parents('li').hide();
+
+        // Main Value
+        } else if (isTranslated && translation_ast.value) {
```
Pike: Should be good enough to just check translation_ast.value?

With that, rs=me.

PR #642, opened 2017-07-25
Prevent DoesNotExist exception

mathjazz:
This is to prevent a possible `DoesNotExist` exception in `Locale.parts_stats`. Is there a better way to do this?

### DIFF
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -134,7 +134,10 @@ def locale_project_parts(request, locale, slug):
     locale = get_object_or_404(Locale, code=locale)
     project = get_object_or_404(Project, slug=slug)
 
-    return JsonResponse(locale.parts_stats(project), safe=False)
+    try:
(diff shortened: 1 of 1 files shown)


adngdb:
I'm not sure what the performance implications are, but I think this adds another database query that we could avoid. The more "pythonic" way of solving this would be to use a ``try`` block. For example: 

```python
try:
    return JsonResponse(locale.parts_stats(project), safe=False)
except DoesNotExist:
    raise Http404('No ProjectLocale matches the given query.')
```

However, maybe it is more "djangonic" to do it the ``get_or_404`` way? I actually don't know.

jotes:
What if we move that logic into fetching project/locale `project = get_object_or_404(Project, slug=slug, locales__in=[locale])`? That would result only in a small join to verify if object with such relation exists.

mathjazz:
It's always a good idea to save a DB request. Updated!

PR #643, opened 2017-07-25
Replace commonware secure cookie patch with Django's built-in security

mastizada:
Commonware is only used for monkey-patch of Django's HttpResponse object for secure cookies. Django's `SESSION_COOKIE_SECURE` and `CSRF_COOKIE_SECURE` should give same result.

### DIFF
--- a/pontoon/settings/base.py
+++ b/pontoon/settings/base.py
@@ -144,7 +144,6 @@ INSTALLED_APPS = (
     'django.contrib.sites',
 
     # Third-party apps, patches, fixes
-    'commonware.response.cookies',
     'django_jinja',
     'django_nose',
     'pipeline',
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,8 +11,6 @@ certifi==14.05.14 \
 chardet==2.3.0 \
     --hash=sha256:aaf514bde38020b4f1e42c6a6e141f2827a8a58ccfc3b22b6ff5a1a4b50be56e
     --hash=sha256:e53e38b3a4afe6d1132de62b7400a4ac363452dc5dfcf8d88e8e0cce663c68aa
-commonware==0.4.3 \
-    --hash=sha256:a7b02a5f76d89a79f861926fb34e029cc4343c13802525c818542a39fe788cce
 configparser==3.5.0b2 \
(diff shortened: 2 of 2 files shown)


docker/config/webapp.env
```diff
@@ -3,6 +3,7 @@ DJANGO_DEV=True
 DJANGO_DEBUG=True
 DATABASE_URL=postgres://pontoon:asdf@postgresql/pontoon
 SESSION_COOKIE_SECURE=False
+CSRF_COOKIE_SECURE=False
```
mathjazz: Why do we need to explicitly disable this in DEV environment, given that it was enabled until now?

pontoon/settings/base.py
```diff
@@ -61,6 +61,7 @@ def path(*args):
 
 SESSION_COOKIE_HTTPONLY = os.environ.get('SESSION_COOKIE_HTTPONLY', 'True') != 'False'
 SESSION_COOKIE_SECURE = os.environ.get('SESSION_COOKIE_SECURE', 'True') != 'False'
+CSRF_COOKIE_SECURE = os.environ.get('CSRF_COOKIE_SECURE', 'True') != 'False'
```
mathjazz: Looking at the [commonware source code](https://github.com/mozilla/commonware/blob/master/commonware/response/cookies/monkeypatch.py), it seems like we also need to add the `CSRF_COOKIE_HTTPONLY` setting?

docker/config/webapp.env
mastizada: It was displayed together with `SESSION_COOKIE_SECURE` in doc and that one was False by default in settings. Should I change that to True or make both True by default?

pontoon/settings/base.py
mastizada: Right, will add tonight.

mathjazz:
After looking into this a little bit deeper, I think we can simply remove `commonware` from the requirements and `INSTALLED_APPS`. Thanks to `SESSION_COOKIE_HTTPONLY` and `SESSION_COOKIE_SECURE`, the session should be fine already. For CSRF we [don't use cookies](https://github.com/mozilla/django-session-csrf).

I tested locally and on stage. Session and POST requests seem to be working fine.

mastizada:
@mathjazz removed only commonware, should be enough

mathjazz:
Thanks!

PR #644, opened 2017-07-26
Bug 1382538: Properly remove comments in FTL translations

mathjazz:
We don't store translation comments in the DB, at least for now. The way we strip them from the translation is hideous. This patch fixes this.

All this should become easier when we'll be able to store AST in the DB instead of the serialized string.


mathjazz:
Thanks! I tested heavily locally. We'll see if heavily enough!

PR #649, opened 2017-07-27
Fix bug 1384846: Fix diff view for FTL strings

mathjazz:
This worked for all formats except for FTL, because getSimplePreview()
returns the fallback value for them, which is passed as the second
argument.

For FTL strings we need to generate their presentation using their
source, which is passed as the first argument. Since the same argument
was passed for the base string and the translate string, the diff
was always empty.

@jotes r?


adngdb:
Oh how I'd love to have front-end unit tests! Though as it is now, it's going to be very hard to do. Let's wait until we rewrite the translate view with something else, and then make sure we have good tests from the start.

PR #651, opened 2017-07-28
Fixes for a few FTL bugs

mathjazz:
Details in commit messages.

@jotes r?

### DIFF
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -317,7 +317,7 @@ var Pontoon = (function (my) {
           } else {
             var attributes = ast.attributes;
             if (attributes && attributes.length) {
-              response = attributes[0].value.elements[0].value;
+              response = this.serializePlaceables(attributes[0].value.elements);
             }
           }
 
@@ -333,7 +333,7 @@ var Pontoon = (function (my) {
                 });
 
             if (isFTLplural) {
-              response = variants[0].value.elements[0].value;
+              response = this.serializePlaceables(variants[0].value.elements);
               entity.isFTLplural = true;
             }
           }
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -118,6 +118,15 @@ var Pontoon = (function (my) {
 
 
       /*
+       * Render editor with given translation
+       */
+      renderEditorWithTranslation: function (translation) {
+        Pontoon.fluent.renderEditor(translation);
+        Pontoon.fluent.focusFirstField();
+      },
+
+
+      /*
        * Serialize value with placeables into a simple strings
        */
       serializePlaceables: function (elements) {
@@ -160,7 +169,7 @@ var Pontoon = (function (my) {
           $('#ftl').addClass('active');
 
           this.renderEditor();
-          $('#ftl-area input.value:visible:first').focus();
+          this.focusFirstField();
 
         } else {
           $('#ftl-area').hide();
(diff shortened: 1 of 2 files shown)


pontoon/base/static/js/fluent_interface.js
```diff
@@ -317,7 +317,7 @@ var Pontoon = (function (my) {
           } else {
             var attributes = ast.attributes;
             if (attributes && attributes.length) {
-              response = attributes[0].value.elements[0].value;
+              response = this.serializePlaceables(attributes[0].value.elements);
```
mathjazz: This fixes this bug:
https://pontoon.mozilla.org/sl/firefox-screenshots/server.ftl/?string=168505

Where the simplified version of the string is cut off if it contains placeables.

pontoon/base/static/js/translate.js
```diff
@@ -1926,12 +1936,22 @@ var Pontoon = (function (my) {
           return;
         }
 
-        var source = $(this).find('.translation-clipboard').text();
+        // FTL Editor
+        if ($('#ftl').is('.active')) {
```
jotes: I think that at some point we'll have to split implementation of editors into separate classes. That's not a blocker, just an observation.

pontoon/base/static/js/translate.js
```diff
@@ -1926,12 +1936,22 @@ var Pontoon = (function (my) {
...
 
-        $('#translation').val(source).focus();
-        self.moveCursorToBeginning();
-        self.updateCurrentTranslationLength();
-        self.updateInPlaceTranslation();
+        // Standard Editor
+        } else {
+          var source = $(this).find('.translation-clipboard').text();
+
+          $('#translation').val(source).focus();
```
jotes: I see that this code pattern occurs few times, maybe it's a good idea to create a method for that kind of render?

pontoon/base/static/js/translate.js
```diff
@@ -1954,45 +1974,53 @@ var Pontoon = (function (my) {
...
+          button.prop('title', 'Approve');
+          button.parents('li.translated').removeClass('translated').addClass('suggested');
+          button.parents('li').find('.info a').prop('title', self.getApproveButtonTitle({
+            approved: false,
+            unapproved_user: self.user.display_name
+          }));
+
+          self.endLoader('Translation has been unapproved.');
+        }, function() {
+          self.endLoader("Couldn't unapprove this translation.");
```
jotes: I wonder if we should inform a user what was a reason of that error (e.g. internal server error vs. you don't have permissions etc.).

pontoon/base/static/js/translate.js
```diff
@@ -1954,45 +1974,53 @@ var Pontoon = (function (my) {
...
+          }
+
+          if (entity.body && pf === 0) {
+            self.postMessage("SAVE", {
+              translation: data.translation.string,
+              id: entity.id
+            });
+          }
+
+          button.removeClass('unapprove').addClass('approve');
```
jotes: I hope that React/Angular will free us from that bloat of changing of state of elements inside the js code.

pontoon/base/static/js/translate.js
mathjazz: Yessss!

pontoon/base/static/js/translate.js
mathjazz: I agree, but I'm not sure if fits into this PR.

pontoon/base/static/js/translate.js
mathjazz: Good point, will update!

PR #652, opened 2017-07-28
Fix bug 1382535: Keep translations on delete

mathjazz:
@flodolo Do you feel comfortable reviewing this?

I'm working on FTL bugfixes and I thought this one was one of them, too. I fixed it by mistake.

### DIFF
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -2045,7 +2045,9 @@ var Pontoon = (function (my) {
 
                   // Last translation deleted, no alternative available
                   } else {
-                    $('#clear').click();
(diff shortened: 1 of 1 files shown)


pontoon/base/static/js/translate.js
```diff
@@ -2045,7 +2045,9 @@ var Pontoon = (function (my) {
 
                   // Last translation deleted, no alternative available
                   } else {
-                    $('#clear').click();
+                    if (self.cachedTranslation === $('#translation').val()) {
+                      $('#clear').click();
+                    }
```
mathjazz: What happens here is now is that we check if the text in the translation input is different than the last active translations. If yes, we don't clear the input.

flodolo:
Testing, should be easy enough.

PR #655, opened 2017-07-29
Bug 1343804: Add ability to save and edit team information

mastizada:
Fix bug 1343804 - Edit team info

### DIFF
--- a/pontoon/settings/base.py
+++ b/pontoon/settings/base.py
@@ -434,7 +434,7 @@ PIPELINE_JS = {
             'js/tabs.js',
             'js/request_projects.js',
             'js/permissions.js',
-            'js/team.js',
+            'js/info.js',
         ),
         'output_filename': 'js/team.min.js',
     },
--- a/pontoon/settings/base.py
+++ b/pontoon/settings/base.py
@@ -728,6 +728,18 @@ MICROSOFT_TERMINOLOGY_LOCALES = [
     'vi-vn', 'wo-sn', 'xh-za', 'yo-ng', 'zh-cn', 'zh-hk', 'zh-tw', 'zu-za',
 ]
 
+# Bleach tags and attributes
+ALLOWED_TAGS = [
+    'a', 'abbr', 'acronym', 'b', 'blockquote', 'br', 'code', 'em', 'i',
+    'li', 'ol', 'p', 'strong', 'ul',
+]
+
+ALLOWED_ATTRIBUTES = {
+    'a': ['href', 'title', 'target'],
+    'abbr': ['title'],
+    'acronym': ['title'],
+}
+
 # Contributors to exclude from Top Contributors list
 EXCLUDE = os.environ.get('EXCLUDE', '').split(',')
 
--- a/pontoon/teams/static/css/team.css
+++ b/pontoon/teams/static/css/team.css
@@ -123,12 +123,43 @@
   color: #7BC876;
 }
 
-.edit-info {
+#info-wrapper .edit-info {
   float: right;
 }
 
-.editing-info {
+#info-wrapper .edit-info .fa {
+  padding-right: 5px;
+}
+
+#info-wrapper .read-write-info textarea {
+  background: #333941;
+  border: 1px solid #4D5967;
+  border-radius: 3px;
+  color: #FFFFFF;
+  font-weight: 300;
+  margin-bottom: 10px;
+  padding: 5px;
   width: 100%;
+
+  -moz-box-sizing: border-box;
+  box-sizing: border-box;
+}
+
+#info-wrapper .read-write-info .toolbar .subtitle {
+  color: #AAA;
+  float: left;
+  text-transform: uppercase;
+}
+
+#info-wrapper .controls {
+  margin: 0;
+  text-align: right;
+}
+
+#info-wrapper .controls .cancel {
+  color: #7BC876;
+  margin: 9px;
+  text-transform: uppercase;
 }
 
 #permissions-form .selector-wrapper {
--- a/pontoon/teams/static/css/team.css
+++ b/pontoon/teams/static/css/team.css
@@ -123,6 +123,14 @@
   color: #7BC876;
 }
 
+.edit-info {
+  float: right;
+}
+
+.editing-info {
+  width: 100%;
+}
+
 #permissions-form .selector-wrapper {
   white-space: nowrap;
 }
--- a/pontoon/teams/static/js/info.js
+++ b/pontoon/teams/static/js/info.js
@@ -6,23 +6,23 @@ $(function() {
     e.preventDefault();
     var infoBox = container.find('.info');
     var content = infoBox.html();
-    var textArea = $('.read-write-info textarea').val($.trim(content));
-    $('.edit-info').hide();
-    $('.read-only-info').hide();
-    $('.read-write-info').toggleClass('hidden');
+    var textArea = container.find('.read-write-info textarea').val($.trim(content));
+    container.find('.edit-info').hide();
+    container.find('.read-only-info').hide();
+    container.find('.read-write-info').toggleClass('hidden');
     textArea.focus();
   });
 
   container.on('click', '.cancel', function(e) {
     e.preventDefault();
-    $('.edit-info').show();
-    $('.read-only-info').show();
-    $('.read-write-info').toggleClass('hidden');
+    container.find('.edit-info').show();
+    container.find('.read-only-info').show();
+    container.find('.read-write-info').toggleClass('hidden');
     return false;
   });
 
   container.on('click', '.save', function(e) {
-    var textArea = $('.read-write-info textarea');
+    var textArea = container.find('.read-write-info textarea');
     var content = textArea.val();
     $.ajax({
       url: textArea.parent().data('url'),
@@ -33,15 +33,15 @@ $(function() {
       },
       success: function(data) {
         container.find('.info').html(data);
-        $('.read-write-info').toggleClass('hidden');
-        $('.edit-info').show();
-        $('.read-only-info').show();
+        container.find('.read-write-info').toggleClass('hidden');
+        container.find('.edit-info').show();
+        container.find('.read-only-info').show();
         Pontoon.endLoader('Team info saved.');
       },
       error: function(request) {
         Pontoon.endLoader(request.responseText, 'error');
       }
-    })
+    });
     return false;
   });
 
--- a/pontoon/teams/static/js/info.js
+++ b/pontoon/teams/static/js/info.js
@@ -4,35 +4,45 @@ $(function() {
 
   container.on('click', '.edit-info', function(e) {
     e.preventDefault();
-    var infobox = $(".info");
-    var content = infobox.html();
-    var text_area = $(".editable-info textarea").val(content);
-    $(".edit-info").hide();
-    $(".read-only-info").hide();
-    $(".editable-info").toggleClass("hidden");
-    text_area.focus();
+    var infoBox = container.find('.info');
+    var content = infoBox.html();
+    var textArea = $('.read-write-info textarea').val($.trim(content));
+    $('.edit-info').hide();
+    $('.read-only-info').hide();
+    $('.read-write-info').toggleClass('hidden');
+    textArea.focus();
+  });
+
+  container.on('click', '.cancel', function(e) {
+    e.preventDefault();
+    $('.edit-info').show();
+    $('.read-only-info').show();
+    $('.read-write-info').toggleClass('hidden');
     return false;
   });
-  container.on('click', '.save.info', function(e) {
-    var text_area = $(".editable-info textarea");
-    var content = text_area.val();
+
+  container.on('click', '.save', function(e) {
+    var textArea = $('.read-write-info textarea');
+    var content = textArea.val();
     $.ajax({
-      url: text_area.parent().data('url'),
-      method: 'POST',
-      beforeSend: function(xhr) {
-        xhr.setRequestHeader("X-CSRFToken", text_area.parent().data('csrf'));
+      url: textArea.parent().data('url'),
+      type: 'POST',
+      data: {
+        csrfmiddlewaretoken: textArea.parent().data('csrf'),
+        team_info: content
       },
-      data: content,
-      dataType: "text",
+      success: function(data) {
+        container.find('.info').html(data);
+        $('.read-write-info').toggleClass('hidden');
+        $('.edit-info').show();
+        $('.read-only-info').show();
+        Pontoon.endLoader('Team info saved.');
+      },
+      error: function(request) {
+        Pontoon.endLoader(request.responseText, 'error');
+      }
     })
-      .done(function(jqXHR) {
-        $(".info").html(jqXHR.responseText);
-        $(".editable-info").toggleClass("hidden");
-        $(".edit-info").show();
-        $(".read-only-info").show();
-      })
-      .fail(function(jqXHR) {
-        $(".editable-info textarea").val(jqXHR.responseText);
-      });
+    return false;
   });
+
 });
--- a/pontoon/teams/static/js/info.js
+++ b/pontoon/teams/static/js/info.js
@@ -1,0 +1,38 @@
+/* global $ */
+$(function() {
+  var container = $('#main .container');
+
+  container.on('click', '.edit-info', function(e) {
+    e.preventDefault();
+    var infobox = $(".info");
+    var content = infobox.html();
+    var text_area = $(".editable-info textarea").val(content);
+    $(".edit-info").hide();
+    $(".read-only-info").hide();
+    $(".editable-info").toggleClass("hidden");
+    text_area.focus();
+    return false;
+  });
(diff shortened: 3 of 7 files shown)


pontoon/teams/static/css/team.css
```diff
@@ -123,6 +123,27 @@
...
+}
+
+.controls {
+  text-align: right;
+}
+
+.controls .cancel {
+  margin: 9px;
+}
+
```
mathjazz: These selectors are not specific enough. Since dashboards are singe page apps, these rules break the UI on other tabs. Which also means we should wrap the template into something like `#info-wrapper`.

pontoon/teams/static/js/info.js
```diff
@@ -0,0 +1,46 @@
...
+    e.preventDefault();
+    var infobox = $(".info");
+    var content = infobox.html();
+    var text_area = $(".editable-info textarea").val($.trim(content));
+    $(".edit-info").hide();
+    $(".read-only-info").hide();
+    $(".editable-info").toggleClass("hidden");
+    text_area.focus();
+    return false;
+  });
```
mathjazz: Add blank lines between the top-level blocks.

pontoon/teams/static/js/info.js
```diff
@@ -0,0 +1,46 @@
+/* global $ */
+$(function() {
+  var container = $('#main .container');
+
+  container.on('click', '.edit-info', function(e) {
+    e.preventDefault();
+    var infobox = $(".info");
+    var content = infobox.html();
+    var text_area = $(".editable-info textarea").val($.trim(content));
```
mathjazz: Please use camelCase in JS - `var textArea`.

pontoon/teams/static/js/info.js
```diff
@@ -0,0 +1,46 @@
...
+    $(".edit-info").hide();
+    $(".read-only-info").hide();
+    $(".editable-info").toggleClass("hidden");
+    text_area.focus();
+    return false;
```
mathjazz: Why should we `return false` here?

pontoon/teams/static/js/info.js
```diff
@@ -0,0 +1,46 @@
+/* global $ */
+$(function() {
+  var container = $('#main .container');
+
+  container.on('click', '.edit-info', function(e) {
+    e.preventDefault();
+    var infobox = $(".info");
```
mathjazz: This selector (and the ones that follow) is too general. It's too dangerous too assume we'll never add elements to any of the other tabs or the page heading with such class name. Please use `container.find('.info')` instead.

And, try to use single quotes in selectors.

pontoon/teams/static/js/info.js
```diff
@@ -0,0 +1,46 @@
...
+      data: {
+        csrfmiddlewaretoken: text_area.parent().data('csrf'),
+        team_info: content
+      },
+      success: function(data) {
+        $(".info").html(data) ;
+        $(".editable-info").toggleClass("hidden");
+        $(".edit-info").show();
+        $(".read-only-info").show();
+        Pontoon.endLoader('Team info saved!');
```
mathjazz: Please use period instead of exclamation mark.

pontoon/teams/static/js/info.js
```diff
@@ -0,0 +1,46 @@
...
+      },
+      error: function(request) {
+        $(".editable-info textarea").val(request.responseText);
```
mathjazz: Why should we put the error message into textarea?

pontoon/teams/static/css/team.css
```diff
@@ -123,6 +123,27 @@
   color: #7BC876;
 }
 
+.info {
+  text-align: center;
```
mathjazz: We should keep the default alignment (left). That's what users expect by default. And that's how we currently align team and project information blocks.

pontoon/teams/templates/teams/includes/info.html
```diff
@@ -1,5 +1,25 @@
...
+      <p>No team info available.</p>
+    {% endif %}
+  </div>
+</div>
+{% if request.user.has_perm('base.can_manage_locale', locale) %}
+<div class="editable-info hidden"
+     data-url="{{ url('pontoon.teams.ajax.update-info', locale.code) }}"
+     {% if csrf_token %}data-csrf="{{ csrf_token }}"{% endif %}>
+
+  <textarea class="editing-info" rows="20"/>
```
mathjazz: 10 lines should be enough. And then users can resize.

pontoon/teams/views.py
```diff
@@ -77,6 +78,25 @@ def ajax_info(request, locale):
     })
 
 
+@require_POST
+@permission_required_or_403('base.can_manage_locale', (Locale, 'code', 'locale'))
+@transaction.atomic
+def ajax_update_info(request, locale):
+    description = request.POST.get('team_info', None)
+    tags = bleach.ALLOWED_TAGS[:]
+    # allow <p>
```
mathjazz: Start comments with the capital letter. We also allow `<br>`.

I'd add blank lines above the `tags =` and `attrs =` lines.

pontoon/teams/views.py
```diff
@@ -77,6 +78,25 @@ def ajax_info(request, locale):
     })
 
 
+@require_POST
+@permission_required_or_403('base.can_manage_locale', (Locale, 'code', 'locale'))
+@transaction.atomic
+def ajax_update_info(request, locale):
+    description = request.POST.get('team_info', None)
```
mathjazz: The param should be called `team_description` to be consistent with the DB.

pontoon/teams/views.py
```diff
@@ -77,6 +78,25 @@ def ajax_info(request, locale):
...
+    tags = bleach.ALLOWED_TAGS[:]
+    # allow <p>
+    tags.extend(('br', 'p'))
+    attrs = bleach.ALLOWED_ATTRIBUTES.copy()
+    # allow <a target="">
+    attrs['a'].append('target')
+    description = bleach.clean(
+        description, tags=tags, attributes=attrs, strip=True)
+    _locale = get_object_or_404(Locale, code=locale)
```
mathjazz: I'm not sure what's the best practice for naming variables like this. I'm not a big fan of the `_` prefix. I'd probably use `l`, but that's ugly either.

mastizada:
Thanks @mathjazz,
The only solution for duplicate new lines that I could find is to replace all new lines with `<br>`. That makes text to look ugly in the editor.

In updating info text after save have 1 issue: In the template there is `<p class="no-results">` if there is no team info, and `<div class="info">` is not added to the template at all. My previous solution for that was to use `<div class="info">` for both and adding text using `if else` block inside the `<div class="info">` element. I implemented that but patch reverted again.

Will be happy to work on new bug for bleach too :)

mathjazz:
> The only solution for duplicate new lines that I could find is to replace all new lines with <br>. That makes text to look ugly in the editor.

That work better now. But it still adds additional lines if you simply edit + save existing text that contain newlines and <br>, e.g. [cs](https://pontoon.mozilla.org/cs/info/). But there aren't many cases like this, so we could manually update them in the DB.

What if we simply ignore new lines? That way the users would need to use HTML tags to add new lines, either the BR or P tag.

> In updating info text after save have 1 issue: In the template there is <p class="no-results"> if there is no team info, and <div class="info"> is not added to the template at all. My previous solution for that was to use <div class="info"> for both and adding text using if else block inside the <div class="info"> element. I implemented that but patch reverted again.

Argh, sorry for the mess. Could you add those elements via JS? Or reverse this part of my patch?

> Will be happy to work on new bug for bleach too :)

Great! Filed: https://bugzilla.mozilla.org/show_bug.cgi?id=1386166

pontoon/teams/static/js/info.js
```diff
@@ -0,0 +1,48 @@
...
+    e.preventDefault();
+    $('.edit-info').show();
+    $('.read-only-info').show();
+    $('.read-write-info').toggleClass('hidden');
+    return false;
+  });
+
+  container.on('click', '.save', function(e) {
+    var textArea = $('.read-write-info textarea');
+    var content = textArea.val().replace(/(?:\r\n|\r|\n)/g, '<br>');
```
adngdb: Why not just remove those newline characters if they're at the end of the string?

pontoon/teams/views.py
```diff
@@ -77,6 +78,27 @@ def ajax_info(request, locale):
...
 
 
+@require_POST
+@permission_required_or_403('base.can_manage_locale', (Locale, 'code', 'locale'))
+@transaction.atomic
+def ajax_update_info(request, locale):
+    team_description = request.POST.get('team_info', None)
+
+    # Allow <p> and <br> tags
+    tags = bleach.ALLOWED_TAGS[:]
```
adngdb: I assume you want to make a copy of ``ALLOWED_TAGS`` here. It is generally more readable to do ``list(bleach.ALLOWED_TAGS)`` instead. The usage of ``[:]`` is quite cryptic and considered bad practice.

pontoon/teams/views.py
```diff
@@ -77,6 +78,27 @@ def ajax_info(request, locale):
...
+    tags.extend(('br', 'p'))
+
+    # Allow <a target=""> attribute
+    attrs = bleach.ALLOWED_ATTRIBUTES.copy()
+    attrs['a'].append('target')
+    team_description = bleach.clean(
+        team_description, tags=tags, attributes=attrs, strip=True)
```
adngdb: Please put the closing parenthesis on a new line.

mastizada:
@mathjazz now showing result works if there is no team info.

As you suggested I removed replace statement at all, that makes sense to allow users decide when to add `<br>`.

mastizada:
@adngdb Thanks for review. Implemented your suggestions, can you review again?

adngdb:
@mastizada when making changes to a branch during a review, please add those changes in a new commit. This way it is easier for reviewers to see what changed since the last review. (Right now, it is near impossible to see what you have changed, and thus I have to read the entire code change again. )

pontoon/teams/views.py
```diff
@@ -77,6 +78,28 @@ def ajax_info(request, locale):
...
 
 
+@require_POST
+@permission_required_or_403('base.can_manage_locale', (Locale, 'code', 'locale'))
+@transaction.atomic
+def ajax_update_info(request, locale):
+    team_description = request.POST.get('team_info', None)
+
+    # Allow <p> and <br> tags
+    tags = bleach.ALLOWED_TAGS
```
adngdb: Why did you remove the copy altogether? I thought it was a pretty good idea to not edit bleach's variables, to avoid conflicting with other parts of the code that might use it in the future.

pontoon/teams/static/js/info.js
```diff
@@ -0,0 +1,48 @@
...
+        container.find('.info').html(data);
+        $('.read-write-info').toggleClass('hidden');
+        $('.edit-info').show();
+        $('.read-only-info').show();
+        Pontoon.endLoader('Team info saved.');
+      },
+      error: function(request) {
+        Pontoon.endLoader(request.responseText, 'error');
+      }
+    })
```
mathjazz: Add missing semicolon.

pontoon/teams/static/js/info.js
```diff
@@ -0,0 +1,48 @@
+/* global $ */
+$(function() {
+  var container = $('#main .container');
+
+  container.on('click', '.edit-info', function(e) {
+    e.preventDefault();
+    var infoBox = container.find('.info');
+    var content = infoBox.html();
+    var textArea = $('.read-write-info textarea').val($.trim(content));
+    $('.edit-info').hide();
```
mathjazz: We should change all `$('.className')` selectors to `container.find('.className')`. This script is active on all team page tabs and it's too dangerous to assume that we will never add an element with one of these class names to any of the tabs.

pontoon/settings/base.py
```diff
@@ -728,6 +728,18 @@ def _allowed_hosts():
     'vi-vn', 'wo-sn', 'xh-za', 'yo-ng', 'zh-cn', 'zh-hk', 'zh-tw', 'zu-za',
 ]
 
+# Bleach tags and attributes
+ALLOWED_TAGS = [
+    'a', 'abbr', 'acronym', 'b', 'blockquote', 'br', 'code', 'em', 'i',
+    'li', 'ol', 'p', 'strong', 'ul',
+]
```
mathjazz: Why not simply extend whatever `bleach` decides is safe? Same for `ALLOWED_ATTRIBUTES`.

```python
ALLOWED_TAGS = list(bleach.ALLOWED_TAGS).extend(['br', 'p'])
```

pontoon/teams/static/js/info.js
```diff
@@ -0,0 +1,48 @@
...
+    container.find('.read-write-info').toggleClass('hidden');
+    textArea.focus();
+  });
+
+  container.on('click', '.cancel', function(e) {
+    e.preventDefault();
+    container.find('.edit-info').show();
+    container.find('.read-only-info').show();
+    container.find('.read-write-info').toggleClass('hidden');
+    return false;
```
mathjazz: `return false` within a jQuery event handler is the same as calling both `e.preventDefault` and `e.stopPropagation`. Why do we need to do this when the handler finishes executing its code? Same goes for the `.save` button handler.

pontoon/teams/static/js/info.js
```diff
@@ -0,0 +1,48 @@
...
+  });
+
+  container.on('click', '.save', function(e) {
+    var textArea = container.find('.read-write-info textarea');
+    var content = textArea.val();
```
mathjazz: You can also pass this directly to the `team_info` parameter, without defining a variable.

pontoon/teams/static/js/info.js
```diff
@@ -0,0 +1,48 @@
+/* global $ */
+$(function() {
+  var container = $('#main .container');
+
+  container.on('click', '.edit-info', function(e) {
+    e.preventDefault();
+    var infoBox = container.find('.info');
+    var content = infoBox.html();
```
mathjazz: You could also do `var content = container.find('.info').html();` and avoid defining a variable.

pontoon/teams/static/js/info.js
```diff
@@ -0,0 +1,48 @@
...
+    var textArea = container.find('.read-write-info textarea').val($.trim(content));
+    container.find('.edit-info').hide();
+    container.find('.read-only-info').hide();
+    container.find('.read-write-info').toggleClass('hidden');
```
mathjazz: You can replace `.hide()` and `.show()` with `.toggle()` to get the same effect. Then you have the pattern of these three lines repeating 3 times. You can turn it into a function, e.g.:

```js
function toggleWidgets() {
  container
    .find('.edit-info').toggle().end()
    .find('.read-only-info').toggle().end()
    .find('.read-write-info').toggleClass('hidden');
}
```

pontoon/teams/templates/teams/includes/info.html
```diff
@@ -1,5 +1,36 @@
...
+    {% endif %}
+    </div>
+  </div>
+
+  {% if request.user.has_perm('base.can_manage_locale', locale) %}
+  <div class="read-write-info hidden"
+       data-url="{{ url('pontoon.teams.ajax.update-info', locale.code) }}"
+       {% if csrf_token %}data-csrf="{{ csrf_token }}"{% endif %}>
+
+    <textarea rows="10" />
```
mathjazz: Add blank line.

pontoon/teams/templates/teams/includes/info.html
```diff
@@ -1,5 +1,36 @@
...
+
+  <div class="read-only-info">
+    <div class="info">
+    {% if locale.team_description %}
+      {{ locale.team_description|safe }}
+    {% else %}
+      <p class="no-results">No team info available.</p>
+    {% endif %}
+    </div>
+  </div>
```
mathjazz: There are 2 issues with locales without `team_description`:
* If you edit info for such locale, you get `<p class="no-results">No team info available.</p>` in the textarea.
* If you save empty team info, the `No team info available.` text doesn't show up.

I suggest that you keep using the same HTML as we do for project info, but keep both HTML elements in the DOM:
```jinja
<div class="info">{{ locale.team_description|safe }}</div>
<p class="no-results {% if locale.team_description %}hidden{% endif %}">No team info available.</p>
```
Then in the `success` function of the `$.ajax call in JS toggle both elements based on the returned data:
```js
container.find('.info').html(data).toggle(data !== '');
container.find('.no-results').toggle(data === '');
```

pontoon/settings/base.py
adngdb: +1

mastizada: I wanted to have better control for allowed tags and attributes as we will use same for other inputs (project info for ex.). It will also help to avoid changes in bleach libraries defaults. Also having complete list of all tags and attributes in settings makes it easy to control them.

mathjazz: OK, let's keep this explicit. It should help with figuring out which elements we need to take care of when styling those outputs (I tested all of them and they look OKish).

PR #656, opened 2017-07-30
 Fix bug 1386193: Use six library for python 3 conflicting methods 	

mastizada:


mastizada:
@mathjazz @jotes Should I rebase all commits to 1?

jotes:
@mastizada During the review process it's better to keep commits separated because it's easier to follow what changed between comments. However, that's required to squash all commits into 1 before PR is merged.

Pike:
FWIW, in this particular case, I'd possibly keep the commits unsquashed.

That said, there are a few minor details where things are in the wrong commit, I think.

mastizada:
@jotes @Pike 1 unicode reference was replaced in urllib commit and urlparse was replaced in last commit as Pike mentioned. Also pylama gave error due to conflicting urlencode functions in helper.py, replaced it with `urllib_urlencode` and squashed commits.

mathjazz:
@adngdb Could you review this one please?

adngdb:
A few things before I start the actual review: 

* is there a bug associated with this? If yes, please add a reference to that bug in your commit (see the [contributing guide](https://mozilla-pontoon.readthedocs.io/en/latest/dev/contributing.html)), otherwise, please file a bug (and add a reference to it).
* there is a conflict on ``pontoon/base/models.py``
* is the goal here to make the entire Pontoon app work with python 3? Or is it a step towards that goal?

pontoon/base/models.py
```diff
@@ -1,11 +1,15 @@
+from __future__ import unicode_literals
 from __future__ import division
+from django.utils.encoding import python_2_unicode_compatible
```
adngdb: This should go with the other ``django`` related imports. Note that this is how we usually sort imports: 

0. ``__future__``
1. standard library
2. dependencies
3. framework
4. local code

And then we sort them by alphabetical order inside each section.

pontoon/base/models.py
```diff
@@ -1,11 +1,15 @@
...
 
 import hashlib
 import logging
 import math
 import os.path
 import re
-import urllib
+from six.moves.urllib.parse import urlencode
+from six.moves import reduce
+from six.moves.urllib.parse import urlparse
```
adngdb: Please rewrite this to:

```python
from six.moves import reduce
from six.moves.urllib.parse import (urlencode, urlparse)
```

and move it to after ``from dirtyfields import DirtyFieldsMixin`` (as ``six`` is a dependency and not from the standard library).

pontoon/base/templatetags/helpers.py
```diff
@@ -1,8 +1,10 @@
 import cgi
 import datetime
 import json
-import urllib
-import urlparse
```
adngdb: Please add a new line here to separate standard and dependency imports.

pontoon/base/templatetags/helpers.py
```diff
@@ -1,8 +1,10 @@
...
+from six.moves.urllib.parse import urlparse, parse_qsl
+from six.moves.urllib.parse import quote_plus, ParseResult
+from six.moves.urllib.parse import urlencode as urllib_urlencode
+from six import text_type
```
adngdb: Please rewrite to: 

```python
from six import text_type
from six.moves.urllib import parse
```

And then use ``parse.xxx`` to replace the various ``urlparse.xxx`` and ``urllib.xxx``. It is generally a better practice to know what package a method comes from, and numerous imports like this should be avoided. The cost of importing the entire module instead is very low.

Note: you can rename ``parse`` to something better if you want. I admit that it is not a very good name and it makes readability harder. Maybe ``from six.moves.urllib import parse as six_parse`` is better?

pontoon/contributors/tests.py
```diff
@@ -1,4 +1,5 @@
 from collections import OrderedDict
+from six.moves import range
```
adngdb: This should be after ``datetime`` in the alphabetical order of imports. :)

pontoon/sync/formats/xliff.py
```diff
@@ -2,6 +2,7 @@
 Parser for the xliff translation format.
 """
 from translate.storage import xliff
+from six import text_type
```
adngdb: This should be before ``translate``.

pontoon/sync/vcs/repositories.py
```diff
@@ -4,6 +4,7 @@
 import os
 import scandir
 import subprocess
+from six import text_type
```
adngdb: Please add a new line before this.

mastizada:
@adngdb thanks again. I had no idea for import sort in pontoon, that will help in future.

pontoon/base/models.py
mathjazz: @adngdb Should be put this into the [CONTRIBUTING](https://github.com/mozilla/pontoon/blob/master/CONTRIBUTING.rst) file?

pontoon/base/templatetags/helpers.py
```diff
@@ -1,8 +1,9 @@
 import cgi
 import datetime
 import json
-import urllib
-import urlparse
+
+from six import text_type
+from six.moves.urllib import parse as urlparse
```
adngdb: I find this a bit worrying, because there is a package that is called ``urlparse``. We might end up with people thinking it's the actual library and try to do things with it that are not supported by ``six.moves.urllib.parse``. Can you name it something different that conveys the fact that it's from the ``six`` library?

mastizada: urlparse module is removed in python3 but yes, it can mislead contributors. Used six_parse as you suggested before.

mathjazz: @adngdb Are there really such things in `urlparse` that are not supported by [`six.moves.urllib.parse`](https://pythonhosted.org/six/#module-six.moves.urllib.parse)? Wouldn't people notice they break if they use them anyways?

Anyhow, if you decide to use `six_parse`, then we should do this across the entire project, not just in this file.

mastizada: Six's documentation shows that all methods that were in python 2 `urlparse` are supported in `six.moves.urllib.parse` and also there is no such library as urlparse in python 3 (became `urllib.parse`). I think it will be fine to use urlparse name.

adngdb: Fair enough, I didn't know everything was supported in ``six``.

PR #659, opened 2017-08-02
Fix bug 1386632 - Update python packages rsa and oauthlib.

adngdb:
@mathjazz I'm not too sure how to test this. Unit tests pass, but I'm afraid the ``oauthlib`` update might break stuff.


mathjazz:
I'm not exactly sure if we still need the two requirements.

Seems like they were [added](https://github.com/mozilla/pontoon/commit/ae7b918f8d026d74640583fa0e51613484da36e3) as `peep` dependencies, which we no longer use.

I'll test a little further.

adngdb:
So ``oauthlib`` is required by ``django-oauth`` which is required by ``django-allauth``, which we notably use for the Firefox Account login. ``rsa``, however, seems to be unused, so I'm going to remove it.

PR #660, opened 2017-08-03
Fix team edit

mathjazz:
Addressing issues reported by @MikkCZ.

* Moving all buttons to the top to avoid overflowing text. It should also be handier than catching buttons on unexpected places.
* Also enabling team info editing on the localization pages (/cs/pontoon-intro/)


pontoon/teams/static/css/team.css
```diff
@@ -123,45 +123,6 @@
...
-  margin: 0;
-  text-align: right;
-}
-
-#info-wrapper .controls .cancel {
-  color: #7BC876;
-  margin: 9px;
-  text-transform: uppercase;
-}
-
```
adngdb: Yay to splitting up files! :)

PR #662, opened 2017-08-03
Fix bug 1374050: Change logo to SVG

MekliCZ:


mathjazz:
Thanks for the patch! r+

PR #663, opened 2017-08-05
Fix bug 1386166: clean html input for notifications and project info

mastizada:


pontoon/projects/forms.py
```diff
@@ -1,7 +1,18 @@
...
 from django.core import validators
+from django.conf import settings
 
 
 class NotificationsForm(forms.Form):
     message = forms.CharField(widget=forms.Textarea)
     selected_locales = forms.CharField(validators=[validators.validate_comma_separated_integer_list])
+
+    def clean_message(self):
+        message = self.cleaned_data['message']
```
jotes: I'd move that code in the validator/django form field class and add tests for views with input sanitization.

mastizada: Bleach can only clean, it doesn't have any exceptions (except missing settings). The only validation that we can implement is to make cleaned text and compare to user input and if the input do not match, then raise exception. Also bleach do not provide which not-allowed tags were user so we will need to find difference and use that for validation error message.

jotes:
Hey, then we can move this logic into a form field class.


6 sie 2017 4:34 PM "Emin Mastizada" <notifications@github.com> napisał(a):

mastizada:
@mathjazz @jotes why not to use TextArea widget for administration form too? Or everything related to administration will be in another PR?

mathjazz:
Changes to Pontoon Admin (/admin) should be limited to maintenance only, because we'll drop it.

pontoon/administration/forms.py
```diff
@@ -12,6 +13,7 @@ def label_from_instance(self, obj):
 
 class ProjectForm(forms.ModelForm):
     contact = ContactChoiceField(None, required=False)
+    info = HtmlField()
```
mathjazz: `info = HtmlField(widget=forms.Textarea)`

It would be even better if you could move the widget definition into the field itself, because we will always want to use textarea instead of the input field.

pontoon/base/forms.py
```diff
@@ -11,6 +14,16 @@
...
 
 
+class HtmlField(forms.CharField):
+    def clean(self, value):
+        value = super(HtmlField, self).clean(value)
+        bleach.clean(
+            value, strip=True,
+            tags=settings.ALLOWED_TAGS, attributes=settings.ALLOWED_ATTRIBUTES
+        )
+        return value
```
mathjazz: This won't work, because you didn't set the value to what bleach returns.

mastizada: sorry, that was stupid mistake :D

pontoon/administration/forms.py
mastizada: used default widget, tested in pontoon/admin

PR #669, opened 2017-08-10
Fix bug 1389084: Add Rejected Suggestions filter

mathjazz:
Let's make it easier to find rejected suggestions.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -1557,7 +1557,7 @@ class EntityQuerySet(models.QuerySet):
     Queryset provides a set of additional methods that should allow us to filter entities.
     """
     user_entity_filters = ('missing', 'fuzzy', 'suggested', 'translated')
-    user_extra_filters = ('has-suggestions', 'unchanged')
+    user_extra_filters = ('has-suggestions', 'rejected', 'unchanged')
 
     def filter_statuses(self, locale, statuses):
         sanitized_statuses = filter(lambda s: s in self.user_entity_filters, statuses)
@@ -1576,12 +1576,16 @@ class EntityQuerySet(models.QuerySet):
         """
         Helper method that returns a set with annotation of following fields:
             * approved_count - a number of approved translations in the entity.
-            * fuzzy_count - a number of fuzzy translations in the entity
-            * suggested_count - a number of translations assigned do the entity.
-            * unreviewed_count - a number of translations that have not yet been
-              approved or rejected in the entity.
-            * expected_count - a number of translations that should cover entity.
-            * unchanged_count - a number of translations that have the same string as the entity.
+            * fuzzy_count - a number of fuzzy translations in the entity.
+            * suggested_count - a number of suggested translations in the
+              entity.
+            * unreviewed_count - a number of translations in the entity that
+              have not yet been approved or rejected.
+            * rejected_count - a number of rejected translations in the entity.
+            * expected_count - a number of translations that should cover
+              entity.
+            * unchanged_count - a number of translations that have the same
+              string as the entity.
         """
         return self.annotate(
             approved_count=Sum(
@@ -1627,6 +1631,21 @@ class EntityQuerySet(models.QuerySet):
                     default=0
                 )
             ),
+            rejected_count=Sum(
+                Case(
+                    When(
+                        Q(
+                            translation__locale=locale,
+                            translation__approved=False,
+                            translation__fuzzy=False,
+                            translation__rejected=True,
+                        ),
+                        then=1
+                    ),
+                    output_field=models.IntegerField(),
+                    default=0
+                )
+            ),
             expected_count=Case(
                 When(
                     Q(string_plural__isnull=True) | Q(string_plural=""), then=1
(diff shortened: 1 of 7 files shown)


pontoon/base/models.py
```diff
@@ -1576,12 +1576,16 @@ def with_status_counts(self, locale):
...
+            * fuzzy_count - a number of fuzzy translations in the entity.
+            * suggested_count - a number of suggested translations in the
+              entity.
+            * unreviewed_count - a number of translations in the entity that
+              have not yet been approved or rejected.
+            * rejected_count - a number of rejected translations in the entity.
+            * expected_count - a number of translations that should cover
+              entity.
+            * unchanged_count - a number of translations that have the same
+              string as the entity.
```
mathjazz: Reworded a little.

pontoon/base/static/css/style.css
```diff
@@ -251,21 +251,6 @@ button {  width: auto; overflow: visible; }
...
-.has-suggestions .status.fa:before {
-  color: #AAAAAA;
-  content: "";
-}
-
-.unchanged .status.fa:before {
-  color: #AAAAAA;
-  content: "";
-}
-
```
mathjazz: I don't think we need these in the general CSS file. They are already defined in the translate.css.

pontoon/base/static/js/translate.js
```diff
@@ -1341,6 +1341,13 @@ var Pontoon = (function (my) {
     attachEntityListHandlers: function() {
       var self = this;
 
+      function isExtraFilter(el) {
+        return el.hasClass('untranslated') ||
+               el.hasClass('unchanged') ||
+               el.hasClass('has-suggestions') ||
+               el.hasClass('rejected');
+      }
+
```
mathjazz: Let's make it reusable.

pontoon/base/models.py
```diff
@@ -1627,6 +1631,21 @@ def with_status_counts(self, locale):
...
                 )
             ),
+            rejected_count=Sum(
+                Case(
+                    When(
+                        Q(
+                            translation__locale=locale,
+                            translation__approved=False,
+                            translation__fuzzy=False,
+                            translation__rejected=True,
```
adngdb: Considering that when ``rejected`` is true, none of ``approved`` or ``fuzzy`` can be true, I think you can avoid the previous 2 filters. Not sure how big an impact on performance that would have though.

pontoon/base/templates/translate.html
```diff
@@ -174,6 +174,7 @@
             {{ Filter.item('untranslated', 'Untranslated') }}
             {{ Filter.item('unchanged', 'Unchanged') }}
             {{ Filter.item('has-suggestions', 'Unreviewed Suggestions') }}
+            {{ Filter.item('rejected', 'Rejected Suggestions') }}
```
adngdb: Why add ``Suggestions`` here? I'm not sure I understand what you mean by ``suggestions``, but to me it's not a suggestion unless it's neither approved nor rejected. Does that make sense?

pontoon/base/tests/test_models.py
```diff
@@ -1542,41 +1542,85 @@ def test_has_suggestions_plural(self):
...
-            plural_form=2,
+            plural_form=0,
         )
         TranslationFactory.create(
             locale=self.plural_locale,
             entity=third_entity,
             approved=True,
             fuzzy=False,
-            plural_form=0,
+            plural_form=1
```
adngdb: Let's keep trailing comas, they make subsequent diffs easier to read if we're ever to add a new argument to this function.

pontoon/base/tests/test_models.py
```diff
@@ -1542,41 +1542,85 @@ def test_has_suggestions_plural(self):
...
+        )
+        assert_equal(
+            {first_entity, third_entity},
+            set(Entity.objects.with_status_counts(self.plural_locale).filter(Entity.objects.has_suggestions()))
+        )
+
+    def test_rejected_plural(self):
+        first_entity, second_entity, third_entity = PluralEntityFactory.create_batch(3,
```
adngdb: This is a PEP8 violation: the first argument should be on a new line.

pontoon/base/tests/test_models.py
```diff
@@ -1542,41 +1542,85 @@ def test_has_suggestions_plural(self):
...
+            string_plural='Unchanged plural string'
         )
         TranslationFactory.create(
             locale=self.plural_locale,
-            entity=third_entity,
+            entity=first_entity,
+            approved=True,
+            fuzzy=False,
+            rejected=False,
+            plural_form=0
```
adngdb: Trailing coma please. :)

pontoon/base/tests/test_views.py
```diff
@@ -306,7 +307,11 @@ def test_entity_filters(self):
                 'limit': 1,
             }
 
-            if filter_ == 'unchanged' or filter_ == 'has-suggestions':
+            if (
+                filter_ == 'unchanged' or
```
jotes: `filter_ in ('unchanged', 'rejected'...)` ?

pontoon/base/models.py
mathjazz: I didn't notice any impact on performance. Since we don't have any DB constraints, I'd rather stay on the safe side and keep all the checks.

pontoon/base/templates/translate.html
mathjazz: Good point. We also don't say *Approved Suggestions*.

I think we should still keep *Unreviewed Suggestions*, though. For easier distinction with *Suggested*. Which will get rid of as part of bug [1377969](https://bugzilla.mozilla.org/show_bug.cgi?id=1377969). And then we start calling them just *Suggestions*.

PR #671, opened 2017-08-11
Consolidate developer docs

mathjazz:
1. Due to overlap, the following pages have been merged:
* Contributing
* Workflow
* Integration with fluent

2. The hashin package to requirements-dev.txt.

3. README has been updated with information that the listed deployments
are used for the localization of Mozilla projects only.

@jotes r?


jotes:
:+1: for  hashin

CONTRIBUTING.rst
```diff
@@ -193,3 +199,48 @@ different depending on the context.
...
+.. code-block:: shell
+
+   # Pull latest code (assuming you've already checked out master).
+   git pull origin master
+
+   # Install new dependencies or update existing ones.
+   pip install -U --force --require-hashes -r requirements.txt
+
+   # Run database migrations.
+   python manage.py migrate
```
adngdb: Those instructions will not work for someone using ``docker``. I suggest we move them to the install docs instead, and add an equivalent to the install with docker doc as well.

CONTRIBUTING.rst
```diff
@@ -193,3 +199,48 @@ different depending on the context.
...
+* python-fluent (responsible for e.g. server-side sync process)
+* fluent-syntax (required by the fluent editor)
+
+How to build the fresh version of fluent-syntax.js
+--------------------------------------------------
+
+.. code-block:: shell
+
+    npm install fluent-syntax
+    cp node_modules/fluent-syntax/compat.js pontoon/base/static/js/lib/fluent-syntax.js
```
adngdb: This is ugly and will probably push me harder towards rewriting dependency management using npm. :)

jotes: or yarn. And I totally agree -> however, changes in js stack are tedious for me ;-)

CONTRIBUTING.rst
mathjazz: Right. Since the rest of this document assumes the docker environment (see Tests section for example), we could also put docker docs here. On the long term we probably want to document developer setup with docker only.

PR #677, opened 2017-08-14
Fix bug 1390111: Unescape \u0027 & \u0022 in Android

mathjazz:


pontoon/sync/tests/formats/test_silme.py
```diff
@@ -221,8 +225,10 @@ def test_quotes(self):
         expected_string = dedent("""
             <!ENTITY SingleQuote "Single Quote \&apos;">
             <!ENTITY SingleQuoteEntity "Double Quote \&quot;">
+            <!ENTITY SingleQuotePython "\&apos;">
             <!ENTITY DoubleQuote '\&quot;'>
```
Pike: The 'Double Quote "' should be inserted to [3], not [1] these days, right?

mathjazz:
Thanks!

I addressed your comments and added a separate commit to fix bug 1390113 with a data migration.

mathjazz:
@Pike Could you also review the fix for bug 1390113?

mathjazz:
Thanks. I tested with the DB dump indeed. Everything worked as expected. Except that it was a little slow. Because `bulk_update` was running within the loop. :1st_place_medal:

Fixed and rebased.

PR #679, opened 2017-08-15
Bug 1381852: Unapprove fuzzy translations on import

mathjazz:
If an existing translation is made fuzzy in the VCS, we mark it as
fuzzy in the DB on next sync, but don't unapprove it. That is because
two separate instances of the same translation are passed to
bulk_update - the first one sets the fuzzy flag and the second one
unsets the approved flag. bulk_update only executes the first instance,
so now we properly set all flags on it and don't pass the second one.

Note: this only applies to translations that are made fuzzy in the
locale file. Most of the time a new source strings is created when
existing translation are made fuzzy on msgmerge. This case already
behaves properly.

@jotes r?

### DIFF
--- a/pontoon/sync/changeset.py
+++ b/pontoon/sync/changeset.py
@@ -48,7 +48,7 @@ class ChangeSet(object):
         }
 
         self.entities_to_update = []
-        self.translations_to_update = []
+        self.translations_to_update = set()
         self.translations_to_create = []
         self.commit_authors_per_locale = defaultdict(list)
         self.locales_to_commit = set()
@@ -218,7 +218,7 @@ class ChangeSet(object):
                 db_translation.extra = vcs_translation.extra
 
                 if db_translation.is_dirty():
-                    self.translations_to_update.append(db_translation)
+                    self.translations_to_update.add(db_translation)
                 if not db_translation.fuzzy:
                     approved_translations.append(db_translation)
                 else:
@@ -248,8 +248,8 @@ class ChangeSet(object):
                 translation.approved_user = None
                 translation.approved_date = None
 
-                if translation not in self.translations_to_update and translation.is_dirty():
-                    self.translations_to_update.append(translation)
+                if translation.is_dirty():
+                    self.translations_to_update.add(translation)
 
         # Any existing translations that are no longer fuzzy get unfuzzied.
         for translation in db_translations:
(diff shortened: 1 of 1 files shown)


pontoon/sync/changeset.py
```diff
@@ -195,6 +195,10 @@ def update_entity_translations_from_vcs(
             db_translations = db_entity.translation_set.filter(
                 locale__code=locale_code,
             )
+
+        if old_translations is None:
+            old_translations = db_translations.filter(approved_date__lte=self.now)
+
```
mathjazz: Just moving this block up, closer to `db_translations`. Doesn't have any impact.

pontoon/sync/changeset.py
```diff
@@ -234,24 +238,27 @@ def update_entity_translations_from_vcs(
...
             if translation not in approved_translations:
+                translation = (
+                    match_attr(self.translations_to_update, pk=translation.pk) or translation
+                )
                 translation.approved = False
                 translation.approved_user = None
                 translation.approved_date = None
 
-                if translation.is_dirty():
+                if translation not in self.translations_to_update and translation.is_dirty():
```
jotes: I have a non-blocking comment: If we'd change a type of translations_to_update into something dict/set like (with translation_pk as the index) that could simplify those changes.

pontoon/sync/changeset.py
```diff
@@ -234,24 +238,27 @@ def update_entity_translations_from_vcs(
...
                     self.translations_to_update.append(translation)
 
         # Any existing translations that are no longer fuzzy get unfuzzied.
         for translation in db_translations:
             if translation not in fuzzy_translations:
+                translation = (
```
jotes: I think that line is worth a comment what's going in here (bugs with references to objects are tricky to find).

mathjazz:
Thanks. Updated!

pontoon/sync/changeset.py
```diff
@@ -234,25 +238,30 @@ def update_entity_translations_from_vcs(
...
 
         # Any existing translations that were not approved get unapproved.
-        if old_translations is None:
-            old_translations = db_translations.filter(approved_date__lte=self.now)
-
         for translation in old_translations:
             if translation not in approved_translations:
+                # Use the translation instance already set for update if it exists
+                translation = (
+                    match_attr(self.translations_to_update, pk=translation.pk) or translation
```
jotes: NIT: Set is nice, but dict could reduce that line into `self.tranlations_to_update.get(translation.pk, translation)` and _probably_ could be faster than match_attr function but _relatively_ less explicit. Still, the PR lgtfm :+1:

mathjazz: Good point!

PR #680, opened 2017-08-15
Fix bug 1390047: Exclude rejected and unreviewed from TM

mathjazz:
@jotes r?

### DIFF
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -886,11 +886,10 @@ def download_translation_memory(request, locale, slug, filename):
     project = get_object_or_404(Project, slug=slug)
 
     tm_entries = (
-        TranslationMemoryEntry.objects.filter(
-            locale=locale,
-            project=project,
-            translation__isnull=False,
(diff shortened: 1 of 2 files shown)


pontoon/base/views.py
```diff
@@ -886,11 +886,10 @@ def download_translation_memory(request, locale, slug, filename):
...
     tm_entries = (
-        TranslationMemoryEntry.objects.filter(
-            locale=locale,
-            project=project,
-            translation__isnull=False,
-        ).exclude(Q(source='') | Q(target=''))
+        TranslationMemoryEntry.objects
+        .filter(locale=locale, project=project, translation__isnull=False)
+        .exclude(Q(source='') | Q(target=''))
+        .exclude(translation__approved=False, translation__fuzzy=False)
```
jotes: Is this happen for newer objects? Maybe that could be handled by a data migration?

mathjazz: See https://bugzilla.mozilla.org/show_bug.cgi?id=1390047#c1

PR #682, opened 2017-08-16
Fix bug 1390805: Always create TM entries on sync

mathjazz:
@jotes r?

See [bug 1390805](https://bugzilla.mozilla.org/show_bug.cgi?id=1390805) for more details.


pontoon/base/migrations/0100_bug_1390805_create_missing_tm_entries.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def create_missing_translaton_memory_entries(apps, schema_editor):
+    Translation = apps.get_model('base', 'Translation')
+    TranslationMemoryEntry = apps.get_model('base', 'TranslationMemoryEntry')
+
+    translations_to_create_translaton_memory_entries_for = (
+        Translation.objects
+        .filter(approved=True, memory_entries__isnull=True)
+        .prefetch_related('entity__resource__project')
```
jotes: `values_list` could be a little less memory/time consuming if that is a problem.

mathjazz: It worked ok on stage, so I'm keeping it.

PR #684, opened 2017-08-17
Bug 1391049 reject approved vcs

mathjazz:
@jotes r?

This is actually easier than it looks - please check out the 2nd commit. The first commit is just code cleanup that makes things a little easier to understand.

### DIFF
--- a/pontoon/sync/changeset.py
+++ b/pontoon/sync/changeset.py
@@ -252,6 +252,13 @@ class ChangeSet(object):
                 translation.approved_user = None
                 translation.approved_date = None
 
+                # Reject translations unless they became fuzzy during sync. Condition is sufficient
+                # because they were approved previously.
+                if not translation.fuzzy:
+                    translation.rejected = True
+                    translation.rejected_user = user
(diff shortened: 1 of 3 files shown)


pontoon/sync/changeset.py
```diff
@@ -190,70 +190,84 @@ def execute_create_db(self):
...
+                # Use the translation instance already set for update if it exists.
                 translation = self.translations_to_update.get(translation.pk, translation)
                 translation.approved = False
                 translation.approved_user = None
                 translation.approved_date = None
 
+                # Reject translations unless they became fuzzy during sync. Condition is sufficient
+                # because they were approved previously.
+                if not translation.fuzzy:
+                    translation.rejected = True
```
jotes: nit: maybe we could create a general methods for such operations? e.g. translatios.reject(user), translation.approve?

mathjazz: Yeah! Maybe that's something we should do together with [bug 1179954](https://bugzilla.mozilla.org/show_bug.cgi?id=1179954).

PR #688, opened 2017-08-23
Bug 1383252: Reduce the number of hg clone operations

mathjazz:
This is 3 things and it's better to look at them commit by commit:

1. Purely cosmetic. But it helps a little with the following ones.
2. If `sync_projects` is ran with the `--locale` parameter, we only sync the specified locale, but we also pull repositories for all other locales within the project. And rely on that fact in a few places. This anomaly is now gone, which and helps with the final commit.
3. The actual bugfix. More info in commit description.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -1391,6 +1391,41 @@ class Repository(models.Model):
         """True if we can commit strings back to this repo."""
         return self.type in ('svn', 'git', 'hg')
 
+    @cached_property
+    def api_config(self):
+        """
+        Repository API configuration consists of:
+        - Endpoint: A URL pattern to get repository metadata from. Used during sync for faster
+          retrieval of latest commit hashes in combination with the Key.
+          Supports {locale_code} wildcard.
+        - Key: A string used to retrieve the latest commit hash from the JSON response.
+        """
+        url = self.url
+
+        if url.startswith('ssh://hg.mozilla.org/'):
+            parsed_url = urlparse(url)
+            endpoint = (
+                'https://{netloc}/{path}/json-rev/default'
+                .format(netloc=parsed_url.netloc, path=parsed_url.path.strip('/'))
+            )
+            return {
+                'endpoint': endpoint,
+                'get_key': lambda x: x['node'],
+            }
+
+        if url.startswith('ssh://hg@bitbucket.org/'):
+            parsed_url = urlparse(url)
+            endpoint = (
+                'https://api.bitbucket.org/2.0/repositories/{path}/commit/default'
+                .format(path=parsed_url.path.strip('/'))
+            )
+            return {
+                'endpoint': endpoint,
+                'get_key': lambda x: x['hash'],
+            }
+
+        return None
+
     def locale_checkout_path(self, locale):
         """
         Path where the checkout for the given locale for this repo is
@@ -1425,7 +1460,7 @@ class Repository(models.Model):
 
         raise ValueError('No repo found for path: {0}'.format(path))
 
-    def pull(self, skip_locales=None):
+    def pull(self, locales=None):
         """
         Pull changes from VCS. Returns the revision(s) of the repo after
         pulling.
@@ -1437,8 +1472,9 @@ class Repository(models.Model):
             }
         else:
             current_revisions = {}
-            skip_locales = skip_locales or []
-            for locale in self.project.locales.exclude(code__in=skip_locales):
+            locales = locales or self.project.locales.all()
+
+            for locale in locales:
                 repo_type = self.type
                 url = self.locale_url(locale)
                 checkout_path = self.locale_checkout_path(locale)
--- a/pontoon/sync/core.py
+++ b/pontoon/sync/core.py
@@ -1,5 +1,6 @@
 from functools import wraps
 import logging
+import requests
 
 from collections import Counter
 
@@ -164,6 +165,27 @@ def update_translated_resources(db_project, vcs_project, locale):
                 translatedresource.calculate_stats()
 
 
+def update_translated_resources_no_files(db_project, locale, changed_resources):
+    """
+    Create/update TranslatedResource entries if files aren't available. This typically happens when
+    originals change and translations don't, so we don't pull locale repositories.
+    """
+    for resource in changed_resources:
+        # We can only update asymmetric (monolingual) TranslatedResources. For bilingual files we
+        # only create TranslatedResources if the file is present in the repository for the locale,
+        # which we cannot check without files.
+        if not resource.is_asymmetric:
+            log.error(
+                'Unable to calculate stats for asymmetric resource: {resource}.'.format(resource)
+            )
+            continue
+
+        translatedresource, _ = (
+            TranslatedResource.objects.get_or_create(resource=resource, locale=locale)
+        )
+        translatedresource.calculate_stats()
+
+
 def get_vcs_entities(vcs_project):
     return {entity_key(entity): entity for entity in vcs_project.entities}
 
@@ -198,20 +220,39 @@ def entity_key(entity):
     return ':'.join([entity.resource.path, key])
 
 
-def pull_changes(db_project, source_only=False):
+def pull_changes(db_project, locales=None):
     """
     Update the local files with changes from the VCS. Returns True
     if any of the updated repos have changed since the last sync.
     """
     changed = False
-    repositories = [db_project.source_repository] if source_only else db_project.repositories.all()
     repo_locales = {}
-    skip_locales = []  # Skip already pulled locales
+
+    # When syncing sources, pull source repository only.
+    if locales is None:
+        repositories = [db_project.source_repository]
+    # When syncing locales and some have changed, pull all project repositories.
+    elif locales:
+        repositories = db_project.repositories.all()
+    # When syncing locales and none have changed, quit early.
+    else:
+        return changed, repo_locales
+
+    locales = locales or db_project.locales.all()
+
+    # Skip already pulled locales. Useful for projects with multiple repositories (e.g. Firefox),
+    # since we don't store the information what locale belongs to what repository.
+    pulled_locales = []
 
     for repo in repositories:
-        repo_revisions = repo.pull(skip_locales)
+        remaining_locales = locales.exclude(code__in=pulled_locales)
+        if not remaining_locales:
+            break
+
+        repo_revisions = repo.pull(remaining_locales)
         repo_locales[repo.pk] = Locale.objects.filter(code__in=repo_revisions.keys())
-        skip_locales += repo_revisions.keys()
+        pulled_locales += repo_revisions.keys()
+
         # If any revision is None, we can't be sure if a change
         # happened or not, so we default to assuming it did.
         unsure_change = None in repo_revisions.values()
@@ -241,3 +282,82 @@ def commit_changes(db_project, vcs_project, changeset, locale):
     locale_path = vcs_project.locale_directory_paths[locale.code]
     repo = db_project.repository_for_path(locale_path)
     repo.commit(commit_message, commit_author, locale_path)
+
+
+def get_changed_locales(db_project, locales):
+    """
+    Narrow down locales to the ones that have changed since the last sync by fetching latest
+    repository commit hashes via API. For projects with many repositories, this is much faster
+    than running VCS pull/clone for each repository.
+    """
+    repos = db_project.translation_repositories()
+
+    # Requirement: all translation repositories must have API configured.
+    for repo in repos:
+        if not repo.api_config:
+            return locales
+
+    log.info('Fetching latest commit hashes for project {0} started.'.format(db_project.slug))
+
+    # If locale has changed in the DB, we need to sync it.
+    changed_locale_pks = list(locales.filter(
+        changedentitylocale__entity__resource__project=db_project
+    ).values_list('pk', flat=True))
+
+    unchanged_locale_pks = []
+    error_locale_pks = set()
+
+    for repo in repos:
+        for locale in locales:
+            # If we already processed the locale, we can move on.
+            if locale.pk in changed_locale_pks + unchanged_locale_pks:
+                continue
+
+            try:
+                locale_api_endpoint = repo.api_config['endpoint'].format(locale_code=locale.code)
+                response = requests.get(locale_api_endpoint)
+
+                # Raise exception on 4XX client error or 5XX server error response
+                response.raise_for_status()
+
+                # If locale has not synced yet, we need to sync it.
+                last_synced_commit_id = repo.get_last_synced_revisions(locale.code)
+                if not last_synced_commit_id:
+                    changed_locale_pks.append(locale.pk)
+                    continue
+
+                # If locale has changed in the VCS, we need to sync it.
+                latest_commit_id = repo.api_config['get_key'](response.json())
+                if not latest_commit_id.startswith(last_synced_commit_id):
+                    changed_locale_pks.append(locale.pk)
+
+                # If locale hasn't changed in the VCS, we don't need to sync it.
+                else:
+                    unchanged_locale_pks.append(locale.pk)
+
+            # Errors and exceptions can mean locale is in a different repository or indicate
+            # an actual network problem.
+            except requests.exceptions.RequestException:
+                error_locale_pks.add(locale.pk)
+
+    # Check if any locale for which the exception was raised hasn't been processed yet.
+    # For those locales we can't be sure if a change happened, so we assume it did.
+    for l in error_locale_pks:
+        if l not in changed_locale_pks + unchanged_locale_pks:
+            log.error(
+                'Unable to fetch latest commit hash for locale {locale} in project {project}'
+                .format(locale=Locale.objects.get(pk=l), project=db_project.slug)
+            )
+            changed_locale_pks.append(locale.pk)
+
+    changed_locales = db_project.locales.filter(pk__in=changed_locale_pks)
+
+    log.info(
+        'Fetching latest commit hashes for project {project} complete. Changed locales: {locales}.'
+        .format(
+            project=db_project.slug,
+            locales=', '.join(changed_locales.values_list("code", flat=True))
+        )
+    )
+
+    return changed_locales
--- a/pontoon/sync/tasks.py
+++ b/pontoon/sync/tasks.py
@@ -2,6 +2,7 @@ import logging
 
 from django.conf import settings
 from django.db import connection, transaction
+from django.db.models import Q
 from django.utils import timezone
 
 from pontoon.base.models import (
@@ -19,7 +20,9 @@ from pontoon.sync.core import (
     serial_task,
     update_originals,
     update_translated_resources,
+    update_translated_resources_no_files,
     update_translations,
+    get_changed_locales,
 )
 from pontoon.sync.models import ProjectSyncLog, RepositorySyncLog, SyncLog
 from pontoon.sync.vcs.repositories import CommitToRepositoryException
@@ -77,6 +80,7 @@ def sync_project(
         pk=project_pk,
         message='Could not sync project with pk={0}, not found.'.format(project_pk)
     )
+
     sync_log = get_or_fail(
         SyncLog,
         pk=sync_log_pk,
@@ -85,8 +89,6 @@ def sync_project(
             .format(db_project.slug, sync_log_pk)
         )
     )
-
-    log.info('Syncing project {0}.'.format(db_project.slug))
 
     # Mark "now" at the start of sync to avoid messing with
     # translations submitted during sync.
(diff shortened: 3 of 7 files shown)


pontoon/sync/core.py
```diff
@@ -241,3 +252,74 @@ def commit_changes(db_project, vcs_project, changeset, locale):
...
+            try:
+                locale_api_endpoint = repo.api_endpoint.format(locale_code=locale.code)
+                response = requests.get(locale_api_endpoint)
+
+                # Raise exception on 4XX client error or 5XX server error response
+                if response.status_code != 200:
+                    response.raise_for_status()
+
+                # If locale has changed in the VCS, we need to sync it.
+                latest_commit_id = get_json_value_by_path(response.json(), repo.latest_commit_path)
```
mathjazz: I wonder if there's a better way of solving this problem.

Currently we have two repository sources for Firefox (and a few other projects):
* https://api.bitbucket.org/2.0/repositories/mozilla-l10n/arn/commits
* https://hg.mozilla.org/l10n-central/ja/rev/default?style=json

Since we need to retrieve commit hash from 2 different JSON files, I added this `latest_commit_path` which tries to simulate what XPath does.

pontoon/sync/core.py
```diff
@@ -241,3 +252,74 @@ def commit_changes(db_project, vcs_project, changeset, locale):
...
+                if latest_commit_id[:12] != repo.get_last_synced_revisions(locale.code):
```
flodolo: Curiosity: isn't it standard to use the first characters of a commit hash?

mathjazz: Yes. This should be the same as `latest_commit_id[0:12]`. I could change for clarity.

mathjazz:
Added another commit that I'll squash with the 3rd one after the review.

pontoon/sync/tasks.py
```diff
@@ -98,40 +97,35 @@ def sync_project(
...
 
-    # Sync resources
-    resource_changes = sync_resources(db_project, now, force, no_pull)
-    if not resource_changes:
-        project_sync_log.skip()
-        return
+    # Sync source strings. We cannot sync sources if locale is specified,
+    # because that would apply any source string changes to the specified locale only.
+    if locale:
+        source_changes = None
```
Pike: Make it an empty dict here? ...

pontoon/sync/tasks.py
```diff
@@ -98,40 +97,35 @@ def sync_project(
...
     sync_translations.delay(
         project_pk,
         project_sync_log.pk,
         now,
-        resource_changes['project_changes'],
-        resource_changes['obsolete_vcs_resources'],
-        resource_changes['new_paths'],
+        source_changes['project_changes'] if source_changes else None,
+        source_changes['obsolete_vcs_resources'] if source_changes else None,
+        source_changes['new_paths'] if source_changes else None,
```
Pike: ... then you can just .get() here.

pontoon/sync/core.py
Pike: The API endpoints should be:

https://hg.mozilla.org/l10n-central/de/json-rev/default -> get result.node
https://api.bitbucket.org/2.0/repositories/mozilla-l10n/arn/commit/default -> get result.hash
https://api.github.com/repos/mozilla-l10n/firefoxios-l10n/git/refs/heads/master -> get result.object.sha

Note the rate limit for gh, we might need to get an API key, or add that in a separate bug.

pontoon/sync/core.py
Pike: .startswith is OK, but we should aim to use full-length shas for this.

pontoon/base/models.py
```diff
@@ -1321,6 +1321,16 @@ class Repository(models.Model):
...
 
+    api_endpoint = models.CharField("API Endpoint", max_length=2000, blank=True, help_text="""
+        A URL pattern to get repository metadata from. Used during sync for faster retrieval of
+        latest commit hashes in combination with `latest_commit_path`.
+        Supports `{locale_code}` wildcard.
+    """)
+    latest_commit_path = models.CharField(max_length=2000, blank=True, help_text="""
+        A slash-separated list of keys and indices used to retrieve the latest commit hash value
+        from the `api_endpoint` JSON response.
+    """)
```
Pike: This one bothers me from a technical POV.

We shouldn't trust admins to know and get API endpoints right. I think we should do something programmatically to transform the repo to API.

pontoon/sync/tasks.py
```diff
@@ -246,6 +234,20 @@ def sync_translations(
         )
         obsolete_vcs_entities = project_changes['obsolete_db']
 
+    # If project repositories have API access, we can retrieve latest commit hashes and detect
+    # changed locales before the expensive VCS pull/clone operations. When performing full scan and
+    # when source strings have changed, we still need to sync all locales.
```
Pike: Or do we? I think we do for a full scan, but for en-US changes, I think we don't. We used to when we sync'ed removed strings and templates, but as we're only syncing files with l10n activity today, we don't?

pontoon/sync/core.py
flodolo: Not sure why I read this the other way around…

pontoon/base/models.py
flodolo: +1 on this. Setting up a new project would be a bunch of "open another project, copy and paste setting from there".

mathjazz: PMs should not bother setting those up. That's why I'm hiding these fields from the currently used and documented project admin interface. I can also hide them from the Django admin.

But I'm not exactly sure if we can automate this.

pontoon/sync/core.py
mathjazz: I believe we should limit this optimization to the 6 projects that use per-locale repositories, which means we don't need to change anything for the GitHub repositories. In > 90% of cases all 27 of those are done in less than a minute.

With that, I wonder if I should drop `get_json_value_by_path` and change `latest_commit_path` to `latest_commit_key`?

pontoon/sync/tasks.py
mathjazz: Mmmm, I love this!

pontoon/sync/tasks.py
mathjazz: Yeah, when I woke up I remembered I needed to add this check and while I was driving to gym I realized I remembered wrong. Will fix.

pontoon/base/models.py
Pike: This shouldn't be hard to do programmatically, it's just a regular expression replacement for each hosting service.

pontoon/sync/core.py
Pike: As with the API end points, I think this shouldn't be in the DB, and depending on how we abstract the hosting API, this might just move into that abstraction.

pontoon/base/models.py
```diff
@@ -1401,6 +1391,37 @@ def can_commit(self):
...
+        Repository API configuration consists of:
+        - Endpoint: A URL pattern to get repository metadata from. Used during sync for faster
+          retrieval of latest commit hashes in combination with the Key.
+          Supports {locale_code} wildcard.
+        - Key: A string used to retrieve the latest commit hash from the JSON response.
+        """
+        url = self.url
+
+        if url.startswith('ssh://hg.mozilla.org/'):
+            endpoint_base = url.replace('ssh://', 'https://')
```
Pike: I find this a bit hard to digest. How about

```
'https://{0.netloc}{0.path}json-rev/default'.format(urlparse(r.url))
```

pontoon/base/models.py
```diff
@@ -1401,6 +1391,37 @@ def can_commit(self):
...
+            return {
+                'endpoint': os.path.join(endpoint_base, 'json-rev/default'),
+                'key': 'node',
```
Pike: Can we return a lambda function instead?

That way we'll be able to easily add gh API support in the future.

pontoon/base/models.py
```diff
@@ -1401,6 +1391,37 @@ def can_commit(self):
...
+            }
+
+        elif url.startswith('ssh://hg@bitbucket.org/'):
```
Pike: omg, what a nitpicker :-(

No else after return. Also below.

pontoon/base/models.py
```diff
@@ -1401,6 +1391,37 @@ def can_commit(self):
...
+            endpoint_base = url.replace(
+                'ssh://hg@bitbucket.org/',
+                'https://api.bitbucket.org/2.0/repositories/',
+            )
```
Pike:
```
https://api.bitbucket.org/2.0/repositories/{0.path}/commit/default'.format(urlparse(url))
```
would work here.

pontoon/sync/core.py
```diff
@@ -264,7 +263,7 @@ def get_changed_locales(db_project, locales):
 
     # Requirement: all translation repositories must have API configured.
     for repo in repos:
-        if not (repo.api_endpoint and repo.latest_commit_path):
+        if not repo.api_config:
```
Pike: I'd cache the api_config here, and not use the getter twice below.

pontoon/base/models.py
```diff
@@ -1401,6 +1391,37 @@ def can_commit(self):
...
+            return {
+                'endpoint': os.path.join(endpoint_base, 'commit/default'),
```
jotes: Do we need to do os.path.join here? I would move `commit/default` to the rest of url at 1415.

pontoon/base/models.py
jotes: Alternative option: https://pypi.python.org/pypi/jsonpath/ . However, that may be overkill.

pontoon/base/models.py
jotes: @Pike :+1: Optionally, we could do something like this:
```
url = urlparse(self.url)
# That dict can be moved into constant somewhere above.
supported_repos =  {
  'hg.mozilla.org': {
       'endpoint': 'https://hg.mozilla.org/json-rev/default',
      'node': 'node',
   },
  (...)
}
if url.netloc in supported_repos:
   repo = supported_repos[url.netloc]
   repo['endpoint'] = repo['endpoint'].format(url)
   return repo
else:
   return None
```
That could reduce number of nested code blocks in this property in the future and even we could make this more reusable for non-mozilla related projects (?) <insert philosophical raptor here>

pontoon/sync/core.py
```diff
@@ -241,3 +251,74 @@ def commit_changes(db_project, vcs_project, changeset, locale):
...
+            if locale in changed_locales_in_db:
+                changed_locale_pks.append(locale.pk)
+                continue
+
+            try:
+                locale_api_endpoint = repo.api_config['endpoint'].format(locale_code=locale.code)
+                response = requests.get(locale_api_endpoint)
+
+                # Raise exception on 4XX client error or 5XX server error response
+                if response.status_code != 200:
```
jotes: that isn't necessary for response.raise_for_status()?

mathjazz: Nice catch!

mathjazz:
I restored production DB on stage and synced Firefox (96 locales) with and without this patch applied. In both cases there were 5 locales that needed to be synced.

Wihtout patch:
* total sync time: 16:10
* hg time (clone + identify): 15:26

With patch:
* total sync time: 01:54
* hg time (clone and id): 00:46
* api time: 00:34

pontoon/sync/core.py
```diff
@@ -241,3 +251,85 @@ def commit_changes(db_project, vcs_project, changeset, locale):
...
+                    changed_locale_pks.append(locale.pk)
+
+                # If locale hasn't changed in the VCS, we don't need to sync it.
+                else:
+                    unchanged_locale_pks.append(locale.pk)
+
+            # Errors and exceptions can mean locale is in a different repository or indicate
+            # an actual network problem.
+            except requests.exceptions.RequestException as e:
+                log.info(
```
Pike: This sounds a bit spammy over time? Like, we'll get this each sync process for all locales on bitbucket from hgmo?

pontoon/sync/tasks.py
```diff
@@ -229,11 +224,18 @@ def sync_translations(
         repo_sync_log.end()
         return
 
+    # If project repositories have API access, we can retrieve latest commit hashes and detect
+    # changed locales before the expensive VCS pull/clone operations. When performing full scan,
+    # we still need to sync all locales.
+    if not full_scan:
+        locales = get_changed_locales(db_project, locales) or locales
```
Pike: This will sync all locales if it should sync none.

This goes back to the changed_locales code returning None if any of the repos doesn't have an API. Maybe we should revert that logic and sync all locales for which we don't have an API return code, similar to what we're doing for http errors?

Then this `or locales` clause would work via get_changed_locales directly, and the "no locales need syncing" bug would be fixed.

Pike:
> when source strings change, we need to pull all locales and sync them to update stats

Maybe we need to "sync" them, but we don't need to pull them?

mathjazz:
> Maybe we need to "sync" them, but we don't need to pull them?

That's the problem - we currently need files for sync, so we need to pull.

Should be a relatively simple fix.

mathjazz:
I reverted the temporary commit and added an actual fix to properly sync locales on project changes, even if we don't pull them.

It's a little bigger than what I was hoping for. I made a few changes unnecessary for this fix, but they should make the code more readable.

@Pike Since you're the reviewer of this PR, I wonder if you could review this commit, too?

pontoon/sync/core.py
```diff
@@ -165,6 +165,26 @@ def update_translated_resources(db_project, vcs_project, locale):
                 translatedresource.calculate_stats()
 
 
+def update_translated_resources_no_files(db_project, locale, changed_resources):
```
mathjazz: This is the main change. It updates translated resources for all sources that have source string changed.

pontoon/sync/core.py
```diff
@@ -165,6 +165,27 @@ def update_translated_resources(db_project, vcs_project, locale):
...
+    """
+    for resource in changed_resources:
+        # We can only update asymmetric (monolingual) TranslatedResources. For bilingual files we
+        # only create TranslatedResources if the file is present in the repository for the locale,
+        # which we cannot check without files.
+        if not resource.is_asymmetric:
+            log.error(
+                'Unable to calculate stats for asymmetric resource: {resource}.'.format(resource)
+            )
+            continue
```
mathjazz: This should never happen, because bilingual files need to be updated for all locales (merged), when the source change.

Still, maybe we should throw an exception here instead, so that we notice the issue?

pontoon/sync/tasks.py
```diff
@@ -238,14 +240,17 @@ def sync_translations(
...
-        resources_changed = (
-            project_changes['update_db'] +
-            project_changes['obsolete_db'] +
-            project_changes['create_db']
+        updated_entity_pks = [x[1].pk for x in project_changes['update_db']]
+        obsolete_entity_pks = project_changes['obsolete_db']
+        changed_resources = db_project.resources.filter(
+            Q(entities__date_created=now) |
+            Q(entities__pk__in=updated_entity_pks + obsolete_entity_pks)
         )
```
mathjazz: * `project_changes['update_db']`: entities to update in DB, if e.g. string or comment changes and key doesn't
* `project_changes['obsolete_db']`: entities to obsolete
* `project_changes['create_db']`: entities to create in DB

Since `project_changes['create_db']` doesn't contain PKs, we need to fetch those entities using `Q(entities__date_created=now)`. All newly created entities have the same timestamp as the sync (`now`).

Alternative approach would be to gather `changed_resources` while syncing sources and pass them all the way here, but I'd rather do this as a followup to this PR (to avoid making it even bigger).

pontoon/sync/tasks.py
```diff
@@ -283,34 +289,13 @@ def sync_translations(
                 if not vcs_project.synced_locales:
                     vcs_project.resources
 
-                if not obsolete_vcs_resources:
```
mathjazz: This check is actually no longer needed. It's role was to prevent skipping the commit when files were removed from source repository - so we could also remove them from locale repository. But we no longer do that.

And, we should rename `obsolete_vcs_resources` to `removed_paths` in a followup.

pontoon/sync/tasks.py
```diff
@@ -283,34 +289,13 @@ def sync_translations(
...
-                        if resources_changed:
-                            update_translated_resources(db_project, vcs_project, locale)
-                            update_locale_project_locale_stats(locale, db_project)
-                            log.info('Skipping locale `{0}` for project {1}, no changes detected.'
-                                     .format(locale.code, db_project.slug))
-                            log.info(vcs_project.resources)
-                        continue
+                # Skip all locales if none of the them has anything to sync
+                if len(vcs_project.synced_locales) == 0:
+                    break
```
mathjazz: And now that we also update stats after the big `for locale in locales` loop (to c), we can simply break instead of return.

pontoon/sync/tasks.py
```diff
@@ -283,34 +289,13 @@ def sync_translations(
...
+
+                # Skip locales that have nothing to sync
+                if vcs_project.synced_locales and locale not in vcs_project.synced_locales:
+                    continue
```
mathjazz: Similarly as above, stats for those locales will be updated at the end of the big for loop.

pontoon/sync/core.py
```diff
@@ -165,6 +165,27 @@ def update_translated_resources(db_project, vcs_project, locale):
                 translatedresource.calculate_stats()
 
 
+def update_translated_resources_no_files(db_project, locale, changed_resources):
+    """
+    Create/update TranslatedResource entries if files aren't available. This tipically happens when
```
Pike: 'tipi' -> 'typically'

pontoon/sync/core.py
Pike: What would happen on next sync if we do?

pontoon/sync/tasks.py
```diff
@@ -238,14 +240,17 @@ def sync_translations(
...
-    resources_changed = []
+    changed_resources = None
     obsolete_vcs_entities = []
+
     if project_changes:
-        resources_changed = (
-            project_changes['update_db'] +
-            project_changes['obsolete_db'] +
-            project_changes['create_db']
+        updated_entity_pks = [x[1].pk for x in project_changes['update_db']]
```
Pike: Can you use a real tuple here that hints at why we're taking the second field?

pontoon/sync/tasks.py
Pike: The 'now' query has a race condition with regular translations, and other sync process, right?

Do we use bulk_insert? https://docs.djangoproject.com/en/1.11/ref/models/querysets/#bulk-create says that starting with 1.10, on postgres, we get the IDs back.

mathjazz: > The 'now' query has a race condition with regular translations, and other sync process, right?

No. Regular translations are stored in a different model (Translation) than source strings (Entity). Other sync processes for the same project [cannot happen at the same time](https://github.com/mozilla/pontoon/blob/master/pontoon/sync/core.py#L37).

> Do we use bulk_insert? https://docs.djangoproject.com/en/1.11/ref/models/querysets/#bulk-create says that starting with 1.10, on postgres, we get the IDs back.

No, but we [should](https://github.com/mozilla/pontoon/blob/master/pontoon/sync/changeset.py#L167). Thanks. Although that wouldn't bring us IDs here.

pontoon/sync/core.py
mathjazz: Regardless of raising an exception, we no longer have the information about the project changes in the following sync. That's [bug 1383271](https://bugzilla.mozilla.org/show_bug.cgi?id=1383271).

pontoon/sync/tasks.py
```diff
@@ -244,7 +244,10 @@ def sync_translations(
     obsolete_vcs_entities = []
 
     if project_changes:
-        updated_entity_pks = [x[1].pk for x in project_changes['update_db']]
+        updated_entity_pks = []
+        for locale_code, db_entity, vcs_entity in project_changes['update_db']:
+            updated_entity_pks = db_entity.pk
```
Pike: .append(db_entity.pk)

pontoon/sync/core.py
Pike: I'd not error out here for now. Also, I'm not 100% confident that we should restrict ourselves to have projects with only monolingual or only bilingual files.

pontoon/sync/tasks.py
Pike: 'k

pontoon/sync/core.py
mathjazz: We have projects with a mix of monolingual or bilingual files (e.g. Fundraising). They should work fine with this code.

If the source .lang file changes, all locales will be updated in the repository, so we'll have files for all locales and use the current `update_translated_resources` method.

If only the source .properties file changes, we'll use `update_translated_resources_no_files` for locales that don't have any locale changes.

PR #690, opened 2017-08-28
Filters optimization for non-PO projects

mathjazz:
Our model has two drawbacks that make it hard (slow) to filter and search strings:

1. Most filters need to identify an *active translation* (approved or latest suggestion) for the string, and there's no easy (fast) way of retrieving active translations from the model.

2. For projects using .po files we store translations for different plural forms as different Translation model instances, which makes it even harder to retrieve active translations: we need to retrieve one for each plural form supported for a locale. To circumvent that we use `with_status_counts` annotations, but that slows things down even further.

This PR solves the issue #2 for non-PO projects. That's currently 26 out of 33 projects enabled in Pontoon. The number also includes our biggest projects in terms of number of source strings, which perform slower when filtering.

It's a quick and dirty hack, meant to reduce the number of Request timeouts we're seeing after we moved locales from Pootle to Pontoon. It's deployed on stage ATM. Applying status and extra filters feels noticably faster than on production. DB was restored from a few days old production backup. 

Note that this is only a temporary hack before we come up with a proper solution for bug 1377099.

Pinging @adngdb and @jotes for feedback.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -1601,17 +1601,23 @@ class EntityQuerySet(models.QuerySet):
     user_entity_filters = ('missing', 'fuzzy', 'suggested', 'translated')
     user_extra_filters = ('has-suggestions', 'rejected', 'unchanged')
 
-    def filter_statuses(self, locale, statuses):
+    def filter_statuses(self, locale, statuses, no_plurals):
         sanitized_statuses = filter(lambda s: s in self.user_entity_filters, statuses)
         if sanitized_statuses:
-            return reduce(lambda x, y: x | y, [getattr(Entity.objects, s)() for s in sanitized_statuses])
+            return reduce(
+                lambda x, y: x | y,
+                [getattr(Entity.objects, s)(locale, no_plurals) for s in sanitized_statuses]
+            )
         return Q()
 
-    def filter_extra(self, locale, extra):
+    def filter_extra(self, locale, extra, no_plurals):
         sanitized_extras = filter(lambda f: f in self.user_extra_filters, extra)
 
         if sanitized_extras:
-            return reduce(lambda x, y: x | y, [getattr(Entity.objects, s.replace('-', '_'))() for s in sanitized_extras])
+            return reduce(
+                lambda x, y: x | y,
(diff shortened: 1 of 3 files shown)


pontoon/base/models.py
```diff
@@ -1854,6 +1899,11 @@ def for_project_locale(self, project, locale, paths=None, statuses=None,
         pre_filters = []
         post_filters = []
 
+        # Temporary optimization: use simpler status & extra filter DB queries for non-PO projects,
+        # which don't store translations for different plural forms in separate Translation model
+        # instances.
+        no_plurals = True if len(project.resources.filter(format='po')) == 0 else False
```
jotes: `no_plurals = (len(project.resources.filter(format='po')) == 0)`

adngdb: You can even do without the parentheses. And you can also do this: 
```python
no_plurals = not len(project.resources.filter(format='po'))
```

Or you can go further and do this (but I'm not sure that's more readable):
```python
no_plurals = not project.resources.filter(format='po')
```

Pike: Or even .exists(), https://docs.djangoproject.com/en/1.8/ref/models/querysets/#django.db.models.query.QuerySet.exists.

adngdb:
I find this quite difficult to review, but I see that we have some failing unit tests. I think it will be easier with them, at least the get a better grasp of the expected behavior.

pontoon/base/models.py
```diff
@@ -1902,7 +1902,7 @@ def for_project_locale(self, project, locale, paths=None, statuses=None,
         # Temporary optimization: use simpler status & extra filter DB queries for non-PO projects,
         # which don't store translations for different plural forms in separate Translation model
         # instances.
-        no_plurals = True if len(project.resources.filter(format='po')) == 0 else False
+        no_plurals = project.resources.filter(format='po').exists()
```
adngdb: I think you meant 
```python
no_plurals = not project.resources.filter(format='po').exists()
```

mathjazz: And I even updated test to compensate for that! Grrrr... (and thanks for the catch!)

adngdb:
Also, I think this should be linked to [bug 1377099](https://bugzilla.mozilla.org/show_bug.cgi?id=1377099) as it is a substantial step towards its completion.

PR #695, opened 2017-09-01
Fix bug 1363164: Show read notifications in popup

mathjazz:
In notifications popup, always show the latest 7 notifications, including the unread. If there are more than 7 unread notifications, show all.

@jotes r?

### DIFF
--- a/pontoon/contributors/templates/contributors/widgets/notifications_menu.html
+++ b/pontoon/contributors/templates/contributors/widgets/notifications_menu.html
@@ -1,14 +1,19 @@
 {% macro menu() %}
 {% if user.is_authenticated() %}
+
 {% set unread_notifications = user.notifications.unread() %}
-<div id="notifications" class="notifications select{% if unread_notifications|length > 0 %} unread{% endif %}">
+{% set unread_notifications_count = unread_notifications.count() %}
+{% set limit = 7 %}
+{% set notifications_to_display_count = unread_notifications_count if unread_notifications_count > limit else limit %}
+
+<div id="notifications" class="notifications select{% if unread_notifications_count > 0 %} unread{% endif %}">
 
   <div class="button selector">
     <i class="icon fa fa-bell-o fa-fw"></i>
   </div>
 
   <div class="menu">
-    {{ list(notifications=unread_notifications) }}
+    {{ list(notifications=user.notifications.all()[:notifications_to_display_count]) }}
 
     <ul>
       <li class="horizontal-separator"></li>
@@ -17,6 +22,7 @@
   </div>
 
 </div>
+
 {% endif %}
 {% endmacro %}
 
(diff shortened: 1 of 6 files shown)


pontoon/contributors/templates/contributors/widgets/notifications_menu.html
```diff
@@ -1,14 +1,19 @@
 {% macro menu() %}
 {% if user.is_authenticated() %}
+
 {% set unread_notifications = user.notifications.unread() %}
-<div id="notifications" class="notifications select{% if unread_notifications|length > 0 %} unread{% endif %}">
+{% set unread_notifications_count = unread_notifications.count() %}
+{% set limit = 7 %}
+{% set notifications_to_display_count = unread_notifications_count if unread_notifications_count > limit else limit %}
+
+<div id="notifications" class="notifications select{% if unread_notifications_count > 0 %} unread{% endif %}">
```
jotes: I think you could remove that '> 0' part and do something like:  `{{ ' unread' if unread_notifications_count else '' }}`

pontoon/contributors/templates/contributors/widgets/notifications_menu.html
```diff
@@ -1,14 +1,19 @@
 {% macro menu() %}
 {% if user.is_authenticated() %}
+
 {% set unread_notifications = user.notifications.unread() %}
-<div id="notifications" class="notifications select{% if unread_notifications|length > 0 %} unread{% endif %}">
+{% set unread_notifications_count = unread_notifications.count() %}
+{% set limit = 7 %}
```
jotes: Sorry if that's a little bit to general question... but shouldn't we keep such numbers in pontoon-settings to avoid hardcoding? That's probably not the thing for this PR but I wonder in general how much pontoon should be extensible for installations outside Mozilla (if there are any).

pontoon/contributors/templates/contributors/widgets/notifications_menu.html
```diff
@@ -1,14 +1,19 @@
...
+{% set notifications_to_display_count = unread_notifications_count if unread_notifications_count > limit else limit %}
```
jotes: Maybe we should move that part into view/model and just display the final list in the template? It will be easier to test later.

pontoon/contributors/templates/contributors/widgets/notifications_menu.html
mathjazz: I agree, updated in the following commit.

pontoon/contributors/templates/contributors/widgets/notifications_menu.html
mathjazz: Good call. Updated.

pontoon/contributors/templates/contributors/widgets/notifications_menu.html
mathjazz: Updated (not exactly as you suggested though).

PR #696, opened 2017-09-01
Fix bug 1390187: Make Transvision opt-in

mathjazz:
@jotes r?


pontoon/base/models.py
```diff
@@ -540,6 +541,11 @@ class Locale(AggregatedStats):
         """
     )
 
+    transvision = models.BooleanField(default=False, help_text="""
```
jotes: btw. Did you consider to add opt-in/opt-out for user profile for all systems supported in machinery tab?

pontoon/base/models.py
```diff
@@ -670,6 +676,7 @@ def serialize(self):
             'script': self.script,
             'ms_translator_code': self.ms_translator_code,
             'ms_terminology_code': self.ms_terminology_code,
+            'transvision': json.dumps(self.transvision),
```
jotes: Is there a reason why do you need to have this boolean as a string?

mathjazz: We read this on the client side and without the json trick, we pass `"True"` instead of `true`. Could also output the entire dict as json.

pontoon/base/models.py
mathjazz: No. Why would we *not* serve suggestions from some source to some people?

pontoon/base/models.py
jotes: That's true, I'm just not sure why you don't serialize the output from `.serialize()` (that could be safer for the rest of the available values).

pontoon/base/models.py
jotes: I just thought that some users may prefer some systems over the other.

pontoon/base/models.py
julen: Out of curiosity, where in the client-side is this loaded specifically? Thanks.

pontoon/base/models.py
mathjazz: We haven't gotten any requests for this.

pontoon/base/models.py
mathjazz: Here's one place the client loads this: https://github.com/mozilla/pontoon/blob/master/pontoon/base/static/js/translate.js#L3488

PR #697, opened 2017-09-01
Fix bug 1396080: Encode non-encoded querystring

mathjazz:
@jotes r?

### DIFF
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -4021,9 +4021,11 @@ var Pontoon = (function (my) {
         return decodeURIComponent(encodedURI);
 
       // If querystring not encoded, we need to encode it first
-      } catch (e if e instanceof URIError) {
-        encodedURI = encodeURIComponent(encodedURI);
-        return decodeURIComponent(encodedURI);
+      } catch (e) {
+        if (e instanceof URIError) {
+          encodedURI = encodeURIComponent(encodedURI);
+          return decodeURIComponent(encodedURI);
+        }
       }
     },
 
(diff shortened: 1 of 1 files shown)


pontoon/base/static/js/translate.js
```diff
@@ -4015,7 +4015,16 @@ var Pontoon = (function (my) {
...
       }
 
-      return decodeURIComponent(results[2].replace(/\+/g, " "));
+      var encodedURI = results[2].replace(/\+/g, " ");
+
+      try {
+        return decodeURIComponent(encodedURI);
+
+      // If querystring not encoded, we need to encode it first
+      } catch (e if e instanceof URIError) {
```
jotes: btw. Did you consider to add raygun.io to the frontend?

julen: This inline conditional catch syntax [is not standard and its use is discouraged](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#Conditional_catch_clauses). You should check for `URIError` in a separate conditional block.

jotes:
thx @julen. I've seen that some of sites use the same exception handling and I assumed that if it works in Fx (that's the browser in which this bug was verified) that means Chrome also will work.
TIL: that syntax is supported only in Firefox and not available in Chrome/Safari/Opera.
@mathjazz @adngdb I would add a bug for javascript linter in pontoon.

pontoon/base/static/js/translate.js
mathjazz: Thanks. I had no idea. And I actually found this syntax in the exact same article. :/

mathjazz:
Updated.

@jotes I support the idea of a JS linter.

PR #698, opened 2017-09-02
Map bools, integers lists to html attributes

jotes:
@mathjazz This was faster than to write large comment in #696  -> feel free to comment/take some parts of it into your pr.

### DIFF
--- a/pontoon/base/templatetags/helpers.py
+++ b/pontoon/base/templatetags/helpers.py
@@ -2,6 +2,8 @@ import cgi
 import datetime
 import json
 
+from xml.sax.saxutils import escape
+
 from six import text_type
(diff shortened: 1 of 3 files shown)


adngdb:
@jotes you can file pull requests directly to @mathjazz's repo. If you do that, he can comment on it and all, and merge your branch into his, which will then automatically update his PR. The only thing is that it's a bit tricky to create a PR to a non-upstream repo with github's interface. The link to that for this would be https://github.com/mathjazz/pontoon/compare/bug-1390187-transvision...jotes:dict_html_attrs_fix?expand=1

jotes:
@adngdb thank you senpai!

PR #700, opened 2017-09-05
Bug 1377099 - Make search faster.

adngdb:
We join the entity table with the translation table in order to search, but that join is done for all the translations, where we only care about translations for the locale the user is looking at. This commit moves the filter on the locale up in the WHERE clause, allowing the database to optimize the join, leading to a faster search time.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -2026,19 +2026,28 @@ class Entity(DirtyFieldsMixin, models.Model):
         if search:
             # https://docs.djangoproject.com/en/dev/topics/db/queries/#spanning-multi-valued-relationships
             search_query = (search, locale.db_collation)
-            entities = (
-                Entity.objects.filter(
-                    (
-                        Q(translation__string__icontains_collate=search_query) |
-                        Q(string__icontains=search) |
-                        Q(string_plural__icontains=search) |
-                        Q(comment__icontains=search) |
-                        Q(key__icontains=search)
-                    ),
-                    pk__in=entities,
+            translation_matches = (
+                entities.filter(
+                    translation__string__icontains_collate=search_query,
                     translation__locale=locale,
                 )
                 .distinct()
+                .values_list('id')
+            )
+            entity_matches = (
+                entities.filter(
+                    Q(string__icontains=search) |
+                    Q(string_plural__icontains=search) |
+                    Q(comment__icontains=search) |
+                    Q(key__icontains=search)
+                )
+                .distinct()
+                .values_list('id')
+            )
+            entities = Entity.objects.filter(
+                pk__in=set(
+                    e[0] for e in list(translation_matches) + list(entity_matches)
+                )
             )
 
         entities = entities.prefetch_resources_translations(locale)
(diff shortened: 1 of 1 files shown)


mathjazz:
This will only search translations and their entities, which means entities without any translations (Missing) will be left out.

That said, searching might be faster if we create a union of two queries - one to search entities and one to search translations - instead of a DB join.

mathjazz:
This seems to be working fine and is still fast: https://gist.github.com/mathjazz/e6d60666a40ad4a07bc857c45402a9af

jotes:
As we discussed at yesterday's water-cooling meeting, that won't help because we have to display entities without translation. Unions may be the solution #1. 
Solution #2: That's what we have to achieve here, is a simple LEFT JOIN between entities and translations (in terms of the SQL), base resources (etc.). I'm not a Django expert, so I'm not so sure if that's possible in ORM, but for the sake of experiment, We could try to rewrite this part into saner (from DB perspective) raw SQL query (but yeah, parametrization will be the hell).

adngdb:
So I went with @mathjazz's solution, and it's even faster than what I had before. Yay!

pontoon/base/models.py
```diff
@@ -2026,12 +2026,28 @@ def for_project_locale(self, project, locale, paths=None, statuses=None,
...
-                Entity.objects.filter(
-                    Q(translation__string__icontains_collate=search_query, translation__locale=locale) | Q(string__icontains=search) | Q(string_plural__icontains=search) | Q(comment__icontains=search) | Q(key__icontains=search),
-                    pk__in=entities
+            translation_matches = (
+                entities.filter(
+                    translation__string__icontains_collate=search_query,
+                    translation__locale=locale,
                 )
                 .distinct()
+                .values_list('id')
```
mathjazz: Nice trick. I'd add `, flat=True` here...

pontoon/base/models.py
```diff
@@ -2026,12 +2026,28 @@ def for_project_locale(self, project, locale, paths=None, statuses=None,
...
+            )
+            entity_matches = (
+                entities.filter(
+                    Q(string__icontains=search) |
+                    Q(string_plural__icontains=search) |
+                    Q(comment__icontains=search) |
+                    Q(key__icontains=search)
+                )
+                .distinct()
+                .values_list('id')
```
mathjazz: ...and here...

pontoon/base/models.py
```diff
@@ -2026,12 +2026,28 @@ def for_project_locale(self, project, locale, paths=None, statuses=None,
...
+            )
+            entities = Entity.objects.filter(
+                pk__in=set(
+                    e[0] for e in list(translation_matches) + list(entity_matches)
+                )
```
mathjazz: ...and then you can just `set(list(translation_matches) + list(entity_matches))`.

adngdb:
On my laptop, we're talking about going from 3 to 4 seconds to a ridiculous 200ms to load the entities. 20 times faster!!! Thanks for the reviews guys!

PR #702, opened 2017-09-05
Pass host of the webapp to docker image.

jotes:
Hi @mathjazz @adngdb 
I'm currently working on a docker environment proxied via virtualbox (joys of the development on windows). And because of there's no way I can access webapp via localhost, because I have to pass IP of a Virtualbox instance, instead Pontoon throws error about mismatched ALLOWED_HOSTS.

Could you review?


mathjazz:
@adngdb I'd rather pass this one to you, since you mastered our Docker install.

@MekliCZ Might be interested in this, too.

docker/config/webapp.env.template
```diff
@@ -3,7 +3,7 @@ DJANGO_DEV=True
 DJANGO_DEBUG=True
 DATABASE_URL=postgres://pontoon:asdf@postgresql/pontoon
 SESSION_COOKIE_SECURE=False
-SITE_URL=http://localhost:8000
+SITE_URL=http://#WEBAPP_HOST#
```
adngdb: I would make that variable contain the entire URL, schema included. In the future we might want to use those docker files for prod deploys, in which case we'll want to use HTTPS.

Makefile
```diff
@@ -1,4 +1,5 @@
 DC := $(shell which docker-compose)
+WEBAPP_HOST ?= localhost:8000
```
adngdb: Make this ``"http://localhost:8000"`` instead.

adngdb: Also, can you add a comment to explain why you want to able to set this variable from the environment?

Pike:
Related question, are there any concrete problems with running pontoon in docker for windows?

I'm using docker for mac, and that works great for me, that's why I'm asking.

jotes:
@adngdb thx! I'll ping you when I'm done.
@Pike I use Windows 8.1 and that may be the part of the problem. So, the problems I encountered on my path:
* Bash is natively supported in Windows 10. In Windows 8.1 you have to use solutions like MinGW, Cygwin etc. -> And for some reason that wasn't playing well with docker toolbox on my machine (I had to modify environment variables etc.). 
* docker-compose was installed in "c:\program files\do..." - a space in path to docker-compose caused error during execution of makefile. That was relatively easy to fix and I've done that locally.
* Build scripts couldn't run (they were failing silently in my case/or couldn't find files) because git downloaded them with windows line endings (\r\n). That was really strange, but after I ran dos2unix on all .sh scripts, they started working fine.
* Make couldn't find docker-compose because there was a missing `.exe` part.
* I don't remember the last error from my breaking point, and unfortunately I don't have more info like e.g. screens.

<joke>After 4 hours of googling my setup still felt a little too mainstream (yeah, not hipster enough). During 5 minutes of meditation, an internal struggle with the forces of the evil and the lesser evil, I decided to come to the dark side and install Ubuntu in Virtualbox with docker inside (on a Windows Host). </joke>

Currently my setup looks like this:
Windows ==> Virtualbox (via NAT) ==> docker 

I may try to figure it out at some point again or just accept advice of someone smarter in Windows arts.

Makefile
adngdb: By this I mean, can you explain the use case that requires you to use a variable here? Like, something about Windows needing to expose the service on a different URL.

Makefile
```diff
@@ -8,8 +13,12 @@ all: dockerrun
 	make dockerbuild
 
 dockerbuild:
+	cp ./docker/config/webapp.env.template ./docker/config/webapp.env
+	sed -i 's/#SITE_URL#/$(subst /,\/,${SITE_URL})/g' ./docker/config/webapp.env
```
julen: This breaks on OS X. It works this way:
```
sed -i '' -e 's/#SITE_URL#/$(subst /,\/,${SITE_URL})/g' ./docker/config/webapp.env
```
Although it might break on other platforms...

docs/dev/install-docker.rst
```diff
@@ -65,7 +71,8 @@ initialize submodules with ``git submodule update --init --recursive``.
    This will ask you to create a superuser, and then will update your Firefox
    account settings.
 
-At that point, you're good to go. Access the webapp via this URL: http://localhost:8000/
+At that point, you're good to go. Access the webapp via this URL: http://localhost:8000/ or the custom SITE_URL.
```
julen: Note that while accessing via the custom `SITE_URL` seems to work on the surface, I'm experiencing some CSP/CORS issues due to the client-side code and HTML markup still referencing `localhost` here and there.

Makefile
jotes: I was running that command under linux envs, that's probably it.

docs/dev/install-docker.rst
jotes: @julen @adngdb That's probably caused due to the fact that our initial migration for pontoon-intro always sets base url to `http://localhost:8000/`. We could parametrize that or change after migrations are applied.

Makefile
jotes: Also, that maybe good to check if that will work on Windows 10 with bash subsystem if someone uses Windows for development.

PR #703, opened 2017-09-06
Faster translate view

mathjazz:
While looking at performance data of the search speedup, New Relic pointed me to a simple fix for improving page load time of the translate view.

It's a simple trick to cache a DB query in a variable insted of querying the DB with the same request in a loop.

Note that the translate view is also used as the homepage for users that are not authenticated or don't have any contributions, which makes it the most popular view.

```
translate view speedup: 40%
```

### DIFF
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -126,6 +126,7 @@ def translate(request, locale, slug, part):
         'download_form': forms.DownloadFileForm(),
         'upload_form': forms.UploadFileForm(),
         'locale': locale,
+        'locale_projects': locale.available_projects_list(),
         'locales': Locale.objects.available(),
         'part': part,
         'project': project,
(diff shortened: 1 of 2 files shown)


adngdb:
I just noticed there are no bugs associated with your three recent pull requests. We should at least have a "performance improvement" bug tracker, so that those can be attached to it. And thus I filed one: https://bugzilla.mozilla.org/show_bug.cgi?id=1397725

adngdb:
> The project selector widget is only loaded inside the translate view,
to which we always pass a project and a locale.

Is that going to change with #520 ?

mathjazz:
> Is that going to change with #520 ?

No. In #520 we should probably just remove the "Resources" selector or make it unselectable. But we'll keep project selector bound to the translate view only.

PR #705, opened 2017-09-06
Faster teams and projects pages

mathjazz:
This one was also discovered using the New Relic data.

We now use python to get top instances and aggregated data from projects and locales QuerySets. It's faster than querying the DB with order_by() and aggregate().

The 2nd commit also has a noticable performance impact. We now list all locales that have at least one project enabled, regardless if it's synced or not.

```
/teams speedup: 70%
/projects speedup: 40%
```

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -412,18 +412,11 @@ class AggregatedStats(models.Model):
         """
         Get sum of stats for all items in the queryset.
         """
-        stats = qs.aggregate(
-            Sum('total_strings'),
-            Sum('approved_strings'),
-            Sum('translated_strings'),
-            Sum('fuzzy_strings'),
-        )
-
         return cls(
-            total_strings=stats['total_strings__sum'],
-            approved_strings=stats['approved_strings__sum'],
-            translated_strings=stats['translated_strings__sum'],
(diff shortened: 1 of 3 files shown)


pontoon/base/models.py
```diff
@@ -475,7 +470,7 @@ def available(self):
         """
         Available locales have at least one TranslatedResource defined.
         """
-        return self.filter(translatedresources__isnull=False).distinct()
+        return self.filter(project_locale__isnull=False).distinct()
```
adngdb: I don't understand this change. Can you explain it, or add a comment?

Pike: ... also noticed that the comment above this is not matching what the code does anymore

mathjazz: This was explained in https://github.com/mozilla/pontoon/pull/705/commits/d0847e7f0be458019ae2b147c21d8c3c3424ace7, but poorly. :) I noticed the query was slow and fixed it by changing it. Instead of showing all locales that have at least one resource of at least one project enabled (synced), we'd show all locales that have at least one project enabled (even if it's not synced yet).

Now I restored the behaviour so it's exactly the same as originally, but the query is much faster. So I'm leaving the comment unchanged.

If we decide to make the change on which locales to list, we can do that in a followup commit.

PR #706, opened 2017-09-08
Bug 1357361: Speed up contributors list

mathjazz:
This brings around 80% speed up to the /contributors view, and up to 90% on more specific contributors pages (/sl/contributors, /projects/firefox/contributors, /sl/firefox/contributors).

On production the load time of these pages averages around 6 seconds, so this should bring us to around 1.5 seconds. I'm not sure we can achieve much more without changing the data model.

Please review each commit separately. I tried to be as verbose as possible in commit messsages. To further clarify the last commit (https://github.com/mozilla/pontoon/pull/706/commits/0f9ce46cc7a9d29d950a22c7e257017b74077023), I'm adding the before and after SQL query:

Before:
```sql
SELECT DISTINCT "auth_user"."id",
                "auth_user"."password",
                "auth_user"."last_login",
                "auth_user"."is_superuser",
                "auth_user"."username",
                "auth_user"."first_name",
                "auth_user"."last_name",
                "auth_user"."email",
                "auth_user"."is_staff",
                "auth_user"."is_active",
                "auth_user"."date_joined",
                Sum(CASE
                      WHEN ( "base_translation"."user_id" IS NOT NULL
                             AND "base_translation"."approved" = true ) THEN 1
                      ELSE 0
                    end) AS "translations_approved_count",
                Sum(CASE
                      WHEN ( "base_translation"."user_id" IS NOT NULL
                             AND "base_translation"."fuzzy" = false
                             AND "base_translation"."approved" = false ) THEN 1
                      ELSE 0
                    end) AS "translations_unapproved_count",
                Sum(CASE
                      WHEN "base_translation"."user_id" IS NOT NULL THEN 1
                      ELSE 0
                    end) AS "translations_count",
                Sum(CASE
                      WHEN ( "base_translation"."user_id" IS NOT NULL
                             AND "base_translation"."fuzzy" = true ) THEN 1
                      ELSE 0
                    end) AS "translations_needs_work_count"
FROM   "auth_user"
       LEFT OUTER JOIN "base_translation"
                    ON ( "auth_user"."id" = "base_translation"."user_id" )
GROUP  BY "auth_user"."id"
HAVING NOT ( Sum(CASE
                   WHEN "base_translation"."user_id" IS NOT NULL THEN 1
                   ELSE 0
                 end) = 0 )
ORDER  BY "translations_count" DESC
LIMIT  100  
```

After:
```sql
SELECT "base_translation"."user_id",
       "base_translation"."approved",
       "base_translation"."fuzzy",
       Count("base_translation"."user_id") AS "count"
FROM   "base_translation"
WHERE  NOT ( "base_translation"."user_id" IS NULL )
GROUP  BY "base_translation"."user_id",
          "base_translation"."approved",
          "base_translation"."fuzzy"  
```

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -91,27 +91,27 @@ class UserTranslationsManager(UserManager):
 
             return self._changed_translations_count(query)
 
-        def extract_locale(name):
-            """Extract locale code from group name."""
-            name = name.split(' ')[0]
-            if '/' in name:
-                name = name.split('/')[1]
-            return name
-
         def with_roles(self):
             """Add user roles."""
             managers = defaultdict(set)
-            manager_groups = Group.objects.filter(name__endswith='managers').prefetch_related('user_set')
-
-            for group in manager_groups:
-                for user in group.user_set.all():
-                    managers[user].add(extract_locale(group.name))
-
             translators = defaultdict(set)
-            translator_groups = Group.objects.filter(name__endswith='translators').prefetch_related('user_set')
-            for group in translator_groups:
-                for user in group.user_set.all():
-                    translators[user].add(extract_locale(group.name))
+
+            locales = Locale.objects.prefetch_related(
+                Prefetch(
+                    'managers_group__user_set',
+                    to_attr='fetched_managers'
+                ),
+                Prefetch(
+                    'translators_group__user_set',
+                    to_attr='fetched_translators'
+                )
+            )
+
+            for locale in locales:
+                for user in locale.managers_group.fetched_managers:
+                    managers[user].add(locale.code)
+                for user in locale.translators_group.fetched_translators:
+                    translators[user].add(locale.code)
 
             for contributor in self:
                 contributor.user_role = contributor.role(managers, translators)
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -43,96 +43,100 @@ from pontoon.sync import KEY_SEPARATOR
 log = logging.getLogger(__name__)
 
 
-# User class extensions
 class UserTranslationsManager(UserManager):
-    """
-    Provides various method to interact with larger sets of translations and their stats for user.
-    """
-
-    def _changed_translations_count(self, *args):
-        """
-        Helper method, returns expression object which allows us to annotate querysets
-        with counts of translations.
-        """
-        translation_query = Q(translation__user__isnull=False)
-        for arg in args:
-            translation_query &= arg
-
-        # For each translation that matches the filter, return 1. Aggregate
-        # the sum of all those results to count the number of matching
-        # translations.
-        return Sum(
-            Case(
-                When(translation_query, then=1), output_field=models.IntegerField(), default=0))
-
     def with_translation_counts(self, start_date=None, query_filters=None, limit=100):
         """
-        Returns contributors list, sorted by count of their translations.
-        Every user instance has added following properties:
+        Returns contributors list, sorted by count of their translations. Every user instance has
+        the following properties:
         * translations_count
         * translations_approved_count
         * translations_unapproved_count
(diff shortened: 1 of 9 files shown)


pontoon/base/models.py
```diff
@@ -91,27 +91,27 @@ def translations_count(query=None):
...
+            locales = Locale.objects.prefetch_related(
+                Prefetch(
+                    'managers_group__user_set',
+                    to_attr='fetched_managers'
+                ),
+                Prefetch(
+                    'translators_group__user_set',
+                    to_attr='fetched_translators'
+                )
+            )
```
adngdb: Maybe you want to make use of ``values()`` here, to only get ``code``  and the two prefetched fields defined before?

pontoon/base/tests/test_models.py
```diff
@@ -544,14 +544,14 @@ def test_period_filters(self):
 
         top_contributors = User.translators.with_translation_counts(aware_datetime(2015, 6, 10))
 
-        assert_equal(top_contributors.count(), 1)
+        assert_equal(len(top_contributors), 1)
```
adngdb: I don't think this is better in terms of perf. The only use you make of ``top_contributors``' values is of the 1st element, so by using ``len()`` here, you force it to select all results, when afterwards you only use the first one. ``count()`` seems better here.

But, the key point is that you're trying to optimize unit tests, which is not a high priority at the moment, so you could probably avoid this altogether.

pontoon/base/models.py
```diff
@@ -43,96 +43,100 @@
...
-                for user in locale.managers_group.fetched_managers:
-                    managers[user].add(locale.code)
-                for user in locale.translators_group.fetched_translators:
-                    translators[user].add(locale.code)
+            if translation['approved']:
+                status = 'approved'
+            elif translation['fuzzy']:
+                status = 'fuzzy'
+            else:
+                status = 'suggested'
```
adngdb: What about rejected translations?

adngdb:
This new versions shows a bit more translations for each user, in each state. Is that expected? Why is it happening?

mathjazz:
> This new versions shows a bit more translations for each user, in each state. Is that expected? Why is it happening?

Yes, this is expected. It because we no longer exclude translations that are the same as the original string. https://github.com/mozilla/pontoon/pull/706/commits/75426ff9815fe8728d101b64ae15b15aca883776

pontoon/base/models.py
mathjazz: We currently don't show them on the dashboard and since this bug is only about optimization, I didn't bother adding rejected. The number of rejected translations has negative connotation, so we should be careful about displaying it publicly and communicate ir properly.

pontoon/base/tests/test_models.py
mathjazz: `top_contributors` is no longer a QuerySet, it's a list.

adngdb: Aha, I missed that, my bad. :)

pontoon/base/models.py
mathjazz: That results in a `FieldError` for the prefetched fields. The performance of this query is pretty good, it takes less than 100 ms.

pontoon/base/models.py
adngdb: Mmm, the thing that bothers me is that you consider rejected translations to be suggested with this code. Isn't that a bit of a problem?

mathjazz: The code in production does that, too. The same thing happens in team, project, localization dashboards, and also filters: Suggested includes rejected. See [bug 1377969](https://bugzilla.mozilla.org/show_bug.cgi?id=1377969).

adngdb: Fair enough then! :)

PR #707, opened 2017-09-08
Fix bug 1398108: Do not duplicate group type in group name

mathjazz:
@jotes r?


pontoon/base/migrations/0103_bug_1398108_rename_groups_with_duplicate_words.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def rename_groups_with_duplicate_words(apps, schema_editor):
+    Group = apps.get_model('auth', 'Group')
+    groups_to_update = []
+
+    for name in ('managers', 'translators'):
+        double = '{} {}'.format(name, name)
+        for group in Group.objects.filter(name__contains=double):
+            group.name = group.name.replace(double, name)
```
jotes: I think you could try to use  `.update()` here.

mathjazz: I tried that in combination with the `F()` expression, but failed to find a solution for search and replace.

jotes: https://gist.github.com/simonw/30d2e626772e84edd688ba717a766c89

mathjazz: Thanks! Updated.

pontoon/base/signals.py
```diff
@@ -94,8 +94,8 @@ def create_locale_permissions_groups(sender, **kwargs):
         return
 
     try:
-        create_group(instance, 'translators', ['can_translate_locale'], '{} translators'.format(instance.code))
-        create_group(instance, 'managers', ['can_translate_locale', 'can_manage_locale'], '{} managers'.format(instance.code))  # noqa
+        create_group(instance, 'translators', ['can_translate_locale'], instance.code)
+        create_group(instance, 'managers', ['can_translate_locale', 'can_manage_locale'], instance.code)  # noqa
```
adngdb: What this ``# noqa`` for? I guess you can drop it now.

mathjazz: The line is too long. I'll fix that.

PR #711, opened 2017-09-11
Fix bug 1397893: Show project and locale name in title element of the…

jotes:
Hi @mathjazz, could you review? Thanks in advance.

### DIFF
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -3666,6 +3666,8 @@ var Pontoon = (function (my) {
       self.entities = entitiesData.entities;
       self.hasNext = hasNext;
 
+      self.updateTitle();
+
       // No entities found
       if (!self.entities.length) {
         if (!self.requiresInplaceEditor()) {
@@ -3973,7 +3975,7 @@ var Pontoon = (function (my) {
           queryParams = {};
 
       self.state = state;
-      self.updateTitle();
+
       // Keep homepage URL
       if (window.location.pathname === '/' && state.project === 'pontoon-intro') {
         url = '/';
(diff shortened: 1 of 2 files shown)


pontoon/base/static/js/translate.js
```diff
@@ -3962,7 +3973,7 @@ var Pontoon = (function (my) {
           queryParams = {};
 
       self.state = state;
-
+      self.updateTitle();
```
mathjazz: You should add a blank line after this statement, to separate the block from the next one, which starts with a comment.

Actually, `pushState` is not the place to put `updateTitle` in - it's called everytime you open a new string for editing or load new entities, and we don't need to update the title on this occasion.

You should move it to the end of the `processEntities()` function.

jotes:
@mathjazz done.

pontoon/base/static/js/translate.js
```diff
@@ -4117,6 +4130,7 @@ Pontoon.attachEntityListHandlers();
 Pontoon.attachEditorHandlers();
 Pontoon.attachBatchEditorHandlers();
 
+Pontoon.updateTitle();
```
mathjazz: We don't need that anymore.

pontoon/base/static/js/translate.js
```diff
@@ -3953,6 +3955,17 @@ var Pontoon = (function (my) {
 
 
     /*
+     * Update title of the current view.
+     */
+    updateTitle: function() {
```
jotes: @mathjazz Do you wish to keep that function or should I move it into  process entities?

mathjazz: I'd keep it. The PR looks OK to me.

PR #713, opened 2017-09-12
[DNM] Bug 1397725 - Add context to newrelic transactions.

adngdb:
@mathjazz I am not really confident about how to test this. I see we have newrelic set up for our staging environment, I guess that would be the best way to test it? What do you think? 

The goal of this is to get more insights about those traces we have for the entities page, and be able to better understand what path is slow in our code. Also, it's a first step making it easier to add more context later on.


pontoon/base/utils.py
```diff
@@ -653,3 +659,14 @@ def build_translation_memory_file(creation_date, locale_code, entries):
...
+
+
+def add_newrelic_context(**kwargs):
+    try:
+        for key, value in kwargs.items():
+            newrelic.agent.add_custom_parameter(key, value)
+    except AttributeError:
+        # If newrelic-admin isn't used, for example in local development,
+        # then newrelic.agent will not be set.
+        log.warning('Unable to add context to newrelic')
```
mathjazz: This might get a little bit too verbose over time once we add more context?

pontoon/base/views.py
```diff
@@ -1,8 +1,8 @@
 import logging
 import os
+from datetime import datetime
 
 from bulk_update.helper import bulk_update
-from datetime import datetime
```
mathjazz: I'm curious, why did you move this?

pontoon/base/utils.py
adngdb: That should only happen in environments where we do not have newrelic enabled, and as far as I know, that means only dev. I guess we can do with excessive amounts of logs in dev?

pontoon/base/views.py
adngdb: Well ``datetime`` is part of the standard library, so it should be with the other imports from that standard library, at the top.

pontoon/base/utils.py
mathjazz: OK, let's keep it.

pontoon/base/views.py
mathjazz: Oh, OK. For some reason I got misled by the `from` into thinking this is not a standard library, but `form` has nothing to do with it of course.

adngdb:
This is not working yet. I'm trying to figure out how we set up NewRelic in heroku, to understand what might be wrong. I suspect we have some incompatibilities around the moving parts related to NewRelic.

adngdb:
I'm still struggling with this, I've [asked for help on NewRelic's forum](https://discuss.newrelic.com/t/newrelic-has-no-attribute-agent-with-newest-versions/51867).

adngdb:
Closing this for now. I can't make any progress here without help from NewRelic, and they don't seem to be willing to help, sadly. Also, this is not as important as I thought it was.

PR #714, opened 2017-09-12
Fix bug 1399166: Use X-channel repo as source

mathjazz:
Cross-channel repository is already limited to localizable files only, and it also includes files from the comm-* repositories.

### DIFF
--- a/bin/mozilla-en-US.py
+++ b/bin/mozilla-en-US.py
@@ -113,7 +113,7 @@ dname = os.path.dirname(abspath)
 os.chdir(dname)
 
 # Clone or update source repository
-url = 'ssh://hg.mozilla.org/users/axel_mozilla.com/gecko-strings/'
+url = 'https://hg.mozilla.org/users/axel_mozilla.com/gecko-strings/'
 target = 'source'
 pull(url, target)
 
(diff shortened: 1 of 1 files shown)


Pike:
I'd not want to merge this just yet, and I also wonder about some edge cases in the script.

In particular, for as long as I might still trash the gecko-strings repo, you should clean up the target directory on a failed pull, so that it's empty for a clone.

For maintainability, it'd be nice to still have the directories in single lines, just in case we add or remove stuff.
And I wonder if we should prune all directories in the target repo, and not just those that we have.

Both of these might make it easier to remove devtools, if they move out of tree before we have actual project configuration up.

bin/mozilla-en-US.py
```diff
@@ -117,45 +112,33 @@ def push(path):
...
-        folders = TARGET_REPOS[repo]['folders']
-        source = TARGET_REPOS[repo]['source'] + '-' + channel
+    # Clone or update target repository
+    pull(url, target)
 
-        for folder in folders:
-            origin = os.path.join('source', source, folder, 'locales/en-US')
-            destination = os.path.join('target', ending, folder)
+    # Prune all subdirectories in target repository in case they get removed from source
+    for root, subdirs, files in os.walk(target, True):
```
Pike: Using os.walk is a bit excessive? How about just listdir? 'cause I think that's what you're ending up with?

PR #715, opened 2017-09-12
Reset screenshot on string navigation

mathjazz:
If a string contains a link to a JPG or PNG file in its comment, we
render the linked image as a thumbnail in the source panel. If the
subsequently openned string does not have a comment, we keep the
screenshot visible instead of removing it. This patch fixes that.


mathjazz:
@jotes r?

PR #717, opened 2017-09-20
Refactor the entities view into smaller functions.

adngdb:
@mathjazz I was reading through this code to make sure I actually understand it well, in order to make sense of our optimization goal, and thought that I could make some things slightly better. One thing leading to another, I ended up with this not-so-small refactoring of the ``entities`` view. 

I do not know what to make of this. Maybe it isn't valuable to merge it, as surely it introduces risk (I'm not confident the current code coverage protects well enough against regression in a code that is that complex). I'm also not sure it's even that much better than the current code. I do like the use of a ``Form`` in that view, but I'm not fond of the separation in functions. I guess what I would really want to do is to split this view into 3 or 4 independent views, instead of having one view doing 3 or 4 different things, but that would be even more work. 

So, I'll defer that to you. What do you think? Is it worth investing more time here?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -1953,20 +1953,21 @@ class Entity(DirtyFieldsMixin, models.Model):
             )[0]
             return translation.serialize()
 
-        else:
-            return {
-                'fuzzy': False,
-                'string': None,
-                'approved': False,
-                'pk': None,
-            }
+        return {
+            'fuzzy': False,
+            'string': None,
+            'approved': False,
+            'pk': None,
+        }
 
     @classmethod
-    def for_project_locale(self, project, locale, paths=None, statuses=None,
-                           search=None, exclude=None, extra=None, time=None, authors=None):
+    def for_project_locale(
+        self, project, locale, paths=None, status=None,
+        search=None, exclude_entities=None, extra=None, time=None, author=None,
+    ):
         """Get project entities with locale translations."""
 
-        # Time & authors filters have to be applied before the aggregation
+        # Time & author filters have to be applied before the aggregation
         # (with_status_counts) and the status & extra filters to avoid
         # unnecessary joins causing performance and logic issues.
         pre_filters = []
@@ -1984,21 +1985,21 @@ class Entity(DirtyFieldsMixin, models.Model):
             else:
                 raise ValueError(time)
 
-        if authors:
-            pre_filters.append(Entity.objects.authored_by(locale, authors.split(',')))
+        if author:
+            pre_filters.append(Entity.objects.authored_by(locale, author.split(',')))
 
         if pre_filters:
             entities = Entity.objects.filter(pk__in=Entity.objects.filter(Q(*pre_filters)))
         else:
             entities = Entity.objects.all()
 
-        if statuses:
+        if status:
             # Apply a combination of filters based on the list of statuses the user sent.
             status_filter_choices = ('missing', 'fuzzy', 'suggested', 'translated')
             post_filters.append(
                 combine_entity_filters(
                     status_filter_choices,
-                    statuses.split(','),
+                    status.split(','),
                     locale,
                     no_plurals
                 )
@@ -2058,8 +2059,8 @@ class Entity(DirtyFieldsMixin, models.Model):
 
         entities = entities.prefetch_resources_translations(locale)
 
-        if exclude:
-            entities = entities.exclude(pk__in=exclude)
+        if exclude_entities:
+            entities = entities.exclude(pk__in=exclude_entities)
 
         return entities.distinct().order_by('resource__path', 'order')
 
--- a/pontoon/base/tests/test_views.py
+++ b/pontoon/base/tests/test_views.py
@@ -31,7 +31,7 @@ from pontoon.base.tests import (
     TestCase,
     UserFactory,
 )
-from pontoon.base.views import build_translation_memory_file
+from pontoon.base.utils import build_translation_memory_file
 
 
 class UserTestCase(TestCase):
@@ -284,7 +284,7 @@ class EntityViewTests(TestCase):
             'project': self.resource.project.slug,
             'locale': self.locale.code,
             'paths[]': [self.resource.path],
-            'inplaceEditor': True,
+            'inplace_editor': True,
             # Inplace mode shouldn't respect paging or limiting page
             'limit': 1,
         })
@@ -334,7 +334,7 @@ class EntityViewTests(TestCase):
             'project': self.resource.project.slug,
             'locale': self.locale.code,
             'paths[]': [self.resource.path],
-            'excludeEntities[]': [self.entities[1].pk],
+            'exclude_entities': [self.entities[1].pk],
             'limit': 1,
         })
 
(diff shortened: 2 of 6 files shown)


pontoon/base/models.py
```diff
@@ -1946,20 +1946,21 @@ def get_translation(self, plural_form=None):
...
+            'approved': False,
+            'pk': None,
+        }
 
     @classmethod
-    def for_project_locale(self, project, locale, paths=None, statuses=None,
-                           search=None, exclude=None, extra=None, time=None, authors=None):
+    def for_project_locale(
+        self, project, locale, paths=None, status=None,
+        search=None, exclude_entities=None, extra=None, time=None, author=None,
```
adngdb: Note that here I'm changing some of those parameters names, to make them coherent throughout the code base.

pontoon/base/tests/test_views.py
```diff
@@ -334,7 +334,7 @@ def test_exclude_entities(self):
             'project': self.resource.project.slug,
             'locale': self.locale.code,
             'paths[]': [self.resource.path],
-            'excludeEntities[]': [self.entities[1].pk],
+            'exclude_entities': [self.entities[1].pk],
```
adngdb: This puzzles me. Why was it ``excludeEntities[]`` and not ``excludeEntities``? I assume it was a mistake, and the test actually doesn't really test this feature, sadly. :(

pontoon/base/views.py
```diff
@@ -72,7 +67,10 @@ def home(request):
             return redirect('pontoon.teams.team', locale=user.profile.custom_homepage)
 
     locale = utils.get_project_locale_from_request(request, project.locales) or 'en-GB'
-    path = Resource.objects.filter(project=project, translatedresources__locale__code=locale)[0].path
+    path = Resource.objects.filter(
+        project=project,
+        translatedresources__locale__code=locale
+    ).values_list('path', flat=True)[0]
```
adngdb: There's a small optimization hidden here: I added ``values_list`` because we only care about the ``path`` field here, and nothing else.

pontoon/base/views.py
```diff
@@ -426,22 +450,29 @@ def get_translations_info(translations, changed_entities=None):
...
+            entity=t.entity,
+            translation=t,
+            project=project,
+        )
+        for t in (
+            Translation.objects
+            .filter(pk__in=changed_translation_pks)
+            .prefetch_related('entity__resource')
+        )
+    ]
```
adngdb: This is simply PEP8-compliant rewriting.

pontoon/base/views.py
```diff
@@ -519,9 +550,14 @@ def unapprove_translation(request):
...
-            request.user == translation.user or
-            translation.approved):
+    if not (
+        request.user.can_translate(
+            project=translation.entity.resource.project,
+            locale=translation.locale
+        ) or
+        request.user == translation.user or
+        translation.approved
+    ):
```
adngdb: Same here, making this PEP8-compliant.

pontoon/base/tests/test_views.py
jotes: @adngdb from what I recall that was because of the way Jquery serializes querystring parameters.

adngdb: That's the case for ``paths`` but ``exclude_entities`` is not a list, it's a string. The JS side does a ``.join(',')`` before sending it to the server, so really that ``[]`` shouldn't exist. I assume it was a leftover from a previous change.

pontoon/base/views.py
```diff
@@ -168,111 +166,137 @@ def authors_and_time_range(request, locale, slug, part):
...
+    # make a new dict with only the keys we want to pass to that function.
+    restrict_to_keys = (
+        'paths', 'status', 'search', 'exclude_entities', 'extra', 'time', 'author',
+    )
+    form_data = {k: form.cleaned_data[k] for k in restrict_to_keys if k in form.cleaned_data}
 
     entities = Entity.for_project_locale(
-        project, locale, paths, status, search, exclude_entities, extra, time, author
+        project, locale, **form_data
     )
```
mathjazz: Nit: this could be in one line.

```python
entities = Entity.for_project_locale(project, locale, **form_data)
```

pontoon/base/views.py
```diff
@@ -168,111 +166,137 @@ def authors_and_time_range(request, locale, slug, part):
...
+                entities_to_map = list(entities_to_map) + list(entities.filter(pk=entity_pk))
+
+    return JsonResponse({
+        'entities': Entity.map_entities(locale, entities_to_map, []),
+        'has_next': has_next,
+        'stats': TranslatedResource.objects.stats(
+            project, form.cleaned_data['paths'], locale
+        ),
+    }, safe=False)
+
```
mathjazz: I wonder if there's a better place to put these `_get_entities*` functions in. I don't have an idea though. :)

pontoon/base/views.py
```diff
@@ -168,111 +166,137 @@ def authors_and_time_range(request, locale, slug, part):
     }, safe=False)
 
 
+def _get_entities_list(locale, project, form):
+    """Return a specific list of entities, as defined by the `entity_ids` field of the form.
```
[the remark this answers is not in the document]
adngdb: Maybe?

pontoon/base/views.py
adngdb: Yeah, ultimately I want the ``translate`` page to live in a different app. It souldn't be in ``base``. It's a bit out of the scope of this, plus if we start rewriting that page, I'll do it in a different app, and in the end we'll be able to remove all of this from ``base``.

pontoon/base/views.py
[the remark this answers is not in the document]
adngdb: More seriously, I think convention is to always have a short description on the same line as the quotes, then add more explanations after a new line. Like this: 

```python
def foo(a, b):
    """Return something that is very useful.

    The reason why we do this is because bar. And we do it in this particular 
    way, because.

    :arg a: list of things we care about
    :arg b: string that contains something valuable

    """
    pass
```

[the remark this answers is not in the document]
adngdb: See for example the way @willkg writes his comments in Antenna: https://github.com/mozilla-services/antenna/blob/master/antenna/util.py

Btw @willkg, do you use a tool to generate API docs automatically from those comments? If so, which one?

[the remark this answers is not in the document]
jotes: I would be in favor of https://www.python.org/dev/peps/pep-0257/ or any lint-able format.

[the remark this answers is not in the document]
willkg: There's a PEP for formatting docstrings that I tend to just follow because I'm used to doing it that way:

https://www.python.org/dev/peps/pep-0257/

That talks about text in the docstring particularly conventions around the first line. Then I couple that with this:

http://www.sphinx-doc.org/en/stable/domains.html#info-field-lists

which is how I do parameters, parameter types, exceptions raised, returned values, and other things.

Doing it that way works nicely with Sphinx autodoc (``sphinx.ext.autodoc``).

Hope that helps!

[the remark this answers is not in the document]
mathjazz: Thanks @willkg @jotes @adngdb, this looks great!

Let's keep this particular comment unchanged and follow the standard.

adngdb:
Thanks for the reviews and comments everyone!

PR #721, opened 2017-09-26
Fix status icon for partially translated strings

mathjazz:
Partially translated strings use a different CSS class than missing
strings, depite the fact that the status icon looks the same. That is
because we have translations to show for the partially translated
strings, so the source strings column can only occupy half the width.

Missing strings on the other hand don't have translations (d'oh), so
the original string can occupy the entire string list width.


adngdb:
Can you provide some steps to test this? Like, where can I see the problem on production?

PR #722, opened 2017-09-27
[WIP] Bug1303066: Remove jQuery from pontoon.js

MikkCZ:
I have (partly) removed jQuery from the first functions of pontoon.js. Hitting the `loadEntities...` functions I wonder, if it wouldn't be better to redesign the script a bit. @mathjazz @jotes what do you think if I rewrite the script using some central mapping between DOM nodes <-> entities, or at least using dedicated object wrapper for passing the data between functions? This way it's not easy to find all places, where the entities are created and consumed.

### DIFF
--- a/pontoon/base/templates/js/pontoon.js
+++ b/pontoon/base/templates/js/pontoon.js
@@ -21,6 +21,25 @@
 
 
       /**
+       * In-place keyboard shortcuts handler
+       */
+      function keydownPontoonHandler(e) {
+        const key = e.keyCode,
+            toolbar = document.querySelector('.pontoon-editable-toolbar'),
+            cancel = toolbar.querySelector('.cancel');
+
+        if (toolbar.style.display !== 'none' && cancel.style.display !== 'none') {
+          if (key === 27) { // Esc: status quo
+            cancel.click();
+            hideToolbar(toolbar);
+            return false;
+          }
+        }
+      }
+
+
+
+      /**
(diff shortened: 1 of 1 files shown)


jotes:
Hey, I'll look to the code later. A single class responsible for communication between sounds interesting but could you tell a little more about it? e.g. How the mapping would look like?

MikkCZ:
At this moment the entities structures containing DOM nodes are created on multiple places. One option would be to at least wrap this to an object and use dedicated methods for the data manipulation a retrieval.
The other option I think about it to use some kind of [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) to hold the mapping and avoid the method calls completely. This can streamline the code to something like _detect entities_ > _put them in a map_ > _read the map_ > _make entities localizable_.

The first approach will probably heat the GC a bit by creating short living instances. The second may on the other hand result in a big structure residing in the memory, but I think both are not a problem in case of a single page.

mathjazz:
Hey @MikkCZ, what's the status of this PR?

We're tidying the open PR list, so If you want to continue at a later time, I'd rather close this one now.

MikkCZ:
Hi @mathjazz . Thank you for ping. I have completely forgot about this PR and probably won't have time to continue before Austin. Feel free to close it and I will create a new PR, once it is ready again.

mathjazz:
Thanks!

PR #724, opened 2017-09-27
Fix bug 1382538 - Bump python-fluent to 0.4.2

jotes:
@mathjazz @stasm 
I tested that upgrade with my version of pontoon-intro. If you remember more suitable projects/edge-cases - feel free to point them out/review.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -37,8 +37,8 @@ py-bcrypt==0.3 \
     --hash=sha256:af12f99454e895e9e430c75723c3b9419c0bccd0ea77803662d6a9ac56f6425b
 pyasn1==0.1.7 \
     --hash=sha256:e4f81d53c533f6bd9526b047f047f7b101c24ab17339c1a7ad8f98b25c101eab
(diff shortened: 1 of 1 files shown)


mathjazz:
I think the latest version is 0.5.0.

We should also update fluent-syntax.js as documented:
http://mozilla-pontoon.readthedocs.io/en/latest/dev/contributing.html#integration-with-fluent

Then I'll test with projects on prod.

stasm:
> I think the latest version is 0.5.0.

The latest version of `fluent-syntax` (in JS) is 0.5.0. The latest version of `python-fluent` is 0.4.2.  The versions of those two packages are not lined up and they implement the same syntax support. Sorry if this is confusing: we follow semver and sometimes a change in the API that's specific to the language of the implementation requires a bigger version bump.

jotes:
@mathjazz fluent-syntax is in the newest stable version (0.5.0)

mathjazz:
@stasm Thanks for clarifying!

@jotes Sorry for misguiding you!

The code is r+. Before I approve the PR, I'll test on stage using prod data, because we don't have sufficient test coverage.

PR #725, opened 2017-09-28
Fixes bug 1384139 - Fix all remaining PEP8 violations in Pontoon.

adngdb:
It's a brave new world! 

Note: I need to verify if we run pylama on builds in travis. @jotes @mathjazz do you know if we do?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ install:
 before_script:
   - psql -c 'create database pontoon;' -U postgres
 script:
-  - pylama
+  - pylama pontoon
   - python manage.py test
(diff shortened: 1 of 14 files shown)


jotes:
@adngdb We do from what I recall.

adngdb:
@jotes I added it a while ago, when I started working on [bug 1384139](). This PR removes the two exceptions that we had (and adds a new one that bothered me). 

Btw, I can't add you to reviewers, but feel free to do it if you like!

jotes:
@adngdb Sorry, I forgot :(

PR #728, opened 2017-10-05
Fix bug 1377099 - Speed up status filters

adngdb:
There is one last test that is failing, and it's a mocking issue. I'll finish this next week, but that shouldn't have any impact on the actual code, so @mathjazz if you want to review this now, feel free! 

Note that I also haven't tested this with the Search across all projects feature yet.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -1721,179 +1721,89 @@ class Subpage(models.Model):
 
 
 class EntityQuerySet(models.QuerySet):
-    """
-    Queryset provides a set of additional methods that should allow us to filter entities.
-    """
-
-    def with_status_counts(self, locale):
-        """
-        Helper method that returns a set with annotation of following fields:
-            * approved_count - a number of approved translations in the entity.
-            * fuzzy_count - a number of fuzzy translations in the entity.
-            * suggested_count - a number of suggested translations in the
-              entity.
-            * unreviewed_count - a number of translations in the entity that
-              have not yet been approved or rejected.
-            * rejected_count - a number of rejected translations in the entity.
-            * expected_count - a number of translations that should cover
-              entity.
-            * unchanged_count - a number of translations that have the same
-              string as the entity.
-        """
-        return self.annotate(
-            approved_count=Sum(
-                Case(
-                    When(
-                        Q(
-                            translation__approved=True,
-                            translation__fuzzy=False,
-                            translation__locale=locale
-                        ), then=1
-                    ), output_field=models.IntegerField(), default=0
-                )
-            ),
-            fuzzy_count=Sum(
-                Case(
-                    When(
-                        Q(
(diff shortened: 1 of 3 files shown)


pontoon/base/models.py
```diff
@@ -1721,179 +1721,256 @@ def __str__(self):
...
+                self
+                .exclude(string_plural='')
+                .filter(pk__in=translations.values('entity'))
+                .prefetch_related(Prefetch(
+                    'translation_set',
+                    queryset=(
+                        Translation.objects
+                        .filter(locale=locale)
+                        .order_by('approved', 'fuzzy', '-date')
                     ),
```
mathjazz: I think `I think `queryset=translations.order_by('approved', 'fuzzy', '-date')` would work here. And then you don't need `.filter(pk__in=translations.values('entity'))` above.

pontoon/base/models.py
```diff
@@ -1721,179 +1721,256 @@ def __str__(self):
...
-                            translation__string=F('string_plural')
-                        ), then=1
-                    ), output_field=models.IntegerField(), default=0
+                    to_attr='fetched_translations'
+                ))
+            )
+
+            # Now that we have all those translations, we'll want to extract just the
+            # active one and then make sure it matches the `rule`. If it does, we store
+            # it to be retrieved in the finale query.
```
mathjazz: Nit: final.

pontoon/base/models.py
```diff
@@ -1721,179 +1721,256 @@ def __str__(self):
...
         )
 
-    def translated(self, locale, no_plurals):
-        if no_plurals:
-            approved = Translation.objects.filter(locale=locale, approved=True).values('entity')
-            return Q(pk__in=approved)
+    def rejected(self, locale):
+        """Return a filter to be used to select entities with rejected translations.
+
+        This filter will return any entities that have a rejected translation, whether
```
mathjazz: s/any/all

adngdb:
@mathjazz comments addressed, tested again with #520, this seems to be working great! And I've filed a bug for the unchanged filter issue.

pontoon/base/models.py
```diff
@@ -1749,22 +1749,21 @@ def get_filtered_entities(self, locale, query, rule, match_all=True):
             # so that the first one for each plural form will be the active one.
             plural_candidates = (
                 self
+                .filter(
+                    resource__translatedresources__locale=locale,
+                    obsolete=False
+                )
```
mathjazz: This makes things slow (e.g. slower than current master when filtering Firefox/All Resources).

If you revert 3f668c7 instead, we're good.

adngdb:
@mathjazz reverting means a test fails, because of mocking issues. I spent a good part of the day fighting it, and I don't see any easy way of solving it without rewriting half the app, so I'm skipping it for now. :)

PR #730, opened 2017-10-09
Fix bug 1406974: Fix file upload

mathjazz:
@adngdb or @jotes r?

See [bug 1406974](https://bugzilla.mozilla.org/show_bug.cgi?id=1406974) for more details.


adngdb:
@mathjazz could you maybe add a unit test that ensures this problem will be caught if it were to happen again? 

Otherwise, r+.

jotes:
:+1: for a regression test. Also I wonder if moving that concatenation into a property could make the testing easier and hide that difference between types.

mathjazz:
This is indeed a little awkward without proper file upload tests, so I did what @jotes suggested.

PR #731, opened 2017-10-11
Fix Bug1387972 Set lang and dir attributes on l20n inputs

nathanielnebel:
Add dir, and lang attributes to l20n inputs https://bugzilla.mozilla.org/show_bug.cgi?id=1387972 .

### DIFF
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -69,7 +69,7 @@ var Pontoon = (function (my) {
                   '<label class="id built-in" for="ftl-id-' + this + '">' +
                     '<span>' + this + ' (e.g. </span><span class="stress">' + Pontoon.locale.examples[pluralForm] + '</span>)<sub class="fa fa-remove remove" title="Remove"></sub>' +
                   '</label>' +
-                  '<input class="value" id="ftl-id-' + this + '" type="text" value="' + value + '" dir="' + Pontoon.locale.direction + '" data-script="' + Pontoon.locale.script + '"  lang="' + Pontoon.locale.code + '">' +
+                  self.inputValueElement(this, value) +
                 '</li>');
           });
 
@@ -96,7 +96,7 @@ var Pontoon = (function (my) {
             maxlength = '1';
             input = '<div class="accesskeys"></div>';
           }
-          input += '<input class="value" id="ftl-id-' + id + '" type="text" value="' + value + '" maxlength="' + maxlength + '" dir="' + Pontoon.locale.direction + '" data-script="' + Pontoon.locale.script + '" lang="' + Pontoon.locale.code + '">';
+          input += self.inputValueElement(id, value, maxlength);
 
           if ($.inArray(id, [entityAttributes])) {
             label = '<label class="id" for="ftl-id-' + id + '">' +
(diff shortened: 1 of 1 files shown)


pontoon/base/static/js/fluent_interface.js
```diff
@@ -106,7 +106,7 @@ var Pontoon = (function (my) {
           } else {
             cls = ' class="custom-attribute clearfix"';
             label = '<div class="wrapper">' +
-              '<input type="text" class="id" placeholder="enter-attribute-id" value="' + id + '">' +
+              '<input type="text" class="id" placeholder="enter-attribute-id" value="' + id + '" dir="' + Pontoon.locale.direction + '" data-script="' + Pontoon.locale.script + '"  lang="' + Pontoon.locale.code + '">' +
```
mathjazz: This element is actually used for the attribute `id`, so no need to add the attributes here.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -69,7 +69,7 @@ var Pontoon = (function (my) {
                   '<label class="id built-in" for="ftl-id-' + this + '">' +
                     '<span>' + this + ' (e.g. </span><span class="stress">' + Pontoon.locale.examples[pluralForm] + '</span>)<sub class="fa fa-remove remove" title="Remove"></sub>' +
                   '</label>' +
-                  '<input class="value" id="ftl-id-' + this + '" type="text" value="' + value + '">' +
+                  '<input class="value" id="ftl-id-' + this + '" type="text" value="' + value + '" dir="' + Pontoon.locale.direction + '" data-script="' + Pontoon.locale.script + '"  lang="' + Pontoon.locale.code + '">' +
```
mathjazz: Double space in front of `  lang=`.

nathanielnebel:
@mathjazz I agree with your suggestion for creating a function to generate the input element as to facilitate code reuse. The only reason I didn't do that was that the input element on line 72 didn't specify a maxlength (nor do I think a variable lives within that scope). The input element on line 99 does specify that value. If you would still believe that to be advisable I can create a function to do it and leave the parameter for max length optional and use that to return the appropriate element for each section.

mathjazz:
Yeah, I'm aware that that input elements are not exactly the same. Let's create such function.

pontoon/base/static/js/fluent_interface.js
nathanielnebel: Resolved, generation of this element is now contained in a function

pontoon/base/static/js/fluent_interface.js
nathanielnebel: Oops, reverted it.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -164,6 +164,21 @@ var Pontoon = (function (my) {
 
 
       /*
+       * Generate Input value element
+       */
+      inputValueElement: function (id, value, maxLength) {
+         var base = '<input class="value" id="ftl-id-' + id + '" type="text" value="' + value + '"';
```
mathjazz: Nit: Use 2 spaces for indentation instead of 3.

nathanielnebel:
@mathjazz Indentation fixed, does Mozilla have a styleguide for Javascript that I can read?

PR #732, opened 2017-10-11
Fixes bug 1395231 - Format requirements files.

adngdb:


adngdb:
@mathjazz ``setuptools`` is not a requirement of ``bleach`` as far as I can tell. The problem here is that we "install" ``django_nose`` in the base settings, but I moved it to the dev requirements. As travis builds for prod, the library is never installed and tests cannot run. Though now that I'm thinking about it, maybe stuff we use to run tests are needed in the "prod" config... I'm about to find out!

adngdb:
@mathjazz It's green now! :)

adngdb:
And I've verified by pushing to stage that it works. Full disclosure: it didn't, and I had to fix something else. But now it does!

PR #735, opened 2017-10-15
Fix bug 1407532 Update git submodules buildpack, refactor code respon…

jotes:
Hey @mathjazz and @adngdb!
Could you look at this patch?
Thanks in advance.


app.json
```diff
@@ -32,6 +32,10 @@
      }
    ],
    "env": {
+     "SITE_URL": {
+       "description": "Access url to the instance e.g. https://<app name>.herokuapp.com without a trailing slash.",
+       "required": true
+     },
```
mathjazz: I remember we had issues setting `SITE_URL` during build. Has that changed?

app.json
```diff
@@ -109,7 +113,7 @@
    },
    "buildpacks": [
      {
-       "url": "https://github.com/dmathieu/heroku-buildpack-submodules#b37ffe4361bb9c975dd8e93068c9d296365d748c"
+       "url": "https://github.com/dmathieu/heroku-buildpack-submodules#0caf30af7737bf1bc32b7aafc009f19af3e603c1"
```
mathjazz: Why do we need to change the buildpack? Shall we do that on production, too?

runtime.txt
```diff
@@ -0,0 +1 @@
+python-2.7.14
```
mathjazz: I wonder why do we need to specify this manually now. When deploying to production, I see this:
```
The latest version of Python 2 is python-2.7.14 (you are using python-2.7.12, which is unsupported).
We recommend upgrading by specifying the latest version (python-2.7.14).
```

app.json
```diff
@@ -32,6 +32,10 @@
      }
    ],
    "env": {
+     "SITE_URL": {
+       "description": "Access url to the instance e.g. https://<app name>.herokuapp.com without a trailing slash.",
```
mathjazz: I would be more strict here. Instead of **e.g.**, this **has to be** "https://<app name>.herokuapp.com", right?

jotes:
@mathjazz Hey, I'll update the commit message and I'll try to respond to all your questions inline.

runtime.txt
jotes: Unfortunately, I can't add any comments to this file :(
Python3 is the default python for the new apps - that probably doesn't affect existing production but it's breaking deployment of new apps if you don't specify runtime explicitly.
Related docs: https://devcenter.heroku.com/changelog-items/1169

app.json
jotes: :+1:

app.json
jotes: Previously, Pontoon was able to detect existing domain automatically after deployment.

I took that approach to simplify deployment as much as it was possible.
However, after a few versions of Django (request.get_host() throws an exception if there's no right SITE_URL defined) the process got harder.
The code was already hard to maintain and required of the magic (lazily loaded django settings).
That's why I decided to remove all unnecessary code in the favor of a small requirement from the user.

app.json
jotes: I'm not sure about the production - Heroku may work differently for already deployed apps and previous version of this particular buildpack will work perfectly fine.
However, for the new builds https://github.com/dmathieu/heroku-buildpack-submodules/blob/master/bin/install_git_submodules.rb throws an error and https://github.com/dmathieu/heroku-buildpack-submodules/commit/35622728df8757a1bacde155290fcd2960ff1a11 seems to fix that.

mathjazz: Gotcha!

mathjazz:
Thanks for explainig all this! +1 for simplifying the codebase!

The only thing I'd change is that `SITE_URL` description, other than that this is r+.

jotes:
@mathjazz could you look again?

app.json
```diff
@@ -32,6 +32,10 @@
      }
    ],
    "env": {
+     "SITE_URL": {
+       "description": "Access url to the instance, has to be https://.herokuapp.com without a trailing slash.",
```
mathjazz: `"Base URL of the site. Has to be https://{app-name}.herokuapp.com."`

jotes: done :-)

PR #736, opened 2017-10-19
Bug 1403861 - Hide Suggested and Fuzzy statistics in the API

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1403861

The Suggested status is pending a refactor in bug 1377969 and until it's
fixed, it's confusing to use `translatedStrings` for Suggested.

The Fuzzy status might get folded into Unreviewed Strings as part of bug
1377969 too. Let's remove it for now and re-add if it stays around.

The Complete flag will now only be true if all strings are Approved.
Suggested and Fuzzy strings don't count towards this flag anymore.

### DIFF
--- a/pontoon/api/schema.py
+++ b/pontoon/api/schema.py
@@ -20,8 +20,7 @@ class ProjectLocale(DjangoObjectType, Stats):
     class Meta:
         model = ProjectLocaleModel
         only_fields = (
-            'total_strings', 'approved_strings', 'translated_strings',
-            'fuzzy_strings', 'project', 'locale'
+            'project', 'locale', 'total_strings', 'approved_strings', 'fuzzy_strings'
         )
 
 
@@ -29,9 +28,8 @@ class Project(DjangoObjectType, Stats):
     class Meta:
         model = ProjectModel
         only_fields = (
-            'name', 'slug', 'disabled', 'info', 'deadline', 'priority',
-            'contact', 'total_strings', 'approved_strings',
-            'translated_strings', 'fuzzy_strings'
+            'name', 'slug', 'disabled', 'info', 'deadline', 'priority', 'contact', 'total_strings',
+            'approved_strings', 'fuzzy_strings'
         )
 
     localizations = graphene.List(ProjectLocale)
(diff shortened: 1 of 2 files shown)


mathjazz:
We have yet to decide what to do with fuzzy, so I wouldn't make changes here just yet.

Also, I'd prefer to keep the API aligned with the model (and the dashboards), so instead of removing `translatedStrings` I'd rename them and by doing that (and renaming `approvedStrings`) fix bug 1403861.

stasm:
I'm in favor of staying as close to the model as possible myself. In this case however, given that there's a change in semantics planned for the near future, I suggest we simply hide the contentious fields until we implement the new names and semantics in bug 1377969. Admittedly the API is more vulnerable to renames.

pontoon/base/models.py
```diff
@@ -479,7 +479,7 @@ def missing_strings(self):
 
     @property
     def complete(self):
-        return self.missing_strings == 0
+        return self.total_strings - self.approved_strings == 0
```
mathjazz: Good call. The current version is off.

Nit: `return self.total_strings == self.approved_strings` reads more natural.

stasm: I don't know what I was thinking :) Of course it does, thanks.

mathjazz:
As we talked on IRC, let's move forward with removing `suggested` from the API and keep `fuzzy` for now. @stas Does that make sense?

stasm:
Yes. I updated the PR accordingly.

pontoon/api/schema.py
```diff
@@ -45,9 +43,8 @@ class Locale(DjangoObjectType, Stats):
     class Meta:
         model = LocaleModel
         only_fields = (
-            'name', 'code', 'direction', 'script', 'population',
-            'total_strings', 'approved_strings', 'translated_strings',
-            'fuzzy_strings'
+            'name', 'code', 'direction', 'script', 'population', 'total_strings',
+            'approved_strings', 'fuzzy_strings'
```
adngdb: Nit, and coming after the battle, but since those are going to evolve over time, I'd recommend we put one field per line, sorted alphabetically. It's going to make following changes much much easier to read.

PR #740, opened 2017-10-24
Fix bug 1411378: Sync projects with non-ASCII names

mathjazz:


pontoon/sync/vcs/models.py
```diff
@@ -177,7 +177,7 @@ def changed_locales_files(self):
 
                 log.info(
                     'Changed files in {} repository, all: {}'.format(
```
flodolo: Python newbie passing by: wouldn't it be enough to do `u'Changed files…`?

mathjazz: That would do, too. But I'm trying to stay consistent with the rest of the code.

Here's how it works: Default string representation of the Project instance is [defined](https://github.com/mozilla/pontoon/blob/master/pontoon/base/models.py#L1059) as `self.name`. Note the [decorator](https://github.com/mozilla/pontoon/blob/master/pontoon/base/models.py#L1002) on top which makes the output `unicode`. Since `str` and `unicode` concatenation (which is what we do in this line) results in a `unicode` string, that's all we need.

HTH

flodolo: > HTH

Yep, thanks.

adngdb:
This is pretty hard to test. Do we have any unit tests for the sync app? If so, please add a new one that tests this case. Otherwise, I'll just trust you that this works and you have tested it, and give you an r+. :)

mathjazz:
We have a decent test coverage in `pontoon.sync`:
https://github.com/mozilla/pontoon/tree/master/pontoon/sync/tests

It would be great if we could find a meaningful way to test this, because its very implicit when it breaks and when it doesn't. OTOH, we should be in a much better position once we switch to python3, where strings are Unicode by default.

Any idea how to test?

mathjazz:
Still no idea how to test this in a sane way.

PR #744, opened 2017-10-30
Add dockerdebug make target to allow pdb debugging

phlax:
as in title

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,4 +40,4 @@ dockerloaddb:
 	docker exec -i `${DC} ps -q postgresql` pg_restore -U pontoon -d pontoon -O < ${DB_DUMP_FILE}
 
 dockerrun:
-	${DC} up webapp
+	${DC} run --service-ports webapp


mathjazz:
Thanks for the patch! @adngdb Do you wanna have a look?

adngdb:
@phlax reading the documentation about ``--service-ports``, I don't really understand what your use case for this is. How is this different from using ``make dockershell``? What are you trying to achieve with that command?

phlax:
@adngdb it does pretty much the same thing as `make dockerrun` - the only difference being that running this way means that you can add a `pdb` break in the code and interact with it.

`--service-ports` just binds the the containers ports to the host so you can browse to eg http://localhost:8000

adngdb:
Ah, OK, but in that case why not just add ``--service-ports`` to the ``dockerrun`` command? This docker environment is made for local development only, so I think it's fine. What do you think?

phlax:
i think that when you use `compose up` you dont need to add `service-ports` - its just when use `compose run` you have to specify the ports.

The question i was more wondering is whether we should just use the command that i added for `dockerdebug` for `dockerrun` itself. Not sure if there any downsides to that

adngdb:
Yeah, I'm wondering the same thing. I propose that you change your patch to reflect that, and then we test this locally over time and if no one ever complains it's a win. :)

phlax:
@adngdb updated, ill squash if you are happy with pr

adngdb:
Yay for your first r+ gif! Congrats! :)

PR #745, opened 2017-10-30
Fix bug 1407623 - New source input for short-term projects

adngdb:
This adds a new way to add source strings to a project, without needing a repository. Any project that has no repositories will have a link to manage strings, which offers an interface to enter new strings in a batch. Once that project has strings, it will show an interface to edit those strings or add new ones, and a link to download all translations in a single CSV file. 

There are few questions pending regarding this: 
 - what should happens if a repository is added after strings have been inserted via this new way?
 - currently, those "short-term" projects are identified by their not having a repo, but Axel mentioned in the bug that we should, in a future iteration, be able to add a repo so that strings are synchronized into it. What should be the correct way of identifying short-lived projects? Should we use a special marker in the database? Should we have a tagging system? 
 - it is not possible to delete a source string at the moment. Should that be a feature? 
 - I have not tested the sync process with those short-term projects. Maybe it will fail? Maybe not? 
 - I'm going to deploy this to stage at some point to have Peiying test the feature. 
 - there is no documentation. Do we have user/manager documentation somewhere? Is this the place? https://github.com/mozilla-l10n/documentation/tree/master/tools/pontoon


flodolo:
> there is no documentation. Do we have user/manager documentation somewhere? Is this the place? https://github.com/mozilla-l10n/documentation/tree/master/tools/pontoon

Yes, that's where we keep documentation for admin tasks

mathjazz:
> what should happens if a repository is added after strings have been inserted via this new way?
> currently, those "short-term" projects are identified by their not having a repo, but Axel mentioned in the bug that we should, in a future iteration, be able to add a repo so that strings are synchronized into it. What should be the correct way of identifying short-lived projects? Should we use a special marker in the database? Should we have a tagging system?

I would leave those questions out of the scope of this PR, unless you think that will make things harder in the future.

> it is not possible to delete a source string at the moment. Should that be a feature?

Yes. Note: in DB sense deleting would mean setting `Entity.obsolete = True`.

> I have not tested the sync process with those short-term projects. Maybe it will fail? Maybe not?

It will probably fail. We'll probably need to exclude "short-term" projects from sync, similarly to how we [exclude disabled projects](https://github.com/mozilla/pontoon/blob/master/pontoon/sync/management/commands/sync_projects.py#L60). I'll have a look as part of the code review.

adngdb:
> I would leave those questions out of the scope of this PR, unless you think that will make things harder in the future.

Yeah, I believe we need to answer those questions now. Your comment about excluding them from sync is yet another reason to do that now rather than later. But we want to think about it so that it's future proof, as it will likely be a database change. We need to consider those few things: 
 - we'll have repositories for those in the future
 - we'll want to squash all short-term projects under a meta project
 - we'll want different rules for syncing those projects (never read from repo for example, only write)

Pike:
I think the meta project work will come when we do [project config support](https://bugzilla.mozilla.org/show_bug.cgi?id=1385834), so that one I'd not worry too much about just now.

I do think we'll want repo sync support in a second step, yes. It'd be nice to set the projects up such that the current contributions are all marked as "to be synced"? I don't think we want different repo rules for them.

If "can-edit" is a flag or a tag, I don't know.

pontoon/administration/static/css/admin_project.css
```diff
@@ -121,8 +122,8 @@ form a:link, form a:visited {
   text-transform: uppercase;
 }
 
-form a.add-inline,
-form a.add-repo {
+#admin-form form a.add-inline,
+#admin-form form a.add-repo {
```
mathjazz: I think you meant:

```css
#admin-form a.add-inline,
#admin-form a.add-repo {
```

Otherwise "Add More" links on the project admin page grow in size.

adngdb:
@mathjazz The last two commits address most of your comments. Thanks, it looks much better now! Note that this commit is well described, to see what's new: https://github.com/mozilla/pontoon/pull/745/commits/f0a7f20e25995acadb5d319da74598eb9ff2719f

adngdb:
@mathjazz you mentioned "deleting" should be marking the string as "obsolete", is that a requirement? Simply deleting the string is much easier, but I assume you have good reasons to use "obsolete" instead? Could you share that with me, for my culture? :)

mathjazz:
> @mathjazz you mentioned "deleting" should be marking the string as "obsolete", is that a requirement? Simply deleting the string is much easier, but I assume you have good reasons to use "obsolete" instead? Could you share that with me, for my culture? :)

For example, if you delete an Entity, all its Translations are also gone by default, which impacts User stats for example.

Obsolete is how we treat deleted Entities for other (Repository based) projects.

mathjazz:
Good work!

I only have one suggestion. If the "Database" is selected as the "Data Source", we don't need to show the Repository section. So we can merge "Data source" and "Repository" into one "Data source" section, where "Database" will be an option under "Type". When selected, Repository fields need to be hidden.

adngdb:
> I only have one suggestion. If the "Database" is selected as the "Data Source", we don't need to show the Repository section. So we can merge "Data source" and "Repository" into one "Data source" section, where "Database" will be an option under "Type". When selected, Repository fields need to be hidden.

That is true right now, but won't be as soon as we add the ability to sync to a repository (which, I assume, will require some tweaks to the sync code). Hence, I am hesitant to add code to do what you ask, as I don't think it's harmful to let the repositories displayed, and we're very likely to remove that exact feature in a few. Do you think we should still do it now, even if we'll remove it afterwards?

mathjazz:
My understanding is that the Data Source is either the Database or the Repository - it cannot be both. So in the future, when we'll add the ability to sync a specific project to the Repository, we'll remove the ability to manage strings in the Admin.

Pike:
> My understanding is that the Data Source is either the Database or the Repository - it cannot be both.

Why should we not be able to do both? We can do bidirectional sync for l10n files, and this would simply be bidirectional sync for en-US reference?

Pike:
The twist in allowing a project to sync back en-US strings to VCS is that it assumes to have write access to VCS.

Which for many projects it might have, but not in a way that's actually constructive. I.e., if we're talking about gettext/.lang projects, the en-US templates are extracted from the code, and us writing back other English strings won't have any impact on the project. And for projects like Firefox, pontoon just won't have write access.

The other aspect wrt to sync of en-US might be that the sync algorithm assumes that there's a reference template to sync with? In that case, yes, syncing a project without a template will be an interesting challenge.

adngdb:
Hey! This is ready! Oh, wait, nope... I forgot to add documentation... :) However @mathjazz the code is complete, I think. Feel free to review at will!

adngdb:
Documentation is written, PR is here: https://github.com/mozilla-l10n/documentation/pull/120

mathjazz:
This looks good!

The only thing I would change is be explicit about excluding non-Repository projects from [sync](https://github.com/mozilla/pontoon/blob/master/pontoon/sync/management/commands/sync_projects.py).

The deal was that I fine tune the frontend design. Can I proceed now (and update the docs if needed)?

adngdb:
@mathjazz How do you want to go about that? Do you want to open a PR to my branch, or should we merge this and you'll make your changes in a subsequent PR to this repo? 

The sync thing is done.

mathjazz:
@adngdb Thanks!

I'd rather open a PR to your branch. I'm not planning to make big changes.

adngdb:
@mathjazz I think this is ready. r? :)

PR #746, opened 2017-10-30
Parametrize docker test runner.

jotes:
Hey @mathjazz & @adngdb 
Sometimes I need to run tests with different verbosity or with `--no-input` to reset the database [when database already exists] between tests runs.
This patch should pass all arguments from the makefile to django test runner.
Could you review?
Thanks in advance!

### DIFF
--- a/docker/run_tests_in_docker.sh
+++ b/docker/run_tests_in_docker.sh
@@ -61,7 +61,7 @@ else
            --env-file ./docker/config/webapp.env \
            -e LOCAL_USER_ID=$UID \
            local/pontoon \
-           python manage.py test
+           python manage.py test $@
 
     echo "Done!"
 fi


mathjazz:
LGTM. @adngdb?

adngdb:
@jotes Can you show me an example of how you would use it? I don't think it works for me, like, if I run ``make dockertest pontoon.administration`` it still runs the entire test suite.

jotes:
@adngdb I started yesterday with `make dockertest ARGS="pontoon.administration --verbose"` but it looks like I can do it with simpler syntax. Give me a moment.

jotes:
@adngdb https://stackoverflow.com/questions/6273608/how-to-pass-argument-to-makefile-from-command-line Could you look if that's what would you want or do you prefer explicit parameter like e.g. ARGS for test arguments?

adngdb:
OK, I see. So my initial reaction to this was: you should use ``make dockershell``. That's what it's made for. Basically, as I thought about this, ``dockertest`` is for running the entire test suite, no questions. Every other use case should be done with ``dockershell``, because otherwise we risk overcomplicating our makefile logic. 

However, this seems to work and doesn't disrupt anything as far as I can tell, so let's merge it.

PR #749, opened 2017-11-07
Fix bug 1392255: Enable source view for FTL messages

mathjazz:
Restore FTL button and enable switching between source and FTL editor:
- Fallback to source view for translations not supported in FTL editor.
- Same for source strings.
- Switching reflects changes in either editor.
- Prevent switching to rich editor if message not supported.
- Show translation length and Copy from source only for simple FTL strings and only in FTL editor.
- For untranslated strings not supported in rich editor, show source view if source string displayed as source.

Always use #translation textarea for source editor and #ftl-area for FTL editor (even for simple strings). Similarly, even simple source strings are displayed in #ftl-original. To make that possible, we keep entity and translation objects unchanged for non-complex FTL strings.

Other improvements:
- Bugfix: Update editor when FTL message unrejected.
- Join renderEditorWithTranslation() and renderEditor() functions.
- Focus first element, regardless if <input> or <textarea>.

To test and review, use the test project containing a good balance of FTL messages:
https://github.com/mathjazz/pontoon-ftl/

### DIFF
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -151,6 +151,7 @@ var Pontoon = (function (my) {
           }
 
           $('#translation').val(value);
+          Pontoon.updateCachedTranslation();
           Pontoon.fluent.toggleEditor(false);
           return;
         }
@@ -582,6 +583,7 @@ $(function () {
       }
 
       $('#translation').val(translation);
+      Pontoon.updateCachedTranslation();
     }
 
     Pontoon.fluent.toggleEditor($(this).is('.active'));
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -374,8 +374,8 @@ var Pontoon = (function (my) {
        * Return translation in the editor as FTL source
        */
       getTranslationSource: function () {
-        var entity = Pontoon.getEditorEntity(),
-            fallback = $('#translation').val();
+        var entity = Pontoon.getEditorEntity();
+        var fallback = $('#translation').val();
 
         // For non-FTL entities, return unchanged translations
         if (entity.format !== 'ftl') {
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -167,6 +167,11 @@ var Pontoon = (function (my) {
 
         Pontoon.fluent.focusFirstField();
         Pontoon.fluent.toggleEditorToolbar();
+
+        Pontoon.moveCursorToBeginning();
+        Pontoon.updateCurrentTranslationLength();
+        Pontoon.updateInPlaceTranslation();
+
         return true;
       },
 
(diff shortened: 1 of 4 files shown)


adngdb:
I found a bug. Steps to reproduce: 

1. load the Pontoon FTL translation page
2. open the "About Localization" entity, you can see the textarea to edit
3. open the "Delete this download?" entity
4. come back to the previous entity: the textarea doesn't appear anymore.

Note that in my case, I have submitted translations for the "Delete this download" entity, in case that matters.

adngdb:
Adding the link to my PR with edits and style changes here: https://github.com/mathjazz/pontoon/pull/6

adngdb:
mathjazz#6

mathjazz:
Nice catch! Updated. If the simple string was untranslated, we didn't open the simple editor when switching back from the complex one.

adngdb:
New bug found, steps to reproduce: 

1. Go to Pontoon FTL, open last entity "to get started"
2. Copy/paste the original content, translate, save
3. Open the entity again, change something, save
4. Open the entity again, notice the key appears twice, like this: `warning-upgradewarning-upgrade = { ...`

Note that, from a user experience standpoint, I find it disturbing to have a blank textarea when I open that entity. Most of the time I expect the user will copy and paste the original content to have the same structure. Could it be possible to do something better? The options imo are: 
1. auto-fill with the English string
2. auto-fill with `key = `
3. auto-fill with the structure of the original string, in this case: `warning-upgrade = { LINK("", title: "") }`

The first is the easiest one, and it might also actually be the best one in terms of UX. What do you think?

mathjazz:
Updated. Regarding the textarea - I agree we should change that. I'm actually leaning towards #2.

#1 would be inconsistent with the default (rich) editor.
#3 might might encourage localizers against the Isolation [principle](https://github.com/projectfluent/fluent/wiki/Design-Principles).

Pike:
Also +1 to option 2. Suggesting the English string will make it a lot easier to end up with left-over English strings or fragments thereof.

mathjazz:
@adngdb Could you review the two commits I added?

adngdb:
I suspect the only reason why you added more code to this PR after having an r+ is because you are disappointed by the gif I used. Fair enough. I'll get you a better one then...

mathjazz:
Found some more bugs. @adngdb Could you please review and find another GIF pretty please?

pontoon/base/static/js/translate.js
```diff
@@ -706,11 +706,11 @@ var Pontoon = (function (my) {
       }
 
       var before = this.cachedTranslation,
-          after = $('#translation').val();
+          after = this.fluent.serializeTranslation(entity, $('#translation').val());
```
adngdb: In most places you've replaced `$('#translation')` with `$('#editor textarea:visible:first')`, why not here?

mathjazz: Here, we compare the old (`before`) and the current (`after`) state of the string by comparing the source of the string, which is stored in `#translation`.

mathjazz:
Updated and ready for the next look.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -370,6 +370,37 @@ var Pontoon = (function (my) {
       },
 
 
+      /*
+       * Return translation in the editor as FTL source
+       */
+      getTranslationSource: function () {
+        var entity = Pontoon.getEditorEntity(),
+            fallback = $('#translation').val();
```
adngdb: Please put one `var` per line. :)

mathjazz:
I keep forgetting this!

PR #751, opened 2017-11-09
Fix bug 1415979: No TM entries for fuzzy imports

mathjazz:
@jotes r?


pontoon/sync/changeset.py
```diff
@@ -371,11 +371,11 @@ def bulk_update_translations(self):
     def bulk_create_translaton_memory_entries(self):
         """
         Create Translation Memory entries for:
-            - new translations
+            - new approved translations
             - updated translations that are approved and don't have a TM entry yet
         """
         translations_to_create_translaton_memory_entries_for = (
-            self.translations_to_create +
+            [t for t in self.translations_to_create if t.approved] +
```
jotes: Existing entries created from fuzzy translations should be removed from the database, am I right? That would require a migration.

mathjazz: Good point. Upadated!

PR #753, opened 2017-11-10
Fix 1416237 - Caighdean integration

phlax:
Integrate the caighdean machine translation engine into Pontoon


pontoon/base/static/js/main.js
```diff
@@ -513,6 +513,33 @@ var Pontoon = (function (my) {
...
+	    key: $('#editor')[0].entity.key,
+            locale: self.locale.code
+          }
+        }).success(function(data) {
+          if (data.translation) {
+            append({
+              url: 'https://github.com/kscanne/caighdean',
+              title: 'Caighdean Machine Translation',
+              source: 'Caighdean Machine Translation',
+              translation: data.translation
```
mathjazz: It would be useful to also add the `original` parameter here, so that we see what string was used as the input.

pontoon/machinery/views.py
```diff
@@ -113,6 +117,32 @@ def machine_translation(request):
...
+    except Entity.DoesNotExist as e:
+        return HttpResponseBadRequest('Bad Request: {error}'.format(error=e))
+    try:
+        text = entity.translation_set.get(locale__code="gd").string
+    except Translation.DoesNotExist:
+        return JsonResponse({})
+    try:
+        return JsonResponse(
+            dict(locale=locale_code,
+                 translation=Translator().translate(text)))
```
mathjazz: You will make Adrian very happy if you format this like that:
```
return JsonResponse(
    dict(
        locale=locale_code,
        translation=Translator().translate(text)
    )
)
```

:-)

I don't have a strong opinion on this, but I'd use `{}` instead of `dict()`.

pontoon/machinery/views.py
```diff
@@ -113,6 +117,32 @@ def machine_translation(request):
...
+    except TranslationError as e:
+        return HttpResponseBadRequest('Bad Request: {error}'.format(error=e))
```
mathjazz: Please ass blank lines between `try-except` blocks. And try use `'` instead of `"` where possible.

requirements.txt
```diff
@@ -272,3 +272,7 @@ argparse==1.3.0 \
     --hash=sha256:b3a79a23d37b5a02faa550b92cbbbebeb4aa1d77e649c3eb39c19abf5262da04
 diff-match-patch==20121119 \
     --hash=sha256:9dba5611fbf27893347349fd51cc1911cb403682a7163373adacc565d11e2e4c
+
+# ----------------------------------------------------------------------------------
+
+-e git+https://github.com/phlax/python-caighdean#egg=caighdean
```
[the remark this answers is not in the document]
phlax: > This fails to install:

It seems that you cant use an edit install if you are checking hashes.

For this repo, ill make a pypi release.

pontoon/base/static/js/main.js
phlax: i think the original would be the `gd` translation that is sent to the MT. We could add it in the response, but it wouldnt be used currently

phlax:
> Tests would be nice. But I wouldn't block on that, because we have no tests in the Machinery app anyways.

Im definitely thinking to add some. I was hoping to land #755 first, as then its easier to see coverage

pontoon/machinery/views.py
```diff
@@ -113,6 +117,30 @@ def machine_translation(request):
...
+    """Get translation from Caighdean machine translation service."""
+    try:
+        entityid = int(request.GET['id'])
+    except (MultiValueDictKeyError, ValueError) as e:
+        return HttpResponseBadRequest('Bad Request: {error}'.format(error=e))
+
+    try:
+        entity = Entity.objects.get(id=entityid)
+    except Entity.DoesNotExist as e:
+        return HttpResponseBadRequest('Bad Request: {error}'.format(error=e))
```
adngdb: Why not use ``get_object_or_404`` here?

pontoon/machinery/views.py
```diff
@@ -113,6 +117,30 @@ def machine_translation(request):
         return HttpResponseBadRequest('Bad Request: {error}'.format(error=e))
 
 
+def machine_translation_caighdean(request):
+    """Get translation from Caighdean machine translation service."""
+    try:
+        entityid = int(request.GET['id'])
+    except (MultiValueDictKeyError, ValueError) as e:
+        return HttpResponseBadRequest('Bad Request: {error}'.format(error=e))
```
adngdb: This is simply a question: when do you use a Form, and when do you not? I guess if there's just one field, then you don't bother creating a new class for handling that? Do you have a rule for when to use a Form?

tests/machinery/views.py
```diff
@@ -0,0 +1,81 @@
...
+
+    response = client.get(url, dict(id="DOESNOTEXIST"))
+    assert response.status_code == 400
+    assert (
+        response.content
+        == ("Bad Request: invalid literal for int() "
+            "with base 10: 'DOESNOTEXIST'"))
+
+    maxid = Entity.objects.values_list(
+        "id", flat=True).order_by("-id").first()
```
adngdb: This is a good example of why I dislike not putting closing brackets/parentheses on a new line. It is quite difficult to understand where the closing parenthesis of the first line is. To me, this is much more readable:

```python
maxid = Entity.objects.values_list(
    "id", flat=True
).order_by("-id").first()
```

I don't have to think to see where the function's arguments end, which is not the case with the way you wrote it.

pontoon/machinery/views.py
phlax: probs that would be better, ill update

pontoon/machinery/views.py
phlax: imho - use a form by default - unless you are using pretty much the unchanged resolved keywords from the url

phlax: i think this would be better implemented in a form, incidentally

pontoon/machinery/views.py
phlax: hmm, im also wondering whether the other bad responses, should be JSON

pontoon/machinery/views.py
```diff
@@ -113,6 +117,30 @@ def machine_translation(request):
...
+    try:
+        text = entity.translation_set.get(locale__code='gd').string
+    except Translation.DoesNotExist:
+        return JsonResponse({})
+
+    try:
+        translation = Translator().translate(text)
+        return JsonResponse({'original': text, 'translation': translation})
+    except TranslationError as e:
+        return HttpResponseBadRequest('Bad Request: {error}'.format(error=e))
```
adngdb: Yeah, it is weird to return a BadRequest here. It's not the user's fault if there are not translations available for that input.

phlax: yep

pontoon/machinery/views.py
mathjazz: So in this particular case (if you go with `get_object_or_404`) `HttpResponseBadRequest` won't be used, right?

In general, I don't know. Is there a good reason to move the output to JSON?

phlax: this is what i was thinking to do...

```
    try:
        entity = get_object_or_404(Entity, id=entityid)
    except http.Http404 as e:
        return http.HttpResponseNotFound(
            json.dumps(dict(error=404, message=str(e))),
            content_type='application/json')
```

I think the reason it *should* respond with JSON, is because (i figure) the client makes the request with `application/json` as the `Content-Type` and expects that - certainly that is what is returned in the case of success

phlax: ...from a practical point of view, its also a pita when XHR requests get full page errors back

if you are debugging it, you have to sift through the unrendered html to find what the error was. 

In this case, ie simple get_object_or_404 it probs wouldnt matter o much

mathjazz: LGTM

PR #754, opened 2017-11-10
[WIP] Fixes #1416256 Add string tiers to Pontoon

phlax:
This is the integration PR for the string tiers, its ~frequently rebased on the other PRs and is used to ensure that they work together correctly.

For initial deployment of string tiers functionality the following are required:

- [X] Add react/webpack to the environment
- [X] Add basic models required
- [x] Add utils for efficient faceting of tag data #766 
- [x] Add project admin view that allows tags to be linked, and for resources to be linked to tags
    - [x] javascript for admin view #815 
    - [x] view and form for api #823 
- [x] Add project view showing a list of linked tags and a breakdown with stats per locale #807 

Following initial deployment the following priorities have been identified
- [X] Filtering with tags in translate views

This has been completed and a new bug for follow up features has been added (https://bugzilla.mozilla.org/show_bug.cgi?id=1449854)


mathjazz:
I wonder if we can simply reuse the currently unused `priority` field, which we also use in the `Project` model.

The question boils down to whether we need the name for a tier/priority or not. Do we, @flodolo?

flodolo:
That's one of the open questions in the doc. I think we should be fine with priority, but I haven't heard back from @Pike. @guerojeff seems to be on board after the explanation.

phlax:
imho it makes much more sense to break out the tiers (or however we name them).

i think the big reason to separate is that the system can be much more extensible. Now we want to mark the tiers for their priority, but in the future we also might want to give tiers other attributes, or functionality.

by putting in a separate model it makes the orm queries more manageable, and allows us to adapt on the Tier class - as here https://github.com/mozilla/pontoon/pull/754/files#diff-ffe4fd5e0ca04672f83d24cecda1a3faR94

im also wondering whether `Tier` should be `Tag` and to allow a resource to be marked with multiple tags. That way it would be much more extensible.

Pike:
I've left an open question in the google doc. I can think of tiers as something that localizations might want to opt out of. Or do we want to use sub-projects for that, strictly?

pontoon/base/static/js/translate.js
```diff
@@ -1137,6 +1137,7 @@ var Pontoon = (function (my) {
         status: [],
         extra: [],
         time: '',
+	tier: [],
```
adngdb: Wrong alignment.

pontoon/base/models.py
```diff
@@ -1654,6 +1654,11 @@ class Resource(models.Model):
     project = models.ForeignKey(Project, related_name='resources')
     path = models.TextField()  # Path to localization file
     total_strings = models.PositiveIntegerField(default=0)
+    tier = models.ForeignKey(
+        "Tier",
```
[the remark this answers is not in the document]
phlax: I tend to use dqs by preference in python - but im not so bothered about it - ill update

pontoon/base/static/js/translate.js
phlax: damn tabs!

pontoon/teams/views.py
```diff
@@ -37,9 +38,11 @@ def teams(request):
 def team(request, locale):
     """Team dashboard."""
     locale = get_object_or_404(Locale, code=locale)
+    tiers = Tier.objects.count() or False
```
Pike: Should this only get the tiers that are enabled for the projects that this locale is active in?

pontoon/teams/views.py
```diff
@@ -74,6 +77,30 @@ def ajax_info(request, locale):
...
+    if code or locale:
+        locale = get_object_or_404(Locale, code=code or locale)
+    else:
+        locale = None
+    if slug:
+        project = get_object_or_404(Project, slug=slug)
+        tiers = Tier.objects.filter(resource__project=project).distinct()
+    else:
+        project = None
+        tiers = Tier.objects.all()
```
Pike: Similar here, should this only list the tiers in the projects a locale is active in, if the locale is given?

mathjazz:
Adding some general suggestions on the project dashboard changes here:

1. I suggest we move the *Tags* tab to the 2nd position, next to *Teams*. They serve the same purpose and are the most important tabs in the project dashboard.
screen shot 2018-02-01 at 15 37 03

2. A click on a specific Tag should open a Project-Tag dashboard.
screen shot 2018-02-01 at 15 56 59

It looks very much like a project dashboard, but broken down by the selected tab:
1. The `<title>` tag (browser tab title) and the header title reflect the Tag name.
1. Tag priority and Project priority can be both seen in the heading section.
1. Stats (pie chart and the legend) reflect the selected tag only.
1. Contents of the Teams tab should look & feel the same as on the Project dashboard, but again - the stats and the latest activity should only reflect the selected tab.
1. We could add additional tabs like Contributors (filtered by the selected tag only) and Info with more details about the tag.

I'm not sure about the URL structure. Should we go with `/projects/firefox/tags/a-tag/` (more consistent with existing URLs like project-locale dashboard and translate view or `/projects/firefox/tags/a-tag/` (less error prone).

@flodolo @phlax @Pike I'd love to hear your thoughts.

flodolo:
I agree on both 1. and 2.
As for the URL, `/projects/firefox/tags/a-tag/` seems indeed clearer, also considering `/projects/firefox/tags/` is the TAGS tab.

phlax:
@mathjazz 1 seems like an #easyfix

2 i have a template for from when i had added the site-wide tags, so shouldnt be too difficult.

The only question im left with is the navigation both in terms of url and in terms of tabs

mathjazz:
I'm trying to collect the exact list of things that remain to be done within this PR - feature wise.

I'm starting with the [Admin](https://mozilla-pontoon-staging.herokuapp.com/admin/projects/firefox/):
- [ ] The slug field should autcomplete after the name is entered. We should use the same logic already available for the Project slug. And switch order of the name and the slug field.
- [ ] If I try to select a resource in the selector, it disappears (together with the manage resources button).
- [ ] CSS enhancements to make the resource selector look & feel more like the rest of the Admin UI. [Matjaž]

Next, I have a few questions for @flodolo:
* Did we say we'll use the same priority levels for files as we use for projects or not? If yes, I suggest we use the same widget in the admin UI for selecting priority - dropdown.
* Could you confirm that the first iteration of string tiers (this patch) doesn't include filtering of strings by tags or priority?
* Could you confirm that the first iteration of string tiers (this patch) doesn't include priority column on the [ProjectLocale](https://pontoon.mozilla.org/sl/firefox/) dashboard?

flodolo:
> Next, I have a few questions for @flodolo:
>     Did we say we'll use the same priority levels for files as we use for projects or not? If yes, I suggest we use the same widget in the admin UI for selecting priority - dropdown.

In the end [we decided](https://docs.google.com/document/d/1jaqRQtq1VSZkzm0E4z6OyOaajkoSvZc-f1HuhNOqdBo/edit?ts=5a033c53#heading=h.p4e48o52qw05) to use 1-5 like projects, even if the original proposal had 3 levels.

>     Could you confirm that the first iteration of string tiers (this patch) doesn't need to include filtering of strings by tags or priority?

I think that's the fundamental part for localizers to be able to actually use tiers. But, in the last meeting, I recall we decided that it would be added after the first landing to make things easier from a development perspective.

To be honest, I'm not sure it makes a lot of sense to start adding resources to tiers before we have that, besides for testing reason.

>     Could you confirm that the first iteration of string tiers (this patch) doesn't need to include priority column on the ProjectLocale dashboard?

What's the ProjectLocale dashboard? The localization page? I assume that's OK, for the same reason as the previous point.

mathjazz:
Thanks for clarifying all that!

PR #755, opened 2017-11-10
Fixes #1416269 Add coverage to Pontoon

phlax:
its really helpful to see what code is/not tested in PRs, to see what is currently not tested, and to monitor over time.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -94,9 +94,6 @@ pytz==2017.2 \
     --hash=sha256:39504670abb5dae77f56f8eb63823937ce727d7cdd0088e6909e6dcac0f89043 \
     --hash=sha256:ddc93b6d41cfb81266a27d23a79e13805d4a5521032b512643af8729041a81b4 \
     --hash=sha256:f5c056e8f62d45ba8215e5cb8f50dfccb198b4b9fbea8500674f3443e4689589
-requests==2.6.2 \
-    --hash=sha256:8f0f56813f82d0c27d9578221268ac9af48f076c71ee69693305ceca6ca355bd \
-    --hash=sha256:0577249d4b6c4b11fd97c28037e98664bfaa0559022fee7bcef6b752a106e505
 scandir==1.5 \
     --hash=sha256:c2612d1a487d80fb4701b4a91ca1b8f8a695b1ae820570815e85e8c8b23f1283
 six==1.10.0 \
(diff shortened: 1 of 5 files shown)


mathjazz:
Let's try this! IIRC @adngdb was also in favour when we discussed this recently?

Shall we also integrate it with GitHub?

phlax:
> Shall we also integrate it with GitHub?

I added submitting to codecov in the test run - we might want to deconfigure it a bit as it can be noisy

requirements.txt
```diff
@@ -272,3 +272,10 @@ argparse==1.3.0 \
     --hash=sha256:b3a79a23d37b5a02faa550b92cbbbebeb4aa1d77e649c3eb39c19abf5262da04
 diff-match-patch==20121119 \
     --hash=sha256:9dba5611fbf27893347349fd51cc1911cb403682a7163373adacc565d11e2e4c
+
+# ----------------------------------------------------------------------------------
+
+coverage==4.4.2	\
+    --hash=sha256:eea9135432428d3ca7ee9be86af27cb8e56243f73764a9b6c3e0bda1394916be
+codecov==2.0.9 \
+    --hash=sha256:ad82f054837b02081f86ed1eb6c04cddc029fbc734eaf92ff73da1db3a79188b
```
adngdb: I assume those have no sub-requirements, can you add them to the first section of this file, in alphabetical order?

phlax:
> I was expecting to find something in the PR checks, but there's nothing that leads to codecov.

codecov is submitted to at the end of the travis run

> how are we going to access the coverage reports?

all pull requests will have a coverage report in the "All checks have passed" (or otherwise) section

> Does it say nothing if coverage goes up or doesn't change, and only "complain" if it goes down?

dep on how we configure it, it will fail PRs that have either low coverage in the PR, or that lower overall coverage. Otherwise it will show as passing and provide a link to view the coverage report

you can also check coverage directly here https://codecov.io/gh/mozilla/pontoon/

phlax:
@adngdb ive separated out the test requirements into a separate file, i dont think its necessary to install this if you are not running tests and checking/submitting coverage

We might want to use `extra_requires` for this, but as we dont currently use `install_requires` i figured follow existing.

requirements.txt
```diff
@@ -117,6 +114,28 @@ https://github.com/mathjazz/silme/archive/v0.9.3.zip#egg=silme==0.9.3 \
 # Requirements with sub-dependencies.
 # ==================================================================================
 
+requests==2.18.4 \
```
adngdb: Ideally this should be lower in the requirements file, in order to follow alphabetical order.

phlax: cool, ill update

PR #756, opened 2017-11-13
[WIP] Fix bug 1229891 - Log every changes in an user roles.

jotes:
Hey @mathjazz, @adngdb 
This isn't fully reviewable yet.
However,  I wanted to ask for your opinion on following things:
* django-enumfields - I added new package and introduced the static way of saving enum values.
* Where I should put methods like e.g. `log_users_roles`? (I'm referring to one of the previous comments of @adngdb)

TODO:
- [ ] Add tests
- [ ] Add missing docstrings
- [ ] Fix pylama


mathjazz:
I'd probably just use the `ChoiceField` here, but I don't have an opinion against `enum`.

Regarding the code structure and where to put which function, maybe `pontoon.sync.models` can be used as an inspiration? It's used for the `Sync` log ATM.

adngdb:
I agree with @mathjazz, I think you are over-engineering the problem. A simple list of strings would suffice, and a ``ChoiceField`` seems good enough. It's fine if it's not the responsibility of the database to enforce that the "action" is correct. 

Regarding the ``log_user*`` functions, they should indeed be functions, or at least be outside of a Manager class. Where to put them, however, I'm not sure. Do you have opinions, @mathjazz and @phlax ?

jotes:
@adngdb Yeah, I decided to use enumfields in the last minute before push of this PR. I'm not a fan of passing harcoded strings (because of mispells, refactoring later etc.) in the code as I think it's safer to pass static identifiers as the general pattern.
That's why I decided to change it/discuss it with you. I can change it to choices.

adngdb:
Closing this for now, please reopen when it's ready for review!

PR #757, opened 2017-11-14
Fixes #1417026 Add pytest

phlax:
Adds py.test setup and some basic fixtures

There are a couple of side effects of adding this:

- slows down test run
- by default py.tests coverage includes test files

re the first, i think we can live with that temporarily.

re including test files, i think that its useful to have coverage on tests/fixtures as sometimes its possible to assume a condition is being matched, or a loop contains some items, but coverage can indicate otherwise.

### DIFF
--- a/tests/fixtures/resources.py
+++ b/tests/fixtures/resources.py
@@ -1,0 +1,16 @@
+
+import pytest
+
+from pontoon.base.models import Resource
+
+
+@pytest.fixture
+def resource0(project0):
+    return Resource.objects.create(
+        project=project0, path="resource0.po", format="po")
+
+
+@pytest.fixture
+def resource1(project1):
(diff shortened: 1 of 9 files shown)


tests/fixtures/resources.py
```diff
@@ -0,0 +1,10 @@
+
+import pytest
+
+
+@pytest.fixture
+def resource0(project0):
+    from pontoon.base.models import Resource
```
adngdb: Why do you import those at runtime?

phlax: it might not be necessary, but often the fixtures are loaded before django has been configured - ill check

phlax: turns out this isnt necessary - imports moved

PR #758, opened 2017-11-15
Touch bug 1417530 Shift remaining base tests to pytest

phlax:
Inititally, shift the base tests suite to py.test


adngdb:
Oh, I didn't notice you added those tests in a ``tests`` folder at the root. I'm not sure I actually like that. All django projects I've worked with have tests in app folders, and I think that makes it easier to discover, understand and maintain. What's your rational for moving them to a different folder?

phlax:
> What's your rational for moving them to a different folder?

putting tests together makes it easier on writing tests, switching between existing tests to crib code or check how something works etc.

previously i treated every app as its own testable unit - as you would in a more component oriented framework like zope. The last few projects i have worked on have tended to do it this way, and tbh i think it makes more sense, because those apps are supposed to work together, and are not so likely to be broken to a separate package.

another benefit if you have coverage on the tests is that then you can see straight away what the percentages for tests are (should never  be below 100% with few exceptions) and what the percentages on production code are

adngdb:
I still think it is a bad idea to move tests to a single folder. However, I am willing to give it a try, so I won't object to merging your work because of that. I do want to explain why I think that though. Here it comes! 

I think doing this embraces the monolithic model of software development, which I dislike. By considering that all parts of our code work together (which they do, of course), we allow ourselves to couple them more tightly. I believe this is opening a door to more and more spaghetti code, with dependencies going all over the place. In my short career as a developer, every time this has happened, we've ended up having to fight it somehow, with a lot of very difficult refactoring. 

We've had an example of that very recently with the Terminology patch. I've raised alarms about the code in that patch adding a lot of complexity to the app by tightly tying existing parts of the code to new ones. We've discussed this and chosen to go for a more encapsulated solution, with one app dedicated to handling the entire terminology logic, and exposing a simple API that other processes can call. In this very case, it would make no sense to have the tests live outside the terminology app. 

So, I believe we developers should aim for more isolated code, because that's easier to maintain and understand. My opinion is that we are not doing a very good job of it (that's been the case in all the django projects I've worked on), and that we should enforce harsher rules on ourselves instead of embracing the easier way. Pretty sure if we were to ask Yoda if this way is stronger, he'd say "No! No…no. Quicker. Easier. More seductive." :smile: 

The one point that makes sense to me is the coverage thing, but it still feels wrong to be adapting your code to the needs of a tool, instead of doing the opposite. 

@phlax, you have a lot more experience than I do, so really, I'm interested in your thoughts on this. That said, once again, I'm not going to r- on that. On to the actual review now! :)

phlax:
> I think doing this embraces the monolithic model of software development, which I dislike.

yep, me too - but im not sure that moving the tests this does so much.

This is not an argument for monolithic development, just to separate tests where they are packaged and tested.

I tend to use a file structure in the test env that mirrors the structure in the codebase, so breaking code/tests out somewhere would not be so difficult from a test perspective at least.

At the same time, Django does tend towards a more monolithic style of development with interdependent apps, mostly because python lacks native `interfaces` and `adapters`, which would be needed to create stable but loose coupling between apps.

Python does now have the `AbstractBaseClass` which goes some way to a native `Interface` but it still does not have the magic :sparkles: adapter resolution that eg `zope.component` provides.

Even if python did have these things, its not a common pattern to use aspect-orientation in Django anyway.

The situation in Pontoon is even more monolithic.

If you look here https://codecov.io/gh/mozilla/pontoon/ - there are mostly 2 apps, and one contains all of the models - none of which are defined as `AbstractModels` nor have any way to be swapped/overridden. Even Django's `get_user_model` is rarely used.

> In this very case, it would make no sense to have the tests live outside the terminology app.

Again, this is not an argument against stable APIs or breaking Pontoon into more discrete apps - just to place the tests separate within the package/repo.

> The one point that makes sense to me is the coverage thing, but it still feels wrong to be adapting your code to the needs of a tool, instead of doing the opposite.

Being able to visualize the code and tests separately is defo handy, but not by any means the only advantage.

Another huge advantage is that you can do...

```
git grep foo tests/
git grep bar pontoon/
```
To search for things in in tests/pontoon respectively. I do this very, very frequently, and commonly rebase tests/code together respectively.

If you practice TDD and are trying to get/maintain high coverage, the tests themselves need some maintenance/refactoring/love from time to time, having them together makes this a less painful task.

The more spread out the tests are the more of a dev pita this is, in zope/plone they have this https://pypi.python.org/pypi/collective.recipe.omelette to mitigate this, but its not perfect (incidentally, i ported this to general python - https://pypi.python.org/pypi/scrambler/0.0.3)

Another upside thats standing out to me is how to approach refactoring. I try to refactor either tests or code - one side or the other, one at a time. Having the code/tests in separate folder trees tends i believe to encourage you to think of the code/tests separately and to be trying to balance each side.

phlax:
another advantage...

You dont need to use the `test_foo.py` automagic, and the paths are shorter easier to navigate

Running some tests you can tab-tab-run eg.
```
py.test -vv -k tests/someapp/views.py
```

adngdb:
Thanks for the lengthy explanations here @phlax! Much appreciated! 

What is the status of this PR? Has it been made obsolete by the smalled ones you've open?

phlax:
This PR is waiting on #762 to land

adngdb:
#762 has landed, but this has a conflict. What's the status now? Is it still a WIP?

phlax:
> #762 has landed, but this has a conflict. What's the status now? Is it still a WIP?

it was waiting for the view tests to move (#796 ) which now has

phlax:
@adngdb can we clear this one ?

tests/base/signals.py
```diff
@@ -0,0 +1,23 @@
...
+    assert project_locale0.project.has_changed is False
+    project_locale0.latest_translation = None
+    project_locale0.project.save()
+    project_locale0.project.refresh_from_db()
+    assert project_locale0.project.has_changed is False
+    assert project_locale0.latest_translation is None
+    project_locale0.latest_translation = translation0
+    project_locale0.save()
+    project_locale0.project.refresh_from_db()
+    assert project_locale0.project.has_changed is False
```
adngdb: All those `x is False` should be replaced with `not x`.

tests/base/utils.py
```diff
@@ -92,3 +95,29 @@ def test_get_m2m_mixed(user0, user1, userX):
         get_user_model().objects.filter(pk__in=[user1.pk]),
         get_user_model().objects.filter(pk__in=[userX.pk, user0.pk]),
     ) == ([user0, userX], [user1])
+
+
+def test_util_base_extension_in():
+    assert extension_in('filename.txt', ['bat', 'txt']) is True
+    assert extension_in('filename.biff', ['biff']) is True
+    assert extension_in('filename.tar.gz', ['gz']) is True
```
adngdb: What is the point of using `is True` here? Functionally, it can be removed.

tests/base/utils.py
```diff
@@ -92,3 +95,29 @@ def test_get_m2m_mixed(user0, user1, userX):
...
+
+    assert extension_in('filename.txt', ['png', 'jpg']) is False
+    assert extension_in('.dotfile', ['bat', 'txt']) is False
+
+    # Unintuitive, but that's how splitext works.
+    assert extension_in('filename.tar.gz', ['tar.gz']) is False
```
adngdb: `not x` instead of `x is False`.

tests/base/adapter.py
```diff
@@ -0,0 +1,43 @@
...
+
+from mock import MagicMock
+
+
+# We have to support customized adapter during the transition of accounts
+# between fxa and persona.
+
+
+@pytest.mark.django_db
+def test_adapter_base_get_connect_normal_auth_account(social_adapter0):
```
adngdb: `social_adapter0` is only used in this file, wouldn't it make sense to define it here? I'm not confident enough with pytest to know if that's even possible, but if it is, I think it would make it easier to understand those tests. I've found fixtures to be pretty obscure in our tests, so putting them closer to where they're being used would help imo.

phlax: i think its a pretty common pattern in pytest to separate out the fixtures - but if its a blocker - i can move here - there's no technical reason why not

tests/base/utils.py
phlax: well - i would make 2 points here - this is what the existing test did under the hood, and sometimes you want to know an api returns a bool - not eg the filename or somesuch

i have no qualms either way tho - so ill update now...

tests/base/adapter.py
adngdb: Not a blocker, it simply was not obvious to me, and I actually made different comments because I misunderstood the fixture -- until I reached the end of the PR and saw the fixture itself, realizing my comments where completely off. 

So, I'm just sending that comment your way, in case you agree it would make it easier to understand, or if you have a better idea.

phlax: i think my reluctance is that most of the fixtures are together now - so you might not see it if you needed to use this. That said - i want to make it as easy as possible to understand, so ill move it here for now, and we can move it to fixtures if it does need to be re-used

phlax: updated

PR #759, opened 2017-11-16
No bug - Enable Sync button in admin for local development.

adngdb:
Running sync locally can be difficult, so allowing to do so from the admin interface is quite convenient for local development. This commit makes that the default behavior.

### DIFF
--- a/docker/config/webapp.env.template
+++ b/docker/config/webapp.env.template
@@ -2,6 +2,7 @@ SECRET_KEY=insert_random_key
 DJANGO_DEV=True
 DJANGO_DEBUG=True
 DATABASE_URL=postgres://pontoon:asdf@postgresql/pontoon
+MANUAL_SYNC=True
 SESSION_COOKIE_SECURE=False
 SITE_URL=#SITE_URL#
 FXA_CLIENT_ID=2651b9211a44b7b2


docker/config/webapp.env.template
```diff
@@ -2,6 +2,7 @@ SECRET_KEY=insert_random_key
 DJANGO_DEV=True
 DJANGO_DEBUG=True
 DATABASE_URL=postgres://pontoon:asdf@postgresql/pontoon
+MANUAL_SYNC=True
```
phlax: if it wants to be `True` by default in a dev env, what about making it equal `DEBUG` or somesuch by default

adngdb:
Huh, @mathjazz, that didn't occur to me... It is a big problem indeed. Well, the big problem is using prod db dumps, to be honest. Maybe we should invest in fixing that problem, and make sure devs cannot accidentally break anything on prod?

mathjazz:
The problem is also that some of us have write access to repos used on prod. So even if we stop using the prod DB, it doesn't prevent us from adding a project manually and syncing by mistake.

adngdb:
Well, if we can create a good dataset of fake data, that covers most of our use cases, we could go without using any of the real projects we have on prod. We can even setup some fake repos to verify sync works locally, etc. It's some work, but as far as I know, it's not undoable (unlike e.g. Socorro).

adngdb:
@mathjazz I was just thinking... What you're saying makes sense with a local install, since the environment contains your SSH key and git configuration. But we're talking docker here, so the sync process would run inside a container that, and I hope I'm not wrong here, has no access to either your SSH key or your git identity. So, I believe this is actually fine to land, as writing by mistake is very unlikely to happen. Do you agree with that?

mathjazz:
That is correct. 👍

PR #763, opened 2017-11-21
Touch bug 1417530 Add fixtures

phlax:
- Add test fixtures 
- move the core fixtures into pytest-django's `django_db_blocker` context

this allows the the fixtures to be created once, and rolled back if required


tests/base/fixtures.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+@pytest.mark.django_db
+def test_fixtures_base0(entity0, resource0, project0, pl0,
+                        translation0, locale0):
+    assert entity0.resource == resource0
+    assert entity0.resource.project == project0
+    assert translation0.entity == entity0
+    assert pl0.project == project0
+    assert pl0.locale == locale0
```
adngdb: I think `pl` should be `project_locale`, especially if it's a fixture. It might be used in contexts where it will be much harder to guess what `pl` means.

tests/fixtures/factory.py
```diff
@@ -0,0 +1,28 @@
+
+import pytest
+
+from django.db.models import Max
+
+
+def _factory(**kwargs):
```
adngdb: I think this function deserves a little docstring. Mostly, the arguments are what I would love to see documented. :)

tests/fixtures/projects.py
```diff
@@ -4,11 +4,17 @@
 from pontoon.base.models import Project
 
 
+@pytest.mark.env
```
adngdb: I don't understand this. Why are you marking a fixture?

tests/fixtures/resources.py
```diff
@@ -6,11 +6,15 @@
...
 
 
 @pytest.fixture
 def resource1(project1):
+    return Resource.objects.get(project=project1, path="resource1.po")
+
+
+@pytest.fixture
+def resourceX(projectX):
     return Resource.objects.create(
```
adngdb: Why is this a `create` still and not a `get` like the others? You are creating `projectX` in the Environment.

tests/base/fixtures.py
phlax: after typing `project_locale` a lot i changed it to this - my assumption is that people reading/writing tests in pontoon will know the convention

tests/fixtures/projects.py
phlax: it was a failed attempt to use markers to automatically include in the `post_db_setup` fixture

phlax: ...ill remove

tests/fixtures/resources.py
phlax: because any fixture with `X` at end has minimal setup in the env, so in this case its not already created

eg. userX has no contributions and resourceX has no entities, etc

phlax: actually, `resourceX` was a bad example, because `projectX` has no resources, and therefore `resourceX` doesnt exist

tests/fixtures/entities.py
```diff
@@ -7,22 +7,33 @@
...
+@pytest.fixture
+def entity_factory(factory):
+    """Entity factory
+    create entities in a hurry!
+
+    Provides an entity factory function that accepts the following args:
+
+    :param `batch`: number of entities to instantiate, defaults to len of
+        `batch_kwargs` or 1
+    :param `batch_kwargs`: a list of kwargs to instantiate the entities
```
adngdb: This is not consistent with what we already do, we use `arg` instead of `param`. See here for example: https://github.com/mozilla/pontoon/blob/master/pontoon/base/models.py#L48-L61

tests/base/fixtures.py
adngdb: I insist that it should be `project_locale`. We all use good IDEs with completion, and `pl0` reads a lot like `p10`, which makes it even more confusing. And I favor making it harder to write than read, because we read a lot more than we write.

phlax:
@adngdb updated

PR #764, opened 2017-11-23
Bug 1411955 - Upgrade to Graphene 2.0

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1411955


stasm:
@Pike I fixed the noisy test output by overwriting `sys.excepthook` as per https://docs.python.org/2/library/sys.html#sys.excepthook.

PR #765, opened 2017-11-26
Fix bug 1416890: Save proper state of translators and managers after …

jotes:
Hey @mathjazz , @adngdb !

Could you review this small fix?

### DIFF
--- a/pontoon/teams/static/js/permissions.js
+++ b/pontoon/teams/static/js/permissions.js
@@ -2,12 +2,15 @@ $(function() {
   var container = $('#main .container');
 
   function inputHidden(name, value) {
-    return $('<input type="hidden" name="' + name + '" value="' + value + '">');
+    return $('<input class="permissions-form-item" type="hidden" name="' + name + '" value="' + value + '">');
   }
 
   container.on('click', '#permissions-form .save', function(e) {
     e.preventDefault();
     var $form = $('#permissions-form');
+
+    // Remove stale permissions items (bug 1416890)
+    $('input.permissions-form-item').remove();
 
     // Before submitting the form, update translators and managers
     $.each(['translators', 'managers'], function(i, value) {


pontoon/teams/static/js/permissions.js
```diff
@@ -2,13 +2,16 @@ $(function() {
...
-    return $('<input type="hidden" name="' + name + '" value="' + value + '">');
+    return $('<input class="permissions-form-item" type="hidden" name="' + name + '" value="' + value + '">');
   }
 
   container.on('click', '#permissions-form .save', function(e) {
     e.preventDefault();
     var $form = $('#permissions-form');
 
+    // Remove stale items.
+    $form.find('input.permissions-form-item').remove();
```
adngdb: Is this different from `$('input.permissions-form-item', $form).remove();`? Note that this is just me being curious, it's the first time I see it written like this.

mathjazz: Same thing. :)

pontoon/teams/static/js/permissions.js
```diff
@@ -2,13 +2,16 @@ $(function() {
...
   function inputHidden(name, value) {
-    return $('<input type="hidden" name="' + name + '" value="' + value + '">');
+    return $('<input class="permissions-form-item" type="hidden" name="' + name + '" value="' + value + '">');
   }
 
   container.on('click', '#permissions-form .save', function(e) {
     e.preventDefault();
     var $form = $('#permissions-form');
 
+    // Remove stale items.
```
mathjazz: Please add a reference to the bug.

jotes: Done :)

pontoon/teams/static/js/permissions.js
jotes: I thought it may be a little faster than `$()` in this case. However, maybe your notation will be easier to read and more explicit.

PR #768, opened 2017-11-28
Fix bug 1421317 - Expose current git revision sha in web app.

adngdb:
@jotes @mathjazz r? 

I haven't tested it on heroku yet, because I don't want to mess with our stage server now. So I'm counting on your knowledge of heroku to confirm that this is the right way to implement that feature. :)

### DIFF
--- a/bin/post_compile
+++ b/bin/post_compile
@@ -5,6 +5,11 @@ export PATH=/app/.heroku/node/bin:$PATH
 ./manage.py migrate --noinput
 ./manage.py collectstatic --noinput
 
+# Create a file with the current git HEAD revision.
+# We use that file to expose the git revision in our web app, allowing
+# us to use services like whatsdeployed to compare what is deployed to stage and prod for example.
+git rev-parse HEAD > static/revision.txt
+
 # Inform New Relic that a deploy is happening.
 if [ -n "${NEW_RELIC_API_KEY}" ] && [ -n "${NEW_RELIC_APP_NAME}" ]; then
     echo "Sending deploy notification to New Relic...."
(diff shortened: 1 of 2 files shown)


mathjazz:
Deployed to stage. File is there, but it's empty:
https://mozilla-pontoon-staging.herokuapp.com/static/revision.txt

Heroku toolbelt says:
```bash
Snowball:mozilla-pontoon mathjazz$ heroku run --app mozilla-pontoon-staging bash
Running bash on ⬢ mozilla-pontoon-staging... up, run.1825 (Standard-1X)
~ $ git rev-parse HEAD
fatal: Not a git repository (or any parent up to mount point /app)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

bin/post_compile
```diff
@@ -5,6 +5,11 @@ export PATH=/app/.heroku/node/bin:$PATH
 ./manage.py migrate --noinput
 ./manage.py collectstatic --noinput
 
+# Create a file with the current git HEAD revision.
+# We use that file to expose the git revision in our web app, allowing
+# us to use services like whatsdeployed to compare what is deployed to stage and prod for example.
+git rev-parse HEAD > static/revision.txt
```
jotes: That probably won't work: https://stackoverflow.com/questions/14583282/heroku-display-hash-of-current-commit

adngdb:
Thanks to @jotes' link, I think I figured out a solution. We would need to enable Dyno Metadata, see https://devcenter.heroku.com/articles/dyno-metadata

@mathjazz can I go ahead and test this on stage?

mathjazz:
Done. And confirmed to be working via Heroku Toolbelt - `HEROKU_SLUG_COMMIT` matched the deployed commit hash.

mathjazz:
I wonder if we should make the commit ID in the revision file link to `https://github.com/mozilla/pontoon/commit/HEROKU_SLUG_COMMIT` for better visibility of what's deployed.

adngdb:
Updated and verified working on stage. Thanks @mathjazz !

PR #769, opened 2017-11-29
Add related_name on base resources

phlax:
this is useful for doing table joins on these fields

TranslatedResource.latest_translation
ProjectLocale.latest_translation
Locale.latest_translation
Project.latest_translation


adngdb:
Sorry to jump in too late, but this should definitely have had a bug. Because it's generally a good practice, there should be very few exceptions of commits not related to a bug, but in this case, there's a migration. A bug with some explanations of what's happening and why really was needed in my opinion.

phlax:
@adngdb ok, fair point. As it was a minor change i thought it would be ok to just PR, but i will defo ticket if there are migrations in future

PR #770, opened 2017-11-29
Touch bug #1417530 Add more fixtures

phlax:
Adds:

- some further fixtures for base models
- a `factories` fixtures for easier accessing of fixtures
- a `site_matrix` fixture that loads a load of related projects/locales/project_locales/resources/translated_resources etc

### DIFF
--- a/tests/fixtures/site.py
+++ b/tests/fixtures/site.py
@@ -1,0 +1,41 @@
+import pytest
+
+
+@pytest.fixture
+def site_matrix(factories):
+    projects = factories['project'](batch=3)
+    locales = factories['locale'](batch=3)
+    project_locale_kwargs = []
+    resource_kwargs = []
+    translated_resource_kwargs = []
+    for project in projects:
+        resources = []
+        for i in range(0, 7):
+            resources.append(
+                dict(project=project,
+                     total_strings=(i + 5) ** 3))
+        for locale in locales:
+            project_locale_kwargs.append(
+                dict(project=project, locale=locale))
+            for i, resource in enumerate(resources):
+                translated_resource_kwargs.append(
+                    dict(locale=locale,
+                         approved_strings=(i + 2) ** 3,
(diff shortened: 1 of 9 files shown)


tests/fixtures/site.py
```diff
@@ -0,0 +1,39 @@
+import pytest
+
+
+@pytest.fixture
+def site_matrix(factories):
+    projects = factories['project'](batch=3)
+    locales = factories['locale'](batch=3)
+    pl_kwa = []
+    resource_kwa = []
+    tr_kwa = []
```
adngdb: I assume `kwa` stands for `kwargs`? And I'd rather have long and clear variable names than acronyms, for the sake of people who will read this later and will have to figure out / remember what `pl` and `tr` stand for. 

As a side note, Elon Musk hates acronyms, too: https://pbs.twimg.com/media/CF4cFtLUgAASqYO.png :P

tests/fixtures/site.py
```diff
@@ -0,0 +1,39 @@
...
+                    dict(locale=locale,
+                         approved_strings=(i + 2) ** 3,
+                         translated_strings=(i + 1) ** 3,
+                         fuzzy_strings=i ** 3))
+        resource_kwa += resources
+    project_locales = factories['project_locale'](batch_kwargs=pl_kwa)
+    resources = factories['resource'](batch_kwargs=resource_kwa)
+    [tr.update(dict(resource=resources[i % len(resources)]))
+     for i, tr
+     in enumerate(tr_kwa)]
```
adngdb: This is weird. Why not just make a for loop if you're not using the data structure you're creating?

phlax: i got carried away with the functional programming - i can update

tests/fixtures/site.py
phlax: these are local variables in a test, surely we can be a little bit relaxed here

PR #771, opened 2017-11-29
Fix bug 1421495: Add SVG favicon

mathjazz:
Adding the HiRes SVG favicon for nicer display in the Firefox's new tab.

For more details, see https://bugzilla.mozilla.org/show_bug.cgi?id=1421495.

We still need to use the old `.ico` favicon:
https://caniuse.com/#feat=link-icon-svg

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ before_script:
   - psql -c 'create database pontoon;' -U postgres
 script:
   - pylama pontoon
+  - python manage.py collectstatic --noinput
   - python manage.py test --with-coverage
   - py.test --cov-append --cov-report=term --cov=. -v
   - codecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ install:
   - pip install -U --force pip
   - pip install --require-hashes -r requirements.txt
   - pip install --require-hashes -r requirements-test.txt
+  - npm install
 before_script:
   - psql -c 'create database pontoon;' -U postgres
 script:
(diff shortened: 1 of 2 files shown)


MikkCZ:
I am not sure if this won't break the favicon in Chromium. If a favicon is not defined in `<head>`, it surely checks `/favicon.ico`. But does it do so if the favicon is explicitly defined, but it's in not supported format? Personally I would expect it won't fallback to `/favicon.ico` anymore.

phlax:
is it not possible to specify a `.ico` first and then browsers fall back to that ?

MikkCZ:
I cannot tell right at the moment, if this will break in Chromium or not. Alternatively you can use `<link rel="icon">` together with an image with bigger resolution in `<link rel="apple-touch-icon">` (but the level of `.svg` support here is the same). But let's check if the current approach works in Chromium or not.

mathjazz:
Nice cacth, @MikkCZ! Indeed it didn't work in Chrome. Now it does.

@phlax The SVG icon is already in the repository.

mathjazz:
The patch is now ready from my side.

I've fixed that tests by adding `collectstatic` to travis. That requires `yuglify`, so I also needed to add `npm install`.

phlax:
not sure if we should use "shortcut icon" for better IE compatibility. I dont have any versions of IE to test tho so lgtm

mathjazz:
Thanks for the review and the [link](https://pontoon.mozilla.org/static/img/favicon.ico)!

We should be good: https://pontoon.mozilla.org/static/img/favicon.ico.

PR #774, opened 2017-12-01
Fix bug 1421495: Define favicon type

mathjazz:
Without the type defined, Firefox uses the last defined favicon. See bug 1422289 for more details.

### DIFF
--- a/pontoon/base/templates/base.html
+++ b/pontoon/base/templates/base.html
@@ -10,7 +10,8 @@
     <meta name="description" content="Live web localization tool">
     <meta name="author" content="Mozilla">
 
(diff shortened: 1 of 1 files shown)


mathjazz:
@MikkCZ Could you review this?

I've deployed the patch to stage:

MikkCZ:
@mathjazz I have tested in current Nightly and it works. The favicon in Chromium is displayed too. Still there is one small caveat, the tile in activity stream looks "broken" because it does not fill the whole square.
snimek z 2017-12-01 15-52-06

mathjazz:
Thanks!

There's a white border surrounding the "P", which is part of the icon.

I agree - we need a better logo. :) But that's a different bug.

PR #775, opened 2017-12-06
Fix bug 1423460: Do not hardcode plural selector

mathjazz:
@flodolo Thanks for noticing and filing the bug! Could you confirm that we also need to re-export all strings that are broken because of this bug?

### DIFF
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -316,7 +316,8 @@ var Pontoon = (function (my) {
             }
 
             if (value) {
-              value = '{ $num ->' + value + '\n  }';
+              var entity_ast = fluentParser.parseEntry(entity.original);
+              value = '{ ' + entity_ast.value.elements[0].expression.id.name + ' ->' + value + '\n  }';
             }
           }
 


flodolo:
> Could you confirm that we also need to re-export all strings that are broken because of this bug?

I'm not sure that's necessary, but I also don't have the full picture of what Pontoon will do:
- Notes changed all the strings to use $num, maybe we need to force an export from Pontoon to make sure everything looks good?
- Screenshots is broken and we need to force an export, so that $number is used instead of $num

pontoon/base/static/js/fluent_interface.js
```diff
@@ -316,7 +316,8 @@ var Pontoon = (function (my) {
             }
 
             if (value) {
-              value = '{ $num ->' + value + '\n  }';
+              var entity_ast = fluentParser.parseEntry(entity.original);
+              value = '{ ' + entity_ast.value.elements[0].expression.id.name + ' ->' + value + '\n  }';
```
Pike: This only works if the select expression is the first thing. Already the examples on playground break this.
Doing a full-depth analysis might be overdoing it, but I'd at least search the top-level:

```
entity_ast.value.elements.filter(node => node.type === 'Placable' && node.expression.type === ''SelectExpression')[0].expression.id.name
````

should work, AFAICT.

stasm:
Is this issue only about the name of the variable? What if English doesn't have any plurals?

Pike:
> Is this issue only about the name of the variable? What if English doesn't have any plurals?

Good question. In the case where we want to create a plural form in just l10n, we'll need to synthesize a variable. We could do that on any external argument in the en-US AST, for just-l10n-plural and multiple arguments, we'd need a UI element to help us.

Pike:
If en-US has select expressions, though, we can take that as a hint.

@stasm, do we have any AST iteration utilities in fluent.js yet?

stasm:
No, not yet. `python-fluent` has [`traverse`](https://github.com/projectfluent/python-fluent/blob/6610a40e5126f6f342914cec12eca287b761c596/fluent/syntax/ast.py#L49), but `fluent.js` doesn't have anything like that yet. I'd be happy to take a PR :)

pontoon/base/static/js/fluent_interface.js
mathjazz: Correct. Are you saying we have examples exposed for localization that break with this? If not, I'd rather ship this and address the issue of "non-first" select expressions in a followup patch.

The reason is https://github.com/mozilla/pontoon/commit/874cb0198775661b3194e3cca1330ddce164d3a9, which I'd like to revert and then add fixes on top of it before making things hard to merge. We need to make changes you're proposing in multiple places, because plurals support in FTL editor is for now very limited - to strings that have plural variants in source string (and nothing else).

mathjazz:
> Is this issue only about the name of the variable? What if English doesn't have any plurals?

Correct. If there's no plural in English, users cannot use plurals for now. Bug [1394110](https://bugzilla.mozilla.org/show_bug.cgi?id=1394110).

pontoon/base/static/js/fluent_interface.js
Pike: @flodolo, anything on your radar where en-US uses select expressions as a part?

Re the relanding of the source editor, I can see that this code snippet is right in the middle of the patch ;-)

flodolo: If I understand the question correctly, I think we're good. I can only think of selectors used for plural forms. BTW, these are all the files we have around:
https://github.com/mozilla/send/blob/master/public/locales/en-US/send.ftl
https://github.com/mozilla-services/screenshots/blob/master/locales/en-US/server.ftl
https://github.com/mozilla/testpilot/blob/master/locales/en-US/app.ftl
https://github.com/mozilla/testpilot/blob/master/locales/en-US/experiments.ftl

Only the first two have plural forms.

Pike: I've checked the first two, and they're of the form

    foo = { $num ->
        }

so we should get along with this. We'll need a follow-up bug, though.

mathjazz:
Thanks.

Filed:
https://bugzilla.mozilla.org/show_bug.cgi?id=1424381

PR #777, opened 2017-12-13
Source view 2nd attempt

mathjazz:
In this PR I'll fix bugs that forced us to revert the source view patches.

The 1st commit (f3380a4) makes reverting the reverts possible and the 4th (b263770) applies what the 1st commit reverted.

So only the commits following the 4th one are in need of a review.

### DIFF
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -5,6 +5,24 @@ var Pontoon = (function (my) {
   var fluentParser = new FluentSyntax.FluentParser({ withSpans: false });
   var fluentSerializer = new FluentSyntax.FluentSerializer();
 
+  function renderPluralEditor(value, name, pluralIndex) {
+    var example = Pontoon.locale.examples[pluralIndex];
+
+    $('#ftl-area .main-value ul')
+      .append(
+        '<li class="clearfix">' +
+          '<label class="id built-in" for="ftl-id-' + name + '">' +
+            '<span>' + name + '</span>' +
+            (Pontoon.fluent.isCLDRplural(name) && example !== undefined ? '<span> (e.g. ' +
+              '<span class="stress">' + example + '</span>' +
+            ')</span>' : '') +
+            '<sub class="fa fa-remove remove" title="Remove"></sub>' +
+          '</label>' +
+          Pontoon.fluent.getTextareaElement(name, value) +
+        '</li>'
+      );
+  }
+
   return $.extend(true, my, {
     fluent: {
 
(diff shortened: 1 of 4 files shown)


mathjazz:
Thanks!

I wonder how should I go about all these comits? Should I just squash them into 1 or 2, e. g. "Fix bug 1392255" and "Fix bug 1382536"?

pontoon/base/static/js/fluent_interface.js
```diff
@@ -57,43 +57,55 @@ var Pontoon = (function (my) {
         }
         // Plurals
         else if (self.isPlural(entityAST)) {
+          function renderPluralEditor(value, name, pluralIndex) {
```
adngdb: Functions inside functions are generally a bad pattern. Can you move that to a different location? Ideally, outside of the `Pontoon` object, it already has too much going on. :)

pontoon/base/static/js/fluent_interface.js
```diff
@@ -57,43 +57,55 @@ var Pontoon = (function (my) {
...
+            var variants = translationAST.value.elements[0].variants;
 
-            if (translationAST) {
-              for (var i = 0; i < variants.length; i++) {
-                if (variants[i].key.name === pluralStr) {
-                  value = self.serializePlaceables(variants[i].value.elements);
-                  break;
-                }
-              }
+            for (var i = 0; i < variants.length; i++) {
```
adngdb: This should be a `forEach` instead. The reason why it wasn't before is because there was a `break` in there, but you've removed it.

mathjazz:
> D'oh! It's cheating to add commits after an r+... ;)

I warned you there's another commit coming on IRC. :)

Let me address this quickly...

pontoon/base/static/js/fluent_interface.js
mathjazz: `Pontoon.fluent` is actually the only meaningful option I can think of. Do you have other ideas?

adngdb: You don't need that function outside of the file, so it can simply be at the top of the file, in the closure.

mathjazz:
All comment have been addressed. I've rebased into a single commit following the revert.

This is now deployed on stage, closely resembling current prod DB. Those are the files that use FTL:
https://mozilla-pontoon-staging.herokuapp.com/sl/firefox-screenshots/server.ftl/?string=168513
https://mozilla-pontoon-staging.herokuapp.com/sl/test-pilot-firefox-send/send.ftl/?string=169011
https://mozilla-pontoon-staging.herokuapp.com/sl/test-pilot-website/app.ftl/?string=153342
https://mozilla-pontoon-staging.herokuapp.com/sl/test-pilot-website/experiments.ftl/?string=159461
https://mozilla-pontoon-staging.herokuapp.com/sl/test-pilot-website/experiments.ftl/?string=159461

mathjazz:
@adngdb I've addressed the [comments](https://public.etherpad-mozilla.org/p/flod-ftl) I got during testing. Could you review the last two commits?

mathjazz:
Rebased.

PR #778, opened 2017-12-14
Bug 1417530 - Install test requirements in docker.

adngdb:
I did not succeed in running tests with pytest in `make dockertest` for the moment, for a reason I don't understand. For now, this allows to manually run them with `make dockershell`, which is all I need.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -9,8 +9,10 @@ RUN DEBIAN_FRONTEND=noninteractive /app/set_up_ubuntu.sh
 # Install Pontoon Python requirements
 COPY requirements.txt /app/requirements.txt
 COPY requirements-dev.txt /app/requirements-dev.txt
+COPY requirements-test.txt /app/requirements-test.txt
 RUN pip install -U 'pip>=8'
 RUN pip install --no-cache-dir --require-hashes -r requirements-dev.txt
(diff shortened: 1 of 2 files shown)


jotes:
@adngdb btw. what do you think about setting DEV settings for pontoon inside docker? I'd like to use shell_plus from time to time.

jotes:
@adngdb oh, interesting - I rebuild docker and everything is fine. Sorry for false alarm.

phlax:
> I did not succeed in running tests with pytest in make dockertest for the moment

the following diff should work

```
diff --git a/docker/run_tests_in_docker.sh b/docker/run_tests_in_docker.sh
index e3dedc30..daac9d77 100755
--- a/docker/run_tests_in_docker.sh
+++ b/docker/run_tests_in_docker.sh
@@ -63,5 +63,17 @@ else
            local/pontoon \
            python manage.py test $@
 
+    echo "Running py.tests..."
+    docker run \
+           --rm \
+           --volumes-from pontoon-tests \
+           --workdir /app \
+           --network pontoon_default \
+           --link "${DC} ps -q postgresql" \
+           --env-file ./docker/config/webapp.env \
+           -e LOCAL_USER_ID=$UID \
+           local/pontoon \
+           py.test -vv
+
     echo "Done!"
 fi
```

PR #779, opened 2017-12-14
 Fix bug 1400620 - Reject previous translations after batch edit.

adngdb:
Solving the original issue was easy enough, but while I was at it, I decided to finish something I started a while ago: moving batch actions logic into a separate app. This removes some 300 lines from the base views file, and refactors the batch code into a lot of smaller functions. It should make it easier to test, but I haven't gotten to the testing part much yet... Still, there were no tests before, so this is really just a straight improvement.


pontoon/batch/actions.py
```diff
@@ -0,0 +1,200 @@
...
+    translated_resources = list(translations.translated_resources(locale))
+    changed_entities = list(
+        Entity.objects.filter(translation__in=translations).distinct()
+    )
+
+    return count, translated_resources, changed_entities
+
+
+def batch_action(form, user, translations, locale):
+    """Empty batch action, does nothing, only used for documentation.
```
jotes: @adngdb what do you think about moving that part into function name I mean, to rename this function into:
`def empty_batch_action` or even `def default_action`  - that would be more explicit imho.

tests/batch/views.py
```diff
@@ -0,0 +1,66 @@
...
+
+def test_batch_edit_translations_no_user(client):
+    """If there are no logged in users, the view redirects to the login page.
+    """
+    response = client.post(reverse('pontoon.batch.edit.translations'))
+    assert response.status_code == 302
+
+
+@pytest.mark.django_db
+def test_batch_edit_translations_bad_request(client, member0, locale0):
```
jotes: @adngdb I would use `pytest.mark.parametrize` here because there's a lot of duplicated assertions - what do you think?

pontoon/batch/views.py
```diff
@@ -0,0 +1,174 @@
...
+    projects = Project.objects.filter(pk__in=projects_pk.distinct())
+    for project in projects:
+        if not request.user.can_translate(project=project, locale=locale):
+            return HttpResponseForbidden(
+                "Forbidden: You don't have permission for batch editing"
+            )
+
+    translation_pks = set()
+
+    # Find all impacted translations, including plural forms.
```
mathjazz: This only collects "active" translations. Impacted can also be non-active translations (in a sense that they get rejected for example).

pontoon/batch/actions.py
```diff
@@ -0,0 +1,200 @@
...
+        )
+        translations.update(
+            approved=False,
+            approved_user=None,
+            approved_date=None,
+            rejected=True,
+            rejected_user=user,
+            rejected_date=timezone.now(),
+            fuzzy=False,
+        )
```
mathjazz: Note that `translations` only contains "active" translations; I've added another comment where `translations` are collected. That means non-active suggestions stay unreviewed after search & replace.

I suggest we use this refactor to get more specific in variable names.

pontoon/batch/views.py
```diff
@@ -0,0 +1,174 @@
...
+    Translation,
+)
+from pontoon.base.utils import (
+    require_AJAX,
+)
+from pontoon.batch import forms
+from pontoon.batch.actions import ACTIONS_FN_MAP
+
+
+log = logging.getLogger('pontoon')
```
mathjazz: `log = logging.getLogger(__name__)`

pontoon/batch/actions.py
```diff
@@ -0,0 +1,200 @@
...
+    content of the `replace` parameter.
+
+    For documentation, refer to the `batch_action` function.
+
+    """
+    find = form.cleaned_data['find']
+    replace = form.cleaned_data['replace']
+
+    try:
+        translations, changed_translations = translations.find_and_replace(
```
mathjazz: Should we move `TranslationQuerySet.find_and_replace()` in this module?

adngdb: If it's only ever used by it, yes, by all means!

mathjazz: Yup, it is.

tests/batch/views.py
adngdb: I can't do that here sadly, it conflicts with using fixtures. I need to have the fixtures data in the parametrization decorator, which is not possible.

pontoon/batch/actions.py
mathjazz: As we figured out, this is actually a feature - we don't want people to reject translations they haven't reviewed yet.

adngdb:
@mathjazz comments addressed, ready for review again!

pontoon/batch/utils.py
```diff
@@ -0,0 +1,78 @@
...
+        translation.pk = None  # Create new translation
+        translation.string = translation.string.replace(find, replace)
+        translation.user = translation.approved_user = user
+        translation.date = translation.approved_date = now
+        translation.approved = True
+        translation.fuzzy = False
+        translations_to_create.append(translation)
+
+    # Unapprove old translations
+    translations.update(approved=False, approved_user=None, approved_date=None)
```
mathjazz: I believe this is now redundand because of https://github.com/mozilla/pontoon/pull/779/files#diff-c0c8d928e6b5c9e664c03f6b16316c48R137.

pontoon/batch/utils.py
```diff
@@ -0,0 +1,78 @@
...
+
+    # Create translations' clones and replace strings
+    now = timezone.now()
+    translations_to_create = []
+    for translation in translations:
+        translation.pk = None  # Create new translation
+        translation.string = translation.string.replace(find, replace)
+        translation.user = translation.approved_user = user
+        translation.date = translation.approved_date = now
+        translation.approved = True
```
mathjazz: We're missing a `translation.rejected = False` here.

We're replacing active translations, which could be rejected. So they will end up being approved and rejected, which is not allowed.

pontoon/batch/utils.py
adngdb: Right, good catch!

PR #781, opened 2017-12-15
Fix bug 1425420: Fix plural form example calculation

mathjazz:
Previous code always set examples for the first two plural forms to 1 and 2.

@jotes r?

### DIFF
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -498,11 +498,17 @@ var Pontoon = (function (my) {
      */
     generateLocalePluralExamples: function () {
       var self = this;
-      var examples = self.locale.examples = {0: 1, 1: 2};
       var nplurals = self.locale.nplurals;
-      var n = 0;
 
-      if (nplurals !== 2) {
+      if (nplurals === 2) {
+        self.locale.examples = {0: 1, 1: 2};
+
+      } else {
+        var examples = self.locale.examples = {};
+
+        // Example candidate: n is a variable used in the eval()
+        var n = 0;
+
         while (Object.keys(examples).length < nplurals) {
           var rule = eval(self.locale.plural_rule);
           if (!examples[rule]) {


pontoon/base/static/js/translate.js
```diff
@@ -498,11 +498,16 @@ var Pontoon = (function (my) {
...
-      var examples = self.locale.examples = {0: 1, 1: 2};
       var nplurals = self.locale.nplurals;
-      var n = 0;
 
-      if (nplurals !== 2) {
+      if (nplurals === 2) {
+        self.locale.examples = {0: 1, 1: 2};
+      }
+
+      else {
```
jotes: `} else {` ?

pontoon/base/static/js/translate.js
```diff
@@ -498,11 +498,16 @@ var Pontoon = (function (my) {
...
+        var examples = self.locale.examples = {};
+        var n = 0;
```
jotes: I would add a comment - it's hard to see that this variable is only required for `eval()` later.

pontoon/base/static/js/translate.js
```diff
@@ -498,11 +498,16 @@ var Pontoon = (function (my) {
...
       var nplurals = self.locale.nplurals;
-      var n = 0;
 
-      if (nplurals !== 2) {
+      if (nplurals === 2) {
+        self.locale.examples = {0: 1, 1: 2};
+      }
+
+      else {
+        var examples = self.locale.examples = {};
```
jotes: Is it a common pattern in javascript to create 2 references and mutate only one of them? (I'm asking not judging).

pontoon/base/static/js/translate.js
mathjazz: Right. I forgot this file hasn't been sanitized by Adrian yet. :)

pontoon/base/static/js/translate.js
mathjazz: Good point!

pontoon/base/static/js/translate.js
mathjazz: I just use `examples` as a shortcut. I hope that's ok.

PR #784, opened 2017-12-19
Fix bug 1393523: Search and replace text values only

mathjazz:
To make search and replace work for FTL strings, we can't just use
Translation.string.replace(keyword, replacement) like for other file
formats, becuse we store the entire source code of the FTL Message in
Translation.string.

Instead, we need to:

  1. Parse each string and get AST
  2. Recursively traverse AST and run find & replace on text values only
  3. Serialize updated AST.

### DIFF
--- a/pontoon/batch/utils.py
+++ b/pontoon/batch/utils.py
@@ -6,6 +6,16 @@ from pontoon.base.models import (
     Translation,
 )
 
+from fluent.syntax import (
+    ast,
+    FluentParser,
+    FluentSerializer,
+)
+
+
+parser = FluentParser()
+serializer = FluentSerializer()
+
 
 def get_translations_info(translations, locale):
     """Return data about a translations set.
@@ -30,6 +40,31 @@ def get_translations_info(translations, locale):
     return count, translated_resources, changed_entities
 
 
+def ftl_find_and_replace(string, find, replace):
+    """Replace text values in an FTL string.
+
+    :arg string string: a serialized FTL string
+    :arg string find: a string to search for, and replace, in translations
+    :arg string replace: what to replace the original string with
+
+    :returns: a serialized FTL string
+    """
+
+    def fun(value):
+        if type(value) == ast.TextElement:
+            value.value = value.value.replace(find, replace)
+        return value
+
+    # Parse FTL string and get AST
+    old_ast = parser.parse_entry(string)
+
+    # Recursively traverse the AST and perform find and replace on text values only
+    new_ast = old_ast.traverse(fun)
+
+    # Serialize updated AST
+    return serializer.serialize_entry(new_ast)
+
+
 def find_and_replace(translations, find, replace, user):
     """Replace text in a set of translation.
 
@@ -45,10 +80,11 @@ def find_and_replace(translations, find, replace, user):
     """
     translations = translations.filter(string__contains=find)
 
+    # No matches found
     if translations.count() == 0:
         return translations, []
 
-    # Empty translations produced by replace might not be always allowed
+    # Empty translations produced by replace are not allowed for all formats
     forbidden = (
         translations.filter(string=find)
         .exclude(entity__resource__format__in=Resource.ASYMMETRIC_FORMATS)
@@ -60,8 +96,14 @@ def find_and_replace(translations, find, replace, user):
     now = timezone.now()
     translations_to_create = []
     for translation in translations:
+        string = translation.string
         translation.pk = None  # Create new translation
-        translation.string = translation.string.replace(find, replace)
+
+        if translation.entity.resource.format == 'ftl':
+            translation.string = ftl_find_and_replace(string, find, replace)
+        else:
+            translation.string = string.replace(find, replace)
+
         translation.user = translation.approved_user = user
         translation.date = translation.approved_date = now
         translation.approved = True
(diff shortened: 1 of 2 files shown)


pontoon/batch/utils.py
```diff
@@ -30,6 +40,31 @@ def get_translations_info(translations, locale):
...
+    """Replace text values in an FTL string.
+
+    :arg string string: a serialized FTL string
+    :arg string find: a string to search for, and replace, in translations
+    :arg string replace: what to replace the original string with
+
+    :returns: a serialized FTL string
+    """
+
+    def fun(value):
```
stasm: Perhaps rename `fun` to something more descriptive, e.g. `replace_text_elements`?

pontoon/batch/utils.py
```diff
@@ -30,6 +40,31 @@ def get_translations_info(translations, locale):
...
+        return value
+
+    # Parse FTL string and get AST
+    old_ast = parser.parse_entry(string)
+
+    # Recursively traverse the AST and perform find and replace on text values only
+    new_ast = old_ast.traverse(fun)
+
+    # Serialize updated AST
+    return serializer.serialize_entry(new_ast)
```
stasm: This is just a nit, but the first and the last comment here kind of repeat exactly what's written in code, and the middle comment could be move to a docstring of the function passed to `traverse` :)

pontoon/batch/utils.py
```diff
@@ -60,19 +96,26 @@ def find_and_replace(translations, find, replace, user):
     now = timezone.now()
     translations_to_create = []
     for translation in translations:
+        string = translation.string
```
stasm: I'd add a comment here explaining why it's helpful to cache the old value of the string.

mathjazz:
Thanks for the quick review! I've addressed your comments and added a minor optimization in https://github.com/mozilla/pontoon/pull/784/commits/ab17493a7f61163ab3d5094e61355a55e8e959d9.

I've also found and fixed a bug in https://github.com/mozilla/pontoon/pull/784/commits/7c2696de89974ae857be70f41c34f44f534d26fc.  This is what happens:

If you select an FTL string for batch editing and then click on another string in the string list, the editor doesn't update - it shows the string you selected for batch editing.

That's because we don't hide the batch editor early enough to be able to detect whether we need to show the rich FTL editor in https://github.com/mathjazz/pontoon/blob/7c2696de89974ae857be70f41c34f44f534d26fc/pontoon/base/static/js/translate.js#L651.

Could you please review this, too? I know that it sounds super complicated, so I'd be happy to provide more details if you need them.

stasm:
I was able to reproduce the bug on the staging server. I don't understand the front-end code well enough to be able to review this fix though, sorry.

pontoon/batch/utils.py
```diff
@@ -6,6 +6,16 @@
...
 
+from fluent.syntax import (
+    ast,
+    FluentParser,
+    FluentSerializer,
+)
+
+
+parser = FluentParser()
+serializer = FluentSerializer()
```
adngdb: Why create these at the module level, and not in functions? Do they take so much time that it's worth optimizing it?

pontoon/batch/utils.py
```diff
@@ -72,7 +117,6 @@ def find_and_replace(translations, find, replace, user):
         translations_to_create.append(translation)
 
     # Create new translations
-    changed_translations = Translation.objects.bulk_create(
-        translations_to_create
-    )
+    changed_translations = Translation.objects.bulk_create(translations_to_create)
```
adngdb: Iirc we've decided that we would go to 80 chars per line during out code-style meeting, didn't we? In which case, this is longer than 80. :)

pontoon/batch/utils.py
```diff
@@ -30,6 +40,29 @@ def get_translations_info(translations, locale):
...
+    :arg string string: a serialized FTL string
+    :arg string find: a string to search for, and replace, in translations
+    :arg string replace: what to replace the original string with
+
+    :returns: a serialized FTL string
+
+    """
+
+    def replace_text_elements(node):
+        """Recursively traverse the AST and perform find and replace on text values only"""
```
adngdb: This is not recursive, it doesn't call itself. It does traverse a tree, it seems, but that's not done by this function. :)

pontoon/batch/utils.py
mathjazz: Ha! :)

pontoon/batch/utils.py
mathjazz: True :)

pontoon/batch/utils.py
mathjazz: Should I move them into `find_and_replace()`? They're used in `ftl_find_and_replace()`, but that's called for each string.

adngdb: Yeah they could be arguments to `ftl_find_and_replace()`.

stasm: There's little use for passing them as arguments unless you'd like to stub them. It's safe to create them globally: they only keep state for the config (e.g. `withSpans`) and otherwise the parsing done via `FluentParser.parse` is stateless.

mathjazz: But that means if you find & replace 1000 strings, you construct parser and serializer 1000 times each.

adngdb: Huh, yeah, well, whatever. It's a detail anyway, let's not sweat too much over it!

PR #785, opened 2017-12-20
Fix bug 1366259: beautify FTL strings in latest activity

mathjazz:
Same as #618, but server side. The idea is to show FTL translations as simple strings instead of source in the latest activity column tooltip on dashboards.

Not pretty, but still better IMHO.

Let me know if this makes sense and I can add some tests.

### DIFF
--- a/pontoon/base/templatetags/helpers.py
+++ b/pontoon/base/templatetags/helpers.py
@@ -274,11 +274,6 @@ def _serialize_elements(elements):
 @library.filter
 def as_simple_translation(source):
     """Transfrom complex FTL-based strings into single-value strings."""
-
-    # Empty string
-    if not source:
-        return source
-
     translation_ast = parser.parse_entry(source)
 
     # Non-FTL string or string with an error
--- a/pontoon/base/templatetags/helpers.py
+++ b/pontoon/base/templatetags/helpers.py
@@ -292,7 +292,7 @@ def as_simple_translation(source):
             return _serialize_elements(variants[0].value.elements)
 
         # Simple strings
-        except (AttributeError, IndexError) as e:
+        except (AttributeError, IndexError):
             return _serialize_elements(translation_ast.value.elements)
 
     else:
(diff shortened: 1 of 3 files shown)


pontoon/base/templatetags/helpers.py
```diff
@@ -250,3 +251,56 @@ def dict_html_attrs(dict_obj):
...
+def _serialize_elements(elements):
+    """Serialize text elements and references into a simple string."""
+    response = ''
+
+    for element in elements:
+        if type(element) == ast.TextElement:
+            response += element.value
+
+        elif type(element) == ast.ExternalArgument:
+            response += '{$' + element.id.name + '}'
```
zbraniecki: we usually try to put spaces around the placeable when we serialize: `{ $name }` rather than `{$name}` as in https://github.com/projectfluent/python-fluent/blob/master/fluent/syntax/serializer.py#L137

pontoon/base/templatetags/helpers.py
```diff
@@ -250,3 +251,56 @@ def dict_html_attrs(dict_obj):
...
+
+        elif type(element) == ast.MessageReference:
+            response += '{' + element.id.name + '}'
```
zbraniecki: Same here - `{ foo }` rather than `{foo}` :)

pontoon/base/tests/test_helpers.py
```diff
@@ -54,3 +60,49 @@ def test_unicode(self):
...
     def test_escape(self):
         assert_equal(str(nospam('<>\'"@&')), '&lt;&gt;&quot;&quot;&#64;&amp;')
+
+
+class AsSimpleTranslationTests(TestCase):
+    """Tests related to the as_simple_translation template filter."""
+
+    def test_empty_string(self):
+        source = ''
+        assert_equal(as_simple_translation(source), '')
```
adngdb: What if `source` is `None`?

mathjazz: The filter is used on `Translation.string`, which is never None.

Should I add protection anyways?

adngdb: Nope, I just wanted to know.

PR #786, opened 2017-12-21
Fix bug 1426676: Support for WebExtensions' JSON format

mathjazz:
We also store `placeholder` data, which includes examples. Might come in handy in the future.

This is not urgent and can wait until @phlax is back.


Pike:
Wondering, are there ways to test this code?

Pike:
> We also store placeholder data

... couldn't find that in the code? Maybe I overlooked.

I'd also like to see us maintaining VCS attribution in the serialized file. AFAICT, json.dumps can serialize in whichever order it wants, and indention would also not be maintained.

mathjazz:
Yup, I should add tests here:
https://github.com/mathjazz/pontoon/tree/e735592e7f0c9768336a7979506a253d0cd2c0ae/pontoon/sync/tests/formats

Looks like I forgot to push the commit which imports placeholder data. Fixing.

Any suggestions on how to keep the output as close to the existing localized file as possible?

pontoon/sync/formats/json_extensions.py
```diff
@@ -0,0 +1,138 @@
...
+                self.json_file = json.loads(content)
+        except IOError:
+            # If the file doesn't exist, but we have a source resource,
+            # we can keep going, we'll just not have any translations.
+            if source_resource:
+                return
+            else:
+                raise
+
+        for order, unit in enumerate(self.json_file.items()):
```
jotes: `for order, (key, data) in enumerate(`

pontoon/sync/formats/json_extensions.py
```diff
@@ -0,0 +1,138 @@
...
+        Load the source resource, modify it with changes made to this
+        Resource instance, and save it over the locale-specific
+        resource.
+        """
+        if not self.source_resource:
+            raise SyncError('Cannot save FTL resource {0}: No source resource given.'
+                            .format(self.path))
+
+        with codecs.open(self.source_resource.path, 'r', 'utf-8') as resource:
+            json_file = json.loads(resource.read())
```
jotes: `json.load(resource)` - `.load` is able to read contents from file-like objects.

pontoon/base/models.py
```diff
@@ -1665,15 +1665,16 @@ class Resource(models.Model):
...
     FORMAT_CHOICES = (
-        ('po', 'po'),
-        ('xliff', 'xliff'),
-        ('xlf', 'xliff'),
-        ('properties', 'properties'),
         ('dtd', 'dtd'),
+        ('ftl', 'ftl'),
         ('inc', 'inc'),
         ('ini', 'ini'),
+        ('json', 'json'),
```
jotes: `.json` is the general format for various types of data. I'm not sure how the layout of the future repositories will look like.
Maybe we should restrict our parsers only to  parse only files named `messages.json` and additionaly use http://json-schema.org/ to validate if the schema is correct?

phlax: ive been using json-schema and it rocks - seems like a good way to validate files are actually messages

pontoon/sync/formats/json_extensions.py
```diff
@@ -0,0 +1,138 @@
...
+
+    @property
+    def comments(self):
+        return [self.data['description']] if 'description' in self.data else []
+
+    @property
+    def fuzzy(self):
+        return False
+
+    @fuzzy.setter
```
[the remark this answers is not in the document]
phlax: mozilla `.lang` files also have a fuzzy logic i think - not to detract from the point about refactoring

pontoon/sync/formats/json_extensions.py
```diff
@@ -0,0 +1,138 @@
...
+                self.entities[key] = JSONEntity(
+                    entity.order,
+                    entity.key,
+                    copy.copy(entity.data),
+                )
+
+        try:
+            with codecs.open(path, 'r', 'utf-8') as resource:
+                content = resource.read()
+                self.json_file = json.loads(content)
```
phlax: unless there is some setting somewhere else this will use dictionaries for json objects - ie unordered

you can set it globally (which it might be, but i cant see it grepping), or in the call to `loads` - eg

```
json.loads(content, object_pairs_hook=OrderedDict)
```

phlax:
definitely could do with some tests - https://codecov.io/gh/mozilla/pontoon/pull/786/diff

pontoon/base/models.py
mathjazz: Thanks, I'll look into that.

pontoon/sync/formats/json_extensions.py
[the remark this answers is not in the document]
mathjazz: I'll look into refactoring the defaults. There's no fuzzy in .lang files.

pontoon/sync/formats/json_extensions.py
```diff
@@ -0,0 +1,138 @@
...
+        except IOError:
```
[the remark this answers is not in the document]
mathjazz: Turns out `json` raises `ValueError` and `simplejson` raised `JSONDecodeError`. Anyhow, I'll update.

mathjazz:
Given that this PR is now less of a priority than it was at the time of openning it, and it's not even clear when we'll start using it, I'd like to finalize it without impacting other file formats too much (e.g. fuzzy refactoring).

I've addressed most of the comments and added a frontend fix.

@phlax @jotes Let me know if this looks better now and then I'll add some tests.

pontoon/sync/formats/json_extensions.py
```diff
@@ -0,0 +1,187 @@
...
+            for key, entity in source_resource.entities.items():
+                self.entities[key] = JSONEntity(
+                    entity.order,
+                    entity.key,
+                    copy.copy(entity.data),
+                )
+
+        try:
+            with codecs.open(path, 'r', 'utf-8') as resource:
+                self.json_file = json.load(resource)
```
phlax: i think it needs object_pairs_hook here too

pontoon/sync/formats/json_extensions.py
```diff
@@ -0,0 +1,187 @@
...
+from pontoon.sync.exceptions import ParseError
+from pontoon.sync.formats.base import ParsedResource
+from pontoon.sync.vcs.models import VCSTranslation
+
+
+log = logging.getLogger(__name__)
+
+SCHEMA = {
+    "type": "object",
+    "patternProperties": {
```
phlax: im wondering if this would be simpler as just properties without the pattern match

pontoon/sync/formats/json_extensions.py
```diff
@@ -0,0 +1,187 @@
...
+                                "example": {
+                                    "type": "string"
+                                }
+                            },
+                            "required": ["content"]
+                        }
+                    }
+                }
+            },
+            "required": ["message", "description"]
```
phlax: this needs to be at the same level as the properties its constraining

pontoon/sync/formats/json_extensions.py
mathjazz: I'm not sure how to do that. The JSON file contains a [list of objects](https://github.com/mathjazz/pontoon-test/blob/master/en-US/example.json) where anything can be a key.

pontoon/sync/formats/json_extensions.py
mathjazz: The nesting level seems to be fine?
http://json-schema.org/examples.html

It also breaks if I move it one level up.

phlax: my bad - you're right

pontoon/sync/formats/json_extensions.py
phlax: just use `properties`

phlax: aplogies - above comments should have said `additionalProperties` rather than `properties`

pontoon/sync/formats/json_extensions.py
```diff
@@ -0,0 +1,179 @@
...
+        return sorted(self.entities.values(), key=lambda e: e.order)
+
+    def save(self, locale):
+        """
+        Load the source resource, modify it with changes made to this
+        Resource instance, and save it over the locale-specific
+        resource.
+        """
+        if not self.source_resource:
+            raise SyncError('Cannot save FTL resource {0}: No source resource given.'
```
Pike: Not FTL, but ... ?

pontoon/sync/formats/json_extensions.py
```diff
@@ -0,0 +1,179 @@
...
+                            .format(self.path))
+
+        with codecs.open(self.source_resource.path, 'r', 'utf-8') as resource:
+            json_file = json.load(resource, object_pairs_hook=OrderedDict)
+
+            try:
+                validate(json_file, SCHEMA)
+            except ValidationError as e:
+                raise ParseError(e)
```
Pike: Should we really refuse to sync if a translated json isn't valid? Or should we rather fix it up?

pontoon/sync/formats/json_extensions.py
```diff
@@ -0,0 +1,179 @@
...
+                del json_file[key]
+
+        # Create parent directory if it doesn't exist.
+        try:
+            os.makedirs(os.path.dirname(self.path))
+        except OSError:
+            pass  # Already exists, phew!
+
+        with codecs.open(self.path, 'w+', 'utf-8') as f:
+            log.debug('Saved file: %s', self.path)
```
Pike: `Saved` is the wrong tense here, as above.

mathjazz:
@Pike @jotes I addressed the review comments and added tests. I also added support for marking up `$PLACEHOLDER$` as placeables and displaying placeholder examples in the metadata section (under the source string) in the editor.

I'd appreciate if you could have another look.

pontoon/base/static/js/translate.js
```diff
@@ -589,10 +589,23 @@ var Pontoon = (function (my) {
...
+        // JSON
+        } else if (typeof(entity.source) === 'object') {
+          var examples = [];
+          $.each(Object.keys(entity.source), function() {
+            var example = entity.source[this].example;
+            if (example) {
+              examples.push('$' + this.toUpperCase() + '$: ' + example);
+            }
           });
+          self.appendMetaData('Placeholder Examples', examples.join(', '));
```
jotes: [nit/question] I'm not sure if `examples` will always contain escaped values (?). What do you think about `self.doNotRender` here?

pontoon/sync/formats/json_extensions.py
```diff
@@ -0,0 +1,179 @@
...
+                validate(self.json_file, SCHEMA)
+
+        except (IOError, ValueError, ValidationError) as err:
+            # If the file doesn't exist or cannot be decoded,
+            # but we have a source resource,
+            # we can keep going, we'll just not have any translations.
+            if source_resource:
+                return
+            else:
+                raise ParseError(err)
```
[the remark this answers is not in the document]
mathjazz: Mostly for consistency with other formats, so it's easier to catch the exception:
https://github.com/mozilla/pontoon/blob/master/pontoon/sync/vcs/models.py#L349

pontoon/sync/formats/json_extensions.py
```diff
@@ -0,0 +1,184 @@
...
+            if entity.strings:
+                json_file[key]['message'] = entity.strings[None]
+            else:
+                del json_file[key]
+
+        # Create parent directory if it doesn't exist.
+        try:
+            os.makedirs(os.path.dirname(self.path))
+        except OSError:
+            pass  # Already exists, phew!
```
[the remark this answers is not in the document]
mathjazz: We use the same block in `silme` and `FTL`, so I'll factor it out. Note that `os.path.isdir(self.path)` will always be false, because `self.path` is the path to a resource file.

pontoon/sync/formats/json_extensions.py
```diff
@@ -0,0 +1,184 @@
...
+
+        with codecs.open(self.path, 'w+', 'utf-8') as f:
+            log.debug('Saving file: %s', self.path)
+            f.write(json.dumps(json_file, indent=4))
```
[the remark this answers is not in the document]
mathjazz: Again, this is consistent with `FTL` and `silme` formats, in which we also use the source file as the basis, remove untranslated and update translated strings, and then dump the tree into a localized file.

That sometimes causes an unnecessarily large diff, but fixing that problem will need to wait for a more extensive refactoring.

mathjazz:
I addressed your comments, @jotes. Thanks for your feedback!

For testing, you can use this:
https://github.com/mathjazz/pontoon-test

PR #790, opened 2017-12-23
Fix bug 1426835: Do not render HTML in FTL strings

mathjazz:
@adngdb or @Pike or @jotes If anyone is still around to have a look at this, I'd appreciate it.

### DIFF
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -258,7 +258,13 @@ var Pontoon = (function (my) {
 
         elements.forEach(function (item) {
           if (item.type === 'TextElement') {
-            translatedValue += item.value;
+            // TODO: We shouldn't rely on markPlaceables in determining when to not render HTML.
+            if (markPlaceables) {
+              translatedValue += Pontoon.doNotRender(item.value);
+            }
+            else {
+              translatedValue += item.value;
+            }
           }
           else if (item.type === 'ExternalArgument') {
             if (markPlaceables) {
@@ -600,6 +606,11 @@ var Pontoon = (function (my) {
           if (this.isPlural(ast)) {
             var variants = ast.value.elements[0].variants;
             response = this.serializePlaceables(variants[0].value.elements);
+          }
+
+          // Update source string markup
+          if (object.hasOwnProperty('original')) {
+            response = Pontoon.doNotRender(response);
           }
         }
 


pontoon/base/static/js/fluent_interface.js
```diff
@@ -258,7 +258,13 @@ var Pontoon = (function (my) {
...
         elements.forEach(function (item) {
           if (item.type === 'TextElement') {
-            translatedValue += item.value;
+            // TODO: We shouldn't rely on markPlaceables in determining when to not render HTML.
+            if (markPlaceables) {
+              translatedValue += Pontoon.doNotRender(item.value);
+            }
+            else {
+              translatedValue += item.value;
+            }
```
mathjazz: This fixes the source string panel.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -601,6 +607,11 @@ var Pontoon = (function (my) {
             var variants = ast.value.elements[0].variants;
             response = this.serializePlaceables(variants[0].value.elements);
           }
+
+          // Update source string markup
+          if (object.hasOwnProperty('original')) {
+            response = Pontoon.doNotRender(response);
+          }
```
mathjazz: This fixes the string list.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -258,7 +258,13 @@ var Pontoon = (function (my) {
 
         elements.forEach(function (item) {
           if (item.type === 'TextElement') {
-            translatedValue += item.value;
+            // TODO: We shouldn't rely on markPlaceables in determining when to not render HTML.
+            if (markPlaceables) {
+              translatedValue += Pontoon.doNotRender(item.value);
```
jotes: @mathjazz is there a case when `value` shouldn't be escaped? If I understand correctly strings coming from db should be always escaped via `doNotRender()`

mathjazz: Yes, we shouldn't escape in editor (textarea).

pontoon/base/static/js/fluent_interface.js
```diff
@@ -258,7 +258,13 @@ var Pontoon = (function (my) {
...
+            }
+            else {
```
jotes: `} else {`

jotes:
@mathjazz Seems okay-ish for me. I'm always concerned when I see the code that concatenates html-strings in javascript :/ On my localhost this patch fixes the issue with html tags in ftl files.

pontoon/base/static/js/fluent_interface.js
mathjazz: I'm keeping the code style consistent with how @adngdb reformatted the file. I like his code style a lot, but I agree this particular case looks odd.

mathjazz:
Thanks for having a look, @jotes! I'll go ahead and merge since this is pretty nasty.

pontoon/base/static/js/fluent_interface.js
adngdb: Ah, this is another case where it's all about personal taste... I personally dislike `} else {`, I find it harder to identify blocks and logic with it. In general I'm a big fan of vertical space, and every attempt at removing some makes me sad. :)

PR #792, opened 2018-01-03
Fix bug 1427048: Fix tab shortcut regression

mathjazz:
@jotes r?

### DIFF
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -1853,13 +1853,13 @@ var Pontoon = (function (my) {
         // Tab: Select suggestions
         if (!$('.menu').is(':visible') && key === 9 && !e.ctrlKey) {
 
-          // Prevent in complex FTL mode
+          // Rich FTL editor: ignore tab key
           if (self.fluent.isFTLEditorEnabled() && self.fluent.isComplexFTL()) {
             return;
           }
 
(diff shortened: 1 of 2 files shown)


pontoon/base/static/js/translate.js
```diff
@@ -1853,13 +1853,13 @@ var Pontoon = (function (my) {
...
-          // Prevent in complex FTL mode
+          // Rich FTL editor: ignore tab key
           if (self.fluent.isFTLEditorEnabled() && self.fluent.isComplexFTL()) {
             return;
           }
 
-          // Special case in FTL source editor
-          if ($('#translation').is(':visible') || $('#ftl').is('.active')) {
+          // Source FTL editor: insert tab character
+          if ($('#ftl:visible').is('.active')) {
```
jotes: isn't that `self.fluent.isComplexFTL()`?

mathjazz: No, FTLEditor is the form-based AKA rich FTL editor as seen here:
https://pontoon.mozilla.org/sl/firefox/browser/browser/preferences/main.ftl/?string=174655

jotes:
Ok, got it.  [Nit] Can we return that 'state' of app by some method?

mathjazz:
Updated with a method.

PR #631, opened 2017-07-16
Bug 1335331: added ability to change email address

karabellyj:
@mathjazz r?

PR #793, opened 2018-01-08
Fix bug 1335331 - Enable users to change their email address

adngdb:
Taking over from #631. This is exactly the same content, plus some changes I've made on top, and rebased against latest master. 

Lots of credits go to @karabellyj! Thanks again for your work! 

@mathjazz, mind making sure I haven't done anything stupid?

### DIFF
--- a/pontoon/contributors/static/css/contributor.css
+++ b/pontoon/contributors/static/css/contributor.css
@@ -47,6 +47,7 @@ a.avatar:hover .desc ~ img {
 }
 #profile-form .field {
   margin-top: 20px;
+  text-align: left;
 }
 
 #profile-form .field input {
@@ -70,7 +71,7 @@ a.avatar:hover .desc ~ img {
   margin-top: 5px;
   color: red;
 }
-#profile-email:focus + span{
+#id_email:focus + span{
   display: block;
 }
 
--- a/pontoon/contributors/static/css/contributor.css
+++ b/pontoon/contributors/static/css/contributor.css
@@ -3,7 +3,7 @@
 }
 
 a.avatar {
-  display: inline-block;
+  display: block;
   position: relative;
 }
 
@@ -30,60 +30,37 @@ a.avatar:hover .desc ~ img {
   opacity: 0.3;
 }
 
-#username {
-  color: #EBEBEB;
-  font-size: 48px;
-  letter-spacing: -2px;
-  text-transform: none;
-}
-
-#username .uneditable {
-  letter-spacing: normal;
-  padding: 8px 0 20px;
-}
-
-#username .container {
-  display: inline-block;
+#profile-form {
+  display: block;
   position: relative;
+  text-align: left;
+  margin: 0 auto;
+  width: 620px;
+}
+#profile-form .field {
+  margin-top: 10px;
 }
 
-#username .container input {
-  background: transparent;
-  border-bottom: 1px dashed transparent;
+#profile-form .field input {
   color: #FFFFFF;
-  font-size: 48px;
-  font-weight: 300;
-  padding: 4px 50px;
-  text-align: center;
-  width: 100%;
+  background: #373C42;
+  border: 1px solid #4F5965;
+  border-radius: 5px;
+  float: none;
+  width: 290px;
+  padding: 4px;
 
   -moz-box-sizing: border-box;
   box-sizing: border-box;
 }
 
-#username .container input:hover,
-#username .container input:focus {
+#profile-form .field input:hover,
+#profile-form .field input:focus {
   border-color: #888888;
 }
 
-#username .container .submit {
-  bottom: -8px;
-  color: #888888;
-  opacity: 0;
-  padding: 20px 0;
-  position: absolute;
-  right: 0;
-  /* Needed or click doesn't trigger */
-  transition: all 0.1s ease-out;
-}
-
-#username .container input:focus + .submit {
-  cursor: pointer;
-  opacity: 1;
-}
-
-#username .container input:focus + .submit:hover {
-  color: #7BC876;
+#profile-form button {
+  margin-top: 10px;
 }
 
 .info {
--- a/pontoon/contributors/static/css/contributor.css
+++ b/pontoon/contributors/static/css/contributor.css
@@ -38,43 +38,6 @@ a.avatar:hover .desc ~ img {
   text-transform: none;
 }
 
-#profile-form {
-  display: block;
-  position: relative;
-  text-align: left;
-  margin: 0 auto 70px auto;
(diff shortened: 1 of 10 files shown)


pontoon/contributors/templates/contributors/settings.html
```diff
@@ -49,13 +44,31 @@ <h2 id="username">
...
+          {{ profile_form.first_name.label_tag(label_suffix='') }}
+          {{ profile_form.first_name }}
+          {{ profile_form.first_name.errors }}
+      </div>
+      <div class="field">
+          {{ profile_form.email.label_tag(label_suffix='') }}
+          {{ profile_form.email }}
+          {{ profile_form.email.errors }}
+          <span id="email-warning">
+            Caution: Changing your email address will cause a logout.<br / />
```
mathjazz: `<br />`

pontoon/contributors/static/css/contributor.css
```diff
@@ -33,57 +33,46 @@ a.avatar:hover .desc ~ img {
...
+#email-warning {
+  display: none;
+  margin-top: 5px;
+  color: red;
 }
-
-#username .container input:focus + .submit:hover {
-  color: #7BC876;
+#id_email:focus + span{
+  display: block;
```
mathjazz: `contributor.css` is shared between the /settings and /profile page. All the newly introduced CSS rules should be moved to the settings.css file, because they don't have any impact on the /profile page.

pontoon/contributors/static/css/contributor.css
```diff
@@ -33,57 +33,46 @@ a.avatar:hover .desc ~ img {
...
+}
+#profile-form .field {
+  margin-top: 20px;
+  text-align: left;
 }
 
-#username .container input {
-  background: transparent;
-  border-bottom: 1px dashed transparent;
+#profile-form .field input {
```
mathjazz: Separate CSS rules with a blank line.

pontoon/contributors/static/css/contributor.css
```diff
@@ -33,57 +33,46 @@ a.avatar:hover .desc ~ img {
...
+  margin-top: 10px;
 }
-
-#username .container input:focus + .submit {
-  cursor: pointer;
-  opacity: 1;
+#email-warning {
+  display: none;
+  margin-top: 5px;
+  color: red;
```
mathjazz: Let's use *Pontoon red* here - `#F36`.

And let's apply the same style (margin + color) to the `.errorlist` (e.g. "Email address must be unique.") and also set `list-style: none;`.

pontoon/contributors/views.py
```diff
@@ -148,14 +134,32 @@ def save_custom_homepage(request):
...
+        user = get_object_or_404(User, username=request.user.username)
+
+        if locales_form.is_valid() and profile_form.is_valid():
+            locales_form.save()
+            profile_form.save()
+
             messages.success(request, 'Settings saved.')
-            return redirect(request.POST.get('return_url', '/'))
+
+            if user.email != request.user.email:
```
mathjazz: Do we need this check? `UserProfileForm` should be invalid if the condition is true, so we never reach this code.

adngdb: If the user didn't change their email, then there's no reason to disconnect them. Unless I'm missing something, this line is needed for that reason, and it does get called. It's possible for the user to edit their name and not their email, in which case the form is valid but the email did not change.

mathjazz: You're correct. Sorry for the noise.

adngdb:
@mathjazz I've added a new commit that does more than just addressing your comments, mind giving it a look? Rational for the change is in the commit message.

pontoon/contributors/static/css/settings.css
```diff
@@ -52,3 +52,45 @@
...
+}
+
+#profile-form button {
+  margin-top: 10px;
+}
+
+#email-warning {
+  display: none;
+}
+#email-warning,
```
mathjazz: Blank line missing!

pontoon/contributors/static/js/settings.js
```diff
@@ -23,4 +23,19 @@ $(function() {
...
+
+  emailField.on('keyup', function () {
+    var newEmail = emailField.val();
+    if (newEmail === originalEmail && warningContent.is(':visible')) {
+      warningContent.hide();
+    }
+    else if (newEmail !== originalEmail && !warningContent.is(':visible')) {
+      warningContent.show();
+    }
+  });
```
mathjazz: I'm not a big fan of this. It doesn't work if you paste the email address using the mouse and the context menu for example. It also adds a lot of code.

Unrelatedly, there's a usability issue with this text - it gets overflown by the autocomplete box. We could move it to the right of the field, and possibly make it fit into two lines.

screen shot 2018-01-15 at 18 28 18

adngdb: I agree about the usability issue, and I'm going to fix it, but I don't understand your point about "a lot of code". That's only 10 lines, and I don't think it can be done with much less. The CSS-only solution had a different bunch of usability issues. Do you think we should do without that feature entirely instead?

mathjazz: A lot of code compared to the CSS solution. What are the usability issues there?

adngdb: Quoting my commit message: 

> It was bad that the warning for a changed email would only appear on focus on the input. First it made it difficult to click the save button, as the layout was changed and thus the button moved. Second, it would show up even if the email was not changed.

I was going to consider going for a tooltip to the right of the input box instead of a red text below. What do you think?

mathjazz: You could circumvent the layout change by toggling the `visibility` property instead of `display`. But that would still make the UI "blink".

Tooltip sounds good. Do you want to do something fancy? Helper text would also work:
screen shot 2018-01-16 at 14 36 39

adngdb:
New proposal: 

screen-email-change-warning

pontoon/contributors/static/css/settings.css
```diff
@@ -83,10 +83,12 @@
   margin-top: 10px;
 }
 
-#email-warning {
-  display: none;
+#profile-form .help {
+  color: #FED271;
+  font-style: italic;
+  font-weight: bold;
```
mathjazz: Now that the text is always visible, I find the styling a little too intrusive.

I'd remove bold (bold+italic at the same time is rarely a good idea) and change color to `#888888`. Please also add `margin-top: 5px;`.

adngdb:
Success! Thanks a lot again to @karabellyj for your contributions! 😃

PR #797, opened 2018-01-15
Touch bug 1417530: Add more fixtures

phlax:
this will also ease the landing of #766

### DIFF
--- a/tests/fixtures/site.py
+++ b/tests/fixtures/site.py
@@ -1,41 +1,184 @@
+import math
+from datetime import datetime, timedelta
+
 import pytest
 
+from django.utils import timezone
 
-@pytest.fixture
-def site_matrix(factories):
+
+def _get_project_resources(project):
+    # this somewhat arbitrarily spreads the string count
+    # among projects
+    return [
+        dict(project=project, total_strings=(i + 5) ** 3)
+        for i in range(0, 7)]
+
+
+def _get_project_locales(locales, project):
+    return [
+        dict(project=project, locale=locale)
+        for locale in locales]
+
+
+def _get_translated_resources(locales, project, resources):
+    _kwargs = []
+    for i, resource in enumerate(resources):
+        for locale in locales:
+            # this again spreads the stats counts in a kinda
+            # arbitrary fashion
+            _kwargs.append(
+                dict(locale=locale,
+                     approved_strings=(i + 2) ** 3,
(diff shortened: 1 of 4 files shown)


tests/fixtures/site.py
```diff
@@ -1,41 +1,150 @@
...
 import pytest
 
+from django.utils import timezone
 
-@pytest.fixture
-def site_matrix(factories):
+
+def _add_project_resources(project):
+    return [
+        dict(project=project, total_strings=(i + 5) ** 3)
```
adngdb: All those numbers in this file look like magic numbers to me. Can you add a comment explaining what they correspond to? Like, why `+ 5` and why `** 3`?

tests/fixtures/site.py
```diff
@@ -1,41 +1,150 @@
...
+def _add_translated_resources(locales, project, resources):
+    _kwargs = []
+    for i, resource in enumerate(resources):
+        for locale in locales:
+            _kwargs.append(
+                dict(locale=locale,
+                     approved_strings=(i + 2) ** 3,
+                     translated_strings=(i + 1) ** 3,
+                     fuzzy_strings=i ** 3))
+    return _kwargs
```
adngdb: These three functions are called `_add_something` but I don't see them adding anything. Shouldn't they be called `_get_something` instead, as they return the items you're interested in?

tests/fixtures/site.py
```diff
@@ -1,41 +1,150 @@
...
+    for kwargs in translated_resource_kwargs:
+        kwargs["latest_translation"] = _get_latest_translation(latest, kwargs)
+    translated_resources = (
+        factories['translated_resource'](
+            batch_kwargs=translated_resource_kwargs))
+    return translated_resources
+
+
+@pytest.fixture
+def site_matrix(factories):
```
adngdb: This file could do with a lot more docstrings on functions, especially this one and `_create_site_root`. I'm not exactly sure what you mean by "site matrix" and "site root", so a short explanation could be quite useful. :)

tests/fixtures/site.py
phlax: yep, renamed

tests/fixtures/site.py
phlax: added, hope its enough to make this code a bit more readable

PR #798, opened 2018-01-16
[WIP] Es6 webpack support

phlax:
adding es6/webpack support in environment, altho this PR doesnt use it


docker/run_webapp.sh
```diff
@@ -2,6 +2,12 @@
 
 # Prepares then runs the webapp.
 
+echo 'getting nvm/node/npm'
+curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.8/install.sh | bash
+export NVM_DIR="$HOME/.nvm"
+[ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh" # This loads nvm
+nvm install node
```
adngdb: This would be done every time we start a container, which I think we want to avoid. It should instead be done at build time. Probably in `set_up_ubuntu.sh` instead thus.

requirements.txt
```diff
@@ -315,3 +315,6 @@ caighdean==0.0.3 \
 # Required by caighdean
 nltk==3.2.5 \
     --hash=sha256:2661f9971d983db314bbebd51ba770811a362c6597fd0f303bb1d3beadcb4834
+
+django-webpack-loader==0.5.0 \
+    --hash=sha256:0a8536e36a30d719018cd4c5da6e9d2377771134e713c14e617bb484b4f0acce
```
adngdb: This should go up in the section for requirements with no dependencies. Also, please keep it alphabetically sorted.

pontoon/settings/base.py
```diff
@@ -219,10 +221,12 @@ def path(*args):
 ANONYMOUS_USER_ID = -1
 GUARDIAN_RAISE_403 = True
 
-PIPELINE_COMPILERS = (
+__PIPELINE_COMPILERS = (
     'pipeline.compilers.es6.ES6Compiler',
 )
 
+# PIPELINE_ENABLED = True
+
```
adngdb: Debug leftover?

pontoon/settings/base.py
```diff
@@ -219,10 +221,12 @@ def path(*args):
 ANONYMOUS_USER_ID = -1
 GUARDIAN_RAISE_403 = True
 
-PIPELINE_COMPILERS = (
+__PIPELINE_COMPILERS = (
```
adngdb: Is this a fix of a previous bug that went undetected, or are you kind of commenting that key out? It's weird that we have keys starting with underscores, is that a thing that some libraries use?

phlax: it was me commenting it out - i left it there because i wasnt sure - but now i think we can safely remove

phlax:
> This does not mean exiting JS bundles are compiled with webpack, right? With the current configuration, if I understand correctly, only files in static/placeholder/ are compiled, right?

yep, thats correct

PR #799, opened 2018-01-16
Touch #1416256 String tiers - just the model

phlax:
This adds the boilerplate for the tags app, and the model


pontoon/tags/models.py
```diff
@@ -0,0 +1,17 @@
+
+import logging
+
+from django.db import models
+
+from pontoon.base.models import Project, Resource
+
+
+log = logging.getLogger(__name__)
```
phlax: probs should remove this

pontoon/tags/models.py
```diff
@@ -0,0 +1,12 @@
...
+
+from pontoon.base.models import Project, Resource
+
+
+class Tag(models.Model):
+    slug = models.CharField(max_length=20)
+    name = models.CharField(max_length=30)
+    project = models.ForeignKey(Project, blank=True, null=True)
+    resources = models.ManyToManyField(Resource)
+    priority = models.IntegerField(blank=True, null=True)
```
mathjazz: We probably want to use choices here, e.g. like this:
https://github.com/mozilla/pontoon/blob/e65d7d59f301f499de144cc638bf2255c4236a23/pontoon/base/models.py#L1682

pontoon/tags/models.py
```diff
@@ -0,0 +1,15 @@
+
+from django.db import models
+
+from pontoon.base.models import Project, Resource
+
+
+class Tag(models.Model):
+    slug = models.CharField(max_length=20)
+    name = models.CharField(max_length=30)
+    project = models.ForeignKey(Project, blank=True, null=True)
```
mathjazz: What's the purpose of the project field?
To be used in `unique_together`?
Optimization?

phlax: to create associated tags in the project admin ui

phlax: well because @flodolo wanted tags to be per-project so they kinda need to be associated

phlax:
> Obviously we can make that change when we actually implement the logic.

yep

PR #802, opened 2018-01-17
Touch #1416256: String tiers - admin ui

phlax:
adds an admin ui for adding/removing tags and their resources

this is the final PR for the string tiers admin widget (yay!)


pontoon/administration/templates/admin_project.html
```diff
@@ -287,8 +287,10 @@ <h3>
 
 {% block extend_css %}
   {% stylesheet 'admin_project' %}
+  {{ render_bundle('tag_admin', 'css') }}
```
mathjazz: There's nothing in there, or is it?

pontoon/administration/static/css/admin_project.css
```diff
@@ -469,3 +469,7 @@ textarea.strings-source {
   width: 156px;
   margin-top: 0;
 }
+
+.tag.inline .tag-resources {
+    padding: 5px;
+}
```
mathjazz: I don't think we need that, but I'll followup with CSS fixes anyways, so you can leave it.

pontoon/static/js/widgets/errors.css
```diff
@@ -0,0 +1,12 @@
...
+    padding: .3em;
+    margin: 0;
+    list-style: none;
+    background: #ccc;
+}
+
+ul.errors li.error {
+    color: #F36;
+    font-size: 1.1em;
+}
```
mathjazz: What do I need to do in the UI to test this CSS?

phlax: its kinda hard to trigger in normal operation - one way is to open 2 admin windows showing same resources. In first window select all and link/unlink - then do the same in second window with same resources - that should throw an error

pontoon/administration/templates/admin_project.html
phlax: hmm, maybe not - i think i have another npm package installed locally that fills it - ill check...

adngdb: Where is `render_bundle` coming from?

pontoon/administration/templates/tag_inline_formset.html
```diff
@@ -31,6 +31,14 @@ <h4>
     </div>
   </div>
   {% endif %}
+
+  {% if form.slug.value() %}
+  <div class="js-tag-resources tag-resources"
+       data-api="{{ url('pontoon.admin.project.ajax.tag', **dict(project=project.slug, tag=form.slug.value())) }}"
```
adngdb: Is this different from 
```python
url('pontoon.admin.project.ajax.tag', project=project.slug, tag=form.slug.value())
```
?

pontoon/administration/templates/admin_project.html
phlax: here https://github.com/ezhome/django-webpack-loader

pontoon/static/js/tags/admin/index.js
```diff
@@ -0,0 +1,15 @@
...
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import TagResourcesButton from './button';
+
+
+document.addEventListener("DOMContentLoaded", () => {
+    document.querySelectorAll('.js-tag-resources').forEach(node => {
+        ReactDOM.render(
+            <TagResourcesButton
```
adngdb: This is kinda weird that this displays a button. It's also strange that the button has a data API (with `api` and all). I would expect there to be a father component here, that would handle the data, and then delegate showing a button and the interface or whatever. 

Not a blocker, but if you agree I'd love to see a follow-up to improve that.

phlax: not sure what the best pattern is - these widgets are exceptional i think in that they are not part of an "app" - just inject to page.

im happy to code/review an update tho - but would need to be a bit clear about how it would work if im to code it

pontoon/administration/templates/tag_inline_formset.html
phlax: no - ill update

pontoon/administration/templates/admin_project.html
phlax: ill remove this as its not necessary for this PR

pontoon/static/js/tags/admin/index.js
julen: If I understood the context properly (hard to tell without seeing something live), components need to be rendered in individual spread-out locations which have no common parent, because the surrounding layout is already rendered by server-side templates.

Unless the surrounding layout is also rendered via components, there will be no common parent, which means rendering via the top-level API (`ReactDOM.render()`) needs to be done N times, where N is the amount of `TagResourcesComponent` components to be rendered.

For passing the data to such components, reading it from the DOM as it's done here works, although IMO it's cleaner and more manageable to populate a data object in JS (i.e. in templates, initialize a JS object with relevant data), and have it read by glue code which is then responsible to render individual components. I have used this approach in the past and works good.

This pattern also allows to better handle situations where these components need to be re-rendered and/or added/removed due to interactions. Note as well that in those situations it's always important to unmount components to avoid keeping alive references to unused objects and event listeners.

phlax: > For passing the data to such components, reading it from the DOM as it's done here works, although IMO it's cleaner and more manageable to populate a data object in JS

for the other UIs i worked on with react i did this - or mostly - there are some tradeoffs

in this case i dont think this is a good pattern - it would force the transfer of lots of data (list of all resources * n) that for the most part wouldnt get used

julen: In terms of data transfer, what I suggested is no different from the current code — again, if I understood the context correctly.

phlax:
in the current code the widgets only get filled - and data used - if the user clicks the button - they arent rendered (or at least the data is not called/rendered) at page load

pontoon/static/js/tags/admin/index.js
phlax: in the current code the widgets only get filled - and data used - if the user clicks the button - they arent rendered (or at least the data is not called/rendered) at page load

pontoon/administration/static/css/admin_project.css
phlax: ill remove and reduce noize on PR - and you can dress it with a clean slate

pontoon/static/js/tags/admin/index.js
phlax: @julen maybe i misunderstand the point about passing it in DOM - in the current implementation the data is held as `state` in the dataManager hoc, and passed down as props. Are you suggesting storing it somewhere else after its retrieved from xhr ?

pontoon/static/js/widgets/errors.css
mathjazz: I get a weird error message if I do this:
screen shot 2018-03-21 at 15 45 12

pontoon/static/js/widgets/errors.css
```diff
@@ -0,0 +1,12 @@
...
+ul.errors {
+    padding: .3em;
+    margin: 0;
+    list-style: none;
+    background: #ccc;
+}
+
+ul.errors li.error {
+    color: #F36;
+    font-size: 1.1em;
```
mathjazz: I would remove that line.

phlax: cool, removing...

pontoon/static/js/widgets/errors.css
phlax: let me check that again

phlax: @mathjazz good spot - fixed

pontoon/static/js/tags/admin/index.js
julen: What I was suggesting is that instead of storing the data in the DOM as data attributes, it can be rendered to a JS object, and use that to initialize a controller/parent module which serves as glue code and is responsible of doing the rendering of components (mounting/unmounting as needed).

phlax: ah - i see - the api/project etc, rather than `data`

Pike: IIRC, inline js data objects don't go well with CSP, or rather, are frowned upon from that school of thought.

phlax: you can add `nonces` to the snippets - i think thats ok - see https://github.com/Bennyoak/django-csp-nonce

phlax: although...

> This code has not been through a third party security audit.

pontoon/static/js/widgets/errors.css
mathjazz: Thanks for the update! I'd still change the message from:
```
data: Unable to link the resources requested
```
to
```
Unable to link the requested resources
```

phlax: its in the context of it being a django form - so its the `data` field that is wrong - but we can make it so that it adds a form error, rather than a field error

mathjazz: As long as we drop the field name (and change word order), I'm fine. It doesn't really tell much to the end user in this case.

PR #803, opened 2018-01-17
Touch Bug 1243129: React utils and widgets

phlax:
Add some basic react widgets (...and tests!)

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,9 @@
       "<rootDir>/pontoon/static/js",
       "<rootDir>/static/placeholder"
     ],
+    "moduleNameMapper": {
+      "\\.(css|less)$": "identity-obj-proxy"
+    },
     "setupFiles": [
       "./tests/js/__setup__"
     ],
@@ -54,14 +57,20 @@
     "babel-preset-env": "^1.6.1",
     "babel-preset-react": "^6.24.1",
     "babel-preset-stage-2": "^6.24.1",
+    "css-loader": "^0.28.8",
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "^1.1.1",
     "eslint": "^4.15.0",
     "eslint-plugin-react": "^7.5.1",
+    "identity-obj-proxy": "^3.0.0",
     "jest": "^22.1.2",
+    "jsdom": "^11.6.0",
     "react": "^16.2.0",
     "react-dom": "^16.2.0",
+    "react-table": "^6.7.6",
     "react-test-renderer": "^16.2.0",
+    "request": "^2.83.0",
+    "style-loader": "^0.19.1",
     "sync-exec": "^0.6.2",
     "webpack": "^3.10.0",
     "webpack-bundle-tracker": "^0.2.1",
--- a/pontoon/static/js/widgets/columns.js
+++ b/pontoon/static/js/widgets/columns.js
@@ -1,0 +1,97 @@
+
+import React from 'react';
+
+
+export class Columns extends React.Component {
+
+    render() {
+        return (
+            <Container
+               columns={this.columns.length}
+               ratios={this.columns.map(([, v]) => v)}>
+              {this.columns.map(([column,], key) => {
+                  return (
+                      <Column key={key}>
+                        {column}
+                      </Column>);
+              })}
+            </Container>);
+    }
+}
+
+
+export class Container extends React.Component {
+
+    createColumnStyle(width) {
+        return {
+            float: "left",
+            overflow: "hidden",
+            boxSizing: "border-box",
+            width: width.toString() + "%"};
+    }
+
+    get containerStyle () {
+        return {
+            content: "",
+            display: "table",
+            width: "100%",
+        };
+    }
+
+    get columnStyles () {
+        let {columns, ratios} = this.props;
+        let result = [];
+        ratios = ratios || [];
+        if (!columns) {
+            return result;
+        }
+        if (ratios.length) {
+            // sum of rations
+            const total = ratios.reduce((a, b) => a + b, 0);
+            // percentage widths of each column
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -3,7 +3,9 @@
   "version": "0.1.0",
   "description": "Mozilla in-place localization tool.",
   "private": true,
-  "dependencies": {},
+  "dependencies": {
+    "node-pre-gyp": "^0.6.39"
```
phlax: not sure if all these explicit dependencies are actually required - npm grumbles about dep issues - but this doesnt seem to help

pontoon/static/js/utils.js
```diff
@@ -0,0 +1,149 @@
...
+    asGetParams (data) {
+        return this.appendParams(new URLSearchParameters(), data);
+    }
+
+    asMultipartForm (data) {
+        return this.appendParams(new FormData(), data);
+    }
+
+    fetch (url, data, params) {
+        const {method} = (params || {})
```
jotes: `const {method} = (params || {'method': 'get'})` because currently if you won't pass params, function will return null.
Also I'd raise  an exception if passed method is unsupported.

tests/js/utils.js
```diff
@@ -0,0 +1,237 @@
...
+test('PontoonDjangoAjax instance', () => {
+    expect(ajax instanceof DjangoAjax).toBe(true);
+    expect(ajax instanceof PontoonDjangoAjax).toBe(true);
+})
+
+
+test('DjangoAjax csrf', () => {
+    const _ajax = new DjangoAjax();
+    document.cookie = "csrf=tripe";
+    expect(_ajax.csrf).toBe('tripe');
```
jotes: I'd try to check if:
* what you will get if there's no csrf cookie.
* there are multiple cookie keys.

tests/js/utils.js
```diff
@@ -0,0 +1,237 @@
...
+})
+
+
+test('PontoonDjangoAjax csrf', () => {
+    const _ajax = new PontoonDjangoAjax();
+    document.querySelector = jest.fn(() => {return {value: 'rhubarb'}})
```
jotes: do arrow functions require return? I think that:
`jest.fn(() => ({value: 'rhubarb'}))`
could be shorter.

tests/js/widgets.js
```diff
@@ -0,0 +1,106 @@
...
+    class MockContainer extends Container {
+
+        get columnStyles () {
+            return styles();
+        }
+    }
+
+    const container = shallow(<MockContainer children={children} />);
+    expect(container.text()).toBe('FOO......BAR');
+    expect(styles.mock.calls).toEqual([[], [], [], []])
```
jotes: I know that We talked about the guidelines in code that if linters don't complain then it's okay.
But I find a little bit challenging to understand what should be and de facto is tested.
I feel that We could improve situation here by doing one of 2 things.:
1. Create more smaller test-cases with bigger and more verbose descriptions.
2. Add newlines and add comments per every section of given test ->

pontoon/static/js/utils.js
```diff
@@ -0,0 +1,149 @@
+
+export class Cookies {
```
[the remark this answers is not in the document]
phlax: well, i think its about finding a balance - i was trying to avoid installing deps, but also not re-inventing the wheel - probs in the case of cookies we could just use something - but i didnt see any nice pythonesque strip libs (at least not after a quick look)

pontoon/static/js/utils.js
phlax: cool, ill add

tests/js/utils.js
phlax: this is actually  a bit wrong - i think it should be `csrftoken`

i kinda added this for completeness, to make it clear what was different about our setup, and to have the code there if we switched to using django's csrf

tests/js/utils.js
phlax: how so ?

phlax: ah - i see the question properly - yep if you have `{}` around the value

phlax: the alternate being  ` jest.fn(() => {return {value: 'rhubarb'}})`

tests/js/widgets.js
phlax: im just switching on eslint for new js code if that helps re linting

- smaller test cases - in some cases ok - but its expensive to build/teardown tests whether its db/rollback or react rendering - so i try and get stuff done while they are built/rendered

- new lines - i can go for a bit of new lineage - but not so much - its a pita try and read more code at the same time

tests/js/utils.js
phlax: ...ill add some comments etc...

phlax:
@jotes i removed the cookies stuff, added (a few!) new lines and addressed other comments.  Ill add some more comments etc tomorrow

tests/js/utils.js
phlax: wow - i think i read this late at night - apologies - i see you suggested the shorter version - updating...

pontoon/static/js/widgets/columns.js
```diff
@@ -0,0 +1,97 @@
...
+
+export class Container extends React.Component {
+
+    createColumnStyle(width) {
+        return {
+            float: "left",
+            overflow: "hidden",
+            boxSizing: "border-box",
+            width: width.toString() + "%"};
+    }
```
adngdb: I am not a big fan of having CSS properties in JS. Since we're using webpack, it would be quite simple to have a CSS file live next to this, and import it here. Also, I'm wondering if the width problem you're solving with JS could be done with CSS instead?

phlax: this is a definite exception. This does one thing and one thing only - it gives columns percentage widths in a grid layout - its the whole point of this component

PR #804, opened 2018-01-17
Touch Bug 1243129: Js testing and linting

phlax:
Adds jest and enzyme testing to the test env, and eslint, altho its set to not fail for now

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,44 @@
+module.exports = {
+    "extends": [
+        "eslint:recommended",
+        "plugin:react/recommended"
+    ],
+    env: {
+        browser: true,
+    },
+    parserOptions: {
+        ecmaVersion: 6,
+        ecmaFeatures: {
+            jsx: true,
+        },
+        sourceType: 'module',
+    },
+    globals: {
+        gettext: false,
+        ngettext: false,
+        interpolate: false,
+        l: false,
+        expect: false,
+        test: false,
+        browser: false,
+        jest: false,
+        Promise: false,
+        require: false,
+        shortcut: false,
+        sorttable: false,
+        $: false,
+        Pontoon: false,
+        jQuery: false
+    },
+    plugins: [
+        'react',
+    ],
+    rules: {
+        'react/prefer-es6-class': 0,
+        'react/prefer-stateless-function': 1,
+        "react/prop-types": 0,
+        "react/jsx-key": 0,
+        "react/jsx-uses-react": 1,
+        'react/jsx-uses-vars': 1,
+    },
+};
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,9 +14,12 @@ before_script:
 script:
   - pylama pontoon
   - python manage.py collectstatic --noinput
-  - python manage.py test --with-coverage
   - ./node_modules/.bin/webpack
+  - npm test
+  - python manage.py test --with-coverage
   - py.test --cov-append --cov-report=term --cov=. -v
+  # This temporarily exits 0 while we cleanup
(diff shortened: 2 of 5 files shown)


mathjazz:
I'll pass this to @adngdb and @jotes.

.eslintrc.js
```diff
@@ -0,0 +1,44 @@
+module.exports = {
+    "extends": [
+	"eslint:recommended",
```
adngdb: The indentation in this file is a mess, it mixes 3-spaces, 4-spaces and tabs. :/

.eslintrc.js
```diff
@@ -0,0 +1,44 @@
...
+	test: false,
+	browser: false,
+	jest: false,
+	Promise: false,
+	require: false,
+	shortcut: false,
+	sorttable: false,
+        $: false,
+        Pontoon: false,
+        jQuery: false
```
adngdb: It's interesting to put this here. I would instinctively put them in each file instead, adding only the globals being used there. Maybe keep just a few exceptions here, like `Promise`. But I find there is value in having, at the top of each file, a list of the globals being used. It makes it easier to spot where something is coming from.

.travis.yml
```diff
@@ -14,9 +14,11 @@ before_script:
 script:
   - pylama pontoon
   - python manage.py collectstatic --noinput
-  - python manage.py test --with-coverage
   - ./node_modules/.bin/webpack
+  - npm test
+  - python manage.py test --with-coverage
   - py.test --cov-append --cov-report=term --cov=. -v
+  - (./node_modules/.bin/eslint .; exit 0)
```
adngdb: I like that, but can you add a comment to notice that it should only be temporary?

tests/js/foo.js
```diff
@@ -0,0 +1,4 @@
+
+test('Post truth', () => {
+    expect(true).toBe(true)
+})
```
adngdb: Damn, we'll have to have the talk about semicolons in JS... As you must have guessed, I'm in favor of semicolons everywhere. The reasons are numerous and many have blogged about excellent reasons why we should use them. Also, this is actually the very first time I'm seeing semicolon-less JS code in my career at Mozilla, and it was surprising. 

So, please, semicolons? :-)

.eslintrc.js
phlax: ill fix

.eslintrc.js
phlax: im not sure i agree - esp for such things like new language primitives

.travis.yml
phlax: for sure

tests/js/foo.js
phlax: k

.eslintrc.js
phlax: rereading, thats what you said re Promise

phlax:
> I wonder how hard it would be to add a make dockertest-js command or something? Or maybe make dockertest accept some argument to run only Python or JS tests?

this seems like a good time to mention makeyfile 🎉 ! https://github.com/phlax/makeyfile

adngdb:
Oh hey `makeyfile` looks pretty cool! That would indeed advantageously replace our current Makefile. What do you think @mathjazz?

tests/js/foo.js
phlax: we need to make sure we are linting for that

adngdb: Isn't that the case with eslint by default?

phlax: not with current settings - i cribbed a pretty relaxed moanset

PR #808, opened 2018-01-19
Touch #1416256 String tiers - add tag test fixtures

phlax:
This provides fixtures for #766 - mostly to separate the test logic from the implementation for easier review.

~~It (will) also provides a glob to regex function that is needed in above PR~~

### DIFF
--- a/tests/fixtures/tags.py
+++ b/tests/fixtures/tags.py
@@ -1,0 +1,351 @@
+
+import fnmatch
+import functools
+from collections import OrderedDict
+from datetime import datetime
+
+import pytest
+
+from django.db.models import F
+from django.utils import timezone
+
+from pontoon.base.models import TranslatedResource, Translation
+from pontoon.tags.models import Tag
+
+
+@pytest.fixture
+def tag0(resource0):
+    """Tag 0"""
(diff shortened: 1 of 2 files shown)


tests/fixtures/tags.py
```diff
@@ -0,0 +1,352 @@
...
+        "approved_strings",
+        "translated_strings",
+        "fuzzy_strings"]
+    totals = {}
+    resource_attrs = [
+        "resource",
+        "locale",
+        "latest_translation__date",
+        "resource__total_strings",
+        "resource__path"]
```
phlax: these 2 should probs be removed as they are annotated

tests/fixtures/tags.py
```diff
@@ -0,0 +1,352 @@
...
+
+@pytest.fixture(params=_tag_kwargs)
+def tag_test_kwargs(request, tag_matrix):
+    """This is a parametrized fixture that provides a range of possible
+    **kwargs for testing the TagsTool against tags in the `tag_matrix`
+    fixture.
+
+    If a parameter values for `path` is an `int` its mangled to the
+    `path` of the `resource` (in the site_matrix) with the corresponding
+    `pk`
```
phlax: this should be `index` in the site_matrix list

PR #814, opened 2018-01-22
Bug 1431436: [FTL] Support PLATFORM()

mathjazz:
To make review easier, I've split the patch for bug [1431436](https://bugzilla.mozilla.org/show_bug.cgi?id=1431436) into two pieces, which can also be landed separately. This is the first piece that contains:

1. Simple preview updates, which all Pontoon to display virtually any valid FTL Message in the string list, latest activity column and other places where simplified preview is used.

2. Original string presentation updates to support any FTL Message that contains any combination of elements of type `TextElement`, `ExternalArgument`, `MessageReference` and `SelectExpression` in their value or attribute. That includes strings with plurals, PLATFORM() and strings that contain both - value and attribute.

The second piece will bring the editor support.

A [pontoon-ftl](https://github.com/mozilla-l10n/pontoon-ftl/blob/master/en-US/demo.ftl) repository is available for testing and set up on stage:
https://mozilla-pontoon-staging.herokuapp.com/fr/test-fluent/demo.ftl/

**NOTE: Stage is not up to date with this branch ATM.**

### DIFF
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -51,16 +51,15 @@ var Pontoon = (function (my) {
     var content = '';
     var simpleElements = [];
 
-    elements.forEach(function (element, i, array) {
+    elements.forEach(function (element, i) {
       // Adjoining simple elements are concatenated and presented as a simple string
       if (Pontoon.fluent.isSimpleElement(element)) {
         simpleElements.push(element);
 
-        if (i === array.length - 1) {
+        if (i === elements.length - 1) {
           content += renderOriginalSimpleElement(simpleElements, title);
         }
       }
-
       // SelectExpression elements are presented as a list of variants
       else if (element.type === 'SelectExpression') {
         content += renderOriginalSimpleElement(simpleElements, title);
@@ -267,7 +266,7 @@ var Pontoon = (function (my) {
               'ExternalArgument',
               'MessageReference',
               'SelectExpression'
-            ].indexOf(item.type) >= 0;
+            ].indexOf(item.type) === 0;
           });
         }
 
@@ -291,7 +290,7 @@ var Pontoon = (function (my) {
           'TextElement',
           'ExternalArgument',
           'MessageReference'
-        ].indexOf(element.type) >= 0;
+        ].indexOf(element.type) === 0;
       },
 
 
@@ -356,8 +355,6 @@ var Pontoon = (function (my) {
        * Is SelectExpression?
        */
       isSelectExpression: function (ast) {
-        var self = this;
-
         if (
           ast &&
           ast.value &&
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -731,11 +731,6 @@ var Pontoon = (function (my) {
             tree = ast.attributes[0];
           }
 
-          // This should never happen! Return fallback.
-          else {
-            return fallback;
-          }
-
           // Simple string: use entire value
           var elements = tree.value.elements;
 
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -354,6 +354,7 @@ var Pontoon = (function (my) {
        * markPlaceables Should placeables be marked up?
        */
       serializePlaceables: function (elements, markPlaceables) {
+        var self = this;
         var translatedValue = '';
         var startMarker = '';
         var endMarker = '';
@@ -381,6 +382,13 @@ var Pontoon = (function (my) {
               endMarker = '</mark>';
             }
             translatedValue += startMarker + '{' + item.id.name + '}' + endMarker;
+          }
+          else if (item.variants) {
+            var variantElements = item.variants.filter(function (item) {
+              return item.default;
+            })[0].value.elements;
+
+            translatedValue += self.serializePlaceables(variantElements);
           }
         });
 
@@ -711,19 +719,7 @@ var Pontoon = (function (my) {
             tree = ast.attributes[0];
           }
 
-          var elements = tree.value.elements;
-
-          // String with variants: use default variant
-          try {
-            elements = elements[0].variants.filter(function (item) {
-              return item.default;
-            })[0].value.elements;
-          }
-
-          // Simple string: use entire value
-          catch (e) {}
-
-          response = this.serializePlaceables(elements);
+          response = this.serializePlaceables(tree.value.elements);
 
           // Update source string markup
           if (object.hasOwnProperty('original')) {
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -34,36 +34,35 @@ var Pontoon = (function (my) {
     '</li>';
   }
 
-  function renderOriginalSimpleElement(simpleElements, title) {
-    if (simpleElements.length) {
-      var response = renderOriginalVariant(title, simpleElements);
-      simpleElements = [];
-      return response;
-    }
-
-    return '';
-  }
-
   /*
    * Render original string elements.
+   *
+   * - Adjoining simple elements are concatenated and presented as a simple string
+   * - SelectExpression elements are presented as a list of variants
    */
   function renderOriginalElements(elements, title) {
     var content = '';
     var simpleElements = [];
 
     elements.forEach(function (element, i) {
-      // Adjoining simple elements are concatenated and presented as a simple string
-      if (Pontoon.fluent.isSimpleElement(element)) {
-        simpleElements.push(element);
+      var isSimpleElement = Pontoon.fluent.isSimpleElement(element);
+      var isLastElement = i === elements.length - 1;
 
-        if (i === elements.length - 1) {
-          content += renderOriginalSimpleElement(simpleElements, title);
+      // Collect simple elements
+      if (isSimpleElement) {
+        simpleElements.push(element);
+      }
+
+      // Render collected simple elements when non-simple or last element is met
+      if (!isSimpleElement || isLastElement) {
+        if (simpleElements.length) {
+          content += renderOriginalVariant(title, simpleElements);
+          simpleElements = [];
         }
       }
-      // SelectExpression elements are presented as a list of variants
-      else if (element.type === 'SelectExpression') {
-        content += renderOriginalSimpleElement(simpleElements, title);
 
+      // Render SelectExpression
+      if (element.type === 'SelectExpression') {
         element.variants.forEach(function (item) {
           content += renderOriginalVariant(item.key.value || item.key.name, item.value.elements);
         });
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -54,11 +54,9 @@ var Pontoon = (function (my) {
       }
 
       // Render collected simple elements when non-simple or last element is met
-      if (!isSimpleElement || isLastElement) {
-        if (simpleElements.length) {
-          content += renderOriginalVariant(title, simpleElements);
-          simpleElements = [];
-        }
+      if ((!isSimpleElement || isLastElement) && simpleElements.length) {
+        content += renderOriginalVariant(title, simpleElements);
+        simpleElements = [];
       }
 
       // Render SelectExpression
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -351,24 +351,6 @@ var Pontoon = (function (my) {
 
 
       /*
-       * Is SelectExpression?
-       */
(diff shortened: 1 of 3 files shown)


Pike:
I'm not too fond of adding more special cases for hopefully frequent cases.

Can't we factor the selector display into serializePlaceables?

Also, I think we should display the default variant, not the first.

pontoon/base/templatetags/helpers.py
```diff
@@ -280,23 +280,27 @@ def as_simple_translation(source):
...
-            return _serialize_elements(attributes[0].value.elements)
-
-        # All other strings: display unchanged
-        except (AttributeError, IndexError):
-            return source
+        return source
+
+    elements = tree.value.elements
+
+    # String with variants: use first variant
```
zbraniecki: should we use the default variant instead of a first variant here?

pontoon/base/static/js/fluent_interface.js
```diff
@@ -23,6 +23,57 @@ var Pontoon = (function (my) {
...
+      '<span class="value">' +
+        Pontoon.fluent.serializePlaceables(elements, true) +
+      '</span>' +
+    '</li>';
+  }
+
+  function renderOriginalSimpleElement(simpleElements, title) {
+    if (simpleElements.length) {
+      var response = renderOriginalVariant(title, simpleElements);
+      simpleElements = [];
```
adngdb: This has no effects. If your goal is to empty the content of the variable that is passed as `simpleElements`, then you'll need to do something else. As it is now, it simply makes the variable point to a different, empty array, without altering the original one. 

To clear the actual original array, you can do `simpleElements.length = 0;` or `simpleElements.splice(0, simpleElements.length);`. See https://stackoverflow.com/questions/1232040/how-do-i-empty-an-array-in-javascript

pontoon/base/static/js/fluent_interface.js
```diff
@@ -23,6 +23,57 @@ var Pontoon = (function (my) {
...
+  }
+
+  /*
+   * Render original string elements.
+   */
+  function renderOriginalElements(elements, title) {
+    var content = '';
+    var simpleElements = [];
+
+    elements.forEach(function (element, i, array) {
```
adngdb: `array` is a poor variable name as it is really close to `Array`. I would favor a different name here.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -23,6 +23,57 @@ var Pontoon = (function (my) {
...
+      // Adjoining simple elements are concatenated and presented as a simple string
+      if (Pontoon.fluent.isSimpleElement(element)) {
+        simpleElements.push(element);
+
+        if (i === array.length - 1) {
+          content += renderOriginalSimpleElement(simpleElements, title);
+        }
```
adngdb: This won't happen if the last element of `elements` isn't a simple element. Is that what you intended? If so, I would love to see a comment that explains this logic. :)

pontoon/base/static/js/fluent_interface.js
```diff
@@ -23,6 +23,57 @@ var Pontoon = (function (my) {
...
+      }
+
```
adngdb: No empty line here, otherwise it looks like the `if` is finished, which is not the case.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -198,25 +249,69 @@ var Pontoon = (function (my) {
...
+       */
+      isSupportedMessage: function (ast) {
+        function elementsSupported(elements) {
+          return elements.every(function(item) {
+            return [
+              'TextElement',
+              'ExternalArgument',
+              'MessageReference',
+              'SelectExpression'
+            ].indexOf(item.type) >= 0;
```
adngdb: Shouldn't that be `=== 0`? Otherwise if there's a `FooTextElement` in there, it will considered supported, but it might have unexpected side effects.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -198,25 +249,69 @@ var Pontoon = (function (my) {
...
+       * - TextElement
+       * - ExternalArgument
+       * - MessageReference
+       */
+      isSimpleElement: function (element) {
+        return [
+          'TextElement',
+          'ExternalArgument',
+          'MessageReference'
+        ].indexOf(element.type) >= 0;
```
adngdb: Same here, I think that should be `=== 0`.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -257,6 +352,26 @@ var Pontoon = (function (my) {
       },
 
 
+      /*
+       * Is SelectExpression?
+       */
+      isSelectExpression: function (ast) {
+        var self = this;
```
adngdb: Unused.

pontoon/base/templatetags/helpers.py
```diff
@@ -280,23 +280,27 @@ def as_simple_translation(source):
...
+    try:
+        print elements
```
adngdb: Debug leftover?

pontoon/base/templatetags/helpers.py
```diff
@@ -280,23 +280,27 @@ def as_simple_translation(source):
...
     else:
-        # Strings with attributes: display first attribute
-        try:
-            attributes = translation_ast.attributes
-            return _serialize_elements(attributes[0].value.elements)
-
-        # All other strings: display unchanged
-        except (AttributeError, IndexError):
-            return source
+        return source
```
adngdb: If this should never happen, maybe it would be worth raising an Exception instead? This way, if it does actually happen, we'll know.

pontoon/base/static/js/fluent_interface.js
mathjazz: Agreed. OTOH it makes the code less readable. I like to separate blocks with an empty line, which is why I did it in other places as well.

So why don't we do this?

```js
// If condition is true
if (true) {
  a = 0;

// If condition is false
} else {
  a = 1;
}
```

pontoon/base/static/js/fluent_interface.js
mathjazz: Good point. Related to your comment below, I think we should be better off without `renderOriginalSimpleElement()`.

pontoon/base/static/js/fluent_interface.js
mathjazz: Good point. Related to your comment above, I think we should be better off without `renderOriginalSimpleElement()`.

pontoon/base/static/js/fluent_interface.js
adngdb: It's really a matter of personal preference. I like to have a clearly marked end of block, with the `}` alone on a new line. That's how I'm used to reading code, and thus what is easiest on me. I've always found `} else {` quite confusing because it kinda hides the end of the block. 

But overall, this is nitpicking and I can live with those newlines and stuff. 

On your example...

```javascript
// Do something.
if () {
    a();

// Do something else.
} else {
    b();
}
```

```javascript
// Do something.
if () {
    a();
}
// Do something else.
else {
    b();
}
```

```javascript
// Do something.
if () {
    a();
}

// Do something else.
else {
    b();
}
```
The second version to me is better because blocks are better identified. The third version hints me that the `else` block isn't related to the previous `if`. I use newlines to separate logical blocks of code, and there those two blocks are related, so they should be kept together. 

Hope I make sense, if you disagree it's fine! :)

pontoon/base/static/js/fluent_interface.js
mathjazz: https://github.com/mozilla/pontoon/pull/814/commits/46c9f2c601fd96ff15c8699665a60dc7e4b5d44e :)

pontoon/base/static/js/fluent_interface.js
mathjazz: Same as above.

pontoon/base/static/js/fluent_interface.js
mathjazz: I see your point, but as you said - it's a matter of personal preference. :)

Since we decided not go beyond linters(:D) in code style reviews, I'll take the liberty of keeping this unchanged.

mathjazz:
Thanks for the comments!

All of them should be addressed now, so this is ready for another look.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -613,27 +696,37 @@ var Pontoon = (function (my) {
...
           }
-          // Attributes
-          else {
-            var attributes = ast.attributes;
-            if (attributes && attributes.length) {
-              response = this.serializePlaceables(attributes[0].value.elements);
-            }
+
+          // Attributes: use ast of the first attribute
+          else if (ast.attributes.length) {
```
Pike: I think checking `attributes.length` isn't required here. That's why we removed the least `else` block.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -613,27 +696,37 @@ var Pontoon = (function (my) {
...
+            tree = ast.attributes[0];
           }
-          // Plurals
-          if (this.isPlural(ast)) {
-            var variants = ast.value.elements[0].variants;
-            response = this.serializePlaceables(variants[0].value.elements);
+
+          // Simple string: use entire value
```
Pike: This comment is down below again, and it's right there. Not here.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -613,27 +696,37 @@ var Pontoon = (function (my) {
...
+          var elements = tree.value.elements;
+
+          // String with variants: use default variant
+          try {
+            elements = elements[0].variants.filter(function (item) {
```
Pike: Still concerned about the assumption that it's the top-level element that makes this string non-simple.

pontoon/base/templatetags/helpers.py
```diff
@@ -280,23 +280,24 @@ def as_simple_translation(source):
...
-            elements = translation_ast.value.elements
-            variants = elements[0].expression.variants
-            return _serialize_elements(variants[0].value.elements)
+        tree = translation_ast
 
-        # Simple strings
-        except (AttributeError, IndexError):
-            return _serialize_elements(translation_ast.value.elements)
+    # Attributes: use first attribute
+    elif translation_ast.attributes:
```
Pike: Same as above. You could assert here, though.

pontoon/base/templatetags/helpers.py
```diff
@@ -280,23 +280,24 @@ def as_simple_translation(source):
...
-
-        # All other strings: display unchanged
-        except (AttributeError, IndexError):
-            return source
+    elements = tree.value.elements
+
+    # String with variants: use default variant
+    try:
+        elements = filter(
+            lambda x: x.default, elements[0].expression.variants
```
Pike: Here we make the assumption on the top-level complexity, too.

mathjazz: We have to - AST's of the used fluent libraries aren't the same.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -23,6 +23,55 @@ var Pontoon = (function (my) {
...
+      if (!isSimpleElement || isLastElement) {
+        if (simpleElements.length) {
+          content += renderOriginalVariant(title, simpleElements);
+          simpleElements = [];
+        }
+      }
+
+      // Render SelectExpression
+      if (element.type === 'SelectExpression') {
+        element.variants.forEach(function (item) {
```
Pike: Forgot, not exactly sure about the context in which this is used, but do we want to find the default variant here instead of going over all?

mathjazz: We need all variants here - this function is used in the source string panel.

pontoon/base/static/js/fluent_interface.js
adngdb: Ahahaha, sorry! 😄

pontoon/base/static/js/fluent_interface.js
```diff
@@ -23,6 +23,55 @@ var Pontoon = (function (my) {
...
+      var isLastElement = i === elements.length - 1;
+
+      // Collect simple elements
+      if (isSimpleElement) {
+        simpleElements.push(element);
+      }
+
+      // Render collected simple elements when non-simple or last element is met
+      if (!isSimpleElement || isLastElement) {
+        if (simpleElements.length) {
```
adngdb: This can be simplified to:
```javascript
if ((!isSimpleElement || isLastElement) && simpleElements.length) {
```

mathjazz:
Thanks!

I've added one more thing as suggest by @Pike in https://github.com/mozilla/pontoon/pull/814#discussion_r163582627.

adngdb:
Code looks good. @Pike, does this answer your concerns?

Pike:
> @Pike, does this answer your concerns?

I think so

PR #818, opened 2018-01-24
Touch Bug 1243129: linting, react env and a widget

phlax:
- sets up linting in more of the code base
- adds a checkbox widget that can have its `indeterminate` property set

### DIFF
--- a/pontoon/static/js/widgets/checkbox.js
+++ b/pontoon/static/js/widgets/checkbox.js
@@ -1,0 +1,36 @@
+
+import React from 'react';
+
+
+export default class Checkbox extends React.Component {
+
+    /* A checkbox which you can set `indeterminate` on
+     *
+     */
+
+    constructor (props) {
+        super(props);
+        this.el = {};
+    }
+
+    get indeterminate () {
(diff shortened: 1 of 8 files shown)


pontoon/static/js/widgets/checkbox.js
```diff
@@ -0,0 +1,38 @@
...
+    constructor (props) {
+        super(props);
+        this.el = {};
+    }
+
+    get indeterminate () {
+        const {indeterminate} = this.props
+        if (indeterminate) {
+            return true;
+        }
```
adngdb: Why don't you just do `return this.props.indeterminate;`? If the goal is to not expose the prop, then maybe `return this.props.indeterminate ? true : false;`?

phlax: i think i wanted to leave it undefined if its not set in the props

phlax: which i think is what it would be without messing with it, i can check again

phlax: actually - i checked its `false` by default - ill update...

PR #822, opened 2018-01-25
Touch #1416256: Add tags_enabled flag for Project

phlax:
this allows tags to be enabled/disabled for projects


mathjazz:
Are those project tags or tags belonging to project resources?

I don't see how you plan to use that, so I wonder if reverse DB lookup would be sufficient?
https://github.com/phlax/pontoon/blob/1d9b354d6f2af84fdb8a3b6ab77cc7106dad569d/pontoon/tags/models.py#L11

phlax:
> I don't see how you plan to use that, so I wonder if reverse DB lookup would be sufficient?

i dont think so - its used to completely disable tags for a project, so that they dont show up in ui, and any configured tags are hidden - i think this was what @flodolo wanted

flodolo:
> i dont think so - its used to completely disable tags for a project, so that they dont show up in ui, and any configured tags are hidden - i think this was what @flodolo wanted

Confirmed. The vast majority of projects won't have tags (only Gecko, and some like mozilla.org will have them), no point in making the UI busy for those.

PR #824, opened 2018-01-25
Add jinja env setup

phlax:
Allows you to use the `url` template tag, to get reversed urls (as you would with django templates)

### DIFF
--- a/pontoon/settings/base.py
+++ b/pontoon/settings/base.py
@@ -177,6 +177,7 @@ TEMPLATES = [
         'NAME': 'jinja2',
         'APP_DIRS': True,
         'OPTIONS': {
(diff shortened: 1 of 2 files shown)


phlax:
@mathjazz not sure how you feel about this one - its a convenience to be able to reverse urls in templates, but i can defo work around

mathjazz:
What does that bring us in addition to e.g.:
https://github.com/mozilla/pontoon/blob/master/pontoon/teams/templates/teams/teams.html#L15?

phlax:
nothing 8/ - im wondering why it didnt work for me, and where the url function comes from

ill follow up and see if i can do without...

adngdb:
@phlax can we close this PR?

phlax:
yep, sorry for noise...

PR #828, opened 2018-01-28
Touch #1416256: String tiers utils 1

phlax:
adds 2 utils for tag data:

- stats tool - gets aggregates for tags or for tagged locales/projects/resources
- translations tool - finds the last translation data for tags or tagged locales/projects/resources
- adds index on translation date, locale for fast lookup of latest translations


phlax:
@mathjazz @adngdb this one is finally ready for review - the next PR #766 has further tools that build on these

mathjazz:
This is very nicely split into commits, which makes things much easier to review. Nice job!

Before I dig into more details, I'd like to sort out why are we adding another approach to handling stats and latest activity, inconsistent with the existing implementaiton in other models. `Project`, `Locale`, `ProjectLocale` and `TranslatedResource` all inherit from `AggregatedStats` model and store data denormalized.

So I wonder if this approach intends to replace the one used by the `AggregatedStats` descendants?

If yes, are we confident we will not be introducing any [performance regressions](https://bugzilla.mozilla.org/show_bug.cgi?id=1214392)?

If no, why should we keep two separate logics that handle virtually the same problem in two different ways?

phlax:
> I'd like to sort out why are we adding another approach to handling stats 

I looked at what was there and concluded firstly that there was no need to denormalize stats onto tags and secondly that the algos there didnt quite do what i required.

> So I wonder if this approach intends to replace the one used by the AggregatedStats descendants?

imho the stats table should have been kept separate from the other models rather than being inherited from them. This would allow for very fast lookup of stats from indexed integer tables.

that said i wasnt looking to replace the implementation there - certainly not as part of this PR anyhow

phlax:
...also...

i then did various timing experiments to figure out which tables were fasted to start the joins from when mining the data (at least with current setup/indeces)

i tried quite hard to combine these 2 functionalities - ie getting aggregated stats, and the latest translation data, but it borked the queries, and had unreliable outcomes, so i gave up and did it as 2 queries instead

mathjazz:
> I looked at what was there and concluded firstly that there was no need to denormalize stats onto tags and secondly that the algos there didnt quite do what i required.

Good to hear there's no need to denormalize. That said, it doesn't mean we shouldn't denormalize if that means we can reuse existing code instead of writing much more code.

I'm interested in what's missing in algorithms, because the dashboards I saw on the demo appeared to be presenting very similar if not the same data `AggregatedStats` descendants use.

phlax:
> That said, it doesn't mean we shouldn't denormalize if that means we can reuse existing code instead of writing much more code

denormalizing means not only writing more code, it also means swellling the databse and having to manage those fields whenever their normal relations are changed

phlax:
also how would we do it in this case - sure we could denorm the *totals* for a tag but mostly we want breakdowns

phlax:
> I'm interested in what's missing in algorithms, because the dashboards I saw on the demo appeared to be presenting very similar if not the same data AggregatedStats descendants use.

they are designed to present the same data - or at least the same format of data.

they are not however showing denormalized fields, they are calculating aggregated stats on the fly according to situation (only project/locale breakdowns so far)

phlax:
...and...

if we were to denorm totals for project/locale/project_locale etc (which i dont actually know where we would put such data) then we would still need the queries provided here to do the actual denormalizatoin

mathjazz:
> denormalizing means not only writing more code, it also means swellling the databse and having to manage those fields whenever their normal relations are changed

That's the thing - the code to do that is already in place. You'd only need to inherit `Tag` from `AggregatedStats` and update the fields in `TranslatedResource.calculate_stats()` as we do for other models.

And you'd need to add the `TagLocale(AggregatedStats)` model and update it the same way.

Am I missing something else?

phlax:
yes - not all resources are associated with a `Tag`

phlax:
well, i suppose that point doesnt matter - but why create more denormalized fields and a bunch of additional models (ProjectTag, ProjectLocaleTag) when we can add 1 model, and efficiently get the stats etc, without having to update all of these tables every time a translation happens.

mathjazz:
I did a quick comparison on stage between how fast the list of teams loads for the [entire project](https://mozilla-pontoon-staging.herokuapp.com/projects/firefox/) and for a [specific tag](https://mozilla-pontoon-staging.herokuapp.com/projects/firefox/tags/another-tag/).

```
/projects/firefox/ajax/................................600 ms
/projects/firefox/ajax/tags/another-tag/..............1150 ms
```

In absolute numbers, this is not terrible. Relatively speaking, it's twice as slow as it could be.

Also: There's another tag called `A tag!`, with less resources, for which the AJAX request is half shorter. Which almost certainly means the code doesn't scale too well and we'll face problems if we'll have bigger projects to deal with in the future.

As I pointed earlier though, my main concern is using two different code paths for the same functionality. In the near future, we'll be making quite significant changes to stats (just take [suggestions](https://bugzilla.mozilla.org/show_bug.cgi?id=1377969) and [errors](https://bugzilla.mozilla.org/show_bug.cgi?id=1237667) for example), and having two code paths means we'll need to make them and maintain them and test them in two different places.

I understand the drawbacks of denormalization regarding data redundancy and maintenance. But we have these mechanisms in place already, so why not build on top of them?

I also think that there are ways to improve the way we handle stats. But I don't think adding another code path is one of them.

pontoon/tags/utils/base.py
```diff
@@ -0,0 +1,193 @@
...
+        clone_kwargs = dict(
+            (k, getattr(self, k))
+            for k
+            in self.clone_kwargs)
+        clone_kwargs.update(kwargs)
+        return self.__class__(**clone_kwargs)
+
+
+class FilteredData(Clonable):
+    """Base Tool for constructing and coalescing aggregating querysets
```
mathjazz: Nit: Given that the comments says it's a tool, should the class name be `FilteredDataTool`, to resemble the naming schema used in descendants?

pontoon/tags/utils/base.py
```diff
@@ -0,0 +1,193 @@
...
+        return Resource.objects
+
+    @property
+    def tag_manager(self):
+        return Tag.objects
+
+    @property
+    def translation_manager(self):
+        """Django object manager for ``Translation``
+        """
```
mathjazz: It's pretty obvious what the code does, but I agree - the comment doesn't hurt. I still wonder why comment here and not for other properties as well?

What would be more interesting to read a comment about is why do we need managers to be available as `TagsDataTool` properites.

pontoon/tags/utils/stats.py
```diff
@@ -0,0 +1,57 @@
...
+    filter_methods = (
+        'tag', 'projects', 'locales', 'path')
+
+    # from the perspective of translated resources
+    _default_annotations = (
+        ('total_strings', Coalesce(
+            Sum('resource__total_strings'),
+            Value(0))),
+        ('fuzzy_strings', Coalesce(
+            Sum("fuzzy_strings"),
```
mathjazz: Single quotes, please. :) Also below.

pontoon/tags/utils/translations.py
```diff
@@ -0,0 +1,50 @@
+
+from django.db.models import Max, Q
+
+from .base import TagsTRTool
+
+
+class TagsTranslationsTool(TagsTRTool):
```
mathjazz: Maybe `TagsLatestTranslationsTool`?

pontoon/tags/utils/stats.py
```diff
@@ -0,0 +1,57 @@
...
+            Value(0))),
+        ('approved_strings', Coalesce(
+            Sum("approved_strings"),
+            Value(0))),
+        ('translated_strings', Coalesce(
+            Sum("translated_strings"),
+            Value(0))))
+
+    def get_data(self):
```
mathjazz: Can we get some comments in here explaining what's going on?

pontoon/tags/utils/translations.py
```diff
@@ -0,0 +1,50 @@
...
+    """For given filters this tool will find the latest ``Translations``
+    for a a ``Tag``. It uses TranslatedResources to find the translations
```
mathjazz: Nit: Double a.

pontoon/tags/utils/base.py
phlax: sounds good - ill update

pontoon/tags/utils/base.py
phlax: I think i started adding comments - and thought similar to what you wrote - that the comments werent helping. Im happy to add/remove as you feel.

re overriding the managers - it lets you customize them (theoretically) - im happy to add a comment to that effect - altho i think probs in the class comment or similar

pontoon/tags/utils/translations.py
phlax: yep - that probs reflects the purpose better

pontoon/tags/utils/stats.py
phlax: yep

PR #829, opened 2018-01-29
Touch #1416256: String tiers utils 2

phlax:
Adds:

- ~~a resources tool that can be used for linking unlinking resources from tags~~ moved to #877 
- tagged proxy wrappers for using values data in templates

~blocked by #828~


phlax:
closed in favour of #766

PR #830, opened 2018-01-30
[WIP] Fix bug 1432179 - Sort contributors by email

jotes:
Hey, @mathjazz 
I know there're missing tests, however I decided to push it to check if you agree with how this should look like.

TODO:

* [x] backend tests

### DIFF
--- a/pontoon/teams/views.py
+++ b/pontoon/teams/views.py
@@ -129,19 +129,26 @@ def ajax_permissions(request, locale):
             }
         )
 
-    managers = l.managers_group.user_set.all()
-    translators = l.translators_group.user_set.exclude(pk__in=managers).all()
-    all_users = User.objects.exclude(pk__in=managers).exclude(pk__in=translators).exclude(email='')
+    managers = l.managers_group.user_set.order_by('email')
+    translators = l.translators_group.user_set.exclude(pk__in=managers).order_by('email')
+    all_users = (
+        User.objects
+            .exclude(pk__in=managers | translators)
+            .exclude(email='')
+            .order_by('email')
+    )
 
-    contributors_qs = User.translators.with_translation_counts(None, Q(locale=l), None)
-    contributors = set(contributor.email for contributor in contributors_qs)
+    contributors_emails = set(
+        contributor.email
+        for contributor in User.translators.with_translation_counts(None, Q(locale=l), None)
+    )
 
     locale_projects = l.projects_permissions
 
     return render(request, 'teams/includes/permissions.html', {
         'locale': l,
         'all_users': all_users,
-        'contributors': contributors,
+        'contributors_emails': contributors_emails,
         'translators': translators,
         'managers': managers,
         'locale_projects': locale_projects,
--- a/tests/teams/views.py
+++ b/tests/teams/views.py
@@ -1,0 +1,111 @@
+import pytest
+
+from django.shortcuts import render
+from mock import patch
+
+
+@pytest.yield_fixture
+def sorted_locale0_translators(
+    locale0,
+    user_factory
+):
+    """
+    Assign (& yield) sorted translators to locale0
+    """
+    sorted_translators = sorted(
+        user_factory(batch=3),
+        key=lambda u: u.email
+    )
+
+    locale0.translators_group.user_set.add(*sorted_translators)
+    yield sorted_translators
+
+
+@pytest.yield_fixture
+def sorted_locale0_managers(
+    locale0,
+    user_factory
+):
+    """
+    Assign (& yield) sorted managers to locale0
+    """
+    sorted_managers = sorted(
+        user_factory(batch=3),
+        key=lambda u: u.email
+    )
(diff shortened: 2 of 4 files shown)


pontoon/teams/views.py
```diff
@@ -129,12 +129,17 @@ def ajax_permissions(request, locale):
...
+    all_users = (
+        User.objects
+            .exclude(pk__in=managers | translators)
+            .exclude(email='')
+            .order_by('email')
+    )
 
     contributors_qs = User.translators.with_translation_counts(None, Q(locale=l), None)
-    contributors = set(contributor.email for contributor in contributors_qs)
+    contributors = {contributor.email for contributor in contributors_qs}
```
mathjazz: Isn't that the same as before?

pontoon/teams/static/js/permissions.js
```diff
@@ -100,6 +100,12 @@ $(function() {
     var clone = $(this).remove();
 
     target.find('ul').prepend(clone.removeClass('hover'));
+
+    // After every change sort translators/managers by email
+    target.find('ul li').sort(function (a, b) {
+      return $(a).text().localeCompare($(b).text());
+    }).appendTo(target.find('ul'));
+
```
mathjazz: I'd rather remove this actually.

That way the newly added user is always on top of the list and stands out. Otherwise, if the list is long, it might get lost. It's also consistent with how locale selector behaves in the Admin.

pontoon/teams/views.py
phlax: also - every time i read this code i get confused because the var has `_qs` and has a `Q` in the args that generate it - but its not really a `_qs` - can we rename it ?

jotes: Actually, Personally I would remove this temporary variable and do a simple inline function call.

jotes: @mathjazz okay... I'm a syntax sugar addict :D I'll revert that.

pontoon/teams/static/js/permissions.js
jotes: Totally makes sense. Actually I even remember our first discussion about this behaviour.
My fault, will redo!

pontoon/teams/views.py
```diff
@@ -129,19 +129,26 @@ def ajax_permissions(request, locale):
...
+    all_users = (
+        User.objects
+            .exclude(pk__in=managers | translators)
+            .exclude(email='')
+            .order_by('email')
+    )
 
-    contributors_qs = User.translators.with_translation_counts(None, Q(locale=l), None)
-    contributors = set(contributor.email for contributor in contributors_qs)
+    contributors_emails = set(
```
jotes: @mathjazz I restored set, removed temporary *_qs variable and renamed contributors variable into (hopefully) more descriptive label.

jotes:
@adngdb could you review code & tests? thx in advance!

adngdb:
Deferring the review to @phlax, I've got too many reviews on my plate. :)

tests/teams/views.py
```diff
@@ -0,0 +1,111 @@
...
+    project_locale0.translators_group.user_set.add(*[u['id'] for u in translators])
+
+    yield sorted_translators
+
+
+@pytest.mark.django_db
+@patch('pontoon.teams.views.render', wraps=render)
+def test_ajax_permissions_locale_translators_managers_order(
+        render_mock,
+        admin_client,
```
phlax: i think its important to test with a non `admin_client` also to test how the view would look to anon/logged in

pontoon/teams/views.py
phlax: @mathjazz, @jotes if we are not using the `translation_counts` and we are not using the ordering supplied in `translation_counts` should we be using this (expensive) method at all?

Im wondering if we should add a manager method specifically to get the emails etc

mathjazz: Yeah, we should totally optimize that. Not as part of this PR, though:
https://bugzilla.mozilla.org/show_bug.cgi?id=1379149

Note that I just recently added `with_translation_counts` here, because the original (and more straightforward) version was even slower: https://github.com/mozilla/pontoon/commit/5a7f766a3cc39fd436539db56ce04396ff3de2df.

tests/teams/views.py
```diff
@@ -0,0 +1,172 @@
...
+
+    # Check project_locale0 id in the permissions form
+    assert locale_projects[0][0] == project_locale0.pk
+
+    # Check project_locale0 translators
+    assert locale_projects[0][4] == sorted_project_locale0_translators
+
+
+@pytest.mark.django_db
+def test_users_permissions_for_ajax_permissions_view(
```
jotes: @phlax Hey, could you check if that's enough for now? I could create a follow-up bug to add more sec-related tests.

phlax: looks good

jotes: @phlax can you r? the rest of the stuff? Thanks in advance :D

PR #831, opened 2018-01-31
Fix bug 1434535: Mark XML tags as placeables

mathjazz:
This is a next iteration of [bug 1382348](https://bugzilla.mozilla.org/show_bug.cgi?id=1382348).

Among other things, it will mark up DOM Overlays and make them easier to insert into the editor (using a mouse click).

### DIFF
--- a/pontoon/base/static/js/main.js
+++ b/pontoon/base/static/js/main.js
@@ -148,6 +148,45 @@ var Pontoon = (function (my) {
     },
 
     /*
+     * Markup XML Tags
+     *
+     * Find any XML Tags in a string and mark them up, while making sure
+     * the rest of the text in a string is displayed not rendered.
+     */
+    markXMLTags: function (string) {
+      var self = this;
+
+      var markedString = '';
+      var startMarker = '<mark class="placeable" title="XML Tag">';
+      var endMarker = '</mark>';
+
+      var re = /(<[^(><.)]+>)/gi;
+      var results;
+      var previousIndex = 0;
+
+      function doNotRenderSubstring(start, end) {
+        return self.doNotRender(string.substring(start, end));
+      }
+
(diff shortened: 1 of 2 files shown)


pontoon/base/static/js/main.js
```diff
@@ -147,6 +147,45 @@ var Pontoon = (function (my) {
       return number.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ",");
     },
 
+    /*
+     * Markup XML Tags
+     *
+     * Find any XML Tags in a string and mark them up, while making sure
+     * the rest of the text in a string is displayed not rendered.
+     */
+    markXMLTags: function (string) {
```
jotes: [nit] I wonder if there's a possibility to move this function to a separate file and create some unit-tests.

I guess that even wouldn't hurt even if that function would "global" but tested. I'm worried about every function/method which touches raw-HTML markup that later could be exploited in security vulnerability.

@phlax probably could help.

mathjazz: I very much agree this code (like any other) needs proper tests.

Still, I don't think we should invest into adding unit tests right now, because we're about to [rewrite the frontend code of the translate page using React](https://bugzilla.mozilla.org/show_bug.cgi?id=1243129). And the logic of marking up placeables specifically needs to be [refactored](https://bugzilla.mozilla.org/show_bug.cgi?id=1368602).

phlax: > I'm worried about every function/method which touches raw-HTML markup that later could be exploited in a security vulnerability.

try injecting some `><script>alert('BAD STRINGS')</script>` into where the HTML would go

> I don't think we should invest into adding unit tests right now, because we're about to

yep, agreed

phlax:
lgtm

PR #833, opened 2018-01-31
Fix bug 1416528: Modified color of links in dashboard header

VishalCR7:
I have changed the color to brightest possible shade of white i.e. #FFFFFF also the links gets underlined when someone hovers on the whole list(ref: https://pontoon.mozilla.org/projects/ ). Also added the same with right ones in one of the pages (ref: https://pontoon.mozilla.org/ach/activity-stream-new-tab/ )

Here is the screenshot and hovering effect as well 
image

image

peek 2018-02-01 13-47

### DIFF
--- a/pontoon/base/static/css/heading_info.css
+++ b/pontoon/base/static/css/heading_info.css
@@ -45,6 +45,7 @@ ul {
 }
 
 #heading .details .value {
+  color: #AAA;
   float: right;
 }
 
@@ -131,7 +132,7 @@ ul {
 }
 
 #heading .legend li span.value {
-  color: #EBEBEB;
+  color: #AAA;
   float: right;
 }
 
--- a/pontoon/base/static/css/heading_info.css
+++ b/pontoon/base/static/css/heading_info.css
@@ -57,10 +57,6 @@ ul {
   color: #7BC876;
 }
 
-#heading .details li:hover .value a {
-  color: #7BC876;
-}
-
 #heading .details .priority .value {
   margin-top: 6px;
 }
--- a/pontoon/base/static/css/heading_info.css
+++ b/pontoon/base/static/css/heading_info.css
@@ -49,10 +49,15 @@ ul {
 }
 
 #heading .details .value a {
-  color: #EBEBEB;
+  color: #FFFFFF;
 }
 
 #heading .details .value a:hover {
+  color: #7BC876;
+}
+
+#heading .details li:hover .value a {
+  text-decoration: underline;
   color: #7BC876;
 }
 
@@ -107,6 +112,15 @@ ul {
   position: relative;
   text-align: left;
   text-transform: uppercase;
+}
+
+#heading .legend li a {
+  color: #FFFFFF;
+}
+
+#heading .legend li:hover a {
+  text-decoration: underline;
+  color: #7BC876;
 }
 
 #heading .legend li.banner {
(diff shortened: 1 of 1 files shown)


VishalCR7:
@mathjazz can you please review this PR

pontoon/base/static/css/heading_info.css
```diff
@@ -49,13 +49,18 @@ ul {
...
-  color: #EBEBEB;
+  color: #FFFFFF;
 }
 
 #heading .details .value a:hover {
   color: #7BC876;
 }
 
+#heading .details li:hover .value a {
+  text-decoration: underline;
```
mathjazz: I don't like the underline effect here. Chaging color should be enough.

pontoon/base/static/css/heading_info.css
```diff
@@ -109,6 +114,15 @@ ul {
   text-transform: uppercase;
 }
 
+#heading .legend li a {
+  color: #FFFFFF;
+}
+
+#heading .legend li:hover a {
+  text-decoration: underline;
```
mathjazz: Same as above.

pontoon/base/static/css/heading_info.css
VishalCR7: @mathjazz done :)

mathjazz:
Thanks for the update! I still think this doesn't solve the bug.

As mentioned above, you should also make non-links `color: #aaa;`, specifically `#heading .legend li span.value` and `#heading .details li .value`.

VishalCR7:
@mathjazz done should I update the screenshots as well ?

mathjazz:
On the project-locale page like `http://localhost:8000/en-GB/pontoon-intro/` the legend numbers (`#heading .legend li a .value`) look grey, despite being links. They should be white and change to green on hover.

pontoon/base/static/css/heading_info.css
```diff
@@ -45,17 +45,22 @@ ul {
...
 }
 
 #heading .details .value a:hover {
   color: #7BC876;
 }
 
+#heading .details li:hover .value a {
+  color: #7BC876;
+}
+
```
mathjazz: I'd remove this, too. Because when you hover over "REPOSITORY" on http://localhost:8000/en-GB/pontoon-intro/, the URL changes to green, but you can't click (because "REPOSITORY" is not a link).

VishalCR7: @mathjazz  the repository seems to be link I even clicked it and it lead me here https://github.com/mozilla/pontoon-intro

VishalCR7: Yup I have also checked it on mozilla.pontoon.org and it indeed is a link

mathjazz: Sure it is. But the title "REPOSITORY" is not. And it turns green on hover. So you'd expect it to be link, too.

VishalCR7: only links are green may be some confusion repository is not turning green at all. I will update all the screenshots with next commit so that it solves all the confusions. :)

VishalCR7: @mathjazz hope this one gets everything right sorry for the blunders but a link or two always tend to miss out

VishalCR7: screenshots updated!

mathjazz: > only links are green may be some confusion repository is not turning green at all.

This patch makes all the links white by default and green on hover, and all non-links grey, consistently across the entire dashboard heading. Making some of the non-links green on hover breaks that consistency. As I already said, it's also misleading - because you expect such text to be a link after you hover.

Please remove this rule and we're good to go.

PR #834, opened 2018-01-31
Fix bug 1418540: Fix overlay image cutoff in all-resources.

VishalCR7:
I couldn't test it locally but this css addition tends to fix it using developer tools
Reference link : https://pontoon.mozilla.org/fr/fundraising/all-resources/?string=174187
Screenshots : 
Before:
image

After
image

### DIFF
--- a/pontoon/base/static/css/translate.css
+++ b/pontoon/base/static/css/translate.css
@@ -21,6 +21,7 @@ body {
   left: 0;
   right: 0;
   margin: 0 auto;
+  max-width: 90%;
 }
 
 mark.placeable {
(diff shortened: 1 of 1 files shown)


VishalCR7:
@mathjazz @adngdb can you please review this PR

VishalCR7:
@mathjazz  I have set max width in the same way If i can set max height that will prevent overflow and may prevent overshoot as well

mathjazz:
Let's try. :)

VishalCR7:
@mathjazz  i have set the max height and set the img to center of overlay div and also updated the PR description See if it looks good

PR #835, opened 2018-02-02
Fix bug 1389724: Fixed author select filter UI

VishalCR7:
Fixed removal of .selected class even though it is selected which
gets removed due to updation each time filter option is clicked.
peek 2018-02-02 18-28

### DIFF
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -976,6 +976,9 @@ var Pontoon = (function (my) {
       if(!$('#from').data('id')) $('#from').val(Highcharts.dateFormat('%d/%m/%Y %H:%M', from));
       if(!$('#to').data('id')) $('#to').val(Highcharts.dateFormat('%d/%m/%Y %H:%M', to));
 
+      if(!$('#from').data('id')||!$('#to').data('id')){
+        return;
+      }
       // Render range selector
       Highcharts.setOptions({
         global: {
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -970,15 +970,17 @@ var Pontoon = (function (my) {
         return;
       }
 
+      if ($('#from').val() || $('#to').val()){
+        return;
+      }
+
       // Set default input values and limits
       var from = counts[0][0],
           to = counts[counts.length - 1][0];
-      if(!$('#from').data('id')) $('#from').val(Highcharts.dateFormat('%d/%m/%Y %H:%M', from));
-      if(!$('#to').data('id')) $('#to').val(Highcharts.dateFormat('%d/%m/%Y %H:%M', to));
+          
+       $('#from').val(Highcharts.dateFormat('%d/%m/%Y %H:%M', from));
+       $('#to').val(Highcharts.dateFormat('%d/%m/%Y %H:%M', to));
 
-      if($('#from').data('id')||$('#to').data('id')){
-        return;
-      }
       // Render range selector
       Highcharts.setOptions({
         global: {
@@ -1027,7 +1029,6 @@ var Pontoon = (function (my) {
             setExtremes: function (e) {
               $('#from').val(Highcharts.dateFormat('%d/%m/%Y %H:%M', e.min));
               $('#to').val(Highcharts.dateFormat('%d/%m/%Y %H:%M', e.max));
-              $('#filter .time-range input').data("id", 1);
             }
           }
         },
@@ -1482,13 +1483,13 @@ var Pontoon = (function (my) {
 
       // Clear error styling on value change
       $('#filter .time-range input').on('input propertychange change', function() {
-        $('#filter .time-range input').data("id", 1);
         var from = self.local2chart($('#from').val()),
             to = self.local2chart($('#to').val());
 
         $('#range-picker').highcharts().xAxis[0].setExtremes(from, to);
         $(this).removeClass('error');
       });
+
       // Do not close the filter menu if clicked inside the menu
       $('#filter').on('click', '.menu', function (e) {
         e.stopPropagation();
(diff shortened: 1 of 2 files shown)


VishalCR7:
@mathjazz  please review.

pontoon/base/static/js/translate.js
```diff
@@ -2992,22 +2992,42 @@ var Pontoon = (function (my) {
       var self = this,
           $forAuthors = $('#filter').find('.for-authors').toggle(authors.length > 0);
 
+      var selectedAuthors = $('#filter .menu li.author.selected').map(function() {
+                                return $.trim($(this).data("type"));
+                            }).get();
+
```
mathjazz: For consistency reasons let's stick to the indentation we use in JS files:
```js
      var selectedAuthors = $('#filter .menu li.author.selected').map(function() {
        return $.trim($(this).data("type"));
      }).get();
```

pontoon/base/static/js/translate.js
```diff
@@ -2992,22 +2992,42 @@ var Pontoon = (function (my) {
...
+                '<img class="rounded" src="' + this.gravatar_url + '">' +
+              '</span>' +
+              '<figcaption>' +
+                '<p class="name">' + this.display_name + '</p>' +
+                '<p class="role">' + this.role + '</p>' +
+              '</figcaption>' +
+              '<span class="count">' + self.numberWithCommas(this.translation_count) + '</span>' +
+            '</figure>' +
+          '</li>');
+        }
```
mathjazz: This is an anti-pattern. You duplicated the entire block, so in case we need to make a change, we need to make it in two places. That's rarelly a good idea.

Instead, you can keep the original block (the entire thing marked in red) and replace the first line with:
```js
var selected = (selectedAuthors.includes(this.email)) ? ' selected' : '';
$forAuthors.after('<li class="author' + selected + '" data-type="' + this.email + '">' +
```

pontoon/base/static/js/translate.js
```diff
@@ -2992,22 +2992,42 @@ var Pontoon = (function (my) {
...
-              '<img class="rounded" src="' + this.gravatar_url + '">' +
-            '</span>' +
-            '<figcaption>' +
-              '<p class="name">' + this.display_name + '</p>' +
-              '<p class="role">' + this.role + '</p>' +
-            '</figcaption>' +
-            '<span class="count">' + self.numberWithCommas(this.translation_count) + '</span>' +
-          '</figure>' +
-        '</li>');
+        if (selectedAuthors.includes(this.email)) {
```
mathjazz: I had no idea we can use `Array.prototype.includes()` safely! Thanks!

VishalCR7:
@mathjazz  I have included the changes and I must say that the use of ternary operator and other imporvements were awesome Thanks a lot!

VishalCR7:
let me see it :P

VishalCR7:
@mathjazz  hope it works now :P

VishalCR7:
@mathjazz  this seems to work fine for me see here : 
peek 2018-02-05 23-16

mathjazz:
This seems like a correct behaviour. I'll look again and get back to you once I figure out why it doesn't work in my case.

mathjazz:
> This seems like a correct behaviour. I'll look again and get back to you once I figure out why it doesn't work in my case.

The patch works fine when you select time range using the date/time picker. It doesn't however, if you select range using the chart.

VishalCR7:
@mathjazz  I didn't get to test it but hope it works

mathjazz:
Now the chart doesn't even render.

To avoid blocking the working part of the patch, I suggest you remove the last two commits and only keep the ones that fix the bug for the authors (1st and 2nd). And then we can fix the bug for the time range selector in a subsequent PR.

VishalCR7:
@mathjazz  just this last time

pontoon/base/static/js/translate.js
```diff
@@ -973,10 +973,12 @@ var Pontoon = (function (my) {
...
           to = counts[counts.length - 1][0];
+      if(!$('#from').data('id')) $('#from').val(Highcharts.dateFormat('%d/%m/%Y %H:%M', from));
+      if(!$('#to').data('id')) $('#to').val(Highcharts.dateFormat('%d/%m/%Y %H:%M', to));
 
-      $('#from').val(Highcharts.dateFormat('%d/%m/%Y %H:%M', from));
-      $('#to').val(Highcharts.dateFormat('%d/%m/%Y %H:%M', to));
-
+      if($('#from').data('id')||$('#to').data('id')){
+        return;
+      }
```
mathjazz: I think not-reseting the time range inputs can be greatly simplified. You only need to wrap the code to set the `#from` and `#to` input fields into an `if` block. Something like this:

```js
      // Set default input values and limits if not set yet
      if (!$('#filter li.time-range').is('.selected')) {
        var from = counts[0][0],
            to = counts[counts.length - 1][0];

        $('#from').val(Highcharts.dateFormat('%d/%m/%Y %H:%M', from));
        $('#to').val(Highcharts.dateFormat('%d/%m/%Y %H:%M', to));
      }
```

And then you can get rid of all the traces of `data('id')` (also in the template file).

pontoon/base/static/js/translate.js
```diff
@@ -970,12 +970,16 @@ var Pontoon = (function (my) {
         return;
       }
 
+      if ($('#from').val() || $('#to').val()){
+        return;
+      }
+
```
mathjazz: We should still update the chart, so we cannot return.

VishalCR7: the #rangepicker chart ?

pontoon/base/static/js/translate.js
```diff
@@ -970,12 +970,14 @@ var Pontoon = (function (my) {
         return;
       }
 
-      // Set default input values and limits
-      var from = counts[0][0],
-          to = counts[counts.length - 1][0];
+      // Set default input values and limits if values not set
+      if (!$('#from').val()||!$('#to').val()){
```
mathjazz: Nit: Please add some whitespace, like this:
```js
if (!$('#from').val() || !$('#to').val()) {
```

PR #836, opened 2018-02-04
Fix bug 1389969 - Fix display of trailing/leading spaces and newlines…

jotes:
… in the diff view.

Major changes:
* markPlaceables() is applied only once in diff(), that allows to mark
  leading/trailing spaces correctly
* fixed issue with missing trailing/leading spaces for multiline strings
* translation diff displays newlines correctly instead of flat string

Minor changes:
* use a switch statement instead of three if blocks
* use constants provided by Diff.js instead of magic numbers

@mathjazz could you look at this? I'm trying to finish things assigned to me ;-)

### DIFF
--- a/pontoon/base/static/js/main.js
+++ b/pontoon/base/static/js/main.js
@@ -187,25 +187,34 @@ var Pontoon = (function (my) {
     },
 
     /*
+     * Get markup for a placeable instance.
+     */
+    getPlaceableMarkup: function (title, replacement) {
+        return '<mark class="placeable" title="' + title + '">' + replacement + '</mark>';
+    },
+
+    /*
+     * Mark single instance of a placeable in string
+     */
+    markPlaceable: function(string, regex, title, replacement) {
+      replacement = replacement || '$&';
+      return string.replace(regex, this.getPlaceableMarkup(title, replacement));
+    },
+
+    /*
      * Markup placeables
      */
-    markPlaceables: function (string) {
-      function getReplacement(title, replacement) {
-        return '<mark class="placeable" title="' + title + '">' + replacement + '</mark>';
-      }
+    markPlaceables: function (string, whiteSpaces) {
+      var self = this;
+      whiteSpaces = whiteSpaces !== false;
 
-      function markup(string, regex, title, replacement) {
-        replacement = replacement || '$&';
-        return string.replace(regex, getReplacement(title, replacement));
-      }
-
-      string = this.doNotRender(string);
+      string = self.doNotRender(string);
 
       /* Special spaces */
       // Pontoon.doNotRender() replaces \u00A0 with &nbsp;
-      string = markup(string, /&nbsp;/gi, 'Non-breaking space');
-      string = markup(string, /[\u202F]/gi, 'Narrow non-breaking space');
-      string = markup(string, /[\u2009]/gi, 'Thin space');
+      string = self.markPlaceable(string, /&nbsp;/gi, 'Non-breaking space');
+      string = self.markPlaceable(string, /[\u202F]/gi, 'Narrow non-breaking space');
+      string = self.markPlaceable(string, /[\u2009]/gi, 'Thin space');
 
       /* Multiple spaces */
       string = string.replace(/  +/gi, function(match) {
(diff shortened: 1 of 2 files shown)


pontoon/base/static/css/style.css
```diff
@@ -1408,6 +1408,10 @@ body > form,
   display: none;
 }
 
+#helpers > section ul li > p.translation-diff {
+  white-space: pre-wrap;
+}
+
```
mathjazz: This is now being fixed in https://github.com/mozilla/pontoon/pull/837, so it can be removed. It turned out to be a separate bug anyways.

pontoon/base/static/js/main.js
```diff
@@ -189,7 +189,8 @@ var Pontoon = (function (my) {
     /*
      * Markup placeables
      */
-    markPlaceables: function (string) {
+    markPlaceables: function (string, unsafe) {
+      var unsafe = unsafe || true;
```
mathjazz: No need for `var`.

pontoon/base/static/js/main.js
```diff
@@ -220,8 +223,8 @@ var Pontoon = (function (my) {
       });
 
       /* Leading and Trailing spaces */
-      string = markup(string, /^ /gi, 'Leading space');
-      string = markup(string, / $/gi, 'Trailing space');
+      string = markup(string, /^ /gmi, 'Leading space')
```
mathjazz: Colon.

pontoon/base/static/js/main.js
mathjazz: Also, this will return `true` if `unsafe` is `false`.

pontoon/base/static/js/main.js
```diff
@@ -243,27 +246,26 @@ var Pontoon = (function (my) {
       diff_obj.diff_cleanupSemantic(diff);
       diff_obj.diff_cleanupEfficiency(diff);
 
-      $.each(diff, function() {
+      $.each(diff, function(index) {
```
mathjazz: Thers's no need for initializing `index`.

pontoon/base/static/js/main.js
```diff
@@ -220,8 +223,8 @@ var Pontoon = (function (my) {
       });
 
       /* Leading and Trailing spaces */
-      string = markup(string, /^ /gi, 'Leading space');
-      string = markup(string, / $/gi, 'Trailing space');
+      string = markup(string, /^ /gmi, 'Leading space');
+      string = markup(string, / $/gmi, 'Trailing space');
```
mathjazz: What's the motivation for adding the `m` modifier here and not in other places within the function?

Should we add a comment?

jotes:
@mathjazz hey, I tried to fix regressions, however that required some restructuring of the code. It would be helpful if you could look if all edge-cases are handled properly e.g. when trailing space is before `</ins>` tag.
About code-style:
I don't feel that adding new methods to parent class is nice, but currently I don't see easy way to share those functions nicely between methods (maybe I need some sleep though).
Could you look again?

pontoon/base/static/js/main.js
```diff
@@ -218,20 +224,37 @@ var Pontoon = (function (my) {
...
-      /* Tab */
-      string = markup(string, /\t/gi, 'Tab character', '&rarr;');
+      }
+
+      return string;
+    },
+
+    /*
+     * Mark leading/trailng spaces in multiline strings (that contains newlines inside).
+     * Should be applied on the fully concatenated string, doesn't handle well substrings.
```
mathjazz:
```
* Mark leading/trailng spaces in multiline strings (that contain newlines inside).
* Should be applied to a fully concatenated string, doesn't handle substrings well.
```

pontoon/base/static/js/main.js
```diff
@@ -218,20 +224,37 @@ var Pontoon = (function (my) {
...
+     */
+    markWhiteSpaces: function (string) {
+      /* 'm' modifier in makes ^/$ applicable to every separate line in string, not the string as the whole. */
```
mathjazz:
```
/* 'm' modifier makes regex applicable to every separate line in string, not the string as the whole. */
```

pontoon/base/static/js/main.js
```diff
@@ -218,20 +224,37 @@ var Pontoon = (function (my) {
...
+      string = string.replace(
+        /^(<(ins|del)>)*( )/gmi,
+        '$1' + this.getPlaceableMarkup('Leading space', ' ')
+      );
+      string = string.replace(
+        /( )(<\/(ins|del)>)*$/gmi,
+        this.getPlaceableMarkup('Trailing space', ' ') + '$2'
+      );
```
mathjazz: It would be nice to split the two rules with a blank line and add comments on top (Leading space and Trailing space) like we did below.

jotes:
@mathjazz I tried to recreate that situation on my local machine (yeah, the funny part) and got following html contents: 
```
<del>Ou</del><ins>Whoo</ins>ps<del> </del>! Assurez-vous d<del>’</del><ins>'</ins>avoir <del>correctement saisi</del><ins>entré</ins> votre numéro de téléphone<ins> correctement</ins>.
```

Could you push me HTML contents which you got on your machine? I tested regexes again and still can't reproduce that effect (or, maybe I don't understand the problem :D).

jotes:
@mathjazz could you look at this again? All comments should be updated.

mathjazz:
Please remove line 232 in `pontoon/base/static/js/main.js` and we're good!

jotes:
@mathjazz fixed and rebased.

mathjazz:
I like the commit message, too! :)

PR #838, opened 2018-02-05
Fix bug 1435010 - Require SSL for connecting to PostgreSQL.

adngdb:
@mathjazz @phlax any idea how I can test that? I pushed the change to stage and nothing broke, but I do not know how to verify that the connection is using SSL.

### DIFF
--- a/pontoon/settings/base.py
+++ b/pontoon/settings/base.py
@@ -49,6 +49,12 @@ DATABASES = {
     'default': dj_database_url.config(default='mysql://root@localhost/pontoon')
 }
 
+# Ensure that psycopg2 uses a secure SSL connection.
+if not DEV and not DEBUG:
+    if 'OPTIONS' not in DATABASES['default']:
+        DATABASES['default']['OPTIONS'] = {}
+    DATABASES['default']['OPTIONS']['sslmode'] = 'require'
+
 
 # Absolute path to the directory static files should be collected to.
 # Don't put anything in this directory yourself; store your static files


jotes:
Hey Adrian,
Did you try to execute `SELECT * FROM pg_sslstatus;` in dbshell?

mathjazz:
You could [try this](https://devcenter.heroku.com/articles/heroku-postgresql#external-connections-ingress) and check for `sslmode=require`:
```
heroku pg:credentials:url DATABASE --app mozilla-pontoon
```

adngdb:
@jotes `pg_sslstatus` requires an extension that we don't have installed afaict. 

@mathjazz your solution works, thanks! Fun fact, on stage and prod, it does use ``sslmode=require``, even though my patch isn't applied. I'm very confused.

mathjazz:
> @mathjazz your solution works, thanks! Fun fact, on stage and prod, it does use sslmode=require, even though my patch isn't applied. I'm very confused.

[Most clients will connect over SSL by default, but on occasion it is necessary to set the sslmode=require parameter on a Postgres connection.](https://devcenter.heroku.com/articles/heroku-postgresql#heroku-postgres-ssl)

adngdb:
Yeah, but I'm curious why we receive an alert. I don't know if it's because we connect without SSL to our database, or if it's only because we have a legacy database. And the only way to be sure is to upgrade the database. Maybe we can do that for stage? After all, we don't care too much if it's broken for a bit, and it will allow us to be sure prod won't break when the upgrade comes for it.

adngdb:
We have decided to do nothing for now, and let tomorrow's brownout happen. At worst, we'll experience 10 minutes of downtime, which is not the end of the world. Hopefully, nothing will happen and everyone will live happily ever after. 

We will revisit this the day after tomorrow.

edmorley:
The 10 minute brownout window for Pontoon has started already:
https://papertrailapp.com/systems/mozilla-pontoon/events?focus=900479095102672903

...so looks like nothing broke.

Though worth noting that the default ssl `prefer` setting is vulnerable to MITM downgrade attacks, so ideally require should still be set.

adngdb:
Thanks @edmorley! @mathjazz, do you agree we should merge this?

PR #840, opened 2018-02-05
Update to python-fluent 0.6.x and fluent-syntax.js 0.6.x

mathjazz:
**Note: Before landing, we'll have to migrate existing projects to use the new syntax.**

This is the PR to fix [bug 1431435](https://bugzilla.mozilla.org/show_bug.cgi?id=1431435).

The `python` updates (https://github.com/mozilla/pontoon/pull/840/commits/159e6a4d3678cd2be862e2ab9f959d667e92a892) include:
* Switch to PyPI install
* Support for Terms (treated same as Messages)
* Support for GroupComments as a replacement for Section comments

We should also start adding back tests for the FTL file format which we once had, either as part of this PR or separately.

The `js` updates are described in commit messages.


pontoon/base/static/js/fluent_interface.js
```diff
@@ -398,7 +398,6 @@ var Pontoon = (function (my) {
         if (
           ast &&
           !ast.attributes.length &&
-          !ast.tags.length &&
```
stasm: Nit: `ast.{attributes,tags}.length == 0` would make this more explicit.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -373,15 +373,25 @@ var Pontoon = (function (my) {
...
+        if (element.type === 'TextElement') {
+          return true;
+        }
+
+        // Placeable
+        if (
+          element.expression &&
+          [
+            'ExternalArgument',
+            'MessageReference'
```
stasm: Perhaps it would be useful to factor this out to a `isSimplePlaceable` function which would check for any of the following [inline expression types](https://github.com/projectfluent/fluent/blob/69ecabcc06cf6f5b3654c439a9defc8575744313/spec/fluent.asdl#L46-L52):

    StringExpression(string)
    NumberExpression(number)
    MessageReference(iden id)
    ExternalArgument(iden id)
    AttributeExpression(iden id, iden name)
    VariantExpression(iden id, varkey key)
    CallExpression(fun callee, arg* arguments)

stasm: Oh, I see, this is about expressions supported in the rich editor right now. Never mind then.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -340,19 +340,17 @@ var Pontoon = (function (my) {
...
-            return [
-              'TextElement',
-              'ExternalArgument',
-              'MessageReference',
-              'SelectExpression'
-            ].indexOf(item.type) >= 0;
+          return elements.every(function(element) {
+            return (
+              self.isSimpleElement(element) ||
+              (element.expression && element.expression.variants)
```
stasm: `variants` is always a non-empty array in `SelectExpression`. Did you mean to test for the selector? That would be `element.expression.expression`.

stasm: (It's a bit weird that it's called `expression` rather than `selector`).

pontoon/base/static/js/fluent_interface.js
```diff
@@ -103,8 +103,8 @@ var Pontoon = (function (my) {
       }
 
       // Render SelectExpression
-      if (element.type === 'SelectExpression') {
-        element.variants.forEach(function (item) {
+      if (element.expression && element.expression.variants) {
```
stasm: Here as well, `variants` is always an non-empty array. Otherwise it's a syntax error.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -140,7 +119,7 @@ var Pontoon = (function (my) {
 
       // Render SelectExpression
       if (element.expression && element.expression.variants) {
-        var expression = serializeExpression(element.expression.expression);
+        var expression = fluentSerializer.serializeExpression(element.expression.expression);
```
stasm: Cool!

pontoon/base/static/js/fluent_interface.js
```diff
@@ -228,7 +228,7 @@ var Pontoon = (function (my) {
           (translationAST && !self.isSupportedMessage(translationAST)) ||
           (!translationAST && !self.isSupportedMessage(entityAST))
         ) {
-          value = entity.key + ' = \n';
+          value = entity.key + ' = ';
```
stasm: Why was this needed?

I'm asking because there's a bug in `fluent-syntax` 0.6.0 in parsing empty messages at EOF. See http://projectfluent.org/play/?gist=dcae41a5096d8fe24cc4b6fa1041534e. I wonder if perhaps you've run into it here as well?

pontoon/base/templatetags/helpers.py
```diff
@@ -268,7 +268,7 @@ def _serialize_elements(elements):
             elif type(element.expression) == ast.MessageReference:
                 response += '{ ' + element.expression.id.name + ' }'
 
-            elif hasattr(element, 'expression') and hasattr(element.expression, 'variants'):
+            elif hasattr(element.expression, 'variants'):
```
stasm: I think all these ifs should use `isinstance`. You wont' need the `hasattr` check then at all.

pontoon/base/static/js/fluent_interface.js
mathjazz: Good point. But since we use such syntax across the entire app a lot, I'd rather keep it.

pontoon/base/static/js/fluent_interface.js
mathjazz: Pontoon only uses `.parseEntry()`, but I guess that doesn't matter?

I feel like this bug was indeed causing problems in Pontoon, because we previously inserted untranslated unsupported messages with trailing `\n` and then they didn't get parsed properly when detecting unsaved changes.

mathjazz: Still, we seem to be good without trailing `\n` in 0.6.0 and also 0.4.2.

pontoon/base/static/js/fluent_interface.js
mathjazz: I'm checking if the element has variants, which effectivelly means if it's a `SelectExpression`. I could be more explicit and say `element.expression.type === 'SelectExpression'`.

But I use the same check on several occasions where `element.expression.variants` is used if the check passes, so I found the code easier to understand that way - otherwise you'd need to know that if the element is of type `SelectExpression`, it must have variants.

I think I should factor this out into something like `isSelectExpression{Element/Placeable}`, use the existing check and add a comment describing this. Thoughts?

And I agree that `element.expression.expression` is confusing. :)

pontoon/base/templatetags/helpers.py
mathjazz: So in this particular I could use this:
```python
elif isinstance(element.expression, ast.SelectExpression):
```

If you agree with the proposal in https://github.com/mozilla/pontoon/pull/840#discussion_r166584643, I'd rather keep it in line with the JS code.

pontoon/sync/formats/ftl.py
```diff
@@ -74,32 +74,30 @@ def __init__(self, path, locale, source_resource=None):
...
                 raise
 
-        def get_comment(obj):
-            return [obj.comment.content] if obj.comment else []
-
-        section_comment = None
+        group_comment = None
         for obj in self.structure.body:
-            if type(obj) == ast.Message:
+            if type(obj) in (ast.Message, ast.Term):
```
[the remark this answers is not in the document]
mathjazz: BTW, what's the benefit of using `isinstance` over `type` in this vase? I could also do:
```python
localizable_entries = (ast.Message, ast.Term)
if type(obj) in localizable_entries:
    # …
```

pontoon/base/static/js/fluent_interface.js
stasm: The bug described in http://projectfluent.org/play/?gist=dcae41a5096d8fe24cc4b6fa1041534e was fixed in `fluent-syntax` 0.6.2. I updated the Playground to use it as well.

pontoon/base/static/js/fluent_interface.js
stasm: The `Cool` was about using the upstream `serializeExpression`.

I do have another comment about `isSelectExpressionElement`: How is it better than the following? :)

```js
if (element.type === `SelectExpression`) {
```

pontoon/base/static/js/fluent_interface.js
```diff
@@ -722,6 +722,12 @@ var Pontoon = (function (my) {
           }
         }
 
+        // Overcome bug in fluent-syntax 0.6.0 that incorrectly parses
+        // empty messages without trailing newlines as Messages instead of Junk
```
stasm: This bug has been fixed in `fluent-syntax` 0.6.2. Do we still need this workaround?

flodolo:
The serialized files for Screenshots, Test Pilot and Send look good.

Here are some issues found during testing, some are probably long standing, other might not be blocking this release.

#### Terms in Advanced Mode

https://mozilla-pontoon-staging.herokuapp.com/it/test-fluent/all-resources/?string=175567

Switch to Advanced Mode, switch the term to a message (remove the leading `-`). Then save.

At this point there are two distinct translations stored in HISTORY, the current being a message (not a term), but when you switch back to advanced mode it shows up as a term.

Note that it's serialized correctly as a message.

#### Error in console when copying placeholder

`TypeError: o is undefined translate.min.7249a872f3dd.js:1387:616`

https://mozilla-pontoon-staging.herokuapp.com/it/test-fluent/all-resources/?string=175554
Switch to Advanced Mode, then back to normal mode, then click the placeholder without focusing the input field

#### Parsing and serialization for the 'multiple selectors' example is broken

This string is completely off
https://mozilla-pontoon-staging.herokuapp.com/it/test-fluent/all-resources/?string=175564

Source

```properties
selector-multi =
   There { $num ->
       [one] is one email
      *[other] are many emails
   } for { $gender ->
      *[masculine] him
       [feminine] her
}
```

Result

```properties
selector-multi = { $gender ->
        [Value] x
        [Value] x
        [one] x
        [other] x
        [Value] x
        [masculine] x
       *[feminine] x
    }
```

Same result in the serialized file.

#### Placeable with expression is not highlighted

https://mozilla-pontoon-staging.herokuapp.com/it/test-fluent/all-resources/?string=175569

#### Blank spaces in source

Significant blank spaces should be highlighted in source, e.g. spaces surrounding "for" in

https://mozilla-pontoon-staging.herokuapp.com/it/test-fluent/all-resources/?string=175564

#### Advanced mode on empty complex strings

Switching to Advanced Mode on a string with attributes is always leading to weird situations.

https://mozilla-pontoon-staging.herokuapp.com/it/test-fluent/all-resources/?string=175566

Without any translations, switching to Advanced Mode brings only up the ID with an equal sign. If you write something, then switch back to normal mode, you lost the UI for attributes.

If you clear the string, than go back and forth you recover the UI.

Should there be some checks to avoid this? Should the structure of the FTL be pre-filled?

For example, now I get:

```properties
delete-all-message = 
```

Maybe I should get:

```properties
delete-all-message = { $num ->
        [one]
       *[other]
    }
```

mathjazz:
I fixed 1, 2 and 6 (the checks part).

I don't think I'll fix the rest as part of this PR, but I'll look into 3.

flodolo:
Now I can't switch back to normal mode.

```
TypeError: e.id is undefined
```

For 6: the message says "Make sure to fill in the value", but there's already a value in my translation.

mathjazz:
Fixed and updated stage.

flodolo:
Seems to work correctly now.

I think it's still worth discussing the behavior for 6, when switching to Advanced mode when the string is still empty, but that's material for a bug.

mathjazz:
This is a map of items from the [bug report](https://github.com/mozilla/pontoon/pull/840#issuecomment-365193253) and their corresponding fixes:

1: dce35a9e7f494c8f2c7ecb269e7b5cff83ae6ada
2: 70b85239b4836f469b8f7548a2078b8960dab254
3: 09577a58d1c7364f7adc961ec475c30a8dba71f4
6: 8b6ea6ad6b98c5ccaecb6c7a9e8a8a23ce3ebed0

I've also filed 2 bugs (5 is already tracked):
4: https://bugzilla.mozilla.org/show_bug.cgi?id=1437859
6: https://bugzilla.mozilla.org/show_bug.cgi?id=1437857

PR #841, opened 2018-02-05
Fix bug 1410569: Translation author not set to user if there are no changes

VishalCR7:
Earlier the user who hits the save button for a translation without making any changes is set as author for the translation if author is not set. I have removed the condition which sets the author for non changing translations

### DIFF
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -545,9 +545,6 @@ def update_translation(request):
                     fuzzy=False,
                 )
 
-                if t.user is None:
-                    t.user = user
-
                 t.approved = True
                 t.approved_date = timezone.now()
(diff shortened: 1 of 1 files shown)


VishalCR7:
@mathjazz  please review.

pontoon/base/views.py
```diff
@@ -545,9 +545,6 @@ def update_translation(request):
                     fuzzy=False,
                 )
 
-                if t.user is None:
-                    t.user = user
-
                 t.approved = True
                 t.approved_date = timezone.now()
```
mathjazz: This line can be removed, because the `approved_date` is set in line 560.

VishalCR7:
@mathjazz I have made the requested changes

PR #842, opened 2018-02-06
Touch Bug 1436149: Entities perf hacks - translated resource idx

phlax:
Adds a unique together for translated resource, (locale, resource)

this gives 10-20 % speed up for (EDIT: *db timings in...*) project locale entities view in my testing.

this PR includes some hacks to debug the sql, and check indeces etc if required, which will be removed before landing


phlax:
the perf improvement was ~10% on db - so barely noticable in total render - but yep agreed the unique together is a good idea,  and i think taking even a little strain from the db for a common view is a good thing.

PR #846, opened 2018-02-09
Fix Bug 1436191: Entity resource order unique

phlax:
adds a migration to fix the current dupes, and then adds the constraint.

There are a lot of dupes in the database (~9505) so im wondering if there is anywhere that could create this problem still.


mathjazz:
I actually see 9765 such cases on prod.

For 9 of them `date_created` says `datetime.datetime(2017, 4, 10, 9, 48, 16, 934267, tzinfo=<UTC>)`, the rest are even older (no `date_created`). We might have fixed the bug since them, but it's hard to tell.

Vast majority of entities are created here:
https://github.com/phlax/pontoon/blob/a299c6e8c27b1c0a3412e999d8df5e13d9887367/pontoon/sync/changeset.py#L132

phlax:
> I actually see 9765 such cases on prod.

hmm that implies to me that they are still being created as there were only 9505 on my local copy of staging.

i think with the addition of the order in `Entities.create` the problem should not recur - unless there is somewhere that could batch create entities without order

phlax:
...either way with this PR the db wont allow it to happen any more. I think my biggest concern was that it could cause a fail eg in sync and mess the db/fs

mathjazz:
Yeah. The good news is that all these entities belong to one file in one project, which is now disabled: https://pontoon.mozilla.org/admin/projects/snippets/.

pontoon/base/migrations/0116_fix_ambiguous_entity_order.py
```diff
@@ -0,0 +1,51 @@
...
+            # update on the boundary
+            order = 0
+            resource = _resource
+        else:
+            order += 1
+        fixed_entities[order] = fixed_entities.get(order, [])
+        fixed_entities[order].append(pk)
+
+    for order, pks in fixed_entities.items():
+        Entity.objects.filter(pk__in=pks).update(order=order)
```
mathjazz: In theory we should reset the order based on the order in the actual file.

Anyhow, since the inconsistencies are [not that important,](https://github.com/mozilla/pontoon/pull/846#issuecomment-365400510) we should be good.

It would be nice to add a comment explaining how the algorithm actually works.

phlax: ill add some comments

PR #847, opened 2018-02-09
Remove filter/distinct from Locale aggregate method 

phlax:
this might be required - im not sure - but in testijng the sql i always got the same results - and it took about 90% off of the query time. This method is used quite a lot i think


mathjazz:
The patch looks OK.

It's mostly used during sync and in project admin, but also in translation batch actions, where it might have the most noticeable impact.

PR #849, opened 2018-02-10
Add index on locale, user, entity

phlax:
This is a massive optimization for the contributors view - at least for locales

In my testing this speeds up the v slow query ~200x

Before:
```
 Execution time: 103.502 ms
```

After:

```
 Execution time: 0.486 ms

```


phlax:
> I wonder if we can do something similar for other /contributors views.

sadly i wasnt able to reproduce this for any other view, nor do i think we can. From my experiments just doing the aggregation for the bigger views with the sql stripped down and every index possible is outside of the desired time.

Imho we are gonna need to denorm user project locale stats - this would allow us to aggregate them sufficiently efficiently, and would not be ~too costly to calculate on translation changes

PR #852, opened 2018-02-11
[WIP] Add react loaders

phlax:
This adds 2 loaders for rendering react components into the DOM

## `Loader`

This will load the component specified in `this.component` to all nodes found by `this.selector`. Each node will get the props that are created by calling `this.getProps` with the `node`

By default the component will be rendered into the found node, but this can be changed by overriding `this.getRenderNode` which is called with the `node`

## `ToggleLoader`

This will listen for (by default click) events on the nodes found by `this.selector`.

When the event happens the component is loaded or unloaded, and a toggle indicator if found is toggled.

Like the generic loader you must specify `this.component`, and can specify `this.getProps` and `this.getRenderNode`

@adngdb , @Pike i know this is not the way that you guys suggested, but i already needed a variant of the loader, so i have separated out the code for further discussion.

for reference here are the 2 uses of the loaders:

https://github.com/phlax/pontoon/blob/bd8ac8bd083a2756dcc23f92dee0777399620e47/pontoon/static/js/tags/locales/index.js

https://github.com/phlax/pontoon/blob/4a629a8522d86d8e431146f0e0a6a0d532236441/pontoon/static/js/tags/admin/index.js

One of the benefits of using these is that we can write some tests for these loaders, and any other loaders just need to be tested for their specific properties

### DIFF
--- a/pontoon/static/js/loaders/click.js
+++ b/pontoon/static/js/loaders/click.js
@@ -1,0 +1,46 @@
+
+import Loader from './loader'
+
+
+export default class ClickLoader extends Loader {
+    // instead of loading components at the nodes identified by
+    // `this.selector`, this Loader listens for clicks (or other events)
+    // on them and loads the components when they happen
+
+    // When this loader first loads it will check if any of the selected
+    // nodes are "active" and simulate the trigger event if it is.
+
+    constructor () {
+        super();
+        this.handle = this.handle.bind(this);
+    }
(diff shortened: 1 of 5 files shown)


Pike:
I guess both Adrian and I are happy to look at is once it's stable. Please request reviews then.

phlax:
this is no longer required

PR #853, opened 2018-02-11
[WIP] String tiers project js

phlax:
Use React to render the project locale tag data.

This can be either done inline or as a more lightweight way to render the data.


mathjazz:
Closing, since this is already in master in another form or part of #874.

PR #854, opened 2018-02-12
Touch Bug 1243129: Mixed bag of JS cleanups/updates

phlax:
this 
- fixes a js test error
- separates out some of the js tests 
- adds a `NotImplementedError`
- adds an l10n Datetime formatter
- ~adds `Humanize` lib for l10n numbers etc~

### DIFF
--- a/pontoon/static/js/utils/datetime.js
+++ b/pontoon/static/js/utils/datetime.js
@@ -1,0 +1,32 @@
+
+export default class Datetime {
+
+    constructor (datetime, locale) {
+        this.datetime = new Date(datetime);
+        this.locale = locale || 'en-GB';
+    }
+
+    get date () {
+        return new Intl.DateTimeFormat(
+            this.locale, this.dateFormat).format(this.datetime);
+    }
+
+    get dateFormat () {
+        return {
+            day: 'numeric',
+            month: 'long',
+            year: 'numeric'}
+    }
+
+    get time () {
+        return new Intl.DateTimeFormat(
+            this.locale, this.timeFormat).format(this.datetime);
+    }
+
+    get timeFormat () {
+        return {
+            hour: '2-digit',
+            minute: '2-digit',
+            second: '2-digit'}
+    }
+}
--- a/tests/js/utils/datetime.js
+++ b/tests/js/utils/datetime.js
@@ -1,0 +1,25 @@
+
+import Datetime from 'utils/datetime';
+
+
+test('Datetime constructor', () => {
+    let dt = new Datetime(0);
+
+    // this.datetime is set with a new Date
+    expect(dt.datetime).toEqual(new Date(0));
+
+    // default locale is en-GB
+    expect(dt.locale).toBe('en-GB');
+
+    // locale can be overridden
+    dt = new Datetime(1, 'B');
+    expect(dt.datetime).toEqual(new Date(1));
+    expect(dt.locale).toBe('B');
+})
+
+
+test('Datetime date', () => {
+    let dt = new Datetime(7723);
+    expect(dt.date).toBe("January 1, 1970");
+    expect(dt.time).toBe("12:00:07 AM");
+})
--- a/pontoon/static/js/errors/index.js
+++ b/pontoon/static/js/errors/index.js
@@ -1,0 +1,10 @@
+
+export class NotImplementedError extends Error {
+
+    constructor(...args) {
+        super(...args)
+        Error.captureStackTrace(this, NotImplementedError)
+    }
+}
+
+NotImplementedError.prototype = Error.prototype;
--- a/tests/js/utils/ajax.js
+++ b/tests/js/utils/ajax.js
@@ -1,6 +1,5 @@
 
 import {DjangoAjax, PontoonDjangoAjax, ajax} from 'utils/ajax';
-import {Strip, strip} from 'utils/strip';
 
 
 test('PontoonDjangoAjax instance', () => {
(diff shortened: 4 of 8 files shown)


pontoon/static/js/utils/datetime.js
```diff
@@ -0,0 +1,32 @@
...
+            this.locale, this.timeFormat).format(this.datetime);
+    }
+
+    get timeFormat () {
+        return {
+            hour: '2-digit',
+            minute: '2-digit',
+            second: '2-digit'}
+    }
+}
```
adngdb: This is something that [`moment.js`](http://momentjs.com/) (for example) does very well. Why reimplement that logic?

package.json
```diff
@@ -62,6 +62,7 @@
     "enzyme-adapter-react-16": "^1.1.1",
     "eslint": "^4.15.0",
     "eslint-plugin-react": "^7.5.1",
+    "humanize-plus": "^1.8.2",
```
adngdb: I don't see this being used anywhere in this PR, why include it?

tests/js/utils/datetime.js
```diff
@@ -0,0 +1,35 @@
...
+
+    // date and time formatting is specified
+    expect(dt.dateFormat).toEqual({
+        day: 'numeric',
+        month: 'long',
+        year: 'numeric'})
+    expect(dt.timeFormat).toEqual({
+        hour: '2-digit',
+        minute: '2-digit',
+        second: '2-digit'})
```
adngdb: These two assertions are basically testing that you didn't make a spelling mistake in your code. I really fail to see what value this provides over just proof-reading your code (instead of writing it down twice).

package.json
phlax: its in the next PR - i wanted to clear as many commits as possible - and use of this lib seemed like a separate conv - so i pushed it up front

tests/js/utils/datetime.js
phlax: im in the DRY-except-in-unit-tests school of thought here (see here for interesting discussion http://codebetter.com/karlseguin/2009/09/12/unit-testing-do-repeat-yourself/)

i would view this kind of test as more harness than logic check

pontoon/static/js/utils/datetime.js
phlax: no real objection to that - i was just trying to replicate code from here https://github.com/mozilla/pontoon/blob/master/pontoon/base/static/js/table.js#L10

tests/js/utils/datetime.js
adngdb: This is not about repeating code, it's about the test doing nothing but verifying that you, the developer, didn't make a spelling mistake. It's not testing a feature, or a side-effect, it's just checking you wrote the right thing by writing that same thing again. 

If I try to simplify it, it's the same as doing this:
```javascript
var answer = 42;

test('answer', () => {
    expect(answer).toEqual(42);
});
```

And I honestly do not see what value that adds.

package.json
adngdb: I'd rather keep things in context, as I said in my review comment. Right now you have me confused because you add a dependency without any explanation of what that dep will be used for. I would prefer reading more code, but understand what the code is doing, than the opposite.

phlax: ill shift this commit to where its required

tests/js/utils/datetime.js
phlax: >  it's about the test doing nothing but verifying that you, the developer, didn't make a spelling mistak

its about future you, me and others not messing with this unless they are prepared to re/move/update tests

mathjazz:
> It just dawned on me why I'm confused by those Tiers PRs: you're separating things into horizontal slices instead of vertical ones. And so a PR like this one is terribly difficult to understand, because I do not see where and how this code is being used.

What so you have in mind as horizontal vs. vertical slices? "Model", "JS cleanups", "utils" vs. "admin ui for string tiers", "string tiers in project dashboard", "string tiers in filters" etc.?

phlax:
> What so you have in mind as horizontal vs. vertical slices? "Model", "JS cleanups", "utils" vs. "admin ui for string tiers", "string tiers in project dashboard", "string tiers in filters" etc.?

both are there, but we are still on the model, cleanups and utils bit

pontoon/static/js/utils/datetime.js
julen: > This is something that moment.js (for example) does very well. Why reimplement that logic?

What are the browser requirements for Pontoon? When browsers can natively handle the basic needs outlined here via the `Intl` API, IMHO it wouldn't be wise to pull in a big dependency such as moment.js.

adngdb: Browser requirements for Pontoon are non-existent at the moment. We're going to work on them alongside those of Fluent in the near future.

adngdb:
@mathjazz that's correct. 

@phlax both are there indeed, but I've only been reviewing horizontal slices so far. I'd rather review that code in a vertical-slice-PR, so that I can understand the context.

pontoon/static/js/utils/datetime.js
phlax: @adngdb can we go with this for now then please, as its basically doing what we do now. We can update if we feel we need some lib to handle this

adngdb: Yeah, I don't care too much about that. I'm using `moment.js` for Translate.Next so I figured we might as well use it here as well, but it's not super important.

tests/js/utils/datetime.js
adngdb: This however is a blocker for me.

phlax: how do you propose to test this - or are you suggesting we dont ?

pontoon/static/js/utils/datetime.js
julen: > I'm using moment.js for Translate.Next

What's this and how does it affect Pontoon?

phlax: its a proposed rewrite of the translation ui

julen: Thanks. Is there a place it can be checked out? Or where can I learn more about it?

phlax: atm i think the discussion is mostly internal - there are some bugs on bugzilla - @adngdb is the person driving the update

ill see if i can find relevant bugs...

phlax: https://bugzilla.mozilla.org/show_bug.cgi?id=1243129

tests/js/utils/datetime.js
adngdb: Yep, I think we should just not have those 2 tests.

phlax: i really dont agree - and im tempted to refactor the class to get implict coverage, but pretty reluctant to do that.

Ill remove these tests for the sake of getting this landed

pontoon/static/js/utils/datetime.js
adngdb: @julen At the moment there's a document you can read about it, that explains pretty well what Translate.Next is: https://docs.google.com/document/d/1HDtCUMeogo2kjD5aNjulnlxSaC5kr928w5VnyXiAN3o/edit#

I haven't made any public announcements yet, but that will come. I want to involve the community early in the process of designing this, but I haven't taken the time to do that just yet.

julen: Thanks @adngdb, that pretty much answers the questions I had with my localizer hat 👍 Will plan to have a deeper look at that in the near future.

PR #857, opened 2018-02-13
Touch Bug 1243129: React data manager

phlax:
Adds 

- ~~a data table component that fetches data and builds a reacttable component onMount~~
- ~~a table with checkboxes that can be selected and submitted~~

Adds a higher order component that wraps the wrapper component with a "manager" that retrieves/submits data from/to an api.

The wrapped component is passed `data`  and `errors` , and functions to `refreshData` and `handleSubmit` as props

~~blocked by #854~~

required for strings tiers project and admin views

### DIFF
--- a/pontoon/static/js/utils/csrf.js
+++ b/pontoon/static/js/utils/csrf.js
@@ -1,0 +1,4 @@
+
+export const getCSRFToken = () => {
+    return document.querySelector('input[name=csrfmiddlewaretoken]').value;
+}
--- a/pontoon/static/js/utils/data.js
+++ b/pontoon/static/js/utils/data.js
@@ -1,0 +1,85 @@
+
+import React from 'react';
+
+import {ajax} from 'utils/ajax';
+
+
+export class DataManager {
+
+    constructor (state) {
+        this.state = state;
+    }
+
+    get data () {
+        return this.state.data;
+    }
+
+    get errors () {
+        return this.state.errors;
+    }
+}
+
+
+export function dataManager(WrappedComponent, Manager, data) {
+
+    return class Wrapper extends React.Component {
+
+        constructor(props) {
+            super(props);
+            this.state = {errors: {}, data: data};
+        }
+
+        get api () {
+            return this.props.api;
+        }
+
+        get requestMethod () {
+            return this.props.requestMethod || 'get';
+        }
+
+        get submitMethod () {
+            return this.props.submitMethod || this.requestMethod;
+        }
+
+        refreshData = async (params) => {
+            return this.handleResponse(
+                await ajax.fetch(
+                    this.api,
+                    params,
+                    {method: this.requestMethod}));
+        }
+
+        handleResponse = async (response) => {
+            const {status} = response;
+            const json = await response.json();
+            if (status === 200) {
+                const {data} = json;
+                this.setState({data, errors: {}});
+                return;
+            }
+            const {errors} = json;
+            this.setState({errors});
+        }
+
+        handleSubmit = async (params) => {
+            return this.handleResponse(
+                await ajax.fetch(
+                    this.api,
+                    params,
+                    {method: this.submitMethod}));
+        }
+
+        render() {
+            Manager = Manager || DataManager;
+            const manager = new Manager(this.state);
+            return (
+                <WrappedComponent
+                   manager={manager}
+                   errors={manager.errors}
+                   data={manager.data}
+                   handleSubmit={this.handleSubmit}
+                   refreshData={this.refreshData}
+                   {...this.props} />);
+        }
+    };
+}
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -7,6 +7,7 @@ module.exports = {
         es6: true,
         browser: true,
     },
+    parser: "babel-eslint",
     parserOptions: {
         ecmaVersion: 2017,
         ecmaFeatures: {
--- a/package-lock.json
+++ b/package-lock.json
@@ -52,6 +52,192 @@
         }
       }
     },
+    "@babel/generator": {
+      "version": "7.0.0-beta.40",
+      "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.0.0-beta.40.tgz",
+      "integrity": "sha512-c91BQcXyTq/5aFV4afgOionxZS1dxWt8OghEx5Q52SKssdGRFSiMKnk9tGkev1pYULPJBqjSDZU2Pcuc58ffZw==",
+      "dev": true,
+      "requires": {
+        "@babel/types": "7.0.0-beta.40",
+        "jsesc": "2.5.1",
+        "lodash": "4.17.4",
+        "source-map": "0.5.7",
+        "trim-right": "1.0.1"
+      },
+      "dependencies": {
+        "jsesc": {
+          "version": "2.5.1",
+          "resolved": "https://registry.npmjs.org/jsesc/-/jsesc-2.5.1.tgz",
+          "integrity": "sha1-5CGiqOINawgZ3yiQj3glJrlt0f4=",
+          "dev": true
+        }
+      }
+    },
+    "@babel/helper-function-name": {
+      "version": "7.0.0-beta.40",
+      "resolved": "https://registry.npmjs.org/@babel/helper-function-name/-/helper-function-name-7.0.0-beta.40.tgz",
+      "integrity": "sha512-cK9BVLtOfisSISTTHXKGvBc2OBh65tjEk4PgXhsSnnH0i8RP2v+5RCxoSlh2y/i+l2fxQqKqv++Qo5RMiwmRCA==",
+      "dev": true,
+      "requires": {
+        "@babel/helper-get-function-arity": "7.0.0-beta.40",
+        "@babel/template": "7.0.0-beta.40",
+        "@babel/types": "7.0.0-beta.40"
+      }
+    },
+    "@babel/helper-get-function-arity": {
+      "version": "7.0.0-beta.40",
+      "resolved": "https://registry.npmjs.org/@babel/helper-get-function-arity/-/helper-get-function-arity-7.0.0-beta.40.tgz",
+      "integrity": "sha512-MwquaPznI4cUoZEgHC/XGkddOXtqKqD4DvZDOyJK2LR9Qi6TbMbAhc6IaFoRX7CRTFCmtGeu8gdXW2dBotBBTA==",
+      "dev": true,
+      "requires": {
+        "@babel/types": "7.0.0-beta.40"
+      }
+    },
+    "@babel/highlight": {
+      "version": "7.0.0-beta.40",
+      "resolved": "https://registry.npmjs.org/@babel/highlight/-/highlight-7.0.0-beta.40.tgz",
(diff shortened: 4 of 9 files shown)


pontoon/static/js/widgets/data.js
```diff
@@ -0,0 +1,104 @@
...
+            await ajax[this.submitMethod](this.api, params));
+    }
+
+    render () {
+        return '';
+    }
+}
+
+
+export class DataTable extends DataManager {
```
julen: While this might work, note how inheritance is not idiomatic React and will result foreign to people familiar with it:
https://reactjs.org/docs/composition-vs-inheritance.html

When willing to separate data handling from visuals, the common approach is to externalize state handling in a parent container component, which then passes all the needed data/handlers down to presentational components as props. These presentational components needn't have any state, and hence can be simple functions (provided they don't need lifecycle methods).

There are many articles online explaining all of this, most notably:
* https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0
* https://medium.com/@dan_abramov/mixins-are-dead-long-live-higher-order-components-94a0d2f9e750

All in all explicit composition is favored over inheritance.

julen: (Another common pattern is to organize components in individual files named after the component, e.g. `DataTable.js`, usually having one component per file.)

phlax: @julen thanks for review

yep, i am mostly trying to use composition and avoid inheritance for react widgets

in this particular case its what happens in `componentDidMount` that is being inherited more than anything else - so i dont really see how that could be done with composition

for the CheckboxTable that i have also added in this PR it could be done through composition - but still im not entirely convinced of the argument for doing so

phlax: re organising 1 per file - again im mostly following that - but in some cases i throw them together, expecting that in the future they wiill be separated and then re-exported from index.js

julen: > in this particular case its what happens in componentDidMount that is being inherited more than anything else - so i dont really see how that could be done with composition

I believe you are looking for Higher-order Components: https://reactjs.org/docs/higher-order-components.html

adngdb: On that topic of HIC, I've found this article quite interesting: http://americanexpress.io/faccs-are-an-antipattern/

pontoon/static/js/utils/data.js
```diff
@@ -0,0 +1,90 @@
...
+            this.component.setState({data, errors: {}});
+            return;
+        }
+        const {errors} = json;
+        this.component.setState({errors});
+    }
+
+    async handleSubmit (params) {
+        return this.handleResponse(
+            await ajax[this.submitMethod](this.api, params));
```
adngdb: Why not use `ajax.fetch` here? It has some built-in protections that might be convenient. It would also make it a bit less difficult to track down code paths while debugging.

pontoon/static/js/utils/data.js
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    async refreshData (params) {
+        if (this.initialData) {
+            this.component.setState({data: this.initialData, errors: []});
+            this.initialData = undefined;
+            return
+        }
+        return this.handleResponse(
+            await ajax[this.requestMethod](this.api, params));
```
adngdb: Same question about using `ajax.fetch`.

pontoon/static/js/utils/data.js
```diff
@@ -0,0 +1,90 @@
...
+        }
+        const {errors} = json;
+        this.component.setState({errors});
+    }
+
+    async handleSubmit (params) {
+        return this.handleResponse(
+            await ajax[this.submitMethod](this.api, params));
+    }
+}
```
adngdb: I'm confused by the design of this class. It seems to be doing 2 things, holding data and querying that data. I'm quite convinced that in your place I would have written that logic as a function instead of a class, and let a different structure hold the data. 

I guess I need to see it in action in order to understand it better. Could you show me an instance of code where you use it?

pontoon/static/js/utils/csrf.js
```diff
@@ -0,0 +1,8 @@
+
+
+export class PontoonCSRF {
+
+    get value () {
+        return document.querySelector('input[name=csrfmiddlewaretoken]').value;
+    }
+}
```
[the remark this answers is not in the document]
phlax: ok - i agree here that func would probs be simpler than class

pontoon/static/js/utils/data.js
```diff
@@ -0,0 +1,90 @@
...
+import {ajax} from 'utils/ajax';
+
+
+export class DataManager {
+
+    constructor (component, data) {
+        this.component = component;
+        this.handleResponse = this.handleResponse.bind(this);
+        this.handleSubmit = this.handleSubmit.bind(this);
+        this.refreshData = this.refreshData.bind(this);
```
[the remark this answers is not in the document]
phlax: yep, like it. well - i feel a bit uncomfortable about the implicitness of the difference between a property function and normal one - but the whole copying out all the methods thing in the constructor is a pita, so wfm!

my only concern re this PR, is yep babel was a pita to set up for class properties - i can look at adding

pontoon/static/js/utils/data.js
phlax: ill update

pontoon/static/js/utils/data.js
phlax: ultimately it provides a "HOC" function that you can pass in an `object` too - that object can be used to:

- specify the components to be used in rendering
- handle data mangling
- be the (adaptable) "single source of truth"

my reading of the HOC docs is that the data provider can be any object

the slight deviation i made from the standard pattern was that instead of embedding all of the data mangling functionality in the react component its pulled out to a class.

its used in both the admin and project uis

in the admin - v little is overridden in the class - the default could be switched in this case and could use the generic datamanager - although it probs makes sense to keep the default as `get`

the manager is set up here:

https://github.com/mozilla/pontoon/pull/815/files#diff-a58b75723f571a686ce4ade32fc3eeffR61

the refreshData and submit methods are then passed down, and are used here:

https://github.com/mozilla/pontoon/pull/815/files#diff-a58b75723f571a686ce4ade32fc3eeffR13

Using a class allows much more to be overriden - as for example here where the dashboard is specified:

https://github.com/mozilla/pontoon/pull/853/files#diff-c29b1fca17c37a5c3c590455ac4986c0R6

Or the very sparse code required for specifying specific dashboards:

https://github.com/mozilla/pontoon/pull/853/files#diff-917435cc33dad6f9255ad48eb8a951a2R10

https://github.com/mozilla/pontoon/pull/853/files#diff-fb5c63ce89f3847b2c2101ebffeb3092R11

phlax: updating the code - parts of this functionality could defo be done *inside* the func, but it would still be in a class - ie the `WrappedComponent` - this pattern keeps the composability but also avoids putting all of the functionality into an anonymous class

pontoon/static/js/utils/data.js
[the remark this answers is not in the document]
phlax: updated to use class properties - i had to add `babel-eslint` , but it otherwise seems to work

pontoon/static/js/utils/data.js
phlax: this has been separated into HOC and and data provider for the sake of landing the code

pontoon/static/js/utils/data.js
```diff
@@ -0,0 +1,85 @@
...
+                <WrappedComponent
+                   manager={manager}
+                   errors={manager.errors}
+                   data={manager.data}
+                   handleSubmit={this.handleSubmit}
+                   refreshData={this.refreshData}
+                   {...this.props} />);
+        }
+    };
+}
```
adngdb: The content of this file is too complex, there are too many concepts and it is not obvious from the outside what this does, or how one uses it. I for one know that I won't use it because I'm sure next time I look at this I will have forgotten how it works. I'm not even 100% sure what problem you're solving with this HOC. In short, I don't think this follows the KISS principle -- it's not simple, and definitely not stupid.

pontoon/static/js/utils/data.js
```diff
@@ -0,0 +1,85 @@
...
+            this.setState({errors});
+        }
+
+        handleSubmit = async (params) => {
+            return this.handleResponse(
+                await ajax.fetch(
+                    this.api,
+                    params,
+                    {method: this.requestMethod}));
+        }
```
adngdb: `refreshData` and `handleSubmit` do the same thing now.

phlax: this should be `{method: this.submitMethod}` - ill update

pontoon/static/js/utils/data.js
phlax: im not sure how i could have done it better - but hopefully time and usage will show us

there are some things that are really common patterns and in react-land HOCs are the way to factor that out - ive tried to stay by the book in what we eventually landed this (even tho it meant you cant do what i had hoped to at a higher level of abstraction. /endmoan)

what is not really addressed/factored into this code is how pages are reloaded with partial content <> full page loads. But my mission is `string_tiers` so im really reluctant to think through those issues in this context.

either way - i think we have something usable - if not, its something to crititque/rubbish/remove/replace - but in the meantime we can use it 8/

phlax: > what is not really addressed/factored into this code is how pages are reloaded with partial content <> full page loads

@adngdb on this point i think it should `Object.assign(state.data, response.data)`

julen: I'm missing the context so forgive my ignorance but since I read the code snippet above, please note [state should be treated as if it was immutable](https://reactjs.org/docs/react-component.html#state):

> Never mutate `this.state` directly, as calling `setState()` afterwards may replace the mutation you made. Treat `this.state` as if it were immutable.

So something like `Object.assign({}, state.data, response.data)` is preferable, and then `setState()` as needed.

phlax: hi - sorry that was more meant to be pseudo code for what should happen, rather than the implementation

julen: Cool, the code drew my attention — sorry for the noise!

adngdb: @phlax I don't follow, what do you mean by
> what is not really addressed/factored into this code is how pages are reloaded with partial content <> full page loads

?

phlax: if you make a json call back - eg for partial content for a dashboard - you dont want to cane the existing `state` object more add to/update it

phlax: to (hopefully) clarify a bit better - here is the implementation/change i had in mind:

```
@@ -54,7 +58,7 @@ export function dataManager(WrappedComponent, Manager, data) {
             const json = await response.json();
             if (status === 200) {
                 const {data} = json;
-                this.setState({data, errors: {}});
+                this.setState(prevState => ({data: Object.assign(prevState.data, {...data}), errors: {}}));
                 return;
             }
```

adngdb: That seems counter intuitive to me. Wouldn't the service you're calling send all the data you need? Maybe it can send back an object that doesn't have a key the previous state had, for good reason. In such a case, your state would still keep the obsolete key.

phlax: i can show you some code, just not yet - but i think this change does make sense in the context of sometimes retrieving all data and sometimes only updating the existing

PR #860, opened 2018-02-14
[WIP] NOBUG - Request for comments - Move markPlaceables outside main.js

jotes:
@mathjazz
Hey, so after my declaration at our irc channel I experimented with code related to markPlaceables (I feel kinda forced by struggles introduced during work on #836).
Important disclaimer: So, this PR isn't like ready to merge thing. I haven't add tests yet, but I feel it can be (easily) done later. `placeables.js` is built by webpack. Placeables are later visible as `placeables.markAll()` in `main.js`

The question is if we want to pursue such way of working with the legacy Pontoon js -> move some parts of `main.js` code to separate modules and expose them globally (via `window`).

From my perspective that approach has following benefits:
* you can use sort-of name-spaces instead of one god-like objects
* unit-testing

I feel that those things still can be valuable during the transition phase to Translate.next and to add some anti-regression tests.
Aaaand there's a huge chance that there are better approaches to handle such legacy code in Javascript -> I'm open to learn something new.
If We want to pursue that, I'll create a proper PR to add support for legacy modules and later add PR for refactoring which will move placeables, fixes issue with leading/trailing spaces (1-2 days of afterhours :D).

Feel free to ping Adrian/Ryan -> I'm not sure who's less busy right now :(

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,13 +1,19 @@
 {
   "presets": [
-     "env",
+     ["env", {
+      "modules": false,
+      "targets": {
+        "yuglify": true,
+      }
+     }],
      "stage-2",
      "react"
   ],
   "plugins": [
-     "transform-runtime"
+     "transform-runtime",
+     "transform-es2015-modules-commonjs"
   ],
   "ignore": [
-     "node_modules",
+     "node_modules"
   ]
 }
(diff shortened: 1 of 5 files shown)


phlax:
imho we should leave functions like this in the legacy code until we have es6 code that uses it.

Trying to mix the 2 is confusing and much as i would like to see some test coverage on this code, i think the way to migrate/test it is with the code that uses it

jotes:
After discussion at work week We decided to wait till `translate.next` because placeables will be rewritten anyway.

PR #861, opened 2018-02-20
Added rel="noopener noreferrer" to header links

psiinon:


jotes:
@psiinon does this issue affects https://github.com/mozilla/pontoon/blob/master/pontoon/base/static/js/translate.js#L152 too?

mathjazz:
Thanks for the patch @psiinon!

@jotes The links in `translate.js` are OK, because they are internal. In `main.js` however we have 4 external `target="_blank"` link references, which all need to be updated. I'll take care of them in a separate patch.

PR #863, opened 2018-02-21
Bug 1439650 - Migrate to analytics.js

jotes:
* Create a template for Google Analytics snippet and point to it in base templates
* Update the inline js snippet
* Update related CSP hashes and update CSP Rules

@mathjazz / @phlax Could you review and then push it to the staging to check if migrations works?

I tested it locally and analytics.js sends data to `https://www.google-analytics.com/collect?v=1&...` (it uses pixel tracking).

### DIFF
--- a/pontoon/base/templates/base.html
+++ b/pontoon/base/templates/base.html
@@ -21,22 +21,7 @@
       {% endblock %}
     {% endblock %}
 
-    {% if settings.GOOGLE_ANALYTICS_KEY %}
-    <!-- Google Analytics -->
-    <script type="text/javascript">
-
-      var _gaq = _gaq || [];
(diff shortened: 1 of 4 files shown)


pontoon/base/templates/google_analytics.html
```diff
@@ -0,0 +1,14 @@
+{% if settings.GOOGLE_ANALYTICS_KEY %}
```
adngdb: I'd rename this file to `analytics`, or even better, `tracker.html` so that in the future, if we're to change tracker, we don't have to rename.

jotes: Done :) I decided to rebase the pull request because it's a small change and it's still merge-able.

phlax:
lgtm

PR #865, opened 2018-02-22
Prevent Fluent UI from breaking into new line

mathjazz:
Finally I was able to see in person how fluent UI breaks for some people:

index

The bug is only reproducible on certain OSes and/or screen resolutions and/or screen densities.

Thanks to @flodolo, this patch should provide a fix (and also remove a duplicated CSS property).

### DIFF
--- a/pontoon/base/static/css/translate.css
+++ b/pontoon/base/static/css/translate.css
@@ -1457,12 +1457,12 @@ body > header aside p {
 }
 
 .ftl-area section .id {
-  font-weight: 300;
   border: 1px solid #5E6475;
   background: transparent;
   color: #AAAAAA;
   display: inline-block;
   font-weight: normal;
+  float: left;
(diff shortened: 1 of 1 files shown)


mathjazz:
@zbraniecki & @Jobava, could you please give it a try on [stage](https://mozilla-pontoon-staging.herokuapp.com/sl/test-fluent/demo.ftl/?string=175566) and check if the patch fixes the bug? Patch will be deployed to stage for the next 24 hours.

mathjazz:
Also @stasm, could you check [stage](https://mozilla-pontoon-staging.herokuapp.com/sl/test-fluent/demo.ftl/?string=175566), too?

stasm:
Looks good!

zbraniecki:
Yup, this is fixed for me on stage. Congrats :)

PR #868, opened 2018-02-24
Touch Bug 1440940: Add tr locale in travis env

phlax:
add collations into travis env

(...so that we can...) add some tests to ensure collation lookups work correctly with db

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,11 @@ cache:
   directories:
     - node_modules
 before_install:
+  - sudo apt-get update
+  - sudo apt-get install language-pack-tr
+  - sudo /etc/init.d/postgresql stop
+  - sudo /etc/init.d/postgresql start 9.4
+  - psql -c "CREATE COLLATION tr_tr (LOCALE = 'tr_TR.utf8');" -U postgres
   - git submodule update --init --recursive
 install:
   - pip install -U --force pip


.travis.yml
```diff
@@ -7,6 +7,12 @@ cache:
   directories:
     - node_modules
 before_install:
+  - echo 'setting up locales...'
+  - sudo apt-get update
+  - sudo apt-get install language-pack-tr
+  - sudo /etc/init.d/postgresql stop
```
jotes: is it necessary to restart postgresql? asking

phlax: yep, apparently so. There are a few places it mentions this - but see here:

https://docs.travis-ci.com/user/database-setup/#PostgreSQL-and-Locales

.travis.yml
```diff
@@ -7,6 +7,11 @@ cache:
   directories:
     - node_modules
 before_install:
+  - sudo apt-get update
+  - sudo apt-get install language-pack-tr
+  - sudo /etc/init.d/postgresql stop
+  - sudo /etc/init.d/postgresql start 9.4
+  - psql -c "CREATE COLLATION tr_tr (LOCALE = 'tr_TR.utf8');" -U postgres
```
phlax: im wondering if this should actually be in a migration, but it would cause a db without necessary locale env to fail i think (which may be a good thing).

either way i think we should probs deal with any migration separately - and potentially sort other locales at the same time

jotes: Can you create a follow-up bug to add collation for the rest of languages or could you add a comment about why `tr_TR` is the only locale added here?

After some thought I feel that support for collations could be a strict requirement and previous decision could be misleading (mea culpa).

From my perspective migrations will be executed in  3 environments:
* docker build - add those locales
* standalone installation/standalone dev setups - requires updates in docs
* existing heroku instances (e.g. prod/staging) will require some manual changes, similar to a standalone dev setup.

It's better than silently failing/broken features (e.g. invalid sorting/search for tr_TR).

Besides, lgtm -> thanks for so much work you put to make it right!

phlax: > Can you create a follow-up bug to add collation for the rest of languages

Added here:

https://bugzilla.mozilla.org/show_bug.cgi?id=1441469

PR #871, opened 2018-03-01
Fix Bug 1442281: Cache python dirs in travis

phlax:
This is a massive optimization for (cached) travis runs. Its pretty much as fast as we will get it without removing the old test suite, or speeding up tests themselves

- cache python dirs
- a workaround for travis' use of aufs which otherwise causes it to upload cache without change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,26 @@ python:
   - '2.7'
 cache:
   bundler: true
+  # cache the python dirs - this provides a very big speedup
   directories:
     - node_modules
+    - $HOME/virtualenv/python2.7.14/bin
+    - $HOME/virtualenv/python2.7.14/lib
+    - $HOME/virtualenv/python2.7/bin
 before_install:
   - git submodule update --init --recursive
+
+  # make a copy of the python libs before running tests. These are then used to restore
+  # specifically the "py" and "pytest" packages (including any pyc etc) at the end of the
+  # test run. This is necessary to prevent cache uploads with no change.
+
+  # see here for discussion - https://github.com/travis-ci/travis-ci/issues/4873
+
+  # bugzilla bug here: https://bugzilla.mozilla.org/show_bug.cgi?id=1442281
+
+  # caches may need to be cleared from time to time https://docs.travis-ci.com/user/caching#Clearing-Caches
+
+  - cp -a $HOME/virtualenv/python2.7.14/lib/python2.7/site-packages/ $HOME/py-workaround/
 install:
   - pip install -U --force pip
   - pip install --require-hashes -r requirements.txt
@@ -35,6 +51,15 @@ addons:
   apt:
     packages:
       - language-pack-tr
+
+before_cache:
+  # workaround for travis/aufs to prevent cache upload if no change
+  # if real change happens, the cache will still be updated
+
+  - pip uninstall py pytest -y
+  - pip install py==1.4.34 pytest==3.2.3
+  - pyclean $HOME/virtualenv/python2.7.14/lib/python2.7/site-packages/
+  - cp -a $HOME/py-workaround/* $HOME/virtualenv/python2.7.14/lib/python2.7/site-packages/
 
 env:
   global:


.travis.yml
```diff
@@ -31,14 +29,25 @@ script:
...
 addons:
   postgresql: "9.4"
+  apt:
+    packages:
+      - language-pack-tr
+
+before_cache:
+  # workaround for travis/aufs to prevent cache upload if no change
+  - pip uninstall py pytest -y
+  - pip install py==1.4.34 pytest==3.2.3
```
adngdb: This looks bad. First it's setting the version number in a different place (so we'll have to remember to update it here when we do in `requirements.txt`) and it doesn't require hashes, opening some security issues. Wouldn't it be better to maybe extract those two deps from the `requirements.txt` file in order to be able to install just those two, and still have only one place where we set the version and hash?

mathjazz:
Passing the review to @adngdb.

tests/env.py
```diff
@@ -10,10 +11,15 @@
...
 
     def setup(self):
+        self.setup_collations()
         for i in range(0, 2):
             self.setup_translations(i)
         self.setup_x()
 
+    def setup_collations(self):
+        cursor = connection.cursor()
+        cursor.execute("CREATE COLLATION tr_tr (LOCALE = 'tr_TR.utf8');")
```
adngdb: Here again I'm worried that we're doing the same thing in (as far as I know) 2 different places. Do you have any ideas why it doesn't work in travis without this fixture?

phlax: no, but after a lot of re/running travis what i could see was adding the collation and then dumping the collation table showed it to be present, but when you actually run the tests the collation was missing

This really was the only way i could find to get this to work in travis, and as it removes the other setup, its only done once

.travis.yml
phlax: Yep, the hardcoding requirements is a tradeoff - but imhe travis doesnt change these very often, probs ~every 6 months or so.

> Wouldn't it be better to maybe extract those two deps from the requirements.txt

Its not about our requirements - its about travis provided versions - i can test this further but i dont think this is optional if you want to save ~30s on every test run

adngdb: Huh, travis uses its own version of pytest, which is not necessarily the one we have decided upon in our `requirements.txt`???

tests/env.py
phlax: ...it also means its defo present in local test runs, so its actually a lot more reliable doing it this way

.travis.yml
phlax: it uses ours, but for this hacky kludgy, but very effective, workaround to work we need to reset to travis' version at the end of each test run

phlax:
@adngdb re-requesting review on this as it allows us all to make better use of the short amount of time we have on this planet 8/

phlax:
see here for ref about not sending cache https://travis-ci.org/mozilla/pontoon/builds/347843129#L1224

phlax:
and further ref https://github.com/travis-ci/travis-ci/issues/4873

adngdb:
@phlax the build is failing now, on a unit test. Any ideas why?

phlax:
@adngdb which fail ?

if you mean this one https://travis-ci.org/mozilla/pontoon/builds/350273685 , i think it was because of the django upgrade, altho its wierd because it passed after rebasing, and the django upgrade happens.

Either way, caning the cache and re-running fixed. IMHE this happens from time to time when there are major changes, but saving ~25% on most other test runs means its a small hardship

phlax:
sorry, i posted wrong test run - the one i posted failed because of linting

this one https://travis-ci.org/mozilla/pontoon/builds/350253068 failed because of cache, but i re-ran it

phlax:
i think this is the fastest run ive seen with this PR in effect https://travis-ci.org/mozilla/pontoon/builds/348908065

there are many others that run close to 5 mins

adngdb:
I have a few concerns with this, but nothing that cannot be solved. 

What happens when Travis updates their version of pytest? Is the build going to fail, or is it simply going to become slower again? Would it be possible to use wildcard, or anything, so that we don't have to update our travis.yml file every time they change stuff on their side? 

We've also just seen that caching can cause some builds to fail, and I don't know how you solved that one. I've rerun the build before telling you about the failure, so that's not it. How do you rebuild the cache on travis? 

This is also a very hacky thing, and a pretty obscure one at that. That combined with my previous points make me believe we need to have super solid documentation around it. I would like to see explanations of why we're doing that, and links to better context and explanations about what's happening, as well as maintenance processes.

phlax:
> What happens when Travis updates their version of pytest? Is the build going to fail, or is it simply going to become slower again?

I think the latter, but if it does fail its an easy fix. In the couple of years of using this hack i havent seen that problem, altho as said builds do fail from time to time when deps change significantly.

> We've also just seen that caching can cause some builds to fail, and I don't know how you solved that one

In this case i cleared individual caches in the travis cache UI - mostly tho we have just deleted all caches if there were any problems.

see here for docs https://docs.travis-ci.com/user/caching#Clearing-Caches

> This is also a very hacky thing, and a pretty obscure one at that.

Yep, agreed! Its very effective tho.

> That combined with my previous points make me believe we need to have super solid documentation around it. I would like to see explanations of why we're doing that, and links to better context and explanations about what's happening, as well as maintenance processes.

Cool, ill update the PR with some info from here.

phlax:
@adngdb ive added some explanation - would you mind reviewing again

PR #872, opened 2018-03-01
No bug - Run all tests and linters during make dockertest.

adngdb:
This commit adds a new run_tests file that runs all our current linters and tests, and puts the docker environment to parity with travis. This way we can easily run the entire test suite locally.


adngdb:
Done!

PR #873, opened 2018-03-01
Touch Bug 1440940: Collation test env

phlax:
- add collation in the pytest env
- remove collation from travis setup as this did not seem to work for the test db

### DIFF
--- a/tests/env.py
+++ b/tests/env.py
@@ -1,4 +1,5 @@
 from django.contrib.auth.models import Group
+from django.db import connection
 
 from pontoon.base.models import (
     Entity, Locale, Project, ProjectLocale, Resource,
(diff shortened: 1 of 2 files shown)


tests/env.py
```diff
@@ -10,10 +11,15 @@
...
 
     def setup(self):
+        self.setup_collations()
         for i in range(0, 2):
             self.setup_translations(i)
         self.setup_x()
 
+    def setup_collations(self):
+        cursor = connection.cursor()
+        cursor.execute("CREATE COLLATION tr_tr (LOCALE = 'tr_TR.utf8');")
```
adngdb: Can you add a short comment either explaining why we need this and/or adding a link to the bug?

PR #875, opened 2018-03-02
Fix bug 1440940:Add tests to prevent regression in collation lookups.

VishalCR7:
I have added the required tests for checking collations. @phlax thanks a lot!

### DIFF
--- a/tests/base/managers/entity.py
+++ b/tests/base/managers/entity.py
@@ -591,3 +591,56 @@ def test_mgr_entity_search_translation(entity_test_search):
     assert search('first translation') == [entities[0]]
     assert search('second translation') == [entities[1]]
     assert search('third translation') == [entities[2]]
+
+
+@pytest.mark.django_db
+def test_lookup_collation(resource0, locale0, translation_factory):
+    """
+    Filter translations according to collation.
+    """
+    entity = Entity.objects.create(
+        resource=resource0,
+        string="string")
+
+    batch_kwargs = sum(
+        [[dict(
+            locale=locale0,
+            entity=entity,
+            string='this is string')],
+         [dict(
+            locale=locale0,
+            entity=entity,
+            string='this is STRİNG')],
+         [dict(
+            locale=locale0,
+            entity=entity,
+            string='this is Strıng')],
+         [dict(
+            locale=locale0,
+            entity=entity,
+            string='this is StrInG')],
+         [dict(
+            locale=locale0,
+            entity=entity,
+            string='this is sTriNg')]],
+        [])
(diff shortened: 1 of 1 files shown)


tests/base/managers/entity.py
```diff
@@ -591,3 +591,56 @@ def test_mgr_entity_search_translation(entity_test_search):
...
+@pytest.mark.django_db
+def test_lookup_collation(resource0, locale0, translation_factory):
+    """
+    Filter translations according to collation.
+    """
+    entity = Entity.objects.create(
+        resource=resource0,
+        string="string")
+
+    batch_kwargs = sum(
```
phlax: i think the sum is unnecessary here and complicates the code a little - why not just use a list of dicts ?

tests/base/managers/entity.py
```diff
@@ -591,3 +591,56 @@ def test_mgr_entity_search_translation(entity_test_search):
...
+            locale=locale0,
+            entity=entity,
+            string='this is sTriNg')]],
+        [])
+    translations = translation_factory(
+        batch_kwargs=batch_kwargs)
+    assert (
+        set( Translation.objects.filter(
+            string__icontains_collate=(u'string', 'tr_tr')))
+        == {translations[0], translations[1], translations[4],})
```
phlax: would you mind using `set()` rather than `{}` for sets as itss easy to confuse with `dicts`

phlax: you could also do this...

```
batch_kwargs = [dict(entity=entity, locale=locale0, string=s) for s in [u'string 1', u'string 2']]
```

tests/base/managers/entity.py
phlax: you could also do this...

```
batch_kwargs = [dict(entity=entity, locale=locale0, string=s) for s in [u'string 1', u'string 2']]
```

tests/base/managers/entity.py
```diff
@@ -591,3 +591,37 @@ def test_mgr_entity_search_translation(entity_test_search):
...
+        string="string")
+    batch_kwargs = [dict(entity=entity, locale=locale0, string=s) 
+        for s in [u'this is string',
+            u'this is STRİNG',
+            u'this is Strıng',
+            u'this is StrInG',
+            u'this is sTriNg']]
+    translations = translation_factory(
+        batch_kwargs=batch_kwargs)
+    assert (
```
phlax: one last request 8)

would you mind adding a comment before each of these asserts explaining what its testing

phlax:
nice work @VishalCR7 !

PR #877, opened 2018-03-02
Touch Bug #1416256: String tiers resources util

phlax:
Adds a Resources util for adding/removing resources from tags according to project and project set up

it does the following

- search for linked resources
- search for not linked resources
- un/link resources

### DIFF
--- a/pontoon/tags/utils/resources.py
+++ b/pontoon/tags/utils/resources.py
@@ -1,0 +1,119 @@
+
+from pontoon.base.utils import glob_to_regex
+
+from pontoon.tags.exceptions import InvalidProjectError
+
+from .base import TagsDataTool
+
+
+class TagsResourcesTool(TagsDataTool):
+    """ Find, link and unlink Resources for Tags
+    """
+
+    filter_methods = ('locales', 'projects', 'slug', 'path')
+
+    @property
+    def data_manager(self):
+        return self.resource_manager
+
+    @property
+    def filtered_data(self):
+        return super(TagsResourcesTool, self).filtered_data.distinct()
+
+    def filter_locales(self, resources):
+        return (
+            resources.filter(translatedresources__locale__in=self.locales)
+            if self.locales
+            else resources)
+
+    def filter_path(self, resources):
+        return (
+            resources.filter(path__regex=glob_to_regex(self.path))
+            if self.path
+            else resources)
+
+    def filter_projects(self, resources):
+        return (
+            resources.filter(project__in=self.projects)
+            if self.projects
+            else resources)
+
+    def filter_slug(self, resources):
+        return (
+            resources.filter(tag__slug__regex=glob_to_regex(self.slug))
+            if self.slug
+            else resources)
+
+    def find(self, glob, include=None, exclude=None):
+        """ Find filtered resources for a glob match, and optionally
+        include/exclude resources linked to given tags
+        """
+
+        if include:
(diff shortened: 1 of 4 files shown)


pontoon/tags/utils/resources.py
```diff
@@ -0,0 +1,108 @@
+
+from pontoon.base.utils import glob_to_regex
+
+from pontoon.tags.exceptions import InvalidProjectError
+
+from .base import TagsDataTool
+
+
+class TagsResourcesTool(TagsDataTool):
```
mathjazz: A short comment explaining the purpose of this class would be helpful.

pontoon/tags/utils/resources.py
```diff
@@ -0,0 +1,108 @@
...
+    filter_methods = ('locales', 'projects', 'slug', 'path')
+
+    @property
+    def data_manager(self):
+        return self.resource_manager
+
+    def link(self, tag, glob=None, resources=None):
```
mathjazz: Please add a quick comment explaining both scenarious.

pontoon/tags/utils/resources.py
```diff
@@ -0,0 +1,108 @@
...
+        return self.locale_manager.filter(
+            pk__in=self.get(slug).values_list(
+                "translatedresources__locale").distinct())
+
+    def get_linked_projects(self, slug):
+        return self.project_manager.filter(
+            pk__in=self.get(slug).values_list(
+                "project").distinct())
+
+    def unlink(self, tag, glob=None, resources=None):
```
mathjazz: Maybe move up, next to the `link()` method?

And also, comment. :)

pontoon/tags/utils/resources.py
```diff
@@ -0,0 +1,108 @@
...
+        else:
+            resources = self.resource_manager
+        if self.projects:
+            resources = resources.filter(project__in=self.projects)
+        return resources.filter(path__regex=glob_to_regex(glob))
+
+    def get(self, tag):
+        return self.filtered_data.filter(tag__slug=tag).distinct()
+
+    def get_linkable_resources(self, slug):
```
mathjazz: Please add a comment explaining linkable means "not linked yet".

pontoon/tags/utils/resources.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def get_linked_locales(self, slug):
+        return self.locale_manager.filter(
+            pk__in=self.get(slug).values_list(
+                "translatedresources__locale").distinct())
+
+    def get_linked_projects(self, slug):
+        return self.project_manager.filter(
+            pk__in=self.get(slug).values_list(
+                "project").distinct())
```
mathjazz: There's an inconsistency between `get_linked_resources` vs. `get_linked_locales` and `get_linked_projects` in that the former returns `values` and the latter two return `values_list`.

pontoon/tags/utils/resources.py
phlax: ill move link down - they were supposed to be in alpha order, but link seems to have floated up

pontoon/tags/utils/resources.py
phlax: ill update the the latter 2 to return values

phlax: in the end i removed the latter 2 methods - they were part of the site-wide tags implementation, and are not required/used - so better remove, at least for now

PR #879, opened 2018-03-04
Upgrade to Django 1.11

VishalCR7:
This, for now, solves the broken collation search in Django-1.11 However its hardcoded for the events objects are filtered based on their string property.

To Do:
- [x]  Check if all the tests passes using Django-1.11.
- [x] Go through all the views to check if any of these breaks.
- [x] Check if sync works fine.

### DIFF
--- a/pontoon/db/__init__.py
+++ b/pontoon/db/__init__.py
@@ -30,6 +30,9 @@ class IContainsCollate(IContains):
         lhs, params = super(IContainsCollate, self).process_lhs(qn, connection)
         if self.collation:
             lhs = lhs.replace('::text', '::text COLLATE "{}"'.format(self.collation))
+            if '::text' not in lhs:
+                lhs = lhs.replace(
+                    '."string"', '."string"::text COLLATE "{}"'.format(self.collation))
         return lhs, params
 
     def get_rhs_op(self, connection, rhs):
--- a/pontoon/db/__init__.py
+++ b/pontoon/db/__init__.py
@@ -32,7 +32,8 @@ class IContainsCollate(IContains):
             lhs = lhs.replace('::text', '::text COLLATE "{}"'.format(self.collation))
             if '::text' not in lhs:
                 lhs = lhs.replace(
-                    '."string"', '."string"::text COLLATE "{}"'.format(self.collation))
+                    '."{}"'.format(self.lhs.target.column), '."{}"::text COLLATE "{}"'.format(
+                        self.lhs.target.column, self.collation))
         return lhs, params
 
     def get_rhs_op(self, connection, rhs):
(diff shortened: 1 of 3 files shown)


pontoon/db/__init__.py
```diff
@@ -30,6 +30,9 @@ def process_lhs(self, qn, connection):
         lhs, params = super(IContainsCollate, self).process_lhs(qn, connection)
         if self.collation:
             lhs = lhs.replace('::text', '::text COLLATE "{}"'.format(self.collation))
+            if '::text' not in lhs:
+                lhs = lhs.replace(
+                    '."string"', '."string"::text COLLATE "{}"'.format(self.collation))
```
jotes: I'll look into that issue today, because I don't have enough context to decide if We should leave this hard-coded field.

VishalCR7: @jotes its used only once in the whole code i.e. in /base/models.py for translation 'string' query. So if its going to be used only there or further only for string query in the future its safe i guess

VishalCR7:
@mathjazz @jotes @phlax I have tried to remove the hardcoded 'string' and and also added the test to check if it works fine with different fields. I tested it using django1.11 locally and compared sql queries and it looks good to me. Can you please take a look

phlax:
really nice work @VishalCR7 - perhaps we should extend the tests to cover other fields etc

VishalCR7:
@phlax I have extended the test to filter comment field of entity

phlax:
wow, i should have checked 8/ - you were already one step ahead!

pontoon/db/__init__.py
jotes: Yes, I'm aware where it's used in Pontoon. However, I don't know what has been changed in Django 1.11

phlax: @jotes for whatever reason the query no longer had `::text` so the collation mangling failed

the latest update to this PR seems to be good i think

pontoon/db/__init__.py
```diff
@@ -30,6 +30,10 @@ def process_lhs(self, qn, connection):
         lhs, params = super(IContainsCollate, self).process_lhs(qn, connection)
         if self.collation:
             lhs = lhs.replace('::text', '::text COLLATE "{}"'.format(self.collation))
+            if '::text' not in lhs:
```
jotes: @VishalCR7 Hey, I tested this locally and it works - a nice work Vishal!

The only thing I'm wondering if We should leave this hard-coded  "base.string" field. Currently, there's only one place that would be affected by this hack, but the whole goal of this class was to reuse this particular behavior in future components.

So, I experimented myself and got something like this:
https://gist.github.com/jotes/a3ee091c4f8d96a9d92d2067e110754d  This allows to handle basically any field in different models etc.
Anyway, I'm posting this only as an optional advice - feel free to express your opinion/ask questions.
Still, The current PR definitely solves the problem.

If that's okay, I'll leave the final decision/merge to @phlax.

phlax: thanks for review @jotes - my thinking is that if we can reliably fix this for any field, then lets do that, however we do it.

Looking at posted gist, its probs more elegant, but i have an aversion to regexes unless absolutely necessary as they are hard to grok. I would be happy with either solution (if tests pass!)

phlax:
Landed in  1c7a491 - great work @VishalCR7 - thanks

PR #888, opened 2018-03-09
Fix bug 1219908: Add split search query for translation search.

VishalCR7:
This PR is regarding the bug : https://bugzilla.mozilla.org/show_bug.cgi?id=1219908.

Right now I have added the split lookup and search according to this : https://pastebin.mozilla.org/9079558

The sql query for the the most recent commit looks like: https://pastebin.com/1UCFFjug

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -8,7 +8,7 @@ import math
 import operator
 import os.path
 import re
-import shlex
+# import shlex
 
 from collections import defaultdict
 from dirtyfields import DirtyFieldsMixin
@@ -2214,10 +2214,23 @@ class Entity(DirtyFieldsMixin, models.Model):
             entities = entities.filter(Q(*post_filters))
 
         # Filter by search parameters
-        if search:
+        if search and search != '"':
             # https://docs.djangoproject.com/en/dev/topics/db/queries/#spanning-multi-valued-relationships
-            search_list = shlex.split(search)
-            search_list = [s.strip() for s in search_list]
+            search = search + '"' if(search.count('"') % 2) else search
+            search_list = re.findall('([^\"]\\S*|\".+?\")\\s*', search)
+            search_list = [s.replace('"', '').strip("'").strip() for s in search_list]
+            """
+            Alternate way to create the above list.
+            def newSplit(value):
+                lex = shlex.shlex(value)
+                lex.quotes = '"'
+                lex.whitespace_split = True
+                lex.commenters = ''
+                return list(lex)
+            search = search+'"' if(search.count('"')%2) else search
+            search_list = newSplit(search)
+            search_list = [s.strip("'").strip(' ').strip('"') for s in search_list]
+            """
             search_query_list = [(s, locale.db_collation) for s in search_list]
             query = reduce(operator.and_, (
                 Q(translation__string__icontains_collate=search_query) &
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -2215,10 +2215,12 @@ class Entity(DirtyFieldsMixin, models.Model):
         # Filter by search parameters
         if search and search != '"':
             # https://docs.djangoproject.com/en/dev/topics/db/queries/#spanning-multi-valued-relationships
+            # Add a double quote at the end of search if closing quote not present
             search = (
                 '%s"' % search
                 if search.count('"') % 2
                 else search)
+            # Split search string on spaces except if between quotes.
             search_list = re.findall('([^\"]\\S*|\".+?\")\\s*', search)
             search_list = [s.strip('"').strip("'").strip() for s in search_list]
             search_query_list = [(s, locale.db_collation) for s in search_list]
@@ -2238,7 +2240,7 @@ class Entity(DirtyFieldsMixin, models.Model):
                   )
                 for search in search_list)
             )
-            entity_matches = (entities.filter(query).values_list('id', flat=True))
+            entity_matches = entities.filter(query).values_list('id', flat=True)
             entities = Entity.objects.filter(
                 pk__in=set(list(translation_matches) + list(entity_matches))
             )
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -2213,34 +2213,45 @@ class Entity(DirtyFieldsMixin, models.Model):
             entities = entities.filter(Q(*post_filters))
 
         # Filter by search parameters
-        if search and search != '"':
-            # https://docs.djangoproject.com/en/dev/topics/db/queries/#spanning-multi-valued-relationships
+        if search:
             # Add a double quote at the end of search if closing quote not present
             search = (
                 '%s"' % search
                 if search.count('"') % 2
                 else search)
+
             # Split search string on spaces except if between quotes.
             search_list = re.findall('([^\"]\\S*|\".+?\")\\s*', search)
             search_list = [s.strip('"').strip("'").strip() for s in search_list]
+
+            # Search for `"` and `'` when entered as search terms
+            if search == '""' and not search_list:
+                search_list = ['"']
+
+            if search == "'" and not search_list:
+                search_list = ["'"]
+
             search_query_list = [(s, locale.db_collation) for s in search_list]
-            query = reduce(operator.and_, (
+
+            translation_filters = (
                 Q(translation__string__icontains_collate=search_query) &
-                Q(translation__locale=locale
-                  )
+                Q(translation__locale=locale)
                 for search_query in search_query_list)
-            )
-            translation_matches = entities.filter(query).values_list('id', flat=True)
-            
-            query = reduce(operator.and_, (
+            entity_filters = (
                 Q(string__icontains=search) |
                 Q(string_plural__icontains=search) |
                 Q(comment__icontains=search) |
-                Q(key__icontains=search
-                  )
+                Q(key__icontains=search)
                 for search in search_list)
-            )
-            entity_matches = entities.filter(query).values_list('id', flat=True)
+
+            # Combine all generated filters with an AND operator.
+            # `operator.and_` is the '&' Python operator, which turns into a logical AND
+            # when used between django ORM query objects.
+            translation_query = reduce(operator.and_, translation_filters)
+            entity_query = reduce(operator.and_, entity_filters)
+
+            translation_matches = entities.filter(translation_query).values_list('id', flat=True)
+            entity_matches = entities.filter(entity_query).values_list('id', flat=True)
             entities = Entity.objects.filter(
                 pk__in=set(list(translation_matches) + list(entity_matches))
             )
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -2215,25 +2215,30 @@ class Entity(DirtyFieldsMixin, models.Model):
         # Filter by search parameters
         if search:
             # https://docs.djangoproject.com/en/dev/topics/db/queries/#spanning-multi-valued-relationships
-            search_query = (search, locale.db_collation)
-            translation_matches = (
-                entities.filter(
-                    translation__string__icontains_collate=search_query,
-                    translation__locale=locale,
+            translation_matches_list = []
+            entity_matches_list = []
+            for search in search.split():
+                search_query = (search, locale.db_collation)
+                translation_matches = (
+                    entities.filter(
+                        translation__string__icontains_collate=search_query,
+                        translation__locale=locale,
+                    )
+                    .values_list('id', flat=True)
                 )
-                .values_list('id', flat=True)
-            )
-            entity_matches = (
-                entities.filter(
-                    Q(string__icontains=search) |
-                    Q(string_plural__icontains=search) |
-                    Q(comment__icontains=search) |
-                    Q(key__icontains=search)
+                translation_matches_list += list(translation_matches)
+                entity_matches = (
+                    entities.filter(
+                        Q(string__icontains=search) |
+                        Q(string_plural__icontains=search) |
+                        Q(comment__icontains=search) |
+                        Q(key__icontains=search)
+                    )
+                    .values_list('id', flat=True)
                 )
-                .values_list('id', flat=True)
-            )
+                entity_matches_list += list(entity_matches)
             entities = Entity.objects.filter(
-                pk__in=set(list(translation_matches) + list(entity_matches))
+                pk__in=set(translation_matches_list + entity_matches_list)
             )
 
         if exclude_entities:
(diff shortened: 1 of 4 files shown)


VishalCR7:
@mathjazz  @phlax : please review

phlax:
@VishalCR7 debugging the sql looks good now - ie it uses a single query. Ive not seen the `operator.and_` construct before and find it a little hard to grok, but i wouldnt block on it. 

My preferred construct for this pattern is like this...

```
qs = Model.objects.all()
_qs = qs.none()

for search_filter in search_filters:
    _qs &= qs.filter(...search_filter...)


```

pontoon/base/static/js/translate.js
```diff
@@ -3857,11 +3857,24 @@ var Pontoon = (function (my) {
 
       item.unmark();
       if (searchQuery) {
-        item.mark(searchQuery, {
+        searchQuery = ((searchQuery.split('"').length-1)%2)? searchQuery+'"':searchQuery;
```
phlax: can we add some space between the `operators`

pontoon/base/models.py
```diff
@@ -2206,25 +2206,29 @@ def for_project_locale(
...
             )
-            entity_matches = (
-                entities.filter(
-                    Q(string__icontains=search) |
-                    Q(string_plural__icontains=search) |
-                    Q(comment__icontains=search) |
-                    Q(key__icontains=search)
-                )
-                .values_list('id', flat=True)
+            translation_matches = (entities.filter(query).values_list('id', flat=True))
```
phlax: the brackets around this qs are not needed

pontoon/base/models.py
```diff
@@ -2206,25 +2206,29 @@ def for_project_locale(
...
+        if search and search != '"':
             # https://docs.djangoproject.com/en/dev/topics/db/queries/#spanning-multi-valued-relationships
-            search_query = (search, locale.db_collation)
-            translation_matches = (
-                entities.filter(
-                    translation__string__icontains_collate=search_query,
-                    translation__locale=locale,
-                )
-                .values_list('id', flat=True)
+            search = search + '"' if(search.count('"') % 2) else search
```
phlax: better use interpolation than `search + '"'`

pontoon/base/models.py
```diff
@@ -2206,25 +2206,32 @@ def for_project_locale(
...
+        if search and search != '"':
             # https://docs.djangoproject.com/en/dev/topics/db/queries/#spanning-multi-valued-relationships
-            search_query = (search, locale.db_collation)
-            translation_matches = (
-                entities.filter(
-                    translation__string__icontains_collate=search_query,
-                    translation__locale=locale,
-                )
-                .values_list('id', flat=True)
+            search = (
```
phlax: can we add a comment explaining why this was necessary

pontoon/base/models.py
```diff
@@ -2206,25 +2206,32 @@ def for_project_locale(
...
+                '%s"' % search
+                if search.count('"') % 2
+                else search)
+            search_list = re.findall('([^\"]\\S*|\".+?\")\\s*', search)
```
phlax: probs best to add a comment explaining what this regex does

pontoon/base/models.py
```diff
@@ -2206,25 +2206,32 @@ def for_project_locale(
...
+            
+            query = reduce(operator.and_, (
+                Q(string__icontains=search) |
+                Q(string_plural__icontains=search) |
+                Q(comment__icontains=search) |
+                Q(key__icontains=search
+                  )
+                for search in search_list)
             )
+            entity_matches = (entities.filter(query).values_list('id', flat=True))
```
phlax: the brackets around this qs are also unnecessary

mathjazz:
Great work @VishalCR7!

I've noticed a regression - letter `š` is not highlighted:
screen shot 2018-03-12 at 20 56 09

phlax:
really nice work @VishalCR7 !

pontoon/base/models.py
```diff
@@ -2206,25 +2206,34 @@ def for_project_locale(
...
+                if search.count('"') % 2
+                else search)
+            # Split search string on spaces except if between quotes.
+            search_list = re.findall('([^\"]\\S*|\".+?\")\\s*', search)
+            search_list = [s.strip('"').strip("'").strip() for s in search_list]
+            search_query_list = [(s, locale.db_collation) for s in search_list]
+            query = reduce(operator.and_, (
+                Q(translation__string__icontains_collate=search_query) &
+                Q(translation__locale=locale
+                  )
```
mathjazz: Nit: Why break the ```?``` into a new line instead of doing this?
```
Q(translation__locale=locale)
```

pontoon/base/models.py
```diff
@@ -2206,25 +2206,34 @@ def for_project_locale(
...
-                )
-                .values_list('id', flat=True)
+            translation_matches = entities.filter(query).values_list('id', flat=True)
+            
+            query = reduce(operator.and_, (
+                Q(string__icontains=search) |
+                Q(string_plural__icontains=search) |
+                Q(comment__icontains=search) |
+                Q(key__icontains=search
+                  )
```
mathjazz: Nit: Why break the ```?``` into a new line instead of doing this?
```
Q(key__icontains=search)
```

pontoon/base/models.py
```diff
@@ -2206,25 +2206,34 @@ def for_project_locale(
...
+            # Add a double quote at the end of search if closing quote not present
+            search = (
+                '%s"' % search
+                if search.count('"') % 2
+                else search)
+            # Split search string on spaces except if between quotes.
+            search_list = re.findall('([^\"]\\S*|\".+?\")\\s*', search)
+            search_list = [s.strip('"').strip("'").strip() for s in search_list]
+            search_query_list = [(s, locale.db_collation) for s in search_list]
+            query = reduce(operator.and_, (
```
mathjazz: Can we replace this with https://github.com/mozilla/pontoon/pull/888#issuecomment-372222015? I find the proposed syntax much easier to read.

VishalCR7: will do. However, operator is used in one more place I the same file but with some helpful comments. would it work if I add similar comments?

mathjazz: You could do that, too.

And putting the second argument of the `reduce()` in a variable would help readability as well:
```python
reduce(operator.and_, something)
```

flodolo:
One issue is that a search for `"` will give you all strings, so you can't really search for that character. Maybe it should at least work if escaped, i.e. `\"`?

mathjazz:
> One issue is that a search for " will give you all strings, so you can't really search for that character. Maybe it should at least work if escaped, i.e. \"?

[It](https://mozilla-pontoon-staging.herokuapp.com/sl/firefox/all-resources/?search=%22) works fine for me. We just fixed that yesterday. Did you test on stage today?

flodolo:
> It works fine for me. We just fixed that yesterday. Did you test on stage today?

That's returning all strings, not just those with a double quote in it.

mathjazz:
It's not. We search strings (original & translated), comments and keys. We'd need an additional syntax and/or UI element to define where exactly to search.

flodolo:
> It's not. We search strings (original & translated), comments and keys. We'd need an additional syntax and/or UI element to define where exactly to search.

Argh, forgot about comments…

TheoChevalier:
Thanks @mathjazz !
I wonder if we will have to update the `<edit in Pontoon>` links on Transvision where we’re performing a search on the full string, and wrap them into double quotes. So far it seems to work even with strings containing quotes, they are just not highlighted anymore. https://transvision.mozfr.org/?recherche=%3Ca&repo=mozilla_org&sourcelocale=en-US&locale=fr&search_type=strings (Checked with those links, and replaced the domain with staging)
There is probably a slightly lower performance cost when wrapping search string with double quotes?

The only regression I can find is with nbsp (how much of a PITA am I with those?):

https://mozilla-pontoon-staging.herokuapp.com/fr/firefox/all-resources/?search=%C2%A0!&string=76048
https://pontoon.mozilla.org/fr/firefox/all-resources/?search=%C2%A0!&string=75350

On staging, you can see that the first string (and many others) shouldn’t be returned

TheoChevalier:
Note that it doesn’t work either with double quotes https://mozilla-pontoon-staging.herokuapp.com/fr/firefox/all-resources/?search=%22%C2%A0!%22&string=76048

MikkCZ:
The functionality looks good to me, thought I haven't tested any special characters queries etc. Still I am looking forward for the ultimate goal to have the all projects search in the Transvision-style.

phlax:
@TheoChevalier i believe @VishalCR7 has fixed the issue with spaces in the double quotes like `" !"` - latest code is on staging.

re the other problem with quotes marks as in `<a href="` is harder to fix. 

Currently this will return an exact match: 

https://pontoon.mozilla.org/fr/mozillaorg/all-resources/?search=%3Ca+href%3D%22&string=7867

with the PR this would match 3 different strings:

https://mozilla-pontoon-staging.herokuapp.com/fr/mozillaorg/all-resources/?search=%3Ca+href%3D%22%25(&string=7867

I dont see an easy way to restore the current behaviour *and* use the `"` to indicate phrase matching. 

One possible solution, at a cost of some complexity, is to allow users to escape the double quotes.

Searches could then distinguish between  `<a href=\"foo\"`, `"<a href=\"foo\""` and `<a href="foo">`

mathjazz:
@MikkCZ Would https://github.com/mozilla/pontoon/pull/885 take us closer to that goal?

MikkCZ:
It won't be search first experience, like something to replace the Machinery page could be (I wonder if we shouldn't remove the Machinery page completely now), but combined with this PR, it will definitely serve the purpose.

mathjazz:
> One possible solution, at a cost of some complexity, is to allow users to escape the double quotes.
> 
> Searches could then distinguish between <a href=\"foo\", "<a href=\"foo\"" and <a href="foo">

I think this is the right way to go here. @VishalCR7 Any chance you could take a stab at that?

VishalCR7:
@mathjazz: Sure! Though it can take some time

phlax:
thry this @VishalCR7 

```
# -*- coding: utf-8 -*-                                                                                                                                       

import re

UNUSABLE = u'☠️'

def mangler(search):
    return [
        x.strip('"').strip("'").replace(UNUSABLE, '"')
        for x
        in re.findall(
            '([^\"]\\S*|\".+?\")\\s*',
            search.replace('\\"', UNUSABLE))]

search = u'foo "s😐mething \\"and 😐 asdf\\" else" bar'

print u'IN: %s' % search
print
print 'OUT:'
for res in mangler(search): print res

```

pontoon/base/models.py
```diff
@@ -2207,24 +2208,42 @@ def for_project_locale(
...
-                entities.filter(
-                    Q(string__icontains=search) |
-                    Q(string_plural__icontains=search) |
-                    Q(comment__icontains=search) |
-                    Q(key__icontains=search)
-                )
-                .values_list('id', flat=True)
-            )
+            # Split search string on spaces except if between quotes.
+            UNUSABLE = u'☠'
```
phlax: probs best to declare constants at the top of the file

pontoon/base/models.py
```diff
@@ -43,6 +44,7 @@
 
 log = logging.getLogger(__name__)
 
+UNUSABLE_SEARCH_CHAR = u'☠'
```
phlax: could you add an extra blank line under the constant definition

tests/base/managers/entity.py
```diff
@@ -570,12 +570,23 @@ def test_mgr_entity_search_entities(entity_test_search):
     entities, search = entity_test_search
 
     assert search(u'e') == entities
-    assert search(u'entity string') == entities
+    assert search(u'entity string') == [entities[i] for i in [0, 1, 2, 3, 5, 6]]
```
phlax: i think to make it clearer what the entity indeces translate to we probs need to add a comment with the strings

pontoon/base/models.py
```diff
@@ -2207,24 +2210,41 @@ def for_project_locale(
...
-            entity_matches = (
-                entities.filter(
-                    Q(string__icontains=search) |
-                    Q(string_plural__icontains=search) |
-                    Q(comment__icontains=search) |
-                    Q(key__icontains=search)
-                )
-                .values_list('id', flat=True)
-            )
+            # Split search string on spaces except if between quotes.
```
phlax: ...if between *unescaped* quotes...

phlax:
@mathjazz @TheoChevalier seems like @VishalCR7 has fixed the issues highlighted - you can now escape quotes in the search - eg

https://mozilla-pontoon-staging.herokuapp.com/fr/mozillaorg/all-resources/?search=%22%3Ca+href%3D%5C%22%25(%22&string=7867

@VishalCR7 again great work!

pontoon/base/models.py
```diff
@@ -2217,7 +2217,7 @@ def for_project_locale(
 
         # Filter by search parameters
         if search:
-            # Split search string on spaces except if between quotes.
+            # Split search string on spaces except if between unescaped quotes.
```
phlax: hmm - im now wondering if this is correct (sorry!) - maybe `non-escaped`

pontoon/base/models.py
```diff
@@ -2214,24 +2217,41 @@ def for_project_locale(
...
+            if search == '"' and not search_list:
+                search_list = ['"']
+
+            if search == "'" and not search_list:
+                search_list = ["'"]
+
+            search_query_list = [(s, locale.db_collation) for s in search_list]
+
+            translation_filters = (
+                Q(translation__string__icontains_collate=search_query) &
```
phlax: i generally try to avoid code style nits that are pep8 friendly, however...

according to pep8 you *should* put operators at end of line - but i really think this is wrong (a lot!)

my pref is to have the operators at beginning - like (using next lines as example) ...

```

            entity_filters = (
+                Q(string__icontains=search)
+                | Q(string_plural__icontains=search)
+                | Q(comment__icontains=search)
+                | Q(key__icontains=search)
+                for search in search_list)
+
````

phlax:
> 2. Searching for <a href="%(url)s"> and <a href=\"%(url)s\"> on /fr/mozillaorg/all-resources/ gives me the same results, but matches in the latter are not marked up.

this seems to work for me - they are slightly different searches in that the second is going to match 2 strings instead of 1 - ie `<a` and `href=\"%(url)s\">` - but it seems to be returning correct results and highlighting correctly

https://mozilla-pontoon-staging.herokuapp.com/fr/mozillaorg/all-resources/?search=%3Ca+href%3D%5C%22%25(url)s%5C%22%3E&string=7867

VishalCR7:
@mathjazz @phlax  thanks a lot! This one was great to work on!

PR #889, opened 2018-03-09
Fix bug 1442201: Update Fluent libraries to 0.6.4

mathjazz:
As @Pike points out, we should be good for bug 1442201 by upgrading to 0.6.4.


mathjazz:
My suspicion is we'll need to land his patch together with the fix for [bug 1441020](https://bugzilla.mozilla.org/show_bug.cgi?id=1441020), otherwise we'll be creating more duplicate translations.

mathjazz:
Landing blocked by #893.

PR #893, opened 2018-03-12
Bug 1441020: Migrate FTL translations to latest syntax and serializer

mathjazz:
In #889 we're introducing a new Fluent serializer, which serializes FTL messages slightly differently than the previously used version. To prevent duplicate translations from being created on save and approval, we're migrating all existing FTL translations to the new serialized output.

The migration will also affect translations using the old (0.4) syntax, so this PR also fixes [bug 1441020](https://bugzilla.mozilla.org/show_bug.cgi?id=1441020).

Note: The PR is branched off of #889. The only change is the data migration in https://github.com/mozilla/pontoon/commit/4794d4e52afaec1b06622bf55a1a872b84803dd5.


mathjazz:
Thanks!

Filed [bug 1444848](https://bugzilla.mozilla.org/show_bug.cgi?id=1444848) for `BaseNode.equals` comparison.

PR #895, opened 2018-03-12
Fix bug 1405256: Fix migration

mathjazz:
Delete TranslationMemoryEntry instances first, because the tms QuerySet gets empty after the Translation instances are deleted.


mathjazz:
r=me

PR #896, opened 2018-03-13
Fix bug 1389820: Break resources into new lines if needed

mathjazz:
This fixes the bug as shown in the screenshot:

screen shot 2018-03-13 at 16 05 17


mathjazz:
I noticed that, too, but I'm not sure how to fix it easily. I'll merge and we can followup if needed.

PR #899, opened 2018-03-15
Fix bug 1445435: Extend access key candidate generators

mathjazz:
FTL editor uses a special [UI for localizing access keys](https://www.youtube.com/watch?v=hnJoWodS3F8). Candidates for access keys can be selected from a list that is generated from the input in surrounding fields. Currently the only such field is the first textarea element of the first attribute of the message.

After this patch, we generate access key candidates from all textarea elements belonging to the `label` attribute. If such attribute doesn't exist, we generate candidates from all textarea elements of the value.

See examples in the [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1445435) for test cases (which should all work now, because the patch is deployed to stage). 

Also included in the patch:
- Exclude space as a possible accesskey candidate
- Add more accurate current accesskey detection
- Make access keys case sensitive

### DIFF
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -902,21 +902,45 @@ $(function () {
     $('#ftl-area .attributes ul:first').append($('#ftl-area .attributes .template li').clone());
   });
 
-  // Generate access key list
-  $('#ftl-area .attributes').on('keyup', 'textarea:first', function () {
-    var active = $('.accesskeys').find('.active').html();
-    var unique = $(this).val()
-      .toUpperCase()
-      .split('')
+  // Generate access key candidates
+  $('#ftl-area').on('keyup', 'textarea', function () {
+    var accessKeyExists = $('#ftl-area .attributes [data-id="accesskey"]').length;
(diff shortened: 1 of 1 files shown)


flodolo:
Since we're touching this UI: why are we converting candidates to uppercase?

Access keys are [case sensitive](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XUL/XUL_Accesskey_FAQ_and_Policies). Using **bold** to highlight the underlined character, since GitHub sanitizes everything else:
* `Test` with accesskey `T`: **T**est
* `Test` with accesskey `t`: Tes**t**
* `test` with accesskey `T`: **t**est 

The last example works because the logic tries to fall back to the alternative case.

mathjazz:
Thanks for explaining this! To clarify, the key combination stays the same, it's just a question of which "T" gets underlined?

Anyhow, I pushed the fix and deployed it to stage.

flodolo:
> To clarify, the key combination stays the same, it's just a question of which "T" gets underlined?

Yes, if there are two you can pick with which one you want to underline between uppercase and lowercase.

PR #900, opened 2018-03-16
Fix Bug 1446295: Fix py linters

phlax:
- updates pylama and deps
- moves pylama requirement to requirements-test.txt
- switch from pep8 to pycodestyle
- ...fix uncovered linting issues


phlax:
> Could you update the install docs

yep

PR #901, opened 2018-03-16
Touch Bug #1416256: String tiers admin form

phlax:
Adds an admin form for un/linking resources to/from tags

~~just tidying tests...~~

used in #802

### DIFF
--- a/pontoon/tags/admin/forms.py
+++ b/pontoon/tags/admin/forms.py
@@ -1,0 +1,110 @@
+
+from django import forms
+from django.utils.functional import cached_property
+
+from pontoon.base.models import Resource
+from pontoon.base.utils import glob_to_regex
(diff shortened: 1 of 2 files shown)


pontoon/tags/admin/forms.py
```diff
@@ -0,0 +1,110 @@
...
+                 else 'unlink'))
+
+    @property
+    def resources(self):
+        return (
+            self.tag_tool.linkable_resources
+            if self.cleaned_data['type'] == 'nonassoc'
+            else self.tag_tool.linked_resources)
+
+    def _clean_paths_for_select(self):
```
mathjazz: You mean `for_search`?

phlax: its kinda `for_select` as the user is selecting these paths for submission

PR #902, opened 2018-03-16
Touch #1416256: String tiers admin widget

phlax:
Just the react admin widget and tests for adding/removing resources from tags

also some cleanup/fixes for checkbox widget

Broken out from #815 

Used in #802

### DIFF
--- a/pontoon/static/js/tags/admin/button.js
+++ b/pontoon/static/js/tags/admin/button.js
@@ -1,0 +1,47 @@
+
+import React from 'react';
+
+import {TagResourceManager} from './manager';
+
+
+export default class TagResourcesButton extends React.Component {
+    state = {open: false};
+
+    handleClick = (evt) => {
+        evt.preventDefault();
+        this.setState((prevState) => ({open: !prevState.open}));
+    }
+
+    renderButton () {
+        const {open} = this.state;
+        let message = "Manage resources for this tag";
+        if (open) {
+            message = "Hide the resource manager for this tag";
+        }
+        return (
+              <button onClick={this.handleClick}>
+                {message}
+              </button>);
+    }
+
+    renderResourceManager () {
+        const {open} = this.state;
+        if (!open) {
+            return '';
+        }
+        return (
+            <TagResourceManager
+               tag={this.props.tag}
+               api={this.props.api}
+               requestMethod="post"
+               project={this.props.project} />);
+    }
+
+    render () {
+        return (
+            <div>
+              {this.renderButton()}
+              {this.renderResourceManager()}
+            </div>);
+    }
+}
--- a/pontoon/static/js/tags/admin/manager.js
+++ b/pontoon/static/js/tags/admin/manager.js
@@ -1,0 +1,53 @@
+
+import React from 'react';
+
+import TagResourceSearch from './search';
+import TagResourceTable from './table';
+import {ErrorList} from 'widgets/errors';
+
+import "./tag-resources.css";
+
+import {dataManager} from 'utils/data';
+
+
+export class TagResourceManagerWidget extends React.Component {
+    state = {type: 'assoc', search: "*"};
+
+    componentDidMount () {
+        this.props.refreshData({...this.state});
+    }
+
+    componentDidUpdate (prevProps, prevState) {
+        if (prevState !== this.state) {
+            this.props.refreshData({...this.state});
+        }
+    }
+
+    handleSearchChange = (change) => {
+        this.setState(change);
+    }
+
+    handleSubmit = (checked) => {
+        return this.props.handleSubmit(
+            Object.assign({}, this.state, checked));
+    }
+
+    render () {
+        const {data, errors} = this.props;
+        const {type} = this.state;
+        return (
+            <div className="tag-resource-widget">
+              <TagResourceSearch
+                 handleSearchChange={this.handleSearchChange} />
+              <ErrorList
+                 errors={errors || {}} />
+              <TagResourceTable
+                 data={data || []}
+                 type={type}
+                 handleSubmit={this.handleSubmit}  />
+            </div>);
+    }
+}
+
+
+export const TagResourceManager = dataManager(TagResourceManagerWidget);
--- a/pontoon/static/js/tags/admin/search.js
+++ b/pontoon/static/js/tags/admin/search.js
@@ -1,0 +1,43 @@
+
+import React from 'react';
+
+import {Columns} from 'widgets/columns';
+
+
+export default class TagResourceSearch extends React.PureComponent {
+
+    get columns () {
+        return [[this.renderSearchInput(), 3],
+                [this.renderSearchSelect(), 2]];
+    }
+
+    handleChange = (evt) => {
+        return this.props.handleSearchChange(
+            {[evt.target.name]: evt.target.value});
+    }
+
+    renderSearchInput () {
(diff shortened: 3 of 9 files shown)


pontoon/static/js/tags/admin/button.js
```diff
@@ -0,0 +1,51 @@
...
+import React from 'react';
+
+import {TagResourceManager} from './manager';
+
+
+export default class TagResourcesButton extends React.Component {
+
+    constructor (props) {
+        super(props);
+        this.state = {open: false};
```
adngdb: Since you've enabled it, you can use an attribute for this instead, and avoid the constructor altogether. 

```javascript
export default class TagResourcesButton extends React.Component {
    state = {
        open: false,
    }

    ...
}
```

pontoon/static/js/tags/admin/button.js
```diff
@@ -0,0 +1,51 @@
...
+               requestMethod="post"
+               project={this.props.project} />);
+    }
+
+    render () {
+        return (
+            <div>
+              {this.renderButton()}
+              {this.renderResourceManager()}
+            </div>);
```
adngdb: I don't understand the indentation you use in your `render` functions. I see some 6-space, 2-space and even 3-space in this file alone. Wouldn't it be much simpler if you used 4-space everywhere?

pontoon/static/js/tags/admin/manager.js
```diff
@@ -0,0 +1,57 @@
...
+import "./tag-resources.css";
+
+import {dataManager} from 'utils/data';
+
+
+export class TagResourceManagerWidget extends React.Component {
+
+    constructor (props) {
+        super(props);
+        this.state = {type: 'assoc', search: "*"};
```
adngdb: You can use an attribute here as well.

pontoon/static/js/tags/admin/manager.js
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    componentDidUpdate (prevProps, prevState) {
+        if (prevState !== this.state) {
+            this.props.refreshData({...this.state});
+        }
+    }
+
+    handleSearchChange = (change) => {
+        return this.setState(change);
```
adngdb: `setState` returns nothing, this `return` is useless.

pontoon/static/js/tags/admin/manager.js
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    handleSubmit = (checked) => {
+        return this.props.handleSubmit(
+            Object.assign({}, {...this.state}, checked));
```
adngdb: No need to deconstruct `state` here: `Object.assign({}, this.state, checked)`.

pontoon/static/js/tags/admin/search.js
```diff
@@ -0,0 +1,47 @@
...
+    handleChange = (evt) => {
+        switch (evt.target.nodeName) {
+        case 'INPUT':
+            return this.props.handleSearchChange(
+                {search: evt.target.value});
+        case 'SELECT':
+            return this.props.handleSearchChange(
+                {type: evt.target.value});
+        }
+    }
```
adngdb: You can make this more generic and more explicit at the same time by adding a `name` attribute to your elements. See the official documentation: https://reactjs.org/docs/forms.html#handling-multiple-inputs

```javascript
handleInputChange(event) {
    const target = event.target;
    this.props.handleSearchChange({
        [target.name]: target.value,
    });
}
```

pontoon/static/js/tags/admin/table.js
```diff
@@ -0,0 +1,42 @@
...
+            Cell: this.renderResource};
+    }
+
+    get columns () {
+        return [this.columnResource];
+    }
+
+    get submitClass () {
+        return "tag-resources-associate";
+    }
```
adngdb: Why are these three getters? It would be much simpler to make them attributes.

pontoon/static/js/widgets/tables/checkbox.js
```diff
@@ -12,14 +12,6 @@ export default class CheckboxTable extends React.Component {
...
         this.visible = [];
         this.state = {checked: new Set()};
-        this.handleCheckboxClick = this.handleCheckboxClick.bind(this);
-        this.handleSelectAll = this.handleSelectAll.bind(this);
-        this.handleSubmit = this.handleSubmit.bind(this);
-        this.handleTableChange = this.handleTableChange.bind(this);
-        this.handleTableSortChange = this.handleTableSortChange.bind(this);
-        this.handleTableResize = this.handleTableResize.bind(this);
-        this.renderCheckbox = this.renderCheckbox.bind(this);
-        this.renderSelectAllCheckbox = this.renderSelectAllCheckbox.bind(this);
```
adngdb: :sun_with_face:

pontoon/static/js/tags/admin/button.js
phlax: nice!

pontoon/static/js/tags/admin/button.js
phlax: im using `rjsx-mode` in emacs which defaults to 4 space on code, 2 on jsx - which i like tbh

pontoon/static/js/tags/admin/table.js
phlax: yep, ill update

pontoon/static/js/tags/admin/button.js
adngdb: Fair enough, but why use 3-space for attributes in JSX?

phlax: well - again - i was going with the defaults for the mode, which i quite like

pontoon/static/js/tags/admin/table.js
phlax: i think only this one can be moved to an attr as the others use `this`

phlax: re js `Class.attrs` im remming the thing that is a pita with these - you cant do python-like...

```
Class A(object):
     foo = "FOO"

class B(A):
    
     @property
     def foo(self):
          return "MANGLED FOO"

```

altho im not clear whether this is an es6+ problem, or whether its just the way webpack is currently set

pontoon/static/js/tags/admin/manager.js
julen: If object spread is available, [`Object.assign()` is not needed either](https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&code_lz=FAYw9gdgzgLgBAQzgXjgbzgMzGAXHAcgCMEAnAuAXwG5RJY4iV0sd9iEAvAgGkbPYluVWnWjwQzNADpZCPrOlEaYqGAA2AU2nqwAcwAUIAJTUgA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=false&lineWrap=false&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=6.26.0&envVersion=):
```
return {...this.state, checked}
```
That will already create and return a new object.

phlax: ah, thanks, nice

adngdb: Oh, I never saw that syntax before! Sweet!

PR #905, opened 2018-03-19
Fix Bug 1440895: Add dj.debug package to requirements-dev

phlax:
This allows the use of a @debug decorator that can be used to
print out the sql used in a method or function.

Also provides a contextmanager to do similar for a block of code.

Ideally, we add this to staging, and then we can check/share performance changes

### DIFF
--- a/bin/post_compile
+++ b/bin/post_compile
@@ -1,5 +1,11 @@
 #!/usr/bin/env bash
 
+# Install requirements-dev.txt on staging
+if [ "$DJANGO_STAGE" = True ]; then
+    echo "Installing dev requirements..."
+    pip install -r requirements-dev.txt
+fi
+
 # Compile static assets
 export PATH=/app/.heroku/node/bin:$PATH
 npm install .
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -13,6 +13,8 @@ django-extensions==1.6.7 \
     --hash=sha256:ebdfc329207231f63b5ccb591407aadbc1a524af752e52690f0646eb7ff52943
 django-sslserver==0.19 \
     --hash=sha256:1363835229a0585f89c42f3beca836572f3f6babdc1508f13352aed84b0924b3
+dj.debug==0.0.2 \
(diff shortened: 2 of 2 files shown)


bin/post_compile
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+pip install -r requirements-dev.txt
```
jotes: I'm not sure if Heroku won't run that in the production environment.

phlax: good spot!

i wonder how we can do this only on staging

bin/post_compile
```diff
@@ -1,5 +1,12 @@
 #!/usr/bin/env bash
 
+
+# Install requirements-dev.txt on staging
+if [ "$HEROKU_APP_NAME" == "mozilla-pontoon-staging" ]; then
```
mathjazz: Instead of hardcoding the instance name, maybe use `DJANGO_DEV` here and set it to True on stage?

phlax: hmm, maybe they are independent

```
DJANGO_DEBUG:                 False
DJANGO_DEV:                   False
DJANGO_LOG_LEVEL:             DEBUG
```

ill update and see what happens...

mathjazz: We might need a new variable - `STAGE`.

phlax: @mathjazz ive updated with `DJANGO_DEV` and checked staging/logs etc - all looks good.

i think this is the right way to go - but i can change to `STAGE` if you want

mathjazz:
Setting `DJANGO_DEV=True` impacts other things, which I don't think are desirable on stage, especially Django Toolbar:

https://github.com/mozilla/pontoon/blob/master/pontoon/settings/base.py#L52
https://github.com/mozilla/pontoon/blob/master/pontoon/settings/dev.py
https://github.com/mozilla/pontoon/blob/master/pontoon/api/urls.py

I think we should add a `STAGE` env variable.

phlax:
> I think we should add a STAGE env variable.

ah, ok, on it - sorry this has taken a few attempts to get right

PR #906, opened 2018-03-19
Bug 1447095: FTL access key UI fixes

mathjazz:
This patch brings two fixes for the access key UI:
1. The list of candidates is now initialized properly - by extending the selector.
2. The access key textarea element no longer transforms text to uppercase.

@phlax Since you reviewed #899, which is related, I'd like to ask you to look at this one as well.

I deployed the patch to stage.

The bug on prod:
https://pontoon.mozilla.org/sl/firefox/browser/browser/preferences/preferences.ftl/?search=accessk&string=176239

The fix on stage:
https://mozilla-pontoon-staging.herokuapp.com/sl/test-fluent/preferences.ftl/?search=accessk&string=176235

### DIFF
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -293,7 +293,7 @@ var Pontoon = (function (my) {
           $('#ftl-area .attributes ul:first').append(attributes);
 
           // Update access keys presentation
-          $('#ftl-area textarea:visible').keyup();
+          $('#ftl-area textarea.value').keyup();
         }
 
         // Ignore editing for anonymous users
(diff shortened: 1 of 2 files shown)


pontoon/base/static/js/fluent_interface.js
```diff
@@ -293,7 +293,7 @@ var Pontoon = (function (my) {
           $('#ftl-area .attributes ul:first').append(attributes);
 
           // Update access keys presentation
-          $('#ftl-area .attributes textarea').keyup();
+          $('#ftl-area textarea:visible').keyup();
```
phlax: this feels like quite a loose selector - `#ftl-area textarea.value:visible` maybe ?

mathjazz:
Nice catch! I updated the selector
* added `.value` to tighten it a little
* removed `:visible` which caused the regression

It's deployed to stage now.

phlax:
cool, thats stopped it caning accessskeys

im still wondering tho whether raw mode should be showing ~same for access keys as the rich edit mode

phlax:
> im still wondering tho whether raw mode should be showing ~same for access keys as the rich edit mode

and after wondering aloud i realize my mistake - source view is showing **selected** access key 8/

PR #907, opened 2018-03-20
Fix bug 1447103: Exclude placeables from access key candidates

mathjazz:
Fluent uses curly brackets to denote Placeables, which must not be used to generate access key candidates, because in in Pontoon UI they don't contain translateable text (but references to external arguments or other entities).

Bug on prod:
https://pontoon.mozilla.org/sl/firefox/browser/browser/preferences/preferences.ftl/?string=176239

Fixed on stage:
https://mozilla-pontoon-staging.herokuapp.com/sl/test-fluent/preferences.ftl/?string=176235

### DIFF
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -921,8 +921,9 @@ $(function () {
       selector = '.attributes [data-id="label"]';
     }
     $('#ftl-area ' + selector + ' textarea').each(function() {
-      // Remove whitespace
-      content += $(this).val().replace(/\s/g,'');
+      content += $(this).val()
+        .replace(/{.*?}/g, '') // Remove placeables: bug 1447103
+        .replace(/\s/g, '');   // Remove whitespace
     });
 
     // Extract unique candidates in a list


pontoon/base/static/js/fluent_interface.js
```diff
@@ -921,8 +921,9 @@ $(function () {
       selector = '.attributes [data-id="label"]';
     }
     $('#ftl-area ' + selector + ' textarea').each(function() {
-      // Remove whitespace
-      content += $(this).val().replace(/\s/g,'');
+      content += $(this).val()
+        .replace(/{.*?}/g, '') // Remove placeables: bug 1447103
```
phlax: im wondering if there are situations where there might be nested `{}` , as this would fail in that case - ie:

```
>>> ' foo {  asdfa asdf {asd} asdf} bar '.replace(/{.*?}/g, '')
" foo  asdf} bar "
```

phlax: i had a quick look at the spec, and there are cases where brackets are nested, but i wasnt sure if they are relevant here

mathjazz: There are such cases in FTL, but they'd be split into multiple `textarea` elements.

For example this:
https://github.com/mozilla-l10n/pontoon-ftl/blob/master/en-US/demo.ftl#L23

Looks like this in Pontoon:
https://mozilla-pontoon-staging.herokuapp.com/sl/test-fluent/demo.ftl/?string=176160

phlax: cool

jotes: I'm not sure if that's a bug, but if you'll type this string:
"Datoteka {"
Then switch to the source mode and then back to the simple mode, inputs are blank.
But, if you'll type this string in the source view directly, you'll get a message: "Expected literal". I think validation should work in both cases.

Sorry if that's already mentioned somewhere.

mathjazz: That happens when the string contains ParseErrors and it's not related to this patch.

zbraniecki: That seems like a footgun in a long turn when you'll encounter escaped patterns for example:

```
key = Hello \{ and then { -brand-short-name} World
```

I think that in all places where you handle Fluent it would be nice to parse it using `fluent-syntax` and rely on it to produce AST for you rather than trying to recreate the parser for custom use cases.

mathjazz: Bummer, I didn't know you could do that. Makes sense.

PR #911, opened 2018-03-23
Fix bug 1433499: Refactor fluent_interface.js

mathjazz:
A dozen or so fixes that should make `fluent_interface.js` easier to review and migrate to React. The refactor is limited to the plan outlined in [the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1433499#c3).

Diff is unreadable, so I strongly suggest looking into each commit diff message separately.

### DIFF
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -82,7 +82,7 @@ var Pontoon = (function (my) {
       }
 
       // Render SelectExpression
-      if (Pontoon.fluent.isSelectExpressionElement(element)) {
+      if (element.expression && element.expression.type === 'SelectExpression') {
         element.expression.variants.forEach(function (item) {
           content += renderOriginalElement(item.key.value || item.key.name, item.value.elements);
         });
@@ -118,7 +118,7 @@ var Pontoon = (function (my) {
       }
 
       // Render SelectExpression
-      if (Pontoon.fluent.isSelectExpressionElement(element)) {
+      if (element.expression && element.expression.type === 'SelectExpression') {
         var expression = fluentSerializer.serializeExpression(element.expression.expression);
         content += '<li data-expression="' + expression + '"><ul>';
 
@@ -328,7 +328,7 @@ var Pontoon = (function (my) {
           return elements.every(function(element) {
             return (
               self.isSimpleElement(element) ||
-              self.isSelectExpressionElement(element)
+              (element.expression && element.expression.type === 'SelectExpression')
             );
           });
         }
@@ -375,25 +375,13 @@ var Pontoon = (function (my) {
 
 
       /*
-       * Is element a SelectExpression?
-       *
-       * We evaluate that by checking if element expression has variants defined.
-       * If yes, the element must be a SelectExpression according to Fluent ASDL:
-       * https://github.com/projectfluent/fluent/blob/master/spec/fluent.asdl
-       */
-      isSelectExpressionElement: function (element) {
-        return element.expression && element.expression.variants;
-      },
-
-
-      /*
        * Is element representing a pluralized string?
        *
        * Keys of all variants of such elements are either
        * CLDR plurals or numbers.
        */
       isPluralElement: function (element) {
-        if (!this.isSelectExpressionElement(element)) {
+        if (!(element.expression && element.expression.type === 'SelectExpression')) {
           return false;
         }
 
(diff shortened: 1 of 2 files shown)


mathjazz:
Thanks for having a look. I'll leave the decision about tests to @adngdb.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -6,21 +6,136 @@ var fluentSerializer = new FluentSyntax.FluentSerializer();
...
+   */
+  function isSupportedMessage(ast) {
+    function elementsSupported(elements) {
+      return elements.every(function(element) {
+        return (
+          isSimpleElement(element) ||
+          (element.expression && element.expression.type === 'SelectExpression')
+        );
+      });
+    }
```
adngdb: As far as I can see, there's no point in defining this function inside another function. By doing that, you'll force the compiler to recreate the `elementsSupported` function every time you invoke `isSupportedMessage`, so it would be better to put it outside (even if it's only ever used inside `isSupportedMessage`).

pontoon/base/static/js/fluent_interface.js
```diff
@@ -6,21 +6,136 @@ var fluentSerializer = new FluentSyntax.FluentSerializer();
...
+   * elements are simple.
+   */
+  function isSimpleMessage(ast) {
+    if (
+      ast &&
+      !ast.attributes.length &&
+      ast.value &&
+      ast.value.elements.every(function(item) {
+        return isSimpleElement(item);
+      })
```
adngdb: You can simplify this to `ast.value.elements.every(isSimpleElement)`.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -6,21 +6,136 @@ var fluentSerializer = new FluentSyntax.FluentSerializer();
...
+    return false;
+  }
+
+  /*
+   * Render textarea element with the given properties
+   */
+  function renderTextareaElement(id, value, maxlength) {
+    var element = '<textarea class="value" id="ftl-id-' + id + '"';
+
+    if (typeof maxlength !== 'undefined' && maxlength !== null) {
```
adngdb: Is `maxlength = 0` a use case we want to support here? It seems weird to have a textarea in which you can't write anything. To me that should be a disabled textarea instead.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -528,70 +455,74 @@ var Pontoon = (function (my) {
...
-        Pontoon.fluent.toggleEditorToolbar();
+        toggleEditorToolbar();
         Pontoon.moveCursorToBeginning();
       },
 
 
       /*
-       * Toggle translation length and Copy button in editor toolbar
+       * Get source string value of a simple FTL message to be used in
+       * tje Copy (original to translation) function
```
adngdb: typo? `the` instead of `tje`?

adngdb:
> I wonder about the move to internal functions. As much as that clarifies the API surface of Pontoon.fluent, how do we envision for them to get tests at some point?

By the time we get to testing that code, we'll probably switch this file to an ES6-style file using imports and exports. When that happens, all those functions will be easier to access and test.

pontoon/base/static/js/fluent_interface.js
mathjazz: No, we shouldn't support `maxlength = 0`. Actually, this condition is always true. I'll remove the `if`.

mathjazz:
@adngdb Thanks for the revew. Comments addressed!

adngdb:
Bug with plurals. Steps to reproduce: 

1. Go to Pontoon FTL project, open `Delete {$num} downloads?` string.
2. Enter translation for `one`.
3. Start typing translation for `other`, click placeable `{$num}` in content.

Expected result:
The string `{$num}` is added to the `other` string, right where the cursor is.

Actual behavior:
The string `other` is replaced with the content from `one`, and `{$num}` is added to the end of it. Both `one` and `other` are changed.

mathjazz:
Nice catch. It's not related to the patch, though.

pontoon/base/static/js/translate.js
```diff
@@ -1764,13 +1764,17 @@ var Pontoon = (function (my) {
...
+        if (!textarea.length) {
+          textarea = $('#editor textarea:visible:first');
+        }
+
+        var selectionStart = textarea.prop('selectionStart');
+        var selectionEnd = textarea.prop('selectionEnd');
+        var placeable = $(this).text();
+        var cursorPos = selectionStart + placeable.length;
+        var before = textarea.val();
+        var after = before.substring(0, selectionStart) + placeable + before.substring(selectionEnd);
```
adngdb: Yay to `var`s! :-)

PR #912, opened 2018-03-26
String tiers django-based views

phlax:
Implementation of the string tiers project views using django/jinja

### DIFF
--- a/pontoon/tags/urls.py
+++ b/pontoon/tags/urls.py
@@ -1,0 +1,17 @@
+
+from django.conf.urls import url
+
+import views
+
+
+urlpatterns = [
+    # Project tag page
+    url(r'^projects/(?P<project>[\w-]+)/tags/(?P<tag>[\w-]+)/$',
+        views.ProjectTagView.as_view(),
+        name='pontoon.tags.project.tag'),
+
+    # AJAX view: Project tag teams
+    url(r'^projects/(?P<project>[\w-]+)/ajax/tags/(?P<tag>[\w-]+)/$',
+        views.ProjectTagView.as_view(),
+        name='pontoon.tags.ajax.teams'),
+]
--- a/pontoon/base/static/css/table.css
+++ b/pontoon/base/static/css/table.css
@@ -28,6 +28,10 @@ table.table {
   width: 420px;
 }
 
+.table th.tag {
+  width: 330px;
+}
+
 .table th.population,
 .table th.deadline {
   width: 90px;
--- a/pontoon/base/templates/widgets/heading_info.html
+++ b/pontoon/base/templates/widgets/heading_info.html
@@ -13,9 +13,9 @@
 </li>
 {% endmacro %}
 
-{% macro details_item_priority(priority) %}
+{% macro details_item_priority(priority, title='Priority') %}
 <li class="priority">
-  <span class="title">Priority</span>
+  <span class="title">{{ title }}</span>
   <span class="value">
     {% for n in range(5) %}
       <span class="fa fa-star{% if loop.index <= priority %}{{ ' active' }}{% endif %}"></span>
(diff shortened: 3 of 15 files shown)


mathjazz:
Nice work!

On the `/projects/PROJECT/tags/` dashboard I see 2 things missing:
1. Priority needs to be represented with stars.
1. Tags entry flickers on hover. You can prevent it by adding the following rule to the `table.css`:
```css
.table th.tag {
  width: 330px;
}
```

On the `/projects/PROJECT/tags/TAG/` dashboard everything looks fine in the tab, but the heading needs to follow the [spec](https://user-images.githubusercontent.com/626716/35684935-a14da4ae-0768-11e8-9c25-0c4c18424dc7.png) more closely, for example:
* change the `<title>` tag: Project Tag
* change dashboard title: Project Tag
* add Tag priority to the heading
* reflect only the tag in stats

mathjazz:
Really nice work!

Two last things I have are:

1. In `/projects/PROJECT/tags/TAG/` "PRIORITY" in the heading should ready "PROJECT PRIORITY".
1. In `/projects/PROJECT/tags/` it would be nice to add a search field, same as style as we use for searching teams, projects and resources in other dashboards. It not an absolute must, but since the UI is so similar to other dashboards, the user kind of expects it to be there.

phlax:
> it would be nice to add a search field

i thought about this, but as the expected number of tags is likely to be low initially, i thought it would be kinda superfluous

pontoon/tags/urls.py
```diff
@@ -0,0 +1,17 @@
...
+import views
+
+
+urlpatterns = [
+    # Project tag page
+    url(r'^projects/(?P<project>[\w-]+)/tags/(?P<tag>[\w-]+)/$',
+        views.ProjectTagView.as_view(),
+        name='pontoon.tags.project.tag'),
+
+    # AJAX view: Project tag
```
mathjazz: Nit: `# AJAX view: Project tag teams`

pontoon/tags/urls.py
```diff
@@ -0,0 +1,17 @@
...
+    url(r'^projects/(?P<project>[\w-]+)/ajax/tags/(?P<tag>[\w-]+)/$',
+        views.ProjectTagView.as_view(),
+        name='pontoon.projects.ajax.tags'),
```
mathjazz: Nit: Looking at the `localization` module, `pontoon.tags.ajax.teams` would be more consistent.

pontoon/tags/templates/tags/includes/tags.html
```diff
@@ -0,0 +1,20 @@
+<div id="info-wrapper">
```
mathjazz: You can remove this wrapper.

pontoon/tags/templates/tags/includes/tags.html
```diff
@@ -0,0 +1,20 @@
...
+
+{% import 'tags/widgets/tag_list.html' as TagList %}
+
+<div class="resources">
```
mathjazz: `class="tags"`

pontoon/tags/templates/tags/includes/tags.html
```diff
@@ -0,0 +1,20 @@
...
+  {% set latest_activity = tag.latest_activity %}
+  {% set chart = tag.chart %}
+  {{ TagList.item(tag, main_link, chart_link, latest_activity, chart) }}
+  {% endfor %}
+
+  {{ TagList.footer() }}
+</div>
+
+
+</div>
```
mathjazz: This file should go into `projects/templates/projects/includes`. Each file in that folder represents a tab on the project dashboard.

pontoon/tags/templates/tags/includes/tag.html
```diff
@@ -0,0 +1,37 @@
+{% import 'widgets/latest_activity.html' as LatestActivity %}
```
mathjazz: As far as I can see this file is never used and so is the imported `tag_locales.html` file.

PR #914, opened 2018-03-26
Fix bug 1448814: Add 3rd-party service copyright notice to Terms

mathjazz:
@guerojeff Could you have a look at the English and legal in this patch?

a-fullpage


gueroJeff:
lgtm, thanks!

PR #915, opened 2018-03-26
String tiers translate filter

phlax:
Adds tag filters to translate filtering dropdown

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -2152,7 +2152,7 @@ class Entity(DirtyFieldsMixin, models.Model):
 
     @classmethod
     def for_project_locale(
-        self, project, locale, paths=None, status=None,
+        self, project, locale, paths=None, status=None, tag=None,
         search=None, exclude_entities=None, extra=None, time=None, author=None,
     ):
         """Get project entities with locale translations."""
@@ -2216,8 +2216,14 @@ class Entity(DirtyFieldsMixin, models.Model):
                 )
             )
 
+        if tag:
+            post_filters.append(Q(resource__tag__slug__in=tag.split(',')))
+
         if post_filters:
             entities = entities.filter(Q(*post_filters))
+            if tag:
+                # only tag needs `distinct` as it traverses m2m fields
+                entities = entities.distinct()
 
         # Filter by search parameters
         if search:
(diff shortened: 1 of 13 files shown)


pontoon/base/models.py
```diff
@@ -2216,8 +2216,11 @@ def for_project_locale(
                 )
             )
 
+        if tag:
+            post_filters.append(Q(resource__tag__slug__in=tag.split(',')))
+
         if post_filters:
-            entities = entities.filter(Q(*post_filters))
+            entities = entities.filter(Q(*post_filters)).distinct()
```
phlax: `distinct` is only needed if tag is set

pontoon/base/views.py
```diff
@@ -102,6 +103,7 @@ def translate(request, locale, slug, part):
         'part': part,
         'project': project,
         'projects': projects,
+        'tags': TagsTool(projects=[project])
```
phlax: this requires `priority=True`

pontoon/base/views.py
```diff
@@ -102,6 +103,7 @@ def translate(request, locale, slug, part):
         'part': part,
         'project': project,
         'projects': projects,
+        'tags': TagsTool(projects=[project], priority=True)
```
phlax: add `if project.tags_enabled`

phlax:
TODO:

- make links in dashboards work with tag filtering
- single click filtering doesnt work atm
- update icons for tags in filter dropdown

mathjazz:
For the icon I suggest we use the [Font Awesome](https://fontawesome.com/cheatsheet) "tag" icon: .

mathjazz:
Instead of using the (translation) `status` parameter, we should introduce the `tag` parameter and use that for filtering by tag.

phlax:
> Instead of using the (translation) status parameter,

this was the bug "single click filtering doesnt work atm"

pontoon/base/static/css/translate.css
```diff
@@ -708,6 +708,12 @@ body > header aside p {
   display: none;
 }
 
+[class^="tag-"] .status.fa:before,
+[class*=" tag-"] .status.fa:before {
```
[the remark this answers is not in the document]
phlax: either should work - but the current selector is a bit more specific/accurate

PR #916, opened 2018-03-28
Fix bug 1449126 - Correctly handle error when producing empty string …

adngdb:
…with search and replace batch action.


adngdb:
I really wanted to add a test for this, but `mock` got the best of me. I couldn't figure out how to mock the `ACTIONS_FN_MAP` dict from the tests. If you manage to do it, please show me how it can be done!

PR #917, opened 2018-03-28
Add query params in lang agnostic redirect

phlax:
Allow query params to be passed through when redirecting to lang translate view


mathjazz:
It would be nice to redirect "All Projects" properly as well. It could also be a separate patch.

https://gist.github.com/mathjazz/e191e7136a37c6dd89696a9840a53c22

PR #919, opened 2018-03-31
Fix Bug 1450464: prevent tags view failing where latest translation is None

phlax:
Im not sure how this situation is created - but this fixes the issue for the tags view, and works the same as in other views


mathjazz:
I reproduced the error locally now - it happens when the translation is empty. In this case we get pass the [condition](https://github.com/mozilla/pontoon/blob/master/pontoon/base/templates/widgets/latest_activity.html#L3) and fail at `latest_activity.translation`.

r++. :-)

PR #922, opened 2018-04-03
Fix Bug 1450863: Use User.with_translation_counts for getting contribs in translate dd

phlax:
using the contribs code path in translate dropdown

### DIFF
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -179,7 +179,7 @@ def authors_and_time_range(request, locale, slug, part):
     translations = Translation.for_locale_project_paths(locale, project, paths)
 
     return JsonResponse({
-        'authors': translations.authors().serialize(),
+        'authors': translations.authors(),
         'counts_per_minute': translations.counts_per_minute(),
     }, safe=False)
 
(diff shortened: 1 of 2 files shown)


pontoon/base/views.py
```diff
@@ -177,9 +177,20 @@ def authors_and_time_range(request, locale, slug, part):
     paths = [part] if part != 'all-resources' else None
 
     translations = Translation.for_locale_project_paths(locale, project, paths)
-
+    filters = Q(entity__resource__project=project)
```
mathjazz: You [also](https://github.com/mozilla/pontoon/blob/master/pontoon/base/models.py#L2434) need to filter by locale and filter out obsolete entities:
`filters = Q(entity__resource__project=project) & Q(entity__obsolete=False) & Q(locale=locale)`

phlax: ah, thanks, ill update

phlax:
> This logic should go into TranslationQuerySet.authors(), otherwise the method is unused.

cool, ill move it - i wasnt sure where it should be

PR #923, opened 2018-04-03
Fix bug 1450265: Filter Translators by project

mathjazz:
The Translation Time and Translation Authors sections of the filter menu in the translate view should only include translations of the project that is currenty being translated if "All Resources" is selected (i.e. `path` not specified).

Note that #922 will fix the issue for Translation Authors, but not for the Translation Time filter, which also depends on `for_locale_project_paths`.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -2437,6 +2437,7 @@ class Translation(DirtyFieldsMixin, models.Model):
         """
         translations = Translation.objects.filter(
             entity__obsolete=False,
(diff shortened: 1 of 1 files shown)


mathjazz:
I'm not sure why the test fails, because it doesn't seem related to the patch.

I tried to change the index from 1 to 0, but it doesn't help:
https://gist.github.com/mathjazz/e8ccf3ce350cb489ab3efa95ab985f17

No idea how to fix this.

PR #925, opened 2018-04-06
Fix bug 1213346: Turn mozilla-en-US.py into a management command

mathjazz:
We'll be closing the SLC3 data center in Q2, so we need to move the `mozilla-en-US.py`script that's currently running on `pontoon1.dev.webapp.scl3.mozilla.com` to Heroku.

This patch turns the script into a management command first, which can then be preiodically executed using Heroku Scheduler (like `sync_projects` for example).

I didn't make any other changes to the script yet, except for replacing the custom logging function. We could also replace `pull` and `push` functions with `pontoon.sync.vcs.repositories`, but I'd rather keep the changeset as minimal as possible and save time to focus on [bug 1385834](https://bugzilla.mozilla.org/show_bug.cgi?id=1385834).

@Pike Given that you were reviewing the last changes made to the script, I wonder if you'd be willing to take a look.

I tested the script on stage by pushing to [-aurora](https://hg.mozilla.org/users/m_owca.info/?sort=lastchange) repos and it seems to be working fine. Total execution time on first load was ~1 minute.


mathjazz:
Actually now that I'm thinking about it, maybe we could simply execute the old script via the Heroku Scheduler. I'll try that.

PR #926, opened 2018-04-11
Fix bug 1377969 - Rename suggested to unreviewed

adngdb:


mathjazz:
Nice work! I tested the patch without looking at the code and it seems to be working fine. I noticed a few things that should be changed:

**Dashboards** (http://localhost:8000/teams/, http://localhost:8000/projects/, http://localhost:8000/fr/, http://localhost:8000/projects/firefox/, http://localhost:8000/fr/firefox/):
* Remove the blue square in front of `Unreviewed strings` in the dashboard heading. This part of the screen is used as a legend for the circular progress chart, and we cannot paint unreviewed on it.
* When you hover the horizontal progress bar in the list, the `Suggested` should use the default color instead of blue (same as `All Strings`) and moved to the right, in front of `All Strings`, to follow the order we have in the dashboard heading.

**Contributors** (http://localhost:8000/contributors/):
* `Suggested strings` in the heading should be renamed to `Unreviewed strings` and moved down, right in front of `All Strings`.
* In the contributor list, `Suggested` should be renamed to `Unreviewed`, colored grey (same as `Total`) and moved to the far right.

**Profile** (http://localhost:8000/profile/):
* `Suggested` should be renamed to `Unreviewed` and moved to the far right.

**Translate** (http://localhost:8000/sl/firefox/all-resources/):
* The blue status icon in the string list should be replaced with grey (missing).
* When you click on the progress chart, `Unreviewed` should be removed from the legend below the progress chart. You can add it under `All strings`.
* `Unreviewed` filter should be moved to the bottom of Translation status filters, because it's not really a status. The blue icon should be replaced, maybe with the circle we use ATM.

adngdb:
@mathjazz this is now ready for your review again! I've (hopefully) addressed all your comments. :-)

adngdb:
@mathjazz After looking at the code, I don't see how `calculate_stats` can create negative numbers. Could you help me reproduce the error so that I can investigate and debug? Can you maybe send me your DB dump? 

I've run the migration against a local prod dump and it passes (even though it takes a very long time).

mathjazz:
Thanks for addressing all the issues!

* OK, let's ignore the migration problem for now. Seems like I have some inconsistencies in the DB. I'll look into it more closely and comment. BTW, do you have any ideas on how to speed the data migration up?
* The `pontoon/base/tests/views/test_entity.py::test_view_entity_exclude_entities`test is failing.
* We should [expose unreviewed in the API](https://github.com/mozilla/pontoon/commit/6a8c517a3759e211e518bd4ad8a91c09244d62da#diff-b93462d9cf6c04d9b6c87a841029e899).

Other than that, we seem to be good to go! I'm still looking at the code, but haven't found anything bad so far. :)

mathjazz:
> OK, let's ignore the migration problem for now. Seems like I have some inconsistencies in the DB. I'll look into it more closely and comment.

Confirming that the problem was caused by negative numbers already being present in my DB.

MikkCZ:
@mathjazz as discussed today, can you please let me know when this is in staging, or any other testing environment? Or if the patch is more or less ready, I can test and fix Pontoon Tools locally. I have realized I actually get the data via API, not parsing from DOM.

adngdb:
> BTW, do you have any ideas on how to speed the data migration up?

Yes, but all of them would require more than 2 hours of work, and 2 hours is the time it should take to complete on prod (probably even less considering that's the time it takes on my computer). I think it's not worth the effort, unless you think we will have to run that migration several times? 

> The pontoon/base/tests/views/test_entity.py::test_view_entity_exclude_entitiestest is failing.

That's unrelated to this PR, it's a weird bug that happens only the first time you run the test suite. If you run it again, it disappears. I introduced it with my pytest work, and haven't looked into it yet. 

@MikkCZ this is very very close to being ready. The one big change that might impact you is the data exposure in the API that @mathjazz mentioned. Other than that, we might see small UI changes but nothing impacting.

mathjazz:
@adngdb OK, thanks for clarifying. I'm re-runnig Travis now.

I wonder if 2 hours of running a migration means we should deploy this in two steps (2nd sept being a data migraion only). We should test on stage.

I'll have another look at the code and if all is well, we should aim for a release on Monday. The patch will also require us to make some changes to docs. (cc @flodolo).

flodolo:
Skimmed through bug and issue, but couldn't see anything related to this question: are we going to expose all unreviewed suggestions as "Unreviewed strings"? I remember we talked about a strategy to clean up most of those, i.e. automatically rejecting suggestions if they were older than the currently approved string. Am I mixing this with a different bug/discussion?

flodolo:
> We mass-rejected such suggestions in the previous patch:
> https://blog.mozilla.org/l10n/2017/08/10/making-unreviewed-suggestions-easier-to-find/

Thanks, that makes sense now (and it's sad to see how many unreviewed suggestions we still have…)

flodolo:
I'm happy to give it a run on stage to start updating docs. As anticipated in SFO, It would help to get a list of changes to look out for.

mathjazz:
Deploying to stage now. It will probably take a couple of hours.

MikkCZ:
Cool. If the deploy is successful, I will start working on Pontoon Tools tonight.

Btw. I don't know, if that's related, but staging now gives 500 (but no tetris).

mathjazz:
That's related - the migration is still running.

mathjazz:
Deployment failed with:
```
remote: Running migrations:
remote:   Applying base.0123_rename_translated_to_unreviewed... OK
error: RPC failed; curl 18 transfer closed with outstanding read data remaining
fatal: The remote end hung up unexpectedly
fatal: The remote end hung up unexpectedly
Everything up-to-date
```

Investigating and in the meantime, trying again.

Alternative approach to a data migration is a management command which we'd manually run after deployment. That means deployment will be as fast as usually, site will not go down, but the unreviewed string numbers will be off until the management command completes.

mathjazz:
This time it failed with:
```
remote: Running migrations:
remote:   Applying base.0124_recalculate-all-stats...-----> Timed out running buildpack Python
remote: 
remote:  !     Timed out compiling app (16 minutes)
remote:  !     See https://devcenter.heroku.com/articles/slug-compiler#time-limit
remote: Verifying deploy...
remote: 
remote: !	Push rejected to mozilla-pontoon-staging.
```

mathjazz:
I moved the data migration to a management command in order to be able to deploy to stage. I succeeded, so stage is now using this branch, ready for @flodolo to consume for documentation.

ATM, I'm running the management command to fix the stats.

mathjazz:
I removed the icon from the Unreviewed filter, which was indeed a weird choice.

Total and Unreviewed use a default UI color, because we don't paint them in charts anymore. Despite that, we could use a consistent color for suggestions across the UI, e.g. blue or purple (I'm keeping orange and red for warnings and errors).

I have a counter proposal though. To make suggestions truly discoverable, we should put them in a separate column, which will make them always visible. Previously, you could (at least the ones in a `Suggested` state) see them from the linear chart, but now they are completely hidden until you hover.

We often hear complaints about suggestions not being acted upon, and making them as visible as possible is IMHO the first step to fixing this.

screen shot 2018-06-19 at 15 38 45

mathjazz:
I that makes sense, I can provide a patch and push it to stage.

flodolo:
> To make suggestions truly discoverable, we should put them in a separate column, which will make them always visible.

It definitely makes them more actionable and visible, but it looks a bit weird. Together with the other numbers, it's clearly "(number of) unreviewed (strings)", here it's harder to understand at first glance.

Make the graph should have a different look when there are unreviewed suggestions? e.g. some sort of overlay, making it different from a project at 100% and without unreviewed suggestions.

MikkCZ:
What about changing the meaning of 100% from "all string have approved
suggestions" to "everything is approved and all suggestions reviewed"?
Would that work? Or may localizers think the project is not complete yet,
while it is in the repository?

--
Michal Stanke

Dne út 19. 6. 2018 16:02 uživatel Francesco Lodolo <notifications@github.com>
napsal:

MikkCZ:
> That seems really hard to calculate in practice. You can be at 100% (all translated), and have 5% of unreviewed translations. Would you display 95% there?

Yes, that was my idea. It depends how critical you think the action on suggestions are.

Or can we paint two graphs/lines below each other for each project? One for project completion (strings having approved suggestions) and second for reviews (strings with all suggestions reviewed)?

mathjazz:
I'm not convinced about mixing translation and review completion percentages. If my project is shipping complete and then I get 20% of weird suggestions, I don't want my completion percentage to show 80%.

I like the idea of adding additional graphic to denote pending reviews. I'll experiment a little.

mathjazz:
A possible solution, along the lines of "the graph should have a different look when there are unreviewed suggestions". I believe a binary representation "has vs doesn't have suggestions" is sufficient, no need for adding more complexity with additional progress chart.

The first one is more subtle, the second one uses the old "blue" color we used for Suggested.

I'm not sure about the icon:
https://fontawesome.com/cheatsheet?from=io

I'm thinking about `exclamation-circle`, `question-circle`, `flag` and `lightbulb`. We could reuse it in the dashboard heading and the filter menu. Ideas welcome!

screenshot_2018-06-19 slovenian sl 1

screenshot_2018-06-19 slovenian sl

flodolo:
I think I like the second option (more visible), with a flag icon.

MikkCZ:
Personally I like [this lightbulb](https://fontawesome.com/icons/lightbulb?style=regular), but I think it does not matter that much if it is a lightbulb or a flag, if it's used consistently both on the dashboard and in the filter view. Only I would suggest is to choose as simple symbol as possible to not break the gentle UI. Or maybe to reuse the bell as a sign for notification?

As for the UI layout, I prefer the second colored proposal too. The only change I would make it to swap the percentage number and the icon, because of the combination with sorting. I would expect the column to be sorted by the right-most object in it. So the whole column would look like:

Firefox   -   ⭐⭐⭐⭐⭐   1 week ago   =========- 🔔 90%

MikkCZ:
This way with the icon directly next to the graph, it would also make sense to show the number of unreviewed strings on hover, as on the screenshot above https://user-images.githubusercontent.com/3644868/41580237-6fcc60a2-739a-11e8-8e14-9533d6b20a9a.png

mathjazz:
I added a sortable "Unreviewed" column (identified by a blue flag) to /projects, /teams, project, locale, localization and tags dashboards to denote if unreviewed suggestions are pending. Hovering over a flag in the table head shows a tooltip with description, to minimize the confusion.

The same blue color (previously used for Suggested) and flag are used for marking Unreviewed across the app, including filters and contributors dashboards & profiles.

Check it out on stage:
https://mozilla-pontoon-staging.herokuapp.com/sl/

Sneak peek:
screen shot 2018-06-19 at 23 02 00

Let me know if this looks good. And then we'll need to update a bunch of screenshots again. :/

Pike:
Looking at https://mozilla-pontoon-staging.herokuapp.com/projects/firefox/, ordering by unreviewed gives me a weird ordering.

Food for thought, do we care about the difference between unreviewed suggestions for missing strings vs unreviewed suggestions for existing ones?

mathjazz:
> Looking at https://mozilla-pontoon-staging.herokuapp.com/projects/firefox/, ordering by unreviewed gives me a weird ordering.

Clicking on the flag doesn't order by the number of Unreviewed strings. It orders by the "Has Unreviewed Strings?" boolean.

Maybe I should upgrade that? Ordering by Unreviewed strings also means ordering by that boolean.

> Food for thought, do we care about the difference between unreviewed suggestions for missing strings vs unreviewed suggestions for existing ones?

Not sure if we do in terms of numbers (stats), but we have the ability to separate them with combinations of filters.

mathjazz:
Updated stage with sorting by Unreviewed count. Also fixed a bug in sorting by progress.

The management command to calculate stats should be a little faster now and also more verbose. It starts with enabled projects, in ascending order of their resource count. Running it now, and it took it 20 minutes to get to Firefox. So we should be done in ~30 minutes for all enabled projects.

Tomorrow, I'll see if I can find a better icon than a flag, I'm still not sure about that. Will experiment with the bulb.

flodolo:
Current stage looks good to me. Need to wait for the new icon to redo the screenshots.

flodolo:
Only note: I'm not fond of a tooltip with a question mark. Should it just be "Unreviewed strings"?

mathjazz:
@adngdb Could you please have a look at https://github.com/mozilla/pontoon/pull/926/commits/cc9d463b431d6fa239b3e3e92bf2c7c0971adcb3? Seems like it worked fine on stage.

mathjazz:
Deployed the latest version to stage, which replces a flag icon with a lightbulb. It seems to be the safest bet given what Google Image search for "Suggestion" hints at. Frankly, a flag is a more general symbol.

To use the lightbulb, I had to update FontAwesome from v4 to v5, which makes the PR look huge:
https://github.com/mozilla/pontoon/pull/926/commits/e8d1a1678e45b7b8e29ce29d8c231a3089229d48

I also updated the tooltip as suggested by @flodolo.

MikkCZ:
I like the current look of the list of projects. However probably because of the Fontawesome update, when you are logged out, the hamburger menu icon is missing (top right corner).

adngdb:
Tested locally, everything works fine. And I really like what you've done with the new icon!

PR #927, opened 2018-04-11
Bug 1453139: Prevent DoesNotExist for non-managers

mathjazz:
Details in commit messages.

### DIFF
--- a/pontoon/base/static/js/main.js
+++ b/pontoon/base/static/js/main.js
@@ -845,7 +845,7 @@ $(function() {
               attribute: 'forceSuggestions',
               value: is_enabled
             });
-            $('[id^="save"]').toggleClass('suggest', is_enabled);
+            Pontoon.updateSaveButtons();
           }
         }
       },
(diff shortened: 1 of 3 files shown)


pontoon/base/static/js/main.js
```diff
@@ -845,7 +845,7 @@ $(function() {
               attribute: 'forceSuggestions',
               value: is_enabled
             });
-            $('[id^="save"]').toggleClass('suggest', is_enabled);
+            Pontoon.updateSaveButtons();
```
adngdb: That function is created in `translate.js`, what happens if you are outside of the translate page and this gets called? As far as I can tell, this is going to be a JS error, potentially blocking the page.

adngdb: If this piece of code cannot be called outside of the Translate page, then it's not in the correct file.

mathjazz: Actually `Pontoon.user` in the condition above makes sure this is only executed in the translate view. The other option being the `/settings` page.

I know this is not ideal and I can revert this change for a later refactor.

adngdb: OK, fair enough then. I think we can live with this being in the wrong place for now, given that we're rewriting our JS anyway.

PR #928, opened 2018-04-16
Fix bug 1454351: Update the right string in string list

mathjazz:
The solution is simple: we need to select the string before the request
is made and not wait until it completes.


mathjazz:
Yeah I had to use `time.sleep()` to reproduce it reliably.

PR #930, opened 2018-04-17
Fix bug 1437875 - Run webpack in docker.

adngdb:
This commit also refactors some of our docker process. nodejs and npm are now installed from a package instead of via nvm. npm install, collectstatic and webpack are now part of the build process. This makes starting the webapp a lot faster.

This commit also changes the interface of make, removing all docker prefixes from commands, as they were all about docker anyway. It is much simpler to use now, and backwards compatibility has been kept.


CONTRIBUTING.rst
```diff
@@ -189,6 +189,13 @@ Running a file of tests::
   app@...:/app$ ./manage.py test pontoon/base/tests/test_views.py
 
 
+Note that currently, we run some tests with `django` and some with `pytest`.
+`make test` runs all of them, but if you want to run just some specific tests,
+and they are using `pytest`, you should run::
+
+  app@...:/app$ pytest path/to/my/test.py
```
mathjazz: The same is true for Django tests, right? Should we add a similar command here?

adngdb: Actually, fun fact: the command in the example doesn't work. That's one of the things I'm trying to fix with `pytest` at the moment. Also, given that I still want to get rid of running tests with django, I'd rather not spend too much time on this, and make a better documentation update with my next PR (it's going to be about `pytest`).

adngdb:
@Pike maybe you want to give this a try, if you have time? As far as I remember you were using docker locally. Or @stasm maybe?

CONTRIBUTING.rst
```diff
@@ -120,7 +120,7 @@ That format is documented more extensively inside the ``requirements.txt`` file.
 
 Once you are done adding or updating requirements, rebuild your docker environment::
 
-  $ make dockerbuild
+  $ make build
```
jotes: :beers:

CONTRIBUTING.rst
jotes: I use: `pytest -vsk 'test name'`

PR #938, opened 2018-04-27
Fix bug 1438575: Mass replace works on non-breakable spaces.

jotes:
Hey @mathjazz, could you look at this?


jotes:
Thx @TheoChevalier!
Unfortunately, I can't reproduce this issue on my local instance :(
I tried following combinations: 
"&nbsp;" -> " " 
" " -> "&nbsp;"

That would be helpful if you could paste the network communication between the frontend and the backend (especially everything that is sent to "/batch-edit-translations/"). That could help me to find the source of the problem. Thanks in advance!

PR #942, opened 2018-05-03
Fix bug 1456846 - Fix docker dev environment.

adngdb:
This commit ensures that:
 - the assets/ folder is always created
 - dependencies (Python and node) do not pollute the user's repository. They are now installed in a different folder inside the container
 - webpack takes resources from the right place, and puts bundles in the right folder

This looks like a small patch but it was actually fairly complex. The reason is that our Dockerfile is made for a production deployment, meaning that it builds all the resources it needs to run in place. That's fine. However, when we run locally for development, using docker-compose, we override the /app folder with the user's local folder. That means all resources built during `make build` are lost. This patch puts those resources in a different folder, and then links to that folder from the user's repository. This allows to keep a clean production build process, while having a working dev environment, and a clean local folder.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,8 @@ tmp/*
 docs/_build
 venv
 /media/
+.cache/
+coverage/
 
 # For docker
 .docker-build
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,10 @@ help:
 	@echo "  build-frontend-w Watches the frontend static files and builds on change\n"
 
 .docker-build:
-	make dockerbuild
+	make build
+
+assets:
+	mkdir -p assets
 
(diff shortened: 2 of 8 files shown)


mathjazz:
@exsilium Would you mind taking this one for a test drive?

exsilium:
will do

exsilium:
@mathjazz, it seems to work well. Thanks @adngdb! (bow) My minor suggestion to this patch would be to squash some of the `RUN`s for less layers that would make the build a bit speedier with less steps. Cheers! 🍻

exsilium:
If so, could we get this pulled? :crossed_fingers: :bowtie:

PR #946, opened 2018-05-04
Fix bug 1450505: [FTL] Re-enable Tab to copy from helpers

mathjazz:
In complex FTL strings, multiple fields are present, and we use Tab to
navigate between them, as this is the browser default shortcut. We also
use Tab to navigate between helper suggestions (history, machinery,
locales), as many other CAT tools do.

This patch re-enables Tab for navigating helper suggestions for complex
FTL strings when the focus is in the last field. After the suggestion
is copied, we keep the focus in the last field to make moving to next
suggestion with subsequent Tabs easier.

@jotes r?

### DIFF
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -1883,8 +1883,12 @@ var Pontoon = (function (my) {
         // Tab: Select suggestions
         if (!$('.menu').is(':visible') && key === 9 && !e.ctrlKey) {
 
-          // Rich FTL editor with complex message: ignore tab key
-          if (self.fluent.isFTLEditorEnabled() && self.fluent.isComplexFTL()) {
+          // Rich FTL editor with complex message and last element not focused: ignore tab key
+          if (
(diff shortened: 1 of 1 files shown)


pontoon/base/static/js/translate.js
```diff
@@ -1883,8 +1883,12 @@ var Pontoon = (function (my) {
         // Tab: Select suggestions
         if (!$('.menu').is(':visible') && key === 9 && !e.ctrlKey) {
 
-          // Rich FTL editor with complex message: ignore tab key
-          if (self.fluent.isFTLEditorEnabled() && self.fluent.isComplexFTL()) {
+          // Rich FTL editor with complex message and last element not focused: ignore tab key
+          if (
+            self.fluent.isFTLEditorEnabled() &&
+            self.fluent.isComplexFTL() &&
+            !$('#editor textarea:visible:last').is(":focus")
```
jotes: [nit] `'focus'`?

PR #948, opened 2018-05-07
Bug bug 1459583: Fix failing migration

mathjazz:
By re-implementing the Model methods in the migration and using frozen models instead of actual ones.

@jotes r?


mathjazz:
We might reuse code in this migration, that's why I commented out `print` statemens instead of removing them.

PR #949, opened 2018-05-07
Fix bug 1459601: Enable download TM for All Projects

mathjazz:
Download TM is now enabled for All Projects.


mathjazz:
@flodolo Could you give it a try on stage?

flodolo:
It seems to work correctly. It's also a 16MB file, with almost half a million lines for Italian. Is that going to create any performance/band issue?

mathjazz:
Thanks!

It doesn't have a noticeable impact on the DB. The amount of such requests is also not expected to be high.

PR #950, opened 2018-05-08
Fix bug 1459596: Fix approval in suggest mode

mathjazz:
When Make Suggestions is turned on and you approve a suggestion that
returns warnings and click `Save anyway`, suggestions needs to get
approved. Previously, `Same translation already exists` message was
getting displayed.


mathjazz:
@jotes r?

@flodolo This is now on stage, could you take a look?

Testing instructions:
https://bugzilla.mozilla.org/show_bug.cgi?id=1459596#c0

flodolo:
Seems to work fine on stage 👍

PR #951, opened 2018-05-08
Fix bug 1269427: Fix title of Approve Anyway button

mathjazz:
When you approve a suggestion and a list of failed checks is displayed, the title of the button to override checks is called Save anyway or Suggest anyway (if you are in suggestions mode).

That misleading, especially in case of Suggest anyway. We change both title to Approve anyway now and color the button green.

### DIFF
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -2587,7 +2587,7 @@ var Pontoon = (function (my) {
     renderFailedChecks: function(failedChecks, messageOnly) {
       if (!messageOnly) {
         $('#save-anyway')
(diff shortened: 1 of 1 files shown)


pontoon/base/static/js/translate.js
```diff
@@ -2586,7 +2586,9 @@ var Pontoon = (function (my) {
      */
     renderFailedChecks: function(failedChecks, messageOnly) {
       if (!messageOnly) {
-        $('#save-anyway').show();
+        $('#save-anyway')
+          .toggleClass('approve', this.approvedNotSubmitted || false)
```
adngdb: `x or false` is always the same a `x`, so `|| false` is useless here.

PR #952, opened 2018-05-09
Fix bug 1460073: Ignore Android DTD escaping checks

mathjazz:
Pontoon takes care of escaping quotes and apostrophes on serialization.
See: pontoon.sync.formats.silme.SilmeResource:escape_quotes()

@jotes r?

### DIFF
--- a/pontoon/checks/libraries/compare_locales.py
+++ b/pontoon/checks/libraries/compare_locales.py
@@ -148,8 +148,7 @@ def run_checks(entity, locale_code, string):
     )
 
     checker = getChecker(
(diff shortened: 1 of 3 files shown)


Pike:
Actually, there's another check that compare-locales runs, and that's for unicode escape sequences, https://hg.mozilla.org/l10n/compare-locales/file/tip/compare_locales/checks/dtd.py#l204. Do we have something in pontoon that enforces those?

mathjazz:
We don't.

So the right solution here is not to remove the `android-dtd` checks, but instead use `escape_quotes()` in `cast_to_compare_locales()`?

jotes:
👯‍♂️

PR #953, opened 2018-05-09
Fix bug 1460081: Group compare-locales messages

mathjazz:
@jotes r?


jotes:
works Messages are grouped - as expected :D 
What do you think about adding test for that behaviour? (I can do that as in a follow-up).

PR #956, opened 2018-05-10
Fix bug 1453319: [FTL] Add support for missing expression types

mathjazz:
This is similar to [bug 1437857](https://bugzilla.mozilla.org/show_bug.cgi?id=1437857), where we added support for `CallExpressions`: https://github.com/mozilla/pontoon/pull/910/commits/2edbadd13d9f305b7cc80fd0901bdb83dc7f6a8f.

Now we're adding support for 4 remaining `*Expression` types.

### DIFF
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -8,8 +8,8 @@ var Pontoon = (function (my) {
   /*
    * Is ast element of type that can be presented as a simple string:
    * - TextElement
-   * - Placeable with expression type CallExpression, ExternalArgument
-   *   or MessageReference
+   * - Placeable with expression type CallExpression, StringExpression,
+   *   ExternalArgument or MessageReference
    */
   function isSimpleElement(element) {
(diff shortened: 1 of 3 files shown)


pontoon/base/static/js/fluent_interface.js
```diff
@@ -339,9 +343,18 @@ var Pontoon = (function (my) {
...
+        else if (
+          [
+            'CallExpression',
+            'StringExpression',
+            'NumberExpression',
+            'VariantExpression',
+            'AttributeExpression',
+          ].indexOf(element.expression.type) >= 0
+        ) {
+          var title = element.expression.type.split('Expression')[0] + ' Expression';
```
adngdb: Interestingly, I would have gone for a `replace` here instead. 

```javascript
var title = element.expression.type.replace('Expression', ' Expression');
```

I think it's slightly more readable, as it expressed better what the code is doing. But it's a nit, so your call!

mathjazz: True. I'll update.

PR #957, opened 2018-05-10
Fix bug 1460597: Approving suggestions not working

mathjazz:
When fixing bug 1453879, we replaced the `click` event with `mouseover`,
but forgot to update the trigger. That caused the translation to be
treated as submitted instead of approved, which resulted in the
`Same translation already exists` erro message instead of approval.


MikkCZ:
Thank you @mathjazz . I have verified it works in staging.

PR #958, opened 2018-05-10
Fix bug 1269427: toggleClass state must be boolean

mathjazz:


mathjazz:
@MikkCZ r?

https://mozilla-pontoon-staging.herokuapp.com/cs/firefox/all-resources/?string=175507

MikkCZ:
Thank you. It works in staging.

As for the code, I personally prefer if boolean variables like `approvedNotSubmitted` are named `is...` or `approvingNotSubmitted` in this case, and are never `null`, but I am not an expert for Python code style.

mathjazz:
Good point! Updated the variable name.

We can't be always sure the variable is defined, so it helps to convert to boolean.

pontoon/base/static/js/translate.js
```diff
@@ -2608,7 +2608,7 @@ var Pontoon = (function (my) {
     renderFailedChecks: function(failedChecks, messageOnly) {
       if (!messageOnly) {
         $('#save-anyway')
-          .toggleClass('approve', this.approvedNotSubmitted)
+          .toggleClass('approve', this.isApprovedNotSubmitted || false)
```
adngdb: Instead of using this confusing `|| false`, if your goal is to ensure you pass a boolean (which I don't understand, JS is a weakly typed language, you shouldn't need to bother doing that), you should instead use either `Boolean(myVar)` or `!!myVar`. I favor the former, which is in my opinion a lot clearer about the intention.

mathjazz: Good call.

Here's why it needs to be a `Boolean` (not just truthy/falsy) value:
http://api.jquery.com/toggleclass/#toggleClass-className-state

adngdb: Oh, OK, that's surprising! In that case, I would really favor using `Boolean(myVar)`.

PR #959, opened 2018-05-11
Fix bug 1455191: [FTL] Convert TM strings to simple strings

mathjazz:
Test case:
https://mozilla-pontoon-staging.herokuapp.com/de/common-voice/all-resources/?string=178208

### DIFF
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -382,7 +382,7 @@ var Pontoon = (function (my) {
     $('#translation-length, #copy').toggle(show);
 
     if ($('#translation-length').is(':visible')) {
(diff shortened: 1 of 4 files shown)


mathjazz:
First example doesn't return anything, because with this patch applied, it hits 501:
https://github.com/mathjazz/pontoon/blob/c6f01684403de915e8ffed1889cd909db25b63a5/pontoon/machinery/views.py#L70

The second example works as expected, except for the percentage (should be 100%). That's because in FTL we store entire messages as TM source and target, which includes the key. And the key is different for different strings. Add a [comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1453341#c1) in Bugzilla. Thanks for bringing this up!

I updated the comment for `Pontoon.isPluralized()`.

PR #960, opened 2018-05-13
Fix bug 1461199 - Translation length check is able to correctly count…

jotes:
@mathjazz could you review?

### DIFF
--- a/pontoon/checks/libraries/pontoon.py
+++ b/pontoon/checks/libraries/pontoon.py
@@ -1,6 +1,9 @@
 from __future__ import absolute_import
 
+import HTMLParser
 import re
+
+import bleach
 
 from collections import defaultdict
 from fluent.syntax import FluentParser, ast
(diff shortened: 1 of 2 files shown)


pontoon/checks/libraries/pontoon.py
```diff
@@ -1,11 +1,13 @@
...
 
+import HTMLParser
 import re
 
+import bleach
+
 from collections import defaultdict
 from fluent.syntax import FluentParser, ast
 
-
```
mathjazz: I would keep this blank line. As you can check in other files, we usually split imports and global variable definitions with two blank lines.

pontoon/checks/libraries/pontoon.py
```diff
@@ -32,9 +34,19 @@ def run_checks(entity, string):
...
+        html_parser.unescape(
+            bleach.clean(
+                string,
+                strip=True,
+                tags=()
+            )
+        )
+    )
 
     # Prevent translations exceeding the given length limit
```
mathjazz: This comment should indicate the entire max length check. So it should be moved above the following line:
```
max_length = get_max_length(entity.comment)
```

jotes:
@mathjazz fixed

mathjazz:
Thanks!

PR #961, opened 2018-05-14
Fix bug 1398339: Add Releasing the queue to docs

mathjazz:
A queue [got stuck](https://pontoon.mozilla.org/sync/log/51004/) after a long time. I believe nobody has been fixing this except for me and @jotes.

So let's document it.

### DIFF
--- a/docs/admin/maintenance.rst
+++ b/docs/admin/maintenance.rst
@@ -22,7 +22,8 @@ Heroku-hosted instance of RabbitMQ:
 Releasing the queue
 -------------------
 If queue gets stuck, tasks don't make it to the worker until manual
-intervention.
+intervention. You can fix this by running the following commands from your
+local development environment.
 
 First, you need to purge the queue:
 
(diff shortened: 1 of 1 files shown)


docs/admin/maintenance.rst
```diff
@@ -18,3 +18,24 @@ Heroku-hosted instance of RabbitMQ:
...
+   celery amqp --broker=`heroku config:get RABBITMQ_URL --app=my-app-name`
+   # Replace my-queue-name with your queue's name (e.g. celery).
+   1> queue.purge my-queue-name
+
+Finally, you need to simply access the worker:
+
+.. code-block:: bash
+
+   # Replace my-app-name with your Heroku app's name.
+   celery worker --broker=`heroku config:get RABBITMQ_URL --app=my-app-name`
```
adngdb: It's unclear where you need to run those. I assume you need to ssh into your heroku instance? Which one? How is that done? 

Note that this is probably outside the scope of this work, so feel free to leave it out.

mathjazz: Updated.

adngdb: Oh nice I wouldn't have guessed you'd run that from the dev env! Glad I asked! :D

PR #963, opened 2018-05-14
[WIP] Fix bug 1458344. Add models for checks module

jotes:
TODO:
* [x] - discussion about models
* [x] - documentation
* [x] - tests

Hi @mathjazz, 
As always, it took a little longer than I expected ;] This patch currently covers saving warnings from the UI.
Could you review the structure of proposed models?


pontoon/checks/models.py
```diff
@@ -0,0 +1,39 @@
...
+            ('tt', 'translate-toolkit'),
+            ('cl', 'compare-locales'),
+        ),
+        db_index=True
+    )
+    message = m.TextField()
+
+    class Meta:
+        abstract = True
+        ordering = ('library', 'message')
```
mathjazz: Looking at [bug 1436985](https://bugzilla.mozilla.org/show_bug.cgi?id=1436985) I'm a little concerned about performance implications. Shall we move ordering to the `QuerySet` instead?

Note that ths is just a guess and could be a premature optimization. It might be better to leave this unchanged and just add a comment. In the final stage of implementing checks we'll see how much data do we have and how does it behave and then we can work on optimizations.

pontoon/base/views.py
```diff
@@ -704,6 +711,7 @@ def update_translation(request):
             t.approved_date = now
 
         t.save()
+        save_failed_checks(t, checks)
```
mathjazz: There's another instance of `t.save()` in this method, where we approve and/or unfuzzy an existing translation. We should add `save_failed_checks(t, checks)` there, too.

pontoon/checks/libraries/__init__.py
```diff
@@ -76,7 +74,14 @@ def run_checks(
     )
 
     checks.update(pontoon_checks)
+    return checks
+
+
+def failed_checks_response(checks, ignore_warnings):
+    """
```
mathjazz: Looks like this blank line can be removed.

pontoon/checks/utils.py
```diff
@@ -0,0 +1,30 @@
+from pontoon.base.models import Translation
+from pontoon.checks.models import (
+    Warning,
+    Error
+)
+
+def save_failed_checks(translation, checks):
+    """
+
```
mathjazz: This looks like a good opportunity to document the function. :)

pontoon/checks/utils.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+    warnings = []
+    errors = []
+    for check_group, failed_checks in checks.items():
+        library = (
+            check_group
+            .replace('Warnings', '')
+            .replace('Errors', '')
+        )
+        severity_cls, messages_list = (Error, errors) if check_group.endswith('Errors') else (Warning, warnings)
```
mathjazz: This line is too long.

pontoon/checks/utils.py
```diff
@@ -0,0 +1,30 @@
...
+    Error
+)
+
+def save_failed_checks(translation, checks):
+    """
+
+    """
+    warnings = []
+    errors = []
+    for check_group, failed_checks in checks.items():
```
mathjazz: `failed_checks` -> `messages`

pontoon/base/views.py
```diff
@@ -655,6 +660,10 @@ def update_translation(request):
                 t.fuzzy = False
 
             t.save()
+            t.warnings.clear()
+            t.errors.clear()
+
+            save_failed_checks(t, failed_checks)
```
mathjazz: This is confusing me. In `save_failed_checks()` we [delete](https://github.com/mozilla/pontoon/pull/963/files#diff-8c894a1d32099aa55fbfe944f8d1ac65R32) translation warnings and errors, why would we need to disassociate them first here?

Also, this call to `save_failed_checks()` is the only one where we might have existig warnings and errors (because translation exists), so it makes more sense to move the logic here.

PR #964, opened 2018-05-15
Bug 1461630 - Set SameSite attributes for `sessionid` and `csrftoken`…

jotes:
… cookies in the custom middleware.

Django 1.11.x won't receive ability to set SameSite attribute in cookies.
That will be possible with Django 2.0, but that also would require to migrate Pontoon into Python3 (and that will take some time).

As a temporary solution Pontoon will use CookiesSameSiteStrict middleware which will be removed after Pontoon will migrate into Django 2.0.

@mathjazz @psiinon could you review/check  if the whole approach is acceptable?

### DIFF
--- a/pontoon/settings/base.py
+++ b/pontoon/settings/base.py
@@ -152,6 +152,7 @@ INSTALLED_APPS = (
 BLOCKED_IPS = os.environ.get('BLOCKED_IPS', '').split(',')
 
 MIDDLEWARE_CLASSES = (
+    'django_cookies_samesite.middleware.CookiesSameSite',
     'django.middleware.gzip.GZipMiddleware',
     'sslify.middleware.SSLifyMiddleware',
     'pontoon.base.middleware.RaygunExceptionMiddleware',
@@ -217,6 +218,8 @@ TEMPLATES = [
         }
     },
 ]
+
+SESSION_COOKIE_SAMESITE = 'Strict'
 
 AUTHENTICATION_BACKENDS = [
     'django.contrib.auth.backends.ModelBackend',
--- a/requirements.txt
+++ b/requirements.txt
@@ -324,3 +324,5 @@ compare-locales==3.2.2 \
 # Required by compare-locales
 pytoml==0.1.14 \
     --hash=sha256:aff69147d436c3ba8c7f3bc1b3f4aa3d7e47d305a495f2631872e6429694aabf
+django-cookies-samesite==0.1.1 \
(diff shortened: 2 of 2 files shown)


pontoon/base/middleware.py
```diff
@@ -48,3 +48,19 @@ def process_request(self, request):
             if user:
                 request.user = user
                 auth.login(request, user)
+
+
+class CookiesSameSiteStrict(object):
+    """
+    SameSite attribute is the new security mechanism that should improve protection against CSRF attacks.
+    Support for SameSite attribute in Cookies is implemented in Django 2.0 and won't be backported to Django 1.11.x
+    This middleware will be obsolete when Pontoon will be ported to Django 2.0 and should as a follow-up action.
```
mathjazz: Looks like a verb is missing at the end of the last sentence.

Also, please make lines shorter than 79 characters.

pontoon/base/middleware.py
```diff
@@ -48,3 +48,20 @@ def process_request(self, request):
...
+    This middleware will be obsolete when Pontoon will be ported to Django 2.0 and
+    should be removed as a follow-up action.
+
+    Related bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1461630
+    """
+    def process_response(self, request, response):
+        for cookie in ('sessionid', 'csrftoken'):
+            if cookie in response.cookies:
+                response.cookies[cookie]._reserved['samesite'] = 'SameSite'
+                response.cookies[cookie]['samesite'] = 'strict'
```
psiinon: 'strict' means that regular links to the application also wont get the session cookie added, which is good for security but not so good for usability. 'lax' is sometimes a better compromise as it still prevents the more common attacks. If 'strick' works for pontoon then great, but being able to choose between them would be even better for reuse.

jezdez:
FWIW, the patch will land only in Django 2.1, not 2.0.

jotes:
@psiinon Hey, could you look again? I updated the PR with the package from here: https://github.com/jotes/django-cookies-samesite

psiinon:
@jotes That looks great :)

mathjazz:
@psiinon Is this r+?

I deployed the patch to stage:

psiinon:
Sorry, yes, its an r+ from me.
https://mozilla-pontoon-staging.herokuapp.com/ looks good :)

mathjazz:
Thanks!

PR #969, opened 2018-05-18
Fix bug 1452647: Add Tags tab to the Localization dashboard

mathjazz:
In https://github.com/mozilla/pontoon/commit/5c4152d2a33be7a0c28d766b8da922a9d1fd35f6 we landed support for the _Tags_ tab in the [project dashboard](https://mozilla-pontoon-staging.herokuapp.com/projects/firefox/tags/), which contains a list of tags and their stats for the project. This patch bring the _Tags_ tab to the [localization dashboard](https://mozilla-pontoon-staging.herokuapp.com/sl/firefox/tags/), which contains a list of tags for the project and their stats for the localization (project-locale).

[First commit](https://github.com/mozilla/pontoon/commit/43fc906227d8442d533b736b7e786553a8c0cf60) sets the stage by generalizing the code for Tags tab on the Project dashboard. The [second one](https://github.com/mozilla/pontoon/commit/4aeaa0b538e7012689bc067ce4c779687b56124a) is an actual bugfix. It will be easier to review them separtely.

Check out the [docs](https://github.com/mozilla-l10n/localizer-documentation/pull/120/) for more information about tags.

PR #970, opened 2018-05-21
Fix bug 1463073: Show tag priority in filter menu

mathjazz:
This (alongside #969) is hopefully the last step before shipping Tags. You can test it on stage:

@jotes r?

### DIFF
--- a/pontoon/base/templates/widgets/filter_item.html
+++ b/pontoon/base/templates/widgets/filter_item.html
@@ -1,9 +1,16 @@
-{% macro item(type, title, count=False) %}
+{% macro item(type, title, count=False, priority=None) %}
 <li class="{{ type }}" data-type="{{ type }}">
   <span class="status fa"></span>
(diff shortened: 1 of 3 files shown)


pontoon/base/templates/widgets/filter_item.html
```diff
@@ -1,9 +1,16 @@
-{% macro item(type, title, count=False) %}
+{% macro item(type, title, count=False, priority=None) %}
 <li class="{{ type }}" data-type="{{ type }}">
   <span class="status fa"></span>
   <span class="title">{{ title }}</span>
   {% if count %}
-    <span class="count"></span>
+  <span class="count"></span>
+  {% endif %}
+  {% if priority %}
```
jotes: Currently, template gets only tags with set priority and because of that this condition will be always true.

mathjazz: It actually gets all the tags. Priority is an optional field.

PR #971, opened 2018-05-22
Fix bug 1463364: Always generate tag slug

mathjazz:
Tag slug is only generated for the first tag. If you add more tags, it's not.


mathjazz:
@flodolo Could you review it on stage?
https://mozilla-pontoon-staging.herokuapp.com/admin/projects/firefox/

flodolo:
Seems to be working fine on stage 👍

PR #972, opened 2018-05-25
Fix bug 1455246 - Improve pytest environment.

adngdb:
This is a heavy change to most of our tests, and to our test infrastructure. Here's a non-exhaustive list of all the things it does:

- move all tests back into the `pontoon` folder;
- run all tests (including the remaining django tests) with `pytest`;
- enable using `--reuse-db`;
- create a `test` app for shared fixtures, infra and setup;
- refactor factories so that they are used in fixtures, and they are not exposed as fixtures anymore;
- simplify fixtures wherever possible;
- do not create any data by default in tests, all data is created by tests on demand;
- rename fixtures from a `fixture0` to a `fixture_a` format (which is more pythonic);
- move fixtures closer to actual tests wherever possible;
- change code style in files where I made changes.

I also hope this makes a lot of things simpler to understand and discover. It might not be where I want it to be yet, and we could do with some documentation (in readme files, probably) but that can come later.

I've tried to split this in commits that make it as easy to review as possible. All file movements and renaming are isolated in their own commits. This way the other commits should have useful diffs.


.travis.yml
```diff
@@ -38,8 +38,7 @@ script:
   - ./node_modules/.bin/webpack
   - python manage.py collectstatic -v0 --noinput
   - npm test
-  - python manage.py test --with-coverage
-  - py.test --cov-append --cov-report=term --cov=. -v
+  - py.test --cov-append --cov-report=term --cov=. -v --create-db --migrations
```
jotes: <3

pontoon/base/tests/forms/test_permission_log.py
```diff
@@ -1,77 +1,20 @@
-import factory
 import pytest
```
mathjazz: While evaluating previous changesets, I already wrote a comment about the need of reusing fixtures. Now, I can delete it. :)

pontoon/tags/tests/conftest.py
```diff
@@ -0,0 +1 @@
+from fixtures import *  # noqa
```
Pike: Make this an explicit relative import via `.fixtures`? To prevent that from happening, maybe also a `from __future__ import absolute_imports` ?

PR #967, opened 2018-05-17
Fix bug 1462419: Restore text selection in helpers

mathjazz:
This is a regression from [bug 1453879](https://bugzilla.mozilla.org/show_bug.cgi?id=1453879).

PR #983, opened 2018-06-24
Fix bug 1470742: Restore clickable contributor links in History tab

mathjazz:
We introduced the removed line while fixing bug 1453879: https://github.com/mozilla/pontoon/pull/944.

After we fixed bug 1462419, the line is no longer needed: https://github.com/mozilla/pontoon/pull/967.

### DIFF
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -293,7 +293,7 @@ var Pontoon = (function (my) {
                     '">' +
                     '<div class="info">' +
                       ((!this.uid) ? '<span>' + this.user + '</span>' :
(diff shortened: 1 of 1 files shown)


mathjazz:
Patch is deployed to stage:

The following functionality is affected / should be tested in History/Machinery/Locales tabs:
* Click on translation author (History tab).
* Click on service name (Machinery tab).
* Click on locale name (Locales tab).
* Copy entry (suggestion) to the translation field.
* Copy entry (suggestion) to the focused translation field for FTL strings.
* (Un)Approve/(Un)Reject.
* Show Diff.
* Select part of an entry (suggestion).

mathjazz:
Good point!

PR #984, opened 2018-06-25
[Experiment] Bug 1344457: Redesign main menu in translate view

mathjazz:
This is a quick attempt at improving the main menu in the translate view by making it work more like it looks like - a breadcrumbs menu.

That means the first two "menus" (locale and project) are no longer menus. Instead, they take you to the locale page (e.g. `/de/`) and the localization page (e.g. `/de/firefox/`) of the current translate view.

The resources menu looks more or less the same, but since it's the only menu now, it takes action immediatelly after you make a selection, so the `Go` button is gone.

The main motivation for the change is the hypothesis (partly proved by analytics) that localizers mostly leave the translate view for their team page or localization page, but rarely switch locales, projects or go to `/teams/` and `/projects/` pages.

TODO:
- [ ] The behaviour changed, but the look didn't. It will take a while before people get used to the new UX. To help with that, we should take a bolder step in differentiating links from dropdowns (add a caret to the dropdown and consider changing the background color of links).
- [ ] Consider blocking the release on [new homepage](https://bugzilla.mozilla.org/show_bug.cgi?id=1378471) to prevent the inability of getting from the homepage to dashboards.
- [ ] Add tooltips to all three items in the main menu.

### DIFF
--- a/pontoon/base/static/css/style.css
+++ b/pontoon/base/static/css/style.css
@@ -573,6 +573,7 @@ tfoot td a {
   display: none;
 }
 
+.project.select.all-projects > .button.breadcrumbs:after,
 .part.select > .button.breadcrumbs:after {
   display: none;
 }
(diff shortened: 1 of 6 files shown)


flodolo:
From a quick look, this seems more intuitive than the current system 👍 

There is one possible confusion: clicking the resource name shows a list of resources, clicking the first two items brings you to a different page (BTW, they definitely need a tooltip). I wonder if the last element of the breadcrumbs should look even more different from the other two, e.g. slightly different background color.

mathjazz:
Adding a few more comments from IRC:

@TheoChevalier
I like it, I think I’ll finally stop constantly editing URLs.  It may take some time before people realize the different behavior between the tree elements. Maybe adding an arrow to the file selector would help differentiate even further?

c8b912fd-6a84-4f68-abe1-4c9e92a107c8

@VishalCR7
This also has same behavior on landing page https://mozilla-pontoon-staging.herokuapp.com/. And it prevents me from reaching All teams/projects dashboard directly from home page.

@jotes
I thought that the staging is broken :D I'm too accustomed to the previous design. Anyway, I think it's the right move.

mathjazz:
Closing for inactivity.

PR #986, opened 2018-06-25
Fix bug 1470337: De-optimize calculate_stats

mathjazz:
To prevent IntegrityError.


mathjazz:
Updated with a note in the help text.

adngdb:
r+ but pylama is failing because you have imports in that file that are not used anymore. https://travis-ci.org/mozilla/pontoon/builds/396919866#L690

mathjazz:
Ahhhhh.

PR #987, opened 2018-06-25
Fix bug 1455099: [FTL] Make selectors with siblings more readable using indentation

mathjazz:
Details in the bug:
https://bugzilla.mozilla.org/show_bug.cgi?id=1455099

### DIFF
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -670,6 +670,13 @@ var Pontoon = (function (my) {
           $('#ftl-area textarea.value').keyup();
         }
 
+        // Indent selector variants if preceeded by other elements
+        $('.ftl-area section > ul:not(.template) > li[data-expression]').each(function(i, node) {
+          if ($(node).prevAll(':visible:not([data-expression])').length) {
+            $(node).addClass('indented');
+          }
+        });
+
         // Ignore editing for anonymous users
         if (!Pontoon.user.id) {
           $('#ftl-area textarea').prop('readonly', true);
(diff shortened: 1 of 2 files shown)


adngdb:
Currently it looks like this:



I dislike the misalignment on the right of those boxes, and I think something like this looks better: 



What do you think? 

To do it, remove the current CSS rule you have and add instead: 

```css
.ftl-area section .indented .id {
    margin-left: 10px;
    width: calc(25% - 15px);
}
```

pontoon/base/static/js/fluent_interface.js
```diff
@@ -670,6 +678,13 @@ var Pontoon = (function (my) {
           $('#ftl-area textarea.value').keyup();
         }
 
+        // Indent selector variants if preceeded by other elements
+        $('.ftl-area section > ul:not(.template) > li[data-expression]').each(function(i, node) {
+          if ($(node).prevAll(':visible:not([data-expression])').length) {
+            $(node).addClass('indented');
+          }
+        });
```
adngdb: Actually, I'm wondering if this is the right solution. Why did you decide to add that class afterwards using selectors rather than during the creation of the elements?

flodolo:
> So I'm undecided and calling in @flodolo to pull the trigger. :)

As someone who doesn't feel happy around misalignments, I'm all for Adrian's proposal, with the input fields aligned.

pontoon/base/static/js/fluent_interface.js
adngdb: Yep, I figured there was a complexity issue involved, just wanted to make sure. That's fine by me!

PR #988, opened 2018-06-26
Fix bug 1447231: [FTL] Use simple string UI for messages with 1 simple attribute

mathjazz:
This patch adds a special case for FTL messages with 1 simple attribute and no value to:
* source string panel
* editor
* serialization

A test case:
https://github.com/mozilla-l10n/pontoon-ftl/blob/master/en-US/demo.ftl#L10

Diff looks bigger than it could, due to change indentation. I suggest reviewing on a commit-by-commit basis.

### DIFF
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -693,25 +693,45 @@ var Pontoon = (function (my) {
 
         // Attributes
         if (attributesTree.length) {
-          attributesTree.forEach(function (attr) {
-            var id = attr.id.name;
+          // Simple single-attribute string: only attribute
+          if (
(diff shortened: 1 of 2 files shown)


pontoon/base/static/js/fluent_interface.js
```diff
@@ -111,6 +111,23 @@ var Pontoon = (function (my) {
     return false;
   }
 
+  /*
+   * Is ast of a message without value and with 1 attribute,
+   * which only contains simple elements?
```
adngdb: Nit, I think those comments are better when they start with `Return`. In this case, something like `Return true when an AST has no value and a single attribute with only simple elements`, for example.

PR #990, opened 2018-06-27
Fix bug 1454003: [FTL] Add ability to manage Term attributes

mathjazz:
Unlike regular Messages, Terms are allowed to contain custom attributes (i.e. attributes not present in source string).

This patch introduces the "Add attribute" button, together with the ability to remove custom attributes and edit their IDs.

screen shot 2018-06-27 at 12 09 25

To learn more about Terms and attributes, see:
https://projectfluent.org/fluent/guide/terms.html

The doc also contains a few examples, but you can also use the following file for testing:
https://github.com/mozilla-l10n/pontoon-ftl/blob/master/en-US/brand.ftl

### DIFF
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -288,7 +288,7 @@ var Pontoon = (function (my) {
               item.value.elements,
               isPluralElement(element),
               isTranslated,
-              isCustomAttribute
+              isCustomAttribute,
             );
           });
         }
@@ -952,7 +952,7 @@ var Pontoon = (function (my) {
               if (showFTLEditor) {
                 var isRichEditorSupported = self.renderEditor({
                   pk: translated, // An indicator that the string is translated
-                  string: translation
+                  string: translation,
                 });
 
                 // Rich FTL editor does not support the translation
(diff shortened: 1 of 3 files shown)


pontoon/base/static/js/fluent_interface.js
```diff
@@ -252,7 +262,8 @@ var Pontoon = (function (my) {
               item.key.value || item.key.name,
               item.value.elements,
               isPluralElement(element),
-              isTranslated
+              isTranslated,
+              isCustomAttribute
```
adngdb: Nit: you can add trailing commas in function arguments, this way next time you make a change the diff will only show one line instead of three.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -826,7 +855,7 @@ var Pontoon = (function (my) {
               if (showFTLEditor) {
                 var isRichEditorSupported = self.renderEditor({
                   pk: translated, // An indicator that the string is translated
-                  string: translation,
+                  string: translation
```
adngdb: I'd argue against removing trailing commas, they make diffs better. I've always considered them a good practice in both Python and JS.

PR #993, opened 2018-06-27
Fix bug 1471534: Truncate missing strings with suggestions

mathjazz:
1st commit removes the unused code.

2nd commit fixes the regression.

### DIFF
--- a/pontoon/base/static/css/translate.css
+++ b/pontoon/base/static/css/translate.css
@@ -1166,6 +1166,7 @@ body > header aside p {
 }
 
 #entitylist > .wrapper > ul > li.translated p span.source-string,
+#entitylist > .wrapper > ul > li.missing.has-translations p span.source-string,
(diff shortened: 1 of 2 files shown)


mathjazz:
Deployed to stage. Things look particularly bad if strings contain placeables.

Prod:
https://pontoon.mozilla.org/sl/common-voice/all-resources/?status=unreviewed&string=179060

Stage:
https://mozilla-pontoon-staging.herokuapp.com/sl/common-voice/all-resources/?status=unreviewed&string=179048

/cc @upwinxp

flodolo:
Ugh, I even checked those, and didn't realize the long strings had a suggestion :-\

mathjazz:
While waking up I realized we don't change the truncation if you submit a suggestion to a missing string without suggestions. Now confirmed it's indeed happening. Fixing...

mathjazz:
Fix deployed to stage. Sorry for the mess.

PR #996, opened 2018-07-02
Fix bug 1472793: Error on non-localizable entries

mathjazz:
For more details, please see the bug description:
https://bugzilla.mozilla.org/show_bug.cgi?id=1472793


mathjazz:
Deployed to stage:
https://mozilla-pontoon-staging.herokuapp.com/pl/firefox-screenshots/server.ftl/?string=168513

jotes:
Hi,
I tested this on staging and your patch works okay (I couldn't submit "## Comment" as a suggestion). Could you add a test-case for this behavior?
r+ from me.

PR #997, opened 2018-07-03
Fix bug 1472995: Dynamically update tags in the filters menu

mathjazz:
Steps to reproduce:

1. Go to https://pontoon.mozilla.org/sl/firefox/all-resources/.
2. Open the filter menu.
3. Observe Tags.
4. Select AMO from the Project menu and click Go.
5. Open the filter menu.
6. Observe Tags.

Expected result: 
There are no tags in step 6.

Actual result:
Tags from step 3 (belonging to Firefox) are displayed in step 6 (when AMO is being translated).

### DIFF
--- a/pontoon/tags/tests/test_models.py
+++ b/pontoon/tags/tests/test_models.py
@@ -1,5 +1,3 @@
-from __future__ import absolute_import
-
 import pytest
 
 from pontoon.tags.models import Tag
--- a/pontoon/tags/tests/test_models.py
+++ b/pontoon/tags/tests/test_models.py
@@ -1,0 +1,37 @@
+from __future__ import absolute_import
+
+import pytest
+
+from pontoon.tags.models import Tag
+from pontoon.test.factories import TagFactory
+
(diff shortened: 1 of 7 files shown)


adngdb:
This is looking really good!

Have you looked at whether `TagTools` was still used? And there should be at least one simple test for the new Python code. :-)

mathjazz:
Yeah, we still need the `TagsTool` (on dashboards). But it's a fair question - as we observe how the tags are [being used](https://bugzilla.mozilla.org/show_bug.cgi?id=1464802#c5), we'll likely need to make UI refactorings, that could lead into code refactorings, too.

Good point about tests! I should test the changes in `pontoon.tags.models`. I'll follow up.

pontoon/tags/tests/test_models.py
```diff
@@ -0,0 +1,37 @@
+from __future__ import absolute_import
```
adngdb: Why do you need this?

PR #998, opened 2018-07-04
Fix bug 1473372: Make Docker setup default in developer docs

mathjazz:
It's time.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,7 +27,8 @@ coverage/
 
 # Docker
 .docker-build
-docker/config/webapp.env*
+docker/config/webapp.env
+docker/config/webapp.env-e
 
 # Webpack
 /assets/
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,3 @@
-# Local settings
-hooks/php/local-settings.php
-settings_local.py
-settings/local.py
-*/settings/local.py
 *.py[co]
 *.sw[po]
 .coverage
@@ -11,7 +6,7 @@ docs/_gh-pages
 build.py
 build
 .DS_Store
-node_modules/
+node_modules
 *-min.css
 *-min.js
 *-all.js
@@ -30,9 +25,9 @@ venv
 .cache/
 coverage/
 
-# For docker
+# Docker
 .docker-build
-docker/config/webapp.env
+docker/config/webapp.env*
 
 # Webpack
 /assets/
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -30,9 +25,9 @@ venv
 .cache/
 coverage/
 
-# For docker
+# Docker
 .docker-build
-docker/config/webapp.env
+docker/config/webapp.env*
```
adngdb: I believe that one is incorrect. We have a `webapp.env.template` that we want to track in git. Why did you change this rule?

CONTRIBUTING.rst
```diff
@@ -24,6 +24,20 @@ If you're not familiar with `Docker <https://docs.docker.com/>`_ and
...
+
+By default, you will have default data loaded for only the Pontoon Intro project.
+If you have a database dump, you can load it into your PostgreSQL database by running::
+
+  $ make loaddb DB_DUMP_FILE=path/to/my/dump
+
+Note that your database container needs to be running while you do that. You
+can start just the postgresql container by runnin::
+
+  $ docker-compose run postgresql
```
adngdb: Might be worth adding `-d` to that command so that the database container runs in the background? I know about it, but others might be frustrated by docker containers showing logs by default.

.gitignore
mathjazz: `docker/config/webapp.env-e` popped up for me.

CONTRIBUTING.rst
```diff
@@ -35,7 +35,7 @@ If you have a database dump, you can load it into your PostgreSQL database by ru
 Note that your database container needs to be running while you do that. You
 can start just the postgresql container by runnin::
 
-  $ docker-compose run postgresql
+  $ docker-compose -d run postgresql
```
adngdb: That's not going to work, the `-d` must be at the end.

.gitignore
adngdb: Huh, OK, that's surprising. Definitely don't see that locally. Can you add specifically `-e` then?

CONTRIBUTING.rst
mathjazz: Whoops!

.gitignore
mathjazz: @adngdb Should I instead list this specific file?

adngdb: Yeah that's what I meant.

mathjazz:
@adngdb One final look, please? :)

adngdb:
Le r+!

PR #1000, opened 2018-07-05
Fix bug 1473575: Fix tags detection for all-projects

mathjazz:
For steps to reproduce, see:
https://bugzilla.mozilla.org/show_bug.cgi?id=1473575#c0

I'll deploy to stage in a bit.


mathjazz:
Stage is now running this branch:

PR #1002, opened 2018-07-07
Bug 1378471: New Pontoon homepage

shkamaru:
The webpage is now a simple 'hello world' and is for the users not logged in. Once the user is logged in it shows the custom_homepage or by default the top_contributed locale.


pontoon/base/templates/home.html
```diff
@@ -0,0 +1,11 @@
+{% extends "base.html" %}
+
+{% block title %}Home Page{% endblock %}
+
+{% block home %}
```
mathjazz: You should override one of the existing blocks (e.g. `middle`) instead of creating a new one. It doesn't make much sense to create page-specific blocks like `home` in a general template.

That would effectivelly mean no changes to the base template are needed.

pontoon/base/views.py
```diff
@@ -58,6 +58,10 @@ def home(request):
     user = request.user
     project = Project.objects.get(id=1)
 
+    # If user is not logged in, the default homepage is set to static webpage
+    if not user.is_authenticated():
+        return render(request, 'home.html')
```
mathjazz: This means we'll still load the all Pontoon Intro page for authenticated users that don't specify a custom homepage. Instead, we want the homepage to be loaded for everybody (regardless if they are logged in or not), unless they set a custom homepage.

To do that, you need to ommit the condition and replace line `return translate...` with this one. Then, you can also remove the lines which set `project`, `locale` and `path`, since they are no longer needed.

node_modules
```diff
@@ -0,0 +1 @@
+/webapp-frontend-deps/node_modules
```
mathjazz: We should get rid of this as part of #998.

pontoon/base/templates/base.html
```diff
@@ -36,9 +36,6 @@
 
       {% include "header.html" %}
 
-      {% block before %}
-      {% endblock %}
-
```
mathjazz: Are you sure this is not needed? If not, please restore it.

shkamaru: Yeah, sorry it was required. It was creating error to select the custom homepage on the settings.

node_modules
adngdb: Right, and you should actually remove this file from your branch please. :) (It's conflicting with my local files. )

pontoon/base/templates/header.html
```diff
@@ -23,6 +23,12 @@
       <div class="right">
         {{ Notifications.menu() }}
 
+        <div class="sign-in-header">
+          {% if not user.is_authenticated() %}
```
mathjazz: I believe you should move this line a line up and wrap the entire `.sign-in-header` block. Otherwise part of the markup stays in regardless of the and the CSS adds unwanted padding.

shkamaru: Sorry, I didn't get you. Which line are you talking about? I removed the extra padding to the notifications icon.

mathjazz: It should look like this:
```jinja2
{% if not user.is_authenticated() %}
<div class="sign-in-header">
    <a id="fxa-sign-in" href="{{ provider_login_url(request) }}" class="button">Sign Up</a>
</div>
{% endif %}
```

pontoon/base/static/css/style.css
```diff
@@ -952,6 +966,27 @@ header nav .right {
   float: right;
 }
 
+header nav .right .sign-in-header {
+  display: inline-block;
+  vertical-align: bottom;
+  font-size: 14px;
+  margin-bottom: 14px;
```
mathjazz: 13px (for proper vertical center-alignment).

pontoon/base/templates/home.html
```diff
@@ -0,0 +1,160 @@
...
+    <section class="snaps section-6" data-section-name="section-6">
+      <div class="footer vertical-align-center flex-direction-col">
+        <div class="title-style">Join us.</div>
+        <p class="content-style">Now that you know what Pontoon is, come on in.</p>
+        <a class="homepage-button primary-button-cta" href="localhost:8000/teams/">
+          <div>Localize Mozilla</div>
+        </a>
+        <p class="content-style">Want to learn more?</p>
+        <a class="homepage-button secondary-button-cta" href="localhost:8000/demo/">
+          <div>Take a tour!</div>
```
mathjazz: Make sure you follow the latest proposal:
https://ibb.co/hAZgK8

The CTAs in the Final section must match the CTAs in the Introduction.

pontoon/base/static/css/style.css
shkamaru: Are you telling me about the margin-bottom? Because I tried that and it was not centrally aligned in the header.

shkamaru:
Regarding
1. I did set width and height of the image in CSS. I am getting single image only. Maybe it has something to do with the OS as in the case of header.

5. The image is different. I will change it.

6. I forgot the links. Will set them. 

Most of the changes which you are seeing is not coming in my webpage. My webpage looks exactly like I designed. I think that most of the placement of elements varied because of change in the OS.

mathjazz:
The page layout should be independent of the OS. For example, if it looks good on Linux, that's not sufficient, because 90% of our users don't use it.

However, I don't think it's the OS that affects the layout, it's mostly the screen resoltution. This is how the Intro section background and the Section 3 layout look for me when I resize the browser window to occupy full screen and when I only use a quarter of the screen:

Full screen:
https://screenshots.firefox.com/0u7lee3mHP04W7dQ/localhost

Quarter of the screen:
https://screenshots.firefox.com/ykf8G5YBOMS89TXh/localhost

I see some `@media` queries in CSS, so that might be a good place to start looking.

shkamaru:
Made all the requires changes except section 3 image, I am still looking for an alternative for that because I don't want to change its design to that of section 4 and if I make it horizontal then it breaks the concept or symmetry being followed in the previous sections of two columns. I have updated the all the rest of the points. 

Screenshot: 1920x1080 -- https://screenshots.firefox.com/qPUIL8uApzZI4kcA/localhost

Screenshot: 1366x768 -- https://screenshots.firefox.com/GMYcEeRwpb5o6l9e/localhost

mathjazz:
Nice work!

Clicking on "Why should I join Mozilla localization?" still appends the "#section-2" hash to the URL.

And, it would be great to replace the image in Section 3 with text + image.

Once this is done, we should deploy to stage and share the link on [Discourse](https://discourse.mozilla.org/t/pontoon-homepage-redesign-ux/29673).

mathjazz:
Bingo!

I noticed two more things that would be nice to fix:
* When you slide from Section 5 (One more thing) to the Final Section, the header background transitions from dark to transparent, which results in an unpleasant blink effect and makes header contents hard to read during scrolling.
* Header: make sure the spacing between the Sign in button and the Profile menu icon is the same as spacing between the Notifications icon (bell) and the Profile menu icon.
* Footer: Links should be vertically aligned. A bit more padding at the bottom would be nice. Footer content should be inside a `.container`, so the width would be the same as in header (980px).

mathjazz:
Another thing I noticed while scrolling from the 1st to the 2nd section: the "Why should I join Mozilla localization?" ribbon and the 2nd section should be connected, instead of being interrupted by the bottom part of the 1st section. That would make scrolling much smoother.

shkamaru:
Hey, in the recent commits I worked mainly on the scrolling.
The jQuery plugin I used for scrolling already had the easing function but it was a default one, so I made changes in that mainly. I have added the link of vishal's tour provided by Vishal only. There has to be a demo project for the tour first, as he described that's why the link was directed to en-GB/demo/database.

pontoon/base/templates/home.html
```diff
@@ -0,0 +1,175 @@
...
+        <a class="homepage-button primary-button-cta" href="/teams/">
+          <div>Start Localizing Now</div>
+        </a>
+        <p class="content-style">Want to learn more?</p>
+        <a class="homepage-button secondary-button-cta" href="/tour/">
+          <div>Take a Tour</div>
+        </a>
+      </div>
+      <div class="homepage-footer flex">
+        <div class="flex-col-4 align-left"><a href="https://mozilla.org/"><img class="moz-logo" src="/static/img/Homepage/moz-logo.svgz"></img></a></div>
```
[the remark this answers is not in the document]
shkamaru: Because it's the one available on the https://mozilla.design/mlogo plus svgz format is compressed svg files which supports graphic data in XML format.

pontoon/base/templates/home.html
```diff
@@ -0,0 +1,175 @@
...
+    <section class="snaps section-1 relative" data-section-name="section-1">
+      <div class="introduction neg-margin--large">
+        <div class="title-style">A place to localize Mozilla, fast and easy.</div>
+        <div class="content-style">Join a global community of thousands in making Firefox and various Mozilla products speak to over a hundred of cultures in their native language.</div>
+        <div class="flex flex--wrap">
+          <a class="flex-col-full homepage-button primary-button-cta" href="/teams/">
+            <div>Start Localizing Now</div>
+          </a>
+          <p class="flex-col-full">or</p>
+          <a class="flex-col-full homepage-button secondary-button-cta" href="/tour/">
```
VishalCR7: Hi @shkamaru  please use `href="{{ url('pontoon.translate.locale.agnostic', 'projects', 'demo', 'all-resources') }}"` as a link to the tour as suggested by matjaz in the PR

mathjazz:
I noticed that after the recent updates, the "Why should I join" section gets cut off.
screen shot 2018-08-03 at 16 01 20

When I click on the first section bullet, it's better, but still not ideal (I'd expect it to be centered):
screen shot 2018-08-03 at 16 01 25

shkamaru:
I am having some trouble with that. Actually the plugin sets the intro-footer based on the screen size, so sometimes it gets set as the design and sometimes it gets cut. I have been looking at this but couldn't find a solution. 
If I set the text in centre and accordingly adjust the height of the intro footer I get the problem where the background image is exceeding. So, I tried to hide the exceeding background image with the increase in height of the intro-footer and make the gave the text a bit of padding from the top. Which was working as  a solution on most of the screens as because the media queries I used had the sizing set to rem.

I have now changed it to the vh unit and thus it is placed as per the design. But the problem is that sometimes still exists that it sometimes loads the intro-footer mostly and sometimes it doesn't load it.

shkamaru:
I am not able to get issue with the position of Notifications icon. I placed it accordingly to the gap in the header of the other links and it is same everywhere. Can you tell me what exactly is the margin you are expecting? Because earlier the icon was very close to the profile icon which was making no space for the user between the click on the notifications and the profile icon, that's why the space was increased.

shkamaru:
Made the separate Django app for the homepage. The Travis CI report is showing some errors in the tests files. I fixed the linting as well but not able to understand why is it failing.

adngdb:
@shkamaru there's an error in the tests of `silme`, and given that @mathjazz updated that library recently, I'd be inclined to think that a rebase of your branch against the latest master could fix the problem.

pontoon/homepage/templates/home.html
```diff
@@ -0,0 +1,175 @@
...
+    <section class="snaps section-1 relative" data-section-name="section-1">
+      <div class="introduction neg-margin--large">
+        <div class="title-style">A place to localize Mozilla, fast and easy.</div>
+        <div class="content-style">Join a global community of thousands in making Firefox and various Mozilla products speak to over a hundred of cultures in their native language.</div>
+        <div class="flex vertical-align-portrait">
+          <a class="homepage-button primary-button-cta" href="{{ url('pontoon.teams') }}">
+            <div>Start Localizing Now</div>
+          </a>
+          <p>or</p>
+          <a class="homepage-button secondary-button-cta" href="/tour">
```
mathjazz: Hardcoded relative URLs in templates are forbidden.

Instead of `/tour`, you should use this:
`{{ url('pontoon.translate.locale.agnostic', 'projects', 'demo', 'all-resources') }}`

Instead of `/in-context`, you should use this:
`{{ url('pontoon.translate.locale.agnostic', 'projects', 'pontoon-intro', 'messages.properties') }}`

Instead of `/static/img/image-name.svg`, you should use this:
`{{ static('img/image-name.svg') }}`

shkamaru:
@mathjazz I have merged the master but still the the silme test is failing. I am not sure why.

mathjazz:
@shkamaru Thanks for the quick update! The only thing that remains to be fixed are the links to the tour and the in-context demo.

pontoon/homepage/static/css/homepage.css
```diff
@@ -2,19 +2,6 @@ html {
...
-  }
-}
-
-@media only screen and (min-width: 1500px) {
-  html {
-    font-size: 10.44px;
-  }
-}
-
-
```
shkamaru: As the different screen resolutions was facing issues with the font-size, I used them as they were working as the ratio system for the other screens. Now, it will be the same for all the screen sizes.

pontoon/homepage/static/css/homepage.css
```diff
@@ -434,98 +417,3 @@ body > header {
...
-  }
-
-  .section-5 a {
-    margin: 5.2rem auto;
-  }
-
-  .section-5 .content-style {
-    padding-right: 0;
-  }
-}
```
shkamaru: This was basically used because the elements in a section in the landscape orientation will be in the same row but for portrait orientation they will be placed one under other so that there is more space for the font-size otherwise it looks very small on those screens which was the major problem initially mentioned by adrian.

PR #1003, opened 2018-07-09
Bug 1473134: Add the ability to add special projects for tour.

VishalCR7:
I have been working on implementing a guided tour for Pontoon and for that we want to have some special projects over which the whole tour will be added. The project will be a short-term project with no repository which stays in the database itself.

We'll use this special project for the Pontoon Tour. The idea is that strings will guide the user through the features of Pontoon, like this: https://imgur.com/a/CQYvvad.

The special project needs to have the following properties:
- [X] No stats calculation for any kind of translations.
- [x] Not visible in dashboards but Accessible through some links on home page 
- [x] Private translations ( A user's translation will not be visible to other users).


pontoon/administration/templates/admin_project.html
```diff
@@ -277,6 +277,11 @@ <h3>
           <label for="id_disabled">
               {{ form.disabled }}{{ form.disabled.label }}
           </label>
+      </div>
+       <div class="checkbox clearfix">
+          <label for="id_is_dummy">
+              {{ form.is_dummy }}{{ form.is_dummy.label }}
+          </label>
```
adngdb: I don't think creating dummy projects is a thing that will happen more than once. Thus I think it would be better to not expose this option in the admin panel, and to create the one dummy project we're going to use as a data migration. That makes sure that anyone using Pontoon will have it, which is good because it will be a requirement for the website to work properly.

pontoon/base/models.py
```diff
@@ -338,7 +342,9 @@ def user_locale_role(self, locale):
 @property
 def contributed_translations(self):
     """Filtered contributions provided by user."""
-    return Translation.objects.filter(user=self)
+    return Translation.objects.filter(
+        user=self,
+        entity__resource__project__is_dummy=False)
```
adngdb: Please put closing parentheses on a new line, and add a comma to the end of this line (to make following diffs smaller).

pontoon/base/models.py
```diff
@@ -1134,6 +1146,8 @@ class Project(AggregatedStats):
 
     tags_enabled = models.BooleanField(default=True)
 
+    is_dummy = models.BooleanField(default=False)
```
adngdb: Please add a comment explaining what a `dummy` project is. By the way, maybe we can find a better name than `dummy`? I'm not a big fan of it. Any suggestions?

pontoon/base/models.py
```diff
@@ -1162,17 +1176,20 @@ def save(self, *args, **kwargs):
         for all project locales.
         """
         disabled_changed = False
+        is_dummy_changed = False
         if self.pk is not None:
             try:
                 original = Project.objects.get(pk=self.pk)
                 if self.disabled != original.disabled:
                     disabled_changed = True
+                if self.is_dummy != original.is_dummy:
```
adngdb: I don't think that's a use case we need to support. I don't see why anyone would change a project to `dummy` or the other way around. Do you expect that to happen?

pontoon/base/models.py
```diff
@@ -2153,7 +2175,6 @@ def for_project_locale(
             entities = Entity.objects.filter(pk__in=Entity.objects.filter(Q(*pre_filters)))
         else:
             entities = Entity.objects.all()
-
```
adngdb: Please leave those blank lines, they make the code a lot easier to read. :-)

pontoon/base/models.py
```diff
@@ -2756,5 +2783,17 @@ def calculate_stats(self, save=True):
...
+        # Locale
+        total_strings_diff = (total_strings_diff, 0)[resource.project.is_dummy]
+        approved_strings_diff = (approved_strings_diff, 0)[resource.project.is_dummy]
+        fuzzy_strings_diff = (fuzzy_strings_diff, 0)[resource.project.is_dummy]
+        unreviewed_strings_diff = (unreviewed_strings_diff, 0)[resource.project.is_dummy]
+        locale.adjust_stats(
+            total_strings_diff, approved_strings_diff,
+            fuzzy_strings_diff, unreviewed_strings_diff,
+            for_dummy_project=False
+        )
```
adngdb: What's the point of this repeated code? Why not just use the same pattern as for other calls to `adjust_stats`?

pontoon/base/views.py
```diff
@@ -210,7 +210,7 @@ def _get_entities_list(locale, project, form):
     }, safe=False)
 
 
-def _get_all_entities(locale, project, form, entities):
+def _get_all_entities(request, locale, project, form, entities):
```
adngdb: Why are you adding `request` here? You don't use it in the function.

pontoon/base/views.py
```diff
@@ -262,8 +262,25 @@ def _get_paginated_entities(locale, project, form, entities):
...
+            print(t[0]['user'])
+            if t[0]['user'] not in [request.user.pk, -1]:
+                entity['translation'] = [{
+                    'fuzzy': False,
+                    'string': None,
+                    'approved': False,
+                    'pk': None,
+                    'user': -1,
+                }]
+        entities_updated.append(entity)
```
adngdb: I don't understand this. What is it supposed to do?

pontoon/base/models.py
```diff
@@ -508,12 +514,16 @@ def adjust_stats(
...
         self.total_strings = F('total_strings') + total_strings_diff
-        self.approved_strings = F('approved_strings') + approved_strings_diff
-        self.fuzzy_strings = F('fuzzy_strings') + fuzzy_strings_diff
-        self.unreviewed_strings = F('unreviewed_strings') + unreviewed_strings_diff
+        self.approved_strings = (F('approved_strings') +
+                                 approved_strings_diff, 0)[for_dummy_project]
+        self.fuzzy_strings = (F('fuzzy_strings') +
+                              fuzzy_strings_diff, 0)[for_dummy_project]
+        self.unreviewed_strings = (F('unreviewed_strings') +
+                                   unreviewed_strings_diff, 0)[for_dummy_project]
```
adngdb: This is quite hard to read. I propose instead: 

```python
if for_dummy_project:
    # We do not count stats for dummy projects. 
    self.approved_strings = 0
    self.fuzzy_strings = 0
    self.unreviewed_strings = 0
else:
    self.approved_strings = F('approved_strings') + approved_strings_diff
    self.fuzzy_strings = F('fuzzy_strings') + fuzzy_strings_diff
    self.unreviewed_strings = F('unreviewed_strings') + unreviewed_strings_diff
```

pontoon/base/models.py
adngdb: I am more and more confident that `dummy` is a bad name. We should find a better one. Here's what my brain can come up with: 

*Context: naming a type of projects that can be seen by everyone, exists for every language, but for which translations can only be seen by users who created them. We plan to use such a project to let users try features without having any impact on other users.*

`demo` -> if we're only ever using this for the one project that we'll use to show the features to users, then `demo` can work. 

`sandbox` -> conveys the fact that this project is a sandbox, meaning that what happens on that project has no impact to the rest of the world. It's more generic than `demo` and thus could be used for other potential use cases.

I'm actually quite satisfied with `sandbox`, but I'd like to get the opinion of @mathjazz and our Master of Naming Things, @stasm.

mathjazz: To add a bit more context, we'll use this special project for the Pontoon Tour. The idea is that strings will guide the user through the features of Pontoon, like this: https://imgur.com/a/CQYvvad.

There is however another special project, which we'll use to demo in-context localization. Perhaps we can join them into one project?

These two projects should also not be visible in dashboards and excluded from stats.

I'm a bit skeptical about only exposing translation to the user that submitted them. I'm afraid it will make the code more complex on many different places, as is already seen in this PR. What if we instead don't store translations at all, and only keep them on the client side until next page load? In this case I'd vote for the name `demo`.

adngdb: I would be deeply confused if my translations were not kept between sessions, honestly. That's how smartcat does it, and it means you can come back to the project and use it as a reference.

We could figure out better ways to handle that special case, or refactor the code, to reduce the complexity, but I don't think we should make something weird because we fear the code would to complex.

stasm: >I'm actually quite satisfied with sandbox, but I'd like to get the opinion of @mathjazz and our Master of Naming Things, @stasm.

_Is this a trap? Am I walking straight into it? I guess I am walking straight into it._

I see how _dummy_ is not a good word. I'm not sold on _demo_ either. To me a _demo_ is a time-limited or feature-limited version of the software. A demo is something to be shown during a presentation or in a screencast on the landing page. The way @mathjazz describes it, _tutorial_ might work well.

I also like _sandbox_, although it sounds rather specific about what happen to the translations. I.e. anyone can do anything and the result will not be saved, or will be saved but ignored ni production.

I realize that I'm probably not helping by adding yet another option, but if this is supposed to be a name of a _type_ of projects, then maybe _system projects_, _standard projects_, _builtin projects_, or _skeleton projects_ (a reference to `/etc/skel`) would work?

Thanks for looping me in and I hope I can help somehow. I'm happy to brainstorm other ideas but I'd like to make it clear that I'll be happy with any option you end up choosing.

pontoon/administration/templates/admin_project.html
VishalCR7: Yeah, I just added it for my ease. Will surely remove it eventually.

pontoon/base/models.py
VishalCR7: :+1:

pontoon/base/models.py
VishalCR7: No, not at all. Will change it

pontoon/base/models.py
VishalCR7: Actually, these are  Locales used for all the projects(normal and dummy ones). Other objects are related to specifically dummy  project which can be easily set to zero but doing the same breaks the stat calculation for other projects. If I m still unable to make it clear enough then we can discuss it once.

pontoon/base/models.py
VishalCR7: Agreed! :D

pontoon/base/views.py
VishalCR7: Will add comments in next commit  Sorry for that. This part is supposed to remove translations of entities connected to the dummy project made by other users. If no translation is present for an entity still a translation goes in this format evident from this part https://github.com/VishalCR7/pontoon/blob/master/pontoon/base/models.py#L2109. So if another user's translation comes into query this part removes it and replaces it with a dummy one.

pontoon/base/models.py
VishalCR7: :+1:

pontoon/base/views.py
VishalCR7: What goes for above will be used here, but since above code was too complex I held it for the time we can come up with something simpler and better.

mathjazz:
Which of these steps cover adding the tour to production?

adngdb:
6. Add the demo project to production. :D

Pike:
Sounds good to me.

mathjazz:
SGTM. Let's do this!

PR #1006, opened 2018-07-10
Bug 1464802: File priority & deadline followup

mathjazz:
This is a collection of fixes and improvements that were discovered as part of landing the file-level priority and deadline information.

https://github.com/mozilla/pontoon/commit/5607c65c20f687c2e650778f5481942c0bb103a6 just exposes Resource deadline in the Django admin and
https://github.com/mozilla/pontoon/commit/fe10de397d01b23f29446e8cbd97306d32b36566 removes the foo code to set the project deadline based on resource deadlines.

The interesting one is https://github.com/mozilla/pontoon/commit/7357daa8477b1b9e78394dab13da6c09a2f6732e, which prevents the `TypeError: datetime.date(2018, 2, 20) is not JSON serializable` triggered in here:
https://github.com/mozilla/pontoon/blob/7357daa8477b1b9e78394dab13da6c09a2f6732e/pontoon/projects/templates/projects/widgets/project_selector.html#L24

So I changed `locale.parts_stats()` to return a JSON serializable data by converting a `datetime.datetime` instance to string and then back where it's required.

@adngdb @jotes I wonder if there's a more elegant way of making values of type `datetime.datetime`  
JSON serializable.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -1820,14 +1820,6 @@ class Resource(models.Model):
     def __str__(self):
         return '%s: %s' % (self.project.name, self.path)
 
-    def save(self, *args, **kwargs):
-        super(Resource, self).save(*args, **kwargs)
-
-        if self.deadline and self.project.deadline:
-            if self.deadline < self.project.deadline:
-                self.project.deadline = self.deadline
-                self.project.save()
-
     @classmethod
     def get_path_format(self, path):
         filename, extension = os.path.splitext(path)
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -924,7 +924,7 @@ class Locale(AggregatedStats):
     def parts_stats(self, project):
         """Get locale-project pages/paths with stats."""
         def get_details(parts):
-            values = parts.order_by('title').values(
+            return parts.order_by('title').values(
                 'url',
                 'title',
                 'resource__path',
@@ -934,14 +934,6 @@ class Locale(AggregatedStats):
                 'unreviewed_strings',
                 'approved_strings',
             )
-
-            # Make output JSON serializable
-            for value in values:
-                deadline = value['resource__deadline']
-                if deadline:
-                    value['resource__deadline'] = str(deadline)
-
-            return values
 
         pages = project.subpage_set.all()
         translatedresources = TranslatedResource.objects.filter(
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -924,7 +924,7 @@ class Locale(AggregatedStats):
     def parts_stats(self, project):
         """Get locale-project pages/paths with stats."""
         def get_details(parts):
-            return parts.order_by('title').values(
+            values = parts.order_by('title').values(
                 'url',
                 'title',
                 'resource__path',
@@ -934,6 +934,14 @@ class Locale(AggregatedStats):
                 'unreviewed_strings',
                 'approved_strings',
             )
+
+            # Make output JSON serializable
+            for value in values:
+                deadline = value['resource__deadline']
+                if deadline:
+                    value['resource__deadline'] = str(deadline)
+
+            return values
 
         pages = project.subpage_set.all()
         translatedresources = TranslatedResource.objects.filter(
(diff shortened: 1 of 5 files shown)


pontoon/base/models.py
```diff
@@ -935,6 +935,14 @@ def get_details(parts):
                 'approved_strings',
             )
 
+            # Make output JSON serializable
+            for value in values:
+                deadline = value['resource__deadline']
+                if deadline:
+                    value['resource__deadline'] = str(deadline)
```
adngdb: I used `datetime.isoformat()` in Socorro. Not sure if that's what `str` does under the hood, but might as well be explicit, and `isoformat` is the best format for sharing dates between computer systems in my opinion.

pontoon/localizations/views.py
```diff
@@ -97,6 +99,10 @@ def ajax_resources(request, code, slug):
     for part in parts:
         part['resource__priority'] = resource_priority_map.get(part['title'], None)
 
+        deadline = part['resource__deadline']
+        if deadline:
+            part['resource__deadline'] = datetime.strptime(deadline, '%Y-%m-%d')
```
adngdb: I would do this in the template instead of the view, and keep the full `datetime` object here. And to turn stuff into that format, you can use `my_datetime.date().isoformat()`. See https://docs.python.org/2.7/library/datetime.html#datetime.date.isoformat

pontoon/base/models.py
```diff
@@ -1812,14 +1820,6 @@ def is_asymmetric(self):
     def __str__(self):
         return '%s: %s' % (self.project.name, self.path)
 
-    def save(self, *args, **kwargs):
-        super(Resource, self).save(*args, **kwargs)
-
-        if self.deadline and self.project.deadline:
-            if self.deadline < self.project.deadline:
-                self.project.deadline = self.deadline
-                self.project.save()
```
adngdb: Why do you remove this? I understand that it doesn't work as expected, but does it mean the intention is was wrong or useless? Don't we need to replace this with something that works?

pontoon/base/models.py
jotes: From my past experiences: `datetime.isoformat` + custom Json encoder to keep that in one place.
e.g.
https://gist.github.com/drmalex07/5149635e6ab807c8b21e

pontoon/base/models.py
mathjazz: We never used this function. The idea was to set the project deadline based on the resource deadline, but that's wrong, because the project deadline in this case should be different for each locale (based on what files they have enabled and what's their progress). I can file a bug, but we could also first see if there's a need for such feature, now that we'll actually start using resource deadlines.

mathjazz:
@jotes @adngdb Thank you for your comments. I've updated the code, which is now much simpler and ready for another look.

pontoon/base/models.py
adngdb: OK, sounds good to me!

pontoon/base/templatetags/helpers.py
```diff
@@ -34,6 +34,9 @@ def default(self, obj):
         if isinstance(obj, QuerySet):
             return list(map(str, obj))
 
+        if isinstance(obj, datetime.date):
+            return obj.isoformat()
+
```
adngdb: If we want to make this really really correct, then I believe this should go into a different class. This is not related to Django's LazyObjects at all. I'd write this I think: 

```python
class DatetimeAwareJSONEncoder(json.JSONEncoder):
    def default(self, obj):
        if isinstance(obj, datetime.date):
            return obj.isoformat()

class LazyObjectsJsonEncoder(DatetimeAwareJSONEncoder):
    # etc.
```

pontoon/base/tests/test_helpers.py
```diff
@@ -8,7 +8,12 @@
 from six import text_type
 
 from pontoon.base.templatetags.helpers import (
-    as_simple_translation, format_datetime, format_timedelta, nospam)
+    to_json,
+    format_datetime,
+    format_timedelta,
+    nospam,
+    as_simple_translation,
```
adngdb: Imports should be alphabetically ordered. :)

mathjazz:
Good call! Updated.

PR #1010, opened 2018-07-12
Add synchroneous sync projects

Natim:
At some point our AMQP calls stopped triggering the task.
We used this trick to force exec the task again.

### DIFF
--- a/pontoon/sync/management/commands/sync_projects.py
+++ b/pontoon/sync/management/commands/sync_projects.py
@@ -43,11 +43,19 @@ class Command(BaseCommand):
         )
 
         parser.add_argument(
-            '--force',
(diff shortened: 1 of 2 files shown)


mathjazz:
Thanks for the patch!

We have an [environment variable](http://mozilla-pontoon.readthedocs.io/en/latest/admin/deployment.html#environment-variables) called `CELERY_ALWAYS_EAGER`, which controls whether asynchronous tasks are sent to Celery or executed synchronously.

Do you think we need a management command in addition to that?

Natim:
Yes it totally works to do: `CELERY_ALWAYS_EAGER=True ./manage.py sync_projects`

PR #1012, opened 2018-07-16
Bug 1458345 sync checks

jotes:
Hey,
So this is the next PR for https://bugzilla.mozilla.org/show_bug.cgi?id=1458345

In this one, We finally run checks on translations during the sync process. 

TODO:
* [x] Discuss the feature scope with @mathjazz
* [x] Rebase against master
* [x] Refactor bulk_run_checks tasks from the previous PR
* [x] Add tests

*Discussion/Questions* 
1. Do We want to ignore Translate Toolkit warnings here?
1. I wonder if the sync process should create TranslateMemory entries for the translations with errors?
1. Do we want to approve translations (from vcs) with errors? (I don't know if every VCS repo is protected with `compare-locales`)

### DIFF
--- a/pontoon/base/migrations/0100_bug_1390805_create_missing_tm_entries.py
+++ b/pontoon/base/migrations/0100_bug_1390805_create_missing_tm_entries.py
@@ -9,7 +9,7 @@ def create_missing_translaton_memory_entries(apps, schema_editor):
     Translation = apps.get_model('base', 'Translation')
     TranslationMemoryEntry = apps.get_model('base', 'TranslationMemoryEntry')
 
-    translations_to_create_translaton_memory_entries_for = (
+    translations_to_create_translation_memory_entries_for = (
         Translation.objects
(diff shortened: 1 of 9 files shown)


Pike:
> Do We want to ignore Translate Toolkit warnings here?

Yes, at least for now.

> I wonder if the sync process should create TranslateMemory entries for the translations with errors?

I'd say no.

> Do we want to approve translations (from vcs) with errors? (I don't know if every VCS repo is protected with compare-locales)

Yes, we need to get the content into the db, so that we can constructively fix it. Also, no repo is actually protected with c-l, they're more like travis tests. Nothing happens to your repo if they're failing.

mathjazz:
> Do We want to ignore Translate Toolkit warnings here?

I also filed a bug to [stop storing failed Translate Toolkit checks in the DB](https://bugzilla.mozilla.org/show_bug.cgi?id=1475962) in general.

PR #1013, opened 2018-07-16
Fix bug 1475927: Center-align header of the percentage column

mathjazz:
Column header is now center aligned with the progress line:

screen shot 2018-07-16 at 13 52 36

It looks odd if center-aligned within the entire column width.

### DIFF
--- a/pontoon/base/static/css/table.css
+++ b/pontoon/base/static/css/table.css
@@ -59,6 +59,10 @@ table.table {
   width: 140px;
 }
 
+.table th.progress {
+  padding-right: 39px;
+}
+
 .table th.unreviewed-status {
   position: relative;
   width: 16px;
@@ -72,10 +76,6 @@ table.table {
   position: absolute;
   font-size: 18px;
   margin: -15px 0 0 -5px;
-}
-
-.table .progress {
-  text-align: left;
 }
 
 .table h4 {


mathjazz:
Deployed to stage:
https://mozilla-pontoon-staging.herokuapp.com/sl/firefox/

Also, I didn't find a way to mark the limits of columns while keeping the minimalistic appearance.

flodolo:
This makes the title centered on the progress bar, not on the progress+percentage. 

I can see why you want to do that, but the whole idea of bug 1475927 is to make evident that the header above the percentage is clickable to sort the column, and I'm not sure centering the text on the progress alone helps with that. That's why I was suggesting to add some vertical hints (or maybe some light :hover effect).

schermata 2018-07-16 alle 14 06 21

mathjazz:
I don't think centering the title within the boundaries of the column makes it much more evident that you can sort the column.

I'd even hope it already is evident, because percentage is just a different representation of what progress line is representing.

The only alternative I can think of is right-alignment:
screen shot 2018-07-16 at 14 12 20

flodolo:
> I'd even hope it already is evident, because percentage is just a different representation of what progress line is representing.

Then I think bug 1475927 is a wontfix/invalid (when I filed, I assume percentage had its own invisible header).

PR #1015, opened 2018-07-16
Fix bug 1475962 Exclude Translate Toolkit checks from DB

jotes:
@mathjazz could you look at this?
`DB_EXCLUDE_LIBRARIES`  definetely needs a better name :-)


jotes:
@mathjazz hey, can you look again?

PR #1016, opened 2018-07-17
 Fix bug 1473866 - Do not show rejected translations in strings list and editor.

adngdb:


pontoon/base/static/js/translate.js
```diff
@@ -35,18 +35,31 @@ var Pontoon = (function (my) {
...
+      }
 
-      var li = $('<li class="entity' +
-        (' ' + status) +
-        (translation.pk ? ' has-translations' : '') +
-        (!entity.body ? ' uneditable' : '') +
-        (this.allEntitiesSelected ? ' selected' : '') +
-        (entity.visible ? ' visible': '') +
+      var classNames = ['entity', status];
+      if (translation.pk && !translation.rejected) {
```
adngdb: I'm not sure about this change. It looks like this: 



I think it's ugly, I'd rather always have strings take only half of the space so that it's quite clear which ones lack a translation.

adngdb:
> The only think I'd change is update the editor after a translation is rejected. For example, if there are only rejected translations in the history, the editor should become empty instantly.

I knew there was something I was missing and that I couldn't possibly think about. :D

pontoon/base/static/js/translate.js
```diff
@@ -2204,6 +2204,10 @@ var Pontoon = (function (my) {
             item.find('.unapprove').removeClass('unapprove').addClass('approve').prop('title', 'Approve');
             button.addClass('unreject').removeClass('reject').prop('title', 'Unreject');
 
+            // Reload the editor so that, if the rejected string was the active one,
+            // it disappears from the textarea.
+            Pontoon.switchToEntity(entity);
+
```
mathjazz: You could replace Pontoon with self, but this code is about to die, soon anyways.

PR #1017, opened 2018-07-18
 Fix bug 1473348 - Implement structure for navigation and URL routing. 

adngdb:
This PR defers the navigation to `react-router`, and adds a selector on top of it to better expose URL parameters (locale, project, resource and string). It adds support for URLs on the Django side, and history support on the browser side. 

I've also used the opportunity to do some CSS work so that the EntitiesList module looks like the one from the current app.


mathjazz:
An empty http://localhost:8000/translate/ loads for me with the following error in the JS Console:

```
SyntaxError: JSON.parse: unexpected character at line 2 column 1 of the JSON data[Več o tem]
```

mathjazz:
Oh, thanks for clarifying. I'm fine with fixing it later.

It works nicely, I'm looking at the code now.

PR #1021, opened 2018-07-18
Fix bug 1350546: Enable read-only ProjectLocales

mathjazz:
This patch adds a `readonly` flag to the `ProjectLocale` model, which allows enabling locales for projects in read-only mode. Such locales will act the same as other (read-write) locales, except that users will not be able to make any changes through Pontoon by:
* submitting translations or suggestions
* reviewing ((un)approving or (un)rejecting) existing translations
* uploading files
* performing mass actions

Find more details in commit messages.

@adngdb @jotes r?


pontoon/base/views.py
```diff
@@ -582,6 +614,18 @@ def update_translation(request):
...
     project = e.resource.project
 
+    # Read-only translations cannot saved
+    if (
+        ProjectLocale.objects.filter(
+            locale=locale,
+            project=project,
+            readonly=True,
+        ).exists()
+    ):
```
adngdb: I would put this in a function, to avoid repeating it so much. 

```python
if is_readonly(locale, project):
    return HttpResponseForbidden(
        "Forbidden: This string is in read-only mode"
    )
```

Or, even better, could that be a decorator? That would remove most of the boilerplate and make it a lot easier to use on all those views that edit translations. 

```python
def check_readonly_projetlocale(fn):
    def wrapper(request, *args, **kwargs):
        locale = request.get('locale')
        project = request.get('project')
        
        if (
            ProjectLocale.objects.filter(
                locale=locale,
                project=project,
                readonly=True,
            ).exists()
        ):
            return HttpResponseForbidden(
                "Forbidden: This string is in read-only mode"
            )

        return fn(request, *args, **kwargs)

    return wrapper
```

Wow I haven't written a decorator in a long while, not sure that's correct, haven't tested it... :stuck_out_tongue:

pontoon/batch/views.py
```diff
@@ -121,10 +121,18 @@ def batch_edit_translations(request):
 
     # Batch editing is only available to translators. Check if user has
     # translate permissions for all of the projects in passed entities.
+    # Also make sure projects are note enabled in read-only mode for a locale.
```
adngdb: `s/note/not`?

pontoon/base/views.py
adngdb: Actually, given that you use the query in a bunch of different places, I'd combine the two and have a function that is called by the decorator.

pontoon/base/static/js/translate.js
```diff
@@ -728,6 +742,12 @@ var Pontoon = (function (my) {
 
       self.fluent.toggleButton();
 
+      // Toggle read-only mode
+      $('#editor #single').toggleClass('readonly', entity.readonly);
+      $('#editor textarea').attr('readonly', function() {
+        return self.isReadonlyEditor();
+      });
```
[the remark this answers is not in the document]
mathjazz: Because when all projects are loaded for a locale (e.g. https://pontoon.mozilla.org/sl/all-projects/all-resources/), some strings could be read-only, and some not, so we need to make a check on each entity switch.

pontoon/base/static/js/translate.js
```diff
@@ -3266,12 +3284,21 @@ var Pontoon = (function (my) {
...
         .toggle(this.project.slug !== 'all-projects');
-      $('#profile .upload').toggle(this.state.paths && this.user.canTranslate() && this.part !== 'all-resources');
+
+      var enableFileUpload = (
+        this.state.paths &&
+        this.user.canTranslate() &&
+        this.part !== 'all-resources' &&
+        !this.entities[0].readonly
+      );
+      $('#profile .upload').toggle(enableFileUpload);
```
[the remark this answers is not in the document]
mathjazz: I'm a fast learner. 🐗

mathjazz:
Thanks for your comments, I think they are all addressed.

I didn't use the decorator, because views use different parameters, so we'd need three different decorators. So I went with a function instead. It's not ideal, but it's better than before.

pontoon/base/utils.py
```diff
@@ -679,3 +679,25 @@ def is_same(same_translations, can_translate):
...
+    if (
+        ProjectLocale.objects.filter(
+            project__in=projects,
+            locale=locale,
+            readonly=True,
+        ).exists()
+    ):
+        return True
+
+    return False
```
adngdb: That's the same as `return ProjectLocale.objects.filter(...).exists()`.

pontoon/base/utils.py
```diff
@@ -679,3 +679,25 @@ def is_same(same_translations, can_translate):
...
+    :arg Locale locale: Locale instance.
+    :returns: True if a read-only ProjectLocale instance for given Projects and
+        Locale exists.
+    """
+    # Avoid circular import; someday we should refactor to avoid.
+    from pontoon.base.models import ProjectLocale
+
+    if (
+        ProjectLocale.objects.filter(
+            project__in=projects,
```
adngdb: You can add a check to make sure `projects` is a list, to simplify the usage (and avoid having all those `[project]`).

```python
if not isinstance(projects, (tuple, list)): 
    projects = [projects]
```

adngdb:
> I didn't use the decorator, because views use different parameters, so we'd need three different decorators. 

That's actually quite sad and makes me wish we did refactor that. Having consistent interfaces is always a good thing. Anyway, that's out of scope for sure, but something to keep in mind.

PR #1022, opened 2018-07-19
Fix bug 1477026: data.translation is not always present

mathjazz:
@VishalCR7 r?


VishalCR7:
:+1:  r+  Thanks! :smile:

PR #1024, opened 2018-07-24
Fix bug 1478066 - Use volumes instead of symlinks in docker dev envir…

adngdb:
…onment.


docker/run_webapp.sh
```diff
@@ -2,9 +2,6 @@
 
 # Prepares then runs the webapp.
 
-echo ">>> Prepare revision file"
-git rev-parse HEAD > static/revision.txt
-
```
adngdb: I'm removing this because it didn't work and is completely useless for dev.

adngdb:
@Pike and @stasm I am very interested in your review given that you use non-Linux OSes. @jotes I don't remember if that was the case for you but anyway, if you're using docker, please do give it a try as well. :)

I hope I have tried all use cases but chances are high that I missed something.

docker/run_tests_in_docker.sh
```diff
@@ -25,7 +25,10 @@ if [ "$1" == "--shell" ]; then
 
     docker run \
            --rm \
-           --volume "$(pwd)":/app \
+           --volume "$(pwd)"/pontoon:/app/pontoon \
```
jotes: Do you remember the reason why do we do this in the separate script instead of adding those volumes to docker-compose setup?
In the theory we could switch into something like:
`docker-compose run local/pontoon /bin/bash "${@:2}"`

If that's not possible, We could move those parameters into some variable and use it in both places.

adngdb: That was a copy and paste from Socorro. It is indeed inconsistent, I'll give it a thought and make it better.

adngdb:
Thanks to your comments @jotes and @Pike I've simplified this even further.  :sparkles:

adngdb:
My favorite kind of PR.

PR #1026, opened 2018-07-26
Fix bug 1465121: Update to Fluent Syntax 0.6

mathjazz:
For testing I suggest setting up the [Pontoon FTL](https://github.com/mozilla-l10n/pontoon-ftl) project locally and comparing the behaviour of the `demo.ftl` file (which tries to capture all syntax details) to [stage](https://mozilla-pontoon-staging.herokuapp.com/sl/fluent/demo.ftl/?string=180475) (which still uses the old libraries).

I enabled 2 locales:
* `sl`: to see how existing translations look
* `af`: to test empty editor and submitting translations

Note that `VariantLists` are not supported yet.


pontoon/base/static/js/fluent_interface.js
```diff
@@ -861,7 +868,7 @@ var Pontoon = (function (my) {
           });
 
           if (attributes) {
-            content = (value ? content : entity.key) + attributes;
+            content = (value ? content : entity.key + ' = ') + attributes;
```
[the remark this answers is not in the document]
mathjazz: Beautiful catch!

mathjazz:
https://github.com/mozilla/pontoon/pull/1026/commits/78b9c355c66f2172f117e338fb9ef97a146882a6 brings support for `VariantList` to backend, where we only implement "simplified string presentation". It also adds some missing test cases.

https://github.com/mozilla/pontoon/pull/1026/commits/d206ae3940fe22b4cbdb3f493da05284e8c82c61 does the same thing on frontend, but for all presentations and related helper functions. It does that by mutating the `VariantList` AST to act like "selector-less `SelectExpression`".

@stasm @Pike What do you think about this approach? I'd like to avoid making big changes to `fluent_interface.js` when we're in the middle of the translate view rewrite.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -1,15 +1,39 @@
...
 var fluentParser = new FluentSyntax.FluentParser({ withSpans: false });
+var fluentParserOriginal = new FluentSyntax.FluentParser({ withSpans: false });
 var fluentSerializer = new FluentSyntax.FluentSerializer();
 
+/*
+ * VariantLists behave the same as "selector-less SelectExpressions" in Pontoon.
+ * Overriding the parseEntry() method appliess that logic to the AST, too. That
+ * allows us to use the same code for both, VariantLists and SelectExpressions.
+ */
+fluentParser.parseEntry = function(source, useDefaultParser) {
```
stasm: The fact that you need to override the official implementation is a red flag. Either the design of `VariantList` is wrong, or there has to be a better way to do this. This approach creates a lot of technical debt. I'll be more than happy to help you find a better solution :)

Looking at the source code, in [line 642](https://github.com/mathjazz/pontoon/blob/d206ae3940fe22b4cbdb3f493da05284e8c82c61/pontoon/base/static/js/fluent_interface.js#L642-L645) there's this snippet:

```js
// Value
else if (ast.value) {
    value = renderOriginalElements(ast.value.elements, 'Value');
}
```

You could check if `ast.value` is of type `Pattern` or `VariantList` here and pick between `renderOriginalElements` and, say, `renderOriginalVariants`. Or abstract this check away into a new `renderOriginalValue` method. There might be some duplication regarding the code which renders variants of `SelectExpressions` and variants of `VariantLists`, but not too much, and I'll take it over a custom `parseEntry` any day :)

mathjazz: Perhaps I should have stressed that I'm looking at this implementation as a way to postpone yet another refactor of `fluent_interface.js`, and try to do it after or even as part of the rewrite in React.

stasm: I might be missing something. I'm not suggesting a full refactor. Is my suggestion about line 642 wrong? I really think implementing `VariantList` the proper way at this time shouldn't be very hard and should make the code simpler.

mathjazz: Your suggestion is correct - actually it's what I did on backend (https://github.com/mozilla/pontoon/pull/1026/commits/78b9c355c66f2172f117e338fb9ef97a146882a6), where I replaced the `*_elements` method with the `*_value` method.

I completely agree that this is the right approach in general, but on frontend we have much more code that requires such transformation. Here's why I call it a refactor:

1. There are 4 methods used for different string presentation styles:
* `Pontoon.fluent.renderOriginal()`
* `Pontoon.fluent.renderEditor()`
* `Pontoon.fluent.serializeTranslation()`
* `Pontoon.fluent.getSimplePreview()`

2. There are 8 helper functions that take `element` or `elements` as an argument:
* `isSimpleElement`
* `isPluralElement`
* `renderEditorElement`
* `renderEditorElements`
* `renderOriginalElement`
* `renderOriginalElements`
* `serializeFTLEditorElements`
* `stringifyElements`

3. There are 18 calls to `value.elements`, most of which we can no longer make safely.

4. There are things that I didn't think about yet. :)

I'm not saying that every single one of these will need to be refactored, but if https://github.com/mozilla/pontoon/pull/1026/commits/78b9c355c66f2172f117e338fb9ef97a146882a6 resulted in a 55-line changeset for just 1 of these functions, I'd be surprised if we don't get to a 10-times bigger changeset for the entire `fluent_interface.js`. Changing 500 lines in a file with 1000 lines is hardly not a refactor. :)

I also understand parts of code will bi simpler, but in general we're adding complexity. I don't see any simplification beyond removing the two [ifs](https://github.com/mozilla/pontoon/pull/1026/commits/6904925ccf6b4a4636fd98daa649860936ee174e). On the other hand, the proper way of doing things on the [backend](https://github.com/mozilla/pontoon/commit/78b9c355c66f2172f117e338fb9ef97a146882a6#diff-af5d1b44e5d90a99d8307abe864f4bcb) resulted in 30% more code.

stasm: I understand that there are many parts here which are affected. It might end up being more work than I say. I still firmly believe that it would be better to do this work now than to start with this workaround and effectively do the work twice. I'd prefer to fix the underlying cause rather than the symptom. Fixes to symptoms tend to accumulate until the code becomes unmaintainable and calls for a refactor :)

I don't know how much `fluent_interface.js` is going to be affected by the translate view re-write. You have better visibility into this and I trust your judgement. I understand that this approach allows us to ship Syntax 0.6 faster. I'm worried that by using this work-around we'll make it harder to add code to `fluent_interface` in the longer run. How soon do you expect the translate view re-write to happen? What is the life expectancy of this work-around?

mathjazz: > I'd prefer to fix the underlying cause rather than the symptom. Fixes to symptoms tend to accumulate until the code becomes unmaintainable and calls for a refactor :)

To go back to the true cause, I wonder if we can solve the problem `VariantLists` are solving by making a less radical change in the AST (i.e. that will shorten the upper list of things that need to be refactored). I don't know enough about `VariantLists` and I fully trust your judgement, so a simple Yes/No answer would be sufficient. :)

> How soon do you expect the translate view re-write to happen? What is the life expectancy of this work-around?

The rewrite is ongoing and is expected to be complete in Q3. So I don't think it makes a big difference maintenance-wise if we do the refactor now or then.

What worries me is how much time we spend refactoring `fluent_interface.js`. And given that this particular refactor is caused by a feature that is known to be pretty obscure, I worry that it might not even be the last refactor needed before 1.0.

Can I get some guarantee that no further refactors will be needed before we ship 1.0? :) I will understand if the answer is no, but in this case I believe it's better to wait with the refactor until the very last.

stasm: > To go back to the true cause, I wonder if we can solve the problem VariantLists are solving by making a less radical change in the AST (i.e. that will shorten the upper list of things that need to be refactored). 

No, I don't think that's possible. They are a different kind of values than Patterns.

> What worries me is how much time we spend refactoring fluent_interface.js. And given that this particular refactor is caused by a feature that is known to be pretty obscure, I worry that it might not even be the last refactor needed before 1.0.

I see. I don't want to make product decisions for Pontoon. My advice is based on what's best for Fluent long-term. I understand that Pontoon might want to choose a more winded path to get to the full support :)

> Can I get some guarantee that no further refactors will be needed before we ship 1.0? :)

No, sorry, I can't guarantee this. There aren't any ground shaking changes planned before 1.0, but this commitment concerns the syntax. AST might be subject to change, still, and depending on the scope of changes, they might require more work in `fluent_interface`.

Which is also why I'm afraid of taking the work-around route right now. If more changes are needed down the pike, the next refactor is going to be bigger and harder. Or, we'll choose to add more work-arounds, which is how technical debt is created.

That said, I see your point about this being an obscure feature of Fluent, rarely used in Pontoon right. For this reason, it makes sense to go with the work-around. I think what I'm really afraid is that we won't be able to fully (and properly) support Fluent 1.0 before the translate view re-write is complete. Are these two related? Is Fluent 1.0 an explicit goal for the rewrite? Or is it independent (but impacted by it, as we see).

stasm: Sounds good, @mathjazz, thanks.

mathjazz: OK. I believe the PR is ready for a review then.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -367,14 +392,19 @@ var Pontoon = (function (my) {
...
             endMarker = '</mark>';
           }
           string += startMarker + '{$' + element.expression.id.name + '}' + endMarker;
         }
-        else if (element.expression.type === 'MessageReference') {
+        else if (
+          [
+            'MessageReference',
+            'TermReference',
+          ].indexOf(element.expression.type) >= 0
```
stasm: Nit: using `||` makes the intent clearer IMO:

    else if (element.expression.type === "MessageReference"
        || element.expression.type === "TermReference") {

pontoon/base/static/js/fluent_interface.js
```diff
@@ -228,8 +253,8 @@ var Pontoon = (function (my) {
       // Render SelectExpression
       if (element.expression && element.expression.type === 'SelectExpression') {
```
stasm: Why does this check for `element.expression` first?

pontoon/base/static/js/fluent_interface.js
```diff
@@ -228,8 +230,8 @@ var Pontoon = (function (my) {
...
         var expression = '';
-        if (element.expression.expression) {
-          expression = fluentSerializer.serializeExpression(element.expression.expression);
+        if (element.expression.selector) {
```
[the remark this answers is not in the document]
stasm: We'll need to keep this check now with the `VariantList` work-around.

pontoon/base/templatetags/helpers.py
```diff
@@ -263,36 +263,43 @@ def dict_html_attrs(dict_obj):
...
+
+
+def _serialize_value(value):
+    """Serialize AST values into a simple string."""
     response = ''
 
-    for element in elements:
-        if isinstance(element, ast.TextElement):
-            response += element.value
+    if hasattr(value, 'variants'):
```
stasm: Please use `isinstance` here and later. I don't think duck typing is the right approach for working with the AST. The logic is specific to a data type, which happens to have fields named this way.

```python
if isinstance(value, ast.VariantList):
```

pontoon/base/templatetags/helpers.py
```diff
@@ -263,36 +263,43 @@ def dict_html_attrs(dict_obj):
...
+        default_variant = _get_default_variant(value.variants)
+        response += _serialize_value(default_variant.value)
 
-        elif isinstance(element, ast.Placeable):
-            if isinstance(element.expression, ast.ExternalArgument):
-                response += '{ $' + element.expression.id.name + ' }'
+    elif hasattr(value, 'elements'):
```
stasm:
```py
if isinstance(value, ast.Pattern):
```

pontoon/base/static/js/fluent_interface.js
```diff
@@ -1,15 +1,39 @@
 /* global FluentSyntax */
 var fluentParser = new FluentSyntax.FluentParser({ withSpans: false });
+var fluentParserOriginal = new FluentSyntax.FluentParser({ withSpans: false });
```
stasm: Rather than instantiating the parser twice and overwriting `parseEntry` in one of the instances, why not write a custom `parseEntry` function and use it throughout the file when needed? It should save some memory and also make this work-around a bit nicer. The custom logic belongs in Pontoon, rather than in `FluentParser` after all.

pontoon/base/static/js/fluent_interface.js
mathjazz: We don't always have `element.expression.type`, e.g.:
https://github.com/mozilla-l10n/pontoon-ftl/blob/master/en-US/demo.ftl#L14

pontoon/base/static/js/fluent_interface.js
mathjazz: I was thinking about that initially, but then I changed my mind, because it's a temporary workaround and eventually we'd need to replace `parseEntry()` with `fluentParser.parseEntry()` again. Do you think I should still do it?

mathjazz:
Thanks for the review! I think I've addressed all the comments but one, for which I've left a question.

pontoon/base/static/js/fluent_interface.js
stasm: We'll need to change all the callsites which use `useDefaultParser = false` anyways :) This might be a personal preference, but if possible, I'd go for wrappers rather than monkey patches.

stasm: Or, in other words:

image

pontoon/base/static/js/fluent_interface.js
stasm: Do you mean the fact that 1) this is an old 0.4 syntax? Or the fact that 2) these are `TextElements`?

1. I think we shouldn't have any 0.4 strings left anywhere across current Mozilla projects. In fact, `fluent-syntax` 0.8.0 removed the Syntax 0.4 support. This string will now parse as `Junk`. It might be worth changing the `demo.ftl` file.

2. `isSimpleElement` should catch `TextElements`. In fact, shouldn't there be an early return in `renderOriginalElements` just before `// Render SelectExpression`? Or at least I'd suggest adding a `if (element.type === "Placeable")` test there. All `Placeables` are guaranteed to have the `expression` field.

pontoon/base/static/js/fluent_interface.js
mathjazz: OK, I'll do that in a bit.

pontoon/base/static/js/fluent_interface.js
mathjazz: Oh, right... Let me first fix the demo.ftl.

mathjazz:
Updated.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -367,28 +390,29 @@ var Pontoon = (function (my) {
...
-            'StringExpression',
-            'NumberExpression',
-            'VariantExpression',
-            'AttributeExpression',
-          ].indexOf(element.expression.type) >= 0
+          element.expression.type === 'AttributeExpression' ||
+          element.expression.type === 'CallExpression' ||
+          element.expression.type === 'VariantExpression' ||
+          element.expression.type === 'NumberLiteral' ||
+          element.expression.type === 'StringLiteral'
```
stasm: Thanks for moving away from `indexOf`. This already looks easier on the eyes. I now realize that an even better way to handle those `ifs` would be a `switch` statement, with multiple cases falling through each other. It could be also used in `isSimpleElement`. I know you didn't want to spend too much time working on `fluent_interface`, so feel free to leave the code as-is for landing.

stasm:
Looks great, thanks!

PR #1027, opened 2018-07-26
Fix bug 1476945: [FTL] Query TM with simple strings

mathjazz:
Details:
https://bugzilla.mozilla.org/show_bug.cgi?id=1476945#c7

### DIFF
--- a/pontoon/base/static/js/main.js
+++ b/pontoon/base/static/js/main.js
@@ -341,15 +341,7 @@ var Pontoon = (function (my) {
           tab = $('#' + loader).addClass('loading'), // .loading class used on the /machinery page
           requests = 0,
           count = 0,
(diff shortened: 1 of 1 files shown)


pontoon/base/static/js/main.js
```diff
@@ -609,7 +605,7 @@ var Pontoon = (function (my) {
       }
 
       // Machine translation (Caighdean)
-      if (self.locale.code === 'ga-IE') {
+      if (!customSearch && self.locale.code === 'ga-IE') {
```
mathjazz: On custom search (also used on `/machinery`) we don't have an entity.id, so we cannot use this service.

mathjazz:
@adngdb says: r+.

I say: Thank you!

PR #1031, opened 2018-07-27
Fix bug 1471354: Make review icons visible to unauthenticated users and contributors

VishalCR7:
For unauthenticated users: 

image


pontoon/base/static/css/translate.css
```diff
@@ -1995,7 +1995,9 @@ body > header aside p {
   background: no-repeat transparent;
   border: none;
   color: #AAAAAA;
-  display: none;
+  display: inline-block;
+  pointer-events: none;
```
mathjazz: Cool stuff!

PR #1032, opened 2018-07-28
Fix bug 1407180: unchanged filter only on active translations

raivisdejus:
Fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1407180

Unchanged filter will check only active translations to avoid false positives that have historically had been translated with unchanged string but have been changed (fixed) since

Added check for "active" attribute of the translation object

Also adjusted tests to match the current state of the application

Tested with tests and on pontoon-intro project by approving some non-identical string and checking that the entry disappears from list of unchanged strings.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -2021,8 +2021,9 @@ class EntityQuerySet(models.QuerySet):
         return Q(
             pk__in=self.get_filtered_entities(
                 locale,
-                Q(Q(string=F('entity__string')) | Q(string=F('entity__string_plural'))),
-                lambda x: x.string == x.entity.string,
+                Q(Q(approved=True) &
+                  Q(Q(string=F('entity__string')) | Q(string=F('entity__string_plural')))),
+                lambda x: x.approved and x.string == x.entity.string,
                 match_all=False,
             )
         )
@@ -2417,7 +2418,7 @@ class Translation(DirtyFieldsMixin, models.Model):
         User, related_name='unrejected_translations', null=True, blank=True)
     unrejected_date = models.DateTimeField(null=True, blank=True)
 
-    # Field contains a concatenated state of the  entity for faster search lookups.
+    # Field contains a concatenated state of the entity for faster search lookups.
     # Due to the nature of sql queries, it's faster to perform `icontains` filter on the same table
     # than OR condition for The Entity and The Translation class joined together.
     entity_document = models.TextField(blank=True)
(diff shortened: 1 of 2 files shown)


mathjazz:
Thanks for the patch, @raivisdejus!

An "Active translation" is the translation displayed in the sidebar, next to the original string. It's either the approved translation, or the latest unreviewed suggestion (if there are no approved translations):
https://github.com/mozilla/pontoon/blob/master/pontoon/base/models.py#L2126

While your patch works fine for strings with approved translations, it doesn't cover the rest of the strings.

adngdb:
Hey @raivisdejus. Active translations can be approved, fuzzy or the latest suggestion (in that order). It is not the translation that gets sent to the product (only approved and fuzzy translations make it to VCS). 

If this is not the correct solution, you're right that addressing in the right manner will be complex. I think we can either code this using the database schema we currently have, and make a pretty messy piece of code, or we can, as I have been thinking about for some time now, [add a new marker to the database for active translations](https://bugzilla.mozilla.org/show_bug.cgi?id=1481175). That would help us address problems like this one with very straightforward code. However, that would require another substantial amount of code refactoring, as we would need to ensure that the active marker is always consistent and correct. 

Considering that what you're trying to achieve right now is doable by combining filters (approved + unchanged, though that does not do exactly what you want, but hopefully that is close enough?), I think we should reject your pull request and instead make sure we take the time, as soon as possible, to make this right by working on that active marker proposal. Sadly this won't be done in the near future as I already have a pretty big refactoring on my hands, but if you (or anyone) is willing to contribute to that feature, I would be very happy to help!

raivisdejus:
**Issue 1 - Sending translations to VCS**
If we send only strings that some admin has also seen it is ok. So sending approved or fuzzy is ok.

**Refactoring**
This is a deeper think that I will not go into as I have not played around the code for long enough, but generally speaking having "approved" status on DB level seems like a useful thing. It has costs, like time for refactoring, so not a strait forward decision.

**Unchanged filter that checks mostly active translations**
For me this patch is quite important as it hinders our work right now. 

I see that my patch can break some workflows if they rely on filtering out unchanged suggestions.

But my patch is really important to me :) How about an idea to have another filter "unchanged_active" that would filter out only unchanged active. We can add it to the list of filter or have as secret "link only" filter to not clutter the UI.

mathjazz:
Hi @raivisdejus, we just raised the priority of https://bugzilla.mozilla.org/show_bug.cgi?id=1481175 and we plan to fix it in the next two weeks. That will allow us to properly fix https://bugzilla.mozilla.org/show_bug.cgi?id=1407180.

raivisdejus:
Great news, can't wait!

Good health and steady hands and minds to you for at least the next few weeks.

PR #1033, opened 2018-07-29
Fix bug 1479208 - cl checks won't return errors for DTD entities with…

jotes:
Hey @mathjazz & @Pike!
Could you take a look?

### DIFF
--- a/pontoon/checks/libraries/compare_locales.py
+++ b/pontoon/checks/libraries/compare_locales.py
@@ -45,7 +45,18 @@ class CompareDTDEntity(DTDEntityMixin):
 
     @property
     def all(self):
-        return u'<!ENTITY {} \"{}\">'.format(self.key, self.raw_val)
+        # Wrap entities with apostrophes when they contain quotes
+        # Related bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1479208
+        if '"' in self.raw_val:
+            wrap = "'"
+        else:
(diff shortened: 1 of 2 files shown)


pontoon/checks/libraries/compare_locales.py
```diff
@@ -45,7 +45,18 @@ def __init__(self, key, raw_val, pre_comment):
 
     @property
     def all(self):
-        return u'<!ENTITY {} \"{}\">'.format(self.key, self.raw_val)
+        # Wrap entities with apostrophes when they contain quotes
+        # Related bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1479208
+        if '"' in self.raw_val:
+            wrap = "'"
+        else:
+            wrap = '"'
```
mathjazz: Thanks for addressing this bug so quickly!

This will wrap `Don't press \"Enter\".` like this: `'Don't press \"Enter\".'`. I checked and we don't have such strings in the DB, but we could, so I suggest we stay on the safe side.

jotes: That's what pondered me too and that's why I decided to start by doing this PR instead of going too deep.

jotes:
@Pike Thanks for https://bugzilla.mozilla.org/show_bug.cgi?id=1402560, it helps a lot and solves some of my doubts ;D

PR #1037, opened 2018-08-01
No bug: Allow empty translations in inc format

mathjazz:
Turns out the `inc` format allows empty translations:
https://hg.mozilla.org/users/m_owca.info/seamonkey-central/file/65cd9bcb5f50/suite/profile/bookmarks.inc#l80

So we shouldn't raise an error when an empty string is submitted as a translation in Pontoon.

### DIFF
--- a/pontoon/checks/libraries/pontoon.py
+++ b/pontoon/checks/libraries/pontoon.py
@@ -65,7 +65,7 @@ def run_checks(entity, string):
                 )
 
     # Prevent empty translation submissions if not supported
-    if resource_ext not in {'properties', 'ini', 'dtd'} and string == '':
+    if resource_ext not in {'properties', 'ini', 'dtd', 'inc'} and string == '':
         checks['pErrors'].append(
             'Empty translations are not allowed'
         )


pontoon/checks/libraries/pontoon.py
```diff
@@ -65,7 +65,7 @@ def run_checks(entity, string):
                 )
 
     # Prevent empty translation submissions if not supported
-    if resource_ext not in {'properties', 'ini', 'dtd'} and string == '':
+    if resource_ext not in {'properties', 'ini', 'dtd', 'inc'} and string == '':
```
jotes: [nit] Maybe it's the time to move to create a constant value e.g. EMPTY_TRANSLATION_FORMATS and move them to `pontoon.checks.__init__`?

mathjazz:
Agreed. I'd prefer if we keep that list in the model, so that we don't forget to update it when new formats are added.

The `DB_FORMATS` is a different beast, because we'll update it when we update `compare-locales`.

pontoon/checks/libraries/pontoon.py
```diff
@@ -65,7 +66,7 @@ def run_checks(entity, string):
                 )
 
     # Prevent empty translation submissions if not supported
-    if resource_ext not in {'properties', 'ini', 'dtd', 'inc'} and string == '':
+    if resource_ext not in Resource.EMPTY_TRANSLATION_FORMATS and string == '':
```
jotes: [nit] Do you think that property here could be a good idea?
e.g. `if entity.resource.allow_empty_translation and string == ''`?

mathjazz: Yes. :) I'll add that.

PR #1038, opened 2018-08-01
Bug 1473134:Keep track of user's progress during the tour.

VishalCR7:
I have added a new IntegerField to UserProfile model to keep track of user's progress during the tour so that the user can start just after the point where he left it. 

A 0 implies either finished or not started yet state and other numbers correspond to the step on which user stopped the tour.


pontoon/base/views.py
```diff
@@ -741,6 +741,21 @@ def update_translation(request):
         })
 
 
+@utils.require_AJAX
+def update_tour_status(request):
+    """Perform quality checks and return a list of any failed ones."""
+    try:
+        tour_status = request.POST['tour_status']
+    except MultiValueDictKeyError:
+        raise Http404
```
adngdb: This shouldn't be a 404 but a 400 (Bad Request Error), with a message explaining that the param is missing. You could get that for "free" by using a django Form.

pontoon/base/views.py
```diff
@@ -741,6 +741,21 @@ def update_translation(request):
         })
 
 
+@utils.require_AJAX
+def update_tour_status(request):
+    """Perform quality checks and return a list of any failed ones."""
+    try:
+        tour_status = request.POST['tour_status']
```
adngdb: You should validate that this is an integer.

pontoon/base/views.py
```diff
@@ -741,6 +741,21 @@ def update_translation(request):
...
+    except MultiValueDictKeyError:
+        raise Http404
+    user = request.user
+    user.profile.tour_status = tour_status
+    user.profile.save()
+    return JsonResponse({
+        'tourStatus': tour_status,
```
adngdb: I think you can simply return `True` here, as the client doesn't need the data.

pontoon/base/static/js/tour.js
```diff
@@ -45,7 +63,16 @@ $(function () {
...
           "localization (l10n) tool. At Mozilla, we currently use Pontoon " +
-          "to localize various Mozilla project."
+          "to localize various Mozilla project.",
+        listeners: {
+          beforeStep: function() {
+            if(tourStatus != 0)
+              Sideshow.gotoStep(tourStatus+1);
+          },
+          afterStep: function() {
+            update_tour_status(1);
```
adngdb: I don't like that you have to set the step numbers manually. If we ever add a step near the beginning, we'll have to change all the following numbers, and that's going to be very annoying and hard to maintain. Can you find a different solution so that the numbers aren't hard-coded?

pontoon/base/static/js/tour.js
```diff
@@ -204,7 +269,12 @@ $(function () {
           "some of which we didn't mention in this introductory tutorial. " +
           "<br> Feel free to explore this demo project to know about these  " +
           "or move forward to translate some live projects.",
-        format: "markdown"
+        format: "markdown",
+        listeners: {
+          afterStep: function() {
+            update_tour_status(0);
```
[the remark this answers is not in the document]
VishalCR7: I have set it up to start from 1st step for a user who completes it. If we want to show the tour only once then a `-1` will be good. Which one do you prefer?

pontoon/base/static/js/translate.js
```diff
@@ -4408,6 +4408,7 @@ Pontoon.user = {
   forceSuggestions: $('#server').data('force-suggestions') === 'True' ? true : false,
   manager: $('#server').data('manager'),
   localesOrder: $('#server').data('locales-order') || {},
+  tourStatus: $('#server').data('tour-status') || 0,
```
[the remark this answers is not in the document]
VishalCR7: Cool!

pontoon/base/views.py
VishalCR7: I think a django form is the way to go then ;-)

pontoon/base/static/js/tour.js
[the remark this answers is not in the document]
adngdb: A user who has completed the tour should not see it again when going to the demo page. We could instead add a button or something if we do want to allow users to take the tour again, but I'm not sure that's needed.

pontoon/base/static/js/tour.js
```diff
@@ -66,11 +66,14 @@ var update_tour_status = function(step) {
...
         listeners: {
           beforeStep: function() {
+            // Take the user directly to next step of where he left.
             if(tourStatus != 0)
               Sideshow.gotoStep(tourStatus+1);
           },
           afterStep: function() {
-            update_tour_status(1);
+            if (tourStatus === 0) {
+              updateTourStatus(tourStatus+1);
```
adngdb: I like this idea! However, if the request to save the status to the server fails, then the status number won't get updated, and on subsequent calls, the number will be off. 

For example: 

1 -> OK, status = 1
2 -> FAIL, status = 1
3 -> OK, status = 2

Next time the user comes to the demo page, they will be sent back to page 2 instead of 3. I think the simple way to solve this is to simply pass `tourStatus++` as a parameter, and not use the value returned by the server at all.

pontoon/base/static/js/tour.js
```diff
@@ -275,14 +275,15 @@ $(function () {
...
+            updateTourStatus(-1);
           }
         },
       }
     ]
   });
 
-  if (Pontoon.state.project === "demo") {
+  // Run the tour only on project with slug 'demo' and if not completed by user
+  if (Pontoon.state.project === "demo"&& tourStatus != -1) {
```
adngdb: Need a space before `&&`.

pontoon/base/static/js/tour.js
```diff
@@ -45,7 +63,19 @@ $(function () {
         text:
           "Pontoon is a web-based, What-You-See-Is-What-You-Get (WYSIWYG), " +
           "localization (l10n) tool. At Mozilla, we currently use Pontoon " +
-          "to localize various Mozilla project."
+          "to localize various Mozilla project.",
+        listeners: {
+          beforeStep: function() {
+            // Take the user directly to next step of where he left.
+            if(tourStatus != 0)
```
adngdb: Space after `if`, and use `!==`.

PR #1040, opened 2018-08-03
Fix bug 1480184: Upgrade to silme 0.9.4

mathjazz:
We upgraded silme to fix its .inc parser to properly support empty translations.


mathjazz:
r=me

PR #1042, opened 2018-08-06
Fix bug 1463173 - Run checks during batch actions

jotes:
# TODO
* [x] Discuss the feature scope
* [x] Implement changes
* [x] Add tests

# Discussion points
* How `find&replace` action should work? I thought about the following approaches:
  * Allow to create suggestions with errors but don't mark them as approved
  * Don't create suggestions at all if the change could create generate errors. This approach could require additional UI changes (to display errors etc.)
* Batch Approve should approve only translations without any errors (?)
* I think that batch reject operation doesn't require checks (feel free to correct me)

@mathjazz @pike hey, could you look at these discussion points?

Thanks in advance!

### DIFF
--- a/pontoon/batch/actions.py
+++ b/pontoon/batch/actions.py
@@ -26,6 +26,8 @@ def batch_action_template(form, user, translations, locale):
               * latest_translation_pk: the id of the latest affected
                 translation
               * changed_translation_pks: a list of ids of affected translations
+              * invalid_translation_pks: a list of ids of translations that have
+                errors or changes on them could introduce new errors e.g. after find & replace
 
     """
     return {
@@ -34,6 +36,7 @@ def batch_action_template(form, user, translations, locale):
         'changed_entities': [],
         'latest_translation_pk': None,
         'changed_translation_pks': [],
+        'invalid_translation_pks': [],
     }
 
 
@@ -43,7 +46,17 @@ def approve_translations(form, user, translations, locale):
     For documentation, refer to the `batch_action_template` function.
 
     """
-    translations = translations.filter(approved=False)
+    invalid_translation_pks = list(
+        translations.filter(
+            approved=False,
+            errors__isnull=False,
+        ).values_list('pk', flat=True)
+    )
+
+    translations = translations.filter(
+        approved=False,
+        errors__isnull=True,
+    )
     changed_translation_pks = list(translations.values_list('pk', flat=True))
 
     latest_translation_pk = None
@@ -69,6 +82,7 @@ def approve_translations(form, user, translations, locale):
         'changed_entities': changed_entities,
         'latest_translation_pk': latest_translation_pk,
         'changed_translation_pks': changed_translation_pks,
+        'invalid_translation_pks': invalid_translation_pks,
     }
 
 
@@ -108,6 +122,7 @@ def reject_translations(form, user, translations, locale):
         'changed_entities': changed_entities,
         'latest_translation_pk': None,
         'changed_translation_pks': [],
+        'invalid_translation_pks': [],
     }
 
 
@@ -125,13 +140,14 @@ def replace_translations(form, user, translations, locale):
     latest_translation_pk = None
 
     try:
-        old_translations, changed_translations = utils.find_and_replace(
+        old_translations, changed_translations, invalid_translation_pks = utils.find_and_replace(
             translations,
             find,
             replace,
             user
         )
         changed_translation_pks = [c.pk for c in changed_translations]
+
         if changed_translation_pks:
             latest_translation_pk = max(changed_translation_pks)
     except Translation.NotAllowed:
@@ -161,6 +177,7 @@ def replace_translations(form, user, translations, locale):
         'changed_entities': changed_entities,
         'latest_translation_pk': latest_translation_pk,
         'changed_translation_pks': changed_translation_pks,
+        'invalid_translation_pks': invalid_translation_pks,
     }
 
 
--- a/pontoon/batch/actions.py
+++ b/pontoon/batch/actions.py
@@ -139,21 +139,16 @@ def replace_translations(form, user, translations, locale):
     replace = form.cleaned_data['replace']
     latest_translation_pk = None
 
-    try:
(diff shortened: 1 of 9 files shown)


mathjazz:
> How find&replace action should work? I thought about the following approaches:
> ** Allow to create suggestions with errors but don't mark them as approved
> ** Don't create suggestions at all if the change could create generate errors. This approach could require additional UI changes (to display errors etc.)

I prefer the latter, because that behaviour is consistent with submitting single translations. We simply block submissions with errors.

> Batch Approve should approve only translations without any errors (?)

Correct.

> I think that batch reject operation doesn't require checks (feel free to correct me)

Correct.

--

I was thinking about a more advanced report after batch actions, with e.g. links to strings skipped due to errors. But at this point it's hard to tell how useful would such feature be, so I suggest we avoid blocking this KR with that and potentially add it in the future.

We could however amend the "5 strings replaced/approved" message to something like "5 strings replaced/approved, 2 skipped due to errors".

jotes:
@mathjazz I think it's ready for your review :)

pontoon/batch/utils.py
```diff
@@ -80,7 +82,7 @@ def find_and_replace(translations, find, replace, user):
 
     # No matches found
     if translations.count() == 0:
-        return translations, []
+        return translations, [], []
 
     # Empty translations produced by replace are not allowed for all formats
```
mathjazz: I think we remove this block now that we run all the checks (including this one)?

Also, the `return` in line 112 should is a bug (luckily prevented on frontend). It should be `continue`.

pontoon/batch/utils.py
```diff
@@ -114,11 +119,32 @@ def find_and_replace(translations, find, replace, user):
...
         translation.fuzzy = False
-        translations_to_create.append(translation)
+
+        errors = run_checks(
+            translation.entity,
+            translation.locale.code,
+            translation.entity.string,
+            translation.string,
+            use_tt_checks=False,
+        )
```
mathjazz: I didn't check, but I wonder in how many DB queries will this result. Looking at the surrounding code, we might need to prefetch `entity__resource` and `locale`.

pontoon/batch/tests/test_views.py
```diff
@@ -2,7 +2,75 @@
...
+        project=translation.entity.resource.project,
+        locale=translation.locale,
+    )
+
+    yield translation
+
+
+@pytest.yield_fixture
+def translation_dtd_invalid_unapproved():
+    # Provide invalid characters in translation to cause checks
```
mathjazz: Nit: "...to fail".

pontoon/batch/tests/test_views.py
```diff
@@ -11,56 +79,151 @@ def test_batch_edit_translations_no_user(client):
...
+        find='Translation',
+        replace='%$#%>',
+    )
+
+    assert response.json() == {
+        'count': 0,
+        'invalid_translation_count': 1,
+    }
+
+    translation_dtd_unapproved.refresh_from_db()
```
mathjazz: The string of `translation_dtd_unapproved` would never change. I'd use the `new_translation` trick from above here, too.

jotes:
@mathjazz can you check it again?

pontoon/batch/actions.py
```diff
@@ -125,13 +140,14 @@ def replace_translations(form, user, translations, locale):
...
             translations,
             find,
             replace,
             user
         )
         changed_translation_pks = [c.pk for c in changed_translations]
+
         if changed_translation_pks:
             latest_translation_pk = max(changed_translation_pks)
     except Translation.NotAllowed:
```
mathjazz: You can remove this try - except block now that we no longer raise `Translation.NotAllowed`. In fact, you can even remove the `TranslationNotAllowed` exception from `pontoon.base.models`.

pontoon/checks/utils.py
```diff
@@ -8,17 +8,13 @@
...
+
     translations = (
         Translation.objects
-        .filter(
-            entity__resource__format__in=DB_FORMATS,
-            **qs_filters
-
-        )
         .prefetch_related(
             'entity',
```
mathjazz: This line is not needed, because it's already covered by the next one.

pontoon/checks/utils.py
```diff
@@ -8,17 +8,13 @@
 from pontoon.checks.libraries import run_checks
 
 
-def get_translations(**qs_filters):
+def prefetch_translations():
     """
-    Prefetch translations with fields required for bulk_run_checks
+    Prefetch translations with fields required for run_checks
     """
+
```
mathjazz: Nit: no space is needed here.

pontoon/checks/utils.py
```diff
@@ -8,17 +8,13 @@
 from pontoon.checks.libraries import run_checks
 
 
-def get_translations(**qs_filters):
+def prefetch_translations():
```
mathjazz: Wouldn't it be more elegant to join both of these functions into a `TranslationQuerySet` method, e.g. `for_checks()`, which would prefetch relevant data and filter only translations of `DB_FORMATS`?

```python
def for_checks(self):
    """
    Prefetch data with fields needed for running checks.
    """
    return self.prefetch_related(
        'entity__resource__entities',
        'locale',
    ).filter(
        entity__resource__format__in=DB_FORMATS,
    )
```

Instead of:
```python
prefetched_translations = prefetch_translations().filter(pk__in=translations)
```

You would then call (note that we only care about `DB_FORMATS` checks on batch actions, too):
```python
translations = translations.for_checks()
```

jotes: That was one of my initial ideas, but I decided to skip this approach because that could later end-up with circular dependencies between `base` and `checks` modules (e.g. because of DB_FORMATS). However, It's not a strong opinion -> I can adjust this patch if you prefer to create `for_checks()` method in the `base` module.

mathjazz: As agreed on IRC, we'll go ahead and see if the circular dependency indeed causes trouble.

PR #1051, opened 2018-08-09
Fix bug 1481766 - Show entity metadata.

adngdb:
This adds a new component to show the metadata of a selected entity. It also refactors the component and the Editor component into a new, high-level (redux-connected) component. This makes the logic a bit easier to test, reduces the number of redux-connected components, and makes more sense semantically.


adngdb:
> Clicking on a link to resource doesn't work (nothing happens).

It works for me, it correctly goes to the path (and you can see it because it removes the `string` parameter and shows "Select an entity" instead of the metadata component).

I'm addressing the rest. Thanks for the great review! :-)

adngdb:
@mathjazz how does this look now? :-)

mathjazz:
Nice work!

1. The `context` entry is still missing.

2. The `source [#: ...]` in PO files is styled slightly differently then on prod:
https://pontoon.mozilla.org/sl/amo/LC_MESSAGES/djangojs.po/?string=66616

3. Thumbnail of the screenshot should also be clickable, because users are more likely to click on it than on the link in the comment.

adngdb:
@mathjazz updated! This time I believe we should be good.

mathjazz:
Confirming that this seems to be working fine now! Still need to have a look at the code in more detail.

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        return <Property title='Context'>{ entity.key }</Property>;
+    }
+
+    renderSourceArray(source: Array<Array<string>>): React.Node {
+        return <ul>{ source.map((value, key) => {
+            return <li key={ key }><span className="title">#:</span>{ value.join(':') }</li>;
+        }) }</ul>;
+    }
```
mathjazz: Nit: this will return `<ul></ul>` if we don't have any source files instead of `null`.

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -0,0 +1,140 @@
...
+
+        return this.renderSourceObject(entity.source);
+    }
+
+    render(): React.Node {
+        const { entity, locale } = this.props;
+
+        return <div className="metadata">
+            <Screenshots source={ entity.comment } locale={ locale } />
+            <p className="original">{ entity.original }</p>
```
mathjazz: The original string is not part of string metadata. It should be a sibling of `div.metadata`, instead of a child.

Which means it should be a separate component. Although very simple right now, it will be much more complex in the case of FTL strings.

adngdb: I understand, but I think I'd rather refactor when the time comes to integrate support for Fluent. It's not harmful to not have a separate component right now, and when that time comes to have one, we might have a different idea for implementation.

frontend/src/modules/entitydetails/components/Metadata.js
adngdb: In the case of an empty `source`, that's handled in `renderSources`. But it might be possible that `source = [[]]` in which case we would indeed return an empty list. I've added a check for that.

adngdb:
@mathjazz My apologies, last night I thought about this and realized the `Lightbox` module could be isolated a lot better. I've made some extensive changes in 263e3f30c0974ee0ae24710d7c7e11643ae4cd1c, but it doesn't change anything functionally. Do you mind giving a look at that commit?

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -81,6 +81,10 @@ export default class Metadata extends React.Component<Props> {
     }
 
     renderSourceArray(source: Array<Array<string>>): React.Node {
+        if (source.length === 1 && !source[0]) {
+            return null;
+        }
+
```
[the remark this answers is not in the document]
mathjazz: So you don't see any `<ul></ul>`'s if you check a random Firefox string?

"[Arrays are particularly weird](https://www.nfriedly.com/techblog/2009/07/advanced-javascript-operators-and-truthy-falsy/)".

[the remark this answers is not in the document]
adngdb: God damn it Javascript!

PR #1052, opened 2018-08-09
Fix bug 1482067 - Do not fail when loading caighdean while offline.

adngdb:


pontoon/machinery/views.py
```diff
@@ -142,6 +146,12 @@ def machine_translation_caighdean(request):
     except Translation.DoesNotExist:
         return JsonResponse({})
 
+    if Translator is None:
+        return http.HttpResponseServerError(
+            json.dumps(dict(error=500, message='Caighdean unavaible offline')),
```
jotes: `unavailable`

pontoon/machinery/views.py
```diff
@@ -148,7 +150,10 @@ def machine_translation_caighdean(request):
 
     if Translator is None:
         return http.HttpResponseServerError(
-            json.dumps(dict(error=500, message='Caighdean unavaible offline')),
+            json.dumps({
+                'error': 500,
+                'message': 'Caighdean unavailable offline'
```
mathjazz: Nit: comma. :P

adngdb:
I realized just before merging that this is exactly the kind of code that requires a comment. Pretty sure 31-year-old me won't understand this without the comment I just added.

mathjazz:
Looks like we forgot to merge this.

PR #1056, opened 2018-08-14
No bug - Fix possible XSS bug in the tags display

jotes:
Hey,
As We discussed, this should fix the possible XSS attack from anyone who's able to create new tags.

I decided to escape only  `slug` and `name` fields because they have `CharField` type.
//cc @mathjazz

### DIFF
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -3216,14 +3216,19 @@ var Pontoon = (function (my) {
       tags.forEach(function(tag) {
         var priority = '';
 
+        // In order to avoid possible XSS attacks these variables can't be rendered as HTML,
+        // because they contain input which comes from users.
+        var name = Pontoon.doNotRender(tag.name);
+        var slug = Pontoon.doNotRender(tag.slug);
+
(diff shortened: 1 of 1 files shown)


pontoon/base/static/js/translate.js
```diff
@@ -3215,15 +3215,16 @@ var Pontoon = (function (my) {
 
       tags.forEach(function(tag) {
         var priority = '';
+        var slug = Pontoon.doNotRender(tag.slug);
```
mathjazz: I'd add `var name = Pontoon.doNotRender(tag.name);` here and a comment on top of these two lines explaining why is that needed. Other than that, this is good to go. :)

pontoon/base/static/js/translate.js
```diff
@@ -3216,14 +3216,19 @@ var Pontoon = (function (my) {
       tags.forEach(function(tag) {
         var priority = '';
 
+        // In order to avoid possible XSS attacks these variables can't be rendered as HTML,
```
jotes: @mathjazz could you check if this comment is okay?

PR #1057, opened 2018-08-14
Bug 1482862: Hide tutorial and pontoon-intro from dashboards

VishalCR7:
Adding datamigration for pontoon-intro remains. This patch hides the "system projects" from teams and projects dashboard and also from permissions tab.


pontoon/base/models.py
```diff
@@ -1370,9 +1382,13 @@ def __str__(self):
...
         """
-        Available project locales belong to available projects.
+        Available project locales belong to translable projects.
         """
-        return self.filter(project__disabled=False, project__resources__isnull=False).distinct()
+        return self.filter(
+            project__disabled=False,
+            project__system_project=False,
+            project__resources__isnull=False,
+        ).distinct()
```
mathjazz: This change implies that we no longer need the old implementation of the `ProjectLocale.available()` method. Are you sure about that?

If yes, you should rename the method to the same name as you use on the `Project` model (e.g. `visible`).

If not, you should add a new method, instead of overriding the excisting one,

pontoon/base/models.py
```diff
@@ -1023,6 +1023,16 @@ def get_details(parts):
...
 class ProjectQuerySet(models.QuerySet):
+    def translable(self):
+        """
+        Translable projects are not disabled and have at least one
+        resource defined and are not system projects.
+        """
+        return self.filter(
+            disabled=False, resources__isnull=False,
+            system_project=False,
+        ).distinct()
```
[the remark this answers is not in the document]
VishalCR7: `visible` sounds good :)

pontoon/base/models.py
```diff
@@ -1106,6 +1115,11 @@ class Project(AggregatedStats):
         Prevent project from syncing with VCS.
     """)
 
+    system_project = models.BooleanField(default=False, help_text="""
+        System projects are built into Pontoon. They are accessible from the
+        translate view, but hidden from dashboards.
+    """)
+
```
mathjazz: Now that you added the `help_text`, you'll also need to update the migration.

If you've already applied it, first unapply it with:
`./manage.py migrate base 0126`

Then delete the `0127_system_project_flag.py` file and create a new migration with:
`./manage.py makemigrations base`

pontoon/base/models.py
```diff
@@ -1372,7 +1386,10 @@ def available(self):
         """
         Available project locales belong to available projects.
         """
-        return self.filter(project__disabled=False, project__resources__isnull=False).distinct()
+        return self.filter(
+            project__disabled=False,
+            project__resources__isnull=False,
+        ).distinct()
```
mathjazz: So I looked into this method and figured we only use it twice. In https://github.com/mozilla/pontoon/pull/1057/files#diff-c18ce63ddb8aca155c0553a0c1b174e3R832 and also in the `pontoon//teams/views.py` file. And, both of these method should exclude system projects.

So I suggest you change the `available()` method to `visible()` and call it in these two places.

mathjazz:
Nice work! :)

PR #1060, opened 2018-08-18
Bug 1483577: Disable tab keydown events during tour.

VishalCR7:
As @mathjazz  suggested that disabling the "tab" key might be the fastest way to solve the bug. I have tried to do the same.

### DIFF
--- a/pontoon/tour/static/js/tour.js
+++ b/pontoon/tour/static/js/tour.js
@@ -56,7 +56,17 @@ $(function () {
       function() {
         return true;
       }
-    ]
+    ],
+    listeners: {
+        beforeWizardStarts: function(){
+            $("body").addClass("tour");
+            $("body *").attr("tabindex", "-1");
+        },
+        afterWizardEnds: function(){
+           $("body").removeClass("tour");
+           $("body *").removeAttr("tabindex");
+        }
+    }
   }).storyLine({
     showStepPosition: true,
     steps: [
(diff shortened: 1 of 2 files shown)


pontoon/tour/static/js/tour.js
```diff
@@ -56,7 +56,17 @@ $(function () {
       function() {
         return true;
       }
-    ]
+    ],
+    listeners: {
+        beforeWizardStarts: function(){
+            $("body").addClass("tour");
+            $("body *").attr("tabindex", "-1");
```
adngdb: This means that absolutely no elements on the page can be navigated using tab. That is a big breaker for accessibility. We want to make sure that while the tour is active, no hidden element (behind the gray background) can be focused, but we want the controls of the tour to still be accessible, as well as the non-hidden elements.

mathjazz:
@adngdb What's the status here?

VishalCR7:
@mathjazz I dont know what exactly needs to be done in this but this solution is not whats expected. How about I close this PR for now and open a new one later?

adngdb:
@VishalCR7 we talked about this today, and we have a better definition of the work that needs to be done now. Quoting myself from [the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1483577): 

> This bug should focus on making sure that, while the tour is active, hidden elements of the page cannot receive the focus. Only the tour elements themselves, and the elements being shown, should be able to receive focus through keyboard navigation.

The current implementation removes all keyboard navigation with Tab entirely, which is a big blocker for accessibility. Feel free to close this PR and open a new one if you want, or to keep working on this  branch, that is entirely up to you. And let us know if you need help, or can't work on this anymore! :-)

VishalCR7:
Thanks @adngdb  I will open a new PR after a while ( busy with midterm exams right now :cry: )

PR #1064, opened 2018-08-21
Fix bug 1376454: Un-submodule pontoon.intro

mathjazz:
This patch removes the [pontoon-intro](https://github.com/mozilla/pontoon-intro/) submodule and replaces it with the new app `pontoon.in-context`. It also updates the `/intro/` page contents to serve as in-context demo page, instead of introducing Pontoon. Docs are also updated.

As a follow-up, I'll remove everything but the `locales` folder from [pontoon-intro](https://github.com/mozilla/pontoon-intro/) and update source strings. Alternativelly, we could simpy replace the Pontoon Intro project with e.g. In-context Tutorial with a DB migration.


mathjazz:
@adngdb Updated and ready for another look.

PR #1065, opened 2018-08-22
Fix Bug 1484223: Add signal for automatically adding system projects for new locales.

VishalCR7:


pontoon/base/signals.py
```diff
@@ -170,3 +170,29 @@ def assign_project_locale_group_permissions(sender, **kwargs):
 def create_user_profile(sender, instance, created, **kwargs):
     if created:
         UserProfile.objects.create(user=instance)
+
+
+@receiver(post_save, sender=Locale)
+def add_locale_to_system_projects(sender, instance, created, **kwargs):
```
mathjazz: There's another `@receiver(post_save, sender=Locale)` above, I suggest we move this signal right after it.

pontoon/base/signals.py
```diff
@@ -170,3 +170,29 @@ def assign_project_locale_group_permissions(sender, **kwargs):
...
+                resource=project.resources.all()[0],
+                locale=instance,
+                total_strings=project.resources.all()[0].total_strings,
+            )
+
+            project.total_strings += project.resources.all()[0].total_strings
+            project.save(update_fields=['total_strings'])
+
+
+
```
mathjazz: One new line at the end if enough, you can remove the other three.

pontoon/base/signals.py
```diff
@@ -170,3 +170,29 @@ def assign_project_locale_group_permissions(sender, **kwargs):
...
+                total_strings=project.resources.all()[0].total_strings,
+            )
+            TranslatedResource.objects.create(
+                resource=project.resources.all()[0],
+                locale=instance,
+                total_strings=project.resources.all()[0].total_strings,
+            )
+
+            project.total_strings += project.resources.all()[0].total_strings
+            project.save(update_fields=['total_strings'])
```
mathjazz: You can avoid the `total_strings` dance by running `calculate_stats()` on the created `TranslatedResource` instance.

pontoon/base/signals.py
```diff
@@ -134,13 +134,24 @@ def create_project_locale_permissions_groups(sender, **kwargs):
...
+    projects = Project.objects.filter(system_project=True)
+    for project in projects:
+        ProjectLocale.objects.create(project=project, locale=instance)
+        for resource in project.resources.all():
+            translated_resource = TranslatedResource.objects.create(
+                resource=resource,
+                locale=instance,
+            )
+            translated_resource.calculate_stats()
+
```
mathjazz: This should still be a separate `add_locale_to_system_projects` function.

mathjazz:
Just confirming that this works fine. Please fix the review comment and we're good to go. :)

PR #1070, opened 2018-08-28
Fix bug 1486117 - Correctly sync when a new locale is added to the tu…

adngdb:
…torial project.

### DIFF
--- a/pontoon/administration/views.py
+++ b/pontoon/administration/views.py
@@ -312,6 +312,41 @@ def _get_project_strings_csv(project, entities, output):
     return output
 
 
+def _get_resource_for_database_project(project):
+    """Return the Resource object of an in database project.
+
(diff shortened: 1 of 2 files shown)


pontoon/administration/views.py
```diff
@@ -365,8 +367,10 @@ def _create_or_update_translated_resources(
         )
 
     if resource is None:
+        is_tutorial = project.slug == 'tutorial'
+        resource_path = 'playground' if is_tutorial else 'database'
         resource, _ = Resource.objects.get_or_create(
-            path='database',
+            path=resource_path,
```
mathjazz: A DB project can only have 1 resource. It would be much more elegant to rely on that instead of hardcoding resource paths (twice).

adngdb:
@mathjazz updated!

PR #1072, opened 2018-08-30
[WIP] Bug 1481175: Active translations

mathjazz:
- [x] Add Translations.active field
- [x] Add unique constraint to the Translation.active field
- [x] Add data migration to set Translation.active to true where needed
- [x] Use `Translation.active` instead of an algorithm to get active translations.
- [x] Ensure `Translation.active` is updated correctly when data changes.
- [x] Remove `print` statements from migrations after deploying to stage.

Note: This patch also removes the `Translation.entity_document` field and related trigger and function, which were introduced [earlier](https://github.com/mozilla/pontoon/blob/master/pontoon/base/migrations/0091_add_fulltext_search.py) and aren't used anymore.

We'll probably need to put Pontoon into maintenance mode while we deploy this. Migration timings on a few weeks old DB dump look like this for me locally:
* Migrations to add `active` field and set indices: ~3 minutes.
* Set `approved` and `fuzzy` translations to active: ~14 minutes.
* Identify `unreviewed` translations to be made active: ~5 minutes.
* Set identified `unreviewed` translations to active: ~1 minute.


pontoon/base/migrations/0131_translation_active_populate.py
```diff
@@ -0,0 +1,68 @@
...
+
+    for candidate in candidates:
+        siblings = Translation.objects.filter(
+            entity=candidate.entity,
+            locale=candidate.locale,
+            plural_form=candidate.plural_form,
+        )
+        if not siblings.filter(active=True).exists():
+            most_recent = siblings.latest('date')
+            unreviewed_pks.add(most_recent.pk)
```
adngdb: I think this block can be optimized. Right now it does a lot of computation several times (if an entity has several suggestions, then this logic is applied to all suggestions). However, given how complex this model is, finding the optimal solution is going to be quite complex, and my brain is definitely not able to do that on a Thursday evening. :P

pontoon/base/migrations/0129_translation_active.py
```diff
@@ -0,0 +1,42 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
mathjazz: I think Django added this name automatically.

pontoon/base/migrations/0131_translation_active_populate.py
```diff
@@ -0,0 +1,68 @@
...
+
+from django.db import migrations
+
+
+def set_active_translations(apps, schema_editor):
+    Translation = apps.get_model('base', 'Translation')
+
+    # All approved translations are active.
+    Translation.objects.filter(approved=True).update(active=True)
+    print(datetime.datetime.time(datetime.datetime.now()))
```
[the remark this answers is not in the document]
mathjazz: Oh! Thanks :)

pontoon/base/migrations/0131_translation_active_populate.py
```diff
@@ -0,0 +1,68 @@
...
+    # siblings.
+    # Note that currenty this logic applies to all suggestions, even rejected.
+    # However, this is a bug, and we're fixing it here.
+    unreviewed_pks = set()
+    candidates = Translation.objects.filter(
+        approved=False,
+        fuzzy=False,
+        rejected=False,
+    )
+    print(datetime.datetime.time(datetime.datetime.now()))
```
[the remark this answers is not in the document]
mathjazz: Yup, just nticed that earlier :)

pontoon/base/migrations/0131_translation_active_populate.py
```diff
@@ -0,0 +1,68 @@
...
+from __future__ import unicode_literals
+
+import datetime
+
+from django.db import migrations
+
+
+def set_active_translations(apps, schema_editor):
+    Translation = apps.get_model('base', 'Translation')
+    print(datetime.datetime.now().time())
```
mathjazz: I'll remove these `print` statements before mering. Only using that for measuring performance during development.

pontoon/base/migrations/0131_translation_active_populate.py
mathjazz: True. I thought about that for 5 minutes and then gave up and tried to measure it first. Check out the [timings](https://github.com/mozilla/pontoon/pull/1072#issue-212123351).

This part takes 8 minutes, which means that even if we get it down to a minute, there's still 75% of the migration time left. We'll likely need to take Pontoon down briefly anyhow.

pontoon/base/migrations/0129_translation_active.py
```diff
@@ -0,0 +1,42 @@
...
+        migrations.RemoveField(
+            model_name='translation',
+            name='entity_document',
+        ),
+        migrations.AddField(
+            model_name='translation',
+            name='active',
+            field=models.BooleanField(default=False),
+        ),
+        migrations.RunPython(
```
jotes: [nit] you can use `migrations.RunSQL` and `migrations.RunSQL.noop` respectively

pontoon/base/migrations/0131_translation_active_populate.py
```diff
@@ -0,0 +1,68 @@
...
+    # (locale, entity, plural_form) is active, unless it has approved or fuzzy
+    # siblings.
+    # Note that currenty this logic applies to all suggestions, even rejected.
+    # However, this is a bug, and we're fixing it here.
+    unreviewed_pks = set()
+    candidates = Translation.objects.filter(
+        approved=False,
+        fuzzy=False,
+        rejected=False,
+    )
```
jotes: [nit] you probably could use `values_list` to return only `entity_id`, `locale` and `plural_form` (that would reduce the time required by Django in order to initialize object instances).

pontoon/base/migrations/0129_translation_active.py
mathjazz: That was the initial implementation, but then it resulted in `ProgrammingError` (which is now catched) as I was applying and unapplying migrations during development.

pontoon/base/migrations/0131_translation_active_populate.py
mathjazz: Wow, that splits the time in half! I'll ammend the commit.

mathjazz:
@adngdb @jotes I'll add some tests, notably for `Entity.reset_active_translation` and `EntityQuerySet.reset_active_translations`. Other than that, this is now ready for a full review.

jotes:
@mathjazz I'll review this tomorrow!

pontoon/base/utils.py
```diff
@@ -680,8 +680,7 @@ def is_same(same_translations, can_translate):
         if st.approved and not st.fuzzy:
             return True
     else:
-        if not st.fuzzy:
-            return True
+        return True
 
     return False
```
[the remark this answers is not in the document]
mathjazz: It will, if `can_translate` is `True` and `st.approved` is `False`. In reality that means that a privileged user adds a suggestion that already exists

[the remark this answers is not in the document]
adngdb: Right, sorry!

pontoon/base/models.py
```diff
@@ -2660,6 +2660,31 @@ def unapprove(self, user, stats=True):
...
+        Reject translation.
+        """
+        # Check if translation was approved or fuzzy.
+        # We must do this before unapproving/unfuzzying it.
+        if self.approved or self.fuzzy:
+            if stats:
+                TranslatedResource.objects.get(
+                    resource=self.entity.resource,
+                    locale=self.locale
+                ).calculate_stats()
```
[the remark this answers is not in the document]
mathjazz: Good point. :) This is useless. Especially because we `calculate_stats` (properly) in the custom `Translation.save()` method. Updating.

pontoon/base/migrations/0131_translation_active_populate.py
```diff
@@ -0,0 +1,71 @@
...
+                plural_form=plural_form,
+            )
+            .exclude(rejected=True)
+            .order_by('-active', '-date')
+        )
+        if siblings and not siblings[0].active:
+            unreviewed_pks.add(siblings[0].pk)
+    print(datetime.datetime.now().time())
+
+    Translation.objects.filter(pk__in=unreviewed_pks).update(active=True)
```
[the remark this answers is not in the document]
mathjazz: Calculation of `unreviewed_pks` relies on the `active` flag already being set for `approved` and `fuzzy` strings. Given the timings above, I doubt changing the calculation would bring important performance wins.

pontoon/base/tests/managers/test_entity.py
```diff
@@ -831,3 +831,125 @@ def test_lookup_collation(resource_a, locale_a):
             string__icontains=u'string'
         )) == [translations[n] for n in [0, 2, 3, 4]]
     )
+
+
+@pytest.mark.django_db
+def test_mgr_entity_reset_active_translations(resource_a, locale_a):
```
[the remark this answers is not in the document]
mathjazz: This test cases tests an `Entity` manager method. First, it creates entities and their translations, and then calls the method - **once**. The only thing that follows at the bottom is assertions.

That's consistent with other test cases in the file, which mostly test filters. I think it's also more robust than slicing the test into multiple test cases, which would use smaller `QuerySets`.

mathjazz:
In https://github.com/mozilla/pontoon/pull/1072/commits/9d2050daade90fc54e5332d75734be19c4359a2c I think it makes sense to split test cases.

adngdb:
Testing locally: 

- I can't see active translations in entities list anymore, nor status. 
- However, if I send a new translation, that one appears correctly.
- Same if I make any changes (e.g. unapprove / approve, add a suggestion... ).

I'm not sure if that's because I missed something, or something is missing in the code. Migrations did run as far as I can tell.

pontoon/sync/changeset.py
```diff
@@ -108,7 +109,40 @@ def execute(self):
...
+                        WHERE locale_id = %(locale_id)s
+                        AND res.project_id = %(project_id)s)
+                    AND approved_date !=
+                    (SELECT max(approved_date)
+                        FROM base_translation
+                        WHERE entity_id = b.entity_id
+                        AND locale_id = b.locale_id
+                        AND (plural_form = b.plural_form OR plural_form IS NULL));
+            """, {
+                'locale_id': self.locale.id,
```
exsilium: I'm getting a NoneType error when syncing with Git repo from `self.locale.id`. DB migrations have been done. Am I missing something when trying to upgrade my dev instance?

mathjazz: Should be fixed in https://github.com/mozilla/pontoon/commit/20bffedb82441e4c7fe5d2813691627221575054. Let me know if the problem persist.

exsilium: Thanks for the quick reply @mathjazz , I guess I should have mentioned from the start that the issue appeared for me in 20bffedb82441e4c7fe5d2813691627221575054 . I then dropped to 6db15d17c9707ac1340119acc0e67dba7839fac0 to try if it works in the original merge - it didn't. 😭

mathjazz: Hm, that's surprising. Could you paste the stack trace?

exsilium: I tried reproducing this issue for the stack trace going from safe state e2fbf0cbeb09cb12e8ea373768dffb9a2101b3e0 to 20bffedb82441e4c7fe5d2813691627221575054 and no longer couldn't reproduce it. Issue solved. Thanks!

PR #1073, opened 2018-09-04
Fix bug 1486516 - Continuous loading of entities. 

adngdb:
This makes sure that the screen always takes 100% of the screen, and that scrolling is handled in panels, notably in the EntitiesList component. It also adds an infinite scroller so that next entities are loaded when a user goes near the bottom of the list (by default, within 250px from the bottom).

### DIFF
--- a/frontend/src/App.css
+++ b/frontend/src/App.css
@@ -1,5 +1,6 @@
 #app {
     display: flex;
+    height: 100%;
 }
 
 #app > section {
--- a/frontend/src/modules/entitieslist/actions.js
+++ b/frontend/src/modules/entitieslist/actions.js
@@ -3,8 +3,9 @@
 import type { Translation } from './reducer';
 
 
-export const REQUEST: 'entitieslist/REQUEST' = 'entitieslist/REQUEST';
 export const RECEIVE: 'entitieslist/RECEIVE' = 'entitieslist/RECEIVE';
+export const REQUEST: 'entitieslist/REQUEST' = 'entitieslist/REQUEST';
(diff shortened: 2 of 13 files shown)


frontend/src/App.css
```diff
@@ -1,5 +1,6 @@
 #app {
     display: flex;
+    height: 100%;
 }
```
mathjazz: Stupid question: can we merge `#root` and `#app` into a single element?

frontend/src/modules/entitieslist/actions.js
```diff
@@ -84,13 +92,25 @@ export function get(
 
         const response = await fetch(url, requestParams);
         const content = await response.json();
-        dispatch(receive(content.entities));
+        dispatch(receive(content.entities, content.has_next));
```
mathjazz: It's a bit confusing to have `has_next` and `hasMore` - should we unify that?

adngdb: `<InfiniteScroll>` accepts that argument as `hasMore`, which is why I called in that in my code, and the API exposes `has_next`. I don't want to mess with the API as part of Translate.Next (not until we're done with rewriting everything), so we're bound to having this difference. The question is, which term do we think is better, and thus want to use in our code? I don't really have an opinion there, if you have one, please let me know.

adngdb:
@mathjazz I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1489054 to add loaders. I want to keep PRs small, and this one is already too big in my opinion.

PR #1074, opened 2018-09-05
Fix bug 1488804: Add unique constraint to approved & fuzzy fields

mathjazz:
Translation.approved and Translation.fuzzy can be True for only one (locale + entity + plural_form) combination. We should encode that into the DB.


mathjazz:
This is not going to work:
https://bugzilla.mozilla.org/show_bug.cgi?id=1488804#c2

PR #1076, opened 2018-09-07
Fix bug 1486608: Aggregate error and warning counts in the model

mathjazz:
This patch simply adds `strings_with_errors` and `strings_with_warnings` to the `AggregatedStats` model. It also updates the related `adjust_stats`, `aggregate_stats`, `calculate_stats` methods accordingly and exposes stats in Django admin. A data migration updates stats calculation, and takes between 2-3 minutes to complete.

Also included is a fix for [bug 1486607](https://bugzilla.mozilla.org/show_bug.cgi?id=1486607) (Expose errors and warnings in the API), because it depends on [bug 1486608](https://bugzilla.mozilla.org/show_bug.cgi?id=1486608) and is super simple.

@jotes r?

**A review note**: 70% of this patch is migrations, one of witch includes a copy and paste of the `TranslatedResource.calculate_stats()`, since model methods are not available on the `apps` models.

**A deployment note**: We should hold off deploying this until [bug 1486606](https://bugzilla.mozilla.org/show_bug.cgi?id=1486606) is fixed, otherwise numbers on dashboards will be off (because approved strings with errors and warnings are remove from the approved count).


pontoon/base/migrations/0133_calculate_strings_with_errors_warnings.py
```diff
@@ -0,0 +1,254 @@
...
+                list(Warning.objects.values_list('translation', flat=True))
+            )
+        )
+        .prefetch_related('entity__resource')
+    )
+
+    # Collect TranslatedResources of translations with errors and warnings
+    # and run calculate_stats() on them.
+    query = Q()
+    for t in translations.values('entity__resource', 'locale', 'pk'):
```
jotes: [nit] there's no reference to 'pk' later from what I see (also, with .distinct() you probably could reduce number of Q() objects)

pontoon/base/models.py
```diff
@@ -2913,30 +2933,94 @@ def calculate_stats(self, save=True):
...
+            locale=locale,
+        ).prefetch_related(
+            'errors',
+            'warnings',
         )
-        approved = translations.filter(approved=True).count()
-        fuzzy = translations.filter(fuzzy=True).count()
-        missing = resource.total_strings - approved - fuzzy
+
+        approved = translations.filter(
```
jotes: [nit] Maybe it's a good moment to add new method, e.g. `Translation.objects.without_errors()`?

pontoon/base/models.py
```diff
@@ -2913,30 +2933,94 @@ def calculate_stats(self, save=True):
...
         # Plural
         nplurals = locale.nplurals or 1
         for e in translated_entities.exclude(string_plural=''):
-            translations = Translation.objects.filter(entity=e, locale=locale)
-            plural_approved_count = translations.filter(approved=True).count()
-            plural_fuzzy_count = translations.filter(fuzzy=True).count()
+            translations = Translation.objects.filter(
+                entity=e,
+                locale=locale,
+            ).prefetch_related(
```
jotes: [nit/questions] I wonder if it's required do prefetch fields if you only call `count()` later.

pontoon/base/migrations/0133_calculate_strings_with_errors_warnings.py
```diff
@@ -0,0 +1,254 @@
...
+        approved=False,
+        fuzzy=False,
+        rejected=False,
+    ).count()
+
+    missing = resource.total_strings - approved - fuzzy - errors - warnings
+
+    # Plural
+    nplurals = len(map(int, locale.cldr_plurals.split(','))) or 1
+    for e in translated_entities.exclude(string_plural=''):
```
jotes: [nit]  `.values_list('pk') and entity_id=` could probably reduce ORM overhead.

pontoon/sync/models.py
```diff
@@ -93,7 +93,12 @@ def fix_stats(self):
                 resource__project__sync_disabled=False,
             )
             .annotate(
-                total=Sum(F('approved_strings') + F('fuzzy_strings'))
+                total=Sum(
```
jotes: [nit] Comment in line 88 needs an update.

pontoon/base/models.py
mathjazz: Possibly, I'll see if it's needed while I fix bug 1486606.

pontoon/base/models.py
mathjazz: Good call!

mathjazz:
@jotes This is now ready for another look.

PR #1077, opened 2018-09-08
Fix bug 1407180: unchanged filter only on active translations

raivisdejus:
Unchanged filter will check only active translations to avoid false
positives that have historically had been translated with unchanged
string but have been changed (fixed) since


pontoon/base/models.py
```diff
@@ -2106,8 +2106,9 @@ def unchanged(self, locale):
         return Q(
             pk__in=self.get_filtered_entities(
                 locale,
-                Q(Q(string=F('entity__string')) | Q(string=F('entity__string_plural'))),
-                lambda x: x.string == x.entity.string,
+                Q(Q(active=True) &
+                  Q(Q(string=F('entity__string')) | Q(string=F('entity__string_plural')))),
```
mathjazz: Nit: we try to use a diff-friendly codestyle (it's not required, just a recommendation), which putch each concept into its own line.  So I suggest you reorganize this code like this:

```python
Q(
    Q(active=True) &
    Q(
        Q(string=F('entity__string')) |
        Q(string=F('entity__string_plural'))
    )
),
```

pontoon/base/models.py
```diff
@@ -2106,8 +2106,9 @@ def unchanged(self, locale):
...
+                lambda x: x.active and x.string == x.entity.string,
```
mathjazz: We should also compare the string to `string_plural` (unrelated to this bug), so this line should read:

```python
lambda x: x.active and (
    x.string == x.entity.string or
    x.string == x.entity.string_plural
),
```

raivisdejus:
Thanks for comments, useful. Updated code.

PR #1078, opened 2018-09-11
 Fix bug 1481767 - Implement the History tab in Translate.Next.

adngdb:
This also refactors the API usage into a core module, exposing functions separated into namespaces per type of objects (current translation and entity). Finally, it separates our big `style.css` file into smaller files, and imports some of those from Translate.Next. 

Some features are missing from this patch. That is on purpose: adding them would add a lot of complexity to an already quite complex diff. I filed bugs for those features: 

* [Change status actions](https://bugzilla.mozilla.org/show_bug.cgi?id=1490351)
* [Support for plural forms](https://bugzilla.mozilla.org/show_bug.cgi?id=1486517) (right now, for PO files, it will show all forms instead of just one)
* [Load translation in Editor on click](https://bugzilla.mozilla.org/show_bug.cgi?id=1490346)
* [Show diff](https://bugzilla.mozilla.org/show_bug.cgi?id=1490343) (this one is mentored, looks like a fairly good first bug to me)


frontend/src/core/api/index.js
```diff
@@ -0,0 +1,10 @@
+/* @flow */
+
+import { default as TranslationAPI } from './translation';
+import { default as EntityAPI } from './entity';
```
adngdb: This could simply be `import TranslationAPI from './translation';`

frontend/src/modules/entitydetails/actions.js
```diff
@@ -8,39 +10,15 @@ export function suggest(
...
-        const requestParams = {
-            method: 'POST',
-            credentials: 'same-origin',
-            headers: headers,
-            body: payload,
-        };
-
-        const response = await fetch(url, requestParams);
-        const content = await response.json();
+): Function {
```
adngdb: This `: Function` might not be needed.

frontend/src/modules/history/actions.js
```diff
@@ -0,0 +1,50 @@
...
+export const RECEIVE: 'history/RECEIVE' = 'history/RECEIVE';
+export const REQUEST: 'history/REQUEST' = 'history/REQUEST';
+
+
+export type ReceiveAction = {|
+    +type: typeof RECEIVE,
+    +entity: number,
+    +translations: Array<Object>,
+|};
+export function receive(entity: number, translations: Array<Object>) {
```
adngdb: Missing return Flow type.

frontend/src/modules/history/actions.js
```diff
@@ -0,0 +1,50 @@
...
+        translations,
+    };
+}
+
+
+export type RequestAction = {|
+    +type: typeof REQUEST,
+    +entity: number,
+|};
+export function request(entity: number) {
```
adngdb: Missing return Flow type.

frontend/src/modules/history/components/Translation.js
```diff
@@ -0,0 +1,99 @@
...
+            return 'rejected';
+        }
+        return 'unreviewed';
+    }
+
+    getApprovalTitle() {
+        const { translation } = this.props;
+
+        if (translation.approved && translation.approved_user) {
+            return 'Approved by ' + translation.approved_user;
```
adngdb: It would probably be better to add a level of indirection between what the API returns and the attributes we use in our code. Here for example we should use `translation.approvedUser` and in the reducer get that from `translation.approved_user`.

adngdb: I also need to see if it's better to use `'a ' + b` or `` `a ${b}` ``.

frontend/src/modules/history/components/Translation.js
```diff
@@ -0,0 +1,99 @@
...
+        }
+        else {
```
julen: You don't need this `else`, nor the other one below.

frontend/src/modules/history/reducer.js
```diff
@@ -0,0 +1,72 @@
...
+
+export default function reducer(
+    state: HistoryState = initialState,
+    action: Action
+): HistoryState {
+    switch (action.type) {
+        case REQUEST:
+            return {
+                ...state,
+                ...{
```
julen: Create another object and then destructuring it is redundant — the extra arguments can be passed directly:
```js
return {
  ...state,
  entity: action.entity,
  fetching: true,
  translations: [],
}
```

frontend/src/modules/history/components/Translation.js
julen: They won't be better or worse unless you are microbenchmarking ;)

adngdb: Nice! Is there one that is more commonly used by the JS community?

frontend/src/modules/history/reducer.js
adngdb: Ooooh right, thank you!

frontend/src/modules/history/components/Translation.js
julen: Ever since they are available, I see template strings used widely. Their usefulness shines as more parameters are used as part of the string.

In your code above you'll want to avoid string concatenation if you ever plan to support i18n anyway.

adngdb:
I've filed bugs for the things I didn't want to address in this PR. :-) Everything else should have received a commit now.

> - You need to set dir, lang and data-script attributes for translations.

[Bug 1492163](https://bugzilla.mozilla.org/show_bug.cgi?id=1492163)

> - Whitespaces should be marked up.

[Bug 1492166](https://bugzilla.mozilla.org/show_bug.cgi?id=1492166)

pontoon/base/static/css/boilerplate.css
```diff
@@ -1,7 +1,7 @@
 /*
   HTML5 ✰ Boilerplate
 
-  style.css contains a reset, font normalization and some base styles.
+  style.css contains font normalization and some base styles.
```
mathjazz: Nit: that comment is still wrong - it talks about `style.css`, but this is `boilerplate.css`.

frontend/src/modules/entitydetails/components/Metadata.css
```diff
@@ -3,6 +3,7 @@
     color: #AAA;
     font-size: 0.9em;
     font-style: italic;
+    line-height: 22px;
```
mathjazz: This adds extra `line-height` to all #metadata elements, but we should also add it to the source string.

You can fix this by fixing this:
https://github.com/mozilla/pontoon/pull/1051#discussion_r213642105

pontoon/base/static/css/boilerplate.css
adngdb: Right, and my change was dumb: this file actually contains a CSS reset…

frontend/src/modules/entitydetails/components/Metadata.css
adngdb: I don't understand. The source string being in the #metadata component right now, this should work fine, right? When I move the source string outside of metadata, then I'll make sure it has the right line-height too.

mathjazz: It doesn't work fine witht the current markup, because it increases the `line-height` of all `#metadata` elements, including the original string.

If you compare the style with prod, you'll notice that we should increase line-height only for the original string, but not for other `#metadata` elements.

PR #1081, opened 2018-09-13
Fix bug 1486606: Show errors & warnings on dashboards

mathjazz:
This is the last patch in the effort to bring `compare-locales` checks to Pontoon. You can take it for a spin on stage:
https://mozilla-pontoon-staging.herokuapp.com/projects/

First, it restores disabled features that landed in previous patches:
* errors & warnings translation status in string list
* errors & warnings filters

Then, it finalizes errors & warnings implementation in translate view by:
* adding counts to filters and including Errors & Warnings within the Untranslated filter
* exposing errors & warnings in progress chart & menu

Finally, it shows errors & warnings on dashboards:
* in headings (progress chart and legend)
* in listings (Progress column)

Due to space limitations, some modifications were needed on dashboards:
* In headings, progress chart legend has been split into two columns.
* Long Progress column titles have been shortened using [Disemvoweling](https://en.wikipedia.org/wiki/Disemvoweling).
* Progress column numbers are smaller and use metric prefixes (e.g. `165.5k`).

Note on translation status: the newly added "Errors" and "Warnings" statuses (represented in string list with distinct colors, in dashboards, and as filters) only include approved and fuzzy translations (i.e. [synced translations](https://github.com/mozilla/pontoon/pull/1075/commits/818839a7e41394027cc8a545bd4d193b3ecad16e)).


flodolo:
I find this confusing. Why not move errors and warnings a separate column?

schermata 2018-09-13 alle 12 52 48

Once you reach a string with error or warning, there's no indication of what the problem is.

mathjazz:
> I find this confusing. Why not move errors and warnings a separate column?

The items in the left column are plotted on the circular progress chart, so they act as a legend. Each translation unit (source - translation pair) can be in one of these 5 states only. *All strings* and *Unreviewed* are not plotted on the chart, so they are treated separately.

I made that distinction more obvious in the latest update. Is that helpful?
screenshot_2018-09-14 firefox browser

The alternative would be to use the existing layout:
aa109217-10e0-40e3-a07e-9540957c3c76

> Once you reach a string with error or warning, there's no indication of what the problem is.

Working on it!

mathjazz:
> > Once you reach a string with error or warning, there's no indication of what the problem is.
> 
> Working on it!

Done! I'd love to hear your thoughts on this. And @Pike's.

flodolo:
> The items in the left column are plotted on the circular progress chart, so they act as a legend. Each translation unit (source - translation pair) can be in one of these 5 states. _All strings_ and _Unreviewed_ are not plotted on the chart, so they are treated separately.

That wasn't really obvious to me, since warnings/errors are orders of magnitude smaller than the other values, so they don't show up. But the current solution is much clearer.

Why ignoring errors/warnings when looking at approved and fuzzy? They're still approved/fuzzy, even with warnings or errors.

mathjazz:
> Why ignoring errors/warnings when looking at approved and fuzzy? They're still approved/fuzzy, even with warnings or errors.

In code they are (at least for now), but not in the UI (dashboards, string list, filters). Translation states are mutually exclusive, so each translation unit must be in one and only one state.

Without this change, you could click on `Approved: 3` in the dashboard and get 9 items in the string list.

mathjazz:
This is now ready for a review. @adngdb @jotes

pontoon/base/models.py
```diff
@@ -509,7 +509,13 @@ def get_top_instances(cls, qs):
             'most_suggestions': sorted(qs, key=lambda x: x.unreviewed_strings)[-1],
             'most_missing': sorted(
                 qs,
-                key=lambda x: x.total_strings - x.approved_strings - x.fuzzy_strings
+                key=lambda x: (
+                    x.total_strings -
```
jotes: [nit] I wonder if that's possible to make the property called `missing` and use it here and in the template.

pontoon/base/models.py
```diff
@@ -2811,8 +2847,8 @@ def serialize(self):
             'approved': self.approved,
             'rejected': self.rejected,
             'fuzzy': self.fuzzy,
-            'error_count': self.errors.count(),
-            'warning_count': self.warnings.count(),
+            'errors': [error.message for error in self.errors.all()],
```
jotes: [nit]  what do you think about`self.errors.values_list('message', flat=True)` here?

flodolo:
Question while doing screenshots: should these labels be shortened too?

string_overview

flodolo:
While testing this morning, there's one thing that is not super obvious: if you [land on a string with errors](https://mozilla-pontoon-staging.herokuapp.com/ilo/firefox/all-resources/?status=errors&string=74958), you wonder where the SUBMIT/SUGGEST button is (it appears as soon as you start typing).

pontoon/base/models.py
mathjazz: That sounds like a good idea, I'll look into it.

pontoon/base/models.py
mathjazz: That was my first idea, but it results in additional DB queries even if errors and warnings are prefetched, which is expensive.

For example, here:
https://github.com/mozilla/pontoon/blob/398e983e09d38bfd4dad2d943ab9bd5e6bddea58/pontoon/base/models.py#L2507

mathjazz:
> I missed this part. I find it really confusing: if a string is translated, and has errors/warning, it's not intuitive that it would show up in the untranslated filter.
> 
> Especially considering that warnings cannot be marked as false positives: I have 3 warnings of my own in Firefox, as have several other locales, and those strings are all but untranslated. It would really bother me that, every time I want to translate new content, these strings would show up in the list of strings to look at.

The Untranslated filter should return all strings that are not translated (in product). That includes Missing and Fuzzy (even if translated in Pontoon UI and VCS), and as of this bug, also Errors. Warnings, should not be included - good point.

Should I just make that change and remove Warnings from the list?

As a followup, we could add a new filters called "Needs Work", which would also include Warnings and Unreviewed.

Thoughts?

flodolo:
> The Untranslated filter should return all strings that are not translated (in product). That includes Missing and Fuzzy (even if translated in Pontoon UI and VCS), and as of this bug, also Errors. Warnings, should not be included - good point.
> 
> Should I just make that change and remove Warnings from the list?

Yep, that works for me. Good point in "not translated in product", I was thinking more on the line of "it's in VCS".

PR #1085, opened 2018-09-20
Fix bug 1492829: Include warnings in completion percent

mathjazz:
We're change the calculation of the completion percentage visible in progress charts on dashboards and the translate view.

**From:**
The number of Translated strings divided by the number of All strings [in %].

**To:**
The number of Translated strings and Strings with warnings divided by the number of All strings [in %].

That's because Warnings (unlike Errors) actually end up in product on build and are often false positives.


mathjazz:
@flodolo Could you please review this on stage?

flodolo:
It seems to be working fine to me 👍

PR #1086, opened 2018-09-21
Fix bug 1493137: Prevent duplicate translation submissions

mathjazz:
The existing pattern to prevent duplicate translation submission is insufficient ([details in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1493137)). Apparenty it aborts the request when the translation is already saved.

Note that the majority of the request time is occupied by the stats calculation, which takes places after saving the translation.

### DIFF
--- a/pontoon/base/static/js/translate.js
+++ b/pontoon/base/static/js/translate.js
@@ -2859,8 +2859,9 @@ var Pontoon = (function (my) {
         submittedPluralForm = 0;
       }
 
+      // Prevent duplicate translation submissions
       if (self.XHRupdateOnServer) {
-        self.XHRupdateOnServer.abort();
+        return;
       }
 
       self.XHRupdateOnServer = $.ajax({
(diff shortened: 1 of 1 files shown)


julen:
Is there any reason why the submission button is not disabled upon clicking it? That'd disallow immediate clicks firing multiple events in the first place (and that seems to be the way to reproduce this issue as mentioned in the bug.)

mathjazz:
The submission button is disabled upon clicking it.

In addition to that, this patch disables the AJAX request (which could also be triggered by the keyboard shortcut for example).

julen:
I might have misread the bug description — I was under the impression the submission event could happen twice.

pontoon/base/static/js/translate.js
```diff
@@ -2910,6 +2911,7 @@ var Pontoon = (function (my) {
           }
         },
         complete: function() {
+          self.XHRupdateOnServer = undefined;
```
adngdb: Setting to `undefined` is very weird and unusual to me. Why not set it to `null` instead?

PR #1088, opened 2018-09-23
implement Microsoft translator API v3 support

ernado:
Seems like v2 is deprecated.
This PR implements the basic support for v3 version of Microsoft translator.

I've just made a copy of `machine_translation` handler and adjusted it to
use the new v3 API, so now `machine_translation` is calling the `machine_translation_v3` if `MICROSOFT_TRANSLATOR_API_VERSION` equals to `3.0`.

### DIFF
--- a/pontoon/machinery/tests/conftest.py
+++ b/pontoon/machinery/tests/conftest.py
@@ -1,0 +1,17 @@
+import pytest
+
+
+@pytest.fixture
+def ms_api_key(settings):
+    """Set the settings.MICROSOFT_TRANSLATOR_API_KEY for this test"""
+    key = '1fffff'
+    settings.MICROSOFT_TRANSLATOR_API_KEY = key
+    return key
+
+
+@pytest.fixture
+def ms_locale(locale_a):
+    """Set the Microsoft API locale code for locale_a"""
+    locale_a.ms_translator_code = 'gb'
+    locale_a.save()
+    return locale_a
--- a/pontoon/machinery/tests/test_views.py
+++ b/pontoon/machinery/tests/test_views.py
@@ -19,6 +19,61 @@ from pontoon.test.factories import (
 
 
 @pytest.mark.django_db
+def test_view_machine_translation(client, ms_locale, ms_api_key):
+    url = reverse('pontoon.machine_translation')
+
+    with requests_mock.mock() as m:
+        data = [
+            {
+                'translations': [
+                    {'text': 'target'}
+                ]
+            }
+        ]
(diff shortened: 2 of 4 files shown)


Pike:
Drive-by question, is there a good reason to keep v2?

ernado:
I can't find one, that's why it is done in the way that we can rename `machine_translation_v3` to `machine_translation` and remove the old one.

But I'm not enough familiar with it to say exactly.

mathjazz:
Thanks for the patch, @ernado! See my comments below.

> Drive-by question, is there a good reason to keep v2?

In general, no. For the time being v2 still works for us, but not [very well](https://bugzilla.mozilla.org/show_bug.cgi?id=1481516). So we either need to upgrade to v3 or start using Google Translate instead, which supports more of the languages we need at Mozilla. I'll start an email thread.

Only once we make the decision, we can remove v2 support and merge this patch, because the v2 API KEY apparently won't work with v3. I'm getting the following response when I try to use it with the new codebase:
```json
{
    "error": {
        "code": 401000,
        "message": "The request is not authorized because credentials are missing or invalid."
    }
}
```

Regardless of the path forward, I'm fine with keeping the support for Microsoft Translator (v3) in the codebase should we or somebody else need it. Maybe you run your own install of Pontoon and need Microsoft Translator API v3 support, @ernado?

@ernado, in addition to removing v2 support, I'd also ask for a unit test of this function.

ernado:
Yes, the v3 does not support the v2 keys. 

And yes, I'm running my own install of Pontoon, but I'm ok to just host the v3-code in my fork which already have the optional support for Discord OAuth instead of FXA that is unlikely needed in codebase.

This PR is the way for me to ask "Do you need the v3 at all?" and allowing others to use V3 API if they are self-hosting too.

Thank you for comments, @mathjazz, I will remove the v2 and add the unit test during the week.

mathjazz:
Excellent, thanks!

ernado:
Hey @mathjazz, I've added some unit tests and completely replaced v2 with v3.
Please take a look.
Note that `MICROSOFT_TRANSLATOR_API_VERSION` is removed as unneeded.

mathjazz:
@ernado Thanks for the patch! We're in the process of obtaining the API key (sigh), and once we get it we can merge this.

mathjazz:
Update: We should be ready to merge this soon.

Deployment is blocked on [bug 1499512](https://bugzilla.mozilla.org/show_bug.cgi?id=1499512).

PR #1089, opened 2018-09-23
Bug 1489054 - Add entity list loader

jotes:
Hi,
I decided to start with the smallest possible component to get familliar with the code.

@mathjazz @adngdb could you review?

### DIFF
--- a/frontend/src/modules/entitieslist/components/EntitiesList.js
+++ b/frontend/src/modules/entitieslist/components/EntitiesList.js
@@ -12,6 +12,7 @@ import type { Navigation } from 'core/navigation';
 
 import { actions, NAME } from '..';
 import Entity from './Entity';
+import EntitiesLoader from './EntitiesLoader'
 
 import type { Entities, DbEntity } from '../reducer';
 
@@ -20,6 +21,7 @@ type Props = {|
     entities: {|
         entities: Entities,
         hasMore: boolean,
+        fetching: boolean,
     |},
     parameters: Navigation,
 |};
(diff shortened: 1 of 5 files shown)


mathjazz:
Makes sense!

frontend/src/modules/entitieslist/components/EntitiesList.js
```diff
@@ -95,12 +105,15 @@ export class EntitiesListBase extends React.Component<InternalProps> {
         const { entities } = this.props;
         const selectedEntity = this.props.parameters.entity;
 
+        // InfiniteScroll will display information about loading during the request
+        const hasMore = entities.fetching || entities.hasMore;
```
adngdb: I don't understand this. What does `fetching` have to do with `hasMore`? How is the fact that we're fetching the next batch of results related to the question of whether it can load more results? And even if that's the case, I don't understand the logic being used here.

jotes: Hey, from what I understand and from what I was able to notice during testing -> `hasMore` is responsible for two things:
* if it's false, InfiniteScroller will not attach event handlers related to handle scrolling and progressive loading
* If it's true, InfiniteScroller will display Loader.

`translate.next` resets hasMore to False at the start of every fetch and sets it again after response from server.

And because of this (presumely) I was unable to catch loading animation at all because before I scrolled to the bottom of the page, `hasMore` was false, and then entities magically jumped into the view.

I may not understand the flow of actions and states correctly, I just tried to make this the same as in in the old translate.view. From what I remember (and checked) previews translate view displays loading animation even during fetching data from the server.

I'm sorry If I overexplained, I'm just not sure if my approach is good and if what I see is really what I should see ;D

adngdb: Thanks @jotes, I suspected that might be the case, I'm glad you investigated that. If it's the behavior of the library, then so be it. It kind of makes sense, too (even though I'm not fond of that behavior, as I dislike knowing that the animation is running pretty much all the time, not sure how that impacts browser performance).

Given that I needed a lengthy explanation though, I think we need a more comprehensive comment here. Can you rephrase what you explained in the code please?

PR #1090, opened 2018-09-25
Fix bug 1490349 - Add current user data to Translate app. 

adngdb:
This data will be used later on by other modules to give access to actions, show special data, and display the name and avatar of the user.

### DIFF
--- a/frontend/src/core/user/components/UserAutoUpdater.js
+++ b/frontend/src/core/user/components/UserAutoUpdater.js
@@ -1,0 +1,60 @@
+/* @flow */
+/* globals IntervalID */
+
+import * as React from 'react';
+import { connect } from 'react-redux';
+
+import { actions } from '..';
+
+
+type Props = {
+};
+
+type InternalProps = {
+    ...Props,
+    dispatch: Function,
+};
+
+type State = {|
+    timer: ?IntervalID,
+|};
+
+
+/**
+ * Regularly fetch user data to keep it up-to-date with the server.
+ */
+export class UserAutoUpdaterBase extends React.Component<InternalProps, State> {
+    constructor() {
+        super();
+        this.state = {
+            timer: null,
+        };
+    }
+
+    fetchUserData = () => {
+        this.props.dispatch(actions.get());
+    }
+
+    componentDidMount() {
+        this.fetchUserData();
+        const timer = setInterval(this.fetchUserData, 2 * 60 * 1000);
+        this.setState({ timer });
+    }
+
+    componentWillUnmount() {
+        if (this.state.timer) {
+            clearInterval(this.state.timer);
+        }
+    }
+
+    render() {
+        return null;
+    }
+}
+
+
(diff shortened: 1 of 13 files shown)


frontend/src/core/user/components/UserAutoUpdater.js
```diff
@@ -0,0 +1,60 @@
...
+    dispatch: Function,
+};
+
+type State = {|
+    timer: ?IntervalID,
+|};
+
+
+/**
+ * Regularly fetch user data to keep it up-to-date with the server.
```
mathjazz: Since we don't have this functionality in the existing translate view, I'm curiouis what's the use case for this?

pontoon/base/views.py
```diff
@@ -788,6 +788,26 @@ def download_translation_memory(request, locale, slug, filename):
...
+        })
+
+    return JsonResponse({
+        'is_authenticated': True,
+        'id': user.id,
+        'email': user.email,
+        'display_name': user.display_name,
+        'manager_for_locales': list(user.managed_locales.values_list('code', flat=True)),
+        'translator_for_locales': list(user.translated_locales.values_list('code', flat=True)),
+        'translator_for_projects': user.translated_projects,
```
mathjazz: There's more user data passed to `translate.html` and `translate.js`:
* `data-force-suggestions`
* `data-tour-status`
* `data-locales-order`

Shouldn't we include that, too?

There are two other pieces of data, but it doesn't look like they are used:
* `data-username`
* `data-manager`

frontend/src/core/user/components/UserAutoUpdater.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    fetchUserData = () => {
+        this.props.dispatch(actions.get());
+    }
+
+    componentDidMount() {
+        this.fetchUserData();
+        const timer = setInterval(this.fetchUserData, 2 * 60 * 1000);
+        this.setState({ timer });
```
julen: I might be missing something, but why is the timer stored as part of state? Setting it here will force a re-render of the component, and even though it renders `null`, IMHO it seems more natural to store the timer reference as a property of the component instance, because state provides no additional value needed to implement this logic.

frontend/src/core/user/components/UserAutoUpdater.js
adngdb: That's a very good question. The reason is that on processeer, I have problems with the session expiring quite quickly, but the local state doesn't know about it. So, it regularly happens that I want to save a change and can't because I'm logged out. I thus wanted to make sure that our data about the user stays up-to-date (to a certain extent), so that we don't give them a false experience (like, a Save button when they're actually logged out of the server).

However, Pontoon sessions run a lot longer, and users are very rarely disconnected. It is possible that this feature is useless. What do you think?

pontoon/base/views.py
mathjazz: SGTM

PR #1099, opened 2018-10-12
Reduce fix_stats() log verbosity

mathjazz:
total_strings mismatch in ProjectLocales of the same Project is
sometimes expected, because not all strings (files) need to be exposed
to all locales. So we shouldn't log every time we find such mismatch.


mathjazz:
r=me

PR #1103, opened 2018-10-16
Fix bug 1497858: Expose localizable files through project config

mathjazz:
We expose localizable files (on dashboards and in translate view) by creating corresponding `TranslatedResource` objects. So if a `TranslatedResource` object exists for a given `Resource` and `Locale`, the file is localizable.

The way this works is documented in the [bug description](https://bugzilla.mozilla.org/show_bug.cgi?id=1497858#c0), alongside with the limitations, which are mitigated with this patch.

This PR is split into 3 pieces:
1. https://github.com/mozilla/pontoon/commit/297295851bf45c5f7c8f9474f1c1c5789ee71047 just adds the `Project.configuration_file` to the data model and admin.
2. https://github.com/mozilla/pontoon/commit/719913b250c1f7dc20fe9e9d003b9a71cb790be0 is even simpler - it updates `compare-locales`.
3. https://github.com/mozilla/pontoon/commit/08b9f265b11324f805f284c962c4bcc29ffa608d is the actual bugfix. It splits the existing `update_translated_resources()` method into two methods, based on how `TranslatedResource` instance are created: using the existing algorithm or through [project configuration](https://moz-l10n-config.readthedocs.io/en/latest/fileformat.html).

Before I add unit tests, I'd like to hear your thoughts on code structure and potentially refactor it.

To test I used [this simple project](https://github.com/mathjazz/moz-l10n-config-test), which should be used to demonstrate the ability to expose monolingual files only to a subset of locales, defined in project configuration file.


pontoon/base/models.py
```diff
@@ -1150,6 +1150,16 @@ class Project(AggregatedStats):
         Allow localizers to request the project for their team.
     """)
 
+    configuration_file = models.CharField(
+        null=True,
+        blank=True,
+        max_length=2000,
+        help_text="""
+        A path to the optional project configuration file, relative to the
+        source string repository.
```
adngdb: nit: this should be indented one more block.

pontoon/sync/vcs/models.py
```diff
@@ -431,6 +438,69 @@ def resources_for_path(self, path):
...
+        return parser.parse(path).configs
+
+    @cached_property
+    def paths(self):
+        """Return paths specified in the configurations."""
+        return [
+            paths
+            for configuration in self.configs
+            for paths in configuration.paths
+        ]
```
Pike: The configurations can also be locale-dependent. For example, we could have a smaller list of locales in `_configs/devtools-client.toml`.

pontoon/sync/vcs/models.py
```diff
@@ -431,6 +438,69 @@ def resources_for_path(self, path):
...
+
+        for resource in self.vcs_project.db_project.resources.all():
+            absolute_resource_path = os.path.join(
+                self.vcs_project.source_directory_path,
+                resource.path,
+            )
+
+            for path in self.locale_paths(locale):
+                if path.match(absolute_resource_path) is not None:
+                    resources.append(resource)
```
Pike: Pretty sure this can create false-positives if `locale` and `l10n_base` aren't set in the matching environment.

pontoon/base/models.py
mathjazz: Huh, we keep making this mistake consistenly across the entire `models.py` file. Maybe we should take care of it as part of https://bugzilla.mozilla.org/show_bug.cgi?id=1415732.

mathjazz:
@Pike Thanks for the pro-tip about `ProjectFiles`! It surely simplifies things. And does its job well.

How do I set `l10n_base`? By passing `mergebase` to `ProjectFiles`?

Is that even necessary here?

Pike:
You add that by passing an env to 

```python
TOMLParser().parser('l10n.toml', env={'l10n_base':'local_path'})
```

pontoon/sync/vcs/models.py
```diff
@@ -459,10 +459,15 @@ def project(self):
...
-            self.configuration_file
+            self.configuration_file,
         )
-        parser = TOMLParser()
-        return parser.parse(path)
+
+        # Use parent directory of the first locale directory as l10n_base
+        l10n_base = get_parent_directory(
+            self.vcs_project.locale_directory_paths.itervalues().next()
+        )
```
mathjazz: Until [bug 1497851](https://bugzilla.mozilla.org/show_bug.cgi?id=1497851) is fixed.

mathjazz:
@Pike This is ready for another look.

I also added a testing project config checkout in https://github.com/mozilla/pontoon/pull/1103/commits/651f18df34ca5c3d5584208b5ec11b929357d90e. Only the `config.toml` file is used for now.

pontoon/sync/vcs/models.py
Pike: I'm hand-waving, but I think this is not right. `source_directory_path` might be closer, and is a less scary code path? Or, `self.vcs_project.db_project.source_repository.checkout_path`, which I think should be correct for non-gecko apps.

mathjazz:
@Pike One more look?

pontoon/sync/vcs/models.py
```diff
@@ -431,6 +441,53 @@ def resources_for_path(self, path):
...
+    def parsed_configuration(self):
+        """Return parsed project configuration file."""
+        db_project = self.vcs_project.db_project
+
+        path = os.path.join(
+            db_project.source_repository.checkout_path,
+            self.configuration_file,
+        )
+
+        l10n_base = db_project.translation_repositories()[0].checkout_path
```
mathjazz: Source code for [`translation_repositories`](https://github.com/mozilla/pontoon/blob/981114359923d7656f563b696af36c94392283b7/pontoon/base/models.py#L1382) and [`checkout_path`](https://github.com/mozilla/pontoon/blob/981114359923d7656f563b696af36c94392283b7/pontoon/base/models.py#L1669).

Example paths for some projects:
```python
>>> Project.objects.get(slug="firefox").translation_repositories()[0].checkout_path
u'/app/media/projects/firefox/l10n-central'

>>> Project.objects.get(slug="mozillaorg").translation_repositories()[0].checkout_path
u'/app/media/projects/mozillaorg/git@github.com:mozilla-l10n/www.mozilla.org.git'

>>> Project.objects.get(slug="amo").translation_repositories()[0].checkout_path
u'/app/media/projects/amo/git@github.com:mozilla/olympia.git'
```

mathjazz: Note: We also have locales coming from bitbucket repositories, for which the checkout path is different.

pontoon/sync/tests/project-config-checkout/l10n.toml
```diff
@@ -0,0 +1,20 @@
+basepath = "."
+
+locales = [
+    "fr",
+    "sl",
+]
+
+[env]
+    l = "{locale}"
```
Pike: This should be `l = "{l10n_base}/{locale}/"` including the trailing dash

pontoon/sync/tests/project-config-checkout/l10n.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[[paths]]
+    reference = "en-US/strings.properties"
+    l10n = "{l}/strings.properties"
+
+[[paths]]
+    reference = "en-US/strings_reality.properties"
+    l10n = "{l}/strings_reality.properties"
```
Pike: In both `[[paths]]`, `{l}strings...`.

PR #1104, opened 2018-10-18
Fix bug 1499512: Add support for Google Cloud Translation API

mathjazz:
This patch doesn't only add support for Google Translate via Google Cloud Translation API, but also makes it possible for Pontoon deployments to use multiple machine translation services (Microsoft Translator is also supported).

We'll be switchin to using Google Translate on `pontoon.mozilla.org`, which will increase the number of supported locales from 48 to 103. All existing locales will remain supported, so we're adding 55 new ones. All locales should also work much more reliably, because

**TODO:** Check with legal if text attribution and Disclaimer notice meet the [attribution requirements](https://cloud.google.com/translate/attribution) or do we also need a graphic. The document says "using the appropriate text or brand elements", but doesn't seem to be coherent.


pontoon/machinery/views.py
```diff
@@ -134,6 +134,53 @@ def microsoft_translator(request):
         return HttpResponseBadRequest('Bad Request: {error}'.format(error=e))
 
 
+def google_translate(request):
+    """Get translation from Google machine translation service."""
+    try:
+        text = request.GET['text']
+        locale_code = request.GET['locale']
+    except MultiValueDictKeyError as e:
+        return HttpResponseBadRequest('Bad Request: {error}'.format(error=e))
```
adngdb: Possible improvement: move this code into a `Form` and share it between `google` and `microsoft` views.

pontoon/machinery/views.py
```diff
@@ -134,6 +134,53 @@ def microsoft_translator(request):
...
+    try:
+        r = requests.post(url, params=payload)
+        root = json.loads(r.content)
+        translation = root['data']['translations'][0]['translatedText']
+        obj = {
+            'locale': locale_code,
+            'translation': translation,
+        }
+
+        return JsonResponse(obj)
```
adngdb: I'd not bother using `obj` here and just return `JsonResponse({ content })`.

pontoon/machinery/views.py
```diff
@@ -134,6 +134,53 @@ def microsoft_translator(request):
...
+
+    except Exception as e:
+        return HttpResponseBadRequest(
```
adngdb: Mmm, that's a bit problematic. What if there's an error in our code? Then instead of rightfully returning a 500 we would return a 400, and that would make it a lot more difficult for us to catch that problem. It's generally a bad idea to `exception Exception` as it hides a lot of potentially important information. Can you catch a better defined list of exceptions here, and let the others pass? I'd rather we raise some 500 and see them in the logs, then improve this code, than the other way around.

pontoon/base/views.py
```diff
@@ -79,6 +81,12 @@ def translate(request, locale, slug, part):
         'part': part,
         'project': project,
         'projects': projects,
+        'is_google_translate_supported':
+            True if settings.GOOGLE_TRANSLATE_API_KEY
+            else False,
+        'is_microsoft_translator_supported':
+            True if settings.MICROSOFT_TRANSLATOR_API_KEY
+            else False,
```
adngdb: Both of these can be simplified to `bool(settings.XXX_SERVICE_API_KEY)`.

pontoon/machinery/views.py
```diff
@@ -41,6 +41,12 @@ def machinery(request):
     return render(request, 'machinery/machinery.html', {
         'locale': Locale.objects.get(code=locale),
         'locales': Locale.objects.all(),
+        'is_google_translate_supported':
+            True if settings.GOOGLE_TRANSLATE_API_KEY
+            else False,
+        'is_microsoft_translator_supported':
+            True if settings.MICROSOFT_TRANSLATOR_API_KEY
+            else False,
```
adngdb: Same here, please use `bool`.

pontoon/base/views.py
mathjazz: Thank you!

pontoon/machinery/urls.py
```diff
@@ -18,11 +18,11 @@
...
     url(r'^translation-memory/$', views.translation_memory,
         name='pontoon.translation_memory'),
-    url(r'^machine-translation/$', views.machine_translation,
-        name='pontoon.machine_translation'),
-    url(r'^machine-translation-caighdean/$',
-        views.machine_translation_caighdean,
-        name='pontoon.machine_translation_caighdean'),
+    url(r'^microsoft-translator/$', views.microsoft_translator,
+        name='pontoon.microsoft_translator'),
+    url(r'^caighdean/$',
```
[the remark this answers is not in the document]
mathjazz: The naming I'm using is based on official service names (e.g. Google Translate, Microsoft Translator), and it's used consistently in the models (locale codes), templates (what service is enabled), views (method names), URLs, JS, even CSS.

There are also other services that could be grouped similarly as MT services as far as naming is concerned (e.g. terminology and TM services). But I'm not sure how much sense does it make to group these services that way. 

I believe Translate.Next (post React rewrite) will be a good opportunity to rethink how we use these services and look at the codebase and refactor it. https://github.com/mozilla/pontoon/pull/1104#discussion_r226244540 also fits there.

pontoon/machinery/views.py
mathjazz: See: https://github.com/mozilla/pontoon/pull/1104#discussion_r226265539.

mathjazz:
@adngdb Thanks for a great review! I've addressed your feedback and added two comments. This is now ready for another look.

mathjazz:
Also added tests now.

PR #1109, opened 2018-10-22
[WIP] Bug 1489054 -  Add loader for the entity list 

jotes:
hey @adngdb,

In order to add the loader for the EntityList, I'm planning to add a reducer which will listen to all redux actions and record all REQUEST/RECEIVE/ERRORS actions and add components connected to those actions to the LoaderState. Later I'm planning to add a Redux state selector (which will wait until EntityList will receive response from the backend) which will render the Loader component.

I think the same pattern could apply to the future loading components.

Feel free to suggest better approaches/point me into the better directions :-)

TODO:
* [ ] - Discuss the approach
* [ ] - Add the Loader Component
* [ ] - Create a selector
* [ ] - Add tests

### DIFF
--- a/frontend/src/modules/loader/index.js
+++ b/frontend/src/modules/loader/index.js
@@ -1,0 +1,6 @@
+export { default as reducer } from './reducer';
+
+
+// Name of this module.
+// Used as the key to store this module's reducer.
+export const NAME: string = "loader";
--- a/frontend/src/modules/loader/reducer.js
+++ b/frontend/src/modules/loader/reducer.js
@@ -1,0 +1,56 @@
+/* @flow */
+
+export const RECEIVE: 'RECEIVE' = 'RECEIVE';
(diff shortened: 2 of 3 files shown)


adngdb:
@jotes I don't understand what you're trying to achieve with this reducer. Are you talking about the loader for the page load, the one that looks like growing rectangles? If so, then I believe it should simply be a test in the `App` component. If the specific data we need to be able to display the app is missing, then return the Loader in the `render` method.

Something _à la_:

```js
class App extends React.Component {
    render() {
        const state = this.props;

        if (state.entities.fetching || state.locales.fetching) {
            return <Loader />;
        }

        return <div>
            Content
        </div>;
    }
}
```

Am I missing something here? I don't understand why you want to use a reducer.

jotes:
I'm closing this for now, I'll open a new PR when Loader will be finished.

PR #1110, opened 2018-10-23
Fix bug 1501088: Use text format for Google Translate

mathjazz:


mathjazz:
r+ed by @adngdb over IRC.

mathjazz:
We got some positive feedback from localizers and no objections against using the new behaviour. Let's go ahead with this.

It's also consistent with how Microsoft Translator API is/was used.

PR #1111, opened 2018-10-24
Fix bug 1473022: Add support for XML file format used in Android

mathjazz:
In progress ...


pontoon/sync/formats/compare_locales.py
```diff
@@ -64,6 +78,32 @@ def __init__(self, path, source_resource=None):
...
 
+    def save(self, locale):
+        if not self.source_resource:
+            raise SyncError(
+                'Cannot save resource {0}: No source resource given.'
+                .format(self.path)
+            )
+
+        # Parse source resource
+        self.source_resource.parser.readFile(self.source_resource.path)
```
mathjazz: I need to investigate why do we need to read the reference file again. I was expecting `list(self.source_resource.parser.walk())` to JustWork™, but for some reason `self.source_resource.parser.ctx.contents` is empty.

pontoon/sync/formats/compare_locales.py
```diff
@@ -39,6 +43,16 @@ def __init__(self, path, source_resource=None):
...
 
+        # Copy entities from the source_resource if it's available.
+        if source_resource:
+            for key, entity in source_resource.entities.items():
+                self.entities[key] = CompareLocalesEntity(
+                    entity.key,
+                    '',
+                    None,
+                    entity.order
+                )
```
mathjazz: TODO: Look into refactoring the code so that this block is not required.

mathjazz: There are several places where the code expects l10n resources to contain all entities, even if they don't exist in the file, e.g.:
* `pontoon.sync.core.py:update_translations()`
* `pontoon.base.utils.py:get_download_content()`
* `pontoon.base.utils.py:handle_upload_content()`

pontoon/sync/formats/compare_locales.py
```diff
@@ -0,0 +1,116 @@
...
+            # If the file doesn't exist, but we have a source resource,
+            # we can keep going, we'll just not have any translations.
+            if source_resource:
+                return
+            else:
+                raise ParseError(err)
+
+        entities = list(self.parser.parse())
+        for order, entity in enumerate(entities):
+            if isinstance(entity, parser.Entity):
```
mathjazz: We need this check in case the file does not exists. In which case the parser returns `Junk`.

pontoon/sync/formats/compare_locales.py
```diff
@@ -0,0 +1,116 @@
...
+        self.fuzzy = False
+        self.source = []
+
+
+class CompareLocalesResource(ParsedResource):
+    def __init__(self, path, source_resource=None):
+        self.path = path
+        self.entities = OrderedDict()  # Preserve entity order.
+        self.source_resource = source_resource
+        self.parser = copy.deepcopy(parser.getParser(self.path))
```
mathjazz: We need to `deepcopy` here or else parser contexts from l10n resource and reference resource override each other.

We should add a comment, but I hope there's a better way of implementing this. :)

mathjazz:
Thanks for the hint.

I created a new structure called `CompareLocalesResource.parsed_objects` to avoid messing with `CompareLocalesResource.entities`, which only contains translations.

pontoon/sync/formats/compare_locales.py
```diff
@@ -0,0 +1,118 @@
...
+class CompareLocalesEntity(VCSTranslation):
+    """
+    Represents an entity in a file handled by compare-locales.
+    """
+    def __init__(self, key, string, comment, order):
+        self.key = key
+        self.source_string = string
+        self.source_string_plural = ''
+        self.strings = {None: self.source_string} if self.source_string is not None else {}
+        self.comments = [c.strip() for c in comment.raw_val.split('\n')] if comment else []
```
mathjazz: Once `compare-locales` is updated to a version incorporating fix for [bug 1502009](https://bugzilla.mozilla.org/show_bug.cgi?id=1502009), this line should be changed to:
```python
self.comments = comment.raw_val.split('\n') if comment else []
```

mathjazz:
This PR is now complete with the exception of bumping `compare-locales` version when it's released and updating a test to accomodate the changes: https://github.com/mozilla/pontoon/pull/1111#discussion_r230048783.

The gist of the PR is the [`pontoon/sync/formats/compare_locales.py`](https://github.com/mozilla/pontoon/pull/1111/files#diff-b9fa87fecf315b0a463be6ca0b4d22cb) file, so I'm predominantly interested in its review.

The related tests are here:
https://github.com/mozilla/pontoon/pull/1111/files#diff-b9fa87fecf315b0a463be6ca0b4d22cb

pontoon/sync/tests/formats/test_compare_locales.py
```diff
@@ -0,0 +1,254 @@
...
+</resources>
+        """)
+
+        self.run_save_no_changes(input_string, input_string, source_string=source_string)
+
+    def test_save_translation_missing(self):
+        source_string = dedent("""<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="String">Source String</string>
+    <string name="Missing String">Missing Source String</string>
```
Pike: Can you add a comment for "Missing String"? It should also show up in the serialized file, it'd be good to have a check for that here as well.

mathjazz:
Thanks for the review! 👍 

@flodolo @TheoChevalier @MikkCZ This patch adds support for the [Android XML file format](https://developer.android.com/guide/topics/resources/string-resource) to Pontoon through a not-yet-released version of `compare-locales` parser and serializer. 

Before we cut the release of `compare-locales` and ship this patch, @Pike and I would appreciate if you could take it for a spin on stage:
https://mozilla-pontoon-staging.herokuapp.com/projects/android/

We're mostly interested in checking if strings are imported and serialized as expected.

TheoChevalier:
Thanks @mathjazz ! Checked French and I can’t see anything wrong. The only thing I’m not 100% sure, is whether the new line is okay here (and at the bottom of the file) and works like en-US https://mozilla-pontoon-staging.herokuapp.com/fr/android/strings.xml/?string=191732

Pike:
> Thanks @mathjazz ! Checked French and I can’t see anything wrong. The only thing I’m not 100% sure, is whether the new line is okay here (and at the bottom of the file) and works like en-US https://mozilla-pontoon-staging.herokuapp.com/fr/android/strings.xml/?string=191732

That should work, I think.

Pike:
One thing I noticed is that we're not yet running compare-locales checks.

mathjazz:
Thanks, @TheoChevalier!

Looks like the source file and the translation file use new line differently, so I think we're good:
https://github.com/mozilla-l10n/mobile-android/blob/master/en-US/strings.xml#L294
https://github.com/mozilla-l10n/mobile-android/blob/master/fr/strings.xml#L18

...working on enabling c-l checks...

mathjazz:
I've updated the branch and stage with compare-locales checks.

Pike:
I think we should also de-duplicate tt checks, like we do for properties here:

https://github.com/mozilla/pontoon/blob/b2a3cd615eb8db614f2617d607d91af349731c60/pontoon/checks/libraries/__init__.py#L49-L56

Maybe be even more agressive on some? Like the double-space and newline checks sound counter-helpful

flodolo:
I'm a bit surprised that this (replacing `%1$s` with `%1$S`) doesn't trigger an error?
https://mozilla-pontoon-staging.herokuapp.com/it/android/all-resources/?string=191740

Also, looks like machinery is a bit broken for [some strings](https://mozilla-pontoon-staging.herokuapp.com/it/android/all-resources/?string=191556)

schermata 2018-11-06 alle 14 40 16

Pike:
> I'm a bit surprised that this (replacing `%1$s` with `%1$S`) doesn't trigger an error?

https://bugzilla.mozilla.org/show_bug.cgi?id=1504636

mathjazz:
> I think we should also de-duplicate tt checks, like we do for properties here:

Done. Stage should be updated in a few minutes.

> Also, looks like machinery is a bit broken for [some strings](https://mozilla-pontoon-staging.herokuapp.com/it/android/all-resources/?string=191556)

I actually think it works as expected: [this string](https://mozilla-pontoon-staging.herokuapp.com/it/all-projects/all-resources/?search=please+contact+the+website+owners&string=78887) is used as source.

mathjazz:
I'm curious, aren't [strings like this](https://mozilla-pontoon-staging.herokuapp.com/it/android/all-resources/?string=191708) localizability bugs? Spaces, newlines, tabs and even markup are not really helping localizers.

Pike:
That string is used in the same context as https://mozilla-pontoon-staging.herokuapp.com/it/android/all-resources/?string=191518, and that's basically the same thing as https://pontoon.mozilla.org/de/firefox/browser/chrome/overrides/netError.dtd/?string=78875

In Fluent, we're just finishing up https://github.com/projectfluent/fluent/issues/162, which will add localizer indent, for use-cases like error fragments.

flodolo:
> I actually think it works as expected: [this string](https://mozilla-pontoon-staging.herokuapp.com/it/all-projects/all-resources/?search=please+contact+the+website+owners&string=78887) is used as source.

I clearly never realized how confusing the machinery is for this type of strings :-\

MikkCZ:
Sorry for my late late reaction.

I went through Czech in staging and what I have noticed is inconsistent handling of newline character. Somewhere it is visualized as `¶` (entities list and history), somewhere as `\n` (source string), somewhere as whitespace (also some source strings).

https://screenshots.firefox.com/zj94asRullMbzrqp/mozilla-pontoon-staging.herokuapp.com
https://screenshots.firefox.com/7W54TEppRRrJ8SBu/mozilla-pontoon-staging.herokuapp.com

The inconsistent view is OK, however the problem I see that `\n`is highlighted as a placeable and clicking it copies it into the translation. Instead there should be a newline and newline should be pasted into the translation. Serialization as `\n` should be something happening on the backend.

Pike:
These are actually inconsistencies in the en-US data for the most part, https://github.com/mozilla-l10n/mobile-android/blob/master/en-US/strings.xml is a hotchpotch of newlines and \n and both :-/

mathjazz:
Thanks @Pike!

@MikkCZ A small part of this differences is also caused by [bug 1368602](https://bugzilla.mozilla.org/show_bug.cgi?id=1368602).

PR #1112, opened 2018-10-24
Fix bug 1501168: Upload FTL strings without comments

mathjazz:
We store FTL translations as serialized FTL messages with removed
comments (which are also part of the AST). Pontoon's FTL parser removes
comments from translations, and keeps them in source strings.

To detect whether a file contains translations or source strings, it
checks if the source resource has been provided. This patch makes sure
that is the case on file load, which means translations will be
imported without comments.

### DIFF
--- a/pontoon/base/utils.py
+++ b/pontoon/base/utils.py
@@ -475,7 +475,11 @@ def handle_upload_content(slug, code, part, f, user):
         for chunk in f.chunks():
             temp.write(chunk)
         temp.flush()
-        resource_file = formats.parse(temp.name)
+        # Bilingual file format parsers could rely on source resources being
+        # passed when parsing translated resource (see bug 1501168). Since we
+        # don't have a source resource here, we can fake it by using translated
+        # resource also as a source resource.
+        resource_file = formats.parse(temp.name, temp.name)
 
     # Update database objects from file
     changeset = ChangeSet(


pontoon/base/utils.py
```diff
@@ -475,7 +475,11 @@ def handle_upload_content(slug, code, part, f, user):
         for chunk in f.chunks():
             temp.write(chunk)
         temp.flush()
-        resource_file = formats.parse(temp.name)
+        # Bilingual file format parsers could rely on source resources being
+        # passed when parsing translated resource (see bug 1501168). Since we
+        # don't have a source resource here, we can fake it by using translated
+        # resource also as a source resource.
+        resource_file = formats.parse(temp.name, temp.name)
```
adngdb: I understand nothing about this code, and that makes me think that this variable is quite poorly named. `temp` doesn't tell me anything about what the variable contains, unless I scroll up to find where it is created.

Also, I don't understand the comment, but that is because I lack knowledge about Fluent and parsers in general, I assume. 

All that to say that: if you feel like taking this opportunity to make the code a bit more readable, that would be lovely (but it's not a blocker), and I am not being a good reviewer here as I don't understand what this is doing at all. :D

mathjazz: If you look 3 lines above the fold, you'll see a comment explaining `temp`. It's a temporary file created from the uploaded file.

I also think this is one of the [rare cases](https://stackoverflow.com/a/21944302) were `temp` is not a bad variable name. Note that `tempfile` is the name of the library. Should I rename it to `temp_resource` or `temp_file`?

`resource_file` is an instance of a `ParsedResource`, which represents a parsed file. Should I rename it to `parsed_resource` or `parsed_file`?

I'll also try to make the comment more explanatory.

adngdb: I'm not sure anymore, as I understand the code better, and it kinda makes sense to use `temp` here. I still don't like it, but once again, I'm not pushing it and I'll be fine if you decide to just merge as is.

mathjazz:
We should solve this problem differently:
https://bugzilla.mozilla.org/show_bug.cgi?id=1501168#c2

Sorry for the confusion. But at least @adngdb understands the code better. :)

PR #1114, opened 2018-10-25
Fix bug 1501262: Upgrade to Fluent Syntax 0.7

mathjazz:
I updated fluent just by bumping the libraries.

The test file behaves as expected:
https://github.com/mozilla-l10n/pontoon-ftl/blob/master/en-US/demo.ftl

I also updated the docs.


mathjazz:
> Does Pontoon use `package-lock.json`? If so, it should probably also get an update in this PR.

Thanks, I also updated the docs.

@adngdb Is there a way to automate updating `package-lock.json` when `package.json` is updated?

In virtualenv `npm install` did the trick.

adngdb:
`make build` won't be able to make any modifications to the lock file, it's not what it's made for. I believe the only solution as of today is indeed to run `yarn install` outside docker, and then remove the `node_modules` folder (so it doesn't collide with the one in docker, though that might not be needed, not sure).

PR #1115, opened 2018-10-28
Fix migration 0120 for Fluent 0.7

jotes:
Hey,
It looks like the latest update of Fluent broke one of migrations.
I see the following error: https://gist.github.com/jotes/6c15a57a152d478a53131772e9eb8f5a

@mathjazz could you confirm the issue and review this patch?


mathjazz:
Thanks!

PR #1116, opened 2018-10-29
No bug - Update to create-react-app 2.

adngdb:
The various changes in this patch are consequences of the updates of React and Flow. The important update was `react-scripts` which is the core piece of `create-react-app`.

### DIFF
--- a/frontend/src/core/plural/components/PluralSelector.js
+++ b/frontend/src/core/plural/components/PluralSelector.js
@@ -80,10 +80,10 @@ export class PluralSelectorBase extends React.Component<InternalProps> {
             <ul>
                 { locale.cldrPlurals.map((item, i) => {
                     return <li key={ item } className={ i === pluralForm ? 'active' : '' }>
-                        <a onClick={ () => this.selectPluralForm(i) }>
+                        <button onClick={ () => this.selectPluralForm(i) }>
                             <span>{ CLDR_PLURALS[item] }</span>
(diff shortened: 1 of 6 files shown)


adngdb:
Note that this is a blocker for #1113.

frontend/src/core/plural/components/PluralSelector.js
```diff
@@ -80,10 +80,10 @@ export class PluralSelectorBase extends React.Component<InternalProps> {
             <ul>
                 { locale.cldrPlurals.map((item, i) => {
                     return <li key={ item } className={ i === pluralForm ? 'active' : '' }>
-                        <a onClick={ () => this.selectPluralForm(i) }>
+                        <button onClick={ () => this.selectPluralForm(i) }>
                             <span>{ CLDR_PLURALS[item] }</span>
                             <sup>{ examples[item] }</sup>
-                        </a>
+                        </button>
```
mathjazz: I'm curious what's the purpose of this change.

adngdb: React raised a warning, using a `<a>` with a `href` attribute is considered a bad pattern, and they are right: `<a>` should be links, leading to a different page or at least a URL change. In this case, it was not doing that, so it should have been a `<button>` from the start.

adngdb:
Tests are passing now! :-)

PR #1119, opened 2018-11-01
No bug - Mark migration 0120 as a irreversible, next fixes for Fluent 0.7

jotes:
Hello again,
I recreate my Pontoon database from time to time and sadly, I spotted the next issue with the migration 0120 (hopefully it's the last one).
In my humble opinion (and after some manual testing) this migration is irreversible because of changes on unique indexes. I couldn't force this migration to run after a downgrade to 0119.

That could overshadow the next issue: `ast.VariantName` has been removed in the Fluent 0.7.

I'm not sure if `ast.Identifier` is the good replacement (based on https://github.com/projectfluent/python-fluent/commit/a76040a1dc5c5f7a2965df4d60f4dd77801aafc8), so I would love to get your feedback.

@mathjazz @Pike could you look at this? Thanks in advance!


mathjazz:
The changes look good to me.

This migration has more potential for breaking in the future due to the changes to the Fluent syntax and AST. @Pike Should we just remove it?

Pike:
How would you do that? Just make this migration a no-op? That might be just as fine.

jotes:
Other possible  solutions:
* add tests for migrations 
* squash all migrations (troublesome/risky)

I feel that the returning question is how much do We want to support standalone instances of Pontoon (I don't have a strong opinion. I only see some of the ideas how to solve this particular problem).

mathjazz:
> How would you do that? Just make this migration a no-op? That might be just as fine.

That's what I had in mind, so it's my favourite option. :)

jotes:
I accidentally closed this PR and can't reopen it. I'll submit the new one.

jotes:
@mathjazz can you look again?

PR #1120, opened 2018-11-02
Fix bug 1497850: Find source (en-US) files through project configuration

mathjazz:
~~**NOTE:** this PR is based on #1111, so the only relevant changeset is https://github.com/mozilla/pontoon/commit/ad2648ed02ea7481c2bf1746941a5c16817708e2.~~

Pontoon finds source (AKA reference) files in two steps:
1. First, it finds the **source directory**, which must be called `templates`, `en-US`, `en-us`, `en` [or some other variations](https://github.com/mozilla/pontoon/blob/ad2648ed02ea7481c2bf1746941a5c16817708e2/pontoon/sync/vcs/models.py#L58), and contain at least one file with supported file extensions ([learn more](https://github.com/mozilla/pontoon/blob/ad2648ed02ea7481c2bf1746941a5c16817708e2/pontoon/sync/utils.py#L58)).
2. Then, it searches for all files with supported file extensions within the source directory. These files are treated as source files.

[Project configuration](https://moz-l10n-config.readthedocs.io/en/latest/fileformat.html) brings more flexibility and reliability in detecting source files by explicitly defining their paths or patterns. This PR adds support for that, while keeping the algorithm described above avaialble for projects that don't use the project configuration file.

Finding source files through project configuration works like this:
1. Source files could be stored in multiple directories, so we just use the source repository checkout directory as the **source directory**. [CLICK](https://github.com/mozilla/pontoon/commit/ad2648ed02ea7481c2bf1746941a5c16817708e2#diff-c3cfdee64b8608d7c2751b1df6d78acdR388)
2. Then, we search for all the paths within that directory that match one of the reference patters in the project configuration file. [CLICK](https://github.com/mozilla/pontoon/commit/ad2648ed02ea7481c2bf1746941a5c16817708e2#diff-c3cfdee64b8608d7c2751b1df6d78acdR434)


mathjazz:
Asking @adngdb for the actual code review and @Pike for the super-review. The only change in this patch (except for the tests) is https://github.com/mozilla/pontoon/commit/ad2648ed02ea7481c2bf1746941a5c16817708e2#diff-c3cfdee64b8608d7c2751b1df6d78acd.

pontoon/sync/vcs/models.py
```diff
@@ -414,20 +420,38 @@ def source_directory_path(self):
...
+        List of absolute paths for all supported source resources
+        as specified through project configuration.
         """
         path = self.source_directory_path
-        for absolute_path in self.resources_for_path(path):
-            absolute_path = source_to_locale_path(absolute_path)
+        project_files = ProjectFiles(None, [self.configuration.parsed_configuration])
 
-            yield os.path.relpath(absolute_path, path)
+        for root, dirnames, filenames in scandir.walk(path):
```
Pike: I'd exclude hidden files here, too. Editors can leave files with weird names around.

PR #1121, opened 2018-11-05
Fix bug 1469010 - Process all strings above 255 characters length in

jotes:
TODO:
* [x] - find real-world examples of strings to test this patch
* [x] - research the performance of python part can be improved
* [x] - add unit tests
* [x] - add docstrings


pontoon/base/models.py
```diff
@@ -2872,30 +2874,106 @@ def serialize(self):
...
+                        then=Value(quality * 100)
+                    )
+                )
+
+        entries = (
+            self.filter(
+                pk__in=matches_pks,
+
+            ).annotate(
+                quallity= Case(
```
mathjazz: That space after `=` is breaking `pylama`.

pontoon/base/migrations/0136_index_on_translationmemoryentry.py
```diff
@@ -0,0 +1,19 @@
...
+    dependencies = [
+        ('base', '0135_project_project_configuration'),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            "CREATE INDEX tm_locale_char_length_idx ON base_translationmemoryentry(locale_id, char_length(source))",
+            "DROP INDEX tm_locale_char_length_idx"
+        )
+    ]
```
mathjazz: Can we create this index via Django? It's not very discoverable when created with a migration.

jotes:
@adngdb feel free to wait for tests, docstrings, and CI. I'll ping you when everything will be ready.

jotes:
@adngdb hey, can you take a look? thanks!

pontoon/base/models.py
```diff
@@ -2872,28 +2874,139 @@ def serialize(self):
...
-        entries = self.extra(
-            where=[
-                '(CHAR_LENGTH(source) BETWEEN %s AND %s)',
-                levenshtein_ratio_equation + ' > %s'
-            ],
-            params=(min_dist, max_dist, text, text, text, min_quality),
-            select={'quality': levenshtein_ratio_equation + '* 100'},
-            select_params=(text, text, text)
-        )
+        if min_dist > 255 or max_dist > 255:
```
adngdb: Can there be cases where `length` is > 255 but neither `min_dist` nor `max_dist` are? Because if that's possible then the code is going to raise a `ValueError`.

pontoon/base/models.py
```diff
@@ -2872,28 +2874,139 @@ def serialize(self):
...
+        else:
+            entries = self.postgres_levenshtein_ratio(
+                locale,
+                text,
+                min_quality,
+                min_dist,
+                max_dist,
+            )
+
         return entries
```
adngdb: This could be simplified to something like this, for less repetition:

```python
get_matches = self.postgres_levenshtein_ratio
if min_dist > 255 or max_dist > 255:
    get_matches = self.python_levenshtein_ratio

return get_matches(
     locale,
     text,
     min_quality,
     min_dist,
     max_dist,
)
```

pontoon/base/tests/managers/test_translationmemory.py
```diff
@@ -0,0 +1,162 @@
+import pytest
+
+from mock import patch, call
+
+from pontoon.base.models import TranslationMemoryEntry
+from pontoon.base.tests import TranslationMemoryFactory
```
adngdb: Please use `pontoon.tests.factories` instead here.

pontoon/base/tests/managers/test_translationmemory.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+@pytest.fixture
+def tm_entry_long():
+    yield TranslationMemoryFactory.create(
```
adngdb: That's surprising, why do you use `yield` instead of a simple `return`?

pontoon/base/tests/managers/test_translationmemory.py
```diff
@@ -0,0 +1,162 @@
...
+            0.7,
+            35,
+            71,
+        ).values_list('pk', 'source', 'target', 'quality')
+    )
+
+    expected_results = [
+        (tm_entry_short.pk, tm_entry_short.source, tm_entry_short.target, 100),
+    ]
+    assert postgresql_queryset == python_queryset == expected_results
```
adngdb: That's going to output confusing messages in case of a failure. I think it would be better to break it into 2 lines.

pontoon/machinery/views.py
```diff
@@ -59,11 +59,9 @@ def translation_memory(request):
     locale = get_object_or_404(Locale, code=locale)
     entries = (
         TranslationMemoryEntry.objects
-        .minimum_levenshtein_ratio(text)
-        .filter(locale=locale)
+        .minimum_levenshtein_ratio(locale, text)
```
adngdb: If you were to move `filter(locale)` before the call to `minimum_levenshtein_ratio`, wouldn't that allow you to simply ignore the `locale` filter altogether in the levenshtein functions? Because you'd trust that whatever `self` has is already filtered for that locale.

pontoon/base/models.py
jotes: I can't find a good example, So I'll remove this check from `postgresql_levenshtein_ratio`. I've been probably too defensive :-)

pontoon/base/models.py
```diff
@@ -2872,28 +2874,139 @@ def serialize(self):
...
+            ).annotate(
+                quality=Case(
+                    *quality_sql_map,
+                    **dict(
+                        default=Value(0),
+                        output_field=IntegerField(),
+                    )
+                )
+            )
+        )
```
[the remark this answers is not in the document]
adngdb: I think it's fine to not optimize to the fullest such a little used function. Keep it this way, but maybe add a comment explaining what you said in your comment?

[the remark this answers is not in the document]
jotes: <3 Sure, I'll ping you after I'll finish (I want to be sure that's understandable for everyone).

pontoon/base/models.py
```diff
@@ -2872,28 +2874,139 @@ def serialize(self):
...
+        if len(text) > 255 or min_dist > 255 or max_dist > 255:
+            raise ValueError("Can't search for strings longer than 255 characters.")
+
+        entries = (
+            self.extra(
+                where=[
+                    '(CHAR_LENGTH(source) BETWEEN %s AND %s)',
+                    self.get_levenshtein_sql('source') + ' > %s'
+                ],
+                params=(min_dist, max_dist, len(text), text, len(text), min_quality),
```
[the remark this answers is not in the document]
jotes: I'm not sure if I can pass named parameters here, because Django doesn't support them or I can't find it :)
https://github.com/django/django/blob/master/django/db/models/sql/query.py#L1877
I have an idea to create queryset filters to remove this `extra` calls, I'll try to do this tomorrow.

jotes:
@adngdb Hey,
I've introduced a couple of changes:
* removed the raw SQL query from the TranslationMemory
* rewrote the main equation into SQL expression
* squashed some of commits

Could you look at this again?

julen:
Thanks for working on this — looking forward to have an improved TM experience! 🎉

PR #1122, opened 2018-11-05
Fix bug 1443551: List custom plural forms

mathjazz:
If source string contains custom (numeric) plural forms, list them as variants in untranslated, along with locale CLDR plurals.

Test case:
https://github.com/mozilla-l10n/pontoon-ftl/blob/master/en-US/demo.ftl#L30

### DIFF
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -85,10 +85,7 @@ var Pontoon = (function (my) {
    */
   function getPluralForms(element) {
     var customPluralForms = element.expression.variants.filter(function(item) {
-      if (item.key.type === 'NumberLiteral') {
-        return true;
-      }
-      return false;
+      return item.key.type === 'NumberLiteral';
     }).map(function (item) {
       return item.key.value;
     });
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -84,11 +84,16 @@ var Pontoon = (function (my) {
    *   - CLDR plurals of the locale
    */
   function getPluralForms(element) {
-    var customPluralForms = element.expression.variants.map(function (variant) {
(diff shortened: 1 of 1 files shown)


pontoon/base/static/js/fluent_interface.js
```diff
@@ -77,6 +77,20 @@ var Pontoon = (function (my) {
...
+   * The list if a union of:
+   *   - any custom (numeric) plural forms specified in the original string (bug 1443551)
+   *   - CLDR plurals of the locale
+   */
+  function getPluralForms(element) {
+    var customPluralForms = element.expression.variants.map(function (variant) {
+      return variant.key.value;  // can be undefined, so we filter it out below
+    });
+
+    return customPluralForms.filter(Boolean).concat(Pontoon.locale.cldr_plurals);
```
Pike: This filters out `0`, too?

Also, would this add `one` to locales that only have `[many]` as plural forms?

Pike: You probably want to exclude the en-US plurals, and you don't want to filter out `0`.

mathjazz: Thanks! It doesn't include en-US plurals (that would be `item.key.name`), but it definitely filters out 0. I'll fix that and make the code more explicit.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -77,6 +77,25 @@ var Pontoon = (function (my) {
...
+   * The list if a union of:
+   *   - any custom (numeric) plural forms specified in the original string (bug 1443551)
+   *   - CLDR plurals of the locale
+   */
+  function getPluralForms(element) {
+    var customPluralForms = element.expression.variants.filter(function(item) {
+      if (item.key.type === 'NumberLiteral') {
+        return true;
+      }
+      return false;
```
Pike: Just `return item.key.type === 'NumberLiteral';` ?

PR #1126, opened 2018-11-09
Fix bug 1505825 - Load locales list via AJAX.

adngdb:
Passing the locales list through the template cannot work with django < 2, so we need to pass them differently. This patch makes a call to a new django endpoint to get the preferred locale for the current user, and then loads translation files when the list is received.

### DIFF
--- a/frontend/public/index.html
+++ b/frontend/public/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="{{ lang }}">
+<html>
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
--- a/frontend/src/core/api/l10n.js
+++ b/frontend/src/core/api/l10n.js
@@ -4,9 +4,28 @@ import APIBase from './base';
 
 
 export default class L10nAPI extends APIBase {
+    /**
+     * Return a string with translations for a given locale.
+     */
     async get(locale: string) {
         const url = this.getFullURL(`/static/locale/${locale}/translate.ftl`);
         const response = await fetch(url);
         return await response.text();
+    }
+
+    /**
+     * Return the preferred locale of the current user.
+     */
+    async getPreferredLocales(): Promise<Array<string>> {
+        const headers = new Headers();
+        headers.append('X-Requested-With', 'XMLHttpRequest');
+
+        const result = await this.fetch('/user-preferred-locale/', 'GET', null, headers);
+
+        const locales = [ result.locale ];
+        if (result.locale !== 'en-US') {
+            locales.push('en-US');
+        }
+        return locales;
     }
 }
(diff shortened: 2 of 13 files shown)


adngdb:
This is deployed on stage right now, and it works! Yay! :-)

adngdb:
@mathjazz, well that's simply because I didn't think about that function. However, lang neg is bound to evolve and to return not just one locale. Also, I think @jotes is right and we should move the logic of adding `en-US` to the list of locales to the server. All that means that we can't really use `<html lang="">` because the locale is most of the time going to be a list. Thus, I believe the API endpoint approach is better in the long run.

mathjazz:
OK, thanks for the clarification.

My only concern is that we shouldn't be making too many AJAX requests on load, since we already need 5 for the History, Machinery, and Locales tabs. But those calls might be made afterwards and thus not be affected. So my question is, can we load this data as part of the [logged-in user data](https://bugzilla.mozilla.org/show_bug.cgi?id=1490349)?

Also, I'm not sure we need to be as sophisticated as to give users the ability to define _a list_ of their preferred UI languages. I have never seen this is any web app, not even Facebook, Twitter or Gmail go beyond a single language.

julen:
> All that means that we can't really use `<html lang="">` because the locale is most of the time going to be a list.

The UI can only be in a single locale at a time, and based on that, `lang` (and `dir`) attributes should be set at the root. I agree with @mathjazz having a list as preferred UI languages seems unnecessarily sophisticated.

The server should already know the preferred user language, and I cannot understand why that couldn't be used when initializing the editor without the need for extra requests.

Pike:
> The UI can only be in a single locale at a time

That's a paradigm we're actually getting rid of with Fluent. With Fluent, a message knows which language it is in, and does choose things like plural forms or directionality accordingly.

adngdb:
There are a few misunderstanding here, so let me try to make things a bit clearer. @stasm or @Pike might want to chime in and complete what I'm saying.

Fluent can handle a list of locales to fallback to, in an ordered manner. That means that for some languages, it might be interesting to not fallback directly to `en-US` when a message is not localized. Let's say you are in Argentina and your locale of choice is `es-AR`. If the translation is incomplete for that locale, it would probably be a lot better for the user to fallback to `es`, or `es-ES`, before showing `en-US`. Those fallback chains should be generated automatically.

That means we won't give users the option to define a list of locales. They would get to choose their primary one, and then we would automatically compute the best possible fallbacks. I hope such a thing already exists out there. 

So, for an end user, they can: 
 - do nothing, get the UI shown in the locale that their browser is configured to show (basic language negociation), and benefit from smart fallbacks ; or
 - go to their settings page, force a specific locale, and benefit from smart fallbacks.

Does that sound better? 

In all cases, that means a page would potentially show messages in various locales, and thus the `lang` attribute becomes less consistent and useful. However, we should indeed update it when we get the preferred user's locale.

mathjazz:
You got me at "That means we won't give users the option to define a list of locales." :-)

adngdb:
Thanks everyone for the feedback! I'm closing this in favor of #1131.

PR #1130, opened 2018-11-14
Fix bug 1497851: Enable finding l10n files via project config

mathjazz:
TODO:

- [x] Make downloading files work.
- [x] Make uploading files work.
- [x] Look into adding new locales for projects with project config.
- [x] Add tests.


mathjazz:
@adngdb We talked about this is the past and since we're close to finalizing this milestone of project config, I'd like to make sure the code is in a satisfactory state. We have a couple of places ([1](https://github.com/mozilla/pontoon/blob/fc3339c93c0990ec9e92c9d4653708168911ce66/pontoon/sync/core.py#L172) & [2](https://github.com/mozilla/pontoon/blob/fc3339c93c0990ec9e92c9d4653708168911ce66/pontoon/sync/vcs/models.py#L428)) where code branches depending on whether project configuration is provided. I wonder what's the best way to name these methods. Some options:

- We could append `_with_pc`, `_with_config`, `_with_project_config` to the project config methods and replace `with` with `without` or `no` in the default methods.
- We could only postfix project config methods, e.g. with `_pc`, `_config` or `_project_config`.
- We could move the project config methods to the [VCSConfiguration](https://github.com/mozilla/pontoon/blob/fc3339c93c0990ec9e92c9d4653708168911ce66/pontoon/sync/vcs/models.py#L478) class.

mathjazz:
@Pike This patch should be feature-complete. I have to add some more tests. I know you're traveling, so I wonder if you'd be able to review it.

The biggest change I made since the initial version is the ability to add new locales to projects, and update the config file accordingly. I use `pytoml` to serialize it, which works, but [doesn't look optimal](https://github.com/mozilla-l10n/mobile-android/commit/a12d75215159e6713b8973499410f318a3d029f3).

I wonder if I should try to be smarter about it and only use this approach as a fallback. By default I'd first check if the file lists locale codes "one per line". If it does, I'd manually insert a line with the new locale code and try to maintain the alphabetic order. If it doesn't I'd fallback to the current behaviour. To avoid blocking, this could also be a followup bug.

mathjazz:
Updated - we no longer commit the file.

Thanks for the link to https://github.com/avakar/pytoml/issues/45 - the best way around this is of course to patch `pytoml` upstream and make it act like `compare-locales` serializer. :)

PR #1131, opened 2018-11-16
Fix bug 1505825 - Rework language negotiation.

adngdb:
This fixes passing the preferred locale of the user to the template for production. It also changes the language negotiation logic: now if the user has defined a preferred locale, they will be shown the UI in that locale, otherwise we use simple langneg in the client.

### DIFF
--- a/frontend/src/core/l10n/components/AppLocalizationProvider.js
+++ b/frontend/src/core/l10n/components/AppLocalizationProvider.js
@@ -3,6 +3,7 @@
 import * as React from 'react';
 import { connect } from 'react-redux';
 import { LocalizationProvider } from 'fluent-react';
+import { negotiateLanguages } from 'fluent-langneg';
 import 'intl-pluralrules';
 
 import * as l10n from 'core/l10n';
@@ -29,9 +30,21 @@ type InternalProps = {|
  */
 export class AppLocalizationProviderBase extends React.Component<InternalProps> {
     componentDidMount() {
-        // $FLOW_IGNORE: we count on the 'lang' attribute being set.
-        const locale = document.documentElement.lang;
-        this.props.dispatch(l10n.actions.get([locale]));
+        if (document.documentElement && document.documentElement.lang) {
+            // The user has chosen a specific locale, show that.
+            this.props.dispatch(
+                l10n.actions.get([ document.documentElement.lang ])
+            );
+        }
+        else {
+            // No specified locale, let's figure out the user's preferences.
+            const languages = negotiateLanguages(
+                navigator.languages,
+                ['fr', 'es', 'en-US'],  // available locales
+                { defaultLocale: 'en-US' },
+            );
+            this.props.dispatch(l10n.actions.get(languages));
+        }
     }
 
     render() {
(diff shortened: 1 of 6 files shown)


frontend/src/core/l10n/components/AppLocalizationProvider.js
```diff
@@ -29,9 +30,21 @@ type InternalProps = {|
  */
 export class AppLocalizationProviderBase extends React.Component<InternalProps> {
     componentDidMount() {
-        // $FLOW_IGNORE: we count on the 'lang' attribute being set.
-        const locale = document.documentElement.lang;
-        this.props.dispatch(l10n.actions.get([locale]));
+        if (document.documentElement && document.documentElement.lang) {
+            // The user has chosen a specific locale, show that.
+            this.props.dispatch(
+                l10n.actions.get([ document.documentElement.lang ])
```
stasm: I think this should have `en-US` as the last item, in case the selected locale is not complete. The UI should fall back to English then. You could use `negotiateLanguages` here as well, actually, or just push it if `documentElement.lang !== "en-US"`.

frontend/src/core/l10n/components/AppLocalizationProvider.js
```diff
@@ -29,9 +30,21 @@ type InternalProps = {|
...
+            );
+        }
+        else {
+            // No specified locale, let's figure out the user's preferences.
+            const languages = negotiateLanguages(
+                navigator.languages,
+                ['fr', 'es', 'en-US'],  // available locales
```
stasm: Is there an established way to put this in some kind of a config file? I'd recommend to also document that new languages need to be added here before they become available in the UI.

adngdb: I've filed https://bugzilla.mozilla.org/show_bug.cgi?id=1507811 to take care of this issue. I didn't want to spend more time on this right now. 

Note: your comments on how to solve this are very welcome in the bug! :-)

mathjazz: Good point! Note that this will be taken care of (at least partly) in https://bugzilla.mozilla.org/show_bug.cgi?id=1507811.

frontend/src/core/l10n/components/AppLocalizationProvider.js
adngdb: That is taken care of in the `get` action, see here: https://github.com/mozilla/pontoon/blob/8d71348cd55f436920e19965e53faa0c08d7af3c/frontend/src/core/l10n/actions.js#L50-L56

frontend/src/core/l10n/components/AppLocalizationProvider.js
mathjazz: Good timing!

frontend/src/core/l10n/actions.js
```diff
@@ -55,7 +58,13 @@ export function get(locales: Array<string>): Function {
             locales.push('en-US');
```
stasm: Remove this if here?

adngdb: Is `negotiateLanguage` taking care of it? If not, why do you now say we shouldn't enforce `en-US` to always be the last fallback language?

stasm: It does, yes! `{ defaultLocale: 'en-US' }` will make sure that `en-US` is on the list of negotiated languages, if needed.

PR #1137, opened 2018-11-29
Fix bug 1490351 - Add actions on status icons in History tab.

adngdb:
@mathjazz I did not want to use `update_translation` to approve a translation, because that view requires a lot of parameters that are not needed in order to only approve the translation. Also, I didn't like that it was incoherent with the other review actions (they all have a dedicated view), so I decided to add a new `approve_translation` view that does only that.

It might be a good idea to refactor `update_translation` so that it only adds a suggestion, and then calls `approve_translation`, but I didn't dare do that because `update_translation` is quite complex. We should probably do that someday.


pontoon/base/views.py
```diff
@@ -348,6 +348,93 @@ def get_translation_history(request):
...
+    if project.slug == 'tutorial':
+        # Checks are disabled for the tutorial.
+        use_ttk_checks = False
+    else:
+        try:
+            use_ttk_checks = UserProfile.objects.get(user=user).quality_checks
+        except UserProfile.DoesNotExist:
+            use_ttk_checks = True
+
+    if use_ttk_checks and translation.errors.all():
```
jotes: Hey :)
1. From what I remember (@mathjazz and @pike can correct me if something has changed in that matter) Pontoon can't allow the users to approve translations with errors.
https://github.com/mozilla/pontoon/blob/master/pontoon/checks/utils.py#L94

So, this line should look like this:
```python
if translation.errors.exist():
```
2. Translations can have warnings and the translator can choose to ignore them during the approval process.

pontoon/base/views.py
```diff
@@ -348,6 +348,93 @@ def get_translation_history(request):
...
+    translation.active = True
+    translation.approved = True
+    translation.approved_user = user
+    translation.approved_date = now
+    translation.unapproved_user = None
+    translation.unapproved_date = None
+    translation.fuzzy = False
+    translation.rejected = False
+    translation.rejected_user = None
+    translation.rejected_date = None
```
mathjazz: Gosh is this ugly (and we have a dozen of similar places). We should prioritize https://bugzilla.mozilla.org/show_bug.cgi?id=1473761 and https://bugzilla.mozilla.org/show_bug.cgi?id=1179954.

pontoon/base/views.py
mathjazz: Agreed. We should use `are_blocking_checks` here.

adngdb:
I am very afraid of modifying the current translate page, because it is less tested, and I don't really want to fall into a pit where I break something and then I have to spend too much time debugging it. I'd rather have this duplicated code for a bit, and once we get rid of the current Translate, we can remove the useless bits in `update_translation`.

pontoon/base/views.py
```diff
@@ -612,8 +606,14 @@ def perform_checks(request):
...
 def update_translation(request):
-    """Update entity translation for the specified locale and user."""
+    """Update entity translation for the specified locale and user.
 
+    Note that this view is also used to approve a translation by the old
+    Translate app. Once we migrate to Translate.Next, we'll want to rework
+    this view to remove the bits about approving a translation, because that
+    has been delegated to the `approve_translation` view.
+
+    """
```
mathjazz: Nice!

pontoon/base/views.py
```diff
@@ -382,16 +382,10 @@ def approve_translation(request):
...
-    else:
-        try:
-            use_ttk_checks = UserProfile.objects.get(user=user).quality_checks
-        except UserProfile.DoesNotExist:
-            use_ttk_checks = True
+    # Checks are disabled for the tutorial.
+    use_checks = project.slug != 'tutorial'
 
-    if use_ttk_checks and translation.errors.all():
+    if use_checks and translation.errors.exists():
```
mathjazz: That's an improvement, but we should actually re-run checks here:

```python
failed_checks = run_checks(...)
if are_blocking_checks(failed_checks, ignore_warnings):
    return JsonResponse({
        'failedChecks': failed_checks,
    })
```

The reviewer should not only be told that translation contains errors, but also be able to see any warnings and bypass them. It should also see Translate Toolkit warnings.

adngdb: There's one thing I don't understand here. Aren't the errors and warnings saved in the database? Why do we need to re-run the checks, instead of just getting them from the database and returning that? 

I assume the reason why you need to return them is because in some cases a user will approve something that is not the active translation.

mathjazz: > There's one thing I don't understand here. Aren't the errors and warnings saved in the database? Why do we need to re-run the checks, instead of just getting them from the database and returning that?

Good question. This is an edge case, but if we introduce new checks between the point when the translation was stored (along with failed checks) and the point when it gets approved, the new checks would never run. Unless we use the `run_checks` management command everytime we introduce important new checks (that's actually why we created it).

@Pike do you have thoughts on whether we should re-run checks when approving, or should we just return the ones in the DB?

> I assume the reason why you need to return them is because in some cases a user will approve something that is not the active translation.

We need to `return` them, because we need to list them in the UI. And let the user ignore them, if they are only warnings.

Pike: We didn't have new checks in a while, outside of native Android.

Arguably, we should rerun checks when we update compare-locales, though, as stated above, that's excessive these days.

Does this bug cover un-obsoleting translations? If so, we probably want to rerun tests?

Not a strong opinion.

mathjazz: Thanks for the update! And sorry for the ping, forgot you were on a PTO.

If by un-obsoleting translations you mean un-rejecting, then no. This patch only adds a new method for approving translations (currently the same method is used for saving and approving translations).

mathjazz: @adngdb If Axel doesn't have a strong opinion, I'm fine with moving on here by simply returning the existing failed checks and not re-running them.

adngdb:
@mathjazz hopefully this is good now. Note that I don't use errors or warnings in T.N yet, so whatever I'm returning doesn't matter much right now.

pontoon/base/views.py
```diff
@@ -348,6 +348,87 @@ def get_translation_history(request):
...
+        )
+
+    # Check for errors.
+    # Checks are disabled for the tutorial.
+    use_checks = project.slug != 'tutorial'
+
+    if use_checks and translation.errors.exists():
+        return JsonResponse({
+            'failedChecks': translation.errors.all(),
+        })
```
mathjazz: We should return all failed checks here, including warnings. And they should be formatted in the same fashion as `run_checks` formats them. 

You can also take care of this in followup bug that will implement checks, in which case you might want to error here, so we don't forget about this use case (and only focus on saving translations by mistake). Seems like we don't have a bug on file for implementing checks in Translate.Next, yet?

pontoon/base/views.py
```diff
@@ -348,6 +348,87 @@ def get_translation_history(request):
...
+    translation.approved_user = user
+    translation.approved_date = now
+    translation.unapproved_user = None
+    translation.unapproved_date = None
+    translation.fuzzy = False
+    translation.rejected = False
+    translation.rejected_user = None
+    translation.rejected_date = None
+
+    translation.save()
```
mathjazz: There are a few things here, inconsistent with the related "review" views (unapprove, reject, unreject, delete):

1. `translation.entity.reset_active_translation()` should take care of the `active` flag instead of mangling with it manually.

2. You should replace lines 393 - 424 with a call to `translation.approve()` model instance method, which you'll need to create. The `# Reject previously approved translations.` block is unnecessary - `Translation.save()` takes care of that.

3. And with those two changes in place, I suspect you should be able to lose `user = request.user` (given that we don't use it in other related methods).

mathjazz:
Sorry, pressed Submit too early. These two problems still occur:
> * After successfully approving a string, the editor should jump to the next string in the string list.
> * In Chrome, a blue box shadow appears around approve/reject icons after you click on them.

Additionaly:
* Rejected strings should have `opacity: 0.4`.
* Now that we added a delete action, we should implement that as well.

pontoon/base/views.py
adngdb: OK, filed https://bugzilla.mozilla.org/show_bug.cgi?id=1518820 then, and I'll remove this to just raise an error.

adngdb:
@mathjazz ready for another look? I should have addressed all your comments.

PR #1139, opened 2018-12-03
Fixes bug 1486497 - Add a Locales tab

adngdb:


pontoon/base/views.py
```diff
@@ -290,12 +290,21 @@ def get_translations_from_other_locales(request):
     try:
         entity = request.GET['entity']
         locale = request.GET['locale']
+        plural_form = request.GET['plural_form']
```
mathjazz: BTW, that is this change about?

Note that it breaks Translate.Current in its current form.

adngdb: This view doesn't take plural form into account. So if you're on a string that has a plural, you can only ever see the singular form from other locales. This change allows to see plural forms from other locales. I've also added a commit to fix my breaking the current Translate page.

adngdb:
> We don't seem to take into account Preferred locales you can set in settings.

Filed [bug 1514722](https://bugzilla.mozilla.org/show_bug.cgi?id=1514722).

adngdb:
@mathjazz I believe all your concerns have been addressed. :-)

frontend/src/modules/otherlocales/components/Locales.css
```diff
@@ -0,0 +1,38 @@
+.otherlocales > ul {
```
mathjazz: Nit: `other-locales`. We use `entity-details` for example.

adngdb:
@mathjazz I haven't been able to figure out the CSS jump problem, if you have some ideas, I'd be glad for your help. Otherwise, I think we can delay this and fix it later. I expect there will come a phase where we'll be fixing all the small issues that made it through the review net anyway.

frontend/src/modules/otherlocales/components/Locales.js
```diff
@@ -0,0 +1,67 @@
...
+    parameters: Navigation,
+|};
+
+
+/**
+ * Shows all translations of an entity in locales other than the current one.
+ */
+export default class Locales extends React.Component<Props> {
+    renderNoResults() {
+        return <section className="otherlocales">
```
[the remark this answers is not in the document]
adngdb: Hah, that's interesting, your merge actually broke that. Not sure how that happened.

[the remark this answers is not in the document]
adngdb: Stupid me cannot read a diff.

PR #1142, opened 2018-12-06
Fix bug 1435179: Add ability to request language from project dashboard

VishalCR7:
The UI remains exactly the same as /$TEAM page and with same user permission (i.e. just being authenticated). The js code remains exactly the same as /$TEAM page as well.

### DIFF
--- a/pontoon/teams/static/js/request.js
+++ b/pontoon/teams/static/js/request.js
@@ -14,11 +14,11 @@ var Pontoon = (function (my) {
             .toggleClass('fa-chevron-right', show)
             .toggleClass('fa-chevron-left', !show);
 
-        if (type === 'projects') {
+        if (type === 'locale-projects') {
           var localeProjects = $('#server').data('locale-projects');
 
           // Hide all projects
-          $('.projects')
+          $('.items')
             .toggleClass('request', !show)
             .find('tbody tr')
               .toggleClass('limited', !show)
@@ -26,14 +26,14 @@ var Pontoon = (function (my) {
 
           // Show requested projects
           $(localeProjects).each(function() {
-            $('.projects')
+            $('.items')
                 .find('td[data-slug="' + this + '"]')
               .parent()
                 .toggleClass('limited', show)
                 .toggle(show);
           });
 
-          Pontoon.requestItem.toggleButton(!show, 'projects');
+          Pontoon.requestItem.toggleButton(!show, 'locale-projects');
         }
         else if (type === 'team') {
           // Hide all teams and the search bar
@@ -52,8 +52,8 @@ var Pontoon = (function (my) {
         condition = condition || true;
         var show = condition
 
-        if (type === 'projects') {
-          show = condition && $('.projects td.check.enabled:visible').length > 0;
+        if (type === 'locale-projects') {
+          show = condition && $('.items td.check.enabled:visible').length > 0;
         }
         else if (type === 'team') {
           show = condition &&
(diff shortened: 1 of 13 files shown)


pontoon/teams/views.py
```diff
@@ -173,15 +173,14 @@ def ajax_permissions(request, locale):
 
 @login_required(redirect_field_name='', login_url='/403')
 @require_POST
-def request_item(request, locale=None):
+def request_item(request, type=None, code=None):
```
mathjazz: `type` is [not a good](https://stackoverflow.com/questions/2417979/can-i-use-the-variable-name-type-as-function-argument-in-python) variable name in Python.

pontoon/teams/templates/teams/widgets/team_list.html
```diff
@@ -68,5 +69,11 @@ <h4>
       </div>
       <button type="submit" id="request-item">Request new team</button>
     </form>
+    {% else %}
+    <div id="request-item-note">
+      <p>The requested projects won’t be enabled automatically: an email will be sent to Pontoon admins and team managers to evaluate the request. We’ll try to respond as soon as possible. Please make sure to get in touch with the localization team if you are a new contributor.</p>
```
mathjazz: Sorry... The requested **teams**

pontoon/teams/urls.py
```diff
@@ -4,13 +4,18 @@
 import views
 
 urlpatterns = [
+    # AJAX: Request projects to be added to locale
+    url(r'^request/(?P<_type>[A-Za-z0-9\-\@\.]+)/(?P<code>[A-Za-z0-9\-\@\.]+)/$',
```
mathjazz: Let's use `request_type` to be more explicit.

pontoon/teams/views.py
```diff
@@ -205,6 +211,43 @@ def request_item(request, locale=None):
...
+        locales = ''.join('- {} ({})\n'.format(l.name, l.code) for l in locale_list)
+
+        mail_subject = u'Team request for {project} ({slug})'.format(
+            project=project.name, slug=project.slug
+        )
+
+        cc = []
+
+        for locale in locale_list:
+            cc += (
```
jotes: Hey!
I know that this PR will have to be reviewed by PMs, but I hope that you won't mind if I'll add some notes/comments - All of them all optional. If you feel that you haven't learned anything new from them or you just disagree - that's also okay (and feel free to say it out loud :-).

So, currently this loop will make n number of queries. I think that can be changed into:
```python
# distinct() is required, because some of managers can be assigned to multiple locales.
cc = locale_list.values_list('managers_group__user__email', flat=True).distinct()
```

pontoon/teams/views.py
```diff
@@ -205,6 +211,43 @@ def request_item(request, locale=None):
             'user_url': request.build_absolute_uri(user.profile_url)
         }
 
+    # Request teams to be enabled for project
+    elif code and request_type == 'teams':
+        code_list = request.POST.getlist('items[]')
+        project = get_object_or_404(Project, slug=code)
```
jotes: [nit] I would use `Project.objects.available` to filter out disabled projects.

mathjazz:
So we got feedback from the PMs, who requested two changed:

**1. Change wording to use Language instead of Team:**
* The toggle button should read "Request new Language" and "Back to Enabled Languages".
* The button on the bottom of the request form should read "Request new Language".
* The notice above that button should read "The requested languages won’t be ...".

**2. Only allow requesting one language at a time:**
* We should use [radio buttons]( https://fontawesome.com/cheatsheet) instead of checkboxes:
   * default: `fa fa-circle`
   * selected: `fa fa-dot-circle`
   * hovered: `far fa-dot-circle`
* Which means we can also use the same code path on the backend (URL, view, email template ...) as for requests on the team page, because request from the project page will be the same as a request on the team page with only one project selected.

mathjazz:
Nice work! I would change the styling of radio buttons a little bit, here's the patch you can apply:
https://gist.github.com/mathjazz/34765a24191da703e549d8e53c391a94

One thing on top of that (not included in the diff) is that we shouldn't change the selected radio button on hover (it should stay the same as in the selected state).

Fix that and change the wording and we should be good to go.

VishalCR7:
There's still a change of wording left which I m a bit confused about the locations where the change is needed i.e. only on /$project page or /$teams page or overall pontoon (to maintain uniformity as suggested by jeff)

mathjazz:
When you hover over a selected radio button, it turns into a solid circle. Instead we should just keep the selected radio button unchanged on hover. Please change that.

And, it seems like the checkboxes on the team page (when requesting projects) are now broken. Probably caused by my gist, sorry about that. Could you fix that, please?

Good point about wording - I'll consult with the team and get back to you about it.

pontoon/teams/static/js/request.js
```diff
@@ -74,22 +74,23 @@ var Pontoon = (function (my) {
             projects: projects,
           },
           success: function() {
-            Pontoon.endLoader('New projects request sent.', '', 5000);
+            Pontoon.endLoader('New ' + type + ' request sent.', '', 5000);
```
mathjazz: This will still produce "team" instead of "language". We can just say: "New request sent".

PR #1144, opened 2018-12-10
Add portable docker image

danehammer:
The Heroku approach at present requires environment variables to be landed in a file by the buildpacks. That means the dev docker approach today does the same. But that prevents being able to publish a Pontoon image to DockerHub that folks can run outside of Heroku. I wrote a new Dockerfile that is a tad smaller and portable. It uses the same environment variables, but gets them from docker at runtime so it can be shared.

I have this running on our Mesos cluster at Udacity. I am not using several of the add-ons though, so there is likely more work to do. I wanted to have a proposed solution for https://bugzilla.mozilla.org/show_bug.cgi?id=1513104 first though.

To review and test:
* build locally with `docker build -f docker/Dockerfile.prod -t mozilla/pontoon:latest .`
* replace the docker-compose config with that image
* deploy somewhere else? If someone could try using Heroku's new support for deploying Docker containers, that would be most excellent!

### DIFF
--- a/docker/Dockerfile.prod
+++ b/docker/Dockerfile.prod
@@ -1,0 +1,66 @@
+FROM node:10.14.1-alpine
+
+# Install build deps
+RUN set -ex \
+  && apk add --no-cache \
+  gcc \
+  make \
+  libc-dev \
+  musl-dev \
+  linux-headers \
+  pcre-dev \
+  postgresql-dev \
+  libxml2-dev \
+  libxslt \
+  libxslt-dev \
+  python2 \
+  py2-pip \
+  python2-dev \
+  git \
+  openssh-client \
+  build-base \
+  && pip install --upgrade pip \
+  && pip install virtualenv
+
+# Create the folder for frontend assets
+RUN mkdir -p /app/assets
+
+COPY requirements.txt /app
+
+# Install python deps
+RUN virtualenv /env \
+  && /env/bin/pip install -U 'pip>=8' -r /app/requirements.txt
+
+COPY . /app
+
+RUN cd /app && npm install
+RUN cd /app/frontend && yarn install
+
+# JavaScript applications paths
+ENV WEBPACK_BINARY /app/node_modules/.bin/webpack
+ENV YUGLIFY_BINARY /app/node_modules/.bin/yuglify
+
+# Run webpack to compile JS files
+RUN cd /app/ && $WEBPACK_BINARY
+
+# Build Translate.Next frontend resources
+RUN cd /app/frontend/ && yarn build
+
+# pontoon service port
+EXPOSE 8000
+
+# Python environment variables
+ENV PYTHONUNBUFFERED 1
+ENV PYTHONDONTWRITEBYTECODE 1
+ENV PYTHONPATH /app
+
+COPY docker/run_gunicorn.sh /app
+RUN chmod +x /app/run_gunicorn.sh
+
+WORKDIR /app
+
+# Call collectstatic (customize the following line with the minimal environment variables needed for manage.py to run):
+RUN DATABASE_URL=none SECRET_KEY=none SITE_URL=none /env/bin/python manage.py collectstatic --noinput
+
+# Start the django app
+CMD ["/app/run_gunicorn.sh"]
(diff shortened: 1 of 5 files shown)


jotes:
Hi @danehammer,
It's astonishing to hear that Udacity uses Pontoon and gives some good energy back ;-)

@mathjazz and @adngdb will take the full review, but after a quick look this looks very similar to things I'm using at my dayjob (gunicorn and stuff).

Maybe We'll consider to switch deployment setup and use containers on Heroku?

docker/Dockerfile.prod
```diff
@@ -0,0 +1,66 @@
+FROM node:10.14.1-alpine
```
Pike: I'm concerned about the change to the base image here.

mathjazz:
Thanks for the patch, @danehammer! 

I'll leave this PR in the good hands of @Pike, @adngdb and @jotes. Please note that we might be a little slow to respond until the middle of the next week due to a community event we're hosting.

danehammer:
Thanks for the quick responses. I was waffling between (a) trying to pull request something small enough to be useful immediately when it hits master, and (b) starting a conversation about supporting this more directly instead of dropping it in on the side. Looks like (b) happened on its own. So if it works for this project's style, I'll add a bunch of comments for what I would like it to look like long term. Let me know when that should be separate PRs or if it should be all merged together.

adngdb:
Hey @danehammer, thanks for your contribution! I by no means have the skills to fully understand what your proposing, but I do not doubt that it is better than what I built (for I basically copy/pasted what others from Mozilla did). However, I have a few questions. 

Theoretically, `docker` and `docker-compose` have different purposes. The images built using our Dockerfiles should be production-ready. Then `docker-compose` makes changes on top of that, like setting volumes, changing default configuration, etc. to adapt those images to make development-ready containers. However, as we do not deploy our images to prod, I have no idea if they actually work or not.

Considering this, I wonder why you did not, or could not, deploy our existing image to production? I also wonder why you went for a separate file. I would assume that the same approach (a production-ready Dockerfile and docker-compose overwrites for dev) should work with the changes you made. Having two different files is worrisome because, as we don't deploy to prod using docker, we would never actually test the prod file. If we kept the current model, we would at least always test the docker files locally, during development. 

I hope my ramblings make sense. Thanks again for your work!

docker/Dockerfile.prod
```diff
@@ -0,0 +1,66 @@
...
+ENV PYTHONDONTWRITEBYTECODE 1
+ENV PYTHONPATH /app
+
+COPY docker/run_gunicorn.sh /app
+RUN chmod +x /app/run_gunicorn.sh
+
+WORKDIR /app
+
+# Call collectstatic (customize the following line with the minimal environment variables needed for manage.py to run):
+RUN DATABASE_URL=none SECRET_KEY=none SITE_URL=none /env/bin/python manage.py collectstatic --noinput
```
adngdb: I'm not sure I understand this comment and those `=none`. Are users expected to edit this file on their local clone before building? Or is this a note that more work is needed in the PR?

danehammer: Sorry, will correct. I was going from a guide I found online for how to make a production ready docker image for a django app and this is copy/paste cruft.

danehammer:
> Considering this, I wonder why you did not, or could not, deploy our existing image to production? I also wonder why you went for a separate file. I would assume that the same approach (a production-ready Dockerfile and docker-compose overwrites for dev) should work with the changes you made. Having two different files is worrisome because, as we don't deploy to prod using docker, we would never actually test the prod file. If we kept the current model, we would at least always test the docker files locally, during development.

The existing image seemed to, at first glance, not yield everything in a production-ready way. The specifics being:
* The node process is backgrounded in the `run_webapp.sh` script - what happens if that process stops? The container should die
* The heroku assumptions of how to get runtime env vars. I don't want the secrets persisted into an image, I want them provided as env vars to the docker runtime when I start the image, otherwise I risk exposing secrets to by docker image registry and anyone with access to the image on that registry

It sounds like everyone is amenable to making improvements like this. My approach early on was to just get something I felt good about running on Udacity's Mesos cluster, and then share with this group. So it sounds like I need to re-work this approach to maximize the re-use of Dockerfile logic, for producing the image you want to run in production, the image you want to run for local development (with docker-compose), and possibly even what could be deployed on Heroku. I'll take a stab at that this week.

mathjazz:
Hi @danehammer, just checking if you managed to make any progress here. Thanks! :)

mathjazz:
Closing due to inactivity. Feel free to reopen if you continue working on this. Thanks!

erickwilder:
@mathjazz @danehammer 
I would like to deploy Pontoon on a Kubernetes cluster and an official Docker image would be much more beneficial to the whole community. 
I'm willing to dedicate some time to help on this. Just let me know from where I can pick up to not duplicate efforts.

danehammer:
My team abandoned their pontoon efforts, so I haven't worked on it. Feel free to take this and run with it.

PR #1147, opened 2018-12-14
Update compare-locales to pick up fixes for Android.

Pike:
Notably, we're picking up the checker file pattern, and bug 1512709, to
keep namespace prefixes on Android documents.

https://github.com/Pike/compare-locales/compare/RELEASE_5_0_0...RELEASE_5_0_3


Pike:
> Fluent update is a separate release, right?

Yes, preparing that right now. Version is going to be 5.1.0, fyi.

PR #1148, opened 2018-12-18
Bug 1486504 - Navigation for Translate.Next

adngdb:
This adds a very simple navigation element at the top. It's a breadcrumb that allows to go back to the home page, to the team page, or to the project dashboard for the current team. Note that this PR doesn't close the bug, as I guess this is insufficient for our users.

### DIFF
--- a/frontend/src/App.css
+++ b/frontend/src/App.css
@@ -1,20 +1,25 @@
 #app {
-    display: flex;
-    height: 100%;
+    display: grid;
+    grid-template-columns: 50% 50%;
+    grid-template-rows: 60px auto;
+    height: 100vh;
+}
+
+#app header {
+    grid-column-start: 1;
+    grid-column-end: 3;
 }
 
 #app .panel-content,
 #app .panel-list {
-    width: 50%;
-    height: 100%;
+    height: calc(100vh - 60px);
 }
 
 #app .panel-list {
-    position: absolute;
+    position: relative;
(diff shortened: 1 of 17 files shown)


frontend/src/App.css
```diff
@@ -1,20 +1,25 @@
 #app {
-    display: flex;
-    height: 100%;
+    display: grid;
```
mathjazz: We dropped the idea of using `CCS grid` a while ago, because it's not supported by our browser support matrix.

adngdb: Can we maybe align our support matrix on the grid feature? I don't remember how we chose those browsers to support, so maybe that's a dumb question. :-)

mathjazz: OK, I'll have a look and PR the branch with an update to the CSS file. Or the browser matrix. :-)

adngdb:
@mathjazz, I'd like to merge this now, because there are some refactoring in this that will impact my other work on T.N. Do you mind if we just merge now and make the needed changes in follow-up PRs?

mathjazz:
I updated the styling but left CSS Grid in. So I also updated the Browser Support Matrix.

CSS is quite different from the one we used in Translate.Current, so the changeset would be quite extensive. Besides, we'll probably use more CSS Grid in the future.

adngdb:
Hurray for Grid!

PR #1149, opened 2018-12-18
No bug: Fix counting unreviewed pluralized strings in calculate_stats

mathjazz:
Each Unreviewed suggestion should be counted separately in stats. Until this patch, all Unreviewed suggestions of a pluralized entity for a given locale were counted as 1.

After landing this patch, we also need to `./manage.py calculate_stats` for all .po projects.

@jotes r? Details on IRC.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -3339,7 +3339,7 @@ class TranslatedResource(AggregatedStats):
                 rejected=False
             ).count()
             if plural_unreviewed_count:
-                unreviewed += 1
+                unreviewed += plural_unreviewed_count
 
         if not save:
             self.total_strings = resource.total_strings


pontoon/base/models.py
```diff
@@ -3339,7 +3339,7 @@ def calculate_stats(self, save=True):
                 rejected=False
             ).count()
             if plural_unreviewed_count:
-                unreviewed += 1
+                unreviewed += plural_unreviewed_count
 
         if not save:
```
Kasbr305, line 3344: 1

pontoon/base/models.py
```diff
@@ -3339,7 +3339,7 @@ def calculate_stats(self, save=True):
                 rejected=False
             ).count()
             if plural_unreviewed_count:
-                unreviewed += 1
```
Kasbr305, line 3342: -

PR #1150, opened 2018-12-19
Bug 1514691: Change wording of the empty translation warning

mathjazz:
I've deployed this to stage as the first step to fixing bug 1514691.

Could you please have a look, @flodolo?


mathjazz:
I've deployed this to stage as the first step to fixing bug 1514691.

Could you please have a look, @flodolo?

mathjazz:
Thanks!

I agree. Commit messages starting with "Fix bug X" close the bug, messages with "Bug X" just reference it. So this one will stay open.

I'd like to check what % of the authors of the empty translations you've fixed have TTK checked enabled or not. Where can I find commits you've made? That would help in identifying those users.

PR #1151, opened 2018-12-19
Bug 1489054 - Add the loader for the initialization phase

jotes:
Hey,
It's the next loader from the list :) Could you review this and share your thoughts?

Thanks in advance!
cc @mathjazz, @adngdb


frontend/src/App.css
```diff
@@ -6,3 +6,124 @@
...
+    75% {
+        display: block;
+        opacity: 0.75;
+    }
+
+    100% {
+        display: block;
+        opacity: 1;
+    }
+}
```
mathjazz: Nit.

adngdb:
I missed that in your previous PR, but I think loaders should all live in a core module instead of next to where they are used. That is because some of those we could reuse in other places. 

I would go for a simple `core/loaders` module, with one component per loader type (`Spinner`, `Wave`... ).

jotes:
@adngdb 
I totally agree. Actually, I've moved the Loader to `core` module yesterday. I can move the previous module.
I refactored the PR a little. Could you look at the changes in `AppLocalizationProvider` and express your opinion?
I'll fix eslint errors and add tests as the last step after your initial review.

@mathjazz 
Could you look again? I fixed all things you mentioned in the previous comment.

mathjazz:
Thanks for addressing the issues, @jotes! This is the only one that remains:

> When you search or filter entities, the new loader appears (instead of `fa-sync`).

jotes:
@adngdb 
re: https://github.com/mozilla/pontoon/pull/1151#issuecomment-449373296
In order to solve this problem: I want to add a reducer for Preloader component. This reducer will recalculate `isLoading` state variable before it will be false for the first time.
This reducer should be called on the top of all reducers (that can be done by `reduce-reducers package).
Simpler ideas are always welcome!

`isLoading` is calculated here: 
https://github.com/mozilla/pontoon/pull/1151/files#diff-5af375ffa05d9305793dfbed74fcff0fR48.

jotes:
@adngdb hey, I've started to implement your previous comment but I have only one doubt:
What should happen when e.g. the users tries to search for a string (look at: https://github.com/mozilla/pontoon/pull/1151#issuecomment-449373296)?
If I understand your idea correctly, `entitieslist.fetching` will be set to true and this will render the Loader again (but shouldn't).

adngdb:
@jotes, you're right, take `entities` out of that. We show a different loader for entities anyway, right? Only show the "application loader" if `l10n` or `locales` are fetching, those only ever fetch once per page load.

jotes:
@adngdb Hey, no problem :-) However, to give you more context why I've started thinking about reducers:
In the current version of the Translate View, the wave loader is being displayed until the server returns requested entities.

adngdb:
Yeah, I think that's fine if we change this behavior. The 2 apps don't need to be exactly similar, especially when it's due to those structural changes.

jotes:
@mathjazz hey, I'm posting the visual reference just to let you know what should be expected from this patch:
* [When loader wait untill entities will load ](https://www.youtube.com/watch?v=gPnVcSGhMyw&feature=youtu.be)
* [When loader doesn't wait for entities](https://www.youtube.com/watch?v=a1HfseXLe5c&feature=youtu.be)
In both videos, I slowed down `/get-entities/` just to make the situation more visible. If that's okay then I'll rebase and apply all comments and push it again to your review.

mathjazz:
@jotes That looks good if I correctly understand what you mean by loader "waits" or "doesn't wait" for entities.

> When loader wait untill entities will load

This should only happen when the translate page is loaded. We don't hide the main loader until entities are fully loaded, otherwise we'd show two different loaders one after another (the main one and the `fa-sync` one in the sidebar).

> When loader doesn't wait for entities

This should happen when entities in the string list change because of searching and filtering. Instead of the main loader, we only show the `fa-sync` loader in the sidebar.

jotes:
@adngdb @mathjazz

mathjazz:
Sure, let's talk about this on Monday.

jotes:
@adngdb Hey!
After our discussion I've decided to reduce number of changes and keep them as small as possible.
The most important changes:
*  I moved all loading-related components to `src/core/loaders`
*  I removed Preloader component . It could be confusing to use in the future because of mixed responsibilities (One component to load them all ;-)
Could you review again?

adngdb:
>     3. The fact that we show two loaders one after another is not optimal. I think we'll need to change this. Still, I think we can add that as a followup, when we'll be able to use the new translate view with all the filters and locale-project-file combinations. Then it'll be easier to tell.

Friendly reminder, we decided during our Jan 14 Pontoon meeting that we would do things this way for now, and see later on if that is problematic or not. :)

adngdb:
> 1. Unlike in Translate.Current, it seems like we show the white background briefly, right after the main loader shows up. Maybe we just need to change the default template background to the same color that's used in the loader background?

That's already the case: https://github.com/mozilla/pontoon/blob/master/frontend/public/index.html#L28

The only white background I see is when waiting for the first load, right before the loader shows up. I don't see white after that.

frontend/src/core/loaders/components/WaveLoader.css
```diff
@@ -0,0 +1,101 @@
...
+    0% {
+        display: None;
+        opacity: 0;
+    }
+
+    100% {
+        display: block;
+        opacity: 1;
+    }
+}
```
adngdb: You should configure your code editor to always add new lines at the end of files. And add a new line here. ;)

frontend/src/core/loaders/components/WaveLoader.test.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import WaveLoader, { WaveLoader as WaveLoaderBase } from './WaveLoader';
+
+import { createReduxStore } from '../../../test/store';
+import { shallowUntilTarget } from '../../../test/utils';
```
adngdb: You can use absolute imports here, e.g. `import from 'test/store';` instead.

frontend/src/core/loaders/components/WaveLoader.test.js
```diff
@@ -0,0 +1,67 @@
...
+
+describe('<WaveLoader>', () => {
+    it('render the loader before l10n bundle will be fetched', () => {
```
adngdb: Nit, these verbs should be 3rd person (`renders`), the goal being to read them as a sentence (`it renders the loader`).

frontend/src/modules/entitieslist/components/EntitiesList.js
```diff
@@ -11,7 +11,7 @@ import * as navigation from 'core/navigation';
 
 import { actions, NAME } from '..';
 import Entity from './Entity';
-import EntitiesLoader from './EntitiesLoader'
+import { CircleLoader } from '../../../core/loaders'
```
adngdb: Here too you can use absolute import.

mathjazz:
> Friendly reminder, we decided during our Jan 14 Pontoon meeting that we would do things this way for now, and see later on if that is problematic or not. :)

That’s not what we decided, although it’s in the docs. :) The decision was to test first and then see if we need to block on this.

jotes:
@adngdb I fixed some stuff, could you look again?

frontend/src/core/loaders/components/WaveLoader.js
```diff
@@ -1,43 +1,45 @@
...
+    const props = this.props;
     // Render children components, because some of fetches happen after components are rendered.
-    <React.Fragment>
-        { props.isLoading ? <WaveLoadingBar /> : '' }
-        { props.children }
-    </React.Fragment>
-);
+        return props.isLoading ? <WaveLoadingBar /> : props.children;
+    }
+}
```
adngdb: This is great, but I would go even further, remove this component, and put the logic in the `App` component. I don't think there's much use in having a component for just this `if`.

jotes:
@mathjazz 
ad 1. I see this issue too. I'm not an expert but from a very quick look at the network tab I see the following things:
* all CSS is inside javascript files
* Before the first render, the browser has to fetch `bundle.js`, `1.chunk.js`, `main.chunk.js`
* For unknown reasons `1.chunk.js` loads very slowly on my machine.
Maybe it's related to Webpack and automatic code-splitting (however I don't see the easy way to fix this bug).

mathjazz:
@jotes I don't know the technical details, so I'm passing this to @adngdb.

frontend/src/core/loaders/components/WaveLoader.js
jotes: done :-)

jotes:
@mathjazz 
Hey,
ad2. In the last commit I removed the dark header from the loading bar component.
You can look at this here:
https://youtu.be/I--zqUFWkFU (Unfortunately, the loading bar is being displayed only for a few milliseconds)
How does it look?

mathjazz:
>  How does it look?

Just tested it locally and it looks great!

frontend/src/App.css
```diff
@@ -23,3 +23,13 @@
 	border-left: 1px solid #5E6475;
 	box-sizing: border-box;
 }
+
+.inner {
+    display: table-cell;
+    vertical-align: middle;
+}
```
adngdb: This should be removed.

frontend/src/App.js
```diff
@@ -4,15 +4,21 @@ import React from 'react';
 import { connect } from 'react-redux';
 
 import './App.css';
-
+import { L10nState } from 'core/l10n';
+import { LocaleState } from 'core/locales';
```
adngdb: These are Flow types and thus should be imported like so: 

```js
import type { L10nState } from 'core/l10n';
```

Also, I try to keep some structure to the imports, even though I haven't documented it yet. Here's what it is:

```js
import 'libraries';

import 'css';

import 'core and modules';

import 'local files';

import type 'types';
```

frontend/src/App.js
```diff
@@ -4,15 +4,21 @@ import React from 'react';
...
 import { Lightbox } from 'core/lightbox';
 import * as locales from 'core/locales';
 import { Navigation } from 'core/navigation';
 import { UserAutoUpdater } from 'core/user';
 import { EntitiesList } from 'modules/entitieslist';
 import { EntityDetails } from 'modules/entitydetails';
 import { SearchBox } from 'modules/search';
+import { WaveLoader } from './core/loaders';
```
adngdb: This should be an absolute import, without `./` thus, and should be placed by alphabetical order of the "from". So, right after `import { Lightbox } from 'core/lightbox';` in this case.

frontend/src/App.js
```diff
@@ -45,4 +57,13 @@ class App extends React.Component<InternalProps> {
...
 
-export default connect()(App);
+const mapStateToProps = (state: Object): Props => {
+    const { l10n, locales } = state;
+
+    return {
+        l10n,
+        locales,
+    };
+};
```
adngdb: I think it's better if we don't hard-code the module names, and instead use their declared `NAME` constant. That will allow us to refactor easier in the future. Take a look at the way I did it in other files, for example here: https://github.com/mozilla/pontoon/blob/master/frontend/src/modules/search/components/SearchBox.js#L104

Note that `router` is an exception because it's an external library and I don't expect it to ever change.

frontend/src/core/loaders/components/WaveLoader.js
```diff
@@ -0,0 +1,27 @@
+/* @flow */
+
+import React from 'react';
+
+import './WaveLoader.css';
+
+export const WaveLoader = (): React.Node => (
+    <div id="project-load" className="overlay">
```
adngdb: In order to make modules as abstract as possible, I would do 2 things here: first remove the `id` and make it a `class` instead, so that it can be reused more easily if we need to, and rename it to not be `project-load`, because that is too specific to the use case, and not to the nature of this component. I'd go with a simple `wave-loader` class name.

frontend/src/core/loaders/components/WaveLoader.css
```diff
@@ -0,0 +1,100 @@
+
+.overlay {
```
adngdb: Please make sure that everything in CSS files is encapsulated in the component's identified (for now an id, but with my proposed change, this would be in `.wave-loader`), so that it doesn't cause unexpected side effects in other components.

frontend/src/core/loaders/components/WaveLoader.js
```diff
@@ -0,0 +1,27 @@
...
+                &nbsp;
+                <div></div>
+                &nbsp;
+                <div></div>
+                &nbsp;
+                <div></div>
+                &nbsp;
+                <div></div>
+            </div>
+            <div className="text">&quot;640K ought to be enough for anybody.&quot;</div>
```
adngdb: @mathjazz do you think we should make that string localizable? :thinking:

mathjazz: I think it's not adding much value, so I'd just remove it for now.

adngdb: Actually, that's displayed before l10n strings are in, so there's no point localizing it.

adngdb:
> @adngdb How can we fix the white background? [#1151 (comment)](https://github.com/mozilla/pontoon/pull/1151#issuecomment-458766411)

Add `!important` in `index.html` after the background color. This flicker happens because in `boilerplate.css` there's a `background: transparent` applied to `body`. That stays until the CSS of the components loads in the JS bundle.

adngdb:
Here's a great, great news: this patch fixes a race issue we're having at the moment on Translate.Next! \o/

mathjazz:
> Here's a great, great news: this patch fixes a race issue we're having at the moment on Translate.Next! \o/

That's a GIF material.

jotes:
@adngdb 
I fixed things you mentioned in previous comments. Could you review them?

PR #1152, opened 2018-12-21
Fix bug 1397377: Restore ability to delete translations

mathjazz:
Translation can only be deleted, if it's rejected. That way we make the
Delete action of secondary importance compared to the first-rate review
actions - Approve & Reject. Hence, if you want to delete a non-rejected
translation, you'll need to make two clicks: first to reject and then to
delete a translation - the second click acting like a sort of confirmation step.


mathjazz:
@jotes r?

pontoon/base/views.py
```diff
@@ -482,6 +482,41 @@ def unreject_translation(request):
...
+    locale = translation.locale
+
+    # Read-only translations cannot be deleted
+    if utils.readonly_exists(project, locale):
+        return HttpResponseForbidden(
+            "Forbidden: This string is in read-only mode"
+        )
+
+    # Only privileged users or authors can delete translations
+    if not translation.rejected or not (
```
jotes: [nit] `get_object_or_404` can take a QuerySet as the first parameter. That's totally subjective but it could be more easier to read/refactor it later.
e.g.
```
translation = get_object_or_404(Translation.objects.rejected, pk=4)
if not(request.user.can_translate(locale,project) or request.user == translation.user):
    ...
```

jotes:
@mathjazz I would add a test... but besides that lgtm.

pontoon/base/views.py
mathjazz: We'd need a manager for `Translation.objects.rejected` to work, though - otherwise this is a an `AttributeError`.

PR #1155, opened 2018-12-28
Fix bug 1516165: Hide email address from profile page for unauthenticated users

mathjazz:
This is available on stage for testing:
https://mozilla-pontoon-staging.herokuapp.com/contributors/24SF1A7jby9lhuuirruQS4xRN3w/

### DIFF
--- a/pontoon/contributors/templates/contributors/profile.html
+++ b/pontoon/contributors/templates/contributors/profile.html
@@ -27,10 +27,19 @@
   <h2 id="username">{{ contributor.first_name }}</h2>
 
   <ul class="info">
-    <li><span class="fa fa-fw fa-envelope"></span><a href="mailto:{{ contributor.email|nospam }}">{{ contributor.email|nospam }}</a></li>
-    <li><span class="fa fa-fw fa-lock"></span>{{ contributor|display_permissions() }}</li>
+    <li>
+      <span class="fa fa-fw fa-envelope"></span>
+      <a href="mailto:{{ contributor.email|nospam }}">{{ contributor.email|nospam }}</a>
+    </li>
+    <li>
+      <span class="fa fa-fw fa-lock"></span>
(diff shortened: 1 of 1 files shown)


mathjazz:
@flodolo Functional review?

pontoon/contributors/templates/contributors/profile.html
```diff
@@ -27,10 +27,28 @@
...
+      <span class="fa fa-fw fa-envelope"></span>
+      {% if user.is_authenticated() %}
+      <a href="mailto:{{ contributor.email|nospam }}">{{ contributor.email|nospam }}</a>
+      {% else %}
+      {% if settings.DJANGO_LOGIN %}
+        {% set login_url = url('standalone_login') %}
+      {% else %}
+        {% set login_url = provider_login_url(request) %}
+      {% endif %}
+      <a id="sidebar-signin" href="{{ login_url }}">Sign in to see email address</a>
```
flodolo: I'm torn if we should just hide the email address, instead of telling that we're going to show it.

Who's supposed to see my email address? Should it be just users with higher privileges (translators/managers/admins), or just anyone logged in? If that's the case, I should probably have a setting in the profile to hide my email address from standard users. But then, the email address of admins (or PM contacts) should be always visible.

I think this is a good first step, but I'd like to have that discussion in the future. We'll also need to touch the profile for the review process, I believe, so this keeps things simple and solves the immediate issue.

mathjazz: For comparison: GitHub also shows "Sign in to view email", but the user can chose to hide the email address from the profile page.

mathjazz:
I'll merge this, because it fixes the bug. We can continue improving this in a followup bug or even as part of the [general profile page overhaul](https://bugzilla.mozilla.org/show_bug.cgi?id=1225109).

PR #1158, opened 2019-01-08
Remove unused priority field and add date created and date obsoleted to resource and project

VishalCR7:


pontoon/base/models.py
```diff
@@ -1269,6 +1272,10 @@ def save(self, *args, **kwargs):
                 original = Project.objects.get(pk=self.pk)
                 if self.disabled != original.disabled:
                     disabled_changed = True
+                    if self.disabled:
+                        self.date_disabled = timezone.now()
+                    else:
+                        self.date_disabled = None
```
mathjazz: We should replace this entire custom `save()` method with a [`signal`](https://docs.djangoproject.com/en/2.1/ref/signals/). That would be consistent with how we update stats when projects get deleted. Checkout the `signals.py` file.

VishalCR7: I was planning to go with signals initially but I found this save function "accidentally" and thought that it  was the right thing to do. No problem I will remove it and convert it into `pre_save` and `post_save` signals.

PR #1160, opened 2019-01-09
Fix bug 1518749 - Update bleach to version 3.0.2.

adngdb:
Note that from version 3, bleach does not require html5lib anymore, as it vendors it. We thus remove it from our requirements as well.


mathjazz:
Can we also remove the `html5lib` requirements?

adngdb:
Well, yes, of course, why not?

PR #1162, opened 2019-01-09
Fix bug 1433994 - Show an Ace editor for Fluent strings

adngdb:
@stasm, @mathjazz, how's that looking?


mathjazz:
Nice work! It works fine and I love how you structured the code. I also tested the changes proposed by @stasm and I like them.

I suggest you link the patch to [bug 1433994](https://bugzilla.mozilla.org/show_bug.cgi?id=1433994), which is unlikely to get fixed in Translate.Current.

I'm curious whether the theme and highlighting in `editor-mode-fluent.js` and `editor-theme-fluent.js` come from `projectfluent.org` or some other external source. Maybe we should reference that in the comment?

We should adapt the theme to make the editor feel more at home in Pontoon, but I suggest we do that as part of the post Translate.Next changes to the translate view.

The only thing I would change at this point is add `border: none;` to `.editor > textarea`. It's been around for a while I suspect, but I only noticed the border now when switching between the editors.

adngdb:
> I suggest you link the patch to [bug 1433994](https://bugzilla.mozilla.org/show_bug.cgi?id=1433994), which is unlikely to get fixed in Translate.Current.

:+1: 

> I'm curious whether the theme and highlighting in `editor-mode-fluent.js` and `editor-theme-fluent.js` come from `projectfluent.org` or some other external source. Maybe we should reference that in the comment?

That's what @stasm is referring to in his previous comment:

> I'll also publish the Fluent editor mode as its own package on npm in the future and I'll open a PR here to update to it once it's done.

> We should adapt the theme to make the editor feel more at home in Pontoon, but I suggest we do that as part of the post Translate.Next changes to the translate view.

Sounds good to me!

> The only thing I would change at this point is add `border: none;` to `.editor > textarea`. It's been around for a while I suspect, but I only noticed the border now when switching between the editors.

Doing that.

PR #1163, opened 2019-01-11
[WIP]Fix bug 1468840: Obsolete Resources instead of deleting them.

VishalCR7:
Entities of obsoleted resources are also obsoleted.
Stat calculation seems to work fine.
The obsoleted resource doesn't seem to appear in any view.

PS: still looking, will be adding more commits in the future.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -889,6 +889,7 @@ class Locale(AggregatedStats):
     def aggregate_stats(self):
         TranslatedResource.objects.filter(
             resource__project__disabled=False,
+            resource__obsolete=False,
             resource__project__system_project=False,
             locale=self,
         ).aggregate_stats(self)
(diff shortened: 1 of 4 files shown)


pontoon/base/models.py
```diff
@@ -1876,6 +1878,10 @@ class ResourceQuerySet(models.QuerySet):
     def asymmetric(self):
         return self.filter(format__in=Resource.ASYMMETRIC_FORMATS)
 
+    def obsolete(self, now):
+        self.update(obsolete=True, date_obsoleted=now)
+        Entity.objects.filter(resource__in=self).update(obsolete=True, date_obsoleted=now)
```
mathjazz: It's probably a good idea to also add `entity.obsolete()` and `entities.obsolete()` methods, which would update these two fields.

VishalCR7:
> Nit: that's probably not needed.

Yeah, Probably that is causing the build to fail as well

mathjazz:
From a quick testing, the changes you've made seem to be working fine.

Next, we should go through all calls to `Resource` instances in the code (especially `project.resources`) and make sure we only return non-obsolete ones instead of all where needed, because until now the code relied on the fact that all `Resource` instances are needed. We should probably introduce `Resource.objects.available()` method to help with that, similar to the one we have on the `Locale` and `Project` class.

adngdb:

Pike:
Going to give this a round trip against current master, just before it's turning 1.

mathjazz:
Continuing in #2351.

PR #1164, opened 2019-01-13
Fix bug 1520050 - Remove custom managers from the User model because they gene…

jotes:
Hey,
It looks like after some small refactorings, I was able to fix bug 1342613.

Can you look if this is mergeable @mathjazz?
Thanks in advance


pontoon/base/models.py
```diff
@@ -90,152 +90,6 @@ def combine_entity_filters(entities, filter_choices, filters, *args):
...
-                'count': day['count'],
-                'project': example.entity.resource.project,
-                'translation': example,
-            })
-
-        return timeline
-
-
-class UserQuerySet(models.QuerySet):
-    def serialize(self):
```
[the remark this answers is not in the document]
jotes: I was surprised too.

jotes:
@mathjazz a good catch! Something is fishy because Travis is green (and tests for contributors shouldn't even pass).

jotes:
@mathjazz Okay, once again you proved that we don't have to rely on Travis because we have you ;-)
Tests related to 'pontoon.contributors' app weren't executed in pytest. I fixed it.

pontoon/base/utils.py
```diff
@@ -737,3 +739,5 @@ def readonly_exists(projects, locale):
         locale=locale,
         readonly=True,
     ).exists()
+
+
```
mathjazz: Nit: we don't need so many new lines at the end of the file.

pontoon/contributors/utils.py
```diff
@@ -0,0 +1,131 @@
...
+        contributor.translations_approved_count = user['approved']
+        contributor.translations_unapproved_count = user['unreviewed']
+        contributor.translations_needs_work_count = user['fuzzy']
+        contributor.user_role = contributor.role(managers, translators)
+
+    contributors_list = sorted(contributors, key=lambda x: -x.translations_count)
+    if limit:
+        contributors_list = contributors_list[:limit]
+
+    return contributors_list
```
mathjazz: Nit: No newline.

PR #1165, opened 2019-01-14
Fix bug 1492163 - Add support for RTL languages to Translate.Next.

adngdb:


adngdb:
@mathjazz  how do I prevent this from happening? It doesn't happen on the prod version and I can't spot the difference in HTML or CSS.

mathjazz:
> @mathjazz  how do I prevent this from happening? It doesn't happen on the prod version and I can't spot the difference in HTML or CSS.

https://github.com/mozilla/pontoon/commit/66bf4ef3f81a0bb5f7fbeb6d91b9c7dab2e17413

frontend/src/modules/entitieslist/components/Entity.css
```diff
@@ -33,10 +33,15 @@
...
-    font-style: italic;
     padding-left: 2%;
     text-align: start;
 }
+.entity .translation-string[data-script="Latin"],
+.entity .translation-string[data-script="Greek"],
+.entity .translation-string[data-script="Cyrillic"],
+.entity .translation-string[data-script="Vietnamese"] {
+    font-style: italic;
+}
```
mathjazz: While you're at it, you should also add fix for [bug 1353135](https://bugzilla.mozilla.org/show_bug.cgi?id=1353135):

```css
#entitylist > .wrapper > ul > li p span.translation-string[lang="mk"],
#entitylist > .wrapper > ul > li p span.translation-string[lang="sr"] {
  font-family: "Ubuntu Regular";
}
```

adngdb: Done!

PR #1167, opened 2019-01-15
Fix bug 1518828: [Translate.Next] Implement delete option in History tab

mathjazz:
This is a WIP patch which:
1. Shows the delete button when needed.
2. Deletes the translation on click.

This is the work that remains to be migrated:
https://github.com/mozilla/pontoon/blob/ba838fa88b787a9323f563987c35a4904fcab942/pontoon/base/static/js/translate.js#L2325-L2347

@adngdb I need some help:

1. How am I doing so far?
2. How do I reference the button to apply the `delete` class to it? Should I do that in `actions.js`?
3. How do I update count? Do I need to update `history.translations` first and then use length?

### DIFF
--- a/frontend/src/modules/history/actions.js
+++ b/frontend/src/modules/history/actions.js
@@ -117,10 +117,20 @@ export function updateStatus(
 }
 
 
+export function deleteTranslation(
+    translation: number,
+): Function {
+    return async dispatch => {
+        await api.translation.delete(translation);
+    }
+}
+
+
 export default {
     get,
     receive,
     request,
     reset,
     updateStatus,
+    deleteTranslation,
 };
--- a/frontend/src/modules/history/actions.js
+++ b/frontend/src/modules/history/actions.js
@@ -118,10 +118,14 @@ export function updateStatus(
 
 
 export function deleteTranslation(
+    entity: number,
+    locale: string,
+    pluralForm: number,
     translation: number,
 ): Function {
     return async dispatch => {
         await api.translation.delete(translation);
+        dispatch(get(entity, locale, pluralForm));
     }
 }
 
--- a/frontend/src/modules/history/components/Translation.css
+++ b/frontend/src/modules/history/components/Translation.css
@@ -116,7 +116,6 @@
 }
 
 .translation > header button.delete {
-    display: none;
     margin-right: 1px;
 }
 
@@ -129,8 +128,4 @@
 
 .translation > header button.delete:hover:before {
     color: #F36;
-}
-
-.translation.rejected.can-reject > header button.delete {
-    display: inline-block;
 }
--- a/frontend/src/modules/history/components/Translation.css
+++ b/frontend/src/modules/history/components/Translation.css
@@ -134,9 +134,3 @@
 .translation.rejected.can-reject > header button.delete {
     display: inline-block;
 }
-
-/* MANDATORY, or transitionend is triggered too late */
-.translation.deleted {
-    transform: scale(0);
-    -webkit-transform: scale(0);
-}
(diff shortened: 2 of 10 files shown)


frontend/src/modules/history/components/Translation.js
```diff
@@ -119,6 +124,21 @@ export default class Translation extends React.Component<Props> {
...
+                    // Delete Button
+                    <Localized
+                        id="history-translation-button-delete"
+                        attrs={{ title: true }}
+                    >
+                        <button
+                            className='delete far'
+                            title='Delete'
+                            onClick={ this.delete }
+                        ></button>
```
adngdb: When an element is empty, you can (and should) use an autoclosing tag (`<button ... />`).

mathjazz:
Thank you for your help @adngdb! My first Translate.Next patch is finally ready for review!

frontend/src/modules/history/components/Translation.js
```diff
@@ -121,6 +126,21 @@ export default class Translation extends React.Component<Props> {
...
+                        attrs={{ title: true }}
+                    >
+                        <button
+                            className='delete far'
+                            title='Delete'
+                            onClick={ this.delete }
+                        />
+                    </Localized>
+                    :
+                    null
```
adngdb: I like to reverse those conditions so that I can put the `? null :` on the first line, and make it overall a little bit more compact. It also puts the `:` operator closer to the condition, which I believe makes it easier to understand.

frontend/src/modules/history/components/Translation.css
```diff
@@ -113,3 +114,23 @@
 .translation > header button.unreject:hover:before {
     font-weight: 400;
 }
+
+.translation > header button.delete {
+    display: none;
```
adngdb: I'm not a big fan of hiding components with CSS. It's a form of logic imho, and I like to keep logic in the JS code. I'd rather you add a condition in the component to not show the button if the user can't reject, rather than having to go into CSS to understand when the button is shown or not.

frontend/src/modules/history/actions.js
```diff
@@ -117,10 +117,24 @@ export function updateStatus(
...
 
 
+export function deleteTranslation(
+    entity: number,
+    locale: string,
+    pluralForm: number,
+    translation: number,
+): Function {
+    return async dispatch => {
+        await api.translation.delete(translation);
```
adngdb: Cool thing about reviewing this code: it makes me realize we don't have any sort of tolerance to errors, and we should. For example, here if the server returns any kind of error, the client will throw an error as well, and that error never being caught will propagate and make the client stop running. It's completely out of scope here, but in the future we should add some try / catch blocks around all our API calls somehow, and show some errors to the user. I might do that as part of [bug 1500871](https://bugzilla.mozilla.org/show_bug.cgi?id=1500871).

frontend/src/modules/history/components/Translation.js
```diff
@@ -131,7 +151,7 @@ export default class Translation extends React.Component<Props> {
                             className='unapprove fa'
                             title='Unapprove'
                             onClick={ this.unapprove }
-                        ></button>
+                        />
```
adngdb: Oh, thanks! I don't know how that slipped by. :-)

frontend/src/modules/history/actions.js
mathjazz: Agreed, code-wise this is very much related to that bug. Adding a reminder comment there.

mathjazz:

Do you feel like reversing the delete button condition makes the code more or less readable? I'm not sure.

And yes, I've dropped the animation. I think it didn't add much (any) value to the UX, because it slows you down. And it doesn't looks so great if you delete a non-last translation in the History.

mathjazz:
Fixing the tests...

mathjazz:
I'm not sure what's the problem here, any idea?
https://travis-ci.org/mozilla/pontoon/builds/488421357?utm_source=github_status&utm_medium=notification

PR #1168, opened 2019-01-17
Fix heroku-deployment link

DamienCassou:


mathjazz:
Thanks for the patch! Could you please explain what's the purpose of this change?

DamienCassou:
> Thanks for the patch! Could you please explain what's the purpose of this change?

The link in the README is supposed to create a Heroku app but doesn't: it just moves the user to the heroku platform without any clue about what to do.

This PR replicates the link at https://mozilla-pontoon.readthedocs.io/en/latest/admin/deployment.html which does the job of creating a Heroku app for Pontoon.

DamienCassou:
Matjaž Horvat <notifications@github.com> writes:
> So if you click on the button at https://github.com/mozilla/pontoon/,
> it doesn't take you to the same page as the button in the docs?


no. The current button in the README (that this PR is trying to fix)
brings the user to a general-purpose Heroku page that has no relation to
Pontoon. The button in Pontoon's documentation brings the user to a page
where Pontoon is already selected as the app to deploy.


> The benefit of not specifying the URL is that you can then deploy a
> branch or fork by just clicking on the button.


I don't know how that would work. I guess someone working on a branch is
probably knowledgeable enough to manually deploy on Heroku.

-- 
Damien Cassou
http://damiencassou.seasidehosting.st

"Success is the ability to go from one failure to another without
losing enthusiasm." --Winston Churchill

DamienCassou:
I can't reproduce anymore.

DamienCassou:
Sorry for the noise

mathjazz:
No problem! :)

PR #1170, opened 2019-01-18
Fix bug 1515159: Visually indicate no. of TM matches

mathjazz:
Show Machinery count as "Number of TM matches" + "Number of other matches",
similarly as we show counts in Locales tab as "Number of preferred locales" + "Number of other locales".


mathjazz:
@MikkCZ Could you please review the behaviour on stage?
https://mozilla-pontoon-staging.herokuapp.com/cs/firefox/browser/chrome/browser/browser.properties/?string=74432

PR #1171, opened 2019-01-18
Fix bug 1514691: Ask for confirmation when submitting empty translations

aniruddha2000:
1. I have disabled the untranslated translate toolkit check https://github.com/mozilla/pontoon/blob/master/pontoon/checks/libraries/__init__.py#L41

2. Created our own check here https://github.com/mozilla/pontoon/blob/master/pontoon/checks/libraries/pontoon.py


aniruddha2000:
Made a mistake by adding previous changes of my PR

PR #1172, opened 2019-01-19
[DO NOT MERGE] Fix bug 1514691: Ask for confirmation when submitting empty translations

aniruddha2000:
1. I have disabled the untranslated translate toolkit check https://github.com/mozilla/pontoon/blob/master/pontoon/checks/libraries/__init__.py#L41

2. Created our own check here https://github.com/mozilla/pontoon/blob/master/pontoon/checks/libraries/pontoon.py

### DIFF
--- a/pontoon/checks/libraries/pontoon.py
+++ b/pontoon/checks/libraries/pontoon.py
@@ -70,12 +70,10 @@ def run_checks(entity, string):
             checks['pWarning'].append(
                 'Empty translation'
             )
-
         else:
             checks['pErrors'].append(
                 'Empty translations are not allowed'
             )
-
 
     # FTL checks
     if resource_ext == 'ftl' and string != '':
--- a/pontoon/checks/libraries/pontoon.py
+++ b/pontoon/checks/libraries/pontoon.py
@@ -65,7 +65,12 @@ def run_checks(entity, string):
                 )
 
     # Prevent empty translation submissions if not supported
-    if string == '' and not entity.resource.allows_empty_translations:
+    if entity.resource.allows_empty_translations:
+        checks['pWarning'].append(
+            'Empty translation'
+        )
+
+    else:
         checks['pErrors'].append(
             'Empty translations are not allowed'
         )
--- a/pontoon/checks/libraries/pontoon.py
+++ b/pontoon/checks/libraries/pontoon.py
@@ -65,15 +65,17 @@ def run_checks(entity, string):
                 )
 
     # Prevent empty translation submissions if not supported
-    if entity.resource.allows_empty_translations:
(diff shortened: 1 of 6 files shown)


pontoon/checks/libraries/pontoon.py
```diff
@@ -65,10 +65,17 @@ def run_checks(entity, string):
...
-    if string == '' and not entity.resource.allows_empty_translations:
-        checks['pErrors'].append(
-            'Empty translations are not allowed'
-        )
+    if string == "":
+        if entity.resource.allows_empty_translations:
+            checks['pWarning'].append(
+                'Empty translation'
+            )
+
```
mathjazz: Remove this blank line.

pontoon/checks/libraries/pontoon.py
```diff
@@ -65,10 +65,17 @@ def run_checks(entity, string):
...
+        else:
+            checks['pErrors'].append(
+                'Empty translations are not allowed'
+            )
+
```
mathjazz: Remove this blank line.

aniruddha2000:
> @aniruddha2000 Thanks for the fix, but 2 other tests are still failing (please make sure you [run tests](https://mozilla-pontoon.readthedocs.io/en/latest/dev/contributing.html#running-tests) locally, before openning a PR):
> 
>     1. In [pontoon/checks/tests/test_libraries.py::test_tt_disabled_checks](https://travis-ci.org/mozilla/pontoon/builds/482070018#L1807) you need to add `untranslated` to the list of disabled checks.
> 
>     2. In [pontoon/checks/tests/test_pontoon.py::test_empty_translations](https://travis-ci.org/mozilla/pontoon/builds/482070018#L1829) you'll need to add `pWarning` with the corresponding message.

Fixed all the tests :)

flodolo:
> @flodolo This patch will now always show a warning when you try to submit an empty translation to a format that supports empty translations (like `.properties`).

That would be good.

> A side effect of this change is that those warnings also get stored in the DB and show up on dashboards, because that's what happens with all errors and warnings that don't come from Translate Toolkit. If we just move forward with this patch, the number of warning will quadruple.

That really wouldn't work. We would have a ton of people doing weird things to those strings just to get rid of warnings, and that's not the outcome I'd like. I'd like people to think twice before submitting an empty string, because most of the times it happens unwillingly, not to care about empty strings that are expected.

mathjazz:
Thanks for the quick response.

Filed a bug:
https://bugzilla.mozilla.org/show_bug.cgi?id=1521601

Note: **DO NOT MERGE**

mathjazz:
This PR has been integrated into #1235 for easier development and deployment.

PR #1174, opened 2019-01-21
Fix bug 1521523: [FTL] Hardcode - in front of Term keys

mathjazz:
This is how the non-imported strings from prod look on stage:
https://mozilla-pontoon-staging.herokuapp.com/ru/firefox/browser/browser/branding/sync-brand.ftl/
https://mozilla-pontoon-staging.herokuapp.com/uk/firefox/browser/browser/branding/sync-brand.ftl/

### DIFF
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -104,7 +104,15 @@ var Pontoon = (function (my) {
     return elements.every(function(element) {
       return (
         isSimpleElement(element) ||
-        (element.type === 'Placeable' && element.expression.type === 'SelectExpression')
+        (
+          element.type === 'Placeable' &&
+          element.expression.type === 'SelectExpression' &&
+          element.expression.variants.every(function(variant) {
+            return variant.value.elements.every(function(element) {
+              return isSimpleElement(element);
+            });
+          })
+        )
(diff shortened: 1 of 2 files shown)


mathjazz:
I didn't add any tests, because... we don't have tests for the ftl format. @adngdb Did we perhaps lost those as part of the pytest rewrite? If not, we should add some.

pontoon/base/static/js/fluent_interface.js
```diff
@@ -104,7 +104,15 @@ var Pontoon = (function (my) {
...
-        (element.type === 'Placeable' && element.expression.type === 'SelectExpression')
+        (
+          element.type === 'Placeable' &&
+          element.expression.type === 'SelectExpression' &&
+          element.expression.variants.every(function(variant) {
+            return variant.value.elements.every(function(element) {
+              return isSimpleElement(element);
+            });
+          })
+        )
```
mathjazz: This changes makes sure that we don't treat nested selectors as supported in rich editor?

adngdb:
I am not aware of us losing tests. Are you positive we had tests at some point in the past?

mathjazz:
OK. It could be that we "lost" them as part of the L10n -> Fluent migration.

PR #1175, opened 2019-01-21
Fix several Cross-Site Scripting Vulnerabilities

tch1bo:
Hello,

I noticed 5 Cross-Site Scripting [(XSS)](https://www.owasp.org/index.php/Cross-site_Scripting_(XSS)) Vulnerabilities in the pontoon project. 

For each of the vulnerabilities please find a proposed fix in the corresponding commit. The commit messages include a Proof of Concept exploit to trigger the vulnerability. 

Fortunately, I couldn't bypass the CSP, so these vulnerabilities might not be easily exploitable. Some of them are additionally protected by CSRF-cookies, which makes exploiting them in a real-world scenario even harder. Nevertheless, each of them is still a major security risk, which someone more skilled than me might successfully exploit, so i would suggest fixing them. Please consider my patches, which should mitigate the vulnerabilities and not harm legit usages of the application.

Also, please note that i am not a regular contributor to your project, so i am not very familiar with it. I found the bug while testing DeepCode’s AI Code Review. The tool can help you automate the process of finding such (and many other types of) bugs. You can sign-up your repo (free for Open Source) to receive notifications whenever new bugs are detected. You can give it a try [here](https://deepcode.ai/app/gh/review?utm_source=github_python&utm_campaign=mozilla_pontoon_XSS).

Any feedback is more than welcome at chibo@deepcode.ai.

Cheers, Victor.

### DIFF
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -247,7 +247,7 @@ def entities(request):
     """Get entities for the specified project, locale and paths."""
     form = forms.GetEntitiesForm(request.POST)
     if not form.is_valid():
-        return HttpResponseBadRequest(form.errors.as_json())
+        return HttpResponseBadRequest(form.errors.as_json(escape_html=True))
 
     locale = get_object_or_404(Locale, code=form.cleaned_data['locale'])
 
--- a/pontoon/batch/views.py
+++ b/pontoon/batch/views.py
@@ -112,7 +112,7 @@ def batch_edit_translations(request):
     """
     form = forms.BatchActionsForm(request.POST)
     if not form.is_valid():
-        return HttpResponseBadRequest(form.errors.as_json())
+        return HttpResponseBadRequest(form.errors.as_json(escape_html=True))
 
     locale = get_object_or_404(Locale, code=form.cleaned_data['locale'])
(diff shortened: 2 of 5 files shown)


mathjazz:
Thanks for the patch, @tch1bo!

@adngdb Could you have a look, please?

adngdb:
Hi @tch1bo, thank you for your contribution! 

As far as I understand, however, these changes are useless. That is because in all but one case, we are treating data generated by django from our Form classes. That data should not contain any user input, only things that were written in code, and thus do not risk exposing us to an XSS attack. 

The last one is about escaping data that comes from an external service, through their API. I understand that it is possible they might get compromised and then in turn it might be possible that we get compromised too. But I'm not even positive that is possible, since that data is exposed in a HTTP Response that is, as far as I know in our case, not rendered in our DOM. Hence, there should be no risk there either.

@mathjazz, you know more about this code than I do. Can you confirm that we do not render those `HttpResponseBadRequest` anywhere? 

Hope that makes sense, if not, please do let me. :)

adngdb:
And now I'm seeing your commit messages and I feel stupid... :man_facepalming:

PR #1177, opened 2019-01-22
Fix bug 1519898 - Add user settings in the Editor. 

adngdb:
This adds settings to the Editor, and makes use of them to show the "Save" or "Suggest" button as well as when sending translations. It also now moves on to the next entity after a translation has been sent.


adngdb:
@stasm mind giving a quick look at https://github.com/mozilla/pontoon/pull/1177/commits/6d4908ba89a61f493add1f8c1355afd757c07367 and letting me know if I did something wrong?

stasm:
I haven't tried it locally, but it LGTM.

PR #1181, opened 2019-01-24
Fix bug 1497572: Improve custom deployment experience 

VishalCR7:
Added a new template instead of 404ing when no projects are present as discussed on IRC.
This is how it looks:
image
 
PS:  Design subject to change :P

### DIFF
--- a/pontoon/base/static/css/style.css
+++ b/pontoon/base/static/css/style.css
@@ -1058,6 +1058,17 @@ noscript p {
   line-height: 162px;
 }
 
+#error #box a {
+  color: #7BC876;
+  font-size: 108px;
+}
+
+#error #box p {
+  font-size: 32px;
+  line-height: 50px;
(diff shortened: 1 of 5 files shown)


pontoon/base/templates/add_project.html
```diff
@@ -0,0 +1,21 @@
...
+  <div class="inner">
+    <div id="box">
+      <a href={{ url('pontoon.admin.project.new') }}>
+        <div class="fas fa-plus-square"></div>
+      </a>
+      <p>There aren't any project set up currently</p>
+    </div>
+    <div id="subtitle">
+      <a href={{ url('pontoon.admin.project.new') }}>
+        {% block description %}Set up your first project{% endblock %}
```
mathjazz: You don't need the `description` block here, and to avoid adding whitespace, you should just put the entire `<a>` node in one line.

PR #1183, opened 2019-01-29
No bug - Enable running only changed frontend tests.

adngdb:
This requires the container to have git, and the local .git folder. With those, running tests with jest automatically only runs the tests that have changed since the last commit.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,5 +3,7 @@
 **/*.pyc
 __pycache__
 .cache
-venv
+assets
+node_modules
 static
+venv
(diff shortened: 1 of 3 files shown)


adngdb:
@mathjazz or @jotes, could one of you give this a quick try to make sure I'm not breaking the dev environment? 

### Steps to reproduce

1. `make build`
2. `make test-frontend`

### Current result

It runs the entire test suite.

### Expected result

It only runs tests that have changed since the last commit. To verify that, you can make a small change to any `.test.js` file and see that it only re-runs that single test.

PR #1184, opened 2019-02-01
Fix bug 1486496 - Add a Machinery tab to Translate.Next. 

adngdb:
This adds the Machinery tab and all its features. On load, fetches translations from various sources (Machine Translation, Translation Memory) and shows it as it arrives. Also allows searching for specific terms.

@mathjazz the commits in this branch are a bit erratic, so I'm not sure there's much value reviewing this one commit at a time. That is because I've been working weirdly on this, going back and forth as I understood more about the feature.

Note that the search bar does nothing for now, as this patch is already huge. I'll do that in follow-up [bug 1524549](https://bugzilla.mozilla.org/show_bug.cgi?id=1524549). The copy on click action will be covered in [bug 1490346](https://bugzilla.mozilla.org/show_bug.cgi?id=1490346).

### DIFF
--- a/frontend/public/static/locale/en-US/translate.ftl
+++ b/frontend/public/static/locale/en-US/translate.ftl
@@ -22,19 +22,37 @@ editor-settings-force-suggestions = <glyph></glyph>Make Suggestions
 editor-settings-change-all = Change All Settings
 
 
+## Machinery
+## Shows a list of translations from machines.
+
+machinery-machinery-search-placeholder =
+    .placeholder = Type to search machinery
+
+machinery-translation-copy =
+    .title = Copy Into Translation (Tab)
+
+machinery-translation-number-occurrences =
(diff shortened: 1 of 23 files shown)


mathjazz:
Works great!

The only difference with Prod is that here we make server requests even for services that are not enabled, e.g. Google Translate or Microsoft Translator. Translate.Current [prevents](https://github.com/mozilla/pontoon/blob/f3fcf08/pontoon/base/static/js/main.js#L534) that and it would be nice to do that here as well, to prevent extra traffic.

I haven't looked at the code yet.

adngdb:
I thought about that and my conclusion was that it doesn't matter much. It's adding a bit of complexity to the app to avoid a few low cost requests. Am I missing something? Do you disagree with this?

mathjazz:
SGTM.

PR #1185, opened 2019-02-01
Fix bug 1519898 - Avoid erroring when user is logged out.

adngdb:


adngdb:
You are very right that this is not satisfying, and that is because it will be solved in a future bug: [bug 1524602](https://bugzilla.mozilla.org/show_bug.cgi?id=1524602). In the meantime, I just wanted to prevent this bug in the simplest way. :)

PR #1191, opened 2019-02-06
Fix bug 1486508: Add entity navigation

mathjazz:
This is a WIP PR, I need some additional information before I can proceed.


frontend/public/static/locale/es/translate.ftl
```diff
@@ -10,6 +10,15 @@ entitydetails-editor-button-save = Guardar
 entitydetails-editor-button-suggest = Sugerir
 
 
+## Entity Navigation
+## Shows next/previous buttons.
+
+entitynavigation-next = <glyph></glyph>Siguiente
+    .title = Go To Next String (Alt + Down)
+entitynavigation-previous = <glyph></glyph>Anterior
+    .title = Go To Previous String (Alt + Up)
```
mathjazz: Hi @guerojeff, Any chance you could translate the `.title` attributes to Spanish? :)

frontend/public/static/locale/fr/translate.ftl
```diff
@@ -35,6 +35,15 @@ machinery-translation-number-occurrences =
     .title = Nombre d'occurrences de la traduction
 
 
+## Entity Navigation
+## Shows next/previous buttons.
+
+entitynavigation-next = <glyph></glyph>Suivant
+    .title = Go To Next String (Alt + Down)
+entitynavigation-previous = <glyph></glyph>Précédent
+    .title = Go To Previous String (Alt + Up)
```
mathjazz: Hi @adngdb, Any chance you could translate the `.title` attributes to French? :)

frontend/src/modules/entitydetails/components/EntityNavigation.js
```diff
@@ -0,0 +1,66 @@
...
+
+
+/**
+ * Component showing entity navigation toolbar.
+ *
+ * Shows next/previous buttons.
+ */
+export default class EntityNavigation extends React.Component<Props> {
+    goToNextEntity = (entity: DbEntity) => {
+        console.log("foo");
```
mathjazz: @adngdb For some reason I see the log messages while the Translate view is loading. Clicking on Previous/Next buttons OTOH doesn't trigger them. What am I doing wrong? :)

frontend/public/static/locale/es/translate.ftl
```diff
@@ -10,6 +10,15 @@ entitydetails-editor-button-save = Guardar
 entitydetails-editor-button-suggest = Sugerir
 
 
+## Entity Navigation
+## Shows next/previous buttons.
+
+entitynavigation-next = <glyph></glyph>Siguiente
+    .title = Go To Next String (Alt + Down)
```
julen: _Ir a la cadena siguiente (Alt + Abajo)_

frontend/public/static/locale/es/translate.ftl
julen: _Ir a la cadena anterior (Alt + Arriba)_

frontend/public/static/locale/fr/translate.ftl
```diff
@@ -35,6 +35,15 @@ machinery-translation-number-occurrences =
     .title = Nombre d'occurrences de la traduction
 
 
+## Entity Navigation
+## Shows next/previous buttons.
+
+entitynavigation-next = <glyph></glyph>Suivant
+    .title = Go To Next String (Alt + Down)
```
adngdb: _Aller à la chaîne suivante (Alt + Bas)_

frontend/public/static/locale/fr/translate.ftl
adngdb: _Aller à la chaîne précédente (Alt + Haut)_

frontend/src/modules/entitydetails/components/EntityNavigation.js
```diff
@@ -0,0 +1,66 @@
...
+        return <div className='entity-navigation clearfix'>
+            <Localized
+                id="entitynavigation-next"
+                attrs={{ title: true }}
+                glyph={
+                    <i className="fa fa-chevron-down fa-lg"></i>
+                }
+            >
+                <button
+                    id="next"
```
adngdb: I like to avoid using ids on React components, especially ones that are as generic as this one. The idea being that, components being reusable, you run the risk of creating conflicts in the future. Thus far I've removed all (I believe) IDs and used on classes. CSS selectors are good enough that we can do that and still not create styling conflicts.

frontend/src/modules/entitydetails/components/EntityNavigation.js
adngdb: This happens because you actually call these functions when you render the component. And when you click, it calls the return value of these functions, which is `undefined`, so it does nothing. You don't need to pass the entity here, as that variable is a prop that's always the same and always available in this component. 

To fix your problem, remove `entity` from the parameters of your `goTo` methods and get it from the props instead, and don't call the methods when you pass them in `onClick`.

adngdb: Btw, if you haven't seen it already, here's a function that might come in handy: https://github.com/mozilla/pontoon/blob/master/frontend/src/modules/entitieslist/selectors.js#L52

mathjazz:
Thanks for the localizations, @julen and @adngdb!

frontend/src/modules/entitydetails/components/EntityDetails.js
```diff
@@ -128,7 +128,10 @@ export class EntityDetailsBase extends React.Component<InternalProps, State> {
 
         return <section className="entity-details">
             <EntityNavigation
-                entity={ state.selectedEntity }
+                nextEntity={ state.nextEntity }
+                previousEntity={ state.previousEntity }
+                dispatch={ this.props.dispatch }
```
adngdb: `dispatch` shouldn't be used outside of connected containers (like this one). Instead you should pass a method that will, in turn, use `dispatch`. This way you can also remove the `router` attribute.

mathjazz:

PR #1192, opened 2019-02-07
Fix bug 1525539: Support multiline everywhere

mathjazz:
This patch extends the ability to use multiline values wherever in a Fluent Message.

ATM, multiline values are only supported in simple messages, e.g.:
https://mozilla-pontoon-staging.herokuapp.com/de/fluent/demo.ftl/?string=195116

However, using them in select expressions and other more complex messages results in errors, e.g.:
https://mozilla-pontoon-staging.herokuapp.com/de/fluent/demo.ftl/?string=195048

To understand the problem better, you can try to save the translation below to the strings above on stage (which still runs the old code) and then test again locally with the patch applied:
```
This is a multiline
message.
```

### DIFF
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -174,6 +174,17 @@ var Pontoon = (function (my) {
   }
 
   /*
+   * Make sure multiline strings meet Fluent Syntax requirements.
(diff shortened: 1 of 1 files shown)


mathjazz:
@stasm Could you have a look?

PR #1193, opened 2019-02-07
Fix bug 1436985: Speed up locale queries

aniruddha2000:


mathjazz:
I've tested `/teams` loading times on stage with and without the patch.

Average load time (10 attempts):
- without patch: 904.9
- with patch: 928.2

Sorry for misleading you, @aniruddha2000. I'll wontfix the bug.

PR #1195, opened 2019-02-07
Fix bug 1490346 - Copy translations into the Editor on click.

adngdb:
This enables copying a translation in one of the helper tabs (History, Machinery, Locales) into the Editor when clicked. It involves refactoring some of those components to make the data flow the right way. It might not be the best way to do this, but it works and is satisfying for now. We might want to refactor this part of the code again later.

This also separates CSS rules for the various Translation components of the tabs, so that they do not impact each other.

### DIFF
--- a/frontend/public/static/locale/en-US/translate.ftl
+++ b/frontend/public/static/locale/en-US/translate.ftl
@@ -28,6 +28,9 @@ editor-settings-change-all = Change All Settings
 machinery-machinery-search-placeholder =
     .placeholder = Type to search machinery
 
+## Machinery Translation
+## Shows a specific translation from machinery
+
 machinery-translation-copy =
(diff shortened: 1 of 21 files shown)


mathjazz:
If I you enter some text in the translation field [here](https://mozilla-pontoon-staging.herokuapp.com/translate/sl/firefox/all/?string=74127) and click on the first item in History, text remains unchanged.

If you move to the Machinery tab and click on the first item, text still remains unchanged. But if you click on the second item, it gets copied into translation field.

And if you the try to click on the first items in Machinery and History, they will be copied as well.

adngdb:
Thanks @mathjazz, that's a great catch, a side effect that I didn't think about. It should be solved now!

PR #1202, opened 2019-02-12
Fix bug 1519921: Add keyboard shortcut list

mathjazz:
Note: the actual keyboard shortcuts are tracked in [bug 1525984](https://bugzilla.mozilla.org/show_bug.cgi?id=1525984).

### DIFF
--- a/frontend/src/modules/editor/components/EditorSettings.css
+++ b/frontend/src/modules/editor/components/EditorSettings.css
@@ -7,7 +7,7 @@
     cursor: pointer;
     color: #aaa;
     font-size: 22px;
-    padding: 9px 5px;
+    padding: 9px 5px 9px 0;
 }
 
 .editor-settings .selector:hover {
--- a/frontend/src/modules/editor/components/KeyboardShortcuts.js
+++ b/frontend/src/modules/editor/components/KeyboardShortcuts.js
@@ -67,21 +67,27 @@ export class KeyboardShortcutsBase extends React.Component<Props, State> {
                             <Localized id="editor-keyboard-shortcuts-save-translation">
                                 <td>Save Translation</td>
                             </Localized>
-                            <td>
-                                <Localized id="editor-keyboard-shortcuts-save-translation-shortcut">
-                                    <span>Enter</span>
-                                </Localized>
-                            </td>
+                            <Localized
+                                id="editor-keyboard-shortcuts-save-translation-shortcut"
+                                accel={ <span/> }
+                            >
+                                <td>
+                                    { '<accel>Enter</accel>' }
+                                </td>
+                            </Localized>
                         </tr>
                         <tr>
                             <Localized id="editor-keyboard-shortcuts-cancel-translation">
                                 <td>Cancel Translation</td>
                             </Localized>
-                            <td>
-                                <Localized id="editor-keyboard-shortcuts-cancel-translation-shortcut">
-                                    <span>Esc</span>
-                                </Localized>
-                            </td>
+                            <Localized
+                                id="editor-keyboard-shortcuts-cancel-translation-shortcut"
+                                accel={ <span/> }
+                            >
+                                <td>
+                                    { '<accel>Esc</accel>' }
+                                </td>
+                            </Localized>
                         </tr>
                         <tr>
                             <Localized id="editor-keyboard-shortcuts-insert-a-new-line">
@@ -193,11 +199,14 @@ export class KeyboardShortcutsBase extends React.Component<Props, State> {
                             <Localized id="editor-keyboard-shortcuts-copy-from-helpers">
                                 <td>Copy From Helpers</td>
                             </Localized>
-                            <td>
-                                <Localized id="editor-keyboard-shortcuts-copy-from-helpers-shortcut">
-                                    <span>Tab</span>
-                                </Localized>
-                            </td>
+                            <Localized
+                                id="editor-keyboard-shortcuts-copy-from-helpers-shortcut"
+                                accel={ <span/> }
+                            >
+                                <td>
+                                    { '<accel>Tab</accel>' }
+                                </td>
+                            </Localized>
                         </tr>
                     </tbody>
                 </table>
--- a/frontend/src/modules/editor/components/KeyboardShortcuts.js
+++ b/frontend/src/modules/editor/components/KeyboardShortcuts.js
@@ -27,7 +27,7 @@ export class EditorSettingsBase extends React.Component<Props, State> {
     }
 
     toggleVisibility = () => {
-        this.setState((state, props) => {
+        this.setState((state) => {
             return { visible: !state.visible };
         });
     }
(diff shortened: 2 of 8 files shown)


frontend/src/modules/editor/components/EditorSettings.css
```diff
@@ -7,7 +7,7 @@
     cursor: pointer;
     color: #aaa;
     font-size: 22px;
-    padding: 9px 5px;
+    padding: 9px 5px 9px 0;
```
mathjazz: This position the settings icon same as in Translate.Prod.

frontend/src/modules/editor/components/KeyboardShortcuts.js
```diff
@@ -0,0 +1,154 @@
...
+                className="overlay"
+                onClick={ this.toggleVisibility }
+            >
+                <Localized id="keyboard-shortcuts-overlay-title">
+                    <h2>Keyboard Shortcuts</h2>
+                </Localized>
+                <table>
+                    <tbody>
+                        <tr>
+                            <Localized id="keyboard-shortcuts-save-translation">
```
adngdb: So far I've added the module name to those IDs, so in this case I would have called it `editor-keyboard-shortcuts-save-translation`. The idea was to avoid name conflicts as much as possible. Now I'm not really sure if there's value in doing that. I also wonder how much we should share strings between components (for example, `History`, `Machinery` and `OtherLocales` all use the same `No translations` string, and so I used the same ID for all three). I'd like to know what you think about all that.

I'm mostly jumping on the opportunity to start a discussion about IDs, but that's definitely not a blocking point for this PR.

frontend/src/modules/editor/components/KeyboardShortcuts.js
```diff
@@ -0,0 +1,154 @@
...
+                />
+            </Localized>
+
+            { !this.state.visible ? null :
+            <div
+                className="overlay"
+                onClick={ this.toggleVisibility }
+            >
+                <Localized id="keyboard-shortcuts-overlay-title">
+                    <h2>Keyboard Shortcuts</h2>
```
adngdb: Iirc @stasm recommended that strings be put in quotes when they are localizable, but I can't remember why... :)

Here that would be `<h2>{ 'Keyboard Shortcuts' }</h2>` for example.

(Once again that might be something that I didn't do elsewhere in the code base, sorry about that. :/ )

adngdb:
Once we agree on the localization questions, I'll add a commit to your branch with the French translations.

frontend/src/modules/editor/components/KeyboardShortcuts.js
stasm: When the whole contents of the element is simple text, it's fine to use both approaches.

The recommendation to use the string literal in quotes comes from my anticipating that in the future we'll want to extract these source strings automatically. When this happens, it will be important that they are all strings (`<p>{"I'm a string with a Fluent {$variable}."}</p>`) rather than a list of React nodes (`<p>I'm a React node with {things.map(item => item.toString())}.</p>`). In the former case, extraction is trivial. In the latter, not so much.

frontend/src/modules/editor/components/KeyboardShortcuts.js
mathjazz: Let's stick to the tradition you started for now. I'll update the IDs.

We'll probably see later on, when we'll have all strings "extracted" - it might even make sense to split them into multiple files.

frontend/src/modules/editor/components/KeyboardShortcuts.js
mathjazz: So I'm keeping that unchanged.

mathjazz:

frontend/src/modules/editor/components/KeyboardShortcuts.js
```diff
@@ -0,0 +1,210 @@
...
+                            </td>
+                        </tr>
+                        <tr>
+                            <Localized id="editor-keyboard-shortcuts-insert-a-new-line">
+                                <td>Insert A New Line</td>
+                            </Localized>
+                            <td>
+                                <Localized
+                                    id="editor-keyboard-shortcuts-insert-a-new-line-shortcut"
+                                    mod1={ <span/> }
```
adngdb: I think the problem here might be that you need to write those tags with a closing tag instead of the autoclosing form, so `<span></span>`.

frontend/src/modules/editor/components/KeyboardShortcuts.js
```diff
@@ -0,0 +1,210 @@
...
+                </Localized>
+                <table>
+                    <tbody>
+                        <tr>
+                            <Localized id="editor-keyboard-shortcuts-save-translation">
+                                <td>Save Translation</td>
+                            </Localized>
+                            <td>
+                                <Localized id="editor-keyboard-shortcuts-save-translation-shortcut">
+                                    <span>Enter</span>
```
adngdb: These `<span>` should be in the string here too, I guess, otherwise they won't be part of the rendered content and that might break the styling?

frontend/src/modules/search/components/FiltersPanel.js
```diff
@@ -37,8 +37,8 @@ export class FiltersPanelBase extends React.Component<Props, State> {
     }
 
     toggleVisibility = () => {
-        this.setState({
-            visible: !this.state.visible,
+        this.setState((state) => {
+            return { visible: !state.visible };
```
adngdb: Thanks! :+1:

frontend/src/modules/editor/components/KeyboardShortcuts.js
stasm: Actually, `<span/>` should work fine.

The problem here is that `Localized` only runs the overlay logic when it wraps an _element_. Here, it wraps a _string_. The solution is to switch the order of the nesting to:

```js
<Localized id="editor-keyboard-shortcuts-insert-a-new-line-shortcut"
    mod1={ <span/> }
    accel={ <span/> }
>
    <td>
        { '<mod1>Shift</mod1> + <accel>Enter</accel>' }
    </td>
</Localized>
```

Sorry about this obscure catch. I'll file an issue about this in `fluent-react`.

frontend/src/modules/editor/components/KeyboardShortcuts.js
```diff
@@ -0,0 +1,210 @@
...
+
+type State = {|
+    visible: boolean,
+|};
+
+
+/*
+ * Shows a list of keyboard shortcuts.
+ */
+export class EditorSettingsBase extends React.Component<Props, State> {
```
stasm: Nit: this should probably be called something else :)

mathjazz:

frontend/src/modules/editor/components/KeyboardShortcuts.js
stasm: I filed https://github.com/projectfluent/fluent.js/issues/339 in `fluent-react`.

adngdb:
(Note that I added French translations. )

PR #1205, opened 2019-02-18
Fix bug 1524602 - Hide Editor actions and settings when user is logged out.

adngdb:


adngdb:
So, this doesn't work yet, because the text `Sign in to translate` doesn't have the link set correctly. 

In the DOM I see this: 

```html
<p class="banner">
    <a href="/accounts/fxa/login/?scope=profile%3Auid+profile%3Aemail+profile%3Adisplay_name"></a>
    Sign in to translate.
</p>
```

I'm at a complete loss there... @stasm, please, help.

adngdb:
It's solved thanks for @stasm! `<link>` was not usable in a Fluent template because it is a reserved HTML tag, as far as I understand. Using `<a>` instead is fine.

stasm:
It's reserved _and_ it's not allowed to have any content. It's _content model_ is defined to be [empty](https://www.w3.org/TR/html50/document-metadata.html#the-link-element). So `<link>Foo</link>` parses as `[<link>, "Foo"]`, and the closing `</link>` is ignored.

frontend/public/static/locale/en-US/translate.ftl
```diff
@@ -4,10 +4,11 @@
...
 
-entitydetails-editor-button-copy = Copy
-entitydetails-editor-button-clear = Clear
-entitydetails-editor-button-save = Save
-entitydetails-editor-button-suggest = Suggest
+editor-editor-sign-in-to-translate = <a>Sign in</a> to translate.
+editor-editor-button-copy = Copy
+editor-editor-button-clear = Clear
+editor-editor-button-save = Save
+editor-editor-button-suggest = Suggest
```
mathjazz: `editor-editor` looks weird... Is that because you're simply following the `$module-$component` convention?

frontend/src/modules/editor/components/Editor.js
```diff
@@ -79,31 +79,44 @@ export default class Editor extends React.Component<Props, State> {
                 updateTranslation={ this.updateTranslation }
             />
             <menu>
+            { !this.props.user.isAuthenticated ?
+                <Localized
+                    id="editor-editor-sign-in-to-translate"
+                    a={
+                        <a href='/accounts/fxa/login/?scope=profile%3Auid+profile%3Aemail+profile%3Adisplay_name'></a>
```
mathjazz: This is breaking tests:
```
./src/modules/editor/components/Editor.js
  Line 86:  Anchors must have content and the content must be accessible by a screen reader  jsx-a11y/anchor-has-content
```

frontend/public/static/locale/en-US/translate.ftl
adngdb: Yeah, I don't like it much either. At some point we should have a conversation about those IDs, and decide on a clear rule.

stasm: I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1528926 to kick off the conversation about the id naming scheme.

PR #1206, opened 2019-02-18
Bug 1522520: Store homepage as html in database

VishalCR7:
This addresses part-1 of the spec provided for bug #1522520 i.e. moving whole html content to database.
Changing the title is still left and needs to be addressed( Probably by adding a title field to Homepage table)


pontoon/homepage/models.py
```diff
@@ -1,6 +1,12 @@
...
+from django.db import models
 
-# Create your models here.
+
+class Homepage(models.Model):
+    text = models.TextField()
+    created_at = models.DateTimeField(auto_now_add=True)
+
+    class Meta:
+        ordering = ('-created_at', )
```
mathjazz: Is that really necessary? Default ordering is by ID, which should result in the same but reverse order as if ordering by `auto_now_add=True` fields.

pontoon/homepage/views.py
```diff
@@ -28,6 +30,23 @@ def homepage(request):
...
+        'slug': 'tutorial',
+        'part': 'playground',
+    })
+
+    intro_url = reverse('pontoon.translate.locale.agnostic', kwargs={
+        'slug': 'pontoon-intro',
+        'part': 'messages.properties',
+    })
+
+    homepage = Homepage.objects.all()[0]
```
mathjazz: That's the same as `Homepage.objects.first()`. But you really want `Homepage.objects.last()` if you change ordering.

mathjazz:
>  Changing the title is still left and needs to be addressed( Probably by adding a title field to Homepage table)

Or we could just make it more general (e.g. "Pontoon") and not customizable.

pontoon/homepage/views.py
```diff
@@ -28,6 +30,23 @@ def homepage(request):
...
+        'slug': 'tutorial',
+        'part': 'playground',
+    })
+
+    intro_url = reverse('pontoon.translate.locale.agnostic', kwargs={
+        'slug': 'pontoon-intro',
+        'part': 'messages.properties',
+    })
+
+    homepage = Homepage.objects.first()
```
mathjazz: You mean `.last()`?

pontoon/homepage/views.py
```diff
@@ -28,6 +30,23 @@ def homepage(request):
...
-    return render(request, 'homepage.html', {
+    playground_url = reverse('pontoon.translate.locale.agnostic', kwargs={
+        'slug': 'tutorial',
+        'part': 'playground',
+    })
+
+    intro_url = reverse('pontoon.translate.locale.agnostic', kwargs={
+        'slug': 'pontoon-intro',
+        'part': 'messages.properties',
+    })
```
adngdb: Why did you move these out of the template? I think it's fine to keep them there, and use the `url` helper to construct the URLs.

I guess you followed the `start_url` pattern, but I think that one is an exception, because we want to be smart with it. The other 2 don't change depending on the context, so they should stay in the template in my opinion.

pontoon/homepage/models.py
adngdb: I agree, this is not useful in the current state, so let's rely on the database ordering by ID by default. Less strain on the database is always good. :-)

adngdb:
> > Changing the title is still left and needs to be addressed( Probably by adding a title field to Homepage table)
> 
> Or we could just make it more general (e.g. "Pontoon") and not customizable.

I actually like that external users would be able to change the main title. I'm not sure if that should be just on the homepage, or if we should instead make the overall title (the `Pontoon` name) configurable? Anyway, that can be kept for later.

pontoon/homepage/views.py
VishalCR7: I was facing problems in rendering the template using template.render() . It was throwing a parse error while using url() or static() words in template.

mathjazz: I think in the case of a customizable homepage it's actually OK to move the logic out of the template, so that editing will be easier. I'd rename `intro_url` to `in_context_url`, though.

pontoon/homepage/views.py
VishalCR7: Since we would like to display the most recent iteration, so shouldn't it be first, considering we used "-created_at" ordering in the models?

mathjazz: I thought you'd remove that:
https://github.com/mozilla/pontoon/pull/1206/files#r257930665

pontoon/homepage/views.py
adngdb: Not being able to create links in the template might be a problem in the future. It might be worth trying to figure out what that is so, and fix it.

pontoon/homepage/templates/homepage_content.html
```diff
@@ -128,7 +128,7 @@ <h2>One more thing.</h2>
           making you the first proofreader and tester.
           Talk about context?
         </p>
-        <a class="button primary" href="{{ intro_url }}">Try In-context Localization</a>
+        <a class="button primary" href="{% url 'pontoon.translate.locale.agnostic' 'pontoon-intro' 'messages.properties' %}">Try In-context Localization</a>
```
adngdb: As I thought, this is django template syntax. I guess it's fine to use that instead of jinja2. :-)

VishalCR7:
@mathjazz  as directed, I have added the edit homepage button according to my very limited design sense. Please have a look and let me know.
image

mathjazz:
@VishalCR7 Nice work! I've moved the button from the header to the homepage, which allows us to:
* make the button more visible,
* keep the notification/profile icon position same as on other pages,
* keep the code where it belongs semantically,
* make the condition for when to show the button simpler.

I'd love to hear your thoughts on this.

What other work remains to be done here (apart from removing `Meta.ordering` in the model)?

VishalCR7:
Ace added!
I have used a separate commit for adding django_ace to requirements.txt. If it conflicts with jotes' current PR then I will just amend the last commit.
image

mathjazz:
That looks good, nice work! The only thing that remains from Step 2 of the [spec](https://docs.google.com/document/d/1TrV6MO-58LyiVpvrpPOvjneDJucoRnlr-8vh13bp1S0/edit#) is this:

_That page should show a simple documentation about the expected structure of the HTML content. We want to keep that content compatible with the current slide structure of the home page, so the content will be expected to be separated into sections, one section per slide._

I suggets you add that as `help_text` in the model and the Django Admin will pick it up like it does for [projects](http://localhost:8000/a/base/project/95/change/).

mathjazz:
Great work!

I'd fix the padding/margin in the Admin. Please also remove the `h3` to make the help text consistent with the rest of the app and future-compatible with any style changes in Django 2.

You can simply apply this patch:

```diff
diff --git a/pontoon/homepage/admin.py b/pontoon/homepage/admin.py
index 555e6c9e..136f658d 100644
--- a/pontoon/homepage/admin.py
+++ b/pontoon/homepage/admin.py
@@ -12,5 +12,9 @@ class HomepageAdmin(admin.ModelAdmin):
     list_display = ('pk', 'created_at',)
     form = forms.HomepageForm
 
+    class Media:
+        css = {
+                'all': ('css/homepage_admin.css',)
+        }
 
 admin.site.register(models.Homepage, HomepageAdmin)
diff --git a/pontoon/homepage/forms.py b/pontoon/homepage/forms.py
index 4840f5ba..5c5a9866 100644
--- a/pontoon/homepage/forms.py
+++ b/pontoon/homepage/forms.py
@@ -12,8 +12,8 @@ class HomepageForm(forms.ModelForm):
             'text': AceWidget(mode='django', theme='tomorrow_night', width="900px", height="400px"),
         }
         help_texts = {
-            'text': """<h3>To keep the content compatible with the current slide structure of
+            'text': """To keep the content compatible with the current slide structure of
                 the home page, the content is expected to be separated into sections with class
-                "section".</h3>"""
+                "section"."""
         }
 
diff --git a/pontoon/homepage/static/css/homepage_admin.css b/pontoon/homepage/static/css/homepage_admin.css
new file mode 100755
index 00000000..1ebbe15e
--- /dev/null
+++ b/pontoon/homepage/static/css/homepage_admin.css
@@ -0,0 +1,4 @@
+form .aligned div.help {
+    margin-left: 0;
+    padding-left: 0;
+}
```

pontoon/homepage/static/css/homepage.css
```diff
@@ -100,6 +99,27 @@ h2 {
   background: #7BC876;
 }
 
+/* Edit Homepage button */
+#edit-homepage {
+	margin: 0 auto;
+	width: 980px;
```
adngdb: There's an indentation bug here. :)

pontoon/homepage/static/css/homepage_admin.css
```diff
@@ -0,0 +1,4 @@
+form .aligned div.help {
+    margin-left: 0;
+    padding-left: 0;
```
adngdb: You should use 2-space indentation here please.

PR #1207, opened 2019-02-19
Fix bug 1490343: Show/hide diff

mathjazz:
TODO:
- [x] Add the actual diff
- [x] Add tests
- [x] ~~Do not copy translation into editor when review buttons and show/hide diff is clicked~~

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -5,6 +5,7 @@
   "dependencies": {
     "brace": "0.11.1",
     "connected-react-router": "4.5.0",
+    "diff-match-patch": "1.0.4",
     "flow-bin": "0.84.0",
     "fluent": "0.9.1",
     "fluent-langneg": "0.1.0",
--- a/frontend/public/static/locale/en-US/translate.ftl
+++ b/frontend/public/static/locale/en-US/translate.ftl
@@ -121,6 +121,11 @@ history-history-no-translations = No translations available.
 history-translation-copy =
     .title = Copy Into Translation (Tab)
 
+history-translation-hide-diff = Hide diff
+    .title = Hide diff against the currently active translation
+history-translation-show-diff = Show diff
+    .title = Show diff against the currently active translation
+
 history-translation-button-delete =
     .title = Delete
 
--- a/frontend/src/core/diff/components/TranslationDiff.js
+++ b/frontend/src/core/diff/components/TranslationDiff.js
@@ -6,7 +6,7 @@ import DiffMatchPatch from 'diff-match-patch';
 import './TranslationDiff.css';
 
 
-type InternalProps = {|
+type Props = {|
     base: string,
     target: string,
 |};
@@ -22,7 +22,7 @@ const dmp = new DiffMatchPatch();
  * Added slices are wrapped in <ins>.
  * Removed slices are wrapped in <del>.
  */
-export class TranslationDiffBase extends React.Component<InternalProps> {
+export default class TranslationDiff extends React.Component<Props> {
     getDiff(base: string, target: string) {
         const diff = dmp.diff_main(base, target);
     
@@ -53,6 +53,3 @@ export class TranslationDiffBase extends React.Component<InternalProps> {
         });    
     }
 }
-
-
-export default TranslationDiffBase;
--- a/frontend/src/core/diff/components/TranslationDiff.js
+++ b/frontend/src/core/diff/components/TranslationDiff.js
@@ -1,7 +1,6 @@
 /* @flow */
 
 import * as React from 'react';
-import { connect } from 'react-redux';
 import DiffMatchPatch from 'diff-match-patch';
 
 import './TranslationDiff.css';
@@ -56,4 +55,4 @@ export class TranslationDiffBase extends React.Component<InternalProps> {
 }
 
 
-export default connect()(TranslationDiffBase);
+export default TranslationDiffBase;
--- a/frontend/src/core/diff/components/TranslationDiff.js
+++ b/frontend/src/core/diff/components/TranslationDiff.js
@@ -1,0 +1,59 @@
+/* @flow */
+
+import * as React from 'react';
+import { connect } from 'react-redux';
+import DiffMatchPatch from 'diff-match-patch';
+
+import './TranslationDiff.css';
+
+
+type InternalProps = {|
+    base: string,
+    target: string,
+|};
+
+
+const dmp = new DiffMatchPatch();
+
+
+/**
+ * Render diff between the base and the target string.
+ * 
+ * Unchanged slices are wrapped in <span>.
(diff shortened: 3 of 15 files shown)


frontend/src/modules/history/components/Translation.js
```diff
@@ -138,6 +158,38 @@ export default class Translation extends React.Component<Props> {
...
+                    { index === 0 ? null :
+                        this.state.isDiffVisible ?
+                            // Hide Diff
+                            <Localized
+                                id='history-translation-hide-diff'
+                                attrs={{ title: true }}
+                            >
+                                <a
+                                    className='toggle-diff'
+                                    href='#'
```
mathjazz: This appends `#` to the URL when clicked. Is there a better workaround then using another element, e.g. `<button>`?

mathjazz: /cc @adngdb

adngdb: Nope, using a `<button>` is the correct way to solve this. `<a>` should only be used when you're actually changing the URL.

frontend/src/modules/history/components/History.js
```diff
@@ -73,7 +73,7 @@ export default class History extends React.Component<Props> {
                         deleteTranslation={ deleteTranslation }
                         updateEditorTranslation={ updateEditorTranslation }
                         updateTranslationStatus={ updateTranslationStatus }
-                        key={ key }
+                        index={ index }
```
adngdb: You'll need to keep the `key` part around, React requires that you pass a unique `key` attribute to elements of a list. You can for example use `key={ key } index={ key }` to also pass that down to the component.

frontend/package.json
```diff
@@ -5,6 +5,7 @@
   "dependencies": {
     "brace": "0.11.1",
     "connected-react-router": "4.5.0",
+    "diff-match-patch": "1.0.4",
```
mathjazz: How do I add packages to frontend? Shouldn't yarn.lock also be updated?

We should document that:
https://mozilla-pontoon.readthedocs.io/en/latest/dev/contributing.html#dependencies

frontend/public/static/locale/en-US/translate.ftl
```diff
@@ -121,6 +121,11 @@ history-history-no-translations = No translations available.
 history-translation-copy =
     .title = Copy Into Translation (Tab)
 
+history-translation-hide-diff = Hide diff
+    .title = Hide diff against the currently active translation
+history-translation-show-diff = Show diff
+    .title = Show diff against the currently active translation
+
```
mathjazz: @adngdb Could you please add fr (and es?) translations? :)

mathjazz:
> Do not copy translation into editor when review buttons and show/hide diff is clicked

Filed: https://bugzilla.mozilla.org/show_bug.cgi?id=1529491

frontend/package.json
adngdb: It's a bit cumbersome but until I figure out how to deal with our docker volumes better, here's what I do:

```shell
cd ./frontend/
yarn add my-package
rm -rf node_modules
cd ..
make build
```

I'm making a note about the docs, I need to update a bunch of them anyway.

frontend/public/static/locale/en-US/translate.ftl
adngdb: Yo no hablo español. :cry:

frontend/src/modules/history/components/Translation.js
```diff
@@ -99,8 +118,50 @@ export default class Translation extends React.Component<Props> {
...
+    getDiff(base: string, target: string) {
+        const diff = dmp.diff_main(base, target);
+
+        dmp.diff_cleanupSemantic(diff);
+        dmp.diff_cleanupEfficiency(diff);
+
+        return diff;
+    }
+
+    renderDiff(translation: DBTranslation, activeTranslation: DBTranslation) {
```
adngdb: Nit: I would directly pass the strings here, instead of the full objects, because you don't need the objects anyway and that makes the function more "dumb" and thus easier to test.

frontend/src/modules/history/components/Translation.js
```diff
@@ -99,8 +118,50 @@ export default class Translation extends React.Component<Props> {
...
+                    return <ins key={ index }>{ slice }</ins>;
+
+                case DiffMatchPatch.DIFF_DELETE:
+                    return <del key={ index }>{ slice }</del>;
+
+                default:
+                    return <span key={ index }>{ slice }</span>;
+            }
+        });
+    }
```
adngdb: These `getDIff` and `renderDiff` function are not strictly related to this component, and I assume we might want to reuse that code in the Machinery tab, too. How would you feel about putting them somewhere else? Maybe it could be a component somewhere in `core`? Then below you would either render the simple string or the component.

frontend/src/modules/history/components/Translation.js
```diff
@@ -138,6 +199,36 @@ export default class Translation extends React.Component<Props> {
                         />
                     </div>
                     <menu className='toolbar'>
+                    { index === 0 ? null :
+                        this.state.isDiffVisible ?
```
adngdb: I think double ternary operators is not a great pattern, it really hinders readability. When that happens in my code, I tend to extract that code to a `renderX` method, and use regular `if` to return the expected result.

frontend/public/static/locale/en-US/translate.ftl
adngdb: French translation: https://github.com/mathjazz/pontoon/pull/8

frontend/src/modules/history/components/Translation.js
```diff
@@ -199,36 +233,9 @@ export default class Translation extends React.Component<Props, State> {
...
-                                    className='toggle-diff show'
-                                    title='Show diff against the currently active translation'
-                                    onClick={ this.toggleDiff }
-                                >
-                                    { 'Show diff' }
-                                </button>
-                            </Localized>
-                    }
+
+                    { index > 0 ? this.renderDiffToggle() : null }
```
adngdb: I would even go further and put only `this.renderDiffToggle(index)` here, and do the `if index` test in that method. But it's a nit, if you prefer this way it's fine by me.

mathjazz:
@adngdb This is now ready for another look.

frontend/src/core/diff/components/TranslationDiff.js
```diff
@@ -0,0 +1,59 @@
...
+    
+                default:
+                    return <span key={ index }>{ slice }</span>;
+            }
+        });    
+    }
+}
+
+
+export default connect()(TranslationDiffBase);
```
adngdb: This component doesn't need to be connected. We only "connect" components that need a direct access to the redux store, which is not the case here.

PR #1208, opened 2019-02-21
Fix bug 1528177 - Add links to Sign in and out in Translate.Next.

adngdb:
This also does a bunch of refactoring, putting all user-related components into the user module and exposing a single, catch-all UserControls component for simplicity of use.

### DIFF
--- a/frontend/src/core/user/components/SignInLink.js
+++ b/frontend/src/core/user/components/SignInLink.js
@@ -1,0 +1,22 @@
+/* @flow */
+
+import * as React from 'react';
+
+
+type Props = {|
+    children?: React.Node,
+|};
+
+
+/*
+ * Render a link to the Sign In process.
+ */
+export default class SignInLink extends React.Component<Props> {
+    render() {
+        return <a
+            href='/accounts/fxa/login/?scope=profile%3Auid+profile%3Aemail+profile%3Adisplay_name'
+        >
+            { this.props.children }
+        </a>;
+    }
+}
--- a/frontend/src/core/user/components/SignInLink.js
+++ b/frontend/src/core/user/components/SignInLink.js
@@ -13,8 +13,21 @@ type Props = {|
  * Render a link to the Sign In process.
  */
 export default class SignInLink extends React.Component<Props> {
+    generateSignInURL() {
+        const base = this.props.url;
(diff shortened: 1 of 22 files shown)


mathjazz:
Nice work! A couple of notes:

1. After successful login, you get redirected to the homepage. Instead, you should end up on the same page, like in Translate.Current:

2. There should be no confirmation step after Signing Out. Not only because such UX is inconsistent with the Translate.Current UX, which we're copying, but also because signing in and out isn't an irreversible step. Additionally, we'll change the Sign in/Sign out UX soon [anyways](https://bugzilla.mozilla.org/show_bug.cgi?id=1486506).

frontend/src/core/user/components/SignInLink.js
```diff
@@ -0,0 +1,34 @@
...
+        const base = this.props.url;
+        const startSign = base.match(/\?/) ? '&': '?';
+
+        return (
+            base +
+            startSign +
+            'next=' +
+            window.location.pathname +
+            window.location.search
+        );
```
mathjazz: Is there a more elegant way of appending a parameter to the URL?

frontend/src/core/user/components/UserControls.js
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    render() {
+        const { router, user } = this.props;
+
+        return <div className='user-controls'>
+            <UserAutoUpdater getUserData={ this.getUserData } />
+
+            { user.isAuthenticated ?
+                <SignOut dispatch={ this.props.dispatch } url={ user.signOutURL } /> :
```
mathjazz: I haven't seen `dispatch` being passed around like this. Am I doing something wrong?

mathjazz:
@adngdb This now works as expected from my perspective. Could you please r? (I cannot assign a review to you, because it's your branch).

frontend/src/core/user/components/UserControls.js
adngdb: Yes, `dispatch` should never be passed around. Instead you should create a new method here that uses `dispatch` and pass that method down. I can take it up from here though, let me review by pushing some commits. :sweat_smile:

adngdb:
@mathjazz I think this is in working order. Wanna take a last look at it and merge if you're happy with it?

PR #1211, opened 2019-02-22
No bug - Refresh our Contributing documentation file.

adngdb:
I'm not exactly sure how to talk about the 2 different frontend apps (Translate.Next and the code for tags). The documentation for Translate.Next is in its README, so I could point there maybe? Or I could not even talk about Translate.Next for now and wait until it's turned on? What do you think?

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -28,14 +28,29 @@ Database
 ========
 
 By default, you will have default data loaded for only the Pontoon Intro project.
-If you have a database dump, you can load it into your PostgreSQL database by running::
+If you have a database dump, you can load it into your PostgreSQL database by running:
 
-  $ make loaddb DB_DUMP_FILE=path/to/my/dump
+.. code-block:: shell
+
+    $ make loaddb DB_DUMP_FILE=path/to/my/dump
 
 Note that your database container needs to be running while you do that. You
-can start just the postgresql container by runnin::
+can start just the postgresql container by running:
 
-  $ docker-compose run postgresql -d
+.. code-block:: shell
+
+    $ docker-compose run postgresql -d
+
+
+Interactive shell
+=================
+
+If you need to run specific commands, that are not covered by our `Makefile`,
+you can start an interactive shell inside a Pontoon container:
+
+.. code-block:: shell
+
+    $ make shell
 
 
 Python code conventions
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -28,50 +28,80 @@ Database
...
+that in most cases, it is better to fix the issues than ignoring them.
 
 
 Javascript code conventions
 ===========================
 
+Outside the ``frontend`` folder, we don't follow strict rules other than using
 2-space indentation.
 
+Inside ``frontend`` (which contains the new Translate.Next app), we use 4-space
```
mathjazz: I'd just say "contains the translate app" to be future-compatible.

CONTRIBUTING.rst
```diff
@@ -122,19 +152,24 @@ Dependencies for production Pontoon are in ``requirements.txt``. Development dep
 
 Note that we use a specific format for our dependencies, in order to make them more maintainable. When adding a new requirement, you should add it to the appropriate section and group it with its sub-dependencies if applicable.
 
-For example, to add ``foobar`` version 5::
+For example, to add ``foobar`` version 5:
```
mathjazz: What's the deal with replacing all `::` with `:`?

CONTRIBUTING.rst
```diff
@@ -122,19 +152,24 @@ Dependencies for production Pontoon are in ``requirements.txt``. Development dep
...
 
-  $ hashin -r requirements.txt foobar==5
+    $ hashin -r requirements.txt foobar==5
 
 Then open ``requirements.txt`` and move the added dependencies to:
+
 * the first section if it has no other requirements
 * the second section if it has sub-dependencies, and add all its dependencies there as well.
 
 That format is documented more extensively inside the ``requirements.txt`` file.
```
mathjazz: Should we explain adding frontend requirements here, as tedious as they are ATM?

adngdb: Hah, well, here's my answer: https://github.com/mozilla/pontoon/pull/1212 :smile:

CONTRIBUTING.rst
adngdb: That's `.rst` for you. If you end a line with `::` then it considers the next one as a block of code. Thus, because I added all those `.. code-block:: shell`, I had to remove the colons, otherwise it would turn `.. code-block:: shell` into a code block and show the actual code as regular text.

CONTRIBUTING.rst
mathjazz: Ah, great! :)

PR #1212, opened 2019-02-22
No bug - Explain process to add JS dependencies.

adngdb:


frontend/README.md
```diff
@@ -106,6 +108,24 @@ As far as we know, it is not possible to make that work in Chrome or Edge. This
 If you can't turn on websockets, you will see errors in the console (that's not very impacting) and you'll have to reload your Django server regularly, because polling requests don't close, and after so many web page reloads, the Django process won't be able to accept new requests.
 
 
+## Dependencies
+
+We manage our JavaScript dependencies with `yarn`. Dependencies and their version are listed in the `package.json` file, under `dependencies` for production and `devDependencies` for development.
```
mathjazz: version -> versions

adngdb: Are you sure? Because each dependency has just one version, so I thought it would be singular. Should we ask a native speaker? :stuck_out_tongue:

mathjazz: I guess both should be OK. I wonder what would a native speaker say. :)

adngdb: @gueroJeff could you help us figure this out please? :-)

gueroJeff: Sorry, didn't see this til now. It would be versions, because although each dependency might have its own version, you're making the dependencies plural, requiring also that the versions would be plural.

adngdb: @mathjazz

PR #1213, opened 2019-02-25
Fix bug 1514722: Highlight user preferred locales in Locales tab

mathjazz:
This should work with the exception of the failing test, but I have a couple of questions:

1. Should we move `MachineryCount` and `OtherLocalesCount` components to corresponding modules (`machinery` and `otherlocales`)?

2. Should we rename `Locales` component to `OtherLocales`?

### DIFF
--- a/frontend/src/modules/entitydetails/components/Tools.js
+++ b/frontend/src/modules/entitydetails/components/Tools.js
@@ -10,6 +10,7 @@ import { Machinery } from 'modules/machinery';
 import { Locales } from 'modules/otherlocales';
 
 import MachineryCount from './MachineryCount';
+import OtherLocalesCount from './OtherLocalesCount';
 
 import type { Locale } from 'core/locales';
 import type { NavigationParams } from 'core/navigation';
@@ -38,6 +39,23 @@ type Props = {|
  * Shows the metadata of the entity and an editor for translations.
  */
 export default class Tools extends React.Component<Props> {
+    preferredCount() {
+        const { otherlocales, user } = this.props;
+
+        if (!user.isAuthenticated) {
+            return 0;
+        }
+
+        const preferred = otherlocales.translations.reduce((count, item) => {
+            if (user.preferredLocales.indexOf(item.code) > -1) {
+                return count + 1;
+            }
+            return count;
+        }, 0);
+
+        return preferred;
+    }
+
     render() {
         const {
             history,
@@ -72,7 +90,10 @@ export default class Tools extends React.Component<Props> {
                 <Tab>
                     Locales
                     { !otherlocalesCount ? null :
-                    <span className={ 'count' }>{ otherlocalesCount }</span>
+                    <OtherLocalesCount
+                        otherlocales={ otherlocales }
+                        preferredCount={ this.preferredCount() }
+                    />
                     }
                 </Tab>
             </TabList>
@@ -98,8 +119,10 @@ export default class Tools extends React.Component<Props> {
             <TabPanel>
                 <Locales
                     otherlocales={ otherlocales }
+                    user={ user }
                     parameters={ parameters }
                     updateEditorTranslation={ updateEditorTranslation }
+                    preferredCount={ this.preferredCount() }
                 />
             </TabPanel>
         </Tabs>;
--- a/frontend/src/modules/otherlocales/components/Locales.js
+++ b/frontend/src/modules/otherlocales/components/Locales.js
@@ -9,12 +9,15 @@ import Translation from './Translation';
 
 import type { Navigation } from 'core/navigation';
 import type { LocalesState } from '..';
+import type { UserState } from 'core/user';
 
 
 type Props = {|
     otherlocales: LocalesState,
+    user: UserState,
     parameters: Navigation,
     updateEditorTranslation: (string) => void,
+    preferredCount: number,
 |};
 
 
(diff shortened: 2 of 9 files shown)


frontend/src/modules/otherlocales/components/Locales.test.js
```diff
@@ -17,18 +17,54 @@ describe('<Locales>', () => {
...
+        };
+        const wrapper = shallow(
+            <Locales
+                otherlocales={ otherlocales }
+                parameters={ params }
+                user={ user }
+            />
+        );
+
+        expect(wrapper.find('Translation:first-child .stress').text()).toContain('br');
```
mathjazz: This test fails with:
```
Method “text” is only meant to be run on a single node. 0 found instead."
```

`Translation:first-child` is found, but for some reason `Translation:first-child .stress` isn't.

frontend/src/modules/entitydetails/components/Tools.js
```diff
@@ -38,6 +39,23 @@ type Props = {|
...
+
+        const preferred = otherlocales.translations.reduce((count, item) => {
+            if (user.preferredLocales.indexOf(item.code) > -1) {
+                return count + 1;
+            }
+            return count;
+        }, 0);
+
+        return preferred;
+    }
```
adngdb: I think this should actually be a selector. This way, we won't recompute it every time there's a render, but only when the depending data changes.

frontend/src/modules/otherlocales/components/Locales.test.js
adngdb: That is because `shallow` only creates the top component, and none of its children. In this case, that means `Translation` is not constructed, and thus you cannot access its content. But that is good, because in tests you should only test the component you're interested in, and no others.

In this case I believe you want to test the outcome of `sortByPreferred` instead of looking at what's rendered. You can do that by calling `wrapper.instance().sortByPreferred()`.

frontend/src/modules/otherlocales/components/Locales.js
```diff
@@ -9,19 +9,56 @@ import Translation from './Translation';
...
+                return 1;
+            }
+            else if (indexA > indexB) {
+                return -1;
+            }
+            else {
+                return 0;
+            }
+        });
+    }
```
adngdb: I think this should also be a selector. There will be a lot of cases where you re-do that computation but the outcome will be exactly the same, because the inputs didn't change.

adngdb: In general, you should avoid having logic in your `render` methods as much as possible. Put as much as makes sense in either actions, reducers or selectors.

frontend/src/modules/otherlocales/components/Locales.test.js
adngdb: Note that moving the ordering logic to a selector will make it a lot simpler to test, which I believe is also a good argument in favor of the selector. :)

mathjazz:
I'm done with moving files around and renaming. I have yet to move `preferredCount` to selectors and reshuffle tests.

@adngdb Any early feedback?

mathjazz:

frontend/src/modules/otherlocales/components/OtherLocales.js
```diff
@@ -43,15 +60,30 @@ export default class Locales extends React.Component<Props> {
...
+const mapStateToProps = (state: Object): Props => {
+    return {
+        orderedOtherLocales: selectors.getOrderedOtherLocales(state),
+        preferredCount: selectors.getPreferredLocalesCount(state),
+        otherlocales: state.otherlocales,
+        user: state.user,
+    };
+};
+
+export default connect(mapStateToProps)(OtherLocalesBase);
```
adngdb: Here too, I believe this shouldn't be a connected component. It might make for a lot of props to pass down from `EntityDetails`, but for the sake of consistency I think we should do that for now. Maybe we will change that later on, but if we do we'll do all three tab components the same way.

frontend/src/modules/otherlocales/components/Count.js
```diff
@@ -39,3 +42,13 @@ export default class Count extends React.Component<Props> {
...
+
+
+const mapStateToProps = (state: Object): Props => {
+    return {
+        otherlocales: state.otherlocales,
+        preferredCount: selectors.getPreferredLocalesCount(state),
+    };
+};
+
+export default connect(mapStateToProps)(CountBase);
```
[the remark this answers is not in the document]
adngdb: On that topic, here's an article that I found useful to understand the difference between connected (containers) and non-connected (components): https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0 (not the header that says this is not "law", and the author changed his mind on it, but I still find that it is a useful way to think about these when beginning).

frontend/src/modules/otherlocales/selectors.js
```diff
@@ -0,0 +1,90 @@
...
+import { createSelector } from 'reselect';
+
+import api from 'core/api';
+
+import type { LocalesState } from '.';
+import type { UserState } from 'core/user';
+
+
+const otherLocalesSelector = (state): LocalesState => state.otherlocales;
+const userSelector = (state): UserState => state.user;
```
adngdb: He he he, pretty sure that's a mistake I introduced myself again, but here we should use the `NAME` constant from each module instead of hard-coding the module names. :grin:

frontend/src/modules/entitydetails/components/EntityDetails.js
```diff
@@ -213,6 +213,8 @@ export class EntityDetailsBase extends React.Component<InternalProps, State> {
                 locale={ state.locale }
                 machinery={ state.machinery }
                 otherlocales={ state.otherlocales }
+                orderedOtherLocales={ state.orderedOtherLocales }
+                preferredCount={ state.preferredCount }
```
adngdb: In this context, it's unclear what `preferredCount` refers to. Could you rename that to `preferredLocalesCount` at least until it's in a context where we know it's about locales?

mathjazz: OK, `preferredOtherLocalesCount` even.

PR #1214, opened 2019-02-26
Fix bug 1528293: Convert read-only AJAX responses to JSON

mathjazz:
This fixes the following error shown in the JS Console when a user tries to submit a translation to a read-only project:
```
SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data
```

While the error will be harder to trigger when [bug 1528273](https://bugzilla.mozilla.org/show_bug.cgi?id=1528273) gets fixed, we should still convert all AJAX responses to JSON, because that's what our new frontend expects. So this is just a first step. I'd be happy to make the next (and the final) one in a follow-up commit.

It shouldn't be too hard to keep compatibility with Translate.Current if we keep the status codes unchanged, because that's what our old frontend expects.

@adngdb Thoughts?

### DIFF
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -15,7 +15,6 @@ from django.db.models import Q
 from django.http import (
     Http404,
     HttpResponse,
-    HttpResponseBadRequest,
     HttpResponseForbidden,
     JsonResponse,
     StreamingHttpResponse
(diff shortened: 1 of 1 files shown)


mathjazz:
I've add 3 additional commits which make sure that all AJAX views in `pontoon.base.views` return JSON. @adngdb Could you have another look, please?

adngdb:
There's one `HttpResponseForbidden` left in `base/views.py`, do you want to remove it? Otherwise, r+ again.

mathjazz:
> There's one `HttpResponseForbidden` left in `base/views.py`, do you want to remove it? Otherwise, r+ again.

That's intentional - it's not an AJAX view, so we rightfully show https://pontoon.mozilla.org/403/.

PR #1215, opened 2019-02-26
Fix Bug-1521820: [sync] Add tests for FTL file format

aniruddha2000:


pontoon/sync/tests/formats/test_ftl.py
```diff
@@ -0,0 +1,107 @@
...
+    def test_parse_with_no_source_path(self):
+        contents = dedent("""text = Arise, awake and do not stop until the goal is reached.""")
+        path = create_named_tempfile(
+            contents,
+            prefix='strings',
+            suffix='.ftl',
+            directory=self.tempdir,
+        )
+        ftl.parse(path, source_path=None, locale=None)
+        assert_true(ftl.FTLResource(path, locale=None, source_resource=None))
```
adngdb: These last 2 tests do nothing. It is expected to test the output of the `parse` function, and you don't do anything with that output.

Also, please configure your code editor so that it automatically adds an empty line at the end of files.

adngdb:
Please take a look at the checks on travis, there's a failure with your code.

pontoon/sync/tests/formats/test_ftl.py
```diff
@@ -0,0 +1,111 @@
...
+    def setUp(self):
+        super(FTLResourceTests, self).setUp()
+        self.tempdir = tempfile.mkdtemp()
+
+    def tearDown(self):
+        super(FTLResourceTests, self).tearDown()
+        shutil.rmtree(self.tempdir)
+
+    def get_nonexistant_file_resource(self, path):
+        contents = dedent("""text = Arise,awake and do not stop until the goal is reached.""")
```
mathjazz: Please add space after comma.

aniruddha2000:
There is still some tests are failing and i am fixing that.

pontoon/sync/tests/formats/test_ftl.py
aniruddha2000: As, [__init__](https://github.com/mozilla/pontoon/blob/master/pontoon/sync/tests/formats/__init__.py#L363) i think this is getting the `key` attribute from the [test_compare_locale](https://github.com/mozilla/pontoon/blob/master/pontoon/sync/tests/formats/test_compare_locales.py#L241) `name` property of this XML

But when i am writing test for the FTL format [test_ftl](https://github.com/aniruddha2000/pontoon/blob/bug-1521820/pontoon/sync/tests/formats/test_ftl.py#L226) the FTL variable name is `Missing-String` but in [__init__](https://github.com/mozilla/pontoon/blob/master/pontoon/sync/tests/formats/__init__.py#L363) case the key is getting from `Missing String` so the `missing_translation` variable will not valid and [this-line](https://github.com/mozilla/pontoon/blob/master/pontoon/sync/tests/formats/__init__.py#L364) will show an error `AttributeError: 'NoneType' object has no attribute 'strings'` 

If i change [this](https://github.com/mozilla/pontoon/blob/master/pontoon/sync/tests/formats/__init__.py#L363) `key` attribute to `Missing-String` then `test_compare_locale` will show an error `AttributeError: 'NoneType' object has no attribute 'strings'` and i also can't give a space in between the FTL variable.
That's I am facing the main problem.

mathjazz:
@adngdb r?

pontoon/sync/tests/formats/test_ftl.py
```diff
@@ -0,0 +1,291 @@
...
+    def setUp(self):
+        super(FTLResourceTests, self).setUp()
+        self.tempdir = tempfile.mkdtemp()
+
+    def tearDown(self):
+        super(FTLResourceTests, self).tearDown()
+        shutil.rmtree(self.tempdir)
+
+    def get_nonexistant_file_resource(self, path):
+        contents = dedent("""text = Arise, awake and do not stop until the goal is reached.""")
```
adngdb: No need to use `"""` here, simple quotes will suffice. And `dedent` is likewise useless.

pontoon/sync/tests/formats/test_ftl.py
```diff
@@ -0,0 +1,291 @@
...
+        save.
+        """
+        path = self.get_nonexistant_file_path()
+        translated_resource = self.get_nonexistant_file_resource(path)
+
+        translated_resource.translations[0].strings = {
+            None: 'New Translated String'
+        }
+        translated_resource.save(LocaleFactory.create())
+        assert_true(os.path.exists(path))
```
adngdb: Just to prove that this test does what it's supposed to do, I would `assert_false` this before the call to `save`.

pontoon/sync/tests/formats/test_ftl.py
```diff
@@ -0,0 +1,291 @@
...
+
+    def test_parse_with_source_path(self):
+        contents = dedent("""text = Arise, awake and do not stop until the goal is reached.""")
```
adngdb: No need to triple quotes nor `dedent`.

pontoon/sync/tests/formats/test_ftl.py
```diff
@@ -0,0 +1,291 @@
...
+        )
+        path = self.get_nonexistant_file_path()
+        obj = ftl.parse(path, source_path=source_path, locale=None)
+        assert_equal(obj.path, path)
+        assert_equal(obj.locale, None)
+        assert_equal(obj.source_resource.path, source_path)
+        assert_equal(obj.source_resource.locale, None)
+
+    def test_parse_with_no_source_path(self):
+        contents = dedent("""text = Arise, awake and do not stop until the goal is reached.""")
```
adngdb: No need to triple quotes nor `dedent`.

pontoon/sync/tests/formats/test_ftl.py
```diff
@@ -0,0 +1,291 @@
...
+
+BASE_FTL_FILE = """
+# Sample comment
+SourceString = Translated String
+
+# First comment
+# Second comment
+MultipleComments = Translated Multiple Comments
+
+NoCommentsorSources = Translated No Comments or Sources
```
adngdb: What does `Commentsor` mean? :wink:  (I'd go for `NoCommentsOrSources` instead. )

pontoon/sync/tests/formats/test_ftl.py
```diff
@@ -0,0 +1,291 @@
...
+            key=self.key('Source String'),
+            strings={None: 'SourceString = Translated String\n'},
+            fuzzy=False,
+            order=translation_index,
+        )
+
+    def test_parse_multiple_comments(self):
+        input_string = BASE_FTL_FILE
+        translation_index = 1
+        path, resource = self.parse_string(input_string)
```
adngdb: Ouch, that hurts a bit. I'm not saying you should change it here, but this is a piece of complex code, that is very hard to understand without already knowing the code base. It depends on the structure of `FormatTestsMixin` a lot, and hides the function that is actually being tested (in this case, `ftl.parse`). To me this is a complexity that we should avoid in test (and in code in general) if possible, even if that means repeating code, in favor of making tests as straightforward as possible. Anyway, just ranting. :grin:

pontoon/sync/tests/formats/test_ftl.py
```diff
@@ -0,0 +1,291 @@
...
+            source_string=input_string,
+            expected_translation=expected_translation,
+        )
+
+    def test_save_remove(self):
+        """Deleting strings removes them completely from the FTL file."""
+        input_string = dedent("""
+            Source-String = Source String
+        """)
+        expected_string = dedent("""""")
```
adngdb: These 2 strings can easily fit on a single line, removing the need for triple quotes and `dedent`.

pontoon/sync/tests/formats/test_ftl.py
mathjazz: Good point. We'll likely migrate most of the formats to `compare-locales`, which would be a good opportunity to make this code more explicit.

PR #1219, opened 2019-03-05
Bug 1492166 - Mark placeables in translations. 

adngdb:
I'm going to pause here for now, and do the finishing bits in a follow-up PR. Remaining work is to add the placeables to the entity in the Metadata component, make those placeables clickable and copy content into the Editor on click.

@mathjazz sadly once again the commits might not be very useful, I've working on this in a weird way, discovering issues after the fact. And I also made some mistakes with my commits which prevent me from reordering them coherently. Sorry. :/


frontend/src/core/placeable/components/WithPlaceables.js
```diff
@@ -0,0 +1,84 @@
...
+
+    // The spaces placeable can match '\n  ' and mask the newline,
+    // so it has to come later.
+    unusualSpace,
+    nonBreakingSpace,
+    narrowNonBreakingSpace,
+    thinSpace,
+    multipleSpaces,
+
+
```
mathjazz: Nit: you can remove the 2nd blank line.

adngdb:
>     1. Have you considered making the "placeable" (or should we call it "placeables"?) module an external library, like `react-content-marker`? That would be nice.

Yes, at first, but then I wondered if that would really be of any use to anyone else? I assumed those would be pretty specific to us, but maybe not so much. Also, I wonder how many tools there are out there that would benefit from this? If you have a strong opinion in favor of extracting it to a library, I can do that at the same time that I do react-content-marker.

>     2. In addition to string list and original string panel, we also need to markup placeables in the History, Machinery and Locales tabs. However, that can be done as part of the followup you mentioned above.

Yes, I'll do that in the follow-up, but thanks for reminding me! :)

>     3. I was able to reproduce all 4 issues collected in [bug 1251186](https://bugzilla.mozilla.org/show_bug.cgi?id=1251186). Which is good, because that means the logic works consistently with Translate.Current. I wonder however, if it's better to fix these now or later.

I think we should simply make [bug 1251186](https://bugzilla.mozilla.org/show_bug.cgi?id=1251186) part of Translate.Next as a P3 or P4. I like the idea of being able to show users that Translate.Next does help us in fixing more bugs, and showing that the new version has benefits over the current one. What do you think?

PR #1221, opened 2019-03-08
bug 1533703, update compare-locales to 6.0

Pike:
Let's pick up more Fluent checks to run on Pontoon.

In particular, this will prevent localizers to add new Variant Lists.


requirements/constraints.txt
```diff
@@ -167,9 +167,10 @@ nltk==3.2.5 \
 requests==2.21.0 \
     --hash=sha256:7bf2a778576d825600030a110f3c0e3e8edc51dfaafe1c146e39a2027784957b
 
-# >=0.10.0,<0.11 compare-locales
-fluent==0.10.0 \
-    --hash=sha256:b5a270e6e4425cdf140ea10d1a7d9c1af713dd00d61b0412f05cd22e1ff84324
+# <0.13,>=0.12.0 compare-locales
```
mathjazz: Nit: could you flip the boundaries to match order in other requirements?
`>=0.12.0,<0.13`

Pike:
That's weird, it should show an error, and it does for me when I run compare-locales locally

requirements/constraints.txt
Pike: Done.

PR #1224, opened 2019-03-11
Fix bug 1486506: Implement user menu

mathjazz:
I'll file a followup bug for upload/download.


frontend/src/core/user/components/UserMenu.js
```diff
@@ -0,0 +1,208 @@
...
+import * as React from 'react';
+import onClickOutside from 'react-onclickoutside';
+import { Localized } from 'fluent-react';
+
+import './UserMenu.css';
+
+import SignOut from './SignOut';
+
+import type { NavigationParams } from 'core/navigation';
+import type { UserState } from 'core/user';
```
adngdb: In imports, I usually put local imports after module imports, so those should be one block up. I'm not sure I have written those rules down anywhere, if I haven't, I should. Feel free to remind me to do so. :stuck_out_tongue:

frontend/src/core/user/components/UserMenu.js
```diff
@@ -0,0 +1,208 @@
...
+    render() {
+        const { parameters, signOut, user } = this.props;
+
+        return <div className="user-menu">
+            <div
+                className="button selector"
+                onClick={ this.toggleVisibility }
+            >
+                { user.isAuthenticated ?
+                <img src={ user.gravatarURLSmall } alt="" /> :
```
adngdb: Shouldn't we set a value to `alt` here? Something like `Avatar of user ${username}`?

frontend/src/core/user/components/UserMenu.js
```diff
@@ -0,0 +1,208 @@
...
+                <div className="menu-icon fa fa-bars" />
+                }
+            </div>
+
+            { !this.state.visible ? null :
+            <div className="menu"><ul>
```
adngdb: Is the `<div>` here useful? Can you not just use `<ul className="menu">` instead of having that double tag?

frontend/src/core/user/components/UserMenu.js
```diff
@@ -0,0 +1,208 @@
...
+                <li className="details">
+                    <a href="/profile/">
+                        <img src={ user.gravatarURLBig } alt="User avatar" />
+                        <p className="name">{ user.nameOrEmail }</p>
+                        <p className="email">{ user.email }</p>
+                    </a>
+                </li>
+                }
+
+                { !user.isAuthenticated ? null :
```
adngdb: Why not merge that with the previous block?

frontend/src/core/user/components/UserMenu.js
```diff
@@ -0,0 +1,208 @@
...
+                }
+
+                <li>
+                    <Localized
+                        id="user-UserMenu--download-tm"
+                        glyph={
+                            <i className="fa fa-cloud-download-alt fa-fw"></i>
+                        }
+                    >
+                        <a href={ `/${parameters.locale}/${parameters.project}/${parameters.locale}.${parameters.project}.tmx` }>
```
adngdb: For the sake of shorter lines, I would move the content of the `href` attribute to a variable.

frontend/src/core/user/components/UserMenu.js
```diff
@@ -0,0 +1,208 @@
...
+                        }
+                    >
+                        <a href="/admin/">
+                            { '<glyph></glyph>Admin' }
+                        </a>
+                    </Localized>
+                </li>
+                }
+
+                { !user.isAdmin ? null :
```
adngdb: Same here, why not have those two be part of the same block since the conditions are the same?

frontend/src/core/user/components/UserMenu.js
```diff
@@ -0,0 +1,208 @@
...
+                        }
+                    >
+                        <a href="/settings/">
+                            { '<glyph></glyph>Settings' }
+                        </a>
+                    </Localized>
+                </li>
+                }
+
+                { !user.isAuthenticated ? null :
```
adngdb: And here again. :)

frontend/src/core/user/components/UserMenu.test.js
```diff
@@ -0,0 +1,67 @@
...
+            <UserMenuBase
+                user={ { isAuthenticated: true } }
+                parameters={ { locale: 'locale', project: 'project' } }
+            />
+        );
+        wrapper.instance().setState({visible: true});
+
+        expect(wrapper.find('.details')).toHaveLength(1);
+        expect(wrapper.find('a[href="/settings/"]')).toHaveLength(1);
+        expect(wrapper.find(SignOut)).toHaveLength(1);
```
adngdb: One thing you can use in tests is the l10n ids. That easily allows you to check if a certain string is exposed. Not that it is needed here, but I wanted to make sure you had that tool in your belt. :)

pontoon/base/views.py
```diff
@@ -979,9 +979,11 @@ def user_data(request):
 
     return JsonResponse({
         'is_authenticated': True,
+        'is_admin': user.has_perm('base.can_manage_project'),
```
adngdb: Shouldn't that be true only if the user is a superuser? There are non-superusers who can access the admin interface?

frontend/src/core/user/components/UserMenu.js
mathjazz: Oh, that's OK. For some reason I thought `type` imports are an exception and always come in last.

pontoon/base/views.py
mathjazz: Yes. Superusers can also access Django Admin, which gives users more power, among other things to delete stuff. We have external PMs (for projects like MDN), which can only access Pontoon Admin.

Still, this is confusing, and hence we have bug 1203788 & 1204659.

frontend/src/core/user/components/UserMenu.test.js
mathjazz: Thanks!

However, I'm trying this and it fails:
```
expect(wrapper.find('[data-l10n-id="user-UserMenu--settings"]')).toHaveLength(1);
```

frontend/src/core/user/components/UserMenu.js
mathjazz: I got `Adjacent JSX elements Parse Error`, but I see I can fix it by using `React.Fragment`. Thanks!

frontend/src/core/user/components/UserMenu.js
adngdb: Damn, sorry, I didn't see those were type imports... You were absolutely correct. :-(

mathjazz: NP, I'll fix it in https://bugzilla.mozilla.org/show_bug.cgi?id=1534708.

frontend/src/core/user/components/UserMenu.js
adngdb: Ah, I should have guessed so. You can either use `React.Fragment` indeed (I've been avoiding the form `<> ... </>` because I find it highly confusing, maybe once it becomes more spread in the React community we could revisit) or you can keep it this way, I think that's fine too as it removes a level of indentation.

frontend/src/core/user/components/UserMenu.test.js
adngdb: Here's an example of how to do it: https://github.com/mozilla/testpilot/blob/master/frontend/test/app/util.js#L4

I haven't yet used that, and so I haven't imported those functions into Pontoon, but I guess we'll do it at some point in the future.

PR #1226, opened 2019-03-13
[WIP] Fix bug 1534708: File download/upload

mathjazz:
Downloading a file directly from an Ajax request is not possible ATM. The workaround is to add a hidden form and submit it behind the scenes to get the browser to trigger the Save dialog.

@adngdb What are your thoughts on the following implementation? It follows Translate.Prod pretty closely, so I suspect it's wrong. :) I'd like to get the design sorted out early though, because we'll need something similar for file upload. And possibly even for Sign Out (I'm not sure using AJAX is the right approach for it).


frontend/src/core/user/components/FileDownload.js
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    render() {
+        const { parameters } = this.props;
+
+        let csrfToken = '';
+        const rootElt = document.getElementById('root');
+        if (rootElt) {
+            csrfToken = rootElt.dataset.csrfToken;
+        }
```
mathjazz: We should factor this out and reuse in `core/api/base.js`.

frontend/src/core/user/components/FileDownload.js
```diff
@@ -0,0 +1,56 @@
...
+    parameters: NavigationParams,
+|};
+
+
+/*
+ * Render a Sign Out link.
+ */
+export default class SignOut extends React.Component<Props> {
+    downloadTranslations = () => {
+        document.getElementById('download-form').submit();
```
mathjazz: I'm getting an interesting `flow` error here:
```
Cannot call document.getElementById(...).submit because:
 • property submit is missing in HTMLElement [1].
 • property submit is missing in null [2].
```

adngdb: Yeah, Flow isn't working very well with DOM elements. You can either work around the error but checking that what's returned by `getElementById` is of the correct type (but meeeh) or simply ignore this error by putting `// $FLOW_IGNORE` on the previous line, ideally with a short comment explaining why we ignore this (in this case, we are pretty sure the element will exist).

frontend/src/core/user/components/FileDownload.js
adngdb: There's a `getCSRF` method in api.base maybe you can just use that? Or maybe that's what you meant and I'm being a fool? :clown_face:

frontend/src/core/user/components/FileDownload.js
adngdb: Actually, you should use `refs` in React in order to reach a DOM element of your component. See https://reactjs.org/docs/refs-and-the-dom.html

adngdb: You could also put your `button` inside the `form` and avoid this function entirely.

frontend/src/core/user/components/FileDownload.js
```diff
@@ -0,0 +1,56 @@
...
+                <button onClick={ this.downloadTranslations }>
+                    { '<glyph></glyph>Download Translations' }
+                </button>
+            </Localized>
+
+            <form
+                id="download-form"
+                action="/download/"
+                className="hidden"
+                method="POST"
```
adngdb: Why does the view `download` require `POST`? It could be so much better if it was a get, then this could simply be a link. Browsers nowadays don't hide the page you where in if the target of the link is a download.

frontend/src/core/user/components/FileUpload.js
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    render() {
+        const { parameters } = this.props;
+
+        let csrfToken = '';
+        const rootElt = document.getElementById('root');
+        if (rootElt) {
+            csrfToken = rootElt.dataset.csrfToken;
+        }
```
mathjazz: I wasn't able to figure out how to re-use `getCSRFToken` from `core.api.base` here. Help?

frontend/src/core/user/components/UserMenu.js
```diff
@@ -49,11 +50,29 @@ export class UserMenuBase extends React.Component<Props, State> {
...
+        const transHref = `/download/?code=${locale}&slug=${project}&part=${resource}`;
+
+        const canDownload = (
+            project !== 'all-projects' &&
+            resource !== 'all-resources'
+        );
+        const canReview = (
+            user.translatorForLocales &&
+            user.translatorForLocales.includes(locale)
+        );
```
mathjazz: This is taken from the [`History`](https://github.com/mozilla/pontoon/blob/master/frontend/src/modules/history/components/History.js#L60) component, but it's not really accurate: it doesn't take into account locales the user is a manager for (`managerForLocales`) and per-project translator permissions (`translatorForProjects`). See how [Translate.Current](https://github.com/mozilla/pontoon/blob/master/pontoon/base/static/js/translate.js#L4553) handles that.

We should amend `canReview` accordingly, and possibly store the information in `UserState` for easier reuse.

My question is, should we already on backend generate a list of (locale_code, project_slug) pairs and pass that to frontend? We'd do that using `managerForLocales`, `translatorForLocales` and `translatorForProjects`, which are all passed to frontend, but never used beyond the "`canReview`" use case.

frontend/src/core/user/components/FileUpload.js
mathjazz: Added a comment.

frontend/src/core/user/components/FileUpload.js
```diff
@@ -0,0 +1,70 @@
...
+            <input name="slug" type="hidden" value={ parameters.project } />
+            <input name="part" type="hidden" value={ parameters.resource } />
+            <label>
+                <Localized
+                    id="user-UserMenu--upload-translations"
+                    glyph={
+                        <i className="fa fa-cloud-upload-alt fa-fw"></i>
+                    }
+                >
+                    <span>{ '<glyph></glyph>Upload Translations' }</span>
```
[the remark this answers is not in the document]
mathjazz: I believe it's needed. If I remove it, `<glyph></glyph>` doesn't render. Other `Localized` elements using `<glyph>` also wrap an element, not the text node directly. I also see the `<span>` in the DOM.

frontend/src/core/user/components/UserMenu.js
mathjazz: Added the [selector](https://github.com/mozilla/pontoon/pull/1226/commits/3c4cbd77d30ee3aac23e214effa1fe841ce958ba).

frontend/src/core/user/selectors.js
```diff
@@ -0,0 +1,53 @@
...
+    if (!user.isAuthenticated) {
+        return false;
+    }
+
+    if (user.managerForLocales.indexOf(locale) !== -1) {
+        return true;
+    }
+
+    if (user.translatorForProjects.hasOwnProperty(localeProject)) {
+        return user.translatorForProjects[localeProject];
```
mathjazz: I'm getting a flow error here that I can't fix:
`Cannot get user.translatorForProjects[localeProject] because string [1] is not an array index.`

frontend/src/core/user/components/FileUpload.js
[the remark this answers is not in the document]
adngdb: Isn't that because the translation was missing in the `.ftl` file? I'm not very clear how fluent-react works internally but I don't understand why it would keep that span in the content. Discard my comment if you think it's not useful.

frontend/src/core/user/selectors.js
adngdb: `user.translatorForProjects` is defined as an `Array<string>` in `core/user/reducer.js`. Hence this code shouldn't work, because that array is not supposed to have your `localeProject` property, nor can it be accessed like an object. I'm not sure which one is wrong, your code or the type declaration, but you'll have to change one of the 2.

frontend/src/core/user/components/FileUpload.js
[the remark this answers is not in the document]
mathjazz: Nope, it's in there already:
https://github.com/mozilla/pontoon/pull/1226/files#diff-141010cc4b360add776bdcffd473a577R160

mathjazz:
And the tests are passing!!! @adngdb One more look?

frontend/src/core/user/components/UserMenu.test.js
```diff
@@ -58,6 +61,75 @@ describe('<UserMenuBase>', () => {
...
     });
 
+    it('shows upload & download menu items', () => {
+        const wrapper = shallow(
+            <UserMenuBase
+                isTranslator={ true }
+                user={ { isAuthenticated: true } }
+                parameters={ { locale: 'mylocale', project: 'myproject', resource: 'myresource' } }
+            />
+        );
```
adngdb: I'd recommend you refactor this into a function at the start of the file, and replace all instances of the same line in your tests. 

Here's an example of such a function, with default values for parameters, and to which you can pass only some of the params: https://github.com/mozilla/pontoon/blob/master/frontend/src/modules/editor/components/Editor.test.js#L23

frontend/src/core/user/components/UserMenu.test.js
```diff
@@ -58,6 +61,75 @@ describe('<UserMenuBase>', () => {
...
+                parameters={ { locale: 'mylocale', project: 'myproject', resource: 'all-resources' } }
+            />
+        );
+        wrapper.instance().setState({visible: true});
+
+        expect(wrapper.find(FileUpload)).toHaveLength(0);
+        expect(findLocalizedById(wrapper, 'user-UserMenu--download-translations')).toHaveLength(0);
+    });
+
+    it('hides upload menu item fors users without permission to review translations', () => {
```
adngdb: s/fors/for

frontend/src/core/user/components/UserMenu.test.js
```diff
@@ -58,6 +61,75 @@ describe('<UserMenuBase>', () => {
...
+                isReadOnly={ true }
+                isTranslator={ true }
+                user={ { isAuthenticated: true } }
+                parameters={ { locale: 'mylocale', project: 'myproject', resource: 'myresource' } }
+            />
+        );
+        wrapper.instance().setState({visible: true});
+
+        expect(wrapper.find(FileUpload)).toHaveLength(0);
+    });
```
adngdb: Those are excellent tests! :sun_with_face:

frontend/src/core/user/reducer.js
```diff
@@ -55,7 +55,7 @@ export type UserState = {|
     +username: string,
     +managerForLocales: Array<string>,
     +translatorForLocales: Array<string>,
-    +translatorForProjects: Array<string>,
+    +translatorForProjects: Object,
```
adngdb: Can you be more precise in this declaration? Based on what you told me over IRC, I think this should be `{ [string]: boolean }` (see documentation: https://flow.org/en/docs/types/objects/#toc-objects-as-maps ).

frontend/src/core/user/selectors.test.js
```diff
@@ -0,0 +1,44 @@
...
+        ).toEqual(false);
+    });
+
+    it('returns true if user is a manager of the locale', () => {
+        expect(
+            _isTranslator(
+                { managerForLocales: ['mylocale'] },
+                { locale: 'mylocale', project: 'myproject' }
+            )
+        ).toEqual(false);
```
adngdb: This test doesn't do what it says it should do. :-) If I understand it correctly, it should have `isAuthenticated: true` in the user part, and assert that the results is true. Btw, you can use `toBeTruthy()` or `toBeFalsy()` here.

Side note, the more I use Jest's assertion API the less I like it. At some point I think I'll introduce [Chai](https://www.chaijs.com/) in Pontoon and switch us to that if you don't mind.

frontend/src/core/user/selectors.test.js
```diff
@@ -0,0 +1,44 @@
...
+    });
+
+    it('returns true if user is a translator of the locale', () => {
+        expect(
+            _isTranslator(
+                { translatorForLocales: ['mylocale'] },
+                { locale: 'mylocale', project: 'myproject' }
+            )
+        ).toEqual(false);
```
adngdb: Same here and for the test below.

frontend/src/core/user/selectors.test.js
mathjazz: Ouch!

mathjazz:
Thanks for a great review! I've pushed updates according to your comments.

frontend/src/core/user/components/UserMenu.test.js
```diff
@@ -8,38 +8,43 @@ import { UserMenuBase } from './UserMenu';
...
+    locale = 'mylocale',
+    project = 'myproject',
+    resource = 'myresource',
+} = {}) {
+    return shallow(
+        <UserMenuBase
+            isReadOnly={ isReadOnly }
+            isTranslator={ isTranslator }
+            user={ { isAuthenticated: isAuthenticated, isAdmin: isAdmin } }
+            parameters={ { locale: locale, project: project, resource: resource } }
```
adngdb: There's a nice trick with ES6: if the name of the key and the name of the variable are the same, you can shorten the syntax with just the key. For example, in this case, you could write: `{ locale, project, resource }`.

mathjazz: Oh, that's cool! Thanks for the pro-tip!

PR #1233, opened 2019-03-21
Fixes bug 1492166 - Add placeables to Metadata and helper tools. 

adngdb:
@mathjazz the commits should be a lot more helpful the time! :-)

Note that there's one remaining task, and that is to add placeables to the helper tabs. If it is trivial for Locales, I am not sure yet how I can combine it with `TranslationDiff` in History and Machinery.


mathjazz:
> If it is trivial for Locales, I am not sure yet how I can combine it with `TranslationDiff` in History and Machinery.

Yup. You might find Translate.Prod useful for that, because we had the same problem there.

frontend/src/modules/editor/components/Editor.js
```diff
@@ -55,6 +55,10 @@ export class EditorBase extends React.Component<InternalProps, State> {
     constructor(props: InternalProps) {
         super(props);
 
+        // This state acts as a buffer between the Editors and the editor state.
+        // We want to avoid updating the redux state at every key stroke, thus
+        // we keep track of changes in this state and only update the store
+        // every few milliseconds using debouce.
```
mathjazz: What's the benefit of updating every few milliseconds vs. updating at every key stroke?

frontend/src/modules/editor/components/FluentEditor.js
```diff
@@ -36,6 +36,35 @@ function annotate(source) {
...
+        // the content differently for each Editor type. Thus each Editor
+        // implements an `updateTranslationSelectionWith` method and that is
+        // used to update the translation.
+        if (this.props.editor.selectionReplacementContent) {
+            this.updateTranslationSelectionWith(
+                this.props.editor.selectionReplacementContent
+            );
+            this.props.resetSelectionContent();
+        }
+    }
```
mathjazz: Since the content of this is the same as in `GenericEditor.js`, I wonder if we can factor it out and reuse somehow? Or is that somehow related to the "redux hack"?

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -66,7 +66,14 @@ export default class Metadata extends React.Component<Props> {
...
             // $FLOW_IGNORE
-            if (e.target.childNodes.length) {
+            if (e.target.dataset['match']) {
+                this.props.addTextToEditorTranslation(
+                    // $FLOW_IGNORE
+                    e.target.dataset['match']
+                );
+            }
+            // $FLOW_IGNORE
+            else if (e.target.childNodes.length) {
```
mathjazz: I wonder if it'd be cleaner to require `data-match` attribute for all placeable types and drop this `else`.

frontend/src/modules/editor/components/Editor.js
adngdb: Users can type very very fast, and for every key stroke, we would trigger a React render. That can completely boast the UI and make the user experience horrible. I have no idea what the correct time here should be, but I'm sure it's better to have something rather than just let the events roam free.

frontend/src/modules/editor/components/FluentEditor.js
adngdb: I've tried to do that but failed. This isn't a case where Higher Order Components work, it requires a class hierarchy, and I'm not too happy about extending classes. Though in this case that might just be the proper solution.

frontend/src/modules/entitydetails/components/Metadata.js
adngdb: That's kinda useless on most of the parsers. I'd rather make the API as clean as possible, and make the code underneath a bit more clunky in such a case. If you feel strongly about it though, I'm open to changing it.

adngdb: It also adds duplicated to the DOM, not sure how much that costs but I'm thinking we might as well stay as slim as we can (says the guy who uses React... ).

adngdb:
@mathjazz re-requesting review because my last changes are not trivial.

frontend/src/modules/machinery/components/Translation.test.js
```diff
@@ -37,8 +37,8 @@ describe('<Translation>', () => {
             entity={ DEFAULT_ENTITY }
         />);
 
-        expect(wrapper.find('.original').find(TranslationDiff)).toHaveLength(1);
```
mathjazz: You'll need to remove TranslationDiff import.

adngdb:
@mathjazz the bug should be fixed now. Mind verifying?

frontend/src/core/placeable/components/WithPlaceablesNoUnusualSpace.js
```diff
@@ -0,0 +1,21 @@
...
+    let newRules = [ ...rules ];
+    newRules.splice(newRules.indexOf(unusualSpace), 1);
+    return newRules;
+}
+
+
+const WithPlaceablesNoUnusualSpace = createMarker(getRulesWithoutUnusualSpace(rules));
+
+
+export default WithPlaceablesNoUnusualSpace;
```
mathjazz: A comment explaining motivation for this component would be nice.

adngdb:
@mathjazz well I made a very stupid mistake &mdash; and yesterday I thought it was a problem with docker... Glad I slept on it, it took me half a minute of thinking to figure out what I did wrong this morning.

Anyway, I also figured I should split out the leading space part of `unusualSpace` and remove only that from the parsers list.

The problem with sending translations for a missing string is out of scope I believe, I'm on it.

frontend/src/core/placeable/components/WithPlaceablesNoLeadingSpace.js
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Component that marks placeables in a string. Same as WithPlaceables but
+ * without the leadingSpace parser.
+ *
+ * The leadingSpace parser checks for spaces at the beginning of a line.
+ * If the input was previously split, then that parser will generate a lot of
+ * false positive (for example, if something else was marked just before a
+ * space, it thus becomes a space the beginning of a line). We thus want to
+ * have a special Placeables component without that parser, for use in
+ * combination with other parsing tools (like diff).
```
mathjazz: Nice!

PR #1236, opened 2019-03-25
Fix bug 1536670 - Update six to 1.12.0

jotes:
Hey,

I updated `six` to 1.12.0 and replaced all missing methods with their equivalents.

@adngdb do you have a moment to look at this?

### DIFF
--- a/pontoon/checks/libraries/pontoon.py
+++ b/pontoon/checks/libraries/pontoon.py
@@ -35,7 +35,7 @@ def run_checks(entity, string):
     Group all checks related to the base UI
 
     :arg pontoon.base.models.Entity entity: Source entity
-    :arg six.string_types string: a translation
+    :arg basestring string: a translation
     """
     checks = defaultdict(list)
     resource_ext = entity.resource.format
--- a/pontoon/checks/libraries/pontoon.py
+++ b/pontoon/checks/libraries/pontoon.py
@@ -1,6 +1,5 @@
 from __future__ import absolute_import
 
-import HTMLParser
 import re
 
 import bleach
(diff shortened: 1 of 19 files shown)


pontoon/checks/libraries/pontoon.py
```diff
@@ -33,7 +35,7 @@ def run_checks(entity, string):
     Group all checks related to the base UI
 
     :arg pontoon.base.models.Entity entity: Source entity
-    :arg basestring string: a translation
+    :arg six.string_types string: a translation
```
jotes: @adngdb Do you prefer the previous version?

adngdb: I think it would better to be less specific here. Hopefully once we're moved to py3 we should remove `six`, right?

pontoon/localizations/views.py
```diff
@@ -89,7 +89,7 @@ def ajax_resources(request, code, slug):
         pages[page.name] = latest_page_translatedresource
 
     translatedresources = {s.resource.path: s for s in translatedresources_qs}
-    translatedresources = dict(translatedresources.items() + pages.items())
+    translatedresources = dict(list(translatedresources.items()) + pages.items())
```
adngdb: I've seen a few of those `list()` added. Why is that needed?

Pike: You can't concat `dict_items`. Interestingly enough, this is not enough, both `items` need to be converted.

Or you create a dict from `translatedresources` first, and then update from `pages`.

adngdb: Ah, `.keys()` doesn't return a simple `list`? I didn't know that. Thanks!

jotes: @Pike Thanks :) Added to my todo list.

pontoon/checks/libraries/pontoon.py
jotes: That's right, I'll revert it then.

PR #1238, opened 2019-03-26
Fix bug 1527291 - Redirect to first entity when none is selected.

adngdb:


adngdb:
Actually, that one will still show up when the entity ID is incorrect or missing. That will happen when users follow links, for example, on an entity that is not in the first batch of entities. It's a whole different problem to solve, and I didn't do it here. I've filed [bug 1539228](https://bugzilla.mozilla.org/show_bug.cgi?id=1539228) for that.

PR #1240, opened 2019-03-28
Bug 1525984 - Add basic keyboard shortcuts for the Editor.

adngdb:
This is part one of bug 1525984. It adds three shortcuts for the Editor: send translation on Enter, copy original on Ctrl + Shift + C and clear on Ctrl + Shift + Backspace.

@mathjazz, I'm going to do this in parts, because it will make it easier for you to review, and also those shortcuts aren't really tied to each others.


adngdb:
Yes, I intend to do the Fluent Editor later on. It will work differently, and require different shortcuts too I think (it doesn't make sense to have Enter send the translation inside a multi-line editor).

mathjazz:
Good point!

PR #1241, opened 2019-03-28
Remove Select an entity note

mathjazz:
Let's remove this unstyled block of text for good.


adngdb:
Haha, you really do hate it, do you? ;-)

PR #1242, opened 2019-03-29
No bug - Update frontend dependencies

adngdb:
This also upgrades our Docker image to use a more recent Debian and Python.


adngdb:
Forgot to add testing steps, here they are: 

1. `make clear`
2. `make build`
3. `make test`

If everything passes then this should be good to go.

PR #1244, opened 2019-03-30
Bug 1429392 - Textarea More Explicitly Locked for Non-Auth Users

Neecolaa:
Changes:
* Added different styling to textarea for non-authenticated users. (translate.css)
* Changed placeholder text to clearly state what is required next. (translate.html)

The textarea in question already was already readonly and disabled focus, this pull request just adds in the formatting changes that were also requested when the bug report was created. 

Rather than removing the placeholder text, the placeholder text now also directs the user to sign in, in case the link below was overlooked.

Testing:
Viewed and interacted with textarea both as an authenticated and non-authenticated user.

### DIFF
--- a/pontoon/base/templates/translate.html
+++ b/pontoon/base/templates/translate.html
@@ -310,7 +310,6 @@
           placeholder="Type translation and press Enter to save"
         {% else %}
           readonly
-          placeholder="Sign in below to begin translating."
         {% endif %}
         ></textarea>
 
@@ -324,7 +323,6 @@
                   placeholder="Type translation and press Enter to save"
                 {% else %}
                   readonly
-                  placeholder="Sign in below to begin translating."
                 {% endif %}
               ></textarea>
             </li>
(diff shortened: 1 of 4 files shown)


pontoon/base/templates/translate.html
```diff
@@ -304,7 +304,8 @@ <h2></h2>
         </ul>
       </nav>
 
-      <textarea {% if not user.is_authenticated() %} readonly {% endif %} id="translation" placeholder="Type translation and press Enter to save"></textarea>
+      <textarea {% if user.is_authenticated() %} id="translation" placeholder="Type translation and press Enter to save" 
+      {% else %} id="translation-no-login"  placeholder="Sign in below to begin translating."  {% endif %}></textarea>
```
mathjazz: Please make sure that you keep the `id` unchanged and that you don't remove the `readonly attribute`.

In CSS you'd then reference editor for non-authenticated users with `#editor textarea[readonly] {`, which will also allow you to only change the background color (and remove `(min-)height`).

Note that there's another `textarea` below (line 314), where you also need to change the placeholder text based on `is_authenticated` status.

Neecolaa: Thank you for the corrections! I've made and committed your suggested changes.

PR #1247, opened 2019-04-03
Fix bug 1541576: Support unicode characters in XML checks

mathjazz:
@adngdb and @jotes r?


mathjazz:
r+ via @jotes via IRC

PR #1248, opened 2019-04-04
Fix bug 1529491 - Do not copy translation when link or button is clicked

adngdb:


adngdb:
Good catch Matjaz! I've added those.

PR #1249, opened 2019-04-04
Fix bug 1524549: Add support for custom search in Machinery

mathjazz:
This gets the job done. (With the exception of changing the source string in machinery diff, which I'm not sure how to take care of yet.)

Before I dig into it, I wonder if I should refactor the machinery API, because I'm using a couple of ugly hacks to make this patch work. See the question below.

### DIFF
--- a/frontend/src/modules/entitydetails/components/EntityDetails.js
+++ b/frontend/src/modules/entitydetails/components/EntityDetails.js
@@ -93,22 +93,23 @@ export class EntityDetailsBase extends React.Component<InternalProps, State> {
         }
 
         dispatch(history.actions.get(parameters.entity, parameters.locale, pluralForm));
-        dispatch(machinery.actions.get(selectedEntity, locale));
+        dispatch(machinery.actions.get(selectedEntity.original, locale, selectedEntity.pk));
         dispatch(otherlocales.actions.get(parameters.entity, parameters.locale));
     }
 
     searchMachinery = (query: string) => {
         const { dispatch, locale, selectedEntity } = this.props;
 
-        // Deep copy entity to avoid mutating state
-        const entity = JSON.parse(JSON.stringify(selectedEntity));
+        let source = query;
+        let pk = null;
 
         // On empty query, use source string as input
-        if (entity && query.length) {
-            entity.original = query;
+        if (selectedEntity && !query.length) {
+            source = selectedEntity.original;
+            pk = selectedEntity.pk;
         }
 
-        dispatch(machinery.actions.get(entity, locale));
+        dispatch(machinery.actions.get(source, locale, pk));
     }
 
     goToNextEntity = () => {
--- a/frontend/src/modules/entitydetails/components/EntityDetails.js
+++ b/frontend/src/modules/entitydetails/components/EntityDetails.js
@@ -97,7 +97,7 @@ export class EntityDetailsBase extends React.Component<InternalProps, State> {
         dispatch(otherlocales.actions.get(parameters.entity, parameters.locale));
     }
 
-    customMachinerySearch = (query: string) => {
+    searchMachinery = (query: string) => {
         const { dispatch, locale, selectedEntity } = this.props;
 
         // Deep copy entity to avoid mutating state
@@ -209,7 +209,7 @@ export class EntityDetailsBase extends React.Component<InternalProps, State> {
                 deleteTranslation={ this.deleteTranslation }
                 updateTranslationStatus={ this.updateTranslationStatus }
                 updateEditorTranslation={ this.updateEditorTranslation }
-                customMachinerySearch={ this.customMachinerySearch }
+                searchMachinery={ this.searchMachinery }
             />
         </section>;
     }
--- a/frontend/src/modules/entitydetails/components/EntityDetails.js
+++ b/frontend/src/modules/entitydetails/components/EntityDetails.js
@@ -97,6 +97,11 @@ export class EntityDetailsBase extends React.Component<InternalProps, State> {
         dispatch(otherlocales.actions.get(parameters.entity, parameters.locale));
     }
 
+    fetchMachinery = () => {
+        const { dispatch, locale, selectedEntity } = this.props;
+        dispatch(machinery.actions.get(selectedEntity, locale));
+    }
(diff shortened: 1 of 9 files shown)


frontend/src/modules/entitydetails/components/EntityDetails.js
```diff
@@ -97,6 +97,20 @@ export class EntityDetailsBase extends React.Component<InternalProps, State> {
...
+    customMachinerySearch = (query: string) => {
+        const { dispatch, locale, selectedEntity } = this.props;
+
+        // Deep copy entity to avoid mutating state
+        const entity = JSON.parse(JSON.stringify(selectedEntity));
+
+        // On empty query, use source string as input
+        if (entity && query.length) {
+            entity.original = query;
+        }
```
mathjazz: Obviously, this is ugly. I believe the proper solution here is to refactor the machinery API:
* now: `machinery.actions.get(entity, locale)`
* after refactor: `machinery.actions.get(source, pk, locale)`

frontend/src/modules/entitydetails/components/EntityDetails.js
```diff
@@ -97,6 +97,20 @@ export class EntityDetailsBase extends React.Component<InternalProps, State> {
         dispatch(otherlocales.actions.get(parameters.entity, parameters.locale));
     }
 
+    customMachinerySearch = (query: string) => {
+        const { dispatch, locale, selectedEntity } = this.props;
+
+        // Deep copy entity to avoid mutating state
+        const entity = JSON.parse(JSON.stringify(selectedEntity));
```
adngdb: A more elegant way to do this with ES6 is to use the spread operator: `const entity = { ...selectedEntity };`

frontend/src/modules/entitydetails/components/EntityDetails.js
adngdb: Yes, I entirely agree. It also makes the `get` action more focused, so that's all the better.

frontend/src/modules/entitydetails/components/EntityDetails.js
```diff
@@ -195,6 +209,7 @@ export class EntityDetailsBase extends React.Component<InternalProps, State> {
                 deleteTranslation={ this.deleteTranslation }
                 updateTranslationStatus={ this.updateTranslationStatus }
                 updateEditorTranslation={ this.updateEditorTranslation }
+                customMachinerySearch={ this.customMachinerySearch }
```
adngdb: This sounds like a variable name and not a function. Maybe rename it to something like `searchMachinery`?

frontend/src/modules/machinery/components/Machinery.js
```diff
@@ -46,7 +57,12 @@ export default class Machinery extends React.Component<Props> {
             <div className="search-wrapper clearfix">
                 <div className="icon fa fa-search"></div>
                 <Localized id="machinery-machinery-search-placeholder" attrs={{ placeholder: true }}>
-                    <input type="search" autoComplete="off" placeholder="Type to search machinery" />
+                    <input
+                        type="search"
+                        autoComplete="off"
+                        placeholder="Type to search machinery"
+                        onKeyDown={ this.handleShortcuts }
+                    />
```
adngdb: I suspect this would be cleaner if it was put inside a form and we used the `onSubmit` event of the form to handle the custom search instead. Less hackish, more standard, probably more accessible too.

mathjazz:
@adngdb I've addressed you review comments.

I've also taken the opportunity to convert all Machinery related AJAX responses to JSON to prevent Error messages in the console.

frontend/src/modules/machinery/components/Machinery.js
```diff
@@ -29,9 +28,13 @@ type Props = {|
  * third-party Machine Translation.
  */
 export default class Machinery extends React.Component<Props> {
+    submitForm = (event: SyntheticKeyboardEvent<>) => {
+        event.preventDefault();
+        this.props.searchMachinery(event.currentTarget[0].value);
```
mathjazz: I'm not sure how to address the flow error here. Should I just ignore it?

adngdb: This might be simpler and more robust (especially if we ever change the DOM structure) by using a ref on the input.

pontoon/machinery/views.py
```diff
@@ -90,16 +100,26 @@ def microsoft_translator(request):
...
+        }, status=400)
 
     api_key = settings.MICROSOFT_TRANSLATOR_API_KEY
 
     if not api_key:
         log.error("MICROSOFT_TRANSLATOR_API_KEY not set")
-        return HttpResponseBadRequest("Missing api key.")
+        return JsonResponse({
+            'status': False,
+            'message': 'Bad Request: {error}'.format(error='Missing api key.'),
```
adngdb: Hehe, I guess this can be simplified to a simple string. :wink:

pontoon/machinery/views.py
```diff
@@ -138,16 +164,26 @@ def google_translate(request):
...
+        }, status=400)
 
     api_key = settings.GOOGLE_TRANSLATE_API_KEY
 
     if not api_key:
         log.error('GOOGLE_TRANSLATE_API_KEY not set')
-        return HttpResponseBadRequest('Missing api key.')
+        return JsonResponse({
+            'status': False,
+            'message': 'Bad Request: {error}'.format(error='Missing api key.'),
```
adngdb: Same here. :)

pontoon/machinery/views.py
```diff
@@ -201,21 +243,19 @@ def caighdean(request):
...
-        return http.HttpResponseServerError(
-            json.dumps({
-                'error': 500,
-                'message': 'Caighdean is unavailable offline',
-            }),
-            content_type='application/json',
-        )
+        return JsonResponse({
+            'status': False,
+            'message': 'Server Error: {error}'.format(error='Caighdean is unavailable offline'),
```
adngdb: And here too!

pontoon/machinery/views.py
mathjazz: Yikes!

PR #1251, opened 2019-04-07
Fix bug 1542575: Update Caighdean endpoint by using fork

mathjazz:
Caighdean API endpoint [has moved](https://github.com/translate/python-caighdean/pull/2). Since the PR hasn't been merged upstream yet, we need to switch to a fork, which fixes the endpoint.

https://mozilla-pontoon-staging.herokuapp.com/ga-IE/firefox/browser/branding/official/brand.dtd/?string=74131

### DIFF
--- a/frontend/src/modules/machinery/components/Translation.js
+++ b/frontend/src/modules/machinery/components/Translation.js
@@ -41,6 +41,9 @@ export default class Translation extends React.Component<Props> {
         const { locale, sourceString, translation } = this.props;
 
         const TranslationPlaceablesDiff = withDiff(WithPlaceablesNoLeadingSpace);
+        const types = translation.sources.map(function(source) {
+            return source.type;
+        });
 
         return <Localized id="machinery-translation-copy" attrs={{ title: true }}>
             <li
@@ -78,11 +81,15 @@ export default class Translation extends React.Component<Props> {
                     </ul>
                 </header>
                 <p className="original">
-                    <TranslationPlaceablesDiff
-                        diffTarget={ translation.original }
-                    >
-                        { sourceString }
-                    </TranslationPlaceablesDiff>
+                    { types.indexOf('Caighdean') === -1 ?
+                        <TranslationPlaceablesDiff
+                            diffTarget={ translation.original }
+                        >
+                            { sourceString }
+                        </TranslationPlaceablesDiff>
+                    :
+                        sourceString
+                    }
                 </p>
                 <p
                     className="suggestion"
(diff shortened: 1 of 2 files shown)


frontend/src/modules/machinery/components/Translation.js
```diff
@@ -78,11 +81,15 @@ export default class Translation extends React.Component<Props> {
...
-                    </TranslationPlaceablesDiff>
+                    { types.indexOf('Caighdean') === -1 ?
+                        <TranslationPlaceablesDiff
+                            diffTarget={ translation.original }
+                        >
+                            { sourceString }
+                        </TranslationPlaceablesDiff>
+                    :
+                        sourceString
+                    }
```
mathjazz: This fix is outside the scope of the bug, but it's easier to test when Caighdean actually works.

frontend/src/modules/machinery/components/Translation.js
```diff
@@ -41,6 +41,9 @@ export default class Translation extends React.Component<Props> {
         const { locale, sourceString, translation } = this.props;
 
         const TranslationPlaceablesDiff = withDiff(WithPlaceablesNoLeadingSpace);
+        const types = translation.sources.map(function(source) {
+            return source.type;
```
adngdb: This can be simplified to `source => source.type`.

frontend/src/modules/machinery/components/Translation.js
adngdb: I don't understand, why disable diff and placeables when the source is Caighdean? Can you add a comment explaining that?

requirements/default.txt
```diff
@@ -147,3 +145,6 @@ https://github.com/mathjazz/silme/archive/v0.9.4.zip#egg=silme==0.9.4 \
 
 https://github.com/mathjazz/translate/archive/2.2.5.zip#egg=translate-toolkit==2.2.5 \
     --hash=sha256:9a23bda0f3ec07f36b6b662bf2a25492b6a354709500c736213229e94f462443
+
+https://github.com/mathjazz/python-caighdean/archive/v0.0.4.zip#egg=caighdean==0.0.4 \
```
adngdb: Any idea how we can make sure we don't forget to switch back to the official package as soon as a new version is released?

mathjazz: I'll get an email notification from GitHub when it's merged? :)

frontend/src/modules/machinery/components/Translation.js
mathjazz: Added explanation.

PR #1254, opened 2019-04-08
Fix bug 1541453: Update to fluent-syntax 0.9

mathjazz:
WIP: still need to address the first bullt point from the bug.


Pike:
I found at least 

https://github.com/mozilla/pontoon/blob/022a62d4eec72e041a218dbb8be16a8736791587/pontoon/checks/tests/test_compare_locales.py#L394

with old syntax in the test, and I changed some error messages in compare-locales, which probably just need fixing of the expected value in 

https://github.com/mozilla/pontoon/blob/022a62d4eec72e041a218dbb8be16a8736791587/pontoon/checks/tests/test_compare_locales.py#L416

Tests are a tad hard to read, no idea what the 4 failures actually are.

stasm:
Should we also update the `fluent-syntax` dependency in `package.json`?

mathjazz:
> Should we also update the `fluent-syntax` dependency in `package.json`?

Isn't that already done?
https://github.com/mozilla/pontoon/pull/1254/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2

stasm:
My bad, it's literally the first change in the PR. Not sure how I missed it. Thanks!

mathjazz:
NP.

This is now ready for a full review.

PR #1255, opened 2019-04-08
Fix bug-1541260: [python3] Add a service in docker-compose with Python 3 as the default interpreter

aniruddha2000:
This bug is about to add the new service in `docker-compose.yml` and set Python 3 as the default interpreter.


docker/Dockerfile.base
```diff
@@ -1,5 +1,7 @@
 FROM python:2.7.16-stretch
 
+FROM python:3.6.8-slim
```
jotes: @aniruddha2000 
AFAIR, you can use multiple `FROM` statements when you're trying to make a multi-stage docker build. If my knowledge is up-to-date, this feature doesn't allow to merge two different images into one.
More info: https://blog.alexellis.io/mutli-stage-docker-builds/ 

In our case, one of possible approaches is to keep two separate services:
* webapp (the existing service which use Python 2)
* webapp.python3 (the service which will run Python 3)
My initial idea was to add a separate file e.g. `docker/Dockerfile.py3.base` and use it as the base image for Python 3-related images.

@adngdb what do you think about this?

aniruddha2000: @jotes ok then doing that.

Pike:
I suggest to use a variable for the version of the base image, like so:

https://github.com/mozilla/elmo/blob/beb74b3629fb7a79603c79a3eb885f30a065d31d/docker/Dockerfile#L1-L3

https://github.com/mozilla/elmo/blob/beb74b3629fb7a79603c79a3eb885f30a065d31d/docker-compose.yml#L4-L8

jotes:
@adngdb I agree about additional commands. Unfortunately, the migration to Python3 is not yet finished.

My personal acceptance criteria was to add a docker service and then start working on Python3. New commands can introduced later  (e.g. https://bugzilla.mozilla.org/show_bug.cgi?id=1542084)

If you think that's not a good idea to keep a not fully functional piece of code on the master branch, this PR will have to wait.

Alternatively, One of possible solutions is to add those commands and mark them as "experimental". But that doesn't solve the problem of the vague acceptance criteria.

adngdb:
I guess I have 2 arguments here. The first one is that I don't like dead code, so having this docker image ready and built every time I `make build` is going to be painful for no real reason. But that can, and should be solved by actually moving the building part of that py3 image out of `make build` and into something like `make build-py3`. 

The second one is that I assume that during your work on py3 you'll want to have those tools available? If that is not the case, or at least not yet, then it might be better not to merge this now indeed. It really depends on you @jotes as you're the one who's going to work on this the most.

jotes:
@adngdb 
Thanks!
ad 1. To avoid such situation we can create a separate config file e.g. `docker-compose.python3.yml` and put `webapp.python3` there. That will give us the ability to run things on Python 3 without making regular development more painful.
ad 2. I'm planning to work on this during the upcoming weeks. However, I don't want to make your work harder :-)

aniruddha2000:
@jotes @adngdb  
Ok then creating a separate `docker-compose.python3.yml` and moving that `webapp.python3` there.

docker/Dockerfile.py3.base
```diff
@@ -0,0 +1,2 @@
+ARG pythonversion
```
jotes: If I understand Pike's intentions correctly, this argument should land in `docker/Dockerfile.base` and `Dockerfile.py3.base` should be removed.

aniruddha2000: Ok

docker-compose.yml
```diff
@@ -12,6 +12,14 @@ services:
       dockerfile: ./docker/Dockerfile.base
     image: local/pontoon_base
 
+  webapp.python3:
+    build:
+      context: .
+      dockerfile: ./docker/Dockerfile.py3.base
+      args:
+        - pythonversion=3.6.8
+    image: local/pontoon_base
```
jotes: I think we can name this image `local/pontoon-py3` to be more explicit about its contents.

adngdb:
Sorry, you all are going too fast for me. :P There's no need to extract the py3 content from the `docker-compose` file. What matters is what happen when I run `make build`. If that command is written to build only the regular `webapp` target, and there's no dependency between `webapp` and `webapp.python3` then the latter won't be built. I believe what needs to be done here is simply to remove the `docker-composer build webapp.python3` part in the `Makefile`.

Once that's done, then this PR won't have any impact on my daily work, and I'm happy to accept it.

docker-compose.python3.yml
```diff
@@ -0,0 +1,11 @@
...
+#
+# Note: Requires docker-compose 1.10+.
+version: "2"
+services:
+  # Pontoon python3 base image
+  webapp.python3:
+    build:
+      context: .
+      dockerfile: ./docker/Dockerfile.py3.base
+    image: local/pontoon_base
```
aniruddha2000: @jotes I think this `docker-compose.python3.yml` file is not doing anything because there is no `Dockerfile.py3.base` file so what should we do now?

jotes: @aniruddha2000 the first of all -> thanks for all your work!

I have some thoughts/questions:
* I can't find any reference to `docker-compose.python3.yml` . Do you want to use it in the Makefile? How it will affect  the targets?
* `docker-compose.yml` allows to parametrize the version of Python.

aniruddha2000: @jotes 

>     * I can't find any reference to `docker-compose.python3.yml` . Do you want to use it in the Makefile? How it will affect  the targets?

Can you provide me some reference how i can `docker-compose.python3.yml` this file in the Makefile?

docker-compose.yml
```diff
@@ -10,14 +10,27 @@ services:
...
+    image: local/pontoon_base_${PYTHON_VERSION}
+
+  # Pontoon python3 base image
+  webapp.python3:
+    build:
+      context: .
+      dockerfile: ./docker/Dockerfile.base
+      args:
+        pythonversion: "${PYTHON_VERSION}"
+    image: local/pontoon-py3
```
aniruddha2000: @jotes I think now it is ok

Makefile
```diff
@@ -36,6 +38,9 @@ build:
 
 	touch .docker-build
 
+build.py3: override PYTHON_VERSION=3.7.3
+build.py3: build
+
```
aniruddha2000: @jotes I think we should target the `webapp.python3` service here so we can get the `PYTHON_VERSION`

docker-compose.yml
aniruddha2000: And we can write the image like this `image: local/pontoon_base_py_${PYTHON_VERSION}`

jotes: @aniruddha2000 do we need a separate service at all if it looks the same as `webapp`?
I mean, this could reduce the size of PR.

docker-compose.python3.yml
jotes: Theoretically, you can create a target in the Makefile and just pass it there via `docker-compose -f`.
However, I think we don't need to have a new service for Python 3 if we will parametrize existing services (web app).

docker-compose.yml
```diff
@@ -10,14 +10,27 @@ services:
...
-    image: local/pontoon_base
+      args:
+        pythonversion: "${PYTHON_VERSION}"
+    image: local/pontoon_base_${PYTHON_VERSION}
+
+  # Pontoon python3 base image
+  webapp.python3:
+    build:
+      context: .
+      dockerfile: ./docker/Dockerfile.base
```
jotes: I think it should be `./docker/Dockerfile`.

Makefile
```diff
@@ -1,12 +1,13 @@
...
 DOCKER := $(shell which docker)
+export PYTHON_VERSION := 2.7.13
 
 # *IMPORTANT*
 # Don't use this instance in a production setting. More info at:
 # https://docs.djangoproject.com/en/dev/ref/django-admin/#runserver
 SITE_URL ?= http://localhost:8000
 
-.PHONY: build setup run clean test shell loaddb build-frontend build-frontend-w
+.PHONY: build build.py3 setup run clean test shell loaddb build-frontend build-frontend-w
```
adngdb: Let's use coherent command names here: `build-py3`.

docker-compose.yml
adngdb: Since the Python version is handled in the Makefile, and passed down when building the images, there is no need to have any differentiated services here. This `webapp.python3` target can be removed.

Makefile
```diff
@@ -36,6 +38,9 @@ build:
 
 	touch .docker-build
 
+build.py3: override PYTHON_VERSION=3.7.3
+build.py3: build
```
adngdb: You forgot to update these to `build-py3`.

PR #1257, opened 2019-04-09
Fix bug 1543193: Changed the url regexp for usernames to accept dots

thiagohirata:
Created a username with a dot and it broke pontoon. Included a dot as a possible character for usernames on django urls configuration.


mathjazz:
Oh, you're already here! Thanks for the patch! :)

Could you please update the regex to match the one used by Django?
https://github.com/django/django/blob/master/django/contrib/auth/validators.py

PR #1258, opened 2019-04-10
Fix bug 1542611 - Always put focus on Editor.

adngdb:
Also make sure that the cursor is at the beginning of the Editor when it loads.


adngdb:
@mathjazz I *finally* figured it out! :man_dancing:

PR #1261, opened 2019-04-12
Switch back to upstream caighdean

mathjazz:
We switched to our own caighdean fork as part of bug 1542575. Since the
patch has been merged, we can now switch back to upstream caighdean.

@jotes One more r?


mathjazz:
Thanks :)

PR #1262, opened 2019-04-12
bug 1544048, support flexible locale handling in project config, r=mathjazz

Pike:


Pike:
Forgot to git add the new file, and then didn't lint. Fixes pushed.

PR #1263, opened 2019-04-15
Bug 1542027: Multiple tries to download GPG keys in docker image

pierotofy:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1542027 using the Jarek's suggested patch.


jotes:
Hi @pierotofy :-)

PR #1266, opened 2019-04-16
Fix bug 1518820: Add errors and warnings

mathjazz:
This is a WIP, see TODO for details. Due to the indent fix, it will be easier to review on a commit-by-commit basis.

TODO:
- [x] Fix showing/hiding of failed checks popup.
- [x] Add ability to `ignore_warnings` (pass param to `update_translation` and add button).
- [x] Close failed checks popup when textarea content changes.

### DIFF
--- a/frontend/src/modules/editor/actions.js
+++ b/frontend/src/modules/editor/actions.js
@@ -61,13 +61,16 @@ export type FailedChecks = {|
 export type UpdateFailedChecksAction = {|
     +type: typeof UPDATE_FAILED_CHECKS,
     +failedChecks: FailedChecks,
+    +source: '' | 'stored' | 'submitted' | 'approved',
 |};
 export function updateFailedChecks(
     failedChecks: FailedChecks,
+    source: '' | 'stored' | 'submitted' | 'approved',
 ): UpdateFailedChecksAction {
     return {
         type: UPDATE_FAILED_CHECKS,
         failedChecks,
+        source,
     };
 }
 
@@ -122,7 +125,7 @@ export function sendTranslation(
         );
 
         if (content.failedChecks) {
-            dispatch(updateFailedChecks(content.failedChecks));
+            dispatch(updateFailedChecks(content.failedChecks, 'submitted'));
         }
         else if (content.same) {
             // The translation that was provided is the same as an existing
--- a/frontend/src/modules/editor/actions.js
+++ b/frontend/src/modules/editor/actions.js
@@ -61,11 +61,11 @@ export type FailedChecks = {|
 export type UpdateFailedChecksAction = {|
     +type: typeof UPDATE_FAILED_CHECKS,
     +failedChecks: FailedChecks,
-    +source: '' | 'stored' | 'submitted' | 'approved',
+    +source: '' | 'stored' | 'submitted' | number,
 |};
 export function updateFailedChecks(
     failedChecks: FailedChecks,
-    source: '' | 'stored' | 'submitted' | 'approved',
+    source: '' | 'stored' | 'submitted' | number,
 ): UpdateFailedChecksAction {
     return {
         type: UPDATE_FAILED_CHECKS,
--- a/frontend/src/modules/editor/actions.js
+++ b/frontend/src/modules/editor/actions.js
@@ -10,6 +10,7 @@ import type { DbEntity } from 'modules/entitieslist';
 
 export const RESET_SELECTION: 'editor/RESET_SELECTION' = 'editor/RESET_SELECTION';
 export const UPDATE: 'editor/UPDATE' = 'editor/UPDATE';
+export const UPDATE_FAILED_CHECKS: 'editor/UPDATE_FAILED_CHECKS' = 'editor/UPDATE_FAILED_CHECKS';
 export const UPDATE_SELECTION: 'editor/UPDATE_SELECTION' = 'editor/UPDATE_SELECTION';
 
 
@@ -42,6 +43,30 @@ export function updateSelection(content: string): UpdateSelectionAction {
     return {
         type: UPDATE_SELECTION,
         content,
+    };
+}
+
+
+/**
+ * Update failed checks in the active editor.
+ */
+export type FailedChecks = {|
+    +clErrors: Array<string>,
+    +pErrors: Array<string>,
+    +clWarnings: Array<string>,
+    +pndbWarnings: Array<string>,
+    +ttWarnings: Array<string>,
+|};
+export type UpdateFailedChecksAction = {|
+    +type: typeof UPDATE_FAILED_CHECKS,
+    +failedChecks: FailedChecks,
+|};
+export function updateFailedChecks(
+    failedChecks: FailedChecks,
+): UpdateFailedChecksAction {
+    return {
+        type: UPDATE_FAILED_CHECKS,
+        failedChecks: failedChecks,
     };
 }
 
@@ -83,7 +108,10 @@ export function sendTranslation(
             forceSuggestions,
         );
 
-        if (content.same) {
+        if (content.failedChecks) {
+            dispatch(updateFailedChecks(content.failedChecks));
+        }
+        else if (content.same) {
             // The translation that was provided is the same as an existing
             // translation for that entity.
             // Show an error.
@@ -115,5 +143,6 @@ export default {
     resetSelection,
     sendTranslation,
     update,
+    updateFailedChecks,
     updateSelection,
 };
--- a/frontend/src/modules/editor/actions.js
+++ b/frontend/src/modules/editor/actions.js
@@ -8,6 +8,7 @@ import { actions as entitiesActions } from 'modules/entitieslist';
 import type { DbEntity } from 'modules/entitieslist';
 
 
+export const RESET_FAILED_CHECKS: 'editor/RESET_FAILED_CHECKS' = 'editor/RESET_FAILED_CHECKS';
 export const RESET_SELECTION: 'editor/RESET_SELECTION' = 'editor/RESET_SELECTION';
 export const UPDATE: 'editor/UPDATE' = 'editor/UPDATE';
 export const UPDATE_FAILED_CHECKS: 'editor/UPDATE_FAILED_CHECKS' = 'editor/UPDATE_FAILED_CHECKS';
@@ -72,8 +73,7 @@ export function updateFailedChecks(
 
 
 /**
- * Update the content that should replace the currently selected text in the
- * active editor.
+ * Reset content to default value.
  */
 export type ResetSelectionAction = {|
     +type: typeof RESET_SELECTION,
@@ -81,6 +81,19 @@ export type ResetSelectionAction = {|
 export function resetSelection(): ResetSelectionAction {
     return {
         type: RESET_SELECTION,
+    };
+}
+
+
+/**
+ * Reset failed checks to default value.
+ */
+export type ResetFailedChecksAction = {|
+    +type: typeof RESET_FAILED_CHECKS,
+|};
+export function resetFailedChecks(): ResetFailedChecksAction {
+    return {
+        type: RESET_FAILED_CHECKS,
     };
 }
 
@@ -140,6 +153,7 @@ export function sendTranslation(
 
 
 export default {
+    resetFailedChecks,
     resetSelection,
     sendTranslation,
     update,
--- a/frontend/src/modules/editor/actions.js
+++ b/frontend/src/modules/editor/actions.js
@@ -66,7 +66,7 @@ export function updateFailedChecks(
 ): UpdateFailedChecksAction {
     return {
         type: UPDATE_FAILED_CHECKS,
-        failedChecks: failedChecks,
+        failedChecks,
     };
 }
 
--- a/frontend/src/modules/editor/actions.js
+++ b/frontend/src/modules/editor/actions.js
@@ -113,6 +113,7 @@ export function sendTranslation(
     forceSuggestions: boolean,
     nextEntity: ?DbEntity,
     router: Object,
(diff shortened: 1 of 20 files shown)


frontend/src/modules/editor/actions.js
```diff
@@ -44,6 +45,30 @@ export function updateSelection(content: string): UpdateSelectionAction {
...
 
+/**
+ * Update failed checks in the active editor.
+ */
+export type FailedChecks = {|
+    +clErrors: Array<string>,
+    +pErrors: Array<string>,
+    +clWarnings: Array<string>,
+    +pndbWarnings: Array<string>,
+    +ttWarnings: Array<string>,
```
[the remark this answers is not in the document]
mathjazz: I know, I know... I looked into refactoring checks before digging into it, but the patch would be pretty big. Maybe I should file a P5 Translate.Next bug, so we don't forget about it?

I'm thinking checks should instead be formatted like this:
```js
failedChecks = {
    errors: [
        {
            message: 'Missing variable %s',
            library: 'compare-locales'
        },
        ...
    ],
    warnings: [...],
}
```

frontend/src/modules/editor/components/Editor.js
```diff
@@ -100,6 +101,27 @@ export class EditorBase extends React.Component<InternalProps> {
...
+        }
+
+        if (editor[type].length) {
+            return editor[type];
+        }
+
+        const plural = pluralForm === -1 ? 0 : pluralForm;
+        const translation = selectedEntity.translation[plural];
+
+        if (!translation || (!translation.approved && !translation.fuzzy)) {
```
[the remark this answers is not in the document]
mathjazz: We only show checks of the active translation if it's approved or fuzzy (because than its status is also colored as error/warning in the string list).

But now that I think about it, maybe we should just show the failed checks by default anyways? Otherwise the user might think it's OK to approve the translation, and once she approves it, she'll see the popup (which is not an optimal UX).

What do you think?

frontend/src/modules/editor/actions.js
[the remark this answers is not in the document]
adngdb: Oh god yes that looks so much better! What I was wondering is if you could maybe add a layer between the API and the reducer to make the data from the server look more like what you show up there? It might be too much work, though, feel free not to do that.

However, yes, please do file a bug about refactoring errors and warnings. :-) Not a Translate.Next though, I think it's a general Pontoon issue.

frontend/src/modules/editor/components/Editor.js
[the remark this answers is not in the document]
adngdb: That sounds reasonable indeed, though maybe we should refrain from changing the behavior from Translate.Current for now?

frontend/src/modules/editor/actions.js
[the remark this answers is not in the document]
mathjazz: I'm thinking the bug I'm about to file would make such API redundant, so I'd rather not spent time working on it.

frontend/src/modules/editor/components/Editor.js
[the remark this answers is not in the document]
mathjazz: Good point! I'll add a comment.

frontend/src/modules/editor/components/FailedChecks.js
```diff
@@ -0,0 +1,65 @@
...
+            <Localized
+                id="editor-FailedChecks--close"
+                attrs={{ ariaLabel: true }}
+            >
+                <button
+                    aria-label="Close failed checks popup"
+                    className="close"
+                    onClick={ this.closeFailedChecks }
+                >×</button>
+            </Localized>
```
mathjazz: @stasm Is there a way to make only the `ariaLabel` localized, without the button text?

stasm: Sadly, not right now. There's an open issue about it: https://github.com/projectfluent/fluent.js/issues/184

mathjazz: Thanks, that's a good enough answer right now.

frontend/src/modules/editor/components/Editor.js
```diff
@@ -101,6 +107,34 @@ export class EditorBase extends React.Component<InternalProps> {
...
+            state.selectedEntity.pk,
+            state.locale.code,
+            state.parameters.resource,
+            state.pluralForm,
+            translationId,
+            state.nextEntity,
+            state.router,
+            ignoreWarnings,
+        ));
+    }
```
mathjazz: This is very ugly, because `updateTranslationStatus` already exists in `EntityDetails` component, but I can't pass it to the `Editor`.

mathjazz:
@adngdb I've addressed your feedback and made two additional changes:
* Failed checks are now always stored in the editor (including the ones coming from the DB). That allows us to use `resetFailedChecks` to close the failed checks popup.
* Save/submit/approve Anyway button is now available.

Could you have another look, please?

There's one open issue remaining - the failed checks popup should close when you change the content of the editor. If I `resetFailedChecks` in `Editor.componentDidUpdate`, I hit the infinite recursion of course. Is the way to go here to factor `FailedChecks` out into their own component?

I wanted to hear your opinion first, because there might be other refactorings needed here. See my [comment](https://github.com/mozilla/pontoon/pull/1266#discussion_r276540834). Thanks!

frontend/src/modules/editor/reducer.js
```diff
@@ -32,7 +32,13 @@ export type EditorState = {|
     +selectionReplacementContent: string,
     +errors: Array<string>,
     +warnings: Array<string>,
-    +source: '' | 'stored' | 'submitted' | 'approved',
+
+    // A source of failed checks (erros and warnings). Possible values:
```
adngdb: s/erros/errors

mathjazz:
> Regarding the panel closing issue, can you maybe show me a commit with the problem you hit? It might just be a matter of finding the right condition.

Pushed https://github.com/mozilla/pontoon/pull/1266/commits/dc1a3fbb1bd90c6c258b6d550ce23952076e1646. You should see the following error in the console on any translate.next view:
```
Error: Maximum update depth exceeded.
This can happen when a component repeatedly calls setState inside
componentWillUpdate or componentDidUpdate.
React limits the number of nested updates to prevent infinite loops.
```

mathjazz:
That prevents failed checks from appearing... ever.

The reason for this is that we show/hide failed checks by updating the **`editor`** state. So when we add checks in `updateFailedChecks`, `editor` state changes, forcing `GenericEditor` component to re-render, because it relies on the `editor` state. After render, `componentDidUpdate` is called, which then calls `resetFailedChecks` (because the condition is `true`).

That's why I'm thinking, maybe we should make `FailedChecks` a separate module, with its state independent of the editor?

mathjazz:
> It seems your patch breaks the stats in the FilterPanel. This happens after I change the status of a translation (approve, reject):

That bug actually happens on master, too.

adngdb:
I've tried locally, and this does the trick:

```js
        if (
            (
                prevProps.editor.errors.length ||
                prevProps.editor.warnings.length
            ) && (
                this.props.editor.errors.length ||
                this.props.editor.warnings.length
            )
        ) {
            this.props.resetFailedChecks();
        }
```

You don't want to reset if the errors and warnings were just added to the editor object, but in all other cases you do.

mathjazz:
> I've tried locally, and this does the trick:

Thanks, that is super helpful! I've added that, now I'll write some tests and then I'll ask you for another round of review.

frontend/src/modules/editor/actions.js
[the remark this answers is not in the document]
mathjazz: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1545740.

@jotes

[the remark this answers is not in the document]
jotes: @mathjazz thanks! I totally agree.  The current structure of the API response is a bit complicated (I didn't consider the later use and scaling). Your proposal looks very good :-)

mathjazz:
@adngdb This is now ready for another look. I can't figure out why one of the tests is failing (left a [comment](https://github.com/mozilla/pontoon/pull/1266/commits/d44c772eb21aa12975ec8226bcfd1fc9eba3cc39) in the commit message). I'd appreciate if you could enlighten me. :)

frontend/src/modules/entitydetails/components/EntityDetails.test.js
```diff
@@ -107,6 +107,43 @@ describe('<EntityDetailsBase>', () => {
...
+                original: 'something',
+                translation: [{
+                    approved: true,
+                    string: 'quelque chose',
+                    errors: ['Error1'],
+                }],
+            },
+        });
+
+        expect(editor.actions.updateFailedChecks.calledOnce.toBeTruthy());
```
adngdb:
```js
expect(editor.actions.updateFailedChecks.calledOnce).toBeTruthy();
```

`toBeTruthy` is a property of the `expect` framework. :-)

adngdb: For this to work, you'll need to mock those two actions though. See for example: https://github.com/mozilla/pontoon/blob/master/frontend/src/modules/entitydetails/components/EntityDetails.test.js#L115

mathjazz: Thanks :) Updated, but the 2nd test is still failing.

frontend/src/modules/editor/components/EditorProxy.test.js
```diff
@@ -24,4 +31,41 @@ describe('<EditorProxy>', () => {
 
         expect(wrapper.type()).toEqual(GenericEditor);
     });
+
+    it('resets failed checks when translation changes, but errors and warnings do not', () => {
+        const resetMock = sinon.stub();
+
+        const wrapper = mount(<EditorProxy
```
adngdb: I'm not sure you need to `mount` here, `shallow` might be sufficient. It's better to avoid using `mount` as much as possible, because `mount` will instantiate all the components in the hierarchy below this one, whether `shallow` only instantiate the one component you're testing, and no others.

frontend/src/modules/editor/components/EditorProxy.test.js
```diff
@@ -24,4 +31,41 @@ describe('<EditorProxy>', () => {
...
+            }
+        });
+
+        expect(resetMock.calledOnce).toBeTruthy();
+    });
+
+    it('keeps failed checks when translation changes, along with errors or warnings', () => {
+        const resetMock = sinon.stub();
+
+        const wrapper = mount(<EditorProxy
```
adngdb: Same here.

frontend/src/modules/entitydetails/components/EntityDetails.test.js
```diff
@@ -107,6 +117,45 @@ describe('<EntityDetailsBase>', () => {
...
+                original: 'something',
+                translation: [{
+                    approved: true,
+                    string: 'quelque chose',
+                }],
+            },
+        });
+
+        expect(editor.actions.updateFailedChecks.calledOnce).toBeFalsy();
+        expect(editor.actions.resetFailedChecks.calledOnce).toBeTruthy();
```
adngdb: Have you checked that the method that calls this action is actually called? I usually run those kinds of debugs with `console.log`, and make sure that the correct path gets ran.

mathjazz: Thanks for the debugging help. :) Indeed, there was a problem with the code and I fixed it.

But I'm only **almost** there... For some reason `editor.actions.updateFailedChecks.calledOnce` is still `truthy`. I thought count should be reset in `afterAll`?

adngdb: Ah, `afterAll` only runs after all tests have run. You want to use `afterEach` as well, see: https://github.com/mozilla/pontoon/blob/master/frontend/src/modules/editor/components/Editor.test.js#L66

frontend/src/modules/entitydetails/components/EntityDetails.test.js
```diff
@@ -96,12 +96,12 @@ function createEntityDetailsWithStore() {
...
 
 describe('<EntityDetailsBase>', () => {
-    beforeAll(() => {
+    beforeEach(() => {
         sinon.stub(editor.actions, 'updateFailedChecks').returns({ type: 'whatever'});
         sinon.stub(editor.actions, 'resetFailedChecks').returns({ type: 'whatever'});
     });
 
-    afterAll(() => {
+    afterEach(() => {
```
adngdb: I think that's the wrong approach. Those were fine. I believe what you want to do is add a third step, `afterEach`, that calls `.reset()`. Otherwise, you mock and de-mock the object after every test. I have no idea if that has an impact in terms of performance, but I'm pretty sure that's not the philosophy of `sinon`.

mathjazz: I tried that first, but it gave errors:
* `TypeError: editor.actions.updateFailedChecks.restore is not a function`
* `Received: undefined`

adngdb: I've just tested locally with this code and it passes: 

```js
    beforeAll(() => {
        sinon.stub(editor.actions, 'updateFailedChecks').returns({ type: 'whatever'});
        sinon.stub(editor.actions, 'resetFailedChecks').returns({ type: 'whatever'});
    });

    afterEach(() => {
        editor.actions.updateFailedChecks.reset();
        editor.actions.resetFailedChecks.reset();
    });

    afterAll(() => {
        editor.actions.updateFailedChecks.restore();
        editor.actions.resetFailedChecks.restore();
    });
```

mathjazz:
Thanks for the advice, the tests should now work as expected!

PR #1268, opened 2019-04-18
Fix bug 1545450: Fix entity status icons

mathjazz:
This fixes two things:

1. It takes into account all plural forms when determining entity status, not just the first one.

2. It adds support for `errors`, `warnings` and `partial` stati.

### DIFF
--- a/frontend/src/modules/entitieslist/components/Entity.js
+++ b/frontend/src/modules/entitieslist/components/Entity.js
@@ -64,19 +64,19 @@ export default class Entity extends React.Component<Props> {
             }
         });
 
-        if (errors > 0) {
+        if (errors) {
             return 'errors';
         }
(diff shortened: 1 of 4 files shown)


frontend/src/modules/entitieslist/components/Entity.js
```diff
@@ -34,14 +37,48 @@ type Props = {
...
+            }
+            else if (translation.fuzzy) {
+                fuzzy++;
+            }
+        });
+
+        if (errors > 0) {
+            return 'errors';
+        }
+        else if (warnings > 0) {
```
adngdb: No need for the `else`s here because you `return`.

adngdb: You can also simplify to `if (errors) {`.

PR #1270, opened 2019-04-19
Fix bug 1500871 - Add support for status notifications.

adngdb:
@mathjazz this is an early look, but the feature "works", and I wanted to have your opinion on the way I handle animations. You can test it by sending an already existing translation.

### DIFF
--- a/frontend/public/index.html
+++ b/frontend/public/index.html
@@ -38,7 +38,7 @@
     <div
         id="root"
         data-csrf-token="{{ csrf_token }}"
-        data-notifications="{{ notifications | tojson | forceescape }}"
+        data-notifications="{{ notifications | to_json | forceescape }}"
     ></div>
     <!--
       This HTML file is a template.
--- a/frontend/public/index.html
+++ b/frontend/public/index.html
@@ -21,7 +21,6 @@
     -->
     <title>Translate.Next</title>
 
-
     <!-- Inlining CSS mitigates/prevents fouc. -->
     <style>
       body {
@@ -36,7 +35,11 @@
     <noscript>
       You need to enable JavaScript to run this app.
     </noscript>
-    <div id="root" data-csrf-token="{{ csrf_token }}"></div>
+    <div
+        id="root"
+        data-csrf-token="{{ csrf_token }}"
+        data-notifications="{{ notifications | tojson | forceescape }}"
+    ></div>
     <!--
       This HTML file is a template.
       If you open it directly in the browser, you will see an empty page.
--- a/frontend/public/index.html
+++ b/frontend/public/index.html
@@ -38,7 +38,7 @@
     <div
         id="root"
         data-csrf-token="{{ csrf_token }}"
-        data-notifications="{{ notifications | to_json | forceescape }}"
+        data-notifications="{{ notifications | to_json }}"
     ></div>
     <!--
       This HTML file is a template.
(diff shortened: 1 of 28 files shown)


frontend/src/core/notification/components/NotificationPanel.js
```diff
@@ -0,0 +1,71 @@
...
+        let hideClass = '';
+        if (notification.messages.length) {
+            hideClass = ' showing';
+        }
+        if (this.state.hiding) {
+            hideClass = ' hide';
+        }
+
+        return <ul className={ 'notification-panel' + hideClass }>
+            { notification.messages.map((notif, index) => {
```
mathjazz: This use case in particular (and also some others) makes me wonder whether we should rename:
`code/notification` -> `code/notifications`
`NotificationPanel` -> `NotificationsPanel`

Which would also be consistent with `notifications` in the `UserState` and `UserNotificationsMenu`.

frontend/src/core/notification/components/NotificationPanel.css
```diff
@@ -0,0 +1,32 @@
...
+    color: #7BC876;
+    cursor: pointer;
+    font-style: italic;
+    left: 0;
+    margin: 0;
+    position: fixed;
+    top: -60px;
+    width: 100%;
+    z-index: 100;
+    transition: 500ms;
```
mathjazz: I'd reduce that to `200ms`, to keep the animation as "fast" as on `Translation.Current`.

frontend/src/core/notification/components/NotificationPanel.js
adngdb: The thing is, the UI can only ever show one notification at a time. I've kept that `map` in here because that's how it was implemented in the original code, but if you look at the reducer, the array always has maximum 1 element. So maybe we should just change this code instead? Or are we likely to change this notification system in the future to make it allow chaining notifications or showing multiple at once?

mathjazz: Good point. I don't think we'll add ability to display multiple notifications anytime soon.

adngdb:
@mathjazz this is now ready for a real review. :-)

frontend/src/modules/history/actions.js
```diff
@@ -102,25 +103,57 @@ export function updateStatus(
...
+
+        // Show a notification to explain what happened.
+        if (results.translation) {
+            let statusChange = change;
+            if (change.endsWith('e')) {
+                statusChange += 'd';
+            }
+            else {
+                statusChange += 'ed';
+            }
```
mathjazz: This is hard to read. Especially if you consider we should make notifications localizable.

adngdb:
>     2. "Can't load specified string" message (when specifying non-existent string ID via URL).

I've added a comment to [bug 1539228](https://bugzilla.mozilla.org/show_bug.cgi?id=1539228#c2) about that, I don't think it makes sense to implement that notification until we have the feature properly working.

> Also, I was somehow expecting this patch to bring support for NProgress. I guess I was wrong, so we should file a bug for that.

Ah yeah I did not think about that and I believe it is out of context. I've filed [bug 1547274](https://bugzilla.mozilla.org/show_bug.cgi?id=1547274).

mathjazz:
A couple of python tests are failing.

frontend/public/index.html
```diff
@@ -36,7 +35,11 @@
     <noscript>
       You need to enable JavaScript to run this app.
     </noscript>
-    <div id="root" data-csrf-token="{{ csrf_token }}"></div>
+    <div
+        id="root"
+        data-csrf-token="{{ csrf_token }}"
+        data-notifications="{{ notifications | tojson | forceescape }}"
```
mathjazz: `to_json`

PR #1271, opened 2019-04-23
Fix bug 1543177: Always show selected entity in the list

mathjazz:


frontend/src/modules/entitieslist/components/EntitiesList.js
```diff
@@ -78,6 +85,32 @@ export class EntitiesListBase extends React.Component<InternalProps> {
         ) {
             this.props.dispatch(actions.reset());
         }
+
+        this.scrollToSelectedElement();
+    }
+
+    scrollToSelectedElement() {
+        const list = this.list.current;
+        const element = list.querySelector('li.selected');
```
adngdb: I think it's possible for `list` to be undefined here, so there should be a test to check that before you call `.querySelector` on it. I'm surprised that Flow didn't fail on that though, so maybe I'm wrong?

mathjazz: The `render()` method executes before `componentDidUpdate()`, which calls `scrollToSelectedElement()`, and always sets the ref. So I think we're good.

PR #1273, opened 2019-04-25
Fix bug 1543158: Prevent multiple review/delete requests

mathjazz:
This solution is inspired by: https://stackoverflow.com/a/49642037. I tried to avoid using `ref`.

I wonder if there's a better approach for preventing multiple in React world. If not, I'll add a similar solution for saving translations/suggestions.

### DIFF
--- a/frontend/src/modules/editor/components/Editor.js
+++ b/frontend/src/modules/editor/components/Editor.js
@@ -161,6 +161,8 @@ export class EditorBase extends React.Component<InternalProps, State> {
             return null;
         }
 
+        const FailedChecksWithActionsDisabled = entitydetails.withActionsDisabled(FailedChecks);
+
         return <div className="editor">
             <plural.PluralSelector />
             <EditorProxy
@@ -176,7 +178,7 @@ export class EditorBase extends React.Component<InternalProps, State> {
                 updateTranslationStatus={ this.updateTranslationStatus }
             />
             <menu>
-                <FailedChecks
+                <FailedChecksWithActionsDisabled
                     source={ this.props.editor.source }
                     user={ this.props.user }
                     errors={ this.props.editor.errors }
@@ -184,6 +186,8 @@ export class EditorBase extends React.Component<InternalProps, State> {
                     resetFailedChecks={ this.resetFailedChecks }
                     sendTranslation={ this.sendTranslation }
                     updateTranslationStatus={ this.updateTranslationStatus }
+                    isActionDisabled={ self.isActionDisabled }
+                    restoreAction={ self.restoreAction }
                 />
                 { !this.props.user.isAuthenticated ?
                     <Localized
--- a/frontend/src/modules/editor/components/Editor.js
+++ b/frontend/src/modules/editor/components/Editor.js
@@ -24,6 +24,7 @@ import TranslationLength from './TranslationLength';
 import type { Locale } from 'core/locales';
 import type { NavigationParams } from 'core/navigation';
 import type { UserState } from 'core/user';
+import { withActionsDisabled } from 'core/utils';
 import type { ChangeOperation } from 'modules/history';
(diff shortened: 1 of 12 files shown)


mathjazz:
I've added tests and extended functionality to submitting translations with both, keyboard and mouse events.

frontend/src/modules/editor/components/Editor.js
```diff
@@ -78,6 +96,11 @@ export class EditorBase extends React.Component<InternalProps> {
     }
 
     sendTranslation = (ignoreWarnings: ?boolean) => {
+        if (this.state.isSendTranslationDisabled) {
+            return;
+        }
+        this.setState({ isSendTranslationDisabled: true });
```
adngdb: These could be refactored into a Higher-Order component, allowing to prevent to pattern repetition. See https://reactjs.org/docs/higher-order-components.html

mathjazz: I gave HOC a [try](https://github.com/mozilla/pontoon/pull/1273/commits/c1c0ecd4705bae4bf8b53d4ab92b749031d9aeb5), and now I'm not sure it's the right tool for the job.

My understanding is that we're passing props from the WrappedComponent to HOC, which then adds additional props, and does something with them. We don't seem to have access to those HOC-specific props in the WrappedComponent.

In our use case here we'd need access, though. E.g. in `render()` and when restoring actions.

adngdb:
@mathjazz I've addressed the HOC code here: https://github.com/mathjazz/pontoon/pull/10

frontend/src/modules/editor/components/Editor.js
```diff
@@ -46,14 +46,32 @@ type InternalProps = {|
     dispatch: Function,
 |};
 
+type State = {|
+    isSendTranslationDisabled: boolean,
+|};
+
```
adngdb: Ewwww sorry I forgot to include that one in my refactor. :( Mind doing it?

frontend/src/modules/entitydetails/components/EntityDetails.test.js
```diff
@@ -119,8 +119,10 @@ describe('<EntityDetailsBase>', () => {
     it('loads the correct list of components', () => {
         const wrapper = createShallowEntityDetails();
 
+        expect(wrapper.text()).toContain('EntityNavigation');
         expect(wrapper.text()).toContain('Metadata');
-        expect(wrapper.text()).toContain('Editor');
+        //expect(wrapper.text()).toContain('Editor');
```
mathjazz: If I uncomment this line, the test fails with the following error:

```
 FAIL  src/modules/entitydetails/components/EntityDetails.test.js (8.491s)
  <EntityDetailsBase>
    ✓ shows an empty section when no entity is selected (30ms)
    ✕ loads the correct list of components (72ms)
    ✓ shows failed checks for approved (or fuzzy) translations with errors or warnings (23ms)
    ✓ hides failed checks for approved (or fuzzy) translations without errors or warnings (4ms)
  <EntityDetails>
    ✓ dispatches the updateStatus action when updateTranslationStatus is called (296ms)

  ● <EntityDetailsBase> › loads the correct list of components

    expect(string).toContain(value)

    Expected string:
      "<EntityNavigation /><Metadata /><_temp /><Helpers />"
    To contain value:
      "Editor"

      122 |         expect(wrapper.text()).toContain('EntityNavigation');
      123 |         expect(wrapper.text()).toContain('Metadata');
    > 124 |         expect(wrapper.text()).toContain('Editor');
          |                                ^
      125 |         expect(wrapper.text()).toContain('Helpers');
      126 |     });
      127 | 
```

The same thing happens if I use the following line at the end of Editor.js:
```js
export default connect(mapStateToProps)(withActionsDisabled(EditorBase));
```

PR #1274, opened 2019-04-25
Fix bug 1543182: Move to next plural, not entity (if exists)

mathjazz:


frontend/src/modules/editor/actions.js
```diff
@@ -147,10 +149,15 @@ export function sendTranslation(
...
-                dispatch(navActions.updateEntity(router, nextEntity.pk.toString()));
+            if (nextEntity) {
+                // The change did work, we want to move on to the next Entity or pluralForm.
+                const nextPluralForm = pluralForm + 1;
+                if (pluralForm !== -1 && nextPluralForm < locale.cldrPlurals.length) {
+                    dispatch(pluralActions.select(nextPluralForm));
+                }
+                else if (nextEntity.pk !== entity) {
+                    dispatch(navActions.updateEntity(router, nextEntity.pk.toString()));
+                }
```
mathjazz: This code is repeated in `frontend/src/modules/history/actions.js`. I'd like to have it on once place. Where should that be? Maybe `navActions.updateEntityOrPlural`?

mathjazz:
@adngdb `import { actions as pluralActions } from 'core/plural';` in `navigation.actions` is crashing the app and causing the tests to fail with `TypeError: Cannot read property 'default' of undefined`.

I don't understand the errors message in failed tests. Any hints how to avoid that?

PR #1275, opened 2019-04-26
Fix bug 1519931: Show string length and countdown

mathjazz:
This patch adds support for translation length information. For `LANG` strings with `MAX_LENGTH` provided in their comment, it shows countdown instead.

### DIFF
--- a/frontend/src/modules/editor/components/Editor.js
+++ b/frontend/src/modules/editor/components/Editor.js
@@ -65,7 +65,7 @@ export class EditorBase extends React.Component<InternalProps> {
     copyOriginalIntoEditor = () => {
         const { selectedEntity, pluralForm } = this.props;
         if (selectedEntity) {
-            if (pluralForm === -1 || pluralForm === 1) {
+            if (pluralForm === -1 || pluralForm === 0) {
                 this.updateTranslation(selectedEntity.original, true);
             }
             else {
@@ -187,6 +187,7 @@ export class EditorBase extends React.Component<InternalProps> {
                         <KeyboardShortcuts />
                         <TranslationLength
                             entity={ this.props.selectedEntity }
+                            pluralForm={ this.props.pluralForm }
                             translation={ this.props.editor.translation }
                         />
                         <div className="actions">
--- a/frontend/src/modules/editor/components/Editor.js
+++ b/frontend/src/modules/editor/components/Editor.js
@@ -19,6 +19,7 @@ import FailedChecks from './FailedChecks';
 import EditorProxy from './EditorProxy';
 import EditorSettings from './EditorSettings';
 import KeyboardShortcuts from './KeyboardShortcuts';
+import TranslationLength from './TranslationLength';
 
 import type { Locale } from 'core/locales';
 import type { NavigationParams } from 'core/navigation';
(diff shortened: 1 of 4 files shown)


frontend/src/modules/editor/components/TranslationLength.js
```diff
@@ -0,0 +1,87 @@
...
+type Props = {|
+    entity: ?DbEntity,
+    translation: string,
+|};
+
+
+/*
+ * Shows translation length vs. original string length, or countdown.
+ * 
+ * Countdown is current only supported for LANG strings, which use special
```
adngdb: `s/current/currently`

frontend/src/modules/editor/components/TranslationLength.js
```diff
@@ -0,0 +1,87 @@
...
+ * syntax in the comment to define maximum translation length. MAX_LENGTH
+ * is provided for strings without HTML tags, so they need to be stripped.
+ */
+export default class TranslationLengthBase extends React.Component<Props> {
```
adngdb: Please call it `TranslationLength`.

frontend/src/modules/editor/components/TranslationLength.js
```diff
@@ -0,0 +1,87 @@
...
+    getLimit() {
+        const entity = this.props.entity;
+
+        if (!entity) {
```
adngdb: If you make this `if (!entity || entity.format !== 'lang') {` you can simplify the logic and avoid doing the `split` when the format is not LANG.

frontend/src/modules/editor/components/TranslationLength.js
```diff
@@ -0,0 +1,87 @@
...
+            return null;
+        }
+
+        const split = entity.comment.split('\n');
```
adngdb: I like the name `parts` better because it's not the same as the function, avoiding confusion when reading this kind of code.

adngdb: Also makes refactoring easier. :-)

frontend/src/modules/editor/components/TranslationLength.test.js
```diff
@@ -0,0 +1,68 @@
...
+    });
+
+    it('does not strips html from translation when calculating length', () => {
+        const wrapper = shallow(<TranslationLengthBase
+            translation='12<span>34</span>56'
+            entity={ LENGTH_ENTITY }
+        />);
+
+        expect(wrapper.find('.translation-vs-original').childAt(0).text()).toEqual('19');
+    });
```
adngdb: These are great, I would just add a test with a more complex `comment` structure, to make sure that the splitting works.

frontend/src/modules/editor/components/Editor.js
```diff
@@ -64,7 +65,7 @@ export class EditorBase extends React.Component<InternalProps> {
     copyOriginalIntoEditor = () => {
         const { selectedEntity, pluralForm } = this.props;
         if (selectedEntity) {
-            if (pluralForm === -1 || pluralForm === 1) {
+            if (pluralForm === -1 || pluralForm === 0) {
```
adngdb: Wooops! Thanks for catching that! :heart:

PR #1280, opened 2019-05-06
Fix bug 1525986: Unsaved changes warning

mathjazz:
~~This is a WIP - the warning is only shown when changes are made in the editor before you go to the Next/Previous entity using the navigation toolbar.~~

@adngdb Could you please have a look if this is the way to go here?

One of the issues I bumped into when I tried to show the warning also when the user clicks on the entity in the entity list was how to check if the editor has unsaved changes. I tried to add `activeTranslation` and `editor` to the `EntitiesList` state, but that causes a cyclic import.


frontend/src/modules/editor/components/UnsavedChanges.js
```diff
@@ -0,0 +1,76 @@
...
+
+    leaveAnyway = () => {
+        const { callback, hide, ignore } = this.props;
+        if (callback) {
+            ignore();
+
+            setTimeout(() => {
+                callback();
+                hide();
+            }, 10);
```
mathjazz: I wonder how to circumvent the `setTimeout` here. Without it, the callback (which depends on state) executes before the state is updated.

frontend/src/modules/editor/components/Editor.js
```diff
@@ -162,6 +171,12 @@ export class EditorBase extends React.Component<InternalProps> {
                     sendTranslation={ this.sendTranslation }
                     updateTranslationStatus={ this.updateTranslationStatus }
                 />
+                <UnsavedChanges
+                    callback={ this.props.editor.unsavedChangesCallback }
+                    hide={ this.hideUnsavedChanges }
+                    ignore={ this.ignoreUnsavedChanges }
+                    show={ this.props.editor.unsavedChanges }
```
adngdb: This sounds like it's an action when it's actually data. How about calling it something like `changes` instead?

frontend/src/modules/editor/components/UnsavedChanges.js
adngdb: I think you want to call the callback after the component's next render, in `componentDidUpdate`. In React the "correct" way to be sure that state has been updated is to wait for the next render to happen.

mathjazz:
After lots of changes, this is finally fully working and tested.

Please review. This time around it probably doesn't make much sense to do it commit-by-commit.

adngdb:
I'm able to consistently produce a "Maximum depth exceeded" issue. Steps:

1. Select any entity, make a change in the editor.
2. Click on the same entity, "Unsaved changes" popup shows up.
3. Click "Leave anyway".
4. Type something in the search bar that includes the selected entity.
5. Wait a bit (or look at the logs) while the `unsavedchanges/SHOW` actions is triggered in a loop.

adngdb:
@mathjazz Sadly now I can get that recursion error by just doing: 

1. Select an entity, make a change to the translation.
2. Start a search.
3. Wait for recursion error to occur (see the logs).

:cry:

mathjazz:
Updated! Hopefully it works now. 😸

PR #1290, opened 2019-05-20
Fix bug 1543162 - Cancel previously running requests when changing entity.

adngdb:
Whenever we fetch new content for the History, Machinery or Locales tabs, this cancels all previously running requests before starting new ones. This will reduce the load on the server and avoid mixed data issues in our tabs.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "abortcontroller-polyfill": "^1.3.0",
     "brace": "0.11.1",
     "connected-react-router": "4.5.0",
     "diff-match-patch": "1.0.4",
--- a/frontend/src/core/api/base.js
+++ b/frontend/src/core/api/base.js
@@ -1,5 +1,7 @@
 /* @flow */
 
+import 'abortcontroller-polyfill/dist/abortcontroller-polyfill-only';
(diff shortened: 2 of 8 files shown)


adngdb:
Ah, note to self: the `Abort` feature is [not supported](https://developer.mozilla.org/en-US/docs/Web/API/AbortController#Browser_compatibility) on all the targets of our support matrix, so I'll need to encapsulate our usage of it in something, to be safe.

adngdb:
> I also noticed moving between entities is much much slower than on Translate.Current. That however is definitely a separate issue.

Are you testing locally? If so, the local build has a lot of features that make it a lot slower than it actually is on stage/prod (logs, debug stuff, etc. ). My experience on stage has been quite good so far, but it's nothing more than a feeling. At some point we should perform actual performance tests.

adngdb:
> I'm not sure if that's related, but if I click the `Next` button around 20 times very fast on a view with FTL strings, I get the following error, which results in making the entire app unresponsive:

I suspect this is a problem in dev, as I cannot reproduce it on stage. Or it is a problem with Ace that doesn't support instantiating so many components so fast. It's out of scope and probably going to be very hard to debug, for something that, hopefully, won't ever hit our users (as I don't expect frantically moving to the next entity is something they would do very often ;-) ).

PR #1291, opened 2019-05-21
No bug - Fix build issue with pip in travis.

adngdb:
'tis a work in progress, I am clueless about this error so I'm trying things. Don't mind me.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: false
 language: python
 python:
-  - '2.7'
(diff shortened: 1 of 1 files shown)


adngdb:
First passing build: `Ran for 10 min 9 sec`. Re-running to see if the cache does something.

adngdb:
Second build: `Ran for 10 min`. Well I guess that makes our builds longer, but at least they pass without having to clear the cache every time.

PR #1292, opened 2019-05-21
Fix bug 1519894: Add Resource menu

mathjazz:
This patch exposes the name of the resource currently localized in the main navigation, and lists the entire batch of project resources with their completion percentages in the corresponding menu.

Selecting a resource in the menu hot-reloads the translate view. Before loading we check for any unsaved changes in the editor, which we now also do upon clicking on the resource path in the Metadata section.

--

I still need to implement search and write some tests, and then I believe navigation will be in a good enough state for Stage 2 testing. We'll still need to figure out what to do with "All Projects" afterwards, because right now that view hides the resource menu and makes it hard to switch to the regular locale - project - resource view.

@adngdb I'd appreciate your help with the "Can't load specified string" message that appears when switching resources. For some reason `connected-react-router`'s `push()` method preserves the `entity` param and I haven't yet figured out how to reset it.

Also, a quick high-level review would be nice. :)

### DIFF
--- a/frontend/src/core/resource/components/ResourceMenu.js
+++ b/frontend/src/core/resource/components/ResourceMenu.js
@@ -1,0 +1,164 @@
+/* @flow */
+
+import * as React from 'react';
+import { Localized } from 'fluent-react';
+import onClickOutside from 'react-onclickoutside';
+
+import './ResourceMenu.css';
+
+import type { NavigationParams } from 'core/navigation';
+import type { Resource, ResourcesState } from '..';
+
+
+type Props = {|
+    parameters: NavigationParams,
+    resources: ResourcesState,
+    navigateToPath: (string) => void,
+|};
+
+type State = {|
+    visible: boolean,
+|};
+
+
+/**
+ * Render a resource menu for the main navigation bar.
+ *
+ * Allows to switch between resources without reloading the Translate app.
+ */
+export class ResourceMenuBase extends React.Component<Props, State> {
+    constructor(props: Props) {
+        super(props);
+        this.state = {
+            visible: false,
+        };
+    }
+
+    toggleVisibility = () => {
+        this.setState((state) => {
+            return { visible: !state.visible };
+        });
+    }
+
+    // This method is called by the Higher-Order Component `onClickOutside`
+    // when a user clicks outside the user menu.
+    handleClickOutside = () => {
+        this.setState({
+            visible: false,
+        });
+    }
+
+    navigateToPath = (event: SyntheticMouseEvent<>) => {
+        event.preventDefault();
+
+        // $FLOW_IGNORE
+        const path = event.currentTarget.pathname;
+        this.props.navigateToPath(path);
+
+        this.setState({
+            visible: false,
+        });
+    }
+
+    renderPercent = (resource: Resource) => {
+        return Math.floor(resource.approved_strings / resource.total_strings * 100) + '%';
+    }
+
+    renderResource = (resource: Resource, i: number) => {
+        const { parameters } = this.props;
+
+        return <li
+            className={ parameters.resource === resource.path ? 'current' : null }
+            key={ i }
+        >
+            <a
+                href={ `/${parameters.locale}/${parameters.project}/${resource.path}/` }
+                onClick={ this.navigateToPath }
+            >
+                <span>{ resource.path }</span>
+                <span className="percent">{ this.renderPercent(resource) }</span>
+            </a>
+        </li>;
+    }
+
+    render() {
+        const { parameters, resources } = this.props;
+
+        if (parameters.project === 'all-projects') {
+            return null;
+        }
+
+        let className = 'resource-menu';
+        if (!this.state.visible) {
+            className += ' closed';
+        }
+
+        let resourceName = parameters.resource.split('/').slice(-1)[0];
+        if (resourceName === 'all-resources') {
+            resourceName = 'All Resources';
+        }
+
+        // Extract All Resources entry for more exposed presentation
+        const allResources = resources.resources.slice(-1)[0];
+
+        return <li>
+            <div className={ className }>
+                <div
+                    className="selector unselectable"
+                    onClick={ this.toggleVisibility }
+                    title={ parameters.resource }
+                >
+                    <span>{ resourceName }</span>
+                    <span className="icon fa fa-caret-down"></span>
+                </div>
+
+                { !this.state.visible ? null :
+                <div className="menu">
+                    <div className="search-wrapper">
+                        <div className="icon fa fa-search"></div>
+                        <input type="search" autoComplete="off" autoFocus />
+                    </div>
+
+                    <ul>
+                        { resources.resources.map((resource, i) => {
+                            // Skip All Resources entry for more exposed presentation below
+                            if (i === resources.resources.length - 1) {
+                                return null;
+                            }
+
+                            return this.renderResource(resource, i);
+                        }) }
+                    </ul>
+
+                    <ul className="static-links">
+                        <li className={ parameters.resource === 'all-resources' ? 'current' : null }>
+                            <a
+                                href={ `/${parameters.locale}/${parameters.project}/all-resources/` }
+                                onClick={ this.navigateToPath }
+                            >
+                                <Localized id='navigation-ResourceMenu-all-resources'>
+                                    <span>All Resources</span>
+                                </Localized>
+                                <span className="percent">{ this.renderPercent(allResources) }</span>
+                            </a>
+                        </li>
+                        <li>
+                            <a
+                                href={ `/${parameters.locale}/all-projects/all-resources/` }
+                                onClick={ this.navigateToPath }
+                            >
+                                <Localized id='navigation-ResourceMenu-all-projects'>
+                                    <span>All Projects</span>
+                                </Localized>
+                                <span className="percent"></span>
+                            </a>
+                        </li>
+                    </ul>
+                </div>
+                }
+            </div>
+        </li>;
+    }
+}
+
+export default onClickOutside(ResourceMenuBase);
--- a/frontend/src/core/resource/components/ResourceMenu.js
+++ b/frontend/src/core/resource/components/ResourceMenu.js
@@ -6,7 +6,8 @@ import onClickOutside from 'react-onclickoutside';
 
 import './ResourceMenu.css';
 
-import { ResourceItem, ResourcePercent } from '..';
+import ResourceItem from './ResourceItem.js';
+import ResourcePercent from './ResourcePercent.js';
 
 import type { NavigationParams } from 'core/navigation';
 import type { ResourcesState } from '..';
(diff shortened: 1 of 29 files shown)


frontend/src/modules/entitydetails/components/EntityDetails.js
```diff
@@ -187,6 +188,17 @@ export class EntityDetailsBase extends React.Component<InternalProps, State> {
...
 
+    navigateToPath = (path: string) => {
+        const { dispatch } = this.props;
+
+        dispatch(
+            unsavedchanges.actions.check(
+                this.props.unsavedchanges,
+                () => { dispatch(push(path)); }
+            )
+        );
```
adngdb: [rant] This pattern is repeated over a lot of places, and it makes me sad. I wished there was a way to refactor that so it's simpler to use (something like a Python decorator) but I don't know how that can be done. [/rant]

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -217,9 +225,12 @@ export default class Metadata extends React.Component<Props> {
             { this.renderSources(entity) }
             <Localized id='entitydetails-metadata-resource' attrs={ { title: true } }>
                 <Property title='Resource' className='resource'>
-                    <Link to={ `/${locale.code}/${entity.project.slug}/${entity.path}/` }>
+                    <a
```
adngdb: Why replace `Link` with `a`? If it's to use `onClick`, I'd be surprised if it didn't work with `Link`.

frontend/src/core/resource/components/ResourceMenu.js
```diff
@@ -0,0 +1,164 @@
...
+        this.setState({
+            visible: false,
+        });
+    }
+
+    renderPercent = (resource: Resource) => {
+        return Math.floor(resource.approved_strings / resource.total_strings * 100) + '%';
+    }
+
+    renderResource = (resource: Resource, i: number) => {
```
adngdb: I've actually come to think that if you create a `renderX` method, you actually want to have an `X` component. In this case, it would be perfect to use a simple function component (instead of a class), because the code is so small. I'm more and more thinking that we should drop the rule of using only class-based components and embrace function components and hooks. What do you think?

frontend/src/modules/entitydetails/components/EntityDetails.js
mathjazz: Yeah. :/ Or a component lifecycle method.

frontend/src/core/resource/components/ResourceMenu.js
mathjazz: Yeah I was thinking about a component and I thought it was an overkill, but that's before I've been introduced to the concept of functional components. :) I'll give it a try.

frontend/src/modules/entitydetails/components/Metadata.js
mathjazz: What's the benefit of using `Link` instead of `a`? I'm preventing default anyways (to check for unsaved changes).

mathjazz:
@adngdb I've addressed your comments, except the ones I commented on. Could you please have another look?

I've also:
* added the ability to search resources
* made percentages in the Resource menu update as you change translations
  * fixed two related bugs
* wrote some tests
* filed a bug for the ["Can't load specified string"](https://bugzilla.mozilla.org/show_bug.cgi?id=1553618) message (it's unrelated to this patch).

frontend/src/modules/entitydetails/components/Metadata.js
adngdb: To me, it helps understanding what the link is about. I'm using `<a>` when I link to an external resource (a different page, a different website) and `<Link>` when I link to an internal resource, meaning when it doesn't trigger a page reload. It doesn't make much difference but I think it makes the intent clearer.

mathjazz: Good point about semantics. I'll update.

mathjazz: Meh, the problem with `<Link>` is that it prepends `/translate` to the `href` attribute, and it doesn't allow me to navigate to the right URL without manually stripping `/translate`.

frontend/src/core/resource/index.js
```diff
@@ -0,0 +1,16 @@
+/* @flow */
+
+export { default as actions } from './actions';
+export { default as reducer } from './reducer';
+
+export { default as ResourceItem } from './components/ResourceItem';
+export { default as ResourceMenu } from './components/ResourceMenu';
+export { default as ResourcePercent } from './components/ResourcePercent';
```
adngdb: `ResourceItem` and `ResourcePercent` should not be exposed, they are not meant to be used outside of the `resource` module.

frontend/src/core/resource/components/ResourceMenu.js
```diff
@@ -0,0 +1,172 @@
+/* @flow */
+
+import * as React from 'react';
+import { Localized } from 'fluent-react';
+import onClickOutside from 'react-onclickoutside';
+
+import './ResourceMenu.css';
+
+import { ResourceItem, ResourcePercent } from '..';
```
adngdb: You should import from the files directly, and not expose these outside this module.

frontend/src/core/resource/components/ResourceMenu.js
```diff
@@ -0,0 +1,172 @@
...
+    handleClickOutside = () => {
+        this.setState({
+            visible: false,
+        });
+    }
+
+    navigateToPath = (event: SyntheticMouseEvent<>) => {
+        event.preventDefault();
+
+        // $FLOW_IGNORE
```
adngdb: Can you explain why you ignore Flow here?

frontend/src/core/resource/components/ResourceMenu.js
```diff
@@ -0,0 +1,172 @@
...
+        if (parameters.project === 'all-projects') {
+            return null;
+        }
+
+        let className = 'resource-menu';
+        if (!this.state.visible) {
+            className += ' closed';
+        }
+
+        let resourceName = parameters.resource.split('/').slice(-1)[0];
```
adngdb: I'm not sure I understand this. If there happens to be a resource `path/to/all-resources` then it will match the next condition, but I don't think it should. Why not just check if `parameters.resource` is equal to `all-resources`?

frontend/src/core/resource/components/ResourceMenu.js
```diff
@@ -0,0 +1,172 @@
...
+        if (resourceName === 'all-resources') {
+            resourceName = 'All Resources';
+        }
+
+        // Extract All Resources entry for more exposed presentation
+        const allResources = resources.resources.slice(-1)[0];
```
adngdb: I assume the API returns a list of resources that also contains `all-resources`? If that's the case, I'd remove it in the reducer instead of here.

frontend/src/core/resource/components/ResourceMenu.js
```diff
@@ -0,0 +1,172 @@
...
+
+        // Search resources
+        const search = this.state.search;
+        const resourceElements = resources.resources.filter(resource =>
+            typeof resource.path === 'string' &&  // Exclude All Resources
+            resource.path.toLowerCase().indexOf(search.toLowerCase()) > -1
+        );
+
+        return <li>
+            <div className={ className }>
```
adngdb: Is this `<div>` needed? Can you move the class to the `<li>` instead and move everything one level down?

frontend/src/core/resource/components/ResourceMenu.test.js
```diff
@@ -0,0 +1,87 @@
...
+        expect(wrapper.find('.resource-menu .menu')).toHaveLength(1);
+        expect(wrapper.find('.resource-menu .menu .search-wrapper')).toHaveLength(1);
+        expect(wrapper.find('.resource-menu .menu > ul')).toHaveLength(2);
+        expect(wrapper.find('.resource-menu .menu > ul').find(ResourceItem)).toHaveLength(3);
+        expect(wrapper.find('.resource-menu .menu .static-links')).toHaveLength(1);
+        expect(wrapper.find('.resource-menu .menu #navigation-ResourceMenu-all-resources')).toHaveLength(1);
+        expect(wrapper.find('.resource-menu .menu #navigation-ResourceMenu-all-projects')).toHaveLength(1);
+    });
+
+    it('renders resource menu correctly', () => {
```
adngdb: You have the same test name twice here. This one should talk about search.

frontend/src/core/resource/components/ResourcePercent.js
```diff
@@ -0,0 +1,25 @@
...
+type Props = {|
+    resource: Resource,
+|};
+
+
+/**
+ * Render a resource item percentage.
+ */
+export default function ResourcePercent(props: Props) {
+    const { approved_strings, strings_with_warnings, total_strings }  = props.resource;
```
adngdb: Double space before the `=` sign.

frontend/src/core/resource/reducer.js
```diff
@@ -0,0 +1,69 @@
...
+    approved_strings: number,
+    strings_with_warnings: number,
+): Array<Resource> {
+    return state.resources.map(item => {
+        if (item.path === resource_path) {
+            const allResources = state.resources.slice(-1)[0];
+
+            const diff_approved = approved_strings - item.approved_strings;
+            item.approved_strings += diff_approved;
+            allResources.approved_strings += diff_approved;
```
adngdb: `item` and `allResources` should be immutable. You should instead copy these objects first, and recreate a new list of resources based on them. Note that you don't need to copy the ones you do not modify.

adngdb: I am very disappointed that Flow didn't catch that. :-(

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -202,6 +202,14 @@ export default class Metadata extends React.Component<Props> {
         return this.renderSourceObject(entity.source);
     }
 
+    navigateToPath = (event: SyntheticMouseEvent<>) => {
+        event.preventDefault();
+
+        // $FLOW_IGNORE
```
adngdb: Please explain why you ignore Flow here.

frontend/src/core/resource/components/ResourceMenu.js
mathjazz: I don't think it's possible to circumvent the "Property pathname is missing in EventTarget [1]." Flow error. Is it?

frontend/src/modules/entitydetails/components/Metadata.js
mathjazz: See above.

frontend/src/core/resource/components/ResourceMenu.js
adngdb: Yeah, I guess so. But can you add a note to say that? :angel:

mathjazz: Oh! Well, I can actually get rid of it. :)

frontend/src/core/resource/components/ResourcePercent.js
adngdb: Still a double-space before the equal sign. :-)

adngdb: Forget that, I'm dumb.

PR #1296, opened 2019-05-23
Bug 1486510 - Show project info panel.

adngdb:
This is part 1 of the bug, only showing the info panel. It's already quite big so I figured I should split it.


frontend/src/modules/projectinfo/components/InfoPanel.css
```diff
@@ -0,0 +1,40 @@
+.info-panel {
+    font-size: 14px;
+}
+
+.info-panel .button {
+    cursor: pointer;
+    font-size: 1.4em;
+    font-weight: 900;
+    line-height: 16px;
```
mathjazz: 18px

frontend/src/modules/projectinfo/components/InfoPanel.css
```diff
@@ -0,0 +1,40 @@
...
+    height: 17px;
```
mathjazz: 18px

frontend/src/modules/projectinfo/components/InfoPanel.css
```diff
@@ -0,0 +1,40 @@
...
+    padding: 21px 7px;
```
mathjazz: 19px 7px 23px

frontend/src/App.js
```diff
@@ -38,6 +42,7 @@ type InternalProps = {
 class App extends React.Component<InternalProps> {
     componentDidMount() {
         this.props.dispatch(locales.actions.get());
+        this.props.dispatch(project.actions.get(this.props.parameters.project));
```
mathjazz: I'm curious if that's also where I should [load resources](https://github.com/mozilla/pontoon/pull/1292/files#diff-9d4d0fc9d1fef199d928b85a28631f6dR38)?

frontend/src/modules/projectinfo/components/InfoPanel.js
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    render() {
+        return <div className="info-panel">
+            <div className="button" onClick={ this.toggleVisibility }>
+                <span className="fa fa-info"></span>
+            </div>
+            { !this.state.visible ? null : <div className="panel">
+                <h2>Project Info</h2>
+                <p dangerouslySetInnerHTML={ { __html: this.props.info } } />
```
mathjazz: A comment would be nice:
https://github.com/mozilla/pontoon/blob/master/frontend/src/core/user/components/UserNotification.js#L79

frontend/src/modules/projectinfo/components/InfoPanel.js
```diff
@@ -0,0 +1,57 @@
...
+            visible: false,
+        });
+    }
+
+    render() {
+        return <div className="info-panel">
+            <div className="button" onClick={ this.toggleVisibility }>
+                <span className="fa fa-info"></span>
+            </div>
+            { !this.state.visible ? null : <div className="panel">
```
mathjazz: Nit: In other instances (except in `FiltersPanel.js`) we break into new line after `:`.

frontend/src/modules/projectinfo/components/ProjectInfo.js
```diff
@@ -0,0 +1,55 @@
...
+
+
+/**
+ * Render the current project's completion status and its info panel.
+ */
+export class ProjectInfoBase extends React.Component<InternalProps> {
+    render() {
+        if (
+            this.props.project.fetching ||
+            !this.props.project.name ||
```
mathjazz: You mean !this.props.project.**info** ||? There's no point in showing the icon and empty panel if there's no project info.

frontend/src/modules/projectinfo/components/ProjectInfo.js
```diff
@@ -0,0 +1,55 @@
...
+
+type Props = {|
+    parameters: NavigationParams,
+    project: ProjectState,
+|};
+
+type InternalProps = {|
+    ...Props,
+    dispatch: Function,
+|};
```
mathjazz: We don't need `InternalProps` here.

frontend/src/modules/projectinfo/components/ProjectInfo.js
```diff
@@ -0,0 +1,55 @@
...
+    render() {
+        if (
+            this.props.project.fetching ||
+            !this.props.project.name ||
+            this.props.parameters.project === 'all-projects'
+        ) {
+            return null;
+        }
+
+        return <div className="project-info">
```
mathjazz: We should be able to remove this `<div>` and join this component with `InfoPanel`.

ProgressBar should be a separate component and the two don't need a dedicated wrapper component. Semantically, they are different. Note that the ProgressBar represents the Resource progress, not Project progres.

frontend/src/App.js
adngdb: Yeah, I think it makes sense to make this the place where we load all initial data. It will also make it easier to optimize in the future (we might want to make all these calls into a single GraphQL query, and split the results afterwards).

frontend/src/modules/projectinfo/components/ProjectInfo.js
adngdb: I did mean `name` here, as my intent was just to check if the data hadn't been fetched yet, but you're right! It also makes the component even more self-contained. :)

frontend/src/modules/projectinfo/components/ProjectInfo.js
adngdb: Ah, I didn't know the progress bar was about the resource. I guess when we're on "all-resources" it still shows the project's data, though?

mathjazz: Yes, in this case it is. It's just a different and more visible presentation of numbers in the filter, that's why we'd like to join the two in the future somehow.

PR #1297, opened 2019-05-24
Fix bug 1553618 - Show the missing entity notification only on first …

adngdb:
…load.


frontend/src/modules/entitieslist/reducer.js
```diff
@@ -39,11 +39,11 @@ export type DbEntity = {
 export type Entities = Array<DbEntity>;
 
 // Read-only state (marked by '+').
-export type State = {
+export type EntitiesListState = {
     +entities: Entities,
     +fetching: boolean,
+    +fetchCount: number,
     +hasMore: boolean,
-    +errors: Array<string>,
```
adngdb: This was not being used anywhere.

PR #1298, opened 2019-05-25
Use correct url for jquery

Manishearth:
Really we should be using `{{settings.SITE_URL}}` here, but then we'd need to strip out the scheme.


mathjazz:
Welcome to Pontoon! :-)

PR #1299, opened 2019-05-26
Bug 1541253 - Remove pylama in favour of flake8

jotes:
`flake8` is the wrapper for a few existing linters:
* pyflakes
* pycodestyle
* mccabe

Additionally, it supports Python 3 and doesn't have issues with ignoring directories and files.
It was able to find a few small issues in Pontoon e.g. trailing spaces and so on.

Also, this change will fix the problems mentioned here: https://github.com/mozilla/pontoon/pull/1288#issuecomment-494589327

@mathjazz @adngdb thanks in advance for your review :-)


adngdb:
Woopsy, you had the wrong bug number there. It's actually [1541252](https://bugzilla.mozilla.org/show_bug.cgi?id=1541252).

PR #1302, opened 2019-05-27
No bug - Use aside instead of div for info panel.

adngdb:


adngdb:
Inspired by https://github.com/mozilla/pontoon/pull/1300 :-)

PR #1304, opened 2019-05-28
Format numbers in resource progress and filters

mathjazz:
I'm hardcoding `en-US` to make number formatting consistent with dashboards.

### DIFF
--- a/frontend/src/modules/resourceprogress/components/ResourceProgress.js
+++ b/frontend/src/modules/resourceprogress/components/ResourceProgress.js
@@ -52,6 +52,10 @@ export class ResourceProgressBase extends React.Component<Props, State> {
         if (total) {
             percent = Math.floor(complete / total * 100);
         }
+        // Do not show resource progress until stats are available
+        else {
+            return false;
+        }
 
         return <div className="progress-chart">
             <div className="selector" onClick={ this.toggleVisibility }>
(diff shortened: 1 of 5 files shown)


frontend/src/modules/resourceprogress/components/ResourceProgress.js
```diff
@@ -52,6 +52,10 @@ export class ResourceProgressBase extends React.Component<Props, State> {
         if (total) {
             percent = Math.floor(complete / total * 100);
         }
+        // Do not show resource progress until stats are available
+        else {
+            return false;
```
adngdb: You should return `null` here instead of false, that's the usual when you want to not render a component.

frontend/src/modules/resourceprogress/components/ResourceProgress.js
```diff
@@ -62,32 +66,38 @@ export class ResourceProgressBase extends React.Component<Props, State> {
             <aside className="menu">
                 <div className="main">
                     <header>
-                        <h2>All strings <span className="value">{ total }</span></h2>
-                        <h2 className="small">Unreviewed <span className="value">{ unreviewed }</span></h2>
+                        <h2>
+                            All strings
+                            <span className="value">{ Number(total).toLocaleString('en-US') }</span>
```
adngdb: That seems fair enough to me for now, but once we have support for more locales in Pontoon, I think I'd like to make this internationalized. Maybe we should file a P5 bug about this?

adngdb: Also, this pattern is repeated a lot, can you perhaps make a function of it? That will make a potential future refactoring easier.

adngdb: Put that function in `core/utils` since it is used across several files. :angel:

PR #1305, opened 2019-05-28
No bug - Split entitieslist module.

adngdb:
This moves the data part (reducer, selectors, actions) of the entitieslist module into a core/entities module, and moves the types into the api module. The goal is to make imports around the app more consistent (we should not import from the modules/ folder, only from core/). I also want to consolidate types for data that comes from the API into the api module, and this is one step towards that goal. This is purely a refactoring, no functionalities have changed.


adngdb:
Hold on. I might have made a mistake there. Actually, we have a choice to make. Do you prefer: 

```js
import type { Entity as DbEntity } from 'core/api/types';
const a: DbEntity;
```

Or:

```js
import api from 'core/api';
const a: api.types.Entity;
```

Both work, as far as I can tell. The former breaks module borders (as we import from `/types` instead of from the public interface). The second makes the type more tedious to type (`api.types.MyType`), and makes modules import `api` even if it's only used for types.

PR #1310, opened 2019-06-07
Fix bug 1552801 : [pre-translation] Implement background task.

VishalCR7:
Hey everyone!
I would like a functional review of the pre-translate task. The way it is done currently is: 
1. Add a `pre_translation_enabled` flag to Project for identification.
2. Once the pre-translation is enabled for a project, Identify the entities without any translations and use machine translations to store translations as suggestions. We do it as part of the save() function for the project.
3. New entities added to the projects are pre-translated as a part of the sync.

Please refer  the [doc](https://docs.google.com/document/d/1PK6Sz7Q_6F7SOYPbZRrI-vJQ73P1xjKIuwrRXGiUOS4/edit) for more details
I would like to know your thoughts about it.  Thanks in advance :)

PS: We are using TM for  100% match  and fall back to MT if no matches exist


pontoon/base/models.py
```diff
@@ -3326,3 +3334,73 @@ def calculate_stats(self, save=True):
             strings_with_warnings_diff,
             unreviewed_strings_diff,
         )
+
+
+from pontoon.machinery.views import translation_memory
+from pontoon.machinery.views import google_translate
+
+
+def pre_translate(project, locales=None, entities=None):
```
mathjazz: We should move this outside `pontoon.base.models`. I suspect we'll need a dedicated module, e.g. `pontoon.pre_translation`

pontoon/machinery/views.py
```diff
@@ -45,17 +45,18 @@ def machinery(request):
...
+    if request:
+        try:
+            text = request.GET['text']
+            locale = request.GET['locale']
+            pk = request.GET['pk']
+        except MultiValueDictKeyError as e:
+            return JsonResponse({
+                'status': False,
+                'message': 'Bad Request: {error}'.format(error=e),
+            }, status=400)
```
mathjazz: I don't like this too much. I'd keep the original function signature and then within it call `a_new_function(text, locale, pk)` that will also be used by `pre_translation`. @adngdb Do you agree?

VishalCR7:
Only work remaining was to make the build pass which was failing because of the way I imported the functions. Other than that it is complete and ready for the review. I will be addressing the comments you left pretty soon! :)

pontoon/machinery/views.py
adngdb: Yes, definitely. This is a django view, and should stay as such. If you want the insides to be shared, then extract the insides to a different function and call that function from the view.

adngdb: This way you also make the function call a lot less painful for the other places, because you don't have the JSON encoding logic and all.

pontoon/base/models.py
adngdb: Agreed. I don't much like the underscore in that name though, I'd simply call it `pretranslation`.

pontoon/base/models.py
```diff
@@ -56,6 +56,7 @@
 from pontoon.checks.utils import save_failed_checks
 from pontoon.db import IContainsCollate, LevenshteinDistance  # noqa
 from pontoon.sync import KEY_SEPARATOR
+from pontoon.pretranslation.utils import pre_translate
```
mathjazz: We should be consistent and drop underscore everywhere (i.e. `pretranslate`).

pontoon/pretranslation/utils.py
```diff
@@ -0,0 +1,70 @@
...
+            resource__project=project,
+            obsolete=False,
+            translation__isnull=True,
+        )
+
+    now = timezone.now()
+
+    resources = project.resources.filter(obsolete=False)
+
+    user = base.models.User.objects.get(pk=2)
```
mathjazz: That's too hardcoded. :) According to the spec, we should create a "Pretranslation" User.

pontoon/pretranslation/utils.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+
+    if not locales:
+        locales = project.locales.filter(project_locale__readonly=False)
+
+    if not entities:
+        entities = base.models.Entity.objects.filter(
+            resource__project=project,
+            obsolete=False,
+            translation__isnull=True,
```
mathjazz: That's going to exclude entities that have translations at least for one locale.

pontoon/pretranslation/utils.py
```diff
@@ -0,0 +1,70 @@
...
+                    string = gt_response['translation']
+
+            if string:
+                t = base.models.Translation(
+                    entity=entity,
+                    locale=locale,
+                    string=string,
+                    user=user,
+                    date=now,
+                    approved=False,
```
mathjazz: We also need to make translations `active`, otherwise they don't appear in the string list and editor.

pontoon/pretranslation/utils.py
```diff
@@ -0,0 +1,70 @@
...
+                )
+                t.save()
+                t.update_latest_translation()
```
mathjazz: We should do these actions in bulk, otherwise this is too slow.

pontoon/pretranslation/utils.py
VishalCR7: oops! :/

pontoon/pretranslation/utils.py
```diff
@@ -0,0 +1,122 @@
+from django.utils import timezone
+from django.db.models import CharField, Value as V
+from django.db.models.functions import Concat
+from bulk_update.helper import bulk_update
+
+import pontoon.base as base
+from pontoon.machinery.utils import google_translate_util, translation_memory_util
+# from pontoon.base.models import Entity, Locale, Translation, TranslatedResource, User
```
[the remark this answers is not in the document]
VishalCR7: As I told you earlier on IRC that using the pretranslate() in project.save() was causing circular imports. This is a workaround to be able to use it. Sadly, I Couldn't find anything better to make it work.

pontoon/machinery/views.py
```diff
@@ -67,31 +66,8 @@ def translation_memory(request):
...
-            entry['quality'] > suggestions[entry['target']]['quality']
-        ):
-            suggestions[entry['target']].update(entry)
-        suggestions[entry['target']]['count'] += 1
-
-    return JsonResponse(
-        sorted(suggestions.values(), key=lambda e: e['count'], reverse=True)[:max_results],
-        safe=False
-    )
+    data = translation_memory_util(text, locale, max_results, pk)
```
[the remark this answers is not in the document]
VishalCR7: :+1:

pontoon/pretranslation/utils.py
```diff
@@ -0,0 +1,122 @@
...
+    # bulk_update(locale_list, update_fields=['latest_translation'])
+    # bulk_update(projectlocale_list, update_fields=['latest_translation'])
+    # bulk_update(translatedresource_dict.values(), update_fields=['latest_translation'])
+
+    for translation in translations:
+        translation.update_latest_translation()
+
+    for translatedresource in translatedresource_dict.values():
+        # translatedresource.save(update_fields=['latest_translation'])
+        translatedresource.calculate_stats()
```
jotes: I don't have measurements/metrics, but every `calculate_stats` call takes some time (I think you should measure how long it takes to recalculate all stats for your test project/locale).
I don't want to overshoot with the optimization (@mathjazz?). If recalculation of the stats will be the slowest part of pre-translation, you can take a look at https://bugzilla.mozilla.org/show_bug.cgi?id=1344658.
The general idea for the optimization in this case would be to generate the difference between stats and later apply them to translated resources, instead of full recalculation.

mathjazz: OTOH, [sync](https://github.com/mozilla/pontoon/blob/2346e09c916b42c628bed59f4a8fd6e03a33a18d/pontoon/sync/core.py#L219) just calls calculate_stats on each changed resource, too.

We could however use something like [batch_actions](https://github.com/mozilla/pontoon/blob/2346e09c916b42c628bed59f4a8fd6e03a33a18d/pontoon/batch/views.py#L39) do.

VishalCR7: This was already on my bucket list for the weekend and I was going to discuss this in #pontoon Thanks a lot for the heads up @jotes.

pontoon/pretranslation/utils.py
```diff
@@ -0,0 +1,122 @@
...
+
+    # for locale in locale_list:
+    #     locale.save(update_fields=['latest_translation'])
+
+    # for projectlocale in projectlocale_list:
+    #     projectlocale.save(update_fields=['latest_translation'])
+
+    # bulk_update(locale_list, update_fields=['latest_translation'])
+    # bulk_update(projectlocale_list, update_fields=['latest_translation'])
+    # bulk_update(translatedresource_dict.values(), update_fields=['latest_translation'])
```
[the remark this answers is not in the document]
VishalCR7: It was being done in the loop. looks like I deleted the code. It was something like this : 
https://gist.github.com/VishalCR7/2f003c9f816ca07674be533f0fa405dc 
@jotes  any way it might still work?

VishalCR7:
Hey @mathjazz  @jotes  I have addressed almost all of the review comments. Fortunately, I m able to do most of the database operations in bulk. Although the code might look ugly, I would like some feedback about it. 

Also, following numbers compare the previous (for loops with function calls) and the current approach for a limited number of strings and this seems to work fine.

Previous ones (looping over each translation and translatedResource)
```
----------Update_latest_activity----------
No of translations: 259
Time: 2.53836798668
-------------------------------------------
----------calculate_stats-----------------
No of TranslatedResources: 49
1.84736204147
------------------------------------------
```

With Bulk Update
```
----------Total-Activity+Stats------------
No of translations:259
No of Resources: 49
Time: 0.0405910015106
------------------------------------------
```

pontoon/pretranslation/utils.py
```diff
@@ -0,0 +1,162 @@
...
+
+    translated_entities = list(translated_entities)
+
+    now = timezone.now()
+
+    user, _ = base.models.User.objects.get_or_create(
+        email="pontoon@mozilla.com",
+        username="pontoon-translator",
+        first_name="Pontoon Translator",
+    )
```
mathjazz: We should create this user with a data migration. I believe however we should create separate users for MT and TM though (`first_name="Google Translate"` & `first_name="Translation Memory"`), because Google Translate API requires [attribution](https://cloud.google.com/translate/attribution#attribution_and_logos).

pontoon/pretranslation/utils.py
```diff
@@ -58,17 +64,29 @@ def pretranslate(project, locales=None, entities=None):
...
+                tm_response = sorted(tm_response, key=lambda e: e['plural_form'])
+
+                if tm_response:
+                    if entity.string_plural == "":
+                        strings = [(tm_response[0]['target'], None)]
+                    else:
+                        for plural_form in plural_forms:
+                            plurals = [t for t in tm_response if t['plural_form'] == plural_form]
+                            if plurals:
+                                strings.append(((plurals[0]['target'], plural_form)))
```
mathjazz: I don't understand how this works. You only make one request to TM (with singular string as text input), but then you use it for plural translations, in case translation (of a singular string) is stored as pluralized.

I think you should simplify the algorithm and use the same logic for MT:
1. Get TM entry for the singular string (`text=entity.string`).
1. `if entity.string_plural != ""`, also get TM entry for the pluralized string (`text=entity.string_plural`).
1. Assign `#1` to the first `plural_form` of the locale.
1. Assign `#1` to all remaining `plural_forms` of the locale.

adngdb: As a general rule, when I have to write a complex algorithm like this one, I tend to start by writing comments describing what it does, then I break it down into functions. 

This here is very difficult to read, and thus difficult both to understand and to maintain. I highly recommend that you break it down into smaller bits of code. It's fine to have multiple small functions in your code. It will also make your code easier to reason about, and might help you address the issues Matjaz pointed.

Also, this is in `utils.py`. That name is a convention that I dislike, as it is very unclear what its content is. For this `pretranslate` function, I would actually put it in a file that tells exactly what it contains (calling the file `pretranslate.py` is fine). This way you know that all functions in that file are strictly related to the pretranslate work, and it will be a lot easier for developers to understand and reason about this module.

pontoon/pretranslation/utils.py
```diff
@@ -77,38 +95,52 @@ def pretranslate(project, locales=None, entities=None):
...
+                            gt_response = get_google_translate_data(
+                                text=entity.string,
+                                locale_code=locale.google_translate_code,
+                            )
+                            if 'translation' in gt_response.keys():
+                                strings.append((gt_response['translation'], plural_forms[1]))
+
+                if strings:
+                    for string, plural_form in strings:
+                        print string
```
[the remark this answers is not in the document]
VishalCR7: I think I removed it :thinking:  in `2e5001e`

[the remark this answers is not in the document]
mathjazz: No. You removed it in https://github.com/mozilla/pontoon/pull/1310/commits/ffe80ad43a5c941e1fb9e2be1a7f991039ec8538. 🤣

VishalCR7:
@mathjazz , @adngdb  please review :)

pontoon/pretranslation/tasks.py
```diff
@@ -0,0 +1,115 @@
...
+            if locale_entity not in translated_entities:
+                strings = get_translations(entity, locale)
+                if strings:
+                    for string, plural_form, user in strings:
+                        t = base.models.Translation(
+                            entity=entity,
+                            locale=locale,
+                            string=string,
+                            user=user,
+                            date=now,
```
mathjazz: That means all translations across all locales and entities will have the same timestamp. Which will also be reflected in the latest activity column. Is that intentional?

`Translate.date` sets the default value to the actual `timezone.now`, why don't you just keep that (by not setting the date here at all)?

pontoon/pretranslation/tasks.py
```diff
@@ -0,0 +1,115 @@
...
+                        locale_dict[locale.code][2] += 1
+        log.info('Fetching pretranslations for locale {} done'.format(locale.code))
+
+    if len(translations) == 0:
+        return
+
+    translations = base.models.Translation.objects.bulk_create(translations)
+
+    # Update latest activity and unreviewed count for the project.
+    project.latest_translation = translations[0]
```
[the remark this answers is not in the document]
VishalCR7: Since the time-stamps are the same and we are using bulk_create it shouldn't  matter

pontoon/pretranslation/tasks.py
VishalCR7: We will leave it default and I will change the latest translation to last one as well

pontoon/pretranslation/tasks.py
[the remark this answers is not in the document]
mathjazz: As of https://github.com/mozilla/pontoon/pull/1310/commits/78567d76f38160f8a18326b18085190a2467682e, timestamps are no longer the same.

mathjazz:
Thanks for the update! The code looks good, but I haven't tested it yet. Could you please address https://github.com/mozilla/pontoon/pull/1310#discussion_r332375966 before I take it for another spin?

mathjazz:
Thanks, all my comments are addressed and everything works as expected in my testing scenario:
https://github.com/mozilla/pontoon/pull/1310#pullrequestreview-298573961

I'm passing this to @adngdb to have another look, at least in the two files of the pretranslate module, because he had opinions there (on naming for example). Once that's done, we'll also need some unit tests for these files.

Before merging, let's also fix https://github.com/mozilla/pontoon/pull/1310#discussion_r332366888.

pontoon/pretranslation/tests/test_tasks.py
```diff
@@ -0,0 +1,125 @@
...
+
+    # 100% match does not exists and locale.google_translate_code is not None.
+    response = get_translations(entities[2], google_translate_locale)
+    assert response == [(entities[2].string, None, gt_user)]
+
+    # Entity.string_plural is not None.
+    response_a = get_translations(entities[1], google_translate_locale)
+    response_b = get_translations(entities[3], google_translate_locale)
+    assert response_b == [(entities[2].string, 0, gt_user), (entities[2].string, 1, gt_user)]
+    assert response_a == [(entities[1].string, 0, tm_user), (entities[1].string, 1, tm_user)]
```
mathjazz: Can you switch the order of these two lines, please?

pontoon/pretranslation/tests/test_tasks.py
```diff
@@ -0,0 +1,125 @@
...
+        TranslationMemoryFactory.create(
+            entity=entity,
+            source=entity.string,
+            target=entity.string,
+            locale=google_translate_locale,
+        )
+    # Mock the return value of get_google_translate_data
+    gt_mock.return_value = {
+        'status': True,
+        'translation': entities[2].string,
```
mathjazz: Wouldn't it be less confusing if you used `gt_translation` here instead of `entities[2].string`? It took me a while to figure out `response_b` in the last test case.

PR #1311, opened 2019-06-08
Fix ISO date format for notifications data and translation history

MikkCZ:
The format of `date_iso` in notifications data returned by `/user-data/` endpoint is not valid. It contains duplicated timezone offset. The same is in `/get-history/`.

This should fix it and also avoid the [parsing magic](https://github.com/nmn/react-timeago/blob/v4.4.0/src/dateParser.js) which happens in _react-timeago_ [when `date` is a string](https://github.com/nmn/react-timeago/blob/v4.4.0/src/index.js#L187). They only [test with `Date` objects](https://github.com/nmn/react-timeago/blob/v4.4.0/__tests__/index.js), so let's use them too.

### DIFF
--- a/frontend/src/core/user/components/UserNotificationsMenu.test.js
+++ b/frontend/src/core/user/components/UserNotificationsMenu.test.js
@@ -68,7 +68,7 @@ describe('<UserNotificationsMenuBase>', () => {
                     description: 'description',
                     verb: 'verb',
                     date: 'Jan 31, 2000 10:20',
-                    date_iso: '2019-01-31T10:20:00+00:00+0000',
+                    date_iso: '2019-01-31T10:20:00+00:00',
                     actor: {
                         anchor: 'actor_anchor',
                         url: 'actor_url',
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -374,7 +374,6 @@ def get_translation_history(request):
     translations = translations.order_by('-active', 'rejected', '-date')
 
     payload = []
-    offset = timezone.now().strftime('%z')
 
     for t in translations:
         u = t.user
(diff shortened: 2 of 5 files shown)


MikkCZ:
Note: I am currently not sitting at a computer where I could run tests. So hopefully they will pass on Travis. If not, I will revisit this PR later.

pontoon/base/views.py
```diff
@@ -384,7 +383,7 @@ def get_translation_history(request):
             "uid": "" if u is None else u.id,
             "username": "" if u is None else u.username,
             "date": t.date.strftime('%b %d, %Y %H:%M'),
-            "date_iso": t.date.isoformat() + offset,
+            "date_iso": t.date.isoformat(),
```
mathjazz: This part of the changeset affects Translate.Current as well. I haven't tested the code yet, but I wonder if we need to make changes to `translate.js` as well.

frontend/src/core/user/components/UserNotificationsMenu.test.js
```diff
@@ -68,7 +68,7 @@ describe('<UserNotificationsMenuBase>', () => {
                     description: 'description',
                     verb: 'verb',
                     date: 'Jan 31, 2000 10:20',
-                    date_iso: '2019-01-31T10:20:00+00:00+0000',
+                    date_iso: '2019-01-31T10:20:00+00:00',
```
mathjazz: Nice catch! :)

pontoon/base/views.py
MikkCZ: The only place I see a usage in `translate.js` is to fill [`<time datetime="...">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time#Valid_datetime_Values), which is never displayed to the user.

MikkCZ: Btw. latest activity tag on dashboards uses just `isoformat()`: `<time datetime="{{ latest_activity.date.isoformat() }}" ...>`.

PR #1315, opened 2019-06-11
Fix bug 1558465: Do not break sync if compare-locale parser not found

mathjazz:
I'm also taking this opportunity to update Django.


pontoon/sync/formats/compare_locales.py
```diff
@@ -42,7 +42,12 @@ def __init__(self, path, source_resource=None):
         self.path = path
         self.entities = OrderedDict()  # Preserve entity order.
         self.source_resource = source_resource
-        self.parser = parser.getParser(self.path)
+
+        try:
+            self.parser = parser.getParser(self.path)
+        except UserWarning as err:
+            raise ParseError(err)
```
mathjazz: As you can see from the first line of [the traceback](https://bugzilla.mozilla.org/show_bug.cgi?id=1558465#c0), raising a `ParseError` here will result in [skipping the file](https://github.com/mozilla/pontoon/blob/master/pontoon/sync/vcs/models.py#L364).

mathjazz:
Thanks!

We don't use `AdminURLFieldWidget` and it works fine locally.

PR #1316, opened 2019-06-11
No bug - Upgrade react-scripts to version 3.0.1.

adngdb:


adngdb:
This does not solve any problems but it's generally good to have anyway. :-)

PR #1320, opened 2019-06-12
Fix bug 1555924: Improve layout on mobile

mathjazz:
Entity details panel will now break less on narrow mobile screens.

The patch includes some other CSS improvements in `EntityDetails`.

### DIFF
--- a/frontend/src/modules/editor/components/Editor.css
+++ b/frontend/src/modules/editor/components/Editor.css
@@ -1,23 +1,18 @@
 .editor {
     background: rgb(39, 42, 47);
-    border-bottom: 1px solid #5E6475;
     display: flex;
     flex-direction: column;
-    height: calc(20% + 60px);
-    min-height: 160px;
+    height: 20%;
+    min-height: 100px;
 }
 
 .editor textarea {
     border: none;
     box-sizing: border-box;
-    display: block;
     line-height: 22px;
-    height: calc(100% - 60px);
-    min-height: 100px;
-    overflow: auto;
+    height: 100%;
     padding: 10px;
     resize: none;
-    width: 100%;
 }
 
 /* Remove highlight in Chrome */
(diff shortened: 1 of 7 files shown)


frontend/src/modules/editor/components/Editor.js
```diff
@@ -163,25 +163,27 @@ export class EditorBase extends React.Component<InternalProps> {
             return null;
         }
 
-        return <div className="editor">
+        return <React.Fragment>
```
adngdb: I don't like that, it makes this component not encapsulated. I'd rather keep the `editor` class at the top, and add another class to the underlying editor, rather than doing this. That also conflicts a lot with my current refactoring, and the code I have will make this easier to do because I've renamed a few components. If you could keep the changes to a minimum here that would be very helpful to me. :angel:

frontend/src/modules/editor/components/Editor.css
```diff
@@ -37,26 +32,25 @@
...
     line-height: 40px;
 }
 
-.editor .banner a {
+.banner a {
     color: #7BC876;
 }
 
-.editor menu {
+.editor-menu {
```
adngdb: I'd rather use `.editor > menu` if you want to enforce a stricter rule. I think we should not abuse CSS classes when they're not needed.

frontend/src/modules/editor/components/Editor.css
```diff
@@ -37,26 +32,25 @@
     font-size: 15px;
 }
 
-.editor .banner {
+.banner {
```
adngdb: This is uncontained and might conflict with other components. :(

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -30,8 +30,11 @@ class Property extends React.Component<PropertyProps> {
     render(): React.Node {
         const { children, className, title } = this.props;
         return <p className={ className }>
-            <span className="title">{ title }</span>
-            <span className="content">{ children }</span>
+            {
+                /* Space between <span> elements also prevents cross
+                   element selection on double click (bug 1228873) */
+            }
+            <span className="title">{ title }</span> <span className="content">{ children }</span>
```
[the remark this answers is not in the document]
mathjazz: Thanks, that looks much nicer!

frontend/src/modules/editor/components/Editor.js
mathjazz: OK. In this case the menu will have to break slightly out of boundaries on narrower screens and we'll postpone making it prettier after your refactor or even until we make the UI responsive.

PR #1332, opened 2019-07-04
Fix bug 1555083: Prevent duplicate helpers requests

mathjazz:
I'd like to add a comment in comment in `componentDidUpdate` to explain why do we need the `selectedEntity !== prevProps.selectedEntity` check. But I forgot the answer. :) Could you please remind me?

Note that I've found some unused code and inconsistencies, which are addressed in followup commits.

### DIFF
--- a/frontend/src/modules/entitydetails/components/EntityDetails.js
+++ b/frontend/src/modules/entitydetails/components/EntityDetails.js
@@ -81,8 +81,8 @@ export class EntityDetailsBase extends React.Component<InternalProps, State> {
     componentDidUpdate(prevProps: InternalProps) {
         const { parameters, pluralForm, selectedEntity, activeTranslation } = this.props;
 
+        // We should not use parameters.entity here, because it's already set from the URL
         if (
-            parameters.entity !== prevProps.parameters.entity ||
             pluralForm !== prevProps.pluralForm ||
             selectedEntity !== prevProps.selectedEntity
         ) {
--- a/frontend/src/modules/entitydetails/components/EntityDetails.js
+++ b/frontend/src/modules/entitydetails/components/EntityDetails.js
@@ -79,9 +79,8 @@ export class EntityDetailsBase extends React.Component<InternalProps, State> {
     }
 
     componentDidUpdate(prevProps: InternalProps) {
-        const { parameters, pluralForm, selectedEntity, activeTranslation } = this.props;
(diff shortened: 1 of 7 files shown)


frontend/src/modules/entitydetails/components/EntityDetails.js
```diff
@@ -92,18 +92,28 @@ export class EntityDetailsBase extends React.Component<InternalProps, State> {
...
+        if (selectedEntity.pk !== this.props.otherlocales.entity) {
+            dispatch(otherlocales.actions.get(parameters.entity, parameters.locale));
+        }
 
-        const source = utils.getOptimizedContent(selectedEntity.original, selectedEntity.format);
-        dispatch(machinery.actions.get(source, locale, selectedEntity.pk));
+        if (selectedEntity.pk !== this.props.machinery.entity) {
+            const source = utils.getOptimizedContent(selectedEntity.original, selectedEntity.format);
+            dispatch(machinery.actions.get(source, locale, selectedEntity.pk));
+        }
```
mathjazz: Or maybe I could just add one big `if` and return in case any of the helpers has old data? It's unlikely they will have data for different entities anyways.

mathjazz: Yeah, especially now that I added the `pluralForm` check, it makes no sense to make that change.

frontend/src/modules/history/reducer.js
```diff
@@ -58,18 +61,18 @@ export default function reducer(
             return {
                 ...state,
                 fetching: true,
+                entity: null,
+                pluralForm: null,
```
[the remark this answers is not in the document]
mathjazz: Good point!

mathjazz:
Updated and removed what indeed seems to be a redundant condition.

frontend/src/modules/entitydetails/components/EntityDetails.js
```diff
@@ -81,8 +81,8 @@ export class EntityDetailsBase extends React.Component<InternalProps, State> {
     componentDidUpdate(prevProps: InternalProps) {
         const { parameters, pluralForm, selectedEntity, activeTranslation } = this.props;
 
+        // We should not use parameters.entity here, because it's already set from the URL
```
adngdb: I don't think that comment is useful anymore, because in a month both of us will have forgotten the context. :P 

Also, you can remove `parameters` from the declared variables I think.

PR #1334, opened 2019-07-05
Fix bug 1563677: Escape selectors in data-expression attributes

mathjazz:
Translating a string like this in the rich editor is not possible - it results in an `Expected literal` error upon saving:
```
number =
    { NUMBER($var, type: "ordinal") ->
        [1] first
        [one] { $var }st
       *[other] { $var }nd
    }
```

**What's going on?**

Pontoon doesn't show selectors (in this case `NUMBER($var, type: "ordinal")`) in the rich editor, but still stores them in DOM to help with serialization. Specifically, it naïvely puts them in `data-expression` attributes, without any escaping.

So when the selector contains quotes, like in our example, content gets stripped upon serialization, resulting in the above error. The serialized string looks like this:

```
number =
    { NUMBER($var, type:  ->
        [1] first
        [one] { $var }st
       *[other] { $var }nd
    }
```

A test case:
https://github.com/mozilla-l10n/pontoon-ftl/blob/master/en-US/demo.ftl#L53

### DIFF
--- a/pontoon/base/static/js/fluent_interface.js
+++ b/pontoon/base/static/js/fluent_interface.js
@@ -263,7 +263,7 @@ var Pontoon = (function (my) {
         if (element.expression.selector) {
           expression = FluentSyntax.serializeExpression(element.expression.selector);
         }
-        content += '<li data-expression="' + expression + '"><ul>';
+        content += '<li data-expression="' + escape(expression) + '"><ul>';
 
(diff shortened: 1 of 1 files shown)


jamesking:
My colleague @elisehein would be best placed to do a review :)

mathjazz:
Thanks for the review, I'm glad this is helpful! :)

PR #1335, opened 2019-07-12
Lodash

mathjazz:
This patch reverts the two failed attempts of removing the outdated `lodash.template` version from `yarn.lock`.

I haven't found a good way of updating an indirect dependency to its latest version, and I'm [not sure there is one](https://github.com/yarnpkg/yarn/issues/4986). Since `yarn upgrade lodash.template` and `yarn upgrade lodash.template --latest` didn't make any changes, I ended up running `yarn upgrade` and manually copying the `lodash.template` block from the updated to the existing `yarn.lock` file.

Very ugly, but at least it works locally and on stage.


mathjazz:
Maybe I should use [selective dependency resolutions](https://yarnpkg.com/lang/en/docs/selective-version-resolutions/) instead? It's still ugly, but less.

mathjazz:
After talking to @adngdb, we should only ever update our direct dependencies. Hence I'm removing the last commit.

`lodash.template` is required by two different packages, both of which are sub-dependencies of `react-scripts`. Dependency chart looks like this:

```
react-scripts -> workbox-webpack-plugin -> workbox-build -> lodash.template
react-scripts -> postcss-preset-env -> postcss-initial -> lodash.template
```

So far only `postcss-preset-env` and `postcss-initial` have released new versions to require the newest `lodash.template`. @adngdb Should we just open an issue in [`react-scripts`](https://github.com/facebook/create-react-app/) to require newest `postcss-preset-env`?

adngdb:
@mathjazz I had a hunch that this issue would not be affecting us, because `react-script` is a build tool and not runtime, it seems gaeron confirms that here: https://github.com/facebook/create-react-app/issues/6928#issuecomment-487218069

I don't think there's any emergency here for us, except hiding that error message on github. It might be worth filing an issue for `react-scripts` though.

mathjazz:
Good point. I guess we should still merge this PR?

PR #1336, opened 2019-07-15
Fix bug 1511357: Highlight search terms in entity list

mathjazz:
This works pretty much the same way as on Translate.Current, sharing most of the code, but replacing the `mark.js` with a similar React lib.

However, marking placeables and highlighting search terms is currently exclusive. Ideas on how to make these two functionalities play with one another welcome!


adngdb:
Can you not use `react-content-marker` instead of adding a new library? You can create a custom parser dynamically, based on the search words, and use that to mark the content. If you do that, then it becomes quite easy to mix it with placeables (it would work just like we did to mix placeables and diff).

frontend/src/modules/search/withSearch.js
```diff
@@ -0,0 +1,56 @@
...
+        const rules = searchWords.map(function(searchWord) {
+            return {
+                rule: new RegExp(searchWord, 'i'),
+                tag: searchWord => <mark className='search'>{ searchWord }</mark>,
+            };
+        });
+
+        const WithSearch = createMarker(rules);
+
+        return <WithSearch>{ base }</WithSearch>;
```
adngdb: You don't want to return a React Component instance here, because that won't be parsable by the next `react-content-marker` parser. Instead you want this function to return an array of strings and elements, basically the result of the `mark` function.

However, I don't think that it is accessible from the module. I'll have to release a new version of the library, will do that shortly.

frontend/src/modules/search/withSearch.js
```diff
@@ -0,0 +1,56 @@
...
+type Props = {
+    search: string,
+};
+
+
+export default function withSearch<Config: Object>(
+    WrappedComponent: React.AbstractComponent<Config>
+): React.AbstractComponent<Config> {
+    return function WithSearch(props: { ...Config, ...Props }) {
+        const base = props.children;
```
adngdb: I think this is more confusing than just using `props.children` directly in the call to `highlightSearch`. I generally dislike creating a variable if that is used only once, unless it really increases readability.

frontend/src/modules/search/withSearch.js
adngdb: See https://github.com/mozilla/react-content-marker/pull/3

frontend/src/core/translation/components/TranslationProxy.js
```diff
@@ -10,21 +10,30 @@ type Props = {|
     content: ?string,
     diffTarget?: ?string,
     format: string,
+    search?: ?string,
```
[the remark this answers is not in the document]
mathjazz: Bad copy-paste from `diffTarget` :)

adngdb:
@mathjazz I'm not sure you meant to add `mark.js` here, did you? Also, `react-content-marker` v1.1.0 has been released, and it exposes `mark`: https://www.npmjs.com/package/react-content-marker

mathjazz:
I might. I'm just playing with it and see if we can use it the we we use it in `translate.js`. We probably won't need `react-content-marker` in case it'll work. :/ But I think it's still good to have `mark()` function exposed.

mathjazz:
Oh, I see I already pushed `mark.js` dependency. Please ignore that for now.

frontend/src/modules/search/withSearch.js
```diff
@@ -0,0 +1,52 @@
...
+            return b.length - a.length;
+        });
+
+        return searchWords.map(function(searchWord) {
+            return mark(
+                searchWord,
+                new RegExp(searchWord, 'i'),
+                searchWord => <mark className='search'>{ searchWord }</mark>,
+            );
+        });
```
mathjazz: This is returning an array of functions (of type `ContentMarker`) for a reason I don't understand.

mathjazz: @adngdb I don't understand why https://github.com/mozilla/react-content-marker/blob/master/src/mark.js would return `ContentMarker`. Any hints?

frontend/src/modules/search/withSearch.js
```diff
@@ -1,7 +1,7 @@
 /* @flow */
 
 import * as React from 'react';
-import createMarker from 'react-content-marker';
+import mark from 'react-content-marker';
```
adngdb: `import { mark } from 'react-content-marker';`

Otherwise you're just importing the same thing (the default export) under a different name. :-)

mathjazz: Of course :)

mathjazz:
This is supposed to work now. It fixes the bug I think, but doesn't have the powers of Translate.Current. To get there, we have two options:
1. Improve `react-content-marker` to support search highlights within placeables and highlights spanning [across elements](https://github.com/julmot/mark.js/blob/9cf9c85f1ccf90fe988f19182329e24221b56d43/src/lib/mark.js).
2. Reactify [`mark.js`](https://markjs.io/), whose API demands DOM elements, which we cannot use easily with our implementation.

adngdb:
@mathjazz I think it's fine to lose some features in the transition to Translate.Next. I don't think it's realistic to go to that big a length (to me both options you propose will require heavy work) for such a small feature. I propose we go with option "3. Accept that Translate.Next will be less good in this area".

mathjazz:
True, but I'd like to be careful when dropping features.

The motivation behind Translate.Next is to be able to develop new things (on top of Translate.Current) faster, and now already existing features require that much more time to implement in Translate.Next. But I guess that's expected - one tech stack will always be better in something than another.

Let's make this exception.

adngdb:
Right, and there's always going to be the next tough thing to do, of course. I think that's why we need to constantly refactor, so that we keep a code base that is as maintained as possible at all time. But in this case, I think it's more of a technical side-effect than an actual feature drop. I assume creating the same feature with React is definitely doable, but the technology makes it really much harder than with simple text + DOM manipulation. This I believe is one of the few trade-offs we need to make for the sake of a better stack overall.

PR #1337, opened 2019-07-16
Fix bug 1565674: Add note on locale folder names and symlinks

mathjazz:


docs/user/localizing-your-projects.rst
```diff
@@ -39,7 +39,10 @@ structure as expected by Pontoon:
 2. Source locale needs to be called ``templates``, ``en-US``, ``en-us`` or
    ``en``. If multiple folders with such name exist in the repository and
    contain files in a supported file format, the first one will be used.
-3. Locale code must not be part of the file name.
+3. Locale folder names must always match `BCP 47`_ locale identifiers used by.
```
flodolo: How is this helping? I believe `fy` is valid identifier as much as `fy-NL`.

mathjazz: The crucial piece is "used by Pontoon". But yeah, linking to that wiki page is then misleading.

I guess I should simply leave out the `BCP 47` part?

flodolo: Yes, just "locales used by Pontoon". There's no enforcing a standard when you create a locale in Pontoon, isn't it?

mathjazz: Updated.

PR #1338, opened 2019-07-22
Fix bug 1486518: Add styled and localizable batch editor

mathjazz:
WIP: The batch editor is always on and doesn't do anything. The single entity editor is inaccessible.


frontend/public/static/locale/en-US/translate.ftl
```diff
@@ -84,6 +84,33 @@ editor-keyboard-shortcuts-copy-from-helpers = Copy From Helpers
...
+
+batcheditor-BatchEditor--warning-title = Warning:
+batcheditor-BatchEditor--warning-content = These actions will be applied to all selected strings and cannot be undone.
+
+batcheditor-BatchEditor--review-heading = Review translations
+batcheditor-BatchEditor--approve-all-button = Approve all
+batcheditor-BatchEditor--reject-all-button = Reject all unreviewed suggestions
+
+batcheditor-BatchEditor--find-replace-heading = Find & Replace in translations
+batcheditor-BatchEditor--find
```
adngdb: Missing an equal sign here.

frontend/public/static/locale/en-US/translate.ftl
```diff
@@ -84,6 +84,33 @@ editor-keyboard-shortcuts-copy-from-helpers = Copy From Helpers
...
+    .placeholder = Find
+batcheditor-BatchEditor--replace-with
```
adngdb: And here as well.

frontend/src/App.js
```diff
@@ -15,6 +15,7 @@ import * as project from 'core/project';
 import * as resource from 'core/resource';
 import * as stats from 'core/stats';
 import { UserControls } from 'core/user';
+import { BatchEditor } from 'modules/batcheditor';
```
adngdb: I think it will be confusing to call this an "editor", as it will overlap with the `genericeditor` and `fluenteditor` modules. If I were to read the list of modules I would assume `batcheditor` was the same kind of module as the other two, which it isn't. Maybe change it to `batchactions`?

frontend/src/modules/batcheditor/components/BatchEditor.js
```diff
@@ -0,0 +1,121 @@
...
+            <div className="main-content">
+                <div className="intro">
+                    <p>
+                        <Localized id="batcheditor-BatchEditor--warning-title">
+                            <span className="warning">Warning:</span>
+                        </Localized>
+                        { ' ' }
+                        <Localized id="batcheditor-BatchEditor--warning-content">
+                            <span className="content">These actions will be applied to all selected strings and cannot be undone.</span>
+                        </Localized>
```
adngdb: Why is this split into 2 strings? I think it would make more sense to have only one, and use some tag to mark the `Warning:` part, maybe like `<mark>` maybe? Otherwise this structure will likely create issues in RTL languages, or other things I can't think of.

mathjazz:
@adngdb I've now added the [ability to select (and unselect) entities for batch editing](https://github.com/mozilla/pontoon/pull/1338/commits/7f13a3b7f12bd49c53632f6c49393e612ce41401) by clicking on their status icon. Performing actions, Select All and Keyboard shortcuts are still missing, but the batch editor now opens and closes when it should.

I've also made some Chrome-related CSS fixes in a separate commit.

I'm mostly wondering what do you think about making the `Entity.js` a connected Component.

frontend/src/modules/batchactions/actions.js
```diff
@@ -0,0 +1,32 @@
...
+        type: RESET,
+    };
+}
+
+
+export type ToggleAction = {|
+    type: typeof TOGGLE,
+    entity: number,
+|};
+export function toggle(entity: number): ToggleAction {
```
adngdb: `toggleSelection` maybe?

frontend/src/modules/entitieslist/components/Entity.js
```diff
@@ -83,21 +90,52 @@ export default class Entity extends React.Component<Props> {
         return 'missing';
     }
 
-    selectEntity = () => {
+    selectEntity = (e: SyntheticMouseEvent<HTMLLIElement>) => {
+        // Flow requires that we use `e.currentTarget` instead of `e.target`.
+        // However in this case, we do want to use that, so I'm ignoring all
+        // errors Flow throws there.
+        // $FLOW_IGNORE
```
adngdb: Is this comment actually accurate?

frontend/src/modules/entitieslist/components/Entity.js
```diff
@@ -123,3 +161,13 @@ export default class Entity extends React.Component<Props> {
...
     }
 }
+
+
+const mapStateToProps = (state: Object): Props => {
+    return {
+        isReadOnlyEditor: entities.selectors.isReadOnlyEditor(state),
+        isTranslator: user.selectors.isTranslator(state),
+    };
+};
```
adngdb: Why would you connect and not just pass this down from `EntitiesList` like you did for the `batchactions` state?

frontend/src/modules/entitieslist/components/Entity.js
mathjazz: I think so. It's a copy and paste, but Flow fails.

adngdb: Sorry, my question wasn't clear. I'm wondering if you can actually use `e.currentTarget` here instead of `e.target`?

mathjazz: I tried that initially, but it always returns the `<li>` element.

adngdb: OK great! Just wanted to make sure it wasn't a wrong copy/paste. :-)

frontend/src/modules/entitieslist/components/Entity.js
mathjazz: I have no idea what I was thinking. Updated!

mathjazz:
@adngdb I've added the ability to Shift-select strings and Select all strings, so now I can start implementing actions. I wonder if you have any comments on the recent changes.

mathjazz:
This is now feature complete, so I'd like to ask for a full review.

I still need to add some unit tests, but I wanted to check out if we should make any refactorings first. Especially the button components and related actions call for something, maybe HOC.

frontend/src/modules/batchactions/components/ReplaceAll.js
```diff
@@ -0,0 +1,103 @@
...
+        }
+        else {
+            return this.renderDefault();
+        }
+    }
+
+    render() {
+        return <button
+            className="replace-all"
+            onClick={ this.replaceAll }
```
adngdb: Same here, pretty sure you can just pass `this.props.replaceAll`.

frontend/src/modules/batchactions/components/BatchActions.js
```diff
@@ -80,6 +80,10 @@ export class BatchActionsBase extends React.Component<InternalProps> {
     }
 
     approveAll = () => {
+        if (this.props.batchactions.fetching) {
+            return;
+        }
```
adngdb: Maybe we should also disable the various buttons while a fetching is in progress? That would make things clearer for users. I think a button that just doesn't do anything when clicked, without any feedback, is not a great user experience.

frontend/src/core/api/entity.js
```diff
@@ -47,6 +47,7 @@ export default class EntityAPI extends APIBase {
         locale: string,
         project: string,
         resource: string,
+        entity_ids: ?Array<number>,
```
adngdb: camelCase please. :-)

frontend/src/modules/batchactions/actions.js
```diff
@@ -2,6 +2,10 @@
 
 import api from 'core/api';
 
+import { actions as entitiesActions } from 'core/entities';
+import { actions as resourceActions } from 'core/resource';
+import { actions as statsActions } from 'core/stats';
```
adngdb: I don't very much like this way of importing things, I prefer the `import * as entities from 'core/entities';` form. I think it makes things clearer in the code when you read `entities.actions`. What do you think?

frontend/src/modules/batchactions/actions.js
```diff
@@ -53,6 +59,50 @@ export function performAction(
...
+                        dispatch(
+                            entitiesActions.updateEntityTranslation(
+                                entity.pk,
+                                pluralForm,
+                                translation,
+                            )
+                        );
+                    });
+                }
+            }
```
adngdb: Maybe put all this code in a different function, to make it easier to read?

frontend/src/modules/entitieslist/components/EntitiesList.js
```diff
@@ -234,6 +234,9 @@ export class EntitiesListBase extends React.Component<InternalProps> {
             return;
         }
 
+        // Do not return a specific list of entities defined by their IDs.
+        const entity_ids = null;
```
adngdb: :camel:Caaaaaaaaaaaase! :smile:

frontend/public/static/locale/en-US/translate.ftl
```diff
@@ -79,6 +142,16 @@ editor-KeyboardShortcuts--copy-from-helpers = Copy From Helpers
...
+## Machinery
+## Shows a list of translations from machines.
+
+machinery-machinery-search-placeholder =
+    .placeholder = Type to search machinery
+
+
+## Machinery Translation
+## Shows a specific translation from machinery
+
```
adngdb: Where did this come from? A wrong rebase I guess?

frontend/src/modules/batchactions/components/BatchActions.js
mathjazz: You mean by setting the `disabled` attribute?

frontend/src/modules/batchactions/components/BatchActions.js
```diff
@@ -68,24 +68,27 @@ export class BatchActionsBase extends React.Component<InternalProps> {
...
+                { this.props.batchactions.fetching === 'select-all' ?
+                    <div className="selecting fa fa-sync fa-spin"></div>
+                    :
+                    <Localized
+                        id="batchactions-BatchActions--header-selected-count"
+                        attrs={{ title: true }}
+                        glyph={
+                            <i className="fa fa-times fa-lg"></i>
+                        }
+                        stress={ <span className="stress" /> }
```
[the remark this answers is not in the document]
mathjazz: Hm, I remember something, but I'm not exactly sure what. :) Maybe @stasm can help?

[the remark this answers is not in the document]
Pike: We have issues with autoclosing in fluent strings (and the html template parser), but this is just jsx here, you should be able to just do what works.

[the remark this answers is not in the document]
stasm: @Pike is correct.

frontend/src/modules/batchactions/components/BatchActions.js
adngdb: Yup. And maybe changing their color or something?

mathjazz:
Still need to fix that bug, flow and add more unit tests, but I think another look would be helpful.

frontend/src/modules/entitieslist/components/Entity.test.js
```diff
@@ -148,4 +148,54 @@ describe('<Entity>', () => {
...
+
+    it('does not call the selectEntity function on click on .status', () => {
+        const selectEntityFn = sinon.spy();
+        const wrapper = mount(<Entity
+            entity={ ENTITY_A }
+            selectEntity={ selectEntityFn }
+            locale={ DEFAULT_LOCALE }
+        />);
+        wrapper.find('.status').simulate('click');
+        expect(selectEntityFn.calledOnce).toEqual(false);
```
adngdb: If you want to check it's never been called, you need to test for `.called`. Otherwise, this will pass if the function is called twice or more.

Note that the same thing happens twice below.

frontend/src/modules/batchactions/actions.js
```diff
@@ -33,6 +33,58 @@ export function checkSelection(
 }
 
 
+function _updateUI(
```
adngdb: There's no need to use the leading underscore here, because modules have a notion of private / public. If it's not `export`ed then it is private, so that's quite easy to figure out.

frontend/src/modules/batchactions/reducer.js
```diff
@@ -48,7 +48,23 @@ const initial: BatchActionsState = {
 };
 
 
-function toggleEntity(entities: Array<number>, entity: number) {
+function checkEntities(
+    state_entities: Array<number>,
```
adngdb: :camel:

frontend/src/modules/batchactions/reducer.js
```diff
@@ -60,6 +76,14 @@ function toggleEntity(entities: Array<number>, entity: number) {
 }
 
 
+function uncheckEntities(
+    state_entities: Array<number>,
```
adngdb: :camel:

frontend/src/modules/batchactions/actions.js
adngdb: Yeah I think it's better to enforce that we don't use module names as variable names in our code, to avoid confusion about what is what. Even though sometimes it makes naming a bit complex, I believe it is beneficial in the long term. It is not too big of a deal though.

frontend/src/modules/batchactions/components/BatchActions.js
adngdb: Sure thing!

PR #1340, opened 2019-07-23
Fix bug 1567113 - Upgrade build dependencies for tags.

adngdb:
This is going to be **great** for my contributor stats!


adngdb:
@mathjazz I left these because they were harmless, but after further inspection, they are actually completely useless to us. So I removed the sources of these errors.

mathjazz:
Now the tests are failing and I'm unable to log into Travis to empty the cache. :/

adngdb:
Yeah, it turns out we rely on one of the 2 deps that I removed, and that's the one that has no patch for the vulnerability. It's a build-time only library so I think it's fine to keep that error around.

PR #1341, opened 2019-07-25
No bug - Reorganize FTL files

adngdb:
This is a maintenance patch to:
1. sort the content of FTL files alphabetically;
2. harmonize our usage of l10n IDs to follow our convention everywhere;
3. add the missing French translations.

Each step is a separate commit, hopefully that's easy to review. :-)

### DIFF
--- a/frontend/public/static/locale/en-US/translate.ftl
+++ b/frontend/public/static/locale/en-US/translate.ftl
@@ -1,5 +1,9 @@
 ### Localization for the Translate page of Pontoon
 
+# Naming convention for l10n IDs: "module-ComponentName--sentence-summary".
+# This allows us to minimize the risk of conflicting IDs throughout the app.
+# Please sort alphabetically by (module name, component name).
+
 
 ## Editor
 ## Allows contributors to modify or propose a translation
@@ -12,7 +16,7 @@ editor-editor-button-save = Save
 editor-editor-button-suggest = Suggest
 
 
-## Failed Checks
+## Editor Failed Checks
 ## Renders the failed checks popup
 
 editor-FailedChecks--close = ×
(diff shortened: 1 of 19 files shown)


frontend/public/static/locale/en-US/translate.ftl
```diff
@@ -1,5 +1,9 @@
 ### Localization for the Translate page of Pontoon
 
+# Naming convention for l10n IDs: "module-ComponentName--sentence-summary".
```
mathjazz: Maybe `module-ComponentName--STRING-summary`?

frontend/public/static/locale/en-US/translate.ftl
```diff
@@ -1,18 +1,11 @@
...
-
-editor-editor-sign-in-to-translate = <a>Sign in</a> to translate.
-editor-editor-read-only-localization = This is a read-only localization.
-editor-editor-button-copy = Copy
-editor-editor-button-clear = Clear
-editor-editor-button-save = Save
-editor-editor-button-suggest = Suggest
+# Naming convention for l10n IDs: "module-ComponentName--sentence-summary".
+# This allows us to minimize the risk of conflicting IDs throughout the app.
+# Please sort alphabetically by (module name, component name).
```
mathjazz: Some items in the `editor` and `entitydetails` modules are not sorted by component (`EditorMenu` and `EditorSettings` should come before `FailedChecks`, `EntityNavigation` should also be moved up).

Pike:
> sort the content of FTL files alphabetically;

I'd prefer to not do that. Ordering of strings does benefit from being in reading order. Sometimes, that's also alphabetical, but not always.

In case of doubt, I'd defer to flod for this.

mathjazz:
> > sort the content of FTL files alphabetically;
> 
> I'd prefer to not do that. Ordering of strings does benefit from being in reading order. Sometimes, that's also alphabetical, but not always.
> 
> In case of doubt, I'd defer to flod for this.

In this patch we only sort by:
1. module name
1. component name

We do not sort strings within the component - they are basically in reading order.

I think that's helpful.

Pike:
> We do not sort strings within the component - they are basically in reading order.

Thanks.

PR #1342, opened 2019-07-30
Fix bug 1510184 - Enable selecting multiple filters

adngdb:


frontend/src/modules/search/components/SearchBox.js
```diff
@@ -49,8 +50,17 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
         super(props);
 
         const search = props.parameters.search;
+
+        const statuses =  this.getInitialStatuses();
```
mathjazz: Nit: Double space.

frontend/src/modules/search/components/SearchBox.js
```diff
@@ -97,51 +143,61 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
...
+            navigation.actions.updateSearch(
+                this.props.router,
+                this.state.search,
             )
         );
-    }, 500)
+    }
+
+    updateStatus() {
+        const statuses = Object.keys(this.state.statuses).filter(f => this.state.statuses[f]);
```
mathjazz: `getSelectedStatuses()`?

frontend/src/modules/search/components/FiltersPanel.js
```diff
@@ -33,23 +36,42 @@ type State = {|
...
+            return null;
+        }
+
+        return (event: SyntheticInputEvent<>) => {
+            event.stopPropagation();
+            this.props.toggleStatus(status);
+        };
+    }
 
+    createSetStatus(status: string) {
```
mathjazz: I believe we need an exception here in case of `All`. Right now clicking on `All` threats it as a filter: when you reopen the filter list, it says "Apply 1 filter" and when you tick "Translated", the count increases and the placeholder reads "All, Translated".

frontend/src/modules/search/components/FiltersPanel.js
```diff
@@ -33,23 +36,42 @@ type State = {|
...
+    applyFilters = () => {
+        this.toggleVisibility();
+        return this.props.update();
+    }
+
+    resetFilters = () => {
+        this.props.resetStatuses();
+    }
+
+    createHandleSelectStatus = (status: string) => {
```
mathjazz: I don't understand the name of this function, and the next one:
* why did you prepend names with `create`?
* why does this name include the word `handle, but the next one doesn't?

frontend/src/modules/search/index.js
```diff
@@ -12,11 +12,34 @@ export const NAME: string = 'search';
...
-    { title: 'All', tag: 'all', stat: 'total' },
-    { title: 'Translated', tag: 'translated', stat: 'approved' },
-    { title: 'Fuzzy', tag: 'fuzzy' },
-    { title: 'Warnings', tag: 'warnings' },
-    { title: 'Errors', tag: 'errors' },
-    { title: 'Missing', tag: 'missing' },
-    { title: 'Unreviewed', tag: 'unreviewed' },
+    {
+        title: 'All',
+        tag: 'all',
```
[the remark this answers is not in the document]
adngdb: Yeah I guess `slug` works better.

frontend/src/modules/search/components/SearchBox.js
```diff
@@ -78,6 +88,42 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
         document.removeEventListener('keydown', this.handleShortcuts);
     }
 
+    getInitialStatuses() {
```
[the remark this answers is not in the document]
adngdb: If flow doesn't complain, that's because it's able to deduce that from the code. In which case, no point in making it explicit, I think. At least that's what the folks from AMO recommended: don't write flow types that are not needed.

frontend/src/modules/search/components/FiltersPanel.js
adngdb: They are called `createX` to make it obvious that they return a function, and don't perform any other actions themselves. The `handle` part I will remove.

PR #1344, opened 2019-08-05
Fix bug 1510187: Add tag filters

mathjazz:
This patch also updates `graphene-django`, so that we can work with the choices more easily.


mathjazz:
Just rebased against latest master (so that I can more reliably base the patch for bug 1510189 on this branch).

frontend/src/modules/search/components/SearchBox.js
```diff
@@ -136,31 +150,45 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
...
-                extras[filter] = true;
+            switch (type) {
+                case 'statuses':
+                    statuses[filter] = true;
+                    break;
+                case 'extras':
+                    extras[filter] = true;
+                    break;
+                case 'tags':
+                    tags[filter] = true;
```
adngdb: You want another `break` here and a `default:` otherwise flow will complain. And I believe rightfully so, it's better to always have a `default` clause in your switch, even if it does nothing.

mathjazz: Good call. Thew build even failed. Updated.

PR #1347, opened 2019-08-06
Fix bug 1570746 - Uncompress Mozilla logo.

adngdb:
Our gzip middleware compresses files before sending them, and thus our mozilla logo, which was already compressed, would get compressed twice but uncompressed only once. That resulted in that file not being readable by browsers.


mathjazz:
I wonder if we can remove this folder:
https://github.com/mozilla/pontoon/tree/master/pontoon/base/static/img/Homepage

adngdb:
I can't find any references to it in our code, I guess it's not being used!

PR #1348, opened 2019-08-09
Fix bug 1510189: Authors filters

mathjazz:


frontend/src/modules/search/components/SearchBox.js
```diff
@@ -268,25 +309,38 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
...
         );
 
         if (!selectedFilters.length) {
             selectedFilters = [{ title: 'All' }];
         }
 
         const selectedFiltersString = selectedFilters.map(
-            item => item.title || item.name
+            item => {
+                if (item.display_name) {
```
adngdb: This would benefit from a comment saying that it's specific for Authors.

frontend/src/modules/search/components/SearchBox.js
```diff
@@ -191,9 +219,18 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
...
         });
     }
 
+    getAuthorsAndTimeRangeData = () => {
+        const { locale, project, resource } = this.props.parameters;
+
+        this.props.dispatch(
+            search.actions.getAuthorsAndTimeRangeData(locale, project, resource)
+        )
+    }
```
adngdb: I think you don't need this function, and you can call this action in the `componentDidMount` method of this class.

frontend/src/modules/search/components/SearchBox.test.js
```diff
@@ -1,3 +1,4 @@
+
```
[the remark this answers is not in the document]
mathjazz: Because I wanted to trigger a test and forgot to remove it. :-)

frontend/src/modules/search/components/FiltersPanel.js
```diff
@@ -85,15 +100,27 @@ export class FiltersPanelBase extends React.Component<Props, State> {
...
+        const {
+            statuses,
+            extras,
+            tags,
+            authors,
+            authorsData,
+            tagsData,
+            stats,
+            parameters,
+        } = this.props;
```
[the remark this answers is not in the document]
mathjazz: Good point!

frontend/src/modules/search/components/FiltersPanel.js
```diff
@@ -47,6 +52,16 @@ export class FiltersPanelBase extends React.Component<Props, State> {
         };
     }
 
+    componentDidUpdate(prevProps: Props, prevState: State) {
+        if (
+            this.state.visible &&
+            !prevState.visible &&
+            this.props.parameters.project !== 'all-projects'
```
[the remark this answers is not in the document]
mathjazz: Submitting translations in the translate view affects the Authors and the Time Range part of the filters menu. So in the past we used to update the data every time the translation was submitted, but that was slow and more annoying.

frontend/src/modules/search/components/SearchBox.js
mathjazz: See https://github.com/mozilla/pontoon/pull/1348#discussion_r312485133.

frontend/src/modules/search/components/SearchBox.js
```diff
@@ -322,16 +322,18 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
...
         if (!selectedFilters.length) {
-            selectedFilters = [{ title: 'All' }];
+            selectedFilters = [{ name: 'All' }];
         }
 
         const selectedFiltersString = selectedFilters.map(
             item => {
+                // Specific case for Translation Authors filters
                 if (item.display_name) {
+                    // $FLOW_IGNORE: I just can't figure out the Flow error here.
```
adngdb: What is it?

frontend/src/modules/search/components/FiltersPanel.test.js
```diff
@@ -16,22 +16,17 @@ describe('<FiltersPanelBase>', () => {
...
-                authorsData={ authorsData }
-                tagsData={ tagsData }
-                stats={ stats }
-                parameters={ parameters }
+                tags={ {} }
+                authors={ {} }
+                authorsData={ []}
+                tagsData={ [] }
+                stats={ {} }
+                parameters={ {} }
```
adngdb: Thank you! :pray:

frontend/src/modules/search/components/SearchBox.js
mathjazz:
```
Cannot call this.props.authorsAndTimeRange.authors.filter because:
 • Either cannot get item.name because property name is missing in Author [1].
 • Or cannot add item.display_name and "'s translations" because property display_name of unknown type [2] could
   either behave like a string or like a number.

     src/modules/search/components/SearchBox.js
     310│         );
     311│
     312│         const authors = this.getSelectedAuthors();
     313│         const selectedAuthors = this.props.authorsAndTimeRange.authors.filter(
     314│             f => authors.includes(f.email)
     315│         );
     316│
     317│         let selectedFilters = [].concat(
     318│             selectedStatuses,
        :
 [2] 331│                 if (item.display_name) {

     src/modules/search/reducer.js
 [1]  13│     +authors: Array<Author>,

```

adngdb: OK, that makes sense to me. Basically here `item` can be either of type `Author` or type X (not sure where that one is defined, but it doesn't matter). Author has `display_name` and not `name`, X has `name` and not `display_name`. The problem here is that Flow is not able to tell, in each branch of the condition, which type of item is being used. Which makes, in its eyes, the instruction below invalid, because there are branches where `item.display_name` is not defined and thus could be anything.

I can't dive into this right now, but it might take roots in the definition of type "X". I think this definitely a solvable issue and we should not be ignoring it.

mathjazz: Thanks! I'll try to change the logic a little bit.

adngdb:
One thing that is missing (because I didn't implement it before) is to make the option toolbar fixed when the filters panel is longer than the page. Right now you have to scroll down a lot if you want to click "Apply filters".

adngdb:
Also, contributors are sorted in order of increasing translations count, instead of decreasing.

mathjazz:
> One thing that is missing (because I didn't implement it before) is to make the option toolbar fixed when the filters panel is longer than the page. Right now you have to scroll down a lot if you want to click "Apply filters".

Nice catch! I'll address this as part of https://bugzilla.mozilla.org/show_bug.cgi?id=1510190, which will also affect toolbar positioning (due to collapsible time range chart).

PR #1350, opened 2019-08-13
Fix bug 1570228: Update History after Save/Review when entity not changed

mathjazz:


frontend/src/core/plural/actions.js
```diff
@@ -26,6 +27,9 @@ export function moveToNextTranslation(
     else if (nextEntity !== entity) {
         dispatch(navActions.updateEntity(router, nextEntity.toString()));
     }
+    else {
+        dispatch(historyActions.get(entity, locale.code, pluralForm));
```
adngdb: I think this should actually be in `EntityDetails`. https://github.com/mozilla/pontoon/blob/master/frontend/src/modules/entitydetails/components/EntityDetails.js#L105-L106 is where the history is updated on change, and I think we should keep that as the only place (otherwise it adds a lot of complexity and we might later on wonder why we have several updates happening). The question thus becomes, why does that test not work now?

adngdb: After reading the bug, I now understand why it doesn't work and why you're doing it this way. It still feels wrong to call `history.get` from different locations though. There's a bigger question here probably of how we organize our action calls (at the moment, we mostly react from state changes, and not much to action triggers, meaning that actions change state which triggers actions instead of having actions trigger other actions themselves). Not sure what the correct answer is. :/

mathjazz: What do you suggest we do? I think I might still be able to fix it using the trick you suggested in https://github.com/mozilla/pontoon/pull/1350/files#r315254605.

adngdb: If you can manage to do that, then it's perfect in my opinion! Otherwise, I don't really know.

mathjazz: Updated.

PR #1351, opened 2019-08-19
Fix bugs 1574889 (<title>) and 1574922 (update navbar on project change)

mathjazz:
This should have been two separate PRs, but since they are small and the 2nd depends on the 1st, I'm squeezing them together. Will avoid doing that in the future.


adngdb:
Hah, except you broke a unit test. :-)

mathjazz:
Nice catch! That error is actually unrelated to this patch, but I need to fix it anyways.

PR #1356, opened 2019-08-23
Fix bug 1576120: Store Fluent strings in TM as simple strings

mathjazz:
I've tested all 4 TM creation scenarios:
1. Submit a translation
2. Approve a suggestion
3. Mass approve suggestions
4. Import translations during sync

TM entries seem to be storing fine. We also don't regress performance (not even increase DB calls).

The migration took <10 minutes on stage and didn't cause any issues.


pontoon/base/utils.py
```diff
@@ -643,7 +643,7 @@ def build_translation_memory_file(creation_date, locale_code, entries):
         }
     )
     for resource_path, key, source, target, project_name, project_slug in entries:
-        tuid = ':'.join((project_slug, slugify(resource_path), slugify(key)))
+        tuid = ':'.join((project_slug, resource_path, slugify(key)))
```
mathjazz: Outside the scope of this patch, but small enough.

This change was introduced as part of bug 1390149. We don't need to slugify resource path, so let's keep it usable.

frontend/src/modules/machinery/components/Machinery.css
```diff
@@ -8,17 +8,6 @@
...
-.machinery p {
-    min-height: 22px;
-}
-
-.machinery > p {
-    border-bottom: 1px solid #5E6475;
-    color: #aaa;
-    font-style: italic;
-    padding: 5px 10px 20px 10px;
-}
```
mathjazz: This rule was never used.

frontend/src/modules/machinery/components/Translation.css
```diff
@@ -44,6 +44,12 @@
     color: #7BC876;
 }
 
+.machinery .translation p {
+    min-height: 22px;
+    text-align: start;
+    white-space: pre-wrap;
+}
```
mathjazz: Moving to the right file and properties to keep formatting of multiline strings in line with Translate.Current.

pontoon/base/migrations/0142_migrate_tm_ftl.py
```diff
@@ -0,0 +1,66 @@
...
+        batch_size=1000,
+    )
+
+
+def revert_simplify_ftl_tm_entries(apps, schema):
+    TranslationMemoryEntry = apps.get_model('base', 'TranslationMemoryEntry')
+
+    tm_entries = (
+        TranslationMemoryEntry.objects
+        .filter(entity__resource__format='ftl')
```
jotes: That's a small thing but you can use `.update` and `F()` fields to make the downgrade migration faster.

mathjazz: Good point! I've been only looking into using `.update` in the "upgrade" direction.

mathjazz: I'll keep that as it is, if you don't mind, to prevent any last-minute breakage.

PR #1357, opened 2019-08-23
Fix bug 1557922 - Keep focus on search input after a request.

adngdb:
@mathjazz I randomly tried this to see what it would do, and, well, it kinda seems to work? Lucky me I guess! :D

### DIFF
--- a/frontend/src/core/editor/components/connectedEditor.js
+++ b/frontend/src/core/editor/components/connectedEditor.js
@@ -129,7 +129,7 @@ export default function connectedEditor<Object>(
         }
 
         clearEditor = () => {
-            this.updateTranslation('');
+            this.updateTranslation('', true);
         }
 
         sendTranslation = (ignoreWarnings?: boolean, translation?: string) => {
--- a/frontend/src/core/editor/components/connectedEditor.js
+++ b/frontend/src/core/editor/components/connectedEditor.js
@@ -22,7 +22,7 @@ import type { Locale } from 'core/locale';
 import type { NavigationParams } from 'core/navigation';
 import type { UserState } from 'core/user';
 import type { ChangeOperation } from 'modules/history';
-import type { AuthorsAndTimeRangeState } from 'modules/search';
+import type { SearchAndFilters } from 'modules/search';
 import type { UnsavedChangesState } from 'modules/unsavedchanges';
 import type { EditorState } from '../reducer';
 import type { Translation } from '../actions';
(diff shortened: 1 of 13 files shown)


adngdb:
@mathjazz I've been thinking about this a bunch, and I don't see a simple solution to this problem. 

The problem comes from React. In jQuery / DOM land, user actions are directly related to view updates, and it is thus very easy to, during the execution of an action, wait for the UI to update and then make some DOM changes (like putting focus on a newly constructed element).

React de-correlates actions from view updates, and that means it becomes very hard to perform DOM changes, directly related to a user action, after all the dependencies of that action have been done. We have to rely on the state to set the focus, and that is very hacky and also very complex.

One solution could be to just stop putting automatic focus on the editor. That way the focus will stay where it was whenever there's a change in the app, effectively solving the issue at hand (while, I guess, creating another inconvenience for users).

if there's another solution, I haven't found it yet, and unless I have a stroke of genius overnight, I don't think I'll be able to fix this before my vacations. I'm thus putting this back into your hands.

mathjazz:
> The problem comes from React. In jQuery / DOM land, user actions are directly related to view updates, and it is thus very easy to, during the execution of an action, wait for the UI to update and then make some DOM changes (like putting focus on a newly constructed element).

You should have seen how much focus-handling code there is `translate.js` and how ugly it is. :)

I think the solution could be to store the previous value of the search parameter in redux store and compare it with the current search parameter before setting focus on the editor. If the values are different, it means the user is searching and thus we shouldn't focus the editor.

mathjazz:
> I think the solution could be to store the previous value of the search parameter in redux store and compare it with the current search parameter before setting focus on the editor. If the values are different, it means the user is searching and thus we shouldn't focus the editor.

I ended up building a variant of that: I put `searchInputFocused` in the store and update it on search input focus/blur. Before setting focus in the editor, I check that the value of it isn't true.

This patch brings several other improvements:
* It also sets focus in the Rich Editor (seems like we forgot about it). 55e4752 
* It sets the `key` attributes in the Rich Fluent Original. f45e771
* It removed the `focus` attribute from the `AceEditor`.

The patch is not as big as it looks, because I also renamed a bunch of variables in 54bd7c0.

mathjazz:
Hi @flodolo,

You mentioned editor stealing focus from the search field was causing you pain. I wonder if you would be willing to take this patch for a spin before it's merged.

It's deployed to stage, you only need to make sure you are in Translate.Next.

flodolo:
> It's deployed to stage, you only need to make sure you are in Translate.Next.

I can do that tomorrow, I'll update here as soon as I'm done.

flodolo:
I haven't looked at stage in a while, but is all that noise in the Error Console expected or caused by this patch?

Schermata 2019-09-05 alle 18 52 45

mathjazz:
That's expected and unrelated to this patch.

flodolo:
Tested yesterday and a bit more today, can't spot any issue on stage 👍

mathjazz:
Thanks!

PR #1360, opened 2019-08-25
Bug 1362323 - Fix tests for the administration app.

jotes:
Hey,

The PR fixes unittests for the administration module.
Major changes :
* I've bumped up the version of `django-cookies-samesite` because of the missing Cookie module
* The `HTTPResponse.content`  returns a bytes object (on Python 3). The `contains` method raises an exception if you check a non-bytes object.

@adngdb @mathjazz r?

### DIFF
--- a/pontoon/administration/tests/test_views.py
+++ b/pontoon/administration/tests/test_views.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-from __future__ import absolute_import
+from __future__ import absolute_import, unicode_literals
 
 import pytest
 
@@ -108,7 +108,7 @@ def test_manage_project_strings_new(client_superuser, locale_a):
         assert entity.order == index
 
     # Verify new strings appear on the page.
-    assert 'Hey, I just met you' in response.content
+    assert b'Hey, I just met you' in response.content
 
 
 @pytest.mark.django_db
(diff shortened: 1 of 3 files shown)


pontoon/administration/tests/test_views.py
```diff
@@ -108,7 +108,7 @@ def test_manage_project_strings_new(client_superuser, locale_a):
         assert entity.order == index
 
     # Verify new strings appear on the page.
-    assert 'Hey, I just met you' in response.content
+    assert b'Hey, I just met you' in response.content
```
adngdb: I can tell I wrote these tests just from this. :muscle:

pontoon/base/models.py
```diff
@@ -99,7 +99,7 @@ def user_profile_url(self):
 
 
 def user_gravatar_url(self, size):
-    email = hashlib.md5(self.email.lower()).hexdigest()
+    email = hashlib.md5(self.email.lower().encode('utf-8')).hexdigest()
```
adngdb: Will we be able to remove those `.encode('utf-8')` once we're on Py3?

jotes: Py3 reduces it to `.encode()`. Hashlib requires bytes here.

PR #1363, opened 2019-08-28
Fix bug 1510190: Time range filter

mathjazz:
Time range filter!

Functionality is the same as in Translate.Current, with one difference - the datepicker widget is no more. The only way you can enter the time range is by sliding the chart or via manual text input. Or by using the shortcuts (e.g. "30 Days"). Or by typing date into the URL.

Well, that's 4 methods, should be sufficient. The datepicker widget is buggy and confusing (only allows for selecting dates, whereas the input field also shows time). I believe we need to refactor the UX in the near future, so I'd move on without the effort of bring in the datepicker.

I think I have also fixed the bug with the 2nd applying of filter having no effect:
https://bugzilla.mozilla.org/show_bug.cgi?id=1469611

Entering text via input is slightly better than on prod, but this bug remains open:
https://bugzilla.mozilla.org/show_bug.cgi?id=1377872

I've fixed a few additional bugs I have found while comparing with how Translate.Current works:
- chart now always resets when the filter menu opens
- chart keeps state when it's toggled using the Edit/Save Range buttons
- bad time range URL no longer results in server error, which no longer breaks Translate.Next

[The toolbar (Apply N filters) position](https://github.com/mozilla/pontoon/pull/1348#issuecomment-519986026) has also been taken care of.

And finally, two things I'd like to hear your opinion on:
- `FiltersPanel.js` is getting pretty big. I'm tempted to create custom components for each filter type. But then again I wonder if I should do that now or wait for the UX refactor.
- I'm thinking about renaming *timeRange* to *time*.


frontend/src/modules/search/components/FiltersPanel.js
```diff
@@ -44,22 +61,163 @@ type State = {|
...
+        return date.format(d, INPUT_FORMAT);
+    }
+
+    handleInputChange = (event: SyntheticInputEvent<HTMLInputElement>) => {
+        const d = date.parse(event.target.value, INPUT_FORMAT);
+
+        // If valid date, update chart extremes
+        if (!isNaN(d)) {
+            this.updateChartExtremes(event.target.name, d.getTime());
+        }
```
adngdb: Maybe show an error notification when the date is not valid?

frontend/src/modules/search/components/FiltersPanel.js
```diff
@@ -73,6 +231,16 @@ export class FiltersPanelBase extends React.Component<Props, State> {
...
 
+    createToggleTimeRangeFilter() {
+        const { chartFrom, chartTo, isChartVisible } = this.state;
+
+        if (isChartVisible) {
+            return;
+        }
+
+        return this.createToggleFilter([chartFrom, chartTo].join('-'), 'timeRange');
+    }
```
adngdb: That is a weird construct that is not required. The reason to make a method that returns a function is to create a closure that preserves a variable. In this case, you don't need that, because your variable is in `state`. So you could here instead create a `toggleTimeRangeFilter` method that combines the content of this method and of `createToggleFilter`'s returned function.

frontend/src/modules/search/components/FiltersPanel.js
```diff
@@ -84,6 +252,16 @@ export class FiltersPanelBase extends React.Component<Props, State> {
...
 
+    createApplyTimeRangeFilter() {
+        const { chartFrom, chartTo, isChartVisible } = this.state;
+
+        if (isChartVisible) {
+            return;
+        }
+
+        return this.createApplySingleFilter([chartFrom, chartTo].join('-'), 'timeRange');
+    }
```
adngdb: The same goes here, just make a simple method that will be called directly, no need to create a closure.

frontend/src/modules/search/components/FiltersPanel.js
```diff
@@ -240,6 +443,83 @@ export class FiltersPanelBase extends React.Component<Props, State> {
...
+                                    highcharts={ Highcharts }
+                                    options={ this.state.chartOptions }
+                                    constructorType = { 'stockChart' }
+                                    allowChartUpdate = { false }
+                                    containerProps = {{ className: 'chart' }}
+                                    ref={ this.chart }
+                                />
+                            }
+                        </li>
+                    </>}
```
adngdb: All of this would indeed probably deserve to be in a separate component.

frontend/src/modules/search/components/SearchBox.js
```diff
@@ -176,6 +224,7 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
         const statuses = this.getInitialStatuses();
         const extras = this.getInitialExtras();
         const tags = this.getInitialTags();
+        let timeRange = this.getInitialTimeRange();
```
adngdb: I like consistency too, but in this case I really think this should be just `let timeRange = null;` because I believe simplicity > consistency. :-)

adngdb:
Holy Molly the way this filter works is incredibly unintuitive to me. Here's what I did: 

1. Click "Edit range".
2. Change time range.
3. Hit "Save range".

Expected result: the "Apply filter" button shows up. At that point I got very confused, tried the same steps on prod, saw that it worked the same and started writing something down here.

Then I tried clicking on one of the dates. I expected to be able to edit that time, instead it applied the filter and reloaded the entities. Hah! That's when I realized it worked as the other filters and you have to select it or click it to apply it.

Well on one hand I'm happy to see that this is consistent with the rest of the filters, but on the other I really really did not expect this filter to work like that and got really confused to see that it did.

This is not a blocker, as this is exactly how it works on prod. But wow. I think that's not a great UX! :)

frontend/src/modules/search/components/FiltersPanel.js
```diff
@@ -44,22 +61,163 @@ type State = {|
...
+    }
+
+    updateChartExtremes = (key: ?string, value: ?number) => {
+        const { chartFrom, chartTo } = this.state;
+
+        if (!chartFrom || !chartTo) {
+            return;
+        }
+
+        let extremes = {
```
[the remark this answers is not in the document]
mathjazz: They are used in a function called `setExtremes`, though.

frontend/src/modules/search/components/FiltersPanel.js
```diff
@@ -3,37 +3,54 @@
...
     resetFilters: () => void,
     toggleFilter: (string, string) => void,
     update: () => void,
+    updateTimeRange: (filter: string) => void,
 |};
 
 type State = {|
+    chartFrom: ?number,
+    chartTo: ?number,
+    chartOptions: Object,
```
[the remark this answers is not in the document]
mathjazz: Honestly, I don't see much value in writing another 130 lines of code to mimic `chart-options.js`.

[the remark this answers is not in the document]
mathjazz: Maybe it's OK if I only define the smallest required set of object properties and their (sub)types?

frontend/src/modules/search/components/FiltersPanel.js
```diff
@@ -44,22 +61,163 @@ type State = {|
...
+            chartFrom = timeRange.from;
+            chartTo = timeRange.to;
         }
+
+        // Set chart data
+        let chartOptions = this.state.chartOptions;
+        chartOptions.series[0].data = timeRangeData;
+
+        // Set chart update function
+        const self = this;
```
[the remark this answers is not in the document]
mathjazz: Cool, I had no idea!

frontend/src/modules/search/components/SearchBox.js
mathjazz: Good point. I'll remove `getInitialTimeRange()`.

mathjazz:
> > ```
> > * I'm thinking about renaming _timeRange_ to _time_.
> > ```
> 
> 
> Not sure about this one. `timeRange` is more precise than `time`. But I don't have a strong opinion here.

I thought we should make it consistent with the URL parameter. But that's OK.

mathjazz:
> Expected result: the "Apply filter" button shows up. At that point I got very confused, tried the same steps on prod, saw that it worked the same and started writing something down here.

I implemented that in a4ee37f: when Save Range is clicked, Time Range filter automatically gets selected (if it's not selected yet) and the toolbar with the "Apply filter" button shows up. The Save Range button is now also slightly more noticeable (it's green).

I the followup patch I'll improve the manual time input (https://bugzilla.mozilla.org/show_bug.cgi?id=1469611).

PR #1365, opened 2019-08-30
Fix bug 1554092: Render Project Menu in All Projects view

mathjazz:
The first 4 commits are refactoring the way we fetch locales: instead of all locales data, we only fetch data for the current locale. That's the only data we actually use. So the request is not only faster, but also allows to be used to fetch data needed for the "All Projects" menu. As a consequence, the "locales" module is now called "locale".


adngdb:
I like this very very much! I was wondering why you didn't show the project menu when a project is selected though? Is that because users rarely switch from one project to another? 

There's a bug when you switch project (or go to All projects) and there's an unsaved change. Steps to reproduce: 

1. Make a change in the editor
2. Click "All project" or any project link if you're already on All projects.
3. Click "Quit anyway" in the editor's unsaved changes popup.

Expected: you get there.

Actual result: JS error and empty screen.

mathjazz:
> I like this very very much! I was wondering why you didn't show the project menu when a project is selected though? Is that because users rarely switch from one project to another?

I'd like to test the current behaviour with the users first, which allows to switch to the localization (locale-project) dashboard with a single click. And it's the same amount of clicks to switch the project if you go to the team dashboard first and select a project there.

> There's a bug when you switch project (or go to All projects) and there's an unsaved change. Steps to reproduce:
> 
>     1. Make a change in the editor
>     2. Click "All project" or any project link if you're already on All projects.
>     3. Click "Quit anyway" in the editor's unsaved changes popup.
> 
> Expected: you get there.
> 
> Actual result: JS error and empty screen.

Nice catch. That's unrelated to this patch, but fixing anyways, it's simple.

mathjazz:
> I'd like to test the current behaviour with the users first, which allows to switch to the localization (locale-project) dashboard with a single click. And it's the same amount of clicks to switch the project if you go to the team dashboard first and select a project there.

Filed a bug with a basic patch:
https://bugzilla.mozilla.org/show_bug.cgi?id=1578235

PR #1366, opened 2019-09-01
Fix Bug 1578057: update locale's latest_translation for visible projects only

MikkCZ:
As of now, translation changes to invisible projects (e.g. system project like tutorial) do update the latest activity column shown on all teams dashboard. That's kind of confusing, since you cannot click through to that system project from any dashboards. As the name suggests, the project is invisible there.

This change omits updating `latest_translation` for a locale if the translation is being saved for a system project. Note, that `latest_translation` is still updated in all other tables, including `project_locale`, because the dashboard of an invisible project is still accessible via direct link or from its translation view and the latest activity should be displayed there.

I have written a small unit test for the behaviour. Also I have tested the patch manually in local setup. Exact steps to reproduce can be found in the bug, though in the local setup you need to add at least one visible project, so the teams dashboard can show something interesting.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -2868,22 +2868,25 @@ class Translation(DirtyFieldsMixin, models.Model):
         resource = self.entity.resource
         project = resource.project
         locale = self.locale
-        translatedresource = TranslatedResource.objects.get(resource=resource, locale=locale)
 
-        instances = [translatedresource, project]
+        to_update = [
+            (TranslatedResource, Q(Q(resource=resource) & Q(locale=locale))),
+            (ProjectLocale, Q(Q(project=project) & Q(locale=locale))),
+            (Project, Q(pk=project.pk)),
+        ]
 
         if not project.system_project:
-            instances.append(locale)
+            to_update.append((Locale, Q(pk=locale.pk)))
 
-        project_locale = utils.get_object_or_none(ProjectLocale, project=project, locale=locale)
-        if project_locale:
-            instances.append(project_locale)
-
-        for instance in instances:
-            latest = instance.latest_translation
-            if latest is None or self.latest_activity['date'] > latest.latest_activity['date']:
-                instance.latest_translation = self
-                instance.save(update_fields=['latest_translation'])
+        for model, query in to_update:
+            model.objects.filter(
+                Q(
+                    query & Q(
+                        Q(latest_translation=None)
+                        | Q(latest_translation__date__lt=self.latest_activity['date'])
+                    )
+                )
+            ).update(latest_translation=self)
 
     def approve(self, user):
         """
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -2882,8 +2882,8 @@ class Translation(DirtyFieldsMixin, models.Model):
             model.objects.filter(
                 Q(
                     query & Q(
-                        Q(latest_translation=None)
-                        | Q(latest_translation__date__lt=self.latest_activity['date'])
+                        Q(latest_translation=None) |
+                        Q(latest_translation__date__lt=self.latest_activity['date'])
                     )
                 )
             ).update(latest_translation=self)
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -2870,7 +2870,10 @@ class Translation(DirtyFieldsMixin, models.Model):
         locale = self.locale
         translatedresource = TranslatedResource.objects.get(resource=resource, locale=locale)
 
-        instances = [project, locale, translatedresource]
+        instances = [translatedresource, project]
+
+        if Project.objects.visible().filter(pk=project.pk).first():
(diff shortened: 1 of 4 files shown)


MikkCZ:
Hi @mathjazz, can you suggest me a suitable reviewer, please?

pontoon/base/tests/models/test_translation.py
```diff
@@ -98,6 +98,44 @@ def test_translation_save_latest_update(locale_a, project_a):
     assert project_locale.latest_translation == newer_approved_translation
 
 
+@pytest.mark.django_db
+def test_translation_save_latest_update_for_system_project(locale_a, system_project_a):
```
MikkCZ: I have noticed in this file, that at least the `test_translation_save_latest_update` attempts to test multiple scenarios at once. It's more like some saga than a single unit test. Would you mind if I split such tests in this file, to cover exactly one scenario?

pontoon/base/models.py
```diff
@@ -2870,7 +2870,10 @@ def update_latest_translation(self):
         locale = self.locale
         translatedresource = TranslatedResource.objects.get(resource=resource, locale=locale)
 
-        instances = [project, locale, translatedresource]
+        instances = [translatedresource, project]
+
+        if Project.objects.visible().filter(pk=project.pk).first():
```
MikkCZ: Is there a better way to do this in Django? I am not that familiar with the DB layer, but it seems to me there are three SELECTs to the database first (`TranslatedResource.objects.`, `Project.objects.` and `utils.get_object_or_none`) to decide whether to update the three tables or not.

Maybe it could be rewritten to something that issues just three UPDATE statements instead, with the proper WHERE clause to ensure the right rows are (not) updated.

mathjazz:
> Hi @mathjazz, can you suggest me a suitable reviewer, please?

Thanks for filing a bug and creating a PR! I'll be happy to review!

pontoon/base/models.py
mathjazz: It's enough to check if the project isn't a system project here like this:
```
if not project.system_project:
```

That's also what we do in `adjust_all_stats()` to make sure we don't update stats on `/teams` when system projects change.

The `Project.visible()` method would also check if the project is available (i.e. has been synced for the first time), but we don't need that since in this case the `Translation.save()` method (which calls `update_latest_translation()`) would not be callable.

pontoon/base/tests/models/test_translation.py
mathjazz: Nice catch! Feel free to do so.

MikkCZ: Done.

pontoon/base/models.py
MikkCZ: Nice, that's one SELECT down.

Do you think I should also get rid of the `get` on `TranslatedResource` and `ProjectLocale`? There wouldn't be the for-loop, just `X.objects.filter(...).update(...)`.

mathjazz: You should be able to save another DB request by getting rid of `TranslatedResource.objects.get()` if you pass `translatedresource` to `self.update_latest_translation()` in `save()`.

I'm not sure about the `ProjectLocale`.

You can use this patch to debug SQL performance and compare the number of DB calls:
https://gist.github.com/mathjazz/92d2f6a36c968e8c1f6496f5e5aa0318

MikkCZ: I have tried to do it via QuerySet update() API. `update_latest_translation` is sometimes called from a context, where the translated resource is not present, or at least not directly.

MikkCZ:
Thanks for the review. Please also check the last two commits. If you like them and CI is green, I intend no more changes.

pontoon/base/models.py
```diff
@@ -2868,22 +2868,25 @@ def update_latest_translation(self):
...
-            latest = instance.latest_translation
-            if latest is None or self.latest_activity['date'] > latest.latest_activity['date']:
-                instance.latest_translation = self
-                instance.save(update_fields=['latest_translation'])
+        for model, query in to_update:
+            model.objects.filter(
+                Q(
+                    query & Q(
+                        Q(latest_translation=None)
+                        | Q(latest_translation__date__lt=self.latest_activity['date'])
```
mathjazz: Nit: Please move that `|` operator to the end of the previous line, so that `Q()`'s will align.

MikkCZ: Done.

pontoon/base/models.py
```diff
@@ -1477,9 +1477,9 @@ def get_chart(cls, self, extra=None):
         Get chart for project, locale or combination of both.
 
         :param self: object to get data for,
-            instance of Projet or Locale
+            instance of Project or Locale
         :param extra: extra filter to be used,
-            instance of Projet or Locale
+            instance of Project or Locale
```
mathjazz: Thank you!

PR #1375, opened 2019-09-10
Fix bug 1580211: Add Support for strings with SelectExpressions

mathjazz:
This patch brings support for strings with `SelectExpressions` in the `RichEditor` and `RichString` components.

The first 4 commits are setting the stage by adapting the reusable functions to support `SelectExpressions`. The last 3 commits bring stylistic changes. Which leaves 197231b (editor) and 54eccdf (source string panel) to be the gist of this patch.

I'll probably block landing of this patch on [bug 1580213](https://bugzilla.mozilla.org/show_bug.cgi?id=1580213), because having just a general SelectExpression UI handle plurals might lead to some damage.

### DIFF
--- a/frontend/src/core/utils/fluent/flattenElements.js
+++ b/frontend/src/core/utils/fluent/flattenElements.js
@@ -8,25 +8,45 @@ import type { FluentElement } from './types';
 /**
  * Return a flattened list of Fluent elements.
  *
- * Takes a list of Fluent SyntaxNode elements, extracts a serialized value for
- * each and returns an array with a single TextElement containing those values.
+ * @param {Array<SyntaxNode>} elements A list of Fluent SyntaxNode elements to flatten.
  *
- * @param {Array<SyntaxNode>} elements A list of Fluent SyntaxNode lements to flatten.
- *
- * @returns {Array<TextElement>} An array containing a single TextElement which
- * contains all elements' values serialized.
+ * @returns {Array<TextElement|Placeable>} An array containing elements of type
+ * TextElement (merging serialized values of neighbour simple elements) and
+ * Placeable (representing select expressions).
  */
 export default function flattenElements(elements: Array<FluentElement>): Array<FluentElement> {
-    const values = elements.map(element => {
-        switch (element.type) {
-            case 'TextElement':
-                return element.value;
-            default:
-                return serializeExpression(element);
+    const flatElements = [];
+    let simpleElements = [];
+
+    elements.forEach((element, index) => {
+        if (element.type === 'Placeable' && element.expression.type === 'SelectExpression') {
+            // Before adding SelectExpression merge collected simple elements into a TextElement
+            if (simpleElements.length) {
+                flatElements.push(new TextElement(simpleElements.join('')));
+                simpleElements = [];
+            }
+
+            // Flatten SelectExpression variant elements
+            element.expression.variants.forEach(variant => {
+                variant.value.elements = flattenElements(variant.value.elements);
+            });
+
+            flatElements.push(element);
+        }
+        else {
+            if (element.type === 'TextElement') {
+                simpleElements.push(element.value);
+            }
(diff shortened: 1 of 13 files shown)


frontend/src/core/utils/fluent/flattenElements.js
```diff
@@ -8,25 +8,45 @@ import type { FluentElement } from './types';
...
+                simpleElements.push(element.value);
+            }
+            else {
+                simpleElements.push(serializeExpression(element));
+            }
+
+            // Before the end of loop merge collected simple elements into a TextElement
+            if (index === elements.length - 1) {
+                flatElements.push(new TextElement(simpleElements.join('')));
+            }
```
Pike: This part should be after the loop, I think.

frontend/src/core/utils/fluent/flattenElements.js
```diff
@@ -8,25 +8,45 @@ import type { FluentElement } from './types';
...
  */
 export default function flattenElements(elements: Array<FluentElement>): Array<FluentElement> {
-    const values = elements.map(element => {
-        switch (element.type) {
-            case 'TextElement':
-                return element.value;
-            default:
-                return serializeExpression(element);
+    const flatElements = [];
+    let simpleElements = [];
```
Pike: As I want changes here below, how about `textFragments`? They're not simple nor elements in the end, right?

frontend/src/core/utils/fluent/types.js
```diff
@@ -1,8 +1,19 @@
...
+
+export type FluentExpression = {
+    type: string,
+    variants: Array<FluentExpressionVariant>,
+};
+
 export type FluentElement = {
     type: string,
     value: string,
+    expression: FluentExpression,
```
Pike: I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1580468 for the more general nomenclature problem, but I'd like to start here by not making it worse ;-)

We should have a `Placeable`, with a `SelectExpression` and just `Variant`s. And then `FluentElement` should probably be `{type,value} | Placeable` ?

Should `Variant` have `default` already, too? I don't think we use that yet, but it's probably nice to have for the next thing?

mathjazz:
Thanks! Updated and listed [bug 1580468](https://bugzilla.mozilla.org/show_bug.cgi?id=1580468) on the [wiki](https://wiki.mozilla.org/L10n:Pontoon#Be_ready_to_rapidly_develop_and_test_new_features_in_the_Pontoon_translation_workbench).

frontend/src/modules/fluenteditor/components/RichTranslationForm.js
```diff
@@ -327,26 +328,38 @@ export class RichTranslationFormBase extends React.Component<InternalProps> {
...
+                { this.renderInput(value, path) }
+            </td>
+        </tr>;
+    }
+
+    renderElements(elements: Array<FluentElement>, path: MessagePath, label: string): React.Node {
         return elements.map((element, index) => {
-            if (element.type !== 'TextElement') {
-                return null;
+            if (element.type === 'Placeable' && element.expression.type === 'SelectExpression') {
```
Pike: Here and below, there's a somewhat mixed up assumption that `flattenMessage` has been called. On one hand, you're testing the inside of the `Placeable`, and on the other hand, you're assuming it's a `TextElement` if it's not a `SelectExpression`.

That puzzles me a bit, but I don't have a good proposal on how to resolve that.

mathjazz: Here (and in `RichString.js ` as well) we expect `flattenMessage` to already been called. That's why we can hardcode `0` in `variant.value.elements[0].value`. For `SelectExpressions` we only flatten variants.

mathjazz:
I've addressed the comments. I've also added the optional `default` parameter.

This is ready for another look.

PR #1379, opened 2019-09-12
Fix bug 1580213: Add Special UI for plurals

mathjazz:
The Plurals UI is actually not so special. `SelectExpressions` are considered plurals if keys of all their variants are either CLDR plural categories or `NumberLiterals`. The only changes for plurals happen in the editor:
1. Variant keys that match CLDR plural categories are equiped with examples.
1. For untranslated strings we render locale-specific variants, instead of copying the source.

### DIFF
--- a/frontend/public/static/locale/en-US/translate.ftl
+++ b/frontend/public/static/locale/en-US/translate.ftl
@@ -195,6 +195,12 @@ entitydetails-GenericOriginalString--plural = Plural
 entitydetails-GenericOriginalString--singular = Singular
 
 
+## Rich editor
+## Renders a Rich editor for Fluent string editing
+
+fluenteditor-RichTranslationForm--plural-example = (e.g. <stress>{ $example }<stress>)
+
+
 ## History
 ## Shows a list of translations for a specific entity
 
--- a/frontend/public/static/locale/en-US/translate.ftl
+++ b/frontend/public/static/locale/en-US/translate.ftl
@@ -198,7 +198,7 @@ entitydetails-GenericOriginalString--singular = Singular
 ## Rich editor
 ## Renders a Rich editor for Fluent string editing
 
-fluenteditor-RichTranslationForm--plural-example = (e.g. <stress>{ $example }<stress>)
+fluenteditor-RichTranslationForm--plural-example = <plural>{ $plural }</plural> <example>(e.g. <stress>{ $example }</stress>)</example>
 
 
 ## History
--- a/frontend/public/static/locale/en-US/translate.ftl
+++ b/frontend/public/static/locale/en-US/translate.ftl
@@ -198,7 +198,7 @@ entitydetails-GenericOriginalString--singular = Singular
 ## Rich editor
 ## Renders a Rich editor for Fluent string editing
 
-fluenteditor-RichTranslationForm--plural-example = <plural>{ $plural }</plural> <example>(e.g. <stress>{ $example }</stress>)</example>
+fluenteditor-RichTranslationForm--plural-example = { $plural } (e.g. <stress>{ $example }</stress>)
 
 
 ## History
--- a/frontend/src/core/locale/getPluralExamples.js
+++ b/frontend/src/core/locale/getPluralExamples.js
@@ -1,0 +1,30 @@
+/* @flow */
+
+import type { Locale } from './actions';
+
+
+export default function getPluralExamples(locale: Locale) {
+    const pluralsCount = locale.cldrPlurals.length;
+    const examples = {};
+
+    if (pluralsCount === 2) {
+        examples[locale.cldrPlurals[0]] = 1;
+        examples[locale.cldrPlurals[1]] = 2;
+    }
+    else {
+        // This variable is used in the pluralRule we eval in the while block.
+        let n = 0;
+        while (Object.keys(examples).length < pluralsCount) {
+            // This `eval` is not so much evil. The pluralRule we parse
+            // comes from our database and is not a user input.
+            // eslint-disable-next-line
+            const rule = eval(locale.pluralRule);
+            if (!examples[locale.cldrPlurals[rule]]) {
+                examples[locale.cldrPlurals[rule]] = n;
+            }
+            n++;
+        }
+    }
+
+    return examples;
+}
--- a/frontend/src/core/locale/getPluralExamples.js
+++ b/frontend/src/core/locale/getPluralExamples.js
@@ -3,6 +3,20 @@
 import type { Locale } from './actions';
 
 
+/**
+ * Return a map of given locale's cldrPlurals and their plural examples.
+ *
+ * For example, for Slovenian (sl) the function returns the following:
+ *     {
+ *          1: 1,
+ *          2: 2,
+ *          3: 3,
+ *          5: 0,
+ *     }
+ *
+ * @param {Locale} locale A Locale object.
+ * @returns {Object} A map of locale's cldrPlurals and their plural examples.
+ */
 export default function getPluralExamples(locale: Locale) {
     const pluralsCount = locale.cldrPlurals.length;
     const examples = {};
--- a/frontend/src/core/utils/fluent/getEmptyMessage.js
+++ b/frontend/src/core/utils/fluent/getEmptyMessage.js
@@ -3,9 +3,59 @@
 import { Transformer } from 'fluent-syntax';
 
 import flattenMessage from './flattenMessage';
+import isPluralElement from './isPluralElement';
 
 import type { FluentMessage } from './types';
+import type { Locale } from 'core/locale';
 
+
+const CLDR_PLURALS = ['zero', 'one', 'two', 'few', 'many', 'other'];
+
+
+/**
+ * Gather custom (numeric) plural variants
+ */
+function getNumericVariants(variants) {
+    return variants.filter(variant => {
+        return variant.key.type === 'NumberLiteral';
+    });
+}
+
+/**
+ * Generate a CLDR template variant
+ */
+function getCldrTemplateVariant(variants) {
+    return variants.find(variant => {
+        return CLDR_PLURALS.indexOf(variant.key.name) !== -1;
+    });
+}
+
+/**
+ * Generate locale plural variants from a template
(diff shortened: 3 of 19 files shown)


mathjazz:
The patch is now ready for a full review.

frontend/src/core/utils/fluent/isPluralElement.js
```diff
@@ -0,0 +1,29 @@
...
+
+import { CLDR_PLURALS } from 'core/plural';
+
+import type { PatternElement } from './types';
+
+
+/**
+ * Return true when AST element represents a pluralized string.
+ *
+ * Keys of all variants of such elements are either CLDR plurals or numbers.
```
Pike: In https://bugzilla.mozilla.org/show_bug.cgi?id=1385863, we decided to drop `other` from the tested plural categories. Notably, 'cause there's often `[other]` used in a non-plural-form sense.

frontend/src/core/utils/fluent/isPluralElement.js
```diff
@@ -0,0 +1,29 @@
...
+            element.expression && element.expression.type === 'SelectExpression'
+        )
+    ) {
+        return false;
+    }
+
+    return element.expression.variants.every(variant => {
+        return (
+            CLDR_PLURALS.indexOf(variant.key.name) !== -1 ||
+            variant.key.type === 'NumberLiteral'
```
Pike: I'd revert the order of the two, as you check for `undefined` in the `NumberLiteral` case. Also, doing the type check first, and then accessing the property of the other is better typed, I guess. Flow might not care, but .. ;-)

frontend/src/core/utils/fluent/isPluralElement.test.js
```diff
@@ -0,0 +1,65 @@
...
+        expect(isPluralElement(element)).toEqual(true);
+    });
+
+    it('returns true if all variant keys are numbers', () => {
+        const input = `
+my-entry =
+    { $num ->
+        [1] Hello!
+       *[2] World!
+    }`;
```
Pike: Is this the current behavior? I don't think that should be the case. In practice, this shouldn't be the case to begin with, but this is more likely a 1 vs many than an actual plural.

frontend/src/core/utils/fluent/isPluralElement.test.js
```diff
@@ -0,0 +1,65 @@
...
+my-entry =
+    { $num ->
+        [one] Hello!
+       *[1] World!
+    }`;
+        const message = parser.parseEntry(input);
+        const element = message.value.elements[0];
+
+        expect(isPluralElement(element)).toEqual(true);
+    });
```
Pike: As I stated on the actual code, `[1] foo *[other] bar` isn't actually a plural form semantically, I'd love to see a test for that.

Also, there's no test for something like `[one] foo *[male] other `, which I think would good to have here.

frontend/src/core/utils/fluent/types.js
```diff
@@ -1,8 +1,13 @@
...
 
+export type Identifier = {
+    type: string,
+    name: string,
+};
+
 export type Variant = {
     default: boolean,
-    key: string,
+    key: Identifier,
```
Pike: `Variant.key` is actually `NumberLiteral | Identifier`. This ties in to my comment about the check for `undefined` as you're treating the poor `NumberLiteral` as if it didn't exist 😢

frontend/src/modules/fluenteditor/components/RichTranslationForm.js
```diff
@@ -328,42 +332,98 @@ export class RichTranslationFormBase extends React.Component<InternalProps> {
         />;
     }
 
-    renderItem(value: string, path: MessagePath, label: string, className: ?string) {
+    renderPluralExample(pluralForm: number) {
+        const examples = locale.getPluralExamples(this.props.locale);
```
Pike: This is pretty excessive, in the case of Arabic, you're counting to 600.

I'd rather compute the examples in `renderElements`, and pass down the `examples | null` ?

frontend/public/static/locale/en-US/translate.ftl
```diff
@@ -195,6 +195,12 @@ entitydetails-GenericOriginalString--plural = Plural
 entitydetails-GenericOriginalString--singular = Singular
 
 
+## Rich editor
+## Renders a Rich editor for Fluent string editing
+
+fluenteditor-RichTranslationForm--plural-example = (e.g. <stress>{ $example }<stress>)
```
Pike: I'd argue that the plural category should be part of this message, too, to allow for ordering or other grammar needed in this phrase.

mathjazz:
Thanks for the review, that was very helpful!

> * can we go with just one transformer

I tried that initially by moving `visitTextElement` under `PluralsTransformer`, but I failed. It didn't replace all `TextElements`.

> * can we do no select expression for Chinese et al? I also didn't check if this patch changes what's a plural compared to old-translate, fixing that might be a follow up?

This patch keeps the behaviour of Translate.Current. Hence, I suggest we leave the following for the followup patch:
* No select expression for Chinese et al.
* Stop recognizing `other` as the plurals case.
* Require at least one variant key to be a CLDR plural (the `[1] one [2] two` case).

This is ready for another look. Sadly the localization change I made in 164f06c "doesn't work", i.e. the markup doesn't render and the test fails. No idea why.

frontend/src/modules/fluenteditor/components/RichTranslationForm.js
```diff
@@ -328,42 +332,102 @@ export class RichTranslationFormBase extends React.Component<InternalProps> {
...
+        const value = element.value;
+        if (typeof(value) !== 'string') {
+            return null;
+        }
+
+        const label = serializeVariantKey(variant.key);
+        let pluralForm = CLDR_PLURALS.indexOf(label);
+
+        if (!isPluralElement || pluralForm === -1) {
+            pluralForm = null;
```
Pike: If you compute the `pluralExample` up here, you can avoid the `pluralExamples[null]` in line 404.

frontend/src/core/locale/getPluralExamples.js
```diff
@@ -0,0 +1,30 @@
+/* @flow */
+
+import type { Locale } from './actions';
+
+
+export default function getPluralExamples(locale: Locale) {
```
Pike: One more thing, it took me a long time to understand what this function returns, i.e., `{"1": 1}`, right? Can you add a comment or maybe even add an indexer type? https://flow.org/en/docs/types/objects/#toc-objects-as-maps

frontend/src/core/locale/getPluralExamples.js
```diff
@@ -0,0 +1,30 @@
...
+        examples[locale.cldrPlurals[1]] = 2;
+    }
+    else {
+        // This variable is used in the pluralRule we eval in the while block.
+        let n = 0;
+        while (Object.keys(examples).length < pluralsCount) {
+            // This `eval` is not so much evil. The pluralRule we parse
+            // comes from our database and is not a user input.
+            // eslint-disable-next-line
+            const rule = eval(locale.pluralRule);
```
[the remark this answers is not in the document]
mathjazz: Sorry, but I can't get this to work. I'm getting "Cannot get locale.cldrPlurals[pluralRule] because new Function is not an array index". I though `pluralRule()` would help.

Note that this particular use of `eval()` has been approved during two of our security audits (bug 1338247 and when defining the CSP policy). I'm curious how is `new Function()` more secure than `eval()`.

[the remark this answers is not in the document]
Pike: Yeah, you need `const rule = pluralRule(n)`, too.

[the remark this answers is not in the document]
Pike: Re the sec impact, https://stackoverflow.com/questions/4599857/are-eval-and-new-function-the-same-thing has good comments. It's not cool, but it's less hot. ;-)

frontend/src/core/utils/fluent/getEmptyMessage.js
```diff
@@ -19,21 +68,45 @@ import type { FluentMessage } from './types';
  * @param {FluentMessage} source A Fluent AST to empty.
```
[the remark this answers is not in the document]
mathjazz: I'll update the comment. The function flattens the message, so it doesn't expect input to already be flat.

frontend/src/core/locale/getPluralExamples.js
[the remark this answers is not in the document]
mathjazz: Nope, still won't work ("no arguments expected by new Function()").

And yes, I've fixed `locale.PluralRule`.

mathjazz:
Thanks for the review. I've addressed all the comments but the [new Function()](https://github.com/mozilla/pontoon/pull/1379#discussion_r324603108) one.

https://github.com/mozilla/pontoon/pull/1379#discussion_r324127764 will be part of the followup.

frontend/src/core/locale/getPluralExamples.js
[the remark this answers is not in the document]
mathjazz: Actually this works now - the "no arguments expected" is a Flow error. I can't figure out how to bypass it beyond adding a series of `// $FLOW_IGNORE` lines.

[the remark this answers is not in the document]
Pike: Let's keep the eval then.

mathjazz:
> This patch keeps the behaviour of Translate.Current. Hence, I suggest we leave the following for the followup patch:
> * No select expression for Chinese et al.
> * Stop recognizing `other` as the plurals case.
> * Require at least one variant key to be a CLDR plural (the `[1] one [2] two` case).

Filed:
https://bugzilla.mozilla.org/show_bug.cgi?id=1581809
https://bugzilla.mozilla.org/show_bug.cgi?id=1581813

In https://github.com/mozilla/pontoon/pull/1379/commits/3f67dc7dc957785041f649f1b9c11b659a052ab0 I've also added a test case to cover the scenario uncovered by https://github.com/mathjazz/pontoon/pull/11/files#diff-0.

PR #1391, opened 2019-09-25
Bug 1543344 - Remove tab shortcut from the panel.

adngdb:


adngdb:
Oops, forgot to remove the translation entries.

PR #1394, opened 2019-09-25
Fix bug 1581594: Re-generate TM entries of FTL strings with comments

mathjazz:
Data migration in this patch re-generates the source columns in TranslationMemory table for FTL entries with comments.


mathjazz:
I've deployed the migration to stage and it fixes the problem:
https://mozilla-pontoon-staging.herokuapp.com/a/base/translationmemoryentry/3129926/change/

PR #1395, opened 2019-09-26
Fix bug 1583764: Properly detect multiple peer Fluent placeables

mathjazz:
I've also checked other similar (but non-Fluent) placeables (e.g. python*, java*) and they work fine.

### DIFF
--- a/frontend/src/core/placeable/parsers/fluentTerm.test.js
+++ b/frontend/src/core/placeable/parsers/fluentTerm.test.js
@@ -18,4 +18,15 @@ describe('fluentTerm', () => {
         expect(wrapper.find('mark')).toHaveLength(1);
         expect(wrapper.find('mark').text()).toEqual(mark);
     });
+
+    each([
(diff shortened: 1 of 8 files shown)


frontend/src/core/placeable/parsers/fluentTerm.test.js
```diff
@@ -18,4 +18,15 @@ describe('fluentTerm', () => {
...
+    each([
+        ['{-brand}', '{-vendor}', 'Hello {-brand} from {-vendor}'],
+    ])
+    .it('marks `%s` and `%s` in `%s`', (mark1, mark2, content) => {
+        const Marker = createMarker([fluentTerm]);
+        const wrapper = shallow(<Marker>{ content }</Marker>);
+        expect(wrapper.find('mark')).toHaveLength(2);
+        expect(wrapper.find('mark').at(0).text()).toEqual(mark1);
+        expect(wrapper.find('mark').at(1).text()).toEqual(mark2);
+    });
```
adngdb: Could you please add that same test to all the other parsers?

PR #1400, opened 2019-09-30
Fix bug 1584472: If translation is copied from Machinery, add text, not overwrite

mathjazz:


frontend/src/modules/fluenteditor/components/Editor.js
```diff
@@ -86,10 +86,18 @@ export class EditorBase extends React.Component<EditorProps, State> {
             this.props.entity &&
             !this.state.forceSource &&
             this.props.editor.translation !== prevProps.editor.translation &&
-            this.props.editor.changeSource === 'external' &&
+            this.props.editor.changeSource !== 'internal' &&
             typeof(this.props.editor.translation) === 'string'
         ) {
-            this.analyzeFluentMessage(this.props.editor.translation);
+            if (this.props.editor.changeSource === 'machinery') {
+                this.props.updateTranslation(prevProps.editor.translation, true);
```
adngdb: I'm not sure I understand this code, but this is an asynchronous action. You cannot expect that `this.props.editor.translation` will have the value in the next lines. You'd have to wait for the next render. And I don't understand why you would want to update the value to previous one?

frontend/src/modules/fluenteditor/components/Editor.js
```diff
@@ -86,10 +86,18 @@ export class EditorBase extends React.Component<EditorProps, State> {
...
+                if (typeof(this.props.editor.translation) === 'string') {
+                    this.props.addTextToEditorTranslation(this.props.editor.translation);
```
adngdb: I don't understand, shouldn't we replace instead of adding text?

mathjazz: Good point, we should replace.

frontend/src/modules/fluenteditor/components/Editor.js
mathjazz: Yeah, that's both, tricky and ugly.

1. On click on the Machinery "translation", we call [`editor.actions.update()`](https://github.com/mozilla/pontoon/pull/1400/commits/0621fbb77a3ba99b51fea6d68e48d067667471ff#diff-44f1b6ba42ed38be107c9c4b1bc18274L222).
2. So we overwrite `this.props.editor.translation` (which is what the bug is about).
3. As a consequence, this component here updates, and then we have the content of the Machinery "translation" in `this.props.editor.translation` and the previous translation in `prevProps.editor.translation`.
4. So we revert the translation to the previous translation and then "add text" to it.

adngdb: Oh, I see! It is very tricky indeed. This is definitely worth a comment!

mathjazz:
I tried to add another action in https://github.com/mozilla/pontoon/pull/1400/commits/fd4a52b325a07731cb2e971c5479c031d040a060, but that looks complex.

I'm playing with an option to select all text in the input and still abuse addTextToEditorTranslation. Any objections?

PR #1403, opened 2019-10-01
Fix bug 1583776 - Remove Translate.Next switch.

adngdb:
This patch removes the user setting controlling whether Translate.Next is enabled or not, and instead use only the waffle flag to choose if Translate.Next or translate current should be shown to the user. It also replaces the link to switch in the front-end with a message and link to our release announcement blog post.


frontend/src/core/user/components/UserControls.js
```diff
@@ -72,8 +72,17 @@ export class UserControlsBase extends React.Component<InternalProps> {
             <SignIn url={ user.signInURL } />
             }
 
-            { /* To be removed as part of bug 1527853. */ }
-            <AppSwitcher router={ router } user={ user } />
+            {/* To be removed as part of bug 1527853. */}
+            <p className='translate-next-notice'>
+                You are currently using Translate.Next.<br />
```
mathjazz: We have (active) Pontoon users on low-resolution displays. This text will make the right part of the header disappear on screens narrower than 900px if the project and resource names are longer than average.

If you change the text to something like "You are on Translate.Next.", the breaking point moves to around 800px, which should be sufficient.

PR #1405, opened 2019-10-03
Fix bug 1564837 - Make contributor's own public profle link more discoverable

abowler2:
When logged into Pontoon it was hard to find your own public profile link.

This fix addresses that by changing the URL from the profile widget to be 
`./contributors/[username]` instead of `./profile`.

### DIFF
--- a/frontend/src/core/user/components/UserMenu.js
+++ b/frontend/src/core/user/components/UserMenu.js
@@ -88,7 +88,7 @@ export class UserMenuBase extends React.Component<Props, State> {
                 { !user.isAuthenticated ? null :
                 <React.Fragment>
                     <li className="details">
-                        <a href={ `/contributors/${user.username}` }/>
+                        <a href={ `/contributors/${user.username}/` }>
                             <img src={ user.gravatarURLBig } alt="User avatar" />
                             <p className="name">{ user.nameOrEmail }</p>
                             <p className="email">{ user.email }</p>
--- a/frontend/src/core/user/components/UserMenu.js
+++ b/frontend/src/core/user/components/UserMenu.js
@@ -88,7 +88,7 @@ export class UserMenuBase extends React.Component<Props, State> {
                 { !user.isAuthenticated ? null :
                 <React.Fragment>
                     <li className="details">
-                        <a href="/{ user }/">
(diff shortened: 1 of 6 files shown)


pontoon/base/templates/django/base.html
```diff
@@ -64,7 +64,7 @@
               <div class="menu">
                 <ul>
                   {% if user.is_authenticated %}
-                  <li><a href="{% url 'pontoon.contributors.profile' %}"><i class="fa fa-user fa-fw"></i>My Profile</a></li>
+                  <li><a href="{% url 'pontoon.contributors.contributor.username', username=user %}"><i class="fa fa-user fa-fw"></i>My Profile</a></li>
```
mathjazz: Arguments of the [`url` template tag](https://docs.djangoproject.com/en/2.2/ref/templates/builtins/#url) in the Django templates are space-separated values, so you have to remove the comma.

pontoon/contributors/views.py
```diff
@@ -32,11 +32,6 @@
 
 
 @login_required(redirect_field_name='', login_url='/403')
```
mathjazz: You have to remove this line as well, otherwise this [decorator](https://docs.djangoproject.com/en/2.2/topics/http/decorators/) will make the `contributor_email` view (which handles http://localhost:8000/contributors/your.name@example.com/ for example) only accessible to logged in users.

frontend/src/core/user/components/UserMenu.js
```diff
@@ -88,7 +88,7 @@ export class UserMenuBase extends React.Component<Props, State> {
                 { !user.isAuthenticated ? null :
                 <React.Fragment>
                     <li className="details">
-                        <a href="/profile/">
+                        <a href={ `/contributors/${user.username}/` }/>
```
mathjazz: Please remove the trailing slash and we're good to go!

abowler2: Done :)

abowler2: Sorry, I removed the wrong one.  Resending now

abowler2:
Thank you. I appreciate all your help, and no worries on the last minute changes.

Glad to be here and look forward to more contributions.

PR #1407, opened 2019-10-04
Fix bug 1504702 - Add pseudo localization tool for development.

adngdb:
To enable the feature, set USE_PSEUDO_LOCALIZATION to `true` in core/l10n/index.js. Then all localized strings become weird, accented versions of themselves. This allows to spot places where we forgot to localized elements, notably.


adngdb:
Note to self: this needs to be documented in `frontend/README.md`.

frontend/src/core/l10n/actions.js
```diff
@@ -50,19 +51,39 @@ export function get(locales: Array<string>): Function {
     return async dispatch => {
         dispatch(request());
 
+        // Pseudo localization shows a weirdly translated UI, based on English.
+        // This is a development only tool that helps verifying that our UI
+        // is properly localized.
+        const usePseudoLocalization = (
+            process.env.NODE_ENV === 'development'
```
Pike: We allow pseudo in release Firefox, any reason to restrict that to `development`?

adngdb: Why is it allowed in Firefox release? 

My reasoning was that I didn't want to run the risk of showing a broken UI to users on production, so I stayed on the safe side. I also don't see the value of doing this on stage or prod, but maybe I'm missing something?

frontend/src/core/l10n/actions.js
```diff
@@ -50,19 +51,39 @@ export function get(locales: Array<string>): Function {
...
 
+        // For pseudo localization, we only want to serve English.
+        if (usePseudoLocalization) {
+            languages = ['en-US'];
+        }
+
         const bundles = await Promise.all(languages.map(locale => {
             return api.l10n.get(locale)
             .then(content => {
                 const bundle = new FluentBundle(locale);
```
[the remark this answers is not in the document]
adngdb: Aha, thanks!

adngdb:
@Pike This is now using the code from Firefox, and pseudolocalization is enabled via the URL. :-) r?

frontend/src/core/l10n/actions.js
Pike: It's entertaining to just put someone's Firefox into pseudo-localized, just to show off that you can.

For Pontoon, it would also be something where you could audit the UI for non-localizable strings w/out having to do that on a local setup.

I'd definitely play around with that on prod ;-)

adngdb: You got me at "entertaining".

PR #1408, opened 2019-10-06
Fix bug 1567402 - Make bugs tab disabled by default

abowler2:
The Bugs tab on the Team page (e.g. https://pontoon.mozilla.org/fr/bugs/) lists should not be enabled for those outside of Mozilla.

This fix adds an environment variable and then utilizes the value set in that variable to determine if the Bugs tab should display or not.

### DIFF
--- a/pontoon/settings/base.py
+++ b/pontoon/settings/base.py
@@ -782,8 +782,8 @@ USE_I18N = False
 # calendars according to the current locale
 USE_L10N = False
 
-#Enable Bugs tab on the team pages, pulling data from bugzilla.mozilla.org.
-#See bug 1567402 for details. A Mozilla-specific variable.
+# Enable Bugs tab on the team pages, pulling data from bugzilla.mozilla.org.
+# See bug 1567402 for details. A Mozilla-specific variable.
 ENABLE_BUGS_TAB = os.environ.get('ENABLE_BUGS_TAB', 'False') != 'False'
 
 # Bleach tags and attributes
--- a/pontoon/settings/base.py
+++ b/pontoon/settings/base.py
@@ -28,11 +28,13 @@ SECRET_KEY = os.environ['SECRET_KEY']
 # Is this a dev instance?
 DEV = os.environ.get('DJANGO_DEV', 'False') != 'False'
 
+MAINTAINER = os.environ.get('MAINTAINER', 'False') != 'False'
+
 DEBUG = os.environ.get('DJANGO_DEBUG', 'False') != 'False'
 
 HEROKU_DEMO = os.environ.get('HEROKU_DEMO', 'False') != 'False'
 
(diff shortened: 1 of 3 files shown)


pontoon/settings/base.py
```diff
@@ -28,11 +28,13 @@ def path(*args):
 # Is this a dev instance?
 DEV = os.environ.get('DJANGO_DEV', 'False') != 'False'
 
+MAINTAINER = os.environ.get('MAINTAINER', 'False') != 'False'
```
mathjazz: Could you move this setting to the bottom of the file, where we keep Pontoon app-specific variables?

I'd give the variable a more self-explanatory name (`ENABLE_BUGS_TAB`?) and add a comment in the line above it:
```python
# Enable Bugs tab on team pages, pulling team data from bugzilla.mozilla.org.
# See bug 1567402 for details. A Mozilla-specific variable.
ENABLE_BUGS_TAB = os.environ.get('ENABLE_BUGS_TAB', 'False') != 'False'
```

We also need to [document](https://github.com/mozilla/pontoon/blob/248facc10aef3e41c1157d0afcbb3549432fc88f/docs/admin/deployment.rst#environment-variables) the variable (note that variables are listed in alphabetic order there).

pontoon/settings/base.py
```diff
@@ -28,11 +28,13 @@ def path(*args):
...
+
 DEBUG = os.environ.get('DJANGO_DEBUG', 'False') != 'False'
 
 HEROKU_DEMO = os.environ.get('HEROKU_DEMO', 'False') != 'False'
 
-DJANGO_LOGIN = os.environ.get('DJANGO_LOGIN', 'False') != 'False' or HEROKU_DEMO
+DJANGO_LOGIN = os.environ.get('DJANGO_LOGIN', 'True') != 'False' or HEROKU_DEMO
```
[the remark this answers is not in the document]
abowler2: Oops...yes, this is for my local environment

pontoon/settings/base.py
abowler2: Within the documentation, do you want me to specify that it is a Mozilla specific variable?

mathjazz: Yeah, the docs should basically says what's written in the comment (you can leave out the reference to the bug).

pontoon/settings/base.py
```diff
@@ -782,6 +782,10 @@ def _allowed_hosts():
 # calendars according to the current locale
 USE_L10N = False
 
+#Enable Bugs tab on the team pages, pulling data from bugzilla.mozilla.org.
+#See bug 1567402 for details. A Mozilla-specific variable.
```
mathjazz: Please add spaces after `#`'s. The lack of them make tests fail, as you can see by the [Travis report](https://travis-ci.org/mozilla/pontoon/builds/594658552?utm_source=github_status&utm_medium=notification).

PR #1411, opened 2019-10-08
Fix bug 1586227 - Localize ProjectInfo

monikamaheshwari:
The ProjectInfo component needs to be localized. Wrapped the unlocalized strings in <Localized> components and gave them proper ids.


mathjazz:
Note: I've updated the title with the right bug number. We should also update it in the commit message before merging.

frontend/src/modules/projectinfo/components/ProjectInfo.js
```diff
@@ -54,11 +55,21 @@ export class ProjectInfoBase extends React.Component<Props, State> {
...
+              </Localized>
             </div>
             { !this.state.visible ? null :
             <aside className="panel">
-                <h2>Project Info</h2>
+              <Localized
+                  id="projectinfo-ProjectInfo--project-info-title"
+              >
+                  <h2>Project Info</h2>
+              </Localized>
```
adngdb: The indentation here is off. We use 4-space indentation in our front-end, please make sure to configure your editor correctly! :)

frontend/src/modules/projectinfo/components/ProjectInfo.js
```diff
@@ -54,11 +55,21 @@ export class ProjectInfoBase extends React.Component<Props, State> {
...
         return <div className="project-info">
             <div className="button" onClick={ this.toggleVisibility }>
-                <span className="fa fa-info"></span>
+              <Localized
+                  id="projectinfo-ProjectInfo--project-info"
+                  glyph={
+                      <span className="fa fa-info"></span>
+                  }
+              >
+              </Localized>
```
adngdb: There is no string content there, only an icon, so there is no need to translate it. Please revert this part and remove then entry in the .ftl file.

frontend/src/modules/projectinfo/components/ProjectInfo.js
```diff
@@ -1,6 +1,7 @@
 /* @flow */
 
 import * as React from 'react';
+import { Localized } from 'fluent-react';
```
adngdb: Oh, and this is incorrect. You might have started working on this a few days ago, and we changed the way we import fluent libraries since then. Please get the latest changes from master to see the correct way of importing.

monikamaheshwari: Done

adngdb:
Quick side-note @monikamaheshwari, you should not work directly on your `master` branch, but instead create a new branch for each bug you take. That will allow you to work on different things at the same time, keep in sync with the master branch, etc. Generally I highly recommend reading some tutorials about git.

monikamaheshwari:
Thanks:) Will keep that in mind to create separate branches for features

PR #1412, opened 2019-10-09
Fix bug 1583821 - Languages are deactivated when they don't have projects

Karskaya:
Change visible() method to available() to remove 404s for locale without non-system project


mathjazz:
Well done! :)

There are three more things that need to be fixed.

1. The "Start Localizing Now" button on the homepage (http://localhost:8000/) takes you to the team that matches your browser language. We only redirect to `visible()` teams, because until this patch other team pages 404ed. We should now redirect to `available()` teams. See https://github.com/mozilla/pontoon/blob/master/pontoon/homepage/views.py.

1. Teams that only have system projects enabled will appear on the http://localhost:8000/teams/ without stats - instead you'll see "Not synced yet". We should display "No projects enabled yet" (but only on the http://localhost:8000/teams/ page). 

1. On team pages without any projects (e.g. http://localhost:8000/af/) we should hide the table header and search box, but keep the "Request more projects" button. Instead of the table we should show a nice title saying "There are no projects enabled for the team yet.". For styling ideas see the `no_projects.html` template. If you want, I can help you with the exact HTML and CSS here.

pontoon/homepage/views.py
```diff
@@ -24,7 +24,7 @@ def homepage(request):
             return redirect('pontoon.teams.team', locale=user.profile.custom_homepage)
 
     # Guess user's team page or redirect to /teams
-    locale = get_project_locale_from_request(request, Locale.objects.visible())
+    locale = get_project_locale_from_request(request, Locale.objects.available())
```
mathjazz: Well done!

pontoon/teams/templates/teams/widgets/team_list.html
```diff
@@ -39,7 +39,7 @@ <h4>
       {% if chart.total_strings %}
         {{ ProgressChart.span(chart, chart_link, link_parameter, has_param) }}
       {% else %}
-        <span class="not-ready">Not synced yet</span>
+        <span class="not-ready">No projects enabled yet</span>
```
mathjazz: This is not OK, because this template is also used by the project dashboard, e.g. https://pontoon.mozilla.org/projects/firefox/. And we want to display "Not synced yet" when a new locale is added to the project and hasn't synced with the VCS yet (which exposes strings for localization for that particular locale within the project).

So you'll need another "else if" here that will figure out which title to use.

Karskaya: Changed. My variant has a new variable boolean for_teams_page. It's true if only /pontoon/teams/templates/teams/teams.html call it. 

Another idea is to give text variable to TeamList.item() with exact message.

mathjazz: Good job!

Could you indeed please use something like `not_ready_message` instead? The default value could be `Not synced yet` and the `teams.html` template will then override it.

Karskaya: @mathjazz Done with not_ready_message. Please, confirm that it's correct.

pontoon/teams/templates/teams/teams.html
```diff
@@ -86,7 +86,7 @@ <h1>
       {% set latest_activity = locale.get_latest_activity() %}
       {% set chart = locale.get_chart() %}
 
-      {{ TeamList.item(locale, main_link, chart_link, latest_activity, chart, link_parameter=True) }}
+      {{ TeamList.item(locale, main_link, chart_link, latest_activity, chart, link_parameter=True, not_ready_message = 'No projects enabled yet') }}
```
mathjazz: Nit: no spaces around `=`.

Karskaya:
> 3. On team pages without any projects (e.g. http://localhost:8000/af/) we should hide the table header and search box, but keep the "Request more projects" button. Instead of the table we should show a nice title saying "There are no projects enabled for the team yet.". For styling ideas see the `no_projects.html` template. If you want, I can help you with the exact HTML and CSS here.

1. `project.html` has new variable `hide_table`(true if no visible projects). It allows me to set display:none for search box, set visibility for `ProjectList.header` and show "There are no projects..." message
2. `ProjectList.header` from `project_list.html` has additional parameter visible(if false set `display:none`)
3. If search box is hidden, "Request new project" jumps down and "There are no projects..." message goes up. To remove this I add `.no-projects` style in `style.css`
4. request.js looks ugly now - i hope you can help me here. `toggleItem()` function:
 - First, I remove all changes for empty project-list.
 - At the end, check if list is empty again and hide all excessive/show message

It doesn't look like `no_projects.html`, but have similarities with empty contributors list(see attach)

New look of team page with no projects

Look of empty contributors page as example:

mathjazz:
Very nice work @Karskaya!

Before I look into the code, could you please make the styling of the following pages consistent when no items are found?
* /af/ (There are no projects enabled for the team yet.)
* /af/contributors/ (No contributors found.)
* /af/info/ (No team info available.)

You'll need to make the following changes:
* PROJECTS: Rename title to "No projects enabled for the team yet".
* CONTRIBUTORS: Remove the border.
* INTO: Remove 10px margin-bottom on `#info-wrapper .controls`.

Karskaya:
Thank you for support @mathjazz!!! It means so much for newbie like me.
I've done requested changes, please, review.

Karskaya:
One more question: if there are no projects available for requesting, how should it looks like?
Now it looks like this:
no-projects-for-request

mathjazz:
> Thank you for support @mathjazz!!! It means so much for newbie like me.
> I've done requested changes, please, review.

Happy to help! :) Thank you for your contributions!

mathjazz:
> One more question: if there are no projects available for requesting, how should it looks like?

Good point! Could you please hide the "Request..." button in this case?

Karskaya:
> > One more question: if there are no projects available for requesting, how should it looks like?
> 
> Good point! Could you please hide the "Request..." button in this case?

Will try my best!

pontoon/projects/templates/projects/widgets/project_list.html
```diff
@@ -3,8 +3,12 @@
 {% import 'widgets/deadline.html' as Deadline %}
 {% import 'widgets/priority.html' as Priority %}
 
-{% macro header(request=False) %}
+{% macro header(request=False, visible=True) %}
+  {% if visible %}
   <table class="table table-sort project-list item-list">
+  {% else %}
+  <table class="table table-sort project-list item-list" style="display:none">
```
mathjazz: We don't use style attribute, please use CSS instead.

pontoon/teams/static/js/request.js
```diff
@@ -33,6 +39,14 @@ var Pontoon = (function (my) {
                 .toggle(show);
           });
 
+          //Hide search box and table, show no result message if no project displayed
```
mathjazz: Nit: Space after `//`.

pontoon/teams/static/js/request.js
```diff
@@ -17,6 +17,12 @@ var Pontoon = (function (my) {
         if (type === 'locale-projects') {
           var localeProjects = $('#server').data('locale-projects');
 
+          //Show table and search box, hide no-result message, remove no=projects style
```
mathjazz: Nit: Space after `//`.

pontoon/teams/static/js/request.js
```diff
@@ -17,6 +17,12 @@ var Pontoon = (function (my) {
...
+            $('.search-wrapper:hidden').toggle();
+            $('.controls.no-projects').toggleClass('no-projects');
+            $('.project-list:hidden').toggle();
+            $('.no-results:visible').toggle();
```
mathjazz: Nit: Dedent this block by two spaces.

pontoon/teams/templates/teams/includes/projects.html
```diff
@@ -1,8 +1,17 @@
 {% import 'projects/widgets/project_list.html' as ProjectList with context %}
 
+{% set hide_table = (locale.project_set.visible().count() == 0) %}
+
 <div class="projects items">
+  
+  {% if hide_table %}
+  <menu class="controls no-projects">
+    <div class="search-wrapper small clearfix" style="display:none">
```
mathjazz: Same as above - avoid using the `style` attribute.

Karskaya:
> I wasn't exactly clear if you expected my review or if you work on hiding the "Request..." button, so I went ahead and left a few comments. :)

@mathjazz I was dive into django. I see it for the first time. Done with button. Will add changes according to your comments and will commit it soon.

> Nice work so far, only a few smaller things are left. Once you address the comments and make the button behave as expected, I'll do another review round.

Thank you! 
I've just realized that this pull request is one week old. Would it have been better to have separate tickets for each group of changes?
 e.g. 1. Remove 404s for locales without projects 2. Fix "No projects enabled yet" and make all similar pages consistent. 3. Hide table and search box for empty project list. 4. Hide "Request new projects button"

mathjazz:
I actually think it's fine to fix this things together. We try to make our codebase better with each patch. There's always a risk that followup patches don't land because of other priorities, and that's how technical debt is created.

Karskaya:
It works! @mathjazz, please, review. I do not feel satisfied about my code. I feel, there could be more elegant solution, but for now that's it. 
So,
1 Remove button, in case there are no projects to request
noRequest
2  Add all changes requested by you above 
3  Rewrite request.js part for Request new projects button. It looks a bit nicer now.

pontoon/teams/templates/teams/includes/projects.html
```diff
@@ -1,22 +1,33 @@
...
+
 <div class="projects items">
+  
+  {% if hide_table %}
+  <menu class="controls no-projects">
+    <div class="search-wrapper small clearfix hidden" >
+  {% else %}
   <menu class="controls">
     <div class="search-wrapper small clearfix">
+  {% endif %}
```
mathjazz: Instead of this, you could do:
```html
<menu class="controls {% if hide_table %}no-projects{% endif %}">
  <div class="search-wrapper small clearfix" >
```

And then change the CSS rule to:
```css
.controls.no-projects > .search-wrapper {
  display: none;
}
```

pontoon/base/static/css/style.css
```diff
@@ -1276,6 +1276,10 @@ body > form,
   position: relative;
 }
 
+.controls.no-projects {
+  height: 28px;
+}
+
```
mathjazz: This rule should go to `team.css`.

pontoon/teams/templates/teams/includes/projects.html
```diff
@@ -1,22 +1,33 @@
 {% import 'projects/widgets/project_list.html' as ProjectList with context %}
 
+{% set locale_projects = locale.available_projects_list() %}
+{# Hide if team has any projects #}
+{% set hide_table = (locale.project_set.visible().count() == 0) %}
+{# Hide if there are any projects available for request #}
+{% set hide_request_toggle = projects.exclude(slug__in=locale_projects).count() == 0 %}
```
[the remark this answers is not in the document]
Karskaya: Thank you for your feedback! Very informative as usual!

> please move this variable definitions to the view and pass them to templates as context

This will take some time. Perhaps, I'll have it tomorrow. Otherwise only on weekends. I hope it's not a problem.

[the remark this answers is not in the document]
mathjazz: Definitely not a problem. :)

pontoon/teams/static/js/request.js
```diff
@@ -33,6 +33,18 @@ var Pontoon = (function (my) {
...
+            $('.project-list').toggleClass('hidden', show);
+            $('menu.controls').toggleClass('no-projects', show);
+            $('.no-results:hidden').toggle();
+          }
+          else {
+            $('.project-list.hidden').toggleClass('hidden');
+            $('menu.controls.no-projects').toggleClass('no-projects');
+            $('.no-results:visible').toggle();
+          }
+
```
mathjazz: You can simplify this block to this:
```js
// Toggle table & search box, show no results message based on project visibility
var noProjects = $('.project-list tr.limited').length === 0;
$('.project-list').toggleClass('hidden', noProjects);
$('menu.controls').toggleClass('no-projects', noProjects);
$('.no-results').toggle();
```

Karskaya:
@mathjazz Please, review. BTW, has pontoon similar cases with projects or contributors pages?

pontoon/teams/views.py
```diff
@@ -77,21 +77,21 @@ def ajax_projects(request, locale):
...
 
-    no_visible_projects = (locale.project_set.visible().count() == 0)
+    locale_projects = locale.available_projects_list()
 
-    no_projects_to_request = (
-        Project.objects.visible().exclude(Q(locales=locale)).count() == 0
-    )
+    no_visible_projects = locale.project_set.visible().count() == 0
+
+    has_projects_to_request = projects.visible().exclude(locales=locale).count() > 0
```
mathjazz: Why use `.visible()` here? Above you'll see that `projects` already includes `Project.objects.visible()`.

Karskaya: Oh, so silly

Karskaya:
@mathjazz Please, review again. 
18 commits! Should I rebase and squash all those commits afterwards? Or as far as "This branch has no conflicts with the base branch" rebase in not necessary?

mathjazz:
That's OK. I'll squash before merging. :)

Karskaya:
@mathjazz Thank you for your time!!!

PR #1413, opened 2019-10-09
Fix bug 1501356 - Add ability to generate a permalink to a string

abowler2:
Add a feature to generate a permanent link to a string, without any (or at least mutable) queryset parameters (e.g. status filters).

This fix adds a link which when clicked will add the correct URL (without additional parameters) to the clipboard and display a message to let the user know the copy has been made.

### DIFF
--- a/frontend/public/static/locale/en-US/translate.ftl
+++ b/frontend/public/static/locale/en-US/translate.ftl
@@ -154,8 +154,10 @@ editor-UnsavedChanges--leave-anyway = Leave anyway
 
 
 ## Entity Details Navigation
-## Shows next/previous buttons.
+## Copies string link - Shows next/previous buttons.
 
+entitydetails-EntityNavigation--link = <glyph></glyph>Copy Link
+    .title = Copy Link to String
 entitydetails-EntityNavigation--next = <glyph></glyph>Next
     .title = Go To Next String (Alt + Down)
 entitydetails-EntityNavigation--previous = <glyph></glyph>Previous
--- a/frontend/public/static/locale/en-US/translate.ftl
+++ b/frontend/public/static/locale/en-US/translate.ftl
@@ -154,7 +154,7 @@ editor-UnsavedChanges--leave-anyway = Leave anyway
 
 
 ## Entity Details Navigation
-## Copies string link - Shows next/previous buttons.
+## Shows Copy Link and Next/Previous buttons.
 
 entitydetails-EntityNavigation--link = <glyph></glyph>Copy Link
     .title = Copy Link to String
@@ -365,6 +365,7 @@ notification--tt-checks-disabled = Translate Toolkit Checks disabled
 notification--make-suggestions-enabled = Make Suggestions enabled
 notification--make-suggestions-disabled = Make Suggestions disabled
 notification--entity-not-found = Can’t load specified string
+notification--string-link-copied = Link copied to clipboard
 
 
 ## OtherLocales Translation
--- a/frontend/src/core/notification/messages.js
+++ b/frontend/src/core/notification/messages.js
@@ -117,4 +117,10 @@ export default {
         </Localized>,
         type: 'error',
     },
+    'STRING_LINK_COPIED': {
+        content: <Localized id='notification--string-link-copied'>
+            Link copied to clipboard
+        </Localized>,
+        type: 'info',
+    },
 };
--- a/frontend/src/modules/entitydetails/components/EntityDetails.js
+++ b/frontend/src/modules/entitydetails/components/EntityDetails.js
@@ -168,6 +168,10 @@ export class EntityDetailsBase extends React.Component<InternalProps, State> {
         dispatch(machinery.actions.get(source, locale, pk));
     }
 
+    copyLinkToClipboard = () => {
+        console.log("hello");
+    }
+
     goToNextEntity = () => {
         const { dispatch, nextEntity, router } = this.props;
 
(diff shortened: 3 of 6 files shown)


frontend/public/static/locale/en-US/translate.ftl
```diff
@@ -154,8 +154,10 @@ editor-UnsavedChanges--leave-anyway = Leave anyway
 
 
 ## Entity Details Navigation
-## Shows next/previous buttons.
+## Copies string link - Shows next/previous buttons.
```
mathjazz: Nit: Shows Copy Link and Next/Previous buttons.

Please also update the docstring of the class in EntityDetails.js.

frontend/src/core/notification/messages.js
```diff
@@ -117,4 +117,10 @@ export default {
         </Localized>,
         type: 'error',
     },
+    'STRING_LINK_COPIED': {
+        content: <Localized id='notification--string-link-copied'>
+            Link copied to clipboard
+        </Localized>,
+        type: 'info',
+    },
```
mathjazz: Please also add this to the .ftl file.

frontend/src/modules/entitydetails/components/EntityDetails.js
```diff
@@ -168,6 +169,17 @@ export class EntityDetailsBase extends React.Component<InternalProps, State> {
         dispatch(machinery.actions.get(source, locale, pk));
     }
 
+    copyLinkToClipboard = () => {
+        const { locale, project, resource, entity } = this.props.parameters;
+
+        let string_link = `/${locale}/${project}/${resource}/?string=${entity}`
```
mathjazz: This is just a relative link, without the `https://example.com/` part, so it won't work when pasted.

frontend/src/modules/entitydetails/components/EntityNavigation.js
```diff
@@ -28,6 +29,10 @@ export default class EntityNavigation extends React.Component<Props> {
         document.removeEventListener('keydown', this.handleShortcuts);
     }
 
+    copyLinkToClipboard = () => {
+        this.props.copyLinkToClipboard();
+    }
+
```
mathjazz: I would actually not create this function and directly use `this.props.copyLinkToClipboard` below. For the same reason, I'd also remove the `goToNextEntity()` and `goToPreviousEntity()` and make sure we always use `this.props.goToNextEntity` and `this.props.goToPreviousEntity`.

frontend/public/static/locale/en-US/translate.ftl
abowler2: Can you please clarify which docstring in EntityDetails.js to update as I did not locate one for the Next/Previous or navigation in general. ~ Thanks

mathjazz: Sorry, I meant `EntityNavigation.js`:
https://github.com/mozilla/pontoon/blob/00f69cf509e0bb3b18d9c36c66d13798cac5c930/frontend/src/modules/entitydetails/components/EntityNavigation.js#L19

frontend/src/modules/entitydetails/components/EntityDetails.js
```diff
@@ -168,6 +169,17 @@ export class EntityDetailsBase extends React.Component<InternalProps, State> {
         dispatch(machinery.actions.get(source, locale, pk));
     }
 
+    copyLinkToClipboard = () => {
+        const { locale, project, resource, entity } = this.props.parameters;
+
+        let string_link = `https://pontoon.mozilla.org/${locale}/${project}/${resource}/?string=${entity}`
```
mathjazz: That's better. But it will only work on Mozilla's instance of Pontoon, which runs on `pontoon.mozilla.org`. There are other organizations that use self-hosted installs of Pontoon.

frontend/src/modules/entitydetails/components/EntityNavigation.js
```diff
@@ -28,14 +29,6 @@ export default class EntityNavigation extends React.Component<Props> {
...
     }
 
-    goToNextEntity = () => {
-        this.props.goToNextEntity();
-    }
-
-    goToPreviousEntity = () => {
-        this.props.goToPreviousEntity();
-    }
-
```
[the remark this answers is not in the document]
abowler2: I saw that... after the test failed :(  

It should be fixed in commit `03b301c` above, but please let me know if it still need attention.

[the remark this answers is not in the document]
mathjazz: You are correct! It's all good now!

mathjazz:
Since I can't find you on IRC... Maybe you want to work on this, next? :)

https://bugzilla.mozilla.org/show_bug.cgi?id=1587910

frontend/src/modules/entitydetails/components/EntityDetails.js
```diff
@@ -168,6 +169,17 @@ export class EntityDetailsBase extends React.Component<InternalProps, State> {
         dispatch(machinery.actions.get(source, locale, pk));
     }
 
+    copyLinkToClipboard = () => {
+        const { locale, project, resource, entity } = this.props.parameters;
+
+        const string_link = `${location.protocol}//${location.host}/${locale}/${project}/${resource}/?string=${entity}`
```
mathjazz: Sorry for misleading you, you actually have to use `window.location...` instead of `location...`, otherwise eslint fails.

Please also keep line length under 100 characters. You could do something like this:
```js
const { locale, project, resource, entity } = this.props.parameters;
const { host, protocol } = window.location;

const link = `${protocol}//${host}/${locale}/${project}/${resource}/?string=${entity}`;
```

Please also add comma at the end.

abowler2:
Looks like I wasn't able to find you on IRC today...  

If you happen upon this before Monday, please just let me know what you would like me to work on next.  
FWIW, I took a look at https://bugzilla.mozilla.org/show_bug.cgi?id=1243137  and am pretty sure I worked out what needs to be done.

PR #1414, opened 2019-10-10
Fix bug 1251186 - Fix multiple placeable detection errors

adngdb:
I made a commit for each error. :-)


frontend/src/core/placeable/parsers/pythonFormattingVariable.test.js
```diff
@@ -15,15 +15,28 @@ describe('pythonFormattingVariable', () => {
         ['%(cows.number)d', 'There were %(cows.number)d cows'],
         ['%(number of cows)d', 'There were %(number of cows)d cows'],
         ['%(number)03d', 'There were %(number)03d cows'],
-        ['%(number) 3d', 'There were %(number) 3d cows'],
         ['%(number)*d', 'There were %(number)*d cows'],
         ['%(number)3.1d', 'There were %(number)3.1d cows'],
         ['%(number)Ld', 'There were %(number)Ld cows'],
+        ['%s', 'path/to/file_%s.png'],
+        ['%s', 'path/to/%sfile.png'],
```
mathjazz: I'm curious - why did you decide to add this test case? Is it related to the patch?

frontend/src/core/placeable/parsers/pythonFormattingVariable.test.js
```diff
@@ -15,15 +15,28 @@ describe('pythonFormattingVariable', () => {
...
         const Marker = createMarker([pythonFormattingVariable]);
         const wrapper = shallow(<Marker>{ content }</Marker>);
         expect(wrapper.find('mark')).toHaveLength(1);
         expect(wrapper.find('mark').text()).toEqual(mark);
     });
+
+    each([
+        ['10 % complete'],
+        // We used to match '%(number) 3d' here, but don't anymore to avoid
+        // false positives.
```
mathjazz: Nit: A ref to the bug would be helpful.

mathjazz: BTW: sadly the issue reported in https://bugzilla.mozilla.org/show_bug.cgi?id=1251186#c0 is not fixed by this changeset, because the same placeable gets falsely detected by the `stringFormattingVariable`. (I also check that no other parser detects it.)

frontend/src/core/placeable/parsers/numberString.js
```diff
@@ -18,7 +18,7 @@ import { Localized } from '@fluent/react';
  * https://github.com/translate/translate/blob/2.3.1/translate/storage/placeables/general.py#L72
  */
 const numberString = {
-    rule: /([-+]?[0-9]+([\u00A0.,][0-9]+)*)/u,
+    rule: /([-+]?[0-9]+([\u00A0.,][0-9]+)*)\b/u,
```
mathjazz: That's a valid fix. But "3D", "A4" and the likes are candidates for placeables, so I wonder if we should make `allCapitalsString` catch them?

frontend/src/core/placeable/parsers/pythonFormattingVariable.test.js
adngdb: Because I was learning more about regexp and thought that what I was doing could break this specific case, which I didn't want to happen. Figured I might as well leave that around.

frontend/src/core/placeable/parsers/pythonFormattingVariable.test.js
adngdb: Woopsy, thanks for catching that!

frontend/src/core/placeable/parsers/pythonFormattingVariable.test.js
mathjazz: OK, makes sense!

frontend/src/core/placeable/parsers/numberString.js
mathjazz: BTW, how hard would it be to make `allCapitalsString` not catch this at all?
https://pontoon.mozilla.org/translate/sl/firefox/all-resources/?string=204038

I mean, would it be easier than catching all Latin characters? Maybe we can require whitespace/start/end around the capitals.

frontend/src/core/placeable/parsers/shortCapitalNumberString.js
```diff
@@ -0,0 +1,30 @@
+/* @flow */
+
+import * as React from 'react';
+import { Localized } from '@fluent/react';
+
+
+/**
+ * Marks 2-letters-long terms containing a combination of a capital letter and
+ * a number.
```
mathjazz: Why 2-letter only? There are longer Numeronyms, e.g. W3C.

adngdb: That will be caught by the `allCapitalsString` placeable.

frontend/src/core/placeable/parsers/numberString.js
adngdb: That would mean we wouldn't catch `CAPS.` for example, because it's followed by a dot, which is effectively a word boundary but not an empty space. We would need to add something very close to the `\b` word boundary rule but not exactly, and I'm not sure that's wise to do as it's bound to lead to other bugs. And to a very complex regepx anyway.

I don't think there's a good way out of this problem with the current state of RegExp in JavaScript. We just need the browsers to improve their support of unicode. :/

frontend/src/core/placeable/parsers/shortCapitalNumberString.js
mathjazz: Right!

PR #1417, opened 2019-10-12
Fix bug 1545964 - Confusing Copy Into Translate tooltip

monikamaheshwari:
[Translate.Next] Button to accept a suggested translation has an inaccurate/confusing tooltip


adngdb:
Hey @monikamaheshwari, this is looking fine, but I'm wondering if we shouldn't do things a bit differently here. I have commented in the bug, and I'm putting the review here on hold until we reach an agreement on how exactly to solve this issue. Thank you! :)

adngdb:
@monikamaheshwari Please see https://bugzilla.mozilla.org/show_bug.cgi?id=1545964#c10 for what to do. What you did there is a good start, but it needs a bit more work. :-)

I would also like you to clean the history of your branch. It shows 5 commits but only one is interesting, the others are junk. Please get rid of the useless commits, then force-push to your branch to replace the history. 

(Note that you should only rarely force-push, as it makes life more difficult for reviewers. But in this case, you'll have to clean the history before we can merge this, so you should do it first thing, before adding other commits to this branch. )

frontend/src/modules/history/components/Translation.js
```diff
@@ -206,6 +206,12 @@ export class TranslationBase extends React.Component<InternalProps, State> {
         );
 
         let className = 'translation ' + this.getStatus();
+
+        if(!user.isAuthenticated || isReadOnlyEditor) {
+            // This user can not copy translate
+            className += ' can-not-copy';
```
adngdb: The proper form would be `cannot` instead of `can-not`.

monikamaheshwari: Done

monikamaheshwari:
@adngdb Thanks. Did the test cases failure changes.

monikamaheshwari:
What styled is needed to be there in 1st and 2nd scenario? I thought they didn't have right to do anything so disabled it.

mathjazz:
> What styled is needed to be there in 1st and 2nd scenario? I thought the user didn't have right to do anything so disabled review buttons.

The style should remain the same as it is now on the `master` branch. Buttons are disabled, but they are still there to show the status of the translation.

mathjazz:
The opacity is lower, so the buttons already look disabled.

monikamaheshwari:
Ooo okay

monikamaheshwari:
> We're getting there, good work! I made a few comments related to code.
> 
> There's a bug though: when I'm logged out, if I hover one of the buttons, it stills shows the cursor pointer, and the icon changes. I'd expect that not to be the case.

Hey can you guide me what should be proper style at that time as I am confused.

adngdb:
When you see the button, it looks like this: 

pontoon-btn-normal

Now, if you're logged in and have permissions to unapprove the translation, if you put your mouse over the button, it turns to this: 

pontoon-btn-hover

But, if you're logged out or don't have sufficient permissions, it should *not* turn like this, and instead should stay like in the first example. That is not the case at the moment.

monikamaheshwari:
Screenshot 2019-10-22 at 6 08 34 PM
Done.

adngdb:
Almost there! The cursor just needs to not turn into a pointer in the same cases, and I think we're good. :-)

frontend/src/modules/history/components/Translation.css
```diff
@@ -6,7 +6,11 @@
     display: flex;
 }
 
-.history .translation:hover {
+.history .translation.cannot-copy {
+    pointer-events: none;
+}
+
+.history .translation.can-copy:hover {
```
adngdb: This should not be limited to `can-copy` only. Please remove the `.can-only` bit, and since I believe that class is never used, please don't set it in the code.

monikamaheshwari: If I am not setting that to `can-copy` then on hovering buttons. Translation background changes

adngdb: Unless I'm misunderstanding, yes, that is the behavior I expect. The background of a translation should become brighter when you are hovering anywhere over it.

monikamaheshwari:
> **In the 1st scenario (User is logged out):**
> Translation background doesn't change on hover (good), but it does when you hover the review buttons. That's confusing.

@adngdb These scenarios

adngdb:
After discussing this with @mathjazz, we agreed that it's better to keep the background brighter on hover, as it is consistent with the rest of the app. However, he did notice something that I missed on the status buttons. 

When logged out, the inactive buttons are filled, like this:

screenshot_2019-10-22_at_17 51 01

And they should actually look like this:

screenshot_2019-10-22_at_17 50 48

monikamaheshwari:
> After discussing this with @mathjazz, we agreed that it's better to keep the background brighter on hover, as it is consistent with the rest of the app. However, he did notice something that I missed on the status buttons.
> 

As I am setting pointer-events:none to not show title on :hover background is not changing.

mathjazz:
Icons look OK now, thanks!

But the Translation hover behavior is now back into the buggy state. There's no hover effect; unless you hover the review buttons; then it appears.

I'm fine with both options when the editor is read-only (hover effect or no hover effect), but we should make it consistent with the Machinery and Locales tabs. Including the mouse pointer style.

monikamaheshwari:
Is not hover effect OK if I implement same on Machinery and Locale tab as both are also getting `copy into translation` title even when user is logged out.

mathjazz:
I've updated my comment, so hopefully it makes more sense now. No hover effect is OK if you also implement it on Machinery and the Locales tab. And good point about the title - we should remove that also from the Machinery/Locales tab when the editor is read-only.

mathjazz:
That works fine! The only thing we should change is to make links in the Machinery and Locales tab work. Please restore that functionality.

monikamaheshwari:
Can you suggest a title on locale link as other wise it is showing `copy into translation` on hover.

mathjazz:
I'd suggest something like this:
"Visit string in German (de)"

Note that you'll also need to make it localizable.

mathjazz:
Could you please replace "Visit" with "Open"?

Other than that this looks good to me. I'll let @adngdb do the final review tomorrow.

frontend/public/static/locale/fr/translate.ftl
```diff
@@ -206,6 +206,9 @@ notification--entity-not-found = Impossible de charger la chaîne spécifiée
 otherlocales-Translation--copy =
     .title = Copier la traduction
 
+otherlocales-Translation--header-link = 
+    .title = Open string in { $locale } ({ $code })
+
```
mathjazz: This file contains French localizations, so we should either remove this block or provide an actual translation.

frontend/src/modules/history/components/Translation.js
```diff
@@ -224,6 +224,12 @@ export class TranslationBase extends React.Component<InternalProps, State> {
         );
 
         let className = 'translation ' + this.getStatus();
+
+        if (!user || !user.isAuthenticated || isReadOnlyEditor) {
```
mathjazz: Note that `isReadOnlyEditor` covers the `!user.isAuthenticated` case, so I think it can be the only condition here.

frontend/src/modules/history/components/Translation.js
```diff
@@ -224,6 +224,12 @@ export class TranslationBase extends React.Component<InternalProps, State> {
...
+            // This user can not copy into translate
```
mathjazz: Nit: "Copying into the editor is not allowed"

frontend/src/modules/otherlocales/components/Translation.js
```diff
@@ -53,15 +57,23 @@ export default class Translation extends React.Component<Props> {
...
+                        $code={ translation.code }
                     >
-                        { translation.locale }
-                        <span>{ translation.code }</span>
-                    </a>
+                        <a
+                            href={ `/translate/${translation.code}/${parameters.project}/${parameters.resource}/?string=${parameters.entity}` }
+                            target='_blank'
+                            rel='noopener noreferrer'
+                            title={ 'Open string in { $locale } ({ $code })' }
```
mathjazz: I don't think we need curly brackets here.

frontend/public/static/locale/fr/translate.ftl
```diff
@@ -206,6 +206,9 @@ notification--entity-not-found = Impossible de charger la chaîne spécifiée
 otherlocales-Translation--copy =
     .title = Copier la traduction
 
+otherlocales-Translation--header-link = 
+    .title = Chaîne ouvre en { $locale } ({ $code })
```
adngdb: Sorry, that's not correct French (I'm guessing Google Translate? :smile: ). Please remove this instead, as we'll want to let Pontoon users do the translations themselves in the future.

monikamaheshwari: Yes sure :)

PR #1420, opened 2019-10-14
Fixed Bug 1243137 - Implement empty string filter

abowler2:
This fix adds a new filter in the "Extra filters" category that will search for empty strings.

### DIFF
--- a/frontend/src/modules/search/components/SearchBox.css
+++ b/frontend/src/modules/search/components/SearchBox.css
@@ -98,6 +98,11 @@
     content: "";
 }
 
+.search-box .empty .status:before {
+    color: #AAAAAA;
+    content: "\f111";
+}
+
 .search-box .rejected .status:before {
     color: #AAAAAA;
     content: "";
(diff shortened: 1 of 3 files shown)


frontend/src/modules/search/components/SearchBox.css
```diff
@@ -98,6 +98,11 @@
     content: "";
 }
 
+.search-box .empty .status:before {
+    color: #AAAAAA;
+    content: "\f111";
+}
+
```
mathjazz: I thought we should use the "empty" circle. The easiest way to achieve it is it to add `font-weight: 100;` here.

And then you'll just need to make sure `font-weight: 900;` is used when the empty icon is selected or hovered, but only in the menu (not in the search field). This should do:
```css
.search-box .menu .empty.selected .status:before,
.search-box .menu .empty .status:hover:before {
    font-weight: 900;
}
```

While you're at it, please also fix the bad indent in line 93.

PR #1422, opened 2019-10-14
Fix bug 1508474 - Add check for unsaved changes before executing review actions

Henikilan:
Displays "You have unsaved changes" message on executing review actions if there is a new translation in the translation input.

### DIFF
--- a/frontend/src/modules/entitydetails/components/EntityDetails.js
+++ b/frontend/src/modules/entitydetails/components/EntityDetails.js
@@ -242,12 +242,19 @@ export class EntityDetailsBase extends React.Component<InternalProps, State> {
 
     deleteTranslation = (translationId: number) => {
         const { parameters, pluralForm, dispatch } = this.props;
-        dispatch(history.actions.deleteTranslation(
-            parameters.entity,
-            parameters.locale,
-            pluralForm,
-            translationId,
-        ));
+        dispatch(
+            unsavedchanges.actions.check(
+                this.props.unsavedchanges,
+                () => {
+                    dispatch(history.actions.deleteTranslation(
+                        parameters.entity,
+                        parameters.locale,
+                        pluralForm,
+                        translationId,
+                    ));
+                }
+            )
+        );
     }
 
     /*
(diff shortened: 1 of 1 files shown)


frontend/src/modules/entitydetails/components/EntityDetails.js
```diff
@@ -242,12 +242,19 @@ export class EntityDetailsBase extends React.Component<InternalProps, State> {
...
+                () => {
+                    dispatch(history.actions.deleteTranslation(
+                        parameters.entity,
+                        parameters.locale,
+                        pluralForm,
+                        translationId,
+                    ));
+                }
+            )
+        );
```
adngdb: This one doesn't change the content of the editor, as far as I can tell, so it's not needed to run the unsaved actions check here.

mathjazz: Correct :)

frontend/src/modules/entitydetails/components/EntityDetails.js
```diff
@@ -257,16 +264,23 @@ export class EntityDetailsBase extends React.Component<InternalProps, State> {
...
-        dispatch(history.actions.updateStatus(
-            change,
-            selectedEntity,
-            locale,
-            parameters.resource,
-            pluralForm,
-            translationId,
-            nextEntity,
-            router,
-        ));
```
mathjazz: I'd also add a comment here along the lines of:
```
No need to check for unsaved changes in `EditorBase.updateTranslationStatus()`,
because it cannot be triggered for the use case of bug 1508474.
```

frontend/src/modules/entitydetails/components/EntityDetails.js
Henikilan: you are right, fixed 👍

frontend/src/modules/entitydetails/components/EntityDetails.js
Henikilan: Done

PR #1416, opened 2019-10-11
Fix Bug 1369976 Add sort by completion in the translation resource menu

nadinda:
Add a feature to sort the resource by completion in the translation interface's resource menu.

This fix adds two buttons placed on top of the resource paths and the translation completion progress in the resource menu drop down. The initial state would sort the resource name alphabetically. Then if the percent completion is clicked, it will sort the list by the total completed translation. User can also cancel the sort by completion and back to initial state by clicking the 'RESOURCE' title above the resource paths.

PR #1424, opened 2019-10-15
Fix Bug 1588468 - Prevent long Locale codes to break layout

nadinda:
This fix make a long locale code in [teams] (https://pontoon.mozilla.org/teams/) page cut out and replaced by ellipsis (...) instead of showing them in two lines. This way long codes will not break the layout of the column or the table.

The outcome of this change :
Bug 1588468

### DIFF
--- a/pontoon/base/static/css/table.css
+++ b/pontoon/base/static/css/table.css
@@ -130,6 +130,14 @@ table.table {
   color: #F36;
 }
 
+.table td.code div{
+  width: 70px;
+  display: block;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
+
 .table td.code a {
   color: #7BC876;
   line-height: 47px;
--- a/pontoon/base/static/css/table.css
+++ b/pontoon/base/static/css/table.css
@@ -130,9 +130,8 @@ table.table {
   color: #F36;
 }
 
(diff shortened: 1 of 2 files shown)


pontoon/base/static/css/table.css
```diff
@@ -130,6 +130,14 @@ table.table {
   color: #F36;
 }
 
+.table td.code div{
```
mathjazz: Nit: please add space between `div` and `{`.

pontoon/base/static/css/table.css
```diff
@@ -130,6 +130,14 @@ table.table {
...
+  width: 70px;
+  display: block;
```
mathjazz: I think we don't need `display: block;`, because the DIV element is a [block-level element](https://developer.mozilla.org/en-US/docs/Web/HTML/Block-level_elements).

nadinda: Ah I see, thanks for the info! And it's done ;)

nadinda:
Wohoo! All thanks to you as a mentor @mathjazz. Let me know If I can work on another bug.

mathjazz:
Let's sync tomorrow to find more bugs for you to work on.

Until then, there's some work left in https://github.com/mozilla/pontoon/pull/1416. :)

PR #1426, opened 2019-10-15
Fix bug 1586233 - Localize search panel statuses and filters

monikamaheshwari:
Localize search panel statuses and filters

### DIFF
--- a/frontend/public/static/locale/en-US/translate.ftl
+++ b/frontend/public/static/locale/en-US/translate.ftl
@@ -464,6 +464,15 @@ search-FiltersPanel--heading-status = Translation Status
 search-FiltersPanel--heading-tags = Tags
 search-FiltersPanel--heading-extra = Extra Filters
 search-FiltersPanel--heading-authors = Translation Authors
+search-FiltersPanel--status-name-all = All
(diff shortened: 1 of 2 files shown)


frontend/public/static/locale/en-US/translate.ftl
```diff
@@ -464,6 +464,15 @@ search-FiltersPanel--heading-status = Translation Status
...
 search-FiltersPanel--heading-authors = Translation Authors
+search-FiltersPanel--status-name-all = All
+search-FiltersPanel--status-name-translated = Translated
+search-FiltersPanel--status-name-fuzzy = Fuzzy
+search-FiltersPanel--status-name-warnings = Warnings
+search-FiltersPanel--status-name-errors = Errors
+search-FiltersPanel--status-name-missing = Missing
+search-FiltersPanel--status-name-unreviewed = Unreviewed
+search-FiltersPanel--extra-name-unchanged = Unchanged
+search-FiltersPanel--extra-name-rejected = Rejected
```
mathjazz: You're missing the newly added Empty Filter here.

monikamaheshwari: Yeah. Will add that also

PR #1427, opened 2019-10-15
Fix bug 1588787 - [translate] Improve unsaved changes message

Karskaya:
Change message and button caption for unsaved changes


Karskaya:
@mathjazz I've left button class name as "proceed anyway", to be similar with other anyway buttons

PR #1429, opened 2019-10-16
Fix bug 1586829 - Show avatars in translations

Henikilan:
image

### DIFF
--- a/frontend/src/core/user/components/UserMenu.css
+++ b/frontend/src/core/user/components/UserMenu.css
@@ -12,7 +12,7 @@
 .user-menu .selector .menu-icon,
 .user-menu .menu li.details img {
     border: 2px solid #4d5967;
-    border-radius: 100%;
+    border-radius: 4px;
 }
 
 .user-menu .selector img {
--- a/frontend/src/core/user/components/UserMenu.css
+++ b/frontend/src/core/user/components/UserMenu.css
@@ -12,7 +12,7 @@
 .user-menu .selector .menu-icon,
 .user-menu .menu li.details img {
     border: 2px solid #4d5967;
-    border-radius: 4px;
+    border-radius: 6px;
 }
 
 .user-menu .selector img {
--- a/frontend/src/modules/history/components/Translation.css
+++ b/frontend/src/modules/history/components/Translation.css
@@ -21,7 +21,7 @@
     border: 1px solid #5E6475;
 }
 
-.history .translation .avatar img:hover {
+.history .translation:hover .avatar img {
     border-color: #637283;
 }
 
--- a/frontend/src/modules/history/components/Translation.css
+++ b/frontend/src/modules/history/components/Translation.css
@@ -3,13 +3,29 @@
     cursor: pointer;
     padding: 5px 10px 20px 10px;
     transition: all 0.1s ease-out;
+    display: flex;
 }
 
 .history .translation:hover {
     background: #4D5967;
 }
 
-.history .translation > header {
+.history .translation .avatar {
+    padding-right: 8px;
+    position: relative;
+    top: 3px;
+}
+
+.history .translation .avatar img {
+    border-radius: 4px;
+    border: 1px solid #4D5967;
+}
+
+.history .translation .content {
+    flex-grow: 1;
+}
+
+.history .translation .content > header {
     color: #AAAAAA;
     display: block;
     font-size: 11px;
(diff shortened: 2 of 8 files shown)


frontend/src/modules/history/components/Translation.css
```diff
@@ -3,13 +3,29 @@
...
-.history .translation > header {
+.history .translation .avatar {
+    padding-right: 8px;
+    position: relative;
+    top: 3px;
+}
+
+.history .translation .avatar img {
+    border-radius: 4px;
+    border: 1px solid #4D5967;
```
mathjazz: Let's use `#5E6475` instead (same as translation borders) and make it `#637283` on hover. Sorry for not pointing that out in the spec!

frontend/src/modules/history/components/Translation.js
```diff
@@ -123,6 +123,24 @@ export class TranslationBase extends React.Component<InternalProps, State> {
         return 'Not reviewed yet';
     }
 
+    renderUserAvatar() {
+        const { translation } = this.props;
+
+        if (!translation.uid) {
+            return <img src={ translation.user_gravatar_url_small } alt="User avatar" />;
```
mathjazz: If translation lacks `uid`, it also lacks `user_gravatar_url_smal`. Which results it:
Screenshot 2019-10-16 at 14 14 53

Obviously, that's not great. To mitigate the problem, we can either:
1. not show the avatar at all (not even the `<div.avatar>`), which will make those translations stand out more
1. use `/static/img/anon.jpg` - our fallback when user doesn't have an image set
1. find a better image to be used for strings without Translation authors (i.e. imported from version control)

Which option would you prefer? I'd probably go with #2 for now.

Henikilan: Yes, I also prefer #2

frontend/src/core/user/components/UserMenu.css
```diff
@@ -12,7 +12,7 @@
 .user-menu .selector .menu-icon,
 .user-menu .menu li.details img {
     border: 2px solid #4d5967;
-    border-radius: 100%;
+    border-radius: 4px;
```
mathjazz: Nice! You still need to make this change for the avatars in the [Filter menu](https://mozilla-l10n.github.io/localizer-documentation/tools/pontoon/search_filters.html).

frontend/src/modules/history/components/Translation.css
```diff
@@ -18,7 +18,11 @@
 
 .history .translation .avatar img {
     border-radius: 4px;
-    border: 1px solid #4D5967;
+    border: 1px solid #5E6475;
+}
+
+.history .translation .avatar img:hover {
```
mathjazz: We should change the border color on `.translation:hover` (when the background changes), not `.img:hover`.

frontend/src/modules/history/components/Translation.js
```diff
@@ -127,7 +127,7 @@ export class TranslationBase extends React.Component<InternalProps, State> {
         const { translation } = this.props;
 
         if (!translation.uid) {
-            return <img src={ translation.user_gravatar_url_small } alt="User avatar" />;
+            return <img src='/static/img/anon.jpg' alt="User avatar" />;
```
mathjazz: Nit: Please also use single quotes in the `alt` attribute.

frontend/src/core/user/components/UserMenu.css
Henikilan: Sorry, missed this one :)

Henikilan: BTW, the filters button is not obvious, it was hard to find it for me.
image
Should it be an image like in the documentation screenshot?

Henikilan:
@mathjazz 
Fixed :)

One question about style.css - how to apply them in local environment?
I made changes, but http://localhost:8000/static/css/style.css do not update.
Restarting the server does not help.

mathjazz:
You'll need to `make build` (sadly), which is a pretty long process. And then `make run` again and you should be good.

Could you please file a bug about changes to static files outside the `frontend` folder requiring `make build`?

frontend/src/modules/history/components/Translation.css
```diff
@@ -3,13 +3,33 @@
...
-.history .translation > header {
+.history .translation .avatar {
+    padding-right: 8px;
+    position: relative;
+    top: 3px;
+}
+
+.history .translation .avatar img {
+    border-radius: 6px;
+    border: 1px solid #5E6475;
```
mathjazz: You will hate me for this I know. :) Could you set the border to 2px instead of 1px? It's against the spec, but also looks inconsistent with the rest of the UI, where we always use 2px.

Henikilan: Fixed :)
Actually the opposite, I appreciate your code review. That's a great way to learn. Thank you for that!

PR #1431, opened 2019-10-16
Fix bug 1374605 - Show file-level comments

abowler2:
I'm not able to get the function to populate the new fields "group_comment" and "file_comment" working.  

Here is a link to the log from my most recent attempt to force sync showing the current error that is occurring: https://pastebin.com/zfRf7M3m


pontoon/sync/vcs/models.py
```diff
@@ -714,7 +714,7 @@ class VCSTranslation(object):
     translation for that plural form.
     """
     def __init__(
-        self, key, strings, comments, fuzzy,
+        self, key, strings, comments, group_comments, file_comments, fuzzy,
```
mathjazz: You'll need to set the default value of `group_comments` and `file_comments` to `None`.

Otherwise other classes that inherit from `VCSTranslation` and don't have the constructor (`__init__`) set accordingly (like in your case `/app/pontoon/sync/formats/po.py`) won't work.

abowler2:
I can get all 3 of the comments to populate, but they only do so in the "Comment" field (So, I can only see one at a time). 

All of my attempts to populate all three comment fields with their respective comments have been unsuccessful.  

So, I know that all three types of comments are available, and they seem to be parsing out correctly; but I can't work out how to get them to populate the correct fields.

mathjazz:
The code looks good, I'm testing it locally. If I understand correctly only the "Comment" field gets stored in the DB (which you can confirm in Django admin)?

abowler2:
Back on track  - Thank you!

abowler2:
OK...so I have this functioning and displaying the comments in the UI.  However, it looks as though the file comments are only storing the last comment listed as opposed to all of them. I'm attaching an image of the file comments present in application.ftl. The top comment ("These strings are used inside the Application....) is not being stored and therefore not being rendered.

Screen Shot 2019-10-17 at 10 50 35 AM

abowler2:
Hey Matjaz ~ I was about to send in my updated PR but I have 2 tests failing with the following message:  
> ProgrammingError: column "group_comment" of relation "base_entity" already exists

mathjazz:
Could you try update the PR anyways? It could be that something went wrong with migrations locally.

abowler2:
In looking at the test that failed it looks as though it just needs the 2 new fields added to it.  Would you like me to do that?

Also, since I haven't been able to connect with you on IRC today, please let me know what you would like for me to work on next. ~ Thanks

mathjazz:
> In looking at the test that failed it looks as though it just needs the 2 new fields added to it. Would you like me to do that?

Yes, please!

> Also, since I haven't been able to connect with you on IRC today, please let me know what you would like for me to work on next. ~ Thanks

Sorry about that. I'll ping you on IRC!

pontoon/sync/vcs/models.py
```diff
@@ -623,6 +623,10 @@ def __init__(self, vcs_project, path, locales=None):
                 string=translation.source_string,
                 string_plural=translation.source_string_plural,
                 comments=translation.comments,
+                group_comments=translation.group_comments if hasattr(
+                    translation, 'group_comments') else None,
+                resource_comments=translation.resource_comments if hasattr(
+                    translation, 'resource_comments') else None,
```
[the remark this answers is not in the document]
adngdb: Matjaz, correct me if I'm wrong but there's no need for the condition here, as `group_comments` and `resource_comments` are defined in `VCSTranslation`. Thus, every instance of `translation` should have them, even is set to `None`.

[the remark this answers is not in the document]
mathjazz: I'm afraid it's needed afterall. :/

While for example `SilmeEntity` inherits from `VCSTranslation`, it doesn't initialize its attributes (with `super()`). To do that, we'd need to refactor pretty much every file format quite a bit, because we use the `@property` decorator, which doesn't play nicely with `super()`. Or maybe you have an idea how to do that more easily?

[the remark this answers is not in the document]
abowler2: Matjaz,  I did have these as one line originally but the linter failed as it was over the 100 character limit. Is there a preferred way for handling this?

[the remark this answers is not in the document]
mathjazz: Good point, in this case let's do this:
```python
group_comments=(
    translation.resource_comments
    if hasattr(translation, 'resource_comments') else None
),
```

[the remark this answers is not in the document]
adngdb: Nope, I don't know the code enough to propose anything else. Forget that I was there! :smile:

abowler2:
For the 'See more' button ~ I was not able to work out how to align it to the right of the truncated text without it overlapping.  Your thoughts and/or suggestions would be greatly appreciated.

frontend/src/modules/entitydetails/components/Metadata.css
```diff
@@ -49,3 +49,22 @@
...
+    border: none;
+    color: #7BC876;
+    float: right;
+    font-weight: 300;
+    padding: 0;
+}
+
+.hide {
+    display: none;
+}
```
mathjazz: Mind the newline at the end of the file.

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -84,6 +86,55 @@ export default class Metadata extends React.Component<Props> {
...
+                title='Resource Comment'
+                className={(this.state.seeMore) ? 'comment' : 'comment truncate'}
+            >
+                <Linkify properties={ { target: '_blank', rel: 'noopener noreferrer' } }>
+                    { comment }
+                </Linkify>
+                <button
+                    id='entitydetails-Metadata--see-more'
+                    className={(this.state.seeMore) ? 'hide' : ''}
+                    title="See More"
```
mathjazz: I wouldn't use a `title` attribute here, since it's the same as the button text. If you still want to keep it, please use single quotes.

pontoon/sync/vcs/models.py
```diff
@@ -688,13 +696,15 @@ class VCSEntity(object):
...
     def __init__(self, resource, key, string, comments, source, string_plural='',
-                 order=0):
+                 order=0, group_comments=None, resource_comments=None):
         self.resource = resource
         self.key = key
         self.string = string
         self.string_plural = string_plural
         self.translations = {}
         self.comments = comments
+        self.group_comments = group_comments
```
mathjazz: `or []`

pontoon/sync/changeset.py
```diff
@@ -181,6 +181,9 @@ def get_entity_updates(self, vcs_entity, db_entity=None):
             'string_plural': vcs_entity.string_plural,
             'key': vcs_entity.key,
             'comment': '\n'.join(vcs_entity.comments),
+            'group_comment': '\n'.join(
+                vcs_entity.group_comments) if vcs_entity.group_comments is not None else '',
```
mathjazz: Same as below: now that `vcs_entity.group_comments` is always a `list`, you can just use `'\n'.join(vcs_entity.group_comments)`.

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -84,6 +86,55 @@ export default class Metadata extends React.Component<Props> {
         </Localized>;
     }
 
+    renderGroupComment(entity: Entity): React.Node {
+        if (!entity.group_comment) {
+            return null;
+        }
+
+        let comment = entity.group_comment;
```
mathjazz: Since this is only used once, you could just use `entity.group_comment` directly. Also below for the Resource comment.

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -84,6 +86,55 @@ export default class Metadata extends React.Component<Props> {
...
+                    { comment }
+                </Linkify>
+                <button
+                    id='entitydetails-Metadata--see-more'
+                    className={(this.state.seeMore) ? 'hide' : ''}
+                    title="See More"
+                    onClick={ this.handleClick }
+                >
+                    { 'See More' }
+                </button>
```
mathjazz: I don't think we'll be able to make the `<button>` localizable that way, given that it lives within the `<Localized>` tag already. @adngdb?

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -84,6 +86,55 @@ export default class Metadata extends React.Component<Props> {
...
+                </Linkify>
+            </Property>
+        </Localized>;
+    }
+
+    state = { seeMore: false }
+
+    handleClick = () => {
+        this.setState({ seeMore: true });
+    };
```
mathjazz: We should use state here. See more and the truncated resource comment should be reset every time you move to the next entity.

frontend/public/static/locale/en-US/translate.ftl
```diff
@@ -190,6 +190,15 @@ entitydetails-Helpers--locales = Locales
 entitydetails-Metadata--comment =
     .title = Comment
 
+entitydetails-Metadata--group_commment =
```
adngdb: Please don't use underscores in l10n IDs but dashes instead. Here it should be `entitydetails-Metadata--group-comment` instead (note also that you put 3 "m" in "comment" :wink: ).

frontend/src/modules/entitydetails/components/Metadata.js
adngdb: Nope, the `<button>` element needs to be put inside another `<Localized>` component in order to make the content localizable. It also needs a new l10n id. Note that `<Localized>` elements are stackable, so that's not a problem.

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -84,6 +86,55 @@ export default class Metadata extends React.Component<Props> {
...
+            <Property
+                title='Resource Comment'
+                className={(this.state.seeMore) ? 'comment' : 'comment truncate'}
+            >
+                <Linkify properties={ { target: '_blank', rel: 'noopener noreferrer' } }>
+                    { comment }
+                </Linkify>
+                <button
+                    id='entitydetails-Metadata--see-more'
+                    className={(this.state.seeMore) ? 'hide' : ''}
```
adngdb: Please keep spaces after `{` and before `}`.

frontend/src/modules/entitydetails/components/Metadata.js
adngdb: I guess you meant "we should **not**"? State is actually fine, but it should be reset every time the entity changes. Also, since this state is specific to only the resource comment part, I think I would recommend to move that part into a new component. It can be a very simple component (maybe even using React Hooks? ).

mathjazz: Yes, I mean we should NOT. And I'm wrong, thanks for the correction and sorry for the confusion, April!

frontend/src/modules/entitydetails/components/Metadata.js
abowler2: @adngdb I have an id of `entitydetails-Metadata--see-more` on the button now. Does this need to be different?

abowler2:
I completed all the requested revisions but the ability to reset the truncated text with each entity change.   All the attempts I made broke the button, hid the Resource comment entirely, or crashed the whole App (yep, I had lots of fun :laughing: )
  If you could point me in the right direction for getting the state to reset without breaking the function of the button I would appreciate it.

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -84,6 +90,56 @@ export default class Metadata extends React.Component<Props> {
...
+    constructor(props: Props) {
+        super(props);
+        this.state = { seeMore: false };
+    }
+
+    handleClick = () => {
+        this.setState((state) => {
+            return { seeMore: !state.seeMore };
+        });
+    };
```
mathjazz: We could simply `this.setState({ seeMore: true });`, because the button is only visible when the text is truncated, i.e. when it can "see more" (and not "see less").

mathjazz: Also, let's move this method up, in front of `handleClickOnPlaceable`, to keep all `render*` methods together. I'd also call it `handleClickOnSeeMore`, to follow the pattern introduced by `handleClickOnPlaceable`.

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -84,6 +90,56 @@ export default class Metadata extends React.Component<Props> {
...
+
+    renderResourceComment(entity: Entity): React.Node {
+        if (!entity.resource_comment) {
+            return null;
+        }
+
+        return <Localized id='entitydetails-Metadata--resource-comment' attrs={ { title: true } }>
+            <Property
+                title='Resource Comment'
+                className={ (this.state.seeMore) ? 'comment' : 'comment truncate' }
```
mathjazz: Nit, you can remove brackets.

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -84,6 +90,56 @@ export default class Metadata extends React.Component<Props> {
...
+            >
+                <Linkify properties={ { target: '_blank', rel: 'noopener noreferrer' } }>
+                    { entity.resource_comment }
+                </Linkify>
+                <Localized id='entitydetails-Metadata--see-more'>
+                    <button
+                        className={ (this.state.seeMore) ? 'hide' : '' }
```
mathjazz: This is a jQueryish approach. In React world we don't even render the button if `seeMore === true`. Then you can also remove the `.hide` rule from CSS.

For the exact code pattern, see: https://github.com/mozilla/pontoon/blob/df0aafcf03336462cfd1419dd2c908e87d7f1a97/frontend/src/modules/entitydetails/components/Helpers.js#L76-L78.

abowler2:
I think all that is left is the styling of the button.  I might be able to achieve what we are looking for with Flexbox.

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -84,6 +90,59 @@ export default class Metadata extends React.Component<Props> {
...
+
+        return <Localized id='entitydetails-Metadata--resource-comment' attrs={ { title: true } }>
+            <Property
+                title='Resource Comment'
+                className={ this.state.seeMore ? 'comment' : 'comment truncate' }
+            >
+                <Linkify properties={ { target: '_blank', rel: 'noopener noreferrer' } }>
+                    { entity.resource_comment }
+                </Linkify>
+                { this.state.seeMore  ? null :
```
mathjazz: Nit: Mind the double space.

abowler2:
Darn it!  I knew I was going to forget to move those once I got it all working 😉 
It is corrected now.

I will play with the styling and let you know ASAP if I'm too far out of my depths on it.

abowler2:
Just wanted to give an update on this.  I have the CSS worked out for the See More button and am working on setting it up to check if there is overflow before displaying the button.

abowler2:
In trying to check if the text is truncated before rendering the button I'm having trouble accessing the widths of the element to do a check.  I pushed what I have so far for your review. ATM, I get a 
`span` is null error.

abowler2:
I was able to work out where I went wrong in checking for truncated text, and the button should work correctly now (Yay!).  However, it does not pass some of the tests since from what I can tell the `ref` is not included in the test's shallow copy so it equates to `null`

frontend/src/modules/entitydetails/components/Metadata.css
```diff
@@ -53,14 +53,19 @@
...
 .metadata button {
     background: none;
     border: none;
     color: #7BC876;
-    float: right;
+    font-style: italic;
     font-weight: 300;
     padding: 0;
+    position: absolute;
+    top: 3px;
```
mathjazz: I think this should be 2px. It would be even more noticeable if the `font-style` would `normal`.

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -25,19 +25,58 @@ type Props = {|
     +navigateToPath: (string) => void,
 |};
 
+type State = {|
+    seeMore: boolean,
+    isNotTruncated: boolean
```
mathjazz: Nit: Please add comma at the end of the line.

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -25,19 +25,58 @@ type Props = {|
...
  *  - a link to the resource
  *  - a link to the project
  */
-export default class Metadata extends React.Component<Props> {
+export default class Metadata extends React.Component<Props, State> {
+    constructor(props: Props) {
+        super(props);
+        this.state = {
+            seeMore: false,
+            isNotTruncated: false
```
mathjazz: Nit: Please add comma at the end of the line.

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -149,7 +231,7 @@ export default class Metadata extends React.Component<Props> {
     render(): React.Node {
         const { entity, locale, openLightbox, pluralForm } = this.props;
 
-        return <div className="metadata">
+        return <div className="metadata" ref={ this.span }>
```
mathjazz: Why do you reference this as a "span", if it's a div?

frontend/src/modules/entitydetails/components/Metadata.css
```diff
@@ -49,3 +49,23 @@
     margin: 0;
     padding: 0;
 }
+
+.truncate {
+    text-overflow: ellipsis;
+    overflow: hidden;
+    padding-right: 50px;
```
mathjazz: We could add a pixel or two more here, because sometimes ... gets really close to "See More". But we really need to make it dynamic, because the width will change with localizations. If that's too hard, we'll need to replace "See More" with "...".

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -25,19 +25,58 @@ type Props = {|
...
+            isNotTruncated: false
+        };
+        this.span = React.createRef();
+    }
+
+    componentDidMount() {
+        this.setState({
+            isNotTruncated: this.isEllispsisActivated()
+            });
+        }
```
mathjazz: This method has weird alignment. See `componentDidUpdate` for reference (including `setState`).

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -25,19 +25,58 @@ type Props = {|
...
+        if (this.props.entity !== prevProps.entity) {
+            this.setState({ seeMore: false });
+        }
+    }
+
+    isEllispsisActivated = () => {
+        const span = this.span.current;
+        const element = span.querySelector('span.resource')
+
+        return  element.getBoundingClientRect().width < element.scrollWidth
```
mathjazz: Nit: double space.

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -25,19 +25,58 @@ type Props = {|
...
+
+    componentDidUpdate(prevProps: Props) {
+        if (this.props.entity !== prevProps.entity) {
+            this.setState({ seeMore: false });
+        }
+    }
+
+    isEllispsisActivated = () => {
+        const span = this.span.current;
+        const element = span.querySelector('span.resource')
```
mathjazz: I understand why you did that, but I wonder if we can reference that span directly somehow. Maybe @adngdb has ideas? Maybe we should not use `Property` here?

mathjazz:
Oh, and the `flow` is failing.

frontend/src/modules/entitydetails/components/Metadata.js
abowler2: I referenced this as a "span" as the original placement I had for it was on a span, and I didn't catch it and rename when I moved it to the div.  It is changed now.

frontend/src/modules/entitydetails/components/Metadata.js
abowler2: This was the only way I was able to work out how to target this specific span as opposed to all the comments.  If @adngdb has another idea, just let me know.

adngdb: Can you not put the `ref` directly on the element you're interested in? Which one is it you're trying to reach?

mathjazz: Hm... It's not in the `Property` component actually, it's here:
https://github.com/mozilla/pontoon/pull/1431/files/8cf204ec53a0de923656981f2a08a1cacfa0c9fe#diff-fff36ad131370bbb16e4d6f466368b64R154

You should be able to put it there directly?

abowler2: I'm trying to reach the resource comment, and when I placed the ref on any of the elements in `renderResourceComment` it would return null.

adngdb: Can you pass the `ref` to `Property` and inside `Property` assign it to the right element?

If you haven't done so already, I recommend reading the doc about Refs: https://reactjs.org/docs/refs-and-the-dom.html

frontend/src/modules/entitydetails/components/Metadata.css
```diff
@@ -49,3 +49,23 @@
     margin: 0;
     padding: 0;
 }
+
+.truncate {
+    text-overflow: ellipsis;
+    overflow: hidden;
+    padding-right: 9%;
```
mathjazz: That's still not OK, see https://github.com/mozilla/pontoon/pull/1431#discussion_r339634630.

@adngdb How should we circumvent the problem of dynamic "See More" button width (due to localizations)? Should we dynamically calulcated it (and set padding in JS) or should we just use something like "..." and hardcode the width? I'm leaning towards latter.

frontend/src/modules/entitydetails/components/Metadata.js
abowler2: @adngdb I have been using the docs for Refs as well as referencing other uses of refs within the code base. 

Here is what I have found so far: 
If I pass the `ref` directly to `Property` I get the following error: 

> "Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?"

and the button does not render as the `ref` would be null.

When looking into forwarding the `ref` it appears that this is recommended for reusable component libraries and this is the only time I'm aware of that this element will need to be referenced. At any rate, my initial attempts at this were not successful, but I'm happy to keep looking into it if this is the direction needed.

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -25,19 +25,62 @@ type Props = {|
...
+                isNotTruncated: this.isEllispsisActivated(),
+             });
+        }
+    }
+
+    isEllispsisActivated = () => {
+        if(this.div.current) {
+            const div = this.div.current;
+            const element = div.querySelector('span.resource')
+            return  element.getBoundingClientRect().width < element.scrollWidth
```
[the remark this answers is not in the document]
abowler2: I don't see the double space...

frontend/src/modules/entitydetails/components/Metadata.css
abowler2: If we change the button text to "..." will it be clear that it is a button to expand even with it colored green?

frontend/src/modules/entitydetails/components/Metadata.js
adngdb: OK, I finally understood and reproduced the problem. I think you are overdoing it here, just using a fixed max length should be fine. I for one don't care too much if it doesn't look pretty (wink wink Matjaz) as long as it works. I'm also not sure why you want to hide that comment by default, and not just show all of it?

mathjazz: The reason we want to hide it by default is that it's usually long and repeats for each string within the file. It was an explicit request, which we took without knowing it would cause so much trouble. :)

mathjazz: @abowler2 As agreed with @adngdb, we should show the first 90 characters of the Resource comment and then add "...", followed by the "See more" button. Based on our screen resolution data, that will fit the Resource comment into one line for >95% of our users.

abowler2: I knew I ran into issues when trying to use the length as the condition for expanding the text before, but it was one of the first things I tried and I've been down many paths since, so I couldn't remember exactly what the issue was. 

I've been able to recreate it and get a little bit more insight (PR has been pushed for your review):  
In the ternary the `comment` can be seen for determining the length and for splicing and rendering the first 85 chars (90 was too wide with the button), BUT it won't render the full comment after the button is clicked. The button hides, but the comment does not change.

However, if the `comment` is less than 85 chars then it displays just fine (without the ellipsis and no button), and if I change the `comment` in the `true` value to a plain string (ie "Hello") the comment changes to 'Hello' when the button is clicked.

I'm not sure why this is happening or even how to research this issue... 🆘

abowler2:
You can disregard my comment from earlier (I deleted it too), I figured out what went wrong and pushed the update.

frontend/src/modules/entitydetails/components/Metadata.css
```diff
@@ -49,3 +49,19 @@
...
+.metadata button {
+    background: none;
+    border: none;
+    color: #7BC876;
+    font-style: italic;
+    font-weight: 300;
+    padding: 0;
+    position: absolute;
+    top: 2px;
+    right: 4px;
```
mathjazz: The button should just follow the text instead of being positioned absolute at the end of the line. So you should remove the last three properties and set `padding: 0 0 0 2px;`.

I also think `font-weight: 300;` doesn't have any impact?

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -84,6 +104,45 @@ export default class Metadata extends React.Component<Props> {
...
+        }
+
+        let comment = entity.resource_comment
+
+        return <Localized id='entitydetails-Metadata--resource-comment' attrs={ { title: true } }>
+            <Property title='Resource Comment' className='comment truncate'>
+                <Linkify properties={ { target: '_blank', rel: 'noopener noreferrer' } } className='resource'>
+                    { comment.length < 85 || seeMore ? comment : comment.slice(0, 85) + "..." }
+                </Linkify>
+                { comment.length < 85 || seeMore ? null :
```
mathjazz: 85 repeats several times, you should define `const MAX_LENGTH = 85;` at the top of the method.

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -84,6 +104,45 @@ export default class Metadata extends React.Component<Props> {
...
+        if (!entity.resource_comment) {
+            return null;
+        }
+
+        let comment = entity.resource_comment
+
+        return <Localized id='entitydetails-Metadata--resource-comment' attrs={ { title: true } }>
+            <Property title='Resource Comment' className='comment truncate'>
+                <Linkify properties={ { target: '_blank', rel: 'noopener noreferrer' } } className='resource'>
+                    { comment.length < 85 || seeMore ? comment : comment.slice(0, 85) + "..." }
```
mathjazz: Instead of "...", we should use the https://en.wikipedia.org/wiki/Ellipsis (…) -> it's a single character.

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -84,6 +104,45 @@ export default class Metadata extends React.Component<Props> {
...
+
+        if (!entity.resource_comment) {
+            return null;
+        }
+
+        let comment = entity.resource_comment
+
+        return <Localized id='entitydetails-Metadata--resource-comment' attrs={ { title: true } }>
+            <Property title='Resource Comment' className='comment truncate'>
+                <Linkify properties={ { target: '_blank', rel: 'noopener noreferrer' } } className='resource'>
```
mathjazz: We should break this into multiple lines:
```js
<Linkify
    properties=...
    className=...
>
```

frontend/src/modules/entitydetails/components/Metadata.css
abowler2: With this no longer set to absolute there is also no need for the `truncate` class with position of relative, so I removed it.

frontend/src/modules/entitydetails/components/Metadata.js
mathjazz: Excellent catch!

We usually try to keep line length under <100 charaters, but in this case I don't think code readability would be much improved if we imposed that limit in the line withing the `<Linkify>` element.

frontend/src/modules/entitydetails/components/Metadata.css
mathjazz: Good call!

abowler2:
No problem. It didn't help that I chose to go in possibly the most complicated direction with it.  

I appreciate both you and @adngdb for patiently guiding me out of the weeds 😄

PR #1432, opened 2019-10-16
No bug: Set width/height on user avatars

mathjazz:
Setting width/height on user avatars prevents the surrounding UI from jumping while the image is loading.

Relatedly, we're also making the alt= attribute empty to prevent (rather meaningless) text from appearing during loading. Gmail, FB, TW, BZ all seem to pull the same trick.

@Henikilan Could you please review this?


Henikilan:
Looks good:)
Should we also apply the same changes to non-translate view pages?

PR #1434, opened 2019-10-16
Fix bug 1344120 - Reset search placeholder icon

sccofield:
Reset search placeholder icon when the filter selection is not applied

[1344120](https://bugzilla.mozilla.org/show_bug.cgi?id=1344120)

### DIFF
--- a/frontend/src/modules/search/components/FiltersPanel.js
+++ b/frontend/src/modules/search/components/FiltersPanel.js
@@ -36,6 +36,7 @@ type Props = {|
     toggleFilter: (string, string) => void,
     update: () => void,
     updateTimeRange: (filter: string) => void,
+    updateFiltersFromURLParams: () => void,
 |};
 
 type State = {|
@@ -78,6 +79,7 @@ export class FiltersPanelBase extends React.Component<Props, State> {
         this.setState(state => {
             return { visible: !state.visible };
         });
+        this.props.updateFiltersFromURLParams();
     }
 
     // This method is called by the Higher-Order Component `onClickOutside`
@@ -86,10 +88,13 @@ export class FiltersPanelBase extends React.Component<Props, State> {
         this.setState({
             visible: false,
         });
+        this.props.updateFiltersFromURLParams();
     }
 
     applyFilters = () => {
-        this.toggleVisibility();
+        this.setState({
+            visible: false,
+        });
         return this.props.update();
     }
 
--- a/frontend/src/modules/search/components/SearchBox.js
+++ b/frontend/src/modules/search/components/SearchBox.js
@@ -65,6 +65,21 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
     constructor(props: InternalProps) {
         super(props);
 
+        this.state = {
+            search : '',
+            statuses: {},
+            extras: {},
+            tags: {},
+            timeRange: null,
+            authors: {},
+        };
+        
+        this.searchInput = React.createRef();
+    }
+
+    updateFiltersFromURLParams = () => {
+        const { props } = this;
+        
         const statuses = this.getInitialStatuses();
         if (props.parameters.status) {
             props.parameters.status.split(',').forEach(f => {
@@ -100,21 +115,20 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
 
         const searchParam = props.parameters.search;
 
-        this.state = {
+        this.setState({
             search: searchParam ? searchParam.toString() : '',
             statuses,
             extras,
             tags,
             timeRange,
             authors,
-        };
-
-        this.searchInput = React.createRef();
+        });
     }
 
     componentDidMount() {
         // $FLOW_IGNORE (errors that I don't understand, no help from the Web)
         document.addEventListener('keydown', this.handleShortcuts);
+        this.updateFiltersFromURLParams();
     }
 
     componentDidUpdate(prevProps: InternalProps) {
@@ -443,6 +457,7 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
                 toggleFilter={ this.toggleFilter }
                 update={ this.update }
                 updateTimeRange={ this.updateTimeRange }
+                updateFiltersFromURLParams={ this.updateFiltersFromURLParams }
             />
         </div>;
     }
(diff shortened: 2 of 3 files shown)


mathjazz:
Thanks for the patch @sccofield! Could you please configure your text editor first so that it doesn't change indent and whitespace when not needed?

Once that's done, please update the PR and I'll have a look. Right now it's [very hard to see](https://github.com/mozilla/pontoon/pull/1434/files) where the actual changes are.

sccofield:
hi @mathjazz I have updated the PR.

frontend/src/modules/search/components/FiltersPanel.js
```diff
@@ -86,6 +86,7 @@ export class FiltersPanelBase extends React.Component<Props, State> {
         this.setState({
             visible: false,
         });
+        this.props.resetFilters();
```
mathjazz: Nice, this fixes the bug for the use case described in the bug:
https://bugzilla.mozilla.org/show_bug.cgi?id=1344120#c9

However, it introduces another problem: If you apply some filters first (e.g. Missing) and then try to reproduce the bug, the filters will be reset to the default state (nothing selected) instead of the previous state.

See how to solve the bug properly:
https://bugzilla.mozilla.org/show_bug.cgi?id=1344120#c13

sccofield:
hi @mathjazz I have worked on the requested changes. I had to call the method on componentDidMount because I was setting a state on the method, and I needed to be sure the state was available before calling the setState method.

frontend/src/modules/search/components/SearchBox.js
```diff
@@ -443,6 +456,7 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
                 toggleFilter={ this.toggleFilter }
                 update={ this.update }
                 updateTimeRange={ this.updateTimeRange }
+                initializeFilters = { this.initializeFilters }
```
mathjazz: Mind the whitespace around `=`.

frontend/src/modules/search/components/SearchBox.js
```diff
@@ -65,73 +65,86 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
...
             });
         }
 
-        const searchParam = props.parameters.search;
+        const searchParam = this.props.parameters.search;
 
-        this.state = {
-            search: searchParam ? searchParam.toString() : '',
+            this.setState({
+            search : searchParam ? searchParam.toString() : '',
```
mathjazz: Nit: Mind the whitespace.

frontend/src/modules/search/components/SearchBox.js
```diff
@@ -65,73 +65,86 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
...
                 authors[f] = true;
             });
         }
 
-        const searchParam = props.parameters.search;
+        const searchParam = this.props.parameters.search;
 
-        this.state = {
-            search: searchParam ? searchParam.toString() : '',
+            this.setState({
```
mathjazz: Nit: Mind the indent.

frontend/src/modules/search/components/SearchBox.js
```diff
@@ -65,73 +65,86 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
...
+        }
+    }
+
+    componentWillUnmount() {
+        // $FLOW_IGNORE (errors that I don't understand, no help from the Web)
+        document.removeEventListener('keydown', this.handleShortcuts);
+    }
+
+    initializeFilters = () => {
+        // Reset filters to state provided by the url parameters
```
mathjazz: You should define a `props` constant at the beginning of the function and then you can keep the diff much shorter by keeping `props...` instead of `this.props...`.

frontend/src/modules/search/components/SearchBox.js
```diff
@@ -65,73 +65,86 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
...
+            });
+        }
+    }
+
+    componentWillUnmount() {
+        // $FLOW_IGNORE (errors that I don't understand, no help from the Web)
+        document.removeEventListener('keydown', this.handleShortcuts);
+    }
+
+    initializeFilters = () => {
```
mathjazz: Let's call the function `updateFiltersFromURLParams` and then you don't need the comment below.

sccofield:
> Nicely done!
> 
> It works fine, except for one edge case - if you close the menu by clicking on the filter icon, the bug still exists.

Hi @mathjazz to solve the edge case, I had to call `updateFiltersFromURLParams` from the `toggleVisibility` method on the `filterpanel` component. 

After doing that I noticed another bug. 

When I click on `apply filter` the placeholder does not update until I click outside the filter panel. 

I don't really know what caused the behaviour so I removed the `togglevisibility` method that was called on the `applyfilter` method, and did a setState for the visibility right in the method, and it worked perfectly.

I don't know if that was the right approach to go about it.

frontend/src/modules/search/components/SearchBox.js
```diff
@@ -100,21 +114,20 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
...
             search: searchParam ? searchParam.toString() : '',
             statuses,
             extras,
             tags,
             timeRange,
             authors,
-        };
-
-        this.searchInput = React.createRef();
+        })
```
adngdb: This is missing a semicolon here.

adngdb:
Sorry, I missed that the tests do not pass. Please make sure they pass locally before you send commits. To run frontend tests, you can use `make test-frontend`. We use [Jest](https://jestjs.io/), see the documentation to learn more about it.

frontend/src/modules/search/components/SearchBox.js
```diff
@@ -65,7 +65,21 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
...
+            timeRange: null,
+            authors: {},
+        };
+        
+        this.searchInput = React.createRef();
+    }
+
+    updateFiltersFromURLParams = () => {
         const statuses = this.getInitialStatuses();
+        const { props } = this
```
mathjazz: Please add ; at the end of the line and we're good! :)

mathjazz: AND please move this line to the top of the function and add a blank line after it.

sccofield:
hi @mathjazz all the requested changes have been effected.

sccofield:
yeaaa. Thanks @mathjazz for the feedback, I will work on my commits in future PRs.

PR #1438, opened 2019-10-18
Fix bug 1589506 - Set width/height on img elements with user avatars

monikamaheshwari:
Set width/height on img elements with user avatars


mathjazz:
Well done!

I'd like to ask you to fix one more thing as part of this PR - avatars on high DPI displays (e.g. [Retina](https://en.wikipedia.org/wiki/Retina_display)) don't look sharp. To mitigate that, we should double the image size, while keeping the height/width attribute on the `` tag unchanged. 

The same trick is used on GitHub for example - if you go to https://github.com/monikamaheshwari and download your avatar, you'll see that the image size is almost twice as big as the actual displayed image.

So please double the size of images used in `src` attribute of all images you have changed height/width for in this PR. And also for the ones fixed here:
https://github.com/mozilla/pontoon/commit/01fd682246c2e63ab058bbbc632ea9b34dc70e48

mathjazz:
Good job!

I found two more places where you need to replace `44` with `88`:
* https://github.com/mozilla/pontoon/blob/e674f075eacdd5bc09ea38910ea0711d3ac9b122/pontoon/base/templates/widgets/latest_activity.html#L8
* https://github.com/mozilla/pontoon/blob/e674f075eacdd5bc09ea38910ea0711d3ac9b122/pontoon/base/models.py#L106

And please replace both `anon` images in https://github.com/mozilla/pontoon/tree/e674f075eacdd5bc09ea38910ea0711d3ac9b122/pontoon/base/static/img with the following:

anon_big

anon

PR #1441, opened 2019-10-18
[WIP] Fix Bug 1588931 - Localize Machinery sources

nadinda:
This fix makes machinery sources and their title attributes in machinery tab on translation menu localizable with the details as follow:
- The only localizable sources.type is "Translation Memory"
- All sources.title are localizable

The output is like the screenshot down below:
machinery localizable


adngdb:
Hey @nadinda, I like what you did there, it's smart! I'd like to ask you to improve it a bit though. My main concern is having `<Localized>` components in the `api` module. I'd like to keep that module as non-React as possible. 

My suggestion is to add a layer of abstraction between the api and the Machinery component. We could have a React component per type of translation source (Machine translation, Translation Memory, Microsoft, etc. ), each defining its render method with localized content and all. Then the Machinery component, instead of generating a structure for each element, would simply instance the right component based on the translation type. Then the api can return the bare minimum of data: the source type, the content and whatever metadata is needed.

Maybe this is a bit cryptic, so feel free to ask questions if you need help with it.

nadinda:
> Hey @nadinda, I like what you did there, it's smart! I'd like to ask you to improve it a bit though. My main concern is having `<Localized>` components in the `api` module. I'd like to keep that module as non-React as possible.
> 
> My suggestion is to add a layer of abstraction between the api and the Machinery component. We could have a React component per type of translation source (Machine translation, Translation Memory, Microsoft, etc. ), each defining its render method with localized content and all. Then the Machinery component, instead of generating a structure for each element, would simply instance the right component based on the translation type. Then the api can return the bare minimum of data: the source type, the content and whatever metadata is needed.
> 
> Maybe this is a bit cryptic, so feel free to ask questions if you need help with it.

Hi @adngdb, thanks for the feedback! You have the right concern for making the `api` module to not have any React component. I'm not sure if I understood your suggestion tho but I thought it would be a lot of changes to be done if we have to break each of the type of translation source to different React components. So I just made the adjustment similar to the `source.title` to get rid of the `<Localized>` component entirely in `api` module. 

Let me know if my solution can be improved. Thanks!

mathjazz:
Please check the failing tests.

nadinda:
> Please check the failing tests.

Yes, sorry I just noticed it. Still working on it.

nadinda:
Hi @adngdb mind to review this fix again? does my solution need more work or it's sufficient enough? Thanks before!

adngdb:
Hey @nadinda, that's a better version of the code for sure, but still not what I had in mind. Let me try to explain with more details.

I'd like you to create 6 new React components, one for each type of machinery source. They would be in the `modules/machinery` module, and be called something like `GoogleTranslation`, `MemoryTranslation` or `TransvisionMemoryTranslation`. 

Each component renders a translation source entry, as done in the `machinery.Translation` component at the moment. But because they are specific to each source, you can put the localizable content directly in there, and remove it from the api.

By doing that, you then can make the api functions return only the data that comes from the requests (`original`, `translation`, `quality`… ), as well as a type (a string like `google-translate`, `translation-memory`, etc. identifying the translation source). 

Then in the `machinery.Translation` component, you would replace the content of the translation sources with an instance of the correct translation source component (e.g. `<GoogleTranslation source={ source } />`).

Advantages of this:
- less data returned by the machinery api;
- localizable content is in a component that is self-contained and well described;
- better separation of concerns.

nadinda:
> Hey @nadinda, that's a better version of the code for sure, but still not what I had in mind. Let me try to explain with more details.
> 
> I'd like you to create 6 new React components, one for each type of machinery source. They would be in the `modules/machinery` module, and be called something like `GoogleTranslation`, `MemoryTranslation` or `TransvisionMemoryTranslation`.
> 
> Each component renders a translation source entry, as done in the `machinery.Translation` component at the moment. But because they are specific to each source, you can put the localizable content directly in there, and remove it from the api.
> 
> By doing that, you then can make the api functions return only the data that comes from the requests (`original`, `translation`, `quality`… ), as well as a type (a string like `google-translate`, `translation-memory`, etc. identifying the translation source).
> 
> Then in the `machinery.Translation` component, you would replace the content of the translation sources with an instance of the correct translation source component (e.g. `<GoogleTranslation source={ source } />`).
> 
> Advantages of this:
> 
>     * less data returned by the machinery api;
> 
>     * localizable content is in a component that is self-contained and well described;
> 
>     * better separation of concerns.

Alright thanks for the more detailed explanation. I'll have a go.

nadinda:
Hi @adngdb, I haven't written the tests for the new components. But can you review my current implementation to make sure whether we are on the same page or not?

frontend/src/core/api/machinery.js
```diff
@@ -42,9 +42,8 @@ export default class MachineryAPI extends APIBase {
         return results.map(item => {
             return {
                 sources: [{
-                    type: 'Translation memory',
+                    type: 'translation-memory',
                     url: '/',
```
adngdb: You can move the `url`s too.

frontend/src/modules/machinery/components/CaighdeanTranslation.js
```diff
@@ -0,0 +1,43 @@
...
+        url: string,
+    },
+    index: number
+|};
+
+
+/**
+ * Show the translation source from Caighdean Machine Translation.
+ */
+export default class CaighdeanTranslation extends React.Component<Props> {
```
adngdb: These components are very simple, they should be functions instead of classes. Look at https://github.com/mozilla/pontoon/blob/master/frontend/src/modules/fluentoriginal/components/SimpleString.js for an example.

frontend/src/modules/machinery/components/CaighdeanTranslation.js
```diff
@@ -0,0 +1,43 @@
...
+    render() {
+        const {source, index} = this.props;
+
+        return <ul className="sources">
```
adngdb: These should return an `<li>` element, the `<ul>` should be set outside, in the `Translation` component.

frontend/src/core/api/machinery.js
nadinda: only the `url` which have fixed string or all of them? Cause I thought about including the url in the react component too. But there are some like for example `MicrosoftTerminology` which has `item.source` and `locale.msTerminologyCode` as its `url`

frontend/src/modules/machinery/components/CaighdeanTranslation.js
nadinda: Alright, will do! I thought about making them as function and was not sure.

frontend/src/modules/machinery/components/CaighdeanTranslation.js
nadinda: Okay!

frontend/src/modules/machinery/components/CaighdeanTranslation.test.js
```diff
@@ -0,0 +1,27 @@
...
+
+
+describe('<CaighdeanTranslation>', () => {
+    it('renders the component properly', () => {
+        const source = {
+            type: 'caighdean',
+            url: 'https://github.com/kscanne/caighdean',
+            id: 'machinery-CaighdeanTranslation--visit-caighdean',
+            string: 'Caighdean',
+            title: 'Visit Caighdean Machine Translation',
```
adngdb: Only the `type` value is needed here. The rest is not used in the component (except `url` but I think it should be removed).

frontend/public/static/locale/en-US/translate.ftl
```diff
@@ -356,15 +356,56 @@ machinery-Machinery--search-placeholder =
...
+## Shows the translation source from Microsoft Translation.
+machinery-MicrosoftTranslation--visit-bing =
+    .title = Visit Bing Translate
+
+
+## Machinery Microsoft Terminology
+## Shows the translation source from Microsoft Terminology.
+machinery-MicrosoftTerminology--visit-microsoft =
+    .title = Visit Microsoft Terminology Service API.
+    © 2018 Microsoft Corporation. All rights reserved.
```
adngdb: I think this needs to be indented to align with the content on the previous line. Is that right @Pike?

frontend/src/core/api/machinery.js
adngdb: I think all of them. For the ones that need some data, you can pass the data down to the component. That makes for a more consistent approach.

frontend/src/modules/machinery/components/CaighdeanTranslation.js
```diff
@@ -0,0 +1,38 @@
+/* @flow */
+
+import React from 'react';
+import { Localized } from '@fluent/react';
+
+
+type Props = {|
+    source: {
+        type: string,
```
adngdb: You don't even need to pass `type`, since you know that it has to be, in this case, a `caidhdean` source. I think you should make `props` flatter, and pass only the data the components needs directly. In this specific case, the component doesn't actually need any props! :-)

frontend/src/modules/machinery/components/CaighdeanTranslation.test.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import CaighdeanTranslation from './CaighdeanTranslation';
+
+
+describe('<CaighdeanTranslation>', () => {
+    it('renders the component properly', () => {
+        const source = {
```
adngdb: You could move this to a constant outside of the `describe` block, called `SOURCE`, and use it just like you did here. Note that you'd not pass anything to the component, since it doesn't actually require any props.

frontend/src/modules/machinery/components/CaighdeanTranslation.js
adngdb: Note that these comments of course apply to all other components as well.

frontend/src/modules/machinery/components/MicrosoftTerminology.js
```diff
@@ -0,0 +1,39 @@
...
+    return <li>
+        <Localized
+            id= "machinery-MicrosoftTerminology--visit-microsoft"
+            attrs={{ title: true }}
+        >
+            <a
+                className="translation-source"
+                href={ source.url }
+                title="'Visit Microsoft Terminology Service API.\n'+
+            '© 2018 Microsoft Corporation. All rights reserved.'"
```
adngdb: Wrong indentation here.

adngdb: And weird quotes. Is that syntax even valid??? The double quotes `"` shouldn't be needed.

frontend/src/modules/machinery/components/Translation.js
```diff
@@ -115,6 +96,6 @@ export default class Translation extends React.Component<Props> {
                     />
                 </p>
             </li>
-        </Localized>;
+        </Localized>
```
adngdb: Please do keep the semicolon here.

frontend/src/modules/machinery/components/TranslationMemory.js
```diff
@@ -0,0 +1,51 @@
...
+                        attrs={{ title: true }}
+                    >
+                        <sup title="Number of translation occurrences">
+                            { source.count }
+                        </sup>
+                    </Localized>
+                }
+            </a>
+        </Localized>
+    </li>
```
adngdb: This is missing a semicolon here.

frontend/src/modules/machinery/components/TranslationMemory.test.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import TranslationMemory from './TranslationMemory';
+
+
+describe('<TranslationMemory>', () => {
+
+    it('renders the component without number of occurance properly', () => {
```
adngdb: "occurance" -> "occurrences"

frontend/src/modules/machinery/components/TranslationMemory.test.js
```diff
@@ -0,0 +1,48 @@
...
+        expect(wrapper.find('Localized').at(0).props().id).toEqual('machinery-TranslationMemory--pontoon-homepage');
+        expect(wrapper.find('li a').props().title).toContain('Pontoon Homepage');
+
+        expect(wrapper.find('Localized').at(1).props().id).toEqual('machinery-TranslationMemory--translation-memory');
+        expect(wrapper.find('li a span').text()).toEqual('Translation memory');
+
+        expect(wrapper.find('Localized')).toHaveLength(2);
+    });
+
+    it('renders the component with number of occurance properly', () => {
```
adngdb: Same here.

frontend/src/modules/machinery/components/TranslationSource.js
```diff
@@ -0,0 +1,47 @@
...
+                    case 'transvision':
+                        return <TransvisionMemory source={ source } key={ index }/>;
+                    case 'caighdean':
+                        return <CaighdeanTranslation source={ source } key={ index }/>;
+                    default:
+                        return null;
+                }
+            });
+
+        return <div>{ translationSource }</div>
```
adngdb: Missing semicolon. And you can use `<></>` (which is the equivalent of `<React.Fragment></React.Fragment>` here to contain the components. Or better yet, you could create the `<ul>` element here instead of in `Translation`.

In any case, I think it's incorrect to have a `<div>` as the child of a `<ul>` element.

frontend/src/modules/machinery/components/TranslationSource.js
```diff
@@ -0,0 +1,47 @@
...
+
+type Props = {|
+    translation: MachineryTranslation,
+|};
+
+
+/**
+ * Shows a list of translation sources.
+ */
+export default function TranslationSource(props: Props) {
```
adngdb: This is great!

frontend/src/modules/machinery/components/TranslationSource.js
```diff
@@ -0,0 +1,47 @@
...
+        const {translation} = props;
```
adngdb: Please add spaces between brackets for consistency with the rest of the code base.

frontend/src/modules/machinery/components/TranslationSource.test.js
```diff
@@ -0,0 +1,57 @@
...
+
+
+describe('<TranslationSource>', () => {
+    [{id: 'translation-memory', component: 'TranslationMemory'},
+        {id: 'google-translate', component: 'GoogleTranslation'},
+        {id: 'microsoft-translator', component: 'MicrosoftTranslation'},
+        {id: 'microsoft-terminology', component: 'MicrosoftTerminology'},
+        {id: 'transvision', component: 'TransvisionMemory'},
+        {id: 'caighdean', component: 'CaighdeanTranslation'},
+    ].forEach(({id, component}) =>
```
adngdb: We use a tool to do just that: `jest-each`. See a usage example here: https://github.com/mozilla/pontoon/blob/master/frontend/src/core/placeable/parsers/camelCaseString.test.js

frontend/public/static/locale/en-US/translate.ftl
Pike: It just needs to be indented. I'd recommend making it a proper multi-line string, though:

```
## Shows the translation source from Microsoft Terminology.
machinery-MicrosoftTerminology--visit-microsoft =
    .title =
        Visit Microsoft Terminology Service API.
        © 2018 Microsoft Corporation. All rights reserved.
```

frontend/src/core/api/machinery.js
```diff
@@ -121,12 +118,12 @@ export default class MachineryAPI extends APIBase {
...
         }
 
-        return results.translations.map(item => {
+        return results.translations.map((item): MachineryTranslation => {
             return {
                 sources: [{
                     type: 'microsoft-terminology',
-                    url: 'https://www.microsoft.com/Language/en-US/Search.aspx?sString=' +
-                        item.source + '&langID=' + locale.msTerminologyCode,
+                    sourceString: item.source,
```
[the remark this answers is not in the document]
nadinda: ah yes! how could I miss it. Sorry.

nadinda:
phew, this changes took longer than I thought. Anyway, it's ready for another check, thanks! @adngdb

frontend/src/core/api/types.js
```diff
@@ -56,24 +56,8 @@ export type OtherLocaleTranslation = {|
...
-        sourceString: string,
-        localeCode: string,
-    }|{
-    type: 'transvision',
-        sourceString: string,
-        localeCode: string,
-    }|{
-        type: 'caighdean'
-    }>,
+    sources: Array<{ type: 'translation-memory' | 'google-translate' | 'microsoft-translator' | 'microsoft-terminology' | 'transvision' | 'caighdean' }>,
```
adngdb: You can make this a bit simpler but making it an array of strings instead of an array of objects.

`sources: Array<'translation-memory' | 'google-translate' | 'microsoft-translator' | 'microsoft-terminology' | 'transvision' | 'caighdean'>,`

What would be even nicer would be to define a new type for the `sourceType`, and use that here instead of the long list of types.

frontend/src/modules/machinery/components/TranslationSource.js
```diff
@@ -14,53 +15,52 @@ import TranslationMemory from './TranslationMemory';
...
 
 
 /**
  * Shows a list of translation sources.
  */
-export default function TranslationSource(props: Props) {
-        const { translation } = props;
+export default function TranslationSource({ translation, locale }: Props) {
+    const translationSource =
+        translation.sources.map((source, index) => {
```
adngdb: Please move this to the previous line (after the `=` sign), and then dedent the whole `switch` block by one level.

frontend/src/core/api/types.js
nadinda: I tried to make it as an array of strings, but sources.type were needed somewhere else. I'll have a look at it and try to change it again.

nadinda: Oh it works with your suggestion `sources: Array<'translation-memory' | 'google-translate' | 'microsoft-translator' | 'microsoft-terminology' | 'transvision' | 'caighdean'>,`
Maybe I just mistyped something before.

frontend/src/modules/machinery/components/Translation.js
```diff
@@ -46,7 +46,7 @@ export default class Translation extends React.Component<Props> {
     render() {
         const { locale, sourceString, translation, isReadOnlyEditor } = this.props;
 
-        const types = translation.sources.map(source => source.type);
+        const types = translation.sources.map(source => source);
```
adngdb: This is equivalent to `const types = translation.sources`. No need to use a `map` here.

frontend/src/core/api/types.js
```diff
@@ -55,8 +55,10 @@ export type OtherLocaleTranslation = {|
  * Translation that comes from a machine (Machine Translation,
  * Translation Memory... ).
  */
+type SourceType = 'translation-memory' | 'google-translate' | 'microsoft-translator' | 'microsoft-terminology' | 'transvision' | 'caighdean'
```
adngdb: Can you split this into several lines, to avoid being longer than 100 characters? Here's an example of how we do it elsewhere: https://github.com/mozilla/pontoon/blob/master/frontend/src/modules/batchactions/reducer.js#L25-L33

frontend/src/modules/machinery/components/Translation.js
nadinda: Yeah I also wondered about this bit. Okay, will change it!

frontend/src/core/api/types.js
nadinda: Ooo definitely looks neater. Thanks for pointing it out!

nadinda:
> This is great! Thanks a lot!
> 
> I did push a last minute commit to correct some mistakes that I missed in the translation file, nothing big (alphabetical sorting, and some content wasn't localized).
> 
> 

yayyy, thanks for bearing with me!

PR #1442, opened 2019-10-19
[WIP] Fix bug 1558484 - Add ability to log users in using GitHub

abowler2:
Add GitHub login thus creating 3 authentication systems:
   * Firefox Accounts
   * GitHub
   * Django (login only, user creation only possible via Django admin) - set this as default.

This fix sets 'django' as the default with the ability to change it to Firefox ('fxa') or GitHub ('github') if needed.

The GitHub authentication is not currently working and gives the following error:  
> SocialApp matching query does not exist.

I believe this is due to needing to have GitHub credentials added in order to gain access to their authentication process.

### DIFF
--- a/docs/admin/deployment.rst
+++ b/docs/admin/deployment.rst
@@ -78,10 +78,12 @@ you create:
    production.
    Adds some additional django apps that can be helpful during day to day development.
 
-``DJANGO_LOGIN``
+``AUTHENTICATION_METHOD`
    Required for non-Mozilla deployments, which cannot use `Firefox Accounts`_ for
-   login. Set to True if you want to use the default Django login instead. This will
+   login. Set to 'django' if you want to use the default Django login instead. This will
    allow you to log in via accounts created using `manage.py shell`.
+   Set to 'fxa' if you want to use 'Firefox Accounts'.
+   Set to 'github' if you want to use 'GitHub Account'.
 
 ``ENABLE_BUGS_TAB``
    Optional. Enables Bugs tab on team pages, which pulls team data from
--- a/docs/dev/setup-virtualenv.rst
+++ b/docs/dev/setup-virtualenv.rst
@@ -182,8 +182,10 @@ The following extra settings can be added to your ``.env`` file.
    Set your `Google Analytics key`_ to use Google Analytics.
 ``MANUAL_SYNC``
    Enable Sync button in project Admin.
-``DJANGO_LOGIN``
-   Set to True if you want to use the default Django login instead of Firefox Accounts. This will allow you to log in via accounts created using `manage.py shell`.
+``AUTHENTICATION_METHOD``
+   Set to 'django' if you want to use the default Django login instead of Firefox Accounts. This will allow you to log in via accounts created using `manage.py shell`.
+   Set to 'fxa' if you want to use the Firefox Accounts.
+   Set to 'github' if you want to use GitHub Account.
 
 .. _Microsoft Translator API key: http://msdn.microsoft.com/en-us/library/hh454950
 .. _Google Analytics key: https://www.google.com/analytics/
--- a/pontoon/allauth_urls.py
+++ b/pontoon/allauth_urls.py
@@ -14,7 +14,7 @@ from django.contrib.auth.views import login, logout
 from allauth.account import views as account_views
 from allauth.socialaccount import views as socialaccount_views, providers
 
-if settings.DJANGO_LOGIN:
+if settings.AUTHENTICATION_METHOD == 'django' or settings.HEROKU_DEMO:
     urlpatterns = [
         url(r'^standalone-login/$', login, name='standalone_login'),
         url(r'^standalone-logout/$', logout, name='standalone_logout', kwargs={'next_page': '/'}),
--- a/pontoon/base/templates/django/base.html
+++ b/pontoon/base/templates/django/base.html
@@ -80,7 +80,7 @@
                   <li><a href="{% url 'pontoon.admin' %}">Admin</a></li>
                   {% endif %}
 
-                  {% if settings.DJANGO_LOGIN %}
+                  {% if settings.AUTHENTICATION_METHOD == 'django' or settings.HEROKU_DEMO %}
                     {% if user.is_authenticated %}
                       <li id="standalone-sign-out"><a href="{% url 'standalone_logout' %}" title="{{ user.email|nospam }}"><i class="fa fa-sign-out-alt fa-fw"></i>Sign out</a></li>
                     {% else %}
--- a/pontoon/base/templates/header.html
+++ b/pontoon/base/templates/header.html
@@ -25,7 +25,7 @@
 
         {% if not user.is_authenticated() %}
         <div class="sign-in-header">
-          {% if settings.DJANGO_LOGIN %}
+          {% if settings.AUTHENTICATION_METHOD == 'django' or settings.HEROKU_DEMO %}
           <a href="{{ url('standalone_login') }}" class="button">Sign in</a>
           {% else %}
           <a id="fxa-sign-in" href="{{ provider_login_url(request) }}" class="button">Sign in</a>
@@ -58,7 +58,7 @@
               {% if user.is_authenticated() %}
                 <li><a href="{{ url('pontoon.contributors.settings') }}"><i class="fa fa-cog fa-fw"></i>Settings</a></li>
 
-                {% if settings.DJANGO_LOGIN %}
+                {% if settings.AUTHENTICATION_METHOD == 'django' or settings.HEROKU_DEMO %}
                 <li id="standalone-sign-out"><a href="{{ url('standalone_logout') }}" title="{{ user.email|nospam }}"><i class="fa fa-sign-out-alt fa-fw"></i>Sign out</a></li>
                 {% else %}
                 <li id="sign-out"><a href="{{ url('account_logout') }}" title="{{ user.email|nospam }}">{{ SignOut.csrf_form() }}<i class="fa fa-sign-out-alt fa-fw"></i>Sign out</a></li>
--- a/pontoon/base/templates/translate.html
+++ b/pontoon/base/templates/translate.html
@@ -152,7 +152,7 @@
               <li><a href="{{ url('pontoon.contributors.settings') }}"><i class="fa fa-cog fa-fw"></i>Settings</a></li>
             {% endif %}
 
-            {% if settings.DJANGO_LOGIN %}
+            {% if settings.AUTHENTICATION_METHOD == 'django' or settings.HEROKU_DEMO %}
               {% if user.is_authenticated() %}
                 <li class="standalone-sign-out"><a href="{{ url('standalone_logout') }}"><i class="fa fa-sign-out-alt fa-fw"></i>Sign out</a></li>
               {% else %}
@@ -429,7 +429,7 @@
         </section>
         {% else %}
           <p class="banner">
-            {% if settings.DJANGO_LOGIN %}
+            {% if settings.AUTHENTICATION_METHOD == 'django' or settings.HEROKU_DEMO %}
               {% set login_url = url('standalone_login') %}
             {% else %}
               {% set login_url = provider_login_url(request) %}
--- a/pontoon/base/templatetags/helpers.py
+++ b/pontoon/base/templatetags/helpers.py
@@ -255,7 +255,7 @@ def nospam(self):
 
 
 @library.global_function
-def provider_login_url(request, provider_id='fxa', **query):
+def provider_login_url(request, provider_id=settings.AUTHENTICATION_METHOD, **query):
     """
     This function adapts the django-allauth templatetags that don't support jinja2.
     @TODO: land support for the jinja2 tags in the django-allauth.
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -982,7 +982,7 @@ def user_data(request):
     user = request.user
 
     if not user.is_authenticated:
-        if settings.DJANGO_LOGIN:
+        if settings.AUTHENTICATION_METHOD == 'django' or settings.HEROKU_DEMO:
             login_url = reverse('standalone_login')
         else:
             login_url = provider_login_url(request)
@@ -992,7 +992,7 @@ def user_data(request):
             'login_url': login_url,
         })
 
-    if settings.DJANGO_LOGIN:
+    if settings.AUTHENTICATION_METHOD == 'django' or settings.HEROKU_DEMO:
         logout_url = reverse('standalone_logout')
     else:
         logout_url = reverse('account_logout')
--- a/pontoon/contributors/templates/contributors/profile.html
+++ b/pontoon/contributors/templates/contributors/profile.html
@@ -32,7 +32,7 @@
       {% if user.is_authenticated() %}
       <a href="mailto:{{ contributor.email|nospam }}">{{ contributor.email|nospam }}</a>
       {% else %}
-      {% if settings.DJANGO_LOGIN %}
+      {% if settings.AUTHENTICATION_METHOD == 'django' or settings.HEROKU_DEMO %}
         {% set login_url = url('standalone_login') %}
       {% else %}
         {% set login_url = provider_login_url(request) %}
--- a/pontoon/settings/base.py
+++ b/pontoon/settings/base.py
@@ -32,7 +32,7 @@ DEBUG = os.environ.get('DJANGO_DEBUG', 'False') != 'False'
 
 HEROKU_DEMO = os.environ.get('HEROKU_DEMO', 'False') != 'False'
 
-DJANGO_LOGIN = os.environ.get('DJANGO_LOGIN', 'False') != 'False' or HEROKU_DEMO
+AUTHENTICATION_METHOD = os.environ.get('AUTHENTICATION_METHOD', 'django')
 
 # Automatically log in the user with username 'AUTO_LOGIN_USERNAME'
 # and password 'AUTO_LOGIN_PASSWORD'
@@ -150,6 +150,7 @@ INSTALLED_APPS = (
     'allauth.account',
     'allauth.socialaccount',
     'allauth.socialaccount.providers.fxa',
+    'allauth.socialaccount.providers.github',
     'notifications',
     'graphene_django',
     'webpack_loader',


abowler2:
The Github sign in is working now and the database appears to be updating with the info. 
However, the success message from `/pontoon/base/adapter.py`  displays showing: “Your Persona account and Firefox Account have been connected.”  How would you like me to adjust that? (assuming I have the other pieces working correctly :wink:)

docker/set_up_webapp.sh
```diff
@@ -7,8 +7,8 @@ cd /app
 echo "Creating a new django superuser"
 python manage.py createsuperuser
 
-echo "Updating Firefox Account provider settings"
-python manage.py updatefxaprovider
+echo "Updating provider settings"
```
mathjazz: Maybe **authentication** provider?

docs/admin/deployment.rst
```diff
@@ -78,10 +78,10 @@ you create:
...
    Adds some additional django apps that can be helpful during day to day development.
 
-``DJANGO_LOGIN``
-   Required for non-Mozilla deployments, which cannot use `Firefox Accounts`_ for
-   login. Set to True if you want to use the default Django login instead. This will
-   allow you to log in via accounts created using `manage.py shell`.
+``AUTHENTICATION_METHOD``
+   The default value is `django`, which allows you to log in via accounts created using `manage.py shell`.
+   Set to 'fxa' if you want to use 'Firefox Accounts'.
+   Set to 'github' if you want to use 'GitHub'.
```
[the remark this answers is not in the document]
mathjazz: Also, it would also be nice to point out that corresponding `FXA_*` and `GITHUB_*` settings must be set for each of these methods.

pontoon/base/management/commands/update_auth_providers.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+FXA_PROVIDER_ID = FirefoxAccountsProvider.id
+GITHUB_PROVIDER_ID = GithubProvider.id
+
+
+class Command(BaseCommand):
+    help = ('Ensures an allauth application exists and has credentials that match settings')
+
+    if FXA_PROVIDER_ID:
```
mathjazz: IIUC this will always be `True`?

There a lot of code repetition here, so I suggest we refactor a little:
```python
def handle():
     if `FXA`:
        update_provider(fxa_data)

     if `GITHUB`:
        update_provider(github_data)

     # Ensure the provider applies to the current default site.

def update_provider(data):
    try:
    except:
```

pontoon/base/management/commands/update_auth_providers.py
```diff
@@ -0,0 +1,94 @@
...
+
+            # Ensure the provider applies to the current default site.
+            sites_count = app.sites.count()
+            if sites_count == 0:
+                default_site = Site.objects.get(pk=settings.SITE_ID)
+                app.sites.add(default_site)
+
+    elif GITHUB_PROVIDER_ID:
+        def handle(self, *args, **options):
+            # Check if Github_* settings are configured, bail if not
```
mathjazz: Nit: GitHub

pontoon/base/management/commands/update_auth_providers.py
```diff
@@ -0,0 +1,94 @@
...
+            if settings.GITHUB_CLIENT_ID is None or settings.GITHUB_SECRET_KEY is None:
+                self.stdout.write("Github_* settings unavailable; "
+                                  "skipping provider config.")
+                return
+
+            data = dict(
+                name='Github',
```
mathjazz: Nit: GitHub

pontoon/base/management/commands/update_auth_providers.py
abowler2: For this and the comment below ~ Does it matter that the django admin setup is named 'Github'?  If so, I'm assuming I should change it there as well.

mathjazz: What do you mean by django admin setup?

abowler2: From the tutorial, within the Django Administration panel I set up the "Social Applications" name to be "Github"  but I can edit that to "GitHub" if needed.

mathjazz: OK. I haven't tested your code yet, but the management command should take care of "everything". Users should not be required to change Django Administration to set up authentication.

abowler2:
Updating SOCIALACCOUNT_PROVIDERS with the settings from here:
https://django-allauth.readthedocs.io/en/latest/providers.html#github
asks for quite a bit more access.  Do we need all of those? If not, what changes do we need to make?

Attached are screenshots of the Rust Team's login (which is how I originally had it set up) and the updated login: 

Rust team - 
Screen Shot 2019-10-23 at 11 37 31 AM

New login - 
Screen Shot 2019-10-23 at 11 36 20 AM

mathjazz:
We definitely don't need that. Sorry for the hiccup. Ignore me. :)

pontoon/base/management/commands/update_auth_providers.py
```diff
@@ -0,0 +1,75 @@
...
+                client_id=settings.FXA_CLIENT_ID,
+                secret=settings.FXA_SECRET_KEY
+            )
+
+            return update_provider(fxa_data)
+
+        if 'GITHUB':
+            # Check if GitHub_* settings are configured, bail if not
+            if settings.GITHUB_CLIENT_ID is None or settings.GITHUB_SECRET_KEY is None:
+                self.stdout.write("Github_* settings unavailable; skipping provider config.")
```
mathjazz: GitHub

pontoon/base/management/commands/update_auth_providers.py
```diff
@@ -0,0 +1,75 @@
...
+
+FXA_PROVIDER_ID = FirefoxAccountsProvider.id
+GITHUB_PROVIDER_ID = GithubProvider.id
+
+
+class Command(BaseCommand):
+    help = ('Ensures an allauth application exists and has credentials that match settings')
+
+    def handle(self, *args, **options):
+        def update_provider(data):
```
mathjazz: Can you make this a standalone function, not part of `handle()`? You'll need to call it with `this.update_provider()`.

pontoon/base/management/commands/update_auth_providers.py
```diff
@@ -0,0 +1,75 @@
...
+                for k, v in data.items():
+                    setattr(app, k, v)
+                app.save()
+            except ObjectDoesNotExist:
+                # Create the provider if necessary.
+                app = SocialApp(**data)
+                app.save()
+                self.stdout.write("Created new authentication provider (pk=%s)" % app.pk)
+
+            # Ensure the provider applies to the current default site.
```
mathjazz: This block shouldn't be part of `update_provider()`. It's enough if it executes once, at the end of `handle()`.

pontoon/base/management/commands/update_auth_providers.py
```diff
@@ -0,0 +1,75 @@
...
+            sites_count = app.sites.count()
+            if sites_count == 0:
+                default_site = Site.objects.get(pk=settings.SITE_ID)
+                app.sites.add(default_site)
+
+        if 'FXA':
```
mathjazz: What I meant with my pseudocode example was to use the invert of the if clause below, something like:
```python
if settings.FXA_CLIENT_ID is not None or settings.FXA_SECRET_KEY is not None:
  # do_something()
```

abowler2: OK.  Would the line "self.stdout.write("FXA_* settings unavailable; skipping provider config.")" not be needed then, or would it go into an `else`

mathjazz: It would go into `else`. We'd also need to make sure at least one provider is available before running `# Ensure the provider applies to the`.

abowler2: Hmm...so if the FXA ID and KEY are `None` then the `else` will trigger the write line even if the GitHub ID and KEY are `not None`? Is that correct??

If so, should it be an `if/elif/else` block with the `else` writing a general "settings unavailable; skipping provider config." message?

pontoon/base/management/commands/update_auth_providers.py
abowler2: Moving this out of `handle()` causes "undefined name 'self' " errors on the 2 write lines in `update_provider`  
1)`self.stdout.write("Updating ....`
2) `self.stdout.write("Created ....`

Removing `self` generates an error of `undefined name 'stdout'`

mathjazz: The first argument of update_provider must be self I guess:
```python
def update_provider(self, data):
```

pontoon/base/management/commands/update_auth_providers.py
mathjazz: If it makes things easier, you can also log on success:
```python
if settings.FXA_CLIENT_ID is not None or settings.FXA_SECRET_KEY is not None:
  # do_something()
  # log_something()
if settings.GITHUB...:
  # do_something()
  # log_something()
```

abowler2: I went ahead and made the changes that I thought would maintain the originally intended function to bail if no settings found.  I just pushed those changes, so just let me know if anything needs revised.  Thanks.

pontoon/base/management/commands/update_auth_providers.py
```diff
@@ -0,0 +1,74 @@
...
+                name='GitHub',
+                provider=GITHUB_PROVIDER_ID,
+                client_id=settings.GITHUB_CLIENT_ID,
+                secret=settings.GITHUB_SECRET_KEY
+            )
+
+            return self.update_provider(github_data)
+
+        # Bail if neither FXA_* or GitHub_* settings are configured
+        else:
```
mathjazz: This should be `if provider_configured` and you would set `provider_configured = True` for each of the provider `if`-blocks above.

pontoon/base/management/commands/update_auth_providers.py
```diff
@@ -0,0 +1,74 @@
...
+            self.stdout.write("Authentication settings unavailable; skipping provider config.")
+            return
+
+        # Ensure the provider applies to the current default site.
+        sites_count = self.app.sites.count()
+        if sites_count == 0:
+            default_site = Site.objects.get(pk=settings.SITE_ID)
+            self.app.sites.add(default_site)
```
mathjazz: These two blocks should be in an `if-else`-block:

```python
if provider_configured:
    # Ensure the provider applies...
   ...
else:
    self.stdout.write()
```

Of course you would need to set `provider_configured = True` in each of the provider `if`-blocks above.

pontoon/base/management/commands/update_auth_providers.py
```diff
@@ -0,0 +1,71 @@
...
+                name='GitHub',
+                provider=GITHUB_PROVIDER_ID,
+                client_id=settings.GITHUB_CLIENT_ID,
+                secret=settings.GITHUB_SECRET_KEY
+            )
+
+            provider_configured = True
+            return self.update_provider(github_data)
+
+        # Bail if neither django-allauth provider configured
```
mathjazz: You can remove this comment now.

pontoon/base/management/commands/update_auth_providers.py
```diff
@@ -0,0 +1,71 @@
...
+            # Grab the credentials from settings
+            fxa_data = dict(
+                name='FxA',
+                provider=FXA_PROVIDER_ID,
+                client_id=settings.FXA_CLIENT_ID,
+                secret=settings.FXA_SECRET_KEY
+            )
+
+            provider_configured = True
+            return self.update_provider(fxa_data)
```
mathjazz: You shouldn't return here (or below), otherwise the rest of the code will never execute.

pontoon/base/management/commands/update_auth_providers.py
```diff
@@ -0,0 +1,70 @@
...
+FXA_PROVIDER_ID = FirefoxAccountsProvider.id
+GITHUB_PROVIDER_ID = GithubProvider.id
+
+
+class Command(BaseCommand):
+    help = ('Ensures an allauth application exists and has credentials that match settings')
+
+    def update_provider(self, data):
+        try:
+            # Update the existing provider with current settings.
```
mathjazz: Nit: I would move this comment above `try`. Same for the comment in the `except` block.

pontoon/base/management/commands/update_auth_providers.py
```diff
@@ -0,0 +1,70 @@
...
+            app = SocialApp.objects.get(provider=data['provider'])
+            self.stdout.write("Updating existing authentication provider (pk=%s)" % app.pk)
```
mathjazz: I would move this to the bottom, after `app.save()` - to follow the patter in the `except` block.

pontoon/base/management/commands/update_auth_providers.py
```diff
@@ -0,0 +1,70 @@
...
+            for k, v in data.items():
+                setattr(app, k, v)
+            app.save()
+        except ObjectDoesNotExist:
```
mathjazz: Nit: I'd add a blank line above for better readability.

docker/config/webapp.env.template
```diff
@@ -9,3 +9,5 @@ FXA_CLIENT_ID=2651b9211a44b7b2
 FXA_SECRET_KEY=a3cafccbafe39db54f2723f8a6f804c337e362950f197b5b33050d784129d570
 FXA_OAUTH_ENDPOINT=https://oauth-stable.dev.lcip.org/v1
 FXA_PROFILE_ENDPOINT=https://stable.dev.lcip.org/profile/v1
+GITHUB_CLIENT_ID=b0f6bd92211f1cfb0f60
+GITHUB_SECRET_KEY=1a123d2c7d567f8d1520b86d89f1d5cf78a00e42
```
mathjazz: We'll need to update those with something within mozilla/pontoon*.

pontoon/base/management/commands/update_auth_providers.py
```diff
@@ -0,0 +1,70 @@
...
+
+from __future__ import absolute_import
+
+from django.core.exceptions import ObjectDoesNotExist
+from django.core.management.base import BaseCommand
+from django.conf import settings
+from django.contrib.sites.models import Site
+from allauth.socialaccount.models import SocialApp
+from allauth.socialaccount.providers.fxa.provider import FirefoxAccountsProvider
+from allauth.socialaccount.providers.github.provider import GithubProvider
```
mathjazz: `GitHubProvider`

pontoon/base/management/commands/update_auth_providers.py
```diff
@@ -0,0 +1,70 @@
...
+                provider=GITHUB_PROVIDER_ID,
+                client_id=settings.GITHUB_CLIENT_ID,
+                secret=settings.GITHUB_SECRET_KEY
+            )
+
+            provider_configured = True
+            self.update_provider(github_data)
+
+        if provider_configured:
+            # Ensure the provider applies to the current default site.
```
mathjazz: Sorry for not testing this earlier properly: we should move this block to the bottom of `update_provider()`, because  we need a reference to `app`.

We should then just drop `provider_configured` and "skipping provider config" log (which isn't very useful anyways).

abowler2:
I'm so happy it worked!!  😄 
Just pushed the changes you requested, let me know if there are any other adjustments needed. 

And, I do still get the "Persona" message. I can't remember if I got it on the very first log in or if it was only after my token was saved though....FWIW, the message comes down and then disappears pretty quickly (which I was going to mention, all the messages are set a bit quick so if you don't know they are coming you can miss them or if they are long there isn't quite enough time to catch what they say)

pontoon/base/management/commands/update_auth_providers.py
```diff
@@ -10,42 +10,47 @@
...
 
     def update_provider(self, data):
+        # Update the existing provider with current settings.
         try:
-            # Update the existing provider with current settings.
             app = SocialApp.objects.get(provider=data['provider'])
-            self.stdout.write("Updating existing authentication provider (pk=%s)" % app.pk)
             for k, v in data.items():
                 setattr(app, k, v)
+            self.stdout.write("Updating existing authentication provider (pk=%s)" % app.pk)
```
mathjazz: Nit: After `save()` please, and then you can say "Updated" instead of "Updating" :)

mathjazz:
It turns out I wasn't seeing that `Your Persona account and Firefox Account have been connected.` message because I have my Pontoon account connected also with FXA.

Questions for @Pike and @jotes:
* Should we just remove that [`if`](https://github.com/mozilla/pontoon/blob/master/pontoon/base/adapter.py#L42-L46) block in [`pontoon/base/adapter.py`](https://github.com/mozilla/pontoon/blob/master/pontoon/base/adapter.py#L42-L46)?
* Or replace it with a more general message along the lines of `You have logged in to your existing Pontoon account.`

I think we should still connect accounts where possible.

docker/config/webapp.env.template
```diff
@@ -9,3 +9,5 @@ FXA_CLIENT_ID=2651b9211a44b7b2
 FXA_SECRET_KEY=a3cafccbafe39db54f2723f8a6f804c337e362950f197b5b33050d784129d570
 FXA_OAUTH_ENDPOINT=https://oauth-stable.dev.lcip.org/v1
 FXA_PROFILE_ENDPOINT=https://stable.dev.lcip.org/profile/v1
+GITHUB_CLIENT_ID=f70b793c5bf6a66aa3d8
+GITHUB_SECRET_KEY=dd5a595346d645b25e6107288298512e92bea9e2
```
mathjazz: Sorry @abowler2. As agreed with @Pike, we should actually remove those two lines, because they aren't meant to be public anyways. Users we'll need to set up their own to use GitHub OAuth, and we have that documented.

abowler2: No worries. Done and pushed.

mathjazz:
Thanks! Let's figure out that [message](https://github.com/mozilla/pontoon/pull/1442#issuecomment-547633064) and we're good to go!

Pike:
Both options sounds good to me. I'd be fine with also just removing it. Then we don't have to think about eventually localizing it ;-)

jotes:
@mathjazz I think that's better to remove the message If Pontoon doesn't need to support multiple authentication services at the same time (e.g. fxa and github).

mathjazz:
Please fix flake8 error and we're ready to merge this!

abowler2:
If it's not one thing, it's another 😆 
Working on it now.

jotes:
@abowler2 congratulations!
superb

PR #1443, opened 2019-10-21
Fix bug 1514555 - Remove ellipsis and allow string to overflow

katherine95:
#### What does this PR do?
This pull request fixes the issue [1514555](https://bugzilla.mozilla.org/show_bug.cgi?id=1514555). It allows the string list to overflow instead of ellipsis.

#### Type of change:
- [x] Bug fix (non-breaking change which fixed an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update.

#### How Has This Been Tested?
- Clone this repository 
- Git checkout to `katherine95:1514555-remove-ellipsis` branch.
- Follow the setup guidelines to set up pontoon locally
- Run `make build` then `make run`
- Login to Pontoon
- Navigate to the tutorial project to load some string, e.g
`http://localhost:8000/projects/tutorial/playground/`
- The strings in the string list should overflow instead of ellipsis.

#### Checklist:
- [x]At the tutorial project the strings in the string list should overflow.

#### Screenshots (if appropriate):
Before
Screenshot from 2019-10-23 18-27-50

After
Screenshot from 2019-10-23 18-23-19

### DIFF
--- a/frontend/src/modules/entitieslist/components/Entity.css
+++ b/frontend/src/modules/entitieslist/components/Entity.css
@@ -33,7 +33,6 @@
 .entity .source-string,
 .entity .translation-string {
     display: inline-block;
-    height: fit-content;
     line-height: 22px;
     margin: 0;
     width: 49%;
--- a/frontend/src/modules/entitieslist/components/Entity.css
+++ b/frontend/src/modules/entitieslist/components/Entity.css
@@ -35,7 +35,7 @@
     display: inline-block;
     line-height: 22px;
     margin: 0;
-    width: 49%;
+    width: 100%;
 }
 
 .entity .translation-string {
--- a/frontend/src/modules/entitieslist/components/Entity.css
+++ b/frontend/src/modules/entitieslist/components/Entity.css
@@ -18,7 +18,6 @@
 }
 
 .entity > div {
-    height: 22px;
     line-height: 22px;
 }
 
@@ -33,9 +32,10 @@
 .entity .source-string,
 .entity .translation-string {
     display: inline-block;
+    height: 22px;
     line-height: 22px;
     margin: 0;
-    width: 100%;
+    width: 49%;
 }
 
 .entity .translation-string {
--- a/frontend/src/modules/entitieslist/components/Entity.css
+++ b/frontend/src/modules/entitieslist/components/Entity.css
@@ -32,10 +32,10 @@
 .entity .source-string,
 .entity .translation-string {
     display: inline-block;
-    height: 22px;
     line-height: 22px;
     margin: 0;
     width: 49%;
+    vertical-align: top;
 }
 
 .entity .translation-string {
--- a/frontend/src/modules/entitieslist/components/Entity.css
+++ b/frontend/src/modules/entitieslist/components/Entity.css
@@ -36,8 +36,6 @@
     height: 22px;
     line-height: 22px;
     margin: 0;
-    overflow: hidden;
-    text-overflow: ellipsis;
     white-space: nowrap;
     width: 49%;
 }
--- a/frontend/src/modules/entitieslist/components/Entity.css
+++ b/frontend/src/modules/entitieslist/components/Entity.css
@@ -33,10 +33,9 @@
 .entity .source-string,
 .entity .translation-string {
     display: inline-block;
-    height: 22px;
+    height: fit-content;
     line-height: 22px;
     margin: 0;
-    white-space: nowrap;
     width: 49%;
 }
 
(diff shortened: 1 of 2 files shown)


katherine95:
> Thanks for the patch, but this is not working as expected. Perhaps the bug report wasn't clear, sorry about that.
> 
> I'm attaching the screenshot of the desired outcome. The idea is that the entire text should be visible.
> 
> Posnetek zaslona 2019-10-21 ob 18 59 55

From which interface? I don't understand which bug fix I am supposed to fix

katherine95:
> > Thanks for the patch, but this is not working as expected. Perhaps the bug report wasn't clear, sorry about that.
> > I'm attaching the screenshot of the desired outcome. The idea is that the entire text should be visible.
> > Posnetek zaslona 2019-10-21 ob 18 59 55
> 
> From which interface? I don't understand which bug fix I am supposed to fix

@mathjazz this is the [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1514555) that this PR fixes. . Just to  seek clarity what do you mean by `the entire text should be visible?`

mathjazz:
With your changes, the text overflows behind the scrollbar, so you need to scroll horizontally to reveal it.

What we want instead is shown on the screenshot I posted: the text should overflow vertically.

katherine95:
> With your changes, the text overflows behind the scrollbar, so you need to scroll horizontally to reveal it.
> 
> What we want instead is shown on the screenshot I posted: the text should overflow vertically.

I now get it. Is it as below?
Screenshot from 2019-10-22 00-50-52

mathjazz:
Also not the [fit-content](https://caniuse.com/#search=fit-content) is an experimental function, not supported in all [browsers Pontoon supports](https://mozilla-pontoon.readthedocs.io/en/latest/dev/contributing.html#browser-support).

katherine95:
> Also not the [fit-content](https://caniuse.com/#search=fit-content) is an experimental function, not supported in all [browsers Pontoon supports](https://mozilla-pontoon.readthedocs.io/en/latest/dev/contributing.html#browser-support).

I have also noticed fixing the CSS in the frontend part of the code does not fix the issue. I mean fixing at `translate.css` fixes the issue locally, but fixing at `entity.css` does not fix.

katherine95:
> The code doesn't work like your screenshot shows:
> Screenshot 2019-10-22 at 09 16 40
> 
> Also note that [fit-content](https://caniuse.com/#search=fit-content) is an experimental function, not supported in all [browsers Pontoon supports](https://mozilla-pontoon.readthedocs.io/en/latest/dev/contributing.html#browser-support).
> 
> You should simply remove `height`, including on the parent `div` element.

My bad, I just noticed I forgot to activate the new feature. I just did so and came across this. Let me work on fixing it

katherine95:
How about this? Is it okay? @mathjazz 
Screenshot from 2019-10-22 12-18-13

mathjazz:
That's not OK sadly, please have a look at my comment above to see the desired outcome:
https://github.com/mozilla/pontoon/pull/1443#pullrequestreview-304696291

katherine95:
> That's not OK sadly, please have a look at my comment above to see the desired outcome:
> [#1443 (review)](https://github.com/mozilla/pontoon/pull/1443#pullrequestreview-304696291)

Between what I have and the desired outcome is the space in between the breaking sentence. Right? I would like to understand if I am fixing the right thing. @mathjazz

mathjazz:
The difference is that you changed the `width` to `100%`, so strings are no longer in two columns, and that `height` is still limited by the parent `<div>` element setting (`22px`).

You should keep the `width` on strings unchanged and as I mentioned before, remove `height` also on the parent `<div>` element.

katherine95:
> The difference is that you changed the `width` to `100%`, so strings are no longer in two columns, and that `height` is still limited by the parent `<div>` element setting (`22px`).
> 
> You should keep the `width` on strings unchanged and as I mentioned before, remove `height` also on the parent `<div>` element.

Thank you for this. I realized I had not make a sample translation for me to figure this out

mathjazz:
Thanks for the update.

If the strings are of different sizes, this will happen:
Screenshot 2019-10-23 at 17 40 10

Which is not great - text should be vertically aligned to the top.

Note that you should also remove `height` from the `.entity .source-string, .entity .translation-string`.

katherine95:
> Thanks for the update.
> 
> If the strings are of different sizes, this will happen:
> Screenshot 2019-10-23 at 17 40 10
> 
> Which is not great - text should be vertically aligned to the top.
> 
> Note that you should also remove `height` from the `.entity .source-string, .entity .translation-string`.

That is true. It's not great. I just fixed it.

PR #1444, opened 2019-10-21
Fix bug 1590165 - Join RESOURCE and PROJECT entries in Metadata panel

monikamaheshwari:
Screenshot 2019-10-21 at 11 38 13 PM
Join RESOURCE and PROJECT entries in Metadata panel


frontend/src/modules/entitydetails/components/Metadata.css
```diff
@@ -49,3 +49,13 @@
     margin: 0;
     padding: 0;
 }
+
+.metadata .resource a:nth-child(2)::before {
+    padding-left: 3px;
+}
+
+.metadata .resource a:nth-child(2)::before {
+    content: "•";
```
mathjazz: That's not the right dot. Please use the [middot](https://en.wikipedia.org/wiki/Interpunct) instead.

frontend/src/modules/entitydetails/components/Metadata.css
```diff
@@ -49,3 +49,13 @@
     margin: 0;
     padding: 0;
 }
+
+.metadata .resource a:nth-child(2)::before {
```
mathjazz: Let's add a `className` and use that instead of `nth-child`. It's easier to maintain.

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -167,19 +167,17 @@ export default class Metadata extends React.Component<Props> {
...
-                        { entity.path }
-                    </a>
-                </Property>
-            </Localized>
-            <Localized id='entitydetails-Metadata--project' attrs={ { title: true } }>
-                <Property title='Project' className='project'>
-                    <a href={ `/${locale.code}/${entity.project.slug}/` }>
-                        { entity.project.name }
-                    </a>
+                    <span>
```
mathjazz: We don't need the `<span>`, do we?

monikamaheshwari: Yeah it is redundant.

frontend/src/modules/entitydetails/components/Metadata.css
```diff
@@ -49,3 +49,13 @@
...
+
+.metadata .resource a.resource-path::before {
+    padding-left: 3px;
+}
+
+.metadata .resource a.resource-path::before {
+    content: "\00b7";
+    color: #AAAAAA;
+    padding-right: 3px;
+}
```
mathjazz: These could be joined into one rule.

frontend/src/modules/entitydetails/components/Metadata.test.js
```diff
@@ -47,7 +47,7 @@ describe('<Metadata>', () => {
         const content = wrapper.find('Linkify').map(item => item.props().children);
         expect(content).toContain(ENTITY.comment);
 
-        expect(wrapper.find('#entitydetails-Metadata--resource a').text()).toContain(ENTITY.path);
+        expect(wrapper.find('#entitydetails-Metadata--resource a').at(1).text()).toContain(ENTITY.path);
```
mathjazz: Let's append `.resource-path` to the selector instead of using `at()` - same reason as for avoiding `nth-child`.

frontend/src/modules/entitydetails/components/Metadata.js
```diff
@@ -167,19 +167,16 @@ export default class Metadata extends React.Component<Props> {
             { this.renderSources(entity) }
             <Localized id='entitydetails-Metadata--resource' attrs={ { title: true } }>
                 <Property title='Resource' className='resource'>
+                    <a href={ `/${locale.code}/${entity.project.slug}/` }>
+                        { entity.project.name }
+                    </a>
                     <a
```
mathjazz: Sorry for not bringing this up earlier: could you instead of using `::before` pseudoclass add a delimiter `<span>` element? Right now the middot is clickable as well (it's part of the resource link), but it shouldn't be.

monikamaheshwari: Yes you are correct.

frontend/src/modules/entitydetails/components/Metadata.css
```diff
@@ -48,4 +48,4 @@
     list-style: none;
     margin: 0;
     padding: 0;
-}
+}
```
mathjazz: Nit: Mind the new line at the end of files.

PR #1445, opened 2019-10-22
Fix bug 1589137 - Autoreload django static files in dev.

adngdb:
Because of the way we set up static files for Translate.Next, in order to load them from the webpack server, we broke autoreloading of django static files. One had to rebuild the docker image in order to get updated static files. The problem was that the WhiteNoise middleware was taking over serving files, and in doing so only served from collected static files, thus never serving updated files in dev. With WhiteNoise disabled in dev, our static files serve function takes over and does the right thing at all time.

@mathjazz to be honest, I'm not sure I understand exactly what's happening there. WhiteNoise is supposed to serve files from disk when they get updated, but for whatever reason it doesn't. Anyway, this seems to work for me locally.

### DIFF
--- a/pontoon/settings/base.py
+++ b/pontoon/settings/base.py
@@ -134,6 +134,7 @@ INSTALLED_APPS = (
     'django.contrib.auth',
     'django.contrib.contenttypes',
     'django.contrib.sessions',
+    'whitenoise.runserver_nostatic',
     'django.contrib.staticfiles',
 
     # Django sites app is required by django-allauth
--- a/pontoon/settings/dev.py
+++ b/pontoon/settings/dev.py
@@ -3,7 +3,8 @@
 from __future__ import absolute_import
 
 import copy
-import base
+from . import base
+
 
 INSTALLED_APPS = base.INSTALLED_APPS + (
     # Provides a special toolbar which helps with tracking performance issues.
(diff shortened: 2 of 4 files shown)


pontoon/settings/base.py
```diff
@@ -134,6 +134,7 @@ def path(*args):
     'django.contrib.auth',
     'django.contrib.contenttypes',
     'django.contrib.sessions',
+    'whitenoise.runserver_nostatic',
```
adngdb: This is better than putting `--nostatic` in the `run_webapp.sh` command, because it means anyone manually running `./manage.py runserver` will get the correct behavior.

pontoon/settings/dev.py
```diff
@@ -3,7 +3,8 @@
 from __future__ import absolute_import
 
 import copy
-import base
+from . import base
```
adngdb: This has been broken for a while now, I guess. :(

pontoon/translate/views.py
```diff
@@ -47,7 +47,7 @@ def static_serve_dev(request, path, insecure=False, **kwargs):
     """
     try:
         # First try to load the file with django's regular serve view.
-        return serve(request, path, insecure=False, **kwargs)
+        return serve(request, path, insecure=True)
```
adngdb: These are just improvements that I found out while reading the documentation, they don't impact anything.

PR #1446, opened 2019-10-22
Bug 1579655 - Write tests for fluentoriginal module

nadinda:
There are React components that are not tested in fluentoriginal module. 

The details for added tests are :
- SourceString Component 
   - [x] render output the same as original input
   - [x] call correct function when clicking placeable
- SimpleString Component 
   - [x] render output from original input as simple string
   - [x] call correct function when clicking placeable
- RichString Component 
   - [x] render correct value and attributes
   - [x] render select expression from the original input properly
   - [x] call correct function when clicking placeable

FluentOriginalString did not get tested because it will be testing getSyntaxType() which already covered in https://github.com/mozilla/pontoon/blob/master/frontend/src/core/utils/fluent/getSyntaxType.test.js

### DIFF
--- a/frontend/src/modules/fluentoriginal/components/RichString.test.js
+++ b/frontend/src/modules/fluentoriginal/components/RichString.test.js
@@ -1,0 +1,63 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import sinon from 'sinon';
+
+import RichString from './RichString';
+
+
+const ORIGINAL = 'song-title = Hello\n    .genre = Pop\n    .album = Hello and Good Bye';
+
+const ENTITY = {
+    original: ORIGINAL,
+};
+
+describe('<RichString>', () => {
+    it('renders correct value and each attribute of original input', () => {
+        const wrapper = shallow(<RichString
+            entity = { ENTITY }
+        />);
+
+        expect(wrapper.find('RichString')).toBeDefined();
+
+        expect(wrapper.find('ContentMarker')).toHaveLength(3);
+        expect(wrapper.find('ContentMarker').at(0).html()).toContain('Hello');
+        expect(wrapper.find('ContentMarker').at(1).html()).toContain('Pop');
+        expect(wrapper.find('ContentMarker').at(2).html()).toContain('Hello and Good Bye');
+    });
+
+    it('renders select expression correctly', () => {
+        const input = `
+user-entry =
+    { PLATFORM() ->
+        [variant-1] Hello!
+       *[variant-2] Good Bye!
+    }`;
+
+        const entity = {
+            original: input,
+        };
+
+        const wrapper = shallow(<RichString
+            entity = { entity }
+        />);
+
+        expect(wrapper.find('ContentMarker')).toHaveLength(2);
+
+        expect(wrapper.find('label').at(0).html()).toContain('variant-1');
+        expect(wrapper.find('ContentMarker').at(0).html()).toContain('Hello!');
+
+        expect(wrapper.find('label').at(1).html()).toContain('variant-2');
+        expect(wrapper.find('ContentMarker').at(1).html()).toContain('Good Bye!');
+    });
+
+    it('calls the handleClickOnPlaceable function on click on .original', () => {
+        const handleClickOnPlaceable = sinon.spy();
+        const wrapper = shallow(<RichString
+            entity = { ENTITY }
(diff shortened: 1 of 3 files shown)


frontend/src/modules/fluentoriginal/components/RichString.test.js
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import sinon from 'sinon';
+
+import RichString from './RichString';
+
+
+const ORIGINAL = 'song-title = Hello\n    .genre = Pop\n    .album = Hello and Good Bye';
```
mathjazz: Let's use the same formatting here as for the `const input` below.

frontend/src/modules/fluentoriginal/components/RichString.test.js
```diff
@@ -0,0 +1,63 @@
...
+        const handleClickOnPlaceable = sinon.spy();
+        const wrapper = shallow(<RichString
+            entity = { ENTITY }
+            handleClickOnPlaceable={ handleClickOnPlaceable }
+        />);
+
+        wrapper.find('.original').simulate('click');
+        expect(handleClickOnPlaceable.called).toEqual(true);
+    });
+});
```
mathjazz: Nit: add newline at the end of the file (for all 3 files).

frontend/src/modules/fluentoriginal/components/RichString.test.js
```diff
@@ -0,0 +1,63 @@
...
+    original: ORIGINAL,
+};
+
+describe('<RichString>', () => {
+    it('renders correct value and each attribute of original input', () => {
+        const wrapper = shallow(<RichString
+            entity = { ENTITY }
+        />);
+
+        expect(wrapper.find('RichString')).toBeDefined();
```
mathjazz: I believe that's redundant (also in tests below), because it's pretty obvious it will be true given the initialization of the `wrapper` above.

frontend/src/modules/fluentoriginal/components/RichString.test.js
```diff
@@ -0,0 +1,63 @@
...
+
+
+const ORIGINAL = 'song-title = Hello\n    .genre = Pop\n    .album = Hello and Good Bye';
+
+const ENTITY = {
+    original: ORIGINAL,
+};
+
+describe('<RichString>', () => {
+    it('renders correct value and each attribute of original input', () => {
```
mathjazz: Nit: `renders value and each attribute correctly`

frontend/src/modules/fluentoriginal/components/RichString.test.js
```diff
@@ -0,0 +1,63 @@
...
+        const wrapper = shallow(<RichString
+            entity = { ENTITY }
+        />);
+
+        expect(wrapper.find('RichString')).toBeDefined();
+
+        expect(wrapper.find('ContentMarker')).toHaveLength(3);
+        expect(wrapper.find('ContentMarker').at(0).html()).toContain('Hello');
+        expect(wrapper.find('ContentMarker').at(1).html()).toContain('Pop');
+        expect(wrapper.find('ContentMarker').at(2).html()).toContain('Hello and Good Bye');
```
mathjazz: You could also check that labels are rendered correctly, same as below.

nadinda:
Hi @mathjazz I've changed it based on your suggestions. Let me know if there's more improvements I should make. Thanks!

frontend/src/modules/fluentoriginal/components/RichString.test.js
```diff
@@ -0,0 +1,71 @@
...
+import { shallow } from 'enzyme';
+import sinon from 'sinon';
+
+import RichString from './RichString';
+
+
+const ORIGINAL = `
+song-title = Hello
+.genre = Pop
+.album = Hello and Good Bye`;
```
mathjazz: A 4-space indent in front of `.` would be nice.

frontend/src/modules/fluentoriginal/components/RichString.test.js
mathjazz: We need blank line at the end of the **file**:
https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file

There shouldn't be one before the final `});`.

nadinda:
Fixed the newline and indentations

PR #1452, opened 2019-10-25
bug 1585251, more test pass on python 3

Pike:


pontoon/base/tests/views/test_tmx.py
```diff
@@ -39,7 +39,7 @@ def test_view_tmx_locale_file_dl(client, entity_a, locale_a):
     )
     assert response.status_code == 200
     _check_xml(
-        ''.join(response.streaming_content).encode('utf-8')
```
jotes: :+1:

adngdb:
Thanks for doing the review @jotes!

PR #1453, opened 2019-10-25
Fix bug 1395179 - Fix status numbers in progress menu to be clickable

katherine95:
#### What does this PR do?
This pull request fixes the issue [1395179](https://bugzilla.mozilla.org/show_bug.cgi?id=1395179). It ensures the status numbers in the progress menu are clickable.

#### Type of change:
- [x] Bug fix (non-breaking change which fixed an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update.

#### How Has This Been Tested?
- Clone this repository 
- Git checkout to `katherine95:1395179-clickable-status-numbers` branch.
- Follow the setup guidelines to set up pontoon locally
- Run `make run`
- Login to Pontoon
- Navigate to the progress menu and click the percentage.
- Click on the status numbers. For example, click  the number under "MISSING"

Expected result:

It navigates to `http://localhost:8000/translate/en-GB/tutorial/all-resources/?status=missing&string=1`

#### Checklist:
- [x] On click on any of the status numbers, it should navigate to the respective url.

#### Screenshot
Screenshot from 2019-11-13 14-50-35

### DIFF
--- a/frontend/src/modules/resourceprogress/components/ResourceProgress.css
+++ b/frontend/src/modules/resourceprogress/components/ResourceProgress.css
@@ -135,4 +135,5 @@
     font-size: 22px;
     font-weight: 300;
     padding: 5px 0 10px;
+    cursor: pointer;
 }
--- a/frontend/src/modules/resourceprogress/components/ResourceProgress.css
+++ b/frontend/src/modules/resourceprogress/components/ResourceProgress.css
@@ -135,5 +135,4 @@
     font-size: 22px;
     font-weight: 300;
     padding: 5px 0 10px;
-    cursor: pointer;
 }
--- a/frontend/src/modules/resourceprogress/components/ResourceProgress.js
+++ b/frontend/src/modules/resourceprogress/components/ResourceProgress.js
@@ -47,6 +47,14 @@ export class ResourceProgressBase extends React.Component<Props, State> {
         });
     }
 
+    // This method is called when a user clicks on the status numbers in progress menu.
+    handleClickStatusNumbers = (statusValue) => {
+        const url = window.location.pathname;
+        const params = new URLSearchParams({'status' : statusValue});
+        const newUrl = url + '?' + params.toString(); 
+        return newUrl;
+    }
+
     render() {
         const { approved, fuzzy, warnings, errors, missing, unreviewed, total } = this.props.stats;
         const complete = approved + warnings;
@@ -92,7 +100,7 @@ export class ResourceProgressBase extends React.Component<Props, State> {
                                 Translated
                             </Localized>
                         </span>
-                        <p className="value">{ asLocaleString(approved) }</p>
+                        <p className="value"><a href={ this.handleClickStatusNumbers('translated') } >{ asLocaleString(approved) }</a></p>
                     </div>
                     <div className="fuzzy">
                         <span className="title">
@@ -100,7 +108,7 @@ export class ResourceProgressBase extends React.Component<Props, State> {
                                 Fuzzy
                             </Localized>
                         </span>
-                        <p className="value">{ asLocaleString(fuzzy) }</p>
+                        <p className="value"><a href={ this.handleClickStatusNumbers('fuzzy') } >{ asLocaleString(fuzzy) }</a></p>
                     </div>
                     <div className="warnings">
                         <span className="title">
@@ -108,7 +116,7 @@ export class ResourceProgressBase extends React.Component<Props, State> {
                                 Warnings
                             </Localized>
                         </span>
-                        <p className="value">{ asLocaleString(warnings) }</p>
+                        <p className="value"><a href={ this.handleClickStatusNumbers('warnings') } >{ asLocaleString(warnings) }</a></p>
                     </div>
                     <div className="errors">
                         <span className="title">
@@ -116,7 +124,7 @@ export class ResourceProgressBase extends React.Component<Props, State> {
                                 Errors
                             </Localized>
                         </span>
-                        <p className="value">{ asLocaleString(errors) }</p>
+                        <p className="value"><a href={ this.handleClickStatusNumbers('errors') } >{ asLocaleString(errors) }</a></p>
                     </div>
                     <div className="missing">
                         <span className="title">
@@ -124,7 +132,7 @@ export class ResourceProgressBase extends React.Component<Props, State> {
                                 Missing
                             </Localized>
                         </span>
-                        <p className="value">{ asLocaleString(missing) }</p>
+                        <p className="value"><a href={ this.handleClickStatusNumbers('missing') } >{ asLocaleString(missing) }</a></p>
                     </div>
                 </div>
             </aside>
--- a/frontend/src/modules/resourceprogress/components/ResourceProgress.js
+++ b/frontend/src/modules/resourceprogress/components/ResourceProgress.js
@@ -11,10 +11,12 @@ import ProgressChart from './ProgressChart';
 
 import { asLocaleString } from 'core/utils';
 
+import type { NavigationParams } from 'core/navigation';
 import type { Stats } from 'core/stats';
 
 
 type Props = {|
+    parameters: NavigationParams,
     stats: Stats,
 |};
 
@@ -49,7 +51,9 @@ export class ResourceProgressBase extends React.Component<Props, State> {
     }
 
     render() {
-        const { approved, fuzzy, warnings, errors, missing, unreviewed, total } = this.props.stats;
+        const { parameters, stats } = this.props;
+        const { approved, fuzzy, warnings, errors, missing, unreviewed, total } = stats;
+        const currentPath = `/${parameters.locale}/${parameters.project}/${parameters.resource}/`;
         const complete = approved + warnings;
 
         let percent = 0;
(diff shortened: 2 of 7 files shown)


katherine95:
@mathjazz @adngdb kindly review.

frontend/src/modules/resourceprogress/components/ResourceProgress.js
```diff
@@ -92,39 +100,39 @@ export class ResourceProgressBase extends React.Component<Props, State> {
                                 Translated
                             </Localized>
                         </span>
-                        <p className="value">{ asLocaleString(approved) }</p>
+                        <p className="value"><a href={ this.handleClickStatusNumbers('translated') } >{ asLocaleString(approved) }</a></p>
```
adngdb: I think these should actually be `react-router` [Links](https://reacttraining.com/react-router/web/api/Link). Look at the examples in the documentation, you can easily manage the same features without using your custom `handleClickStatusNumbers` method.

katherine95: okay on it

katherine95: The `react-router` Links are working good, only that I am facing a blocker with the URL, the base URL is `domain:translate/` so the new URL turns to have `translate` twice in the URL as follows `http://localhost:8000/translate/translate/en-GB/tutorial/all-resources/?status=translated` I am working on finding a fix for this though.

adngdb: Have you tried setting only the query string? Anyway if you want help, the best way would be to show code. :-)

katherine95: Got it. Actually I was not supposed to use the `pathname` property.

frontend/src/modules/resourceprogress/components/ResourceProgress.js
```diff
@@ -100,39 +93,39 @@ export class ResourceProgressBase extends React.Component<Props, State> {
                                 Translated
                             </Localized>
                         </span>
-                        <p className="value"><a href={ this.handleClickStatusNumbers('translated') } >{ asLocaleString(approved) }</a></p>
+                        <p className="value"><Link to={{ search:"?status=translated" }}>{ asLocaleString(approved) }</Link></p>
```
adngdb: White space after colon (`:`) please.

adngdb: Also, please break these into several lines, like so:
```js
<p>
    <Link>
        { content }
    </Link>
</p>
```

adngdb:
> For case (1), the placeholder on the search bar only doesn't get updated until after clicking the resource panel.

Indeed, and that's a problem. Please fix it! :-)

> For case (2) the link location works fine

Not for me. When I copy a link location, it gives `http://localhost:8000/translate/?status=translated`. Notice that it lacks the `locale/project/resource` part, and that link leads to a broken UI.

katherine95:
> > For case (1), the placeholder on the search bar only doesn't get updated until after clicking the resource panel.
> 
> Indeed, and that's a problem. Please fix it! :-)
> 
> > For case (2) the link location works fine
> 
> Not for me. When I copy a link location, it gives `http://localhost:8000/translate/?status=translated`. Notice that it lacks the `locale/project/resource` part, and that link leads to a broken UI.

Working on it to fix case (1) and reproduce case (2)

mathjazz:
Hi @katherine95, let us know if you need any help. Thanks! :)

katherine95:
> Hi @katherine95, let us know if you need any help. Thanks! :)

Hi @mathjazz. Thanks, I am currently working on it. Actually I noticed the issue is the status on the search box placeholder is delayed to be updated by the state.

katherine95:
I have managed to fix updating the state of the status immediately.I am currently facing one issue, that is having one status at the search box, I working on that.
Screenshot from 2019-11-13 14-10-08

katherine95:
> I have managed to fix updating the state of the status immediately.I am currently facing one issue, that is having one status at the search box, I working on that.
> Screenshot from 2019-11-13 14-10-08

fixed

frontend/src/modules/resourceprogress/components/ResourceProgress.css
```diff
@@ -135,4 +135,5 @@
     font-size: 22px;
     font-weight: 300;
     padding: 5px 0 10px;
+    cursor: pointer;
```
adngdb: This is not needed now (because the elements are links, so the pointer automatically gets turned into a pointer).

frontend/src/modules/resourceprogress/components/ResourceProgress.js
```diff
@@ -92,39 +93,59 @@ export class ResourceProgressBase extends React.Component<Props, State> {
                                 Translated
                             </Localized>
                         </span>
-                        <p className="value">{ asLocaleString(approved) }</p>
+                        <p className="value" onClick={ this.toggleVisibility }>
+                            <Link to={{ search: "?status=translated" }}>
```
adngdb: This still doesn't work, but I'm not sure how best to solve the issue here. I think we'll want to upgrade `react-router` to be able to use a new feature of theirs. Give me some more time, I'll investigate.

frontend/src/modules/search/components/SearchBox.js
```diff
@@ -3,6 +3,7 @@
 import * as React from 'react';
 import { connect } from 'react-redux';
 import debounce from 'lodash.debounce';
+import _ from 'lodash';
```
adngdb: We try not to import `lodash` directly, but instead install just the module that we need.

frontend/src/modules/search/components/SearchBox.js
```diff
@@ -143,6 +144,22 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
         }
     }
 
+    static getDerivedStateFromProps(prevProps: InternalProps, nextProps: Object) {
```
adngdb: This could do with some comments to explain what it's doing and why.

frontend/src/modules/search/components/SearchBox.js
katherine95: Can I just import `_`  from `lodash-core`?

adngdb: No, please look at what we do just on the previous line. We want to import only the `mapValues` function, and nothing else. Just like, on the line before, we import only `debounce`. For that to work, you need to install `lodash.mapValues` in our frontend dependencies. That process is documented here: https://github.com/mozilla/pontoon/tree/master/frontend#dependencies

frontend/src/modules/resourceprogress/components/ResourceProgress.js
adngdb: OK, figured it out. No need to update anything. Just add `pathname` here. To set its value correctly, you'll need to pass `parameters` from `App.js` and reconstitute it in a variable. For example:

```js
const currentPath = `/${params.locale}/${params.project}/${params.resource}/`;
```

frontend/src/modules/search/components/SearchBox.js
adngdb: After reading the documentation for `getDerivedStateFromProps`, I'm getting wary about using it. Can you please explain precisely what you're achieving here, and maybe we can find a different, simpler way to do it?

frontend/src/modules/search/components/SearchBox.js
katherine95: I now see. Thank you for this.

frontend/src/modules/resourceprogress/components/ResourceProgress.js
katherine95: I will work on this feedback.

frontend/src/modules/search/components/SearchBox.js
katherine95: I just added comments but I have not pushed the code yet.
The local state in that component is delayed to get updated, so after clicking on the status numbers, the status value on the search box placeholder does not get updated  not until one refreshes the page. 
I used the `getDerivedStateFromPops` to ensure the state of the statuses is updated upon changes in the DOM.

katherine95: I couldn't use `componentDidUpdate` to set the new state of statuses because it caused an infinite loop. 
In this function I also need to change the parameter `nextProps` to `state` to be more readable.
The setState called in `componentDidUpdate` does not do anything since `search` key from props is always null. I can use the lifecycle method to update state of the statuses. 
I understand the caveats of using `getDerivedStateFromProps`. Let me see what I can do about it. It should be the last option.

frontend/src/modules/search/components/SearchBox.js
```diff
@@ -3,9 +3,8 @@
 import * as React from 'react';
 import { connect } from 'react-redux';
 import debounce from 'lodash.debounce';
-import _ from 'lodash';
 import isEqual from 'lodash.isequal';
-
+import mapValues from 'lodash.mapvalues';
 import './SearchBox.css';
```
adngdb: Please keep the empty line before this CSS import.

frontend/src/modules/search/components/SearchBox.js
```diff
@@ -132,7 +131,7 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
         this.updateFiltersFromURLParams();
     }
 
-    componentDidUpdate(prevProps: InternalProps) {
+    componentDidUpdate(prevProps: InternalProps, state: Object) {
```
adngdb: This is not needed if `state` is never used in the function.

frontend/src/modules/search/components/SearchBox.js
katherine95: @adngdb what's your take on this? 
I have not found another way yet of updating the state of statuses immediately.

adngdb: I thought about it more, and I now think this is actually the correct solution. Congrats on finding it! :-)

PR #1454, opened 2019-10-28
Bug 1555259: Copying from Machinery to Source editor results in a broken string

Karskaya:
In `editor.js->componentDidUpdate` add `else if` to check the content copied from machinery tab.

In case parsed to fluent message content is Junk, convert it to complex syntax.

Change FTL strings with attribute to add translation for key id in `getReconstructedMessage.js` to has the same behavior with rich editor.

Change test `getReconstructedMessage` to pass translation in key id instead of attribute.

### DIFF
--- a/frontend/src/modules/entitieslist/components/Entity.css
+++ b/frontend/src/modules/entitieslist/components/Entity.css
@@ -18,7 +18,6 @@
 }
 
 .entity > div {
-    height: 22px;
     line-height: 22px;
 }
 
@@ -33,13 +32,10 @@
 .entity .source-string,
 .entity .translation-string {
     display: inline-block;
-    height: 22px;
     line-height: 22px;
     margin: 0;
-    overflow: hidden;
-    text-overflow: ellipsis;
-    white-space: nowrap;
     width: 49%;
+    vertical-align: top;
 }
 
 .entity .translation-string {


Karskaya:
In this PR I've changed `getReconstructedMessage` and 'getReconstructedMessage.test` and I'm not sure that it was a right decision.
If user click on machinery translation while rich editor is on, translation is placed into the `Value` field. Also translations in machinery tab are suitable for string placed in the Value field of the original string. So, I made a conclusion that translation was put in a attribute by mistake.
If I'm right, than for multilines the same changes should be done here: https://github.com/mozilla/pontoon/blob/a0c716122dbebba6fb2f8d221713e9928542ef2b/frontend/src/core/utils/fluent/getReconstructedMessage.js#L31
Please, confirm @adngdb @mathjazz 

Also, I have a question. 
In case of fluent string with attribute, if one of the fields(attribute or value) is empty, system count it as Junk. So, it's possible to convert from rich editor to source editor with only one value, but if you want to convert back - system will make both fields empty. Is it an expected result or no?

frontend/src/core/utils/fluent/getReconstructedMessage.js
```diff
@@ -31,7 +31,7 @@ export default function getReconstructedMessage(original: string, translation: s
             translation.split('\n').forEach(t => content += `\n        ${t}`);
         }
         else {
-            content = `${key} =\n    .${attribute} = ${translation}`;
+            content = `${key} = ${translation}\n    .${attribute} =`;
```
adngdb: That change is not correct. It creates an invalid Fluent message. There is no bug here, so you definitely shouldn't make any change of that sort.

PR #1457, opened 2019-10-28
bug 1585251, more test fixes for python3

Pike:
There's an intentional change to the serialization of the json for web extensions, it's now dropping the trailing `" "` in `, \n`.

### DIFF
--- a/pontoon/batch/forms.py
+++ b/pontoon/batch/forms.py
@@ -26,8 +26,8 @@ class BatchActionsForm(forms.Form):
 
         Related bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1438575
         """
-        field_val = str(self.cleaned_data.get(param_name, ''))
-        return urllib.parse.unquote(field_val).decode('utf-8')
+        field_val = self.cleaned_data.get(param_name, '')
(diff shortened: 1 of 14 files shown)


pontoon/batch/forms.py
```diff
@@ -26,8 +26,8 @@ def decode_field(self, param_name):
 
         Related bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1438575
         """
-        field_val = str(self.cleaned_data.get(param_name, ''))
-        return urllib.parse.unquote(field_val).decode('utf-8')
+        field_val = self.cleaned_data.get(param_name, '')
+        return urllib.parse.unquote(field_val)
```
Pike: @jotes, would you be able to dig up in your brain why this code was like this?

You originally wrote this for https://bugzilla.mozilla.org/show_bug.cgi?id=1438575.

jotes: If I remember correctly, that's needed because the form gets data from the query-string of a request. Django forms weren't decoding those populated fields correctly and the backend was searching for `%26nbsp%3B` instead of `&nbsp;`.
But you're right, decoding to UTF-8 isn't required here.

PR #1459, opened 2019-10-28
Bug 1555259: Copying from Machinery to Source editor results in a broken string

Karskaya:
In editor.js->componentDidUpdate add else if to check the content copied from machinery tab.
In case parsed to fluent message content is Junk, convert it to complex

[Bug 1555259](https://bugzilla.mozilla.org/show_bug.cgi?id=1555259)

### DIFF
--- a/frontend/src/modules/fluenteditor/components/Editor.js
+++ b/frontend/src/modules/fluenteditor/components/Editor.js
@@ -92,6 +92,24 @@ export class EditorBase extends React.Component<EditorProps, State> {
         ) {
             this.analyzeFluentMessage(this.props.editor.translation);
         }
+        // If translation changes from machinery tab,
+        // check if it's valid and convert syntax to comlex if no.
+        else if (
+            this.props.entity &&
+            this.state.forceSource &&
+            this.props.editor.translation !== prevProps.editor.translation &&
+            this.props.editor.changeSource === 'machinery' &&
+            typeof(this.props.editor.translation) === 'string'
+        ) {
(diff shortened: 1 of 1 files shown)


Karskaya:
from this [PR](https://github.com/mozilla/pontoon/pull/1454)
> The value of a Fluent message can be empty if the message has attributes. However, an attribute cannot be empty. It is expected that Fluent complains about that (and we've had to cheat a bit around that in some places).

@adngdb This made me confusing and I did a wrong conclusion about correct system behaviour.

What about rich editor? Machinery value will be inserted into Fluent message, not into an attribute. Should that stay as it is? Would it be rewritten with this issue - [[FTL] Make machinery usable for complex strings](https://bugzilla.mozilla.org/show_bug.cgi?id=1453341 )?

rich

Karskaya:
This fix cause an issue.
If the string in Source editor looks like this:
`serviceworker-worker-debug-forbidden =`
`  .title = Iskanje napak`
After user toggle from advanced to simple FTL mode, instead of rich syntax string receives simple.

adngdb:
Hey @Karskaya, this is a good catch, but it's not related to the bug you are working on. It can be reproduced on master. Also, I don't know if that's actually a bug or not. Anyway, I'll file a bug and we'll see. Thanks!

Karskaya:
> Hey @Karskaya, this is a good catch, but it's not related to the bug you are working on. It can be reproduced on master. Also, I don't know if that's actually a bug or not. Anyway, I'll file a bug and we'll see. Thanks!

@adngdb Thank you! I'll be glad to help with that bug, just in case.

PR #1460, opened 2019-10-28
No bug: Fix Django login avatar regression

mathjazz:
Fix regression introduced in https://github.com/mozilla/pontoon/commit/ff4f1d2158278b95b55f338fb0d0ae4e7a883a90#diff-91c36325f8496b8a88e5688322739387.

### DIFF
--- a/pontoon/base/templates/django/base.html
+++ b/pontoon/base/templates/django/base.html
@@ -55,8 +55,7 @@
             <div id="profile" class="select">
               <div class="button selector">
                 {% if user.is_authenticated %}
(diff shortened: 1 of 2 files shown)


pontoon/base/templates/django/base.html
```diff
@@ -55,8 +55,7 @@
             <div id="profile" class="select">
               <div class="button selector">
                 {% if user.is_authenticated %}
-                  <span class="title noselect">{{ user.name_or_email|nospam }}</span>
```
mathjazz: We got rid of that in other places looong time ago.

PR #1462, opened 2019-10-29
Fix bug 1377887: Selecting multiple filters/entities should be more intuitive

monikamaheshwari:
Ability to select multiple filters/entities should be more intuitive

### DIFF
--- a/frontend/src/modules/entitieslist/components/Entity.css
+++ b/frontend/src/modules/entitieslist/components/Entity.css
@@ -79,6 +79,11 @@
     content: "";
 }
 
+.entity:hover .status:before {
+    content: "";
+    font-weight: 100;
+}
+
 .entity.batch-editable .status:hover:before {
     content: "";
 }
--- a/frontend/src/modules/entitieslist/components/Entity.css
+++ b/frontend/src/modules/entitieslist/components/Entity.css
@@ -85,13 +85,13 @@
     font-weight: bold;
 }
 
-.entity:hover .status:before {
+.entity:hover .status:before,
+.entity.selected:hover .status:before {
     content: "";
     font-weight: 100;
 }
 
-.entity.checked:hover .status:before,
-.entity.selected:hover .status:before {
+.entity.checked:hover .status:before {
     content: "";
     font-weight: bold;
 }
--- a/frontend/src/modules/entitieslist/components/Entity.css
+++ b/frontend/src/modules/entitieslist/components/Entity.css
@@ -79,30 +79,22 @@
     content: "";
 }
 
-.entity.checked .status:before,
-.entity.selected .status:before {
-    content: "";
-    font-weight: bold;
-}
-
-.entity:hover .status:before,
-.entity.selected:hover .status:before {
+.entity.batch-editable:hover > .status:before,
+.entity.batch-editable.selected:hover > .status:before {
     content: "";
     font-weight: 100;
 }
 
-.entity.checked:hover .status:before {
-    content: "";
-    font-weight: bold;
-}
-
-.entity.batch-editable .status:hover:before {
+.entity.batch-editable > .status:hover:before,
+.entity.batch-editable:not(.checked) > .status:hover:before,
+.entity.batch-editable.selected > .status:hover:before {
     content: "";
     font-weight: bold;
 }
 
 .entity.batch-editable.checked > .status:before {
     content: "";
+    font-weight: bold;
 }
 
 .entity.approved .status:before {
--- a/frontend/src/modules/entitieslist/components/Entity.css
+++ b/frontend/src/modules/entitieslist/components/Entity.css
@@ -79,9 +79,21 @@
     content: "";
 }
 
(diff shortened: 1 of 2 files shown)


monikamaheshwari:
I have made the icon into regular squares on whole list item hover. Let me know if you want to change it on only text hover.

mathjazz:
Hi Monika, where the icon changes is OK (no need to make the change only when you hover text).

But the icon is not a regular square, in most cases it's solid (you'll need to set `font-weight` to 100 to force it as a regular square).

The square also moves and grows as you move between icons. It should always keep the same position relative to the line and the same size.

mathjazz:
Thanks for the update. I still see some jumping while hovering over filters - I feel like the square in front of All, Empty, Rejected and Translation Time filters should be moved 1 px to the right.

mathjazz:
Thanks for the update!

I believe the square icon that appears when hovering all three icons of the Extra filters and the Time Range filter should still be moved 1px to the right. It feels like the icons are jumping left and right on hover and unhover.

Getting there!

mathjazz:
Thanks for the update, Monika!

@flodolo I've deployed this to stage, what are your thoughts on the UX?
https://mozilla-pontoon-staging.herokuapp.com/translate/it/lockwise-ios/lockwise-ios.xliff/?string=199220

flodolo:
Doesn't look too bad
Schermata 2019-10-31 alle 11 35 18

mathjazz:
I can't decide which variant to pick... The first one (empty square) maybe looks more consistent when you take translation authors into account (avatar borders also look like empty squares).

Now I wonder what should we do with the string list and batch editing actions? Try the behaviour there as well?

@monikamaheshwari I noticed the tick icon is no longer centered in the Translation Authors avatars. And, the small square appears withing the avatar when you slowly move the mouse to the upper part of it. Could you please fix that?

monikamaheshwari:
@mathjazz I was able to reproduce and fixed it.

mathjazz:
Hi @monikamaheshwari, ticks no longer appear when you hover icons (except for the Authors).

Could you also try chaning the icon to ?

mathjazz:
> Now I wonder what should we do with the string list and batch editing actions? Try the behaviour there as well?

@monikamaheshwari Thanks for the update! Could you please also make the change in the string list? The status icon of the hovered string should change to the checked square - .

mathjazz:
@monikamaheshwari I'm getting a b0rken icon when I hover the checkbox:

Screenshot 2019-11-12 at 20 16 19

Could you please fix that?

Also, we shouldn't change the icon on hover if it's checked. Behaviour should be the same as in filter menu.

String list (regular checkbox - not OK):
Screenshot 2019-11-12 at 20 18 54

Filter menu (solid checkbox - OK):
Screenshot 2019-11-12 at 20 18 46

mathjazz:
Thanks for the update, please fix the issue with the [b0rken icon](https://github.com/mozilla/pontoon/pull/1462#issuecomment-553072701):
```diff
 .entity.batch-editable .status:hover:before {
     content: "";
+    font-weight: bold;
 }
```

Also, when you click on the string to select it, the checkbox should not be selected.

mathjazz:
@monikamaheshwari I noticed the tick icon no longer appears when you hover Translation Author fitlers - only if you hover the avatar. Could you please fix that?

mathjazz:
Hey @monikamaheshwari, let us know if you need any help here.

monikamaheshwari:
Fixed the issue. Please check @mathjazz

mathjazz:
Thanks for the quick update!

Please also make sure that the icon changes to the selected checkbox when hovering selected (not checked) string in the string list, like this:
Screenshot 2019-11-12 at 20 18 54

mathjazz:
Thanks! One thing I didn't notice in your previous update - when hovering translation authors, we should just show the tick (instead of a squared tick) within the avatar, because the avatar itself already acts as a square.

mathjazz:
Thanks for the update! While it fixes the issue, it sadly introduces a new one - hovering other filter icons now shows a squared tick (instead of just a tick).

mathjazz:
@monikamaheshwari We'd like to move forward here and I figured it's easier if I commit fixes directly.

I've deployed the patch to stage to let users say how they feel about the change before we ship it:

mathjazz:
We've got some feedback:
* Lan figured we show batch selection hints even when entities are not batch editable (2310f70).
* Artem thinks there are two many different icons. He suggests we change color instead of using the tick icon when hovering the checkbox.
* Michal also thinks there are two many different icons. He comes with an alternative proposal which interestingly also drops the tick icon. He also notes that the proposal might not work on other pages like dashboards.
photo6041865611115606598

Using status icons as checkboxes is a pretty unusual UX, so it'll be hard to come up with a solution that is "very good" without making more radical changes. Personally I see the current implementation as "not bad" and this patch as a slight improvement at best.

mathjazz:
Closing due to inactivity. We need to address open questions before continuing.

PR #1468, opened 2019-11-04
Bug 1593747 - [sync] Pontoon incorrectly considers .pot files to be .…

TheoChevalier:
…po in some cases

### DIFF
--- a/pontoon/sync/vcs/models.py
+++ b/pontoon/sync/vcs/models.py
@@ -563,11 +563,12 @@ class VCSConfiguration(object):
             ProjectFiles(locale_code, [self.parsed_configuration]),
         )
 
-    def l10n_path(self, locale, reference_path):
+    def l10n_path(self, locale, absolute_resource_path):
         """
         Return l10n path for the given locale and reference path.
         """
         project_files = self.get_or_set_project_files(locale.code)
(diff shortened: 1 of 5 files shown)


mathjazz:
I've also made the similar change in `l10n_path` and added some tests.

@TheoChevalier Could you test if the patch works?

@Pike Do you have cycles for a review?

TheoChevalier:
@mathjazz It still works with the latest changes (pulled your changes, ran `make build`, `make run`, then synced one of my test projects that were not syncing properly)

pontoon/sync/vcs/models.py
```diff
@@ -563,11 +565,12 @@ def get_or_set_project_files(self, locale_code):
             ProjectFiles(locale_code, [self.parsed_configuration]),
         )
 
-    def l10n_path(self, locale, reference_path):
+    def l10n_path(self, locale, absolute_resource_path):
         """
         Return l10n path for the given locale and reference path.
         """
         project_files = self.get_or_set_project_files(locale.code)
+        reference_path = locale_to_source_path(absolute_resource_path)
```
Pike: `l10n_path` is only ever called with p-c on, so this should be here anymore, I think.
Or it should at least be guarded by an `if`.

PR #1470, opened 2019-11-05
Fix bug 1588471 - Remove `translate/` from the Translate.Next URLs.

adngdb:
This makes the old Translate page unaccessible, as the Translate.Next URL now replaces the old URL. It also ensures that previous Translate.Next URLs are correctly redirected.


pontoon/base/tests/views/test_translation.py
```diff
@@ -225,11 +228,13 @@ def test_view_translate_not_authed_public_project(
...
         resource=resource, locale=locale_a,
     )
-    response = client.get(
-        '/%s/%s/%s/'
-        % (locale_a.code, project.slug, resource.path)
-    )
-    assert response.status_code == 200
+
+    with override_flag('translate_next', active=True):
+        response = client.get(
```
adngdb: @mathjazz This test fails, and I don't understand why. There seems to be an infinite redirect here. Also, I don't understand what this test does. Can you perhaps shed some light here?

mathjazz: It's actually `Failed to establish a new connection: [Errno 111] Connection refused` error. The test will pass if you'd pass project_locale_a and resource_a, but then it's effectivelly the same as `test_translate_behind_flag` in `pontoon.translate.tests.test_views`. Let's remove it.

PR #1473, opened 2019-11-05
Fix bug 1593237 Replace String list spinner

sccofield:
**What does this PR do?**
This pull request fixes the issue [1593237](https://bugzilla.mozilla.org/show_bug.cgi?id=1593237). It replaces the string list spinner with a content loading placeholder (skeleton).

Screenshot 2019-11-05 at 11 26 07 PM


sccofield:
hi, @mathjazz I decided to use HTML and CSS to build the skeleton to make it more responsive.

frontend/src/core/loaders/components/SkeletonLoader.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+
+import './SkeletonLoader.css';
+
+const list = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
```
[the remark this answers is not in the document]
mathjazz: Ideally, we should fill the entire string list with placeholders. So I would increase the number to 30 and hide the scrollbar:
```css
.skeleton-loader {
    height: 100%;
    overflow: hidden;
}
```

frontend/src/core/loaders/components/SkeletonLoader.js
```diff
@@ -0,0 +1,24 @@
...
+
+const SkeletonLoader = () => (
+    <div className="skeleton-loader entities unselectable">
```
mathjazz: You can remove the `entities unselectable` classes.

frontend/src/core/loaders/components/SkeletonLoader.js
```diff
@@ -0,0 +1,24 @@
...
+        <ul>
+            { list.map((i) => {
+                const classes = `entity missing ${i ===1 ? 'selected' : null}`
```
mathjazz: Please add space after `===` and change 1 to 0 now that the `list` will start from 0.

frontend/src/core/loaders/components/SkeletonLoader.js
```diff
@@ -0,0 +1,24 @@
...
+                return <li className={ classes }>
+                    <span className='status fa'/>
+                    <div>
+                        <p className='source-string'></p>
+                        <p className='text-2'></p>
+                    </div>
```
mathjazz: Could you also add skeleton for the translations?

frontend/src/core/loaders/components/SkeletonLoader.js
sccofield: Hi, @mathjazz I noticed that when I removed the entities class, the list does not load, it just stops on the skeleton page. 
I can't really figure out why it's behaving like that.

sccofield:
> Once that's done we'll also need to look into the editor and figure out if we should add some placeholders there as well.

It will look nice to add some placeholders where the editor is, I will work on a design on it first to show you what it will look like. Though I noticed that the editor is empty when there is no strings to translate. 
Screenshot 2019-11-05 at 11 26 07 PM

sccofield:
> Once that's done we'll also need to look into the editor and figure out if we should add some placeholders there as well.

It will look nice to add some placeholders where the editor is, I will work on a design on it first to show you what it will look like. Though I noticed the editor is not visible when they are no strings to translate.

Screenshot 2019-11-07 at 1 03 02 AM

frontend/src/core/loaders/components/SkeletonLoader.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+
+import './SkeletonLoader.css';
+
+const list = [...Array(30).keys()];
+
+const SkeletonLoader = () => (
+    <div className='skeleton-loader entities'>
```
adngdb: Why the `<div>` here? Why not return the `<ul>` directly?

frontend/src/core/loaders/components/SkeletonLoader.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+
+import './SkeletonLoader.css';
+
+const list = [...Array(30).keys()];
+
+const SkeletonLoader = () => (
```
adngdb: To stay consistent with the rest of the app, please use the `function` syntax here. That way you can write…

```js
export default function SkeletonLoader() {
```

… and avoid the export on a different line.

frontend/src/core/loaders/components/SkeletonLoader.css
```diff
@@ -0,0 +1,33 @@
+.skeleton-loader p {
+    height: 0.875rem;
+    background-color: #E2E2E2;
+    animation: fading 1.5s infinite;
+    overflow: hidden;
```
mathjazz: Not sure this is needed here. But you should add a new rule on top to prevent scrollbar on the loader:

```
.skeleton-loader.entities {
    height: 100%;
    overflow: hidden;
}
```

sccofield:
> I think we'll also need to use a different loader (None? Spinner?) when you scroll down the list. The current experience with the skeleton is not so great for that use case.

Hi, @mathjazz I need more clarifications here on where we need a different loader.

mathjazz:
> Hi, @mathjazz I need more clarifications here on where we need a different loader.

Sure: when you scroll down the string list, more strings are loaded. There we shouldn't show the loader.

sccofield:
Hi @mathjazz looking at the code, I discovered that we used `react-infinite-scroller` to load the list. This package accepts just one loader so it will be difficult to use a separate loader when scrolling down the list. I was also thinking if there is any way we can edit the present skeleton to improve the users' experience or use a different type of loader all together.

mathjazz:
I see. Let's keep the skeleton loader then.

I'd suggest to make two changes though:

1. Let's just have two lines - one for source string and one for translation - both with the same thickness.

1. If you scroll to the very bottom of the list, so that you only see the skeleton, strings keep loading in batches, one after another. We should only load one batch of strings and move the scrollbar up so that the first newly loaded string is shown on top of the list.

mathjazz:
Thanks for the update! Could you also unify the line thickness and use different widths? I like how the skeleton looked originally: https://github.com/mozilla/pontoon/pull/1473#issue-337048376.

--

To test string loading, I suggest you [set up a test project](https://mozilla-pontoon.readthedocs.io/en/latest/user/localizing-your-projects.html#adding-a-new-project-to-pontoon). Go to http://localhost:8000/admin/, click "Add new project" and then fill out the following fields:
* Name: Test
* Slug: (autocompletes)
* Locales: Select Slovenian (sl)
* Repository URL: https://github.com/mathjazz/pontoon-test.git
* Download prefix: .
* Public Repository Website: (autocompletes)

Click Save Project and once the page reloads, click the red Sync button at the bottom. After a few seconds, you should be able to access http://localhost:8000/projects/test/ and then get to the translate view. You should now have multiple resources in the menu. Let me know how it goes. :)

--

I haven't checked, but you might be right about the solution for the infinite string loading issue - increasing the threshold in the React Infinite Scroller. The problem could also be that the loader now occupies the entire string list height.

sccofield:
hi @mathjazz Thanks, I was able to add a new project with your guide. 

I have also worked on the requested changes.

mathjazz:
Nice work! The skeleton now looks exactly as it should.

I think we still have a UX problem when loading more strings on scroll. If you scroll to the bottom of the list, entities keep loading and you don't even realize that, because the loader is static.

The current loader is smaller in height and disappears when the new batch of entities load. We should do the same with the skeleton loader. Maybe we can only occupy the lower 20-30% of the string list with the skeleton loader when requesting more strings on scroll?

sccofield:
Hi, @mathjazz I finally figured a way to use another loader, when loading more strings after the first load. I connected the component to the redux store to know if they are strings already and conditionally render the spinner loader or the skeleton.

In this case, I think I might need to change the name of the component from SkeletonLoader to maybe something like "EntityLoader".

Also after connecting the component to the redux store, I started having issues with the test suite. I have not been able to figure a way to test the component after connecting it to the redux store. It seems the shallow.() does not work with connected components. 

I will need some advice on how to proceed in this scenerio.

mathjazz:
Great work @sccofield! In this case I'd suggest to just use the skeleton loader even on scroll, but let's only show one skeleton entity (maybe two?). Note that you'll also need to set `.skeleton-loader.entities` `height` to `auto` when scrolling and to `100%` in other cases.

That might also fix the tests.

I guess you can also revert the `InfiniteScroll` treshod now.

sccofield:
Hi @mathjazz Thanks for the suggestion, The skeleton looks much better now when scrolling. I also passed the entities as props to the skeleton so that fix the tests.

frontend/src/core/loaders/components/SkeletonLoader.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+
+import './SkeletonLoader.css';
+
+let list;
+
+export default function SkeletonLoader (props) {
+    { list = props.entities.length === 0 ? [...Array(30).keys()] : [...Array(2).keys()] }
```
mathjazz: I would slice this into a few variables, which will be used later on:

```js
const firstLoad = props.entities.length === 0;
const entityCount = firstLoad ? 30 : 2;
const list = [...Array(entityCount).keys()];
```

You can also drop the `{}` and `let list` above;

frontend/src/core/loaders/components/SkeletonLoader.js
```diff
@@ -0,0 +1,21 @@
...
+    return <ul className={`skeleton-loader entities ${props.entities.length === 0 ? 'entities' : 'loading'}`}>
```
mathjazz: You are appending the `entities` class twice. It should be `null : 'scroll'`.

Also: spaces around content within `{}` and `${}`.

frontend/src/core/loaders/components/SkeletonLoader.js
```diff
@@ -0,0 +1,21 @@
...
+        { list.map((i) => {
+            const classes = `entity missing ${i === 0 ? 'selected' : null}`
```
mathjazz: We should only mark selected entity on `firstLoad`.

frontend/src/core/loaders/components/SkeletonLoader.css
```diff
@@ -0,0 +1,37 @@
+.skeleton-loader.entities {
+    pointer-events: none;
+    overflow: hidden;
+    height: 100%;
+}
+
+.skeleton-loader.loading {
+    pointer-events: none;
+    overflow: hidden;
```
mathjazz: You don't need to repeat these two properties here. You should change the selector to `.skeleton-loader.entities.loading`.

sccofield:
> Nice work! The only weird thing is that after applying several filters (which reload the string list), the animation seems to trigger at different times for different entities in the skeleton loader.

Hi, @mathjazz I have worked on the other corrections. I need more clarification on how to reproduce the animation triggering at different times.

sccofield:
> Thanks for the update! Adding a video to explain the issue with animation:
> https://drive.google.com/open?id=16DrjsoxoEK9A_G1MCvEAwAW__NrL-K-1

Hi, @mathjazz I just requested access for the video.

mathjazz:
You should have access now.

sccofield:
hi, @mathjazz I added some code to improve the skeleton animation. It looks better now.

adngdb:
I like this very much!

PR #1474, opened 2019-11-06
Fix bug 1380356: Add ability to use localization as source lang for translation

abowler2:
Issue: Some localization communities have a hard time to directly translate from English, and prefer to do their work based on an existing translation.

This fix is working to add a Preferred Source Locale selection to allow users to select existing translations to work from.

____
Current status: I have added the new field to the UserProfile model and added the new setting to the UI of the Settings page.  

However, the form is not functioning properly. It is currently rendering the correct selections in the menu, but any change made effects both the `Custom Homepage` selector and the `Preferred source locale` selector. Selections are not being saved in the model field for `preferred_source_locale` (the custom_homepage setting still saves when changes are made with the custom homepage selector).

If you could take a look at my current progress and help me to find my missteps or where I need to go from here that would be appreciated.


mathjazz:
Nice work so far!

The settings page will need a slight redesign, but for now I suggest we move the Preferred source locale selector under the Select your homepage selector - it's a similar UX and it's hard to predict which of the features will be used more often.

I would also change the label to "Select preferred source locale" and add the default value to the selector: "Default project locale".

You'll need to update the `settings.js` file to make sure it properly handles selectors. Similarly, you'll need to update `settings.css` to make sure styling of both selectors is consistent.

abowler2:
I made progress and am now able to save the preferred source locale. However, I have not been able to track down why when a selection is made it changes the selection for both items. This is only on the UI as the user's selections are being properly saved in the models. I tried giving unique class names to the elements, but that did not alter the behavior. 

If you have a moment to take a look and offer suggestions on this I would appreciate it. 

In the meantime, I'll continue to work on it as well as move forward with the next step of using this setting to determine which strings to render in the translation panel.

pontoon/contributors/static/css/settings.css
```diff
@@ -1,25 +1,35 @@
-#homepage {
+#homepage,
+#preferred-locale {
   margin-top: 30px;
 }
 
-#homepage .label {
+#homepage .label,
+#preferred-locale .label {
```
mathjazz: What if you simply use `#heading X` instead of `#homepage X, #preferred-locale` here and below?

abowler2:
> Nice work!
> 
>     1. To fix the problem with selectors, please look into the `pontoon/teams/static/js/team_selector.js` file (sorry for misleading you), which handles the actual menu selector. You'll need to make sure that the element we update is withing the selector by prepending something like `$(this).parents(".locale-selector")...`.

Thank you! 

> 
>     2. When I select a source locale and save it, the /settings page errors after a reload:
>        `'None' has no attribute 'serialize'`

I just realized this myself  😞  

> 
>     3. There's no way to select "Default Project Locale" once you select another locale. Unlike what I said initially when you were creating a menu, the behaviour (and the model) of the preferred locale selector should be exactly the same as for the homepage selector.

OK, I originally had appended it but then since it is a Foreignkey was pretty sure it would cause an issue.  Should I change the model for `preferred_source_locale` to CharField and then mimic the behavior of the Homepage selector then?
If so, do you want me to revert the last migration or just migrate the changes?

mathjazz:
> OK, I originally had appended it but then since it is a Foreignkey was pretty sure it would cause an issue. Should I change the model for `preferred_source_locale` to CharField and then mimic the behavior of the Homepage selector then?

Sadly, yes.

> If so, do you want me to revert the last migration or just migrate the changes?

I'll leave that to you. If possible, remove the migration and add a new one, and before you do so, make sure you unapply the current migration.

abowler2:
All seems to be working as expected now on the first parts of this task.  Please let me know if there are any changes or adjustments needed.

Working on the next step now, or after I fix that failing test anyway 😃

abowler2:
So, I actually don't know why that test failed.  All tests pass locally, and I'm not entirely sure what the error message is......

adngdb:
@abowler2 It seems it was a random artifact from travis, or pytest, or anything. @mathjazz restarted the build and it's now passing. :)

abowler2:
My attempt to access the data in the preferred_source_locale field on the frontend has not been successful (it comes back as  `undefined`). I would appreciate it if you could take a look at what I have and advise on whether I'm on the right track or not and if so, where I need to make changes in order to get that data. 

Of course, if I'm completely off track then if you could please provide information on how to go about accessing the data I would appreciate it.

frontend/src/core/user/reducer.js
```diff
@@ -83,6 +83,7 @@ export type UserState = {|
     +translatorForProjects: { [string]: boolean },
     +settings: SettingsState,
     +preferredLocales: Array<string>,
+    +userPreferredLocale: string,
```
mathjazz: `preferredSourceLocale`

pontoon/contributors/views.py
```diff
@@ -168,6 +168,14 @@ def save_custom_preferred_source_locale(request):
     return HttpResponse('ok')
 
 
+def user_preferred_source_locale(request):
+    preferred_locale = request.user.profile.preferred_source_locale
+
+    return JsonResponse({
+        'preferred-locale': preferred_locale,
```
mathjazz: `'preferred-source-locale': preferred_source_locale,`

pontoon/contributors/templates/contributors/settings.html
```diff
@@ -31,10 +31,17 @@ <h2 id="username">{{ user.first_name }}</h2>
...
-    <span class="label">Select your homepage</span>
-    {% include 'teams/widgets/team_selector.html' %}
+  <div id="heading" class="clearfix">
+    <div id="homepage">
+      <span class="label">Select your homepage</span>
+      <span class="homepage">{% include 'teams/widgets/team_selector.html' %}</span>
+    </div>
+    <div id="preferred-locale">
+      <span class="label preferred-locale">Select preferred source locale</span>
+      <span class="preferred-locale">{% include 'teams/widgets/preferred_locale_selector.html' %}</span>
```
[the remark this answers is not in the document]
abowler2: Within `pontoon/contributors/views.py` there are 2 separate lists created with different default values for `custom_homepage` and `preferred_locale` as well as separate custom values saved.  This widget renders using the values from the `preferred_locales` whereas `teams/widgets/team_selector.html` renders based on the `custom_homepage` values.

pontoon/teams/static/js/team_selector.js
```diff
@@ -1,5 +1,10 @@
 $(function() {
   $('.locale-selector .locale .menu li:not(".no-match")').click(function () {
-    $('.locale .selector').html($(this).html());
+    if ($(this).closest('#homepage').length == 1) {
+      $('#homepage .locale .selector').html($(this).html());
+    }
+    if ($(this).closest('#preferred-locale').length == 1) {
+      $('#preferred-locale .locale .selector').html($(this).html());
+    }
```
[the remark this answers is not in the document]
abowler2: Ahhhh....I tried many a combination with .parents() and .find() with no luck....clearly I did not use this combo 😄

abowler2:
> In [b67fd68](https://github.com/mozilla/pontoon/commit/b67fd68ea78beb664be1ef1b5afd499aad4ad12d) I don't see `getUserPreferredLocale` ever being called. 

I had been using a console.log() in the Translation.js file to show all the values in the `user` object, as it had `user` imported, in order to check if the value was being accessed.  It would show the variable but then have its value as `undefined` 

However, as you stated that I may not even need to use the call in the translate app, it might not matter.

pontoon/contributors/templates/contributors/settings.html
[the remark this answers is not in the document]
mathjazz: I see what the problem is - the "locale" variable names are different.

Now that we're [removing](https://github.com/mozilla/pontoon/pull/1486/files#diff-0f32e96ffcaa02bd0293a2c31c056d24) `translate.html`, I suggest you simplify `teams/widgets/team_selector.html` by replacing it with the contents of `teams/widgets/preferred_locale_selector.html`. You can also remove the `if project and preferred_locale and preferred_locale == l` block.

Then you'll need to turn it into a [macro](https://jinja.palletsprojects.com/en/2.10.x/templates/#macros) and pass the right variable to it. See `multiple_team_selector.html` for example.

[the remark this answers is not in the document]
abowler2: Just pushed those changes.

pontoon/contributors/templates/contributors/settings.html
```diff
@@ -31,14 +31,14 @@ <h2 id="username">{{ user.first_name }}</h2>
...
+  <div id="user-preference" class="clearfix">
     <div id="homepage">
       <span class="label">Select your homepage</span>
       <span class="homepage">{% include 'teams/widgets/team_selector.html' %}</span>
     </div>
     <div id="preferred-locale">
-      <span class="label preferred-locale">Select preferred source locale</span>
-      <span class="preferred-locale">{% include 'teams/widgets/preferred_locale_selector.html' %}</span>
+      <span class="label">Select preferred source locale</span>
+      <span class='preferred-locale'>{% include 'teams/widgets/preferred_locale_selector.html' %}</span>
```
mathjazz: Nit: please use double quotes.

pontoon/contributors/views.py
```diff
@@ -193,11 +220,26 @@ def settings(request):
...
         custom_homepage_locale = default_homepage_locale
 
+    default_preferred_source_locale = Locale(name='Default project locale', code='')
+    preferred_locales = list(Locale.objects.all())
+    preferred_locales.insert(0, default_preferred_source_locale)
+
+    # Set preferred source locale
+    preferred_source_locale = request.user.profile.preferred_source_locale
+    if preferred_source_locale:
+        custom_preferred_source_locale = (
```
mathjazz: Nit: I would just call this `preferred_source_locale`.

And above, I'd rename `custom_homepage` to `custom_homepage_locale`.

pontoon/contributors/templates/contributors/settings.html
abowler2: I actually removed that class in the refactor of team selector (https://github.com/mozilla/pontoon/pull/1474/commits/c9b672fe23836c08aaec3c0cbffd12161946f880) as it was no longer needed but will note the need for double quotes on classes.

pontoon/contributors/views.py
mathjazz: Whoops, false alarm. Sorry!

pontoon/contributors/templates/contributors/settings.html
mathjazz: Whoops, false alarm. Sorry!

abowler2:
Pushed my progress so far. Wanted to touch base on a couple of things:

1) I'm not quite sure how to access the plural strings when displaying the preferred source locale as the Translation model `plural_form` field is a small integer indicating the Index of `Locale.cldr_plurals_list()` (per the comment in the code).

2) There is a bug that I can't track down as I have not yet been able to pinpoint the trigger. It occurs at times when loading entities or when selecting an entity but I'm not sure why as it currently seems somewhat random.  Maybe it has to do with the plural forms since that piece is not complete????  At any rate, when it happens the screen goes blank with the following error: 
Screen Shot 2019-11-17 at 10 49 10 AM


It also appears that I have broken some tests 😄   Will work on fixing those once this piece is working properly.

pontoon/base/models.py
```diff
@@ -2588,7 +2591,22 @@ def map_entities(cls, locale, entities, visible_entities=None):
...
+                preferredSourceLocale = Locale.objects.filter(
+                    code=preferred_source_locale).values_list('pk')[0][0]
+                string = Translation.objects.filter(locale=preferredSourceLocale, entity=entity)
+                if not string:
+                    string = entity.string
+                    string_plural = entity.string_plural
+                else:
+                    string = string.values_list('string')[0]
+                    # replace with Source Locale Translated Plural String
+                    string_plural = entity.string_plural
```
mathjazz: You should remove this entire if-else block and only keep default string values:
```python
original = entity.string
original_plural = entity.string_plural
```

pontoon/base/models.py
```diff
@@ -2588,7 +2591,22 @@ def map_entities(cls, locale, entities, visible_entities=None):
...
+
+            if string_plural == "":
```
mathjazz: Inside this if-block, you should set the source string to the alternative source locale if available:
```python
if preferred_source_locale != '' and entity.alternative_originals:
    string = entity.alternative_originals[0].string
```

Note that this will only work if you above the `entities` for loop do:

```python
if preferred_source_locale != '':
    entities = entities.prefetch_alternative_originals(preferred_source_locale)
```

With contents of the `prefetch_alternative_originals()` being:

```python
def prefetch_alternative_originals(self, code):
    return self.prefetch_related(
        Prefetch(
            'translation_set',
            queryset=(
                Translation.objects.filter(locale__code=code, approved=True)
            ),
            to_attr='alternative_originals'
        )
    )
```

pontoon/base/views.py
```diff
@@ -999,6 +1005,7 @@ def user_data(request):
             'force_suggestions': user.profile.force_suggestions,
         },
         'preferred_locales': user.profile.sorted_locales_codes,
+        'preferred_source_locale': user.profile.preferred_source_locale,
```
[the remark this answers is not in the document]
abowler2: No, we don't need that. That was from when I was trying to access the preferred_source_locale from the frontend.  I removed it.

pontoon/base/models.py
```diff
@@ -2585,12 +2602,19 @@ def map_entities(cls, locale, entities, visible_entities=None):
...
             translation_array = []
+            string = entity.string
+            string_plural = entity.string_plural
 
-            if entity.string_plural == "":
+            if string_plural == "":
                 translation = entity.get_active_translation().serialize()
                 translation_array.append(translation)
+                if preferred_source_locale != '' and entity.alternative_originals:
+                    string = entity.alternative_originals[0].string
```
mathjazz: Now to handle plurals, we should actually move this block below the `if string_plural == '' - else` block and then use the last available plural form for the pluralized string:
```python
if string_plural != '':
    string_plural = entity.alternative_originals[-1].string
```

abowler2: This works for all the files I have in my Test project except the `amo.po` file. It works when only changing `original` to the `alternative_original` but the plural form causes an error with the JSON. ("The response content is not JSON-compatible")

When I try to simulate the query in the shell I get an `AssertionError: Negative indexing is not supported.`

mathjazz: Hm, I'm not able to reproduce your issues. 

What do you mean by "changing original to the alternative_original"?

I've made the following change:
```diff
diff --git a/pontoon/base/models.py b/pontoon/base/models.py
index 105a600a7..820741e54 100644
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -2621,6 +2621,11 @@ class Entity(DirtyFieldsMixin, models.Model):
                     translation = entity.get_active_translation(plural_form).serialize()
                     translation_array.append(translation)
 
+            if preferred_source_locale != '' and entity.alternative_originals:
+                string = entity.alternative_originals[0].string
+                if string_plural != '':
+                    string_plural = entity.alternative_originals[-1].string
+
             entities_array.append({
                 'pk': entity.pk,
                 'original': string,
```

abowler2: Ahh, OK.  I had them set up as stand-alone `if` statements instead of nesting the plurals `if` statement.  It is working fine now ~ Thank you.

abowler2:
Made requested updates and fixed tests.

Working on how to set Machinery to use en-US original now. Let me know if there is anything in particular I need to know to complete this.

mathjazz:
Very nice work, thanks!

To fix Machinery, I think you'll need to replace `selectedEntity.original` in `fetchHelpersData` with `selectedEntity.machineryOriginal`, which you'll need to pass from `Entity.map_entities()`.

abowler2:
That seems to have worked!  Thanks

Let me know if anything further is needed.

abowler2:
I'll definitely work on showing the original source in the Locals tab, and let you know if I have questions.

abowler2:
> A nice addition would be to also show the original source (i.e. `machinery_original`) in the Locales tab. Could you try adding that as well?

I pushed what I have so far for this....but I just want to say right off that while it is working...it's pretty ugly 🤣 
In order to refactor it I needed some info (and I also wanted to be sure I'm even on the right track with it).

1. There is no ForeignKey between `Locale` and `Entity`, and I couldn't find a common field. Also, the entity does not exist in Translation so making use of the existing code does not work. Given this, I hard coded the `pk` in order to access the properties for the `original source`. If there is a way for this to be dynamic, or if there is another way to go about this, please let me know.

2. For the `original source` to display the same as the other translations, I had to rename the `keys` in the `dict`, and I'm not sure if there is a more efficient way to do this.

3. I was unsure how to add the `original source` in a way that would allow it to be sorted within the other translations, so I prepended it to the beginning.

4. When the `Locale/code` link is clicked on the frontend it gives a 404-error as this was manually entered so it is not a valid link. How should this be handled?

I think that sums up the current issues.

TIA 😄

pontoon/base/forms.py
```diff
@@ -247,6 +247,23 @@ def __init__(self, *args, **kwargs):
...
+    """
+    Form is responsible for saving preferred source locale
+    """
+    class Meta:
+        model = UserProfile
+        fields = ('preferred_source_locale',)
+
+    def __init__(self, *args, **kwargs):
+        super(UserPreferredSourceLocaleForm, self).__init__(*args, **kwargs)
+        preferred_source_locales = list(Locale.objects.all().values_list('code', 'name'))
```
mathjazz: Nit: We should actually call this `all_locales`, same as above.

pontoon/base/models.py
```diff
@@ -2205,6 +2208,20 @@ def prefetch_active_translations(self, locale):
             )
         )
 
+    def prefetch_alternative_originals(self, code):
+        """
+        Prefetch active translations for given preferred source locale.
```
mathjazz: active -> approved

pontoon/base/views.py
```diff
@@ -208,15 +208,20 @@ def _get_all_entities(locale, project, form, entities):
     visible_entities = entities.values_list('pk', flat=True)
 
     return JsonResponse({
-        'entities': Entity.map_entities(locale, entities_to_map, visible_entities),
+        'entities': Entity.map_entities(
+            locale,
+            preferred_source_locale,
+            entities_to_map,
+            visible_entities
```
mathjazz: Nit: please add comma at the end of the line. That code style allows us to add/remove parameters by making 1-line diffs.

pontoon/contributors/views.py
```diff
@@ -149,6 +149,25 @@ def save_custom_homepage(request):
...
 
 
+@login_required(redirect_field_name='', login_url='/403')
+@require_POST
+@transaction.atomic
+def save_preferred_source_locale(request):
+    """Save preferred source locale """
+    form = forms.UserPreferredSourceLocaleForm(
+        request.POST,
+        instance=request.user.profile
```
mathjazz: Nit: comma at the end of the line.

mathjazz:
Bug filed:
https://bugzilla.mozilla.org/show_bug.cgi?id=1600134

You can also split it in two patches if you want (refactor + displaying source string).

PR #1478, opened 2019-11-07
Fix bug 1586820: 3-column layout

mathjazz:
The gist of the patch is in the first two commits.

The rest is a couple of related CSS improvements.

### DIFF
--- a/frontend/src/App.css
+++ b/frontend/src/App.css
@@ -1,6 +1,6 @@
 #app {
     display: grid;
-    grid-template-columns: 50% 50%;
+    grid-template-columns: 25% 75%;
     grid-template-rows: 60px auto;
     height: 100vh;
 }
--- a/frontend/src/modules/entitydetails/components/Helpers.js
+++ b/frontend/src/modules/entitydetails/components/Helpers.js
@@ -6,7 +6,6 @@ import { Localized } from '@fluent/react';
 
 import './Helpers.css';
 
-import { History } from 'modules/history';
 import { Machinery, MachineryCount } from 'modules/machinery';
 import { OtherLocales, OtherLocalesCount } from 'modules/otherlocales';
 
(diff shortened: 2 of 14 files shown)


frontend/src/modules/entitydetails/components/Helpers.js
```diff
@@ -47,36 +41,23 @@ export default class Helpers extends React.Component<Props> {
...
         } = this.props;
 
-        const historyCount = history.translations.length;
         const machineryCount = machinery.translations.length;
         const otherlocalesCount = otherlocales.translations.length;
 
         return <Tabs>
             <TabList>
-                <Tab>
-                    <Localized id='entitydetails-Helpers--history'>
```
adngdb: This can be removed from our various .ftl files.

frontend/src/App.css
```diff
@@ -1,6 +1,6 @@
 #app {
     display: grid;
-    grid-template-columns: 50% 50%;
+    grid-template-columns: 25% 75%;
```
adngdb: Wouldn't it be better to split the `EntityDetails` component into 2 parts, one with metadata and history, the other with the helpers, and make this a grid of `25% 50% 25%`? `Helpers` would become a container module (connected to the store), which might also make the code a bit simpler.

This is absolutely a question btw, if you think it's not a good idea or is too hard, feel free to say so. :)

mathjazz: I was thinking about that and decided for the current implementation, because it made switching between the single and batch editor slightly easier.

frontend/src/modules/entitydetails/components/Helpers.js
mathjazz: Good point!

frontend/src/App.css
adngdb: Good point!

adngdb: Good point!

mathjazz:
> I'm amazed by how simple this PR is. I expected something a lot more hairy. I'm thinking this is thanks to our work on Translate.Next, and that makes me happy. 🙂

Indeed, I was surprised myself that it turned out to be a one-day task. In fact, I implemented it while I was waiting for stage to update to the newest PostgreSQL. :)

PR #1480, opened 2019-11-07
No bug - Disallow failures for Python 3 tests.

adngdb:
From now on, all changes will have to ensure they do not break our Python 3 support!


adngdb:
Thanks for the reviews @jotes!

PR #1482, opened 2019-11-11
Fix bug #1594177: Indicate visually when Machinery has found some TM …

Karskaya:
…matches

Fix name of translation source translation-memory to count such translations as preferredCount and shows their amount in green number


Karskaya:
Hello @mathjazz! I hope I get the problem right, because this fix is suspiciously easy.

Karskaya:
@mathjazz Thank you!

PR #1487, opened 2019-11-12
Fix bug 1552802: Add support for pretranslation in admin panel.

VishalCR7:
We are looking to trigger the pretranslation in 3 cases:

- [x] Pretranslate now: starts the async job immediately.
- [x] Pretranslation enabled: execute the job as part of sync when new strings arrive.
- [x] Pretranslation enabled: also execute the job when new locales are added to the project.

I would like to know your views about the UI on the admin page and check if the all the cases work fine.  Thanks (:

### DIFF
--- a/pontoon/administration/views.py
+++ b/pontoon/administration/views.py
@@ -537,7 +537,7 @@ def manually_sync_project(request, slug):
 def manually_pretranslate_project(request, slug):
     if not request.user.has_perm('base.can_manage_project') or not settings.MANUAL_SYNC:
         return HttpResponseForbidden(
-            "Forbidden: You don't have permission for syncing projects"
+            "Forbidden: You don't have permission for pretranslating projects"
         )
 
     project = Project.objects.get(slug=slug)
--- a/pontoon/administration/views.py
+++ b/pontoon/administration/views.py
@@ -36,6 +36,7 @@ from pontoon.base.models import (
 )
 from pontoon.sync.models import SyncLog
 from pontoon.sync.tasks import sync_project
+from pontoon.pretranslation.tasks import pretranslate
 
 
 log = logging.getLogger(__name__)
@@ -527,5 +528,19 @@ def manually_sync_project(request, slug):
     sync_log = SyncLog.objects.create(start_time=timezone.now())
     project = Project.objects.get(slug=slug)
     sync_project.delay(project.pk, sync_log.pk)
+
+    return HttpResponse('ok')
+
+
+@login_required(redirect_field_name='', login_url='/403')
+@require_AJAX
+def manually_pretranslate_project(request, slug):
+    if not request.user.has_perm('base.can_manage_project') or not settings.MANUAL_SYNC:
+        return HttpResponseForbidden(
+            "Forbidden: You don't have permission for syncing projects"
+        )
+
+    project = Project.objects.get(slug=slug)
+    pretranslate(project)
 
     return HttpResponse('ok')
--- a/pontoon/administration/views.py
+++ b/pontoon/administration/views.py
@@ -535,7 +535,7 @@ def manually_sync_project(request, slug):
 @login_required(redirect_field_name='', login_url='/403')
 @require_AJAX
 def manually_pretranslate_project(request, slug):
-    if not request.user.has_perm('base.can_manage_project') or not settings.MANUAL_SYNC:
+    if not request.user.has_perm('base.can_manage_project'):
         return HttpResponseForbidden(
             "Forbidden: You don't have permission for pretranslating projects"
         )
--- a/pontoon/pretranslation/tasks.py
+++ b/pontoon/pretranslation/tasks.py
@@ -25,15 +25,17 @@ def pretranslate(project, locales=None, entities=None):
 
     log.info('Fetching pretranslations for project {} started'.format(project.name))
 
-    if not locales:
-        locales = project.locales.filter(
-            project_locale__readonly=False,
-        ).prefetch_project_locale(project)
+    if locales:
+        locales = project.locales.filter(pk__in=locales)
     else:
-        locales = project.locales.filter(
-            project_locale__readonly=False,
-            pk__in=locales,
-        ).prefetch_project_locale(project)
+        locales = project.locales
+
+    locales = (
+        locales
+        .filter(project_locale__readonly=False)
+        .distinct()
+        .prefetch_project_locale(project)
+    )
 
     if not entities:
         entities = Entity.objects.filter(
--- a/pontoon/pretranslation/tasks.py
+++ b/pontoon/pretranslation/tasks.py
@@ -32,7 +32,7 @@ def pretranslate(project, locales=None, entities=None):
     else:
         locales = project.locales.filter(
             project_locale__readonly=False,
-            pk__in = locales,
+            pk__in=locales,
         ).prefetch_project_locale(project)
 
     if not entities:
--- a/pontoon/pretranslation/tasks.py
+++ b/pontoon/pretranslation/tasks.py
@@ -29,6 +29,11 @@ def pretranslate(project, locales=None, entities=None):
         locales = project.locales.filter(
             project_locale__readonly=False,
         ).prefetch_project_locale(project)
+    else:
+        locales = project.locales.filter(
+            project_locale__readonly=False,
+            pk__in = locales,
+        ).prefetch_project_locale(project)
 
     if not entities:
         entities = Entity.objects.filter(
--- a/pontoon/sync/changeset.py
+++ b/pontoon/sync/changeset.py
@@ -236,7 +236,7 @@ class ChangeSet(object):
                     ))
 
         self.send_notifications(new_entities)
-        self.new_entities = self.new_entities + new_entities
+        self.new_entities = new_entities
 
     def update_entity_translations_from_vcs(
             self, db_entity, locale_code, vcs_translation,
--- a/pontoon/sync/changeset.py
+++ b/pontoon/sync/changeset.py
@@ -54,6 +54,7 @@ class ChangeSet(object):
         self.entities_to_update = []
         self.translations_to_update = {}
         self.translations_to_create = []
+        self.new_entities = []
         self.commit_authors_per_locale = defaultdict(list)
         self.locales_to_commit = set()
 
@@ -235,6 +236,7 @@ class ChangeSet(object):
                     ))
 
         self.send_notifications(new_entities)
+        self.new_entities = self.new_entities + new_entities
 
     def update_entity_translations_from_vcs(
             self, db_entity, locale_code, vcs_translation,
--- a/pontoon/sync/core.py
+++ b/pontoon/sync/core.py
@@ -27,14 +27,15 @@ log = logging.getLogger(__name__)
 
 def update_originals(db_project, now, full_scan=False):
     vcs_project = VCSProject(db_project, locales=[], full_scan=full_scan)
-
+    new_entities = []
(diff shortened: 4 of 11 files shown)


pontoon/sync/tasks.py
```diff
@@ -297,7 +311,11 @@ def sync_translations(
                 changeset = ChangeSet(db_project, vcs_project, now, locale)
                 update_translations(db_project, vcs_project, locale, changeset)
                 changeset.execute()
-                update_translated_resources(db_project, vcs_project, locale)
+
+                new_entities = new_entities + changeset.new_entities
```
mathjazz: `changeset.new_entities` should always be an empty list when syncing translations, so we should only collect new entities during the `sync_sources` task.

pontoon/sync/tasks.py
```diff
@@ -127,6 +128,15 @@ def sync_project(
         full_scan=force
     )
 
+    new_entities = new_entities.result
```
mathjazz: If `new_entities` is a list, it can't have a `result` property.

VishalCR7: Calling function.delay() returns an EagerResult object which contains the return value of function() in its `result` attribute (which will be a list that we need). That was something new which I learned while working on this. and to differentiate between return types I should have used different variable names and added comments to it :/ 

Either way, we will not be needing it based on your previous comments :)

pontoon/sync/core.py
```diff
@@ -173,48 +174,63 @@ def update_translations(db_project, vcs_project, locale, changeset):
...
         )
+    return flag
 
 
 def update_translated_resources_with_config(db_project, vcs_project, locale):
     """
     Create/update the TranslatedResource objects for each Resource instance
     that is enabled for the given locale through project configuration.
     """
+    flag = True
```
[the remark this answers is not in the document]
VishalCR7: oops!

pontoon/sync/tasks.py
mathjazz: Ah, thanks for clarifying!

pontoon/sync/changeset.py
```diff
@@ -235,6 +236,7 @@ def execute_create_db(self):
                     ))
 
         self.send_notifications(new_entities)
+        self.new_entities = self.new_entities + new_entities
```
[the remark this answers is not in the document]
VishalCR7: Apparently, I didn't go through the whole code to check if the function is invoked only a single time or multiple time but in both the cases, this works fine, thus saving a little time :P 

But I will go ahead and change it, now that I have a confirmation. Thanks :D

pontoon/sync/core.py
```diff
@@ -173,48 +172,49 @@ def update_translations(db_project, vcs_project, locale, changeset):
...
+
     for resource in vcs_project.configuration.locale_resources(locale):
         translatedresource, created = (
             TranslatedResource.objects.get_or_create(resource=resource, locale=locale)
         )
 
-        if created and resource.project.pretranslation_enabled:
-            flag = True
-
+        tr_created |= created  # set to true if at least one instance is created
```
mathjazz: This would be simpler to understand:
```pythonn
if created:
    tr_created = True
```

pontoon/sync/core.py
```diff
@@ -224,12 +224,11 @@ def update_translated_resources_without_config(db_project, vcs_project, locale):
                 translatedresource, created = (
                     TranslatedResource.objects.get_or_create(resource=resource, locale=locale)
                 )
-                if created and resource.project.pretranslation_enabled:
-                    flag = True
 
+                tr_created |= created  # set to true if at least one instance is created
```
mathjazz: Same as above.

pontoon/sync/tasks.py
```diff
@@ -128,13 +128,10 @@ def sync_project(
...
-    new_entities = new_entities.result
+    new_entities = source_changes.get('new_entities')
 
-    if source_changes.get('new_entities'):
-        new_entities = new_entities + source_changes.get('new_entities')
-
-    if db_project.pretranslation_enabled and len(new_entities):
-        new_entities = list(set(new_entities))  # take only unique entities
+    if db_project.pretranslation_enabled and new_entities:
+        new_entities = list(set(source_changes.get('new_entities')))
```
mathjazz: Can't you move this line above the `if` and remove line 131?

VishalCR7: For uniformity, we return None and the set() doesn't take non-iterables. So first we need to check the case of value not being none and then proceed.

pontoon/pretranslation/tasks.py
```diff
@@ -29,6 +29,11 @@ def pretranslate(project, locales=None, entities=None):
         locales = project.locales.filter(
             project_locale__readonly=False,
         ).prefetch_project_locale(project)
+    else:
+        locales = project.locales.filter(
+            project_locale__readonly=False,
+            pk__in=locales,
+        ).prefetch_project_locale(project)
```
mathjazz: You could repeat less code by doing this:
```python
if locales:
    locales = project.locales.filter(pk__in=locales)
else:
    locales = project.locales

locales = (
    locales
    .filter(project_locale__readonly=False)
    .prefetch_project_locale(project)
)
```

pontoon/sync/tasks.py
```diff
@@ -127,6 +128,12 @@ def sync_project(
         full_scan=force
     )
 
+    new_entities = source_changes.get('new_entities')
+
+    if db_project.pretranslation_enabled and new_entities:
+        new_entities = list(set(new_entities))
+        pretranslate(db_project, entities=new_entities)
```
mathjazz: We sometimes "migrate" translations, e.g. from one format to another (recently that happens often with Firefox, where we're replacing legacy file formats with Fluent). In practice that's done via a script and committed to VCS. On next sync, we then create new entities as part of `sync_sources()` and new translations as part of `sync_translations()`.

The logic above will pretranslate such entities, even though it shouldn't (because we'll create translations from VCS immediately after within `sync_translations()`. So we should delay execution of `pretranslate()` once all translations are created.

pontoon/administration/views.py
```diff
@@ -529,3 +530,17 @@ def manually_sync_project(request, slug):
     sync_project.delay(project.pk, sync_log.pk)
 
     return HttpResponse('ok')
+
+
+@login_required(redirect_field_name='', login_url='/403')
+@require_AJAX
+def manually_pretranslate_project(request, slug):
+    if not request.user.has_perm('base.can_manage_project') or not settings.MANUAL_SYNC:
```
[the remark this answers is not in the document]
VishalCR7: That part was copy-pasted from manual sync so it stayed. I will remove it. :sweat_smile:

pontoon/sync/tasks.py
```diff
@@ -125,15 +125,10 @@ def sync_project(
         locale=locale,
         no_pull=no_pull,
         no_commit=no_commit,
-        full_scan=force
+        full_scan=force,
+        new_entities=source_changes.get('new_entities')
```
mathjazz: Could you instead pass `source_changes.get('new_entities')` as a positional argument, consistently with other `source_changes` arguments?

pontoon/sync/tasks.py
```diff
@@ -279,6 +283,8 @@ def sync_translations(
     synced_locales = set()
     failed_locales = set()
 
+    new_locales = []  # store the locales for which new resources are added
```
mathjazz: Nit: Please change to:
```python
# Store newly added locales and locales with newly added resources
new_locales = [] 
```

pontoon/sync/tasks.py
```diff
@@ -389,3 +400,13 @@ def sync_translations(
...
+
+    if db_project.pretranslation_enabled:
+        # Pretranslate all entities for newly added locales
+        if len(new_locales):
+            pretranslate(db_project, locales=new_locales)
+
+        # Pretranslate newly added entities for all locales
+        if new_entities:
+            new_entities = list(set(new_entities))
+            pretranslate(db_project, entities=new_entities)
```
mathjazz: Could you please add a basic test to check when `pretranslate()` is called?

pontoon/sync/tasks.py
```diff
@@ -389,3 +400,13 @@ def sync_translations(
             locales=repo_locales[r.pk].exclude(code__in=failed_locales)
         )
     repo_sync_log.end()
+
+    if db_project.pretranslation_enabled:
+        # Pretranslate all entities for newly added locales
```
mathjazz: Nit: `...and locales with newly added resources`.

pontoon/sync/tasks.py
```diff
@@ -389,3 +400,13 @@ def sync_translations(
             locales=repo_locales[r.pk].exclude(code__in=failed_locales)
         )
     repo_sync_log.end()
+
+    if db_project.pretranslation_enabled:
```
mathjazz: If a new file is added, both conditions below will be true. So I suggest we exclude `new_locales` from the 2nd `pretranslate()` call.

VishalCR7: In the 2nd call as well, we will only loop through the entity locale pairs without actually pretranslating them ( as well have pretranslated them above). Although we can pass the excluded locales to make it bit fast it doesn't create any difference in the pretranslations created. What do you think??

mathjazz: I'm aware of that. I'd still avoid calling the method if it's not needed.

pontoon/sync/tests/test_tasks.py
```diff
@@ -259,6 +260,66 @@ def test_create_repository_log(self):
...
+        }]
+
+        with self.patch('pontoon.sync.tasks.update_translated_resources', return_value=False):
+            sync_translations(self.db_project.pk, self.project_sync_log.pk,
+                              self.now, [], [], [], ['new_entity'])
+
+        assert_true(self.mock_pretranslate.called)
+        assert_equal(self.mock_pretranslate.call_args[1]['entities'], ['new_entity'])
+
+    def test_new_resource_pretranslation(self):
```
mathjazz: `test_new_translated_resource_pretranslation`

pontoon/sync/tests/test_tasks.py
```diff
@@ -259,6 +260,66 @@ def test_create_repository_log(self):
...
+        """
+        Test if pretranslate is called for locales with newly added resource.
```
mathjazz: pretranslate -> pretranslation
resource -> TranslatedResource

pontoon/sync/tests/test_tasks.py
```diff
@@ -259,6 +260,66 @@ def test_create_repository_log(self):
...
+        self.mock_pull_changes.return_value = [True, {
+            self.repository.pk: Locale.objects.filter(pk=self.translated_locale.pk)
+        }]
+
+        with self.patch('pontoon.sync.tasks.update_translated_resources', return_value=False):
+            sync_translations(self.db_project.pk, self.project_sync_log.pk,
+                              self.now, [], [], [], ['new_entity'])
+
+        assert_true(self.mock_pretranslate.called)
+        assert_equal(self.mock_pretranslate.call_args[1]['entities'], ['new_entity'])
```
mathjazz: Can you also ensure the `locales` argument of `pretranslate()` is accurately set?

pontoon/sync/tasks.py
```diff
@@ -389,3 +401,16 @@ def sync_translations(
...
+        # Pretranslate all entities for newly added locales
+        # and locales with newly added resource
```
mathjazz: resource**s** :)

PR #1489, opened 2019-11-15
Bug 1527853 - Move review action views to a different module.

adngdb:


adngdb:
> I wonder if we should also move `Delete` action under review?

I want to move `update` and `delete` into the `translate` module in a follow-up. They are not technically "review" actions.

PR #1491, opened 2019-11-17
Fix bug 1597284 - Add support for authenticating with Google

anandthakker:
Bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1597284

So straightforward thanks to the groundwork in https://github.com/mozilla/pontoon/pull/1442 to generalize the socialapp setup h/t @abowler2

### DIFF
--- a/pontoon/base/templatetags/helpers.py
+++ b/pontoon/base/templatetags/helpers.py
@@ -263,8 +263,6 @@ def provider_login_url(request, provider_id=settings.AUTHENTICATION_METHOD, **qu
     provider = providers.registry.by_id(provider_id)
 
     auth_params = query.get('auth_params', None)
-    if provider_id != 'google':
-        query['scope'] = 'profile:uid profile:email profile:display_name'
     process = query.get('process', None)
 
     if auth_params == '':
--- a/pontoon/base/templatetags/helpers.py
+++ b/pontoon/base/templatetags/helpers.py
@@ -263,7 +263,8 @@ def provider_login_url(request, provider_id=settings.AUTHENTICATION_METHOD, **qu
     provider = providers.registry.by_id(provider_id)
 
     auth_params = query.get('auth_params', None)
-    query['scope'] = 'profile:uid profile:email profile:display_name'
+    if provider_id != 'google':
+        query['scope'] = 'profile:uid profile:email profile:display_name'
     process = query.get('process', None)
 
     if auth_params == '':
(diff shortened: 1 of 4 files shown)


abowler2:
Thanks, @anandthakker. 😊

pontoon/base/templatetags/helpers.py
```diff
@@ -263,7 +263,8 @@ def provider_login_url(request, provider_id=settings.AUTHENTICATION_METHOD, **qu
     provider = providers.registry.by_id(provider_id)
 
     auth_params = query.get('auth_params', None)
-    query['scope'] = 'profile:uid profile:email profile:display_name'
+    if provider_id != 'google':
+        query['scope'] = 'profile:uid profile:email profile:display_name'
```
mathjazz: I think [we can remove](https://github.com/mozilla/pontoon/blob/9353da04929858811867f60961dbc2765ffc1d40/pontoon/settings/base.py#L814) both of these lines completely. I tested locally and it works.

@jotes Could you confirm that?

pontoon/settings/base.py
```diff
@@ -825,6 +826,10 @@ def account_username(user):
     }
 }
 
+# Google Accounts
+GOOGLE_CLIENT_ID = os.environ.get('GOOGLE_CLIENT_ID', '')
+GOOGLE_SECRET_KEY = os.environ.get('GOOGLE_SECRET_KEY', '')
+
```
mathjazz: Nit: Please move this block under the `#GitHub` block (in front of `# All settings related to the AllAuth`).

pontoon/base/templatetags/helpers.py
jotes: I remember that some docs with suggestions to make the scope of OAUTH request explicit. However, the Jarek from future sees: https://github.com/pennersr/django-allauth/blob/e84a64cc51cd892ecab507f37b0e873be8d81b5d/allauth/socialaccount/providers/fxa/provider.py#L29 and says this line can be removed.

mathjazz: Thanks for confirming! @anandthakker So let's indeed remove this.

anandthakker:
@mathjazz pushed two commits to address your comments and filed a bug

mathjazz:
Thanks! Could you please also address the issue with `query['scope']`?

https://github.com/mozilla/pontoon/pull/1491/files#r347345771

anandthakker:
Oops!  I thought I'd committed that, but hadn't.

anandthakker:
@mathjazz removed.

PR #1494, opened 2019-11-18
No bug - Update PostgreSQL version to use on new Heroku deploys.

adngdb:


mathjazz:
I guess we need to use "11"? Right now deployment fails with:

"An error was encountered when contacting the add-on partner to create heroku-postgresql:hobby-dev: Available versions are 9.5, 9.6, 10, 11 (the default is 11)"

adngdb:
@mathjazz can you please try this again?

adngdb:
Thanks a lot for doing the testing!

PR #1495, opened 2019-11-19
Bug #1219431 - Add word count to Stats

Karska-dev:
Add methods get_or_create and bulk_update to EntityQuerySet for future changes related with word count
Bug [#1219431](https://bugzilla.mozilla.org/show_bug.cgi?id=1219431)


Karskaya:
@mathjazz, please review.
Made changes according to https://bugzilla.mozilla.org/show_bug.cgi?id=1219431#c14

pontoon/base/models.py
```diff
@@ -2245,6 +2246,12 @@ def reset_active_translations(self, locale):
 
         translations.filter(pk__in=unreviewed_pks).update(active=True)
 
+    def get_or_create(self, defaults=None, **kwargs):
+        return super(EntityQuerySet, self).get_or_create(defaults, **kwargs)
+
+    def bulk_update(self, objs, fields=None, batch_size=None):
```
jotes: Hey @Karskaya,
That's exactly what I've meant. The only thing that's missing for me is a small test to check if the method works as we expect.

Additionally, It could be a good idea to wrap this PendingDeprecationWarning and throw a warning if the version of Django is equal to or greater than 2.

What do you think about it? (cc: @mathjazz)

pontoon/base/models.py
```diff
@@ -2245,6 +2246,12 @@ def reset_active_translations(self, locale):
 
         translations.filter(pk__in=unreviewed_pks).update(active=True)
 
+    def get_or_create(self, defaults=None, **kwargs):
```
jotes: I'm not sure if you need to override this method. Do you have a use-case for this?

mathjazz: I'll guess we'll add logic to calculate and store word count in a followup PR?

pontoon/base/models.py
mathjazz: Sounds good to me!

Karskaya: Hello @jotes!
Could you please explain one more time this part with PendingDeprecationWarning?

Did you mean something like this?
```
def bulk_update(self, objs, fields=None, batch_size=None):
        if django.get_version() >= LooseVersion("2"):
            msg = "Warning message about Django version"
            warnings.warn(msg, PendingDeprecationWarning)
        return bulk_update(objs, fields, batch_size)
```

cc: @mathjazz

mathjazz: Something like that, year. Or just add a reference to [bug 1600344](https://github.com/mozilla/pontoon/blob/1c90890/pontoon/base/middleware.py#L23).

Karskaya:
Hello @mathjazz !
Sorry for holding this issue for 22 days already!
Need your help here. 

Q1. 
> Something like that, year. Or just add a reference to [bug 1600344](https://github.com/mozilla/pontoon/blob/1c90890/pontoon/base/middleware.py#L23).

Here is topic about Python version, not Django version. Does it make sense to add reference anyway?

Q2
This PendingDeprecationWarning is only warning and later code need to be updated or it's possible to try-catch QuerySet.bulk_update and if no luck to call bulk_update.helper function?

Q3
Is it ok to add two more imports for checking version? (import LooseVersion, import get_version) 

Q4
Function for counting words
`def get_word_count(string)`
What is the right place for it? helper.py?

Q5
I'm pretty sure that this part doesn't work:
```
for obj in objs:
            if hasattr(obj, 'string'):
                obj.word_count = get_word_count(obj.string)
                obj.save()
```
All I found about is add field to the class and run migration.
What was an idea here? I thought I got it, but I didn't. It looked easier at first glance. 

Q6 
overrided get_or_create function should make the same as bulk_update^ but only for one entity, right?

Q7
How all of this I will test? I will need to run a migration?

Q8 Tests.
Can I leave them for now and come back after all above will be clear?

Thank you!

cc: @jotes

mathjazz:
> Q1.
> 
> > Something like that, year. Or just add a reference to [bug 1600344](https://github.com/mozilla/pontoon/blob/1c90890/pontoon/base/middleware.py#L23).
> 
> Here is topic about Python version, not Django version. Does it make sense to add reference anyway?

Good point! No need for that reference.

> Q2
> This PendingDeprecationWarning is only warning and later code need to be updated or it's possible to try-catch QuerySet.bulk_update and if no luck to call bulk_update.helper function?

I think warning is good enough in this case.

> Q3
> Is it ok to add two more imports for checking version? (import LooseVersion, import get_version)

I'd just use:
```python
import django
if django.VERSION[0] > 1:
    # warn
```

> Q4
> Function for counting words
> `def get_word_count(string)`
> What is the right place for it? helper.py?

I'd just keep it in this file, but indent it so that it fits under `EntityQuerySet`. I'd also prepend the method name with `_` to denote that it's an internal function, same as `_repo_compat_attr`.

However, @jotes suggested this goes into a followup PR, so I'd remove it:
https://bugzilla.mozilla.org/show_bug.cgi?id=1219431#c14

> Q5
> I'm pretty sure that this part doesn't work:
> 
> ```
> for obj in objs:
>             if hasattr(obj, 'string'):
>                 obj.word_count = get_word_count(obj.string)
>                 obj.save()
> ```
> 
> All I found about is add field to the class and run migration.
> What was an idea here? I thought I got it, but I didn't. It looked easier at first glance.

Yes, you'd need to add a new `Entity.word_count` field. Also see inline comments.

And same as above - this should be a followup PR, so I'd remove it:
https://bugzilla.mozilla.org/show_bug.cgi?id=1219431#c14

> Q6
> overrided get_or_create function should make the same as bulk_update^ but only for one entity, right?

Correct.

> Q7
> How all of this I will test? I will need to run a migration?
> 
> Q8 Tests.
> Can I leave them for now and come back after all above will be clear?

Yeah, let's wait with tests for now.

pontoon/base/models.py
Karskaya: @mathjazz Please, review again.
I've remove all word_count related code for now and changed version check.
To make this PR finished I need a test, according to 
  
> The only thing that's missing for me is a small test to check if the method works as we expect.

Any advice with test? Should it looks the same as tests in test_entity.py? How can I create `objs` data? Can it be done like this? - https://github.com/mozilla/pontoon/blob/3462eff11975cf5442bc03d3ba6b617666f94fd2/pontoon/base/tests/managers/test_entity.py#L63-L71



>

mathjazz:
@jotes Could you please take a look?

pontoon/base/models.py
```diff
@@ -2245,6 +2248,15 @@ def reset_active_translations(self, locale):
...
+        return super(EntityQuerySet, self).get_or_create(defaults, **kwargs)
+
+    def bulk_update(self, objs, fields=None, batch_size=None):
+        if django.VERSION[0] >= 2:
```
jotes: Great work! that's what I meant.

Karskaya: @jotes Thank you!

mathjazz:
@Karskaya Given that the changeset is pretty small so far, let's continue fixing the bug in this PR.

Karskaya:
> @Karskaya Given that the changeset is pretty small so far, let's continue fixing the bug in this PR.

ok!

Karskaya:
Hello @mathjazz! Please, take a look. What do think about this attempt?

pontoon/base/models.py
```diff
@@ -2248,13 +2248,24 @@ def reset_active_translations(self, locale):
 
         translations.filter(pk__in=unreviewed_pks).update(active=True)
 
+    def _get_word_count(self, string):
+        return len(re.findall(r'\w+', string))
```
mathjazz: I'd simplify this to what @Pike does in [`compare-locales`](https://github.com/Pike/compare-locales/blob/c3c59dae92a717014c23c4708675a4555064f6fa/compare_locales/parser/fluent.py#L37):
```python
return len(string.value.split())
```

pontoon/base/models.py
```diff
@@ -2245,6 +2248,26 @@ def reset_active_translations(self, locale):
...
+
+    def get_or_create(self, defaults=None, **kwargs):
+        if 'string' in kwargs:
+            kwargs['word_count'] = self._get_word_count(kwargs['string'])
+        return super(EntityQuerySet, self).get_or_create(defaults, **kwargs)
+
+    def bulk_update(self, objs, fields=None, batch_size=None):
+        if django.VERSION[0] >= 2:
+            msg = "Django version is 2 or higher. Function bulk_update need to be removed"
```
mathjazz: need**s**

pontoon/base/models.py
```diff
@@ -2245,6 +2248,26 @@ def reset_active_translations(self, locale):
 
         translations.filter(pk__in=unreviewed_pks).update(active=True)
 
+    def _get_word_count(self, string):
+        return len(re.findall(r'\w+', string))
+
+    def get_or_create(self, defaults=None, **kwargs):
+        if 'string' in kwargs:
```
mathjazz: Why do we need this `if` clause?

pontoon/base/models.py
```diff
@@ -2245,6 +2248,26 @@ def reset_active_translations(self, locale):
...
+
+    def bulk_update(self, objs, fields=None, batch_size=None):
+        if django.VERSION[0] >= 2:
+            msg = "Django version is 2 or higher. Function bulk_update need to be removed"
+            warnings.warn(msg, PendingDeprecationWarning)
+        if objs:
+            for obj in objs:
+                if hasattr(obj, 'string'):
+                    obj.word_count = self._get_word_count(obj.string)
+                    obj.save()
```
mathjazz: Why do we need the two `if` clauses?

You shouldn't `save()` each item individually (slow) and instead let bulk_update make the change in the DB.

adngdb:

Karska-dev:
> The Black formatting commit has been merged to master, causing conflict on this branch. Please rebase and fix the conflicts as soon as you can, to avoid potentially more conflicts later on! And my apologies for disturbing your workflow, but I strongly believe now was the best time to do this. :-)

Done, thank you!

pontoon/base/models.py
Karska-dev: @mathjazz ok, I'll change it. 

Just want to notice that cases with dash will be counted in a wrong way. 
e.g. My brothers—Richard and John—are visiting Hanoi. 
using split method get_word_count will count 'brothers—Richard' as one word. 
or in case with spaces 'brothers — Richard' as three words.

All separately stand digits and symbols will be count as word as well(here is hard to understand if it's an expected result or no).

mathjazz: Good point. I think the first example lacks a space, but the 2nd example seems legit.

@Pike I'd like to keep the word count method consistent with [`compare-locales`](https://github.com/Pike/compare-locales/blob/c3c59dae92a717014c23c4708675a4555064f6fa/compare_locales/parser/fluent.py#L37), which in the example given by @Karska-dev counts `—` as words I believe. Are we fine with that?

Pike: Seems that there isn't just one answer, and the c-l wordcount isn't right. flod found
http://unicode.org/reports/tr29/#Word_Boundaries and https://docs.sdl.com/LiveContent/content/en-US/SDL%20WorldServer-v7/GUID-376E123B-1C7E-4D64-82B0-1D33F088ABD5#addHistory=true&filename=GUID-20199935-0650-461B-99E1-F9C45FCC8A43.xml&docid=GUID-20199935-0650-461B-99E1-F9C45FCC8A43&inner_id=&tid=&query=&scope=&resource=&toc=false&eventType=lcContent.loadDocGUID-20199935-0650-461B-99E1-F9C45FCC8A43

Jeff said that he'd like hyphenated words to be one.

My suggestion would be to use `r'[\w,.-]+'`

Karska-dev: Done

pontoon/base/models.py
Karska-dev: @mathjazz If there are no 'string' arguments passed to this method - I cannot count words for this entity. Is this situation real? Should I remove this 'if'?

pontoon/base/models.py
Karska-dev: @mathjazz Save() removed.

> Why do we need the two if clauses?
First clause check that objs exist and second that obj has string attribute. Are they excessive?

pontoon/base/models.py
```diff
@@ -2287,6 +2309,7 @@ class Entity(DirtyFieldsMixin, models.Model):
     order = models.PositiveIntegerField(default=0)
     source = JSONField(blank=True, default=list)  # List of paths to source code files
     obsolete = models.BooleanField(default=False)
+    word_count = models.PositiveIntegerField(default=0)
```
Karska-dev: I've added word_count property, but to make it work I need to run migration, am I right?

pontoon/base/tests/models/test_entity.py
```diff
@@ -397,3 +397,18 @@ def test_entity_project_locale_tags(entity_a, locale_a, tag_a):
...
+    """
+    created = True
+    assert created == True
+
+
+@pytest.mark.django_db
+def test_bulk_update():
+    """ Test
+    """
+    assert True
```
Karska-dev: I have this two empty tests for get_or_create and bulk_update accordingly. What about test data? I need EntityQuerySet to call both of them. How I can create it?

Karska-dev:
Hello @mathjazz !
Could you please help me here again?

pontoon/base/models.py
mathjazz: Yes, `string` should always be there:
https://github.com/mozilla/pontoon/blob/cc3c3267e8f279f1724d1f79fa0d351cb1c544fb/pontoon/sync/changeset.py#L181

pontoon/base/models.py
mathjazz: Good opint - the first one should stay. The second should always be `True`:
https://github.com/mozilla/pontoon/blob/cc3c3267e8f279f1724d1f79fa0d351cb1c544fb/pontoon/sync/changeset.py#L181

pontoon/base/models.py
mathjazz: Correct:
https://docs.djangoproject.com/en/3.0/topics/migrations/#workflow

pontoon/base/tests/models/test_entity.py
mathjazz: I would add a test case for the `_get_word_count` method, to test if we compute the word count as expected.

Then, I'd use your two tests to see if the `word_count` gets assigned during their execution. The tests should be moved to `pontoon/base/tests/managers/test_entity.py`, where you can find some inspiration on how to write such tests.

pontoon/base/tests/managers/test_entity.py
```diff
@@ -930,3 +930,49 @@ def test_mgr_entity_reset_active_translations(resource_a, locale_a):
...
+
+
+@pytest.mark.django_db
+def test_get_word_count(resource_a, locale_a):
+    """ How many words in a string
+    """
+    testEntitiesQuerySet = Entity.for_project_locale(resource_a.project, locale_a)
+    count = testEntitiesQuerySet._get_word_count("String 123 =+& string hh-gg object.string")
+    assert count == 5
+
```
Karska-dev: Hello @mathjazz ! Could you please review this PR again?

Do we need one more test for get_word_count for empty string?

pontoon/base/tests/managers/test_entity.py
```diff
@@ -930,3 +930,49 @@ def test_mgr_entity_reset_active_translations(resource_a, locale_a):
...
+    testEntitiesQuerySet = Entity.for_project_locale(resource_a.project, locale_a)
+    arguments = {
+        "resource": resource_a,
+        "string": 'String 123 =+& string hh-gg object.string',
+    }
+    obj, created = testEntitiesQuerySet.get_or_create(arguments)
+
+    assert created
+    assert obj.word_count == 5
+
```
Karska-dev: Is this a correct test data? 
Do we need test where object is already exists?(get_or_create make get, not create like in this test)

pontoon/base/tests/managers/test_entity.py
```diff
@@ -930,3 +930,49 @@ def test_mgr_entity_reset_active_translations(resource_a, locale_a):
...
+        "string",
+        "string_plural",
+        "key",
+        "comment",
+        "group_comment",
+        "resource_comment",
+        "order",
+        "source",
+    ]
+    assert True
```
Karska-dev: This one is not finished yet and I struggle with test data here.
bulk_update needs objs as argument, is list(testEntitiesQuerySet) a good candidate here?

mathjazz: Yup, I think it is!

pontoon/base/tests/managers/test_entity.py
mathjazz: This test should check if the `get_or_create()` updates the `word_count`. So you should first assert 0, and then after you run `get_or_create()` make sure it changes. Nit: I'd use a simpler string parameter. :)

adngdb:
Please also rebase this branch against `master` and make sure you use `black` to format Python code. See [documentation about black](https://mozilla-pontoon.readthedocs.io/en/latest/dev/contributing.html#python-code-conventions).

mathjazz:
Hey @Karska-dev, let us know if you need any help here. :)

Karska-dev:
Hi @mathjazz I had ambitious plan to continue work on it during my vacation, but something went wrong. I hope it's not a problem if I'll continue with this ticket after 23rd of February.
Also, I have a question. As I understood Pontoon using python 3 now, so does Pontoon upgraded to Django 2 already or not? If not, when are you going to do that?
I afraid, when I will finish with this issue part of it will be out of date.

mathjazz:
Hey @Karska-dev, thanks for the update. It's definitely not a problem to continue after Feb 23.

Pontoon is on Python 3, but still uses Django 1. We're in the process of upgrading to Django 2. We expect to get there before April. Don't worry about it for now, we'll update the patch if needed.

pontoon/base/tests/managers/test_entity.py
Karska-dev: > So you should first assert 0, and then after you run get_or_create() make sure it changes.

@mathjazz  I didn't get it. How I can assert 0 first?
Now I'm creating new entity via get_or_create() and check after if word_count is correct. 
How I can create first without involving get_or_create()?

pontoon/base/tests/managers/test_entity.py
```diff
@@ -928,3 +928,67 @@ def test_mgr_entity_reset_active_translations(resource_a, locale_a):
...
+    assert obj.word_count == 2
+
+
+@pytest.mark.django_db
+def test_mgr_bulk_update(resource_a, locale_a):
+    """
+    Update entities method.
+    """
+    objs = [
+        EntityFactory.create(resource=resource_a, string="testentity %s" % i,)
```
Karska-dev: @mathjazz I have a question here. In case I'm using EntityFactory to create a list of entities for bulk_update, I receive an error: 

> AttributeError: 'list' object has no attribute 'pk'

How I can fix this? May I just assign to pk attribute numbers '1,2,3 etc' during EntityFactory.create() ?

Karska-dev:
Hello @mathjazz ! Sorry for delay! Please, review and give me an answers for my two questions above. Thank you!

pontoon/base/tests/managers/test_entity.py
mathjazz: When the tests fails, you should see values of the relevant variables, which show that the `update_fields = None` and that the value of `meta` is what should be in `update_fields`:
```
objs = [<Entity: testentity 0>, <Entity: testentity 1>], meta = ['resource', 'string', 'string_plural', 'key', 'comment', 'group_comment', ...], update_fields = None, exclude_fields = None
using = 'default', batch_size = 2, pk_field = 'pk'
```

So we have a bug in https://github.com/mozilla/pontoon/pull/1495/files#diff-c18ce63ddb8aca155c0553a0c1b174e3R2315 - you should return `bulk_update` with keyword arguments, not positional:
```python
return bulk_update(objs, update_fields=update_fields, batch_size=batch_size)
```

That's for proving how useful the unit tests are! :-)

mathjazz:
Thanks for the update! Left a couple of comments.

pontoon/base/tests/managers/test_entity.py
Karska-dev: > you should return bulk_update with keyword arguments, not positional:

Oh, I see now.

Karska-dev:
Hello @mathjazz! Please, review my PR one more time!

mathjazz:
Excellent work, @Karska-dev!

One last thing you need to do is to properly update the migration:
1. Remove migration 0152.
2. Rename migration 0149 to 0152 and change its dependency to `0151_permalink_prefix_optional`.

mathjazz:
@Karska-dev Since I couldn't reach you here or on Matrix, I went ahead and fixed the migration (02f8e92) and added a data migration to calculate stats for all existing entities (7971eda). For future compatibility, I've also overriden the `Entity.save()` method to set `word_count` when a single Entity is created (1c29a64).

In the earlier three commits I've made smaller changes to make the changes described above easier to make.

Karska-dev:
> Thank you very much for fixing one of our oldest bugs, @Karska-dev!

Thank you for your wonderful mentoring! Stay safe!!!

PR #1497, opened 2019-11-19
Fixes bug 1597733 - Update celery to latest version.

adngdb:


mathjazz:
The tests are failing. Restarted, to see if it'll help.

adngdb:
Hah, seems this breaks support of Python 2. And tests fail with Python 3. I've done a bad job at testing my own changes!

adngdb:
It seems this isn't needed for sync to run on stage. Let's close for now and revisit later if needed.

PR #1500, opened 2019-11-21
Fix Bug 1597523 Add skeleton loader to editor and helper column

sccofield:
**What does this PR do?**
This pull request fixes the issue [1597523](https://bugzilla.mozilla.org/show_bug.cgi?id=1597523). It adds a skeleton loader to the editor and helper column.

Screenshot 2019-11-21 at 5 12 37 PM

### DIFF
--- a/frontend/src/core/loaders/components/EditorLoader.css
+++ b/frontend/src/core/loaders/components/EditorLoader.css
@@ -1,0 +1,45 @@
+.editor-loader {
+    animation: fading 1.5s infinite;
+    pointer-events: none;
+    overflow: hidden;
+    height: 100%;
+}
+
+.editor-loader .entity-navigation{
+    padding: 18px;
+}
+
+.editor-loader p {
+    height: 0.8125rem;
+    background-color: #E2E2E2;
+    padding-bottom: 3px;
+    margin: 20px 15px;
+}
+
+.editor-loader .metadata p {
+    margin: 5px 5px;
+}
+
+.editor-loader .text-2 {
+    width: 50%;
+    display: inline-block;
+}
+
+.editor-loader .editor {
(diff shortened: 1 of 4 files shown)


sccofield:
Hi, @mathjazz It was a bit difficult coming up with a good design for the skeleton loader for the remaining two columns. I don't know if this is okay.

Also, I would like to know if it is better to keep the two columns blank when no entity is available after filtering.

mathjazz:
Nice work, @sccofield!

I think the overall design is good, but I'd still make a few changes:
* In the 3rd column, I'd just use the skeleton from the string list (but without the checkboxes). slightly wider.
* Same for the history elements.
* I'd use two elements in the source string panel - one for the string and anther one (shorter) for metadata.
* I'd also make sure placeholders are aligned vertically and horizontally (see for example how the editor size is chosen).
* The placeholder for the topbar (string navigation + Machinery/Locales tabs) should use the same color and size as it does when entities are loaded.

Screenshot 2019-11-22 at 10 12 32

The loading animation between columns should be in sync - right now the string list and the rest of the UI animate with different starting times. Please also make sure we don't animate the horizontal bar on the top and the vertical line splitting the 2nd and the 3rd column.

Also, please hide the loader when no matches are found.

sccofield:
Hi @mathjazz I am really sorry for the delay in coming up with a fix. I have been a bit indispose and would not be able to continue with the issue.

mathjazz:
Not a problem at all, @sccofield! Thanks for the update, we'll take it from here.

All the best!

mathjazz:
Note to self: see http://farmdev.com/thoughts/108/the-art-of-ui-skeletons/.

mathjazz:
Closing due to inactivity.

PR #1503, opened 2019-11-22
Fixes bug 1473761 - Add an ActionLog table to log user actions.

adngdb:
I'm showing off where I'm at with this work. I haven't tested it locally yet (but tests do pass). The bulk of the work is done, what's left to do is:

- [x] Testing locally
- [x] Logging sync translation creation action
- [x] ~~Exposing data in public API~~ (I've decided to do this in a follow-up PR)


pontoon/actionlog/models.py
```diff
@@ -0,0 +1,104 @@
...
+        models.CASCADE,
+        blank=True,
+        null=True,
+    )
+    locale = models.ForeignKey(
+        Locale,
+        models.CASCADE,
+        blank=True,
+        null=True,
+    )
```
mathjazz: Should we use this opportunity to introduce the `deleted` flag on the `Translation` model instead of actually deleting entries from the DB? Eventually, that should also allow us to drop `ChangedEntityLocale` model.

pontoon/actionlog/models.py
```diff
@@ -0,0 +1,104 @@
...
+    Translation,
+    User,
+)
+
+
+class ActionLog(models.Model):
+    ACTIONS_TYPES = (
+        # A translation has been created.
+        ('translation:created', 'Translation created'),
+
```
mathjazz: There are three different actions in in `pontoon.base.views:update_translation()` (add first translation, add additional translation, update translation). We should log all three of them, possibly as 2 or 3 different types.

pontoon/actionlog/models.py
```diff
@@ -0,0 +1,104 @@
...
+    if instance.action_type not in valid_types:
+        raise ValidationError(
+            'Action type "{}" is not one of the permitted values: {}'.format(
+                instance.action_type,
+                ', '.join(valid_types)
+            )
+        )
+
+
+models.signals.pre_save.connect(validate_action_type_choice, sender=ActionLog)
```
mathjazz: We usually put signals in a separate file (see `pontoon.base.signals`). But here, maybe we don't even need signals? Maybe we can simply override the `save()` method?

pontoon/actionlog/utils.py
```diff
@@ -0,0 +1,34 @@
...
+    :returns: None
+    """
+    action = ActionLog(
+        action_type=action,
+        performed_by=user,
+        translation=translation,
+        entity=entity,
+        locale=locale,
+    )
+    action.save()
```
mathjazz: Do we really need this util function, given how simple it is?

pontoon/actionlog/models.py
adngdb: Yeah, I like signals because they're more flexible but after reading some opinions on the web, I think it's better to use `save()` here.

pontoon/actionlog/models.py
adngdb: What's "update translation"? Is that even possible? Why should we differentiate between adding the first translation to an entity and adding a non-first translation?

pontoon/actionlog/models.py
adngdb: Do we have a bug filed for that? I can't find one.

I think it's a good idea but it's going to have some impact on our code. We'll need to change a bunch of queries to exclude `deleted=true`.

If `ChangedEntityLocale` is only used to check what locales / entities have changed in a given period of time, then the `ActionLog` table as it is right now is enough to replace it, I believe.

pontoon/actionlog/utils.py
adngdb: It's simple but this feature will be used in a lot of places, and I find it easier to import and use this function than import the model and remember to use `.save()` every time. I think it's also more maintainable over time, especially since we know that this feature is going to grow (to log more actions, to be used in more places, etc. ).

adngdb:
>     1. Is this new table also going to replace the `PermissionChangelog`?
I guess it will in the future, yes. It certainly has the data structure to do so.

>     2. We'll likely need to denormalize some fields to speedup DB queries, e.g. for the Latest Activity columns on dashboards. I guess this will be done in a followup?
I'm not good enough with database queries to anticipate that. I'd rather keep it simple and handle performance issues when they arise. Maybe there won't be any problems!
 
>     3. We'll need to migrate existing data to the new model. I guess this will be done in a followup?
I hadn't thought about that. To me we'd start with just the data we have now. But that was not accounting for using this table for the latest activity feature. I guess back-filling could be done as part of the latest activity work?

pontoon/actionlog/models.py
mathjazz: Update happens when you unfuzzy or approve a suggestion by submitting identical translation:
https://github.com/mozilla/pontoon/blob/3d5d5130309b4119313aa4af7cdc907d6bd7d937/pontoon/base/views.py#L552

We don't need to differentiate between first/non-first translations. But we should log all 3 actions in `pontoon.base.views:update_translation()`.

Pike:
IIRC, bulk actions don't go through the model save method nor send signals, right? So we'll need to manually hook the action logs to bulk actions? Not sure if you've done so already, just glanced at the code.

pontoon/actionlog/models.py
mathjazz: We don't have a bug on file for that.

If you agree it's a good idea, I can file it. Fixing it would simplify this patch, but it's not a trivial fix, so we can also do it in a followup.

mathjazz:
> I hadn't thought about that. To me we'd start with just the data we have now. But that was not accounting for using this table for the latest activity feature. I guess back-filling could be done as part of the latest activity work?

I was thinking we could drop also `(un)approved_user/_date` and `(un)rejected_user/_date`:

https://github.com/mozilla/pontoon/blob/3d5d5130309b4119313aa4af7cdc907d6bd7d937/pontoon/base/models.py#L2736-L2752

pontoon/actionlog/models.py
adngdb: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1599377

pontoon/actionlog/models.py
```diff
@@ -0,0 +1,95 @@
...
+from django.db import models
+
+
+class ActionLog(models.Model):
+    ACTIONS_TYPES = (
+        # A translation has been created.
+        ('translation:created', 'Translation created'),
+
+        # A translation has been deleted.
+        ('translation:deleted', 'Translation deleted'),
```
jotes: [question] @adngdb what do you think about using enums in Python?
An enum with all possible actions could minimize some of the problems of "string-based" etc. (sorry if I'm stating the obvious)

Additionally, https://pypi.org/project/django-enum-choices/ gives you helpers to generate a choice field based on an enum-like class.

adngdb: I'm unsure about this. On one hand, I recognize that it's better to use an enum for this use case. On the other, it means more `import`s, and we already have this behavior coded with the type validation on `save`. 

Given that `enum` has been added in py3.4 and we'd need to add one more dependency and rewrite a bunch of code, I think I'd rather not do it. But if anyone has strong opinions in favor of enums, I'm happy to read them.

pontoon/base/models.py
```diff
@@ -2858,15 +2859,24 @@ def save(self, update_stats=True, failed_checks=None, *args, **kwargs):
...
-                    rejected_date=self.approved_date,
-                    fuzzy=False,
+            )
+
+            # Log that all those translations are rejected.
+            for t in approved_translations:
+                log_action(
+                    'translation:rejected',
+                    self.approved_user or self.user,
+                    translation=t,
```
mathjazz: We should use `bulk_create` here.

pontoon/sync/changeset.py
```diff
@@ -397,6 +410,12 @@ def bulk_update_entities(self):
 
     def bulk_create_translations(self):
         Translation.objects.bulk_create(self.translations_to_create)
+        for translation in self.translations_to_create:
+            log_action(
+                'translation:created',
+                translation.user or self.sync_user,
+                translation=translation,
+            )
```
mathjazz: We must use `bulk_create` here, same as we do for TM and checks.

Otherwise adding a locale to Firefox or running a Fluent migration will result in tens of thousands of new DB calls.

pontoon/sync/changeset.py
```diff
@@ -259,6 +262,11 @@ def update_entity_translations_from_vcs(
             # Modify existing translation.
             if db_translation:
                 if not db_translation.approved and not vcs_translation.fuzzy:
+                    log_action(
+                        'translation:approved',
+                        user or self.sync_user,
+                        translation=db_translation,
+                    )
```
mathjazz: We should use `bulk_create` instead (also for the `'translation:rejected'` action below).

pontoon/base/models.py
adngdb: I object, in this specific case, there should always be only one translation there. It's done in a loop because it's easier to do it this way, but there should never be more than one approved translation when approving a new one, so there should never be more than one rejection event to log.

mathjazz: Good point, sorry for the confusion!

mathjazz:
Could you please fix the tests first?

pontoon/sync/tests/test_changeset.py
```diff
@@ -306,13 +311,39 @@ def test_update_db_unapprove_existing(self):
...
+        self.main_vcs_translation.strings[None] = 'New Translated String'
+
+        self.update_main_db_entity()
+        self.main_db_translation.refresh_from_db()
+        assert_attributes_equal(
+            self.main_db_translation,
+            approved=False,
+            approved_user=None,
+            approved_date=None,
+            rejected=False,
```
mathjazz: You should also add `fuzzy=False`.

pontoon/sync/tests/test_changeset.py
```diff
@@ -226,6 +227,11 @@ def test_update_db_approve_translation(self):
             approved_date=aware_datetime(1970, 1, 1)
         )
 
+        assert ActionLog.objects.filter(
+            action_type='translation:approved',
+            translation=self.main_db_translation.pk
```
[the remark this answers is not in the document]
adngdb: Well that makes me want to have Black even more! :D

pontoon/batch/actions.py
```diff
@@ -69,6 +70,16 @@ def approve_translations(form, user, translations, locale):
         translations,
         locale,
     )
+
+    # Log actions
```
mathjazz: Did you want to make this comment and the one below more specific, to make them in line with the 3rd and the 4th one, e.g. "Log approving actions"?

pontoon/actionlog/admin.py
```diff
@@ -18,6 +19,14 @@ class ActionLogAdmin(admin.ModelAdmin):
...
         'locale_id',
     )
+    fieldsets = (
+        (None, {
+            'fields': (
+                'action_type',
+                'performed_by',
+            ),
+        }),
+    )
```
mathjazz: Instead of this, I'd use:
```
raw_id_fields = ('entity', 'translation',)
```

adngdb: Sweet! I went looking for something like this in the docs but failed. Thank you! :slightly_smiling_face:

PR #1506, opened 2019-11-28
Fix bug 1597558 - Send errors to Raygun.io on Python 3.

jotes:
Unfortunately, I couldn't find a better way to handle this.
To give you the fuller background:
On Python 2, Raygun4py fails to send an exception with the Unicode message and throws the UnicodeDecodeError. To avoid this, the middleware casts the exception to the `str` object.
On Python 3, Raygun4py requires an instance of an Exception instead of a string.
@pike @adngdb r?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,11 @@ shell-py3: shell
 pytest-py3: override PYTHON_VERSION=$(PYTHON_3_VERSION)
 pytest-py3: pytest
 
+run-py3: override PYTHON_VERSION=$(PYTHON_3_VERSION)
+run-py3: run
+
+setup-py3: override PYTHON_VERSION=$(PYTHON_3_VERSION)
+setup-py3: setup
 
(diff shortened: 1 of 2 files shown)


adngdb:
That looks good to me, but could you please add a comment near the code that explains what you said in this PR?

mathjazz:
Should we also file a bug to remove the py2 specific code once we update to dj2?

adngdb:
I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1600344. @jotes can you also please reference this bug in your code comment, so that it's easy later to find it back?

jotes:
> I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1600344. @jotes can you also please reference this bug in your code comment, so that it's easy later to find it back?

thanks @adngdb, I'll add it tomorrow.

PR #1507, opened 2019-11-29
Bug 1599075 - Show Approve button when translation already exists.

adngdb:
This makes the actions behind the "Save" button of the Editor clearer: only two UI states where shown so far, but in reality three kinds of actions could happen. The "Approve" action was hidden, and this commit makes it explicit by changing the button label to show what is going to happen in the back-end. This also makes the front-end control the behavior instead of relying on the back-end, allowing us to make more focus backend views in a follow-up.

@mathjazz in order to avoid conflicts with my ActionLog PR, I want to do the back-end part later.


frontend/src/core/editor/components/EditorMainAction.js
```diff
@@ -0,0 +1,107 @@
...
+        if (translation === initialTranslation) {
+            sameTranslationAs = activeTranslation;
+        }
+        else if (history.translations.length) {
+            sameTranslationAs = history.translations.find(t => t.string === translation)
+        }
+    }
+
+    function approveTranslation() {
+        if (sameTranslationAs) {
```
mathjazz: Is this check necessary? Seems like `approveTranslation()` can only be called if the condition is true.

frontend/src/core/editor/components/EditorMainAction.js
```diff
@@ -0,0 +1,107 @@
...
+            updateTranslationStatus(sameTranslationAs.pk, 'approve');
+        }
+    }
+
+    if (isTranslator && sameTranslationAs && !sameTranslationAs.approved) {
+        // Approve button, will approve the translation.
+        return <Localized
+            id="editor-EditorMenu--button-approve"
```
mathjazz: Please also update the FTL file accordingly.

frontend/src/core/editor/components/EditorMainAction.test.js
```diff
@@ -0,0 +1,124 @@
...
+            translation: 'hello, world!',
+        }
+        const wrapper = shallow(<EditorMainAction { ...params } />);
+
+        expect(wrapper.find('.action-save')).toHaveLength(1);
+        expect(wrapper.find('.action-suggest')).toHaveLength(0);
+        expect(wrapper.find('.action-approve')).toHaveLength(0);
+    });
+
+    it('renders the Save button when user does not have permission', () => {
```
mathjazz: I think we have a bug (here and in prod): we should show the "Suggest" button for non-privileged users.

https://mozilla-l10n.github.io/localizer-documentation/tools/pontoon/translate.html#submit-a-suggestion

frontend/src/core/editor/components/EditorMainAction.test.js
```diff
@@ -0,0 +1,124 @@
...
+        expect(wrapper.find('.action-suggest')).toHaveLength(0);
+        expect(wrapper.find('.action-approve')).toHaveLength(0);
+    });
+
+    it('renders the Save button when same translation exists but is approved', () => {
+        const params = {
+            initialTranslation: 'something',
+            isTranslator: true,
+            history: HISTORY,
+            translation: 'hello, world!',
```
mathjazz: Translations are not the same, which is what the test title says.

frontend/src/core/editor/components/EditorMainAction.js
```diff
@@ -0,0 +1,107 @@
...
+        initialTranslation,
+        isTranslator,
+        forceSuggestions,
+        history,
+        translation,
+        sendTranslation,
+        updateTranslationStatus,
+    } = props;
+
+    let sameTranslationAs = null;
```
mathjazz: Maybe `existingTranslation`?

frontend/src/core/editor/components/EditorMainAction.js
adngdb: It's needed for Flow. :sweat_smile:

frontend/src/core/editor/components/EditorMainAction.test.js
adngdb: It is the same as one that exists in the history. :)

mathjazz: Ah... Maybe add "exists **in history**" to make it more obvious?

frontend/src/core/editor/components/EditorMainAction.test.js
adngdb: Nice catch! Thank you unit tests! :D

mathjazz:
Excellent job! There's just one thing we should fix - if a check fails when submitting a translation, we show the green "Save Anyway" button instead of the blue "Suggest Anyway" button to non-privileged users.

adngdb:
I realized while working on the "Save Anyway" feature that I forgot to code the correct behavior for the "Enter" shortcut. It's a big problem as it will break the feature when we remove the back-end equivalent feature.

adngdb:
@mathjazz it turned out to require a bunch more work, but hopefully everything works now! (Main editor action button, Enter shortcut, and FailedChecks. )

adngdb:
@mathjazz done! :-)

PR #1508, opened 2019-11-29
Fix Bug 1600134 - Refactor get_translations_from_other_locales() API

abowler2:
Initial work on refactoring get_translations_from_other_locales() API
_____
Can you verify if this is what you had in mind for creating a dict with 2 lists: `preferred_locales` and `other_locales`?

I know there is one issue ATM with `preferred_locales` in that I haven't worked out how to successfully exclude the source locale if it is also a preferred locale.


pontoon/base/views.py
```diff
@@ -295,6 +295,7 @@ def get_translations_from_other_locales(request):
 
     entity = get_object_or_404(Entity, pk=entity)
     locales = entity.resource.project.locales.exclude(code=locale)
+    selected_locales = request.user.profile.sorted_locales
```
mathjazz: I doubt you'll be able to use `sorted_locales` and keep the order. Instead, you'll need to manually sort `preferred_locales` below.

pontoon/base/views.py
```diff
@@ -304,9 +305,23 @@ def get_translations_from_other_locales(request):
         approved=True
     ).order_by('locale__name')
 
-    payload = list(translations.values(
+    other_locales = list(translations.values(
```
mathjazz: You will need to exclude preferred locales.

pontoon/base/views.py
```diff
@@ -304,9 +305,23 @@ def get_translations_from_other_locales(request):
...
+    preferred_source_translations = Translation.objects.filter(
+        entity=entity,
+        locale__in=selected_locales,
+        plural_form=plural_form,
+        approved=True
+    )
+
+    preferred_locales = list(preferred_source_translations.values(
+        'locale__code', 'locale__name', 'locale__direction', 'locale__script', 'string'
+    ))
```
mathjazz: I would create a helper function that would return a list of requested values for both, preferred and other locales. That way you only need to list values once.

pontoon/base/views.py
```diff
@@ -304,9 +305,23 @@ def get_translations_from_other_locales(request):
...
+
+    payload = {'preferred_locales': preferred_locales, 'other_locales': other_locales}
```
mathjazz: Please put attributes in separate lines:

```python
x = {
    `a`: a,
    `b`: b,
}
```

pontoon/base/views.py
```diff
@@ -295,6 +295,7 @@ def get_translations_from_other_locales(request):
 
     entity = get_object_or_404(Entity, pk=entity)
     locales = entity.resource.project.locales.exclude(code=locale)
```
mathjazz: This is unrelated to your patch, but instead of using `locales` and `locale__in=locales`, we should use `locale = get_object_or_404(Locale, code=locale)` and `.exclude(locale=locale)` .

abowler2: Can you please clarify. If I eliminate `locales` and replace with `locale = get_object_or_404(Locale, code=locale)` I would only get the one locale that was passed into the method as opposed to all the locales in the project and then if I exclude with `locale=locale` it would produce an empty array.

mathjazz: You also need to remove the `locale__in=locales` filter:
```python
translations = (
    Translation.objects
    .filter(
        entity=entity,
        plural_form=plural_form,
        approved=True,
    )
    .exclude(locale=locale)
    .order_by('locale__name')
)
```

abowler2: So, this works if I don't add `.exclude(locale=locale)`, but then it does not exclude the current locale. When I add the exclusion the app crashes and I get `undefined` as the value of `translations`

Also, while we are here....the list I am getting locally is not ordered by `locale__name` (this was the case even with the original code)

I have made some adjustments to the code in `entity.js` to parse the new data structure, which is is still a WIP but it seems to be parsing the lists correctly. I can push that if you would like to see as it's possible that is part of the problem.

abowler2: Nevermind.....I had a typo 😞

abowler2: However, it as I stated above, it's not sorting correctly for me and this was true even for the original code.  Which is why I did not have the `.order_by` on the `preferred_locales list`

mathjazz: > Nevermind.....I had a typo 😞

Welcome to my World. :)

Feel free to push your WIP.

And yes - you should drop `.order_by('locale__name')`.

abowler2:
Just pushed my work so far.  Still working on excluding `preferred_locales` from the `other_locales` list and sorting the lists by `locale_name`

Also, I'm only returning one list in `entity.js` ATM as I haven't refactored the rest of the code in response to the changes made thus far.

abowler2:
Updated work ~ 
Still need to work out how to sort the lists properly and finish refactoring the frontend code.

pontoon/base/views.py
```diff
@@ -294,19 +300,36 @@ def get_translations_from_other_locales(request):
...
+    locale = get_object_or_404(Locale, code=locale)
+    selected_locales = request.user.profile.sorted_locales
     plural_form = None if entity.string_plural == "" else 0
 
+    preferred_source_translations = Translation.objects.filter(
+        entity=entity,
+        locale__in=selected_locales,
+        plural_form=plural_form,
+        approved=True
+    ).exclude(locale=locale)
```
mathjazz: If you move this block under `translations =`, you can avoid code duplication and do:
```python
preferred = translations.filter(locale__in=preferred_locales)
other = translations.exclude(locale__in=preferred_locales)
```

pontoon/base/views.py
```diff
@@ -294,19 +300,36 @@ def get_translations_from_other_locales(request):
         }, status=400)
 
     entity = get_object_or_404(Entity, pk=entity)
-    locales = entity.resource.project.locales.exclude(code=locale)
+    locale = get_object_or_404(Locale, code=locale)
+    selected_locales = request.user.profile.sorted_locales
```
mathjazz: I'd call the variable `preferred_locales`.

pontoon/base/views.py
```diff
@@ -294,19 +300,36 @@ def get_translations_from_other_locales(request):
...
         plural_form=plural_form,
         approved=True
-    ).order_by('locale__name')
+    ).exclude(locale=locale)
+
+    other_locales = list(get_translation_values(translations))
+
+    preferred_locale_codes = [element['locale__code'] for element in preferred_locales]
+    for code in preferred_locale_codes:
+        other_locales = [l for l in other_locales if not (l['locale__code'] == code)]
```
mathjazz: You can drop these three lines if you define `preferred/other` as described above.

abowler2:
In working on refactoring the frontend code as per the [description](https://bugzilla.mozilla.org/show_bug.cgi?id=1600134#c0) on the Bug I'm running into issues since the data structure has changed.

As I have it now `entity.js` returns an Array with 2 Arrays of Objects instead of the original structure of just one Array of Objects. Every effort to access and then use the inner Arrays has thrown errors. I can concatenate the lists and then the code works fine, but I'm pretty sure that the intent was to use the lists separately as needed.

I can get access to the inner Arrays, but the moment I try to do anything with them I get errors such as: `Cannot call otherlocales.translations[0].reduce because property reduce is missing in OtherLocaleTranslation`

If you could direct me to information on how to tackle this I would appreciate it.

pontoon/base/views.py
```diff
@@ -281,6 +281,16 @@ def entities(request):
...
 
 
+def _get_translation_values(query):
+    return query.values(
+        'locale__code',
+        'locale__name',
+        'locale__direction',
+        'locale__script',
+        'string',
+    )
```
mathjazz: To simplify the logic on frontend, I would convert the output to:
```python
    return [{
        'code': translation.locale.code,
        'direction': translation.locale.direction,
        'locale': translation.locale.name,
        'locale_pk': translation.locale.pk,
        'script': translation.locale.script,
        'translation': translation.string,
    } for translation in query]
```

Mind the `locale_pk`, which we'll need for sorting below.

pontoon/base/views.py
```diff
@@ -294,19 +304,27 @@ def get_translations_from_other_locales(request):
...
-        locale__in=locales,
         plural_form=plural_form,
         approved=True
-    ).order_by('locale__name')
+    ).exclude(locale=locale)
+
+    preferred = translations.filter(locale__in=preferred_locales)
+    other = translations.exclude(locale__in=preferred_locales)
+
+    preferred_locales_list = list(_get_translation_values(preferred))
```
mathjazz: We should keep the order of locales the same. Taking into account the suggested naming changes, you can do it like this:
```python
preferred_translations = sorted(
    _get_translation_values(preferred),
    key=lambda t: request.user.profile.locales_order.index(t['locale_pk'])
)
```

pontoon/base/views.py
```diff
@@ -294,19 +304,27 @@ def get_translations_from_other_locales(request):
         }, status=400)
 
     entity = get_object_or_404(Entity, pk=entity)
-    locales = entity.resource.project.locales.exclude(code=locale)
+    locale = get_object_or_404(Locale, code=locale)
+    preferred_locales = request.user.profile.sorted_locales
```
mathjazz: In `models.py` we should introduce a new `preferred_locales` property and use that instead of `sorted_locales` (sorting will be lost below, so we'll have to sort again anyways):
```python
@property
def preferred_locales(self):
    locales = Locale.objects.filter(pk__in=self.locales_order)
```

We should then also make `sorted_locales` property use `preferred_locales`.

frontend/src/core/api/entity.js
```diff
@@ -141,18 +141,22 @@ export default class EntityAPI extends APIBase {
 
         const results = await this.fetch('/other-locales/', 'GET', payload, headers);
 
-        if (!Array.isArray(results)) {
+        if (typeof results !== 'object' || Array.isArray(results) || results == null) {
             return [];
```
mathjazz: We shouldn't return `[]`, because `LocalesState.translations` will not be an `Array` anymore, it will be an nobject. I suggest you return `null`.

frontend/src/core/api/entity.js
```diff
@@ -141,18 +141,22 @@ export default class EntityAPI extends APIBase {
...
+            return locale_lists.map(entry => {
+                return {
+                    code: entry.locale__code,
+                    locale: entry.locale__name,
+                    direction: entry.locale__direction,
+                    script: entry.locale__script,
+                    translation: entry.string,
+                }
+            })
+        } );
```
mathjazz: With https://github.com/mozilla/pontoon/pull/1508#discussion_r353735751, you can get rid of this entire block.

You'll need to update actions, reducer and component logic in the `otherlocales` module though, to take into account the new format of the `results` AKA `translations`.

frontend/src/core/api/entity.js
abowler2: When I return `null` here I get an error that `null is incompatible with array type`

mathjazz: Right, that's why you'll also need to update `actions` and the `reducer`: https://github.com/mozilla/pontoon/pull/1508#discussion_r353759554.

frontend/src/core/api/entity.js
abowler2: In changing the format type from Array to Object I keep coming up against errors regarding `inexact` values being incompatible with `exact OtherLocaleTranslation`. 

For example when I try changing the initial state of `translations` in reducer.js from an empty array to an empty object I get:  `Inexact object literal [1] is incompatible with exact OtherLocaleTranslation [2] in property translations.`

Should `results` not be set as an Object?

mathjazz: Could you try defining `translations` like this?
```js
+translations: {
    preferred: Array<OtherLocaleTranslation>,
    other: Array<OtherLocaleTranslation>,
},
```

Now that I look at this, I'm thinking maybe we should rename preferred_translations -> preferred and other_translations -> other.

abowler2: That seems to have worked, but I did have to make that change in actions.js as well as for the Promise in entity.js as well.

The errors I'm getting now are all in relation to the format change with the exception of 
```javascript
if (results.status === false) {
            return null;
     }
```
which throws:  `Cannot return null because inexact null [1] is incompatible with exact
OtherLocaleTranslation [2] in type argument R`

mathjazz: Glad to hear it worked. Well done!

You can either:
* return `{}` instead of null or
* use `null` as the default value of the `initialState`, in which case you should change `+translations: {` to `+translations: ?{` to make it [optional](https://flow.org/en/docs/types/maybe/).

abowler2:
It looks as though the payload object from views.py is empty. I'm working on tracking down the issue but pushed my code in case you have a moment to check it out (or I'm not able to find the issue). *Fair warning: this is early stage, just-trying-to-get-it-to-render code 😁 

Also, here is a link to the pastbin for the error log I got: https://pastebin.com/0rcmAyGr

pontoon/base/views.py
```diff
@@ -282,13 +282,14 @@ def entities(request):
...
-        'string',
-    )
+    return [{
+        'code': translation.locale.code,
+        'direction': translation.locale.direction,
+        'locale': translation.locale.name,
+        'locale_pk': translation.locale.pk,
+        'script': translation.locale.script,
+        'translation': translation.string,
+    } for translation in query]
```
mathjazz: Sorry for changing my mind, but since we're refactoring code, let's make sure we do it properly. I believe it will be way less confusing if we introduce another level of nesting in the output to hold all locale-specific values:
```python
return [
    {
        'locale': {
            'code': translation.locale.code,
            'direction': translation.locale.direction,
            'name': translation.locale.name,
            'pk': translation.locale.pk,
            'script': translation.locale.script,
        },
        'translation': translation.string,
    }
    for translation in query
]
```

pontoon/base/views.py
```diff
@@ -294,19 +305,33 @@ def get_translations_from_other_locales(request):
         }, status=400)
 
     entity = get_object_or_404(Entity, pk=entity)
-    locales = entity.resource.project.locales.exclude(code=locale)
+    locale = get_object_or_404(Locale, code=locale)
+    preferred_locales = request.user.profile.preferred_locales
```
mathjazz: We'll need to make sure the code also runs for anonymous users - right now it breaks.

So we'll need to introduce some check ([example](https://github.com/mozilla/pontoon/commit/c4bacd58e74cdcb1cbfeb8b56fc03692398d87ad)) before executing the code related to preferred locales.

pontoon/base/views.py
```diff
@@ -281,6 +281,17 @@ def entities(request):
     return _get_paginated_entities(locale, preferred_source_locale, project, form, entities)
 
 
+def _get_translation_values(query):
```
mathjazz: Maybe `_serialize_translation_values` is even better.

frontend/src/core/api/entity.js
```diff
@@ -131,7 +131,10 @@ export default class EntityAPI extends APIBase {
     async getOtherLocales(
         entity: number,
         locale: string,
-    ): Promise<Array<OtherLocaleTranslation>> {
+    ): Promise<{
```
mathjazz: Since we're repeating this patter a lot, I suggest you change this to `Promise<OtherLocaleTranslations>` and define a new `OtherLocaleTranslations` type:
```js
export type OtherLocaleTranslations = {|
    +preferred: Array<OtherLocaleTranslation>,
    +other: Array<OtherLocaleTranslation>,
|};
```

abowler2: This change brings back the Flow error that the `object literal is incompatible with exact OtherLocaleTranslations`.

Also, would this be the correct typing for the new format of the `translation`: 
```javascript
export type OtherLocaleTranslation = {|
    +locale: {|
        +code: string,
        +locale: string,
        +locale_pk: number,
        +direction: string,
        +script: string,
    |},
    +translation: string,
|};
```
Or would I just set `locale: Object`

mathjazz: Make sure you prepend the `OtherLocaleTranslations` with `?` to make it [optional](https://flow.org/en/docs/types/maybe/) when you use it (e.g. `+translations: ?OtherLocaleTranslations`) and then use `null` instead of `{}`.

The `OtherLocaleTranslation` type looks good (using `Object` is cheating hehe), but you'll need to use the same keys as in https://github.com/mozilla/pontoon/pull/1508#discussion_r354178848.

abowler2:
Updated code based on your previous feedback. I'm having a couple of issues though (I know, big surprise right 😂 )

1) If I don't use `?OtherLocaleTranslations` then I get the "inexact incompatible with exact" errors when setting empty objects or null. But, when I use the `?` then I get the following Flow error:
```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/modules/otherlocales/components/OtherLocales.js:55:14

null or undefined [1] is not an object.

     src/modules/otherlocales/components/OtherLocales.js
     52│             return null;
     53│         }
     54│
     55│         if (!Object.keys(otherlocales.translations).length) {
     56│             return this.renderNoResults();
     57│         }
     58│

     src/modules/otherlocales/reducer.js
 [1] 18│     +translations: ?OtherLocaleTranslations
```
Which highlights `(!Object.keys(otherlocales.translations).length)` as the issue. However, when I console.log this it logs exactly what is expected.

2) Somewhat related, when I try accessing anything past `otherlocales.translations` Flow gives me this error: `Cannot get otherlocales.translations.other because property other is missing in null or undefined` but again console.log shows the correct values.

Your thoughts on fixing this are appreciated.

pontoon/base/views.py
```diff
@@ -316,16 +322,13 @@ def get_translations_from_other_locales(request):
     ).exclude(locale=locale)
 
     preferred = translations.filter(locale__in=preferred_locales)
```
mathjazz: If the user isn't authenticated, this will break, because `preferred_locales` isn't defined. You'll need to execute all the code related to preferred locales only for authenticated users.

frontend/src/core/api/types.js
```diff
@@ -43,14 +43,26 @@ export type Entities = Array<Entity>;
 
 
 /*
- * Translation of an entity in a locale other than the currently selected.
+ * Preferred and Other Translations of an entity in a locale other than
+ * the currently selected locale.
```
mathjazz: Nit: "A collection of translations of an entity to a locale..."

frontend/src/core/api/types.js
```diff
@@ -43,14 +43,26 @@ export type Entities = Array<Entity>;
...
+ */
+export type OtherLocaleTranslations = {|
+    +preferred: Array<OtherLocaleTranslation>,
+    +other: Array<OtherLocaleTranslation>,
+|};
+
+
+/*
+ * Translation of an entity in a locale.
```
mathjazz: Nit: "Translation of an entity to a locale other than the currently selected locale."

abowler2:
Successfully refactored to have the Locales tab render 'other translations' and append the 'en-US' string to the preferred list when an alternate source locale is selected.

Couple things to note:
1) The 'en-US' string is not linked when appended to the preferred list as intended, but since the pointer is set for the whole section it is still present when you hover over `name/code` (it does not highlight green though). It feels a little 'weird' from a UX perspective, but I'm not sure how much it matters.
2) My local environment orders the 'other' translations by `code` no matter what value I put in `order_by` in the `views.py` file. I have it set to 'locale__name' as that is what it was originally sorted by.

Let me know if any changes are needed.

frontend/src/modules/otherlocales/components/OtherLocales.js
```diff
@@ -37,41 +35,65 @@ export default class OtherLocales extends React.Component<Props> {
...
-                        lastPreferred={ lastPreferred }
-                        key={ index }
-                    />;
-                }) }
-            </ul>
+        return <section>
+            <section className="other-locales">
+                <ul>
+                    { translation.preferred.map((translation, index) => {
+                        const lastPreferred = (index === preferredCount -1);
```
mathjazz: You can remove the `lastPreferred` logic from JS completely and use CSS to draw the green border bottom color on the last item in the first `<ul>`.

frontend/src/modules/entitydetails/components/Helpers.js
```diff
@@ -45,16 +43,19 @@ export default class Helpers extends React.Component<Props> {
...
-            preferredLocalesCount,
             parameters,
             user,
             updateEditorTranslation,
             searchMachinery,
         } = this.props;
 
+        if (!otherlocales.translations) {
+            return null;
+        }
```
mathjazz: Supernit: you could move this under declarations and then use `if (!translations) {`. :)

frontend/src/modules/otherlocales/components/OtherLocales.js
```diff
@@ -37,41 +35,56 @@ export default class OtherLocales extends React.Component<Props> {
...
-            preferredLocalesCount,
-            otherlocales,
-            parameters,
-            updateEditorTranslation,
-        } = this.props;
+        const { otherlocales } = this.props;
+
+        if (!otherlocales.translations) {
+            return null;
+        }
```
mathjazz: Same Übernit as above. :)

frontend/src/modules/otherlocales/components/OtherLocales.js
```diff
@@ -37,41 +35,56 @@ export default class OtherLocales extends React.Component<Props> {
...
-                        updateEditorTranslation={ updateEditorTranslation }
-                        lastPreferred={ lastPreferred }
-                        key={ index }
-                    />;
-                }) }
-            </ul>
-        </section>;
+                <ul className="preferred-list">
+                    { translation.preferred.map((translation, index) => {
+
```
mathjazz: I'd remove this blank line (and the one below). And maybe add one to separate `<ul>`'s.

frontend/src/modules/otherlocales/components/OtherLocales.js
```diff
@@ -37,41 +35,56 @@ export default class OtherLocales extends React.Component<Props> {
...
-                        translation={ translation }
-                        parameters={ parameters }
-                        updateEditorTranslation={ updateEditorTranslation }
-                        lastPreferred={ lastPreferred }
-                        key={ index }
-                    />;
-                }) }
-            </ul>
-        </section>;
+                <ul className="preferred-list">
```
mathjazz: You can dedent this entire block by 1 level.

pontoon/base/views.py
```diff
@@ -281,6 +281,22 @@ def entities(request):
...
+def _serialize_translation_values(query):
+    return [
+        {
+            'locale': {
+                'code': translation.locale.code,
+                'direction': translation.locale.direction,
+                'name': translation.locale.name,
+                'pk': translation.locale.pk,
+                'script': translation.locale.script,
+            },
```
mathjazz: I just realized we have a `serialize()` method on the `Locale` class. If you add `pk` to it, you could just to this:
```python
'locale': translation.locale.serialize(),
```

And for the source string:
```python
'locale': Locale.objects.get(code='en-US').serialize(),
```

frontend/src/modules/entitydetails/components/Helpers.js
abowler2: I actually already tried that but I get errors from Flow about `translations` missing in null or undefined.

mathjazz: Oh. Right. Mr Flow. :/ Let's keep it, then.

abowler2:
Made the requested changes and fixed it so that it no longer breaks when logged out.

Let me know if any further changes are needed.

mathjazz:
Thanks for the update! I went ahead and pushed a commit that uses `.values()` instead of `.serialize()`, which serializes data much faster. Before that I tried using `prefetch_related()` on `translations` to reduce the number of DB calls, but it didn't have such a big impact.

I've noticed one bug though - why you click on strings in the string list fast one after another, the app breaks with an error:
```
TypeError: translations.other is undefinedHelpers.js:58
```

abowler2:
So, I forgot to pull your changes before making mine and then tried to fix my mistake. Clearly I didn't do this correctly 😂 . I don't think anything is broken, and in fact I was able to fix the bug you found.

Let me know if I need to roll back or make any other fixes/changes.

mathjazz:
Sorry about that, I should have let you commit my patch instead.

Could you rebase the branch against the current master?

frontend/src/modules/entitydetails/components/Helpers.js
```diff
@@ -55,7 +55,7 @@ export default class Helpers extends React.Component<Props> {
 
         const machineryCount = machinery.translations.length;
         const translations = otherlocales.translations;
-        const otherlocalesCount = translations.other.length || translations.preferred.length;
+        const otherlocalesCount = Object.keys(translations).length === 0;
```
mathjazz: `translations` should always have keys, unless we're `fetching`. In practice however, since we abort the request
when switching strings quickly, `translations` might indeed become [`{}`](https://github.com/mozilla/pontoon/blob/1174c7027c91bc85f601d90f2cbed07bfc50a5fb/frontend/src/core/api/base.js#L71).

I've created a patch to store `null` instead of `{}` when request gets aborted. Please download and `git apply` it:
https://gist.github.com/mathjazz/3486ce4c41ffe96e330db046ea4b3de4

The patch also moves the checks from this file to corresponding `Count` modules, so that we don't for example block presenting Machinery count while waiting for Locales.

frontend/src/modules/otherlocales/components/OtherLocales.js
```diff
@@ -37,39 +35,53 @@ export default class OtherLocales extends React.Component<Props> {
...
-            parameters,
-            updateEditorTranslation,
-        } = this.props;
+        const { otherlocales } = this.props;
+
+        if (!otherlocales.translations) {
+            return null;
+        }
+
+        const translation = otherlocales.translations;
```
mathjazz: One more nit I have found: this should actually be called `translations`.

abowler2:
Status: Rebased with current master, applied your patch, and made requested changes. 

However, the patch caused the app to break for me with the following error: `Error: Cannot find module 'lodash.isempty'`.  Is there another step I need to take for this such as updating a dependency?

Also, when there are no translations in other locales I gave the section the `history` className instead of `other-locales` to replicate the styling, but I'm not sure if that is what you had in mind or if you preferred that I add a separate className and just change the padding.

Finally, I am soooo sorry for the mess that I made with the commits on this PR :weary:

mathjazz:
Thanks for the update! Sadly it seems like something went wrong with the rebase (check out the commit count). I wasn't able to fix it easily, so I'm closing this on behalf of #1517.

The reason for the error you were seeing is that you need to run `make build` after new [dependencies](https://github.com/mozilla/pontoon/tree/master/frontend#dependencies) are introduced.

PR #1509, opened 2019-12-02
Bug 1585252 - Fix string encoding errors for sync with Python 3.

adngdb:
@jotes mind giving this a quick look too, please?

### DIFF
--- a/pontoon/sync/vcs/repositories.py
+++ b/pontoon/sync/vcs/repositories.py
@@ -173,9 +173,6 @@ class CommitToGit(CommitToRepository):
         if code != 0:
             raise CommitToRepositoryException(text_type(error))
 
-        if type(error) is bytes:
-            error = error.decode('utf-8')
-
         if 'Everything up-to-date' in error:
             return self.nothing_to_commit()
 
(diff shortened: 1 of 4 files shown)


pontoon/sync/vcs/repositories.py
```diff
@@ -327,6 +330,8 @@ def normalize_revision(rev):
             ['svn', 'diff', '-r', '{}:{}'.format(from_revision, 'HEAD'), '--summarize'],
             cwd=path
         )
+        if type(output) is bytes:
```
jotes: Do you want to drop this on Python 3?

adngdb: I'm not sure, I've seen cases where this was an instance of `str` and I wasn't able to figure out when or why.

pontoon/sync/vcs/repositories.py
```diff
@@ -173,6 +173,9 @@ def commit(self, path=None, message=None, user=None, branch=None):
         if code != 0:
             raise CommitToRepositoryException(text_type(error))
 
+        if type(error) is bytes:
+            error = error.decode('utf-8')
```
mathjazz: Why do you only add the check for the `commit()` method of the `CommitToGit` class, and not also other `CommitToX` classes?

adngdb: Ah, that's because this is the only one I tested for, and I didn't think to look at the other classes. Thanks!

mathjazz:
I guess we should do the same for SVN, then?

What about `PullFromX.pull()`?

adngdb:
@mathjazz this is a lot better tested now. I was able to sync all files from `pontoon-test` and write to all file formats as well — see https://github.com/adngdb/pontoon-test. Can you review again?

PR #1512, opened 2019-12-04
Fix bug 1599056: Do not allow newlines in PO translations

mathjazz:


pontoon/checks/libraries/__init__.py
```diff
@@ -68,10 +68,16 @@ def run_checks(
                     'singlequoting',
                     'startwhitespace',
                 ])
+        # We override newlines check for lang and po files in
+        # pontoon.checks.libraries.pontoon_db
         elif resource_ext == 'lang':
             tt_disabled_checks.update([
                 'newlines',
             ])
+        elif resource_ext == 'po':
```
adngdb: How about merging the two blocks? 
```python
elif resource_ext in ('lang', 'po'):
```

Pike:
The caveat about the TT check is that it also enforces in-content differences in newlines, not only end-of-content ones. Which is what we need here, AFAICT.

mathjazz:
> The caveat about the TT check is that it also enforces in-content differences in newlines, not only end-of-content ones. Which is what we need here, AFAICT.

Oh, wow. This patch is completely wrong then! I though newlines are not allowed at all.

[Instead](https://www.gnu.org/software/gettext/FAQ.html#newline), we need to make sure the translations ends or doesn't end with the newline, depending on what the source string does. Axel, can you confirm that?

Pike:
That's how I read the docs, but they're vague, to say the least, https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html, at the very bottom

```po
msgid ""
"Here is an example of how one might continue a very long string\n"
"for the common case the string represents multi-line output.\n"
```

mathjazz:
@Pike does https://github.com/mozilla/pontoon/pull/1512/files#diff-3b42f6d8a5b9478290a0658743092871R69 make sense?

mathjazz:
@Pike Any updates?

pontoon/checks/libraries/pontoon_db.py
```diff
@@ -65,6 +66,17 @@ def run_checks(entity, string):
                     'Translation too long'
                 )
 
+    # Bug 1599056: Original and translation must either both end in a newline,
+    # or none of them should.
+    if resource_ext == 'po':
+        if (
+            original.endswith('\n') and not string.endswith('\n') or
+            not original.endswith('\n') and string.endswith('\n')
```
Pike: I think this is easier written as `original.endswith('\n') == string.endswith('\n')`?

mathjazz: Ohhhhhhhh. :-) `!=` instead of `==`, but yeah. Much easier.

PR #1513, opened 2019-12-05
Fix bug 1601002: Show full-width input if no accesskey candidates available

mathjazz:
Should we also generate accesskeys from the attribute positioned right before the "accesskey" attribute if none of the attributes is called "label"?

That would cover cases like https://github.com/mozilla-l10n/pontoon-ftl/commit/5a67087243f6017f4bf5c66aacda30f0b5ec80c4.

### DIFF
--- a/frontend/src/modules/fluenteditor/components/RichTranslationForm.js
+++ b/frontend/src/modules/fluenteditor/components/RichTranslationForm.js
@@ -345,6 +345,7 @@ export default class RichTranslationForm extends React.Component<EditorProps> {
             candidates = fluent.extractAccessKeyCandidates(message);
         }
 
+        // Access Key UI
         if (candidates && candidates.length) {
             this.accessKeyElementId = path.join('-');
             return <td>
@@ -352,6 +353,8 @@ export default class RichTranslationForm extends React.Component<EditorProps> {
                 { this.renderAccessKeys(candidates) }
             </td>;
         }
+
+        // Default UI
         else {
             return <td>
                 { this.renderTextarea(value, path) }
--- a/frontend/src/modules/fluenteditor/components/RichTranslationForm.js
+++ b/frontend/src/modules/fluenteditor/components/RichTranslationForm.js
@@ -288,7 +288,7 @@ export default class RichTranslationForm extends React.Component<EditorProps> {
         );
     }
 
-    renderInput(value: string, path: MessagePath, maxlength: ?number) {
+    renderTextarea(value: string, path: MessagePath, maxlength: ?number) {
         return <textarea
             id={ `${path.join('-')}` }
             key={ `${path.join('-')}` }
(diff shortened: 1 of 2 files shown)


Pike:
Let me comment and ask a bit, 'cause I'm unsure what this does.

Here's what I think the criteria is:

The en-US accesskey is non-trivial, or the l10n accesskey is non-trivial.

The way I read the patch is that it triggers if the localization is just a reference or placeable?

What happens with a `PLATFORM` select expression in an accesskey right now?

Also, given the en-US accesskey was non-trivial, but l10n is, and the localizer presses the message reference in en-US to copy over to the accesskey, what would happen?

mathjazz:
> Here's what I think the criteria is:
> 
> The en-US accesskey is non-trivial, or the l10n accesskey is non-trivial.
> 
> The way I read the patch is that it triggers if the localization is just a reference or placeable?

So the bug we're fixing is that the input for the `accesskey` attribute is too short if its value isn't trivial. The patch thus more strictly splits the Default UI (regular `<textarea>`) and the Access Key UI (`<textarea maxlength="1">` with candidate keys).

Currently, if we [cannot generate candidate keys](https://github.com/mozilla/pontoon/blob/de490f70ebed9586a3f8686197ff64cb3de3b4bf/frontend/src/core/utils/fluent/extractAccessKeyCandidates.test.js#L52), we still show the `<textarea maxlength="1">` as long as the attribute ID is `accesskey`. The patch OTOH requires at least one candidate being present before adding the `maxlength=1` attribute.

> What happens with a `PLATFORM` select expression in an accesskey right now?

Default UI is used:
https://pontoon.mozilla.org/sl/firefox/all-resources/?search=PLATFORM&string=200969

If `PLATFORM` is only used in label, Access Key UI is used:
https://pontoon.mozilla.org/sl/firefox/all-resources/?search=PLATFORM&string=200883

> Also, given the en-US accesskey was non-trivial, but l10n is, and the localizer presses the message reference in en-US to copy over to the accesskey, what would happen?

Currently: Access Key UI is used.
In the patch: Default UI is used.

Pike:
I think I know what this patch does, let me try to break it w/out actually having to apply the patch locally ;-)

Let's assume this reference:

```
jazz = Something
    .accesskey = {funk.accesskey}
```

A couple of scenarios:

There's no translation yet. This patch goes back to default UI, as there's no candidates?

There's a translation, 

```
jazz = Irgendwas
    .accesskey = {funk.accesskey}
```

so now there are candidates, and it shows `{` in the accesskey UI?

mathjazz:
Right - you'd need to switch to source view or use cursor to reveal the entire accesskey in the 2nd example. Is the 2nd scenario realistic, though? I haven't seen it in bug 1600528.

If we need to account for that scenario, I don't know how to fix it without compromising on the most common scenario too much. :/

mathjazz:
I've added another requirement for using the Access Key UI (in addition to access key candidates being present): `.accesskey` translation must not be longer in length than 1 character.

I you want to submit an `.accesskey` translation that is longer, you either need to enter it before translating the label or use the source editor. That's because such scenario is very unlikely and because I'd rather keep the benefit of `maxlength=1` on the input field for the most common scenario.

PR #1515, opened 2019-12-09
Fix bug 1601564: Allows reading dotenv config in WSGI application

flozz:
This PR allows to use a dotenv file to configure Pontoon when deployed as a WSGI app. The path of the dotenv file can be configured through the `DOTENV_PATH` environment variable.

To test this PR, you can do the following:

1. Write a dotenv file with the Pontoon configuration in it,
2. install gunicorn or any other WSGI server (`pip install gunicorn`)
3. run the application (`DOTENV_PATH=./dotenv.conf gunicorn pontoon.wsgi:application`)

Pontoon should have its config read from the file.

### DIFF
--- a/docs/admin/deployment.rst
+++ b/docs/admin/deployment.rst
@@ -45,6 +45,20 @@ Environment Variables
 The following is a list of environment variables you'll want to set on the app
 you create:
 
+.. NOTE::
+
+   Alternatively, you can put all variables below in a `dotenv
+   <https://github.com/theskumar/python-dotenv>`_ text file::
+
+      VAR="value 1"
+      OTHER_VAR="other value"
+
+   If you do so, you will only have to give the path of this file to Pontoon
+   through the ``DOTENV_PATH`` environment variable::
+
+      DOTENV_PATH=/path/to/my/config.env
+
+
 ``ADMIN_EMAIL``
    Optional. Email address for the ``ADMINS`` setting.
 
--- a/pontoon/wsgi.py
+++ b/pontoon/wsgi.py
@@ -8,9 +8,13 @@ from __future__ import absolute_import
 
 import os
 
+import dotenv
 from django.core.wsgi import get_wsgi_application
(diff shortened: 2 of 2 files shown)


adngdb:
Hi @flozz, thank you for your contribution! Could you please document this new feature in https://github.com/mozilla/pontoon/blob/master/docs/admin/deployment.rst#environment-variables ?

flozz:
Done. But please double check my english 😅️

paolodirollo:
Hi Guys,

I becoming crazy trying to pass a config.env file to the docker version of Pontoon. 

This is from your documentation:

"
If you do so, you will only have to give the path of this file to Pontoon through the DOTENV_PATH environment variable:
DOTENV_PATH=/path/to/my/config.env
"

I added in the config.env file the SSH_KEY and the SSH_CONFIG variables and I put this file in the pontoon root directory. Now how can initialize Pontoon in order to take in consideration this file? I did a lot of test but without any success. When I build the software and run it, if I try to do a sync with git the system always ask for a git account and password so I imagine that Pontoon is not taking care of the config.env file.

I'm running Pontoon from a Google Cloud Compute Engine with Ubuntu, where I installed docker. To start the project I'm just using this command: "make run".

Please give me a suggestion on of to move ahead, I'm new with Dijango and Python....

jotes:
Hey @paolodirollo,

I think that an important thing is to mention here that `make run` starts a development version of Pontoon and shouldn't be used as something that's exposed to the real user traffic.

Having said that, I'm going to give you a few hints on what's the possible root of your problem. Pontoon's sync process runs shell commands internally, so they use the environment variables not from the dot env file, but from the environment on which the Pontoon's instance process runs.
So, in your case, you should probably add those environment variables to the Pontoon's docker-compose configuration.

I think that our `k8s` setup does a similar thing, you [can check it out](https://github.com/mozilla/pontoon/blob/master/docker/k8s-first-steps/k8s-pontoon-example.yaml).

PR #1516, opened 2019-12-10
Fix bug 1602732: GitLab authentication method

flozz:
This PR adds the support of the GitLab authentication method to Pontoon.

To try it out, you will need a GitLab instance (you can probably use gitlab.com).

1. Set following environment variable: 
  * `AUTHENTICATION_METHOD=gitlab`
  * `GITLAB_URL=https://gitlab.com`

2. Follow the instructions of [this article](https://www.wanadev.fr/196-pontoon-sauthentifier-avec-gitlab/) to setup the applications, but skip the part about patching Pontoon as this patch is provided by this PR (the article is in French but [Google Translate should help](https://translate.google.com/translate?sl=auto&tl=en&u=https%3A%2F%2Fwww.wanadev.fr%2F196-pontoon-sauthentifier-avec-gitlab%2F)). If you do not have an admin account on the GitLab server (on gitlab.com for example), you may be able to [create an app](https://docs.gitlab.com/ce/integration/oauth_provider.html) linked to your own account. 

3. You should now be able to sign in using a GitLab account.

:)

### DIFF
--- a/docs/admin/deployment.rst
+++ b/docs/admin/deployment.rst
@@ -69,7 +69,7 @@ you create:
    The default value is `django`, which allows you to log in via accounts created using `manage.py shell`.
    Set to 'fxa' if you want to use 'Firefox Accounts' (corresponding FXA_* settings must be set).
    Set to 'github' if you want to use 'GitHub' (corresponding GITHUB_* settings must be set).
-   Set to 'gitlab' if you want to yse 'GitLab' (corresponding GITLAB_* settings must be set if required).
+   Set to 'gitlab' if you want to use 'GitLab' (corresponding GITLAB_* settings must be set if required).
    Set to 'google' if you want to use 'Google' (corresponding GOOGLE_* settings must be set).
 
 ``CELERY_ALWAYS_EAGER``
@@ -108,7 +108,7 @@ you create:
    a system error. See `Heroku Reference`_ for more information.
 
 ``GITLAB_URL``
-   Optional. URL of your GitLab instace when ``AUTHENTICATION_METHOD=gitlab``.
+   Optional. URL of your GitLab instance when ``AUTHENTICATION_METHOD=gitlab``.
    If not defined, ``https://gitlab.com`` will be used by default.
 
 ``GOOGLE_ANALYTICS_KEY``
--- a/docs/admin/deployment.rst
+++ b/docs/admin/deployment.rst
@@ -69,6 +69,7 @@ you create:
    The default value is `django`, which allows you to log in via accounts created using `manage.py shell`.
    Set to 'fxa' if you want to use 'Firefox Accounts' (corresponding FXA_* settings must be set).
    Set to 'github' if you want to use 'GitHub' (corresponding GITHUB_* settings must be set).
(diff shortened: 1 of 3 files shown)


docs/admin/deployment.rst
```diff
@@ -69,6 +69,7 @@ you create:
    The default value is `django`, which allows you to log in via accounts created using `manage.py shell`.
    Set to 'fxa' if you want to use 'Firefox Accounts' (corresponding FXA_* settings must be set).
    Set to 'github' if you want to use 'GitHub' (corresponding GITHUB_* settings must be set).
+   Set to 'gitlab' if you want to yse 'GitLab' (corresponding GITLAB_* settings must be set if required).
```
adngdb: `use` instead of `yse` ;)

docs/admin/deployment.rst
```diff
@@ -106,6 +107,10 @@ you create:
    Optional. URL to the page displayed to your users when the application encounters
    a system error. See `Heroku Reference`_ for more information.
 
+``GITLAB_URL``
+   Optional. URL of your GitLab instace when ``AUTHENTICATION_METHOD=gitlab``.
```
adngdb: `instance` instead of `instace`.

docs/admin/deployment.rst
```diff
@@ -106,6 +107,10 @@ you create:
...
+   If not defined, ``https://gitlab.com`` will be used by default.
+
```
mathjazz: I'd leave that out as we don't have docs for other provider-specific variables.

pontoon/settings/base.py
```diff
@@ -827,7 +828,11 @@ def account_username(user):
         'SCOPE': FXA_SCOPE,
         'OAUTH_ENDPOINT': FXA_OAUTH_ENDPOINT,
         'PROFILE_ENDPOINT': FXA_PROFILE_ENDPOINT,
-    }
+    },
+    'gitlab': {
+        'GITLAB_URL': os.environ.get('GITLAB_URL', 'https://gitlab.com'),
```
mathjazz: Nit: Let's keep this consistent with other providers, i.e.:
```python
...
# GitLab
GITLAB_URL = os.environ.get('GITLAB_URL', 'https://gitlab.com')
...
'gitlab': {
    'GITLAB_URL': GITLAB_URL,
...
```

docs/admin/deployment.rst
flozz: Hum,

* Google: does not seems to require any additional env var to work (only some config in the Django backoffice),

* Github: the allauth doc say that the URL can be configured to support Github Entreprise but this is not supported by the current implementation in Pontoon so there is no specific env var for this provider too.

* FXA: can be configured using env var, and... this is not documented... Maybe we should add some documentation for the FXA env vars instead?

flozz:
@adngdb If you cannot test using Framagit, I can create an account on our Gitlab for you ;)

docs/admin/deployment.rst
flozz: Oops, didn't saw the vars for google and github, I will look at this tomorrow  ^^'

PR #1517, opened 2019-12-11
Fix bug 1600134: Refactor get_translations_from_other_locales()

mathjazz:
This is a continuation of #1508, which we closed due to a bad rebase.

* Make API return separate lists of translations: from preferred and other locales
* No longer pass preferred locales as part of user data
* Show original string in Other Locales if alternative source locale is used


mathjazz:
Great work @abowler2! This will not only make the code much more manageable, but also allow people to finally use alternative source languages.

PR #1518, opened 2019-12-11
Fixed Bug 1602374: Columns should be sorting descending with first click

abowler2:
Priority and Unreviewed columns for projects had an initial ascending sort order.

This fix reverses that so the priority and unreviewed columns now sort by descending order.

### DIFF
--- a/pontoon/base/static/css/style.css
+++ b/pontoon/base/static/css/style.css
@@ -187,16 +187,28 @@ th {
   position: absolute;
 }
 
-.table-sort th.asc i:after {
+.table-sort th.sort-order i:after {
   content: "";
   display: inline-block;
   margin-top: 5px;
 }
 
-.table-sort th.desc i:after {
+.table-sort th.rev-order i:after {
   content: "";
   display: inline-block;
   margin-top: -5px;
+}
+
+.table-sort th.priority.sort-order i:after {
+  content: "";
+  display: inline-block;
+  margin-top: -5px;
+}
+
+.table-sort th.priority.rev-order i:after {
+  content: "";
+  display: inline-block;
+  margin-top: 5px;
 }
 
 td {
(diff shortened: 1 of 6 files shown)


pontoon/base/static/css/style.css
```diff
@@ -187,18 +187,30 @@ th {
...
+  content: "";
+  display: inline-block;
+  margin-top: -5px;
+}
+
+.table-sort th.priority.rev-order i:after {
+  content: "";
+  display: inline-block;
+  margin-top: 5px;
+}
```
mathjazz: I guess you `display` and `margin-top` don't need to be repeated.

And you also need to add a selector for the `.unreviewed-status`.

pontoon/base/static/css/style.css
```diff
@@ -187,18 +187,30 @@ th {
   position: absolute;
 }
 
-.table-sort th.asc i:after {
+.table-sort th.sort-order i:after {
```
mathjazz: Maybe `default-order` and `reverse-order` is more accurate?

pontoon/base/static/css/style.css
abowler2: If I remove those then the alignment for the arrow is off. 

So, I opted to add a new class (`desc-sort`) to those columns that need to be in descending order as default, and then I added  a selector for that class. I think this will make it more future proof too.

mathjazz:
Everything works great now, thanks! But I have a naming suggestion, as I usually do. :)

I'd keep the old class names, as they describe what the CSS rules actually do:
`default-order` -> `asc`
`reverse-order` -> `desc`
`desc-sort` -> `inverted`

And a nit: I'd make `desc-sort` the 2nd (not 1st) class name in the CSS selector, i.e. in class names suggested above: `.desc.inverted`.

PR #1519, opened 2019-12-12
Use Python 3.8 with heroku

Natim:


adngdb:
Merci Rémy ! We're not ready yet to use Python 3 though, but I'm really close to being done with all my testing. There are still some potential issues with sync that I need to test, but it's coming really soon. :-)

You can follow the state of this process here: https://wiki.mozilla.org/L10n:Pontoon#Upgrade_to_python_3.6

PR #1520, opened 2019-12-13
No bug - Upgrade react-scripts to latest version.

adngdb:


adngdb:
Hold on, I think this is breaking the `make test-frontend` command. I need to fix that first.

adngdb:
I just spent a few hours trying to figure out why `make test-frontend` doesn't work anymore. I failed. Here's what I found: 

The problem only happens inside our docker container. It works if you run `jest` directly in the container (but tests fail, because of missing babel transpiling). It fails only if you run `react-scripts test`, or react-scripts' `test.js` file directly. It seems it's the config that this script passes to `jest` that causes the issue, but that config is complex and I have been unable to identify the source of the problem. Maybe there's an integration issue with babel? Maybe not? Maybe I should go fork myself? 

Anyway, I forfeit, and will replace the `--watch` option with `--watchAll`. It's not as convenient as `--watch` but at least it works.

adngdb:
@mathjazz you can review now. :-)

docker/Dockerfile
```diff
@@ -57,7 +57,7 @@ RUN npm install
 RUN mkdir frontend
 WORKDIR frontend
 COPY frontend/package.json .
-COPY frontend/yarn.lock .
+COPY --chown=pontoon:pontoon frontend/yarn.lock .
```
adngdb: I have no idea why, but docker stopped building without this. 

Side-note: I'm starting to hate create-react-app more and more…

PR #1521, opened 2019-12-13
Fix bug 1602973 - Make sure users cannot send the same translation twice.

adngdb:
This prevents users from spamming the "save" button, which if done fast enough can lead to the same translation being sent several times. Now as soon as the user clicks the Save button, it becomes disabled until the `sendTranslation` action is entirely finished. I am also using this opportunity to introduce spinners and label update on buttons, as a proposal. It is a pattern that I like, so I would like to generalize it on the app.

TODO:
- [ ] File a bug to replace `withActionsDisabled` with this kind of approach.

### DIFF
--- a/frontend/public/static/locale/en-US/translate.ftl
+++ b/frontend/public/static/locale/en-US/translate.ftl
@@ -94,8 +94,12 @@ editor-EditorMenu--button-approve = Approve
     .title = Approve Translation (Enter)
 editor-EditorMenu--button-save = Save
     .title = Save Translation (Enter)
+editor-EditorMenu--button-saving = Saving
+    .title = Saving Translation…
 editor-EditorMenu--button-suggest = Suggest
     .title = Suggest Translation (Enter)
+editor-EditorMenu--button-suggesting = Suggesting
+    .title = Suggesting Translation…
 
 
 ## Editor Settings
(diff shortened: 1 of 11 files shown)


mathjazz:
The tests are failing.

adngdb:
> The code looks good, I wonder however if we should use `withActionsDisabled`, which does the same thing in a few other places.

The way `withActionsDisabled` works is prone to bugs. This PR does it the "proper" way, because it makes sure that no actions can be taken until the server is done responding (where `withActionsDisabled` only waits for the next re-render I think? ). This is also much cleaner, and a lot easier to understand (to prove it, I already don't remember how `withActionsDisabled`  exactly works :/ ). 

My plan with this was to show it to you, get your approval, then slowly replace all traces of `withActionsDisabled` with something like this. :smirk: 

I think we should just ship this indeed, and improve the rest of the app later on.

mathjazz:
I was hoping for such answer. Thanks! :)

mathjazz:
@adngdb Seems like the spinner no longer appears?

I've also notice two other annoyances:
1. We don't change the text to Approving when the button reads Approve.
1. If you are reviewing strings, the button will jump between Approve and Save as you move between strings with Enter and as results are fetched from History.

PR #1524, opened 2019-12-18
Fix Bug 1603610 - Enable users to leave comments on translations

abowler2:
Initial work to add the ability to leave comments on translations.

___________________
At the moment I'm just printing the comments into the history panel. Moving forward from here I believe I will need to add a Comment component in order to render and style all the information properly. If that is indeed the case should I add that component to the History module or the Translation module?

Also, I'm not sure if a Prefetch is needed for retrieving the comments in `map_entities` within `models.py`, but I was unsuccessful in my attempts at that. Would appreciate your thoughts and insights on this.

TODO:
- [ ] [File a bug](https://github.com/mozilla/pontoon/pull/1524#discussion_r360291627) to rename Translation.user to Translation.author.
- [ ] File a bug to make the Add comment fields [grow as you type](https://www.npmjs.com/package/react-textarea-autosize).
- [x] `./manage.py makemigrations` says we need to make some migrations.
- [x] Use `UserAvatar` component wherever possible.


frontend/src/core/api/types.js
```diff
@@ -32,6 +43,7 @@ export type Entity = {|
     +project: Object,
     +source: Array<Array<string>> | Object,
     +translation: Array<EntityTranslation>,
+    +translation_comments: Array<TranslationComments>,
```
mathjazz: I'd move this property under `HistoryTranslation` and rename it to `comments`. The comments data should be loaded and stored together with the rest of the history panel data.

pontoon/base/models.py
```diff
@@ -2640,6 +2646,7 @@ def map_entities(cls, locale, preferred_source_locale, entities, visible_entitie
                 'source': entity.source,
                 'obsolete': entity.obsolete,
                 'translation': translation_array,
+                'translation_comments': translation_comments_array,
```
mathjazz: Same as on frontend: this should be moved to `pontoon.base.views:get_translation_history()`.

pontoon/base/models.py
```diff
@@ -3425,3 +3432,13 @@ def calculate_stats(self, save=True):
...
+
+
+class Comment(models.Model):
+    author = models.ForeignKey(User)
+    timestamp = models.DateTimeField(default=timezone.now)
+    translation = models.ForeignKey(Translation)
+    comment = models.TextField(blank=True)
+
+    def __unicode__(self):
+        return self.translation.string
```
mathjazz: Why are you rendering a translation?

pontoon/base/models.py
```diff
@@ -3425,3 +3432,13 @@ def calculate_stats(self, save=True):
             strings_with_warnings_diff,
             unreviewed_strings_diff,
         )
+
+
+class Comment(models.Model):
+    author = models.ForeignKey(User)
+    timestamp = models.DateTimeField(default=timezone.now)
+    translation = models.ForeignKey(Translation)
+    comment = models.TextField(blank=True)
```
mathjazz: As noted above, we should rename this to `content`.

Two questions for @adngdb:
1. Should we add fields needed for Team Comments now or when we need them? I'm leaning towards the later. Anyhow, I think we could make use of your foreign key validation logic from #1503.
2. Should we make comments a separate pontoon app?

frontend/src/modules/history/components/Translation.js
```diff
@@ -414,6 +414,11 @@ export class TranslationBase extends React.Component<InternalProps, State> {
                             format={ entity.format }
                         />
                     </p>
+                    <p>
+                        { entity.translation_comments.map(comment => {
```
mathjazz: This should be `translation.comments`...

pontoon/base/models.py
abowler2: This was actually another question I had and forgot to add. When accessing the model from `admin` this is what displays, without this it defaults to `Comment object` for all comments.  I went back and forth on whether this should be the translation or the comment, and settled on the translation since I figured that would be the reference that most would look for when accessing through the admin panel.

frontend/src/core/api/types.js
```diff
@@ -14,6 +14,17 @@ export type EntityTranslation = {|
 |};
 
 
+/**
+ * Comments pertaining to a translation.
+ */
+export type TranslationComments = {|
+    +author_id: number,
+    +comment: ?string,
```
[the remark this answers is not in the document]
abowler2: Probably not. I think I was trying to guard against making it required, but if someone selects to leave a comment then they shouldn't be able to send a blank one.  I'll make the changes.

frontend/src/modules/history/components/Translation.js
```diff
@@ -414,6 +414,11 @@ export class TranslationBase extends React.Component<InternalProps, State> {
...
+                            return comment.comment
```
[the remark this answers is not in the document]
abowler2: Do you prefer that I make a new migration with the changes or would you rather I roll back this migration, make changes, and then re-migrate?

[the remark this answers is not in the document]
mathjazz: The 2nd option would be ideal. Make sure you revert the migration first. Since this can be slightly trickier, I'm also fine with option 1.

pontoon/base/admin.py
```diff
@@ -204,6 +204,10 @@ class TranslationAdmin(admin.ModelAdmin):
     raw_id_fields = ('entity',)
 
 
+class CommentAdmin(admin.ModelAdmin):
+    row_id_fields = ('comment',)
```
mathjazz: r**a**w_id_fields

pontoon/base/models.py
mathjazz: I'd prefer `self.content` to make it consistent with the `Entity` and `Translation` admin.

abowler2:
Current progress for review. There is obviously still a lot to be done but wanted to check in to make sure all is on track (Note: I used some temporary filler values for titles and inline styling in the components that is not intended to stay 😄 ) 

In addition to the obvious need for styling, I still have to add the Prefetch in `get_translation_history`, the `Delete` and `Share` buttons in the `Comment` component, and build out the `AddComment` component. 

Please let me know what changes are needed at this point.

pontoon/base/models.py
```diff
@@ -3425,3 +3425,21 @@ def calculate_stats(self, save=True):
             strings_with_warnings_diff,
             unreviewed_strings_diff,
         )
+
+
+class Comment(models.Model):
+    author = models.ForeignKey(User)
```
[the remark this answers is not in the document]
adngdb: `author` is a better name in both contexts, so I'd go for the bug to rename it in `Translation`.

pontoon/base/models.py
```diff
@@ -3425,3 +3425,21 @@ def calculate_stats(self, save=True):
...
+    timestamp = models.DateTimeField(default=timezone.now)
```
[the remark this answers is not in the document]
adngdb: … and `date` is a terrible name in my opinion! :D It's too generic, and can conflict with reserved keywords in languages. I'd rather use something like `created_at`.

frontend/src/core/api/types.js
```diff
@@ -14,6 +14,21 @@ export type EntityTranslation = {|
...
+ */
+export type TranslationComment = {|
+    +author: string,
+    +author_id: number,
+    +username: string,
+    +user_gravatar_url_small: string,
+    +timestamp: string,
+    +date_iso: string,
+    +content: string,
+    +translation: number,
```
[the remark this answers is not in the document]
adngdb: (Hah, I now realize that you haven't implemented the `api` yet :P So please do change this type, and when you add the api method, please make sure to map the data to these names. )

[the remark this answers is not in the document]
mathjazz: BTW, this patch only implements translation comments, which will be loaded as part of the existing `getHistory()` API call. Team comments (the ones in the sidebar) will be implemented in a followup patch (and will require a new API call).

Since we'll need this type for team comments as well, I think it's fine to keep it here (as opposed to history reducer), but I'd rename it to just `Comment`.

pontoon/base/views.py
```diff
@@ -396,6 +413,7 @@ def get_translation_history(request):
             "date_iso": t.date.isoformat(),
             "approved_user": User.display_name_or_blank(t.approved_user),
             "unapproved_user": User.display_name_or_blank(t.unapproved_user),
+            "comments": translation_comments_array,
```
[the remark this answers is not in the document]
adngdb: OK so I'm dumb and I didn't see that @mathjazz had said it all before me. Sorry for making this more confusing! 

You can indeed just do what Matjaz said, and it will work. You should use [`prefetch_related`](https://docs.djangoproject.com/en/1.11/ref/models/querysets/#django.db.models.query.QuerySet.prefetch_related) though to improve query performance, and you *can* set `related_name` to use a different name than `comment_set` but you don't have to.

frontend/src/core/translation/components/Comment.js
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    render() {
+        const { comment } = this.props;
+
+        if (!comment) {
+            return null;
+        }
+
+        var divStyle = { display: 'inline' }
```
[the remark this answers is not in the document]
abowler2: Yes 😁

frontend/src/core/api/types.js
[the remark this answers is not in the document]
abowler2: Renaming from `TranslationComment` to `Comment` in `types` causes conflicts do to the `Comment` component.  Maybe something like `CommentDetails` could work??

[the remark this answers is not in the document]
adngdb: I'd go for `CommentType`, I think. I remember resolving such a conflict by calling the type `DbThing` instead of `Thing` but that wasn't great and I have been able to rename that recently. 

Note though that this type here has a `translation` key that will **not** be in the data for other types of comments. So maybe it's just better to keep this as `TranslationComment` for now, and refactor later on?

frontend/src/core/translation/components/Comment.js
```diff
@@ -0,0 +1,88 @@
...
+import { WithPlaceables } from 'core/placeable';
+import type { TranslationComment } from 'core/api';
+
+
+type Props = {|
+    comment: TranslationComment,
+|};
+
+
+export default class Comment extends React.Component<Props> {
```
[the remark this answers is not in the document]
abowler2: So I should also make `AddComment` functional, correct?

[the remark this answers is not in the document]
mathjazz: Correct!

frontend/src/core/translation/components/Comment.js
```diff
@@ -0,0 +1,88 @@
...
+    renderUserAvatar() {
+        const { comment } = this.props;
+
+        if (!comment.author_id) {
+            return <img
+                src='/static/img/anon.jpg'
+                alt=''
+                height='44'
+                width='44'
+            />;
```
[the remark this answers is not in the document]
abowler2: Can  you please clarify what you mean by "factor this out"?  Should it be a functional component and if so under which file?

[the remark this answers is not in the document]
mathjazz: Initially I meant a `renderUserImg()` function, but now that we decided to turn all `renderX` method in components, I'd suggest a component.

Probably in `core/user/components`, so that we can reuse it in `history`'s `Translation.js`, `UserMenu.js`, `FiltersPanel.js` and later on in Team Comments.

pontoon/base/views.py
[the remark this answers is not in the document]
abowler2: To clarify:  This is the array from `line:396` above which has been serialized and updated with the needed values (at least that is what I thought I did 😀 )

I'm aware of the need to prefetch this data, but am still fuzzy on how to do this and so will be continuing to research on this.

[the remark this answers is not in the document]
mathjazz: Yeah, that array and the entire `for` loop on top of it can go away if you use this: :)
https://github.com/mozilla/pontoon/pull/1524/files/cba6a6eb3f1d7344451d49d1b24173fad1b5cb2a#r360300975.

To prefetch, you  just need to add `.prefetch_related('comments')` to the existing `Translation.objects.filter...` query. To make it work, you need to set the [`related_name`](https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.ForeignKey.related_name) on the ForeignKey, though.

abowler2:
Made the requested changes. Please let me know if any further changes are needed.

RE: the failing `Translation` test ~ I refactored the user's avatar to use the new `UserImage` component which causes the test to fail based on only one `a` tag being present now. Not sure if I should adjust the test to account for the change or if there is something I missed when using the new component.

pontoon/base/models.py
```diff
@@ -3430,16 +3430,19 @@ def calculate_stats(self, save=True):
...
 
     def serialize(self):
         return {
-            'author': self.author_id,
-            'timestamp': self.timestamp,
-            'content': self.content,
-            'translation': self.translation_id
+            "author": self.author.name_or_email,
+            "username": self.author.username,
+            "userGravatarUrlSmall": self.author.gravatar_url(88),
```
adngdb: Ah, here's the trick: Python variables should use `underscore_case`. :) This is confusing, sorry. We effectively have two different languages with different naming conventions. We want to strictly enforce those naming conventions. The only time when they conflict is through the API. And that's where we need some specific code to manually convert those names.

So, here you should revert this change. But keep your type definition in JS. Then on the API function, in front-end, where you get the data, you should have some logic to transform the data, mapping one convention (underscore_case from Python) to the other (camelCase for JS).

Here's where the data gets fetched: https://github.com/mozilla/pontoon/blob/6cc116bfff32d1c9ac7133535d4cb8055074b5a2/frontend/src/core/api/entity.js#L115

_Side note: if this data was exposed with GraphQL, we wouldn't have that problem, as it handles that conversion automatically for us. Sadly, this is not the case yet, but doing this conversion now means it will be easier later on to switch to fetching that data through GraphQL._

frontend/src/core/comments/components/Comment.js
```diff
@@ -0,0 +1,54 @@
...
+|};
+
+
+export default function Comment(props: Props) {
+    const { comment } = props;
+
+    function renderUser() {
+        return <a
+            href={ `/contributors/${comment.username}` }
+            title='Author'
```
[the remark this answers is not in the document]
abowler2: As per the discussion this morning, I no longer have a title for this or UserAvatar.

abowler2:
Sending up the updates from the last review. I'm planning to work on the `delete` and `share` button functionality next and then styling the elements. 

The one thing I did have issues with was the `Copy to Translation` title which displays on hover that we discussed last week. It's being displayed from the Localization tag, and I was not able to successfully change it up without breaking it. I figured I would work on it more when I begin the styling, but your thoughts on how I can approach it are always welcome 😄

frontend/src/core/comments/components/Comment.test.js
```diff
@@ -0,0 +1,36 @@
...
+        content: 'Comment for testing',
+        translation: 0,
+        id: 1,
+    }
+
+    describe('renderUser', () => {
+        it('returns a link for the author', () => {
+            const comments = {
+                ...DEFAULT_COMMENTS,
+                ...{ username: 'id_Sarevok', author: 'Sarevok' }
```
adngdb: One thing that is **extremely** important with tests is to leave your mark. Here, I know that you copied a test I wrote because I'm a *Baldur's Gate* fan and Sarevok is a character from that game. It is important that you make these tests you write truly **yours**. This is the place where you are absolutely free to do whatever you want, so you should use that opportunity. Surprise us! Our selves from the future will be glad to enjoy a smile or a laugh when they read these tests again.

(This is, of course, not serious. You don't have to do anything special, and using whatever fake content you want is fine. Buuuut it's true that this is the place where you can go a little crazy. :wink: )

frontend/src/core/comments/components/Comment.test.js
```diff
@@ -0,0 +1,36 @@
...
+        username: '',
+        userGravatarUrlSmall: '',
+        createdAt: '',
+        dateIso: '',
+        content: 'Comment for testing',
+        translation: 0,
+        id: 1,
+    }
+
+    describe('renderUser', () => {
```
adngdb: When testing a small component like this one you can skip this level of abstraction and put all your tests in the top-level `describe`.

frontend/src/core/comments/components/Comment.test.js
```diff
@@ -0,0 +1,36 @@
...
+                key={ comments.id }
+            />);
+
+            const link = wrapper.find('a');
+            expect(link).toHaveLength(1);
+            expect(link.at(0).props().children).toEqual('Sarevok');
+            expect(link.at(0).props().href).toEqual('/contributors/id_Sarevok');
+        });
+    });
+});
```
adngdb: Still missing newlines at end of file. ;)

frontend/src/core/comments/components/Comment.test.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import Comment from './Comment';
+
+
+describe('<Comment>', () => {
+    const DEFAULT_COMMENTS = {
```
adngdb: This contains a single comment, it shouldn't have the `S` at the end.

frontend/src/core/api/entity.js
```diff
@@ -125,7 +125,45 @@ export default class EntityAPI extends APIBase {
...
+
+        const renamedCommentsKeys = results[0].comments.map(comment => {
+            return renameKeys(keysMap, comment);
+        });
+
+        if (renamedResultsKeys.comments) {
+            renamedResultsKeys.comments = renamedCommentsKeys;
+        }
+
+        return [renamedResultsKeys];
```
adngdb: I find this code too complex, and too specific. Once again, apologies for opening a can of worms here. Let's still get to the end of it, but by using a more generic solution that we can then apply to other places around the API.

I found this article that shows a nice solution to transform code from snake_case to camelCase: https://matthiashager.com/converting-snake-case-to-camel-case-object-keys-with-javascript

I'd take that code and put in the `api/base.js` file, add a method there to transform objects, and then use that method here to convert the results.

abowler2:
Updated the camelCase conversion as per @adngdb's note. However, I'm not sure if I 'typed' the argument that is passed in properly. Technically it is being passed an `Array<Object>` from `getHistory`, but if I type it as that I get Flow errors within the function. Also, when this is used for other api's it might not be the same.  So, I have it just as an `Object` which does not throw any errors from Flow.

pontoon/base/models.py
```diff
@@ -3425,3 +3425,25 @@ def calculate_stats(self, save=True):
...
+        return self.content
+
+    def serialize(self):
+        return {
+            "author": self.author.name_or_email,
+            "username": self.author.username,
+            "user_gravatar_url_small": self.author.gravatar_url(88),
+            "created_at": self.timestamp.strftime('%b %d, %Y %H:%M'),
+            "date_iso": self.timestamp.isoformat(),
+            'translation': self.translation_id,
```
mathjazz: Do we even need `translation`?

frontend/src/core/comments/components/Comment.js
```diff
@@ -0,0 +1,49 @@
...
+type Props = {|
+    comment: TranslationComment,
+|};
+
+
+export default function Comment(props: Props) {
+    const { comment } = props;
+
+    if (!comment) {
+            return null;
```
mathjazz: Nit: too much indent.

frontend/src/core/comments/components/CommentsList.js
```diff
@@ -0,0 +1,35 @@
...
+
+
+export default function CommentsList(props: Props) {
+    const { comments } = props;
+
+    if (!comments) {
+        return null;
+    }
+
+    return (
```
mathjazz: No need to wrap the return in brackets.

mathjazz:
> Also, when this is used for other api's it might not be the same. So, I have it just as an `Object` which does not throw any errors from Flow.

I believe this is one of those rare cases where using `Object` is actually permissible - for the reason you mentioned.

pontoon/base/models.py
abowler2: It doesn't look like we do, so I'll remove it.

abowler2:
Early styling of Comment element for review. 

One of the issues that came up with the change in structure is that the bottom border is now above the comments since they have been separated into individual `divs`. I was thinking this could be set up as a conditional class that would be placed based on the existence of comments or not. Let me know if you think a different approach would be better.

frontend/public/static/locale/en-US/translate.ftl
```diff
@@ -563,6 +563,12 @@ search-TimeRangeFilter--edit-range = <glyph></glyph>Edit Range
 search-TimeRangeFilter--save-range = Save Range
 
 
+## User Avatar
+## Shows user Avatar with alt text
+
+user-avatar-anonymous = Anonymous User
+user-avatar = User Profile
```
adngdb: These do not respect our [naming convention for l10n IDs](https://github.com/mozilla/pontoon/tree/master/frontend#adding-localized-content).

frontend/src/core/api/base.js
```diff
@@ -87,4 +87,39 @@ export default class APIBase {
...
+            .forEach((key) => {
+              newObj[toCamelCase(key)] = this.keysToCamelCase(results[key]);
+            });
+
+          return newObj;
+        } else if (isArray(results)) {
+          return results.map((i) => {
+            return this.keysToCamelCase(i);
+          });
+        }
```
adngdb: Please re-indent this block of code so that it uses 4-spaces.

frontend/src/core/api/base.js
```diff
@@ -87,4 +87,39 @@ export default class APIBase {
...
+                .replace('_', '');
+            });
+        };
+
+        const isArray = function (arr) {
+            return Array.isArray(arr);
+        };
+
+        const isObject = function (obj) {
+            return obj === Object(obj) && !isArray(obj) && typeof obj !== 'function';
```
adngdb: I think the `isArray` function is superfluous, we can do without it. You can simply use `Array.isArray` here instead.

frontend/src/core/api/base.js
```diff
@@ -87,4 +87,39 @@ export default class APIBase {
...
+        };
```
adngdb: Can you move `toCamelCase` and `isObject` to outside this method? Maybe at the top of this file? 

I'm not exactly sure how compilers optimize this, but there's a risk that these functions will be re-created every time `keysToCamelCase` is called, taking unnecessary memory and execution time. It's good practice to declare functions inside functions only when strictly needed. :-)

frontend/src/core/user/components/UserAvatar.js
```diff
@@ -13,30 +14,34 @@ export default function UserAvatar(props: Props) {
...
-        >
-            <img
-                src={ imageUrl }
-                alt='User Profile'
-                height='44'
-                width='44'
-            />
-        </a>
-    </div>
+    return <Localized id='user-avatar' attrs={{ title: true }} >
```
adngdb: In this case, what you want to localize is the `alt` attribute of the `img` tag. This `Localized` component should thus be directly around the `` tag, and you'll want to change the `attrs` prop to set `alt: true` instead.

Currently, what this is effectively doing is localizing the title of the `<a>` tag, but that title being passed from outside, you don't actually know what it contains and thus how it should be localized. It will be up to the components using `UserAvatar` to pass it a localized title attribute.

frontend/src/core/api/base.js
```diff
@@ -88,7 +88,7 @@ export default class APIBase {
         }
     }
 
-    keysToCamelCase(results) {
+    keysToCamelCase(results: Object) {
```
adngdb: `Object` is an alias of `any` and it is recommended by Flow not to use it. You can instead just use `any` here, as we don't really care what the type of value is.

frontend/src/core/api/base.js
```diff
@@ -106,7 +106,7 @@ export default class APIBase {
         };
 
         if (isObject(results)) {
-          const newObj = {};
+          const newObj: Object = {};
```
adngdb: This should not be needed, Flow should be able to guess the data type from the assignment. Is it not the case?

adngdb:

frontend/src/core/comments/components/Comment.js
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    return <li className='comment'>
+        <UserAvatar
+            user={ comment.author }
+            username={ comment.username }
+            title=''
+            imageUrl={ comment.userGravatarUrlSmall }
+        />
+        < div className='container'>
```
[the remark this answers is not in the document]
abowler2: So weird, I deleted that file and fixed the indent on a previous commit. I'm starting to think I have a syntax ghost trying to mess with me 🤣

frontend/src/core/api/base.js
abowler2: That's what I would think too, but I kept getting an error 🤷‍♀

abowler2:
> The avatars don't show up yet and the comment styling also applies to the metadata comments:

That is really weird ~ Here is what it looks like on my side: 

Screen Shot 2019-12-31 at 8 04 56 AM

frontend/src/core/api/base.js
adngdb: What's the error?

abowler2:
``` 
Cannot call await with api.entity.getHistory(...) bound to p because:
 • Either cannot call receive with content bound to translations because
   object literal [1] is incompatible with array type [2].
 • Or cannot call receive with content bound to translations because
   Promise [3] is incompatible with array type [2].

     src/modules/history/actions.js
 [2]  28│     translations: Array<Object>,
        :
      70│         // Abort all previously running requests.
      71│         await api.entity.abort();
      72│
      73│         const content = await api.entity.getHistory(entity, locale, pluralForm);
      74│         dispatch(receive(content));
      75│     }
      76│ }

     /tmp/flow/flowlib_12cd5d1e/core.js
 [3] 623│ declare class Promise<+R> {

     src/core/api/base.js
 [1] 105│             const newObj = {};

Comment (adngdb) [reply frontend/src/core/api/base.js]: Eeeew. OK. Then let's do what you did but replace `Object` with `any`. 

:grimacing: Flow :grimacing:

File: frontend/src/modules/history/components/Translation.js [hunk 34]
```diff
@@ -254,169 +228,177 @@ export class TranslationBase extends React.Component<InternalProps, State> {
...
+                                    this.state.isDiffVisible ? activeTranslation.string : null
+                                }
+                                format={ entity.format }
+                            />
+                        </p>
+                    </div>
+                </div>
+            </Localized>
+                <div>
+                    <CommentsList comments={ translation.comments } />
```
mathjazz: You can remove the surrounding `<div>` (we already have one inside the `CommentsList`) and dedent the `CommentsList` to be on the same level as `Localized`.

frontend/src/core/comments/components/Comment.css
```diff
@@ -0,0 +1,65 @@
+.history-comment {
```
mathjazz: We are likely to use this component for the Team Comments as well, so `history-comment` is not the best name. I'd suggest you add a `className` on the outer `div` element in the `CommentsList` (e.g. `comment-list`) and then change this selector to `.comment-list comment`.

frontend/src/core/api/base.js
```diff
@@ -35,6 +35,18 @@ export default class APIBase {
...
 
+    toCamelCase = (s: string) => {
+        return s.replace(/([-_][a-z])/ig, ($1) => {
+            return $1.toUpperCase()
+                .replace('-', '')
+                .replace('_', '');
+        });
+    };
+
+    isObject = function (obj: Object) {
```
mathjazz: I guess this one should also be `any` instead of `Object`.

frontend/src/core/api/base.js
```diff
@@ -87,4 +99,24 @@ export default class APIBase {
             return {};
         }
     }
+
+    keysToCamelCase(results: any) {
+        if (this.isObject(results)) {
+            const newObj: any = {};
+
+          Object.keys(results)
```
mathjazz: Indentation of `keysToCamelCase` is still off.

frontend/src/core/user/components/UserAvatar.js
```diff
@@ -0,0 +1,48 @@
...
+    title: string,
+    imageUrl: string,
+|};
+
+export default function UserAvatar(props: Props) {
+    const { user, username, title, imageUrl } = props;
+
+    if (!user) {
+        return <div className='user-avatar'>
+            <Localized id='user-UserAvatar--anon-alt-text' attrs={{ title: true }} >
```
mathjazz: s/title/alt

frontend/src/core/comments/components/Comment.js
```diff
@@ -0,0 +1,59 @@
...
+                    { comment.content }
+                </p>
+            </div>
+            <div className='info'>
+                <ReactTimeAgo
+                    dir='ltr'
+                    date={ new Date(comment.dateIso) }
+                    title={ `${comment.createdAt} UTC` }
+                />
+                { '\u2022' }
```
mathjazz: These are purely decorative, so I suggest we handle them with [CSS psuedo-classes](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes) like for the Translation metadata.

frontend/src/core/comments/components/Comment.js
```diff
@@ -0,0 +1,59 @@
...
+                <button>DELETE</button>
```
mathjazz: Let's use CSS's `text-transform: uppercase` here and below.

abowler2:
Just pushed the delete comment functionality for initial review.

Couple of known potential issues - 

- There is no api specifically for comments, so I chose to put the logic within the translation api.  This causes a bit of a naming issue in the `history/actions.js` file in that the function is named `deleteComment` which then calls `deleteComment` from `core/api/translation.js` which is also named `deleteTranslation` in order to distinguish it from the existing `delete` function. - Suggestions on naming changes welcome 😄 
-  Still need to add the check to prevent multiple click actions
-  Still need to address the Action Log

pontoon/actionlog/migrations/0002_auto_20200103_0313.py
```diff
@@ -0,0 +1,22 @@
...
+        ('actionlog', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='actionlog',
+            name='performed_by',
+            field=models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='actions', to=settings.AUTH_USER_MODEL),
+        ),
+    ]
```
adngdb: What's this? I don't understand where it's coming from (I don't see changes in the models that would justify it).

pontoon/base/migrations/0150_auto_20200103_0313.py
```diff
@@ -0,0 +1,21 @@
...
+
+    dependencies = [
+        ('base', '0149_comment'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='comment',
+            name='translation',
+            field=models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='comments', to='base.Translation'),
```
adngdb: Same here, not sure where it's coming from. Also, it would be better to just update the previous migration, that adds the comment table, rather than creating several migrations. It might be a bit more tricky for you, but for everyone else it will make a lot more sense to have only one (as we're going to squash all commits before merging).

frontend/src/core/comments/components/Comment.js
```diff
@@ -7,30 +7,39 @@ import { Localized } from '@fluent/react';
...
         return null;
     }
 
+    // Does the currently logged in user own this Comment?
+    const ownComment = (
+        user && user.username &&
+        user.username === comment.username
+    );
+
+    let canDelete = (canReview || ownComment);
```
adngdb: Eeeeww, `canReview` is a bad name, sorry :/ (it was my doing ;) ). I think what it means is `user can manage this project`, but I don't remember precisely. In the case of history translation, it kinda made sense because it was only used to tell if the user could review or not, but in this case it doesn't really work. Can you please call this something different? Not sure exactly what it should be. The goal is just to make reading this piece of code easier, but having clearer variable names.

frontend/src/core/comments/components/Comment.js
```diff
@@ -58,30 +67,19 @@ export default function Comment(props: Props) {
...
+                { !canDelete ? null :
+                    <Localized
+                        id='comments-Comment--delete-button'
+                        attrs={{ title: true }}
                     >
-                        SHARE
-                    </button>
-                </Localized>
+                        <button
+                            title='Delete'
```
adngdb: The title here is useless because it's just a repeat of the button label.

frontend/src/modules/history/actions.js
```diff
@@ -223,11 +223,30 @@ export function deleteTranslation(
     }
 }
 
+export function deleteComment(
```
adngdb: Shouldn't this be in the `comments` module?

pontoon/base/views.py
```diff
@@ -485,6 +486,28 @@ def delete_translation(request):
...
 
 
+@utils.require_AJAX
+@login_required(redirect_field_name="", login_url="/403")
+@transaction.atomic
+def delete_comment(request):
+    """Delete given comment."""
+    try:
+        c = request.POST["comment"]
+    except MultiValueDictKeyError as e:
```
adngdb: I don't think this can happen from the previous line. You probably just want to check for a missing value here, or an invalid integer.

frontend/src/core/comments/components/Comment.test.js
```diff
@@ -16,19 +17,52 @@ describe('<Comment>', () => {
...
+    const DEFAULT_USER = {
+        username: 'Leslie_Knope',
+    };
+
+    const DEFAULT_CANREVIEW = {
+        canReview: true,
+    }
+
+    it('returns a string', () => {
+        const wrapper = shallow(<p>{ DEFAULT_COMMENT.content }</p>);
```
adngdb: This is testing the `<p>` tag from JSX. I assume you want to use the `Comment` component here instead?

pontoon/actionlog/migrations/0002_auto_20200103_0313.py
abowler2: When I rebased my code after the changes from your PR adding Black I was prompted to do the migration.

adngdb: Oh, crap. Sorry about that. It's a bug, please remove these files. I'm not sure what is triggering that warning though. :/

abowler2: Will do.

frontend/src/core/comments/components/Comment.js
abowler2: In order to avoid repeating code, I passed this in as a prop from 'Translation`. Would you prefer I re-write it here with a new name?

frontend/src/core/comments/components/Comment.js
abowler2: Does it need the title for Localization though?

frontend/src/modules/history/actions.js
abowler2: The comments module does not have an `actions` or `reducer` it has been coming in through 'history', which is why this is here.

pontoon/base/views.py
abowler2: Will that change if we add the `created-at` parameter as discussed for the action log?

abowler2:
I have the action log recording the deletion of comments, but the way it is currently set up it is using the current time for the `created_at` field and does not record any identifying information such as which translation it was attached to.  

Let me know what modifications, if any, are needed for this action.

pontoon/base/views.py
```diff
@@ -480,6 +486,28 @@ def delete_translation(request):
...
+        return JsonResponse(
+            {"status": False, "message": "Bad Request: {error}".format(error=e)},
+            status=400,
+        )
+
+    comment = get_object_or_404(Comment, pk=c)
+
+    comment.delete()
+
+    log_action("comment:deleted", request.user)
```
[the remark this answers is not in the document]
adngdb: Why would we want to keep logs about things that don't exist anymore? In the case of a Translation, Entity or Locale, I don't think it makes sense, since we should never request that element ever again, and thus logs become useless.

For users, it's a different scenario, but as far as I know we can't actually properly delete users right now. If we want that to become possible, we should indeed change the on_delete property for the User foreign key to set to null, I guess.

On the comment about this line, this should not work. There should be a validation error when you delete something, because an ActionLog entry needs at least one of `translation` or `entity and locale`.

pontoon/actionlog/models.py
```diff
@@ -54,7 +56,7 @@ def validate_foreign_keys_per_action(self):
 
         if self.action_type != "translation:deleted" and (
             not self.translation or self.entity or self.locale
-        ):
+        ) and self.action_type != "comment:deleted":
```
adngdb: Aha, that's not correct. When a comment is deleted, we still want to link to the translation that is concerned. :)

pontoon/base/views.py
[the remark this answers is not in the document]
mathjazz: I guess keeping all the activity in the Project/Locale/User timelines permanently would benefit us in monitoring Project/Locale/User activity. OTOH, if I submit a translation by mistake and delete it, I wouldn't mind if traces of that get removed. :)

I wonder what @flodolo and @Pike think. Note that we log the delete action itself.

And to avoid off-topic'ing too much, I also wonder what they think about logging of comment actions (add/remove). Is it good enough if the log action is assigned to a translation (or entity-locale in the case of [team comments](https://docs.google.com/document/d/1W_MEYwtNn-_J0Uiu7gTSwc2B3oux75rjVtjGu-YoFi4/edit#), or do we want to have a comment associated with it?

pontoon/actionlog/models.py
abowler2: Yep, I no longer needed this check as I am passing in the associated translation (haven't pushed it yet as I'm finishing up some areas).

abowler2:
Made changes from previous review and initial steps for adding comment
______
Want to ensure the way I'm approaching `Add Comment` is on track before I go too much further.  Also, I'm getting a Flow error for the `input.value` regarding it possibly being a null value but all my attempts to put a check in have been unsuccessful.  Here is the Flow error: 
```
Cannot assign empty string to document.getElementById(...).value because:
 • property value is missing in HTMLElement [1].
 • property value is missing in null [2].

        src/core/comments/components/AddComment.js
         27│             return null
         28│         }
         29│         let comment=document.querySelector('#comment-input').value
         30│         document.getElementById('comment-input').value = ''
         31│         console.log(comment);
         32│     };
         33│

        /tmp/flow/flowlib_3128ff73/dom.js
 [1][2] 785│   getElementById(elementId: string): HTMLElement | null;

File: frontend/src/core/comments/components/AddComment.js [hunk 50]
```diff
@@ -0,0 +1,90 @@
...
+        <UserAvatar
+            user={ user }
+            username={ username }
+            title=''
+            imageUrl={ imageURL }
+        />
+        <form className='container'>
+            <div className='content'>
+                <Localized
+                    id='comments-AddComment--input'
```
mathjazz: These need to go to the `.ftl` file, too.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -0,0 +1,90 @@
...
+
+    return <div className='comment add-comment'>
+        <UserAvatar
+            user={ user }
+            username={ username }
+            title=''
+            imageUrl={ imageURL }
+        />
+        <form className='container'>
+            <div className='content'>
```
mathjazz: I don't think we need this `<div>`. It should be easier to style `<input>` directly.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -0,0 +1,90 @@
...
+export default function AddComments(props: Props) {
+    const { user, username, imageURL } = props;
+
+    if (!user) {
+        return null;
+    }
+
+    const submitComment = (event: SyntheticMouseEvent<HTMLButtonElement>) => {
+        event.preventDefault();
+        let comment=document.querySelector('#comment-input').value
```
mathjazz: You should use a [Ref](https://reactjs.org/docs/refs-and-the-dom.html#adding-a-ref-to-a-dom-element) instead. Check out how we do that in [Machinery](https://github.com/mozilla/pontoon/blob/master/frontend/src/modules/machinery/components/Machinery.js).

frontend/src/core/comments/components/AddComment.js
```diff
@@ -0,0 +1,90 @@
...
+            <div className='options'>
+                <Localized id='comments-AddComment--comment-button'>
+                    <button
+                        className='comment-btn'
+                        onClick={ (e) => submitComment(e) }
+                    >
+                        Comment
+                    </button>
+                </Localized>
+                <Localized id='comments-AddComment--reject-button'>
```
mathjazz: I would ignore Reject/Approve & Comment buttons for now, since they'll require quite some effort. We'd need logic to decide when and to whom to show or not show them, and this PR is already getting big.

I'd even be fine with shipping comments without this feature in the first version.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -0,0 +1,90 @@
...
+            </div>
+            <div className='options'>
+                <Localized id='comments-AddComment--comment-button'>
+                    <button
+                        className='comment-btn'
+                        onClick={ (e) => submitComment(e) }
+                    >
+                        Comment
+                    </button>
+                </Localized>
```
mathjazz: Since this would be the only remaining button, I'd remove it (for now) and just let comments be submitted with Enter (LinkedIn and Facebook do the same).

frontend/src/core/comments/components/AddComment.js
abowler2: Yeah, I was waiting until I knew for sure what items were staying and if the naming was okay 😊

abowler2:
Comments can now be added (Yay!)

Known issue:  The button to toggle comments works as expected; however, when a comment is added or deleted the state of the button is reset and the comments close.  I was unsuccessful in fixing this behavior

Questions:

1. The specs indicate that the comments should grow as you type. Is this still the desired behavior? (they are not set up that way ATM)
2. Not sure if the test I wrote is of value for the `AddComment` component and was unsure how/if to test this component since it is pretty much just an input form. 
3. The input is currently only being checked for an empty value. What other checks are needed to verify that the comment is valid, and should those checks be in the component, in the view, or somewhere else?
4. Do the comments need to be set up to accommodate `rtl` languages?

mathjazz:
> * The specs indicate that the comments should grow as you type. Is this still the desired behavior? (they are not set up that way ATM)

Yes. It could be a followup PR, though.

> * Not sure if the test I wrote is of value for the `AddComment` component and was unsure how/if to test this component since it is pretty much just an input form.

You should simulate an event, see https://github.com/mozilla/pontoon/blob/master/frontend/src/core/editor/components/connectedEditor.test.js.

> * The input is currently only being checked for an empty value. What other checks are needed to verify that the comment is valid, and should those checks be in the component, in the view, or somewhere else?

We should sanitize comments using bleach, same as we do for manual notifications, team descriptions, etc.
https://github.com/mozilla/pontoon/blob/master/pontoon/teams/views.py#L111

> * Do the comments need to be set up to accommodate `rtl` languages?

We're likely to see comments in RTL languages. The best we can do I think is to set `dir=auto` on the comment element and input form.

frontend/src/modules/history/components/Translation.js
```diff
@@ -251,172 +240,202 @@ export class TranslationBase extends React.Component<InternalProps, State> {
...
-                                    id='history-Translation--button-approved'
-                                    attrs={{ title: true }}
-                                >
-                                    <button
-                                        className='unapprove fa'
-                                        title='Approved'
-                                        disabled
-                                    />
-                                </Localized>
+                                    { commentCount === 0 ?'Comment' : `${commentCount} Comment` }
```
mathjazz: This text will need to be localized. For inspiration, see https://github.com/mozilla/pontoon/blob/master/frontend/src/modules/batchactions/components/ApproveAll.js#L57. In addition to the `one` and `other` forms, you should also add `0` with a value `Comment`.

adngdb:
The reason why the comment section hides whenever a change happens is, when changes are finished, you [call the `get` action](https://github.com/abowler2/pontoon/blob/translation-comments/frontend/src/modules/history/actions.js#L239) to refresh the data and show the results of the change. But, [that `get` action](https://github.com/abowler2/pontoon/blob/translation-comments/frontend/src/modules/history/actions.js#L66) resets the data of the history section (through [the `request` action](https://github.com/abowler2/pontoon/blob/translation-comments/frontend/src/modules/history/actions.js#L54), which has [effect in the reducer](https://github.com/abowler2/pontoon/blob/translation-comments/frontend/src/modules/history/reducer.js#L63)).

What thus happens is that data disappears, so React re-renders the History component, with new data. It is considered a new component, thus the state is lost. Then it receives updated data from the server, and re-renders again, this time with the correct data, and yet again a new state. 

Possible solutions to this could be: 
- don't reset data when calling the `get` action — but we do that so that, when switching entities, you don't get a mismatch of translations belonging to the wrong entity. Maybe then, call the reset action only when changing entities, and not when adding or removing content (or performing review actions, as that has the same downside).
- don't hide comments at all (but I guess @mathjazz had a good reason for adding that to the list of features? )
- instead of refreshing all data from the server, update data locally with the appropriate changes. For example, in case of `addComment`, make the API call return the new comment's data, and manually add that to the store through a new action/reducer. For `deleteComment`, remove just that single comment from the store data.

mathjazz:
> don't hide comments at all (but I guess @mathjazz had a good reason for adding that to the list of features? )

The purpose of the toggle is to reduce the clutter - we definitely don't want to show the Add comment form by default under each translation.

I'm not sure if it helps, but we could render all *existing* comments by default (along with the add comment form).

frontend/public/static/locale/en-US/translate.ftl
```diff
@@ -270,6 +278,14 @@ history-Translation--button-not-rejected =
 history-Translation--button-rejected =
     .title = Rejected
 
+history-Translation--button-comment =
+    .title = Toggle translation comment
```
mathjazz: The attribute should come after the value.

In order to easily verify that a string is effectively localized, you can turn on [pseudo-localization](https://github.com/mozilla/pontoon/tree/master/frontend#pseudo-localization).

abowler2: I actually meant to write a comment about this section. I did test it with the pseudolocalization, and it wasn't working (even after a rebuild). 

I just made the changes you stated here, rebuilt, and it still is not showing that element as localized, neither the button text or the title.

abowler2: Hold that thought....I just saw that when I stashed some changes I had in order to run the build that change was stashed as well.

abowler2: Nope...It still doesn't appear to be working.

frontend/src/core/comments/components/Comment.js
```diff
@@ -0,0 +1,83 @@
...
+import * as React from 'react';
+import ReactTimeAgo from 'react-time-ago';
+import { Localized } from '@fluent/react';
+
+import './Comment.css';
+
+import type { TranslationComment } from 'core/api';
+import type { UserState } from 'core/user';
+
+import { UserAvatar } from 'core/user'
```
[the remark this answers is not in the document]
mathjazz: BTW, is this documented somewhere?

frontend/src/core/comments/components/AddComment.js
```diff
@@ -0,0 +1,69 @@
...
+            username={ username }
+            title=''
+            imageUrl={ imageURL }
+        />
+        <form className='container' onSubmit={ submitComment }>
+            <Localized
+                id='comments-AddComment--input'
+                attrs={{ placeholder: true }}
+            >
+                <input
```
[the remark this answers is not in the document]
abowler2: The specs indicate that this will have an [autogrow](https://goodies.pixabay.com/jquery/auto-grow-input/demo.html) feature which would apply to an input field. As per @mathjazz's [previous comment](https://github.com/mozilla/pontoon/pull/1524#issuecomment-573549432), this will be added in a new PR.

frontend/src/core/api/entity.js
```diff
@@ -125,7 +125,13 @@ export default class EntityAPI extends APIBase {
         const headers = new Headers();
         headers.append('X-Requested-With', 'XMLHttpRequest');
 
-        return await this.fetch('/get-history/', 'GET', payload, headers);
+        const results = await this.fetch('/get-history/', 'GET', payload, headers);
+
+        if (results.length === 0) {
+            return [];
+        }
```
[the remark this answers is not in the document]
abowler2: No, it is no longer needed and has been removed :)

frontend/src/core/comments/components/AddComment.js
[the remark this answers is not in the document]
mathjazz: Note that the spec links to a jQuery lib, which is no longer useful for us now that we switched to React. I'll update the spect with a link to https://www.npmjs.com/package/react-textarea-autosize.

frontend/src/core/comments/components/Comment.css
```diff
@@ -0,0 +1,83 @@
+.comments-list .comment {
+    padding: 8px 10px;
+    transition: all 0.1s ease-out;
+    display: flex;
+    margin-left: 60px;
+}
+
+.comments-list .comment .user-avatar {
```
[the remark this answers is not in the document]
abowler2: @mathjazz You mentioned that you were going to make CSS changes via a patch. Should I first split the CSS that I have into individual files then?

frontend/src/core/comments/components/AddComment.test.js
```diff
@@ -0,0 +1,16 @@
...
+import AddComment from './AddComment';
+
+describe('<AddComment>', () => {
+    it('calls submitComment function', () => {
+        const submitCommentFn = sinon.spy();
+        const wrapper = shallow(<AddComment
+            submitComment={ submitCommentFn }
+        />);
+
+        wrapper.prop('onSubmit') === submitCommentFn;
```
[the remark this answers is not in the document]
abowler2: So this component has a Form with an `onSubmit` that calls the `submitComment` function. My attempts to test the submit event have been unsuccessful.  The main error I get is: `Method “simulate” is meant to be run on 1 node. 0 found instead.` 

I've tried to find it by 'input' and 'form' as well as by 'id' all with the same result.

frontend/public/static/locale/en-US/translate.ftl
abowler2: This still did not work.  I just realized though that none of the elements I localized in this section are showing as localized with the pseudolocalizer active so maybe its something with my local environment??

frontend/src/core/comments/components/Comment.css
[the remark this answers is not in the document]
mathjazz: Yes, please.

frontend/src/core/comments/components/Comment.js
```diff
@@ -0,0 +1,83 @@
...
+                        <button
+                            className='btn'
+                            onClick={ _delete }
+                        >
+                            Delete
+                        </button>
+                    </Localized>
+                }
+            </div>
+        </div>
```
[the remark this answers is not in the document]
mathjazz: Note that we also agreed to remove the Delete Comment feature for now.

frontend/public/static/locale/en-US/translate.ftl
mathjazz: Did you `make build` before testing? I'm not sure this is documented anywhere, but it's a requirement before testing localizations.

frontend/src/core/comments/components/AddComment.test.js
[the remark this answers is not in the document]
adngdb: That means there was a problem with either the component or the selector you wrote, I guess. Can you find or write back that code and show it to us?

frontend/src/core/comments/components/Comment.js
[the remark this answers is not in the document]
adngdb: No… :cry: Sorry, that's very much my fault. Making a note to document that next week.

frontend/src/core/comments/components/AddComment.test.js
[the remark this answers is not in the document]
mathjazz: You can trigger the error by replacing the line above with:
```js
wrapper.find('form').simulate('submit', { submitCommentFn });
```

I guess you hit the `return null;` in the component, because you don't set the `user` prop, hence the `<form>` is never rendered.

frontend/public/static/locale/en-US/translate.ftl
abowler2: Yes, I ran make build but it still doesn't show as localized.

frontend/src/core/comments/components/Comment.js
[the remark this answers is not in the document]
abowler2: I have the Delete Comment functionality removed now but was waiting to push until Adrian finished his review so as not to confuse things too much.

abowler2:
I've made requested changes except changing the AddComment from `input` to `textarea`.  It looks like that will break some things, so I will work on that today along with splitting the CSS up into separate files.

Also, the test I've written passes but I had to `mount` the component due to the use of `ref` and now it gives the following console warning:
```  console.error node_modules/prop-types/checkPropTypes.js:20
    Warning: Failed context type: The l10n context field must be an instance of ReactLocalization.
        in Localized (at UserAvatar.js:38)
        in a (at UserAvatar.js:31)
        in div (at UserAvatar.js:30)
        in UserAvatar (at AddComment.js:47)
        in div (at AddComment.js:46)
        in AddComments (created by WrapperComponent)
        in WrapperComponent

File: frontend/public/static/locale/en-US/translate.ftl (hunk 56 above)
Comment (abowler2) [reply frontend/public/static/locale/en-US/translate.ftl]: It must have been a caching issue because it is all working fine now (I did clear my cache before, or at least I thought I did 😄 )

**Comment by mathjazz** [comment]
Nice update: I only looked at the code so far and noticed that we need to add migrations ob the `base` and `actionlog` models. For details, run `./manage.py makemigrations` within `make shell`.

File: frontend/src/core/comments/components/AddComment.js [hunk 63]
```diff
@@ -0,0 +1,76 @@
...
+    } = props;
+
+    let commentInput: any = React.useRef();
+
+    if (!user) {
+        return null;
+    }
+
+    const onEnterSubmit = (event: SyntheticKeyboardEvent<HTMLTextAreaElement>) => {
+        if(event.keyCode === 13 && event.shiftKey === false) {
```
mathjazz: Nit: space after if.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -0,0 +1,76 @@
...
+            event.preventDefault();
+            submitComment(event);
+          }
```
mathjazz: Nit: indent.

mathjazz:
Status update:
* We figured out how to fix the Warning in the test (use `shallow`).
* We also figured out how to fix the problem with the logged out state (use `pointer-events: auto`, also for the Show Diff button).

Once that's done, the only thing that remains is fixing the Collapsing issue. As Adrian proposed, you should:
* Remove the [call to `request()`](https://github.com/abowler2/pontoon/blob/translation-comments/frontend/src/modules/history/actions.js#L68) in the `get()` action (fixes the problem immediately).
* Call `request()` where `get()` is called and it makes sense to remove the old data immediately:
  - [When we change entity or plural form](https://github.com/abowler2/pontoon/blob/translation-comments/frontend/src/modules/entitydetails/components/EntityDetails.js#L112).
  - [When batch action takes place](https://github.com/abowler2/pontoon/blob/translation-comments/frontend/src/modules/batchactions/actions.js#L86-L92).

Then I'll push my CSS work. I also got an idea for a slightly different UX, which I'll try to propose as an alternative.

frontend/src/core/comments/components/CommentsList.js
```diff
@@ -0,0 +1,63 @@
...
+
+import * as React from 'react';
+
+import './CommentList.css';
+
+import { Comment, AddComment } from 'core/comments';
+
+import type { TranslationComment } from 'core/api';
+import type { HistoryTranslation } from 'modules/history'
+import type { UserState } from 'core/user';
```
mathjazz: Nit: we sort imports by module, i.e. the value following `from`.

mathjazz:
Excellent job!

Please hide the "Comment" button completely when we don't show the Add Comment component and `commentCount === 0`.

frontend/src/modules/history/components/Translation.js
```diff
@@ -251,172 +238,204 @@ export class TranslationBase extends React.Component<InternalProps, State> {
             className += ' can-reject';
         }
 
-        let canDelete = (canReview || ownTranslation) && !isReadOnlyEditor;
-        let canReject = (canReview || (ownTranslation && !translation.approved)) && !isReadOnlyEditor;
+        let canDelete = (isTranslator || ownTranslation) && !isReadOnlyEditor;
+        let canReject = (isTranslator || (ownTranslation && !translation.approved)) && !isReadOnlyEditor;
+        let canComment = (isTranslator || ownTranslation);
```
mathjazz: According to the "Permissions" section of the [spec](https://docs.google.com/document/d/1W_MEYwtNn-_J0Uiu7gTSwc2B3oux75rjVtjGu-YoFi4/edit#heading=h.1frxkg572534), any logged in contributor `canComment`, so the value should be `user.isAuthenticated`. And then you don't need the `!!` part of the `!!canComment` below.

I think it's still valuable to have the `canComment` variable defined along with other `CanSomething` permissions.

abowler2:
Sorry, pushed and then realized the translation test was broken.  Working on it now

frontend/src/modules/history/components/Translation.test.js
```diff
@@ -286,6 +302,7 @@ describe('<TranslationBase>', () => {
                 translation={ translation }
                 entity={ DEFAULT_ENTITY }
                 locale={ DEFAULT_LOCALE }
+                user={ 'Andy_Dwyer' }
```
mathjazz: While this probably makes the test pass, this is better: `user={ { username: 'Andy_Dwyer' } }`.

mathjazz:
[These](https://gist.github.com/mathjazz/8ed7f0b8047ef9d4f81325c92a021e92) are the CSS fixes I promised, which:
1. Make the Comment toggle and Show diff toggle look & feel the same way.
1. Unify the display of comments and the Add comment form.
1. Fix a few bugs.

mathjazz:
@abowler2 One more patch please - it makes the input fields styled consistently.

```diff
diff --git a/frontend/src/core/comments/components/AddComment.css b/frontend/src/core/comments/components/AddComment.css
index a3fd9f5d1..0f0915bfd 100644
--- a/frontend/src/core/comments/components/AddComment.css
+++ b/frontend/src/core/comments/components/AddComment.css
@@ -1,10 +1,10 @@
 .comments-list .add-comment textarea {
-    background-color: transparent;
-    border: solid 1px #5E6475;
+    background: #333941;
+    border: none;
     border-radius: 4px;
     color: #FFFFFF;
     font-size: 11px;
-    height: 22px;
-    line-height: 22px;
+    height: 24px;
+    line-height: 24px;
     padding: 6px;
 }
diff --git a/frontend/src/modules/machinery/components/Machinery.css b/frontend/src/modules/machinery/components/Machinery.css
index b988b51a0..438089d2e 100644
--- a/frontend/src/modules/machinery/components/Machinery.css
+++ b/frontend/src/modules/machinery/components/Machinery.css
@@ -23,7 +23,7 @@
 }
 
 .machinery > .search-wrapper input[type=search] {
-    background: #4D5967;
+    background: #333941;
     border: none;
     border-radius: 20px;
     box-sizing: border-box;
diff --git a/frontend/src/modules/search/components/SearchBox.css b/frontend/src/modules/search/components/SearchBox.css
index 0817c68cb..fa4e2b90f 100644
--- a/frontend/src/modules/search/components/SearchBox.css
+++ b/frontend/src/modules/search/components/SearchBox.css
@@ -4,7 +4,7 @@
 }
 
 .search-box input[type="search"] {
-    background: #4D5967;
+    background: #333941;
     border: none;
     border-radius: 20px;
     box-sizing: border-box;
```

frontend/src/core/comments/components/Comment.css
```diff
@@ -0,0 +1,29 @@
...
+}
+
+.comments-list .comment .content p {
+    color: #CCCCCC;
+    margin-left: 4px;
+}
+
+.comments-list .comment .info {
+    color: #AAAAAA;
+    font-size: 11px;
```
mathjazz: Please also add `font-weight: 300;` here.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -0,0 +1,76 @@
...
+        <form className='container'>
+            <Localized
+                id='comments-AddComment--input'
+                attrs={{ placeholder: true }}
+            >
+                <textarea
+                    id='comment-input'
+                    name='comment'
+                    dir='auto'
+                    placeholder={ `Write a comment ${'\u2026'}` }
```
adngdb: No space before the ellipsis, and I believe you can simply use the actual character, no need to use the Unicode representation: `…` (you can just copy and paste it ;) ).

adngdb: Note that, whenever you change localized content, what actually matters is what is in the `en-US` FTL file, *not* what is in the code. That's there only to help developers understand what the content is without having to go look into the `.ftl` files, but that content will never be shown to users. We should aim to have content in code always in sync with content in the `en-US` reference FTL files, even if that's a bit painful to do.

frontend/src/core/comments/components/AddComment.test.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import sinon from 'sinon';
+
+import AddComment from './AddComment';
+
+
+const DEFAULT_USER = {
+    user: 'RSwanson',
+    username: 'Ron_Swanson',
```
adngdb: :heart_eyes:

frontend/src/core/comments/index.js
```diff
@@ -0,0 +1,5 @@
+/* @flow */
+
+export { default as CommentsList } from './components/CommentsList';
+export { default as Comment } from './components/Comment';
+export { default as AddComment } from './components/AddComment';
```
adngdb: Nit: these three items should be in reverse order (sorted alphabetically by their module name).

frontend/src/core/user/components/UserAvatar.js
```diff
@@ -0,0 +1,48 @@
...
+            <Localized id='user-UserAvatar--anon-alt-text' attrs={{ alt: true }} >
+                <img
+                    src='/static/img/anon.jpg'
+                    alt='Anonymous User'
+                    height='44'
+                    width='44'
+                />
+            </Localized>
+        </div>;
+
```
adngdb: Nit: please remove this empty line.

frontend/src/modules/history/components/Translation.js
```diff
@@ -251,172 +239,206 @@ export class TranslationBase extends React.Component<InternalProps, State> {
...
+                                    $commentCount={ commentCount }
                                 >
                                     <button
-                                        className='approve fa'
-                                        title='Not approved'
-                                        disabled
-                                    />
+                                        className={ commentCount === 0 ? 'toggle-comments' :
+                                            'toggle-comments'
+                                        }
```
abowler2: Since the className is now the same regardless of `commentCount` this should just set the className to `toggle-comments` without the ternary, correct?

abowler2: I went ahead and made this change. Just let me know if I'm missing something and the logic needs to be reapplied.

frontend/src/modules/history/components/Translation.js
```diff
@@ -274,19 +274,33 @@ export class TranslationBase extends React.Component<InternalProps, State> {
 
                             { (!canComment && commentCount === 0) ? null :
                                 <Localized
-                                    id='history-Translation--button-comment'
+                                    id={ commentCount === 0 ?
+                                        'history-Translation--button-comment' :
+                                        'history-Translation--button-comments'
+                                    }
```
mathjazz: It would be more readable if each `<button>` would be wrapped in their own `<Localized>` element.

frontend/src/modules/history/components/Translation.js
mathjazz: That ternary was weird anyways. :) "If true X else X". :)

frontend/src/modules/history/components/Translation.js
```diff
@@ -274,19 +274,33 @@ export class TranslationBase extends React.Component<InternalProps, State> {
...
+                                    :
+                                        <button
+                                            className='toggle-comments'
+                                            title='Toggle translation comments'
+                                            onClick={ this.toggleComments }
+                                        >
+                                            { commentCount === 1 ?
+                                                `${commentCount} Comment` :
+                                                `${commentCount} Comments`
+                                            }
```
mathjazz: You should simplify this to a simple string, e.g.:
```
{ `${commentCount} Comments` }
```

See:
https://github.com/mozilla/pontoon/pull/1524#discussion_r370056156

frontend/src/modules/history/components/Translation.js
abowler2: Originally there were two different classes based on the condition, which made more sense 😄

frontend/src/modules/history/components/Translation.js
abowler2: I tried that first but couldn't get it working. I'll have another go to see if I can work it out as I agree it would be much easier to understand what is happening there.

frontend/src/modules/history/components/Translation.js
```diff
@@ -142,31 +142,52 @@ export class TranslationBase extends React.Component<InternalProps, State> {
...
+    toggleComments = (event: SyntheticMouseEvent<>) => {
+        event.stopPropagation();
+        this.setState((state) => {
+            return { isCommentVisible: !state.isCommentVisible };
+        });
+    }
 
-        if (!translation.uid) {
-            return <span>{ translation.user }</span>;
+    renderCommentButton(commentCount: number) {
```
mathjazz: `renderCommentToggle` would be consistent with the naming used by the Diff toggle.

frontend/src/modules/history/components/Translation.js
```diff
@@ -142,31 +142,52 @@ export class TranslationBase extends React.Component<InternalProps, State> {
...
+        if(commentCount === 0) {
```
mathjazz: Nit: Missing space.

frontend/src/modules/history/components/Translation.js
```diff
@@ -142,31 +142,52 @@ export class TranslationBase extends React.Component<InternalProps, State> {
...
+            return <Localized
+                id='history-Translation--button-comment'
+                attrs={{ title: true }}
+                $commentCount={ commentCount }
```
mathjazz: We don't use this variable.

frontend/src/modules/history/components/Translation.js
```diff
@@ -142,31 +142,52 @@ export class TranslationBase extends React.Component<InternalProps, State> {
...
+            >
+                    <button
+                        className='toggle-comments'
+                        title='Toggle translation comments'
+                        onClick={ this.toggleComments }
+                    >
+                        { 'Comment' }
+                    </button>
+            </Localized>
         }
```
mathjazz: Nit: remove blank line.

frontend/src/modules/history/components/Translation.js
```diff
@@ -142,31 +142,52 @@ export class TranslationBase extends React.Component<InternalProps, State> {
...
+            >
+                <button
+                    className='toggle-comments'
+                    title='Toggle translation comments'
+                    onClick={ this.toggleComments }
+                >
+                    { commentCount === 1 ?
+                        `1 Comment` :
+                        `${commentCount} Comments`
+                    }
```
mathjazz: See https://github.com/mozilla/pontoon/pull/1524#discussion_r370312943.

mathjazz:
Alright, we're done here! Great work, April!

Before we merge, please disable the feature until we land Team Comment by:
- [ ] Returning HTTP error 501 in the [beginning of the `add_comment()` view](https://github.com/mozilla/pontoon/pull/1524/files#diff-d964cbd7ee26bf4860da2c98511ccb58R494).
- [ ] Returning null in the [beginning of the `renderCommentToggle()` function](https://github.com/mozilla/pontoon/pull/1524/files#diff-3bd40de47215a9570c4279c2542ad9c4R156).

PR #1525, opened 2019-12-19
Fix bug 1605064: Make /machinery work again by restoring team selector

mathjazz:
This is a regression from #1474.

There's another bugfix in a followup commit.


mathjazz:
🙏

PR #1529, opened 2019-12-30
Bug 1594793 - Reformat Pontoon Python code with black.

adngdb:
This is step 1, where we reformat all existing code.


adngdb:
> Do you want to check for incorrectly formatted files on TravisCI?

Yes, that will be part of a follow-up PR. :)

PR #1533, opened 2020-01-08
no bug, fix reference files in sync with excludes

Pike:
This picks up a bug fix in compare-locales in projects
that exlcude files for some locales, so that they're still
available in other locales, and the reference files.


Pike:
I've tested this in my local docker setup, and there sync of `www-l10n` worked as expected.

PR #1535, opened 2020-01-13
Update Docker makefile and documentation for Windows

flodolo:
Can someone verify that this doesn't break anything for Linux and macOS? 

I plan to verify this also on WIndows Professional later this week, where it's possible to use the official Docker installer.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ build:
 	cp ./docker/config/webapp.env.template ./docker/config/webapp.env
 	sed -i -e 's/#SITE_URL#/$(subst /,\/,${SITE_URL})/g' ./docker/config/webapp.env
 
-	${DC} build webapp
+	"${DC}" build webapp
 
 	touch .docker-build
 
@@ -59,50 +59,50 @@ setup-py3: override PYTHON_VERSION=$(PYTHON_3_VERSION)
 setup-py3: setup
 
 setup: .docker-build
-	${DC} run webapp /app/docker/set_up_webapp.sh
+	"${DC}" run webapp /app/docker/set_up_webapp.sh
 
 run: .docker-build
-	${DC} run --rm --service-ports webapp
+	"${DC}" run --rm --service-ports webapp
 
 clean:
 	rm .docker-build
 
 test:
-	${DC} run --rm webapp /app/docker/run_tests.sh
+	"${DC}" run --rm webapp /app/docker/run_tests.sh
 
 test-frontend: jest
 jest:
-	${DC} run --rm -w /app/frontend webapp yarn test
+	"${DC}" run --rm -w /app/frontend webapp yarn test
 
 pytest:
-	${DC} run --rm -w /app webapp pytest --cov-append --cov-report=term --cov=. $(opts)
+	"${DC}" run --rm -w /app webapp pytest --cov-append --cov-report=term --cov=. $(opts)
 
 flake8:
-	${DC} run --rm -w /app webapp flake8
+	"${DC}" run --rm -w /app webapp flake8
 
 flow:
-	${DC} run --rm -w /app/frontend -e SHELL=/bin/bash webapp yarn flow:dev
+	"${DC}" run --rm -w /app/frontend -e SHELL=/bin/bash webapp yarn flow:dev
 
 lint-frontend:
-	${DC} run --rm -w /app/frontend webapp ./node_modules/.bin/eslint src/
+	"${DC}" run --rm -w /app/frontend webapp ./node_modules/.bin/eslint src/
 
 shell:
-	${DC} run --rm webapp /bin/bash
+	"${DC}" run --rm webapp /bin/bash
 
 loaddb:
 	# Stop connections to the database so we can drop it.
-	-${DC} stop webapp
+	-"${DC}" stop webapp
 	# Make sure the postgresql container is running.
-	-${DC} start postgresql
-	-${DC} exec postgresql dropdb -U pontoon pontoon
-	${DC} exec postgresql createdb -U pontoon pontoon
+	-"${DC}" start postgresql
+	-"${DC}" exec postgresql dropdb -U pontoon pontoon
+	"${DC}" exec postgresql createdb -U pontoon pontoon
 	# Note: docker-compose doesn't support the `-i` (--interactive) argument
 	# that we need to send the dump file through STDIN. We thus are forced to
 	# use docker here instead.
-	${DOCKER} exec -i `${DC} ps -q postgresql` pg_restore -U pontoon -d pontoon -O < ${DB_DUMP_FILE}
+	${DOCKER} exec -i `"${DC}" ps -q postgresql` pg_restore -U pontoon -d pontoon -O < ${DB_DUMP_FILE}
 
 build-frontend:
-	${DC} run --rm webapp npm run build
+	"${DC}" run --rm webapp npm run build
 
 build-frontend-w:
-	${DC} run --rm webapp npm run build-w
+	"${DC}" run --rm webapp npm run build-w
--- a/Makefile
+++ b/Makefile
@@ -81,13 +81,13 @@ flake8:
 	"${DC}" run --rm -w /app webapp flake8
 
 flow:
-	"${DC}" run --rm -w /app/frontend -e SHELL=/bin/bash webapp yarn flow:dev
+	"${DC}" run --rm -w /app/frontend -e SHELL=//bin/bash webapp yarn flow:dev
 
 lint-frontend:
 	"${DC}" run --rm -w /app/frontend webapp ./node_modules/.bin/eslint src/
 
 shell:
-	"${DC}" run --rm webapp /bin/bash
+	"${DC}" run --rm webapp //bin/bash
 
 loaddb:
 	# Stop connections to the database so we can drop it.
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ setup-py3: override PYTHON_VERSION=$(PYTHON_3_VERSION)
 setup-py3: setup
 
 setup: .docker-build
-	"${DC}" run webapp /app/docker/set_up_webapp.sh
+	"${DC}" run webapp //app/docker/set_up_webapp.sh
 
 run: .docker-build
 	"${DC}" run --rm --service-ports webapp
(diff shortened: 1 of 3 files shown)


flodolo:
I've only tested `make build`, `make shell`, `make run`. It would be worth to test at least `loaddb`.

flodolo:
Leaving also my notes here, feel free to do what you prefer with it :) (I don't plan to update the docs, this has already taken away a lot of time).

Maybe @jotes wants to give it a try

### Installing Docker on Windows Home (deprecated)

It's only possible to install an deprecated tool, Docker Toolbox
https://docs.docker.com/toolbox/toolbox_install_windows/

Once installed, it will create Docker Quickstart Terminal icon that can be use to start a terminal.

While the package already installs `git`, you need to disable the config `core.autocrlf` before cloning the Pontoon repository, otherwise all files will use Windows line-endings (CRLF). To do so, open a Powershell as Admin (right click on the Start Menu), and run:


```
git config --system --unset core.autocrlf
git config --global core.autocrlf false
```

#### Install make

The easiest way is to use a package manager like Chocolatey
https://chocolatey.org/install

Follow the instructions forr Windows Powershell (Admin), then run

`choco install make`

Follow the prompt requests allowing script execution. At the end, verify that make is available with `make --version`, it should return a version (e.g. GNU Make 4.2.1)

#### Repository path caveat

Make sure to clone the repository in a folder where the user has write permissions. The procedure has been tested with the clone inside the user's home (`c:\Users\username`).

flodolo:
For the record, if you have a Mac and enough space left (~60 gb) on the disk, it's pretty straightforward to test Windows (takes around 40 minutes, most of them going through a crapton of privacy dialogs in the setup finalization phase).

1) Download ISO from Microsoft https://www.microsoft.com/en-us/software-download/windows10ISO (select English International, 64 bit)
2) Open BootCamp Assistant on Mac, follow the instructions to create a partition.
3) During the install, when asked for a product key select "I don't have a product ID". In the next screen you can select the version of Windows you want to install (the ISO contains all of them).

The product won't be activated, but it will work for 30 days, with reasonable limitations (e.g. not being able to customize stuff).

Once you're done, you restart with macOS and remove the Bootcamp partition.

flodolo:
I was hoping Win Pro would be easier… Turns out the makefile doesn't work at all in cmd/Powershell, I think the easiest solution is to use CygWin on top.

I'll keep the loaner for a few more days in case there's more testing needed, but I think I've done all I could here :)

---

### Installing Docker on Windows Pro/Enterprise/Education

Install [Docker Desktop for Windows](https://docs.docker.com/docker-for-windows/install/).

#### Install tools (git, make, cygwin)

The easiest way is to use a package manager like [Chocolatey](https://chocolatey.org/install). Follow the installation instructions for Windows Powershell (Admin), then run `choco install make git cygwin` to install all packages.

Follow the prompt requests allowing script execution. At the end, verify that packages are available with `make --version` and `git --version`, it should return a version for each command.

At this point you need to disable the config `core.autocrlf` before cloning the Pontoon repository, otherwise all files will use Windows line-endings (CRLF), and docker images will fail to build. To do so, open a Powershell as Admin (right click on the Start Menu, select *Windows Powershell (Admin)*), and run:

```
git config --system --unset core.autocrlf
git config --global core.autocrlf false
```

You can use `git config -l` to verify that the value for `core.autocrlf` is correctly set.

At this point, you can open the *Cygwin64 Terminal* and proceed with the installation (the content of `C:` will be available in `/cygdrive/c`).

flodolo:
Proof that it works 🦄 

Capture

jotes:
@flodolo AWSOMETASTIC! You basically figured out the thing that blocked me from trying to go further. :bowing_man: 
I'll take a look today on my laptop with Windows Home.

Makefile
```diff
@@ -36,7 +36,7 @@ build:
 	cp ./docker/config/webapp.env.template ./docker/config/webapp.env
 	sed -i -e 's/#SITE_URL#/$(subst /,\/,${SITE_URL})/g' ./docker/config/webapp.env
 
-	${DC} build webapp
+	"${DC}" build webapp
```
jotes: [nitpick/question/note to myself] I would try to quote this variable once during its initialization.

Pike: Anecdote: We used to have `\\\\\\\\` or `\\\\\\\\\\\\\\\\` in the Firefox build system to ensure that expanding a variable N times would still have a path separator on Windows.

I think it's good practice to ensure quoting at the call sites, if possible.

jotes:
@flodolo Pontoon is working on Windows 💃  I have a question though: I couldn't access http://localhost:8000. To fix this I had to forward ports between the host and the guest machine (like here: https://www.jhipster.tech/tips/020_tip_using_docker_containers_as_localhost_on_mac_and_windows).html). Is this expected?

flodolo:
> @flodolo Pontoon is working on Windows 💃 I have a question though: I couldn't access http://localhost:8000. To fix this I had to forward ports between the host and the guest machine (like here: https://www.jhipster.tech/tips/020_tip_using_docker_containers_as_localhost_on_mac_and_windows).html). Is this expected?

Good point, that's something I didn't add to my notes.

On Windows Pro it was accessible via localhost. 

On Windows Home, when the terminal starts, it shows you the IP of the underlying virtualbox machine, close to the ASCII whale (can you confirm the text or add a screenshot here? I had to replace Win Home with Pro). 

I built the VM with `make build SITE_URL="https://THAT_IP:8000"`, and it was accessible in the host via `https://THAT_IP:8000`

jotes:
@flodolo 
That's true, the IP is displayed below the ASCII whale (but I can't do a screenshot right now).
It can be retrieved after that if you run:
```
docker-machine ip default
```
source: https://devilbox.readthedocs.io/en/latest/howto/docker-toolbox/find-docker-toolbox-ip-address.html

docs/dev/setup.rst
```diff
@@ -108,25 +108,30 @@ run::
...
-You can use ``git config -l`` to verify that the value for ``core.autocrlf`` is correctly set.
+You can use ``git config -l`` to verify that the value for ``core.autocrlf`` is
+correctly set.
 
-At this point, you can open the *Cygwin64 Terminal* and proceed with the installation (the content of
-``C:`` will be available in ``/cygdrive/c``).
+At this point, you can open the *Cygwin64 Terminal* and proceed with the
+installation (the content of ``C:`` will be available in ``/cygdrive/c``). Once
+the Docker image is running, Pontoon's instance will be available at
+`http://localhost:8000`.
```
flodolo: The last sentence is the only part added for Win Pro

docs/dev/setup.rst
```diff
@@ -135,16 +140,29 @@ select *Windows Powershell (Admin)*), and run::
...
+
+Repository Path and SITE_URL
+++++++++++++++++++++++++++++
+
+Make sure to clone the repository in a path where the user has write
+permissions. The procedure has been tested with the clone inside the user's
+home (``c:\Users\username``). Otherwise, the image might fail to load the
+correct volumes.
+
+Since Pontoon will be running inside a VirtualBox machine, in order to access
```
flodolo: From here on, the text is new.

flodolo:
Side question: when I tried, I used https for the SITE_URL, and it didn't create any problem. Should it say http or https, given the rest of references to `localhost` in the doc?

jotes:
I don't think that Pontoon is configured to handle HTTPS traffic on the development environment, I would leave HTTP as the default protocol (for now).
Side question #2: Have you noticed any performance improvements when you switched from Windows Home to Windows Pro? Does the support of Hyper-V make any difference?

flodolo:
>  Side question #2: Have you noticed any performance improvements when you switched from Windows Home to Windows Pro? Does the support of Hyper-V make any difference?

OK, dropping the https:// in the examples then.

I didn't test it enough. The `make build` felt a bit faster though.

jotes:
@flodolo I'll take it for a final round of review today and check if the rest of commands work correctly.

jotes:
@flodolo 

Some tasks are still broken on Windows Home e.g.:
```
$ make pytest                                                                                                           "/c/Program Files/DockerToolbox/docker-compose" run --rm -w /app webapp pytest --cov-append --cov-report=term --cov=.   Starting pontoon_postgresql_1 ... done                                                                                  ERROR: Cannot create container for service webapp: b"the working directory 'C:/Program Files/Git/app' is invalid, it needs to be an absolute path"                                                                                              make: *** [Makefile:78: pytest] Error 1                                                                                                                         
```
Adding slashes to the workdir argument fixes the issue:
```
"/c/Program Files/DockerToolbox/docker-compose" run --rm -w "//app" webapp pytest --cov-append --cov-report=term --cov=.
Starting pontoon_postgresql_1 ... done                                                                                  ================================================= test session starts ==================================================
platform linux2 -- Python 2.7.17, pytest-3.2.3, py-1.4.34, pluggy-0.4.0
Django settings: pontoon.settings (from ini file)
rootdir: /app, inifile: pytest.ini
plugins: django-3.1.2, cov-2.5.1, celery-4.3.0
collected 734 items

pontoon/actionlog/tests/test_models.py ......
pontoon/administration/tests/test_views.py ..........
pontoon/api/tests/test_schema.py ......
```

flodolo:
> Adding slashes to the workdir argument fixes the issue:

So, we need to change all `/app/something` to `//app/something`? I though so (and I did the first iteration), but then thought it was overdoing, since the rest still worked (e.g. `make setup`).

I'm going to push the change as a separate commit, let me know if that fixes it for you.

flodolo:
> I'm going to push the change as a separate commit, let me know if that fixes it for you.

Reinstalled Windows Home, so I tried once again the procedure. I tested:
- pytest
- test
- flake8
- jest
- flow (not sure it's expected, I hat to CTRL+C after `No errors!` showed up)
- lint-frontend (11 errors, 3 warnings?)

And they all work

mathjazz:
> * flow (not sure it's expected, I hat to CTRL+C after `No errors!` showed up)
> * lint-frontend (11 errors, 3 warnings?)

Confirming that it's the same for me on latest master, running Docker on macOS.

flodolo:
I think this should be good to merge at this point, and we can fix other things later as needed.

adngdb:
Ctrl + C to stop flow is expected, the 11 errors are not (and they are not actually valid errors). But I can see them as well locally. I wonder what the hell is going on. Feel free to ignore them, they're not related.

mathjazz:
Great work, @flodolo!

PR #1536, opened 2020-01-14
Bug 1473761 - Expose the ActionLog data in the public API.

adngdb:
This also exposes dependencies of the ActionLog: the User and Translation models. This means a lot more data is available through the API with this change.

### DIFF
--- a/pontoon/api/schema.py
+++ b/pontoon/api/schema.py
@@ -1,17 +1,20 @@
 from __future__ import absolute_import
 
+from django.contrib.auth.models import User as UserModel
+
 import graphene
 from graphene_django import DjangoObjectType
 from graphene_django.debug import DjangoDebug
 
 from pontoon.api.util import get_fields
 
+from pontoon.actionlog.models import ActionLog as ActionLogModel
 from pontoon.base.models import (
     Locale as LocaleModel,
     Project as ProjectModel,
     ProjectLocale as ProjectLocaleModel,
+    Translation as TranslationModel,
 )
-
 from pontoon.tags.models import Tag as TagModel
 
 
@@ -112,6 +115,54 @@ class Locale(DjangoObjectType, Stats):
         return qs.filter(project__disabled=False)
 
 
+class ActionLog(DjangoObjectType):
+    class Meta:
+        model = ActionLogModel
+        only_fields = (
+            "action_type",
+            "created_at",
+            "performed_by",
+            "translation",
+        )
+
+
+class User(DjangoObjectType):
+    class Meta:
+        convert_choices_to_enum = False
+        model = UserModel
+        only_fields = (
+            "username",
+            "first_name",
+            "last_name",
+        )
+
+    actions = graphene.List(ActionLog)
+
+    def resolve_actions(obj, info):
+        return ActionLogModel.objects.filter(performed_by=obj)
+
+
+class Translation(DjangoObjectType):
+    class Meta:
+        model = TranslationModel
+        only_fields = (
+            "user",
+            "locale",
+            "string",
+            "plural_form",
+            "date",
+            "active",
+            "fuzzy",
+            "approved",
+            "rejected",
+        )
+
+    actions = graphene.List(ActionLog)
+
+    def resolve_actions(obj, info):
+        return ActionLogModel.objects.filter(translation=obj)
+
+
 class Query(graphene.ObjectType):
     debug = graphene.Field(DjangoDebug, name="__debug")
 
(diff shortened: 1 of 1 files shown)


pontoon/api/schema.py
```diff
@@ -112,6 +115,54 @@ def resolve_localizations(obj, _info, include_disabled):
...
+        only_fields = (
+            "user",
+            "locale",
+            "string",
+            "plural_form",
+            "date",
+            "active",
+            "fuzzy",
+            "approved",
+            "rejected",
```
mathjazz: Until https://bugzilla.mozilla.org/show_bug.cgi?id=1179954 is fixed, it'd be nice to have a `status` property on the `Translation` model. We did something similar for `missing_strings` and `complete` and make use of that in the schema.

pontoon/api/schema.py
```diff
@@ -112,6 +115,54 @@ def resolve_localizations(obj, _info, include_disabled):
...
+
+
+class User(DjangoObjectType):
+    class Meta:
+        convert_choices_to_enum = False
+        model = UserModel
+        only_fields = (
+            "username",
+            "first_name",
+            "last_name",
```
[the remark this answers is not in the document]
adngdb: It's just so that we don't expose user ids in the API. I'll remove `last_name`, and then we have `first_name` for users who have filled it, and `username` as a backup. Note that I intend this API to be mostly used to generate dashboards, which is why it matters to be able to expose users.

+1 on email addresses, though. That will make this a lot more useful.

pontoon/api/schema.py
```diff
@@ -112,6 +115,54 @@ def resolve_localizations(obj, _info, include_disabled):
...
+    def resolve_actions(obj, info):
+        return ActionLogModel.objects.filter(performed_by=obj)
+
+
+class Translation(DjangoObjectType):
+    class Meta:
+        model = TranslationModel
+        only_fields = (
+            "user",
+            "locale",
```
stasm: Do we need to expose `locale` here? It seems like the only way to query for translations is with `project` and `locale` parameters anyways.

pontoon/api/schema.py
```diff
@@ -176,5 +235,23 @@ def resolve_locale(obj, info, code):
...
+        qs = TranslationModel.objects
+        fields = get_fields(info)
+
+        if "user" in fields:
+            qs = qs.prefetch_related("translation__user")
+
+        if "locale" in fields:
+            qs = qs.prefetch_related("translation__locale")
+
+        return qs.filter(locale__code=locale, entity__resource__project__slug=project,)
```
stasm: I think this should forbid the following cyclic queries:

    translations.user.actions.performed_by
    translations.locale.localizations.project.localizations
        (or remove Translation.locale)

Again, I assume this one can be called valid?

    translations.user.actions.translation

"Show me the translations which the author of each translation on the list somehow modified."

pontoon/api/schema.py
```diff
@@ -112,6 +115,54 @@ def resolve_localizations(obj, _info, include_disabled):
...
+        only_fields = (
+            "username",
+            "first_name",
+            "last_name",
+        )
+
+    actions = graphene.List(ActionLog)
+
+    def resolve_actions(obj, info):
+        return ActionLogModel.objects.filter(performed_by=obj)
```
stasm: Should there be a reasonable max limit to this?

pontoon/api/schema.py
```diff
@@ -176,5 +235,23 @@ def resolve_locale(obj, info, code):
...
+        if "user" in fields:
+            qs = qs.prefetch_related("translation__user")
+
+        if "locale" in fields:
+            qs = qs.prefetch_related("translation__locale")
+
+        return qs.filter(locale__code=locale, entity__resource__project__slug=project,)
+
+    def resolve_actions(obj, info):
+        return ActionLogModel.objects.all()
```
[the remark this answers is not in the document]
stasm: Also, you probably want to forbid the following queries:

    actions.performed_by.actions
    actions.translation.locale.localizations.project.localizations
        (or remove Translation.locale)

pontoon/api/schema.py
stasm: Ah, no, that's wrong. It's possible to do `actions.translation` from the root, from `translations`, and from `user`. Is it needed?

pontoon/api/schema.py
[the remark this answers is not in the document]
adngdb: We already have 47119 registered actions… So yeah, we need to set a limit! :grin:

adngdb:
I'm going to put this on hold until we have figured out data story out. Thanks for the comments everyone!

PR #1537, opened 2020-01-14
Fix bug 1527853: Remove mark_placeables()

mathjazz:
Translate.Next uses its own placeables library, so we can remove the `mark_placeables()` method.


adngdb:
You forgot to remove a few things. :)

mathjazz:
Ah, thanks. Updated.

PR #1541, opened 2020-01-23
Validate priority before comparison

mathjazz:
Python 3.0 has simplified the rules for ordering comparisons: The ordering
comparison operators (<, <=, >=, >) raise a TypeError exception when the
operands don’t have a meaningful natural ordering.

Thus, expressions like 0 > None are no longer valid, and raise a TypeError
instead of returning False.


mathjazz:
Thanks for the fast review! Confirmed that it's working on stage.

PR #1545, opened 2020-01-28
Bug 1594793 - Enforce black formatting in CI.

adngdb:
This requires Python 2 to be dropped, so blocked on #1544.


adngdb:
This is failing because of black! Yay! :tada:

PR #1546, opened 2020-01-28
Fix  Bug 1611512  - Add ability for Add Comment input to grow as you type

abowler2:
This fix allows for the textarea in Add Comment to grow as the user types.

________________
I have actually set up a way to grow the textarea either in a downward direction or an upward direction (similar to Slack). However, both options pose additional challenges that I have not yet found solutions for.

**Downward growth** (which is what is active): If there are multiple comments and the Add Comment textarea grows to the point that the outer `history` div begins scrolling then every keypress causes the `history` scroll to jump back up which is not a good experience.

**Upward growth** (can be activated by uncommenting the CSS in `AddComent.css`): This has a fluid growth action, but it does not effect the `history` div which means the overflow is never switched to scroll and the textarea grows up and covers the comment above it.

I have tried implementing lifecycle methods/hooks using a `ref` on the History `ul` and/or on the existing `ref` on the textarea to set or remember the scroll position but have not had success. The closest I came was with the downward growth and calling a function with `onScroll` that scrolled to the bottom, but this option disabled the ability to scroll back up.

Let me know which option is preferable (upward or downward growth) and your thoughts on how I can address the associated scrolling issue for that option. ~ Thanks

### DIFF
--- a/frontend/src/core/comments/components/AddComment.js
+++ b/frontend/src/core/comments/components/AddComment.js
@@ -26,7 +26,7 @@ export default function AddComments(props: Props) {
     } = props;
 
     const commentInput: any = React.useRef();
-    const [rows, setRows] = React.useState(1);
+    const rows = 1;
     const minRows = 1;
     const maxRows = 3;
 
@@ -49,12 +49,12 @@ export default function AddComments(props: Props) {
             if (currentRows === previousRows) {
                 commentInput.current.rows = currentRows;
             }
-
-            if (currentRows >= maxRows) {
+            else if (currentRows < maxRows) {
+                commentInput.current.rows = currentRows;
+            }
+            else if (currentRows >= maxRows) {
                 commentInput.current.rows = maxRows;
             }
-
-            currentRows < maxRows ? setRows(currentRows) : setRows(maxRows);
         }
     }
 
--- a/frontend/src/core/comments/components/AddComment.js
+++ b/frontend/src/core/comments/components/AddComment.js
@@ -26,9 +26,9 @@ export default function AddComments(props: Props) {
     } = props;
 
     const commentInput: any = React.useRef();
-    const rows = 1;
     const minRows = 1;
     const maxRows = 3;
+    const rows = minRows;
 
     if (!user) {
         return null;
--- a/frontend/src/core/comments/components/AddComment.js
+++ b/frontend/src/core/comments/components/AddComment.js
@@ -41,15 +41,11 @@ export default function AddComments(props: Props) {
         }
         else {
             const textAreaLineHeight = 24;
-            const previousRows = commentInput.current.rows;
             commentInput.current.rows = minRows;
 
             const currentRows = Math.trunc(commentInput.current.scrollHeight / textAreaLineHeight);
 
-            if (currentRows === previousRows) {
-                commentInput.current.rows = currentRows;
-            }
-            else if (currentRows < maxRows) {
+            if (currentRows < maxRows) {
                 commentInput.current.rows = currentRows;
             }
             else if (currentRows >= maxRows) {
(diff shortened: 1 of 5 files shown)


abowler2:
Additional issue discovered is that when the user uses the `shift + enter` option to add a new line without submitting it is not honored when the comment renders.  It shows in the database with the new lines intact but then when rendered it all runs together.

mathjazz:
Hi @abowler2, could you please resolve the conflict?

abowler2:
Third time's the charm....right 😉

frontend/src/core/comments/components/AddComment.js
```diff
@@ -67,8 +84,9 @@ export default function AddComments(props: Props) {
                     name='comment'
                     dir='auto'
                     placeholder={ `Write a comment…` }
+                    rows={ rows }
                     ref={ commentInput }
-                    onKeyDown={ onEnterSubmit }
+                    onKeyUp={ handleKeyUp }
```
mathjazz: If you paste text instead of typing it, textarea doesn't resize until you release the last key.

I think using `onChange` instead of `onKeyUp` would fix that.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -67,8 +84,9 @@ export default function AddComments(props: Props) {
                     name='comment'
                     dir='auto'
                     placeholder={ `Write a comment…` }
+                    rows={ rows }
```
mathjazz: Should this actually be `minRows`? Then you can remove the `rows` variable.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -25,17 +25,33 @@ export default function AddComments(props: Props) {
...
+        else {
+            const textAreaLineHeight = 24;
+            commentInput.current.rows = minRows;
+
+            const currentRows = Math.trunc(commentInput.current.scrollHeight / textAreaLineHeight);
+
+            if (currentRows < maxRows) {
+                commentInput.current.rows = currentRows;
+            }
+            else if (currentRows >= maxRows) {
```
mathjazz: That's the same as `else`, right?

abowler2:
With the current changes the issue I was having with the downward growth is no longer present. However, it is allowing for an empty input to be submitted into the database even though there is a check for empty within the `submitComment` function.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -67,8 +79,10 @@ export default function AddComments(props: Props) {
                     name='comment'
                     dir='auto'
                     placeholder={ `Write a comment…` }
+                    rows={ minRows }
                     ref={ commentInput }
-                    onKeyDown={ onEnterSubmit }
+                    onChange={ handleOnChange }
+                    onKeyUp={ submitComment }
```
mathjazz: I believe you should use `onKeyDown` here, same as we did before. Otherwise the newline (`%0A`) character enters the `<textarea>` before we call `submitComment()`, which means `commentInput.current.value` isn't empty.

abowler2: Ahhh, yes that makes sense.  I made the change and it indeed fixed the issue.  Thanks!

abowler2:
> Additional issue discovered is that when the user uses the `shift + enter` option to add a new line without submitting it is not honored when the comment renders. It shows in the database with the new lines intact but then when rendered it all runs together.

Did you have thoughts on this??

mathjazz:
> > Additional issue discovered is that when the user uses the `shift + enter` option to add a new line without submitting it is not honored when the comment renders. It shows in the database with the new lines intact but then when rendered it all runs together.
> 
> Did you have thoughts on this??

We should use [`white-space: pre-wrap;`](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space) when rendering comments.

OTOH, we should use `white-space: nowrap;` in the comment author element to prevent wrapping it into multiple lines.

Please also revert the changes in Translation.js and views.py when the PR is complete, so that the comments functionality remains disabled.

PR #1548, opened 2020-01-29
Fix bug 1603380: Search broken when filtering for more than one string status

mathjazz:


pontoon/base/models.py
```diff
@@ -1997,17 +1999,20 @@ def get_filtered_entities(self, locale, query, rule, match_all=True, prefetch=No
                 ):
                     plural_pks.append(candidate.pk)
 
+        translations = Translation.objects.filter(locale=locale)
+
+        # Additional filter on the project field speeds things up because it makes faster
+        # to execute a SQL subquery generated by Django.
+        if project and project.slug != "all-projects":
+            translations = translations.filter(entity__resource__project=project,)
```
mathjazz: Nit: trailing comma.

mathjazz:
Great work! r+

Regarding the followup performance improvement it's worth noting that:
1. Filtering for Translated and Missing separately takes less time than if filters are combined.
2. Filtering for Translated and Missing across all projects is faster than in e.g. Firefox only.

PR #1550, opened 2020-01-30
Fix bug 1606302 - Show Approve button for Unreviewed FTL strings.

adngdb:


frontend/src/core/editor/selectors.js
```diff
@@ -60,9 +62,25 @@ export function _existingTranslation(
...
+                // We thus store a simplified version of the Fluent message in memory,
+                // but the history has actual Fluent-formatted strings (with ID).
+                // Thus, we need to reconstruct the Fluent message in order to be able
+                // to compare it with items in history.
+                const reconstructed = fluent.serializer.serializeEntry(
+                    fluent.getReconstructedMessage(entity.original, translation)
+                );
+                existingTranslation = history.translations.find(
+                    t => t.string === reconstructed
+                )
```
mathjazz: You can move this block out of the `if-else` block and if you call `reconstructed` and `translation` by the same name, you don't need to repeat it below.

adngdb: As we talked about yesterday, this is very low value, and I actually like that we have different variable names in there, to make the difference more explicit. Hence, I'm not going to do it. :angel:

PR #1552, opened 2020-01-30
No bug - On some of Docker images, the `pontoon` user won't have id=1000 b…

jotes:
…y default.

Changes:
* Explicitly set the user id (to avoid confusions when maintainers of the docker image will change something).
* Parametrize the user/group id in the Makefile to give people an easy way to get around that problem.

To give you more context -> I rebuild the webapp image today and had problems with permissions (e.g. making new migrations).
My user id was:
```bash
pontoon@63acee8f6019:/app$ id
uid=999(pontoon) gid=1000(pontoon) groups=1000(pontoon)
```
@adngdb @Pike can you take a look?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ help:
 	@echo "  setup            Configures a local instance after a fresh build"
 	@echo "  run              Runs the whole stack, served on http://localhost:8000/"
 	@echo "  clean            Forces a rebuild of docker containers"
-	@echo "  shell            Opens a Bash shell in a webpp docker container"
+	@echo "  shell            Opens a Bash shell in a webapp docker container"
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -15,7 +18,7 @@ help:
 	@echo "  setup            Configures a local instance after a fresh build"
 	@echo "  run              Runs the whole stack, served on http://localhost:8000/"
 	@echo "  clean            Forces a rebuild of docker containers"
-	@echo "  shell            Opens a Bash shell in a webapp docker container"
+	@echo "  shell            Opens a Bash shell in a webpp docker container"
```
jotes: Mispell, I'll fix it.

PR #1553, opened 2020-01-31
Fix Bug 1584180: run quality checks and save pretranslations as fuzzy.

VishalCR7:


mathjazz:
All the problems seem to be fixed! Well done!

I still need to have a closer look at the code.

pontoon/pretranslation/pretranslate.py
```diff
@@ -79,26 +79,59 @@ def update_changed_instances(tr_filter, tr_dict, locale_dict, translations):
...
-        projectlocale.latest_translation = translations[index]
+        locale.latest_translation = translations[data["latest_translation_index"]]
+        projectlocale.latest_translation = translations[
+            data["latest_translation_index"]
+        ]
 
-        # Since the translations fall into unreviewed category.
-        locale.unreviewed_strings += diff
-        projectlocale.unreviewed_strings += diff
+        # Since the translations fall into fuzzy category.
```
mathjazz: Nit: Not sure if that comment is necessary here?

pontoon/pretranslation/tasks.py
```diff
@@ -102,44 +114,112 @@ def pretranslate(project, locales=None, entities=None):
...
+    for t in translations:
+        key = (t.entity.pk, t.locale.pk)
+        # Remove duplicate changes to prevent unique constraint violation
+        if key not in existing:
+            changed_entities[key] = ChangedEntityLocale(
+                entity=t.entity, locale=t.locale
+            )
+    ChangedEntityLocale.objects.bulk_create(changed_entities.values())
+
+    # Update latest activity and fuzzy count for the project.
```
mathjazz: Nit: fuzzy count -> stats

VishalCR7:
> I wonder if you could use the `bulk_run_checks()` method from `pontoon.checks.utils` to run checks more elegantly?

We were using `bulk_run_checks()` earlier [here](https://github.com/mozilla/pontoon/pull/1553/commits/0d545732125694b78543d7f880de674a37d422fa#diff-411c4ccc71efd731fab9583b132ed846R140) But that makes it difficult to keep count of strings with errors/warnings. We **might be able** to do that using bulk_run_checks as well but it would definitely cost us some DB calls to find number of translations in each category (fuzzy, error, warning) for each locale (& projectLocale) and translated resource.

mathjazz:
Closing on behalf of #1573.

PR #1554, opened 2020-01-31
Fix Bug 1605206: Make pretranslation a background task.

VishalCR7:


mathjazz:
Early observation:

Screenshot 2020-01-31 at 15 51 29

We should remove the "..." now that this is a background task, because it doesn't look very nice. :-)

pontoon/administration/static/js/admin_project.js
```diff
@@ -83,7 +83,7 @@ $(function() {
       return;
     }
 
-    button.addClass('in-progress').html('Pretranslating...');
+    button.addClass('in-progress').html('Pretranslating');
```
mathjazz: Now I see that we're doing the exact same thing with the Sync button. :/

So I suggest you revert this change (i.e. restore ...) and increase the width of buttons from 130px to 150px.

VishalCR7: I should have pointed it out earlier. What about the actual async behavior? Is it working as expected?

pontoon/pretranslation/tasks.py
```diff
@@ -2,18 +2,22 @@
...
+from pontoon.base.tasks import PontoonTask
+from pontoon.sync.core import serial_task
 
 
 log = logging.getLogger(__name__)
 
 
-def pretranslate(project, locales=None, entities=None):
+@serial_task(settings.SYNC_TASK_TIMEOUT, base=PontoonTask, lock_key="project={0}")
+def pretranslate(self, project, locales=None, entities=None):
```
mathjazz: I haven't tested this yet, but I wonder if this works as expected when called from `pontoon/sync/tasks.py`.

pontoon/sync/core.py
```diff
@@ -53,7 +53,13 @@ def wrapper(func):
         @shared_task(bind=True, **celery_args)
         @wraps(func)
         def wrapped_func(self, *args, **kwargs):
-            lock_name = "serial_task.sync[{}]".format(lock_key.format(*args, **kwargs))
+            if func.__name__ == "pretranslate":
+                lock_name = "task.pretranslate[{}]".format(lock_key.format(args[0].pk))
+            else:
+                lock_name = "serial_task.sync[{}]".format(
+                    lock_key.format(*args, **kwargs)
+                )
```
mathjazz: We should generalize this, and that shouldn't be too hard now that we only have 1 sync task:
https://github.com/mozilla/pontoon/commit/dd5e48051abb2f02f4d3591c1566ca7f9ab9f1ec#diff-2be3ba079451045c9badba06c960b1d0

pontoon/pretranslation/tasks.py
VishalCR7: I don't remember exactly, but I might have tested both cases before opening the PR during All hands. I will do it once again.

VishalCR7: @mathjazz : It is working as expected for me in both the cases.

pontoon/pretranslation/tasks.py
```diff
@@ -2,18 +2,22 @@
...
 )
+from pontoon.base.tasks import PontoonTask
+from pontoon.sync.core import serial_task
 
 
 log = logging.getLogger(__name__)
 
 
-def pretranslate(project, locales=None, entities=None):
+@serial_task(settings.SYNC_TASK_TIMEOUT, base=PontoonTask, lock_key="project={0}")
```
mathjazz: I'm getting the following warning when I start the pretranslate process while another pretranslate process for the same project is already running:

```
/usr/local/lib/python3.7/site-packages/django/core/cache/backends/base.py:245:
CacheKeyWarning: Cache key contains characters that will cause errors if used with memcached:
':1:serial_task.pontoon.pretranslation.tasks.pretranslate[project=Pontoon Test]'
  'used with memcached: %r' % key, CacheKeyWarning
```

That's because `{0}` gets replaced with the representation of the Project instance (i.e. project.name), which in my case includes space.

In the case of sync task, we pass the project PK.

We should either pass the project PK here as well or remove white spaces from the lock in the `serial_task()` function.

VishalCR7: Moving forward with passing project pk.

pontoon/administration/static/css/admin_project.css
```diff
@@ -136,7 +136,7 @@ form .button {
   border-radius: 3px;
   font-weight: 400;
   text-transform: uppercase;
-  width: 130px;
+  width: 150px;
```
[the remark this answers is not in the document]
VishalCR7: Went with 1st option.

PR #1555, opened 2020-02-03
Fix bug 1612814 - Harcode UID and GID to fix builds on macOS

jotes:
Uids and Gids on macOS start from very low numbers (e.g. 20) and they overlap with the ids of the existing services.
This commit hardcodes the UID and GID but keeps them parametrizable via commandline.

@Pike @adngdb r?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,8 +6,8 @@ DOCKER := $(shell which docker)
 # https://docs.djangoproject.com/en/dev/ref/django-admin/#runserver
 SITE_URL ?= http://localhost:8000
 
-USER_ID?=$(shell id -u)
-GROUP_ID?=$(shell id -g)
+USER_ID?=1000
+GROUP_ID?=1000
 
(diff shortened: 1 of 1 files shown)


jotes:
Okay, I have mixed thoughts about this patch. Do you have any reports if the previous setup (after Gosu was removed) works on MacOS systems (@mathjazz probably could tell more about it)?
If I understand everything correctly, it's possible to build Pontoon image on macOS but permissions of files on the volume aren't mapped between a container and the host machine (?).

jotes:
Also, I'll update the dev docs as the last step (after we agree on the solution to this problem).

hbinkle:
I can confirm that the state before this fix causes the build also fail on windows wsl (ubuntu).
With this fix (_USER_ID?=1000,GROUP_ID?=1000_) it works like a charm again.

mathjazz:
Leaving it to @adngdb or @Pike to decide if this is ready to merge.

PR #1557, opened 2020-02-04
Show default avatar for users without Gravatar URLs

mathjazz:
@abowler2 Could you please review this?

We broke user avatars for users that don't have Gravatar URLs, e.g. if Translation is Imported:
https://pontoon.mozilla.org/it/firefox/all-resources/?string=74127


mathjazz:
Thanks!

PR #1558, opened 2020-02-05
deploy pontoon on k8s (kubernetes)

hbinkle:
…//github.com/mozilla/pontoon/pull/1555

* base.py: added ability to include ALLOWED_HOSTS from environment variable
* run_webapp.sh:
   - added log output
   - added dumping SSH_KEY and KNOWN_HOSTS to pontoon users .ssh files
   - added call syncprojects.sh
* syncprojects.sh: script running sync_projects in a loop using environment variable SYNC_INTERVAL [minutes]
* k8s-pontoon-example.yaml: example file for deploying pontoon with nginx on a kubernetes cluster


hbinkle:
I don't know what the "black" check mean with "reformatted".
I only changed few lines since the original pull:
image

adngdb:
Hi @hbinkle, we use black for Python code formatting now, and you did change a Python file without running black afterwards. Please see our documentation about it here: https://mozilla-pontoon.readthedocs.io/en/latest/dev/contributing.html#python-code-conventions (Note that in this case, the code is very short so you might even be able to fix the errors manually.)

hbinkle:
@adngdb Thanks.
As you can see at my last two commits for this PR, I already did fix the errors.
But the check still fails with a message I don't understand:
```
$ black pontoon/ --check
would reformat /home/travis/build/mozilla/pontoon/pontoon/settings/base.py
Oh no! 💥 💔 💥
1 file would be reformatted, 446 files would be left unchanged.
The command "black pontoon/ --check" exited with 1.
```

adngdb:
@hbinkle you fixed errors from `flake8`, but not from `black`. They are two different tools. Make me realize that the document is not entirely clear, and we could benefit from having something like `make black`. 

In the mean time, what you can do is run this: 
```
make shell
black pontoon/
```

That should fix your issue without having to integrate black into your code editor.

hbinkle:
@adngdb  I'm sorry, "make shell --> black pontoon/" doesn't work for me and I'm also unable to install black in WSL
I give up at this point.

mathjazz:
Hey @hbinkle, could you please clarify what you mean by "doesn't work"? Is there an error message you get?

You can also rebase against latest master and run the recently added `make black` command.

hbinkle:
> Hey @hbinkle, could you please clarify what you mean by "doesn't work"? Is there an error message you get?

I can't install black in WSL:
```shell
root@P-DSEW002772b:/mnt/c/Projekte/pontoon/latest# pip install black
WARNING: pip is being invoked by an old script wrapper. This will fail in a future version of pip.
Please see https://github.com/pypa/pip/issues/5599 for advice on fixing the underlying issue.
To avoid this problem you can invoke Python with '-m pip' instead of running pip directly.
DEPRECATION: Python 2.7 reached the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 is no longer maintained. A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support
ERROR: Could not find a version that satisfies the requirement black (from versions: none)
ERROR: No matching distribution found for black
root@P-DSEW002772b:/mnt/c/Projekte/pontoon/latest# python --version
Python 3.6.8
root@P-DSEW002772b:/mnt/c/Projekte/pontoon/latest#
```

mathjazz:
"All checks have passed" 🏆 👍

pontoon/settings/base.py
```diff
@@ -520,15 +520,20 @@ def _allowed_hosts():
...
     # to local development (usually the host is localhost:8000).
     if ":" in host:
         host_no_port = host.rsplit(":", 1)[0]
-        return [host, host_no_port]
+        result = [host, host_no_port]
+
+    # add values from environment variable. Needed in case of URL/domain redirections
+    envVarsStr = os.getenv("ALLOWED_HOSTS", "127.0.0.1:8000")
+    envVars = [x.strip() for x in envVarsStr.split(",")]
+    result.extend(envVars)
```
adngdb: Please use `underscore_names` for your variables here.

docker/syncprojects.sh
```diff
@@ -0,0 +1,23 @@
...
+        do
+           echo "calling sync_projects for user pontoon"
+           python /app/manage.py sync_projects > /app/lastSync.log
+           # multiply by 60 to have minutes in the env var
+           sleep $((syncInterval*60))
+        done
+    fi
+fi
+echo "SYNC_INTERVAL is not defined or <5, doing nothing."
+echo "SYNC_INTERVAL is not defined or <5, doing nothing." >> /app/run_webApp.log
```
adngdb: This looks like a very unreliable way to run sync. The first failure will break the whole process, and you'll have to restart the app manually to recover. Aren't there better solution for running jobs regularly out there, that don't require a fragile script like this?

docker/run_webapp.sh
```diff
@@ -1,12 +1,36 @@
...
+echo "out/logs is going into '/app/run_webApp.log'"
 
-# Prepares then runs the webapp.
+#user:
+(echo ">>> running as user: " && whoami) >> /app/run_webApp.log
 
-echo ">>> Setting up the db for Django"
-python manage.py migrate
+#user pontoon stuff --> get enc vars
+(echo ">>> running as user: " && whoami) >> /app/run_webApp.log
```
adngdb: This is repeated twice.

pontoon/settings/base.py
hbinkle: done

docker/run_webapp.sh
hbinkle: removed

docker/syncprojects.sh
hbinkle: right, I tried to use a cron job, but with no success. For me this works for now. This is better than no sync at all. Feel free to improve. 
BTW. in k8s you will not try to restart the app or similar, you simply drop the entire pod.

adngdb:
Hey @hbinkle, sorry for coming back to you so late. Thanks a lot for the work you put in here!

I am concerned about merging this though. I like the `k8s-pontoon-example.yaml` file, as it is branded as an example and might serve others, while making it clear that it shouldn't be used as is. I don't think that's the case for the other files, notably the scripts. They don't meet the security standards that I want to keep in the repo, and thus I am not willing to merge them, and run the risk that other people might use them in production without knowing that they can be harmful. 

My other concerns are that this might actually be of little value without a proper documentation to go with it, and that this impacts our development workflow (which is another reason why I'd wanted to see a separate `run_webapp` file dedicated to production). Oh and I don't understand you latest changes, related to Azure? 

I understand that might not have the time to invest in this work, and that's fine. I'm wondering what to do with this, as I'd like for it not to be wasted. One option would be for someone to take this over and make it so that it's good enough for me to merge. Another might be to just slice it, and merge only the k8s example part? I'm open to other suggestions. I also won't have time to work on this personally for the foreseeable future, sadly.

Once again, thanks a lot for your contribution!

hbinkle:
Hello @adngdb ,
thanks for comming back on this.
The commit regarding to azure was not intended.
For saving the work and not to provide an instable state I suggest to cancel this PR and have a look at:
https://github.com/mozilla/pontoon/pull/1580
This does also contain a readme.md how to apply the changes when doing fruther work towards running pontoon on a k8s environment

adngdb:
Thanks @hbinkle! Closing this in favor of #1580.

PR #1561, opened 2020-02-07
Fix Bug 1611508: Implement team comments

abowler2:
Adding the ability for users to leave Team comments

___
Sending up my initial work for review to ensure I'm on the right track. 
(I have left comments and/or TODOs where further work or changes are needed)

I could also use a little direction on the query for the Team Comments within the `get_team_comments` in `views.py` as my initial attempts are not producing results.

TODO:

- [ ] Update https://discourse.mozilla.org/t/translation-comments-help-us-pick-the-best-ux/54808/12


frontend/src/core/comments/components/TeamComment.css
```diff
@@ -0,0 +1,6 @@
+.team-comment > p {
+    border-bottom: 1px solid #5E6475;
+    color: #aaa;
+    font-style: italic;
+    padding: 15px 10px;
+}
```
mathjazz: Please make sure this is styled exactly the same as in other two tabs. Right now it is not - I see that the `line-height` is not the same as in other-locales.

frontend/src/modules/entitydetails/components/Helpers.js
```diff
@@ -63,6 +64,11 @@ export default class Helpers extends React.Component<Props> {
                     </Localized>
                     <OtherLocalesCount otherlocales={ otherlocales } />
                 </Tab>
+                <Tab>
+                    <Localized id='entitydetails-Helpers--comments'>
+                        { 'Comments' }
+                    </Localized>
```
mathjazz: We'll also need count here.

pontoon/base/views.py
```diff
@@ -436,6 +436,25 @@ def get_translation_history(request):
...
+    try:
+        entity = request.GET["entity"]
+        locale = request.GET["locale"]
+    except MultiValueDictKeyError as e:
+        return JsonResponse(
+            {"status": False, "message": "Bad Request: {error}".format(error=e)},
+            status=400,
+        )
+
+    # TODO: query for team comments
```
mathjazz: You'll need something like this here:
```python
comments = Comment.objects.filter(entity=entity, locale=locale)
```

abowler2:
Comments render in the tab 🎉   Moving on to implement the ability to add team comments.
___
Question: I'm using the type `TranslationComment` for the team comments as it is the correct type but that feels wrong, and making a new type would not be DRY since it would be exactly the same as `TranslationComment`. I know we had to rename this when implementing comments because there is a component named `Comment` which caused conflict.  So, should we rename this again to something more generalized since it is now being used for both translation and team comments?

Regarding the failing tests - I can't work out what it wants reformatted. I have `black` set as my formatter and have run `Format Document` and it made no changes. In addition, I have tried to visually find what might be causing the failure but my formatting appears to be the same as similar bits of code for both the listed files.

adngdb:
I don't think it's too bad to use `TranslationComment` for team comments, though I agree it might be confusing. A simple thing you can do is create an alias. For example, `type TeamComment = TranslationComment;`, and then use `TeamComment`. If one day the two types differ, it will be easy to adapt, and if they don't, then it's easy to evolve both at the same time too.

pontoon/base/views.py
```diff
@@ -436,6 +436,30 @@ def get_translation_history(request):
...
+        )
+
+    entity = get_object_or_404(Entity, pk=entity)
+    locale = get_object_or_404(Locale, code=locale)
+    comments = Comment.objects.filter(entity=entity, locale=locale)
+
+    payload = []
+
+    for c in comments:
+        payload.append(c.serialize())
```
mathjazz: These 4 lines look like a good candidate for [list comprehension](https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions):
```python
payload = [c.serialize() for c in comments]
```

frontend/src/core/comments/components/TeamComment.js
```diff
@@ -0,0 +1,44 @@
...
+
+    if (!teamComments.comments.length) {
+        return <section className="no-team-comments">
+            <Localized id="entitydetails-Helpers--no-comments">
+                <p>No comments available.</p>
+            </Localized>
+        </section>
+    }
+
+    return <div className='comments-list team-comment-list'>
```
mathjazz: Nit: team-comment**s**-list (because `comments-list`).

frontend/src/core/comments/components/Comment.js
```diff
@@ -22,7 +22,7 @@ export default function Comment(props: Props) {
         return null;
     }
 
-    return <li className='comment'>
+    return <li className='comment team-comment'>
```
mathjazz: `Comment` is a reusable component, so there should be no code specific to the component that uses it in it. You can simply omit `team-comment` here and use `.comment` in CSS.

abowler2: Forgot about that one 😄 , I originally put that in as I was having trouble getting my margin changes for the Team Comments to work.

abowler2:
Updated status of this fix with the AddComment component implemented.
___
Originally `addComment` passed the `content` and the `translationId` I changed that up to just `id` as the Team Comments often won't have a translation ID yet and I don't want those comments to cross-over with those in the `History` panel.  ATM, I set the team comments to pass in the `entityId` and kept the `translationId` on those in history. So, my question is:  How/where do I differentiate between the `translation comments` and the `team comments`?

abowler2:
> > Also, IIRC the `action log` requires a translation ID.
> 
> It [doesn't](https://github.com/abowler2/pontoon/blob/team-comments/pontoon/actionlog/models.py#L31-L70) - when we delete the translation for example, we pass `Entity` and `Locale` instead of `Translation` to the `ActionLog`.
> 
> We should add a similar validation to the `Comment` model.

Can you please clarify what validation would be needed in the `Comment` model?  Did you mean that we need to add similar validation for `Comment added` within the `ActionLog` or something else?

mathjazz:
> @mathjazz Can you please clarify what validation would be needed in the `Comment` model? Did you mean that we need to add similar validation for `Comment added` within the `ActionLog` or something else?

Sure. :) We should override the `save()` method of the `Comment` model to validate foreign keys before saving a comment. The validation should check that the comment is:
- either saved with `translation` and without `entity` and `locale`
- or with `entity` and `locale` and without `translation` 

In other cases, `ValidationError` should be raised.

frontend/src/core/comments/actions.js
```diff
@@ -0,0 +1,91 @@
...
+    return async dispatch => {
+        dispatch(request(entity));
+
+        // Abort all previously running requests.
+        await api.entity.abort();
+
+        let content = await api.entity.getTeamComments(entity, locale);
+
+        // The default return value of aborted requests is {},
+        // which is incompatible with reducer
```
adngdb: If the request was aborted, it's probably better to not call the `receive` action, isn't it? Since the content will be bogus anyway?

frontend/src/core/comments/components/CommentsList.js
```diff
@@ -44,7 +44,7 @@ export default function CommentsList(props: Props) {
                 user={ translation.user }
                 username={ translation.username }
                 imageURL={ translation.userGravatarUrlSmall}
-                translationId={ translation.pk }
+                id={ translation.pk }
```
adngdb: I would avoid using `id` as a prop name, as it can be confused with the `id` attribute of DOM nodes. I thought `translationId` was fine here, why change it?

frontend/src/core/comments/components/TeamComment.js
```diff
@@ -0,0 +1,60 @@
+/* @flow */
+
+import * as React from 'react';
+import { Localized } from '@fluent/react';
+
+import './CommentList.css'
```
adngdb: That looks fishy. Why do you import that CSS file, instead of the module itself? Can you make it so you use the `CommentsList` module directly?

adngdb: (Also, it should be `Comment*s*List.css` ;) )

frontend/src/core/comments/actions.js
abowler2: That makes sense.  Since this is included in the `get` function of `otherlocales/actions.js` I thought it was a needed check.

I should still call `api.entity.abort()` though, correct?
(history, otherlocales, and machinery all make this call but entities does not)

frontend/src/core/comments/components/TeamComment.js
abowler2: `CommentsList` renders differently than `TeamComments` in regards to users who are not authorized to leave comments and what renders when there are no comments. However, they share most of the CSS for displaying the list of comments and the add comment textarea.

I can copy the styling into the `TeamComments.css` file but that doesn't feel right either. 
@mathjazz Do you have a suggestion here?

mathjazz: I think you should be able to modify `CommentsList` such that you could use it in `TeamComments`.

abowler2:
Latest changes 😊 
___
Notes/questions:

- I successfully merged the `TeamComments` and the `CommentsList` files and chose to keep the `CommentsList` naming as it felt more accurate. Let me know if you prefer a different name there.

- When changing `translation` (formerly `translationId`) to optional I ran into issues with Flow in the `Helpers.js` and `History.js` files requiring that I use `any` instead of `?number` within the `addComment` function prop.  Let me know if this is what should be used everywhere or if there is some other way to handle this.

- I'm not sure about the way I handled the validation in ActionLog. It was the only way I could successfully check the comment arguments to verify they logged correctly, but it doesn't seem scalable if any additional items need to be added to the log.

- When `TeamComments` save it feels a little 'rough' in that you can visibly see the refresh (it is especially noticeable when there are multiple comments). Is there a way to smooth that out?

- Also, there were a few times when I was testing that the comments did not render in the correct order (ie...the newer comment rendered above the older comment, see image below). This only happened a few times, so I'm not sure how to reproduce it or what caused it. Perhaps there is a way to set the order so this does not happen?

Screen Shot 2020-02-12 at 4 02 36 PM

- Lastly, I'm pretty sure the failing tests are not on my end as I ran all tests successfully before pushing and they seem to be related to `setup.py` which I have not touched.  However, if it is in fact something I need to fix, just let me know 😁

mathjazz:
> Lastly, I'm pretty sure the failing tests are not on my end as I ran all tests successfully before pushing and they seem to be related to `setup.py` which I have not touched. However, if it is in fact something I need to fix, just let me know 😁

I'm afraid I have some bad news... :-) If you look at the [Travis log](https://travis-ci.org/mozilla/pontoon/builds/650269728?utm_source=github_status&utm_medium=notification) you'll notice there's a conflicting migration (0151), because we recently merged one to master. So you'll need to rename yours to 0152 and set its dependency to `0151_permalink_prefix_optional`.

pontoon/base/models.py
```diff
@@ -3441,3 +3445,14 @@ def serialize(self):
...
             "id": self.id,
         }
+
+    def save(self, *args, **kwargs):
+        """
+        Validate Comments before saving.
+        """
+        if (self.translation is not None and (not self.locale and not self.entity)) or (
+            self.translation is None and (self.locale and self.entity)
+        ):
```
mathjazz: You don't need to use `None` for the `translation` either, so I suggest you unify and simplify the checks:
```python
if (self.translation and not self.locale and not self.entity) or (
    not self.translation and self.locale and self.entity
):
```

frontend/src/modules/commentslist/components/CommentsList.css
```diff
@@ -23,3 +23,19 @@
...
+    margin-left: .5em;
+}
+
+.no-team-comments > p {
+    border-bottom: 1px solid #5E6475;
+    color: #aaa;
+    font-style: italic;
+    line-height: 22px;
+    padding: 15px 10px;
+}
```
mathjazz: This is specific to TeamComments, so it should go to `TeamComments.css`.

frontend/src/core/comments/actions.js
adngdb: Yes. The reason why we have this abort mechanism is that, if a user changes entities quickly, several requests are sent for helpers tabs content, and because network is unreliable, sometimes the results for request `t` would arrive after results for request `t+1`, thus showing the wrong content. To avoid that, when we start a new request for data, we abort the previously running request.

Note that if you copy/pasted this from somewhere else, it's likely that somewhere else is incorrect as well… :see_no_evil:

pontoon/base/views.py
```diff
@@ -436,39 +435,62 @@ def get_translation_history(request):
...
+    if translationId:
+        translation = get_object_or_404(Translation, pk=translationId)
 
-    c = Comment(author=user, translation=translation, content=comment,)
+    if translationId:
+        c = Comment(author=user, translation=translation, content=comment)
+        log_action("translation_comment:added", user, translation=translation)
+    else:
+        c = Comment(author=user, entity=entity, locale=locale, content=comment)
+        log_action("team_comment:added", user, entity=entity, locale=locale)
```
mathjazz: I don't think we need to log two separate actions - we can see what comment type it is by looking at the arguments (translation vs. entity+locale). So I'd revert this change (and related model, validation and migration changes).

Pike:
Fly-by comment, we should start using `on_delete` on all ForeignKeys to prep for the django update.

frontend/src/modules/commentslist/components/CommentsList.css
abowler2: `TeamComments` is now merged with `CommentsList` as per [suggested](https://github.com/mozilla/pontoon/pull/1561#discussion_r378891678). 
So there currently is no `TeamComments.css`, and if I add that file back in then we are at a similar place with 2 CSS modules being imported.

pontoon/base/views.py
abowler2: That is how I originally set this up, but had issues because what is True for one comment is False for the other. So, what validation error do I apply?

frontend/src/modules/commentslist/components/CommentsList.css
mathjazz: They shouldn't be merged. There should be a `TeamComments` component, which renders the `CommentsList` component or the "No comments message".

pontoon/base/views.py
mathjazz: The logic should be the same as in the `Comment` model:
https://github.com/mozilla/pontoon/pull/1561/files#diff-c18ce63ddb8aca155c0553a0c1b174e3R3449

abowler2:
> I'm afraid I have some bad news... :-) If you look at the [Travis log](https://travis-ci.org/mozilla/pontoon/builds/650269728?utm_source=github_status&utm_medium=notification) you'll notice there's a conflicting migration (0151), because we recently merged one to master. So you'll need to rename yours to 0152 and set its dependency to `0151_permalink_prefix_optional`.

Sigh....of course it was me 😐 
Since I'll be reverting and eliminating my migration, do I need to do anything here to avoid this conflict?

mathjazz:
>  Since I'll be reverting and eliminating my migration, do I need to do anything here to avoid this conflict?

First, roll back to previous migration (0150).
Next, rebase against master.
Then, rename your migration (to 0152) and edit it to change its dependency.
Finally, apply migrations.

frontend/src/core/comments/components/CommentsList.js
```diff
@@ -2,51 +2,80 @@
...
+            { !canComment ? null :
+                <AddComment
+                    user={ user.nameOrEmail }
+                    username={ user.username }
+                    imageURL={ user.gravatarURLSmall}
+                    addComment={ addComment }
+                />
+            }
+        </div>
+    }
```
mathjazz: You are repeating the gist of the component for each of its use cases, which destroys the purpose of factoring this code out into a reusable component.

One return should be sufficient, and you should modify the props accordingly:
```js
return <div className='comments-list'>
    <ul>
        { comments.map(comment =>
            <Comment
                comment={ comment }
                key={ comment.id }
            />
        )}
    </ul>
    { !canComment ? null :
        <AddComment
            user={ user }
            username={ username }
            imageURL={ imageURL}
            translation={ translation }
            addComment={ addComment }
        />
    }
</div>;
```

abowler2:
> Thanks for the update. I don't think `commentslist` should be a module - it is shared between two components, so it should be moved back to `core/comments`. Once that's done I'll have another look, because the diff will change a lot I have some other suggestions. :)

OK.  I think it is ready for your 'other suggestions' 😁

frontend/src/modules/history/components/History.js
```diff
@@ -21,7 +21,7 @@ type Props = {|
     locale: Locale,
     user: UserState,
     deleteTranslation: (number) => void,
-    addComment: (string, number) => void,
+    addComment: (string, any) => void,
```
mathjazz: Maybe @adngdb has an idea why `?number` makes flow unhappy here and also in `Helpers.js`?
```
Cannot create Translation element because number [1] is incompatible
with null or undefined [2] in the second argument of property addComment.
```

frontend/src/modules/teamcomments/components/TeamComment.js
```diff
@@ -0,0 +1,59 @@
...
+    const { teamComments, user, addComment } = props;
+
+    let canComment = user.isAuthenticated;
+    const comments = teamComments.comments;
+
+    if (!comments) {
+        return null;
+    }
+
+    if (!teamComments.comments.length) {
```
mathjazz:
```js
if (!comments.length) {
```

frontend/src/modules/teamcomments/components/TeamComment.js
```diff
@@ -0,0 +1,59 @@
...
+import type { TeamCommentState } from 'modules/teamcomments';
+
+type Props = {|
+    teamComments: TeamCommentState,
+    user: UserState,
+    addComment: (string, ?number) => void,
+|};
+
+
+export default function TeamComment(props: Props) {
```
mathjazz: This component should be called `TeamComments`.

frontend/src/modules/teamcomments/components/TeamComment.test.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import TeamComment from './TeamComment';
+
+describe('<TeamComment>', () => {
+    const DEFAULT_USER = 'AndyDwyer';
+
```
mathjazz: We should add a test case that the `CommentsList` is rendered if there are comments.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -56,7 +56,13 @@ export default function AddComments(props: Props) {
                 return null;
             }
 
-            addComment(comment, translationId);
+            if (translation) {
+                addComment(comment, translation);
+            }
+            else {
+                addComment(comment);
+            }
```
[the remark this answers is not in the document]
abowler2: Not anymore. My original setup did not always pass a value for `translation` but with the changes it will either be a number or null, so this is no longer needed 😊

frontend/src/modules/history/components/History.js
adngdb: Because the function is defined like this:

```js
addComment = (comment: string, translation?: number) => {
```

and instead it should be defined like this:

```js
addComment = (comment: string, translation: ?number) => {
```

However, please don't ask me why, as I have no freaking clue. :/

frontend/src/core/comments/index.js
```diff
@@ -1,16 +1,5 @@
...
 export { default as Comment } from './components/Comment';
-export { default as CommentCount } from './components/CommentCount';
 export { default as CommentsList } from './components/CommentsList';
-export { default as TeamComment } from './components/TeamComment';
-
-export type { TeamCommentState } from './reducer';
-
-// Name of this module.
-// Used as the key to store this module's reducer.
-export const NAME: string = 'comments';
```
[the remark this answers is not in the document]
adngdb: Scratch that, it's outdated.

mathjazz:
Not sure if this was ready for another look, but the code looks great. Thanks for the update!

Things that remain:
- my CSS work
- comment flashing on submission
- nice to have: `on_delete` attribute on `ForeignKeys`

abowler2:
> Not sure if this was ready for another look, but the code looks great. Thanks for the update!
> 
> Things that remain:
> 
>     * my CSS work
> 
>     * comment flashing on submission
> 
>     * nice to have: `on_delete` attribute on `ForeignKeys`

Yes, sorry I should have commented that it was ready for review.

Regarding, the `on_delete` on ForeignKeys:  Those were added as default in the migration process (`on_delete=django.db.models.deletion.CASCADE,`), but do we need to explicitly add it in the model as well?

Pike:
> Regarding, the `on_delete` on ForeignKeys: Those were added as default in the migration process (`on_delete=django.db.models.deletion.CASCADE,`), but do we need to explicitly add it in the model as well?

Yes, https://docs.djangoproject.com/en/2.2/releases/2.0/#features-removed-in-2-0 mentions that they required in both models and migrations starting in 2.0. In our process to migrate to that, we should not add more ForeignKeys w/out explicit `on_delete`.

abowler2:
I haven't been able to work out what is causing the flashing on the team comments. The only major difference between them and the translation comments is that they process through a form when submitted.  

If that is indeed the cause then I'm not sure how to fix that and would appreciate your thoughts and suggestions.

abowler2:
Update - found the issue and just pushed the fix

mathjazz:
The code looks great and the patch works fine!

As promised, here are my CSS fixes. I've also disabled the "No comments" message for users that can submit the comments - for them we now only show the `AddComment` form.

```diff
diff --git a/frontend/src/core/comments/components/CommentsList.css b/frontend/src/core/comments/components/CommentsList.css
index 72ee8cb4..34d8b1e6 100644
--- a/frontend/src/core/comments/components/CommentsList.css
+++ b/frontend/src/core/comments/components/CommentsList.css
@@ -1,4 +1,5 @@
 .comments-list {
+    line-height: 22px;
     padding: 8px;
 }
 
diff --git a/frontend/src/modules/teamcomments/components/TeamComments.css b/frontend/src/modules/teamcomments/components/TeamComments.css
index 50d7a842..4ee4bd7e 100644
--- a/frontend/src/modules/teamcomments/components/TeamComments.css
+++ b/frontend/src/modules/teamcomments/components/TeamComments.css
@@ -1,11 +1,11 @@
-.team-comments-list ul .comment,
-.team-comments-list div .add-comment {
-    margin-left: 0;
+.team-comments .comments-list {
+    margin-top: 2px;
 }
 
-.team-comments-list ul,
-.team-comments-list div .add-comment {
-    margin-left: .5em;
+.team-comments ul,
+.team-comments ul .comment,
+.team-comments div .add-comment {
+    margin-left: 0;
 }
 
 .no-team-comments > p {
diff --git a/frontend/src/modules/teamcomments/components/TeamComments.js b/frontend/src/modules/teamcomments/components/TeamComments.js
index 53987e3d..55e17253 100644
--- a/frontend/src/modules/teamcomments/components/TeamComments.js
+++ b/frontend/src/modules/teamcomments/components/TeamComments.js
@@ -28,26 +28,27 @@ export default function TeamComments(props: Props) {
     }
 
     if (!comments.length) {
-        return <section>
-            <div className="no-team-comments">
-                <Localized id="entitydetails-Helpers--no-comments">
-                    <p>No comments available.</p>
-                </Localized>
-            </div>
-            <div className='comments-list team-comments-list'>
-                { !canComment ? null :
+        return <section className="team-comments">
+            { !canComment ?
+                <div className="no-team-comments">
+                    <Localized id="entitydetails-Helpers--no-comments">
+                        <p>No comments available.</p>
+                    </Localized>
+                </div>
+                :
+                <div className="comments-list">
                     <AddComment
                         user={ user.nameOrEmail }
                         username={ user.username }
                         imageURL={ user.gravatarURLSmall}
                         addComment={ addComment }
                     />
-                }
-            </div>
+                </div>
+            }
         </section>
     }
 
-    return <section className="team-comments-list">
+    return <section className="team-comments">
         <CommentsList
             comments={ comments }
             user={ user }
```

frontend/src/core/comments/actions.js
```diff
@@ -0,0 +1,38 @@
...
+
+import api from 'core/api';
+import * as notification from 'core/notification';
+import * as history from 'modules/history';
+import * as teamcomments from 'modules/teamcomments';
+
+
+export function addComment(
+    entity: number,
+    locale: string,
```
adngdb: `entity` and `locale` are only needed when adding a team comment, and `translation` is only needed when adding a translation comment, right? If so, why always require `entity` and `locale`? Could it work to only require either `entity` + `locale` or `translation`?

frontend/src/modules/teamcomments/components/TeamComments.js
```diff
@@ -0,0 +1,60 @@
...
+    if (!comments.length) {
+        return <section className="team-comments">
+            { !canComment ?
+                <div className="no-team-comments">
+                    <Localized id="entitydetails-Helpers--no-comments">
+                        <p>No comments available.</p>
+                    </Localized>
+                </div>
+                :
+                <div className='comments-list'>
```
adngdb: This case is already handled by `CommentsList`. Simply pass an empty list of comments and it will display the form only. You thus should not be repeating this code here, and count on `CommentsList` to display the form when there are no comments.

pontoon/base/models.py
```diff
@@ -3442,3 +3454,14 @@ def serialize(self):
...
+    def save(self, *args, **kwargs):
+        """
+        Validate Comments before saving.
+        """
+        if (self.translation and not self.locale and not self.entity) or (
+            not self.translation and self.locale and self.entity
+        ):
+            super(Comment, self).save(*args, **kwargs)
+        else:
+            raise ValidationError("Invalid comment arguments")
```
adngdb: Generally, it's better to reverse those conditions: first look for the failures, and let the normal behavior resolve outside your condition block. Since you're using a `raise` here, you don't need to `else` blocks, allowing to indent code a little less. I've also found it more natural to reason about. 

```py
if not (condition):
    raise ValidationError()

super(Comment, self).save()
```

frontend/src/core/comments/actions.js
abowler2: Within that function the calls to `get` require `locale` and `entity`

mathjazz: Right, we need to update history.

I'd be nice to make them optional in `AddCommentsForm`.

frontend/src/modules/teamcomments/components/TeamComments.js
```diff
@@ -36,14 +36,14 @@ export default function TeamComments(props: Props) {
...
-                        imageURL={ user.gravatarURLSmall}
+                <section className="team-comments">
+                    <CommentsList
+                        comments={ comments }
+                        user={ user }
+                        canComment={ canComment }
                         addComment={ addComment }
                     />
-                </div>
+                </section>
```
mathjazz: This entire `section.team-comments` is repeated below. We should only have it once. You can achieve that by changing the condition.

abowler2:
Sorry....fixing test right now 😊

adngdb:
@mathjazz Shouldn't we put the email address on top of the comment? It's a lot of empty space at the moment.

Screenshot_2020-02-21 French (fr) · Pontoon test

mathjazz:
> @mathjazz Shouldn't we put the email address on top of the comment? It's a lot of empty space at the moment.

No, it should be in the same line, BUT the following lines should start at the same position as the user identifier. See how Facebook does that. Nice catch!

Screenshot 2020-02-21 at 10 53 45

abowler2:
Updated the layout so the text wraps under the email/username.  Let me know if I need to adjust the spacing between them (and any other needed adjustments of course 😄 )

frontend/src/core/comments/components/Comment.css
```diff
@@ -3,6 +3,7 @@
 }
 
 .comments-list .comment a {
+    margin-right: 6px;
```
mathjazz: This needs to go into a more specific selector, otherwise it adds margin to the avatar as well.

Also, 4px seems enough.

mathjazz:
OK, one final thing - as @adngdb suggested, we should add a add button for submitting comments.

Design:
Screenshot 2020-02-21 at 17 34 43

The icon is `fa-paper-plane`. The `color` should change from `#CCC` to `#FFF` on hover.

abowler2:
Added the button, but you are most likely going to have to work your CSS magic on it 🤣 

I couldn't get the initial height of the `textarea` to match that of the comments, and not sure if the button is supposed to grow with the `textarea` but it does ATM, so if not that will need to change.

frontend/src/core/comments/components/CommentsList.css
```diff
@@ -18,9 +18,3 @@
     width: 32px;
     height: 32px;
 }
-
-.comments-list .comment .container {
-    display: flex;
-    flex-flow: column wrap;
-    width: 100%;
-}
```
mathjazz: Why only move this CSS rule to `Comment.css`, but not the other `.comments-list .comment` rules?

Also, please remove the `margin-left: 4px;` from the `.comments-list .comment .content p` rule in the `Comment.css` file, because it's no longer needed and makes passing inconsistent between submitted comments and the add comment form.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -47,28 +47,26 @@ export default function AddComments(props: Props) {
...
+        commentInput.current.rows = minRows;
     };
 
     return <div className='comment add-comment'>
         <UserAvatar
             username={ username }
             imageUrl={ imageURL }
         />
-        <form className='container'>
+        <form className='form-container'>
```
mathjazz: I'd leave the `.container` class here, and then just override 1 differing CSS rule (`flex-flow`) with a more specific selector, e.g. `comments-list .comment.add-comment .container`.

abowler2:
>     1. I wonder if we should keep the ability to "Submit with Enter". I'd say yes, because that's what we do in pretty much all other forms. What do you think? I also wonder what @adngdb thinks who requested this feature. :)
> 

I personally prefer to submit on 'enter', but the hesitation I would have is that we don't have a way to edit or delete comments. So, if one intended to just go to a new line but forgot to `shift-enter` they may find it frustrating.

I'll wait until we make a final decision on this before tackling the CSS notes as they will no longer be relevant if we revert back to "submit on enter"

mathjazz:
I agree that submitting with Enter by mistake is likely to happen. I'd say that's more of an argument for the ability to Edit/Delete then against the keyboard shortcut.

Note that the button will always be there - we're just wondering if we should restore the keyboard shortcut. So the decision should not affect the CSS work.

frontend/src/core/comments/components/CommentsList.css
abowler2: > Why only move this CSS rule to `Comment.css`, but not the other `.comments-list .comment` rules?
> 
> Also, please remove the `margin-left: 4px;` from the `.comments-list .comment .content p` rule in the `Comment.css` file, because it's no longer needed and makes passing inconsistent between submitted comments and the add comment form.

I only moved this one because it was the only one no longer being shared by both comment and add comment.  However, the latest changes reverses this and they now share most of these styles and so I moved it back.

abowler2:
Just pushed the updated CSS per your notes.
____

I'm not sure why that test would have failed...all tests pass locally, and I only made CSS changes

mathjazz:
> I'm not sure why that test would have failed...all tests pass locally, and I only made CSS changes

Fixed :)

mathjazz:
This looks good to me. Leaving the decision about "Submit with Enter" to @adngdb:
https://github.com/mozilla/pontoon/pull/1561#pullrequestreview-363601921

adngdb:
I vote for submit on Enter. That's the default behavior now in many systems with "small" comments. We can revisit if users complain, though, as it might turn out that people will write longer comments than we expect, in which case we would need to change the behavior. Let's ship this thing simple, and change only if / when needed.

abowler2:
Ability to "Submit on Enter" is now active again 😃

frontend/src/core/comments/components/AddComment.js
```diff
@@ -80,9 +79,26 @@ export default function AddComments(props: Props) {
                     rows={ minRows }
                     ref={ commentInput }
                     onChange={ handleOnChange }
-                    onKeyDown={ submitComment }
+                    onKeyDown={ e =>
+                        (e.keyCode === 13 && e.shiftKey === false) ? submitComment(e) : null
+                    }
```
mathjazz: Please change this to `handleOnKeyDown` for consistency and clarity. I believe it will be easier to read.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -47,6 +47,12 @@ export default function AddComments(props: Props) {
         }
     }
 
+    const handleOnKeyDown = (event: SytheticKeyboardEvent) => {
+        if (event.keyCode === 13 && event.shiftKey === false) {
+            submitComment(event)
```
mathjazz: Nit: please add colon. :-)

frontend/src/core/comments/components/AddComment.css
```diff
@@ -1,11 +1,36 @@
...
+    color: #ccc;
+    display: flex;
+    align-items: center;
+    height: 36px;
+    margin-left: 6px;
+    padding: 0 12px;
+}
+
+.comments-list .add-comment .submit-button:hover {
+    color: #fff;
```
mathjazz: Let's make the highlight a bit more obvious and in line with the other buttons of similar style on the dashboard:
```css
color: #272A2F;
background: #7BC876;
```

That's my last request for change. I'm working on the "Threaded view" of Translation comments, and once I'm done I plan to conduct a quick round of testing before shipping either of these two versions.

mathjazz:
As already mentioned, I'll merge this as soon as we decide on the Translation Comments UX:
https://github.com/mozilla/pontoon/pull/1561#discussion_r384060696

mathjazz:
Hi @abowler2, could you please apply this path?

It's a collection of small improvements found during testing:
- Autofocus AddComment textarea
- Simplify Team comments DOM
- Square Send Comment button
- Highlight comment toggle if comments exist
- Show divider if comments exist and cannot comment

```diff
diff --git a/frontend/src/core/comments/components/AddComment.css b/frontend/src/core/comments/components/AddComment.css
index d2c2c61c..cd773dc2 100644
--- a/frontend/src/core/comments/components/AddComment.css
+++ b/frontend/src/core/comments/components/AddComment.css
@@ -27,8 +27,9 @@
     display: flex;
     align-items: center;
     height: 36px;
+    width: 36px;
     margin-left: 6px;
-    padding: 0 12px;
+    padding-left: 10px;
 }
 
 .comments-list .add-comment .submit-button:hover {
diff --git a/frontend/src/core/comments/components/AddComment.js b/frontend/src/core/comments/components/AddComment.js
index 8746c4dc..1c9487c3 100644
--- a/frontend/src/core/comments/components/AddComment.js
+++ b/frontend/src/core/comments/components/AddComment.js
@@ -78,6 +78,7 @@ export default function AddComments(props: Props) {
                 attrs={{ placeholder: true }}
             >
                 <textarea
+                    autoFocus
                     id='comment-input'
                     name='comment'
                     dir='auto'
diff --git a/frontend/src/modules/history/components/Translation.css b/frontend/src/modules/history/components/Translation.css
index dd9118da..3ead3ca5 100644
--- a/frontend/src/modules/history/components/Translation.css
+++ b/frontend/src/modules/history/components/Translation.css
@@ -82,8 +82,9 @@
 }
 
 .history .translation .content > header .toggle-diff:hover,
+.history .translation .content > header .toggle-diff.active,
 .history .translation .content > header .toggle-comments:hover,
-.history .translation .content > header .toggle-diff.active {
+.history .translation .content > header .toggle-comments.active {
     color: #7BC876;
 }
 
diff --git a/frontend/src/modules/history/components/Translation.js b/frontend/src/modules/history/components/Translation.js
index 41bcd793..4ffa7b96 100644
--- a/frontend/src/modules/history/components/Translation.js
+++ b/frontend/src/modules/history/components/Translation.js
@@ -175,7 +175,7 @@ export class TranslationBase extends React.Component<InternalProps, State> {
                 $commentCount={ commentCount }
             >
                 <button
-                    className='toggle-comments'
+                    className='toggle-comments active'
                     title='Toggle translation comments'
                     onClick={ this.toggleComments }
                 >
@@ -301,7 +301,7 @@ export class TranslationBase extends React.Component<InternalProps, State> {
 
                             { this.renderDiffToggle() }
 
-                            { (index === 0 || !canComment) ? null : <span className='divider'>&bull;</span> }
+                            { (index === 0 || (!canComment && commentCount === 0)) ? null : <span className='divider'>&bull;</span> }
 
                             { (!canComment && commentCount === 0) ? null : this.renderCommentToggle(commentCount) }
 
diff --git a/frontend/src/modules/teamcomments/components/TeamComments.css b/frontend/src/modules/teamcomments/components/TeamComments.css
index 4ee4bd7e..14898c24 100644
--- a/frontend/src/modules/teamcomments/components/TeamComments.css
+++ b/frontend/src/modules/teamcomments/components/TeamComments.css
@@ -8,7 +8,7 @@
     margin-left: 0;
 }
 
-.no-team-comments > p {
+.no-team-comments {
     border-bottom: 1px solid #5E6475;
     color: #aaa;
     font-style: italic;
diff --git a/frontend/src/modules/teamcomments/components/TeamComments.js b/frontend/src/modules/teamcomments/components/TeamComments.js
index 4a4a16eb..0e613258 100644
--- a/frontend/src/modules/teamcomments/components/TeamComments.js
+++ b/frontend/src/modules/teamcomments/components/TeamComments.js
@@ -29,20 +29,16 @@ export default function TeamComments(props: Props) {
 
     return <section className="team-comments">
         { !comments.length && !canComment ?
-            <div className="no-team-comments">
-                <Localized id="entitydetails-Helpers--no-comments">
-                    <p>No comments available.</p>
-                </Localized>
-            </div>
+            <Localized id="entitydetails-Helpers--no-comments">
+                <p className="no-team-comments">No comments available.</p>
+            </Localized>
             :
-            <section className="team-comments">
-                <CommentsList
-                    comments={ comments }
-                    user={ user }
-                    canComment={ canComment }
-                    addComment={ addComment }
-                />
-            </section>
+            <CommentsList
+                comments={ comments }
+                user={ user }
+                canComment={ canComment }
+                addComment={ addComment }
+            />
         }
     </section>
 }
```

abowler2:
> Hi @abowler2, could you please apply this path?
> 
> It's a collection of small improvements found during testing:
> 
>     * Autofocus AddComment textarea
> 
>     * Simplify Team comments DOM
> 
>     * Square Send Comment button
> 
>     * Highlight comment toggle if comments exist
> 
>     * Show divider if comments exist and cannot comment
> 
> 

Done :)

mathjazz:
@adngdb I've implemented 2 concepts from the "Variant B" implementation:
- I've added the "Comment Summary widget"
- I've removed comment count from "Comment toggle"

Could you please review?

mathjazz:
I wonder if we should add "Collapse comments" button also at the bottom (under the `AddComment` form).

frontend/src/modules/history/components/Translation.js
```diff
@@ -168,6 +168,78 @@ export class TranslationBase extends React.Component<InternalProps, State> {
         </Localized>
     }
 
+    renderCommentSummary() {
```
adngdb: That totally should be a component! :-)

frontend/src/modules/history/components/Translation.js
```diff
@@ -168,6 +168,78 @@ export class TranslationBase extends React.Component<InternalProps, State> {
...
+        const comments = this.props.translation.comments;
+        const commentCount = comments.length;
+
+        if (commentCount === 0) {
+            return null;
+        }
+
+        const authors = [...new Set(comments.map(comment => comment.username))];
+        const lastComment = [...comments].pop();
```
adngdb: There's no need to copy the `comments` array here. You can simply do `const lastComment = comments[commentCount - 1];`.

frontend/src/modules/history/components/Translation.js
```diff
@@ -168,6 +168,78 @@ export class TranslationBase extends React.Component<InternalProps, State> {
...
 
+    renderCommentSummary() {
+        const comments = this.props.translation.comments;
+        const commentCount = comments.length;
+
+        if (commentCount === 0) {
+            return null;
+        }
+
+        const authors = [...new Set(comments.map(comment => comment.username))];
```
adngdb: You could return an object in your `map` call here, including both `username` and `userGravatarUrlSmall`, then not need to do another recursion below when displaying user avatars.

frontend/src/modules/history/components/Translation.js
```diff
@@ -168,6 +168,78 @@ export class TranslationBase extends React.Component<InternalProps, State> {
...
+        const lastComment = [...comments].pop();
+
+        return <div
+            className='comment-summary'
+            onClick={ this.toggleComments }
+            title=''
+        >
+            <ul>
+                { authors.map((author, index) => {
+                    const comment = comments.filter(comment => (comment.username === author))[0];
```
adngdb: This could be removed, see previous comment.

mathjazz:
> * The "COMMENT" link is very hard to discover. The contrast is very low, and I find it almost impossible to notice next to the review actions.

That's by design. An action to add a comment should be deprioritized in the UI compared to the review actions, because it's going to be used significantly less often. We've decided against using the icon in Paris. I've been thinking about adding a hamburger menu on a longer term to hold items like Show Diff and Add Comment.

> * The comments section doesn't get the same background color as the translation on hover.

That's also by design. :-) We copy translation into the textarea on click and we use hover affect (and a tooltip) as a hint for that.

> * It's a bit confusing that the comments widget appears only when there are comments. It creates an inconsistent user experience.

We want to make it obvious that there are comments, hence the contrast. It's as inconsistent as the History panel is when there are / aren't any comments. :-)

What I see an inconsistency is that there are two comment toggles when comments are present. I'll see if I can come up with some minimalistic widget when there aren't any comments. We shouldn't hurt the review process by making the UI more complex than necessary for the most common (i.e. no comments) scenario.

> * How does the comments widget behave when there are many different comment authors? I haven't tried it but it seems avatars just pile up. It might break at some point? (Note that's an edge case that I'm happy to leave aside for the moment. )

I was tested for this purpose, but I think I broken it a little with latest CSS changes. :-)

mathjazz:
@adngdb Thanks for the feedback!

I haven't addressed your suggestions for improving the code, because I'm not sure about the UI yet. The latest commit follows the spec closely (i.e. no Comment Summary widget), the only thing it changes compared to April's implementation is that when comments are present, the button to reveal comments gets highlighted.

If you revert the last two commits (so you end up on 7da7f55), you'll see the next iteration of the Comment Summary approach. It makes the Comment Summary visible also when there aren't any comments, but only on hover. I find this unusable, because it makes the UI jumpy and review icons hard to reach. Still wanted to hear your thoughts, because that's the best I could do with unifying the Comment Summary experience for the case  when there are or aren't any comments.

Sorry for making this more confusing to test than necessary.

adngdb:
@mathjazz I don't like the jumpy comment summary widget either. The latest changes seem like a fine compromise to me.

PR #1562, opened 2020-02-10
Bug 1602490 - Specification for the private projects feature.

adngdb:
https://bugzilla.mozilla.org/show_bug.cgi?id=1602490

This feature has originally been asked for by Wanadev. Mozilla would benefit from a part of this feature (making projects private for the sake of configuration), and even if we don't need the second part (controlling a list of approved users), we estimated that it would not hurt Pontoon, so we are happy to accept this feature.

Note that this is our very first specification proposal, we're testing this process out. Comments both on the content *and* on the form are very welcome.


adngdb:
@jotes and @VishalCR7, I'd be happy to have your feedback on this as well.

0100-private-projects.md
```diff
@@ -0,0 +1,36 @@
+- Feature Name: Private Projects
+- Created: 2020-02-10
```
mathjazz: Is this information needed, given that we're storing everything in the repository?

0100-private-projects.md
```diff
@@ -0,0 +1,36 @@
...
+
+# Summary
+
+Add an option to make a project "private", restricting who can see and interact with it.
+
+# Motivation
+
+It can be useful in some cases to be able to hide a project from users. For example, when a superuser creates a new project, they might want to make sure that everything is correctly set up before exposing it — running sync, verifying that it is enabled for the right locales, etc. In such a case, the project being not visible from non-superusers is enough. Once the superuser is done with configuration, they simply make the project public.
+
+For a different use-case, let's consider an agency working with several clients. They have a number of projects created on Pontoon, and they want to have specific localizers working on specific projects. It would be beneficial for them to able to restrict seeing a project to only a defined list of users. This requires being able to mark a project as private, and also having a configuration option to select a list of users who will be able to see, and interact, with that project. That would allow a clear separation of projects, reducing the risk of localizers translating the wrong set of projects.
```
mathjazz: Grammar: beneficial for them to **be**  able to.

0100-private-projects.md
```diff
@@ -0,0 +1,36 @@
...
+- a simple checkbox to make the project private or public;
+- a formset controlling which users can see the project when it is private.
+
+The "approved users" formset is visible only when the project is marked as private. It has two elements: a list of users, showing who currently has access (or will upon saving), and a search input, used to look for a specific user in the database and adding that user to the list. Clicking a user in the list removes that user from the list.
+
+Administrators and managers of a project always have access to it, regarless of whether they are in the approved users list or not.
+
+On each dashboard listing projects (`/projects/`, `/{locale}/`), private projects will not appear unless the user is authenticated and is in the list of approved users.
+
+Pages related to a project (`/{project}/`, `\{locale}/{project}/`, `/{locale}/{project}/{resource}/`) will return a 404 Not Found error unless the user is authenticated and is in the list of approved users. That should also be the case for back-end endpoints used to modify data, as a user should not be able to interact at all with a private project if they are not approved for that project.
```
mathjazz: `/{project}/` -> `/projects/{project}/`

mathjazz: We should also exclude private projects from permissions pages: `\{locale}/permissions/`.

0100-private-projects.md
```diff
@@ -0,0 +1,36 @@
...
+
+For a different use-case, let's consider an agency working with several clients. They have a number of projects created on Pontoon, and they want to have specific localizers working on specific projects. It would be beneficial for them to able to restrict seeing a project to only a defined list of users. This requires being able to mark a project as private, and also having a configuration option to select a list of users who will be able to see, and interact, with that project. That would allow a clear separation of projects, reducing the risk of localizers translating the wrong set of projects.
+
+# Feature explanation
+
+When creating or editing a project, a superadmin has two new options:
+
+- a simple checkbox to make the project private or public;
```
mathjazz: If PMs will usually create projects as private first for testing purposes, the "private" setting should be "True" by default. Once the project is synced and private, we can make it more obvious that it needs to be made "public" before others can see them.

0100-private-projects.md
```diff
@@ -0,0 +1,36 @@
...
+- a formset controlling which users can see the project when it is private.
```
mathjazz: We should probably use the similar UX to locale selector in settings/admin or user selector in the permissions page. So two lists - one for "approved" (is "privileged" less confusing?) and the other for available users. There should also be an option to copy users from another project (same as we have for copying locales), because the list of privileged users is likely to repeat if I understand the use case correctly.

0100-private-projects.md
```diff
@@ -0,0 +1,36 @@
...
+
+The "approved users" formset is visible only when the project is marked as private. It has two elements: a list of users, showing who currently has access (or will upon saving), and a search input, used to look for a specific user in the database and adding that user to the list. Clicking a user in the list removes that user from the list.
+
+Administrators and managers of a project always have access to it, regarless of whether they are in the approved users list or not.
```
jotes: `regardless`

0100-private-projects.md
jotes: Private projects shouldn't be visible (for the non-authorized users) in the sync log too?

0100-private-projects.md
jotes: Additionally, what if a translator decides to pre-translate some strings when the project is private? IMHO Such contributions shouldn't be visible in the action log.

0100-private-projects.md
```diff
@@ -0,0 +1,36 @@
...
+
+On each dashboard listing projects (`/projects/`, `/{locale}/`), private projects will not appear unless the user is authenticated and is in the list of approved users.
+
+Pages related to a project (`/{project}/`, `\{locale}/{project}/`, `/{locale}/{project}/{resource}/`) will return a 404 Not Found error unless the user is authenticated and is in the list of approved users. That should also be the case for back-end endpoints used to modify data, as a user should not be able to interact at all with a private project if they are not approved for that project.
+
```
jotes: What about Translation Memory? Strings from private projects shouldn't be visible until they get public, is it true?

0100-private-projects.md
Pike: Do we need the concept of a Client here?

0100-private-projects.md
```diff
@@ -0,0 +1,36 @@
...
+For a different use-case, let's consider an agency working with several clients. They have a number of projects created on Pontoon, and they want to have specific localizers working on specific projects. It would be beneficial for them to able to restrict seeing a project to only a defined list of users. This requires being able to mark a project as private, and also having a configuration option to select a list of users who will be able to see, and interact, with that project. That would allow a clear separation of projects, reducing the risk of localizers translating the wrong set of projects.
+
+# Feature explanation
+
+When creating or editing a project, a superadmin has two new options:
+
+- a simple checkbox to make the project private or public;
+- a formset controlling which users can see the project when it is private.
+
+The "approved users" formset is visible only when the project is marked as private. It has two elements: a list of users, showing who currently has access (or will upon saving), and a search input, used to look for a specific user in the database and adding that user to the list. Clicking a user in the list removes that user from the list.
```
Pike: Our Pontoon instance has a massive amount of users, most of which are dormant. We probably can't expose "everyone" on that list?

0100-private-projects.md
VishalCR7: Along with the option to copy users from another projects, Can we also have the ability to form a "user group" and allow copying from a user group. In fact, users from another project can be treated as a "user group" as well. I think this will add more flexibility to how a user is approved for a project.

0100-private-projects.md
VishalCR7: Also, how are we going to handle the Stats and their calculations for a privileged user and a non-privileged user? The stats will be different on the individual team page for a privileged and non-privileged user.

0100-private-projects.md
mathjazz: Groups would be ideal. They also require much more work than (reusing) the copy function.

adngdb:
>     1. Should we move the file to a `specifications` folder or something?

Totally, I had intended to do that and messed up.

>     2. Does `0100` in the file name represent a version? What do we need it for?

It's a way to sort files in the folder, I assume. Both PEPs and RFCs have that. 

Starting with `0100` is to have 100 "reserved" specs numbers, in case we want to have some specific ones in the future. For example, we might use `0000` as the template, `0001` as the spec to write specs, etc. 

>     3. I would add a "Success criteria" section, which we'd use to confirm if the feature is completed. It would consist of measurable goals and instructions for measuring them.

I'm not sure about that. Could you give me an example of criteria, for Private projects for example? What else would there be other than "build the feature"? 

> On top of that, I'd like us to be more formal on roles. I think this proposal has three names for one role, and it might not be what we're using in practice.

Should that be a specification? :innocent:

0100-private-projects.md
adngdb: I guess it's nice to have that there, instead of having to look at git history?

0100-private-projects.md
adngdb: I don't think it matters too much for sync log. I don't expect regular users to look at them (it's not even linked from anywhere in the site as far as I know), and it's not giving away too much information.

But as a side note, why are sync logs public? I'd expect them to be accessible only to superusers.

0100-private-projects.md
adngdb: Latest activity probably needs to be computed differently indeed, to not show private projects activity. 

Blocking translation however defeats the purpose of having this feature for a company like wanadev, as the ultimate goal to have only a selected group of users able to translate.

0100-private-projects.md
adngdb: Yes, that's why there are two elements to the UI: a list of currently approved users (which should be small), and an input to look for a specific user in the DB. It's never showing the full list of all users.

0100-private-projects.md
adngdb: We might, I'm not sure. How would you define it?

0100-private-projects.md
adngdb: Groups would indeed be great, but I guess that's another feature entirely? 

@mathjazz regarding the two lists, I assume one of them has the full list of all users in the instance? That would not work in our case, as we have several thousand users. The list would be too long to load and near unusable.

Copy from a project sounds good though, and important enough.

0100-private-projects.md
adngdb: I think it's fine to exposed strings in the Translation Memory. The goal here is not to pretend the projects do not exist, but to make sure only the right set of users can translate them.

Stats computation will be tricky indeed. @mathjazz don't we already have something similar, where stats can appear "wrong" because of hidden data?

0100-private-projects.md
mathjazz: We can limit access to sync log, but there's no real need for that as they aren't "giving away too much information". :)

Sometimes it's helpful it's public so that we can share links to particular sync logs in bugzilla.

0100-private-projects.md
mathjazz: Trying the permissions page, your assumption seems correct - it gets pretty slow. We should however consolidate the selector UX with the permissions page, because this is basically the same thing.

0100-private-projects.md
mathjazz: > I think it's fine to exposed strings in the Translation Memory. The goal here is not to pretend the projects do not exist, but to make sure only the right set of users can translate them.

That's good to know. I though private project would be secret (because "we don't want to reveal who our clients are"). We should clarify in the spec, either in the *Motivation* section or by adding the [*Out of scope*](https://docs.google.com/document/d/1W_MEYwtNn-_J0Uiu7gTSwc2B3oux75rjVtjGu-YoFi4/edit#) section.

> Stats computation will be tricky indeed. @mathjazz don't we already have something similar, where stats can appear "wrong" because of hidden data?

We exclude system projects and disabled projects from stats on most of the dashboards. Bogus stats calculations (due to which we need to fix stats after each sync job) are not related to that.

mathjazz:
> I'm not sure about that. Could you give me an example of criteria, for Private projects for example? What else would there be other than "build the feature"?

An example for our use case: "PMs no longer verify project setup on the stage server". Or something less binary, more measurable (but IMHO also more stupid 😛): "Project setup takes 30% less time".

We could also add some goals for 3rd parties, e.g. "Wanadev uses at least 2 private projects on their instance".

adngdb:
> > I'm not sure about that. Could you give me an example of criteria, for Private projects for example? What else would there be other than "build the feature"?
> 
> An example for our use case: "PMs no longer verify project setup on the stage server". Or something less binary, more measurable (but IMHO also more stupid stuck_out_tongue): "Project setup takes 30% less time".
> 
> We could also add some goals for 3rd parties, e.g. "Wanadev uses at least 2 private projects on their instance".

I struggle to understand the benefits of having that in the specification itself. It sounds great for goals and OKRs and stuff, and it would fit well in a KR to measure the success of a feature, but I don't see what benefits it brings to writing the spec itself. Also, I find those very vague, or nearly impossible to actually measure with our current solution. :/

mathjazz:
I think defining success criteria can:
* clarify the motivation for developing the feature better and
* help assess if the feature achieves the desired results

Since we no longer do OKRs, I think spec is the right home for this. Or the Roadmap?

Also note that I don't think we should sweat it too much. Without feature tracking in place to monitor how users are engaging with features and where improvements are needed, I suspect we'll often struggle to find success metrics.

I think it's worth including the part about dropping stage for project setup being the factor somewhere. It could also be part of the Motivation section, or part of an optional Success Criteria section, or part of the Roadmap.

adngdb:
I've added roles (and a note about TM). I'm not sure that brings much value, but maybe it does? What do you think, @mathjazz and @Pike?

BThibaut:
Permissions access of TM and Translators for private projects are good for us.

adngdb:
@mathjazz 
> One last thing here is to resolve the last chapter - the UI for selecting users.

I don't know what to do about that. Should we leave it as a question to deal with during development of the feature? 

@BThibaut 
> Permissions access of TM and Translators for private projects are good for us.

Thank you! :)

mathjazz:
> I don't know what to do about that. Should we leave it as a question to deal with during development of the feature?

I'd expect specifications to define the core pieces of the feature precisely enough for developers to start working on. I think this specification is doing that job very well, except for this particular piece. When the developer gets there, she'll basically need a specification for it.

specs/0100-private-projects.md
```diff
@@ -0,0 +1,52 @@
...
+- a simple checkbox to make the project private or public (private by default);
+- a formset controlling which users can see the project when it is private.
+
+The "approved users" formset is visible only when the project is marked as private. It has three elements:
+
+- a list of users, showing who currently has access (or will upon saving). Clicking a user in the list removes that user from the list;
+- a search input, used to look for a specific user in the database and adding that user to the list;
+- a "copy from project" button, allowing to copy the approved users list from an existing project.
+
+Locale managers and localizers of a project always have access to it, regardless of whether they are in the approved users list or not.
```
mathjazz: s/localizers/translators

mathjazz:
A quick mockup of how the admin UI can look like:

Screenshot_2020-03-27 Edit project(2)

We need a distinction between All Users and Project Managers, to prevent Locale Managers and Translators from having access while Project Managers are testing the setup. Project Managers should be the default.

Note that we intend to refactor the project admin significantly, so we'll don't need to sweat the design at this stage too much.

mathjazz:
Pasting this here from conversation between me and @adngdb:

> Yeah, that widget is likely to have performance issues, but we already use it in other places. Creating a new multi select widget (at least for long list like users, I think it works OK for locales) is a separate bug and spec I think, even if it blocks the "3rd party part" of the private projects feature.

I just noticed today that Django Admin has the exact same UI and it performs well with 1,000s of items: https://pontoon.mozilla.org/a/auth/user/4400/change/.

adngdb:
@mathjazz on that mockup, I don't understand why you added three visibility options. I think it would be easier to have a single checkbox, "Private", that if unchecked means everyone can see it, and if checked means only superusers + project admins + selected users can see it.

I'm not sure the distinction between "Project admins" and "Selected users" adds much value, but I'm quite sure it will add some complexity to the code. Why not go for the simplest option?

mathjazz:
Good point. Updated the mockup with select box: when "Private" option is selected, the user selector is revealed - a concept similar to the one we use for "Data source".

We could also use the checkbox instead of the select box.

77749552-7f96a700-7022-11ea-8803-3897581d9317

mathjazz:
Hm, now I remember why I used three options:

> Locale managers and localizers of a project always have access to it, regardless of whether they are in the selected users list or not.

I guess we need all three anyways? We don't want to expose Private projects during the setup phase to anyone except for Project Managers (not even Locale Managers and Localizers). So we need to distinct between two types of Private projects:
1. Accessible to Project Managers only.
1. Accessible to the selected list of users, including Locale Managers and Localizers.

Alternatively, we could keep the two options only (Private/Public) and add a checkbox: "Visible for Locale managers and Localizers".

PR #1563, opened 2020-02-11
Fix bug 1614593: Do not error if project created with empty Download prefix field

mathjazz:
The patch is deployed to stage and ready for testing.


mathjazz:
Hey @flodolo, could you please test this on stage?

flodolo:
It doesn't save, I get the error that the field is required. Is that the expected behavior?

mathjazz:
Yes. I suspect making the filed optional will simplify things you, so I should do it?

flodolo:
> Yes. I suspect making the filed optional will simplify things you, so I should do it?

I see two options:
1) It should be populated automatically, because right now it's a mess to find out.
2) It should be optional.

I'd prefer 2 at this point, with 1 as a long term fix.

mathjazz:
Updated with option #2 and deployed to stage for another round of testing.

mathjazz:
Thanks!

mathjazz:
Pinging @adngdb for a **code** review. :-)

PR #1565, opened 2020-02-12
Add timelines specification.

Pike:
This is for [bug 1614618](https://bugzilla.mozilla.org/show_bug.cgi?id=1614618)


specifications/1614618-timelines.md
```diff
@@ -0,0 +1,95 @@
...
+
+The basic model for Timelines is a Point in Time.
+
+| Key | Type |
+| -- | -- |
+| project | ForeignKey(Project) |
+| when | datetime |
+| label | string |
+| author | ForeignKey(User) |
+| created_when | datetime |
```
mathjazz: Nit: our existing models call such fields `created_at`.

specifications/1614618-timelines.md
```diff
@@ -0,0 +1,95 @@
+- Feature Name: Timelines
```
mathjazz: Note that we have another feature proposal with the same name:
https://bugzilla.mozilla.org/show_bug.cgi?id=1473760

I suggest we rename this feature to "Milestones".

specifications/1614618-timelines.md
```diff
@@ -0,0 +1,95 @@
...
+
+# New Roles
+
+| Role | Impact |
+| -- | -- |
+| Project Manager (PPM) | PPM can set multiple points in time for a project|
+
+This feature introduces a role of the upstream project manager. In contrast to
+the Localization Project Manager (LPM), this is a Project Project Manager, thus
+we're introducing the acronym PPM.
```
mathjazz: I like this section!

I think adding such role is a separate feature, though:
- I don't see the need for it in the Motivation section.
- We should let PPMs edit other project details.
- I'd even let them [verify/submit projects](https://bugzilla.mozilla.org/show_bug.cgi?id=1204659) to be enabled on Pontoon.

specifications/1614618-timelines.md
```diff
@@ -0,0 +1,95 @@
...
+| author | ForeignKey(User) |
+| created_when | datetime |
+
+PITs are deleted when projects are deleted.
+
+*TBD*: Modifications to existing PITs should be traced in a log. Is ActionLog
+the right place?
+
+*TBD*: What to do when a User is deleted?
+
```
mathjazz: I like that section. :-)

specifications/1614618-timelines.md
```diff
@@ -0,0 +1,95 @@
...
+| project | ForeignKey(Project) |
+| when | datetime |
+| label | string |
+| author | ForeignKey(User) |
+| created_when | datetime |
+
+PITs are deleted when projects are deleted.
+
+*TBD*: Modifications to existing PITs should be traced in a log. Is ActionLog
+the right place?
```
mathjazz: I think it's the right place, yes. What about additions? I think that's even more important to show in e.g. the "Timelines" tab.

specifications/1614618-timelines.md
```diff
@@ -0,0 +1,95 @@
...
+
+*TBD*: What to do when a User is deleted?
```
mathjazz: `SET_NULL`? We don't want to delete PITs.

specifications/1614618-timelines.md
```diff
@@ -0,0 +1,95 @@
...
+every two weeks, a PIT for the next cut-off for a Fenix release, and a PIT for
+the next cut-off for a Lockwise release.
+
+*TBD*: Should metadata like labels be shown to Localizers?
+
+*TBD*: Should we show the latest PIT if there's no upcoming one? Should we do
+that only for some amount of time?
+
+*TBD*: Can there be both a Deadline and a Timeline for a project?
+
```
mathjazz: I would expect "Feature Explanation" to define the UX implications of the future precisely enough for developers to be able to start coding. I think https://github.com/mozilla/pontoon/pull/1562/ succeeds in that.

It might be pretty straightforward where to put managing of PITs and how it should work, but it's still worth writing it down. OTOH, localizer implications are missing: how does the feature affect dashboards and notifications? Should we mark PITs in the Time Range chart? Etc.

specifications/1614618-timelines.md
adngdb: I don't understand this PPM / LPM thing at all. I agree that it seems good to have a section about roles though, if only for the benefit of me being confused by this content. :smile: 

I'd suggest merging the two roles-related sections into one. And I think we should also consolidate the list of roles in Pontoon, putting that list in the future template, asking each spec to explain clearly how a feature impacts each role. That means we need to take some time to clarify what the roles currently are in Pontoon, though, because I for one am totally not sure what they are.

specifications/1614618-timelines.md
adngdb: Agreed, with this description I have difficulty picturing what the feature will look like, and thus how it will work.

specifications/1614618-timelines.md
adngdb: I'm not sure we need to go all the way to data models in specifications. I would leave that to the developer in charge of implementation.

specifications/1614618-timelines.md
adngdb: I also like the questions asked here. Not sure it should be "Data models", maybe "Technical implementation questions"? Or maybe it should just be part of the "Feature Explanation"?

Pike:
I've done the rename, and addressed some of the review comments.

I guess I should've clarified the intent of the *TBD* initially, they're basically a self-review piece, and not intended to be part of the final landing. I'm generally asking for more opinions there.

Another interesting question is if I should attempt to write UX specs. U know, I'm at most half-eyed when it comes down to UX.

specs/0103-milestones.md
```diff
@@ -0,0 +1,62 @@
...
+Milestones introduce a concept of multiple points in time that have impact to
+localizers. The milestones shown to localizers are always relevant to what a
+localizer does right now.
+
+For PMs these milestones will have labels that help to recognize which
+milestone is which. That way, if schedules change, PMs can update the corresponding milestone easily.
+
+Example: In Android L10n, we might have milestones for releases of Android Components
+every two weeks, a milestone for the next cut-off for a Fenix release, and a milestone for
+the next cut-off for a Lockwise release.
```
mathjazz, line 37: question: We've split Android L10n into per-product projects. Maybe there's a more generic example we can use in a spec?

specs/0103-milestones.md
```diff
@@ -0,0 +1,62 @@
...
+
+*TBD*: Should metadata like labels be shown to Localizers?
```
mathjazz, line 39: suggestion: If not elsewhere, we could show them in the Timelines tab?
https://bugzilla.mozilla.org/show_bug.cgi?id=1473760

specs/0103-milestones.md
```diff
@@ -0,0 +1,62 @@
...
+
+*TBD*: Should we show the latest milestone if there's no upcoming one? Should we do
+that only for some amount of time?
+
+*TBD*: Can there be both a Deadline and Milestones for a project?
```
mathjazz, line 44: thought: I see Deadline as the next Milestone, but that really depends on how you envision the UX. We should have some form of informing users which of the projects is the most urgent to work on. A combination of the Priority and the Deadline column does that job pretty well (apart from limitations listed in the Motivation section).

mathjazz (commented):
Thanks for the update, it's a big step forward! I understand the feature much better now.

Not sure you'll like it, but I think the only missing piece now is the UX part. I think you should give it a try. :)

specs/0103-milestones.md
```diff
@@ -0,0 +1,62 @@
...
+
+| Key | Type |
+| -- | -- |
+| project | ForeignKey(Project, CASCADE) |
+| when | datetime |
+| label | string |
+| author | ForeignKey(User, SET_NULL) |
+| created_at | datetime |
+
+Milestones are deleted when projects are deleted.
```
Delphine, line 60: I'm not sure I understand this part.
It's just saying that if we delete a project we delete the corresponding milestones?

specs/0103-milestones.md
```diff
@@ -0,0 +1,62 @@
...
+
+Example: In Android L10n, we might have milestones for releases of Android Components
+every two weeks, a milestone for the next cut-off for a Fenix release, and a milestone for
+the next cut-off for a Lockwise release.
+
+*TBD*: Should metadata like labels be shown to Localizers?
+
+*TBD*: Should we show the latest milestone if there's no upcoming one? Should we do
+that only for some amount of time?
+
```
Delphine, line 43: I can't really think of a reason why we would want to keep the previous milestone if there is no upcoming one.
Thoughts?
Side-note in the case where there is an upcoming milestone: I do think the milestones should disappear and switch to the upcoming one each time

specs/0103-milestones.md
mathjazz: Yup, that's the idea behind labels.

specs/0103-milestones.md
mathjazz: Correct. :)

specs/0103-milestones.md
Delphine: cool! Ok so perhaps a more generic example here for labels could be something along the lines of:
"Example: for mobile products, Firefox for Android and Focus for Android follow the same l10n milestones - but Firefox for iOS and Focus for iOS follow different ones (as does Firefox). Labels would help identify these easily, such as "android v99", "iOS v99". and "desktop v99"."

specs/0103-milestones.md
Delphine: Do we have actual deadlines for some projects these days (as in, a project that is static and does not have any other releases planned)?
I feel like we only have ongoing projects in Pontoon with Milestones at this point.
Might be worth just renaming every instance of "deadline" as Milestone, unless there is really a reason not to

specs/0103-milestones.md
Delphine: Ah ok, perhaps a good place to point out what we want to do with the Milestones when there is no upcominug ones - as well as when there is an upcoming one (like discussed in a comment above)

specs/0103-milestones.md
Delphine: Another thought around this: adding something additional around notifications in Pontoon, about the milestones coming up. That would be more for PMs (and locale managers?)

mathjazz: Yeah, we should hash that out. Currently, we send notifications when projects are incomplete for the locale 7 and 2 days before the deadline.

Pike:
On a general note, I think it's perfectly fine to prioritize Mozilla's current needs and priorities.

For Pontoon, though, it's also OK to think about what Mozilla needed at some point, and consider that other deployments might have the same needs, and if it's not the way of what Mozilla needs, provide for those.

That said, maybe something that's on an agile sprint-oriented timeline is actually rare in the world of localization.

peiying2:
Coming late to the party, :). Another use for deadline these days in my case has been new product launch for `select markets`. The info needs to be shared to the communities but I couldn't use Pontoon. Instead, I communicated to the target communities through email. If we are able to set deadlines for the targeted locales and only to those locales that the deadline is shown. For the rest, the project is available for localization just like any others.

flodolo:
> The info needs to be shared to the communities but I couldn't use Pontoon.

Why? You can send notifications to specific locales in a project. They're not the best, since they only live in Pontoon, but that's one way to do it.

In any case, how would milestones help in your scenario? Different milestones for different locales?

amirmozla:
> My personal take is that we're at a point where we're using deadlines less and less, because it's clear that people localize when they want and can. On top of that, some projects are not that great at sticking to the deadlines they set. So, my instinct would be to simplify things, not to add complexity.
> 
> If the word deadline is scary, we should change it (personally, I don't think it is). Instead of creating multiple milestones, we could just add an optional description for the deadline.
> 
> What we need is to make sure a deadline is automatically hidden once it's passed, e.g. a couple of days after. That's currently the main problem I see in Pontoon: you get a notification for new strings in a project, go in Pontoon and see that the deadline is set in the past, which is frustrating and confusing for new contributors.

I agree with optional description for the deadline, but it would be nice to have this whole deadline/due-date process automated somehow.

mathjazz:
Thanks everyone for your comments!

I'm closing this PR as explained in #2210. Let's continue the effort with #2433 and #2434 and file more issues to improve the behavior of the deadlines.

PR #1570, opened 2020-02-20
[RFC] Bug 1616869 - Documentation overhaul specification

adngdb:
This is a request for comments at the moment, I intend to evolve this into an actual specification based on feedback I will receive here and elsewhere.

Let's not discuss the *form* here, but only the *content* (form can be discussed in other specification PRs like #1562).


flodolo:
I think a first step should be avoiding using role names that have an established connotation in Pontoon (contributor, administrator, manager).

Contributor -> Occasional developer?
Administrator -> System Administrator?
Manager -> Administrator

```
- **Contributor** — New developer looking to contribute to Pontoon's code base. Wants accessible setup and contributing documentation.
- **Administrator** — System administrator looking to host their own instance of Pontoon. Wants platform-specific setup documentation.
- **Localizer** — User of Pontoon looking to learn how to efficiently translate projects. Wants detailed features documentation.
- **Manager** — User of Pontoon looking to learn how to manage projects and locales. Wants detailed administration tasks documentation.
- **Core developer** — Core developer of Pontoon. Wants processes and technical documentation.
```

I'm also not clear what the expectation here is. I'd expect the specification to only have the agreed plan (i.e. the outcome of the discussion)?

mathjazz:
First of all, I really like that you are working on this issue.

I don't see the need for distinction between two groups of developers (new and core). As a core developer, I still look into the Local setup and Contributing guidelines from time to time. I can also see (more senior) new developers jumping to technical documentation.

I'd call Managers "Project Managers", because that's how we call them. :-) I've seen the same terminology in other localization platforms, in addition to "localization managers".

mathjazz:
As a small team I don't think we can afford spending time developing our own documentation solution, not even maintaining our own wiki deployment.

I'd go with a hosted documentation solution like ReadTheDocs, because it's the cheapest option. It's also not that much harder to click twice to edit than clicking once to edit the wiki page. :-)

adngdb:
> I'm also not clear what the expectation here is. I'd expect the specification to only have the agreed plan (i.e. the outcome of the discussion)?

Ultimately yes, that's what's going to happen, but I wanted to try to use this like Python does, as a "Request for comments". This PR should be a living thing, and the document will evolve with comments, until we reach a consensus.

adngdb:
> I'd go with a hosted documentation solution like ReadTheDocs, because it's the cheapest option. It's also not that much harder to click twice to edit than clicking once to edit the wiki page. :-)

I'm not sure I understand this. The problem with RTD is that documentation is in a git repo, on GitHub, and thus requires some technical skills to be able to edit it (a wiki is a much simpler tool to use for the average person). It's not really about the number of clicks, unless I misunderstood you? :curious:

mathjazz:
> I'm not sure I understand this. The problem with RTD is that documentation is in a git repo, on GitHub, and thus requires some technical skills to be able to edit it (a wiki is a much simpler tool to use for the average person). It's not really about the number of clicks, unless I misunderstood you? :curious:

There's "Edit on GitHub" link on top of each page on RTD, and once you get to GitHub, you just click the Edit (pencil) again which takes you to the editor:
https://github.com/mozilla/pontoon/edit/master/docs/index.rst

It's harder than wiki, it's not WYSIWYG, and someone needs to merge the PR, but I think it's not a bad editing experience. In fact, a confirmation step before updating technical documentation is IMO a plus.

specs/0102-documentation-overhaul.md
```diff
@@ -0,0 +1,106 @@
...
+
+**Pros**
+
+- Each audience has a more focused documentation, with less risks of confusion.
+- Allows us to translate only the localizer-facing documentation, and not the technical one.
+
+**Cons**
+
+- Two distinct documentations to maintain, which is more difficult.
+- There might be confusion about what to put where.
```
gueroJeff: We'll need to define criteria, for sure.

specs/0102-documentation-overhaul.md
```diff
@@ -0,0 +1,106 @@
...
+## Where to store documentation
+
+### ReadTheDocs
+
+We build the documentation automatically and push it to a readthedocs.org page on each update to the master branch.
+
+**Pros**
+
+- Hosted with Pontoon's code base, allowing to enforce documentation updates along with code changes.
+- Localization is technically possible, but I don't know how difficult it would be.
```
gueroJeff: Actually performing localization or are we talking about multilingual publishing, or both?

specs/0102-documentation-overhaul.md
```diff
@@ -0,0 +1,106 @@
...
+- Easy to deploy and maintain, as ReadTheDocs is a mature solution and we have experience with it.
+
+**Cons**
+
+- Contributing to the documentation is difficult as it requires making a pull request on GitHub.
```
gueroJeff: That's the current barrier to entry for localizer-documentation too.

specs/0102-documentation-overhaul.md
```diff
@@ -0,0 +1,106 @@
...
+- Requires building our own documentation solution, with localization support.
+
+### Wiki
+
+**Pros**
+
+- Very easy to edit, by anyone interested in doing so.
+- Some wikis support content localization out of the box.
+
+**Cons**
```
gueroJeff: Moderation may add overhead too.

specs/0102-documentation-overhaul.md
adngdb: This is about the technical side, enabling localization. I didn't consider the translation work in this document.

Pike:
Leaving some personal opinions here:

I'd prefer for all docs to be in the pontoon repo.

I'm not sure how much impact localizing docs should have on our decision. I think it falls into the same bucket of "yes and" as localizing Pontoon itself.

We shouldn't pay too much attention on hosting. Anything that can be built can be put into an S3 bucket and served from there in the end.

There is the idea that PMs can write markdown and not restructured text. Support for markdown in Sphinx exists, but it doesn't cover all features, IIRC, https://www.sphinx-doc.org/en/master/usage/markdown.html.

adngdb:
I'm not sure about Sphinx's support of Markdown, but in theory, Markdown supports HTML, so if a feature is not supported it should be possible to just build it with HTML markup, right? I'm think about tables for example, usually poorly supported in most markdown tools, but easy to set up with some good old HTML tags.

Pike:
We should probably try on a dummy doc. Things I'd like to see tested are various cross references, indices, and search.

mathjazz:
Not directly related, but I don't have a better place to dump it.

Nice docs on writing docs:
https://documentation.divio.com/

specs/0104-documentation-overhaul.md
```diff
@@ -0,0 +1,54 @@
...
+
+That's easier to set up for external deployments than using a 3rd party documentation service like ReadTheDocs. It also maps each Pontoon deployment to its corresponding documentation more explicitly, making sure deployment version and docs version match.
+
+## What format to use
+
+Documentation is writen using the Markdown syntax.
+
+## Maintenance
+
+TBD
```
mathjazz, line 54: I wonder if there's something we can do to "enforce documentation updates along with code updates".

Something automated, something like documentation coverage.

specs/0104-documentation-overhaul.md
```diff
@@ -0,0 +1,54 @@
...
+
+It is also unclear who is responsible for maintaining these documentations. That can lead to big delays between a change in a feature and its associated documentation being updated.
+
+# Feature explanation
+
+## Target audiences
+
+- **Localizer** — User of Pontoon wanting to learn how to efficiently translate and review projects. Looks for detailed feature documentation.
+- **Administrator** — User of Pontoon wanting to learn how to manage projects and teams. Looks for detailed administration tasks documentation.
+- **Project Owner** — Anyone wanting to make their project translated into several languages. Looks for internationalization and deployment documentation.
```
Pike, line 27: I'm not sure we should include this target audience. If I understand this correctly, I'd assume this to be very specific to each org deploying Pontoon.

Pike: Also, is this deploying Pontoon or projects localized with Pontoon?

mathjazz: This covers deploying Pontoon as well as prerequisites on the side of the project to be localized (i.e. information about supported l10n frameworks, file formats and VCS systems, required access, file and folder structure):
* https://mozilla-pontoon.readthedocs.io/en/latest/user/localizing-your-projects.html (without Adding a new project section, which is *Administrator*)
* https://mozilla-pontoon.readthedocs.io/en/latest/admin/deployment.html
* https://mozilla-pontoon.readthedocs.io/en/latest/admin/maintenance.html

flodolo (commented):
I think this is still a bit too vague.

* We want to use Markdown, great. How do we convert it to static HTML? How do we deal with images?
* Will the solution support search? That's a blocker.

I'm OK with having all docs in one place, but not in exposing them together. Localizers are interested in user docs, not the rest, so they shouldn't have to navigate through the entire documentation, or be exposed to it.

Side note: I don't think there's anything preventing us from deploying parts of the documentation to GH pages, like we're currently doing for user docs.

specs/0104-documentation-overhaul.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Where to host
+
+Documentation is hosted in a dedicated section within Pontoon app (`/docs`) and built on each deployment.
+
+That's easier to set up for external deployments than using a 3rd party documentation service like ReadTheDocs. It also maps each Pontoon deployment to its corresponding documentation more explicitly, making sure deployment version and docs version match.
+
+## What format to use
+
+Documentation is writen using the Markdown syntax.
```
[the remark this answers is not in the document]
jotes: Do we plan to use a static site generator?

specs/0104-documentation-overhaul.md
jotes: What do we exactly expect from such software? e.g. print what parts of the documentation should be updated in a PR?

Maybe we should involve more people in the review process, e.g. ask a localizer/project manager to check the documentation about a feature/bugfix (?)

specs/0104-documentation-overhaul.md
```diff
@@ -0,0 +1,54 @@
...
+
+All documentation is put in one central place. It is organized by audience, with clearly defined sections, and an accessible front page allowing users to quickly find the documents they need.
+
+A single place for all documentation means less questions about where to put what, and is easier to maintain.
+
+## Where to store
+
+Documentation is stored in the `docs` folder within the Pontoon code repository.
+
+Having documentation in the same place as the code base allows us to enforce documentation updates along with code updates.
```
jotes, line 40: I see a problem there related to the fact that every link is going to point to the latest version? e.g. when a translator discusses a document that changes over time and there's no way to reference the previous version.

srl295:
Should the buglink be #2214 ?

mathjazz:
This patch has been stalled for a while.

Let's reopen when work continues.

PR #1573, opened 2020-02-27
Fix Bug 1584180: run quality checks using bulk_run_checks and simplify stats.

VishalCR7:
Hey! There are two changes from #1553 
* Run checks "elegantly" using the `bulk_run_checks` method the downside of which I thought was time taken for stats calculation
* Use [bulk_update](https://github.com/mozilla/pontoon/blob/master/pontoon/batch/views.py#L33-L57) logic for stats calculation (We can also try to modify the original method a bit to make it suitable for pretranslation use case i.e. single project & multiple locales. 

I tested both versions on [pontoon-test](https://github.com/mathjazz/pontoon-test/) for a single locale. Here are the time taken for different parts:

**The current PR** 
- Collecting Translations Finished: 50.1863968372345
- Checks Finished: 0.5103518962860107
- Mark as changed Finished: 0.08880019187927246
- Update Stats Finished: 0.5324735641479492

#1553 
- Collecting Translations Finished: 43.390445709228516
- Checks Finished: 0.0789637565612793
- Mark Translation Active Finished: 0.023960351943969727
- Stats and latest activity Finished: 0.04465889930725098

As @mathjazz pointed out earlier, the most time-consuming part is collecting pretranslation, We can simplify the code for stats calculation (and may be latest_translation)

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -17,6 +17,7 @@ from django.db.models.functions import Length, Substr, Cast
 from partial_index import PartialIndex, PQ
 from six.moves import reduce
 from six.moves.urllib.parse import urlencode, urlparse
+from bulk_update.helper import bulk_update
 
 from django.conf import settings
 from django.contrib.auth.models import User, Group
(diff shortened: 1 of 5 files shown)


mathjazz:
Excellent job!

Just tested this locally with pontoon-test and 2 locales and the stats calculation time is insignificant compared to the entire execution time. The amount of code compared to #1553 OTOH is significantly smaller, so I'd close #1553.

I still need to have a closer look at the code.

pontoon/base/models.py
```diff
@@ -3260,6 +3261,40 @@ def stats(self, project, paths, locale):
 
         return translated_resources.aggregated_stats()
 
+    def update_stats(self, locales):
+        """Update stats on a list of TranslatedResource.
+        """
+        translated_resources = list(self)
```
mathjazz: You should only convert `translated_resources` to a list within `bulk_update`, otherwise it will have negative effect on performance. That will also allow you to prefetch related objects needed for faster calculations.

We should remove the `locales` parameter, because you can get it from the `TranslatedResources`, same as you already do for projects. I'd change the way you retrieve them though, and also use the same logic for collecting `ProjectLocales`.

```python
def update_stats(self):
    self = self.prefetch_related('resource__project', 'locale')

    locales = Locale.objects.filter(
        translatedresources__in=self,
    ).distinct()

    projects = Project.objects.filter(
        resources__translatedresources__in=self,
    ).distinct()

    projectlocales = ProjectLocale.objects.filter(
        project__resources__translatedresources__in=self,
        locale__translatedresources__in=self,
    ).distinct()

    for translated_resource in self:
        translated_resource.calculate_stats(save=False)

    # bulk update TR

    # aggregate stats for P, L and PL
```

pontoon/pretranslation/tasks.py
```diff
@@ -107,44 +113,49 @@ def pretranslate(self, project_pk, locales=None, entities=None):
...
     if len(translations) == 0:
         return
 
     translations = Translation.objects.bulk_create(translations)
 
-    # Update latest activity and unreviewed count for the project.
-    project.latest_translation = translations[-1]
-    project.unreviewed_strings += len(translations)
-    project.save(update_fields=["latest_translation", "unreviewed_strings"])
+    translation_pks = {translation.pk for translation in translations}
```
mathjazz: A comment above this line would be nice, e.g. `# Run checks on all translations`.

PR #1574, opened 2020-02-27
Fix bug 1616918 - Enforce newline at end of Fluent strings when compa…

adngdb:
…ring.

This solves a detection problem with simple Fluent strings, where the newline is missing from some Fluent string in history. This enforces trailing newlines, making the selector find the correct similar translation.

### DIFF
--- a/frontend/src/core/editor/selectors.js
+++ b/frontend/src/core/editor/selectors.js
@@ -62,27 +62,25 @@ export function _existingTranslation(
         }
         // If translation is a string, from the generic editor.
         else {
-            // Except it might actually be a simple Fluent message from the SimpleEditor.
+            // Except it might actually be a Fluent message from the Simple or Source
+            // editors.
             if (entity.format === 'ftl') {
-                // This case happens when the format is Fluent and the string is simple.
-                // We thus store a simplified version of the Fluent message in memory,
-                // but the history has actual Fluent-formatted strings (with ID).
-                // Thus, we need to reconstruct the Fluent message in order to be able
-                // to compare it with items in history.
-                const reconstructed = fluent.serializer.serializeEntry(
-                    fluent.getReconstructedMessage(entity.original, translation)
-                );
+                // For Fluent files, the translation can be stored as a simple string
+                // when the Source editor or the Simple editor are on. Because of that,
+                // we want to turn the string into a Fluent message, as that's simpler
+                // to handle and less prone to errors. We do the same for each history
+                // entry.
+                let ftlMessage = fluent.parser.parseEntry(translation);
+                if (ftlMessage.type === 'Junk') {
+                    // If the message was junk, it means we are likely in the Simple
+                    // editor, and we thus want to reconstruct the Fluent message.
+                    // Note that if the user is actually in the Source editor, and
+                    // entered an invalid value (which creates this junk entry),
+                    // it doesn't matter as there shouldn't be anything matching anyway.
+                    ftlMessage = fluent.getReconstructedMessage(entity.original, translation);
+                }
                 existingTranslation = history.translations.find(
-                    t => {
-                        // Some Fluent strings are missing the trailing newline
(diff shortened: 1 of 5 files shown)


frontend/src/core/editor/selectors.js
```diff
@@ -73,8 +73,17 @@ export function _existingTranslation(
...
+                        // Some Fluent strings are missing the trailing newline
+                        // character while stored in the database. In order for this
+                        // comparison to work, we need to make sure it is there.
+                        let str = t.string;
+                        if (!str.endsWith('\n')) {
+                            str += '\n';
+                        }
+                        return str === reconstructed;
+                    }
+                );
```
mathjazz: Why not instead parse & serialize as we do in line 52?

I'd also file a bug to investigate why the strings are stored inconsistently and reference it here.

frontend/src/modules/fluenteditor/components/Editor.js
```diff
@@ -146,7 +146,11 @@ export class EditorBase extends React.Component<EditorProps, State> {
             }
         }
 
-        props.setInitialTranslation(translationContent);
+        // Update the initial translation content only when the entity changed, not
+        // when new content is loaded from an external source.
+        if (typeof translation === 'undefined') {
+            props.setInitialTranslation(translationContent);
+        }
```
mathjazz: The code change makes sense, but I'm unable to figure out what problem was the old code causing in the UI.

frontend/src/core/editor/selectors.js
adngdb: Because it's a lot more work, for just this simple fix.

frontend/src/modules/fluenteditor/components/Editor.js
adngdb: Right, sorry, that's unrelated. I thought it was a cause of issue but it wasn't, and I didn't revert it. This was causing a failure in the unsaved changes code: if you load a translation from for example the History content, the initial translation would be updated to that translation, and thus when leaving the entity, the unsaved changes warning would not be triggered. Do you want me to move this to a different PR?

mathjazz: Nope, feel free to leave this in. I'm unable to reproduce the problem, though:
https://pontoon.mozilla.org/sl/firefox/all-resources/?string=203763

frontend/src/core/editor/selectors.js
mathjazz: OK. I'd still reference the bug you filed here (and/or vice-versa).

mathjazz:
(Unless you want to also fix the Source Editor, then I'll review again) :-)

adngdb:
Yeah, I wanted to fix it because it was actually caused by this. Now it should work in all fluent cases. I've also added a bit more code to fix another issue with unsaved changes (I do it here because I find them as the initial translation is also used in the same translation selector, sorry for the review mess).

Note that I can reproduce the unsaved changes issue I talked about locally, not sure why it doesn't happen in the case you linked to.

frontend/src/core/editor/selectors.js
```diff
@@ -62,19 +62,26 @@ export function _existingTranslation(
...
-                // to compare it with items in history.
-                const reconstructed = fluent.serializer.serializeEntry(
-                    fluent.getReconstructedMessage(entity.original, translation)
-                );
+                // For Fluent files, the translation can be stored as a simple string
+                // when the Source editor or the Simple editor are on. Because of that,
+                // we want to turn the string into a Fluent message, as that's simpler
+                // to handle and less prone to errors. We do the same for each history
+                // entry.
+                let ftlMessage = fluent.parser.parseEntry(translation);
```
mathjazz: Nit: I don't like that we're using two inconsistent names for the seamingly same concept: `ftlMessage` and `fluentTranslation`.

PR #1575, opened 2020-02-28
[WIP] Fix Bug 1611510 - Implement source string comments 

abowler2:
Initial work to implement source string comments.

____
Questions/issues:
1) I have it set up with two buttons (`Request context` and `Report issue`) but they both call the same function on click.  Should this stay as 2 buttons or become one?
2) I'm having trouble serializing the `contact` value for the project manager. I've been trying to send it up along with the entities from within `map_entities` in `models.py` but keep running into the error:
```
TypeError: Object of type User is not JSON serializable
```


mathjazz:
This was just a temporary code share, closing for now.

PR #1578, opened 2020-03-02
Source comments

abowler2:
@mathjazz Here is what I have so far in my attempt to send the `contact` up with the `get-entities` request

I'm clearly missing a step (or more) as I'm getting a `'dict' object is not callable` error.

Your thoughts on this are appreciated.


abowler2:
Closing this as it was a temporary code share.

PR #1579, opened 2020-03-03
Fix bug 1619650: Add Issue tracking to autocomplete list of External Resources

mathjazz:
Deployed to stage for testing:
https://mozilla-pontoon-staging.herokuapp.com/admin/projects/firefox/

### DIFF
--- a/pontoon/administration/templates/admin_project.html
+++ b/pontoon/administration/templates/admin_project.html
@@ -258,7 +258,6 @@
     <option value="Development build">
     <option value="Production build">
     <option value="Screenshots">
-    <option value="Language pack">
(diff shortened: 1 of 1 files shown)


pontoon/administration/templates/admin_project.html
```diff
@@ -259,6 +259,7 @@ <h3>Project info <span class="small stress">(optional)</span></h3>
     <option value="Production build">
     <option value="Screenshots">
     <option value="Language pack">
```
flodolo: Can we remove "Language Pack" at this point? As far as I can tell, that's not supported anywhere at this point.

PR #1580, opened 2020-03-04
first steps for running in k8s environment

hbinkle:
* adjusted base.py to enabled using some more env vars needed for running in k8s environment

* example files and instructions to get pontoon running on k8s.
--- it's not yet a stable solution ---


adngdb:
Hey @hbinkle, thanks a lot, this looks much better, and the documentation is much appreciated! I wonder, why did you duplicate the Dockerfile? Can you not reference the official one? I'm sure we will eventually forget to update the duplicated one, and it might also cause confusion for people working on Pontoon.

hbinkle:
Hello @adngdb,
I'm sorry in some steps inbetween I needed to adjust the docker file.
But finally I recongnized that these adjustment are no longer necessarry.
I simply reverted the adjustments and forget to remove the file.
It's done now.

PR #1583, opened 2020-03-10
Remove redundand test decorators and params

mathjazz:
@julen r?

### DIFF
--- a/pontoon/base/tests/managers/test_entity.py
+++ b/pontoon/base/tests/managers/test_entity.py
@@ -938,7 +938,7 @@ def test_mgr_entity_reset_active_translations(resource_a, locale_a):
     [
         ("There are 7 words in this string", 7),
         ("String 123 =+& string hh-gg object.string", 5),
-    ]
+    ],
 )
 def test_get_word_count(input, expected_count):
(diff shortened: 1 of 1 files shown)


julen:
Looks good to me.

As a tip, one thing I love about pytest is being able to [parametrize test functions](https://docs.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions) to quickly add multiple expected input/output parameters; example at https://github.com/evernote/zing/blob/952c1cac4782e1252e45c24ab6d6cac23c51da9f/tests/core/utils/json.py#L28-L44

So this particular test function could be written as (didn't test it):
```py
@pytest.mark.parametrize(
    "input, expected_count",
    [
        ("There are 7 words in this string", 7),
        ("String 123 =+& string hh-gg object.string", 5),
    ]
)
def test_get_word_count(input, expected_count):
    assert get_word_count(input) == expected_count
```

And with that it becomes pretty easy to add more test cases, and to focus thinking on the edge cases rather than the body of the test function itself.

mathjazz:
Thanks for the tip! Updated.

PR #1584, opened 2020-03-10
Prototype of a new interface for translations in history.

adngdb:
This is a design research proposal, based on work by @abowler2 and @mathjazz, forked from #1561. The intent is to provoke conversation around the current UX of the history panel, by challenging some design choices that were made long ago. This is not meant to be merged as is. It is also not high priority.

This proposal changes entries in the History panel, making review actions more explicit, always showing the comments widget, and simplifying the translation status icons.

With this change, the translation status is separated from review actions. Users have access to only the review actions they are allowed to perform. Main actions (approve, reject, delete) are done with a colored button containing a single, clear icon. Secondary actions (unapprove, unreject) are done with a transparent button containing explicit text. 

Hovering over anything related to a translation will highlight that translation's content, to make it clear what actions and comments are about. 

_What it looks like for a superuser:_
Screenshot_2020-03-10 French (fr) · Pontoon Test HG ssh

_What it looks like for a logged out user:_
Screenshot_2020-03-10 French (fr) · Pontoon Test HG ssh(1)


mathjazz:
High-level observations:
1. At this stage, this is better suited for something like a Discourse discussion. I think it's too early for a specification or even a bug.
1. We should have a clear problem statement written down, covering what doesn't work and what works in the current implementation.
1. We should try to analyse DB and New Relic data to support the proposed changes.
1. Last but not least, we should take into account the urgency of these changes vis-à-vis our Roadmap.

adngdb:
I'm closing this as it's unlikely we'll pursue it. But I was happy to take some time thinking about that UX! :relaxed:

PR #1585, opened 2020-03-11
Fix bug 1621720: Send notifications on comment submissions

mathjazz:
This patch sends notifications when comments get submitted.

On translation comment (bound to translation), we notify:
- translation author
- authors of any other translation comments in the thread

On team comment (bound to entity-locale pair), we notify:
- any locale managers
- any locale translators
- any project-locale translators
- authors of any other team comments in the thread
- any translation authors
- authors of any translation comments

### DIFF
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -25,6 +25,7 @@ from django.views.decorators.csrf import csrf_exempt
 from django.views.decorators.http import condition, require_POST
 from django.views.generic.edit import FormView
 
+from notifications.signals import notify
 from six.moves.urllib.parse import urlparse
 
 from pontoon.actionlog.utils import log_action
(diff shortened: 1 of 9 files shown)


flodolo:
> On translation comment (bound to translation), we notify:
>     * translation author
>     * authors of any other translation comments in the thread

We should add the reviewer (approver) here, if the string is approved.

> On team comment (bound to entity-locale pair), we notify:

First thought: are we notifying too many people for this? Do we plan to add the ability to tailor the notifications via profile settings at some point, like you do on Bugzilla or Phabricator?

>     * any locale translators
>     * any project-locale translators

What does this mean in practice? 

If a locale has a general group of translators (defined at locale level), and a per-project custom configs, we should only notify users with translators rights on that specific project, not include the general one.

mathjazz:
> We should add the reviewer (approver) here, if the string is approved.

Good point!

> First thought: are we notifying too many people for this? Do we plan to add the ability to tailor the notifications via profile settings at some point, like you do on Bugzilla or Phabricator?

I don't know if that's too many. The thing is that unlike suggestions, you don't see comments on dashboards, so they are impossible to find without notifications. We're also considering adding filters for strings with comments and including user comments in search.

Yes, we plan the ability to [opt-out](https://bugzilla.mozilla.org/show_bug.cgi?id=1364069) (and obviously back in) of notifications.

> If a locale has a general group of translators (defined at locale level), and a per-project custom configs, we should only notify users with translators rights on that specific project, not include the general one.

Good point again!

pontoon/base/views.py
```diff
@@ -456,6 +457,82 @@ def get_team_comments(request):
     return JsonResponse(payload, safe=False)
 
 
+def _send_add_comment_notifications(user, comment, entity, locale, translation):
+    # On translation comment, notify:
+    #   - authors of other translation comments in the thread
+    #   - translation author
+    #   - translation reviewers
+    if translation:
+        recipients = list(translation.comments.values_list("author__pk", flat=True))
```
adngdb: Should `recipients` be a `set` instead of a list? There's no point sending a notification to the same user several times.

pontoon/base/views.py
```diff
@@ -456,6 +457,82 @@ def get_team_comments(request):
...
+        project_locale = ProjectLocale.objects.get(
+            project=entity.resource.project, locale=locale,
+        )
+        translations = Translation.objects.filter(entity=entity, locale=locale)
+
+        # Only notify translators of the project if defined
+        translators = project_locale.translators_group.user_set.values_list(
+            "pk", flat=True
+        )
+        if len(translators) == 0:
```
adngdb: `if not translators:` is more idiomatic I think.

pontoon/base/views.py
```diff
@@ -456,6 +457,82 @@ def get_team_comments(request):
...
+            )
+        )
+
+        recipients.append(translations.values_list("user__pk", flat=True))
+        recipients.append(translations.values_list("approved_user__pk", flat=True))
+        recipients.append(translations.values_list("unapproved_user__pk", flat=True))
+        recipients.append(translations.values_list("rejected_user__pk", flat=True))
+        recipients.append(translations.values_list("unrejected_user__pk", flat=True))
+
+        recipients = [item for sublist in recipients for item in sublist]
```
adngdb: What? I don't understand this. Isn't it just copying the list? Or is it flattening it? If so, you can probably avoid that by using `extend` instead of `append`.

pontoon/base/views.py
```diff
@@ -456,6 +457,82 @@ def get_team_comments(request):
...
+
+    for recipient in (
+        User.objects.filter(pk__in=recipients).exclude(pk=user.pk).distinct()
```
adngdb: If you make `recipients` a `set`, you can remove the call to `distinct` here. It might also be worth removing the current user's pk from the `recipients` set directly, instead of filtering them out in SQL.

adngdb:
I can't comment with a new user. 

## Steps to reproduce:
1. Create a new user through the django admin interface
2. Log in with that user
3. Go to the translate page of any project
4. Click the "COMMENT" link of any translation.

## Result
Nothing appears, including the comment form.

## Expected
I can see the comment form and use to add a comment.

Note that it works fine with my superuser account.

mathjazz:
Hm, I'm unable to reproduce this. Do you see any errors in the JS or Django console?

PR #1587, opened 2020-03-15
[WIP] Fix Bug 1611510 - Implement Source comments

abowler2:
Initial work to implement Source Comments
___
There are a few areas around the mentions feature that I need feedback and help on:

1) The ability to center the text (and placeholder) via `align-items` no longer works with the mentions component. I had to use padding to achieve this.
2) When moving to line 2+ there is unneeded padding added to the bottom of the textarea
3) The padding that is added to the bottom on line 2+ means that the color added to the text no longer lines up between `mentions__input` and the `mentions__highlighter` (which is what adds color to the text).
4) The mentions text is transferred to the DB in markdown format and thus displays with markdown syntax. This may allow us to have highlighted email addresses when the comments render if there is a way to utilize the markdown, otherwise I will need help on how to remove that before it is saved and/or rendered.
5) I have added a function that retrieves the needed data for `users` that is triggered when the `@` is triggered, but I don't think I set this up correctly as the data does not get sent back to the `AddComment` component to be used. I can see that the data is being received and is in the correct format, but can't access it within `AddComment`

Also, I set it up to request the full user list from the comments module as that is where it will be used (at least initially), but I named it as `CommentsState` in case future items are needed within comments. Happy to change this out if not correct.

Pending major changes on this review, I will have additional questions/issues to bring up regarding auto-adding the project manager to the comment via the button in `metadata`.


mathjazz:
Hey @abowler2, thanks for the patch. Nice work so far!

1. Could you please remove the `yarn.lock` file in the root of the repo? It's there by mistake and it adds some 7,000+ lines to the PR.

2. There are also 2 JS errors:
```
./src/core/comments/components/AddComment.js
  Line 29:9:  'getUsers' is assigned a value but never used        no-unused-vars
  Line 30:9:  'projectManager' is assigned a value but never used  no-unused-vars
```

3. Once that's fixed, I can review the patch. I've also seen some other issues, e.g.:
- There's no scrollbar anymore in the `textarea` when it's needed.
- The user list is moving as you type. It should remain in the fixed position, same as on GH.
- Why aren't we also showing full names, when they exist (in addition to emails)?
- If you add mention a user in line 2 or more, the mention appears twice (once white and once green).

abowler2:
>     1. Could you please remove the `yarn.lock` file in the root of the repo? It's there by mistake and it adds some 7,000+ lines to the PR.

Done (sorry about that 😊 )

> 
>     2. There are also 2 JS errors:
> 
`projectManager` was unused because I haven't completed the ability to add the contact to the textarea when using the `Report Issue/Request Context` button. - for now I've just console logged the value.

For `getUsers` - that is in relation to # 5 of my original [comment](https://github.com/mozilla/pontoon/pull/1587#issue-388495913). For now I added it in for the `data` attribute and commented out the manual data structure used to test the mentions. 

>     * Why aren't we also showing full names, when they exist (in addition to emails)?

I haven't fully implemented the data as I was working to get the functionality first. With that said, I still have not quite worked out how/if I can have 2 options given the data structure that is required for `mentions` to work.
> 
> 
>     * There's no scrollbar anymore in the `textarea` when it's needed.
> 
>     * The user list is moving as you type. It should remain in the fixed position, same as on GH.
> 
>     * If you add mention a user in line 2 or more, the mention appears twice (once white and once green).

These are all related (see #s 1-4 from the original [comment](https://github.com/mozilla/pontoon/pull/1587#issue-388495913)) and are due to the way the styling comes in from `react-mentions`. Many items end up as `inline` and I have not been able to override them (see image below regarding the `overflow`).  In addition, I can't work out why the spacing changes as you go to line 2 and beyond. 

Screen Shot 2020-03-18 at 9 05 00 PM

If you could take a look to see if you have suggestions on how to overcome these issues that would be great.  It may just be that this package isn't going to be compatible after all 🤷‍♀

mathjazz:
Thanks for the update and clarifications!

Note that the mentions popup doesn't appear at the moment - I had to uncomment the content of the `data` attribute of the `Mention` element. I've also noticed that it's helpful to set `allowSuggestionsAboveCursor={ true }` on the `<MentionsInput>`.

That is indeed a lot of issues that this package brings. Maybe [this article](https://medium.com/upshotly/implementing-using-react-mentions-from-react-to-database-and-vice-versa-657c936cfc42) can help? There's a "Changing the UGLY UI" section towards the bottom and I see the components also take the `style` attribute.

The article also mentions [draft-js-mention-plugin](https://www.npmjs.com/package/draft-js-mention-plugin), which you might want to check out in case the article won't help.

I also came across an article about [implementing a custom mentions component](https://blog.campvanilla.com/reactjs-input-trigger-github-twitter-mentions-8ad1d878110d), but I'd try the existing solutions first.

abowler2:
@mathjazz Thank you for the references. Those were some of the ones I had been working with and, in fact, the Upshot suggestions for styling were what got me as far as I had gotten 😊 . Although, based on what that article said about using `draft-js-mention-plugin` I didn't look too closely at [draft.js](https://github.com/facebook/draft-js) until yesterday. 

Now that I have, I think `draft.js` would be a better way to go as I think it will handle the mentions functionality much better with a lot fewer side effects in addition to addressing much of the additional functionality that we may want to add (such as bold text, italics, etc.)

So far it has not caused any issues with the UI, which is much appreciated. Where I'm a little uncertain is how to save the data to the DB and then render it again.  The best I can tell there are a couple of options. 1) Use the built-in option of converting it to and from JSON, which I believe would require changing the `comment` file to properly render that. 2) Save it as HTML, which would require an additional dependency and I think also requires changes with rendering given the way the comments are currently handled. 

For reference: 
[Post on saving the content with the built in functionality](https://codepulse.blog/how-to-store-draft-js-content/)
[Post on converting content to HTML](https://codepulse.blog/how-to-display-draft-js-content-as-html/) 

So, before I go too far down this road.  What are your thoughts on this? Are you  OK moving forward with draft.js? If so, which route do you recommend for saving the content?

TIA 😄

mathjazz:
The fact that you used `draft.js` and "much better" in the same sentence sounds great! :)

I think we should store HTML, because we'll also need to render comments in notifications, where we don't use React.

abowler2:
No longer moving forward with these changes.
Current direction can be found on #1623

PR #1591, opened 2020-03-22
Bug 1607696 - Bump the Django version to 2.2.11

jotes:
TODO:
* [x] - reorder the dependencies in requirements files
* [x] - write a changelog
* [x] - discuss `on_delete`
* [x] - bump `django-model-utils`


pontoon/base/models.py
```diff
@@ -3252,8 +3282,10 @@ class TranslationMemoryEntry(models.Model):
     translation = models.ForeignKey(
         Translation, null=True, on_delete=models.SET_NULL, related_name="memory_entries"
     )
-    locale = models.ForeignKey(Locale)
-    project = models.ForeignKey(Project, null=True, related_name="memory_entries")
+    locale = models.ForeignKey(Locale, models.CASCADE)
+    project = models.ForeignKey(
+        Project, models.CASCADE, null=True, related_name="memory_entries"
+    )
```
mathjazz: I suggest we `SET_NULL` if `project` is deleted, same as we do for `entity` and `translation`. The TM entry would still be useful.

jotes:
@mathjazz I've noticed that we have 2 styles of passing the `on_delete` argument:
* `field = models.ForeignKey('model', models.CASCADE)`
* `field = models.ForeignKey('model', on_delete=models.CASCADE)`
Which one is preferred?

mathjazz:
Thanks for the update!

My personal preference in this case is the positional argument, so I'd go with the first option.

jotes:
*IMPORTANT*

You have to deploy the commit with squashed migrations (be748906aab82d5e57db7273ea2e74abe1b5aba3) and run `manage.py migrate` on your Pontoon instance before this commit.

I know that this PR is huge, mostly because of the removed migrations.
It introduces a couple of small changes. This commit:
* removes old migrations
* bumps up the dependencies that couldn't run on Django 2.2
* adds missing on_delete clauses
* removes all `authenticated()` calls and uses the property instead
* uses the Django's constraints mechanism and removes PartialIndex
* updates the config of `django-notifications-hq` 

@mathjazz r?

mathjazz:
I also just noticed I can't log out.

mathjazz:
I think we'll also need to update this:
https://github.com/jotes/pontoon/blob/bug-1607696-bump-django-version/pontoon/base/models.py#L2342

jotes:
@mathjazz I fixed the issues with `django-notifications` and `allauth`, can you look again?

requirements/constraints.txt
```diff
@@ -371,3 +377,6 @@ Werkzeug==0.15.5 \
 # django-debug-toolbar
 sqlparse==0.1.19 \
     --hash=sha256:d896be1a1c7f24bffe08d7a64e6f0176b260e281c5f3685afe7826f8bada4ee8
+Jinja==1.2 \
```
mathjazz: Is this really a requirement of `django-debug-toolbar`?

jotes: It's not, I'm removing it right now.

PR #1593, opened 2020-03-24
Bug 1624557: Add specification for terminology presentation

mathjazz:
This is a specification of the first step out of at least three of bringing full terminology support to Pontoon. It covers the ability to show terms present in the source string, whereas term creation and translation aren't in scope here.

This patch renders the [old specification](https://docs.google.com/document/d/1fT3Wyk-rYOCXwV51pmscJORPxUmY41olPJolmhzdJOI/edit#) deprecated.


specs/0101-terminology-presentation.md
```diff
@@ -0,0 +1,41 @@
...
+# Feature explanation
+
+As the first step, we need the ability to store terms in the database. Data model should allow for importing terms stored in the TBX format and the [initial Mozilla terms list](https://docs.google.com/spreadsheets/d/1MAPD8WBnstR6pwKbNEDKOpw5CTPnl3qAobDgomdmtdY/edit?ts=5e79126c#gid=1146590716).
+
+The list of fields should include at least the following fields:
+- term
+- part of speech
+- definition
+- usage
+- translations
```
Pike: One or many translations? Jeff said that one-to-one is good enough, but we should be precise about that.

specs/0101-terminology-presentation.md
```diff
@@ -0,0 +1,41 @@
...
+
+Next, we need to identify stored terms in any original string used in translation workbench. We should take into account that terms can take various grammatical forms when used in strings.
```
Pike: What's the role of stemming here? How do we implement that? For which languages? Only source/English, or also target?

specs/0101-terminology-presentation.md
```diff
@@ -0,0 +1,41 @@
...
+
+Terms should be highlighted in the original string, similarly to how we highlight placeables. On term hover, a popup should appear with a list of matching terms and the corresponding metadata (description, part of speech and translation, if exists). On click, translation of the first term should be inserted into the editor.
```
Pike: Similar to the previous question, would we consider finding Terms in the target? Also Terms not in the source?

specs/0101-terminology-presentation.md
mathjazz: I was thinking of Term Translations as a separate table. I'll clarify that.

specs/0101-terminology-presentation.md
mathjazz: I think stemming is the solution to that problem, yeah. The `nltk` library should help with that. We probably only need it for matching original strings and terms, i.e. English.

specs/0101-terminology-presentation.md
mathjazz: If you mean to implement some sort've checks, then I'd say we should do it in a followup.

Pike:
Looking at http://www.nltk.org/howto/stem.html, it seems that stemming is hard (TM). And maybe not reasonable to implement, at least outside of English.

mathjazz:
> Looking at http://www.nltk.org/howto/stem.html, it seems that stemming is hard (TM). And maybe not reasonable to implement, at least outside of English.

I don't see the use case for stemming outside English as part of this bug.

And now that I'm thinking about it again, I'm wondering whether tokenization and stemming isn't too much of an implementation detail to be written down in a specification.

specs/0101-terminology-presentation.md
```diff
@@ -0,0 +1,42 @@
...
+- **Improving translation consistency**: it’s hard to expect all authors to always use the same translations for the same words and phrases without any help from the tool.
+- **Speeding up translation process**: localizers need to manually search for existing translations of terms found in the strings they are translating.
+- **Fixing brand translations**: trademarks present a special case for localization as they have legal and semantic significance, which isn't always obvious to localizers.
+- **Advancing content comprehension**: it's not always clear what the (standard) expression means, especially if the comment isn't provided.
+
+# Out of scope
+
+- **Terminology creation**: the ability to manually add, remove and review terms.
+- **Terminology translation**: the ability to manually add, remove and review term translations.
+- **VCS intergration**: the ability to sync terminology with VCS.
```
adngdb: s/intergration/integration

specs/0101-terminology-presentation.md
```diff
@@ -0,0 +1,42 @@
...
+- part of speech
+- definition
+- usage
+- translations (separate table)
+
+Next, we need to identify stored terms in any original string used in translation workbench. We should take into account that terms can take various grammatical forms when used in strings and use tokenization and stemming to mitigate that.
+
+Terms should be highlighted in the original string, similarly to how we highlight placeables. On term hover, a popup should appear with a list of matching terms and the corresponding metadata (description, part of speech and translation, if exists). On click, translation of the first term should be inserted into the editor.
+
+Aditionally, we could also introduce the Terminology panel in the 3rd column, next to Machinery, Locales and Comments. It would allow translators to see terms and the corresponding metadata all the time, without the need to hover. Translations of terms (if exist) would be inserted into the editor on click.
```
adngdb: s/Aditionally/Additionally

specs/0101-terminology-presentation.md
```diff
@@ -0,0 +1,42 @@
...
+
+# Mockup
+
+![](0101/mockup.png)
```
adngdb: The 3rd column is getting very crowded with this (your example doesn't show the Comments tab), should we maybe exchange the History panel with the Helpers tabs?

specs/0101-terminology-presentation.md
```diff
@@ -0,0 +1,42 @@
...
+- **VCS intergration**: the ability to sync terminology with VCS.
+- **Checks**: the ability to check if translations include term translations.
+
+# Feature explanation
+
+As the first step, we need the ability to store terms in the database. Data model should allow for importing terms stored in a TBX file and the [initial Mozilla term list](https://docs.google.com/spreadsheets/d/1MAPD8WBnstR6pwKbNEDKOpw5CTPnl3qAobDgomdmtdY/edit?ts=5e79126c#gid=1146590716).
+
+At least the following information needs to be saved for each term:
+- term itself
+- part of speech
```
adngdb: What is this? In general, this list could provide a bit more information about what each field is about.

adngdb:
> And now that I'm thinking about it again, I'm wondering whether tokenization and stemming isn't too much of an implementation detail to be written down in a specification.

I'd argue that it's quite a substantial detail, as it can turn a week of work into several. I think it needs to be very clear in the specification so that the developer can anticipate and look for appropriate tools.

specs/0101-terminology-presentation.md
mathjazz: Yeah. I was thinking more towards using icons instead of text when needed (and making tabs responsive), or the accordion menu.

specs/0101-terminology-presentation.md
mathjazz: https://en.wikipedia.org/wiki/Part_of_speech

Those are standard TBX fields.

adngdb: Ah, OK. Then maybe add a link to a description of the TBX format?

mathjazz: Done.

mathjazz:
I'd like to avoid blocking this feature on finding the perfect algorithm (which is a non-trivial task), and instead move forward with a good enough solution we can always iterate on.

As it turned out in our [previous attempt](https://github.com/mozilla/pontoon/pull/535), stemming tends to result in false positives (e.g. "extension" matches "extensibility"). I went through the entire `Mozilla_initial_terms_import` and the following algorithm finds all terms in their usage examples:

```
Order terms by length from longest to shortest.
If case_sensitive = True, lowercase the string and the terms.
Check if a perfect match of the term exists within the string.
If exact_match != True: 
  Tokenize the string.
  Singularize every token and every terms.
  Check if any term matches any of the tokens.
  Check if any term is a substring of any of the tokens.
```

Ordering of terms by length should prevent us to mark up "home" and "page" in "This is my home page", while "home page" is also available.

Without Singularization, we didn't find one term in its example:
* curriculum in "...curricula..."

Without Substring matching, 10 out of 213 terms are not found in their examples, e.g.:
* authenticated in "...unauthenticated..."
* corrupt in "...corrupted..."
* track in "...tracks..."
* surf in "...surfing..."

Before I fine-tune the algorithm above (e.g. define, what to do when we find the match), I'd like to hear your thoughts on this proposal.

mathjazz:
@Pike I've started working on this feature 😄. Do you have further requests before this is merged?

mathjazz:
Thanks. I just realized the spec doesn't say where the algorithm should be used. Given the conversation we had, maybe for the better. :)

PR #1600, opened 2020-03-30
Bug 1624557: Add Data model for Terminology

mathjazz:
This is the first part of the [Terminology Presentation](https://github.com/mozilla/pontoon/blob/master/specs/0101-terminology-presentation.md#feature-explanation) feature. It creates the data model for storing terms collected in the [initial Mozilla term list](https://docs.google.com/spreadsheets/d/1MAPD8WBnstR6pwKbNEDKOpw5CTPnl3qAobDgomdmtdY/edit?ts=5e79126c#gid=1146590716).

Since it's not clear yet how exactly we'll be importing/updating the terms (e.g. by importing TBX files or some 3rd party API), I tested the patch by simply converting that spreadsheet to CSV and then manually importing it. I'm not sure we should use data migration for that purpose, because we probably don't want other Pontoon installations to come preloaded with a set of terms.


pontoon/actionlog/admin.py
```diff
@@ -6,7 +6,6 @@
 
 
 class ActionLogAdmin(admin.ModelAdmin):
-    list_display = ("action_type",)
```
mathjazz: Noticed this duplication while copying the Admin code.

Pike:
I agree that the actual content shouldn't be landed in Pontoon.

PR #1173, opened 2019-01-19
 Bug 1521288 Pontoon should commit with multiple authors

tomer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1521288

PR #1601, opened 2020-03-30
Use Co-authored-by in the commit message.

XAMPPRocky:
The Co-authored-by trailer allows you to state that PR has multiple authors, and using this trailer allows platforms like GitHub to reflect that in the UI.

reference: https://help.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors

### DIFF
--- a/pontoon/sync/templates/sync/commit_message.jinja
+++ b/pontoon/sync/templates/sync/commit_message.jinja
@@ -1,7 +1,6 @@
 Pontoon: Update {{ locale.name }} ({{ locale.code }}) localization of {{ project.name }}
-{%- if authors %}
 
-Localization authors:
+{%- if authors %}
 {% for author in authors -%}
 Co-authored-by: {{ author.display_name_and_email|safe }}
 {% endfor %}
--- a/pontoon/sync/templates/sync/commit_message.jinja
+++ b/pontoon/sync/templates/sync/commit_message.jinja
@@ -3,6 +3,6 @@ Pontoon: Update {{ locale.name }} ({{ locale.code }}) localization of {{ project
 
 Localization authors:
 {% for author in authors -%}
-- {{ author.display_name_and_email|safe }}
+Co-authored-by: {{ author.display_name_and_email|safe }}
 {% endfor %}
 {%- endif -%}


mathjazz:
Thanks for the patch!

Please note that we WONTFIX-ed [bug 1521288](https://bugzilla.mozilla.org/show_bug.cgi?id=1521288) and closed the related PR #1173. Since we failed to implement any of the proposed solutions since then, I'm OK with merging this patch. Provided we list @tomer under Co-authored-by. 😄

I'll leave the final decision to @Pike, though.

XAMPPRocky:
Thank you, I didn't check. What I would say against the original motivation for the wontfix, is that you shouldn't block low cost quality of life improvements behind larger solutions, for exactly that reason that often the larger solution is not done.

And what I can say towards whether l10n contributions are appropriate under `co-authored-by`  is as one of the maintainers of the rust-lang pontoon. Is that it feels quite odd that commits that could be authored by four or five individuals are shown as authored by one in the UI.

Pike:
I'm not terribly against this, but I'm also not a fan. I haven't seen any concrete information on how this makes things better, outside the gh UI.

Can you rebase your PR onto the current master of Pontoon? Right now it's a fork of the rust-lang fork, and that's tricky. At least I think that's why the tests are failing.

XAMPPRocky:
@Pike Oh sorry, that explains the CI failure. I made the change from the GitHub UI, and apparently it doesn't behave like you expect if you already have a fork. I'll make the changes and add the previous PR author as a co author.

> I'm not terribly against this, but I'm also not a fan. I haven't seen any concrete information on how this makes things better, outside the gh UI.

I'm not arguing that the change would be useful outside of the gh UI, I'd still want this even if it was called `X-Github-Co-Authored-By`. Our entire our workflow is based around GitHub, for me the more pontoon feels integrated with GitHub, the better it feels to use it.

XAMPPRocky:
@Pike Actually investigating it, the history of the branch is correct and points to mozilla/pontoon, so it seems like this failure is a problem with the current master. You might have been confused because my repo is technically a fork of the rust-lang fork.

Screenshot 2020-03-30 at 22 37 09

tomer:
> I'm not arguing that the change would be useful outside of the gh UI, I'd still want this even if it was called `X-Github-Co-Authored-By`. Our entire our workflow is based around GitHub, for me the more pontoon feels integrated with GitHub, the better it feels to use it.

The co-authored-by is used in Github to credit people who provided help in the commit such as reviewers etc. In the Pontoon case it'd credit translators who translated some strings in the commit, and will be visible in their personal profile dashboards. 

Commit 9d0777f148f546b0827881e65d21292745fa238b can be used as demonstration, as I've credited @mozilla-pontoon as co-author there. (hover the link to the commit to see it)

XAMPPRocky:
@tomer This PR, is also now an example with you credited. 😄

pontoon/sync/templates/sync/commit_message.jinja
```diff
@@ -3,6 +3,6 @@ Pontoon: Update {{ locale.name }} ({{ locale.code }}) localization of {{ project
 
 Localization authors:
```
mathjazz: I'd say we can remove this line now.

XAMPPRocky: Resolved

mathjazz:
In case you are interested in making rust-lang pontoon commits even better, we'd appreciate a patch to de-hardcode VCS identity:
https://bugzilla.mozilla.org/show_bug.cgi?id=1567400

PR #1607, opened 2020-04-05
Fix bug 1609870: Pull the repos only once

VishalCR7:
As stated in [bug-1609870](https://bugzilla.mozilla.org/show_bug.cgi?id=1609870) I have tried to `pull_changes` only once for the source repo. 

While most of the code and logic remains the same, I had to let go of a single optimization [here](https://github.com/mozilla/pontoon/blob/master/pontoon/sync/tasks.py#L243), which indirectly depends upon the results of 1st `pull_changes` call.  

I would like to know your thoughts about it.

I'm also looking to fix [bug-1406374](https://bugzilla.mozilla.org/show_bug.cgi?id=1406374) once current changes are approved.


pontoon/sync/tasks.py
```diff
@@ -99,38 +98,74 @@ def sync_project(
...
     # because that would apply any source string changes to the specified locale only.
+    if locale:
+        locales = db_project.locales.filter(pk=locale.pk)
+    else:
+        locales = db_project.locales.all()
+
+    if not locales:
+        log.info(
+            "Skipping syncing translations for project {0}, no locales to sync "
+            "found within.".format(db_project.slug)
```
mathjazz, line 109: We're now skipping the entire sync (not just translations, but also source strings), so let's change this to:
```
log.info("Skipping project {0}, no locales to sync found.".format(db_project.slug))
```

pontoon/sync/tasks.py
```diff
@@ -172,88 +203,36 @@ def sync_sources(db_project, now, force, no_pull):
...
-        locales = get_changed_locales(db_project, locales, now)
-
-    readonly_locales = db_project.locales.filter(project_locale__readonly=True)
     added_and_changed_resources = db_project.resources.filter(
         path__in=list(added_paths or []) + list(changed_paths or [])
     ).distinct()
 
-    # We should also sync files for which source file change - but only for read-only locales.
-    # See bug 1372151 for more details.
-    if added_and_changed_resources:
```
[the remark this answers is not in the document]
VishalCR7: So added_or_changed resources depend on `update_originals` in `sync_sources` which depends on 1st pull_changes call in original code i.e. the 1st call to sync the source repo actually has an impact at 2nd call, which makes it hard to `pull_changes` only once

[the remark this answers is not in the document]
VishalCR7: However, what can be done is keep the original code and just pull the changes for source repo only once, i.e. call the 2nd pull only  when we have the locale parameter passed to `sync_translation` or when we have multiple repositories ( but do it for translation_repo only)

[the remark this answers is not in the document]
mathjazz: Ah, I see the problem. Right now we'll pull read-only locales all the time, which isn't optimal.

I guess we can move that block of code (starting with `if not locale:`) right above the call to `sync_translations()` and then pull repositories of read-only locales if not yet pulled and `if added_and_changed_resources == True`?

pontoon/sync/tasks.py
```diff
@@ -172,100 +199,36 @@ def sync_sources(db_project, now, force, no_pull):
...
-            pk__in=changed_locales_pks + readonly_locales_pks
-        )
-
-    # Pull VCS changes in case we're on a different worker than the one
-    # sync started on.
-    if not no_pull:
-        log.info("Pulling changes for project {0} started.".format(db_project.slug))
-        repos_changed, repo_locales = pull_changes(db_project, locales)
+    if repo_locales:
         repos = repos.filter(pk__in=repo_locales.keys())
```
mathjazz, line 229: Looks like `repos` isn't used before line 362, so we can join both `if repo_locales:` into one down under.

pontoon/sync/tasks.py
```diff
@@ -97,62 +92,61 @@ def sync_project(
...
-    if locale:
-        source_changes = {}
-    else:
-        source_changes = sync_sources(db_project, now, force, no_pull)
-        # Skip syncing translations if we already know there's nothing to sync
-        if not source_changes:
-            project_sync_log.skip()
-            return
+    source_changes = sync_sources(db_project, now, force, no_pull)
+    # Skip syncing translations if no source directory found
```
mathjazz, line 96: Nit: This comment should now also say what it said originally:
```
Skip syncing translations if we already know there's nothing to sync
or if no source directory found.
```

pontoon/sync/tasks.py
```diff
@@ -97,62 +92,61 @@ def sync_project(
...
+        log.info(
+            "Pulling source changes for project {0} started.".format(db_project.slug)
+        )
+        source_changed = pull_source_repo_changes(db_project)
+        log.info(
+            "Pulling source changes for project {0} complete.".format(db_project.slug)
+        )
 
-    # If the only repo hasn't changed since the last sync and there are
+    # If the repos haven't changed since the last sync and there are
```
mathjazz, line 130: Nit: You can revert this comment to the previous version now.

pontoon/sync/tasks.py
```diff
@@ -97,62 +92,61 @@ def sync_project(
...
+        force=force,
     )
 
 
 def sync_sources(db_project, now, force, no_pull):
-    # Pull source repository
+    # Pull from source repository
     if no_pull:
-        source_repo_changed = True  # Assume changed
+        source_changed = True
```
mathjazz, line 120: What's the purpose of this change? The previous name seems more verbose and less likely to be confused with `source_changes`.

If you still want to rename the variable, I'd suggest prepending booleans with is/are/has/have like this:
- `source_repo_changed` -> `has_source_repo_changed`

pontoon/sync/tasks.py
```diff
@@ -97,62 +92,61 @@ def sync_project(
...
             added_paths, removed_paths, changed_paths, new_entities = update_originals(
-                db_project, now, full_scan=force
+                db_project, now, force=force
             )
         except MissingSourceDirectoryError as e:
             log.error(e)
             return False
 
-        if not db_project.has_single_repo:
-            db_project.source_repository.set_last_synced_revisions()
```
mathjazz, line 155: It seems like you can keep revert this change now.

And then you don't need any of this:
https://github.com/mozilla/pontoon/pull/1607/files#diff-98331b44e2eb35ae952822657cdd5f6bR173
https://github.com/mozilla/pontoon/pull/1607/files#diff-98331b44e2eb35ae952822657cdd5f6bR221
https://github.com/mozilla/pontoon/pull/1607/files#diff-98331b44e2eb35ae952822657cdd5f6bR236

pontoon/sync/tasks.py
```diff
@@ -396,10 +378,12 @@ def sync_translations(
...
-    for r in repos:
-        r.set_last_synced_revisions(
-            locales=repo_locales[r.pk].exclude(code__in=failed_locales)
-        )
+    if repo_locales:
+        repos = db_project.repositories.filter(pk__in=repo_locales.keys())
+        for r in repos:
+            r.set_last_synced_revisions(
+                locales=repo_locales[r.pk].exclude(code__in=failed_locales)
+            )
```
mathjazz, line 386: I think you can also keep the original code here, if you set the last synced revision for the source repository separately.

pontoon/sync/tasks.py
VishalCR7: Since the code was built upon the previous changes, this stayed. I will follow  "has_changed" convention that is used elsewhere

pontoon/sync/tasks.py
mathjazz: As agreed via chat, it's actually better to set last synced revision at the end of the process to prevent bugs like https://bugzilla.mozilla.org/show_bug.cgi?id=1383271.

pontoon/sync/tasks.py
mathjazz: This comment is now also obsolete.

pontoon/sync/tasks.py
```diff
@@ -247,18 +218,30 @@ def sync_translations(
...
+        # Pull repos of locales in case of multi_locale_project
+        if not db_project.has_single_repo:
+            log.info(
+                "Pulling locale repos for project {0} started.".format(db_project.slug)
+            )
+            locale_repos_changed, not_pulled_repo_locales = pull_locale_repo_changes(
+                db_project, locales
+            )
+
+            repos_changed = repos_changed or locale_repos_changed
```
mathjazz, line 233: I'd change these names a little:
```python
have_repos_changed = has_source_repo_changed or have_locale_repos_changed
```

pontoon/sync/tasks.py
```diff
@@ -247,18 +218,30 @@ def sync_translations(
...
-        repos = repos.filter(pk__in=repo_locales.keys())
-        log.info("Pulling changes for project {0} complete.".format(db_project.slug))
+        repo_locales = {db_project.source_repository.pk: Locale.objects.none()}
+
+        # Pull repos of locales in case of multi_locale_project
+        if not db_project.has_single_repo:
+            log.info(
+                "Pulling locale repos for project {0} started.".format(db_project.slug)
+            )
+            locale_repos_changed, not_pulled_repo_locales = pull_locale_repo_changes(
```
mathjazz, line 229: Why `not_pulled_repo_locales`, if they are actually pulled?

`not_yet_pulled_repo_locales` would be more accurate, but that's pretty long. Simply `pulled_repo_locales` will do.

pontoon/sync/tasks.py
```diff
@@ -247,20 +218,33 @@ def sync_translations(
...
+            log.info(
+                "Pulling locale repos for project {0} started.".format(db_project.slug)
+            )
+            have_locale_repos_changed, pulled_repo_locales = pull_locale_repo_changes(
+                db_project, locales
+            )
+
+            have_repos_changed |= have_locale_repos_changed
+            log.info(
+                "Pulling locale repos for project {0} complete.".format(db_project.slug)
```
mathjazz, line 238: Sorry, one more: can you move this line right after line 234, where we pull?

PR #1609, opened 2020-04-06
Fix Bug 1611510 - Implement Source String Comments

abowler2:
This initial work towards implementing source string comments replaces the textarea for the comments with a [draft.js](https://github.com/facebook/draft-js) editor. 

______
**OF NOTE:**  
Flow threw a lot of errors directed at the draft.js `lib` file.  For the moment, I have added that to `ignore` within the `flowconfig`.  Happy to address this another way but might need you to direct me on where to look for a solution to that. 

**CURRENT KNOWN FIXES NEEDED:**
1. Linkify is currently not working due the way comments are now rendered.  There is a plugin that I will add which will replicate this functionality.
2. The comments are now displaying below the username/email when rendered. I have not been able to work out how to fix this just yet, or even if it can be.
3. The editor handles new lines differently.  I have removed `white-space: pre-wrap` from the CSS as a result, but that will cause any existing comments to ignore any new lines that are present. Not sure the best approach to deal with this.

____
### TODO:

- [ ] Convert existing comments to [new format](https://github.com/mozilla/pontoon/pull/1609#discussion_r406739073)
- [ ] Fix or file [notification/focus bug](https://github.com/mozilla/pontoon/pull/1609#discussion_r405893246)


frontend/src/core/comments/components/Comment.css
```diff
@@ -19,7 +19,6 @@
     display: flex;
     font-size: 11px;
     padding: 6px;
-    white-space: pre-wrap;
```
mathjazz, line 22: I don't understand your comment - why do we need to remove this declaration?

frontend/src/core/comments/components/AddComment.css
```diff
@@ -13,13 +13,20 @@
...
+    max-height: 144px;
     line-height: 24px;
     overflow: auto;
-    padding: 6px;
+    padding: 5px;
     resize: none;
 }
 
+.public-DraftEditorPlaceholder-inner {
+  position: absolute;
```
mathjazz, line 25: That causes the placeholder to stay at the same position as I scroll the History panel:

Screenshot 2020-04-07 at 17 35 37

frontend/src/core/comments/components/AddComment.js
```diff
@@ -25,69 +34,87 @@ export default function AddComments(props: Props) {
...
-                    autoFocus
-                    name='comment'
-                    dir='auto'
-                    placeholder={ `Write a comment…` }
-                    rows={ minRows }
-                    ref={ commentInput }
-                    onChange={ handleOnChange }
-                    onKeyDown={ handleOnKeyDown }
+        <div className='container'>
+            <div className='comment-div' onClick={ focusEditor }>
```
mathjazz, line 108: Why do we need this wrapper (instead of putting the `onClick` to `Editor` directly)?

Or even: why do we even need to set focus? Shouldn't it be set automatically on click, like in any editor? It also surprises me cursor isn't properly set (to `help`) when you hover the editor before it has focus.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -25,69 +34,87 @@ export default function AddComments(props: Props) {
...
 
-    const submitComment = (event: SyntheticEvent<>) => {
-        event.preventDefault();
-        const comment = commentInput.current.value;
+    React.useEffect(() => {
+        focusEditor()
+    }, []);
 
-        if (!comment) {
-            return null;
```
mathjazz, line 61: Looks like we lost that - empty comments can be submitted.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -25,69 +34,87 @@ export default function AddComments(props: Props) {
...
 
-        commentInput.current.value = '';
-        commentInput.current.rows = minRows;
-    };
+        clearEditorContent();
+    }
+
+    if (!user) {
+        return null;
+    }
```
mathjazz, line 100: This one should be somewhere up, so that we don't execute all the code for non-authenticated users.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -25,69 +34,87 @@ export default function AddComments(props: Props) {
...
+    }, []);
 
-        if (!comment) {
-            return null;
+    const handleKeyCommand = (command: string) => {
+        if (command === 'submitOnEnter') {
+            submitComment();
+            return 'handled';
         }
+        return 'not-handled';
```
mathjazz, line 68: Interesting - is that a requirement of draftjs to return these values?

frontend/src/core/comments/components/AddComment.js
```diff
@@ -25,69 +34,87 @@ export default function AddComments(props: Props) {
...
     } = props;
 
-    const commentInput: any = React.useRef();
-    const minRows = 1;
-    const maxRows = 6;
+    let [editorState, setEditorState] = React.useState(
+        EditorState.createEmpty()
+    )
+    const editor: any = React.useRef(null);
+    const rawContentState = convertToRaw(editorState.getCurrentContent());
```
mathjazz, line 41: Should we only do the conversion before saving?

frontend/src/core/comments/components/AddComment.js
```diff
@@ -25,69 +34,87 @@ export default function AddComments(props: Props) {
...
-            commentInput.current.rows = maxRows;
-        }
+    function focusEditor() {
+        editor.current.focus();
     }
 
-    const handleOnKeyDown = (event: SyntheticKeyboardEvent<>) => {
-        if (event.keyCode === 13 && event.shiftKey === false) {
-            submitComment(event);
+    function keyBindingFn(e: SyntheticKeyboardEvent<>) {
```
mathjazz, line 51: It seems quite complicated how you handle the keyboard shortcut for Enter. Do we need to separate methods?

frontend/.flowconfig
```diff
@@ -1,4 +1,5 @@
 [ignore]
+.*\/node_modules\/draft-js\/lib\/.*.js.flow.*
```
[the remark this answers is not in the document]
abowler2: When I remove the line and run `make build` I get the errors again. 

I put a copy of my terminal log [HERE](https://pastebin.com/xiZLqYfa)

frontend/src/core/comments/components/Comment.css
abowler2: The comments are saved as strings in the original implementation of comments. However, with this implementation they are saved as html, so each line is a separate `<p>` . The `pre-wrap`, of course, handles each differently.  So, consideration needs to be made on how to handle previous comments that have line breaks in them.

without-prewrap


with-prewrap

frontend/src/core/comments/components/AddComment.js
abowler2: As I understand it, yes.  Since we are in essence creating a custom key binding to trigger the submit function on `enter`

Docs:  https://draftjs.org/docs/advanced-topics-key-bindings

frontend/src/core/comments/components/AddComment.js
abowler2: This relates to the above [comment](https://github.com/mozilla/pontoon/pull/1609#discussion_r405242400).  I can look into this further, but as I understand it, yes.

frontend/src/core/comments/components/AddComment.js
abowler2: Yes. I have moved it into the submit function.

frontend/src/core/comments/components/AddComment.js
abowler2: This check is no longer working.  I will continue to work on this to find out how to prevent an empty entry.

frontend/src/core/comments/components/AddComment.js
abowler2: The editor gets assigned its own `className` and it does not add a secondary class if one is given. I tried to set the styling to that `className`, but that caused additional issues. So, the only way to get the proper styling is to have the wrapper. 

Removing the `onClick` makes it so that you can't focus on the input area at all, and moving the `onClick` into the `Editor` causes the cursor to disappear if you click anywhere else and it does not refocus when you click back.

frontend/src/core/comments/components/AddComment.css
abowler2: I wasn't able to find an immediate solution here. I will continue to work on this.

frontend/src/core/comments/components/Comment.js
```diff
@@ -41,7 +41,9 @@ export default function Comment(props: Props) {
                         { comment.author }
                     </a>
                     <Linkify properties={ { target: '_blank', rel: 'noopener noreferrer' } }>
-                        { comment.content }
+                        {/* We can safely use comment.content because it is validated by
+                        bleach before being saved into the database. */}
+                        <span dangerouslySetInnerHTML={ { __html: comment.content } } />
```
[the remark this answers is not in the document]
abowler2: draftToHtml() actually saves the content **AS** `html`. 


Screen Shot 2020-04-07 at 9 29 54 PM

abowler2:
Thanks for taking a look at the progress so far.  I was able to make some of the suggested changes and addressed the other questions/comments you had within the comments above.

> The comments are now displaying below the username/email when rendered. I have not been able to work out how to fix this just yet, or even if it can be.
> 
> That's because they are wrapped in `<p>`. We should prevent that.

This is how the comments are saved as they are converted to `html` using `draftToHtml()`

frontend/src/core/comments/components/AddComment.js
abowler2: Found the fix for this.

However, in the process I discovered a bug that only occurs in the translation comments but happens on both the original comment implementation and with this editor version. 
>When you submit an empty comment, then submit a normal comment, and then try to submit an empty comment again the `Same translation already exists` error displays.

frontend/src/core/comments/components/AddComment.css
abowler2: I believe this is now fixed.

frontend/.flowconfig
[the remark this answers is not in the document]
mathjazz: Gotcha. I see a lot of reports on the interwebs about this issue, and ignoring seems to be the only workaround. Maybe @adngdb has better ideas?

I tried upgrading flow to the latest version, but it didn't help either.

frontend/src/core/comments/components/AddComment.js
mathjazz: Thanks! The bug happens because focus is set to the text editor after you successfully submit a comment. It shouldn't be related to preventing empty comments from being submitted.

If it's a non-trivial fix, we can add it to the list of bugs to be filed and fixed in a followup. I usually keep such list in the first comment of the PR and file bugs after merging.

frontend/src/core/comments/components/AddComment.js
mathjazz: Nit: It can go further up, right after you declare props consts. No need to use state, ref or focus if you don't have a user.

frontend/src/core/comments/components/AddComment.js
mathjazz: Thanks for the pointer, it makes sense now!

frontend/src/core/comments/components/AddComment.js
mathjazz: Thanks! Please mind the indent and the colon.

Super nit: it sees more "drafty" to name the command with "kebab-case".

frontend/src/core/comments/components/Comment.js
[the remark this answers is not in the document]
mathjazz: Oh, I meant the manually entered HTML. But I forgot it's a rich editor. I see what it's doing now. https://www.npmjs.com/package/draftjs-to-html

Please ignore my comment.

We should remove the Linkify component though. Eventually we should replace it with a plugin: https://www.draft-js-plugins.com/

frontend/src/core/comments/components/Comment.css
mathjazz: OK, I see why removing `pre-wrap makes sense. We can convert existing comments to the new format with a data migration. I think that answer your point 3.

mathjazz:
> This is how the comments are saved as they are converted to `html` using `draftToHtml()`

I see. It seems like we can't easily remove the `<p>`'s. What if we just make the first one `display: inline`?

```css
.comments-list .comment .content p:first-child {
    display: inline;
}
```

frontend/src/core/comments/components/AddComment.css
mathjazz: Thanks. Setting width to a fixed value will likely cause problems with localizability. Can we make it relative, i.e. to occupy the entire width available?

I wonder if [Draft.css](https://draftjs.org/docs/advanced-topics-issues-and-pitfalls/#missing-draftcss) can help with styling issues? I just came across it in the docs, and I didn't dig deeper, so I'm likely to be off. :)

frontend/src/core/comments/components/AddComment.js
mathjazz: OK. I'd change the class name to something more semantic, e.g. `.editor-wrapper`.

I noticed that with the following CSS
```css
.DraftEditor-root {
    height: 100%;
    width: 100%;
}
```

I solve a few issues:
- Focus on the editor JustWorks(TM). It's still a good idea to set it onClick, in case user clicks on the padding.
- Cursor is properly set to `cursor: text;`.
- Placeholder no longer requires a [fixed width](https://github.com/mozilla/pontoon/pull/1609#discussion_r406755592).

So I wonder if it's expected for `.DraftEditor-root` to occupy zero space and hence we need to manually set focus, cursor, placeholder width...

Looking at this article, I'd say no:
https://medium.com/@khcarr/styling-the-draft-js-editor-e14152a2472b

frontend/src/core/comments/components/AddComment.js
abowler2: To clarify the 'super nit': Do you mean change from `submitOnEnter` to `submit-on-enter`?

mathjazz: Correct :)

frontend/src/core/comments/components/AddComment.js
abowler2: Actually, I can't per the "rules of hooks" 😦 

Screen Shot 2020-04-10 at 8 14 01 AM

mathjazz: Oh, wow! I had no idea. Sorry. 🤷‍♂️ 😸

abowler2:
I believe we now have a stable editor to begin expanding on  🎉 

Of course, let me know if you find any issues or have any further changes you would like.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -25,69 +34,91 @@ export default function AddComments(props: Props) {
...
+        const lastBlock = contentState.getLastBlock();
+        const allSelected = new SelectionState({
+            anchorKey: firstBlock.getKey(),
+            anchorOffset: 0,
+            focusKey: lastBlock.getKey(),
+            focusOffset: lastBlock.getLength(),
+            hasFocus: true
+        });
+        contentState = Modifier.removeRange(contentState, allSelected, 'backward');
+        editorState = EditorState.push(editorState, contentState, 'remove-range');
```
adngdb, line 86: Unless I missed something, this is replacing the `editorState` that you defined with `useState` outside this function. I assume that's not what you intended to do, especially considering you're using `setEditorState` right below. I'd recommend using a different variable name here, or at least declaring it with `let` to make sure it's not the same as in the component's scope.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -25,69 +34,91 @@ export default function AddComments(props: Props) {
...
-                    rows={ minRows }
-                    ref={ commentInput }
-                    onChange={ handleOnChange }
-                    onKeyDown={ handleOnKeyDown }
+        <div className='container'>
+            <div className='editor-wrapper' onClick={ focusEditor }>
+                <Editor
+                    ref={ editor }
+                    editorState={ editorState }
+                    placeholder='Write a comment…'
```
adngdb, line 116: Why is this not localized anymore? A limitation of this new tool?

frontend/src/core/comments/components/AddComment.js
```diff
@@ -25,69 +34,91 @@ export default function AddComments(props: Props) {
...
+    }
 
     return <div className='comment add-comment'>
         <UserAvatar
             username={ username }
             imageUrl={ imageURL }
         />
-        <form className='container'>
-            <Localized
-                id='comments-AddComment--input'
```
adngdb, line 77: Note that if this string is not being used anymore, it should be removed from the en-US source FTL file.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -25,69 +34,91 @@ export default function AddComments(props: Props) {
...
-                    autoFocus
-                    name='comment'
-                    dir='auto'
-                    placeholder={ `Write a comment…` }
-                    rows={ minRows }
-                    ref={ commentInput }
-                    onChange={ handleOnChange }
-                    onKeyDown={ handleOnKeyDown }
+        <div className='container'>
+            <div className='editor-wrapper' onClick={ focusEditor }>
```
mathjazz, line 112: If we're focusing on click on the wrapper, we should also set `cursor: text` on it.

frontend/src/core/comments/components/AddComment.js
mathjazz: This method seems pretty cumbersome to just clear the content. Can we instead use something like this?
https://stackoverflow.com/questions/37463757/how-to-clear-input-field-in-draft-js/39029383#39029383

I've also found clearEditorContent in https://github.com/jpuri/draftjs-utils.

abowler2: Sadly, there is no easy way to reset the state and clear the editor. This solution was from an [issue submitted](https://github.com/facebook/draft-js/issues/445) about this on the draft.js repo. (I had to correct line 87 to be `...forceSelection(**editorState**, ....) as the solution given did not work as written and was admittedly (via the disclaimer) not tested prior to posting)

I did see that the issue post mentioned above had `let { editorState } = this.state;` , but when I tried to adjust given that we are using hooks it did not work. Also, giving this a different variable name does not work either. In essence, neither of those options clears the contents and allows for a new comment to be posted.

If you have an alternate way of approaching this, I would be more than happy to pursue it.

frontend/src/core/comments/components/AddComment.js
abowler2: I didn't realize I had removed it ☹️ 

It is corrected now.

frontend/src/core/comments/components/AddComment.js
abowler2: @mathjazz I tried that first solution with no luck, and I'm pretty sure I also tried the second solution (drafjs-utils), also with no luck, but will try again as I can't quite remember (I tried MANY options for this problem... 😄 )

mathjazz: Maybe use [createFromText()](https://draftjs.org/docs/api-reference-content-state/#createfromtext)?

```
editorState = EditorState.createWithContent(ContentState.createFromText(''));
setEditorState(editorState);
```

abowler2: @mathjazz   `createFromText()` introduced 2 new bugs: 
1. the comment now renders on top of the existing comments instead of the bottom.
2. The editor exhibits weird behaviors on the next entry (cursor moves in front of the first char entered and the ability to delete text is gone as well as possible other issues)

Also, after taking another look at `draftjs-utils` it appears that `clearrEditorContent` is essentially the same implementation that I have (I still can't remember if added and tested it out, but I feel like I did.  Either way, I will still give it a go, but wanted you to be aware that while it would cleaner on the surface it would essentially be the same 'under the hood')

abowler2:
>    1. The submit button starts to shrink after the editor increases to occupy two lines.

I wasn't able to reproduce this.

> 
>     2. Very long word without spaces (e.g. 200 characters) break the layout (which they don't on master). If you submit such comment, it's broken in both, your branch and master.

Fixed and pushed 😄

mathjazz:
> I wasn't able to reproduce this.

Interesting. So if you add enough text to the input to make it span say 3 lines, the submit button size is exactly the same as if the editor is empty?

abowler2:
> Interesting. So if you add enough text to the input to make it span say 3 lines, the submit button size is exactly the same as if the editor is empty? For me it shrinks for a few pixels.

Yes.  It doesn't appear to change for me (either visually or from the inspector)

Screen Shot 2020-04-10 at 1 29 45 PM

Screen Shot 2020-04-10 at 1 30 20 PM

mathjazz:
Note: it doesn't shrink for me either if I break into new lines by manually adding a line break, like it seems you did in the screenshot. It shrinks however when I add enough text to occupy more than 1 or 2 lines.

mathjazz:
Excellent, it no longer shrinks! :-)

Note that `cursor:text;` should be set in CSS, not React. :-)

What are the next steps here? Integrating plugins (linkify, mentions)?

frontend/src/core/comments/components/AddComment.js
```diff
@@ -25,69 +34,96 @@ export default function AddComments(props: Props) {
...
-                    dir='auto'
-                    placeholder={ `Write a comment…` }
-                    rows={ minRows }
-                    ref={ commentInput }
-                    onChange={ handleOnChange }
-                    onKeyDown={ handleOnKeyDown }
-                />
-            </Localized>
+        <div className='container'>
+            <div className='editor-wrapper' onClick={ focusEditor } cursor={ 'text' }>
```
mathjazz, line 112: We should remove that `cursor={ 'text' }`.

abowler2:
Yes, this should be ready for adding plugins.

Did you still want me to try to implement `clearEditorContent`?

mathjazz:
I think adding a TODO comment explaining we’re aware it’s ugly but unaware of less ugly approach is sufficient for the time being. Let’s first make sure we can actually fullfill the required implementation and then make it prettier.

abowler2:
@mathjazz @adngdb In moving on to add plugins I decided to add the `linkify` plugin first as it was the most straightforward.  However, it was not working, and I couldn't find any reason why it wouldn't work. 

In digging around I noticed that the docs for the plugins only referenced `class` components so out of curiosity I put together 2 different sandboxes on CodeSandbox, one with [hooks](https://codesandbox.io/s/draftjs-linkify-plugin-hooks-3r22l) and one with [class](https://codesandbox.io/s/draftjs-linkify-plugin-class-l7rdy?file=/src/index.js) (I'm pretty sure I have the versions correct on the dependencies too 😊 ). The `hooks` sandbox does not work, but the `class` one does.

Not sure what would need to be done to have hooks work, or if I will need to convert this to a `class` component. Your thoughts?

frontend/src/core/comments/components/AddComment.css
```diff
@@ -1,32 +1,53 @@
...
     border-radius: 4px;
     color: #FFFFFF;
+    cursor: text;
     display: flex;
     flex-grow: 2;
     font-size: 11px;
-    height: auto;
+    min-height: 26px;
+    max-height: 144px;
     line-height: 24px;
```
mathjazz, line 19: Let's bump the `line-height` to `26px`, otherwise the text jumps a little when the comment breaks into the 2nd line.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -2,8 +2,20 @@
...
+    EditorState, 
+    getDefaultKeyBinding, 
+    convertToRaw,
+    SelectionState,
+    Modifier, 
+} from 'draft-js';
+import draftToHtml from 'draftjs-to-html';
+import createLinkifyPlugin from 'draft-js-linkify-plugin';
 
 import './AddComment.css';
```
mathjazz, line 16: I think we should also add the default `draft.js` CSS file:
```
import 'draft-js/dist/Draft.css';
```

That will greatly reduce our custom CSS rules, to only include ones that are absolutely needed to adapt the style for our use case. Here's the patch, which also prevents comment text in the editor from jumping by a pixel or two after text breaks into the 2nd line.

```diff
diff --git a/frontend/src/core/comments/components/AddComment.css b/frontend/src/core/comments/components/AddComment.css
index 2bdb48d3d..e59baed7a 100644
--- a/frontend/src/core/comments/components/AddComment.css
+++ b/frontend/src/core/comments/components/AddComment.css
@@ -11,7 +11,6 @@
     border-radius: 4px;
     color: #FFFFFF;
     cursor: text;
-    display: flex;
     flex-grow: 2;
     font-size: 11px;
     min-height: 26px;
@@ -22,18 +21,8 @@
     resize: none;
 }
 
-.DraftEditor-root {
-    height: 100%;
-    width: 100%;
-}
-
-.public-DraftEditorPlaceholder-root {
-    position: relative;
-    color: #aaa;
-}
-
 .public-DraftEditorPlaceholder-inner {
-    position: absolute;
+    color: #aaa;
 }
 
 .public-DraftEditor-content {
diff --git a/frontend/src/core/comments/components/AddComment.js b/frontend/src/core/comments/components/AddComment.js
index a8653a633..23c260790 100644
--- a/frontend/src/core/comments/components/AddComment.js
+++ b/frontend/src/core/comments/components/AddComment.js
@@ -14,6 +14,7 @@ import draftToHtml from 'draftjs-to-html';
 import createLinkifyPlugin from 'draft-js-linkify-plugin';
 
 import './AddComment.css';
+import 'draft-js/dist/Draft.css';
 import 'draft-js-linkify-plugin/lib/plugin.css';
 
```

mathjazz:
> Not sure what would need to be done to have hooks work, or if I will need to convert this to a `class` component. Your thoughts?

This seems unrelated to hooks/class. It's focus that's causing the problem:
https://github.com/draft-js-plugins/draft-js-plugins/issues/800

The workaround requires another comment explaining "why so ugly", but it's a workaround:
```diff
diff --git a/frontend/src/core/comments/components/AddComment.js b/frontend/src/core/comments/components/AddComment.js
index 23c260790..8f83414f8 100644
--- a/frontend/src/core/comments/components/AddComment.js
+++ b/frontend/src/core/comments/components/AddComment.js
@@ -44,7 +44,9 @@ export default function AddComments(props: Props) {
     const editor: any = React.useRef(null);
 
     React.useEffect(() => {
-        focusEditor()
+        setTimeout(() => {
+            focusEditor();
+        }, 0);
     }, []);
 
     if (!user) {
```

abowler2:
No longer moving forward with these changes.
Current direction can be found on #1623

PR #1612, opened 2020-04-08
Fix bug 1628256 - Update @fluent/react library and our usage of its API

adngdb:
Hopefully my regex talents have not failed me and I have not missed anything in this update, because there won't be any errors if I did, but things will be broken. :-)

Assigning @stasm for review because it's Fluent-specific, but @mathjazz you might want to give it a look and check that I haven't forgotten anything.

Note that I made organized commits to make it easier to review. :angel:


frontend/src/modules/otherlocales/components/Translation.js
```diff
@@ -59,22 +59,18 @@ export default class Translation extends React.Component<Props> {
...
-                        <Localized
-                            id='otherlocales-Translation--header-link'
-                            $locale={ translation.locale.name }
-                            $code={ translation.locale.code }
-                        >
-                            <div>
-                                { translation.locale.name }
-                                <span>{ translation.locale.code }</span>
-                            </div>
-                        </Localized>
```
mathjazz, line 71: Oops :) Nice catch!

frontend/package.json
```diff
@@ -3,10 +3,10 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@fluent/bundle": "^0.15.0",
-    "@fluent/langneg": "^0.3.0",
-    "@fluent/react": "^0.10.0",
-    "@fluent/syntax": "^0.15.0",
+    "@fluent/bundle": "0.15.0",
+    "@fluent/langneg": "0.3.0",
```
stasm, line 7: `@fluent/langneg` 0.4.0 is out, too, now rewritten to TypeScript. No API changes.

frontend/package.json
```diff
@@ -3,10 +3,10 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@fluent/bundle": "^0.15.0",
-    "@fluent/langneg": "^0.3.0",
-    "@fluent/react": "^0.10.0",
-    "@fluent/syntax": "^0.15.0",
+    "@fluent/bundle": "0.15.0",
```
stasm, line 6: I published a small fix in `@fluent/bundle` 0.15.1 on Tuesday.

frontend/src/core/l10n/components/AppLocalizationProvider.js
```diff
@@ -47,7 +47,9 @@ export class AppLocalizationProviderBase extends React.Component<InternalProps>
     render() {
         const { children, l10n } = this.props;
 
-        return <LocalizationProvider bundles={ l10n.bundles }>
+        const localizations = new ReactLocalization(l10n.bundles);
```
stasm, line 50: My intent in the new API was that `ReactLocalization` instances are stored in the app's state, because I've seen issues with storing just the `bundles` (esp. when it's a generator rather than an array, which doesn't apply in Pontoon's case).

I suggest to make the `core/l10n.actions:get` action return an instance of `ReactLocalization` instead of `bundles`. You won't have to instantiate it in `AppLocalizationProvider.render` then.

adngdb:
All comments should be addressed! :-)

frontend/src/core/l10n/actions.js
```diff
@@ -31,9 +32,9 @@ export function request(): RequestAction {
  */
 export type ReceiveAction = {
     +type: typeof RECEIVE,
-    +bundles: Array<FluentBundle>,
+    +bundles: ReactLocalization,
```
stasm, line 35: Nit: I'd call it `l10n` or `localization`, since `bundles` has a specific meaning inside of it. But it's not very important.

frontend/src/core/l10n/reducer.js
```diff
@@ -14,13 +14,13 @@ type Action =
...
     +fetching: boolean,
-    +bundles: Array<FluentBundle>,
+    +bundles: ReactLocalization,
 |};
 
 
 const initial: L10nState = {
     fetching: false,
-    bundles: [],
+    bundles: new ReactLocalization([]),
```
stasm, line 23: Same nit about naming as above.

adngdb:
All done! @mathjazz wanna merge this whenever you're ready for it?

mathjazz:
Nice work!

BTW, you can still merge yourself, right?

PR #1614, opened 2020-04-10
Fix bug 1625443: Show attribute label in variant labels

mathjazz:
This patch brings support for showing attribute names in select expression variants:
https://bugzilla.mozilla.org/show_bug.cgi?id=1625443

You can test it on stage:
https://mozilla-pontoon-staging.herokuapp.com/de/fluent-test/demo.ftl/?string=205407

Posnetek zaslona 2020-04-10 ob 21 12 23

### DIFF
--- a/frontend/src/modules/fluenteditor/components/RichTranslationForm.js
+++ b/frontend/src/modules/fluenteditor/components/RichTranslationForm.js
@@ -366,6 +366,7 @@ export default class RichTranslationForm extends React.Component<EditorProps> {
         value: string,
         path: MessagePath,
         label: string,
+        attributeLabel: ?string,
         className: ?string,
         example: ?number,
     ) {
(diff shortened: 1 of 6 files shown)


mathjazz:
@stasm Would you mind giving it a look?

frontend/src/modules/fluenteditor/components/RichTranslationForm.js
```diff
@@ -466,14 +482,19 @@ export default class RichTranslationForm extends React.Component<EditorProps> {
             return null;
         }
 
+        let attributeLabel = null;
         if (!label) {
             label = 'Value';
         }
+        else {
+            attributeLabel = label;
```
stasm, line 490: When `label` is truthy, both `label` and `attributeLabel` have the same string value, which I think is a bit misleading. At least it was for me. I don't have a good suggestions about how to fix this, however. Perhaps find a better name for `label`? Or add a comment about its purpose?

frontend/src/modules/fluenteditor/components/RichTranslationForm.js
```diff
@@ -466,14 +482,19 @@ export default class RichTranslationForm extends React.Component<EditorProps> {
             return null;
         }
 
+        let attributeLabel = null;
         if (!label) {
             label = 'Value';
```
stasm, line 487: I think the code could be refactored a bit to make this assignment later on, in `renderElements`.

mathjazz:
Thanks for the review, @stasm!

I've moved the code to set the label to `renderElements()`, simplified it a little and added a short comment. For both, the editor and the original string. Let me know if that makes things any better.

The more serious refactor I'd leave to Adrian, since it's already in progress.

stasm:
Nice, the latest changes definitely made this code easier to follow for me. Thanks!

PR #1616, opened 2020-04-16
Fix bug 1624557: Terminology presentation popup

mathjazz:
This is a continuation of #1608 and a final piece of the [Terminology Presenation](https://github.com/mozilla/pontoon/blob/master/specs/0101-terminology-presentation.md) feature. It's branched off of #1608, on top of which it only adds [one new commit](https://github.com/mozilla/pontoon/commit/cabc4d01c868cb126bd98c651c73db6f34de3025), which implements term highlighting in the source string and the term popup.

**TESTING**
The patch is deployed to stage. I've put the `Usage` column of the [imported Mozilla's terms](https://mozilla-pontoon-staging.herokuapp.com/a/terminology/term/) into a [`terminology.ftl`](https://github.com/mathjazz/pontoon-test/commit/feb8a1521f4fb8fd0b47c05ef475acb5eb01cf05) file, which should help with real-life testing of term matching in the strings. See:
https://mozilla-pontoon-staging.herokuapp.com/fr/pontoon-test-2/terminology.ftl/
https://mozilla-pontoon-staging.herokuapp.com/de/pontoon-test-2/terminology.ftl/

### DIFF
--- a/frontend/src/core/term/withTerms.js
+++ b/frontend/src/core/term/withTerms.js
@@ -1,0 +1,50 @@
+/* @flow */
+
+import * as React from 'react';
+import shortid from 'shortid';
+import escapeRegExp from 'lodash.escaperegexp';
+import { mark } from 'react-content-marker';
+
+import type { TermState } from 'core/term';
+
+
+export function markTerms(base: string, terms: TermState) {
+    if (terms.fetching || !terms.terms) {
+        return base;
+    }
+
+    // Sort terms by length descendingly. That allows us to mark multi-word terms
+    // when they consist of words that are terms as well. See test case for the example.
+    const sortedTerms = terms.terms.sort((a, b) => (a.text.length < b.text.length) ? 1 : -1);
+
+    for (let term of sortedTerms) {
+        const text = escapeRegExp(term.text);
+        const rule = new RegExp(`\\b${text}[a-zA-z]*\\b`, 'gi');
+        const tag = (x: string) =>
+            <mark
+                className='term'
+                data-term={ term.text }
+                key={ shortid.generate() }
+            >{ x }</mark>;
+
+        base = mark(base, rule, tag);
+    }
+
+    return base;
+}
+
+
+type Props = {|
+    terms: TermState,
+|};
+
+
+export default function withTerms<Config: Object>(
+    WrappedComponent: React.AbstractComponent<Config>
+): React.AbstractComponent<Config> {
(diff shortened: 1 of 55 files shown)


frontend/src/core/term/withTerms.js
```diff
@@ -0,0 +1,54 @@
...
+
+
+export function markTerms(base: string, terms: TermState) {
+    if (terms.fetching) {
+        return base;
+    }
+
+    if (!terms.terms) {
+        return base;
+    }
```
adngdb, line 18: Might as well write

```js
    if (terms.fetching || !terms.terms) {
        return base;
    }
```

right? :innocent:

frontend/src/modules/entitydetails/components/TermsPopup.js
```diff
@@ -0,0 +1,49 @@
+/* @flow */
```
adngdb, line 1: Why is this file not in `modules/terms`?

pontoon/terminology/models.py
```diff
@@ -1,31 +1,21 @@
...
-            else:
-                if re.search("".join(terms_words), "".join(source_words), flags):
-                    term_pks.append(term.pk)
-
-        return self.filter(pk__in=term_pks)
+
+            if re.search(term_text, string, flags):
+                terms.append(term)
+
+        return terms
```
adngdb, line 18: What's this? I don't quite understand the previous code, but wasn't it related to tokenizing? Why did you remove it?

frontend/src/modules/entitydetails/components/TermsPopup.js
mathjazz: I've put the files specific for the Terms panel into modules/terms (we use the same logic for the history, machinery, otherlocales and teamcomments panels). Since the popup is not part of the panel, I've put it in modules/entitydetails, where it belongs UI-wise.

frontend/src/core/term/withTerms.js
mathjazz: Good point. Updated along with several other instances across different modules.

pontoon/terminology/models.py
mathjazz: I should have explained it, thanks for asking. :-)

We no longer tokenize strings into words and then search for matches within tokes. Instead we just search for substrings within the string that start with the term.

That greatly simplifies the code (because we don't need a special code path for multi-word terms) and works equally good for within our test file with 200+ strings (the same terms are matched for every string).

PR #1619, opened 2020-04-19
Fix 1627877 Private projects

jotes:
TODO:
* [x] add docs
* [x] add tests
* [x] fix linters


pontoon/base/models.py
```diff
@@ -1031,6 +1031,29 @@ def get_details(parts):
...
+                user, "base.can_manage_locale", accept_global_perms=False
+            ) | get_objects_for_user(
+                user, "base.can_translate_locale", accept_global_perms=False
+            )
+            project_locales = get_objects_for_user(
+                user, "base.can_translate_project_locale", accept_global_perms=False
+            )
+            visibility |= Q(project_locale__locale__in=locales) | Q(
+                project_locale__in=project_locales
+            )
```
mathjazz, line 1054: At this stage we should only distinct between superusers and all other users.

pontoon/administration/templates/admin_project.html
```diff
@@ -286,16 +286,24 @@ <h3>
...
-      <div class="checkbox clearfix">
-        <label for="id_tags_enabled">
+      <label for="id_tags_enabled">
           {{ form.tags_enabled }}{{ form.tags_enabled.label }}
-        </label>
-      </div>
+      </label>
+      {% if tag_formset %}
+      {{ tag_inline_formset.render(title='Linked tags', type='tag', formset=tag_formset, project=project) }}
+      {% endif %}
```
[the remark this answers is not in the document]
jotes: Okay, I'll fix this.

pontoon/administration/templates/admin_project.html
```diff
@@ -286,16 +286,24 @@ <h3>
...
+  </section>
+
+  <section class="section_visibility">
+    <div class="checkbox clearfix id_visibility">
```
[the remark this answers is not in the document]
jotes: Thanks!

jotes:
@mathjazz Hey, I don't know if you had a chance to talk with Adrian about https://github.com/mozilla/pontoon/pull/1562#issuecomment-610210482. Nonetheless, I've fixed a few things and added unit tests for this functionality.

mathjazz:
> @mathjazz Hey, I don't know if you had a chance to talk with Adrian about [#1562 (comment)](https://github.com/mozilla/pontoon/pull/1562#issuecomment-610210482). Nonetheless, I've fixed a few things and added unit tests for this functionality.

Thanks for the update, nice work!

I've opened the PR to update the spec:
https://github.com/mozilla/pontoon/pull/1624

pontoon/base/models.py
```diff
@@ -1215,6 +1253,7 @@ class Project(AggregatedStats):
 
     class Meta:
         permissions = (("can_manage_project", "Can manage project"),)
+        ordering = ("pk",)
```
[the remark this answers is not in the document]
jotes: https://docs.djangoproject.com/en/3.0/ref/models/options/#ordering
```
If a query doesn’t have an ordering specified, results are returned from the database in an unspecified order. A particular ordering is guaranteed only when ordering by a set of fields that uniquely identify each object in the results.
```

pontoon/base/models.py
```diff
@@ -1147,6 +1171,20 @@ class Project(AggregatedStats):
     """,
     )
 
+    VISIBILITY_TYPES = (
+        ("public", "Public"),
+        ("private", "Private"),
+    )
+    visibility = models.CharField(
+        max_length=20,
+        default=VISIBILITY_TYPES[0][0],
```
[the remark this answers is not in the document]
jotes: Okay, I'll fix this.

pontoon/base/admin.py
```diff
@@ -179,6 +179,7 @@ class ProjectAdmin(admin.ModelAdmin):
         "sync_disabled",
         "system_project",
         "pretranslation_enabled",
+        "visibility",
```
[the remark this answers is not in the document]
jotes: Sure, no problem.

pontoon/base/models.py
```diff
@@ -1147,6 +1157,17 @@ class Project(AggregatedStats):
...
+        choices=VISIBILITY_TYPES,
+        help_text="""Project managers can see all projects, the rest of users can view only public projects""",
```
mathjazz, line 1168: Can we make this text future-compatible? It will need to change with the next iteration of the spec implementation, if we keep it like this.

How about something like "Change project visibility between private and public"? If you think that's too vague, I'm also fine with removing the `help_text`.

In any case, you'll also need to update the DB migration.

pontoon/base/utils.py
```diff
@@ -541,3 +541,27 @@ def readonly_exists(projects, locale):
...
+    return (
+        User.objects.filter(pk__in=users)
+        .filter(
+            groups__in=[
+                ProjectLocale.objects.get(
+                    project=project, locale=locale
+                ).translators_group,
+                locale.translators_group,
+                locale.managers_group,
+            ]
```
mathjazz, line 564: I don't understand this filter in the context of bug 1627877.

In fact, I think we should remove this entire function. If the Admin submits a comment in a Private project, that probably means they forgot to make the project accessible to the recipient. Without the check recipient will at least be able to complain about not being able to see the comment/project.

If anything, we should prevent submission of comments on Private projects, but I think that's slightly overengineered at this stage, since only Admins would be able to make such mistake.

mathjazz:
Thanks for the update! Code look good. I'll test once the stats work is done.

Searching for `system_project` in `pontoon.base.models` should hint at some other places that need to be updated for aggregated stats to work properly (i.e. only include public projects).

pontoon/base/templates/widgets/heading_info.html
```diff
@@ -110,7 +110,7 @@
     class='missing',
     value=(obj.total_strings - obj.approved_strings - obj.fuzzy_strings - obj.strings_with_warnings - obj.strings_with_errors),
     link=(link + '?status=missing') if link else None)
-  }}
+  }} {{ obj.total_strings }}
```
jotes, line 113: to remove

pontoon/base/models.py
```diff
@@ -3427,6 +3456,25 @@ def calculate_stats(self, save=True):
...
+                    -self.unreviewed_strings,
+                )
+            else:
+                self.total_strings = 0
+                self.approved_strings = 0
+                self.fuzzy_strings = 0
+                self.strings_with_errors = 0
+                self.strings_with_warnings = 0
+                self.unreviewed_strings = 0
+            return
```
mathjazz, line 3476: I think this is an interesting approach, but it has a few drawbacks:
1. If the changes are made to translations of a private project (e.g. by an admin), stats will be stored again (and not stay at 0).
1. Users who see private projects should see the actual stats, not 0.
1. If you flip the visibility switch for an existing (partially translated) project (which is the use case we'll need as early as next week if we ship the feature in time), stats don't update correctly right now.

Instead I propose we revert the changes to the stats processing you have made in the last two commits and treat private projects similarly as `disabled` and `system_projects`:
1. When the [visibility status changes](https://github.com/mozilla/pontoon/pull/1619/files?file-filters%5B%5D=.css&file-filters%5B%5D=.py#diff-51dc0e413b1d5b6b4a182d2e14ab4909R62), run `locale.aggregate_stats()` for each project-locale.
1. In [`Locale.aggregate_stats()`](https://github.com/mozilla/pontoon/blob/master/pontoon/base/models.py#L1617), only include `TranslatedResources` with `resource__project__visibility="public"`.
1. In [`TranslatedResourceQuerySet.stats()`](https://github.com/mozilla/pontoon/blob/master/pontoon/base/models.py#L3323): if `project.slug == "all-projects"`, only include `TranslatedResources` with `resource__project__visibility="public"`.

pontoon/base/admin.py
```diff
@@ -210,9 +212,10 @@ def enabled(self, obj):
                     "disabled",
```
mathjazz, line 212: This should be moved to the very bottom of the list for consistency with the list view.

pontoon/administration/forms.py
```diff
@@ -61,8 +61,9 @@ def __init__(self, *args, **kwargs):
     def save(self, *args, **kwargs):
         obj = super(ProjectForm, self).save(*args, **kwargs)
         if "disabled" in self.changed_data or "visibility" in self.changed_data:
-            for locale in self.locales.all():
-                locale.aggregate_stats()
+            for project_locale in obj.project_locale.all():
+                project_locale.locale.aggregate_stats()
```
mathjazz, line 66: I don't understand this. Why did you change locale to project_locale? Stats of the project_locale shouldn't change when you change disabled or visibility flags.

pontoon/base/admin.py
```diff
@@ -230,14 +229,14 @@ def enabled(self, obj):
         ExternalProjectResourceInline,
     )
 
-    def save(self, obj):
-        if "visibility" in self.changed_data or "disabled" in self.changed_data:
-            for locale in obj.project_locale.all():
-                locale.aggregate_stats()
+    def save_model(self, request, obj, form, change):
```
mathjazz, line 232: Hmmmmmm... Of course, now that you aren't using Project.save(), you need to repeat the same code twice. :-(

I think that's even worse than using "disabled_changed" in Project.save(), because it will lead to inconsistencies. You've already introduced one by flipping the order of conditions in the if clause. :-)

So I'd restore Project.save() and move all the logic there. That way we'll have everything in one place and the changeset will be smaller.

pontoon/administration/forms.py
jotes: As we discussed IRL: this code executes `aggregate_stats()` on locales assigned to a project.

pontoon/projects/management/commands/send_deadline_notifications.py
```diff
@@ -45,7 +45,12 @@ def handle(self, *args, **options):
             ).distinct()
 
             for contributor in contributors:
-                notify.send(project, recipient=contributor, verb=verb)
+                if (
+                    Project.objects.filter(pk=project.pk)
```
mathjazz, line 49: This will run a DB query for each user... Maybe we can instead tweak the `contributors` QuerySet above with the same trick as here https://github.com/mozilla/pontoon/pull/1619/commits/c864d025b712313f4efc83b5931f31b4d3305450#diff-c18ce63ddb8aca155c0553a0c1b174e3?

jotes: I can optimize this, but I don't think there's an easy way to re-use `visibility_for` in this context. I can create a separate function, similar to this on -> https://github.com/mozilla/pontoon/pull/1619/files/edbd0672ab405c8f64158821eb24c6276323f627#diff-cfad39fede62fdc24a97edc6622b538dR546

pontoon/base/models.py
```diff
@@ -1559,6 +1559,7 @@ class ProjectLocale(AggregatedStats):
 
     class Meta:
         unique_together = ("project", "locale")
+        ordering = ("pk",)
```
mathjazz, line 1562: Does that change require a migration?

pontoon/projects/utils.py
```diff
@@ -0,0 +1,5 @@
+def filter_users_by_project_visibility(project, users):
```
mathjazz, line 1: Can we instead move this to `pontoon.base.models` and use `User.add_to_class`?
https://stackoverflow.com/questions/1642779/django-custom-managers-for-user-model

pontoon/base/models.py
jotes: It does, I've updated the migrations.

pontoon/projects/utils.py
jotes: Done :-)

pontoon/base/migrations/0004_make_old_projects_public_again.py
```diff
@@ -0,0 +1,19 @@
...
+    """Old projects were public by default."""
+
+    Project = apps.get_model("base", "Project")
+    Project.objects.update(visibility="public")
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("base", "0003_auto_20200505_1808"),
```
mathjazz, line 16: I'm a fan of verbose migration names, so I'd suggest to rename migration `0003` to something more meaningful than `auto_XYZ`.

pontoon/base/models.py
```diff
@@ -1031,6 +1043,23 @@ def get_details(parts):
...
+        The visiblity of projects is determined by the role of the user:
+        * Administrators can access all public and private projects
+        * Other users can see only public projects
+        """
+        if user.is_superuser:
+            return self
+
+        return self.public()
+
+    def public(self):
```
mathjazz, line 1057: We don't seem to need this method. I'll leave it to you to decide if we should still keep it.

mathjazz:
You'll also need to rename the migration reference. :)

PR #1622, opened 2020-04-24
Bug 1381959: Add specification for terminology translation

mathjazz:


specs/0102-terminology-translation.md
```diff
@@ -0,0 +1,29 @@
...
+
+- **Terminology creation**: the ability to manually add, remove and review terms.
+- **VCS integration**: the ability to sync terminology with VCS.
+- **Checks**: the ability to check if translations include term translations.
+
+# Feature explanation
+
+We need a new built-in translation project (called "Terminology"), which will allow for translation of all terms using the translation workbench. The project would be similar to system projects (e.g. "Tutorial") in that it would be enabled automatically for any locale and any newly added locale, but would (unlike system projects) remain visible in dashboards and included in team stats.
+
+For every `Term` with the `do_not_translate` and `forbidden` field set to False, an `Entity` will be created in the "Terminology" project with the following data:
```
Pike, line 23: Should `forbidden` be localizable? @gueroJeff?

specs/0102-terminology-translation.md
```diff
@@ -0,0 +1,29 @@
...
+* `Entity.string` with content of `Term.text`.
+* `Entity.comment` with joint content of `Term.part_of_speech`, `Term.definition` and `Term.usage`.
```
Pike, line 25: How is this content joined? Do we have example comments for this?

specs/0102-terminology-translation.md
gueroJeff: Within this scope, `forbidden` is only useful for source content authors. This is one area where n:n mapping comes into play. For now, the entries as a whole should be `forbidden` so that they don't enter the pipeline. We can figure out the mapping in a different iteration.

specs/0102-terminology-translation.md
mathjazz: For example:
```
Part of speech. Definition. E.g.: Usage.
```

I can include that in the spec.

mathjazz:
> How is sync going to work? I'm wondering if the vagueness of "sync" indicates a schema problem in the Terminology?

I'll clarify this in the spec. I think the schema is OK, we'll to reference the Entity though.

PR #1623, opened 2020-04-27
Fix Bug 1611510 - Implement Source String Comments 

abowler2:
Initial work towards implementing source string comments. Replaced the textarea for the comments with a [Slate](https://github.com/ianstormtaylor/slate) editor and added the ability to `@` mention users.

_________________
Work still needed:

- [x] Add ability to `linkify` links
- [x] Add actual users from database for mentions
- [x] Fix Flow errors
- [x] Fix crash when submitting on `enter`
- [x] Find a way for editor to grow when a `\n` is added
- [x] Find out why height limit is not being honored
- [x] Find a way to check if only `\n` added to avoid empty comments
- [x] See if there is a way to remove `!important` from CSS change
- [x] Fix suggestions staying fixed when div scrolls

____
### TODO:

- [ ] Convert existing comments to [new format](https://github.com/mozilla/pontoon/pull/1609#discussion_r406739073) (this may not be needed)
- [ ] Fix or file [notification/focus bug](https://github.com/mozilla/pontoon/pull/1609#discussion_r405893246)
- [ ] Add the "Request context" / "Report issue" link in the metadata section 
- [ ] Add the ability for PMs to "Pin" their responses
- [x] See if there is a way to select a mention with mouse click


frontend/src/core/comments/components/AddComment.css
```diff
@@ -17,7 +17,35 @@
...
+}
+
+.comments-list .add-comment .comment-editor span {
+    vertical-align: baseline !important;
+}
+
+.mention-list {
+    top: -9999px;
+    left: -9999px;
+    position: absolute;
```
mathjazz, line 30: We'll need to set the parent element to relative, otherwise the mention list is fixed on scroll.

abowler2: Since this is inside a `<Portal />`  that is going to take a bit of research as it sits outside of the DOM hierarchy. 

https://reactjs.org/docs/portals.html

abowler2: Also need to work out how reposition it to hover on top when it hits the bottom of the view

Screen Shot 2020-04-27 at 9 53 54 AM

mathjazz: That's definitely another UX glitch - if you have existing translation comments, the popup sometimes opens so low in the screen that you don't even see a pixel of it.

mathjazz: I randomly bumped into this, maybe it'll help:
https://blog.logrocket.com/controlling-tooltips-pop-up-menus-using-compound-components-in-react-ccedc15c7526/

abowler2: I just pushed changes that address both of these issues.  Pretty sure they can be improved but wanted to see if I was even on the right track 😄

frontend/src/core/comments/components/AddComment.js
```diff
@@ -25,69 +41,207 @@ export default function AddComments(props: Props) {
...
+
+        mentionsActive 
+        ? element.style.overflow = 'hidden' 
+        : element.style.overflow = 'auto';
     }
 
-    const submitComment = (event: SyntheticEvent<>) => {
-        event.preventDefault();
-        const comment = commentInput.current.value;
+    const serialize = (node: any) => {
```
abowler2, line 170: I originally tried to define a type within 'types.js' but was unsuccessful. Given the way the data is handled I'm not quite sure how to tackle it. 

**Examples:**
With only a line of text this is what `node` logs:
Screen Shot 2020-05-03 at 6 00 11 PM

However, when there is a line of text and a mention this is what `node` logs:
Screen Shot 2020-05-03 at 6 02 30 PM

frontend/src/core/comments/components/AddComment.js
```diff
@@ -25,69 +41,207 @@ export default function AddComments(props: Props) {
...
+        return ReactDOM.createPortal(children, document.body);
     }
 
-    const handleOnKeyDown = (event: SyntheticKeyboardEvent<>) => {
-        if (event.keyCode === 13 && event.shiftKey === false) {
-            submitComment(event);
+    const disableHistoryScroll = (mentionsActive: boolean) => {
+        const element = document.querySelector(
+            '.history .mentions-disable-scroll'
+        );
```
abowler2, line 159: This only accounts for the current scroll behavior within the `History` panel. 

I was going to set it up with a `ref` on the `history ul` but that caused conflicts with `Team Comments` and before I went too far with that I wanted to be sure this was even a viable solution.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -25,69 +41,212 @@ export default function AddComments(props: Props) {
...
+    );
+
+    const onKeyDown = (event: SyntheticKeyboardEvent<>) => {
+        if (event.key === 'Enter' && !event.shiftKey) {
+            event.preventDefault();
+            submitComment();
+        }
+        if (event.key === 'Enter' && event.shiftKey) { 
+            event.preventDefault();
+            editor.insertText("\n\uFEFF");
```
abowler2, line 118: This allows for the new lines to render while adding comments. The good news is that it might not interfere with how `\n` have been handled in current comments.

To avoid an issue with the cursor placement and an error when navigating with arrows that occurs in Firefox I couldn't have `\n` be the last character so I added the BOM as per the recommendation given in Slate's Slack channel. 

I'm still working on how to strip this when it is the only thing on a line as it still renders a blank line (or an entire blank comment) ATM.

frontend/src/core/comments/components/AddComment.js
mathjazz: You could define type like this:
```diff
-
-    const serialize = (node: any) => {
-        if (Text.isText(node)) {
+    type TextType = {|text: string|};
+    type MentionType = {|type: string, character: string, url: string, children: Array<TextType>|};
+    type InitialType = {|type: string, children: Array<TextType>|};
+    const serialize = (node: TextType | MentionType | InitialType) => {
+        if (Text.isText(node) && node.text) {
             return escapeHtml(node.text);
         }
 
+        if (!node.type || !node.children) {
+            return;
+        }
+
         const children = node.children.map(n => serialize(n)).join('');
 
         switch (node.type) {
             case 'paragraph':
                 return `<p>${children}</p>`;
             case 'link':
-                return `<a href="${escapeHtml(node.url)}">${children}</a>`;
+                if (node.url) {
+                    return `<a href="${escapeHtml(node.url)}">${children}</a>`;
+                }
             case 'mention':
-                return `<a href="${escapeHtml(node.url)}">${children}</a>`;
+                if (node.url) {
+                    return `<a href="${escapeHtml(node.url)}">${children}</a>`;
+                }
             default:
                 return children;
         }
```

frontend/src/core/comments/components/AddComment.js
mathjazz: Nice trick! Please also add (most of) this comment as the code comment.

abowler2:
Rebased with `master`

abowler2:
After spending some time working out how to detect links while comments are added I ran into issues finding the right solution for detecting links that did not include a protocol.

In the end I decided to use [html-react-parser](https://www.npmjs.com/package/html-react-parser) to parse the html into react elements which allowed me to again use `Linkify` when rendering the comments.

I was also able to handle the added `\n` when checking for empty comments as well as remove any trailing `\n` to prevent blank lines being added to the end of a comment.

abowler2:
Just pushed up my most recent work which adds the ability to get user data from the DB for the suggestions, keeps the suggestions from overflowing the edge of the window, and removes the work for adding the `Request context / Report issue` button as it will be worked on as a separate patch.

The final known piece remaining is to fix the positioning of the suggestions when scroll is active.

Also, there are now conflicts with master around the terminology changes. However, I'm not sure how to handle the conflict in `AddComment.js` as I changed the way focus is handled due to `autofocus` not working.  Please advise on how I should handle that.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -22,72 +43,289 @@ export default function AddComments(props: Props) {
...
+            }
+
+            el.style.top = `${setTop}px`;
+            el.style.left = `${setLeft}px`;
+        }
+    }, [chars.length, editor, index, search, target, scrollPosition]);
+
+    React.useEffect(() => {
+       // $FLOW_IGNORE 
+        const handleScroll = (e: SyntheticEvent<HTMLElement>) => {
```
abowler2, line 132: I was not able to figure out what Flow was asking for here.

abowler2:
Mentions are now working 😃 

I left a note above regarding issues with Flow, and I didn't tackle the conflicts because IIRC you were going to do a rebase for those since they involve the Terminology changes (let me know if you want me to go ahead and tackle that though). Otherwise, the tests pass for me locally so this should be ready for a review.

mathjazz:
Thanks for the update, April!

The most important thing in rebasing this patch is to move the condition we currently use on the `autoFocus` param to `ReactEditor.focus(editor);`. Other than that, it seems pretty straightforward (just a little more manual work than usually, because it's a lot of commits).

Let me know if you still want me to do that myself.

abowler2:
>     2. When I'm typing a new team comment the text input grows properly, but the panel doesn't scroll 
>     down fully. Hence, after the field grows, the padding under the input field is cut off.
> 

I can see what is happening, and that is the section `.team-comments` is bigger than the containing div; however, I'm not sure how to tackle this. I'll keep digging but thought I would bring it up in case you have an idea of how to handle it. 

Here are screenshots to illustrate:

### Containing Div

Screen Shot 2020-07-23 at 5 37 37 PM

### Section

Screen Shot 2020-07-23 at 5 37 19 PM

frontend/src/core/api/comment.js
```diff
@@ -25,4 +25,14 @@ export default class CommentAPI extends APIBase {
...
+
+    /**
+     * Get all users from server.
+     */
+    async getUsers(): Promise<Object> {
+        const headers = new Headers();
+        headers.append('X-Requested_With', `XMLHttpRequest`);
+
+        return await this.fetch('get-users', 'GET', null, headers);
+    }
```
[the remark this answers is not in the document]
abowler2: I went back and forth on this and ultimately decided to put it in `comment.js` because it is ALL users as opposed to data specific to an individual user and comments is the only place it is used.  

With that said, it did still feel a bit strange and anti-intuitive to have it there; so, I'll move it to `user.js`

abowler2:
Here are the latest changes.  Of note, when using the mouse to select a mention the original highlight is still present. Initial work to change that was unsuccessful, but I will tackle it again when I implement the additional style changes that are forthcoming.

frontend/src/core/comments/components/Comment.css
```diff
@@ -5,7 +5,11 @@
 .comments-list .comment a {
     color: #7BC876;
     font-weight: 400;
-    white-space: nowrap;
+}
+
+.comments-list .comment a[href^="/contributors/"]:not(.comment-author) {
+    background-color: rgba(51, 57, 65, .4);
+    padding: 0 3px;
```
mathjazz, line 12: Let's use 2px.

And add `border-radius: 2px;`.

frontend/src/core/comments/components/Comment.css
```diff
@@ -5,7 +5,11 @@
 .comments-list .comment a {
     color: #7BC876;
     font-weight: 400;
-    white-space: nowrap;
+}
+
+.comments-list .comment a[href^="/contributors/"]:not(.comment-author) {
```
mathjazz, line 10: This selector will also catch the avatar. Let's add `.container .content` after `.comment`.

frontend/src/core/comments/components/Comment.css
```diff
@@ -5,7 +5,11 @@
...
+    background-color: rgba(51, 57, 65, .4);
```
mathjazz, line 11: What if we just use `#3f4752`? I don't think we need transparency and it will be easier for theming to reuse colors.

mathjazz:
Here are the proposed CSS changes for the mentions popup:
Screenshot 2020-07-29 at 21 43 06

Diff:
```diff
diff --git a/frontend/src/core/comments/components/AddComment.css b/frontend/src/core/comments/components/AddComment.css
index 2f6fbebc4..e068f330d 100644
--- a/frontend/src/core/comments/components/AddComment.css
+++ b/frontend/src/core/comments/components/AddComment.css
@@ -32,18 +32,31 @@
     padding: 10px 12px;
     background-color: #272a2f;
     border: 1px solid #333941;
-    border-radius: 4px;
-    box-shadow: 0 1px 5px rgba(0,0,0,.2);
 }
 
 .mention {
-    padding: 0 4px;
+    padding: 4px 4px 0;
     cursor: pointer;
 }
 
 .mention:hover, 
 .active-mention {
-    background: #aaa;
+    background: #3F4752;
+}
+
+.mention .user-avatar {
+    display: inline-block;
+    vertical-align: middle;
+}
+
+.mention .user-avatar img {
+    border-radius: 2px;
+    border: 1px solid #5E6475;
+    margin-right: 8px;
+}
+
+.mention .name {
+    font-weight: 300;
 }
 
 .mention-element {
diff --git a/frontend/src/core/comments/components/AddComment.js b/frontend/src/core/comments/components/AddComment.js
index 5cb8696ec..ec58347d6 100644
--- a/frontend/src/core/comments/components/AddComment.js
+++ b/frontend/src/core/comments/components/AddComment.js
@@ -335,7 +335,10 @@ export default function AddComments(props: Props) {
                                     className={ i === index ? 'mention active-mention' : 'mention' }
                                     onMouseDown={event => handleMouseDown(event)}
                                 >
-                                    {char}
+                                    <span className="user-avatar">
+                                        <img src="//www.gravatar.com/avatar/0345e1d1fab27d251f6bae37c7088cfc?s=88&d=https%3A%2F%2Fui-avatars.com%2Fapi%2FFrancesco%2520Lodolo%2F88%2F333941%2FFFFFFF" alt="User Avatar" width="22" height="22" />
+                                    </span>
+                                    <span className="name">{ char }</span>
                                 </div>
                             ))}
                         </div>
```

Of course, the user avatar must not be hardcoded. :)

I think you're already aware of it, but in case you are note: if no comments are present in the Team comments section, the mentions list gets cut off at the top of the page. We should probably show it at the bottom instead.

abowler2:
I've added the style changes and fixed up a few things. In doing that I reduced the number of suggestions displayed in order to avoid many of the overflow issues. FWIW, they fit the best when there are only 3, but that felt a bit limiting so I set it to 5.  However, with that setting it can still get cut off at the top of the team comments if the window is smaller (mostly this occurred for me when I had the dev tools open at the bottom)

I did try to set a max-height and then allow all the suggestions to populate with the ability to scroll through the names, but I was not able to get the div to scroll with the use of the arrow keys. Let me know if this is something you would like me to dig into and I can try to see if there is a solution for that.

Of course, if you find any further functional issues just let me know 😊

pontoon/base/views.py
```diff
@@ -596,6 +596,24 @@ def add_comment(request):
     return JsonResponse({"status": True})
 
 
+def get_users(request):
```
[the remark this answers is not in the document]
abowler2: Ah, I had forgotten about this and wanted to discuss options with you. I originally added the AJAX decorator, but it returns `status 400`.

The login decorator does not appear to cause any issues.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -26,72 +46,331 @@ export default function AddComments(props: Props) {
...
+                            {chars.map((char, i) => (
+                                <div
+                                    key={char}
+                                    className={ i === index ? 'mention active-mention' : 'mention' }
+                                    onMouseDown={ handleMouseDown }
+                                >
+                                    <span className='user-avatar'>
+                                        <img 
+                                            src={ getUserGravatar(char) } 
+                                            alt="User Avatar"
```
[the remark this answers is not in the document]
abowler2: When localizing this, I realized that I had not set it up properly for the UserAvatar component and so I fixed that one as well.

pontoon/base/views.py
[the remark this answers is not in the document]
mathjazz: Yeah, [there](https://github.com/mozilla/pontoon/pull/1623/files?file-filters%5B%5D=.css&file-filters%5B%5D=.json&file-filters%5B%5D=.lock&file-filters%5B%5D=.py#diff-54ad6848e541cb011eb00575cfbfbb8cR28) is a `_` instead of `-`.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -26,72 +46,331 @@ export default function AddComments(props: Props) {
...
+                setLeft = rect.right - suggestionsWidth;
+            }
+
+            el.style.top = `${setTop}px`;
+            el.style.left = `${setLeft}px`;
+        }
+    }, [chars.length, editor, index, search, target, scrollPosition]);
+
+    React.useEffect(() => {
+       // $FLOW_IGNORE
```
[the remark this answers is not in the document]
abowler2: I found the comment you linked to previously, and it does eliminate the Flow errors if used as written. However, as I understand it React uses `SyntheticEvents` and this is where the errors come in:
1. `SynthticEvent<>` does not exist on `EventListener`
1. There is no SynthticEventListener
1. Without `EventListener` typed on the event handler then I get errors on all of the listeners re: `Event` being incompatible with `SynthticEvent` 

So, the below code eliminates the Flow errors but uses `Event` in place of the `SyntheticEvent`.  Is that OK, or should I continue to ignore flow and add a comment?
```js
        const handleScroll: EventListener = (e: Event) => {
            if (e.currentTarget instanceof HTMLElement) {
                const element = e.currentTarget;
                setScrollPosition(element.scrollTop);
            }
        }

        const historyScroll: HTMLElement | null = document.querySelector('#history-list');
        const teamsScroll: HTMLElement | null = document.querySelector('#react-tabs-3');
```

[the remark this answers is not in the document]
mathjazz: Thanks for looking into it. I think we should be using the SyntheticEvent wrapper, since it works the same way across browsers. We use it everywhere else.

frontend/src/core/comments/components/AddComment.css
```diff
@@ -1,42 +1,76 @@
...
     line-height: 24px;
+    margin: auto;
     overflow: auto;
     padding: 6px;
-    resize: none;
+    word-break: break-all;
+}
+
+.comments-list .add-comment .comment-editor span {
+    vertical-align: baseline !important;
```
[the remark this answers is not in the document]
abowler2: I can't remember what led me to this solution, but the issue was that the editor would shrink a few pixels once the user began to type and this solved it. I did try at one point earlier to remove it but was still having issues so I put it back in.

However, in taking another look at it I *think* I fixed it without needing this.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -26,72 +46,331 @@ export default function AddComments(props: Props) {
...
+            const rect = domRange.getBoundingClientRect();
+            const teamCommentsEl = document.querySelector('.top');
+            const teamCommentsRect = !teamCommentsEl ? null : teamCommentsEl.getBoundingClientRect();
+            const teamCommentsActive = !teamCommentsEl ? false : teamCommentsEl.contains(document.activeElement);
+            const translateCommentsEl = document.querySelector('.history');
+            const translateCommentsRect = !translateCommentsEl ? null : translateCommentsEl.getBoundingClientRect();
+            const translateCommentsActive = !translateCommentsEl ? false : 
+                translateCommentsEl.contains(document.activeElement);
+
+            let setTop = (rect.top + window.pageYOffset) + 21;
```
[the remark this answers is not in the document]
abowler2: The good news is that I found ways to programmatically set these values. The bad news is there are now quite a few more `const` variables that have been added to that block of code 😁

frontend/src/core/comments/components/AddComment.js
```diff
@@ -26,72 +46,331 @@ export default function AddComments(props: Props) {
...
+        const user = usersList.find(user => user.name === name);
+        if (!user) {
+            return;
+        }
+        return user.gravatar;
+    }, [usersList]);
+    
     if (!user) {
         return null;
     }
```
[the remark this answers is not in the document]
abowler2: I don't believe so, this is the only place the `user` value is used and there is a check in `CommentList` to ensure the user is allowed to leave comments before the `AddComment` is even rendered which would imply that there was a `user`

frontend/src/core/comments/components/AddComment.js
```diff
@@ -1,97 +1,393 @@
...
+        const tabIndexHeight = !tabIndex ? 0 : tabIndex.clientHeight;
+        const commentEditor = document.querySelector('.comments-list .add-comment .comment-editor');
+        const commentEditorLineHeight = parseInt(window.getComputedStyle(commentEditor).lineHeight) || 0;
+        const commentEditorTopPadding = parseInt(window.getComputedStyle(commentEditor).paddingTop) || 0;
+        const commentEditorBottomPadding = parseInt(window.getComputedStyle(commentEditor).paddingBottom) || 0;
+        const commentEditorSpan = document.querySelector('.comments-list .add-comment .comment-editor p span');
+        const commentEditorSpanHeight = !commentEditorSpan ? 0 : commentEditorSpan.offsetHeight;
+        const setTopAdjustment = commentEditorBottomPadding + commentEditorSpanHeight;
+        const suggestionsHeightAdjustment = (
+            commentEditorTopPadding + ((commentEditorLineHeight - commentEditorSpanHeight) / 2));
```
adngdb, line 109: This whole function, and this block in particular, are quite complex. I wished this was handled by an external library, but I assume that might not be a possible option here? 

If so, then for the sake of maintainability (and the sanity of developers, including you, who might need to look back at this code in the future :wink: ), I would highly recommend to take special care to make this as easy to read and understand as possible. That can mean comments, logical blocks of lines, variable names, and probably a combination of all those. If the complexity of the code cannot be lowered, then the number of comments needs to be much higher, as it is easier to read English than to read code.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -1,97 +1,393 @@
...
 
-        const currentRows = Math.trunc(commentInput.current.scrollHeight / textAreaLineHeight);
+        el.style.top = `${ setTop }px`;
+        el.style.left = `${ setLeft }px`;
+    
+    }, [ chars.length, editor, index, search, target, scrollPosition ]);
 
-        if (currentRows < maxRows) {
-            commentInput.current.rows = currentRows;
+    React.useEffect(() => {
```
adngdb, line 149: I think effects should all have a description comment, just like we would do for functions. In a component like this one, where the complexity is very high, it's quite hard to navigate all the hooks without getting lost. Can you please add a description of what each complex hook does in this file?

frontend/src/core/comments/components/AddComment.js
adngdb: I'm pretty sure the number of people who fully understand Flow is lower than the number of fingers on the hand of an arm-less person.

frontend/src/core/comments/components/AddComment.js
```diff
@@ -1,97 +1,393 @@
...
+            if (historyScroll) {
+                historyScroll.removeEventListener('scroll', handleScroll);
+            }
+            if (teamsScroll) {
+                teamsScroll.removeEventListener('scroll', handleScroll);
+            }
+        }
+    }, [])
+
+    const handleMentionsKeyDown = React.useCallback(
```
adngdb, line 182: I don't think this needs to be a callback either. It should be fine as a simple function. It's not too big of a deal to recreate the function on each render, as it depends on many things that are going to trigger a re-render anyway. Also, it's not a dependency of another hook, so it doesn't have to be "callback'ed".

frontend/src/core/comments/components/AddComment.js
```diff
@@ -1,97 +1,393 @@
...
+            * To avoid an issue with the cursor placement and an error when 
+            * navigating with arrows that occurs in Firefox '\n' can't be 
+            * the last character so the BOM was added
+            */
+            editor.insertText('\n\uFEFF');
+            return null;
         }
     }
+    
+    const handleMentionsMouseDown = React.useCallback((event: SyntheticMouseEvent<HTMLDivElement>) => {
```
adngdb, line 236: With the same arguments, I don't think this has to be a callback. (Sorry if I missed something and am wrong about all of those! )

frontend/src/core/comments/components/AddComment.js
```diff
@@ -1,97 +1,393 @@
...
+        event.preventDefault();
+        if (target !== null) {
+            const charIndex = chars.indexOf(event.currentTarget.innerText)
```
adngdb, line 239: Missing semicolon here. :wink:

frontend/src/core/comments/components/AddComment.js
```diff
@@ -1,97 +1,393 @@
...
+            Transforms.select(editor, target);
+            insertMention(editor, chars[charIndex], usersList);
+            return setTarget(null);
+        }
+    }, [ editor, target, chars, usersList ]);
+    
+    const getUserGravatar = React.useCallback((name: string) => {
```
adngdb, line 246: This one on the other hand I think is smart, because it would only change when `usersList` changes, and should never happen more than once (on the first load of the component, with the current implementation, and even less if you move that loading to the `App` component).

frontend/src/core/comments/components/AddComment.js
```diff
@@ -1,97 +1,393 @@
...
+    React.useEffect(() => {
+        getUsers();
+    }, [getUsers])
 
-    if (!user) {
-        return null;
-    }
+    const usersList = users.users;
+    const USERS = usersList.map(user => user.name);
+    const chars = USERS.filter(c =>
```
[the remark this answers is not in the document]
abowler2: I based much of my initial code off of the example given within the Slate editor documentation. This is a carryover from that. 

With that said, it is technically just `chars` as this is what is rendered as the user is typing in the mention name, thus it may only ever be 1-3 chars until they find the person they wanted and make their selection. At that point it is a `mention` 🤷‍♀️ 

Happy to rename it, but not sure what a better name would be given what it is.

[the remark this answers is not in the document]
adngdb: This `chars` variable actually is a list of user objects, based on `USERS` filtered to only contain the top 5 users whose names begin with `search`. Unless I'm misunderstanding, `search` is the variable that contains a list of characters. 

Intuitively I would call it something like `suggestedUsers`, as it contains the list of users to pick from that are going to be shown in the interface.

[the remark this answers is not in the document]
abowler2: And there in lies the danger of responding to reviews late in the day coupled with bad variable naming....Thank you for the friendly response.  😄

abowler2:
I've addressed @adngdb comments and made the needed changes.

adngdb (approved):

mathjazz:
Anyone else having problems building this patch locally?

```
Step 32/35 : RUN yarn build
 ---> Running in 7960c46406e7
yarn run v1.22.4
$ react-scripts build
Creating an optimized production build...
Browserslist: caniuse-lite is outdated. Please run next command `yarn upgrade`
Failed to compile.

./node_modules/slate/dist/index.es.js
Attempted import error: 'createDraft' is not exported from 'immer'.


error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
ERROR: Service 'webapp' failed to build: The command '/bin/sh -c yarn build' returned a non-zero code: 1
make: *** [build] Error 1
Leopold:pontoon mathjazz$
```

It seems like [upgrading](https://github.com/mozilla/pontoon/tree/master/frontend#dependencies) `slate`, `slate-react` and `html-react-parser` to latest versions solves the problem:

- slate: 0.58.4
- slate-react: 0.58.4
- html-react-parser: 0.13.0

adngdb:
It happens to me too :hand:

PR #1624, opened 2020-04-27
0100 Private Projects spec update

mathjazz:
Split private projects into two groups


mathjazz:
@flodolo Just to clear this out: we don't want anyone to have access to the projects while they are being set up (tested), except for PMs, right?

The previous iteration of the spec said Locale Managers and Translators can have access.

flodolo:
Commenting on the resulting document as a whole, since I didn't see the first iteration.

Do we plan to introduce a "Project Manager" role in Pontoon? If not, I think it's quite confusing to use that reference, because it's not available in Pontoon's UI.

Even *superuser* is confusing. Should it be *Admin*, with an explanation of what they are, given that's how it shows up currently in the UI?

mathjazz:
https://github.com/mozilla/pontoon/pull/1624

> Do we plan to introduce a "Project Manager" role in Pontoon? If not, I think it's quite confusing to use that reference, because it's not available in Pontoon's UI.
> 
> Even _superuser_ is confusing. Should it be _Admin_, with an explanation of what they are, given that's how it shows up currently in the UI?

Yes, we'll need to clearly define user roles within Pontoon.

For now I went with Project Manager, because that's what we use at Mozilla, so it's easier to relate to. We also use in the [Documentation specification](https://github.com/mozilla/pontoon/pull/1570/files).

specs/0100-private-projects.md
```diff
@@ -39,9 +42,10 @@ All instances of the "latest activity" feature, throughout the website, should a
 
 | Role | Impact |
 | -- | -- |
-| Locale Manager | Can see all private projects of locales they are managing |
-| Translator | Can see all private projects of locales they are translating |
-| Contributor | Cannot access private projects unless they are given explicit permission |
+| Project Manager | Can see all projects |
+| Locale Manager | Can see all projects visible to "Selected Users" for locales they are managing |
+| Translator | Can see all projects visible to "Selected Users" for locales they are translating |
```
Pike, line 47: I'd think that we would want the opposite to be true, if we used this feature at Mozilla.
We'd use it to restrict projects to contributors under NDA, for example, and most locale managers aren't. Let alone translators.

mathjazz: This changes everything again, then. :-) Is there a use case for this?

We can go back to the simple Private/Public switch if this also works for Wanadev. If it doesn't, we'll need a checkbox to include locale managers and translators or not.

@BThibaut Is it OK for you if Locale Managers and Translators are *not* automatically granted permission to access the private projects and you need to manually select them?

mathjazz:
I've updated the spec which makes visibility for team managers and translators an option, which should allow for more flexibility in how different organizations use private projects.

That actually keeps the spec closer to the version on master. The true difference are the checkboxes to enable projects for team managers and translators.

@Pike Could you r?

specs/0100-private-projects.md
BThibaut: Yes it is ok for us, this will be a little extra step to set up a project but not a regular action.

mathjazz: Thanks! We've now made it easier - there will be a check to include Locale Managers and Translators.

PR #1625, opened 2020-05-08
Add a filter for "Stale" translation

anandthakker:
Adds a new "Stale" search filter, showing translations that have not been updated since before a source string was changed.

Note that this required adding a date_updated field to `Entity`


pontoon/base/models.py
```diff
@@ -2245,6 +2245,29 @@ def unchanged(self, locale, project=None):
...
+        :arg Locale locale: a Locale object to get translations for
+
+        :returns: a django ORM Q object to use as a filter
+
+        """
+        return Q(
+            pk__in=self.get_filtered_entities(
+                locale,
+                Q(
+                    Q(approved=True) & Q(date__lt=F("entity__date_updated"))
```
anandthakker, line 2263: `approved=True` is probably not right here, since that can be filtered separately.  Should this include an `active=True` filter?  I saw that on the `unchanged` query, but wasn't sure what it meant.

anandthakker:
Note: if it feels like the "Stale" filter here is not a general enough use case to be worth including in the filter panel, I'd be happy to resubmit a PR with just 22278ed and d5e90c9, adding the date_updated field to Entity.  That's data that currently isn't being captured, and could have a number of uses going forward

abowler2:
Just a heads up, Prettier formatting has been merged to master, causing conflict on this branch. Please rebase and fix the conflicts as soon as you can, to avoid potentially more conflicts later on. Thank you.

mathjazz:
I'm afraid we should close this PR.

The Stale filter shows translations whose approval happened prior to the most recent update to their corresponding source string. If a stale translation doesn't need to be changed, it cannot be "unstaled" (unless you unapprove and approve translations again, but that’s a really poor UX). That means over time the filter will become pretty useless, catching more and more translations that don't require any action.

PR #1627, opened 2020-05-13
Fix bug 1637660: Implement Terminology translation

mathjazz:
This is Terminology translation implementation:
https://github.com/mozilla/pontoon/blob/master/specs/0102-terminology-translation.md

I've made 3 changes which (if accepted) require an update of the spec:
1. Terms without definitions are not localizable (because we also don't show them in the UI).
1. Terminology Presentation features (Panel and Popup) are disabled in the Terminology project, because all the relevant data is already included in the comments.
1. Whenever relevant Term changes are synced to the Translation project, we also update stats.

TODO:
- [x] Do not focus comment input by default
- [x] Implement sync
- [x] Update spec


frontend/src/modules/entitydetails/components/EntityDetails.js
```diff
@@ -122,7 +122,7 @@ export class EntityDetailsBase extends React.Component<InternalProps, State> {
 
         const source = utils.getOptimizedContent(selectedEntity.machinery_original, selectedEntity.format);
 
-        if (source !== this.props.terms.sourceString) {
+        if (source !== this.props.terms.sourceString && parameters.project !== 'terminology') {
```
Pike, line 125: Is the project name `terminology` a mozilla-ism of Pontoon, or is that hard-coded, and other instances can't have a non-terminology project anyway?

frontend/src/modules/entitydetails/components/TermsPopup.js
```diff
@@ -34,9 +36,11 @@ export function TermsPopup(props: Props) {
 
     return <div className="terms-popup" onClick={ props.hide }>
         <TermsList
-            terms={ terms }
             isReadOnlyEditor={ props.isReadOnlyEditor }
+            locale={ locale }
```
Pike, line 40: This could just be `{ props.locale }` here like in the other parts?

frontend/src/modules/terms/components/Terms.js
```diff
@@ -35,9 +37,11 @@ export default function Terms(props: Props) {
             </Localized>
             :
             <TermsList
-                terms={ terms }
                 isReadOnlyEditor={ isReadOnlyEditor }
+                locale={ locale }
```
Pike, line 41: Also `props` here and below?

pontoon/base/migrations/0003_add_terminology_project.py
```diff
@@ -0,0 +1,131 @@
...
+    term_translations = TermTranslation.objects.filter(term__in=terms)
+
+    Locale = apps.get_model("base", "Locale")
+    locales = Locale.objects.exclude(code__in=["en-US", "en"]).annotate(
+        term_count=Count("terms")
+    )
+
+    # Create the "Terminology" project
+    Project = apps.get_model("base", "Project")
+    project, _ = Project.objects.get_or_create(
```
Pike, line 23: I wonder if this should fail if there's already an existing project with the `terminology` slug.

Also what happens if a different slug had the "Terminology" name?

frontend/src/modules/entitydetails/components/EntityDetails.js
mathjazz: It's hardcoded for all locales. Same as the "Tutorial" project.

pontoon/base/migrations/0003_add_terminology_project.py
mathjazz: Good point, this should fail in case some instance uses the "Terminology" name or "terminology" slug already. That way they will have to (and will be able to) modify their existing projects before upgrading.

frontend/src/core/term/components/Term.js
```diff
@@ -48,6 +58,14 @@ export default function Term(props: Props) {
         <header>
             <span className='text'>{ term.text }</span>
             <span className='part-of-speech'>{ term.partOfSpeech }</span>
+            <a
+                href={ `/${locale}/terminology/common/?string=${term.entityId}` }
+                onClick={ navigateToPath }
+                className='translate'
+            >
```
adngdb, line 65: Wouldn't this work with a simple `<Link>` tag from `react-router`? I'm guessing I'm the one who built this complicated `navigateToPath` thing, but looking at this code I wonder if it's needed.

frontend/src/modules/entitydetails/components/EntityDetails.js
adngdb: Maybe there should be a "reserved words" list somewhere in the documentation to make this explicit to external users?

frontend/src/modules/teamcomments/actions.js
```diff
@@ -47,6 +49,12 @@ export function get(entity: number, locale: string): Function {
 
         let content = await api.entity.getTeamComments(entity, locale);
 
+        // The default return value of aborted requests is {},
+        // which is incompatible with reducer
+        if (isEmpty(content)) {
```
Pike, line 54: Is this a ride-along?

mathjazz: Yes and no. It fixes the JS error when you e.g. quickly move between strings (which aborts requests), but it only occurs now that the Comments panel is visible by default.

frontend/src/modules/entitydetails/components/EntityDetails.js
mathjazz: Obviously, the Admin will complain if you add a project which already exists, but I agree it's helpful to document that.

Maybe we can list internal projects here?
https://mozilla-pontoon.readthedocs.io/en/latest/user/localizing-your-projects.html#adding-a-new-project-to-pontoon

frontend/src/core/term/components/Term.js
mathjazz: I don't think that's possible, because we want to check for unsaved changes.

mathjazz:
I've made the remaining changes to:
- [Sync Terminology project changes to Terminology model](https://github.com/mozilla/pontoon/pull/1627/commits/3f66ac9168739c3be48db30553bc2719d89f9a5b)
- [Sync Terminology model changes to Terminology project](https://github.com/mozilla/pontoon/pull/1627/commits/d2d9dc3c3fb725858991c1162e228f95de74f18c)

I've also made the `Term.entity` field [optional](https://github.com/mozilla/pontoon/pull/1627/commits/fbf34790e69f0fa509d88641fa5ad5db91fade00), because some Terms aren't localizable, i.e. don't have Entity assigned.

mathjazz:
@Pike Let me know if you'd rather see someone else review the [followup](https://github.com/mozilla/pontoon/pull/1627#issuecomment-638973983).

pontoon/base/models.py
```diff
@@ -2585,6 +2585,29 @@ def reset_active_translation(self, locale, plural_form=None):
         else:
             return Translation()
 
+    def reset_term_translation(self, locale):
```
Pike, line 2588: I noticed the loop in the bulk edit, which made me wonder if it's worth to refactor this to have a query of Entities?

Only if this would be a new bottleneck, I guess?

pontoon/terminology/models.py
```diff
@@ -1,7 +1,44 @@
...
+
+
+def update_terminology_project_stats():
+    resource = Resource.objects.get(project__slug="terminology")
+    project = resource.project
+    total_strings = Entity.objects.filter(resource=resource, obsolete=False).count()
+    resource.total_strings = total_strings
+    resource.save(update_fields=["total_strings"])
+
+    translated_resources = TranslatedResource.objects.filter(resource=resource)
```
Pike, line 17: As you get a list anyway, I'd get the list here. Then you don't rely on the query object caching it for you when you use the list in `bulk_update`.

mathjazz:
> I have one question: Say, we want to update the description of a Term. Does that trigger retranslation in all languages for that Term?

No.

As per the [spec](https://github.com/mozilla/pontoon/blob/master/specs/0102-terminology-translation.md#synchronization-between-translation-project-and-data-model):
"If `Term.part_of_speech`, `Term.definition` or `Term.usage` change, `Entity.comment` gets updated". So that's like updating a Fluent message without changing the ID.

As opposed to:
"If `Term.text` changes, a new Entity instance gets created and the previous one becomes obsolete". That triggers retranslation.

Pike:
Ah, cool, found the corresponding code now.

pontoon/terminology/models.py
```diff
@@ -62,6 +101,143 @@ def translation(self, locale):
...
+        entity, created = Entity.objects.get_or_create(
+            string=self.text, comment=self.entity_comment(), resource=resource,
+        )
+
+        # Using update() to avoid circular Term.save() call
+        Term.objects.filter(pk=self.pk).update(entity_id=entity.id)
+
+        if created:
+            entity.order = (
+                Entity.objects.filter(resource=resource).order_by("-order")[0].order
```
jotes, line 152: [question] What happens with `entity.order` when you call create_entity two times in a row? I think it will stay the same for the new entities.

pontoon/base/models.py
```diff
@@ -2585,6 +2585,29 @@ def reset_active_translation(self, locale, plural_form=None):
...
+        function.
+        """
+        approved_translations = self.translation_set.filter(
+            locale=locale, approved=True
+        )
+        term = self.term
+
+        if approved_translations:
+            term_translation, _ = term.translations.get_or_create(locale=locale)
+            term_translation.text = approved_translations[0].string
```
jotes, line 2606: [question] Is there a moment when there's more than 1 approved translation for a given locale? Maybe calling `.get()` could be more explicit here (?).

pontoon/terminology/models.py
mathjazz: Hmmmm, right. That should be +1.

mathjazz:
Updated and made a few additional improvements.

@jotes I'd appreciate another look.

PR #1628, opened 2020-05-13
Bug 1459440: Support Translate Toolkit checks for FTL strings 

anandthakker:
Specifically, this implements the approach suggested in https://bugzilla.mozilla.org/show_bug.cgi?id=1459440#c1


abowler2:
Just a heads up, Prettier formatting has been merged to master, causing conflict on this branch. Please rebase and fix the conflicts as soon as you can, to avoid potentially more conflicts later on. Thank you.

Pike:
Sorry for the lack of response here. I don't think this is the right approach, so I'm going to close this one out. I don't know if what the right approach would be, but also, we should integrate less tightly with tt in general.

I had this as an open question in the bug, but that was mostly me being polite.

PR #1632, opened 2020-05-20
Fix bug 1604038 Use initial letter of display name as unique avatars …

vishnutk:
Gravatar never return a 404 unless we pass them explicitly to send it. Thus changed the code to for it. If we get 404, we call ui-avatars to get image with initials.

Gravatar has default parameter where you pass an image path and it will return you that image if gravatar is not configure for user. This is how we were getting that red panda image. But this works only with static image. If you pass ui-avatar url as default url, we didnt get any image. Probably the reason is, gravatar first pull that image on their server and its not working in ui-avatar case.

Reference : [https://en.gravatar.com/site/implement/images/](https://en.gravatar.com/site/implement/images/)

Some sample urls:

http://www.gravatar.com/avatar/b3beeb5a7ff97a4247b8e92806a708b0?s=88&d=https%3A%2F%2Fpontoon.mozilla.org%2Fstatic%2Fimg%2Fanon.b314dab655e0.jpg

http://www.gravatar.com/avatar/b3beeb5a7ff97a4247b8e92806a708b0?s=88&d=https%3A%2F%2Fui-avatars.com%2Fapi%2F%3Fname%3DJohn%2BDoe

http://www.gravatar.com/avatar/b3beeb5a7ff97a4247b8e92806a708b0?s=88&d=https://eu.ui-avatars.com/api/?uppercase=false&name=different+Case

http://www.gravatar.com/avatar/b3beeb5a7ff97a4247b8e92806a708b0?s=88&d=https://interactive-examples.mdn.mozilla.net/media/examples/grapefruit-slice-332-332.jpg


pontoon/base/models.py
```diff
@@ -109,16 +109,20 @@ def user_profile_url(self):
 
 def user_gravatar_url(self, size):
     email = hashlib.md5(self.email.lower().encode("utf-8")).hexdigest()
-    data = {"s": str(size)}
+    data = {"s": str(size), "d": str(404)}
```
mathjazz, line 112: Instead of `404`, you can set the ui-avatars.com URL directly (see https://eu.ui-avatars.com/):

```python
data = {
    "s": str(size),
    "d": "https://ui-avatars.com/api/{name}/{size}/{background}/{color}".format(
        name=quote(self.display_name),
        size=size,
        background="333941",
        color="FFFFFF",
    ),
}
```

It would also help to use consistent styling.

pontoon/base/models.py
```diff
@@ -109,16 +109,20 @@ def user_profile_url(self):
...
     )
 
+    r = requests.get(local_gravatar_url)
+    if r.status_code == 404:
+        return "https://eu.ui-avatars.com/api/?uppercase=true&name={name}".format(
+            name=self.display_name
+        )
+    else:
+        return local_gravatar_url
+
```
mathjazz, line 125: And then you can remove this entire block.

vishnu-onshape: This is more efficient than my changes. I tried the same previously with below url:

https://eu.ui-avatars.com/api/?uppercase=false&name=different+Case

But it was not working, probably due to query string. But in your case, parameters are part of url path and am able to load the image. 

Will make the changes. Thanks.

vishnutk: This is more efficient than my changes. I tried the same previously with below url:

https://eu.ui-avatars.com/api/?uppercase=false&name=different+Case

But it was not working, probably due to query string. But in your case, parameters are part of url path and am able to load the image.

Will make the changes. Thanks.

pontoon/base/models.py
```diff
@@ -109,13 +108,17 @@ def user_profile_url(self):
...
+        "d": "https://ui-avatars.com/api/{name}/{size}/{background}/{color}".format(
+            name=quote(self.display_name),
+            size=size,
+            background="333941",
+            color="FFFFFF",
+        ),
+    }
 
-    return "//www.gravatar.com/avatar/{email}?{data}".format(
+    return "https://www.gravatar.com/avatar/{email}?{data}".format(
```
mathjazz, line 121: You can remove `https:`, so that `http:` is used on localhost.

mathjazz:
Nice work! Could you please also remove `anon_big.jpg` and `anon.jpg` from the repo? They are no longer needed.

vishnutk:
We can remove `anon_big.jpg`, but `anon.jpg` is referenced in UserAvatar component for anonymous user

https://github.com/mozilla/pontoon/blob/01c66927318adbeeae62a754f4cb25ed0e147cc8/frontend/src/core/user/components/UserAvatar.js#L19

Do we want to change there and show `AU` as a image from ui-avatar?

mathjazz:
> We can remove `anon_big.jpg`, but `anon.jpg` is referenced in UserAvatar component for anonymous user

That's an excellent catch!

I think we can remove both images and also that whole `if (!imageUrl)` block if you do something like this:
```python
diff --git a/pontoon/base/views.py b/pontoon/base/views.py
index b32bb8250..431d3deac 100755
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -421,14 +421,14 @@ def get_translation_history(request):
     payload = []
 
     for t in translations:
-        u = t.user
+        u = t.user or User(id="", username="", first_name="Imported", email="imported")
         translation_dict = t.serialize()
         translation_dict.update(
             {
-                "user": "Imported" if u is None else u.name_or_email,
-                "uid": "" if u is None else u.id,
-                "username": "" if u is None else u.username,
-                "user_gravatar_url_small": "" if u is None else u.gravatar_url(88),
+                "user": u.name_or_email,
+                "uid": u.id,
+                "username": u.username,
+                "user_gravatar_url_small": u.gravatar_url(88),
```

pontoon/base/models.py
```diff
@@ -108,11 +107,17 @@ def user_profile_url(self):
...
-    data = {"s": str(size)}
-
-    if not settings.DEBUG:
-        append = "_big" if size > 88 else ""
-        data["d"] = settings.SITE_URL + static("img/anon" + append + ".jpg")
+    data = {
+        "s": str(size),
+        "d": "https://ui-avatars.com/api/{name}/{size}/{background}/{color}".format(
+            name=quote(
+                self.display_name if self.display_name is not "" else "Imported"
```
vishnutk, line 114: This is the corner case where I logged in with pontoon generated AnonymousUser and it doesnt have first_name, email etc. So just to handle this add if for display_name.

mathjazz: Thanks for checking so deep! AnonymousUser is not supposed to be used as the actual use logging into Pontoon, so we can remove this. Also, if we were to keep this, I think you could simplify the code to: `self.display_name or "Imported"`.

vishnutk: Is every user is goint to have first_name or email, then we are good. Removed the fallback.

PR #1637, opened 2020-05-26
No bug - update the docs about private projects

jotes:
@mathjazz Hey, I'm not sure if that's enough. Should I extend this section and explain the visibility settings in depth?

### DIFF
--- a/docs/user/localizing-your-projects.rst
+++ b/docs/user/localizing-your-projects.rst
@@ -74,9 +74,9 @@ following required fields:
    strings. You can monitor the progress in the Sync log (``/sync/log/``).
 1. When the synchronization is finished, you should check the imported resources
     and the entities. If everything went okay, you can proceed to the next step.
-1. Go to the project's admin page and change the visibility option to make the project public.
-   It's required because all new projects in Pontoon are private by default and aren't visible to
-   localizers and locale managers.
+1. Go to the project's admin page and change the visibility option to make
+   the project public. It's required because all new projects in Pontoon are private
+   by default and aren't visible to localizers and locale managers.
 
 For complete documentation of the Admin form, please refer to Mozilla's
 `new project documentation`_.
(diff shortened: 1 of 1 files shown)


docs/user/localizing-your-projects.rst
```diff
@@ -69,9 +69,14 @@ following required fields:
 5. **Download prefix**: used for manually downloading files. Select any
    localizable file on GitHub, click ``Raw`` and replace locale code and the
    following bits in the URL with {locale_code}.
-6. Click **SAVE PROJECT** at the bottom of the page.
-7. After the page reloads, click **SYNC** and wait for Pontoon to import
-   strings. You can monitor the progrss in the Sync log (``/sync/log/``).
+7. Click **SAVE PROJECT** at the bottom of the page.
```
mathjazz, line 72: That's actually 6, not 7?

docs/user/localizing-your-projects.rst
```diff
@@ -69,9 +69,14 @@ following required fields:
...
+8. All new projects in Pontoon are private by default and aren't visible to
+   localizers and locale managers.
```
mathjazz, line 74: This is not an actual step, so I would instead append it to the last step.

docs/user/localizing-your-projects.rst
adngdb: It's doesn't matter much, as when processed, those numbers are not used (they are only cosmetic, for when you read the source file). You could have all items start with `1.`, the end result will be a correctly numbered list.

mathjazz: Yeah, converting all to 1. is even better.

docs/user/localizing-your-projects.rst
jotes: I moved it to the last step, can you check if the phrase still sounds okay?

PR #1639, opened 2020-05-27
Fix bug 1635743 [API] Exclude system projects from default

vishnutk:
Introduces includeSystem parameter which is by default false. Tested with below graphql query on local box and getting expected result.

```
query {
  projects(includeDisabled: true, includeSystem: false) {
    name
  }
  locale(code:"mr") {
    name
    googleTranslateCode
    
    localizations(includeSystem: true) {
      project {
        name
      }
    }
  }
}
```
@mathjazz r?

### DIFF
--- a/pontoon/api/schema.py
+++ b/pontoon/api/schema.py
@@ -112,14 +112,14 @@ class Locale(DjangoObjectType, Stats):
     def resolve_localizations(obj, _info, include_disabled, include_system):
         qs = obj.project_locale
 
-        if include_disabled and include_system:
-            return qs.all()
-        elif include_disabled and not include_system:
-            return qs.filter(project__system_project=False)
-        elif not include_disabled and include_system:
-            return qs.filter(project__disabled=False)
+        records = qs.filter(project__disabled=False, project__system_project=False)
 
-        return qs.filter(project__disabled=False, project__system_project=False)
+        if include_disabled:
+            records = records | qs.filter(project__disabled=True)
+        if include_system:
+            records = records | qs.filter(project__system_project=True)
+
+        return records.distinct()
 
 
 class Query(graphene.ObjectType):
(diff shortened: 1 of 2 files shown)


pontoon/api/schema.py
```diff
@@ -100,29 +100,40 @@ class Meta:
...
-        if include_disabled:
+        if include_disabled and include_system:
             return qs.all()
+        elif include_disabled and not include_system:
+            return qs.filter(project__system_project=False)
+        elif not include_disabled and include_system:
+            return qs.filter(project__disabled=False)
 
-        return qs.filter(project__disabled=False)
+        return qs.filter(project__disabled=False, project__system_project=False)
```
mathjazz, line 118: I think you can simplify this entire block to:
```python
return qs.filter(
    project__disabled=include_disabled,
    project__system_project=include_system,
)
```

pontoon/api/schema.py
```diff
@@ -132,10 +143,14 @@ def resolve_projects(obj, info, include_disabled):
...
-        if include_disabled:
+        if include_disabled and include_system:
             return qs.all()
+        elif include_disabled and not include_system:
+            return qs.filter(system_project=False)
+        elif not include_disabled and include_system:
+            return qs.filter(disabled=False)
 
-        return qs.filter(disabled=False)
+        return qs.filter(disabled=False, system_project=False)
```
mathjazz, line 153: I think you can simplify this entire block to:
```python
return qs.filter(disabled=include_disabled, system_project=include_system)
```

pontoon/api/schema.py
vishnutk: This is not giving correct result. Reason is QuerySet filter does AND on the condition. As per my understanding we need below result, correct?

// disabled = true and system = true - should return all the project
// disabled = true and system = false - should return all the project except system
// disabled = false and system = true - should return all the project except disabled
// disabled = false and system = false - should return all the project except disabled and system

I am not able to write complex query which will give this result with .filter/.exlude. Thus its better to keep this block.

mathjazz: That's a great point!

I'm afraid the current solution doesn't scale very well. Imagine we'll add one or more conditions we'd like to exclude by default.

How about something like this?

```python
qs = obj.project_locale

records = qs.filter(disabled=False, system_project=False)

if include_disabled:
    records = records | qs.filter(project__disabled=True)
if include_system:
    records = records | qs.filter(project__system_project=True)

return records.distinct()
```

PR #1640, opened 2020-06-02
No bug - Fix the lists in the "Localizing your projects" document.

jotes:
@mathjazz r?


Pike:
Looking at https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html?#lists-and-quote-like-blocks, the RST way is to use `#. `. Go  figure.

jotes:
@Pike Thanks, I found some piece of documentation that didn't include this version of numbered lists and I incorrectly assumed that's not supported.

PR #1643, opened 2020-06-08
Fix bug 1624408 - Increase the Gunicorn's worker timeout config option.

jotes:
@mathjazz a smalll patch to check if the hypothesis is correct.


mathjazz:
I've deployed the patch to stage. I didn't find any regressions in the (very) quick test I performed.

Interestingly, download of the test file (and some other I've tried) completes way ahead of the timeout (even the old one). Apparently it's much faster on stage compared to prod. What would be the reason for that?

Given that increasing the timeout doesn't scale too good, I wonder if the short-term bugfix is to document the use of the `--compressed` flag while using `curl`?

jotes:
> Apparently it's much faster on stage compared to prod.
Without a performance report, it's hard to say. My wild guess is: the network traffic is smaller and the CPU load on the dynos differs.

PR #1647, opened 2020-06-16
Fix bug 1561663 Admins should be able to remove an account

vishnutk:
@mathjazz r+

### DIFF
--- a/pontoon/actionlog/models.py
+++ b/pontoon/actionlog/models.py
@@ -1,14 +1,8 @@
 from __future__ import absolute_import, unicode_literals
 
 from django.core.exceptions import ValidationError
-from django.contrib.auth import get_user_model
+from pontoon.base.utils import get_sentinel_user
 from django.db import models
-
-
-def get_sentinel_user():
-    return get_user_model().objects.get_or_create(
-        username="pontoon-deleted-user", email="pontoon-deleted-user@mozilla.com"
-    )[0]
 
 
 class ActionLog(models.Model):
--- a/pontoon/actionlog/models.py
+++ b/pontoon/actionlog/models.py
@@ -1,7 +1,14 @@
 from __future__ import absolute_import, unicode_literals
 
 from django.core.exceptions import ValidationError
+from django.contrib.auth import get_user_model
 from django.db import models
+
+
+def get_sentinel_user():
+    return get_user_model().objects.get_or_create(
+        username="pontoon-deleted-user", email="pontoon-deleted-user@mozilla.com"
+    )[0]
 
 
 class ActionLog(models.Model):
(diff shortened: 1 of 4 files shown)


vishnutk:
Once admin deletes user from admin console, below are the references gets set to `pontoon-deleted-user`. I was not willing to put email to this user, but front end code use email in comments and other places. Also gravatar broke since our assumption that user will always have email failed :-) Thus added the email which gets displayed for the deleted user in front end.

ActionLog - performed_by
PermissionChangelog - performed_by, performed_on
Project - contact
Translation - user, approved_user, unapproved_user, rejected_user, unrejected_user
Comment - author
Term - created_by

pontoon/base/models.py
```diff
@@ -64,6 +65,12 @@
 UNUSABLE_SEARCH_CHAR = "☠"
 
 
+def get_sentinel_user():
+    return get_user_model().objects.get_or_create(
+        username="pontoon-deleted-user", email="pontoon-deleted-user@mozilla.com"
```
mathjazz, line 70: We should actually create a [unique](https://bugzilla.mozilla.org/show_bug.cgi?id=1561663#c5) user for each deleted user, so you'll need to create a unique `username` and `email`, e.g. `username=deleted-user-{hash}` and `email=deleted-user-{hash}@mozilla.com`.

Sorry for not making that requirement clearer.

We should also set `first_name` to `Deleted User` to make attributions more readable.

mathjazz: Note that we should use the example.com domain name for the email:
https://bugzilla.mozilla.org/show_bug.cgi?id=1646209

pontoon/actionlog/models.py
```diff
@@ -1,9 +1,16 @@
 from __future__ import absolute_import, unicode_literals
 
 from django.core.exceptions import ValidationError
+from django.contrib.auth import get_user_model
 from django.db import models
 
 
+def get_sentinel_user():
```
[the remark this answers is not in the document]
vishnutk: I should have followed DRY :-(. Done now and moved it to utils

pontoon/base/models.py
vishnutk: Updated code with all except hash comment.

pontoon/base/models.py
```diff
@@ -63,7 +63,6 @@
 
 UNUSABLE_SEARCH_CHAR = "☠"
 
-
```
mathjazz, line 66: Nit: you should restore this line. We keep two blank lines between imports and initializations, and functions/classes.

vishnutk: Restored.

mathjazz:
Thanks! Please fix `black` (you can simply run `make black`) and `flake8`.

vishnutk:
With new changes in requirement, this PR is no more relevent. Thus closing it.

PR #1650, opened 2020-06-22
Fix bug 1638341 - Prevent editor focus  after user-data request.

VishalCR7:


frontend/src/modules/genericeditor/components/GenericTranslationForm.js
```diff
@@ -59,7 +61,12 @@ export default class GenericTranslationForm extends React.Component<EditorProps>
             this.props.resetSelectionContent();
         }
 
-        this.focusInput(editor.changeSource !== 'internal');
+        const prevPropsWithoutUser = (({ user, ...rem }) => rem)(prevProps)
+        const propsWithoutUser = (({ user, ...rem }) => rem)(this.props)
+
+        if(!isEqual(prevPropsWithoutUser, propsWithoutUser)){
```
mathjazz, line 67: I guess we can simplify this by comparing all props (without excluding the `user`).

VishalCR7: If there is a change in user-data isEqual(prevPropsWithoutUser, propsWithoutUser) will return false, and the editor gets focused.
As discussed, yesterday, we don't want to focus the editor even when props.user changes. Isn't this the case?

mathjazz: Good point. Let's keep it that way.

The code to exclude user from props is pretty hard to understand. I wonder if we can make it more straightforward. Maybe we can use `delete`?

PR #1651, opened 2020-06-25
 Fix bug 1402679 - Add machinery_sources to translation. 

VishalCR7:
Added `machinerySources` to editor state which gets set while copying translation from machinery.


mathjazz:
Thanks for the update!

I've noticed one issue that didn't existing before - FTL translations aren't marked as coming from Machinery anymore.

frontend/src/core/editor/components/connectedEditor.js
```diff
@@ -250,21 +250,27 @@ export default function connectedEditor<Object>(
...
-        sendTranslation = (ignoreWarnings?: boolean, translation?: string) => {
+        sendTranslation = (ignoreWarnings?: boolean, translation?: string, simpleTranslation?: Translation) => {
             const props = this.props;
 
             if (props.editor.isRunningRequest || !props.selectedEntity || !props.locale) {
                 return;
             }
 
             const content = translation || props.editor.translation;
+            const simpleContent = simpleTranslation || props.editor.translation;
```
mathjazz, line 261: Not sure `simpleTranslation` is a good name - it's not simple when it comes from the RichEditor. Maybe `editorTranslation`? And editorContent instead of `simpleContent`?

PR #1652, opened 2020-06-26
Fix bug 1561663 Admins should be able to remove an account

vishnutk:
If we didn't modify the ForeignKey constraint on the Comment, Action Log etc, django admin ui shows these entities on the confirmation page. Thus marked it as Null.

### DIFF
--- a/pontoon/actionlog/models.py
+++ b/pontoon/actionlog/models.py
@@ -25,7 +25,7 @@ class ActionLog(models.Model):
     action_type = models.CharField(max_length=50, choices=ACTIONS_TYPES)
     created_at = models.DateTimeField(auto_now_add=True)
     performed_by = models.ForeignKey(
-        "auth.User", models.CASCADE, related_name="actions"
+        "auth.User", models.SET_NULL, related_name="actions", null=True
     )
 
     # Used to track on what translation related actions apply.
--- a/pontoon/actionlog/models.py
+++ b/pontoon/actionlog/models.py
@@ -25,7 +25,7 @@ class ActionLog(models.Model):
     action_type = models.CharField(max_length=50, choices=ACTIONS_TYPES)
     created_at = models.DateTimeField(auto_now_add=True)
     performed_by = models.ForeignKey(
-        "auth.User", models.SET_NULL, related_name="actions", null=True
+        "auth.User", models.CASCADE, related_name="actions"
     )
 
     # Used to track on what translation related actions apply.
--- a/pontoon/base/admin.py
+++ b/pontoon/base/admin.py
@@ -1,5 +1,7 @@
 from __future__ import absolute_import
 
+import uuid
+
 from django.contrib import admin
 from django.contrib.auth import get_user_model
 from django.contrib.auth.admin import (
@@ -11,8 +13,10 @@ from django.forms.models import ModelForm
 from django.forms import ChoiceField
 from django.urls import reverse
 
+from pontoon.actionlog.models import ActionLog
 from pontoon.base import models
 from pontoon.base import utils
+from pontoon.terminology.models import Term
 
 from pontoon.teams.utils import log_user_groups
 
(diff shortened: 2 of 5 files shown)


pontoon/base/admin.py
```diff
@@ -55,6 +59,44 @@ def save_model(self, request, obj, form, change):
 
             log_user_groups(request.user, obj, (add_groups, remove_groups))
 
+    def delete_model(self, request, obj):
+        random_hash = uuid.uuid4().hex
+        new_user = User.objects.create_user(
+            username="deleted-user-" + random_hash,
+            email="deleted-user-" + random_hash + "@example.com",
+            first_name="Deleted User",
```
mathjazz, line 67: Let's also set [`is_active`](https://docs.djangoproject.com/en/3.0/ref/contrib/auth/#django.contrib.auth.models.User.is_active) flag to `False`. That way it will be easier to filter deleted users.

pontoon/base/admin.py
```diff
@@ -55,6 +59,44 @@ def save_model(self, request, obj, form, change):
 
             log_user_groups(request.user, obj, (add_groups, remove_groups))
 
+    def delete_model(self, request, obj):
```
mathjazz, line 62: It would be nice to add a comment here explaining what we're doing here and why (creating a new user to replace the deleted user in order to change the ID).

pontoon/actionlog/models.py
```diff
@@ -25,7 +25,7 @@ class ActionLog(models.Model):
     action_type = models.CharField(max_length=50, choices=ACTIONS_TYPES)
     created_at = models.DateTimeField(auto_now_add=True)
     performed_by = models.ForeignKey(
-        "auth.User", models.CASCADE, related_name="actions"
+        "auth.User", models.SET_NULL, related_name="actions", null=True
```
mathjazz, line 28: That shouldn't be necessary. If Django still tries to delete related objects (same for Comments and Terms), the custom delete code probably doesn't execute. Maybe you need to add an [action](https://docs.djangoproject.com/en/3.0/ref/contrib/admin/actions/#adding-actions-to-the-modeladmin)?

vishnutk:
current confirmation page
New_confirmation_page

Attached the current and new screenshot for changes display

pontoon/base/admin.py
vishnutk: Added it

pontoon/base/admin.py
```diff
@@ -55,6 +62,94 @@ def save_model(self, request, obj, form, change):
 
             log_user_groups(request.user, obj, (add_groups, remove_groups))
 
+    # This method overrite the default delete process. We are creating new user with random hash
+    # and updating the cascade relationships on the objects with this new user. Then actual user gets
+    # deleted. By this way we persist translations, comments etc for deleted user.
+    # Bug 1561663 Admin should be able to remove an account
```
mathjazz, line 68: Let's rephrase this a little bit:
```
# Before deleting the user, we need to make sure data we care about is kept,
# especially translations the user authored. Hence we create a new anonymized user,
# move user data we care about to it and then delete the user.
# See bug 1561663 for details.
```

pontoon/base/admin.py
```diff
@@ -55,6 +62,94 @@ def save_model(self, request, obj, form, change):
...
+        )
+        models.Translation.objects.filter(unrejected_user=obj).update(
+            unrejected_user=new_user
+        )
+        Term.objects.filter(created_by=obj).update(created_by=new_user)
+        models.Comment.objects.filter(author=obj).update(author=new_user)
+
+        super(UserAdmin, self).delete_model(request, obj)
+
+    # This method is to overrite bulk delete method from the user list page
```
mathjazz, line 104: Grammar: overri**d**e.

pontoon/base/admin.py
vishnutk: Modified

PR #1656, opened 2020-07-07
Fix bug 1651105: Add specification for concordance search

mathjazz:


mathjazz:
@MikkCZ I wonder what your thoughts are.

mathjazz:
> Can you detail a bit more on how this will interact with MT machinery, and TM matches?

Same as the existing "Search in Machinery". The default Machinery suggestions will be replaced by the concordance search results after you hit Enter.

> I just tried the "Search in Machinery", and found it confusing. Also seems hard to get back to the initial state? So I wouldn't be tied to the status quo.

I'm curious, what specifically did you find confusing?

To get back to the original state, you need to empty the search box and press Enter. I understand that's not obvious. I'm not sure it needs to be very obvious (we could add a Clear button for example).

I realize we have an inconsistent behavior (and similar looks) between the String list search (searches instantly) and the Machinery search (searches after you press Enter). We should either add a button to the Machinery search, or (my favorite) require Enter to be pressed in the String list as well.

> But to some extent, this search sounds like an extension to the normal results? Replacing the current kinda-filtering behavior?

Would it be less confusing if we moved the search functionality to a new tab and only kept the default Machinery behaviour in the Machinery tab?

MikkCZ:
If I understand it correctly, the proposed behaviour is to keep the current in-translation Machinery to work as is (approximate string search for the whole string in TM), until you start typing some text into the search field. When you type text there, the results from TM are fetched using fulltext-like search instead. Is that correct?

If so, I would also suggest to search TM not just against the original English strings (e.g. _file_), but also in the localized strings (e.g. _soubor_). It would also be nice to have quote to work the same way as they do in the translation view search (exact match).

One general note, the situation in bug I reported originally got a little bit better since we have the terminology feature available. However it didn't help when you need to search anything more complex than a single term, or a term that is not in the terminology yet.

mathjazz:
> I realize we have an inconsistent behavior (and similar looks) between the String list search (searches instantly) and the Machinery search (searches after you press Enter). We should either add a button to the Machinery search, or (my favorite) require Enter to be pressed in the String list as well.

Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1659491 and PRed https://github.com/mozilla/pontoon/pull/1670.

mathjazz:
OK, time to finally move forward here. :-)

I've added two things:
- support for searching for multi-word phrases (within quotes).
- a button for quick access to the initial Machinery state.

Searching within translations (in addition to source strings) was already included.

@MikkCZ @Pike Could you have another look please?

julen:
>> I just tried the "Search in Machinery", and found it confusing. Also seems hard to get back to the initial state? So I wouldn't be tied to the status quo.
>
> I'm curious, what specifically did you find confusing?
>
> To get back to the original state, you need to empty the search box and press Enter. I understand that's not obvious. I'm not sure if it needs to be very obvious (we could add a Clear button for example).

I always found this confusing too, in two ways:
1. There's no way to get back to the initial state: emptying the search box and pressing Enter doesn't currently work unfortunately. Being able to revert to the initial set of results will be a good improvement as outlined in your latest changes 👍 
2. If a term is searched, the state of the input element persists when navigating away to another translation unit; however, the Machinery results for the new unit are displayed. So the term that was previously searched should be cleared from the input when navigating away.

mathjazz:
Thank you for your feedback @julen!

I've filed two bugs:
* [Unable to get back to the initial Machinery state after search](https://bugzilla.mozilla.org/show_bug.cgi?id=1676750) (seems FTL specific)
* [Clear Search in Machinery field after switching strings](https://bugzilla.mozilla.org/show_bug.cgi?id=1676751)

specs/0106-concordance-search.md
```diff
@@ -0,0 +1,25 @@
...
+- Created: 2020-07-07
+- Associated Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1514698
+
+# Summary
+
+Add ability to search existing translations while translating or reviewing strings.
+
+# Motivation
+
+Having the ability to search for existing translations is a crucial tool for assuring translation consistency. It allows you to see how was the same expression translated in the past or verify that the translation you intend to use is consistent with the rest of the corpora. Pontoon tries to automate that as much as possible with Machinery suggestions by querying Translation Memory as soon as a string is opened for translation.
```
Pike, line 11: _corpora_ is plural, I think within this document, you just want _corpus_.

specs/0106-concordance-search.md
```diff
@@ -0,0 +1,25 @@
...
+
+# Feature explanation
+
+This feature changes the way "Search in Machinery" searches Translation Memory. Instead of searching for fuzzy matches of whole strings stored in Translation Memory, `source` and `target` fields of the `TranslationMemoryEntry` model are searched for each keyword and each phrase within quotes separately.
+
+For inspiration, see `Entity.for_project_locale()` in `pontoon.base.models`, which is used in string list search. Note that only `source` and `target` fields are used for searching (unlike in `Entity.for_project_locale()`, which also searches comments).
+
+Placeholder text of the search box is changed from "Search in Machinery" to "Concordance Search". When results for Concordance search are displayed, the magnifier icon in the search box changes from fa-search to fa-times. Clicking on it reverts to the default Machinery results. In output, project name is used as the title instead of "Translation Memory", unless `TranslationMemoryEntry.project` field is None.
```
Pike, line 23: Is there a less linguistic term to use instead of _concordance search_ ? Maybe just _search_?

specs/0106-concordance-search.md
```diff
@@ -0,0 +1,25 @@
...
+
+Search behaviour of other sources (e.g. Machine Translation) remains unchanged.
```
Pike, line 25: I wouldn't mind just searching TM, and to drop MT. Or to leave the MT results untouched, if that's easy enough.

Pike (commented):
I think I had an epiphany just now. You're not adding another search method, you're ripping out the old one, too? I've been banging my head against the wall of how the two would work together.

One thing that I'd also change is the click behavior, it shouldn't replace the input field, I'd say. Maybe just write to the clipboard via https://developer.mozilla.org/en-US/docs/Web/API/Clipboard ?

specs/0106-concordance-search.md
mathjazz: I'd say let's ask @guerojeff, but he'll most likely prefer "Concordance Search". :-)

While Concordance Search is more accurate, I'm afraid many users are not familiar with the term. So I'm fine with either. Or Search Translation Memory?

specs/0106-concordance-search.md
mathjazz: Untouched compared to the default Machinery behaviour?

mathjazz:
> I think I had an epiphany just now. You're not adding another search method, you're ripping out the old one, too?

Correct. :-)

> One thing that I'd also change is the click behavior, it shouldn't replace the input field, I'd say. Maybe just write to the clipboard via https://developer.mozilla.org/en-US/docs/Web/API/Clipboard ?

That's a good point. Or maybe just add to the editor, like with the [Terms](https://pontoon.mozilla.org/de/firefox/all-resources/?string=214061)?

specs/0106-concordance-search.md
Pike: Yeah, keeping the MT results for the original string, and search matches for TM. But also confusing, I guess.

mathjazz: I'd find that confusing, too.

SDL for example includes MT as an option:
https://docs.sdl.com/791656/39128/sdl-worldserver-11-0/concordance-search

MikkCZ (approved):
Thanks you for the ping. I have no new comments or thoughts other than what was already said or addressed here.

specs/0106-concordance-search.md
```diff
@@ -0,0 +1,25 @@
...
+
+That being said, Translation Memory is only queried for strings as a whole, even if you "Search in Machinery". If you're interested in the translation of a word "file", but your corpora only uses that word as part of longer strings like "Select a file..." or "Source file uploaded", Machinery won't find any matches.
+
+Alternativelly, you can switch to the All Projects view and use the string list search, which searches for substrings within existing translations for each word separatelly, but that also comes with a few drawbacks. You either lose translation progress in the editor and the state of the string list, or you need to use a separate tab. You also won't be able to search for translations in the projects that aren't active anymore.
+
+# Feature explanation
+
+This feature changes the way "Search in Machinery" searches Translation Memory. Instead of searching for fuzzy matches of whole strings stored in Translation Memory, `source` and `target` fields of the `TranslationMemoryEntry` model are searched for each keyword and each phrase within quotes separately.
```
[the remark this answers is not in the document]
mathjazz: That's a good question. As outlined below, we can base the code on `Entity.for_project_locale()`, which orders results by project name, resource name and finally order. I don't think that's the best order here.

We could either use Levenshtein distance - i.e. show "Bookmarks" above "Bookmark this page" when searching for "bookmark" (added to the spec) - or show most recent translations first.

mathjazz:
To quote Axel on CMO: "no further comments on the spec from my side, thanks".

Thanks everyone for a very valuable feedback.

Let's merge!

specs/0106-concordance-search.md
```diff
@@ -0,0 +1,27 @@
...
+# Motivation
+
+Having the ability to search for existing translations is a crucial tool for assuring translation consistency. It allows you to see how was the same expression translated in the past or verify that the translation you intend to use is consistent with the rest of the corpus. Pontoon tries to automate that as much as possible with Machinery suggestions by querying Translation Memory as soon as a string is opened for translation.
+
+That being said, Translation Memory is only queried for strings as a whole, even if you "Search in Machinery". If you're interested in the translation of a word "file", but your corpus only uses that word as part of longer strings like "Select a file..." or "Source file uploaded", Machinery won't find any matches.
+
+Alternativelly, you can switch to the All Projects view and use the string list search, which searches for substrings within existing translations for each word separatelly, but that also comes with a few drawbacks. You either lose translation progress in the editor and the state of the string list, or you need to use a separate tab. You also won't be able to search for translations in the projects that aren't active anymore.
```
flodolo, line 15: nit: a few typos here (adverbs ending in -ely, like alternatively and separately, only have on l).

specs/0106-concordance-search.md
```diff
@@ -0,0 +1,27 @@
...
+
+# Feature explanation
+
+This feature changes the way "Search in Machinery" searches Translation Memory. Instead of searching for fuzzy matches of whole strings stored in Translation Memory, `source` and `target` fields of the `TranslationMemoryEntry` model are searched for each keyword and each phrase within quotes separately.
```
flodolo, line 19: I'm completely lost on this explanation: can you provide an example? To quote the example above, would you need to search for `"file"`? Would `"select" "file"` work?

specs/0106-concordance-search.md
```diff
@@ -0,0 +1,27 @@
...
+
+For inspiration, see `Entity.for_project_locale()` in `pontoon.base.models`, which is used in string list search. Note that only `source` and `target` fields are used for searching (unlike in `Entity.for_project_locale()`, which also searches comments). Results are sorted by Levenshtein distance in descending order.
```
flodolo, line 21: For what it's worth, I'm not sure it helps to have code references in specs. There should be a description of the behavior accessible to non developers, or developers not familiar with Pontoon's codebase.

specs/0106-concordance-search.md
```diff
@@ -0,0 +1,27 @@
...
+
+Two changes are made to the search box. Placeholder text is changed from "Search in Machinery" to "Search Translation Memory" and the magnifier icon is changed from fa-search to fa-times when results for Concordance search are displayed. Clicking on it reverts to the default Machinery results.
```
flodolo, line 23: Similar note here, it would be great to see what `fa-search` and `fa-times` actually looks like.

specs/0106-concordance-search.md
```diff
@@ -0,0 +1,27 @@
...
+
+In output, project name is used as the title instead of "Translation Memory", unless `TranslationMemoryEntry.project` field is None. Clicking on each result adds the text to the editor (like when inserting Terms) instead of replacing it (like when using default Machinery results).
```
flodolo, line 25: What happens if there are matches in multiple projects?

specs/0106-concordance-search.md
mathjazz: Thanks! I'll fix that.

specs/0106-concordance-search.md
mathjazz: Check out the next paragraph. In essence, search will work the same as in the string list, except that we'll only search `source` and `target` fields. I'll try to come up with a clearer explanation.

specs/0106-concordance-search.md
mathjazz: I don't think code references should be *the only* way to explain a feature, but I'm curious to learn why do you think code references aren't helpful in general.

Developers are a target audience of the specification. So in cases like this, where we have an existing code that can be reused, it's beneficial to call it out. Especially for developers not familiar with Pontoon's codebase.

specs/0106-concordance-search.md
mathjazz: Good idea. Icon references should be links, e.g.:
https://fontawesome.com/v4.7.0/icon/search
https://fontawesome.com/v4.7.0/icon/times

specs/0106-concordance-search.md
mathjazz: Same as in the string list - it's a separate entry.

specs/0106-concordance-search.md
flodolo: Because code references can become obsolete, while a description of the feature doesn't? But yeah, if there are both, I guess it doesn't hurt to have specific code references, and they serve different targets.

PR #1658, opened 2020-07-13
Fix bug 1604843: Add support for SYSTRAN Translate service

mathjazz:
The patch is deployed to [stage](https://mozilla-pontoon-staging.herokuapp.com/de/firefox/all-resources/). Systran is only enabled for German, French and Spanish (es and es-ES).

Followups:
- [ ] Refactor Machinery so we don't need to add a new engine in so many different places
- [ ] Showing original string in Machine Translation services results


Pike:
I glanced at the results on stage, and there's one where I don't know why, https://mozilla-pontoon-staging.herokuapp.com/de/firefox/all-resources/?string=192281. There's a leading `<meta http-equiv="Content-Type" content="text/html; charset=UTF-8">` ?

mathjazz:
Addressed review comments in https://github.com/mozilla/pontoon/pull/1658/commits/74b25d6a1f6a16e6a57f0f72c8c4491b603c2d89 and https://github.com/mozilla/pontoon/pull/1658/commits/649277e3c25d9253ba149c2ea7dbe20402e100da.

PR #1661, opened 2020-08-03
Fix bug 1638239: Do not highlight terms within placeables

mathjazz:
This patch reverses the order in which we mark placeables and terms using [react-content-marker](https://github.com/mozilla/react-content-marker/) to prevent marking terms within placeables ([bug 1638239](https://bugzilla.mozilla.org/show_bug.cgi?id=1638239)).

It also drops Higher-Order Marker in favor of appending the list of Term rules rules to the list of Placeable rules.


frontend/src/core/term/withTermsAndPlaceables.js
```diff
@@ -0,0 +1,52 @@
...
+import {
+    getRulesWithFluent,
+    getRulesWithoutLeadingSpace,
+    rules,
+} from 'core/placeable';
+
+import type { TermState } from 'core/term';
+
+
+export default function withTermsAndPlaceables(terms: TermState, fluent: ?boolean) {
```
adngdb, line 17: I'd rename `fluent` into `usingFluent` or `isFluent`. Just "fluent" is a bit confusing.

adngdb: And I would rename this function as well. The word "with" made sense when it was an HOC, but now that it returns a Marker, I doesn't work as good. Maybe call it something like `getMarkerWithTermsAndPlaceables`? Maybe just `getMarker`, as it is in the `terms` namespace already.

PR #1670, opened 2020-08-17
Fix bug 1659491: Trigger string list search with Enter

mathjazz:
Deployed to stage for user feedback:

### DIFF
--- a/frontend/src/modules/search/components/SearchBox.js
+++ b/frontend/src/modules/search/components/SearchBox.js
@@ -4,8 +4,6 @@ import * as React from 'react';
 import { connect } from 'react-redux';
 import isEqual from 'lodash.isequal';
 
-import { Localized } from '@fluent/react';
-
 import './SearchBox.css';
 
 import * as navigation from 'core/navigation';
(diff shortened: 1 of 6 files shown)


frontend/src/modules/search/components/SearchBox.js
```diff
@@ -330,12 +327,15 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
...
         this.setState({
             search: event.currentTarget.value,
-        }, () => this.updateSearchParams());
+        });
     }
 
-    updateSearchParams = debounce(() => {
-        this.update();
-    }, 500)
+    performSearch = (event: SyntheticKeyboardEvent<HTMLInputElement>) => {
```
adngdb, line 333: This is a bit confusing as it doesn't, strictly speaking, performs the search. I'd call it "handleKeyDown" and add a comment that says it performs the search on Enter.

frontend/src/modules/search/components/SearchBox.js
```diff
@@ -450,6 +450,7 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
                 onBlur={ this.unsetFocus }
                 onChange={ this.updateSearchInput }
                 onFocus={ this.setFocus }
+                onKeyDown={ this.performSearch }
```
adngdb, line 453: The reason for my previous is that here, on reading this line, I think that you perform the search on each key stroke.

frontend/src/modules/search/components/SearchBox.js
```diff
@@ -330,12 +327,15 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
...
+        });
     }
 
-    updateSearchParams = debounce(() => {
-        this.update();
-    }, 500)
+    performSearch = (event: SyntheticKeyboardEvent<HTMLInputElement>) => {
+        // On Enter
+        if (event.keyCode === 13) {
+            this.update()
```
adngdb, line 336: Missing semicolon. :innocent:

adngdb (approved):

mathjazz:
We've got positive feedback on [Discourse](https://discourse.mozilla.org/t/proposal-to-change-search-behavior-in-the-string-list/65901) and chat. Let's move on!

PR #1672, opened 2020-08-25
Nobug - Remove unused UserProjectsManager to fix the problem with the…

jotes:
This should fix the issue with the missing migration. @mathjazz r?


mathjazz:
Remove unused UserProjectsManager.

PR #1675, opened 2020-08-29
Fix bug 1658420: Enable downloads of Terminology in TBX 2008 (v2) format

mathjazz:
We're adding the ability to download Terminology in the legacy TBX 2.0 format (URL-only).

### DIFF
--- a/pontoon/terminology/utils.py
+++ b/pontoon/terminology/utils.py
@@ -13,7 +13,7 @@ def build_tbx_v2_file(term_translations, locale):
     yield (
         u'<?xml version="1.0" encoding="UTF-8"?>'
         u'\n<!DOCTYPE martif SYSTEM "TBXcoreStructV02.dtd">'
-        u'\n<martif type="TBX" xml:lang="en">'
+        u'\n<martif type="TBX" xml:lang="en-US">'
         u"\n\t<martifHeader>"
         u"\n\t\t<fileDesc>"
         u"\n\t\t\t<titleStmt>"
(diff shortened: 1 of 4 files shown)


pontoon/terminology/views.py
```diff
@@ -41,13 +41,30 @@ def get_terms(request):
 
 
 @condition(etag_func=None)
-def download_terminology(request, locale):
+def download_terminology_tbx_v2(request, locale):
```
jotes, line 44: [nitpick] A class-based view can reduce the amount of the duplicated code

pontoon/terminology/utils.py
```diff
@@ -3,8 +3,76 @@
 from xml.sax.saxutils import escape, quoteattr
 
 
-def build_terminology_file(term_translations, locale):
+def build_tbx_v2_file(term_translations, locale):
     """
+    Generates contents of the TBX 2008 (v2) file (TBX-Default dialect):
```
jotes, line 8: Do you know about command-line linters for the TBX dialects?

jotes (approved):
I see only one nitpick related to the DRY, but the rest looks good.

pontoon/terminology/utils.py
mathjazz: There might be some here:
https://www.tbxinfo.net/tbx-downloads/
https://www.tbxinfo.net/tbx-downloads-2/

mathjazz:
Thanks, I'd appreciate another look on changes related to the class-based view.

jotes (approved):

PR #1676, opened 2020-08-31
Fix Bug 1611510 - Insert PM into team comment via button

abowler2:
This is my initial work for implementing the ability to use the `Request context / Report issue` button to auto-insert a mention for the project manager into the team comments.

I've been able to get as far as successfully inserting the mention when the button is pressed, but my approach has presented a couple of issues that I haven't been able to work out.  I'm opening this PR to get some ideas and thoughts from @mathjazz and/or @adngdb. 

### The main issue:
The only success I have had in inserting the mention and still being able to continue typing in the editor is to utilize the `useEffect()` hook. However, after the mention for the project manager is inserted via the button it will continue to be inserted on every instance of Team Comments. So, for example, if you change entities and then open the Team Comments it will be inserted, if you switch to the Terms tab and then come back it will be inserted. 

#### The two main approaches I tried - (I tried many variations on these approaches, and there could be a correct path in here, but I wasn't  successful):

- setting a Ref on `AddComment` and forwarding that `ref` back to `EntityDetails` in an effort to trigger a function in `AddComment` when the button is clicked  (ie.... `editorRef.current.insertProjectManager()`) - I was not able to get this working because `current` was always `null`
- Since I can't insert the mention prior to setting the initial value I tried to call an async function that would set the initial value and then insert the mention if there was a `project manager`. - This produced errors regarding not being able to use `map` on a promise within the serialize function (that is the error that prevents the mention from being inserted prior to the initial value as well only it states that it can't map undefined in that case).

The other issue is that if the Team Comments section is open when the button is pressed the mention is not inserted unless you switch out of the panel and come back again (of course, it may not appear after coming back if the above issue is fixed).

If you have suggestions on where to go from here I would really appreciate it. 😊

### DIFF
--- a/frontend/src/modules/entitydetails/components/ContextIssueButton.css
+++ b/frontend/src/modules/entitydetails/components/ContextIssueButton.css
@@ -1,7 +1,3 @@
-.metadata .source-string-comment {
-    flex: 1 0 auto;
-}
-
 .metadata .source-string-comment .context-issue-button {
     background: #333941;
     border: 1px solid #3f4752;
@@ -21,17 +17,4 @@
 
 .metadata .source-string-comment .context-issue-button {
     color: #aaa;
-}
-
-@media (max-width: 780px) {
-    .metadata .source-string-comment {
-        flex: 0 1 auto;
-    }
-}
-
-@media (max-width: 640px) {
-    .metadata .source-string-comment {
-        flex: 1 0 auto;
-        margin-bottom: 8px;
-    }
 }
--- a/frontend/src/modules/entitydetails/components/ContextIssueButton.css
+++ b/frontend/src/modules/entitydetails/components/ContextIssueButton.css
@@ -22,3 +22,16 @@
 .metadata .source-string-comment .context-issue-button {
     color: #aaa;
 }
+
+@media (max-width: 780px) {
+    .metadata .source-string-comment {
+        flex: 0 1 auto;
+    }
+}
+
+@media (max-width: 640px) {
+    .metadata .source-string-comment {
+        flex: 1 0 auto;
+        margin-bottom: 8px;
+    }
+}
--- a/frontend/src/modules/entitydetails/components/ContextIssueButton.css
+++ b/frontend/src/modules/entitydetails/components/ContextIssueButton.css
@@ -11,7 +11,7 @@
     font-size: 11px;
     font-style: normal;
     font-weight: 100;
-    margin: -1px -1px 0 4px; /* Compensate for invisible border in default state */
+    margin: -1px -1px -1px 4px; /* Compensate for invisible border in default state */
     padding: 2px 4px;
 }
 
--- a/frontend/src/modules/entitydetails/components/ContextIssueButton.css
+++ b/frontend/src/modules/entitydetails/components/ContextIssueButton.css
@@ -1,0 +1,24 @@
+.metadata .source-string-comment {
+    flex: 1 0 auto;
+}
+
+.metadata .source-string-comment .context-issue-button {
+    background: #333941;
+    border: 1px solid #3f4752;
+    border-radius: 4px;
+    color: #ccc;
+    float: right;
+    font-size: 11px;
+    font-style: normal;
+    font-weight: 100;
+    margin: -1px -1px 0 4px; /* Compensate for invisible border in default state */
+    padding: 2px 4px;
+}
+
+.metadata .source-string-comment .context-issue-button:hover {
+    border-color: #637283;
+}
+
+.metadata .source-string-comment .context-issue-button {
+    color: #aaa;
+}
--- a/frontend/src/modules/entitydetails/components/ContextIssueButton.css
+++ b/frontend/src/modules/entitydetails/components/ContextIssueButton.css
@@ -14,7 +14,3 @@
 .metadata .source-string-comment .context-issue-button:hover {
     border-color: #637283;
 }
-
-.metadata .source-string-comment .context-issue-button {
-    color: #aaa;
-}
--- a/frontend/src/modules/entitydetails/components/ContextIssueButton.js
+++ b/frontend/src/modules/entitydetails/components/ContextIssueButton.js
@@ -1,0 +1,25 @@
+/* @flow */
+
+import * as React from 'react';
+import { Localized } from '@fluent/react';
+
+import './ContextIssueButton.css';
+
+type Props = {|
+    openTeamComments: () => void,
+|};
+
+export default function ContextIssueButton(props: Props) {
+    return (
+        <div className='source-string-comment'>
+            <Localized id='entitydetails-Metadata--context-issue-button'>
+                <button
+                    className='context-issue-button'
+                    onClick={props.openTeamComments}
+                >
+                    {'REQUEST CONTEXT or REPORT ISSUE'}
+                </button>
+            </Localized>
+        </div>
+    );
+}
--- a/frontend/src/modules/entitydetails/components/ContextIssueButton.js
+++ b/frontend/src/modules/entitydetails/components/ContextIssueButton.js
@@ -10,12 +10,23 @@ type Props = {|
 |};
 
 export default function ContextIssueButton(props: Props) {
+    const [isDisabled, setIsDisabled] = React.useState(false);
+
+    const handleClick = () => {
+        setIsDisabled(true);
+        props.openTeamComments();
+
+        setTimeout(() => {
+            setIsDisabled(false);
+        }, 3000);
+    };
     return (
         <div className='source-string-comment'>
             <Localized id='entitydetails-Metadata--context-issue-button'>
                 <button
                     className='context-issue-button'
-                    onClick={props.openTeamComments}
+                    onClick={handleClick}
+                    disabled={isDisabled}
                 >
                     {'REQUEST CONTEXT or REPORT ISSUE'}
                 </button>
--- a/frontend/src/modules/entitydetails/components/ContextIssueButton.js
+++ b/frontend/src/modules/entitydetails/components/ContextIssueButton.js
@@ -22,7 +22,7 @@ export default function ContextIssueButton(props: Props) {
     };
     return (
         <div className='source-string-comment'>
-            <Localized id='entitydetails-Metadata--context-issue-button'>
+            <Localized id='entitydetails-ContextIssueButton--context-issue-button'>
                 <button
                     className='context-issue-button'
                     onClick={handleClick}
--- a/frontend/src/modules/entitydetails/components/ContextIssueButton.js
+++ b/frontend/src/modules/entitydetails/components/ContextIssueButton.js
@@ -10,23 +10,12 @@ type Props = {|
 |};
 
 export default function ContextIssueButton(props: Props) {
-    const [isDisabled, setIsDisabled] = React.useState(false);
-
-    const handleClick = () => {
-        setIsDisabled(true);
-        props.openTeamComments();
-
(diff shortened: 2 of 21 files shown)


pontoon/base/models.py
```diff
@@ -360,6 +360,17 @@ def serialized_notifications(self):
     }
 
 
+def user_serialize(self):
+    """ Serialize Project contact """
```
abowler2, line 364: I set this up to send all the needed data for a mention, but ATM I'm only using the `name`

abowler2:
Rebased with master to fix merge conflicts and incorporate prettier changes.

frontend/src/modules/entitydetails/components/EntityDetails.js
```diff
@@ -78,6 +80,16 @@ type State = {|
...
 export class EntityDetailsBase extends React.Component<InternalProps, State> {
+    tabRef: { current: Object };
+    constructor(props: InternalProps, state: State) {
+        super(props);
+        this.state = {
+            ...state,
+            tabIndex: 0,
+            projectManager: '',
+        };
+        this.tabRef = React.createRef();
```
adngdb, line 91: This should probably be called `commentTabRef`, to be more precise.

adngdb (commented):
Hey @abowler2, sorry I took so long to take a look at this. And thanks for reminding me (through Matjaz :smile: ) to do it! 

Here's an example of how I solved a similar problem: https://github.com/adngdb/pontoon/blob/686f11bc5b853344ac52cf0c0d54364fb2296062/frontend/src/core/editor/hooks/useReplaceSelectionContent.js

I put the value in a state. When that value changes, I have an effect that triggers the desired change in the other component, and then reset the state to an empty value. This way, the effect happens only once per user action, instead of happening every time a dependency of the effect changes.

I believe this pattern can solve your issue here. Store the `projectManager` in a state (which you already do), and pass down a function to reset that value along with the variable. In your effect, verify that the value is not null / not empty, and reset it after you've applied the desired changes.

abowler2:
> I believe this pattern can solve your issue here. Store the `projectManager` in a state (which you already do), and pass down a function to reset that value along with the variable. In your effect, verify that the value is not null / not empty, and reset it after you've applied the desired changes.

Hey @adngdb  - so this appears to have worked 🎉  Thank you again for your help.

abowler2:
@mathjazz - I've implemented Adrian's suggestion and it looks to have solved the issues I was having. This should now be ready for a review.

mathjazz (commented):
Very well done!

I left a few notes, mostly suggestions for cosmetic code improvements.

I'd style the button differently as demonstrated in one of the comments.

Two other things I notices:
1. We don't add space after the mention of the contact person, which is annoying.
1. If the Comments tab is open before you click on the "Request..." button, comment field doesn't get a focus.

abowler2 (commented):
I was able to address adding a space after the mention and making sure the focus/cursor is placed properly even when the tab is already opened.

I've also addressed all but one of the remaining comments, that one just needs your thoughts and will be a quick fix.

frontend/src/modules/entitydetails/components/EntityDetails.js
```diff
@@ -70,6 +70,8 @@ type InternalProps = {|
 
 type State = {|
     translation: string,
+    tabIndex: number,
```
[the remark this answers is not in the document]
mathjazz: Good call!

mathjazz (commented):
Great work!

I have a few minor stylistic suggestions, which I find easier to communicate with a diff:

```css
diff --git a/frontend/src/modules/entitydetails/components/Metadata.css b/frontend/src/modules/entitydetails/components/Metadata.css
index 5c98b0d97..a77366a9d 100644
--- a/frontend/src/modules/entitydetails/components/Metadata.css
+++ b/frontend/src/modules/entitydetails/components/Metadata.css
@@ -46,6 +46,7 @@
     font-size: 14px;
     font-style: normal;
     line-height: 22px;
+    margin-top: -2px; /* Align with the source string comment button */
     padding-bottom: 6px;
     text-align: start;
     white-space: pre-wrap;
@@ -78,18 +79,19 @@
     padding: 0 0 0 2px;
 }

-.metadata .source-string-comment {
-    float: right;
+.metadata .source-string-comment .context-button {
     background: #333941;
     border: 1px solid #3f4752;
     border-radius: 4px;
+    color: #ccc;
+    float: right;
+    font-size: 11px;
+    font-style: normal;
+    font-weight: 100;
+    margin: -1px -1px 0 2px; /* Compensate for invisible border in default state */
     padding: 2px 4px;
 }

-.metadata .source-string-comment:hover {
+.metadata .source-string-comment .context-button:hover {
     border-color: #637283;
 }
-
-.metadata .source-string-comment .context-button {
-    color: #aaa;
-}
```

Please also make sure that the Request / Report button doesn't push the Rich Fluent Editor downwards.

abowler2:
@mathjazz I've made the requested changes and for the Rich Fluent Editor I was able to utilize Flexbox to keep it from slipping under the button.

Let me know if you need any further changes.

frontend/src/modules/entitydetails/components/Metadata.css
```diff
@@ -76,3 +85,28 @@
...
+.metadata .source-string-comment .context-issue-button {
+    background: #333941;
+    border: 1px solid #3f4752;
+    border-radius: 4px;
+    color: #ccc;
+    float: right;
+    font-size: 11px;
+    font-style: normal;
+    font-weight: 100;
+    margin: -1px -1px 0 2px; /* Compensate for invisible border in default state */
```
mathjazz, line 102: Sorry, could you please replace 2px with 4px to add some more spacing between text and the button?

mathjazz (commented):
Nice work!

The Rich Editor works very well now. But it seems that the regular source string is now narrower - every line breaks into a new one.

Posnetek zaslona 2020-10-23 ob 10 04 45

I also just noticed another issue I haven't seen before - if you click the button N times, the mention also gets inserted N times. We should only insert it once.

abowler2:
> The Rich Editor works very well now. But it seems that the regular source string is now narrower - every line breaks into a new one.

Yes, but only if the button is present. If there is no button then the text displays across the whole div. This was the compromise with using Flexbox to account for the Rich Editor.  The other option I tried was to set a `max-width` on the Table itself, but this did not hold up under smaller screens and pushed the table below the button again. 

If you have another thought on how to approach this I'd be happy to give it a try.

mathjazz:
> If you have another thought on how to approach this I'd be happy to give it a try.

The only other option I can think of is to use different CSS for different editors. Before this latest changeset, the generic source string wrapped as expected.

abowler2:
I've restructured a bit in order to be able to style the Rich Editor separately, so now the table displays correctly and the generic source strings wrap as expected. I don't think this broke the styling, but happy to make changes if they are needed.

I also addressed the issue of multiple insertions if the context button is pressed multiple times. For this I chose to disable the button via a timeout since the Metadata section does not re-render with changes in the comments. If you prefer a different approach, want a different duration, or want the button styled to reflect that it is disabled just let me know.

frontend/src/modules/entitydetails/components/ContextIssueButton.js
```diff
@@ -0,0 +1,36 @@
...
+        setIsDisabled(true);
+        props.openTeamComments();
+
+        setTimeout(() => {
+            setIsDisabled(false);
+        }, 3000);
+    };
+    return (
+        <div className='source-string-comment'>
+            <Localized id='entitydetails-Metadata--context-issue-button'>
```
mathjazz, line 25: s/Metadata/ContextIssueButton

frontend/src/modules/fluentoriginal/components/FluentOriginalString.js
```diff
@@ -8,15 +8,19 @@ import RichString from './RichString';
 import SimpleString from './SimpleString';
 import SourceString from './SourceString';
 
+import ContextIssueButton from 'modules/entitydetails/components/ContextIssueButton';
```
mathjazz, line 11: "Outside of the module, always import from the module's index, and never from specific files. This allows us to easily evolve modules and keep things decoupled, which reduces code complexity."

https://github.com/mozilla/pontoon/tree/master/frontend#modules

frontend/src/modules/entitydetails/components/ContextIssueButton.css
```diff
@@ -0,0 +1,24 @@
...
+.metadata .source-string-comment .context-issue-button {
+    background: #333941;
+    border: 1px solid #3f4752;
+    border-radius: 4px;
+    color: #ccc;
+    float: right;
+    font-size: 11px;
+    font-style: normal;
+    font-weight: 100;
+    margin: -1px -1px 0 4px; /* Compensate for invisible border in default state */
```
mathjazz, line 14: I noticed that `-1px -1px -1px 4px` allows the 2nd line of the source string to occupy full width.

frontend/src/modules/entitydetails/components/ContextIssueButton.js
```diff
@@ -0,0 +1,36 @@
...
+export default function ContextIssueButton(props: Props) {
+    const [isDisabled, setIsDisabled] = React.useState(false);
+
+    const handleClick = () => {
+        setIsDisabled(true);
+        props.openTeamComments();
+
+        setTimeout(() => {
+            setIsDisabled(false);
+        }, 3000);
```
mathjazz, line 21: I don't think that's a reliable solution. What if I click after 3 seconds? Also, if I move to the next string quickly, the button remains blocked. We should instead check if the mention already exists and not insert it again.

mathjazz (commented):
Thanks for the update!

The generic source string now works as expected. In the Rich Editor however the button overflows the longer string:

Screenshot 2020-10-24 at 23 20 45

abowler2:
Latest attempt to fix the remaining concerns.  Let me know what changes you would like 😊

mathjazz (commented):
Thanks for the update and addressing my comments. 👍 

I don't think the media queries fix the issue in the Rich Editor. We don't need to make the button fit into smaller screens - the entire Translate page is not really responsive anyways and requires wider screens.

The issue is that the text doesn't wrap around the button like in the Generic source string:

Screenshot 2020-10-25 at 22 50 24

abowler2:
OK, I switched from flexbox to grid for the Rich Editor and it works better at keeping the button and table lined up, without overlapping, even on smaller screens. However, given that the Rich Editor is in a `table` it does not wrap the subsequent lines across the full div.

The only other thought I had was to try incorporating the button into the first row of the table, but I'm not sure that is possible or wise. 

If you have further ideas or can point me in the direction of another option I can look into that would be appreciated.

mathjazz (approved):
Thanks for the update!

That grid solution is also not ideal - it wraps the button in two lines on a laptop screen (1440px). Positioning of the button in the Rich Editor is a hard problem to fix and with that editor not being used so often, I suggest we go ahead with the original solution that pushes the rich editor a line downwards. I'll make those changes. They will also make the code much simpler.

**Thanks for the great work in this patch and for all the work you've put into implementing translation, team and source string comments, as well as pinned comments and mentions! Thanks to your effort, collaboration and communication features of Pontoon have come a long way - from basically not existing to a level unmatched by basically any other TMS. Excellent job!**

mathjazz:
@abowler2 Could you please review the behavioral and stylistic changes in https://github.com/mozilla/pontoon/pull/1676/commits/2a738851739febf750292a0c416ab9b4156c3a7b?

abowler2:
LGTM and works as expected.

PR #1678, opened 2020-08-31
 Fix Bug 1598313 change checkbox colour for missing strings

anujpandey785:
Three Changes Made to implement new colour for missing strings
1)Entity.css
Added- .entity.missing .status:before with new colour ie #5F7285
2)ResourceProgress.css
Changed-.progress-chart .menu .details div.missing border colour to #5F7285
3)style.css
Added- .missing .status.fa:before with new colour ie #5F7285

### DIFF
--- a/frontend/src/modules/entitieslist/components/Entity.css
+++ b/frontend/src/modules/entitieslist/components/Entity.css
@@ -95,6 +95,11 @@
     color: #FED271;
 }
 
+.entity.missing .status:before {
+    color: #5F7285;
+}
+
+
 .entity.errors .status:before {
     color: #F36;
 }
--- a/frontend/src/modules/entitieslist/components/Entity.css
+++ b/frontend/src/modules/entitieslist/components/Entity.css
@@ -75,7 +75,7 @@
 }
 
 .entity .status:before {
-    color: #4D5967;
+    color: #5F7285;
     content: "";
 }
 
@@ -95,24 +95,10 @@
     color: #FED271;
 }
 
-.entity.missing .status:before {
-    color: #5F7285;
-}
-
-
 .entity.errors .status:before {
     color: #F36;
 }
 
 .entity.warnings .status:before {
     color: #FFA10F;
-}
-
-.entity.missing.checked .status:before,
-.entity.missing.selected .status:before,
-.entity.missing:hover .status:before,
-.entity.partial.checked .status:before,
-.entity.partial.selected .status:before,
-.entity.partial:hover .status:before {
-    color: #3F4752;
 }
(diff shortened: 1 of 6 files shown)


frontend/src/modules/entitieslist/components/Entity.css
```diff
@@ -95,6 +95,11 @@
     color: #FED271;
 }
 
+.entity.missing .status:before {
+    color: #5F7285;
+}
+
+
```
mathjazz, line 102: We shouldn't need to add new rules. Instead, you can just modify the value of the `.entity .status:before {` rule.

Please also remove the very last rule at the bottom of the file. It's no longer needed, because the new "missing" color works well even on lighter background used when the entity is selected or hovered.
```css
.entity.missing.checked .status:before,
.entity.missing.selected .status:before,
.entity.missing:hover .status:before,
.entity.partial.checked .status:before,
.entity.partial.selected .status:before,
.entity.partial:hover .status:before {
    color: #3F4752;
}
```

pontoon/base/static/css/style.css
```diff
@@ -21,6 +21,10 @@
   content: "";
 }
 
+.missing .status.fa:before {
+  color:#5F7285;
+}
+
```
mathjazz, line 27: Same as in `Entity.css`: no need to add a new rule, just change the color on `.status.fa:before {` to `#5F7285`.

Please also search for `.missing` within the `style.css` file and change background and border values to `5F7285`. I'm not exactly sure these rules are still being used, but let's at least keep the values consistent.

frontend/src/modules/resourceprogress/components/ResourceProgress.css
```diff
@@ -122,7 +122,7 @@
 }
 
 .progress-chart .menu .details div.missing {
-    border-color: #4D5967;
+    border-color:#5F7285;
```
mathjazz, line 125: Mind the space after `:`.

mathjazz (commented):
Thanks for the patch!

There are a few other places where we need to change the "missing" color:

1. Rounded progress chart in the header in the translate view: https://github.com/mozilla/pontoon/blob/master/frontend/src/modules/resourceprogress/components/ProgressChart.js#L75

2. Missing icon in the filter menu in the translate view: https://github.com/mozilla/pontoon/blob/master/frontend/src/modules/search/components/SearchBox.css#L81 Please also remove the rule for the `hover` state at the end of the file. It's no longer needed, same as in the entity list.

3. Horizontal progress bar on dashboards: https://github.com/mozilla/pontoon/blob/master/pontoon/base/static/css/table.css#L303

anujpandey785:
Pushed the required Changes and removed the new rules added in the previous commit

mathjazz (approved):
Nice job!

PR #1682, opened 2020-09-13
Fix bug 1640235 - Return only public projects from the GraphQL API en…

jotes:
@mathjazz r? I think it's better to return only public projects because the GraphQL endpoint doesn't support authentication/authorization.


jotes:
TIL: The translate.next editor freezes when the user tries to access a private project. I'll try to solve this in this patch.

pontoon/tour/migrations/0001_squashed_0001_initial.py
```diff
@@ -40,6 +40,7 @@ def create_tutorial_project(apps, schema_editor):
         admin_notes=(
             "Do not remove, this is required in code to show the guided tour."
         ),
+        visibility="public",
```
mathjazz, line 43: Why is this change needed?

We already made existing projects public in a data migration:
https://github.com/mozilla/pontoon/blob/master/pontoon/base/migrations/0004_make_old_projects_public_again.py

mathjazz (commented):
Thanks for the patch!

I can imagine that we'd want to include private projects sometimes, as well. Could you add an `include_private` parameter, similar to `include_disabled` and `include_system`?

https://github.com/mozilla/pontoon/commit/4cd387da855b67072b847c3d7b18000a4cd5b18d

jotes:
`I can imagine that we'd want to include private projects sometimes`
I implemented `include_private` as the first solution but then I realized that the nature of private projects is to not expose them for certain queries (even when they want to include them).
E.g. a GraphQL query from a translator shouldn't ever return private projects, but a query from an Administrator/Manager should return all projects.
Nonetheless, our GraphQL endpoint doesn't have such an authorization/authentication layer, does it?

mathjazz:
Good point. Please ignore my comment.

jotes:
@mathjazz I've learned that Graphene has access to the user who sends the query (my previous understanding of Graphene in Pontoon was invalid).

I've updated the code and the UI is working fine. Also, I refactored a small parts of the code.

mathjazz:
Thanks for the update! I still don't understand the change in the migration file:
https://github.com/mozilla/pontoon/pull/1682#discussion_r488431250

pontoon/api/schema.py
```diff
@@ -111,16 +112,16 @@ class Meta:
...
-
-        records = qs.filter(project__disabled=False, project__system_project=False)
+        records = (
+            obj.project_locale.visible_for(_info.context.user)
+            .filter(
+                Q(project__disabled=include_disabled)
+                | Q(project__system_project=include_system)
+            )
+            .distinct()
+        )
```
mathjazz, line 122: I fell into a similar trap:
https://github.com/mozilla/pontoon/pull/1639#discussion_r431067188

If you'll check results, you'll notice it doesn't really work. We could also add tests to catch this.

pontoon/tour/migrations/0001_squashed_0001_initial.py
jotes: On my machine (and on TravisCI too), migration `base.0004`  executes before migration `tour.001`.

However, I've decided to revert this change and add a data-migration to fix it.

pontoon/api/schema.py
jotes: Okay, I see my flaw in understanding this problem. I think that without a testing matrix, anybody can make the same mistake.
I'll create a test and then try to figure if that's solvable in an ORM query.

pontoon/tour/migrations/0002_make_tutorial_public.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 2.2.13 on 2020-10-10 22:57
+
+from django.db import migrations
+
+
+def make_tutorial_public(apps, schema_editor):
+    """Tutorial should be public by default."""
+
+    Project = apps.get_model("base", "Project")
+    Project.objects.filter(slug="tutorial").update(visibility="public")
```
mathjazz, line 10: We've already made all projects public in https://github.com/mozilla/pontoon/blob/master/pontoon/base/migrations/0004_make_old_projects_public_again.py#L10.

Why are these migrations needed?

jotes: https://github.com/mozilla/pontoon/pull/1682#discussion_r502843775

mathjazz: Ah, missed first part of your comment.

Now I see where the problem see. I'll leave it to you to decide what's best to do here - update the migration or add a new one.

pontoon/api/schema.py
jotes: @mathjazz I added two tests to reduce the risk of that regression.

pontoon/api/schema.py
```diff
@@ -109,14 +111,17 @@ class Meta:
     )
 
     def resolve_localizations(obj, _info, include_disabled, include_system):
-        qs = obj.project_locale
+        projects = obj.project_locale.visible_for(_info.context.user)
```
mathjazz, line 114: Nit: I guess `_info` should be `info` now that we actually use it?

pontoon/api/schema.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import absolute_import
+
```
mathjazz, line 2: Shouldn't we remove this? https://github.com/mozilla/pontoon/commit/142f75013c691762e60ec1e69af0a9320cc14b45

pontoon/tour/migrations/0002_make_tutorial_public.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 2.2.13 on 2020-10-10 22:57
+
+from django.db import migrations
+
+
+def make_tutorial_public(apps, schema_editor):
+    """Tutorial should be public by default."""
+
```
mathjazz, line 8: I would add a clarifying comment here:
```
We need this migration, because migration base.0004 executes before migration tour.0001.
```

pontoon/api/schema.py
```diff
@@ -137,26 +142,27 @@ class Query(graphene.ObjectType):
...
 
     def resolve_projects(obj, info, include_disabled, include_system):
-        qs = ProjectModel.objects
         fields = get_fields(info)
 
+        projects = ProjectModel.objects.visible_for(info.context.user)
+        records = projects.filter(disabled=False, system_project=False)
+
+        if include_disabled:
+            records |= projects.filter(disabled=include_disabled)
```
mathjazz, line 151: I'd be consistent with `resolve_localizations()` here and below (`if include_system`) and either hardcode `True` or use `include_disabled` / `include_system` also in `resolve_localizations()`.

mathjazz (commented):
That works very well, thanks!

Left a few notes, nits.

pontoon/api/schema.py
jotes: I couldn't find other places that follow this convention so I'm going to replace `_info` with `info`.

pontoon/tour/migrations/0002_make_tutorial_public.py
mathjazz: Please also add the comment. :)

mathjazz (approved):
A thing of beauty! :-)

PR #1684, opened 2020-09-20
Fix  Bug 1611510  - Add ability to pin team comments

abowler2:
This PR builds on the source comment work by adding the ability for PMs to "Pin" comments within the Team Comment section and make the pinned comment visible globally for all locales in the metadata section.


frontend/src/modules/teamcomments/components/Count.js
```diff
@@ -18,8 +18,32 @@ export default function Count(props: Props) {
...
+        !pinnedCommentCount ? null :
+        <span className='pinned'>{ pinnedCommentCount }</span>
+    )
+
+    const commentTotal = (
+        !commentCount ? null :
+        <span>{ commentCount }</span>
+    )
+
+    const dash = (
```
abowler2, line 39: I wasn't sure what character would be best as a separator here. Went with the dash but let me know if you prefer something different.

pontoon/base/views.py
```diff
@@ -603,6 +604,24 @@ def add_comment(request):
     return JsonResponse({"status": True})
 
 
+@login_required(redirect_field_name="", login_url="/403")
+@require_POST
+@transaction.atomic
+def update_comment(request):
+    """ Update pinned status of comment """
```
abowler2, line 611: I chose to keep the naming neutral (rather than something like update_pinned_comment) in case we ever want to add additional update actions, but then added the note to indicate that only the pinned status is being updated ATM.

mathjazz (commented):
Some early comments. I haven't tested the patch locally yet due to some local setup issues.

frontend/src/core/comments/components/Comment.js
```diff
@@ -59,7 +71,27 @@ export default function Comment(props: Props) {
                     date={ new Date(comment.dateIso) }
                     title={ `${comment.createdAt} UTC` }
                 />
+                { canPin ?
+                    <Localized
+                        id="comments-Comment--pin-button"
+                        attrs={{ title: true }}
+                        vars={{
+                            pinnedTitle: !comment.pinned ? 'Pin comment' : 'Unpin comment',
+                            pinButton: !comment.pinned ? 'PIN' : 'UNPIN',
```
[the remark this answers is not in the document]
Pike: You could also change the ID for the button.

Re UPPERCASE, using CSS isn't good for l12y, I think.

frontend/public/static/locale/en-US/translate.ftl
```diff
@@ -90,6 +90,8 @@ comments-AddComment--mention-avatar-alt =
     .alt = User Avatar 
 comments-AddComment--submit-button = <glyph></glyph>
     .title = Submit comment
+comments-Comment--pin-button = { $pinButton }
+    .title = { $pinnedTitle }
```
Pike, line 94: Should we have distinct files for Admin-only strings? Being able to tell if a string  is needed for  users or  admins has  been frequently requested for AMO, for example.

frontend/src/core/comments/components/Comment.css
```diff
@@ -44,3 +45,25 @@
...
+    background-color: transparent;
+    border: none;
+    color: #AAAAAA;
+    font-size: 11px;
+    font-weight: 300;
+    padding: 0;
+}
+
+.comments-list .comment .comment-pin {
+    color: orange;
```
mathjazz, line 64: I'd change this block a little to make the pin icon smaller and its color same as the pinned comment count:

```css
.comments-list .comment .comment-pin {
    color: #7BC876;
    position: absolute;
    right: 4px;
    top: -6px;
}
```

frontend/src/core/comments/components/Comment.js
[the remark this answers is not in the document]
mathjazz: > Re UPPERCASE, using CSS isn't good for l12y, I think.

Interesting. I suggested CSS because using uppercase here is a stylistic choice that might not be carried over to a future redesign.

I'm curious to learn more about the l12y concerns of using CSS to change case. Is this one of them? https://stackoverflow.com/questions/28783259/how-do-i-make-text-transformuppercase-work-properly-with-greek

frontend/public/static/locale/en-US/translate.ftl
mathjazz: I'd leave translations in one file until we make the decision in https://bugzilla.mozilla.org/show_bug.cgi?id=1504710.

frontend/src/core/comments/components/Comment.js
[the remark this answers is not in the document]
Pike: The https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices#Create_localizable_UI linked from there is more to the point. Greek might be solved by now, not sure Gaelic is. IIRC, Turkish also has letters that are "interresting" to uppercase.

And then there's the question around uppercase being the right choice for a language to begin with. Similar to the Italics comment for CJKT on that page.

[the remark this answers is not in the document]
mathjazz: Thanks. We use `text-transform: uppercase` quite a lot in the Translate app, so I'll file a bug to change that.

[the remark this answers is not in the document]
mathjazz: Filed: https://bugzilla.mozilla.org/show_bug.cgi?id=1666277

frontend/src/modules/teamcomments/components/Count.js
abowler2: Just to clarify ~ I currently have the counts as `<# of pinned comments>-<total comments>`.  
Should that be `<# of pinned comments>+<# of remaining comments>` instead?

I didn't originally set it up this way since we are only adding the pin icon and not sorting the pinned comments to the top so it felt more like "there are pinned comments within these comments" and less like "here are the pinned comments and here is everything else".

mathjazz: >  Should that be `<# of pinned comments>+<# of remaining comments>` instead?

Yes. That would be consistent with the Machinery and Locales tabs and allow for consistent styling.

mathjazz:
Thanks for the update!

> Not entirely sure why it worked when passing in `pinned` or why `pin_comment` wasn't also broken, but nonetheless it appears to be fixed now.

Maybe the old URL schema resulted in always calling `pin_comment`?

frontend/src/modules/teamcomments/actions.js
```diff
@@ -38,6 +39,19 @@ export function request(
     };
 }
 
+export type UpdatePinnedAction = {|
+    +type: typeof UPDATE_PINNED,
+    +pinned: boolean,
+    +commentId: number,
+|};
+export function updatePinned(pinned: boolean, commentId: number):UpdatePinnedAction {
```
mathjazz, line 47: Nit: missing space after colon.

frontend/src/core/comments/components/Comment.js
```diff
@@ -14,16 +15,25 @@ import type { TranslationComment } from 'core/api';
...
 
 export default function Comment(props: Props) {
-    const { comment } = props;
+    const { comment, canPin, updatePinnedComment } = props;
 
     if (!comment) {
         return null;
     }
 
+    const handlePinnedComment = () => {
```
mathjazz, line 30: We should either have two separate methods or change the name to something like `handlePinAndUnpin`.

frontend/src/modules/teamcomments/actions.js
```diff
@@ -59,9 +73,24 @@ export function get(entity: number, locale: string): Function {
     }
 }
 
+export function savePinnedStatus(pinned: boolean, commentId: number): Function {
```
mathjazz, line 76: Maybe `togglePinnedStatus`, to clarify that we're flipping the value?

frontend/src/modules/teamcomments/actions.js
```diff
@@ -59,9 +73,24 @@ export function get(entity: number, locale: string): Function {
...
+    return async dispatch => {
+        if (pinned) {
+            await api.comment.pinComment(commentId);
+        }
+        else {
+            await api.comment.unpinComment(commentId);
+        }
+
+        dispatch(updatePinned(pinned, commentId))
```
mathjazz, line 85: I'd also use `toggle` instead of `update` here and in `UpdatePinnedAction` and `UPDATE_PINNED`.

I'd also call `updatePinnedComment` `togglePinnedStatus`.

mathjazz (commented):
Nice work! Everything works as expected.

I'm proposing some changes to make naming (hopefully) clearer.

frontend/src/modules/teamcomments/reducer.js
```diff
@@ -1,14 +1,15 @@
 /* @flow */
 
-import { RECEIVE, REQUEST } from './actions';
+import { RECEIVE, REQUEST, UPDATE_PINNED } from './actions';
```
mathjazz, line 3: TOGGLE_PINNED

frontend/src/modules/teamcomments/reducer.js
```diff
@@ -1,14 +1,15 @@
...
 
 import type { TeamComment } from 'core/api';
-import type { ReceiveAction, RequestAction } from './actions';
+import type { ReceiveAction, RequestAction, UpdatePinnedAction } from './actions';
```
mathjazz, line 6: TogglePinnedAction

mathjazz (commented):
A couple of übernits left (should have made a separate comment in the previous round of review).

abowler2:
> A couple of übernits left (should have made a separate comment in the previous round of review).

Or.... I could have had read your original comment more carefully 😄

frontend/src/modules/entitydetails/components/Metadata.css
```diff
@@ -30,6 +30,10 @@
     text-decoration: none;
 }
 
+.metadata .comment .content {
+    display: inline-block;
+}
+
```
mathjazz, line 36: Just realized this change is breaking comments into new lines:

Why do we need it?

abowler2: Since the pinned comments are coming in as `<p></p>` those comments were breaking into 2 lines without this. What's interesting is that locally `Comment` is not split:

Screen Shot 2020-09-25 at 1 15 10 PM

It looks like if I move it to the `<p>` then both comments render correctly. Let me know if that is what I should do or if you have an alternate solution.

mathjazz: SGTM

mathjazz (approved):
\o/

PR #1688, opened 2020-09-26
Fix Bug 1615565: Use prettier to format frontend code

abowler2:
This PR uses the [Prettier documentation](https://prettier.io/docs/en/install.html) along with the [previous work](https://github.com/mozilla/pontoon/compare/master...adngdb:prettier?expand=1) from @adngdb to add, configure, and run Prettier to format the frontend code.

Pending feedback and review on this I can add it to the CI as well as create a `make` command to make it easier to enforce the changes.


frontend/.prettierignore
```diff
@@ -0,0 +1,19 @@
+# dependencies
+/node_modules
```
abowler2, line 2: This file is a replication of the .gitignore file for the frontend as I was uncertain which files we might want to ignore, and felt this was a good representation.

jotes:
I don't modify the JS code lately, but I'm in favor of using an auto-formatter. The hanges look good for me :-)

 I can take a look later when the Makefile/Travis part is done, or help if there's any problem with that part.

mathjazz:
I like!

Let's add it to CI and add some make commands.

mathjazz (approved):
Excellent job!

PR #1690, opened 2020-09-28
Fix bug 1667765: Enable Sync button by default in production

mathjazz:
Sync button in Project Admin is not enabled by default on production instances (unlike in development instances), which is inconsistent with the [docs](https://mozilla-pontoon.readthedocs.io/en/latest/user/localizing-your-projects.html#adding-a-new-project-to-pontoon).


jotes (approved):
LGTM

PR #1692, opened 2020-10-01
Fix Bug 1666199 Unify download URL schema

anujpandey785:
1)Changed the url for translation memory (line 133)  in UserMenu.js file
2)Changed the url for translations (line 153)  in UserMenu.js file
3)Removed the parameters locale and slug passed in the url for
  downloading translation memory (line 59)  in pontoon/base/ urls.py
4)Modified the view download_translation_memory in pontoon/base/views.py
  to accept only 2 parameters i.e. request and filename and introduced a
  new variable vars to extract locale and slug using split function.
5)Changed the view name download in pontoon/base/views.py to translations
  (line 694) and changed the urls,name corresponding to it in
  pontoon/base/ urls.py (line 81)

### DIFF
--- a/pontoon/base/urls.py
+++ b/pontoon/base/urls.py
@@ -56,7 +56,7 @@ urlpatterns = [
     ),
     # Download translation memory
     url(
-        r"^translation-memory/(?P<filename>.+)\.tmx$",
+        r"^translation-memory/(?P<locale>[A-Za-z0-9\-\@\.]+)\.(?P<slug>[\w-]+)\.tmx$",
         views.download_translation_memory,
         name="pontoon.download_tmx",
     ),
--- a/pontoon/base/urls.py
+++ b/pontoon/base/urls.py
@@ -78,7 +78,11 @@ urlpatterns = [
         views.get_translations_from_other_locales,
         name="pontoon.other_locales",
     ),
-    url(r"^translations/", views.translations, name="pontoon.translations"),
+    url(
+        r"^translations/",
+        views.download_translations,
+        name="pontoon.download.translations",
+    ),
     url(r"^upload/", views.upload, name="pontoon.upload"),
     url(r"^user-data/", views.user_data, name="pontoon.user_data"),
 ]
--- a/pontoon/base/urls.py
+++ b/pontoon/base/urls.py
@@ -56,7 +56,7 @@ urlpatterns = [
     ),
     # Download translation memory
     url(
-        r"^(?P<locale>[A-Za-z0-9\-\@\.]+)/(?P<slug>[\w-]+)/(?P<filename>.+)\.tmx$",
+        r"^translation-memory/(?P<filename>.+)\.tmx$",
         views.download_translation_memory,
         name="pontoon.download_tmx",
     ),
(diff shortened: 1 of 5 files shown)


pontoon/base/urls.py
```diff
@@ -56,7 +56,7 @@
     ),
     # Download translation memory
     url(
-        r"^(?P<locale>[A-Za-z0-9\-\@\.]+)/(?P<slug>[\w-]+)/(?P<filename>.+)\.tmx$",
+        r"^translation-memory/(?P<filename>.+)\.tmx$",
```
jotes, line 59: hey @anujpandey785 
The patch looks good, but I have a small suggestion that potentially can reduce the amount of code.

The regex for this contains a variable that can be removed because it [gets overridden in the view](https://github.com/mozilla/pontoon/pull/1692/files#diff-d964cbd7ee26bf4860da2c98511ccb58R783).
```
r"^translation-memory/(?P<locale>[A-Za-z0-9\-\@\.]+)\.(?P<slug>[\w-]+)\.tmx$",
```

anujpandey785: Hii @jotes 
Thanks for reviewing the patch .Sorry, I had overlooked the fact that the variable filename is being overridden. Yeah that helps us reducing 5 lines of code I have made the suggested changes , pushing the changes now. Tysm :)

anujpandey785: @jotes Done 
But as we have changed the url for downloading the translational memory it would have its effect on button for downloading translation memory located in the project dashboard as well , we should change the url there too ?
Also, I am not sure about this but the tests which are failing do we also need to make changes in the test_tmx.py file for both failed cases?

pontoon/base/urls.py
```diff
@@ -78,7 +78,7 @@
         views.get_translations_from_other_locales,
         name="pontoon.other_locales",
     ),
-    url(r"^download/", views.download, name="pontoon.download"),
+    url(r"^translations/", views.translations, name="pontoon.translations"),
```
mathjazz, line 81: I'd make that less ambiguous. Maybe `views.download_translations` and `pontoon.download.translations`?

mathjazz (commented):
Thanks for the patch.

I've noticed the following issues:
- I'm unable to download TM from the dashboards (locale code is included in the URL)
- ~I'm unable to download translations (404)~

Please also fix the tests.

pontoon/base/urls.py
anujpandey785: * Yeah , i should have taken a hint from ```download_translation_memory```

anujpandey785:
> * I'm unable to download translations (404)

@mathjazz Could you please elaborate about the error in this issue this part as i am able to download translations in my local copy.

mathjazz:
> Is the Download translations option in the user menu giving a 404 error?

Correct.

pontoon/base/urls.py
jotes: I'll take a look at it today :-)

jotes: Okay, the tests are failing because of the changes in the URLs. I think it's a good idea to use [reverse](https://docs.djangoproject.com/en/3.1/topics/http/urls/#reverse-resolution-of-urls) and replace the hardcoded URLs.

anujpandey785: > Okay, the tests are failing because of the changes in the URLs. I think it's a good idea to use [reverse](https://docs.djangoproject.com/en/3.1/topics/http/urls/#reverse-resolution-of-urls) and replace the hardcoded URLs.

Thanks for the suggestion. I am working on it.

mathjazz (approved):
Well done!

PR #1693, opened 2020-10-04
Fix Bug 1601987 - Fix ability to resize translation textarea

abowler2:
The textarea within the editor was growing horizontally and overflowing into the third column. This PR fixes the textarea by setting it to grow vertically. 

However, when doing this the growth of the textarea would trigger a scroll event on the window body which would in turn prevent the suggestion box in the translation comments from appearing when below the original window height (also the borders on the columns would not render beyond the initial size). To fix this I moved the scroll event to the `main-column` and removed it from the `history-list`.

### DIFF
--- a/frontend/src/core/editor/components/connectedEditor.css
+++ b/frontend/src/core/editor/components/connectedEditor.css
@@ -3,7 +3,6 @@
     border-bottom: 1px solid #5e6475;
     display: flex;
     flex-direction: column;
-    height: calc(20% + 60px);
     min-height: 160px;
 }
 
@@ -18,6 +17,7 @@
     overflow: auto;
     padding: 10px;
     width: 100%;
+    resize: vertical;
 }
 
 /* Remove highlight in Chrome */
--- a/frontend/src/core/editor/components/connectedEditor.css
+++ b/frontend/src/core/editor/components/connectedEditor.css
@@ -3,6 +3,7 @@
     border-bottom: 1px solid #5e6475;
     display: flex;
     flex-direction: column;
+    flex-shrink: 0;
     min-height: 160px;
 }
 
--- a/frontend/src/core/comments/components/AddComment.js
+++ b/frontend/src/core/comments/components/AddComment.js
@@ -193,7 +193,7 @@ export default function AddComment(props: Props) {
             setScrollPosition(element.scrollTop);
         };
 
-        const historyScroll = document.querySelector('.main-column');
+        const historyScroll = document.querySelector('#history-list');
         const teamsScroll = document.querySelector('#react-tabs-3');
 
         if (!historyScroll && !teamsScroll) {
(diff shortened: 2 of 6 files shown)


frontend/src/core/editor/components/connectedEditor.css
```diff
@@ -3,7 +3,6 @@
     border-bottom: 1px solid #5e6475;
     display: flex;
     flex-direction: column;
-    height: calc(20% + 60px);
     min-height: 160px;
 }
```
abowler2, line 7: Since we no longer start with a calculated height, should this value be increased?

mathjazz (commented):
Thanks for the patch!

My concern is that the entire main (middle) column is now scrollable, whereas on prod you can only scroll within the History panel.

I think we should keep the toolbar (copy, previous, next), metadata and the editor always visible.

Same as e.g. in String list, where the search box is always visible and we only scroll what makes sense to be scrolled, i.e. the entities.

abowler2:
I don't disagree, however, moving the scroll event up one level to the `history` section does not work as the comments will expand without forcing the section to scroll which brings about the issues with the border not extending and the comment suggestions not rendering.

Another approach could be to remove the ability to resize the textarea and set it to scroll if the contents overflow.  This is what we did in the comments when we were originally using a textarea there.

Of course, if you have another solution that could work I would be more than happy to apply it.

mathjazz:
Let's see if I understand things correctly. Here's what I did (on master branch):
* set `resize: vertical` on the `textarea`
* removed `height: calc(20% + 60px);` from the `.editor`

Resizing of the `textarea` now works as expected (vertical only). But the content of the `.main-column` overflows if editor height is increased beyond its initial value, which causes two problems you pointed out:
* mentions popup in translation comments doesn't appear if below initial window height
* border on the `.main-column` do not render beyond the initial size

We should prevent overflow in `.main-column` by changing the `.history` height on `textarea` instead of keeping it hardcoded at `calc(60% - 125px)`. I wonder if we can do that without using JS (maybe flexbox or grid can help?).

abowler2:
Was able to eliminate the scrolling on the main column by utilizing flexbox on that column instead. (I swear I tried that before without luck 🤷‍♀️).

Let me know if you have any other concerns or changes you would like.

mathjazz:
Nice job!

Can you please prevent textarea height from changing when translation comments (causing history overflow) are collapsed / expanded?

frontend/src/core/editor/components/connectedEditor.css
mathjazz: I'd leave that for now. I think editor is mostly too big anyways, and it still fits 3 lines in the case of rich FTL editor.

mathjazz (approved):
Great job! Thanks for solving this old issue so elegantly!

PR #1694, opened 2020-10-04
Update README clarifying chat channel scope

zanona:
Many end-users who enter Pontoon's chat channel, probably do so by obtaining the link here _(as I did! :raised_eyebrow: )_
This helps to guide users to the right place for discussing their questions.

*Thanks to @flodolo,  for pointing that out.*

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,8 @@ If you want to go further, you can:
 * See Mozilla's Pontoon servers:
     * [Staging](https://mozilla-pontoon-staging.herokuapp.com/)
     * [Production](https://pontoon.mozilla.org/)
-* Get in touch with us on [chat.mozilla.org](https://chat.mozilla.org/#/room/#pontoon:mozilla.org) _(**Please note:** This channel is dedicated exclusively for discussing Pontoon's development. Questions about third-party deployment, syntax of localizable resources, general support, etc; are better suited for [our discourse forum](https://discourse.mozilla.org/c/pontoon/)_.
+* For discussing Pontoon's development, get in touch with us on [chat.mozilla.org](https://chat.mozilla.org/#/room/#pontoon:mozilla.org)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,7 +45,7 @@ If you want to go further, you can:
 * See Mozilla's Pontoon servers:
     * [Staging](https://mozilla-pontoon-staging.herokuapp.com/)
     * [Production](https://pontoon.mozilla.org/)
-* Get in touch with us on [chat.mozilla.org](https://chat.mozilla.org/#/room/#pontoon:mozilla.org).
+* Get in touch with us on [chat.mozilla.org](https://chat.mozilla.org/#/room/#pontoon:mozilla.org) _(**Please note:** This channel is dedicated exclusively for discussing Pontoon's development. Questions about third-party deployment, syntax of localizable resources, general support, etc; are better suited for [our discourse forum](https://discourse.mozilla.org/c/pontoon/)_.
```
mathjazz, line 48: I wouldn't be so "strict" here. :-)

I'd just add another bullet point along the lines of:
```
* For feedback, support and 3rd party deployments check out [Discourse](https://discourse.mozilla.org/c/pontoon/).
```

zanona: Sure, thing.

I believe that, if the chat is to be used exclusively for discussing Pontoon's development, it might be well worth establishing it, otherwise it could get a bit confusing to users in general as the line is actually pretty inviting as  _“Get in touch with us on chat.mozilla.org”_ :wink: 

Perhaps being transparent of the main scope could benefit everybody?

So perhaps:

```markdown
* For discussing Pontoon's development, get in touch with us on chat.mozilla.org
* For feedback, support and 3rd party deployments, checkout Discourse
```

What do you reckon?

mathjazz: I like that proposal, let's go with it!

mathjazz (approved):
Thanks for the patch!

PR #1695, opened 2020-10-06
Rename AddComments function

mathjazz:


abowler2:
It looks like the plural form is also used [here](https://github.com/mozilla/pontoon/blob/master/pontoon/base/forms.py#L318).
Not sure if that is appropriate there either as each comment is sent individually through that form on submission.

If it does need to be changed here then that Form is called in [`base/views.py`](https://github.com/mozilla/pontoon/blob/master/pontoon/base/views.py#L566) as well as used as a reference in comments in `Metadata.js`, `UserNotification.js`, and `Comment.js` which means those will all need to change as well.

mathjazz:
Excellent catch! I've updated the PR with requested changes.

abowler2 (approved):
All looks good!

PR #1697, opened 2020-10-09
Bug 1627655: Update python testing dependencies

PFischbeck:
In preparation for the Django upgrade and other package upgrades, the current testing framework should be up-to-date to catch all warnings.
This PR updates `pytest`, `pytest-django`, `coverage`, `pytest-cov`, `codecov` and `requests_mock`.
As a side effect, the coverage reporting to codecov, which fails currently, will hopefully work again now.


mathjazz:
Thanks for the patch, Philipp!

@jotes Would you have cycles for a review?

jotes (approved):
LGTM :-)

PR #1698, opened 2020-10-09
Fix bug 1668362: Use text cursor on search box icon

mathjazz:
This patch changes the cursor on the search box magnifier icon from cursor to text in order to prevent users from thinking the icon acts as a button.

The patch can be tested on stage:


mathjazz:
r+ed in https://bugzilla.mozilla.org/show_bug.cgi?id=1668362#c7

PR #1699, opened 2020-10-10
Bug 1627655: Remove django-nose dependency

PFischbeck:
The package doesn't officially support Django >=3.0, and `nose` is not maintained anymore. All usages were replaced by appropriate `pytest` equivalents.

I deleted the `TEST_RUNNER` definition in the settings. We could change it to run `pytest` instead when calling `./manage.py test` (see instructions [here](https://pytest-django.readthedocs.io/en/latest/faq.html#how-can-i-use-manage-py-test-with-pytest-django)), but I don't think that's worth it since we use the `pytest` command directly.

### DIFF
--- a/pontoon/base/tests/__init__.py
+++ b/pontoon/base/tests/__init__.py
@@ -13,7 +13,6 @@ from django.template.defaultfilters import slugify
 from django.test import TestCase as BaseTestCase, Client as BaseClient
 
 import factory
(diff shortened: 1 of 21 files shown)


mathjazz (approved):
Less dependencies and less lines of code. Woohoo! :-) 

Thanks for the patch, great work!

PR #1700, opened 2020-10-10
Authentication Error bug fix

abowler2:
Working on [Bug 1669192](https://bugzilla.mozilla.org/show_bug.cgi?id=1669192) uncovered a syntax error within the `authentication_error.html` file that was causing the app to crash instead of showing the "Sign in Failure" page.

This patch fixes that error and also includes adding `*html`  to the `prettierignore` file so that those are not reformatted with Prettier.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -13,6 +13,7 @@
 *.env.development.local
 *.env.test.local
 *.env.production.local
+*.html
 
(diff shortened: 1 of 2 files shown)


mathjazz (approved):
Nice catch!

PR #1701, opened 2020-10-11
Fix 1417530 - Convert the api tests from Unittest to Pytest.

jotes:
I didn't want to reinvent the wheel in #1682. There's no need for parametrization in Unittest when it's available in Pytest OOTB.


jotes:
@mathjazz r?

mathjazz (approved):
Nice work!

Changing the bug number and keeping the bug open - there are still places that use the old testing framework.

PR #1702, opened 2020-10-11
Fix Bug 1669192 - Unable to authorize with GitHub

abowler2:
Using GitHub to authenticate a signin was causing an error. The authentication would complete on GitHub but when redirected to Pontoon an error occurred.

In looking into this it was discovered that GitHub changed the way it handles authentication from using query params to using headers instead. The django-allauth package was updated with this change in v0.42.0.

Pontoon was still using v0.41.0 and updating to v0.42.0 fixes the issue.


jotes (approved):

mathjazz (approved):
Woohoo, just signed in using GitHub!

PR #1703, opened 2020-10-11
Fix 1417530 - Convert the tests from pontoon/db/test_lookups from Uni…

jotes:
@mathjazz r?


pontoon/db/tests/test_lookups.py
```diff
@@ -1,48 +1,56 @@
+"""
+Collation lookups allows a user to set text-collation to search queries.
```
mathjazz, line 2: Nit: allow.

mathjazz (commented):
Thanks for modernizing our codebase!

PR #1704, opened 2020-10-13
Bug 1600344: Remove six and __future__ usage

PFischbeck:
This removes the direct `six` dependency and removes all` __future__` imports. However, `six` is still an indirect dependency through several packages.
The templatetags helpers `urlparams` and `urlencode` were removed since they were unused (`base/templatetags/helpers.py`).

I think the only Python-2-specific part remaining is in the documentation, namely `setup-virtualenv.rst`. I didn't touch it, since much of it seems outdated anyways (e.g., assuming Ubuntu 16.04 or 17.04), so someone with more insight should update/remove this at some point.

Sorry about the 233 changed files! :sweat_smile:


Pike:
This looks nice and easy for like 300 of the 233 files ;-).

I looked at the test coverage for `base/templatetags/helpers.py`, are the other template tags w/out test coverage used?

And, probably a good other PR is to drop `mock` for `unittest.mock` ?

jotes:
:+1: to deprecate `setup-virtualenv.rst` in favor of the Docker setup.

Also, great work @PFischbeck!

PFischbeck:
> I looked at the test coverage for `base/templatetags/helpers.py`, are the other template tags w/out test coverage used?
> 
Looks like `DatetimeAwareJSONEncoder` and `LazyObjectsJSONEncoder` can be replaced by `DjangoJSONEncoder`, and `thisyear` and `naturalday` are unused. The rest is used, as far as I can tell. Should I do this in an extra PR?

> And, probably a good other PR is to drop `mock` for `unittest.mock` ?
> 
Sure, that's on my todo list too :smiley:

mathjazz:
> Should I do this in an extra PR?

Sounds like a good idea to me. In case something goes wrong, it's easier to revert a small patch.

PR #1705, opened 2020-10-13
Bug 1627655: Update django-guardian

PFischbeck:
The new version is compatible with Django 3.1. The `ANONYMOUS_USER_ID` setting has been removed since django-guardian 1.4.2.


mathjazz (approved):
Thanks for the patch!

mathjazz:
And yay for less settings. :)

PR #1706, opened 2020-10-16
Remove more unused templatetags helpers

PFischbeck:
As a follow-up to #1704, I removed some more helpers.
The DjangoJSONEncoder already supports datetime and Promise objects, and thisyear+naturalday were unused.
The previous encoder also handled QuerySet objects, but I couldn't find any place where this is actually used.


mathjazz (approved):
Thanks for the patch! Great work!

PR #1707, opened 2020-10-16
Fix failing tmx tests

PFischbeck:
Two tmx tests were marked with `xfail`. I fixed them by moving the test files to the right directory and updating the expected contents.
It seems that at some point, the entity key was additionally slugified, which is why I changed it to `pontoon:aa/bb/ccc:xyz` in the expected output, without all the extra encoded characters.


mathjazz (approved):
Nice! \o/

Just recently, xfail tests were a source of confusion for one of our new contributors.

PR #1708, opened 2020-10-16
Remove backports.csv dependency

PFischbeck:
All CSV features we use are natively supported by Python. In particular, the DictWriter takes care of missing row entries and treats them as an empty string.

### DIFF
--- a/pontoon/administration/views.py
+++ b/pontoon/administration/views.py
@@ -1,6 +1,5 @@
+import csv
 import logging
-
-from backports import csv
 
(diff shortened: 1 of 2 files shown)


mathjazz (approved):
A dependency less is a dependency less.

A note for the future: please file a bug first and reference it in the commit message. We try to only land patches without a bug reference if they are completely trivial (e.g. one-liners, minor CSS changes).

PFischbeck:
Okay, will do!

PR #1709, opened 2020-10-18
Bug 1627655: Update graphe-django

PFischbeck:
I updated `graphene-django` to ensure compatibility for the update to Django 3.1.
Using `only_fields` has been deprecated, so I replaced it with `fields` instead.


mathjazz (approved):
LGTM

Thanks!

PR #1710, opened 2020-10-20
Bug 1627655: Update django-cors-headers

PFischbeck:
The new version supports Django 3.1. The CORS_ORIGIN_ALLOW_ALL setting was renamed to CORS_ALLOW_ALL_ORIGINS.

### DIFF
--- a/pontoon/settings/base.py
+++ b/pontoon/settings/base.py
@@ -763,7 +763,7 @@ CELERY_ACCEPT_CONTENT = ["pickle"]
 # For the sake of integration with other sites,
 # some of javascript files (e.g. pontoon.js)
 # require Access-Control-Allow-Origin header to be set as '*'.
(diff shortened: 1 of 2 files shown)


pontoon/settings/base.py
```diff
@@ -763,7 +763,7 @@ def _allowed_hosts():
 # For the sake of integration with other sites,
 # some of javascript files (e.g. pontoon.js)
```
jotes, line 764: @PFischbeck Great work!
@mathjazz I think that was required by the legacy Translation view and the in-place editor. Does any Pontoon feature still needs it?

jotes (approved):

pontoon/settings/base.py
mathjazz: A line under hints at GraphQL.

mathjazz (approved):
Thanks for the patch!

PR #1711, opened 2020-10-20
Fix bug 1671295 - Fix the User Admin form and show the email field on…

jotes:


pontoon/base/models.py
```diff
@@ -1525,7 +1524,9 @@ def available_locales_list(self):
 
 class UserProfile(models.Model):
     # This field is required.
-    user = models.OneToOneField(User, models.CASCADE, related_name="profile")
+    user = models.OneToOneField(
+        User, models.CASCADE, related_name="profile", primary_key=True
```
jotes, line 1528: `primary_key=True` is required to make Django to figure out the relation between the objects and remove the IntegrittyError

pontoon/base/admin.py
```diff
@@ -43,6 +43,13 @@ class UserAdmin(AuthUserAdmin):
     list_filter = ("is_staff", "is_superuser")
     inlines = (UserProfileInline,)
 
+    add_fieldsets = (
+        (None, {
+            'classes': 'wide',
+            'fields': ('username', 'email', 'password1', 'password2')
```
jotes, line 49: I added this for the sake of UI because email is the first field on the  users list view.

jotes:
@mathjazz could you r?

mathjazz (approved):
Works great! Thanks!

PR #1712, opened 2020-10-21
Bug 1627655: Remove django-bulk-update dependency

PFischbeck:
The package has not been updated for 3 years, and is not tested on current Django versions. Thus, we use the Django-native `bulk_update` method instead.

The Django-native method, as opposed to the package method, does not return the number of updated objects, so I got rid of that in the single test where it was used. Also, I adjusted the `EntityQuerySet.bulk_update` method to not throw a `DeprecationWarning`, and take care of the word count.


mathjazz (approved):
Another one bites the dust. :-)

Thanks for getting rid of another legacy dependency.

PR #1716, opened 2020-10-23
Fix bug 1671904: Add specification for Community Health Dashboard

mathjazz:
This is a spec for the Community Health Dashboard, developed as part of the _Use Pontoon to create and display community health data visualizations for l10n community managers_ L10n-drivers OKR. It's based on the metrics we've identified in the [scoping document](https://docs.google.com/document/d/1I5Sq2SB3fTe9apDDCdtdm-F-hY6ReURS6WdMqkA8MOw/edit?ts=5f735c63#).

See [Pontoon Roadmap](https://wiki.mozilla.org/L10n:Pontoon_Roadmap#Community_Health_Dashboard) for development tracking. Once the spec lands, implementation bugs will be filed.

You can test the dashboard with the fake data on the [testing instance](https://pontoon-chd.herokuapp.com/en-GB/insights/).

Flagging all L10n drivers for a review.

@peiying2 r?
@delphine r?
@flodolo r?
@gueroJeff r?


specs/0108-community-health-dashboard.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Translation activity
+
+A combination of a line chart showing completion over time and a column chart showing approved translation submissions and source string additions over time in two separate columns. The translation submission column is a stack of two columns - human translations and Machinery translations. A tooltip showing exact data at the given time appears when hovering a chart.
+
+See the [existing script](https://github.com/flodolo/scripts/blob/master/pontoon/self_approval_ratio.py) for calculating self-approval ratio.
+
+## Review activity
+
+A combination of a line chart showing volume of unreviewed strings over time and a column chart showing review actions and suggestion submissions over time in two separate columns. The review actions column is a stack of three columns - peer-approvals, self-approvals and rejections. A tooltip showing exact data at the given time appears when hovering a chart.
```
flodolo, line 39: It's a bit confusing that one axis is "strings", the other is "volume", but they count the same thing.

specs/0108-community-health-dashboard.md
```diff
@@ -0,0 +1,49 @@
...
+
+# Out of scope
+
+At this stage we’re only interested in presenting data we already collect in Pontoon. We're primarily focusing on exposing data to team managers; project managers will be able to obtain data across all teams if needed, e.g. by using scripts and the API.
+
+# Feature explanation
+
+A new "Insights" tab is made available on the Team page, positioned next to the "Contributors" tab. It consists of several sections, presenting data for the period of the last 12 months. Each of the sections in described below. Fore more details about the design, see the [Mockup](#mockup) section.
+
+## Active users
```
flodolo, line 23: Title doesn't match the mockup.

flodolo (changes requested):
All graphs seem to be about the last 12 months, maybe it should be indicated in a "main title" at the beginning of the page, e.g. "Locale insights over the last 12 months"?

This is the first spec I check for Pontoon, so it's not clear to me to what level of detail it should go into. I would expect it to get very deep (e.g. what each axis' label), to spare the developer any doubt when it comes to implementation.

specs/0108-community-health-dashboard.md
```diff
@@ -0,0 +1,49 @@
...
+
+Pie charts showing ratios of active vs. all managers, reviewers and contributors. See the existing scripts ([1](https://github.com/flodolo/scripts/blob/master/pontoon/active_contributors.py), [2](https://github.com/flodolo/scripts/blob/master/pontoon/list_reviewers_with_contribution_stats.py)) for criteria of defining active users and for details on how to retrieve data.
+
+## Average review lag
+
+A line chart showing average unreviewed suggestion lifespans over time. See the [existing script](https://github.com/flodolo/scripts/blob/master/pontoon/unreviewed_suggestions_lifespan.py) for details on how to retrieve data. A tooltip showing exact data at the given time appears when hovering a chart.
+
+## Translation activity
+
+A combination of a line chart showing completion over time and a column chart showing approved translation submissions and source string additions over time in two separate columns. The translation submission column is a stack of two columns - human translations and Machinery translations. A tooltip showing exact data at the given time appears when hovering a chart.
```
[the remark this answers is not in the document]
Delphine: Yeah, a quick summary in a few words of how "human" and "machine" differentiate would be good.

I'm more though about the second item Flod mentions, which is "show which axis belongs to which series".

Delphine (commented):
Left a few comments. I think this is mostly straight-forward. A bit more explanations would be welcome, as Flod already called out. But otherwise I think this is good.
I'll wait for others to chime in and for current changes to be made to give this another round of review. Thanks M! This is cool 🦄

peiying2 (changes requested):
Added some of my own comments, and comments to other review comments for your consideration, :). Otherwise, I agree with them.

specs/0108-community-health-dashboard.md
```diff
@@ -0,0 +1,49 @@
...
+
+# Feature explanation
+
+A new "Insights" tab is made available on the Team page, positioned next to the "Contributors" tab. It consists of several sections, presenting data for the period of the last 12 months. Each of the sections in described below. Fore more details about the design, see the [Mockup](#mockup) section.
+
+## Active users
+
+Pie charts showing ratios of active vs. all managers, reviewers and contributors. See the existing scripts ([1](https://github.com/flodolo/scripts/blob/master/pontoon/active_contributors.py), [2](https://github.com/flodolo/scripts/blob/master/pontoon/list_reviewers_with_contribution_stats.py)) for criteria of defining active users and for details on how to retrieve data.
+
+## Average review lag
```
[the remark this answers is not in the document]
mathjazz: Another one that comes to my mind is "Average unreviewed suggestion lifespan".

specs/0108-community-health-dashboard.md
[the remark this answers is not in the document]
mathjazz: I haven't found good practices of matching Y axes to their corresponding series. I think you'll get used to it. 😃

Kidding aside, maybe the tooltip will help a little. I'll deploy charts to stage.

specs/0108-community-health-dashboard.md
mathjazz: "Reviewed strings" is also not a great title, because it also covers "Added suggestions". I'm changing it to "Suggestions" and the other axis to "Unreviewed Suggestions".

mathjazz:
> All graphs seem to be about the last 12 months, maybe it should be indicated in a "main title" at the beginning of the page, e.g. "Locale insights over the last 12 months"?

They are about the last 12 months as per spec, although charts are dynamic, so one can also change range using mouse. I've added explicit range selector we already use on Contributors pages instead. Alternatively we could also add such selector to each chart separately, which would allow for more flexibility. Thoughts?

> This is the first spec I check for Pontoon, so it's not clear to me to what level of detail it should go into. I would expect it to get very deep (e.g. what each axis' label), to spare the developer any doubt when it comes to implementation.

We haven't finalized (i.e. documented) the spec process yet, so we're still learning. The idea is to have an agreement between stakeholders what needs to be implemented in enough detail to for an engineer to estimate the amount of work and file implementation bugs.

mathjazz:
Thanks for the feedback so far!

I've updated the spec and deployed the design with fake data to stage for easier testing:
https://mozilla-pontoon-staging.herokuapp.com/fr/insights/

The patch is ready for another round of reviews.

specs/0108-community-health-dashboard.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Unreviewed suggestion lifespan
+
+A line chart showing time before suggestion gets approved or rejected. See the [existing script](https://github.com/flodolo/scripts/blob/954fa85/pontoon/unreviewed_suggestions_lifespan.py) for details on how to retrieve data. A tooltip showing the exact data at a given time appears when hovering over a chart.
+
+## Translation activity
+
+A combination of a line chart showing completion over time and a column chart showing approved translation submissions and source string additions over time in two separate columns. The translation submission column is a stack of two columns - human translations (`Translation.machinery_sources` not set) and Machinery translations (`Translation.machinery_sources` set). A tooltip showing the exact data at a given time appears when hovering over a chart.
```
flodolo, line 33: The explanation doesn't really help: `Translation.machinery_sources` set means that the translation was taken from machinery and saved without changes? How do we plan to explain this in the graph, for locale managers?

Should it be "Machinery matches", or something along that line?

specs/0108-community-health-dashboard.md
```diff
@@ -0,0 +1,49 @@
...
+
+See the [existing script](https://github.com/flodolo/scripts/blob/954fa85/pontoon/self_approval_ratio.py) for calculating self-approval ratio.
```
flodolo, line 35: This paragraph belongs to the next section.

specs/0108-community-health-dashboard.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Review activity
+
+A combination of a line chart showing the number of unreviewed strings over time and a column chart showing review actions and suggestion submissions over time in two separate columns. The review actions column is a stack of three columns - peer-approvals, self-approvals and rejections. A tooltip showing the exact data at a given time appears when hovering over a chart.
```
flodolo, line 39: The graph on stage is missing the right axis at this point.

specs/0108-community-health-dashboard.md
mathjazz: Oops, now I understand the original question was more about terminology than definition. "Copied from Machinery" is probably the most accurate. Not sure what to replace Human Translations with, though.

Or we can at least give the hint in the legend and clarify it in the snippet under the chart as pointed out above: https://github.com/mozilla/pontoon/pull/1716#discussion_r512501074.

specs/0108-community-health-dashboard.md
mathjazz: That's by design. See https://github.com/mozilla/pontoon/pull/1716#discussion_r512563932.

flodolo: I missed that part. At this point, why not have a group with 3 bars? The line is clearly following a different scale, so it's even more confusing than trying to figure out which graph belongs to which axis.

mathjazz: I thought about that, too. The thing is that the amount of Unreviewed suggestions is a result of items in bars (some sort of summary), so I think it deserves a special attention. The same is true for the Completion line in the Translation Activity chart.

Should I put the secondary Y axis back? That would be consistent with the Translation Activity chart.

flodolo: > Should I put the secondary Y axis back? That would be consistent with the Translation Activity chart.

I think so, otherwise the graphic looks wrong (vs confusing) ;-)

mathjazz: Updated the screenshot and the stage.

flodolo (approved):

Delphine (approved):
r+

specs/0108-community-health-dashboard.md
```diff
@@ -0,0 +1,49 @@
...
+- Created: 2020-10-23
+- Associated Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1672966
+
+# Summary
+
+Create a team-specific health dashboard targeted at Pontoon users with Manager rights.
+
+# Motivation
+
+Team managers as well as project managers should have the ability to understand how localization communities are performing. Team dashboards already provide some of the answers, including the overall status of completion and translation statistics, but the information is only available in the form of a current snapshot.
```
gueroJeff, line 11: This is a little bit of a nit, but could we reframe this as "engagement" instead of "performance"? I know that I've historically used that term, but I've realized that there's a nuance to the term "performance" that relates to an employer - employee relationship. We should seek to avoid that and speak of this in terms of how much the community is engaged in our localization initiatives.

specs/0108-community-health-dashboard.md
```diff
@@ -0,0 +1,49 @@
...
+
+At this stage we’re only interested in presenting data we already collect in Pontoon. We're primarily focusing on exposing data to team managers; project managers will be able to obtain data across all teams if needed, e.g. by using scripts and the API.
+
+# Feature explanation
+
+A new "Insights" tab is made available on the Team page, positioned next to the "Contributors" tab. It consists of several sections, presenting data for the period of the last 12 months. Each section is described below. Fore more details about the design, see the [Mockup](#mockup) section.
+
+## Active users
+
+Pie charts showing ratios of active vs. all managers, reviewers and contributors. See the existing scripts ([1](https://github.com/flodolo/scripts/blob/954fa85/pontoon/active_contributors.py), [2](https://github.com/flodolo/scripts/blob/954fa85/pontoon/list_reviewers_with_contribution_stats.py)) for criteria of defining active users and for details on how to retrieve data.
```
gueroJeff, line 25: We should expect some confusion around introducing the term "reviewers" as a role, when that role doesn't existing within Pontoon. We could either formalize that change by renaming the "Translator" role to "Reviewer" or lean more heavily into the ambiguity and change "reviewer" to "translator" for consistency.

specs/0108-community-health-dashboard.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Unreviewed suggestion lifespan
+
+A line chart showing time before suggestion gets approved or rejected. See the [existing script](https://github.com/flodolo/scripts/blob/954fa85/pontoon/unreviewed_suggestions_lifespan.py) for details on how to retrieve data. A tooltip showing the exact data at a given time appears when hovering over a chart.
+
+## Translation activity
```
gueroJeff, line 31: Ok, looking at the dashboard, why are we including a bar for new source strings added? New source strings come in independent of localizer engagement and don't really serve as an interesting point of reference within this context (especially when the translation activity stats are significantly higher volume than the new source strings). Could we make human v machinery into two columns and remove the new source strings column? Also, instead of raw volume, could the tooltip be a percentage of the total, so we can see the ratio of human translations vs machinery translations more clearly?

specs/0108-community-health-dashboard.md
```diff
@@ -0,0 +1,49 @@
...
+
+A combination of a line chart showing completion over time and a column chart showing approved translation submissions and source string additions over time in two separate columns. The translation submission column is a stack of two columns - human translations (`Translation.machinery_sources` not set) and Machinery translations (`Translation.machinery_sources` set). A tooltip showing the exact data at a given time appears when hovering over a chart.
+
+## Review activity
+
+A combination of a line chart showing the number of unreviewed suggestions over time and a column chart showing review actions and suggestion submissions over time in two separate columns. The review actions column is a stack of three columns - peer-approvals, self-approvals and rejections. A tooltip showing the exact data at a given time appears when hovering over a chart.
```
gueroJeff, line 37: What does the added column represent and what is its purpose?

specs/0108-community-health-dashboard.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Unreviewed suggestion lifespan
+
+A line chart showing time before suggestion gets approved or rejected. See the [existing script](https://github.com/flodolo/scripts/blob/954fa85/pontoon/unreviewed_suggestions_lifespan.py) for details on how to retrieve data. A tooltip showing the exact data at a given time appears when hovering over a chart.
+
+## Translation activity
+
+A combination of a line chart showing completion over time and a column chart showing approved translation submissions and source string additions over time in two separate columns. The translation submission column is a stack of two columns - human translations (`Translation.machinery_sources` not set) and Machinery translations (`Translation.machinery_sources` set). A tooltip showing the exact data at a given time appears when hovering over a chart.
+
+## Review activity
```
gueroJeff, line 35: I'm conflicted about the use of raw volume numbers instead of ratios. It feels like self-approval here is out of place, as the focus appears to be a depiction of what happens to unreviewed translations over time. As a community manager, I would be most interested in understanding of these unreviewed translations, our ratio of peer-approving translations to peer-rejecting translations is X, and this is what that means in terms of volume. Self-approval feels like it should be its own column maybe to compare peer review results against self-approval results?

gueroJeff (changes requested):

specs/0108-community-health-dashboard.md
flodolo: IMO the new string bar is useful to understand what happened: did they translate more because there was more to translate, or are they doing QA?

gueroJeff: We have other, more concrete, ways of depicting what happened though. For instance, each string has a translation history. If that history log increases, then we can be sure its QA. If history is created for the first time, we can be sure that it's a new translation. With the bar, we're left to make assumptions rather than knowing for certain.

specs/0108-community-health-dashboard.md
flodolo: I personally think the latter would be more confusing, because the review is done by both manager and translators. Maybe "Active reviewers", to imply that it's an action and not a role?

specs/0108-community-health-dashboard.md
flodolo: True that, although I don't know if that information can be collected easily.

Having said that, knowing how many strings are "requested" every month is something I'd like to know, so maybe it should be in its own graph?

specs/0108-community-health-dashboard.md
gueroJeff: That could work, with a tooltip that explains how active reviewers are defined?

specs/0108-community-health-dashboard.md
gueroJeff: I don't know either, and maybe that's something to include in the out-of-scope list (Matjaz can make that call).

Agreed, that is an interesting metric that might be better suited to its own graph.

specs/0108-community-health-dashboard.md
mathjazz: Yup, that is confusing. I don't know if adding "Active" would solve the problem, because the whole section is already called "Active (users)" and the charts imply "Active" managers/contributors/reviewers using the "(out of N)" block.

If we're splitting users by activity type, Contributors is also not a great name - it implies any kind of activity, but it only includes users that submitted translations or suggestions. What about Authors? That should also reduce the confusion with user roles.

And I just realized the spec should explicitly define what is an active manager, because the scripts we link to don't cover that category. I assume active manager is a manager who authored or reviewed a translation in a period of time?

I'm also in favor of adding a [tooltip or some other form of description](https://github.com/mozilla/pontoon/pull/1716#discussion_r512501074) with more details.

gueroJeff: I've thought we needed to revisit the naming for roles for a while to make them more descriptive of our expectation of their principle function (e.g., Contributor -> Translator, Translator -> Reviewer). If this is the opportunity to talk about it, let's go for it. Probably in a different forum though.

We previously had defined an active manager as a manager who logged into Pontoon within a certain time frame.

mathjazz: > I've thought we needed to revisit the naming for roles for a while to make them more descriptive of our expectation of their principle function (e.g., Contributor -> Translator, Translator -> Reviewer). If this is the opportunity to talk about it, let's go for it. Probably in a different forum though.

I'm open for that. Are you also suggesting we should split these charts by role instead of activity type?

> We previously had defined an active manager as a manager who logged into Pontoon within a certain time frame.

I saw that in the scoping doc. Isn't that somewhat low barrier to become active compared to other two categories?

specs/0108-community-health-dashboard.md
mathjazz: "Suggestion submissions". Maybe New or Submitted would be more self-explanatory than Added?

The chart shows how the amount on Unreviewed suggestions changes over time. Without showing newly added suggestions, it could be confusing to see Unreviewed suggestions grow for example, despite the evident review actions.

specs/0108-community-health-dashboard.md
gueroJeff: I more mean that the role names should reflect their contribution/activity. The way the chart is split is fine, it's the larger issue of role names. Not sure what Flod thinks?

specs/0108-community-health-dashboard.md
gueroJeff: What other ways would cause unreviewed suggestions to grow? Is the uptick self-explanatory?

specs/0108-community-health-dashboard.md
gueroJeff: I'm basing naming on role expectations. We don't expect a manager's primary contribution to be a translation, it's to keep their community healthy and overcome bottlenecks. We expect "Translators" to be the peer reviewers. We expect "contributors" to suggest translations.

mathjazz: I added a note about detailed descriptions for each data series.

To avoid blocking the OKR, let's save renaming of roles (and possibly other related terms) for another occasion.

specs/0108-community-health-dashboard.md
mathjazz: I'd leave that for locale managers to tell, because I'm biased. 😄

Note that each data series can be enabled or disabled by clicking on the corresponding item in the legend. They can also be disabled by default.

specs/0108-community-health-dashboard.md
mathjazz: The charts actually show the ratios already:
- blue vs red: approvals vs rejections
- light blue vs dark blue: peer-approvals vs self-approvals

We can add the ratios in numeric form to the tooltip (in any combination, e.g. peer-approvals vs. (peer-)rejections and self-approvals vs. other peer review actions).

gueroJeff: I caught that, I should have been more clear. I'm wondering if a percentage ratio in the tooltip to reinforce the visual ratio is more helpful.

mathjazz:
In addressing Jeff's comments, I've put them into three groups, because they are somewhat related to each other:

1. **More clarity on what each data series means ([1](https://github.com/mozilla/pontoon/pull/1716#discussion_r514386991))**: tooltips have been added to legend items as well as to each chart. Note that this is just to demonstrate two of the options available. We probably don't need both. Content in the tooltips also needs work after we pick which variant we want to use. Also: "Added" are now called "New".

2. **Show rations instead of raw volume numbers ([1](https://github.com/mozilla/pontoon/pull/1716#discussion_r514396889), [2](https://github.com/mozilla/pontoon/pull/1716#discussion_r514403499))**: ratios have been added to tooltips where requested. Raw numbers are still there.

3. **Remove auxilary data, not directly related to the OKR ([1](https://github.com/mozilla/pontoon/pull/1716#discussion_r514396889), [2](https://github.com/mozilla/pontoon/pull/1716#discussion_r514398627))**: Newly added source strings and newly added suggestions are now disabled by default in the Translation activity and Review activity charts. Relatedly, I've made it cleared legend items are toggles.

I haven't updated the spec, screenshot or the demo. Instead, I've pushed the new iteration:
https://pontoon-chd-v2.herokuapp.com/en-GB/insights/

You can compare that with the old version still available here:
https://pontoon-chd.herokuapp.com/en-GB/insights/

flodolo:
A few nits on the tooltips:
* I think they should say "within *the* selected timeframe", instead of "a timeframe"?
* Translation activity tooltip: disabled by default -> hidden by default

Besides that, I really like the updated dashboard.

gueroJeff (approved):
I really like the changes, thanks Matjaz!

Delphine:
Thanks for the updates Matjaz! looks really good as is 
r+ for me

peiying2 (approved):
Great work! Thank you!

PR #1717, opened 2020-10-25
Fix bug 1672713: Remove django-session-csrf dependency

PFischbeck:
The package is natively replaced by Django. We use CSRF_USE_SESSIONS to force Django to save the CSRF  tokens in the session instead of in a cookie. We also get rid of the patching of the AppConfig and AdminSite.

The changes to the admin site were needed because after deleting the `SessionCsrfAdminSite`, Django complained about the re-registering of the `User`. I fixed this as suggested in the [Django docs](https://docs.djangoproject.com/en/3.1/topics/auth/customizing/#extending-the-existing-user-model), and I also removed the redundant Group/GroupAdmin registration.

### DIFF
--- a/pontoon/base/admin.py
+++ b/pontoon/base/admin.py
@@ -2,11 +2,8 @@ import uuid
 
 from django.contrib import admin
 from django.contrib.auth import get_user_model
-from django.contrib.auth.admin import (
-    UserAdmin as AuthUserAdmin,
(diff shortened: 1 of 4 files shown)


mathjazz (approved):
Thanks for the patch and explanation of the changes.

LGTM.

I tested on stage and it works as expected.

PR #1718, opened 2020-10-25
Bug 1627655: Update factory-boy

PFischbeck:
Since this is a testing dependency, it is moved from requirements/default.txt to requirements/test.txt.

One new dependency of factory-boy is python-dateutil. We previously used the py-dateutil package, which is an old, unmaintained version of python-dateutil, so both packages would use the same namespace. Thus, we only use python-dateutil now.


mathjazz (approved):
Thanks!

PR #1719, opened 2020-10-26
Bug 1627655: Update django-pipeline dependency

PFischbeck:
The new version supports Django 3.1, and futures is not a dependency any more.

I'm not sure how to properly test this locally, so this should probably be tested in staging.


mathjazz (approved):
Works fine on stage. Thanks!

PR #1720, opened 2020-10-28
Bug 1611510: Make pinned comments globally visible

mathjazz:
This patch makes Pinned comments globally visible to all locales.

The 2nd commit unifies team comments tab count behavior with other tabs counts.


abowler2 (approved):
Other than needing to run Prettier, this works as expected and the code all looks good to me.

PR #1721, opened 2020-10-28
Bug 1611510: Send Pin Comment notifications

mathjazz:
Notify:
- authors of existing translations across all locales
- reviewers of existing translations across all locales

Send separate notification for each locale the user has contributed to.
Each notification is then linked to corresponding translate view.

### DIFF
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -1,5 +1,7 @@
 import logging
 import re
+
+from collections import defaultdict
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


mathjazz:
No idea why the `tests/js/utils/datetime.js` is suddenly failing since the previous commit. Can't reproduce the problem locally.

abowler2 (approved):
Works as expected and the code looks good to me.

I also was not able to reproduce the test failure locally.

PR #1722, opened 2020-10-29
Fix bug 1674173: Add support for non-ASCII characters in mentions

mathjazz:
One should be able to mention people with fancy names like `@théo` or `@matjaž`.


abowler2 (approved):
That's some fancy Regex ✨ to use for those fancy names 😄 

LGTM

mathjazz:
Thanks!

I was able to mention @TheoChevalier on stage:
https://mozilla-pontoon-staging.herokuapp.com/sl/firefox/all-resources/?string=74128

abowler2:
> I was able to mention @TheoChevalier on stage:
> https://mozilla-pontoon-staging.herokuapp.com/sl/firefox/all-resources/?string=74128

I thought for sure you were Rick Rollin' us there.... but it was even better 😂

PR #1724, opened 2020-10-30
Fix bug 1671836: Remove django-partial-index

PFischbeck:
The dependency was only used in one migration, and the introduced partial indexes were already removed again in the next migration. Thus, this deletes the creation in the first migration, and deletes them in the second migration if they are present. This second step can be removed if the two migrations are squashed.


Pike (approved):
lgtm.

jotes (approved):

PR #1725, opened 2020-10-31
Fix Bug 1624370 -  Rich Editor crash

abowler2:
When transitioning from a translation that is open in a simple editor to a suggestion that is open in a rich editor the app would crash. The problem happened when trying to serialize a translation that is already serialized

To fix this I am only sending the data to be serialized if it is updated.

### DIFF
--- a/frontend/src/modules/fluenteditor/components/RichEditor.js
+++ b/frontend/src/modules/fluenteditor/components/RichEditor.js
@@ -76,20 +76,23 @@ export default class RichEditor extends React.Component<Props> {
         this.props.sendTranslation(ignoreWarnings, fluentString, message);
     };
 
-    updateUnsavedChanges = (
-        translation?: Translation,
-        initial?: Translation,
-    ) => {
+    updateUnsavedChanges = () => {
         const props = this.props;
 
-        this.props.updateUnsavedChanges(
-            !translation
-                ? props.editor.translation
-                : fluent.serializer.serializeEntry(translation),
-            !initial
-                ? props.editor.initialTranslation
-                : fluent.serializer.serializeEntry(initial),
-        );
+        let translation = props.editor.translation;
+        let initial = props.editor.initialTranslation;
+
+        // If we're switching from or to a simple editor, we will not be able
+        // to serialize the current or initial translation
+        try {
+            translation = fluent.serializer.serializeEntry(translation);
+        } catch (e) {}
+
+        try {
+            initial = fluent.serializer.serializeEntry(initial);
+        } catch (e) {}
+
+        props.updateUnsavedChanges(translation, initial);
     };
 
     render() {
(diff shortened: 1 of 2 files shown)


mathjazz:
Thanks for the patch! It fixes the bug, but I see an issue I can't reproduce on master:

1. I add a new project linked to https://github.com/mozilla-l10n/pontoon-ftl and enabled it for `sl`.
2. I open the following string: https://github.com/mozilla-l10n/pontoon-ftl/blob/master/sl/demo.ftl#L15.
3. I open the next string.
4. Unsaved changes warning appears (it shouldn't).

abowler2:
So, I was tracking down the likely cause for the issue you raised and it seems that something within the process of flattening the message registers as a change for that particular string's format. In investigating that further I found a solution that fixes the issue; however, it doesn't make sense to me why it would. 

Right now I've added a check after the message is flattened to see if the message 'equals' the translation and if so then update the translation. It feels like the opposite check should be true. I've gone over the code in both files and can't wrap my head around why I need to check for equality and not lack of equality.

Of course, the real issue could be that I haven't in fact found the true cause of the bug 😆, so any insight you can add would be appreciated.

mathjazz:
There's a problem with this fix - values that need to be flattened, don't get flattened. You'd probably get to the same result by simply removing the two `updateTranslation()` calls.

To demonstrate the problem, open https://github.com/mozilla-l10n/pontoon-ftl/blob/master/sl/demo.ftl#L21 in the editor and see that only first part of the value of the `one` variant appears in the editor.

I think we need to change the initial patch. I'll add another comment in the code.

frontend/src/modules/fluenteditor/components/RichEditor.js
```diff
@@ -82,17 +82,13 @@ export default class RichEditor extends React.Component<Props> {
     ) => {
         const props = this.props;
 
-        if (!translation) {
```
mathjazz, line 85: First of all, `updateUnsavedChanges()` is always called without arguments, so we should:
1. Remove arguments from the function signature.
2. Set the default values.
3. Only serialize content when we know it's serializable (and explain that in a comment).

A possible solution:
```js
updateUnsavedChanges = () => {
    const props = this.props;

    let translation = props.editor.translation;
    let initial = props.editor.initialTranslation;

    // If we're switching from or to a simple editor, we will not be able
    // to serialize the current or initial translation
    try {
        translation = fluent.serializer.serializeEntry(translation);
    } catch (e) {}

    try {
        initial = fluent.serializer.serializeEntry(initial);
    } catch (e) {}

    props.updateUnsavedChanges(translation, initial);
};
```

abowler2:
Thank you for taking a look at this and offering your suggestion. That does in fact look like it has fixed the original bug without causing additional issues. 

As always, let me know if anything additional is needed for this.

mathjazz (commented):
Everything works fine.

Please fix eslint and we're good to merge.

mathjazz (approved):
Excellent job!

PR #1726, opened 2020-11-02
Fix Bug 1658867: Fix for tags in the filter panel 

Mitch4sho:
Addressing Bug 1658867 Tag priority. Fix for tags in the filter panel to put in order of highest priority to lowest, Changes tested by flow.

### DIFF
--- a/frontend/src/modules/search/components/FiltersPanel.js
+++ b/frontend/src/modules/search/components/FiltersPanel.js
@@ -299,64 +299,59 @@ export class FiltersPanelBase extends React.Component<Props, State> {
                                     </Localized>
 
                                     {this.props.tagsData
-                                        
-                                        .sort((a:  Tag, b:  Tag) =>  {
-                                                return b.priority - a.priority;
-                                            })
-                                        
+                                        .sort((a: Tag, b: Tag) => {
+                                            return b.priority - a.priority;
+                                        })
                                         .map((tag, i) => {
-                                                const selected =
-                   
-                                                                            props.tags[tag.slug];
+                                            const selected =
+                                                props.tags[tag.slug];
 
-                                                let className = tag.slug;
-                                                if (selected) {
-                                                    className += ' selected';
-                                                }
+                                            let className = tag.slug;
+                                            if (selected) {
+                                                className += ' selected';
+                                            }
 
-                                                return (
-                                                    <li
-                                                        className={`tag ${className}`}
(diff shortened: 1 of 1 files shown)


abowler2:
Hey @Mitch4sho, congrats on your first PR for Pontoon! 🎉 

I noticed that your Prettier settings made changes that aren't consistent with the settings for Pontoon. You can run `make prettier` on your branch to have the correct formatting applied and then push those changes (you might need to disable your editor from formatting on save in order to not have those changes overridden).

Let me know if you need any help with that.

frontend/src/modules/search/components/FiltersPanel.js
```diff
@@ -298,53 +298,60 @@ export class FiltersPanelBase extends React.Component<Props, State> {
...
-                                                onClick={this.createApplySingleFilter(
-                                                    tag.slug,
-                                                    'tags',
-                                                )}
-                                            >
-                                                <span
-                                                    className='status fa'
-                                                    onClick={this.createToggleFilter(
+                                    {this.props.tagsData
+                                        .sort((a: Tag, b: Tag) => {
```
abowler2, line 302: From working with you on this I know that the `Tag` types were added in the attempt to work out what was causing Flow errors. I'm not sure if they are needed here now and will defer to @mathjazz on that.

abowler2 (commented):
Nice work!  

This sorts the tags as expected, and pending the below comment re: typing the arguments in Sort, the code looks good to me too.

frontend/src/modules/search/components/FiltersPanel.js
mathjazz: Yeah, please remove the `Tag` types - the rule of the thumb is not to use them as long as Flow doesn't complain.

Please also replace `this.props.tagsData` with `props.tagsData`.

There's a lot of indent in here. One day we should refactor and probably move each Filter section into a separate component.

Mitch4sho: Okay, will remove the `Tag` types, replace with `props.tagData`and recommit the following. Thank you for reviewing!

mathjazz (approved):
Nice work!

Thanks for the patch. And welcome to Pontoon! 🥳

Mitch4sho:
Thank you!

mathjazz:
Works beautifully on stage:

PR #1648, opened 2020-06-16
Fix bug 1646010 - Refactor Editor code to use React hooks

adngdb:
Finally! It took months but I got there: this is, as far as I can tell, fully functional and ready to be pulled in. :relaxed: 

I did my best to rework my previous branch and split the work into smaller and more meaningful commits. The exception is the last commit, because that one includes all the components that depend on each other. But most of the preparation work is in previous commits, which should make that last one easier to review.

This is a lot to take in, as hooks really are a different approach to writing React components, but I like it overall. I had to learn a lot, and did not do any work to pass that knowledge on, so please do ask questions during review. That should serve a good basis for updating the documentation and providing resources to learn more. 

A good place to start learning about hooks is [React's documentation about them](https://reactjs.org/docs/hooks-intro.html). I also appreciated reading this [article going deeper on the `useEffect` hook](https://overreacted.io/a-complete-guide-to-useeffect/) (thanks April :wink: ).

Note that we are losing a bunch of unit tests, for those in `connectedEditor.test.js` I haven't ported elsewhere. If anyone is up for adding more test coverage to this part of the code, that'd be great!

PR #1727, opened 2020-11-05
Fix bug 1646010 - Refactor Editor code to use React hooks

adngdb:
This is a continuity of #1648 but rebased down to one commit, because that was the simplest way to rebase after the prettier change. Tests are passing, hopefully this is fully functional, but there were a few issues that I haven't been able to reproduce.

@mathjazz would you have some time to test this and give me reproduction cases for errors you find? Thanks!


mathjazz:
Thanks for your hard work on this huge patch!

Here are the issues I see:

1. Leaving [delete-all-message-special-cases](https://pontoon-1646010.herokuapp.com/sl/test-fluent/demo.ftl/?string=182) results in a false positive unsaved changes warning. [It should not](https://mozilla-pontoon-staging.herokuapp.com/sl/test-fluent/demo.ftl/?string=216478).

1. [Default cursor position](https://pontoon-1646010.herokuapp.com/sl/test-fluent/demo.ftl/?string=113) in the generic editor is at the end. [It should be at the beginning](https://mozilla-pontoon-staging.herokuapp.com/sl/test-fluent/demo.ftl/?string=216436).

1. Clicking on a [rejected suggestion](https://pontoon-1646010.herokuapp.com/sl/test-fluent/demo.ftl/?string=113) opens it in a source editor. [It should open in a rich editor](https://mozilla-pontoon-staging.herokuapp.com/sl/test-fluent/demo.ftl/?string=216436).

adngdb:
>     3. Clicking on a [rejected suggestion](https://pontoon-1646010.herokuapp.com/sl/test-fluent/demo.ftl/?string=113) opens it in a source editor. [It should open in a rich editor](https://mozilla-pontoon-staging.herokuapp.com/sl/test-fluent/demo.ftl/?string=216436).

I cannot reproduce that one. Can you make me an account on that test instance, or give me more precise steps to reproduce? Thanks!

I'm working on the other two problems in the meantime. :)

mathjazz:
Sure thing!

Username: https://github.com/mozilla/pontoon/blob/master/app.json#L39
Password: https://github.com/mozilla/pontoon/blob/master/app.json#L47

adngdb:
I am very hopeful that this is finally done (yay, isn't it, like, the third time I wrote this? :laughing: ). I've fixed all the problems you addressed, @mathjazz, and I also fixed the performance issues while typing. Please do give it another sound testing if you don't mind, because I've had to make another round of substantial changes, and I might have missed something, or created new side-effects. Thanks!

mathjazz:
Thanks!

I assume you intentionally left tests broken?

adngdb:
Hum… Yes, yes. Of course… :eyes:

mathjazz:
I can't break it! All issues seem to be fixed!

Do you want to fix the tests before I do the actual **code** review?

adngdb:
Yeah, I've fixed the main issue (a circular dependency) but I'm seeing other failures that I don't quite understand yet. Hopefully it won't have a big impact but I can't be sure yet.

adngdb:
Alright, this is it! Tests are finally passing!

frontend/src/modules/fluenteditor/components/FluentEditor.js
```diff
@@ -0,0 +1,281 @@
...
+import SourceEditor from './source/SourceEditor';
+import SimpleEditor from './simple/SimpleEditor';
+import RichEditor from './rich/RichEditor';
+
+/**
+ * Hook to update the editor content whenever the entity changes.
+ *
+ * This hook is in charge of formatting the translation content for each type of Fluent Editor.
+ * It does *not* update that content when the user switches the "force source" mode though,
+ * as that is dealt with with the `useForceSource` hook.
```
abowler2, line 23:
```suggestion
 * as that is dealt with  by using the `useForceSource` hook.
```

frontend/src/modules/fluenteditor/components/FluentEditor.js
```diff
@@ -0,0 +1,281 @@
...
+            ),
+        );
+    }
+
+    // Show a button to allow switching to the source editor.
+    let ftlSwitch = null;
+    // But only if the user is logged in and the string is not read-only.
+    if (user.isAuthenticated && !isReadOnlyEditor) {
+        if (syntaxType === 'complex') {
+            // TODO: To Localize
```
abowler2, line 256: Were you planning to do the localization here and on #L267 in this patch?

frontend/src/modules/fluenteditor/components/rich/RichEditor.js
```diff
@@ -0,0 +1,94 @@
...
+    ftlSwitch: React.Node,
+};
+
+/**
+ * Rich Editor for supported Fluent strings.
+ *
+ * This shows the Fluent translation based on its AST, presenting a nicer
+ * interface to the user. The translation is stored as an AST, and changes
+ * are made directly to that AST. That is why lots of Editor methods are
+ * overwritten, to handle the convertion from AST to string and back.
```
abowler2, line 22:
```suggestion
 * overwritten, to handle the conversion from AST to string and back.
```

abowler2 (commented):
This is really nice work! I actually enjoyed going through this as it was very useful for me to see your use of hooks here. 

I left a few comments, but overall it looks good to me.

frontend/src/modules/fluenteditor/components/FluentEditor.js
adngdb: No, that's outside the scope of this, which is why I put those comments instead. I think there are still a few parts that are not localized throughout the app, and we'll need to make a thorough audit of the app before enabling translations. But that should be left for a different bug! :-)

adngdb:
Thanks for the review April! Comments addressed (or answered :grinning: ).

mathjazz (commented):
Wow! This is massive. Excellent job!

I like so many things:
- I dig that we no longer need to pass so={many} props={from} one={component} to={another} one={whoops}!
- Big plus on how you structured components into subfolders!
- Testing the patch on stage it seems much more responsive!
- And - you removed 10% of the code!

The only thing I don't like is that I don't understand all of the code fully yet. There are just too many changes. But things will get better over time. :-)

A few questions:

1. What happened to isActionDisabled?
https://github.com/mozilla/pontoon/blob/master/frontend/src/core/editor/components/FailedChecks.js#L42

1. What rule do you use to decide when to put the factored out component into a separate file vs. when to keep it in the same file?
https://github.com/mozilla/pontoon/pull/1727/files#diff-ecb26bc37dcade3ddee0f6aa7e828506cd0b7c90f840d6f1aaa71cf8e5050664R44

1. You left some editor components as class components (e.g. KeyboardShortcuts). I assume that's intentional and we simply follow the [Gradual Adoption Strategy](https://reactjs.org/docs/hooks-intro.html#gradual-adoption-strategy)?

First question is really the only thing that prevents me to just go ahead and approve this.

adngdb:
>     1. What happened to isActionDisabled?

If I remember correctly, there were two elements: first was that it broke components, and I'd have to reimplement it entirely. Second was that I'm convinced this is a bad pattern and we should use something better instead, like I did in several places already (like having a marker in the store telling if a request is already running). So I decided to kill that "action disabled" think entirely, to force us to replace it with proper code, but as you can see I didn't do the latter part. Maybe we could track the places where this mechanism is needed and file follow-up bugs? (Note that this is fairly easy code to write and could be a good opportunity to create "good first bugs".) What do you think? 

>     2. What rule do you use to decide when to put the factored out component into a separate file vs. when to keep it in the same file?

I've probably used the "wet finger" rule here. Meaning I just went with the flow and didn't look back too  much. But, here's what should drive this decision: 
- if the hook is complex enough (more than a few lines of code), move it to a separate function;
- if the hook is being used by several components, move it to a separate file.

I very likely have a few hooks that are in external files while being used only once. I'll take a pass at that and make it more coherent. 

>     3. You left some editor components as class components (e.g. KeyboardShortcuts). I assume that's intentional and we simply follow the [Gradual Adoption Strategy](https://reactjs.org/docs/hooks-intro.html#gradual-adoption-strategy)?

Yeah, I did the ones that were essential, because they would break or otherwise defeat the purpose of this patch, but I didn't touch more than was necessary. We should indeed rewrite to hooks slowly over time, but we don't have to do that right away.

mathjazz (approved):
Thanks for the updates and clarifications!

Let's file followup bugs to address changes to `withActionsDisabled` (and other changes if needed). It's still in use in the History panel and I think we only need to replace it in the `FailedChecks` component. I haven't seen any issues in the Editor, probably because it uses isRunningRequest.

mathjazz:
Wait. This one deserves a GIF:

adngdb:
Me right now:



Thanks everyone for the reviews!

PR #1729, opened 2020-11-08
Fix Bug 1675438 - Poorly wrapped text in comment

abowler2:
When adding a comment if the text extended beyond the end of the line it would split the word and continue on the next line. This was being caused by the `word-break` property that was originally set when the comment input was a textarea. With the switch to the Slate editor this is no longer needed and long text now wraps to the next line without breaking.

### DIFF
--- a/frontend/src/core/comments/components/AddComment.css
+++ b/frontend/src/core/comments/components/AddComment.css
@@ -17,6 +17,7 @@
     margin: auto;
     overflow: auto;
     padding: 7px;
+    word-break: break-word;
 }
 
 .comments-list .add-comment .comment-editor .mention-element {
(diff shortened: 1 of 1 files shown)


frontend/src/core/comments/components/AddComment.css
```diff
@@ -17,7 +17,6 @@
     margin: auto;
     overflow: auto;
     padding: 7px;
-    word-break: break-all;
```
mathjazz, line 20: `word-break: break-word;` would be consistent with Comment.css and prevent long words from breaking the editor. For example, try pasting:
```
mmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm
```

PR #1732, opened 2020-11-12
Fix Bug 1607158 Upgrade PostgreSQL database version used in travis

anujpandey785:
Updated PostgreSQL 9.4 to PostgreSQL 11.5 in travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ script:
   - cat app.json | python -m json.tool > /dev/null
 
 addons:
-  postgresql: "11.5"
+  postgresql: "11.10"
   apt:
     packages:
       - language-pack-tr
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ script:
   - cat app.json | python -m json.tool > /dev/null
 
 addons:
-  postgresql: "9.4"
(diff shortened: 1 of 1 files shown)


mathjazz:
Tests are failing with "could not connect to server: Connection refused".

Seems like 11.5 is not supported and we should use 11 instead:
https://docs.travis-ci.com/user/database-setup/#using-a-different-postgresql-version
https://wiki.postgresql.org/wiki/Apt

mathjazz (commented):
Looks like you'll also need to update the dist to `bionic`, which [supports PostgreSQL 11](https://docs.travis-ci.com/user/reference/bionic/#databases-and-services).

I tried it myself and it passed:
https://github.com/mozilla/pontoon/pull/1742/files

I still haven't figured out how to make 11.5 work. Maybe @adngdb has ideas?

adngdb:
I would assume that what they use behind version "11" is the latest dot release for that major version, but I have no proof of that. According to the docs, there's not other version available than "11". If what we use in prod is 11.X, then I think it's fine to use 11 in travis.

mathjazz:
Thanks, that make sense.

mathjazz (approved):
Good work!

PR #1733, opened 2020-11-12
Fix bug 1676927: Enable force sync for projects with sync disabled

mathjazz:
This patch ignores the `sync_disabled` flag on force-sync and makes error messages clearer.

@jotes Could you help with the review, pretty please?

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -1138,14 +1138,17 @@ class ProjectQuerySet(models.QuerySet):
         """
         return self.available().filter(system_project=False)
 
+    def force_syncable(self):
+        """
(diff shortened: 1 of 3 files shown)


jotes (approved):
The patch looks good! I have only one small doubt about the naming convention, because `sync_disabled` sounds like `automatic synchronization is disabled`, maybe it should be called `autosync_disabled`? Also, this could require additional changes and a migration, so that's not something to fix in this PR.

mathjazz:
When sync is disabled, manual sync is also disabled (not just automatic).

Only force sync (which in its essence skips detecting changes and syncs everything) overrides the setting.

PR #1734, opened 2020-11-13
Fix bug 1674159: Fix slow terminology tests

PFischbeck:
The tests were slow because `update_terminology_project_stats` was called when creating a new `Term`. We now mock the calls to this expensive function. The whole `pytest` suite now only takes about 2.5 minutes.

I couldn't find an easy way to pass parameters from `TermFactory.create` to `Term.save` in order to optionally skip the stat updates. This is why I chose the mocking/patching option.

### DIFF
--- a/pontoon/terminology/tests/test_models.py
+++ b/pontoon/terminology/tests/test_models.py
@@ -7,7 +7,8 @@ from pontoon.test.factories import EntityFactory, TermFactory, TermTranslationFa
 
 
 @pytest.fixture
-def available_terms():
+@patch("pontoon.terminology.models.update_terminology_project_stats")
+def available_terms(_):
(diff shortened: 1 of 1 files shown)


pontoon/terminology/tests/test_models.py
```diff
@@ -63,11 +66,12 @@ def test_terms_for_string(string, found_terms, available_terms):
     assert len(terms) == len(found_terms)
 
     for i, term in enumerate(terms):
-        term.text = found_terms[i]
+        assert term.text == found_terms[i]
```
mathjazz, line 69: Nice catch!

mathjazz (approved):
Thanks, it works very well and very fast! :-)

PR #1735, opened 2020-11-13
Fix bug 1676857: Add template for writing specifications

mathjazz:
This patch adds a template for writing specifications, which must be used when submitting a feature specification for Pontoon. It builds upon specifications that have landed so far.

It's also the last step before we can mark [Specification Process](https://wiki.mozilla.org/L10n:Pontoon_Roadmap#Specification_Process) as done on our Roadmap.

We'll include more details about the process of writing specifications in the [Development Process](https://wiki.mozilla.org/L10n:Pontoon_Roadmap#Development_Process) document.

@adngdb Would you be able to review this?

### DIFF
--- a/specs/0000-template.md
+++ b/specs/0000-template.md
@@ -28,4 +28,3 @@ Feel free to add additional sections as needed. Common examples include (and are
 - **Out of scope:** In describing a feature scope it's sometimes helpful to explaining what doesn't fit in.
 
 Note that you are welcome to use all features of the [Markdown](https://guides.github.com/features/mastering-markdown/) syntax in formatting your specification.
-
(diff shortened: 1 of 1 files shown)


specs/0000-template.md
```diff
@@ -0,0 +1,30 @@
...
+
+Note that we use present simple tense across the entire specification.
+
+# Optional sections
+
+Feel free to add additional sections as needed. Common examples include (and are not limited to):
+
+- **Mockup(s):** Sometimes it's easier to explain the UI with a screenshot or a drawing.
+- **Algorithm:** Explain parts of the feature with pseudocode (or Python or JS code).
+- **Roles:** Provide details on how the feature impacts different user roles?
```
adngdb, line 27: Why the question mark here?

adngdb (commented):
This looks good, it's simple, which I like, and has worked well for the few specifications that have been written so far. I would simply add two notes:

1. Explain the file naming convention.
2. Explain that assets need to go in a folder with the spec number.

adngdb (approved):

PR #1737, opened 2020-11-14
Bug 1417530 - Migrate Contributors unit-tests to Pytest.

jotes:
Misc:
* Improve the performance of Contributors tests - achieved by decreasing the number of objects produced by the test fixtures.
* Introduce a new configuration option to set the page size of the events on the Contributor's timeline.

@mathjazz r? The execution time of the tests in this module dropped from ~30s to ~10s.


mathjazz (approved):
Well done. And yay for the speedup!

PR #1741, opened 2020-11-15
Fix Bug 1647124 - RTL support in comments

abowler2:
The comment editor handled the RTL text properly but, when the comment displayed it miss-ordered the sentence if English letters were inside the arab/kurdish setence.

It appears that having the direction set in the parent div caused the direction to be set by the first element, which is the username/email.  I've moved the attribute `dir: auto` from the parent div to a div that directly wraps the comment.

I've not been able to test that this picks up the correct direction with `auto`, but it does display correctly when directly set to the `rtl` value.

### DIFF
--- a/frontend/src/core/comments/components/Comment.js
+++ b/frontend/src/core/comments/components/Comment.js
@@ -39,36 +39,34 @@ export default function Comment(props: Props) {
                 imageUrl={comment.userGravatarUrlSmall}
             />
             <div className='container'>
-                <div className='content' dir='auto'>
-                    <div>
-                        <a
-                            className='comment-author'
-                            href={`/contributors/${comment.username}`}
-                            target='_blank'
-                            rel='noopener noreferrer'
-                            onClick={(e: SyntheticMouseEvent<>) =>
-                                e.stopPropagation()
-                            }
-                        >
-                            {comment.author}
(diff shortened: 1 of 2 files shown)


mathjazz:
Please check the tests. :)

mathjazz:
Thanks for the patch!

I think we actually need that `<div>` I told you to remove. Check out the 2nd comment:
https://mozilla-pontoon-staging.herokuapp.com/ar/firefox/all-resources/?string=74127

Regarding the actual issue, I think this patch solves the problem. Ideally I believe RTL comments should be aligned right (check out the 1st comment above), but I'm not sure we can fix that with this approach.

abowler2:
I know I've said it before.....but, one day I will remember to test multi-line comments 🙄 

I also had to change the comment wrapper from a `div` to a `span` in order to keep the proper inline display of the first line.

abowler2:
> Ideally I believe RTL comments should be aligned right (check out the 1st comment above), but I'm not sure we can fix that with this approach.

Yeah, my first thought would be to set a ref on that element and check for RTL direction. If found, then set a className that aligns the contents to 'flex-end'. Not sure if that would work though... 🤔

mathjazz (approved):
Well done!

PR #1743, opened 2020-11-16
Fix bug 1614588: Make Save/Suggest/Approve spinner appear again

mathjazz:
Seems like we forgot to port this node to the new react-fluent syntax.


adngdb (approved):
Ah, that's probably something I forgot in an old Fluent migration. Sorry!

Also, stealing this review from @abowler2. :smiling_imp:

PR #1746, opened 2020-11-19
Fix Bug 1673238: Interactive tour shows wrong elements for Machinery/Locales

Mitch4sho:
Addressing bug 1673238: Fixed interactive tour, added two tabs for comments and terms that point to the comments and terms sections. Pointed the machinery and locales tabs to the machinery and locales section. I was not sure what description I should put for the comments and terms tabs in the interactive tour.

### DIFF
--- a/frontend/public/static/locale/en-US/translate.ftl
+++ b/frontend/public/static/locale/en-US/translate.ftl
@@ -379,11 +379,19 @@ interactivetour-InteractiveTour--history-content =
 
 interactivetour-InteractiveTour--terms-title = Terms
 interactivetour-InteractiveTour--terms-content =
-    Test
+    The Terms panel contains any specialized words
+    found in the source string, along with their
+    definition, usage example, part of speech and
+    translation. By clicking on the term, its
+    translation gets inserted into the editor
 
 interactivetour-InteractiveTour--comments-title = Comments
 interactivetour-InteractiveTour--comments-content =
-    Test
+    In the Comments tab you can discuss how to
+    translate a string with your fellow team
+    members. It's also the place where you can
+    request more context about or report an issue in
+    the source string.
     
 interactivetour-InteractiveTour--machinery-title = Machinery
 interactivetour-InteractiveTour--machinery-content =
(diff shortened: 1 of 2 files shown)


mathjazz:
Hi @Mitch4sho. Thanks for your patch!

Could you please rebase your branch against latest master? If you'll look into the [Diff](https://github.com/mozilla/pontoon/pull/1746/files), you'll notice that this patch also contains all the changes from your "uppercase" patch.

Instead, it should only contain your last commit: https://github.com/mozilla/pontoon/pull/1746/commits/1ddd15f3f9f3dcf78d80ae6281adaeae23f20ed6.

Let us know if you need help rebasing!

Mitch4sho:
Yes will make sure fix this as soon as possible

frontend/public/static/locale/en-US/translate.ftl
```diff
@@ -377,6 +377,14 @@ interactivetour-InteractiveTour--history-content =
     each entry indicate its review status (Approved, Rejected or
     Unreviewed).
 
+interactivetour-InteractiveTour--terms-title = Terms
+interactivetour-InteractiveTour--terms-content =
+    Test
```
mathjazz, line 382: _The Terms panel contains any specialized words found in the source string, along with their definition, usage example, part of speech and translation. By clicking on the term, its translation gets inserted into the editor._

frontend/public/static/locale/en-US/translate.ftl
```diff
@@ -377,6 +377,14 @@ interactivetour-InteractiveTour--history-content =
...
+
+interactivetour-InteractiveTour--comments-title = Comments
+interactivetour-InteractiveTour--comments-content =
+    Test
```
mathjazz, line 386: _In the Comments tab you can discuss how to translate a string with your fellow team members. It's also the place where you can request more context about or report an issue in the source string._

mathjazz (commented):
Great work!

Please update the contents of the new tour steps as proposed and we can merge this.

mathjazz (approved):
Great job, @Mitch4sho!

I found it easier to just push the changes (fixing the failing test, adding missing period and changing a couple of words) instead of calling them out.

Mitch4sho:
Thank you appreciate I'll dive in more in why the test failed I'm not sure why. Thank you once again!

mathjazz:
Sorry, that's hard to see now: `eslint` was failing with "HTML entities must be escaped", because ' was used instead of ` in "It’s also the place...".

PR #1748, opened 2020-11-22
Prettier updates

abowler2:
The original implementation of Prettier only covered the files in the `/frontend` folder. This moves Prettier to the main folder so that it covers the `.js` and `.css` files in `/frontend`, `/pontoon`, and `/tests` folders.


abowler2:
Welp, I clearly broke some stuff here.... Gonna close this PR while I work it all out.

PR #1749, opened 2020-11-22
Move prettier to main folder

abowler2:
The original implementation of Prettier only covered the files in the `/frontend` folder. This moves Prettier to the main folder so that it covers the `.js` and `.css` files in` /frontend`, `/pontoon`, and `/tests` folders.

In addition, I added a script to give the option to just run a format 'check' with Prettier.

_______
At the moment there is a conflict with one of the build scripts that causes tests to fail. I've tried to track it down but have been unsuccessful. I would appreciate your thoughts on how to fix this.


mathjazz:
For reference, I'm quoting the error in question below.

Sadly it's hard to tell which files are breaking from the error message. I've ran a few manual tests and was able to proof that the `/tests` folder and the CSS files in the `/pontoon` folder are OK.

Multiple JS files in the `/pontoon` folder are causing breakage, one of them being `pontoon/teams/static/js/permissions.js`.

```
pontoon@bc462b7b52f2:/app$ ./manage.py collectstatic --noinput
  File "./manage.py", line 20, in <module>
  File "/usr/local/lib/python3.8/site-packages/django/core/management/__init__.py", line 401, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.8/site-packages/django/core/management/__init__.py", line 395, in execute
  File "/usr/local/lib/python3.8/site-packages/django/core/management/base.py", line 330, in run_from_argv
  File "/usr/local/lib/python3.8/site-packages/django/core/management/base.py", line 371, in execute
  File "/usr/local/lib/python3.8/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 194, in handle
    collected = self.collect()
  File "/usr/local/lib/python3.8/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 132, in collect
    for original_path, processed_path, processed in processor:
  File "/usr/local/lib/python3.8/site-packages/pipeline/storage.py", line 37, in post_process
    packager.pack_javascripts(package)
  File "/usr/local/lib/python3.8/site-packages/pipeline/packager.py", line 122, in pack_javascripts
    return self.pack(package, self.compressor.compress_js, js_compressed, templates=package.templates, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/pipeline/packager.py", line 116, in pack
    content = compress(paths, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/pipeline/compressors/__init__.py", line 66, in compress_js
    js = getattr(compressor(verbose=self.verbose), 'compress_js')(js)
  File "/usr/local/lib/python3.8/site-packages/pipeline/compressors/yuglify.py", line 15, in compress_js
    return self.compress_common(js, 'js', settings.YUGLIFY_JS_ARGUMENTS)
  File "/usr/local/lib/python3.8/site-packages/pipeline/compressors/yuglify.py", line 12, in compress_common
    return self.execute_command(command, content)
  File "/usr/local/lib/python3.8/site-packages/pipeline/compressors/__init__.py", line 250, in execute_command
    raise CompressorError(stderr)
pipeline.exceptions.CompressorError: b'\n/app/node_modules/yuglify/bin/yuglify:77\n                throw(err);\n                ^\nSyntaxError: Unexpected token: punc ())\n    at JS_Parse_Error.get (eval at <anonymous> (/app/node_modules/yuglify/node_modules/uglify-js/tools/node.js:21:1), <anonymous>:75:23)\n'
```

Pike:
https://github.com/mozilla/pontoon/blob/28802fba2e8fa444ae6f62f8042bc828a6eded9c/pontoon/teams/static/js/permissions.js#L116 seems to have an extra trailing `)`, and no test coverage?

jotes:
@Pike `permissions.js` is an ancient piece of JQuery-based code. It's not tested, sadly.

Pike:
I think I found the culprit, it's trailing commas on function args, https://github.com/prettier/prettier/issues/152.

I think the only way to "fix" this is to disable trailing commas in the pontoon folder. Or figure out a new compressor pipeline.

A `pontoon/.prettierrc.json` like this fixed it locally for me:

```json
{
    "trailingComma": "none"
}
```

abowler2:
> I think I found the culprit, it's trailing commas on function args, [prettier/prettier#152](https://github.com/prettier/prettier/issues/152).
> 
> I think the only way to "fix" this is to disable trailing commas in the pontoon folder. Or figure out a new compressor pipeline.

Yep, the trailing commas do in fact appear to be the issue (thank you @Pike!)

I currently have it set to `all` but I can change it to `es5` for the `/pontoon` file (which as per the [documentation](https://prettier.io/docs/en/options.html#trailing-commas) will put trailing commas where allowed as per es5 (such as objects, arrays, etc...)) and this allows the build to succeed.

abowler2:
I went ahead and added a separate Prettier config file for the `/pontoon` folder changing the settings to reflect `trailingComma: "es5"`  in order to eliminate the trailing comma in functions that was causing issues with the compressor. 

Let me know if any further changes are needed.

pontoon/administration/static/js/admin_project.js
```diff
@@ -1,252 +1,284 @@
...
+    $('html')
+        .unbind('keydown.pontoon')
+        .bind('keydown.pontoon', function (e) {
+            if (
+                $('input[type=text]:focus').length > 0 ||
+                $('input[type=url]:focus').length > 0
+            ) {
+                var key = e.keyCode || e.which;
+                if (key === 13) {
+                    // Enter
```
Pike, line 49: I think this comment is off on its own line.

Maybe just remove it? Add a `keyCode == 13` to the comment in line 39 instead, if we want to keep the semantics?

pontoon/base/static/css/boilerplate.css
```diff
@@ -16,59 +16,168 @@
...
+header,
+hgroup,
+menu,
+nav,
+section,
+summary,
+time,
+mark,
+audio,
+video {
```
Pike, line 87: Not a fan of this change.

Also, this stylesheet is vendored in, we might want to exclude it from reformatting.

PS: There's also a newer version of this file. Maybe look at an update?

pontoon/base/static/css/fontawesome-all.css
```diff
@@ -7,3197 +7,4250 @@
...
-  text-rendering: auto;
-  line-height: 1; }
+    -moz-osx-font-smoothing: grayscale;
+    -webkit-font-smoothing: antialiased;
+    display: inline-block;
+    font-style: normal;
+    font-variant: normal;
+    text-rendering: auto;
+    line-height: 1;
+}
```
Pike, line 17: This is also a vendored CSS file, also probably best to exclude it.

pontoon/homepage/static/js/lib/fullpage.js
```diff
@@ -1,3869 +1,4391 @@
...
+            return root.fullpage;
+        });
+    } else if (typeof exports === 'object') {
+        // Node. Does not work with strict CommonJS.
+        module.exports = factory(window, document);
+    } else {
+        // Browser globals.
+        window.fullpage = factory(window, document);
+    }
+})(this, window, document, function (window, document) {
```
Pike, line 25: This extra closing `)` is surprising?

Also, vendored.

pontoon/in_context/static/css/agency.css
```diff
@@ -1,834 +1,853 @@
...
\ No newline at end of file
+/*!
+ * Start Bootstrap - Agency Bootstrap Theme (http://startbootstrap.com)
+ * Code licensed under the Apache License v2.0.
+ * For details, see http://www.apache.org/licenses/LICENSE-2.0.
+ */
+
+body {
+    overflow-x: hidden;
+    font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif;
```
Pike, line 9: This file is also vendored.

pontoon/in_context/static/js/agency.js
```diff
@@ -1,26 +1,31 @@
...
+    $('a.page-scroll').bind('click', function (event) {
+        var $anchor = $(this);
+        $('html, body')
+            .stop()
+            .animate(
+                {
+                    scrollTop: $($anchor.attr('href')).offset().top,
+                },
+                1500
+            );
```
Pike, line 18: ... so is this.

Pike (commented):
Gave this a glance-over. I think we should exclude vendored files, I know of one (jquery-ui) that I didn't comment on explicitly, there might be more.

I'm torn on the single-quote choice, not sure if we should bite that bullet now that we bite that bullet?

I'm not sure if there were more comments that moved to weird places beyond the one I found at the top.

abowler2:
Removed the vendored files and modified the comment in `admin_project.js` as @Pike suggested. 

As always, let me know if any additional changes are needed.

.prettierignore
```diff
@@ -14,6 +14,13 @@
 *.yml
 **/base/static/js/lib*
 **/base/templates/js/pontoon.js
+**/base/static/css/boilerplate.css
+**/base/static/css/fontawesome-all.css
+**/base/static/css/jquery-ui.css
```
mathjazz, line 19: Please also add `**/base/static/css/nprogress.css`.

mathjazz (approved):
Nice work!

One last file to ignore and we're good.

mathjazz (approved):
Good work! Thanks!

PR #1750, opened 2020-11-23
Bug 1602896 - Force-sync the project, when the project config files c…

jotes:
@mathjazz r?

### DIFF
--- a/pontoon/sync/vcs/models.py
+++ b/pontoon/sync/vcs/models.py
@@ -177,7 +177,7 @@ class VCSProject(object):
 
     @cached_property
     def changed_files(self):
-        if self.force:
+        if self.force or self.changed_config_files:
             # All files are marked as changed
             return None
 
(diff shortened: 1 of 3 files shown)


pontoon/sync/vcs/models.py
```diff
@@ -556,6 +556,25 @@ def resource_paths_without_config(self):
...
+        if not self.db_project.configuration_file:
+            return False
+
+        config_files = set(
+            pc.path.replace(os.path.join(self.source_directory_path, ""), "")
+            for pc in self.configuration.parsed_configuration.configs
+        )
+        changed_files = set(self.changed_source_files[0])
+
+        if changed_files & config_files:
```
mathjazz, line 573: Ohhhhh, now I see what you did here: `&` causes the result to be an [intersection](https://docs.python.org/2/library/sets.html#set-objects) of the two sets!

Can you please use `a.intersection(b)` or at least add a comment? I find this syntax very confusing for beginners like myself. :-)

mathjazz: I would do this to be more explicit:
```python
changed_config_files = changed_files.intersection(config_files)

if changed_config_files:
    return True

return False
```

mathjazz (commented):
Nice work!

Could you add a unit test?

I think this will fix not one, not two, but three bugs!
https://bugzilla.mozilla.org/show_bug.cgi?id=1602896
https://bugzilla.mozilla.org/show_bug.cgi?id=1610284
https://bugzilla.mozilla.org/show_bug.cgi?id=1605421

mathjazz (approved):
This is big!

Excellent work!

PR #1751, opened 2020-11-25
Fix bug 1633148: Support alternative file extenions on upload

mathjazz:
XLIFF file format supports .xliff and .xlf extensions. Previously, we only support the extension of the target file.

Also, due to bug 1530139, projects using TOML configuration files require Gettext files to use the .pot extension on upload, which are used for templates (source files). With this fix, .po (used for l10n files is also supported).

### DIFF
--- a/pontoon/base/forms.py
+++ b/pontoon/base/forms.py
@@ -12,8 +12,8 @@ from pontoon.base.models import (
     User,
     UserProfile,
 )
+from pontoon.sync.formats import are_compatible_formats
 from pontoon.teams.utils import log_group_members
-from pontoon.sync.formats import SUPPORTED_FORMAT_PARSERS
 
 
 class HtmlField(forms.CharField):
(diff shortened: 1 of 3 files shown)


pontoon/base/forms.py
```diff
@@ -65,16 +65,13 @@ def clean(self):
...
             if part:
-                file_extension = os.path.splitext(uploadfile.name)[1].lower()
-                part_extension = os.path.splitext(part)[1].lower()
-
-                # For now, skip if uploading file while using subpages
-                if (
-                    part_extension in SUPPORTED_FORMAT_PARSERS.keys()
-                    and part_extension != file_extension
-                ):
+                uploadfile_ext = os.path.splitext(uploadfile.name)[1].lower()
```
jotes, line 68: [nitpick/advice] In Python 3, you can use [Pathlib](https://docs.python.org/3/library/pathlib.html) objects.
```
Path(uploadfile.name).suffix
```

jotes (approved):

PR #1752, opened 2020-11-27
Fix bug 1678452: Fix Django deprecation warnings

PFischbeck:
This fixes all Django deprecation warnings caused by our code. It is split up into 3 commits, which you can review individually.

The largest change concerns all `urls.py` files. The previously used `django.conf.urls.url` is deprecated, and almost all of our regex uses are now replaced by explicit path strings. In the process, I greatly simplified the regex usage.

Note that the previous GraphQL regex made the API available at any URL starting with `/graphql` (including `/graphql/`), but now it is only exactly that URL.

### DIFF
--- a/pontoon/api/urls.py
+++ b/pontoon/api/urls.py
@@ -1,7 +1,7 @@
 from csp.decorators import csp_exempt
 from graphene_django.views import GraphQLView
 
-from django.conf.urls import url
+from django.urls import path
 from django.views.decorators.csrf import csrf_exempt
 
 from pontoon.api.schema import schema
@@ -10,8 +10,8 @@ from pontoon.settings import DEV
 
 urlpatterns = [
     # GraphQL endpoint. In DEV mode it serves the GraphiQL IDE if accessed with Accept: text/html
-    url(
-        r"^graphql",
+    path(
+        "graphql",
         csp_exempt(csrf_exempt(GraphQLView.as_view(schema=schema, graphiql=DEV))),
     ),
 ]
--- a/pontoon/localizations/urls.py
+++ b/pontoon/localizations/urls.py
@@ -1,4 +1,4 @@
-from django.conf.urls import url
+from django.urls import include, path, re_path
 
 from pontoon.projects import views as projects_views
 from pontoon.teams import views as teams_views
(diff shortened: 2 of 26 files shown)


mathjazz (commented):
Wow, great work - especially with all the URL pattern changes!

I'm still testing this, but I've noticed some additional deprecation warnings in Travis. I wonder if that's expected.

Thanks for another substantial improvement to Pontoon codebase!

PFischbeck:
> I'm still testing this, but I've noticed some additional deprecation warnings in Travis. I wonder if that's expected.

As far as I can tell, all remaining Django deprecation warnings are caused by dependencies (e.g., django-allauth), so these will have to be fixed upstream at some point.

pontoon/localizations/urls.py
```diff
@@ -1,67 +1,68 @@
...
+                    views.localization,
+                    name="pontoon.localizations.project-info",
+                ),
+                # Team info
+                path(
+                    "team-info/",
+                    views.localization,
+                    name="pontoon.localizations.team-info",
+                ),
+                # AJAX view: Localization resources
```
mathjazz, line 37: Should we add another nesting level for the AJAX views, like in the `projects` module?

pontoon/api/urls.py
```diff
@@ -10,8 +10,8 @@
 
 urlpatterns = [
     # GraphQL endpoint. In DEV mode it serves the GraphiQL IDE if accessed with Accept: text/html
-    url(
-        r"^graphql",
+    path(
+        "graphql",
```
mathjazz, line 14: Apparently this change breaks URLs like https://pontoon.mozilla.org/graphql/.

I guess if we add slash at the end, both https://pontoon.mozilla.org/graphql and https://pontoon.mozilla.org/graphql/ will work, which would also be consistent with other URL patters across the app.

pontoon/teams/urls.py
```diff
@@ -1,78 +1,68 @@
...
+                path("permissions/", views.team, name="pontoon.teams.permissions",),
+                # Legacy url for permissions management
+                path(
+                    "manage/",
+                    RedirectView.as_view(
+                        url="/%(locale)s/permissions/", permanent=True
+                    ),
+                    name="pontoon.teams.manage",
+                ),
+                # AJAX view: Team projects
```
mathjazz, line 37: Should we add another nesting level for the AJAX views, like in the projects module?

mathjazz (commented):
Thanks for clarifying it. Makes sense.

And thanks again for this great work. Our URL patterns are now much more readable.

I only left one question and 2 minor suggestions.

pontoon/localizations/urls.py
PFischbeck: Good idea, done!

pontoon/teams/urls.py
PFischbeck: Done

pontoon/api/urls.py
PFischbeck: If we have the slash at the end, requests to https://pontoon.mozilla.org/graphql would be redirected to https://pontoon.mozilla.org/graphql/, which might break existing workflows/scripts. As far as I can tell, the URL endpoint without the slash is canonical for GraphQL (e.g., see the [graphene-django docs](https://docs.graphene-python.org/projects/django/en/latest/installation/#id1)), and is already used in scripts (e.g., [here](https://github.com/flodolo/scripts/blob/954fa857af8422b71662470106415924e4524800/mozilla_l10n/cldr_data/cldr_coverage.py)).
Is https://pontoon.mozilla.org/graphql/ actually used, except for the two JS files in the frontend that I modified?

I see two options here:
1. https://pontoon.mozilla.org/graphql is the only endpoint: `path("graphql", ...)`
2. Both versions with and without slash reach the endpoint: `re_path(r"^graphql/?$", ...)` 

What do you think?

mathjazz: Thanks for clarifying!

Another important user of the API is the Pontoon Add-on, which also doesn't use the slash:
https://github.com/MikkCZ/pontoon-addon/blob/d8d023788444a698f0cab37ac8d41caecd3aa3d9/src/packages/background/src/RemotePontoon.js#L139

Stas also used links without slash in the blog post:
https://blog.mozilla.org/l10n/2017/10/10/exposing-pontoon-data-through-api/

Since we don't have a lot of other API consumers ATM, let's move on with this change.

mathjazz (approved):
Excellent work!

PR #1753, opened 2020-11-27
Fix bug 1633149: Detect changes to translated resources on sync

mathjazz:
When detecting changes to translated resources in VCS, we need to identify which of the changed paths belongs to a localizable resource (and ignore others).

The logic that does that makes assumptions about the file and folder structure which are not valid for projects using project configuration. Hence this patch adds a new logic for such projects, which makes use of the ProjectFiles matcher of the compare-locales library.

The logic also converts the changed paths to the canonical form for representing resources used across Pontoon codebase (reference path insted of l10n path) and assigns the list of changed locales to each path.

As a side effect, this patch also fixes [bug 1608045](https://bugzilla.mozilla.org/show_bug.cgi?id=1608045), which is another symptom of the very same problem.

### DIFF
--- a/pontoon/sync/vcs/models.py
+++ b/pontoon/sync/vcs/models.py
@@ -265,21 +265,13 @@ class VCSProject(object):
                     )
                 )
 
-                # Find relevant changes in repository by matching changed
-                # paths against locale repository paths
-                locale_path_locales = self.locale_path_locales(repo.checkout_path)
-                locale_paths = locale_path_locales.keys()
-
-                for path in changed_files:
-                    if is_hidden(path):
-                        continue
-
-                    for locale_path in locale_paths:
-                        if path.startswith(locale_path):
-                            locale = locale_path_locales[locale_path]
-                            path = path[len(locale_path) :].lstrip(os.sep)
-                            files.setdefault(path, []).append(locale)
-                            break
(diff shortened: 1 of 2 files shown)


pontoon/sync/vcs/models.py
```diff
@@ -311,6 +303,49 @@ def changed_config_files(self):
...
+        for path in paths:
+            if is_hidden(path):
+                continue
+
+            for locale_path in locale_paths:
+                if path.startswith(locale_path):
+                    locale = locale_path_locales[locale_path]
+                    path = path[len(locale_path) :].lstrip(os.sep)
+                    files.setdefault(path, []).append(locale)
+                    break
```
mathjazz, line 345: This is the existing code, not changed by this patch.

pontoon/sync/vcs/models.py
```diff
@@ -665,6 +700,15 @@ def l10n_path(self, locale, reference_path):
         m = project_files.match(reference_path)
         return m[0] if m is not None else None
 
+    def reference_path(self, locale, l10n_path):
```
mathjazz, line 703: This is the opposite function of `VCSConfiguration.l10n_path()`.

jotes (approved):

PR #1754, opened 2020-11-28
Fix Bug 1679654 - Add ability to reset machinery search field

abowler2:
The ability to reset the Machinery Search input to its original state is not currently intuitive (user must clear the input and hit enter). 

This PR follows the [specification for Concordance Search](https://github.com/mozilla/pontoon/blob/master/specs/0106-concordance-search.md) and switches the 'search' icon to a 'close' button (`x`) when a search is made. When the 'close' button is clicked the search input will be cleared and returned to its original state.

This does not currently work for the FTL files due to [Bug 1676750](https://bugzilla.mozilla.org/show_bug.cgi?id=1676750). The input is cleared but the original state is not restored.

### DIFF
--- a/frontend/src/modules/machinery/components/Machinery.css
+++ b/frontend/src/modules/machinery/components/Machinery.css
@@ -23,6 +23,12 @@
     top: 16px;
 }
 
+.machinery > .search-wrapper > label > button {
+    background-color: transparent;
+    border: none;
+    color: #aaaaaa;
+}
+
 .machinery > .search-wrapper input[type='search'] {
     background: #333941;
     border: none;
--- a/frontend/src/modules/machinery/components/Machinery.css
+++ b/frontend/src/modules/machinery/components/Machinery.css
@@ -27,6 +27,7 @@
     background-color: transparent;
     border: none;
     color: #aaaaaa;
+    padding: 0 4px;
 }
 
 .machinery > .search-wrapper input[type='search'] {
--- a/frontend/src/modules/machinery/components/Machinery.js
+++ b/frontend/src/modules/machinery/components/Machinery.js
@@ -55,7 +55,6 @@ export default class Machinery extends React.Component<Props> {
     }
 
     handleResetSearch = () => {
-        this.searchInput.current.value = '';
         this.props.searchMachinery('');
     };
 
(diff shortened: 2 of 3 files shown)


frontend/src/modules/machinery/components/Machinery.css
```diff
@@ -23,6 +23,12 @@
     top: 16px;
 }
 
+.machinery > .search-wrapper > label > button {
+    background-color: transparent;
+    border: none;
+    color: #aaaaaa;
```
mathjazz, line 29: Seems like `<button>` gets some padding assigned (although I don't know from where?), which moves the times icon towards left compared to the magnifier icon.

Adding `padding: 0 4px;` here should fix the unexpected jump.

frontend/src/modules/machinery/components/Machinery.js
```diff
@@ -54,6 +54,12 @@ export default class Machinery extends React.Component<Props> {
         }
     }
 
+    handleResetSearch = (event: SyntheticMouseEvent<>) => {
+        event.preventDefault();
+        this.searchInput.current.value = '';
+        this.props.searchMachinery(this.searchInput.current.value);
```
mathjazz, line 60: I'd expect `this.props.searchMachinery('');` to be sufficient here, because `componentDidUpdate` should set `this.searchInput.current.value` accordingly.

And I don't think we need to prevent default?

mathjazz (commented):
Great job!

Works like a charm.

Just two nits / questions I found along the way.

frontend/src/modules/machinery/components/Machinery.js
abowler2: Yep, much cleaner to just set with an empty string.

I wasn't sure about the prevent default either. I added it to protect against an action that I might have overlooked, but it does not appear to be needed.

abowler2:
I just noticed that when switching tabs the search term still stays in the form as it should but the 'close' button is no longer present.  I will work on fixing that.

frontend/src/modules/machinery/components/Machinery.js
```diff
@@ -72,12 +77,23 @@ export default class Machinery extends React.Component<Props> {
...
+        const showResetButton = !machinery.entity && machinery.sourceString;
+
         return (
             <section className='machinery'>
                 <div className='search-wrapper clearfix'>
-                    <label htmlFor='machinery-search'>
-                        <div className='fa fa-search'></div>
-                    </label>
+                    {showResetButton ? (
+                        <label htmlFor='machinery-search'>
```
mathjazz, line 86: Looks like the `<label>` element is the same, so you can move the condition inside of it.

mathjazz (commented):
Thanks for the update and spotting the issue the review hasn't. :-)

A test would be nice.

abowler2:
Updated based on comments and added a test.

Not sure what is happening with the CI tests though. Locally the tests pass. I rebased to ensure I had the latest changes to `master` but it is still failing unrelated to my changes.

mathjazz (approved):
Great job!

That Travis error is unrelated, we can tackle it separately if it resurfaces.

PR #1756, opened 2020-12-03
Fix bug 1680429: Update compare-locales and fluent

mathjazz:
Updating compare-locales requires updating python-fluent. I also updated frontend Fluent libraries.

I also introduced the temporary fix for the `pip` issue in order to be able to install libraries.


Pike (approved):
lgtm.

mathjazz:
That's not fair. I was just about to flag you for a review! 😅

PR #1757, opened 2020-12-03
Fix bug 1646010 - Correct various bugs around the Fluent editor

adngdb:
This fixes the following problems:

- The content becoming empty when clicking on the selected entity in the list.
- The content being empty in the next entity after approving.
- The content not being correctly parsed when using the Copy button.

### DIFF
--- a/frontend/src/core/editor/actions.js
+++ b/frontend/src/core/editor/actions.js
@@ -265,6 +265,7 @@ export function sendTranslation(
 
             if (nextEntity) {
                 // The change did work, we want to move on to the next Entity or pluralForm.
+                dispatch(reset());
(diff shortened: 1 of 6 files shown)


mathjazz (commented):
Nice work!

I found a regression in the last(!) string I tested: seems like strings that need to be opened in the source editor ([example](https://github.com/mozilla-l10n/pontoon-ftl/blob/master/en-US/demo.ftl#L110)) open in generic editor by default. Or at least the FTL button doesn't get highlighted.

adngdb:
@mathjazz good catch! Should be fixed now. :)

mathjazz (approved):
Well done!

PR #1759, opened 2020-12-06
Ignore Prettier formatting commits on blame

abowler2:
Added the Prettier commits and some comments to the git blame ignore file


mathjazz (approved):
Thanks! 👍

PR #1760, opened 2020-12-08
Fix bug 1680053: Create data model for storing community health data

mathjazz:
This is [Step 1/4](https://wiki.mozilla.org/L10n:Pontoon_Roadmap#Community_Health_Dashboard) of the [Community Health Dashboard](https://github.com/mozilla/pontoon/blob/fddf18f45546cb3c9d59285437e18c72de7ed577/specs/0108-community-health-dashboard.md) implementation. It adds data model for storing insights, and a utility function for retrieving them in the format required by the dashboard. We'll populate the model and create the actual dashboard in the following steps.

To make the review less abstract, I've included a commit which populates the data model with fake data and renders it on the dashboard. Most of the dashboard code was created as part of prototyping anyways.

I deployed that commit [to stage](https://mozilla-pontoon-staging.herokuapp.com/en-GB/insights/). You can also test the dashboard locally at http://localhost:8000/en-GB/insights/ if you checkout commit 1cfd6cc. Note that you'll need to run migrations (e.g. via `make build`) and that the data is only available for the en-GB locale.

**Finally, I've reverted that extra commit to make it clear that commit 1cfd6cc should NOT be reviewed.**


pontoon/base/utils.py
```diff
@@ -429,6 +422,11 @@ def handle_upload_content(slug, code, part, f, user):
         changeset.translations_to_create[-1].update_latest_translation()
 
 
+def aware_datetime(*args, **kwargs):
+    """Return an aware datetime using Django's configured timezone."""
+    return timezone.make_aware(datetime(*args, **kwargs))
+
+
```
mathjazz, line 429: Moving all date/time helpers together in the file.

pontoon/insights/models.py
```diff
@@ -0,0 +1,47 @@
...
+    active_reviewers = models.ManyToManyField(
+        User, related_name="active_reviewers_insights"
+    )
+    all_reviewers = models.ManyToManyField(User, related_name="all_reviewers_insights")
+    active_contributors = models.ManyToManyField(
+        User, related_name="active_contributors_insights"
+    )
+    all_contributors = models.ManyToManyField(
+        User, related_name="all_contributors_insights"
+    )
```
mathjazz, line 25: Active users data is different from all the other data in that we can't simply aggregate daily numbers and use them as the input for the charts. We need to observe the entire period (in our case 12 months) and count how many different users did we have in each of the observed groups.

Hence the use of the `many-to-many` relationship with the User model. Which is tedious, because we're saving bug amounts of data in order to get to 6 numbers eventually and because it adds complexity to `get_insights()`.

I'm sort of hoping to be able remove these fields completely as part of [bug 1680054](https://bugzilla.mozilla.org/show_bug.cgi?id=1680054) if it turns out we can calculate the numbers on the fly (on each dashboard load) quickly enough.

mathjazz:
@Pike @adngdb @jotes Anyone willing to review this? :)

Pike:
> @Pike @adngdb @jotes Anyone willing to review this? :)

Not really in the loop on this one myself. I glanced at it, and noticed that the reverse names in `models.py` seem to be off?

A note on the UI prototype, I don't think the licensing is solid enough for Pontoon. In particular for 3rd-party deployments.

mathjazz:
> Not really in the loop on this one myself. I glanced at it, and noticed that the reverse names in `models.py` seem to be off?

Thank you for your honesty. :)

Reverse names are required, but I'm struggling to come up with good names. Any suggestions? I was also thinking about just using `+`.

> A note on the UI prototype, I don't think the licensing is solid enough for Pontoon. In particular for 3rd-party deployments.

Phew, that's a surprise. We already use the same library in frontend, so we'll need to tackle that.

Pike:
> Reverse names are required, but I'm struggling to come up with good names. Any suggestions? I was also thinking about just using `+`.

Sorry, I got confused by the patch. Stupid black formatting each line the other way than the one before. I had seen that as inconsistent naming, and it was just the formatting distracting me.

pontoon/insights/models.py
mathjazz: > I'm sort of hoping to be able remove these fields completely as part of [bug 1680054](https://bugzilla.mozilla.org/show_bug.cgi?id=1680054) if it turns out we can calculate the numbers on the fly (on each dashboard load) quickly enough.

It turns out we should still keep the numbers in the DB, but I was able to greatly simplify the data model by storing active user numbers for various periods separately.

pontoon/insights/utils.py
```diff
@@ -0,0 +1,120 @@
...
+        active_reviewers = active_users["reviewers"]
+        active_contributors = active_users["contributors"]
+
+    # Unreviewed suggestions lifespan
+    unreviewed_lifespans = [
+        x["unreviewed_suggestions_lifespan_avg"].days for x in insights
+    ]
+
+    # Translation activity
+    completion = [round(float(x["completion_avg"])) for x in insights]
```
jotes, line 86: I can't find any place that uses  `get_insights` function. Do you have a specific use-case in mind when you think about the results of this function? I see that it returns the lists of the values instead of their aggregates (?). Also, I feel that a large portion of this function could be handled in a database query (?)

pontoon/insights/utils.py
```diff
@@ -0,0 +1,120 @@
...
+    # Active users
+    total_managers = 0
+    total_reviewers = 0
+    total_contributors = 0
+    active_managers = 0
+    active_reviewers = 0
+    active_contributors = 0
+
+    if snapshots:
+        latest_snapshot = snapshots.latest("created_at")
```
jotes, line 69: How often Pontoon is going to create insight snapshots? I ponder over the reasons behind the denormalization.

pontoon/insights/utils.py
mathjazz: We use the ` get_insights` function when we [render the dashboard](https://github.com/mozilla/pontoon/pull/1760#issue-534386013). Check out the reverted "[DO NOT REVIEW]" commit.

We plot some numbers (like `completion`) in a chart over time, so we need a list of items, each representing a point in chart (e.g. a month). See the [spec mockup](https://raw.githubusercontent.com/mozilla/pontoon/master/specs/0108/mockup.png). I currently use stage to test my next patch, so it might be less reliable.

I didn't find an elegant way to get to the final results via a DB query directly.

pontoon/insights/utils.py
mathjazz: Once a day.

It's almost impossible to get the exact numbers for the period of the past 12 months from a certain point in time. We simply don't collect all the required information today to tell you what was the overall completion of German for example.

We'll still try to do that as part of https://bugzilla.mozilla.org/show_bug.cgi?id=1680055, but those numbers will be as good estimates as we'll be able to make. They will also be pushed off the charts after 365 days.

pontoon/insights/utils.py
mathjazz: We could make the code following the `QuerySet` more condensed:
https://gist.github.com/mathjazz/3dc4ba0fbaeb545d32897a05f12629bb

jotes (approved):

PR #1767, opened 2020-12-10
FIx Bug 1676750: [FTL] Unable to get back to the initial Machinery state after search

Mitch4sho:
Fixed the initial machinery state by added in function to optimize the FTL files


mathjazz (approved):
Excellent job!

Mitch4sho:
I'm embarrass to say this took me way longer than it should to figure out 😅

PR #1769, opened 2020-12-11
Revert mistakenly pushed commits

mathjazz:
This reverts commits:
- 6f82b65d91ac13b18f460ad141ace6465f875d91
- 37c65ca55fa78b0dec1327a7e0639a6f6c570631

Since the latter is a merge commit, it takes us to the state of commit 5084c1cf979aec538c8d2ace0902de04638e60f2.


mathjazz:
@Pike r?

Pike (approved):
I wonder if blame would go easier if this was a merge commit, but  as I'm expecting pretty similar code to land again, there shouldn't be much of a  difference.

PR #1770, opened 2020-12-12
Fix bug 1682122: Validate GET parameters that need to be integers

mathjazz:
Passing GET parameters of wrong type can sometimes result in server errors. That happens particularly often when crawlers or attackers make requests to the views which take integers as inputs (e.g. IDs), causing ValueErrors.

Some of our views have protection against that, but many don't. This patch fixes that by validating GET parameters that need to be integers.

### DIFF
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -342,9 +342,9 @@ def _serialize_translation_values(query):
 def get_translations_from_other_locales(request):
     """Get entity translations for all but specified locale."""
     try:
-        entity = request.GET["entity"]
+        entity = int(request.GET["entity"])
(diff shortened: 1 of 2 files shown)


mathjazz:
self-r+

Pushing this to prod since we're getting a lot of these errors right not.

Pike:
The changes looked good to me, the only question I have is if they're exhaustive?

I found at least some `cleaned_data` access, where I haven't figured out where the errors would go

mathjazz:
It should be exhaustive for the integers.

We probably still have a bug with the errors the `split_ints` produces - we might be able to use `validate_comma_separated_integer_list` there.

PR #1771, opened 2020-12-13
Fix Bug 1679709 - Clicking on each machinery search result adds the text to the editor

abowler2:
Currently when searching within the Machinery tab, the results will replace the contents of the editor when clicked on.  As per the [concordance search spec](https://github.com/mozilla/pontoon/blob/master/specs/0106-concordance-search.md), the search results should just be added instead.  This patch makes that change.

### DIFF
--- a/frontend/src/modules/machinery/components/Translation.js
+++ b/frontend/src/modules/machinery/components/Translation.js
@@ -59,11 +59,14 @@ export default function Translation(props: Props) {
         // added to the editor instead of replacing the editor content
         if (!entity) {
             addTextToEditorTranslation(translation.translation);
-        } else if (typeof editorContent !== 'string') {
-            // This is a Fluent Message, thus we are in the RichEditor.
-            // Handle machinery differently.
+        }
+        // This is a Fluent Message, thus we are in the RichEditor.
+        // Handle machinery differently.
+        else if (typeof editorContent !== 'string') {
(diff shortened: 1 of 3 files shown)


frontend/src/modules/machinery/components/Translation.js
```diff
@@ -53,7 +55,9 @@ export default function Translation(props: Props) {
             return;
         }
 
-        if (typeof editorContent !== 'string') {
+        if (!entity) {
+            addTextToEditorTranslation(translation.translation);
```
mathjazz, line 59: Since this is a special case, it would be nice to add a short comment on top explaining why we are doing it.

mathjazz (commented):
Great work! I'd just add a unit test which checks the right method is called.

frontend/src/modules/machinery/components/Translation.js
```diff
@@ -53,7 +55,11 @@ export default function Translation(props: Props) {
             return;
         }
 
-        if (typeof editorContent !== 'string') {
+        // If there is no entity then it is a search term and it is
+        // added to the editor instead of replacing the editor content
```
mathjazz, line 59: Übernit: could you please make sure this comment and the next one both use the same indentation level?

mathjazz (approved):
Great work! Thanks!

PR #1773, opened 2020-12-14
Fix bug 1680054: Regularly collect data needed for community health dashboard

mathjazz:
This is [Step 2/4](https://wiki.mozilla.org/L10n:Pontoon_Roadmap#Community_Health_Dashboard) of the [Community Health Dashboard](https://github.com/mozilla/pontoon/blob/fddf18f45546cb3c9d59285437e18c72de7ed577/specs/0108-community-health-dashboard.md) implementation.

It adds a scheduled job, which will populate the data model introduced in f2258ed4aba57731983b5e126cc8b3b10e351c29 every day. The job takes around 15 seconds to run on stage. Historic data for the past year needed to plot over-the-time charts before enough data is collected using this job will be added as part of [bug 1680055](https://bugzilla.mozilla.org/show_bug.cgi?id=1680055).

Deployment docs have been updated to cover the new scheduled job and the not yet documented `send_deadline_notifications`. A few smaller fixes discovered during development have been added in a separate commit.

### DIFF
--- a/docs/admin/deployment.rst
+++ b/docs/admin/deployment.rst
@@ -280,6 +280,10 @@ RabbitMQ add-on:
 
 Scheduled Jobs
 --------------
+Pontoon requires several scheduled jobs to run regularly.
+
+Sync projects
+~~~~~~~~~~~~~
 While internal Pontoon DB can be used for storing localizable strings, Pontoon
 specializes in using version control systems for that purpose. If you choose
 this option as well, you'll need to run the following scheduled job:
@@ -291,6 +295,37 @@ this option as well, you'll need to run the following scheduled job:
 It's recommended to run this job at least once an hour. It commits any string
 changes in the database to the remote VCS servers associated with each project,
 and pulls down the latest changes to keep the database in sync.
+
+Send Deadline Notifications
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Pontoon allows you to set deadlines to projects. This job sends deadline
+reminders to contributors of projects when they are due in 7 days. If 2 days
+before the deadline project still isn't complete for the contributor's locale,
+notifications are sent again. The command is designed to run daily.
+
+.. code-block:: bash
+
+   ./manage.py send_deadline_notifications
+
+Collect insights
+~~~~~~~~~~~~~~~~
+The Insights tab in the dashboards presents data that cannot be retrieved from
+the existing data models efficiently upon each request. This job gathers all
+the required data and stores it in a dedicated denormalized data model. The job
+is designed to run in the beginning of the day, every day.
+
+.. code-block:: bash
+
+   ./manage.py collect_insights
+
+Sync Log Retention
+~~~~~~~~~~~~~~~~~~
+You may also optionally run the ``clear_old_sync_logs`` management command on a
+schedule to remove sync logs from the database that are over 90 days old:
+
+.. code-block:: bash
+
+   ./manage.py clear_old_sync_logs
 
 Provisioning Workers
 ~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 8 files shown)


flodolo (commented):
Just by passing by and taking a look

tenor

docs/admin/deployment.rst
```diff
@@ -314,6 +327,19 @@ schedule to remove sync logs from the database that are over 90 days old:
 
    ./manage.py clear_old_sync_logs
 
+Provisioning Workers
```
jotes, line 330: <3

pontoon/insights/tasks.py
```diff
@@ -0,0 +1,391 @@
...
+
+log = logging.getLogger(__name__)
+
+
+@shared_task(bind=True)
+def collect_insights(self):
+    """
+    Gather data needed for the Insights tab and store it in the DB.
+    """
+    start_of_today = (
```
jotes, line 23: [nitpick] I'm not sure about the types in the rest of the patch, but `timezone.now().date()` could produce the same results.

pontoon/insights/tasks.py
```diff
@@ -0,0 +1,391 @@
...
+    total_reviewers = len(all_reviewers)
+    total_contributors = len(all_contributors)
+
+    active_users_last_month = get_active_users_data(
+        start_of_today,
+        privileged_users,
+        active_users_actions,
+        all_managers,
+        all_reviewers,
+        days=365 / 12,
```
jotes, line 191: I'm not sure about the required accuracy of the Insights dashboard, but the number of days is going to be wrong for leap years. Also, the number of days in a month changes too. The same comment applies to the calculation above.

pontoon/insights/tasks.py
```diff
@@ -0,0 +1,391 @@
...
+        new_source_strings=len(entities),
+        # Review activity
+        peer_approved=len(peer_approved),
+        self_approved=len(self_approved),
+        rejected=len(rejected),
+        new_suggestions=len(new_suggestions),
+    )
+
+
+def get_all_users_data(privileged_users):
```
jotes, line 264: [nitpick] `get_privileged_users_data` ?

pontoon/insights/tasks.py
```diff
@@ -0,0 +1,391 @@
...
+        )
+    )
+
+    return group_dict_by(privileged_users, "pk")
+
+
+def get_contributors():
+    """Get all contributors without system users.
+
+    Note that excluding system users directly in the QuerySet is slow.
```
jotes, line 78: I wonder what's the reason behind that issue -> in theory, it shouldn't be a performance problem when you don't use a subquery. 
Is there a difference between `exclude(pk__in=list(excluded_users))` and `exclude(pk__in=excluded_users)`?

jotes (changes requested):

pontoon/insights/tasks.py
mathjazz: I feel stupid.

mathjazz: Oh wait, that results in the naive `datetime`. It should be timezone aware instead.

pontoon/insights/tasks.py
mathjazz: I'll see if I can use `relativedelta` instead.

pontoon/insights/tasks.py
mathjazz: What I mean to say is that this would be slow:

```python
contributors = (
    Translation.objects
    .filter(user__isnull=False)
    .exclude(user__email__regex=r"^pontoon-(\w+)@example.com$")
    .values("locale", "user")
    .distinct()
)
```

I assume that's because all translations are joined with users.

I simplified the query with your approach.

pontoon/insights/tasks.py
jotes: I'm sorry, I thought that Django somehow handles that case.

jotes (approved):

PR #1774, opened 2020-12-16
Title missing in Terms panel and popup

codershona:
# Fixed Bugs: Bug 1679565 
## Bug Name : Title missing in Terms panel and popup
**_(BUGZILLA)_**
### Explanation: I had added title string into term panel and added in python files also.

### DIFF
--- a/frontend/src/core/api/types.js
+++ b/frontend/src/core/api/types.js
@@ -64,6 +64,7 @@ export type UsersList = {|
  * Term entry with translation.
  */
 export type TermType = {|
+    +title: string,
     +text: string,
     +partOfSpeech: string,
     +definition: string,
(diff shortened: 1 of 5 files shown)


mathjazz (changes requested):
Thanks for the patch.

Sadly it doesn't solve the problem. The gist of the bug is to add a "Copy Into Translation" tooltip for each term on the list.

For inspiration, see:
https://github.com/mozilla/pontoon/blob/master/frontend/src/modules/machinery/components/Translation.js#L80
https://github.com/mozilla/pontoon/blob/master/frontend/public/static/locale/en-US/translate.ftl#L456

mathjazz:
Closing since it doesn't solve the problem. Feel free to open a new PR.

PR #1775, opened 2020-12-17
Update Dockerfile

quimicefa:
we're running from debian buster, so we need to use nodeJS buster repos (not jessie)


mathjazz (approved):
LGTM

Thanks for the patch!

jotes (approved):

PR #1776, opened 2020-12-17
Fix bug 1680056: Implement community health dashboard frontend

mathjazz:
This is [Step 4/4](https://wiki.mozilla.org/L10n:Pontoon_Roadmap#Community_Health_Dashboard) of the [Community Health Dashboard](https://github.com/mozilla/pontoon/blob/fddf18f45546cb3c9d59285437e18c72de7ed577/specs/0108-community-health-dashboard.md) implementation.

It adds the frontend code: the actual dashboard with charts that visualize the data collected and stored in previous steps. Two changes have been made that are not part of the spec:

1. The `ENABLE_INSIGHTS_TAB` setting has been introduced and documented to enable the Insights tab. It's enabled by default in dev environment and disabled in prod. That's because `collect_insights` job needs to run in order to be able to show any data. With the new setting admins have the ability to only enable the feature when they start collecting data.

1. Ability to change time periods (1, 3, 6 or 12 months) has only been added to the Active users panel (instead of all panels). That's because other panels show data over a period of 12 months, and the 6-, 3- or 1- month periods are already seen. In the future we might add the ability to show last month in other panels as well, but we need to first collect enough data and see if there's an actual need.

The patch is deployed to stage without the last commit (which reverts the previous commit with fake data hardcoded).

### DIFF
--- a/pontoon/insights/templates/insights/widgets/insights.html
+++ b/pontoon/insights/templates/insights/widgets/insights.html
@@ -1,3 +1,5 @@
+{% import "insights/widgets/tooltip.html" as Tooltip %}
+
 {# Widget to display insights. #}
 {% macro display() %}
 <div id="insights" data-dates="{{ dates }}">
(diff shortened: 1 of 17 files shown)


pontoon/insights/utils.py
```diff
@@ -54,17 +54,28 @@ def get_insights(query_filters=None):
...
-        "total_contributors": latest.total_contributors if latest else 0,
-        "active_managers": active_users["managers"] if active_users else 0,
-        "active_reviewers": active_users["reviewers"] if active_users else 0,
-        "active_contributors": active_users["contributors"] if active_users else 0,
+        "total": {
+            "managers": latest.total_managers if latest else 0,
+            "reviewers": latest.total_reviewers if latest else 0,
+            "contributors": latest.total_contributors if latest else 0,
+        },
+        "active_users_last_month": latest.active_users_last_month
```
jotes, line 67: [nitpick] `black` formats this part in a weird way. Also, Subjectively, I see code repetition in the conditions like `if latest`. Maybe it would be better to move the initialization of the default values outside of this dict, and create a separate block? e.g.
```
if latest:
   results['total'] = {}
else:
   results['total'] = {}
```

pontoon/insights/templates/insights/widgets/insights.html
```diff
@@ -0,0 +1,172 @@
...
+
+    <section class="translation-activity">
+        <figure class="block">
+            <h3 class="controls">
+                Translation activity
+                <div class="fa fa-info selector"></div>
+                <div class="tooltip">
+                    <p>Impact of adding translations and source strings on the overal translation completion.</p>
+                    <ul>
+                        <li class="completion">
```
jotes, line 95: [nitpick] What do you think about using macros for the columns like that? They could reduce the size of this html.

mathjazz: Good point! I should add macros for tooltips, and probably also for the three active users "pie" charts.

mathjazz:
I pass variable directly to templates now (without the high-level `data` variable).

I've added tooltip and active users chart macros.

I've also reduced the number of `if latest` repetitions, although I'm not sure this is what you had in mind.

We should probably also use the JSON field for the `total_users` and format it the same as various `active_users` fields, but I haven't made that change (yet).

jotes (approved):

jotes (approved):

PR #1777, opened 2020-12-18
Bug 1679904 Add a Django view that handles Concordance search.

jotes:


pontoon/base/utils.py
```diff
@@ -573,3 +577,37 @@ def readonly_exists(projects, locale):
     return ProjectLocale.objects.filter(
         project__in=projects, locale=locale, readonly=True,
     ).exists()
+
+
+def get_search_phrases(search):
+    """
+    Split the search phrase into separate search queries, and assign the database collation.
```
mathjazz, line 584: We're not assigning database collation here, are we?

pontoon/machinery/utils.py
```diff
@@ -56,18 +59,37 @@ def get_google_translate_data(text, locale_code):
...
+    search_list = base.utils.get_search_phrases(text)
+    search_query_list = [(s, locale.db_collation) for s in search_list]
+    search_filters = (
+        Q(target__icontains_collate=(phrase, collation)) | Q(source__icontains=phrase)
+        for phrase, collation in search_query_list
+    )
+    search_filters = (
+        Q(target__icontains_collate=(phrase, collation)) | Q(source__icontains=phrase)
+        for phrase, collation in search_query_list
+    )
```
mathjazz, line 72: I've seen these 4 lines 4 lines ago. :)

pontoon/machinery/utils.py
```diff
@@ -56,18 +59,37 @@ def get_google_translate_data(text, locale_code):
...
+    ).search_result_quality(
+        # Calculate the quality of the results based on the Levenhstein's distance
+        " ".join(search_list),
+        F("target"),
     )
 
-    # Exclude existing entity
     if pk:
-        entries = entries.exclude(entity__pk=pk)
+        results = results.exclude(pk=pk)
```
mathjazz, line 84: No need to exclude the current entity here. We want the behavior to mimic string list search.

pontoon/machinery/utils.py
```diff
@@ -56,18 +59,37 @@ def get_google_translate_data(text, locale_code):
...
 
-    entries = entries.values("source", "target", "quality")
+    entries = results.values("source", "target", "quality").annotate(
+        project_name=F("project__name")
+    )
+    return sort_query_results(entries)
```
mathjazz, line 89: And also no need to group (count) results.

mathjazz (commented):
Nice work! Thanks for the refactoring and docstrings.

Left some comments, but nothing major. Can't wait to take it for a spin!

pontoon/machinery/utils.py
jotes: Yeah, that's why this PR is marked as a WIP :-) I'm going to clean up all of that.

pontoon/machinery/utils.py
```diff
@@ -56,6 +59,29 @@ def get_google_translate_data(text, locale_code):
...
+        Q(Q(target__icontains_collate=(phrase, locale.db_collation)), locale=locale,)
+        for phrase in search_phrases
+    )
+    search_query = reduce(operator.and_, search_filters)
+
+    search_query_results = base.models.TranslationMemoryEntry.objects.filter(
+        search_query
+    ).values_list("target", "project__name")
+
+    search_results = [
```
jotes, line 74: @abowler2 Hey!
Could you take a look and check the fields that are going to be returned in concordance search results? They defined are below.
Ping me on chat if you have any questions :-) Thanks!

pontoon/machinery/utils.py
```diff
@@ -56,6 +59,29 @@ def get_google_translate_data(text, locale_code):
         }
 
 
+def get_concordance_search_results(text, locale):
+    search_phrases = base.utils.get_search_phrases(text)
+    search_filters = (
+        Q(Q(target__icontains_collate=(phrase, locale.db_collation)), locale=locale,)
```
mathjazz, line 65: We should add `| Q(source__icontains=phrase),` here. According to the [spec](https://github.com/mozilla/pontoon/blob/a4f6e09/specs/0106-concordance-search.md#feature-explanation), CS "searches source strings and translations".

pontoon/machinery/utils.py
```diff
@@ -56,6 +59,29 @@ def get_google_translate_data(text, locale_code):
...
+    )
+    search_query = reduce(operator.and_, search_filters)
+
+    search_query_results = base.models.TranslationMemoryEntry.objects.filter(
+        search_query
+    ).values_list("target", "project__name")
+
+    search_results = [
+        {
+            "target": target,
```
mathjazz, line 76: We should also return the `source`.

pontoon/machinery/views.py
```diff
@@ -69,6 +70,27 @@ def translation_memory(request):
...
+        )
+
+    try:
+        locale = Locale.objects.get(code=locale)
+    except Locale.DoesNotExist as e:
+        return JsonResponse(
+            {"status": False, "message": "Not Found: {error}".format(error=e)},
+            status=404,
+        )
+    return JsonResponse(get_concordance_search_results(text, locale), safe=False)
```
mathjazz, line 91: Nitpicking: I'd make the style and naming consistent with `get_translation_memory_data` and `get_google_translate_data`:

```python

data = get_concordance_search_data(text, locale)
return JsonResponse(data, safe=False)
```

You could also rename `*_data` function names to `*_results` at the end.

mathjazz (commented):
This looks great! Well done with the unit tests as well!

The only thing that needs to change is to include the `source` field in search and response.

I also tested performance with prod DB and it's very good, incomparable to the string list. Which is why:
* I don't think we should use a quality threshold.
* We should probably add paging, but it could be done in a followup.
* We should reconsider using a denormalized table for searching and filtering in a string list.

pontoon/machinery/views.py
jotes: I'm going to rename the function to match the convention. I have a question about the last variable before the return statement -> do we use it to only increase the readability?

mathjazz: Yeah. I don't have a strong preference for either style. But I'd prefer consistency.

mathjazz (approved):
Well done!

I'll wait for @abowler2 to comment on the API before merging. As far as I'm concerned this is good to go.

pontoon/machinery/utils.py
abowler2: This looks good to me.

abowler2 (approved):

PR #1779, opened 2020-12-20
Add guide to run pontoon in production on k8s

michaellzc:
Right now this project is lacking proper instruction to run a production-ready instance on platforms other than Heroku.

We are running a production pontoon deployment on k8s ourselves, and figure it would be helpful for those who are also interested in self-hosting pontoon.

We periodically build the docker image from [Dockerfile](https://github.com/mozilla/pontoon/blob/master/docker/Dockerfile) and publish to [Docker Hub](https://hub.docker.com/r/skillsnetwork/pontoon) at https://github.com/ibm-skills-network/pontoon.

We also publish a helm chart that gives users the flexibility to deploy pontoon however they want. https://github.com/ibm-skills-network/charts/tree/master/charts/pontoon


ExiaSR:
> We have [bugzilla.mozilla.org/show_bug.cgi?id=1513104](https://bugzilla.mozilla.org/show_bug.cgi?id=1513104) on the problems we see when using the current development docker image on production installs.

Left a reply with more detail at https://bugzilla.mozilla.org/show_bug.cgi?id=1513104#c7

mathjazz:
As explained in https://github.com/mozilla/pontoon/pull/2320#issuecomment-956472603, our path forward WRT 3rd party deployments needs some clarity before we can move forward with support for Docker. I suggest we close this patch until then.

(Also, there's a lot in this patch that should not land in tree.)

PR #1780, opened 2020-12-21
Bug 1683619 - Move Pontoon automation from Travis CI to GitHub Actions

flodolo:


flodolo:
This is the final product of 75 attempts



You'll notice 3 separate commits.

At first, I tried to follow [GitHub instructions for PostgreSQL](https://docs.github.com/en/free-pro-team@latest/actions/guides/creating-postgresql-service-containers). The first issue was the use of `sslmode`, then the missing collation for `tr_TR`. It might be possible to still use this method, but you'd need a customized docker image.

So, decided to just install PostgreSQL on the image running tests (2nd commit).

Turns out ubuntu-latest already has postgresql-13 (13.1 IIRC) running. I couldn't find a more elegant solution to eradicate it then the current one. If you just remove the package, postgresql will refuse to start because "PostgreSQL 13 is missing" 🙄 
FYI, tests fail badly with PS 13, with errors like: `cannot ALTER TABLE "base_projectlocale" because it has pending trigger events`. Might be worth keeping it in mind for an update from 11.

After discovering that I was running the wrong version of PostgreSQL, I ended up with *only* 4 tests running (3rd commit). I'm not exactly sure why this would happen, but I assume the underlying cause is the same. I'd be curious to understand why they pass on Travis.

flodolo:
I don't think GitHub Actions will run automatically on this PR, probably because of my permissions.

This is the resulting run on my fork
https://github.com/flodolo/pontoon/runs/1587902007?check_suite_focus=true

Pike:
I have tons of thoughts on the actual automation. What really bothers me is that we have to change tests here.
That indicates that the tests here work differently than prod, and that wouldn't be good.

flodolo:
Replaced by #1784

PR #1782, opened 2020-12-21
bug 1683619, move linting into gh actions

Pike:
Refactoring requirements to allow only linting reqs to be installed.
Exposed requirements to docker, so pip-compile can be run there.


.github/workflows/python.yml
```diff
@@ -0,0 +1,44 @@
...
+          python-version: 3.8
+      - name: Install Dependencies
+        run: |
+          pip install -U pip
+          pip install -r requirements/lint.txt
+      - name: flake8
+        run: flake8 pontoon
+
+  black:
+    runs-on: ubuntu-latest
```
flodolo, line 33: Is there a benefit in splitting the two jobs? One job with multiple steps would reuse the checkout and the running image.

Pike: The benefit is that it runs both if one fails. That's a thing I really despise about the travis setup we have right now.

Also, it's easier to understand which tool failed if one doesn't have to dig that up in the logs.

flodolo: Makes sense. I assume it comes with a cost, i.e. more time to run the whole test?

Pike: I followed up on this conversation in the bug, seems worthwhile to document the rationales there.

Pike:
Closing this in favor of #1784

PR #1784, opened 2020-12-21
bug 1683619, migrate automation from travis to github actions

Pike:
Refactoring requirements to allow only linting reqs to be installed.
Exposed requirements to docker, so pip-compile can be run there.

Testing in-repo branch


.github/workflows/python.yml
```diff
@@ -0,0 +1,44 @@
+name: Python linters
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - pontoon/**.py
+      - .github/workflows/python.yml
```
PFischbeck, line 9: I suggest we also run the workflow whenever `requirements/lint.txt` changes

Pike: Added

.travis.yml
```diff
@@ -35,12 +33,6 @@ script:
   - py.test --cov-append --cov-report=term --cov=. -v --create-db --migrations
   - npm test
 
-  # Run all frontend checks
-  - pushd frontend
-  - npm run flow
-  - npm test
-  - popd
-
   - codecov
```
flodolo, line 36: I think you can remove this one too?

Pike: That'll come when I remove the pytest

flodolo: I though you were already running it at the end of `frontend.yml`. Are those two different things? I noticed a coverage report at the end of pytest, but assumed it was one of the arguments passed.

Pike:
This is looking good now. Wall-clock time is around 13 mins, from 11 mins on Travis. There's some good wins to be expected by using the gh docker registry as cache, but I'd rather do that as a follow-up.

Any takers on a review here?

Points of interest: New coverage on frontend. There's an odd part about a reduced test coverage on `settings/base.py`, not sure what's up with that. https://app.codecov.io/gh/mozilla/pontoon/compare/1784 with details.
Also, I tried my best to get the dependencies right, not sure if I missed stuff.
And, I haven't done the README/CONTRIBUTE changes yet, now that I think about that.

jotes (approved):

flodolo:
You might want to edit the PR title, since it covers more than linter.

Also, there a couple of edits to READMEs mentioning Travis (you can find them in #1780).

mathjazz (commented):
This looks really impressive!

Do we expect subsequent PRs from forks to trigger automation after this PR lands? Or is that question still to be resolved?

I only have a couple of naming (🤭) questions/suggestions:

1. Python lint**ers** vs. JavaScript lint**ing**. I just slightly prefer linting (in which case I'd also go with "Heroku app.json validation"), but I'm definitely in favor of consistency. 😄
1. Docker automation also looks more like Backend automation?

If that makes sense to you, you might also want to rename the files a little bit:
* docker.yml -> **backend.yml**
* eslint.yml -> **js-lint.yml**
* frontend.yml -> frontend.yml
* heroku.yml -> heroku.yml
* js-test.yml -> **non-frontend-js.yml**
* python.yml -> **py-lint.yml**

flodolo:
> Do we expect subsequent PRs from forks to trigger automation after this PR lands?

They should just work without any change, because the workflow will be already available in the repo once this merges.

mathjazz (approved):
Woohoo!

I assume `non-frontend-js.yml` and `js-lint.yml` didn't run because they were just renamed without any changes.

Pike:
Good  catch, they didn't run 'cause I didn't update the deps in the workflow to the rename

mathjazz:
> Good catch, they didn't run 'cause I didn't update the deps in the workflow to the rename

Ah!

PR #1786, opened 2020-12-22
Bug 1680056: Join Y-Axis of the Review activity chart

mathjazz:
Using two Y axes with different scale and the same unit is deceiving.


mathjazz:
self-r+

PR #1789, opened 2020-12-23
Fix bug 1684056: Regularly collect stats for Project and ProjectLocale models

mathjazz:
We don't yet have a strong use case for the Insights tab on the `Project` and `ProjectLocale` dashboards, similar to the one we put on the [`Locale` dashboard](https://github.com/mozilla/pontoon/blob/master/specs/0108-community-health-dashboard.md).

If we'll ever have one however, it will be helpful (to say the least) to have statistics snapshots available for each day for the past year or at least a month. We've learned that the hard way when building the aforementioned Locale Insights tab, for which we had to retrieve historic data from DB dumps, and we didn't have many available.

This patch adds `Project` and `ProjectLocale` stats collection to the `collect_insights` management command. We don't collect any other data for these two models for now (e.g. user actions) because it can be retrieved for the past dates using the `ActionLog` model.


pontoon/insights/models.py
```diff
@@ -45,3 +45,15 @@ class Meta:
 
 class LocaleInsightsSnapshot(InsightsSnapshot):
     locale = models.ForeignKey("base.Locale", models.CASCADE)
+
+
+class ProjectInsightsSnapshot(AggregatedStats):
```
mathjazz, line 50: I haven't made the new models inherit from `InsightsSnapshot` yet, because they are highly likely to change when we'll have the actual use case. That it will be a good opportunity to refactor all Insights models.

pontoon/base/admin.py
```diff
@@ -282,6 +282,12 @@ def enabled(self, obj):
     )
 
 
+class ProjectLocaleAdmin(admin.ModelAdmin):
+    search_fields = ["project__name", "project__slug", "locale__name", "locale__code"]
+    list_display = ("pk", "project", "locale", "readonly")
+    ordering = ("-pk",)
```
mathjazz, line 288: I figured we haven't had ProjectLocales exposed in the Admin. Fixing that!

jotes (approved):
Lgtm

PR #1790, opened 2020-12-23
Minor styling improvements on Insights charts

mathjazz:
* More space between a bullet and text in the legend items
* Bigger area to fill with gradient under line charts


mathjazz:
self-r+

PR #1791, opened 2020-12-23
Fix bug 1684086: Prevent resource path overflow in resource menu

mathjazz:
Before:
Screenshot 2020-12-23 at 20 30 47

After:
Screenshot 2020-12-23 at 20 37 37

I was also thinking about adding `direction:rtl`, which adds ellipsis at the beginning. But it seemed even more confusing.


abowler2 (approved):
This all looks good.

PR #1792, opened 2020-12-23
Bug 1672966: Update spec with details on what data is presented

mathjazz:


specs/0108-community-health-dashboard.md
```diff
@@ -22,21 +22,44 @@ A new "Insights" tab is made available on the Team page, positioned next to the
...
+* Active managers have logged into Pontoon.
+* Active reviewers have performed a review activity (approve, unapprove, reject, or unreject a translation).
+* Active contributors have submitted a translation.
+
+Note that only users included in the all users counts can be active users. If a project contact person rejects a suggestion or submits a translation, it doesn't count as an active reviewer or contributor of the team.
 
 ## Unreviewed suggestion lifespan
 
-A line chart showing time before suggestion gets approved or rejected. See the [existing script](https://github.com/flodolo/scripts/blob/954fa85/pontoon/unreviewed_suggestions_lifespan.py) for details on how to retrieve data. A tooltip showing the exact data at a given time appears when hovering over a chart.
+A line chart showing the average age of the unreviewed suggestions in a period of last 12 months.
```
flodolo, line 38: I'm still lost on what this means in practice.

Does it mean that, in December, it will show the data for `2019-12-01 to 2020-12-01`, instead of `2020-12-01 to 2020-12-31` (which is what I would expect)? 

Does it still look at all suggestions submitted, and how long it took for them to be approved/rejected? If so, I'm lost on the huge difference between our previous data (script) and this one. Several locales look in trouble (fr, it), and I know they're not.

mathjazz: It's the latter. In December it will show what the average lifespan (age) of the unreviewed suggestions was between 2020-12-01 and 2020-12-31. More precisely, it looks for unreviewed suggestions once a day and calculates their average age. It then aggregates all the daily averages for a particular month into a monthly average and renders it in the chart.

I now see the above description is confusing. Maybe I should change it to this: `A line chart showing the average age of the unreviewed suggestions in a particular month for a period of last 12 months.`?

> Does it still look at all suggestions submitted, and how long it took for them to be approved/rejected? If so, I'm lost on the huge difference between our previous data (script) and this one. Several locales look in trouble (fr, it), and I know they're not.

It doesn't look at suggestions that have been reviewed, which is what the script does.

mathjazz: Pressed Enter too quickly... Adding more details...

flodolo: Thanks for the explanation. I would suggest to table the conversation for after the holidays, because it's not trivial.

In the end, this is not the "average lifespan of suggestions", it is indeed the "average lifespan of suggestions still unreviewed". I personally see a lot of value in knowing the first data point, and a lot less in the second, mostly because it's included in the first and it doesn't tell us much about how healthy the community is.

> I now see the above description is confusing. Maybe I should change it to this: `A line chart showing the average age of the unreviewed suggestions in a particular month for a period of last 12 months.`?

Yes, this sounds clearer.

mathjazz:
self-r+.

Let's continue the conversation about "Unreviewed suggestion lifespan" changes and other followups at the team meeting.

PR #1795, opened 2020-12-29
Bug 1679710 - Remove duplicates from Concordance search results.

jotes:
@mathjazz r?

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "python.linting.pylintEnabled": false,
+    "python.linting.flake8Enabled": true,
+    "python.linting.enabled": true
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,0 @@
-{
-    "python.linting.pylintEnabled": false,
-    "python.linting.flake8Enabled": true,
-    "python.linting.enabled": true
-}
--- a/pontoon/machinery/tests/test_views.py
+++ b/pontoon/machinery/tests/test_views.py
@@ -299,27 +299,35 @@ def test_view_concordance_search(client, project_a, locale_a, resource_a):
         "/concordance-search/", {"text": "cdd", "locale": locale_a.code},
     )
     result = json.loads(response.content)
-    assert result == [
-        {
-            u"project_name": project_a.name,
-            u"quality": 86,
-            u"source": u"abaf",
-            u"target": u"ccdd",
-        }
-    ]
+    assert result == {
+        "results": [
+            {
+                u"entity_id": entities[1].pk,
+                u"project_name": project_a.name,
+                u"project_slug": project_a.slug,
+                u"source": u"abaf",
+                u"target": u"ccdd",
+            }
+        ],
+        "has_next": False,
+    }
 
     response = client.get(
         "/concordance-search/", {"text": "abaa", "locale": locale_a.code},
     )
     result = json.loads(response.content)
-    assert result == [
-        {
-            u"project_name": project_a.name,
-            u"quality": 100,
-            u"source": u"abaa",
-            u"target": u"ccc",
-        }
-    ]
+    assert result == {
+        "results": [
+            {
+                u"entity_id": entities[0].pk,
+                u"project_name": project_a.name,
+                u"project_slug": project_a.slug,
+                u"source": u"abaa",
+                u"target": u"ccc",
+            }
+        ],
+        "has_next": False,
+    }
 
 
 @pytest.mark.django_db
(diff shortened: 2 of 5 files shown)


mathjazz (commented):
Thanks for the patch!

If I understand correctly, only one result per project will be returned, even if various TranslationMemoryEntries are found and both, source and target are different.

I've changed the 2nd `TranslationMemoryEntry`'s `target` from "ccc1" and the test still passed. I believe it shouldn't - we should get two matches. And I also think we should get two matches in the existing test, because `source` is different.

We should only group results where all three columns are identical: `source`, `target` and `project__name`. The frontend could then also group the results where `source` and `target` are the same, and list all the projects (it already does that for the default Machinery results).

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "python.linting.pylintEnabled": false,
+    "python.linting.flake8Enabled": true,
+    "python.linting.enabled": true
+}
```
mathjazz, line 5: 👀

pontoon/machinery/utils.py
```diff
@@ -71,9 +71,11 @@ def get_concordance_search_data(text, locale):
     )
     search_query = reduce(operator.and_, search_filters)
 
-    search_query_results = base.models.TranslationMemoryEntry.objects.filter(
-        search_query
-    ).values_list("source", "target", "project__name")
+    search_query_results = (
+        base.models.TranslationMemoryEntry.objects.filter(search_query)
+        .values_list("source", "target", "project__name")
```
mathjazz, line 76: While we're at it, maybe also add these two fields?
https://github.com/mozilla/pontoon/pull/1794#discussion_r549899420

pontoon/machinery/tests/test_views.py
```diff
@@ -320,3 +320,61 @@ def test_view_concordance_search(client, project_a, locale_a, resource_a):
...
+            "project_name": "Project A",
+            "quality": 100,
+        },
+        {
+            "source": "abaf",
+            "target": "cccbbb",
+            "project_name": "Project A",
+            "quality": 67,
+        },
+    ]
```
mathjazz, line 380: This test will pass even without the changes you've made to the `get_concordance_search_data()`.

pontoon/machinery/utils.py
```diff
@@ -73,26 +73,34 @@ def get_concordance_search_data(text, locale):
...
         {
             "source": source,
             "target": target,
+            "entity_id": entity_id,
             "project_name": project_name,
+            "project_slug": project_slug,
             "quality": max(
                 int(round(Levenshtein.ratio(text, target) * 100)),
                 int(round(Levenshtein.ratio(text, source) * 100)),
             ),
```
mathjazz, line 90: We're just adding `quality` to `search_query_results` here... And then we remove it again below.

I wonder if there's a more elegant way to sort, something like:
```python
def key(e):
    quality = max(int(...), int(...))
    return quality, e["target"]

return search_query_results.sort(key=key, reverse=True)
```

Or maybe that's slower? In this case I'd add a comment why we're doing it the way we're doing it.

pontoon/machinery/views.py
```diff
@@ -75,6 +76,8 @@ def concordance_search(request):
     try:
         text = request.GET["text"]
         locale = request.GET["locale"]
+        page_results_limit = int(request.GET.get("limit", 100))
+        page = int(request.GET.get("page", 1))
     except MultiValueDictKeyError as e:
```
mathjazz, line 81: We should also catch a `ValueError` here in case `limit` or `page` aren't integers. Then we also don't need to catch `PageNotAnInteger` below.

mathjazz (commented):
Nice job! I only added a couple of ideas for potential improvements.

pontoon/machinery/utils.py
jotes: I tried to avoid using too much lambdas and internal functions because they often decrease readability for other people I've worked with. I can switch into your version.

pontoon/machinery/utils.py
```diff
@@ -73,26 +73,34 @@ def get_concordance_search_data(text, locale):
 
     search_query_results = (
         base.models.TranslationMemoryEntry.objects.filter(search_query)
-        .values_list("source", "target", "project__name")
+        .values_list("source", "target", "entity__pk", "project__name", "project__slug")
```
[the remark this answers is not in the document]
jotes: I think that `pk` is more suitable because it's suggested by the Django project. I unified the rest of the patch to use it.

mathjazz (approved):
Great work! Thanks!

PR #1796, opened 2021-01-01
Fix Bug 1633150  Add ability to enable projects with read-only locales only

Mitch4sho:
fixed the problem where you couldn't save a project with read-only locales in the admin project

### DIFF
--- a/pontoon/administration/forms.py
+++ b/pontoon/administration/forms.py
@@ -1,6 +1,7 @@
 from django import forms
 from django.contrib.auth.models import User
 from django.forms.models import inlineformset_factory
+from django.core.exceptions import ValidationError
 
 from pontoon.base.models import (
     Entity,
@@ -28,6 +29,14 @@ class ProjectForm(forms.ModelForm):
     locales = forms.ModelMultipleChoiceField(
         queryset=Locale.objects.all(), required=False,
     )
+
+    def clean(self):
+        cleaned_data = super().clean()
+        locales_readonly = cleaned_data.get("locales_readonly")
+        locales = cleaned_data.get("locales")
+        if not (locales or locales_readonly):
+            raise ValidationError("This field is required")
+        
 
     class Meta:
         model = Project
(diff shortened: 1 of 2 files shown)


pontoon/administration/forms.py
```diff
@@ -25,6 +25,9 @@ class ProjectForm(forms.ModelForm):
     locales_readonly = forms.ModelMultipleChoiceField(
         queryset=Locale.objects.all(), required=False,
     )
+    locales = forms.ModelMultipleChoiceField(
+        queryset=Locale.objects.all(), required=False,
+    )
```
mathjazz, line 30: Good work!

We should still require that at least one locale is enabled for the project before it can be saved (regardless if it's localizable or read-only).

Resources:
https://stackoverflow.com/questions/31831620/can-i-make-at-least-one-field-a-requirement-on-a-django-model
https://docs.djangoproject.com/en/dev/ref/forms/validation/#cleaning-and-validating-fields-that-depend-on-each-other

Mitch4sho: Alright will go ahead and implement this

pontoon/administration/forms.py
```diff
@@ -25,6 +26,16 @@ class ProjectForm(forms.ModelForm):
...
+
+    def clean(self):
+        cleaned_data = super().clean()
+        locales_readonly = cleaned_data.get("locales_readonly")
+        locales = cleaned_data.get("locales")
+        if not (locales or locales_readonly):
+            raise ValidationError("This field is required")
```
mathjazz, line 38: The message should read "At least one locale must be selected.", otherwise the UI implies the localizable locale field is required.

pontoon/administration/templates/admin_project.html
```diff
@@ -53,9 +53,8 @@ <h3>Locales</h3>
   <div class="locales clearfix">
       {{ admin_team_selector.render(locales_available, locales_selected, locales_readonly) }}
       {{ form.locales }}
-      {{ form.locales.errors }}
+      {{ form.non_field_errors() }}
       {{ form.locales_readonly }}
-      {{ form.locales_readonly.errors }}
```
mathjazz, line 58: I think we should keep `{{ form.locales.errors }}` and `{{ form.locales_readonly.errors }}` in case there are any field-specific errors.

mathjazz (commented):
Thanks for the update! Almost there!

mathjazz (approved):
Great job! Thanks for the patch!

Mitch4sho:
Thanks!

PR #1799, opened 2021-01-07
Fix Bug 1543344: Implement Tab shortcut in editor

julen:
* Reimplements the tab shortcut
* Brings back the keyboard shortcuts panel entry

### DIFF
--- a/frontend/public/static/locale/en-US/translate.ftl
+++ b/frontend/public/static/locale/en-US/translate.ftl
@@ -454,7 +454,7 @@ machinery-MicrosoftTerminology--visit-microsoft = MICROSOFT
 ## Machinery Translation
 ## Shows a specific translation from machinery.
 machinery-Translation--copy =
-    .title = Copy Into Translation
+    .title = Copy Into Translation (Tab)
 
 
 ## Machinery Translation Memory
--- a/frontend/src/core/editor/actions.js
+++ b/frontend/src/core/editor/actions.js
@@ -20,8 +20,12 @@ export const END_UPDATE_TRANSLATION: 'editor/END_UPDATE_TRANSLATION' =
 export const RESET_EDITOR: 'editor/RESET_EDITOR' = 'editor/RESET_EDITOR';
 export const RESET_FAILED_CHECKS: 'editor/RESET_FAILED_CHECKS' =
     'editor/RESET_FAILED_CHECKS';
+export const RESET_HELPER_INDEX: 'editor/RESET_HELPER_INDEX' =
+    'editor/RESET_HELPER_INDEX';
 export const RESET_SELECTION: 'editor/RESET_SELECTION' =
     'editor/RESET_SELECTION';
+export const SELECT_HELPER_INDEX: 'editor/SELECT_HELPER_INDEX' =
+    'editor/SELECT_HELPER_INDEX';
 export const SET_INITIAL_TRANSLATION: 'editor/SET_INITIAL_TRANSLATION' =
     'editor/SET_INITIAL_TRANSLATION';
 export const START_UPDATE_TRANSLATION: 'editor/START_UPDATE_TRANSLATION' =
@@ -91,6 +95,29 @@ export function updateMachinerySources(
         type: UPDATE_MACHINERY_SOURCES,
         machinerySources,
         machineryTranslation,
+    };
+}
+
+/**
+ * Reset selected helpers index to its initial value.
+ */
+export type ResetHelperIndexAction = {|
+    +type: typeof RESET_HELPER_INDEX,
+|};
+export function resetHelperIndex(): ResetHelperIndexAction {
+    return {
+        type: RESET_HELPER_INDEX,
+    };
+}
+
+export type SelectHelperIndexAction = {|
+    +type: typeof SELECT_HELPER_INDEX,
+    +index: number,
+|};
+function selectHelperIndex(index: number): SelectHelperIndexAction {
+    return {
+        type: SELECT_HELPER_INDEX,
+        index,
     };
 }
 
@@ -286,8 +313,10 @@ export default {
     endUpdateTranslation,
     reset,
     resetFailedChecks,
+    resetHelperIndex,
     resetSelection,
     sendTranslation,
+    selectHelperIndex,
     setInitialTranslation,
     startUpdateTranslation,
     update,
--- a/frontend/src/core/editor/components/KeyboardShortcuts.css
+++ b/frontend/src/core/editor/components/KeyboardShortcuts.css
@@ -14,7 +14,7 @@
     border: 1px solid #333941;
     font-size: 13px;
     width: 400px;
-    height: 370px;
+    height: 400px;
     left: 0;
     top: 0;
     bottom: 0;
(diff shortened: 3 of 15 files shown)


mathjazz (commented):
Thanks for the patch, @julen!

Two high-level comments:

1. Compared to the old Translate app, we now have two lists visible at the same time that could be copied from: History, and Machinery or Locales. That represents a UX issue if we want to copy from all three lists, because the user would need to somehow pick between History and the other visible panel. I see that you only implemented the Tab shortcut for the Machinery panel, which I think is the right choice, because the other two are less likely to be copied from. I'm fine with dropping support for copying from the History panel. Should we still allow copying from the Locales panel when it visible? In any case, we should revert changes made to the `title` attribute accordingly and update the text in the keyboard shortcut panel.

1. It would be helpful to implement a visual feedback when copying happens - in the old Translate app the copied item (e.g. in the Machinery panel) got highlighted (as if it was hovered over), and the list of items scrolled if needed to make sure the copied item is always visible.

julen:
Thanks for the quick feedback @mathjazz!

Ah true, I blindly reverted the old patch that removed the shortcuts panel entry and didn't pay attention to the History and Locales panels (should be fixed now). My personal experience (and motivation to fix this bug) is that I use the Machinery entries very often, whereas the other two I barely touch, therefore I'd also be inclined to only support copying Machinery entries only.

Regarding the second point, do you mean to temporarily highlight it? If you instead mean to highlight it permanently, then for consistency that should also maybe be the behavior when users select a machinery entry. What do you think?

mathjazz:
> Ah true, I blindly reverted the old patch that removed the shortcuts panel entry and didn't pay attention to the History and Locales panels (should be fixed now). My personal experience (and motivation to fix this bug) is that I use the Machinery entries very often, whereas the other two I barely touch, therefore I'd also be inclined to only support copying Machinery entries only.

For consistency between both panels (and to justify the name "Copy from Helpers"), I'd suggest we also support copying from the Locales tab when it's active. It can also be a followup patch.

> Regarding the second point, do you mean to temporarily highlight it? If you instead mean to highlight it permanently, then for consistency that should also maybe be the behavior when users select a machinery entry. What do you think?

These are both interesting ideas. I'm fine with both.

julen:
Oh, now I also see manually copying a machinery entry does a bit more than just copying, e.g. it [updates the machinery sources](https://github.com/mozilla/pontoon/blob/008a6ac85b0ca9b8e6cda74401179fc66a56f06b/frontend/src/modules/machinery/components/Translation.js#L72), which is then kept in the backend to determine if an entry was copied from TM. This will take a bit longer to implement, but it's probably worth factoring out that logic so the end result/experience is the same regardless of where copying is triggered from.

mathjazz:
Correct, Tab would effectively need to call `copyTranslationIntoEditor()` for the selected translation:
https://github.com/mozilla/pontoon/blob/008a6ac85b0ca9b8e6cda74401179fc66a56f06b/frontend/src/modules/machinery/components/Translation.js#L48-L81

julen:
I extracted the copying logic into a hook and now the logic is reused for shortcuts. What's left is the visual feedback; support for the Locales tab can be worked on separately.

julen:
Highlighting is now added as a visual feedback. This only highlights machinery entries upon explicit selection, i.e. it won't highlight an entry if it was just loaded from the entity list and it matches a machinery translation — this, if deemed useful, could be worked on in a follow-up change.

mathjazz (commented):
The feature works very well and the code is super clean. Excellent job!

The only concern I have on the code quality is the drop in test coverage. Could you replace the removed tests with the equivalent unit tests for the `useCopyMachineryTranslation` hook?

On the UX side, it would be nice to scroll the Machinery panel when the selected item overflows. If it's a bigger effort, we can also take care of it in a followup (same as we decided to do with the Locales panel).

julen:
> The only concern I have on the code quality is the drop in test coverage. Could you replace the removed tests with the equivalent unit tests for the `useCopyMachineryTranslation` hook?

How do you replace them? I'm asking because I found zero tests for hooks in the codebase and I have no clue how Pontoon wants to go forward about this. I removed existing tests because of that, and also because they were testing implementation details, rather than functionality.

> On the UX side, it would be nice to scroll the Machinery panel when the selected item overflows. If it's a bigger effort, we can also take care of it in a followup (same as we decided to do with the Locales panel).

Not super hard, but would be happy to handle it as a follow-up.

mathjazz:
Thanks for the update!

> How do you replace them? I'm asking because I found zero tests for hooks in the codebase and I have no clue how Pontoon wants to go forward about this. I removed existing tests because of that, and also because they were testing implementation details, rather than functionality.

All good points.

@adngdb I'm curious if you've made any plans for testing hooks.

adngdb:
Testing hooks is a bit complex, and I haven't done much of it as you can see. Generally, I propose the following guidelines:

- if the hook is used in a single component, test the component and not the hook;
- if the hook is used in several components but is simple, create a fake component to test it;
- in other cases, and only if necessary, use [react-hooks-testing-library](https://react-hooks-testing-library.com/).

Hope that makes sense and helps!

mathjazz:
Thanks @adngdb!

@julen There's one final bug we should fix before merging: say you have 3 Machinery item and you select the 2nd one with the mouse. After you press Tab, we the 1st one (instead of the 3rd one). The other symptom of (probably) the same bug is that if you have a [string with plural forms](https://pontoon.mozilla.org/sl/amo/all-resources/?string=66616), tab index doesn't reset as you move between plural forms.

I'll file bugs to:
* Add support for the Locales tab
* Move scrollbar to make sure selected item is always visible
* Add tests.

julen:
> in other cases, and only if necessary, use [react-hooks-testing-library](https://react-hooks-testing-library.com/).

Thanks for the feedback @adngdb. I've had a look at the testing situation, and I believe we fall under _in other cases_, although I don't see how the testing library can help because the currently existing hooks have many side-effects; none that could be verified by checking a return value (like the examples in the hooks testing library).

If behavior wants to be tested, we need to be aware these particular hooks rely on multiple parts of the rendering tree and can have different triggers too (keyboard shortcut, element click), and to my untrained eye this looks like we will need to deeply render a whole tree down from `EntityDetails`.

Happy to hear if any of this makes sense and overall any other thoughts in general.

julen:
> @julen There's one final bug we should fix before merging: say you have 3 Machinery item and you select the 2nd one with the mouse. After you press Tab, we the 1st one (instead of the 3rd one). The other symptom of (probably) the same bug is that if you have a [string with plural forms](https://pontoon.mozilla.org/sl/amo/all-resources/?string=66616), tab index doesn't reset as you move between plural forms.

Ah, I see, good catch. To solve this we'll need to share state between the shortcut handler and the UI.

julen:
> @julen There's one final bug we should fix before merging: say you have 3 Machinery item and you select the 2nd one with the mouse. After you press Tab, we the 1st one (instead of the 3rd one). The other symptom of (probably) the same bug is that if you have a [string with plural forms](https://pontoon.mozilla.org/sl/amo/all-resources/?string=66616), tab index doesn't reset as you move between plural forms.

@mathjazz these two should now be addressed.

mathjazz (approved):
Great work!

I'm sure many users will appreciate resurrection of this old handy feature, myself included.

mathjazz:
Bugs filed:
https://bugzilla.mozilla.org/show_bug.cgi?id=1686369
https://bugzilla.mozilla.org/show_bug.cgi?id=1686370
https://bugzilla.mozilla.org/show_bug.cgi?id=1686371

PR #1801, opened 2021-01-07
bug 1685558, use carryforward flags to deal with skipped test runs

Pike:


Pike:
I have no idea why this fails to get the reports processed, opened https://community.codecov.io/t/trying-to-introduce-flags-triggers-error-on-processing-reports/2450

.codecov.yml
```diff
@@ -1 +1,18 @@
 comment: off
+
+flags:
+  backend:
+    paths:
+      - /pontoon/**.py
```
thomasrockhu, line 6:
```suggestion
      - pontoon/**.py
```
I think this might work

Pike: Thanks, that seems to have done the trick. I pushed each path individually, and the backend tests already are marked as carry-forward until the ones from the current test run made it in.

Pike:
This is ready for review, per Tom's comment on discourse.

mathjazz (approved):
\o/

PR #1802, opened 2021-01-07
Add link to Slate docs to Frontend ReadMe documentation

abowler2:
It felt more appropriate to add this to the `Tools documentation` section at the end of the ReadMe, but let me know if you think it needs to also be in the `Tools` table at the top


mathjazz (approved):
Thanks!

PR #1804, opened 2021-01-08
Fix Bug 1684013 Document environment variables used for custom error and maintenance pages

anujpandey785:
Fix Bug 1684013 Document environment variables used for custom error and maintenance pages

Modified the Deployment Documentation with respect to Environment Variables used to configure Custom Error Pages :
1)Documented ``ERROR_PAGE_URL`` Environment variable.
2)Documented ``MAINTENANCE_PAGE_URL`` Environment variable.


mathjazz:
It turns out these are already documented. I don't know what was I thinking. Sorry about that!

PR #1805, opened 2021-01-08
Update lxml to 4.6.2 and python to 3.8.7

mathjazz:
I've tested locally and on stage and didn't spot any issues.

The amount of diff in the `.txt` files is probably caused by the latest version of `pip-tools` I installed locally.

Could we replace `compile-all.sh` with a `make compile-requirements` and compile requirements in Docker?

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.8.3-buster
+FROM python:3.8-buster
 
 ARG USER_ID=1000
 ARG GROUP_ID=1000
--- a/requirements/default.in
+++ b/requirements/default.in
@@ -35,7 +35,7 @@ graphene-django==2.13.0
 gunicorn==19.9.0
 jsonfield==3.1.0
 jsonschema==2.6.0
-lxml==4.3.3
+lxml==4.6.2
 mercurial==5.2.2
 newrelic==2.106.1.88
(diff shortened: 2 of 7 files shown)


Pike:
I've actually run it inside docker the  past few times I updated deps. I wouldn't mind that being the default.

There's a  bit of trickery involved in that, though, as all other dependencies are installed by root,  and the Pontoon user  doesn't have access or its path set up right for local installs of piptools.

And yes, the formatting gone different with https://github.com/jazzband/pip-tools/pull/1237

PFischbeck (approved):
Looks good to me! This commit might also be the right time to update the Python version in the Dockerfile. Right now, it is at `3.8.3-buster`. We could use either `3.8.7-buster` or  `3.8-buster`, which always refers to the latest patch version.

Compiling requirements in Docker sounds good, but unfortunately I don't know enough about Docker to help much.

mathjazz:
> Looks good to me! This commit might also be the right time to update the Python version in the Dockerfile. Right now, it is at `3.8.3-buster`. We could use either `3.8.7-buster` or `3.8-buster`, which always refers to the latest patch version.

Thanks, I'll update that.

> Compiling requirements in Docker sounds good, but unfortunately I don't know enough about Docker to help much.

I have a patch ready :)

PR #1807, opened 2021-01-10
Fix bug 1678462 - Use new enumeration types

PFischbeck:
We now use the new enumeration types wherever possible, in order to have more readable and maintainable code.

This is best reviewed per commit, as every commit introduces a single enum. The final commit (`Repository.Type`) includes some small drive-by refactoring, and it also introduced a cyclic dependency between `base/models.py` and `sync/vcs/repositories.py`. I solved it by making all imports in `base/models.py` local, but I am open to alternative solutions. At some point, some more refactoring of `sync/vcs/repositories.py` probably wouldn't hurt,

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -951,7 +951,7 @@ class Locale(AggregatedStats):
         TranslatedResource.objects.filter(
             resource__project__disabled=False,
             resource__project__system_project=False,
-            resource__project__visibility="public",
+            resource__project__visibility=Project.Visibility.PUBLIC,
             locale=self,
         ).aggregate_stats(self)
 
@@ -1151,7 +1151,7 @@ class ProjectQuerySet(models.QuerySet):
         if user.is_superuser:
             return self
 
-        return self.filter(visibility="public")
+        return self.filter(visibility=Project.Visibility.PUBLIC)
 
     def available(self):
         """
@@ -1271,12 +1271,12 @@ class Project(AggregatedStats):
     """,
     )
 
-    VISIBILITY_TYPES = (
-        ("private", "Private"),
-        ("public", "Public"),
-    )
+    class Visibility(models.TextChoices):
+        PRIVATE = "private", "Private"
+        PUBLIC = "public", "Public"
+
     visibility = models.CharField(
-        max_length=20, default=VISIBILITY_TYPES[0][0], choices=VISIBILITY_TYPES,
+        max_length=20, default=Visibility.PRIVATE, choices=Visibility.choices,
     )
 
     # Website for in place localization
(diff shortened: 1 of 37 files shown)


pontoon/base/models.py
```diff
@@ -3035,18 +3037,19 @@ class Translation(DirtyFieldsMixin, models.Model):
...
-    SOURCE_TYPES = (
-        ("translation-memory", "Translation Memory"),
-        ("google-translate", "Google Translate"),
-        ("microsoft-translator", "Microsoft Translator"),
-        ("systran-translate", "Systran Translate"),
-        ("microsoft-terminology", "Microsoft"),
-        ("transvision", "Mozilla"),
-        ("caighdean", "Caighdean"),
-    )
+    class Source(models.TextChoices):
```
mathjazz, line 3040: I'd rename the `Source` class to `MachinerySource` to be more explicit.

pontoon/base/models.py
```diff
@@ -1867,7 +1861,7 @@ def checkout_path(self):
     @property
     def can_commit(self):
         """True if we can commit strings back to this repo."""
-        return self.type in ("svn", "git", "hg")
+        return self.type in (self.Type.SVN, self.Type.GIT, self.Type.HG)
```
mathjazz, line 1864: We should drop the `can_commit` logic entirely (can be done in a followup), because this will always be True.

pontoon/base/tests/models/test_repository.py
```diff
@@ -2,26 +2,10 @@
...
-    assert repo_file.checkout_path == os.path.join(
-        *[repo_file.project.checkout_path] + urlparse(repo_file.url).path.split("/")
-    )
-    settings.MEDIA_ROOT = "/media/root"
-    assert repo_file.checkout_path == os.path.join(
-        *[repo_file.project.checkout_path] + urlparse(repo_file.url).path.split("/")
-    )
-    assert repo_file.project.checkout_path.startswith("/media/root")
-
-
```
mathjazz, line 24: I'm curious why did you choose to remove this test case.

mathjazz (commented):
Nice work!

Only left one note about naming and another one about `can_commit` (which could be done in a followup).

pontoon/base/models.py
PFischbeck: Done :+1:

pontoon/base/models.py
PFischbeck: Sure, I can do this in a follow-up PR.

pontoon/base/tests/models/test_repository.py
PFischbeck: It was in error - when skimming the test, I thought it did something `file`-specific. I have re-added the test (now using `repo_git`).

pontoon/base/models.py
mathjazz: Thanks, filed https://bugzilla.mozilla.org/show_bug.cgi?id=1686278.

mathjazz (approved):
\o/

PR #1808, opened 2021-01-12
Fix bug 1686278 - Drop can_commit logic

PFischbeck:


mathjazz (approved):
That was fast. 👍 

Nice job!

PR #1809, opened 2021-01-13
Fix bug 1686370: scroll selected helper item into view

julen:
Follow-up of #1799.

### DIFF
--- a/frontend/src/modules/machinery/components/Translation.js
+++ b/frontend/src/modules/machinery/components/Translation.js
@@ -57,12 +57,26 @@ export default function Translation(props: Props) {
         className += ' selected';
     }
 
+    const translationRef = React.useRef();
+    React.useEffect(() => {
+        if (
+            editorState.selectedHelperIndex === index &&
+            translationRef.current
+        ) {
+            translationRef.current.scrollIntoView({
+                behavior: 'smooth',
+                block: 'nearest',
+            });
+        }
+    }, [editorState.selectedHelperIndex, index]);
+
     return (
         <Localized id='machinery-Translation--copy' attrs={{ title: true }}>
             <li
                 className={className}
                 title='Copy Into Translation (Tab)'
                 onClick={copyTranslationIntoEditor}
+                ref={translationRef}
             >
                 <header>
                     {!translation.quality ? null : (


frontend/src/modules/machinery/components/Translation.js
```diff
@@ -57,12 +57,26 @@ export default function Translation(props: Props) {
...
+    const translationRef = React.useRef();
+    React.useEffect(() => {
+        if (
+            editorState.selectedHelperIndex === index &&
+            translationRef.current
+        ) {
+            translationRef.current.scrollIntoView({
+                behavior: 'smooth',
+                block: 'nearest',
+            });
```
mathjazz, line 69: OMG, I had no idea `scrollIntoView` is a thing! Thanks!

mathjazz: I wonder what over places could we use it. Two come to my mind:
* https://github.com/mozilla/pontoon/blob/448bf00900923f980f9decfd8a53f966f3b0928c/frontend/src/modules/entitieslist/components/EntitiesList.js#L154
* Mentions list position, although I'm not sure about this one (cc @abowler2).

mathjazz (approved):
This is great, thanks!

frontend/src/modules/machinery/components/Translation.js
julen: If you file a bug for those it seems like low-hanging fruit.

mathjazz: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1686569 for the first one. :)

abowler2: @mathjazz Can you clarify what you mean by the Mentions list position? If you mean using this in place of the position calculations we use, I don't think this will work as I want to say I tried this among the many attempts I made to solve that problem. 

With that said, the Mentions list does have a `ref` already assigned that could be used to test it out.

mathjazz: Yeah, that's what I meant. Thanks for confirming. I don't think `scrollIntoView` could be of much help here.

PR #1810, opened 2021-01-13
Bugifx: Rename Source to MachinerySource

mathjazz:


mathjazz:
Pushing this to prod. We should add a test in a followup.

PR #1811, opened 2021-01-13
Fix bug 1686369: implement Tab shortcut for 'Locales' tab

julen:
Overall it resembles the implementation of the Machinery tab, although some extra changes have been necessary to be able to use hooks (that's why there's extra noise in the diff).

### DIFF
--- a/frontend/src/modules/otherlocales/selectors.js
+++ b/frontend/src/modules/otherlocales/selectors.js
@@ -1,0 +1,48 @@
+/* @flow */
+
+import { createSelector } from 'reselect';
+
+import type { OtherLocaleTranslation, OtherLocaleTranslations } from 'core/api';
+
+import { NAME } from '.';
+
+const otherLocalesSelector = (state): string => state[NAME].translations;
+
+/**
+ * Return the Entity that follows the current one in the list.
+ */
+
+function _flattenOtherLocales(
+    otherLocales: OtherLocaleTranslations,
+): Array<OtherLocaleTranslation> {
+    if (!otherLocales) {
+        return [];
+    }
+
+    let result = [];
+    if (otherLocales.preferred) {
+        otherLocales.preferred.forEach((translation) => {
+            result.push(translation);
+        });
+    }
+
+    if (otherLocales.other) {
+        otherLocales.other.forEach((translation) => {
+            result.push(translation);
+        });
+    }
+
+    return result;
+}
+
+/**
+ * Return the Entity that preceeds the current one in the list.
+ */
+export const getTranslationsFlatList: Function = createSelector(
+    otherLocalesSelector,
+    _flattenOtherLocales,
+);
+
+export default {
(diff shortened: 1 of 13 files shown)


frontend/src/modules/otherlocales/selectors.js
```diff
@@ -0,0 +1,48 @@
...
+
+import { NAME } from '.';
+
+const otherLocalesSelector = (state): string => state[NAME].translations;
+
+/**
+ * Return the Entity that follows the current one in the list.
+ */
+
+function _flattenOtherLocales(
```
julen, line 15: With the current backend response and redux state shape, I had to do this in order to have all other locale translations in a single list. I feel like the backend could already provide this list as-is, and the fact that some locales are preferred could be indicated via an optional `"is_preferred": true` flag perhaps. So I wonder what the motivation of the current backend response is. Any ideas?

mathjazz (commented):
Another excellent patch, thanks so much!

I only noticed a couple of issues with selecting (highlighting) items in the Locales tab:
* If there are multiple items in the Locales tab with the same value, they all get selected. This will happen quite often in strings like [brand names](https://pontoon.mozilla.org/fr/firefox/browser/branding/official/brand.dtd/?string=74127).
* In the case of FTL strings, items don't get highlighted at all. That's because we're comparing the value shown in the editor with the source value of the FTL string (in the form of `key = value`).

Why should probably just check if `editorState.selectedHelperElementIndex === index` to determine if the translation needs to be highlighted.

frontend/src/modules/otherlocales/selectors.js
julen: Cool, I could handle that in a follow-up as well.

julen:
> Why should probably just check if `editorState.selectedHelperElementIndex === index` to determine if the translation needs to be highlighted.

Thanks @mathjazz, this works, and I updated the PR accordingly. One thing to note however is that because selection indexes are shared across tabs, changing tabs will make selections go away.

Example:
1. Select `element #2` from Locales tab --> `element #2` is highlighted
2. Switch to Machinery tab --> no elements are highlighted, as expected
3. Switch back to Locales tab --> no elements are highlighted

If we are OK with this last step, then we can go ahead with the existing implementation. Otherwise, I believe we'd need to either:
* Make the selection index state specific to each tab, or
* Have some sort of uuid to uniquely identify machinery/locales results (as far as I understand they don't map 1:1 to entity IDs, because a single machinery entry can have multiple sources)

frontend/src/modules/otherlocales/selectors.js
mathjazz: Great! Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1686690.

mathjazz (approved):
I think this works good enough already and adding more code complexity would bring much benefit. Besides, focus in the editor is also lost when you switch tabs.

PR #1812, opened 2021-01-14
Fix bug 1686711: Add basic unit test for Translation.machinery_sources_values

mathjazz:
We had to land a hotfix for the error we'd catch with a test like this: https://github.com/mozilla/pontoon/commit/7181fd430bac94725eab7880d1eb8e54f8192eb6.

### DIFF
--- a/pontoon/base/tests/models/test_translation.py
+++ b/pontoon/base/tests/models/test_translation.py
@@ -273,3 +273,23 @@ def test_translation_not_in_terminology_saved(
     TranslationFactory.create(locale=locale_a, entity=entity_a)
 
     assert not reset_term_translation_mock.called
+
+
+@pytest.mark.django_db
+def test_machinery_sources_values(locale_a, entity_a):
+    """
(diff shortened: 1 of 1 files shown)


pontoon/base/tests/models/test_translation.py
```diff
@@ -273,3 +273,23 @@ def test_translation_not_in_terminology_saved(
...
+    """
+    translation_no_machinery_sources = TranslationFactory.create(
+        locale=locale_a, entity=entity_a,
+    )
+    assert translation_no_machinery_sources.machinery_sources_values == ""
+
+    translation = TranslationFactory.create(
+        locale=locale_a,
+        entity=entity_a,
+        machinery_sources=["translation-memory", "google-translate"],
```
PFischbeck, line 291: You could replace this with `Translation.MachinerySource.TranslationMemory/GoogleTranslate`.

PFischbeck (approved):
Looks good to me! Thanks also for the hotfix. I only have a small optional suggestion.

PR #1813, opened 2021-01-14
No bug - Adjust code comment to reflect actual 'Tab' behavior

julen:
Sorry I had this change stashed for the last PR but failed to update the commit accordingly.


mathjazz (approved):
Nice catch!

PR #1814, opened 2021-01-15
Switch to a MIT-licensed version of fullpage.js

mathjazz:
The version we are downgrading two was released 2 months before the GPL3 version we've been using.


Pike (approved):
LGTM.

I consider this one a tech debt issue we should resolve by removing the library eventually.

PR #1815, opened 2021-01-16
Bug 1682033: Improve visibility of pinned comments

Mitch4sho:
Added ability for the pinned comments to be set at the top when pin is selected. Fixed metadata pinned comments to be pinned at the top of this section.


frontend/src/core/comments/components/CommentsList.css
```diff
@@ -24,3 +24,8 @@
     flex-flow: column wrap;
     width: 100%;
 }
+
+.pinned-comments {
+    border-bottom: 1px solid #5e6475;
```
mathjazz, line 29: Let's follow the styling of Preferred Locales in the Locales tab and use `#7bc876` as the border color.

Same as in the Locales panel, I'd also use the full-width the border line, by moving padding from the parent element to `ul` and `div` elements.

frontend/src/core/comments/components/CommentsList.js
```diff
@@ -42,17 +42,35 @@ export default function CommentsList(props: Props) {
...
-                        key={comment.id}
-                        togglePinnedStatus={togglePinnedStatus}
-                    />
-                ))}
+            <ul className='pinned-comments'>
+                {pinnedComments.map((comment) => renderComment(comment))}
+            </ul>
+
+            <ul className='unpinned-comments'>
+                {unpinnedComments.map((comment) => renderComment(comment))}
```
mathjazz, line 73: For better readability in the chronologic order, I'd repeat pinned comments below. So it should be pinned comments and then all comments, instead of pinned and unpinned.

frontend/src/core/comments/components/CommentsList.css
```diff
@@ -1,12 +1,19 @@
...
-    padding: 8px;
 }
 
 .comments-list .comment {
     display: flex;
     margin-left: 58px;
 }
+.comments-list ul li {
+    border-bottom: 1px solid #5e6475;
+    padding: 8px;
```
mathjazz, line 11: This will add 16px padding between comments (8px above and 8px below).

frontend/src/core/comments/components/CommentsList.css
```diff
@@ -1,12 +1,19 @@
...
     line-height: 22px;
-    padding: 8px;
 }
 
 .comments-list .comment {
     display: flex;
     margin-left: 58px;
 }
+.comments-list ul li {
+    border-bottom: 1px solid #5e6475;
```
mathjazz, line 10: Only the last child among the pinned comments should have a border.

frontend/src/core/comments/components/CommentsList.css
mathjazz: And we also need to add left and right padding to the add comment form.

mathjazz (commented):
Left some notes.

Also make sure translation comments styling isn't impacted.

frontend/src/core/comments/components/CommentsList.css
Mitch4sho: For the padding did you want to not have that much padding between the comments?

mathjazz: No, I'd like it to remain the same as on master.

frontend/src/core/comments/components/CommentsList.css
Mitch4sho: I'm not too sure what to do here. Do you want the pinned comments to have no borders besides the last pinned comment?

mathjazz: Correct. The apperance should be the same as on master. The only change would be the border between pinned comments and all comments.

Mitch4sho:
fixing testing

frontend/src/core/comments/components/CommentsList.js
mathjazz: This isn't resolved - we still iterate over `unpinnedComments` comments, instead of (all) `comments`.

frontend/src/core/comments/components/CommentsList.js
```diff
@@ -42,17 +42,35 @@ export default function CommentsList(props: Props) {
...
+
+    const [pinnedComments, allComments] = comments.reduce(
+        (comment, currentComment) => {
+            comment[currentComment.pinned === true ? 0 : 1].push(
+                currentComment,
+            );
+            return comment;
+        },
+        [[], []],
+    );
```
mathjazz, line 64: We should replace this block with something simpler like:
```js
const pinnedComments = comments.filter(comment => comment.pinned === true);
```

And then use `comments` instead of `allComments` below.

Mitch4sho: I want to make sure I have this right before I push the code. I replace the block of code with the one you suggested. But know it doesn't Filter the `pinnedComments` from the original comments. It is now duplicating the comments instead of pinning the one to the top when the pin is selected. Do you want me to create a filter to filter the pinned comments from the original comments, or was this code you suggested to achieve this function?

julen: `.filter(comment => comment.pinned)` should suffice if I read that correctly.

abowler2: @mathjazz Just an FYI, I'm working with Mitch on this.

mathjazz: Yup, comments should be duplicated. Pinned comments on top, then all comments in chronological order.

If you extract pinned comments, it's hard to follow the discussion.

Mitch4sho: Got it! Do you want to remove the pin for all comments because when pinned it shows the pin in both places?

Screen Shot 2021-01-21 at 10 15 15 AM

mathjazz: I think it's helpful to see the pin in the list of all comments as well, to get the sense of which comments are important when you're going through a list of comments, especially if it's long.

frontend/src/core/comments/components/CommentsList.js
```diff
@@ -42,17 +42,28 @@ export default function CommentsList(props: Props) {
...
+            <Comment
+                comment={comment}
+                canPin={canPin}
+                key={comment.id}
+                togglePinnedStatus={togglePinnedStatus}
+            />
+        );
+    };
+
+    const pinnedComments = comments.filter((comment) => comment.pinned == true);
```
mathjazz, line 57: You can remove ` == true`, which should also fix the test.

Mitch4sho:
hopefully, this is the last time

mathjazz (approved):
Works great! Nice job!

I'll kindly ask you to review a couple of related CSS changes after I push them.

mathjazz:
@ Mitch4sho I've added some CSS changes as agreed on chat. Could you please have a look and review?

The most notable change is the title for the pinned and all comments sections. I'm afraid without it, users would be wondering why some comments are repeated. I've also made pinned comments slightly more notable by adding the "PINNED" title.

I also tried changing the background color, but wasn't happy with the result, so I dropped that effort.

Mitch4sho:
Awesome I will review the changes later today!

Mitch4sho:
Everything seems great! you made the pinned section look way better

PR #1816, opened 2021-01-16
No bug - ESLint: remove fullpage.js globals

julen:
These two globals are not present anymore. This is a follow-up of #1814.


mathjazz (approved):
Thanks!

I wonder how did you catch these?

julen:
> I wonder how did you catch these?

I was curious what fullpage.js is and just grepped its (former) API usage in the codebase.

PR #1817, opened 2021-01-16
Fix bug 1687112: remove fullpage.js dependency

julen:
* Re-implements the logic used from fullpage.js in ~120 LoC
* Removes fullpage.js
* Changes `homepage_content.html` in a backwards-compatible way
* Keeps existing visual layout and design
* Keeps existing behavior

I'm still playing around the scroll/wheel acceleration in order to have a behavior closer to what it was before. Other than that, the whole thing should be working.


mathjazz:
Thanks for the patch!

I've notice two issues:
- Quick scrolling though the page doesn't work well. There's a slight delay after the section stops scrolling before you can scroll again.
- The Edit Homepage button (seen if you are an admin) should be aligned right and visible on each page.

Also note that the tests are failing, because pipeline/yuglify chokes on modern JS features. :/

julen:
Yes, as I said I'm still working with the scrolling behavior. The Edit button should be fixed now.

How do you suggest to move forward with the failing build? I can definitely write ES5, but given [Pontoon's browser support matrix](https://mozilla-pontoon.readthedocs.io/en/latest/dev/contributing.html#browser-support), that seems quite a questionable limitation rather than a valid case for a _failing test_.

julen:
Scrolling behavior should now be fixed.

julen:
> How do you suggest to move forward with the failing build? I can definitely write ES5, but given [Pontoon's browser support matrix](https://mozilla-pontoon.readthedocs.io/en/latest/dev/contributing.html#browser-support), that seems quite a questionable limitation rather than a valid case for a _failing test_.

re. django-pipeline, I've checked the status of things and it doesn't look very difficult to switch JS compression over to terser (ES2015+ compatible successor of uglifyjs), and that's what I'd suggest myself. Any input on that @mathjazz?

abowler2:
> re. django-pipeline, I've checked the status of things and it doesn't look very difficult to switch JS compression over to terser (ES2015+ compatible successor of uglifyjs), and that's what I'd suggest myself. Any input on that @mathjazz?

If we went this route I believe we could eliminate the additional `prettierrc` file in `/pontoon` and unify the frontend and backend JS formatting.

julen:
> If we went this route I believe we could eliminate the additional `prettierrc` file in `/pontoon` and unify the frontend and backend JS formatting.

I might be missing something or maybe I didn't fully understand your comment in the context of django-pipeline, so I wonder how prettier's config location is relevant here.

In the meantime, I've opened a PR at #1819 implementing what I suggested earlier.

abowler2:
> > If we went this route I believe we could eliminate the additional `prettierrc` file in `/pontoon` and unify the frontend and backend JS formatting.
> 
> I might be missing something or maybe I didn't fully understand your comment in the context of django-pipeline, so I wonder how prettier's config location is relevant here.
> 

To clarify it would unify the JS formatting for both the `/frontend` and `/pontoon` folders. 

When I [moved Prettier into the main folder](https://github.com/mozilla/pontoon/pull/1749) we ran into the same issue with Uglify. The way we fixed it at that time was to add a `prettierrc` for `/pontoon` that changes the `trailing comma` rule from `all` to `es5`. But that means there is a slight difference in the JS formatting between `/frontend` and `/pontoon`. 

If we change the JS compression to be compatible with ES2015+ that should eliminate the need for that file.

mathjazz (approved):
Great work! Everything looks great now.

I'm approving this, and also blocking merging on #1819.

Good point @abowler2! We could also do that in a followup to #1819.

julen:
@abowler2 thanks for the extra context and explanation, that makes perfect sense now 👍

julen:
242 insertions(+), 3200 deletions(-)

58wT

PR #1818, opened 2021-01-17
No bug - package-lock.json: remove obsolete fluent-syntax

julen:
This should've been removed as part of c52a603.


mathjazz (approved):
Another excellent catch!

PR #1820, opened 2021-01-18
Fix bug 1687328: Replace Tab keyboard shortcut with Ctrl + Shift + Up/Down

mathjazz:
That leaves the default and expected behavior of the Tab key unchanged,
which is critical for accessibility and usability.

Also included:
- Add shortcut to the Locales tab tooltip
- Fix keyboard shortcut display in the panel

Deployed to stage:

### DIFF
--- a/frontend/src/core/editor/reducer.js
+++ b/frontend/src/core/editor/reducer.js
@@ -196,6 +196,7 @@ export default function reducer(
             return {
                 ...initial,
                 isRunningRequest: state.isRunningRequest,
+                selectedHelperTabIndex: state.selectedHelperTabIndex,
             };
         case UPDATE_MACHINERY_SOURCES:
             return {
--- a/frontend/public/static/locale/en-US/translate.ftl
+++ b/frontend/public/static/locale/en-US/translate.ftl
@@ -177,8 +177,11 @@ editor-KeyboardShortcuts--search-strings-shortcut = <mod1>Ctrl</mod1> + <mod2>Sh
 editor-KeyboardShortcuts--select-all-strings = Select All Strings
 editor-KeyboardShortcuts--select-all-strings-shortcut = <mod1>Ctrl</mod1> + <mod2>Shift</mod2> + <accel>A</accel>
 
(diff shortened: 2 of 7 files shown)


mathjazz:
@julen r?
@bidaian r?

julen:
I have added some thoughts in the original bug, and I'd hold off merging this until discussion has been clarified there.

julen:
Am I supposed to do something else than restarting the server? FWIW this is what I get when opening the shortcuts dialog:
Shortcuts window

Aside from this, and not strictly related to this particular issue, but I noticed how reading the dialog from top to bottom, down arrows come before up arrows (i.e. 'next' before 'prev'), which feels a bit counter-intuitive.

frontend/src/core/editor/reducer.js
```diff
@@ -196,6 +196,7 @@ export default function reducer(
             return {
                 ...initial,
                 isRunningRequest: state.isRunningRequest,
+                selectedHelperTabIndex: state.selectedHelperTabIndex,
```
julen, line 199: This change/fix seems to be unrelated to changing the keyboard shortcut.

mathjazz:
Sadly you need to `make build` first and then restart the server to apply the localization changes.

I've changed the order of Next/Previous.

frontend/src/core/editor/reducer.js
mathjazz: Correct, I thought it's a small enough fix to sneak it in as a ride-a-long. :)

julen: My personal preference is to keep such changes aside in order to make `git blame` useful in the future. Your call nonetheless.

julen (approved):
Looks and works good.

As a follow-up question remains if a focused textarea should be required in order to trigger these actions or not.

mathjazz:
> As a follow-up question remains if a focused textarea should be required in order to trigger these actions or not.

I'd leave that unchanged for now and open a separate bug if needed. I'd be open to investigate pros/cons of such change.

PR #1821, opened 2021-01-19
Fix bug 1686690: simplify 'otherlocales' response payload

julen:
* Unifies the initial state values to match what machinery uses
* Flattens the response payload, simplifies frontend code


pontoon/base/views.py
```diff
@@ -334,6 +334,7 @@ def _serialize_translation_values(query):
                 "script": translation["locale__script"],
             },
             "translation": translation["string"],
+            "is_preferred": translation["locale__code"] in preferred_values,
```
julen, line 337: We could leave this out entirely if the value is `False` and hence reduce the response payload.

mathjazz (commented):
Nice job, this is much cleaner.

It seems to be working very well. I only caught two `eslint` warnings (note that GH Actions are down ATM):
```
./src/modules/otherlocales/components/OtherLocales.js
  Line 64:60:  Expected to return a value at the end of arrow function  array-callback-return
  Line 72:60:  Expected to return a value at the end of arrow function  array-callback-return
```

mathjazz (approved):
Thanks for the update! This is good to go.

P.S.: I saw the eslint errors in the output of `make run`.

mathjazz:
@julen Do you want to squash commits and push them to your branch before I merge?

julen:
There you go!

PR #1822, opened 2021-01-22
Fix bug 1688174: Unify prettier configuration

mathjazz:
In addition to removing the config file in the `pontoon` folder and reformatting the code, I've also:
- removed the seemingly redundant `.prettierrc.json` file
- simplified the ` .prettierignore` file


.prettierignore
```diff
@@ -1,34 +1,13 @@
...
-
-# production
-/build
-
-# misc
-*.html
+# minified code
 *.min.js
 *.min.css
-*.yml
```
abowler2, line 14: I found that when `.html` and `.yml` files were not in the ignore file then if I made any changes in them with VSCode set to `format on save` it would reformat those files.  Of course, that could just be something I need to change in my local setup

.prettierignore
```diff
@@ -1,34 +1,13 @@
...
-**/base/templates/js/pontoon.js
 **/base/static/css/boilerplate.css
 **/base/static/css/fontawesome-all.css
 **/base/static/css/jquery-ui.css
 **/base/static/css/nprogress.css
+**/base/templates/js/pontoon.js
 **/in_context/static/css/agency.css
 **/in_context/static/js/agency.js
-**/frontend/build*
-**/package.json*
```
abowler2, line 24: This was added for the same reason as above.

.prettierignore
mathjazz: Gotcha. I'll restore these and the two you called out below. We don't want to make your life harder. :)

abowler2: Well now you don't need to cater to me (though I do appreciate when my live is easier 😁 ) I bring it up in case it isn't me and other run into the same issue.

With that said, I didn't intend for GitHub to highlight both lines 23 and 24 below.  The `**/frontend/build/*` file is likely fine to remove. Sorry about that.

I'll approve either way as having it in or out of the ignore file is not a big deal.

abowler2 (approved):

.prettierignore
mathjazz: :-)

Gotcha, makes sense re `**/frontend/build/*`.

abowler2 (approved):

PR #1823, opened 2021-01-24
Fix Add Comment input height to match submit button

mathjazz:
Add Comment input box in the Comments tab is smaller than the submit button.
We don't have that problem in translation comments, because line-height is
defined in History.css.


abowler2 (approved):
I'm pretty sure at the time there was a reason for setting the line-height there, but this all looks good so I'm either misremembering or another change somewhere along the line removed the need.  🤷🏻‍♀️

mathjazz:
Yeah, I checked blame and I didn't  get anything more meaningful out of it. I also confirmed that it works fine with mentions.

PR #1826, opened 2021-01-26
Fix bug 1681688 - Correctly update generic editor content when the editor is reset.

adngdb:
The issue here was that we did not trigger the `useLoadTranslation` effect when the editor was reset. We do need to trigger it when, and only when, the editor is reset, as that covers all the cases when the editor content needs to be set back to its initial value.

I'm using the opportunity to apply my own rules for hooks: the `useLoadTranslation` hook was only used by one component, so it went back into that component's file.


mathjazz:
> I'm using the opportunity to apply my own rules for hooks: the `useLoadTranslation` hook was only used by one component, so it went back into that component's file.

Yessssssssss!

mathjazz (commented):
Nice work! I'm no longer able to reproduce the bug in any scenario.

However, there's a regression. If you have a string with just one translation, which is approved, and you unapprove it and then try to approve it again, you get a failed check (Unchanged) warning.

adngdb:
> However, there's a regression. If you have a string with just one translation, which is approved, and you unapprove it and then try to approve it again, you get a failed check (Unchanged) warning.

I cannot reproduce that locally. Can you give me more context, or a string that triggers this?

mathjazz:
> I cannot reproduce that locally. Can you give me more context, or a string that triggers this?

Do you have Translate Toolkit checks disabled?

mathjazz:
Sorry, that's my bad. I was approving a translation that is identical to the source string, so it makes sense to show that warning. 😬

mathjazz (approved):

mathjazz:
*It's the least I can do...*

PR #1827, opened 2021-01-26
Fix bug 1688570: improve performance of initial editor change

julen:
Changing from/to initial textarea values would result in redundant
re-renders due to subscribing to `unsavedChangesExist` and
`unsavedChangesIgnored` redux state changes.

Implementation-wise, this accesses `unsavedchanges` state directly,
without subscribing to its changes because the state members are not
used for rendering purposes.

For direct store access react-redux strongly suggests to leverage the
`useStore()` API, because accessing the store via context is an
implementation detail and it might change in the future. That's why the
components connecting redux state -> props have been converted to hooks.

It'd also have been possible to avoid these intermediate components and
co-locate redux state directly within components, however that would
require more invasive changes, creeping the scope of this change
considerably.

Refs. https://react-redux.js.org/using-react-redux/accessing-store

----

Before:
before

After:
after

### DIFF
--- a/frontend/src/core/plural/components/PluralSelector.test.js
+++ b/frontend/src/core/plural/components/PluralSelector.test.js
@@ -2,8 +2,7 @@ import React from 'react';
 import { shallow } from 'enzyme';
 import sinon from 'sinon';
 
-import { createReduxStore } from 'test/store';
-import { shallowUntilTarget } from 'test/utils';
+import { createReduxStore, mountComponentWithStore } from 'test/store';
 
 import { actions } from '..';
 
@@ -78,20 +77,31 @@ describe('<PluralSelector>', () => {
             },
             router: {
                 location: {
-                    pathname: '/kg/pro/all/',
+                    pathname: '/kg/firefox/all-resources/',
                     search: '?string=42',
                 },
+                // `action` is required because
+                // https://github.com/supasate/connected-react-router/issues/312#issuecomment-500968504
+                // Please note the initial `LOCATION_CHANGE` action can and must
+                // be supressed via the `noInitialPop` prop in
+                // `ConnectedRouter`, otherwise it'll cause side-effects like
+                // executing reducers and hence altering initial state values.
+                action: 'some-string-to-please-connected-react-router',
             },
         };
         const store = createReduxStore(initialState);
         const dispatchSpy = sinon.spy(store, 'dispatch');
 
-        const wrapper = shallowUntilTarget(
-            <PluralSelector store={store} resetEditor={sinon.spy()} />,
-            PluralSelectorBase,
-        );
+        const root = mountComponentWithStore(PluralSelector, store, {
+            resetEditor: sinon.spy(),
+        });
+        const wrapper = root.find(PluralSelectorBase);
 
-        wrapper.find('button').first().simulate('click');
+        const button = wrapper.find('button').first();
+        // `simulate()` doesn't quite work in conjunction with `mount()`, so
+        // invoking the `prop()` callback directly is the way to go as suggested
+        // by the enzyme maintainer...
+        button.prop('onClick')();
 
         const expectedAction = actions.select(0);
         expect(dispatchSpy.calledWith(expectedAction)).toBeTruthy();
--- a/frontend/src/modules/entitieslist/components/EntitiesList.test.js
+++ b/frontend/src/modules/entitieslist/components/EntitiesList.test.js
@@ -1,8 +1,6 @@
-import React from 'react';
 import sinon from 'sinon';
 
-import { createReduxStore } from 'test/store';
-import { shallowUntilTarget } from 'test/utils';
+import { createReduxStore, mountComponentWithStore } from 'test/store';
 
 import * as entities from 'core/entities';
 import * as navigation from 'core/navigation';
@@ -11,7 +9,10 @@ import * as batchactions from 'modules/batchactions';
 import EntitiesList, { EntitiesListBase } from './EntitiesList';
 
 // Entities shared between tests
-const ENTITIES = [{ pk: 1 }, { pk: 2 }];
+const ENTITIES = [
+    { pk: 1, translation: [{ string: '', errors: [], warnings: [] }] },
+    { pk: 2, translation: [{ string: '', errors: [], warnings: [] }] },
+];
 
 describe('<EntitiesList>', () => {
     beforeAll(() => {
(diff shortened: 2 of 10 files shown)


julen:
Tests for changed components are currently broken. I've been banging my head to find a solution but I can't manage to shallow render components that use react-redux hooks.

@adngdb any hints? I'm really out of ideas.

adngdb:
Hey @julen, I don't remember if it's possible to "shallow" components. There are however several examples of mounting components, like this one: https://github.com/mozilla/pontoon/blob/master/frontend/src/modules/genericeditor/components/GenericEditor.test.js#L44

Does that help, or are you looking for something else?

julen:
Thanks @adngdb, that set me in the right track and I've been able to fix the tests... not without sweating though! I'd be grateful if you could now review the changes for me.

frontend/src/core/plural/components/PluralSelector.test.js
```diff
@@ -78,20 +78,28 @@ describe('<PluralSelector>', () => {
             },
             router: {
                 location: {
-                    pathname: '/kg/pro/all/',
+                    pathname: '/kg/firefox/all-resources/',
                     search: '?string=42',
                 },
+                // `action` is required because
+                // https://github.com/supasate/connected-react-router/issues/312#issuecomment-500968504
```
julen, line 85: Super leaky implementation detail 🤦

frontend/src/core/plural/components/PluralSelector.test.js
```diff
@@ -78,20 +78,28 @@ describe('<PluralSelector>', () => {
...
-            PluralSelectorBase,
-        );
+        const root = mountComponentWithStore(PluralSelector, store, {
+            resetEditor: sinon.spy(),
+        });
+        const wrapper = root.find(PluralSelectorBase);
 
-        wrapper.find('button').first().simulate('click');
+        const button = wrapper.find('button').first();
+        button.prop('onClick')();
```
julen, line 102: I really struggled with this one. Apparently `.simulate()` doesn't quite work when components are mounted (as opposed to shallow-rendered). If you read enzyme GitHub issues, the maintainer strongly suggests to stay away from `.simulate()` because it doesn't do what it appears to be doing, and instead suggests to directly invoke the component's handler function.

frontend/src/modules/entitieslist/components/EntitiesList.test.js
```diff
@@ -11,7 +11,10 @@ import * as batchactions from 'modules/batchactions';
 import EntitiesList, { EntitiesListBase } from './EntitiesList';
 
 // Entities shared between tests
-const ENTITIES = [{ pk: 1 }, { pk: 2 }];
+const ENTITIES = [
+    { pk: 1, translation: [{ string: '', errors: [], warnings: [] }] },
```
julen, line 15: Because we are now deep rendering (i.e. `.mount()`), the passed in state needs to have more members so that children components don't crash and can render.

frontend/src/modules/entitydetails/components/EntityDetails.test.js
```diff
@@ -21,6 +22,8 @@ const ENTITIES = [
                 warnings: [],
             },
         ],
+        project: { contact: '' },
```
julen, line 25: Same like before, this is to be able to deep render the component tree under test.

frontend/src/modules/entitydetails/components/EntityDetails.test.js
```diff
@@ -209,9 +213,11 @@ describe('<EntityDetails>', () => {
     });
 
     it('dispatches the updateStatus action when updateTranslationStatus is called', () => {
-        const [wrapper] = createEntityDetailsWithStore();
+        const [wrapper, store] = createEntityDetailsWithStore();
 
         wrapper.instance().updateTranslationStatus(42, 'fake translation');
+        // Proceed with changes even if unsaved
+        store.dispatch(unsavedchanges.actions.ignore());
```
julen, line 220: I couldn't explain why this wasn't needed before, any ideas @adngdb?

frontend/src/test/store.js
```diff
@@ -31,7 +31,7 @@ export function createReduxStore(initialState = {}) {
 export function mountComponentWithStore(Component, store, props = {}) {
     return mount(
         <Provider store={store}>
-            <ConnectedRouter history={history}>
+            <ConnectedRouter history={history} noInitialPop>
```
julen, line 34: This prop is required to omit an initial navigation dispatch from the router, which could have side-effects like resetting some initial state passed to the root reducer factory function. Please check the comment in _PluralSelector.test.js_ for further details.

frontend/src/modules/entitydetails/components/EntityDetails.test.js
adngdb: That's weird. Maybe the unsavedchanges state is "polluted" and thus must be ignored before the action is actually called? You could check that by printing the value of `store.getState().unsavedchanges` before this line, to see if it has `exists` as true. As for why that would be, I do not know! Either a precedent test failed to clean the state, or there is something in the setup that triggers the unsavedchanges action.

frontend/src/test/store.js
adngdb: Here again, can you write this down in the code too? Thanks!

adngdb (approved):
At first I was wary of the mix of class components and functional components, but I actually think this is a nice step towards rewriting everything to functions, so yay! I made a few comments, mostly about adding more comments. I like the graphs you showed, and I didn't find any bug in my quick testing, so great! :-)

frontend/src/modules/entitydetails/components/EntityDetails.test.js
julen: Hmmm I went through the possible code paths again, and the main difference now being how `EntityDetails` is mounted, this also means underlying components are mounted, so the logic under `GenericTranslationForm` will be executed, which in turn fires the `unsavedchanges.actions.update()` action via the `useUpdateUnsavedChanges` hook.

And that brings us now to have the `unsavedchanges.actions.exists` populated to `true`, and because we are invoking `EntityDetail` component's `.updateTranslationStatus()` directly, that will only execute the `history.actions.updateStatus()` action when no unsaved changes exist (as part of the callback function).

Long story short: mounting brings us closer to real user interactions, and the line I added makes sense in this context.

frontend/src/core/plural/components/PluralSelector.test.js
julen: Sure, done.

frontend/src/test/store.js
julen: Done as well 👍

julen:
> At first I was wary of the mix of class components and functional components, but I actually think this is a nice step towards rewriting everything to functions, so yay!

Yeah I'm not a big fan either (life was simpler when `React.createClass()` was the sole API!), but as I explained in the original PR comment, this was necessary in order to keep the scope of changes as small as possible. Initially I started to rewrite the components as functional components (as per the react-redux `useState()` hook requirement), but the scope of changes was going way too far away.

adngdb (approved):

mathjazz:
Cool work, @julen! 👏 

Could you please squash commits? Same for #1833.

julen:
Sure, done!

PR #1829, opened 2021-01-28
 Bug 1686569 Use Element.scrollIntoView() to scroll to the selected entity

Mitch4sho:
Added the new method scrollIntoView and remove some redundant code

### DIFF
--- a/frontend/src/modules/entitieslist/components/EntitiesList.js
+++ b/frontend/src/modules/entitieslist/components/EntitiesList.js
@@ -107,7 +107,6 @@ export class EntitiesListBase extends React.Component<InternalProps> {
 
         // Scroll to selected entity when entity changes
         // and when entity list loads for the first time
-
         if (
             previous.entity !== current.entity ||
             (!prevProps.entities.entities.length &&
@@ -116,13 +115,10 @@ export class EntitiesListBase extends React.Component<InternalProps> {
             const list = this.list.current;
             const element = list.querySelector('li.selected');
 
-            if (!element) {
-                return;
-            } else
-                element.scrollIntoView({
-                    behavior: 'smooth',
-                    block: 'nearest',
-                });
+            element.scrollIntoView({
+                behavior: 'smooth',
+                block: 'nearest',
+            });
         }
     }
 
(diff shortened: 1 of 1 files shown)


frontend/src/modules/entitieslist/components/EntitiesList.js
```diff
@@ -107,12 +107,22 @@ export class EntitiesListBase extends React.Component<InternalProps> {
 
         // Scroll to selected entity when entity changes
         // and when entity list loads for the first time
+
```
mathjazz, line 110: Nit: you can remove this line. :-)

frontend/src/modules/entitieslist/components/EntitiesList.js
```diff
@@ -107,12 +107,22 @@ export class EntitiesListBase extends React.Component<InternalProps> {
...
             (!prevProps.entities.entities.length &&
                 this.props.entities.entities.length)
         ) {
-            this.scrollToSelectedElement();
+            const list = this.list.current;
+            const element = list.querySelector('li.selected');
+
+            if (!element) {
+                return;
+            } else
```
mathjazz, line 121: You can remove that `else` and dedent the `element.scrollIntoView()` block underneath.
(You might add a blank line after the `if` block though.)

After the `if` with a `return` statement, there's no need for the `else` part.

```js
if (condition) {
  return something;
} else {
  // do another thing
}
```

is logically the same as

```js
if (condition) {
  return something;
}

// do another thing
```

because `// do another thing` only executes when the condition is `false` anyways (because you `return` early if condition is `true`).

mathjazz (commented):
Excellent job! Works like a charm. 👍 

Just a couple of stylistic nits and we merge!

frontend/src/modules/entitieslist/components/EntitiesList.js
Mitch4sho: I got it. I'm going to correct it right away

julen:
@mathjazz isn't the `if (!element) return;` now missing?

Mitch4sho:
Oh, was I not supposed to remove it?

mathjazz:
Whoops, my bad. @Mitch4sho Could you please restore that?

PR #1831, opened 2021-01-29
Fix bug 1689653: avoid lightbox actions when closed

julen:
The editor would constantly fire `lightbox/CLOSE` actions upon
pressing any of space/esc/enter keys.


mathjazz (approved):
Great work!

PR #1833, opened 2021-01-30
Bug 1688571: add discard functionality hook

julen:
This will be used as a replacement for react-onclickoutside, because the functionality only needs a few lines of code and it's more explicit than using a HoC.

Because 10 components of different complexity are affected, it's going to be way easier to review if we move with individual PRs, one component at a time. So the plan would be:
1. Introduce this hook
2. Replace every single use of `onClickOutside()` with `useOnDiscard()`
3. Remove react-onclickoutside

### DIFF
--- a/frontend/src/core/plural/selectors.js
+++ b/frontend/src/core/plural/selectors.js
@@ -24,7 +24,7 @@ export function _getPluralForm(pluralForm: number, entity: ?Entity): number {
  */
 export const getPluralForm: Function = createSelector(
     pluralSelector,
-    entities.selectors.getSelectedEntity,
+    (state) => entities.selectors.getSelectedEntity(state),
     _getPluralForm,
 );
 
@@ -55,7 +55,7 @@ export function _getTranslationForSelectedEntity(
  * most recent non-rejected one.
  */
 export const getTranslationForSelectedEntity: Function = createSelector(
-    entities.selectors.getSelectedEntity,
+    (state) => entities.selectors.getSelectedEntity(state),
     getPluralForm,
     _getTranslationForSelectedEntity,
 );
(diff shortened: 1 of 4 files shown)


frontend/src/core/plural/selectors.js
```diff
@@ -24,7 +24,7 @@ export function _getPluralForm(pluralForm: number, entity: ?Entity): number {
  */
 export const getPluralForm: Function = createSelector(
     pluralSelector,
-    entities.selectors.getSelectedEntity,
+    (state) => entities.selectors.getSelectedEntity(state),
```
adngdb, line 27: Why are you making this change?

adngdb (approved):
This is looking real good! Nice and simple, with tests, great work. :-)

frontend/src/core/plural/selectors.js
julen: As I explained in the commit message, this change is necessary to avoid circular import errors.

In subsequent PRs this import would fail:
```
import { useOnDiscard } from 'core/utils';
```
with a very cryptic error message. Delaying the evaluation of `entities.selectors` solved the issue.

adngdb: Ah, sorry, I didn't look at the commit messages. Thanks! :)

adngdb (approved):

julen:
@mathjazz squashed this one too. Once merged into master, I'll make sure to rebase the rest of the open PRs so that reviewing them becomes way easier.

mathjazz:
Thanks (I was just about to ask you for that). :-)

julen:
@mathjazz done — it'd be great to slowly but surely start moving forward with them.

PR #1834, opened 2021-01-30
Bug 1688571: fix negative performance side-effects of `UserMenu`

julen:
This now only listens for `click` events when the menu is visible by the means of leveraging the `useOnDiscard` hook.


mathjazz (approved):
Very well done! Apart from the performance win, we can also drop a dependency after the last component is migrated.

Adding a high-level comment that refers to all related PRs: The code looks good. I've tested the PR locally and didn't spot an issues.

PR #1842, opened 2021-02-02
Update pontoon.js

Krshivam25:
Updated Line 669 as there is one Syntax Error i.e; unexpected tokens.


mathjazz:
This is jinja2 code: the file changes dynamically for each deployment, which is also why it's locates in the `templates` folder.

PR #1846, opened 2021-02-03
Bug 1690485: Use HTTPS for the Microsoft Terminology endpoint

mathjazz:
Microsoft Terminology added support for HTTPS, so let's use it! We still have to use HTTP in the SOAP header, though - otherwise it 500s.

Note that the issue with the service availability has been resolved independently of this change.

### DIFF
--- a/pontoon/machinery/templates/machinery/microsoft_terminology.jinja
+++ b/pontoon/machinery/templates/machinery/microsoft_terminology.jinja
@@ -1,9 +1,9 @@
 <?xml version='1.0' encoding='utf-8'?>
 <soap-env:Envelope xmlns:soap-env="http://schemas.xmlsoap.org/soap/envelope/">
     <soap-env:Header>
-        <wsa:Action xmlns:wsa="http://www.w3.org/2005/08/addressing">http://api.terminology.microsoft.com/terminology/Terminology/GetTranslations</wsa:Action>
-        <wsa:MessageID xmlns:wsa="http://www.w3.org/2005/08/addressing">urn:uuid:{{ uuid }}</wsa:MessageID>
-        <wsa:To xmlns:wsa="http://www.w3.org/2005/08/addressing">http://api.terminology.microsoft.com/Terminology.svc</wsa:To>
+        <wsa:Action xmlns:wsa="https://www.w3.org/2005/08/addressing">https://api.terminology.microsoft.com/terminology/Terminology/GetTranslations</wsa:Action>
+        <wsa:MessageID xmlns:wsa="https://www.w3.org/2005/08/addressing">urn:uuid:{{ uuid }}</wsa:MessageID>
+        <wsa:To xmlns:wsa="https://www.w3.org/2005/08/addressing">https://api.terminology.microsoft.com/Terminology.svc</wsa:To>
     </soap-env:Header>
     <soap-env:Body>
         <ns0:GetTranslations xmlns:ns0="http://api.terminology.microsoft.com/terminology">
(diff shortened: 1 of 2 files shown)


pontoon/machinery/templates/machinery/microsoft_terminology.jinja
```diff
@@ -1,9 +1,9 @@
 <?xml version='1.0' encoding='utf-8'?>
 <soap-env:Envelope xmlns:soap-env="http://schemas.xmlsoap.org/soap/envelope/">
     <soap-env:Header>
-        <wsa:Action xmlns:wsa="http://www.w3.org/2005/08/addressing">http://api.terminology.microsoft.com/terminology/Terminology/GetTranslations</wsa:Action>
-        <wsa:MessageID xmlns:wsa="http://www.w3.org/2005/08/addressing">urn:uuid:{{ uuid }}</wsa:MessageID>
-        <wsa:To xmlns:wsa="http://www.w3.org/2005/08/addressing">http://api.terminology.microsoft.com/Terminology.svc</wsa:To>
+        <wsa:Action xmlns:wsa="https://www.w3.org/2005/08/addressing">https://api.terminology.microsoft.com/terminology/Terminology/GetTranslations</wsa:Action>
```
Pike, line 4: Is there documentation that says that these xml namespaces are OK to be `https`?
I only find `http` ad hoc.

mathjazz: Good point! Updated.

Pike (changes requested):
Ugh, this stuff is hard. Harder than it should be, actually.

Which is why I'd like to scope-creep a little, sorry.

I'd prefer the fix to the template to be a removal of the `<soap-env:Header>`, or at least its children. The Action name is namespace/nodeName, thus the fix is wrong, but also AFAICT, not needed at all. Just an empty element should do.

I'm also having a nit about `ns0`, we should just use the default namespace, or the `tns` prefix.

Interestingly, the HTTP headers seem to be required. Maybe either an HTTP header or a SOAP header would work, but I'm all for doing one and not both, and I guess I prefer the http header, as we need to set the charset anyway (according to my experiments locally).

I'm also unhappy about the xpath logic in the response processing, but maybe I'll take that as a follow-up myself.

mathjazz:
I have no idea what I'm doing. 🤷‍♂️

I think I've addressed the comments, except for the one quoted below. Or maybe there's no action required there?

> Interestingly, the HTTP headers seem to be required. Maybe either an HTTP header or a SOAP header would work, but I'm all for doing one and not both, and I guess I prefer the http header, as we need to set the charset anyway (according to my experiments locally).

pontoon/machinery/templates/machinery/microsoft_terminology.jinja
```diff
@@ -1,20 +1,16 @@
 <?xml version='1.0' encoding='utf-8'?>
 <soap-env:Envelope xmlns:soap-env="http://schemas.xmlsoap.org/soap/envelope/">
-    <soap-env:Header>
-        <wsa:Action xmlns:wsa="http://www.w3.org/2005/08/addressing">http://api.terminology.microsoft.com/terminology/Terminology/GetTranslations</wsa:Action>
-        <wsa:MessageID xmlns:wsa="http://www.w3.org/2005/08/addressing">urn:uuid:{{ uuid }}</wsa:MessageID>
```
Pike, line 5: You can drop the `uuid` from `views.py` as well.

Pike (approved):
Clean-up nit left on `views.py`, but this looks good.

As you mentioned, I also don't know what I'm doing, and thus I researched quite a bit. And removing stuff I didn't research sounded like a good idea, instead of doing more research :-)

PR #1847, opened 2021-02-03
Bug 1688571: fix negative performance side-effects of `FiltersPanel`

julen:
This now only listens for `click` events when the menu is visible by the means of leveraging the `useOnDiscard` hook

### DIFF
--- a/frontend/src/modules/search/components/FiltersPanel.css
+++ b/frontend/src/modules/search/components/FiltersPanel.css
@@ -59,10 +59,6 @@
     position: relative;
 }
 
-.filters-panel.fixed .menu ul {
-    margin-bottom: 54px;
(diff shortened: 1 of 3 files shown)


julen:
I haven't been able to verify the proper functioning of the time range filter (not enough data available on my side perhaps?), so another pair of eyes checking that would be appreciated.

mathjazz:
Seems like this was the last component to use `react-onclickoutside`, so we can remove the dependency.

julen:
Yes, I have a separate PR for that.

mathjazz (approved):
Really nice work, julen! Everything works as expected.

Bonus points for the CSS implementation of the sticky toolbar!

PR #1848, opened 2021-02-03
Fix bug 1690500: Only collect and pass user data once

mathjazz:
We currently collect user data twice in the `EntityDetails` component and pass it as two different props (`user` and `users`):

1. EntityDetails > History > Translation > CommentsList > AddComment
1. EntityDetails > Helpers > TeamComments > CommentsList > AddComment

It's enough to collect and pass the same data once. :)


mathjazz:
@julen Would you be up for a review?

julen (approved):
Less is more!

PR #1852, opened 2021-02-03
Fix bug 1690563: Restore TM in custom Machinery search

mathjazz:
In separate commits, I've made two additional changes:
* Removed Caighdean service from searching on the /machinery page, because it takes `gd` string as input, so it's only usable in the translate view, where we have an entity and we can get to a `gd` translation through it.
* Removed `customSearch` attribute, which is always true. We needed it when /machinery and the legacy translate page shared the same Machinery frontend code.


pontoon/base/static/js/main.js
```diff
@@ -391,10 +391,9 @@ var Pontoon = (function (my) {
          * Get suggestions from machine translation and translation memory
          *
          * original Original string
-         * customSearch Instead of source string, use custom search keyword as input
          * loader Loader element id
          */
-        getMachinery: function (original, customSearch, loader) {
+        getMachinery: function (original, loader) {
```
Pike, line 396: You'll also need to fix https://github.com/mozilla/pontoon/blob/a9aacb6ce812a9f95fbdc107986a416cfe193f96/pontoon/machinery/static/js/machinery.js#L11 for this. I haven't found more call sites?

Pike (changes requested):
This looks good overall, but there's one bug left, you changed the signature of `getMachinery`, but not the call site(s).

pontoon/base/static/js/main.js
mathjazz: Nice catch. There is indeed only this call site remaining.

Which means we should move `getMachinery` to machinery.js. I could do also do it as part of this PR.

Pike: I agree, and I think we should do this as part of this PR. Much of the changes are really about `getMachinery` not being a multi-purpose API anymore. Guess this is fallout from TN? That this used to be shared code between translate and machinery UI?

mathjazz: That's correct. OK, I'll push a separate commit.

I've also identified several other unused pieces of code in main.js. These will go into a separate PR.

mathjazz: I'll rebase #1851 once this is merged.

mathjazz: No I won't.

Pike (approved):

PR #1853, opened 2021-02-03
Fix bug 1690642: Cleanup main.js file

mathjazz:
`main.js` file is loaded on each page, except for the translate view.

It contains code, which is no longer used, because it has been replaced by the new translate app (Translate.Next).

There's also code, which is only used by one view (as discovered in bug 1690563), so it should be moved to view-specific files.

And then there's code that is obsolete, or can be rewritten using native DOM APIs.

Let's clean it up!

### DIFF
--- a/pontoon/base/static/css/style.css
+++ b/pontoon/base/static/css/style.css
@@ -512,8 +512,7 @@ tfoot td a {
     padding: 2px 4px;
 }
 
-.menu li.hover,
-.menu .static-links div.hover {
+.menu li.hover {
     color: #ffffff;
     background: #3f4752;
 }
--- a/pontoon/base/static/css/style.css
+++ b/pontoon/base/static/css/style.css
@@ -282,6 +282,10 @@ tfoot td a {
     z-index: 20; /* Must be higher than iframe-cover */
 }
 
+.select .menu.permanent {
+    z-index: 19; /* Must be lower than for the (popup) .menu */
+}
+
 .select > .button.breadcrumbs,
 #go {
     background: #3f4752;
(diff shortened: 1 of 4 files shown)


mathjazz:
Axel, this should be the last one. 

Before we hit CSS next week. 😆

pontoon/base/static/js/main.js
```diff
@@ -398,97 +269,43 @@ $(function () {
...
-            success: function () {
-                self.toggleClass('enabled');
-                var is_enabled = self.is('.enabled'),
-                    status = is_enabled ? 'enabled' : 'disabled';
-
-                Pontoon.endLoader(self.text() + ' ' + status + '.');
-
-                if (self.is('.force-suggestions') && Pontoon.user) {
-                    Pontoon.user.forceSuggestions = is_enabled;
-                    Pontoon.postMessage('UPDATE-ATTRIBUTE', {
```
Pike, line 444: You dropped the postMessage here. I guess you know why, but should you also drop listening to `UPDATE-ATTRIBUTE` in `pontoon.js`, given this was the only send location?

pontoon/base/static/css/style.css
```diff
@@ -282,6 +282,10 @@ tfoot td a {
     z-index: 20; /* Must be higher than iframe-cover */
```
Pike, line 282: This comment is wrong these days?

Pike (approved):
Looks good to me, the question about the message passing is related to in-context UI, right? Which is broken anyway?

pontoon/base/static/js/main.js
mathjazz: We've removed all other `postMessage` calls as part of removing Translate.Dead - most of them were in the `translate.js` file, which we removed completely.

We haven't changed pontoon.js as part of that process, so I suggest we keep it unchanged until we take on https://bugzilla.mozilla.org/show_bug.cgi?id=1536884.

PR #1854, opened 2021-02-05
Fix bug 1688573: avoid redundant editor re-renders

julen:
This PR avoids major part of redundant re-renders in the helper components. As I have observed, the components making use of diff functionality as well as the `Localized` components incur in perceptible performance cost.

Before:
before

After:
after

I think the result is good enough for the sake of bug 1688573.

It's worth noting though, there are a few other areas that could still be improved:
* The first keystroke produces a re-render in the helpers components because the editor's `changeSource` changes.
* `EditorMenu` > `MenuContent` constantly re-renders due to translation and translation length being changed. This becomes more expensive than expected because the button texts are wrapped in `Localized` elements.
* Every keystroke of the translation textarea triggers a series of redux actions, which could potentially be minimized to reduce the amount of work done per keystroke.

### DIFF
--- a/frontend/src/core/diff/withDiff.js
+++ b/frontend/src/core/diff/withDiff.js
@@ -37,11 +37,11 @@ type Props = {
 export default function withDiff<Config: Object>(
     WrappedComponent: React.AbstractComponent<Config>,
 ): React.AbstractComponent<Config> {
-    return function WithDiff(props: { ...Config, ...Props }) {
+    return React.memo(function WithDiff(props: { ...Config, ...Props }) {
         return (
             <WrappedComponent {...props}>
                 {getDiff(props.diffTarget, props.children)}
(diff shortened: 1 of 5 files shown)


mathjazz:
LGTM, thanks!

Tested locally quickly and didn't spot any issues.

@adngdb Could you also have a look please?

adngdb (approved):
These are all very sound and smart changes! :relaxed:

mathjazz:
Thanks, Adrian!

@julen It's be great if you could squash commits and reference the bug and the issue in the commit message :)

julen:
Sure, done. (I guess you meant PR# reference).

mathjazz:
> Sure, done. (I guess you meant PR# reference).

I did. :) Thanks!

PR #1855, opened 2021-02-06
Update to translate toolkit 3.3.2

mathjazz:
* New API: `translate.lang.data.normalized_unicode` -> `translate.lang.data.normalize` ([click](https://github.com/translate/translate/commit/cc4ed15f79eb233e500c8a8c178f733d134fbb63))
* Drop deprecated `get_rich_source()` method ([click](https://github.com/translate/translate/commit/19abe9f19d4b5d3b645ce1986e5fa9402173fe8d))
* Ensure xml:space="preserve" is set when updating target ([click](https://github.com/translate/translate/commit/a17c23bbf344d3b614667e361562aab98d696774))

### DIFF
--- a/pontoon/checks/libraries/translate_toolkit.py
+++ b/pontoon/checks/libraries/translate_toolkit.py
@@ -5,8 +5,8 @@ from translate.storage import base as storage_base
 
 def run_checks(original, string, locale_code, disabled_checks=None):
     """Check for obvious errors like blanks and missing interpunction."""
-    original = lang_data.normalized_unicode(original)
-    string = lang_data.normalized_unicode(string)
+    original = lang_data.normalize(original)
+    string = lang_data.normalize(string)
     disabled_checks = disabled_checks or []
 
     unit = storage_base.TranslationUnit(original)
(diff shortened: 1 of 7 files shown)


Pike:
I've followed the different versions of the patch here, and I'm a bit surprised how often the tests pass, and yet the tests changed in some ways or not.

That makes me wonder, are the tests just to easy on things, or was there something else going on?

Asking because I recall how fishy XCode is about xliff and its structure.

mathjazz:
We could have more (or stricter) tests, for sure. However...

The two changes I've included in the test by mistake (and later removed), both affected the template we use to generate the XLIFF files **before and after** saving translation changes in the tests. That these changes don't have impact on the result of the tests, proves that Pontoon keeps the changed part of the file unchanged on export.

As part of [bug 1676336](https://bugzilla.mozilla.org/show_bug.cgi?id=1676336) we might want to become more opinionated about the export structure.

Note that motivation for this update is the same as in https://github.com/mozilla/pontoon/commit/8c463bad78fa2d9ce04ff9bc37d3c8e8628d04da.

Pike (approved):

PR #1857, opened 2021-02-06
Fix bug 1691180: avoid native shortcut when pressing ctrl+shift+c

julen:
Please note the cleanup commit is to avoid things like this from happening in the future.

### DIFF
--- a/frontend/src/core/editor/hooks/useHandleShortcuts.js
+++ b/frontend/src/core/editor/hooks/useHandleShortcuts.js
@@ -50,7 +50,6 @@ export default function useHandleShortcuts() {
             copyOriginalIntoEditorCustom || copyOriginalIntoEditor;
 
         const key = event.keyCode;
-        let handledEvent = false;
 
         // Disable keyboard shortcuts when editor is in read only.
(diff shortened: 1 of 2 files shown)


mathjazz (approved):
Nice job!

Given that the 2nd commit cancels out the 1st one, do you want to squash them?

Please also add a reference to the PR in the commit message. I know this is awkward, because it's hard to do it in advance, but I find it to be a very valuable feature of the "Squash and merge" action in GH.

julen:
Done.

By the way, regarding the second point, you should note the GH UI already offers merged PR# references even if these are not part of commit messages ([example](https://github.com/mozilla/pontoon/commit/8caa910ac493f4c2f2b1ab9675e5fcbb63b7061e)).

mathjazz:
> By the way, regarding the second point, you should note the GH UI already offers merged PR# references even if these are not part of commit messages ([example](https://github.com/mozilla/pontoon/commit/8caa910ac493f4c2f2b1ab9675e5fcbb63b7061e)).

Oh, thanks! I wasn't aware of that. No need for PR references in commit messages in this case. :)

PR #1859, opened 2021-02-08
Update section about code formatting

mathjazz:
- Remove differences in /pontoon and /fronten JS code conventions
- Factor out motivation for code formatting and generic make format command to a separate section
- Add note aboute ignoring eslint errors

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -77,6 +77,23 @@ The following is the `browser support matrix of Pontoon <https://browserl.ist/?q
     iOS >= 10.3
 
 
+Code style
+==========
+
+We use code formatters so that we cannot fight over code style.
+You are free to write code however you like, because in the end formatter is the one
+that will format it. We thus don't need to pay attention to style during
+code reviews, and are free from those never-ending code style discussions.
+
+To format the Python and the JavaScript code at once you can use:
+
+.. code-block:: shell
+
+    $ make format
+
+Code formatting is explained in more detail in the following sections.
+
+
 Python code conventions
 =======================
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -77,6 +77,23 @@ The following is the `browser support matrix of Pontoon <https://browserl.ist/?q
     iOS >= 10.3
 
 
+Code style
+==========
+
+We use code formatters so that we cannot fight over code style.
```
abowler2, line 83:
```suggestion
We use code formatters so that we do not have to fight over code style.
```

CONTRIBUTING.rst
```diff
@@ -77,6 +77,23 @@ The following is the `browser support matrix of Pontoon <https://browserl.ist/?q
...
+You are free to write code however you like, because in the end formatter is the one
```
abowler2, line 84:
```suggestion
You are free to write code however you like, because in the end the formatter is the one
```

CONTRIBUTING.rst
```diff
@@ -130,24 +137,16 @@ Alternatively, you can format your code using:
...
-For more specifics about the ```frontend`` folder, look at the README.md file there.
-
-
-.. A note about formatting...::
-
-    To format both the frontend and Python code at once you can use:
-
-    .. code-block:: shell
+In the rare case when you cannot fix an eslint error, use ``// eslint-disable`` to make the linter
+ignore that error. Note that in most cases, it is better to fix the issues than ignoring them.
```
abowler2, line 147:
```suggestion
ignore that error. Note that in most cases, it is better to fix the issues than ignore them.
```

abowler2 (commented):
Just a few adjustments for clarity

abowler2 (approved):

PR #1861, opened 2021-02-09
Remove unused brace package

mathjazz:
The package was introduced by bug 1433994 (#1162), which brought support for the Ace Editor.

The editor was removed by bug 1563457 (#1382), but the package remained.


adngdb (approved):
Good catch!

PR #1862, opened 2021-02-09
Fix bug 1691805: Do not make requests to MT services if locale not supported

mathjazz:
We're getting lots of 400 errors:
https://bugzilla.mozilla.org/show_bug.cgi?id=1691805

### DIFF
--- a/pontoon/machinery/views.py
+++ b/pontoon/machinery/views.py
@@ -123,18 +123,14 @@ def microsoft_translator(request):
         if not locale_code:
             raise ValueError("Locale code is empty")
 
+        api_key = settings.MICROSOFT_TRANSLATOR_API_KEY
+        if not api_key:
+            raise ValueError("Missing api key")
+
     except (MultiValueDictKeyError, ValueError) as e:
         return JsonResponse(
             {"status": False, "message": "Bad Request: {error}".format(error=e)},
             status=400,
-        )
-
-    api_key = settings.MICROSOFT_TRANSLATOR_API_KEY
-
-    if not api_key:
-        log.error("MICROSOFT_TRANSLATOR_API_KEY not set")
-        return JsonResponse(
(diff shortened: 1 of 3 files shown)


mathjazz:
@Pike r pretty please?

pontoon/machinery/views.py
```diff
@@ -134,7 +134,10 @@ def microsoft_translator(request):
         )
 
     # Validate if locale exists in the database to avoid any potential XSS attacks.
-    if not Locale.objects.filter(ms_translator_code=locale_code).exists():
+    if (
+        not locale_code
+        or not Locale.objects.filter(ms_translator_code=locale_code).exists()
+    ):
```
Pike, line 140: Now that we only allow these requests for logged-in users, this query lost its purpose. I think we should just remove them.

pontoon/machinery/views.py
```diff
@@ -240,6 +247,8 @@ def systran_translate(request):
             status=404,
         )
 
+    locale = Locale.objects.get(systran_translate_code=locale_code)
```
Pike, line 250: This could be ambiguous?

Pike (changes requested):
I wish this stuff worked differently, but that's also out of scope, I guess. If we'd eventually chain the MT engines server-side, we'd send the original locale code, and not the custom one. And then bridge that gap on the server side.

I think the frontend changes are OK, but the python parts bother me.

mathjazz:
I've addressed comments in https://github.com/mozilla/pontoon/pull/1862/commits/f654924aa8c08785e18ec31c7dc17eb5ddc3fa33 and https://github.com/mozilla/pontoon/pull/1862/commits/cffb0ea6368a5e070fd01d813c4bd481e05c1147.

In between I've made an additional refactoring.

Pike (approved):
Thanks, this makes more sense now :-)

PR #1864, opened 2021-02-10
Bug 1691959, upgrade dependencies, and script updates to make that easy

Pike:
I've spent a while on making sure the upgrades make sense, by looking at the release notes and dependency chains. Sadly, `pip-compile` doesn't put the ranges into the requirements file, so that's a rather manual process.

The most funky thing is the `MarkupSafe` hashes, but I confirmed that they've been uploading more wheels for a single release over time. 🤷


mathjazz (commented):
Nice work!

Could you document `make requirements --upgrade` in the "Dependencies" section of the `CONTRIBUTING` file?

Pike:
Oh, sorry, I looked for docs before I found the makefile entry, and then didn't look again 😄

mathjazz (approved):
Thanks, Axel!

Bummer regarding the ranges, though.

PR #1865, opened 2021-02-10
Fix Bug 1679565 Title Missing In Terms Panel and Popup

anujpandey785:
1) Wrapped the LI element in a LOCALIZED element with a Localized id='terms-Translation--copy'
2) Added the title 'Copy Into Translation' for the tooltip
3) Made an entry for the corresponding title in the translate.ftl file (Line 462)

### DIFF
--- a/frontend/public/static/locale/en-US/translate.ftl
+++ b/frontend/public/static/locale/en-US/translate.ftl
@@ -459,11 +459,6 @@ machinery-MicrosoftTerminology--visit-microsoft = MICROSOFT
         Visit Microsoft Terminology Service API.
         © 2018 Microsoft Corporation. All rights reserved.
 
-## Machinery Microsoft Terminology
-## Shows the translation source from Microsoft Terminology.
-terms-MicrosoftTerminology--visit-microsoft = MICROSOFT
-    .title = Copy Into Translation
-
 ## Machinery Translation
 ## Shows a specific translation from machinery.
 machinery-Translation--copy =
@@ -633,6 +628,9 @@ search-TimeRangeFilter--save-range = SAVE RANGE
 
 ## Term
 ## Shows term entry with its metadata
+
+term-Term--copy = 
+    .title = Copy Into Translation
 
 term-Term--for-example = E.G.
 
--- a/frontend/public/static/locale/en-US/translate.ftl
+++ b/frontend/public/static/locale/en-US/translate.ftl
@@ -459,6 +459,10 @@ machinery-MicrosoftTerminology--visit-microsoft = MICROSOFT
         Visit Microsoft Terminology Service API.
         © 2018 Microsoft Corporation. All rights reserved.
 
+## Machinery Microsoft Terminology
+## Shows the translation source from Microsoft Terminology.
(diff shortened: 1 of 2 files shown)


anujpandey785:
Forgot to run prettier, Adding another commit in a moment :-)

frontend/src/core/term/components/Term.js
```diff
@@ -52,33 +52,36 @@ export default function Term(props: Props) {
...
-                    <p className='usage'>
-                        <Localized id='term-Term--for-example'>
-                            <span className='title'>E.G.</span>
-                        </Localized>
-                        <span className='content'>{term.usage}</span>
-                    </p>
-                )}
-            </div>
-        </li>
+        <Localized id='terms-Translation--copy' attrs={{ title: true }}>
```
mathjazz, line 55: String identifiers [should follow this form](https://github.com/mozilla/pontoon/tree/master/frontend#adding-localized-content): `${module}-${ComponentName}--${string-description}`, so in this case `term-Term--copy`.

frontend/public/static/locale/en-US/translate.ftl
```diff
@@ -459,6 +459,10 @@ machinery-MicrosoftTerminology--visit-microsoft = MICROSOFT
         Visit Microsoft Terminology Service API.
         © 2018 Microsoft Corporation. All rights reserved.
 
+## Machinery Microsoft Terminology
+## Shows the translation source from Microsoft Terminology.
+terms-MicrosoftTerminology--visit-microsoft = MICROSOFT
+    .title = Copy Into Translation
```
mathjazz, line 465: You can remove the comment, because it's pasted from elsewhere :) and move the string further down the file, just above `term-Term--for-example`.

You'll also need to update the id to match the one used in the component (`term-Term--copy`) and remove the message value (`MICROSOFT`), because it's not used.

frontend/src/core/term/components/Term.js
anujpandey785: Thanks for sharing the documentation link. Noted :-)

frontend/public/static/locale/en-US/translate.ftl
```diff
@@ -459,7 +459,6 @@ machinery-MicrosoftTerminology--visit-microsoft = MICROSOFT
         Visit Microsoft Terminology Service API.
         © 2018 Microsoft Corporation. All rights reserved.
 
-
```
mathjazz, line 462: Please revert this empty line you removed. :)

mathjazz (commented):
Please fix the nit and we're good to go. :)

frontend/public/static/locale/en-US/translate.ftl
anujpandey785: Oops!

mathjazz (approved):
Well done, thanks for the patch!

anujpandey785:
> Well done, thanks for the patch!
> 
> 

image

anujpandey785:
> Well done, thanks for the patch!
> 
> 

image

anujpandey785:
> Well done, thanks for the patch!
> 
> 

I'm glad I could be of help 😃☺️

PR #1866, opened 2021-02-11
Add ability to reveal adjacent strings 

### LINKED ISSUE #2066: Ability to view adjacent strings to a search result
## [Bug 1370173](https://bugzilla.mozilla.org/show_bug.cgi?id=1370173)
Bug Reporter: @StoyanDimitrov
CC:   @jordis, @julen, @mathjazz, @Mitch4sho, @MikkCZ
See also:     https://bugzilla.mozilla.org/show_bug.cgi?id=1370200

The idea is to have search result that shows few strings before and few stings after it.

Pootle has it. While not needed for every day use I find it very convenient at times and critical at others (basically no other way around it).

The idea is that related strings are close together in the source file. The classic example is the strings for pagination. You have "First", "Prev", "Next" and "Last" strings filed one after the other. I know that "Prev" and "Next" are together, but currently have no easy way to identify the correct "Prev"/"Next" pair I want to edit, let alone all four mentioned.

In a projects like AMO that lacks any kind of searchable meta-data attached to strings it is the only way to navigate to different sections of this 2 file, 2.8k strings project.

Mitch4sho:
Fixes #2066.

### DIFF
--- a/frontend/public/static/locale/en-US/translate.ftl
+++ b/frontend/public/static/locale/en-US/translate.ftl
@@ -268,6 +268,13 @@ entitydetails-GenericOriginalString--plural = PLURAL
 entitydetails-GenericOriginalString--singular = SINGULAR
 
 
+## Entities List Entity
+## Displays a single Entity as a list element
+
+entitieslist-Entity--sibling-strings-title =
+    .title = Click to reveal sibling strings
+
+
 ## Rich editor
 ## Renders a Rich editor for Fluent string editing
 
--- a/frontend/src/core/api/types.ts
+++ b/frontend/src/core/api/types.ts
@@ -74,6 +74,14 @@ export type Entity = {
 };
 
 /**
+ * Lists of preceding and succeeding entities
+ */
+export type EntitySiblings = {
+    readonly preceding: Array<Entity>,
+    readonly succeeding: Array<Entity>,
+};
+
+/**
  * List of Entity objects.
  */
 export type Entities = Array<Entity>;
@@ -111,6 +119,7 @@ export type SourceType =
     | 'systran-translate'
     | 'microsoft-terminology'
     | 'caighdean';
+
 export type MachineryTranslation = {
     sources: Array<SourceType>;
     itemCount?: number;
--- a/frontend/src/core/api/types.ts
+++ b/frontend/src/core/api/types.ts
@@ -71,14 +71,15 @@ export type Entity = {
     readonly source: Array<Array<string>> | Record<string, any>;
     readonly translation: Array<EntityTranslation>;
     readonly readonly: boolean;
+    readonly sibling: boolean;
 };
 
 /**
  * Lists of preceding and succeeding entities
  */
 export type EntitySiblings = {
-    readonly preceding: Array<Entity>,
-    readonly succeeding: Array<Entity>,
+    readonly preceding: Array<Entity>;
+    readonly succeeding: Array<Entity>;
 };
 
 /**
@@ -119,7 +120,6 @@ export type SourceType =
     | 'systran-translate'
     | 'microsoft-terminology'
     | 'caighdean';
-
 export type MachineryTranslation = {
     sources: Array<SourceType>;
     itemCount?: number;
--- a/frontend/src/core/api/types.ts
+++ b/frontend/src/core/api/types.ts
@@ -71,8 +71,7 @@ export type Entity = {
     readonly source: Array<Array<string>> | Record<string, any>;
     readonly translation: Array<EntityTranslation>;
     readonly readonly: boolean;
-    readonly sibling: boolean;
-    readonly displaySiblings: boolean;
+    readonly isSibling: boolean;
 };
 
 /**
--- a/frontend/src/core/api/types.ts
+++ b/frontend/src/core/api/types.ts
@@ -72,6 +72,7 @@ export type Entity = {
     readonly translation: Array<EntityTranslation>;
     readonly readonly: boolean;
     readonly sibling: boolean;
+    readonly displaySiblings: boolean;
 };
 
 /**
--- a/frontend/src/core/entities/reducer.ts
+++ b/frontend/src/core/entities/reducer.ts
@@ -55,19 +55,21 @@ function injectSiblingEntities(
     entity: number,
 ): Entities {
     const index = entities.findIndex((item) => item.pk === entity);
-    const currentPK = entities.map((e) => e.pk);
+    let parentEntity = entities[index];
+    const currentPKs = entities.map((e) => e.pk);
     let newEntitiesState = entities.slice();
+    parentEntity = { ...parentEntity, displaySiblings: true };
 
     // filtering out parent entities already present in the list
-    let siblingsTestOne = siblings.preceding.filter(
-        (sibling) => !currentPK.includes(sibling.pk),
+    const precedingSiblings = siblings.preceding.filter(
+        (sibling) => !currentPKs.includes(sibling.pk),
     );
 
-    let siblingsTestTwo = siblings.succeeding.filter(
-        (sibling) => !currentPK.includes(sibling.pk),
+    const succeedingSiblings = siblings.succeeding.filter(
+        (sibling) => !currentPKs.includes(sibling.pk),
     );
 
-    let list = [...siblingsTestOne, entities[index], ...siblingsTestTwo];
+    let list = [...precedingSiblings, parentEntity, ...succeedingSiblings];
 
     newEntitiesState.splice(index, 1, ...list);
 
--- a/frontend/src/core/entities/reducer.ts
+++ b/frontend/src/core/entities/reducer.ts
@@ -55,15 +55,28 @@ function injectSiblingEntities(
     entity: number,
 ): Entities {
     const index = entities.findIndex((item) => item.pk === entity);
+    const currentPK = entities.map((e) => e.pk);
+    let newEntitiesState = entities.slice();
 
-    const list = [
-        ...siblings.preceding,
-        entities[index],
-        ...siblings.succeeding,
-    ];
-    entities.splice(index, 1, ...list);
+    // filtering out parent entities already present in the list
+    let siblingsTestOne = siblings.preceding.filter(
+        (sibling) => !currentPK.includes(sibling.pk),
+    );
 
-    return entities;
+    let siblingsTestTwo = siblings.succeeding.filter(
+        (sibling) => !currentPK.includes(sibling.pk),
+    );
+
+    let list = [...siblingsTestOne, entities[index], ...siblingsTestTwo];
+
+    newEntitiesState.splice(index, 1, ...list);
+
+    // This block removes duplicated siblings
+    newEntitiesState = newEntitiesState.filter(
+        (currentValue, index, array) =>
+            array.findIndex((e) => e.pk === currentValue.pk) === index,
+    );
+    return newEntitiesState;
 }
 
 const initial: EntitiesState = {
--- a/frontend/src/core/entities/reducer.ts
+++ b/frontend/src/core/entities/reducer.ts
@@ -1,14 +1,20 @@
-import { RECEIVE, REQUEST, RESET, UPDATE } from './actions';
+import { RECEIVE, REQUEST, RESET, UPDATE, RECEIVE_SIBLINGS } from './actions';
 
-import type { Entities, EntityTranslation } from 'core/api';
+import type { Entities, EntityTranslation, EntitySiblings } from 'core/api';
 import type {
     ReceiveAction,
     RequestAction,
     ResetAction,
     UpdateAction,
+    ReceiveSiblingsAction,
 } from './actions';
 
-export type Action = ReceiveAction | RequestAction | ResetAction | UpdateAction;
+export type Action =
+    | ReceiveAction
+    | RequestAction
+    | ResetAction
+    | UpdateAction
+    | ReceiveSiblingsAction;
 
 // Read-only state.
 export type EntitiesState = {
@@ -41,6 +47,23 @@ function updateEntityTranslation(
             translation: translations,
         };
     });
+}
+
+function injectSiblingEntities(
+    entities: Entities,
+    siblings: EntitySiblings,
+    entity: number,
+): Entities {
+    const index = entities.findIndex((item) => item.pk === entity);
+
+    const list = [
+        ...siblings.preceding,
+        entities[index],
+        ...siblings.succeeding,
+    ];
+    entities.splice(index, 1, ...list);
+
+    return entities;
 }
 
 const initial: EntitiesState = {
@@ -84,6 +107,15 @@ export default function reducer(
                     action.entity,
                     action.pluralForm,
                     action.translation,
+                ),
+            };
+        case RECEIVE_SIBLINGS:
+            return {
+                ...state,
+                entities: injectSiblingEntities(
+                    state.entities,
+                    action.siblings,
+                    action.entity,
                 ),
             };
         default:
--- a/frontend/src/core/entities/reducer.ts
+++ b/frontend/src/core/entities/reducer.ts
@@ -58,7 +58,6 @@ function injectSiblingEntities(
     let parentEntity = entities[index];
     const currentPKs = entities.map((e) => e.pk);
     let newEntitiesState = entities.slice();
-    parentEntity = { ...parentEntity, displaySiblings: true };
 
     // filtering out parent entities already present in the list
     const precedingSiblings = siblings.preceding.filter(
--- a/frontend/src/modules/entitieslist/components/EntitiesList.tsx
+++ b/frontend/src/modules/entitieslist/components/EntitiesList.tsx
@@ -38,10 +38,6 @@ type InternalProps = Props & {
     store: Record<string, any>;
 };
 
-type State = {
-    entitiesWithActiveSiblings: Array<number>;
-};
-
 /**
  * Displays a list of entities and their current translation.
  *
@@ -49,15 +45,12 @@ type State = {
  * entities. It interacts with `core/navigation` when an entity is selected.
  *
  */
-export class EntitiesListBase extends React.Component<InternalProps, State> {
+export class EntitiesListBase extends React.Component<InternalProps> {
     list: { current: any };
 
     constructor(props: InternalProps) {
         super(props);
         this.list = React.createRef();
-        this.state = {
-            entitiesWithActiveSiblings: [],
-        };
     }
 
     componentDidMount() {
@@ -104,7 +97,6 @@ export class EntitiesListBase extends React.Component<InternalProps, State> {
             previous.author !== current.author ||
             previous.time !== current.time
         ) {
-            this.setState({ entitiesWithActiveSiblings: [] });
             this.props.dispatch(entities.actions.reset());
         }
 
@@ -232,12 +224,6 @@ export class EntitiesListBase extends React.Component<InternalProps, State> {
 
     getSiblingEntities: (entity: number) => void = (entity: number) => {
         const { dispatch, locale } = this.props;
-        this.setState({
-            entitiesWithActiveSiblings: [
-                ...this.state.entitiesWithActiveSiblings,
-                entity,
-            ],
-        });
         dispatch(entities.actions.getSiblingEntities(entity, locale.code));
     };
 
@@ -371,14 +357,10 @@ export class EntitiesListBase extends React.Component<InternalProps, State> {
                 >
                     {hasMore || props.entities.entities.length ? (
                         <ul>
-                            {props.entities.entities.map((entity, i) => {
+                            {props.entities.entities.map((entity) => {
                                 const selected =
                                     !props.batchactions.entities.length &&
                                     entity.pk === props.parameters.entity;
-
-                                const areSiblingsRevealed = this.state.entitiesWithActiveSiblings.includes(
-                                    entity.pk,
-                                );
 
                                 return (
                                     <Entity
@@ -396,12 +378,9 @@ export class EntitiesListBase extends React.Component<InternalProps, State> {
                                         locale={props.locale}
                                         selected={selected}
                                         selectEntity={this.selectEntity}
-                                        key={i}
+                                        key={entity.pk}
                                         getSiblingEntities={
                                             this.getSiblingEntities
-                                        }
-                                        areSiblingsRevealed={
-                                            areSiblingsRevealed
                                         }
                                         parameters={parameters}
                                     />
--- a/frontend/src/modules/entitieslist/components/EntitiesList.tsx
+++ b/frontend/src/modules/entitieslist/components/EntitiesList.tsx
@@ -38,6 +38,10 @@ type InternalProps = Props & {
     store: Record<string, any>;
 };
 
+type State = {
+    entitiesWithActiveSiblings: Array<number>;
+};
+
 /**
  * Displays a list of entities and their current translation.
  *
@@ -45,12 +49,15 @@ type InternalProps = Props & {
  * entities. It interacts with `core/navigation` when an entity is selected.
  *
  */
-export class EntitiesListBase extends React.Component<InternalProps> {
+export class EntitiesListBase extends React.Component<InternalProps, State> {
     list: { current: any };
 
     constructor(props: InternalProps) {
         super(props);
         this.list = React.createRef();
+        this.state = {
+            entitiesWithActiveSiblings: [],
+        };
     }
 
     componentDidMount() {
@@ -97,6 +104,7 @@ export class EntitiesListBase extends React.Component<InternalProps> {
             previous.author !== current.author ||
             previous.time !== current.time
         ) {
+            this.setState({ entitiesWithActiveSiblings: [] });
             this.props.dispatch(entities.actions.reset());
         }
 
@@ -224,6 +232,12 @@ export class EntitiesListBase extends React.Component<InternalProps> {
 
     getSiblingEntities: (entity: number) => void = (entity: number) => {
         const { dispatch, locale } = this.props;
+        this.setState({
+            entitiesWithActiveSiblings: [
+                ...this.state.entitiesWithActiveSiblings,
+                entity,
+            ],
+        });
         dispatch(entities.actions.getSiblingEntities(entity, locale.code));
     };
 
@@ -362,8 +376,9 @@ export class EntitiesListBase extends React.Component<InternalProps> {
                                     !props.batchactions.entities.length &&
                                     entity.pk === props.parameters.entity;
 
-                                const sibling = entity.sibling;
-                                const displaySiblings = entity.displaySiblings;
+                                const areSiblingsRevealed = this.state.entitiesWithActiveSiblings.includes(
+                                    entity.pk,
+                                );
 
                                 return (
                                     <Entity
@@ -385,8 +400,9 @@ export class EntitiesListBase extends React.Component<InternalProps> {
                                         getSiblingEntities={
                                             this.getSiblingEntities
                                         }
-                                        sibling={sibling}
-                                        displaySiblings={displaySiblings}
+                                        areSiblingsRevealed={
+                                            areSiblingsRevealed
+                                        }
                                         parameters={parameters}
                                     />
                                 );
--- a/frontend/src/modules/entitieslist/components/EntitiesList.tsx
+++ b/frontend/src/modules/entitieslist/components/EntitiesList.tsx
@@ -222,6 +222,16 @@ export class EntitiesListBase extends React.Component<InternalProps> {
         );
     };
 
+    getSiblingEntities = (entity: number) => {
+        const entityIds = this.props.entities.entities.map(
+            (entity) => entity.pk,
+        );
+        const { dispatch, locale } = this.props;
+        dispatch(
+            entities.actions.getSiblingEntities(entity, locale.code, entityIds),
+        );
+    };
+
     toggleForBatchEditing: (
         entity: number,
         shiftKeyPressed: boolean,
@@ -330,7 +340,7 @@ export class EntitiesListBase extends React.Component<InternalProps> {
 
     render(): React.ReactElement<'div'> {
         const props = this.props;
-        const search = props.parameters.search;
+        const parameters = props.parameters;
 
         // InfiniteScroll will display information about loading during the request
         const hasMore = props.entities.fetching || props.entities.hasMore;
@@ -371,10 +381,13 @@ export class EntitiesListBase extends React.Component<InternalProps> {
                                         }
                                         isTranslator={props.isTranslator}
                                         locale={props.locale}
-                                        search={search}
                                         selected={selected}
                                         selectEntity={this.selectEntity}
                                         key={i}
+                                        getSiblingEntities={
+                                            this.getSiblingEntities
+                                        }
+                                        parameters={parameters}
                                     />
                                 );
                             })}
--- a/frontend/src/modules/entitieslist/components/EntitiesList.tsx
+++ b/frontend/src/modules/entitieslist/components/EntitiesList.tsx
@@ -222,14 +222,9 @@ export class EntitiesListBase extends React.Component<InternalProps> {
         );
     };
 
-    getSiblingEntities = (entity: number) => {
-        const entityIds = this.props.entities.entities.map(
-            (entity) => entity.pk,
-        );
+    getSiblingEntities: (entity: number) => void = (entity: number) => {
         const { dispatch, locale } = this.props;
-        dispatch(
-            entities.actions.getSiblingEntities(entity, locale.code, entityIds),
-        );
+        dispatch(entities.actions.getSiblingEntities(entity, locale.code));
     };
 
     toggleForBatchEditing: (
@@ -367,6 +362,8 @@ export class EntitiesListBase extends React.Component<InternalProps> {
                                     !props.batchactions.entities.length &&
                                     entity.pk === props.parameters.entity;
 
+                                const sibling = entity.sibling;
+
                                 return (
                                     <Entity
                                         checkedForBatchEditing={props.batchactions.entities.includes(
@@ -387,6 +384,7 @@ export class EntitiesListBase extends React.Component<InternalProps> {
                                         getSiblingEntities={
                                             this.getSiblingEntities
                                         }
+                                        sibling={sibling}
                                         parameters={parameters}
                                     />
                                 );
--- a/frontend/src/modules/entitieslist/components/EntitiesList.tsx
+++ b/frontend/src/modules/entitieslist/components/EntitiesList.tsx
@@ -363,6 +363,7 @@ export class EntitiesListBase extends React.Component<InternalProps> {
                                     entity.pk === props.parameters.entity;
 
                                 const sibling = entity.sibling;
+                                const displaySiblings = entity.displaySiblings;
 
                                 return (
                                     <Entity
@@ -385,6 +386,7 @@ export class EntitiesListBase extends React.Component<InternalProps> {
                                             this.getSiblingEntities
                                         }
                                         sibling={sibling}
+                                        displaySiblings={displaySiblings}
                                         parameters={parameters}
                                     />
                                 );
--- a/frontend/src/modules/entitieslist/components/Entity.css
+++ b/frontend/src/modules/entitieslist/components/Entity.css
@@ -13,6 +13,7 @@
 
 .entity.checked,
 .entity.selected,
+.entity.sibling:hover,
 .entity:hover {
     background-color: #333941;
 }
--- a/frontend/src/modules/entitieslist/components/Entity.css
+++ b/frontend/src/modules/entitieslist/components/Entity.css
@@ -103,7 +103,11 @@
     color: #ffa10f;
 }
 
-.entity .sibling-entities {
+.entity.sibling {
+    background-color: #4d5967;
+}
+
+.entity .sibling-entities-icon {
     color: #aaa;
     left: 19px;
     top: 39px;
@@ -113,6 +117,6 @@
     font-size: 16px;
 }
 
-.entity .sibling-entities:hover {
+.entity .sibling-entities-icon:hover {
     color: #7bc876;
 }
--- a/frontend/src/modules/entitieslist/components/Entity.css
+++ b/frontend/src/modules/entitieslist/components/Entity.css
@@ -109,12 +109,12 @@
 
 .entity .sibling-entities-icon {
     color: #aaa;
-    left: 19px;
+    left: 17px;
     top: 39px;
     position: absolute;
-    margin: -13px -13px -13px -16px;
-    padding: 13px 13px 13px 16px;
-    font-size: 16px;
+    margin: 0 -13px -13px -17px;
+    padding: 0 13px 13px 17px;
+    font-size: 14px;
 }
 
 .entity .sibling-entities-icon:hover {
--- a/frontend/src/modules/entitieslist/components/Entity.css
+++ b/frontend/src/modules/entitieslist/components/Entity.css
@@ -102,3 +102,17 @@
 .entity.warnings .status:before {
     color: #ffa10f;
 }
+
+.entity .sibling-entities {
+    color: #aaa;
+    left: 19px;
+    top: 39px;
+    position: absolute;
+    margin: -13px -13px -13px -16px;
+    padding: 13px 13px 13px 16px;
+    font-size: 16px;
+}
+
+.entity .sibling-entities:hover {
+    color: #7bc876;
+}
--- a/frontend/src/modules/entitieslist/components/Entity.tsx
+++ b/frontend/src/modules/entitieslist/components/Entity.tsx
@@ -19,9 +19,8 @@ type Props = {
     selected: boolean;
     selectEntity: (...args: Array<any>) => any;
     getSiblingEntities: Function;
-    sibling: boolean;
-    displaySiblings: boolean;
     parameters: NavigationParams;
+    areSiblingsRevealed: Boolean;
 };
 
 /**
@@ -138,9 +137,8 @@ export default class Entity extends React.Component<Props> {
             isTranslator,
             locale,
             selected,
-            sibling,
-            displaySiblings,
             parameters,
+            areSiblingsRevealed,
         } = this.props;
 
         const classSelected = selected ? 'selected' : '';
@@ -150,7 +148,7 @@ export default class Entity extends React.Component<Props> {
 
         const classChecked = checkedForBatchEditing ? 'checked' : '';
 
-        const classSibling = sibling ? 'sibling' : '';
+        const classSibling = entity.isSibling ? 'sibling' : '';
         return (
             <li
                 className={`entity ${this.status} ${classSelected} ${classBatchEditable} ${classChecked} ${classSibling}`}
@@ -164,7 +162,7 @@ export default class Entity extends React.Component<Props> {
                     <div>
                         {parameters.search || this.areFiltersApplied() ? (
                             <Localized id='entitieslist-Entity--sibling-strings-title'>
-                                {displaySiblings ? null : (
+                                {!areSiblingsRevealed && (
                                     <i
                                         className={
                                             'sibling-entities-icon fas fa-expand-arrows-alt'
--- a/frontend/src/modules/entitieslist/components/Entity.tsx
+++ b/frontend/src/modules/entitieslist/components/Entity.tsx
@@ -23,7 +23,7 @@ type Props = {
 };
 
 type State = {
-    siblingsAreActive: Boolean;
+    areSiblingsActive: Boolean;
 };
 
 /**
@@ -47,7 +47,7 @@ type State = {
 export default class Entity extends React.Component<Props, State> {
     constructor(props) {
         super(props);
-        this.state = { siblingsAreActive: false };
+        this.state = { areSiblingsActive: false };
     }
     get status(): string {
         const translations = this.props.entity.translation;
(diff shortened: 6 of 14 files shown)


Mitch4sho:
Thank you Pike!

frontend/src/modules/entitieslist/components/Entity.js
```diff
@@ -151,6 +152,12 @@ export default class Entity extends React.Component<Props> {
                         />
                     </p>
                 </div>
+                <button
+                    className='sibling-Entities'
```
mathjazz, line 156: Nit: Please use lowercase class names.

frontend/src/modules/entitieslist/components/Entity.css
```diff
@@ -102,3 +102,13 @@
 .entity.warnings .status:before {
     color: #ffa10f;
 }
+
+/* button styling for entities sibling */
+
+.entity .noButton {
+    display: none;
+}
```
mathjazz, line 110: I guess this is temporary?

We shouldn't need this rule, because we don't use the `.noButton` class. We'll show/hide the button using React.

Mitch4sho: Yes, this temporary, show would you use react to show or hide an element?

Mitch4sho: Ignore the last statement I figured it out :)

frontend/src/modules/entitieslist/components/Entity.js
```diff
@@ -152,12 +152,16 @@ export default class Entity extends React.Component<Props> {
                         />
                     </p>
                 </div>
-                <button
-                    className='sibling-Entities'
-                    onClick={this.props.getSiblingEntities}
-                >
-                    TEST
-                </button>
+                {search ? (
```
mathjazz, line 155: You'll need to extend the check to also include filters.

To do that, you'll need to pass the `parameters` prop from `EntityList` to `Entity` and then check if either of the filter parameters (`status`, `extra`, `tag`, `time`, `author`) is set. You could add a helper function, so that the expression when rendering the button will be cleaner, e.g. `search || areFiltersApplied()`.

frontend/src/modules/entitieslist/components/EntitiesList.js
```diff
@@ -217,6 +217,12 @@ export class EntitiesListBase extends React.Component<InternalProps> {
         );
     };
 
+    getSiblingEntities = () => {
+        const { entity, locale } = this.props.parameters;
```
mathjazz, line 221: We should retrieve data from parameters - the `entity` parameter reflects the entity that is currently open in the editor, but users could click on others.

You can get `locale` directly from props.

And you'll need to pass entity from the `Entity` component where you click on it. See `selectEntity` in Entity component for inspiration.

mathjazz (commented):
Getting there!

frontend/src/modules/entitieslist/components/Entity.js
```diff
@@ -20,6 +20,8 @@ type Props = {
     selected: boolean,
     selectEntity: Function,
     getSiblingEntities: Function,
+    areFiltersApplied: Function,
```
mathjazz, line 23: You can remove this line. `areFiltersApplied` is not passed as a prop from another component. It's a function **within** the component.

frontend/src/modules/entitieslist/components/Entity.js
```diff
@@ -20,6 +20,8 @@ type Props = {
...
+    parameters: Object,
```
mathjazz, line 24: We should be as specific as possible when declaring type and avoid generic types like `Object`. Here, you should use `NavigationParams` type, which needs to be imported from `core/navigation`.

frontend/src/modules/entitieslist/components/EntitiesList.js
```diff
@@ -329,6 +329,7 @@ export class EntitiesListBase extends React.Component<InternalProps> {
     render() {
         const props = this.props;
         const search = props.parameters.search;
```
mathjazz, line 330: Now that you pass all parameters as props, it doesn't make much sense to pass the search parameter separately. So you can remove this line and the prop from the `Entity` component, where you'll need to declare search from parameters.

frontend/src/modules/entitieslist/components/Entity.js
Mitch4sho: I see that what was causing the error for Flow `Cannot create Entity element because property areFiltersApplied is missing in props [1] but exists in Props [2].`

frontend/src/modules/entitieslist/components/Entity.js
Mitch4sho: I understand because the `NavigationParams` that will be imported will set the types to be what is exported out of the file `selector.js`

Mitch4sho: and with a generic type like `Object`, it's not setting any rules for us to follow

frontend/src/modules/entitieslist/components/Entity.js
```diff
@@ -95,6 +97,14 @@ export default class Entity extends React.Component<Props> {
         this.props.selectEntity(this.props.entity);
     };
 
+    getSiblingEntities = (e: SyntheticMouseEvent<HTMLLIElement>) => {
+        if (e.currentTarget && e.currentTarget.classList.contains('status')) {
+            return null;
+        }
```
mathjazz, line 103: We don't need this block here: it's used in `selectEntity()` to (poorly, see point 2 below) prevent entity selection if user clicks on the `status` icon (which toggles batch editing). So you can simply remove it.

However, there are two things we should do:

1. As you'll notice, when you click on the TEST button on the entity that's not currently selected for editing, it gets selected. We want to prevent that. There's a [`stopPropagation`](https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation) on the event object that helps with that.
2. Unrelated to this patch (but since we're already at it, let's do it now) you should also call `e.stopPropagation()` early in the `toggleForBatchEditing()` function. Then you can remove the `if (e.currentTarget)` block from `selectEntity`, including the weird flow comment. That's not only more proper and elegant solution to the problem - it also works better. Right now, if you're logged out (i.e. unabled to do batch editing) and click on the status icon, the entity also doesn't get selected, but it should.

frontend/src/modules/entitieslist/components/Entity.js
```diff
@@ -95,6 +97,14 @@ export default class Entity extends React.Component<Props> {
...
+
+        this.props.getSiblingEntities(this.props.entity);
```
mathjazz, line 105: You should actually pass `entity.pk`, not the entire `entity` object.

frontend/src/modules/entitieslist/components/Entity.js
```diff
@@ -103,15 +113,30 @@ export default class Entity extends React.Component<Props> {
...
+            parameters.status != null ||
+            parameters.extra != null ||
+            parameters.tag != null ||
+            parameters.time != null ||
+            parameters.author != null
+        ) {
+            return true;
+        } else {
+            return false;
+        }
```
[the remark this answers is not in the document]
Mitch4sho: I knew I was doing something wrong here will fix it

frontend/src/modules/entitieslist/components/Entity.js
```diff
@@ -83,35 +85,47 @@ export default class Entity extends React.Component<Props> {
...
     };
 
+    getSiblingEntities = () => {
+        this.props.getSiblingEntities(this.props.entity.pk);
+    };
+
     toggleForBatchEditing = (e: SyntheticMouseEvent<HTMLSpanElement>) => {
         const { entity, isReadOnlyEditor, isTranslator } = this.props;
 
+        e.stopPropagation();
```
mathjazz, line 99: Let's move this within the `if` clause: that way we'll select the entity when clicking on the status icon in read-only projects and for non-privileged (incl. non-authenticated) users.

frontend/src/modules/entitieslist/components/Entity.js
```diff
@@ -83,35 +85,47 @@ export default class Entity extends React.Component<Props> {
...
-        // $FLOW_IGNORE
-        if (e.target && e.target.classList.contains('status')) {
-            return null;
-        }
-
+    selectEntity = () => {
         this.props.selectEntity(this.props.entity);
     };
 
+    getSiblingEntities = () => {
```
mathjazz, line 92: We'll need to `stopPropagation` first. See Point 1 in https://github.com/mozilla/pontoon/pull/1866#discussion_r579644071.

pontoon/base/views.py
```diff
@@ -377,6 +377,28 @@ def get_translations_from_other_locales(request):
     return JsonResponse(payload, safe=False)
 
 
+# Get entities preceding and succeeding the current entity
```
mathjazz, line 380: Please move the comment further down (see other functions).

pontoon/base/views.py
```diff
@@ -377,6 +377,28 @@ def get_translations_from_other_locales(request):
...
+@utils.require_AJAX
+def get_sibling_entities(request):
+    print("TEST-------------------")
+    try:
+        entity = int(request.GET["entity"])
+        locale = request.GET["locale"]
+        print(entity, locale)
+
```
mathjazz, line 388: Nit: You can remove the blank line here for consistency. :)

mathjazz (commented):
Onwards!

frontend/src/modules/entitieslist/components/Entity.js
Mitch4sho: I have it set up like this passing the event object and then used `stopPropagation() `is this correct?

```    
getSiblingEntities = (e: SyntheticMouseEvent<HTMLSpanElement>) => {
        e.stopPropagation();
        this.props.getSiblingEntities(this.props.entity.pk);
    };
```

mathjazz: Yeah, except it's not a span element the event is triggered on, it's a button. So you'll need to change the type (to `HTMLButtonElement` I think).

mathjazz (commented):
Nice work!

Please remove the `does not call the selectEntity function` test, because we removed that functionality.

You'll need to define `const selectEntityFn = sinon.spy();` and pass it as a prop to `Entity` in both `does not call the toggleForBatchEditing...` test cases.

pontoon/base/views.py
```diff
@@ -377,6 +377,27 @@ def get_translations_from_other_locales(request):
     return JsonResponse(payload, safe=False)
 
 
+@utils.require_AJAX
+def get_sibling_entities(request):
+    # Get entities preceding and succeeding the current entity
```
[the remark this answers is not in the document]
Mitch4sho: got it!

pontoon/base/views.py
```diff
@@ -377,6 +377,27 @@ def get_translations_from_other_locales(request):
...
+        return JsonResponse(
+            {"status": False, "message": "Bad Request: {error}".format(error=e)},
+            status=400,
+        )
+
+    entity = get_object_or_404(Entity, pk=entity)
+    locale = get_object_or_404(Locale, code=locale)
+
+    payload = []
+    return JsonResponse(payload, safe=False)
```
mathjazz, line 398: Great job so far! Now comes the fun part, where we'll actually return the data that we need. :)

Instead of an empty list, we should return a list of preceding entities and a list of succeeding entities. You should use a [`dictionary`](https://docs.python.org/3/tutorial/datastructures.html#dictionaries) with two keys ("preceding" and "succeeding") to combine the lists into one data structure. You'll see we use that pattern a lot in other `JsonResponse` instances.

To generate the actual content for the two lists, you should use `Entity.map_entities()` for each, which returns all the data required to render entities in the string list for the given locale and entities, which is exactly what we need here. See how it's used in `_get_entities_list()` for example.

`Entity.map_entities()` takes 3 arguments:
- `locale`: we already have it
- `preferred_source_locale`: see the `entities()` view to see how to define it
- `entities`: we'll need an `Entity` [QuerySet](https://docs.djangoproject.com/en/3.1/ref/models/querysets/) that gets the relevant data from the DB. See below.

1. You'll need to [filter](https://docs.djangoproject.com/en/3.1/ref/models/querysets/#filter) entities for which `resource` (i.e. file) equals the resource of the entity the user has clicked on, and for which the `obsolete` flag is set to `False` (i.e. the entity hasn't been removed).
2. Then you'll need to [`order_by`](https://docs.djangoproject.com/en/3.1/ref/models/querysets/#django.db.models.query.QuerySet.order_by) the `order` field and again filter entities for which `order` value is less (for preceding) or more (for succeeding) than the order of the given entity.
3. Finally, you'll only need to retrieve the first two matching instances, which you can do by simply adding `[:2]` at the end of the QuerySet.

Note that filter() and order_by() expressions can be chained, so you can do `Entity.objects.filter(...).order_by(...).filter(...)...`.

pontoon/base/views.py
```diff
@@ -407,20 +405,19 @@ def get_sibling_entities(request):
...
     )[:2]
 
-
-    print(entities)
-    print(preceding_entities)
-    print(succeeding_entities)
-
     entity = get_object_or_404(Entity, pk=entity)
     locale = get_object_or_404(Locale, code=locale)
     preferred_source_locale = ""
```
Mitch4sho, line 410: I wasn't sure how you wanted me to handle the `preferred_source_locale`. I didn't know if you wanted me to copy exactly how the `entities()` view defines it. Let me know if I did it correctly :)

pontoon/base/views.py
```diff
@@ -380,21 +379,20 @@ def get_translations_from_other_locales(request):
...
         entity = int(request.GET["entity"])
         locale = request.GET["locale"]
-        print(entity, locale)
     except (MultiValueDictKeyError, ValueError) as e:
         return JsonResponse(
             {"status": False, "message": "Bad Request: {error}".format(error=e)},
             status=400,
         )
 
     entities = Entity.objects.get(id=entity)
```
Mitch4sho, line 394: can you show me the magic to make this more DRY. I can't figure out how to use the QuerySets properly

pontoon/base/views.py
Mitch4sho, line 410: I wasn't sure how you wanted me to handle the preferred_source_locale. I didn't know if you wanted me to copy exactly how the entities() view defines it. Let me know if I did it correctly :)

mathjazz: Correct - we need the same thing as in the `entities()` view.

pontoon/base/views.py
```diff
@@ -377,6 +377,53 @@ def get_translations_from_other_locales(request):
...
+            {"status": False, "message": "Bad Request: {error}".format(error=e)},
+            status=400,
+        )
+
+    entities = Entity.objects.get(id=entity)
+
+    preceding_entities = (
+        Entity.objects.filter(resource=entities.resource, obsolete=False)
+        .order_by("order")
+        .filter(order__gt=entities.order)
```
mathjazz, line 399: Preceding entities have an `order`that is **less than (`lt`)** the order of the current entity, so `order__lt=entity.order`. And inversely below for the `succeeding_entities`.

pontoon/base/views.py
mathjazz: 1. Remove this line. :)
2. As you can see below, `preceding_entities` and `succeeding_entities` QuerySets share the first two lines, so you can factor them out into an `entities` QuerySet and reuse it, e.g. `preceding = entities.filter(...)`.
3. Also, move both `get_object_or_404` declarations after the `try - except` block, because you'll need `entity` in `entity.resource` and `entity.order`.

pontoon/base/views.py
```diff
@@ -377,6 +377,53 @@ def get_translations_from_other_locales(request):
...
+    entities_dict = {
+        "preceding": Entity.map_entities(
+            locale, preferred_source_locale, preceding_entities
+        ),
+        "succeeding": Entity.map_entities(
+            locale, preferred_source_locale, succeeding_entities
+        ),
+    }
+
+    payload.append(entities_dict)
```
mathjazz, line 423: You can directly return `entities_dict`, no need to put it into a list.

mathjazz (commented):
Getting there - almost done with the view!

pontoon/base/views.py
Mitch4sho: This makes more sense after the question you answered on Element.
I didn't realize that `entity` and `locale` were getting the model object when using the `get_object_or_404` method
but hence the name of the method.  so it came with all the models plus the one that is automatically added the `id` which I was looking for so I can filter out the entities in the first place :)

pontoon/base/views.py
Mitch4sho: I was tired :)

pontoon/base/views.py
```diff
@@ -377,6 +377,44 @@ def get_translations_from_other_locales(request):
...
+        locale = request.GET["locale"]
+    except (MultiValueDictKeyError, ValueError) as e:
+        return JsonResponse(
+            {"status": False, "message": "Bad Request: {error}".format(error=e)},
+            status=400,
+        )
+
+    entity = get_object_or_404(Entity, pk=entity)
+    locale = get_object_or_404(Locale, code=locale)
+    preferred_source_locale = ""
```
mathjazz, line 394: Nit: I'd add a blank line above this line to visually separate the blocks.

Mitch4sho: Thank you, I'll keep it in mind when I'm working. Where did you learn your best practice methods when writing code.

mathjazz: I wouldn't call this a best practice, just a personal preference. :)

Ideally, we should just leave it for the linters to do their job and not talk about code style during reviews. So I'd be totally fine if you ignore my comment. :)

frontend/src/modules/entitieslist/components/Entity.js
```diff
@@ -130,12 +144,19 @@ export default class Entity extends React.Component<Props> {
                     className='status fa'
                     onClick={this.toggleForBatchEditing}
                 />
+                {parameters.search || this.areFiltersApplied() ? (
+                    <i
+                        className={'sibling-entities fas fa-arrows-alt-v'}
+                        title='Get more entities'
```
mathjazz, line 150: Please replace text with the one used in the mockup: "Click to reveal surrounding strings".

And make it [localizable](https://github.com/mozilla/pontoon/tree/master/frontend#localization), i.e. wrap in the Localized element and create an entry for it in the .ftl file.

frontend/src/modules/entitieslist/components/Entity.js
```diff
@@ -130,12 +144,19 @@ export default class Entity extends React.Component<Props> {
                     className='status fa'
                     onClick={this.toggleForBatchEditing}
                 />
+                {parameters.search || this.areFiltersApplied() ? (
```
mathjazz, line 147: I'm not sure if we should make the icon only visible on hover or only for selected entities. Let's try with the selected entities first and see how it'll work (I'm not sure if there's a use case for seeing surrounding strings of strings that are not being edited).

frontend/src/modules/entitieslist/components/Entity.css
```diff
@@ -102,3 +102,17 @@
 .entity.warnings .status:before {
     color: #ffa10f;
 }
+
+.entity .sibling-entities {
+    color: #838383;
```
mathjazz, line 107: We use #aaa as the neutral color, let's stick to that.

mathjazz (commented):
This is getting shape! Well done!

frontend/public/static/locale/en-US/translate.ftl
```diff
@@ -258,6 +258,12 @@ entitydetails-ContextIssueButton--context-issue-button = REQUEST CONTEXT or REPO
 entitydetails-GenericOriginalString--plural = PLURAL
 entitydetails-GenericOriginalString--singular = SINGULAR
 
+## Entities List
+## Show a list of entities
```
mathjazz, line 262: This section refers to a component (as opposed to a module), so we should tweak the comments a little bit:

```
## Entities List Entity
## Displays a single Entity as a list element
```

Nit: Please also make sure there are 2 blank lines between this and the previous section (instead of just 1).

frontend/public/static/locale/en-US/translate.ftl
```diff
@@ -258,6 +258,12 @@ entitydetails-ContextIssueButton--context-issue-button = REQUEST CONTEXT or REPO
...
+
+entitieslist-Entity--title =
+    .title = Click to reveal surrounding strings
```
mathjazz, line 265: Let's use the word sibling (instead of surrounding) here for consistency. Sorry for the confusion in previous review!

frontend/public/static/locale/en-US/translate.ftl
```diff
@@ -258,6 +258,12 @@ entitydetails-ContextIssueButton--context-issue-button = REQUEST CONTEXT or REPO
 entitydetails-GenericOriginalString--plural = PLURAL
 entitydetails-GenericOriginalString--singular = SINGULAR
 
+## Entities List
+## Show a list of entities
+
+entitieslist-Entity--title =
```
mathjazz, line 264: Let's make this more specific `entitieslist-Entity--sibling-strings-title`.

frontend/src/core/entities/actions.js
```diff
@@ -102,6 +117,18 @@ export function get(
     };
 }
 
+export function getSiblingEntities(entity: number, locale: string): Function {
+    return async (dispatch) => {
+        dispatch(request());
```
mathjazz, line 122: You can remove this line - it's only needed when loading entities for the entire entity list.

frontend/src/core/entities/actions.js
```diff
@@ -37,6 +39,19 @@ export function receive(entities: Entities, hasMore: boolean): ReceiveAction {
...
     };
 }
+/**
+ * Update the siblings of an entity.
+ */
+export type ReceiveSiblingsAction = {
+    type: typeof RECEIVE_SIBLINGS,
+    entities: Entities,
+};
+export function receiveSiblings(entities: Entities): ReceiveSiblingsAction {
```
mathjazz, line 49: `Entities` type here and in line 47 needs to be replaced with a new flow type that will match the backend response. You can define it in `core/api/types`.

Something like this:
```
export type EntitySiblings = {|
    +preceding: Array<Entity>;,
    +succeeding: Array<Entity>;,
|};
```

frontend/src/core/entities/reducer.js
```diff
@@ -88,6 +96,12 @@ export default function reducer(
                     action.translation,
                 ),
             };
+        case RECEIVE_SIBLINGS:
+            return {
+                ...state,
+                preceding: state.entities.concat(action.entities.preceding),
+                succeeding: state.entities.concat(action.entities.succeeding),
```
mathjazz, line 103: We shouldn't extend the `EntitiesState` (not even above in the initial state). Instead, we should extend the `entities` key with the new entities.

You could create a new function that will inject preceding entities before and succeeding entities after the given entity within the entities list. You could call it like this:
```js
entities: injectSiblingEntities(state.entities, action.entities, entity),
```

mathjazz (commented):
Nice work with the localization and the actions! Just a couple of minor improvements needed there.

In the reducer, we'll need to extend the **existing** entities. See details below. 💪

frontend/src/core/entities/reducer.js
Mitch4sho: I'm not too sure what or how I would use `entity` in this function

mathjazz: You'd need to pass it here first from the `getSiblingEntities` action.

It's the selected entity - the one the user clicks on - and you'll need it to determine where to insert surrounding entities.

frontend/src/core/entities/reducer.js
```diff
@@ -51,10 +55,36 @@ function updateEntityTranslation(
...
 }
 
+function injectSiblingEntities(
+    entities: Entities,
+    siblings: EntitySiblings,
+    entity: number,
+) {
+    const copy = [...entities];
+
+    const index = entities.findIndex((item) => item.pk === entity);
```
mathjazz, line 65: Now that you have index, how about you try this approach:

1. Create a [concatenated list](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat) of preceding siblings, entity item (i.e. `entities[index]`) and succeeding siblings.
2. Replace entity item in the entities array with a concatenated list - `entities.splice(index, 1, concatenated);`.
3. Return a [flat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) list with `entities.flat()`.

I also don't think you'll need to create a `copy`.

mathjazz (commented):
Getting there! Please also run `make pyupgrade`.

frontend/src/core/entities/reducer.js
Mitch4sho: You are a wizard

Mitch4sho:
`Flow` did not like that we `splice()` the entities array, I made a copy and  `splice()` the copy.

mathjazz (commented):
It works! Well done!

I'm thinking about enhancements to the UX. I don't like:
- ~That the selected entity is moved after sibling entities are loaded. It should stay in the same place.~
- That there's no clear indication of sibling strings after they load. I'd try applying the dark background (same as selected), which would reset after another entity is selected or after a period of time.

And we'll also need to prevent sibling entities to load if they are already loaded (e.g. if you have entities 1, 2, 3, 5, 6, 7 in the list and you click reveal sibling entities for entity 5, only entity 4 should load).

frontend/public/static/locale/en-US/translate.ftl
```diff
@@ -259,6 +259,20 @@ entitydetails-GenericOriginalString--plural = PLURAL
...
+
+entitieslist-Entity--sibling-strings-title =
+    .title = Click to reveal sibling strings
+
+
+## Entities List Entity
+## Displays a single Entity as a list element
+
+entitieslist-Entity--sibling-strings-title =
+    .title = Click to reveal sibling strings
```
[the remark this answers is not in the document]
Mitch4sho: Got it resolved

frontend/src/modules/entitieslist/components/EntitiesList.js
```diff
@@ -218,8 +218,13 @@ export class EntitiesListBase extends React.Component<InternalProps> {
     };
 
     getSiblingEntities = (entity: number) => {
+        const entityIds = this.props.entities.entities.map(
+            (entity) => entity.pk,
+        );
```
mathjazz, line 223: That could be a lot of entities, which means a lot of data, and a big loop in the action to check if the list of entities contains any slings.

It should be enough (and scale much better) if you just pass the immediate preceding and succeeding entity.

Mitch4sho: Can I have more context on how to implement this, I thought we needed to use the previous list to filter out the `siblingsEntities`?

mathjazz: I think you could do something even simpler, at least as the first step: you could just remove duplicates at the end of `injectSiblingEntities`.

frontend/src/core/entities/reducer.js
```diff
@@ -63,8 +63,26 @@ function injectSiblingEntities(
...
+    siblings.succeeding.forEach((sibling) => {
+        const i = entities.findIndex((item) => item.pk === sibling.pk);
+        entities.forEach((e) => {
+            if (entities.indexOf(e) === i) {
+                list.splice(list.indexOf(sibling), 1);
+            }
+        });
+    });
+
+    entities.splice(index, 1, ...list);
```
mathjazz, line 85: That should work, but we should optimize by reducing the number of loops and the lines of code. You could keep the original code, and just remove duplicates before returning.

See https://stackoverflow.com/a/56757215 for inspiration.

```js
entities = entities.filter((v, i, a) => a.findIndex(t => (t.id === v.id)) === i)
```

Mitch4sho: Thanks! I didn't know you can add that many parameters in `filter()` learning something new every day 👍🏾

mathjazz:
> That the selected entity is moved after sibling entities are loaded. It should stay in the same place.

Let's ignore this, because it can't work if the scrollbar is not visible (because the list is too short).

Pike:
Bouncing this to rerun tests.

docker-compose.yml
```diff
@@ -5,6 +5,7 @@ version: "2"
 services:
   # Webapp
   webapp:
+    platform: linux/amd64
```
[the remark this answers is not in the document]
mathjazz: This should no longer be needed.

frontend/src/modules/entitieslist/components/EntitiesList.tsx
```diff
@@ -219,10 +249,15 @@ export class EntitiesListBase extends React.Component<InternalProps> {
...
     getSiblingEntities: (entity: number) => void = (entity: number) => {
         const { dispatch, locale } = this.props;
-
         dispatch(entities.actions.getSiblingEntities(entity, locale.code));
     };
 
+    // need to be able to remove siblings when an new parent is selected
+    removeSiblingEntities: () => void = () => {
+        console.log('*** we are removing siblings');
+        this.props.dispatch(entities.actions.reset());
```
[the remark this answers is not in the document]
mathjazz: I don't think we want this. I couldn't test it, because all entities get reloaded when you switch entities, but I assume it would be unexpected and confusing.

frontend/src/modules/entitieslist/components/Entity.css
```diff
@@ -103,8 +103,8 @@
     color: #ffa10f;
 }
 
-.entity .sibling-entities {
-    color: #333941;
+.entity.sibling {
+    background-color: #313940;
```
mathjazz, line 107: Seems like we don't use this color. Why now stick to `#333941` instead?

frontend/src/core/entities/reducer.ts
```diff
@@ -43,6 +49,25 @@ function updateEntityTranslation(
...
+    siblings: EntitySiblings,
+    entity: number,
+): Entities {
+    const index = entities.findIndex((item) => item.pk === entity);
+    let newEntitiesState = entities.slice();
+    let list = [...siblings.preceding, entities[index], ...siblings.succeeding];
+
+    newEntitiesState.splice(index, 1, ...list);
+
+    newEntitiesState = newEntitiesState.filter(
```
mathjazz, line 63: A comment above explaining that this block eliminates dupes would be nice.

frontend/src/modules/entitieslist/components/Entity.tsx
```diff
@@ -89,13 +89,22 @@ export default class Entity extends React.Component<Props> {
...
+            this.areFiltersApplied();
+
         if (
             e.target instanceof HTMLElement &&
             e.target.classList.contains('status')
         ) {
             return null;
         }
-        if (!this.props.selected && !this.props.sibling) {
+        if (parameterOne || parameterTwo) {
```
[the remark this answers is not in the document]
mathjazz: There's a bug in this iteration. I didn't check if it's caused by this, but the symptom is: when the Missing filter is applied, after selecting any entity all entities in the sidebar are reloaded.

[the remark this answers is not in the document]
mathjazz: Also, it's better to put all conditions in the if than use generic variable names like `parameterOne`. :)

frontend/src/modules/entitieslist/components/EntitiesList.tsx
```diff
@@ -116,6 +123,29 @@ export class EntitiesListBase extends React.Component<InternalProps> {
...
+                    if (!PK.includes(e.pk)) {
+                        siblingPKs.push(e.pk);
+                    }
+                });
+
+                this.setState({
+                    siblingIds: siblingPKs,
+                });
+            }
+        }
```
mathjazz, line 148: A more elegant solution would be to set a new `sibling` flag in the `map_entities` function in Djangoon the when siblings are returned.

mathjazz (commented):
Nice work implementing indication of siblings and preventing siblings to load if they are already loaded. I found some bugs and left some comments.

docker-compose.yml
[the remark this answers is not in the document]
Mitch4sho: with the next push,  code will be removed; we might want to update the docs to reflect the change => https://mozilla-pontoon.readthedocs.io/en/latest/dev/setup.html

frontend/src/core/api/types.ts
```diff
@@ -110,6 +118,7 @@ export type SourceType =
     | 'systran-translate'
     | 'microsoft-terminology'
     | 'caighdean';
+
```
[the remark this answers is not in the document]
Mitch4sho: Do we want to remove the blank line or leave it in?

frontend/src/modules/entitieslist/components/EntitiesList.tsx
Mitch4sho: To clarify this solution, is it to let Django handle all the filtering with the new `sibling` flag?

frontend/src/modules/entitieslist/components/Entity.css
Mitch4sho: I have no idea why I didn't use the previous color, changed it 👍🏾

frontend/src/modules/entitieslist/components/Entity.tsx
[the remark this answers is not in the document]
Mitch4sho: Will work on this

frontend/src/modules/entitieslist/components/EntitiesList.tsx
Mitch4sho: April and I just solved this it was way more elegant 😄

frontend/src/modules/entitieslist/components/Entity.css
Mitch4sho: image 
I think I got the color from the mock because it is a little bit lighter than the default color let me know if this matters.

docker-compose.yml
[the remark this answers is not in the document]
mathjazz: Good point! I'll PR that now.

frontend/src/core/api/types.ts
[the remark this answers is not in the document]
mathjazz: Remove (sorry for phrasing that awkwardly). :)

frontend/src/modules/entitieslist/components/Entity.css
mathjazz: Hmmmm... Right. Since the mockup was created, we changed the background color of the selected entity from lighter to darker than default.

I'll take the patch for a spin once it's done to see if there are further CSS changes needed.

Mitch4sho:
@mathjazz, can you test out the code and see if I need anything else before moving on to the next steps.

mathjazz:
The patch works as expected. You've addressed all the issues!

I've deployed it to stage to collect more feedback about the UX:
https://mozilla-pontoon-staging.herokuapp.com/sl/firefox/all-resources/?status=missing&string=228369

mathjazz:
Two more changes were proposed on chat:
- usa background color that is different from selected strings: `#4d5967`
- use `fa-expand-arrows-alt` as the icon

Mitch4sho:
hopefully, this is the last step before polishing :)

mathjazz:
Just a slight polish :)

```
.entity .sibling-entities-icon {
	color: #aaa;
	left: 17px;
	top: 39px;
	position: absolute;
	margin: 0 -13px -13px -17px;
	padding: 0 13px 13px 17px;
	font-size: 14px;
}
```

Mitch4sho:
@mathjazz I know this test failed because of the linting for python, but when I ran the test locally, it changed over 20 files styling. Is this because something in the test has changed?

mathjazz:
Yeah, we upgraded black which reformatted most of the python files. Please rebase against master to resolve the conflict.

Mitch4sho:
Let me know if this is exactly what you @mathjazz

frontend/src/core/entities/reducer.ts
```diff
@@ -55,19 +55,21 @@ function injectSiblingEntities(
...
     );
 
-    let siblingsTestTwo = siblings.succeeding.filter(
-        (sibling) => !currentPK.includes(sibling.pk),
+    const succeedingSiblings = siblings.succeeding.filter(
+        (sibling) => !currentPKs.includes(sibling.pk),
     );
 
-    let list = [...siblingsTestOne, entities[index], ...siblingsTestTwo];
+    let list = [...precedingSiblings, parentEntity, ...succeedingSiblings];
```
mathjazz, line 72: Much better names! :)

frontend/src/modules/entitieslist/components/Entity.css
```diff
@@ -109,12 +109,12 @@
...
-    left: 19px;
+    left: 17px;
     top: 39px;
     position: absolute;
-    margin: -13px -13px -13px -16px;
-    padding: 13px 13px 13px 16px;
-    font-size: 16px;
+    margin: 0 -13px -13px -17px;
+    padding: 0 13px 13px 17px;
+    font-size: 14px;
```
mathjazz, line 117: 👍

frontend/src/modules/entitieslist/components/EntitiesList.tsx
```diff
@@ -357,6 +362,9 @@ export class EntitiesListBase extends React.Component<InternalProps> {
                                     !props.batchactions.entities.length &&
                                     entity.pk === props.parameters.entity;
 
+                                const sibling = entity.sibling;
+                                const displaySiblings = entity.displaySiblings;
```
mathjazz, line 366: Since you already pass the whole `entity` object as the prop to `Entity`, there's no need to also pass `displaySiblings` and `sibling` as separate props.

pontoon/base/models.py
```diff
@@ -2911,7 +2911,13 @@ def for_project_locale(
 
     @classmethod
     def map_entities(
-        cls, locale, preferred_source_locale, entities, visible_entities=None
+        cls,
+        locale,
+        preferred_source_locale,
+        entities,
+        visible_entities=None,
+        sibling=False,
```
mathjazz, line 2919: Nit: We usually prefix boolean variables and function names with "is" or "has", so `is_sibling` would be better: https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig

pontoon/base/models.py
```diff
@@ -2911,7 +2911,13 @@ def for_project_locale(
...
+        display_siblings=False,
```
mathjazz, line 2920: The backend will always pass `display_siblings` as `False` to frontend, so to save on web traffic it should instead not pass anything and the frontend would just assume the default value of `False`.

Even better, I believe the logic to show/hide the reveal button would be cleaner if you put something like `areSiblingsRevealed` in a state, similarly to how we handle [`isDiffVisible`](https://github.com/mozilla/pontoon/blob/0f2382606f9a892c670133ae2d4f295e5603c3a4/frontend/src/modules/history/components/Translation.tsx#L39).

frontend/src/modules/entitieslist/components/Entity.css
```diff
@@ -102,3 +102,21 @@
 .entity.warnings .status:before {
     color: #ffa10f;
 }
+
+.entity.sibling {
+    background-color: #4d5967;
```
mathjazz, line 107: Please make sure this background color is overridden on hover with the hover background color.

frontend/src/modules/entitieslist/components/EntitiesList.tsx
```diff
@@ -38,19 +38,26 @@ type InternalProps = Props & {
...
  */
-export class EntitiesListBase extends React.Component<InternalProps> {
+export class EntitiesListBase extends React.Component<InternalProps, State> {
     list: { current: any };
 
     constructor(props: InternalProps) {
         super(props);
         this.list = React.createRef();
+        this.state = {
+            entitiesWithActiveSiblings: [],
```
mathjazz, line 59: I wonder why did you decide to keep the state of all entities with revealed siblings in the parent component, instead of in `Entity`?

It should be simpler to keep the logic in the `Entity` component, without the need to passing it downwards from the parent `EntitiesList` component.

mathjazz (commented):
Nice work!

Unless we have a good reason, we should keep the "revealed" state in the Entity component. And change background of siblings on hover.

Other than that, this is good to go.

frontend/src/modules/entitieslist/components/EntitiesList.tsx
Mitch4sho: I couldn't explain it properly in text so I made a video 
https://www.loom.com/share/e6f681e83f2748a9ad9e733b31bc5ed8

mathjazz: That's a great explanation, nice work! Do you might pushing or pasting a patch with the code used in the video?

mathjazz: My first bet would be that we use index as the `key`, which is problematic when order of items in the list changes (and in our cases it does, because we inject):
https://robinpokorny.medium.com/index-as-a-key-is-an-anti-pattern-e0349aece318

Can you try using `entity.pk` instead of `i` as the `key` attribute of the `Entity` component?

abowler2: @Mitch4sho  After thinking on this some more I realized that when you and I worked on this I got too focused on 'finding a working solution' and I should have taken a moment to reflect more on WHY the state was changing based on the index location and thus changing the state for the newly injected sibling instead of the correct entity.

When taking that time, I saw that within the `map()` the `key` is being set to the index number. Since the `key` is what [React](https://reactjs.org/docs/lists-and-keys.html#keys) uses to identify which element to apply the state change to it would then apply the change to the newly injected sibling at that index/key location.

By setting the `key` to the `entity.pk` you should be able to successfully handle the state in `Entity` in the way you demonstrate in your video.

As a side note, this beautifully demonstrates why it is recommended to use a unique identifier other than index whenever possible 😁

Mitch4sho: This makes so much sense now thinking about it! Thank you should be able to complete the refactor tomorrow!

Mitch4sho:
alright, let me know if anything else I need to look out for @mathjazz :)

frontend/src/modules/entitieslist/components/Entity.tsx
```diff
@@ -14,9 +16,14 @@ type Props = {
...
     locale: Locale;
-    search: string | null | undefined;
     selected: boolean;
     selectEntity: (...args: Array<any>) => any;
+    getSiblingEntities: Function;
+    parameters: NavigationParams;
+};
+
+type State = {
+    siblingsAreActive: Boolean;
```
mathjazz, line 26: Nit: `areSiblingsActive` would be consistent with our boolean variable naming convention which starts with a  verb (is, are, has ...).

frontend/src/modules/entitieslist/components/Entity.tsx
```diff
@@ -122,21 +155,39 @@ export default class Entity extends React.Component<Props> {
...
             >
                 <span
                     className='status fa'
                     onClick={this.toggleForBatchEditing}
                 />
+                {classSelected && !classSibling ? (
+                    <div>
+                        {parameters.search || this.areFiltersApplied() ? (
+                            <Localized id='entitieslist-Entity--sibling-strings-title'>
+                                {!this.state.siblingsAreActive && (
```
mathjazz, line 172: Is there are reason to have this condition within the `<Localized>` element instead of putting it in line 170?

Logically all three of them serve the same purpose - showing or hiding the same button. So they should be all together, and if it make code easier to read, you could also factor them out into a function (e.g. `showSiblingEntitiesButton`).

mathjazz (approved):
Great work, Mitch!

Everything works as expected! I left two nits where I believe we could improve code readability slightly.

frontend/src/modules/entitieslist/components/Entity.tsx
Mitch4sho: there was no reason for it, and you are correct; it does do the same thing to show the button
here is the refactor in a function 
```
showSiblingEntitiesButton = () => {
        const isSearched = this.props.parameters.search;
        const areFiltersApplied = this.areFiltersApplied();
        const areSiblingsActive = !this.state.areSiblingsActive;

        return (isSearched || areFiltersApplied) && areSiblingsActive;
    };
    
    {this.showSiblingEntitiesButton() && (
                            <Localized id='entitieslist-Entity--sibling-strings-title'>
                                <i
                                    className={
                                        'sibling-entities-icon fas fa-expand-arrows-alt'
                                    }
                                    title='Click to reveal sibling strings'
                                    onClick={this.getSiblingEntities}
                                ></i>
                            </Localized>
                        )} 
```

Mitch4sho: Let me know if the way I have the function is readable.

mathjazz: LGTM.

Mitch4sho: I think I'm getting really old, it took me way to long to decipher what this meant 😭

mathjazz: Hehe :) Ok, let's make these two changes and we ship.

mathjazz (approved):
It. Is. DONE!

What a ride! Congrats for pushing till the end and keeping up with a tedious review process, spec changes along the way and... yours truly.

This feature is big and so will be the joy of people using it.

Mitch4sho:
I loved every bit of the process! 😭 time for the next issue to handle

PR #1868, opened 2021-02-11
Fix bug 1692423: Replace broken CSS compressor with a no-op compressor

mathjazz:
The + and - operators in CSS calc() must be surrounded by whitespace:
https://developer.mozilla.org/en-US/docs/Web/CSS/calc()#notes

The yuglify CSS compressor we use removes them and breaks CSS.
I tried yuicompressor instead, which has the same bug: https://github.com/yui/yuicompressor/issues/59

Other CSS compressors I've checked look (even more) outdated. I haven't tested any.

The only other instance of calc() in our code uses - instead of +, which is fine.

### DIFF
--- a/pontoon/settings/base.py
+++ b/pontoon/settings/base.py
@@ -476,6 +476,7 @@ PIPELINE = {
     "STYLESHEETS": PIPELINE_CSS,
     "JAVASCRIPT": PIPELINE_JS,
     "JS_COMPRESSOR": "pipeline.compressors.terser.TerserCompressor",
+    "CSS_COMPRESSOR": "pipeline.compressors.terser.NoopCompressor",
     "YUGLIFY_BINARY": path(
         os.environ.get("YUGLIFY_BINARY", "node_modules/.bin/yuglify")
     ),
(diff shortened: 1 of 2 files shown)


Pike:
I mentioned this on matrix, I don't think we should constrain our CSS by bugs in that pipeline. In particular for CSS, where there's just not much to win, IMHO.

https://django-pipeline.readthedocs.io/en/latest/compressors.html#no-op-compressors suggests that we can just use a no-op css compressor to cat the files together.

If we're really concerned, maybe there's a trick to compress the cat'ed css files with a transfer encoding.

mathjazz:
> https://django-pipeline.readthedocs.io/en/latest/compressors.html#no-op-compressors suggests that we can just use a no-op css compressor to cat the files together.

Agreed. I've tested this on stage and it works as expected.

pontoon/settings/base.py
```diff
@@ -476,6 +476,7 @@ def path(*args):
     "STYLESHEETS": PIPELINE_CSS,
     "JAVASCRIPT": PIPELINE_JS,
     "JS_COMPRESSOR": "pipeline.compressors.terser.TerserCompressor",
+    "CSS_COMPRESSOR": "pipeline.compressors.terser.NoopCompressor",
```
Pike, line 479: That one doesn't exist, it's in the parent, https://github.com/jazzband/django-pipeline/blob/master/pipeline/compressors/__init__.py#L256

Weird that that does something constructive on stage

Pike (changes requested):

Pike (approved):

julen (approved):
I fully agree with Axel. Changes look good to me 👍.

PR #1870, opened 2021-02-15
Bug 1584178: Wrap the placeables when you send a string to Google Translate API.

jotes:
TODO:
* [x] add unit tests
* [x] add docs
* [x] remove the debug prints
* [x] add a link to the comparison matrix

I'm opening this PR because it's easier to test the feature on Staging. Nonetheless, it still needs some work.


mathjazz:
Deployed to stage:

mathjazz:
I've done some testing ...

1. Python format string: the bug is fixed!
- PROD: https://pontoon.mozilla.org/sl/firefox/all-resources/?string=199692
- STAGE: https://mozilla-pontoon-staging.herokuapp.com/sl/firefox/all-resources/?string=199692

2. Python format string in quotes from [bug 1630407](https://bugzilla.mozilla.org/show_bug.cgi?id=1630407): the bug is fixed, but `"` becomes `&quot;` and space is added around the placeable if not already surrounded by it.

- PROD: https://pontoon.mozilla.org/fi/lightning/calendar/calendar/calendar-uri-redirect-dialog.ftl/?string=209313
- STAGE: https://mozilla-pontoon-staging.herokuapp.com/fi/lightning/calendar/calendar/calendar-uri-redirect-dialog.ftl/?string=209313

3. HTML tags: the bug is fixed, but space is added around the placeable if not already surrounded by it.

- PROD: https://pontoon.mozilla.org/ka/firefox/devtools/client/tooltips.ftl/?string=204020
- STAGE: https://mozilla-pontoon-staging.herokuapp.com/ka/firefox/devtools/client/tooltips.ftl/?string=204020

4. HTML tags: content within tags is missing (mind the bug in the last closing `<strong>` on stage).
- PROD: https://pontoon.mozilla.org/el/firefox/devtools/client/tooltips.ftl/?string=204020
- STAGE: https://mozilla-pontoon-staging.herokuapp.com/el/firefox/devtools/client/tooltips.ftl/?string=204020

mathjazz:
I've deployed the latest update to stage. I can still reproduce #2.

jotes:
I was struggling with this bug for quite some time because depending on the selected locale/format/value of the text input, Google Translate API produces different results, and in many cases returns broken placeables.
So, I created a small comparison matrix that tries to find a combination of the parameters that doesn't produce translations with broken placeables.
You can find the script that produces the matrix and the generated results in [this Gist](https://gist.github.com/jotes/da50cf04b6c309a6a330e418ccaf8060).
I selected 3 locales:
* pl
* de
* th

Additionally, I added a few pre-processing methods of the tex t that goes to the API:
* plain - no changes, placeables are sent as they are
* wrapped_with_span - wrap placeable with HTML Span tag to mark them as non-translateable
  * `%s` -> `<span translate="no">$s</span>`
* wrapped_with_span_and_replaced_with_integer - replace placeables with HTML span tags with random integers inside, this method assumes that you do some processing on the frontend side.
  * `%s` -> `<span translate="no">1</span>`
* a few additional variations of the "wrapped_with_span" method but they don't seem to add much to the research

The summary table that compares all of the methods:


The results say that no matter on the case, GTA always breaks placeables in when the input format is `text`. Changing it to `HTML` seems to help in some cases, but requires thinking about the additional processing of the HTML entities.
It seems that the best solution is to go with the additional pre-processing of input text,  I mean `wrapped_with_span_and_replaced_with_integer` or any of its variants.

jotes:
I've added a couple additional pre-processing methods of the input text:
* plain_replaced_with_integers -> e.g. replaces `%s`  to `1234947`
* plain_replaced_with_integers -> e.g. replaces `%s`  to `123e4567-e89b-12d3-a456-426614174000`

Google Translate API interprets them correctly and doesn't mess up them for the PL locale. I'll run them the rest of the locales.

frontend/src/modules/machinery/components/source/GoogleTranslation.tsx
```diff
@@ -1,5 +1,189 @@
...
+    );
+
+    let newText: string = '';
+    let placeableOccurrence = placeablesRegex.exec(text);
+    let pos: number = 0;
+
+    while (placeableOccurrence) {
+        let placeable: string = placeableOccurrence[0];
+
+        // Wrap the placeable with spaces, in order to avoid concatenating hashes of placeables with the words that
```
jotes, line 96: the placeables

frontend/src/modules/machinery/components/source/GoogleTranslation.tsx
```diff
@@ -1,5 +1,189 @@
...
+
+    const checkAndReplace = (acc, placeable): string => {
+        let placeableHash = getPlaceableHash(placeable);
+
+        if (response.translation.indexOf(placeableHash) == -1) {
+            throw new Error(
+                `Google Translate API removed the placeable: ${placeable}`,
+            );
+        }
+        return acc.replace(placeableHash, placeable);
```
eemeli, line 183: If the same placeable shows up more than once in a message, will it also show up multiple times in `placeables`? This `replace()` call will only replace the first instance.

eemeli:
Would it make sense to encode the added spaces in the hash? For instance, if we start with a Fluent message
```
{$foo}text {$bar}
```
Assuming that the SHA1 hashes are `FOO` and `BAR`, correspondingly, that could be represented as
```
FOOA text BARN
```
Where the suffix `A` is indicating that a space was added after the placeable, and `N` that no space was added. This would allow for the added spaces to be trimmed with more certainty.

Not sure if this is something that ought to be considered separately from this PR, though.

frontend/src/modules/machinery/components/source/GoogleTranslation.tsx
jotes: That's something I've missed, and I'll add a test case for that too. Thanks!

jotes:
@eemeli I was thinking about something similar but in a different way, e.g. change the GoogleTranslateInput function to return an array of objects like:
```
 { "lspace": true, "rspace": false, "index": 1, "text": "placeable hash" }
```
And later use that in order to determine how to post-process the response from GTA.

However, I 've decided to not jump right into it because it could make this PR hard to understand, without a good explanation why the PR does that (I'm aware that what I'm trying to create is a hack that tries to fix GTA's bug). If what you see here is readable and understable enough I move further.

eemeli:
The array of objects would make a lot of sense, actually. That would also allow you to avoid the `sha.js` dependency, and use a random identifier for the "hash", as you'd be able to look it up from the table rather than needing to recalculate it later. The hash generation could then check that it's unique and be re-run if necessary.

mathjazz:
I pushed latest changes to [stage](https://mozilla-pontoon-staging.herokuapp.com/) and ran it against test strings [from above](https://github.com/mozilla/pontoon/pull/1870#issuecomment-779764225).

Seems like most of the issues are fixed, but there's a couple of new ones we should look into:

1. Python format string: the bug is fixed, but there is an extra space after the 2nd placeable (before comma).

- PROD: https://pontoon.mozilla.org/sl/firefox/all-resources/?string=199692
- STAGE: https://mozilla-pontoon-staging.herokuapp.com/sl/firefox/all-resources/?string=199692

2. Python format string in quotes from [bug 1630407](https://bugzilla.mozilla.org/show_bug.cgi?id=1630407): the bug is fixed, but it seems to impact the translation (`URI -tunnuksen` becomes `URI: n`).

- PROD: https://pontoon.mozilla.org/fi/lightning/calendar/calendar/calendar-uri-redirect-dialog.ftl/?string=209313
- STAGE: https://mozilla-pontoon-staging.herokuapp.com/fi/lightning/calendar/calendar/calendar-uri-redirect-dialog.ftl/?string=209313

3. HTML tags: the bug is fixed!

- PROD: https://pontoon.mozilla.org/ka/firefox/devtools/client/tooltips.ftl/?string=204020
- STAGE: https://mozilla-pontoon-staging.herokuapp.com/ka/firefox/devtools/client/tooltips.ftl/?string=204020

4. HTML tags (same as above, but different target locale): the bug is fixed!

- PROD: https://pontoon.mozilla.org/el/firefox/devtools/client/tooltips.ftl/?string=204020
- STAGE: https://mozilla-pontoon-staging.herokuapp.com/el/firefox/devtools/client/tooltips.ftl/?string=204020

eemeli:
@jotes Regarding the representation of placeables when passed to Google Translate, I somewhat randomly remembered just now about `<span translate="no">`, which allows stuff to [not be translated](https://cloud.google.com/translate/troubleshooting), at least for HTML input.

Would that make it more certain to pass placeables through the API without modification, or does it possibly change how content around them is handled? Or should this be a different/separate discussion?

jotes:
@mathjazz
Thanks! So, to clarify, there are two bugs:
* [ ] handling the whitespaces that surround a placeable
  * relatively easy to fix
* [ ] the current solution affects the quality of translation when the string is in `text` format
  * I'm going to check if `<span translate=no>` is going to help with that

@eemeli Actually, using `<span translate=no>` had been one of the possible options and was investigated here: https://github.com/mozilla/pontoon/pull/1870#issuecomment-874600686. I've used it in one of the previous versions of this PR.

That kind of marking placeables works only in `HTML` strings and requires passing format param in API calls to GTA (here: https://github.com/mozilla/pontoon/pull/1870/files#diff-59e1d12f6420b5134545b45d68695382bcddd0484e4c459bf4ecc2ab9734aebfR35).

I was experimenting with using `HTML` as the default input format in API calls to GTA, and it had helped in some cases but introduced bugs in others:
e.g.
https://gist.github.com/jotes/ed4b904ecfc8df37ab44ee6c3f0203e6 <- Google Translate API removes the newline character

Additionally, there's a case in which a string contains the reserved characters (like `<`, `>` e.g. `1 < 2 < 3`) and isn't HTML. In such a case, the backend has to post-process the response from Google Translate API and try to figure out if the string is HTML and then unquote the reserved characters.

That's why I've decided to detect the format of a string and use it in the API call to GTA, to avoid introducing too much complexity related to the post-processing functions.

------
I need your help with figuring out what are the next steps for this patch. I don't want to create tension on anybody to merge it, but here there are my thoughts/doubts/questions based on my experience:
* My solution is a hack that may get obsolete with a new release of Google Translate API
* The patch solves the main problem (broken placeables) - what's the severity of the other bugs?
* I don't want to create a patch that's going to be rejected/stalled because it's too huge and complicated to review (for everyone's sake).
  * Do you have an idea how to split it into mergeable parts or ideas about how to the patch easier to read?
  * Do I explain the problems/solutions/choices clearly in the comments?

mathjazz:
That's a great update @jotes, thanks!

"Handling the whitespaces that surround a placeable" issue should be addressed, because it's a regression. Especially if it's "relatively easy to fix".

And then we should wrap up (pun not intended) the work here. As you are hinting at, we're not going to get to a perfect solution, but at least we can fix the most critical translation errors caused by the the Google Translate API. And the patch has already gotten big (but not too big to be reviewed in one piece).

So I suggest you fix the regression, them we test again, and if all goes well, review the code and land.

mathjazz:
@jotes Seems like that extra space after the 2nd placeable from the [1st example](https://github.com/mozilla/pontoon/pull/1870#issuecomment-891839729) is still there.

Also, please rebase against master to fix the linting issue.

jotes:
@mathjazz hey, I've fixed the issue with the extra space. From now on, when a peaceable is wrapped with punctuation characters (look at `isPunctuationCharacter` function for more info) it will "stick" to them.
I also did some cleanups related to the unit tests and moved the code around to a more suitable module (outside of the components layer).

Can you check it again on staging? Thanks in advance.

mathjazz:
Deployed to stage. Now there's a missing space between a period and the placeable in the [1st example](https://github.com/mozilla/pontoon/pull/1870#issuecomment-891839729).

mathjazz:
Let's reopen when the work continues here.

PR #1872, opened 2021-02-16
Fix Bug 1692539: Implementing dark background on hover for all panels

anujpandey785:
Changed Background to : #272a2f

### DIFF
--- a/frontend/src/core/term/components/Term.css
+++ b/frontend/src/core/term/components/Term.css
@@ -9,7 +9,7 @@
 }
 
 .terms-list .term:not(.cannot-copy):hover {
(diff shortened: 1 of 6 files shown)


mathjazz:
Thanks for the patch. Could you please try `#333941` instead?

And also remove the extra padding in the Locales tab?

mathjazz (approved):
I think this is great. Nice work!

anujpandey785:
> I think this is great. Nice work!
> 
> 

Thank-you, Means a lot coming from you.
But you should really thank my Mentor and other awesome people I meet on Thursdays.  XD!!

PR #1874, opened 2021-02-23
Fix bug 1694504: Find & Replace fails for some strings

mathjazz:


Pike (approved):
Nice to see added test coverage here.

One thing I noticed is that the resulting strings aren't necessarily valid fluent. I.e., one could create leading dots or `[` via find-n-replace.

Not a regression, so that's OK. Just wondering.

mathjazz:
Yeah. At least we run checks.

PR #1876, opened 2021-02-25
Bug 1694911 - Add specs for user infobar component

flodolo:
CC @guerojeff


flodolo:
This is mostly a braindump of how I think this component should work, plenty of space for discussion and changes.

Other ideas I played with, but didn't include:
* Type: for example, having an _alert_ vs _info_, associated to different colors and iconography.
* Create notifications that would be displayed only a maximum number of times, or at specific intervals.

There are probably some implementation details I'm missing, for example how we'd manage cohorts/experiments.

Also didn't include mock-ups, because I'm not sure they help. But just to the idea.

infobar_mockup

modal_mockup

mathjazz (commented):
Thanks!
I left some questions and proposals.

I also think it makes sense to look a little closer into existing solutions for legal consents and AB testing / experiments to help determine the scope.

bsam1212 (approved):

flodolo:
@bsam1212 Playing with other people's production repositories is not OK.

specs/0109-user-notifications-component.md
```diff
@@ -0,0 +1,55 @@
...
+* Dismissed (boolean).
+
+If there are multiple notifications active, we look at them in order of creation date, from the most recent to the oldest, to make sure we only display one.
+
+We create a row in the table for the user the first time the notification is displayed, and keep updating the table until it’s dismissed.
+
+For `infobar`, the notification won’t be displayed to the user again (i.e. the `dismissed` field is set to True) if:
+* They click the close (x) button.
+* They click any of the links included.
+* They click the button, if present.
```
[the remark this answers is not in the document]
flodolo: Uhm, I think that would be annoying for users. They should have a choice to stop getting the notification.

But this makes me realize that the infobar for the add-on would require logic (is the add-on installed?).
So, we could use the logic to set up the cohort field, and display the message accordingly?

flodolo (commented):
Thanks for the feedback, waiting for a few more answers before making any update.

In general, I think it makes sense to focus on infobars and drop the modal dialogs.

specs/0109-user-notifications-component.md
```diff
@@ -0,0 +1,55 @@
...
+* We want to promote the Pontoon add-on. The notification will include a button opening the AMO page to install the add-on.
+* We make changes to Pontoon terms and conditions, and want to require users to acknowledge them. The infobar will include a link to the updated terms, and a button to explicitly accept them.
+
+# Feature explanation
+
+When a logged in user opens Pontoon, we check against a list of notifications. If the user matches the necessary conditions, we display the notification.
+
+Information about each notification is stored in a table with the following fields:
+* Notification ID (automatically generated).
+* Identifier: short text identifier, e.g. `new_terms_2020`.
```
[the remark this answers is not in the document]
mathjazz: OK, I see the value. Maybe we could pre-fill it from the title as a slug.

specs/0109-user-notifications-component.md
```diff
@@ -0,0 +1,55 @@
...
+* Active (boolean): indicates if the notification is currently active or not.
+* Creation date (date): stores when the notification was created.
+* Cohort: if defined, we only display the notification for users part of a specific experiment or cohort (A/B testing).
+* Type: `infobar`, `modal`. Infobar is displayed at the top of the window, while modal is displayed in the middle of the window.
+* Title: (optional) title for the notification.
+* Text: (mandatory) text for the notification. Basic HTML should be allowed.
+* Button: text for the button. If empty, the button will not be displayed. The only action associated to a button would be opening the associated URL.
```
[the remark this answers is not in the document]
mathjazz: I mean semantically (not stylistically) - this should be called Anchor or LinkText, because it will be used in a combination with the Link / URL.

specs/0109-user-notifications-component.md
```diff
@@ -0,0 +1,55 @@
...
+# Feature explanation
+
+When a logged in user opens Pontoon, we check against a list of notifications. If the user matches the necessary conditions, we display the notification.
+
+Information about each notification is stored in a table with the following fields:
+* Notification ID (automatically generated).
+* Identifier: short text identifier, e.g. `new_terms_2020`.
+* Active (boolean): indicates if the notification is currently active or not.
+* Creation date (date): stores when the notification was created.
+* Cohort: if defined, we only display the notification for users part of a specific experiment or cohort (A/B testing).
```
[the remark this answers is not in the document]
flodolo: Correct.

flodolo (commented):
I'm leaving the specs as is for today, in case Jeff wants to add some feedback, and then I'm going to update them tomorrow based on the feedback.

specs/0109-user-notifications-component.md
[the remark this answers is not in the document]
mathjazz: I don't think this can be implemented without logic efficiently. If you'd use a cohort, you'd need to update it regularly (with some logic) for every user to check if they installed the add-on or not.

specs/0109-user-notifications-component.md
```diff
@@ -0,0 +1,55 @@
...
+* Type: `infobar`, `modal`. Infobar is displayed at the top of the window, while modal is displayed in the middle of the window.
+* Title: (optional) title for the notification.
+* Text: (mandatory) text for the notification. Basic HTML should be allowed.
+* Button: text for the button. If empty, the button will not be displayed. The only action associated to a button would be opening the associated URL.
+* URL: URL to open when the user clicks the button.
+
+We also need to store information on the last time we displayed the infobar to each user, to determine if it should be displayed again in the future:
```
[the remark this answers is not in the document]
mathjazz: Then it's easier to track `dismissed_date` (same as we'd like to track `read_date` for notifications.

specs/0109-user-notifications-component.md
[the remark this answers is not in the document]
flodolo: This is how I imagined it working, but I guess it might be incorrect or not realistic.

Normal flow:
- User opens Pontoon.
- After the page is fully loaded, we send a request to an API that checks if there are active infobars that should be displayed for the user, and we display them as needed. 
- It should be possible to hook additional logic before this call, to support A/B testing and more complex conditions.

For the add-on promotion, the logic checking for the add-on would sit before the standard infobar API.

flodolo (commented):
I've made changes to address some of the comments.

Not changing the filename, because we would lose comments, and I'm not sure we settled on "infobars".

Not addressed:
- Count times an infobar is displayed, and how expensive it is.
- Can we use cohorts for the add-on?
- Button/anchor conversation.

specs/0109-user-notifications-component.md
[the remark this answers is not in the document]
flodolo: Note: the title is optional, so generating from that wouldn't work. Since we're removing the modal option, should we remove the title too? Would the text be enough?

mathjazz (commented):
After thinking about this for a while, I think add-on promotion (A) and new feature introduction (B) are quite different use cases:

- A cannot be executed by admins only, becuase it requires a custom code that checks if the add-on is installed.
- For A, keeping state of users in a cohort is harder than in the User object, and we'd need to update it every time a new user is created (or deleted).
- B should ideally be displayed with a popup near the feature UI. A URL path and XPath could be used for that.

Similarly to the Terms update notice, I think Add-on promotion is a specific use case that should be treated separately.

flodolo:
Questions on my side:
- Dropping the second use case (add-on) would solve the conversation around this spec?
- Do we agree on "infobar"?
- What about the "Number of times it’s been displayed." discussion?

flodolo:
Questions on my side:
- Would dropping the second use case (add-on) from the specs allow to move forward with this?
- Do we agree on calling these "infobars"?
- What about the "Number of times it’s been displayed." data point? Is it solvable, do we need to drop it?

flodolo:
> For changes to an existing infobar, do we track changes?

No, I don't think so. We want to start with a very basic model, so a change would be a new infobar in my head.

> Or we could now call them "Feature announcements" now.

I still feel like this can be used for more than just announcing a feature. For example, it could be easily used to warn about a planned downtime, so I'd prefer to keep a more generic approach.

Let me know if I've addressed all your notes. I agree on having start/end times as part of the infobar (removed the active field), I assume we'd still need different tables for tracking dismissal and all the visits.

specs/0109-infobar-component.md
```diff
@@ -0,0 +1,58 @@
...
+
+Create a reusable component to display infobars to users.
+
+# Motivation
+
+We need a flexible way to notify users without writing ad-hoc code for every situation:
+* Infobars should be manageable by admins without intervention from developers.
+* Unlike notifications, infobars are more visible and require interaction from users to get dismissed.
+
+Example use case: we introduce a brand new feature, like string comments, and we want to make sure that users are aware of it. The infobar will include a short message and a link to documentation.
```
mathjazz, line 15: I'd add the planned downtime announcement as another example.

specs/0109-infobar-component.md
```diff
@@ -0,0 +1,58 @@
...
+
+When a logged in user opens Pontoon, we check against a list of infobars. If the user matches the necessary conditions, we display the infobar.
+
+Information about each infobar is stored in a table with the following fields:
+* Identifier: short text identifier, e.g. `new_terms_2020`.
+* Start time (datetime): when we should start displaying the infobar to users.
+* End time (datetime): when we should stop displaying the infobar to users.
+* Creation date (date): stores when the infobar was created.
+* Cohort: if defined, we only display the infobar for users part of a specific experiment or cohort (A/B testing).
+* Type: `infobar`, `modal`. Infobar is displayed at the top of the window, while modal is displayed in the middle of the window.
```
mathjazz, line 27: Seems like an infobar can be an infobar or modal. Can we come up with better names? Maybe "announcement" could be of type infobar or modal.

Also, what's the use case for modal now that we dropped terms update?

mathjazz (commented):
Thanks for the update. Left a couple of comments.

Should we include the ability to position the announcement in a custom place, using something like a URL pattern and XPath?

specs/0109-infobar-component.md
flodolo: Uh, that's a very good point. The only use case I had in mind for "modal" was the terms update. Should we just get rid of this? It would definitely simplifies things.

flodolo:
> Should we include the ability to position the announcement in a custom place, using something like a URL pattern and XPath?

Initially I thought of adding a position field (e.g. bottom, top), but then I wasn't sure how helpful it would be. Do you have any example of "URL pattern/XPath"? Are you thinking of providing a DOM Path, so that the infobar can be placed anywhere in the page? How much would that complicate things in terms of styling the infobar?

specs/0109-infobar-component.md
mathjazz: If we don't have a use case for it, we should.

mathjazz:
> Initially I thought of adding a position field (e.g. bottom, top), but then I wasn't sure how helpful it would be. Do you have any example of "URL pattern/XPath"? Are you thinking of providing a DOM Path, so that the infobar can be placed anywhere in the page? How much would that complicate things in terms of styling the infobar?

I was thinking about something like [reactour](https://reactour.js.org/) (used on Pontoon tour), where you give it a CSS selector and it renders a popup around the element you want to highlight.

flodolo:
Updated, but I've left out the position for now. If I understand correctly, the suggestion is to have a CSS selector, and the infobar will be displayed as a pop-up there.

I don't mind adding it, but I wonder if it's broadening the scope too much:
* What's the use case? I can think of highlighting a specific section for the current first use case (new feature).
* How do we limit scope? Would it be OK to define it like this?
  * If empty, use default style and display at the top.
  * If defined, it will be displayed as a pop-up using a non customizable style.

mathjazz:
True about broadening the scope, but OTOH we can implement the feature according to the spec in several iterations, i.e. we could ship position as a followup. And it's good to be aware of potential followups in advance.

The use case is almost any new feature announcement. Imagine we ship Concordance search (or Comments, Terms, Notifications...). It's easier to understand the new feature if the popup is displayed near the newly introduced panel, tab, button or icon (versus in a generic place, which could be far away from the feature UI).

I agree with your proposal about definitions and defaults. We'll probably also need to define the URL (pattern) for pages on which to display the popup.

flodolo:
OK, added both fields. Let me know if they make sense.

specs/0109-infobar-component.md
```diff
@@ -0,0 +1,71 @@
...
+
+For each user, we need to store if an infobar has been dismissed, to avoid displaying it again. We store that information in a table with the following fields:
+* User ID.
+* Infobar ID.
+* Action date (datetime).
+
+We also want to keep track of every time we display infobars to each user. Every display action will be logged in a table with the following fields:
+* User ID.
+* Infobar ID.
+* Action date (datetime).
```
mathjazz, line 52: I am a bit torn if these tables should be new tables, or if we should use ActionLog instead.

mathjazz (approved):
I'd probably call it announcement instead of infobar, because it's not always a bar. But I don't have strong opinions. Other than that, LGTM! Thanks!



@Pike Let us know if you have further comments.

specs/0109-infobar-component.md
```diff
@@ -0,0 +1,71 @@
+* Feature Name: Infobar Component
+* Created: 2021-02-25
+* Associated Bugs: [bug 1694911](https://bugzilla.mozilla.org/show_bug.cgi?id=1694911)
+
+# Summary
+
+Create a reusable component to display infobars to users.
```
Pike, line 7: I'd drop the *reusable*.

specs/0109-infobar-component.md
```diff
@@ -0,0 +1,71 @@
...
+# Feature explanation
+
+When a logged in user opens Pontoon, we check against a list of infobars. If the user matches the necessary conditions, we display the infobar.
+
+Information about each infobar is stored in a table with the following fields:
+* Identifier: short text identifier, e.g. `new_terms_2020`.
+* Start time (datetime): when we should start displaying the infobar to users.
+* End time (datetime): when we should stop displaying the infobar to users.
+* Creation date (date): stores when the infobar was created.
+* Cohort: if defined, we only display the infobar for users part of a specific experiment or cohort (A/B testing).
```
Pike, line 28: Give this an `(optional)` to align with the other fields?

Also, should we have a modifier for things that we can't implement right now, as they're blocked by other things to be specified?

specs/0109-infobar-component.md
```diff
@@ -0,0 +1,71 @@
...
+* Selector (optional): a CSS selector.
+    * If empty, the infobar will be displayed at the top of the page.
+    * If defined, the infobar will be displayed as a pop-up above the element identified by the selector.
+* Pages pattern (optional): regular expression to indicate on which pages the infobar should be displayed.
+    * If empty, the infobar will be displayed on all pages.
+    * If defined, the infobar will be displayed only on pages whose URL matches the regular expression.
+* Title (optional): title for the infobar.
+* Text (mandatory): text for the infobar.
+    * Basic HTML should be allowed, including anchors.
```
Pike, line 37: This should be defined, as we need to instruct bleach on what to accept. Do we enforce a link target?

specs/0109-infobar-component.md
```diff
@@ -0,0 +1,71 @@
...
+* Button (optional): text for an optional button.
+    * The button will not be displayed if this field is empty.
+    * The only action associated to a button would be opening the URL specified in the URL field.
```
Pike, line 40: Same tab or new tab?

specs/0109-infobar-component.md
```diff
@@ -0,0 +1,71 @@
...
+We also want to keep track of every time we display infobars to each user. Every display action will be logged in a table with the following fields:
+* User ID.
+* Infobar ID.
+* Action date (datetime).
+
+# Dismissing infobars
+
+The infobar won’t be displayed again to the user if:
+* They click the close (x) button.
+* They click the button, if present.
```
Pike, line 58: Should the infobar be dismissed if a user follows the link in the text?

specs/0109-infobar-component.md
```diff
@@ -0,0 +1,71 @@
...
+
+When one of these actions are performed, the information is stored in the dismissals table.
+
+# Managing multiple infobars
+
+If there are multiple infobars active, we only display the most recent based on creation date.
```
Pike, line 64: Does that mean that once we stop showing the newest, we show the next-oldest one? If so, when?

I'm a bit concerned about this user experience.

This makes me think that we should limit the number of concurrently available infobars. And if that number is greater than 1, we should probably also throttle the rate at which we show them.

specs/0109-infobar-component.md
```diff
@@ -0,0 +1,71 @@
...
+
+# Out of scope
+
+The following features are out of scope:
+* Writing a dashboard to manage infobars. For the initial phase, working directly via the Django Admin interface would be sufficient.
+* Tracking changes to existing infobars.
+* Customizing the aspect of the infobars.
```
Pike, line 71: The aspect ratio only matters for over-element popups, right? What's the aspect ratio we're starting with for those?

specs/0109-infobar-component.md
```diff
@@ -0,0 +1,71 @@
...
+* Action date (datetime).
+
+We also want to keep track of every time we display infobars to each user. Every display action will be logged in a table with the following fields:
+* User ID.
+* Infobar ID.
+* Action date (datetime).
+
+# Dismissing infobars
+
+The infobar won’t be displayed again to the user if:
```
Pike, line 56: I think we should have one section that defines when to show the infobar, and leave this section just to define when to add Dismiss actions.

Pike (changes requested):
Sorry for the late arrival of this many comments, the earlier iterations had more conversations in flight than I was comfortable in following.

specs/0109-infobar-component.md
flodolo: I think it's OK to define a field in the spec for future use, without explicitly call it out? The alternative would be to drop it, and add it later if/when we decide to implement something that uses it.

specs/0109-infobar-component.md
flodolo: Listing `<p>`, `<strong>`, `<em>`, `<a>`, I don't think we'd need more.

specs/0109-infobar-component.md
flodolo: I can't really answer that ;-) Does it need to be solved at spec level, or that's an implementation detail?

specs/0109-infobar-component.md
flodolo: Not exactly sure how to address this comment :-| 

This section describes when we consider an infobar dismissed, and what happens when we do. 

When we display the infobar is at the beginning of the feature explanation section (which I'll rephrase a bit).

specs/0109-infobar-component.md
flodolo: Per Matjaz previous suggestion, we shouldn't (that was my original version, but I agree with him).

If someone clicks on a link, it doesn't mean it has read the link and acknowledged the infobar. Also, what if there are multiple links?

specs/0109-infobar-component.md
flodolo: I think having more than one infobar active is an exception. This is just describing how to deal with it without picking one at random.

specs/0109-infobar-component.md
flodolo: I consider that an implementation detail, since that's not customizable. Likewise, we're not defining aspects like color or font for the normal infobar.

Pike: TBH, I don't understand what your original sentence is about. I guessed "aspect ratio", but based on your reply, that's not what that is.

flodolo: What is "aspect ratio" then? 
https://en.wikipedia.org/wiki/Aspect_ratio_(image)

mathjazz (approved):
I think this is good as is.

If needed, we can iterate further.

PR #1878, opened 2021-02-25
Upgrade flow to 0.145.0

mathjazz:
We know migrating to TypeScript will be time consuming and will probably require frontend code freeze:
https://bugzilla.mozilla.org/show_bug.cgi?id=1685565

Given [flow](https://github.com/facebook/flow/commits/master) is far from being dead and we use a very old version of it, let's see if it makes sense to update to the latest version.

Maybe it will help with the migration to TS?
Maybe it will be almost as hard to do?
Maybe if we upgrade flow, we don't need to migrate to TS?

This is just a start. 791 errors. :)

Probably mostly caused by https://flow.org/en/docs/lang/types-first/. Although... there are 423 errors already when upgrading to 0.124.0 (Types-First was introduced by 0.125.0).

### DIFF
--- a/frontend/.flowconfig
+++ b/frontend/.flowconfig
@@ -11,8 +11,4 @@ module.system=node
 module.system.node.resolve_dirname=node_modules
 module.system.node.resolve_dirname=./src
 
-# This is used to ignore some Flow errors if needed.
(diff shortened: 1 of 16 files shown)


mathjazz:
Closing on behalf of https://bugzilla.mozilla.org/show_bug.cgi?id=1696517.

PR #1880, opened 2021-02-26
Fix bug 1695226: Sort projects in Concordance search results by priority and disabled status

mathjazz:
This works if I remove the `distinct=True` (but then projects are duplicated), otherwise results in an error shown below. I guess that's because ordering of duplicate items could be ambiguous if items would be ordered by some other field.

I wonder if there's a more elegant way around this than implementing distinct or ordering in python.

```
  File "/usr/local/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 179, in _get_response
  File "/app/pontoon/machinery/views.py", line 80, in concordance_search
    paginator = Paginator(get_concordance_search_data(text, locale), page_results_limit)
  File "/app/pontoon/machinery/utils.py", line 107, in get_concordance_search_data
    return sorted(search_results, key=sort_by_quality, reverse=True)
  File "/usr/local/lib/python3.8/site-packages/django/db/models/query.py", line 287, in __iter__
    self._fetch_all()
  File "/usr/local/lib/python3.8/site-packages/django/db/models/query.py", line 1308, in _fetch_all
    self._result_cache = list(self._iterable_class(self))
  File "/usr/local/lib/python3.8/site-packages/django/db/models/query.py", line 111, in __iter__
    for row in compiler.results_iter(chunked_fetch=self.chunked_fetch, chunk_size=self.chunk_size):
  File "/usr/local/lib/python3.8/site-packages/django/db/models/sql/compiler.py", line 1108, in results_iter
    results = self.execute_sql(MULTI, chunked_fetch=chunked_fetch, chunk_size=chunk_size)
  File "/usr/local/lib/python3.8/site-packages/django/db/models/sql/compiler.py", line 1156, in execute_sql
    cursor.execute(sql, params)
  File "/usr/local/lib/python3.8/site-packages/django/db/backends/utils.py", line 98, in execute
    return super().execute(sql, params)
  File "/usr/local/lib/python3.8/site-packages/django/db/backends/utils.py", line 66, in execute
    return self._execute_with_wrappers(sql, params, many=False, executor=self._execute)
  File "/usr/local/lib/python3.8/site-packages/django/db/backends/utils.py", line 75, in _execute_with_wrappers
    return executor(sql, params, many, context)
  File "/usr/local/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
  File "/usr/local/lib/python3.8/site-packages/django/db/utils.py", line 90, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/usr/local/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
django.db.utils.ProgrammingError: in an aggregate with DISTINCT, ORDER BY expressions must appear in argument list
LINE 1: ...ARRAY_AGG(DISTINCT "base_project"."name" ORDER BY "base_proj...
                                                             ^
```

### DIFF
--- a/pontoon/machinery/utils.py
+++ b/pontoon/machinery/utils.py
@@ -78,7 +78,7 @@ def get_concordance_search_data(text, locale):
         .values("source", "target")
         .annotate(
             project_names=ArrayAgg(
-                "project__name", ordering=["project__disabled", "-project__priority",],
+                "project__name", ordering=["project__disabled", "-project__priority"]
             )
         )
         .distinct()
--- a/pontoon/machinery/utils.py
+++ b/pontoon/machinery/utils.py
@@ -76,7 +76,13 @@ def get_concordance_search_data(text, locale):
     search_results = (
         base.models.TranslationMemoryEntry.objects.filter(search_query)
         .values("source", "target")
-        .annotate(project_names=ArrayAgg("project__name", distinct=True))
+        .annotate(
+            project_names=ArrayAgg(
+                "project__name",
+                distinct=True,
+                ordering=["-project__priority", "project__disabled"],
+            )
+        )
         .distinct()
     )
 
--- a/pontoon/machinery/utils.py
+++ b/pontoon/machinery/utils.py
@@ -78,13 +78,18 @@ def get_concordance_search_data(text, locale):
         .values("source", "target")
         .annotate(
             project_names=ArrayAgg(
-                "project__name",
-                distinct=True,
-                ordering=["-project__priority", "project__disabled"],
+                "project__name", ordering=["project__disabled", "-project__priority",],
             )
         )
         .distinct()
     )
+
+    # We need to remove duplicates manually - ArrayAgg does not support using distinct=True
+    # in combination with ordering.
+    for s in search_results:
+        # Fastest way to eliminate hashable duplicates while retaining order:
+        # https://twitter.com/raymondh/status/944125570534621185
+        s["project_names"] = list(dict.fromkeys(s["project_names"]))
 
     def sort_by_quality(entity):
         """Sort the results by their best Levenshtein distance from the search query"""
--- a/pontoon/machinery/utils.py
+++ b/pontoon/machinery/utils.py
@@ -76,11 +76,7 @@ def get_concordance_search_data(text, locale):
     search_results = (
         base.models.TranslationMemoryEntry.objects.filter(search_query)
         .values("source", "target")
-        .annotate(
-            project_names=ArrayAgg(
-                "project__name", ordering=["project__disabled", "-project__priority"]
-            )
-        )
+        .annotate(project_names=ArrayAgg("project__name", distinct=True))
         .distinct()
     )
 
--- a/pontoon/machinery/utils.py
+++ b/pontoon/machinery/utils.py
@@ -84,13 +84,6 @@ def get_concordance_search_data(text, locale):
         .distinct()
     )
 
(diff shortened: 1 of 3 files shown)


mathjazz:
That's the best I could do.

Related thread from PostgreSQL forum:
https://www.postgresql.org/message-id/flat/02b001cc6837%24555dea90%240019bfb0%24%40yahoo.com

@Pike or @jotes r?

pontoon/machinery/utils.py
```diff
@@ -76,10 +76,21 @@ def get_concordance_search_data(text, locale):
...
+            project_names=ArrayAgg(
+                "project__name", ordering=["project__disabled", "-project__priority"]
+            )
+        )
         .distinct()
     )
 
+    # We need to remove duplicates manually - ArrayAgg does not support using distinct=True
+    # in combination with ordering.
+    for s in search_results:
```
jotes, line 89: Do we have to remove duplicates before the pagination? I think we can move this after the pagination is applied to reduce the number of processed rows.

Pike: What jotes says, and also, would it be useful to just shell out the project names into a separate query and order there? So, first get project IDs with distinct, and then query with ordering.

Maybe also too complicated, I'd make the call depending on how many duplicates there are in practice.

jotes (approved):
I guess that a better/performance-wise way would involve a (complicated?) sub-query. Unfortunately, I don't have a straightforward alternative.

pontoon/machinery/utils.py
mathjazz: @Pike Not sure I understand your part. I'd annotate distinct project IDs (instead of ordered project names) in the `search_results` QuerySet? And then I'd order them and convert them to project names in the loop, using the ordered list of project name values?

Pike: Right. I noticed that I couldn't just write out what I had in mind, but the gist of it was "do two queries, one for distinct, and one for ordering, and then stitch the results together in a loop"

mathjazz: Updated.

I've also included a CSS fix to reveal all project names on hover even if the user is logged out or the project is read-only.

pontoon/machinery/views.py
```diff
@@ -84,6 +84,15 @@ def concordance_search(request):
     except EmptyPage:
         return JsonResponse({"results": [], "has_next": False})
 
+    # ArrayAgg (used in get_concordance_search_data()) does not support using
+    # distinct=True in combination with ordering, so we need to do one of them
+    # manually - after pagination, to reduce the number of rows processed.
+    projects = Project.objects.order_by("disabled", "-priority").values_list(
+        "name", flat=True
```
jotes, line 91: [performance nitpick] -> the query can filter them by the project names returned in search results?

Pike: I'm OK with just a single list, I wonder how much work we'd invest into gathering the projects set from all the object_list first.

Pike (approved):
Adding an r=me.

pontoon/machinery/views.py
mathjazz: Something like this would work I think:
```python
.filter(name__in=[x for r in data.object_list for x in r["project_names"]])
```

But I don't see any performance wins locally, so I'd rather keep the code less ugly. :)

PR #1881, opened 2021-02-26
Bug 1694598, separate schema and data migrations

Pike:


Pike:
This is ready for "review", if you're willing to call it that. @jotes , do you want to take this?

I'm sure we want this on stage, too, just to make sure it upgrades nicely.

@flozz , this passes tests on postgresql 13 for me locally, it'd be great if you could give this branch a spin, too.

jotes:
Sure, i'll take a look later!

mathjazz:
\o/

Deployed to stage. Didn't spot any issues. New migration applied without a hassle.

flozz:
It from PostgreSQL 9.6 to 13 \o/

→ https://github.com/wanadev/pontoon-debian/runs/1990646006?check_suite_focus=true

(Python 3.7 fails, it is "normal", I have not looked yet to its dependency issue)

Pike:
FTR, I'm having some self-concerns about more ordering problems.

jotes:
@Pike AFAIR, I had to split migrations due to a strange circular-like dependency between `base` and `checks`/`tutorial`(?) migrations. Is that related to your concerns? The migrations look okay at the first glance, but I need to take a break and look at this patch today again.

Pike:
Nah, I'm talking about 

```
            options={
                "ordering": ("pk",),
            },
```

Pike:
> 
> 
> FTR, I'm having some self-concerns about more ordering problems.

I cross-checked orderings, and I think those should be good.

Ready for review.

pontoon/base/migrations/0001_squashed_0154_auto_20200206_1736.py
```diff
@@ -3378,320 +842,76 @@ class Migration(migrations.Migration):
...
+                        on_delete=django.db.models.deletion.SET_NULL,
                         related_name="changed_permissions_log",
                         to=settings.AUTH_USER_MODEL,
                     ),
                 ),
                 (
                     "performed_on",
                     models.ForeignKey(
-                        on_delete=django.db.models.deletion.CASCADE,
+                        null=True,
```
jotes, line 854: This will require a few additional changes e.g. [here](
https://github.com/mozilla/pontoon/blob/d73095890882b65e6c7d5c73da1546ae8bdf130c/pontoon/base/admin.py#L355).

jotes (approved):
The overall idea of keeping the data migrations separately is okay. Nonetheless, I didn't test the whole patch extensively on a fresh Pontoon's instance (but I feel it's needed).

pontoon/base/migrations/0001_squashed_0154_auto_20200206_1736.py
mathjazz: I don't think so. We're not actually making changes to the model, we're just squashing the schema part of migrations 0001 and 0002 into 0001, and the data part of migrations 0001 and 0002 into 0002.

See:
https://github.com/mozilla/pontoon/pull/1881/files#diff-eb0ed65c4398ca63ffc99e9d41a7417e1539dd0976ce378ad8219647e7f29b56L38

jotes: Ah, I've got confused by the way how the diff was displayed.

PR #1883, opened 2021-02-27
Fix bug 1681653: Remove invalid characters from TM strings while writing tmx files.

VishalCR7:
Following string in translation memory has a special character considered invalid for XML format hence which prevents tmx file from parsing 

`"Portions of this content are c.1998–2019 by individual mozilla.org "
"contributors. Content available under a <a rel=\"external noopener\" "
"href=\"https://www.mozilla.org/foundation/licensing/website-"
"content/\">Creative Commons license</a>."`

Got rid of such characters from the source and target strings to generate a valid tmx file

Reference: https://stackoverflow.com/questions/1707890/fast-way-to-filter-illegal-xml-unicode-chars-in-python

### DIFF
--- a/pontoon/base/utils.py
+++ b/pontoon/base/utils.py
@@ -3,6 +3,7 @@ import functools
 import io
 import os
 import re
+import sys
 
 import pytz
 import requests
@@ -493,6 +494,50 @@ def get_last_months(n):
         start_date += relativedelta(months=-1)
 
 
+def sanitise_xml_input_string(string):
+    """
+    The XML specification (http://www.w3.org/TR/xml11/#charsets) lists a bunch of Unicode characters
+    that are either illegal or "discouraged". Relace these characters to get valid XML strings
+    """
+
+    illegal_unichrs = [
+        (0x00, 0x08),
+        (0x0B, 0x0C),
+        (0x0E, 0x1F),
+        (0x7F, 0x84),
+        (0x86, 0x9F),
+        (0xFDD0, 0xFDDF),
+        (0xFFFE, 0xFFFF),
+    ]
+    if sys.maxunicode >= 0x10000:  # not narrow build
+        illegal_unichrs.extend(
+            [
+                (0x1FFFE, 0x1FFFF),
+                (0x2FFFE, 0x2FFFF),
+                (0x3FFFE, 0x3FFFF),
+                (0x4FFFE, 0x4FFFF),
+                (0x5FFFE, 0x5FFFF),
+                (0x6FFFE, 0x6FFFF),
+                (0x7FFFE, 0x7FFFF),
+                (0x8FFFE, 0x8FFFF),
+                (0x9FFFE, 0x9FFFF),
+                (0xAFFFE, 0xAFFFF),
+                (0xBFFFE, 0xBFFFF),
+                (0xCFFFE, 0xCFFFF),
+                (0xDFFFE, 0xDFFFF),
+                (0xEFFFE, 0xEFFFF),
+                (0xFFFFE, 0xFFFFF),
+                (0x10FFFE, 0x10FFFF),
+            ]
+        )
+
+    illegal_ranges = [fr"{chr(low)}-{chr(high)}" for (low, high) in illegal_unichrs]
+    xml_illegal_character_regex = "[" + "".join(illegal_ranges) + "]"
+    illegal_xml_chars_re = re.compile(xml_illegal_character_regex)
+
+    return illegal_xml_chars_re.sub("", string)
+
+
 def build_translation_memory_file(creation_date, locale_code, entries):
     """
     TMX files will contain large amount of entries and it's impossible to render all the data with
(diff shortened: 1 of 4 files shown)


pontoon/base/utils.py
```diff
@@ -493,6 +494,50 @@ def get_last_months(n):
...
+                (0xDFFFE, 0xDFFFF),
+                (0xEFFFE, 0xEFFFF),
+                (0xFFFFE, 0xFFFFF),
+                (0x10FFFE, 0x10FFFF),
+            ]
+        )
+
+    illegal_ranges = [fr"{chr(low)}-{chr(high)}" for (low, high) in illegal_unichrs]
+    xml_illegal_character_regex = "[" + "".join(illegal_ranges) + "]"
+    illegal_xml_chars_re = re.compile(xml_illegal_character_regex)
```
mathjazz, line 536: Seems like the following regex does the job as well:
```python
illegal_xml_chars_re = re.compile(u'[\x00-\x08\x0b\x0c\x0e-\x1F\uD800-\uDFFF\uFFFE\uFFFF]')
```

Source: https://lsimons.wordpress.com/2011/03/17/stripping-illegal-characters-out-of-xml-in-python/

mathjazz: I haven't tested it, but I wonder if we could also use this approach:
https://github.com/translate/translate/blob/master/translate/misc/xml_helpers.py#L178

pontoon/base/utils.py
```diff
@@ -508,31 +553,34 @@ def build_translation_memory_file(creation_date, locale_code, entries):
...
+        ' creationtool="pontoon"'
+        ' datatype="plaintext"'
+        ' segtype="sentence"'
+        ' o-tmf="plain text"'
+        ' srclang="en-US"'
+        ' creationdate="%(creation_date)s">'
+        "\n\t</header>"
+        "\n\t<body>" % {"creation_date": creation_date.isoformat()}
     )
     for resource_path, key, source, target, project_name, project_slug in entries:
```
mathjazz, line 570: This is not strictly related to the bug, and the TMX output could be refactored into something more structured:

We don't seem to use `project_name` anywhere, so we could remove it while we're at it.

mathjazz (commented):
Nice work, this fixes the bug!

I'm surprised I couldn't find validation logic we could just import from an existing library. I found a couple of alternative implementation we could use.

pontoon/base/utils.py
```diff
@@ -493,6 +494,50 @@ def get_last_months(n):
         start_date += relativedelta(months=-1)
 
 
+def sanitise_xml_input_string(string):
```
flodolo, line 497: Should we stick to American spelling? (sanitize)

pontoon/base/utils.py
```diff
@@ -493,6 +494,50 @@ def get_last_months(n):
...
+    """
+    The XML specification (http://www.w3.org/TR/xml11/#charsets) lists a bunch of Unicode characters
+    that are either illegal or "discouraged". Relace these characters to get valid XML strings
```
flodolo, line 500: typo: replace

Also bunch -> set

mathjazz:
@flodolo Deploying to stage if you want to take it for a spin - should be up in around 10 minutes.

flodolo:
Seems to work. Tested with xmllint, no errors. This is the output from prod

```
$ xmllint --noout it.all-projects.tmx
it.all-projects.tmx:539393: parser error : PCDATA invalid Char value 3
				<seg>Portions of this content are c.1998–2019 by individual mozilla.org c
				                                  ^
it.all-projects.tmx:539401: parser error : PCDATA invalid Char value 3
				<seg>Portions of this content are c.1998–2019 by individual mozilla.org c
				                                  ^
it.all-projects.tmx:539404: parser error : PCDATA invalid Char value 3
				<seg>Parte di questi contenuti sono c.1998–2019 di singoli collaboratori
				                                    ^
it.all-projects.tmx:649838: parser error : Opening and ending tag mismatch: body line 5 and tu
		</tu>
		     ^
it.all-projects.tmx:655528: parser error : Opening and ending tag mismatch: tmx line 2 and body
	</body>
	       ^
it.all-projects.tmx:655529: parser error : Extra content at the end of the document
</tmx>
^
```

Note to self: when the fix is deployed, I need to remove the hacks from the mt-training-data repo and my own scripts.

Pike:
Question: If the intent is to use this data for training, should we train with content that has encoding errors, and then fixups to that that skip individual bytes?

Or should TUs with encoding errors be skipped as a whole?

flodolo:
> Question: If the intent is to use this data for training, should we train with content that has encoding errors, and then fixups to that that skip individual bytes?

This is about Pontoon generating broken XML that can't be used anywhere, since you can't parse them as XML.

mathjazz:
> Or should TUs with encoding errors be skipped as a whole?

@flodolo What's your preference? Skip illegal characters only or skip entire TUs with illegal characters?

flodolo:
> > Or should TUs with encoding errors be skipped as a whole?
> 
> @flodolo What's your preference? Skip illegal characters only or skip entire TUs with illegal characters?

Given they're control characters that are not supposed to influence the translation, I'd just strip the character.

mathjazz:
Please also remove `project_name` from the docstring of `build_translation_memory_file` and we're good!

PR #1884, opened 2021-03-01
Bug 1694598 Check that Pontoon is installable on standard Linux disto

flozz:
This PR adds a Github Actions workflow that installs Pontoon using various Python and PostgreSQL version to check that it not breaks.

→ see https://bugzilla.mozilla.org/show_bug.cgi?id=1694598 for more details

*This PR should not be merged before PR #1881 :)*

### DIFF
--- a/.github/workflows/linux-install.yml
+++ b/.github/workflows/linux-install.yml
@@ -2,6 +2,8 @@ name: Linux Install Test
 
 on:
   push:
+    branches:
+      - master
     paths:
       - requirements/**
       - pontoon/**
--- a/.github/workflows/linux-install.yml
+++ b/.github/workflows/linux-install.yml
@@ -13,39 +13,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        include:
-          - name: "Debian 10 (buster)"  # -> 2024-06
-            os: ubuntu-18.04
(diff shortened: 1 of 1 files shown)


mathjazz (commented):
Thanks for the patch!

@Pike Would you have cycles for a review?

A few high-level questions and suggestions:
1. How long does it take for the workflow to complete?
2. Why do we only run it on push, and not also on pull_requests?
3. We should add some paths when to run it. Changing `/frontend` for example shouldn't trigger it.
4. Should the step be just to `make build`?

flozz:
1. It takes about 5-6min to complete
2. Because I forgot to add PR and to remove my test branch ^^' (shout it run only on master or on any push?)
3. Ok I will look at this :)
4. I think not: the `Makefile` uses Docker to build to frontend, I try to test an end-to-end build + install on a "classic" Linux server (generally we do not want to install Docker only to build an app on a production server)

flozz:
Ready to review :)

* I rebased my branch on the currnet `master`
* The workflow now run on both `push` and PR
* I added some path to avoid running it on frontend changes

mathjazz:
Closing as per https://bugzilla.mozilla.org/show_bug.cgi?id=1694598#c8.

PR #1885, opened 2021-03-03
Bug 1696146, pyupgrade --py36-plus, black, flake8

Pike:
As is often the case, adding `--py38-plus` didn't impact the results. I tried to find which codemods run there, only found some generator stuff, but also didn't find exactly how all that logic works.

We could also run pyupgrade as a linter, to make sure we don't regress. @mathjazz , what do you think?


mathjazz (approved):
LGTM.

Definitely a good idea to run it as a linter.

PR #1886, opened 2021-03-03
bug 1696146, add linting for pyupgrade

Pike:


mathjazz (commented):
In #1885 you had to run `black` and `flake8` **after** `pyupgrade`. Do we need to make the jobs sequential?

Also, I'd extend `make format` to also include `pyupgrade` and `flake8`.

Other than that it looks good!

Pike:
> In #1885 you had to run `black` and `flake8` **after** `pyupgrade`. Do we need to make the jobs sequential?

As these are just linters, no, they can just run in parallel. Actually has the benefit that you're getting all linters' feedback even if one fails.

> Also, I'd extend `make format` to also include `pyupgrade` and `flake8`.

I added `pyupgrade`, and before `black`. Also had to tweak the pyupgrade target to not fail when it transforms. `flake8` is just a linter, not a transformer, so there's no point adding it to `format`.

mathjazz (approved):
Oh, good point about the order. We're only checking. :-)

And yeah, `flake8` doesn't transform.

PR #1887, opened 2021-03-05
Fix bug 1478554: Add missing (no suggestions) filter to translate app

VishalCR7:
Added new filter to get strings with no or unrejected suggestions.

We first filter out entities with either approved or not rejected translations (for all plurals) and return a filter to not consider these.

### DIFF
--- a/frontend/src/modules/search/components/SearchBox.css
+++ b/frontend/src/modules/search/components/SearchBox.css
@@ -120,6 +120,7 @@
 .filters-panel .menu .missing-without-unreviewed.selected .status::before,
 .filters-panel .menu .missing-without-unreviewed .status:hover:before {
     font-size: 16px;
+    margin-left: 0;
 }
 
 .search-box .tag .status:before {
--- a/frontend/src/modules/search/components/SearchBox.css
+++ b/frontend/src/modules/search/components/SearchBox.css
@@ -113,6 +113,13 @@
 
 .search-box .missing-without-unreviewed .status:before {
     content: '\f0eb';
+    font-size: 18px;
+    margin-left: 1px;
+}
+
+.filters-panel .menu .missing-without-unreviewed.selected .status::before,
+.filters-panel .menu .missing-without-unreviewed .status:hover:before {
+    font-size: 16px;
 }
 
 .search-box .tag .status:before {
--- a/frontend/src/modules/search/components/SearchBox.css
+++ b/frontend/src/modules/search/components/SearchBox.css
@@ -111,8 +111,8 @@
     content: '';
 }
 
-.search-box .missing-no-suggestions .status:before {
-    content: 'Ⓜ';
+.search-box .missing-without-unreviewed .status:before {
+    content: '\f0eb';
 }
 
 .search-box .tag .status:before {
--- a/frontend/src/modules/search/components/SearchBox.css
+++ b/frontend/src/modules/search/components/SearchBox.css
@@ -111,6 +111,10 @@
     content: '';
 }
 
+.search-box .missing-no-suggestions .status:before {
+    content: 'Ⓜ';
+}
+
 .search-box .tag .status:before {
     content: '';
 }
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -2328,8 +2328,8 @@ class EntityQuerySet(models.QuerySet):
         return ~Q(
             pk__in=self.get_filtered_entities(
                 locale,
-                Q(approved=True) | Q(rejected=False),
-                lambda x: x.approved or not x.rejected,
+                Q(approved=True) | Q(fuzzy=True) | Q(rejected=False),
+                lambda x: x.approved or x.fuzzy or not x.rejected,
                 match_all=True,
                 project=project,
             )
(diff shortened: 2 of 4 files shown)


VishalCR7:
@mathjazz, please review (:

pontoon/base/models.py
```diff
@@ -2316,6 +2316,25 @@ def rejected(self, locale, project=None):
...
+
+        :returns: a django ORM Q object to use as a filter
+
+        """
+        return ~Q(
+            pk__in=self.get_filtered_entities(
+                locale,
+                Q(approved=True) | Q(rejected=False),
+                lambda x: x.approved and not x.rejected,
+                match_all=True,
```
mathjazz, line 2333: `False`: even if one plural form has missing strings only, the filter should catch it.

pontoon/base/models.py
```diff
@@ -2316,6 +2316,25 @@ def rejected(self, locale, project=None):
...
+        :arg Locale locale: a Locale object to get translations for
+
+        :returns: a django ORM Q object to use as a filter
+
+        """
+        return ~Q(
+            pk__in=self.get_filtered_entities(
+                locale,
+                Q(approved=True) | Q(rejected=False),
+                lambda x: x.approved and not x.rejected,
```
mathjazz, line 2332: This rule must match the query above (`and not` => `or`).

pontoon/base/models.py
```diff
@@ -2316,6 +2316,25 @@ def rejected(self, locale, project=None):
...
+        This filter will return all entities that have no or only rejected translations.
+        :arg Locale locale: a Locale object to get translations for
+
+        :returns: a django ORM Q object to use as a filter
+
+        """
+        return ~Q(
+            pk__in=self.get_filtered_entities(
+                locale,
+                Q(approved=True) | Q(rejected=False),
```
mathjazz, line 2331: We also need to exclude Fuzzies.

mathjazz (commented):
Very nice work!

Please add tests in `pontoon/base//tests/managers/test_entity.py`.

pontoon/base/models.py
VishalCR7: `True` in this case will get us entity ids for which has an approved or unreviewed string is present for "all" plurals (not of this will be at least one) . So I think this is correct. 

I started with False as first instinct but testing it out with True gives the right behaviour

pontoon/base/models.py
VishalCR7: Fuzzies I think will fall in one of the two categories. Can you please give me an example where this is the case?  Rejected=False is good enough to check the presence of Fuzzies as well

pontoon/base/models.py
mathjazz: Good point! Note that you can also remove the attribute (to be consistent with the `missing()` filter we're mimicing here), because the default value is `True`.

pontoon/base/models.py
mathjazz: You're right again! Could you please add a comment about `rejected=False` also catching `fuzzy=True`? I got confused when comparing with the `missing()` filter, maybe somebody else will, too.

frontend/src/modules/search/components/SearchBox.css
```diff
@@ -111,6 +111,10 @@
     content: '';
 }
 
+.search-box .missing-without-unreviewed .status:before {
+    content: '\f0eb';
+}
```
mathjazz, line 116: Please make sure font-size is the same as for Unreviewed and that the icon is aligned with other Extra filters icons (probably needs margin-left of 1px).

And then also make sure the selected state icon is styled properly (font size and alignment).

pontoon/base/tests/managers/test_entity.py
```diff
@@ -650,6 +650,107 @@ def test_mgr_entity_filter_rejected_plural(resource_a, locale_a):
...
+    TranslationFactory.create(
+        locale=locale_a, entity=entities[1], approved=False, fuzzy=False, rejected=True,
+    )
+    TranslationFactory.create(
+        locale=locale_a,
+        entity=entities[1],
+        approved=False,
+        fuzzy=False,
+        rejected=False,
+    )
```
mathjazz, line 681: Hmmm, I find it weird that black formats this block differently than the previous ones. 👀

mathjazz (commented):
Just a couple of nits and we're good!

pontoon/base/tests/managers/test_entity.py
VishalCR7: Length of string probably

pontoon/base/models.py
VishalCR7: I still added fuzzies to make it easy to understand. Let me know if we want to change it

pontoon/base/tests/managers/test_entity.py
mathjazz: Ah, now I see it's one more character in this line. Good point!

mathjazz:
Please also fix margin in selected (and hover) state:

Screenshot 2021-03-07 at 11 32 41

mathjazz (approved):
Yay!

PR #1888, opened 2021-03-07
Fix Bug 1696517 - Upgrade flow

abowler2:
This work begins the process of upgrading Flow Type from version 0.95.1 to 0.146.0

To accomplish this I first updated to 0.125.1 as that was the first stable release that introduced 'Types-First'. With this change I needed to make an adjustment into the config file to allow the way we import from our `core` and `modules` files.

From there it became apparent that the main issue we were going to have was directly related to 'Types-First' and that since it had been fully implemented at version 0.134.0 there was [documentation](https://flow.org/en/docs/lang/types-first/) and a [codemod](https://flow.org/en/docs/cli/annotate-exports/) in place to help with the transition. So, I opted to just go straight to the latest version.

This required an additional change to the config file and after running the codemod there were still a few errors to contend with (32 total).

I made an initial attempt at the errors and was able to bring it down to 19 errors between 3 files (SearchBox.js has 16 of them due to how some of the props are typed). I was not able to work out those type errors on my first pass, so figured I would open up this PR so others could take a look.

Of note, many of the changes that were put in place by both the codemod and myself use the `any` option. It would likely be a good idea to audit the use of `any` within the code (either in this PR or a separate one) to ensure that we have proper type coverage.


Pike:
I got the SearchBox.js done, by using `{ [string]: boolean }` instead of `{...}`

frontend/src/core/editor/hooks/useHandleShortcuts.js
```diff
@@ -148,7 +148,7 @@ export default function useHandleShortcuts(): ((
             }
 
             const numTranslations =
-                translations.length + (searchResults && searchResults.length);
+                translations.length + (searchResults?.length || 0);
```
mathjazz, line 151: I love optional chaining!

We'll need to update our [browser support matrix](https://github.com/mozilla/pontoon/blob/master/CONTRIBUTING.rst#browser-support) to start [using](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#browser_compatibility) it, though. Which is probably out of date, anyways.

Given Firefox ESR supports it, I'm fine with the change.

mathjazz (commented):
Nice catch with the `codemod`!

Thanks to Axel, the only remaining errors are caused by `ReactLocalization`.

Pike:
I got this down to one error via the following libdef in `frontend/flow-typed`:

```js
declare module "@fluent/react" {

    declare export class ReactLocalization {
        constructor(Array<any>) : ReactLocalization
    }

    declare export function LocalizationProvider(props:any) : React$Element<any>;

    declare export interface LocalizedProps {
        id: string;
        attrs?: {[string]: boolean};
        children?: React$Node;
        vars?: {[string]: any};
        elems?: {[string]: ?React$Element<any>}
    }
    declare export function Localized(props:LocalizedProps): React$Element<any>;
}
```

That's inspired from the typedefs in typescript that come with `@fluent/react`, which I don't understand, though.

Oh, and I had to tweak ` glyph: ?React.Element<any>,` in `frontend/src/core/editor/components/EditorMainAction.js`

Pike:
Remaining error:

```
yarn run v1.22.10
$ flow
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/core/notification/components/NotificationPanel.js:73:51

Cannot create span element because in property children: [incompatible-type]
 • Either inexact function type [1] is incompatible with exact object type [2].
 • Or function type [1] is incompatible with React.Portal [3].
 • Or property @@iterator is missing in statics of function type [4] but exists in $Iterable [5].

     src/core/notification/components/NotificationPanel.js
     70│                 {!notif ? (
     71│                     <span />
     72│                 ) : (
     73│                     <span className={notif.type}>{notif.content}</span>
     74│                 )}
     75│             </div>
     76│         );

     flow-typed/react.js
 [4] 16│     declare export function Localized(props:LocalizedProps): React$Element<any>;

     /tmp/flow/flowlib_545003c912489407_1000/react.js
 [2] 18│   | React$Element<any>
 [3] 19│   | React$Portal
 [5] 20│   | Iterable<?React$Node>;

     src/core/notification/actions.js
 [1] 17│     +content: string | typeof Localized,



Found 1 error

Only showing the most relevant union/intersection branches.
To see all branches, re-run Flow with --show-all-branches
error Command failed with exit code 2.
```

frontend/src/core/editor/hooks/useHandleShortcuts.js
abowler2: Yay!!!

abowler2:
Thank you for your help @Pike!




Sadly, I'm stuck at this last error message as well. If I remove `typeof Localized` from `content` in `type NotificationMessage` it clears, but I'm uncertain of the impact of that.

Pike:
This works for me:

```diff
diff --git a/frontend/src/core/notification/actions.js b/frontend/src/core/notification/actions.js
index 968ddb61..3af37b59 100644
--- a/frontend/src/core/notification/actions.js
+++ b/frontend/src/core/notification/actions.js
@@ -14,7 +14,7 @@ export type NotificationType =

 export type NotificationMessage = {|
     +type: NotificationType,
-    +content: string | typeof Localized,
+    +content: string | React$Element<any>,
     +key?: string,
 |};
```

abowler2:
Removing the `WIP` status. This is now ready for review 😅

frontend/src/core/editor/hooks/useHandleShortcuts.js
Pike: tl;dr: chaining works fine w/out any action required.

I've spent a few hours to find out what's going on here. `react-scripts` hard-code transpiling optional chaining away, due to webpack due to babel due to acorn. Or so. There's tons of major updates out there in our frontend build stack.

frontend/src/modules/entitieslist/components/EntitiesList.js
```diff
@@ -1,5 +1,6 @@
 /* @flow */
 
+import type { Node, Element } from 'React';
 import React from 'react';
```
Pike, line 4: I'd prefer to do this the other way around, use `React` as type namespace, and explicitly import `createRef`.

frontend/src/modules/entitieslist/components/EntitiesList.js
```diff
@@ -384,7 +395,7 @@ export class EntitiesListBase extends React.Component<InternalProps> {
     }
 }
 
-export default function EntitiesList() {
+export default function EntitiesList(): Node {
```
Pike, line 398: `Element<typeof EntitiesListBase>`

frontend/src/modules/entitydetails/components/EntityDetails.js
```diff
@@ -509,7 +526,7 @@ export class EntityDetailsBase extends React.Component<InternalProps, State> {
     }
 }
 
-export default function EntityDetails() {
+export default function EntityDetails(): React.Node {
```
Pike, line 529: `React.Element<typeof EntityDetailsBase>`

frontend/src/modules/entitydetails/components/GenericOriginalString.js
```diff
@@ -56,7 +56,7 @@ function getOriginalContent(props: Props) {
  * Based on the plural form, show either the singular or plural version of the
  * string, and also display which form is being rendered.
  */
-export default function GenericOriginalString(props: Props) {
+export default function GenericOriginalString(props: Props): React.Node {
```
Pike, line 59: Exception, this might be an actual `Node` as this is more than one element. AFAICT.
Though, it seems it's an element, but I think in this case Node is more like it.

frontend/src/modules/otherlocales/components/OtherLocales.js
```diff
@@ -1,5 +1,6 @@
 /* @flow */
 
+import type { Node, Element } from 'React';
 import React from 'react';
```
Pike, line 4: This file just needs types, convert this import to `import type * as React`?

frontend/src/modules/otherlocales/components/Translation.js
```diff
@@ -1,5 +1,6 @@
 /* @flow */
 
+import type { Node } from 'React';
 import React from 'react';
```
Pike, line 4: I'd use a full blown `import * as React`

Pike (changes requested):
This is really nice, and I think it's going to make us get closer to TS.

I've got a general concern about the over use (IMHO) of `React.Node`. I think those places are gonna hurt in a post-flow world. Thus I'd prefer `React.Element<React.ElementType>` instead, unless there's an obvious candidate. I'd make the opposite call in the places where there's actually a fragment of content we return. I think the definition of `React.Node` matches that well.

I kept on annotating that for quite a while, but gave up in the middle. I only kept adding when I found a concrete inner part for `Element`.

If you'd like help with this, lmk. I had most edits in my local tree at some point (but also threw most of them away again) to test my claims.

Pike (approved):
This looks good to me, with my fixes reviewed by April.

mathjazz (approved):
This is beautiful, thank you so much!

FWIW, I've deployed the branch to stage and it seems to be working fine.

PR #1891, opened 2021-03-12
Bug 1698074, run flow-to-ts in automation to keep track of errors

Pike:


Pike:
I think this is a good-enough starting point. I haven't yet figured out how to get an error count, but right now this is at 767 TS errors.

frontend/no-tsconfig.json
```diff
@@ -0,0 +1,72 @@
...
+      // "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
+      // "emitDecoratorMetadata": true,         /* Enables experimental support for emitting type metadata for decorators. */
+  
+      /* Advanced Options */
+      "skipLibCheck": true,                     /* Skip type checking of declaration files. */
+      "forceConsistentCasingInFileNames": true  /* Disallow inconsistently-cased references to the same file. */
+    },
+    "include": ["src"]
+  }
+
```
mathjazz, line 72: I guess the empty line is not empty?

mathjazz (commented):
Thanks!

Why do you need to `git commit` the changes before running the script? And why do you need to run `git diff`?

Also, a bug explaining the motivation or a comment in [bug 1685565](https://bugzilla.mozilla.org/show_bug.cgi?id=1685565) would be welcome.

Pike:
> Why do you need to `git commit` the changes before running the script? And why do you need to run `git diff` (to debug?)?

I want to add visibility into what the steps do, and what the code looks like that is creating errors. The commit and diff on the first step is just for the sake of completeness. I really care about exposing the changes that `flow-to-ts` does, but sadly git didn't cooperate with me at all so far. That's what that's commented out.

Maybe I'll use https://git-scm.com/docs/git-bundle, but I didn't want to spend too much time on this workflow for now, and rather iterate. I found it more interesting to get the flow up, so that we can use it in April's PR.

mathjazz (approved):
Thanks for clarifying!

PR #1894, opened 2021-03-13
Fix bug 1693319: Reduce number of links in comment notifications

VishalCR7:


VishalCR7:
@mathjazz please review :)

pontoon/base/static/css/style.css
```diff
@@ -821,11 +822,20 @@ header .right .select .menu {
...
     white-space: nowrap;
 }
 
+.notifications .menu li.notification-item .message.trim a {
+    color: #aaaaaa;
+}
+
+.notifications .menu li.notification-item.hover .message.trim a {
+    color: #ebebeb;
+}
```
mathjazz, line 837: We should have the same rule in the Translate view CSS file.

mathjazz (commented):
Great work!

I'm thinking about a possible improvement: how about we make the entire notification block (`li` element) clickable now that we only have one link?

I saw that in several popular apps, and it's quite a comfy user experience to not be needing to catch links.

Thoughts?

pontoon/base/static/css/style.css
VishalCR7: The description on hover turns from #aaaaaa to #ebebeb on the normal notification bar; hence this rule needs to be added. 

Change of color does not happen on hovering in general for Translate view CSS so the rule is not needed.

mathjazz:
> I'm thinking about a possible improvement: how about we make the entire notification block (`li` element) clickable now that we only have one link?

Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1700303.

pontoon/base/static/css/style.css
mathjazz: Behavior is inconsistent, though. If you have a comment notification in the Translate view, comment text doesn't become lighter, whereas in the /notifications page and in the notifications menu in a non-translate view, it does.

Styling should be consistent.

frontend/src/core/user/components/UserNotification.js
```diff
@@ -39,6 +39,9 @@ export default class UserNotification extends React.Component<Props, State> {
         const { notification } = this.props;
 
         let className = 'user-notification';
+        let is_comment = !notification.description.content
```
mathjazz, line 42: We use camelCase in JS - snake_case is only used in python.

frontend/src/core/user/components/UserNotification.js
```diff
@@ -53,18 +56,34 @@ export default class UserNotification extends React.Component<Props, State> {
             >
                 <div className='item-content'>
                     <span className='actor'>
-                        <a href={notification.actor.url}>
-                            {notification.actor.anchor}
-                        </a>
+                        {is_comment ? (
```
mathjazz, line 59: I think this would be cleaner with only one, high-level `is_comment` check. Even though we'd be repeating slightly more code, I find it hard to follow it with three `is_comment` checks. And we should do the same in the Django templates.

mathjazz (commented):
Thanks for the update!

Left some suggestions to unify the experience and improve code readability.

frontend/src/core/user/components/UserNotification.js
```diff
@@ -39,12 +39,62 @@ export default class UserNotification extends React.Component<Props, State> {
...
+                        <span className='target'>
+                            {notification.target.anchor}
+                        </span>
+
+                        <ReactTimeAgo
+                            className='timeago'
+                            date={new Date(notification.date_iso)}
+                            title={`${notification.date} UTC`}
+                        />
+                        <div className='message trim'>
```
mathjazz, line 78: Nit: Add blank line above this line for consistency.

pontoon/contributors/templates/contributors/widgets/notifications_menu.html
```diff
@@ -36,41 +39,48 @@
           {% set actor_url = url('pontoon.contributors.contributor.username', notification.actor.username) %}
         {% endif %}
 
-        <span class="actor">
-          <a href="{{ actor_url }}">{{ actor_anchor }}</a>
-        </span>
+        {% if target and target.resource %}
```
mathjazz, line 42: Nit: blank line under this line would be consistent with the rest of the code.

mathjazz (commented):
Just a couple of nits and we're good!

mathjazz (approved):
Awesome!

PR #1896, opened 2021-03-15
Bug 1698074, follow-up for prettier and diff

Pike:
The problems I had with the diffs were a rather obscure limit of files that are allowed when diffing with moves. Bump that.

Also, the prettified version from flow-to-ts strips newlines, move the prettify call separate.

### DIFF
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -67,6 +67,8 @@ jobs:
       - name: Install globals
         run: |
           npm install -g yarn @khanacademy/flow-to-ts
+      - name: npm install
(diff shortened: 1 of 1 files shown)


Pike:
Got the diff to become useful. https://github.com/Khan/flow-to-ts/issues/237 is all over the place, as well as some newly-multiline statements, I think.

Adding ` --ignore-blank-lines` would reduce the size of the output, I guess. But it'd also make it harder to tell line numbers, so I'm not sure which is better.

mathjazz (approved):
LGTM.

Nice catch with the file limit!

PR #1897, opened 2021-03-17
manage.py: read dotenv files from alternative path

flozz:
Hello,

A very small PR that allows `manage.py` to read dotenv files from an alternative path defined by the `DOTENV_PATH` environment variable (like `wsgi.py` does).

This feature is already documented:

Capture d’écran de 2021-03-17 11-06-20

:)

### DIFF
--- a/docs/admin/deployment.rst
+++ b/docs/admin/deployment.rst
@@ -48,7 +48,7 @@ you create:
 .. NOTE::
 
    Alternatively, you can put all variables below in a `dotenv
-   <https://github.com/theskumar/python-dotenv>`_ text file::
+   <https://github.com/jpadilla/django-dotenv>`_ text file::
 
       VAR="value 1"
       OTHER_VAR="other value"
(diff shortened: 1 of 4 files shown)


mathjazz (commented):
Thanks!

Could you please make two corrections in the doc?
1. It links to the wrong library, we use https://github.com/jpadilla/django-dotenv.
2. `DOTENV_PATH` only needs to be passed optionally.

Also, could you please update to the latest version of https://github.com/jpadilla/django-dotenv?

flozz:
> It links to the wrong library, we use https://github.com/jpadilla/django-dotenv.

→ Fixed :)

> DOTENV_PATH only needs to be passed optionally.

→ For `manage.py` you are right, it will try to read a `.env` file in the current directory if this parameter is not available. But for `wsgi.py` this parameter is mandatory if you want to use a dotenv file: https://github.com/mozilla/pontoon/blob/master/pontoon/wsgi.py#L14-L15

I do not know how to change the doc to reflect this difference... 

> Also, could you please update to the latest version of https://github.com/jpadilla/django-dotenv?

→ Updated to v1.4.2

mathjazz (approved):
Thanks for the update!

And good point about the docs - these are deployment docs and `DOTENV_PATH` is required if you want to use a dotenv file in a production environment.

PR #1900, opened 2021-03-17
Prepare for typescript while still in flow

Pike:
* Prepare for typescript in core/api/base


Pike:
April, would you be up for a review on this one? `core/entities` is half-way, fluent is partly in fluent land and also a different PR.

Just starting chopping away at some flow type fixes that get typescript ahead.

abowler2 (approved):
These changes all look good to me.

PR #1901, opened 2021-03-18
Fix bug 1699381: make the media folder configurable

flozz:
This PR makes the path of the `media/` folder configurable through the `MEDIA_ROOT` environment variable.

* By default this is configured to a `media/` folder located next to the `pontoon` Python module (like the currently hardcoded value)
* The documentation was updated accordingly

This fixes the bug #1699381 → https://bugzilla.mozilla.org/show_bug.cgi?id=1699381

### DIFF
--- a/docs/admin/deployment.rst
+++ b/docs/admin/deployment.rst
@@ -145,6 +145,10 @@ you create:
 ``MANUAL_SYNC``
    Optional. Enable Sync button in project Admin.
 
+``MEDIA_ROOT``
+   Optional. The absolute path of the "media" folder where the projects will be
+   cloned (it located next to the "pontoon" Python module by default).
+
 ``MICROSOFT_TRANSLATOR_API_KEY``
    Optional. Set your `Microsoft Translator API key`_ to use machine translation
    by Microsoft.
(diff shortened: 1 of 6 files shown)


docs/admin/deployment.rst
```diff
@@ -145,6 +145,10 @@ you create:
 ``MANUAL_SYNC``
    Optional. Enable Sync button in project Admin.
 
+``MEDIA_ROOT``
+   Optional. The absolute path of the "media" folder where the projects will be
+   cloned (it located next to the "pontoon" Python module by default).
```
mathjazz, line 150: Nit: folder ~where~ the projects will be cloned **into** (it **is** located...

mathjazz (commented):
Thanks!

While we're at it... Could you please also remove the subfolders from the `media` folder? They are a relic of the past and I think we are safe to get rid of them.

docs/admin/deployment.rst
flozz: Thank you for the english check :)

mathjazz (approved):
Great work!

PR #1902, opened 2021-03-18
Type @fluent/syntax

Pike:
Very WIP.

Practically, I'm down to 2 flow errors, and passing tests.

Theoretically, I'm doing typing quite a bit different than it's done on the typescript side.

I remember conversations with @stasm about the use of base classes for Expressions etc and their incompleteness. Which I've done the other way around now so that refinements work.


frontend/src/core/utils/fluent/types.js
```diff
@@ -1,56 +1,12 @@
...
 export type FluentAttributes = Array<FluentAttribute>;
 
-export type FluentMessage = {
-    type: string,
-    value: Pattern,
-    attributes: ?FluentAttributes,
-    clone: () => FluentMessage,
-    equals: (any) => boolean,
-};
+export type FluentMessage = Entry;
```
Pike, line 9: I'm torn on whether to keep these type aliases or not.

In particular the mismatch between calling this a message when it actually is a term, or even a Junk entry, bothers me.

Pike (commented):
This is ready for review. I left an inline comment on `core/utils/fluent/types.js` with an open question.

The tests `expect and if not` pattern is because neither flow nor typescript narrow types on failed test assertions. Thus I went for test-and-return instead.

frontend/src/modules/fluenteditor/components/rich/RichTranslationForm.js
```diff
@@ -415,31 +422,31 @@ export default function RichTranslationForm(
         let indent = false;
 
         return elements.map((element, eIndex) => {
+            let expression;
             if (
                 element.type === 'Placeable' &&
-                element.expression &&
-                element.expression.type === 'SelectExpression'
+                (expression = element.expression) &&
```
abowler2, line 428: Can we just set `const expression = element.expression` above and then just check for `expression` here?

frontend/src/core/utils/fluent/types.js
abowler2: What would be the value of keeping these aliases vs just using the types directly from `@fluent/syntax`

frontend/src/modules/fluenteditor/components/rich/RichTranslationForm.js
Pike: Flow already doesn't like that, because `element` can be a `TextElement`, which doesn't have `.expression`. The order of narrowing matters here.

frontend/src/core/utils/fluent/flattenMessage.test.js
```diff
@@ -13,6 +18,10 @@ describe('flattenMessage', () => {
     it('does not modify attributes with single element', () => {
         const message = parser.parseEntry('title =\n    .foo = Bar');
         const res = flattenMessage(message);
+        expect(res.type).toEqual('Message');
+        if (res.type !== 'Message') {
```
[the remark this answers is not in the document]
Pike: The `expect` is to actually test the type, and the early `return` is to switch on type narrowing in TS for the code that follows. I could've just used returns w/out testing, but that felt even weirder.

What I'd really want is a pythonic `assert` that's supported by TS to narrow types, but I didn't find one.

frontend/src/core/utils/fluent/types.js
Pike: TBH, I've not asked anybody when they got introduced. I'm not even sure if that was TN or translate.old 🤷 

My guess is that it's namespaced as Fluent-specific, but I haven't found that to be much of a problem in the code I touched, it was usually very straight-forward that it's code handling Fluent.

Pike: We talked about this, and we should drop those aliases

frontend/src/core/utils/fluent/flattenMessage.test.js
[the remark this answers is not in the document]
Pike: Let's not add typescript to tests at this point, per today's meeting. I'll do an independent PR to exclude test files, and then remove the crutch here.

Pike:
Addressed the comments and discussions, TS errors are going from 137 to 105 now that we don't convert tests.

There's one in-test fix about the mocked bundles, which is one of those cases where having type checks avoids having junk in tests. Just adding that as a data point, I don't think this puts a dent in our decision to leave the tests in JavaScript.

frontend/src/modules/fluenteditor/components/FluentEditor.js
```diff
@@ -80,11 +82,7 @@ function useLoadTranslation(forceSource) {
             activeTranslationString || entity.original,
         );
 
-        if (syntax === '') {
```
abowler2, line 83: Why are these checks no longer needed? (there is the same check line#208)

Pike: `getSyntaxType` never returns an empty string, it's always one of the `SyntaxType` enum.

> This condition will always return 'false' since the types 'SyntaxType' and '""' have no overlap.

is what https://github.com/mozilla/pontoon/runs/2520885795?check_suite_focus=true#step:16:782 says on master right now.

mathjazz:
Flagging myself for review here to take it through my standard manual editor testing process.

frontend/src/core/utils/fluent/extractAccessKeyCandidates.js
```diff
@@ -36,12 +36,16 @@ function getTextElementsRecursivelly(elements: Array<PatternElement>) {
...
  *
- * @param {FluentMessage} message A (flat) Fluent message to extract access key candidates from.
+ * @param {Entry} message A (flat) Fluent message to extract access key candidates from.
  * @returns {?Array<string>} A list of access key candidates.
  */
 export default function extractAccessKeyCandidates(
-    message: FluentMessage,
+    message: Entry,
 ): ?Array<string> {
+    // Safe guard against non-message Fluent entries
```
mathjazz, line 45: Nit: I'd be consistent and spell "safeguard" as one word across the board.

mathjazz (approved):
I haven't spot any issues.

The code also looks good to my newbie TS self.

@abowler2 Did you want to have another look here?

abowler2:
@Pike and @mathjazz 

Took another look and all looks good to me on this one.

PR #1909, opened 2021-03-25
Update react-content-marker to 2.0.0

mathjazz:


Pike (approved):
This is good as is.

There's still more opportunities around type safety in the marker scene, we have a couple of `any` left. But that's also good for another day.

mathjazz:
Good point!

PR #1913, opened 2021-04-07
Update Python dependencies

mathjazz:
Result of `make requirements opts=--upgrade`, which I ran to update `urllib3`.


Pike (approved):
I did some release notes hunting, and only the asgiref one looks concerning. Though we're probably not using the stuff they're going back and forth for over last few days, https://github.com/django/asgiref/blob/main/CHANGELOG.txt.

PR #1914, opened 2021-04-08
Bug 1703748: Prevent sync breakage due to a typo

mathjazz:
This is a hotfix for an obvious error.

I'll keep the bug open to increase test coverage and catch the error early.


Pike (approved):
I'm concerned that there's a different bug behind this bug, but the affected data `new_locales` is only used in pretranslation, which isn't widely used.

One thing to check is that I *think* it's declaring a project to be brand-new for a locale if only one resource in it is new. That might be bad. Or just a waste of compute cycles.

PR #1920, opened 2021-04-12
Fix bug 1701632: Implement metrics for user testing

mathjazz:
This patch allows for collecting data about any UX action:
https://bugzilla.mozilla.org/show_bug.cgi?id=1701632

We might turn `experiment` into a `ForeignKey` or replace it with a cohort (which would imply experiment) as part of the AB testing capability:
https://bugzilla.mozilla.org/show_bug.cgi?id=1701631


Pike (commented):
I think there's a case to be made for non-experiment UX metrics, for example to get a baseline usage.

I'd make experiment optional, at least. Or maybe add it once we know what it's going to be? I think it's easier to add a new field later than to migrate an existing field.

mathjazz:
Thanks!

I'm keeping the `experiment` field in (and making it optional), because we already know we'll need to collect baseline metrics and then again the same metrics for the experiments. We're still undecided though if we'll run those experiments in parallel (AB testing) or in series. 

I'm also making the `data` field optional.

Pike (approved):

PR #1923, opened 2021-04-16
Fix bug 1697583 : Filter out invalid machinery translations

VishalCR7:
### Issue
If you perform a concordance search and submit a translation immediately after, the previous source string (the one that was just translated) gets used as an input for Machinery along with the current string when the next string opens. i.e. we actually make two requests, and results from both appear in machinery.

### Fix
Filter out the translations for which source != entity except for concordance search translations.


abowler2 (approved):
Looks good to me.

PR #1924, opened 2021-04-17
Bug 1704038: Update newrelic to 6.2.0.156

PFischbeck:
The previous version was incompatible with Django 3.2, see details in [my Bugzilla comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1704038#c3).

Since this is 3 years worth of updates, and I don't know how to test it locally, please thoroughly test in staging whether this behaves as expected. You can find all `newrelic` release notes [here](https://docs.newrelic.com/docs/release-notes/agent-release-notes/python-release-notes/python-agent-30089/) (3.0.0.89 through 6.2.0.156).
Additionally, 404 pages on Pontoon should not cause newrelic to throw an error itself :sweat_smile:


Pike:
I'm going to defer the review to the master of stages. I did read docs and release notes of newrelic wondering about this update, and there wasn't anything that sounded like requesting changes. I'm not sure if there's anything about the version of newrelic we use. I haven't found an indication there is, but also none that there isn't.

mathjazz:
Thanks for the update!

I've deployed the patch to stage about an hour ago and couldn't reproduce the errors I reported in the bug anymore. I also haven't seen any errors since the deployment.

I'll wait for a day before merging in case any other errors pop up.

mathjazz:
Everything looks good on stage for the last 24 hours. \o/

PR #1895, opened 2021-03-15
Bug 1697462, update node dependencies

Pike:
* Update babel and eslint

Drafting this PR for now, as I expect it to conflict with the update-flow PR on package-lock.json.

And once that's done, I'll add more commits for react-scripts and caniuse and such.

PR #1927, opened 2021-04-19
Update webpack

mathjazz:
Updating to the latest version of webpack as hinted at in https://github.com/mozilla/pontoon/pull/1895#issuecomment-801241351. Which also reduces a list of dependencies, some with potential security vulnerabilities.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1756,6 +1756,12 @@
         "minimist": "^1.2.0"
       }
     },
+    "@discoveryjs/json-ext": {
+      "version": "0.5.2",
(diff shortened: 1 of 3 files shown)


Pike (commented):
AFAICT, this is documented to not work :-(

The bundle tracker is a peer of django-webpack-loader, and the alpha version of the js sibling doesn't have a python equiv. And likely won't, as both are now in search of new maintainers.

https://github.com/django-webpack/django-webpack-loader#alternatives-to-django-webpack-loader might be worth considering. Or more drastic actions.

mathjazz:
Thanks for looking into it. Seems like something I could pick at a better time. ;)

PR #1929, opened 2021-04-24
Bug 1704038: Make use of new admin.display decorator

PFischbeck:
This makes use of the new admin.display decorator from Django 3.2. Additionally, we use `format_html` instead of `mark_safe`.

### DIFF
--- a/pontoon/checks/admin.py
+++ b/pontoon/checks/admin.py
@@ -15,15 +15,19 @@ class FailedCheckModelAdmin(admin.ModelAdmin):
         "resource",
     )
 
+    @admin.display
     def project(self, obj):
         return obj.translation.entity.resource.project.name
 
+    @admin.display
     def locale(self, obj):
         return obj.translation.locale.code
 
+    @admin.display
     def resource(self, obj):
         return obj.translation.entity.resource.path
 
+    @admin.display
     def translation(self, obj):
         return obj.translation.string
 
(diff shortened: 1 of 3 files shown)


pontoon/checks/admin.py
```diff
@@ -15,15 +15,19 @@ class FailedCheckModelAdmin(admin.ModelAdmin):
         "resource",
```
mathjazz, line 15: We only show translation in the list view, I wonder why is that.

Note that we have the same problem on master.

mathjazz (approved):
Nice work, thanks!

Left a question regarding the checks admin. Other that that, this look good.

pontoon/checks/admin.py
mathjazz: @PFischbeck Hey, just curious if you'd like to have a look into this.

No pressure, we can also fix it at a later occasion.

PFischbeck: @mathjazz I'm not sure what you mean. I can see the other attributes in the list view too:
grafik
And in the edit view, the translation is enough to infer all other attributes:
grafik
Or do you mean we should list other translation attributes, like the translation author?

mathjazz: Interesting, that's how the list looks for me:

Posnetek zaslona 2021-05-04 ob 14 48 20

PFischbeck: That's weird. Are the other columns present in the HTML? Maybe this is only caused by translations containing special characters (newlines etc.).
Or maybe someone else can try reproducing?

mathjazz: It turns out some translations are very long and they break the layout. So one just needs to scroll horizontally and the fields get revealed. :)

Sorry for the noise, this is outside the scope of this PR anyways.
https://bugzilla.mozilla.org/show_bug.cgi?id=1709788

PR #1931, opened 2021-04-27
Fix bug 1706936: Collect baseline notifications metrics

mathjazz:
This patch introduces collection of user interaction metrics of the existing notifications UI.

It also includes the following improvements:
* CSRF token is now globally accessible on all pages using the base template (https://github.com/mozilla/pontoon/pull/1931/commits/21aad3d05234269f17c1c144e1f5bea88f2f7a15), which allows us to collect metrics on all pages that contain the notification icon/menu.
* Click-area to open the notification menu and to mark notifications as read is now exactly the same (https://github.com/mozilla/pontoon/pull/1931/commits/40b218874fe6cd1ce0f1ff21af1a7245216955b8). That is currently not the case, which can result in notifications not being marked as read in some edge cases, even if the menu is opened.
* Notification animation times in the old and new frontend code have been unified (https://github.com/mozilla/pontoon/pull/1931/commits/8c3be510d0a128698a0deb97d99e75d03ce9401d).

### DIFF
--- a/frontend/src/core/api/user.js
+++ b/frontend/src/core/api/user.js
@@ -29,6 +29,35 @@ export default class UserAPI extends APIBase {
     }
 
     /**
+     * Log UX action.
+     */
+    async logUxAction(
+        action_type: string,
+        experiment: ?string,
+        data: ?any,
+    ): Promise<any> {
+        const csrfToken = this.getCSRFToken();
+
+        const payload = new URLSearchParams();
+        payload.append('csrfmiddlewaretoken', csrfToken);
+        payload.append('action_type', action_type);
+
+        if (experiment) {
+            payload.append('experiment', experiment);
+        }
+
+        if (data) {
+            payload.append('data', JSON.stringify(data));
+        }
+
+        const headers = new Headers();
+        headers.append('X-Requested-With', 'XMLHttpRequest');
+        headers.append('X-CSRFToken', csrfToken);
+
+        return await this.fetch('/log-ux-action/', 'POST', payload, headers);
+    }
+
+    /**
      * Mark all notifications of the current user as read.
      */
     async markAllNotificationsAsRead(): Promise<any> {
--- a/frontend/src/core/api/user.js
+++ b/frontend/src/core/api/user.js
@@ -31,7 +31,7 @@ export default class UserAPI extends APIBase {
     /**
      * Log UX action.
      */
-    async logUxAction(
+    logUxAction(
         action_type: string,
         experiment: ?string,
         data: ?any,
@@ -54,7 +54,7 @@ export default class UserAPI extends APIBase {
         headers.append('X-Requested-With', 'XMLHttpRequest');
         headers.append('X-CSRFToken', csrfToken);
 
-        return await this.fetch('/log-ux-action/', 'POST', payload, headers);
+        return this.fetch('/log-ux-action/', 'POST', payload, headers);
     }
 
     /**
--- a/frontend/src/core/user/components/UserNotificationsMenu.js
+++ b/frontend/src/core/user/components/UserNotificationsMenu.js
@@ -23,26 +23,16 @@ type State = {|
 |};
 
 type UserNotificationsMenuProps = {
-    has_unread: boolean,
-    logUxAction: (string, ?string, ?any) => void,
     notifications: Array<Notification>,
     onDiscard: (e: SyntheticEvent<any>) => void,
 };
 
 export function UserNotificationsMenu({
-    has_unread,
-    logUxAction,
     notifications,
     onDiscard,
 }: UserNotificationsMenuProps): React.Element<'div'> {
     const ref = React.useRef(null);
     useOnDiscard(ref, onDiscard);
-
-    function handleClickSeeAll() {
-        logUxAction('See all Notifications link clicked', 'Notifications 1.0', {
-            unread: has_unread,
-        });
-    }
 
     return (
         <div ref={ref} className='menu'>
@@ -72,9 +62,11 @@ export function UserNotificationsMenu({
                 )}
             </ul>
 
-            <div className='see-all' onClick={handleClickSeeAll}>
+            <div className='see-all'>
                 <Localized id='user-UserNotificationsMenu--see-all-notifications'>
-                    <a href='/notifications'>See all Notifications</a>
+                    <a href='/notifications?referrer=ui'>
+                        See all Notifications
+                    </a>
                 </Localized>
             </div>
         </div>
@@ -130,7 +122,7 @@ export default class UserNotificationsMenuBase extends React.Component<
         }
     }
 
-    handleClickIcon: () => void = () => {
+    handleClick: () => void = () => {
         if (this.state.markAsRead) {
             this.setState({
                 markAsRead: false,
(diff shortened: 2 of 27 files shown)


frontend/src/core/user/components/UserNotificationsMenu.js
```diff
@@ -22,17 +23,27 @@ type State = {|
...
     notifications,
     onDiscard,
 }: UserNotificationsMenuProps): React.Element<'div'> {
     const ref = React.useRef(null);
     useOnDiscard(ref, onDiscard);
 
+    function handleClickSeeAll() {
+        logUxAction('See all Notifications link clicked', 'Notifications 1.0', {
+            unread: has_unread,
+        });
```
mathjazz, line 44: This needs tweaking (also on the old frontend): right now, the request could get interrupted by redirecting to `/notifications`, in which case the action will not be logged.

pontoon/contributors/views.py
```diff
@@ -278,6 +279,13 @@ def mark_all_notifications_as_read(request):
     """Mark all notifications of the currently logged in user as read"""
     request.user.notifications.mark_all_as_read()
 
+    utm_source = request.GET.get("utm_source")
```
Pike, line 282: I think we should collect this ux action on all calls, and add `utm_source` as data.

frontend/src/core/api/user.js
```diff
@@ -28,6 +28,35 @@ export default class UserAPI extends APIBase {
         return await this.fetch('get-users', 'GET', null, headers);
     }
 
+    /**
+     * Log UX action.
+     */
+    async logUxAction(
```
Pike, line 34: I don't think this needs to be async, but it can just be a sync function returning a promise. The `return await` is the only await, and it seems unnecessary.

pontoon/uxactionlog/views.py
```diff
@@ -0,0 +1,35 @@
...
+@require_AJAX
+@transaction.atomic
+def log_ux_action(request):
+    """Save a new UX action in the database."""
+    form = forms.UXActionLogForm(request.POST)
+
+    if not form.is_valid():
+        return JsonResponse(
+            {
+                "status": False,
```
Pike, line 21: The `status` part of the response is a duplicate of the http status, I'd just choose one of the two.

Pike (changes requested):
There doesn't seem to be a naming convention for ux action types yet? I'd argue we should have one.

I'm also wondering if we should document the ux action metrics and experiments?

I haven't really reviewed the frontend code, but I am wondering if there's a better paradigm to use for `logUxAction`. 

In the old UI and python parts, I'm mostly OK. One thing that I noticed is that nothing checks the responses the uxlog view sends?

pontoon/uxactionlog/views.py
mathjazz: True... but we do this in pretty much every `JsonResponse` call, so I'd rather not break this "convention" partially.

mathjazz:
> There doesn't seem to be a naming convention for ux action types yet? I'd argue we should have one.
> 
> I'm also wondering if we should document the ux action metrics and experiments?

Correct and agreed. We should include the convention in the docs. I'm wondering where to put the docs. Ideas for naming convention also welcome!

> I haven't really reviewed the frontend code, but I am wondering if there's a better paradigm to use for `logUxAction`.

Could you elaborate on that?

> In the old UI and python parts, I'm mostly OK. One thing that I noticed is that nothing checks the responses the uxlog view sends?

True. What should we do on error/success?

frontend/src/core/user/components/UserNotificationsMenu.js
Pike: How about just adding a flag to the link on the buttons, and do the ux action when loading the all-notifications page? That way, you don't need the UX state (I guess it'd need to be state), and we'd have one data set for people using UI controls vs just loading the page from history, too.

I'd suggest to unset the flag on load of the notifications page, though, so that reloads get flagged as just loads, and not UI control actions.

Pike:
> > I'm also wondering if we should document the ux action metrics and experiments?
> 
> Correct and agreed. We should include the convention in the docs. I'm wondering where to put the docs. Ideas for naming convention also welcome!

As for the convention, I'd go for something with an ID feel rather than free form text. But that's just a feeling, really.

For documentation, maybe even open a new top-level category? One page with all the metrics, and one page per experiment or so?

> 
> > I haven't really reviewed the frontend code, but I am wondering if there's a better paradigm to use for `logUxAction`.
> 
> Could you elaborate on that?

I'm looking at the ux actions as side effects, and I wonder if they're implemented as state transitions. But this is a blind person painting.



> True. What should we do on error/success?

Maybe at least `.catch(reason => {})` ? No idea if this should be done even on the API layer, as there shouldn't be a reason for the UI to handle those any differently anywhere, right?

mathjazz:
Moved See all Notifications link clicks collection to the backend. I'm only collecting the ones with the referrer (not sure I can remove GET parameters in Django without a redirect).

I've also filed a [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1708927) to add docs.

Pike:
> Moved See all Notifications link clicks collection to the backend. I'm only collecting the ones with the referrer (not sure I can remove GET parameters in Django without a redirect).

My thinking was to do that onload in javascript. Either `document.location` or `window.location` allow to change query params w-out reloading the page.

mathjazz:
I didn't like that, because it flashes the query parameter in the URL. OTOH it allows us to log all Notifications page views. I'm torn.

mathjazz:
> > > There doesn't seem to be a naming convention for ux action types yet? I'd argue we should have one.
> > > 
> > > I'm also wondering if we should document the ux action metrics and experiments?
> > 
> > 
> > Correct and agreed. We should include the convention in the docs. I'm wondering where to put the docs. Ideas for naming convention also welcome!
> 
> As for the convention, I'd go for something with an ID feel rather than free form text. But that's just a feeling, really.
> 
> For documentation, maybe even open a new top-level category? One page with all the metrics, and one page per experiment or so?

I wonder if @flodolo has ideas for naming conventions. Right now we log the following actions:
* Unread notifications icon displayed
* Notifications icon clicked
* mark_all_notifications_as_read (separately tracked for add-on and in-app actions)
* Notifications page loaded (separately tracked for clicks from the Notifications menu and direct pageloads)

The experiment name for the baseline metrics is "Notifications 1.0".

flodolo:
> I wonder if @flodolo has more concrete ideas for naming convention. Right now we log actions with the following names:
> 
>     * "Unread notifications icon displayed"
> 
>     * "Notifications icon clicked"
> 
>     * "mark_all_notifications_as_read" (separately tracked for add-on and in-app actions)
> 
>     * "Notifications page loaded" (separately tracked for clicks from the Notifications menu and direct pageloads)

What about starting with the action? That's the only thing I could think of, the third is a bit tricky though
- Display: unread notification icon
- Click: notifications icon
- Action: mark notifications as read
- Load: notifications page

mathjazz:
I've updated actions names as suggested by @flodolo and tweaked them slightly to make them more specific:

* Render: Unread notifications icon
* Click: Notifications icon
* Background action: Mark all notifications as read
* Page load: Notifications

I'm merging this as we'd like to ship.

Thanks for the review and testing, Axel and Flod!

PR #1932, opened 2021-04-29
Fix bug 1706959: Promote Pontoon Add-On in Pontoon

mathjazz:
Patch is deployed to stage for testing:

Note that there are two separate frontend implementations due to different codebases being in use - one is used on the translate app (`/frontend` folder) and the other one on other pages (`/pontoon` folder).


pontoon/base/static/js/main.js
```diff
@@ -143,6 +143,32 @@ $(function () {
...
+                downloadLink.attr('href', 'https://addons.mozilla.org/sl/firefox/addon/pontoon-tools/')
+            }
+            if (isChrome) {
+                downloadLink.attr('href', 'https://chrome.google.com/webstore/detail/pontoon-add-on/gnbfbnpjncpghhjmmhklfhcglbopagbb')
+            }
+            $('body').addClass('addon-promotion-active');
+        }
+    // window.PontoonAddon is made available by the Pontoon Add-On,
+    // but not immediatelly after the DOM is ready
+    }, 1000);
```
mathjazz, line 170: @MikkCZ Is this expected?

MikkCZ: Not sure, probably. The extension injects it's script when the tab loading is completed, and it's all async, so who actually know when that happens.

I will try to change the way the script is injected to a little more abstract API, that would allow to inject the script earlier.

MikkCZ: Can you please try with the artifact from _.github/workflows/build.yml_ action of https://github.com/MikkCZ/pontoon-addon/pull/142? It should be faster.

pontoon/base/static/js/main.js
```diff
@@ -143,6 +143,38 @@ $(function () {
...
+    var downloadLink = $('#addon-promotion').find('.get');
+    setTimeout(function () {
+        if (!window.PontoonAddon || !window.PontoonAddon.installed) {
+            if (!isFirefox && !isChrome) {
+                return;
+            }
+            if (isFirefox) {
+                downloadLink.attr(
+                    'href',
+                    'https://addons.mozilla.org/sl/firefox/addon/pontoon-tools/',
```
MikkCZ, line 163: Should be locale independent link `https://addons.mozilla.org/firefox/addon/pontoon-tools/`.

pontoon/base/static/js/main.js
```diff
@@ -143,6 +143,38 @@ $(function () {
...
+     *
+     * - Pontoon Add-On not installed
+     * - Page loaded on Firefox or Chrome (add-on not available for other browsers)
+     */
+    var isFirefox = navigator.userAgent.indexOf('Firefox') !== -1;
+    var isChrome = navigator.userAgent.indexOf('Chrome') !== -1;
+    var downloadLink = $('#addon-promotion').find('.get');
+    setTimeout(function () {
+        if (!window.PontoonAddon || !window.PontoonAddon.installed) {
+            if (!isFirefox && !isChrome) {
```
MikkCZ, line 157: Personally I would prefer the conditions like this
```js
let downloadLinkHref;
if (isFirefox) {
  downloadLinkHref = '...';
} else if (isChrome) {
  downloadLinkHref = '...';
}
if (downloadHref) {
  downloadLink.attr(...);
  $('body').addClass('addon-promotion-active');
}
```
Adding more browsers will be easier like this in the future.

pontoon/base/static/js/main.js
```diff
@@ -143,6 +143,38 @@ $(function () {
...
+     * Display Pontoon Add-On promotion, if:
+     *
+     * - Pontoon Add-On not installed
+     * - Page loaded on Firefox or Chrome (add-on not available for other browsers)
+     */
+    var isFirefox = navigator.userAgent.indexOf('Firefox') !== -1;
+    var isChrome = navigator.userAgent.indexOf('Chrome') !== -1;
+    var downloadLink = $('#addon-promotion').find('.get');
+    setTimeout(function () {
+        if (!window.PontoonAddon || !window.PontoonAddon.installed) {
```
MikkCZ, line 156: You also need to react on when the extension is installed after the page is loaded, something like
```js
window.addEventListener('message', (e) => {
  const data = JSON.parse(e.data);
  if (data._type === 'PontoonAddonInfo') {
    const pontoonAddonInfo = data.value;
    if (pontoonAddonInfo.installed) {
      $('body').removeClass('addon-promotion-active');
    }
  }
});
```

MikkCZ (changes requested):

pontoon/base/static/js/main.js
mathjazz: I tried that and couldn't access `window.PontoonAddon` from console.

pontoon/base/static/js/main.js
mathjazz: Ugh, nice catch!

pontoon/base/static/js/main.js
mathjazz: Happy to learn you're willing to add more browsers! 😆

pontoon/base/static/js/main.js
mathjazz: Yup, that's on my TODO list.

pontoon/base/static/js/main.js
MikkCZ: I found a bug, the Pontoon URL in the options must not end with trailing slash. I submitted a hotfix for that behaviour. I would like to use some library with validation for that eventually.

mathjazz: I wasn't using a trailing slash, I used http://localhost:8000, so the new version still doesn't work.

mathjazz: Don't be sorry, please. :) Which API? contentScripts?

MikkCZ: So, I hope I found it this time. See the pontoon-addon PR. This version I tested with some arbitrary page running on localhost, and it should work. I am not sure about the speed of the `window` object modification, but it can probably be any faster now.

mathjazz: Thanks, that works now. I still need to wait for the modification to happen, though.

That said, it seems like I can drop the `setTimeout` anyways now that I have `postMessage` listener in place. At least locally, the promotional banner does not show up, even if the criteria to display it is met - because the message is sent so quickly it closes the banner.

I'll deploy to stage and see how it works there.

MikkCZ: I did one last change to inject the script into `<head>`, so it will hopefully get executed before any frontend JS.

mathjazz: I don't see any difference, but there's definitely an improvement over the version that's currently released. The banner appears briefly with that one.

So if you ship this version today and we start the promotion on Monday... I wonder if we should keep the setTimeout in or do we expect the new version to hit users before Monday.

MikkCZ: It's Friday and there is a fair chance that people won't turn on their computer until Monday again. Anyway I would keep the timeout there, maybe not a full second, to avoid any blicking. That's never good and it would be happening with every page load.

mathjazz: Agreed! Thanks for your help!

mathjazz:
@MikkCZ Since you've already started looking, I wonder if you're up for a review of this patch.

frontend/src/core/api/user.js
```diff
@@ -102,4 +102,14 @@ export default class UserAPI extends APIBase {
             headers,
         );
     }
+
+    /**
+     * Dismiss Add-On Promotion.
+     */
+    dismissAddonPromotion(): Promise<any> {
+        const headers = new Headers();
+        headers.append('X-Requested-With', 'XMLHttpRequest');
```
MikkCZ, line 111: Wouldn't be easier to disable that "security by obscurity" feature in Django?

frontend/src/core/api/user.js
```diff
@@ -102,4 +102,14 @@ export default class UserAPI extends APIBase {
...
+
+        return this.fetch('/dismiss-addon-promotion/', 'GET', null, headers);
```
MikkCZ, line 113: For anything REST-like, this request should definitely be `POST` or even better `PATCH`, and should go to the same endpoint the information is retrieved from. But if we're not building an API here, I think that's fine.

MikkCZ: One idea, should Pontoon Add-on itself call this endpoint upon installation or regularly with fetching new data to dismiss the promotion even before the user visits any page?

If that would be nice to do, I would advise not to call it `/dismiss-addon-promotion/`, but rather `/pontoon-addon-installed/` and POST the JSON object equivalent to what's now injected into the `window` there.

frontend/src/modules/addonpromotion/components/AddonPromotion.js
```diff
@@ -0,0 +1,139 @@
...
+        };
+    }
+
+    componentDidMount() {
+        window.addEventListener('message', this.handleMessages);
+    }
+
+    componentWillUnmount() {
+        window.removeEventListener('message', this.handleMessages);
+    }
```
MikkCZ, line 43: If there are implemented with `useEffect`, this can be a nice functional component. ;)

frontend/src/modules/addonpromotion/index.js
```diff
@@ -0,0 +1,7 @@
+/* @flow */
+
+export { default as AddonPromotion } from './components/AddonPromotion';
+
+// Name of this module.
+// Used as the key to store this module's reducer.
+export const NAME: string = 'addonpromotiom';
```
MikkCZ, line 7: Here is a typo, it should be `addonpromotion`.

pontoon/base/static/js/main.js
```diff
@@ -143,6 +143,62 @@ $(function () {
...
+    /*
+     * Display Pontoon Add-On Promotion, if:
+     *
+     * - Promotion not dismissed
+     * - Add-On not installed
+     * - Page loaded on Firefox or Chrome (add-on not available for other browsers)
+     */
+    setTimeout(function () {
+        var notDismissed = $('#addon-promotion').length;
+        var notInstalled =
```
MikkCZ, line 155: Personally I would prefer `var installed = window.PontoonAddon && window.PontoonAddon.installed`, which is a little easier to read.

MikkCZ:
> @MikkCZ Since you've already started looking, I wonder if you're up for a review of this patch.

Thank you for your confidence in me. I added my few comments, however they may not be that relevant, because I am not familiar with Pontoon codebase nor the expected code style. I don't see why this wouldn't work, but another pair of eyes will be better for this review.

frontend/src/modules/addonpromotion/components/AddonPromotion.css
```diff
@@ -0,0 +1,50 @@
...
+    height: 44px;
+    position: fixed;
+    top: 0;
+    width: 100%;
+    z-index: 20;
+}
+
+.addon-promotion .container {
+    align-items: center;
+    background: #ff3366dd;
```
MikkCZ, line 17: This color (associated with error state) might be a littler too much drawing attention. But it's probably best from the state palette in terms of contrast.

mathjazz:
All good points, Michal!

Maybe @Pike and @abowler2 have cycles over the weekend to have a look? :-)

frontend/src/modules/addonpromotion/components/AddonPromotion.css
mathjazz: Yup, this is already down tuned a little bit from the early feedback. I'll make it slightly more transparent, i.e. dark.

pontoon/base/static/js/main.js
mathjazz: Good call. I should probably also `notDismissed` -> `dismissed`.

frontend/src/core/api/user.js
mathjazz: Our API story is a joke. :) I should make that request POST at least, I'd leave the rest for a followup.

Regarding the Add-on calling into that API I'm not sure. We have `postMessage` in place that effectively serves the same purpose, and use the `dismissed` flag to track which user has explicitly dismissed the promotion. Perhaps we should keep these actions separate.

PR #1938, opened 2021-05-07
Use native events for addEventListener

Pike:
This is another cleanup for TS, but also fixes some flow-ignores.

The gist is that the Synthetic events are only if you use the handler as jsx prop, but when you do `addEventHandler` (and `remove`), you want to use native event types.


Pike:
137 -> 126 (I wanna see 2 digit numbers!)

Pike:
Sorry for the ride-along, 124 was too tempting.

mathjazz (approved):
Nice job with bringing that number down and figuring out the difference between Synthetic and Native events!

PR #1942, opened 2021-05-07
Bug 1685565: various typing fixes related to React

Pike:


Pike:
Matjaz, it'd be great if you could drive this through your testing regime, not that my type narrowing broke stuff.

abowler2:
Took a look through this and all looks good to me.

PR #1943, opened 2021-05-08
Bug 1685565: Improve types for Slate and TypeScript

Pike:


Pike:
We'll fix Slate after the conversion to TS.

PR #1944, opened 2021-05-10
Bug 1685565, various typescript fixes

Pike:
* Don't mix synthetic and plain dom events in onDiscard hook by hiding the plain click event.


mathjazz (approved):
Nice work! I also tested manually and didn't spot any issues.

PR #1945, opened 2021-05-14
Bug 1685565: TypeScript fixes and expected errors

Pike:


frontend/no-tsconfig.json
```diff
@@ -4,7 +4,7 @@
 
     /* Basic Options */
     // "incremental": true,                   /* Enable incremental compilation */
-    "target": "es5",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */
+    "target": "es2015",                       /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */
```
Pike, line 7: I don't think this is going to affect our browser compat, looking at https://webhint.io/docs/user-guide/hints/hint-typescript-config/target/.

This is required to allow for array spreads on iterables.

frontend/src/modules/addonpromotion/components/AddonPromotion.js
```diff
@@ -68,6 +68,7 @@ export class AddonPromotionBase extends React.Component<InternalProps, State> {
         // Add-On installed
         if (
             this.state.installed ||
+            // @ts-expect-error
```
Pike, line 71: I haven't figured out how to amend the global type in TypeScript, let alone doing so in a way that wouldn't surprise flow.

frontend/src/modules/batchactions/actions.js
```diff
@@ -110,8 +110,12 @@ export function performAction(
...
         );
 
-        const response: ResponseType = {};
-        response.action = action;
+        const response: ResponseType = {
+            changedCount: 0,
+            invalidCount: 0,
+            error: false,
+            action,
+        };
```
Pike, line 118: Meh, my previous attempt to fix this didn't work, so I'm just starting off with a full object literal.

The alternative would be to do a bunch of `let`, and then create the response as `{action, changedCount, invalidCount, error}`. But I feel that's harder to read than this?

frontend/src/modules/history/actions.js
```diff
@@ -11,18 +11,38 @@ import { actions as pluralActions } from 'core/plural';
...
+    +rejected: boolean,
+    +string: string,
+    +uid: ?number,
+    +unapprovedUser: string,
+    +machinerySources: string,
+    +user: string,
+    +username: string,
+    +userGravatarUrlSmall: string,
+    +comments: Array<TranslationComment>,
+|};
```
Pike, line 37: I moved this type from `reducer` to avoid circular imports. And it seems that this is the better way looking at other modules.

But this is mostly me guessing.

frontend/src/modules/otherlocales/components/Translation.js
```diff
@@ -113,7 +114,7 @@ export default function Translation(
                 <p
                     lang={translation.locale.code}
                     dir={translation.locale.direction}
-                    script={translation.locale.script}
+                    data-script={translation.locale.script}
```
Pike, line 117: I've tried to find if this is used anywhere, but I couldn't find it. I did find `data-script` selectors in https://github.com/mozilla/pontoon/blob/master/frontend/src/modules/entitieslist/components/Entity.css#L48-L53, but no corresponding CSS in `otherlocales`. So maybe the way we render them isn't great?

frontend/src/modules/search/components/SearchBox.js
```diff
@@ -239,6 +239,7 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
             return this.setState({ timeRange });
         }
 
+        // @ts-expect-error
```
Pike, line 242: The variables on object keys are making TypeScript unhappy. I tried enum types a little bit on the keys, but that didn't work great, at least not in `TimeRangeFilter` below.

frontend/src/modules/search/components/TimeRangeFilter.js
```diff
@@ -387,6 +388,7 @@ export default class TimeRangeFilterBase extends React.Component<Props, State> {
                     {!this.state.visible ? null : (
                         <HighchartsReact
                             highcharts={Highcharts}
+                            // @ts-expect-error
```
Pike, line 391: highcharts comes with types, probably best to do as a follow-up. If at all.

frontend/src/modules/terms/components/Terms.js
```diff
@@ -27,19 +27,19 @@ export default function Terms(props: Props): null | React.Element<'section'> {
         return null;
     }
 
-    terms = terms.terms;
```
Pike, line 30: This redefined the type of `terms`. splitting this out.

Pike (commented):
Some comments on the changes for context

frontend/src/modules/otherlocales/components/Translation.js
mathjazz: Nice catch! There's also https://github.com/mozilla/pontoon/blob/master/frontend/src/index.css#L41, which will now also impact `otherlocales`.

mathjazz (approved):
Thanks. Changes look good.

PR #1946, opened 2021-05-17
Fix bug 1707819 by reverting fix for bug 1697583

mathjazz:
This reverts commit 9e16ba387706aeb7cf837973abca07c79a235f5a.

Bug 1697583 happens less often than 1707819, so let's return the problem solving to the original cause.


Pike (approved):
Based on our conversation, I think this is the right first step.

PR #1947, opened 2021-05-17
Fix bug 1710384: Prevent batch editor layout breakage

mathjazz:
By preventing a class name clash.

### DIFF
--- a/frontend/src/App.css
+++ b/frontend/src/App.css
@@ -14,10 +14,6 @@
     position: relative;
 }
 
-#app > header .user-controls {
-    float: right;
-}
-
 #app > .main-content {
     display: flex;
     flex: 1;
--- a/frontend/src/App.css
+++ b/frontend/src/App.css
@@ -37,11 +37,6 @@
     width: 75%;
 }
 
-#app .inner {
-    display: table-cell;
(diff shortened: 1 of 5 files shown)


Pike:
Wondering how this pieces together, is this https://github.com/mozilla/pontoon/pull/1932/files#diff-a335e6772310196bfa0c52a745f82a727e22185ac26025ee94a46364f7fe09aaR21 hitting more than one `.main-content`?

If so, any idea on how we could be systematic about preventing clashes in css selectors?

mathjazz:
> Wondering how this pieces together, is this https://github.com/mozilla/pontoon/pull/1932/files#diff-a335e6772310196bfa0c52a745f82a727e22185ac26025ee94a46364f7fe09aaR21 hitting more than one `.main-content`?

Correct (I should have pointed that out).

> If so, any idea on how we could be systematic about preventing clashes in css selectors?

Maybe https://create-react-app.dev/docs/adding-a-css-modules-stylesheet can help.

I'm also making App.css selectors more specific (and moving two CSS selectors out of the file).

Pike (approved):
Thanks, this looks much better.

PR #1949, opened 2021-05-17
Bug 1685565: convert frontend production code to TypeScript

Pike:
I'm working around `withActionsDisabled` by adding a `@ts-nocheck` to the top of the file.

I tried that on the continuous integration, but that puts the import before the flag, and then the flag doesn't work.

Status quo right now: Frontend builds, eslints, prettiers, and runs and passes tests. I didn't actually run the frontend.


Pike:
Things I'm not sure about here:

`?string` -> `string | null | undefined` shows up a lot, and that's both very verbose and very vague.
Same for `Function`, which is more verbose in TS with `(...args: Array<any>) => any`.

It's not that Flow is less vague on these, it just doesn't yell at you in the same way that TS does. Probably still OK to do as follow-up?

I've also had to tweak the defaults for TS and eslint to get them to be green. I think they're easier to be done in follow-ups, too. Maybe some of them would help with the vague types above (Looking at `no-inferrable-types`, for example). On the TS front, I had to disable JS as our test files have some optional function params, which TS complains about.

Pike:
Linter stats for good measure:

check | count
------- | ------
`prefer-const` | 20
`no-var` | 1
`@typescript-eslint/ban-ts-comment` | 9
`@typescript-eslint/ban-types` | 2
`@typescript-eslint/explicit-module-boundary-types` | 2
`@typescript-eslint/no-empty-function` | 9
`@typescript-eslint/no-explicit-any` | 270
`@typescript-eslint/no-inferrable-types` | 28
`@typescript-eslint/prefer-as-const` | 61

Pike:
> On the TS front, I had to disable JS as our test files have some optional function params, which TS complains about.

I landed a fix for this one already.

frontend/src/core/api/types.ts
```diff
@@ -30,23 +30,6 @@ export type TranslationComment = {
...
-    type: string;
-    character: string;
-    url: string;
-    children: Array<TextType>;
-};
-export type InitialType = {
-    type: string;
-    children: Array<TextType>;
-};
-
```
mathjazz: 👏

Makefile
```diff
@@ -28,6 +28,7 @@ help:
 	@echo "  prettier         Runs the prettier formatter on the frontend code"
 	@echo "  check-prettier   Runs a check for format issues with the prettier formatter"
 	@echo "  format           Runs formatters for both the frontend and Python code"
+	@echo "  types            Runs the tsc compiler to check TypeScript on all frontend code"
```
mathjazz, line 31: You'll also need to replace flow with types in the `.PHONY` line above

frontend/README.md
```diff
@@ -88,9 +88,6 @@ A common case during development is to have 3 terminals open: one for the dev se
 
     # terminal 2
     $ make test-frontend
-
-    # terminal 3
-    $ make flow
```
mathjazz, line 93: Nice. There's one last mention of `flow` in the README that still needs to be removed.

mathjazz (commented):
I still need to do a bit more in-depth look, so just a few early comments for now:
- there are still 3 `FlowIgnore` occurences
- there's a reference to flow in `SimpleEditor.tsx`

abowler2 (commented):
Left a few questions/comments.

I agree that the way `Function` came across is a bit 'odd'. It would probably be worth going back over those as a follow up to try and be more explicit.

Thanks again for taking on this monumental task.

frontend/src/App.tsx
```diff
@@ -34,19 +32,18 @@ import type { NavigationParams } from 'core/navigation';
...
+    parameters: NavigationParams;
+    project: ProjectState;
+    stats: Stats;
+};
 
-type InternalProps = {
-    ...Props,
-    dispatch: Function,
+type InternalProps = Props & {
+    dispatch: (...args: Array<any>) => any;
```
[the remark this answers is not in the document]
Pike: Possibly, I'm scared to take this one as part of this PR. Filed a potential follow-up https://bugzilla.mozilla.org/show_bug.cgi?id=1712447 for now.

frontend/src/core/api/entity.ts
```diff
@@ -51,17 +49,17 @@ export default class EntityAPI extends APIBase {
...
-        pkOnly?: ?boolean,
-    ): Promise<Object> {
+        entity?: string | null | undefined,
+        search?: string | null | undefined,
+        status?: string | null | undefined,
+        extra?: string | null | undefined,
+        tag?: string | null | undefined,
+        author?: string | null | undefined,
+        time?: string | null | undefined,
+        pkOnly?: boolean | null | undefined,
```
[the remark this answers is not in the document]
Pike: There's a pattern where we pass parameters from function to function, and they're optional all around. And then the optional string becomes and optional string | undefined.

Right now, tsconfig disabled many of those checks, as strict is off, in particular strictNullChecks.

I wonder if a an options object would be a better coding pattern for these?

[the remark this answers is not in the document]
abowler2: I think looking at implementing an options object in these cases would be beneficial. 

For the sake of this PR,  I'm good with keeping these here and revisiting them if we ever decide to tackle the options object.

Pike:
I had another look at our undefined typings and fixed some easy ones. Sadly, the hardest one seems to be `AddComment`, which is hard for two reasons: It's everywhere w/out the parent components inheriting the ac props, so it's really hard to be consistent. The other reason is that navigation params in there are coming from `URLParams.get`, which returns `string | null`, so you can't just rely on `foo | undefined`.

I'd prefer to have those problems be factored in to a patch that uses type inheritance for the parent props.

mathjazz:
Sorry for the conflicts I caused!

I'm getting an error on local build:

```
 => ERROR [22/24] RUN yarn build                                                                                                                                           124.6s
------
 > [22/24] RUN yarn build:
#26 1.895 yarn run v1.22.5
#26 2.198 $ react-scripts build
#26 7.016 Creating an optimized production build...
#26 123.8 Failed to compile.
#26 123.8
#26 123.8 /app/frontend/src/core/comments/components/AddComment.tsx
#26 123.8 TypeScript error in /app/frontend/src/core/comments/components/AddComment.tsx(57,29):
#26 123.8 Property 'character' does not exist on type 'Node'.
#26 123.8   Property 'character' does not exist on type 'BaseEditor'.  TS2339
#26 123.8
#26 123.8     55 |         const [isMentioned] = Editor.nodes(editor, {
#26 123.8     56 |             at: [],
#26 123.8   > 57 |             match: (n) => n.character === contactPerson,
#26 123.8        |                             ^
#26 123.8     58 |         });
#26 123.8     59 |
#26 123.8     60 |         if (contactPerson) {
#26 123.8
#26 123.8
#26 124.3 error Command failed with exit code 1.
#26 124.3 info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
------
executor failed running [/bin/sh -c yarn build]: exit code: 1
ERROR: Service 'webapp' failed to build
make: *** [build] Error 1
```

Pike:
I saw that coming when you updated slate. If you're on the pristine branch, make sure to clobber the frontend packages.

mathjazz:
Yup, just noticed that I broke that myself, too.

Pike:
Found https://docs.slatejs.org/concepts/12-typescript on how slate changed custom element types.

frontend/src/core/comments/components/AddComment.tsx
```diff
@@ -1,30 +1,55 @@
...
+    BaseEditor,
+    Editor,
+    Element as SlateElement,
+    Transforms,
+    Range,
+    createEditor,
+    Text,
+    Node,
+} from 'slate';
+import type { Descendant } from 'slate';
```
abowler2, line 14: This can be included in the above 'slate' imports unless we want to specifically call out that it is a `type` import

frontend/src/core/comments/components/AddComment.tsx
```diff
@@ -1,30 +1,55 @@
...
+    addComment: (arg0: string, arg1: number | null | undefined) => void;
+    resetContactPerson?: () => void;
+};
+
+type Paragraph = {
+    type: 'paragraph';
+    children: Descendant[];
+};
+
+type Mention = {
```
abowler2, line 39: The `Mention` type also includes a `url` which is type `string` (see original typing from `/core/api/types` [HERE](https://github.com/mozilla/pontoon/blob/master/frontend/src/core/api/types.js#L42))

abowler2 (commented):
Great work!  I started looking into this yesterday, but was not able to get to a final working solution. So happy to see that you were successful.

I did note a couple of things but overall it looks good to me.

frontend/src/core/comments/components/AddComment.tsx
Pike: Good catch, there's something fishy about TS not complaining about that type being wrong.

frontend/src/core/comments/components/AddComment.tsx
Pike: Yeah, pointing out that this is just a type import was intentional. I cross-checked the other imports that they're actually calling into the API layer there.

frontend/src/core/comments/components/AddComment.tsx
Pike: I force-pushed an update for this.

The `as Paragraph` cast is necessary, it seems. I haven't grokked why, but this way works.

PR #1950, opened 2021-05-20
Bug 1712086 validate messages from pontoon addon

MikkCZ:
I just typed this code blindly, didn't have time to run compilation or tests locally yet.


Pike (changes requested):
We should wrap the `JSON.parse` in a try-catch block to allow for non-JSON string messages to be posted.

MikkCZ:
After this PR (and reading several pages of the Flow docs) I know I will stay with TypeScript. :)

MikkCZ:
@Pike please re-review, I updated the branch from master.

frontend/src/modules/addonpromotion/components/AddonPromotion.tsx
```diff
@@ -39,13 +46,33 @@ export class AddonPromotionBase extends React.Component<InternalProps, State> {
...
+                try {
+                    data = JSON.parse(
+                        event.data as string,
+                    ) as PontoonAddonInfoMessage;
+                } catch (_) {
+                    return;
+                }
+                break;
+        }
+        if (data && data._type === 'PontoonAddonInfo' && data.value) {
```
Pike, line 73: Can you use optional chaining here and below?

```
if (data?._type === 'PontoonAddonInfo' && data.value?.installed === true)
```

I doubt we can do so in our js code base yet, but in the TS one, this should be transpiled away.

MikkCZ: It's done. I touched another place in the code as well, that deserved to be polished after migration to TypeScript.

Pike (approved):

frontend/src/modules/addonpromotion/components/AddonPromotion.tsx
Pike: Sweet, thanks for removing the ts-expect-error.

mathjazz (approved):
\o/

PR #1951, opened 2021-05-23
Bug 1712507: Use tsarch to validate frontend architecture constraints

Pike:
This PR also includes the current state of the TS PR, as it is based on that conversion. Only the last commit in this PR is of actual interest here.


mathjazz:
As Pike pointed out at the call, https://github.com/MaibornWolff/ts-arch has been archived, so let's close this.

PR #1955, opened 2021-05-27
Update frontend dependencies

mathjazz:
* slate
* react-time-ago
* [~css-loader~](https://github.com/webpack-contrib/mini-css-extract-plugin/issues/156) (will save that for later due to a bug in mini-css-extract-plugin)


mathjazz:
self-r+

PR #1969, opened 2021-07-06
Fix bug 1710475: Fix interactive tour selectors

mathjazz:
We changed the layout in https://github.com/mozilla/pontoon/pull/1932, so we need to update tour selectors accordingly.


abowler2:
Tested functionality and all is working. Code looks good.

PR #1970, opened 2021-07-14
Upgrade Python to 3.8.11

ranisalt:
[Changelog](https://docs.python.org/release/3.8.11/whatsnew/changelog.html#python-3-8-11-final)

Consider upgrading to 3.9, as 3.8 is in maintenance mode


mathjazz (approved):
Thanks for the patch!

Would you be for creating a followup patch to upgrade to 3.9.6?

ranisalt:
> Would you be for creating a followup patch to upgrade to 3.9.6?

Sure, I can do it

PR #1971, opened 2021-07-14
Upgrade Python to 3.9.6

ranisalt:
I have also bumped the Dockerfile to `3.9-buster`. After that, I ran `make build` then `make test` and it passed.

Would you also be interested in a general bump of dependencies as well? I'm sure upgrading the frontend dependencies would lead to faster build times and smaller bundles too.


Pike:
Anything in the 3.9 release notes that we might benefit from? Maybe also in https://github.com/asottile/pyupgrade ?

ranisalt:
@Pike several builtins are sped up due to [PEP 590 vectorcall](https://www.python.org/dev/peps/pep-0590), Unicode support was upgraded to [v13.0.0](https://www.unicode.org/versions/Unicode13.0.0/), [removeprefix and removesuffix on strings](https://www.python.org/dev/peps/pep-0616/) are the notable improvements. [There are some more improvements](https://docs.python.org/release/3.9.0/whatsnew/3.9.html) but those don't seem that useful around here.

ranisalt:
@Pike pyupgrade makes 2 fixes, should I include it in this PR or open a separate one?

mathjazz:
My personal take is to try to be on the latest stable release (similarly to our Django upgrade strategy).

I'd include pyupgrade fixes in this PR.

mathjazz:
Thanks for the update! Please run `make black` to fix the linting issue.

mathjazz:
Thanks!

One last thing - please also update automation config - at least https://github.com/mozilla/pontoon/blob/master/.github/workflows/py-lint.yml still references python 3.8.

ranisalt:
Aha, my bad, I grepped for `3.8` but it was on a hidden folder :laughing:

mathjazz (approved):
Great work!

PR #1973, opened 2021-07-16
Bug 1720875: Add warning for M1 MacBook users to Developer Setup docs

eemeli:
This doesn't fix the problem, but at least it documents a workaround for those affected.


mathjazz (approved):
Nice job!

PR #1975, opened 2021-07-26
Fix bug 1720977: Honour prefers-reduced-motion on homepage

mathjazz:
This patch disables the smooth scrolling animation on Pontoon homepage when `prefers-reduced-motion` is used.

See MDN article on how to test it on your OS:
https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion#user_preferences

### DIFF
--- a/frontend/src/modules/machinery/components/Translation.tsx
+++ b/frontend/src/modules/machinery/components/Translation.tsx
@@ -64,11 +64,14 @@ export default function Translation(
     }
 
     const translationRef = React.useRef();
+    const mediaQuery = window.matchMedia('(prefers-reduced-motion: reduce)');
+
     React.useEffect(() => {
         if (selectedHelperElementIndex === index) {
+            const behavior = mediaQuery.matches ? 'auto' : 'smooth';
             // @ts-expect-error: What can be undefined here?
             translationRef.current?.scrollIntoView?.({
-                behavior: 'smooth',
+                behavior: behavior,
                 block: 'nearest',
             });
         }
--- a/frontend/src/modules/machinery/components/Translation.tsx
+++ b/frontend/src/modules/machinery/components/Translation.tsx
@@ -66,7 +66,9 @@ export default function Translation(
     const translationRef = React.useRef();
     React.useEffect(() => {
         if (selectedHelperElementIndex === index) {
-            const mediaQuery = window.matchMedia('(prefers-reduced-motion: reduce)');
+            const mediaQuery = window.matchMedia(
+                '(prefers-reduced-motion: reduce)',
+            );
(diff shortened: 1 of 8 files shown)


eemeli:
Why is this special-casing the front page only? As in, should this preference also be applied to other transitions that currently use `behavior: 'smooth'`?

mathjazz:
The bug was filed for the homepage specifically, and it might be that the issue is noticeable only there due to the parallax scrolling. It's also possible we should do the same thing wherever we use smooth scrolling behavior.

I don't have a lot of experience in this area, so I'll ask in the bug.

mathjazz:
As confirmed by the reporter, we should disable smooth animations wherever possible.

All `scrollIntoView` calls now take `prefers-reduced-motion` into account.

@eemeli This is ready for another look.

frontend/src/modules/machinery/components/Translation.tsx
```diff
@@ -64,11 +64,14 @@ export default function Translation(
     }
 
     const translationRef = React.useRef();
+    const mediaQuery = window.matchMedia('(prefers-reduced-motion: reduce)');
+
     React.useEffect(() => {
         if (selectedHelperElementIndex === index) {
```
eemeli, line 70:
```suggestion
    React.useEffect(() => {
        if (selectedHelperElementIndex === index) {
            const mediaQuery = window.matchMedia('(prefers-reduced-motion: reduce)');
```

frontend/src/modules/otherlocales/components/Translation.tsx
```diff
@@ -62,11 +62,14 @@ export default function Translation(
     }, [dispatch, index, translation, copyOtherLocaleTranslation]);
 
     const translationRef = React.useRef();
+    const mediaQuery = window.matchMedia('(prefers-reduced-motion: reduce)');
+
     React.useEffect(() => {
         if (selectedHelperElementIndex === index) {
```
eemeli, line 68:
```suggestion
    React.useEffect(() => {
        if (selectedHelperElementIndex === index) {
            const mediaQuery = window.matchMedia('(prefers-reduced-motion: reduce)');
```

eemeli (approved):
Looks good! Presuming here that this is now ready for review?

The `mediaQuery` values in functional components don't need to be re-determined on every render, but only when they're actually needed. If that's a sufficiently heavy operation that it should be done only once (as is done in `homepage.js`, then `useMemo()` is your friend. Or just assigning the query to a variable at the file root?

mathjazz:
Thanks for the review and the suggestions!

I'll fix the prettier and merge.

PR #1976, opened 2021-07-26
Fix bug 1720895: Disable Django admin login form

mathjazz:
Django admin login form appears when Django admin (`/a/`) is accessed by users that aren't authenticated or don't have permission to access it (i.e. non-admins).

On https://pontoon.mozilla.org/ we use Firefox Accounts, which is based on OAuth, so we don't store user passwords, which means the form is useless.

Other installs might use authentication methods that store usernames and passwords, in which case the form could be useful, but it also poses a security risk. Brute-force attack could be used to find valid credentials.

This patch disables the Django admin login form and shows the standard Pontoon 403 page instead.

### DIFF
--- a/pontoon/urls.py
+++ b/pontoon/urls.py
@@ -73,6 +73,8 @@ urlpatterns = [
     path("accounts/", include("pontoon.allauth_urls")),
     # Admin
     path("admin/", include("pontoon.administration.urls")),
+    # Django admin: Disable the login form
+    path("a/login/", permission_denied_view),
     # Django admin
     path("a/", admin.site.urls),
     # Logout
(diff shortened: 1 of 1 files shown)


jotes (approved):

jotes:
The PR solves the issue. An additional security measure would be parametrizing the root of the URL via an environment variable and use a different value per instance - just to avoid accessing the Django admin views at all.

Pike (approved):
I've tried to find more information, and got to https://code.djangoproject.com/ticket/15755, which suggests to subclass the admin site. But I'm also not sure if that's going to work better.

Let's go with this.

PR #1977, opened 2021-07-27
Fix bug 1722562: Update FxA local configuration

mathjazz:
We use https://stable.dev.lcip.org FxA endpoint for local Pontoon development, but it's no longer recommended to use that server, which is also often down.

Instead, we should use the stage server:
https://wiki.mozilla.org/TestEngineering/Services/FxATestEnvironments#FxA_Stage_Environment

This patch does that.


Pike:
This secret is bound to the server that we're logging in to? There's no real way to test this ?

mathjazz:
The secret is specific for the stage FxA endpoint URL and our `redirectUri`.

Forgot to add testing instructions:
1. On this branch, run `make build`, which runs `./manage.py update_auth_providers`. If you don't want to wait for the new build, you can also create/update the `SocialApp` DB entry in http://localhost:8000/a/socialaccount/socialapp/.
1. Set `settings.AUTHENTICATION_METHOD` to `fxa`.
1. Log in on http://localhost:8000.

Pike (approved):
Tested locally, with a ton of hiccups. `make build` wasn't as reliable as I had hoped

PR #1978, opened 2021-08-02
Use constraints for python deps, fix some deprecation warnings

Pike:


mathjazz:
LGTM.

Could you also update the comment in the header of the `.in` files to also include `lint.in` and replace `compile_all.sh` with `make requirements`?

Pike:
Like so?

mathjazz (approved):
Like so.

PR #1981, opened 2021-08-09
Bug 1724658 - Concordance search term should be added to editor

abowler2:
As per the [Concordance search specs](https://github.com/mozilla/pontoon/blob/master/specs/0106-concordance-search.md) when clicking on a search term the term should be added to the editor and not replace the editor contents.  

This change was made in this [PR](https://github.com/mozilla/pontoon/pull/1771); however, changes made in this [patch](https://github.com/mozilla/pontoon/pull/1799) altered that functionality.

This work restores the expected behavior.

### DIFF
--- a/frontend/src/modules/machinery/components/Machinery.tsx
+++ b/frontend/src/modules/machinery/components/Machinery.tsx
@@ -123,6 +123,7 @@ export default class Machinery extends React.Component<Props, State> {
                             return (
                                 <Translation
                                     index={index}
+                                    entity={machinery.entity}
                                     sourceString={machinery.sourceString}
                                     translation={translation}
                                     key={index}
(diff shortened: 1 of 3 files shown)


frontend/src/modules/machinery/components/Machinery.tsx
```diff
@@ -123,6 +123,7 @@ export default class Machinery extends React.Component<Props, State> {
                             return (
                                 <Translation
                                     index={index}
+                                    entity={machinery.entity}
```
abowler2, line 126: We need to pass the actual value of entity to `Translation`. This is because a search string will not have an entity value. 

We then later check within the `useCopyMachineryTranslation` hook to see if `entity` has a value and if it is `null` we know it is a search string and we add the term to the editor instead of replacing the editor content.

mathjazz (approved):
Excellent research and fix!

PR #1984, opened 2021-08-17
Fix bug 1726219: Prevent infinite loop on generating plural examples

mathjazz:
More context in the [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1726219) and the [Discourse post](https://discourse.mozilla.org/t/string-freezes-when-clicked/84627).

### DIFF
--- a/frontend/src/core/locale/getPluralExamples.test.js
+++ b/frontend/src/core/locale/getPluralExamples.test.js
@@ -18,12 +18,20 @@ describe('getPluralExamples', () => {
     });
 
     it('prevents infinite loop if locale plurals are not configured properly', () => {
+        const spy = jest.spyOn(console, 'error').mockImplementation(() => {});
         const locale = {
             cldrPlurals: [0, 1, 2, 3, 4, 5],
             pluralRule: '(n != 1)',
         };
-        const res = getPluralExamples(locale);
-        const expected = { undefined: 1 };
-        expect(res).toEqual(expected);
+        try {
+            const res = getPluralExamples(locale);
+            const expected = { 0: 1, 1: 2 };
+            expect(res).toEqual(expected);
+            expect(spy).toHaveBeenCalledWith(
+                'Unable to generate plural examples.',
+            );
+        } finally {
+            spy.mockRestore();
+        }
     });
 });
--- a/frontend/src/core/locale/getPluralExamples.test.js
+++ b/frontend/src/core/locale/getPluralExamples.test.js
@@ -1,0 +1,29 @@
+import getPluralExamples from './getPluralExamples';
+
+describe('getPluralExamples', () => {
+    it('returns a map of Slovenian plural examples', () => {
+        const locale = {
+            cldrPlurals: [1, 2, 3, 5],
+            pluralRule:
+                '(n%100==1 ? 0 : n%100==2 ? 1 : n%100==3 || n%100==4 ? 2 : 3)',
+        };
+        const res = getPluralExamples(locale);
+        const expected = {
+            1: 1,
+            2: 2,
+            3: 3,
+            5: 0,
+        };
+        expect(res).toEqual(expected);
+    });
+
+    it('prevents infinite loop if locale plurals are not configured properly', () => {
+        const locale = {
+            cldrPlurals: [0, 1, 2, 3, 4, 5],
+            pluralRule: '(n != 1)',
+        };
+        const res = getPluralExamples(locale);
+        const expected = { undefined: 1 };
+        expect(res).toEqual(expected);
+    });
+});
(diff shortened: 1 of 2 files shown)


mathjazz:
@eemeli Let me know if you need more context to review this.

Pike:
I'm wondering if there's a way to avoid this problem from being created. Where does the data enter? Can we do some sort of form validation via static analysis of the plural form? Or something something yadda yadda?

jotes:
There's also a security bug related to this: https://bugzilla.mozilla.org/show_bug.cgi?id=1311053. I think that's something that can be handled on the server-side (Mathjazz also suggests it, as a solution).

mathjazz:
Data typically gets in through the Admin, where PMs add or edit locales. [CLDR data](https://unicode-org.github.io/cldr-staging/charts/37/supplemental/language_plural_rules.html) is used as a source, unless the locale isn't covered by CLDR, in which case PMs cunsult with localizers.

Since CLDR data also includes examples, maybe the right solution is to simply add a comma separated integer field to the Locale model to store examples, and let the PMs type it manually. It's not a very common task after all.

Separately from that and maybe not particularly interesting for Mozilla's instance with already 200+ locales entered, we could also try to pull as much locale data as possible from CLDR automatically. [Related bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1396915).

eemeli:
Sorry, had missed the review notification on this (my feed got simultaneously swamped by stuff from [yaml/yaml-spec](https://github.com/yaml/yaml-spec)).

This is okay as a patch, but a proper-er fix would be for me to include the example data in the `intl-pluralrules` (already a dependency), which is parsing it from CLDR data. It's used there internally to generate tests for its generated plural rule functions, so it's just a matter of exposing the data.

Also, the `eval()` is in fact a bit evil here, and its use ought to be replaced by `new Function()`, which at least limits the context to which the executed code has access.

@mathjazz, it's probably a few hours' work for me to do all that. Should I do so, and then update this PR?

mathjazz:
Thanks, that sounds great!

Re-assigned the bug to you!

eemeli:
@mathjazz I rebased your branch on `master` and replaced `eval()` with `new Function()`. This solves the reported bug more cleanly; I'll post a separate one about getting the examples from `make-plural`.

eemeli:
Also, apparently I can't request a review from @mathjazz because he's the one that originally filed this PR. 😄

mathjazz:
My understanding is that we'll get rid of all this code by using `make-plural`.

Is that correct? If yes, does it even make sense to merge this patch?

eemeli:
`make-plural/examples` solves this for all locales that are in CLDR, but wasn't it so that Pontoon supports others as well?

frontend/src/core/locale/getPluralExamples.ts
```diff
@@ -24,15 +24,14 @@ export default function getPluralExamples(
...
+        ) => number | boolean;
         let n = 0;
         while (Object.keys(examples).length < pluralsCount) {
-            // This `eval` is not so much evil. The pluralRule we parse
-            // comes from our database and is not a user input.
-            // eslint-disable-next-line
-            const rule = eval(locale.pluralRule);
-            if (!examples[locale.cldrPlurals[rule]]) {
-                examples[locale.cldrPlurals[rule]] = n;
+            const rule = locale.cldrPlurals[Number(getRule(n))];
```
mathjazz, line 32: Suggestion: Rename `rule` to `category`, because `locale.cldrPlurals` holds plural categories.

frontend/src/core/locale/getPluralExamples.test.js
```diff
@@ -18,12 +18,20 @@ describe('getPluralExamples', () => {
...
-        const res = getPluralExamples(locale);
-        const expected = { undefined: 1 };
-        expect(res).toEqual(expected);
+        try {
+            const res = getPluralExamples(locale);
+            const expected = { 0: 1, 1: 2 };
+            expect(res).toEqual(expected);
+            expect(spy).toHaveBeenCalledWith(
+                'Unable to generate plural examples.',
+            );
```
mathjazz, line 32: Very nice work!

mathjazz (commented):
I see. My (mis)understanding from last week's call was non-CLDR locales would be handled differently.

Could you please also remove the `unsafe-eval` from our CSP policy. It'd be great to push Mozilla Observatory score beyond 100 points. :)

mathjazz:
Update: we decided not to go forward with `make-plural/examples`, because implementation in this patch is sufficient (it's not slow and makes our bundle smaller).

To only thing that remains to be done here is to change the CSP.

eemeli:
@mathjazz Unfortunately, it looks like we can't drop `unsafe-eval` because that's also needed for `new Function()`: https://www.w3.org/TR/CSP3/#directive-script-src

If we do want to drop it, then we would need to generate the JS function source completely in the backend, and deliver it via some HTTP endpoint that's directly interpreted as JS by the frontend. That sounds like a valid, but different bug.

mathjazz (commented):
That's a bummer...

And I can't even approve my own pull request.

PR #1985, opened 2021-08-20
Fix bug 1726295: Show XLIFF string ID as Context

mathjazz:
Introduces the `Entity.context` field for storing Context data directly in the model instead of extracting it from `Entity.key`. The field is populated for existing Entity objects with a data migration, and gets set properly on sync for new Entities.

This patch also makes the `Entity.key` required, which we should have done long time ago.


pontoon/base/migrations/0018_populate_entity_context.py
```diff
@@ -0,0 +1,41 @@
...
+
+def add_entity_context(apps, schema_editor):
+    Entity = apps.get_model("base", "Entity")
+    entities = Entity.objects.all().prefetch_related("resource")
+
+    for e in entities:
+        if e.resource.format == "lang":
+            e.context = ""
+        if e.resource.format == "po":
+            e.context = e.key.split(KEY_SEPARATOR)[0]
```
jotes, line 16: *idea* Postgresql has `split_part` function, maybe something like this could be a little bit faster (in terms of the time and memory):
```
 split_key_po  = Func(F('key'), Value(KEY_SEPARATOR), Value(1), function='split_part', output_field=TextField())
 split_key_xliff = Func(F('key'), Value(KEY_SEPARATOR), Value(2), function='split_part', output_field=TextField())
Entity.objects.filter(resource__format='lang').update(context='')
Entity.objects.filter(resource__format='po').update(context=split_key_po)
Entity.objects.filter(resource__format='xliff').update(context=split_key_xliff)
Entity.objects.exclude(resource__format__in=['lang', 'po', 'xliff']).update(context=F('key'))
```
You can also try to create a raw SQL to make everything in one query:
```
UPDATE base_entity SET context=CASE
WHEN format='po' THEN split_part(key, '<KEY_SEPARATOR> ', 1)
WHEN format='xliff' THEN split_part(string, '<KEY_SEPARATOR> ', 2)
WHEN format='lang' THEN ''
ELSE key
END
FROM base_resource WHERE base_entity.resource_id=base_resource.id
```

pontoon/base/models.py
```diff
@@ -2503,7 +2503,10 @@ class Entity(DirtyFieldsMixin, models.Model):
     resource = models.ForeignKey(Resource, models.CASCADE, related_name="entities")
     string = models.TextField()
     string_plural = models.TextField(blank=True)
-    key = models.TextField(blank=True)
+    # Unique identifier, used to compare DB and VCS objects
+    key = models.TextField()
```
jotes, line 2507: *question* This is probably going to be detected during testing on staging, but I wonder if there isn't an ancient bug/scenario which makes this field empty or null.

jotes (approved):

pontoon/base/models.py
mathjazz: I checked and there isn't.

pontoon/base/migrations/0018_populate_entity_context.py
mathjazz: Thanks, I'll update!

This is a one-off data migration and it only took about a minute to complete on stage with prod data, but I expect it to be much faster with your approach. Maybe someone else will benefit even more than our instance.

PR #1987, opened 2021-08-20
Fix bug 1726749: Fail gracefully on parsing errors in XLIFF files

mathjazz:
We already fail gracefully on parsing errors for all other file formats, i.e. we log the error and continue syncing the next file. So let's do that for XLIFF files, too.

I'm also making two other related changes along the way:
- FTL and Silme libs handle parsing errors gracefully by themselves on a message level, so I'm just adding a note to the error handling blocks for these two formats.
- For FTL and Silme we now raise `ParseError` when we hit OSError, to be consistent with other monolingual formats.

### DIFF
--- a/pontoon/sync/formats/silme.py
+++ b/pontoon/sync/formats/silme.py
@@ -12,7 +12,7 @@ from silme.format.ini import FormatParser as IniParser
 from silme.format.inc import FormatParser as IncParser
 from silme.format.properties import FormatParser as PropertiesParser
 
-from pontoon.sync.exceptions import SyncError
+from pontoon.sync.exceptions import ParseError, SyncError
 from pontoon.sync.utils import (
     create_parent_directory,
     escape_quotes,
@@ -102,13 +102,13 @@ class SilmeResource(ParsedResource):
                     uncomment_moz_langpack=parser is IncParser and not source_resource,
                 )
             )
-        except OSError:
+        except OSError as err:
             # If the file doesn't exist, but we have a source resource,
             # we can keep going, we'll just not have any translations.
             if source_resource:
                 return
             else:
-                raise
+                raise ParseError(err)
 
         comments = []
         current_order = 0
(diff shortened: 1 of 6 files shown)


flodolo:
Passing by notes: we should probably add these to Papertrail alerts once this merges? Otherwise they'll go unnoticed for ages.

mathjazz:
Good idea!

mathjazz:
Jotes, would you have cycles to review this?

pontoon/sync/formats/silme.py
```diff
@@ -102,13 +102,15 @@ def __init__(self, parser, path, source_resource=None):
                     uncomment_moz_langpack=parser is IncParser and not source_resource,
                 )
             )
-        except OSError:
+        # Parse errors are handled gracefully by fluent
```
jotes, line 105: *nitpick* by `silme` (?)

pontoon/sync/formats/xliff.py
```diff
@@ -125,5 +127,10 @@ def save(self, locale):
 def parse(path, source_path=None, locale=None):
     with open(path) as f:
         xml = f.read().encode("utf-8")
-        xliff_file = xliff.xlifffile(xml)
+
+        try:
+            xliff_file = xliff.xlifffile(xml)
+        except etree.XMLSyntaxError as err:
+            raise ParseError(f"Failed to parse {path}: {err}")
```
jotes, line 134: *question* What do you think about passing path and err as parameters to ParseError class and implementing a method to display a unified error message in other formats? I mean, something like this:
```
class ParseError(RuntimeError):
   def __init__(path, err):
      super().__init__(f"Failed to parse {path}: {err}")
 
ParseError(path, error)
```

jotes (approved):

pontoon/sync/formats/xliff.py
mathjazz: I see your point. I agree it'd be good to keep messaging in one place. But I'd rather do that separately for all custom error classes and all places where we log relevant messages.

PR #1988, opened 2021-08-22
Introduce redux toolkit

Pike:
This is maybe good in its own right, maybe not. More opinions to deal with for sure.

Yet, `redux-thunk` is a hot mess to type, and redux toolkit does that, at least once you update redux.

Eemeli, what's your take?

Also, CC @abowler2


eemeli (approved):
Looks pretty straightforward. Appears to contain what it says on the tin, and gets rid of a good number of `any` types.

PR #1989, opened 2021-08-23
Fix bug 1693321: Link to actionable strings from "new strings" notifications

eemeli:
This turned out to be a decent intro to Pontoon's structure. The logic I went with is that if the user has their homepage locale set, that's the one that's used for the strings; otherwise we default to the project's frontpage, as before. The notification itself is locale-less, so that info needs to come from somewhere.

The URL that's actually used is the locale-agnostic one, though, as that will check if the homepage locale actually exists for the project. The redirect should be really fast, but this allows for that check to only be done when a notification link is clicked. It would be possible to use this even if the homepage locale is unset, in which case the request's `Accept-Language` instead controls the redirect target. Would that be better than going to the project's frontpage?

In some cases it might be possible to use the current rather than preferred locale by parsing it from the `Referer` HTTP request field, but it doesn't look like any of the endpoints currently vary depending on that field, so this would be somewhat novel behaviour.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -290,12 +290,26 @@ def serialized_notifications(self):
         is_comment = False
 
         if hasattr(notification.actor, "slug"):
-            actor = {
-                "anchor": notification.actor.name,
-                "url": reverse(
-                    "pontoon.projects.project", kwargs={"slug": notification.actor.slug}
-                ),
-            }
+            if "new string" in notification.verb and self.profile.custom_homepage:
+                actor = {
+                    "anchor": notification.actor.name,
+                    "url": reverse(
+                        "pontoon.translate.locale.agnostic",
+                        kwargs={
+                            "slug": notification.actor.slug,
+                            "part": "all-resources",
+                        },
+                    )
+                    + "?status=missing,unreviewed,fuzzy",
+                }
+            else:
+                actor = {
+                    "anchor": notification.actor.name,
+                    "url": reverse(
+                        "pontoon.projects.project",
+                        kwargs={"slug": notification.actor.slug},
+                    ),
+                }
         elif hasattr(notification.actor, "email"):
             actor = {
                 "anchor": notification.actor.name_or_email,
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -300,7 +300,7 @@ def serialized_notifications(self):
                             "part": "all-resources",
                         },
                     )
-                    + "?status=missing,unreviewed,fuzzy",
+                    + "?status=missing,fuzzy",
                 }
             else:
                 actor = {
--- a/pontoon/contributors/templates/contributors/widgets/notifications_menu.html
+++ b/pontoon/contributors/templates/contributors/widgets/notifications_menu.html
@@ -33,7 +33,11 @@
 
         {% if notification.actor.slug %}
           {% set actor_anchor = notification.actor %}
-          {% set actor_url = url('pontoon.projects.project', notification.actor.slug) %}
+          {% if "new string" in notification.verb and user.profile.custom_homepage %}
+            {% set actor_url = url('pontoon.translate.locale.agnostic', notification.actor.slug, "all-resources") + '?status=missing,unreviewed,fuzzy' %}
+          {% else %}
+            {% set actor_url = url('pontoon.projects.project', notification.actor.slug) %}
+          {% endif %}
 
         {% elif notification.actor.email %}
           {% set actor_anchor = notification.actor.name_or_email|nospam %}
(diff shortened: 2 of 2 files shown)


pontoon/base/models.py
```diff
@@ -290,12 +290,26 @@ def serialized_notifications(self):
         is_comment = False
 
         if hasattr(notification.actor, "slug"):
-            actor = {
-                "anchor": notification.actor.name,
-                "url": reverse(
-                    "pontoon.projects.project", kwargs={"slug": notification.actor.slug}
-                ),
-            }
+            if "new string" in notification.verb and self.profile.custom_homepage:
```
mathjazz, line 293: As you hinted at in the description, I'd remove the check for `self.profile.custom_homepage` here, and also in the other frontend below, of course.

The locale-agnostic translate view already handles the case where `custom_homepage` isn't set and redirects to the exact same URL we generate in the `else`-block, which you could then drop.

pontoon/base/models.py
```diff
@@ -290,12 +290,26 @@ def serialized_notifications(self):
...
+                actor = {
+                    "anchor": notification.actor.name,
+                    "url": reverse(
+                        "pontoon.translate.locale.agnostic",
+                        kwargs={
+                            "slug": notification.actor.slug,
+                            "part": "all-resources",
+                        },
+                    )
+                    + "?status=missing,unreviewed,fuzzy",
```
mathjazz, line 303: I'd prefer that to be `missing` only, since we only send notifications for newly created strings, but I'll let @flodolo make the call. :)

mathjazz (commented):
Nice work!

Let's go with the locale-agnostic translate view everywhere.

pontoon/base/models.py
mathjazz: One more thing I forgot to mention - it'd be slightly more reliable if we detected new string notifications with a regex match on the notification verb.

pontoon/base/models.py
flodolo: Fuzzy don't show up with "missing", and fuzzy are typically added as new strings. So, we need at least missing+fuzzy.

mathjazz: Fuzzies are actually updated *existing* strings (which allows us to keep translations).

flodolo: > Fuzzies are actually updated _existing_ strings (which allows us to keep translations).

You mean for Pontoon's internal logic? Because for users they're brand new strings.

mathjazz: Correct - that's also why they aren't included in the count shown in the new string notification. Which makes me think including fuzzy there will be confusing.

flodolo: I think they are… I've received a notification from AMO Frontend for 4 new strings this week, and 3 of them were fuzzy.

mathjazz: My bad. They are indeed created as new strings, and translations are re-imported and marked as Fuzzy. Sorry for the confusion and thanks for explaining me how my stuff works. :)

mathjazz: @eemeli Let's remove `unreviewed` from the list of stati.

pontoon/base/models.py
eemeli: Can do. The reason why I didn't do this initially is that if `custom_homepage` isn't set, locale-agnostic goes to the HTTP header's first `Accept-Language` locale. Now, if the user wants to get from there to whatever they consider the "right" locale, they need to 1. click on the current locale, 2. select another locale, 3. select the project where they were. That seems like it makes the cost of being wrong kinda high here, no?

Re: string matching, do I understand right that you'd like for the test condition to be stricter by requiring a closer match?

mathjazz: While the price might be hard, very few people will need to pay for it. We have a localizer though that translates two locales (`dsb` and `hsb`). For him the current behaviour might actually be better.

@flodolo What's your take?

(Note that one day locales will be [easier to switch](https://bugzilla.mozilla.org/show_bug.cgi?id=1578235). 😄 We used to have a locale menu in the old Translate view, but then people were requesting an easier access to the dashboards.)

---

Re string matching, yes, that's what I'd like, but I'll leave the decision to you. It would be better if you could match `updated with {} new string` more closely, but as we discussed, this is only a temporary solution anyways. Going forward, we should group notifications properly with a dedicated field.

flodolo: I kind of like the idea of checking the custom homepage first. I'm afraid that we have a bunch of localizers translating content and using a different accept-languages in the browser.

mathjazz: OK, let's keep the logic here!

mathjazz (approved):
Well done!

PR #1993, opened 2021-08-29
No bug, add automation to go through some stricter TSC settings

Pike:


Pike:
I hinted at doing this, let's keep tabs on how we're doing towards stricter typescript checks.

I recall @abowler2 mentioning strict null, so I did that explicitly. Also implicit any, which is a mix of missing typedefs and bad types on our side. And then just straight strict.

Not sure if this is a good set.

Obviously, all of those failing tsc are supposed to pass, thus the `|| true`

abowler2 (approved):
I think these are a great place to start.  I like that they are separated so that we can focus on them independently (of course, some fixes may impact more than one check).

PR #1994, opened 2021-08-29
No bug, use in-tree action to group tsc checks output

Pike:
I wanted to make it easier to see the numbers of errors in a PR, and thought this could be done in two ways:

Add an extra log group for the tsc invocation, which collapses the full log. You can still expand the twisty if you want to know the details. It repeats the error count afterwards.

Collect the error count as step output, and thus enabling a summary step that just prints all the numbers together.

The best way to do that that I could come up with was to actually create an action to wrap tsc invocations.

### DIFF
--- a/.github/actions/check-tsc/index.js
+++ b/.github/actions/check-tsc/index.js
@@ -1,0 +1,31 @@
+const { promisify } = require('util');
+const {exec} = require('child_process');
+
+
+async function run() {
+    console.log("::group::tsc");
+    let errors = '0';
+    const run = process.env['INPUT_RUN'] || 'yarn types --pretty ';
+    const cwd = process.env['INPUT_WORKING-DIRECTORY'] || 'frontend';
+    let stdout, stderr;
+    try {
+        ({stdout, stderr} = await asyncExec(run, {
+            cwd,
+        }));
+    } catch (failed_proc) {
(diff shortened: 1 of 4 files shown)


.github/actions/check-tsc/index.js
```diff
@@ -0,0 +1,31 @@
+const { promisify } = require('util');
+const {exec} = require('child_process');
```
eemeli, line 2: Should probably run Prettier on this file?

.github/actions/check-tsc/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "check-tsc",
+  "version": "0.0.0",
```
eemeli, line 3:
```suggestion
```
With `private: true` this shouldn't be required.

eemeli (approved):
Looks good, though a bit overengineered. Added a couple of nitpicks as line comments.

Pike:
True story, I actually de-engineered this quite a bit. Not that it ever was in TypeScript, but it was close to being so a few times. And it did have external dependencies for exec and console.log 💣 .

The other part of me was tempted to put this into an exercise to learn PowerShell.

And yes, I started with bash, and failed, and then overengineered it. And then stripped it back down. Who cares about the order of stdout and stderr anyway, right?

Off to get some help.

PR #1996, opened 2021-08-30
Add a separate frontend container

eemeli:
This is intended to make development work a bit easier by separating the TS React app and the Django server into their own Docker containers. There's also an env var `FRONTEND_URL` that allows for using e.g. a non-dockerised `frontend` where that makes sense.

A few relevant make tasks are added; the previous `build-frontend` is renamed to `build-tagadmin` for clarity. `make run` now internally uses `docker-compose up` rather than `docker-compose run`, as it needs to start two services rather than just one. `make test` has two sub-tasks that it runs: `make test-webapp` and `make test-frontend`. (**Edit:** Last sentence updated after 6969965)

The `yarn build` command is left out completely from the initial container build, as the dev system actually uses `yarn start` to generate and serve the relevant files at runtime. The `frontend` container is based on `node:14-buster`, which has the same parent image `buildpack-deps:buster` as the webapp's parent `python:3.9-buster`.

The `webapp` image's `RUN apt-get` line is not updated here, but the yarn stuff could be left out of it. It might be a good idea to separately from this PR update that image to also use Node.js v14 rather than the current v10.

Also separately from this change, it might be good to use a named volume for the dev database so that it might be retained between instantiations. That would allow for `docker-compose up` to also handle its container.


jotes:
I think it's a good choice to split the monolithic docker image into two smaller ones. It will be easier to rebuild the images for the backend/frontend separately.
It also makes the docker files easier to read, easier to deploy, and to understand their purpose.

eemeli:
I added a frontend build stage to the `webapp` dockerfile, so that it continues to provide a prod-ready frontend as previously. This also made the last tests pass.

As this container effectively shares a base with the `frontend` dev container, it would be possible to use that also in the build stage (rather than duplicating its contents), if the here added `local/pontoon-frontend` image were available during the `webapp` build. Adding this requirement could be argued to be a change in the project's external API, as someone who's using the dockerfile directly would need to build the frontend as a first step.

.gitignore
```diff
@@ -28,6 +28,8 @@ coverage/
 
 # Docker
 .docker-build
+.frontend-build
+.webapp-build
```
mathjazz, line 32: Should we call them `frontend` and `backend`?

Or `react` and `django`?

CONTRIBUTING.rst
```diff
@@ -381,16 +381,19 @@ steps, as they don't affect your setup if nothing has changed:
...
+We use webpack to build our JavaScript files for some pages
+(currently only the tag admin UI).
+While `make build` will build those files for you,
+you might want to rebuild them while programming on the front.
+To build the files just once, run:
 
 .. code-block:: shell
 
-    $ make build-frontend
+    $ make build-tagadmin
```
mathjazz, line 392: Thank you!

docker/run_tests.sh
```diff
@@ -43,15 +43,6 @@ $WEBPACK_BINARY
...
 
-echo ""
-echo "--------------------------------------------------------------------------------------------"
-echo "Running JavaScript tests"
-$NPM test
-pushd frontend
-yarn test --watchAll=false
-popd
-
-
```
mathjazz, line 54: Seems like this means we can no longer run the entire test suite with one `make` command as [documented](https://github.com/mozilla/pontoon/blob/3ff1829fc20129df076401da39fe90d9adcc3304/CONTRIBUTING.rst#running-tests)?

mathjazz:
As agreed on the call, let's give a heads-up about this change on https://github.com/ibm-skills-network/pontoon.

.gitignore
eemeli: Whatever the naming system, it ought to be consistent. Which is why I went with `frontend` as the name for the new thing, as that matches the directory.

`react` and `django` feel like they're mixing concerns a bit by referring to the implementation, when we ought to care more about its usage or external API.

docker/run_tests.sh
eemeli: Good point. Separated `make test-webapp` from `make test`, which now also runs the `frontend` tests.

.gitignore
mathjazz: Good point about avoiding the implementation details in naming conventions.

For container names to be consistent with directory names, we'd also need to rename the top-level `pontoon` folder. Even if not 100% accurate, I'd prefer `backend` over `webapp`. I worry though that renaming the `pontoon` folder will require changes to be made in too many places, and that we might want to split webapp/pontoon further in the future.

So how about we just call the `webapp` container `backend` for now, and leave the directory structure unchanged? That would at least fix the current situation, where `webapp` sounds like something `frontend` is a part of.

eemeli: To bikeshed this a bit a more, would you be ok with `frontend` and `server` containers, rather than calling the latter `backend`? As it is serving the Django parts of the frontend, that might be more a bit more appropriate.

mathjazz: Ok, let's do that.

Makefile
```diff
@@ -32,37 +37,45 @@ help:
...
 	"${DC}" run webapp //app/docker/set_up_webapp.sh
 
-run: .docker-build
-	"${DC}" run --rm --service-ports webapp
+run: .frontend-build .webapp-build
+	"${DC}" up --no-build frontend webapp
 
 clean:
-	rm .docker-build
+	rm -f .docker-build .frontend-build .webapp-build
```
adngdb, line 69: I guess you can remove `.docker-build` here, as it's not being used anymore.

Makefile
```diff
@@ -32,37 +37,45 @@ help:
...
+
+test: test-webapp test-frontend
 
-test:
+test-webapp:
 	"${DC}" run --rm webapp //app/docker/run_tests.sh
```
adngdb, line 74: This looks wrong. The `run_tests.sh` file is actually meant to run the full test suite, frontend and webapp/server included. It's basically what the CI runs, but in docker. 

If I understand things correctly, `test-webapp` should actually be an alias of `pytest` just like `test-frontend` is an alias of `jest`. And `test`, I think, should stay as it was. 

If that's not the case and I missed something, then the `run_tests.sh` file should probably be revisited. :-)

adngdb (approved):
From my "I-haven't-touched-at-pontoon-for-a-year" point of view, this looks great! I think there's some minor work needed in making the various testing make endpoints coherent, but the rest looks good.

Good Lord, I can send a gif here again! Sweet joy!

Makefile
eemeli: I left it in for now, to make it easier for anyone that has a current Pontoon checkout. If this gets cleaned up sometime later rather than now, anyone who's run `make clean` at least once in the interim won't need to manually remove the old empty target file.

Makefile
eemeli: At least at the moment, `run_tests.sh` also lints the code, i.e. it's not just an alias for `pytest`. Because it's run within the container, it can't after this change also run the tests for the translate frontend, as that's now in a different container. But it can still run the rest of its commands.

To keep the old behaviour of `make test`, we need these new targets, so that `test-webapp` and `test-frontend` can run all the tests that their corresponding containers offer.

mathjazz: I also think `test-webapp` should only run the unit tests (i.e.  `pytest`) to be consistent with its `test-frontend` counterpart.

And `test` should run unit tests across the board, currently `test-webapp` and `test-frontend`.

I don't consider linting and formatting as tests, so I'd add a new command (i.e. `ci`) to run everything we run in CI.

eemeli:
@mathjazz Merged master into the PR branch & renamed `webapp` to `server`, and dropped the separate frontend Dockerfile.

docker/k8s-first-steps/run_server.sh
```diff
@@ -1,10 +1,10 @@
 #!/bin/bash
-echo "out/logs is going into '/app/run_webApp.log'"
+echo "out/logs is going into '/app/run_server.log'"
```
mathjazz, line 2: Not sure about the log file, but I guess we should rename this file to `server_run.sh` to be consistent with the naming in the parent folder.

Makefile
```diff
@@ -33,69 +38,77 @@ help:
...
-	"${DC}" run --rm webapp pyupgrade --exit-zero-even-if-changed --py38-plus *.py `find pontoon -name \*.py`
+	"${DC}" run --rm server pyupgrade --exit-zero-even-if-changed --py38-plus *.py `find pontoon -name \*.py`
 
 types:
-	"${DC}" run --rm -w //app/frontend webapp yarn types
+	"${DC}" run --rm -w //frontend frontend yarn types
 
 prettier:
-	"${DC}" run --rm webapp npm run prettier
+	"${DC}" run --rm server npm run prettier
```
mathjazz, line 97: How do I run `prettier` (and `check-prettier` and `eslint`) in the frontend container?

Makefile
```diff
@@ -33,69 +38,77 @@ help:
...
 clean:
-	rm .docker-build
+	rm -f .docker-build .frontend-build .server-build
+
+test: test-server test-frontend
 
-test:
-	"${DC}" run --rm webapp //app/docker/run_tests.sh
+test-server:
+	"${DC}" run --rm server //app/docker/server_tests.sh
```
mathjazz, line 75: I think `test-server` should only run unit tests (i.e. pytest): https://github.com/mozilla/pontoon/pull/1996#discussion_r706076167.

Makefile
eemeli: Made `make test-server` effetively an alias for `make pytest`, and added a new `make ci` that tests & lints the frontend & server, as is actually done in CI.

docker/k8s-first-steps/run_server.sh
eemeli: Renamed.

Makefile
eemeli: I added the frontend commands to the mentioned tasks.

Makefile
eemeli: Done.

mathjazz (approved):
Great work!

I think there are some checks we'd need to add to the `ci` command (namely `types`, `eslint` on server and `check-pyupgrade`), but let's keep it as it is for now and change it to `ci: test lint` and drop `server_tests.sh` as part of #2011.

PR #1997, opened 2021-08-30
No Bug, fix some easy TypeScript implicit anys

Pike:
There's obviously hard parts with implicit anys, but there's also low hanging fruit.

noImplicitAny: 137 -> 85
strictNullChecks: 103 -> 102
strict: 200 -> 168

I haven't actually tested this locally yet.


frontend/src/modules/teamcomments/reducer.ts
```diff
@@ -25,15 +25,14 @@ function togglePinnedComment(
             return comment;
         }
 
-        comment.pinned = pinned;
-
         return {
             ...comment,
+            pinned: pinned,
```
Pike, line 30: 🧨

frontend/src/modules/search/components/SearchBox.tsx
```diff
@@ -162,25 +162,25 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
     }
 
     getInitialStatuses(): Record<string, boolean> {
-        const statuses = {};
+        const statuses: Record<string, boolean> = {};
```
Pike, line 165: I'm not happy with how typing works here, but it also seemed very involved.

Basically, I think we need to get the flag lists to be some sort of enums to get better typing here, but I also didn't find any easy wins here.

frontend/src/core/comments/components/AddComment.tsx
```diff
@@ -413,7 +413,7 @@ export default function AddComment(props: Props): React.ReactElement<'div'> {
 
     const submitComment = () => {
         if (Node.string(editor).trim() === '') {
-            return null;
+            return;
```
Pike, line 416: 🧨

frontend/src/core/comments/components/AddComment.tsx
```diff
@@ -321,7 +321,7 @@ export default function AddComment(props: Props): React.ReactElement<'div'> {
              * the last character so the BOM was added
              */
             editor.insertText('\n\uFEFF');
-            return null;
+            return;
```
Pike, line 324: 🧨

frontend/src/modules/resourceprogress/components/ProgressChart.tsx
```diff
@@ -92,7 +92,7 @@ export default class ProgressChart extends React.Component<Props> {
         const x = canvas.width / 2;
         const y = canvas.height / 2;
         const radius = (canvas.width - context.lineWidth) / 2;
-        let end = null;
+        let end: number = null;
```
Pike, line 95: I was wondering if I could set `end = -0.5` and be done, but I had no idea how to actually check if my reading of the code was right.

Pike (commented):
Some self-review comments. I left some 🧨 to indicate the places where I could have introduced behavioral changes.

Pike (commented):
I think I was able to test all changes that actually imply changes, in particular multi-line comments, just whitespace comments (do not submit), and pinning a comment.

frontend/src/modules/resourceprogress/components/ProgressChart.tsx
mathjazz: Yup, I think you are right. Please also update the logic in `progress-chart.js`.

mathjazz (approved):
Everything works as expected. I'll leave it to you to update the Progress Chart logic now or do it in a followup.

PR #2001, opened 2021-09-02
Persist dev env postgres data in a named volume

eemeli:
Right now, data retention effectively works because the `postgresql` container is mentioned in `webapp`'s depends entry in `docker-compose.yml`, and the `make run` command does not explicitly start it. This means that it's started in the background, and then stays running because it's never actually stopped. But it may go away when e.g. the developer's computer is restarted.

This adds a named `pgdata` volume which explicitly persists the data, and a `make dropdb` command for explicitly wiping the database when that's desirable.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -34,3 +34,9 @@ services:
       - POSTGRES_USER=pontoon
       - POSTGRES_PASSWORD=asdf
       - POSTGRES_DB=pontoon
+    volumes:
+      - pgdata:/var/lib/postgresql/data
+
+# Persist database
+volumes:
+  pgdata: null
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,7 @@ help:
 	@echo "  format           Runs formatters for both the frontend and Python code"
 	@echo "  types            Runs the tsc compiler to check TypeScript on all frontend code"
 	@echo "  eslint           Runs a code linter on the JavaScript code"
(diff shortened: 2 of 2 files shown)


jotes (approved):

docker-compose.yml
```diff
@@ -34,3 +34,9 @@ services:
       - POSTGRES_USER=pontoon
       - POSTGRES_PASSWORD=asdf
       - POSTGRES_DB=pontoon
+    volumes:
+      - pgdata:/var/lib/postgresql/data
+
+# Persist database
+volumes:
+  pgdata: null
```
Pike, line 42: Post-merge question, would it be a good idea to use a volume name that's version-specific?

With the data migration we're doing now to 13, would that help a local update?

eemeli: I think in any case you need to be aware of what's happening when applying a Postgres version update, so I'm not really sure that it would really help. Maybe reconsider this when applying the next update, which in any case is the next time when it could make sense to change the volume name?

Pike: Yeah. Background on that from the pontoon chat:

`docker-compose build` doesn't update the postgres container, that's only done once you `up`. At which point the new version of postgres is likely gonna try to do something with the old data and get folks and nervous and stuff. If at that point, it had a new volume, it'd just not find the old data, and you could import it.

It might also help to work on branches before and after the pg update. Even though the data won't sync, it might stick around? Not sure how graceful dc is about dangling volumes.

PR #2003, opened 2021-09-02
Fix bug 1728762: Update list of supported browsers

eemeli:
The current list has not been updated since #1116, at which point it was made to match the support list of fluent.js. That project has since (projectfluent/fluent.js#472) switched its support to target `ES2018` instead, and so is no longer a valid base for Pontoon to use.

This proposed list issomewhat conservative, but still well ahead of the previous one. It also supports practically all current visitors of https://pontoon.mozilla.org according to its Google Analytics data.

Updating the minimum support also allows dropping a couple of polyfills. Altogether these changes reduce the minzipped size of Pontoon's translate frontend assets by about 21 kB, i.e. about 4% of its entire prod bundle size.


mathjazz (approved):
LGTM

While we have 15% mobile users (as per GA), `/frontend` is unusable there.

eemeli:
Just to be clear, Pontoon should continue to work as it does currently with relatively modern mobile clients; there just won't be an explicitly defined minimum supported version for them.

Pike (approved):
LGTM, can you remove the link in line 75, too? The data is wrong, and it seems that the web-facing part of that is fubar.

PR #2008, opened 2021-09-03
Fix bug 1720875: Update Debian Buster → Bullseye, Postgres 11 → 13, Node.js 10 → 14

eemeli:
To persist a database across this change, dump & load it.

Filing this as a draft inititially, as I'm uncertain why (at least on my machine) the pip-installed lint requirements aren't included in the image, so e.g. `black` and `flake8` aren't available.

@Mitch4sho At least for me the update to Bullseye lets me drop the docker-compose `platform: linux/amd64` patch that was previously required on M1 macbooks.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.9-buster
+FROM python:3.9-bullseye
 
 ARG USER_ID=1000
 ARG GROUP_ID=1000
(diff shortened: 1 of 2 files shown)


eemeli:
It looks like the missing lint dependencies are an entirely separate issue, and @jotes said he might look at them. So I think this could be ready for review?

mathjazz (approved):
LGTM, thanks!

I tested locally and it works fine. (Just need to figure out why I get "pg_restore: [archiver] input file does not appear to be a valid archive" on `make loaddb`).

Let's wait for @Mitch4sho to confirm.

mathjazz:
Interesting, if I replace `pg_restore` with `psql` in `make dumpdb`(and drop the -O flag), it works.

Maybe the problem is that I made `make dumpdb` (while still on PG11), than `make build` (which installed PG13) and then `make loaddb`.

Mitch4sho:
Everything works well on my end with this system update for M1 👍

PR #2009, opened 2021-09-03
Fix bug 1713058: Only set target-language if not set

mathjazz:
The original bug report was to avoid updating values in the `target-language` attribute for non-iOS projects. As per the discussion in l10n-community on Matrix, we've decided to go a step further and never update the `target-language` attribute. We only set it, if it's not set yet.

### DIFF
--- a/pontoon/sync/tests/formats/test_xliff.py
+++ b/pontoon/sync/tests/formats/test_xliff.py
@@ -40,8 +40,7 @@ BASE_XLIFF_FILE = """
 
 XLIFF_TEMPLATE = """
 <xliff>
-    <file original="filename" source-language="en" datatype="plaintext"
-          target-language="{locale_code}">
+    <file original="filename" source-language="en" datatype="plaintext" {attr}>
         <body>
             {body}
         </body>
@@ -103,15 +102,17 @@ class XLIFFTests(FormatTestsMixin, TestCase):
                 )
             )
 
-    def generate_xliff(self, body, locale_code="en"):
-        return XLIFF_TEMPLATE.format(body=body, locale_code=locale_code)
+    def generate_xliff(self, body, attr=""):
+        return XLIFF_TEMPLATE.format(body=body, attr=attr)
 
     def test_save_basic(self):
         """
         Test saving changes to an entity with a single translation.
 
-        Also happens to test if the locale code is saved correctly.
+        Make sure the target-language attribute is set when not present.
         """
+        attr = f'target-language="{self.locale.code}"'
+
         input_string = self.generate_xliff(
             dedent(
                 """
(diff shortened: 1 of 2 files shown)


Pike (commented):
So, the intent is that if the file is tagged as `de`, but in the `fr` locale, it stays tagged as `de`?

Would force-sync change anything here?

Also, I had a hard time figuring out the differences between the tests, some more detailed comments would be good.

mathjazz:
We're solving two issues with this patch:

1. Pontoon doesn't set `target-language` attribute if it's not set, but our iOS projects require it.

2. Pontoon rewrites `target-language` attributes and in return [fights with automation](https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/pull/153/files).

Force-sync would not change anything. Simply changing the logic to set attributes does not trigger a commit - for that we need translation changes.

mathjazz:
I've updated the `test_save_basic` a little to make it easier to digest that it's not changing the attribute.

Only `test_save_target_language` is.

pontoon/sync/tests/formats/test_xliff.py
```diff
@@ -103,15 +102,49 @@ def test_parse_invalid_translation(self):
...
             )
 
-    def generate_xliff(self, body, locale_code="en"):
-        return XLIFF_TEMPLATE.format(body=body, locale_code=locale_code)
+    def generate_xliff(self, body, attr=""):
+        return XLIFF_TEMPLATE.format(body=body, attr=attr)
 
     def test_save_basic(self):
         """
         Test saving changes to an entity with a single translation.
```
Pike, line 110: Adjust the comment here to highlight that the `target-language` attribute exists?

Not sure, should this test actually test that we're not modifying the in-file locale code? That is, not use `self.locale.code`, at least not w/out a suffix?

pontoon/sync/tests/formats/test_xliff.py
```diff
@@ -103,15 +102,49 @@ def test_parse_invalid_translation(self):
...
+            ),
+            attr=attr,
+        )
+
+        super().run_save_basic(input_string, expected_string)
 
-        Also happens to test if the locale code is saved correctly.
+    def test_save_target_language(self):
+        """
+        Test if the target-language attribute is set.
```
Pike, line 144: Maybe: Test that the target-language attribute is set if not present.

Pike (changes requested):
I think this could use some better tests. And maybe a more thorough commit message?

Also, does this mean that we could drop https://github.com/mozilla/pontoon/blob/3f4565dcfd923048db16ca23f382cadc1d497efb/pontoon/sync/formats/xliff.py#L108 now?

mathjazz:
I've updated the tests a little bit by adding more explicit comments and flipping the state of the `target-language` attribute on the `input_string`.

We still need that [iOS locale code map](https://github.com/mozilla/pontoon/blob/3f4565dcfd923048db16ca23f382cadc1d497efb/pontoon/sync/formats/xliff.py#L108) for now (in case files for these locales are added by Pontoon), although @st3fan mentioned we might be able to use Pontoon-style locale codes in iOS projects soon. Not sure if he has any updates yet?

Pike (approved):
Thanks, LGTM.

PR #2314, opened 2021-09-20
Remove unused option from entities api

Pike:
This option used to be used by translate.old, but isn't used
in the translate frontend app.

Remove it to make the code a bit easier to reason about.


mathjazz (approved):
Thanks, nice catch!

Pike:
If I want to see that, I look into the mirror.

PR #2318, opened 2021-09-21
Fix #2313: Add ability to show project stats for a subset of teams

### LINKED ISSUE #2313: Add ability to show project stats for a subset of teams
Devs and PMs would like to see project stats for a subset of teams only, especially for projects we ship in several tiers of markets. This issue tracks the implementation of:

1. Ability to define subsets of teams in the project admin.
2. Ability to show project stats for each subset of teams.

In a followup, we'd also like to have the: 

1. Ability to switch between the subset(s) of teams and all teams in the project dashboard.
2. Reflect selected subset of locales in the Contributor and Insights tabs.

Mockup:
Screenshot 2021-09-20 at 15 55 18

mathjazz:
Deployed to stage:
https://mozilla-pontoon-staging.herokuapp.com/projects/firefox-for-ios/?teams=ar,zh-CN,fr,de,it,pl,sl

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -1617,6 +1617,16 @@ class ExternalResource(models.Model):
 
 
 class ProjectLocaleQuerySet(models.QuerySet):
+    def aggregated_stats(self):
+        return self.aggregate(
+            total_strings=Sum("total_strings"),
+            approved_strings=Sum("approved_strings"),
+            fuzzy_strings=Sum("fuzzy_strings"),
+            strings_with_errors=Sum("strings_with_errors"),
+            strings_with_warnings=Sum("strings_with_warnings"),
+            unreviewed_strings=Sum("unreviewed_strings"),
+        )
+
     def visible_for(self, user):
         """
         Filter project locales by the visibility of their projects.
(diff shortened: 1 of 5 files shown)


Delphine:
Tried it out and seems good to me, thank you.
I've played around using the URL too, and seems like both projects and locales can be updated via there, so that's really neat.

flodolo:
LGTM. What's up with the failing test?

shipit

mathjazz:
Thanks for testing! I've fixed the broken test.

@Pike, do you have capacity to review this?

pontoon/base/models.py
```diff
@@ -1617,6 +1617,16 @@ def __str__(self):
 
 
 class ProjectLocaleQuerySet(models.QuerySet):
+    def aggregated_stats(self):
```
Pike, line 1620: I was wondering if this belongs into a base class, but there seems to be only one other query set that implements this, and differently (huh? on the other one).

So probably not, or at least not as part of this patch.

Pike (approved):

pontoon/base/models.py
mathjazz: Yeah, we probably want to apply [something like this](https://gist.github.com/mathjazz/a4f0a2023cbb8e8ddebf665cd2325aa7), but I worry that will cause (even more) issues with stats calculation. The reason we use `total_strings` from the `Resource` model in the other method is probably because that value is different from the value in the TranslatedResource when the method is called during sync.

PR #2320, opened 2021-09-22
Fix #2155: Create portable Docker image

### LINKED ISSUE #2155: Create portable Docker image
## [Bug 1513104](https://bugzilla.mozilla.org/show_bug.cgi?id=1513104)
Bug Reporter: dane.molotok@gmail.com
CC:   @adngdb, @Pike, @eemeli, erickwilder@gmail.com, keunes@mailbox.org, @mathjazz, me@avm99963.com, mlzc@hey.com, santosh.prajapati@ibriz.ai

Pontoon is heavily tailored to be deployed on Heroku. With some massaging, the work done for the dev Docker container could be used to create a production-ready Docker image. This could be published on DockerHub and used by folks who want to run Pontoon in their existing infrastructure.

Freebios:
# Create a production-ready Docker image(s) (resolves #2155)
## Changes made

- Add of a production Docker image
- Tweak a small like in pontoon/translate/views.py

## Quick overview on performances gain

|                          | Development image | Production image                           |
|--------------------------|-------------------|--------------------------------------------|
| Size                     | 2.6GB             | 616MB                                      |
| Runtime memory footprint | ~800MB            | From 180 with the lowest amount of workers |

## Behavior

The dockerfile will first build the project with the bare minimum environment:
- DJANGO_DEV=False
- SECRET_KEY=somerandomstuff
- DATABASE_URL=randomURI

Then when running the Docker image you will have to inject some of the environment variables that you are used in dev.

Finally the Docker image will execute the entrypoint script containing a few things:
- A setup of SSH on the container
- Injection of $(env) into a .env file
- Database migration
- gunicorn launch
- and celery launch

Gunicorn and Celery are configurable through environment variables GUNICORN_WORKERS, GUNICORN_THREADS and CELERY_CONCURRENCY.
I would appreciate some insight on what default values I should apply here.

## Small file tweak
I've changed a thing in the `pontoon/translate/views.py` file.
While testing some stuff using the DEBUG env, I wasn't able to reach the frontend as it was trying to proxy it as in a dev environment. I guess this would make sense to change the variable used to DEV instead.

### DIFF
--- a/docker.production/Dockerfile
+++ b/docker.production/Dockerfile
@@ -1,0 +1,112 @@
+# Frontend build
+FROM node:14-bullseye-slim AS frontend
+WORKDIR /frontend
+COPY frontend/package.json frontend/yarn.lock ./
+RUN yarn install
+COPY frontend/ .
+RUN yarn build
+
+
+FROM python:3.9-slim-bullseye AS webapp_build
+
+ENV DEBIAN_FRONTEND=noninteractive
+ENV HGPYTHON3=1
+
+# Python environment variables
+ENV PYTHONUNBUFFERED 1
+ENV PYTHONDONTWRITEBYTECODE 1
+ENV PYTHONPATH /app
+
+# JavaScript applications paths
+ENV WEBPACK_BINARY /app/node_modules/.bin/webpack
+ENV YUGLIFY_BINARY /app/node_modules/.bin/yuglify
+ENV TERSER_BINARY /app/node_modules/.bin/terser
+
+# Django environment variables
+ENV DJANGO_DEV False
+ENV SECRET_KEY override_me_at_runtime
+ENV DATABASE_URL postgres://pontoon:asdf@postgresql/pontoon
+
+# Install required software.
+RUN apt-get update \
+    # Enable downloading packages over https
+    && apt-get install -y --no-install-recommends \
+        apt-transport-https \
+        curl \
+    # Get source for node.js
+    && curl -fsSL https://deb.nodesource.com/setup_14.x | bash - \
+    # Get source for yarn
+    && curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - \
+    && echo 'deb https://dl.yarnpkg.com/debian/ stable main' > /etc/apt/sources.list.d/yarn.list \
+    # Install software
+    && apt-get update \
+    && apt-get install -y --no-install-recommends \
+        build-essential \
+        libmemcached-dev \
+        nodejs \
+        postgresql-client \
+        postgresql-server-dev-13 \
+        yarn \
+    # Clean up what can be cleaned up.
+    && apt-get autoremove -y
+
+WORKDIR /app
+
+# Install Pontoon Python requirements
+COPY requirements/* /app/requirements/
+RUN pip install -U 'pip>=8' && \
+    pip install --no-cache-dir --require-hashes -r requirements/default.txt
+
+
+COPY package.json .
+COPY package-lock.json .
+RUN npm install
+
+COPY ./pontoon /app/pontoon
+COPY --from=frontend /frontend/build /app/frontend/build
+COPY webpack.config.js .babelrc manage.py setup.py /app/
+
+# Create the folder for front-end assets
+RUN mkdir -p assets
+# Run webpack to compile JS files
+RUN $WEBPACK_BINARY
+
+# Run collectstatic in container which puts files in the default place for
+# static files.
+RUN python manage.py collectstatic --noinput
+
+RUN rm -rf node_modules
+
+
+FROM python:3.9-slim-bullseye AS server
+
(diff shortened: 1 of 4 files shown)


mathjazz:
@Freebios Thanks for the patch! We'll get back to you early next week.

eemeli:
After a brief look, this looks interesting but might be a bit much to take on. I have a couple of questions that might help clear things up:
- Why replace some of the Django asset build with Gulp? Sure it works, but how is it better? The build is rather complicated already, so adding a whole new complex of dependencies requires some really good reasons.
- Does it really make sense to include the whole docker-compose production config, including a reverse proxy? I would expect that in practice this would end up sufficiently customised by any real-word users to make it rather ineffective.

It's likely that we'll be able to merge #1996 soon, which also changes the Docker config. As you rebase this PR on those changes, could you consider splitting it into smaller chunks? It'd be easier to have multiple specific discussions on smaller changes, rather than one wider-ranging discussion.

Freebios:
- Two reasons why creating a new asset Pipeline :
  - The first one, when trying to deploy an image where Django does the static assets serving, I ended with very slow page response time: 1+ minute to load a single page.
  - Second one, the Django asset pipeline relies heavily on Node binaries, leading to long (re)build time and heavy images.
  - => The solution that came to my mind was to treat assets separately only in a Node environment as it is almost already done with Django.
- I should not have included the docker-compose, I will remove it as it was mainly there for test purposes. It could be adapted for production but it's a single use-case among many others, so probably not worth it.

I will wait for the merge of #1996 and then, I can try to find a way to ditch the Gulp pipeline, use Django pipeline again while still reducing image size and serving assets through a CDN.

pontoon/administration/templates/admin.html
```diff
@@ -38,9 +38,9 @@
 {% endblock %}
 
 {% block extend_css %}
-  {% stylesheet 'admin' %}
+  <link href="{{ static('css/admin.min.css') }}" rel="stylesheet" type="text/css">
```
Pike, line 41: Do these changes drop the hashes from the css and js links?

If so, that'd be bad for cached resources in production. Even if we're having a deployment monolith, but more so if we'd go for a CDN.

docker.production/run_webapp.sh
```diff
@@ -0,0 +1,28 @@
...
+
+echo ">>> Setting up the db for Django"
+python manage.py migrate
+
+# syncing projects if env SYNC_INTERVAL is set, if it is set and you need to "work" on the bash, kill the process syncprojects.sh
+echo ">>> starting continuous syncing projects"
+/app/docker.production/syncprojects.sh &
+
+echo ">>> Starting local server"
+python manage.py runserver 0.0.0.0:8000
```
Pike, line 28: Please make sure to not use django's runserver in production. It's dangerous.

See `Procfile` for what we're using right now as production server.

That might also solve the performance issues you encountered.

Pike (changes requested):
I've spotted two problems with the current state of this PR, too.

They might also affect how you approach this.

mathjazz:
> I will wait for the merge of #1996 and then, I can try to find a way to ditch the Gulp pipeline, use Django pipeline again while still reducing image size and serving assets through a CDN.

Note that we just merged #1996.

pontoon/administration/templates/admin.html
Freebios: Hmmm, now that you bring this up, the static() method lookup into a map generated by the gulp pipeline but fallback on regular file path without hash when not using gulp, so I may have broken the previous asset serving.
Though I will check if this can be reverted when applying the suggested changes.

docker.production/run_webapp.sh
Freebios: I'll test performances again

mathjazz:
Thank you for your effort towards running Pontoon in production through Docker! While I strongly believe this is the right way forward in making Pontoon more approachable to users beyond Mozilla, I think it's best to pause the effort for the time being due to too many unknowns.

For more details why, please see my and Eemeli's thoughts from #pontoon from Oct 25:

---

@mathjazz

Sorry for waiting for so long for the review of your patch. We talked about it at the team call last Thursday. The challenge is, we currently don't use Docker in production and don't have immediate plans to do so, so we worry we'll break things as we go, without noticing.

On top of that, we're planning some changes to the (frontend) build process, which will likely require changes to Docker as well. We made an awful job of not making up our mind and communicating our position early on, but we think we should at least wait until the build process refactor before moving forward with a production-ready Docker image. Sorry about that!

---

@eemeli

- The prod and dev Docker builds should avoid duplicating config, so that e.g. the dev container works as the build stage for prod.
- It probably doesn't make sense to include all the services in a single container, so that e.g. the db and message queues can live outside it.
- It's okay to start with providing building blocks rather than a whole setup.
- Having actual users for the container would be great.

PR #2334, opened 2021-10-19
Set user-select: none for non-printing chars in original string

### LINKED ISSUE #2293: Pilcrow in message
## [Bug 1711736](https://bugzilla.mozilla.org/show_bug.cgi?id=1711736)
Bug Reporter: @clouserw
CC:   @Pike, @flodolo, @mathjazz, @peiying2

This was reported at https://github.com/mozilla/fxa/issues/8483 but it seems like a bug in Pontoon that this is getting passed through so I'm filing here.  We've run into it a couple of times.

> On the Firefox Accounts Profile page, the Secondary Email note contains a pilcrow where it should just be a space:
> 
> >   Note: a secondary email won’t restore your information — you’ll¶need a recovery key for that.
> 
> This message comes from the FTL for se-content-note, which looks like a correctly formatted line continuation according to the Fluent spec, so it's quite likely this is a bug in whatever Fluent implementation this page is using.
> Steps to reproduce
> 
> URL: https://accounts.firefox.com/settings#secondary-email
> 
> Add a secondary email to your account. Then open your account profile and look at the grey note in the Secondary email section.
> Expected result
> 
> The message should end with "you'll need a recovery key for that."
> Actual result
> 
> The message ends with "you’ll¶need a recovery key for that."
> Environment
> 
> Using Firefox 88, production Firefox Accounts.

eemeli:
This fixes #2293 by removing the original source of the problem. This does not fix any issues with existing corrupted message data.

Essentially, rather than doing complicated JavaScript magic around copy/paste events, let's instead never allow non-printing characters like `→` for tab, `¶` for newline or `·` for multiple spaces to be selected in the first place. Also, include the actual original source in their place.

This also means that we won't cause an issue for how to handle those characters when they're actually intentionally used in a message.

### DIFF
--- a/frontend/src/core/placeable/components/WithPlaceables.css
+++ b/frontend/src/core/placeable/components/WithPlaceables.css
@@ -7,3 +7,12 @@ mark.placeable {
     font-weight: normal;
     margin: 0 1px;
 }
+
+mark.placeable [aria-hidden='true'] {
+    user-select: none;
+}
+
+mark.placeable .hidden-source {
+    display: inline-block;
+    width: 0;
+}
--- a/frontend/src/core/placeable/parsers/multipleSpaces.tsx
+++ b/frontend/src/core/placeable/parsers/multipleSpaces.tsx
@@ -17,10 +17,8 @@ const multipleSpaces = {
                     title='Multiple spaces'
                     data-match={x}
                 >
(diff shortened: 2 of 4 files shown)


mathjazz:
Thanks, looks even better now!

Issue #2251, opened 2021-01-12
Warnings behave inconsistently when working with FTL string


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1686402](https://bugzilla.mozilla.org/show_bug.cgi?id=1686402)
Bug Reporter: @flodolo
CC:   iann_bugzilla@blueyonder.co.uk, @mathjazz
See also:     https://bugzilla.mozilla.org/show_bug.cgi?id=1618920

I can't totally figure out what's going, but this string should have a warning displayed (duplicate attribute, the error is also in the source).
https://pontoon.mozilla.org/en-GB/all-projects/all-resources/?status=warnings&string=206464

If I try to enter a new suggestion, I should get a warning that the attribute is duplicated, but it doesn't happen. The warning is displayed if I try to approve a suggestion with the error afterwards, or after I reject the suggestion.

Even worse: if I submit a suggestion, then switch to "submit" mode and submit another translation, the string gets approved without any warning.

PR #2348, opened 2021-10-25
Add missing-placeholder warnings for Android strings.xml & don't always ignore warnings

### LINKED ISSUE #2323: Android-l10n: checks for improper placeholders
Localizers often submit improperly written placeholders, which in turns impedes string import on mobile side.
We would like that Pontoon notify localizers of these errors, and so that they can not submit them.

Example of placeholders:
%1$s
%s
%d
%2$s
%1$d

eemeli:
Fixes #2323 
Related: #2251 

Keeping as a draft for now, as the required compare-locales version has not yet been published yet; see the upstream PR mozilla/compare-locales#3.

This work also revealed a bug that meant that `sendTranslation(ignoreWarnings)` was ~always getting called with a true-ish first argument (a React event object), so warnings were being ignored. That should be fixed now.


eemeli:
I got compare-locales 8.2.0 published & specified the EditorMainAction types a bit, so this is should now be ready for review.

mathjazz (approved):

Issue #2133, opened 2018-06-14
Obsolete Resources instead of deleting them


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1468840](https://bugzilla.mozilla.org/show_bug.cgi?id=1468840)
Bug Reporter: @mathjazz
CC:   @VishalCR7
Depends on:   [Bug 1473749](https://bugzilla.mozilla.org/show_bug.cgi?id=1473749)

When a source string is removed from the VCS, we mark corresponding Entity object as obsolete in the DB.

When a source file is removed from the VCS, we delete the corresponding Resource object from the DB.

That's inconsistent.

We should unify the behaviour by marking corresponding Resource objects as obsolete when source files are removed. One of the consequences will be more accurate contributor stats.

PR #2351, opened 2021-10-31
Bug 1468840: Obsolete Resources instead of deleting them

Pike:


mathjazz:
Closing the PR due to inactivity. #2133 is less of a priority since we implemented ActionLog, but it would be still nice to fix.

Issue #2322, opened 2021-09-27
Add support for key-value JSON


Freebios:
Currently, Pontoon only supports WebExtension JSON files.
Key-Value JSON files seem to be used by people (including me :eyes:), adding support for them would be nice.

## Implementation

As it's been pointed out by @mathjazz on Mozilla chat, the key-value files parser could be selected seamlessly alongside the WebExtension one. Indeed files in WebExtension must be named `messages.json`.

The only constraint to use the key-value parser would be to use a filename different than `messages.json`. Additionally, we could fall back on the key-value parser if the `messages.json` structure could not be validated, but I guess this would lead to UX issues in case of structure error in an actual WebExtension file.

12rambau:
Any idea when this PR will be merged to the master branch? I don't really like the idea of relying on a personal fork of the project and I think this could super useful for many projects (e.g. anything based on the Symfony framework).

BTW: thank you @Freebios for your work !

mathjazz:
Sorry @12rambau, it was me who dropped the ball on merging the PR. I just requested some changes there.

Freebios:
This can be closed now :)

mathjazz:
Thanks for the great work! 👍

Issue #2041, opened 2016-10-17
Add ability to write custom plugins for the custom formats of resource files.


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1310735](https://bugzilla.mozilla.org/show_bug.cgi?id=1310735)
Bug Reporter: @jotes
CC:   @mathjazz

Hi,

I want to prepare a demo site for The Piwik project (http://piwik.org). Unfortunately, piwik uses custom format (.json) for keeping source strings and translation.

And because of that I think that we should add a possiblity to write plugins for the new formats of files

I think we can achieve this via the setuptools mechanism for plugins: http://docs.pylonsproject.org/projects/pylons-webframework/en/latest/advanced_pylons/entry_points_and_plugins.html

mathjazz:
We're moving parsers and serializers to https://github.com/mozilla/moz-l10n/.

PR #2352, opened 2021-11-01
Add support for key-value JSON #2322 

Freebios:
# Key value JSON support #2322 
## Purpose of this PR
Add key value JSON support to Pontoon

## What changed
As there is already support for webextension json, I had to update the `formats` extension matching logic.
To do so, I've added support for fnmatch, allowing us to do full or partial match using wildcards and other operators.
This easily let us separate messages.json files and any other .json

## Key value json parser
The parser is based on webextension json parser. Because key-value json allow nested values, the json exploration is based recursive functions.


pontoon/sync/formats/json_keyvalue.py
```diff
@@ -0,0 +1,158 @@
+"""
+Parser for the .json translation format as used by the WebExtensions API:
+https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Internationalization
+
+See also:
+https://www.chromium.org/developers/design-documents/extensions/how-the-extension-system-works/i18n
```
eemeli, line 6: This is a remnant from the original `json_extensions.py` that should be updated, yes?

pontoon/sync/tests/formats/test_json_keyvalue.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class JsonKeyValueTests(FormatTestsMixin, TestCase):
+    parse = staticmethod(json_keyvalue.parse)
+    supports_keys = False
+    supports_source = False
+    supports_source_string = False
+
+    def key(self, source_string):
+        """JSON keys can't contain spaces."""
```
eemeli, line 28: Why not?

eemeli (commented):
Is the shape and structure of this "key-value JSON" format defined somewhere? Without a spec, it's rather hard to assess its implementation.

@mathjazz, should we consider making individual format handlers some sort of Pontoon plugins, so that they could be implemented and used without needing to apply changes to the core?

mathjazz:
> Is the shape and structure of this "key-value JSON" format defined somewhere? Without a spec, it's rather hard to assess its implementation.

We should validate it against schema, same as `json_extensions`.

> @mathjazz, should we consider making individual format handlers some sort of Pontoon plugins, so that they could be implemented and used without needing to apply changes to the core?

Yes :)
* https://github.com/mozilla/pontoon/issues/2041

pontoon/sync/formats/json_keyvalue.py
Freebios: Yes indeed

mathjazz:
Pytest is faling.

Please use `make pytest` to run the tests locally:
https://mozilla-pontoon.readthedocs.io/en/latest/dev/contributing.html#running-tests

Freebios:
Oh sorry, I was off last week.
I've changed extension compatibility into file name compatibility to differentiate *.json and messages.json, tests should pass now

eemeli (changes requested):
To clarify my earlier comment, this PR needs to make explicit reference to some external definition of this "key-value JSON" format. In order for Pontoon to support a file format, that file format needs a specification. Without that, it's not possible to assess the validity of its implementation.

It would also be very helpful if it could be shown that the file format is used relatively widely or is otherwise specifically relevant for Pontoon.

mathjazz:
FWIW, we had (have?) projects at Mozilla that used key-value JSON internally for storing translations, which had to be converted to some supported format and back. Firefox Accounts and Persona come to my mind.

I also found support for such format in Crowdin and Weblate:
https://support.crowdin.com/file-formats/json/
https://docs.weblate.org/en/latest/formats.html#json-files

eemeli:
@mathjazz Sure, but we still need a clear definition of the format. For instance, your Weblate link explicitly lists six different JSON formats that they support. Is this one of those, or something yet different? I can't tell.

mathjazz:
Right, I think what @Freebios has in mind is this:

```
{
  "Key1": "Source string 1",
  "Key2": "Source string 2",
  "Key3": "Source string 3"
}
```

Formal definition should be defined in the Schema.

pontoon/sync/formats/json_keyvalue.py
```diff
@@ -0,0 +1,165 @@
...
+    # Callback a function when reaching the end of a branch
+    def traverse_json(self, data, function, keys=[]):
+        for key, value in data.copy().items():
+            currentKey = keys.copy()
+            currentKey.append(key)
+            if isinstance(value, dict):
+                self.traverse_json(value, function, keys=currentKey)
+            elif type(value) == str:
+                dot_key = ".".join(currentKey)
+                function(dot_key, value)
```
eemeli, line 133: @mathjazz This would suggest that the file format isn't just flat, though?

mathjazz: @Freebios ? :)

Freebios:
Not sure why the black pipeline is failing.

About the definition of the "key-value" format, it seems that there is none available. Everyone who is referencing a spec is actually sending you back to the JSON spec, which is not helpful. Also, everyone seems to add their own features above the basic key-value mapping.

About who is implementing it, here are some more examples:
- https://docs.lokalise.com/en/articles/1400773-json-nested-json
- https://poeditor.com/localization/files/key-value-json
- https://help.phrase.com/help/nested-json (they seem to have included support for plurals here)

Onto who is using it, I could not find many libraries around this format, unfortunately.

Freebios:
I had in mind nest key-value format @mathjazz

pontoon/sync/formats/json_keyvalue.py
Freebios: correct

mathjazz:
Thanks for clarifying. In this case I suggest we document the format ourselves, possibly on top of the file, and include some examples.

Please rebase against master to fix black.

Freebios:
- I've added a short specification.
- I've also fixed a limitation with the naming of keys. Until now it wasn't possible to use dots in keys as they were internally used to represent nested values (the key would have been unwrapped when exporting).
I remembered that there is some funky usage of JSON where the key is the actual entire source string, including punctuation, so I've added support for it just in case.

I'm now using "<dot>" as a nesting separator in the internal keys stored in Pontoon, I did not risk myself using escape sequences (I'm not sure of the impact it would have of storing them in db). If you have a better idea I'm interested!

mathjazz:
Please fix `flake8` and `black`. :)

Freebios:
Sorry about this :sweat_smile:

pontoon/sync/formats/__init__.py
```diff
@@ -19,35 +21,41 @@
...
+
+
+def are_compatible_files(file_a, file_b):
     """
-    Return True if given file extensions belong to the same file format.
-    We test that by comparing parsers used by each file extenion.
+    Return True if five file names belong to the same file format.
+    We test that by comparing parsers used by each file extension.
     Note that some formats (e.g. Gettext, XLIFF) use multiple file extensions.
+    Some use fullname resolution
```
eemeli, line 52:
```suggestion
    Return True if the given file names correspond to the same file format.
    Note that some formats (e.g. Gettext, XLIFF) use multiple file name patterns.
```

eemeli (commented):
Do I understand right that at least the JSONResource code is mostly a copy of the pre-existing `json_extensions.py` code? If so, it might be a good idea to find a way to let these two parsers share some code, rather than duplicating it.

I'm not seeing any documentation updates here. Is there an expectation therefore of this happening separately?

I'm a bit concerned that the use of the string `<dot>` as a separator is setting things up for rather unexpected behaviour if that sequence ever shows up in an actual key. Is that actually revealed in the UI, btw? Or is it rendered there as an actual `.`?

It might be cleaner/safer to use `.` as a separator even internally, and explicitly mask either `"foo.bar": ...` or `"foo": { "bar": ... }` if both show up in the same file.

pontoon/sync/formats/__init__.py
```diff
@@ -19,35 +21,41 @@
...
+        if fnmatch.fnmatch(file_name, format):
+            return parser
+    return None
+
+
+def are_compatible_files(file_a, file_b):
     """
-    Return True if given file extensions belong to the same file format.
-    We test that by comparing parsers used by each file extenion.
+    Return True if five file names belong to the same file format.
```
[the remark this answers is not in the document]
Freebios: typo

Freebios:
I'll look into merging common code.

I'm using `.` in the context. The escape sequence never appears to the end-user.
My issue with using dots internally is when it comes to exporting back translations to a file. How do I choose between unwrapping keys to nested structure and keeping it as is?
I could somehow escape dots in keys rather than using a custom escape sequence.

pontoon/sync/formats/json_keyvalue.py
```diff
@@ -0,0 +1,184 @@
...
+    # Recursively read json object
+    # Callback a function when reaching the end of a branch
+    def traverse_json(self, data, function, keys=[]):
+        for key, value in data.copy().items():
+            currentKey = keys.copy()
+            currentKey.append(key)
+            if isinstance(value, dict):
+                self.traverse_json(value, function, keys=currentKey)
+            elif type(value) == str:
+                internal_key = "<dot>".join(currentKey)
```
eemeli, line 150: Rather than using a custom format like this, why not use a string list as the `internal_key`? So e.g. the key in `{ "foo": { "bar": "msg" } }` would get encoded as `['foo', 'bar']`, while the key in `{ "foo.bar": "msg" } }` would be `['foo.bar']`.

Freebios: I've implemented this solution using JSON.dumps and loads!

pontoon/sync/tests/formats/test_json_keyvalue.py
```diff
@@ -0,0 +1,124 @@
...
+            input_string,
+            expected_string,
+            source_string=input_string,
+        )
+
+    def test_save_dot_key(self):
+        input_string = dedent(
+            """
+            {
+              "Source.String": "Source String"
```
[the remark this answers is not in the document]
Freebios: Not really relevant anymore. I still implemented a test where I put some sort of JSON into a key.

Freebios:
We are getting somewhere :ok_hand: 
- I've changed the internal key format to use a JSON array, eg. ["my", "key.with.dots"]
- I've merged code where possible
- I've added some tests (to check multi-level nested keys, dot keys, JSON keys, ...)

While merging code, I've regrouped two way of opening a JSON file into a single one:
I basically replaced https://github.com/mozilla/pontoon/blob/3154e0cea7a34a39e3dc53446dda0fb428dea936/pontoon/sync/formats/json_extensions.py#L142-L148
with something similar to this https://github.com/mozilla/pontoon/blob/3154e0cea7a34a39e3dc53446dda0fb428dea936/pontoon/sync/formats/json_extensions.py#L104-L116

pontoon/sync/formats/json_keyvalue.py
```diff
@@ -0,0 +1,143 @@
...
+    "additionalProperties": False,
+}
+
+
+class JSONKVEntity(VCSTranslation):
+    """
+    Represents an entity in a Key Value JSON file.
+    """
+
+    def __init__(self, order, key, context, source_value, value):
```
eemeli, line 34: What's the `order` argument used for?

eemeli (approved):
Looks good!

Would be good to add a test or two to showcase how e.g. the following won't cause any conflicts:
```json
{
  "Source": { "String": "foo" },
  "Source.String": "bar"
  "[\"Source\", \"String\"]": "eek"
}
```

I also suspect that the JSON traversal ends up copying data a bit too much and the JSON load/dump of keys could be circumvented by using tuples as dict keys, but those can be addressed separately later if there's actual indication of them being any sort of bottlenecks.

mathjazz:
@Freebios Did you plan to add more tests here?

pontoon/sync/formats/json_keyvalue.py
Freebios: Fixed in the last commit

Freebios:
Here we are, I'll run some tests on my self hosted Pontoon to check if everything is in order and we should be good to go!

Freebios:
Reimporting my strings on several projects worked, export seems to work fine too 🙌

eemeli (approved):

pontoon/sync/formats/json_keyvalue.py
```diff
@@ -0,0 +1,144 @@
+"""
+Parser for key-value JSON, a nested Object structure of String values.
+This implementation does not support plurals.
+
+Each key can be associated with either a String or an Object value.
+Therefore, the format support nested values.
```
mathjazz, line 6: support > supports

mathjazz:
Sorry I dropped the ball on merging this patch.

I just took it for a spin with the [pontoon-test](https://github.com/mathjazz/pontoon-test) project and noticed that sync breaks with the following error:
```
[DEBUG:pontoon.sync.vcs.models] 2022-02-08 13:14:30,535 Parsing resource file: /app/media/projects/key-value-json/mathjazz/pontoon-test.git/sl/key_value.json
[ERROR:pontoon.base.errors] 2022-02-08 13:14:30,547 'NoneType' object has no attribute 'copy'
...
File "/app/pontoon/sync/formats/json_keyvalue.py", line 107, in traverse_json
    for key, value in data.copy().items():
AttributeError: 'NoneType' object has no attribute 'copy'
```

Looks like the problem is that we now expect the translated files to be available. @Freebios Could you please restore the original behavior?

Please also add an entry for the new file format to the [docs](https://github.com/mozilla/pontoon/blob/55579b9c20c6a2835835ecc03accb47053f31e2b/docs/index.rst#pontoon---mozillas-localization-platform).

pontoon/sync/formats/base_json_file.py
```diff
@@ -0,0 +1,56 @@
...
+                json_file = json.load(resource, object_pairs_hook=OrderedDict)
+                validate(json_file, SCHEMA)
+                return json_file
+
+        except (OSError, ValueError, ValidationError) as err:
+            # If the file doesn't exist or cannot be decoded,
+            # but we have a source resource,
+            # we can keep going, we'll just not have any translations.
+            if source_resource:
+                return
```
12rambau, line 33: if nothing is returned, the syncing process will crash along the way:

```
2022-02-09T13:22:27.591077+00:00 app[worker.1]: File "/app/pontoon/sync/formats/json_keyvalue.py", line 107, in traverse_json
2022-02-09T13:22:27.591078+00:00 app[worker.1]: for key, value in data.copy().items():
2022-02-09T13:22:27.591080+00:00 app[worker.1]: AttributeError: 'NoneType' object has no attribute 'copy'
```

Isn't returning an empty dict sufficient ? 
```python
except (OSError, ValueError, ValidationError) as err:
     # If the file doesn't exist or cannot be decoded,
     # but we have a source resource,
     # we can keep going, we'll just not have any translations.
         if source_resource:
             return {}

Freebios: Thank you :)
It seem that this will prevent the issue from happening on the web extension json parser too!

Freebios:
Should be good, sorry for the delay, this wasn't big changes to do :pray:

Freebios:
@mathjazz :wave:

mathjazz (approved):
Apologies for yet another delay!

This looks great and I can also confirm it to works fine.

Issue #2146, opened 2018-10-18
Mixed RTL and LTR content is hard to read in text inputs


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1500333](https://bugzilla.mozilla.org/show_bug.cgi?id=1500333)
Bug Reporter: Mahtab Alam [:alamM] &lt;<mahtab.ceh@gmail.com>&gt;
CC:   @amire80, @flodolo, @guerojeff, @mathjazz, myaseen.khan@hotmail.com, safa1996alfulaij@gmail.com, shahbazalam17feb@gmail.com
See also:     https://bugzilla.mozilla.org/show_bug.cgi?id=1602426

Created attachment 9018524
Screenshot (61).png

User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:62.0) Gecko/20100101 Firefox/62.0

Steps to reproduce:

For using XML Tag and External Argument as it is I clicked on it but it got mixed with one another.

Actual results:

Both got mixed with one another.

Expected results:

They should have remained separate.

Issue #2069, opened 2017-06-13
Mark placeables in translations editor


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1372861](https://bugzilla.mozilla.org/show_bug.cgi?id=1372861)
Bug Reporter: Reza Habibi (rehb) &lt;<reza.hbi@gmail.com>&gt;
CC:   @mathjazz, reza.hbi@gmail.com, safa1996alfulaij@gmail.com

User Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:53.0) Gecko/20100101 Firefox/53.0
Build ID: 20170524180630

Expected results:

Currently, following cases are highlighted after saving translate In Pontoon editor:
- newline character
- tab character
- multiple spaces
- leading and trailing spaces

And This is a great feature for the Pontoon.
I was wondering is it possible to display this feature real-time, before saving strings ( like Pootle)?

PR #2353, opened 2021-11-02
Fix #2267 Set dir attribute on placeables

### LINKED ISSUE #2267: Set direction on placeables
## [Bug 1692620](https://bugzilla.mozilla.org/show_bug.cgi?id=1692620)
Bug Reporter: @mathjazz

Created attachment 9202997
Screenshot

Everywhere we mark up placeables (i.e. everywhere we show strings in the translate view except in the textarea), we need to set the `dir` attribute to `ltr` in order to prevent bad display of placeables in RTL strings (as seen on the screenshot).

nick-zanetti:
This PR sets the dir attribute to ltr for placeable instances in parsers folder. This prevents bad display of placeables in RTL strings.

Anywhere there are placeables in the frontend app when viewing a translation for a RTL locale (for example Arabic) should now display correctly, i.e. `{$placeable}` instead of `{placeable$}`

This fixes #2267 

See [Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1692620) post for screenshot of bug.

### DIFF
--- a/frontend/src/core/placeable/parsers/altAttribute.tsx
+++ b/frontend/src/core/placeable/parsers/altAttribute.tsx
@@ -23,6 +23,7 @@ const altAttribute = {
                 <mark
                     className='placeable'
                     title="'alt' attribute inside XML tag"
+                    dir='ltr'
                 >
                     {x}
                 </mark>
--- a/frontend/src/core/placeable/parsers/camelCaseString.tsx
+++ b/frontend/src/core/placeable/parsers/camelCaseString.tsx
@@ -22,7 +22,7 @@ const camelCaseString = {
                 id='placeable-parser-camelCaseString'
                 attrs={{ title: true }}
             >
-                <mark className='placeable' title='Camel case string'>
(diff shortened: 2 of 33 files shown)


eemeli (approved):
I wonder if we could/should also apply this change to the `<textarea>` e.g. using Unicode RTL & LTR marks? OTOH, a proper solution would potentially be really rather complex and might fail rather horribly when it fails, so there's some risk here.

mathjazz:
> I wonder if we could/should also apply this change to the `<textarea>` e.g. using Unicode RTL & LTR marks? OTOH, a proper solution would potentially be really rather complex and might fail rather horribly when it fails, so there's some risk here.

I think we should. Related issues:
* https://github.com/mozilla/pontoon/issues/2146
* https://github.com/mozilla/pontoon/issues/2069

I was also thinking about a `<Placeable>`component to replace <mark> element (to avoid changing 33 files next time), but that could be done in a followup.

mathjazz:
Oh and BTW @nick-zanetti: welcome to Pontoon! 😄

PR #2358, opened 2021-11-10
Change SYSTRAN Translate URL

mathjazz:
SYSTRAN service currently returns 500 errors, because the endpoint URL, Credentials and Locale Profile IDs have been changed without prior notice.

This patch changes the URL. Credentials and Locale Profile IDs are obviously not stored in the codebase.

### DIFF
--- a/pontoon/machinery/views.py
+++ b/pontoon/machinery/views.py
@@ -187,9 +187,7 @@ def systran_translate(request):
             {"status": False, "message": f"Bad Request: {e}"}, status=400,
         )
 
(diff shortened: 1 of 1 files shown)


mathjazz:
Working on a fix for the black breakage.

eemeli (approved):
Aproving this at least while it only contains the actual URL change and doesn't also reformat 134 other files.

mathjazz:
This is OK to break on black!

The fix for that is in #2359.

PR #2359, opened 2021-11-10
Update black to 21.10b0

mathjazz:
Because the current version is old and more importantly breaks with the latest version of python used by GitHub automation, we should upgrade to latest black.

The changelog is big, mostly because arguments now break into new lines. I love that! :)

This will need a followup to ignore the commit id from git blame.


eemeli (approved):
I trust that this is changing just the code style.

PR #2360, opened 2021-11-10
Fix broken CSS on the Profile page

mathjazz:
CSS in the heading of the User Profile page is broken since #2349.

Let's fix it.

Screenshot 2021-11-10 at 14 25 05


eemeli (approved):
Code changes look fine. Have not checked the actual effect on appearance of the changes.

PR #2364, opened 2021-11-13
Upgrade to python 3.9.8

mathjazz:


Pike:
@mathjazz this doesn't run any CI? Should we get this tested somehow?

PS: I did the reopen dance to make sure we didn't have an infra snafu.

mathjazz:
This is a minor release, but in general it makes sense for this change to trigger CI.

Maybe run [pytest](https://github.com/mozilla/pontoon/blob/master/.github/workflows/backend.yml) with the python version from `runtime.txt`?

Issue #2190, opened 2019-10-28
Can switch from Rich editor to Simple via Source


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1592254](https://bugzilla.mozilla.org/show_bug.cgi?id=1592254)
Bug Reporter: @adngdb
CC:   @Pike, @mathjazz

It is possible to start with a string in the rich editor, edit it in the source editor, and get back to the simple editor when switching source mode off. I am not sure if that's a bug or a feature though.

## Steps to reproduce

1. Go to an FTL string with a value and a single attribute (e.g. https://pontoon.mozilla.org/translate/fr/firefox/devtools/client/application.ftl/?string=179386 ).
2. Clear the content of the value field.
3. Switch to Source mode (click the FTL button).
4. Switch back to Rich mode (click the FTL button again).

## Expected result

You get back to the rich editor with an empty Value field.

## Actual result

The Simple editor shows up, with only the content of the attribute.

mathjazz:
This issue has been fixed.

Issue #2173, opened 2019-07-16
Run checks when switching Fluent Editor mode


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1566476](https://bugzilla.mozilla.org/show_bug.cgi?id=1566476)
Bug Reporter: @adngdb
CC:   @mathjazz
Depends on:   [Bug 1501187](https://bugzilla.mozilla.org/show_bug.cgi?id=1501187)
Blocker for:  [Bug 1554705](https://bugzilla.mozilla.org/show_bug.cgi?id=1554705)

We need to run some checks before switching from a rich editor mode to the source mode and back, to ensure that no data the user has entered will be lost.

PR #2367, opened 2021-11-17
Flatten Fluent select messages more

### LINKED ISSUE #2366: When possible, select messages should only use top-level selectors
In the Unicode MessageFormat WG, one of the key aspects of making life simpler for localisers that we've identified is trying to present complete messages for translation whenever possible.

This seems like it might be a good idea to start doing in Pontoon as well. So, while Fluent supports message-internal selectors, we could (and should) start presenting such messages using top-level selectors. This doesn't work so cleanly for messages with more than one selector, so for those we could initially keep them as they are currently, and later consider more complex solutions, while keeping in mind compatibility with the evolving MF2 spec.

The first stage should not require any UI changes, it'd just allow localisers to have a cleaner experience with most select messages.

eemeli:
Closes #2366 

This refactors the behaviour of `flattenPatternElements()` to, well, flatten them some more. See the above issue for why this is a good idea.

If a Fluent message has a source like:
```fluent
msg = before { $foo ->
   [one] one
  *[other] other
} after
```
this will now be rendered and, if modified, saved as:
```fluent
msg = { $foo ->
   [one] before one after
  *[other] before other after
}
```

With the RichEditor, this means that for the above message a translator will see the fields `one` and `other`, rather than `Value`, `one`, `other`, `Value` as happens currently.

If a message has more than one selector, any whitespace immediately before each selector after the first will be kept out of the selector variants, so for instance this:
```fluent
my-entry =
    There { $num ->
        [one] is one email
       *[other] are { $num } emails
    } for { $awesome } { $gender ->
       *[masculine] him
        [feminine] her
    }.
```
gets parsed as:
```fluent
my-entry =
    { $num ->
        [one] There is one email for { $awesome }
       *[other] There are { $num } emails for { $awesome }
    } { $gender ->
       *[masculine] him.
        [feminine] her.
    }
```

There aren't all that many messages affected by this change; most of those are under Common Voice, e.g. `help-reach-hours-pluralized` and `activity-needed-calculation-plural`. For a number of languages, this change makes it clear that current translations are missing whitespace around the selector. Outside of Common Voice, only one locale (Czech) has many such constructs, see e.g. `toolkit/toolkit/global/resetProfile.ftl` in Firefox.

Editing the FTL source of a message to a different form is still fully supported. Switching from complex to rich editor will re-flatten the message according to the same logic.

The robustness of change detection in messages is slightly improved, and e.g. toggling rich -> source -> rich will now no longer always consider the message to have changed.

### DIFF
--- a/frontend/src/core/editor/hooks/useUpdateUnsavedChanges.ts
+++ b/frontend/src/core/editor/hooks/useUpdateUnsavedChanges.ts
@@ -19,16 +19,14 @@ export default function useUpdateUnsavedChanges(richEditor: boolean) {
 
     // When the translation or the initial translation changes, check for unsaved changes.
     React.useEffect(() => {
-        let exist;
-        if (richEditor) {
-            if (typeof translation === 'string') {
-                return;
-            }
+        let exist: boolean;
+        if (typeof translation === 'string') {
+            if (richEditor) return
+            exist = translation !== initialTranslation;
+        } else {
             exist =
                 typeof initialTranslation !== 'string' &&
                 !translation.equals(initialTranslation);
-        } else {
-            exist = translation !== initialTranslation;
         }
 
         if (exist !== unsavedChangesExist) {
--- a/frontend/src/core/editor/hooks/useUpdateUnsavedChanges.ts
+++ b/frontend/src/core/editor/hooks/useUpdateUnsavedChanges.ts
@@ -21,7 +21,7 @@ export default function useUpdateUnsavedChanges(richEditor: boolean) {
     React.useEffect(() => {
         let exist: boolean;
         if (typeof translation === 'string') {
-            if (richEditor) return
+            if (richEditor) return;
             exist = translation !== initialTranslation;
         } else {
             exist =
--- a/frontend/src/core/utils/fluent/convertSyntax.test.js
+++ b/frontend/src/core/utils/fluent/convertSyntax.test.js
@@ -27,7 +27,7 @@ describe('getComplexFromRich', () => {
 
         const res = getComplexFromRich(current, original, initial);
 
-        expect(res[1]).toEqual('title = Mien titre');
+        expect(res[1]).toEqual('title = Mien titre\n');
     });
 
     it('returns the correct initial translation when none exist', () => {
--- a/frontend/src/core/utils/fluent/convertSyntax.ts
+++ b/frontend/src/core/utils/fluent/convertSyntax.ts
@@ -59,7 +59,7 @@ export function getRichFromComplex(
     initial: string,
     locale: Locale,
 ): [Entry, Entry] {
-    let translationContent = parser.parseEntry(current);
+    let translationContent = flattenMessage(parser.parseEntry(current));
 
     // If the parsed content is invalid, create an empty message instead.
     // Note that this should be replaced with a check that prevents
@@ -90,18 +90,20 @@ export function getComplexFromRich(
     initial: string,
     locale: Locale,
 ): [string, string] {
-    let initialContent = initial;
-
     const translationContent = serializer.serializeEntry(current);
+
+    let initialEntry: Entry;
 
     // If there is no active translation (it's an untranslated string)
     // we make the initial translation an empty fluent message to avoid
     // showing unchanged content warnings.
-    if (!initialContent) {
-        initialContent = serializer.serializeEntry(
-            getEmptyMessage(parser.parseEntry(original), locale),
-        );
+    if (!initial) {
+        initialEntry = getEmptyMessage(parser.parseEntry(original), locale);
+    } else {
+        initialEntry = flattenMessage(parser.parseEntry(initial));
     }
+
+    const initialContent = serializer.serializeEntry(initialEntry);
 
     return [translationContent, initialContent];
 }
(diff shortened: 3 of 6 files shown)


mathjazz:
Nice work!

Everything works as expected, except the following error I get when I try to access the string with key `activity-needed-calculation-plural`. The strings works OK on master (if you can call that monster of a UI "OK").

```
Uncaught TypeError: can't access property "value", element is undefined
    renderVariant RichTranslationForm.tsx:373
    rendered_variants RichTranslationForm.tsx:431
    renderElements RichTranslationForm.tsx:430
    renderElements RichTranslationForm.tsx:418
    renderValue RichTranslationForm.tsx:472
    RichTranslationForm RichTranslationForm.tsx:499
    React 11
    unstable_runWithPriority scheduler.development.js:653
    React 4
    Redux 6
    routerMiddleware middleware.js:22
    dispatch Redux
    get actions.ts:62
    createThunkMiddleware Redux
    routerMiddleware middleware.js:22
    componentDidMount App.tsx:57
    React 6
    unstable_runWithPriority scheduler.development.js:653
    React 10
    tsx index.tsx:22
    tsx main.chunk.js:33676
    Webpack 7
RichTranslationForm.tsx:373
The above error occurred in the <RichTranslationForm> component:
    in RichTranslationForm (at RichEditor.tsx:78)
    in RichEditor (at FluentEditor.tsx:251)
    in FluentEditor (at EditorSelector.tsx:18)
    in div (at EditorSelector.tsx:17)
    in EditorSelector (at EntityDetails.tsx:482)
    in section (at EntityDetails.tsx:459)
    in section (at EntityDetails.tsx:458)
    in EntityDetailsBase (at EntityDetails.tsx:567)
    in EntityDetails (at App.tsx:127)
    in section (at App.tsx:125)
    in section (at App.tsx:120)
    in div (at App.tsx:103)
    in App (created by ConnectFunction)
    in ConnectFunction (at src/index.tsx:26)
    in LocalizationProvider (at AppLocalizationProvider.tsx:46)
    in AppLocalizationProviderBase (created by ConnectFunction)
    in ConnectFunction (at src/index.tsx:25)
    in Router (created by ConnectedRouter)
    in ConnectedRouter (created by Context.Consumer)
    in ConnectedRouterWithContext (created by ConnectFunction)
    in ConnectFunction (at src/index.tsx:24)
    in Provider (at src/index.tsx:23)

Consider adding an error boundary to your tree to customize error handling behavior.
Visit https://fb.me/react-error-boundaries to learn more about error boundaries.
```

eemeli:
Added a fix for that. The code in RichTranslationForm was assuming that a variant's elements array would always have at least one item, and that wasn't true here. So now it can handle that case, and we're avoiding generating it in the first place.

Playing around with this, I think there are a couple of deeper issues around the state changes between rich & complex editing of a message that should be addressed separately, if we care about them:
1. When the original message includes a selector and there isn't a translation available, `getEmptyMessage()` generates something matching the structure of the original, but with empty TextElements. Stringifying that does generate output that looks Fluent-ish, but it gets parsed as Junk on the way back in due to the empty cases.
2. The simple/rich/complex shape of the editor seems to depend on state that can end up mismatched e.g. if you start with a select message, go to edit its source, delete all the contents, and then go back to the simple/rich view.

mathjazz:
Thanks for the update, I can load the message now.

But I see an inconsistency in the way we present the source string (with the Value field) and the editor (without the Value field). I believe we should remove the Value everywhere, but I worry that makes strings as a whole harder to read than currently. What's your take?

Screenshot 2021-11-17 at 14 26 41

Additionally, we should impose max-height on the editor, otherwise the History panel is not accessible.

Regarding your 2nd point: #2190 and #2173 are related issues.

mathjazz:
Yeah, I'm fine if we keep this as it is for now, except for the inconsistency part. We should either show "Value" in both panels or in none of them.

eemeli:
Took me a little while to figure out what's happening here. In the original, the "Value" lines are showing up because the flattening is leaving the whitespace between the selectors outside them; they each have a value `' '`.

In the editor, an empty message is generated from the flattened original, and then this effect re-flattens it: https://github.com/mozilla/pontoon/blob/5aedad86771ccc051bed9d6bbe918fe2fae5596e/frontend/src/modules/fluenteditor/components/rich/RichTranslationForm.tsx#L160-L169

When that happens, the now-empty TextElements between the selectors are dropped because, well, they're empty. As a result, no corresponding Value fields are included in the editor.

@mathjazz Can we leave out that effect, or would doing so cause problems elsewhere?

mathjazz:
We use the [demo.ftl](https://github.com/mozilla-l10n/pontoon-ftl/blob/master/en-US/demo.ftl) file to test important changes to handling FTL files. If you don't notice the difference in it before or after the change (in untranslated (e.g. pl) and translated (e.g. sl) locale), we're likely safe.

eemeli:
@mathjazz Checked with the demo messages that removing that effect should have no ill effects.

frontend/src/core/editor/hooks/useUpdateUnsavedChanges.ts
```diff
@@ -19,16 +19,14 @@ export default function useUpdateUnsavedChanges(richEditor: boolean) {
...
-        let exist;
-        if (richEditor) {
-            if (typeof translation === 'string') {
-                return;
-            }
+        let exist: boolean;
+        if (typeof translation === 'string') {
+            if (richEditor) return;
+            exist = translation !== initialTranslation;
+        } else {
```
mathjazz, line 26: 👍

mathjazz (approved):
Nice work!

...and good first step towards MF 2.0 support. :)

PR #2370, opened 2021-11-23
Fix #2277 Folder format fix

### LINKED ISSUE #2277: Pontoon picks the wrong folder format for locales (underscore instead of hyphens)
## [Bug 1698636](https://bugzilla.mozilla.org/show_bug.cgi?id=1698636)
Bug Reporter: @flodolo
CC:   @mathjazz

I created the repository with the source in `en-US` folder.
https://github.com/mozilla-l10n/mozilla-vpn-website-l10n/

Pontoon then synced `pt-BR` as `pt_BR`. One possible explanation is the presence of a file with underscores (`CODE_OF_CONDUCT.md`) in the repository. If that's the case, it's a bug, because we really care about folders, not files.

I then renamed `pt_BR` to use hyphens, synced translations for `en-CA` and it still used the wrong format (`en_CA`). Fixed this one too, synced a string for `en-GB`,  and again wrong format.

At this point, I don't know if it will keep using _ for the next locales.

nick-zanetti:
This PR changes the `uses_underscore_as_separator` function to only count characters in folders, rather than files and folders. 

The PR also includes a unit test for the function inside test_utils.py. The test uses a dummy directory inside sync/tests to simulate the bug in #2277. 

This fixes #2277


nick-zanetti:
Changed this to WIP because it looks like pytest failed. When I tested locally from a shell using `app@...:/app$ pytest pontoon/sync/tests/test_utils.py` it passed, so not sure what would have caused that

pontoon/sync/tests/test_utils.py
```diff
@@ -0,0 +1,7 @@
+
+from pontoon.sync.utils import uses_undercore_as_separator
+dir = "pontoon/sync/tests/dummy_directory"
```
mathjazz, line 3: We need an absolute path rather than a relative one. I suggest you define it [along with other paths](https://github.com/mozilla/pontoon/blob/408655961d7577dc5e6d7b80c3ff258e6915682c/pontoon/sync/tests/__init__.py#L22) and then import it from `pontoon.sync.tests`.

mathjazz (commented):
Nice work!

Please use the following three commands to fix the linting and code style tests:
https://mozilla-pontoon.readthedocs.io/en/latest/dev/contributing.html#python-code-conventions

As for the failing unit test, the problem is that you use the relative path instead of the absolute one. I left the inline comment on how to fix that.

pontoon/sync/tests/test_utils.py
nick-zanetti: Thanks, made a new commit where I tried making a variable with `os.path.abspath()`, but unfortunately pytest failed again. Passes on my local and when I do `print(os.path.abspath("pontoon/sync/tests/dummy_directory"))` I get the correct absolute directory for my machine. 

Also I ran `make black` and 100+ files on my local branch updated, so I specifically only added the two files where I made changes for the issue rather than having all of them on there. Could that be causing that test to still fail?

mathjazz: I'm not sure abspath is the right function to use here, so to stay on the safe side, I'd do what other PATHs do, i.e.:
```python
DUMMY_DIRECTORY_PATH = os.path.join(
    os.path.dirname(__file__),
    "dummy_directory",
)
```

But more importantly - as of this patch `uses_undercore_as_separator()` only compares underscore and dash counts in directories, so you need to also add directories. Note that git cannot track empty directories, so you also need to put empty `.gitignore` files in each of them.

I'd also give `dummy_directory` a more meaningful name, e.g. `locale-separator-tests` (mind the dash instead of underscore for consistency with other folder names).

Once that's all done, you can run `make black`, but please note that we have updated the library recently, so you need to rebase your branch from upstream and run `make build-server` to install it first and only then run `make black`. Sorry for the inconvenience.

pontoon/sync/utils.py
```diff
@@ -50,8 +50,14 @@ def uses_undercore_as_separator(directory):
     """
     Return True if any subdirectory contains underscore.
```
mathjazz, line 51: We'll need to update the comment to reflect what the function actually does.

pontoon/sync/tests/test_utils.py
nick-zanetti: For the empty directories, did you mean to say add an empty `.gitkeep` in each? Saw [this](https://www.theserverside.com/blog/Coffee-Talk-Java-News-Stories-and-Opinions/gitkeep-push-empty-folders-git-commit) explainer on the issue and sounds like what I need to do. I assume adding empty `.gitignore` would make the directories still be ignored.

mathjazz: It doesn't matter how the file is called, this is just a workaround git's inability to track empty directories. See this explanation:
https://stackoverflow.com/a/7229996

I propose `.gitignore`, because we already use it elsewhere:
https://github.com/mozilla/pontoon/tree/master/media

mathjazz (approved):
Great work, thanks for the patch!

Issue #2262, opened 2021-02-09
Add Time to review suggestions chart to Community health dashboard


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1691770](https://bugzilla.mozilla.org/show_bug.cgi?id=1691770)
Bug Reporter: @mathjazz
Depends on:   [Bug 1687529](https://bugzilla.mozilla.org/show_bug.cgi?id=1687529)

This bug tracks the implementation of the "Time to review suggestions" chart and changes of the "Unreviewed suggestion lifespan" chart on the Community Health Dashboard.

See the spec:
https://github.com/mozilla/pontoon/blob/master/specs/0108-community-health-dashboard.md

Changes:
https://github.com/mozilla/pontoon/pull/1858

PR #2375, opened 2021-11-29
New copy for the Unreviewed suggestions insights chart

mathjazz:
This is a fix for the "change existing chart" part of #2262.

@flodolo could you please review, since it also changes the wording from the spec you wrote?

The implementation of the new chart will follow in a separate patch.

### DIFF
--- a/pontoon/insights/templates/insights/widgets/insights.html
+++ b/pontoon/insights/templates/insights/widgets/insights.html
@@ -51,9 +51,9 @@
         <section class="half">
             <div class="block clearfix">
                 <h3 class="controls">
-                    Unreviewed suggestions lifespan
+                    Age of unreviewed suggestions
                     {{ Tooltip.display(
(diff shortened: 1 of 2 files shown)


pontoon/insights/templates/insights/widgets/insights.html
```diff
@@ -51,9 +51,9 @@ <h3 class="controls">
         <section class="half">
             <div class="block clearfix">
                 <h3 class="controls">
-                    Unreviewed suggestions lifespan
+                    Age of unreviewed suggestions
                     {{ Tooltip.display(
-                        intro='How much time it takes on average to review a suggestion.',
+                        intro='How old are the unreviewed suggestions on average in particular point in time.',
```
flodolo, line 56: This sounds a bit off to me, I think is the more common expression "at any particular point in time", or "at a particular point in time"?

flodolo (approved):

PR #2376, opened 2021-11-29
Flatten Fluent message when copying from original

### LINKED ISSUE #2374: "Copy" does not copy variables for fluent
Take [this FTL string](https://pontoon.mozilla.org/eo/firefox/all-resources/?string=233939), the original is:

bildo

but after "Copy" this is shown (note that the variables are missing):

bildo

although switching to the FTL editor will show the whole text:

bildo

I'm running Firefox 94.0.2 on Linux. This happens, for me, with "Copy" for all FTL strings with variables. This is [another example](https://pontoon.mozilla.org/eo/firefox/all-resources/?string=192290).

eemeli:
Fixes #2374

It would appear that the hook discussed at https://github.com/mozilla/pontoon/pull/2367#issuecomment-972871389 was the only thing handling the message flattening when copying from the original while in the rich editor, and I'd missed it previously.

This is fixed, and I've gone through and reviewed that (hopefully!) all other code paths updating a Fluent translation are not similarly missing an otherwise expected flattening.

### DIFF
--- a/frontend/src/modules/fluenteditor/components/rich/RichEditor.tsx
+++ b/frontend/src/modules/fluenteditor/components/rich/RichEditor.tsx
@@ -47,7 +47,7 @@ export default function RichEditor(props: Props): React.ReactElement<any> {
             if (syntax !== 'rich') {
                 return;
             }
-            updateTranslation(message, changeSource);
+            updateTranslation(fluent.flattenMessage(message), changeSource);
         }
     }, [translation, changeSource, updateTranslation, dispatch]);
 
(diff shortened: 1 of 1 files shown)


mathjazz:
Thanks, that fixes the problem!

But I hit the same bug when I copy translations from the History and Locale panels. It seems to work fine when I copy from Translation Memory.

eemeli:
@mathjazz D'oh, you're right. There was another point at which the flattening was needed.

We really ought to refactor the translation state management at latest when start supporting MF2, so that we can lose the `string | Entry` typing for translations.

mathjazz (approved):
Thanks, I can't break it anymore. 😊

I agree about the refactor as much as I agree we should not do it before adding support for MF2.

PR #2380, opened 2021-12-01
Fix #2262 - Add Time to review suggestions chart

### LINKED ISSUE #2262: Add Time to review suggestions chart to Community health dashboard

mathjazz:
This is a 2nd and final part to fix #2262, which adds the Time to review suggestions chart to the locale Insights tab.

As explained in ones of the commit messages, I've joined Time to review suggestions and Age of unreviewed suggestions charts into one panel. The latter is not visible by default, but that should be OK, since it's probably the least frequently used widget on the page.

The alternative to using the same panel for both charts would be to introduce a new one, which would need to be full-width in order to keep the page layout, but that's an overkill for a chart with relatively low amount of data.

The patch is deployed to stage which uses the prod DB snapshot from yesterday:
https://mozilla-pontoon-staging.herokuapp.com/it/insights/

Data migration only took a minute or so to complete.

@flodolo Could you please take it for a spin, too? Especially from the spec, UX and data perspective. I've moved the Review activity chart right next to the Time to review suggestions chart to make it easier to check the validity of the data. We can move it back to the original place if you think that works better.


pontoon/insights/static/js/insights.js
```diff
@@ -183,6 +184,107 @@ var Pontoon = (function (my) {
...
+                var chart = $('#time-to-review-suggestions-chart');
+
+                new Chart(chart, {
+                    type: 'bar',
+                    data: {
+                        labels: $('#insights').data('dates'),
+                        datasets: [
+                            {
+                                type: 'line',
+                                label: 'Within a month',
```
flodolo, line 197:
```suggestion
                                label: 'Within this month',
```

Since the user is hovering a month, I think it makes more sense to say `this` than `a`.

pontoon/insights/templates/insights/widgets/insights.html
```diff
@@ -48,63 +47,59 @@ <h3 class="controls">
...
-                </canvas>
+                <div class="suggestions-age-items">
+                    <div class="suggestions-age-item time-to-review-suggestions">
+                        <h3 class="controls">
+                            Time to review suggestions
+                            {{ Tooltip.display(
+                                intro='How much time it takes on average to review a suggestion.',
+                                items=[{
+                                    'class': 'within-a-month',
+                                    'name': 'Within a month',
```
flodolo, line 60:
```suggestion
                                    'name': 'Within this month',
```

pontoon/insights/templates/insights/widgets/insights.html
```diff
@@ -48,63 +47,59 @@ <h3 class="controls">
...
+                                    'definition': 'Average age of suggestions reviewed during each month.',
+                                }, {
+                                    'class': 'twelve-month-average',
+                                    'name': '12-month average',
+                                    'definition': 'Average age of suggestions reviewed in the 12 months before each month.',
```
flodolo, line 65:
```suggestion
                                    'definition': 'Average age of suggestions reviewed in the 12 months before the selected month.',
```

Not sure if `selected` is the best choice, but `hovered` might be more confusing than helping. For this one, I'm fine also to keep as-is.

flodolo (approved):
Thanks, the order of graphs makes sense to me.

Looking at the graphs live, I only have a few suggestions on wording. Unrelated and pre-existing: would it be possible to dismiss the infobox by clicking anywhere, instead of having to click again on the i button?

pontoon/insights/templates/insights/widgets/insights.html
flodolo: "Current month" would work for me 👍🏼

pontoon/insights/templates/insights/widgets/insights.html
mathjazz: The spec says "specific", shall we stick to that? Not sure why I changed it in first place...

pontoon/insights/static/css/insights.css
```diff
@@ -109,6 +111,42 @@
...
+    right: 0;
+    margin-left: auto;
+    margin-right: auto;
+}
+
+#insights .suggestions-age nav ul li {
+    color: #aaaaaa;
+    cursor: pointer;
+    display: inline-block;
+}
```
eemeli, line 144: This toggle ends up _tiny_. If the intent is not to hide this toggle, it needs to be bigger. Could maybe use the same style as in the bigger activity charts?

pontoon/insights/static/js/insights.js
```diff
@@ -183,6 +191,107 @@ var Pontoon = (function (my) {
...
+                            callbacks: {
+                                label: function (items, chart) {
+                                    return (
+                                        chart.datasets[items.datasetIndex]
+                                            .label +
+                                        ': ' +
+                                        items.value +
+                                        ' days'
+                                    );
+                                },
```
eemeli, line 253:
```suggestion
                                label(items, chart) {
                                    const { label } =
                                        chart.datasets[items.datasetIndex];
                                    return `${label}: ${items.value} days`;
                                },
```

pontoon/insights/static/js/insights.js
```diff
@@ -183,6 +191,107 @@ var Pontoon = (function (my) {
...
+                                        display: false,
+                                    },
+                                    position: 'right',
+                                    ticks: {
+                                        beginAtZero: true,
+                                        maxTicksLimit: 3,
+                                        precision: 0,
+                                        callback: function (value) {
+                                            return value + ' days';
+                                        },
```
eemeli, line 287:
```suggestion
                                        callback: (value) => `${value} days`,
```

pontoon/insights/static/js/insights.js
```diff
@@ -637,3 +746,14 @@ var Pontoon = (function (my) {
...
+/* Main code */
+$(function () {
+    $('body').on('click', '#insights .suggestions-age nav li', function () {
+        var items = $('.suggestions-age nav li').removeClass('active');
+        $(this).addClass('active');
+        var index = items.index(this);
+        var itemWidth = $('.suggestions-age-item').first().outerWidth();
+        $('.suggestions-age-items').css('marginLeft', -index * itemWidth);
+    });
+});
```
eemeli, line 759:
```suggestion
$('#insights .suggestions-age nav li').on('click', function () {
    var items = $('.suggestions-age nav li').removeClass('active');
    $(this).addClass('active');
    var index = items.index(this);
    var itemWidth = $('.suggestions-age-item').first().outerWidth();
    // Show the selected graph view
    $('.suggestions-age-items').css('marginLeft', -index * itemWidth);
});
```

The bundle is included at the end of the page, so there's no need to add a `$()` wrapper. Also see other suggestions in the code snippet above.

eemeli (approved):
Looks good. Approved pending some fix to make the toggle bigger. JS suggestions are mostly really nits.

I presume we have no active plans for adding this info to the per-project views?

pontoon/insights/static/js/insights.js
mathjazz: Good point. I have to keep the click event handler on the body element, though, because the Insights tab DOM is not there yet when the handler get attached.

pontoon/insights/static/css/insights.css
mathjazz: In the activity charts we have legends, so I worry it'll be confusing to use the same style for a different functionality. Also, chart title are long and wouldn't fit into one line.

Unless you mean we should just makes the dots bigger to match the size of the activity charts?

eemeli: I don't think it'd be enough to just make the dots bigger. I think some sort of titles are needed. While there is a difference to the activity chart toggles, I don't think it's really _all_ that different?

mathjazz: Screenshot 2021-12-01 at 14 24 25

Something like that? It obviously needs fine-tuning (margin etc.).

mathjazz:
Stage updated with latest version.

pontoon/insights/static/css/insights.css
eemeli: Yes!

eemeli (approved):

PR #2381, opened 2021-12-01
Fix #2372 wrong docker-compose version number

### LINKED ISSUE #2372: Incorrect docker-compose.yaml version number
If you try to _docker-compose up -d_ or _make build_, docker will inform you that the **target:** property can not be used in that context. The specified docker-compose version is 2, but **target** property wasn't available until 2.4. I can confirm that after setting the version number to 2.4, the containers were recreated succesfully

Dazag:
Docker won't start up since the use of the property
target: and the mismatch with the version of the
docker-compose.yml (version: "2"). The minimum version
required to use target: is 2.3.

Changing the value of the version to 2.3 of the docker-compose.yml will
allow for the docker containers to start up.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 # docker-compose for Pontoon development.
 #
 # Note: Requires docker-compose 1.10+.
-version: "2"
+version: "2.3"
 services:
   frontend:
     build:


mathjazz (approved):
Thanks for the patch!

PR #2382, opened 2021-12-02
Add spacing to time range buttons

### LINKED ISSUE #2319: Labels overlap in DATE/TIME range picker
Schermata 2021-09-22 alle 11 02 29

Inspecting, the `transform="translate(80,0)"` seems to be at least 6 pixel (?) short. If I substract 6 pixel from the first two elements, it looks OK.

Schermata 2021-09-22 alle 11 21 50

nick-zanetti:
Fixes #2319 

The buttons at the bottom of the DATE/TIME range picker were having overlapping text. This adds spacing between them. Note that this adds spacing between _all_ of the buttons, not just the overlapping one from the issue, however it looks like we can't change the positioning of individual buttons with Highcharts, see [here](https://www.highcharts.com/docs/chart-design-and-style/style-by-css?_ga=2.63616175.1495014264.1638378248-1071920560.1638378248#what-can-be-styled).

### DIFF
--- a/frontend/src/modules/search/components/chart-options.ts
+++ b/frontend/src/modules/search/components/chart-options.ts
@@ -80,6 +80,7 @@ export const CHART_OPTIONS = {
             x: -4,
             y: 95,
         },
+        buttonSpacing: 20,
         buttonTheme: {
             fill: 'none',
             stroke: 'none',


eemeli (approved):
This looks like the right fix, given the limitations of the library we're using.

PR #2385, opened 2021-12-12
Some noImplicitAny TS fixes

Pike:
This is mostly about fixing selectors to work on the `RootState`.

This reduces the TS errors for `--noImplicitAny`, but increases `--strictNullChecks`. The culprit there is `getSelectedEntity`, which used to be `any` and is `Entity | undefined` now.

I wonder what state we're in when that's undefined.

I also made things just `undefined` instead of `null | undefined`

### DIFF
--- a/frontend/src/core/editor/selectors.ts
+++ b/frontend/src/core/editor/selectors.ts
@@ -33,8 +33,8 @@ export function _existingTranslation(
     ) {
         return activeTranslation;
     }
-    if (!history.translations.length) {
-        return;
+    if (history.translations.length === 0) {
+        return undefined;
     }
     // If translation is a FluentMessage, from the fluent editor.
     if (typeof translation !== 'string') {
--- a/frontend/src/core/editor/selectors.ts
+++ b/frontend/src/core/editor/selectors.ts
@@ -6,22 +6,21 @@ import { fluent } from 'core/utils';
 import * as history from 'modules/history';
 import { NAME } from '.';
 
-import type { EntityTranslation, Entity } from 'core/api';
-import type { HistoryTranslation, HistoryState } from 'modules/history';
-import type { EditorState } from '.';
+import type { RootState } from '../../store';
 
-const editorSelector = (state): EditorState => state[NAME];
-const historySelector = (state): HistoryState => state[history.NAME];
+const editorSelector = (state: RootState) => state[NAME];
+const historySelector = (state: RootState) => state[history.NAME];
 
 export function _existingTranslation(
-    editor: EditorState,
-    activeTranslation: EntityTranslation,
-    history: HistoryState,
-    entity: Entity,
-): (EntityTranslation | HistoryTranslation) | null | undefined {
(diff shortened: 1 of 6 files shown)


frontend/src/core/entities/selectors.ts
```diff
@@ -8,21 +8,22 @@ import { NAME } from '.';
...
-const userSelector = (state): UserState => state[user.NAME];
+const entitiesSelector = (state: RootState) => state[NAME].entities;
+const userSelector = (state: RootState) => state[user.NAME];
 
 export function _getSelectedEntity(
-    entities: Entities,
+    entities: ReturnType<typeof entitiesSelector>,
     params: NavigationParams,
-): Entity | null | undefined {
+): Entity | undefined {
```
Pike, line 19: I wonder what kinda state we're in if this doesn't have an Entity?

There's a bunch of code in the frontend that assumes this exists.

frontend/src/core/editor/selectors.ts
```diff
@@ -32,58 +31,51 @@ export function _existingTranslation(
...
-            const fluentTranslation = fluent.parser.parseEntry(
-                fluent.serializer.serializeEntry(translation),
-            );
-            existingTranslation = history.translations.find((t) =>
-                fluentTranslation.equals(fluent.parser.parseEntry(t.string)),
+        return activeTranslation;
+    }
+    if (!history.translations.length) {
+        return;
+    }
```
eemeli, line 38:
```suggestion
    if (history.translations.length === 0) {
        return undefined;
    }
```

It's better to be explicit about noting that we care about the return value, even if it's `undefined`.

frontend/src/core/editor/selectors.ts
```diff
@@ -6,22 +6,21 @@ import { fluent } from 'core/utils';
...
-    history: HistoryState,
-    entity: Entity,
-): (EntityTranslation | HistoryTranslation) | null | undefined {
+    editor: ReturnType<typeof editorSelector>,
+    activeTranslation: ReturnType<
+        typeof plural.selectors.getTranslationForSelectedEntity
+    >,
+    history: ReturnType<typeof historySelector>,
+    entity: ReturnType<typeof entities.selectors.getSelectedEntity>,
+) {
```
eemeli, line 21: Is the return type intentionally dropped here?

eemeli (approved):

frontend/src/core/editor/selectors.ts
Pike: Yes, TS is good about deducing it. And leaving out `null` and `undefined` should help when we transition to `strictNull`

frontend/src/core/entities/selectors.ts
eemeli: A very state. I _think_ it's unreachable unless the state or the store get horribly corrupted somehow.

Really this might be taken as a sign that we should deeply refactor the state handling to reflect this. Not sure if it's worth the trouble, though.

PR #2390, opened 2021-12-17
Update dependencies

mathjazz:
1. Update `lxml`.
2. Update indirect dependencies in `/frontend` within constraints of `/frontend/package.json`.
3. Update indirect frontend dependencies in `/` within constraints of `package.json`.


Pike (approved):
lgtm

Pike (changes requested):
Actually, I think there's a bustage in the frontend, which also breaks the python tests as it needs to build it for the py tests to run.

Pike (approved):
lgtm now. There's one new strict error, I think in src/core/api/base.ts, but I have no clue as to why that would be. Also, it's unclear if it's one new, I only saw that the strict count went up by one, and that one is new. There might be more changes.

mathjazz:
Nice catch with that strict error! Yeah, let's address this separately.

Issue #2270, opened 2021-02-25
Implement infinite scrolling in concordance search


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1695231](https://bugzilla.mozilla.org/show_bug.cgi?id=1695231)
Bug Reporter: @mathjazz

In https://github.com/mozilla/pontoon/pull/1794 we dropped the idea of using react-infinite-scroller (used in string list), because it's buggy.

We should create our own solution and in a followup also extend it to be used in the string list (i.e. drop react-infinite-scroller completely).

PR #2391, opened 2021-12-20
Refactor Machinery.tsx to functional

### LINKED ISSUE #2388: Concordance search text disappears when switching Machinery/Locales tabs
If you enter text in the Concordance search input field, switch to the Locales tab and back to the Machinery tab, the input field is empty again.

abowler2:
In looking into options for the issue to [Implement infinite scrolling in concordance search](#2270) I kept finding solutions suggesting the implementation of a hook to handle this. This was an approach that I had worked on when I first attempted to address changes to the search functionality, but it was beginning to take too much time to find a working solution so we opted to use a load button and circle back around to it.

At that time I had also found [an existing hook solution](https://github.com/onderonur/react-infinite-scroll-hook) that I attempted to utilize, but when I tried it there were functionality issues when using it in the Concordance search.  That hook has since been updated, and the issues that were present appear to have been addressed.

In preparation for either using the above mentioned hook or creating my own, I would need to convert the `Machinery.tsx` file to a functional component.  This PR makes that change, and as a byproduct it resolves #2388

### DIFF
--- a/frontend/src/modules/machinery/components/Machinery.tsx
+++ b/frontend/src/modules/machinery/components/Machinery.tsx
@@ -5,19 +5,16 @@ import './Machinery.css';
 
 import { Translation } from './Translation';
 import { SkeletonLoader } from 'core/loaders';
+import { usePrevious } from 'hooks/usePrevious';
 
 import type { Locale } from 'core/locale';
 import type { MachineryState } from '..';
 
-type Props = {
+interface Props {
     locale: Locale | null | undefined;
     machinery: MachineryState;
     searchMachinery: (query: string, page?: number) => void;
-};
-
-type State = {
-    page: number;
-};
+}
 
 /**
  * Show translations from machines.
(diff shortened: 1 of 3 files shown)


frontend/src/modules/machinery/components/Machinery.tsx
```diff
@@ -26,142 +23,128 @@ type State = {
...
 
-    handleResetSearch: () => void = () => {
-        this.props.searchMachinery('');
-        this.setState({ page: 1 });
+    React.useEffect(() => {
+        // Fetch next page of search results
+        if (prevPage && page !== prevPage) {
+            searchMachinery(searchInput.current.value, page);
+        }
+    }, [page]);
```
eemeli, line 57: Do we need to care about running this hook if `searchMachinery` changes?

frontend/src/modules/machinery/components/Machinery.tsx
```diff
@@ -26,142 +23,128 @@ type State = {
...
+
+    const handleResetSearch = () => {
```
eemeli, line 59:
```suggestion
    const resetSearch = () => {
```

"reset" is a perfectly fine verb to use here.

frontend/src/modules/machinery/components/Machinery.tsx
```diff
@@ -26,142 +23,128 @@ type State = {
...
+    const sourceString = machinery.searchString || machinery.sourceString;
+
+    return (
+        <section className='machinery'>
+            <div className='search-wrapper clearfix'>
+                <label htmlFor='machinery-search'>
+                    {machinery.searchString ? (
+                        <button
+                            className='fa fa-times'
+                            onClick={handleResetSearch}
```
eemeli, line 88:
```suggestion
                            onClick={resetSearch}
```

frontend/src/modules/machinery/components/Machinery.tsx
```diff
@@ -26,142 +23,128 @@ type State = {
...
-            (state) => ({ page: state.page + 1 }),
-            () =>
-                this.props.searchMachinery(
-                    this.searchInput.current.value,
-                    this.state.page,
-                ),
-        );
+    const getMoreResults = () => {
+        setPage(page + 1);
     };
```
eemeli, line 72:
```suggestion
    const getMoreResults = () => setPage(page => page + 1);
```

A bit clearer about what's being done, and less liable to break later in odd ways. Tbh, I'd probably also just inline this directly in the prop value as the function name adds very little context to just reading its body.

eemeli (approved):
Looks good! A few nitpicky mostly stylistic comments/questions below.

frontend/src/modules/machinery/components/Machinery.tsx
abowler2: Yeah, we should probably include it to avoid any potential bugs down the line.

frontend/src/modules/machinery/components/Machinery.tsx
```diff
@@ -26,142 +23,125 @@ type State = {
...
-            (state) => ({ page: state.page + 1 }),
-            () =>
-                this.props.searchMachinery(
-                    this.searchInput.current.value,
-                    this.state.page,
-                ),
-        );
+    const submitForm = (event: React.SyntheticEvent<HTMLFormElement>) => {
+        event.preventDefault();
+        setPage(1)
```
abowler2, line 67: In working on #2394 I discovered that if a user performs a new search without resetting the input or changing entities then the page needs to be reset in order to get accurate results.

PR #2393, opened 2021-12-21
Avoid resetting errors & warning during editor mount

### LINKED ISSUE #2251: Warnings behave inconsistently when working with FTL string

eemeli:
Fixes #2251 by adding an unsaved-changes check to the failed-checks reset that's meant to only be called during editing, rather than also applying when switching between different strings.

This is a little bit of a hack as it relies on `state.unsavedchanges.exist` to change after `state.editor.translation` changes in order to handle the corner cases of going from original -> broken -> original. But 1) it works, and 2) a truly proper fix would have me refactor editor state handling completely, and that's a big enough job that we'll need to prioritise it appropriately.

### DIFF
--- a/frontend/src/modules/entitieslist/components/EntitiesList.tsx
+++ b/frontend/src/modules/entitieslist/components/EntitiesList.tsx
@@ -199,29 +199,32 @@ export class EntitiesListBase extends React.Component<InternalProps> {
     ) => {
         const { dispatch, parameters, router, store } = this.props;
 
+        // Do not re-select already selected entity
+        if (entity.pk === parameters.entity) {
+            return;
+        }
+
         const state = store.getState();
         const unsavedChangesExist = state[unsavedchanges.NAME].exist;
         const unsavedChangesIgnored = state[unsavedchanges.NAME].ignored;
 
-        if (entity.pk !== parameters.entity) {
-            dispatch(
-                unsavedchanges.actions.check(
-                    unsavedChangesExist,
-                    unsavedChangesIgnored,
-                    () => {
-                        dispatch(batchactions.actions.resetSelection());
-                        dispatch(editor.actions.reset());
-                        dispatch(
-                            navigation.actions.updateEntity(
-                                router,
-                                entity.pk.toString(),
-                                replaceHistory,
-                            ),
-                        );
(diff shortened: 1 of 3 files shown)


mathjazz:
Nice work! We need something similar in the `RichTranslationForm` component to catch cases like this:
https://pontoon.mozilla.org/fr/all-projects/all-resources/?status=warnings&string=178286

The only other issue I found is that when I click on the string in the string list for the second time, warning disappears.

eemeli:
Changes rebased to the current `master`, and expanded to also cover `RichTranslationForm`.

To avoid the disappearing warning when re-selecting an entity from the list, that re-selection is now detected, and nothing happens then.

frontend/src/modules/entitieslist/components/EntitiesList.tsx
```diff
@@ -197,29 +197,31 @@ export class EntitiesListBase extends React.Component<InternalProps> {
...
-                        navigation.actions.updateEntity(
-                            router,
-                            entity.pk.toString(),
-                            replaceHistory,
-                        ),
-                    );
-                },
-            ),
-        );
+        if (entity.pk !== parameters.entity) {
```
mathjazz, line 206: Can you invert the condition and return when true? Then you can restore the original (already pretty deep) indentation of the following block.

mathjazz (approved):

PR #2396, opened 2022-01-03
More TypeScript fixes

Pike:
This is a PR for more general TS fixups.

Some test helpers go back to JS. And then a ton of `AppDispatch` instead of a catch-all function in actions.js.

I added a store hook and type for use in `unused`. I'm not particularly fond of that code, but it's perf optimizations from 1be742e2170a928a9f70ae9b87618a689def1e21. Maybe there's a good fix, but once more, I'd do that outside of this patch.

As for the various drops of return types for the functions returning functions, they're well typed by TS internally by adding the type information to the argument itself.


abowler2 (approved):
This looks good.

mathjazz (approved):
I trust @abowler2 :)

Also, LGTM.

PR #2397, opened 2022-01-05
refactor pinned comments to be only in pinned comment section

### LINKED ISSUE #2266: Only list pinned comments under all comments where posted
## [Bug 1692619](https://bugzilla.mozilla.org/show_bug.cgi?id=1692619)
Bug Reporter: @mathjazz
CC:   @flodolo

Created attachment 9202996
Screenshot

We list pinned comments in the Pinned comments and All comments sections. That works fine in the translate view of the locale the pinned comment has been posted in, but not in translate views of other locales.

See screenshot for details.

Mitch4sho:
Fixes #2266 this PR is supposed to set pinned comments in the pinned section only; this is illustrated in the screenshot below.

Screen Shot 2022-01-05 at 5 27 13 AM

### DIFF
--- a/frontend/src/core/comments/components/CommentsList.tsx
+++ b/frontend/src/core/comments/components/CommentsList.tsx
@@ -37,6 +37,7 @@ export default function CommentsList(props: Props): React.ReactElement<'div'> {
         resetContactPerson,
     } = props;
 
+    console.log({ user });
     const translationId = translation ? translation.pk : null;
 
     // rendering comment
@@ -53,6 +54,10 @@ export default function CommentsList(props: Props): React.ReactElement<'div'> {
 
     const pinnedComments = comments.filter((comment) => comment.pinned);
     const unpinnedComments = comments.filter((comment) => !comment.pinned);
+    const produceAllComments =
+        !user.isAuthenticated &&
+        unpinnedComments.length === 0 &&
+        pinnedComments.length;
 
     return (
         <div className='comments-list'>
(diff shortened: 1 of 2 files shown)


mathjazz:
Nice work Mitch!

Please make sure we don't render the All Comments title and All Comments panel if the user cannot add a comment and we only have pinned comments. Otherwise we'll render a blank panel with a title, which looks odd.

Mitch4sho:
I'm ready for the name changes 👍🏾

frontend/src/core/comments/components/CommentsList.tsx
```diff
@@ -67,13 +67,13 @@ export default function CommentsList(props: Props): React.ReactElement<'div'> {
...
-
-                    <Localized id='comments-CommentsList--all-comments'>
-                        <h2 className='title'>ALL COMMENTS</h2>
-                    </Localized>
                 </section>
             ) : null}
             <section className='all-comments'>
+                <Localized id='comments-CommentsList--all-comments'>
+                    <h2 className='title'>ALL COMMENTS</h2>
+                </Localized>
```
mathjazz: This will render the All Comments title even when we don't have any Pinned comments, which is the case most of the time. I don't think we need this change, so I'd revert.

frontend/src/core/comments/components/CommentsList.tsx
```diff
@@ -37,6 +37,7 @@ export default function CommentsList(props: Props): React.ReactElement<'div'> {
         resetContactPerson,
     } = props;
 
+    console.log({ user });
```
mathjazz, line 40: Please remove.

frontend/src/core/comments/components/CommentsList.tsx
```diff
@@ -53,6 +54,10 @@ export default function CommentsList(props: Props): React.ReactElement<'div'> {
 
     const pinnedComments = comments.filter((comment) => comment.pinned);
     const unpinnedComments = comments.filter((comment) => !comment.pinned);
+    const produceAllComments =
+        !user.isAuthenticated &&
```
mathjazz, line 58: WEeshould instead use `canComment` here

frontend/src/core/comments/components/CommentsList.tsx
Mitch4sho: I'm a bit confused on this, 

1. when the user is signed in, we want the All Comments title and section where they can comment regardless of whether there are Pinned comments or not?

2. when there is no user signed in, we want the All Comment title and section to be hidden when there are only Pinned Comments. When there are no pinned comments and no comments, All Comment titles and sections are hidden?

mathjazz: 1. Only the section (because `<AddComment>`is within it). We only add titles when we have Pinned comments.

2. Yes and yes.

Everything needs to be stay the same as it was in your original patch, except what I wrote in my comment above: "Don't render the All Comments title and All Comments panel if the user cannot add a comment and we only have pinned comments.".

frontend/src/core/comments/components/CommentsList.tsx
```diff
@@ -66,14 +69,17 @@ export default function CommentsList(props: Props): React.ReactElement<'div'> {
                             renderComment(comment),
                         )}
                     </ul>
-
-                    <Localized id='comments-CommentsList--all-comments'>
-                        <h2 className='title'>ALL COMMENTS</h2>
-                    </Localized>
+                    {!produceAllComments ? (
```
mathjazz, line 72: Seems like we "produce" All Comments if `produceAllComments` is **False**? We should either rename the flag to reflect that (hide instead of produce?) or invert the logic of `produceAllComments`.

mathjazz (commented):
Works great! The only thing I'd change is semantics, see below.

frontend/src/core/comments/components/CommentsList.tsx
Mitch4sho: got it!

Mitch4sho:
Let me know if `hideComments` is a good name, or should it be `hideAllComments`

mathjazz:
Indeed, `hideAllComments` is the most accurate choice here.

Mitch4sho:
alright, should be good to go 👍🏾  @mathjazz

mathjazz (approved):
Nice job! :)

PR #2398, opened 2022-01-05
Refactor linkify to have a single entry point and integration tests

Pike:


Pike:
@abowler2 what do you think about this idea? Do you think the tests are worthwhile?

I didn't actually hook up this wrapper, nor did I factor our the screenshots parser, which I'd also put into this.

mathjazz:
I like the idea and tests look good to me!

abowler2:
I like the idea of pulling this out so that it can be tested on its own, and I do think the tests are good as well.

The only question I have is regarding your comment on the Screenshots parser as that uses `linkify-it`, are you thinking you would do the same thing with that library, convert to use `react-linkify` or ????

Pike:
I'd pull in at least https://github.com/mozilla/pontoon/blob/d3f4b0db5ace30493638b32c8450b1a92a83f3c9/frontend/src/modules/entitydetails/components/Screenshots.tsx#L7-L9, not exactly sure if it'd make sense to add more.

I'd keep that out of react, intending to replace it with https://linkify.js.org/docs/linkifyjs.html#linkifyfind-str--type in a latter step.

mathjazz (commented):
Nice work!

There are a few other places in the code that import Linkify from `'react-linkify'`. Shouldn't they now import from `'core/linkify'`?

Pike:
> Shouldn't they now import from `'core/linkify'`?

Indeed they should.

mathjazz (approved):

PR #2399, opened 2022-01-06
Fix bug 1748145: Temporarily remove ability to change contributor email address

mathjazz:
Details in the bug.

### DIFF
--- a/pontoon/base/forms.py
+++ b/pontoon/base/forms.py
@@ -201,26 +201,10 @@ class UserProfileForm(forms.ModelForm):
         max_length=30,
         strip=True,
     )
-    email = forms.EmailField(
-        help_text=(
-            "Changing your email address will cause a logout. "
-            "Make sure the new one is correct before saving!"
-        )
-    )
 
     class Meta:
         model = User
-        fields = ("first_name", "email")
-
-    def clean_email(self):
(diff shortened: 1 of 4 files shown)


eemeli (approved):

avm99963:
The bug is private, is this related to a security vulnerability?

I'm asking because as a happy downstream user of Pontoon I'd like to know whether I should apply this patch or the issue is related to the Mozilla deployment. I told a while back my l10n contributors to change their email addresses if they don't want to share it publicly since it is included in Git commits.

Thanks in advance, and thanks for developing such good software! :) I don't have a lot of experience with Python, but if there's anything you might need a hand with, I'd be happy to contribute to Pontoon and also learn some Python on the way!

mathjazz:
@avm99963 Thanks for reaching out and for the kind words!

Indeed, this patch addresses a security issue. I've filed an issue to establish a proper release process for Pontoon, which should make it easier to follow updates and keep track of the changelog:
https://github.com/mozilla/pontoon/issues/2402

I've also published a post on Discourse about the fix:
https://discourse.mozilla.org/t/security-fix-changing-email-address-temporarily-disabled/91386

avm99963:
Thanks @mathjazz! I've just subscribed to the Pontoon category.

PR #2409, opened 2022-01-15
Do not overflow dashboard tabs into new line

mathjazz:
On https://pontoon.mozilla.org/fr/firefox/contributors/, this happens:

Screenshot 2022-01-15 at 22 27 25

This patch reduces the tab width to prevent the above ugliness, and prevents dashboard tabs from overflowing into a new line in general.


eemeli (approved):
If you merge this before the Python upgrade, its diff should get cleaned up automagically.

PR #2394, opened 2021-12-21
Add infinite scroll hook to Machinery tab

abowler2:
Resolves: #2270 

This PR replaces the `Load more` button used for pagination in the Machinery tab with an infinite scroll utilizing a [React hook](https://github.com/onderonur/react-infinite-scroll-hook).

PR #2413, opened 2022-01-24
#2412: Refactor EntitiesList.tsx to a function component

### LINKED ISSUE #2412: Refactor EntitiesList.tsx to a function component
Motivation is the same as with https://github.com/mozilla/pontoon/pull/2391.

Switching to https://github.com/onderonur/react-infinite-scroll-hook would let us unify the infinite scroll code with the one used for Concordance search, which was introduced in https://github.com/mozilla/pontoon/pull/2394.

yaya-usman:
**What changed with this PR:**

Fixes #2412 by Refactoring the EntitiesList Component from a class-based component to a functional component and a little replacement of deprecated syntax or properties

### DIFF
--- a/frontend/src/modules/entitieslist/components/EntitiesList.tsx
+++ b/frontend/src/modules/entitieslist/components/EntitiesList.tsx
@@ -1,4 +1,4 @@
-import * as React from 'react';
+import React, { useRef, useEffect } from 'react';
 import InfiniteScroll from 'react-infinite-scroller';
 
 import './EntitiesList.css';
@@ -22,6 +22,7 @@ import type { Entity as EntityType } from 'core/api';
 import type { EntitiesState } from 'core/entities';
 import type { Locale } from 'core/locale';
 import type { NavigationParams } from 'core/navigation';
+import { usePrevious } from 'hooks/usePrevious';
 
 type Props = {
     batchactions: BatchActionsState;
@@ -45,91 +46,78 @@ type InternalProps = Props & {
  * entities. It interacts with `core/navigation` when an entity is selected.
  *
  */
-export class EntitiesListBase extends React.Component<InternalProps> {
-    list: { current: any };
+export const EntitiesListBase: React.FC<InternalProps> = (
+    props,
+): null | React.ReactElement<'div'> => {
+    const list = useRef<any>();
+    const isFirstRendered = useRef<boolean>(true);
+    const entitiesPrev = usePrevious(props.entities.entities);
 
-    constructor(props: InternalProps) {
-        super(props);
-        this.list = React.createRef();
-    }
+    const {
+        locale,
+        project,
+        resource,
+        search,
+        status,
+        extra,
+        tag,
+        author,
+        time,
+        entity,
+    } = props.parameters;
 
-    componentDidMount() {
-        document.addEventListener('keydown', this.handleShortcuts);
+    useEffect(() => {
+        document.addEventListener('keydown', handleShortcuts);
 
-        this.selectFirstEntityIfNoneSelected();
-    }
+        return () => {
+            document.removeEventListener('keydown', handleShortcuts);
+        };
+    }, []);
 
-    componentWillUnmount() {
-        document.removeEventListener('keydown', this.handleShortcuts);
-    }
+    useEffect(() => {
+        //a workaround preventing the first rendering
+        if (isFirstRendered.current) {
+            isFirstRendered.current = false;
+            selectFirstEntityIfNoneSelected();
+        } else {
+            props.dispatch(entities.actions.reset());
 
-    componentDidUpdate(prevProps: InternalProps) {
-        this.selectFirstEntityIfNoneSelected();
-
-        // Whenever the route changes, we want to verify that the user didn't
-        // change locale, project, resource... If they did, then we'll have
-        // to reset the current list of entities, in order to start a fresh
-        // list and hide the previous entities.
-        //
-        // Notes:
-        //  * It might seem to be an anti-pattern to change the state after the
-        //    component has rendered, but that's actually the easiest way to
-        //    implement that feature. Note that the first render is not shown
-        //    to the user, so there should be no blinking here.
-        //    Cf. https://reactjs.org/docs/react-component.html#componentdidupdate
-        //  * Other solutions might involve using `connected-react-router`'s
-        //    redux actions (see https://stackoverflow.com/a/37911318/1462501)
-        //    or using `history.listen` to trigger an action on each location
-        //    change.
-        //  * I haven't been able to figure out how to test this feature. It
-        //    is possible that going for another possible solutions will make
-        //    testing easier, which would be very desirable.
-        const previous = prevProps.parameters;
-        const current = this.props.parameters;
-        if (
-            previous.locale !== current.locale ||
-            previous.project !== current.project ||
-            previous.resource !== current.resource ||
-            previous.search !== current.search ||
-            previous.status !== current.status ||
-            previous.extra !== current.extra ||
-            previous.tag !== current.tag ||
-            previous.author !== current.author ||
-            previous.time !== current.time
-        ) {
-            this.props.dispatch(entities.actions.reset());
-        }
-
-        // Scroll to selected entity when entity changes
-        // and when entity list loads for the first time
-        if (
-            previous.entity !== current.entity ||
-            (!prevProps.entities.entities.length &&
-                this.props.entities.entities.length)
-        ) {
-            const list = this.list.current;
-            const element = list.querySelector('li.selected');
-            const mediaQuery = window.matchMedia(
-                '(prefers-reduced-motion: reduce)',
-            );
-            const behavior = mediaQuery.matches ? 'auto' : 'smooth';
-
-            if (element) {
-                element.scrollIntoView({
-                    behavior: behavior,
-                    block: 'nearest',
-                });
+            // Scroll to selected entity when entity changes
+            // and when entity list loads for the first time
+            if (!entitiesPrev.length && props.entities.entities.length) {
+                const element = list.current.querySelector('li.selected');
+                const mediaQuery = window.matchMedia(
+                    '(prefers-reduced-motion: reduce)',
+                );
+                const behavior = mediaQuery.matches ? 'auto' : 'smooth';
+                if (element) {
+                    element.scrollIntoView({
+                        behavior: behavior,
+                        block: 'nearest',
+                    });
+                }
             }
         }
-    }
+    }, [
+        locale,
+        project,
+        resource,
+        search,
+        status,
+        extra,
+        tag,
+        author,
+        time,
+        entity,
+    ]);
 
-    handleShortcuts: (event: KeyboardEvent) => void = (
+    const handleShortcuts: (event: KeyboardEvent) => void = (
         event: KeyboardEvent,
     ) => {
-        const key = event.keyCode;
+        const key = event.code;
 
         // On Ctrl + Shift + A, select all entities for batch editing.
-        if (key === 65 && !event.altKey && event.ctrlKey && event.shiftKey) {
+        if (key === '65' && !event.altKey && event.ctrlKey && event.shiftKey) {
             event.preventDefault();
 
             const {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -91,5 +91,8 @@
     "tabWidth": 4,
     "jsxSingleQuote": true,
     "singleQuote": true
+  },
+  "dependencies": {
+    "react-infinite-scroller": "^1.2.4"
```
mathjazz, line 96: This file controls parts of our admin UI, so it doesn't have an impact on `EntitiesList`. Please revert the change, as well as changes to `yarn.lock`

mathjazz (commented):
Thanks for the patch!

I haven't looked closely yet. Please make sure to verify the tests pass locally by runnning `make test-frontend` as [documented](https://github.com/mozilla/pontoon/tree/master/frontend#testing). 😄

package.json
yaya-usman: Hey @mathjazz , i can't seem to pass through two test cases(as attached in the image below) due to the fact that in the EntitiesList.test.js wrapper.instance().anyfn() returns null for stateless components,so i can't access the function getMoreEntities or toggleForBatchEditing,i have tried using forwardRefs to expose the function(in this case getMoreEntities), it was able to refer to it but, i still can't seem to get it to  access the actions property(giving me TypeError: actions is undefined), maybe you can help me out with your suggestions, i have also attached (getMoreEntities.png) how i modified the test. 

testCases
getMoreEntities



**PS: This is my first time contributing to an open source project, so kindly take this as a way of me learning from you. : )**
.

abowler2:
Hey @yaya-usman ~

I took a quick look at the refactor, and its looking good. I plan to take a closer look at this patch tomorrow and will try to work out what might be happening with the tests as well.

yaya-usman:
> Hey @yaya-usman ~
> 
> I took a quick look at the refactor, and its looking good. I plan to take a closer look at this patch tomorrow and will try to work out what might be happening with the tests as well.

Thanks @abowler2 ...Let me push my latest changes , so you can get the review the current version

mathjazz:
Please note that the `yarn.lock` file still needs to be reverted to its state in `master`.

frontend/src/modules/entitieslist/components/EntitiesList.tsx
```diff
@@ -45,91 +47,79 @@ type InternalProps = Props & {
...
-        ) {
-            this.props.dispatch(entities.actions.reset());
-        }
-
-        // Scroll to selected entity when entity changes
-        // and when entity list loads for the first time
-        if (
-            previous.entity !== current.entity ||
-            (!prevProps.entities.entities.length &&
-                this.props.entities.entities.length)
```
abowler2, line 108: In this case we need to actually compare the value of `entities.entities.length` between its previous value and its current value, so only watching for a change of any kind will not get the results needed.

If you utilize the `usePrevious` hook (see example of implementation [here](https://github.com/mozilla/pontoon/blob/master/frontend/src/modules/machinery/components/Machinery.tsx#L55)) to store the previous value of `props.entities.entities`, check for changes in `props.entities.entities` in the dependency array, and then wrap the below code in an `if` statement that makes the same checks as above the test errors around the mediaQuery are no longer present. 

There are still failing tests but those appear to be related to receiving improper values (which are likely legitimate fails that are occurring due to the changes made)

Note: I tested this off of commit [ed684ff](https://github.com/mozilla/pontoon/pull/2413/commits/ed684ffabe147686250d7bfa482c270db764d966)

yaya-usman: Thanks for the feedback @abowler2 but i have resolved that already, i am only having issues testing the arguments contained within a dispatched action from the method getMoreEntities() and toggleBatchEdititng(). I can't seem to access the actions which is valid because, it's a functional component so i can't get an instance of it, functions within is not exposed to other files. I am basically trying to workaround this and test it in other way if possible. Any suggestions from you will be welcomed.

PS: I am not good with tests, i am currently learning so i could get around this.

yaya-usman:
> Please note that the `yarn.lock` file still needs to be reverted to its state in `master`.

Done that already, check my latest commit sir

frontend/src/modules/entitieslist/components/EntitiesList.tsx
abowler2: My apologies. I do see that you made that change in a later commit (when I first ran the test suite locally I was failing around the media query, so I opted to start at the commit before your test changes. Clearly my build was off and did not pick up all of your changes as running it from the latest commit now is fine)

That said, it looks like you are on the right track in that you seem to have noted that it is the use of Enzyme's `instance()` method that is failing since that [returns null](https://enzymejs.github.io/enzyme/docs/api/ShallowWrapper/instance.html#instance--reactcomponent) in functional components.

Sadly, I don't have enough testing experience either to be able to steer you towards a clear path.  I suspect that we may want to think about approaching the test differently rather than shoehorning the older method with a work around.

Maybe @eemeli or @mathjazz could provide some guidance here?

yaya-usman: @abowler2 Exactly, that was the problem i was facing.

mathjazz:
> > Please note that the `yarn.lock` file still needs to be reverted to its state in `master`.
> 
> Done that already, check my latest commit sir

Actually it's still in, see this huge diff: https://github.com/mozilla/pontoon/pull/2413/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2de.

frontend/src/test/store.jsx
```diff
@@ -4,7 +4,6 @@
  * Notably, this one doesn't have any logging, and supports an initialState.
  */
 
-import React from 'react';
```
mathjazz, line 7: Please revert this change.

yaya-usman: yes, just figured that i mistakenly removed it, the test failed cos of this. I'll do that and revert the yarn.lock as well ASAP

mathjazz: Excellent, thanks for your help!

Pike:
If this is too tricky, it might just be OK not to do it. In the end, `EntitiesList` is already a functional component (`EntitiesListBase` isn't, but that shouldn't matter?)

frontend/src/modules/entitieslist/components/EntitiesList.tsx
eemeli: In order to get access to the "instance" of a functional component, [React.useImperativeHandle](https://reactjs.org/docs/hooks-reference.html#useimperativehandle) is your friend.

The code around this should include a comment noting that this exposure is solely for testing, and really the upstream component `react-infinite-scroller` ought to play more nicely with test environment, so that the test could be made more real.

yaya-usman: @eemeli I did that as well using forwardRefs and imperativehandle if you check the comments above, already stated that... But the issue now lies in getting the actions dispatched within those functions exposed.. When I did that, it's sayings actions not defined... Maybe you can help check it out.. Nonetheless I have skipped this test by wrapping setTimeOut() around it,that is why I could pass all test cases till I figure that out... My take is that, we should find another way of testing it rather than trying to access the arguments dispatched by an action.

abowler2: > really the upstream component react-infinite-scroller ought to play more nicely with test environment, so that the test could be made more real.

@eemeli Given that the underlying reason we are converting this component is to replace this infinite scroll component (which requires a work around to properly load new content and is no longer maintained well) with the [hook version](https://github.com/mozilla/pontoon/pull/2394) that is now used in Concordance search (Machinery.tsx) should we consider just replacing it within this work?  

With Machinery we had a button to load more content so the conversion could be decoupled from the implementation of the infinite scroll hook.

mathjazz: As @Pike [pointed out](https://github.com/mozilla/pontoon/pull/2413#issuecomment-1025546448), this patch is not even required for using `react-infinite-scroll-hook`, because `EntitiesList` actually already is a functional component. Which I overlooked, sorry about that. :/

I'd still proceed (because hooks seem to be better supported by the React ecosystem), unless it will make transition to `react-infinite-scroll-hook` harder. @abowler2 Thoughts?

We should not worry about the skipped `excludes current entities when requesting new entities` test, because it's only required by `InfiniteScroll`, which we'll drop. I'm fine with dropping that test. I'm also fine with doing the switch within the same patch if that makes things easier.

abowler2: > As @Pike [pointed out](https://github.com/mozilla/pontoon/pull/2413#issuecomment-1025546448), this patch is not even required for using `react-infinite-scroll-hook`, because `EntitiesList` actually already is a functional component. Which I overlooked, sorry about that. :/

`EntitiesList` is a functional component; however, the infinite scroll component exists in `EntitiesListBase` which is a class component. Unless I'm missing something, `EntitiesListBase` would need to be functional in order to utilize the hook

> I'd still proceed (because hooks seem to be better supported by the React ecosystem), unless it will make transition to `react-infinite-scroll-hook` harder. @abowler2 Thoughts?

Given that current and future iterations of React are hook based I would agree that the effort to convert this component is worth it.

> We should not worry about the skipped `excludes current entities when requesting new entities` test, because it's only required by `InfiniteScroll`, which we'll drop. I'm fine with dropping that test. I'm also fine with doing the switch within the same patch if that makes things easier.

If that test is only a result of the `InfiniteScroll` component, then we should definitely just make the switch now rather than spend any more time trying to 'fix' a test that will soon be obsolete or require a new approach.

Pike: AFAICT, all of

https://github.com/mozilla/pontoon/blob/f867702b8f917fbb4ceda5fbd16996a2a6783d2e/frontend/src/modules/entitieslist/components/EntitiesList.tsx#L406-L420

is just unhooking hooks, and then passing them as props to the class component. We'd do the same thing for `react-infinite-scroll`

yaya-usman:
> I took the patch for a spin and I only found two issues:
> 
> 1. Whenever I select an entity, the entire entity list and entity details panel refresh.
> 2. When I load the translate URL without specifying an entity (e.g. http://localhost:8000/sl/terminology/common/), the first entity does not get selected (as it does on prod: https://pontoon.mozilla.org/sl/terminology/common/?string=212212).

:(  let me try and set it up on my PC and run it, will update my patch if i figure out the problem.

frontend/src/modules/entitieslist/components/EntitiesList.tsx
mathjazz: Correct. We could put the `react-infinite-scroll`-related logic into `EntitiesList`, including the `getMoreEntities` function, and keep the majority of code in `EntitiesListBase` as a class component. In the position we're in right now with this patch, I'd do that as a last resort, though.

> If that test is only a result of the `InfiniteScroll` component, then we should definitely just make the switch now rather than spend any more time trying to 'fix' a test that will soon be obsolete or require a new approach.

It is. We'll need to change the API call on backend to only return the requested page (instead of exluding already received entities).

yaya-usman:
@mathjazz finally, it's working perfectly, thanks for giving me

> I took the patch for a spin and I only found two issues:
> 
> 1. Whenever I select an entity, the entire entity list and entity details panel refresh.
> 2. When I load the translate URL without specifying an entity (e.g. http://localhost:8000/sl/terminology/common/), the first entity does not get selected (as it does on prod: https://pontoon.mozilla.org/sl/terminology/common/?string=212212).


@mathjazz finally😅, it is fixed , it is working perfectly now, i have tested it on my local environment. Thanks for the tip on what happened when you ran it on your local environment.

mathjazz:
I can confirm this now works as expected!

As agreed on Pontoon call, let's fix #2420 first and then continue with this patch. That will allow us to drop tests that are only needed because of the old library.

yaya-usman:
> 

ok, let me handle #2420  as well

yaya-usman:
The continuation will be from

yaya-usman:
This issue has been extended to issue #2420, which will cover all the changes from there.

PR #2415, opened 2022-01-30
Make Screenshots deal with state directly

Pike:


frontend/src/modules/entitydetails/components/Screenshots.test.js
```diff
@@ -1,23 +1,20 @@
...
-        );
-
+        const wrapper = mountComponentWithStore(Screenshots, store, {
+            locale: 'kg',
+            source,
+        });
         wrapper.find('img').simulate('click');
 
-        expect(openLightboxFake.calledOnce).toEqual(true);
+        const newState = store.getState()[lightbox.NAME];
```
Pike, line 16: I'm not too fond of the test here.

In a perfect world, I'd love the test here to pair with the test in https://github.com/mozilla/pontoon/blob/f000a47d7ff18cdae1c84b306711d34816f1db52/frontend/src/core/lightbox/components/Lightbox.test.js#L34-L45, i.e., catch the call to `dispatch`.

Mocking `store.dispatch` didn't seem to do that, much to my surprise.

Pike: Figured it out, had to mock before mounting.

frontend/src/modules/entitydetails/components/Screenshots.tsx
```diff
@@ -15,37 +16,26 @@ type Props = {
...
+        return null;
+    }
+
+    return (
+        <div className='screenshots'>
+            {images.map((urlWithLocale, i) => (
                 <img
                     src={urlWithLocale}
                     alt=''
                     key={i}
```
eemeli, line 35:
```suggestion
            {images.map((urlWithLocale) => (
                <img
                    src={urlWithLocale}
                    alt=''
                    key={urlWithLocale}
```

Nitpick: Am I right in presuming that `urlWithLocale` will not be repeated within `images`? If so, it's a much better unique identifer than `i`.

eemeli (approved):
The functional code looks _much_ better.

frontend/src/modules/entitydetails/components/Screenshots.tsx
Pike: I agree that the index is probably the worst-practice choice here :-/

I guess it's safe to assume that folks are not going to have a comment with multiple references to the same image in there.

PR #2419, opened 2022-02-02
Add ability to Re-apply existing filter

Mitch4sho:
Selecting the same filter will refresh the list.


mathjazz (approved):
Excellent work!

Looks like some blank lines have been removed from both files. Could you please revert them? Perhaps its a config option in your editor that caused that?

Mitch4sho:
I thought to make prettier would take care of the whitespaces. All those locations are where I had the console.logs at 😄

PR #2422, opened 2022-02-09
Restore CSP and CSRF protection on /graphql

mathjazz:
Content Security Policy should be only exempt if it absolutely cannot be used. It doesn't seem like this is the case for our GraphQL endpoint. The same goes for CSRF.

### DIFF
--- a/pontoon/settings/dev.py
+++ b/pontoon/settings/dev.py
@@ -47,6 +47,7 @@ CSP_SCRIPT_SRC = base.CSP_SCRIPT_SRC + (
     "http://ajax.googleapis.com",
     # Needed for GraphiQL
     "https://cdn.jsdelivr.net",
+    # Needed for GraphiQL (inline script)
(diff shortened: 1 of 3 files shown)


pontoon/settings/dev.py
```diff
@@ -42,7 +42,16 @@
     re.VERBOSE,
 )
 
-CSP_SCRIPT_SRC = base.CSP_SCRIPT_SRC + ("http://ajax.googleapis.com",)
 CSP_IMG_SRC = base.CSP_IMG_SRC + ("data:",)
+CSP_SCRIPT_SRC = base.CSP_SCRIPT_SRC + (
+    "http://ajax.googleapis.com",
+    # Needed for GraphiQL
+    "https://cdn.jsdelivr.net",
+    "'sha256-gp1+DqtmqR6gC56O1TE7F+GuoHAHHbXyN+gaBi8gcjo='",
```
eemeli, line 50: Please add a code comment also explaining this.

eemeli (approved):

PR #2428, opened 2022-02-14
Port frontend build from react-scripts to vite

eemeli:
This is likely to require some testing and development still. It's currently at a stage where it builds, passes tests & can be run embedded in the Django app as before. [Vite](https://vitejs.dev/) mostly uses [Rollup](https://rollupjs.org/) internally for its build, rather than the Webpack used by CRA.

A couple of notable changes, which may need to be expanded upon later:
- The production code is packaged as ES modules.
- `yarn test` doesn't automatically start watching; use `yarn test --watch` for that.
- The frontend `index.html` source file is now in `frontend/` rather than `frontend/public/`, and its `<html>` element now always sets its `lang` attribute -- Vite's HTML parser was not happy with the conditional template markup within the tag.
- The files and directories at the root of `frontend/src/` are no longer available for import via their bare names, but need to use a prefix `~` as the first import path part.
- Explicitly installing Jest also updated it, revealing a few new bugs. These are mostly fixed, but a test in `EntityDetails.test.js` needed to be skipped for now due to it throwing uncaught async errors.
- A few libraries needed to be updated, including React 16.13 -> 16.14. Nothing too drastic, though.


eemeli:
Hit a Vite bug with its handling of `<link rel="stylesheet">` HTML tags pointing to external resources. Filed upstream as vitejs/vite#6931. Not sure yet about any workaround.

eemeli:
Similarly to the change to tag-admin done in #2430, the package management for the frontend is here also updated to use npm rather than yarn, and to be configured as a workspace. This requires npm v7 or later. You will want to wipe out the `/node_modules` and `/frontend/node_modules` directories before running `npm install` at the repo root after this PR.

The `local/pontoon-frontend` Docker image now uses `node:16` as a base and `/app/frontend` rather than `/frontend` for its contents & workdir, in order to accommodate the changed topography.

eemeli:
Rather than using Vite, it looks like we can simplify things a lot by using Rollup directly, and by dropping the separate frontend dev server. Will file a separate PR for those changes.

PR #2429, opened 2022-02-15
Add workflow_dispatch triggers for all actions

eemeli:
Sometimes, it's useful to be able to manually start a CI job, e.g. when working on a feature branch. By adding an empty `workflow_dispatch` value to the GitHub actions' `on` object, this becomes possible: https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -17,6 +17,7 @@ on:
       - docker**
       - pontoon/**
       - .github/workflows/backend.yml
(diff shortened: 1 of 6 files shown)


Pike:
This is fine, but I'm personally also encouraging to just have draft PRs instead of semi-hidden branches and manual triggers.

Pike (approved):

PR #2437, opened 2022-02-21
Refactor the frontend build

eemeli:
This PR supersedes #2428.

This replaces the CRA `react-scripts` with Rollup, and drops yarn completely. Some other significant changes:

- The `frontend` service is dropped from the docker-compose config. Rather than proxying requests to `localhost:3000`, during development django-pipeline takes care of updating the served files when they're built by Rollup.
- The frontend is now a workspace of the root package.json, which contains all dev dependencies for all workspaces.
- The files and directories at the root of `frontend/src/` are no longer available for import via their bare names, but need to use a prefix `~` as the first import path part.
- Explicitly installing Jest also updated it, revealing a few new bugs. These are mostly fixed, but a test in `EntityDetails.test.js` needed to be skipped for now due to it throwing uncaught async errors.
- A few libraries needed to be updated, including React 16.13 -> 16.14. Nothing too drastic, though.

A couple of open questions remain:
- [x] @mathjazz should look at the Heroku config changes to make sure that I've not messed anything up, in particular wrt re-enabling default static file handling
- [x] I'm still working on a good watch system that can handle both Rollup builds transparently. Putting them in docker containers is possible, but at least on my machine its slows them down 5-8x when run directly on the host, which means that the `frontend` build in particular takes about 15s to refresh; that's too long.


app.json
```diff
@@ -63,7 +63,7 @@
     },
     "DISABLE_COLLECTSTATIC": {
       "description": "Disables running ``./manage.py collectstatic`` during the build. Should be set to ``1``.",
-      "value": "1"
+      "value": ""
```
mathjazz, line 66: We also need to add `"required": false,` here.

By default the field will be required and without a default value it's not possible to just "Deploy app" with a click of a button.

eemeli: Added. This part of the build still feels a bit too mysterious to me.

mathjazz:
The Heroku build using the purple "Deploy to Heroku" button on your branch fails:

```
-----> $ python manage.py collectstatic --noinput
         File "/tmp/build_0865ae7b/manage.py", line 16, in <module>
         File "/app/.heroku/python/lib/python3.9/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
           utility.execute()
         File "/app/.heroku/python/lib/python3.9/site-packages/django/core/management/__init__.py", line 413, in execute
         File "/app/.heroku/python/lib/python3.9/site-packages/django/core/management/base.py", line 354, in run_from_argv
         File "/app/.heroku/python/lib/python3.9/site-packages/django/core/management/base.py", line 398, in execute
         File "/app/.heroku/python/lib/python3.9/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 187, in handle
           collected = self.collect()
         File "/app/.heroku/python/lib/python3.9/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 105, in collect
           for path, storage in finder.list(self.ignore_patterns):
         File "/app/.heroku/python/lib/python3.9/site-packages/django/contrib/staticfiles/finders.py", line 130, in list
           for path in utils.get_files(storage, ignore_patterns):
         File "/app/.heroku/python/lib/python3.9/site-packages/django/contrib/staticfiles/utils.py", line 23, in get_files
           directories, files = storage.listdir(location)
         File "/app/.heroku/python/lib/python3.9/site-packages/django/core/files/storage.py", line 323, in listdir
           for entry in os.scandir(path):
       FileNotFoundError: [Errno 2] No such file or directory: '/tmp/build_0865ae7b/frontend/dist'
 !     Error while running '$ python manage.py collectstatic --noinput'.
       See traceback above for details.
       You may need to update application code to resolve this error.
       Or, you can disable collectstatic for this application:
          $ heroku config:set DISABLE_COLLECTSTATIC=1
       https://devcenter.heroku.com/articles/django-assets
 !     Push rejected, failed to compile Python app.
 !     Push failed
```

What's the motivation for the change. See the `DISABLE_COLLECTSTATIC` docs:
https://mozilla-pontoon.readthedocs.io/en/latest/admin/deployment.html#environment-variables

eemeli:
@mathjazz Aah! That clarifies things. I'll revert the `app.json` changes. I thought they were necessary, given the change to how static files are served.

mathjazz:
Deployment (without the last commit, which only changes GH workflow) seems to have worked:
https://pontoon-rollup.herokuapp.com/

docker/k8s-first-steps/server_run.sh
```diff
@@ -22,7 +22,7 @@ echo ">>> Setting up the db for Django" >> /app/server_run.log
 python manage.py migrate >> /app/server_run.log
 
 echo ">>> Starting frontend build process in the background" >> /app/server_run.log
-cd frontend && yarn start &
+cd frontend && npm start &
```
Pike, line 25: This looks surprising. Not in the sense that it's new, but that it's not symmetric to `tag-admin`

frontend/tsconfig.json
```diff
@@ -44,8 +44,10 @@
 
     /* Module Resolution Options */
     "moduleResolution": "node",                     /* Specify module resolution strategy: 'node' (Node.js) or 'classic' (TypeScript pre-1.6). */
-    "baseUrl": "src/",                              /* Base directory to resolve non-absolute module names. */
-    // "paths": {},                                 /* A series of entries which re-map imports to lookup locations relative to the 'baseUrl'. */
+    // "baseUrl": "src/",                           /* Base directory to resolve non-absolute module names. */
+    "paths": {                                      /* A series of entries which re-map imports to lookup locations relative to the 'baseUrl'. */
+      "~/*": ["./src/*"]
+    },
```
Pike, line 50: I can't say that I'm fond of this change, but that's just an opinion.

The whole tilde import stuff .. I just find it confusing.

In another project, there's actual names to this, like, `_core/*` and `_modules/*`, though I'm not sure this is really needed.

Is this just to avoid name clashes between nodejs packages and our source code?

pontoon/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "server",
+  "private": true,
+  "dependencies": {
+    "terser": "^5.10.0",
+    "yuglify": "^2.0.0"
+  }
+}
```
Pike, line 8: I remember something flying by that might be related to this, but is this still needed?

If so, is this documented? I haven't read all the docs, I'm afraid

pontoon/translate/views.py
```diff
@@ -21,70 +16,14 @@
...
-    )
+    return render(request, "frontend.html", context=context, using="jinja2")
 
 
-# For production, we've added the front-end static files to our list of
-# static folders. We can thus simply return a template view of index.html.
 @ensure_csrf_cookie
 def catchall_prod(request, context=None):
-    return render(request, "index.html", context=context, using="jinja2")
+    return render(request, "frontend.html", context=context, using="jinja2")
```
Pike, line 26: Seems we only need one of `catchall_*` now?

Pike (commented):
I was able to run things locally on wsl2, npm tests work on both frontend and tags. I wasn't able to get `--watchAll` to just run the tests for changed files, sadly. Neither for frontend nor for tags.

I also left some fly-by comments. Not really a review by any means, though, either way.

docker/k8s-first-steps/server_run.sh
eemeli: Good point. I'll need to review this script in a bit more detail.

frontend/tsconfig.json
eemeli: Yeah, pretty much. Tbh, the actual fix I'd prefer would be to de-complicate the structure enough to not need anything except for relative paths.

Very much open to discussing alternatives on this, or to reverting this change.

pontoon/package.json
eemeli: These are required for django-pipeline, which needs the executables to be available somewhere.

pontoon/translate/views.py
eemeli: The CSRF decorators are different on `catchall_dev` and `catchall_prod`. As I wasn't completely certain about what they did, I left them here as separate.

pontoon/package.json
mathjazz: Right:
https://github.com/eemeli/pontoon/blob/rollup-all/pontoon/settings/base.py#L533

Axel you might remember either of these:
https://bugzilla.mozilla.org/show_bug.cgi?id=1687371
https://bugzilla.mozilla.org/show_bug.cgi?id=1692423

frontend/tsconfig.json
mathjazz: Not a fan either. If this can be reverted without causing clashes, I'd be in favor of that.

mathjazz:
> I wasn't able to get `--watchAll` to just run the tests for changed files, sadly. Neither for frontend nor for tags.

Does that work currently (i.e. on master)? I never used it since we dropped it from `make test-frontend`.

I must say however that outside docker... test and the app run sooo much faster! \o/

pontoon/package.json
Pike: I'm thrown off because there's no `pontoon/node_modules`. But I also don't understand workspaces.

In any way, though, should these be `devDependencies`?

frontend/tsconfig.json
eemeli: One small DX issue with using undecorated imports for the root-level directories is that automated tooling won't be able to differentiate between them, and so e.g. will think that `core/utils` ought to come before `react`. For example, in VS Code if the imports are in order, then its tooling will also add new imports to appropriate places from autocompletion, and `~` is after the letters in Latin-1 (though `_` appears to be special-cased to work the same way). Otherwise imports are added to the end of the list, and need to be manually moved around.

So while yes, we can keep with the old pattern, it does mean mixing the app's internal and dependency namespaces, and it means that tooling won't work quite as well.

eemeli:
Both `--watch` and `--watchAll` seem to work for me? Running e.g. this from the root of the repo:
```
npm test -w frontend -- --watch
```

or this while in `frontend/`:
```
npm test -- --watchAll
```

As one different to `yarn`, `npm` requires the `--` to separate the options given to a script. One way to avoid that is to call jest more directly, again while in `frontend/`:
```
npx jest --watch
```

Pike:
Tried again with a bit more intent, and `npm test -- --watch` works, too. I tried it with `Comments.tsx`, and that just boils down to test everything under the sun, so that's why I thought it didn't work. Testing with `EntitiesList.tsx` actually yields running just its test.

frontend/tsconfig.json
mathjazz: That's sounds like a reasonable argument to me. Let's go ahead, but please add a comment about the imports. Maybe in the CONTRIBUTING file.

Makefile
```diff
@@ -62,7 +62,7 @@ build-frontend: node_modules
 build-tagadmin: node_modules
 	npm run build -w tag-admin
 
-build-server: server-env
+build-server: server-env frontend/dist tag-admin/dist
```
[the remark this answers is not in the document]
eemeli: Not quite; the `build-server` target depends on `frontend/dist` and `tag-admin/dist`, which are non-phony targets. This means that `build-server` needs for the front-end builds to have been run at some point, but it doesn't need to build them on every call.

frontend/src/modules/entitydetails/components/EntityDetails.test.js
```diff
@@ -194,8 +196,17 @@ describe('<EntityDetailsBase>', () => {
     });
 });
 
-describe('<EntityDetails>', () => {
+/**
+ * This test ends up calling fetch(), and expects it to actually work.
+ * It needs to be skipped for now, because it triggers async errors when
+ * the API calls fail, and the calling code doesn't handle the errors.
```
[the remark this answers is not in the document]
eemeli: At the moment, skipped tests are showing up in the test suite results. I'll file issues for these after this burst of PRs has landed, so that it's easier to point at the skipped test in the code.

Makefile
```diff
@@ -4,19 +4,18 @@ DOCKER := $(shell which docker)
...
 
-.PHONY: build build-frontend build-server server-env setup run clean shell ci test test-frontend test-server jest pytest format lint types eslint prettier check-prettier flake8 pyupgrade check-pyupgrade black check-black dropdb dumpdb loaddb build-tagadmin build-tagadmin-w sync-projects requirements
+.PHONY: build build-frontend build-tagadmin build-server server-env setup run clean shell ci test test-frontend test-tagadmin test-server jest pytest format lint types eslint prettier check-prettier flake8 pyupgrade check-pyupgrade black check-black dropdb dumpdb loaddb sync-projects requirements
 
 help:
 	@echo "Welcome to Pontoon!\n"
 	@echo "The list of commands for local development:\n"
 	@echo "  build            Builds the docker images for the docker-compose setup"
-	@echo "  build-frontend   Builds just the frontend image"
+	@echo "  build-frontend   Builds just the translate frontend component"
```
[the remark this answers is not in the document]
eemeli: Done; `frontend/` renamed as `/translate` everywhere.

docs/dev/setup.rst
```diff
@@ -39,6 +39,13 @@ Quickstart
 
 2. From the root of the repository, run::
 
+     $ npm install
```
mathjazz, line 42: Can we run this as part of `make build`?

Please also remove the warning below about localhost:8000

eemeli: Yes, at the cost of adding a couple of seconds to every `make build` call as `npm` will need to check that everything that should be installed is installed.

Going to presume that you meant the note about `localhost:3000`, which I'll remove.

mathjazz: I think a couple of seconds is fine for a command that's not used very often.

3000, yeah.

eemeli: D'oh, I had already set it up to run as a part of `make build`. So I just removed this useless addition to the docs.

mathjazz (approved):

PR #2438, opened 2022-02-26
Set `tabWidth: 2` in Prettier config

eemeli:
As discussed on Matrix & last week's call.

This PR will need to be followed by a separate change updating the `.git-blame-ignore-revs` file.

If you need to **rebase work past this style change**, do as follows:

1. Consider this to be commit `commitA`, replacing that with its id in the following.
2. To make sure mistakes aren't fatal, assign a second branch to your current work.
3. Rebase your branch on the commit immediately before this one, commitA~
4. Run the following command at the root of the repo:

        git rebase --strategy-option=theirs \
          --exec 'npx prettier --write . && git add -u && git commit --amend --no-edit' \
          commitA

That will take a short while esp. if you have multiple commits, as it runs Prettier on everything for every commit. If you've deleted files, the rebase may drop down to interactive mode and have you `git rm` as appropriate, then `git rebase --continue`.

You should end up with just your changes in your branch, prettily formatted. To validate that, apply the same Prettier config change to your original branch, reformat the files with `npm run prettier`, and then compare the results with the rebased branch.


Makefile
```diff
@@ -122,13 +122,13 @@ pyupgrade:
 	"${DC}" run --rm server pyupgrade --exit-zero-even-if-changed --py38-plus *.py `find pontoon -name \*.py`
 
 check-pyupgrade:
-	"${DC}" run --rm webapp pyupgrade --py38-plus *.py `find pontoon -name \*.py`
+	"${DC}" run --rm server pyupgrade --py38-plus *.py `find pontoon -name \*.py`
```
mathjazz, line 125: 😅

mathjazz (approved):

Issue #2300, opened 2021-06-03
Should not depend on React internals when targeting comment tab


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1714370](https://bugzilla.mozilla.org/show_bug.cgi?id=1714370)
Bug Reporter: @mathjazz
Depends on:   [Bug 1685565](https://bugzilla.mozilla.org/show_bug.cgi?id=1685565)
See also:     https://bugzilla.mozilla.org/show_bug.cgi?id=1712446

According to bug 1712446, understanding of `commentTabRef` would be helpful to fix this bug.

eemeli:
The current use of `commentTabRef.current._reactInternalFiber.index` was introduced in #1676, and should be replaced with an approach that uses public interfaces only.

Issue #2295, opened 2021-05-21
Tighten eslint and tsc configs to be strict


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1712449](https://bugzilla.mozilla.org/show_bug.cgi?id=1712449)
Bug Reporter: @Pike
Depends on:   [Bug 1685565](https://bugzilla.mozilla.org/show_bug.cgi?id=1685565)

The PR in bug 1685565 has pretty lax configs for tsc and eslint in the frontend config files.

We should tighten those towards strict, filing this bug to track that.

eemeli:
TS config is tightened in #2439.

PR #2439, opened 2022-02-26
Enable `strict: true` in TypeScript config, refactor as necessary

eemeli:
~This builds on top of #2437 and #2438, so marking as draft & rebasing once they're merged.~ Edit: Done; rebased.

This is a set/sequence of changes that fixes/improves the TS types sufficiently to enable `strict: true`. Mostly the changes are really minor, and should not be observable in the compiled JS code. Some refactoring was needed to avoid patterns that aren't TypeScript-y enough to pass muster.

It may be easiest to review this commit-by-commit (starting from "chore: Specify TS typings better (no JS changes)"), as I've bunched up similar sets of changes. In a couple of cases, fixes were applied to code that may be externally visible during some error conditions, but the effect should be that potential unhandled errors are avoided.


translate/src/modules/machinery/components/ConcordanceSearch.tsx
```diff
@@ -11,8 +11,8 @@ type Props = {
   translation: MachineryTranslation;
 };
 
-function ProjectList({ projects }: { projects: string[] }) {
-  const notEmpty = projects.filter(Boolean);
+function ProjectList({ projects }: { projects: (string | null)[] }) {
+  const notEmpty = projects.filter(Boolean) as string[];
```
Pike, line 15: We should filter on something better here than asserting type, I think.

translate/src/modules/resourceprogress/components/ProgressChart.tsx
```diff
@@ -49,6 +49,10 @@ export default class ProgressChart extends React.Component<Props> {
       this.props.stats;
     const dpr = window.devicePixelRatio || 1;
     const canvas = this.canvas.current;
+    if (!canvas) return;
+
+    const context = canvas.getContext('2d');
+    if (!context) return;
```
Pike, line 55: Just a spoon full of sugar...

```suggestion
    const context = canvas?.getContext('2d');
    if (!context) return;
```

translate/src/modules/search/components/SearchBox.test.js
```diff
@@ -91,47 +91,6 @@ describe('<SearchBoxBase>', () => {
...
-      extras: {
-        unchanged: true,
-        rejected: false,
-      },
-    });
-
-    const selected = wrapper.instance().getSelectedExtras();
-    expect(selected).toEqual(['unchanged']);
-  });
-
```
Pike, line 134: I haven't gotten to review this, but if it's easy for you to disclose, why is it good to remove these tests?

translate/src/store.ts
```diff
@@ -6,15 +6,18 @@ import history from './historyInstance';
 import createRootReducer from './rootReducer';
 
 const store = configureStore({
+  // @ts-ignore Here be dragons
```
Pike, line 9: This smells like something we should have a follow-up for?

My gut says we have a typing bug here. In the sense of 1+.

Pike (commented):
Not yet at a full review, but figured I should share some notes.

I'm torn on the single-line-if-return. I'm torn for my personal projects. I'm often tempted doing it and then get into an argument between my inner me and my inner stas.

translate/src/core/editor/hooks/useUpdateTranslationStatus.ts
```diff
@@ -43,7 +43,9 @@ export default function useUpdateTranslationStatus(): (
       await dispatch(
         history.actions.updateStatus(
           change,
-          entity,
+          // The EditorMainAction tests fail if this dispatch() is skipped on an empty entity
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          entity!,
```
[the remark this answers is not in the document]
eemeli: I was kind of hoping to be able to avoid the problem later on by getting rid of all redux.

translate/src/modules/machinery/components/ConcordanceSearch.tsx
eemeli: To be fair, this is a bit on TS for not capturing the result of the filtering always being a `string[]`. As in, it doesn't matter what we `filter()` on, the typing result will still presume it could be null.

translate/src/modules/search/components/SearchBox.test.js
eemeli: Because they're testing an internal function that shouldn't be public in the first place, and is inlined in the change. These are the lines of code this is testing, in their entirety: https://github.com/mozilla/pontoon/blob/9faea69c202376f3f3515ef665be04a12b61a130/translate/src/modules/search/components/SearchBox.tsx#L189-L193

translate/src/store.ts
eemeli: How about if I file a catch-all issue after this lands for tracking all remaining tsc & eslint error workarounds? Would be easier afterwards, as then could point at specific lines of code on `master`.

Pike: So, I updated VS Code. Which ships TS 4.6, and this error suddenly is pretty omnipresent. That makes it kinda awkward. Only two errors on master (but still two), one is shared, the other is commentTabRef which I think you're fixing on this branch.

I think the problem is that we don't have an app action type. The reducers just take the local action type, and that doesn't align with the app action type.

Found 46 errors in 15 files.

Errors  Files
     3  src/App.tsx:57
     2  src/core/comments/actions.ts:24
     1  src/core/editor/actions.ts:312
     1  src/core/editor/hooks/useUpdateTranslationStatus.ts:44
     1  src/core/l10n/components/AppLocalizationProvider.tsx:39
     1  src/core/plural/actions.ts:23
     1  src/core/plural/components/PluralSelector.tsx:48
     2  src/modules/batchactions/actions.ts:88
     4  src/modules/batchactions/components/BatchActions.tsx:76
     6  src/modules/entitieslist/components/EntitiesList.tsx:144
    14  src/modules/entitydetails/components/EntityDetails.tsx:135
     2  src/modules/history/actions.ts:185
     3  src/modules/navbar/components/Navigation.tsx:46
     3  src/modules/search/components/SearchBox.tsx:184
     2  src/store.ts:9

eemeli: Ugh. Yeah, I see that too. On top of #2444 these are down to 37 errors in 12 files with TS 4.6.

As a workaround, in VS Code you can set it to use the local TS version: https://code.visualstudio.com/docs/typescript/typescript-compiling#_using-the-workspace-version-of-typescript

Could we punt all of these to be a separate issue?

Pike (commented):
Overall, this looks good.

I've got some nits inline.

I've also struggled to find a pattern in imports and curly braces. @mathjazz , does that matter to you?

And last, I still think we should specify a version of typescript in `package.json` that works. The current version range is covering versions that we know that work and versions that don't.

translate/src/modules/search/components/FiltersPanel.tsx
```diff
@@ -2,17 +2,17 @@ import * as React from 'react';
...
 import TimeRangeFilter from './TimeRangeFilter';
+import type { FilterType } from './SearchBox';
 
-import { asLocaleString, useOnDiscard } from '~/core/utils';
-
-import type { Author, TimeRangeType } from '..';
-import type { NavigationParams } from '~/core/navigation';
-import type { Tag } from '~/core/project';
-import type { Stats } from '~/core/stats';
+import './FiltersPanel.css';
```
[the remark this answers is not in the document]
eemeli: Ah, yeah, this should be just after the `../constants` import.

translate/src/core/editor/hooks/useUpdateTranslationStatus.ts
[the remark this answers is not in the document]
eemeli: Issue filed: #2454.

translate/src/core/resource/actions.ts
```diff
@@ -62,7 +62,8 @@ export function get(locale: string, project: string) {
 
     const allResources = resources.pop();
 
-    dispatch(receive(resources, allResources));
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    dispatch(receive(resources, allResources!));
```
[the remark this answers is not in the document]
eemeli: Issue filed: #2454.

translate/src/store.ts
eemeli: Issue filed: #2453.

translate/src/modules/search/components/FiltersPanel.tsx
```diff
@@ -98,7 +98,10 @@ export function FiltersPanel({
         </Localized>
 
         {FILTERS_STATUS.map((status, i) => {
-          const count = status.stat ? stats[status.stat] : stats[status.slug];
+          const count =
+            'stat' in status && status.stat
```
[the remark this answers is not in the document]
eemeli: This exact change actually gets applied in 46e28fa4, as a part of #2444. I'd prefer not to fix it here, as it'd require a manual rebase for the later change.

translate/src/modules/search/components/SearchBox.tsx
```diff
@@ -399,50 +263,22 @@ export class SearchBoxBase extends React.Component<InternalProps, State> {
...
+    const selected: string[] = [];
+    for (const { name, slug } of FILTERS_STATUS)
+      if (statuses[slug]) selected.push(name);
+    for (const { name, slug } of FILTERS_EXTRA)
+      if (extras[slug]) selected.push(name);
+    for (const { name, slug } of project.tags)
+      if (tags[slug]) selected.push(name);
+    if (timeRange) selected.push('Time Range');
+    for (const { display_name, email } of searchAndFilters.authors)
+      if (authors[email]) selected.push(`${display_name}'s translations`);
```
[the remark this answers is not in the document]
eemeli: This ends up improving quite a bit with the ESLint fixes in #2452, here: https://github.com/mozilla/pontoon/pull/2452/commits/f0b12cd717508a91d7ce3647a366032883892f31#diff-c7a8290c5d0a33d346b9b36743b941153d29629fe24fbf702b729fcf3cd0c56aR226-R248

Is that good enough?

eemeli:
> I've got some nits inline.

These should now all be addressed.

> I've also struggled to find a pattern in imports and curly braces. @mathjazz , does that matter to you?

Added #2452 for this.

> And last, I still think we should specify a version of typescript in `package.json` that works. The current version range is covering versions that we know that work and versions that don't.

While the `package.json` range is rather liberal, `typescript` is a dev dependency and as such we ought to be able to rely on the exact specification in `package-lock.json`.

translate/src/modules/search/components/SearchBox.tsx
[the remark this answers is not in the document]
Pike: Yes, thanks.

translate/src/modules/search/components/FiltersPanel.tsx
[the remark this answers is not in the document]
Pike: That's cool, I had a similar idea of potentially inlining that into the lookup.

Pike (approved):
Sweet, I think we're good with the follow-ups coming in follow-up PRs.

mathjazz:
Great work! I'll let your merge this, Eemeli.

I've added #2295 to the list of linked issues. Maybe there are other [typescript](https://github.com/mozilla/pontoon/labels/typescript) issues we can close? #2300?

eemeli:
#2300 is not solved here, though it is made more obvious. I rephrased that issue to make it clearer.

Issue #2440, opened 2022-02-26
In practice, `withActionsDisabled()` does not disable actions


eemeli:
While refactoring code, I started investigating what `withActionsDisabled()` claims to do, and what it actually does. And came to the conclusion that at least at the moment, it doesn't.

The core issue is that its operation is driven by this: https://github.com/mozilla/pontoon/blob/8d2714d5dbfd69e3188e620d02d28adcc06483a6/frontend/src/core/utils/components/withActionsDisabled.tsx#L30-L38

So when `disableAction()` is called:
1. We get one render of the wrapped component with `isActionDisabled: true`
2. `componentDidUpdate()` is called due to the state change (also props change, via the redux store)
3. `isActionDisabled` is set `false`
4. The wrapped component is rendered again, with `isActionDisabled: false`

All of that happens too fast to affect human actions, which means that e.g. double clicking a status-change button sends the action twice. To show that to myself, I put together a [proof of concept](https://github.com/eemeli/pontoon/commit/47eef08840417008b03061e63291356f429bf6cd). Running that, and double clicking a reject button, I see this in the console:

```
wAD enable
{ action: "reject" }
wAD disable { sameProps: false, sameState: true }
wAD disable { sameProps: true, sameState: false }
wAD enable
{ action: "reject" }
wAD disable { sameProps: false, sameState: true }
wAD disable { sameProps: true, sameState: false }
```

I was also able to duplicate the same issue on the staging server by observing multiple REST calls being sent from a double-click.

We should do something about this, but I'm not really sure what?

1. If this is working as intended, and it's disabling some other category of error cases, this should be clarified.
2. If this isn't needed any more, it should be dropped.
3. If double-clicking should only result in one action, and this should prevent that, it should be refactored/fixed to enable that.

eemeli:
Related: #2292

PR #2444, opened 2022-03-04
 Move location & plural forms from Redux to React Context

eemeli:
~This builds on top of PRs up to #2443, so marking as draft & rebasing once they're merged.~ Edit: Done.

This first refactors a whole bunch of components to be functional and turns all selectors into hooks, so that they may more easily depend on non-redux state. Most custom HOCs are also gotten rid of, with an exception noted in #2440.

With that done, the location state is moved from Redux to its own React Context, providing a custom `Link` component (currently used only in `ResourceProgress`) and concentrating there most of the URL parsing/writing.

As a related next step, a context is also added for pluralForm data, in order to replace its connected-react-router `LOCATION_CHANGE` reducer with a useEffect hook depending on the Location. This allows for dropping all of `translate`'s react-router dependencies.

A minor fix is included project tags, which were being sorted during render, modifying a prop value.


mathjazz (commented):
I'm still looking at some bits of the changeset. So far and on a general note it looks good!

Since it touches pretty much every bit of the translate view, we should test it extensively before shipping. For that purpose, I've deployed it to stage.

Seems like it's broken for non-authenticated users, e.g.:
https://mozilla-pontoon-staging.herokuapp.com/sl/firefox/browser/branding/official/brand.dtd/

I'm also unable to submit translations using the Enter keyboard shortcut.

mathjazz:
Two more things I noticed:
* Keyboard shortcuts for copying from helpers or source string, as well as clearing the editor also don't work.
* Clicking on the notification icon (the bell) breaks the app.

eemeli:
Keyboard shortcuts were broken due to relying on `ev.code`, which is only available starting from React v17. I've now downgraded the TS types to match our actual React v16.

Notification menu was broken due to notifications being typed as `Record<string, any>`, which meant that the in-code presumption of `description.content` always being a string wasn't caught; apparently it's `null` for due dates. Added a test suite for the UserNotification component, but it'd be nice if that typing could be made much stricter, esp. as some fields are currently expected by the code to hold specific shapes. That might need some back-end fixes too, though.

mathjazz (approved):
We haven't discovered new issues on stage so far. I'll approve and merge to unblock other PRs, but hold off pushing to prod until tomorrow (when the community testing time is due).

PR #2446, opened 2022-03-08
Validate Repository URLs

mathjazz:
We should validate Repository URLs before they are stored into the DB.

The `repository_url_validator()` has been tested with all Repository URLs on pontoon.mozilla.org. It supports all HTTP(S), FTP(S), SSH and SVN+SSH URLs, as well as Git SCP-like SSH URL.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -20,7 +20,7 @@ from django.contrib.auth.models import User, Group
 from django.contrib.postgres.fields import ArrayField
 
 from django.core.exceptions import ValidationError
-from django.core.validators import validate_email
+from django.core.validators import URLValidator, validate_email
 from django.db import models
 from django.db.models import (
     Count,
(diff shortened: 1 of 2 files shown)


pontoon/base/models.py
```diff
@@ -1796,6 +1796,20 @@ def aggregate_stats(self):
         ).distinct().aggregate_stats(self)
 
 
+def repository_url_validator(url):
+    # Regular URLs
+    validator = URLValidator(["http", "https", "ftp", "ftps", "ssh", "svn+ssh"])
+
+    # Git SCP-like URL
+    pattern = r"(?P<user>git@([\w\.@]+)(/|:))(?P<server>[\w,\-,\_]+)/(?P<project>[\w,\-,\_]+)(.git){0,1}((/){0,1})"
```
eemeli, line 1804:
```suggestion
    pattern = r"git@[\w\.@]+[/:][\w-]+/[\w-]+(.git)?/?"
```

Simpler this way, and the previously named "server" and "project" groups shouldn't include `,`.

eemeli (approved):
The SCP-like pattern is a bit too complex, otherwise looks fine.

pontoon/base/models.py
mathjazz: Good catch!

eudaldv:
The pattern doesn't take into account dashes on the URL format.
Suggestion:
`git@(((?!\-))(xn\-\-)?[a-z0-9\-_]{0,61}[a-z0-9]{1,1}\.)*(xn\-\-)?([a-z0-9\-]{1,61}|[a-z0-9\-]{1,30})\.[a-z]{2,}[/:][\w-]+/[\w-]+(.git)?/?`

Best regards,
Eudald

PR #2452, opened 2022-03-10
Add `curly: error` to ESLint configs

eemeli:
This builds on top of PRs up to #2450, so marking as draft & rebasing once they're merged.

As discussed on today's Pontoon call, let's enforce curly brace style via ESLint.

Issue #2049, opened 2016-12-20
Add an option to link to multiple strings


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1324933](https://bugzilla.mozilla.org/show_bug.cgi?id=1324933)
Bug Reporter: @tomer
CC:   @gaby2300, @ItielMaN, @mathjazz
Blocker for:  [Bug 1390931](https://bugzilla.mozilla.org/show_bug.cgi?id=1390931)

In dxr.mozilla.org I can select multiple lines in a file and than have a link that will show these lines highlighted. I'd suggest having a similar feature in pontoon, so we could send a team member a link that contains multiple strings, instead of one of the following workarounds:

* Sending one string per link
* Sending link to a search results or a resource, and asking them to look on the second string, the third from the bottom, etc. 

I'd like to have a link similar to this one: 
https://pontoon.mozilla.org/he/firefox-aurora/all-resources/?string=75295,75296 
… Which will show me both strings in the list.

As for the UI for creating such lists, I guess having control-click to toggle multiple selections will be enough; this is an advanced feature after all. ☺

eemeli:
This should make it easier to resolve #2276 as well.

PR #2456, opened 2022-03-15
Add notifications for approved & rejected suggestions

### LINKED ISSUE #2094: Notify suggestion authors when their suggestions get accepted/rejected
## [Bug 1398862](https://bugzilla.mozilla.org/show_bug.cgi?id=1398862)
Bug Reporter: @mathjazz
CC:   @flodolo

We should probably do that periodically and aggregate notifications.

eemeli:
Closes #2094 

Adds a script for sending notifications on approved & rejected suggestions to their authors; this should be run every 24 hours by a cron job:

```sh
./manage.py send_review_notifications
```

This follows the same structure and pattern established for suggestion notifications as used in #2329 and #2330.

The message formatting is done all in Python, as it effectively uses a two-variable top level case selector that isn't really supported by current languages; MF2 will deal with it fine.


pontoon/projects/management/commands/send_approved_notifications.py
```diff
@@ -0,0 +1,88 @@
...
+                msg = f"{len(rejected)} of your suggestions were rejected"
+        else:
+            if len(approved) == 1:
+                msg = "one of your suggestions was approved"
+            else:
+                msg = f"{len(approved)} of your suggestions were approved"
+            if len(rejected) == 1:
+                msg += ", and one was rejected"
+            elif len(rejected) > 1:
+                msg += f", and {len(rejected)} were rejected"
```
mathjazz, line 45: I would simplify the logic and only list project-locales with reviewed strings, same as we do in the unreviewed suggestions notifications.

Count is useful, I agree, but I'd rather not mess with the sentences, and instead use a graphic, e.g. Screenshot 2022-03-15 at 19 49 32 

Maybe do that in a follow-up, also for unreviewed suggestion notifications?

eemeli:
Renamed as "review notifications" and applied other suggestions from @mathjazz.

pontoon/projects/management/commands/send_approved_notifications.py
eemeli: The sentence construction isn't really complicated, it's just complex.

I think this and the discussion on due-date notifications is sounding like an impending review of all notifications.

pontoon/projects/management/commands/send_review_notifications.py
```diff
@@ -0,0 +1,91 @@
...
+            & Q(user__profile__review_notifications=True)
+        ):
+            author = suggestion.user
+            locale = suggestion.locale
+            project = suggestion.entity.resource.project
+            key = (author, locale, project)
+
+            if (
+                suggestion.approved
+                and suggestion.active
```
mathjazz, line 72: All approved translations must be active, so we can remove this condition.

pontoon/projects/management/commands/send_review_notifications.py
```diff
@@ -0,0 +1,91 @@
...
+                and suggestion.approved_user != author
+            ):
+                data[key][0].append(suggestion.entity.pk)
+            elif suggestion.rejected and suggestion.rejected_user != author:
+                data[key][1].append(suggestion.entity.pk)
+
+        for ((author, locale, project), (approved, rejected)) in data.items():
+            # Filter out rejections where the author's own suggestion replaced the previous
+            rejected = [x for x in rejected if x not in approved]
```
mathjazz, line 81: I'm not sure about this one. I see why it could make sense when you add a suggestion to an existing approved translation and that suggestion gets accepted. But if I make two suggestions, and one gets approved and the other one rejected, that should count as 1 rejected and 1 approved.

pontoon/projects/management/commands/send_approved_notifications.py
mathjazz: For unreviewed suggestions notifications we also send just one notification per user, here we send one per project-locale combination. (And add counts.)

I'm also curious what @flodolo thinks about the message structure.

flodolo: > I'm also curious what @flodolo thinks about the message structure.

It feels very complex to build. As the recipient of such message, it also forces my brain to do some extra work, because the structure changes depending on the data, and it's not as immediate to parse.

I'd be happy with something minimal (and less elegant/natural) like.

```
Some of your suggestions were reviewed:
* Approved: XX strings
* Rejected: XX strings
```

And drop approved/rejected if the number is 0.

pontoon/projects/management/commands/send_review_notifications.py
eemeli: If you make two suggestions for the same entity and one of them is approved, I would think that the context makes it kinda obvious what's happened?

One alternative that I considered was counting only rejections of suggestions that were never approved, but there doesn't appear to be a way of filtering for only those.

pontoon/projects/management/commands/send_approved_notifications.py
mathjazz: Thanks, Flod. What is your take on 1 notification per user vs one pre each project-locale-user combination?

flodolo: I think 1 notification would be ideal, as long as we can link to specific strings for each project in the message.

I also don't expect notifications to include a lot of projects, so maybe the end result is the same in most cases (for most users, it will be about 1 project).

pontoon/projects/management/commands/send_review_notifications.py
mathjazz: Agreed with both points.

pontoon/projects/management/commands/send_approved_notifications.py
eemeli: As it would appear from this discussion that the spec for the notification contents that I proposed in https://github.com/mozilla/pontoon/issues/2094#issuecomment-1066754638 isn't good enough after all, how should we proceed with re-specifying it?

mathjazz: I suggest we follow the lead of Unreviewed suggestions notifications and send only 1 notification per user with the following content:

```
Your suggestions in the following projects have been reviewed:
* Firefox (X Approved, Y Rejected)
* Common Voice (Z Approved)
```

I would also implement a new `review_time` filter, similar to the `time` filter, and use it in the link.

In a followup, we could also add count to Unreviewed suggestions notifications.

eemeli: Okay; is this how the notification would look like for a single approved string in one project?

```
Your suggestion in the following project has been reviewed:
* Firefox (1 Approved)
```

That seems a bit clumsy. Would it make sense to use a different format if all the strings are in one project, like this?

```
Your suggestions in Firefox have been reviewed (2 Approved, 1 Rejected)
```

And with a special case for a single string:

```
Your suggestion in Firefox has been Approved
```

Also, if we're listing projects and not project+locale combos, how should we create the links if a user had strings reviewed for one project in more than one language?

flodolo: Personally, I prefer consistent message structure for readability (I can quickly scan the relevant info, which is project and number of strings review).

Review for multiple locales+projects within two cron job runs is an edge case, but we clearly need to deal with it somehow. Would it be easier to send separate notifications for each locale?

mathjazz: Agreed with Flod about the message structure.

For Unreviewed, if more than one locales are affected, we list project for each locale and set the URL accordingly. We should show the locale identifier, as well. For example:

```
Your suggestions in the following projects have been reviewed:
* Firefox (Slovenian · sl): X Approved, Y Rejected
* Common Voice (french · fr): Z Approved
```

eemeli: Notification contents refactored as per above. Left out "in the following projects" as redundant, and leading to complexity in needing to match the plural case to the actual count of projects.

pontoon/projects/management/commands/send_review_notifications.py
```diff
@@ -0,0 +1,83 @@
...
+
+            if len(approved) == 0:
+                msg = f"{len(rejected)} Rejected"
+            else:
+                msg = f"{len(approved)} Approved"
+                if len(rejected) > 0:
+                    msg += f", {len(rejected)} Rejected"
+
+            desc += f'\n* {project.name} ({locale.name} · {locale.code}): <a href="{url}">{msg}</a>'
```
mathjazz, line 45: Let's use HTML lists here, which gives as more styling powers and structure. It might be easier to do with a template, see `suggestion_notification.jinja`.

pontoon/projects/management/commands/send_review_notifications.py
```diff
@@ -0,0 +1,83 @@
...
+                    "project": project.slug,
+                    "resource": "all-resources",
+                },
+            )
+            if len(approved) == 1 and len(rejected) == 0:
+                url += "?" + urlencode({"string": approved[0]})
+            elif len(approved) == 0 and len(rejected) == 1:
+                url += "?" + urlencode({"string": rejected[0]})
+            else:
+                url += "?" + urlencode({"author": author.email})
```
mathjazz, line 33: I would suggest we filter by author in all cases. Or block this whole patch on #2049?

eemeli: Added author filter; that does make sense always. Still including the `&string=` when there's only one.

Recommend against blocking on #2049, because the solution to that is likely to be built on top of #2452, and that would require that whole stack of things to resolve before we get to merge this.

pontoon/projects/management/commands/send_review_notifications.py
```diff
@@ -42,9 +42,9 @@ def get_description(self, author, notifyData):
                 if len(rejected) > 0:
                     msg += f", {len(rejected)} Rejected"
 
-            desc += f'\n* {project.name} ({locale.name} · {locale.code}): <a href="{url}">{msg}</a>'
+            desc += f'\n<li>{project.name} ({locale.name} · {locale.code}): <a href="{url}">{msg}</a></li>'
```
mathjazz, line 44: After testing this layout a bit, I propose we make two changes:
1. Drop Locale name and only leave locale code in the brackets. That should make notification data easier to digest and more condensed. Also, some locale names are long. As in 4 words long.
2. Use `PROJECT (LOCALE)` piece as anchor instead of the `msg` piece.

mathjazz:
Other than the proposal in the comment above, I think this patch is good to go!

pontoon/projects/management/commands/send_review_notifications.py
eemeli:
```suggestion
            desc += f'\n<li><a href="{url}">{project.name} ({locale.code})</a>: {msg}</li>'
```

mathjazz:
Fix black and we're good!

mathjazz (approved):
Excellent work!

Issue #2017, opened 2015-07-02
Add translation status field to replace approved & fuzzy


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1179954](https://bugzilla.mozilla.org/show_bug.cgi?id=1179954)
Bug Reporter: @Osmose
CC:   kosiacka@uni.coventry.ac.uk, @mathjazz, @Osmose, @yfdyh000
See also:     https://bugzilla.mozilla.org/show_bug.cgi?id=1409724

The routine that updates the translated_count attribute on Stats instances is wrong; it sometimes comes up with negative counts, which is why we currently have a workaround in place to handle negative counts (https://github.com/mozilla/pontoon/blob/cf6b55b375c9a696325103ede7f28b21f0f1b2ff/pontoon/base/models.py#L550).

We should probably investigate this to see if there's a more correct way of calculating it.

PR #2463, opened 2022-03-25
Add Pretranslated status and drop Fuzzy as status

### LINKED ISSUE #2237: Make distinction between Pretranslated and Fuzzy strings
## [Bug 1666320](https://bugzilla.mozilla.org/show_bug.cgi?id=1666320)
Bug Reporter: @mathjazz

Pretranslations are assigned translation status Fuzzy, which allows us to:
1. Immediately send them to version control system.
2. Make them available in the product.
3. Inform localizers that they should be given another look.
4. Run quality checks on them to identify any errors or warnings.

The caveat however is that translation status Fuzzy is also used for [fuzzy](https://www.gnu.org/software/gettext/manual/html_node/Fuzzy-Entries.html) translations of the Gettext system (filenames ending in .po), which are **not** used in the product. That means bullet point #2 isn't valid for projects using the Gettext system - pretranslated strings in .po files will not make it to the product unless human translator approves them.

We should mitigate that.

mathjazz:
Fixes #2237 by implementing [feature spec 0107](https://github.com/mozilla/pontoon/blob/master/specs/0107-pretranslated-status.md). The patch is ready for review. In the meantime I need to figure out why `pytest` is failing in automation (but not locally).

Summary of changes related to Fuzzy strings:
- Fuzzy status is gone, which means it no longer appears on dashboards, progress charts and string list.
- Fuzzy filter has been moved from Status filters to Extra filters.
- Strings with status Fuzzy are from now on assigned status Missing.
- Fuzzy stats have been removed from Contributors dashboards and Profile pages.
- Fuzzy flag is still supported, and visual representation of Fuzzy strings in the History panel remains unchanged.

Summary of changes related to Pretranslation:
- Pretranslated strings will now use the newly introduced status Pretranslated (rather than Fuzzy).
- Pretranslated status will appear on dashboards, progress charts and string list alongside other stati.
- A new Pretranslated status filter has been added.
- Pretranslated status will be hidden from dashboards, filters and progress charts until we start using pretranslaion.


eemeli (commented):
Some inline comments, at least to start with. I need to consider the full implementation a bit more; not sure how accommodating the changes made here are to making the automatic acceptance of pretranslated strings configurable.

pontoon/base/models.py
```diff
@@ -2405,8 +2415,28 @@ def missing_without_unreviewed(self, locale, project=None):
...
-                Q(approved=True) | Q(fuzzy=True) | Q(rejected=False),
-                lambda x: x.approved or x.fuzzy or not x.rejected,
+                Q(approved=True) | Q(pretranslated=True) | Q(rejected=False),
+                lambda x: x.approved or x.pretranslated or not x.rejected,
+                project=project,
+            )
+        )
+
+    def fuzzy(self, locale, project=None):
+        """Return a filter to be used to select entities marked as "fuzzy".
```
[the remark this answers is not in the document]
mathjazz: We're keeping the Fuzzy filter around and moving it from Status to Extra filters in case users want to list them separately from other Missing strings.

As for the Missing filter, we're applying the change actually - it catches Fuzzy strings now.

mathjazz:
> Some inline comments, at least to start with. 

🥳 

> not sure how accommodating the changes made here are to making the automatic acceptance of pretranslated strings configurable.

That will be a separate patch, but the changes are are not in conflict with it.

pontoon/base/models.py
```diff
@@ -2261,17 +2263,17 @@ def missing(self, locale, project=None):
...
 
-    def fuzzy(self, locale, project=None):
-        """Return a filter to be used to select entities marked as "fuzzy".
+    def warnings(self, locale, project=None):
+        """Return a filter to be used to select entities with translations with warnings.
 
-        An entity is marked as "fuzzy" if all of its plural forms have a fuzzy
-        translation.
+        This filter will return an entity if at least one of its plural forms
+        has an approved, pretranslated or fuzzy translation with a warning.
```
eemeli, line 2276: Why do we still care about fuzzy strings for e.g. this filter? As in, I'm struggling a bit to figure out how to conceptually fit in "fuzzy" together with "pretranslated" and "missing".

Fuzzy suggestions come from gettext's translation memory, so to me there seems to be two different possibilities for how we ought to treat them:
1. We ignore them, because Pontoon is also providing TM results, and we think they're better.
2. We take them as a sort of pretranslation.

Could we do one of the above, rather than this weird special thing? To put that another way, I think I'm arguing that we should drop this part from the [proposal](https://github.com/mozilla/pontoon/blob/master/specs/0107-pretranslated-status.md#treat-gettext-fuzzy-strings-as-missing):
> Internally, we keep using the fuzzy=True flag for Fuzzy strings, which allows us to [...]
> 2. Run quality checks on them.

mathjazz: > Fuzzy suggestions come from gettext's translation memory, so to me there seems to be two different possibilities for how we ought to treat them:
> 
>     1. We ignore them, because Pontoon is also providing TM results, and we think they're better.

I'm not sure we're sure fuzzy strings are completely useless for Mozilla and other users of Pontoon. Fuzzy is a ~bug~feature of Gettext, so it should be supported.

>     2. We take them as a sort of pretranslation.

We need to differentiate between fuzzy and pretranslated in order to serialize them differently in PO files (with or without the `fuzzy` flag).

> Could we do one of the above, rather than this weird special thing?

Could you clarify more precisely which part do you find weird?

eemeli: The way I understand the term, "pretranslation" provides automated suggestions for translations from TM & MT. In some situations, we may feel that these suggestions are likely to be good enough that we could use them in production even without explicit human approval for each one.

Separately from that, "fuzzy" is how gettext encodes and communicates TM suggestions to a translator. It's effectively a subset/implementation of "pretranslated" suggestions, which allows for their representation in the `.po` file format even when unreviewed.

Therefore, if we make the automatic use-in-production of pretranslated suggestions configurable on a project+locale level, we can consider fuzzy strings as pretranslation suggestions on the input, and depending on the config, either include or not the "fuzzy" property when writing pretranslated suggestions into `.po` files.

With this approach, the whole concept of "fuzzy" goes away internally, subsumed within "pretranslated". On the other hand, if "fuzzy" remains at some level internally, that seems weird and to me communicates that we have different understandings of these terms.

mathjazz: > The way I understand the term, "pretranslation" provides automated suggestions for translations from TM & MT. In some situations, we may feel that these suggestions are likely to be good enough that we could use them in production even without explicit human approval for each one.

Agreed.

> Separately from that, "fuzzy" is how gettext encodes and communicates TM suggestions to a translator. It's effectively a subset/implementation of "pretranslated" suggestions, which allows for their representation in the `.po` file format even when unreviewed.

I see concepts as similar, but I don't see "fuzzy" as a subset of "pretranslated":
1. I believe "fuzzy" strings are never perfect matches of past translations, whereas our pretranslation algorithm only takes perfect matches from the TM. 
2. "Fuzzies" are strictly avoided in production, wheres "pretranslated" are predominantly designed to be used in production.

> Therefore, if we make the automatic use-in-production of pretranslated suggestions configurable on a project+locale level, we can consider fuzzy strings as pretranslation suggestions on the input, and depending on the config, either include or not the "fuzzy" property when writing pretranslated suggestions into `.po` files.

Under this proposal, if the project is configured to store pretranslation suggestions as Unreviewed instead of Pretranslated, we need to make (non-trivial) changes to the sync process to also sync Unreviewed strings, which includes the ones suggested by humans. I find that weird. :)

> With this approach, the whole concept of "fuzzy" goes away internally, subsumed within "pretranslated". On the other hand, if "fuzzy" remains at some level internally, that seems weird and to me communicates that we have different understandings of these terms.

If we want to get rid of Fuzzy internally (because it doesn't bring much bang for the buck), I'd be more in favor of ignoring it completely than treating it as pretranslated. But I see value in landing this patch first, which makes Fuzzy almost invisible. It allows us to monitor the reactions from our users without making breaking changes internally.

Note that there's also #2017, which would help reduce the code ugliness.

eemeli (approved):
As discussed on the Pontoon call, let's take this PR as a first step, and then evaluate whether we can drop "fuzzy" some more if/as our presumptions about its usage pan out in practice.

@mathjazz Could you file an issue for that, so we won't forget?

Issue #2203, opened 2019-12-17
Allow users to customize their profile URL


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1604554](https://bugzilla.mozilla.org/show_bug.cgi?id=1604554)
Bug Reporter: Mehul Patel &lt;<iamrowdymehul@gmail.com>&gt;
CC:   @flodolo, @mathjazz

User Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:71.0) Gecko/20100101 Firefox/71.0

Steps to reproduce:

Login to Pontoon >> Navigate to your Profile picture and look at the URL in the browser. 

Actual results:

It shows like this: E.g. https://pontoon.mozilla.org/contributors/wNLHdL7tCYar1cu5Q3OvsfS6qGY/

Expected results:

It should show like: E.g. https://pontoon.mozilla.org/contributors/mehulpatel

Another example: https://mozillians.org/en-US/u/rowdymehul/

It would be great if we can do something like this. Let me if I can help in anyways.

mathjazz:
A dupe of #2483.

Issue #2112, opened 2017-12-17
Pontoon profile should have a GitHub and Bugzilla field


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1425801](https://bugzilla.mozilla.org/show_bug.cgi?id=1425801)
Bug Reporter: @flodolo
CC:   @mathjazz, @MikkCZ

Not sure if this should be a generic input field used to point to external services, but I know it would be extremely useful to have at least:
* The email address used for Bugzilla. A lot of new localizers don't have one, in some cases it doesn't match the email used on Pontoon.
* Their profile on GitHub.

mathjazz:
This has been implemented as part of the Profile redesign.

Issue #2028, opened 2015-11-15
Improving profile page


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1225109](https://bugzilla.mozilla.org/show_bug.cgi?id=1225109)
Bug Reporter: @a-polivanchuk
CC:   drashtipandya37@gmail.com, etrapani@gmail.com, @mathjazz, @MikkCZ, @jotes, safa1996alfulaij@gmail.com
Depends on:   [Bug 1243086](https://bugzilla.mozilla.org/show_bug.cgi?id=1243086), [Bug 1473761](https://bugzilla.mozilla.org/show_bug.cgi?id=1473761)

Currently homepage looks not as good as it could be, especially the contribution history part (vertical timeline).
You have to scroll down for a long time to see the whole history, which also is not very informative.
I think that graph view or even tabbed view of the contribution history would be much better and informative.

mathjazz:
The profile page redesign is now being tracked in a [project](https://github.com/mozilla/pontoon/projects/4).

PR #2464, opened 2022-03-30
Add specs for new Profile page

flodolo:


specs/0111-redesign-profile-page.md
```diff
@@ -0,0 +1,108 @@
...
+In order to support this new profile, users should be able to provide a contact email address:
+* This is the address that will be displayed in the profile page (pending personal visibility settings), and can be different from the one used to log in to Pontoon.
+* Locale managers and administrators must provide a contact email address, and its visibility can only be public, since contributors need to be able to contact them.
+
+### External accounts
+
+Users should be able to provide a series of external accounts to help locale managers and administrator contacting them:
+* Bugzilla (email address).
+* GitHub (username).
+* Telegram (username).
```
flodolo, line 99: The original proposal only had Bugzilla and GitHub.

In the feedback, someone suggested to add Matrix.

Given we often use Telegram, I wonder if it makes sense to add this as well. One concern is about visibility, as we probably want this to be private.

specs/0111-redesign-profile-page.md
```diff
@@ -0,0 +1,108 @@
...
+
+Each square in the graph represents a day, with color changing depending on the amount of contributions. It’s possible to click on a column to select a specific week, or change the year currently displayed using the links in the rightmost column.
+
+When selecting a specific time range, the data displayed below the graph is updated accordingly. This doesn’t affect the approval and self-approval graphs.
+
+## Other data in the sidebar
+
+The sidebar displays contact information for the contributor:
+* Name.
+* Description.
```
flodolo, line 64: This wasn't part of the proposal, but I noticed the mock-up had name, nick, and description.

Do we want to support also a nickname (didn't add it yet) and a description? It feels like a description would be taking away quite a bit of space.

Adding a nickname on top of the name would probably complicate the comments code, but I can see it as beneficial, as quoting some of the names is particularly complicated :-(

specs/0111-redesign-profile-page.md
```diff
@@ -0,0 +1,108 @@
+- Feature Name: Redesign of the Profile Page
+- Created: 2022-03-30
+- Associated Issue: -
```
mathjazz, line 3: I'd use #2028 here. Also related are #2203 and #2112.

specs/0111-redesign-profile-page.md
mathjazz: I thought username (nickname) would be used for the vanity URL outlined below.

No strong thoughts on Description. It could add a personal touch or a summary if the page was used as a CV. It should be limited to e.g. 50 characters.

flodolo: > I thought username (nickname) would be used for the vanity URL outlined below.

I think it might be useful to have a nickname, especially if we support mentions in comments, but I would keep it separate from the vanity URL (e.g. I might want to use my full name for that, vs something silly in the nickname).

specs/0111-redesign-profile-page.md
```diff
@@ -0,0 +1,113 @@
...
+
+Examples:
+* If a person submits 100 translations, but 20 are rejected, the approval ratio will be 80%.
+* If a person submits 100 translations, 20 are rejected, 20 are self-approved, the approval ratio will be 75% (60 over 80).
+
+The graph displays both the selected month’s ratio and the average for the previous 12 months.
+
+## Self-approval ratio
+
+Displayed in a graph, it’s the ratio between the number of translations submitted directly — or self-approved after submitting them as suggestions — over the total number of translations submitted. This data point is meaningful only for users with translation rights (translators, locale managers), as it will always be 0 for contributors.
```
mathjazz, line 39: Good point about that chart only being relevant for Translators. The layout will look odd for Contributors with one chart only. I wonder if we can change the layout to account for this (or find an alternative chart for contributors).

specs/0111-redesign-profile-page.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Approval ratio
+
+Displayed in a graph, it’s the ratio between the number of translations approved over the total number of translations submitted, excluding self-approved translations (either submitted directly as translations, or provided as suggestions and later approved by the translator).
+
+Examples:
+* If a person submits 100 translations, but 20 are rejected, the approval ratio will be 80%.
+* If a person submits 100 translations, 20 are rejected, 20 are self-approved, the approval ratio will be 75% (60 over 80).
+
+The graph displays both the selected month’s ratio and the average for the previous 12 months.
```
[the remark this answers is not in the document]
mathjazz: Reading the rest of the spec made me realize this chart is not supposed to update with the contribution chart.

I'd replace "selected" with "particular" or "specific" to avoid the confusion.

specs/0111-redesign-profile-page.md
```diff
@@ -0,0 +1,113 @@
...
+* Approval ratio.
+
+Limited means that non authenticated users, and other contributors, won’t be able to see this information when they visit the profile page. But this information will be always visible to users with translator rights for any locale in Pontoon (translators, locale managers, administrators).
+
+## Dependencies and other changes
+
+### Contact email address
+
+In order to support this new profile, users should be able to provide a contact email address:
+* This is the address that will be displayed in the profile page (pending personal visibility settings), and can be different from the one used to log in to Pontoon.
```
mathjazz, line 92: "the one used to log in to Pontoon", which will be used by default.

specs/0111-redesign-profile-page.md
mathjazz: Anecdotally speaking optional fields rarely end up being filled out.

specs/0111-redesign-profile-page.md
mathjazz: It works for most platforms, though? I'd rather start small (nickname == vanity URL) and add a separate field in a followup if there's interest from the users. It's an optional field, which rarely get filled out.

flodolo: At that point, I think it makes more sense to only support the field which was originally part of the specs (vanity URL), instead of conflating the two. Splitting them up later would only create issues.

Also, properly supporting a nickname would require more work, and there's already plenty in this spec.

specs/0111-redesign-profile-page.md
flodolo: I personally think it's OK, I wouldn't go as far as making the layout dynamic based on role, and that's data that we want to know about managers/translators.

specs/0111-redesign-profile-page.md
```diff
@@ -0,0 +1,113 @@
...
+## Contribution graph
+
+The contribution graph displays data over time about the following actions:
+* `Submissions`: translations submitted, including both suggestions and direct translations.
+* `Reviews performed`: reviews performed on pending suggestions.
+* `Reviews received`: reviews received from another translator.
+* `Submissions and reviews`: submissions + reviews performed.
+* `All activities`: submissions + reviews performed + reviews received.
+
+Only one data point is displayed at a time, and it can be selected using the dropdown list above the graph. The default view is `All activities`, since it provides a good picture of the user’s activity in Pontoon.
```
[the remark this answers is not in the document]
flodolo: Yes, that's probably a better pick.

specs/0111-redesign-profile-page.md
flodolo: That wasn't the intent behind the current spec though.

The email used to log in won't be displayed by default. We display an email only if the user provides a different email, specifically for that purpose. But I realize this would allow users to have no email displayed, which makes it impossible for managers to contact them, unless we display both login and contact email addresses.

flodolo:
> Great work!

Thanks for the review! Leaving merge to you, also not sure if you want to update the mockup before that.

Issue #2476, opened 2022-04-17
Enter-key event submits translation without accepting IME candidates first


petercpg:
## STR
1. Open Pontoon, go to any translation
2. Enter translation for the string using [IME](https://en.wikipedia.org/wiki/Input_method) (just type some random words, do not hit Enter yet)

## Expected
1. After correcting candidate words in the IME buffer, the content in IME buffer is cleared and inserted to translation field on Pontoon, but the translation was not submitted to Pontoon yet
2. Hitting another Enter or by clicking Save (when there no IME typing occur) submits the translation

## Actual
1. Hitting Enter inserts current buffer, and the translation is submitted to Pontoon altogether

## Environment:
* Microsoft Windows 11 21H2 22000.556
* Firefox Nightly 101.0a1 (2022-04-17) (64 bit)
* Both OS and Firefox locales are in zh-TW
* Only tested with Microsoft Phonetic (for zh-TW) / Microsoft IME (for ja) and it reproduces with both IMEs.

mathjazz:
Seems related to #2475. Both issues were likely caused by #2444.

mathjazz:
@eemeli Will you have cycles to look into this tomorrow?

petercpg:
Testing with latest Relay website strings, I can confirm this issue is now gone in production.

PR #2477, opened 2022-04-19
Base keyboard shortcuts on key values rather than positions

### LINKED ISSUE #2475: Hitting the Enter key in the numpad area now enters a newline instead of submitting the suggestion

eemeli:
Fixes #2475 

In #2444, the keyboard shortcuts were switched from using the deprecated KeyboardEvent `keyCode` to `code`, when they should have been switched to using `key`. Both `code` and `key` use string identifiers for the keys, but `code` is concerned with physical keys while `key` with logical ones. This means that `code` has different values for the different return keys; `Enter` and `NumpadEnter`.

@mathjazz Are you able to test if this does anything for #2476, or if that's a different issue?


mathjazz (approved):
LGTM.

Also tested with keyboard shortcuts I have on my physical keyboard and it seems to work fine.

PR #2488, opened 2022-04-21
Clarify visibility of data in the profile page

mathjazz:
- Email address should never we visible to all users.
- Replace the term "limited visibility" with more self-explanatory "users with translator rights".
- Move Latest activity section closer to the Other data in the sidebar section.
- Use the term "translator rights" consistently.
- Move email address visibility details from the Email address section to the Visibility section.


flodolo (approved):
Thanks, looks a lot clearer this way.

PR #2492, opened 2022-04-22
Delete TM entries of implicitly rejected translations

### LINKED ISSUE #2491: When implicitly rejecting a translation, delete any of its TM entries
We delete TM entries of translations that get explicitly unapproved or rejected.

However, we keep the TM entries when translations get rejected implicitly by approving a different suggestion or submitting a new translation. That means it's currently also possible to delete a translation, and the TM entry remains.

As part of this patch, we should also cleanup all TM entries that belong to rejected translations.

mathjazz:
Fixes #2491.

With this patch we delete Translation Memory entries of translations that get implicitly rejected:
- when a new approved translation is submitted directly or another suggestion gets approved
- as part of the find & replace batch action
- during file upload, which overrides them
- during sync, which overrides them

Existing Translation Memory entries of implicitly rejected translations are deleted by a data migration.

Sadly we don't have good unit tests in place. I did test pretty heavily, though.


pontoon/base/models.py
```diff
@@ -3306,6 +3306,12 @@ def save(self, update_stats=True, failed_checks=None, *args, **kwargs):
                     translation=t,
                 )
 
+            # Remove any TM entries of old translations that will get rejected.
+            # Must be executed before translations set changes.
+            TranslationMemoryEntry.objects.filter(
+                translation__in=approved_translations
+            ).delete()
```
eemeli, line 3313: This is fine, but I did have to double-check that `approved_translations` is the right thing to filter by; while the name isn't strictly speaking misleading, it doesn't quite tell the whole story either.

eemeli (approved):
Non-blocking comments inline.

PR #2501, opened 2022-04-28
Update JS dependencies, including Fluent

### LINKED ISSUE #2453: After enabling `strict: true` in TS config, updating to TS 4.6 reveals new errors
As identified in https://github.com/mozilla/pontoon/pull/2439#discussion_r821712965, there's something a bit fishy in the app action/dispatch types. This should first get a workaround (probably a bunch of `@ts-expect-error` comments) so that we have a clear update path to TS 4.6, and then a more proper fix.

For now, users of VS Code may need to manually set the TS version to use the one set in `package-lock.json` for the workspace to prevent complaints, rather than letting VS Code use its default, i.e. the latest release.

eemeli:
~This is a WIP as projectfluent/fluent.js#590 is needed to avoid a useless cast.~ Edit: Done.

The `@fluent/react` update in particular required quite a bit of work adding a mock LocalizationProvider to a number of tests, as it now throws if `Localized` is used outside a provider.

~Will also include a Jest 28 update if kulshekhar/ts-jest#3453 is resolved soonish.~ Edit: Not included here.

It would also appear that this fixes #2453: no compiler errors show up with the `typescript@4.6.3` version included here.


mathjazz (approved):
Nice work!

PR #2511, opened 2022-05-04
Add ability to pretranslate subset of locales only

### LINKED ISSUE #2509: Add ability to enable pretranslation for a subset of locales only
Currently, pretranslation is a per-project setting - when enabled, all locales enabled for the project get pretranslated.

Instead, we'd like pretranslation to be a per-project-locale setting: admins should select which of the locales enabled for the project should be pretranslated.

mathjazz:
Fixes #2509 by adding a locale selector widget to project Admin to select locales to be pre-translated.

Project-level pretranslation switch remains in place.


pontoon/administration/views.py
```diff
@@ -164,16 +168,29 @@ def manage_project(request, slug=None, template="admin_project.html"):
...
+                locales_pretranslate_form = form.cleaned_data.get(
+                    "locales_pretranslate", []
+                )
+                locales_pretranslate_pks = [l.pk for l in locales_pretranslate_form]
+                project_locales.exclude(locale__pk__in=locales_pretranslate_pks).update(
+                    pretranslation_enabled=False
+                )
+                project_locales.filter(
+                    locale__pk__in=locales_pretranslate_pks,
+                ).update(pretranslation_enabled=True)
```
eemeli, line 192: non-blocking, nitpick: This _feels_ like it's doing two loops where one might be enough. But optimising this almost certainly isn't worthwhile.

Also, is the code style formatter doing something weird here as these two calls place newlines in different places?

pontoon/pretranslation/tasks.py
```diff
@@ -37,19 +37,31 @@ def pretranslate(self, project_pk, locales=None, entities=None):
...
     if locales:
         locales = project.locales.filter(pk__in=locales)
     else:
         locales = project.locales
 
     locales = (
-        locales.filter(project_locale__readonly=False)
+        locales.filter(
+            project_locale__readonly=False, project_locale__pretranslation_enabled=True
+        )
```
eemeli, line 52: Just to check: This means that all projects which currently use pretranslation would have it disabled by this PR, right?

eemeli (changes requested):
I think this needs an additional migration setting `ProjectLocale.pretranslation_enabled` to `True` for users that currently do have pretranslation enabled. Otherwise, this would be a breaking change for them.

Or we need to somehow ascertain that no such users exist, which might not be possible.

pontoon/administration/views.py
mathjazz: Ah, there's an extra comma.

pontoon/pretranslation/tasks.py
mathjazz: Nice catch, there might be 3rd party instances that already use the feature.

pontoon/administration/views.py
mathjazz: Re the optimization - how can I get to the same result with just one SQL UPDATE?

eemeli: I could figure out the SQL for that (the assigned value can be an expression, not just a constant), but I'm not sure how to get there through Django's ORM.

mathjazz: Makes sense. I'd opt for sticking with Django ORM unless absolutely necessary.

eemeli: Agreed.

PR #2514, opened 2022-05-05
Move unsaved changes from Redux to React Context

eemeli:
~Built on #2513, will be rebased once that's merged; only the last commit is actually meant for this PR.~ Edit: Done.

Unsaved-changes are a bit tricky, as they're checked in a number of event handlers. To allow for that usage to continue to not need the parent component to get re-rendered on all unsaved-changes changes, a second ref provider is used.

The popup component is renamed as `UnsavedChangesPopup` and the ignore-handler is moved to the provider.


translate/src/core/editor/actions.ts
```diff
@@ -240,7 +240,7 @@ export function endUpdateTranslation(): EndUpdateTranslationAction {
 /**
  * Save the current translation.
  */
-export function sendTranslation(
+export function sendTranslation_(
```
eemeli, line 243: With the re-export via `default`, this was proving impossible to search for as there are too many components with `sendTranslation` as a prop.

mathjazz: Can we find a more meaningful name? Maybe `sendTranslationEditor`?

mathjazz (approved):
I tested and didn't sport any issues.

Once #2513 is merged, this one can be merged, too.

translate/src/core/editor/actions.ts
eemeli: I can rename it if you want, but really I'd rather work towards getting rid of the whole action, at which point it'll make sense to reconsider the name of its replacement.

Also, fair warning: This doesn't turn out to be the only case where a wrapping function has the same name as the API method. The worst is probably with `addComment`, which ends up with a second layer of wrapping, resulting in `addComment__`. I'm currently working on untangling that as a part of the fixes for #2519.

PR #2517, opened 2022-05-06
Unbundle API methods from class structures

eemeli:
Built on #2514, will be rebased once that'd merged; only the last commit is actually meant for this PR.

This is a pretty thorough cleanup of the API methods:
- Code is moved from `src/core/api/` to `src/api/`. Import orders are fixed globally to account for the re-alphabetisation, and a note is added to the README.
- The `APIBase` class is replaced by two custom wrappers `GET()` and `POST()`, in `api/_base`.
- Methods in classes that used to extend `APIBase` are now exported as standalone functions, with globally unique names. Common code is moved up to avoid duplication.
- A few methods are moved away from `api/entity`:
  - `fetchTeamComments` is now in `api/comment` (was `getTeamComments`)
  - `fetchOtherLocales` is now in `api/other-locales` (was `getOtherLocales`)
  - `fetchTerms` is now in `api/terminology` (was `getTerms`)
- Also in `api/entity`, the previous `getEntities` is split into three: `fetchEntities`, `fetchEntitiesById`, and `fetchEntityIds`
- In `api/translation`, `setTranslationStatus` replaces the previous `approve`, `unapprove`, `reject` and `unreject` methods.
- Where appropriate, type definitions are moved to the corresponding `api/` files. All API responses are typed and do not use `any`.
- The files `api/index.ts` and `api/types.ts` are dropped.
- The dependency on `lodash.isempty` is removed as it's made unnecessary.

The behaviour of the request aborts introduced in #1290 needed to be changed, as this refactor identified it to be rather fragile: Using a single AbortController for multiple endpoints and calling it before each such request meant that when the current Entity changes, some of those requests can get aborted. Now abort controllers are per-URL, and may be enabled via the `singleton` option given to `GET()` and `POST()`. For machinery requests, the abort calls come from a single source, and the previous behaviour is maintained via the `abortMachineryRequests()` function.


translate/src/api/comment.ts
```diff
@@ -0,0 +1,59 @@
+import { GET, POST } from './_base';
```
mathjazz, line 1: File names starting with `_` are a new pattern within the `translate` app I believe.

Care to explain when do we use it in the `README`?

mathjazz (approved):
Great work! The `api` module is now much much cleaner.

Tested locally and didn't spot any issues.

PR #2518, opened 2022-05-06
Remove all default exports

eemeli:
Built on #2517, will be rebased once that's merged; only the last two commits are actually meant for this PR.

Default exports and in particular re-exports are made of pain. They needlessly complicate dependency graphs and make it harder to find all uses of the re-exported things.

So this PR gets rid of all of them, switching to named exports instead. And it adds an ESLint tule to prevent them from being used at all.

As a side benefit, this allows all of the 15 or so dependency loops to be resolved, which is nice. They weren't horrible, but a bit messy.

Where applicable, nested components are named more clearly as Foo and FooDialog; this still leaves the following four Foo + FooBase component pairs, which should be cleaned up separately:
- `modules/entitydetails/components/EntityDetails`
- `modules/genericeditor/components/PluralSelector`
- `modules/history/components/Translation`
- `modules/search/components/SearchBox`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -55,6 +55,7 @@
         "enzyme": "^3.11.0",
         "enzyme-adapter-react-16": "^1.15.6",
         "eslint": "^8.9.0",
+        "eslint-plugin-import": "^2.26.0",
         "eslint-plugin-react": "^7.22.0",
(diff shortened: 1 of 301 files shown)


mathjazz (approved):
Nice job.

Tested locally and didn't spot any issues.

PR #2520, opened 2022-05-07
Add support for list query parameter

### LINKED ISSUE #2049: Add an option to link to multiple strings

eemeli:
Built on #2518, will be rebased once that's merged; only the last two commits are actually meant for this PR.

Fixes #2049 by adding support for a `list=13,42,99` URL query parameter. This is then used by the review-notification links, as discussed previously in https://github.com/mozilla/pontoon/pull/2456#discussion_r831972129.

In the UI, the list view is indicated by a custom search-box icon, [`list-alt`](https://fontawesome.com/v5/icons/list-alt?s=solid):

Screenshot 2022-05-07 at 15 02 06

Internally, the argument uses the existing `entity_ids` payload parameter used by batch actions, with which `/get-entities/` returns an unpaginated list of matching entities. It's not possible to use the parameter together with any other filters like `status` or `search`. This allows for re-merging the just-added `fetchEntitiesById()` API wrapper method into `fetchEntities()`.

An errant missing FiltersPanel FTL string is added, as we're meddling here with that code in any case.

### DIFF
--- a/pontoon/projects/management/commands/send_review_notifications.py
+++ b/pontoon/projects/management/commands/send_review_notifications.py
@@ -1,6 +1,5 @@
 from collections import defaultdict
 from datetime import timedelta
-from urllib.parse import urlencode
 
 from django.core.management.base import BaseCommand
 from django.db.models import Q
@@ -13,7 +12,7 @@ from pontoon.base.models import Translation
 class Command(BaseCommand):
     help = "Notify translators about their newly reviewed suggestions"
 
-    def get_description(self, author, notifyData):
+    def get_description(self, notifyData):
         desc = "Your suggestions have been reviewed:\n<ul>"
 
         for (locale, project), (approved, rejected) in notifyData.items():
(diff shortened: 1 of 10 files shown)


mathjazz:
I tested this and I couldn't make it work.

The following are links to my existing strings:
http://localhost:8000/sl/firefox/all-resources/?string=74127
http://localhost:8000/sl/firefox/all-resources/?string=74128

When I request both of them using the `list` parameter:
http://localhost:8000/sl/firefox/all-resources/?list=74127,74128

I get redirected to the first string:
http://localhost:8000/sl/firefox/all-resources/?string=74127

What am I doing wrong?

mathjazz (approved):

Issue #2050, opened 2016-12-22
Show indication for good results from Translation Memory


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1325525](https://bugzilla.mozilla.org/show_bug.cgi?id=1325525)
Bug Reporter: @tomer
CC:   @ItielMaN, @mathjazz

Sometimes you start translating something that is very similar to previous translations. For example, I'm currently translating Firefox for iOS, which has large amount of strings that are identical to strings from the Android equivalent or the desktop browser. 

Currently we have few options:
* Translate everything manually.
* Use Translation Memory for hints, manually copy strings that are identical to previous translations. 
* Use Translation Memory to blind translate automatically, maybe never actually make sure that everything is correct. 

The second option is a good balance between automatic translation and manual, and  allow us to make sure we won't mess with bad translation as the interaction require human review. 

I'd suggest adding a visible indication when there is a good chance that the same string have been already translated in the past. I mean that there is an item in the Machinery tab that received from a trusted source (we trust only ourselves, so only Mozilla stings are trustfully) with a good score (no less than 100 percents!), a visible icon should appear in the panel above the tabs, with an option to copy the translation in one click. 

I'd also suggest highlighting good results, because most of the times we WANT the translator to at least read the suggestions, as the translated terms are already there. 

Having better TM usage will make sure that we do less mistakes and keep the translations identical between platforms and different resources, and at the same time speed up translations.

eemeli:
Starting to consider this one. Seems to me that we ought to get the desired outcome if, when a new string is added, we were to immediately generate a new suggestion from a 100% TM match. That would then also show up in the edit field for translators.

@flodolo, @mathjazz Do you see any reason why that would be a bad idea?

mathjazz:
When would such suggestion be created?

eemeli:
Primarily I was thinking of doing it when new entities are created. May need a couple of other places as well, such as when a project is added or a new locale is enabled for a project.

mathjazz:
OK, so this would effectively be a special use case of pre-translation (enabled across all projects and locales implicitly, and only engaging TM). It would also affect Unreviewed counts.

One alternative would be to visually "fake" such suggestions in the History panel (and the editor?).

flodolo:
This report is from a time when the MACHINERY tab wasn't displayed by default, so it would be easy to miss a 100% match. 

[...]

PR #2527, opened 2022-05-11
Simplify `Entity.for_project_locale` and `Entity.map_entities`

eemeli:
While working on #2050, I noticed that there were a couple of minor inefficiencies in `Entities` methods; they're fixed here.

This also drops the unused always-false `visible` field from the JSON output.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -2917,15 +2917,17 @@ class Entity(DirtyFieldsMixin, models.Model):
 
         # Filter by search parameters
         if search:
-            # Split search string on spaces except if between non-escaped quotes.
             search_list = utils.get_search_phrases(search)
-            search_query_list = [(s, locale.db_collation) for s in search_list]
 
             translation_filters = (
-                Q(translation__string__icontains_collate=search_query)
+                Q(translation__string__icontains_collate=(search, locale.db_collation))
                 & Q(translation__locale=locale)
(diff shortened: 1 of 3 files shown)


pontoon/base/models.py
```diff
@@ -3021,11 +3014,6 @@ def map_entities(
                     "obsolete": entity.obsolete,
                     "translation": translation_array,
                     "readonly": entity.resource.project.projectlocale[0].readonly,
-                    "visible": (
-                        False
-                        if entity.pk not in visible_entities or not visible_entities
-                        else True
-                    ),
```
mathjazz, line 3028: Took me a while to investigate where this comes from - good old (deprecated) in-context l10n!

pontoon/base/models.py
```diff
@@ -2917,15 +2917,17 @@ def for_project_locale(
...
-            search_query_list = [(s, locale.db_collation) for s in search_list]
 
             translation_filters = (
-                Q(translation__string__icontains_collate=search_query)
+                Q(translation__string__icontains_collate=(search, locale.db_collation))
                 & Q(translation__locale=locale)
-                for search_query in search_query_list
+                for search in search_list
+            )
+            translation_matches = entities.filter(*translation_filters).values_list(
```
mathjazz, line 2927: 👍

mathjazz (approved):

PR #2530, opened 2022-05-16
Update pip-tools to v6.6.1

### LINKED ISSUE #2529: Dev dependencies: pip-tools v5.5.0 not compatible with pip v22.1
The pip-tools version currently used by Pontoon is no more compatible with the latest pip version. The issue is fixed by pip-tools v6.6.1.

More info: https://github.com/jazzband/pip-tools/issues/1617

flozz:
Closes #2529


flozz:
Hum... I do not know how to resolve the `tomli` version issue... :/

mathjazz:
Thanks for the patch!

Can you try upgrading `black`, which seems to require the other `tomli`?

flozz:
I tried but I still have tomli==1.2.2 in the generated requirement file...

I found this packages that depends en tomli
```
black → setuptools_scm[toml]>=6.3.1 → tomli>=1.0.0
pip-tools → pep517 → tomli >=1.1.0
```
But I don't know how tomli 2.x pop in the file :/

flozz:
@mathjazz If I edit the `dev.txt` file by hand to set the tomli version to 1.2.2, the next time I regenerate the file, it is kept... Do you think it is OK to do it like this?

flozz:
> What happens if you run `make requirements` afterwards?

The tomli version stay the same.

The only changes I have are the order of the packages in the file and the comment line 2 (changes I do not included in the PR as they are not relevant):

Capture d’écran de 2022-05-18 10-09-08

mathjazz:
That looks good!

I guess the new output is due to the new pip-compile (part of pip-tools) version. Please update the branch with the output of make requirements and we're good to go.

flozz:
Done. And checks are ok :)

PR #2536, opened 2022-05-24
Various refactors

eemeli:
~Built on top of #2534, so will need a rebase once it's merged. Actual changes here are the last five commits.~ Edit: Done.

This is mostly wrapping up "obvious" fixes that were identified while working on the preceding PR, but which could be left out from it.
- An `EntityView` context replaces the preceding `PluralForm` context, and includes a direct references to the current entity. This allows for the `useSelectedEntity()` hook users not to need to loop through the list on entities when looking for the current one, and collects the related `entity` and `pluralForm` states into one view.
- When the entity has no plural forms, some of the API endpoints expect a `pluralForm` value of `-1`, while others expect a `0`. This determination is now made much closer to the API endpoint in question, and with later refactors should become even less visible from the UI code, where there's now a separate `hasPluralForms: boolean` flag.
- History and notifications now work through contexts, rather than Redux.
- The selected-entity dependent context providers are now all in `Entity`, and none are in `EntityDetails`.


mathjazz (approved):
@eemeli I suggest we merge this and rebase #2537 (which might be block on the UI changes for a while). Should have reviewed it separately after all.

eemeli:
I'll rebase #2537 once I can, hopefully later today.

PR #2537, opened 2022-05-24
Automatically fill editor for missing values with a perfect TM match

### LINKED ISSUE #2050: Show indication for good results from Translation Memory

eemeli:
~Built on top of #2536, so will need a rebase once it's merged. Actual changes here are the last two commits.~ Edit: Done.

Fixes #2050, so ping its original reporter @tomer.

Screenshot of the editor with the "Perfect match from translation memory" indicator shown:
Screenshot 2022-05-25 at 0 09 02

UX is based on discussions with @mathjazz, as a noticeable but discreet indicator that should not move other elements in the view when its shown or hidden. Currently it's only shown for this specific case, but while playing around with it, it seems like it could work for all machinery matches. I'd be very happy to accept suggestions for better/alternative wording for it.

The indicator is only shown when in the "simple" view, and the automatic fill also only happens when the string may be initially shown in that view.

### DIFF
--- a/translate/public/locale/en-US/translate.ftl
+++ b/translate/public/locale/en-US/translate.ftl
@@ -206,6 +206,12 @@ editor-KeyboardShortcuts--copy-from-next-helper = Copy From Next Helper
 editor-KeyboardShortcuts--copy-from-next-helper-shortcut = <mod1>Ctrl</mod1> + <mod2>Shift</mod2> + <accel>Down</accel>
 
 
+## Editor machinery source indicator
+## Shown when a perfect match is provided automatically from translation memory
+
+editor-MachinerySourceIndicator--text = Perfect match from translation memory
+
+
 ## Editor New Contributor Tooltip
 ## Renders the guidelines for new contributors
 
--- a/translate/public/locale/en-US/translate.ftl
+++ b/translate/public/locale/en-US/translate.ftl
@@ -209,7 +209,7 @@ editor-KeyboardShortcuts--copy-from-next-helper-shortcut = <mod1>Ctrl</mod1> + <
 ## Editor machinery source indicator
 ## Shown when a perfect match is provided automatically from translation memory
 
-editor-MachinerySourceIndicator--text = Perfect match from translation memory
+editor-MachinerySourceIndicator--text = <stress>100%</stress> MATCH FROM TRANSLATION MEMORY
 
 
 ## Editor New Contributor Tooltip
--- a/translate/src/context/Editor.tsx
+++ b/translate/src/context/Editor.tsx
@@ -9,6 +9,7 @@ import React, {
 } from 'react';
 
 import type { SourceType } from '~/api/machinery';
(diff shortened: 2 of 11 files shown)


mathjazz (commented):
Great work with the refactor - thanks to which the actual patch looks really trivial. Code looks good and I didn't sport any issues while testing. r+

I need to think more about changes to the UI. I'm considering moving the Perfect match message under the editor, into the History panel.

mathjazz:
I propose the following design changes:
Screenshot 2022-06-27 at 20 34 36

Here's the patch that needs to be applied:
https://gist.github.com/mathjazz/dbdf54626b3ae58f480fdd5d44826d2f

mathjazz:
I see your point.

@flodolo suggested moving the banner below the text area. Here's one take that also addresses the jumping text editor (I've updated the diff, didn't want to push to your branch directly).

Screenshot 2022-06-29 at 21 09 05

eemeli:
That kinda has the same issue, but with slightly less impact. :/

How about this: Let's apply your second design, and then see if we get any complaints?

mathjazz:
@flodolo Did you want the banner to appear above or below the editor menu (dark area where the Save button is)?

Regardless of the position, we should be able to prevent jumping of the UI completely by decreasing the height of the editor or the history panel when the banner is visible.

flodolo:
> @flodolo Did you want the banner to appear above or below the editor menu (dark area where the Save button is)?

Below the text field. Above it's easier to ignore, below is in the way between the entered text and the save button.

mathjazz:
I've rebased and included the design update as shown in the latest screenshot from the last week.

mathjazz (approved):
From my perspective this is good to go.

@eemeli One last look?

mathjazz:
Merging as agreed on the team call today.

Issue #2157, opened 2018-12-22
Users should have a way to contact existing locale managers


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1516164](https://bugzilla.mozilla.org/show_bug.cgi?id=1516164)
Bug Reporter: @flodolo
CC:   @mathjazz

I don't think there's an easy way for contributors to get in touch with existing managers or translators, even if it's something we suggest to do when they ask for new projects.

There's an INFO tab, but few of them customize it. Maybe we should have a TEAM tab, that lists all managers and translators for a team.

We could also have a "contacts" tab, with a form to get in touch with the team, accessible only to registered users.

mathjazz:
At least as the first step, I'd group contributors by role in the Team Contributor tab.

Issue #2305, opened 2021-08-20
Provide new contributors with guidelines before adding their first suggestion in a locale


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1726881](https://bugzilla.mozilla.org/show_bug.cgi?id=1726881)
Bug Reporter: @flodolo
CC:   @Delphine, @mathjazz, reza.hbi@gmail.com

Right now, a new contributor can start translating without looking at style guides, or info on the locale (which is not exactly visible).

It would be valuable for locale managers to be able to set up a message, explaining what the workflow for the locale is, if there are style guides, etc.

For example:
- Locale managers for `it` set up this type of information, and enable this INTRO feature for their locale (maybe it's enabled automatically if the information is provided). The message should allow links, potentially we could display it as a subsection of the locale INFO tab.
- When users are trying to contribute for the first time to a specific locale, if the INTRO feature is enabled, we display a message pointing the user to the locale INFO tab (we could also hint at the fact that someone needs to approve their suggestion). The user can dismiss the notification, and it won't be displayed again.

PR #2542, opened 2022-05-31
Spec Guidelines for new contributors

mathjazz:
This is a spec for #2305.

~I made one significant change from what is documented in the original issue: in addition to locale-specific content, which needs to be set up by locale managers, the intro message also consists of generic content, which is the same across all locales. That way sending the message doesn't depend on managers, many of which I'm afraid will not set up the locale-specific content.~

~We should still encourage managers to fill-out the Team Info tab with relevant info. Sending messages to them when new contributors join might encourage them yo do so implicitly (because otherwise they'll need to repeat similar messages all over again).~

As agreed on the team call last week, I'm flagging all PMs for a review.

@flodolo r?
@Delphine r?
@peiying2 r?
@bcolsson r?


bcolsson (commented):
Left one comment about team info tab. Nothing major as a concern stands out, but may want to follow-up with locale managers to update team info tabs before feature roll out.
(I'll leave it for others with a bit more experience to approve.)

specs/0112-new-contributor-guidelines.md
```diff
@@ -0,0 +1,25 @@
...
+
+# Motivation
+
+It has been suggested on various occasions and from various sources, including at the localizer workshops, that the onboarding experience for new contributors needs improvements.
+
+Pontoon homepage and tour lead users towards making their first contribution, but they don't get much guidance since that point nor are they connected with their team. Locale Managers also don't get any direct information when new contributors join their teams.
+
+# Feature explanation
+
+After the user makes her first suggestion in Pontoon, we send an in-app notification with the following goal:
```
Delphine, line 17: "her first suggestion" > "their"?

specs/0112-new-contributor-guidelines.md
```diff
@@ -0,0 +1,25 @@
...
+
+- Welcome user to Pontoon.
+- Clarify what will happen with the suggestion.
+- Link to Pontoon docs.
+- Explain how to get in touch with locale managers.
+- If set, include links to team Info tab and locale style guide.
+
+After the user makes her first suggestion in a particular locale, we send an in-app notification with just the locale-specific bits of the above defined message. We also send an in-app notification to locale managers, informing them about the new user.
```
Delphine, line 25: "her first suggestion" = "their" here as well?

specs/0112-new-contributor-guidelines.md
```diff
@@ -0,0 +1,25 @@
...
+
+# Feature explanation
+
+After the user makes her first suggestion in Pontoon, we send an in-app notification with the following goal:
+
+- Welcome user to Pontoon.
+- Clarify what will happen with the suggestion.
+- Link to Pontoon docs.
+- Explain how to get in touch with locale managers.
+- If set, include links to team Info tab and locale style guide.
```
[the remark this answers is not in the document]
Delphine: Good point. Agree here as well

flodolo:
This is collapsing two separate issues into one, and I don't think it should:
1. Guidelines for new contributors.
2. Guidelines for new contributors to a specific locale (which is what #2305 was really about).

I think these specs cover 1, but:
* I disagree that we should wait for the user to make a contribution to send the in-app notification. We should send it right after creating the account. 
* We should actually discuss the text and links to use for that notification, unless you plan to make it dynamic (i.e. admins can set it up in Pontoon), which seems like a good idea for maintainability. Either way, at a minimum specs should specify where that description is stored, and how it's going to be updated when necessary.

For 2 (locale specific):
* We should only send it if a locale manager set up specific instructions. They're likely going to be in the native language, and that's one more reason not to mix 1 and 2.
* I think using notifications for this might be too passive. Users will sign up, put in a bunch of random translations before even realizing there are notifications in Pontoon. Ideally, we should show a notification to the user before they try to localized something into that language, and record when the notification has been dismissed.

specs/0112-new-contributor-guidelines.md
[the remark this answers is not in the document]
mathjazz: Could you please clarify that do you mean with the "template" info here?

mathjazz:
> This is collapsing two separate issues into one, and I don't think it should:
> 
>     1. Guidelines for new contributors.
> 
>     2. Guidelines for new contributors to a specific locale (which is what [Provide new contributors with guidelines before adding their first suggestion in a locale #2305](https://github.com/mozilla/pontoon/issues/2305) was really about).

Correct. As explained above, this is intentional, because we don't have locale-specific guidelines. On top of that, joining the two will more likely prevent locale managers to repeat generic content in locale-specific messaging.

> I think these specs cover 1, but:
> 
>     * I disagree that we should wait for the user to make a contribution to send the in-app notification. We should send it right after creating the account.

If we split generic and locale-specific guidelines we should do that, yes. (Otherwise we obviously can't).

>     * We should actually discuss the text and links to use for that notification, unless you plan to make it dynamic (i.e. admins can set it up in Pontoon), which seems like a good idea for maintainability. Either way, at a minimum specs should specify where that description is stored, and how it's going to be updated when necessary.

Agreed about discussing the actual content, but I wanted to first reach an agreement on the feature itself.

> For 2 (locale specific):
> 
>     * We should only send it if a locale manager set up specific instructions. They're likely going to be in the native language, and that's one more reason not to mix 1 and 2.
> 
>     * I think using notifications for this might be too passive. Users will sign up, put in a bunch of random translations before even realizing there are notifications in Pontoon. Ideally, we should show a notification to the user before they try to localized something into that language, and record when the notification has been dismissed.

What about something like the GH does?

Screenshot 2022-06-02 at 10 11 03

flodolo:
> What about something like the GH does?

That would definitely work for me. We should probably put a limit to the amount of content that locale managers can put in there, and ask them to use the INFO tab if they need more things.

mathjazz:
I'll refactor the spec and add a mockup.

specs/0112-new-contributor-guidelines.md
```diff
@@ -1,25 +1,36 @@
...
+Pontoon homepage and tour lead users towards making their first contribution, but don't provide any team-specific information.
+
+On top of that, new users don't get connected with their team, and team managers don't get any explicit information about new contributors joining their teams.
 
 # Feature explanation
 
-After the user makes her first suggestion in Pontoon, we send an in-app notification with the following goal:
+## Translate view tooltip
+
+Before users submits their first suggestion to any team, a tooltip appears near the text editor, with the following content:
```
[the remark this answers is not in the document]
mathjazz: We need a tangible value proposition for that, because it's not trivial to implement editing ability. And even if we had it, judging by the Info tab edits we can expect the majority of locales will not use it.

Why can't we put the same information into the resources we link to, like GitHub does (Info tab, Style Guides)?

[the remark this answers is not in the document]
flodolo: Because, for one, I would like to display that message in my own language, and be able to try different variants if it doesn't work, or add/remove information.

GitHub doesn't have anything localized, and it doesn't deal with localization as a primary function, so I don't think we should use it as a reference besides UI/UX ideas.

If we can't have it localizable, then we should keep this info to a minimum, and just link to the team info panel. 
* We link to style guides, but very few locales have one, and it's not necessarily maintained (which is a different conversation…).
* We link to contributors, but I don't believe that's helpful for newcomers, as there's no way to only display the managers for that locale (lot of people listed, including admins).

specs/0112-new-contributor-guidelines.md
```diff
@@ -0,0 +1,36 @@
...
+
+On top of that, new users don't get connected with their team, and team managers don't get any explicit information about new contributors joining their teams.
+
+# Feature explanation
+
+## Translate view tooltip
+
+Before users submit their first suggestion to a particular team, a tooltip appears near the text editor, with the following content:
+
+It looks like you haven't contributed to this locale yet. Be sure to:
```
flodolo, line 23: I'd replace this with

```
It looks like you haven't contributed to this locale yet. 

Make sure to check the <a href="/locale-code/info">team information</a> page before starting, as it might contain important information and language resources.

If you want to contact the team managers, search for “manager” in the list of <a href="/locale-code/contributors"> active contributors for this locale</a>.
```

specs/0112-new-contributor-guidelines.md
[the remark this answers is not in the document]
mathjazz: That sounds more like we should finally start [localizing](https://github.com/mozilla/pontoon/issues?q=is%3Aopen+is%3Aissue+label%3Al10n) Pontoon or at least the translate view, where most of the i18n work has already been done.

I agree about the issue with the Contributors tab. To include that link, we should fix #2157 first.

I would keep the style guides, though, but only where applicable, obviously. The concerns you raise are valid for the Info tab, too. And, we already link to them from the locale dashboard.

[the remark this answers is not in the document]
flodolo: > I would keep the style guides, though, but only where applicable, obviously. The concerns you raise are valid for the Info tab, too. And, we already link to them from the locale dashboard.

Good point on the team info. Maybe we should add information in English for all locales if it's empty 🤔 

I really think that we shouldn't link to the style guides, because IMO that was done in the wrong way and it might even contain misleading information :-(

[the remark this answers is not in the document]
mathjazz: Should we remove the style guides links from the DB (i.e. dashboards), too?

specs/0112-new-contributor-guidelines.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Notification for team managers
+
+After users submit their first suggestion to a particular team, an in-app notification is sent to team managers, informing them about the new user. Example text:
+
+John Smith has made their first contribution to Locale X (locale-code). Please welcome them to the team!
+
+# Prerequisites
+
+1. Before launch, we give locale managers a heads-up of how the Team Info tab content will be used in the new feature and offer them an opportunity for any updates.
```
flodolo, line 37:
```suggestion
1. Before launch, we give locale managers a heads-up of how the Team Info tab content will be used in the new feature and offer them an opportunity to update it.
```

specs/0112-new-contributor-guidelines.md
```diff
@@ -0,0 +1,45 @@
...
+
+Make sure to check the <a href="/locale-code/info">team information</a> page before starting, as it might contain important information and language resources.
+
+If you want to contact the team managers, check out the <a href="/locale-code/contributors">team contributors</a> page.
+
+## Notification for team managers
+
+After users submit their first suggestion to a particular team, an in-app notification is sent to team managers, informing them about the new user. Example text:
+
+John Smith has made their first contribution to Locale X (locale-code). Please welcome them to the team!
```
flodolo, line 33:
```suggestion
John Smith has made their first contribution to Locale X (locale-code). Please welcome them to the team, and make sure to review their suggestions.
```

Maybe more explicit. Too much?

flodolo (approved):

specs/0112-new-contributor-guidelines.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Translate view tooltip
+
+Before users submit their first suggestion to a particular team, a tooltip appears near the text editor, with the following content:
+
+It looks like you haven't contributed to this locale yet.
+
+Make sure to check the <a href="/locale-code/info">team information</a> page before starting, as it might contain important information and language resources.
+
+If you want to contact the team managers, check out the <a href="/locale-code/contributors">team contributors</a> page.
```
bcolsson, line 27: Would we want to include examples of why someone might want to contact a team manager? If for the purpose of driving engagement/retention of a new potential volunteer, it might be better to phrase this as a call to action. 
Perhaps something like:

`Reach out to your language's team managers from the <a href="/locale-code/contributors">team contributors</a> page if you have questions or want to learn more about contributing.`

specs/0112-new-contributor-guidelines.md
```diff
@@ -0,0 +1,45 @@
...
+
+1. Before launch, we give locale managers a heads-up of how the Team Info tab content will be used in the new feature and offer them an opportunity to update it.
+
+2. [Group contributors by role](https://github.com/mozilla/pontoon/issues/2157) in the Team Contributors page.
+
+# Mockup
+
+Please note that the text in the bubble might be out of date. The canonical version is specified above.
+
+![](0112/mockup.png)
```
peiying2, line 45: The mockup should indicate that this is a Suggestion, not a Translation, assuming to an existing locale.

mathjazz: Nice catch!

Delphine (commented):
This looks really good!
One thing I was thinking: could PMs (admins) also get a notification that someone new has started contributing? Might be good just to be aware so we can follow-up on locales we think really need help (and perhaps with not many that active managers around).
Maybe we mentioned this already, can't recall now. Not sure what the noise ratio would be, either

peiying2:
This is completely out of scope so I don't want to distract us from completing the feature.  I wonder if there is a way to track whether the locale manager follows up with welcoming a new member. I don't doubt new contributors are most likely to reach out, but is there a way to measure the improvement of responsiveness and timeliness from managers.

flodolo:
> could PMs (admins) also get a notification that someone new has started contributing?

I would really prefer not to do that. I'm already receiving way too many notifications, and this is not actionable for l10n PMs.

What would be the benefit of knowing there are new contributors? We still can't determine if their contributions are good.

mathjazz:
> This is completely out of scope so I don't want to distract us from completing the feature. I wonder if there is a way to track whether the locale manager follows up with welcoming a new member. I don't doubt new contributors are most likely to reach out, but is there a way to measure the improvement of responsiveness and timeliness from managers.

Not directly, because managers need to reach out to contributors externally.

We can however observe how quickly managers review suggestions from new contributors. I think that's indeed out of scope, though.

mathjazz:
> > could PMs (admins) also get a notification that someone new has started contributing?
> 
> I would really prefer not to do that. I'm already receiving way too many notifications, and this is not actionable for l10n PMs.
> 
> What would be the benefit of knowing there are new contributors? We still can't determine if their contributions are good.

Since we're not aligned on this one, I'm keeping it out of spec for now.

PR #2544, opened 2022-06-03
Correctly set types of ApiFailedChecks members as optional

### LINKED ISSUE #2543: Warnings don't show up in the Translate View
When trying to approve this suggestion, nothing happens:
https://pontoon.mozilla.org/sl/mozillaorg/all-resources/?string=244396

Backend properly returns a warning, but the frontend is unable to process it.

Changes are unrelated to 835934b72368a5b71665369c5dfa94d0c4af9705.

/cc @eemeli

eemeli:
Fixes #2543, which was failing due to not validating that the API response matches the declared TS type; that lead to code making invalid assumptions, and failing to display the proceed-anyway dialog.

It wasn't only failing for approvals, but also submissions of messages with warnings.

Sorry, this one was my fault.

### DIFF
--- a/translate/src/api/translation.ts
+++ b/translate/src/api/translation.ts
@@ -6,11 +6,11 @@ import type { SourceType } from './machinery';
 export type ChangeOperation = 'approve' | 'unapprove' | 'reject' | 'unreject';
 
 export type ApiFailedChecks = {
-  readonly clErrors: string[];
-  readonly pErrors: string[];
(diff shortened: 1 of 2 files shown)


mathjazz (approved):
Thanks for the fix!

It seems like we could benefit from a bit better test coverage here to avoid similar situations in the future.

eemeli:
Agreed in general, but adding tests for TS types is difficult.

PR #2546, opened 2022-06-09
Allow reading configuration from the repository

### LINKED ISSUE #2545: Reading configuration toml file from the repository
I have a private GitLab repository and trying to setup Pontoon for localization. I have added needed SSH keys and sync is working. 

I'm trying to add a project configuration file now as I need to configure `[[paths]]`.

Problem that I'm facing is that Pontoon only allows using toml config file when repository permalink is set. I cannot do that for a private repository.

Is it possible to make Pontoon read config file from the repository itself?

Demivan:
Fixes #2545

I have no idea how to write tests for this as I'm not a Python developer :slightly_smiling_face:


mathjazz:
Thanks for the patch @Demivan!

Sorry for the delat in a review. We're on work week this week. We'll have a look ASAP.

Demivan:
@eemeli @mathjazz Sorry for pinging. Is there anything I need to improve to have this merged?

Demivan:
@eemeli @mathjazz Sorry for pinging. Is there anything I need to improve to have this merged?

mathjazz:
Putting this on my TODO. Sorry it's taking so long.

Demivan:
Thanks. I have updated branch with latest fix from master

Demivan:
@mathjazz Sorry for a ping, any chance this getting merged?

mathjazz:
Hi @Demivan!

I am very sorry for how long it took to take a look at your patch. There really are no excuses for this.

The code looks good (a unit test would be nice 😊). I also tested the patch with our projects that use l10n.toml file and didn't spot any issues.

Downloading files will not work for projects that don't set "Download prefix or path to TOML file". Are you OK with that? We should fix that, too, or at least file an issue.

Demivan:
@mathjazz Thanks for reviewing the pull request.

> Downloading files will not work for projects that don't set "Download prefix or path to TOML file"

All my projects have TOML configuration file, so this is not an issue for me personally. But that is a much bigger change and depends on how Pontoon wants to handle repositories without the config. Probably some kind of default configuration can be generated in that case.

mathjazz:
To clarify - the `Repositoy.permalink_prefix` field, which shows up as "Download prefix or path to TOML file" in the Project Admin, must be set for downloads to work. 

My understanding was that your projects are in private repos, so with this patch applied, you would leave that field empty in and take the TOML file directly from the repository. Is that not the case?

Demivan:
Yes. I leave "Download prefix or path to TOML file" field empty but I set "Configuration file (optional)" and Pontoon now reads config and strings directly from the repository.

mathjazz:
Right. And the "Download Translations" option (available from the profile menu in the translate view) doesn't work (i.e. you get an error page)?

Demivan:
Yes, it fails. But I have never actually used it :)

Looks like support for reading everything from the repository needs even more changes than I thought. But this fix is enough for me.
I don't know why Pontoon does not read from the repository by default. And I think it would be a good idea to support this use case properly at some point.

mathjazz (approved):
OK, just wanted to confirm that. :)

Downloading translation files from Pontoon is essentially a hack, because we don't have persistent storage on our instance and hence cannot rely on the repository clone to be accessible.

I filed an issue to track implementation of Download translations support for projects like yours:
https://github.com/mozilla/pontoon/issues/2755

Thanks you for the patch and sorry once again it took forever to land it.

PR #2553, opened 2022-06-18
Fix rich editor focus handling

### LINKED ISSUE #2549: When editing existing strings in rich editor cursor jumps to the end
As reported on Matrix:

"When editing existing strings which have translations for "One" and "More", after each input, the cursor jumps to the end of the string."

The problem no longer appears after reverting back to f765fa994d71f6ea21c99dcd527ead4d0c4e1ea0.

eemeli:
Fixes #2549 by not "fixing" the cursor position during manual edits. Spent ages looking in entirely the wrong place for this.

Python changes are completely separate, and were required to account for errors discovered when running the server locally with no network connectivity.

### DIFF
--- a/translate/src/context/Editor.tsx
+++ b/translate/src/context/Editor.tsx
@@ -120,7 +120,7 @@ export function EditorProvider({ children }: { children: React.ReactElement }) {
       setEditorFromInput: (value) =>
         setState((prev) => {
           if (prev.view === 'rich' && typeof value === 'string') {
-            const next = updateRichValue(prev, value, false);
+            const next = updateRichValue(prev, value, false, false);
             if (next) {
               return { ...prev, value: next };
             }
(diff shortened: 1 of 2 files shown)


mathjazz:
> Python changes are completely separate, and were required to account for errors discovered when running the server locally with no network connectivity.

We need to code without internet access more often. :)

mathjazz (approved):
Nice work!

A comment or two would be nice, but not blocking.

translate/src/context/Editor.tsx
```diff
@@ -350,10 +352,12 @@ function updateRichValue(
     }
 
     // Need to let react-dom "fix" the select position before setting it right
```
eemeli, line 354: @mathjazz This comment is actually already in the code as an explainer. Now it's just more accurate.

mathjazz: Yes, I meant on the backend.

PR #2560, opened 2022-07-04
Send new contributor notifications to team managers

mathjazz:
This is a partial fix for #2305.

Specifically, it [sends notifications to team managers](https://github.com/mozilla/pontoon/blob/master/specs/0112-new-contributor-guidelines.md#notification-for-team-managers) when users make their first contributions to the team.

Fix for [the other half of #2305 (AKA the tooltip)](https://github.com/mozilla/pontoon/blob/master/specs/0112-new-contributor-guidelines.md#translate-view-tooltip) will follow in a separate commit or PR.

I've also reworded notification type labels in the Notification subscription section of the user settings. Hopefully the labels are a bit clearer now.

PR #2564, opened 2022-07-07
Implement New Contributor tooltip

### LINKED ISSUE #2305: Provide new contributors with guidelines before adding their first suggestion in a locale

mathjazz:
This is a followup from #2560. It fixes #2305.

It [implements a tooltip](https://github.com/mozilla/pontoon/blob/master/specs/0112-new-contributor-guidelines.md#translate-view-tooltip) that appears before a new contributor tries to make a suggestion to a locale.

### DIFF
--- a/translate/public/locale/en-US/translate.ftl
+++ b/translate/public/locale/en-US/translate.ftl
@@ -209,7 +209,7 @@ editor-KeyboardShortcuts--copy-from-next-helper-shortcut = <mod1>Ctrl</mod1> + <
 ## Editor New Contributor Tooltip
 ## Renders the guidelines for new contributors
 
-editor-NewContributorTooltip--intro = It looks like you haven't contributed to this locale yet.
+editor-NewContributorTooltip--intro = It looks like you haven’t contributed to this locale yet.
 editor-NewContributorTooltip--team-info = Check the <a>team information</a> before starting, as it might contain important information and language resources.
 editor-NewContributorTooltip--team-managers = Reach out to <a>team managers</a> if you have questions or want to learn more about contributing.
 
--- a/translate/public/locale/en-US/translate.ftl
+++ b/translate/public/locale/en-US/translate.ftl
@@ -95,8 +95,8 @@ resourceprogress-ResourceProgress--missing = MISSING
 
 comments-AddComment--input =
     .placeholder = Write a comment…
-comments-AddComment--mention-avatar-alt = 
-    .alt = User Avatar 
+comments-AddComment--mention-avatar-alt =
+    .alt = User Avatar
 comments-AddComment--submit-button = <glyph></glyph>
     .title = Submit comment
 comments-Comment--pin-button = PIN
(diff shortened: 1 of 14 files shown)


translate/public/locale/en-US/translate.ftl
```diff
@@ -206,6 +206,14 @@ editor-KeyboardShortcuts--copy-from-next-helper = Copy From Next Helper
 editor-KeyboardShortcuts--copy-from-next-helper-shortcut = <mod1>Ctrl</mod1> + <mod2>Shift</mod2> + <accel>Down</accel>
 
 
+## Editor New Contributor Tooltip
+## Renders the guidelines for new contributors
+
+editor-NewContributorTooltip--intro = It looks like you haven't contributed to this locale yet.
```
eemeli, line 212:
```suggestion
editor-NewContributorTooltip--intro = It looks like you haven’t contributed to this locale yet.
```

translate/src/core/editor/components/NewContributorTooltip.tsx
```diff
@@ -0,0 +1,60 @@
...
+    !entity.readonly &&
+    user.isAuthenticated &&
+    user.isNewContributor;
+
+  return show ? (
+    <div ref={ref} className='new-contributor-tooltip'>
+      <p className='title'>👋</p>
+
+      <Localized id='editor-NewContributorTooltip--intro'>
+        <p>It looks like you haven't contributed to this locale yet.</p>
```
eemeli, line 32:
```suggestion
        <p>It looks like you haven’t contributed to this locale yet.</p>
```

translate/src/core/user/reducer.ts
```diff
@@ -69,6 +75,7 @@ export type Notifications = {
 export type UserState = {
   readonly isAuthenticated: boolean;
   readonly isAdmin: boolean;
+  readonly isNewContributor: boolean;
```
eemeli, line 78: This seems to be mixing together data that is and is not locale-dependent. Could the API and the data match instead the structure of the `managerForLocales` and `translatorForLocales` fields, with something like `contributorForLocales`? Or is there a reason why the approach proposed here is e.g. much more efficient?

translate/src/api/user.ts
```diff
@@ -64,6 +65,12 @@ export const fetchUserData = (): Promise<ApiUserData> => GET('/user-data/');
 /** Get all users from server. */
 export const fetchUsersList = (): Promise<UsersList[]> => GET('get-users');
 
+/** Check if the user is a new locale contributor. */
+export async function fetchIsNewContributor(locale: string): Promise<boolean> {
+  const results = await GET(`/is-new-contributor/${locale}/`);
+  return keysToCamelCase(results).isNewContributor;
```
eemeli, line 71:
```suggestion
  return results.is_new_contributor;
```

Do I understand right that the only reason for the `keysToCamelCase()` call here is to be able to avoid the `_` in the API field name? That seems rather clumsy.

eemeli (changes requested):
While I see now how you get to re-use the `is_new_contributor()` call by adding a separate REST API endpoint for it, is that really the best choice? As in, if that structure makes sense for determining the locale-specific contributor-ness of a user, by extension wouldn't it also make sense to use a similar structure for the translator-ness and manager-ness of a user?

To be clear, I'm not saying that those APIs ought to be changed, but trying to point out that this approach is different from what's previously used in the REST API. It would be more consistent to match that pattern, and to add a field to the user data API rather than a separate endpoint.

eemeli (approved):

PR #2565, opened 2022-07-07
Always show Age of unreviewed chart

mathjazz:
On some screen resolutions, the Age of unreviewed suggestions chart,
which is hidden by default, didn't render at all when being switch to.


flodolo:
> @flodolo Could you please verify this works on stage?

Yep, this works now 👍🏼

PR #2570, opened 2022-07-08
Control cursor position less, trusting the browser to handle it right

### LINKED ISSUE #2557: Clicking a placeholder in the editor moves the cursor to the beginning of the textbox
STR:
1. Start localizing the first few words of this: https://pontoon.mozilla.org/he/firefox/all-resources/?string=246220
2. When you get to `{ -brand-short-name }`, click it from the original string above
3. Continue typing

AR:
You find out you're typing from the beginning of the string and not when you left off before clicking the placeholder

ER:
You should be typing where you left off

eemeli:
Fixes #2557

The editor state refactoring had carried along (and possibly broken?) some prior behaviour where focus handling was actively driven using `input.setSelectionRange()`. When doing certain editing operations, in particular when mixing RTL & LTR content, this lead to surprising/unhelpful results.

The right answer here isn't to add more customisation and more active control, it's to realise that the browser will almost always do the right thing if you let it. So most of said active control is removed here, and the `setEditorSelection()` action is told to always target the "end" of the selection, as that appears to correspond to the right thing even when mixing RTL & LTR content.

Automated testing of this behaviour is not really possible outside of actual browsers, and we don't have any instrumented browser tests. So this has been and needs to be tested manually to determine if there's a sequence of user actions that produces a weird result with input focus.

### DIFF
--- a/translate/src/context/Editor.tsx
+++ b/translate/src/context/Editor.tsx
@@ -164,7 +164,12 @@ export function EditorProvider({ children }: { children: React.ReactElement }) {
           if (typeof prev.value === 'string') {
             const input = prev.activeInput.current;
             if (input) {
-              input.setRangeText(content);
+              input.setRangeText(
+                content,
+                input.selectionStart,
+                input.selectionEnd,
+                'end',
+              );
               return { ...prev, value: input.value };
             }
           } else {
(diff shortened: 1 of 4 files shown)


mathjazz:
A test is failing.

mathjazz (approved):
Tested and didn't spot any issues.

The code looks good, too.

PR #2571, opened 2022-07-13
fix(docker): copy using chown for package.json and .env

bufferoverflow:
Setup as described [here](https://mozilla-pontoon.readthedocs.io/en/latest/dev/setup.html) does not work as file permissions are wrong, using `COPY` with `--chown=pontoon:pontoon` fixes this.


eemeli (approved):
This is surprising, and I've not experienced the issues myself, but using `pontoon:pontoon` does seem like a good idea here.

Issue #2573, opened 2022-07-19
Clicking item in MACHINERY doesn't focus the translation area


flodolo:
I'm confident that, before, I could click an element in MACHINERY and press enter to confirm. Now, I need to click the SAVE button with the mouse, or manually give focus to the translation area.

mathjazz:
I just tried that with a couple of strings and couldn't reproduce. Could you provide an example?

flodolo:
OK,  I see what happened 🤦🏼 

It only happens when there's a 100% match and the content is pre-filled ([example](https://pontoon.mozilla.org/ta/firefox/all-resources/?extra=missing-without-unreviewed&string=237850)). Clicking the element in machinery hides the banner, but doesn't give focus to the text area.

PR #2581, opened 2022-07-29
Use document.activeElement rather than custom redux value to track SearchBox focus

### LINKED ISSUE #2572: After searching, clicking on the text area doesn't show the cursor
1. Search for a string
2. Click in the text area to change the translation

The cursor won't be displayed, although the text area is focused (typing will add text).

If you click to one of the translation below, then in the text area again, the cursor will reappear.

eemeli:
Fixes #2572

The timing of redux state updates was a bit off, causing the weird UX. Rather than fix the timing, let's just get rid of the state, and trust the browser to track whether the search box is focused or not.


eemeli:
Now also fixes #2573, which turned out to be highly correlated.

mathjazz (approved):
LGTM

PR #2584, opened 2022-07-31
Do not render EntityDetails or EditorMenu while still loading data

### LINKED ISSUE #2561: Flashing of unnecessary text during loading of the translate page
Recently, when the translate page is loading, "Sign in to translate" and "This is a read-only localization" started flashing in the editor menu before the page is fully loaded.

That's annoying. Let's fix that.

eemeli:
Fixes #2561 

While moving data handling from Redux to React Context, some of their default values changed from `null` to objects containing empty/initial values. This meant that checks like `!entity` or `!user` would never actually succeed, and were in fact dropped. Some of these initial values would then bleed through to the UI as e.g. `user.isAuthenticated` would be `false` and `entity.readonly` would be `true` while the relevant data was being fetched from the server, leading to an initial flash of content.

To fix this, a couple of checks are added for `entity.pk === 0` and `user.isAuthenticated === null` (a new state) to not render components while data is still being fetched.

### DIFF
--- a/translate/src/core/editor/components/EditorMenu.test.js
+++ b/translate/src/core/editor/components/EditorMenu.test.js
@@ -63,6 +63,11 @@ function expectHiddenSettingsAndActions(wrapper) {
 }
 
 describe('<EditorMenu>', () => {
+  it('does not render while loading', () => {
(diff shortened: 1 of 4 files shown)


mathjazz (approved):
Nice work!

Issue #2548, opened 2022-06-14
@user search result in Comment field limited to 5 results only


peiying2:
When responding to a comment, @user search result is limited to only 5 people when there could be a lot more. `Michael` is a very common contributor name and you may never find the right person in the drop-down list. 

Workaround appears to be searching by last name, and you have better luck in finding the person you are looking for.

mathjazz:
The list of results will need to remain limited, because we might have too many matches. GitHub here also seems to only show 5 matches.

Looking at this example, we should improve sorting:
<img width="320" alt="Screenshot 2022-06-14 at 17 38 50" src="https://user-images.githubusercontent.com/626716/173694753-57a488ff-7c8a-4096-8e7e-8c89da6e98a5.png">

eemeli:
We should rethink the back-end for this. At the moment, on pontoon.mozilla.org we're fetching 780 kB compresed / 2.85MB raw data from `/get-users/` during/after the translate view's initial load, and then only using it for @-completion when adding a comment. This is... sub-optimal.

As an alternative, we could/should add query parameters to the API endpoint and only call it when actively doing @-completion, which would allow us to filter and prioritise the list according to the current locale, project and even entity. This might add a negligible delay to the dialog when editing, but would also cut 35% from Pontoon's transferred filesize.

PR #2588, opened 2022-07-31
Only fetch /get-users/ when adding a comment

eemeli:
This is a patch for the bad initial loading behaviour identified in #2548. It's not a fix for the issue, but it's a step in a better direction.

With this change, data for @-mentionable users is only fetched when the first AddComment component is rendered, i.e. when one of the following happens:
- an authenticated user clicks on a "COMMENT" button next to a History translation,
- the "REQUEST CONTEXT or REPORT ISSUE" button is clicked, or
- when then team comments tab is opened.

To effect the change, the relevant data is moved to a new React Context.

This reduces the initial data load of Pontoon by about 35% on pontoon.mozilla.org. As a minor regression, there is now a short delay after clicking "REQUEST CONTEXT or REPORT ISSUE" before the contact person's data is entered into the edit field due to the mention data loading in the background.

### DIFF
--- a/translate/src/App.tsx
+++ b/translate/src/App.tsx
@@ -5,13 +5,14 @@ import './App.css';
 
 import { initLocale, Locale, updateLocale } from './context/Locale';
 import { Location } from './context/Location';
+import { MentionUsersProvider } from './context/MentionUsers';
+import { NotificationProvider } from './context/Notification';
 
 import { WaveLoader } from './core/loaders';
 import { NotificationPanel } from './core/notification/components/NotificationPanel';
 import { getProject } from './core/project/actions';
 import { getResource } from './core/resource/actions';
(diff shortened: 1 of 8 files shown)


mathjazz:
@eemeli Could you please resolve the conflict?

mathjazz:
> This reduces the initial data load of Pontoon by about 35% on pontoon.mozilla.org. 

Wow, this is such an improvement! Page load time of the translate view is noticeably improved.

> As a minor regression, there is now a short delay after clicking "REQUEST CONTEXT or REPORT ISSUE" before the contact person's data is entered into the edit field due to the mention data loading in the background.

Yes, I'll file an issue for this, but it's not a blocker.

mathjazz (approved):

mathjazz:
Filed #2591.

PR #2589, opened 2022-07-31
Drop redux-logger

eemeli:
As much of the application state is already out of Redux, the verbosity of `redux-logger` is becoming more and more annoying, and less useful. So let's remove it.


mathjazz (approved):
Good call. It's very verbose, indeed.

PR #2594, opened 2022-08-04
Add username field and use it in profile URL

### LINKED ISSUE #2483: Allow user to set a vanity URL for their profile

In order to share a profile more easily, users should be able to set up a vanity URL — a unique identifier, easier to remember, that can be used in the URL instead of the random ID assigned by Pontoon.

mathjazz:
Fix #2483.

Add an optional username field to settings, which can be used in the profile URL instead of the email address or auto-generated hash. This is an extension of #2483, in a sense that it's not just meant to customize a profile URL into something easier to remember, but a general-purpose username, which can also be used in the UI and mentions.

I'll file two follow-up issues:
* Add support for usernames in the UI (when display name not set)
* Add support for usernames in mentions


pontoon/contributors/views.py
```diff
@@ -42,7 +42,11 @@ def contributor_email(request, email):
 
 
 def contributor_username(request, username):
-    user = get_object_or_404(User, username=username)
+    try:
+        user = User.objects.get(username=username)
```
eemeli, line 46: It would be good to here check if `UserProfile.username` is set, and if so, redirect there.

eemeli (approved):

PR #2596, opened 2022-08-04
Add User Bio field

### LINKED ISSUE #2595: Add User Bio field
Reference: [spec](https://github.com/mozilla/pontoon/blob/master/specs/0111-redesign-profile-page.md)

Users should be able to provide a short bio.

mathjazz:
Fix #2595.

We also update the spec accordingly (renaming Description to Bio).


pontoon/contributors/templates/contributors/settings.html
```diff
@@ -65,6 +65,11 @@ <h3>Personal information</h3>
             {{ user_profile_form.username.errors }}
             <p class="help">{{ user_profile_form.username.help_text }}</p>
         </div>
+        <div class="field">
+            {{ user_profile_form.bio.label_tag(label_suffix='') }}
+            {{ user_profile_form.bio }}
+            {{ user_profile_form.bio.errors }}
+        </div>
```
eemeli, line 72: This seems rather terse, if/as we're displaying only the string "Bio" as a descriptor here. We should note here at least that this information will (eventually) be displayed publicly on the profile page.

eemeli (changes requested):

eemeli (approved):

PR #2598, opened 2022-08-08
Allow users to set visibility of specific profile fields

### LINKED ISSUE #2481: Allow users to set visibility of specific profile fields

Users should be able to set the visibility of specific fields in their profile to either **limited** or **public**.

**Limited** means that non authenticated users, and other contributors, won’t be able to see this information when they visit the profile page. But this information will be always visible to users with translator rights for any locale in Pontoon (translators, locale managers, administrators).

**Public** means that the field is visible to all users, including non authenticated visitors (i.e. users not logged in to Pontoon).

Blocked by:
* #2480

mathjazz:
Fixes #2481.

This patch introduces visibility settings for the profile page, currently only taken into account for the email address on the profile page. Other data for which visibility can be controlled is not shown anywhere yet.

The settings page has been slightly re-organized, and the new toggle button widget has been introduced.


pontoon/base/models.py
```diff
@@ -1586,9 +1586,47 @@ class UserProfile(models.Model):
...
+        "Email address",
+        max_length=20,
+        default=VisibilityLoggedIn.TRANSLATORS,
+        choices=VisibilityLoggedIn.choices,
+    )
+
+    visibility_external_accounts = models.CharField(
+        "External accounts",
+        max_length=20,
+        default=Visibility.TRANSLATORS,
```
mathjazz, line 1613: @flodolo Why is the default value not set to all users here? I'm afraid many people will set the external accounts and forgot to make them visible.

mathjazz:
Sorry, this patch need some updates.

mathjazz:
The tests are now passing. I've also fixed a bug, which is present in production.

The patch is now ready for a review.

pontoon/base/models.py
flodolo: I originally thought it would be better to err on the side of privacy.

1. The users who would be interested in using this data is likely admins (I want to cc someone to an issue or bug). The only exception would be Matrix, which was added later, but an all-or-nothing approach seems easier to implement.
2. Users might sign up with an "anonymous" account on Pontoon, and not tie it publicly to other accounts that might reveal their identity (to avoid issues at their day job, or because translating that language might lead to problems in real life).

I guess for 2 they might simply not add their external accounts…

So, I'd be OK with having these public by default, but maybe have some sort of indicator that these fields are public (we didn't think of that in the specs).

flodolo: Two more thoughts:
* The Bugzilla username is actually an email address. So, it might be problematic to have that as "all users". Can it be set as `VisibilityLoggedIn.LOGGED_IN`?
* `max_length=20`. Is that the max-length of each of these accounts? GitHub has 39 characters, but the email on Bugzilla can be much longer.

pontoon/contributors/static/css/settings.css
```diff
@@ -64,42 +60,98 @@
...
   background: #333941;
   border: 1px solid #4d5967;
   border-radius: 3px;
   float: none;
   padding: 4px;
   box-sizing: border-box;
+  width: 295px;
+}
+
+#main .field .toggle-button {
```
mathjazz, line 90: I wanted to move the toggle-button CSS to a separate toggle.css file, but when I added it to the `PIPELINE_CSS`, the file didn't get included in the settings.css.

Do I need to do something special to re-build the file since the recent build refactor? It's been a while since I added a new CSS file in the `/pontoon` folder. I also didn't find anything relevant in the docs.

pontoon/base/models.py
mathjazz: I guess there are arguments for both options, so let's stick to the spec. We can also make a change afterwards, although probably only for users that wouldn't have set the external accounts yet.

With the default being `TRANSLATORS`, we can also keep the Bugzilla visibility unchanged. Having a separate visibility setting for each external account would also require us to rethink the UX for visibility settings.

`max_length=20` applies to the visibility field value.

pontoon/contributors/views.py
```diff
@@ -141,7 +145,10 @@ def toggle_user_profile_attribute(request, username):
         )
 
     profile = user.profile
-    setattr(profile, attribute, json.loads(value))
+    try:
+        setattr(profile, attribute, json.loads(value))
+    except json.decoder.JSONDecodeError:
+        setattr(profile, attribute, value)
```
eemeli, line 151: What's the reason for this change? It seems a bit weird to accept data as either JSON-packaged and not; this seems like a prime way of writing bad data into the db.

pontoon/teams/static/js/multiple_team_selector.js
```diff
@@ -52,7 +52,7 @@ $(function () {
     });
   }
 
-  $('body').on('submit', '.form.user-locales-settings', function () {
+  $('body').on('submit', 'form.user-locales-settings', function () {
     updateSelectedLocales();
   });
```
eemeli, line 57:
```suggestion
  $('body').on('submit', 'form.user-locales-settings', updateSelectedLocales);
```

pontoon/contributors/static/js/settings.js
```diff
@@ -1,4 +1,42 @@
...
+
+    $.ajax({
+      url: '/api/v1/user/' + $('#server').data('username') + '/',
+      type: 'POST',
+      data: {
+        csrfmiddlewaretoken: $('body').data('csrf'),
+        attribute: attribute,
+        value: value,
+      },
+      success: function () {
```
eemeli, line 23:
```suggestion
      success() {
```

pontoon/contributors/static/js/settings.js
```diff
@@ -1,4 +1,42 @@
...
+        var label = self.parents('.field').find('label').text();
+        Pontoon.endLoader(`${label} visibility set to ${value}.`);
+      },
+      error: function (request) {
+        if (request.responseText === 'error') {
+          Pontoon.endLoader('Oops, something went wrong.', 'error');
+        } else {
+          Pontoon.endLoader(request.responseText, 'error');
+        }
+      },
```
eemeli, line 36:
```suggestion
      error({ responseText }) {
        if (responseText === 'error') {
          Pontoon.endLoader('Oops, something went wrong.', 'error');
        } else {
          Pontoon.endLoader(responseText, 'error');
        }
      },
```

eemeli (changes requested):
The "All users" bit is blocking; other comments are nitpicky.

pontoon/base/models.py
```diff
@@ -1586,9 +1586,47 @@ class UserProfile(models.Model):
...
-    quality_checks = models.BooleanField(default=True)
-    force_suggestions = models.BooleanField(default=False)
+    # External accounts
+    bugzilla = models.EmailField("Bugzilla email address", blank=True, null=True)
+    matrix = models.CharField("GitHub username", max_length=255, blank=True, null=True)
+    github = models.CharField("Matrix username", max_length=255, blank=True, null=True)
+
+    # Visibility
+    class Visibility(models.TextChoices):
+        ALL = "All users", "All users"
```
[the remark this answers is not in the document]
mathjazz: I don't have a strong opinion. @flodolo ?

pontoon/base/models.py
```diff
@@ -1611,11 +1653,6 @@ class UserProfile(models.Model):
         blank=True,
     )
 
-    # External accounts
-    bugzilla = models.EmailField("Bugzilla email address", blank=True, null=True)
-    matrix = models.CharField("GitHub username", max_length=255, blank=True, null=True)
-    github = models.CharField("Matrix username", max_length=255, blank=True, null=True)
-
```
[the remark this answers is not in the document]
mathjazz: Yeah, my bad. These are (for now 😄) final positions.

pontoon/base/models.py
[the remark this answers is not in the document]
flodolo: That's a good point, given this includes visitors (which are not users). "Public" is more accurate.

pontoon/contributors/views.py
mathjazz: `json.loads` is used to convert JS Booleans to Python Booleans. 

The newly introduced fields aren't Boolean.

I'll be more explicit about this.

pontoon/contributors/static/js/settings.js
mathjazz: We use this pattern all over the place. If we were to make a change, let's do it in a separate PR.

pontoon/contributors/static/js/settings.js
mathjazz: Same as above.

eemeli (approved):
I'm pretty sure all's good now? It looks like the PR was rebased or force-pushed so I wasn't able to review the changes made since my last review.

pontoon/contributors/static/css/settings.css
eemeli: During development, if you apply a change like this:
```diff
--- a/pontoon/settings/base.py
+++ b/pontoon/settings/base.py
@@ -359,6 +359,7 @@ PIPELINE_CSS = {
             "css/multiple_team_selector.css",
             "css/contributor.css",
             "css/team_selector.css",
+            "css/toggle-button.css",
             "css/settings.css",
         ),
         "output_filename": "css/settings.min.css",
```
you should be able to see a separate request for a `toggle-button.css`, rather than having its contents inlined into `settings.css`. It's only during the prod build that they're merged into `settings.min.css`.

mathjazz: Hmmm, not sure why CSS didn't get applied the first time I tried. Now everything is fine. 

Sorry for the false alarm!

PR #2601, opened 2022-08-11
Use POST requests for social login endpoints

mathjazz:
With the recent upgrade of `django-allauth`, when you try to log into Pontoon, you get first redirected to a log in page, on which you need to click on a button and then get redirected to the actual Firefox Accounts login page.

That's because POST requests are now required by default to initiate the handshake:
https://django-allauth.readthedocs.io/en/latest/release-notes.html#section-5

The following patch removes that extra step during the log in process by replacing sign in links with buttons, which submit a POST form.

### DIFF
--- a/pontoon/base/static/js/main.js
+++ b/pontoon/base/static/js/main.js
@@ -277,17 +277,6 @@ $(function () {
     return window.location.pathname + window.location.search;
   }
 
-  // Sign in button action
-  $('#fxa-sign-in, #standalone-signin a, #sidebar-signin').on(
-    'click',
-    function () {
-      var $this = $(this);
-      var loginUrl = $this.prop('href'),
-        startSign = loginUrl.match(/\?/) ? '&' : '?';
-      $this.prop('href', loginUrl + startSign + 'next=' + getRedirectUrl());
-    },
-  );
-
   // Sign out button action
   $('.sign-out a, #sign-out a').on('click', function (ev) {
     var $this = $(this),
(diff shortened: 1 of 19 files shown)


pontoon/base/static/js/main.js
```diff
@@ -277,17 +277,6 @@ $(function () {
...
-  // Sign in button action
-  $('#fxa-sign-in, #standalone-signin a, #sidebar-signin').on(
-    'click',
-    function () {
-      var $this = $(this);
-      var loginUrl = $this.prop('href'),
-        startSign = loginUrl.match(/\?/) ? '&' : '?';
-      $this.prop('href', loginUrl + startSign + 'next=' + getRedirectUrl());
-    },
-  );
```
mathjazz, line 289: This is ugly and is now handled in Django.

translate/src/core/editor/components/EditorMenu.tsx
```diff
@@ -55,17 +55,17 @@ function MenuContent() {
     return (
       <Localized
         id='editor-EditorMenu--sign-in-to-translate'
-        elems={{ a: <user.SignInLink url={signInURL} /> }}
+        elems={{ form: <user.SignInForm url={signInURL} /> }}
       >
-        <p className='banner'>{'<a>Sign in</a> to translate.'}</p>
+        <div className='banner'>{'<form>Sign in</form> to translate.'}</div>
```
mathjazz, line 60: `<form>` element cannot be nested within a `<p>` element.

eemeli (approved):

PR #2602, opened 2022-08-11
Turn Sign Out into a synchronous POST request

mathjazz:
#2601 introduces a regression: if you sign out and try to sign in again, CSRF verification fails.

Because the page is not reloaded and token remains the same (sign out request is async).

This patch fixes that.


eemeli (approved):
I'm not sure that I really like the `SignInOutForm` abstraction here, but I can't immediately think of a better solution.

mathjazz:
Maybe we should just give it a more generic name. :)

PR #2613, opened 2022-08-22
New Profile page layout

### LINKED ISSUE #2486: Implement new design layout (without contribution graph)

This issue tracks the implementation of the new profile layout, minus the contribution graph.

Blocked by:
* #2480 
* #2482 
* #2484

mathjazz:
Fixes #2486.

This is the new user profile page without the new contribution graph.

There are some changes compared to the mockup in the spec:

1. The Approval and Self-Approval charts have been put on the same panel, with only one visible at a time. That's because the self-approval chart will be full of zeros for most users anyways (because they don't have permission to self-approve translations). Such layout also makes it easier to handle various visibility scenarios (both charts visible, only one visible, none visible).

2. The "Administrator" label in the sidebar has been replaced with the "Project Manager" section the includes all enabled and public projects the user is a contact person for.

@flodolo I'd appreciate your review on this before we move on to the code review.

I also have a few questions:

1. Should we make the self-approval ratio chart hidden for users without translator permission?

2. Should we call them Approval **Rate** instead of Approval **Ratio** charts? I've used percentages instead of decimal numbers on the chart, because I think they are easier to reason about. 

3. The dominant (lighter) color on the Approval charts is used for the 12-month average, as opposed to the Insights charts (Time to review suggestions / Age of unreviewed), where it's used for the current month. I think we should unify that and use lighter color for the 12-month average. Or should it be the other way around?


mathjazz:
@flodolo To make testing easier, this patch is on stage:
https://mozilla-pontoon-staging.herokuapp.com/contributors/mathjazz/

flodolo:
>     2. The "Administrator" label in the sidebar has been replaced with the "Project Manager" section the includes all enabled and public projects the user is a contact person for.

That is useful, but:
* It doesn't match the terminology used in the header (CONTACT INFO). Maybe we should update that?
* Nothing is displayed for someone who is an admin ([example](https://mozilla-pontoon-staging.herokuapp.com/contributors/pmz0uSCe_Mk9Td1cksHLI1y471k/)). That's useful information, as it explains why a person can make changes without translator rights on the locale (and the PM status has no relation with that).

>     1. Should we make the self-approval ratio chart hidden for users without translator permission?

If it's not too much work, probably a good idea.

>     2. Should we call them Approval **Rate** instead of Approval **Ratio** charts? I've used percentages instead of decimal numbers on the chart, because I think they are easier to reason about.

Yes, I think _rate_ was actually the [right term](http://www.differencebetween.net/science/mathematics-statistics/difference-between-rate-and-ratio/) from the beginning.

>     3. The dominant (lighter) color on the Approval charts is used for the 12-month average, as opposed to the Insights charts (Time to review suggestions / Age of unreviewed), where it's used for the current month. I think we should unify that and use lighter color for the 12-month average. Or should it be the other way around?

I think the main point of interest is the trend over time, so using the approach you currently have here makes sense (focus on the 12-month average = dominant color).

mathjazz:
@flodolo Thanks for your feedback.

I've deployed an updated version, which should address all the items you've called out.

mathjazz:
@eemeli This one is a bit bigger than the previous ones, so I've split it into multiple commits.

Let me know if you need any help!

pontoon/base/models.py
```diff
@@ -136,13 +136,58 @@ def user_display_name_or_blank(cls, user):
...
+    """A list of locales, in which the user is assigned Manager permissions.
+
+    Only includes explicitly assigned locales for superusers.
+    """
+    locales = []
+
+    for group in self.groups.all():
+        locale = group.managed_locales.first()
+        if locale:
+            locales.append(locale)
```
eemeli, line 166: Same as previous comment.

pontoon/contributors/static/js/profile.js
```diff
@@ -0,0 +1,137 @@
+var Pontoon = (function (my) {
+  return $.extend(true, my, {
+    insights: {
+      renderCharts: function () {
```
eemeli, line 4:
```suggestion
      renderCharts() {
```

Prettier with this sugar. Same for later in the file as well.

pontoon/contributors/static/js/profile.js
```diff
@@ -0,0 +1,137 @@
...
+                  },
+                  position: 'right',
+                  ticks: {
+                    beginAtZero: true,
+                    maxTicksLimit: 3,
+                    max: 100,
+                    precision: 0,
+                    callback: function (value) {
+                      return value + '%';
+                    },
```
eemeli, line 119:
```suggestion
                    callback: (value) => nf.format(value/100),
```
See also change to `nf` below.

pontoon/contributors/static/js/profile.js
```diff
@@ -0,0 +1,137 @@
...
+              borderWidth: 1,
+              caretPadding: 5,
+              xPadding: 10,
+              yPadding: 10,
+              callbacks: {
+                label: function (items, chart) {
+                  const label = chart.datasets[items.datasetIndex].label;
+                  const value = items.yLabel;
+
+                  return label + ': ' + nf.format(value) + '%';
```
eemeli, line 83:
```suggestion
                  const value = nf.format(items.yLabel/100);

                  return `${label}: ${value}`;
```

See also change to `nf` below.

pontoon/contributors/static/js/profile.js
```diff
@@ -0,0 +1,137 @@
...
+        });
+      },
+    },
+  });
+})(Pontoon || {});
+
+/* Main code */
+const nf = new Intl.NumberFormat('en', {
+  maximumFractionDigits: 2,
+});
```
eemeli, line 134:
```suggestion
const nf = new Intl.NumberFormat('en', {
  style: 'percent',
});
```

The `percent` style also implicitly sets `maximumFractionDigits: 2`. Note that `nf.format(1) === '100%'`.

Personally, I'd also move this near the top of the file as it's a `const`, and it's a bit surprising to not see usage of it later in the file.

eemeli (changes requested):
Inline comments are mostly nitpicky/stylistic things, but where are the tests here? Some are removed, but is all the new code covered by pre-existing tests?

eemeli:
To me, the "administrator" badge looks like a button. That's probably not intentional?

Screenshot 2022-08-24 at 14 37 04

pontoon/base/models.py
```diff
@@ -136,13 +136,58 @@ def user_display_name_or_blank(cls, user):
...
+    """A list of locales, in which the user is assigned Translator permissions.
+
+    Only includes explicitly assigned locales for superusers.
+    """
+    locales = []
+
+    for group in self.groups.all():
+        locale = group.translated_locales.first()
+        if locale:
+            locales.append(locale)
```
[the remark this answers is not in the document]
mathjazz: I find it less readable that way:

```python
list(
    filter(
        lambda x: x is not None,
        map(lambda g: g.managed_locales.first(), self.groups.all()),
    )
)
```

pontoon/base/migrations/0034_rename_matrix_to_chat.py
```diff
@@ -0,0 +1,31 @@
...
+
+    dependencies = [
+        ("base", "0033_userprofile_contact_email"),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name="userprofile",
+            name="matrix",
+        ),
```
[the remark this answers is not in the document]
mathjazz: Good call.

pontoon/contributors/static/js/profile.js
mathjazz: I'd like to avoid adding more inconsistencies to our code style, so let's do this at a later point across the entire module.

mathjazz:
> To me, the "administrator" badge looks like a button. That's probably not intentional?

It's not intentional and I agree. But I also don't have great ideas on what style to use instead.

I'd like to make something that looks more like a tag (or label). I've made some changes in that direction, but it still looks "button-ish", even though less consistent with the button style used across the app.

Screenshot 2022-08-24 at 18 29 00

mathjazz:
I'd addressed your feedback with the exception of tests. Feel free to comment on these changes or wait until I also push the tests.

pontoon/contributors/utils.py
```diff
@@ -252,16 +272,16 @@ def _get_12_month_average(monthly_rates):
...
+        ),
     )
 
-    approval_rates = _get_monthly_rates(peer_approvals, peer_rejections)
-    approval_rates_12_month_avg = _get_12_month_average(approval_rates)
-    self_approval_rates = _get_monthly_rates(peer_approvals, self_approvals)
-    self_approval_rates_12_month_avg = _get_12_month_average(self_approval_rates)
+    approval_rates = get_shares_of_totals(peer_approvals, peer_rejections)
+    approval_rates_12_month_avg = get_sublist_averages(approval_rates, 12)
+    self_approval_rates = get_shares_of_totals(self_approvals, peer_approvals)
```
mathjazz, line 280: Phew, thanks to writing the unit tests, I found a bug. Self-approval rates were previously (line 260) actually calculated as peer approval rates.

mathjazz:
@eemeli All review comments have now been addressed.

eemeli (approved):
The remaining code style stuff we can figure out separately from this PR.

PR #2615, opened 2022-08-26
Implement contribution graph

### LINKED ISSUE #2487: Implement contribution graph

This issue tracks the implementation of the contribution graph.

Blocked by:
* #2485
* #2486

mathjazz:
This patch partially fixes #2487. It implements the contribution graph, with the ability to display different contribution types (translations, reviews, both, peer reviews, etc.),

~I still need to add a unit test for the utils function and optimize the request to render all contributions.~

The missing piece is the timeline that renders under the graph on page load and updates after selecting a day on the graph.


pontoon/contributors/static/js/profile.js
```diff
@@ -127,9 +136,171 @@ var Pontoon = (function (my) {
...
+          var item = monthPosition[x];
+          var monthName = monthNameFormat.format(
+            new Date(2022, item.monthIndex),
+          );
+          graphHTML += `<text x="${item.x}" y="-5" class="month">${monthName}</text>`;
+        }
+
+        // Add day labels
+        graphHTML += `
+            <text text-anchor="middle" class="wday" dx="695" dy="22">M</text>
```
flodolo, line 236: I wonder if there's a way to show a tooltip for these (e.g. Monday here).

pontoon/contributors/static/js/profile.js
```diff
@@ -127,9 +136,171 @@ var Pontoon = (function (my) {
...
+       * authored by bachvtuan.
+       *
+       * @see {@link https://github.com/bachvtuan/Github-Contribution-Graph}
+       */
+      renderContributionGraph: function () {
+        const graph = $('#contribution-graph');
+        const contributions = graph.data('contributions');
+
+        // Set start date to 365 days before now
+        var startDate = new Date();
```
eemeli, line 151: Mixing `const` and `var` like this is weird. Just make them all `const` and `let`, as appropriate.

pontoon/contributors/static/js/profile.js
```diff
@@ -127,9 +136,171 @@ var Pontoon = (function (my) {
...
+        startDate.setMonth(startDate.getMonth() - 12);
+        startDate.setUTCDate(startDate.getUTCDate() + 1);
+        startDate.setUTCHours(0, 0, 0, 0);
+        var endDate = new Date();
+
+        var graphHTML = '';
+        var step = 13;
+
+        var monthPosition = [];
+        monthPosition.push({ monthIndex: startDate.getMonth(), x: 0 });
```
eemeli, line 161:
```suggestion
        const monthPosition = [{ monthIndex: startDate.getMonth(), x: 0 }];
```

pontoon/contributors/static/js/profile.js
```diff
@@ -127,9 +136,171 @@ var Pontoon = (function (my) {
...
+
+          if (startDate.getDay() == 6) {
+            weekHTML += '</g>';
+            graphHTML += weekHTML;
+            weekHTML = null;
+            week++;
+            gx = week * step;
+          }
+
+          startDate.setUTCDate(startDate.getUTCDate() + 1);
```
eemeli, line 211: This should not be called `startDate` if/as it's being used as an iterator.

pontoon/contributors/static/js/profile.js
```diff
@@ -127,9 +136,171 @@ var Pontoon = (function (my) {
...
+              break;
+            case count < 10:
+              color = colors[1];
+              break;
+            case count < 25:
+              color = colors[2];
+              break;
+            case count < 50:
+              color = colors[3];
+          }
```
eemeli, line 197:
```suggestion
          let color;
          switch (true) {
            case count === 0:
              color = '#333941';
              break;
            case count < 10:
              color = '#41554c';
              break;
            case count < 25:
              color = '#4f7256';
              break;
            case count < 50:
              color = '#64906D';
              break;
            default:
              color = '#7bc876';
          }
```

pontoon/contributors/static/js/profile.js
```diff
@@ -127,9 +136,171 @@ var Pontoon = (function (my) {
...
+
+        // Handle tooltip
+        $(graph)
+          .find('.day')
+          .hover(
+            function (e) {
+              var count = $(e.target).attr('data-count');
+              var date = shortDateFormat.format($(e.target).attr('data-date'));
+              var action = count == 1 ? 'contribution' : 'contributions';
+              var text = `${count} ${action} on ${date}`;
```
eemeli, line 253: Alas, if only we had a localization system that could handle this sort of stuff. 😇

pontoon/contributors/utils.py
```diff
@@ -287,3 +290,68 @@ def get_approval_rates(user):
...
+    }
+
+    if contribution_type not in action_map.keys():
+        contribution_type = "all_user_contributions"
+
+    contributions = get_daily_action_counts(action_map[contribution_type])
+    total = sum(contributions.values())
+
+    return {
+        "contributions": json.dumps(contributions),
```
eemeli, line 355: Why does this need to get double-encoded like this? If there is a need for this, some code comment here explaining the situation would be a really good idea.

eemeli (changes requested):
A few small-ish changes. If using Fluent for our own localisation would be too hard, then we ought to fix that separately.

pontoon/contributors/static/js/profile.js
mathjazz: :-)

We don't use it outside the translate app, yet.

pontoon/contributors/templates/contributors/profile.html
```diff
@@ -264,6 +274,42 @@ <h3 class="controls">
...
+                        <div class="button selector">
+                            <div class="value">
+                                <span data-type="all_user_contributions" title="User translations and reviews">All user contributions</span>
+                            </div>
+                            <span class="icon fa fa-caret-down"></span>
+                        </div>
+                        <div class="menu" style="display: none;">
+                            <ul>
+                                <li class="clearfix">
+                                    <span data-type="user_translations" title="Translations submitted by the user">User translations</span>
```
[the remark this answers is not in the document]
mathjazz: I changed them, because the ones in the spec weren't clear to me. :-)

E.g. We don't use the term "Submissions", "performed" and "received" sound unusual when used together with Review and "Submissions and reviews" isn't clear about excluding reviews of user's translations.

I tried to make thing clearer in the tooltip, but I understand that's not very visible. Maybe we should show the tooltip text underneath the chart?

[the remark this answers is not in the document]
flodolo: * The concept of "peer" doesn't exist anywhere. Funny enough, it's almost never a peer who reviews your work.
* "User review" doesn't say much: are those done by the user, or received by the user.

Submissions was used to make it clear that's something entered by the user, and translation can't be used.

Since this went through review internally and externally, I'd stick to the specs. If those turn out to be unclear in everyday use, changing them should be very quick.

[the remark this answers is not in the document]
mathjazz: > Since this went through review internally and externally, I'd stick to the specs. If those turn out to be unclear in everyday use, changing them should be very quick.

Fair point. Updated the wording.

pontoon/contributors/static/js/profile.js
flodolo: It might be possible with a [title child](https://stackoverflow.com/questions/19637443/how-to-show-tooltip-for-text-in-svg)?

flodolo: Never mind, you already did.

flodolo (approved):
I'm not convinced of using uppercase in the dropdown above the graph, but the rest looks good to me.

eemeli (approved):
I've added a commit with a proposed change that would allow contributor data to not get double-encoded when passed through the `GET /update-contribution-graph` REST endpoint.

@mathjazz I hope that's ok with you? This seemed like the easiest way for me to explain what I think should be done here.

mathjazz:
> I've added a commit with a proposed change that would allow contributor data to not get double-encoded when passed through the `GET /update-contribution-graph` REST endpoint.
> 
> @mathjazz I hope that's ok with you? This seemed like the easiest way for me to explain what I think should be done here.

Ohhhh, now I see what you meant with double-encoding. Thanks for the fix!

I'm refactoring the outputs of the utils functions in the 2nd half of #2487, but I'll take that commit and update there.

mathjazz:
> I'm not convinced of using uppercase in the dropdown above the graph, but the rest looks good to me.

That consistent with the rest of the app. There are a few other things to improve in Pontoon dropdowns (e.g. some lack and arrow and hence don't look like dropdowns, some don't work well with long texts, etc.), so let's address this afterwards.

Issue #2216, opened 2020-02-27
Fluent translations stored in database without the trailing newline


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1618841](https://bugzilla.mozilla.org/show_bug.cgi?id=1618841)
Bug Reporter: @adngdb
CC:   @flodolo, @jordis, @mathjazz, @petercpg

It seems that some Fluent messages are stored in the translation table without the usual trailing newlines. That leads to comparison problems in the code, where translations are considered different because the one coming from the database lacks the newline character.

This bug covers figuring out why and where the newline character gets dropped, and fixing the issue. 

Note that the Translate app correctly sends Fluent translations with the trailing newline.

flodolo:
Coming from #2710 

Can we re-prioritize this? Before it was annoying, now it's skewing data that we need to evaluate pre-translation.

mathjazz:
Reopened #2710.

PR #2635, opened 2022-09-22
Normalise initial FTL values to always end in newline

### LINKED ISSUE #2607: If a string has multiple pending suggestions, asked to confirm existing even without making any change
Example: https://mozilla-pontoon-staging.herokuapp.com/si/firefox/all-resources/?status=unreviewed&string=199679

Just click to another string without making any change, you'll be asked if you want to save unchanged changes.

eemeli:
Fixes #2607 with a workaround

The detected "change" is effectively due to bad data; some FTL strings are stored in the database without the terminal newline that's always included in the output of our Fluent serialiser. I suspect that these may have been entered by editing and submitting strings using the advanced/raw view.

So one way to fix the problem would be to go through the data and normalise all of it to always include a newline, but I've opted here to add that normalisation in the front-end. It's likely that we'll refactor the internal representation of messages further as a part of the MF2 work, at which stage this may be dropped thanks to comparing parsed message structures rather than strings.

### DIFF
--- a/translate/src/context/Editor.tsx
+++ b/translate/src/context/Editor.tsx
@@ -224,7 +224,7 @@ export function EditorProvider({ children }: { children: React.ReactElement }) {
       } else if (!initial.endsWith('\n')) {
         // Some Fluent translations may be stored without a terminal newline.
         // If the data is cleaned up, this conditional may be removed.
-        // https://github.com/mozilla/pontoon/issues/2607
+        // https://github.com/mozilla/pontoon/issues/2216
         initial += '\n';
       }
       const next = getFtlViewAndValue(initial);
--- a/translate/src/context/Editor.tsx
+++ b/translate/src/context/Editor.tsx
@@ -211,23 +211,29 @@ export function EditorProvider({ children }: { children: React.ReactElement }) {
   }, [readonly]);
 
   useEffect(() => {
-    let format: 'ftl' | 'simple' = 'simple';
-    let initial = '';
-    let value: string | Entry = '';
-    let view: 'simple' | 'rich' | 'source' = 'simple';
+    let initial = activeTranslation?.string || '';
 
-    initial = activeTranslation?.string || '';
+    let format: 'ftl' | 'simple';
+    let value: string | Entry;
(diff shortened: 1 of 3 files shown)


translate/src/core/editor/components/Editor.test.js
```diff
@@ -26,6 +26,7 @@ const NESTED_SELECTORS_STRING = ftl`
           }
           [select] WOW
       }
+
```
mathjazz, line 29: Is this intentional? Prettier?

translate/src/context/Editor.tsx
```diff
@@ -211,23 +211,29 @@ export function EditorProvider({ children }: { children: React.ReactElement }) {
...
+    let view: 'simple' | 'rich' | 'source';
     if (entity.format === 'ftl') {
       format = 'ftl';
       if (!initial) {
         const entry = parseEntry(entity.original);
         initial = serializeEntry(getEmptyMessage(entry, locale));
+      } else if (!initial.endsWith('\n')) {
+        // Some Fluent translations may be stored without a terminal newline.
+        // If the data is cleaned up, this conditional may be removed.
+        // https://github.com/mozilla/pontoon/issues/2607
```
mathjazz, line 227: Maybe instead link to https://github.com/mozilla/pontoon/issues/2216?

mathjazz (approved):
Nice work!

Just 2 non-blocking comments.

translate/src/core/editor/components/Editor.test.js
eemeli: It's intentional, and required for the test to pass with the normalisation that's now applied.

translate/src/context/Editor.tsx
eemeli:
```suggestion
        // https://github.com/mozilla/pontoon/issues/2216
```

translate/src/core/editor/components/Editor.test.js
mathjazz: Oh, that's a message, not code. :)

PR #2636, opened 2022-09-22
Refactor accesskey extraction to include more cases

### LINKED ISSUE #2603: Default state of the accesskey message shows "0"
There is a regression in the rich Fluent editor with access key UI:

Screenshot 2022-08-12 at 22 49 33

Example:
https://pontoon.mozilla.org/sl/firefox/all-resources/?string=247270

Perhaps related: if there are multiple fields shown (more than 2), the access key UI does not appear:
https://pontoon.mozilla.org/sl/firefox/all-resources/?string=247262

@eemeli Could you please take a look?

eemeli:
Fixes #2603

The logic for detecting an accesskey is fixed and expanded to cover more cases. Previously, either the `label` attribute or the message value needed to have non-empty contents for the detection to succeed. Now:
- The input is rendered with accesskey styling independently of the number of candidate keys, but buttons next to it are rendered only if there at least two candidates.
- If a message has no `label` and no value, the `value` and `aria-label` attributes are also inspected.
- All non-letter and non-number candidates are dropped, though they may still be manually entered.
- Prefixed access keys are detected. If e.g. a message has attributes `buttonlabel` and `buttonaccesskey`, the former is used to generate the candidates for the latter.

### DIFF
--- a/translate/src/core/utils/fluent/extractAccessKeyCandidates.test.js
+++ b/translate/src/core/utils/fluent/extractAccessKeyCandidates.test.js
@@ -1,69 +1,108 @@
+import ftl from '@fluent/dedent';
 import { extractAccessKeyCandidates } from './extractAccessKeyCandidates';
 import { parseEntry } from './parser';
 
 describe('extractAccessKeyCandidates', () => {
-  it('returns null if the message has no attributes', () => {
-    const message = parseEntry('title = Title');
-    const res = extractAccessKeyCandidates(message);
-
-    expect(res).toEqual(null);
-  });
-
-  it('returns null if the message has no accesskey attribute', () => {
-    const input = 'title =' + '\n    .foo = Bar';
+  it('returns an empty list if the message has no label attribute or value', () => {
+    const input = ftl`
+      title =
+        .foo = Bar
+        .accesskey = B
+      `;
     const message = parseEntry(input);
-    const res = extractAccessKeyCandidates(message);
+    const res = extractAccessKeyCandidates(message, 'accesskey');
 
-    expect(res).toEqual(null);
-  });
-
-  it('returns null if the message has no label attribute or value', () => {
-    const input = 'title =' + '\n    .foo = Bar' + '\n    .accesskey = B';
-    const message = parseEntry(input);
-    const res = extractAccessKeyCandidates(message);
-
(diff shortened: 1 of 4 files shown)


mathjazz:
> The input is rendered with accesskey styling independently of the number of candidate keys, but buttons next to it are rendered only if there at least two candidates.

Not sure if that happens a lot, but why not show the button even for the single candidate? It would make the candidate, even if it's the only one, even more obvious. And I feel it'd be less surprising.

translate/src/core/utils/fluent/extractAccessKeyCandidates.test.js
```diff
@@ -1,69 +1,108 @@
...
-    const message = parseEntry('title = Title');
-    const res = extractAccessKeyCandidates(message);
+  it('returns an empty list if the message has no label attribute or value', () => {
+    const input = ftl`
+      title =
+        .foo = Bar
+        .accesskey = B
+      `;
+    const message = parseEntry(input);
+    const res = extractAccessKeyCandidates(message, 'accesskey');
```
mathjazz, line 13: I would actually expect B, a, r to be the candidates here. What if we just take the attribute preceding the "accesskey" attribute in cases when none of the attribute names matches the allowed list?

eemeli: Then we would at least currently end up misleading the localiser to presume that the preceding attribute is a good candidate for access keys. I looked through all uses of `.accesskey` in m-c, and the list that's proposed here is exhaustive for our current usage.

So while using an effectively random preceding attribute would technically work and be a bit more future-proof, I'm not so sure that it would be currently useful.

eemeli:
> Not sure if that happens a lot, but why not show the button even for the single candidate? It would make the candidate, even if it's the only one, even more obvious. And I feel it'd be less surprising.

@mathjazz Good point; removed the limit.

translate/src/core/utils/fluent/extractAccessKeyCandidates.test.js
mathjazz: OK. You got me at "I looked through all uses of `.accesskey` in m-c".

mathjazz (approved):

PR #2647, opened 2022-10-31
[WIP] Upgrade jquery to v3.6.1

### LINKED ISSUE #2642: Upgrade jQuery to v3.0
Pontoon is still using [jQuery v1.x](https://github.com/mozilla/pontoon/blob/master/pontoon/settings/base.py#L428). Since this is an [EOL/Obsolete Software](https://github.com/jquery/jquery.com/issues/162), is it possible to bump to 3.x?  Do you have this on your roadmap?

willianveiga:
Fixes #2642.


willianveiga:
What features should I test manually? How should I test it? I would need you help to move it forward.
Thanks.

willianveiga:
What features should I test manually? How should I test it? I would need your help to move it forward.
Thanks.

eemeli:
The part of the UI that depends on jQuery unfortunately isn't well covered by automated tests.

@mathjazz Could you comment here with a list of pages/features for @willianveiga to try out?

mathjazz:
I'd start with top-level links:
http://localhost:8000/teams/
http://localhost:8000/projects/
http://localhost:8000/contributors/
http://localhost:8000/machinery/
http://localhost:8000/admin/
http://localhost:8000/profile/
http://localhost:8000/settings/
http://localhost:8000/terms/

And then follow to all subsequent subpages, e.g.:
http://localhost:8000/{locale}/
http://localhost:8000/projects/{project}/
http://localhost:8000/{locale}/{project}/
http://localhost:8000/admin/{project}/

And click on all tabs and buttons on each of the pages.

willianveiga:
Everything seems to be working fine but your test is essential.

What about jQuery UI? Where is exactly it's used. I didn't update it, just jQuery.

Thanks for the help @eemeli and @mathjazz.

mathjazz:
Thank you very much for this important patch! I haven't found any issues myself. r+

jQuery UI is used in the [Admin](https://github.com/willianveiga/pontoon/blob/feature/upgrade-jquery-to-v3.6.1/pontoon/settings/base.py#L446) (but it doesn't look like we need it there, so I think it should be removed) and in the [settings](https://github.com/willianveiga/pontoon/blob/feature/upgrade-jquery-to-v3.6.1/pontoon/settings/base.py#L522), where we need it to drag chosen Preferred locales up and down (see the selector widget at the bottom of the page).

I'll hold on merging in case you wanted to update that, too?

mathjazz:
Thanks for updating jQuery UI!

I think we should only update the `jquery-ui.js` file with the latest version and not include the entire content of the ZIP.

mathjazz (commented):
Nice work!

Seems like `jquery-ui-1.13.2.css` is not used in any of the files, so we should remove it. That will also fix the prettier error.

willianveiga:
> Nice work!
> 
> Seems like `jquery-ui-1.13.2.css` is not used in any of the files, so we should remove it. That will also fix the prettier error.

It's referenced [here](https://github.com/mozilla/pontoon/pull/2647/files#diff-2c7387f9f77183098df1d7ff005de415b24d0517e866c4c2df0bdd557a0d9f2eR379). Should I remove it? :thinking: 

Thanks @mathjazz.

mathjazz:
Sorry - I meant the current (old) version is not referenced anywhere.

So yeah, I don't think we should reference it and keep the file around.

mathjazz:
@willianveiga I think we misunderstood each other. :)

I think the file should be removed and then we also don't need to reference it in settings, as well as in ignore files.

willianveiga:
Hey @mathjazz. Sorry about the confusion :sweat_smile: 
Hopefully I got it right now.
Thanks!

mathjazz (approved):
Great work, thanks!

40k lines of code... not bad for the second (?) patch. 😄😄😄

Issue #2052, opened 2017-01-04
Allow alternative source languages in new project configuration


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1328879](https://bugzilla.mozilla.org/show_bug.cgi?id=1328879)
Bug Reporter: @guerojeff
CC:   @flodolo, @Pike, @mathjazz, @peiying2

Many projects will be coming through in 2017 that will have en-US as the target language, not the source language. The new project configuration dialogue should allow admins to specify the source language, in a similar way that it allows definition of a project's target languages.

PR #2648, opened 2022-11-02
Add support for Google AutoML Translation

mathjazz:
This patch brings Google AutoML Translation support to Pontoon in addition to the existing Google Translation API support. [Read more about the differences between the two services.](https://cloud.google.com/translate/automl/docs/beginners-guide#is-the-translation-api-or-automl-the-right-tool-for-me)

AutoML Translation will be used instead of Translation API for locales for which we have trained models available. For that purpose, a new `Locale.google_automl_model` field has been added to the DB. If the value is set, Pontoon will use AutoML Translation with the given model ID, if not, it will use the Translation API with the generic model.

For now, we only have trained models available for Italian (it) and Slovenian (sl) locale, but we can easily add more. We chose these two locales, because we plan to use them to evaluate the Pretranslation service. In addition to that, AutoML Translation will be used in Machinery as well.

--

A couple of tweaks were required to make AutoML Translation work properly using the Google's client library:

1. An environment variable (`PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION`) is used to configure one of the libraries required by `google-cloud-translate`. See [039d903](https://github.com/mozilla/pontoon/pull/2648/commits/039d90335bb34e191b55e77d734b8623c1324c4c).

2. Authentication using [service accounts](https://cloud.google.com/translate/docs/setup?hl=en_US#auth) required adding a new buildpack to Heroku in order to bypass the limitations of the ephemeral storage. See [24a18c0](https://github.com/mozilla/pontoon/pull/2648/commits/24a18c0268a1d9d80968734f7465c8002af5dd19).


pontoon/machinery/utils.py
```diff
@@ -61,6 +79,45 @@ def get_google_translate_data(text, locale_code):
...
+    try:
+        client = translate.TranslationServiceClient()
+    except DefaultCredentialsError as e:
+        log.error("Google AutoML Translation error: {e}")
+        return {
+            "status": False,
+            "message": f"{e}",
+        }
+
+    project_id = "85591518533"
```
eemeli, line 92: This magic constant should probably be a well commented config value, yes?

pontoon/machinery/utils.py
```diff
@@ -61,6 +79,45 @@ def get_google_translate_data(text, locale_code):
...
+    model_id = locale.google_automl_model
+    location = "us-central1"
```
eemeli, line 94: This also looks like a magic constant. At the very least it should have a code comment explaining itself.

pontoon/machinery/utils.py
```diff
@@ -61,6 +79,45 @@ def get_google_translate_data(text, locale_code):
...
+    parent = f"projects/{project_id}/locations/{location}"
+    model_path = f"{parent}/models/{model_id}"
+
+    response = client.translate_text(
+        request={
+            "contents": [text],
+            "target_language_code": locale.google_translate_code,
+            "model": model_path,
+            "source_language_code": "en",
```
eemeli, line 103: Just to check, is this always the case in Pontoon?

eemeli (changes requested):
In general looks good; see inline comments regarding magic reduction.

pontoon/machinery/utils.py
mathjazz: Yeah, and the same value is [hardcoded](https://github.com/mozilla/pontoon/blob/24a18c0268a1d9d80968734f7465c8002af5dd19/pontoon/machinery/utils.py#L51) in the generic API.

We track support for other source locales in https://github.com/mozilla/pontoon/issues/2052.

pontoon/machinery/utils.py
mathjazz: I'll add a comment here - AutoML requires `us-central1`.

Although the docs are not [explicit](https://cloud.google.com/translate/automl/docs/before-you-begin#create_a_cloud_storage_bucket) about it, I tried other locations and got an error stating that.

pontoon/machinery/utils.py
mathjazz: Of course, nice catch!

eemeli (approved):

PR #2652, opened 2022-11-07
Include pretranslated in completion calculations

mathjazz:
As of this patch, completion percentage calculations used in dashboards, charts and menus treat pretranslated strings as completed, since they end up being used in product.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -583,7 +583,11 @@ class AggregatedStats(models.Model):
 
     @property
     def completed_strings(self):
-        return self.approved_strings + self.strings_with_warnings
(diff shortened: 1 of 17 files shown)


eemeli (approved):
The number of places where this sum is calculated makes me think it should be centralised somewhere, as it's hard to tell if e.g. this PR's change covers all of them.

mathjazz:
> The number of places where this sum is calculated makes me think it should be centralised somewhere, as it's hard to tell if e.g. this PR's change covers all of them.

Agreed, that was my thought, too. Filed https://github.com/mozilla/pontoon/issues/2654.

PR #2656, opened 2022-11-08
Add script to generate source-only repos for Thunderbird and Seamonkey

mathjazz:
This script uses https://hg.mozilla.org/projects/comm-l10n/ repository to create repositories with source strings only for Thunderbird and Seamonkey.

It's a standalone script, rather than being included in [`mozilla-en-US.py`](https://github.com/mozilla/pontoon/blob/60e49bf07d16540a18501c934b310b30f46c067b/bin/mozilla-en-US.py), because the latter will be obsolete and removed soon; mobile has been long gone, Firefox can just use https://hg.mozilla.org/l10n/gecko-strings/, and this script will cover Thunderbird (including Lightning) and Seamonkey.

This is the output of the script:
https://hg.mozilla.org/users/m_owca.info/thunderbird
https://hg.mozilla.org/users/m_owca.info/seamonkey/

### DIFF
--- a/bin/make_comm_source_repos.py
+++ b/bin/make_comm_source_repos.py
@@ -8,122 +8,137 @@ import datetime
 import os
 import shutil
 import subprocess
+import sys
 
 TARGET_REPOS = {
-    'thunderbird': [
-        'calendar',
-        'chat',
-        'mail',
+    "thunderbird": [
+        "calendar",
+        "chat",
+        "mail",
     ],
-    'seamonkey': [
-        'suite',
+    "seamonkey": [
+        "suite",
     ],
 }
 
 
-def write(text):
-    timestamp = datetime.datetime.now().strftime('[%Y-%m-%d %H:%M:%S] ')
-    print(timestamp + text)
+def logmsg(text):
+    timestamp = datetime.datetime.now().strftime("[%Y-%m-%d %H:%M:%S]")
+    print(f"{timestamp} {text}")
 
 
 def execute(command, cwd=None):
     try:
         st = subprocess.PIPE
-        proc = subprocess.Popen(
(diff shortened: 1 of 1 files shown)


mathjazz:
@flodolo I figured you like scripting, so...

bin/make_comm_source_repos.py
```diff
@@ -0,0 +1,129 @@
...
+    url = 'ssh://hg.mozilla.org/users/m_owca.info/' + repo
+    target = os.path.join('target', repo)
+
+    # Clone or update target repository
+    code = pull(url, target)
+
+    quit_or_pass(code)
+
+    # Prune all subdirectories in target repository in case they get removed from source
+    for subdir in os.listdir(target):
```
flodolo, line 116: This fails is there is a file in the root…

flodolo (changes requested):
I confess I'm not a fan of this script, mostly because it looks like a very old script with some logic jumps.

Here's an alternative version with some changes in logic, although the general flow remains the same.
https://gist.github.com/flodolo/52cdf68eb5019a75460df8172efdf957

I'm not super happy about checking the output of `hg status` to see if there's something to commit, but it doesn't look like hg provides any better way. In case, we could commit and push all the time like the previous script was doing 🤷🏼 

If you feel strongly about keeping something similar to what we already have, at least we should pass it through black.

mathjazz:
Admittedly, I wanted to stay away from the refactor here, because this is not a long-term solution and refactors always increase risk. Yes, the code is old and ugly, but OTOH didn't require a meaningful update in over 4 years.

However, the script you proposed is much more modern, so I'm happy to take it.
Especially considering you'll now own it. 😛

flodolo (approved):
I knew it…

star-wars-its-a-trap

PR #2669, opened 2022-11-30
Set up custom Clock Process (see #2670)

mathjazz:
We currently use Heroku Scheduler add-on for job scheduling. It's easy to use, but it doesn't allow running jobs more frequently than every 10 minutes, which is needed to keep Google AutoML Translation Models warm.

Hence this PR creates a [custom clock process](https://devcenter.heroku.com/articles/scheduled-jobs-custom-clock-processes#custom-clock-processes). It replaces the management command introduced in #2659 with a Celery task (than runs on the worker).

The patch is currently deployed to stage to monitor if the current frequency of 5 minutes is good enough.


eemeli (changes requested):
A code comment in `pontoon/base/clock.py` would be good for explaining why/what that script is doing. The file is named really generically but only includes the AutoML warmup task, so explaining that discrepancy would be nice.

eemeli:
@mathjazz Something weird has happened with your clone, with a different history starting from a 2017 commit e4749e05 onwards. Please fix?

mathjazz:
wooow :)

I had to rewrite history and seems like this was an unintended side-effect.

I'd rather stay safe, so I opened a new PR - #2670.

PR #2673, opened 2022-12-07
Update to jest 29 & refresh npm lockfile

eemeli:
Ran `npm update` and installed latest versions of `jest` and related packages. A few test config changes were required, and `jest-environment-jsdom` is now needed as an explicit dependency.

This makes Node.js v14 the minimum required for the build & tests, but doesn't change any of the runtime requirements.


mathjazz (approved):
Tested locally a bit and verified nothing breaks on stage. LGTM!

PR #2675, opened 2022-12-07
Add warning for empty Fluent entries

### LINKED ISSUE #2599: No warnings when trying to submit empty translations
I've noticed an increase in the number of empty strings in Firefox, where I have [special checks](https://test.flod.org/checks/).

Apparently, we don't warn anymore when someone tries to submit an empty translation.

eemeli:
**Edit:** Implementation completely changed from the original description; see comments below.

Fixes #2599 ~~using the solution discussed in mozilla/compare-locales#6.~~
CC @flodolo 

~~Adds a Fluent utility function `isEmptyMessage()` and uses it to disable the save action when the simple or rich editor inputs are empty.~~

~~The basic appearance of the button is kept unchanged, but the hover state is affected; the cursor for the disabled state uses a 🚫 indicator and the tooltip notes that saving an empty string is only possible using the advanced editor.~~

~~The label of the button makes more sense to be controlled with CSS `text-transform` rather than being hard-coded, and it doesn't really make sense to duplicate the formatted strings that are provided by FTL.~~

### DIFF
--- a/pontoon/checks/libraries/pontoon_db.py
+++ b/pontoon/checks/libraries/pontoon_db.py
@@ -5,6 +5,7 @@ import bleach
 
 from collections import defaultdict
 from fluent.syntax import FluentParser, ast
+from fluent.syntax.visitor import Visitor
 
 from pontoon.sync.formats.ftl import localizable_entries
 
@@ -23,6 +24,24 @@ def get_max_length(comment):
         return int(max_length[0][1])
 
     return None
+
+
+class IsEmptyVisitor(Visitor):
+    def __init__(self):
+        self.is_empty = True
+
+    def visit_Placeable(self, node):
+        if isinstance(node.expression, ast.Literal):
+            if node.expression.parse()["value"]:
+                self.is_empty = False
+        elif isinstance(node.expression, ast.SelectExpression):
+            self.generic_visit(node.expression)
+        else:
+            self.is_empty = False
+
+    def visit_TextElement(self, node):
+        if node.value:
+            self.is_empty = False
 
 
 def run_checks(entity, original, string):
@@ -79,5 +98,13 @@ def run_checks(entity, original, string):
         # Message ID mismatch
         elif entity_ast.id.name != translation_ast.id.name:
             checks["pErrors"].append("Translation key needs to match source string key")
+
+        # Empty translation entry warning; set here rather than pontoon_non_db.py
+        # to avoid needing to parse the Fluent message twice.
+        else:
+            visitor = IsEmptyVisitor()
+            visitor.visit(translation_ast)
+            if visitor.is_empty:
+                checks["pndbWarnings"].append("Empty translation")
 
     return checks
--- a/pontoon/checks/tests/test_pontoon_db.py
+++ b/pontoon/checks/tests/test_pontoon_db.py
@@ -132,6 +132,25 @@ def test_empty_translations(get_entity_mock):
         "pErrors": ["Empty translations are not allowed"]
     }
 
+    assert run_checks(
+        get_entity_mock("ftl", string="key = value"), "", 'key = { "" }'
+    ) == {"pndbWarnings": ["Empty translation"]}
+
+    assert (
+        run_checks(
+            get_entity_mock("ftl", string="key =\n  .attr = value"),
+            "",
+            """key =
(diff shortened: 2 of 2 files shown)


translate/public/locale/en-US/translate.ftl
```diff
@@ -113,22 +113,24 @@ comments-CommentsList--all-comments = ALL COMMENTS
 
 editor-EditorMenu--sign-in-to-translate = <form>Sign in</form> to translate.
 editor-EditorMenu--read-only-localization = This is a read-only localization.
-editor-EditorMenu--button-copy = COPY
+editor-EditorMenu--button-copy = Copy
```
mathjazz, line 116: We use uppercase intentionally here. We should add a comment explaining why.

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1666277
https://mozilla-l10n.github.io/documentation/localization/dev_best_practices.html#css-issues

mathjazz:
> The basic appearance of the button is kept unchanged, but the hover state is affected; the cursor for the disabled state uses a 🚫 indicator and the tooltip notes that saving an empty string is only possible using the advanced editor.

I haven't tested this yet, but it sounds like the behavior when submitting an empty string in FTL format is different compared to submitting an empty string in other formats. Communicating this information only via tooltip is also not ideal in the light of our upcoming effort to support mobile devices.

translate/public/locale/en-US/translate.ftl
eemeli: I've reverted my changes on this; sounds like adding that note would be a separate change?

eemeli:
> I haven't tested this yet, but it sounds like the behavior when submitting an empty string in FTL format is different compared to submitting an empty string in other formats. Communicating this information only via tooltip is also not ideal in the light of our upcoming effort to support mobile devices.

@mathjazz The button is now also dimmed when disabled, and pressing Enter while editing applies the same checks as the button does. I also got rid of the l10n-id & classname concatenation & ensured that the disabled button label always matches the active one.

mathjazz:
> The button is now also dimmed when disabled

That's solves the problem with mobile devices, but my main concern was the first part of the paragraph. Why do we need to invent this behavior in first place, given we already have an established UX practice of preventing empty string submissions?

Why is the UX completely different here compared to when submitting an empty string to a PO or DTD file? I would expect the UX is the same, perhaps with a different error message (the one that is used in a tooltip).

mathjazz:
@flodolo We just had a Pontoon call and wanted to ask you for an opinion here as a continuation of the discussion in https://github.com/mozilla/compare-locales/pull/6.

We talked about 3 UX options:

1. Make the Submit button in the Rich editor disabled unless there's content in the text field - this PR ATM.
2. Just show a UI warning - the approach you recommended in your [original comment](https://github.com/mozilla/compare-locales/pull/6#issuecomment-1253924214) in the c-l PR.
3. Only allow empty strings to be submitted via the UI - agreed upon in the c-l PR.

Which is your preference?

flodolo:
> * Make the Submit button in the Rich editor disabled unless there's content in the text field - this PR ATM.

So, it's not possible to enter an empty translation at all? I think we don't want that, we still want to be able to consider the string translated (and empty) in some edge cases, and we want to make it difficult to submit empty translation by mistake (which is what is happening right now).

> * Just show a UI warning - the approach you recommended in your [original comment](https://github.com/mozilla/compare-locales/pull/6#issuecomment-1253924214) in the c-l PR.

I suggested that because I assumed it would be easy (that's what happens for DTDs, for example). 

> * Only allow empty strings to be submitted via the UI - agreed upon in the c-l PR.

via the UI -> via the Advanced Fluent UI?

Personally, 2 is preferable to 3, while 1 seems a no-go.

mathjazz:
> via the Advanced Fluent UI?

Yes!

flodolo:
> @flodolo Does rephrasing the 3rd option change your preference?

No, for me it's still 2 > 3, if nothing else for consistency.

eemeli:
I've replaced the earlier PR with a completely different solution that follows Option 2 as described above, and generates a "non-db" warning from the Pontoon checks when attempting to submit an empty Fluent entry.

mathjazz (approved):
Nice work!

PR #2682, opened 2022-12-13
Set up a dedicated dyno to schedule and execute Google AutoML Translate Warmup process

mathjazz:
Usually, [separating](https://devcenter.heroku.com/articles/scheduled-jobs-custom-clock-processes#overview) scheduler and execution worker is recommended, but we break that pattern intentionally in order to run the task exactly when scheduled.

I wasn't able to figure out how to use celery to run periodic tasks locally, so I added a new dependency - `APScheduler`.


mathjazz:
I wasn't able to figure out how to use celery to run periodic tasks locally, so I added a new dependency - `APScheduler`.

pontoon/machinery/automl_warmup.py
```diff
@@ -0,0 +1,47 @@
...
+
+import django
+import dotenv
+import logging
+import os
+
+from apscheduler.schedulers.blocking import BlockingScheduler
+
+
+# Read dotenv file and inject it's values into the environment
```
flodolo, line 15:
```suggestion
# Read dotenv file and inject its values into the environment
```

pontoon/machinery/automl_warmup.py
```diff
@@ -0,0 +1,47 @@
+"""
+Google Cloud AutoML Translation has latency of ~15s, caused by the loading time of a
+custom model into the chip. To keep latency low, we need to make regular dummy warm-up
```
flodolo, line 3:
```suggestion
custom model into the chip. To keep latency low, we need to make regular dummy warmup
```

That's how you spell it in the rest of the files. Personally, I prefer the hyphenated version (`warm-up`).

pontoon/machinery/automl_warmup.py
```diff
@@ -0,0 +1,47 @@
...
+
+@sched.scheduled_job("interval", seconds=settings.GOOGLE_AUTOML_WARMUP_INTERVAL)
+def warm_up_automl_models():
+    log.info("Google AutoML Warmup process started.")
+
+    locales = Locale.objects.exclude(google_automl_model="").order_by("code")
+
+    for locale in locales:
+        get_google_automl_translation("t", locale)
+        log.info(f"Google AutoML Warmup for {locale.code} complete.")
```
flodolo, line 42: We don't really know if it's been completed though, do we? It could error in the function called.

I wonder if we need this level of details, and this could be just a message with the list of locales (`f"Starting Google AutoML Warmup for: {', '.join(locales)}"), and drop the log at line 36.

mathjazz: We don't know if the warmup was successful, but when the log message is emitted, it's complete. It could also complete by raising an exception, which we don't catch, but can detect it implicitly through a missing log message.

We hit quite a few issues with the warmup so far, so I prefer to keep logging verbose, including timing for each locale. At least until we keep adding locales.

eemeli (approved):

PR #2683, opened 2022-12-14
Cleanup placeables markup, "Translation" components & originalstring module

eemeli:
This PR is in preparation for further data model changes, and its changes should not be externally visible while making later development easier. The three commits here could be considered as separate changes:

### Clarify names of "Translation" components

Previously, we had three different components called `Translation` and one which wasn't, but logically should be. This makes it sometimes rather difficult to track where an issue really is.

### Use updated react-content-marker to refactor placeable markup

This depends on mozilla/react-content-marker#20, which allows moving the parser rules and their tests there. This makes it a bit easier to also effect a merge of `<WithPlaceables>`, `<WithPlaceablesForFluent>`, `<WithPlaceablesNoLeadingSpaces>`, `<WithPlaceablesForFluentNoLeadingSpaces>`, and `getMarker()` all into one `<Marked>` component.

### Refactor OriginalString into its own module, including TermsPopup

The collection of functions and components that make up `<OriginalString>` was previously spread over `modules/entitydetails/` and `modules/fluentoriginal/`. This collects all of that now under `modules/originalstring/` while simplifying in particular `<Metadata>`.


translate/src/modules/entitydetails/components/Metadata.tsx
```diff
@@ -258,19 +210,7 @@ export function Metadata({
...
-      />
-      {popupTerms.length > 0 && (
-        <TermsPopup
-          isReadOnlyEditor={isReadOnlyEditor}
-          terms={popupTerms}
-          hide={hidePopupTerms}
-          navigateToPath={navigateToPath}
-        />
-      )}
+      <OriginalString navigateToPath={navigateToPath} terms={terms} />
```
mathjazz, line 213: `OriginalString` (as well as `ContextIssueButton` above) are not really string Metadata. We should rename this component to something that also covers these two components. That said, I don't have a great idea what would that name be.

translate/src/core/placeable/components/Marked.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+import './Marked.css';
+
+let keyCounter = 0;
+const MarkerCache = new Map<string, ReturnType<typeof createMarker>>();
+
+/**
+ * Component that marks placeables in a string.
+ */
+export function Marked({
```
mathjazz, line 15: `Highlight(ed)` might be a better component name than `Marked`?

eemeli:
Now that mozilla/react-content-marker#20 is available in `react-content-marker@2.2.0`, I've switched the dependency to that rather than the prerelease version. Also renamed the component as `Highlight`, as that is indeed a better name.

The `Metadata` rename should be a separate change from this PR.

mathjazz (approved):
Nice work.

I tested on stage and didn't spot any issues (for some reason couldn't get the build to run locally).

translate/src/core/placeable/components/Highlight.tsx
```diff
@@ -0,0 +1,71 @@
...
+import createMarker, { getRules, Parser } from 'react-content-marker';
+import { TermState } from '~/core/term';
+
+import './Highlight.css';
+
+let keyCounter = 0;
+const MarkerCache = new Map<string, ReturnType<typeof createMarker>>();
+
+/**
+ * Component that marks placeables in a string.
```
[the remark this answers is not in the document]
eemeli:
```suggestion
 * Component that marks placeables and terms in a string.
```

`react-content-marker` doesn't handle wrapping itself really well, so using two separate components would require some further changes there, as well as some way to avoid terms being mistakenly identified within placeables. In other words, it gets Complicated.

eemeli:
> I tested on stage and didn't spot any issues (for some reason couldn't get the build to run locally).

Did you run `npm install` locally, to account for the updated dependency?

mathjazz:
> Did you run `npm install` locally, to account for the updated dependency?

Ah, that was it!

It'd be great if that was part of `make build-frontend` or better documented. I believe that step wasn't necessary in the past.

PR #2684, opened 2022-12-14
Replace Fluent AST with MF2 data model

eemeli:
~~This adds on top of #2683; the actual diff compared to that is [this](https://github.com/eemeli/pontoon/compare/tx-components...mf2-entry?w=1).~~ **Edit:** Parent PR has been merged.

Now that we've somewhat reduced the surface of `utils/message/`, let's refactor it to represent Fluent messages using the MF2 data model rather than the Fluent one. All of the relevant implementations, tests, and call sites are modified as necessary.

It's probably easiest to start looking at this from `parseEntry()` and `serializeEntry()`, which (as before) transform entries between Fluent string representation and the data model. At least for now, these transforms go through the Fluent AST. Next, see the rest of the changes in `utils/message/` for working with a different data model. A small utility library `utils/pojo.ts` is added to replace the Fluent `BaseNode.clone()` and `BaseNode.equals()` methods.

Some of the utility functions should eventually get upstreamed to the `messageformat` & `@messageformat/fluent` packages, but for now they make senes to develop and test here.

Of the usage sites, `<RichString>` needed the greatest amount of structural change, as its `render*()` functions are now more properly components.

The only user-visible change from all this should be that nested selectors are now fully supported in the rich editor. For example, with a source like:
```fluent
-brand-shorter-name =
    { $case ->
        [genitive]
            { $count ->
                [one] Firefoxin
               *[other] Firefoxien
            }
       *[nominative] Firefox
    }
    .case-status = with-cases
```
the rich editor looks like this:

Screenshot 2022-12-14 at 11 07 01

The selectors `$case` and `$count` are lifted up to the top level, and entries found for each variant. When re-stringiying this, the result is:
```fluent
-brand-shorter-name =
    { $case ->
        [genitive]
            { $count ->
                [one] Firefoxin
               *[other] Firefoxien
            }
       *[nominative]
            { $count ->
                [one] Firefox
               *[other] Firefox
            }
    }
    .case-status = with-cases
```

Note how the `nominative` case now has a selector value; matches like this aren't automatically flattened, though that could be done.

One other potentially observable change is that for Fluent, plural categories are identified via `Intl.PluralRules` rather than the manually input locale data. This matches the runtime behaviour of Fluent.


eemeli:
Rebased on latest `master`.

mathjazz:
Nice work!

I found issues with the following string:
https://github.com/mozilla-l10n/pontoon-ftl/blob/master/en-US/demo.ftl#L50-L55

1. The "other" variant renders in the rich editor as if it were a plural variant: "other⁩ (e.g. ⁨0⁩)".
2. The order of the variants is different than in the source file ("other" is rendered first).

mathjazz:
I should have mentioned that I've used the translation from the VCS, which has the first variant set as default:
https://github.com/mozilla-l10n/pontoon-ftl/blob/master/sl/demo.ftl#L31-L35

If you open the editor without any translations and submit one, everything works as expected.

eemeli:
Ah, thanks; that explains things. Should be patched now.

mathjazz:
Thanks!

One more thing: variants order in rich editor for this translation is not the same as in master or in the file:
https://github.com/mozilla-l10n/pontoon-ftl/blob/master/sl/demo.ftl#L56-L61

eemeli:
That kinda counts as a feature, actually. As in, the MF2 selection is order-dependent, so a default choice is always listed last because any earlier it'd get picked before an explicit match. Therefore, the variant order when coming from e.g. Fluent may get rearranged.

While this is a difference to the previous, the original source goes through the same parsing in OriginalString as well as Editor, so the UI is consistent.

There are some, but not many, selectors that use something other than the last variant as the default. Is this an actual problem, i.e. do we have an actual need to maintain the variant order that the original source has?

mathjazz:
Thanks for explaining. This is not a problem per se. I called it out, because it's different from the current behavior and I thought it wasn't intentional.

translate/src/modules/entitydetails/components/FluentAttribute.test.js
```diff
@@ -0,0 +1,30 @@
+import { shallow } from 'enzyme';
+import React from 'react';
+import { FluentAttribute } from './FluentAttribute';
+
+describe('isSimpleSingleAttributeMessage', () => {
+  it('renders nonempty for a string with a single attribute', () => {
+    const original = 'my-entry =\n    .an-atribute = Hello!';
```
mathjazz, line 7: It would be nice to use `@fluent/dedent` here.

mathjazz (approved):
This was pretty tough to review, sorry it took so long.

I tested quite extensively, though.

AFAICT, this is good to go. Just one minor suggestion for one of the test files.

PR #2686, opened 2022-12-14
In the simple view, use an empty string for empty Fluent literals

### LINKED ISSUE #2685: Empty string suggested in the input area for a term with 100% matches
https://pontoon.mozilla.org/sl/firefox-monitor-website/app.ftl/?status=missing&string=266485

Firefox Monitor has 100% matches in MACHINERY (both Google Translate and TM), but when you click the link, the input area has an empty string (`{ "" }`). 

I've only seen this happening with Fluent. Hard to tell when it started, since we don't normally have a lot of new terms, but I think the bug was already there a couple of weeks ago.

eemeli:
Fixes #2685 

We need to use empty literals `{ "" }` on occasion to represent empty messages, but these should not be rendered in the simple view. This change should also allow for automatic 100% TM matches to work for missing terms, as the field value will then be empty by the time we get the TM results.

### DIFF
--- a/translate/src/utils/message/getSimplePreview.test.js
+++ b/translate/src/utils/message/getSimplePreview.test.js
@@ -1,4 +1,6 @@
 import { getSimplePreview } from './getSimplePreview';
+import { parseEntry } from './parser';
+import { serializeEntry } from './serializer';
 
 describe('getSimplePreview', () => {
   it('works for an empty string', () => {
@@ -36,8 +38,9 @@ describe('getSimplePreview', () => {
   });
 
   it('returns an empty string for an empty literal value', () => {
-    const message = 'empty = { "" }';
-    const res = getSimplePreview(message);
+    const entry = parseEntry('empty = { "" }\n')
+    serializeEntry(entry)
+    const res = getSimplePreview(entry);
     expect(res).toEqual('');
   });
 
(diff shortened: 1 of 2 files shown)


translate/src/utils/message/getSimplePreview.test.js
```diff
@@ -35,6 +35,12 @@ describe('getSimplePreview', () => {
     expect(res).toEqual('Avengers');
   });
 
+  it('returns an empty string for an empty literal value', () => {
+    const message = 'empty = { "" }';
+    const res = getSimplePreview(message);
+    expect(res).toEqual('');
+  });
+
```
mathjazz, line 43: Seems like this test cases also passes on master?

mathjazz:
Thanks for the update. Could you please fix prettier?

translate/src/utils/message/getSimplePreview.test.js
eemeli: Yeah. Fixed now to break there, once I figured out how the `StringLiteral` was sneaking in. Not entirely sure exactly how that's happening, but I'd rather not spend time fixing it, as this entire whole mode should disappear with the data model changes.

mathjazz (approved):

Issue #2691, opened 2022-12-23
Don't submit pre-translated strings with warnings or errors


flodolo:
If possible, it would be great to have 2 different behaviors:
* Strings with warning should be suggested, instead of submitted, because they might be OK (i.e. the warning could be ignored), or at least it might be still useful to post-edit them instead of starting from scratch.
* String with errors shouldn't be even suggested.

If not possible, I guess storing them both as suggestions would work, since the localizer wouldn't be able to confirm a string with errors (but it will be confusing).

The important piece is not storing them in VCS, because it will create a chain of errors, especially in parts that we don't control (the final product).

flodolo:
I have another case (caused by #2690)

<img width="1043" alt="Schermata 2023-01-19 alle 21 16 19" src="https://user-images.githubusercontent.com/3644868/213550951-61270929-f2f9-4181-a12a-de0765993828.png">

flodolo:
I think we need to bump priority.

14 errors generated for zh-TW.
https://pontoon.mozilla.org/zh-TW/mozillaorg/all-resources/?time=202304021244-202304021344&string=215305

What's worse, is that I've rejected them in Pontoon, but it didn't remove them from the repository (filing a separate issue to investigate that).

flodolo:
Android projects are another source of errors, because of unescaped quotes (we have 2 for Welsh)

PR #2716, opened 2023-01-19
Only machine translate the part that follows Id in fluent string

### LINKED ISSUE #2690: Pre-translated Fluent string has the ID translated or modified
It happened for at least two strings.

```
remember-pw-link = Remember your password? Sign in
```

Became

```
Remember-pw-link = Ricordi la password? Accedi
```

No clue why it changed to uppercase.

On the other hand, for 

```
plan-price-interval-year =
    { $intervalCount ->
        [one] { $amount } all’anno
       *[other] { $amount } ogni { $intervalCount } anni
    }
    .title =
        { $intervalCount ->
            [one] { $amount } all’anno
           *[other] { $amount } ogni { $intervalCount } anni
        }
```

The id was translated to `piano-prezzo-intervallo-anno`(but the translation was good besides that).

vishalol:
Fixes #2690

I think the same approach can be applied to prevent attribute translation as well.

### DIFF
--- a/pontoon/pretranslation/pretranslate.py
+++ b/pontoon/pretranslation/pretranslate.py
@@ -6,7 +6,7 @@ from functools import reduce
 from django.db.models import CharField, Value as V
 from django.db.models.functions import Concat
 
-from pontoon.base.models import User, TranslatedResource, Resource
+from pontoon.base.models import User, TranslatedResource
 from pontoon.machinery.utils import (
     get_google_translate_data,
     get_translation_memory_data,
@@ -37,9 +37,6 @@ def get_translations(entity, locale):
     """
     tm_user = User.objects.get(email="pontoon-tm@example.com")
     gt_user = User.objects.get(email="pontoon-gt@example.com")
-
-    if entity.resource.format != Resource.Format.FTL:
-        return
 
     strings = []
     plural_forms = range(0, locale.nplurals or 1)
@@ -74,19 +71,19 @@ def get_translations(entity, locale):
 
     # Else fetch from google translate
     elif locale.google_translate_code:
-        entity_string = (
+        key_substituted_string = (
             entity.string.replace(entity.key, UNTRANSLATABLE_KEY, 1)
-            if entity.resource.format == Resource.Format.FTL
+            if entity.resource.format == "ftl"
             else entity.string
         )
 
         gt_response = get_google_translate_data(
-            text=entity_string,
+            text=key_substituted_string,
             locale=locale,
         )
 
         if gt_response["status"]:
-            if entity.string != entity_string:
+            if entity.string != key_substituted_string:
                 gt_response["translation"] = gt_response["translation"].replace(
                     UNTRANSLATABLE_KEY, entity.key
                 )
--- a/pontoon/pretranslation/pretranslate.py
+++ b/pontoon/pretranslation/pretranslate.py
@@ -6,7 +6,7 @@ from functools import reduce
 from django.db.models import CharField, Value as V
 from django.db.models.functions import Concat
 
-from pontoon.base.models import User, TranslatedResource
+from pontoon.base.models import User, TranslatedResource, Resource
 from pontoon.machinery.utils import (
     get_google_translate_data,
     get_translation_memory_data,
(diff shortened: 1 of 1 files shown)


pontoon/pretranslation/pretranslate.py
```diff
@@ -37,6 +38,9 @@ def get_translations(entity, locale):
     tm_user = User.objects.get(email="pontoon-tm@example.com")
     gt_user = User.objects.get(email="pontoon-gt@example.com")
 
+    if entity.resource.format != Resource.Format.FTL:
+        return
+
```
mathjazz, line 43: That needs to be removed probably?

pontoon/pretranslation/pretranslate.py
```diff
@@ -18,6 +18,7 @@
     get_reconstructed_message,
 )
 
+UNTRANSLATABLE_KEY = "AIzaSyDX3R5Y1kxh_8lJ4OAO"
```
mathjazz, line 21: Why don't we simply cut the `key` off and add it back when build a pretranslated message?

pontoon/pretranslation/pretranslate.py
```diff
@@ -70,12 +74,23 @@ def get_translations(entity, locale):
 
     # Else fetch from google translate
     elif locale.google_translate_code:
+        entity_string = (
+            entity.string.replace(entity.key, UNTRANSLATABLE_KEY, 1)
+            if entity.resource.format == Resource.Format.FTL
```
mathjazz, line 79: We mostly just use `ftl` instead of `Resource.Format.FTL` across the codebase.

pontoon/pretranslation/pretranslate.py
```diff
@@ -70,12 +74,23 @@ def get_translations(entity, locale):
 
     # Else fetch from google translate
     elif locale.google_translate_code:
+        entity_string = (
```
mathjazz, line 77: This `entity_string` is different from the one above, which is confusing. They should probably have different names.

pontoon/pretranslation/pretranslate.py
vishalol: No extra code to cut off and reinsert the key (replace works fine). I will need to handle cases where there might be spaces around the `=` separator and then add them back (probably). 

A variation will work well for attributes as well. While removing attributes might be a hassle

eemeli: We really ought to extract each flat pattern, and have the ability to re-integrate it in its context. Consider for instance the following real-world message:

```fluent
detail-show-preferences =
    .label =
        { PLATFORM() ->
            [windows] Options
           *[other] Preferences
        }
    .tooltiptext =
        { PLATFORM() ->
            [windows] Change this add-on’s options
           *[other] Change this add-on’s preferences
        }
```

Here, we should be able to find and separate the four patterns in this message, translate each separately, and then feed them back into the message structure at the appropriate slots.

mathjazz: The ID getting translated is [the most pressing issue](https://github.com/mozilla/pontoon/issues/2691#issuecomment-1397546569), and #2690 is only about that. We'd like to fix that ASAP.

We've talked about "protecting" other parts of the syntax, and my suggestion was to only focus on the message ID.

mathjazz: Let's continue with `UNTRANSLATABLE_KEY` for now.

pontoon/pretranslation/pretranslate.py
```diff
@@ -70,12 +71,23 @@ def get_translations(entity, locale):
...
-            text=entity.string,
+            text=gt_input,
             locale=locale,
         )
 
         if gt_response["status"]:
+            if entity.string != gt_input:
+                gt_response["translation"] = gt_response["translation"].replace(
+                    UNTRANSLATABLE_KEY, entity.key
+                )
```
mathjazz, line 89: We could log cases where `UNTRANSLATABLE_KEY` doesn't exist in the translated string. But since this is a temporary workaround until we extend Pretranslation beyond internal testing, it's fine if we catch these cases during the review process.

mathjazz (approved):
Thanks, nice work!

PR #2720, opened 2023-01-20
Refactor Editor context

eemeli:
tl;dr: Internal changes that should be nearly invisible for users.

The key change here is the re-formulation of the editor state by defining and using an EditorMessage array to represent the `initial` and current `value` of the editor, along with a MessageEntry `entry` that's used to reconstruct the edited message.

The EditorMessage shape is shared across all source formats, reducing the special-ness of Fluent in the editor; the `view` indicator is now replaced by a boolean `sourceView` that controls the use of the source editor. One small UI change is that in the source view the FTL button is rendered with red text when the current message is invalid Fluent. The toggling between the `'simple'` and `'rich'` view is now controlled just by the number of fields that the message needs to display.

The purpose of these changes is to provide a better foundation for rich editor improvements, which now don't need to consider Fluent as a special case but just as one among a number of supported formats.

Many of the subsequent/related changes help reduce the amount of calculation that's required during input, as these data model changes make some of the comparisons easier and effectively cache more data.


mathjazz:
Just noticed that the editor prepends variant identifiers with "feminine" and "other" (instead of "masculine") for the following string: https://github.com/mozilla-l10n/pontoon-ftl/blob/master/en-US/demo.ftl#L111.

Seems like the bug is already present on `master`, but I wanted to bring it up in case you'd like to fix it as part of this patch for some reason. I haven't found any other issues.

eemeli:
Really? I'm not seeing that on either pontoon.mozilla.org or a local instance from this PR's branch. This is how the message renders for me:
Screenshot 2023-01-25 at 22 32 55

mathjazz:
Have you tried with an untranslated message?

eemeli:
Yes. I consistently get the same view as in the previous comment's screenshot.

mathjazz:
That's how it looks over here:

Screenshot 2023-01-25 at 21 50 47

mathjazz:
Noticed two changes in the source editor:
1. When I copy from Machinery, Locales or the source string, the whole content of the source editor is replaced, including the key. That results in the FTL icon going red.
2. There is no blank line at the end of the source string in the source editor. Is that intentional?

eemeli:
Updated with further preparatory steps for improving the editor, by merging the `genericeditor` and `fluenteditor` into one `translationform` module and making them both use a single EditField component to wrap the `<textarea>`. Also added `EditAccesskey` as a separate component, now based on `<input>` rather than `<textarea>`.

> 1. When I copy from Machinery, Locales or the source string, the whole content of the source editor is replaced, including the key. That results in the FTL icon going red.

That's a bug, will need to fix.

> 2. There is no blank line at the end of the source string in the source editor. Is that intentional?

Yes, it's intentional. The newline will still be there when adding to the database.

eemeli:
Looks like I'll still need to update this some to reduce the API conflict with CodeMirror.

Effectively, we need the Editor context to continuously provide the current value in a form that can be used for at least FtlSwitch to be able to depend on it, but if we can make that value always and only calculated from the `<textarea>` values rather than being directly set, we can make them uncontrolled components, and can let them (and later, CodeMirror instances) maintain their own current values.

mathjazz (approved):
I was struggling a little with the review here, because the initial commit is quite big. Going forward, splitting bigger PRs like this into smaller commits would be very appreciated.

I did test the code quite heavily and didn't spot any issues beyond the ones already reported and fixed.

Nice job!

eemeli:
I at least have a strong preference for each individual commit leaving a repo in a working state, which in this case meant that a deep change to the context shape can't really be split into smaller units.

Going forward, for any later large changes, should I relax that condition when packaging the work into commits?

mathjazz:
> Going forward, for any later large changes, should I relax that condition when packaging the work into commits?

Yes, please. :-)

mathjazz:
> When I copy from Machinery, Locales or the source string, the whole content of the source editor is replaced, including the key. That results in the FTL icon going red.

Looks like we missed this? You can reproduce it by going to https://pontoon.mozilla.org/sl/firefox-accounts/settings.ftl/?string=268291 and clicking FTL and then Copy.

PR #2727, opened 2023-01-23
Fix Pretranslation Locale QuerySet

mathjazz:
The QuerySet to select locales to perform pretranslation for a specific project on selects all locales that have pretranslation enabled for at least one project. This patch fixes that to only select locales that have pretranslation enabled for the given project.


eemeli (approved):

eemeli:
A test would've been nice.

PR #2729, opened 2023-01-26
Update URL of pontoon-test repo

mathjazz:
Remove optional fields from the instructions.


eemeli (approved):
Manually tested that this works. Needing to separately sync after saving was a bit unexpected.

mathjazz:
Yeah. I guess clicking on Sync if the project is not saved yet should Save + Sync.

PR #2736, opened 2023-01-28
add support for implicit TLS emails

ImUrX:


ImUrX:
Sorry forgot to link about more info: Django docs [specify this](https://docs.djangoproject.com/en/4.1/ref/settings/#email-use-ssl)

mathjazz:
Thanks for the patch!

Just curious if you could tell a bit more about the motivation for this addition?

ImUrX:
My provider prefers implicit TLS. Just that

mathjazz (approved):
Thanks for clarifying!

Going forward, I'd suggest to file an issue before opening a PR.

We also have a few [good first issues](https://github.com/mozilla/pontoon/labels/good%20first%20issue) in case you're interested!

Thanks for the patch!

PR #2738, opened 2023-01-30
Never include empty literals in simple editor view

### LINKED ISSUE #2737: Translation area is prefilled with {""} for terms
Example
https://pontoon.mozilla.org/mix/firefox-accounts/all-resources/?string=268239

The translation area shows `{ "" }`, although there are no matches whatsoever in MACHINERY.

eemeli:
Fixes #2737

### DIFF
--- a/translate/src/utils/message/getSimplePreview.test.js
+++ b/translate/src/utils/message/getSimplePreview.test.js
@@ -37,8 +37,15 @@ describe('getSimplePreview', () => {
     expect(res).toEqual('Avengers');
   });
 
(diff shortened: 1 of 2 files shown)


flodolo (approved):

mathjazz:
Messages like [this one](https://github.com/mozilla-l10n/pontoon-ftl/blob/master/en-US/demo.ftl#L102) are no longer represented as `{ "" }`. 

Instead, the text field and the source string element are empty.

That's expected as of this patch, right?

eemeli:
Yes, that's expected behaviour.

PR #2750, opened 2023-02-14
Fix date format in latest activity tooltip inconsistency with the rest of the app

### LINKED ISSUE #2611: Date format in Latest activity tooltip is inconsistent with the rest of the app
Screenshot 2022-08-19 at 15 05 51

We use the same format as in the Target Date column everywhere else (History Panel, Sync Log, Profile page, etc.).

To clarify: we should use "Aug 12, 2022" instead of "12 August 2022".

willianveiga:
Fixes #2611.

### DIFF
--- a/pontoon/base/static/js/table.js
+++ b/pontoon/base/static/js/table.js
@@ -9,22 +9,18 @@ $.expr[':'].containsi = function (a, i, m) {
 };
 
 /* Latest activity tooltip */
-var time_formatter = new Intl.DateTimeFormat('en-GB', {
+var date_formatter = new Intl.DateTimeFormat('en-US', {
+    day: 'numeric',
+    month: 'long',
+    year: 'numeric',
+  }),
+  time_formatter = new Intl.DateTimeFormat('en-GB', {
     hour: '2-digit',
     minute: '2-digit',
     second: '2-digit',
   }),
   timer = null,
-  delay = 500,
-  format_date = (date) => {
-    var date_formatter = new Intl.DateTimeFormat('en-GB', {
-        day: 'numeric',
-        month: 'short',
-        year: 'numeric',
-      }),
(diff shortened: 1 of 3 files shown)


pontoon/base/static/js/table.js
```diff
@@ -9,18 +9,22 @@ $.expr[':'].containsi = function (a, i, m) {
 };
 
 /* Latest activity tooltip */
-var date_formatter = new Intl.DateTimeFormat('en-GB', {
```
mathjazz, line 12: I'd just replace `en-GB` with `en-US` here and leave everything else as it is on master.

I'd even keep `month: 'long'`, because the space is not limited. But to be consistent, we then also need to set `month` value to `long` in `shortDateFormat` in pontoon/contributors/static/js/profile.js.

mathjazz (commented):
Thanks for the patch!

Please see my in-line comment - I think we can fix the bug with much smaller changeset.

mathjazz:
I just realized we also need to replace `en-GB` with `en-US` in pontoon/teams/static/js/bugzilla.js.

willianveiga:
Thanks, @mathjazz.

Do we want to change `en-GB` to `en-US` [here](https://github.com/mozilla/pontoon/pull/2750/files#diff-79b6a0398959ea93fb1e1c5c9028bdfbca94d1e7d2d884a9ab3e4024e89add95L17) as well?

mathjazz:
The current patch looks good!

> Do we want to change `en-GB` to `en-US` [here](https://github.com/mozilla/pontoon/pull/2750/files#diff-79b6a0398959ea93fb1e1c5c9028bdfbca94d1e7d2d884a9ab3e4024e89add95L17) as well?

Nice catch, let's do that and we're good!

mathjazz (approved):
Nice work!

PR #2754, opened 2023-02-15
Update JS dependencies

eemeli:
Refresh the lockfile & update rollup v2 -> v3, along with its plugins. `rollup-plugin-terser` is dropped as it's unused -- the minification is actually done by the server calling `terser` directly.


mathjazz (approved):
LGTM

PR #2756, opened 2023-02-16
Only exclude system users from contributor dashboards and permissions selectors

mathjazz:
This PR returns system users to the Authors filters in translate view.

I've also added a couple of minor tweaks to user role handling.

### DIFF
--- a/pontoon/base/tests/managers/test_user.py
+++ b/pontoon/base/tests/managers/test_user.py
@@ -5,8 +5,6 @@ from django.db.models import Q
 from pontoon.base.utils import aware_datetime
 from pontoon.contributors.utils import users_with_translations_counts
 
-from django.contrib.auth.models import User
-
 from pontoon.test.factories import (
     EntityFactory,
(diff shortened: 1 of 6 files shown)


pontoon/base/tests/managers/test_user.py
```diff
@@ -325,22 +323,3 @@ def test_mgr_user_query_args_filtering(
     assert top_contribs[0].translations_approved_count == 11
     assert top_contribs[0].translations_rejected_count == 0
     assert top_contribs[0].translations_unapproved_count == 3
-
-
-@pytest.mark.django_db
-def test_mgr_system_user_contributors_excluded(
```
eemeli, line 331: Why is this test being dropped, and why are no tests added for the changes made here?

mathjazz: This test is no longer relevant, because we no longer exclude system users in that method.

But good point, I should add some test.

eemeli (approved):

PR #2759, opened 2023-02-16
Set refs correctly in EditAccesskey

### LINKED ISSUE #2758: UI is broken with label and accesskey
https://pontoon.mozilla.org/id/firefox/browser/browser/browser.ftl/?search=urlbar-result-menu-learn-more&string=274827

Try to edit the accesskey, the label will be replaced. Even clicking the button character will replace the label.

Please treat this as P1, as errors are starting to come in :-(

eemeli:
Fixes #2758

### DIFF
--- a/translate/src/context/Editor.tsx
+++ b/translate/src/context/Editor.tsx
@@ -101,7 +101,9 @@ export type EditorData = Readonly<{
   entry: MessageEntry;
 
   /** Editor input components */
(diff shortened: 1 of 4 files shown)


mathjazz (approved):
Confirming the problem can no longer be reproduced.

Can we test this somehow so we won't hit the same problem again?

eemeli:
> Can we test this somehow so we won't hit the same problem again?

Not easily with our current solution. Filed a follow-up issue to make it easier going forward.

PR #2762, opened 2023-02-16
Show "All contributions" by default when accessing own profile

### LINKED ISSUE #2624: Show "All contributions" by default when accessing own profile
We currently use "Suggestions and reviews" as the default contribution type filter on the profile page. That allows managers, admins and also external users (e.g. employers) to get a quick overview over user's own activity right after the page is loaded.

But when the user is looking at their own profile, they are more likely interested in "All contributions", which also includes activity of other users on their translations (received reviews).

Such page has a potential to be used as the user homepage, because it aggregates all recent activity relevant for the user in one place.

willianveiga:
Fixes #2624.

### DIFF
--- a/pontoon/contributors/templates/contributors/profile.html
+++ b/pontoon/contributors/templates/contributors/profile.html
@@ -288,7 +288,7 @@
                     <div class="type-selector controls select">
                         <div class="button selector">
                             <div class="value">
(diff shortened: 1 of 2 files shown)


mathjazz:
Thanks for the patch!

The title, selector and timeline are updated as expected, but the contribution-graph (GitHub-style calendar) is not.

mathjazz (approved):
Very well done! 👍

PR #2765, opened 2023-02-20
Do not mark translations of DB projects as needing sync

### LINKED ISSUE #2620: Do not mark translations of DB projects as needing sync
When a translation status changes in a way that requires sync, we store it into a `ChangedEntityLocale` table.

We currently do that for all projects, but we should only do it for projects that are synced (i.e. use code repository as their store instead of internal pontoon DB, like Terminology and Tutorial projects).

willianveiga:
Fixes #2620.


pontoon/base/migrations/0040_remove_changed_entity_locale_entries_for_repository_projects.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 3.2.15 on 2023-02-23 16:02
+from django.db import migrations
+from pontoon.base.models import Project, Resource, Entity, ChangedEntityLocale
+
+
+def remove_changed_entity_locale_entries_for_repository_projects(apps, schema_editor):
+    projects = Project.objects.filter(data_source=Project.DataSource.DATABASE)
+    resources = Resource.objects.filter(project__in=projects)
+    entities = Entity.objects.filter(resource__in=resources)
+    ChangedEntityLocale.objects.filter(entity__in=entities).delete()
```
mathjazz, line 10: You can [follow relationship](https://docs.djangoproject.com/en/4.1/topics/db/queries/#lookups-that-span-relationships) in QuerySet lookups:

```python
ChangedEntityLocale.objects.filter(
    entity__resource__project__data_source=Project.DataSource.DATABASE
).delete()
```

mathjazz (commented):
Very nice work! Love the test!

Just one stylistic suggestion, other than that this is good to go.

pontoon/base/migrations/0040_remove_changed_entity_locale_entries_for_repository_projects.py
```diff
@@ -0,0 +1,24 @@
...
+
+    initial = True
+
+    dependencies = [
+        ("base", "0039_mark_system_users"),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            code=remove_changed_entity_locale_entries_for_repository_projects,
```
mathjazz, line 22: Sorry just realized that we should add a reverse_code here, [noop](https://docs.djangoproject.com/en/4.1/ref/migration-operations/#django.db.migrations.operations.RunPython.noop) is fine.

See migration [`0023_fuzzy_to_missing.py`](https://github.com/mozilla/pontoon/blob/1ae0a58ad0537eb35d80393187202fd523df8bec/pontoon/base/migrations/0027_keep_pretranslation_enabled.py) for example.

mathjazz:
@willianveiga I've tested the patch and it works as expected.

There are however a few places where we create `ChangedEntityLocale` objects in batch:
* https://github.com/mozilla/pontoon/blob/master/pontoon/base/utils.py#L431
* https://github.com/mozilla/pontoon/blob/master/pontoon/batch/views.py#L29
* https://github.com/mozilla/pontoon/blob/master/pontoon/pretranslation/tasks.py#L172

The first is not in use for DB projects, because it handles file uploads, but the 2nd and the 3rd are, so they need to ignore entities belonging to DB projects.

Also, we should centralize this code with a single `ChangedEntityLocale.objects.bulk_mark_changed()` [custom manager](https://docs.djangoproject.com/en/4.1/topics/db/managers/#custom-managers), but we can also do that in a followup patch.

pontoon/pretranslation/tasks.py
```diff
@@ -173,6 +173,9 @@ def pretranslate(self, project_pk, locales=None, entities=None):
     changed_entities = {}
     existing = ChangedEntityLocale.objects.values_list("entity", "locale").distinct()
     for t in translations:
+        if t.entity.resource.project.data_source == Project.DataSource.DATABASE:
```
mathjazz, line 176: We should add project to the [`prefetch_related()`](https://docs.djangoproject.com/en/4.1/ref/models/querysets/#prefetch-related) call [above](https://github.com/mozilla/pontoon/blob/f3727dd6afe8d621d61d2a4311818d5781dc517d/pontoon/pretranslation/tasks.py#L68), which will retrieve project data from the DB in one call, instead of doing that for each translation separately. DB queries are slow, so we should reduce them.

Note that you can simply replace `resource` with `resource__project` in the call above.

pontoon/batch/views.py
```diff
@@ -31,6 +31,9 @@ def mark_changed_translation(changed_entities, locale):
     changed_entities_array = []
     existing = ChangedEntityLocale.objects.values_list("entity", "locale").distinct()
     for changed_entity in changed_entities:
+        if changed_entity.resource.project.data_source == Project.DataSource.DATABASE:
```
mathjazz, line 34: Same as below, let's make sure we don't query the DB for each translation by prefetching project data here:
https://github.com/mozilla/pontoon/blob/f3727dd6afe8d621d61d2a4311818d5781dc517d/pontoon/batch/utils.py#L37

mathjazz (approved):
Nicely done! Thanks for the patch and sorry again for the back and forth in the review process.

PR #2770, opened 2023-02-25
CC requester to the email sent when requesting a new project

### LINKED ISSUE #2769: CC requester to the email sent when requesting a new project
Hopefully that will reduce the number of duplicate requests we receive.

willianveiga:
Fixes #2769.

### DIFF
--- a/pontoon/teams/views.py
+++ b/pontoon/teams/views.py
@@ -270,17 +270,18 @@ def request_item(request, locale=None):
     if settings.PROJECT_MANAGERS[0] != "":
         template = get_template("teams/email_request_item.jinja")
         mail_body = template.render(payload)
-        cc = set(
-            list(locale.managers_group.user_set.values_list("email", flat=True))
-            + [user.contact_email()]
-        )
+        cc = {user.contact_email()}
+        if locale:
+            cc.update(
+                set(locale.managers_group.user_set.values_list("email", flat=True))
+            )
 
         EmailMessage(
             subject=mail_subject,
             body=mail_body,
             from_email=settings.LOCALE_REQUEST_FROM_EMAIL,
             to=settings.PROJECT_MANAGERS,
-            cc=cc if locale else "",
+            cc=cc,
             reply_to=[user.contact_email()],
         ).send()
     else:
--- a/pontoon/teams/views.py
+++ b/pontoon/teams/views.py
@@ -270,10 +270,10 @@ def request_item(request, locale=None):
     if settings.PROJECT_MANAGERS[0] != "":
         template = get_template("teams/email_request_item.jinja")
         mail_body = template.render(payload)
-
-        cc = list(locale.managers_group.user_set.values_list("email", flat=True))
-        if user.profile.contact_email:
-            cc = set(cc + [user.profile.contact_email])
+        cc = set(
+            list(locale.managers_group.user_set.values_list("email", flat=True))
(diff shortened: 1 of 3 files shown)


pontoon/teams/views.py
```diff
@@ -276,9 +276,12 @@ def request_item(request, locale=None):
...
             from_email=settings.LOCALE_REQUEST_FROM_EMAIL,
             to=settings.PROJECT_MANAGERS,
-            cc=locale.managers_group.user_set.exclude(pk=user.pk).values_list(
-                "email", flat=True
+            cc=list(
+                locale.managers_group.user_set.exclude(pk=user.pk).values_list(
+                    "email", flat=True
+                )
             )
+            + [user.email]
```
mathjazz, line 284: We should use contact email address here if it is set.

pontoon/teams/views.py
```diff
@@ -276,9 +276,12 @@ def request_item(request, locale=None):
             body=mail_body,
             from_email=settings.LOCALE_REQUEST_FROM_EMAIL,
             to=settings.PROJECT_MANAGERS,
-            cc=locale.managers_group.user_set.exclude(pk=user.pk).values_list(
-                "email", flat=True
+            cc=list(
+                locale.managers_group.user_set.exclude(pk=user.pk).values_list(
```
mathjazz, line 280: We no longer need to exclude the person that makes the request.

pontoon/teams/views.py
```diff
@@ -276,9 +276,12 @@ def request_item(request, locale=None):
             body=mail_body,
             from_email=settings.LOCALE_REQUEST_FROM_EMAIL,
             to=settings.PROJECT_MANAGERS,
-            cc=locale.managers_group.user_set.exclude(pk=user.pk).values_list(
-                "email", flat=True
+            cc=list(
```
mathjazz, line 279: It's possible that the person that makes the request is also a locale manager, in which case we'll list their email address twice. Let's prevent this by wrapping the entire value of `cc` in `set()`.

mathjazz (commented):
Nice work! Left a few suggestions for improvements, nothing major.

pontoon/teams/views.py
```diff
@@ -271,16 +271,16 @@ def request_item(request, locale=None):
         template = get_template("teams/email_request_item.jinja")
         mail_body = template.render(payload)
 
+        cc = list(locale.managers_group.user_set.values_list("email", flat=True))
+        if user.profile.contact_email:
+            cc = set(cc + [user.profile.contact_email])
```
mathjazz, line 276: We should CC the user that has made the request even if their `contact_email` address is not set.

pontoon/teams/views.py
```diff
@@ -270,18 +270,18 @@ def request_item(request, locale=None):
...
         EmailMessage(
             subject=mail_subject,
             body=mail_body,
             from_email=settings.LOCALE_REQUEST_FROM_EMAIL,
             to=settings.PROJECT_MANAGERS,
-            cc=locale.managers_group.user_set.exclude(pk=user.pk).values_list(
-                "email", flat=True
-            )
-            if locale
-            else "",
```
mathjazz, line 283: Have you removed this check intentionally? I think we should keep it around to support the "New locale" requests.

mathjazz (approved):
Great work! 👍

PR #2778, opened 2023-03-02
Factor out bulk ChangedEntityLocale creation

### LINKED ISSUE #2771: Factor out bulk ChangedEntityLocale creation
There are a few places where we create `ChangedEntityLocale` objects in batch:
* https://github.com/mozilla/pontoon/blob/master/pontoon/base/utils.py#L431
* https://github.com/mozilla/pontoon/blob/master/pontoon/batch/views.py#L29
* https://github.com/mozilla/pontoon/blob/master/pontoon/pretranslation/tasks.py#L172

Since there's a lot of duplication in these three blocks, we should replace them with a single [custom manager](https://docs.djangoproject.com/en/4.1/topics/db/managers/#custom-managers) (e.g. `ChangedEntityLocale.objects.bulk_mark_changed()`).

_Originally posted by @mathjazz in https://github.com/mozilla/pontoon/issues/2765#issuecomment-1448292500_

willianveiga:
Fixes #2771.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -3207,9 +3207,7 @@ class ChangedEntityLocaleManager(models.Manager):
     def bulk_mark_changed(self, translations, locale=None):
         """Mark entities as changed, for later sync."""
         changed_entities = {}
-        existing = ChangedEntityLocale.objects.values_list(
-            "entity", "locale"
-        ).distinct()
+        existing = self.values_list("entity", "locale").distinct()
 
         for translation in translations:
             if (
@@ -3226,7 +3224,7 @@ class ChangedEntityLocaleManager(models.Manager):
                     entity=translation.entity, locale=used_locale
                 )
 
-        ChangedEntityLocale.objects.bulk_create(changed_entities.values())
+        self.bulk_create(changed_entities.values())
 
 
 class ChangedEntityLocale(models.Model):
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -3203,6 +3203,32 @@ class Entity(DirtyFieldsMixin, models.Model):
         return entities_array
 
 
+class ChangedEntityLocaleManager(models.Manager):
+    def bulk_mark_changed(self, translations, locale=None):
+        """Mark entities as changed, for later sync."""
+        changed_entities = {}
+        existing = ChangedEntityLocale.objects.values_list(
+            "entity", "locale"
+        ).distinct()
+
+        for translation in translations:
+            if (
+                translation.entity.resource.project.data_source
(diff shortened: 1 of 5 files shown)


pontoon/base/models.py
```diff
@@ -3203,6 +3203,30 @@ def map_entities(
         return entities_array
 
 
+class ChangedEntityLocaleManager(models.Manager):
+    def bulk_mark_changed(self, translations, locale=None):
```
mathjazz, line 3207: Sorry, I mislead you with my note in https://github.com/mozilla/pontoon/issues/2771#issue-1603963121. Since Translation QuerySet is used as a parameter, it would make more sense to move this logic under `TranslationQuerySet`. 

The API call would then be simply `translations.bulk_mark_changed()`, without any parameters, because we can get locale from each translation object. Before generating the translations QuerySet, we'd need to prefetch related `entity__resource__project` data (and revert prefetching established in #2765).

Going forward, we should file an issue to also move `mark_changed()` [model method](https://docs.djangoproject.com/en/4.1/topics/db/models/#model-methods.) from Entity to Translation model. It's currently always used as `translation.entity.mark_changed()`. And finally, migrate ChangedEntityLocale model to ChangedTranslation. The reason this is currently bound to Entity rather than Translation model is a relic of the past when we had the ability to delete translations without rejecting them first.

pontoon/base/models.py
```diff
@@ -3282,6 +3282,28 @@ def for_checks(self, only_db_formats=True):
 
         return translations
 
+    def bulk_mark_changed(self):
+        changed_entities = {}
+        existing = ChangedEntityLocale.objects.values_list(
+            "entity", "locale"
+        ).distinct()
+
+        for translation in self:
```
mathjazz, line 3291: You can `self.exclude()` DB-projects here instead of using the if-block below. That will also save us the need to prefetch data.

pontoon/batch/utils.py
```diff
@@ -35,9 +35,7 @@ def get_translations_info(translations, locale):
     count = translations.count()
     translated_resources = list(translations.translated_resources(locale))
     changed_entities = list(
-        Entity.objects.filter(translation__in=translations)
-        .distinct()
-        .prefetch_related("resource__project")
```
mathjazz, line 40: We should preserver this one, because it was there before. :)

mathjazz (commented):
Thanks for the updates!

The patch looks good.

Please restore that prefetch, which was probably removed by mistake as a whole. I'll take it for a spin once that's update.

mathjazz (approved):
Works great, thanks!

Issue #2780, opened 2023-03-06
Fluent Plurals UI broken for untranslated strings in some locales


flodolo:
https://pontoon.mozilla.org/tg/firefox/browser/browser/aboutLogins.ftl/?string=275911

First, when you open the string, the input field is empty and there's no variants UI.

If you copy the English string, then things get even weirder: you type in the "other" line, and the text ends up in "one" (but only one character).

mathjazz:
Some early findings...

Same set of problems can be reproduced for `id` or `th`: 
https://pontoon.mozilla.org/id/firefox/all-resources/?string=275911
https://pontoon.mozilla.org/th/firefox/all-resources/?string=275911

But not for `nb-NO` or `or`:
https://pontoon.mozilla.org/nb-NO/firefox/all-resources/?string=275911
https://pontoon.mozilla.org/or/firefox/all-resources/?string=275911

Special case is Spanish, which shows an unexpected plural form - `many` - without an example:
https://pontoon.mozilla.org/es-ES/firefox/all-resources/?string=275911

flodolo:
French has the same behavior as Spanish, but it doesn't do the weird typing bug.

So, it looks like we're picking the wrong number of plural forms, and break the UI when there's only one?

mathjazz:
Right, for `id` and `th` 1 field is actually expected. But not for `tg`.

eemeli:
One part of the disparity is coming from relying on `Intl.PluralRules`. In today's Firefox Nightly 112.0a1:

```js
new Intl.PluralRules('tg').resolvedOptions().pluralCategories 
// [ "other" ]

new Intl.PluralRules('es').resolvedOptions().pluralCategories
// [ "many", "one", "other" ]
```
This effectively matches the behaviour at runtime, which also relies on `Intl.PluralRules`.

As at least Chrome appears to not only rely on CLDR data, a `tg` plural string will work as expected there: https://pontoon.mozilla.org/tg/firefox/all-resources/?string=275911

The `many` in Spanish and all other Romance languages is a valid [CLDR category](https://unicode-org.github.io/cldr-staging/charts/latest/supplemental/language_plural_rules.html#es) for millions; previously only available in French. It doesn't contain an example because our [iteration](https://github.com/mozilla/pontoon/blob/master/translate/src/hooks/usePluralExamples.ts) stops at 1000.

The editor bug is a real bug that needs to be fixed ASAP, and separate from all of the above.

flodolo:
[...]

PR #2782, opened 2023-03-07
Keep editor context fields value updated on relevant changes

eemeli:
Fixes the field entry bug part of #2780.

When the shape of a message changes, the Editor context's `fields` array was not being updated like it should be. This manifested in the reported issue as copying the source changed the count 1 -> 2, and editing the second variant broke things.


mathjazz (approved):
Confirming this fixes the problem for me!

PR #2794, opened 2023-03-12
Move mark_changed() model method from Entity to Translation model

### LINKED ISSUE #2792: Move `mark_changed()` model method from Entity to Translation model
Let's move `mark_changed()` [model method](https://docs.djangoproject.com/en/4.1/topics/db/models/#model-methods.) from Entity to Translation model. It's currently always used as `translation.entity.mark_changed()`. 

~And finally, migrate ChangedEntityLocale model to ChangedTranslation. The reason this is currently bound to Entity rather than Translation model is a relic of the past when we had the ability to delete translations without rejecting them first.~

_Originally posted by @mathjazz in https://github.com/mozilla/pontoon/pull/2778#discussion_r1126242719_

willianveiga:
Fixes #2792.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -2904,16 +2904,6 @@ class Entity(DirtyFieldsMixin, models.Model):
         """
         return locale in self.changed_locales.all()
 
-    def mark_changed(self, locale):
-        """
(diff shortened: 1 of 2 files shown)


mathjazz (approved):
Excellent work! 👍

PR #2795, opened 2023-03-13
SYNC button should be disabled when a project is disabled

### LINKED ISSUE #2724: SYNC button should be disabled when a project is disabled
I didn't realize I was on a disabled project this morning, and clicked the SYNC button. I think it would make sense to disable the button if the project is disabled.

willianveiga:
Fixes #2724.

### DIFF
--- a/pontoon/administration/templates/admin_project.html
+++ b/pontoon/administration/templates/admin_project.html
@@ -337,8 +337,8 @@
               {{ form.sync_disabled }}{{ form.sync_disabled.label }}
           </label>
       </div>
-      {% if settings.MANUAL_SYNC %}
-      <button class="sync button"{% if project.disabled %} disabled{% endif %}>Sync</button>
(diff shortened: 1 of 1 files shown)


mathjazz (commented):
Nice work!

We should have some visual feedback that the button is hidden. Maybe we should simply hide it?

Also, we should disabled sync also in case Sync Disabled is True.

mathjazz (commented):
Thanks, works great!

Nit: perhaps we should remove the brackets, since they aren't really needed?

mathjazz (approved):

flodolo:
> Also, we should disabled sync also in case Sync Disabled is True.

I'm not sure about this. 

Not sure how other PMs work, but I typically disable SYNC temporarily (e.g. for migrations), but then I want to sync manually once before re-enabling sync.

mathjazz:
Oh, right. You can still sync manually, even if sync is disabled. Let's restore that.

PR #2799, opened 2023-03-15
Move translate modules from core/ to modules/

eemeli:
This finally cleans up the effectively useless split of some parts of the translate front-end code being hosted under `translate/src/core/`, and some under `translate/src/modules/`. The `term` and `terms` modules are also merged.


mathjazz (approved):
Very nice!

PR #2800, opened 2023-03-15
Add support for React Testing Library

### LINKED ISSUE #2760: Add support for react-testing-library tests in the translate front-end
Testing the front-end is in general hard, and expanding the enzyme test suite does not allow for as much flexibility as we really need (cf. #2759, and many other places where front-end tests would be desirable).

We should start introducing the [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/) in parallel with enzyme, and later consider porting our existing tests to it. That would remove at least one blocker from moving onwards from React v16.

eemeli:
Fixes #2760 

~This conflicts a bit with #2799, but either should be pretty easy to rebase on the other~ _Edit: conflicts resolved._

Minimal steps are included for being able to write tests using the [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/), and an exemplary few of the tests are ported to it. The style of said tests isn't really updated here at all, so they're not really testing the UX yet.

The version of `@testing-library/react` that's installed here is v12 rather than the most recent v14, as that's the last that supports React v16. The upstream doesn't look like it's gotten too far yet in the last year of updates, but that'll indubitably happen eventually if we don't get rid of our React v16 hangups.

I also added a `MockStore` to `test/store.jsx` as a more useful abstraction than `mountComponentWithStore()`.


mathjazz (approved):
Nice work! And sorry for the delay...

We should track work  needed to move onwards from React v16 somewhere.

PR #2801, opened 2023-03-15
Use correct match for highlighted terms

### LINKED ISSUE #2706: Unable to reveal term popup
Sometimes the term popup doesn't show up after clicking on an underlined term in the source string:

https://pontoon.mozilla.org/sl/focus-for-android/all-resources/?string=267696
https://pontoon.mozilla.org/sl/focus-for-android/all-resources/?string=266420

It seems like it happens when the highlighted phrase isn't a perfect match of the term it belongs to.

eemeli:
Fixes #2706


mathjazz (approved):
Please fix prettier.

PR #2802, opened 2023-03-16
Pretranslate Fluent plural expression with target locale plural forms

### LINKED ISSUE #2734: Pretranslate Fluent plural messages for each plural form used by the locale
Some locales have different plurals forms than English (one, many). We should take that into accounts when pretranslating plural messages and add, remove or modify plural forms in the pretranslation accordingly.

If plural message contains any literal variants (e.g. 1), we should preserve them in pretranslation.

We'll have to identify the right source plural form to be used as the source for each target plural form.

mathjazz:
Fix #2734.

For each plural expression, we change plural forms to match those of the target locale.

The code in `fluent.py` is inspired by the ([legacy](https://github.com/mozilla/pontoon/blob/26c5f709ce9ddec66ec214c4ec8db722196ea24c/translate/src/utils/message/getEmptyMessage.ts#L31)) frontend code.

### DIFF
--- a/pontoon/base/fluent.py
+++ b/pontoon/base/fluent.py
@@ -189,7 +189,7 @@ def create_locale_plural_variants(node, locale):
         key = variant.key
         if isinstance(key, ast.NumberLiteral):
             numeric.append(variant)
-        elif isinstance(key, ast.Identifier) and key.name in locale.cldr_plurals_list():
+        else:
             source_plurals[key.name] = variant
         if variant.default:
             default = variant
--- a/pontoon/base/fluent.py
+++ b/pontoon/base/fluent.py
@@ -194,7 +194,7 @@ def create_locale_plural_variants(node, locale):
         if variant.default:
             default = variant
 
-    for plural in locale.cldr_plurals_list().split(", "):
+    for plural in locale.cldr_plurals_list():
         if plural in source_plurals.keys():
             variant = source_plurals[plural]
         else:
--- a/pontoon/base/fluent.py
+++ b/pontoon/base/fluent.py
@@ -202,7 +202,7 @@ def create_locale_plural_variants(node, locale):
             variant.key.name = plural
         variants.append(variant)
 
-    if len([v for v in variants if v.default]) == 0:
+    if any(v.default for v in variants):
         variants[-1].default = True
 
     node.variants = variants
--- a/pontoon/base/fluent.py
+++ b/pontoon/base/fluent.py
@@ -158,3 +158,18 @@ def get_simple_preview(content):
         tree = translation_ast.attributes[0]
 
     return serialize_value(tree.value)
+
+
+def is_plural_expression(expression):
+    from pontoon.base.models import Locale
+
+    CLDR_PLURALS = [c for _, c in Locale.CLDR_PLURALS]
+
+    if isinstance(expression, ast.SelectExpression):
+        return all(
+            isinstance(variant.key, ast.NumberLiteral)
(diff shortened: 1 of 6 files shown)


pontoon/base/fluent.py
```diff
@@ -158,3 +159,53 @@ def get_simple_preview(content):
...
+    for plural in locale.cldr_plurals_list().split(", "):
+        if plural in source_plurals.keys():
+            variant = source_plurals[plural]
+        else:
+            variant = copy.deepcopy(default)
+            variant.default = False
+            variant.key.name = plural
+        locale_plurals.append(variant)
+
+    variants = numeric + locale_plurals
```
eemeli, line 206: Why are `numeric` and `locale_plurals` accumulated separately and then joined, rather than just `variants` being built directly?

pontoon/base/fluent.py
```diff
@@ -158,3 +159,53 @@ def get_simple_preview(content):
...
+
+    if len([v for v in variants if v.default]) == 0:
+        variants[-1].default = True
```
eemeli, line 209: This results in the wrong default case for some Slavic languages. For plurals, the default should be defined by the locale rather than the source message.

If this isn't possible right now, then a code comment somewhere around this function should note this need for later improvement.

pontoon/base/tests/test_fluent.py
```diff
@@ -193,3 +199,145 @@ def test_flat_transformer_value_and_attributes():
...
+
+def test_is_plural_expression_all_numbers():
+    # Return True if all variant keys are numbers
+    input = dedent(
+        """
+        my-entry =
+            { $num ->
+                [1] Hello!
+               *[2] World!
+            }`
```
eemeli, line 244: If we do have selectors like this, should we really generate plurals even though the source doesn't include even `other`? To me, this looks like a selector on a number, but not a plural selector.

eemeli (changes requested):
Looks pretty decent as a starting point; some detaily things below.

Later on, we should start replacing the formatted `{ $count }` placeholders in what we're sending with some actually appropriate numbers for each category, and then re-integrating the placeholders in what comes back.

pontoon/base/fluent.py
mathjazz: How do we know which is the default variant for the locale?

pontoon/base/tests/test_fluent.py
mathjazz: I'm open to that change, however I'm just following the logic we use in the frontend. If we were to make a change, it needs to happen on both ends.

pontoon/base/fluent.py
eemeli: It should be the last entry in `cldr_plurals_list()`, as long as that's either `many` or `other`. If it's neither, then we've bad data and really ought to either fix the data or add an `other` at the end.

pontoon/base/tests/test_fluent.py
eemeli: Added #2805 for this.

pontoon/base/fluent.py
mathjazz: That sounds like we should always just use the last plural category (implemented in [1f64847](https://github.com/mozilla/pontoon/pull/2802/commits/1f648473290c5fc143e148f2b6d88db2ec5a4395)) or fix data (filed https://github.com/mozilla/pontoon/issues/2807).

eemeli (approved):

PR #2811, opened 2023-03-28
copy translation source content to clipboard

gr8b:
PR adds new button "COPY CONTENT" next to "COPY LINK". Button allow to copy translation source to clipboard.

### DIFF
--- a/translate/public/locale/en-US/translate.ftl
+++ b/translate/public/locale/en-US/translate.ftl
@@ -235,6 +235,8 @@ editor-UnsavedChanges--proceed = PROCEED
 
 entitydetails-EntityNavigation--link = <glyph></glyph>COPY LINK
     .title = Copy Link to String
+entitydetails-EntityNavigation--copysource = <glyph></glyph>COPY CONTENT
+    .title = Copy Source to Clipboard
 entitydetails-EntityNavigation--next = <glyph></glyph>NEXT
     .title = Go To Next String (Alt + Down)
 entitydetails-EntityNavigation--previous = <glyph></glyph>PREVIOUS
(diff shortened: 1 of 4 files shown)


mathjazz:
Hey @gr8b, thanks for the patch!

Could you explain a little bit more, why do you think we should add this feature?

eemeli:
> Currently, multiline translation strings in Pontoon are automatically displayed with pilcrow paragraph breaks (¶) [...] But to send the translation text into an external software, for example, ChatGPT or some other translation wizard you will need to delete the ¶ manually.

Really? This might be a browser issue. What OS & browser are you using?

At least for me, copying source text that includes pilcrows will leave them out of what ends up on the clipboard.

Here's an example multiline string to play around with: https://pontoon.mozilla.org/fi/firefox/toolkit/toolkit/about/aboutMozilla.ftl/?string=213848

gr8b:
@eemeli  Your link works, I successfully copied selection to clipboard without pilcrows characters. You can try to select and copy [this](https://translate.zabbix.com/lv/documentation-60/manual/api/reference/authentication/update.xliff/?string=64212) translation and you will see it will be copied with pilcrows. Is it possible because of outdated version of pontoon?

eemeli:
> Is it possible because of outdated version of pontoon?

Yes. If you take a look at the source, on pontoon.mozilla.org you'll see that the pilcrow is within a `<span aria-hidden>`, while on translate.zabbix.com it isn't.

gr8b:
Thank you, then I don't think there's any point in this PR.

rostislav-palivoda:
I think its useful functionality because big text blocks are hard to select with mouse and CTRL+A does not work in this input window.

PR #2817, opened 2023-04-17
Update to compare-locales 9.0.1, fluent-syntax 0.19.0, pip-tools 6.13.0

eemeli:
Tested this locally, seems to work as intended. Did check that no local code appears to depend on `pytoml`, which is here replaced by `toml`.

The `pip-tools` update is necessary for pip 23.1 compatibility.


mathjazz (approved):
LGTM

PR #2818, opened 2023-04-19
Bulk enable custom translators for project locales

haodave:
Add "Enable custom translators" action for Project Locales in admin panel

### DIFF
--- a/pontoon/base/admin.py
+++ b/pontoon/base/admin.py
@@ -1,21 +1,18 @@
 import uuid
 
-from django.contrib import admin
+from django.contrib import admin, messages
 from django.contrib.auth import get_user_model
 from django.contrib.auth.admin import UserAdmin as AuthUserAdmin
 from django.contrib.auth.models import User
-from django.forms.models import ModelForm
 from django.forms import ChoiceField
+from django.forms.models import ModelForm
 from django.urls import reverse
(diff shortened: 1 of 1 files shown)


mathjazz:
Thanks for the patch! Since there's no issue assigned to this PR, could you please elaborate on what's the motivation for this change?

haodave:
> Thanks for the patch! Since there's no issue assigned to this PR, could you please elaborate on what's the motivation for this change?

When there are 20+ or 30+ project locales in a project, it's a painful process to enable custom translators for all of them.

As for the reason of enabling custom translators, we want the users in the project/locale translators groups to be able to save and approve the translations directly. And this seems the way to go.

mathjazz:
Sorry for dropping the ball on this!

If I understand things correctly, by clicking on locales in the Django admin, you'd only make the widget for per-project permissions in /LOCALE_CODE/permissions/ visible. Currently, you need to instead select the project from the list in that very Permissions project.

In both cases, you still need to actually set permissions there.

I don't think this change makes settings permissions much faster, but I understand how the current UX doesn't work well for your workflow. I suggest we continue the conversation by filing an issue, describing the problem, possibly even write a spec and then working on the implementation.

Sorry again this took forever to review!

Issue #2225, opened 2020-04-26
Authentication with Keycloak on Pontoon


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1633281](https://bugzilla.mozilla.org/show_bug.cgi?id=1633281)
Bug Reporter: nathan.strobbe@gmail.com
CC:   @mathjazz

User Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.132 Safari/537.36

Steps to reproduce:

I tried to login with a custom SSO provider such as Keycloak, but it's not implemented yet. Currently, it's only support Google, Github and Gitlab

Actual results:

Keycloak provider is not implemented

Expected results:

You can login with Keycloak, providing in configuration :
- Keycloak URL
- Realm name
- Client id
- Client secret

PR #2820, opened 2023-04-19
Allow using Keycloak for authentication

haodave:


mathjazz:
Thanks for the patch!

Please fix `black` and `flake8` issues.

Adding a reference to #2225, which this patch fixes.

haodave:
> Thanks for the patch!
> 
> Please fix `black` and `flake8` issues.
> 
> Adding a reference to #2225, which this patch fixes.

Glad to contribute. :)

Reformatted the files. Could you check again?

PR #2826, opened 2023-05-03
Convert multiline messages into single line before pretranslation

### LINKED ISSUE #2810: Pretranslation issues with multiline strings
Pretranslation struggles to translate Fluent strings where line breaks are used to limit the line width, as it treats each line as a separate sentence.

```
rec-pw-1-2 =
  Make this password unique and different from any others you use.
  A good strategy to follow is to combine two or more unrelated
  words to create an entire pass phrase, and include numbers and symbols.
```

I don't think we can completely ignore line breaks, because there are some cases where these have an effect on display ([example](https://searchfox.org/mozilla-central/rev/169bf38e150667afac81ab73ef8b5ace8f1dfa8d/browser/locales/en-US/browser/downloads.ftl#244-248)). But maybe we can strip line breaks (i.e. replace them with a whitespace) when sending the translation to the machine translation engine, hoping that translators will catch edge cases when reviewing?

Cc @eemeli because that's a topic that we discussed at some point for the Fluent editor.

mathjazz:
Fixes #2810.

### DIFF
--- a/pontoon/pretranslation/pretranslate.py
+++ b/pontoon/pretranslation/pretranslate.py
@@ -32,11 +32,15 @@ class PretranslationTransformer(FlatTransformer):
         return self.generic_visit(node)
 
     def visit_TextElement(self, node):
-        pretranslation, service = get_pretranslated_data(node.value, self.locale)
+        # Machine translation treats each line as separate sentence,
+        # hence we replace newline characters with spaces.
+        source = node.value.replace("\n", " ")
+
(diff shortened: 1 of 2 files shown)


eemeli (approved):
I'll need to rebase my current work on this, but that should be pretty straightforward.

mathjazz:
> I'll need to rebase my current work on this, but that should be pretty straightforward.

Sorry about that!

PR #2828, opened 2023-05-03
Replace LOCALE_REQUEST_FROM_EMAIL with DEFAULT_FROM_EMAIL

mathjazz:
We have a `DEFAULT_FROM_EMAIL` variable, but we use a different setting for the FROM email address when requesting locales to be added to projects. There's no need for two different values, really.

Additionally, this patch changes the `noreply` part of the `DEFAULT_FROM_EMAIL` because such addresses canot be verified and because that's [apparently a bad idea in general](https://mailchimp.com/what-to-know-about-using-no-reply-email/).

### DIFF
--- a/docs/admin/deployment.rst
+++ b/docs/admin/deployment.rst
@@ -83,7 +83,7 @@ you create:
 
 ``DEFAULT_FROM_EMAIL``
    Optional. Default email address to send emails from. Default value:
-   ``Pontoon <noreply@hostname>``.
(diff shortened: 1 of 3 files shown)


eemeli (approved):
`docs/admin/deployment.rst` also needs to be updated.

mathjazz:
> `docs/admin/deployment.rst` also needs to be updated.

Nice catch!

PR #2829, opened 2023-05-03
Add `<span translate="no">` wrapper for Fluent expressions passing through Google

### LINKED ISSUE #2181: Google Translate modifies whitespace within placeables, causing errors
## [Bug 1584178](https://bugzilla.mozilla.org/show_bug.cgi?id=1584178)
Bug Reporter: @mathjazz
CC:   @Pike, @guerojeff, @Pike, @lliehu, @jotes

We should prevent certain pieces of text from being changed in the machine translation output in order to prevent translation errors.

For example: Google Translate translates
`fileSizeProgress = ({ $partialSize } of { $totalSize })`
as
`fileSizeProgress = ({$ partialSize} od {$ totalSize})`
(mind the spacing).

Such translations often result in errors, because they mess with the developer variables. Additionally, changes are hard to spot.

We should check if there's a way to annotate substrings that should not be modified in the MT output. Looking at the comments in https://issuetracker.google.com/issues/119256504, we might need to post-process the output instead.

eemeli:
Fixes #2181 

Fluent expressions are wrapped in a `<span translate="no">...</span>` for machinery and pretranslation calls to Google Translate. The wrapping done in the front-end is pretty simplistic, but updating that to e.g. handle plurals right is a whole different job. In the back-end, the wrapping is a bit more sophisticated, as the inserted wrappers carry an `id` parameter that's used in message reconstruction. This allows sanitising the contents that are left within the `notranslate`.

I've tested the front-end parts of this manually (there's rather little there that's properly testable in automation) and played around a bit with the API. @mathjazz, I trust you can take the pretranslate parts for a spin? They pass unit tests, but I've not put them up against the real thing.

It looks like `notranslate` contents do affect the resulting translation somewhat, at least for plurals. I'm also seeing custom elements passing through as would be expected, though spacing does appear to sometimes get "fixed" for elements that aren't recognised as inline HTML. I have not added any escaping for HTML elements because that doesn't seem necessary. There's clearly a proper parser at the other end, so feeding in an incomplete or broken fragment will result in something technically less broken, but therefore structurally different from the input. Given Fluent's own expectations, this should not cause any issues for valid messages.

The PR includes some refactoring, as some Fluent transforms are only used for pretranslation work and really ought to (eventually) get replaced by MF2 equivalents rather than being further promulgated. I also couldn't help myself and added type hints to the new `pretranslation/transformer.py` file.


mathjazz:
Nice work!

I deployed the patch to stage and for now only tested with the [examples](https://github.com/mozilla/pontoon/pull/1870#issuecomment-779764225) from the old bug:

(1)
https://pontoon.mozilla.org/sl/firefox/all-resources/?string=199692
https://mozilla-pontoon-staging.herokuapp.com/sl/firefox/all-resources/?string=199692

(2)
https://pontoon.mozilla.org/fi/thunderbird/calendar/calendar/calendar-uri-redirect-dialog.ftl/?string=209313
https://mozilla-pontoon-staging.herokuapp.com/fi/thunderbird/calendar/calendar/calendar-uri-redirect-dialog.ftl/?string=209313

(3)
https://pontoon.mozilla.org/el/firefox/devtools/client/tooltips.ftl/?string=204020
https://mozilla-pontoon-staging.herokuapp.com/el/firefox/devtools/client/tooltips.ftl/?string=204020

(4)
https://pontoon.mozilla.org/ka/firefox/devtools/client/tooltips.ftl/?string=204020
https://mozilla-pontoon-staging.herokuapp.com/ka/firefox/devtools/client/tooltips.ftl/?string=204020

The good news is that all 4 test cases are now actually fixed in production. 😄

Sadly it seems like the results for the 1st (extra trailing space) and the 2nd (missing trailing quote) example are worse on stage.

mathjazz:
More good news: if I modify the first test case to use the Finnish locale, which (unlike Slovenian) uses the generic engine, I can reproduce the bug on prod, but not on stage. Well done!

https://pontoon.mozilla.org/fi/firefox/all-resources/?string=199692
https://mozilla-pontoon-staging.herokuapp.com/fi/firefox/all-resources/?string=199692

eemeli:
The [missing-end-quote example](https://mozilla-pontoon-staging.herokuapp.com/fi/thunderbird/calendar/calendar/calendar-uri-redirect-dialog.ftl/?string=209313) is weird. If I test that string manually, the trailing quote _is_ included in the translation, but it's clearly getting lost somewhere en route. Will need to look into this some more.

Might also see if adding `notranslate` on all quotes would work, and possibly keep them curly.

mathjazz:
Adding links to the zh-TW errors that Flod called out in #2691:

https://pontoon.mozilla.org/zh-TW/mozillaorg/all-resources/?time=202304021244-202304021344&string=215305
https://mozilla-pontoon-staging.herokuapp.com/zh-TW/mozillaorg/all-resources/?time=202304021244-202304021344&string=215305

Seems like curly braces are added.

@flodolo Do you have other strings to test this patch with?

eemeli:
There was a silly bug in the generic Google Translate calls, as it wasn't setting the format right.

I added locale-aware quote fixing. It's based on data from CLDR, via the `cldr-misc-full` npm package, and uses straight `"quotes"` as an ultimate fallback, though it does try to match the right locale. This requires an additional build step to repackage the data in a more useful format, and is currently only applied to Fluent strings that are complex enough to need to use the HTML format.

@mathjazz Could deploy this to staging?

mathjazz:
Deploy to stage in progress. Should be done in 5-10 minutes.

mathjazz:
https://github.com/mozilla/pontoon/pull/2837 is a continuation of this, let's continue there.

PR #2835, opened 2023-05-05
Move contact person @mention delay to comment submission

### LINKED ISSUE #2591: Delay before contact person data appears in the comment field
Since #2588, there is now a short delay after clicking "REQUEST CONTEXT or REPORT ISSUE" before the contact person's data is entered into the edit field due to the mention data loading in the background.

eemeli:
Fixes #2591 

We don't actually need the `/get-users/` data for the rendering of the `@mention`, but we do ultimately need the URL for comment submission. So let's just insert the mention right away, but block submission (with a `cursor: wait` indicator when hovering on the button) until we have the data.

This still doesn't fix the bad API, but it hides its effects further.

### DIFF
--- a/translate/src/modules/comments/components/AddComment.css
+++ b/translate/src/modules/comments/components/AddComment.css
@@ -77,7 +77,11 @@
   padding-left: 10px;
 }
 
-.comments-list .add-comment .submit-button:hover {
+.comments-list .add-comment .submit-button:hover:not(:disabled) {
   color: #272a2f;
   background: #7bc876;
+}
+
+.comments-list .add-comment .submit-button:disabled {
+  cursor: wait;
 }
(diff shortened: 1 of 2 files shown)


eemeli:
> PM mentions no longer render as links for me: https://mozilla-pontoon-staging.herokuapp.com/sl/common-voice/all-resources/?string=179345

Drat, I hadn't accounted for the `mention` object being apparently frozen. Should be fixed now.

> Also, it would be nice if we added space after mentions. Should I file an issue for that?

The code actually attempts to do that already; I'm not sure why that isn't happening. Yes, this ought to have a new issue.

mathjazz:
> The code actually attempts to do that already; I'm not sure why that isn't happening. Yes, this ought to have a new issue.

https://github.com/mozilla/pontoon/issues/2840

mathjazz (approved):

PR #2838, opened 2023-05-05
Speed up project notifications tab

mathjazz:
The form for sending manual notifications to project contributors can get slow quickly due to the way past notifications are retrieved.

This patch fixes that by reducing the number of DB calls.

### DIFF
--- a/pontoon/projects/views.py
+++ b/pontoon/projects/views.py
@@ -149,7 +149,9 @@ def ajax_notifications(request, slug):
     project = get_object_or_404(
         Project.objects.visible_for(request.user).available(), slug=slug
     )
-    available_locales = project.locales.order_by("name")
+    available_locales = project.locales.prefetch_project_locale(project).order_by(
+        "name"
+    )
 
     # Send notifications
     if request.method == "POST":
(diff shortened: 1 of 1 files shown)


pontoon/projects/views.py
```diff
@@ -149,7 +149,9 @@ def ajax_notifications(request, slug):
     project = get_object_or_404(
         Project.objects.visible_for(request.user).available(), slug=slug
     )
-    available_locales = project.locales.order_by("name")
+    available_locales = project.locales.prefetch_project_locale(project).order_by(
+        "name"
+    )
```
mathjazz, line 154: Brings moderate performance gains by significantly replacing the number of DB calls - one for each Locale - with a Prefetch command.

pontoon/projects/views.py
```diff
@@ -186,22 +188,18 @@ def ajax_notifications(request, slug):
...
+    for notification in Notification.objects.filter(
+        description__isnull=False,
+        target_content_type=ContentType.objects.get_for_model(project),
+        target_object_id=project.id,
+    ):
+        identifier = notification.data["identifier"]
+        if identifier not in notifications_map:
+            notifications_map[identifier] = notification
+
+    notifications = list(notifications_map.values())
```
mathjazz, line 202: Brings big performance gains by dropping slow DB calls - one for each past notification.

eemeli: It would be nice if there were a way for Django to generate a `DISTINCT ON (data->'identifier')` select query, but it looks like the postgres type for that field is `text` rather than `jsonb`.

eemeli (approved):

pontoon/projects/views.py
mathjazz: Yeah, as mentioned in line 190, `django-notifications` is quite out of date and we should write our own notifications system. This field is not even using the Django's [JSONField](https://docs.djangoproject.com/en/4.2/ref/models/fields/#django.db.models.JSONField).

PR #2846, opened 2023-05-10
Also pretranslate strings that only have rejected suggestions submitted by humans

### LINKED ISSUE #2813: Refine pretranslation logic for strings with rejected suggestions
Currently, a string with suggestions is ignored by pretranslation.

We should:
* Pretranslate if all suggestions are submitted by users and rejected. 
* Not pretranslate if a pretranslation was already rejected.

mathjazz:
Fix #2813.

Before this patch, we skipped pretranslation for strings that have had any kind of translation.

### DIFF
--- a/pontoon/pretranslation/tests/test_tasks.py
+++ b/pontoon/pretranslation/tests/test_tasks.py
@@ -8,6 +8,7 @@ from pontoon.test.factories import (
     EntityFactory,
     ResourceFactory,
     TranslatedResourceFactory,
+    TranslationFactory,
     ProjectLocaleFactory,
 )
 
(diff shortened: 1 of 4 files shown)


pontoon/pretranslation/tests/test_tasks.py
```diff
@@ -61,3 +62,51 @@ def test_pretranslate(gt_mock, project_a, locale_a, resource_a, locale_b):
...
     # latest_translation belongs to pretranslations.
     assert project_a.latest_translation in translations
+
+
+@patch("pontoon.pretranslation.tasks.get_pretranslations")
+@pytest.mark.django_db
+def test_which_strings_to_pretranslate(gt_mock, project_a, locale_a, resource_a):
+    """
+    Verify that we only pretranslate:
+    - strings wihtout any translations
```
eemeli, line 72:
```suggestion
    - strings without any translations
```

PR #2837, opened 2023-05-05
Use HTML format for Google Translate calls of Fluent messages

eemeli:
Fixes #2181 
Fixes #2804 

Fluent expressions are wrapped in a `<span translate="no">...</span>` for machinery and pretranslation calls to Google Translate. The wrapping done in the front-end is pretty simplistic, but updating that to e.g. handle plurals right is a whole different job. In the back-end, the wrapping is a bit more sophisticated, as the inserted wrappers carry an `id` parameter that's used in message reconstruction. This allows sanitising the contents that are left within the `translate=no`.

There are now three interconnected transformers working on messages for pretranslation:
- `NotranslateWrapper` replaces inline expressions with `<span id="pt-{idx}" translate="no">{clean}</span>` text elements. Within plural selectors, for references to the plural selector variable, `{clean}` is an example number for the current numerical key or plural category. This uses CLDR data, prepackaged by my `make-plural` npm package. Otherwise, `{clean}` is the serialization of the source, stripped of control characters `{}<>`. The actual serialization is stored in a list for later retrieval according to `{idx}`.
- `PreparePretranslation` flattens select expressions and adjusts the plural selectors for the current locale.
- `ApplyPretranslation` is the public interface, internally calling the other two. Ultimately, it calls `callback(raw, wrapped, locale) -> (translation, service)` for each pattern, and replaces the contents as appropriate.

It looks like `translate=no` contents do affect the resulting translation somewhat, at least for plurals. Hence the replacement with appropriate numbers for plural cases.

I have not added any escaping for HTML or custom elements because that doesn't seem necessary. There's clearly a proper parser at the other end, so feeding in an incomplete or broken fragment will result in something technically less broken, but therefore structurally different from the input. Given Fluent's own expectations, this should not cause any issues for valid messages.

Extra spacing does appear to sometimes get added around elements. To deal with that, spaces are fixed for trailing punctuation in a locale-appropriate manner.

I added locale-aware quote fixing. It's based on data from CLDR, via the `cldr-misc-full` npm package, and uses straight `"quotes"` as an ultimate fallback, though it does try to match the right locale. This requires an additional build step to repackage the data in a more useful format.

The PR includes some refactoring, as some Fluent transforms are only used for pretranslation work and really ought to (eventually) get replaced by MF2 equivalents rather than being further promulgated. I also couldn't help myself and added type hints to the added files.

[...]

PR #2847, opened 2023-05-11
Add ability to opt-out of access key localization

### LINKED ISSUE #2717: Add ability for locales to opt out of access key localization
Fluent Rich editor has a special UI for messages with access keys, which lists access key candidates.

Some locales don't translate access keys, mostly because they use non-Latin scripts.

We should add the ability for Locales to opt out of access key localization.

For those, we should use the original string as an access key candidate or even prefill the field with its value.

mathjazz:
Fix #2717.

This patch adds the ability for locales to opt-out of access key localization. For such locales, pretranslation keeps the accesskeys unchanged.


pontoon/pretranslation/pretranslate.py
```diff
@@ -27,6 +27,16 @@ def __init__(self, locale):
         self.services = []
         self.locale = locale
 
+    def visit_Attribute(self, node):
+        is_accesskey = (
+            node.id.name.endswith("accesskey")
+            and isinstance(node.value.elements[0], ast.TextElement)
+            and len(node.value.elements[0].value) <= 1
```
mathjazz, line 34: This part of the check is taken from the frontend:
https://github.com/mathjazz/pontoon/blob/2717-accesskey-opt-out/translate/src/modules/translationform/components/TranslationForm.tsx#L49

eemeli (changes requested):
In principle this looks fine, but we need to figure out what's happening with #2837 until we land further changes in `pontoon/pretranslation/`.

pontoon/pretranslation/transformer.py
```diff
@@ -135,6 +135,16 @@ def __init__(
         self.locale = locale
         self.services: list[str] = []
 
+    def visit_Attribute(self, node):
+        is_accesskey = (
+            node.id.name.endswith("accesskey")
+            and isinstance(node.value.elements[0], FTL.TextElement)
+            and len(node.value.elements[0].value) <= 1
```
eemeli, line 142: This condition is either insufficient or too strict; we've plenty of accesskeys that are selectors on `PLATFORM()`. For example: https://pontoon.mozilla.org/en-GB/firefox/browser/browser/menubar.ftl/?string=220792

Given that this is gated on `accesskey_localization` which defaults to `True`, I don't think we need to look at the `.accesskey` value if the flag is set to `False`.

pontoon/pretranslation/transformer.py
```diff
@@ -96,8 +96,8 @@ def create_locale_plural_variants(node: FTL.SelectExpression, locale: Locale):
 class PreparePretranslation(Transformer):
     """
     Flattens the given Pattern, uplifting selectors to the highest possible level and
-    duplicating shared parts in the variants. All other Placeables are serialised as
-    TextElements.
+    duplicating shared parts in the variants. Transforms plural variants to match the
+    locale. All other Placeables are serialised as TextElements.
```
eemeli, line 100:
```suggestion
    locale.
```

Fixing my own mistake here. Nothing happens to the placeables in this transform.

eemeli (changes requested):

eemeli (approved):

PR #2854, opened 2023-05-17
Hide "Make Suggestions" for Non-Translators

### LINKED ISSUE #2848: Make suggestions setting should be hidden in editor for contributors
For contributors, the `Make suggestions` setting is hidden in the profile settings, but it's displayed (and can be toggled) in the editor menu. As a consequence, it will say that  `Make suggestions is disabled` when it's not.

We should probably hide this menu (for consistency) or, alternatively, disable it.

ayanaar:
Fixes #2848 

This pull request addresses issue #2848, where the "Make Suggestions" setting was visible in the editor menu for contributors, even though it was hidden in their profile settings. 

With this change, the "Make Suggestions" button will only be rendered if the user has translator permissions. If the user does not have translator permissions, the button will not appear.

This ensures a more consistent and less confusing user experience, as the visibility of the "Make Suggestions" setting now correctly reflects the user's permissions.

### DIFF
--- a/translate/src/modules/editor/components/EditorSettings.test.js
+++ b/translate/src/modules/editor/components/EditorSettings.test.js
@@ -2,7 +2,6 @@ import React from 'react';
 import { shallow } from 'enzyme';
 import sinon from 'sinon';
 
-import { USER } from '~/modules/user/reducer';
 import { EditorSettings, EditorSettingsDialog } from './EditorSettings';
 
 jest.mock('react', () => ({
--- a/translate/src/modules/editor/components/EditorSettings.test.js
+++ b/translate/src/modules/editor/components/EditorSettings.test.js
@@ -2,7 +2,30 @@ import React from 'react';
 import { shallow } from 'enzyme';
 import sinon from 'sinon';
 
+import { USER } from '~/modules/user/reducer';
 import { EditorSettings, EditorSettingsDialog } from './EditorSettings';
+
+jest.mock('react', () => ({
+  ...jest.requireActual('react'),
+  useContext: () => ({ code: 'en' }),
+}));
+
+jest.mock('~/modules/project/hooks', () => ({
+  useProject: () => ({ slug: 'test' }),
+}));
+
+jest.mock('react-redux', () => ({
+  ...jest.requireActual('react-redux'),
+  useSelector: (selector) =>
+    selector({
+      user: {
+        isAuthenticated: true,
+        managerForLocales: ['en'],
+        translatorForLocales: ['en'],
+        translatorForProjects: { 'en-test': true },
+      },
+    }),
+}));
 
 function createEditorSettingsDialog() {
   const toggleSettingMock = sinon.stub();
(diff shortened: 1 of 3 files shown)


translate/src/modules/editor/components/EditorSettings.test.js
```diff
@@ -4,6 +4,28 @@ import sinon from 'sinon';
...
+  ...jest.requireActual('react'),
+  useContext: () => ({ code: 'en' }),
+}));
+
+jest.mock('~/modules/project/hooks', () => ({
+  useProject: () => ({ slug: 'test' }),
+}));
+
+jest.mock('react-redux', () => ({
+  ...jest.requireActual('react-redux'),
```
mathjazz, line 17: You can leave this line one.

translate/src/modules/editor/components/EditorSettings.test.js
```diff
@@ -4,6 +4,28 @@ import sinon from 'sinon';
...
+  useSelector: (selector) =>
+    selector({
+      user: {
+        isAuthenticated: true,
+        managerForLocales: ['en'],
+        translatorForLocales: ['en'],
+        translatorForProjects: { 'en-test': true },
```
mathjazz, line 24: The last two lines are not technically necessary, because if the user is a manager, it's also a translator.

translate/src/modules/editor/components/EditorSettings.test.js
```diff
@@ -19,6 +41,10 @@ function createEditorSettingsDialog() {
 }
 
 describe('<EditorSettingsDialog>', () => {
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
```
mathjazz, line 47: We don't need this block.

translate/src/modules/editor/components/EditorSettings.test.js
```diff
@@ -53,6 +79,9 @@ describe('<EditorSettingsDialog>', () => {
 });
 
 describe('<EditorSettings>', () => {
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
```
mathjazz, line 84: Same as above: this block can be removed.

mathjazz (commented):
Nice work!

It would be great to add a test case to check the new behavior. So if the user is not a translator, we should check that the Make suggestions entry is not visible.

translate/src/modules/editor/components/EditorSettings.tsx
```diff
@@ -43,21 +45,23 @@ export function EditorSettingsDialog({
...
-        </li>
-      </Localized>
+          <li
+            className={
+              'check-box' + (settings.forceSuggestions ? ' enabled' : '')
+            }
+            title='Save suggestions instead of translations'
+            onClick={() => toggleSetting('forceSuggestions')}
+          >
+            {'<glyph></glyph>Make Suggestions'}
```
mathjazz, line 61: While we're in here, we should also fix the capitalization to match the /settings page (also in the .ftl files):
https://github.com/mozilla/pontoon/commit/d8696c9deb54835272b6219ab97fd8a9c9f70802#diff-a8bb92e441421dd7d5626b58519e1827a26d8635c8884cfd17dd0bcd4303f68fR120

eemeli (approved):
Looks fine. The amount of custom mocking that's used the tests here makes them pretty brittle; I find it's often clearer & easier to use e.g. `<Locale.Provider>` rather than mocking its context.

In particular for new tests, we should also be preferring `@testing-library/react` rather than `enzyme`. Support for this was added in #2800. Eventually we'll need to port all the tests to allow us to move on from React v16.

PR #2859, opened 2023-05-25
Fix localization in HistoryTranslation

eemeli:
The Fluent references were broken in #2803 (can't format just an attribute), and we should prefer declarative rather than imperative forms.

@ayanaar Feel free to pass this review on to @mathjazz if it's too far outside your comfort zone.


eemeli:
Bah, forgot to update the tests...

ayanaar (approved):
Tested the modifications locally and everything works as expected!

mathjazz:
@eemeli Could you please rebase?

Issue #2726, opened 2023-01-23
New string notifications should always link to translate view


mathjazz:
If the user hasn't set their custom [homepage](https://mozilla-l10n.github.io/localizer-documentation/tools/pontoon/users.html#locale-settings) (2/3 of the PMO user base), [new string notifications](https://mozilla-l10n.github.io/localizer-documentation/tools/pontoon/notifications.html#new-strings) link to project dashboards, rather than to the translate view.

We should always link to the [locale-agnostic translate view](https://github.com/mozilla/pontoon/blob/master/pontoon/base/views.py#L57), which only links to the dashboard in case the custom homepage isn't set and the user locale cannot be determined using the `Accept-Language` header.

See also:
https://github.com/mozilla/pontoon/pull/1989/files/b404f21cfe312f2020a33265706aad8a56fd00a5#diff-4ebc5e9888b2962aecd470dd76f7d410b4e246f2d8323df57ed7dd0c8d5e3ade

PR #2860, opened 2023-05-25
Correct Redirection for New String Notifications

ayanaar:
Addresses #2726 

Previously, the redirection was dependent on whether the 'new string' was present in the user's custom homepage. This check was implemented in both the template and the models. However, this was found to be redundant and led to incorrect redirections because the [translate_locale_agnostic](https://github.com/mozilla/pontoon/blob/master/pontoon/base/views.py#L57) function already handles the redirection based on the user's custom homepage.

This change simplifies the code and ensures that users are correctly redirected when they click on a notification about a new string.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -351,7 +351,7 @@ def serialized_notifications(self):
         is_comment = False
 
         if hasattr(notification.actor, "slug"):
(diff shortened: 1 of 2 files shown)


mathjazz (approved):
Nicely done! Works very well!

PR #2861, opened 2023-05-27
Update packaged Open Sans font to use WOFF2 format

eemeli:
This gets rid of the `kern: Too large subtable` console errors and reduces the file size of the font files by about 50kB each. All our supported browsers [support WOFF2](https://caniuse.com/woff2), so we don't need to provide other formats.

The fonts were downloaded as TTF from [fonts.google.com](https://fonts.google.com/specimen/Open+Sans) and converted to WOFF2 using [google/woff2](https://github.com/google/woff2).

I have not investigated if this version of the font could also be used for Serbian and Macedonian Cyrillic (#571).


mathjazz:
Excellent work!

I'm getting the following warning when using a Cyrillic translation in the sidebar (e.g. Просечно задовољство).

```
downloadable font: glyf: Number of composite points in glyph 1265 exceeds maxp maxCompositePoints:
183 vs 168, adjusting limit. (font-family: "Ubuntu Regular" style:italic weight:400 stretch:100 src index:0)
source: http://localhost:8000/static/fonts/Ubuntu-RI.ttf 
```

However, it seems like we no longer need the exception introduced in #571 and can use Open Sans everywhere.

mathjazz (approved):
Good news: this warning is there also on master.

eemeli:
Scope creep: Dropped the included Ubuntu Regular font, and the exception that used it for Serbian & Macedonian, and dropped the non-WOFF2 Font Awesome files.

mathjazz (approved):

PR #2863, opened 2023-05-27
Speed up server stop in development

eemeli:
I got annoyed always waiting for the 10 second timeout when stopping the dev server. Two things were going wrong, both fixed here:
1. Docker was sending the process its default `SIGTERM` signal, but Django expects `SIGINT`.
2. The `server_run.sh` script was not replacing its process with Django, and not passing on to it any signals.


mathjazz (approved):
Nice catch!

PR #2864, opened 2023-05-27
Drop `zoom: 1` from CSS clearfix

eemeli:
We don't need to accommodate ancient browsers, esp. as these cause the following console warning:

> This page uses the non standard property “zoom”. Consider using calc() in the relevant property values, or using “transform” along with “transform-origin: 0 0”.

### DIFF
--- a/pontoon/base/static/css/boilerplate.css
+++ b/pontoon/base/static/css/boilerplate.css
@@ -209,5 +209,3 @@ button {  width: auto; overflow: visible; }
 }
 
 .clearfix:after { clear: both; }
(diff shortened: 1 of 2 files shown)


mathjazz:
This code is part of a [library](https://github.com/h5bp/html5-boilerplate). We should update it to the latest version, which doesn't use the `zoom` property anymore.

eemeli:
That sounds like a much bigger change that would require more testing than I've bandwidth for atm. Would you like me to close this PR and file an issue instead?

mathjazz (approved):
Let's file a (good first?) issue and merge.

PR #2865, opened 2023-05-27
Update dependencies

eemeli:
Refresh `package-lock.json`, and apply updates to `typescript` and Fluent packages.

I hit an odd Rollup bundling issue that was easiest solved by updating to newer versions.

The huge diff in `package-lock.json` is due to switching from lockfileVersion 2 to 3, which no longer needs to duplicate all its contents.


eemeli:
Turns out that this didn't fix my bundling issue (a new dependency redeclaring the global var `top`) as that needed us to set `output.format` to `'iife'` in our Rollup config, as we've configured django-pipeline not to do that and it can't be configured to set `<script type=module>`.

But this remains a good idea.

mathjazz (approved):
LGTM

PR #2866, opened 2023-05-28
Replace editor `<textarea>` with CodeMirror v6

### LINKED ISSUE #2180: Improve Fluent source editor with syntax highlighting, error annotations
## [Bug 1581911](https://bugzilla.mozilla.org/show_bug.cgi?id=1581911)
Bug Reporter: @mathjazz
CC:   @stasm, v+mozbug@g.nevcal.com

As part of fixing bug 1563457 we replaced Ace Editor with `<textarea>`  element in order to [support all scripts properly](https://github.com/ajaxorg/ace/issues/460). 

While that fixed the bug, it also removed syntax highlighting and error annotations from Fluent source editor.

We should find a solution similar to Ace Editor, that will support:

1. All scripts.
1. Fluent syntax highlighting.
1. Fluent error annotations.
1. (If possible) browser's spell- and grammar checkers (bug 1556355).

### LINKED ISSUE #2781: Clicking a placeable in source string moves the cursor to the end of the string
Find a string with placeables, click to add it in the middle of the translation. The cursor then jumps at the end of the translation, which I believe is an unwanted regression.

eemeli:
Fixes #2180 
Fixes #2781 

Replaces the translation editor `<textarea>` using [CodeMirror](https://codemirror.net/), an editor we'll be able to extend to provide e.g. syntax and error highlighting as well as autocompletion.

As CodeMirror isn't React, but does manage its internal state in a similar manner, matching that within the app requires some finessing. A new context `EditorResult` is added to `EditorProvider` as an output; it's updated by the editor, but not read by it. The pre-existing `EditorData` now no longer directly contains the value being edited, but does have references to `EditFieldHandle`s, an interface that abstracts the get/set/focus operations for each field.


eemeli:
Rebased to not include other PR contents. Got tests to pass, though needed to skip a couple due to the difference in selection handling. Will need to work on that to help ensure that we also fix #2781.

eemeli:
Key shortcuts and default selection should now work. This now fixes #2781.

translate/rollup.config.mjs
```diff
@@ -10,7 +10,7 @@ import css from 'rollup-plugin-css-only';
 /** @type {import('rollup').RollupOptions} */
 const config = {
   input: 'src/index.tsx',
-  output: { file: 'dist/translate.js' },
+  output: { file: 'dist/translate.js', format: 'iife' },
```
eemeli, line 13: One of the new dependencies includes a module with a top-level statement `const top = ...`. With the default `output.format`, Rollup presumes that the code will be included in a `<script type=module>` so leaves it at the top level. However, as we don't do that, the output needs to be wrapped in an IIFE.

translate/src/context/Editor.tsx
```diff
@@ -148,71 +115,111 @@ export type EditorActions = {
...
-  const { exist } = useContext(UnsavedChanges);
-  const { resetFailedChecks } = useContext(FailedChecksData);
+  const { resetFailedChecks, source: failedChecksSource } =
+    useContext(FailedChecksData);
 
-  const [state, setState] = useState<EditorData>(initEditorData);
+  const [state, setState] = useState(initEditorData);
+  const [result, setResult] = useState<EditorResult>([]);
 
   const actions = useMemo<EditorActions>(() => {
```
eemeli, line 190: The actions that directly set a new value for `fields` need to also update `result`. Those that call `field.handle.current.setValue()` don't, as the EditField and EditAccessKey will take care of that.

translate/src/modules/editor/hooks/useHandleShortcuts.ts
eemeli, line 1: Contents refactored & moved to `translate/src/modules/translationform/utils/editFieldShortcuts.ts`.

translate/src/modules/editor/components/Editor.css
eemeli, line 1: `textarea` styles refactored & moved to `translate/src/modules/translationform/components/TranslationForm.css`

translate/src/modules/translationform/components/EditAccesskey.tsx
```diff
@@ -1,68 +1,145 @@
...
-  inputRef,
-  name,
-  onFocus,
-  userInput,
-  value,
-}: EditFieldProps & { name: string }) {
-  const locale = useContext(Locale);
-  const { setEditorFromInput } = useContext(EditorActions);
-  const message = useEditorValue();
+function useHandleShortcuts(): (event: React.KeyboardEvent) => void {
```
eemeli, line 21: CodeMirror has its own way of specifying key handlers, so this is now specific to EditAccesskey.

As a follow-up change, we should probably move these handlers out of the editors, so that they're applied also when the focus isn't on the input field, rather like `Alt/Option+ArrowDown/ArrowUp`.

translate/src/modules/translationform/components/TranslationForm.css
```diff
@@ -81,6 +74,46 @@
...
-.translationform .accesskeys .key:hover {
+.translationform .accesskeys .key:hover:not(:disabled) {
   border-color: #7bc876;
 }
+
+.translationform .accesskeys .key:disabled {
+  cursor: default;
+}
+
+.cm-editor {
```
eemeli, line 85: CodeMirror packages in some styling for itself via CSS-in-JS; these are overrides on those defaults. It does include its own [styling system](https://codemirror.net/examples/styling/), but relying on CSS is just as valid, and more in line with what we've been doing so far.

translate/src/modules/translationform/components/TranslationForm.tsx
```diff
@@ -1,27 +1,31 @@
...
   getExample,
-  htmlFor,
   labels,
 }: {
   getExample: (label: string) => number | undefined;
-  htmlFor: string;
   labels: Array<{ label: string; plural: boolean }>;
 }) => (
-  <label htmlFor={htmlFor}>
+  <label>
```
eemeli, line 28: Sadly, HTML `<label> and `contentEditable` doesn't really play well together, so clicking on the label will no longer focus the appropriate field.

mathjazz: That's a nasty UX issue. Why do we need `contenteditable`?

mathjazz:
> Fixes #2180

This PR doesn't fix the whole issue. We should either file a separate issue to track the implementation of the rich editor, or a separate issues to track syntax highlighting and error annotations.

translate/src/modules/translationform/components/TranslationForm.tsx
eemeli: Because it's used by CodeMirror, and effectively the basis on which any solution other than `<textarea>` is built.

eemeli:
> > Fixes #2180
> 
> This PR doesn't fix the whole issue. We should either file a separate issue to track the implementation of the rich editor, or separate issues to track syntax highlighting and error annotations.

Point taken. Removed the closing link.

translate/src/modules/translationform/components/TranslationForm.tsx
mathjazz: Oh, ok - I thought we were also using `contenteditable` in the label.

translate/src/modules/translationform/components/TranslationForm.css
```diff
@@ -81,6 +74,46 @@
...
+  background: white;
+  color: #444;
+  font-size: 14px;
+  line-height: 22px;
```
mathjazz, line 89: This setting doesn't have any effect. It needs to be moved to `.cm-scroller`.

mathjazz (commented):
Works very well, excellent job!

The only thing we should fix in addition to the CSS comment inline is add a short delay before we trigger trailing space indicator. Currently it's annoying to see the red color blinking as you type.

translate/src/modules/translationform/components/TranslationForm.css
```diff
@@ -81,6 +74,46 @@
...
+  padding: 6px 8px 6px 4px;
+}
+
+.cm-editor .cm-scroller {
+  font-family: inherit;
```
eemeli, line 94:
```suggestion
  padding: 6px 8px 6px 4px;
}

.cm-editor .cm-scroller {
  font-family: inherit;
  line-height: 22px;
```

translate/src/modules/translationform/utils/editFieldExtensions.ts
```diff
@@ -0,0 +1,90 @@
...
+  useEffect(() => {
+    ref.current = handlers;
+  });
+  return ref;
+}
+
+export const getExtensions = (
+  ref: ReturnType<typeof useKeyHandlers>,
+): Extension[] => [
+  highlightTrailingWhitespace(),
```
eemeli, line 60:
```suggestion
```

eemeli:
> The only thing we should fix in addition to the CSS comment inline is add a short delay before we trigger trailing space indicator. Currently it's annoying to see the red color blinking as you type.

The default extension for this isn't configurable like that, so I removed it for now.

mathjazz (approved):

PR #2868, opened 2023-05-31
Scroll entire main column all together

eemeli:
Currently, the translate view main column included separately scrollable sections for the metadata and for history. This approach mostly works well, but has a couple of failure modes:

- When there's a lot of metadata, it overflows. Often, it's not obvious that this panel contains more content.
- When the screen height is limited, the history may become completely hidden inaccessible.

To improve on this, I'm here proposing to make the whole column scrollable as one, allowing each of its parts to always be shown. This fixes both of the issues mentioned above and at least to me feels less surprising than what we've had so far.

One small negative of the new approach is that when switching between messages, the vertical alignment of the translation form may now vary, as the metadata section's height is no longer constant. On the other hand, all of the metadata is now shown.

### DIFF
--- a/translate/public/locale/en-US/translate.ftl
+++ b/translate/public/locale/en-US/translate.ftl
@@ -276,9 +276,6 @@ entitydetails-Metadata--context =
 entitydetails-Metadata--placeholder =
     .title = PLACEHOLDER EXAMPLES
 
-entitydetails-Metadata--resource =
-    .title = RESOURCE
-
 entitydetails-Metadata--attribute =
     .title = FLUENT ATTRIBUTE
 
--- a/translate/public/locale/fr/translate.ftl
+++ b/translate/public/locale/fr/translate.ftl
@@ -111,9 +111,6 @@ entitydetails-Metadata--context =
 entitydetails-Metadata--placeholder =
     .title = EXEMPLES BOUCHE-TROUS
 
-entitydetails-Metadata--resource =
-    .title = RESSOURCE
-
 
 ## Entity Details PluralString
 ## Shows the original string of an entity with plural forms
--- a/translate/src/modules/entitydetails/components/Metadata.css
+++ b/translate/src/modules/entitydetails/components/Metadata.css
@@ -3,7 +3,7 @@
   color: #aaa;
   font-size: 12px;
   font-style: italic;
-  min-height: 100px;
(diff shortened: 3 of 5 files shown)


mathjazz:
Technically this looks good and is r+.

I am however struggling with the editor jumping up and down as I translate a selection of strings. Yes, sometimes not all metadata is visible, but I don't remember when any of the relevant information I needed was below the fold.

eemeli:
> I am however struggling with the editor jumping up and down as I translate a selection of strings. Yes, sometimes not all metadata is visible, but I don't remember when any of the relevant information I needed was below the fold.

@mathjazz Could you give some examples of metadata that you think we should be fine putting below the fold?

For a couple of other alternatives, we could do one or both of the following:

1. Expand the metadata `min-height`. Right now it's set at 100px; going to 114px would capture a majority of cases that don't fit into that.

2. Squeeze more metadata into fewer vertical pixels. For example, we now always include something like this:<br>
Screenshot 2023-06-16 at 12 01 55<br>
We could in most cases fit that onto a single line:<br><br>Screenshot 2023-06-16 at 12 06 09

mathjazz:
> @mathjazz Could you give some examples of metadata that you think we should be fine putting below the fold?

All the metadata here:
https://pontoon.mozilla.org/sl/amo/LC_MESSAGES/djangojs.po/?string=66617

Project and Resource can be seen from the Main Menu, and the source code reference I never look at.

eemeli:
> Project and Resource can be seen from the Main Menu, and the source code reference I never look at.

Just to clarify, are you asking for the following to take place as a part of this PR?
1. Remove the reference line from the metadata.
2. Leave out `#:` gettext comment lines from the metadata.

mathjazz:
No, no - I'd keep them in. For sure there are cases when they are useful.

mathjazz (approved):
Nice work!

eemeli:
I've applied the following changes as discussed on our call:
- Set a `min-height: 114px` on the metadata. This allows for up to five lines of content without extending its height, and most now use only 3-4 lines.
- Combine the resource line into context, so that it's now in most cases a path `message-id · file · project`. For gettext messages that do not provide a `msgctx`, it's just `file · project`.
- De-emphasize the links on this line, so that they're green only on hover.

PR #2871, opened 2023-06-02
Fix Missing Strings in Downloaded XLIFF Files

### LINKED ISSUE #2239: Revisit sync process for XLIFF files
## [Bug 1676336](https://bugzilla.mozilla.org/show_bug.cgi?id=1676336)
Bug Reporter: @flodolo
CC:   @guerojeff, @mathjazz, sstreich@mozilla.com

So far we've been using XLIFF only for iOS projects. For both projects (Firefox, Focus) we built an infrastructure that updates all locales when adding new strings to the reference file (en-US).

Currently, if you add new strings only to the reference file:
- Strings show up as missing for all locales
- The new translations are not submitted back to the repository, since the XLIFF is missing them
- "Download Translations" gives you a file that is missing translations, like the one in the repository.

The downside of this approach, besides being very confusing, is that we need to replicate the infrastructure for every repository. For example, this is the work happening for the new VPN client
https://github.com/mozilla-l10n/mozilla-vpn-client-l10n/pull/39

Unlike GetText, XLIFF doesn't come with native commands to propagate new strings to all locales. 

I think it would make more sense to be able to add new strings only to the reference locale, and have Pontoon add them to the resulting XLIFF file when serializing it.

ayanaar:
Fixes #2239 

This Pull Request addresses the issue of missing strings in downloaded XLIFF files.

The problem surfaced when a new string was added to the reference file and subsequently translated. Despite the string being present in the Pontoon interface, it was missing in the downloaded XLIFF file.

This PR fixes the issue by ensuring that all new strings and their corresponding translations are correctly parsed and written into the downloaded XLIFF file. Now, when a user adds a new string to the reference file, translates it, and downloads the translated file, they will find the new string and its translation present as expected.

Details of main changes:
* Rewrote the `XLIFFEntity` and `XLIFFResource` classes
* Adjusted the writing logic in the `save` function to ensure all translations are included in the downloaded file
* Modified the parsing logic in the `parse` function to correctly register new strings.

Testing strategy:
* Added new strings to the reference file, translated them and downloaded the XLIFF file to verify the correct behaviour
* Ran existing unit tests to ensure no changes in other functionalities

### DIFF
--- a/pontoon/base/utils.py
+++ b/pontoon/base/utils.py
@@ -295,11 +295,15 @@ def get_download_content(slug, code, part):
             ).filter(locale=locale)
 
         print(resource_file.translations)
-        for key, vcs_translation in resource_file.translations.items():
-            entity_key = vcs_translation.key
-            if entity_key in entities_dict:
-                entity = entities_dict[entity_key]
-                vcs_translation.update_from_db(entity)
+        # for key, vcs_translation in resource_file.translations.items():
+        #     entity_key = vcs_translation.key
+        #     if entity_key in entities_dict:
+        #         entity = entities_dict[entity_key]
+        #         vcs_translation.update_from_db(entity)
+        for vcs_translation in resource_file.translations:
+            key = vcs_translation.key
+            if key in entities_dict:
+                entity = entities_dict[key]
 
 
         resource_file.save(locale)
--- a/pontoon/base/utils.py
+++ b/pontoon/base/utils.py
@@ -298,6 +298,7 @@ def get_download_content(slug, code, part):
             key = vcs_translation.key
             if key in entities_dict:
                 entity = entities_dict[key]
+                vcs_translation.update_from_db(entity)
 
 
         resource_file.save(locale)
--- a/pontoon/base/utils.py
+++ b/pontoon/base/utils.py
@@ -294,12 +294,6 @@ def get_download_content(slug, code, part):
                 Q(approved=True) | Q(pretranslated=True)
             ).filter(locale=locale)
 
-        print(resource_file.translations)
-        # for key, vcs_translation in resource_file.translations.items():
-        #     entity_key = vcs_translation.key
-        #     if entity_key in entities_dict:
-        #         entity = entities_dict[entity_key]
-        #         vcs_translation.update_from_db(entity)
         for vcs_translation in resource_file.translations:
             key = vcs_translation.key
             if key in entities_dict:
--- a/pontoon/base/utils.py
+++ b/pontoon/base/utils.py
@@ -260,7 +260,7 @@ def get_download_content(slug, code, part):
 
         # Get source file if needed
         source_path = None
-        if resource.is_asymmetric:
+        if resource.is_asymmetric or resource.format == "xliff":
             dirnames = VCSProject.SOURCE_DIR_NAMES
             source_path = _download_file(
                 source_prefixes, dirnames, vcs_project, resource.path
--- a/pontoon/base/utils.py
+++ b/pontoon/base/utils.py
@@ -300,7 +300,6 @@ def get_download_content(slug, code, part):
                 entity = entities_dict[key]
                 vcs_translation.update_from_db(entity)
 
-
         resource_file.save(locale)
 
         if not locale_path:
--- a/pontoon/base/utils.py
+++ b/pontoon/base/utils.py
@@ -294,11 +294,13 @@ def get_download_content(slug, code, part):
                 Q(approved=True) | Q(pretranslated=True)
             ).filter(locale=locale)
 
-        for vcs_translation in resource_file.translations:
-            key = vcs_translation.key
-            if key in entities_dict:
-                entity = entities_dict[key]
+        print(resource_file.translations)
+        for key, vcs_translation in resource_file.translations.items():
+            entity_key = vcs_translation.key
+            if entity_key in entities_dict:
+                entity = entities_dict[entity_key]
                 vcs_translation.update_from_db(entity)
+
 
         resource_file.save(locale)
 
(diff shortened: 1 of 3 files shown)


pontoon/sync/formats/xliff.py
```diff
@@ -88,44 +93,98 @@ def sync_changes(self):
...
+        
+        # print(f"Keys in entities_dict: {entities_dict.keys()}")
+
+        # Iterate over each unit in the reference XLIFF file
+        for reference_unit in reference_xliff_file.units:
+            reference_entity = XLIFFEntity(0, reference_unit)
+            #If we have a translation for this entity, add it to the reference entity
+
+            reference_entity_key = reference_entity.key.replace('\x04', '')
+            if reference_entity_key in entities_dict:
```
mathjazz, line 143: This condition will always be True: we are looping over items of `reference_xliff_file.units` and `entities_dict` also contains all items of `reference_xliff_file.units`.

pontoon/sync/formats/xliff.py
```diff
@@ -88,44 +93,98 @@ def sync_changes(self):
...
+            reference_entity = XLIFFEntity(0, reference_unit)
+            #If we have a translation for this entity, add it to the reference entity
+
+            reference_entity_key = reference_entity.key.replace('\x04', '')
+            if reference_entity_key in entities_dict: 
+                entity = entities_dict[reference_entity_key]
+                print(f"Entity strings: {entity.strings}")  # Print entity.strings
+
+                if entity.strings and locale in entity.strings:
+                    reference_entity.target_string = entity.strings[locale]
```
mathjazz, line 148: `entity` is an element of `entities_dict`, which is a derivative of `reference_xliff_file.units`. There are no translations here, source strings only.

pontoon/base/utils.py
```diff
@@ -294,12 +294,14 @@ def get_download_content(slug, code, part):
...
 
-        for vcs_translation in resource_file.translations:
-            key = vcs_translation.key
-            if key in entities_dict:
-                entity = entities_dict[key]
+        print(resource_file.translations)
+        for key, vcs_translation in resource_file.translations.items():
+            entity_key = vcs_translation.key
+            if entity_key in entities_dict:
+                entity = entities_dict[entity_key]
```
mathjazz, line 301: I don't think this change is necessary. Can you try to revert it?

pontoon/sync/formats/xliff.py
```diff
@@ -13,52 +14,30 @@ class XLIFFEntity(VCSTranslation):
...
-        self.order = order
+    def __init__(
+        self,
+        key,
+        source_string,
+        source_string_plural,
+        strings,
+        comments=None,
+        order=None,
+        unit=None
```
mathjazz, line 25: We don't seem to be using `unit` in `XLIFFEntity`, so I'd remove it.

OTOH, we should add `context` parameter, since we also had it before.

pontoon/sync/formats/xliff.py
```diff
@@ -13,52 +14,29 @@ class XLIFFEntity(VCSTranslation):
...
+        context,
+        source_string,
+        source_string_plural,
+        strings,
+        comments=None,
+        order=None,
+    ):
+        super().__init__(
+            key=key,
+            context=key,
```
mathjazz, line 29: This should be `context=context`.

pontoon/sync/formats/xliff.py
```diff
@@ -88,33 +68,123 @@ def sync_changes(self):
...
+                # Parse the xml content of the file into an XLIFF file object
+                self.xliff_file = xliff.xlifffile(xml)
+            except etree.XMLSyntaxError as err:
+                # If there is an error parsing the file, raise a ParseError 
+                raise ParseError(f"Failed to parse {self.source_resource.path}: {err}")
+
+            # Loop through each unit in the XLIFF file
+            for order, unit in enumerate(self.xliff_file.units):
+                 # Apply any changes made to this object to the backing unit in the xliff file.
+                if None in unit.strings:
```
mathjazz, line 160: In `sync_changes()`, `self` refers to the `XLIFFEntity` instance, not `unit`. You need to modify this line to `if None in entity.strings:` and get `entity` from `self.entities`. See [ftl.py](https://github.com/mozilla/pontoon/blob/2ad1cd1e987073bc0d175a75d9b1299550e3afa1/pontoon/sync/formats/ftl.py#L143-L144) for reference. You can get the key from the unit.

pontoon/sync/formats/xliff.py
```diff
@@ -14,107 +15,148 @@ class XLIFFEntity(VCSTranslation):
...
+                key = unit.getid()
+                entity = self.entities.get(key)
+                if None in entity.strings:
+                    # Store updated nodes
+                    xml = unit.xmlelement
+                    target = xml.find(unit.namespaced("target"))
+
+                    # If there's no existing target, create a new one
+                    if target is None:
+                        target = etree.SubElement(xml, unit.namespaced("target"))
```
mathjazz, line 134: We didn't have this logic in the old code and removing it doesn't seem to break when downloading pontoon-test files. I'm curious why do we need it.

The expectation is that the source file always have the `<target>` node.

pontoon/sync/formats/xliff.py
```diff
@@ -14,107 +15,148 @@ class XLIFFEntity(VCSTranslation):
...
+                    unit.settarget(entity.strings[None])
```
mathjazz, line 135: We should set the target at the top of this `if`-block, same as in the old code, in order to keep the `target` value up-to-date.

mathjazz (commented):
Looks great! Just added two minor comments.

pontoon/sync/formats/xliff.py
mathjazz: Sorry, this is actually not the case:
https://github.com/mozilla-l10n/firefoxios-l10n/blob/main/templates/firefox-ios.xliff

The old code didn't need to do this, because we always have the `target` node in the localized files. But here we are working with the source file. 

So let's keep the block in. Apologies, my bad.

ayanaar: No worries!

mathjazz (approved):
Great work! This improvement will greatly simplify the infrastructure requirements for XLIFF projects.

PR #2873, opened 2023-06-07
Use Glossary to prevent placeables modifications by machine translation

### LINKED ISSUE #2181: Google Translate modifies whitespace within placeables, causing errors

mathjazz:
Fixes #2181.

This patch prevents placeables modifications by machine translation. Any placeable found in the input string is stored in the glossary with equivalent translations for all locales supported by Google Translate.

~I [tried to](https://github.com/mozilla/pontoon/pull/2873/commits/8b03a884ca551102e3d92b2643900daca3c2c522) use the glossary only on strings containing errors and warnings, but I [reverted](https://github.com/mozilla/pontoon/pull/2873/commits/1ff9dce6ca4eba7ef4d5b98137c1eb127763b7bd) the commit, because the code looks ugly. A bigger refactor will be needed should we want to proceed with that.~

~Which means, at least for now, any string containing placeables is translated using the glossary. Placeables are detected using the most common parsers taken from `react-content-marker`. I still need to port a few more.~

_**EDIT**: We only use glossary to re-pretranslate strings for which the default (non-glossary) machine translation output resulted in a failed check (error or warning)._

### DIFF
--- a/pontoon/base/placeables.py
+++ b/pontoon/base/placeables.py
@@ -9,7 +9,7 @@ class PythonFormatNamedString:
 
 
 class PythonFormatString:
-    parser = re.compile(r"({{?[\w!.,[\]%:$<>+\-= ]*}?})")
+    parser = re.compile(r"{{|}}|{[\w!.,[\]%:$<>+\-= ]*}")
 
 
 class PythonFormattingVariable:
--- a/pontoon/base/placeables.py
+++ b/pontoon/base/placeables.py
@@ -1,22 +1,23 @@
 import re
 
 
-class PythonFormatNamedString:
+class PythonPrintfString:
     parser = re.compile(
-        r"(%\([[\w\d!.,[\]%:$<>+\-= ]*\)[+-0\d+#]?[.\d+]?[sdefgoxc%])",
-        re.IGNORECASE,
+        r"""
+            %
+            (?:\(.*?\))?  # mapping key
+            [\#0\-\ +]*   # conversion flags
+            [\d*]*        # minimum field width
+            (?:\.[\d*])?  # precision
+            [hlL]?        # length modifier
+            [acdEeFfGgiorsuXx%]  # conversion type
+        """,
+        re.X,
     )
 
 
 class PythonFormatString:
     parser = re.compile(r"{{|}}|{[\w!.,[\]%:$<>+\-= ]*}")
-
-
-class PythonFormattingVariable:
-    parser = re.compile(
-        r"(%(%|(\([^)]+\))?[-+0#]?(\d+|\*)?(\.(\d+|\*))?[hlL]?[diouxXeEfFgGcrs]))"
-    )
-    match_index = 0
 
 
 class FluentTerm:
@@ -34,9 +35,8 @@ class JsonPlaceholder:
 def get_placeables(text):
     """Return a list of placeables found in the given string."""
     parsers = [
-        PythonFormatNamedString.parser,
+        PythonPrintfString.parser,
         PythonFormatString.parser,
-        PythonFormattingVariable.parser,
         FluentTerm.parser,
         FluentFunction.parser,
         JsonPlaceholder.parser,
--- a/pontoon/base/placeables.py
+++ b/pontoon/base/placeables.py
@@ -3,7 +3,7 @@ import re
 
 class PythonFormatNamedString:
     parser = re.compile(
-        r"(%\([[\w\d!.,[\]%:$<>+\-= ]*\)[+|-|0\d+|#]?[.\d+]?[s|d|e|f|g|o|x|c|%])",
+        r"(%\([[\w\d!.,[\]%:$<>+\-= ]*\)[+-0\d+#]?[.\d+]?[sdefgoxc%])",
         re.IGNORECASE,
     )
 
--- a/pontoon/base/placeables.py
+++ b/pontoon/base/placeables.py
@@ -14,7 +14,7 @@ class PythonFormatString:
 
 class PythonFormattingVariable:
     parser = re.compile(
-        r"(%(%|(\([^)]+\)){0,1}[-+0#]{0,1}(\d+|\*){0,1}(\.(\d+|\*)){0,1}[hlL]{0,1}[diouxXeEfFgGcrs]{1}))"
+        r"(%(%|(\([^)]+\))?[-+0#]?(\d+|\*)?(\.(\d+|\*))?[hlL]?[diouxXeEfFgGcrs]))"
     )
     match_index = 0
 
--- a/pontoon/base/placeables.py
+++ b/pontoon/base/placeables.py
@@ -2,7 +2,10 @@ import re
 
 
 class PythonFormatNamedString:
-    parser = r"(%\([[\w\d!.,[\]%:$<>+\-= ]*\)[+|-|0\d+|#]?[.\d+]?[s|d|e|f|g|o|x|c|%])"
+    parser = re.compile(
+        r"(%\([[\w\d!.,[\]%:$<>+\-= ]*\)[+|-|0\d+|#]?[.\d+]?[s|d|e|f|g|o|x|c|%])",
+        re.IGNORECASE,
+    )
 
 
 class PythonFormatString:
--- a/pontoon/base/placeables.py
+++ b/pontoon/base/placeables.py
@@ -31,9 +31,9 @@ def get_placeables(text):
     parsers = [
         PythonFormatNamedString.parser,
         PythonFormatString.parser,
-        # PythonFormattingVariable.parser,
+        PythonFormattingVariable.parser,
         FluentTerm.parser,
-        # FluentParametrizedTerm.parser,
+        FluentParametrizedTerm.parser,
         FluentFunction.parser,
     ]
 
--- a/pontoon/base/placeables.py
+++ b/pontoon/base/placeables.py
@@ -1,0 +1,45 @@
+import re
+
+
+class PythonFormatNamedString:
+    parser = r"(%\([[\w\d!.,[\]%:$<>+\-= ]*\)[+|-|0\d+|#]?[.\d+]?[s|d|e|f|g|o|x|c|%])"
+
+
+class PythonFormatString:
+    parser = r"(\{{?[\w\d!.,[\]%:$<>+-= ]*\}?})"
+
+
+class PythonFormattingVariable:
+    parser = r"(%(%|(\([^)]+\)){0,1}[-+0#]{0,1}(\d+|\*){0,1}(\.(\d+|\*)){0,1}[hlL]{0,1}[diouxXeEfFgGcrs]{1}))"
+    match_index = 0
+
+
+class FluentTerm:
+    parser = r"({ ?-[^}]* ?})"
+
+
+class FluentParametrizedTerm:
+    parser = r"({ ?-[^}]*([^}]*: ?[^}]*) ?})"
+
+
+class FluentFunction:
+    parser = r"({ ?[A-W0-9\-_]+[^}]* ?})"
+
+
+def get_placeables(text):
+    """Return a list of placeables found in the given string."""
+    parsers = [
+        PythonFormatNamedString.parser,
+        PythonFormatString.parser,
+        # PythonFormattingVariable.parser,
+        FluentTerm.parser,
+        # FluentParametrizedTerm.parser,
+        FluentFunction.parser,
+    ]
+
+    placeables = []
+
+    for parser in parsers:
+        placeables += re.findall(parser, text)
+
+    return list(dict.fromkeys(placeables))
--- a/pontoon/base/placeables.py
+++ b/pontoon/base/placeables.py
@@ -9,7 +9,7 @@ class PythonFormatNamedString:
 
 
 class PythonFormatString:
-    parser = re.compile(r"(\{{?[\w\d!.,[\]%:$<>+-= ]*\}?})")
+    parser = re.compile(r"({{?[\w!.,[\]%:$<>+\-= ]*}?})")
 
 
 class PythonFormattingVariable:
--- a/pontoon/base/placeables.py
+++ b/pontoon/base/placeables.py
@@ -1,6 +1,10 @@
 import re
 
 
+class MozillaPrintfString:
+    parser = re.compile(r"%(?:[1-9]\$)?[S@]")
+
+
 class PythonPrintfString:
     parser = re.compile(
         r"""
@@ -35,6 +39,7 @@ class JsonPlaceholder:
 def get_placeables(text):
     """Return a list of placeables found in the given string."""
     parsers = [
+        MozillaPrintfString.parser,
         PythonPrintfString.parser,
         PythonFormatString.parser,
         FluentTerm.parser,
--- a/pontoon/base/placeables.py
+++ b/pontoon/base/placeables.py
@@ -9,24 +9,26 @@ class PythonFormatNamedString:
 
 
 class PythonFormatString:
-    parser = r"(\{{?[\w\d!.,[\]%:$<>+-= ]*\}?})"
+    parser = re.compile(r"(\{{?[\w\d!.,[\]%:$<>+-= ]*\}?})")
 
 
 class PythonFormattingVariable:
-    parser = r"(%(%|(\([^)]+\)){0,1}[-+0#]{0,1}(\d+|\*){0,1}(\.(\d+|\*)){0,1}[hlL]{0,1}[diouxXeEfFgGcrs]{1}))"
+    parser = re.compile(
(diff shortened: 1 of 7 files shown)


pontoon/base/placeables.py
```diff
@@ -0,0 +1,45 @@
...
+class PythonFormattingVariable:
+    parser = r"(%(%|(\([^)]+\)){0,1}[-+0#]{0,1}(\d+|\*){0,1}(\.(\d+|\*)){0,1}[hlL]{0,1}[diouxXeEfFgGcrs]{1}))"
+    match_index = 0
+
+
+class FluentTerm:
+    parser = r"({ ?-[^}]* ?})"
+
+
+class FluentParametrizedTerm:
```
flodolo, line 21:
```suggestion
class FluentParameterizedTerm:
```

pontoon/base/placeables.py
flodolo, line 1: Maybe dumb question, but could we add at least placeables for Android? (this is [what I use](https://github.com/mozilla-l10n/mozl10n-linter/blob/main/l10n/android_l10n.py#L130) in the linter)

I also have [placeables](https://github.com/mozilla-l10n/mozl10n-linter/blob/main/l10n/xliff_l10n.py#L71) used in XLIFF for both iOS and VPN.

pontoon/base/placeables.py
```diff
@@ -0,0 +1,45 @@
...
+    parser = r"(\{{?[\w\d!.,[\]%:$<>+-= ]*\}?})"
+
+
+class PythonFormattingVariable:
+    parser = r"(%(%|(\([^)]+\)){0,1}[-+0#]{0,1}(\d+|\*){0,1}(\.(\d+|\*)){0,1}[hlL]{0,1}[diouxXeEfFgGcrs]{1}))"
+    match_index = 0
+
+
+class FluentTerm:
+    parser = r"({ ?-[^}]* ?})"
```
eemeli, line 18:
```suggestion
    parser = r"({ *-[^}]*})"
```

The spec allows for arbitrary amounts of whitespace, not up to one character. Also, the `[^}]` captures whitespace as well as any valid parameters, so FluentParametrizedTerm form is not required.

pontoon/base/placeables.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class FluentParametrizedTerm:
+    parser = r"({ ?-[^}]*([^}]*: ?[^}]*) ?})"
+
+
+class FluentFunction:
+    parser = r"({ ?[A-W0-9\-_]+[^}]* ?})"
```
eemeli, line 26:
```suggestion
    parser = r"({ *[A-W0-9\-_]+[^}]*})"
```

pontoon/base/placeables.py
```diff
@@ -0,0 +1,45 @@
...
+        # PythonFormattingVariable.parser,
+        FluentTerm.parser,
+        # FluentParametrizedTerm.parser,
+        FluentFunction.parser,
+    ]
+
+    placeables = []
+
+    for parser in parsers:
+        placeables += re.findall(parser, text)
```
eemeli: As `re.findall` doesn't care about groups, the `(...)` around the expressions above could be dropped.

pontoon/base/tests/test_placeables.py
eemeli, line 1: Would be good to have at least one test capturing more than one variable.

pontoon/base/placeables.py
mathjazz: Yes. I'm working on porting more placeables used in the frontend:
https://github.com/mozilla/react-content-marker/tree/master/src/parsers

Are the ones that you use catching stuff that Pontoon doesn't? It would be nice to keep all placeable parses in one place.

eemeli: Suggestion: This should probably also include HTML tags, i.e. something like `r"<.*?>"`.

Nitpick: "Placeables" is a _very_ Fluent term. It might be better to use "placeholder", which is the generic one.

pontoon/machinery/utils.py
```diff
@@ -96,37 +101,122 @@ def get_google_automl_translation(text, locale, format="text"):
...
+        new_terms = {term for term in placeables if term not in existing_terms}
+
+        if new_terms:
+            store_new_terms(url, headers, new_terms)
+
+    glossary_config = translate.TranslateTextGlossaryConfig(glossary=glossary_path)
+    request_params["glossary_config"] = glossary_config
+
+
+def get_existing_terms(url, headers):
```
eemeli, line 176: This seems like a thing that really ought to be cached. If that's too much to add to this PR, then please add a separate issue for it and a code comment here pointing at it.

Probably a new database table that we trust as the local source of truth that's updated during `store_new_terms()` and sync'd daily or weekly from upstream to account for any drift.

eemeli (changes requested):
See inline.

pontoon/base/placeables.py
flodolo: At least on iOS, they catch more than Pontoon: https://pontoon.mozilla.org/it/firefox-for-ios/firefox-ios.xliff/?search=%25%40&string=143714

mathjazz:
I haven't accepted any changes to the placeable regexes, because they were copied verbatim from the library:
https://github.com/mozilla/react-content-marker/tree/master/src/parsers

We should fix them at the source first.

pontoon/base/placeables.py
flodolo: I've never seen parametrized before, but apparently it's a thing.

For sure that's not the terminology we used in Fluent
https://projectfluent.org/fluent/guide/terms.html

mathjazz:
You can test on stage using the Machinery UI or the following command (TM is disabled):

```python
from pontoon.pretranslation.pretranslate import get_pretranslations

get_pretranslations(entity, locale)[0][0]
```

mathjazz:
Curious what do you both think about https://github.com/mozilla/pontoon/pull/2873/commits/8b03a884ca551102e3d92b2643900daca3c2c522?

pontoon/base/placeables.py
eemeli: > What if the HTML tag contains localizable text?
> 
> Example: https://pontoon.mozilla.org/sl/firefox-accounts/all-resources/?string=237723

Ugh, that's unfortunate. Why isn't the `data-l10n-id` on the ``, and the string in a Fluent attribute? But your point is valid: messages exist with localizable content within HTML elements.

So I think this is a question about how certain we want to be about not breaking stuff vs. making sure that localizable things are localized.

One choice would to use a regexp like `r'<("[^ >"]*"|[^>"])*>'`, which won't capture tags that contain quoted values that include spaces.

pontoon/base/placeables.py
```diff
@@ -2,39 +2,56 @@
...
 
 class FluentFunction:
-    parser = r"({ ?[A-W0-9\-_]+[^}]* ?})"
+    parser = re.compile(r"({ ?[A-W0-9\-_]+[^}]* ?})")
+
+
+class JavaFormattingVariable:
+    parser = re.compile(
+        r"(?x){[0-9]+(,\s*(number(,\s*(integer|currency|percent|[-0#.,E;%\u2030\u00a4']+)?)?|(date|time)(,\s*(short|medium|long|full|.+?))?|choice,([^{]+({.+})?)+)?)?}"
+    )
```
eemeli, line 37:
```suggestion
    parser = re.compile(
        r"""
        {
          [0-9]+
          (,\s*(number
            (,\s*(integer | currency | percent | [-0#.,E;%\u2030\u00a4']+)?)?
            | (date | time)(,\s*(short | medium | long | full | .+?))?
            | choice,([^{]+({.+})?)+
          )?)?
        }
        """,
        re.X,
    )
```

When using `re.compile()`, it's a bit cleaner to replace the inline `(?x)` with [re.X](https://docs.python.org/3/library/re.html#re.X), which allows whitespace.

The above is equivalent to the given regex, but isn't it a bit broken for date/time/choice?

pontoon/base/tests/test_placeables.py
```diff
@@ -22,15 +27,66 @@ def test_PythonFormatString():
...
     placeables = get_placeables(input)
 
     assert placeables == ["{ -name }"]
 
 
+def test_FluentParametrizedTerm():
+    input = "My { -name(foo-bar: 'now that's a value!') } is Luka."
+    placeables = get_placeables(input)
+
+    assert placeables == ["{ -name(foo-bar: 'now that's a value!') }"]
```
eemeli, line 48:
```suggestion
    input = "My { -name(foo-bar: \"now that's a value!\") } is Luka."
    placeables = get_placeables(input)

    assert placeables == ["{ -name(foo-bar: \"now that's a value!\") }"]
```

pontoon/base/tests/test_placeables.py
```diff
@@ -22,15 +27,66 @@ def test_PythonFormatString():
...
+
+
 def test_FluentFunction():
     input = "My { NAME() } is Luka."
     placeables = get_placeables(input)
 
     assert placeables == ["{ NAME() }"]
+
+
+def test_JavaFormattingVariable():
```
eemeli, line 58: These are succeeding because they're matching `PythonFormatString`, not `JavaFormattingVariable`.

eemeli:
I've added some inline comments; didn't have time for a proper review at least yet.

> I haven't accepted any changes to the placeable regexes, because they were copied verbatim from the library: https://github.com/mozilla/react-content-marker/tree/master/src/parsers

Ouch, that's a fair burn. I did not review them when moving the code from Pontoon.

> We should fix them at the source first.

I will not have the bandwidth for doing that until maybe July. Rather than using known-bad code here, would prefer doing the right thing first here, then fixing `react-content-marker` later.

pontoon/base/placeables.py
mathjazz: Making sure localizable things are localized should take priority, otherwise we almost certainly produce bad translations. As per our testing data, breaking HTML in machine translations is much less likely.

mathjazz:
Stage updated with latest fixes that address latest comments.

pontoon/base/placeables.py
```diff
@@ -3,38 +3,32 @@
 
 class PythonFormatNamedString:
     parser = re.compile(
-        r"(%\([[\w\d!.,[\]%:$<>+\-= ]*\)[+|-|0\d+|#]?[.\d+]?[s|d|e|f|g|o|x|c|%])",
+        r"(%\([[\w\d!.,[\]%:$<>+\-= ]*\)[+-0\d+#]?[.\d+]?[sdefgoxc%])",
         re.IGNORECASE,
```
eemeli: Presuming that this is attempting to capture [printf-style strings](https://docs.python.org/3/library/stdtypes.html#printf-style-string-formatting), it doesn't really make sense to have both this and `PythonFormattingVariable`; really both should be merged into one with a rewritten regex like this, and call it something like `PythonPrintfString`.

```suggestion
        r"""
            %
            (\(.*?\))?  # mapping key
            [\#0\-\ +]* # conversion flags
            [\d*]*      # minimum field width
            (\.[\d*])?  # precision
            [hlL]?      # length modifier
            [acdEeFfGgiorsuXx%]  # conversion type
        """,
        re.X,
```

pontoon/base/placeables.py
```diff
@@ -3,38 +3,32 @@
...
     )
 
 
 class PythonFormatString:
-    parser = re.compile(r"(\{{?[\w\d!.,[\]%:$<>+-= ]*\}?})")
+    parser = re.compile(r"({{?[\w!.,[\]%:$<>+\-= ]*}?})")
```
eemeli, line 12: Based on my reading of the [Python spec](https://docs.python.org/3/library/string.html#format-string-syntax), doubled `{{` and `}}` are parsed as standalone escaped characters.

Also, this is a really lax implementation that's also capturing e.g. most `FluentTerm` values as well.

```suggestion
    parser = re.compile(r"{{|}}|{[\w!.,[\]%:$<>+\-= ]*}")
```

pontoon/base/placeables.py
eemeli: Let's make it easy for ourselves and remove this; `FluentTerm` is already capturing practically all valid parameterized terms, and the ones that are valid that it isn't capturing aren't captured by this either.

pontoon/base/placeables.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def get_placeables(text):
+    """Return a list of placeables found in the given string."""
+    parsers = [
+        PythonFormatNamedString.parser,
+        PythonFormatString.parser,
+        PythonFormattingVariable.parser,
+        FluentTerm.parser,
+        FluentParametrizedTerm.parser,
```
eemeli, line 45: Following my preceding inline suggestions, I think we end up with:

```suggestion
        PythonPrintfString.parser,
        PythonFormatString.parser,
        FluentTerm.parser,
```

mathjazz:
> Also I think [this comment](https://github.com/mozilla/pontoon/pull/2873#discussion_r1221313834) is still to be addressed.

Is [this](https://github.com/mozilla/pontoon/pull/2873/files#diff-f38038a411f0c005e8060013652366803b7c2880f31bd4ef8d4cff40cc3d518cR70) not what you had in mind?

pontoon/base/placeables.py
mathjazz: `test_PythonFormatNamedString()` and `test_PythonFormattingVariable()` are both failing with this change applied:

```
AssertionError: assert [('(name)', '')] == ['%(name)s']
AssertionError: assert [('(name)', '')] == ['%(name)d']
```

eemeli:
> > Also I think [this comment](https://github.com/mozilla/pontoon/pull/2873#discussion_r1221313834) is still to be addressed.
> 
> Is [this](https://github.com/mozilla/pontoon/pull/2873/files#diff-f38038a411f0c005e8060013652366803b7c2880f31bd4ef8d4cff40cc3d518cR70) not what you had in mind?

Yes, sorry, had managed to somehow miss that previously.

> > Btw, I'm not sure that variables like `%S` or `%1$S` are being captured by anything. Should they be?
> 
> It would be nice, but I'd do that as a followup, because we don't have any errors related to that. Unless you're willing to provide a regex?

Here you go: `re.compile(r"%([1-9]\$)?S")`. That'll match the pattern commonly used in m-c .properties files that's not captured by the generic printf regex (which doesn't like the capital `S`).

pontoon/base/placeables.py
eemeli: I updated the suggestion in the above comment to remove the capturing groups (i.e. it's now using `(?: ... )` rather than `( ... )`), which should fix the issue.

pontoon/base/placeables.py
eemeli: To correct myself, [re.findall](https://docs.python.org/3/library/re.html#re.findall) _does_ care about groups, but it treats having one capturing group for the whole expression exactly the same as not having any capturing groups.

pontoon/base/placeables.py
```diff
@@ -0,0 +1,64 @@
...
+class PythonFormatString:
+    parser = re.compile(r"{{|}}|{[\w!.,[\]%:$<>+\-= ]*}")
+
+
+class FluentTerm:
+    parser = re.compile(r"({ *-[^}]*})")
+
+
+class FluentFunction:
+    parser = re.compile(r"({ *[A-W0-9\-_]+[^}]*})")
```
eemeli, line 32:
```suggestion
class FluentPlaceable:
    parser = re.compile(r"{ *[A-Z0-9\-_][^}]*}")
```

Terms are also captured by the function regexp, so let's fix the regexp's remaining bugs and merge them.

pontoon/base/placeables.py
```diff
@@ -0,0 +1,64 @@
...
+
+def get_placeables(text):
+    """Return a list of placeables found in the given string."""
+    parsers = [
+        MozillaPrintfString.parser,
+        PythonPrintfString.parser,
+        PythonFormatString.parser,
+        FluentTerm.parser,
+        FluentFunction.parser,
+        JsonPlaceholder.parser,
```
eemeli, line 47:
```suggestion
        JsonPlaceholder.parser,
        MozillaPrintfString.parser,
        PythonPrintfString.parser,
        PythonFormatString.parser,
        FluentPlaceable.parser,
```

Merge the Fluent parsers, try `JsonPlaceholder` first so that it'll capture e.g. `$S$` correctly.

eemeli (approved):
This is fine. A couple of minor inline fixes for remaining corner cases.

PR #2876, opened 2023-06-15
Always set target-language on serialization

mathjazz:
~Previously we only set the `target-language` if it wasn't set. From now on we will always set it to the target locale code, because we use the source file for template, so we don't have the existing `target-language` from the locale file readily available.~

~Update:
Most of the changes are [Pass project object to ParsedResource.save()](https://github.com/mozilla/pontoon/pull/2876/commits/881043ec07273b22bd4991df2ff58056efacab26), which is a preparation for [Always set target-language, but only map locales for iOS projects](https://github.com/mozilla/pontoon/pull/2876/commits/9e9ef16b8a9610c2acb6cf594b85207486e3cc61).~

~Previous commits are either reverted or cosmetic changes. The test is removed, because we now *always* set `target-language`.~

Update 2:
We now (again) only set the `target-language` if it wasn't set. We store the value when parsing the l10n file, and if it exists, we use it on serialization in order to preserve it. If `target-language` isn't set, we set it to the current locale code, taking into account the ios mapping.

### DIFF
--- a/pontoon/sync/formats/xliff.py
+++ b/pontoon/sync/formats/xliff.py
@@ -145,41 +145,23 @@ class XLIFFResource(ParsedResource):
                 if target is not None and "state" in target.attrib:
                     del target.attrib["state"]
 
+        locale_mapping = {
+            "bn-IN": "bn",
+            "ga-IE": "ga",
+            "nb-NO": "nb",
+            "nn-NO": "nn",
+            "sv-SE": "sv",
+        }
+
+        locale_code = locale_mapping.get(locale.code, locale.code)
+
+        # Set target-language if not set
+        file_node = self.xliff_file.namespaced("file")
+        for node in self.xliff_file.document.getroot().iterchildren(file_node):
+            node.set("target-language", locale_code)
+
         # Serialize and save the updated XLIFF file.
-        with open(self.path, "rb+") as f:
-            xml = f.read()
-
-            try:
-                locale_xliff_file = xliff.xlifffile(xml)
-            except etree.XMLSyntaxError as err:
-                raise ParseError(f"Failed to parse {self.path}: {err}")
-
-            # Read target-language if set
-            target_language = None
-            l10n_file = locale_xliff_file.namespaced("file")
-            for node in locale_xliff_file.document.getroot().iterchildren(l10n_file):
-                target_language = node.get("target-language")
-                if target_language:
-                    break
-
-            if target_language:
-                # TODO: should be made iOS specific
-                # Map locale codes for iOS: http://www.ibabbleon.com/iOS-Language-Codes-ISO-639.html
-                locale_mapping = {
-                    "bn-IN": "bn",
-                    "ga-IE": "ga",
-                    "nb-NO": "nb",
-                    "nn-NO": "nn",
-                    "sv-SE": "sv",
-                }
-
-                target_language = locale_mapping.get(locale.code, locale.code)
-
-                file_node = self.xliff_file.namespaced("file")
-                for node in self.xliff_file.document.getroot().iterchildren(file_node):
-                    node.set("target-language", target_language)
-
-            f.seek(0)
+        with open(self.path, "wb") as f:
             f.write(bytes(self.xliff_file))
 
 
--- a/pontoon/sync/formats/xliff.py
+++ b/pontoon/sync/formats/xliff.py
@@ -100,7 +100,7 @@ class XLIFFResource(ParsedResource):
     def translations(self):
         return sorted(self.entities.values(), key=lambda e: e.order)
 
-    def save(self, locale, project):
+    def save(self, locale):
         if not self.source_resource:
             raise SyncError(
                 "Cannot save XLIFF resource {}: No source resource given.".format(
--- a/pontoon/sync/formats/xliff.py
+++ b/pontoon/sync/formats/xliff.py
@@ -145,25 +145,20 @@ class XLIFFResource(ParsedResource):
                 if target is not None and "state" in target.attrib:
                     del target.attrib["state"]
 
-        locale_mapping = (
-            {
-                "bn-IN": "bn",
-                "ga-IE": "ga",
-                "nb-NO": "nb",
-                "nn-NO": "nn",
-                "sv-SE": "sv",
-            }
-            if "ios" in project.slug
-            else {}
-        )
+        locale_mapping = {
+            "bn-IN": "bn",
+            "ga-IE": "ga",
+            "nb-NO": "nb",
+            "nn-NO": "nn",
+            "sv-SE": "sv",
+        }
 
-        # Set target-language and map locale codes for iOS according to:
-        # http://www.ibabbleon.com/iOS-Language-Codes-ISO-639.html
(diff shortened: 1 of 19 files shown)


pontoon/sync/formats/xliff.py
```diff
@@ -152,15 +152,13 @@ def save(self, locale):
...
-        locale_code = locale.code
-        if locale_code in locale_mapping:
-            locale_code = locale_mapping[locale_code]
+
+        locale_code = locale_mapping.get(locale.code, locale.code)
 
         # Set target-language if not set
         file_node = self.xliff_file.namespaced("file")
         for node in self.xliff_file.document.getroot().iterchildren(file_node):
-            if not node.get("target-language"):
```
flodolo, line 162: We don't want to always set a target-language, because then we're back to https://bugzilla.mozilla.org/show_bug.cgi?id=1713058

I know it's hack-ish, but could we do something like this instead?

```
if node.get("target-language") not in ["en", "en-US"]
```

flodolo (changes requested):

pontoon/sync/formats/xliff.py
flodolo: Actually, we would still need to check if the value is set, not sure if this works in lxml-land

```
if node.get("target-language", "en-US") not in ["en", "en-US"]:
```

pontoon/sync/tests/formats/test_xliff.py
```diff
@@ -142,42 +142,6 @@ def test_save_basic(self):
             input_string, expected_string, source_string=input_string
         )
 
-    def test_save_preserve_target_language(self):
```
flodolo, line 145: Shouldn't we restore this test, and test for both iOS and non-iOS projects?

flodolo (commented):
The approach makes sense to me, but I think we need to add (restore) tests?

pontoon/sync/tests/formats/test_xliff.py
mathjazz: This test will fail, because it expects `target-language` to not be changed if it exists. This is no longer the case, because we always set it (the `if` from https://github.com/mozilla/pontoon/pull/2876#pullrequestreview-1482109243 is gone).

We should add a test to check different behavior for iOS projects. I didn't include it, because I'd like to handle it differently in a followup - I'd like to introduce `Locale.ios_code` and read iOS locales from the data model rather than hardcode them. We do something similar for Google Translate and Microsoft Translator locales.

eemeli (commented):
I don't think I understand how we get to a mismatching state of affairs, or what we're really looking to do here to fix it. I also don't really like the special-casing of a small set of locales based on the presence of a substring in the project slug.

Could we instead compare only the primary subtags of the current and previous locale code, and set it if they don't match? So for instance if we're saving a `sv-SE` file and we find a `target-language="sv"` then we leave it alone, but if it's empty, then we set it as `target-language="sv-SE"`.

Presumably this could result in up to one round of clobbering before agreeing on `"sv"`, but it'd be a generic solution.

flodolo:
> So for instance if we're saving a `sv-SE` file and we find a `target-language="sv"` then we leave it alone, but if it's empty, then we set it as `target-language="sv-SE"`.

I thought the same. The problem is that we don't have the localized file at this point, only source.

mathjazz:
> I don't think I understand how we get to a mismatching state of affairs, or what we're really looking to do here to fix it. 

As of https://github.com/mozilla/pontoon/commit/b46ec36ce0910d4a52373bb567ca78d6af8d8fb2, we're using the source file as a template on serialization, in order to avoid the need to merge l10n files when new strings are added. That means the old code, which checks if the `target-language` is set, always returns `True` (because the value is `en-US`).

> I also don't really like the special-casing of a small set of locales based on the presence of a substring in the project slug.

The idea is to land this PR today as a hotfix and work on a more proper solution in a followup. Part of that would be to [`Locale.ios_code`](https://github.com/mozilla/pontoon/pull/2876#discussion_r1232084179) and `Project.use_ios_locales`.

> Could we instead compare only the primary subtags of the current and previous locale code, and set it if they don't match? So for instance if we're saving a `sv-SE` file and we find a `target-language="sv"` then we leave it alone, but if it's empty, then we set it as `target-language="sv-SE"`.

Is that essentially the alternative approach from https://github.com/mozilla/pontoon/pull/2876/commits/73c1da36793c5f22d21fe37372eba8be779fe2db? It's also not pretty, because we need to read the l10n file and make sure it's available.

flodolo (approved):

eemeli:
> > I don't think I understand how we get to a mismatching state of affairs, or what we're really looking to do here to fix it.
> 
> As of [b46ec36](https://github.com/mozilla/pontoon/commit/b46ec36ce0910d4a52373bb567ca78d6af8d8fb2), we're using the source file as a template on serialization, in order to avoid the need to merge l10n files when new strings are added. That means the old code, which checks if the `target-language` is set, always returns `True` (because the value is `en-US`).

Ok, so wouldn't the right check be to see if `target-language` is the same as `source-language`, rather than special-casing `en` or `en-US`?

> > Could we instead compare only the primary subtags of the current and previous locale code, and set it if they don't match? So for instance if we're saving a `sv-SE` file and we find a `target-language="sv"` then we leave it alone, but if it's empty, then we set it as `target-language="sv-SE"`.
> 
> Is that essentially the alternative approach from [73c1da3](https://github.com/mozilla/pontoon/commit/73c1da36793c5f22d21fe37372eba8be779fe2db)? It's also not pretty, because we need to read the l10n file and make sure it's available.

I'm not really sure what you're asking here. This code is already reading the previous `target-language` and in some conditions setting it. My suggestion is to set it if the previous `target-language` value's primary subtag is not the one we expect it to be.

mathjazz:
> I'm not really sure what you're asking here. This code is already reading the previous `target-language` and in some conditions setting it. My suggestion is to set it if the previous `target-language` value's primary subtag is not the one we expect it to be.

OK, I think we understand each other. :) I'll switch to that solution then.

mathjazz:
It turns out preserving `target-language` is trivial. I understand my code better now. Sorry for the noise!

pontoon/sync/formats/xliff.py
```diff
@@ -96,6 +97,12 @@ def __init__(self, path, locale, source_resource=None):
                 # Add the entity to the entities dictionary using its key as the dictionary key
                 self.entities[entity.key] = entity
 
+            # Store target-language, needed for serialization
+            file_tag = self.xliff_file.namespaced("file")
+            file_nodes = self.xliff_file.document.getroot().iterchildren(file_tag)
+            if file_nodes:
+                self.target_language = next(file_nodes).get("target-language")
```
eemeli, line 104: Is it possible for this to be incorrectly set to the source language (i.e. `en-US`), as was discussed earlier?

eemeli (approved):
This seems fine to me, but I still don't have a deep understanding of how other tools may produce and modify these XLIFF files.

pontoon/sync/formats/xliff.py
mathjazz: No, unless of course the l10n file is not set up properly in first place.

Issue #2881, opened 2023-06-20
Spellchecking not working anymore


gion-andri:
It appears that there has been a recent update in Pontoon, containing a modification of the translation input field from a `textarea` to a `div` element. 

Consequently, this change has caused the integrated spellchecking feature of Firefox (and probably other browsers) to cease functioning.

As correct spelling is quite an important aspect in Ponton, please find a was to re-enable that.

flodolo:
I see `spellcheck="false"` analyzing the code 🤔 

Manually flipping that it seems to work, I wonder why they disable it.

eemeli:
Based on [this comment](https://discuss.codemirror.net/t/upgrading-from-codemirror-5/2584/2), enabling spellcheck via [contentAttributes](https://codemirror.net/docs/ref/#view.EditorView^contentAttributes) ought to "just work".

> EDIT: [this](https://discuss.codemirror.net/t/inputstyle-contenteditable-we-may-hope-for-browser-spell-checking/608/6) is probably useful (and worrying)? The spellchecking kicks in very late in [this demo](https://plnkr.co/edit/oy23zvjHqpLd7DPbBwQL?p=preview&preview), when you leave the editor or go on a new line.

Both of those examples use CM5, and CM6 is completely different.

flodolo:
I'm not seeing spellchecking enabled in Firefox, even if I force reload the page.
https://pontoon.mozilla.org/it/mozillaorg/all-resources/?status=unreviewed&string=281232

I also still see the `spellcheck="false"` in the code.

mathjazz:
Are logged in?

Is it the same in the private mode?

What about in the Firefox troubleshooting mode?

flodolo:
> Are logged in?

Yes, otherwise the editor is disabled (not editable).

> Is it the same in the private mode?
> What about in the Firefox troubleshooting mode?

It works on a different profile with Firefox release. 

Apparently, it doesn't work with this add-on, while it was working just fine with a textarea
https://addons.mozilla.org/firefox/addon/languagetool/

I suppose that's a problem in Firefox and not Pontoon, still annoying :-(

PR #2882, opened 2023-06-20
Revert codemirror

mathjazz:
This PR reverts the CodeMirror editor and the followup commit which introduces highlighting placeables.

Reason: spellcheker doesn't work with CodeMirror (see #2881).


mathjazz:
Superseded by #2884.

PR #2884, opened 2023-06-20
Re-enable spellchecker in translation editor

mathjazz:


translate/src/modules/translationform/components/EditField.tsx
```diff
@@ -66,6 +66,7 @@ export const EditField = memo(
                   setResultFromInput(index, update.state.doc.toString());
                 }
               }),
+              EditorView.contentAttributes.of({ spellcheck: 'true' }),
```
eemeli, line 69: This should be set in `getExtensions()` rather than here, as it's not dependent on any of the EditField state. But presuming that you've tested this to work as intended, this should be the right fix.

eemeli (approved):

translate/src/modules/translationform/components/EditField.tsx
mathjazz: I thought this would also trigger the spellcheck in the read-only mode. But it doesn't, so let's move it.

Issue #2152, opened 2018-11-14
Redirect old project slugs to the new one


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1507422](https://bugzilla.mozilla.org/show_bug.cgi?id=1507422)
Bug Reporter: @mathjazz
CC:   @Pike, @Delphine

When the project slug changes, we should keep record of the old one and redirect it to the new slug. I like how that works on GitHub.

Without such mechanism, we break links and awesomebars when the slug changes. For more important projects we currently mitigate that with hardcoded redirects:
https://github.com/mozilla/pontoon/blob/8d71348cd55f436920e19965e53faa0c08d7af3c/pontoon/urls.py#L22-L62

But that's ugly and non-maintainable.

Implementation note: we could transform a slug field into a separate table of unique slugs with a Project ForeignKey.

ayanaar:
**Below is my proposed solution:**

1. Create a new Model `ProjectSlugHistory`

This model will store the historical data about a project's slugs. It will have a foreign key relationship to the `Project` model (meaning each `Project` can have many associated `ProjectSlugHistory` instances), the old slug, and a timestamp marking when the slug was replaced.

2. Create a Post-Save Signal for the Project Model

This is a functionality provided by Django that triggers an action after a model instance has been saved. In our case, we want to capture when the slug of a `Project` changes. When the slug changes, we'll create a new `ProjectSlugHistory` instance storing the old slug and linking it to the `Project`. This will automatically record any slug changes when a project is saved.

3. Update the URL Configuration

We'll add a new URL pattern that captures requests to the old project slugs. This pattern will call a view (a function that handles the request and returns a response) which we will define in the next step.

4. Create a New View to Handle the Redirect

This view will receive a request to an old project slug, look up the current slug for the project using the `ProjectSlugHistory` model, and then redirect the request to the URL with the current slug.

This approach maintains a history of past slugs, automatically updates this history when the slug changes, and provides a means to redirect requests from old slugs to the current one. It doesn't require hardcoding each slug redirect and is easier to maintain and extend. It's also more resilient to errors because it's based on the actual history of slug changes.

mathjazz:
I love your proposal. Let's do this!

PR #2888, opened 2023-06-21
2512

ayanaar:
#2152


mathjazz:
This branch needs to be rebased against latest master. There are a lot of conflicts, so it might be easier to create a new branch out of master and manually copy relevant changes there, because there aren't that many.

I did that locally and then I was able to access `ProjectSlugHistory` in Django Admin.

PR #2889, opened 2023-06-21
Add Slug History and Redirect for Project Model

ayanaar:
#2152


pontoon/projects/views.py
```diff
@@ -43,9 +44,23 @@ def projects(request):
...
+            ProjectSlugHistory.objects.filter(old_slug=slug)
+            .order_by("-created_at")
+            .first()
+        )
+
+        if slug_history is not None:
+            # Redirect to the current project slug
+            return redirect("pontoon.projects.project", slug=slug_history.project.slug)
+        else:
+            raise Http404("Project not found.")
```
mathjazz, line 63: There are quite a few views in which we'd need a check like this. For example, we'd also want to redirect the localization dashboard (/{locale}/{project}), translate view (/{locale}/{project}/all-resources/), as well as several AJAX views spread across multiple urls.py files.

Can we instead introduce a few URL patters that will catch old project slugs as proposed [initially](https://github.com/mozilla/pontoon/issues/2152#issuecomment-1591899132)?

mathjazz (commented):
Nice work!

There's two minor changes I requested, as well as a more holistic approach to redirecting several URLs.

eemeli (changes requested):
I'm a bit concerned with the proposed order of operations, for two reasons:
1. This adds a ProjectSlugHistory lookup to every request, even when the slug is the current one used by a project. This seems rather wasteful.
2. What happens if a slug is re-used? I can think of two potentially problematic scenarios:
   1. If a project's slug is renamed `aa` → `bb` → `aa`, there's potential redirect loop that may be entered.
   2. If one project's slug is renamed `aa` → `bb`, and then a different project starts using the `aa` slug, requests for it get redirected to the `bb` project.

Both of the above concerns would be resolved if the ProjectSlugHistory lookup were only done after an initial project slug lookup failed. Is there a reason this would not work?

Test cases should also be included for the scenarios mentioned above.

pontoon/base/utils.py
```diff
@@ -642,29 +646,34 @@ def wrapper(request, *args, **kwargs):
             # If slug is still not defined, raise an error.
             if slug is None:
                 raise ValueError("Cannot extract slug from view arguments")
+            # Try to find a project using the current slug
+            try:
+                Project.objects.get(slug=slug)
+                # If found, continue with the original view function
+                return view_func(request, *args, **kwargs)
```
eemeli, line 653: This is better, but still adds a project lookup to each request that's then discarded. Is there a way to do as I suggested earlier, and to include this redirection in the error handling rather than proactively?

In each of the three functions that use this decorator, there's a call to `get_object_or_404()` for something like `Project.objects.get(slug=slug)`. Could that call be changed to trigger this flow, instead of the current decorator approach?

eemeli (changes requested):

pontoon/base/utils.py
```diff
@@ -606,3 +609,33 @@ def is_email(email):
...
+def get_project_or_redirect(
+    slug, redirect_view_name, slug_arg_name, request_user, **kwargs
+):
+    """
+    Attempts to get a project with the given slug. If the project doesn't exist, it checks if the slug is in the
+    ProjectSlugHistory and if so, it redirects to the current project slug URL. If the old slug is not found in the
+    history, it raises an Http404 error.
+    """
+    Project = apps.get_model("base", "Project")
+    ProjectSlugHistory = apps.get_model("base", "ProjectSlugHistory")
```
eemeli, line 623: Is there a reason to use this pattern rather than the function-local import in e.g. `get_download_content()` above?

pontoon/base/utils.py
```diff
@@ -606,3 +609,33 @@ def is_email(email):
...
+    try:
+        project = get_object_or_404(
+            Project.objects.visible_for(request_user).available(), slug=slug
+        )
```
eemeli, line 627: This wrapper provides no use, as the potential error is being immediately caught; see the example in the docs for what it's doing: https://docs.djangoproject.com/en/3.2/topics/http/shortcuts/#get-object-or-404

pontoon/base/utils.py
```diff
@@ -21,10 +21,13 @@
...
 from django.db.models.query import QuerySet
 from django.http import HttpResponseBadRequest
-from django.shortcuts import get_object_or_404
 from django.utils import timezone
 from django.utils.text import slugify
 from django.utils.translation import trans_real
+from django.shortcuts import redirect, get_object_or_404
+from django.apps import apps
+from django.urls import reverse
+from django.http import Http404
```
eemeli, line 30: nitpick: Would be nice to keep the imports sorted and un-duplicated, but I don't think we've been very strict about that.

eemeli (approved):
This looks good! Some questions and nitpicks on line comments.

@mathjazz Do you want to look at this again, as its shape has evolved somewhat, or should we just merge & push to staging?

mathjazz:
This looks good, nice work!

I'm pushing the branch to stage.

The only thing I'm left wondering about is whether we should redirect just the three main URLs (current state), or all.

eemeli (approved):

flodolo:
This seems to work correctly, although there's an edge case which is not considered, and to me it feels like a bug.

I changed the slug for AMO from `amo` to `amo1`, then to `amo2`. There's a check preventing me to use a slug already in use, which is good (BTW, the message `Project with this Slug already exists.` feels a bit weird with an article at the beginning, i.e. `A project with this Slug already exists.`). Links with both `amo`  and `amo1` correctly redirect to `amo2`.

Then I went to a different project (AMO Linter), and switched its slug to `amo1`. Surprisingly, it let me do that. I assumed we should prevent reusing an existing slug, even if it's not actively used by a project?

eemeli:
> Then I went to a different project (AMO Linter), and switched its slug to `amo1`. Surprisingly, it let me do that. I assumed we should prevent reusing an existing slug, even if it's not actively used by a project?

This should be fine, and already handled by this PR. Historical slugs are not required to be unique, but on access the most recent slug wins.

For an example of a different system that works this way, see GitHub repo renames: They'll redirect old URLs until and unless you define a new repo with the same name as a past one had.

In my opinion, no further work is required on this PR before finally merging it.

mathjazz:
On Pontoon Call we've decided to move on with the current state of the patch. Nice work, @ayanaar!

Issue #2880, opened 2023-06-20
Can't type in editor with pending suggestion


flodolo:
https://pontoon.mozilla.org/it/mozillaorg/en/products/vpn/platforms/android_v2.ftl/?string=281300

The string in question has a pending suggestion. If you click the editor, nothing happens.

If I move to a missing string, then back to the string in question, then I can edit it.

flodolo:
Not sure what I did differently, but clicking that link now works as expected 🤔 

Leaving it open in case someone else can reproduce this.

mathjazz:
@flodolo Seems like this is now also fixed?

https://mozilla-pontoon-staging.herokuapp.com/it/mozillaorg/en/products/vpn/platforms/android_v2.ftl/?string=281300

flodolo:
I hope so. It was reproducing randomly even before, so hard to tell.

PR #2890, opened 2023-06-22
Revert CodeMirror

### LINKED ISSUE #2885: Translators can't approve pending suggestions
I don't have more details, but we received reports from at least 2 locales (fy-NL, da) that they can't approve pending suggestions.

For those affected: it would be great if you could check the brower console and see if anything useful error shows up when you load the page, or when you try to approve a suggestion.

mathjazz:
This patch reverts the CodeMirror editor and the 2 followup patches.

We need to fix #2885 and ideally also #2880 before enabling CodeMirror again.


mathjazz:
Deployed to stage.

flodolo (approved):

PR #2891, opened 2023-06-22
Add light theme spec

mathjazz:
This spec introduces the Pontoon light theme using the mockups of the two commonly used views:
* the translate view
* the team dashboard

In the implementation phase, we might make changes to the proposed color scheme, because it might not work across all the views. We might also uncover incosistencies in the default (dark) theme and make changes accordingly.

That means this is not the exact proposal, but more the direction we'd like to take with the light theme.


specs/0113-light-theme.md
```diff
@@ -0,0 +1,35 @@
...
+
+Implement light theme and enable users to switch between the (current) dark theme and the light theme.
+
+# Motivation
+
+Over the years, many users have expressed preference for a light color theme over the Pontoon default dark theme. Having the ability to switch themes will also allow us to implement high contrast themes, which benefit low vision users, increase readability and reduce website’s visual noise.
+
+# Implementation notes
+
+In the first step, we'll replace existing hardcoded color values with CSS variables. We'll store those in separate CSS files, with one version of each file for each theme. Users will be able to switch theme in their settings.
```
flodolo, line 15: Should there be a third option to follow browser/OS settings? Dark, Light, Automatic.

I also think the specs should be a bit more prescriptive on where this setting would be, and what it would say. Without it, we're deferring the decision to the implementation phase.

specs/0113-light-theme.md
```diff
@@ -0,0 +1,35 @@
...
+
+See below for the actual light theme color values.
```
flodolo, line 17: This looks very limited given how large Pontoon is. Should these specs be done after we `replace existing hardcoded color values with CSS variables`, and have an actual measure of how many colors we need to replace (and maybe change for consistency)?

mathjazz: I've changed the format of the color values, but the list actually covers the translate view and dashboards pretty well. The expectation is that we'll be able to keep colors like status icons the same across both themes.

I wouldn't block the spec on factoring out the color values, because that won't take us to the finish line either. It's totally possible we'll discover that we need a broader spectrum of colors as part of developing the light theme.

These spec is a bit special, because unlike in the feature specs, the idea here is to have a foundation on which to work on the designs for the whole app. Creating a spec with all the designs would be great, but that would essentially solve the vast majority of the implementation.

specs/0113-light-theme.md
```diff
@@ -0,0 +1,43 @@
...
+
+# Motivation
+
+Over the years, many users have expressed preference for a light color theme over the Pontoon default dark theme. Having the ability to switch themes will also allow us to implement high contrast themes, which benefit low vision users, increase readability and reduce website’s visual noise.
+
+# Implementation notes
+
+In the first step, we'll replace existing hardcoded color values with CSS variables. We'll store those in separate CSS files, with one version of each file for each theme.
+
+Users will be able to switch between themes in their settings. Each theme is represented with a title and a radio box. As soon as the theme is selected, it will get applied and the settings will be saved. A "Sync with system theme" checkbox allows users to let the browser pick the theme from the OS setting using the [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) CSS media feature.
```
flodolo, line 17: I don't think 2 radio buttons plus a checkbox is very intuitive. What's the concern about using 3 radio buttons? See Firefox settings for an example.

mathjazz: No particular concern. I like how Slack (and to some extent GitHub) show which theme is selected by the system.

flodolo: Let's see what Eemeli thinks, I'm OK with either.

For reference: Slack, GitHub, Jira, Firefox.

slack
github
jira
firefox

flodolo: Note: either way (3rd radio button, or checkbox), I think the specs should say which settings are the default.

To me, match the system sounds like a good default.

mathjazz: Matching the system would surprise existing users that use the light theme in their OS, which is most users.

Another argument for the dark default theme is that all our docs use the dark theme.

mathjazz: Data shows that 100% of our users use the dark theme ATM. 😛😛😛

specs/0113-light-theme.md
```diff
@@ -0,0 +1,47 @@
...
+
+Implement light theme and enable users to switch between the (current) dark theme and the light theme.
+
+# Motivation
+
+Over the years, many users have expressed preference for a light color theme over the Pontoon default dark theme. Having the ability to switch themes will also allow us to implement high contrast themes, which benefit low vision users, increase readability and reduce website’s visual noise.
+
+# Feature explanation
+
+Users are able to switch between themes in their settings. Each theme is represented with a title and a radio box. As soon as the theme is selected, it gets applied and the settings get saved. The default theme is the dark theme.
```
flodolo, line 15:
```suggestion
Users are able to switch between `Dark theme` and `Light theme` in their settings. Each theme is represented with a title and a radio box. As soon as the theme is selected, it gets applied and the setting get saved. The default theme is the dark theme.
```

I would call out that we'll only have Dark and Light to start

specs/0113-light-theme.md
```diff
@@ -0,0 +1,47 @@
...
+
+A "Sync with system theme" checkbox allows users to let the browser pick the theme from the OS setting using the [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) CSS media feature.
```
flodolo, line 17:
```suggestion
A "Sync with system theme" checkbox allows users to let the browser pick the theme from the OS setting using the [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) CSS media feature. By default, the checkbox is unchecked.
```

flodolo (approved):

specs/0113-light-theme.md
eemeli: Ultimately, the automatic option should be the default one. Maybe not initially though, to let us effectively beta-test the light theme with those who actively switch to it.

specs/0113-light-theme.md
eemeli: I have no strong opinion on the shape of the toggle, though it would be nice if it fit in the hamburger menu.

eemeli (approved):
The toggle is a bit underspecified as discussed in the comments, but otherwise it looks good.

The list of colors would really benefit from being able to see color swatches, but that might be more hassle than it's worth.

mathjazz:
Thanks for the review.

Let's simplify and only use radio boxes.

I've also included the theme selector in the profile menu.

PR #2906, opened 2023-07-11
Update psycopg2 to 2.9.6

flozz:
This PR updates the psycopg2 package to its latest version to make Pontoon compatible with Python 3.11.

Debian 12 Bookworm that was released a month ago is shipped with Python 3.11. The psycopg2 version used by Pontoon (2.8.5) does not work with Python 3.11 ([compilation error](https://github.com/wanadev/pontoon-debian/actions/runs/5514876065/jobs/10054678484)).


mathjazz (approved):
Tested on stage. LGTM. Thanks!

Issue #2662, opened 2022-11-20
Update Pontoon DB


mathjazz:
Our current PostgreSQL version reaches End of Life on 9 November, 2023. Therefore, the Pontoon database will need to be updated before that date.

We should make sure DB versions are unified as much as possible in all our environments:
- dev (Docker, also used by GitHub workflows)
- stage
- production
- app.json (custom deployment)

Resources:
https://devcenter.heroku.com/articles/upgrading-heroku-postgres-databases

Issue #2460, opened 2022-03-21
Broken heroku deploy button


Goddino-Repo:
When i try to deploy pontoon on heroku, i got an infinite charge..
![image](https://user-images.githubusercontent.com/39674510/159341794-4393827a-5b36-49b7-89b4-52b9602d1e72.png)

I just clicked on button from README file and logged in to heroku platform

mathjazz:
Thanks for reporting!

Could you please confirm that you can deploy from the button in the docs?
https://mozilla-pontoon.readthedocs.io/en/latest/admin/deployment.html

mathjazz:
You mean Heroku login or Pontoon login?

If Pontoon, you need to create an admin account first:
https://mozilla-pontoon.readthedocs.io/en/latest/admin/deployment.html#creating-an-admin-user

Goddino-Repo:
Ok Thanks, there was my mistake

mathjazz:
The issue with the Heroku Deploy button in the repository is still a valid issue.

jamesking:
I came across this problem too. When I use the button, I get taken to the Heroku `Create New App` page, which loads UI, but no resources show in the `Resources` section of the page and the `Add to pipeline…` button does nothing.

![CleanShot 2023-03-14 at 14 11 33@2x](https://user-images.githubusercontent.com/11338/225029391-773667b3-553a-4f0d-acb9-53a0a4a13d50.png)

There are some errors in the JS console for the Heroku site which makes me think this is a Heroku issue. 

<img width="800" alt="CleanShot 2023-03-14 at 14 16 37@2x" src="https://user-images.githubusercontent.com/11338/225029579-28b69b81-1fe7-45f8-bde0-b74898bfc179.png">

mathjazz:
Confirming I can reproduce the issue.

Workaround with using the button on https://mozilla-pontoon.readthedocs.io/en/latest/admin/deployment.html#creating-an-admin-user is no longer possible.

PR #2907, opened 2023-07-14
Update README.md with the right deploy URL

julienw:
The deploy to heroku button didn't work. I used the same URL as the one in https://mozilla-pontoon.readthedocs.io/en/latest/admin/deployment.html instead, that seems to work better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ If you are looking to host your own instance of Pontoon, there are several ways 
 To deploy Pontoon to Heroku without leaving your web browser, click the **Deploy to
 Heroku** button below.
 
-[![Deploy](https://www.herokucdn.com/deploy/button.svg)](https://heroku.com/deploy)
+[![Deploy](https://www.herokucdn.com/deploy/button.svg)](https://heroku.com/deploy?template=https://github.com/mozilla/pontoon/tree/master)
 
 Alternatively, you can deploy to Heroku manually by following our
 [Deployment Documentation](https://mozilla-pontoon.readthedocs.io/en/latest/admin/deployment.html).


mathjazz:
Thanks for the patch, @julienw!

I tried the buttons at https://github.com/mozilla/pontoon and https://mozilla-pontoon.readthedocs.io/en/latest/admin/deployment.html, and they both result in an error:

```
Configure environment
There was an issue setting up your app environment.
    We tried to create heroku-postgresql:mini, but received an error from the add-on provider.
    Try the request again, or log a support ticket and include this message:
    Available versions are 12, 13, 14, 15 (the default is 15)
```

Does the one on RTD work OK for you?

Also see #2460.

mathjazz:
Correct, the error happens when you try to deploy.

julienw:
I see, it's strange that the list of resources looks OK in the template form.

mathjazz:
Does the current button at https://github.com/mozilla/pontoon take you to [that template](https://github.com/mozilla/pontoon/pull/2907#issuecomment-1640957729), too?

mathjazz:
Let's revisit this after #2662 is fixed.

surapunoyousei:
> Thanks for the patch, @julienw!
> 
> I tried the buttons at https://github.com/mozilla/pontoon and https://mozilla-pontoon.readthedocs.io/en/latest/admin/deployment.html, and they both result in an error:
> 
> ```
> Configure environment
> There was an issue setting up your app environment.
>     We tried to create heroku-postgresql:mini, but received an error from the add-on provider.
>     Try the request again, or log a support ticket and include this message:
>     Available versions are 12, 13, 14, 15 (the default is 15)
> ```
> 
> Seems like we need to upgrade Postgres first.
> 
> Also see #2460.

@mathjazz 

I am currently unable to deploy due to this error. Is there another way?

mathjazz:
Hey @surapunoyousei, we need to fix #2662 before Deploy to Heroku button will work again.

Until then I recommend following the official instructions for manual deployment to Heroku:
https://mozilla-pontoon.readthedocs.io/en/latest/admin/deployment.html

surapunoyousei:
Thanks  @mathjazz

PR #2912, opened 2023-07-19
Add `date_created` Field to Entity and Display in Metadata Component 

### LINKED ISSUE #2903: Show when source string was created in Metadata
We store Entity creation timestamp in `Entity.date_created`. We should show that information in the Metadata section of the translate view.

ayanaar:
Fixes #2903 

This pull request introduces a new `date_created` attribute to the Entity type and displays this date in the Metadata component in a human-readable format. 

## Key changes:

- `date_created` attribute is added to the Entity type in the API.
- The Metadata component now includes a new `EntityCreatedDate` function that displays the creation date of the entity.
- The `map_entities` function now includes `date_created` as a `datetime` object.
- Corresponding tests in `test_entity.py` have been adjusted to reflect the changes in `map_entities`.


translate/src/modules/entitydetails/components/Metadata.tsx
```diff
@@ -86,6 +86,44 @@ function GroupComment({ comment }: { comment: string }) {
...
+  // Create a Date object from the dateCreated string
+  const date = new Date(dateCreated);
+
+  // Check if date is a valid Date object
+  if (isNaN(date.getTime())) {
+    console.error(
+      `EntityCreatedDate component received invalid date string: ${dateCreated}`,
+    );
+    return null;
+  }
```
mathjazz, line 111: We don't allow anything but a date to be entered in the data model, so I wonder if we can drop this check.

translate/src/context/EntityView.tsx
```diff
@@ -31,6 +31,11 @@ const emptyEntity: Entity = {
   translation: [],
   readonly: true,
   isSibling: false,
+  date_created: new Date().toLocaleDateString('en-US', {
+    day: 'numeric',
+    month: 'long',
+    year: 'numeric',
+  }),
```
mathjazz, line 38: Why not just use `''`, given `date_created` is a string?

translate/src/modules/entitydetails/components/Metadata.tsx
mathjazz: Seems like adding `date_created: new Date().toISOString(),` to the `ENTITY` objects in both failing test files solves the problem.

translate/src/modules/entitydetails/components/Metadata.tsx
```diff
@@ -233,6 +263,7 @@ export function Metadata({
       <GroupComment comment={entity.group_comment} />
       <ResourceComment comment={entity.resource_comment} key={entity.pk} />
       <FluentAttribute entity={entity} />
+      <EntityCreatedDate dateCreated={entity.date_created} />
```
mathjazz, line 266: Please move this element to the very bottom, under `EntityContext`, because it's the least relevant.

translate/src/modules/entitydetails/components/Metadata.tsx
```diff
@@ -86,6 +86,36 @@ function GroupComment({ comment }: { comment: string }) {
...
+
+  // Format the date and time
+  const formattedDate = dateFormatter.format(date);
+  const formattedTime = timeFormatter.format(date);
+
+  // Combine the formatted date and time into one string
+  const formattedDateTime = `${formattedDate} ${formattedTime}`;
+
+  return (
+    <Datum id='entity-created-date' title='CREATED DATE'>
```
mathjazz, line 113: Please make sure you add `entitydetails-Metadata--entity-created-date` item to the [`translate.ftl`](https://github.com/mozilla/pontoon/blob/master/translate/public/locale/en-US/translate.ftl#L273) file. I would just use CREATED or CREATED ON, because it also includes time.

mathjazz (approved):
Great work!

PR #2916, opened 2023-07-24
Update parsimonious to v0.10

flozz:
This PR updates parsimonious from 0.8.1 to 0.10 that is compatible with Python 3.11.


mathjazz (approved):
LGTM

PR #2922, opened 2023-07-31
Only spellcheck string content, not placeholders or html tags

eemeli:
With the new editor and its syntax highlighting, we can limit the spellcheck to just the literal translatable content of a message, rather than needing to apply it to all of it.

This allows us to e.g. not have a fluent placeholder `{ $tabCount }` get marked as a spelling error by the browser.

This isn't a workaround for [Firefox bug 1837268](https://bugzilla.mozilla.org/show_bug.cgi?id=1837268), though tbf seeing if it were was a part of the reason why I implemented this.


mathjazz (commented):
LGTM

PR #2926, opened 2023-08-08
Improve Pontoon RTL editing

### LINKED ISSUE #2146: Mixed RTL and LTR content is hard to read in text inputs

eemeli:
Fixes #2146 

This should make editing RTL content in Pontoon much more pleasant with two changes:
1. Spans of syntax are explicitly marked to have LTR directionality. It's not enough to mark each individual bit of syntax as LTR, because they'd still be concatenated as RTL, which gets messy.
2. Syntax is styled with `white-space: pre`, so that it's not soft-wrapped on multiple lines. This may force the editor box to scroll horizontally on narrow screens and with long spans of syntax, but it's effectively required to not make some RTL messages really really messy.

It would be good to get feedback on these changes from our localisers of RTL languages. I'll be deploying the change to staging shortly.


eemeli:
It turns out that to solve this right, I needed to go deeper and refactor the editor's parsing and decorators pretty deeply. The maximal situation we need to support is having a left-to-right HTML element in a right-to-left message with an auto-directional attribute value which contains a left-to-right placeholder. This gets a bit tricky in CodeMirror, where the decorators for each of these layers needs to get the right precedence in order for the right DOM to be rendered. It's solved here by having multiple decorator plugins.

Quoted attributes are now also styled like text, but they're still left out of spellchecking.

I still need to look into arrow-key navigation near direction boundaries: That at least initially looks like a CodeMirror bug, esp. as the same behaviour shows up in multiple browsers.

suheekims:
> Fix

eemeli:
It's better now, but still a bit wonky around the start of a LTR span within a RTL context. But even if we don't manage to fix that, this seems clearly better than the current situation.

eemeli:
Looks like something in the frontend is no longer happy with Node.js 14, so I updated all the JS actions to the current LTS, Node.js 18.

eemeli:
The recent CodeMirror fixes by @marijnh have let us get this to a state where I think it's ready for prod deployment. There are still a few corner-case issues that could be improved, but it's unclear if these are possibly browser bugs. Each of these only shows up for RTL scripts:
- With Firefox, if a translations ends with a placeholder, the "end of string" position for the cursor is at the right side of that placeholder.
- With Chrome and Safari, the cursor position at the beginning and end of a `dir=auto` span within a `dir=ltr` span is effectively at the wrong end.
- With all of the above browsers, when a `dir=auto` span within a `dir=ltr` span itself contains a `dir=ltr` span as its last element, the left/right arrow navigation treats it as if it was at the other end.

This PR is currently deployed on our staging server, where [this message](https://mozilla-pontoon-staging.herokuapp.com/he/firefox-accounts/all-resources/?string=237718) demonstrates the above issues.

@mathjazz Given that the above are relatively rare, and that the overall change here is so much better than what we currently have, I think this is ready to land. Would you agree?

mathjazz:
Thanks for the great work here!

Comparing with the same string [on prod](https://pontoon.mozilla.org/he/firefox-accounts/all-resources/?string=237718), this is a big step forward in terms of presentation.

Yes, there is a regression if you move the cursor between chunks with different directionalities. To me that looks like a smaller issue, but let's wait to hear back from the localizers.

mathjazz (approved):
We didn't get any pushback so let's go ahead.

Issue #2923, opened 2023-07-31
Implement pretranslation monitoring


mathjazz:
This issue tracks the implementation of Pretranslation Monitoring as specified:
https://github.com/mozilla/pontoon/blob/master/specs/0115-pretranslation-monitoring.md

PR #2927, opened 2023-08-09
Pretranslation monitoring: Time to review pretranslations chart

mathjazz:
Part 1/3 of [#2923](https://github.com/mozilla/pontoon/issues/2923).

This is just the first part of Pretranslation Monitoring, namely the new "[Time to review pretranslations](https://github.com/mozilla/pontoon/blob/master/specs/0115-pretranslation-monitoring.md#new-charts-on-the-insights-tabs)" chart.

I'm opening it in a separate PR, because I found a potential bug in the way we calculate monthly "Time to review" averages, which are (in addition to the new chart) also used in the "Time to review suggestions" chart, and I'd like to resolve it ASAP.

The following 9 locales have pretranslation enabled ATM:
* zh-TW
* fr
* de
* hu
* id
* it
* sl
* es-AR
* cy

But the "Time to review pretranslations" chart is only showing non-zero values for [id](https://mozilla-pontoon-staging.herokuapp.com/id/insights/).

That is because [monthly averages](https://github.com/mozilla/pontoon/blob/7cc79d6fd393580a3b3aa9c2db55bc8ba856c6d6/pontoon/insights/utils.py#L91) are skewed by the [daily default value of 0](https://github.com/mozilla/pontoon/blob/7cc79d6fd393580a3b3aa9c2db55bc8ba856c6d6/pontoon/insights/models.py#L35). For instance, if a month has 30 days and only 1 pretranslation was reviewed in that month and it was 30 days old, the value plotted on the chart will be 1 day (30 + 0 + 0 ... + 0 / 30).

The monthly averages should only be calculated for days on which pretranslations were actually reviewed. If no pretranslations were reviewed in a month, we should plot 0. @eemeli Do you have objections or an alternative proposal?

The issue is much more obvious in the new chart than in the "Time to review suggestions" chart, because pretranslations are reviewed less often than human suggestions, meaning there are more zeros in a month.


mathjazz:
> The monthly averages should only be calculated for days on which pretranslations were actually reviewed. If no pretranslations were reviewed in a month, we should plot 0.

@flodolo Do you have thoughts on this? I can't remember if we already discussed this in past.

flodolo:
I'm confused. This is the definition from the specs

> Current month: Average age of pretranslations reviewed during the specific month.

In your example

> For instance, if a month has 30 days and only 1 pretranslation was reviewed in that month and it was 30 days old, the value plotted on the chart will be 1 day (30 + 0 + 0 ... + 0 / 30).

I would expect the value to be 30. There is one pretranslation reviewed, and it's 30 days old, so the average is 30 (30/1).

If there are 2 pretranslations reviewed, one 20 days old, one 30 days old, the average would be 25, because (20+30)/2. The number of days in the month is not relevant, the way I interpret the specs.

mathjazz:
Correct.

That's not the current behavior in the Time to review suggestions chart, though (which has the exact same spec).

pontoon/insights/tasks.py
```diff
@@ -500,14 +513,18 @@ def build_activity_charts_data(start_of_today, sync_user):
             else:
                 data["peer_approved"].add(translation)
                 if action["approved_date"]:
-                    data["times_to_review"].append(
-                        action["approved_date"] - action["date"]
-                    )
+                    review_time = action["approved_date"] - action["date"]
+                    data["times_to_review_suggestions"].append(review_time)
```
eemeli, line 517: Would it be more useful to not include the pretranslation suggestions in this data, going forward?

pontoon/insights/templates/insights/widgets/insights.html
```diff
@@ -46,65 +46,94 @@ <h3 class="controls">
...
+                        data-lifespans="{{ unreviewed_lifespans }}"
+                        width="410"
+                        height="160">
+                    </canvas>
+                </div>
+            </div>
+        </section>
+        <section class="half">
+            <div class="block clearfix">
+                <div class="chart-item time-to-review-pretranslations">
```
eemeli, line 108: Why is this gated by `{% if unreviewed_lifespans %}` while `time-to-review-suggestions` isn't?

pontoon/insights/static/js/insights.js
```diff
@@ -245,6 +246,109 @@ var Pontoon = (function (my) {
           },
         });
       },
+      renderTimeToReviewPretranslatons: function () {
+        var chart = $('#time-to-review-pretranslations-chart');
```
eemeli, line 250: Should prefer `const`. Don't we have a linter rule for that?

eemeli (commented):
See questions inline.

pontoon/insights/migrations/0012_time_to_review_pretranslations_data.py
```diff
@@ -0,0 +1,85 @@
...
+
+from datetime import datetime, timedelta
+from django.db import migrations
+from django.db.models import F
+
+
+def populate_time_to_review_pretranslations(apps, schema_editor):
+    ActionLog = apps.get_model("actionlog", "ActionLog")
+
+    actions = (
```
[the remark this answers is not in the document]
mathjazz: I followed the pattern used by the migration 0008, that collects times to review suggestions. Your comment makes sense, though. I suggest we file an issue and address both charts together without blocking this PR (I expect this to be quite an edge case).

pontoon/insights/templates/insights/widgets/insights.html
mathjazz: It's gated by `unreviewed_lifespans`, because we only want to show it on the Locale Insights pages and the check is sufficient.

Are you suggesting we change it into `{% if unreviewed_lifespans and time_to_review_pretranslations %}`?

pontoon/insights/tasks.py
mathjazz: I'd like to say no, but I also think it would be useful to see time to review "human" suggestions only. Maybe there can be a filter for that on the Time to review suggestions chart.

@flodolo ?

flodolo: My gut reaction was that we don't want to differentiate, because a review is a review, independently from the source of the suggestion.

On second thoughts though, it might make sense to have one chart only dedicated to time-to-review for standard suggestions, given we have the equivalent for pretranslations. One chart with a selector would work (not sure how you'd call non-pretranslations though, standard suggestions?)

pontoon/insights/migrations/0012_time_to_review_pretranslations_data.py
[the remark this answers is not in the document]
eemeli: Is this something that _can_ be fixed afterwards? Or would the fix be to effectively overwrite all the historical data with a new migration?

pontoon/insights/templates/insights/widgets/insights.html
eemeli: If the current check is sufficient, there's no need to change things. Was just curious about the discrepancy.

pontoon/insights/migrations/0012_time_to_review_pretranslations_data.py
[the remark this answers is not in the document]
mathjazz: Yes, we'd need a data migration that would fix all the existing Time to review suggestions and future Time to review pretranslations data points.

flodolo:
Looking at the graphs for various locales, I wonder if **hours** might be a better unit of measure than **days**. All graphs but Indonesian are completely, or almost flat.

mathjazz:
> Looking at the graphs for various locales, I wonder if **hours** might be a better unit of measure than **days**. All graphs but Indonesian are completely, or almost flat.

That's mostly caused by the bug mentioned [above](https://github.com/mozilla/pontoon/pull/2927#issue-1843953588). I have now added a [commit](https://github.com/mozilla/pontoon/pull/2927/commits/88029ac724c30df999bb6f47290d57af70c2cca7) that fixes it, so averages should be calculated as expected (only among actual review times). 

I've rebased #2929 to also include that fix and pushed it to stage. Most charts should no longer be flat.

flodolo:
> I've rebased #2929 to also include that fix and pushed it to stage. Most charts should no longer be flat.

it, de look flat, probably because the review happens within the day (but that shouldn't be 0, still).

[id](https://mozilla-pontoon-staging.herokuapp.com/id/insights/) also looks funny?

mathjazz:
> > I've rebased #2929 to also include that fix and pushed it to stage. Most charts should no longer be flat.
> 
> it, de look flat, probably because the review happens within the day (but that shouldn't be 0, still).

It's rounded to the number of days. It was always like that for the Time to review chart.

> [id](https://mozilla-pontoon-staging.herokuapp.com/id/insights/) also looks funny?

Why?

flodolo:
> Why?

Because one of the two graphs stops. In hindsight, I suppose it's because nothing was reviewed that month :-( 

Screenshot 2023-08-14 alle 14 35 44

mathjazz:
> Because one of the two graphs stops. In hindsight, I suppose it's because nothing was reviewed that month :-(

Yes. That is expected.

flodolo:
> It's rounded to the number of days. It was always like that for the Time to review chart.

How complex would it be to store hours, and display days in the chart without rounding up to the day? e.g. store 4 h but displaying 0.2 days on the graph, 400 h → 16.7 days

mathjazz:
That's very simple.

Even better would be a dynamic scale, that adapts to data and uses more human friendly units (4 hours and 20 minutes rather than 0.18 days). I was looking into that but didn't get to a working solution instantly. I'll dig deeper and fallback to your proposal if I don't find a solution quickly.

pontoon/insights/migrations/0012_time_to_review_pretranslations_data.py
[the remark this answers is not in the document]
eemeli: Filed an issue: #2930.

pontoon/insights/static/js/insights.js
eemeli: Am confused by the 👍 reaction. So is anything happening here about this?

pontoon/insights/tasks.py
eemeli: @mathjazz Are you considering applying further changes within this PR, or as potential follow-up work?

mathjazz: No in this PR. We should file an issue, though.

pontoon/insights/static/js/insights.js
mathjazz: Removed. :-)

When we introduced linters and formatters, the main argument was to let the tools style the code and speed up the review process. I suggest we file an issue to implement the linter rule if that's possible and keep that as is until then. There is a lot of vars in our codebase.

eemeli: Added #2931.

eemeli (approved):
Looks fine, provided that follow-on issues are filed for https://github.com/mozilla/pontoon/pull/2927#discussion_r1291551046 and the insights review & potential refactor.

mathjazz:
Filed:
https://github.com/mozilla/pontoon/issues/2937
https://github.com/mozilla/pontoon/issues/2938

PR #2929, opened 2023-08-11
Pretranslation monitoring: Pretranslation quality chart

mathjazz:
Part 2/3 of [#2923](https://github.com/mozilla/pontoon/issues/2923).

This is just the second part of Pretranslation Monitoring, namely the new "[Pretranslation quality](https://github.com/mozilla/pontoon/blob/master/specs/0115-pretranslation-monitoring.md#new-charts-on-the-insights-tabs)" chart.

The PR is based on #2927, the first commit specific to this PR is https://github.com/mozilla/pontoon/pull/2929/commits/e7355c5a326e2c3465efe0ade3f2a0adb4ad3ea8.

It is deployed to stage.


flodolo:
Should that graph be hidden if the locale doesn't have any data?
https://mozilla-pontoon-staging.herokuapp.com/cs/insights/

mathjazz:
I would leave the chart in, because no data is also data. Also, if we hide the chart, we'd then also need to hide "Time to review pretranslations", which would break the UI.

We should however change the default values. 100% and 100 obviously don't work well as defaults for the Approval rate and the chrf++ score. It should be null. (And we can use [spanGaps](https://www.chartjs.org/docs/latest/charts/line.html#line-styling) to prevent line charts from being interrupted by missing data).

mathjazz:
Pushed changes to use `null` as the default value for the Approval rate and the chrf++ score (rather than 100% and 100, which were used before commit [14c9b03](https://github.com/mozilla/pontoon/pull/2929/commits/14c9b03ced877f1ef9c53be1eb86b65c2bb53c84)). That's now also deployed to stage.

We could also add "No data yet." or similar inscription over the chart.

And, we should do something similar with the "Time to review pretranslations" chart. Model changes introduced in 14c9b03ced877f1ef9c53be1eb86b65c2bb53c84 applied to that chart will also help resolve the averages bug pointed out in #2927.

eemeli (commented):
I'm a bit concerned that this pattern (also used in #2927 and for other data previously) is effectively premature optimization. As evidenced by the data-filling migration, we can calculate the required values for these graphs from existing data.

Why do we need to log this data continuously, rather than regenerating it on request? If it's desirable to cache it, why not cache the final results for e.g. 24h? The approach of this and the preceding PR seem much more complex, with equivalent data being generated by two different processes, numerous additional database fields, and a daily timed script. It's eager, when we could be lazy.

mathjazz:
Could you clarify what exactly do you mean by "this pattern"?

eemeli:
In order to present a chart like "pretranslation quality", we need the quality data for the team/project/project-locale's pretranslated messages for the past year. So the "pattern" that I'm referring to is that to get there, we:
1. Define new fields to summarise the data, which need to be added to the db.
2. Define a generator for the historical data for all views for the past year.
3. Fill the summary fields with historical data.
4. Write or extend a cron job task that appends new data to the db daily.
5. When requested, we construct the results based on the daily summary data.

The alternative that I'm suggesting be considered is to instead:
1. On request, generate the current team/project/project-locale's data from the pre-existing ActionLog data.
2. Cache the final computed result for 24 hours, so it doesn't need to be recalculated.

mathjazz:
Definitely idea worth checking out. I'm skeptical about performance of generating data on request, though. ActionLog is pretty slow. And these pages don't get that many views, so a great deal of request would be cold I believe.

Should it prove to be feasible nevertheless, it would be a massive refactor of the Insights module, so out of scope here.

eemeli:
Do you have any numbers for the performance cost of generating the data on request?

mathjazz:
Not for the particular views, but It takes 2-3 minutes to complete the daily Insights generation.

pontoon/insights/tasks.py
```diff
@@ -438,8 +460,22 @@ def get_time_to_review_data(category, activities, locale):
...
 
-def query_activity_actions(start_of_today):
-    """Get actions of the previous day, needed for the Translation and Review activity charts."""
+def get_chrf_score(action):
+    try:
+        approved_translation = Translation.objects.get(
+            entity=action["translation__entity"],
+            locale=action["translation__locale"],
+            approved=True,
+        ).string
```
eemeli, line 469: It seems pretty inefficient to look up each approved translation individually. It's fine for the single-run migration, but couldn't all of the rejected pretranslations' approved versions translations be looked up at once?

pontoon/insights/tasks.py
```diff
@@ -526,6 +575,11 @@ def build_activity_charts_data(start_of_today):
                 data["times_to_review_suggestions"].append(review_time)
                 if action["user"] in pretranslation_users:
                     data["times_to_review_pretranslations"].append(review_time)
+            if action["user"] in pretranslation_users:
+                data["pretranslations_rejected"].add(translation)
+                score = get_chrf_score(action)
+                if score:
+                    data["pretranslations_chrf_scores"].append(score)
```
eemeli, line 582: This ends up assigning the pretranslation quality to the date when it's approved or rejected, rather than its creation date. Would it not be more appropriate to use the date when the pretranslation was created?

mathjazz: That's an interesting point. OTOH, it would be much more difficult to backdate scores to when the pretranslation was created.

@flodolo Should we add a note to the chart tooltip clarifying that the chrF++ data belongs to translations **reviewed** (not created) in a particular month?

flodolo: To be honest, I'm not sure it's necessary to point it out, since all the data refers to the specific month. With that said, adding it to the tooltip works for me.

eemeli: I'd be ok with a tooltip note for now, as long as there's a code comment also added somewhere around here about the current date solution being used not because we think it's good, but because it's convenient. Then if we do end up refactoring insights due to #2938, it'll be much easier to switch to the pretranslation's date.

mathjazz: Is that really the case, though? Maybe the expectation is that the Approval rate line and the chrF++ line are aligned on time? I'll let @flodolo decide what's the actual expectation, because I suspect he'll be the most common use of these charts.

I'll add a code comment and tooltip in any case afterwards (at least the former depends on the expectation).

mathjazz: Tooltip updated. I'll add a code comment if the expectation is to calculate the score for the creation month.

eemeli (approved):

pontoon/insights/tasks.py
flodolo: Here's how I see it:
* Tying a metric to the pretranslation date can help us understand if we make improvements to the system (effect of code changes, retraining models, etc.). 
* Tying a metric to the review date can help us understand (also) what's going on in the locale in a specific period.

Since the quality metric (cHRF++) is an automated score, I think we should tie it to the pretranslation date.

Approval data, on the other hand, is a human score, and it can be influenced by a new person coming in to review for that locale from a specific month, so it makes sense to have it tied to the review date.

PR #2932, opened 2023-08-15
Pretranslation monitoring: Insights page

mathjazz:
Part 3/3 of https://github.com/mozilla/pontoon/issues/2923.

This PR adds the third and final part of Pretranslation Monitoring, namely the new "[Insights](https://github.com/mozilla/pontoon/blob/master/specs/0115-pretranslation-monitoring.md#new-insights-page)" page.

The PR is based on https://github.com/mozilla/pontoon/pull/2929, the first commit specific to this PR is https://github.com/mozilla/pontoon/pull/2932/commits/86283b6235a33ca76eae1691401140684fda5e00, which (together with https://github.com/mozilla/pontoon/pull/2932/commits/e48f0d054767194458bda027f56c22adc4287a0b) is just a preparation for the actual changes introduced in https://github.com/mozilla/pontoon/pull/2932/commits/f4f535535522af8ed1a7556582258aa21174336d.

It is deployed to stage:
https://mozilla-pontoon-staging.herokuapp.com/insights/

TODO:

- [x] Filed an issue to upgrade Chart.js: https://github.com/mozilla/pontoon/issues/2939


mathjazz:
Investigating the funkiness of the sl (and All) charts. Seems like an issue with the X-axis.

mathjazz:
> Investigating the funkiness of the sl (and All) charts. Seems like an issue with the X-axis.

Switched from a built-in time-based scale on an X-axis to a default one and that solved the problem. 

That charts no longer looks like this:

Screenshot 2023-08-15 at 21 44 13

🤯 

Also, we should upgrade the chart library. Will file a bug for that.

flodolo:
No familiar with the chart library, but would it be possible to have a "select/deselect all" in the legend (or have the "double click = disable all but this" behavior)?

mathjazz:
Yeah, took me about a second of using the chart to had the same wish. :)

We can render the legend manually in a custom div, which allows for such customization. It will also allow us to align both charts horizontally and enable to scale better when more locales / projects are added. I'll give it a shot.

mathjazz:
> We can render the legend manually in a custom div, which allows for such customization. It will also allow us to align both charts horizontally and enable to scale better when more locales / projects are added. I'll give it a shot.

Done. Cmd + Click or Alt + Click shows only the clicked item. Let me know how it feels.

flodolo:
> Done. Cmd + Click or Alt + Click shows only the clicked item. Let me know how it feels.

Works great. Doesn't feel like it needs a "select all", since you can always reload the page to go back.

pontoon/insights/static/js/insights_charts.js
```diff
@@ -40,6 +40,60 @@ var Pontoon = (function (my) {
         Chart.defaults.global.datasets.bar.barPercentage = 0.7;
         Chart.defaults.global.datasets.bar.categoryPercentage = 0.7;
       },
+      // Legend configuration doesn't allow for enough flexibility,
+      // so we build our own legend
+      // eslint-disable-next-line no-unused-vars
+      customLegend: function (chart) {
+        return function (chart) {
+          function renderLabels(chart) {
+            return chart.data.datasets
```
eemeli, line 49: This is why we have lambdas. Also, a code comment would be Really Nice for why this wrapping is required.

```suggestion
      customLegend: (chart) => (chart) => (chart) => chart.data.datasets
```

pontoon/insights/static/js/insights_charts.js
```diff
@@ -40,6 +40,60 @@ var Pontoon = (function (my) {
...
+                  disabled +
+                  '">' +
+                  '<i class="icon" style="background-color:' +
+                  color +
+                  '"></i>' +
+                  '<span class="label">' +
+                  dataset.label +
+                  '</span>' +
+                  '</li>'
+                );
```
eemeli, line 65: This is why we have template literals. Same comment applies to other string concatenations here. If the lines get too long, you can use `\` to escape newlines that are then ignored.
```suggestion
                return (
                  `<li class="${disabled}"><i class="icon" style="background-color:${color}"></i><span class="label">${dataset.label}</span></li>`
                );
```

pontoon/insights/static/js/insights_tab.js
```diff
@@ -0,0 +1,953 @@
...
+                  const human = chart.datasets[1].data[items.index];
+                  const machinery = chart.datasets[2].data[items.index];
+
+                  const label = chart.datasets[items.datasetIndex].label;
+                  const value = items.yLabel;
+                  const base = label + ': ' + nf.format(value);
+
+                  switch (label) {
+                    case 'Completion':
+                      return base + '%';
```
eemeli, line 457: I don't remember the history here, but there is no need manually format percentages like this; that's why there's the `style: 'percent'` option on `Intl.NumberFormat`.

pontoon/insights/static/js/insights_tab.js
```diff
@@ -0,0 +1,953 @@
...
+                  gridLines: {
+                    display: false,
+                  },
+                  ticks: {
+                    beginAtZero: true,
+                    max: 100,
+                    stepSize: 20,
+                    callback: function (value) {
+                      return value + ' %';
+                    },
```
eemeli, line 511: Using `Intl.NumberFormat` would make sure that we don't end up with differences in how we format percentage values like this: There's a space here, while there isn't in the previous.

pontoon/insights/static/js/insights_tab.js
```diff
@@ -0,0 +1,953 @@
...
+              },
+              callbacks: {
+                label: function (items, chart) {
+                  const label = chart.datasets[items.datasetIndex].label;
+                  const value = items.yLabel;
+                  const base = label + ': ' + nf.format(value);
+
+                  switch (label) {
+                    case 'Approval rate':
+                      return base + '%';
```
eemeli, line 862: Yet another place for the percent formatter.

pontoon/insights/static/js/insights_tab.js
```diff
@@ -0,0 +1,953 @@
...
+                  gridLines: {
+                    display: false,
+                  },
+                  ticks: {
+                    beginAtZero: true,
+                    max: 100,
+                    stepSize: 20,
+                    callback: function (value) {
+                      return value + ' %';
+                    },
```
eemeli, line 908: More manual percent formatting, again with the leading space.

pontoon/insights/static/js/insights_tab.js
eemeli, line 1: Is this renamed from elsewhere? In any case, this file shows up as new when looking at this PR's range of commits, so I looked at it. I have a suspicion I'm rehashing a previous conversation with some of the comments here.

eemeli (changes requested):
Mostly minor JS issues, see inline. Otherwise fine.

pontoon/insights/static/js/insights_tab.js
mathjazz: The file has been renamed, yeah.

pontoon/insights/static/js/insights_charts.js
mathjazz: BTW, this block was also copied from elsewhere.

mathjazz:
Fixed percentages in https://github.com/mozilla/pontoon/pull/2932/commits/db320558e280816c6c2733229753f849c90949e2.

Honestly, I'm not sure I like this approach better, because the values are actual percentages.

pontoon/insights/static/js/insights_charts.js
```diff
@@ -43,25 +43,13 @@ var Pontoon = (function (my) {
       // Legend configuration doesn't allow for enough flexibility,
       // so we build our own legend
       // eslint-disable-next-line no-unused-vars
-      customLegend: function (chart) {
-        return function (chart) {
-          function renderLabels(chart) {
-            return chart.data.datasets
+      customLegend: (chart) => (chart) => (chart) => chart.data.datasets
```
eemeli, line 46: Please review this in an editor. My earlier code suggestions were intended as indicative, not complete. At least this function will now produce a syntax error, and the other string concatenations are untouched. Also, I'd missed on the first pass that the function returned from the outermost function internally calls the innermost one, so that at least should be factored out.

eemeli (changes requested):

eemeli:
Would recommend that you spin off the "Upgrade Chart.js" todo item into a separately tracked issue, so it's not blocking this PR.

eemeli (changes requested):

eemeli (approved):

Issue #2936, opened 2023-08-15
Add ability to switch between themes


ayanaar:
In the concluding step of the light theme implementation process, a feature will be introduced to allow users to seamlessly toggle between the existing dark theme and the newly introduced light theme.

Issue #2935, opened 2023-08-15
Create a new light-theme.css file


ayanaar:
For the third step in the theme implementation process, a new light-theme.css file will be created to house the variables and styles specific to the light theme.

Issue #2934, opened 2023-08-15
Store all extracted CSS variables into a dark-theme.css file


ayanaar:
As the second step in the light theme implementation process, all extracted CSS variables will be consolidated and stored in a dedicated dark-theme.css file.

Issue #2933, opened 2023-08-15
Replace hard-coded color values with CSS variables


ayanaar:
Replace all hardcoded color values in Pontoon with CSS variables to facilitate the upcoming light theme implementation. The variables will be stored in dedicated CSS files, one for each theme.

PR #2953, opened 2023-09-09
Extracting & Replacing Hard-Coded CSS Color Variables for Theme Support

ayanaar:
This PR addresses two key steps in the process of implementing theme support for Pontoon:

1. **CSS Variable Introduction (Issue #2933)**:
   - All hardcoded color values in Pontoon have been replaced with CSS variables. This change facilitates the upcoming introduction of a light theme for Pontoon.
   - By converting colors to variables, we've made the codebase more maintainable, readable, and primed for theming.

2. **Consolidation of CSS Variables (Issue #2934)**:
   - All the extracted CSS variables have been consolidated and stored in a dedicated `dark-theme.css` file. This structure not only provides a centralized location for theme-related colors but also lays the groundwork for the addition of more themes in the future.

### Additional Notes:
   - This change is foundational for the upcoming light theme implementation. By structuring our CSS around variables, we can ensure a smoother transition and switch between themes, enhancing the user experience.
   - **Future Work**:
     - **Theme Toggling Feature (Issue #2936)**: In the concluding phase, we'll introduce a feature allowing users to seamlessly toggle between the existing dark theme and the newly introduced light theme.


mathjazz:
Thanks for the patch!

Could you please rebase against `master`? Looks like there are multiple commits in this PR and only the last one is relevant.

pontoon/static/css/dark-theme.css
```diff
@@ -0,0 +1,68 @@
+/* Dark Theme Variables */
+:root {
+  --black-1: #1c1e21;
+  --black-2: #000000;
+  --black-3: #000;
```
mathjazz, line 5: Note that to convert a 3-value hex syntax to a 6-value syntax, each value needs to be duplicated:
https://developer.mozilla.org/en-US/docs/Web/CSS/hex-color#syntax

So:
* `#000` is the same as `#000000`
* `#FFF` is the same as `#FFFFFF`
* `#123` is the same as `#112233`
* etc.

I suggest we use the six-value (or eight-value when _alpha_ is requried) hex syntax everywhere.

pontoon/static/css/dark-theme.css
```diff
@@ -0,0 +1,68 @@
...
+  --black-4: #272a2f;
+  --black-5: rgba(0, 0, 0, 0.8);
```
mathjazz, line 7: Let's use the hex syntax everywhere:
https://rgbacolorpicker.com/rgba-to-hex

pontoon/static/css/dark-theme.css
```diff
@@ -0,0 +1,68 @@
...
+  --black-6: black;
```
mathjazz, line 8: Same as above, let's remove this on behalf of `--black-2`.

pontoon/static/css/dark-theme.css
```diff
@@ -0,0 +1,68 @@
...
+  --black-7: rgb(39, 42, 47);
```
mathjazz, line 9: That's the same as  [#272a2f](https://www.rgbtohex.net/), which is already defined above.

mathjazz:
Superceeded by #2957.

PR #2957, opened 2023-09-12
Extract and replace all dark theme color variables

ayanaar:
This PR addresses two key steps in the process of implementing theme support for Pontoon:

1. **CSS Variable Introduction (Issue #2933)**:
   - All hardcoded color values in Pontoon have been replaced with CSS variables. This change facilitates the upcoming introduction of a light theme for Pontoon.
   - By converting colors to variables, we've made the codebase more maintainable, readable, and primed for theming.

2. **Consolidation of CSS Variables (Issue #2934)**:
   - All the extracted CSS variables have been consolidated and stored in a dedicated `dark-theme.css` file. This structure not only provides a centralized location for theme-related colors but also lays the groundwork for the addition of more themes in the future.

### Additional Notes:
   - This change is foundational for the upcoming light theme implementation. By structuring our CSS around variables, we can ensure a smoother transition and switch between themes, enhancing the user experience.
   - **Future Work**:
     - **Theme Toggling Feature (Issue #2936)**: In the concluding phase, we'll introduce a feature allowing users to seamlessly toggle between the existing dark theme and the newly introduced light theme.


translate/dark-theme.css
mathjazz, line 1: I would keep the file in the original location. We also keep some other files used in both pontoon and translate in the pontoon.base module.

error_pages/css/blockrain.css
mathjazz, line 1: We should not be making changes to the `error_pages` folder, because these pages are hosted elsewhere and will not be controlled by the user theme settings.

pontoon/base/static/css/nprogress.css
```diff
@@ -70,5 +70,4 @@
 @keyframes nprogress-spinner {
   0%   { transform: rotate(0deg); }
   100% { transform: rotate(360deg); }
-}
-
+}
```
mathjazz, line 73: Please restore.

pontoon/teams/static/css/multiple_team_selector.css
```diff
@@ -1,6 +1,7 @@
 /**
  * All styles are related to forms visible for the contributor.
  */
+
```
mathjazz, line 4: Just an observation: I thought prettier would catch such changes.

pontoon/insights/static/css/insights.css
mathjazz, line 1: Something is off in both insights.css files, perhaps it was caused by a bad merge.

Please make sure we only replace color values with variables and don't add any other code.

error_pages/css/blockrain.css
```diff
@@ -160,4 +160,3 @@
   font-size: 18px;
   margin-bottom: 30px;
 }
-
```
mathjazz, line 163: No need for this change.

error_pages/css/blockrain.css
```diff
@@ -159,5 +159,4 @@
 .blockrain-start-msg, .blockrain-game-over-msg {
   font-size: 18px;
   margin-bottom: 30px;
-}
-
+}
```
mathjazz, line 162: Please make sure the file is exactly the same as before this PR.

pontoon/base/static/css/nprogress.css
mathjazz: Note that this still needs to be addressed.

mathjazz:
This works very well!

The only page where I see the CSS isn't applied is the login page on localhost:
http://localhost:8000/accounts/standalone-login/?next=/

It is powered by the Django templates (rather than Jinja2 like most other views), so we probably need to load the theme file in [the pontoon/base](https://github.com/mozilla/pontoon/blob/master/pontoon/base/templates/django/base.html).

mathjazz:
We should also de-hardcode CSS colors in JS files.

error_pages/css/blockrain.css
mathjazz: This file is still not looking good... There should be now changes made in it as part of this PR.

pontoon/contributors/static/js/profile.js
```diff
@@ -37,7 +37,12 @@ var Pontoon = (function (my) {
         var ctx = chart[0].getContext('2d');
 
         var gradient = ctx.createLinearGradient(0, 0, 0, 160);
-        gradient.addColorStop(0, '#7BC87633');
+        gradient.addColorStop(
+          0,
+          getComputedStyle(document.documentElement)
+            .getPropertyValue('--dark-green')
+            .trim(),
```
mathjazz, line 44: We should declare `style = getComputedStyle(document.documentElement);` only once per file, and then reference colors just by using `style.getPropertyValue('--dark-green')` and the likes.

We should drop [`trim()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim). Is doesn't make a functional difference, does it?

pontoon/insights/static/js/insights.js
```diff
@@ -28,19 +28,19 @@ var Pontoon = (function (my) {
...
-          '#ffbed1',
-          '#9c9290',
-          '#c5bfbe',
-          '#9b93c9',
-          '#ddb5d5',
-          '#7c7270',
-          '#f498b6',
-          '#b173a0',
-          '#c799bc',
+          'getComputedStyle(document.documentElement).getPropertyValue("--white-1").trim()',
```
mathjazz, line 31: We should remove the outer quotes, otherwise http://localhost:8000/insights/ doesn't really use these colors.

mathjazz (approved):
Great work!

As discussed, we'll be making further changes within the `dark-theme.css` file as part of #2935.

PR #2959, opened 2023-09-18
Re-render EditField on readonly status change

### LINKED ISSUE #2955: Impossible to edit the content in translate view (editor has contenteditable set to false)
I believe it only happens when I click a notification from the Pontoon add-on, but I'm not 100% sure if that's the only case. This happens randomly, and I can't spot anything in the browser console to explain the issue.

This is a screenshot of the browser with DevTools showing how the editor has `contenteditable=false`. Setting it to true is also not enough (cursor appears, but still can't type).

editor

eemeli:
Hopefully fixes #2955 

There was an obvious place where a dependency declaration was missed. The original problem does not replicate for me locally, but this should be safe to deploy and then observe a lack of the issue recurring.

### DIFF
--- a/translate/src/modules/translationform/components/EditField.tsx
+++ b/translate/src/modules/translationform/components/EditField.tsx
@@ -44,39 +44,42 @@ export const EditField = memo(
       const keyHandlers = useKeyHandlers();
       const [view, setView] = useState<EditorView | null>(null);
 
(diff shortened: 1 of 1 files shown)


mathjazz (approved):
Looks good.

Looks like we don't have any unit tests here?

Issue #2021, opened 2015-09-10
Replace custom administration interface


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1203788](https://bugzilla.mozilla.org/show_bug.cgi?id=1203788)
Bug Reporter: @Osmose
CC:   @mathjazz

Currently our admin interface is entirely custom-made and is a bit of a mess. Functionally it works fine, but modifying or adding new fields is difficult. Not to mention we have a bunch of code to support widgets that already exist in the built-in Django admin.

We should figure out what things our interface needs that the Django admin doesn't provide, and figure out how difficult it would be to add support for them. Off the top of my head, the only thing missing would be the sync button on the repo field for manually triggering project sync, and adding that in wouldn't be terribly hard.

Oh, and the fact that the built-in admin is kinda ugly and doesn't support easy dashboards, but those are both fixable as well (either manually or by libraries).

mathjazz: Thoughts?

PR #2971, opened 2023-10-02
Add theme selector mockups

mathjazz:
Add mockups for the theme selector to the light theme spec.


flodolo (commented):
One major concern: do we still have plans to localize Pontoon in other languages? If so, the current design for the menu selector is not going to work.

The buttons should also have a tooltip to provide more information, e.g. `Use a dark/light theme in Pontoon`, or `Use a them that matches your system settings`.

For the settings, I feel like the order of selection and explanation should be flipped (you read, then you select).

mathjazz:
> One major concern: do we still have plans to localize Pontoon in other languages? If so, the current design for the menu selector is not going to work.

Added more whitespace. If needed, we can also increase the menu width before we start using localization.

> The buttons should also have a tooltip to provide more information, e.g. `Use a dark/light theme in Pontoon`, or `Use a them that matches your system settings`.

That will not be visible on the mobile UI, so I've also added text above the buttons. I'm not sure it necessary though, especially because the price you pay to click on the wrong button is very low.

> For the settings, I feel like the order of selection and explanation should be flipped (you read, then you select).

Help text position is consistent with other fields and buttons.

flodolo:
> Help text position is consistent with other fields and buttons.

Usually there is a label before a field (left ⇾ right equals top ⇾ bottom). What's below some of them is an additional explanation, but there's already sufficient information in the label itself.

To me, this text feels like the equivalent of the label, not an additional description.

mathjazz:
> but there's already sufficient information in the label itself.

In this case it's best if we just remove it.

flodolo:
> In this case it's best if we just remove it.

I don't understand: are you suggesting removing the description displayed in the current image or something else?

mathjazz:
Yeah.

flodolo:
Personally, I think the description is important (I also never realized how visually confusing I find the selectors for radio buttons).

Let's see if someone else can break the stall.

mathjazz:
I thought by saying "there's already sufficient information in the label itself" you meant that it's not really relevant?

flodolo:
> I thought by saying "there's already sufficient information in the label itself" you meant that it's not really important?

I wonder if the confusion might be due to the fact that I don't consider the current section title (`Appearance`) as a label, because that's not the role section titles have in the rest of the Settings page (they group settings, and each setting has label + optional description).

You wrote:

> Help text position is consistent with other fields and buttons.

My counterargument is that, while such "help text" is placed under a field, the field also has a label that already explains what the field is about. And the label comes before the field. Example:

Screenshot 2023-10-03 alle 11 56 37

In your proposal for settings there is no label, there is only a help description displayed under the selector. I think that description should be above the selector, and basically act as a label. As additional food for thoughts, imagine adding a second setting in this section for "Editor's appearance" (e.g. I want Pontoon Dark but not the editor). How would that work?

BTW, the current menu mock-up uses the same order (description followed by selector).

mathjazz:
OK, so now I think by saying "there's already sufficient information in the label itself" you were talking about the general tune of the Settings page. And not the theme selector specifically, which I thought you were talking about (by "label" I understood the label in the theme buttons). 😄 

Sadly there is no space for the label consistent with the rest of the fields and I honestly don't want to refactor the whole settings UI just for the sake of the theme selector and before taking #2021. And, we actually do have another widget at the bottom of the settings page with only section title serving as the label. I changed its styling in my latest update.

I'd think about placing the "Editor's appearance" if this actually becomes a feature. As for the Profile menu, it's a different story, because we don't have a section title there and there's no other existing UI piece we could reuse there.

All that being said, knowing we'll be refactoring this page sooner or later, I don't have strong opinions about the position of the help text for the theme selector. It's just a theme selector, it's pretty straightforward and if the user for some reason still clicks the wrong button, they will quickly find a way out trouble.

flodolo (approved):

PR #2972, opened 2023-10-03
Theme Toggling Feature 

ayanaar:
This PR addresses the last step in the process of implementing theme support for Pontoon. It introduces the capability for users to seamlessly switch between the platform's existing dark theme and the newly added light theme (Issue #2936 ). 

### Changes:
- **New UI Element**: Added buttons in the settings allowing users to select their preferred theme: Dark, Light, or System Default.
- **Theme Persistence**: User theme preference is stored in their profile, ensuring a consistent experience across sessions.
- **Dynamic Theme Application**: Leveraging JavaScript, the theme is applied dynamically without requiring a page reload.
- **Fallback Mechanism**: In case of any issues, the system defaults to the dark theme, ensuring no disruption to the user experience.

### Additional Notes:
 - **Future Work**:
     - The light-mode.css file used in this PR is a placeholder.


mathjazz:
Nice work!

Please note that the spec has now been updated with the theme selector mockups:
https://github.com/mozilla/pontoon/commit/96051544b3badd7cd0b31da3d1d58c1281474081

I'll share the corresponding HTML and CSS with you.

We should also change the tooltips as noted in the spec.

pontoon/base/templates/base.html
```diff
@@ -33,9 +33,9 @@
   </head>
 
   <body
-    class="{% block class %}{% endblock %}"
+    class="{% block class %}{% endblock %} {% if request.user.is_authenticated %}{{ request.user.profile.theme }}-theme{% endif %}"
     {% if csrf_token %}data-csrf="{{ csrf_token }}"{% endif %}
-  >
+>
```
mathjazz, line 38: Please revert changes to this line.

pontoon/base/templates/django/base.html
```diff
@@ -18,7 +18,7 @@
     <link rel="stylesheet" href="{% static 'css/fonts.css' %}" title="" type="text/css" />
     <link rel="stylesheet" href="{% static 'css/boilerplate.css' %}" title="" type="text/css" />
     <link rel="stylesheet" href="{% static 'css/style.css' %}" title="" type="text/css" />
-    <link rel="stylesheet" href="{% static 'css/dark-theme.css' %}" title="" type="text/css" />
+
```
mathjazz, line 21: Please revert changes to this line.

pontoon/base/templates/django/base.html
```diff
@@ -121,5 +121,6 @@
     <script type="text/javascript" charset="utf-8" src="{% static 'js/lib/jquery-3.6.1.js' %}"></script>
     <script type="text/javascript" charset="utf-8" src="{% static 'js/lib/nprogress.js' %}"></script>
     <script type="text/javascript" charset="utf-8" src="{% static 'js/main.js' %}"></script>
+    <script type="text/javascript" charset="utf-8" src="{% static 'js/theme-switcher.js' %}"></script>
```
mathjazz, line 124: Please revert changes to this line.

pontoon/contributors/templates/contributors/settings.html
```diff
@@ -64,6 +64,18 @@ <h3>Personal information</h3>
         </div>
       </section>
 
+   <section class="appearance" >
+        <h3>Appearance</h3>
+        <div class="field">
+            <p class="help">Choose if Pontoon’s appearance should be dark, light, or follow your system’s settings</a></p>
```
mathjazz, line 70: Please make sure the text matches the spec (we dropped "Pontoon").

pontoon/contributors/templates/contributors/settings.html
```diff
@@ -64,6 +64,18 @@ <h3>Personal information</h3>
...
+        </div>
+        <span class="toggle-button">
+            <button type="button" class="dark active" title="Use a dark theme in Pontoon"><i class="icon far fa-moon"></i>Dark theme</button>
+            <button type="button" class="light" title="Use a light theme in Pontoon"><i class="icon fa fa-sun"></i>Light theme</button>
+            <button type="button" class="system" title="Use a theme in Pontoon that matches your system settings"><i class="icon fa fa-laptop"></i>Match system</button>
```
mathjazz, line 75: Please make sure the text matches the spec. We dropped "Pontoon". Please also do so with the other two themes.

pontoon/contributors/templates/contributors/settings.html
```diff
@@ -160,4 +172,5 @@ <h3><span class="small stress">Preferred locales to get suggestions from</span><
 
 {% block extend_js %}
   {% javascript 'settings' %}
+  {% javascript 'theme-switcher' %}
```
mathjazz, line 175: The theme-switcher.js is already loaded as part of the base template and we also don't have a 'theme-switcher' defined in PIPELINE_JS, so let's revert this line.

pontoon/contributors/views.py
```diff
@@ -187,6 +187,36 @@ def toggle_user_profile_attribute(request, username):
...
+        )
+
+    theme = request.POST.get("theme", None)
+    allowed_themes = ["light", "dark", "system"]
+
+    if theme not in allowed_themes:
+        return JsonResponse(
+            {"status": False, "message": "Bad Request: Invalid theme"},
+            status=400,
+        )
```
mathjazz, line 211: We should only have themes defined in one places. [Looks like](https://stackoverflow.com/questions/60351804/no-validation-on-field-choices-django-postgres) calling `full_clean()` before `save()` will do the validation "for free"?

pontoon/contributors/views.py
```diff
@@ -250,7 +280,6 @@ def settings(request):
             request.POST,
             instance=profile,
         )
-
```
mathjazz, line 253: Please revert changes to this line.

mathjazz:
Excellent work with the updates! Added a few more minor requests.

pontoon/contributors/urls.py
```diff
@@ -62,6 +62,12 @@ class UsernameConverter(StringConverter):
         views.toggle_user_profile_attribute,
         name="pontoon.contributors.toggle_user_profile_attribute",
     ),
+    # API: Toggle user theme preference
+    path(
+        "api/v1/user/theme/<username:username>/",
```
mathjazz, line 67: We should instead use `"api/v1/user/<username:username>/theme/"` (user identifier should follow the "user").

pontoon/contributors/templates/contributors/settings.html
```diff
@@ -64,6 +64,18 @@ <h3>Personal information</h3>
         </div>
       </section>
 
+   <section class="appearance" >
+        <h3>Appearance</h3>
+        <div class="field">
+            <p class="help">Choose if appearance should be dark, light, or follow your system’s settings</a></p>
+        </div>
+        <span class="toggle-button">
+            <button type="button" class="dark active" title="Use a dark theme"><i class="icon far fa-moon"></i>Dark theme</button>
```
mathjazz, line 73: We need to dynamically set `active` instead of hard-coding it.

Inspiration:
https://github.com/mozilla/pontoon/blob/main/pontoon/base/templates/widgets/toggle.html

mathjazz:
Thanks for another set of updates!

I had a closer look at how the toggle logic actually works. Here's my list of proposals:

1. System theme does not get applied. Themes are applied by setting the corresponding class on the `<body>` element, which results in the use of the right theme file. Since we don't have a `system-theme.css` file, setting `system-theme` on the `<body>` element doesn't apply any of the theme files. If the system theme is set, we should [detect what it is](https://stackoverflow.com/questions/56393880/how-do-i-detect-dark-mode-using-javascript) and set `light-theme` / `dark-theme` class accordingly. It would also be nice to add an event handler (as used in the Stackoverflow) that will change the theme immediately after it is changed in the OS.

1. Changing the theme doesn't seem to apply in the Translate view.

1. In order to prevent the error triggered on theme switch, you'll need to make the selector at https://github.com/mozilla/pontoon/blob/main/pontoon/contributors/static/js/settings.js#L3 more specific, to make sure it doesn't apply to the theme selector.

1. Changes to the `specs/` folder need to be excluded from the PR. You'll need to rebase against the `main` branch.

pontoon/base/static/js/themeSwitcher.js
```diff
@@ -0,0 +1,77 @@
+$(function () {
+  function getSystemTheme() {
+    if (
+      window.matchMedia &&
+      window.matchMedia('(prefers-color-scheme: dark)').matches
+    ) {
+      return 'dark';
+    } else {
+      return 'light';
```
mathjazz, line 9: We should default to dark, not light. So if the user sets "System" in Pontoon and "Auto" on macOS, she should get the dark theme. Right now she gets the light theme.

pontoon/base/static/js/themeSwitcher.js
```diff
@@ -0,0 +1,77 @@
...
+    }
+  }
+
+  function updateTheme(newTheme) {
```
mathjazz, line 13: Maybe `applyTheme()` is more accurate?

pontoon/base/static/js/themeSwitcher.js
```diff
@@ -0,0 +1,77 @@
...
+    if (newTheme === 'system') {
+      newTheme = getSystemTheme();
+    }
+    $('body')
+      .removeClass('dark-theme light-theme system-theme')
+      .addClass(`${newTheme}-theme`);
+
+    if (newTheme !== 'system') {
+      Pontoon.endLoader(`Theme changed to ${newTheme}.`);
```
mathjazz, line 22: I would move the notification down to the `ajax.success()` call and call it regardless of the theme being set (i.e. remove the condition).

pontoon/base/static/js/themeSwitcher.js
```diff
@@ -0,0 +1,77 @@
...
+    }
+  }
+
+  window
+    .matchMedia('(prefers-color-scheme: dark)')
+    .addEventListener('change', function (e) {
+      if ($('.appearance .toggle-button button.system').hasClass('active')) {
```
mathjazz, line 29: This will work on the /settings page. But what if another page is opened when the user changes system settings? We could for example store the actual user theme setting in a data-theme attribute on the body element and use that to check if the system theme is enabled.

pontoon/base/static/js/themeSwitcher.js
```diff
@@ -0,0 +1,77 @@
...
+        if (e.matches) {
+          updateTheme('dark');
+        } else {
+          updateTheme('light');
+        }
+      }
+    });
+
+  $(document).ready(function () {
```
mathjazz, line 38: We are already in the `$(document).ready()` as of the first line of this file, so you can remove this wrapper:
https://learn.jquery.com/using-jquery-core/document-ready/

pontoon/base/static/js/themeSwitcher.js
```diff
@@ -0,0 +1,77 @@
...
+  $('.appearance .toggle-button button').click(function (e) {
+    e.preventDefault();
+
+    var self = $(this);
+
+    if (self.is('.active')) {
+      return;
+    }
+
+    var theme = self.attr('class').split(' ')[0];
```
mathjazz, line 54: This looks fragile. I suggest we also put values in button `value` attributes and then just use `self.val()` here.

pontoon/base/templates/base.html
```diff
@@ -33,9 +33,10 @@
   </head>
 
   <body
-    class="{% block class %}{% endblock %}"
+    class="{% block class %}{% endblock %} {% if request.user.is_authenticated %}{% if user.profile.theme == 'system' %}system{% else %}{{ user.profile.theme }}{% endif %}-theme{% endif %}"
```
mathjazz, line 36: Isn't this the same as this?

```jinja2
<body class="{% block class %}{% endblock %} {% if request.user.is_authenticated %}{{ user.profile.theme }}-theme{% endif %}">
```

pontoon/settings/base.py
```diff
@@ -515,11 +516,12 @@ def _default_from_email():
             "js/lib/jquery.color-2.1.2.js",
             "js/lib/nprogress.js",
             "js/main.js",
+            "js/themeSwitcher.js",
```
mathjazz, line 519: We use kebab-case filenames for static files in `/pontoon`.

pontoon/settings/base.py
```diff
@@ -515,11 +516,12 @@ def _default_from_email():
...
         ),
         "output_filename": "js/base.min.js",
     },
     "translate": {
-        "source_filenames": ("translate.js",),
+        "source_filenames": ("translate.js", "js/themeSwitcher.js"),
```
mathjazz, line 524: We don't need that code in the translate view. We should re-implement the logic in React.

pontoon/translate/views.py
```diff
@@ -64,6 +68,7 @@ def translate(request, locale, project, resource):
     context = {
         "locale": get_preferred_locale(request),
         "notifications": [],
+        "user_theme": user_theme,
```
mathjazz, line 71: This will fail if the user is not authenticated. Do we even use it?

translate/public/translate.html
```diff
@@ -16,15 +16,18 @@
     <title>Pontoon</title>
 
     <!-- Inlining CSS mitigates/prevents fouc. -->
-    <style> body { background: #3f4752 !important; } </style>
+    <style> body.dark-theme { background: #3f4752 !important; } </style>
```
mathjazz, line 19: What is the purpose of this change?

translate/public/translate.html
```diff
@@ -16,15 +16,18 @@
...
     <link href="/static/css/fontawesome-all.css" rel="stylesheet" />
     <link href="/static/css/boilerplate.css" rel="stylesheet" />
     <link href="/static/css/fonts.css" rel="stylesheet" />
+    <link href="/static/css/dark-theme.css" rel="stylesheet" />
+    <link href="/static/css/light-theme.css" rel="stylesheet" />
```
mathjazz, line 24: Why not include these files in PIPELINE_CSS? We already include dark-theme.css there.

translate/public/translate.html
```diff
@@ -16,15 +16,18 @@
...
     {% stylesheet 'translate' %}
 
     {% include "tracker.html" %}
   </head>
-  <body>
+  <body class="{% block class %}{% endblock %} {% if request.user.is_authenticated %}{% if user.profile.theme == 'system' %}system{% else %}{{ user.profile.theme }}{% endif %}-theme{% endif %}"
```
mathjazz, line 29: See the base.html on how we could possible simplify this line.

translate/public/translate.html
```diff
@@ -16,15 +16,18 @@
...
+  {% if csrf_token %}data-csrf="{{ csrf_token }}"{% endif %}>
```
mathjazz, line 30: No need for the `csrf_token` here.

translate/public/translate.html
```diff
@@ -34,6 +37,6 @@
         data-notifications="{{ notifications | to_json }}"
         translate="no"
     ></div>
-    {% javascript 'translate' %}
+    {% javascript 'translate' %}
```
mathjazz, line 40: Please revert this change.

pontoon/base/static/js/theme-switcher.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  function applyTheme(newTheme) {
+    if (newTheme === 'system') {
+      newTheme = getSystemTheme();
+    }
+    $('body')
+      .removeClass('dark-theme light-theme system-theme')
+      .addClass(`${newTheme}-theme`)
+      .data('theme', newTheme);
```
mathjazz, line 20: We should move this line to the `ajax.success` function. `applyTheme()` is also called from the event handler, which should not set the data attribute.

mathjazz: And, we should also set the `body[data-theme]` attribute in the templates, otherwise changing the system theme in the OS doesn't work.

pontoon/base/static/js/theme-switcher.js
```diff
@@ -0,0 +1,79 @@
+$(function () {
+  function getSystemTheme() {
+    if (
+      window.matchMedia &&
+      window.matchMedia('(prefers-color-scheme: light)').matches
```
mathjazz, line 5: We should change this back to dark (or switch the returns). Sorry for misleading you - the Auto setting on macos changes based on time and has no effect here.

pontoon/contributors/templates/contributors/settings.html
```diff
@@ -64,6 +64,18 @@ <h3>Personal information</h3>
         </div>
       </section>
 
+   <section class="appearance" >
```
mathjazz, line 67: Mind the extra space at the end.

pontoon/base/templates/base.html
```diff
@@ -33,8 +33,9 @@
   </head>
 
   <body
-    class="{% block class %}{% endblock %}"
+    class="{% block class %}{% endblock %} {% if request.user.is_authenticated %}{{ user.profile.theme }}-theme{% endif %}"
     {% if csrf_token %}data-csrf="{{ csrf_token }}"{% endif %}
+    data-theme="{% if user.profile.theme %}{{ user.profile.theme }}{% else %}system{% endif %}">
```
mathjazz, line 38: Mind the > at the end of the line.

mathjazz (commented):
Just a couple of comments left!

What is the status of applying the theme in the translate view?

Please also look into the failing tests.

mathjazz:
Just noticed that the insights pages (e.g. http://localhost:8000/sl/insights/ and http://localhost:8000/projects/firefox/insights/ appear to be broken (JS error).

mathjazz:
> Just noticed that the insights pages (e.g. http://localhost:8000/sl/insights/ and http://localhost:8000/projects/firefox/insights/ appear to be broken (JS error).

To fix that, we should use `document.body` instead of `document.documentElement` in places like:
https://github.com/mozilla/pontoon/blob/main/pontoon/insights/static/js/insights_tab.js#L7

And also get rid of `trim()` after `style.getPropertyValue()`.

pontoon/base/templates/base.html
```diff
@@ -33,9 +33,9 @@
   </head>
 
   <body
-    class="{% block class %}{% endblock %}"
+    class="{% block class %}{% endblock %} {% if request.user.is_authenticated %}{{ user.profile.theme }}{% else %}dark{% endif %}-theme"
```
mathjazz, line 36: We now repeat this very same line 4 times in total:
```
{% if request.user.is_authenticated %}{{ user.profile.theme }}{% else %}dark{% endif %}
```

It would be nice to factor it out, perhaps into a [custom template tag](https://github.com/mozilla/pontoon/blob/05b851058a0f33d5fc60ef528471d9490898bac0/pontoon/base/templatetags/helpers.py#L30-L36), so we'd just call something like` {{ theme(request.user) }}`.

translate/src/context/Theme.tsx
```diff
@@ -16,36 +16,44 @@ function getSystemTheme() {
...
 
 export function ThemeProvider({ children }: { children: React.ReactElement }) {
-  const [theme] = useState(() => {
-    const currentTheme = document.body.getAttribute('data-theme') || 'system';
-    if (currentTheme === 'system') {
-      return getSystemTheme();
-    }
-    return currentTheme;
-  });
+  const [theme] = useState(() => document.body.getAttribute('data-theme')!);
```
mathjazz, line 19: Why is there `!` at the end of the line here and in a couple of places below? There are warning if you run `make eslint`.

translate/src/context/Theme.tsx
```diff
@@ -16,36 +16,44 @@ function getSystemTheme() {
...
-        document.body.getAttribute('data-theme') || 'system';
+      let userThemeSetting = document.body.getAttribute('data-theme')!;
 
       if (userThemeSetting === 'system') {
-        applyTheme(e.matches ? 'dark' : 'light');
+        if (e.matches) {
+          applyTheme('dark');
+        } else {
+          applyTheme('light');
+        }
```
mathjazz, line 43: Actually what I wanted to say is that I like the previous version and we should use it in both files. :-)))

translate/src/context/Theme.tsx
```diff
@@ -16,36 +16,44 @@ function getSystemTheme() {
...
     mediaQuery.addEventListener('change', handleThemeChange);
-
-    applyTheme(theme);
+    if (document.body.classList.contains('system-theme')) {
+      let systemTheme = getSystemTheme();
+      document.body.classList.remove('system-theme');
+      document.body.classList.add(`${systemTheme}-theme`);
+    } else {
+      applyTheme(document.body.getAttribute('data-theme')!);
+    }
```
mathjazz, line 56: Could you explain why did you make this change?

mathjazz:
Nice work with the updates!

* I've kicked the pytest - I only have 3 failing tests locally.
* As discussed, there is a JS error on the /profile when the system theme is selected. I'll take a look.

translate/src/context/Theme.tsx
ayanaar: I did this because of the expectation that the data-theme attribute will always be present on the document.body and thus will not return a null. I used it to bypass the TypeScript type-checking as document.body.getAttribute('data-theme') can potentially return null, and I get an error otherwise. Should I explicitly set a default value instead?

translate/src/context/Theme.tsx
ayanaar: Removed, this was a previous iteration of the code.

mathjazz (commented):
Great work - the core functionality is there!

Pytest is failing due to a conflicting migration. We have recently landed an existing migration 0047, so you'll need to rename yours to 0048 and update the dependency migration: [pontoon/base/static/css/light-theme.css](https://github.com/mozilla/pontoon/tree/main/pontoon/base/migrations).

We also need to get rid of the eslint warnings.

translate/src/context/Theme.tsx
mathjazz: Yeah, seems like we'll need the explicit value anyways.

mathjazz (commented):
Excellent job!

Now we need to fix the tests and possibly rebase and we're ready to merge.

mathjazz:
As discussed on Slack, we should probably revert the last patch and remove "theme" from the form.

As for the rebase, I suggest you create a fresh branch out of the current `main`, `git apply` [the patch](https://github.com/mozilla/pontoon/pull/2972.patch), remove the changes in the `specs` folder and push to a new PR.

mathjazz:
Works like a charm! Woohoo! 🥳 

Continuing in #2996.

PR #2975, opened 2023-10-04
Do not break comment layout when pasting multiline strings

### LINKED ISSUE #2974: Pasting a multiline text in the comment input text is broken
Copy and paste a multiline text in the comment box, either in the sidebar or below the editor.

If it's a short text, e.g. two short lines, they will be collapsed into one line. If it's a longer text, the editor will suddenly become very confusing.

Example:

```
This is a line.

This is a second line with much longer text repeated a few times. This is a second line with much longer text repeated a few times. This is a second line with much longer text repeated a few times.
```

Note that I copied the same text, the result changes depending on where I copy it to.

Screenshot 2023-10-04 alle 15 40 39

Screenshot 2023-10-04 alle 15 40 58

mathjazz:
Fix #2974.

### DIFF
--- a/translate/src/modules/comments/components/AddComment.css
+++ b/translate/src/modules/comments/components/AddComment.css
@@ -9,13 +9,12 @@
   border: none;
   border-radius: 4px;
   color: var(--white-1);
-  display: flex;
-  flex-grow: 2;
   font-size: 11px;
   max-height: 144px;
   margin: auto;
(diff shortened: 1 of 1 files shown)


mathjazz:
@flodolo Could you please test on stage?

flodolo (approved):
Seems to work correctly 👍🏼

Issue #2949, opened 2023-09-02
Two strings on Pontoon cannot be approved successfully


Olvcpr423:
### The problem:
There are two strings that cannot be approved successfully:
1. 278410 (AMO project)
2. 263159 (SUMO project)

### The behavior:
- **1st string**: Though you can approve a translation, the translation you approved will automatically turn "rejected" several days later, and the status of the string turns "Unreviewed".
- **2nd string**: Though you can approve a translation, the status of the string will automatically turn "Unreviewed" several days later. However, unlike the 1st string, the translation you approved before won't turn "rejected". ([screenshot](https://drive.google.com/file/d/1mZJNT6WVodjY-8EeyczNG5WvrVykkm01/view))

### Affected languages:
As far as I know, this problem occurs in all languages that words don't have a plural form, including Chinese Simplified ([1st](https://pontoon.mozilla.org/zh-CN/all-projects/all-resources/?string=278410) & [2nd](https://pontoon.mozilla.org/zh-CN/all-projects/all-resources/?string=263159) string), Chinese Traditional ([1st](https://pontoon.mozilla.org/zh-TW/all-projects/all-resources/?string=278410) & [2nd](https://pontoon.mozilla.org/zh-TW/all-projects/all-resources/?string=263159) string) and Vietnamese ([1st](https://pontoon.mozilla.org/vi/all-projects/all-resources/?string=278410) & [2nd](https://pontoon.mozilla.org/vi/all-projects/all-resources/?string=263159) string) (Japanese and Korean haven't translated these strings yet).

mathjazz:
It turns out this is actually a duplicate of #2744.

PR #2982, opened 2023-10-10
Fix gettext plural forms for non-plural locales

### LINKED ISSUE #2744: Pontoon fails to approve string translation
Pontoon: https://pontoon.mozilla.org/zh-TW/sumo/all-resources/?string=263159
String repo: https://github.com/mozilla-l10n/sumo-l10n/blob/master/zh_TW/LC_MESSAGES/django.po#L3313-L3317

# STR
1. Open the string on Pontoon.
2. Submit and approve the translation (there's already one with 100% TM match)
3. Reload Pontoon page for the same string

# Expected 
The translation was saved and approved correctly

# Actual 
1. The string seems to be not approved in the translation area, however it is approved in the submission area.
1. By submitting the translation with the existing translation using Enter key, "Same translation already exist" displays every time.
2. By un-approving and approving the string again, the string seems to be approved. However refreshing the page the string became unapproved again.

eemeli:
Fixes #2744 

Includes an SQL data migration to fix the ~100 instances of bad data in the pontoon.mozilla.org DB. Most of those were not actually due directly to the parent issue, but the AMO string `Version {0}` getting a plural form added to it about half a year ago.

After this lands, previously suggested and auto-rejected translations should be re-submitted. As the migration is non-reversible and I managed to apply it locally before taking note of exactly which ones those may be (in addition to the entities mentioned in #2949), here's at least a query that can list them:

```sql
select entity_id, lc.code, tx.string
from base_translation as tx
join base_locale as lc on locale_id = lc.id
join base_entity as ent on entity_id = ent.id
where tx.plural_form is NULL and ent.string_plural != '';
```


pontoon/base/migrations/0047_fix_pluralform_data.py
```diff
@@ -0,0 +1,24 @@
...
+        migrations.RunSQL(
+            sql="""
+UPDATE base_translation AS tx
+SET plural_form = 0
+FROM base_entity AS ent
+WHERE
+  tx.plural_form IS NULL AND
+  tx.entity_id = ent.id AND
+  ent.string_plural != '';
+"""
```
mathjazz, line 22: Please rewrite this in Django ORM. We use SQL as the last resort - in case the query is impossible or much more complex to write in Django ORM, or for performance reasons.

Something like this should work:

```python
Translation.objects.filter(plural_form__isnull=True).exclude(entity__string_plural="").update(plural_form=0)
```

mathjazz (commented):
Thanks!

Could you elaborate why do we need to re-submit previously suggested and auto-rejected translations?

pontoon/base/migrations/0047_fix_pluralform_data.py
eemeli: Can do, but with the caveat that I'm pretty much out of bandwidth atm for Pontoon work given other deadlines. So if I need to rewrite this, that'll _hopefully_ be sometime next week.

eemeli:
> Could you elaborate why do we need to re-submit previously suggested and auto-rejected translations?

Because there are only a few of these, and automating that would be more work than doing it manually.

mathjazz:
Thanks for clarifying.

Leaving the output here for reference:
https://gist.github.com/mathjazz/fc683bf5e5a12ba055e70c483484b049

eemeli:
@mathjazz The migration is now Python. Based on the data dump I still had from early September, these are the messages that ought to be manually reviewed once this lands:

- https://pontoon.mozilla.org/ja/sumo/all-resources/?string=263159
- https://pontoon.mozilla.org/ko/sumo/all-resources/?string=263159
- https://pontoon.mozilla.org/th/amo/all-resources/?string=64906
- https://pontoon.mozilla.org/th/amo-frontend/all-resources/?string=195066
- https://pontoon.mozilla.org/vi/amo/all-resources/?string=278410
- https://pontoon.mozilla.org/vi/sumo/all-resources/?string=263159
- https://pontoon.mozilla.org/zh-CN/amo/all-resources/?string=278410
- https://pontoon.mozilla.org/zh-CN/sumo/all-resources/?string=263159
- https://pontoon.mozilla.org/zh-TW/amo/all-resources/?string=278410
- https://pontoon.mozilla.org/zh-TW/amo/all-resources/?string=64150
- https://pontoon.mozilla.org/zh-TW/sumo/all-resources/?string=263159

PR #2984, opened 2023-10-11
Update Firefox Accounts references to Mozilla Accounts

### LINKED ISSUE #2983: Rename Firefox accounts in Mozilla accounts
There are a few references in the documentation, we should update them.
I think renaming internal references (e.g. fxa) goes beyond the scope, as most products will keep that for the time being.

We also need to make sure to update whatever package is responsible for the text in [this page](https://pontoon.mozilla.org/accounts/fxa/login/)

mathjazz:
Fix #2983.

### DIFF
--- a/pontoon/settings/base.py
+++ b/pontoon/settings/base.py
@@ -956,7 +956,7 @@ ACCOUNT_EMAIL_REQUIRED = True
 ACCOUNT_EMAIL_VERIFICATION = "none"
 ACCOUNT_USER_DISPLAY = account_username
 
-# Mozilla Accounts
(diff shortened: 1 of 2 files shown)


pontoon/settings/base.py
```diff
@@ -956,7 +956,7 @@ def account_username(user):
 ACCOUNT_EMAIL_VERIFICATION = "none"
 ACCOUNT_USER_DISPLAY = account_username
 
-# Firefox Accounts
+# Mozilla Accounts
```
flodolo, line 959: Worth mentioning that the keys use FXA for legacy reasons?

flodolo (approved):

Issue #2987, opened 2023-10-13
Can't type a space at the end of a string (removed automatically)


flodolo:
1. Open this string https://pontoon.mozilla.org/it/firefox-accounts/all-resources/?string=286060
2. Go to the last word (Mozilla), type a space then a character

The space gets automatically removed.

It's pretty bad, because it's easy to miss, and it's hard to understand how to type at the end of the string. But:
* It doesn't happen if I empty the text area and start typing from scratch.
* It doesn't happen if I place the cursor at the very end of the string above (empty line).

CC @eemeli, not sure if this can be related to the latest editor changes.

mathjazz:
I've reverted the latest changes on stage and can no longer reproduce the original issue there.

mathjazz:
I was just doing some translation work and noticed the problem in practice. It is indeed annoying and easy to miss. It was also reported in #pontoon. I'm thinking we should revert it.

Thoughts?

flodolo:
I think we should back out. Another issue, potentially related, since I've never seen anything like that before.

1. Open this string https://pontoon.mozilla.org/it/mozillaorg/all-resources/?status=unreviewed&string=285944
2. Select the 98% match
3. Select `href="` and press delete on the keyboard
4. Instead of the selected text, half of the string gets removed


https://github.com/mozilla/pontoon/assets/3644868/1649d290-d806-4394-ad4c-cb33b388c0c3

flodolo:
Correction: it happens even if I try to delete characters one at a time, it's pretty much impossible to translate in Pontoon at the moment.

mathjazz:
In addition to opening the PR to revert https://github.com/mozilla/pontoon/pull/2926, I've rolled back to the previous deployment on prod (459d4366a5c09449d91bc2eb4533b479304921d1).

PR #2990, opened 2023-10-14
Revert "Improve Pontoon RTL editing (#2926)"

mathjazz:
Temporary fix for #2987:
This reverts commit 5b6af5fcfdf49d0dc0f0ba558a143311f3945fc2.


flodolo (approved):
Maybe this should go through @eemeli on Monday, since we have "fixed" the problem on prod?

PR #2998, opened 2023-10-21
Editor fixes: preserve whitespace, exclude ` ` as a printf conversion flag, match brackets better

### LINKED ISSUE #2987: Can't type a space at the end of a string (removed automatically)

### LINKED ISSUE #2988: Bad placeable detection in the editor
Space and "d" are detected and highlighted as part of the placeable in the following string:
https://mozilla-pontoon-staging.herokuapp.com/sl/all-projects/all-resources/?string=66620

Screenshot 2023-10-13 at 10 15 15

It's also interesting that the space following "%" is wider than the one preceding it.

eemeli:
Fixes #2987
Fixes #2988
Pre-empts #2990

The space issue was due to using an incorrect CSS `white-space: pre-line` value for plain text, which was collapsing whitespace and thereby messing up everything: the right value here is `pre-wrap`. I could replicate this locally, but it doesn't show up in tests (which are finally added here for EditField) as that browser feature is not emulated by JSDOM.

The odd editing experience when removing one of the `""` in a tag property value is fixed by making property values also `pre-wrap`, and bailing out of literals on `}` and `>`.

The printf regexp is updated as discussed in the issue.

I also included some improvements here for how paired syntax characters like `{}` and `""` are handled. The matching pair highlighting is dropped, as that's a bit too buggy with some RTL messages.

All of the above should let us re-deploy the main branch including the #2926 changes once this is merged, and close #2990 without merging.

### DIFF
--- a/translate/src/modules/translationform/components/EditField.test.tsx
+++ b/translate/src/modules/translationform/components/EditField.test.tsx
@@ -1,7 +1,7 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
 
 import * as Fluent from '@fluent/react';
-import { act, render } from '@testing-library/react';
+import { act, render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import React from 'react';
 import sinon from 'sinon';
@@ -135,5 +135,14 @@ describe('<EditField>', () => {
       ref.current!.setSelection('bar');
     });
     expect(spy.getCalls()).toMatchObject([{ args: [0, 'foobar'] }]);
+  });
+
+  it('does not highlight `% d` as code (#2988)', () => {
+    render(<MockEditField defaultValue='{0}% done' format='plain' />);
+    const placeholder = screen.getByText(/0/);
+    const notPrintf = screen.getByText(/% d/);
+    const certainlyText = screen.getByText(/one/);
+    expect(notPrintf).not.toBe(placeholder);
+    expect(notPrintf).toBe(certainlyText);
   });
 });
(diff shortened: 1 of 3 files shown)


mathjazz:
Thanks!

Nice work with adding the unit tests! I looked very briefly, but it seems like we don't test the fix for the original issue reported in https://github.com/mozilla/pontoon/issues/2987?

Deployed to stage for testing and the issues seem fixed!

@flodolo Could you help test this on stage? You mentioned a couple of other issues in the bug report and I'm not sure I was able to reproduce them.

eemeli:
> Nice work with adding the unit tests! I looked very briefly, but it seems like we don't test the fix for the original issue reported in #2987?

That's because all the bugs reported in that issue are fundamentally caused by CSS, and therefore not testable outside a real browser. We don't currently have any browser tests, and they're not supported by Jest which we're using for all the tests that we do have.

flodolo:
I can't seem to be able to reproduce any of the previous issues on stage.

mathjazz (approved):
Looks good!

PR #2997, opened 2023-10-19
Add light theme CSS file

mathjazz:
Fix #2935. In addition to adding the light theme, this patch also includes consistency changes on the dark theme.

**Note**: after deployment, a small change is required to the homepage contents stored in the database, because a Mozilla logo in the footer has been moved from HTML to CSS. I've already done it on stage to where this patch is currently deployed.

**How to test themes in Pontoon?**
1. Go to https://mozilla-pontoon-staging.herokuapp.com/settings/ and select Light theme.
2. Explore Pontoon and comment here if you notice anything unusual. See a list of pages to check below.
3. Go to [Settings](https://mozilla-pontoon-staging.herokuapp.com/settings/) again, select Dark theme and repeat the test. We've also made minor changes to the Dark theme.

**What pages to test?**

1. Dashboards:
- https://mozilla-pontoon-staging.herokuapp.com/projects/
- https://mozilla-pontoon-staging.herokuapp.com/projects/firefox/
- https://mozilla-pontoon-staging.herokuapp.com/teams/
- https://mozilla-pontoon-staging.herokuapp.com/fr/
- https://mozilla-pontoon-staging.herokuapp.com/fr/firefox/

2. User pages:
- https://mozilla-pontoon-staging.herokuapp.com/contributors/
- https://mozilla-pontoon-staging.herokuapp.com/profile/
- https://mozilla-pontoon-staging.herokuapp.com/settings/
- https://mozilla-pontoon-staging.herokuapp.com/notifications/

3. Admin:
- https://mozilla-pontoon-staging.herokuapp.com/admin/
- https://mozilla-pontoon-staging.herokuapp.com/admin/projects/firefox/
- https://mozilla-pontoon-staging.herokuapp.com/admin/projects/tutorial/strings/

4. Other assets:
- https://mozilla-pontoon-staging.herokuapp.com/machinery/
- https://mozilla-pontoon-staging.herokuapp.com/terms/
- https://mozilla-pontoon-staging.herokuapp.com/insights/
- https://mozilla-pontoon-staging.herokuapp.com/sync/
- https://mozilla-pontoon-staging.herokuapp.com/ (in [Settings](https://mozilla-pontoon-staging.herokuapp.com/settings/) select "Default Homepage" as your Homepage)
- Add-On Promotion banner

5. Translate View:
* https://mozilla-pontoon-staging.herokuapp.com/fr/firefox/all-resources/

**Known blocking issues:**

[...]

PR #3002, opened 2023-10-24
Add Theme Switching Functionality to Translate View

ayanaar:
This pull request introduces theme switching functionality directly within the translate view, allowing users to easily switch between light, dark, and system themes in the profile menu. (Issue https://github.com/mozilla/pontoon/issues/2936 )

### Changes:
- **New UI Element**: Added buttons in the profile menu allowing users to select their preferred theme: Dark, Light, or Match System.
- **Theme Persistence**: User theme preference is stored in their profile, ensuring a consistent experience across sessions.


translate/src/modules/user/components/UserMenu.css
```diff
@@ -106,3 +106,72 @@
   margin: 5px 0;
   padding: 0;
 }
+
+/* Profile menu */
```
mathjazz, line 110: This whole CSS file belongs to a User (Profile) Menu. Theme toggle would make more sense.

translate/src/modules/user/actions.ts
```diff
@@ -39,6 +41,21 @@ export type UpdateSettingsAction = {
...
 
+/**
+ * Update the user theme.
+ */
+export type UpdateThemeAction = {
+  readonly type: typeof UPDATE_THEME;
+  readonly theme: string;
+};
+
+export function updateTheme(theme: string) {
```
mathjazz, line 52: This function is not used anywhere.

translate/src/modules/user/actions.ts
```diff
@@ -65,6 +82,14 @@ export function saveSetting(
   };
 }
 
+export function saveTheme(theme: string, username: string): AppThunk {
+  return async (dispatch) => {
+    await updateUserTheme(username, theme);
+
+    dispatch({ type: UPDATE_THEME, theme });
```
mathjazz, line 89: We should also show notification (see above), same as on /pontoon.

translate/src/modules/user/reducer.ts
```diff
@@ -140,12 +142,18 @@ export function reducer(state: UserState = initial, action: Action): UserState {
           notifications: [],
           unread_count: '0',
         },
+        theme: action.data.theme ?? state.theme,
```
mathjazz, line 145: Can we replace `state.theme` with `''`, to follow the example of other keys?

In any case, we should include `theme` in the output of https://github.com/mozilla/pontoon/blob/main/pontoon/base/views.py#L846.

translate/src/modules/user/components/UserMenu.css
```diff
@@ -106,3 +106,72 @@
...
+.toggle-button button:hover {
+  color: var(--light-grey-6);
+}
+
+.toggle-button button.active {
+  background: var(--dark-grey-1);
+  border-color: var(--grey-3);
+  color: var(--light-grey-7);
+  font-weight: 400;
+}
```
mathjazz, line 172: We'll need to sync these colors with the ones used in https://github.com/mozilla/pontoon/pull/2997/files#diff-30336dc4648b06dd8f0228e1c1013ab8260d8a5404cb86cba8187ed141519f1bR1461, but that should be done in the theme CSS PR.

translate/src/modules/user/components/UserMenu.tsx
```diff
@@ -14,6 +14,8 @@ import './UserMenu.css';
 
 type Props = {
   user: UserState;
+  theme: string;
```
mathjazz, line 17: No need to pass that. We can get it from the `user`.

translate/src/modules/user/components/UserMenu.tsx
```diff
@@ -37,6 +40,57 @@ export function UserMenuDialog({
...
+
+    applyTheme(theme);
+
+    return () => {
+      mediaQuery.removeEventListener('change', handleThemeChange);
+    };
+  }, [theme]);
+
+  const handleThemeButtonClick = (selectedTheme: string) => {
+    setTheme(selectedTheme); // Update the local state
```
mathjazz, line 90: State already gets updated through the `onThemeChange()` method, which ends up calling the `saveTheme()` action. 

The only thing we should call here is applyTheme(), ideally from the same place that [Theme.tsx](https://github.com/mozilla/pontoon/blob/91977b087f6fc36a5d821eec1e37091471a4bf53/translate/src/context/Theme.tsx) should be calling it. Then you can remove lines 43-87.

translate/src/hooks/useTheme.ts
```diff
@@ -0,0 +1,45 @@
...
+      if (newTheme === 'system') {
+        newTheme = getSystemTheme();
+      }
+      document.body.classList.remove(
+        'dark-theme',
+        'light-theme',
+        'system-theme',
+      );
+      document.body.classList.add(`${newTheme}-theme`);
+    }
```
mathjazz, line 26: That and getSystemTheme() is all that should remain in this hook.

Everything else should go back to Theme.tsx, because it's only needed there.

mathjazz (commented):
Getting there! Let's factor out the applyTheme() as suggested.

translate/src/modules/user/components/UserMenu.tsx
```diff
@@ -37,6 +40,19 @@ export function UserMenuDialog({
   const ref = useRef<HTMLUListElement>(null);
   useOnDiscard(ref, onDiscard);
 
+  const [theme, setTheme] = useState<string>(
+    () => document.body.getAttribute('data-theme') || 'dark',
+  );
```
mathjazz, line 45: We can get the theme from the user state, so no need for this block.

translate/src/modules/user/components/UserMenu.tsx
mathjazz: We should call `applyTheme()` here.

translate/src/modules/user/components/UserMenu.tsx
```diff
@@ -49,6 +65,39 @@ export function UserMenuDialog({
...
+          <span className='toggle-button'>
+            <button
+              type='button'
+              value='dark'
+              className={`dark ${theme === 'dark' ? 'active' : ''}`}
+              title='Use a dark theme'
+              onClick={() => handleThemeButtonClick('dark')}
+            >
+              <i className='icon far fa-moon'></i>Dark
+            </button>
```
mathjazz, line 79: We should make the buttons and the label above localizable. Ideally, we should also factor out the button template.

translate/src/context/Theme.tsx
```diff
@@ -1,55 +1,17 @@
...
-    }
-
-    mediaQuery.addEventListener('change', handleThemeChange);
-
-    applyTheme(theme);
-
-    return () => {
-      mediaQuery.removeEventListener('change', handleThemeChange);
-    };
-  }, [theme]);
```
mathjazz, line 52: We still need the even handlers for system theme changes.

mathjazz (approved):

PR #3003, opened 2023-10-25
Fix validation of strings with leading and trailing newlines

### LINKED ISSUE #2991: GetText check fails incorrectly on newline
https://pontoon.mozilla.org/en-GB/all-projects/all-resources/?string=286055

If you copy the source string, an extra line is added at the back, and that fails the checks for GetText.

mathjazz:
Fix #2991.

We use CharFields for validating original strings and translations when translations are being saved.

By default, CharField [strips the value of the leading and trailing whitespace](https://docs.djangoproject.com/en/4.2/ref/forms/fields/#django.forms.CharField).

Until now, we've been only bypassing this default for translations. 

From now on, we also do that for original strings.

Deployed to stage for testing:
https://mozilla-pontoon-staging.herokuapp.com/fi/all-projects/all-resources/?string=192448

### DIFF
--- a/pontoon/translations/forms.py
+++ b/pontoon/translations/forms.py
@@ -51,5 +51,8 @@ class CreateTranslationForm(forms.Form):
             return None
         return self.cleaned_data["plural_form"]
 
+    def clean_original(self):
+        return self.data.get("original", "")
+
     def clean_translation(self):
         return self.data.get("translation", "")


pontoon/translations/forms.py
```diff
@@ -51,5 +51,8 @@ def clean_plural_form(self):
             return None
         return self.cleaned_data["plural_form"]
 
+    def clean_original(self):
+        return self.data.get("original", "")
```
mathjazz, line 55: Mind the `data` instead of `cleaned_data`.

eemeli (approved):

PR #3004, opened 2023-10-26
Prevent FOUT when using System Theme

mathjazz:
When system theme is selected, we apply the right theme in the client, because system settings can only be accessed from the client. That results in FOUT (flash of unstyled text) when navigating between pages if the system theme is Light, because Pontoon defaults to the dark theme and then immediatelly changes it to light when the JS executed.

This patch prevents that by storing the system theme setting in a cookie. Cookie is read by the server, so we set the theme accordingly already in a template.

Deployed to https://mozilla-pontoon-staging.herokuapp.com/.

### DIFF
--- a/pontoon/base/static/js/theme-switcher.js
+++ b/pontoon/base/static/js/theme-switcher.js
@@ -13,6 +13,7 @@ $(function () {
   function applyTheme(newTheme) {
     if (newTheme === 'system') {
       newTheme = getSystemTheme();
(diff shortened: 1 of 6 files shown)


eemeli (dismissed):
If a user has set `system` as preference, a cookie is set for a year. If within that year the user changes their preference, the cookie is not removed, and I _think_ then there's a possible flash of light content before going dark?

ayanaar (approved):

Issue #2999, opened 2023-10-22
Charts in global Insights page end in August


flodolo:
See screenshot. I'm pretty sure I've seen these graphs showing data for September, maybe even October before.

![Screenshot 2023-10-22 at 19-35-19 Insights](https://github.com/mozilla/pontoon/assets/3644868/5e8b8e9a-8836-4a57-bf7f-cab5e99025b3)

Issue #2979, opened 2023-10-09
Application error when loading the global Insights page for the first time


flodolo:
https://pontoon.mozilla.org/insights/

I assume not a lot of people check this, and some caching is involved. Every time I try to load this page, I get an application error. Reloading after a few seconds will work.

Issue #2938, opened 2023-08-17
Investigate the ability to generate Insights data on request rather than via cron job


mathjazz:
In order to present a chart like "pretranslation quality", we need the quality data for the team/project/project-locale's pretranslated messages for the past year. To get there, we:

1. Define new fields to summarise the data, which need to be added to the db.
2. Define a generator for the historical data for all views for the past year.
3. Fill the summary fields with historical data.
4. Write or extend a cron job task that appends new data to the db daily.
5. When requested, we construct the results based on the daily summary data.

The alternative that I'm suggesting be considered is to instead:
1. On request, generate the current team/project/project-locale's data from the pre-existing ActionLog data.
2. Cache the final computed result for 24 hours, so it doesn't need to be recalculated.

_Originally posted by @eemeli in https://github.com/mozilla/pontoon/issues/2929#issuecomment-1679757176_
            
See also:
https://github.com/mozilla/pontoon/pull/2929#pullrequestreview-1578723947

PR #3007, opened 2023-10-29
Optimize and fix the /insights page

### LINKED ISSUE #2980: Prevent bots from checking Insights page
I noticed in the logs Google Bot snooping around in Insights pages. 

I don't think there is any value there for a search engine, and given they're quite heavy, we should prevent them via robots.txt.

mathjazz:
Several Insights pages improvements are included in this PR:

* Fix #2980: Prevent bots from checking the /insights page
* Optimize ProjectLocaleInsightsSnapshot Admin view/editing page 
* Refactor data gathering for the /insights page:
  1. Fix https://github.com/mozilla/pontoon/issues/2979: Optimize /insights page. To render the charts on the Insights pages, we use periodically gathered ProjectLocaleInsightsSnapshot data. That works fine on Project, Locale and ProjectLocale pages, but not on the /insights page which shows data across the entire app. The amount of data requested from the DB quickly becomes big and the request starts timing out. In this patch, we gather data directly from the ActionLog as hinted at in https://github.com/mozilla/pontoon/issues/2938.
  2. Fix https://github.com/mozilla/pontoon/issues/2999: Show insights for all months.
  3. Calculate monthly averages rather than averages of averages. Lines for the "All" data point on both "Team pretranslation quality" and "Project pretranslation quality" are the same now.

Deployed to stage:
https://mozilla-pontoon-staging.herokuapp.com/insights/

Note that performance is hard to test reliably due to caching.

### DIFF
--- a/pontoon/base/templates/robots.txt
+++ b/pontoon/base/templates/robots.txt
@@ -2,3 +2,4 @@ User-agent: *
 Disallow: /a/
 Disallow: /admin/
 Disallow: /contributors/
+Disallow: /*insights/
--- a/pontoon/insights/admin.py
+++ b/pontoon/insights/admin.py
@@ -52,6 +52,7 @@ class ProjectLocaleInsightsSnapshotAdmin(admin.ModelAdmin):
         "completion",
         "unreviewed_strings",
     )
+    readonly_fields = ("project_locale",)
 
 
 admin.site.register(LocaleInsightsSnapshot, LocaleInsightsSnapshotAdmin)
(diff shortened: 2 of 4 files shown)


eemeli:
Shouldn't this include some tests?

mathjazz:
Yeah. Since this file is likely to be refactored as part of https://github.com/mozilla/pontoon/issues/2938, I think we should do it as part of that.

mathjazz:
As agreed in-person, I filed https://github.com/mozilla/pontoon/issues/3015.

eemeli (approved):

PR #3014, opened 2023-11-09
Replace react-content-marker with a regexp

### LINKED ISSUE #2515: Placeholders in XLIFF files are not clickable in source strings
Currently in Pontoon, we can not click on a placeholder in the source string in order to reproduce it in localized view (and avoid mistyping it) - vs in Android products, for example.

Example [here](https://pontoon.mozilla.org/hy-AM/firefox-for-ios/firefox-ios.xliff/?search=Menu.TrackingProtection.Details.Verifier&string=231665) and [here](https://pontoon.mozilla.org/hy-AM/firefox-for-ios/firefox-ios.xliff/?search=LoginsHelper.PromptUpdateLogin.Title.TwoArg&string=210088) (where it is partially clickable).

### LINKED ISSUE #3006: Handle all placeables using internal logic used in the editor
We use two different implementations of placeable detection.

Let's drop https://github.com/mozilla/react-content-marker/ and only use the internal logic that is currently only used in the editor.

That should also fix https://github.com/mozilla/pontoon/issues/2515.

eemeli:
Fixes #2515 
Fixes #3006

Ports all the rules from `react-content-marker` into an inlined implementation. Results are cleaner, more efficient, and less buggy. Also, for placeholders we now match the rules used in the editor highlighting.

In a few places the rules are relaxed a bit and/or made more sane, but these are unlikely to actually effect real-world messages noticably. Except where they fix current bugs. Also, the placeholder rules now match the editor.


translate/public/locale/en-US/translate.ftl
```diff
@@ -570,75 +570,35 @@ otherlocales-Translation--header-link =
...
-## Placeable parsers
+## Message terms
 ## Used to mark specific terms and characters in translations.
 
-placeable-parser-altAttribute =
-    .title = 'alt' attribute inside XML tag
-placeable-parser-camelCaseString =
-    .title = Camel case string
-placeable-parser-emailPattern =
+term-cli-option =
```
mathjazz, line 576: Why are the strings prefixed with "term"? We already use the expression "term" for something else and we have a term module with [corresponding term-* entries](https://github.com/mozilla/pontoon/blob/2ac367a4bfafcac22b31687fa1b26d29926232f9/translate/public/locale/en-US/translate.ftl#L663-L669) in the FTL file.

I suggest we use highlight and follow the naming convention from the top of the file.

mathjazz (commented):
There's a regression in string and concordance search - highlighting of placeables takes priority over highlighting the search match. It should be the other way around, because when searching, you care about search matches more:

https://pontoon.mozilla.org/sl/firefox/all-resources/?search=brand-full&string=192278

Other than that:

mathjazz (approved):
🎉

PR #3023, opened 2023-11-17
 Implement Redirection for Updated Locale Codes

### LINKED ISSUE #2302: Pontoon settings not accessible if profile has non existing locale as custom home page
## [Bug 1725310](https://bugzilla.mozilla.org/show_bug.cgi?id=1725310)
Bug Reporter: @flodolo
CC:   @mathjazz

https://discourse.mozilla.org/t/i-cannot-open-my-settings/84380

Background: we used to have `sat-Olck`, which was removed as the team started using `sat` instead. At that point, the Settings page became impossible to access.

To fix it, I had to manually change the setting in Django admin. Not sure if we have other users in the same condition at this point.

ayanaar:
Fixes #2302 

This pull request introduces a solution for handling changes to locale codes within our Pontoon application. Updating a locale code (e.g., changing `sat-Olck` to `sat`) previously could lead to broken links and inaccessible pages for users who had settings or preferences tied to the old locale code. To address this, I've implemented a redirection mechanism that ensures continuity even when locale codes are changed.

**Changes Made:**
1. **LocaleCodeHistory Model:** Introduced a new model to track the history of changes to locale codes. This model records the old and new codes whenever a locale is updated.
2. **Signals for Locale Updates:** Implemented Django signals to capture changes in locale codes and record them in the LocaleCodeHistory model.
3. **Redirection Utility Function (`get_locale_or_redirect`):** Developed a utility function that checks if a requested locale exists. If an entry is found, it either redirects to the view specified by redirect_view_name using the new locale code or returns the Locale object if no redirect_view_name is provided. This function is now used across various views where locales are handled.
4. **Updates to Views:** Modified several views, including the settings, translate, team, and localization views, to incorporate the `get_locale_or_redirect` function. This ensures that any references to outdated locale codes are correctly redirected to the new codes.
5. **Test Cases:** Comprehensive test cases have been added to validate the functionality of the `get_locale_or_redirect` utility function in various views. These tests cover scenarios like code changes, missing codes, and non-existent codes to ensure proper redirection and error handling.


pontoon/base/utils.py
```diff
@@ -639,29 +639,32 @@ def get_project_or_redirect(
...
-            redirect_url = reverse(redirect_view_name, kwargs=redirect_kwargs)
-            return redirect(redirect_url)
+        if code_history:
+            if redirect_view_name and url_arg_name:
+                redirect_kwargs = {url_arg_name: code_history.locale.code}
+                redirect_kwargs.update(kwargs)
+                redirect_url = reverse(redirect_view_name, kwargs=redirect_kwargs)
+                return redirect(redirect_url)
+            else:
+                return code_history.locale  # Return the locale instead of redirecting
```
mathjazz, line 668: Nit: You can dedent the main block like this:
```python
if not redirect_view_name or not url_arg_name:
    return code_history.locale

redirect_kwargs = {url_arg_name: code_history.locale.code}
redirect_kwargs.update(kwargs)
redirect_url = reverse(redirect_view_name, kwargs=redirect_kwargs)
return redirect(redirect_url)
...
```

Also, no need for the comment, it's already well clarified in the docstring above.

pontoon/contributors/views.py
```diff
@@ -310,24 +310,30 @@ def settings(request):
     all_locales = list(Locale.objects.all())
     all_locales.insert(0, default_homepage_locale)
 
-    # Set custom homepage selector value
-    custom_homepage_locale = profile.custom_homepage
-    if custom_homepage_locale:
+    # Set default for custom homepage locale based on name
+    custom_homepage_locale_object = get_locale_or_redirect(profile.custom_homepage)
```
mathjazz, line 314: I would just leave the name `custom_homepage_locale`.

pontoon/contributors/views.py
```diff
@@ -310,24 +310,30 @@ def settings(request):
...
+    custom_homepage_locale_name = custom_homepage_locale_object.name
```
mathjazz, line 315: Why `name` (not unique), not `code` (unique)?

pontoon/contributors/views.py
mathjazz: Also, you should first check if we have `profile.custom_homepage`.

If we don't, use `default_homepage_locale`.
Only if we do, call `get_locale_or_redirect()`, otherwise it will fail (redirect).

pontoon/contributors/views.py
```diff
@@ -310,24 +310,30 @@ def settings(request):
...
+    # Similar logic for preferred source locale
     default_preferred_source_locale = Locale(name="Default project locale", code="")
     preferred_locales = list(Locale.objects.all())
     preferred_locales.insert(0, default_preferred_source_locale)
 
-    # Set preferred source locale
-    preferred_source_locale = profile.preferred_source_locale
-    if preferred_source_locale:
+    # Set preferred source locale based on name
+    preferred_source_locale_object = get_locale_or_redirect(
```
mathjazz, line 330: Same comments as above.

mathjazz (commented):
Thanks for the update!

Left a few notes. It's mostly nits, but we should fix the logic in the settings view.

pontoon/contributors/views.py
ayanaar: From what I understood from our conversation yesterday, the name of the local rarely would ever be edited (in comparison to the locale code) so I thought that setting the custom and default homepage using the locale name would be more resilient to changes in the locale code. However, if it is not unique then that will cause issues so I will change the logic to use the locale code.

mathjazz: Sorry, sounds like a misunderstanding. Locale names are probably edited more often than locale codes.

When you address the comment [above](https://github.com/mozilla/pontoon/pull/3023#discussion_r1402386812), you might even be good without either.

pontoon/contributors/views.py
```diff
@@ -310,27 +311,38 @@ def settings(request):
...
+    try:
+        if profile.custom_homepage:
+            custom_homepage_locale = get_locale_or_redirect(profile.custom_homepage)
+        else:
+            custom_homepage_locale = default_homepage_locale
+    except Http404:
         custom_homepage_locale = default_homepage_locale
+        messages.info(
+            request,
+            "Your previously selected custom homepage locale is no longer available. Reverted to default.",
```
mathjazz, line 324: We don't really revert the locale, we just show the default homepage when that happens. The user still needs to press Save. So I'd replace "Reverted to default." with something along the lines of "Please pick a different one.".

pontoon/contributors/views.py
```diff
@@ -310,27 +311,38 @@ def settings(request):
...
-        custom_homepage_locale = Locale.objects.filter(
-            code=custom_homepage_locale
-        ).first()
-    else:
+    # Set default for custom homepage locale based on code
+    try:
+        if profile.custom_homepage:
+            custom_homepage_locale = get_locale_or_redirect(profile.custom_homepage)
+        else:
+            custom_homepage_locale = default_homepage_locale
```
mathjazz, line 319: You can move this line in front of the `try - except` block and then you can remove it from line 321. 

At that point you no longer need the `else` block and you can move the `try - except` block within the `if` block.

pontoon/contributors/views.py
```diff
@@ -310,27 +311,38 @@ def settings(request):
...
-        ).first()
-    else:
+    # Set preferred source locale based on code
+    try:
+        if profile.preferred_source_locale:
+            preferred_source_locale = get_locale_or_redirect(
+                profile.preferred_source_locale
+            )
+        else:
+            preferred_source_locale = default_preferred_source_locale
```
mathjazz, line 339: Same as above.

pontoon/contributors/views.py
```diff
@@ -310,27 +311,38 @@ def settings(request):
...
+    except Http404:
         preferred_source_locale = default_preferred_source_locale
+        messages.info(
+            request,
+            "Your previously selected preferred source locale is no longer available. Reverted to default.",
```
mathjazz, line 344: Same as above.

mathjazz (commented):
Nice work! Please also add the missing tests as we discussed.

mathjazz (approved):
Great job! This is much more than just a bugfix.

I'll file an issue to consider storing `custom_homepage` as `models.ForeignKey(Locale)`.

Issue #2032, opened 2016-01-26
Make translate view responsive


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1243132](https://bugzilla.mozilla.org/show_bug.cgi?id=1243132)
Bug Reporter: @mathjazz
CC:   anweshknayak@gmail.com, ranjithraj@riseup.net
Depends on:   [Bug 1586820](https://bugzilla.mozilla.org/show_bug.cgi?id=1586820)

Show entity list by default and open editor when entity selected. No in-page localization.

mathjazz:
Work split into multiple issues:
https://github.com/mozilla/pontoon/issues/3026
https://github.com/mozilla/pontoon/issues/3027
https://github.com/mozilla/pontoon/issues/3028
https://github.com/mozilla/pontoon/issues/3029

PR #3024, opened 2023-11-20
Implement a single-column UI for smaller screens

### LINKED ISSUE #3026: Implement a single-column translate view for smaller screens
Translation workspace is not really usable on mobile phones right now.

As the first step to fix that, let's convert it from a 3-column to a single-column layout.

mathjazz:
I've replaced #2032 with multiple sub-issues: https://github.com/mozilla/pontoon/issues/2032#issuecomment-1823048210.

This PR is a fix for the first one and a first step towards making Pontoon translation workspace responsive.

Fix #3026.

There's more things I'd like to do as part of this PR:
- [x] Open editor column after clicking on the string list item
- [x] ~Change entities list toggle icon when entities list is open~


translate/src/modules/entitieslist/components/EntitiesListToggle.tsx
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Render entities list toggle.
+ *
+ * Used in a single-column layout.
+ */
+export function EntitiesListToggle(): React.ReactElement<'div'> {
+  const toggleEntitiesList = () => {
+    document
+      .querySelector('#app > .main-content')
+      .classList.toggle('entities-list');
```
mathjazz, line 14: This is just a temporary hack to show the functionality. Will replace with proper state management before merging.

translate/src/modules/entitieslist/components/EntitiesListToggle.tsx
```diff
@@ -0,0 +1,22 @@
...
+  };
+
+  return (
+    <div className='entities-list-toggle' onClick={() => toggleEntitiesList()}>
+      <div className='icon fa fa-bars'></div>
+    </div>
```
eemeli, line 20: This should be a `<button>`.

eemeli (changes requested):
At least initially, it'd probably be much easier to iterate on the styles if all the `@media max-width` rules were in  a single CSS file, rather than spread all over.

The right column should jump under the main already around 800px. Could we add a first 3 -> 2 column transition at that point?

The `#app > header` has a `min-width: 700px` that adds an unnecessary transition. It would be better if it matched one of the other transition points.

The header's nav path is now completely unavailable on narrow screens. Is fixing that going to be a part of the follow-on menu adaptation PR? I'm rather hesitant about merging this to main without the path being visible _somewhere_ on narrow screens.

_comment_: Eventually, we'll want CSS transitions. Animating the width doesn't really work, though. But this is a later problem.

mathjazz:
Thanks for the feedback.

Sorry, I made some additional changes right after getting your review.

> At least initially, it'd probably be much easier to iterate on the styles if all the `@media max-width` rules were in a single CSS file, rather than spread all over.

Do you mean in e.g. App.css or a separate file with media queries only?

> The right column should jump under the main already around 800px. Could we add a first 3 -> 2 column transition at that point?

I'd do that as part of the "Implement a custom UI on middle screens (i.e. smartphones in landscape mode, tablets)." task.

> The `#app > header` has a `min-width: 700px` that adds an unnecessary transition. It would be better if it matched one of the other transition points.

Probably also best to handle as part of the above task when more transition points will be available.

> The header's nav path is now completely unavailable on narrow screens. Is fixing that going to be a part of the follow-on menu adaptation PR? I'm rather hesitant about merging this to main without the path being visible _somewhere_ on narrow screens.

Now that the string list toggle has moved, I've restored the logo, which leads to the homepage (usually locale dashboard).

> _comment_: Eventually, we'll want CSS transitions. Animating the width doesn't really work, though. But this is a later problem.

Agreed.

eemeli (changes requested):
The changes made it worse. The only way now to return from the strings list to the entity view is to re-select the message that I had open; previously I could just close the list. Effectively you've removed any "back" button functionality.

> > At least initially, it'd probably be much easier to iterate on the styles if all the `@media max-width` rules were in a single CSS file, rather than spread all over.
> 
> Do you mean in e.g. App.css or a separate file with media queries only?

Your pick, as long as it's just one place rather than many.

> > The right column should jump under the main already around 800px. Could we add a first 3 -> 2 column transition at that point?
> 
> I'd do that as part of the "Implement a custom UI on middle screens (i.e. smartphones in landscape mode, tablets)." task.

I would prefer that done already here, as it should be a rather small change.

> > The `#app > header` has a `min-width: 700px` that adds an unnecessary transition. It would be better if it matched one of the other transition points.
> 
> Probably also best to handle as part of the above task when more transition points will be available.

Same as above.

> > The header's nav path is now completely unavailable on narrow screens. Is fixing that going to be a part of the follow-on menu adaptation PR? I'm rather hesitant about merging this to main without the path being visible _somewhere_ on narrow screens.
> 
> Now that the string list toggle has moved, I've restored the logo, which leads to the homepage (usually locale dashboard).

This doesn't address the part of my concern where the UI no longer tells me where I am when looking at the entity view. What language, what project, what resource am I looking at?

mathjazz:
> The changes made it worse. The only way now to return from the strings list to the entity view is to re-select the message that I had open; previously I could just close the list. Effectively you've removed any "back" button functionality.

Clicking anything (but the search/filter) in the string will take you back to the editor. I was thinking about showing [a right arrow](https://fontawesome.com/icons/chevron-right?f=class) on string hover. I'll also try to make the UI move left and right as you switch between the editor and the string list in order to give the user more sense what is actually going on.

> This doesn't address the part of my concern where the UI no longer tells me where I am when looking at the entity view. What language, what project, what resource am I looking at?

You should see project and resource in the Metadata section. I'll think about what to do with the locale code.

eemeli:
Oh, also: Not sure if this was the case when I first reviewed, but in the stringlist there are two overlapping vertical scrollbars, and starting a scroll with the cursor on top of the searchbox and starting one from the entity list produce different results.

mathjazz:
Made a few updates, see commit messages for more details. I'll take a look at the failing tests with the fresh eyes tomorrow.

All review comments should be addressed, except for the requested changes to the navigation menu. I'll do that as part of the "Adapt menus for the smaller screens" task, where I'll also try to make the target locale visible. As part of that I'll also experiment with moving the string list toggle to the header.

translate/src/App.css
```diff
@@ -62,3 +59,65 @@
...
+  }
+
+  #app > .main-content.entities-list > .panel-list,
+  #app > .main-content.entities-list > .panel-content {
+    left: 0%;
+  }
+
+  #app .entity:hover .indicator {
+    display: block;
+  }
```
eemeli, line 102: This is a nice touch.

eemeli (changes requested):
There's an expression in Finnish that seems appropriate here: "väärin sammutettu", i.e. "incorrectly extinguished" when referring to a fire. The job got done, but not the right way.

The styling of the narrow-screen transitions looks good now, and I'm fine with the missing location info being addressed later. But let's not add any new Redux state. This really needs to be implemented with a React Context instead.

translate/src/context/EntitiesList.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export function EntitiesListProvider({
+  children,
+}: {
+  children: React.ReactElement;
+}) {
+  const [state, setState] = useState(() => ({
+    ...initEntitiesList,
+    toggleEntitiesList: () =>
+      setState((prev) => ({ ...prev, visible: !prev.visible })),
```
eemeli, line 23: Looking at how this is used, we don't really want a toggle function, but something like `showEntityList(visible: boolean)` -- in the places we want to set the visibility, we want to explicitly show or hide it.

eemeli (changes requested):

translate/src/context/EntitiesList.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export function EntitiesListProvider({
+  children,
+}: {
+  children: React.ReactElement;
+}) {
+  const [state, setState] = useState(() => ({
+    ...initEntitiesList,
+    show: (visible: boolean) => setState((prev) => ({ ...prev, visible })),
+  })) as [EntitiesList, React.Dispatch<React.SetStateAction<EntitiesList>>];
```
eemeli, line 23: Shouldn't need a typecast here.

```suggestion
  const [state, setState] = useState<EntitiesList>(() => ({
    ...initEntitiesList,
    show: (visible: boolean) => setState((prev) => ({ ...prev, visible })),
  }));
```

eemeli (approved):
See line comment for a final suggestion.

PR #3033, opened 2023-11-23
Restore notification panel delay back to 2 seconds

### LINKED ISSUE #3011: Hovering the notification bar makes it stick for over 10 seconds, with no way to dismiss it
Talking about the notification bar displayed after submitting a translation, or switching between suggest and translate mode.

If you hover the notification with your mouse, it will remain displayed and it's impossible to dismiss. 
It seems to disappear after about 10 seconds without user interaction.

mathjazz:
Fix #3011.

During the development of the light theme, I needed the notification popup to be shown for a longer period than just 2 seconds. And I committed that change by mistake. 🤦 

https://github.com/mozilla/pontoon/commit/b97cbb7fc2cec6731680476f880ec643ab4831ec#diff-c0aa89be6ec58edea23eb616736c9bdf130d5ce3063749733db72e24c020e8c0R28

### DIFF
--- a/translate/src/modules/notification/components/NotificationPanel.tsx
+++ b/translate/src/modules/notification/components/NotificationPanel.tsx
@@ -25,7 +25,7 @@ export function NotificationPanel(): React.ReactElement<'div'> {
   useEffect(() => {
     window.clearTimeout(timeout.current);
     if (message) {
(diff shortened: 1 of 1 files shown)


eemeli (approved):
Isn't 5E3 the modern choice for a delay like this? 😇

PR #3038, opened 2023-11-29
Reposition panels in a single-column translate view

### LINKED ISSUE #3027: Reposition panels in a single-column UI
Metadata, Machinery, Locales, Terms, Comments and Batch Actions panels need to be repositioned to better fit the single-column UI.

mathjazz:
Fix #3027.

This is a WIP to check if the approach to reposition Helpers using hooks makes sense. If it does, I'll factor out tabs and panels into components and reuse them in both views, desktop and mobile.

I've also included a CSS change to make a visual distinction between the History panel and the Helpers. Would love to hear your opinion on this.

### DIFF
--- a/translate/src/modules/entitydetails/components/Helpers.tsx
+++ b/translate/src/modules/entitydetails/components/Helpers.tsx
@@ -5,6 +5,7 @@ import { Tab, TabList, TabPanel, Tabs } from 'react-tabs';
 import type { Entity } from '~/api/entity';
 import { HelperSelection } from '~/context/HelperSelection';
 import type { Location } from '~/context/Location';
+import { useViewport } from '~/hooks/useViewport';
 import type { TermState } from '~/modules/terms';
 import type { UserState } from '~/modules/user';
 import { Machinery, MachineryCount } from '~/modules/machinery';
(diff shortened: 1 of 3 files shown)


mathjazz:
> When factoring out the tabs/panels, if they're pretty small they might not need their own separate files?

No, I'd keep them in the same file, something similar as we do in the Metadata component.

translate/src/hooks/useNarrowScreen.ts
```diff
@@ -0,0 +1,16 @@
+import { useEffect, useState } from 'react';
+
+/**
+ * Return true if the screen is narrower than 600 px. Useful in Responsive Web Design.
+ */
+export function useNarrowScreen(): boolean {
+  const [width, setWidth] = useState(window.innerWidth);
+
+  useEffect(() => {
+    const handleWindowResize = () => setWidth(window.innerWidth);
```
eemeli, line 10: This state change is going to trigger a re-render of the whole component. The state variable should be the boolean result of the comparison to avoid that.

translate/src/hooks/useNarrowScreen.ts
```diff
@@ -0,0 +1,16 @@
...
+    window.addEventListener('resize', handleWindowResize);
+    return () => window.removeEventListener('resize', handleWindowResize);
+  }, []);
+
+  return width <= 600;
```
eemeli, line 15: This magic number should be defined as a constant at the top of the file.

translate/src/modules/entitydetails/components/EntityDetails.tsx
```diff
@@ -88,21 +93,35 @@ export function EntityDetails(): React.ReactElement<'section'> | null {
     [dispatch],
   );
 
+  const { code } = useContext(Locale);
+
+  const openTeamComments = useCallback(() => {
+    const teamCommentsTab = commentTabRef.current;
+    const index = teamCommentsTab?._reactInternalFiber.index ?? 0;
```
eemeli, line 100: We should properly fix this at some point. But for now let's at least highlight it a bit better in the code.

```suggestion
    // FIXME: This is an ugly hack.
    // https://github.com/mozilla/pontoon/issues/2300
    const index = teamCommentsTab?._reactInternalFiber.index ?? 0;
```

translate/src/modules/entitydetails/components/Helpers.tsx
```diff
@@ -32,6 +33,57 @@ type Props = {
...
+  commentTabRef: any;
+}) {
+  return (
+    <Tab ref={commentTabRef}>
+      <Localized id='entitydetails-Helpers--comments'>{'COMMENTS'}</Localized>
+      <CommentCount teamComments={teamComments} />
+    </Tab>
+  );
+}
+
```
eemeli, line 86: I'd consider defining all of the above inline within the `Helpers()` function body. Then they wouldn't need to get their props passed in separately.

eemeli (changes requested):
Overall looks good. Some minor stuff, see inline comments.

eemeli (approved):

PR #3044, opened 2023-12-06
Fix positions of menus and overlays on narrow screens

### LINKED ISSUE #3028: Adapt menus in a single-column translate view
Some menus are not accessible in the single-column translate view.

As part of this task we also need to replace the main locale > project > resource navigation.

### LINKED ISSUE #3029: Implement a custom translate view for middle screens (i.e. tablets, smartphones in landscape mode)
We should be able to make better use of the screen space in middle-sized screens than by just using the single-column layout.

Perhaps 2-column layout would make sense.

mathjazz:
Fix #3028.
Fix #3029.

This is actually three things:
1. Menus and overlays are now always visible in the viewport.
1. Target locale is now always visible in the navbar.
1. A two-column UI is shown for screen width between 600 and 800 pixels.

Deployed to stage for [testing](https://mozilla-pontoon-staging.herokuapp.com/sl/firefox/all-resources/).


translate/src/hooks/useNarrowScreen.ts
```diff
@@ -1,21 +1,17 @@
...
-
 /**
- * Return true if the screen is narrower than 600px. Useful in Responsive Web Design.
+ * Return true if the screen is narrower than MAX_WIDTH. Useful in Responsive Web Design.
  */
-export function useNarrowScreen(): boolean {
-  const [isNarrow, setIsNarrow] = useState(
-    window.innerWidth <= NARROW_SCREEN_MAX_WIDTH,
-  );
+export function isNarrowerThan(MAX_WIDTH: number): boolean {
```
eemeli, line 6: Two problems here:
1. https://react.dev/learn/reusing-logic-with-custom-hooks#hook-names-always-start-with-use
2. Exposing the `MAX_WIDTH` number means that our JS code gets littered with magic constants.

I'd suggest refactoring this as either
```js
useWindowWidth(): 'narrow' | 'mid' | 'wide'
```
or
```js
useViewportColumns(): 1 | 2 | 3
```

eemeli:
The styles themselves look pretty good, mind. The tab contents should have something like a `min-height: 50%`, though, as there's an odd jump if you scroll down and then switch to a tab with no contents.

mathjazz:
> The tab contents should have something like a `min-height: 50%`, though, as there's an odd jump if you scroll down and then switch to a tab with no contents.

Tried to fix that in https://github.com/mozilla/pontoon/pull/3044/commits/ab76c2d26d80be1e92467712fe4a86dc5deb8c42.

translate/src/hooks/useWindowWidth.ts
```diff
@@ -0,0 +1,26 @@
...
+  narrow: 600,
+  medium: 800,
+};
+
+/**
+ * Return true if the screen is narrower than the given width.
+ *
+ * Useful in Responsive Web Design.
+ */
+export function useWindowWidth(width: 'narrow' | 'medium'): boolean {
```
eemeli, line 13: This hook name is misleading, because `narrow` and `medium` sound like they would be exclusive with each other, but with this implementation `narrow` is a subset of `medium`. That is not how it usually works.

In my earlier proposal, I had this returning effectively an enum of `narrow`/`medium`/`wide`, and for the "narrower than" type of comparison to be done in the calling code.

eemeli (changes requested):

translate/src/hooks/useWindowWidth.ts
```diff
@@ -1,26 +1,33 @@
...
- * Return true if the screen is narrower than the given width.
+ * Return window width range: narrow, medium or wide.
  *
  * Useful in Responsive Web Design.
  */
-export function useWindowWidth(width: 'narrow' | 'medium'): boolean {
-  const [isWindowWidth, setIsWindowWidth] = useState(
-    window.innerWidth <= MAX_WIDTH[width],
-  );
+export function useWindowWidth(): string {
```
eemeli, line 13:
```suggestion
export function useWindowWidth(): 'narrow' | 'medium' | 'wide' {
```

eemeli (approved):
Let's narrow the type a bit; otherwise looks good.

PR #3046, opened 2023-12-11
Improve translate view experience on mobile devices

mathjazz:
This PR merges the `rwd` branch into the `main` branch.

**1. Implement a single-column UI for smaller screens (#3024)**
* Move third column under the middle column
* Add Entities list toggle
* Animate switching between the editor and the string list
* When hovering over strings in the string list, show indicator to open the editor
* Hide main navigation
* For easier development, move all `@media max-width` to App.css temporarily

**2. Reposition panels in a single-column translate view (#3038)**
* Only show one Helpers panel at the same time
* Use dot rather than count as the tab indicator
* Make visual distinction between the History panel and the Helpers.
* Add support for batch actions in the single-columm UI

**3. Reposition menus on narrow screens and add 2-column view (#3044)**
* Fix positions of menus and overlays on narrow screens
* Always show target locale code in the navbar
* Implement a 2-column UI for middle screen widths between 600 and 800 pixels

**Other changes:**
* Remove unused flex properties
* Drop 700px transition on the header
* Move OriginalString component and friends out of Metadata component
* Unify Term tab count behaviour with other Helper tabs: do not show 0
* Unify dark-theme.css and light-theme.css structure
* De-hardcode and fix color value in the FiltersPanel


eemeli (changes requested):
I think it's a bit early to merge this to main, but I won't block this if at least the following issues are either fixed or have explicit issues filed for them:
- When looking at the list of messages, navigation state is not obvious.
- Opening translate view always focuses on a single string rather than the list of strings.

mathjazz:
Filed #3047 and #3048.

eemeli (approved):
In this one case, would recommend considering using rebase rather than squash when merging.

mathjazz:
> In this one case, would recommend considering using rebase rather than squash when merging.

Agreed.

PR #3056, opened 2023-12-18
Uniform Checks for Locale, Project, and ProjectLocale in AJAX Views

### LINKED ISSUE #2338: Project+locale AJAX endpoint logic is not uniform
At the moment, `/:code/:slug/ajax/` and `/:code/:slug/ajax/tags/` check if the locale & project exist and are visible to the user, while `/:code/:slug/ajax/contributors/` and `/:code/:slug/ajax/insights/` check that the ProjectLocale exists. [All of these](https://github.com/mozilla/pontoon/blob/master/pontoon/localizations/views.py) should check both conditions.

ayanaar:
Fixes #2338 

This PR addresses an inconsistency in our AJAX views regarding the checks for the existence and visibility of `Locale` and `Project`, as well as the existence of `ProjectLocale`. Previously, `ajax_resources` and `ajax_tags` only checked for the `Locale` and `Project`, while `ajax_contributors` and `ajax_insights` checked for the `ProjectLocale`. This update ensures that all these views consistently perform both sets of checks.

### DIFF
--- a/pontoon/localizations/tests/test_views.py
+++ b/pontoon/localizations/tests/test_views.py
@@ -10,6 +10,7 @@ from pontoon.base.tests import (
     ResourceFactory,
     TranslationFactory,
     TranslatedResourceFactory,
+    ProjectLocaleFactory,
 )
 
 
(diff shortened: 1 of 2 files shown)


mathjazz (approved):
Greak work!

The code makes sense. I also tested locally and didn't spot any issues.

PR #3060, opened 2024-01-11
Use correct Google Translate AutoML locale

mathjazz:
For locales that use the same locale code in Google Translate (e.g. es-AR and es-ES both use es), use the correct locale code in Machinery Google Translate if AutoML is enabled. Right now, we use the first locale found in the DB.


docker/config/server.env.template
```diff
@@ -9,4 +9,6 @@ FXA_CLIENT_ID=727f0251c388a993
 FXA_SECRET_KEY=e43fd751ca5687d28288098e3e9b1294792ed9954008388e39b1cdaac0a1ebd6
 FXA_OAUTH_ENDPOINT=https://oauth.stage.mozaws.net/v1
 FXA_PROFILE_ENDPOINT=https://profile.stage.mozaws.net/v1
+GOOGLE_APPLICATION_CREDENTIALS=/app/pontoon/pontoon-prod.json
+GOOGLE_AUTOML_PROJECT_ID=85591518533
```
eemeli, line 13: These seem surprising to include in the template; aren't these Mozilla instance details? Also, the `pontoon-prod.json` file isn't included in the repo.
```suggestion
```

eemeli (approved):
Actual code looks fine, but the `server.env.template` change seems spurious.

PR #3061, opened 2024-01-11
Properly serialize Fluent placeables generated by Pretranslation

### LINKED ISSUE #3072: Fluent pretranslations not serialized in canonical form
Fluent pretranslations are not stored in a canonical form. For example, `key = {example}` should actually be stored as `key = { example }` (mind the whitespace).

In the exported files in the repository, the strings appear as they should, which results in sync trying to import pretranslations, because they aren't the same as in Pontoon DB. That also triggers #2284. 

In the UI (editor, string list) the strings are also rendered properly, which only makes the problem more difficult to spot.

We should make sure that pretranslations are always stored them in the canonical form.

mathjazz:
After Fluent pretranslations are created, we need to parse and serialize them again in order to make sure they are stored in the canonical form (e.g. with spaces within curly braces).

Fix #3072.

### DIFF
--- a/pontoon/pretranslation/pretranslate.py
+++ b/pontoon/pretranslation/pretranslate.py
@@ -5,7 +5,7 @@ import re
 from django.db.models import CharField, Value as V
 from django.db.models.functions import Concat
 
-from fluent.syntax import ast, FluentParser, FluentSerializer
+from fluent.syntax import FluentParser, FluentSerializer
 from functools import reduce
 
 from pontoon.base.models import User, TranslatedResource
@@ -60,13 +60,6 @@ def get_pretranslations(entity, locale, preserve_placeables=False):
 
         # Parse and serialize pretranslation again in order to assure cannonical style
         parsed_pretranslation = parser.parse_entry(pretranslation)
-
-        if isinstance(parsed_pretranslation, ast.Junk):
-            log.info(
-                f"Fluent pretranslation error: Invalid translation: {pretranslation}"
-            )
-            return []
-
         pretranslation = serializer.serialize_entry(parsed_pretranslation)
 
         authors = [services[service] for service in pretranslate.services]
--- a/pontoon/pretranslation/pretranslate.py
+++ b/pontoon/pretranslation/pretranslate.py
@@ -93,7 +93,7 @@ def get_pretranslations(entity, locale, preserve_placeables=False):
 
 
 def get_pretranslated_data(source, locale, preserve_placeables):
-    # Empty strings do not need translation
+    # Empty strings and strings containing whitespace only do not need translation
     if re.search("^\\s*$", source):
         return source, "tm"
 
--- a/pontoon/pretranslation/pretranslate.py
+++ b/pontoon/pretranslation/pretranslate.py
@@ -5,7 +5,7 @@ import re
 from django.db.models import CharField, Value as V
 from django.db.models.functions import Concat
 
-from fluent.syntax import FluentParser, FluentSerializer
+from fluent.syntax import ast, FluentParser, FluentSerializer
 from functools import reduce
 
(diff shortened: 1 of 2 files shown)


pontoon/pretranslation/pretranslate.py
```diff
@@ -57,6 +57,15 @@ def get_pretranslations(entity, locale, preserve_placeables=False):
             return []
 
         pretranslation = serializer.serialize_entry(entry)
+        print(pretranslation)
+        # error-not-subscribed = Esta dirección de correo electrónico no está suscrita a {-product-name}.
+
+        # Parse and serialize pretranslation again in order to assure cannonical style
+        parsed_pretranslation = parser.parse_entry(pretranslation)
+        pretranslation = serializer.serialize_entry(parsed_pretranslation)
+        # error-not-subscribed = Esta dirección de correo electrónico no está suscrita a { -product-name }.
```
mathjazz, line 66: Leaving actual translations in as comments just for the review purpose (I'll remove that before mering).

@eemeli Is there a more elegant way to achieve the same results?

eemeli: Not at the moment, as the results of the pretranslation are not properly validated: https://github.com/mozilla/pontoon/blob/c63b0c5f6526a4e3507d4e4e7af1463522ff06d3/pontoon/pretranslation/transformer.py#L239-L246

pontoon/pretranslation/pretranslate.py
```diff
@@ -57,6 +57,15 @@ def get_pretranslations(entity, locale, preserve_placeables=False):
             return []
 
         pretranslation = serializer.serialize_entry(entry)
+        print(pretranslation)
+        # error-not-subscribed = Esta dirección de correo electrónico no está suscrita a {-product-name}.
+
+        # Parse and serialize pretranslation again in order to assure cannonical style
+        parsed_pretranslation = parser.parse_entry(pretranslation)
```
eemeli, line 64: The serializer doesn't check for junk at its `serialize_entry()` enpoint, so this should be included:

```suggestion
        parsed_pretranslation = parser.parse_entry(pretranslation)
        if isinstance(parsed_pretranslation, FTL.Junk):
            log.info(f"Fluent pretranslation error: Invalid translation: {pretranslation}")
            return []
```

Also need a new import at the top of the file:
```python
from fluent.syntax import ast as FTL
```

eemeli (changes requested):

pontoon/pretranslation/pretranslate.py
```diff
@@ -58,6 +58,17 @@ def get_pretranslations(entity, locale, preserve_placeables=False):
...
         pretranslation = serializer.serialize_entry(entry)
 
+        # Parse and serialize pretranslation again in order to assure cannonical style
+        parsed_pretranslation = parser.parse_entry(pretranslation)
+
+        if isinstance(parsed_pretranslation, ast.Junk):
+            log.info(
+                f"Fluent pretranslation error: Invalid translation: {pretranslation}"
+            )
+            return []
```
mathjazz, line 68: On a second thought, I'm not sure we should make this change:
- Pretranslation with an Error is still better than no pretranslation
- It's not really in scope of this feature (with change in this PR, no additional messages will become Junk)

eemeli: I disagree, because pretranslations are allowed to go to production, and we should not allow invalid messages to go to production. And I would say that this is well within "Properly serialize Fluent placeables generated by Pretranslation".

eemeli (approved):
Approving, provided that the added check for junk remains included.

pontoon/pretranslation/pretranslate.py
mathjazz: Pretranslations with errors don't go to production.

Issue #3015, opened 2023-11-09
Add unit tests for the pontoon.insights module


mathjazz:
Test coverage of the pontoon.insights module is at 0 %. Let's fix that, before or as part of #2938.

PR #3079, opened 2024-01-28
Add tests for the pontoon.insights module (#3015)

potat0:
Fix https://github.com/mozilla/pontoon/issues/3015.

This is my first contribution to this project. Let me know if there are any improvements I can do. 
* Added tests for pontoon.insights module
* Added test for calling the endpoint without arguments and no data
* Added test with approved and rejected translations per month and assert that the translation quality percents match
* Ran `make test`
  * image
  * image
  * image


pontoon/insights/tests/test_views.py
```diff
@@ -0,0 +1,237 @@
...
+from pontoon.insights import views
+from pontoon.test.factories import (
+    UserFactory,
+    ResourceFactory,
+    TranslationFactory,
+)
+
+
+def approve(translation, user, timestamp):
+    translation.approved = True
```
mathjazz, line 23: This line doesn't have any impact on the rest of the code, so it can be removed.

At that point you might want to drop these two functions and call `perform_action()` directly.

pontoon/insights/tests/test_views.py
```diff
@@ -0,0 +1,237 @@
...
+    return perform_action(
+        ActionLog.ActionType.TRANSLATION_APPROVED, translation, user, timestamp
+    )
+
+
+def reject(translation, user, timestamp):
+    translation.approved = False
```
mathjazz, line 30: Same as above in the `approve()` function.

pontoon/insights/tests/test_views.py
```diff
@@ -0,0 +1,237 @@
...
+        action_type=action_type,
+        performed_by=user,
+        translation=translation,
+    )
+    action.created_at = timestamp
+    action.save()
+    return action
+
+
+def create_translation(locale, resource, resource2, pretranslation_user):
```
mathjazz, line 47: Seems like this is not used.

mathjazz (commented):
Nice work! 

Only left a few comments to hopefully make the code slightly more manageable.

potat0:
Thank you for the review! I'll try to get around to making changes this weekend.

pontoon/insights/tests/test_views.py
potat0: deleted

pontoon/insights/tests/test_views.py
potat0: done

pontoon/insights/tests/test_views.py
potat0: done

potat0:
Hi @mathjazz , let me know if there are any additional changes you would like to see.

pontoon/test/fixtures/base.py
```diff
@@ -19,6 +20,17 @@ def client_superuser(client, admin):
     return client
 
 
+@pytest.fixture
+def system_user():
```
mathjazz, line 24: Nit: `sync_user`.

pontoon/test/fixtures/base.py
```diff
@@ -19,6 +20,17 @@ def client_superuser(client, admin):
...
+    """Add the system user to the test data."""
+    return factories.UserFactory(email="pontoon-sync@example.com")
+
+
+@pytest.fixture
+def pretranslation_user():
```
mathjazz, line 30: Nit: `tm_user`.

mathjazz: As mentioned, we should also use this user in https://github.com/mozilla/pontoon/blob/main/pontoon/pretranslation/tests/test_pretranslate.py.

Please also move the `gt_user` here.

mathjazz (commented):
Thanks for the update (and sorry for the slow response from my end!).

I only added a couple more comments.

Going forward, could you please just add commits with changes, rather than force-push? That way it'll be much easier to see what exactly has changed since the code has been looked at.

potat0:
Hi @mathjazz . No problem. Sorry for missing moving the fixtures from test_pretranslate.py. Hopefully I got everything this time. I'll use normal commits going forward instead of force pushes.

mathjazz:
Thanks for the update!

Please see the failing tests.

mathjazz:
Please note that the pytest is still failing - we just introduced caching that causes it:
https://github.com/mozilla/pontoon/commit/7bdcaccc3d4a6c7ec36a62d5f0f4fd2b7a00fe84#diff-78a0f2e1d3a0a671add455c921328565557bfd1c242000b244fb2a9677a696bcR10

potat0:
Thank you for pointing out the latest caching update. I was stumped why it was failing in ci/cd. After pulling and merging latest, I was able to reproduce the pytest failure occurring in the GitHub action. Clearing the cache prior to running the tests seems to fix it locally. I was looking into how to clear specific views out of the cache, but these solutions (https://stackoverflow.com/questions/2268417/expire-a-view-cache-in-django/2268583#2268583) are quite involved and could be brittle since they depend on the implementation within Django.

mathjazz (approved):

PR #3091, opened 2024-02-09
Cache contributors views

### LINKED ISSUE #3082: Cache pontoon.contributors.views:ContributorsView.get view
Lists of top contributors are generated on the fly from the translation data. That means we must process all translations to generate the /contributors view, which takes time. 

We should cache pontoon.contributors.views:ContributorsView.get for one day or fetch data periodically (e.g. once per day) via cron job, and store it in the DB. 

We should also investigate if it makes sense to retrieve some data asynchronously (e.g. stats, user role). That’s particularly important for the Authors and Time Range filter, which cannot use cached (and potentially outdated) data and shares the code with contributors views.

mathjazz:
This patch caches all Contributors views for a period of 1 day:
- /contributors
- /LOCALE/contributors
- /projects/PROJECT/contributors
- /PROJECT/LOCALE/contributors

It mostly fixes #3082, but I'll keep the issue open until I check if there's something we can do to speed up the Authors and Time Range filters.

If this works well, we should cache the Insights views.

### DIFF
--- a/pontoon/contributors/urls.py
+++ b/pontoon/contributors/urls.py
@@ -3,7 +3,7 @@ from django.urls.converters import StringConverter
 from django.views.decorators.cache import cache_page
 from django.views.generic import RedirectView
 
-from pontoon.settings import PER_VIEW_CACHE_TIMEOUT
+from pontoon.settings import VIEW_CACHE_TIMEOUT
 
 from . import views
 
@@ -28,7 +28,7 @@ urlpatterns = [
     # List contributors
     path(
         "contributors/",
-        cache_page(PER_VIEW_CACHE_TIMEOUT)(views.ContributorsView.as_view()),
+        cache_page(VIEW_CACHE_TIMEOUT)(views.ContributorsView.as_view()),
         name="pontoon.contributors",
     ),
     # Contributor profile by username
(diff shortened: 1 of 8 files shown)


pontoon/contributors/urls.py
```diff
@@ -1,7 +1,10 @@
 from django.urls import path, register_converter
 from django.urls.converters import StringConverter
+from django.views.decorators.cache import cache_page
```
mathjazz, line 3: [Django docs](https://docs.djangoproject.com/en/5.0/topics/cache/#specifying-per-view-cache-in-the-urlconf) recommend to specify the per-view cache in the URLconf rather than next to the view functions themselves.

pontoon/teams/urls.py
```diff
@@ -70,7 +73,9 @@
                             # Team contributors
                             path(
                                 "contributors/",
```
mathjazz, line 75: [As per the docs](https://docs.djangoproject.com/en/5.0/topics/cache/#the-per-view-cache), if multiple URLs point at the same view, each URL will be cached separately. In this example, `/de/contributors/` and `/it/contributors/` will be cached separately, as you may expect. But once a particular URL has been requested, subsequent requests to that URL will use the cache.

eemeli (approved):
Looks pretty straight-forward. My only (very minor) quibble is over the "per view" part of the `PER_VIEW_CACHE_TIMEOUT` variable, which is ultimately clear in context, but I was initially parsing it as something like "per user visit", which wouldn't really make sense.

mathjazz:
Renamed `PER_VIEW_CACHE_TIMEOUT` to `VIEW_CACHE_TIMEOUT`.

PR #3100, opened 2024-02-15
Implementation of LLM-Assisted Translations Backend 

ayanaar:
Addresses Issue #3068 

This PR introduces the backend infrastructure for the LLM-Assisted Translations in Pontoon project. The implementation is divided into two components:

1. **`openai_connection.py`**: This module serves as the interface to the OpenAI API, using the `gpt-4-0125-preview` model. It defines prompts for three translation refinements—informal, formal, and alternative—tailoring the output to meet specific stylistic or formal requirements.

2. **`refine_translation.py`**: A Django management command that facilitates user interaction with the translation refinement feature directly from the command line. Users can input an English source string along with its Slovenian machine translation and specify the desired refinement type. The command returns a refined Slovenian translation.

This backend setup allows users to use the  text generation capabilities of GPT-4 for enhancing translation quality and adaptability.

## Usage Instructions:

To utilize the translation refinement feature, execute the following command in the terminal, substituting `<formality>` with `informal`, `formal`, or `alternative` based on the required refinement type:

```bash
python manage.py refine_translation <formality> "<English text>" "<Slovenian translation>"
```

#### For an Alternative Translation:

```bash
python manage.py refine_translation alternative "Email resent. Add { $accountsEmail } to your contacts to ensure smooth delivery." "E-pošta ponovno poslana. Dodajte { $accountsEmail } med svoje stike in si zagotovite nemoteno dostavo." 
```

#### For an Informal Translation:
```bash
python manage.py refine_translation informal "Your subscription has been confirmed." "Vaša naročnina je bila potrjena."
```

#### For an Formal Translation:
```bash
python manage.py refine_translation formal "We appreciate your feedback." "Vaše povratne informacije cenimo."
```

## Handling Placeholders and Special Characters:
When entering commands, it's essential to handle placeholders `({ placeholder })` and special characters correctly:

- Placeholders: Ensure placeholders within the text are enclosed in quotes to prevent the shell from interpreting them as commands or variables.
- Special Characters: Use single quotes `'` around texts containing special characters or when placeholders are used. If your text includes single quotes, escape them using `'\''`.

**Example with Placeholders:**
```bash
python manage.py refine_translation alternative "Please update your profile here: { $profileUrl }" 'Prosimo, posodobite svoj profil tukaj: { $profileUrl }'
```

## Next Steps:
[...]

PR #3111, opened 2024-02-22
Update actions in workflows to latest versions

flodolo:


.github/workflows/py-lint.yml
```diff
@@ -49,8 +49,8 @@ jobs:
   pyupgrade:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
         with:
           python-version: 3.9
```
flodolo, line 55: Is there a reason to keep using 3.9, and not a more recent version (e.g. 3.12)? I noticed black reformats several files locally, but not in the workflow.

mathjazz: No, we should upgrade. We might even _need_ to upgrade in order to land #3100.

mathjazz (approved):

.github/workflows/py-lint.yml
flodolo: I'll do it in a separate PR to see if the workflows actually fail as I would expect.

mathjazz: Makes sense, we should do that also for Docker and Heroku in one go:
https://github.com/mozilla/pontoon/blob/main/docker/Dockerfile
https://github.com/mozilla/pontoon/blob/main/runtime.txt

Issue #3059, opened 2024-01-05
Upgrade Django to 4.2 (LTS)


mathjazz:
Support for Django 3.2 (LTS) that we currently use ends at the end of ~March~ April.

Let's upgrade to the next LTS release - 4.2.

Resources:
https://docs.djangoproject.com/en/5.0/howto/upgrade-version/
https://medium.com/@awaisq/upgrade-django32-to-django42-d581d272540e

PR #3112, opened 2024-02-22
Update workflows and docker to Python 3.10

flodolo:


flodolo:
Not sure how to test the changes outside of the workflows.

I've quoted the Python version, since that's the example in the action's README, and I had issue in the past myself.

flodolo:
Well, 3.12 doesn't seem to work (fails to install requirements)

mathjazz:
> Well, 3.12 doesn't seem to work (fails to install requirements)

You'll probably need to rebuild requirements with `make requirements`.

flodolo:
I think I need to leave this to someone who actually understands what they're doing.

* Both `make build` and `make requirements` fail locally.
* `flake8` fails in the workflow but passes locally.

I believe the version of numpy is incompatible with 3.12
https://numpy.org/doc/stable/release/1.26.0-notes.html

mathjazz:
Sure thing, thanks for starting it!

At least 3.11 on Bookworm seems to work:
https://wanadev.github.io/pontoon-debian/foreword.html#targeted-systems

flodolo:
OK, I managed to get the flake8 errors locally by upgrading from 3.12 to 3.12.2

flodolo:
Trying with 3.11

flake8 seems very unreliable with 3.12, but we are also very behind (3.7.7 vs 7.0.0). Also black is a couple major versions behind.

flodolo:
What a rabbit hole.

tears

I think the failure is [due to celery](https://github.com/DataDog/dd-trace-py/pull/4125). But updating Celery to 5.3.6 makes `make requirements` fail because of conflicting dependencies.

flodolo:
To summarize.

* Python 3.12: package installation fails because of `numpy` (not sure if others might fail later). We're at 1.24.2 and we'd need 1.26
* Python 3.11: tests fail because of `Celery`. Updating it to 5.3.6 makes `make requirements` fail because of conflicting dependencies, which seem to belong to core `Django` packages. So, for that, #3059 might be a requirement.
* Python 3.10: seems to work?

pontoon/sync/formats/json_keyvalue.py
```diff
@@ -37,9 +37,9 @@ def __init__(self, order, key, context, source_value, value):
             key=key,
             context=context,
             source_string=source_value,
-            strings={None: value}
-            if value
-            else {},  # No plural support in key value JSON
+            strings=(
+                {None: value} if value else {}
+            ),  # No plural support in key value JSON
```
flodolo, line 42: I kept this one in, although not strictly necessary (it was done by black 24, while we use 22). The previous version doesn't seem particularly readable.

requirements/lint.txt
```diff
@@ -66,7 +66,3 @@ tomli==1.2.2 \
     --hash=sha256:c6ce0015eb38820eaf32b5db832dbc26deb3dd427bd5f6556cf0acac2c214fee \
     --hash=sha256:f04066f68f5554911363063a30b108d2b5a5b1a010aa8b6132af78489fe3aade
     # via black
-typing-extensions==4.5.0 \
-    --hash=sha256:5cb5f4a79139d699607b3ef622a1dedafa84e115ab0024e0d9c044a9479ca7cb \
-    --hash=sha256:fb33085c39dd998ac16d1431ebc293a8b3eedd00fd4a32de0ff79002c19511b4
-    # via black
```
mathjazz, line 72: This is great news for https://github.com/mozilla/pontoon/pull/3100.

flodolo:
Discussing on the call, I realized the difference between bookworm and bullseye (EOL on July 2024)
https://wiki.debian.org/DebianReleases

I will try to use the bookwork version (Debian 12), hopefully everything keeps working.

flodolo:
> Discussing on the call, I realized the difference between bookworm and bullseye (EOL on July 2024) https://wiki.debian.org/DebianReleases
> 
> I will try to use the bookwork version (Debian 12), hopefully everything keeps working.

Correction: let's stick to the current version. Locally `make build`fails for me when using `python:3.10-bookworm`, so that might be for a different PR.

```
E: Unable to locate package postgresql-server-dev-13
```

mathjazz (commented):
Thanks for the patch!

Are a5fa14382d404d8c54be6fa7566fb1a764de33b1 and 29bd60a323f347bb7402d6623cbe1237881f9fdd caused by flake8 and black performing differently on different versions of Python? I'm surprised by the `l -> loc` change, which I'm unable to reproduce locally.

I'm curious how did you discover https://github.com/mozilla/pontoon/pull/3112/commits/b459b60accebcd3dd2e57e893e438467ea8d9c69.

Please also replace `--py39-plus` with `--py310-plus` in the `Makefile`.

Going forward, I'd appreciate if you can make a PR from your fork.

mathjazz:
The patch is deployed to stage and seems to work ok!

flodolo:
> Are [a5fa143](https://github.com/mozilla/pontoon/commit/a5fa14382d404d8c54be6fa7566fb1a764de33b1) and [29bd60a](https://github.com/mozilla/pontoon/commit/29bd60a323f347bb7402d6623cbe1237881f9fdd) caused by flake8 and black performing differently on different versions of Python? I'm surprised by the `l -> loc` change, which I'm unable to reproduce locally.

The l warning (unclear variable name) was showing up in 3.11 and 3.12, I kept it there.

> I'm curious how did you discover https://github.com/mozilla/pontoon/commit/b459b60accebcd3dd2e57e893e438467ea8d9c69.

[ruff ](https://github.com/astral-sh/ruff)

flodolo:
> Going forward, I'd appreciate if you can make a PR from your fork.

Sure.

Out of curiosity, is there a reason to avoid using the main repo?

mathjazz:
> The l warning (unclear variable name) was showing up in 3.11 and 3.12, I kept it there.
> 
> [ruff ](https://github.com/astral-sh/ruff)

I'd keep the changes made by `flake8` on newer versions of python, because we'll get there eventually.

I'd refrain from integrating changes made by `ruff`, because we don't use it.

mathjazz:
> Out of curiosity, is there a reason to avoid using the main repo?

Just a convenience - it makes the workflow to test the patch locally and on stage a little different.

flodolo:
> I'd keep the changes made by `flake8` on newer versions of python, because we'll get there eventually.

Why not keep them in? They're still valid changes, and very minor (also unlikely to create rebase problems).

> I'd refrain from integrating changes made by `ruff`, because we don't use it.

Same, they're valid fixes (that are also reported by my VS Code when I open a file). I understand not taking all changes from `ruff`, but if they're valid, why not keeping them?

The changes in both commit seem reasonable, at this point it's extra work (and hard to justify) to remove them just to add them back in a few months.

flodolo:
I'm checking if those changes aren't actually detected by a newer version of flake8, because I might be mixing things up (lots of tests today).

mathjazz (approved):
Nice work, thanks!

Tested latest changed again on stage and locally and they seem to work fine.

> They're valid fixes (that are also reported by my VS Code when I open a file). I understand not taking all changes from `ruff`, but if they're valid, why not keeping them?

They are still out of scope of the PR (which is about python upgrade), so I'd prefer to address issues fixed by `ruff` more robustly (e.g. by introducing `ruff` support).

PR #3115, opened 2024-02-22
Use Debian 12 in docker image, update PostgreSQL and Node versions

flodolo:


docker/Dockerfile
```diff
@@ -17,12 +17,16 @@ RUN apt-get update \
     && apt-get install -y apt-transport-https \
     # Get source for node.js
     && curl -fsSL https://deb.nodesource.com/setup_16.x | bash - \
+    # Add PostgreSQL repository
+    && sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list' \
```
flodolo, line 21: This is based on https://www.postgresqltutorial.com/postgresql-getting-started/install-postgresql-linux/

Package is not available otherwise.

docker/Dockerfile
```diff
@@ -17,12 +17,16 @@ RUN apt-get update \
...
+    && curl -fsSL https://www.postgresql.org/media/keys/ACCC4CF8.asc | gpg --dearmor -o /etc/apt/trusted.gpg.d/postgresql.gpg \
     # Install software
     && apt-get update \
     && apt-get install -y --no-install-recommends \
         build-essential \
         libmemcached-dev \
         nodejs \
+        npm \
```
flodolo, line 29: `npm` is unavailable if not installed explicitly.

flodolo:
Just noticed `SCRIPT DEPRECATION WARNING` on the node install script. I'll take a look at that tomorrow.

docker/Dockerfile
mathjazz: Can you also update Postgres to version 15, which we use in Heroku? Maybe that'll even make thing easier?

flodolo: Side question: should we also update to Node 18, since that's what we use in frontend tests?
https://github.com/mozilla/pontoon/blob/main/.github/workflows/frontend.yml

mathjazz: Sounds good.

flodolo:
Switching to PostgreSQL 15 indeed helped. Looks like using Node 18 creates failures though, I'll need to see if I can understand why tomorrow 🤔

docker/Dockerfile
```diff
@@ -14,17 +14,18 @@ ENV PYTHONPATH /app
 # Install required software.
 RUN apt-get update \
     # Enable downloading packages over https
-    && apt-get install -y apt-transport-https \
-    # Get source for node.js
-    && curl -fsSL https://deb.nodesource.com/setup_16.x | bash - \
+    && apt-get install -y apt-transport-https ca-certificates curl gnupg; \
```
mathjazz, line 17: Seems like we don't need to install `ca-certificates curl gnupg`?

docker/Dockerfile
```diff
@@ -14,17 +14,18 @@ ENV PYTHONPATH /app
...
+    # Add source for node.js
+    curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg \
```
mathjazz, line 19: I'd just keep `&&` at the beginning of each line and drop `;` to make the code easier to edit.

mathjazz (commented):
Good stuff, thanks!

mathjazz (approved):

PR #3117, opened 2024-02-23
404 if tags don't have resources assigned

mathjazz:
Fix #3116.

PR #3119, opened 2024-02-23
Revert "Update to Debian 12, PostgreSQL 15, Node.js 18 in Docker (#3115)"

mathjazz:
This reverts commit fb25c53ddc8b984d5ccc6d445b23dddd5e9e71d8.

@flodolo Could you please give this an r+, so that I can merge it and see if collectstatic doesn't break in the GH workflow of https://github.com/mozilla/pontoon/pull/3117?

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.10-bookworm AS server
+FROM python:3.10-bullseye AS server
 
 ARG USER_ID=1000
 ARG GROUP_ID=1000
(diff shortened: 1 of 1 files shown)


flodolo:
I don't understand…
1. It is still failing even after you backed out the changes in that PR.
2. At least one of them (using latest action versions in workflow) is as safe as a change could get.

flodolo:
This has exactly the same changes you had and it worked on the first try? 🤔 
https://github.com/mozilla/pontoon/pull/3120

flodolo:
Since even this PR failed, I'm trying to re-run it to understand if it was a temporary problem in GitHub infra

flodolo:
> Since even this PR failed, I'm trying to re-running it to understand if it was a temporary problem in GitHub infra

It worked this time, so I'd suggest to restore #3117 to the single commit and see if checks pass.

mathjazz:
Interesting - seems like it was a temp issue with GH Actions.

PR #3121, opened 2024-02-23
Add warning for _any_ empty Fluent entries

### LINKED ISSUE #2977: Warn on empty Fluent rich editor fields
> In addition to the bug, I believe this is a UX regression? We should show a warning that one of the values is empty when trying to save such message.

The current practice is to consider any content in a Fluent value or attribute to qualify the whole message as not-empty. The check for this was added in #2675. Previously empty attributes may indeed have resulted in errors, due to stringifying as invalid Fluent.

_Originally posted by @eemeli in https://github.com/mozilla/pontoon/issues/2973#issuecomment-1750344569_

eemeli:
Fixes #2977

### DIFF
--- a/pontoon/checks/libraries/pontoon_db.py
+++ b/pontoon/checks/libraries/pontoon_db.py
@@ -28,20 +28,27 @@ def get_max_length(comment):
 
 class IsEmptyVisitor(Visitor):
     def __init__(self):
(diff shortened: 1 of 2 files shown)


mathjazz (approved):
LGTM

PR #3125, opened 2024-02-27
Avoid caching page headers

### LINKED ISSUE #3124: Headers on /contributors and /insights pages are cached
Headers on /contributors and /insights pages should not be cached, because they could contain user-specific information.

mathjazz:
Fix #3124.

### DIFF
--- a/pontoon/insights/views.py
+++ b/pontoon/insights/views.py
@@ -17,17 +17,27 @@ def insights(request):
     if not settings.ENABLE_INSIGHTS:
         raise ImproperlyConfigured("ENABLE_INSIGHTS variable not set in settings.")
 
-    team_pretranslation_quality = cache.get_or_set(
-        f"/{__name__}/team_pretranslation_quality",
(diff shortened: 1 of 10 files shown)


pontoon/insights/views.py
```diff
@@ -16,17 +17,25 @@ def insights(request):
...
+        f"/{__name__}/team_pretranslation_quality",
+        get_global_pretranslation_quality("locale", "code"),
+        settings.VIEW_CACHE_TIMEOUT,
+    )
+
+    project_pretranslation_quality = cache.get_or_set(
+        f"/{__name__}/project_pretranslation_quality",
+        get_global_pretranslation_quality("entity__resource__project", "slug"),
+        settings.VIEW_CACHE_TIMEOUT,
+    )
```
eemeli, line 30: Why write these in the cache? This appears to be the only place where the values are used, and with `get_or_set()` the `get_global_..._quality()` function is always called.

eemeli (approved):

PR #3126, opened 2024-02-28
Fix #2931 - New ESLint Rules & Fixes

### LINKED ISSUE #2931: JS linting should prefer const > let > var
The ESLint rules should be amended to include errors for the [no-var](https://eslint.org/docs/latest/rules/no-var) and [prefer-const](https://eslint.org/docs/latest/rules/prefer-const) rules, and the errors under `pontoon/` fixed. Most of those can be taken care of automatically with the `--fix` CLI argument.

cedricium:
Fix https://github.com/mozilla/pontoon/issues/2931.

Introduces two new ESLint rules (with their default options supplied) to be enforced when linting:

 - `no-var`
 - `prefer-const`

Additionally, fixes the ~230 issues raised by these two new rules when linting against the "pontoon/" subdirectory when running the `make eslint` command.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -70,6 +70,14 @@ module.exports = {
       { vars: 'all', args: 'after-used', ignoreRestSiblings: true },
     ],
     'no-console': 1,
+    'no-var': 'error',
+    'prefer-const': [
+      'error',
+      {
+        destructuring: 'any',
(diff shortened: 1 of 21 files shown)


eemeli (approved):

mathjazz:
Thanks for the patch and welcome to Pontoon, @cedricium! :-)

Issue #3069, opened 2024-01-12
Frontend Implementation for LLM-Assisted Translations


ayanaar:
Develop frontend components to display the new features, including alternative translations and formality options. This involves integrating the backend developments with the Pontoon frontend.

PR #3127, opened 2024-03-05
LLM-Assisted Translations Frontend 

ayanaar:
Addresses Issue #3069 

This pull request introduces the LLM-Assisted Translation feature to Pontoon, enhancing the translation workflow by incorporating LLM-powered suggestions as discussed in the [specification](https://github.com/mozilla/pontoon/blob/main/specs/0116-LLM-assisted-translations.md). The implementation spans both frontend modifications and backend integration.

**Key Features:**
- **Frontend Enhancements:** Added a new dropdown menu in the machinery tab, allowing users to select from three LLM-assisted refinement options: Rephrase, Make Formal, and Make Informal. This dropdown is integrated with Google Translate's machinery translation outputs, providing an interface for refining translations.
- **Backend Integration:** Extended the existing translation machinery API to support LLM-assisted transformations. This involves fetching original translations from Google Translate and passing them through an LLM model based on the selected refinement option. The backend ensures that the refined translations are accurately returned to the frontend for display.
- **Logging Mechanism:** Implemented a logging mechanism to track the usage of LLM-assisted translations. This includes logging the selection of refinement options, andvthe copying of LLM-refined translations into the editor. The logs aim to gather insights into feature adoption and user interaction patterns, assisting in further refinement and improvement of the tool.


translate/src/modules/machinery/components/source/dropdown.css
```diff
@@ -0,0 +1,49 @@
...
+  background-color: var(--toggle-color-1);
+  color: var(--status-translated);
+  min-width: 160px;
+  box-shadow: 0px 8px 16px 0px var(--toggle-color-1);
+  z-index: 1;
+  border-radius: 4px;
+  overflow: hidden;
+}
+
+.dropdown-menu li {
```
mathjazz, line 39: The CSS rules selector is not specific enough, so it gets overridden. Prepending it with `.machinery .translation > header ul li` should make the menu items vertical. I'd also move the CSS rules to `MachineryTranslation.css`, which will make it more obvious how specific the selectors need to be.

Issue #2965, opened 2023-09-21
Using the filter ‘Missing’ on Pontoon takes a long time.


Fjoerfoks:
I use https://pontoon.mozilla.org/fy-NL/all-projects/all-resources/?status=missing and https://pontoon.mozilla.org/nl/all-projects/all-resources/?status=missing to get to the latest missing strings in Pontoon. Since the last update loading of these results seems to have slowed down significantly.

mathjazz:
The following view was also reported to be slower:
https://pontoon.mozilla.org/he/firefox/all-resources/?extra=missing-without-unreviewed&string=275204

mathjazz:
It seems like response time of the `pontoon.base.entities` increased after we upgraded the DB from version 11 to 15.

<img width="1270" alt="Screenshot 2023-09-28 at 16 22 55" src="https://github.com/mozilla/pontoon/assets/626716/11b8eba4-12f2-4654-81f1-9f8cda47494f">

_Vertical lines represent deployments._

PR #3128, opened 2024-03-07
Only make one DB query when string param provided in URL

mathjazz:
Currently, the main DB query to load entities in the entity list is executed twice if the `string` parameter is provided in the URL. That's particularly bad, because that query is often very slow (see #2965).

This patch drops that redundant DB query and reorganizes the code to prefetch data.

### DIFF
--- a/pontoon/base/models.py
+++ b/pontoon/base/models.py
@@ -2576,11 +2576,9 @@ class EntityQuerySet(models.QuerySet):
             )
         )
 
-    def prefetch_active_translations(self, locale):
-        """
-        Prefetch active translations for given locale.
-        """
-        return self.prefetch_related(
+    def prefetch_entities_data(self, locale, preferred_source_locale):
+        # Prefetch active translations for given locale
+        entities = self.prefetch_related(
             Prefetch(
                 "translation_set",
                 queryset=(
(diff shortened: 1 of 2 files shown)


pontoon/base/views.py
```diff
@@ -188,21 +188,13 @@ def _get_paginated_entities(locale, preferred_source_locale, project, form, enti
     has_next = entities_page.has_next()
     entities_to_map = entities_page.object_list
 
-    # If requested entity not on the first page
-    if form.cleaned_data["entity"]:
-        entity_pk = form.cleaned_data["entity"]
-        entities_to_map_pks = [e.pk for e in entities_to_map]
```
mathjazz, line 194: This is where the query was executed for the first time.

pontoon/base/models.py
```diff
@@ -3040,20 +3051,19 @@ def map_entities(
...
-            entities = entities.prefetch_alternative_originals(preferred_source_locale)
+        # If requested entity not in the current page
+        if requested_entity and requested_entity not in [e.pk for e in entities]:
+            entities = list(entities) + list(
+                Entity.objects.filter(pk=requested_entity).prefetch_entities_data(
+                    locale, preferred_source_locale
+                )
+            )
 
         for entity in entities:
```
mathjazz, line 3068: This is where the query (with prefetched data) was executed for the 2nd time.

eemeli (approved):
The code looks sensible, but it's very ORM-y so a bit outside my comfort zone. If it works as intended, then it should be good to go.

PR #3129, opened 2024-03-08
Use proper paging for entities

### LINKED ISSUE #2522: Implement proper paging for entities
We currently use a weird way of paginating entities - in each request we include a list of already loaded entities, which we exclude from the matching entities in Django and then return the first page of the remaining entities.

This is a relic of the [`react-infinite-scroller`](https://github.com/mozilla/pontoon/commit/9f39c4f49597aeda95376b026e4b54c9a515d4df) library which we no longer use.

We should implement proper paging, i.e. drop `exclude_entities` and only return the relevant page.

mathjazz:
Fix #2522.

We currently use a weird way of paginating entities - in each request we include a list of already loaded entities, which we exclude from the matching entities in Django and then return the first page of the remaining entities.

This is a relic of the react-infinite-scroller library which we no longer use.

We should implement proper paging, i.e. drop exclude_entities and only return the relevant page.

That should also speed up loading of entities, because NOT IN DB queries tend to be slow and because less data will be transferred.

### DIFF
--- a/pontoon/base/forms.py
+++ b/pontoon/base/forms.py
@@ -294,6 +294,7 @@ class GetEntitiesForm(forms.Form):
     locale = forms.CharField()
     paths = forms.MultipleChoiceField(required=False)
     limit = forms.IntegerField(required=False, initial=50)
+    page = forms.IntegerField(required=False, initial=1)
     status = forms.CharField(required=False)
     extra = forms.CharField(required=False)
     tag = forms.CharField(required=False)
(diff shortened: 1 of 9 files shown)


eemeli (dismissed):
Code looks fine, but the dropped front-end and back-end tests should be replaced by equivalent ones for the new paging API.

mathjazz:
@ayanaar Since Eemeli is out, would you be able to review the added tests?

https://github.com/mozilla/pontoon/pull/3129/files#diff-194250aeeff44dce1e4085100973955c1bb0d21dbbda314e2f9c1bc1c8b82b94
https://github.com/mozilla/pontoon/pull/3129/files#diff-82a3c826352f1bde381c53dd7597bafa656276a9364c72e55360ead72f51736f

ayanaar (approved):
The tests look good to me and work as expected.

PR #3130, opened 2024-03-08
Optimize filters by introducing Resource.order

### LINKED ISSUE #2965: Using the filter ‘Missing’ on Pontoon takes a long time.

mathjazz:
_Might fix #2965. We'll see on prod if further optimizations are needed for the Missing filter._

String filters can be slow, especially because ordering entities by Resource.path is slow. It turns out that:
- Adding DB index to `Resource.path` doesn't bring performance gains.
- Sorting entities by any other `Resource.field` is fast.

The reason for slowness is probably that `Resource.path` can be a pretty long text field.

So we're now introducing `Resource.order`, to represent the alphabetic order of resources in the project. As a consequence, we now have to reorder resources when new resources are added to the project.


mathjazz:
@eemeli I still need to add tests, but might not be able to finish them before you're gone.

pontoon/base/models.py
```diff
@@ -2147,6 +2163,9 @@ def asymmetric(self):
 class Resource(models.Model):
     project = models.ForeignKey(Project, models.CASCADE, related_name="resources")
     path = models.TextField()  # Path to localization file
+    order = models.PositiveIntegerField(
+        default=0
+    )  # Index in the alphabetically sorted list of resources
```
eemeli, line 2168: As a docstring, this can show up in code completion.

```suggestion
    )
    """
    Index in the alphabetically sorted list of resources

    Sorting resources by path is a heavy operation, so we use this field
    to represent the alphabetic order of resources in the project.
    """
```

eemeli (approved):

PR #3138, opened 2024-03-12
Add .readthedocs.yaml configuration file

### LINKED ISSUE #3137: Docs are failing to build
> Error
> 
> The configuration file required to build documentation is missing from your project. Add a configuration file to your project to make it build successfully. Read more at https://docs.readthedocs.io/en/stable/config-file/v2.html
>

mathjazz:
Fix #3137.

Just adding a default configuration file, which is now required for the docs build to succeed:
https://docs.readthedocs.io/en/stable/config-file/v2.html#id1

I've uncommented the python reqs.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -8,7 +8,7 @@ version: 2
 build:
   os: ubuntu-22.04
   tools:
-    python: "3.12"
+    python: "3.10"
     # You can also specify other tool versions:
     # nodejs: "20"
     # rust: "1.70"
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,35 @@
+# Read the Docs configuration file for Sphinx projects
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the OS, Python version and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.12"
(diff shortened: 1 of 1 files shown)


.readthedocs.yaml
```diff
@@ -0,0 +1,35 @@
...
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the OS, Python version and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.12"
```
flodolo, line 11: Is there a reason to not keep this consistent with the rest of Pontoon and use 3.10?

mathjazz: I just took it from https://docs.readthedocs.io/en/stable/config-file/v2.html, assuming this is the latest version. 

Since docs environment is not running the main codebase and has different set of requirements, I wonder if we should keep the python versions of the two in sync?

flodolo: > Since docs environment is not running the main codebase and has different set of requirements, I wonder if we should keep the python versions of the two in sync?

My main concern is the fact that we have a generated list of requirements, and that might change depending on the Python version 🤔

mathjazz: These requirements are not generated using `make requirements`. I was thinking of taking care of this as part of updating the deps, but that's a separate patch.

I'm changing python version to 3.10.

flodolo (approved):

PR #3143, opened 2024-03-13
Add specs for Notification Center

bcolsson:


specs/0118-notification-center.md
```diff
@@ -0,0 +1,119 @@
...
+      - Error if `Filter by Locale` enabled but no locales selected: `You must select at least one locale.`
+  - Checkbox: `Projects`
+    - Use existing 2-panel widget already used for selecting locales in Pontoon
+    - Choose projects by moving them from `Available` to `Chosen` (all projects available)
+    - Include a `Move All` option
+  - Subheader `Filter by Activity`
+    - Button: `Add activity filter`
+      - Creates a new section of the form labeled `Activity filter` and an associated button `Remove filter`.
+      - Creates a dropdown list `Activity type` associated with this filter
+      - `Add activity filter` button persists at the bottom of the form and will still be visible after being clicked, allowing multiple filters to be created
```
mathjazz, line 54: Do we actually use the ability to combine multiple activities?

specs/0118-notification-center.md
```diff
@@ -0,0 +1,119 @@
...
+  - This “Transactional” option flags emails as transactional and sends emails to users regardless of their opt-in status. However, this email content would be restricted to content transactional in nature (e.g. account status notifications, etc.).
+  - Checkbox (default empty): `This is a transactional email`
+  - Text description: `Transactional emails are sent to users who have not opted in to email communication. Transactional emails are restricted in the type of content that can be included.`
+  - Validation:
+    - Error if `Transactional email` enabled but `Message type` does not include `Email`: `You cannnot enable the transactional email option if the Message type does not include email.`
+- Button: `Review message`
+  - A send button is not displayed on the same page as the message editor and recipient selection to avoid sending an incomplete message or a message to the wrong recipients.
+  - When pressed a validation step is performed on inputs to ensure there are no issues with the form. If there are no issues, the user proceeds to the Review/Confirmation page.
+
+- Title: `Review message`
```
mathjazz, line 90: I'm not sure there's a lot of value in having such Review step in addition to Preview.

Instead, having the ability to send the message to herself would let the user see the in-app notification or the email in the same environment the recipients will see it, revealing potential issues the Review step would hide.

specs/0118-notification-center.md
```diff
@@ -0,0 +1,119 @@
...
+ - A rendered version of the message to ensure things like Markdown tags or html don’t have issues.
+ - A summary of who will be receiving the nessage:
+    - The number of recipients
+    - The filters being applied
+  - Whether this is a transactional email. If enabled, a warning will be shown: `Warning: transactional emails are sent to users who have not opted in to email communication. Transactional emails are restricted in the type of content that can be included. When in doubt, please review with legal.`
+  - Button: `Edit`
+    - Clicking the `Edit` button or navigating back in the browser brings you to the previous page
+    - The previously entered content is preserved when `Edit` or the browser's back button is useds so work is not lost
+  - Checkbox (default empty): `I’ve reviewed the work and this is ready to be sent.`
+    - This is another safeguard to avoid accidentally sending messages without thorough review.
```
mathjazz, line 101: I like the idea of preventing sending messages accidentally. 

A checkbox in addition to the Review step seems quite involved, though.

specs/0118-notification-center.md
```diff
@@ -0,0 +1,119 @@
...
+    - Initiates the sending of the email and/or notification
+
+## Email Content
+
+When sending an email to users via the Notification Center, emails include the following content:
+- Subject: Content filled from “Subject” field in Notification Center
+- Body: Content filled from “Body” field in Notification Center (rendered in email compliant HTML)
+- Footer: Added to the bottom of all non-transactional emails: `You’re receiving this email as a contributor to Mozilla localization on Pontoon. <br>To no longer receive emails like these, unsubscribe here: <unsubscribe-link>Unsubscribe</unsubscribe-link>.`
+
+## Other
```
mathjazz, line 115: Are we going to remove the ability to see messages that have already been sent? Keeping the list of sent items would let us recycle and even make changes (in case of in-app messages).

specs/0118-notification-center.md
```diff
@@ -0,0 +1,119 @@
...
+
+Introduce a Notification Center feature to make Pontoon the best tool to send communications to localizers directly, allowing sending of targeted in-app notifications and emails within Pontoon itself.
+
+# Motivation
+
+Currently, we can only send in-app notifications with limited targeting capabilities (selected locales within a single product). For any other communications targeting a subset of users we need to rely on engineers to extract email addresses and send emails to the targeted recipients. This feature would allow administrators of a Pontoon instance managing localization projects to send notifications or emails to targeted sets of recipients with more refined criteria such as locales, projects, roles, recency of activity, and level of activity.
+
+# Feature explanation
+
+## Notification Center
```
mathjazz, line 15: Where in the UI do I access the Notification Center?

mathjazz (commented):
Nice work! Left some thoughts.

specs/0118-notification-center.md
bcolsson: At the very least we would combine `Last login` with one of the others to find people who used to be active contributors but haven't logged in recently. (As in the previous inactive contributor experiment.)

specs/0118-notification-center.md
```diff
@@ -0,0 +1,119 @@
...
+  - Creates a subject and body for emails. Subject line will be required for in-app messages; in app-messages have the subject line combined with the body message as the first line of the message separated by a newline.
+  - Field: `Subject` (text input)
+  - Field: `Body` (text input with formatting: markdown or html)
+  - Button: `Preview` - shows a rendered version of the text in `Body` for checking formatting, similar to how the preview for GitHub comments work
+  - Validation: 
+    - Error if `Subject` field empty: `Your message must include a subject.`
+    - Error if `Body` field empty: `Your message must include a body.`
+- Header: `Recipients`
+  - Select who will receive a message based on a combination of the below filters.
+  - Checkbox: `Filter by User Role` (default empty) - enables User Role as a filter.
```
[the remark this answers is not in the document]
bcolsson: That's how I had originally planned it. I'm fine either way, but I suggested this when creating the spec as it would mean constantly having to click all the user roles. remembering to move all the locales / projects to chosen if we didn't want to filter by these.
With a checkbox we the could enable them only when applicable.

specs/0118-notification-center.md
```diff
@@ -0,0 +1,119 @@
...
+- Title: `Notification Center`
+- Header: `Message type`
+  - Selects the type of message to be sent
+  - Checkboxes: `Email`, `Notification` (multiselect, default: both boxes empty)
+  - Validation: Error if no checkbox selected - `You must select at least one message type.`
+- Header: `Message editor`
+  - Creates a subject and body for emails. Subject line will be required for in-app messages; in app-messages have the subject line combined with the body message as the first line of the message separated by a newline.
+  - Field: `Subject` (text input)
+  - Field: `Body` (text input with formatting: markdown or html)
+  - Button: `Preview` - shows a rendered version of the text in `Body` for checking formatting, similar to how the preview for GitHub comments work
```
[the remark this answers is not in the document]
bcolsson: This is a good call, I'll change to `Tab` in the spec.

specs/0118-notification-center.md
bcolsson: The review step isn't just for reviewing the message, it's also a sanity check to make sure that you are sending to the people you meant to and catch any other errors - and sending to yourself isn't going to check that. Also, from a lot of past experience I will catch errors during a review step so personally I see value in it. Especially since we can't take back sending mass emails with potential errors / or incorrect recipients - the chance of making a mistake is large and the potential for impact is large as well, so being extra cautious is warranted here.

specs/0118-notification-center.md
bcolsson: See my answer above.

specs/0118-notification-center.md
bcolsson: I had considered this and thought the scope of implementing a feature like this might be too big and opted for a MVP. If you don't consider that too large, I'll add it to this spec.
(Also do you have any thoughts @flodolo based on our teams overall state in terms of bandwidth?)

specs/0118-notification-center.md
bcolsson: Very good callout - I'll need to think of this.

specs/0118-notification-center.md
```diff
@@ -0,0 +1,119 @@
...
+
+# Motivation
+
+Currently, we can only send in-app notifications with limited targeting capabilities (selected locales within a single product). For any other communications targeting a subset of users we need to rely on engineers to extract email addresses and send emails to the targeted recipients. This feature would allow administrators of a Pontoon instance managing localization projects to send notifications or emails to targeted sets of recipients with more refined criteria such as locales, projects, roles, recency of activity, and level of activity.
+
+# Feature explanation
+
+## Notification Center
+
+This feature has the ability to compose a message with a subject and body, send a message as either an in-app notification or email, and target a set of recipients based on commonly used criteria.
```
flodolo, line 17:
```suggestion
This feature has the ability to compose a message with a subject and body, send a message as an in-app notification and/or email, and target a set of recipients based on commonly used criteria.
```

Nitpick, but I think _either_ makes it exclusive.

specs/0118-notification-center.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Required functionality:
+
+- Title: `Notification Center`
+- Header: `Message type`
+  - Selects the type of message to be sent
+  - Checkboxes: `Email`, `Notification` (multiselect, default: both boxes empty)
```
flodolo, line 24:
```suggestion
  - Checkboxes: `Email`, `Notification` (default: both checkboxes deselected)
```

Checkboxes are by nature multiselect (there's no concept of group, that would become a radio buttons group).

Same comment applies where this formula is repeated later on.

specs/0118-notification-center.md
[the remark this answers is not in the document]
flodolo: +1 on using checkboxes on the "rule". If we don't want to filter by user role, we don't select the specific rule, and target all of them. That's a lot faster and, at least for me, clearer.

P.S. empty ⇾ deselected.

specs/0118-notification-center.md
```diff
@@ -0,0 +1,119 @@
...
+- Header: `Message editor`
+  - Creates a subject and body for emails. Subject line will be required for in-app messages; in app-messages have the subject line combined with the body message as the first line of the message separated by a newline.
+  - Field: `Subject` (text input)
+  - Field: `Body` (text input with formatting: markdown or html)
+  - Button: `Preview` - shows a rendered version of the text in `Body` for checking formatting, similar to how the preview for GitHub comments work
+  - Validation: 
+    - Error if `Subject` field empty: `Your message must include a subject.`
+    - Error if `Body` field empty: `Your message must include a body.`
+- Header: `Recipients`
+  - Select who will receive a message based on a combination of the below filters.
```
flodolo, line 35:
```suggestion
  - Select who will receive a message based on a combination of the following filters.
```

specs/0118-notification-center.md
flodolo: `Notifications` in the top menu, since we're going to get rid of the notifications tabs in project page?

specs/0118-notification-center.md
```diff
@@ -0,0 +1,119 @@
...
+  - Checkbox: `Filter by User Role` (default empty) - enables User Role as a filter.
+    - Checkboxes: `Manager`, `Translator`, `Contributor` (multiselect, default: all boxes empty)
+    - Validation:
+      - Error if `Filter by User Role` enabled but no roles selected: `You must select at least one user role.`
+  - Checkbox: `Filter by Locale` (default empty) - enables Locale as a filter.
+    - Use existing 2-panel widget already used for selecting locales in Pontoon
+    - Choose locales by moving them from `Available` to `Chosen`(all locales available)
```
flodolo, line 42:
```suggestion
    - Choose locales by moving them from `Available` to `Chosen` (all locales available)
```

specs/0118-notification-center.md
```diff
@@ -0,0 +1,119 @@
...
+    - Include a `Move All` option
+    - Validation:
+      - Error if `Filter by Locale` enabled but no locales selected: `You must select at least one locale.`
+  - Checkbox: `Projects`
+    - Use existing 2-panel widget already used for selecting locales in Pontoon
```
flodolo, line 47: Did you actually mean "locales" here?

specs/0118-notification-center.md
```diff
@@ -0,0 +1,119 @@
...
+      - Creates a new section of the form labeled `Activity filter` and an associated button `Remove filter`.
+      - Creates a dropdown list `Activity type` associated with this filter
+      - `Add activity filter` button persists at the bottom of the form and will still be visible after being clicked, allowing multiple filters to be created
+    - Dropdown list: `Activity type`
+      - Options: `Select activity type`, `Submitted translations`, `Performed reviews`, `Last login`
+      - Select one
+      - Default value: `Select activity type`
+      - Form dynamically displays the appropriate filter inputs below based on the selection in the dropdown. `Select activity type` does not display any filter fields.
+      - Validation:
+        - Multiple activity filters can used at the same time
```
flodolo, line 61:
```suggestion
        - Multiple activity filters can be used at the same time
```

specs/0118-notification-center.md
```diff
@@ -0,0 +1,119 @@
...
+        - Error if more than one filter of the same type used: `You cannot use multiple of the same activity type.`
```
flodolo, line 62: Can we prevent this by populating the dropdown dynamically? If you've already selected "Last login", that disappears from the list.

specs/0118-notification-center.md
```diff
@@ -0,0 +1,119 @@
...
+        - Error if `Select activity type` selected as a filter in the dropdown: `You must select an activity type when you add an activity filter.`
```
flodolo, line 63: Similar: the button stays disabled until you select an actual filter.

specs/0118-notification-center.md
```diff
@@ -0,0 +1,119 @@
...
+    - Filter inputs:
+      - Inputs for `Submitted translations` / `Performed reviews`
+        - `Activity level`
+          - Input box: `Minimum`  (value required, default value: 0, only accepts non-negative integers)
```
flodolo, line 67:
```suggestion
          - Input box: `Minimum` (value required, default value: 0, only accepts non-negative integers)
```

specs/0118-notification-center.md
```diff
@@ -0,0 +1,119 @@
...
+          - Input box: `Maximum` (value optional, default value: empty, leaving empty will mean there is no maximum and all numbers greater than or equal to the value in `Minimum` will be included.
+          - Tooltip on `Maximum`: `Leave empty to include all numbers greater than or equal to Minimum.`
+        - `Timeframe filter`
+          - Checkbox (default empty): `Enable timeframe filter`
+            - Checking the box displays datepickers
+          - Datepicker: `From:` (Default value: today's date)
```
flodolo, line 73: We should pick a date in the past for `From` (1 year?)

specs/0118-notification-center.md
```diff
@@ -0,0 +1,119 @@
...
+        - `Timeframe filter`
+          - Datepicker: `From:` (Default value: today's date)
+          - Datepicker: `To:` (Default value: today's date)
+- Header: `Transactional email`
+  - By default, emails are sent to users who have enabled opt-in consent. (Does not apply to in-app notifications.)
+  - This “Transactional” option flags emails as transactional and sends emails to users regardless of their opt-in status. However, this email content would be restricted to content transactional in nature (e.g. account status notifications, etc.).
+  - Checkbox (default empty): `This is a transactional email`
+  - Text description: `Transactional emails are sent to users who have not opted in to email communication. Transactional emails are restricted in the type of content that can be included.`
+  - Validation:
+    - Error if `Transactional email` enabled but `Message type` does not include `Email`: `You cannnot enable the transactional email option if the Message type does not include email.`
```
flodolo, line 85: Again: keep the checkbox disabled unless email is selected?

specs/0118-notification-center.md
flodolo: +1 on having an explicit review step on top of preview (note that it will display the number of recipients, which we probably don't want to calculate on the fly).

With that said, I like the idea of also having a button to send the message to yourself.

specs/0118-notification-center.md
```diff
@@ -0,0 +1,119 @@
...
+- Title: `Review message`
+- A review step is required to reduce errors because messages could be sent to a large number of people. This page allows you to review all aspects of the email before sending, including:
+ - A rendered version of the message to ensure things like Markdown tags or html don’t have issues.
+ - A summary of who will be receiving the nessage:
+    - The number of recipients
+    - The filters being applied
+  - Whether this is a transactional email. If enabled, a warning will be shown: `Warning: transactional emails are sent to users who have not opted in to email communication. Transactional emails are restricted in the type of content that can be included. When in doubt, please review with legal.`
+  - Button: `Edit`
+    - Clicking the `Edit` button or navigating back in the browser brings you to the previous page
+    - The previously entered content is preserved when `Edit` or the browser's back button is useds so work is not lost
```
flodolo, line 99:
```suggestion
    - The previously entered content is preserved when `Edit` or the browser's back button is used so work is not lost
```

specs/0118-notification-center.md
flodolo: Every single one of us has made mistakes before, especially with notifications. I prefer to err on the side of caution.

specs/0118-notification-center.md
flodolo: I agree that we shouldn't include this in the first iteration:
* In-app notifications: right now they're duplicated across the DB (one copy for recipient). Are you thinking of refactoring notifications as a preliminary step for this feature? Otherwise, I don't see how you would be able to correct a past notification.
* I expect more complex messages to live in documents, if they need to be discussed.

flodolo (commented):
Overall it looks good to me, mostly nitpicks.

specs/0118-notification-center.md
flodolo, line 47:
```suggestion
    - Use existing 2-panel widget like the one already used for selecting locales in Pontoon
```

Maybe this is less confusing, since this is about projects.

specs/0118-notification-center.md
mathjazz: I didn't phrase my initial point exactly as I wanted. Let's try again. :)

I agree the Review step is important, but I'm wondering if we need Review AND Preview (AND Ability to send a test message). I know in the ideal World we'd have all, but this spec if already quite big.

Sending a **test message** is IMO a must, because that's the only way to see how the email or the notification actually renders. I use it for every mass email I send from Pontoon and we make corrections in about 50% of cases.

**Review**, as Bryan mentions, also checks for errors outside the message content. I'd keep that, too.

**Preview** on the other hand is not as reliable as a test message, and is enforced by the Review step. Yes, it lets you check you message instantly, but that's not as critical as the other two. So I'd treat is as optional.

specs/0118-notification-center.md
```diff
@@ -0,0 +1,122 @@
...
+  - This “Transactional” option flags emails as transactional and sends emails to users regardless of their opt-in status. However, this email content would be restricted to content transactional in nature (e.g. account status notifications, etc.).
+  - Checkbox (default: deselected): `This is a transactional email`
+  - Text description: `Transactional emails are sent to users who have not opted in to email communication. Transactional emails are restricted in the type of content that can be included.`
+  - Validation:
+    - Disable `Review message` button and show error message if `Transactional email` enabled but `Message type` does not include `Email`: `You cannnot enable the transactional email option if the Message type does not include email.`
+- Button: `Review message`
+  - A send button is not displayed on the same page as the message editor and recipient selection to avoid sending an incomplete message or a message to the wrong recipients.
+  - When pressed a validation step is performed on inputs to ensure there are no issues with the form. If there are no issues, the user proceeds to the Review/Confirmation page.
+
+- Title: `Review message`
```
mathjazz, line 91: Also, the list of bullet points in this section is very long and hard to follow. It should help a bit to split the Feature explanation into Compose and Review sections.

specs/0118-notification-center.md
mathjazz: Can you share an example of such UX pattern from a popular application? Where you are already on the Review step, and you need to click a ("This is ready to be sent") checkbox before clicking Send?

I don't remember anything like that from my email clients, web shops or online banks (which, TBF, usually require an extra step on another device, but that's due to authentication, not sanity check).

Not even professional marketing service [MailerLite](https://www.mailerlite.com/) uses that (I didn't check others). And we're building just an MVP of a tool like MailerLite.

I 100% understand we'd like to prevent mistakes, but I don't see how this checkbox helps with that. I'm happy to be convinced otherwise with good examples. 😄

specs/0118-notification-center.md
```diff
@@ -0,0 +1,122 @@
...
+    - Checkboxes: `Manager`, `Translator`, `Contributor` (multiselect, default: all boxes empty)
+    - Validation:
+      - Error if `Filter by User Role` enabled but no roles selected: `You must select at least one user role.`
+  - Checkbox: `Filter by Locale` (default: deselected) - enables Locale as a filter.
+    - Use existing 2-panel widget like the one already used for selecting locales in Pontoon
+    - Choose locales by moving them from `Available` to `Chosen` (all locales available)
+    - Include a `Move All` option
+    - Validation:
+      - Error if `Filter by Locale` enabled but no locales selected: `You must select at least one locale.`
+  - Checkbox: `Projects`
```
mathjazz, line 47: Filter by Project

specs/0118-notification-center.md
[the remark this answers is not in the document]
mathjazz: I have a few concerns about checkboxes in the current version of the spec:
* You can select "Filter by Role" and then don't select any Role. Not great UX, which requires validation.
* Some headers have checkboxes, some don't. It's not clear why the difference.
* The current version of the spec suggests lots of checkboxes (10).
* Having all checkboxes selected yields the same results as as not having any checkbox selected.

_This is the default state of the User Role filters according to the spec:_
Screenshot 2024-03-29 at 18 21 53

_This is the default state of the User Role filters without the checkbox in the header:_
Screenshot 2024-03-29 at 18 19 37

> +1 on using checkboxes on the "rule". If we don't want to filter by user role, we don't select the specific rule, and target all of them. That's a lot faster and, at least for me, clearer.

To target all roles, you don't do anything in either case, so both options are equally fast.

Are empty checkboxes in front of all user roles really making it clearer that you're targeting all roles?

specs/0118-notification-center.md
mathjazz: For pretty much every email I have sent from Pontoon we had to chase Google Docs to see what exactly was the target audience for a previous similar email. Having all that stored in one place would be very helpful, and it's not that hard to implement, given we already have a similar functionality.

We can edit in-app notifications without changing the model (although we should do that, too). But I would put the editing part in the [Out of scope](https://github.com/mozilla/pontoon/blob/main/specs/0000-template.md#optional-sections) section. With all the Review capabilities I also hope there will be less need to actually edit notifications.

specs/0118-notification-center.md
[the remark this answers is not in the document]
flodolo: It's clearer for me, because I'm effectively choosing not to "filter by user role" if the checkbox is unchecked (and the checkboxes for each user role would be disabled in that case). Same for other rules: if it's not selected, the associated controls would be disabled. 

With that said, as long as the functionality is there, I can live with either solutions, also considering the users are internal and just need to learn how to use the feature. I also see how the second requires less validation.

Looking at your screenshots, I think we need something to clarify that the selected roles are included when selected (especially in the second). Filter is ambiguous (include or exclude?).

specs/0118-notification-center.md
bcolsson: So what's your suggestion to add here? Specifically, what are you imaging when you say it's not hard to implement/have similar functionality?

specs/0118-notification-center.md
bcolsson: I've removed this check for now. We could argue either way but we have at least 2 clicks before an email gets sent which is at least the bare minimum.

specs/0118-notification-center.md
bcolsson: Changed to `Messaging Center`

specs/0118-notification-center.md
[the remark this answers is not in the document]
bcolsson: I've eliminated the checkboxes and changed `Filter by user role` to `Included User Roles`.

specs/0118-notification-center.md
bcolsson: So, to clarify, are you suggesting to remove the rendered view of the message in the Review step, and the "Preview" tab from the messaging editor?

specs/0118-notification-center.md
```diff
@@ -0,0 +1,134 @@
...
+- A review step is required to reduce errors because messages could be sent to a large number of people. This page allows you to review all aspects of the email before sending, including:
+ - A rendered version of the message to ensure things like Markdown tags or html don’t have issues.
+ - A summary of who will be receiving the nessage:
+    - The number of recipients
+    - The filters being applied
+  - Whether this is a transactional email. If enabled, a warning will be shown: `Warning: transactional emails are sent to users who have not opted in to email communication. Transactional emails are restricted in the type of content that can be included. When in doubt, please review with legal.`
+  - Button: `Send test to myself`
+    - Sends the current message to the user creating the message in the same format(s) as it is being sent.
+  - Button: `Edit`
+    - Clicking the ` Edit` button or navigating back in the browser brings you to the previous page
```
mathjazz, line 117:
```suggestion
    - Clicking the `Edit` button or navigating back in the browser brings you to the previous page
```

specs/0118-notification-center.md
```diff
@@ -0,0 +1,134 @@
+- Feature Name: Messaging Center
```
mathjazz, line 1: Nit: we should also rename the file to reflect the name name.

specs/0118-notification-center.md
```diff
@@ -0,0 +1,134 @@
...
+- Created: 2024-03-12
+- Associated Issue:
```
mathjazz, line 3: We should assign an issue.

specs/0118-notification-center.md
```diff
@@ -0,0 +1,122 @@
...
+
+This feature has the ability to compose a message with a subject and body, send a message as an in-app notification and/or email, and target a set of recipients based on commonly used criteria.
+
+- Top menu item: `Notification Center`
+  - Item that appears in the top menu only for those with administrator priveleges
+  - Navigates to the `Notification Center` page.
+- Title: `Notification Center`
+- Header: `Message type`
+  - Selects the type of message to be sent
+  - Checkboxes: `Email`, `Notification` (default: both boxes deselected)
```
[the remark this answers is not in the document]
mathjazz: That makes sense, thanks!

specs/0118-notification-center.md
```diff
@@ -0,0 +1,134 @@
...
+
+#### Selecting Recipients with Filters
+
+- Header: `Recipients`
+  - Select who will receive a message based on a combination of the following filters.
+  - `Included User Roles` - enables User Role as a filter.
+    - Checkboxes: `Manager`, `Translator`, `Contributor` (multiselect, default: all boxes selected)
+    - Validation:
+      - Error if `Included User Roles` has no roles selected: `You must select at least one user role.`
+  - Subheader: `Filter by Locale`
```
mathjazz, line 48: Default value is not specified. I assume it should be all Locales selected. And same for Projects below.

specs/0118-notification-center.md
```diff
@@ -0,0 +1,134 @@
...
+  - Button: `Edit`
+    - Clicking the ` Edit` button or navigating back in the browser brings you to the previous page
+    - The previously entered content is preserved when `Edit` or the browser's back button is used so work is not lost
+  - Button: `Send`
+    - Triggers validation step
+    - Initiates the sending of the email and/or notification
+
+## Email Content
+
+When sending an email to users via the Notification Center, emails include the following content:
```
mathjazz, line 125: Nit: We should also rename "Notification Center" to "Messaging Center" in this section. Or even drop it?

specs/0118-notification-center.md
mathjazz: Just the "Preview" tab.

specs/0118-notification-center.md
mathjazz: "Inbox" - the ability to see messages that have already been sent.

bcolsson: @mathjazz - I've added a basic idea of what this "inbox" could be. Could you add suggestions to this to align with what you were envisioning?

mathjazz (approved):
Looks good!

The only thing I'd keep is the ability to switch between Compose and Sent view:
https://pontoon.mozilla.org/projects/firefox/notifications/

The Compose screen (Editor) would be the default view when the Messaging page loads and one would be able to switch to the Sent view.

bcolsson:
@mathjazz - Thanks, I've updated the spec to reflect your last comment. Talked with @flodolo who said this doesn't need additional review by him. So this spec should be okay to proceed with. If you're okay with this can you merge? (I don't have write to Pontoon project.)

mathjazz (approved):
Thanks for the updates and for keeping up with the review process. :)

Great work!

PR #3146, opened 2024-03-13
Speed up Project Notifications tab

### LINKED ISSUE #3145: Speed up Project Notifications tab
The Notifications tab takes a while to load if multiple manual notifications have been sent to project contributors.

mathjazz:
Fix #3145.

We make the Project Notifications tab faster by loading past Project Notifications more efficiently. We do so by retrieving only one past project Notification per instance, and by prefetching related Project and User data.

Each project notification is stored in one Notification instance per user. To identify unique project Notifications, we now use the identifier stored in the Notification.data field.

PostgreSQL allows us to retrieve Notifications with unique Notification.data fields by combining .order_by(*fields) and .distinct(*fields) calls. Read more: https://docs.djangoproject.com/en/3.2/ref/models/querysets/#distinct

That approach doesn't allow us to order Notifications by their timestamp, so we have to do that with Python.

### DIFF
--- a/pontoon/projects/views.py
+++ b/pontoon/projects/views.py
@@ -186,20 +186,27 @@ def ajax_notifications(request, slug):
                 identifier=identifier,
             )
 
-    # Detect previously sent notifications using a unique identifier
-    # TODO: We should simplify this with a custom Notifications model
-    notifications_map = {}
+    notifications = list(
+        Notification.objects.filter(
+            description__isnull=False,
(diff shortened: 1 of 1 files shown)


eemeli (approved):
With pure Postgres the sorting could be done using a [`WITH` query](https://www.postgresql.org/docs/current/queries-with.html), but I don't think that's supported by Django.

PR #3147, opened 2024-03-14
Speed up /notifications page

### LINKED ISSUE #2278: Notifications page takes forever to load, or causes application error
## [Bug 1699610](https://bugzilla.mozilla.org/show_bug.cgi?id=1699610)
Bug Reporter: @flodolo
CC:   @mathjazz

My notifications page takes forever to load, and in some cases triggers an application error. I guess I'm an edge case, having too many notifications (2703 right now).

The resulting HTML is 1.65 MB, and takes 17 seconds to load.

In particular, it looks like the page has no pagination for results, which might be the main cause.

mathjazz:
Fix #2278.

The patch comes in 2 parts:

Part 1 improves performance by significantly reducing the number of DB calls through prefetching related data.

Part 2 improves perceived performance by only loading the first 100 notifications initially. Any remaining notifications are loaded afterwards via AJAX. That makes the page usable much faster, and by the time the user scrolls to the bottom of the page, the remaining notifications are likely to be loaded already.

### DIFF
--- a/pontoon/contributors/static/js/notifications.js
+++ b/pontoon/contributors/static/js/notifications.js
@@ -1,4 +1,6 @@
 $(function () {
+  const self = Pontoon;
+
   window.history.replaceState(
     {},
     document.title,
@@ -36,5 +38,25 @@ $(function () {
     setTimeout(function () {
       Pontoon.markAllNotificationsAsRead();
     }, 1000);
+  }
+
+  // Load remaining notifications
+  if ($('#server').data('hasMore')) {
+    self.NProgressUnbind();
+
+    $.ajax({
+      url: '/ajax/notifications/',
+      success: function (data) {
+        $('#main .notification-list').append(data);
+
+        // Re-apply notification filters
+        $('.left-column .selected:not(.all) a').trigger('click');
+      },
+      error: function () {
+        Pontoon.endLoader('Oops, something went wrong.', 'error');
+      },
+    });
+
+    self.NProgressBind();
   }
 });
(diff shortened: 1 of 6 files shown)


mathjazz:
@flodolo Could you please compare the performance of https://pontoon.mozilla.org/notifications/ and https://mozilla-pontoon-staging.herokuapp.com/notifications/ (where this patch is deployed)?

flodolo:
On prod the document is 3.9 MB and takes on average 8s to load. 
On stage it's ~130 kB and takes about 1s, so definitely better.

Not sure I understand how that's possible, given all notifications are still there? 🤯

mathjazz:
Thanks for testing!

See the 2nd commit message for explanation (spoiler: the request it split into two).

pontoon/contributors/static/js/notifications.js
```diff
@@ -37,4 +39,24 @@ $(function () {
       Pontoon.markAllNotificationsAsRead();
     }, 1000);
   }
+
+  // Load remaining notifications
+  if ($('#server').data('hasMore')) {
```
eemeli, line 44: The jQuery `.data()` does not parse the dataset contents, so that if the attribute value is `data-has-more="false"` or any other valid JSON value that's not an empty string, this condition will evaluate as true. An explicit check might be required:

```suggestion
  if ($('#server').data('hasMore') === 'true') {
```

eemeli (approved):
See inline quibble, otherwise looks ok.

pontoon/contributors/static/js/notifications.js
mathjazz: _Every attempt is made to convert the attribute's string value to a JavaScript value (this includes booleans, numbers, objects, arrays, and null)._

https://api.jquery.com/data/#data-html5

PR #3164, opened 2024-04-03
Do not use an inner container for CI backend tests

eemeli:
Relying on the normal CI env and not building a docker container and running the backend tests in it reduces their duration from about 6.5 minutes to 3.5 minutes.

The special-casing of Turkish collation is also dropped here from test setup as unnecessary; the same behaviour is available using the `"tr-x-icu"` collation.


flodolo:
I've seen the Turkish collation error [before](https://github.com/mozilla/pontoon/pull/1780)

mathjazz (approved):
LGTM

@flodolo Do you want to take a peek?

.github/workflows/backend.yml
```diff
@@ -23,13 +23,48 @@ jobs:
   test:
     name: pytest
     runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:13.4
```
flodolo, line 28: Should this use 15 instead of 13? We're using 15.6 in production, and I'm surprised to still see 13.4 [here](https://github.com/mozilla/pontoon/blob/ee29b881b24d6382ec6fbbad84d3779cd4c5b105/docker/postgres/Dockerfile#L1) (but 15 [here](https://github.com/mozilla/pontoon/blob/ee29b881b24d6382ec6fbbad84d3779cd4c5b105/docker/Dockerfile#L28)).

mathjazz: Good catch, let's use PG 15 everywhere.

eemeli:
```suggestion
        image: postgres:15
```

flodolo (approved):
This has already plenty of changes.

As mentioned, I'm not sure if we should update [this file](https://github.com/mozilla/pontoon/blob/main/docker/postgres/Dockerfile), but I'm OK in doing it separately:
* 13 -> 15
* Do we need that tr-TR line once this lands?

eemeli:
> As mentioned, I'm not sure if we should update [this file](https://github.com/mozilla/pontoon/blob/main/docker/postgres/Dockerfile), but I'm OK in doing it separately:
> * 13 -> 15
> * Do we need that tr-TR line once this lands?

Once this lands, we don't need any tuning for Turkish, so the PR for #3165 can switch us to using the base `postgres:15` Docker image directly.

PR #3160, opened 2024-04-02
Project Admin page optimization

mathjazz:
Fix #3083.

The main change here is to [Retrieve project locales asynchronously to speed up loading of the page](https://github.com/mozilla/pontoon/commit/f43e01863abbf36dcdbf0bdabe88899b7879bee4).

_Updated: deployed to stage._

PR #3166, opened 2024-04-03
Disable loading animation on Project Admin page load

mathjazz:
This is a followup from #3160.

There should be no loading animation on page load while data is being fetched in the background.


mathjazz:
Asuming the build failing as usual.

PR #3168, opened 2024-04-04
Keep cache warm

### LINKED ISSUE #3155: Keep cache for Contributors views warm
We cache data for all Contributors pages for a day.

Some of the Contributors pages don't get a lot of visits, not even one per day. So the visitors of these pages often hit the cold cache.

We should keep the cache warm via a cron job.

mathjazz:
Fix #3155.

We cache data for some of the views (e.g. Contributors) for a day. Some of them don't get a lot of visits, not even one per day, meaning that the visitors of these pages often hit the cold cache.

This command refreshes data in the cache every day, because it changes often. The command is designed to run daily.

_Deployed to stage._

### DIFF
--- a/docs/admin/deployment.rst
+++ b/docs/admin/deployment.rst
@@ -400,6 +400,16 @@ is designed to run in the beginning of the day, every day.
 
    ./manage.py collect_insights
 
+Warm up cache
+~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


eemeli (approved):
Thanks, I hate this.

eemeli (approved):
Hackity hack hack.

PR #3169, opened 2024-04-05
Drop the custom tr_tr collation

### LINKED ISSUE #3165: Collation specialization appears to be out of date
We currently build our own [postgres docker image](https://github.com/mozilla/pontoon/blob/main/docker/postgres/Dockerfile) and do [extra steps during test setup](https://github.com/mozilla/pontoon/blob/main/pontoon/test/setup.py) to help ensure that we have a `tr_tr` collation available in Postgres. This was added to resolve [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1346180) in 2017.

Reviewing the situation now, it looks like the same behaviour is available "out of the box" with our current Django and Postgres versions by using the `tr-x-icu` collation. We should despecialize what we're doing and use that (and other ICU-backed locales) where appropriate.

eemeli:
Fixes #3165 

This is a prerequisite to being able to simplify our dev env postgres container.

Because we currently include a
```sql
CREATE COLLATION tr_tr (LOCALE = 'tr_TR.utf8');
```
command in our database initialization, it means that the results of a `pg_dump` or `pg_upgrade` will include that as well, and that means that any environment where the data is restored is expected to include a `tr_TR.utf8` locale, which is itself non-standard. In other words, migrating dev environment data to a not-custom postgres instance is a pain.

So let's drop the collation, and wherever it's used, switch to `tr-x-icu` instead.

This also needs to be a SQL migration, because it's touching stuff that's not accessible via Django.

### DIFF
--- a/pontoon/base/migrations/0058_remove_tr_tr_collation.py
+++ b/pontoon/base/migrations/0058_remove_tr_tr_collation.py
@@ -1,0 +1,20 @@
+# Generated manually on 2024-04-04
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("base", "0057_remove_lang_format"),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            sql="UPDATE base_locale SET db_collation='tr-x-icu' WHERE db_collation='tr_tr';",
+        ),
+        migrations.RunSQL(
+            sql="DROP COLLATION IF EXISTS tr_tr;",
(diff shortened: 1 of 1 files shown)


eemeli:
Deployed to staging, where I can confirm that e.g. this search still produces the expected results: https://mozilla-pontoon-staging.herokuapp.com/tr/amo/LC_MESSAGES/django.po/?search=itiraz&string=286656

mathjazz (approved):
Nice work!

PR #3170, opened 2024-04-05
Use postgres:15 instead of custom postgres:13.4 build for dev env

eemeli:
This drops the custom postgres dev container build originally introduced in #866 in favour of using the upstream `postgres` image directly. While at it, let's also update to v15.

This change should not be applied until #3169 has landed, and anyone who cares has had an opportunity to apply its migration, so that the `pg_upgrade` or `pg_dump`/`pg_restore` this PR necessitates applies cleanly.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -24,9 +24,7 @@ services:
 
   # Database
   postgresql:
-    build:
-      context: ./docker/postgres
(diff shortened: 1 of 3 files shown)


mathjazz (approved):

flodolo:
Should this land, so that the LLM PR can be rebased?

PR #3179, opened 2024-04-17
During sync, fall back to reporting all files as changed if the hg/git/svn cli command fails

### LINKED ISSUE #3178: The repo last_synced_revisions value is not validated
Pontoon keeps track of the last synced revision identifier, and uses this when [determining](https://github.com/mozilla/pontoon/blob/bfc3f8e0d6748aafd71f8e8155a517308fa2bbc3/pontoon/sync/vcs/repositories.py#L458) what files have changed. However, if the repo has changed and that commit/changeset id is no longer valid, the current fallback behaviour is to swallow the error and to return an empty list. This means that any changed or removed files won't be detected and updated until they next change.

On error, we should probably default to returning all the files, which is what's done if the revision id is missing.

eemeli:
Fixes #3178 

Tested manually by setting the data to be invalid in a local install:

```sql
update base_repository set last_synced_revisions = '{"single_locale": "foobar"}' where id = ...;
```

### DIFF
--- a/pontoon/sync/tests/test_vcs.py
+++ b/pontoon/sync/tests/test_vcs.py
@@ -69,7 +69,7 @@ class VCSChangedFilesTests:
         with patch.object(
             self.vcsrepository, "execute", side_effect=self.execute_failure
         ) as mock_execute:
-            assert self.vcsrepository.get_changed_files("path", "1") == []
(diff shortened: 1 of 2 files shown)


mathjazz (approved):
Nice work!

We should monitor sync performance to see if this change results in any false positives.

PR #3159, opened 2024-04-02
LLM-Assisted Translations Frontend (rebased)

mathjazz:
This is a continuation of https://github.com/mozilla/pontoon/pull/3127, which got stuck due to a bad rebase.

I've rebased it against main and pushed to a new branch to be able to roll back. For all conflicts, I just accepted incoming changes and added [40bf35e](https://github.com/mozilla/pontoon/pull/3159/commits/40bf35e3c94f43ccf3e330661fad75911571f1f5).

@ayanaar Could you please check that this works as expected and contains all and only relevant changes? If yes, we should close the original PR.

PR #3181, opened 2024-04-17
Do not make requests to non-configured Machinery services

mathjazz:
Translate view makes requests to Google Translate and Microsoft Translator backend code even if API keys are not set. While the requests to external services are not made in this case, we should still prevent the redundant call to Pontoon backend.

### DIFF
--- a/translate/src/context/MachineryTranslations.tsx
+++ b/translate/src/context/MachineryTranslations.tsx
@@ -99,15 +99,24 @@ export function MachineryProvider({
 
       // Only make requests to paid services if user is authenticated
       if (isAuthenticated) {
-        if (locale.googleTranslateCode) {
+        const root = document.getElementById('root');
+
+        const isGoogleTranslateSupported =
+          root?.dataset.isGoogleTranslateSupported === 'true';
+        const isMicrosoftTranslatorSupported =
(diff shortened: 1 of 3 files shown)


mathjazz:
This will make it easier to disable the LLM-specific code in https://github.com/mozilla/pontoon/pull/3159.

translate/src/context/MachineryTranslations.tsx
```diff
@@ -99,15 +99,24 @@ export function MachineryProvider({
 
       // Only make requests to paid services if user is authenticated
       if (isAuthenticated) {
-        if (locale.googleTranslateCode) {
+        const root = document.getElementById('root');
+
+        const isGoogleTranslateSupported =
+          root?.dataset.isGoogleTranslateSupported === 'true';
```
mathjazz, line 105: Probably better to use `JSON.parse()`? It'll be safer, but the there will be a bit more code.

eemeli: No, this is better

eemeli (approved):
You get extra bonus points for finding good excuses to use kebab-case, lower_case, _and_ camelCase for the same variables.

PR #3183, opened 2024-04-22
Upgrade Django to 4.2.11

### LINKED ISSUE #3059: Upgrade Django to 4.2 (LTS)

mathjazz:
Fix #3059.

The first commit addresses the deprecation warning in Django 3.2.

The next batch of commits updates Django dependencies.

Each _Upgrade Django to 4.X.Y_ commit only updates Django. Commits that follow address issues on build time, on run time and in tests.

The patch is currently deployed to [stage](https://mozilla-pontoon-staging.herokuapp.com/) for testing. I don't see any issues in the app or sync.

The main (and perhaps only) breaking change for existing deployments is that the [ALLOWED_HOSTS](https://github.com/mozilla/pontoon/pull/3183/commits/add84da1a9759e3d758d8978c48734086a892101) variable needs to be explicitly set. It is no longer retrieved from the `SITE_URL` variable.


pontoon/base/utils.py
```diff
@@ -4,7 +4,6 @@
 import os
 import re
 
-import pytz
```
eemeli, line 7: I think this is the only place where we use `pytz` directly, so could drop it from `requirements/default.in`?

eemeli (approved):
Changes look reasonable, and if tests pass and it runs fine on staging, it's probably fine.

PR #3184, opened 2024-04-22
Drop version from docker-compose.yml, because it is obsolete

mathjazz:
This removes the following warning on `make run`:

> WARN[0000] ~/pontoon/docker-compose.yml: `version` is obsolete

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,6 @@
 # docker-compose for Pontoon development.
 #
-# Note: Requires docker-compose 1.10+.
+# Note: Requires docker-compose 1.27+.
 services:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -1,7 +1,6 @@
 # docker-compose for Pontoon development.
 #
 # Note: Requires docker-compose 1.10+.
```
eemeli, line 3: The `version`-less version has a higher minimum bar.

```suggestion
# Note: Requires docker-compose 1.27+.
```

eemeli (approved):
See inline for a small required fix.

PR #3186, opened 2024-04-23
A light refactor of pontoon/sync/ code

eemeli:
This is a first pass at making the sync code a bit more readable to an outsider.

A smattering of type hints is added, but these aren't (yet) checked or anywhere near complete; they're just to help out the reader and editors get a better understanding of the shapes of the data that's being dealt with. Once #3183 lands, it should get easier to see if [django-stubs](https://github.com/typeddjango/django-stubs) could prove useful for better typing of e.g. querysets.

A few functions are refactored to be more pythonic.

The git/hg/svn implementations under `pontoon/sync/vcs/repositories` are each split into their own modules, and their unnecessary wrapper classes are dropped. Previously, the classes were being found using `globals()` and by string-concatenating repo type identifiers.

The `pontoon/sync/tests/test_vcs.py` tests are updated to match, and expanded to cover the `revision()` functions.

### DIFF
--- a/pontoon/sync/changeset.py
+++ b/pontoon/sync/changeset.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 import logging
 
 from collections import defaultdict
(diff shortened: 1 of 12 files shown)


eemeli:
Added some further refactoring of `pontoon.sync.vcs`, just moving code around.

mathjazz (commented):
Changes look good, thanks!

The only thing I believe needs a broader discussion is the introduction of type hints, so I'd leave that out of this PR.

mathjazz (approved):

mathjazz:
We agreed to move on with the type hints and revisit the discussion on them once the refactor is complete.

PR #3206, opened 2024-05-08
Resolved fixable warnings when running pytest

### LINKED ISSUE #3190: Resolve deprecation warnings when running pytest
There are quite a few warnings when running `make pytest` (future Django deprecations, dates without timezones).

We should get rid of them.

harmitgoswami:
Fix #3190

Removed deprecated packages and updated naive datetime objects to become aware. The warnings that remain are from dependencies used within the project. Run `make pytest` from the root repository to view all the warnings.


pontoon/base/utils.py
```diff
@@ -10,7 +10,7 @@
 import zipfile
 
 from collections import defaultdict
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
```
mathjazz, line 13: We should be able to use `timezone` from `django.utils` everywhere I believe. So you could drop this import and also the `dj_timezone` alias below.

pontoon/contributors/tests/test_utils.py
```diff
@@ -27,7 +27,8 @@ def action_a(translation_a):
         action_type=ActionLog.ActionType.TRANSLATION_CREATED,
         translation=translation_a,
     )
-    action.created_at = datetime(2020, 1, 1)
+    aware_datetime = timezone.make_aware(datetime(2020, 1, 1))
+    action.created_at = aware_datetime
```
mathjazz, line 31: Nit: I think it would be clearer if you put everything in one line:
```python
action.created_at = timezone.make_aware(datetime(2020, 1, 1))
```

Also below.

pontoon/base/utils.py
mathjazz: Oh, I see now that `django.utils.timezone.utc` is deprecated.

mathjazz (commented):
Excellent work! I see that the internal `RemovedInDjango51Warning: 'index_together'` warnings are now gone!

I only left one suggestion to avoid the import alias, I think the code will be clearer that way. Please fix that and we can merge!

flodolo:
> Excellent job, Harmit!
> 
> Left some comments and recommendations for possible improvements.
> 
> I'm a bit puzzled with some of the remaining warnings. For example, I'm not sure what's the root cause of this:
> 
> ```
> pontoon/actionlog/tests/test_models.py: 217 warnings
>   /opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/django/db/models/options.py:210: RemovedInDjango51Warning: 'index_together' is deprecated. Use 'Meta.indexes' in 'base.Translation' instead.
>     warnings.warn(
> 
> pontoon/actionlog/tests/test_models.py: 102 warnings
>   /opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/django/db/models/options.py:210: RemovedInDjango51Warning: 'index_together' is deprecated. Use 'Meta.indexes' in 'base.Entity' instead.
>     warnings.warn(
> 
> pontoon/administration/tests/test_views.py::test_manage_project_strings
> pontoon/administration/tests/test_views.py::test_manage_project_strings
>   /opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/site-packages/notifications/templatetags/notifications_tags.py:35: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
>     if StrictVersion(get_version()) >= StrictVersion('2.0'):
> ```

Dumb questions:
* Can the first 2 be caused by older migrations still using `index_together`?
* The latter is caused by the notifications package? (checked their repo, v2.0 should fix most of them, but seems to be moving quite slowly).

mathjazz:
The first two are gone:
https://github.com/mozilla/pontoon/pull/3206#pullrequestreview-2048221870

And yes, the last one is external.

flodolo:
> The first two are gone:

I still see them, e.g. in the run on the last commit (but also locally)
https://github.com/mozilla/pontoon/actions/runs/9029543615/job/24812073586

Issue #2840, opened 2023-05-08
Add space after mentions


mathjazz:
After inserting a mentioned user in the comment field, the cursor is positioned right after the user's name. 

We should also add space after mentions.

eemeli:
The code attempts to do this already, but presumably this bit is somehow buggy? https://github.com/mozilla/pontoon/blob/570121addb91f7c618d247f27f3b844611a0fe69/translate/src/modules/comments/components/AddComment.tsx#L96-L98

PR #3210, opened 2024-05-10
Fixed commenting bug for right-to-left languages

### LINKED ISSUE #2400: Comment layout looks broken when mentioning Arabic nickname
Check the comment on one of the translations
https://pontoon.mozilla.org/ar/firefox/browser/browser/aboutPocket.ftl/?search=pocket-panel-signup-signup-firefox&string=224278

Schermata 2022-01-07 alle 16 37 42

The mention is at the beginning, but it's displayed at the rightmost end. Also, typing after that is really weird (try typing @, then select Safa's nickname, then type something).

harmitgoswami:
Fix #2400

Added the CSS changes proposed by @safaalfulaij (thank you!) to comments in Pontoon. In particular, directionality for comments is set to "plaintext" to treat it as simple text. Directionality for mentions is separated from the rest of the comment content, ensuring that Arabic and Hebrew nicknames maintain their directionality. To improve comment readability, the proposed change to move all comment content to a newline after the author's name was added.

**To reproduce results**: Populate your database as in [step 3 here](https://mozilla-pontoon.readthedocs.io/en/latest/dev/first-contribution.html). Create a superuser account, and set the nickname to be in either Arabic or Hebrew. Create a mock translation for one of the strings in a locale. Create another superuser account, and navigate back to the mock translation. Mention the first superuser using @[Arabic/Hebrew nickname], and post the comment. 

See attached photo below for reference.

image

### DIFF
--- a/translate/src/modules/comments/components/Comment.css
+++ b/translate/src/modules/comments/components/Comment.css
@@ -15,6 +15,15 @@
   background-color: var(--dark-grey-2);
   padding: 0 2px;
   border-radius: 2px;
+  unicode-bidi: isolate;
+}
+
+.comments-list .comment .container .content p {
+  unicode-bidi: plaintext;
+}
+
+.comments-list .comment .container .content .Linkify {
+  display: block;
 }
 
 .comments-list .comment-author {


mathjazz (approved):
Beautiful work, @harmitgoswami!

Also, thanks for the detailed description of the solution and steps to reproduce.

Just realized in the screenshots that we should also fix #2840.

PR #3212, opened 2024-05-10
Add ability to opt in to Email communication

### LINKED ISSUE #3211: Add Email communication settings
Spec:
https://github.com/mozilla/pontoon/blob/main/specs/0117-email-opt-in-unsubscribe.md#email-communication-user-settings

mathjazz:
Fix #3211. 

Adds _Email_ section to the user settings, which includes the existing _Contact email address_ field and the newly added _Email communication_ toggle. To match the existing UI patterns, I've used the shorter Checkbox label as required by the [spec](https://github.com/mozilla/pontoon/blob/main/specs/0117-email-opt-in-unsubscribe.md#email-communication-user-settings) and joined the checkbox label and subtitle from the spec into the help text, which is configurable by instance.

https://mozilla-pontoon-staging.herokuapp.com/settings/

We'll also need to update the docs:
https://mozilla-l10n.github.io/localizer-documentation/tools/pontoon/users.html


harmitgoswami (commented):
LGTM! However, when I try to access the stage, I'm denied permission. Is this supposed to happen?

mathjazz:
Can you try to Sign in?

harmitgoswami:
Okay it works now! All good on my end

harmitgoswami (approved):

mathjazz:
> To match the existing UI patterns, I've used the shorter Checkbox label as required by the [spec](https://github.com/mozilla/pontoon/blob/main/specs/0117-email-opt-in-unsubscribe.md#email-communication-user-settings) and joined the checkbox label and subtitle from the spec into the help text, which is configurable by instance.

@bcolsson As the author of the [spec](https://github.com/mozilla/pontoon/blob/main/specs/0117-email-opt-in-unsubscribe.md#email-communication-user-settings), do you approve these changes? Please check out the implementation on [stage](https://mozilla-pontoon-staging.herokuapp.com/settings/)!

pontoon/contributors/templates/contributors/settings.html
```diff
@@ -39,32 +39,39 @@ <h3>Personal information</h3>
...
         </div>
-        <div class="field">
-            {{ user_profile_form.bio.label_tag(label_suffix='') }}
-            {{ user_profile_form.bio }}
-            {{ user_profile_form.bio.errors }}
-            <p class="help">Displayed on the <a href="{{ url('pontoon.contributors.profile') }}">Profile page</a></p>
-        </div>
-      </section>
+        <ul class="check-list">
+            {{ Checkbox.checkbox('Email communication', class='field email-communications-enabled', attribute='email_communications_enabled', is_enabled=user.profile.email_communications_enabled, title='Receive email updates', help=settings.EMAIL_COMMUNICATION_HELP_TEXT) }}
```
flodolo, line 63: This should still be plural (communications)?

flodolo (commented):
According to the specs, `Contact email address` -> `Contact email address (optional)`

> Receive the latest updates about localization at Mozilla, announcements about new Pontoon features, receive invitations to contributor events and more.

Good call on removing the duplicate "receive" here, but

> Agree to Mozilla handling your personal information as explained in this Privacy Notice.

Missing a `I`.

mathjazz:
> According to the specs, `Contact email address` -> `Contact email address (optional)`

Are you suggesting we append _(optional)_ to the label?

All fields in the settings are optional. So I don't think it makes sense.

> > Agree to Mozilla handling your personal information as explained in this Privacy Notice.
> 
> Missing a `I`.

The first sentence is in the Imperative(?) mood. The second should be in the same. Hence - "Agree to ... your", rather than "I agree to ... my".

flodolo:
> Are you suggesting we append _(optional)_ to the label?

That's how I interpret the specs, but let's wait for Bryan.

> The first sentence is in the Imperative(?) mood. The second should be in the same. Hence - "Agree to ... your", rather than "I agree to ... my".

Good point, but it still sounds weird to me. Maybe add "By selecting/checking this,"?

> Receive the latest updates about localization at Mozilla, announcements about new Pontoon features, invitations to contributor events and more. By selecting this, you agree to Mozilla handling your personal information as explained in this [Privacy Notice](https://www.mozilla.org/privacy/websites/).

mathjazz:
Updated the text. I used the same copy as in the standalone page:
https://github.com/mozilla/pontoon/blob/main/specs/0117-email-opt-in-unsubscribe.md#standalone-page-for-email-opt-in-consent-shown-after-existing-user-login

Accordingly, I also replaced "Receive" with "Get".

bcolsson:
> > According to the specs, `Contact email address` -> `Contact email address (optional)`
> 
> Are you suggesting we append _(optional)_ to the label?
> 
> All fields in the settings are optional. So I don't think it makes sense.
> 
> > > Agree to Mozilla handling your personal information as explained in this Privacy Notice.
> > 
> > 
> > Missing a `I`.
> 
> The first sentence is in the Imperative(?) mood. The second should be in the same. Hence - "Agree to ... your", rather than "I agree to ... my".

These were both requests from legal. 

Let's leave "I agree.." as that's the wording the specifically requested.

As for "Optional", legal was concerned that users would be confused in the instance where they want to have their marketing and login email be the same. In practice they just need to enable the check box to for the consent, but they requested to add something to clarify that specifying an email address in the `Contact email address` field is optional. I added it to the label, but we could just as easily add it to the text underneath.

bcolsson:
> > To match the existing UI patterns, I've used the shorter Checkbox label as required by the [spec](https://github.com/mozilla/pontoon/blob/main/specs/0117-email-opt-in-unsubscribe.md#email-communication-user-settings) and joined the checkbox label and subtitle from the spec into the help text, which is configurable by instance.
> 
> @bcolsson As the author of the [spec](https://github.com/mozilla/pontoon/blob/main/specs/0117-email-opt-in-unsubscribe.md#email-communication-user-settings), do you approve these changes? Please check out the implementation on [stage](https://mozilla-pontoon-staging.herokuapp.com/settings/)!

Is there a way to add a line break to separate out the privacy notice to make it more distinct?

mathjazz:
> As for "Optional", legal was concerned that users would be confused in the instance where they want to have their marketing and login email be the same. In practice they just need to enable the check box to for the consent, but they requested to add something to clarify that specifying an email address in the `Contact email address` field is optional. I added it to the label, but we could just as easily add it to the text underneath.

OK, I see the confusion. How about we put the login email as the placeholder, so it would appear if contact email address is not set?

How should we change the text underneath, which already implies the email address is optional ("If you would like to...")? Replace "please do so here" with "you can optionally do so here"?

> @mathjazz Is there a way to add a line break to separate out the privacy notice to make it more distinct?

Done. Please check on stage if that's better. I think it _looks_ and _reads_ even better with a double line break (you can use devtools to test that), but then it's not immediately clear the text is related to the checkbox.

bcolsson:
> OK, I see the confusion. How about we put the login email as the placeholder, so it would appear if contact email address is not set?

I think that's a great solution.

> Done. Please check on stage if that's better. I think it looks and reads even better with a double line break (you can use devtools to test that), but then it's not immediately clear the text is related to the checkbox.

Agreed - I like the double line break. Also, I think the double line break is still understandable it relates to enabling email communications / the related checkbox due to indenting and the language "By enabling...".

mathjazz:
Stage updated with the placeholder and the additional line break:
https://mozilla-pontoon-staging.herokuapp.com/settings/

bcolsson:
Staging looking good to me 👍

docker-compose.yml
```diff
@@ -23,7 +23,7 @@ services:
 
   # Database
   postgresql:
-    image: postgres:15
+    image: postgres:13.4
```
flodolo, line 26: Why this change? 🤔

mathjazz: Nice catch. Fixed!

PR #3215, opened 2024-05-13
Added Tags search field to both Project and Localization dashboards

### LINKED ISSUE #3105: Add search field to Tags dashboards
We should add search field to the Tags tab in the Project / Localization dashboards.

That would make the UI consistent with the Team / Projects / Resources tabs and prevent the visual jump when switching between Tags.

harmitgoswami:
Fix #3105

Added a search field for Tags in both the Localization and Project dashboards. Reused same component structure and styling as the search field in the Resources/Teams tab located in the Localization/Project dashboards respectively.

**To reproduce results:** First, you will need to download a database dump that includes the "Firefox" and "Mozilla.org" projects. Load the database dump as described [here](https://mozilla-pontoon.readthedocs.io/en/latest/dev/contributing.html#database). Next, create a new superuser account using `make setup`. Run the webapp and locate either the "Firefox" or "Mozilla.org" projects in either the Teams>Language>Projects tab or directly through the Projects tab at the top of the Pontoon website header. Finally, click on the Tags tab to view the new search bar.

See images below for reference, and directly type in the URLs shown in the images if you have trouble finding page:

image

image

### DIFF
--- a/pontoon/localizations/templates/localizations/includes/tags.html
+++ b/pontoon/localizations/templates/localizations/includes/tags.html
@@ -1,6 +1,14 @@
 {% import 'tags/widgets/tag_list.html' as TagList %}
 
 <div class="tags">
+  <menu class="controls">
+    <div class="search-wrapper small clearfix">
+      <div class="icon fa fa-search"></div>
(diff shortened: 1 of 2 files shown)


mathjazz (approved):
Another excellent job, Harmit!

PR #3218, opened 2024-05-14
Drop bin/mozilla-en-US.py

eemeli:
As Firefox sources are now coming from [mozilla-l10n/firefox-l10n-source](https://github.com/mozilla-l10n/firefox-l10n-source), this script and the following repositories that it manages are no longer needed:
- https://hg.mozilla.org/users/m_owca.info/firefox-central
- https://hg.mozilla.org/users/m_owca.info/firefox-for-android-central
- https://hg.mozilla.org/users/m_owca.info/lightning-central
- https://hg.mozilla.org/users/m_owca.info/seamonkey-central
- https://hg.mozilla.org/users/m_owca.info/thunderbird-central

Before this PR is merged, the Heroku Scheduler job calling this script every 10 mins should be dropped.

The [`bin/make_comm_source_repos.py`](https://github.com/mozilla/pontoon/blob/main/bin/make_comm_source_repos.py) script and the repositories it maintains are still necessary at least until [bug 1877804](https://bugzilla.mozilla.org/show_bug.cgi?id=1877804) is resolved.


flodolo:
I was thinking that we should keep this until we're 100% sure, but in hindsight, I realized that the only relevant repo is the one for Firefox, and we can use `gecko-strings` directly if that scenario happens.

mathjazz (approved):

mathjazz:
I've dropped the script from the Scheduler in production.

PR #3233, opened 2024-05-17
Upgraded Chart.js to version 4.4.2

### LINKED ISSUE #2939: Upgrade Chart.js
We use an outdated Chart.js library. Let's upgrade it to latest version.

harmitgoswami:
Fix #2939 

Upgraded Chart.js dependency to latest version (4.4.2). Previously, Chart.js was included via a `Chart.bundle.js` static file. Support for this method has ended, so the dependency was included via a CDN. Changed all instances of deprecated functions to the new equivalents according to the [migration guide](https://www.chartjs.org/docs/latest/migration/v3-migration.html).

image


pontoon/base/templates/base.html
```diff
@@ -29,6 +29,10 @@
       {% endblock %}
     {% endblock %}
 
+    <!-- Add Chart.js and Moment CDN link -->
+    <script src="https://cdn.jsdelivr.net/npm/chart.js@4.4.2/dist/chart.umd.min.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/moment@^2.30.1"></script>
+    <script src="https://cdn.jsdelivr.net/npm/chartjs-adapter-moment@^1.0.1"></script>
```
mathjazz, line 35: We try to avoid loading static files from external sources. Please instead include the files in the repo.

pontoon/base/templates/base.html
```diff
@@ -29,6 +29,12 @@
       {% endblock %}
     {% endblock %}
 
+    <!-- Import Chart.js and Moment.js from static files -->
+    <script language="javascript" type="text/javascript" src="{{ static('js/lib/chart.umd.min.js') }}"></script>
+    <script language="javascript" type="text/javascript" src="{{ static('js/lib/moment.min.js') }}"></script>
+    <script language="javascript" type="text/javascript" src="{{ static('js/lib/chartjs-adapter-moment.min.js') }}"></script>
```
mathjazz, line 35: This will load all these libs on most of the views, even when they aren't needed.

We should load them, like other static files, through Django pipeline in settings, and only in the view where Chart.js was previously used.

Also, seems like pontoon/base/static/js/lib/chart.umd.js.map is not used?

harmitgoswami: You're right, I was getting a console error when I didn't include this file through base.html, but now that I've shifted over to local imports it seems this file isn't necessary.

pontoon/base/templates/base.html
```diff
@@ -28,7 +28,7 @@
       {% block extend_css %}
       {% endblock %}
     {% endblock %}
-
+
```
mathjazz, line 31: Please restore this change.

pontoon/contributors/static/js/profile.js
```diff
@@ -107,39 +107,35 @@ var Pontoon = (function (my) {
...
+              },
+              y: {
+                grid: {
+                  display: false,
+                },
+                ticks: {
+                  beginAtZero: true,
+                  maxTicksLimit: 3,
+                  precision: 2,
+                  callback: (value) => nf.format(value),
```
mathjazz, line 134: Why the change in the `precision` and the `callback` settings?

There are a few things that don't work as expected in the new Profile chart:
1. The curve doesn't have a background.
2. The Y scale data is off.
3. The tooltip: items overlap, the data format is off and there is no border.
4. The chart is cut-off at the top.
5. Can we keep the line curved?

before
after

mathjazz (commented):
The libraries are now loaded as expected. Nice work!

Please see my comment inline for the Profile chart.

I've noticed that the charts on http://localhost:8000/insights/ don't load.

Also, the Age of unreviewed suggestions and the Time to review pretranslations charts on http://localhost:8000/sl/insights/ shouldn't have legends.

mathjazz (commented):
I haven't looked at the code yet, but the changes look good. Nice work!

I noticed a few minor things, which should still be changed:
* On the X scales, there should be no day number (e.g. "Jun" instead of "Jun 1").
* We don't use a thousands separator on the Y scale ATM. I think it's a good idea to use it, but it should be `,` rather than `.` (same as in the tooltip).
* Borders in bullet points in the tooltips don't look the same as in production - they have a 3Dish look and the first one is bigger than the rest.
 
Screenshot 2024-05-23 at 16 07 13

harmitgoswami:
> I haven't looked at the code yet, but the changes look good. Nice work!
> 
> I noticed a few minor things, which should still be changed:
> 
>     * On the X scales, there should be no day number (e.g. "Jun" instead of "Jun 1").
> 
>     * We don't use a thousands separator on the Y scale ATM. I think it's a good idea to use it, but it should be `,` rather than `.` (same as in the tooltip).
> 
>     * Borders in bullet points in the tooltips don't look the same as in production - they have a 3Dish look and the first one is bigger than the rest.
> 
> 
> Screenshot 2024-05-23 at 16 07 13

Thank you! It seems that both the day number and thousands separator styling issues are dependent on browser/location configuration (i.e. the thousands separator is indeed `,` instead of `.` and day numbers already appear in production for Firefox v126.0 en-US).

mathjazz:
> Thank you! It seems that both the day number and thousands separator styling issues are dependent on browser/location configuration (i.e. the thousands separator is indeed `,` instead of `.` and day numbers already appear in production for Firefox v126.0 en-US).

True, but I believe the situation is worse than on prod ATM, where e.g. for me all charts show months as "May" rather than "May 1", which is the case in **some** charts now. E.g. the chart on the /profile page shows months rightfully as "May", whereas the Time to review suggestions chart on /sl/insights shows months as "May 1".

Also, it seems like charts on the /insights page are broken again.

pontoon/insights/static/js/insights_tab.js
```diff
@@ -104,58 +164,55 @@ var Pontoon = (function (my) {
...
-                  },
-                  ticks: {
-                    source: 'data',
+              x: {
+                type: 'time',
+                time: {
+                  displayFormats: {
+                    month: 'MMM',
                   },
+                  tooltipFormat: 'MMM YYYY',
```
mathjazz, line 195: This (and other instances of the same value) should remain `MMMM YYYY`, which will show the full month name in the tooltip, where we have enough space.

harmitgoswami:
The decision to switch from `Moment.js` over to `Date-fns.js` was mainly in hopes that the date display bug for the x-axis on charts will be resolved. The [documentation](https://momentjs.com/docs/#/-project-status/) for `Moment.js` states that no further features will be added, so it's possible that the visual consistency between that and `Chart.js` got lost at some point.

pontoon/insights/static/js/insights_charts.js
```diff
@@ -36,52 +36,15 @@ var Pontoon = (function (my) {
...
-          style.getPropertyValue('--light-grey-7');
-        Chart.defaults.global.defaultFontFamily = 'Open Sans';
-        Chart.defaults.global.defaultFontStyle = '100';
-        Chart.defaults.global.datasets.bar.barPercentage = 0.7;
-        Chart.defaults.global.datasets.bar.categoryPercentage = 0.7;
-      },
-      // Legend configuration doesn't allow for enough flexibility,
-      // so we build our own legend
-      // eslint-disable-next-line no-unused-vars
-      customLegend: (chart) => (chart) => {
```
mathjazz, line 49: Why don't we replace this with the contents of the generalized `htmlLegendPlugin`, which is currently defined in two different places, and thus more prone to errors than if it were kept in one place?

pontoon/insights/static/js/insights.js
```diff
@@ -56,90 +106,104 @@ var Pontoon = (function (my) {
...
+                borderWidth: 1,
+                caretPadding: 5,
+                padding: {
+                  x: 10,
+                  y: 10,
+                },
+                callbacks: {
+                  labelColor: function (context) {
+                    return {
+                      borderColor: '#fff',
```
mathjazz, line 182: Same as above - no color hardcoding. Please update all instances.

pontoon/insights/static/js/insights.js
```diff
@@ -56,90 +106,104 @@ var Pontoon = (function (my) {
             borderWidth: item.name === 'All' ? 3 : 1,
             pointBackgroundColor: color,
             pointHitRadius: 10,
-            pointRadius: 4,
+            pointRadius: 3.25,
```
mathjazz, line 109: Nit: Could you please change all values of pointRadius to `3.25` across all charts?

pontoon/insights/static/js/insights_tab.js
```diff
@@ -316,61 +389,73 @@ var Pontoon = (function (my) {
...
+                intersect: false,
+                borderColor: style.getPropertyValue('--dark-pink'),
+                borderWidth: 1,
+                caretPadding: 5,
+                padding: {
+                  x: 10,
+                  y: 10,
+                },
+                callbacks: {
+                  labelColor: function (context) {
```
mathjazz, line 410: This function is also repeated many times. I would factor it out in the `insights.js` file and then call it from all the places where it's used.

pontoon/insights/static/js/insights_tab.js
```diff
@@ -449,125 +536,136 @@ var Pontoon = (function (my) {
...
+                intersect: false,
+                position: 'nearest',
+                borderColor: style.getPropertyValue('--status-translated'),
+                borderWidth: 1,
+                caretPadding: 5,
+                padding: {
+                  x: 10,
+                  y: 10,
+                },
+                itemSort: function (a, b) {
```
mathjazz, line 618: Please restore the comment.

mathjazz (commented):
Very well done!
 
All charts work as expected.

Left some comments inline, mostly just to improve the code maintainability.

pontoon/contributors/static/js/profile.js
```diff
@@ -82,68 +84,79 @@ var Pontoon = (function (my) {
...
+                  callback: (value) => nf.format(value / 100),
+                },
+                beginAtZero: true,
+                max: 100,
+              },
             },
           },
         });
       },
+      setLabelValue: function (context) {
```
mathjazz, line 151: We can simplify this further I believe by only having one `setLabelValue()` located in `insights_charts.js`, similar to how `customLegend()` was used.

Also, is `setLabelColor()` a better name?

pontoon/insights/static/js/insights_charts.js
mathjazz: This one still needs to be addressed.

pontoon/contributors/static/js/profile.js
```diff
@@ -82,64 +84,68 @@ var Pontoon = (function (my) {
...
+                borderWidth: 1,
+                caretPadding: 5,
+                padding: {
+                  x: 10,
+                  y: 10,
+                },
+                callbacks: {
+                  labelColor: (context) => {
+                    return Pontoon.insights.setLabelColor(context);
+                  },
```
mathjazz, line 108: You can [drop](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#description) the body braces and the word "return":

```suggestion
                  labelColor: (context) =>
                    Pontoon.insights.setLabelColor(context),
                  },
```

The same applies for all other calls of this function.

mathjazz (commented):
Excellent work!

Just that minor Arrow function trick and we're good!

mathjazz (approved):
\o/

Very well done!

PR #3237, opened 2024-05-22
Implement email unsubscribe page

### LINKED ISSUE #3220: Implement email unsubscribe page
Spec:
https://github.com/mozilla/pontoon/blob/main/specs/0117-email-opt-in-unsubscribe.md#unsubscribe-page

mathjazz:
Fix #3220: Implement email unsubscribe page.

This patch adds a unique user identifier to existing and new users, and uses it to create personalized unsubscribe links.

The functionality follows the (edited: "As a user I can:" part of the) [spec](https://github.com/mozilla/pontoon/blob/main/specs/0117-email-opt-in-unsubscribe.md#unsubscribe-page), but the copy is much simpler and so is the UX. Clicking on the unsubscribe link instantly unsubscribes the user, which still has the ability to easily subscribe again. 

Most of the Unsubscribe links work that way and Eemeli and I both believe we are expected to be as good citizens as possible when it comes to respecting user choice. With this approach there's also no need for the deployment-specific copy.


pontoon/messaging/static/css/email_consent.css
```diff
@@ -65,10 +69,13 @@ body > header.menu-opened {
...
   font-weight: 300;
   line-height: 36px;
-  margin-bottom: 60px;
   width: 900px;
 }
 
+#main p.main-text {
+  margin-bottom: 60px;
+}
+
```
mathjazz, line 78: Changes to this file drop the scrollbar on the Email consent page, so they are indirectly related to the issue the patch is fixing.

pontoon/base/models/user_profile.py
```diff
@@ -20,6 +22,7 @@ class UserProfile(models.Model):
     contact_email_verified = models.BooleanField(default=False)
     email_communications_enabled = models.BooleanField(default=False)
     email_consent_dismissed_at = models.DateTimeField(null=True, blank=True)
+    unsubscribe_token = models.UUIDField(default=uuid.uuid4, editable=False)
```
eemeli, line 25: Given that this is a unique user id, why not call it such?

pontoon/messaging/templates/messaging/subscribe.html
eemeli, line 1: Is this actually used anywhere, or included just as a parallel for the unsubscribe? If the former, "Unsubscribe again" is weird. If the latter, YAGNI.

eemeli (changes requested):

pontoon/messaging/templates/messaging/subscribe.html
mathjazz: Yeah, this page loads if you click on the "Subscribe again" link on the Unsubscribe page.

eemeli: Ah, ok, that makes sense.

eemeli (approved):

mathjazz:
@bcolsson Could you please verify that this is OK? 

Please see my comment [above](https://github.com/mozilla/pontoon/pull/3237#issue-2309993123) why the UX is different compared to the spec.

This is your unsubscribe link on stage:
https://mozilla-pontoon-staging.herokuapp.com/unsubscribe/17e9d1fa-5593-4427-a0de-1bdc8c2679d0/

bcolsson:
Thanks @mathjazz - based on our conversation in slack I'll approve this, but this will require approval by legal before we merge.

bcolsson:
Hi @mathjazz - I got legal approval for the changes in this PR.

PR #3241, opened 2024-05-30
Implement UX Logging and Fix LLM Translation Shortcut Functionality

### LINKED ISSUE #3204: LLM translations are not correctly attributed
Click the AI button and select rephrase, then click to copy the text and submit the translation.

The UX action log is correctly recorded (first `LLM dropdown select`, then `LLM dropdown copied`).

The translation is incorrectly attributed to Google Translate in the UI, and this code doesn't return any match

```
from pontoon.base.models import Translation

# List Translation copied from the LLM suggestions
Translation.objects.filter(machinery_sources__contains=['gpt-transform']).values("locale__code", "date")

>>> <TranslationQuerySet []>
```

ayanaar:
Fully fixes #3204 

This PR includes the following changes:

- **Refactored the LLM hook**: The LLM hook (`useLLMTranslation`) was refactored to return a function that is called outside of the callback. This change was made to address issues with the `useHandleCtrlShiftArrow` function, which previously caused the keyboard shortcut to stop functioning correctly after an LLM translation was generated. 

- **Added UX logging**: Implemented UX logging for when LLM translations are copied into the editor using the keyboard shortcut. This helps track the adoption and usage of the LLM translation feature.

### Issue Resolved

- The keyboard shortcut now works as expected for LLM translations and other translations, including when an LLM translation is selected from the dropdown.

### Note

As discussed with @eemeli, there is a need for further refactoring to fully address this issue. Specifically, issue #3305 - "When the main edit box is not in focus, the global arrow key combinations (ArrowDown or ArrowUp) do not function as expected" - should be addressed in the future. While the current changes enable the feature to function as expected, a more comprehensive solution is still needed to address the underlying issue in the existing code.

### DIFF
--- a/translate/src/context/TranslationContext.tsx
+++ b/translate/src/context/TranslationContext.tsx
@@ -4,7 +4,7 @@ import { type MachineryTranslation, fetchGPTTransform } from '~/api/machinery';
 export type SelState = {
   loading: boolean;
   selectedOption: string;
-  llmTranslations: Record<string, string>;
+  llmTranslation: string;
 };
 
 interface LLMTranslationContextType {
@@ -20,7 +20,7 @@ interface LLMTranslationContextType {
 const initSelState = () => ({
   loading: false,
   selectedOption: '',
-  llmTranslations: {},
+  llmTranslation: '',
 });
 
 const LLMTranslationContext = createContext<LLMTranslationContextType>({
@@ -59,10 +59,7 @@ export const LLMTranslationProvider: React.FC = ({ children }) => {
       stateRef.current.set(mt, {
         loading: false,
         selectedOption: characteristic,
-        llmTranslations: {
-          ...currentState.llmTranslations,
-          [characteristic]: machineryTranslations[0].translation,
-        },
+        llmTranslation: machineryTranslations[0].translation,
       });
     } else {
       stateRef.current.set(mt, {
@@ -78,7 +75,7 @@ export const LLMTranslationProvider: React.FC = ({ children }) => {
     stateRef.current.set(mt, {
       ...currentState,
       selectedOption: '',
-      llmTranslations: {},
+      llmTranslation: '',
     });
     setVersion((v) => v + 1);
   };
--- a/translate/src/context/TranslationContext.tsx
+++ b/translate/src/context/TranslationContext.tsx
@@ -1,7 +1,7 @@
 import React, { createContext, useContext, useState, useRef } from 'react';
 import { type MachineryTranslation, fetchGPTTransform } from '~/api/machinery';
 
-export type SelState = {
+type SelState = {
   loading: boolean;
   selectedOption: string;
   llmTranslation: string;
--- a/translate/src/context/TranslationContext.tsx
+++ b/translate/src/context/TranslationContext.tsx
@@ -1,7 +1,7 @@
 import React, { createContext, useContext, useState, useRef } from 'react';
 import { type MachineryTranslation, fetchGPTTransform } from '~/api/machinery';
 
-type SelState = {
+export type SelState = {
   loading: boolean;
   selectedOption: string;
   llmTranslations: Record<string, string>;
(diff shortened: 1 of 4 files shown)


MpcOS77 (changes requested):
The

MpcOS77 (approved):

translate/src/context/TranslationContext.tsx
```diff
@@ -59,7 +59,10 @@ export const LLMTranslationProvider: React.FC = ({ children }) => {
       stateRef.current.set(mt, {
         loading: false,
         selectedOption: characteristic,
```
eemeli, line 61: It's a bit confusing that "selected option" and "characteristic" are names for essentially the same value, which is typed only as a `string`. These values should be typed more strictly, with something like
```ts
type Characteristic = 'rephrased' | 'formal' | 'informal'
```

Also, the variable names should be consistent, so `selectedCharacteristic` instead of `selectedOption`.

translate/src/context/TranslationContext.tsx
```diff
@@ -1,10 +1,10 @@
 import React, { createContext, useContext, useState, useRef } from 'react';
 import { type MachineryTranslation, fetchGPTTransform } from '~/api/machinery';
 
-type SelState = {
+export type SelState = {
   loading: boolean;
   selectedOption: string;
-  llmTranslation: string;
+  llmTranslations: Record<string, string>;
 };
```
eemeli, line 8: It's unclear why this refactor is included in a fix for keyboard shortcuts.

translate/src/modules/translationform/utils/editFieldShortcuts.ts
```diff
@@ -91,13 +94,19 @@ export function useHandleCtrlShiftArrow(): (
...
+  const isMachinery = tab === 0;
+  const numTranslations = isMachinery
+    ? machineryTranslations.length + concordanceSearchResults.length
+    : otherLocaleTranslations.length;
+
+  // Precompute LLM state at the top level
+  let llmState: SelState | null = null;
+  if (isMachinery && element < machineryTranslations.length) {
+    llmState = useLLMTranslation(machineryTranslations[element]);
+  }
```
eemeli, line 107: Why is this now outside the callback, when previously it was in it?

translate/src/modules/translationform/utils/editFieldShortcuts.ts
```diff
@@ -109,13 +118,31 @@ export function useHandleCtrlShiftArrow(): (
...
       const len = machineryTranslations.length;
-      const { translation, sources } =
+      const translationObj =
         nextIdx < len
           ? machineryTranslations[nextIdx]
           : concordanceSearchResults[nextIdx - len];
-      setEditorFromHelpers(translation, sources, true);
+
+      // If we have a valid LLM state, process it
+      if (llmState) {
```
eemeli, line 127: And what if we don't have a valid LLM state?

eemeli (commented):
This seems like a combination of 2-3 different sets of changes that would probably be clearer in separate PRs, but in general the direction for each of them seems fine.

translate/src/modules/translationform/utils/editFieldShortcuts.ts
ayanaar: I had to move the code outside the callback to resolve the error: "Invalid hook call. Hooks can only be called inside the body of a function component." When it is inside the callback the `useLLMTranslation` hook is being called inside a callback function returned by the `useHandleCtrlShiftArrow` custom hook. However, this change of moving it outside, led to issues with the indexing not working correctly (off by 1). I’m currently looking for a way to address both the hook placement and the indexing functionality. Any suggestions on how to mitigate this would be appreciated.

eemeli: Ah, I didn't notice the hook call. Yeah, that definitely needs to be outside the callback, and it depends on the preceding lines.

eemeli (approved):

PR #3248, opened 2024-05-30
Replace black, flake8, pyupgrade with ruff for Python linting and formatting

flodolo:


flodolo:
Thoughts on adding `ruff` on top of `black`? Does it need to be called out in [contributing.rst](https://github.com/mozilla/pontoon/blob/main/CONTRIBUTING.rst#python-code-conventions)?

flodolo:
Looking at AMO, they completely dropped black for ruff, and use some configuration options
https://github.com/eviljeff/olympia/blob/9c570b5180b02a6e5d9e0d1d1dc550d76b3c249e/pyproject.toml

I guess it should be replacing both flake8 and black at this point 🤔 
https://docs.astral.sh/ruff/

mathjazz:
Seems like `ruff` would let us drop `black`, `flake8` and even [`pyupgrade`](https://github.com/astral-sh/ruff/issues/827).

All that while being faster.

Sounds like a win-win to me.

flodolo:
Oh, I missed pyupgrade.

Let me play with it a bit more, especially with the configuration part, and see how far I can go.

flodolo:
There are quite a few changes running `ruff format` to replace black:
* Space after the class definition: https://github.com/astral-sh/ruff/issues/8566
* Some are due to the fact that we were using black 22. The latest version already reformatted 17 files (e.g. adding an empty line after the docstring).

Not completely sure what we want to add/enable in the TOML file.

mathjazz:
>     * Space after the class definition: [formatter removes blank line after class definition astral-sh/ruff#8566](https://github.com/astral-sh/ruff/issues/8566)

Disble that?

https://docs.astral.sh/ruff/rules/one-blank-line-after-class/

flodolo:
> > Space after the class definition: [formatter removes blank line after class definition astral-sh/ruff#8566](https://github.com/astral-sh/ruff/issues/8566)
> 
> Disable that?
> 
> https://docs.astral.sh/ruff/rules/one-blank-line-after-class/

I don't see a way to disable it in the format part (while it would be possible in the linting), very much like black wasn't configurable.

flodolo:
See also https://github.com/astral-sh/ruff/issues/9745

mathjazz (approved):
> I don't see a way to disable it in the format part (while it would be possible in the linting), very much like black wasn't configurable.

Oh, I see...

We talked about this PR at the call today and we all agreed we should go ahead.

mathjazz:
Also, nobody expressed strong opinions the new line after block open and/or about excluding migration files, so let's leave that as it is for now.

Eemeli suggested we also look into https://pycqa.github.io/isort/.

flodolo:
> Eemeli suggested we also look into https://pycqa.github.io/isort/.

In case, looks like it can be fine-tuned for Django (see [AMO file](https://github.com/eviljeff/olympia/blob/9c570b5180b02a6e5d9e0d1d1dc550d76b3c249e/pyproject.toml#L28-L33))

PR #3251, opened 2024-06-04
Use isort to format Python imports

flodolo:


flodolo:
This looks like the closest formatting to what the code is currently doing.

Dropping `no-sections` will create groups (Django, local imports, etc.), the orderd can be decided.

```
section-order = ["future", "standard-library", "django", "third-party", "first-party", "local-folder"]

[lint.isort.sections]
"django" = ["django"]
```

Setting `lines-between-types = 0` will remove the empty line between import and from.

mathjazz:
Looks good! Leaving the final decision to @eemeli.

ruff.toml
```diff
@@ -11,9 +11,16 @@ ignore = [
...
     "F",  # pyflakes
+    "I",  # isort
     "W",  # pycodestyle warnings
 ]
 
+[lint.isort]
+combine-as-imports = true
+lines-after-imports = 2
+lines-between-types = 1
+no-sections = true
```
eemeli, line 22: I would much prefer not setting `no-sections`, as it doesn't differentiate between local and non-local imports. I don't have a specific section order I prefer (as in, the default would be fine for me), but having _some_ section order would be good.
```suggestion
```

eemeli (changes requested):

ruff.toml
flodolo: > I would much prefer not setting no-sections, as it doesn't differentiate between local and non-local imports.

I'm confused. Removing `no-sections` means that it **will** differentiate (and separate) different type of imports.

For example, `models/projects.py` becomes

```python
from os.path import basename, join, normpath
from urllib.parse import urlparse

from django.conf import settings
from django.db import models
from django.db.models import Prefetch
from django.utils.functional import cached_property

from pontoon.base import utils
from pontoon.base.models.aggregated_stats import AggregatedStats
from pontoon.base.models.changed_entity_locale import ChangedEntityLocale
from pontoon.base.models.locale import Locale

class…
```

Is that what you want?

flodolo: OK. I'm going to use the section order that I found in docs, let me know if you prefer to drop that.

```
section-order = ["future", "standard-library", "django", "third-party", "first-party", "local-folder"]

[lint.isort.sections]
"django" = ["django"]
```

eemeli (approved):
> Setting `lines-between-types = 0` will remove the empty line between import and from.

Looking at the updated diff, I'm finding the extra lines separating `import` and `from` lines distracting, as we're now assigning two different meanings to empty lines between imports.

So I'd have a preference to do as you suggest, and set `lines-between-types = 0`. But my feelings aren't too strong here, so I'm going to approve this and pass it back to @mathjazz for his feedback. 😇

ruff.toml
eemeli: No, that's good.

mathjazz (approved):
Looks good.

The only thing I'd do differently is to move the `"django"` section after the `"third-party"` section.

PR #3252, opened 2024-06-05
Fixed spacing bug after mentions in comments

### LINKED ISSUE #2840: Add space after mentions

harmitgoswami:
Fix #2840 

Fixed styling bug for comments, in which there was no automatic space appended after mentioning a user using `@[name]`. The bug was resolved by explicitly setting the cursor to the end of the editor's content, thereby ensuring that the space was added in the correct location within the editor.

To reproduce results: After starting up the server, navigate to any translated string within any team/project. Then, click the "COMMENT" button on that string, and type "@[any letter]" to bring up a list of users to mention in your comment. Choose any user using either your cursor or the arrow keys + tab, then type the rest of your comment. There should be a space automatically added after the mention, before the content of your comment.

image

### DIFF
--- a/translate/src/modules/comments/components/AddComment.tsx
+++ b/translate/src/modules/comments/components/AddComment.tsx
@@ -108,7 +108,7 @@ export function AddComment({
         children: [{ text: name }],
       };
       Transforms.insertNodes(editor, mention);
-      Transforms.move(editor);
+      Transforms.select(editor, Editor.end(editor, []));
       Transforms.insertText(editor, ' ');
     },
     [],
--- a/translate/src/modules/comments/components/AddComment.tsx
+++ b/translate/src/modules/comments/components/AddComment.tsx
@@ -111,7 +111,7 @@ export function AddComment({
       Transforms.select(editor, Editor.end(editor, []));
       Transforms.insertText(editor, ' ');
     },
-    [],
+    [editor],
   );
 
   useEffect(initMentions, []);


mathjazz:
I wonder if we can avoid the edge case - if you add a mention in the middle of the comment.

harmitgoswami:
> I wonder if we can avoid the edge case - if you add a mention in the middle of the comment.

Seems like simply adding editor to the dependency array of the useCallback did the trick :D

image

translate/src/modules/comments/components/AddComment.tsx
```diff
@@ -108,10 +108,10 @@ export function AddComment({
         children: [{ text: name }],
       };
       Transforms.insertNodes(editor, mention);
-      Transforms.move(editor);
+      Transforms.select(editor, Editor.end(editor, []));
       Transforms.insertText(editor, ' ');
     },
-    [],
+    [editor],
```
eemeli, line 114: This addition is fine, but unnecessary in this case as the value is coming from `useMemo`.

eemeli (approved):

translate/src/modules/comments/components/AddComment.tsx
mathjazz: So let's remove that? 

It doesn't seem to fix the problem where you add a mention in the middle of the comment - when you insert it, the space is added to the end of the string and the cursor jumps there, too.

eemeli: Including the `editor` as a dependency is nominally good practice; skipping it is such a marginal optimization that it doesn't matter. Either way goes, and neither is wrong.

PR #3253, opened 2024-06-06
Added capability to mention users by username

### LINKED ISSUE #2625: Add ability to mention users using their username
As part of the redesigned profile page, we've introduced usernames.

We should make use of them in user mentions in comments.

harmitgoswami:
Fix #2625 

Added functionality for mentioning users by typing their username instead of only by display name. This feature was achieved by modifying the `/get-users/` API to return the usernames of all UserProfile objects in addition to the preexisting fields. The suggestedUsers filter function was modified to first check for a matching username (if it exists), then checking for a matching display name.

**To reproduce results**: After starting up the server, navigate to any team/project. Select any string that has an existing translation. Click the "COMMENT" button, and then inside the text editor field, type `@[any string of letters]` to display a list of users whose display name or username contains that string of letters. If you know a specific user's username, try typing their full username or a subset of it, and their profile will appear just as if you typed their display name instead. 

image

image

### DIFF
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -676,6 +676,8 @@ def get_users(request):
         .exclude(profile__system_user=True)
         # Exclude deleted users
         .exclude(email__regex=r"^deleted-user-(\w+)@example.com$")
+        # Prefetch profile for retrieving username
+        .prefetch_related("profile")
     )
     payload = []
 
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -685,6 +685,7 @@ def get_users(request):
                 "gravatar": u.gravatar_url(44),
                 "name": u.name_or_email,
                 "url": u.profile_url,
+                "username": u.profile.username,
             }
         )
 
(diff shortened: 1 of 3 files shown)


pontoon/base/views.py
```diff
@@ -685,6 +685,7 @@ def get_users(request):
                 "gravatar": u.gravatar_url(44),
                 "name": u.name_or_email,
                 "url": u.profile_url,
+                "username": u.profile.username,
```
mathjazz, line 688: This will result in N additional DB queries, where N is the number of `users` (currently close to 20k on production), which makes the `/get-users/` request quite slow compared to its current performance.

This happens because the username is stored in a separate table (`UserProfile` rather than `User`). The issue can be mitigated by using [`prefetch_related()`](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#prefetch-related).

You can debug DB queries using [`debug_sql()`](https://github.com/mozilla/pontoon/blob/main/pontoon/dev/utils.py#L5).

harmitgoswami: That's a million times faster! Thank you for pointing that out

mathjazz (approved):
Excellent job!

PR #3255, opened 2024-06-13
Add button to show older activity in activity log

### LINKED ISSUE #2622: Add a "Show more activity" button in the activity log
Currently it's not possible to show more data, for example when an entire year is selected.

For comparison, GH has such a button.

harmitgoswami:
Fix #2622 

Implemented a button to show more (i.e. older) activity in the user profile activity log. This feature was achieved by modifying the existing `get_contribution_timeline_data` function to retrieve data from the past N months, where N is a counter that the user can increase by one via clicking on the "Show more activity" button. Once the number of months shown on the activity log predates the user's account creation date, the "Show more activity" button is hidden, similar to how GitHub implements this feature.

**To reproduce results**: After starting the server, navigate to any user's profile `(localhost:8000/contributors/[username])`. Once there, click the show more activity button to display an additional month's worth of data. When choosing to view a specific contribution type, the counter resets to 0, and the "Show more activity" button is displayed again (if it was hidden before). 

image
image


pontoon/contributors/static/js/profile.js
```diff
@@ -296,8 +296,12 @@ var Pontoon = (function (my) {
             .find('.selector .value')
             .html($(this).html());
 
+          // Rerender the show more button in case it was hidden
+          $('#show-more').show();
+
           const type = $('#contributions .type-selector span').data('type');
           const user = $('#server').data('user');
+          let months_shown = $(this).data('months_shown');
```
mathjazz, line 304: Nit: Note that we use camelCase in JS, kebab-case in HTML and snake_case in python. 🥳

pontoon/contributors/static/css/profile.css
```diff
@@ -119,6 +119,25 @@ h2 {
...
+  font-weight: normal;
+  height: 32px;
+  padding: 8px 12px;
+  text-align: center;
+  text-transform: uppercase;
+  width: 100%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  gap: 8px;
```
mathjazz, line 134: Are these necessary?

pontoon/contributors/static/css/profile.css
```diff
@@ -119,6 +119,25 @@ h2 {
   width: 100%;
 }
 
+.right-column .button {
+  background: var(--dark-grey-1);
+  font-size: 16px;
+  font-weight: normal;
+  height: 32px;
+  padding: 8px 12px;
+  text-align: center;
```
mathjazz, line 128: Are these necessary?

pontoon/contributors/static/css/profile.css
```diff
@@ -119,6 +119,25 @@ h2 {
   width: 100%;
 }
 
+.right-column .button {
```
mathjazz, line 122: We should make the selector stricter, otherwise it also impacts the contribution type selector button.

flodolo:
I think it's a bit confusing.

https://mozilla-pontoon-staging.herokuapp.com/contributors/flod/

1. I click the button, it shows more activity (`Contribution activity in the last 2 months`), so that works.
2. I click another time, nothing happens, because there is no activity between month 2 and 3 (only the title changes).
3. I click a third time and the content changes because there is activity between month 3 and 4.

When I click the button, I would expect more activity to be loaded if available (e.g. load X more events, not expand the time filter).

It's even more confusing when I select a day with activity, e.g. in February, and then click the button.

mathjazz:
> When I click the button, I would expect more activity to be loaded if available (e.g. load X more events, not expand the time filter).
> 
> It's even more confusing when I select a day with activity, e.g. in February, and then click the button.

@harmitgoswami Can we change the logic to load the data from the beginning of the preceding month? So if the currently displayed data ends in June 10, we load data from May 1 to June 10 and append it. We'd also need to add month indicators to the timeline.

harmitgoswami:
That makes sense to me. So then do we still want the button to add another month to the timeline per click, or like Flod suggested, retrieve latest X events?

mathjazz:
> That makes sense to me. So then do we still want the button to add another month to the timeline per click, or like Flod suggested, retrieve latest X events?

What flod suggests would be even better - so you don't click in vain.

mathjazz:
Thanks for the update!

When I test the patch locally, my [profile page](http://localhost:8000/profile/) says "Contribution activity since May 2024: ". It's not clear why since _May_ and if that's since the beginning or since the end of May. My [original suggestion](https://github.com/mozilla/pontoon/pull/3255#issuecomment-2167696380) was just for the data that we load additionally with the button. Sorry for not making that clearer.

Ideally, I'd refactor `get_contribution_timeline_data()` to work the way it works currently on page load, and to append additional data on each click on the "Show more activity" button.

mathjazz:
> When I click the button, I would expect more activity to be loaded if available (e.g. load X more events, not expand the time filter).

@flodolo It turns out it'll be much easier for us to change the functionality to load events for the current month initially, and then append events for each preceding month after clicking the Show more activity button. That's also the behavior on the GH profile page. For less active users, that could mean no new events will be loaded after clicking the button, which we could address subsequently by automatically extending the period on the backend if no events are returned. Thoughts?

flodolo:
By default, we're showing the graph for the last 12 months, the activity below should match that interval of time.
* We load the activity for the last month by default.
* Clicking `Show more activity` should load some data (if the previous month is empty, we should load stuff until we have something). I don't have a strong preference between "load 1 more month" vs "load X more events".
* We stop when we reach the end of the interval (while I have the feeling now we're not stopping at all, I got to `Contribution activity in the last 13 months` on stage).

mathjazz:
@flodolo How about the button then just loads all the remaining data rendered on the graph, i.e. for the last 365 days?

We can also group events by month in the timeline.

flodolo:
> @flodolo How about the button then just loads all the remaining events rendered on the graph, i.e. for the last 365 days?
> 
> We can also group events by month in the timeline.

This works for me. And grouping would help visually.

harmitgoswami:
Not sure what the best way is to display this information... any thoughts?

flodolo:
> Not sure what the best way is to display this information... any thoughts?

Can you expand on "this information"?

harmitgoswami:
Sorry, forgot to add the images!

This order of information makes the most sense to me (`Title -> Project/Locale -> Month -> Action`), but I was wondering what you two thought

Initial contents:
image

After clicking "Show more activity":
image

flodolo:
I think we need @mathjazz here, to avoid going too far in the wrong direction, but this looks like too much data. 

My expectation is that we keep the current structure, but group it by month (i.e. take the current look of activities, repeat it for each month if there is data).

mathjazz:
Agreed with Flod. 

And we should follow the GitHub profile page UI - add "Month YEAR" (e.g. May 2024) to the vertical timeline.

harmitgoswami:
In this design iteration, it's a little difficult to discern which link you're clicking on when hovering over an action. I've reserved fixing this bug for now in case this design isn't what we were looking for either.

image

mathjazz:
This is the design we're aiming for:

Posnetek zaslona 2024-06-24 ob 23 59 52

eemeli:
@harmitgoswami This is currently marked as a draft PR; could you clarify what sort of review you'd like of it, and/or what's still intended to be done to it?

harmitgoswami:
> @harmitgoswami This is currently marked as a draft PR; could you clarify what sort of review you'd like of it, and/or what's still intended to be done to it?

Hi Eemeli, I’m mainly looking for code style review for this PR. 

I’ve implemented the design and functionality that @mathjazz had requested, but am looking for code style/performance improvements before I mark this PR as Open once again.

pontoon/contributors/utils.py
```diff
@@ -427,24 +427,89 @@ def get_project_locale_contribution_counts(contributions_qs):
...
     """
     end = timezone.now()
-    start = end - relativedelta(months=1)
-    timeline_title = "Contribution activity in the last month"
+
+    if not year_shown:
+        start = end - relativedelta(day=1)
+        timeline_title = "Contribution activity in this month"
+    else:
+        timeline_title = "Contribution activity in the past year"
```
eemeli, line 442: If there's no clear reason why to test `if not year_shown`, the test should be `if year_shown`.

pontoon/contributors/utils.py
```diff
@@ -427,24 +427,89 @@ def get_project_locale_contribution_counts(contributions_qs):
     return counts
 
 
-def get_contribution_timeline_data(user, contribution_type=None, day=None):
+def get_contribution_timeline_data(
+    user, year_shown=None, contribution_type=None, day=None
```
eemeli, line 431: This applies to all the places where this variable name is used: Why "year shown"? Why not "show year"? And if this is a boolean, why does it default to `None`?

pontoon/contributors/utils.py
```diff
@@ -427,24 +427,89 @@ def get_project_locale_contribution_counts(contributions_qs):
...
+    if year_shown:
+        # Perform query 12 times and append to contributions
+        for i in range(12):
+            month_start = end - relativedelta(months=i, day=1)
+            month_end = end - relativedelta(months=i, day=31)
+            contribution_period = Q(
+                created_at__gte=month_start, created_at__lte=month_end
+            )
+            month_start_name = month_start.strftime("%B %Y")
+            contributions_map = get_contributions_map(user, contribution_period)
```
eemeli, line 461: Let's not. `get_contributions_map` is already doing five separate ActionLog queries, so with this loop we get to 60 queries. All of which is then used only for numerical summaries.

All of this should really be one query, and we make the database do all the math. I could write the SQL for that, but making Django generate the same could be trickier. The next best alternative would be to query once for the actions, and then do the filtering in Python.

pontoon/contributors/views.py
```diff
@@ -115,6 +116,8 @@ def update_contribution_timeline(request):
     try:
         user = User.objects.get(pk=request.GET["user"])
         contribution_type = request.GET["contribution_type"]
+        year_shown = request.GET.get("year_shown", None)
+        year_shown = True if year_shown == "true" else False
```
eemeli, line 120:
```suggestion
        year_shown = request.GET.get("year_shown", None) == "true"
```

pontoon/contributors/utils.py
```diff
@@ -427,24 +427,89 @@ def get_project_locale_contribution_counts(contributions_qs):
...
+                else:
+                    restructured[month_name][title]["data"][key]["actions"].extend(
+                        value["actions"]
+                    )
+                    restructured[month_name][title]["data"][key]["count"] += value[
+                        "count"
+                    ]
+        return (restructured, timeline_title, year_shown)
+
+    return (contributions, timeline_title, year_shown)
```
eemeli, line 507: Does `year_shown` ever change here? If not, why is it included in the output?

pontoon/contributors/templates/contributors/includes/timeline.html
```diff
@@ -1,24 +1,51 @@
...
-                localization.locale.name }} <span class="stress">{{ localization.locale.code }}</span>
-                <span class="contribution-count">{{ ", ".join(localization.actions) }}</span>
-            </a>
-        </li>
+{% if contribution_timeline.year_shown %}
+    {% for month, month_data in contribution_timeline.contributions.items() %}
+    <div class="contribution-year">
+        <span class="month-label">{{ month }}</span>
+        {% for title, project_data in month_data.items() %}
+        <div class="contribution-group">
```
eemeli, line 8: It looks like this element is repeated exactly below? Could it be defined once and used twice? DRY isn't an absolute rule, but for something this complex it should be followed unless there's a good reason not to.

eemeli (changes requested):
A bunch of small stuff, but the big issue is the inefficiency in how the data is queried from the DB.

pontoon/contributors/utils.py
harmitgoswami: Yes, `year_shown` on line 505 is always True but `year_shown` on line 507 is always False. I included it in the output since the `timeline.html` template uses `year_shown` as an indicator on what object structure it should expect to render

eemeli: I meant that it doesn't change within the `get_contribution_timeline_data()` function. The code calling it already has access to the `show_year` value, and doesn't need to have it returned from the function.

pontoon/contributors/templates/contributors/includes/timeline.html
```diff
@@ -1,25 +1,76 @@
...
+                {% endfor %}
+            </ul>
+        </a>
+    </li>
+
+    {% endfor %}
+</ul>
+
+{% endfor %}
+{% endif %}
```
mathjazz, line 76: Nit: mind the blank line.

pontoon/insights/utils.py
```diff
@@ -460,6 +460,6 @@ def get_global_pretranslation_quality(category, id):
     ]
 
     return {
-        "dates": list({convert_to_unix_time(x["month"]) for x in actions}),
+        "dates": sorted(list({convert_to_unix_time(x["month"]) for x in actions})),
```
mathjazz, line 463: This is already merged, so it shouldn't appear in the diff:
https://github.com/mozilla/pontoon/commit/136e382a1137ae748d0668ed406b04538d30e709

pontoon/contributors/static/css/profile.css
```diff
@@ -446,6 +446,27 @@ svg.js-calendar-graph-svg {
...
+  background: var(--dark-grey-1);
+  font-size: 16px;
+  font-weight: normal;
+  height: 32px;
+  text-transform: uppercase;
+  width: 100%;
+  border: none;
+  border-radius: 3px;
+  color: var(--light-grey-7);
+  margin-top: 5px;
```
mathjazz, line 463:
```suggestion
  margin-top: 20px;
```

pontoon/contributors/utils.py
```diff
@@ -420,20 +465,81 @@ def get_project_locale_contribution_counts(contributions_qs):
...
     """
     Get data required to render the Contribution timeline on the Profile page
     """
     end = timezone.now()
-    start = end - relativedelta(months=1)
-    timeline_title = "Contribution activity in the last month"
+
+    if year_shown:
+        start = end - relativedelta(years=1, day=1)
+        timeline_title = "Contribution activity in the last year"
```
mathjazz, line 539: How about we just hardcode this to "Contribution activity" in the template?

Then we'd need to always show month + year in the timeline - even for this month's data - which would result in a more consistent UX and simpler code with less IF-ELSE blocks.

pontoon/contributors/templates/contributors/includes/timeline.html
```diff
@@ -1,25 +1,46 @@
...
+        <div class="month-group">
+            <h4>{{ month }}</h4>
+            {% for contribution_type, details in contribution_types.items() %}
+                {{ render_contribution_group(details.title, contribution_type, details.entries) }}
+            {% endfor %}
+        </div>
+    {% endfor %}
+{% else %}
+    {% for title, project_data in contribution_timeline.contributions.items() %}
+        {{ render_contribution_group(title, project_data.type, project_data.data.values() | list) }}
```
mathjazz, line 44: That `list` at the end should be dropped - a more proper approach would be to set the default value to `entries` in `render_contribution_group()`.

pontoon/contributors/templates/contributors/includes/timeline.html
```diff
@@ -1,25 +1,46 @@
...
+    {% for month, contribution_types in contribution_timeline.contributions.items() %}
+        <div class="month-group">
+            <h4>{{ month }}</h4>
+            {% for contribution_type, details in contribution_types.items() %}
+                {{ render_contribution_group(details.title, contribution_type, details.entries) }}
+            {% endfor %}
+        </div>
+    {% endfor %}
+{% else %}
+    {% for title, project_data in contribution_timeline.contributions.items() %}
```
mathjazz, line 43: I find it confusing that we have different "APIs" for this month's data and for the full year data. I would expect the backend functions to return data in the exact same format, but for different periods, so we wouldn't need to IF-ELSE here and in one or two other places.

eemeli (commented):
I tried removing myself as a reviewer, but that didn't seem to work. I'm fine with this if Matjaž is, but please explicitly re-request a review from me if one is desired.

pontoon/contributors/utils.py
```diff
@@ -478,39 +488,70 @@ def get_contribution_timeline_data(user, contribution_type=None, day=None):
...
+        for month, data in contribution_data.items():
+            total_count = 0
+            projects = set()
+            for key, info in data.items():
+                total_count += info["count"]
+                # Count number of unique project/locale pairs
+                projects.add(key)
+
+            # Generate title for the localizations belonging to the same contribution type
+            p_count = len(projects)
```
mathjazz, line 501: This is the logic we use currently, I think it's simpler:

```suggestion
            total_count = sum([info["count"] for _, info in data.items()])
            p_count = len(data)

            # Generate title for the localizations belonging to the same contribution type
```

pontoon/contributors/utils.py
```diff
@@ -478,39 +488,70 @@ def get_contribution_timeline_data(user, contribution_type=None, day=None):
...
 
         contributions.update(
             {
-                title: {
+                contribution_type: {
                     "data": contribution_data,
                     "type": contribution_type.replace("_", "-"),
                 }
             }
         )
```
mathjazz, line 536: We should directly structure the `contributions` dict the way we need data to be structured. We should be able to do that without the additional `restructured` dict.

pontoon/contributors/utils.py
```diff
@@ -478,39 +488,70 @@ def get_contribution_timeline_data(user, contribution_type=None, day=None):
...
-            contribution_data[key]["url"] = f"{url}?{urlencode(params)}"
-
-        # Generate title for the localizations belonging to the same contribution type
-        if contribution_type == "user_translations":
-            title = f"Submitted { intcomma(c_count) } translation{ pluralize(c_count) } in { intcomma(p_count) } project{ pluralize(p_count) }"
-        elif contribution_type == "user_reviews":
-            title = f"Reviewed { intcomma(c_count) } suggestion{ pluralize(c_count) } in { intcomma(p_count) } project{ pluralize(p_count) }"
-        elif contribution_type == "peer_reviews":
-            title = f"Received review for { intcomma(c_count) } suggestion{ pluralize(c_count) } in { intcomma(p_count) } project{ pluralize(p_count) }"
+        for key, value in contribution_data.items():
```
mathjazz, line 516: We already have this loop above. Why don't we move this block back there?

pontoon/contributors/utils.py
```diff
@@ -478,39 +488,70 @@ def get_contribution_timeline_data(user, contribution_type=None, day=None):
...
+                title = f"Submitted { intcomma(total_count) } translation{ pluralize(total_count) } in { intcomma(p_count) } project{ pluralize(p_count) }"
+            elif contribution_type == "user_reviews":
+                title = f"Reviewed { intcomma(total_count) } suggestion{ pluralize(total_count) } in { intcomma(p_count) } project{ pluralize(p_count) }"
+            elif contribution_type == "peer_reviews":
+                title = f"Received review for { intcomma(total_count) } suggestion{ pluralize(total_count) } in { intcomma(p_count) } project{ pluralize(p_count) }"
+
+            # Add title as a field
+            if month not in contribution_data:
+                contribution_data[month] = {}
+            contribution_data[month]["title"] = title
```
mathjazz, line 512: We should avoid such mutations of existing dicts, otherwise we need to do things like `if field != "title":`, which aren't pretty.

Instead, we should be able to store title directly to the new `contributions` dict.

mathjazz (commented):
Nice work!

Left some comments inline to further simplify the code.

Please also note that the bubbles within the same month are stacked together - we should add margin.

Screenshot 2024-07-18 at 22 29 28

pontoon/contributors/templates/contributors/includes/timeline.html
```diff
@@ -1,22 +1,27 @@
...
-        {% endfor %}
-    </ul>
+    {% for _, info in data.items() %}
+        <div class="contribution-type">
+            <h4 class="heading">
+                <span class="icon far {{ info.type }}"></span>
+                {{ info.title }}
+            </h4>
+            <ul class="localizations">
+                {% for key, localization in info.data.items() %}
```
mathjazz, line 13: Since we don't use `key`, let's do this:

```suggestion
                {% for _, localization in info.data.items() %}
```

pontoon/contributors/utils.py
```diff
@@ -478,39 +490,55 @@ def get_contribution_timeline_data(user, contribution_type=None, day=None):
...
+
+            # Generate title for the localizations belonging to the same contribution type
+            if contribution_type == "user_translations":
+                title = f"Submitted { intcomma(total_count) } translation{ pluralize(total_count) } in { intcomma(p_count) } project{ pluralize(p_count) }"
+            elif contribution_type == "user_reviews":
+                title = f"Reviewed { intcomma(total_count) } suggestion{ pluralize(total_count) } in { intcomma(p_count) } project{ pluralize(p_count) }"
+            elif contribution_type == "peer_reviews":
+                title = f"Received review for { intcomma(total_count) } suggestion{ pluralize(total_count) } in { intcomma(p_count) } project{ pluralize(p_count) }"
+
+            data["title"] = title
```
mathjazz, line 505: Please remove this line...

pontoon/contributors/utils.py
```diff
@@ -478,39 +490,55 @@ def get_contribution_timeline_data(user, contribution_type=None, day=None):
...
+
+            # Generate localization URL and add it to the data dict
+            params = params_map[contribution_type]
+            for field, val in data.items():
+                if field != "title":
```
mathjazz, line 510: and this one...

pontoon/contributors/utils.py
```diff
@@ -478,39 +490,55 @@ def get_contribution_timeline_data(user, contribution_type=None, day=None):
...
+                        "pontoon.translate",
+                        args=[
+                            val["locale"]["code"],
+                            val["project"]["slug"],
+                            "all-resources",
+                        ],
+                    )
+                    val["url"] = f"{url}?{urlencode(params)}"
+
+        for month, data in contribution_data.items():
```
mathjazz, line 521: and this one...

pontoon/contributors/utils.py
```diff
@@ -478,39 +490,55 @@ def get_contribution_timeline_data(user, contribution_type=None, day=None):
...
+                contributions[month] = {}
+
+            if contribution_type not in contributions[month]:
+                contributions[month][contribution_type] = {}
+
+            contributions[month][contribution_type].update(
+                {
+                    "data": data,
                     "type": contribution_type.replace("_", "-"),
+                    "title": data.pop("title"),
```
mathjazz, line 532: And just do this here:
```suggestion
                    "title": title,
```

mathjazz:
@harmitgoswami I've [updated the CSS slightly](https://github.com/mozilla/pontoon/pull/3255/commits/b6ae8d0b057e78a65c3ab2ed62d4dc865fcf0c7d) as we agreed. Let me know if changes look good to you.

@flodolo Could you please test this a little on stage?

flodolo:
Seems to work correctly as far as I can tell (but stage doesn't have recent activity, so the list is always empty until I click the button).

harmitgoswami:
> @harmitgoswami I've [updated the CSS slightly](https://github.com/mozilla/pontoon/pull/3255/commits/b6ae8d0b057e78a65c3ab2ed62d4dc865fcf0c7d) as we agreed. Let me know if changes look good to you.
> 
> @flodolo Could you please test this a little on stage?

LGTM!

Issue #3244, opened 2024-05-30
Add ability to filter users by various criteria when sending messages


mathjazz:
Various filters can be used to limit the target audience for the message.

Spec:
https://github.com/mozilla/pontoon/blob/main/specs/0118-messaging-center.md#selecting-recipients-with-filters

Issue #3246, opened 2024-05-30
Add ability to see the Sent messages


mathjazz:
Display a list of all previously sent emails or in-app messages from the Messaging Center.

Spec:
https://github.com/mozilla/pontoon/blob/main/specs/0118-messaging-center.md#review-past-message

RafaelJohn9:
Hello @mathjazz  I would like to be assigned this issue

mathjazz:
Hey @RafaelJohn9, sadly this issue is not ready yet, because we still need designs for it. Also, it's part of our team's plan for Q3, so we'd prefer to implement it ourselves.

Let me know if you need help picking another issue!

RafaelJohn9:
Sure @mathjazz , which one has been bothering?

PR #3259, opened 2024-06-19
Add ability to send emails and notifications from the Messaging Center

### LINKED ISSUE #2072: Add ability to send manual notifications via email
## [Bug 1374577](https://bugzilla.mozilla.org/show_bug.cgi?id=1374577)
Bug Reporter: @mathjazz
CC:   @TheoChevalier
Depends on:   [Bug 1379158](https://bugzilla.mozilla.org/show_bug.cgi?id=1379158)

Send should have an option (a checkbox) to send manual notifications via email.

As part of this bug, we should also investigate the feasibility of sending push notifications.

### LINKED ISSUE #3243: Add ability to send messages to Pontoon users
This issue only tracks the implementation of the message editor and the ability to send its contents as a (transactional) email and/or an in-app notification.

An unsubscribe link is appended to messages as per the spec.

Spec:
https://github.com/mozilla/pontoon/blob/main/specs/0118-messaging-center.md#messaging-center

mathjazz:
Fix #3243.
Fix #2072.

This is the first part of the implementation of the Messaging Center. It allows for sending emails and in-app notifications from `/messaging/`. Locally emails are sent to stdout.

Note that until we implement recipient filters (#3244), the recipient of the messages is the sender.

While the development is in progress, the page is not linked from the UI.

Also included:
- Factor out check-box CSS
- Make .notification selectors in style.css and main.js more specific

### DIFF
--- a/pontoon/base/static/css/check-box.css
+++ b/pontoon/base/static/css/check-box.css
@@ -4,33 +4,33 @@
   margin: 0;
   display: inline-block;
   text-align: left;
-}
 
-#main .check-list .check-box {
-  padding: 8px 0;
-}
+  .check-box {
+    padding: 8px 0;
+  }
 
-#main .check-list .check-box:last-child {
-  padding-bottom: 0;
-}
+  .check-box:last-child {
+    padding-bottom: 0;
+  }
 
-#main .check-list .check-box-wrapper {
-  color: var(--light-grey-7);
-  display: table;
-  font-size: 16px;
-  font-weight: 300;
-  overflow: hidden;
-  text-align: right;
-  text-overflow: ellipsis;
-  white-space: nowrap;
-  width: 342px;
-}
+  .check-box-wrapper {
+    color: var(--light-grey-7);
+    display: table;
+    font-size: 16px;
+    font-weight: 300;
+    overflow: hidden;
+    text-align: right;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    width: 342px;
+  }
 
-#main .check-list .check-box-wrapper:hover {
-  color: var(--white-1);
-}
+  .check-box-wrapper:hover {
+    color: var(--white-1);
+  }
 
-#main .check-list .fa {
-  margin-left: 27px;
-  margin-right: 0;
+  .fa {
+    margin-left: 27px;
+    margin-right: 0;
+  }
 }
--- a/pontoon/base/static/css/check-box.css
+++ b/pontoon/base/static/css/check-box.css
@@ -1,0 +1,36 @@
+#main .check-list {
+  cursor: pointer;
+  list-style: none;
+  margin: 0;
+  display: inline-block;
+  text-align: left;
+}
+
+#main .check-list .check-box {
+  padding: 8px 0;
+}
+
+#main .check-list .check-box:last-child {
+  padding-bottom: 0;
+}
+
+#main .check-list .check-box-wrapper {
+  color: var(--light-grey-7);
+  display: table;
+  font-size: 16px;
+  font-weight: 300;
+  overflow: hidden;
+  text-align: right;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  width: 342px;
+}
+
+#main .check-list .check-box-wrapper:hover {
+  color: var(--white-1);
+}
+
+#main .check-list .fa {
+  margin-left: 27px;
+  margin-right: 0;
+}
--- a/pontoon/messaging/utils.py
+++ b/pontoon/messaging/utils.py
@@ -1,0 +1,10 @@
+import html2text
+
+
+def html_to_plain_text_with_links(html):
+    h = html2text.HTML2Text()
+
+    h.wrap_links = False
+    h.body_width = 0
+
+    return h.handle(html)
--- a/pontoon/messaging/views.py
+++ b/pontoon/messaging/views.py
@@ -1,6 +1,12 @@
 import json
+import uuid
+
+from guardian.decorators import permission_required_or_403
+from notifications.signals import notify
 
 from django.contrib.auth.decorators import login_required
+from django.core.exceptions import PermissionDenied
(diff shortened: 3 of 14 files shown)


pontoon/base/static/css/check-box.css
eemeli, line 1: CSS nesting should now be sufficiently [well supported](https://caniuse.com/css-nesting) that it could be used here. And in other files as well, of course.

pontoon/messaging/utils.py
```diff
@@ -0,0 +1,10 @@
+import html2text
+
+
+def html_to_plain_text_with_links(html):
```
eemeli, line 4: I would just inline this and consider DRY much later, but that's mostly just me.

pontoon/messaging/views.py
```diff
@@ -1,13 +1,93 @@
...
+def send_message(request):
+    # Send notifications
+    if request.method == "POST":
+        form = forms.MessageForm(request.POST)
+
+        if not form.is_valid():
+            return JsonResponse(dict(form.errors.items()))
+
+        # TODO: implement recipient filters
+        from django.contrib.auth.models import User
```
eemeli, line 44: Why is this not a top-level import?

pontoon/messaging/views.py
```diff
@@ -1,13 +1,93 @@
...
+
+        recipients = User.objects.filter(pk=request.user.pk)
+
+        is_notification = form.cleaned_data.get("notification")
+        is_email = form.cleaned_data.get("email")
+        subject = form.cleaned_data.get("subject")
+        body = form.cleaned_data.get("body")
+
+        if is_notification:
+            identifier = uuid.uuid4().hex
```
eemeli, line 54: Is this actually used for anything?

pontoon/messaging/views.py
```diff
@@ -1,13 +1,93 @@
...
+
+            for recipient in recipients.distinct():
+                unique_id = str(recipient.profile.unique_id)
+                text = text_template.replace("{ uuid }", unique_id)
+                html = html_template.replace("{ uuid }", unique_id)
+
+                msg = EmailMultiAlternatives(
+                    subject=subject,
+                    body=text,
+                    from_email="Mozilla L10n Team <team@pontoon.mozilla.com>",
```
eemeli, line 80: This probably should not be hard-coded?

eemeli (approved):
Overall looks fine. A bit sad that we're extending the jQuery dependency, but it's probably the least worst option. Some inline comments might require some fixing.

pontoon/messaging/views.py
mathjazz: Yes, that allows us to find all instances of the same notification (one instance is created for each user as per `django-notifications-hq`).

To implement #3246, we'll need a new data model and than we can drop this.

pontoon/messaging/utils.py
mathjazz: I'm planning to use this function in the [email script](https://github.com/mozilla-l10n/pontoon-scripts/blob/c06950b4761f4208eacf39496b4388980a1f3dda/admin/email_script_events.py), at least for the time being.

pontoon/messaging/views.py
mathjazz: It should be.

pontoon/messaging/views.py
mathjazz: Indeed.

mathjazz:
@eemeli Please let me know if the CSS changes look good.

pontoon/base/static/css/check-box.css
```diff
@@ -4,33 +4,33 @@
...
   display: inline-block;
   text-align: left;
-}
 
-#main .check-list .check-box {
-  padding: 8px 0;
-}
+  .check-box {
+    padding: 8px 0;
+  }
```
eemeli, line 10: The syntax for CSS nesting requires an `&` nesting selector, unlike how it works with SCSS & less: https://kilianvalkhof.com/2021/css-html/css-nesting-specificity-and-you/

mathjazz: That seems to be optional according to MDN:
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_nesting/Using_CSS_nesting

bcolsson:
It looks like we're hardcoding the unsubscribe footer for all emails here. In the [spec](https://github.com/mozilla/pontoon/blob/main/specs/0118-messaging-center.md#flagging-as-transactional) the plan is to have the ability to flag a message as transactional or not, then only attach the footer when an email is not transactional.

I also realize that this was not covered by the spec, but do we also plan to adjust how messages sent as notifications appear in a users Notifications window? Testing this out it seems like a project is required and is defaulting to `Pontoon Intro`. Perhaps we could use only `[user] has sent a message.` in this instance?
Screenshot 2024-06-21 at 17 03 45

mathjazz:
> It looks like we're hardcoding the unsubscribe footer for all emails here. In the [spec](https://github.com/mozilla/pontoon/blob/main/specs/0118-messaging-center.md#flagging-as-transactional) the plan is to have the ability to flag a message as transactional or not, then only attach the footer when an email is not transactional.

I was planning to add support for transactional email in a separate PR, which would also include special handling of the footer. I'll see if I can find some time today to amend this PR with this change.

> I also realize that this was not covered by the spec, but do we also plan to adjust how messages sent as notifications appear in a users Notifications window? Testing this out it seems like a project is required and is defaulting to `Pontoon Intro`. Perhaps we could use only `[user] has sent a message.` in this instance?

I assume you tested on stage (which uses an old version of this patch). I'll deploy the current state.

bcolsson:
I'm okay with adding it to a later PR, but I didn't see an issue tracking this, so wanted to make sure this would be covered.

>I assume you tested on stage (which uses an old version of this patch). I'll deploy the current state of the PR. Should be up in 5 minutes.

Thanks, yes that was what I was doing. Looks good now. 👍

PR #3265, opened 2024-06-28
Revert using bulk_create for new projectlocales

### LINKED ISSUE #3262: Translator rights don't seem to work
CC @canova 

I've added a new project for Turkisk (Firefox Profiler), and added [Nazim](https://pontoon.mozilla.org/contributors/canova/) as a translator. For some reason, he doesn't get translator rights.

I've checked the database, and everything looks correct (he's added to the group).

I even tried adding him to the full Translators group, and still no luck.

I think we have plenty of locales with permissions per-project, and we didn't get bug reports. Is this a caching problem?

eemeli:
Fixes #3262

### DIFF
--- a/pontoon/administration/views.py
+++ b/pontoon/administration/views.py
@@ -183,6 +183,8 @@ def manage_project(request, slug=None, template="admin_project.html"):
                 )
 
                 for locale in locales:
+                    # The implicit pre_save and post_save signals sent here are required
+                    # to maintain django-guardian permissions.
                     ProjectLocale.objects.get_or_create(project=project, locale=locale)
 
                 project_locales = ProjectLocale.objects.filter(project=project)
--- a/pontoon/administration/views.py
+++ b/pontoon/administration/views.py
@@ -182,12 +182,8 @@ def manage_project(request, slug=None, template="admin_project.html"):
                     .delete()
                 )
 
-                project_locales = [
-                    ProjectLocale(project=project, locale=locale) for locale in locales
-                ]
-                ProjectLocale.objects.bulk_create(
-                    project_locales, ignore_conflicts=True
-                )
+                for locale in locales:
+                    ProjectLocale.objects.get_or_create(project=project, locale=locale)
 
                 project_locales = ProjectLocale.objects.filter(project=project)
 
--- a/pontoon/base/migrations/0065_fix_projectlocale_permissions.py
+++ b/pontoon/base/migrations/0065_fix_projectlocale_permissions.py
@@ -1,0 +1,45 @@
+from django.db import migrations
+
+from pontoon.base.models import ProjectLocale
+from pontoon.base.signals import assign_group_permissions, create_group
+
+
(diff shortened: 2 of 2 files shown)


pontoon/administration/views.py
```diff
@@ -182,12 +182,8 @@ def manage_project(request, slug=None, template="admin_project.html"):
...
                 )
 
-                project_locales = [
-                    ProjectLocale(project=project, locale=locale) for locale in locales
-                ]
-                ProjectLocale.objects.bulk_create(
-                    project_locales, ignore_conflicts=True
-                )
+                for locale in locales:
+                    ProjectLocale.objects.get_or_create(project=project, locale=locale)
```
mathjazz, line 186: Could you add a comment here why don't we use `bulk_create` here, to avoid making the same mistake in the future?

mathjazz (approved):
Looks good, I'd just add a comment.

Nice work! And sorry for the extra work. 🦆

pontoon/base/migrations/0065_fix_projectlocale_permissions.py
```diff
@@ -0,0 +1,45 @@
...
+def fix_projectlocale_permissions(apps, schema_editor):
+    for project, locale in [
+        ("common-voice", "nqo"),
+        ("firefox-bridge", "de"),
+        ("firefox-bridge", "fr"),
+        ("firefox-bridge", "it"),
+        ("firefox-multi-account-containers", "fur"),
+        ("firefox-profiler", "en-CA"),
+        ("firefox-profiler", "fur"),
+        ("firefox-profiler", "tr"),
```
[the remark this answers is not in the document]
eemeli: No; it doesn't track group permission creations. These are from the following SQL query, which isn't pretty but works:

```sql
select a.id as group_id, a.name
from auth_group as a
left join guardian_groupobjectpermission as g on a.id=g.group_id
where g.id is null
  and a.name not like 'terminology%'
  and a.name not like 'tutorial%';
```

pontoon/administration/views.py
eemeli:
```suggestion
                    # The implicit pre_save and post_save signals sent here are required
                    # to maintain django-guardian permissions.
                    ProjectLocale.objects.get_or_create(project=project, locale=locale)
```

pontoon/base/migrations/0065_fix_projectlocale_permissions.py
```diff
@@ -0,0 +1,45 @@
+from django.db import migrations
+
+from pontoon.base.models import ProjectLocale
+from pontoon.base.signals import assign_group_permissions, create_group
+
+
+def fix_projectlocale_permissions(apps, schema_editor):
+    for project, locale in [
+        ("common-voice", "nqo"),
```
eemeli, line 9: Missed a few earlier, apparently.

```suggestion
        ("common-voice", "cdo"),
        ("common-voice", "dar"),
        ("common-voice", "nqo"),
        ("common-voice", "shn"),
```

PR #3263, opened 2024-06-27
Add reviewed time information to reviewed strings tooltip

harmitgoswami:
Fix #2763 

Added a `reviewedDate` field to the tooltip when hovering over rejected or approved strings.

**To reproduce results**: After starting the server, navigate to any team/project (e.g. `http://localhost:8000/sq/facebook-container/messages.json/?string=196694` ). Hover over the User Avatar or Username to show the tooltip. The tooltip should now display `[Approved/Rejected] by { $user } on { $reviewedDate }` or `[Approved/Rejected] on { $reviewedDate }`

Screenshot 2024-06-27 at 12 15 53 PM

Screenshot 2024-06-27 at 12 16 42 PM

PR #3270, opened 2024-07-02
Corrected tooltip hover information

### LINKED ISSUE #2763: When hovering an approved string, the tooltip should also say when the strings was approved, not just by whom
Currently, it also says "Approved by X". To understand when it was reviewed, I need to look at VCS.

harmitgoswami:
Fix #2763 

Made changes to previous pull request to correctly display the last day an action was taken upon a translation. Additionally, removed the Weekday within the information displayed in the tooltip. See #3263 for more context.

Updated Tooltip:

Screenshot 2024-07-02 at 1 06 28 PM


pontoon/base/views.py
```diff
@@ -418,7 +418,13 @@ def get_translation_history(request):
                 "date": t.date.strftime("%b %d, %Y %H:%M"),
                 "date_iso": t.date.isoformat(),
                 "approved_user": User.display_name_or_blank(t.approved_user),
+                "approved_date": t.approved_user.latest_action.created_at
+                if t.approved_user and t.approved_user.latest_action
+                else None,
```
mathjazz, line 423: We can simply use `t.approved_date` here.

pontoon/base/views.py
```diff
@@ -418,7 +418,13 @@ def get_translation_history(request):
...
                 "rejected_user": User.display_name_or_blank(t.rejected_user),
+                "rejected_date": t.rejected_user.latest_action.created_at
+                if t.rejected_user and t.rejected_user.latest_action
+                else None,
```
mathjazz, line 427: Similar as above.

translate/public/locale/en-US/translate.ftl
```diff
@@ -369,13 +369,13 @@ history-Translation--span-copied =
     .title = Copied ({ $machinerySources })
 
 history-translation--approved =
-    .title = Approved by { $user } on { DATETIME($reviewedDate, dateStyle:"long", timeStyle:"medium") }
+    .title = Approved by { $user } on { DATETIME($reviewedDate, month: "long", year: "numeric", day: "numeric", hour: "numeric", minute: "numeric", second: "numeric") }
```
mathjazz, line 372: What's the purpose of this change?

translate/src/modules/history/components/HistoryTranslation.tsx
```diff
@@ -196,17 +196,23 @@ export function HistoryTranslationBase({
 
   const review = {
     id: 'history-translation--unreviewed',
-    vars: { user: '', reviewedDate: new Date(translation.dateIso) },
+    vars: { user: '', reviewedDate: new Date() },
```
mathjazz, line 199: Can we completely drop `reviewedDate` here?

translate/public/locale/en-US/translate.ftl
harmitgoswami: This change allows me to print the date without the Weekday as well. Is there a more succinct way of doing this?

translate/src/modules/history/components/HistoryTranslation.tsx
harmitgoswami: Potentially... is there a way for me to pass `approvedDate` or `rejectedDate` to the .ftl file if we do remove this variable?

translate/public/locale/en-US/translate.ftl
mathjazz: This format looks OK. It's the tooltip shown when you hover the translation creation date that is showing the weekday:
https://github.com/mozilla/pontoon/issues/2763#issuecomment-2203002727

translate/src/modules/history/components/HistoryTranslation.tsx
mathjazz: Ah, I missed that this was part of an FTL message.

pontoon/base/views.py
```diff
@@ -418,7 +418,9 @@ def get_translation_history(request):
                 "date": t.date.strftime("%b %d, %Y %H:%M"),
                 "date_iso": t.date.isoformat(),
```
mathjazz, line 419: We no longer use `translation.date` on frontend, so we can drop it. We can also drop `.isoformat()`. At that point we should rename `date_iso` to `date` (also on the frontend).

```suggestion
                "date": t.date,
```

translate/src/modules/history/components/HistoryTranslation.tsx
```diff
@@ -227,6 +233,17 @@ export function HistoryTranslationBase({
...
   let canComment = user.isAuthenticated;
 
+  const customDateFormat = (date: Date) => {
+    return new Intl.DateTimeFormat('en', {
+      year: 'numeric',
+      month: 'long',
+      day: 'numeric',
+      hour: 'numeric',
+      minute: 'numeric',
+      second: 'numeric',
```
mathjazz, line 243: To be consistent with the format used for the review timestamp and more concise, let's use this:
```suggestion
      dateStyle: 'long',
      timeStyle: 'medium',
```

translate/src/modules/history/components/HistoryTranslation.tsx
mathjazz: Can we use `''` instead of `new Date()`?

harmitgoswami: I seem to get a type error if I try to switch to `''`. Should I still try to move forward with using `''` instead of `new Date()`?

Screenshot 2024-07-03 at 1 53 47 PM

mathjazz: Makes sense. Let's keep `new Date()`.

translate/src/modules/history/components/HistoryTranslation.tsx
```diff
@@ -282,8 +295,9 @@ export function HistoryTranslationBase({
                 </Localized>
                 <ReactTimeAgo
                   dir='ltr'
-                  date={new Date(translation.dateIso)}
-                  title={`${translation.date} UTC`}
+                  date={new Date(translation.date)}
+                  title=''
```
mathjazz, line 299: Seems like we can remove the `title` attribute?

mathjazz (approved):
Nice work! Only one comment left.

PR #3272, opened 2024-07-03
Add ability to send transactional emails

### LINKED ISSUE #3268: Add ability to send Transactional emails
This issue tracks the implementation of the ability to send _transactional_ emails, which was not implemented as part of #3259.

Spec:
https://github.com/mozilla/pontoon/blob/main/specs/0118-messaging-center.md#messaging-center

mathjazz:
Fix #3268.

If the email is marked as "transactional", we don't include the Unsubscribe footer.

Spec:
https://github.com/mozilla/pontoon/blob/main/specs/0118-messaging-center.md#flagging-as-transactional

I've made the "Transactional" flag only visible when the Email option is checked. This patterns prevents selecting the Transactional option when Email is not checked.

### DIFF
--- a/pontoon/base/static/css/check-box.css
+++ b/pontoon/base/static/css/check-box.css
@@ -25,6 +25,7 @@
     width: 342px;
   }
 
+  .check-box.hover .check-box-wrapper,
   .check-box-wrapper:hover {
     color: var(--white-1);
   }
--- a/pontoon/messaging/forms.py
+++ b/pontoon/messaging/forms.py
@@ -6,5 +6,6 @@ from pontoon.base.forms import HtmlField
 class MessageForm(forms.Form):
     notification = forms.BooleanField(required=False)
     email = forms.BooleanField(required=False)
+    transactional = forms.BooleanField(required=False)
     subject = forms.CharField()
     body = HtmlField()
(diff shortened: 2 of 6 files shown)


eemeli (approved):
Changes here look fine.

I had not looked at the UI for this before, is this intended as final?
Screenshot 2024-07-04 at 11 32 05

Some comments:
- It's not at all obvious that the 🚫 is a checkbox indicator; to me this seems to be saying "Sending notifications and emails is not available". Having some different indicator for an unselected option could be much clearer.
- The `#1` and `#2` parts of the title don't seem to mean anything?
- Is the "Message editor" title necessary?

mathjazz:
> I had not looked at the UI for this before, is this intended as final?

There are massive UI changes coming in the subsequent patches (especially user filters), so I'd revisit the UI when we have the entire functionality in place.

mathjazz:
@bcolsson Gentle ping for the review. I'd be especially interested in your take on this deviation from the spec:

> I've made the "Transactional" flag only visible when the Email option is checked. This patterns prevents selecting the Transactional option when Email option is not checked and makes the extra validation required by the spec redundant.

bcolsson (approved):
Thanks for the ping. The proposed pattern makes sense, so I'm ok with the PR as proposed.

PR #3273, opened 2024-07-03
Enhance user action logging

### LINKED ISSUE #3269: Enhance User Action Logging for LLM Feature
Enhance user action logging for the LLM feature by adding locale information (locale name and locale code) when an LLM translation is copied. Additionally, update the logging for dropdown selections to record the locale code instead of the locale name for the selected option.

ayanaar:
Fix #3269 

Updated user action logging by adding locale name and locale code when an LLM translation is copied. Additionally, updated the logging for dropdown selections to record the locale code instead of the locale name for the selected option.

### DIFF
--- a/translate/src/modules/machinery/components/MachineryTranslation.tsx
+++ b/translate/src/modules/machinery/components/MachineryTranslation.tsx
@@ -40,6 +40,7 @@ export function MachineryTranslationComponent({
   const isSelected = element === index;
 
   const { llmTranslation } = useLLMTranslation(translation);
+  const locale = useContext(Locale);
 
   const copyTranslationIntoEditor = useCallback(() => {
     if (window.getSelection()?.isCollapsed !== false) {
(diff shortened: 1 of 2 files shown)


translate/src/modules/machinery/components/MachineryTranslation.tsx
```diff
@@ -52,6 +53,8 @@ export function MachineryTranslationComponent({
       if (llmTranslation) {
         logUXAction('LLM Translation Copied', 'LLM Feature Adoption', {
           action: 'Copy LLM Translation',
+          localeCode: locale.code,
+          targetLanguage: locale.name,
```
mathjazz, line 57: I wouldn't include `targetLanguage`, because `localeCode` is already unambiguous. It's also better to keep both related UX actions that are logged consistent in that regard.

mathjazz (approved):
LGTM.

@flodolo Do you have any additional requests for logging UX actions?

flodolo:
> @flodolo Do you have any additional requests for logging UX actions?

No, that's enough for now.

PR #3278, opened 2024-07-11
Sorted date labels for insights chart

### LINKED ISSUE #3276: Charts on Insights page jump back and forth between Dec and Oct 2023
https://pontoon.mozilla.org/insights/

I've only kept one locale to show the issue

Screenshot 2024-07-10 alle 11 01 11

harmitgoswami:
Fixes #3276 

Fixed the bug on the [Insights page](https://pontoon.mozilla.org/insights/) where the graph shown was not a [function](https://en.wikipedia.org/wiki/Vertical_line_test) (i.e. line graph would jump back and forth horizontally). 

The issue seemed to be that the dates being passed through `chart.data('dates')` were not chronologically ordered by default.

Before:
image
image

After:
image
image

### DIFF
--- a/pontoon/insights/static/js/insights.js
+++ b/pontoon/insights/static/js/insights.js
@@ -63,10 +63,18 @@ var Pontoon = (function (my) {
           };
         });
 
+        // Sort each timestamp chronologically and reorder the dataset
+        const sortedIndices = chart
+          .data('dates')
+          .map((timestamp, index) => ({ timestamp, index }))
+          .sort((a, b) => a.timestamp - b.timestamp)
+          .map((dateObj) => dateObj.index);
+        const sortedDates = sortedIndices.map((i) => chart.data('dates')[i]);
+
         new Chart(chart, {
           type: 'bar',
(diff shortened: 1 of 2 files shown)


mathjazz (commented):
Excellent catch!

We actually order `actions` by month:
https://github.com/mozilla/pontoon/blob/e9e9f5f3dec40bb5f34c58086c267922ab1fd405/pontoon/insights/utils.py#L412

But then we use `set()` when extracting unique month values, which drops the order:
https://github.com/mozilla/pontoon/blob/e9e9f5f3dec40bb5f34c58086c267922ab1fd405/pontoon/insights/utils.py#L463

We should fix the order on backend, so that it always returns the expected data. It also simpler to do it there - just wrap the `list()` expression in `sorted()`:
https://github.com/mozilla/pontoon/blob/e9e9f5f3dec40bb5f34c58086c267922ab1fd405/pontoon/insights/utils.py#L463

mathjazz (approved):

PR #3288, opened 2024-07-31
Implement search filter for context identifiers

### LINKED ISSUE #3221: Search: exclude message identifiers by default, add option to include them
Per conversation on improvements to Pontoon search, we should stop including message identifiers by default.

A dropdown should be available in the search field (likely on the lens icon), with a `SEARCH OPTIONS` title group, with a checkbox to `Search in string identifiers`. Selecting this checkbox will include message identifiers in the search.

harmitgoswami:
Fixes #3221 

Excluded context identifiers by default when searching strings on the `translate` view. Now, there is a search panel, identical in style to the existing filter panel, which allows users to explicitly include context identifiers.

**To reproduce results**: Head to any string within the `translate` view upon running the server. Then, click on the magnifying glass icon on the right side of the search box. This action will open up the search panel, which currently only has one option. Either click the check mark icon to toggle the filter, or click the name of the filter and enter your search again (See note below).

Side note: The click behaviour for filters seems a little wonky, in both `SearchPanel` and `FiltersPanel`. I've ignored simplifying the behaviour for this PR.

image


mathjazz (commented):
Very nice work!

As mentioned inline, I'd keep "search options" separate from filters, both in the codebase and the UX behavior:

- I wouldn't include search options in the placeholder, because "Search in Search context identifiers" doesn't sounds right. (At some point, we should refactor the placeholder to incorporate search options and fix issues with the filters.)

- I also wouldn't trigger search after selecting a search options, so clicking on the icon or text should result in the same behavior, i.e. the text should be a label for the checkbox.

- I'd rename the "Apply N filter" button to just "Search".

- I'd drop the Clear button.

translate/src/modules/search/components/SearchPanel.css
eemeli, line 1: Let's start using [CSS nesting](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_nesting/Using_CSS_nesting), esp. when adding new stylesheets.

translate/src/modules/search/constants.ts
```diff
@@ -58,3 +58,26 @@ export const FILTERS_EXTRA = [
     slug: 'missing-without-unreviewed',
   },
 ] as const;
+
+export const FILTERS_SEARCH = [
```
mathjazz, line 62:
```suggestion
export const SEARCH_OPTIONS = [
```

translate/src/modules/search/constants.ts
```diff
@@ -58,3 +58,26 @@ export const FILTERS_EXTRA = [
...
+  {
+    name: 'Search context identifiers',
```
mathjazz, line 64:
```suggestion
    name: 'Search in string identifiers',
```

translate/src/modules/search/constants.ts
```diff
@@ -58,3 +58,26 @@ export const FILTERS_EXTRA = [
...
+    slug: 'exclude_identifiers',
+  },
+  // {
+  //   name: 'Search translations only',
```
mathjazz, line 68:
```suggestion
  //   name: 'Search in translations only',
```

translate/src/modules/search/constants.ts
```diff
@@ -58,3 +58,26 @@ export const FILTERS_EXTRA = [
...
+  //   slug: 'translations',
+  // },
+  // {
+  //   name: 'Search rejected translations',
```
mathjazz, line 72:
```suggestion
  //   name: 'Search in rejected translations',
```

mathjazz (commented):
I didn't get the chance to have a closer look at the code yet, but this seems to be working pretty close to what we have in the spec. Nice work!

As mentioned on Slack, let's make the Search button look like a button and always visible.

I've also updated the label text in https://github.com/mozilla/pontoon/issues/3221, so please rename it to "Search in string identifiers".

To better reflect that in the URL parameter, let's call it `search_identifiers`.

translate/src/modules/search/components/SearchBox.tsx
```diff
@@ -278,6 +291,13 @@ export function SearchBoxBase({
         setTimeRange={setTimeRange}
         updateFiltersFromURL={updateFiltersFromURL}
       />
+      <SearchPanel
+        filters={filters}
+        resetFilters={resetFilters}
+        applyFilters={applyFilters}
+        applySingleFilter={applySingleFilter}
+        toggleFilter={toggleFilter}
```
mathjazz, line 299: Why do we need all these Filters data and logic in the search options? It's a separate panel, so I would expect it to store data separately.

translate/src/modules/search/components/SearchBox.tsx
```diff
@@ -38,7 +39,12 @@ type InternalProps = Props & {
   dispatch: AppDispatch;
 };
 
-export type FilterType = 'authors' | 'extras' | 'statuses' | 'tags';
+export type FilterType =
+  | 'authors'
+  | 'extras'
+  | 'statuses'
+  | 'search_identifiers'
```
mathjazz, line 46: I find it confusing that Filters and Search options are two separate things in the UI, but in the code we mix them.

How difficult would it be to move search options out of `FilterType` and `FilterState`, and then also stop calling `applyFilters` and `toggleFilter` from the `SearchPanel`?

pontoon/base/models/entity.py
```diff
@@ -843,24 +845,60 @@ def for_project_locale(
...
+            else:
+                entities = entities.annotate(
+                    filtered_string=Substr("string", StrIndex("string", Value("=")) + 1)
+                )
+                translation_filters = (
+                    Q(filtered_string__icontains_collate=(search, locale.db_collation))
+                    & Q(translation__locale=locale)
+                    for search in search_list
+                )
+                translation_matches = entities.filter(*translation_filters).values_list(
```
mathjazz, line 871: You can avoid duplicating `translation_matches` by simply dedenting it and removing it from the if block above.

Similarly, you also don't need to repeat `translation_filters` if you express the first `Q()` as `Q({'keyword': value})`.

pontoon/base/models/entity.py
```diff
@@ -843,24 +845,60 @@ def for_project_locale(
...
+                    )
+                    & Q(translation__locale=locale)
+                    for search in search_list
+                )
+                translation_matches = entities.filter(*translation_filters).values_list(
+                    "id", flat=True
+                )
+            else:
+                entities = entities.annotate(
+                    filtered_string=Substr("string", StrIndex("string", Value("=")) + 1)
```
mathjazz, line 864: This is only relevant for Fluent strings. Not sure if adding a condition for this will make the query faster or slower.

mathjazz: Also, maybe we should call it `string_without_identifier` or `string_without_id` rather than `filtered_string`?

mathjazz: Wait, shouldn't we here be annotating `filtered_string` (or whatever we call it) to translations rather than entities?

pontoon/base/models/entity.py
```diff
@@ -843,24 +844,41 @@ def for_project_locale(
...
+            if search_identifiers:
+                entity_matches = entities.filter(*entity_filters).values_list(
+                    "id", flat=True
+                )
+            else:
+                entity_matches = (
+                    entities.filter(*entity_filters)
+                    .exclude(Q(string__iregex=ftl_regex))
+                    .values_list("id", flat=True)
+                )
```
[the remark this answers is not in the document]
mathjazz: This if-else separation seems to be redundant again?

pontoon/base/models/entity.py
```diff
@@ -843,24 +845,60 @@ def for_project_locale(
...
+                    filtered_string=Substr("string", StrIndex("string", Value("=")) + 1)
+                )
+                entity_filters = (
+                    Q(filtered_string__icontains=search) for search in search_list
+                )
+            else:
+                entity_filters = (
+                    Q(string__icontains=search)
+                    | Q(string_plural__icontains=search)
+                    | q_key
```
mathjazz, line 888: You can drop `q_key` now:
```suggestion
                    | Q(key__icontains=search)
```

pontoon/base/models/entity.py
mathjazz: https://stackoverflow.com/a/8144600

translate/src/modules/search/components/SearchBox.tsx
```diff
@@ -92,7 +104,29 @@ export function SearchBoxBase({
...
-    { authors: [], extras: [], statuses: [], tags: [] },
+    {
+      authors: [],
+      extras: [],
+      statuses: [],
+      tags: [],
+    },
+  );
+
+  const [options, updateOptions] = useReducer(
```
mathjazz, line 115: Should we be more explicit, and always prepend `option(s)` with `search`, so `searchOptions`, `updateSearchOptions`?

translate/src/modules/search/components/SearchBox.tsx
```diff
@@ -118,9 +152,18 @@ export function SearchBoxBase({
...
+  const updateOptionsFromURL = useCallback(() => {
+    const { search_identifiers, time } = parameters;
+    updateOptions([
+      { option: 'search_identifiers', value: search_identifiers },
+    ]);
+    setTimeRange(time);
+  }, [parameters]);
+
   // When the URL changes, for example from links in the ResourceProgress
   // component, reload the filters from the URL parameters.
```
mathjazz, line 164: Nit: update the comment so also include search options.

translate/src/modules/search/components/SearchPanel.tsx
```diff
@@ -0,0 +1,140 @@
...
+import type { SearchState, SearchType } from './SearchBox';
+
+import { useOnDiscard } from '~/utils';
+
+import './SearchPanel.css';
+
+type Props = {
+  options: SearchState;
+  applyOptions: () => void;
+  toggleOption: (value: string, filter: SearchType) => void;
```
mathjazz, line 14: There are still a few instances of `filter` and `SearchFilter` in this file. Let's use searchOptions.

mathjazz (commented):
Nice work! Just a few cosmetic changes.

We should test `SearchPanel`, but let's do that when adding the next search options, so that we aren't stuck here for too long.

mathjazz (commented):
Good job!

We should also disable the feature on backend. Right now the default (and with the search panel disabled only possible) behavior is to not search among the string identifiers.

mathjazz (approved):
Thanks for keeping up with the tedious review process. :)

Well done!

PR #3293, opened 2024-08-01
Use default Gunicorn worker timeout setting

mathjazz:
The Heroku router drops a long-running request after 30 seconds, but the dyno behind it continues processing the request until completion.

Adding a timeout to the webserver ensures that the dyno itself drops the long-running request, creating capacity for other requests.

A while ago the webserver timeout was set to 120 seconds, in order to potentially help with long request like downloading TMX files via curl. It's not clear if this was actually helpful, but it increases a chance of Applcation Errors under heavy load.


mathjazz:
It would be nice to deploy this while Cloud Translation API is timing out to see if the number of [H12](https://devcenter.heroku.com/articles/request-timeout) errors decreases. 

We might need to reduce the timeout even below 30 seconds, but let's first test the experience and potential reduction in Application Errors with the default setting.

eemeli (approved):

PR #3294, opened 2024-08-02
Fix failing test case in contributors/tests/test_utils.py

harmitgoswami:
This PR fixes the existing bug with `test_get_contribution_timeline_data_with_actions` that occurs on the first day of the month (e.g. August 1st, September 1st, etc).

The `test_get_contribution_timeline_data_with_actions` test case inside of `contributors/tests/test_utils.py` was failing on the first of the first day of each month since actions are grouped by the month they are created in. The `yesterdays_action_user_a` variable was setting the `action.created_at` to be the last day of the previous month, so the contribution timeline data was returning an empty object.

I've also manually set the start date of the query to be midnight, to ensure that all actions within a month are captured, regardless of the time they were created.

### DIFF
--- a/pontoon/contributors/tests/test_utils.py
+++ b/pontoon/contributors/tests/test_utils.py
@@ -83,12 +83,18 @@ def action_user_b(translation_a, user_b):
 
 @pytest.fixture
 def yesterdays_action_user_a(translation_a, user_a):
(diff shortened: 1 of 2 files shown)


mathjazz (approved):
Nice catch!

PR #3295, opened 2024-08-07
Add specs for transactional emails

bcolsson:


specs/0119-transactional-emails.md
```diff
@@ -0,0 +1,428 @@
...
+
+### Inactive account: Contributor
+
+#### Trigger
+
+All of the following conditions met:
+
+1. Contributor role   
+2. At least 1 non-ML translation submitted   
+3. No log in for 6 months
```
flodolo, line 279:
```suggestion
3. No login for 6 months  
```

flodolo (approved):

specs/0119-transactional-emails.md
```diff
@@ -0,0 +1,428 @@
...
+
+* Header: “Monthly activity summary”  
+  * Located under “Email” section  
+  * Contributors are opted-out by default, translators/team managers are opted-in by default in order to reduce the amount of emails being sent.  
+* Checkbox (default: see below):  “Receive monthly activity summary email”  
+  * Each user model has a flag that corresponds to this checkbox. The flag is referenced by the transactional email trigger to determine if an email is sent. This flag is disabled by default for users except translators and managers.  
+  * Flag is enabled for all current users with global translator or manager permissions at time of implementation.  
+  * Flag is automatically enabled for a user who receives new global translator or manager permissions. If a user's translator or manager permissions are removed, the user would have to manually disable these emails from their settings.
+
+## Transactional emails
```
mathjazz, line 51:
```suggestion
## New transactional emails
```

(to make a distinction against the following section called "Existing Transactional Emails")

specs/0119-transactional-emails.md
```diff
@@ -0,0 +1,428 @@
...
+
+In addition, this document contains a number of transactional emails to be sent to Pontoon users. The motivation for these are based on an [experiment](https://docs.google.com/document/d/1MY4H4jlxgYDmIsu8CW5q8VN0wo_nSWVCtRSdy119EGA/edit?usp=sharing) and work done in 2023 around transactional emails being sent to inactive users — these emails encouraged re-engagement of some lapsed/dormant users.
+
+Finally, some additional transactional emails are intended to complement upcoming initiatives for improving the onboarding experience for new contributors.
+
+# Feature explanation
+
+## Settings changes
+
+### Sending comments as email
```
mathjazz, line 21:
```suggestion
### Sending notifications as email
```

specs/0119-transactional-emails.md
```diff
@@ -0,0 +1,428 @@
...
+Once
+
+### Onboarding: Week 1 - more localization resources
+
+#### Trigger
+
+When both conditions are met: 
+
+1. At least 1 week after account creation  
+2. 5 or more translation suggestions submitted
```
mathjazz, line 239: I assume we should only check these two conditions once? Otherwise we could end up sending this "final introductory email" years after the account creation.

specs/0119-transactional-emails.md
```diff
@@ -0,0 +1,428 @@
...
+
+* [Community roles](https://mozilla-l10n.github.io/localizer-documentation/community/l10n_community_roles.html)  
+* Specific [Mozilla product](https://mozilla-l10n.github.io/localizer-documentation/products/index.html) localization (such as Firefox, Firefox for Android, etc.)  
+* Tools such [Pontoon](https://mozilla-l10n.github.io/localizer-documentation/tools/pontoon/index.html) and [Fluent](https://mozilla-l10n.github.io/localizer-documentation/tools/fluent/index.html) (the advanced translation format used by Mozilla)
+
+**Pontoon Add-on:**   
+You can use the Pontoon Add-on for [Firefox](https://mozilla-l10n.github.io/localizer-documentation/index.html) or [Chrome](https://chromewebstore.google.com/detail/pontoon-add-on/gnbfbnpjncpghhjmmhklfhcglbopagbb) to be instantly notified of any Pontoon notifications right in your browser, as well as some helpful tools to help with proofreading, bug reporting, and fixing. 
+
+**Mozilla community participation guidelines:**
```
mathjazz, line 261: This is already mentioned in the welcome email.

specs/0119-transactional-emails.md
```diff
@@ -0,0 +1,428 @@
...
+
+**Pontoon Add-on:**   
+You can use the Pontoon Add-on for [Firefox](https://mozilla-l10n.github.io/localizer-documentation/index.html) or [Chrome](https://chromewebstore.google.com/detail/pontoon-add-on/gnbfbnpjncpghhjmmhklfhcglbopagbb) to be instantly notified of any Pontoon notifications right in your browser, as well as some helpful tools to help with proofreading, bug reporting, and fixing. 
+
+**Mozilla community participation guidelines:**  
+As an active member of our community, we’d like you to be aware of the [Mozilla community participation guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/) to ensure the wellbeing of all members of the community.
+
+**Email settings:**  
+Finally, if you’re interested in hearing about upcoming events organized for our localization community, be sure to enable email communications in your [settings](https://pontoon.mozilla.org/settings/) so you can be the first to get notified.
```
mathjazz, line 265:
```suggestion
Finally, if you’re interested in latest updates, announcements about new Pontoon features, invitations to contributor events and more, be sure to enable email communications in your [settings](https://pontoon.mozilla.org/settings/) so you can be the first to get notified.
```

Since only people that contribute in the first N days will get this email, should we move this section to one of the previous intro emails?

specs/0119-transactional-emails.md
```diff
@@ -0,0 +1,428 @@
...
+
+This updates the “Notification subscriptions” section to include:
+
+* 2 columns with checkboxes for each notification type.  
+* Column 1 label: “Enable”  
+  * This maps to the current functionality — enabling this will enable notifications in Pontoon.  
+  * Specs / defaults should be preserved from current implementation.  
+* Column 2 label (default value: disabled): “Include in email digest”  
+  * When enabled, this type of notification will be included in an email sent to the user’s designated email address (either their login account, or the email address they define in the email preferences section.)  
+* Dropdown list (default value: Daily): “Email digest frequency”
```
mathjazz, line 33: Is Daily the right default here? I assume most of our active localizers don't localize every day, and daily emails will be much more annoying to them than weekly (which are also sent by SUMO).

specs/0119-transactional-emails.md
bcolsson: Yes, but still worthwhile to include in this since it's 1: important and 2: this is sort of a final collection of useful resources.

specs/0119-transactional-emails.md
bcolsson: My intention is to send this to anybody who makes 5 or more suggestions, but sent 1 week or later after account creation (whenever a user reaches 5 suggestions). With that basis, I'd like to leave this in the final email since the user has a higher chance of being serious about contributing versus someone who just signed up.

I think that's a good callout that I should match the legal approved messaging around the opt-in.

specs/0119-transactional-emails.md
mathjazz: OK, so essentially we should change line 236 to "When either condition is met: "?

bcolsson: Summarizing our discussion offline: with my initial conditions there's a cost associated with needing to check every user that registers to see if they have made 5 translation suggestion. 

Reviewing the spec again I'd like to suggest some changes - I'd like to change the number of translations from 5 to 1. (The reasoning behind this is that the Inactive account: Contributor email triggers with just 1 translation). Then for how to handle the sending so its at least 1 week away from account creation:

I'd like to propose two alternatives:
Alternative 1:
When user makes their first translation, trigger an email to be sent 1 week from date of trigger.

Alternative 2:
Check once a week, starting 1 week from account creation date until 6 months from account creation date. (We consider an account inactive at 6 months, so we can assume they are unlikely to be a contributor at this point.)

I'm fine with either approach, but would like your input @mathjazz  on which is more realistic from a technical standpoint.

specs/0119-transactional-emails.md
bcolsson: That's a good point. I was thinking the main use case here would be to get more timely notifications, but it makes sense to be more conservative with the default choice and making the less noisy version as default here.

specs/0119-transactional-emails.md
```diff
@@ -0,0 +1,428 @@
...
+Body:  
+Hello! 
+
+It's been a while since we’ve last seen you, but we wanted to thank you for contributing to Mozilla’s community-driven localization efforts at [https://pontoon.mozilla.org/](https://pontoon.mozilla.org/). 
+
+Your contributions have been important in furthering Mozilla’s mission and fostering an open and accessible internet for all readers in your language. We understand that life can get busy, and finding time to contribute to projects isn't always easy. We get it. 
+
+As part of our efforts to keep our localization community active and engaged, we periodically review account activity. We’ve noticed you haven’t signed in to your Pontoon account in the past 6 months. While we completely understand that circumstances change, we wanted to inform you that we may deactivate dormant accounts in the future to ensure the security and efficiency of our platform.
+
+If you’d like to keep your Pontoon account active (we hope you will!) please sign in to Pontoon within the next 30 days by visiting [https://pontoon.mozilla.org/](https://pontoon.mozilla.org/) and click the “Sign in” button in the top right corner.
```
flodolo, line 294: I kind of missed this (@Delphine raised it in a separate conversation). Since we're not going to disable accounts, and the sentence before is "we may deactivate", should we remove the `within the next 30 days`?

bcolsson: Good point, we can always update this if we implement deactivation.

specs/0119-transactional-emails.md
bcolsson: Okay, let's just go with the simple implementation. I'll adjust the text accordingly.

mathjazz (approved):

PR #3299, opened 2024-08-16
Update: added migration script for migration of historic data

### LINKED ISSUE #2196: Migrate historic data into ActionLog table
## [Bug 1599387](https://bugzilla.mozilla.org/show_bug.cgi?id=1599387)
Bug Reporter: @adngdb
CC:   @mathjazz

We intend to use the ActionLog table to fuel a bunch of features, such as latest activity, various timelines (user, project, team… ) and so on. In order not to lose data, we want to migrate as much data as we can from our existing tables into the ActionLog table.

Most of the data we want to migrate is in the Translation table, and that's the review actions data (approved date, approved user, and so on).

Matjaz, is there anything else we want to migrate?

RafaelJohn9:
Fix  #2196

It migrates data from the previous years up to `(2020/1/ 7,  9:25:11:829125)`


pontoon/actionlog/migrations/0004_auto_20240816_1045.py
```diff
@@ -0,0 +1,48 @@
...
+
+    for translation in Translation.objects.filter(date__lt=end_date):
+        for attr, action_type in traslation_info.items():
+            if getattr(translation, attr) is not None:
+                actions_to_log.append(
+                    ActionLog(
+                        action_type=action_type,
+                        created_at=translation.date,
+                        entity_id=translation.entity_id,
+                        locale_id=translation.locale_id,
```
mathjazz, line 35: We should remove lines for these two fields. They are only used [when a translation has been deleted or a team comment has been added](https://github.com/mozilla/pontoon/blob/4fe3b6bd8d1cbbeae34e5cbbb8d582a2cedef35f/pontoon/actionlog/models.py#L36), neither of which is relevant for this migration.

pontoon/actionlog/migrations/0004_auto_20240816_1045.py
```diff
@@ -0,0 +1,48 @@
...
+                )
+    ActionLog.objects.bulk_create(actions_to_log)
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("actionlog", "0003_existing_pretranslation_action"),
+    ]
+
+    operations = []
```
mathjazz, line 48: We should call the function from here, as well as the reverse migration ([RunPython.noop](https://docs.djangoproject.com/en/5.1/ref/migration-operations/#django.db.migrations.operations.RunPython.noop) should be fine).

pontoon/actionlog/migrations/0004_auto_20240816_1045.py
```diff
@@ -0,0 +1,48 @@
...
+    # date to end migration
+    end_date = datetime.datetime(
+        2020, 1, 7, 9, 25, 11, 829125, tzinfo=datetime.timezone.utc
+    )
+
+    actions_to_log = []
+
+    for translation in Translation.objects.filter(date__lt=end_date):
+        for attr, action_type in traslation_info.items():
+            if getattr(translation, attr) is not None:
```
mathjazz, line 29: We should also check that any date is before `end_date` in order to avoid duplicate entries. It could be that translation was created before we started using ActionLog, but it could have been approved afterwards.

mathjazz (commented):
Nice work!

Left a few comments below.

pontoon/actionlog/migrations/0004_auto_20240816_1045.py
RafaelJohn9: researched on this and saw that `django.db.models.Q` is a good way to do this
You'll see it in my latest update
Let me know what you think about it

RafaelJohn9:
Hello @mathjazz  I was trying a way not to hardcode the `actiontypes` but right now they are just leading to Attribute errors. Is there a way to achieve this or should I just hardcode the `action types`

pontoon/actionlog/migrations/0004_auto_20240816_1045.py
```diff
@@ -24,15 +25,21 @@ def migrate_translation_to_actionlog(apps, schema_editor):
...
+        | Q(unrejected_date__lt=end_date)
+    )
+
+    for translation in translations:
         for attr, action_type in traslation_info.items():
             if getattr(translation, attr) is not None:
                 actions_to_log.append(
                     ActionLog(
                         action_type=action_type,
                         created_at=translation.date,
```
RafaelJohn9, line 42: Also I noticed that since this is actionlog the creation date is not neccessarilly the creation of the translation date but it varies with the  action type:
- approved date
- unapproved date
- rejected date
- unrejected date
- creation date

Should I also change this to match that?

pontoon/actionlog/migrations/0004_auto_20240816_1045.py
mathjazz: If the `Translation.date` is before `end_date` and `Translation.approved_date` isn't, we'll still create the (duplicate) `ActionLog` item.

It should be sufficient to check `Translation.date` only in the QuerySet (as we did before) and then make sure that `getattr(translation, attr)` is before `end_date`.

RafaelJohn9:
Hello @mathjazz  made the  changes that you recommended, Looking forward to your review

pontoon/actionlog/migrations/0004_auto_20240816_1045.py
```diff
@@ -0,0 +1,54 @@
...
+import datetime
+
+from django.db import migrations
+
+
+def migrate_translation_to_actionlog(apps, schema_editor):
+    Translation = apps.get_model("base", "Translation")
+    ActionLog = apps.get_model("actionlog", "ActionLog")
+
+    traslation_info = {
```
mathjazz, line 12:
```suggestion
    translation_info = {
```

pontoon/actionlog/migrations/0004_auto_20240816_1045.py
```diff
@@ -0,0 +1,54 @@
...
+        2020, 1, 7, 9, 25, 11, 829125, tzinfo=datetime.timezone.utc
+    )
+
+    actions_to_log = []
+
+    for translation in Translation.objects.filter(date__lt=end_date):
+        for attr, action_type in traslation_info.items():
+            if (
+                getattr(translation, attr) is not None
+                and translation["attr"] < end_date
```
mathjazz, line 31: I'd expect this line to raise a `TypeError`. Doesn't it?

Since we already get the value with `getattr(translation, attr)`, I'd move that call before the `if` clause:

```python
value = getattr(translation, attr)
if value is not None and value < end_date:
    ...
```

RafaelJohn9: on it

RafaelJohn9: > I'd expect this line to raise a `TypeError`. Doesn't it?
> 
> Since we already get the value with `getattr(translation, attr)`, I'd move that call before the `if` clause:
> 
> ```python
> value = getattr(translation, attr)
> if value is not None and value < end_date:
>     ...
> ```

thanks on this :)

RafaelJohn9 (commented):
``` 
   traslation_info = {
    translation_info = {
    ```
Would've gone unnotice for centuries and beyond  😅 thanks for that... Vs code auto giving variables made me miss that, might use vim next time 😅

File: pontoon/actionlog/migrations/0004_auto_20240816_1045.py [hunk 7]
```diff
@@ -0,0 +1,52 @@
...
+    actions_to_log = []
+
+    for translation in Translation.objects.filter(date__lt=end_date):
+        for attr, action_type in translation_info.items():
+            value = getattr(translation, attr)
+            if value is not None and value < end_date:
+                actions_to_log.append(
+                    ActionLog(
+                        action_type=action_type,
+                        created_at=translation["attr"],
```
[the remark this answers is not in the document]
RafaelJohn9: yea missed that too

pontoon/actionlog/migrations/0004_auto_20240816_1045.py
```diff
@@ -0,0 +1,52 @@
...
+
+    for translation in Translation.objects.filter(date__lt=end_date):
+        for attr, action_type in translation_info.items():
+            value = getattr(translation, attr)
+            if value is not None and value < end_date:
+                actions_to_log.append(
+                    ActionLog(
+                        action_type=action_type,
+                        created_at=value,
+                        performed_by_id=translation.user_id,
```
mathjazz, line 35: Sorry, I missed one more thing - `performed_by` should be a different value for different values of `attr`:
* `date`: `translation.user`
* `approved_date`: `translation.approved_user`
* `unapproved_date`: `translation.unapproved_user`
* `rejected_date`: `translation.rejected_user`
* `unrejected_date`: `translation.unrejected_user`

Additionally, we should only store and entry to the `ActionLog` if the said user value is not `None`.

RafaelJohn9:
Fix changes you requested :)

pontoon/actionlog/migrations/0004_auto_20240816_1045.py
```diff
@@ -0,0 +1,68 @@
...
+
+    # date to end migration
+    end_date = datetime.datetime(
+        2020, 1, 7, 9, 25, 11, 829125, tzinfo=datetime.timezone.utc
+    )
+
+    actions_to_log = []
+
+    for translation in Translation.objects.filter(date__lt=end_date):
+        for attr, action_user in translation_info.items():
```
mathjazz, line 40: I think this will be much clearer:
```suggestion
        for attr, (action_type, action_user) in translation_info.items():
```

pontoon/actionlog/migrations/0004_auto_20240816_1045.py
```diff
@@ -0,0 +1,68 @@
...
+            value = getattr(translation, attr)
```
mathjazz, line 41: Now that we use `getattr()` twice, we should change the name of the variable to something more specific, e.g.:

```suggestion
            date = getattr(translation, attr)
```

mathjazz (commented):
OK, last two comments, then we're ready to test this with real data.

RafaelJohn9:
`done`  👍

pontoon/actionlog/migrations/0004_auto_20240816_1045.py
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    # date to end migration
+    end_date = datetime.datetime(
+        2020, 1, 7, 9, 25, 11, 829125, tzinfo=datetime.timezone.utc
+    )
+
+    actions_to_log = []
+
+    for translation in Translation.objects.filter(date__lt=end_date):
```
mathjazz, line 39: To improve performance, we can already here exclude translations that don't have any user set, i.e. for which `user`, `approved_user`, `unapproved_user`, `rejected_user` and `unrejected_user` are `None`.

On production instance that reduces the amount of relevant Translation objects by ~65%, because a lot of old translations were imported from VCS and don't have any attribution data associated.

pontoon/actionlog/migrations/0004_auto_20240816_1045.py
```diff
@@ -0,0 +1,68 @@
...
+        for attr, (action_type, action_user) in translation_info.items():
+            date = getattr(translation, attr)
```
mathjazz, line 41: One more edge case we should cover:

If `approved_date` value is the same as the `date` value, we shouldn't store the `translation:approved` action to `ActionLog`. That means translation was directly submitted as approved, in which case we only log `translation:created` action.

mathjazz (commented):
I tested the code on stage and it turns out it was too slow to actually complete.

Inline I provided an idea how to improve performance and another edge case that we need to take into account.

RafaelJohn9:
Fixed them 👍

mathjazz (commented):
We'll need to optimize the code further, because it still doesn't complete. I have some ideas and I'll do a bit of testing on stage first, so we don't run in circles for too long.

pontoon/actionlog/migrations/0004_auto_20240816_1045.py
```diff
@@ -0,0 +1,81 @@
...
+        rejected_user=None,
+        unrejected_user=None,
+    )
+
+    for translation in translations:
+        for attr, (action_type, action_user) in translation_info.items():
+            date = getattr(translation, attr)
+            user_id = getattr(translation, action_user)
+
+            # Skip logging 'translation:approved' if approved_date is the same as the date
```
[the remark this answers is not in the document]
RafaelJohn9: Thank you for the feedback. I agree that comments should be more focused on explaining the reasoning behind the code rather than just describing its functionality.

I’ll update the comments to clarify why we handle logging in this way. It should make the code’s intent clearer and provide better context for future readers.

Thanks again for pointing this out!

RafaelJohn9:
> We'll need to optimize the code further, because it still doesn't complete. I have some ideas and I'll do a bit of testing on stage first, so we don't run in circles for too long.

I was thinking a potential reason why it may be slow is due to creating the objects in bulk at the end instead of single processing each? Though this is just a theory, I'll have to prove it first, then I get back to you

flodolo:
> We'll need to optimize the code further, because it still doesn't complete.

I'm curious: what is killing the migration? Is it a setting on Heroku, a setting in Django? Can it be overridden?

mathjazz:
> I'm curious: what is killing the migration? Is it a setting on Heroku, a setting in Django? Can it be overridden?

Could be a setting, but I don't think this is the right approach, at least not right now.

It took at least 30 minutes before the process was killed, and I comment out the `bulk_create()` action. Such long migrations are risky and most likely require app downtime. 

We should try to make the code faster first.

RafaelJohn9:
1. Filtering takes $$O(n)$$
2. Going through translations:
   - Each translation requires an inner loop running 5 times due to (translation_info). Let's call it $$m$$
   - Worst case going through all translations $$O(n)$$
   so $$O(nm)$$

Total is  $$n + nm$$ times

What if we cut it down to  one loop?

pontoon/actionlog/migrations/0004_auto_20240816_1045.py
```diff
@@ -0,0 +1,81 @@
...
+            ("translation:rejected", "Translation rejected"),
+            "rejected_user",
+        ),
+        "unrejected_date": (
+            ("translation:unrejected", "Translation unrejected"),
+            "unrejected_user",
+        ),
+    }
+
+    # date to end migration
```
mathjazz, line 32:
```suggestion
    # Timestamp of the first Translation created after the introduction of ActionLog
```

pontoon/actionlog/migrations/0004_auto_20240816_1045.py
```diff
@@ -0,0 +1,81 @@
...
+    actions_to_log = []
+
+    # Exclude translations that don't have any user set to improve performance
+    translations = Translation.objects.filter(date__lt=end_date).exclude(
+        user=None,
+        approved_user=None,
+        unapproved_user=None,
+        rejected_user=None,
+        unrejected_user=None,
+    )
```
mathjazz, line 46: This seems to improve performance by at least 90%:

```suggestion
    translations = (
        Translation.objects.filter(date__lt=end_date)
        .exclude(
            user=None,
            approved_user=None,
            unapproved_user=None,
            rejected_user=None,
            unrejected_user=None,
        )
        .values(
            "id",
            "date",
            "approved_date",
            "unapproved_date",
            "rejected_date",
            "unrejected_date",
            "user",
            "approved_user",
            "unapproved_user",
            "rejected_user",
            "unrejected_user",
        )
    )
```

mathjazz: Note that you'll need to make adjustments to the code below to make it work.

RafaelJohn9: aight :handshake:  on it

mathjazz: Sorry, I already went ahead and pushed the changes.

mathjazz: I deployed to our stage server and the migration completed fine! 👍

I'll just check now if the app somewhere relies on the ActionLog entries to be created in chronological order. If not, we can just merge this.

RafaelJohn9: okay, if you need me just let me know :) :+1:

mathjazz:
@RafaelJohn9 r? https://github.com/mozilla/pontoon/pull/3299/commits/7ac05b9eca910665231788f0bdcc5e82c14d5bbc

Without this change, `create_at` was always set to "now", i.e. when the migration was executed.

RafaelJohn9 (commented):
Yea this fixes it  :+1:

RafaelJohn9 (commented):
r+

PR #3306, opened 2024-08-28
Allow IP ranges in BLOCKED_IPS setting

### LINKED ISSUE #3304: Extend BLOCKED_IPS functionality to block a range of IPs
Right now we're [checking for a perfect match](https://github.com/mozilla/pontoon/blob/aafdf527f1adf7266f34e8a6bb231df355e6ca7d/pontoon/base/middleware.py#L37), it would be useful to be able to block a range of IPs.

flodolo:
IP Ranges can be defined with the CIDR notation (e.g. 192.168.1.0/24).

Fixes #3304

### DIFF
--- a/docs/admin/deployment.rst
+++ b/docs/admin/deployment.rst
@@ -73,9 +73,11 @@ you create:
    Set to 'google' if you want to use 'Google' (corresponding GOOGLE_* settings must be set).
 
 ``BLOCKED_IPS``
-   A comma-separated list of IP addresses or IP ranges (expressed using the CIDR
-   notation, e.g. `192.168.1.0/24`) to be blocked from accessing the app, for
+   A comma-separated list of IP addresses or IP ranges (expressed using the
+   `CIDR notation`_, e.g. `192.168.1.0/24`) to be blocked from accessing the app, for
    example because they are DDoS'ing the server.
+
+   .. _CIDR notation: https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing#CIDR_notation
 
 ``CELERY_ALWAYS_EAGER``
    Controls whether asynchronous tasks (mainly used during sync) are sent to
--- a/docs/admin/deployment.rst
+++ b/docs/admin/deployment.rst
@@ -73,8 +73,9 @@ you create:
    Set to 'google' if you want to use 'Google' (corresponding GOOGLE_* settings must be set).
 
 ``BLOCKED_IPS``
-   A comma-separated list of IP addresses to be blocked from accessing the app,
-   for example because they are DDoS'ing the server.
+   A comma-separated list of IP addresses or IP ranges (expressed using the CIDR
+   notation, e.g. `192.168.1.0/24`) to be blocked from accessing the app, for
+   example because they are DDoS'ing the server.
 
 ``CELERY_ALWAYS_EAGER``
    Controls whether asynchronous tasks (mainly used during sync) are sent to
--- a/pontoon/base/middleware.py
+++ b/pontoon/base/middleware.py
@@ -1,6 +1,6 @@
 import logging
 
-from ipaddress import ip_address, ip_network
+from ipaddress import ip_address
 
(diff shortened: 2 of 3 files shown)


flodolo:
Can't test this locally at the moment, but adding a few notes:
* I simplified the creation of BLOCKED_IP_RANGES in `base.py`. Technically, you could do it with a set of `try…except`, but I'm not sure it's worth it (also considering performances).
* Added logging in case one of the ranges (i.e. IPs that included `/`) is invalid. Not sure if I did it correctly, or if that's unnecessary.

flodolo:
On second thoughts, I'm going to move the logic to `base.py` in a separate commit, in case it's easy to back out.

My assumption is that the code in base.py is called less frequently, and it would be good to have it a bit more robust. Not sure how much that's true.

pontoon/settings/base.py
```diff
@@ -277,8 +280,27 @@ def _default_from_email():
...
-# A list of IP addresses to be blocked from accessing the app, because they are DDoS'ing the server
-BLOCKED_IPS = os.environ.get("BLOCKED_IPS", "").split(",")
+# A list of IP addresses or IP ranges to be blocked from accessing the app,
+# as a DDoS countermeasure.
+blocked_ip_settings = os.environ.get("BLOCKED_IPS", "").split(",")
+BLOCKED_IPS = []
+BLOCKED_IP_RANGES = []
+for ip in blocked_ip_settings:
+    ip = ip.strip()
+    if ip == "":
```
flodolo, line 290: This wasn't obvious to me, but noticed the log always showed the warning with an empty IP. 

Turns out splitting an empty string results in a list with an empty string 🤷🏼 

```
>>> "".split(",")
['']
```

I could check before running the for cycle, but maybe it's useful to keep it here, for example if the list ends with a comma.

```
"a,b,".split(",")
['a', 'b', '']
```

Also noticed that we weren't stripping the list item, so added that as well.

pontoon/base/middleware.py
```diff
@@ -38,6 +42,19 @@ def process_request(self, request):
         if ip in settings.BLOCKED_IPS:
             return HttpResponseForbidden("<h1>Forbidden</h1>")
 
+        # Convert request IP string to an IP object, check if it belongs to an
+        # IP range
+        try:
+            ip_obj = ip_address(ip)
+        except ValueError:
+            log = logging.getLogger(__name__)
+            log.error(f"Invalid IP detected in BlockedIpMiddleware: {ip}")
```
mathjazz, line 51: Why do we want to log this?

In case an invalid IP address is reported, what can we do about it?

docs/admin/deployment.rst
```diff
@@ -73,8 +73,9 @@ you create:
    Set to 'google' if you want to use 'Google' (corresponding GOOGLE_* settings must be set).
 
 ``BLOCKED_IPS``
-   A comma-separated list of IP addresses to be blocked from accessing the app,
-   for example because they are DDoS'ing the server.
+   A comma-separated list of IP addresses or IP ranges (expressed using the CIDR
+   notation, e.g. `192.168.1.0/24`) to be blocked from accessing the app, for
```
mathjazz, line 77: A link to CIDR notation article on e.g. Wikipedia would be nice.

mathjazz (commented):
Looks good. Left a minor suggestion and a question.

pontoon/base/middleware.py
flodolo: I think there is value in knowing that the resulting IP is not valid, and we might need to update the code for edge cases. Right now, we wouldn't even know they exist.

```
        try:
            ip = request.META["HTTP_X_FORWARDED_FOR"]
            # If comma-separated list of IPs, take just the last one
            # http://stackoverflow.com/a/18517550
            ip = ip.split(",")[-1]
        except KeyError:
            ip = request.META["REMOTE_ADDR"]
```

mathjazz (approved):

PR #3307, opened 2024-08-28
Add specification for Translation Memory Management

mathjazz:
What the title says.


specs/0121-translation-memory-management.md
```diff
@@ -0,0 +1,38 @@
...
+
+TM entries can be searched by the source string and translation using the Search field. Search is triggered by pressing Enter or 500 ms after the user has stopped typing.
+
+The following actions are available to team managers and translators and disabled for other users:
+
+1. **Edit**. After clicking the Edit button in the Actions column, the translation of the TM entry transforms into a textarea element and becomes editable. Next to it, Save and Cancel buttons appear. Clicking Save saves any changes and clicking Cancel stops editing.
+
+1. **Delete**. After clicking the Delete button in the Actions column, the button changes into the Are you sure? button. Clicking on that button deletes the TM entry.
+
+1. **Upload**. After clicking the Upload button above the TM entry list, an Open file dialog pops up. The user then selects the TMX file to upload and afterwards a success or error message appears.
```
bcolsson, line 32: I assume the `Upload TMX` button is for adding new TM entries, but that's not explained in the spec. Also do we need to explain this in the UI somewhere? For example, what happens if someone downloads the whole TM, edits some of the TMs, then attempts to upload the edited version (with mostly duplicate entries)?

bcolsson:
Should we consider a way to add a new entry without uploading a .tmx file?

bcolsson:
Do we need a way to identify strings with the exact same text (and potentially same translation) - for example which project the string is associated with?

specs/0121-translation-memory-management.md
mathjazz: > I assume the `Upload TMX` button is for adding new TM entries, but that's not explained in the spec. 

Added some clarity to the definition.

> Also do we need to explain this in the UI somewhere? 

I think that's clear? We also don't do anything special in the UI for uploading translations in the translation workspace.

> For example, what happens if someone downloads the whole TM, edits some of the TMs, then attempts to upload the edited version (with mostly duplicate entries)?

I wouldn't exclude duplicates, because the number of matches in TM is also a data point. But I was thinking about flagging uploaded TM entries as such, so that we can delete them easily if something goes south.

specs/0121-translation-memory-management.md
```diff
@@ -0,0 +1,43 @@
...
+Translation memory is a fundamental tool for faster and more consistent translations, and has become even more important with pretranslation, since it's used to train custom machine translation engines in Google AutoML Translation.
+
+Support for TM in Pontoon is limited:
+* Localizers cannot edit or delete existing TM entries directly.
+* TM entries can be changed indirectly via translation workspace, but only if the corresponding strings are still available for translation.
+* Localizers cannot upload their own translation memories.
+
+# Feature explanation
+
+A new "Translation Memory" tab is added to the Team pages, which lists all TM entries for the team. For each entry, source string and translation are provided, clicking on which leads to the translation workspace for the corresponding string. Entries are listed in alphabetic order.
```
eemeli, line 20: Why not use the existing translation view for this? Why not list "Translation Memory" among the projects for the Team, and use a slightly modified translate view for it?

mathjazz:
> Do we need a way to identify strings with the exact same text (and potentially same translation) - for example which project the string is associated with?

To some extent, sorting the strings in alphabetic order should help with that. Anything beyond that I would keep for a follow-up, so I added it to the Out of scope section.

> Should we consider a way to add a new entry without uploading a .tmx file?

IIRC the conversation from Dublin, we didn't find a good reason why somebody would want to do this? If the string is in Pontoon, why wouldn't you just add a translation for it and as such also a TM entry? If it's not in Pontoon, why do you need a TM entry? Uploading .TMX files is different, because a large set of TM entries has a match higher chance of affecting partial matches.

My preference would be to keep this out of scope, because there's a workaround (just create a .TMX file and upload it). If there's an actual demand from users, we can always add it later.

specs/0121-translation-memory-management.md
mathjazz: I was thinking about this initially, because it would match the behavior of https://pontoon.mozilla.org/projects/terminology/. We could even store uploaded .tmx files as files for that project.

But I worry about the backend complexity and speed. Unlike Terminology, TM entries are constantly added, so we'd need to sync between the actual TM and the TM Project a lot. The fact that original strings in Terminology are the same for all locales, and different for each locale for TM also doesn't help.

eemeli: To clarify, I was not suggesting having a separate TM Project, but to write the changes into TM the same way that they would be written in the current proposal.

Using the same front-end as normal projects shouldn't mean that the back-end needs to also match.

specs/0121-translation-memory-management.md
```diff
@@ -0,0 +1,43 @@
...
+1. **Edit**. After clicking the Edit button in the Actions column, the translation of the TM entry transforms into a textarea element and becomes editable. Next to it, Save and Cancel buttons appear. Clicking Save saves any changes and clicking Cancel stops editing.
+
+1. **Delete**. After clicking the Delete button in the Actions column, the button changes into the Are you sure? button. Clicking on that button deletes the TM entry.
+
+1. **Upload**. After clicking the Upload button above the TM entry list, an Open file dialog pops up. The user then selects the TMX file to upload, the TM entries from the file start to import into the team's TM, and eventually a success or error message appears. Uploaded strings have `is_personal` flag said to `True` in the DB.
+
+# Out of scope
+
+* Ability to batch delete uploaded TM entries from the "Translation Memory" tab.
+* Ability to show all TM entries with the matching source string.
```
flodolo, line 37: This feels like a major limitation, and raises other questions: we store multiple translation units for a determined source string. For example, I have 5 translations for `Register`, each of them with a different `tuid`.

```
<tuv xml:lang="en-US">
  <seg>Register</seg>
</tuv>
```

What are we showing if the view only has source and target, and there are multiple term units with the same source?

specs/0121-translation-memory-management.md
flodolo, line 20: Who has access to this tab? Is there a log of the changes made somewhere?

mathjazz: The tab is public.

We cal log changes in the ActionLog.

specs/0121-translation-memory-management.md
mathjazz: We show each TM entry separately, i.e. what https://pontoon.mozilla.org/a/base/translationmemoryentry/ shows. 

For example, there are 71 TM entries for "Bookmarks" translated as "Segnalibri" into Italian. That's 71 lines in the TM tab table.

specs/0121-translation-memory-management.md
```diff
@@ -0,0 +1,43 @@
...
+
+TM entries can be searched by the source string and translation using the Search field. Search is triggered by pressing Enter or 500 ms after the user has stopped typing.
+
+The following actions are available to team managers and translators and disabled for other users:
+
+1. **Edit**. After clicking the Edit button in the Actions column, the translation of the TM entry transforms into a textarea element and becomes editable. Next to it, Save and Cancel buttons appear. Clicking Save saves any changes and clicking Cancel stops editing.
+
+1. **Delete**. After clicking the Delete button in the Actions column, the button changes into the Are you sure? button. Clicking on that button deletes the TM entry.
+
+1. **Upload**. After clicking the Upload button above the TM entry list, an Open file dialog pops up. The user then selects the TMX file to upload, the TM entries from the file start to import into the team's TM, and eventually a success or error message appears. Uploaded strings have `is_personal` flag said to `True` in the DB.
```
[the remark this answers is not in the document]
mathjazz: If we don't make a distinction between Pontoon TM entries and personal TM entries (uploaded or even manually added), it's impossible to remove them.

Size limit: TMX yes, but I'm not sure about custom TM.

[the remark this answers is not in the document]
mathjazz: I've made some changes to the spec that make the `is_personal` flag redundant by introducing the new action type.

specs/0121-translation-memory-management.md
mathjazz: Changed that in the spec. Also realized we have the `?list=` query parameter. 🤦

specs/0121-translation-memory-management.md
```diff
@@ -0,0 +1,43 @@
...
+
+1. **Delete**. After clicking the Delete button in the Actions column, the button changes into the Are you sure? button. Clicking on that button deletes the TM entry.
+
+1. **Upload**. After clicking the Upload button above the TM entry list, an Open file dialog pops up. The user then selects the TMX file to upload, the TM entries from the file start to import into the team's TM, and eventually a success or error message appears. A newly introduced `TM_ENTRIES_UPLOADED` action is logged in the `ActionLog` in case of success, referencing all newly created TM entries. Based on the performance of the upload function, a maximum file upload size will be imposed.
+
+# Out of scope
+
+* Ability to batch delete uploaded TM entries from the "Translation Memory" tab.
+* Ability to update corresponding localizable translations when the TM entry is edited.
```
mathjazz, line 37: Wanted to call out that currently, if a TM entry is updated, any related translations stay the same.

specs/0121-translation-memory-management.md
```diff
@@ -0,0 +1,43 @@
...
+
+Initially, only the first 100 entries are loaded. Additional entries are loaded in batches of 100 using infinite scroll, a UX pattern that is also used in the string list in translation workspace.
+
+TM entries can be searched by the source string and translation using the Search field. Search is triggered by pressing Enter or 500 ms after the user has stopped typing.
+
+The following actions are available to team managers and translators and disabled for other users:
+
+1. **Edit**. After clicking the Edit button in the Actions column, the translation of the TM entry transforms into a textarea element and becomes editable. Next to it, Save and Cancel buttons appear. Clicking Save saves any changes and logs the newly introduced `TM_ENTRY_EDITED` action in the `ActionLog`, referencing the edited TM entry. Clicking Cancel stops editing.
+
+1. **Delete**. After clicking the Delete button in the Actions column, the button changes into the Are you sure? button. Clicking on that button deletes the TM entry.
```
mathjazz, line 30: Note the Delete action is special in that it is not logged. That's because there is no `TMEntry` object left after it is deleted, and because I believe "soft delete" (set `is_deleted` flag rather than actually delete the object) is even worse (because it makes queries more complex and the table more bloated).

The alternative would be to store `Entity` and `Locale` as ForeignKeys.

specs/0121-translation-memory-management.md
```diff
@@ -0,0 +1,43 @@
...
+Translation memory is a fundamental tool for faster and more consistent translations, and has become even more important with pretranslation, since it's used to train custom machine translation engines in Google AutoML Translation.
+
+Support for TM in Pontoon is limited:
+* Localizers cannot edit or delete existing TM entries directly.
+* TM entries can be changed indirectly via translation workspace, but only if the corresponding strings are still available for translation.
+* Localizers cannot upload their own translation memories.
+
+# Feature explanation
+
+A new "Translation Memory" tab is added to the Team pages, which lists all TM entries for the team. For each entry, source string and translation are provided, clicking on which leads to the translation workspace for the corresponding strings (using the list `query` parameter). Entries are listed in alphabetic order and only shown once for each unique combination of source string and translation.
```
mathjazz, line 20: Should the entries be listed in alphabetic order of source strings (easier to find similar strings) or in the chronological order of translations (easier to fix recent mistakes)?

bcolsson: Personal opinion: alphabetical doesn't seem useful considering the number of entries there will be. If I know which string I'm looking for I'd be using the search function instead of scrolling. At least with chronological there might be some utility.

specs/0121-translation-memory-management.md
```diff
@@ -0,0 +1,43 @@
...
+
+Initially, only the first 100 entries are loaded. Additional entries are loaded in batches of 100 using infinite scroll, a UX pattern that is also used in the string list in translation workspace.
+
+TM entries can be searched by the source string and translation using the Search field. Search is triggered by pressing Enter or 500 ms after the user has stopped typing.
+
+The following actions are available to team managers and translators and disabled for other users:
```
bcolsson, line 26:
```suggestion
The following actions are available to team managers and translators for their locales and disabled for other users:
```

specs/0121-translation-memory-management.md
flodolo, line 20: Going back to visibility, I'm concerned about resources and DDoS here (given we just experienced one), also considering the infinite scrolling UX.

Would it make sense to make it visible only to people signed in?

flodolo: +1 on reverse chronological order

specs/0121-translation-memory-management.md
```diff
@@ -0,0 +1,43 @@
...
+* TM entries can be changed indirectly via translation workspace, but only if the corresponding strings are still available for translation.
+* Localizers cannot upload their own translation memories.
+
+# Feature explanation
+
+A new "Translation Memory" tab is added to the Team pages, which lists all TM entries for the team. For each entry, source string and translation are provided, clicking on which leads to the translation workspace for the corresponding strings (using the list `query` parameter). Entries are listed in alphabetic order and only shown once for each unique combination of source string and translation.
+
+Initially, only the first 100 entries are loaded. Additional entries are loaded in batches of 100 using infinite scroll, a UX pattern that is also used in the string list in translation workspace.
+
+TM entries can be searched by the source string and translation using the Search field. Search is triggered by pressing Enter or 500 ms after the user has stopped typing.
```
flodolo, line 24: Why the 500ms? I think it doesn't match the UX in the rest of Pontoon, and it makes sense to react only to enter.

specs/0121-translation-memory-management.md
mathjazz: Actually it will be hard to sort by the chronological order when items are grouped.

mathjazz: Listing TM entries is much faster than querying TM for fuzzy matches (which is available to all users). 

I have the feature as presented in the initial version of the spec implemented locally and it's even (and quite a bit) faster than the Projects tab (which is also public), probably because there aren't any JOINs.

That said, I expect the view to get slower when we start grouping by unique combinations of source strings and translations. I wouldn't want to speculate in the spec though whether making the view private will be needed to mitigate that.

specs/0121-translation-memory-management.md
mathjazz: True, it doesn't match the rest of Pontoon, but requiring Enter also doesn't. The search field in the Projects tab looks exactly the same and searches as you type. 

We can't have that in TM, because it needs to be server side due to the much bigger amount of data.

Should I just drop the 500 ms part and file a bug acknowledging the inconsistency?

flodolo: While they're closer, I think the comparison shouldn't be with the project tab, but with the search in Translate view. I'm OK with dropping and aknowledging.

flodolo:
One thing that is missing is what happens when you edit or delete with entries that are grouped together:
* Delete: I assume we remove all of them
* Edit: edit all of them?

The latter brings up some doubts on how we store translation memories in the first place, but not sure if that's in scope for this or should be a separate conversation (it would definitely impact this).

specs/0121-translation-memory-management.md
mathjazz: OK, I'll drop the 500 ms part.

(The reason why I'm bringing up the Project tab is because the widgets have the exact same designs. For a good UX, it's important that widgets that look the same also behave the same to avoid confusion and maintain usability.)

specs/0121-translation-memory-management.md
mathjazz: Dropped the mention of the order.

mathjazz:
> One thing that is missing is what happens when you edit or delete with entries that are grouped together:
> 
>     * Delete: I assume we remove all of them
>     * Edit: edit all of them?

Made it clearer in the spec the actions apply to all entries.

eemeli (approved):

bcolsson (approved):

specs/0121-translation-memory-management.md
```diff
@@ -0,0 +1,43 @@
...
+
+1. **Edit**. After clicking the Edit button in the Actions column, the translation of the TM entry transforms into a textarea element and becomes editable. Next to it, Save and Cancel buttons appear. Clicking Save saves any changes to all corresponding TM entries and logs the newly introduced `TM_ENTRY_EDITED` action in the `ActionLog`, referencing the edited TM entry. Clicking Cancel cancels the editing action.
+
+1. **Delete**. After clicking the Delete button in the Actions column, the button changes into the Are you sure? button. Clicking on that button deletes all corresponding TM entries.
+
+1. **Upload**. After clicking the Upload button above the TM entry list, an Open file dialog pops up. The user then selects the TMX file to upload, the TM entries from the file start to import into the team's TM, and eventually a success or error message appears. A newly introduced `TM_ENTRIES_UPLOADED` action is logged in the `ActionLog` in case of success, referencing all newly created TM entries. Based on the performance of the upload function, a maximum file upload size will be imposed.
```
flodolo, line 32: For the last paragraph, it's unclear what it means: will we add a limit if performances are bad, or will we decide the limit value based on performances?

flodolo (approved):

specs/0121-translation-memory-management.md
mathjazz: Updated.

mathjazz:
I've made [this change](https://github.com/mozilla/pontoon/pull/3307/commits/5e27cb82cc274ceeeec25ff73acff605020d765b) after the PR was approved, which I've previously mentioned [in the comment](https://github.com/mozilla/pontoon/pull/3307#discussion_r1736962188). That way all 3 actions are logged, and we at least have some data to track changes for the Delete action.

Since it's a minor change, I'll go ahead and merge, but please let me know if you disagree with the change.

PR #3314, opened 2024-09-03
Add ability to filter message recipients by role, project and locale

mathjazz:
This is the first half of fixing #3244 as per the [spec](https://github.com/mozilla/pontoon/blob/main/specs/0118-messaging-center.md#selecting-recipients-with-filters).

The remaining bit are Activity filters, as part of which the Composing UI will be finalized and adjusted.


pontoon/messaging/static/js/messaging.js
```diff
@@ -16,10 +16,24 @@ $(function () {
 
   const container = $('#main .container');
 
-  function isValidForm($form, isNotification, isEmail, subject, body) {
+  function isValidForm() {
```
eemeli, line 19: This should have a name like `validateForm()` so that it's clearer that it has side effects.

pontoon/projects/static/css/multiple_project_selector.css
eemeli, line 1: Nesting...

pontoon/projects/static/js/multiple_project_selector.js
```diff
@@ -0,0 +1,56 @@
...
+   */
+  function updateSelectedProjects() {
+    const $selectedList = $('.multiple-project-selector .project.selected'),
+      $selectedProjectsField = $selectedList.find('input[type=hidden]'),
+      selectedProjects = $selectedList
+        .find('li[data-id]')
+        .map(function () {
+          return $(this).data('id');
+        })
+        .get();
```
eemeli, line 14: Not really sure why we persist with jQuery; in usage like this it's not really any better than the pure JS alternative:
```js
const listItems = document.querySelectorAll('.multiple-project-selector .project.selected li[data-id]');
const selectedProjects = Array.from(listItems, li => li.dataset['id']);
```

This isn't intended as a comment on just this specific place, but overall we should not be making any later transition away from jQuery progressively harder

pontoon/projects/static/js/multiple_project_selector.js
```diff
@@ -0,0 +1,56 @@
...
+
+    $selectedProjectsField.val(selectedProjects.join());
+  }
+
+  // Choose projects
+  $('body').on(
+    'click',
+    '.multiple-project-selector .project.select li',
```
eemeli, line 22: This too doesn't actually need jQuery, but if it's jQuery, it should not be any more complex than necessary.

Same applies to handler below as well.

```suggestion
  $('.multiple-project-selector .project.select li').on(
    'click',
```

eemeli (changes requested):
See above. Sorry, misclicked and meant to file them as review comments rather than one by one.

pontoon/projects/static/js/multiple_project_selector.js
mathjazz: `multiple_project_selector` widget is reusing HTML, JS and CSS code from the `multiple_locale_selector` widget. Should I refactor that into a generic `multiple_item_selector` widget?

eemeli: I think we should have a discussion about what we're doing with the non-translate parts of the UI. I'm not really concerned about this or any other specific instance, but more the overall tech stack that we're using, and whether we might want to move towards less jQuery use.

pontoon/projects/static/js/multiple_project_selector.js
mathjazz: I believe this will break where the selected element is not in the DOM yet when the handler is attached, e.g. in dashboard tabs.

pontoon/projects/static/js/multiple_project_selector.js
mathjazz: Filed https://github.com/mozilla/pontoon/issues/3315.

mathjazz:
I've made two changes:
* [Implemented user role filter on backend](https://github.com/mozilla/pontoon/pull/3314/commits/ae297d0acb0c35e3b6be35b98d72bee7b0a61017)
* [Added generic multiple item selector widget and used it for filtering by project](https://github.com/mozilla/pontoon/pull/3314/commits/0631f973c6fb2997691a29a06681abb8926513b9)

As part of the latter, I've used CSS nesting. I didn't change the JS/jQuery logic, as this is blocked on #3315.

bcolsson (approved):

eemeli (approved):

PR #3316, opened 2024-09-04
Implement Search Filter for searching within translations only

### LINKED ISSUE #3222: Search: add options to search only translations
By default, Pontoon searches both source strings and translations. On top of the work in #3221 (to add the search options panel), we should provide an option to `Search in translations only`. When selected, Pontoon will ignore source strings when searching.

harmitgoswami:
Fix #3222 

Added an option to limit a string search within the `translate` view to only translations (i.e. exclude matches within source strings). Extends the search panel feature from #3288.

**To reproduce results**: Head to any string within the translate view upon running the server. Then, click on the magnifying glass icon on the right side of the search box. Click on the `Search in translations only` option to toggle the filter, then click the `APPLY SEARCH OPTIONS` button. The results shown on the left column will now only include search matches within translations, excluding all matches within source strings.

In the below example, the first image shows `http://localhost:8000/fr/facebook-container/messages.json/?search=web&string=198559`, where the source string contains 'web' but the translation does not. In the second image, after the filter is applied, this string is no longer displayed in the search results.

image

image


mathjazz (commented):
Excellent work!

Could we make the orange/yellow highlight visible only in translations when the option is selected?

translate/src/modules/entitieslist/components/Entity.tsx
```diff
@@ -135,6 +135,7 @@ export function Entity({
             content={entity.original}
             format={entity.format}
             search={parameters.search}
```
mathjazz, line 137: I think if you do this, you don't need to use `translations_only` and pass it further:

```suggestion
            search={!parameters.search_translations_only && parameters.search}
```

mathjazz: Everything else looks and works great!

mathjazz (approved):
Well done!

PR #3317, opened 2024-09-06
Implement filter for searching in rejected translations

### LINKED ISSUE #3226: Search: add option to include rejected translations
By default, Pontoon searches all strings, independently of their status. This means that also rejected translations are included.

Pontoon search should exclude rejected translations by default. On top of the work in #3221 (to add the search options panel), we should provide an option to `Search in rejected translations`, which would include also rejected translations.

harmitgoswami:
Fix #3226 

Added a filter to expand a user's search to include rejected translations. Currently, by default, rejected translations are included in the QuerySet returned after a search. This feature will remove that default behaviour to only search in approved translations, and if the filter is set, then rejected translations are included in the QuerySet as well.

**To reproduce results**: 

Head to any string within the translate view upon running the server. Ensure that the string has an approved translation, as well as a rejected translation. Find the difference between the two strings, then enter that within the Search Box input. The left side column will show no results, as long as the searched string does not appear in any other **approved** translation.

Then, click on the magnifying glass icon on the right side of the search box. Click on the `Search in rejected translations` option to toggle the filter, then click the `APPLY SEARCH OPTIONS` button. The results shown on the left column will now include the rejected translation.

Screenshot 2024-09-06 at 8 45 33 AM
Screenshot 2024-09-06 at 8 45 49 AM

### DIFF
--- a/pontoon/base/models/entity.py
+++ b/pontoon/base/models/entity.py
@@ -733,6 +733,7 @@ class Entity(DirtyFieldsMixin, models.Model):
         extra=None,
         search_identifiers=None,
         search_translations_only=None,
+        search_rejected_translations=None,
         time=None,
         author=None,
         review_time=None,
@@ -845,9 +846,13 @@ class Entity(DirtyFieldsMixin, models.Model):
         if search:
             search_list = utils.get_search_phrases(search)
 
+            q_rejected = (
+                Q()
+            )  # if search_rejected_translations else Q(translation__approved=True)
             translation_filters = (
                 Q(translation__string__icontains_collate=(search, locale.db_collation))
                 & Q(translation__locale=locale)
+                & q_rejected
                 for search in search_list
             )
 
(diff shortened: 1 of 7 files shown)


pontoon/base/models/entity.py
```diff
@@ -845,9 +846,13 @@ def for_project_locale(
         if search:
             search_list = utils.get_search_phrases(search)
 
+            q_rejected = (
+                Q()
+            )  # if search_rejected_translations else Q(translation__approved=True)
```
mathjazz, line 851: That's not a completely accurate condition, because we should also search in suggestions, so:
```suggestion
            )  # if search_rejected_translations else Q(translation__rejected=False)
```

pontoon/base/models/entity.py
```diff
@@ -845,9 +846,13 @@ def for_project_locale(
         if search:
```
mathjazz, line 846: I'd also add a TODO here with a bullet point for each of the (currently 3) search options that are disabled ATM and instructions on how to enable them. That would also supersede the TODO in lines 867-868.

mathjazz (commented):
Great work! Almost there!

mathjazz (approved):
Brilliant!

PR #3318, opened 2024-09-06
Implement search filter for matching case

### LINKED ISSUE #3224: Search: add option to match case
By default, Pontoon search is case insensitive. On top of the work in #3221 (to add the search options panel), we should provide an option to `Match case`, which would only return matches with the same capitalization as the search terms.

harmitgoswami:
Fix #3224 

Added a search filter for matching exact casing. Currently, by default, searches are case insensitive (e.g. `Foo` and `foo` would both appear when searching for `Foo`). This feature retains the default behaviour, but adds the option to strictly match casing between the search term and matches.

**To reproduce results**:

Head to any string within the translate view upon running the server. Ensure that the string appears at least twice in the string list (left-side column), once with an uppercase letter and once with all lowercase letters.

Then, click on the magnifying glass icon on the right side of the search box. Click on the `Match case` option to toggle the filter, then click the APPLY SEARCH OPTIONS button. The results shown on the left column will now only include instances of the search term that match the casing you inputted. Try changing the casing of your search term, and observe how the results on the string list change.

Screenshot 2024-09-06 at 4 51 06 PM
Screenshot 2024-09-06 at 4 51 29 PM
Screenshot 2024-09-06 at 4 51 43 PM

### DIFF
--- a/pontoon/base/models/entity.py
+++ b/pontoon/base/models/entity.py
@@ -843,10 +843,6 @@ class Entity(DirtyFieldsMixin, models.Model):
                 # only tag needs `distinct` as it traverses m2m fields
                 entities = entities.distinct()
 
-        # TODO: Uncomment the following lines to reactivate the
-        #       context identifiers filter once .ftl bug is fixed (issue #3284):
-        #       - 883-888
-
         # Filter by search parameters
         if search:
             search_list = utils.get_search_phrases(search)
@@ -856,16 +852,19 @@ class Entity(DirtyFieldsMixin, models.Model):
                 Q() if search_rejected_translations else Q(translation__rejected=False)
             )
 
+            # Modify query based on case sensitivity filter
+            translation_case_lookup = (
+                "contains" if search_match_case else "icontains_collate"
+            )
+
             translation_filters = (
-                (
-                    Q(translation__string__contains=(search, locale.db_collation))
-                    if search_match_case
-                    else Q(
-                        translation__string__icontains_collate=(
+                Q(
+                    **{
+                        f"translation__string__{translation_case_lookup}": (
                             search,
                             locale.db_collation,
                         )
-                    )
+                    }
                 )
                 & Q(translation__locale=locale)
                 & q_rejected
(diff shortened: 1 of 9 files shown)


translate/src/modules/search/components/SearchPanel.tsx
```diff
@@ -11,7 +11,6 @@ import './SearchPanel.css';
 // TODO: Remove the variable below to reactivate the feature once
 //       all search options are implemented
 // Disable SearchPanel component until fully complete
-const disable: Boolean = true;
```
mathjazz, line 14: We should also remove the comment.

pontoon/base/models/entity.py
```diff
@@ -843,20 +844,29 @@ def for_project_locale(
                 entities = entities.distinct()
 
         # TODO: Uncomment the following lines to reactivate the
-        #       feature once all search options are implemented:
-        #       - 857 (rejected translations)
-        #       - 869-870 (context identifiers)
-        #       - 883-884 (translations only)
+        #       context identifiers filter once .ftl bug is fixed (issue #3284):
+        #       - 883-888
```
mathjazz, line 848: Now it's probably best if you move this comment where these lines are, otherwise you'll need to keep the line numbers accurate. :)

pontoon/base/models/entity.py
```diff
@@ -866,22 +876,39 @@ def for_project_locale(
...
+                entity_filters = (
+                    (
+                        Q(string__contains=search)
+                        | Q(string_plural__contains=search)
+                        | q_key
+                        if search_match_case
+                        else Q(string__icontains=search)
+                        | Q(string_plural__icontains=search)
+                        | q_key
+                    )
```
mathjazz, line 899: Same as above, we shouldn't repeat all this code.

pontoon/base/models/entity.py
```diff
@@ -866,22 +875,36 @@ def for_project_locale(
...
-            q_key = Q(key__icontains=search)  # if search_identifiers else Q()
-            entity_filters = (
-                Q(string__icontains=search) | Q(string_plural__icontains=search) | q_key
-                for search in search_list
-            )
+            # Search in source strings
+            if not search_translations_only:
+                # Search in string (context) identifiers
+
+                case_lookup = "contains" if search_match_case else "icontains"
```
mathjazz, line 882: Can we simply use the `translation_case_lookup` here (and rename it to `case_lookup`)?

mathjazz (commented):
Looking good! Resolve the inline comment and we're good to go!

pontoon/base/models/entity.py
harmitgoswami: Leaving in the `collate` and re-suing the same variable results in an error with `/get-entities/` for me :(

mathjazz: Is the error still there if you add the 2nd parameter, i.e. `locale.db_collation`?

mathjazz: OTOH, we have en-US hardcoded in a few places and collation is not needed for it, so let's keep it like this.

mathjazz (approved):

pontoon/base/models/entity.py
harmitgoswami: image
image
image


It seems I no longer get any results when adding in the second argument

mathjazz: (In this case, locale would need to be `Locale.objects.get(code="en-US")`).

PR #3319, opened 2024-09-09
Fix bug when Match Case and Translations Only filter were active

harmitgoswami:
Currently in production, if both `Match case` and `Translations only` are active Search Options, then no results will be displayed, no matter what the search term is

Before: https://pontoon.mozilla.org/fr/facebook-container/messages.json/?search=Facebook&search_translations_only=true&search_match_case=true

After: http://localhost:8000/fr/facebook-container/messages.json/?search=Facebook&search_translations_only=true&search_match_case=true

This PR fixes that bug, and also adjusts the placement of the Search Options panel to display under the magnifying glass icon.

The bug was fixed by removing the second parameter (`locale.db_collation`) from the query when `search_match_case` is set to true.

### DIFF
--- a/pontoon/base/models/entity.py
+++ b/pontoon/base/models/entity.py
@@ -864,7 +864,7 @@ class Entity(DirtyFieldsMixin, models.Model):
 
             translation_filters = (
                 q_match_case & Q(translation__locale=locale) & q_rejected
-                for search in search_list
+                for s in search_list
             )
 
             translation_matches = entities.filter(*translation_filters).values_list(
--- a/pontoon/base/models/entity.py
+++ b/pontoon/base/models/entity.py
@@ -853,22 +853,10 @@ class Entity(DirtyFieldsMixin, models.Model):
             )
 
             # Modify query based on case sensitivity filter
-
-            translation_case_lookup = (
-                "contains" if search_match_case else "icontains_collate"
-            )
-
             translation_filters = (
-                Q(
-                    **{
-                        f"translation__string__{translation_case_lookup}": (
-                            s,
-                            locale.db_collation,
-                        )
-                        if not search_match_case
-                        else s
-                    }
-                )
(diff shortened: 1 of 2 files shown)


pontoon/base/models/entity.py
```diff
@@ -853,21 +853,17 @@ def for_project_locale(
...
-                        f"translation__string__{translation_case_lookup}": (
-                            search,
-                            locale.db_collation,
-                        )
-                    }
-                )
-                & Q(translation__locale=locale)
-                & q_rejected
+                q_match_case & Q(translation__locale=locale) & q_rejected
                 for search in search_list
```
mathjazz, line 867: We should rename this variable, because `search` is already used:

```suggestion
                for s in search_list
```

That will also reveal the syntax error in the new code.

mathjazz: As mentioned, this will now break, because `s` is not declared.

pontoon/base/models/entity.py
```diff
@@ -861,14 +862,16 @@ def for_project_locale(
...
                     **{
                         f"translation__string__{translation_case_lookup}": (
-                            search,
+                            s,
                             locale.db_collation,
                         )
+                        if not search_match_case
+                        else s
                     }
                 )
```
mathjazz, line 871: At this point it's more readable if you drop `translation_case_lookup` and have just one `if - else` block with the same condition:

```python
Q(translation__string__contains=s) if search_match_case else
Q(translation__string__icontains_collate=(s, locale.db_collation))
```

pontoon/base/models/entity.py
```diff
@@ -853,6 +853,7 @@ def for_project_locale(
             )
 
             # Modify query based on case sensitivity filter
+
```
mathjazz, line 856: Please also remove this blank line.

mathjazz (approved):

Issue #3284, opened 2024-07-29
Exclude message identifiers when searching Fluent strings


mathjazz:
Fixing #3221 by dropping https://github.com/mozilla/pontoon/blob/2d5872e78b043c254ed06489ace91b565fad6e78/pontoon/base/models/entity.py#L858 is not going exclude searching Message IDs for Fluent strings, because they are also part of `Entity.string`.

We should change the data model accordingly to only search for the relevant bits of the message.

mathjazz:
Closed by mistake.

PR #3322, opened 2024-09-10
Fixed variable scoping in Match case query

### LINKED ISSUE #3221: Search: exclude message identifiers by default, add option to include them

### LINKED ISSUE #3321: Search with Match case enabled returns unrelated results
Example: https://pontoon.mozilla.org/it/firefox/browser/browser/aboutPrivateBrowsing.ftl/?search=Cookie&search_match_case=true&string=243111

harmitgoswami:
Fix #3321

Simplifying the query creation seemed to cause an issue with the casing Search Option. 

To fix the issue, I simply added a bracket around the conditional query in `translation_filters`, which corrects the variable scope and produces the correct QuerySet.

Before:
image

After (Note: the three unhighlighted results are from #3284):
image

### DIFF
--- a/pontoon/base/models/entity.py
+++ b/pontoon/base/models/entity.py
@@ -854,9 +854,13 @@ class Entity(DirtyFieldsMixin, models.Model):
 
             # Modify query based on case sensitivity filter
             translation_filters = (
-                Q(translation__string__contains=s)
-                if search_match_case
-                else Q(translation__string__icontains_collate=(s, locale.db_collation))
+                (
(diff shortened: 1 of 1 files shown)


mathjazz (approved):
LGTM

We should add unit tests as part of the next PR.

PR #3324, opened 2024-09-11
Replace pip and pip-tools with uv, update ruff

flodolo:


flodolo:
Meh 

> astral-sh/setup-uv@v2 is not allowed to be used in mozilla/pontoon.

flodolo:
Testing on my fork to see if this is valuable.

Issue #3323, opened 2024-09-11
Pontoon does not correctly differentiate between Turkish dotted and dotless "i"


harmitgoswami:
Currently, Pontoon doesn't differentiate between the Turkish 'ı' and 'i' (capital I and İ respectively), despite these being different characters.

For example, these two queries produce the exact same results (in addition to incorrect highlighting): 

https://pontoon.mozilla.org/tr/firefox/browser/browser/browser.ftl/?search=%C4%B1&string=246376
https://pontoon.mozilla.org/tr/firefox/browser/browser/browser.ftl/?search=i&string=246376

This bug has been bought up and addressed before: https://bugzilla.mozilla.org/show_bug.cgi?id=1346180

harmitgoswami:
It seems after some research that database collation is the correct and recommended way to go: http://www.i18nguy.com/unicode/turkish-i18n.html

However, even after reverting to our [previous approach](https://github.com/mozilla/pontoon/blob/6df98ce4796630bbf8b1f3fc0a4fc6371033ce0b/pontoon/base/models/entity.py#L859), I can confirm that Pontoon still doesn't detect the difference between the 'i' and 'ı' characters.

[Collation in Django](https://docs.djangoproject.com/en/5.1/ref/models/database-functions/#collate) does seem to be supported, but the way we invoke `entities.filter` and `entities.order_by` makes me think we'd need a pretty large refactor to properly use Django's `Collate` function.

mathjazz:
Nice catch - this seems to be fixed, indeed.

PR #3326, opened 2024-09-11
Implement filter for matching whole words

### LINKED ISSUE #3225: Search: add option to match whole words
By default, Pontoon search returns partial matches. For example, searching for "cookie" will also match "cookies".

On top of the work in #3221 (to add the search options panel), we should provide an option to `Match whole words`. This would be equivalent to a regular expression search, with the search terms surrounded by word boundary delimiter (`\b`). The search would remain case insensitive, as there is a separate option for that.

harmitgoswami:
Fix #3225 

Added a filter to only return search results that are an exact match to the search term (case insensitive). Currently, searing for "cookie" would also return results that contain "cookies". Now, if the filter is activated, only results that exactly match "cookie".

**To reproduce results:**

Head to any string within the translate view upon running the server. Ensure the string has both a singular and plural version present in the resource (e.g. cookie and cookies). Search for the singular version using the search box on the left side, and observe how both the singular and plural versions appear in the string list (left side column).

Then, click on the magnifying glass icon on the right side of the search box. Click on the `Match whole words` option to toggle the filter, then click the `APPLY SEARCH OPTIONS` button. Observe how the results in the string list now only include the singular version.  

image

image

### DIFF
--- a/pontoon/base/models/entity.py
+++ b/pontoon/base/models/entity.py
@@ -735,7 +735,7 @@ class Entity(DirtyFieldsMixin, models.Model):
         search_translations_only=None,
         search_rejected_translations=None,
         search_match_case=None,
-        search_match_word=None,
+        search_match_whole_word=None,
         time=None,
         author=None,
         review_time=None,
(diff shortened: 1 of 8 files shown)


pontoon/base/models/entity.py
```diff
@@ -855,10 +856,19 @@ def for_project_locale(
...
+                    if search_match_word
+                    else (
+                        Q(translation__string__contains=s)
+                        if search_match_case
+                        else Q(
+                            translation__string__icontains_collate=(
+                                s,
+                                locale.db_collation,
+                            )
+                        )
```
mathjazz, line 871: If you declared this before line 855:
```python
i = "" if search_match_case else "i"
y = "" if search_match_word else r"\y"
```
You could replace this whole block with this:
```python
Q(**{f"translation__string__{i}regex": rf"{y}{s}{y}"})
```

We can drop `icontains_collate`, because it doesn't work anyways apparently (#3323). And to fix that bug, we might want to implement regex_collate instead.

pontoon/base/models/entity.py
```diff
@@ -884,12 +894,28 @@ def for_project_locale(
...
+                        | q_key
+                        for s in search_list
+                    )
+                else:
+                    entity_filters = (
+                        Q(**{f"string__{case_lookup}": (s)})
+                        | Q(**{f"string_plural__{case_lookup}": (s)})
+                        | q_key
+                        for s in search_list
+                    )
```
mathjazz, line 918: See comment above for an idea how to simplify this.

mathjazz (commented):
Very nice work!

Left a couple of comments to simplify the code a little.

mathjazz (approved):
Excellent work!

In one of the next PRs, we should add unit tests for the search backend changes to test_entity.py.

PR #3331, opened 2024-09-12
Replace html2text with BeautifulSoup

### LINKED ISSUE #3330: Replace html2text package
I'm looking into replacing `html2text`, and decided to go with BeautifulSoup4 (a large package, but it's well maintained).

What I find confusing is the current behavior: we're not generating plain text, but markdown? That doesn't seem to make a lot of sense to me.

This is the output difference between the existing code and the new code (PR incoming) for a few basic examples.

-------------
#### Original
```

<p><strong>Foo</strong> bar, <em>foo1</em> bar1.</p>
<p>Another paragraph.</p>
<p>Yet Another <br><br> paragraph.</p>
```

#### Result from html2text
```
**Foo** bar, _foo1_ bar1.

Another paragraph.

Yet Another   
  
paragraph.

```

#### Result from BeautifulSoup
```

Foo bar, foo1 bar1.

Another paragraph.

Yet Another 

paragraph.

```
-------------
#### Original
```
<p><strong>Foo</strong> bar, <em>foo1</em> bar1. This is a very long text that shouldn't wrap. This is a very long text that shouldn't wrap. This is a very long text that shouldn't wrap.</p>
```

#### Result from html2text
```
**Foo** bar, _foo1_ bar1. This is a very long text that shouldn't wrap. This is a very long text that shouldn't wrap. This is a very long text that shouldn't wrap.

```

#### Result from BeautifulSoup
```
Foo bar, foo1 bar1. This is a very long text that shouldn't wrap. This is a very long text that shouldn't wrap. This is a very long text that shouldn't wrap.

```
-------------
#### Original
```
<div style='test'>Foo bar</div> something else.
```

#### Result from html2text
```
Foo bar

something else.

```

#### Result from BeautifulSoup
```
Foo bar
something else.
```
-------------
#### Original
```
<span class='foo'>Foo bar</span> something else.
```

#### Result from html2text
```
Foo bar something else.

```

#### Result from BeautifulSoup
```
Foo bar something else.
```
-------------
#### Original
```
<p>This is a <a href='https://mozilla.org'>test</a> link.</p>
```

#### Result from html2text
```
This is a [test](https://mozilla.org) link.

```

#### Result from BeautifulSoup
```
This is a test (https://mozilla.org) link.

```

flodolo:
Fixes #3330

### DIFF
--- a/pontoon/messaging/utils.py
+++ b/pontoon/messaging/utils.py
@@ -1,10 +1,20 @@
-import html2text
+from bs4 import BeautifulSoup
 
 
 def html_to_plain_text_with_links(html):
-    h = html2text.HTML2Text()
+    soup = BeautifulSoup(html, "html.parser")
 
-    h.wrap_links = False
-    h.body_width = 0
+    # Replace links with `link_text (url)`
+    for tag in soup.find_all("a"):
+        tag.replace_with(f"{tag.text} ({tag.get('href')})")
 
-    return h.handle(html)
+    # Add empty lines after <p>, </div>, and <br>. Not using the `separator`
+    # parameter, as it would apply to other inline tags like <span>.
+    for tag in soup.find_all(["p", "div", "br"]):
+        tag.insert_after("\n")
+
+    text = soup.get_text()
+    # Remove empty spaces at the beginning of lines
+    text = "\n".join([line.lstrip() for line in text.split("\n")])
+
+    return text
--- a/requirements/default.in
+++ b/requirements/default.in
@@ -15,6 +15,7 @@
 # Dependencies that do not come from pypi (eg. eggs from github) are listed at the end of the list.
 # -------------------------------------------------------------------------------------------------
 APScheduler==3.9.1.post1
+beautifulsoup4==4.12.3
 bleach==6.1.0
 celery==5.4.0
 compare-locales==9.0.4
@@ -33,7 +34,6 @@ django-pipeline==3.0.0
 google-cloud-translate==3.8.4
 graphene-django==3.2.1
 gunicorn==19.9.0
-html2text==2024.2.26
 jsonfield==3.1.0
 jsonschema==2.6.0
 lxml==4.9.1
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -32,6 +32,10 @@ asgiref==3.8.1 \
     # via
     #   django
     #   django-cors-headers
+beautifulsoup4==4.12.3 \
+    --hash=sha256:74e3d1928edc070d21748185c46e3fb33490f22f52a3addee9aee0f4f7781051 \
+    --hash=sha256:b80878c9f40111313e55da8ba20bdba06d8fa3969fc68304167741bbf9e082ed
+    # via -r requirements/default.in
 billiard==4.2.0 \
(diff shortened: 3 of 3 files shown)


mathjazz (commented):
Thanks for the patch!

I assume outputs in #3331 are generated using this configuration rather than default?

pontoon/messaging/utils.py
```diff
@@ -1,10 +1,20 @@
...
 
-    return h.handle(html)
+    # Add empty lines after <p>, </div>, and <br>. Not using the `separator`
+    # parameter, as it would apply to other inline tags like <span>.
+    for tag in soup.find_all(["p", "div", "br"]):
+        tag.insert_after("\n")
+
+    text = soup.get_text()
+    # Remove empty spaces at the beginning of lines
+    text = "\n".join([line.lstrip() for line in text.split("\n")])
```
mathjazz, line 18: Can you use the strip parameter for that?

https://www.crummy.com/software/BeautifulSoup/bs4/doc/#get-text

flodolo:
> I assume outputs in #3331 are generated using this configuration rather than default?

Yes, the output is generated with the existing code and this function

```python
import html2text
from bs4 import BeautifulSoup

strings = [
    """
<p><strong>Foo</strong> bar, <em>foo1</em> bar1.</p>
<p>Another paragraph.</p>
<p>Yet Another <br><br> paragraph.</p>""",
    "<p><strong>Foo</strong> bar, <em>foo1</em> bar1. This is a very long text that shouldn't wrap. This is a very long text that shouldn't wrap. This is a very long text that shouldn't wrap.</p>",
    "<div style='test'>Foo bar</div> something else.",
    "<span class='foo'>Foo bar</span> something else.",
    "<p>This is a <a href='https://mozilla.org'>test</a> link.</p>",
]

h = html2text.HTML2Text()
h.wrap_links = False
h.body_width = 0

def html_to_plain_text_with_links(html):
    soup = BeautifulSoup(html, "html.parser")

    # Replace links with `link_text (url)`
    for tag in soup.find_all("a"):
        tag.replace_with(f"{tag.text} ({tag.get('href')})")

    # Add empty lines after <p>, </div>, and <br>. Not using the `separator`
    # parameter, as it would apply to other inline tags like <span>.
    for tag in soup.find_all(["p", "div", "br"]):
        tag.insert_after("\n")

    text = soup.get_text()
    # Remove empty spaces at the beginning of lines
    text = "\n".join([l.lstrip() for l in text.split("\n")])

    return text

for s in strings:
    print("-------------")
    print(f"#### Original\n```\n{s}\n```\n")
    print(f"#### Result from html2text\n```\n{h.handle(s)}\n```\n")
    print(f"#### Result from BeautifulSoup\n```\n{html_to_plain_text_with_links(s)}\n```")
```

pontoon/messaging/utils.py
flodolo: It doesn't work as well. For example, `<div style='test'>Foo bar</div> something else.` becomes `Foo barsomething else.`, while with the current code is 

```
Foo bar
something else.
```

Without stripping, there would be a weird white space

```
Foo bar
 something else.
```

mathjazz (approved):

PR #3332, opened 2024-09-12
Make search panel visible on all screen widths

mathjazz:
The search panel could get cut-off on the left side on screen narrower than ~1400 px.

### DIFF
--- a/translate/src/modules/search/components/SearchPanel.css
+++ b/translate/src/modules/search/components/SearchPanel.css
@@ -26,8 +26,9 @@
     box-sizing: border-box;
     position: absolute;
     top: 44px;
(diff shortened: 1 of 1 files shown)


harmitgoswami:
It seems that the Search Options panel spans across the entire left column of the page, but the Filter panel leaves a slight gap on the right side.

Did we want to leave a gap for the Search Options panel too? I think it'd be more clear to users which UI element toggled the panel, on top of being more consistent with our existing design.

image

image

harmitgoswami:
In that case, I think it looks fine as it is currently!

harmitgoswami (approved):

PR #3336, opened 2024-09-12
Add regex to solve .ftl search bug

### LINKED ISSUE #3284: Exclude message identifiers when searching Fluent strings

harmitgoswami:
Fix #3284 

This PR adds additional regex to the `entity_filters` and `translation_filters` objects to solve the `.ftl` bug stemming from how Fluent translations are stored in the backend. As such, the `exclude_identifiers` filter has been re-enabled (See #3221).

**To reproduce results**: observe how for any fluent file, the Entity.string has the format of `{key} = {value}`. 

Now, [this query](http://localhost:8000/it/firefox/browser/branding/official/brand.ftl/?search=brand) will produce no results, since "brand" is only contained within the key of each Entity. Once the `exclude_identifiers` [filter is set](http://localhost:8000/it/firefox/browser/branding/official/brand.ftl/?search=brand&search_identifiers=true&string=192265), then all the keys containing "brand" will be displayed.

image

image

### DIFF
--- a/pontoon/base/models/entity.py
+++ b/pontoon/base/models/entity.py
@@ -858,7 +858,20 @@ class Entity(DirtyFieldsMixin, models.Model):
             y = r"\y" if search_match_whole_word else ""
 
             translation_filters = (
-                Q(**{f"translation__string__{i}regex": rf"=.*{y}{s}{y}.*"})
+                (
+                    Q(
+                        Q(resource__format="ftl")
+                        & (
+                            Q(**{f"translation__string__{i}regex": rf"=.*{y}{s}{y}.*"})
+                            if search_identifiers
+                            else Q(**{f"translation__string__{i}regex": rf"{y}{s}{y}"})
+                        )
+                    )
+                    | Q(
+                        ~Q(resource__format="ftl")
+                        & Q(**{f"translation__string__{i}regex": rf"{y}{s}{y}"})
+                    )
+                )
                 & Q(translation__locale=locale)
                 & q_rejected
                 for s in search_list
@@ -879,8 +892,30 @@ class Entity(DirtyFieldsMixin, models.Model):
                 )
 
                 entity_filters = (
-                    Q(**{f"string__{i}regex": rf"=.*{y}{s}{y}.*"})
-                    | Q(**{f"string_plural__{i}regex": rf"=.*{y}{s}{y}.*"})
+                    Q(
+                        Q(resource__format="ftl")
+                        & (
+                            Q(**{f"string__{i}regex": rf"=.*{y}{s}{y}.*"})
+                            if search_identifiers
+                            else Q(**{f"string_plural__{i}regex": rf"{y}{s}{y}"})
+                        )
+                    )
+                    | Q(
+                        ~Q(resource__format="ftl")
+                        & Q(**{f"string__{i}regex": rf"{y}{s}{y}"})
+                    )
+                    | Q(
+                        Q(resource__format="ftl")
+                        & (
+                            Q(**{f"string_plural__{i}regex": rf"=.*{y}{s}{y}.*"})
+                            if search_identifiers
+                            else Q(**{f"string_plural__{i}regex": rf"{y}{s}{y}"})
+                        )
+                    )
+                    | Q(
+                        ~Q(resource__format="ftl")
+                        & Q(**{f"string_plural__{i}regex": rf"{y}{s}{y}"})
+                    )
                     | q_key
                     for s in search_list
                 )
--- a/pontoon/base/models/entity.py
+++ b/pontoon/base/models/entity.py
@@ -892,6 +892,10 @@ class Entity(DirtyFieldsMixin, models.Model):
                         ~Q(resource__format="ftl")
                         & Q(**{f"string__{i}regex": rf"{y}{s}{y}"})
                     )
+                    | Q(
+                        ~Q(resource__format="ftl")
+                        & Q(**{f"string_plural__{i}regex": rf"{r}{y}{s}{y}{o}"})
+                    )
                     | (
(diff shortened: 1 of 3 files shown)


mathjazz:
Could you please check the failing tests?

pontoon/base/models/entity.py
```diff
@@ -858,7 +858,20 @@ def for_project_locale(
             y = r"\y" if search_match_whole_word else ""
```
mathjazz, line 858:
```suggestion
            y = r"\y" if search_match_whole_word else ""
            r = "" if search_identifiers else "=.*"
            o = "" if search_identifiers else ".*"
```

pontoon/base/models/entity.py
```diff
@@ -858,7 +858,20 @@ def for_project_locale(
...
 
             translation_filters = (
-                Q(**{f"translation__string__{i}regex": rf"{y}{s}{y}"})
+                (
+                    Q(
+                        Q(resource__format="ftl")
+                        & (
+                            Q(**{f"translation__string__{i}regex": rf"=.*{y}{s}{y}.*"})
+                            if not search_identifiers
+                            else Q(**{f"translation__string__{i}regex": rf"{y}{s}{y}"})
```
mathjazz, line 867:
```suggestion
                            Q(**{f"translation__string__{i}regex": rf"{r}{y}{s}{y}{o}"})
```

pontoon/base/models/entity.py
mathjazz: I don't have a better idea on what variable names to use, I went with p**r**efix and p**o**stfix. Which is lame.

pontoon/base/models/entity.py
mathjazz: We should do the same below for entities.

pontoon/base/models/entity.py
```diff
@@ -871,19 +884,38 @@ def for_project_locale(
...
 
                 entity_filters = (
-                    Q(**{f"string__{i}regex": rf"{y}{s}{y}"})
-                    | Q(**{f"string_plural__{i}regex": rf"{y}{s}{y}"})
+                    Q(
+                        Q(resource__format="ftl")
+                        & (
+                            Q(**{f"string__{i}regex": rf"=.*{y}{s}{y}.*"})
+                            if not search_identifiers
+                            else Q(**{f"string_plural__{i}regex": rf"{y}{s}{y}"})
```
mathjazz, line 900: Mind the "plural".

pontoon/base/models/entity.py
```diff
@@ -871,19 +884,38 @@ def for_project_locale(
...
+                        ~Q(resource__format="ftl")
+                        & Q(**{f"string__{i}regex": rf"{y}{s}{y}"})
+                    )
+                    | Q(
+                        Q(resource__format="ftl")
+                        & (
+                            Q(**{f"string_plural__{i}regex": rf"=.*{y}{s}{y}.*"})
+                            if not search_identifiers
+                            else Q(**{f"string_plural__{i}regex": rf"{y}{s}{y}"})
+                        )
```
mathjazz, line 913: `Entity.string_plural` is only used for Gettext (.po) strings (I know... lame... but we'll refactor that within the next 6 months), so we don't need this code.

mathjazz (commented):
Works great!

The only (minor) inconsistency I've found is that Match case also applies when searching for string identifiers, but Match whole word doesn't.

I would do two more things since this is the last search option:
* Move both "Match" options to the top of the list
* Add background to the search options on hover to match the behavior of filters

pontoon/base/models/entity.py
```diff
@@ -870,21 +883,20 @@ def for_project_locale(
...
-                    Q(**{f"string__{i}regex": rf"{y}{s}{y}"})
-                    | Q(**{f"string_plural__{i}regex": rf"{y}{s}{y}"})
-                    | q_key
+                    Q(
+                        Q(resource__format="ftl")
+                        & (Q(**{f"string__{i}regex": rf"{r}{y}{s}{y}{o}"}))
+                    )
+                    | Q(
+                        ~Q(resource__format="ftl")
+                        & Q(**{f"string__{i}regex": rf"{y}{s}{y}"})
```
mathjazz, line 893: For non-FTL strings (which include PO strings) we need to search in `string_plural`.

mathjazz (commented):
Almost there!

pontoon/base/models/entity.py
```diff
@@ -892,6 +892,10 @@ def for_project_locale(
                         ~Q(resource__format="ftl")
                         & Q(**{f"string__{i}regex": rf"{y}{s}{y}"})
                     )
+                    | Q(
+                        ~Q(resource__format="ftl")
+                        & Q(**{f"string_plural__{i}regex": rf"{r}{y}{s}{y}{o}"})
```
mathjazz, line 897: That doesn't seem right?
* This is not Fluent, so why do we need the prefix/postfix for ignoring context identifiers within the string?
* Why do we need a separate block instead of just adding line 897 after line 893?

harmitgoswami: Apologies about the first bullet point!

As for the second one, the test case seemed to fail when I tried to coalesce both queries into one block, but it turned out I was incorrectly ANDing both queries instead of ORRing them.

mathjazz (approved):
Great work!

This is it - Search24 improvements are now feature complete!

PR #3337, opened 2024-09-12
Update doc for virtualenv

flodolo:
Not sure if this should be removed, but the doc is currently wildly outdated.


mathjazz (approved):
This is definitely better than what we have.

But things we don't use quickly break or get out of date. So we should either remove it or start using it.

PR #3340, opened 2024-09-13
Add Review message panel

### LINKED ISSUE #3245: Add ability to review messages before sending them
In the Review phase, one can still go back and make changes.

There's also an option to "Send a test message to myself".

Spec:
https://github.com/mozilla/pontoon/blob/main/specs/0118-messaging-center.md#review

mathjazz:
Fix #3245.

WIP

* Add ability to switch between editing and review mode
* Add ability to send a test message to the current user
* Render message preview


mathjazz:
@bcolsson @eemeli Could you please take [this](https://mozilla-pontoon-staging.herokuapp.com/messaging) for a spin for a few moments? I'd appreciate your feedback by the end of the day if possible.

In addition to the Review panel, I've also added the ability to compose messages in Markdown format and "Send the message to myself" for testing purposes.

Please note that sending emails on stage is not possible, but you should be able to send a notification to yourself using the Send to myself button.

eemeli:
Initial thoughts:
- For low-ish locale & project counts, should list them all rather than something like "Locales: 2".
- Markdown passes through without turning into HTML in notifications.
- The explicit review step is good; that page should show the count of how many recipients it matches.

mathjazz:
> Initial thoughts:
> 
>     * For low-ish locale & project counts, should list them all rather than something like "Locales: 2".

Yeah. I also don't like the tooltip too much.

>     * Markdown passes through without turning into HTML in notifications.
>     * The explicit review step is good; that page should show the count of how many recipients it matches.

These to are covered in the above TODO list already.

bcolsson:
Can we apply different styling for the text in the editing review page? Doesn't need to necessarily be perfect, but a more WYSIWYG experience would be nice. Here's a screenshot of some testing:
Screenshot 2024-09-13 at 11 25 55

mathjazz:
> Can we apply different styling for the text in the editing review page? Doesn't need to necessarily be perfect, but a more WYSIWYG experience would be nice. Here's a screenshot of some testing:

Good point! Updated stage with styling changes for commonly used elements.

pontoon/messaging/static/css/messaging.css
```diff
@@ -109,4 +136,115 @@
...
+        }
+
+        ul {
+          overflow: inherit;
+          list-style: disc;
+        }
+
+        li {
+          color: var(--light-grey-6);
+        }
```
bcolsson, line 221:
```suggestion
        li {
          color: var(--light-grey-6);
          a {
            color: var(--status-translated);
            font-weight: inherit;
          }
        }
```

Let's use the same formatting for links in lists.

mathjazz: Pushing changes to stage. Now styling for links doesn't only apply to links within paragraphs, but in general.

bcolsson (approved):

eemeli (commented):
Getting rather odd results while experimenting with the "Send to me" function:
1. Open https://mozilla-pontoon-staging.herokuapp.com/messaging/
2. Select Message type: Notification, and enter Message subject `t1`, body `b`.
3. Select Review Message, then Send To Me.
4. Reload the page (Cmd+R)
5. Observe the notification dot for the `t1` message.
6. Re-select Message type: Notification, and edit Message subject to `t2`.
7. Select Review Message, then Send To Me.
8. Reload the page (Cmd+R)
9. Observe **no** notification dot for the `t2` message.
10. Re-select Message type: Notification, and edit Message subject to `t3`.
11. Select Review Message, then Send To Me.
12. Reload the page (Cmd+R)
13. Observe the notification dot for the `t3` message.
14. Select "See all Notifications".
15. Note that that `t2` message is not present.

This appears to work with further repetition cycles as well, with every second notification arriving.

pontoon/messaging/static/js/messaging.js
```diff
@@ -191,6 +162,40 @@ $(function () {
...
+    .click()
+    .end()
+    .find('.selected ul')
+    .scrollTop(0);
+  $('.multiple-item-selector')
+    .find('.available .move-all')
+    .click()
+    .end()
+    .find('.selected ul')
+    .scrollTop(0);
```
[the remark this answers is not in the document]
mathjazz: Fixed.

mathjazz:
Sorry, I made a mess locally and had to rebase. The only change since review is the last commit.

mathjazz:
Fixed the problem with Notification not always being sent and filed https://github.com/mozilla/pontoon/issues/3346. The problem was that the custom checkbox that we use didn't reflect the state of the actual (hidden) checkbox.

Updated version deployed to stage.

eemeli (approved):
Looks like it works now.

PR #3341, opened 2024-09-13
Add i18next format support to editFieldModes

flx-sta:
I've added highlighting support for the i18next interpolation format to the `common` mode of editFieldModes.

~~It's a quite extensive regex as there is quite some stuff you can do inside a i18next interpolation:~~
<sub>Thank you @eemeli for simplifying it</sub>

```
/{{.*?}}/
```
https://regex101.com/r/gokJ09/1

image

### DIFF
--- a/translate/src/modules/translationform/utils/editFieldModes.ts
+++ b/translate/src/modules/translationform/utils/editFieldModes.ts
@@ -88,12 +88,15 @@ const printf =
 
 const pythonFormat = /^{[\w.[\]]*(![rsa])?(:.*?)?}/;
 
+// https://www.i18next.com/translation-function/interpolation
+const i18nextFormat = /{{\s*[\w.[\]]*\s*(?:,\s*\w+(?:\([^)]*\))?)*\s*(?:![rsa])?(?::.*?)?\s*}}/;
+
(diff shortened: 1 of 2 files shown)


translate/src/modules/translationform/utils/editFieldModes.ts
```diff
@@ -88,12 +88,20 @@ const printf =
 
 const pythonFormat = /^{[\w.[\]]*(![rsa])?(:.*?)?}/;
 
+// https://www.i18next.com/translation-function/interpolation
+const i18nextFormat =
+  /{{\s*[\w.[\]]*\s*(?:,\s*\w+(?:\([^)]*\))?)*\s*(?:![rsa])?(?::.*?)?\s*}}/;
```
eemeli, line 93: This does seem a bit complex. Are there cases where something simpler like this would not work, or would produce misleading results?
```suggestion
const i18nextFormat = /{{.*?}}/;
```

flx-sta: I don't see why it should produce wrong ones.
Thanks for the suggestion.
Applied ✅ 

<sub>I guess using GPT made me silly and lazy that I didn't think of the most obvious one</sub>

eemeli (approved):

PR #3342, opened 2024-09-16
Fix Pontoon search for terms with special characters

### LINKED ISSUE #3333: Pontoon search fails for search terms with $
Example: searching for `moco_link` [finds one result](https://pontoon.mozilla.org/it/sumo/all-resources/?search=moco_link&string=222592), while searching for `$moco_link` [finds none](https://pontoon.mozilla.org/it/sumo/all-resources/?search=%24moco_link).

harmitgoswami:
Fix #3333 

Currently, Pontoon search fails if any search term contains a special character that's used in a regex pattern.

This PR surrounds all search terms with `re.escape`, which makes sure that any combination of special characters are included in the search term rather than affecting the preexisting regex pattern.

For example, this query will now work as intended: http://localhost:8000/it/sumo/all-resources/?search=%24moco_link&search_rejected_translations=true&string=222592

Additionally, the server error previously caused by this query is now fixed: http://localhost:8000/it/sumo/all-resources/?search=%28moco_link&string=222592

### DIFF
--- a/pontoon/base/models/entity.py
+++ b/pontoon/base/models/entity.py
@@ -1,6 +1,6 @@
 from functools import reduce
 from operator import ior
-from re import findall, match
+from re import escape, findall, match
 
 from dirtyfields import DirtyFieldsMixin
 from jsonfield import JSONField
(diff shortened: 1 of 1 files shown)


mathjazz:
Deployed to stage, but it seems like it doesn't work for the case reported in the original issue?

https://mozilla-pontoon-staging.herokuapp.com/it/sumo/all-resources/?search=moco_link
https://mozilla-pontoon-staging.herokuapp.com/it/sumo/all-resources/?search=%24moco_link

harmitgoswami:
Is there a translation containing `$moco_link` in that example? Since Flod's translation is pretty recent, it wasn't in my db file, so I had to manually add the translation containing `$moco_link` to test

mathjazz:
Ah, sorry, false alarm. Fixed the link by adding a suggestion. :)

mathjazz (approved):

PR #3349, opened 2024-09-17
Add ability to search for keys within translations only

### LINKED ISSUE #3344: String not found when searching for ID with both "search in string identifiers" and "search in translations only" enabled
[Example for Firefox](https://pontoon.mozilla.org/it/firefox/browser/browser/aboutPrivateBrowsing.ftl/?search=about-private-browsing-focus-promo-text&search_identifiers=true&search_translations_only=true&string=243111): searching for `about-private-browsing-focus-promo-text` shows 3 results

That doesn't seem to work for Firefox for Android: I'm searching for `mozac_browser_errorpages_redirect_loop_message`
* If I enable the search in string identifiers, it works ([1 result](https://pontoon.mozilla.org/it/firefox-for-android/all-resources/?search=mozac_browser_errorpages_redirect_loop_message&search_identifiers=true&string=196837)).
* If I enable also the `search in translations only` option, it doesn't return [any result](https://pontoon.mozilla.org/it/firefox-for-android/all-resources/?search=mozac_browser_errorpages_redirect_loop_message&search_identifiers=true&search_translations_only=true).

Is this a bug or am I missing something obvious?

harmitgoswami:
Fix #3344 

This PR adds the `search_identifers` query to `translation_filters`, effectively allowing users to search for keys regardless of if they include source strings in their search.

Before, toggling the `Search in translations only` filter would only search within the `Translation.string` field, meaning that a user could not exclude source strings and still search within both the `key` and `Translation.string` fields.

From the filed issue, these two queries now produce the same results, as intended:

http://localhost:8000//it/firefox-for-android/all-resources/?search=mozac_browser_errorpages_redirect_loop_message&search_identifiers=true&string=196837

http://localhost:8000/it/firefox-for-android/all-resources/?search=mozac_browser_errorpages_redirect_loop_message&search_identifiers=true&search_translations_only=true&string=196837

### DIFF
--- a/pontoon/base/models/entity.py
+++ b/pontoon/base/models/entity.py
@@ -889,7 +889,7 @@ class Entity(DirtyFieldsMixin, models.Model):
             # Search in source strings
             entity_filters = (
                 (
-                    Q(pk=None) # Ensures that no source strings are returned 
+                    Q(pk=None)  # Ensures that no source strings are returned
                     if search_translations_only
                     else (
                         Q(
--- a/pontoon/base/models/entity.py
+++ b/pontoon/base/models/entity.py
@@ -879,11 +879,6 @@ class Entity(DirtyFieldsMixin, models.Model):
                     if search_rejected_translations
                     else Q(translation__rejected=False)
                 )
-                | (
-                    Q(**{f"key__{i}regex": rf"{y}{escape(s)}{y}"})
-                    if search_identifiers
-                    else Q()
-                )
                 for s in search_list
             )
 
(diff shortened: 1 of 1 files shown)


pontoon/base/models/entity.py
```diff
@@ -879,7 +874,16 @@ def for_project_locale(
...
+                & (
+                    Q()
+                    if search_rejected_translations
+                    else Q(translation__rejected=False)
+                )
+                | (
+                    Q(**{f"key__{i}regex": rf"{y}{escape(s)}{y}"})
+                    if search_identifiers
+                    else Q()
+                )
```
mathjazz, line 886: I'm curious why did you include this in `translation_filters` rather than `entity_filters`?

harmitgoswami: I thought that all `Translations` also have a `key` field, but upon checking the admin panel, it seems I was wrong!

I think the reason why this works is because if we leave out the `search_identifiers` query from `translations_filters`, and both `search_identifiers` and `search_translations_only` are set to true, then `entity_filters` is never reached, so the key lookup fails.

Would it be better to instead remove the `if not search_translations_only` statement on line 895 and refactor accordingly?

mathjazz: Removing that check would be preferable... Unless it's a massive refactor, in which case I'd just leave it as it is and add a comment. It's possible Eemeli will need to change this code quite a lot as part of the data model unification.

pontoon/base/models/entity.py
```diff
@@ -888,36 +887,43 @@ def for_project_locale(
...
-                        & (Q(**{f"string__{i}regex": rf"{r}{y}{escape(s)}{y}{o}"}))
-                    )
-                    | Q(
-                        ~Q(resource__format="ftl")
-                        & (
-                            Q(**{f"string__{i}regex": rf"{y}{escape(s)}{y}"})
-                            | Q(**{f"string_plural__{i}regex": rf"{y}{escape(s)}{y}"})
+            entity_filters = (
+                (
+                    Q(pk=None)  # Ensures that no source strings are returned
```
mathjazz, line 892: Why is this not simply Q(), like in line 878?

harmitgoswami: During testing, it seemed that using Q() returns _all_ objects.

I found this [StackOverflow question](https://stackoverflow.com/questions/29900386/how-to-construct-django-q-object-matching-none) which was about the same problem I was encountering, and switching to Q(pk=None) now correctly returns no matches.

mathjazz: Oh. Oh! 🤦

mathjazz (approved):

PR #3352, opened 2024-09-18
Tweaked search panel design

### LINKED ISSUE #3345: Make search options clearer
> Wouldn't that create the same confusion you have about `Search in string identifiers`?

Why? If `Search in source strings` is disabled it doesn't imply `Search in string identifiers` can't be enabled. Which is the problem

> UPDATE: As for the checkboxes, since we have other 2 "Match…" options, I'm not a fan of the idea.

I meant something like this:

Posnetek zaslona 2024-09-17 ob 09 47 01

_Originally posted by @mathjazz in https://github.com/mozilla/pontoon/issues/3344#issuecomment-2354818812_

harmitgoswami:
Fix #3345 

As per the discussion in the related issue, this PR renames the following in `translate/src/modules/search/constants.ts`:

- `search_translations_only` --> `search_exclude_source_strings`
- `Search in translations only` --> `Exclude source strings`
- `Search in string identifiers` --> `Include string identifiers`
- `Search in rejected translations` --> `Include rejected translations`

Renaming the variables was done to improve clarity and more accurately reflect how the query changes in the backend. Additionally, I've added visual separation between the 'Match' and 'Include/Exclude' filters, reflecting the difference in their behaviours.

image

### DIFF
--- a/pontoon/base/models/entity.py
+++ b/pontoon/base/models/entity.py
@@ -732,7 +732,7 @@ class Entity(DirtyFieldsMixin, models.Model):
         search=None,
         extra=None,
         search_identifiers=None,
-        search_translations_only=None,
+        search_exclude_source_strings=None,
         search_rejected_translations=None,
         search_match_case=None,
         search_match_whole_word=None,
@@ -889,8 +889,8 @@ class Entity(DirtyFieldsMixin, models.Model):
             # Search in source strings
             entity_filters = (
                 (
-                    Q(pk=None)  # Ensures that no source strings are returned
-                    if search_translations_only
+                    Q(pk__in=[])  # Ensures that no source strings are returned
+                    if search_exclude_source_strings
                     else (
                         Q(
                             Q(resource__format="ftl")
--- a/translate/src/modules/search/components/SearchPanel.tsx
+++ b/translate/src/modules/search/components/SearchPanel.tsx
@@ -32,7 +32,7 @@ const SearchOption = ({
   onToggle: () => void;
 }) => {
   // Find the first option that doesn't have "Match" in the name, and apply top margin for visual seperation
-  const seperator = SEARCH_OPTIONS.find(
+  const separator = SEARCH_OPTIONS.find(
     (option) => !option.name.includes('Match'),
   )?.name;
 
@@ -41,7 +41,7 @@ const SearchOption = ({
       className={classNames(
         'check-box',
         selected && 'enabled',
-        name === seperator && 'option-seperator',
+        name === separator && 'option-separator',
       )}
       onClick={(ev) => {
         ev.stopPropagation();
--- a/translate/src/modules/search/components/SearchPanel.tsx
+++ b/translate/src/modules/search/components/SearchPanel.tsx
@@ -31,18 +31,9 @@ const SearchOption = ({
   selected: boolean;
   onToggle: () => void;
 }) => {
-  // Find the first option that doesn't have "Match" in the name, and apply top margin for visual seperation
-  const separator = SEARCH_OPTIONS.find(
-    (option) => !option.name.includes('Match'),
-  )?.name;
-
   return (
     <li
-      className={classNames(
-        'check-box',
-        selected && 'enabled',
-        name === separator && 'option-separator',
-      )}
+      className={classNames(`check-box ${slug}`, selected && 'enabled')}
       onClick={(ev) => {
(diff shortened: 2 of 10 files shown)


mathjazz:
Deploying this and #3349 to stage...

pontoon/base/models/entity.py
```diff
@@ -889,8 +889,8 @@ def for_project_locale(
             # Search in source strings
             entity_filters = (
                 (
-                    Q(pk=None)  # Ensures that no source strings are returned
-                    if search_translations_only
+                    Q(pk__in=[])  # Ensures that no source strings are returned
```
mathjazz, line 892: Why this change?

translate/src/modules/search/components/SearchPanel.tsx
```diff
@@ -31,9 +31,18 @@ const SearchOption = ({
   selected: boolean;
   onToggle: () => void;
 }) => {
+  // Find the first option that doesn't have "Match" in the name, and apply top margin for visual seperation
```
flodolo, line 34: At risk of asking a dumb question, this feels brittle. 

Can we have IDs associated to these checkboxes, and add a margin to that element?

harmitgoswami: So for example, `Include string identifiers` would get ID = 3, and then we apply the top margin to the element whose ID = 3? Or did I understand incorrectly

flodolo (approved):
Behavior looks correct to me on stage, only question about how to add the margin (assuming we even want one?)

translate/src/modules/search/components/SearchPanel.tsx
```diff
@@ -31,9 +31,18 @@ const SearchOption = ({
...
+  const separator = SEARCH_OPTIONS.find(
+    (option) => !option.name.includes('Match'),
+  )?.name;
+
```
mathjazz, line 38: That looks complex and fragile. I'd just use slug as the className and add margin to `.search_identifiers`.

Alternatively you could have two lists, once for Matches and one for Incudes/Excludes.

translate/src/modules/search/components/SearchPanel.tsx
flodolo: Not necessarily a numeric ID, a CSS ID (e.g. `id="option-include-ids"`). Let's wait for @mathjazz though, as I'm not sure if the space is wanted (it made sense when there was text between the two group of options).

translate/src/modules/search/components/SearchPanel.tsx
mathjazz: Deleting due to https://github.com/mozilla/pontoon/pull/3352#pullrequestreview-2313063497.

translate/src/modules/search/components/SearchPanel.tsx
mathjazz: My deleted (didn't notice this thread earlier) comment:

"That looks complex and fragile. I'd just use slug as the className and add margin to .search_identifiers.

Alternatively you could have two lists, once for Matches and one for Incudes/Excludes."

pontoon/base/models/entity.py
harmitgoswami: I read over the [StackOverflow post](https://stackoverflow.com/questions/29900386/how-to-construct-django-q-object-matching-none) again, it seems that using `pk__in=[]` is a safer and potentially less costly alternative than `pk=None`

mathjazz: Nice work with the research!

translate/src/modules/search/components/SearchPanel.tsx
mathjazz: Much clearer now, thanks!

Yeah, it's OK if some classes are unused.

mathjazz:
Could you please also apply these CSS changes copied from Firefox Devtools? Commented out rules need to be removed.

This will make padding and margins consistent with the Filters panel, otherwise switching between the two makes text jump a little.

```
.search-panel {
  & .menu {
    & .title {
      margin: 5px 0;
    }
    & ul {
      /* margin: 10px 0; */
      /* padding: 4px 0; */
      margin: 0;
    }
  }
  & .search-button {
    margin-bottom: 5px;
    margin-top: 8px;
  }
}

&:hover, & .applied-count {
  margin-top: 35px;
}

```

mathjazz (approved):

PR #3353, opened 2024-09-18
Add ability to see and recycle sent messages

### LINKED ISSUE #3246: Add ability to see the Sent messages

### LINKED ISSUE #3247: Add ability to recycle Sent messages
An old message can easily be used as a template for the new one.

Depends on #3246.

mathjazz:
Fix #3246: After the message is sent, it gets stored to the DB and displayed in the `Sent` panel. 
Fix #3247: Add ability to edit sent messages as new

Since @eemeli is on a PTO, I'll be integrating my work on #3247 directly into this PR. That will allow opening sent messages in the Compose panel, revealing more details about them, like the selected filters. It'll also make Sent and Compose panels accessible via URL, which will allow redirecting to the (updated) Sent panel after the message is sent (right now you need to refresh the page to see the sent message, which is annoying).

@bcolsson Could you please take this for a spin on [stage](https://mozilla-pontoon-staging.herokuapp.com/messaging/)? Note that temporarily messages are always sent to the sender, regardless if you click `Send to myself` or `Send`. You can also manage messages via [Django Admin](https://mozilla-pontoon-staging.herokuapp.com/a/messaging/message/).

Also included are some under the hood changes that will make code more maintainable:
* [Create MessageForm from the Message model](https://github.com/mozilla/pontoon/pull/3353/commits/b5b8ff0b70d86e4974db4097c62790f687881aa0)
* [Use select element instead of input in multiple item selector widget](https://github.com/mozilla/pontoon/pull/3353/commits/a41a697dec5f20ddb331048ba464658b080dd940)
* [Use form tags in the template](https://github.com/mozilla/pontoon/pull/3353/commits/8452802dca98f7435e8ea526002e402e5f779101)

**Updated on Sep 19, 14:20 UTC**: I've now integrated the fix for #3247. Related to that:
* Compose and Sent panels can now be linked to.
* After the message is sent, show it in the Sent panel.


mathjazz:
Known bugs:
* Checking Email as message type will result in an error (sending emails on stage is disabled by design)
* Messages are stored and sent as (sanitized) HTML for security reasons. When you edit sent messages, HTML is converted back to Markdown, which might not be exactly the same as the Markdown you have created when sending the message.

bcolsson:
@mathjazz - testing from a user perspective everything seems nice and functional. Liking it!

bcolsson (approved):

flodolo:
When previewing the message, it would be nice to visually differentiate the text from the rest the the UI, e.g.

Screenshot 2024-09-20 alle 07 51 43

Questions:
* Do we want all locales and all projects selected by default?

Issues:
* There are no checks on the fact that `Maximum` should be bigger than `Minimum`, or `From` earlier than `To`.

mathjazz:
> When previewing the message, it would be nice to visually differentiate the text from the rest the the UI, e.g.

Makese sense, I'll play with it a little.

> Do we want all locales and all projects selected by default?

That's what the spec says. It's also what we used in the [recent emails](https://github.com/mozilla-l10n/pontoon-scripts/blob/main/pm/email_script_events.py). But not in manual notifications. Maybe leave it for now and adapt if data shows it's not the right default?

> There are no checks on the fact that `Maximum` should be bigger than `Minimum`, or `From` earlier than `To`.

This is checked [implicitly](https://github.com/mozilla/pontoon/pull/3325#issuecomment-2348142324), because the number of recipients in cases like this will be zero. Same as above, I'd add an explicit check if this actually turns out to be a problem.

mathjazz:
> > When previewing the message, it would be nice to visually differentiate the text from the rest the the UI, e.g.
> 
> Makese sense, I'll play with it a little.

Made some minor adjustments to the styling, including this.

pontoon/messaging/static/js/messaging.js
```diff
@@ -162,13 +163,87 @@ $(function () {
     $('#review .message-type .transactional').toggle(isTransactional);
   }
 
+  function loadPanelContent(path) {
+    if (inProgress) {
+      inProgress.abort();
+    }
```
eemeli, line 169: Is `inProgress` ever set back to `false`, or is `loadPanelContent()` only meant to be callable once?

pontoon/messaging/static/js/messaging.js
```diff
@@ -162,13 +163,87 @@ $(function () {
...
+
+    const panel = container.find('.right-column');
+    const relative_path = path.split('/messaging/')[1];
+    const menu_path = path === '/messaging/sent/' ? path : '/messaging/';
```
eemeli, line 173: Further down there's a check for `if (relative_path === 'sent/')`, why is the check here different?

pontoon/messaging/static/js/messaging.js
```diff
@@ -162,13 +163,87 @@ $(function () {
...
+      .parents('li')
+      .addClass('selected')
+      .siblings()
+      .removeClass('selected');
+
+    panel.empty();
+
+    inProgress = $.ajax({
+      url: `/messaging/ajax/${relative_path}`,
+      success: function (data) {
```
eemeli, line 187: Is there a reason to persist with this verbose style?
```suggestion
      success(data) {
```

pontoon/messaging/static/js/messaging.js
```diff
@@ -162,13 +163,87 @@ $(function () {
...
+
+  // Load panel conent on page load
+  loadPanelContent(window.location.pathname);
+
+  // Load panel conent on history change
+  window.onpopstate = function () {
+    loadPanelContent(window.location.pathname);
+  };
+
+  // Load panel conent on menu click
```
eemeli, line 223:
```suggestion
  // Load panel content on page load
  loadPanelContent(window.location.pathname);

  // Load panel content on history change
  window.onpopstate = function () {
    loadPanelContent(window.location.pathname);
  };

  // Load panel content on menu click
```

pontoon/messaging/static/js/messaging.js
```diff
@@ -162,13 +163,87 @@ $(function () {
...
+  container.on(
+    'click',
+    '.left-column a, .right-column .edit-as-new',
+    function (e) {
+      // Keep default middle-, control- and command-click behaviour (open in new tab)
+      if (e.which === 2 || e.metaKey || e.ctrlKey) {
+        return;
+      }
```
eemeli, line 231: How about shift+click?

eemeli (changes requested):
Sorry for the time it took me to get to this. A few small things inline, but I'd appreciate a chance to see this on stage again and take it for a spin?

pontoon/messaging/static/js/messaging.js
mathjazz: We'd need to file a bug for that and change it across all pages with tabs.

mathjazz:
> Sorry for the time it took me to get to this. A few small things inline, but I'd appreciate a chance to see this on stage again and take it for a spin?

Deployed to stage.

pontoon/messaging/static/js/messaging.js
mathjazz: Updated.

pontoon/messaging/templates/messaging/includes/sent.html
```diff
@@ -0,0 +1,39 @@
...
+                    <span class="types">
+                    {% if message.notification %}<i class="type far fa-bell" title="Sent as Notification"></i>{% endif %}
+                    {% if message.email %}<i class="type far fa-envelope" title="Sent as Email"></i>{% endif %}
+                    </span>
+                </div>
+                <p class="subject">{{ message.subject }}</p>
+                <div class="body"><div class="value">{{ message.body|safe }}</div></div>
+            </div>
+            <div class="footer">
+                <a href="{{ url('pontoon.messaging.edit_as_new', message.pk) }}" class="edit-as-new"><i class="fa fa-pencil-alt"></i>Edit as new</a>
```
eemeli, line 33: "Edit as new" is not clear and misleading. Something like "Use as template" would more accurately signal that you'll be creating a new message based on this message.

pontoon/messaging/templates/messaging/includes/compose.html
```diff
@@ -0,0 +1,224 @@
...
+        <p class="transactional"><i>Warning:</i> Transactional emails are also sent to users who have
+            not opted in to email communication. They are restricted in the type of content that can be
+            included. When in doubt, please review with legal.</p>
+    </section>
+    <menu class="controls">
+        <button class="button toggle edit" data-target="#compose"><span class="fa fa-chevron-left"></span>Back to
+            editing</button>
+        <div class="right">
+            <button class="button send to-myself">Send to myself</button>
+            <button class="button active send">Send</button>
```
eemeli, line 221: It's not clear that "Send to myself" will not save the message as "Sent", like "Send" does.

Will a later update be introducing something like "Drafts" for messages that have not been sent?

pontoon/messaging/templates/messaging/includes/sent.html
```diff
@@ -0,0 +1,39 @@
...
+                <time>{{ message.sent_at|format_for_inbox }}</time>
```
eemeli, line 34: I initially didn't see this "sent at" info at all. I would expect for this to be near the top of the message, where we also have the sender, the recipient count & message type.

eemeli (changes requested):

pontoon/messaging/static/js/messaging.js
eemeli: The translate frontend already checks for `shiftKey` when it checks for `ctrlKey`, so we shouldn't deviate from that more in the places where we do special stuff on clicks.
```suggestion
      // Keep default middle-, control- and command-click behaviour (open in new tab)
      if (e.which === 2 || e.metaKey || e.ctrlKey || e.shiftKey) {
        return;
      }
```

pontoon/messaging/templates/messaging/includes/sent.html
mathjazz: It's common pattern to show timestamp under the message bubble in messaging apps. I think you'll get used to it quickly.

It'll be busy in the header.

pontoon/messaging/templates/messaging/includes/compose.html
mathjazz: I overlooked that the spec actually says "Send **test** to myself". That should make it clearer.

We'll see if there's actually a need for something like Drafts and the prioritize accordingly. Even showing Sent messages and adding the ability to recycle them wasn't initially planned.

pontoon/messaging/static/js/messaging.js
mathjazz: I meant this: https://github.com/mozilla/pontoon/blob/main/pontoon/base/static/js/tabs.js#L22

mathjazz: I'll update both files.

pontoon/messaging/templates/messaging/includes/sent.html
mathjazz: I like "Use as template".

mathjazz:
Updated and re-deployed to stage. I also had to rebase.

eemeli (approved):
Looks good to go. Instead of rebasing, doing a merge from `main` to the PR branch would make the changes easier to review.

Issue #3355, opened 2024-09-19
Clicking on "APPLY SEARCH OPTIONS" will reload the page, but won't use the input field text to search


flodolo:
1. Type `cookie` in the search field.
2. Change the search options and select `APPLY SEARCH OPTIONS`.
3. Pontoon will reload the page, `cookie` will still be in the input field, but the page actually didn't search for `cookie`.

We should either clear the input field, or actually search for `cookie`. My preference would be for the latter.

PR #3356, opened 2024-09-19
Fixed bugs regarding state preservation

### LINKED ISSUE #3354: Clicking outside of the search options will lose selected checkboxes
1. Click on the search options and select some of them.
2. Click outside of the dropdown to make it disappear.
3. Open the search options again.
4. All options are deselected.

We should probably keep the state here, otherwise it's confusing for users.

harmitgoswami:
Fix #3354 and #3355 

This PR addresses the current bugs that relate to preserving state when the search panel is discarded and search terms are typed into the search bar. 

Previously, discarding the panel with a search option toggled on will toggle off that option. Also, if a search term was entered into the search bar and the Enter/Return key is not clicked, but the `APPLY SEARCH OPTIONS` button was clicked, then the term inside the search bar will not be included in the query.

This PR fixes the first issue by introducing a new state to track which options are selected, and syncing those with searchOptions whenever the panel is visible. The second issue is addressed by moving the `search` variable outside of the spread operator in `SearchBox.tsx`.

### DIFF
--- a/translate/src/modules/search/components/SearchBox.test.js
+++ b/translate/src/modules/search/components/SearchBox.test.js
@@ -270,7 +270,7 @@ describe('<SearchBox>', () => {
     // Wait until Enter is pressed.
     wrapper.find('input#search').simulate('keydown', { key: 'Enter' });
 
-    expect(spy.callCount).toBe(1);
+    expect(spy.callCount).toBe(2);
     expect(spy.firstCall.args).toMatchObject([
       {
         pathname: '/kg/firefox/all-resources/',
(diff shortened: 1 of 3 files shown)


mathjazz:
Nice work!

While #3354 is technically fixed, I think this is still a bug:

1. Click on the search options and select some of them.
1. Click outside of the dropdown to make it disappear.
1. Open the search options again.
1. Previously selected options are now selected. (good job, #3354 is fixed!)
1. Type cookie in the search field.
1. Press Enter.
1. Search results don't reflect the previously selected options.
1. If you open the search options again, all options will be deselected.

harmitgoswami:
What's our intended behaviour here, just so I'm clearly understanding? Do we want the selected search options to still apply when you hit Enter, even without clicking the `APPLY SEARCH OPTIONS` button?

mathjazz (approved):
Thanks, it works great!

mathjazz:
@flodolo Since you filed the bugs, could you please test it on stage?

flodolo (approved):
Works great for me as well, great job 👍🏼

PR #3359, opened 2024-09-20
Remove django-sslserver

mathjazz:
I don't remember ever using this library, which allows running Django apps locally via HTTPS. The library hasn't been updated in years and the same can be achieved using `django-extensions`.

I also fixed `make requirements`.


docker/compile_requirements.sh
```diff
@@ -5,7 +5,7 @@
 
 export CUSTOM_COMPILE_COMMAND="./docker/compile_requirements.sh"
 
-uv pip compile --generate-hashes $@ requirements/default.in
-uv pip compile --generate-hashes $@ requirements/dev.in
-uv pip compile --generate-hashes $@ requirements/lint.in
-uv pip compile --generate-hashes $@ requirements/test.in
+uv pip compile --generate-hashes $@ requirements/default.in -o requirements/default.txt
```
flodolo, line 8: D'oh 👼🏼

flodolo (approved):

PR #3360, opened 2024-09-20
Collect coverage in Backend workflow (Python)

flodolo:
Fixes issue #3362


flodolo:
One more note: I tested it locally outside of docker, and the command `pytest --cov-report=xml:pontoon/coverage.xml --cov=.` only generates one `coverage.xml` file, no clue where the 2nd one comes from.

mathjazz (approved):

PR #3361, opened 2024-09-20
Created inital test cases for SearchPanel

### LINKED ISSUE #3335: Create tests for Pontoon search options
As part of the Search enhancement, we should create a series of `Jest` test cases to ensure any future changes to `SearchPanel.tsx` don't cause unexpected behaviour.

The test cases should be similar to the ones found in `FiltersPanel.test.js`.

harmitgoswami:
Fix #3335 

This PR creates a file, called `SearchPanel.test.js`, which includes all the `jest` test cases for the new `SearchPanel` component. By adding test cases, any new modifications to the `SearchPanel` feature will hopefully not cause any breaking changes.

Most of the test cases added are very similar to the ones found in `FiltersPanel.test.js`. 

**To test changes**: Open a new terminal and make sure you are in the root directory (`~/pontoon`). Then, execute the following command in your terminal to run only the newly added test cases: 

`npm test -w translate translate/src/modules/search/components/SearchPanel.test.js`


translate/src/modules/search/components/SearchPanel.test.js
```diff
@@ -0,0 +1,96 @@
...
+    const wrapper = mountComponentWithStore(SearchPanelDialog, store, {
+      options: selectedSearchOptions,
+      selectedSearchOptions,
+    });
+
+    for (let option of SEARCH_OPTIONS) {
+      expect(wrapper.find(`.menu .${option.slug}`).hasClass('enabled')).toBe(
+        selectedSearchOptions[option.slug],
+      );
+    }
```
mathjazz, line 30: You could use the same pattern as below:

```suggestion
    for (let { slug } of SEARCH_OPTIONS) {
      expect(wrapper.find(`.menu .${slug}`).hasClass('enabled')).toBe(
        selectedSearchOptions[slug],
      );
    }
```

translate/src/modules/search/components/SearchPanel.test.js
```diff
@@ -0,0 +1,96 @@
...
+
+describe('<SearchPanelDialog>', () => {
+  it('correctly sets search option as selected', () => {
+    const selectedSearchOptions = {
+      search_identifiers: false,
+      search_exclude_source_strings: false,
+      search_rejected_translations: true,
+      search_match_case: true,
+      search_match_whole_word: false,
+    };
```
mathjazz, line 18: We use these defaults a few times. I'd only declare it once and then reuse.

mathjazz (commented):
Nice work!

Why does the title say "initial", but the PR closes the issue? Are there more test coming?

What about the backend tests? Are we going to track them under a separate bug?

It would be great to also cover https://github.com/mozilla/pontoon/pull/3356 with the tests.

pontoon/base/tests/managers/test_entity.py
```diff
@@ -1080,6 +1080,113 @@ def test_mgr_entity_filter_combined(admin, resource_a, locale_a, user_a):
...
+                admin,
+                resource_a.project,
+                locale_a,
+                search="testentity",
+                search_match_case=True,
+                author=user_a.email,
+            )
+        )
+        == []
+    )
```
mathjazz, line 1135: For each search option, it would be nice to have two tests - one where the option is set to True, and one where it is not. I would also organize each search option into a separate test case (function).

pontoon/base/tests/managers/test_entity.py
```diff
@@ -1080,6 +1080,113 @@ def test_mgr_entity_filter_combined(admin, resource_a, locale_a, user_a):
...
+
+    # Test search_rejected_translations
+    assert list(
+        Entity.for_project_locale(
+            admin,
+            resource_a.project,
+            locale_a,
+            search="TestEntity0",
+            search_exclude_source_strings=True,
+            search_rejected_translations=True,
```
mathjazz, line 1169: We should have a separate test case that only tests the `search_rejected_translations` parameter. Separately, I'd create a test case that tests combined filters.

harmitgoswami:
@mathjazz are the test cases that I wrote in `test_mgr_entity_option_combined` what you had in mind when you said "combined filters"?

pontoon/base/models/entity.py
```diff
@@ -924,6 +924,8 @@ def for_project_locale(
             entities = Entity.objects.filter(
                 pk__in=set(list(translation_matches) + list(entity_matches))
             )
+            for e in entities:
+                print(e.string)
```
mathjazz, line 928: Please remove :)

pontoon/base/tests/managers/test_entity.py
```diff
@@ -1080,6 +1080,357 @@ def test_mgr_entity_filter_combined(admin, resource_a, locale_a, user_a):
...
+    )
+
+    # Base case
+    assert list(
+        Entity.for_project_locale(
+            admin,
+            resource_a.project,
+            locale_a,
+            search="testentity",
+            author=user_a.email,
```
mathjazz, line 1114: Ideally, we should construct Base test as:

```python
assert list(Entity.for_project_locale(*args, **kwargs)) == ...
```

And then we'd just modiffy kwargs accordingly. That would mean less repetition and less chance for mistakes.

mathjazz (commented):
Great job! I'd try to reduce more code, see my comment inline.

mathjazz (approved):
\o/

Issue #2766, opened 2023-02-20
Drop apscheduler


mathjazz:
We only use apscheduler to schedule a periodic AutoML warmup task.

We could probably use something simpler instead, especially because one of its dependencies (`pytz-deprecation-shim`) requires an additional dependency (`backports.zoneinfo`) on Python 3.8, so our current installation in that environment fails.

flodolo:
I think this was fixed via #3933

PR #3369, opened 2024-09-24
Update APScheduler to 3.10.4

flodolo:
Stumbled upon #2766. Looks like the [project](https://github.com/agronholm/apscheduler) is actively maintained, they're even working on a 4.0 release, so maybe we should close that issue?

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -18,9 +18,9 @@ anyio==4.3.0 \
     # via
     #   httpx
     #   openai
-apscheduler==3.9.1.post1 \
-    --hash=sha256:b2bea0309569da53a7261bfa0ce19c67ddbfe151bda776a6a907579fdbd3eb2a \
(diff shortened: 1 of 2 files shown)


requirements/default.txt
```diff
@@ -1218,10 +1218,6 @@ sacrebleu==2.3.1 \
 sacremoses==0.0.43 \
     --hash=sha256:123c1bf2664351fb05e16f87d3786dbe44a050cfd7b85161c09ad9a63a8e2948
     # via -r requirements/default.in
-setuptools==75.1.0 \
```
flodolo, line 1221: This got in with your last update, but I'm not sure how/why.

`make requirements` removes it for me, and I'm not sure why yours added it.

mathjazz (approved):

mathjazz:
> Stumbled upon #2766. Looks like the [project](https://github.com/agronholm/apscheduler) is actively maintained, they're even working on a 4.0 release, so maybe we should close that issue?

The issue is still valid - it's a dependency we don't really need. I removed the obsolete part.

PR #3376, opened 2024-09-25
Replace unmaintained wsgi-sslify with Django settings

mathjazz:
One more. :)

Deployed to stage. Also tested locally.

### DIFF
--- a/pontoon/settings/base.py
+++ b/pontoon/settings/base.py
@@ -873,9 +873,9 @@ X_FRAME_OPTIONS = "DENY"
 # Use correct header for detecting HTTPS on Heroku.
 SECURE_PROXY_SSL_HEADER = ("HTTP_X_FORWARDED_PROTO", "https")
 
-# Do not set SECURE_HSTS_SECONDS.
-# HSTS is being taken care of in pontoon/wsgi.py.
-# SECURE_HSTS_SECONDS = 63072000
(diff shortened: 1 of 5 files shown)


pontoon/settings/base.py
```diff
@@ -873,9 +873,9 @@ def _default_from_email():
 # Use correct header for detecting HTTPS on Heroku.
 SECURE_PROXY_SSL_HEADER = ("HTTP_X_FORWARDED_PROTO", "https")
 
-# Do not set SECURE_HSTS_SECONDS.
-# HSTS is being taken care of in pontoon/wsgi.py.
-# SECURE_HSTS_SECONDS = 63072000
+# If set to a non-zero integer value, the SecurityMiddleware sets the
+# HTTP Strict Transport Security header on all responses that do not already have it.
+SECURE_HSTS_SECONDS = 31536000
```
flodolo, line 878: Can we put in the comment that this is 365 days?

flodolo (approved):

pontoon/settings/base.py
mathjazz: Updated.

mathjazz:
I came across https://bugzilla.mozilla.org/show_bug.cgi?id=1529895.

It seems HSTS header is now set also for static files.

PR #3377, opened 2024-09-25
Update reactour to latest version

mathjazz:
The important bit here is the removal of (the outdated version of) `lodash.pick`.

Deployed to stage.


flodolo (approved):
I don't feel as confident about JS dependencies, but this feels safe-ish.

PR #3383, opened 2024-09-27
CI: add checks for Python and npm package licenses

flodolo:
A few notes:
1. I originally wanted to use [licensecheck](https://pypi.org/project/licensecheck/), because you don't need to install the dependencies with that tool. Unfortunately I can't seem to make it check other requirements files ([filed issue here](https://github.com/FHPythonUtils/LicenseCheck/issues/90)).
2. For npm, `license-checker` hasn't been updated in a while, there's a fork. One general issue is that `--failOn` will print nothing when it fails (i.e. there is a GPL package), but if it doesn't fail it prints again the same output as `license-checker`. Opted to print a summary on the 2nd run.
3. Given you need to install the dependencies for both, I opted to have the check as a step in existing workflows (originally I had them in a new separate workflow), and to install the 2 packages directly in the workflow.


flodolo:
Not sure why checks are not being triggered here 🤔

flodolo:
Modern development…

Npm 842 packages

├─ MIT: 705
├─ ISC: 42
├─ BSD-2-Clause: 28
├─ BSD-3-Clause: 27
├─ Apache-2.0: 25
├─ (MIT OR CC0-1.0): 3
├─ 0BSD: 2
├─ BSD: 2
├─ (Unlicense OR Apache-2.0): 1
├─ Python-2.0: 1
├─ CC-BY-4.0: 1
├─ (Apache-2.0 OR MPL-1.1): 1
├─ Custom: <LICENSE>: 1
├─ Custom: https://www.highcharts.com/license: 1
├─ UNLICENSED: 1
└─ CC0-1.0: 1

Python only 108 packages.

mathjazz (approved):

PR #3385, opened 2024-09-28
Update to eslint v9

### LINKED ISSUE #3384: Update ESLint to version 9
Looks like ESLint v8 (the one we currently use) is going EOL next week (2024-10-05).

At least one plugin doesn't seem to be compatible:
* `eslint-plugin-import` ([issue](https://github.com/import-js/eslint-plugin-import/issues/2948))

This likely needs someone more familiar with JS than me to fix.

flodolo:
Fixes #3384

Going through the individual commits might help:
1. I moved the excluded paths from `.eslintignore` into the config. The first surprise there is that I had to add `ignorePatterns` also to the config file in `/translate/.eslintrc.js`.
2. Updated `eslint` up to 8.40. Beyond that, the update was not working out of the box.
3. Updated to `eslint` 8.57.1 (had to install also the `globals` package, see [SO thread](https://stackoverflow.com/questions/75524079/eslint-flat-config-eslint-config-js-throwing-parsing-error-the-keyword-imp/75524936#75524936)).
4. Converted configs to new flat format (via [migration tool](https://eslint.org/docs/latest/use/configure/migration-guide)).
5. Fixed remaining errors. This one needs someone who truly understands JS to tell if I just patched the problem in the wrong place. Again, surprising the previous `ignores` rule for `/static` wasn't working 🤷🏼 
6. Updated `caniuse` because I was tired of seeing the warning.

Unrelated:
* `*venv` doesn't seem to work in ignore files, ended up adding both `venv` (virtualenv) and `.venv` (uv style).
* Not sure why `package-lock.json` still had references to `tag-admin`?


flodolo:
7. Manually remove one more zombie piece of `tag-admin` from `package-lock.json`.

package.json
```diff
@@ -35,6 +35,9 @@
     "@babel/preset-env": "^7.13.10",
     "@babel/preset-react": "^7.12.13",
     "@babel/runtime": "^7.13.10",
+    "@eslint/compat": "^1.1.1",
```
flodolo, line 38: These are required by the new flat config (you get the instruction to install them after running the migration).

flodolo:
I actually realized that the update to node 9 is relatively straightforward at this point, so trying to use this PR to do the full jump to v9.

eslint.config.mjs
```diff
@@ -0,0 +1,145 @@
...
+  allConfig: js.configs.all,
+});
+
+export default [
+  {
+    ignores: [
+      '**/.vscode/',
+      '.github/',
+      'media/',
+      '**/static/',
```
mathjazz, line 24: Will that not exclude e.g. `pontoon/base/static`?

flodolo: Yes it would 🤦🏼

pontoon/administration/static/js/admin_project.js
```diff
@@ -194,7 +194,7 @@ $(function () {
 
     try {
       projectLocales = JSON.parse($(this).val());
-    } catch (error) {
+    } catch {
```
flodolo, line 197: Fixing this via eslint comment turned out to be a nightmare

```js
        try {
          data = JSON.parse(event.data);
        }
        // eslint-disable-next-line no-unused-vars
        catch (_) {
          return;
        }
```

This works, but prettier will move the comment with the catch block, breaking eslint.

Started looking into https://github.com/prettier/eslint-plugin-prettier, but that's a rabbit hole I don't want to go down for these 2 errors.

eslint.config.mjs
```diff
@@ -0,0 +1,145 @@
...
+      // Vendored code
+      'error_pages/css/blockrain.css',
+      'error_pages/js/',
+      'pontoon/base/static/css/boilerplate.css',
+      'pontoon/base/static/css/fontawesome-all.css',
+      'pontoon/base/static/css/nprogress.css',
+      'pontoon/base/static/js/lib/',
+    ],
+  },
+  ...compat.extends('eslint:recommended', 'plugin:react/recommended'),
```
eemeli, line 43: This should be pretty easy to replace with a proper non-compat solution, something like this:
```suggestion
  js.configs.recommended,
  reactPlugin.configs.flat.recommended,
```

flodolo: Removed (this was generated by the conversion tool).

translate/eslint.config.mjs
```diff
@@ -0,0 +1,61 @@
...
+  baseDirectory: __dirname,
+  recommendedConfig: js.configs.recommended,
+  allConfig: js.configs.all,
+});
+
+export default [
+  ...compat.extends(
+    'eslint:recommended',
+    'plugin:@typescript-eslint/recommended',
+  ),
```
eemeli, line 23: Same here as with the root config; the compat stuff should not be necessary.

Note that for TS the package to import has changed: https://typescript-eslint.io/getting-started/

eemeli (changes requested):

pontoon/administration/static/js/admin_project.js
eemeli: As we're not using the variables, eliding the `(error)` or `(_)` completely is the right choice.

mathjazz (approved):

eslint.config.mjs
```diff
@@ -0,0 +1,135 @@
...
+    plugins: {
+      react,
+      import: importPlugin,
+    },
+
+    languageOptions: {
+      globals: {
+        ...globals.browser,
+        ...globals.jest,
+        gettext: false,
```
eemeli, line 44: I had to check what `false` meant here, and it's a [deprecated synonym](https://eslint.org/docs/latest/use/configure/language-options#specifying-globals) for `'readonly'`, which would be clearer here.

translate/eslint.config.mjs
```diff
@@ -0,0 +1,49 @@
...
+
+      '@typescript-eslint/no-unused-vars': [
+        'error',
+        {
+          varsIgnorePattern: '^_',
+        },
+      ],
+
+      '@typescript-eslint/prefer-as-const': 0,
+      'import/no-default-export': 'error',
```
eemeli, line 46: This doesn't need a plugin: https://eslint.org/docs/latest/rules/no-restricted-exports#restrictdefaultexports

translate/eslint.config.mjs
```diff
@@ -0,0 +1,49 @@
+import typescriptEslint from '@typescript-eslint/eslint-plugin';
+import _import from 'eslint-plugin-import';
```
eemeli, line 2: This is a different (unnecessary) import plugin than in the root config.

eslint.config.mjs
```diff
@@ -0,0 +1,135 @@
...
+      'pontoon/base/static/css/nprogress.css',
+      'pontoon/base/static/js/lib/',
+    ],
+  },
+  eslint.configs.recommended,
+  react.configs.flat.recommended,
+  {
+    plugins: {
+      react,
+      import: importPlugin,
```
eemeli, line 37: This plugin isn't used, and could/should be dropped.

eslint.config.mjs
```diff
@@ -0,0 +1,135 @@
...
+        DIFF_DELETE: false,
+        ga: false,
+        process: false,
+        generalShortcutsHandler: true,
+        traversalShortcutsHandler: true,
+        editorShortcutsHandler: true,
+        showdown: true,
+      },
+
+      parser: babelParser,
```
eemeli, line 80: Do we really still need Babel to parse the JS sources? That seems unlikely, esp. as it's not required for JSX.

eemeli (changes requested):
Sorry for not doing the full review all at once; this should be the last batch of comments.

eslint.config.mjs
flodolo: That I don't really know. If I use `tsParser` like in the main config it works, if I drop it I get 22 errors of these types

```
error  Parsing error: Unexpected token {
error  Parsing error: Unexpected token ..
error  Parsing error: Unexpected token ?
```

eemeli: When testing this, those go away when we also drop the unnecessary `ecmaVersion`. I'll add a suggestion below for how to simplify this that seems to work for me.

eslint.config.mjs
```diff
@@ -0,0 +1,133 @@
...
+          jsx: true,
+          experimentalObjectRestSpread: true,
+        },
+
+        babelOptions: {
+          presets: ['@babel/preset-react'],
+        },
+
+        requireConfigFile: false,
+      },
```
eemeli, line 93:
```suggestion
      parserOptions: {
        ecmaFeatures: {
          jsx: true,
        },
      },
```

eslint.config.mjs
```diff
@@ -0,0 +1,133 @@
...
+    ignores: [
+      '**/.vscode/',
+      '.github/',
+      'media/',
+      'static/',
+      '**/coverage/',
+      'docs/_build/',
+      '**/package-lock.json',
+      '**/specs/',
+      '**/?venv/',
```
eemeli, line 17: This does not appear to catch `.../venv/...` paths in practice.

flodolo: Is `**/*venv*/` too broad? The alternative is to list both `.venv` and `venv`, technically the former is the suggested way (using `uv`).

flodolo: Thinking back about it, `?venv` is not supposed to catch `venv`, because it needs one character 🤦🏼 

I'm guessing the options are:
- Keep as is
- List both `venv` and `.venv` (and fix all the other ignore files).

eemeli: Listing both seems like the easiest choice.

eemeli (approved):

PR #3388, opened 2024-09-29
Fix graphql on DEV server, update django-graphene to 3.2.2

### LINKED ISSUE #3292: GraphQL does not load
http://localhost:8000/graphql fails with a JS error.

flodolo:
Fixes #3292

### DIFF
--- a/pontoon/settings/dev.py
+++ b/pontoon/settings/dev.py
@@ -47,7 +47,7 @@ CSP_SCRIPT_SRC = base.CSP_SCRIPT_SRC + (
     # Needed for GraphiQL
     "https://cdn.jsdelivr.net",
     # Needed for GraphiQL (inline script)
-    "'sha256-gp1+DqtmqR6gC56O1TE7F+GuoHAHHbXyN+gaBi8gcjo='",
+    "'sha256-HHh/PGb5Jp8ck+QB/v7zeWzuHf3vYssM0CBPvYgEHR4='",
 )
 CSP_STYLE_SRC = base.CSP_STYLE_SRC + (
     # Needed for GraphiQL
--- a/pontoon/settings/dev.py
+++ b/pontoon/settings/dev.py
@@ -41,6 +41,7 @@ TEMPLATES[0]["OPTIONS"]["match_regex"] = re.compile(
     re.VERBOSE,
 )
 
+CSP_FONT_SRC = base.CSP_FONT_SRC + ("data:",)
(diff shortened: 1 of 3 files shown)


mathjazz (commented):
The UI now loads, but there's still CSP errors in the Console (related to fonts). It would be nice to clear them while we're here.

Also, I realized http://localhost:8000/graphql/ doesn't load and http://localhost:8000/graphql does. Should we change that, given all other URLs are defined with a trailing `/`?

flodolo:
> Also, I realized [http://localhost:8000/graphql](http://localhost:8000/graphql/) doesn't load and http://localhost:8000/graphql does. Should we change that, given all other URLs are defined with a trailing `/`?

How can you avoid breaking existing scripts using Pontoon API ([example](https://github.com/mozilla-l10n/pontoon-scripts/blob/main/stats/health_report/extract_completion_data.py#L43))?

Is this code enough?

```python
urlpatterns = [
    # Redirect from /graphql to /graphql/
    path("graphql", RedirectView.as_view(url="/graphql/", permanent=True)),

    # GraphQL endpoint. In DEV mode it serves the GraphiQL IDE if accessed with
    # Accept: text/html
    path(
        "graphql/",
        GraphQLView.as_view(schema=schema, graphiql=DEV),
    ),
]
```

mathjazz:
> How can you avoid breaking existing scripts using Pontoon API ([example](https://github.com/mozilla-l10n/pontoon-scripts/blob/main/stats/health_report/extract_completion_data.py#L43))?

I would expect that to "just work": https://docs.djangoproject.com/en/5.1/ref/settings/#append-slash

flodolo:
Should the documentation be updated? 

> I would expect that to "just work": [docs.djangoproject.com/en/5.1/ref/settings#append-slash](https://docs.djangoproject.com/en/5.1/ref/settings/#append-slash)

I don't think it does?

As DEV, `http://localhost:8000/graphql?query={projects{name}}` works and get changed to `http://localhost:8000/graphql/#query=%7B%0A%20%20projects%20%7B%0A%20%20%20%20name%0A%20%20%7D%0A%7D`. 

But If I set it as `DEV=False`, that first URL gets an error (`"Must provide query string."`).

Can we move this part to a separate issue? I'd be happy to have someone more familiar with Django to look into that ;-)

mathjazz:
> Can we move this part to a separate issue? I'd be happy to have someone more familiar with Django to look into that ;-)

Sure, let's leave this out of this PR.

mathjazz (approved):

PR #3400, opened 2024-10-07
Add role banner under user avatar

### LINKED ISSUE #3393: Implement status icons in user avatars
As per the [spec on gamification](https://github.com/mozilla/pontoon/blob/3b840b967932e898b4ed398f7c7ea3d847106dd2/specs/0119-gamification-badges.md#status-icons), there should be four different status icons that users can have, which will be displayed next to their name:

- **New User Icon**: For users who joined less than 3 months ago.
- **Translator Icon**: For users contributing as translators.
- **Manager Icon**: For users serving as locale managers.
- **Admin Icon**: For users holding administrative roles.

For now, we've agreed to keep the icons as simple stylized text, and will iterate on the design as needed.

harmitgoswami:
Fix #3393 

This PR adds the user role as a banner under their avatar in the `translate` view. As per discussions in team calls, the design for the banner is to be iterated on within this PR.

**To reproduce results**: Head to any string within any team/project that has more than just the imported translation available (i.e. a user has submitted a translation for that string). At the bottom of their avatar, their role will be displayed within a banner.

image

### DIFF
--- a/pontoon/base/models/user.py
+++ b/pontoon/base/models/user.py
@@ -194,7 +194,7 @@ def user_locale_role(self, locale):
         return "Contributor"
 
 
-def user_locale_status(self, locale):
+def user_status(self, locale):
     if self.username == "Imported":
         return ("", "")
     if self.is_superuser:
@@ -439,7 +439,7 @@ User.add_to_class("managed_locales", user_managed_locales)
 User.add_to_class("translated_projects", user_translated_projects)
 User.add_to_class("role", user_role)
 User.add_to_class("locale_role", user_locale_role)
-User.add_to_class("locale_status", user_locale_status)
+User.add_to_class("status", user_status)
 User.add_to_class("contributed_translations", contributed_translations)
 User.add_to_class("top_contributed_locale", top_contributed_locale)
 User.add_to_class("can_translate", can_translate)
--- a/pontoon/base/models/user.py
+++ b/pontoon/base/models/user.py
@@ -195,18 +195,17 @@ def user_locale_role(self, locale):
 
 
 def user_locale_status(self, locale):
-    if self is None:
-        return ""
+    if self.username == "Imported":
+        return ("", "Imported")
     if self.is_superuser:
-        return "Admin"
+        return ("ADMIN", "Admin")
     if self in locale.managers_group.user_set.all():
-        return "Manager"
+        return ("MGNR", "Manager")
     if self in locale.translators_group.user_set.all():
-        return "Translator"
+        return ("TRNSL", "Translator")
     if self.date_joined >= timezone.now() - relativedelta(months=3):
-        return "New User"
-    else:
-        return ""
+        return ("NEW", "New User")
+    return ("", "")
 
 
 @property
--- a/pontoon/base/models/user.py
+++ b/pontoon/base/models/user.py
@@ -196,11 +196,11 @@ def user_locale_role(self, locale):
 
 def user_locale_status(self, locale):
     if self.username == "Imported":
-        return ("", "Imported")
+        return ("", "")
     if self.is_superuser:
         return ("ADMIN", "Admin")
     if self in locale.managers_group.user_set.all():
-        return ("MGNR", "Manager")
+        return ("MNGR", "Manager")
     if self in locale.translators_group.user_set.all():
         return ("TRNSL", "Translator")
     if self.date_joined >= timezone.now() - relativedelta(months=3):
--- a/pontoon/base/models/user.py
+++ b/pontoon/base/models/user.py
@@ -1,6 +1,7 @@
 from hashlib import md5
 from urllib.parse import quote, urlencode
 
+from dateutil.relativedelta import relativedelta
 from guardian.shortcuts import get_objects_for_user
 
 from django.conf import settings
@@ -8,6 +9,7 @@ from django.contrib.auth.models import User
 from django.contrib.contenttypes.models import ContentType
 from django.db.models import Count
 from django.urls import reverse
+from django.utils import timezone
 
 from pontoon.actionlog.models import ActionLog
 
@@ -190,6 +192,21 @@ def user_locale_role(self, locale):
         return "System User"
     else:
         return "Contributor"
+
+
+def user_locale_status(self, locale):
+    if self is None:
+        return ""
+    if self.is_superuser:
+        return "Admin"
+    if self in locale.managers_group.user_set.all():
+        return "Manager"
+    if self in locale.translators_group.user_set.all():
+        return "Translator"
+    if self.date_joined >= timezone.now() - relativedelta(months=1):
+        return "New User"
+    else:
+        return ""
 
 
 @property
@@ -423,6 +440,7 @@ User.add_to_class("managed_locales", user_managed_locales)
 User.add_to_class("translated_projects", user_translated_projects)
 User.add_to_class("role", user_role)
 User.add_to_class("locale_role", user_locale_role)
+User.add_to_class("locale_status", user_locale_status)
 User.add_to_class("contributed_translations", contributed_translations)
 User.add_to_class("top_contributed_locale", top_contributed_locale)
 User.add_to_class("can_translate", can_translate)
--- a/pontoon/base/models/user.py
+++ b/pontoon/base/models/user.py
@@ -203,7 +203,7 @@ def user_locale_status(self, locale):
         return "Manager"
     if self in locale.translators_group.user_set.all():
         return "Translator"
-    if self.date_joined >= timezone.now() - relativedelta(months=1):
+    if self.date_joined >= timezone.now() - relativedelta(months=3):
         return "New User"
     else:
         return ""
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -869,6 +869,7 @@ def user_data(request):
             "display_name": user.display_name,
             "name_or_email": user.name_or_email,
             "username": user.username,
+            "date_joined": user.date_joined,
             "contributor_for_locales": list(
                 user.translation_set.values_list("locale__code", flat=True).distinct()
             ),
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -414,17 +414,13 @@ def get_translation_history(request):
     for t in translations:
         u = t.user or User(username="Imported", first_name="Imported", email="imported")
         translation_dict = t.serialize()
-        if t.user:
-            user_role = User.locale_role(t.user, locale)
-        else:
-            user_role = ""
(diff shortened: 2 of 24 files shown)


pontoon/base/views.py
```diff
@@ -414,12 +414,17 @@ def get_translation_history(request):
...
+            user_role = User.locale_role(t.user, locale)
+        else:
+            user_role = ""
         translation_dict.update(
             {
                 "user": u.name_or_email,
                 "uid": u.id,
                 "username": u.username,
                 "user_gravatar_url_small": u.gravatar_url(88),
+                "user_role": user_role,
```
mathjazz, line 427: I believe the following should be the same (and then you can drop the new code above):
```suggestion
                "user_role": u.locale_role,
```

However, `User.locale_role()` is not exactly what we need here according to the spec. We also need the "New" icon and don't need the "System User" and "Contributor" roles.

translate/public/locale/en-US/translate.ftl
```diff
@@ -704,6 +704,8 @@ user-UserAvatar--anon-alt-text =
     .alt = Anonymous User
 user-UserAvatar--alt-text = 
     .alt = User Profile
+user-UserAvatar--userRole =
+    { $userRole }
```
mathjazz, line 708: This doesn't make the user roles localizable.

translate/src/modules/history/components/HistoryTranslation.tsx
```diff
@@ -129,7 +129,7 @@ const User = ({
       {user}
     </a>
   ) : (
-    <span title={title}>{user}</span>
+    <span title={title}> {user} </span>
```
mathjazz, line 132: Is there a reason for added whitespace?

mathjazz (commented):
Nice work!

The spec is not exactly clear where to show the status icons. It mentions translations (suggestions) and comments and shows the status icon in the Profile. It also says "Status Icons: Show status icons universally in the user avatar".

I am not sure we should show them everywhere we show the avatar. In the Contributors dashboards and user filters, we already show the roles. I also don't think it's useful to show it in the Profile icon or in the Settings page, since that's only visible to the user.

How about we keep it just next to translations and comments (covered by this patch) and also add them to the Profile page? @flodolo Given that you were the reviewer of the spec, what's your take on this?

flodolo:
> The spec is not exactly clear where to show the status icons. It mentions translations (suggestions) and comments and shows the status icon in the Profile. It also says "Status Icons: Show status icons universally in the user avatar".
> 
> I am not sure we should show them everywhere we show the avatar. In the Contributors dashboards and user filters, we already show the roles. I also don't think it's useful to show it in the Profile icon or in the Settings page, since that's only visible to the user.
> 
> How about we keep it just next to translations and comments (covered by this patch) and also add them to the Profile page? @flodolo Given that you were the reviewer of the spec, what's your take on this?

Looks like I didn't actually review the final version 🤔 

I agree on showing the status only in translations and comments:
* That information is useful when reviewing (is this person new? Is the person leaving a comment/suggestion a translator for that locale?)
* Same for comments (am I getting a comment from someone who doesn't know how Pontoon works?).

harmitgoswami:
As we discussed, it seems like some of the statuses do indeed overflow when displayed in the banner. 

image

image

flodolo:
Thoughts on using just letters to keep labels short (A/T/M + New), and rely on tooltip for explanation?

harmitgoswami:
I think it would work for 'New', however, since a lot of users have their avatar as the default (the initials of their username), I personally feel that it might look odd to have up to three letters displayed in the avatar container.

flodolo:
> I think it would work for 'New', however, since a lot of users have their avatar as the default (the initials of their username), I personally feel that it might look odd to have up to three letters displayed in the avatar container.

We could make it more "badge-y".

Screenshot 2024-10-08 alle 21 37 53

mathjazz:
I like flod's idea for the status icon. 

Below is a rough variant of this proposal, which uses different colors for different statuses and shortens the titles using the similar method that we use on the dashboards.

Posnetek zaslona 2024-10-09 ob 10 09 03

pontoon/base/models/user.py
```diff
@@ -192,6 +194,21 @@ def user_locale_role(self, locale):
         return "Contributor"
 
 
+def user_locale_status(self, locale):
+    if self is None:
```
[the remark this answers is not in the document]
mathjazz: See https://github.com/mozilla/pontoon/pull/3400#discussion_r1792993102.

translate/src/modules/user/components/UserAvatar.tsx
```diff
@@ -5,10 +5,13 @@ type Props = {
...
   title?: string;
   imageUrl: string;
+  userStatus?: string;
 };
 
 export function UserAvatar(props: Props): React.ReactElement<'div'> {
-  const { username, title, imageUrl } = props;
+  const { username, title, imageUrl, userStatus } = props;
+  const role = userStatus ? userStatus[0] : '';
+  const tooltip = userStatus ? userStatus[1] : '';
```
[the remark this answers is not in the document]
harmitgoswami: Would `const [role, tooltip] = userStatus || [];` work instead? I had the suggested change before, but TypeScript seems to not like that :(

`Type 'string | undefined' must have a '[Symbol.iterator]()' method that returns an iterator.ts(2488)
`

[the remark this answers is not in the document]
mathjazz: Can we instead make `userStatus` a required prop (by removing `?` in line 8)?

Right now the `AddComment` component doesn't use it, so we need to make it optional. But I'd show the user status there as well, given that it'll appear in the comment list after you post the comment.

Or at least it should. Seems like user statuses don't appear in the Comment list?

translate/src/modules/comments/components/Comment.tsx
```diff
@@ -32,10 +32,13 @@ export function Comment(props: Props): null | React.ReactElement<'li'> {
 
   return (
     <li className='comment'>
-      <UserAvatar
-        username={comment.username}
-        imageUrl={comment.userGravatarUrlSmall}
-      />
+      <div className='avatar-container'>
```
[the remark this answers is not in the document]
harmitgoswami: Sorry, I misread the original comment!

harmitgoswami:
I've held off on creating `useLocaleRole.test.js` for now, in case we need to tweak the hook any further

translate/src/hooks/useLocaleRole.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  if (!isAuthenticated) {
+    return ['', ''];
+  }
+
+  if (isAdmin) {
+    return ['ADMIN', 'Admin'];
+  }
+
+  if (Object.hasOwnProperty.call(managerForLocales, code)) {
```
mathjazz, line 28: I believe this is clearer:

```suggestion
  if (managerForLocales.includes(code)) {
```

translate/src/hooks/useLocaleRole.ts
```diff
@@ -0,0 +1,44 @@
...
+    return ['MNGR', 'Manger'];
+  }
+
+  if (Object.hasOwnProperty.call(translatorForLocales, code)) {
```
mathjazz, line 32: Same as for `managerForLocales`.

translate/src/modules/comments/components/CommentsList.tsx
```diff
@@ -18,6 +19,8 @@ export function CommentsList({
   user,
 }: Props): React.ReactElement<'div'> {
   const onAddComment = useAddCommentAndRefresh(translation);
+  const role = useLocaleRole();
```
mathjazz, line 22: This seems unused, so changes to this file are not needed?

translate/src/modules/user/components/UserAvatar.tsx
```diff
@@ -19,9 +21,19 @@ export function UserAvatar(props: Props): React.ReactElement<'div'> {
...
         onClick={(e: React.MouseEvent) => e.stopPropagation()}
       >
-        <Localized id='user-UserAvatar--alt-text' attrs={{ alt: true }}>
-          <img src={imageUrl} alt='User Profile' height='44' width='44' />
-        </Localized>
+        <div className='avatar-container'>
+          <Localized id='user-UserAvatar--alt-text' attrs={{ alt: true }}>
+            <img src={imageUrl} alt='User Profile' height='44' width='44' />
+          </Localized>
+          {userStatus && (
```
mathjazz, line 28: Now that `userStatus` is always set, the check is no longer sufficient:

```suggestion
          {role && (
```

translate/src/modules/user/components/UserAvatar.tsx
```diff
@@ -5,10 +5,12 @@ type Props = {
   username: string;
   title?: string;
   imageUrl: string;
+  userStatus: string | string[];
```
mathjazz, line 8: `userStatus` should actually always be `string[]`:

```suggestion
  userStatus: string[];
```

To make the `make types` pass, you'll also need to fix the original sin - the type set in both files in the `api` folder.

translate/src/modules/user/components/UserAvatar.css
```diff
@@ -0,0 +1,59 @@
...
+.user-avatar .avatar-container .user-status-banner.translator {
+  color: var(--status-translated);
+}
+
+.user-avatar .avatar-container .user-status-banner.new-user {
+  color: var(--status-fuzzy);
+}
+
+.history .translation:hover .user-status-banner {
+  border-color: var(--translation-border);
```
mathjazz, line 49: Is this needed? Why only for translations?

translate/src/modules/user/components/UserAvatar.css
```diff
@@ -0,0 +1,59 @@
...
+}
+
+.history .translation .user-avatar img {
+  border-radius: 6px;
+  border: 2px solid var(--icon-border-1);
+}
+
+.history .translation:hover .user-avatar img {
+  border-color: var(--translation-border);
+}
```
mathjazz, line 59: This two rules should be moved to the top, after the first rule.

mathjazz (commented):
Works great! A few comments, but they are all small.

translate/src/modules/comments/components/CommentsList.tsx
```diff
@@ -18,6 +18,7 @@ export function CommentsList({
   user,
 }: Props): React.ReactElement<'div'> {
   const onAddComment = useAddCommentAndRefresh(translation);
+
```
mathjazz, line 21: Nit: Please restore this change and we can ship it! :)

mathjazz (approved):

PR #3401, opened 2024-10-08
Clean up .gitignore, dropping /translate/.gitignore as unnecessary

eemeli:
Reorganise the rules into sections, and sort alphabetically: prefixes, suffixes, matches anywhere, and matches at root. Adds `.*` to catch all hidden files by default (they can still be added with `--force`). With that, `/translate/.gitignore` becomes unnecessary.

Filing this, as I want to add a `.python-version` file in my local pontoon dir to have pyenv automatically activate the right virtualenv when `cd`'ing into the project.


.gitignore
```diff
@@ -1,34 +1,27 @@
...
+*~
+
+build.py
+build/
 coverage/
-*.log
+dist/
+node_modules/
+tmp/
+venv/
```
flodolo, line 18: We lose `.venv` with this?

eemeli: No, that's covered by `.*`.

.gitignore
```diff
@@ -1,34 +1,27 @@
...
 
-# Docker
-.server-build
-docker/config/server.env
-docker/config/server.env-e
+/docker/config/server.env
+/docker/config/server.env-e
```
mathjazz, line 21: Is that still a thing?

mathjazz (approved):

.gitignore
eemeli: Good point, I'll drop the rule. Note that you'll need to manually remove the file if you still have it; the one in my local clone was last touched in February 2022.
```suggestion
```

PR #3402, opened 2024-10-09
Add traffic throttling middleware

mathjazz:
A simple IP throttling middleware.

A maximum of 100 requests from the same IP is allowed within 1 minute. Subsequent requests are blocked for 10 minutes.


pontoon/settings/base.py
```diff
@@ -301,11 +301,16 @@ def _default_from_email():
...
+THROTTLE_MAX_COUNT = 100  # A maximum of 100 requests
+THROTTLE_PERIOD = 60  # allowed in 1 minute.
+THROTTLE_BLOCK_DURATION = 600  # Subsequent requests are blocked for 10 minutes.
+
 MIDDLEWARE = (
     "django.middleware.security.SecurityMiddleware",
     "whitenoise.middleware.WhiteNoiseMiddleware",
     "django.middleware.gzip.GZipMiddleware",
     "pontoon.base.middleware.RaygunExceptionMiddleware",
+    "pontoon.base.middleware.ThrottleIpMiddleware",
```
flodolo, line 313: Should we have a setting, e.g. THROTTLE_ENABLED, and include this conditionally?

flodolo:
Thoughts? As pointed out on Slack, I don't think we're currently using the `test` settings anywhere.

```python
diff --git a/pontoon/base/tests/test_middleware.py b/pontoon/base/tests/test_middleware.py
index 1579c97f..ae8de0ef 100644
--- a/pontoon/base/tests/test_middleware.py
+++ b/pontoon/base/tests/test_middleware.py
@@ -34,6 +34,7 @@ def test_EmailConsentMiddleware(client, member, settings):
 
 
+@override_settings(THROTTLE_ENABLED=True)
 class ThrottleIpMiddlewareTest(TestCase):
     def setUp(self):
         self.client = Client()
diff --git a/pontoon/settings/base.py b/pontoon/settings/base.py
index eae4da67..7f71a36b 100644
--- a/pontoon/settings/base.py
+++ b/pontoon/settings/base.py
@@ -301,8 +301,9 @@ for ip in blocked_ip_settings:
             log.error(f"Invalid IP or IP range defined in BLOCKED_IPS: {ip}")
 
+THROTTLE_ENABLED = True
 THROTTLE_MAX_COUNT = 100  # A maximum of 100 requests
-THROTTLE_PERIOD = 60  # allowed in 1 minute.
+THROTTLE_PERIOD = 60  # allowed within 1 minute.
 THROTTLE_BLOCK_DURATION = 600  # Subsequent requests are blocked for 10 minutes.
 
 MIDDLEWARE = (
@@ -310,7 +311,6 @@ MIDDLEWARE = (
     "whitenoise.middleware.WhiteNoiseMiddleware",
     "django.middleware.gzip.GZipMiddleware",
     "pontoon.base.middleware.RaygunExceptionMiddleware",
-    "pontoon.base.middleware.ThrottleIpMiddleware",
     "pontoon.base.middleware.BlockedIpMiddleware",
     "corsheaders.middleware.CorsMiddleware",
     "django.middleware.common.CommonMiddleware",
@@ -323,6 +323,9 @@ MIDDLEWARE = (
     "pontoon.base.middleware.EmailConsentMiddleware",
 )
 
+if THROTTLE_ENABLED:
+    MIDDLEWARE += ("pontoon.base.middleware.ThrottleIpMiddleware",)
+
 CONTEXT_PROCESSORS = (
     "django.contrib.auth.context_processors.auth",
     "django.template.context_processors.debug",
diff --git a/pontoon/settings/dev.py b/pontoon/settings/dev.py
index 338157ad..28eddf6c 100644
--- a/pontoon/settings/dev.py
+++ b/pontoon/settings/dev.py
@@ -1,4 +1,4 @@
-"""A various settings that enable additional packages that are helpful in day to day development."""
+"""Various settings that enable additional packages that are helpful in day to day development."""
 
 import copy
 import re
diff --git a/pontoon/settings/test.py b/pontoon/settings/test.py
index 59385675..1a219dd1 100644
--- a/pontoon/settings/test.py
+++ b/pontoon/settings/test.py
@@ -1,4 +1,12 @@
+# ruff: noqa: F403, F405
 """Django settings for the test suite."""
 
-PIPELINE_ENABLED = False
-CELERY_ALWAYS_EAGER = True
+from .base import *
+
+
+# In tests, remove throttling middleware.
+MIDDLEWARE = tuple(
+    middleware
+    for middleware in MIDDLEWARE
+    if middleware != "pontoon.base.middleware.ThrottleIpMiddleware"
+)
diff --git a/pytest.ini b/pytest.ini
index 13fe229f..1e7f07dc 100644
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,5 +1,4 @@
 [pytest]
 addopts = --doctest-glob='*.tst' --reuse-db
 norecursedirs = .git _build tmp* requirements commands/* node_modules/*
-DJANGO_SETTINGS_MODULE = pontoon.settings
-
+DJANGO_SETTINGS_MODULE = pontoon.settings.test
```

flodolo:
Code seems to work well locally, I can't figure out why the tests are failing (it almost looks like the cache stays empty).

pontoon/base/tests/test_middleware.py
```diff
@@ -29,3 +32,69 @@ def test_EmailConsentMiddleware(client, member, settings):
...
+        # 6th request should be throttled
+        response = self.client.get(url, REMOTE_ADDR="192.168.0.2")
+        self.assertEqual(response.status_code, 429)
+
+        # Wait for block duration to pass
+        time.sleep(3)
+
+        # Make another request after block duration
+        response = self.client.get(url, REMOTE_ADDR="192.168.0.2")
+        self.assertEqual(response.status_code, 200)
```
flodolo, line 81: This fails because `blocked_ip_192.168.0.2` was cleared out, but `throttle_ip_192.168.0.2` is still there (`THROTTLE_PERIOD` is the default setting, so 60s).

flodolo:
Alternative approach, with throttling disabled by default and a few tweaks to the test (this passes locally).

```diff
diff --git a/pontoon/base/middleware.py b/pontoon/base/middleware.py
index e79dc937..7f541252 100644
--- a/pontoon/base/middleware.py
+++ b/pontoon/base/middleware.py
@@ -95,43 +95,44 @@ class ThrottleIpMiddleware:
         self.block_duration = settings.THROTTLE_BLOCK_DURATION
 
     def __call__(self, request):
-        # Get the client's IP address
-        try:
-            ip = request.META["HTTP_X_FORWARDED_FOR"]
-            # If comma-separated list of IPs, take just the last one
-            # http://stackoverflow.com/a/18517550
-            ip = ip.split(",")[-1]
-        except KeyError:
-            ip = request.META["REMOTE_ADDR"]
-
-        ip = ip.strip()
-
-        # Generate cache keys
-        throttle_key = f"throttle_ip_{ip}"
-        blocked_key = f"blocked_ip_{ip}"
-
-        # Check if IP is currently blocked
-        if cache.get(blocked_key):
-            return JsonResponse({"detail": "Too Many Requests"}, status=429)
-
-        # Fetch current request count and timestamp
-        request_data = cache.get(throttle_key)
-        now = time.time()
-
-        if request_data:
-            request_count, first_request_time = request_data
-            if request_count >= self.max_count:
-                # Block further requests for block_duration seconds
-                cache.set(blocked_key, True, self.block_duration)
+        if settings.THROTTLE_ENABLED:
+            # Get the client's IP address
+            try:
+                ip = request.META["HTTP_X_FORWARDED_FOR"]
+                # If comma-separated list of IPs, take just the last one
+                # http://stackoverflow.com/a/18517550
+                ip = ip.split(",")[-1]
+            except KeyError:
+                ip = request.META["REMOTE_ADDR"]
+
+            ip = ip.strip()
+
+            # Generate cache keys
+            throttled_key = f"throttled_ip_{ip}"
+            blocked_key = f"blocked_ip_{ip}"
+
+            # Check if IP is currently blocked
+            if cache.get(blocked_key):
                 return JsonResponse({"detail": "Too Many Requests"}, status=429)
+
+            # Fetch current request count and timestamp
+            request_data = cache.get(throttled_key)
+            now = time.time()
+
+            if request_data:
+                request_count, first_request_time = request_data
+                if request_count >= self.max_count:
+                    # Block further requests for block_duration seconds
+                    cache.set(blocked_key, True, self.block_duration)
+                    return JsonResponse({"detail": "Too Many Requests"}, status=429)
+                else:
+                    # Increment the request count and update cache
+                    cache.set(
+                        throttled_key, (request_count + 1, first_request_time), self.period
+                    )
             else:
-                # Increment the request count and update cache
-                cache.set(
-                    throttle_key, (request_count + 1, first_request_time), self.period
-                )
-        else:
-            # Reset the count and timestamp if first request in the period
-            cache.set(throttle_key, (1, now), self.period)
+                # Reset the count and timestamp if first request in the period
+                cache.set(throttled_key, (1, now), self.period)
 
         response = self.get_response(request)
         return response
diff --git a/pontoon/base/tests/test_middleware.py b/pontoon/base/tests/test_middleware.py
index 1579c97f..3d1a92d5 100644
--- a/pontoon/base/tests/test_middleware.py
+++ b/pontoon/base/tests/test_middleware.py
@@ -2,6 +2,7 @@ import time
 
 import pytest
 
+from django.core.cache import cache
 from django.test import Client, TestCase, override_settings
 from django.urls import reverse
 from django.utils import timezone
@@ -34,67 +35,71 @@ def test_EmailConsentMiddleware(client, member, settings):
     assert response.status_code == 200
 
 
+@override_settings(THROTTLE_ENABLED=True)
 class ThrottleIpMiddlewareTest(TestCase):
     def setUp(self):
         self.client = Client()
+        self.url = "/"
+
+    def tearDown(self):
+        cache.clear()
+        return super().tearDown()
 
     @override_settings(THROTTLE_MAX_COUNT=5)
     @override_settings(THROTTLE_BLOCK_DURATION=60)
     def test_throttle_limit_reached(self):
         """Test that requests are throttled after the limit is reached."""
-        url = "/"
 
         # Make 5 requests within the limit
         for _ in range(5):
-            response = self.client.get(url, REMOTE_ADDR="192.168.0.1")
+            response = self.client.get(self.url, REMOTE_ADDR="192.168.0.1")
             self.assertEqual(response.status_code, 200)
 
         # 6th request should be throttled
-        response = self.client.get(url, REMOTE_ADDR="192.168.0.1")
+        response = self.client.get(self.url, REMOTE_ADDR="192.168.0.1")
         self.assertEqual(response.status_code, 429)
         self.assertEqual(response.json(), {"detail": "Too Many Requests"})
 
         # Check that the IP remains blocked for the block duration
-        response = self.client.get(url, REMOTE_ADDR="192.168.0.1")
+        response = self.client.get(self.url, REMOTE_ADDR="192.168.0.1")
         self.assertEqual(response.status_code, 429)
 
     @override_settings(THROTTLE_MAX_COUNT=5)
+    @override_settings(THROTTLE_PERIOD=2)
     @override_settings(THROTTLE_BLOCK_DURATION=2)
     def test_throttle_reset_after_block_duration(self):
         """Test that requests are allowed after the block duration expires."""
-        url = "/"
 
         # Make 5 requests within the limit
         for _ in range(5):
-            response = self.client.get(url, REMOTE_ADDR="192.168.0.2")
+            response = self.client.get(self.url, REMOTE_ADDR="192.168.0.2")
             self.assertEqual(response.status_code, 200)
 
         # 6th request should be throttled
-        response = self.client.get(url, REMOTE_ADDR="192.168.0.2")
+        response = self.client.get(self.url, REMOTE_ADDR="192.168.0.2")
         self.assertEqual(response.status_code, 429)
 
         # Wait for block duration to pass
         time.sleep(3)
 
         # Make another request after block duration
-        response = self.client.get(url, REMOTE_ADDR="192.168.0.2")
+        response = self.client.get(self.url, REMOTE_ADDR="192.168.0.2")
         self.assertEqual(response.status_code, 200)
 
     @override_settings(THROTTLE_MAX_COUNT=5)
     @override_settings(THROTTLE_BLOCK_DURATION=60)
     def test_throttle_with_different_ips(self):
         """Test that throttling is applied separately for different IPs."""
-        url = "/"
 
         # Make 5 requests from IP1
         for _ in range(5):
-            response = self.client.get(url, REMOTE_ADDR="192.168.0.3")
+            response = self.client.get(self.url, REMOTE_ADDR="192.168.0.3")
             self.assertEqual(response.status_code, 200)
 
         # 6th request from IP1 should be throttled
-        response = self.client.get(url, REMOTE_ADDR="192.168.0.3")
+        response = self.client.get(self.url, REMOTE_ADDR="192.168.0.3")
         self.assertEqual(response.status_code, 429)
 
         # Requests from IP2 should not be throttled
-        response = self.client.get(url, REMOTE_ADDR="192.168.0.4")
+        response = self.client.get(self.url, REMOTE_ADDR="192.168.0.4")
         self.assertEqual(response.status_code, 200)
diff --git a/pontoon/settings/base.py b/pontoon/settings/base.py
index eae4da67..141c0b0e 100644
--- a/pontoon/settings/base.py
+++ b/pontoon/settings/base.py
@@ -301,8 +301,9 @@ for ip in blocked_ip_settings:
             log = logging.getLogger(__name__)
             log.error(f"Invalid IP or IP range defined in BLOCKED_IPS: {ip}")
 
+THROTTLE_ENABLED = False
 THROTTLE_MAX_COUNT = 100  # A maximum of 100 requests
-THROTTLE_PERIOD = 60  # allowed in 1 minute.
+THROTTLE_PERIOD = 60  # allowed within 1 minute.
 THROTTLE_BLOCK_DURATION = 600  # Subsequent requests are blocked for 10 minutes.
 
 MIDDLEWARE = (
```

pontoon/settings/base.py
```diff
@@ -301,11 +301,16 @@ def _default_from_email():
             log = logging.getLogger(__name__)
             log.error(f"Invalid IP or IP range defined in BLOCKED_IPS: {ip}")
 
+THROTTLE_MAX_COUNT = 100  # A maximum of 100 requests
+THROTTLE_PERIOD = 60  # allowed in 1 minute.
```
flodolo, line 305: Looking at today's attach, this might have been too optimistic. 

With multiple IPs creating less than 50 requests in 3–4 minutes Pontoon goes down. Maybe 50 requests over 60s is a more sane value.

mathjazz: 50/m is easy to hit for legitimate users.

I also see up to 1000 RPS during the latest incident.

I'd start with a conservative number and then tune it.

docker-compose.yml
```diff
@@ -22,7 +22,7 @@ services:
 
   # Database
   postgresql:
-    image: postgres:15
+    image: postgres:13.4
```
flodolo, line 25: I feel like I've seen this before ;-)

pontoon/base/middleware.py
```diff
@@ -83,3 +77,54 @@ def __call__(self, request):
...
+
+        # Fetch current request count and timestamp
+        request_data = cache.get(throttle_key)
+        now = time.time()
+
+        if request_data:
+            request_count, first_request_time = request_data
+            if request_count >= self.max_count:
+                # Block further requests for block_duration seconds
+                cache.set(blocked_key, True, self.block_duration)
```
flodolo, line 116: Do you think it would be worth to log this action, to get a sense of how often we block IPs?

docker-compose.yml
mathjazz: 🤦

flodolo:
This looks good to me, but it hsould be documented (`deployment.rst`, `maintenance.rst`)

mathjazz:
The patch is deployed to stage and throttling is enabled. Try searching for "Too Many Requests" and "Blocked IP" in the log.

I've managed to get one of my devices blocked. And unblocked 10 minutes later.

I think we should show a simple static page (like https://pontoon.mozilla.org/404/) instead of JSON to blocked users.

Other than that I think this is indeed good to go as designed. I'll document the variables.

flodolo:
Tested on stage and managed to get myself blocked. I think you're right in being conservative, didn't realize how many requests we send on a translate page.

pontoon/base/middleware.py
```diff
@@ -83,3 +79,55 @@ def __call__(self, request):
...
+            self.observation_period = self.block_duration
+
+    def __call__(self, request):
+        if settings.THROTTLE_ENABLED is False:
+            return self.get_response(request)
+
+        ip = get_ip(request)
+
+        # Generate cache keys
+        throttle_key = f"throttle_ip_{ip}"
```
flodolo, line 103: nitpick: I would go with `throttled_key` for consistency on these (throttled/blocked).

flodolo (approved):

eemeli (commented):
We should include a [Retry-After](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) header in the 429 response.

I'm also a bit concerned about 300 requests/minute being way too easy to hit with valid usage patterns. For example, in the translate view selecting a new string triggers 6 requests, and so navigating in the string list with option+up/down arrow hits that limit really fast.

mathjazz:
> We should include a [Retry-After](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) header in the 429 response.

Done.

eemeli (approved):
Approving with the understanding that when this gets deployed we'll inform users about it, and provide ways (both manual and automated) to ensure that false positives are caught.

PR #3404, opened 2024-10-09
Reduce the number of DB queries when retrieving translation history

mathjazz:
Noticed this while reviewing #3400.


harmitgoswami (changes requested):
Everything looks good, although I think adding line 427 in this PR is causing some error?

Removing line 427 seems to fix the issue

image

mathjazz:
> Everything looks good, although I think adding line 427 in this PR is causing some error?

Oops, great catch!

harmitgoswami (approved):

PR #3405, opened 2024-10-10
Remove pontoon/settings/test.py

mathjazz:


flodolo:
Looks like it's indeed unused. Question is if we actually want to start using it.
https://github.com/mozilla/pontoon/pull/3402#issuecomment-2404281922

flodolo (approved):

PR #3397, opened 2024-10-02
Added badge fields to user_profile.py

harmitgoswami:
Fix #3394 

This PR adds the relevant fields for the upcoming user badges to `user_profile.py`. 

As required from the `choices` field, each field is a tuple, with the first value being the actual value, and the second value being the human-readable version of the field. 

Not sure if `IntegerField` is the best choice here, so please let me know if you see a cleaner way of adding these fields!

Issue #2195, opened 2019-11-25
Replace PermissionChangelog table with ActionLog


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1599384](https://bugzilla.mozilla.org/show_bug.cgi?id=1599384)
Bug Reporter: @adngdb

The ActionLog model should be the one place where we log events happening in Pontoon. It currently doesn't have all the fields required to do what PermissionChangelog does, but we can easily add them. And so we shall do.

## Steps

1. Add missing fields to the ActionLog model.
2. Backfill data from PermissionChangelog to ActionLog.
3. Replace all queries using PermissionChangelog to use ActionLog instead.
4. Remove the PermissionChangelog table.

PR #3413, opened 2024-10-16
Added backend logic to award badges

### LINKED ISSUE #3396: Add logic to award badges on relevant activities
Following the work from #3394, we need to award/update a user's badges according to the [milestones listed here in the spec](https://github.com/mozilla/pontoon/blob/3b840b967932e898b4ed398f7c7ea3d847106dd2/specs/0119-gamification-badges.md#feature-explanation). 

~~We should also send a notification to the user upon the update or achievement of a badge, congratulating them for their hard work :^)~~ (see #3412)

harmitgoswami:
Fix #3396 

This PR adds the relevant stat tracking needed to award badges based on the criteria defined in the [specification](https://github.com/mozilla/pontoon/blob/3b840b967932e898b4ed398f7c7ea3d847106dd2/specs/0119-gamification-badges.md#feature-explanation). If these criteria are met, the relevant badge field in `pontoon/base/models/user_profile.py` is updated to reflect the user's new badge level.

Currently, I've only began tracking activity on or after today's date (October 16 2024), but in the future, this will be changed to the date of the official badges release in Pontoon. I've also held off on creating tests within the initial commit in case adjustments need to be made.

**To reproduce results**: Badges are awarded when users either submit translations, review translations, or promote users. Submitting and reviewing translations are all done in the `translate` view, so adding `print(user.profile.[BADGE TYPE])` statements before and after the updating logic in `pontoon/translations/views.py` should indicate if a badge was successfully awarded.

Similarly, for promoting users, head to the `{locale}/permissions` page for any team (with admin permissions) and observe how promoting users updates the fields using print statements in `pontoon/base/forms.py::assign_users_to_groups`.

Alternatively, the Django admin panel should also serve as an easy way to observe the changes in the `user_profile` fields.

### DIFF
--- a/pontoon/base/forms.py
+++ b/pontoon/base/forms.py
@@ -100,6 +100,15 @@ class UserPermissionLogFormMixin:
         if users:
             group.user_set.add(*users)
 
+        badge_thresholds = [1, 2, 5]
+        for level, threshold in enumerate(badge_thresholds, start=1):
+            if (
+                self.user.profile.community_builder_level < level
+                and len(self.user.badges_promoted_users) >= threshold
+            ):
+                self.user.profile.community_builder_level = level
+                # TODO: Send a notification to the user
+
         log_group_members(self.user, group, (add_users, remove_users))
 
 
--- a/pontoon/base/models/user.py
+++ b/pontoon/base/models/user.py
@@ -12,6 +12,7 @@ from django.urls import reverse
 from django.utils import timezone
 
 from pontoon.actionlog.models import ActionLog
+from pontoon.base.utils import aware_datetime
 
 
 @property
@@ -214,6 +215,47 @@ def contributed_translations(self):
     from pontoon.base.models.translation import Translation
 
     return Translation.objects.filter(user=self)
+
+
+@property
+def badges_translations_count(self):
+    """Contributions provided by user that count towards their badges."""
+    from pontoon.base.models.translation import Translation
+
+    badges_inception = aware_datetime(2024, 10, 16)
+
+    return Translation.objects.filter(user=self, date__gte=badges_inception)
+
+
+@property
+def badges_reviewed_translations(self):
+    """Translation reviews provided by user that count towards their badges."""
+    from pontoon.base.models.translation import Translation
+
+    badges_inception = aware_datetime(2024, 10, 16)
+
+    approvals = Translation.objects.filter(
+        approved_user=self, date__gte=badges_inception
+    )
+    rejections = Translation.objects.filter(
+        rejected_user=self, date__gte=badges_inception
+    )
+
+    return list(approvals) + list(rejections)
+
+
+@property
+def badges_promoted_users(self):
+    """Role promotions performed by user that count towards their badges"""
+    from pontoon.base.models import PermissionChangelog
+
+    badges_inception = aware_datetime(2024, 10, 16)
+
+    return PermissionChangelog.objects.filter(
+        performed_by=self,
+        action_type=PermissionChangelog.ActionType.ADDED,
+        created_at__gte=badges_inception,
+    )
 
 
 @property
@@ -441,6 +483,9 @@ User.add_to_class("role", user_role)
 User.add_to_class("locale_role", user_locale_role)
 User.add_to_class("status", user_status)
 User.add_to_class("contributed_translations", contributed_translations)
+User.add_to_class("badges_translations_count", badges_translations_count)
+User.add_to_class("badges_reviewed_translations", badges_reviewed_translations)
+User.add_to_class("badges_promoted_users", badges_promoted_users)
 User.add_to_class("top_contributed_locale", top_contributed_locale)
 User.add_to_class("can_translate", can_translate)
 User.add_to_class("is_new_contributor", is_new_contributor)
--- a/pontoon/translations/views.py
+++ b/pontoon/translations/views.py
@@ -169,6 +169,16 @@ def create_translation(request):
                 description=desc,
             )
 
+    # Update Translation Champion Badge stats
+    badge_thresholds = [5, 50, 250, 1000]
+    for level, threshold in enumerate(badge_thresholds, start=1):
+        if (
+            user.profile.translation_champion_level < level
+            and len(user.badges_translations_count) >= threshold
+        ):
+            user.profile.translation_champion_level = level
+            # TODO: Send a notification to the user
+
     return JsonResponse(
         {
             "status": True,
@@ -309,6 +319,16 @@ def approve_translation(request):
         plural_form=translation.plural_form,
     )
 
+    # Update Review Master Badge stats
+    badge_thresholds = [5, 50, 250, 1000]
+    for level, threshold in enumerate(badge_thresholds, start=1):
+        if (
+            user.profile.review_master_level < level
+            and len(user.badges_reviewed_translations) >= threshold
+        ):
+            user.profile.review_master_level = level
+            # TODO: Send a notification to the user
+
     return JsonResponse(
         {
             "translation": active_translation.serialize(),
@@ -438,6 +458,17 @@ def reject_translation(request):
         locale=locale,
         plural_form=translation.plural_form,
     )
+
+    # Update Review Master Badge stats
+    user = request.user
+    badge_thresholds = [5, 50, 250, 1000]
+    for level, threshold in enumerate(badge_thresholds, start=1):
+        if (
+            user.profile.review_master_level < level
+            and len(user.badges_reviewed_translations) >= threshold
+        ):
+            user.profile.review_master_level = level
+            # TODO: Send a notification to the user
 
     return JsonResponse(
         {


pontoon/translations/views.py
```diff
@@ -309,6 +319,16 @@ def approve_translation(request):
         plural_form=translation.plural_form,
     )
 
+    # Update Review Master Badge stats
+    badge_thresholds = [5, 50, 250, 1000]
+    for level, threshold in enumerate(badge_thresholds, start=1):
+        if (
+            user.profile.review_master_level < level
+            and len(user.badges_reviewed_translations) >= threshold
```
mathjazz, line 327: These two lines will result in 12 additional DB queries, which is a lot for each approve / reject operation, because DB queries are slow.

I'm thinking if it'll be better to just store translation / review / promotion counts in the DB for each user, and then use these counts here to check if we need to send a "badge awarded" notification to the user and again in the Profile page to determine which badge to show.

By doing that we wouldn't need to maintain the `badges_inception` date.

Thoughts?

mathjazz: It turns out checking translation/review counts is fast and can be done as part of saving/reviewing translations. The same also applies for promotions, of course, which happens much less often.

```python
>>> from django.utils import timezone
>>> from datetime import datetime
>>> from pontoon.base.models import User
>>> from pontoon.actionlog.models import ActionLog
>>>
>>> user = User.objects.get(email="matjaz.horvat@gmail.com")
>>> ActionLog.objects.filter(
    performed_by=user,
    action_type="translation:created",
    created_at__gte=timezone.make_aware(datetime(2024,10,1))
).count()
```

So I suggest we complerely avoid writing to the DB and don't use fields introduced in https://github.com/mozilla/pontoon/pull/3397.

harmitgoswami:
After some testing I found out the following:

Updates to a user's role always happen in the order of contributor -> translator -> manager. This means that if they get promoted, they're first removed from the old group then added to the new group. If they get demoted, however, they are very briefly in two groups simultaneously, then removed from the old group.

The fix I added in the latest commit now seems to properly account for this, but the code seems too fragile in my opinion, especially considering we'd need to rewrite the same logic again in the frontend for the profile page. 

Do you have any thoughts?

pontoon/base/models/user.py
```diff
@@ -216,6 +217,47 @@ def contributed_translations(self):
...
 
 
+@property
+def badges_translations_count(self):
+    """Contributions provided by user that count towards their badges."""
+    from pontoon.base.models.translation import Translation
+
+    badges_inception = aware_datetime(2024, 10, 18)
+
+    return Translation.objects.filter(user=self, date__gte=badges_inception).count()
```
mathjazz, line 227: Translations can be removed. We should use `ActionLog`:
https://github.com/mozilla/pontoon/pull/3413#discussion_r1805000532

Additionally, start day of the observation period should be defined in one place only, possibly in settings.

pontoon/translations/views.py
```diff
@@ -169,6 +173,14 @@ def create_translation(request):
...
             )
 
+    # Update Translation Champion Badge stats
+    if after_count < before_count:
+        badge_thresholds = [5, 50, 250, 1000]
+        for level, threshold in enumerate(badge_thresholds, start=1):
+            if after_count >= threshold and before_count < threshold:
+                # TODO: Send a notification to the user
+                print(f"New level achieved: Level {level}")
+
```
mathjazz, line 183: Why do we need to get the `after_count` from the DB, when we know it's `before_count + 1`? Additionally, comparison in line 176 should be inverted, otherwise condition in line 180 will never be `True`.

I think we can simplify this to the following:
```suggestion
    # Award Translation Champion Badge stats
    if user.badges_translations_count in [5, 50, 250, 1000]:
        # TODO: Send a notification to the user
        pass
```

Tresholds should be defined in one place only, possibly in settings.

pontoon/base/forms.py
```diff
@@ -97,11 +99,36 @@ def assign_users_to_groups(self, group_name, users):
...
+
         if users:
             group.user_set.add(*users)
 
         log_group_members(self.user, group, (add_users, remove_users))
 
+        after_count = self.user.badges_promoted_users
+
+        # Check if user was demoted from Manager to Translator
+        # In this case, it doesn't count as a promotion
```
mathjazz, line 113: Given that this code seems to be the only consumer of the `PermissionChangelog` model, we should refactor it to make retrieving promotions more straightforward. At least we should add a TODO explaining that and reference https://github.com/mozilla/pontoon/issues/2195.

harmitgoswami:
I've left in the check for if `after_count > before_count` inside of `assign_users_to_groups`, because the way I understand is the same function is also used for demotions. Thus, we should only check if the badge should be rewarded if the number of `ADDED` actions increased. Also, multiple users can be promoted/demoted in the same action.

Although let me know if you still think `before_count` is still unnecessary here, I can definitely still be misunderstanding.

pontoon/base/models/user.py
```diff
@@ -220,6 +224,49 @@ def has_approved_translations(self):
...
+        performed_by=self,
+        action_type="translation:created",
+        created_at__gte=badges_inception,
+    ).count()
+
+
+@property
+def badges_reviewed_translations(self):
+    """Translation reviews provided by user that count towards their badges."""
+    from pontoon.actionlog.models import ActionLog
```
mathjazz, line 242: `ActionLog` is already imported at the top.

pontoon/base/models/user.py
```diff
@@ -220,6 +224,49 @@ def has_approved_translations(self):
...
+    approvals = ActionLog.objects.filter(
+        performed_by=self,
+        action_type="translation:created",
+        created_at__gte=badges_inception,
+    ).count()
+    rejections = ActionLog.objects.filter(
+        performed_by=self,
+        action_type="translation:rejected",
+        created_at__gte=badges_inception,
+    ).count()
```
mathjazz, line 253: We should be able to turn this into a single DB query by performing an OR query:
https://djangocentral.com/how-to-perform-or-queries-in-django-orm/

And we should also use `self.actions...`.

pontoon/translations/views.py
```diff
@@ -21,6 +21,9 @@
 from pontoon.translations import forms
 
 
+badge_thresholds = [5, 50, 250, 1000]
```
mathjazz, line 24:
```suggestion
BADGE_THRESHOLDS = [5, 50, 250, 1000]
```

pontoon/base/models/user.py
```diff
@@ -12,6 +12,10 @@
 from django.utils import timezone
 
 from pontoon.actionlog.models import ActionLog
+from pontoon.base.utils import aware_datetime
+
+
+badges_inception = aware_datetime(2024, 10, 18)
```
mathjazz, line 18:
```suggestion
BADGE_START_DATE = aware_datetime(2024, 10, 18)
```

pontoon/base/models/user.py
```diff
@@ -445,6 +492,9 @@ def latest_action(self):
 User.add_to_class("locale_role", user_locale_role)
 User.add_to_class("status", user_status)
 User.add_to_class("contributed_translations", contributed_translations)
+User.add_to_class("badges_translations_count", badges_translations_count)
+User.add_to_class("badges_reviewed_translations", badges_reviewed_translations)
+User.add_to_class("badges_promoted_users", badges_promoted_users)
```
mathjazz, line 497: Nit: For consistency, I'd use these names:

```suggestion
User.add_to_class("badges_translation_count", badges_translation_count)
User.add_to_class("badges_review_count", badges_review_count)
User.add_to_class("badges_promotion_count", badges_promotion_count)
```

pontoon/base/models/user.py
```diff
@@ -220,6 +224,49 @@ def has_approved_translations(self):
...
+@property
+def badges_translations_count(self):
+    """Contributions provided by user that count towards their badges."""
+    from pontoon.actionlog.models import ActionLog
+
+    return ActionLog.objects.filter(
+        performed_by=self,
+        action_type="translation:created",
+        created_at__gte=badges_inception,
+    ).count()
```
mathjazz, line 236: I think this is clearer, since it's called on the `user` instance:

```suggestion
    return self.actions.filter(
        action_type="translation:created",
        created_at__gte=badges_inception,
    ).count()
```

pontoon/base/models/user.py
```diff
@@ -220,6 +224,49 @@ def has_approved_translations(self):
...
+@property
+def badges_promoted_users(self):
+    """Role promotions performed by user that count towards their badges"""
+    from pontoon.base.models import PermissionChangelog
+
+    return PermissionChangelog.objects.filter(
+        performed_by=self,
+        action_type=PermissionChangelog.ActionType.ADDED,
+        created_at__gte=badges_inception,
+    ).count()
```
mathjazz, line 267: I think this is clearer, since it's called on the `user` instance:

```suggestion
    return self.changed_permissions_log.filter(
        action_type=PermissionChangelog.ActionType.ADDED,
        created_at__gte=badges_inception,
    ).count()
```

mathjazz (commented):
Thanks for update! Left some more suggestions.

Please also remove the `print()` statements.

pontoon/base/models/user.py
```diff
@@ -220,6 +224,40 @@ def has_approved_translations(self):
...
+
+
+@property
+def badges_promotion_count(self):
+    """Role promotions performed by user that count towards their badges"""
+    from pontoon.base.models import PermissionChangelog
+
+    return self.changed_permissions_log.filter(
+        performed_by=self,
```
mathjazz, line 255: You can drop this line - it's already implied by `self` at the beginning of the expression:

```suggestion

```

pontoon/base/models/user.py
```diff
@@ -220,6 +224,40 @@ def has_approved_translations(self):
...
+        action_type=PermissionChangelog.ActionType.ADDED,
```
mathjazz, line 256: To be consistent with other `badge_*` properties above (and be able to drop the import), I'd use the string here:

```suggestion
        action_type="added",
```

pontoon/base/models/user.py
mathjazz: Also in the other two `badge_*` properties above.

pontoon/base/models/user.py
```diff
@@ -7,11 +7,15 @@
...
-from django.db.models import Count
+from django.db.models import Count, Q
 from django.urls import reverse
 from django.utils import timezone
 
 from pontoon.actionlog.models import ActionLog
+from pontoon.base.utils import aware_datetime
+
+
+BADGE_START_DATE = aware_datetime(2024, 10, 18)
```
mathjazz, line 18: We should move this variable to settings, so we don't need to deploy new code to change the date. The default value could be None and the code should be adjusted accordingly.

pontoon/base/models/user.py
```diff
@@ -220,6 +224,36 @@ def has_approved_translations(self):
     return self.translation_set.filter(approved=True).exists()
 
 
+@property
+def badges_translation_count(self):
+    """Contributions provided by user that count towards their badges."""
+
```
mathjazz, line 230: Nit: Please remove this blank line (and in the following two functions, too). :)

pontoon/base/models/user.py
```diff
@@ -220,6 +224,36 @@ def has_approved_translations(self):
...
+        created_at__gte=BADGE_START_DATE,
+    ).count()
+
+
+@property
+def badges_review_count(self):
+    """Translation reviews provided by user that count towards their badges."""
+
+    return self.actions.filter(
+        Q(action_type="translation:created") | Q(action_type="translation:rejected"),
```
mathjazz, line 242:
```suggestion
        Q(action_type="translation:approved") | Q(action_type="translation:rejected"),
```

harmitgoswami: Whoops! good catch, sorry about that

PR #3415, opened 2024-10-18
Improve `make shell`, add instructions for SSH in Docker

flodolo:
Currently, `make shell` spins up a brand-new container, which is not particularly useful. Also added a `make shell-root` option to connect as root.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ shell:
   		echo "Error: No container found based on local/pontoon" >&2; \
   		exit 1; \
 	else \
-  		${DOCKER} exec -it $$container /bin/bash; \
+  		DOCKER_CLI_HINTS="false" ${DOCKER} exec -it $$container /bin/bash; \
 	fi
 
 shell-root:
@@ -92,7 +92,7 @@ shell-root:
   		echo "Error: No container found based on local/pontoon" >&2; \
   		exit 1; \
 	else \
-  		${DOCKER} exec -u 0 -it $$container /bin/bash; \
+  		DOCKER_CLI_HINTS="false" ${DOCKER} exec -u 0 -it $$container /bin/bash; \
 	fi
 
 ci: test lint
--- a/Makefile
+++ b/Makefile
@@ -77,23 +77,26 @@ run: translate/dist .server-build
 clean:
 	rm -rf translate/dist .server-build
 
-shell:
+.run-container:
 	@container=$$(${DOCKER} ps -q --filter ancestor=local/pontoon | head -n 1); \
 	if [ -z "$$container" ]; then \
-  		echo "Error: No container found based on local/pontoon" >&2; \
-  		exit 1; \
-	else \
-  		DOCKER_CLI_HINTS="false" ${DOCKER} exec -it $$container /bin/bash; \
-	fi
+  		echo "Trying to start the container" >&2; \
+		"${DC}" up --detach; \
+  		container=$$(${DOCKER} ps -q --filter ancestor=local/pontoon | head -n 1); \
+		if [ -z "$$container" ]; then \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -77,7 +78,22 @@ clean:
 	rm -rf translate/dist .server-build
 
 shell:
-	"${DC}" run --rm server //bin/bash
+	@container=$$(${DOCKER} ps -q --filter ancestor=local/pontoon | head -n 1); \
+	if [ -z "$$container" ]; then \
+  		echo "Error: No container found based on local/pontoon" >&2; \
+  		exit 1; \
```
mathjazz, line 84: That means one needs to `make run` before `make shell`, which shouldn't be a requirement.

Makefile
```diff
@@ -77,7 +78,22 @@ clean:
...
+	else \
+  		${DOCKER} exec -it $$container /bin/bash; \
```
mathjazz, line 86: When I exit the shell with Ctrl+D, I get a promotional message:

```
What's next:
    Try Docker Debug for seamless, persistent debugging tools in any container or image → docker debug ccebe29e4262
    Learn more at https://docs.docker.com/go/debug-cli/
```

Could you please see if [`DOCKER_CLI_HINTS`](https://forums.docker.com/t/how-to-turn-off-the-whats-new-message/140860/6) indeed removes it?

docs/dev/setup.rst
```diff
@@ -77,6 +77,37 @@ The app should now be available at http://localhost:8000 or the custom SITE_URL.
 
 And with that, you're ready to start :doc:`contributing`!
 
+Writing to external repositories
```
mathjazz, line 80: In order to keep the setup instructions short and simple, I would move this section to the CONTRIBUTING file, which also documents how to restore DB backups, which seems to be from the same mold.

flodolo: Let me know if that's the right place in the doc.

Makefile
flodolo: Actually, that doesn't work 🤔

mathjazz: > What's the reason for running `make shell` now? I feel like nobody is using it, so maybe assuming that the container is already running is not excessive.

I use it **a lot**. Mostly to then also run `./make shell` (but not always).

flodolo: But that container is currently spun, then removed. I'm guessing you're running commands against the DB, or just checking other things, not interested in preserving those changes in the filesystem?

mathjazz: Correct.

We could either use a separate command (e.g. `make shell-server`) to connect to the existing container or fallback to the current behavior (i.e. spin up a new container) if the container is not running.

flodolo: > We could either use a separate command (e.g. `make shell-server`) to connect to the existing container or fallback to the current behavior (i.e. spin up a new container) if the server is not running.

Isn't that what the current code is doing? It only fails if there is no container to spin up.

Makefile
```diff
@@ -76,8 +77,26 @@ run: translate/dist .server-build
 clean:
 	rm -rf translate/dist .server-build
 
-shell:
-	"${DC}" run --rm server //bin/bash
+.run-container:
+	@container=$$(${DOCKER} ps -q --filter ancestor=local/pontoon | head -n 1); \
+	if [ -z "$$container" ]; then \
+  		echo "Trying to start the container" >&2; \
+		"${DC}" up --detach; \
```
flodolo, line 84: This is spinning up an existing container if there isn't one already. 
I've tested the code by manually stopping the container before running it.

Makefile
mathjazz: Yup, I didn't realize you already updated the code.

mathjazz (approved):

PR #3416, opened 2024-10-18
Do not throttle IPs of legitimate users

mathjazz:
Change IP throttling to never block legitimate users (identified by having at least 1 approved translation).

For blocked IPs we now also log the username if the request is coming from an authenticated user.

Log message is also changed a little so it's easier to search for all IPs with excessive requests (blocked and legitimate).

### DIFF
--- a/pontoon/base/models/user.py
+++ b/pontoon/base/models/user.py
@@ -210,10 +210,14 @@ def user_status(self, locale):
 
 @property
 def contributed_translations(self):
-    """Filtered contributions provided by user."""
-    from pontoon.base.models.translation import Translation
+    """Contributions provided by user."""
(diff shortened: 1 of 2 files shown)


pontoon/base/models/user.py
```diff
@@ -210,10 +210,14 @@ def user_status(self, locale):
 
 @property
 def contributed_translations(self):
-    """Filtered contributions provided by user."""
-    from pontoon.base.models.translation import Translation
+    """Contributions provided by user."""
+    return self.translation_set.all()
 
-    return Translation.objects.filter(user=self)
```
mathjazz, line 216: Not directly related to the patch, but using similar pattern as the newly added code.

PR #3420, opened 2024-10-21
Release Messaging center

### LINKED ISSUE #3162: Create a Messaging Center for sending emails and in-app notifications
Managers of Pontoon projects need a more robust way of sending targeted communication to Pontoon users. A messaging center feature implemented within Pontoon would enable project managers to send emails or in-app messages to a subset of users (based on criteria such as user role, locale, project, etc.) without needing an engineer.

mathjazz:
Fix #3162.

This is the final step of functional changes to the Messaging center before we can start using it in production:
* Messaging link is added to the main menu, visible only to users that can actually access the page
* As soon as you hit the Review page, contributors start getting fetched, as indicated in the button in the bottom right part of the screen. 
* Once fetching is complete, the button to send the messages is made available with the recipient count written over it. 
* While messages are being sent or if the recipient count is 0, the Send button is disabled. 
* If fetching contributors fails, the button turns red and offers you the ability to retry fetching.

Note that you can only send notifications on stage, sending emails will fail. Also note that neither fetching nor sending is a background task, because testing on stage has shown that splitting fetching and sending in separate steps no longer results in request timeouts.

### DIFF
--- a/pontoon/messaging/static/js/messaging.js
+++ b/pontoon/messaging/static/js/messaging.js
@@ -74,6 +74,30 @@ $(function () {
     );
   }
 
+  function fetchRecipients() {
+    $('#review .controls .fetching').show();
+    $('#review .controls .fetch-again').hide();
+    $('#review .controls .send.active').hide().find('.value').html('');
+
+    $.ajax({
+      url: '/messaging/ajax/fetch-recipients/',
+      type: 'POST',
+      data: $('#send-message').serialize(),
+      success: function (data) {
+        const count = nf.format(data.recipients.length);
+        $('#review .controls .send.active').show().find('.value').html(count);
+        $('#compose [name=recipient_ids]').val(data.recipients);
+      },
+      error: function () {
+        Pontoon.endLoader('Fetching recipients failed.', 'error');
+        $('#review .controls .fetch-again').show();
+      },
+      complete: function () {
+        $('#review .controls .fetching').hide();
+      },
+    });
+  }
+
   function updateReviewPanel() {
     function updateMultipleItemSelector(source, target, item) {
       const allProjects = !$(`${source}.available li:not(.no-match)`).length;
@@ -116,27 +140,7 @@ $(function () {
     const html = converter.makeHtml(bodyValue);
     $('#compose [name=body]').val(html);
 
-    // Fetch recipients
-    $('#review .controls .fetching').show();
-    $('#review .controls .error').hide();
-    $('#review .controls .send.active').hide().find('.value').html('');
-
-    $.ajax({
-      url: '/messaging/ajax/fetch-recipients/',
-      type: 'POST',
-      data: $('#send-message').serialize(),
-      success: function (data) {
-        const count = nf.format(data.recipients.length);
-        $('#review .controls .send.active').show().find('.value').html(count);
-        $('#compose [name=recipient_ids]').val(data.recipients);
-      },
-      error: function () {
-        $('#review .controls .error').show();
-      },
-      complete: function () {
-        $('#review .controls .fetching').hide();
-      },
-    });
+    fetchRecipients();
 
     // Subject
     $('#review .subject .value').html($('#id_subject').val());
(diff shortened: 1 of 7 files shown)


pontoon/messaging/static/js/messaging.js
```diff
@@ -309,12 +362,19 @@ $(function () {
       success: function () {
         Pontoon.endLoader('Message sent.');
         if (!sendToMyself) {
-          container.find('.left-column .sent a').click();
+          const count = container.find('.left-column .sent .count');
+          // Update count in the menu
+          count.html(parseInt(count.html(), 10) + 1);
```
eemeli, line 367: You'll need to store the non-formatted serialization e.g. in a `data-count` field, because this will fail for values over a thousand:
```js
new Intl.NumberFormat('en').format(1000) === '1,000'
parseInt('1,000', 10) === 1
```

pontoon/messaging/views.py
```diff
@@ -196,22 +206,38 @@ def get_recipients(form):
...
+        or translation_maximum is not None
+    ):
+        submission_filters = list(submitted.values_list("user", flat=True).distinct())
+        recipients = recipients.filter(pk__in=submission_filters)
+
+    if review_from or review_to or review_minimum or review_maximum is not None:
+        review_filters = list(
+            approved.values_list("approved_user", flat=True).distinct()
+        ) + list(rejected.values_list("rejected_user", flat=True).distinct())
+        recipients = recipients.filter(pk__in=review_filters)
```
eemeli, line 240: Why listify the `submission_filters` and `review_filters`? Leaving them as just queries should also work?

pontoon/messaging/views.py
```diff
@@ -257,10 +303,9 @@ def send_message(request):
                 description=f"{subject}<br/><br/>{body}",
                 identifier=identifier,
             )
+            notification_recipients_count += 1
 
-        log.info(
-            f"Notifications sent to the following {recipients.count()} users: {recipients.values_list('email', flat=True)}."
-        )
+        log.info(f"Notifications sent to {notification_recipients_count} users.")
```
eemeli, line 308: Why not just get the length directly?
```suggestion
        log.info(f"Notifications sent to {len(recipients)} users.")
```

pontoon/messaging/views.py
```diff
@@ -288,10 +335,9 @@ def send_message(request):
             )
             msg.attach_alternative(html, "text/html")
             msg.send()
+            email_recipients_count += 1
 
-        log.info(
-            f"Email sent to the following {email_recipients.count()} users: {email_recipients.values_list('email', flat=True)}."
-        )
+        log.info(f"Emails sent to {email_recipients_count} users.")
```
eemeli, line 340: Same as above.
```suggestion
        log.info(f"Emails sent to {len(recipients)} users.")
```

eemeli (changes requested):

pontoon/messaging/static/js/messaging.js
mathjazz: The number is not formatted, so it'll be:
```js
parseInt('1000', 10) === 1000
```

pontoon/messaging/views.py
mathjazz: Updated. We should test if this affects performance.

pontoon/messaging/views.py
mathjazz: 🤦

pontoon/messaging/static/js/messaging.js
eemeli: This is a different count? https://github.com/mozilla/pontoon/blob/9ae8f44bd2bde67a8eb04b320a23c36c9c9c85c8/pontoon/messaging/static/js/messaging.js#L91-L96

mathjazz: Yeah, this is the recipient count. The one you commented on originally is the sent message count.

eemeli (approved):

PR #3421, opened 2024-10-21
Check if users are managers and translators first, then admins (frontend code)

mathjazz:
This is a fix for an inconsistency I caused in #3414, which only changes the order in which user statuses are assigned in the backend code.


mathjazz:
Closing on behalf of #3422, which already contains the same fix.

PR #3422, opened 2024-10-21
Add banner for project managers, give priority to roles within locale

### LINKED ISSUE #3418: Unify admin user status definition
Admin user status introduced in https://github.com/mozilla/pontoon/pull/3400 is defined in two different ways: 

1. As a [superuser](https://github.com/mozilla/pontoon/commit/3910a994f7a07ea3d530a49879355909841d80e8#diff-3e822354a9d9b515f80ad2940a7fb3685084a1ec090224d30fc5ede87f4b0174R200-R201) (user with all permissions).

2. As a [user with access to Pontoon admin](https://github.com/mozilla/pontoon/blob/3910a994f7a07ea3d530a49879355909841d80e8/pontoon/base/views.py#L876) (these users can also be selected from the list of Project Managers).

We should unify that and only use one definition. Which one?

I think the latter, because these users add comments and fix translations, so their avatars will actually appear in the History and Comments panels.

They will also be listed as Project Managers in the Project dashboards. So we should also change the status name from Admin to PM.

### LINKED ISSUE #3425: "NEW" banner should not be displayed for System users (like Google Translate)
Screenshot 2024-10-24 alle 14 54 11

It won't happen on the official Pontoon instance, but it can happen on a new one.

flodolo:
This adds a banner for users defined as "Project Manager" within a project. To reduce the space for confusion, I changed the tooltip of MNGR from "Manager" to "Team Manager".

This also makes the priority of roles consistent between front-end and back-end:
- If a user as a role within the locale (translator, manager), we use that for the banner.
- If a user is set as PM, we use that even if the user is an Admin.

Finally, this adds CSS variables for users, instead of reusing the ones for translation status.

Fixes #3418 
Fixes #3425

TODO:
- [ ] Update spec

### DIFF
--- a/pontoon/base/models/comment.py
+++ b/pontoon/base/models/comment.py
@@ -28,10 +28,15 @@ class Comment(models.Model):
 
     def serialize(self):
         locale = self.locale or self.translation.locale
+        project = (
+            self.translation.entity.resource.project
+            if self.translation
+            else self.entity.resource.project
+        )
         return {
             "author": self.author.name_or_email,
             "username": self.author.username,
-            "user_status": self.author.status(locale),
+            "user_status": self.author.status(locale, project),
             "user_gravatar_url_small": self.author.gravatar_url(88),
             "created_at": self.timestamp.strftime("%b %d, %Y %H:%M"),
             "date_iso": self.timestamp.isoformat(),
--- a/pontoon/base/models/comment.py
+++ b/pontoon/base/models/comment.py
@@ -26,17 +26,12 @@ class Comment(models.Model):
     def __str__(self):
         return self.content
 
-    def serialize(self):
+    def serialize(self, project_contact):
         locale = self.locale or self.translation.locale
-        project = (
-            self.translation.entity.resource.project
-            if self.translation
-            else self.entity.resource.project
-        )
         return {
             "author": self.author.name_or_email,
             "username": self.author.username,
-            "user_status": self.author.status(locale, project),
+            "user_status": self.author.status(locale, project_contact),
             "user_gravatar_url_small": self.author.gravatar_url(88),
             "created_at": self.timestamp.strftime("%b %d, %Y %H:%M"),
             "date_iso": self.timestamp.isoformat(),
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -886,7 +886,7 @@ def user_data(request):
             "manager_for_locales": list(
                 user.managed_locales.values_list("code", flat=True)
             ),
-            "pm_for_projects": user.pm_for_projects,
+            "pm_for_projects": user.contact_for.all(),
             "translator_for_locales": list(
                 user.translated_locales.values_list("code", flat=True)
             ),
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -893,8 +893,8 @@ def user_data(request):
             "can_translate_locales": list(
                 user.can_translate_locales.values_list("code", flat=True)
             ),
-            "manager_for_locales": [l.code for l in user.manager_for_locales],
-            "translator_for_locales": [l.code for l in user.translator_for_locales],
+            "manager_for_locales": [loc.code for loc in user.manager_for_locales],
+            "translator_for_locales": [loc.code for loc in user.translator_for_locales],
             "pm_for_projects": list(user.contact_for.values_list("slug", flat=True)),
             "translator_for_projects": user.translated_projects,
             "settings": {
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -889,9 +889,7 @@ def user_data(request):
             "manager_for_locales": list(
                 user.managed_locales.values_list("code", flat=True)
             ),
-            "pm_for_projects": list(
-                user.contact_for.values_list("slug", flat=True)
-            ),
+            "pm_for_projects": list(user.contact_for.values_list("slug", flat=True)),
             "translator_for_locales": list(
                 user.translated_locales.values_list("code", flat=True)
             ),
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -889,7 +889,9 @@ def user_data(request):
             "manager_for_locales": list(
                 user.managed_locales.values_list("code", flat=True)
             ),
-            "pm_for_projects": [p.slug for p in user.contact_for.all()],
+            "pm_for_projects": list(
+                user.contact_for.values_list("slug", flat=True)
+            ),
             "translator_for_locales": list(
                 user.translated_locales.values_list("code", flat=True)
             ),
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -430,7 +430,7 @@ def get_translation_history(request):
                 "uid": u.id,
                 "username": u.username,
                 "user_gravatar_url_small": u.gravatar_url(88),
-                "user_status": u.status(locale),
+                "user_status": u.status(locale, t.entity.resource.project),
                 "date": t.date,
                 "approved_user": User.display_name_or_blank(t.approved_user),
                 "approved_date": t.approved_date,
@@ -886,6 +886,7 @@ def user_data(request):
             "manager_for_locales": list(
                 user.managed_locales.values_list("code", flat=True)
             ),
+            "pm_for_projects": user.pm_for_projects,
             "translator_for_locales": list(
                 user.translated_locales.values_list("code", flat=True)
             ),
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -886,7 +886,7 @@ def user_data(request):
             "manager_for_locales": list(
                 user.managed_locales.values_list("code", flat=True)
             ),
-            "pm_for_projects": user.contact_for.all(),
+            "pm_for_projects": [p.slug for p in user.contact_for.all()],
             "translator_for_locales": list(
                 user.translated_locales.values_list("code", flat=True)
             ),
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -893,6 +893,8 @@ def user_data(request):
             "can_translate_locales": list(
                 user.can_translate_locales.values_list("code", flat=True)
             ),
+            "manager_for_locales": [l.code for l in user.manager_for_locales],
+            "translator_for_locales": [l.code for l in user.translator_for_locales],
             "pm_for_projects": list(user.contact_for.values_list("slug", flat=True)),
             "translator_for_projects": user.translated_projects,
             "settings": {
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -399,6 +399,7 @@ def get_translation_history(request):
 
     entity = get_object_or_404(Entity, pk=entity)
     locale = get_object_or_404(Locale, code=locale)
+    project_contact = entity.resource.project.contact
 
     translations = Translation.objects.filter(
         entity=entity,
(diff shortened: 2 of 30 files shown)


flodolo:
Follow-up on conversation: I tested with `with debug_sql():`, and I'm seeing the same number of db queries with `main` and this code?

* 1 translation comment: 6 db calls
* 1 source comment: 10 db calls

flodolo:
Deleted the previous comment because I realize I was testing the same code 🤦🏼 

Follow-up on conversation, I tested with `with debug_sql():`, if I did it correctly:

* 1 translation comment: new code 6 db calls, previous 5.
* 1 source comment: 10 db calls, previous 6.

flodolo:
The second commit fixes #3425

pontoon/base/models/comment.py
```diff
@@ -28,10 +28,15 @@ def __str__(self):
 
     def serialize(self):
         locale = self.locale or self.translation.locale
+        project = (
+            self.translation.entity.resource.project
+            if self.translation
+            else self.entity.resource.project
+        )
```
mathjazz, line 35: This is not very pretty and not very fast. Similarly, the path to get to `project` in [pontoon/base/views.py](https://github.com/mozilla/pontoon/pull/3422/files#diff-f4cdb2e09580f4e478d1804c3b86f26eed7e43f1f2fb859e83d326074c52d560), while a bit prettier, is still slow.

There are ways to get to the project object in a more scalable way, e.g. by passing the project from frontend all the way to here or by passing `contact = entity.resource.project.contact` instead.

My personal preference would be to just use `user.groups.filter(name="project_managers").exists()` in `user_status()`, which would show all PMs, regardless of the project, as PMs. 

I think that's even better, because when PMs of other projects add translations or comments, they are effectively acting PMs (e.g. because they step in temporarily or because they have managed the project in the past).

mathjazz (commented):
Nice work!

Added a few suggestions for minor improvements, and suggested a couple of ways forward for the PM status.

pontoon/base/models/comment.py
flodolo: > My personal preference would be to just use user.groups.filter(name="project_managers").exists() in user_status(), which would show all PMs, regardless of the project, as PMs.

That kind of defies the main point of having the PM banner. 

I want the PM banner to show up only within the project for which that user is set as PM, not in all of them.

An admin that is also a PM should show up as an Admin. A person that is not an Admin, but is set as a PM for a specific project, shouldn't be displayed as a PM for any other projects (sure, they're unlikely to do anything outside of their project, but we should still avoid presenting them as something they're not).

Regarding this being slow: can we live with it, considering it's only used when there are actually comments (so, not in the majority of cases)? We have 12k comments for almost 8M translations, and I bet most comments are in group of at least 2 (so, ballpark of 4–5k translations affected? I'm sure actual data can be extracted if we think it's important).

flodolo: > There are ways to get to the project object in a more scalable way, e.g. by passing the project from frontend all the way to here or by passing contact = entity.resource.project.contact instead.

I'd be happy to have someone else pick up this PR and pursue this path, because I'm not sure that's something I can do with my current understanding of the codebase 😉

flodolo: (investigating the `passing contact = entity.resource.project.contact instead` in the meantime)

flodolo:
My previous version of the code was running without issues, with the last commit (review comments) I see a weird error when I load the first time?

```
[server] [ERROR:django.request] 2024-10-25 13:34:52,063 Internal Server Error: /user-data/
[server] Traceback (most recent call last):
[server]   File "/usr/local/lib/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner
[server]     response = get_response(request)
[server]                ^^^^^^^^^^^^^^^^^^^^^
[server]   File "/usr/local/lib/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response
[server]     response = wrapped_callback(request, *callback_args, **callback_kwargs)
[server]                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[server]   File "/app/pontoon/base/utils.py", line 139, in wrap
[server]     return f(request, *args, **kwargs)
[server]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[server]   File "/app/pontoon/base/views.py", line 873, in user_data
[server]     return JsonResponse(
[server]            ^^^^^^^^^^^^^
[server]   File "/usr/local/lib/python3.11/site-packages/django/http/response.py", line 731, in __init__
[server]     data = json.dumps(data, cls=encoder, **json_dumps_params)
[server]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[server]   File "/usr/local/lib/python3.11/json/__init__.py", line 238, in dumps
[server]     **kw).encode(obj)
[server]           ^^^^^^^^^^^
[server]   File "/usr/local/lib/python3.11/json/encoder.py", line 200, in encode
[server]     chunks = self.iterencode(o, _one_shot=True)
[server]              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[server]   File "/usr/local/lib/python3.11/json/encoder.py", line 258, in iterencode
[server]     return _iterencode(o, 0)
[server]            ^^^^^^^^^^^^^^^^^
[server]   File "/usr/local/lib/python3.11/site-packages/django/core/serializers/json.py", line 106, in default
[server]     return super().default(o)
[server]            ^^^^^^^^^^^^^^^^^^
[server]   File "/usr/local/lib/python3.11/json/encoder.py", line 180, in default
[server]     raise TypeError(f'Object of type {o.__class__.__name__} '
[server] TypeError: Object of type ProjectQuerySet is not JSON serializable
[server] [ERROR:django.server] 2024-10-25 13:34:52,066 "GET /user-data/ HTTP/1.1" 500 22414
```

I also tried to move `project` to the caller code, but that seems to be more complicated than I expected? Does that require a migration?

```diff
diff --git a/pontoon/base/migrations/0067_comment_project.py b/pontoon/base/migrations/0067_comment_project.py
new file mode 100644
index 00000000..ee28506c
--- /dev/null
+++ b/pontoon/base/migrations/0067_comment_project.py
@@ -0,0 +1,19 @@
+# Generated by Django 4.2.16 on 2024-10-25 13:22
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('base', '0066_userprofile_community_builder_level_and_more'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='comment',
+            name='project',
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, related_name='comments', to='base.project'),
+        ),
+    ]
diff --git a/pontoon/base/models/comment.py b/pontoon/base/models/comment.py
index 474a7ab3..1e1e05d3 100644
--- a/pontoon/base/models/comment.py
+++ b/pontoon/base/models/comment.py
@@ -20,6 +20,9 @@ class Comment(models.Model):
     entity = models.ForeignKey(
         "Entity", models.CASCADE, related_name="comments", blank=True, null=True
     )
+    project = models.ForeignKey(
+        "Project", models.CASCADE, related_name="comments", blank=True, null=True
+    )
     content = models.TextField()
     pinned = models.BooleanField(default=False)
 
@@ -28,15 +31,10 @@ class Comment(models.Model):
 
     def serialize(self):
         locale = self.locale or self.translation.locale
-        project = (
-            self.translation.entity.resource.project
-            if self.translation
-            else self.entity.resource.project
-        )
         return {
             "author": self.author.name_or_email,
             "username": self.author.username,
-            "user_status": self.author.status(locale, project),
+            "user_status": self.author.status(locale, self.project),
             "user_gravatar_url_small": self.author.gravatar_url(88),
             "created_at": self.timestamp.strftime("%b %d, %Y %H:%M"),
             "date_iso": self.timestamp.isoformat(),
diff --git a/pontoon/base/models/user.py b/pontoon/base/models/user.py
index 09c22a0c..c2068521 100644
--- a/pontoon/base/models/user.py
+++ b/pontoon/base/models/user.py
@@ -201,7 +201,7 @@ def user_status(self, locale, project):
         return ("MNGR", "Team Manager")
     if self in locale.translators_group.user_set.all():
         return ("TRNSL", "Translator")
-    if project.contact and self.username == project.contact.username:
+    if project and self.username == project.contact.username:
         return ("PM", "Project Manager")
     if self.is_superuser:
         return ("ADMIN", "Admin")
diff --git a/pontoon/base/views.py b/pontoon/base/views.py
index 5864830b..03c513c5 100755
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -630,12 +630,23 @@ def add_comment(request):
 
     # Translation comment
     if translation:
-        c = Comment(author=user, translation=translation, content=comment)
+        c = Comment(
+            author=user,
+            translation=translation,
+            content=comment,
+            project=translation.entity.resource.project,
+        )
         log_action(ActionLog.ActionType.COMMENT_ADDED, user, translation=translation)
 
     # Team comment
     else:
-        c = Comment(author=user, entity=entity, locale=locale, content=comment)
+        c = Comment(
+            author=user,
+            entity=entity,
+            locale=locale,
+            content=comment,
+            project=entity.resource.project,
+        )
         log_action(
             ActionLog.ActionType.COMMENT_ADDED, user, entity=entity, locale=locale
         )
```

pontoon/base/views.py
```diff
@@ -886,6 +886,7 @@ def user_data(request):
             "manager_for_locales": list(
                 user.managed_locales.values_list("code", flat=True)
             ),
+            "pm_for_projects": user.contact_for.all(),
```
flodolo, line 889: This returns a ProjectQuerySet. Not sure if there's a better way to get a list of slugs besides list comprehension.

pontoon/base/views.py
```diff
@@ -430,7 +430,7 @@ def get_translation_history(request):
                 "uid": u.id,
                 "username": u.username,
                 "user_gravatar_url_small": u.gravatar_url(88),
-                "user_status": u.status(locale),
+                "user_status": u.status(locale, t.entity.resource.project),
```
mathjazz, line 433: This makes 3 additional DB queries for each translation loaded in the history panel.

pontoon/base/models/comment.py
mathjazz: I'll push updates to your branch directly.

Performance degrades for every string that has translations. The PR makes 3 additional DB queries in [pontoon/base/views.py](https://github.com/mozilla/pontoon/pull/3422/files#diff-f4cdb2e09580f4e478d1804c3b86f26eed7e43f1f2fb859e83d326074c52d560) for each translation loaded in the History panel if translation **doesn't** have any comments. 4 DB queries are made here **on top of that** for translations with comments, for a total of 7 additional DB queries. We should avoid that.

pontoon/base/views.py
mathjazz: Yikes!

mathjazz:
@flodolo Could you please take it for a spin on stage?

In addition to addressing the review comments, I've also fix the color for the MNGR status, which was broken and simplified the logic behind it.

flodolo (commented):
I think we have a problem with admins in the front-end: they always show up as "MNGR" when writing a comment, because apparently `managerForLocales` is the full list of Pontoon locales (same for `translatorForLocales`). 

Is that needed to enable/disable editing?

flodolo:
(I'm testing locally because it's easier to mess with projects and users)

Another inconsistency: new user that is not set up as PM for the project, but it's part of project_managers. When commenting (front-end) it shows up as ADMIN, after commenting it shows up as NEW.

mathjazz:
Yeah, we have inconsistent definitions of user statuses between:
https://github.com/mozilla/pontoon/blob/da0c8ec105289c5d31ddfc9e46ed7773438c0b05/pontoon/base/views.py#L860 (used when adding a comment)

https://github.com/mozilla/pontoon/blob/da0c8ec105289c5d31ddfc9e46ed7773438c0b05/pontoon/base/models/user.py#L197 (used everywhere else)

mathjazz:
Stage updated with the new code.

flodolo:
This works great for me locally. Thanks for taking over, so much larger than I expected

mathjazz:
Could you please also take a look at the code? I can't flag you for a review, since you're the author of the PR.

flodolo:
> Could you please also take a look at the code? I can't flag you for a review, since you're the author of the PR.

I should have called out that I did review the various commits 👍🏼

mathjazz (approved):

PR #3428, opened 2024-10-25
Added documentation for new badge env variables

harmitgoswami:
Title

### DIFF
--- a/docs/admin/deployment.rst
+++ b/docs/admin/deployment.rst
@@ -73,15 +73,15 @@ you create:
    Set to 'google' if you want to use 'Google' (corresponding GOOGLE_* settings must be set).
 
 ``BADGES_PROMOTION_THRESHOLDS``
-   A comma-separated list of numeric thresholds for different levels of the 
+   Optional. A comma-separated list of numeric thresholds for different levels of the 
    Community Builder badge.
 
 ``BADGES_START_DATE``
-   Specifies the start date from which user activities count towards badge achievements. 
+   Optional. Specifies the start date from which user activities count towards badge achievements. 
    This variable should be in YYYY-MM-DD format.
 
 ``BADGES_TRANSLATION_THRESHOLDS``
-   A comma-separated list of numeric thresholds for different levels of the 
+   Optional. A comma-separated list of numeric thresholds for different levels of the 
    Review Master and Translation Champion badges.
 
 ``BLOCKED_IPS``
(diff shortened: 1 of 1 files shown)


mathjazz (commented):
LGTM. The only thing I'd change is that the variables are not really required.

I'm also thinking that we should probably have separate lists for Review Master and Translation Champion lists, but that's a separate issue.

harmitgoswami:
Just out of curiosity, when is a variable considered to be required?

mathjazz:
> Just out of curiosity, when is a variable considered to be required?

When the app breaks if it's not set.

All variables are required, unless the descriptions says they are optional. So please start the descriptions with `Optional.`. Sorry for not making it clearer earlier.

mathjazz (approved):

docs/admin/deployment.rst
```diff
@@ -72,6 +72,18 @@ you create:
    Set to 'gitlab' if you want to use 'GitLab' (corresponding GITLAB_* settings must be set if required).
    Set to 'google' if you want to use 'Google' (corresponding GOOGLE_* settings must be set).
 
+``BADGES_PROMOTION_THRESHOLDS``
+   Optional. A comma-separated list of numeric thresholds for different levels of the
```
flodolo, line 76: I'm not sure these explanations are clear without looking at the code.

For each of these options: 
* Do we support any number of levels? 
* Providing an example would also help, maybe with the default value.

docs/admin/deployment.rst
```diff
@@ -72,6 +72,18 @@ you create:
...
 
+``BADGES_PROMOTION_THRESHOLDS``
+   Optional. A comma-separated list of numeric thresholds for different levels of the 
+   Community Builder badge.
+
+``BADGES_START_DATE``
+   Optional. Specifies the start date from which user activities count towards badge achievements. 
+   This variable should be in YYYY-MM-DD format.
+
+``BADGES_TRANSLATION_THRESHOLDS``
```
flodolo, line 83: Question for a potential follow-up: I'm not sure we should have one setting shared between two different badges.
The numbers happen to match in our proposal, but now we're enforcing that for anyone wanting to use badges.

flodolo: Somehow missed this comment

> I'm also thinking that we should probably have separate lists for Review Master and Translation Champion lists, but that's a separate issue.

PR #3442, opened 2024-11-15
Add notifications when a badge is awarded

### LINKED ISSUE #3412: Send notification upon user receiving new badge
Offshoot from #3396 

From discussions with the team, we decided it would be nice to have an animation play on the user's screen when they are awarded a badge, rather than just a simple notification or banner.

Due to the added complexity, it's probably better to tackle the notifications in a separate issue.

harmitgoswami:
Fix #3412 

This PR adds notifications to signal to a user that they received a new badge/badge level. Notifications will appear in both the "Notifications Menu" in the header (first image), and as a notification banner across the top of the page. Notification banners for new badge levels will trigger a confetti animation, imported from https://github.com/catdad/canvas-confetti.

The initial commit for this PR has a noticeable bug, in that demotions and promotions will both count towards the Community Builder Badge. This issue stems from how `assign_users_to_groups` will always update translators then managers, in that order. Thus, there is no way to distinguish if an action is promoting a user from Contributor -> Translator or if an action is demoting a user from Manager -> Translator, especially since multiple users can be demoted/promoted in the same action.

**TODO**: 

- [x] With the newly added messaging center, all templates and logic for handling notifications should be moved into `pontoon.messaging` (see [this comment](https://github.com/mozilla/pontoon/pull/3442#discussion_r1865853616) below). This includes moving `_send_badge_notification` as well. We should remember to file an issue for this before this PR is merged.

image

image


mathjazz:
> The initial commit for this PR has a noticeable bug, in that demotions and promotions will both count towards the Community Builder Badge. This issue stems from how `assign_users_to_groups` will always update translators then managers, in that order. Thus, there is no way to distinguish if an action is promoting a user from Contributor -> Translator or if an action is demoting a user from Manager -> Translator, especially since multiple users can be demoted/promoted in the same action.

Is this not working as expected then?

https://github.com/mozilla/pontoon/blob/8a16a6568601c40551b64922fc0e310194f99885/pontoon/base/forms.py#L125-L136

pontoon/settings/base.py
```diff
@@ -655,6 +655,7 @@ def _default_from_email():
         "source_filenames": (
             "js/lib/chart.umd.min.js",
             "js/lib/chartjs-adapter-date-fns.bundle.min.js",
+            "js/lib/confetti.browser.min.js",
```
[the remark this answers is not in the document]
mathjazz: Please use this file:
https://cdn.jsdelivr.net/npm/canvas-confetti@1.9.3/dist/confetti.browser.js

(Removed `.min` from the link [here](https://github.com/catdad/canvas-confetti?tab=readme-ov-file#install)).

mathjazz:
> Is this not working as expected then?

This patch fixes the `badges_promotion_count` to properly exclude false positives and uses it when permissions get saved:
https://gist.github.com/mathjazz/bc5a9fa6359179fa1677c43b4d8f2080

pontoon/teams/static/js/permissions.js
```diff
@@ -50,8 +50,64 @@ $(function () {
...
+              .find('.controls:nth-child(2)')
+              .text(
+                `Community Builder Badge level gained: Level ${badgeLevel}`,
+              );
+
+            tooltip.addEventListener('click', function hideTooltip() {
+              tooltip.style.display = 'none';
+              tooltip.removeEventListener('click', hideTooltip);
+            });
+          }
```
mathjazz, line 75: It seems unusal not to use jQuery here.

pontoon/teams/templates/teams/includes/permissions.html
```diff
@@ -3,9 +3,15 @@
 <form id="permissions-form" method="POST" action="{{ url('pontoon.teams.ajax.permissions', locale.code) }}">
     <input type="hidden" value="{{ csrf_token }}" name="csrfmiddlewaretoken">
 
-    {% if community_builder_badge %}
-        <input type="hidden" id="community-builder-level" value="{{ community_builder_badge }}">
-    {% endif %}
+    <input type="hidden" id="community-builder-level" value="{{ community_builder_badge }}">
+    <div id="badge-tooltip-container" class="badge-tooltip" style="display: none;">
```
mathjazz, line 7: Please use CSS instead of inline styling.

mathjazz: We should also include the actual badge in the popup.

pontoon/teams/views.py
```diff
@@ -251,6 +251,7 @@ def ajax_permissions(request, locale):
             "project_locales": project_locales,
             "hide_project_selector": hide_project_selector,
             "community_builder_badge": community_builder_level,
+            "username": request.user,
```
mathjazz, line 254: The `request` object is passed to the templates via `pontoon.base.context_processor`, so no need to pass it again here.

translate/src/modules/notification/components/BadgeTooltip.css
```diff
@@ -0,0 +1,32 @@
...
+.badge-tooltip.showing {
+  top: 0;
+}
+
+.badge-tooltip button {
+  background: none;
+  border: none;
+  color: var(--tooltip-color-2);
+  font-weight: 300;
+}
```
mathjazz, line 27: Eemeli would suggest that we use CSS nesting in the new files: 😄
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_nesting/Using_CSS_nesting

mathjazz (commented):
Nice work! I didn't actually test the code yet, but left some notes on things that stood out.

It seems like the badge popup HTML and CSS are not the same in preference page and the translate view?

pontoon/teams/static/js/permissions.js
```diff
@@ -56,21 +56,21 @@ $(function () {
...
-          if (tooltip) {
-            tooltip.style.display = 'block';
+          if ($tooltip.length) {
+            $tooltip.show();
 
             // Force a re-render of the text with the proper badge level
-            $(tooltip)
-              .find('.controls:nth-child(2)')
+            $tooltip
+              .find('p:nth-of-type(2)')
```
mathjazz, line 66: This looks brittle. Can we add a class to the element and reference it by it?

pontoon/teams/static/js/permissions.js
```diff
@@ -50,8 +50,64 @@ $(function () {
...
-      success: function () {
+      success: function (response) {
         Pontoon.endLoader('Permissions saved.');
+        const badgeLevel = $(response).find('#community-builder-level').val();
+
+        // Check for new badge notification
+        if (badgeLevel > 0) {
+          const $tooltip = $('#badge-tooltip-container');
+
+          if ($tooltip.length) {
```
mathjazz, line 61: Why do we need this check?

pontoon/base/forms.py
```diff
@@ -91,6 +92,8 @@ class UserPermissionLogFormMixin:
     def __init__(self, *args, **kwargs):
         self.user = kwargs.pop("user")
         super().__init__(*args, **kwargs)
+        # Track if user reached new level for Community Builder Badge
+        self.community_builder_level = 0
```
mathjazz, line 96: If we're only setting this variable when the new level has been reached, we should reflect that in the variable name, otherwise it becomes confusing, especially on frontend.

For example:
```suggestion
        self.new_community_builder_level = 0
```

Or:
```suggestion
        self.community_builder_level_reached = 0
```

pontoon/settings/base.py
```diff
@@ -789,10 +790,12 @@ def _default_from_email():
     "django.contrib.staticfiles.finders.FileSystemFinder",
     "django.contrib.staticfiles.finders.AppDirectoriesFinder",
 )
-STATICFILES_DIRS = [
+
+STATICFILES_DIRS = (
+    ("badges", "pontoon/contributors/static/img"),
```
mathjazz, line 795: Why do we need this change?

pontoon/teams/static/js/permissions.js
```diff
@@ -50,8 +50,64 @@ $(function () {
...
+            $tooltip
+              .find('p:nth-of-type(2)')
+              .text(
+                `Community Builder Badge level gained: Level ${badgeLevel}`,
+              );
+
+            $tooltip.find('button').one('click', function (e) {
+              e.preventDefault();
+              $tooltip.hide();
+            });
```
mathjazz, line 74: I'd move the event handler outside the AJAX function.

pontoon/teams/templates/teams/includes/permissions.html
```diff
@@ -3,6 +3,17 @@
 <form id="permissions-form" method="POST" action="{{ url('pontoon.teams.ajax.permissions', locale.code) }}">
     <input type="hidden" value="{{ csrf_token }}" name="csrfmiddlewaretoken">
 
+    <input type="hidden" id="community-builder-level" value="{{ community_builder_badge }}">
+    <div id="badge-tooltip-container" class="badge-tooltip" >
+        <p >New badge level gained!</p>
+        <p >
+        </p>
```
mathjazz, line 10: Whitespace looks odd in a few places.

pontoon/teams/views.py
```diff
@@ -168,6 +168,7 @@ def ajax_permissions(request, locale):
     locale = get_object_or_404(Locale, code=locale)
     project_locales = locale.project_locale.visible().visible_for(request.user)
 
+    community_builder_level = 0
```
mathjazz, line 171: Same as above - this variable does not actually contain the value of the community builder badge level per, and the variable name should reflect that.

pontoon/base/forms.py
```diff
@@ -158,9 +130,45 @@ def save(self, *args, **kwargs):
...
+                    "pontoon.contributors.contributor.username",
+                    kwargs={
+                        "username": self.user.username,
+                    },
+                ),
+            )
+            notify.send(
+                sender=self.user,
+                recipient=self.user,
+                verb="",  # Triggers render of description only
```
mathjazz, line 166: Hmmm, this will clash with https://github.com/mozilla/pontoon/pull/3456/files#diff-3e822354a9d9b515f80ad2940a7fb3685084a1ec090224d30fc5ede87f4b0174R458...

I know, the code I'm referencing is brittle, but we should figure something out as a short-term solution. Probably use a unique verb and then handle it when rendering notifications.

pontoon/translations/views.py
```diff
@@ -450,17 +499,41 @@ def reject_translation(request):
...
+                "pontoon.contributors.contributor.username",
+                kwargs={
+                    "username": request.user.username,
+                },
+            ),
+        )
+        notify.send(
+            sender=request.user,
+            recipient=request.user,
+            verb="",  # Triggers render of description only
```
mathjazz, line 532: Isn't this the same notification we already send in [`pontoon/base/forms.py`](https://github.com/mozilla/pontoon/pull/3442/files#diff-3f6e1499881b2916dc46a23905b95100528c9221f79f5f4c1827f486bec596ad)?

translate/public/translate.html
```diff
@@ -11,6 +11,8 @@
     <link rel="icon" href="/favicon.ico" sizes="32x32">
     <link rel="icon" href="/static/img/logo.svg" type="image/svg+xml">
     <link rel="apple-touch-icon" href="/static/img/apple-touch-icon.png">
+    <link rel="review-master-badge" href=" {{ static('badges/review_master_badge.svg') }}" type="image/svg+xml">
+    <link rel="translation-champion-badge" href="{{ static('badges/translation_champion_badge.svg') }}" type="image/svg+xml">
```
mathjazz, line 15: Why do we need that?

pontoon/settings/base.py
harmitgoswami: As I understand it, Django currently only serves static files from `base/static`. I believe adding this change allows the 3 badge svg files to also be served in our app.

The tuple is just to for [better readability](https://docs.djangoproject.com/en/5.1/ref/settings/#prefixes-optional).

Is there a better way of doing this?

pontoon/translations/views.py
harmitgoswami: Yes, should there be a generic helper function for sending out badge notifications?

pontoon/base/forms.py
```diff
@@ -158,9 +128,29 @@ def save(self, *args, **kwargs):
...
+                settings.BADGES_PROMOTION_THRESHOLDS.index(after_count) + 1
+            )
+            _send_badge_notification(
+                self.user,
+                "Community Builder Badge",
+                self.community_builder_level_reached,
+            )
+
+        return self.community_builder_level_reached
+
```
mathjazz, line 153: Will this code work now if you move it back to `assign_users_to_groups()`? 

That way promotions of users in per-project permissions would also count against badges. The spec isn't clear if this is a requirement, but it would be nice to check if the current code supports it.

pontoon/settings/base.py
```diff
@@ -936,7 +939,7 @@ def _default_from_email():
 # Needed if site not hosted on HTTPS domains (like local setup)
 if not (HEROKU_DEMO or SITE_URL.startswith("https")):
     CSP_IMG_SRC = CSP_IMG_SRC + ("http://www.gravatar.com/avatar/",)
-    CSP_WORKER_SRC = CSP_FRAME_SRC = CSP_FRAME_SRC + ("http:",)
+    CSP_WORKER_SRC = CSP_FRAME_SRC = CSP_FRAME_SRC + ("http:",) + ("blob:",)
```
mathjazz, line 942: Why is this needed?

pontoon/settings/base.py
mathjazz: We use https://docs.djangoproject.com/en/5.1/ref/contrib/staticfiles/#collectstatic to collect the static files.

You can check https://docs.djangoproject.com/en/5.1/ref/contrib/staticfiles/#findstatic to check the paths:

```bash
make shell
./manage.py findstatic img
```

pontoon/translations/views.py
```diff
@@ -22,6 +22,31 @@
...
+                "username": user,
+            },
+        ),
+    )
+    notify.send(
+        sender=user,
+        recipient=user,
+        verb="ignore",  # Triggers render of description only
+        description=desc,
+    )
```
mathjazz, line 47: Good idea to factor this code out! 

I'd suggest to use a template to format the notification, like here:
https://github.com/mozilla/pontoon/blob/0d15df6dc01cda21a97c2dbb90cceb7b8a73d2c9/pontoon/projects/management/commands/send_suggestion_notifications.py#L118-L125

Also, `translations` module is not the best place to keep this code, because it's also used for promotions. Let's at least add a TODO item to the PR description to file an issue before this PR is merged to move the code to pontoon.messaging, where I'd like to keep all emails and notifications in one place.

pontoon/translations/views.py
```diff
@@ -170,18 +195,27 @@ def create_translation(request):
...
+        "status": True,
+        "translation": translation.serialize(),
+        "stats": TranslatedResource.objects.stats(project, paths, locale),
+    }
+
+    translation_count = user.badges_translation_count
+    if translation_count in settings.BADGES_TRANSLATION_THRESHOLDS:
+        response_data["badge_update"] = {
+            "name": "Translation Champion Badge",
+            "level": settings.BADGES_TRANSLATION_THRESHOLDS.index(translation_count)
```
mathjazz, line 209: Let's store these two values in variables, because they are repeated below.

pontoon/translations/views.py
```diff
@@ -170,18 +195,27 @@ def create_translation(request):
...
-            "status": True,
-            "translation": translation.serialize(),
-            "stats": TranslatedResource.objects.stats(project, paths, locale),
+    # Send Translation Champion Badge notification information
+    response_data = {
+        "status": True,
+        "translation": translation.serialize(),
+        "stats": TranslatedResource.objects.stats(project, paths, locale),
+    }
+
```
mathjazz, line 204: You could factor out the code below and call it like this:

```python
_add_badge_data(response_data, request.user, badge_name)
```

And the code would look something along the lines of:
```python
def _add_badge_data(response_data, user, badge_name):
    count = user.badges_review_count
    level = settings.BADGES_TRANSLATION_THRESHOLDS.index(count) + 1

    if count in settings.BADGES_TRANSLATION_THRESHOLDS:
        response_data["badge_update"] = {
            "name": badge_name,
            "level": level,
        }
        _send_badge_notification(
            user,
            badge_name,
            level,
        )
```

pontoon/contributors/templates/contributors/widgets/notifications_menu.html
```diff
@@ -76,7 +76,9 @@
           </span>
 
           <span class="verb">
+            {% if notification.verb != 'ignore' %}
```
mathjazz, line 79: We should instead add another `or` in line 48, because we want that condition to be true.

```
or notification.verb == "has reviewed suggestions" or notification.verb == "ignore"
```

translate/src/modules/user/components/UserNotification.tsx
```diff
@@ -39,7 +39,7 @@ const Comment = ({
     <span className='actor'>{actor.anchor}</span>
 
     <span className='verb'>
-      <a href={target.url}>{verb}</a>
+      {verb !== 'ignore' && <a href={target.url}>{verb}</a>}
```
mathjazz, line 42: See my comment in the permissions.html file.

pontoon/messaging/templates/messaging/badge_notification.jinja
```diff
@@ -0,0 +1,5 @@
+You have gained a new badge level!
+<br>
+{{ badge }}: Level {{ level }}
+<br>
+You can view this badge on your <a href={{ profile_href }}> profile page </a>.
```
mathjazz, line 5: Mind the missing empty line. I suggest you tweak your editor settings to add them automatically.

pontoon/settings/base.py
```diff
@@ -789,10 +790,11 @@ def _default_from_email():
     "django.contrib.staticfiles.finders.FileSystemFinder",
     "django.contrib.staticfiles.finders.AppDirectoriesFinder",
 )
-STATICFILES_DIRS = [
+
+STATICFILES_DIRS = (
     os.path.join(TRANSLATE_DIR, "dist"),
     os.path.join(TRANSLATE_DIR, "public"),
-]
+)
```
mathjazz, line 797: Nit: I'd leave that as it was, since it's not introducing any meaningful change.

pontoon/translations/views.py
```diff
@@ -23,21 +24,18 @@
...
+        "messaging/badge_notification.jinja",
+        {
+            "badge": badge,
+            "level": level,
+            "profile_href": reverse(
+                "pontoon.contributors.contributor.username",
+                kwargs={
+                    "username": user,
+                },
+            ),
```
mathjazz, line 37: Now that we're using a template, we can construct the URL there (see `suggestion_notification.jinja`), and only pass `user` in context.

mathjazz (commented):
Almost there!

harmitgoswami:
@mathjazz I've also changed the badge display order for the changes requested in documentation. Just curious, are you still planning to take a look at the CSS for the popup?

mathjazz:
> @mathjazz I've also changed the badge display order for the changes requested in documentation. Just curious, are you still planning to take a look at the CSS for the popup?

Thanks! The code looks good now. I'll work on the CSS tomorrow.

pontoon/translations/views.py
```diff
@@ -22,6 +23,33 @@
...
+def _send_badge_notification(user, badge, level):
+    desc = render_to_string(
+        "messaging/badge_notification.jinja",
+        {"badge": badge, "level": level, "user": user},
+    )
+    notify.send(
+        sender=user,
+        recipient=user,
+        verb="ignore",  # Triggers render of description only
+        description=desc,
```
mathjazz, line 35: Please add a [category](https://github.com/mozilla/pontoon/pull/3456/commits/b1e80b02a7b2c57b7a631629e5c828f1c6643f8f) to the notification:

```suggestion
        description=desc,
        category="badge",
```

mathjazz:
I'm getting these errors on `make build` and `make run`:

```
(!) Plugin typescript: @rollup/plugin-typescript TS2307: Cannot find module 'react-canvas-confetti/dist/presets/fireworks' or its corresponding type declarations.
src/modules/notification/components/BadgeTooltip.tsx: (10:23)

10 import Fireworks from 'react-canvas-confetti/dist/presets/fireworks';
                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

(!) Unresolved dependencies
https://rollupjs.org/troubleshooting/#warning-treating-module-as-external-dependency
react-canvas-confetti/dist/presets/fireworks (imported by "src/modules/notification/components/BadgeTooltip.tsx")
(!) Missing global variable name
https://rollupjs.org/configuration-options/#output-globals
Use "output.globals" to specify browser global variable names corresponding to external modules:
react-canvas-confetti/dist/presets/fireworks (guessing "Fireworks")
created dist/translate.js in 7.4s
```

Translate view is broken with:

```
Uncaught ReferenceError: Fireworks is not defined
<anonymous> http://localhost:8000/static/translate.js:99050
```

mathjazz:
Here's the patch to change the CSS and copy of the badge popup and notification. Please apply it before addressing the review comments. The same changes are required also in the React frontend. Please mind the [`disableForReducedMotion`](https://github.com/catdad/canvas-confetti?tab=readme-ov-file#reduced-motion).

I've also reorganized the JS code a little.

```diff
diff --git a/pontoon/messaging/templates/messaging/badge_notification.jinja b/pontoon/messaging/templates/messaging/badge_notification.jinja
index 600747d8c..db00f628a 100644
--- a/pontoon/messaging/templates/messaging/badge_notification.jinja
+++ b/pontoon/messaging/templates/messaging/badge_notification.jinja
@@ -1,5 +1,5 @@
-You have gained a new badge level!
-<br>
-{{ badge }}: Level {{ level }}
-<br>
-You can view this badge on your <a href="{{ url('pontoon.contributors.contributor.username', user.username) }}">profile page</a>.
+You've been awarded a new badge!
+<br><br>
+<span class="badge">🏆 {{ badge }}: Level {{ level }} 🏆</span>
+<br><br>
+Congratulations! View the badge in your <a href="{{ url('pontoon.contributors.contributor.username', user.username) }}">profile</a>.
diff --git a/pontoon/teams/static/css/team.css b/pontoon/teams/static/css/team.css
index e702e49a4..228afacea 100644
--- a/pontoon/teams/static/css/team.css
+++ b/pontoon/teams/static/css/team.css
@@ -72,39 +72,77 @@
   display: none;
 }

-.badge-tooltip {
+#permissions-form .badge-tooltip {
   background: var(--black-3);
   border: 1px solid var(--dark-grey-1);
   border-radius: 10px;
   box-shadow: 0 0 20px -2px var(--tooltip-background);
   box-sizing: border-box;
   text-align: center;
-  margin: 0;
+  margin: auto;
   padding: 35px;
-  position: absolute;
+  position: fixed;
   display: none;
-  bottom: 350px;
-  left: 575px;
-  width: 375px;
+  top: 0;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  width: 360px;
+  height: 560px;
   z-index: 100;

+  .title {
+    color: var(--tooltip-color-2);
+    font-weight: 100;
+    font-size: 20px;
+    text-transform: uppercase;
+  }
+
   img.badge {
     border: 2px solid var(--main-border-1);
     border-radius: 50%;
-    margin-bottom: 4px;
+    margin: 60px auto 20px;
+    width: 200px;
+    height: 200px;
   }

-  button {
-    position: absolute;
-    background: none;
-    border: none;
+  .badge-name {
+    font-weight: bold;
+    font-size: 24px;
+    margin-bottom: 5px;
+  }
+
+  .badge-level {
+    color: var(--tooltip-color-2);
+    font-weight: 100;
+    font-size: 20px;
+    margin-bottom: 45px;
+  }
+
+  .notice {
+    color: var(--tooltip-color-2);
+    font-size: 16px;
+    font-weight: 100;
+  }
+
+  .notice a {
+    color: var(--status-error);
+  }
+
+  .continue {
     background: var(--button-background-1);
-    color: var(--light-grey-7);
-    padding: 2px 4px;
-    font-weight: 300;
+    border: none;
     border-radius: 3px;
-    top: 5px;
-    right: 5px;
+    color: var(--light-grey-7);
+    font-size: 15px;
+    font-weight: 400;
+    margin-top: 10px;
+    padding: 10px 20px;
+    text-transform: uppercase;
+  }
+
+  .continue:hover {
+    color: var(--white-1);
   }
 }

diff --git a/pontoon/teams/static/js/permissions.js b/pontoon/teams/static/js/permissions.js
index 71844e73c..81741edf4 100644
--- a/pontoon/teams/static/js/permissions.js
+++ b/pontoon/teams/static/js/permissions.js
@@ -1,5 +1,6 @@
 $(function () {
   const container = $('#main .container');
+  const style = getComputedStyle(document.body);

   function inputHidden(name, value, cssClass) {
     return $(
@@ -13,6 +14,48 @@ $(function () {
     );
   }

+  function awardBadge(badgeLevel) {
+    // Update and show badge tooltip
+    container
+      .find('.badge-tooltip')
+      .show()
+      .find('.badge-name')
+      .html('Community Builder')
+      .end()
+      .find('.level')
+      .html(badgeLevel);
+
+    // Throw confetti!
+    const end = Date.now() + 5 * 1000;
+
+    const config = {
+      disableForReducedMotion: true,
+      colors: [
+        style.getPropertyValue('--status-error'),
+        style.getPropertyValue('--white-1'),
+      ],
+      particleCount: 2,
+      spread: 55,
+    };
+
+    (function frame() {
+      confetti({
+        ...config,
+        angle: 60,
+        origin: { x: 0 },
+      });
+      confetti({
+        ...config,
+        angle: 120,
+        origin: { x: 1 },
+      });
+
+      if (Date.now() < end) {
+        requestAnimationFrame(frame);
+      }
+    })();
+  }
+
   container.on('click', '#permissions-form .save', function (e) {
     e.preventDefault();
     const $form = $('#permissions-form');
@@ -46,76 +89,22 @@ $(function () {
       });
     });

-    const $tooltip = $('#badge-tooltip-container');
-
     $.ajax({
       url: $('#permissions-form').prop('action'),
       type: $('#permissions-form').prop('method'),
       data: $('#permissions-form').serialize(),
       success: function (response) {
         Pontoon.endLoader('Permissions saved.');
-        const badgeLevel = $(response).find('#community-builder-level').val();

-        // Check for new badge notification
+        const badgeLevel = $(response).find('#community-builder-level').val();
         if (badgeLevel > 0) {
-          $tooltip.show();
-
-          // Handle discard
-          $(document).on('click', function (e) {
-            if (!$tooltip.is(e.target) && $tooltip.has(e.target).length === 0) {
-              $tooltip.hide();
-            }
-          });
-
-          // Force a re-render of the text with the proper badge level
-          $tooltip
-            .find('.badge-tooltip-info')
-            .text(`Community Builder Badge level gained: Level ${badgeLevel}`);
-
-          const duration = 2000;
-          const animationEnd = Date.now() + duration;
-          const defaults = {
-            startVelocity: 30,
-            spread: 360,
-            ticks: 60,
-            zIndex: 0,
-          };
-
-          function randomInRange(min, max) {
-            return Math.random() * (max - min) + min;
-          }
-
-          const interval = setInterval(function () {
-            const timeLeft = animationEnd - Date.now();
-
-            if (timeLeft <= 0) {
-              return clearInterval(interval);
-            }
-
-            const particleCount = 50 * (timeLeft / duration);
-            // since particles fall down, start a bit higher than random
-            confetti({
-              ...defaults,
-              particleCount,
-              origin: { x: randomInRange(0.1, 0.3), y: Math.random() - 0.2 },
-            });
-            confetti({
-              ...defaults,
-              particleCount,
-              origin: { x: randomInRange(0.7, 0.9), y: Math.random() - 0.2 },
-            });
-          }, 250);
+          awardBadge(badgeLevel);
         }
       },
       error: function () {
         Pontoon.endLoader('Oops, something went wrong.', 'error');
       },
     });
-
-    $tooltip.find('button').one('click', function (e) {
-      e.preventDefault();
-      $tooltip.hide();
-    });
   });

   // Switch available users
@@ -221,4 +210,26 @@ $(function () {
       $permsForm.find('.select.available ul').append($(this).remove());
     });
   });
+
+  // Hide badge tooltip on click on the Continue button or outside of the tooltip
+  $(document).on('click', function (e) {
+    const tooltip = container.find('.badge-tooltip');
+
+    // Tooltip not visible
+    if (!tooltip.is(':visible')) {
+      return;
+    }
+
+    // Click inside the tooltip
+    if (tooltip.is(e.target) || tooltip.has(e.target).length) {
+      // Click on the Continue button
+      if (e.target.className === 'continue') {
+        e.preventDefault();
+      } else {
+        return;
+      }
+    }
+
+    tooltip.hide();
+  });
 });
diff --git a/pontoon/teams/templates/teams/includes/permissions.html b/pontoon/teams/templates/teams/includes/permissions.html
index d15a5aac8..3f4de3d6c 100644
--- a/pontoon/teams/templates/teams/includes/permissions.html
+++ b/pontoon/teams/templates/teams/includes/permissions.html
@@ -4,14 +4,15 @@
     <input type="hidden" value="{{ csrf_token }}" name="csrfmiddlewaretoken">

     <input type="hidden" id="community-builder-level" value="{{ community_builder_level_reached }}">
-    <div id="badge-tooltip-container" class="badge-tooltip">
-        <p>New badge level gained!</p>
-        <p class="badge-tooltip-info"></p>
+
+    <div class="badge-tooltip">
+        <p class="title">Achievement unlocked</p>
         <img class="badge" src="{{ static('img/community_builder_badge.svg') }}" title="Community Builder Badge" />
-        <p>You can view your new badge on your <a href="{{ url('pontoon.contributors.contributor.username', request.user) }}">profile page</a>.</p>
-        <button>Dismiss</button>
+        <p class="badge-name"></p>
+        <p class="badge-level">Level <span class="level"></span></p>
+        <p class="notice">View your new badge in your <a href="{{ url('pontoon.contributors.contributor.username', request.user) }}">profile</a>.</p>
+        <button class="continue">Continue</button>
     </div>
-

     <div class="permissions-groups general clearfix">
         <h3 class="controls">General <span class="small stress">(default team permissions for all projects)</span></h3>
```

pontoon/teams/templates/teams/includes/permissions.html
```diff
@@ -4,13 +4,14 @@
...
-        <button>Dismiss</button>
-    </div>
+        <div class="badge-tooltip">
+            <p class="title">Achievement unlocked</p>
+            <img class="badge" src="{{ static('img/community_builder_badge.svg') }}" title="Community Builder Badge" />
+            <p class="badge-name"></p>
+            <p class="badge-level">Level <span class="level"></span></p>
+            <p class="notice">View your new badge in your <a href="{{ url('pontoon.contributors.contributor.username', request.user) }}">profile</a>.</p>
+            <button class="continue">Continue</button>
+        </div>
```
mathjazz, line 14: Nit: looks like I indented this block too far.

translate/public/locale/en-US/translate.ftl
```diff
@@ -215,6 +215,16 @@ editor-NewContributorTooltip--team-info = Check the <a>team information</a> befo
 editor-NewContributorTooltip--team-managers = Reach out to <a>team managers</a> if you have questions or want to learn more about contributing.
 
 
+## Editor Badge Tooltip
+## Popup notification when users gain a new badge level
+
+editor-BadgeTooltip--intro = Achievement unlocked
+editor-BadgeTooltip--name = {{ badge }}
```
mathjazz, line 222: We can't really localize this.

translate/public/locale/en-US/translate.ftl
```diff
@@ -215,6 +215,16 @@ editor-NewContributorTooltip--team-info = Check the <a>team information</a> befo
...
+editor-BadgeTooltip--level = Level {{ level }}
```
mathjazz, line 223: Fluent uses single curly braces. :)

mathjazz:
In addition to the issues mentioned above (proposed fixes are deployed to stage), I found another one while testing on stage:

If I submit a translation to a string that already has a translation, the old translation gets rejected. That means I get 1 action logged against the Translation Champion badge (translation submitted) and 1 against the Review Master badge (translation rejected, i.e. reviewed). If I do that 5 times in a row, I can get both badges awarded at the same time. 

Our UI only shows one badge popup and sends one notification (both for the Translation Champion badge). But on the Profile page I see both badges.

We need to account for that scenario in the UI or figure out a way to count reviews differently (e.g. as explicit actions only).

pontoon/base/models/user.py
```diff
@@ -234,19 +235,62 @@ def badges_translation_count(self):
...
+    )
+
+    # Exclude auto-rejections caused by creating a new translation or approving an existing one
+    rejected_reviews = rejected_reviews.exclude(
+        Exists(
+            self.actions.filter(
+                performed_by=OuterRef("performed_by"),
+                action_type__in=["translation:created", "translation:approved"],
+                created_at__gt=OuterRef("created_at"),
+                created_at__lte=OuterRef("created_at") + timedelta(milliseconds=100),
```
mathjazz, line 255: Auto-rejections actually happen before?

```suggestion
                created_at__lte=OuterRef("created_at") - timedelta(milliseconds=100),
```

pontoon/base/models/user.py
```diff
@@ -234,19 +235,62 @@ def badges_translation_count(self):
...
+            )
+        )
+    )
+
+    return approved_reviews.count() + rejected_reviews.count()
```
mathjazz, line 260: You could reduce the number of DB queries like this:

```python
    closely_preceded_created = ActionLog.objects.filter(
        performed_by=OuterRef("performed_by"),
        ...
    )

    return self.actions.filter(
        Q(action_type="translation:approved") |
        Q(
            action_type="translation:rejected",
            ~Exists(closely_preceded_direct_action)
        ),
        created_at__gte=settings.BADGES_START_DATE,
    ).count()
```

pontoon/base/models/user.py
harmitgoswami: Yes but in that case wouldn't we want to look if a `created` or `approved` happened within 100 milliseconds after the auto-rejection? Or am I understanding incorrectly?

Changing to subtraction no longer catches the `created` or `approved` action for me anymore

mathjazz: Oh, good point. Sorry!

pontoon/base/models/user.py
harmitgoswami: would switching the order of the `~Exists` like this cause poor performance? If I keep as the second argument then pylance throws an error that "Positional argument cannot appear after keyword arguments"

```
return self.actions.filter(
        Q(action_type="translation:approved") |
        Q(  
            ~Exists(closely_preceded_action),
            action_type="translation:rejected",
        ),
        created_at__gte=settings.BADGES_START_DATE,
    ).count()
```

mathjazz: No, that shouldn't have performance implications.

mathjazz:
I noticed that confetti on the translate page appear behind the editor, editor menu and concordance search field.

pontoon/base/models/user.py
```diff
@@ -234,19 +235,55 @@ def badges_translation_count(self):
...
+        created_at__lte=OuterRef("created_at") + timedelta(milliseconds=100),
+    )
+
     return self.actions.filter(
-        Q(action_type="translation:approved") | Q(action_type="translation:rejected"),
+        Q(action_type="translation:approved")
+        | Q(
+            ~Exists(closely_preceded_action),
+            action_type="translation:rejected",
+        ),
```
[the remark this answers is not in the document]
mathjazz: This code should be reverted.

pontoon/settings/base.py
```diff
@@ -1127,13 +1130,20 @@ def account_username(user):
...
 
-# Used for Review Master and Translation Champion
+# Used for Translation Champion
 BADGES_TRANSLATION_THRESHOLDS = list(
     map(
         int,
         os.environ.get("BADGES_TRANSLATION_THRESHOLDS", "5, 50, 250, 1000").split(","),
     )
 )
+# Used for Review Master
```
mathjazz, line 1140:
```suggestion
# Used for Review Master badge
```

pontoon/settings/base.py
```diff
@@ -1127,13 +1130,20 @@ def account_username(user):
...
+BADGES_REVIEW_THRESHOLDS = list(
+    map(
+        int,
+        os.environ.get("BADGES_REVIEW_THRESHOLDS", "7, 75, 375, 1500").split(","),
```
mathjazz, line 1144: Please add a comment explaining why are the numbers higher than in the Translation Threshold.

docs/admin/deployment.rst
```diff
@@ -76,13 +76,17 @@ you create:
    Optional. A comma-separated list of numeric thresholds for different levels of the 
    Community Builder badge.
 
+``BADGES_TRANSLATION_THRESHOLDS``
```
mathjazz, line 79:
```suggestion
``BADGES_REVIEW_THRESHOLDS``
```

pontoon/translations/views.py
```diff
@@ -19,9 +19,27 @@
...
 from pontoon.checks.utils import are_blocking_checks
+from pontoon.messaging.notifications import send_badge_notification
 from pontoon.translations import forms
 
 
+def _add_badge_data(response_data, user, badge_name, stat_count):
+    if badge_name == "Review Master Badge":
+        badge_level = settings.BADGES_REVIEW_THRESHOLDS.index(stat_count) + 1
+    else:
+        badge_level = settings.BADGES_TRANSLATION_THRESHOLDS.index(stat_count) + 1
```
mathjazz, line 30: Seems like the threshold should be passed as a parameter to `_add_badge_data()`.

pontoon/base/models/user.py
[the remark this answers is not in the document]
mathjazz: I meant this whole change introduced in https://github.com/mozilla/pontoon/pull/3442/commits/2b696be27e371766a0aeb92f02f945c198989cac should be reverted, because it turned out not to scale well. That's why we are making the current changes in favour of potentially 2 badges awarded at the same time.

pontoon/settings/base.py
```diff
@@ -1127,14 +1130,23 @@ def account_username(user):
...
 BADGES_TRANSLATION_THRESHOLDS = list(
     map(
         int,
         os.environ.get("BADGES_TRANSLATION_THRESHOLDS", "5, 50, 250, 1000").split(","),
     )
 )
-# Used for Community Builder
+# Used for Review Master badge
+# Thresholds are higher than Translation Champion to minimize
+# the chances that both badges are awarded at the same exact time.
```
mathjazz, line 1142:
```suggestion
# Thresholds are higher than Translation Champion because implicit
# rejections on translations submissions and approvals also count against the Review Master badge
```

pontoon/batch/actions.py
```diff
@@ -75,6 +75,7 @@ def approve_translations(form, user, translations, locale):
             action_type=ActionLog.ActionType.TRANSLATION_APPROVED,
             performed_by=user,
             translation=t,
+            is_explicit_action=False,
```
mathjazz, line 78: Why is this action not explicit? Same question for lines 134 and 227.

pontoon/settings/base.py
```diff
@@ -1127,14 +1130,23 @@ def account_username(user):
...
 BADGES_TRANSLATION_THRESHOLDS = list(
     map(
         int,
         os.environ.get("BADGES_TRANSLATION_THRESHOLDS", "5, 50, 250, 1000").split(","),
     )
 )
-# Used for Community Builder
+# Used for Review Master badge
+# Thresholds are higher than Translation Champion because implicit rejections
+# on translations submissions and approvals also count against the Review Master badge.
```
mathjazz, line 1142: This is no longer the case? So we should use the same values as in BADGES_TRANSLATION_THRESHOLDS.
```suggestion
```

pontoon/sync/changeset.py
```diff
@@ -318,6 +318,7 @@ def update_entity_translations_from_vcs(
                         action_type=ActionLog.ActionType.TRANSLATION_APPROVED,
                         performed_by=user or self.sync_user,
                         translation=db_translation,
+                        is_explicit_action=False,
```
mathjazz, line 321: Why False?

pontoon/sync/changeset.py
```diff
@@ -494,6 +497,7 @@ def bulk_create_translations(self):
                     created_at=translation.date,
                     performed_by=translation.user or self.sync_user,
                     translation=translation,
+                    is_explicit_action=False,
```
mathjazz, line 500: Why False?

pontoon/base/models/user.py
```diff
@@ -237,16 +238,41 @@ def badges_review_count(self):
     return self.actions.filter(
         Q(action_type="translation:approved") | Q(action_type="translation:rejected"),
         created_at__gte=settings.BADGES_START_DATE,
+        is_implicit_action=True,
```
mathjazz, line 241:
```suggestion
        is_implicit_action=False,
```

mathjazz (commented):
Nice work! Please fix the two outstanding issues and then we can deploy to stage for another round of testing.

Relatedly, I think https://github.com/mozilla/pontoon/issues/3481 and https://github.com/mozilla/pontoon/issues/3482 need to be rephrased: their actions are already included in the badge counts, we just need to award the badge when they trigger it.

mathjazz (approved):
The code looks good and I also didn't spot any issues on stage. \o/

Thanks for keeping up with this roller coaster, Harmit! And nice job with the solution you've proposed for tracking explicit review actions only!

I believe we should now focus on https://github.com/mozilla/pontoon/issues/3481 and https://github.com/mozilla/pontoon/issues/3482, but I wouldn't block a release on these two issues.

Issue #2031, opened 2016-01-26
Allow users to upload personal translation memory


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1243120](https://bugzilla.mozilla.org/show_bug.cgi?id=1243120)
Bug Reporter: @mathjazz
CC:   jm-gailis@fai-vianet.fr

We should allow users to upload their personal translation memories in TMX format.

PR #3452, opened 2024-11-20
Add ability to upload personal translation memory

mathjazz:
Fix https://github.com/mozilla/pontoon/issues/2031.

This is the final step towards Translation Memory Management ([spec](https://github.com/mozilla/pontoon/blob/main/specs/0121-translation-memory-management.md)), which adds the ability for team managers and translators to upload .TMX files to their locale's TM. The PR is deployed to [stage](https://mozilla-pontoon-staging.herokuapp.com/sl/translation-memory/) for testing.

Also included is the [minor change](https://github.com/mozilla/pontoon/pull/3452/commits/91cbd86cb81b1d8df47cead08ba0fe5d62129fe8) in how error messages are presented (as errors, rather than regular messages).

TODO:
- [x] "A maximum file upload size will be imposed, with the exact size to be determined based on the performance of the upload function."

### DIFF
--- a/pontoon/teams/views.py
+++ b/pontoon/teams/views.py
@@ -399,6 +399,15 @@ def ajax_translation_memory_upload(request, locale):
             status=400,
         )
 
+    if file.size > 20 * 1024 * 1024:
+        return JsonResponse(
+            {
+                "status": False,
+                "message": "File size limit exceeded. The maximum allowed size is 20 MB.",
+            },
+            status=400,
+        )
+
     if not file.name.endswith(".tmx"):
         return JsonResponse(
             {
--- a/pontoon/teams/views.py
+++ b/pontoon/teams/views.py
@@ -431,7 +431,7 @@ def ajax_translation_memory_upload(request, locale):
 
     # Extract TM entries
     file_entries = []
-    srclang_pattern = re.compile(r"^en(?:[-_].+)?$", re.IGNORECASE)
+    srclang_pattern = re.compile(r"^en(?:[-_](us))?$", re.IGNORECASE)
     ns = {"xml": "http://www.w3.org/XML/1998/namespace"}
 
     header = root.find("header")
--- a/pontoon/teams/views.py
+++ b/pontoon/teams/views.py
@@ -422,7 +422,7 @@ def ajax_translation_memory_upload(request, locale):
 
     # Extract TM entries
     file_entries = []
-    srclang_pattern = re.compile(r"^en([-_].+)?$", re.IGNORECASE)
+    srclang_pattern = re.compile(r"^en(?:[-_].+)?$", re.IGNORECASE)
     ns = {"xml": "http://www.w3.org/XML/1998/namespace"}
 
     header = root.find("header")
--- a/pontoon/teams/views.py
+++ b/pontoon/teams/views.py
@@ -1,4 +1,6 @@
 import json
+import logging
+import xml.etree.ElementTree as ET
 
 import bleach
 
@@ -35,6 +37,9 @@ from pontoon.base.utils import get_locale_or_redirect, require_AJAX
 from pontoon.contributors.views import ContributorsMixin
 from pontoon.insights.utils import get_locale_insights
 from pontoon.teams.forms import LocaleRequestForm
+
+
+log = logging.getLogger(__name__)
 
 
 def teams(request):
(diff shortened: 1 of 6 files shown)


pontoon/teams/views.py
```diff
@@ -379,6 +384,90 @@ def ajax_translation_memory_delete(request, locale):
...
+        try:
+            source = tu.find("./tuv[@xml:lang='en-US']/seg", namespaces=ns).text.strip()
+            target = tu.find(
+                f"./tuv[@xml:lang='{code}']/seg", namespaces=ns
+            ).text.strip()
+
+            if source and target:
+                file_entries.append({"source": source, "target": target})
+        except Exception as e:
+            log.info(f"Error processing <tu>: {e}")
```
mathjazz, line 435: @eemeli Do you plan to add support for TMX parsing / serializing to moz.l10n? Until this is clear, I just baked the parser in the view, because it's fairly simple.

pontoon/teams/views.py
```diff
@@ -379,6 +384,90 @@ def ajax_translation_memory_delete(request, locale):
...
+            source=entry["source"],
+            target=entry["target"],
+            locale=locale,
+        )
+        for entry in file_entries
+    ]
+
+    created_entries = TranslationMemoryEntry.objects.bulk_create(
+        tm_entries, batch_size=1000
+    )
```
mathjazz, line 455: Should we ignore duplicates? There aren't many downsides of not doing so, safe for making the TranslationMemoryEntry larger.

pontoon/teams/views.py
mathjazz: Also, should we be less strict about the TMX format?

eemeli: No current plans for TMX support in moz.l10n. The format is sufficiently different in structure that it doesn't really fit into the message & resource data models, and so would need a separate, different representation.

pontoon/teams/views.py
```diff
@@ -379,6 +384,90 @@ def ajax_translation_memory_delete(request, locale):
...
+        return JsonResponse(
+            {"status": False, "message": f"Invalid XML file: {e}"}, status=400
+        )
+
+    # Extract TM entries
+    file_entries = []
+    ns = {"xml": "http://www.w3.org/XML/1998/namespace"}
+    for tu in root.findall(".//tu"):
+        try:
+            source = tu.find("./tuv[@xml:lang='en-US']/seg", namespaces=ns).text.strip()
```
mathjazz, line 427: We should read the source locale code from the `<header srclang>` attribute and use it here. We should support `en`, `en-*` and `en_*` in any capitalization.

We should also support the `lang` attribute (in addition to `xml:lang`).

flodolo: I'm confused by this comment: what does "we should" mean? This patch needs to be updated, follow-up?

> We should read the source locale code from the <header srclang> attribute and use it here.

The translation unit can also have a `srclang` attribute.

pontoon/teams/views.py
flodolo: > There aren't many downsides of not doing so, safe for making the TranslationMemoryEntry larger.

You lost me at the double negative. The only downside is making TranslationMemoryEntry larger? But why would we want that, if we can avoid it?

mathjazz: Not ignoring dupes makes the code simpler and faster and thus the max. file upload size larger. 

It's also less confusing than uploading a TMX with 1000 entries and getting a "531 entries uploaded" success message. But we can include the number of skipped entries in the message as well.

I'll take a stab at that.

pontoon/teams/views.py
mathjazz: > I'm confused by this comment: what does "we should" mean? This patch needs to be updated, follow-up?

I'll update this PR.

> The translation unit can also have a `srclang` attribute.

I would ignore that, at least for the first iteration.

mathjazz: Thanks, I'll build upon that.

pontoon/teams/views.py
mathjazz: Updated.

mathjazz:
Made a few updates:
- [Read srclang from `<header>` and `<tu>`, rather than hardcoding it](https://github.com/mozilla/pontoon/pull/3452/commits/a5bd224589277aabc484ed49c608f56b700c7732)
- [Add support for lang attribute in `<tuv>` in addition to xml:lang](https://github.com/mozilla/pontoon/pull/3452/commits/8f2244b3da8fcbe0cff7cdb4f2609d0044819cc2)
- [Ignore duplicates on import](https://github.com/mozilla/pontoon/pull/3452/commits/bc17f6f06c9db090950bc72121ea3ea40f06ff55) (also updated response JSON and message)

pontoon/teams/views.py
```diff
@@ -379,6 +385,143 @@ def ajax_translation_memory_delete(request, locale):
...
+        tree = ET.parse(file)
+        root = tree.getroot()
+    except ET.ParseError as e:
+        return JsonResponse(
+            {"status": False, "message": f"Invalid XML file: {e}"}, status=400
+        )
+
+    # Extract TM entries
+    file_entries = []
+    srclang_pattern = re.compile(r"^en([-_].+)?$", re.IGNORECASE)
```
flodolo, line 425: This feels a bit overkill.

Do we care about excluding English variants from this? That was my understanding: we want `en` and `en-US`, the latter with possible weird spelling, not `en-GB` and `en-CA`.

In case, the group should probably be non-capturing (`^en(?:[-_].+)?$`).

flodolo (approved):

pontoon/teams/views.py
flodolo: We don't do that when it [comes to VCS](https://github.com/mozilla/pontoon/blob/bc2f3642a93d797553f662720ed86e68cf0b00bf/pontoon/sync/vcs/project.py#L50-L57), why would we want a different behavior for TM import?

eemeli: Note that with the sync refactor, the VCS source discovery moves to moz.l10n and becomes [configurable](https://github.com/mozilla/moz-l10n/blob/615ee50f034203bd9fe9a853e98804fcd8995dd8/moz/l10n/paths/discover.py#L83-L86).

But that's not really relevent, tbh. The cases are different because the project paths are controlled by admins, while the TMX values are controlled by individual translators. The former is set when the project is initially configured, while the latter can happen at any time by ~anyone.

But my thoughts here are not very strong, esp. as this parser is getting inlined within the view code like this.

pontoon/teams/views.py
```diff
@@ -379,6 +385,152 @@ def ajax_translation_memory_delete(request, locale):
...
+
+        return seg.text.strip() if seg is not None and seg.text else None
+
+    tu_elements = root.findall(".//tu")
+    for tu in tu_elements:
+        try:
+            srclang = tu.attrib.get("srclang", header_srclang)
+
+            if not srclang_pattern.match(srclang):
+                log.info(f"Skipping <tu> with unsupported srclang: {srclang}")
```
eemeli, line 456:
```suggestion
                tu_str = ET.tostring(tu, encoding='unicode')
                log.info(f"Skipping <tu> with unsupported srclang: {tu_str}")
```

eemeli (approved):

PR #3455, opened 2024-11-21
Add an API endpoint to retrieve User action data

mathjazz:
Introduce an API endpoint to retrieve User action data on a given date for a given project:
`/api/v1/user-actions/YYYY-MM-DD/project/PROJECT-SLUG/`

Example:
[/api/v1/user-actions/2024-01-07/project/thunderbird/](https://mozilla-pontoon-staging.herokuapp.com/api/v1/user-actions/2024-01-07/project/thunderbird/)

The following actions are tracked:
* Translation created
* Translation deleted
* Translation approved
* Translation unapproved
* Translation rejected
* Translation unrejected

Notes:
* Authentication is required.
* The links above point to the stage server, which only has data until 2024-11-20. 
* Performance should be a bit better on the production instance.

### DIFF
--- a/pontoon/api/urls.py
+++ b/pontoon/api/urls.py
@@ -1,7 +1,8 @@
 from graphene_django.views import GraphQLView
 
-from django.urls import re_path
+from django.urls import include, path, re_path
 
+from pontoon.api import views
 from pontoon.api.schema import schema
 from pontoon.settings import DEV
 
@@ -12,5 +13,33 @@ urlpatterns = [
     re_path(
         r"^graphql/?$",
         GraphQLView.as_view(schema=schema, graphiql=DEV),
+    ),
+    # API v1
+    path(
+        "api/v1/",
+        include(
+            [
+                path(
+                    # User actions
+                    "get-user-actions/",
+                    include(
+                        [
+                            # General
+                            path(
+                                "<str:date>/",
+                                views.get_user_actions,
+                                name="pontoon.api.get_user_actions",
+                            ),
+                            # Project-specific
+                            path(
+                                "<str:date>/project/<slug:slug>/",
+                                views.get_user_actions,
+                                name="pontoon.api.get_user_actions.project",
+                            ),
+                        ]
+                    ),
+                ),
+            ]
+        ),
     ),
 ]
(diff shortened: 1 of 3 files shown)


pontoon/api/urls.py
```diff
@@ -13,4 +14,32 @@
...
         GraphQLView.as_view(schema=schema, graphiql=DEV),
     ),
+    # API v1
+    path(
+        "api/v1/",
+        include(
+            [
+                path(
+                    # User actions
+                    "get-user-actions/",
```
eemeli, line 24: The HTTP protocol already includes the GET verb.

```suggestion
                    "user-actions/",
```

pontoon/api/views.py
```diff
@@ -0,0 +1,102 @@
...
+            "resource": {
+                "pk": resource.pk,
+                "path": resource.path,
+                "format": resource.format,
+            },
+            "project": {
+                "pk": project.pk,
+                "slug": project.slug,
+                "name": project.name,
+            },
```
eemeli, line 90: This seems like a lot of duplication, but presumably the data volumes here aren't large enough to really care?

eemeli (approved):
Presuming that the suggested API URL change is acceptable.

sduenas:
Would it be possible that, instead of requesting the actions for a given day requesting them for a range? Probably with a parameter instead of having that on the url. This way it's easier to request the latest actions since the last time they were retrieved or to look for something in particular.

mathjazz:
Hi @sduenas, thank you for your feedback!

The primary goal of this changeset was to minimize the performance impact of retrieving contribution data by implementing a streamlined solution. Introducing the ability to request data for arbitrary time ranges increases the risk of potential performance degradation, which would necessitate additional safeguards such as limits and pagination—measures we are currently unable to commit to.

Would it be feasible for you to retrieve historical data using a one-off script and then set up a cron job to collect data incrementally on a daily basis, similar to the hourly process you employed in the past?

We can certainly make `date` a parameter instead of including it in the URL.

mathjazz:
Thanks everyone for the feedback. Let's merge this PR as it has been approved.

Please file issues for any further enhancements.

PR #3474, opened 2024-12-05
Add ability to send Monthly activity summary

### LINKED ISSUE #3467: Implement Monthly activity summary
Implement ability to send a summary of a user’s activity as per the [spec](https://github.com/mozilla/pontoon/blob/main/specs/0120-transactional-emails.md#monthly-activity-summary).

mathjazz:
Fix #3467.

This PR implements the ability to send monthly activity summary as per the [spec](https://github.com/mozilla/pontoon/blob/main/specs/0120-transactional-emails.md#settings-changes). It also factors out the base HTML email template, which will be shared by all HTML emails we send.

The following work required by the [spec](https://github.com/mozilla/pontoon/blob/main/specs/0120-transactional-emails.md#settings-changes) remains:
* Send onboarding transactional emails
* Send transactional emails to inactive accounts


pontoon/messaging/templates/messaging/emails/monthly_activity_summary.html
```diff
@@ -0,0 +1,106 @@
...
+    background: #c0ff00;
+    border-color: #c0ff00;
+}
+{% endblock %}
+
+{% block header %}Hello.{% endblock %}
+
+{% block subheader %}
+<p>Here's your monthly activity summary for the month of {{ month }}.</p>
+<p>A heartfelt thank you from the Mozilla Localization team and the countless users who benefit from your translations. Your dedication makes Mozilla’s products and services more accessible and enjoyable for people all over the world.</p>
```
mathjazz, line 61: I thought something like this would be nice, saw something similar in the emails I'm getting from SUMO.

mathjazz:
> Below is an example of how the email looks:

@bcolsson Flagging you for the review of the content and the design of the email.

pontoon/messaging/emails.py
```diff
@@ -1,23 +1,246 @@
...
+    # Get contributors that started contributing to the locale in the given month
+    first_contributions = actions.values(
+        "performed_by", "translation__locale"
+    ).annotate(first_contribution_date=Min("created_at"))
+
+    new_locale_contributors = {
+        (entry["translation__locale"], entry["performed_by"])
+        for entry in first_contributions
+        if entry["first_contribution_date"].month == month_date.month
+        and entry["first_contribution_date"].year == month_date.year
```
eemeli, line 100: This requires fetching and annotating the entire actionlog, and filtering it in Python. Is there no way of doing the work in the DB?

pontoon/messaging/emails.py
```diff
@@ -1,23 +1,246 @@
...
+        locale.previous_state = previous_locale_state.get(locale.pk, {})
+        locale.contributors = locale_contributors.get(locale.pk, {})
+
+    # Create a map of users to locales in which they are managers or translators,
+    # which determines if the user should receive the Team activity section of the email
+    user_locales = defaultdict(set)
+    for locale in locales:
+        # Skip locales without data
+        if not locale.data or not locale.previous_data:
+            log.error(f"Locale {locale} has no Monthly activity data.")
```
eemeli, line 206: Why is this considered an error?

pontoon/messaging/management/commands/send_monthly_activity_emails.py
```diff
@@ -0,0 +1,25 @@
...
+            help="Force sending regardless of the current date.",
+        )
+
+    def handle(self, *args, **options):
+        # Only send on the given day of the month or when --force is used
+        if options["force"] or now().day == settings.MONTHLY_ACTIVITY_SUMMARY_DAY:
+            send_monthly_activity_summary()
+        else:
+            self.stdout.write(
+                "This command can only be run on the first day of the month. Use --force to bypass."
```
eemeli, line 24: `MONTHLY_ACTIVITY_SUMMARY_DAY` is configurable.

pontoon/messaging/templates/messaging/emails/monthly_activity_summary.html
```diff
@@ -0,0 +1,109 @@
...
+.bullet.pretranslated {
+    background: #c0ff00;
+    border-color: #c0ff00;
+}
+{% endblock %}
+
+{% block header %}Hello.{% endblock %}
+
+{% block subheader %}
+<p>Here's your monthly activity summary for the month of {{ month }}.</p>
```
eemeli, line 60: Include the year? As in,
> Here's your monthly activity summary for December 2024.

eemeli (changes requested):

pontoon/messaging/emails.py
mathjazz: This is not intuitive at all, but: when `values()` and `annotate()` are [combined](https://docs.djangoproject.com/en/5.1/topics/db/aggregation/#values), unique combinations of the fields specified in the `values()` clause are grouped. Only these combinations are then returned and annotated. In our case that means instead of returning 7 million items, we're only returning 7 thousand.

That said, I can filter entries by the given month and year already in the QuerySet, which brings the number of items to about 100.

eemeli (approved):

bcolsson (approved):
This looks great, I like the additions you made.
The only change I'd like is to change `month before` to any one of the following alternatives:
`last month`
`previous month`
`month previous`

bcolsson:
I don't know if this would add technical complexity, if so we can consider it for the future, but it might also be nice to have 
"No recent manager activity"
"No recent translator activity"
"No recent contributor activity"
appear instead of nothing if there aren't any active managers/translators/contributors.

PR #3475, opened 2024-12-09
Configurable day to send weekly notification digest emails on

mathjazz:
Let's not hardcode the day on which to send weekly notification digest emails. That way we can also make sure it includes the (weekly) suggestion notifications, for which the sending day can also be configured.

### DIFF
--- a/pontoon/messaging/management/commands/send_notification_emails.py
+++ b/pontoon/messaging/management/commands/send_notification_emails.py
@@ -1,5 +1,6 @@
+from django.conf import settings
 from django.core.management.base import BaseCommand
-from django.utils.timezone import now
+from django.utils.timezone import datetime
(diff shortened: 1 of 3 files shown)


pontoon/messaging/management/commands/send_notification_emails.py
```diff
@@ -11,5 +12,5 @@ def handle(self, *args, **options):
         send_notification_digest(frequency="Daily")
 
         # Only send weekly digests on Saturdays
```
eemeli, line 14:
```suggestion
```

eemeli (approved):
Presuming inline incorrect comment removal is included.

Might be good to grep the codebase for any other weekday mentions, if you've not done so yet.

PR #3477, opened 2024-12-09
Fix new contributor detection logic

### LINKED ISSUE #3476: New contributor notifications are not being sent
We broke sending new contributor notifications in #2677 by moving the `is_new_contributor()` check _after_ the point when translation is saved, while keeping its logic (return True whether the user has not made any contributions yet). That way the function always returns False.

mathjazz:
Fix #3476.

I've added a test that checks whether the new contributor notification is sent.

I've also excluded the notification to be sent when project manager makes the contribution to the locale.

### DIFF
--- a/pontoon/base/models/user.py
+++ b/pontoon/base/models/user.py
@@ -282,11 +282,12 @@ def can_translate(self, locale, project):
 
 
 def is_new_contributor(self, locale):
-    """Return True if the user hasn't made contributions to the locale yet."""
+    """Return True if the user has made just 1 contribution to the locale."""
     return (
-        not self.translation_set.filter(locale=locale)
+        self.translation_set.filter(locale=locale)
         .exclude(entity__resource__project__system_project=True)
-        .exists()
+        .count()
+        == 1
     )
 
 
--- a/pontoon/base/models/user.py
+++ b/pontoon/base/models/user.py
@@ -281,7 +281,7 @@ def can_translate(self, locale, project):
     return self.has_perm("base.can_translate_locale", locale)
 
 
-def is_new_contributor(self, locale):
+def has_one_contribution(self, locale):
     """Return True if the user has made just 1 contribution to the locale."""
     return (
         self.translation_set.filter(locale=locale)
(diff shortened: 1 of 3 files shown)


pontoon/translations/tests/test_views.py
```diff
@@ -210,6 +212,54 @@ def test_run_checks_during_translation_update(
...
+
+    assert response.status_code == 200
+    assert response.json()["status"]
+
+    mock_notify.assert_called_once()
+
+    # Check that the new contributor notification was sent to the correct manager
+    notify_args = mock_notify.call_args[1]
+    assert notify_args["recipient"] == user_a
+    assert notify_args["category"] == "new_contributor"
```
eemeli, line 260: After this, should create a second translation and check that no `new_contributor` notification is sent for that.

pontoon/translations/views.py
```diff
@@ -126,7 +126,11 @@ def create_translation(request):
         )
 
     # When user makes their first contribution to the team, notify team managers
-    first_contribution = not project.system_project and user.is_new_contributor(locale)
+    first_contribution = (
+        not project.system_project
+        and user.is_new_contributor(locale)
+        and user != project.contact
```
eemeli, line 132: Probably imperceptible optimization, but checking the project data doesn't need a DB query.
```suggestion
        and user != project.contact
        and user.is_new_contributor(locale)
```

pontoon/base/models/user.py
```diff
@@ -282,11 +282,12 @@ def can_translate(self, locale, project):
 
 
 def is_new_contributor(self, locale):
-    """Return True if the user hasn't made contributions to the locale yet."""
+    """Return True if the user has made just 1 contribution to the locale."""
```
eemeli, line 285: Mild preference for using a different function name, as this is setting up a surprising definition of "new contributor" that works in the context where it's used, but is here declared for User in general.

eemeli (changes requested):

pontoon/base/models/user.py
mathjazz: I thought about that, but couldn't come up with a better name (the name was already misleading with the old logic). Do you have a suggestion? `has_one_contribution`?

eemeli: `has_one_contribution` is better.

eemeli (approved):

PR #3483, opened 2024-12-11
Update ruff (0.8.2) and uv (0.5.7)

flodolo:


docker/compile_requirements.sh
```diff
@@ -5,7 +5,7 @@
 
 export CUSTOM_COMPILE_COMMAND="./docker/compile_requirements.sh"
 
-uv pip compile --generate-hashes $@ requirements/default.in -o requirements/default.txt
-uv pip compile --generate-hashes $@ requirements/dev.in -o requirements/dev.txt
-uv pip compile --generate-hashes $@ requirements/lint.in -o requirements/lint.txt
-uv pip compile --generate-hashes $@ requirements/test.in -o requirements/test.txt
+uv pip compile --generate-hashes --no-strip-extras $@ requirements/default.in -o requirements/default.txt
```
flodolo, line 8: This is a change that has been sitting in #3312 for a while.

.github/workflows/backend.yml
```diff
@@ -42,10 +42,12 @@ jobs:
         with:
           python-version: '3.11'
       - name: Set up uv
-        uses: astral-sh/setup-uv@v2
+        uses: astral-sh/setup-uv@v4
         with:
           enable-cache: true
-          version: "0.4.13"
+          cache-dependency-glob: |
+            **/requirements/*.txt
```
flodolo, line 49: This is needed because the default in v4 is searching for `uv.lock`. Even the new default in the upcoming v5 won't work for our requirements.
https://github.com/astral-sh/setup-uv/issues/181

mathjazz (approved):
LGTM

PR #3479, opened 2024-12-10
Update Transactional emails spec

mathjazz:
Minor copy updates and fixes in Onboarding and Inactive account emails.

Please see the updated text in the designs of all transactional emails below. Note that the http://localhost:8000/ URLs will be replaced with https://pontoon.mozilla.org/ in production.

Onboarding 1:
onboarding1

Onboarding 2:
onboarding2

Onboarding 3:
onboarding3

Inactive contributor: 
inactive-contributor

Inactive translator: 
Inactive-translator

Inactive manager: 
Inactive-manager

PR #3485, opened 2024-12-11
Onboarding and inactive account transactional emails

mathjazz:
This PR implements all remaining parts of the [transactional email spec](https://github.com/mozilla/pontoon/blob/main/specs/0120-transactional-emails.md#activity-summary), namely the ability to send onboarding and inactive account transactional emails.

See https://github.com/mozilla/pontoon/pull/3479 for how the emails look.

The PR also moves contact email verification code to `pontoon.messaging` and sends the email as HTML, using the common base email template.

- [ ] Transactional email content in the templates is generic and stored in the DB, so we need to updated it with the Mozilla-specific content after deployment.


pontoon/messaging/templates/messaging/emails/content/inactive_contributor.html
```diff
@@ -0,0 +1,12 @@
+<p>It's been a while since we’ve last seen you, but we wanted to thank you for contributing to localization efforts at <a href="{{ full_url('pontoon.homepage') }}">{{ full_url('pontoon.homepage') }}</a>.</p>
+
+<p>Your contributions have been important for all readers in your language. We understand that life can get busy, and finding time to contribute to projects isn't always easy. We get it.</p>
+
+<p>As part of our efforts to keep our localization community active and engaged, we periodically review account activity. We’ve noticed you haven’t signed in to your Pontoon account in the past 6 months. While we completely understand that circumstances change, we wanted to inform you that we may deactivate dormant accounts in the future to ensure the security and efficiency of our platform.</p>
```
mathjazz, line 5: I hardcoded `6 months` here since the content can be changed.

Alternatively, we could use the `INACTIVE_CONTRIBUTOR_PERIOD` variable here and similar for the inactive translator / manager emails. But then we should store months (not days) in the variable.

pontoon/messaging/models.py
```diff
@@ -53,3 +53,12 @@ class Meta:
 
     def is_new(self):
         return self.sent_at > timezone.now() - timezone.timedelta(minutes=1)
+
+
+class EmailContent(models.Model):
+    email = models.CharField(max_length=255, unique=True)
```
mathjazz, line 59: Not sure about the naming here. Maybe that's better?

```suggestion
class EmailTemplate(models.Model):
    type = models.CharField(max_length=255, unique=True)
```

pontoon/messaging/management/commands/send_inactive_account_emails.py
```diff
@@ -0,0 +1,111 @@
...
+            .distinct()
+        )
+
+        inactive_translators = (
+            users.filter(
+                pk__in=translators.keys(),
+            )
+            .exclude(
+                pk__in=active_user_ids,
+            )
```
eemeli, line 80: This means that someone who's a translator in one locale and a manager of another locale could get two emails, right?

pontoon/messaging/emails.py
```diff
@@ -310,4 +313,259 @@ def send_notification_digest(frequency="Daily"):
...
+    now = timezone.now()
+    UserProfile.objects.filter(user__in=pks).update(last_inactive_reminder_sent=now)
+
+    log.info(f"Inactive translator emails sent to { len(users) } users.")
+
+
+def send_inactive_manager_emails(users, manager_map):
+    """
+    Sends an email to an inactive manager.
+    """
```
eemeli, line 507: Is a manager becoming inactive significant enough that someone else should also be notified about it?

eemeli (approved):

pontoon/messaging/management/commands/send_inactive_account_emails.py
mathjazz: Yes.

pontoon/messaging/emails.py
mathjazz: Adding @bcolsson for visibility and a potential followup.

bcolsson: Since these don't repeat after the first instance of it being sent, I'm not sure how useful it would be. Since we have our locale health dashboard and will be looking at those more proactively going forward, I don't think we need to add anything here.

PR #3499, opened 2024-12-18
Add functionality for counting batch actions

### LINKED ISSUE #3481: Award badges for batch actions
PR #3442 introduced a new field to ActionLog, called `is_implicit_action`, that distinguishes between user-completed actions and background actions. 

Currently, batch updates to translations count are counted in badge stats. If a batch action triggers a new badge level, then we should send the pop-up tooltip and notification as we do for individual actions.

harmitgoswami:
Fix #3481 

This PR adds badge stat counting logic for batch actions. 

The initial commit is able to send the badge update logic to `translate/src/context/BadgeTooltip.tsx`, but the tooltip component doesn't seem to be rendering.

### DIFF
--- a/pontoon/batch/actions.py
+++ b/pontoon/batch/actions.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from django.utils import timezone
 
 from pontoon.actionlog.models import ActionLog
@@ -6,6 +7,14 @@ from pontoon.base.models import (
     TranslationMemoryEntry,
 )
 from pontoon.batch import utils
+from pontoon.messaging.notifications import send_badge_notification
+
+
+def _get_badge_level(thresholds, action_count):
+    for level in range(len(thresholds) - 1):
+        if thresholds[level] <= action_count < thresholds[level + 1]:
+            return level + 1
+    return 0
 
 
 def batch_action_template(form, user, translations, locale):
@@ -69,6 +78,10 @@ def approve_translations(form, user, translations, locale):
         locale,
     )
 
+    before_level = _get_badge_level(
+        settings.BADGES_REVIEW_THRESHOLDS, user.badges_review_count
+    )
+
     # Log approving actions
     actions_to_log = [
         ActionLog(
@@ -79,6 +92,16 @@ def approve_translations(form, user, translations, locale):
         for t in translations
     ]
     ActionLog.objects.bulk_create(actions_to_log)
+
+    # Send Review Master Badge notification information
+    after_level = _get_badge_level(
+        settings.BADGES_REVIEW_THRESHOLDS, user.badges_review_count
+    )
+    badge_update = {}
+    if after_level > before_level:
+        badge_update["level"] = after_level
+        badge_update["name"] = "Review Master Badge"
+        send_badge_notification(user, badge_update["name"], badge_update["level"])
 
     # Approve translations.
     translations.update(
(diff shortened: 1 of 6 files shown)


pontoon/batch/actions.py
```diff
@@ -80,6 +93,16 @@ def approve_translations(form, user, translations, locale):
...
 
+    # Send Review Master Badge notification information
+    after_level = _get_badge_level(
+        settings.BADGES_REVIEW_THRESHOLDS, user.badges_review_count
+    )
+    badge_update = {}
+    if after_level > before_level:
+        badge_update["level"] = after_level
+        badge_update["name"] = "Review Master Badge"
+        send_badge_notification(user, badge_update["name"], badge_update["level"])
```
mathjazz, line 104: Nit:
```suggestion
        send_badge_notification(user, "Review Master Badge", after_level)
```

pontoon/batch/actions.py
```diff
@@ -69,6 +78,10 @@ def approve_translations(form, user, translations, locale):
         locale,
     )
 
+    before_level = _get_badge_level(
+        settings.BADGES_REVIEW_THRESHOLDS, user.badges_review_count
+    )
```
mathjazz, line 83: Could we instead call this as something along the lines of `user.badges_review_level`? Where `badges_review_level` would be a user property with the logic similar to `_get_badge_level()`?

pontoon/batch/actions.py
```diff
@@ -135,6 +149,14 @@ def reject_translations(form, user, translations, locale):
     ]
     ActionLog.objects.bulk_create(actions_to_log)
 
+    # Send Review Master Badge notification information
+    after_level = user.badges_review_level
+    badge_update = {}
+    if after_level > before_level:
+        badge_update["level"] = after_level
+        badge_update["name"] = "Review Master Badge"
```
mathjazz, line 157: Not sure why this works in other actions, but the `BadgeTooltip` component expects badge names without the word Badge at the end in `switch` - `case`.

mathjazz: In `useSendTranslation` we seem to hardcode the `badgeName` rather than using the one returned from the backend?

harmitgoswami: Great catch, I was so confused 🥲

pontoon/batch/actions.py
```diff
@@ -80,6 +83,14 @@ def approve_translations(form, user, translations, locale):
...
+        send_badge_notification(user, "Review Master Badge", after_level)
```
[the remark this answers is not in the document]
harmitgoswami: Should I revert this change then?

pontoon/batch/actions.py
mathjazz: Let's use the same message style everywhere, i.e. drop "Badge" at the end of the badge name.

mathjazz: That way we can also use the `badgeName` in `useSendTranslation`.

pontoon/batch/actions.py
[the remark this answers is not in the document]
mathjazz: Yes, please! :)

translate/src/modules/editor/hooks/useSendTranslation.ts
```diff
@@ -87,9 +87,9 @@ export function useSendTranslation(): (ignoreWarnings?: boolean) => void {
       );
 
       const badgeLevel = content.badge_update?.level;
-      if (badgeLevel) {
+      if (content.badge_update?.name && badgeLevel) {
```
mathjazz, line 90: Since this pattern is used twice, could you please instead declare it above?

`const badgeName = content.badge_update?.name;`,

mathjazz (approved):
Great work! 👍

PR #3506, opened 2025-01-03
Fix sync - updating source string from VCS not working

haodave:


flodolo:
Can you add more details on the issue/behavior you're trying to fix, so that it can be reproduced?

pontoon/sync/core/entities.py
```diff
@@ -162,23 +163,22 @@ def update_resources(
...
-            and not entities_same(ent, prev_entities[key])
-        ),
-        [
-            "string",
-            "string_plural",
-            "comment",
-            "source",
-            "group_comment",
-            "resource_comment",
-            "context",
```
eemeli, line 179: Why is `context` dropped?

pontoon/sync/core/entities.py
```diff
@@ -333,6 +333,18 @@ def entities_same(a: Entity, b: Entity) -> bool:
     )
 
 
+def entity_update(
+    current: Entity, update_from: Entity, fields: list[str]
+) -> Optional[Entity]:
```
eemeli, line 338: This is replacing the only use of `entities_same()` with a new function; we should not leave orphaned code lying around.

eemeli (changes requested):
This needs to include a test case that fails with the current code, but passes with the changes. It's not at all clear what entity change is not getting picked up by the current code.

There are also two parts to this change: How differences are identified, and whether the previous or next entity is used in `Entity.objects.bulk_update()`. Are both changes required?

pontoon/sync/core/entities.py
haodave: It's a mistake. Resolved

haodave:
> As mentioned, can you clarify the behavior you're trying to fix?
> 
> Do you have a specific example (ideally a minimal case) of what is not working and with what content?

Yes. I updated the PR description earlier. Here are more details of the behavior:
1. In the source file `en.json`, update `"key": "value"` to `"key": "value2"` with a PR.
2. Sync the Pontoon project.
3. The source string should have been changed from `value` to `value2` but it's not.

mathjazz:
I was able to reproduce the problem in https://github.com/mozilla-l10n/pontoon-test/commit/bcad64712cc5f2dd6bddb8c9059bc43453baa698.

eemeli:
@haodave Could you allow edits from maintainers for this branch, or pick up the additional changes from [here](https://github.com/eemeli/pontoon/tree/fix/update-source-string-from-repo)? I added a test case for this, and applied a couple of other minor fixes. Once that's done, this should be good to merge.

eemeli (approved):

Issue #3507, opened 2025-01-03
Possible issues in use of Font Awesome


flodolo:
## fa vs fas

The prefix `fa` was deprecated in favor of `fas` in v5 (which we're currently using)
https://docs.fontawesome.com/v5/web/reference-icons

> The fa prefix has been deprecated in version 5. The new default is the fas solid style and the fab style for brands.

## Use of far

~We seem to be using `far` in a few places (e.g. [admin](https://github.com/mozilla/pontoon/blob/main/pontoon/administration/templates/admin_inline_formset.html#L15)), but as far I can tell we only get "solid" (`fas`) and "brand" (`fab`) with a free plan.~

~Also, is [this line](https://github.com/mozilla/pontoon/blob/main/pontoon/teams/static/js/request.js#L215) correct?~

```
.toggleClass('fa far fa-circle fa-dot-circle enabled');
```

## Other

Is `fa-w` still valid? I think it might have been dropped from v5

https://github.com/mozilla/pontoon/blob/main/translate/src/modules/search/components/SearchPanel.tsx#L41-L45

flodolo:
Regarding `far`, this is clearly working in `messaging/sent/`, so I'm either missing something, or the documentation is incorrect.

flodolo:
> Also, is [this line](https://github.com/mozilla/pontoon/blob/main/pontoon/teams/static/js/request.js?rgh-link-date=2025-01-03T15%3A39%3A22Z#L215) correct?
> 
> ```
> .toggleClass('fa far fa-circle fa-dot-circle enabled');
> ```

Yes, because it toggles any of the classes listed 🙄 
https://api.jquery.com/toggleClass/

flodolo:
If we want to replace `fa` with  `fas`, I have a work-in-progress branch (it's a lot more work than I expected)
https://github.com/mozilla/pontoon/compare/main...flodolo:update_fa_references?expand=1

flodolo:
> Is `fa-w` still valid? I think it might have been dropped from v5

One more discovery: the current version of `font-awesome-all.css` in Pontoon doesn't have a CSS rule for `fa-w`.

But v6 from https://github.com/mozilla/pontoon/pull/3504 has a rule

```
.fa-w {
  --fa: "\57";
}
```

That's just the letter W (0x0057).

flodolo:
All issues were covered by
https://github.com/mozilla/pontoon/pull/3504
https://github.com/mozilla/pontoon/pull/3508
[...]

PR #3508, opened 2025-01-04
Fix advanced search options (missing Fluent strings, CSS)

flodolo:
For `fa-w`, see #3507 

The console is full or React errors when opening the panel. The edit removed a bunch of trailing whitespaces, which I think it's useful to remove anyway.

### DIFF
--- a/translate/public/locale/en-US/translate.ftl
+++ b/translate/public/locale/en-US/translate.ftl
@@ -154,7 +154,7 @@ editor-FtlSwitch--active =
 editor-FailedChecks--close = ×
     .aria-label = Close failed checks popup
 editor-FailedChecks--title = THE FOLLOWING CHECKS HAVE FAILED
-editor-FailedChecks--save-anyway = SAVE ANYWAY 
+editor-FailedChecks--save-anyway = SAVE ANYWAY
 editor-FailedChecks--suggest-anyway = SUGGEST ANYWAY
 editor-FailedChecks--approve-anyway = APPROVE ANYWAY
 
@@ -468,7 +468,7 @@ interactivetour-InteractiveTour--comments-content =
     In the Comments tab you can discuss how to translate content with
     your fellow team members. It’s also the place where you can request
     more context about or report an issue in the source string.
-    
+
 interactivetour-InteractiveTour--machinery-title = Machinery
 interactivetour-InteractiveTour--machinery-content =
     The Machinery tab shows automated translation suggestions from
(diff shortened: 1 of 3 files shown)


translate/src/modules/search/components/SearchPanel.tsx
```diff
@@ -38,7 +38,7 @@ const SearchOption = ({
         onToggle();
       }}
     >
-      <i className='fa fa-w'></i>
```
flodolo, line 41: Was this meant to be `fa-fw` (fixed-width)? In any case, the menu seems to work fine without it.

translate/public/locale/en-US/translate.ftl
```diff
@@ -682,11 +682,15 @@ search-FiltersPanel--apply-filters =
 
 ## Search Options Panel
 
-search-SearchPanel--option-name-identifiers = Search in string identifiers
+search-SearchPanel--option-name-search_match_case = Match case
+search-SearchPanel--option-name-search_match_whole_word = Match whole word
+search-SearchPanel--option-name-search_identifiers = Search in string identifiers
+search-SearchPanel--option-name-search_rejected_translations = Include rejected translations
+search-SearchPanel--option-name-search_exclude_source_strings = Exclude source strings
```
mathjazz, line 689: It would be nice to be consistent with the rest of the file and the README and use the kebab-case:
https://github.com/mozilla/pontoon/blob/0788ed9de8b71b3ab9cb1fb76bb40fe0fe9fa5e0/translate/README.md#localization

translate/public/locale/en-US/translate.ftl
```diff
@@ -682,11 +682,15 @@ search-FiltersPanel--apply-filters =
...
 
 search-SearchPanel--heading = SEARCH OPTIONS
 
-search-SearchPanel--apply-options = APPLY SEARCH OPTIONS
+search-SearchPanel--apply-searchOptions = APPLY SEARCH OPTIONS
```
mathjazz, line 693: Same as above - `...search-options`.

mathjazz:
Thanks for the update!

Didn't notice this before - I don't get a hover effect on the Apply search options button - neither the text color nor the mouse cursor change.

flodolo:
I think I've fixed it, but I'm not sure I understand the intent of the original code (why was the margin-top changing? Was the extra class ever used?)

mathjazz (approved):

PR #3509, opened 2025-01-06
Use 'fas' instead of 'fa' for Font Awesome

flodolo:
Prefix was deprecated in v5
https://docs.fontawesome.com/v5/web/reference-icons

> Prefix Changes in V5
> The fa prefix has been deprecated in version 5. The new default is the fas solid style and the fab style for brands.

### DIFF
--- a/pontoon/administration/templates/admin.html
+++ b/pontoon/administration/templates/admin.html
@@ -15,7 +15,7 @@
   <div class="container">
     <menu class="controls">
       <div class="search-wrapper big clearfix">
-        <div class="icon fa fa-search"></div>
+        <div class="icon fas fa-search"></div>
         <input class="table-filter" type="search" autocomplete="off" autofocus placeholder="Filter projects">
       </div>
       <a class="add button small" href="{{ url('pontoon.admin.project.new') }}">Add new project</a>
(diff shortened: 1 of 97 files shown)


mathjazz:
Seems like the position of the Unreviewed (bulb) icon has changed.

Prod:
https://pontoon.mozilla.org/af/amo/

Stage (running this patch):
https://mozilla-pontoon-staging.herokuapp.com/af/amo/

The misalignment is also visible in the filter menu in the translate view.

The change might have also been caused by the library update.

flodolo:
The previous version had some horizontal padding

Screenshot 2025-01-06 alle 14 13 52

flodolo:
It doesn't feel amazing, because I can get the exact same position and have to use half pixel, but close enough? Not sure what an alternative approach is.

mathjazz:
To make it "pixel perfect" (avoid icon jumping when I switch between https://mozilla-pontoon-staging.herokuapp.com/sl/ and https://pontoon.mozilla.org/sl/) I had to do this:
 
Screenshot 2025-01-06 at 17 49 41

flodolo:
Updated. The columns in the tables are slightly off, but match the current status

Senza nome copia

mathjazz (approved):

PR #3510, opened 2025-01-07
Make advanced search options more discoverable

### LINKED ISSUE #3502: Improve visibility of advanced search options
While the lens icon copies the behavior of the funnel icon (there is no visual indication that it will open a dropdown), part of the feedback during the survey was that the feature is not discoverable.

I see two options to improve the situation.

## Move the advanced search options to the main dropdown

This was suggested during implementation, and I still feel like a separate menu is the right approach. The idea is that users should touch these options very rarely, because the defaults cover most of the cases.

## Add a visual hint

Looking into Font Awesome, it looks like it [supports icon stacking](https://docs.fontawesome.com/web/style/stack), so we could stack a down arrow with the lens? I did some quick experiments, but it doesn't look amazing.

flodolo:
Fixes #3502

### DIFF
--- a/translate/src/modules/search/components/SearchBox.css
+++ b/translate/src/modules/search/components/SearchBox.css
@@ -12,7 +12,7 @@
   font-size: 14px;
   font-weight: 300;
   line-height: 23px;
-  padding: 5px 33px 5px 37px;
+  padding: 5px 36px 5px 37px;
   width: 100%;
 }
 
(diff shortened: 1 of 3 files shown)


flodolo:
The linter should run on these jobs 🤔

mathjazz (approved):
It would be nice to highlight the icons on hover. Other than that, it looks good.

flodolo:
> It would be nice to highlight the icons on hover. Other than that, it looks good.

Added highlighting (white). I think that piece of code was effectively unused?
https://github.com/mozilla/pontoon/pull/3510/commits/df344b1345747b897a91897a4c40af833ff41e21

PR #3522, opened 2025-01-10
Add functionality for counting file uploads

### LINKED ISSUE #3482: Award badges for file uploads
PR #3442 introduced a new field to ActionLog, called `is_implicit_action`, that distinguishes between user-completed actions and background actions. 

Similar to batch updates (#3481), actions from file uploads count towards badge stats. If a file upload triggers a new badge level, then we should send the pop-up tooltip and notification as we do for individual actions.

harmitgoswami:
Fix #3482 

This PR adds badge stat counting logic for file uploads. 

The communication between frontend and backend is done using the Django messaging framework.

**To reproduce results**: Head to any resource, click on your profile in the top right, and then click "Download Translations". Save this file in a text editor, and edit 5 translations. Then, head back to the resource, click on your profile again, and click "Upload Translations". Upload the file you edited. Once the upload is complete, you should be awarded the Translation Champion Badge for this action.

### DIFF
--- a/pontoon/sync/core/translations_from_repo.py
+++ b/pontoon/sync/core/translations_from_repo.py
@@ -77,12 +77,12 @@ def sync_translations_from_repo(
 def write_db_updates(
     project: Project, updates: Updates, user: User | None, now: datetime
 ) -> tuple[str, int]:
-    badge_update, updated_translations, new_translations = update_db_translations(
-        project, updates, user, now
+    badge_name, badge_level, updated_translations, new_translations = (
+        update_db_translations(project, updates, user, now)
     )
     add_failed_checks(new_translations)
     add_translation_memory_entries(project, new_translations + updated_translations)
-    return badge_update
+    return badge_name, badge_level
 
 
 def delete_removed_bilingual_resources(
@@ -435,31 +435,24 @@ def update_db_translations(
             f"[{project.slug}] Created {str_n_translations(created)} from repo changes"
         )
 
-    badge_update = ("", 0)
+    badge_name = ""
+    badge_level = 0
     if actions:
         translation_before_level = log_user.badges_translation_level
         review_before_level = log_user.badges_review_level
 
         ActionLog.objects.bulk_create(actions)
+        if log_user.username != "pontoon-sync":
+            if log_user.badges_translation_level > translation_before_level:
+                badge_name = "Translation Champion"
+                badge_level = log_user.badges_translation_level
+                send_badge_notification(log_user, badge_name, badge_level)
+            if log_user.badges_review_level > review_before_level:
+                badge_name = "Review Master"
+                badge_level = log_user.badges_review_level
+                send_badge_notification(log_user, badge_name, badge_level)
 
-        if (
-            log_user.badges_translation_level > translation_before_level
-            and log_user.username != "pontoon-sync"
-        ):
-            send_badge_notification(
-                log_user, "Translation Champion", log_user.badges_translation_level
-            )
-            badge_update = ("Translation Champion", log_user.badges_translation_level)
-        if (
-            log_user.badges_review_level > review_before_level
-            and log_user.username != "pontoon-sync"
-        ):
-            send_badge_notification(
-                log_user, "Review Master", log_user.badges_review_level
-            )
-            badge_update = ("Review Master", log_user.badges_review_level)
-
-    return badge_update, created, list(suggestions.values())
+    return badge_name, badge_level, created, list(suggestions.values())
 
 
 def str_n_translations(n: int | Sized) -> str:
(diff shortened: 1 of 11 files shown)


pontoon/sync/core/translations_from_repo.py
```diff
@@ -433,10 +435,31 @@ def update_db_translations(
...
     if actions:
+        translation_before_level = log_user.badges_translation_level
+        review_before_level = log_user.badges_review_level
+
         ActionLog.objects.bulk_create(actions)
 
-    return created, list(suggestions.values())
+        if (
+            log_user.badges_translation_level > translation_before_level
+            and log_user.username != "pontoon-sync"
```
mathjazz, line 447: You could add `if log_user.username == "pontoon-sync"` above this if, so you don't need to write the condition twice.

pontoon/sync/core/translations_from_repo.py
```diff
@@ -433,10 +435,31 @@ def update_db_translations(
...
+        ):
+            send_badge_notification(
+                log_user, "Translation Champion", log_user.badges_translation_level
+            )
+            badge_update = ("Translation Champion", log_user.badges_translation_level)
```
mathjazz, line 452: The following would honor the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle:
```suggestion
            badge = "Translation Champion"
            level = log_user.badges_translation_level
            send_badge_notification(log_user, badge, level)
            badge_update = (badge, level)
```

translate/src/context/BadgeTooltip.tsx
```diff
@@ -21,6 +21,26 @@ export function BadgeTooltipProvider({
...
+            } else {
+              return { content: notification.content, type: notification.type };
+            }
+          },
+        );
+        setMessage(parsed[1]);
+      }
+    }
+  }, []);
+
```
mathjazz, line 43: Can we instead just modify the Notification context?

(Also, it's confusing that we refer to Django Messaging framework messages as Notifications (the name we already use for something completely different). Should possibly file a task to rename that.)

https://github.com/mozilla/pontoon/blob/a0b431cff0b6844dd1ba9018010db4010e11497e/translate/src/context/Notification.tsx#L25-L38

mathjazz (commented):
I haven't tested the code yet, but it make sense. Left some suggestions.

translate/src/context/BadgeTooltip.tsx
mathjazz: Perhaps the right way forward here is to just factor out a function that checks if we have any notifications to display.

mathjazz:
Sorry for the confusion - in https://github.com/mozilla/pontoon/pull/3522/files#r1913781900 and the following comment I meant to communicate that the approach I used in the linked gist is not good enough to be used and we should just factor out a function that checks if we have any notifications to display.

translate/src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,47 @@
...
+  // Note that we only show it once, and only when the UI has already
+  // been rendered, to make sure users do see it.
+  useEffect(() => {
+    const rootElt = document.getElementById('root');
+    if (rootElt?.dataset.notifications) {
+      const notifications = JSON.parse(rootElt.dataset.notifications);
+      if (notifications.length > 0) {
+        const generalNotification = notifications.find(
+          (notification: { type: string }) =>
+            notification.type !== 'badge info',
```
mathjazz, line 21: Where is that `badge info` value coming from? I thought we'd use the custom tag here to differentiate between badge messages and the rest?

harmitgoswami: Oh as in create a custom [message level](https://docs.djangoproject.com/en/5.1/ref/contrib/messages/#creating-custom-message-levels), not a message tag? If I’m understanding correctly

translate/src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,50 @@
...
+    null,
+  );
+
+  // If there's a notification in the DOM set by Django, show it.
+  // Note that we only show it once, and only when the UI has already
+  // been rendered, to make sure users do see it.
+  useEffect(() => {
+    const rootElt = document.getElementById('root');
+    if (rootElt?.dataset.notifications) {
+      console.log(rootElt?.dataset.notifications);
```
mathjazz, line 17: Nit: Please remove.

translate/src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,50 @@
...
+      const notifications = JSON.parse(rootElt.dataset.notifications);
+      if (notifications.length > 0) {
+        // Extra tags from the Django messages framework are combined
+        // with the level tag into a single string as notification.type
```
mathjazz, line 21:
```suggestion
        // with the level into a single string as notification.type
```

mathjazz (approved):

PR #3527, opened 2025-01-14
Drop SVN support

### LINKED ISSUE #3303: Unused repo sync features should be dropped (per-locale repositories, svn support)
On `pontoon.mozilla.org`, we've recently migrated the Firefox, Thunderbird, and SeaMonkey localization repositories from hg to git. After this change, we have no projects using hg or svn as their repositories, and we have no projects using per-locale repositories (i.e. with `{locale_code}` in the repo URL).

We should probably drop support for these features, as they're a source of unnecessary complexity and we currently can't validate their behaviour in production. Mercurial (hg) repo support should probably be retained, as it's pretty plausibly used by some third party.

Given that I'm currently working on a sync refactor, this change is liable to be made relatively soon. If you use Pontoon with per-locale repositories or svn, now would be a Really Good Time to speak up.

eemeli:
Closes #3303 

The prod dependency on `heroku-buildpack-apt` is also dropped as unnecessary, including what appears to be an unused dependency on `libsqlite3-dev`.


mathjazz (commented):
Please add the missing migration. Apart from that looks good!

eemeli:
Done. Didn't expect for a migration to be needed, as the `base_repository` table only shows the field as `character varying(255)`, rather than having its values actually limited to the Charfield choices. Which is good, as that means that any remaining `'svn'` values in it shouldn't cause any catastrophic error.

mathjazz (approved):

eemeli:
Note about deployment with this change: The buildpack also needs to be dropped from the Heroku app settings, via the web UI.

Issue #3514, opened 2025-01-08
Stats are off at locale level after sync (Italian)


flodolo:
Locale was at 100%. Mozilla accounts was updated with 25 new strings, which were pretranslated, and then translated.

The locale currently shows 43 missing strings (no clue where that number comes from). The following sync didn't fix the number.

CC @eemeli

flodolo:
After a new batch of strings, stats look good.

New batch of strings (80 in notifications, but actually missing only 34, since they were migrated to a new file).

Right after sync, the number matches (at locale-level in the header, and at project-level). 
After pretranslation, missing went down to 0.
After translation, everything is at 0.

Not sure if `calculate_stats` fixed it for good?

flodolo:
Correction, I should have looked at the locale, which is way worse than before: now it says I have 542 strings missing (!!), and 9 unreviewed. There were never unreviewed strings at any point.

eemeli:
I can't see any current error with the stats for Italian, Mozilla Accounts, or Mozilla Accounts in Italian. Is the issue here replicable anywhere?

flodolo:
> I can't see any current error with the stats for Italian, Mozilla Accounts, or Mozilla Accounts in Italian. Is the issue here replicable anywhere?

Not in Italian. @mathjazz mentioned that Slovenian was also broken, but I think that was fixed by me recalculating stats.

mathjazz:
Slovenian looks good now.

flodolo:
For the record, this happened again this morning.

Mozilla accounts was updated overnight, with a bunch of other projects (63 new strings in FxA, 3 new strings in Firefox iOS).

<img width="391" alt="Image" src="https://github.com/user-attachments/assets/98a779d9-5596-4498-a734-f6f69d50df74" />

- Missing and unreviewed were completely off (there were only 59 missing, 0 unreviewed), pretranslated was correct.
- Translating strings didn't lower the number for either, but lowered the number for pretranslated.
- Once I was done translating, the project-level stats were correct, the locale-level stats didn't change.

I checked about 30 minutes later, and they seem to have fixed themselves, possibly because another project updated?

```
Jan 16 20:42:57 mozilla-pontoon app/worker.1 [firefox-for-ios] Updated stats for 98 translated resources, 98 projectlocales, and 367 locales
Jan 16 20:52:50 mozilla-pontoon app/worker.1 [firefox-for-android] Updated stats for 4057 translated resources, 134 projectlocales, and 367 locales
[...]

PR #3536, opened 2025-01-17
Refactor stats, removing aggregate fields from project, projectlocale & locale

eemeli:
Based on my unscientific local testing, the rendering speed benefits from eagerly populating the aggregated stats into Project, ProjectLocale & Locale don't exist, compared to collating the data on the fly from TranslatedResource objects. Therefore, we should not be doing this complex extra work. This significantly simplifies the logic around stats updates, and makes the stats much easier to reason about.

The AggregatedStats class no longer extends Model, and it's mostly a crutch for reducing the risks in this change; hence its move from `base/models/` to `base/`. With it, fields like `total_strings` and `strings_with_errors` will continue to work for the Django models from which the corresponding DB fields were removed.


eemeli:
It's likely (but not certain) that this will also provide a solution for #3514.

mathjazz:
The branch is deployed to stage for testing.

flodolo:
> The branch is deployed to stage for testing.

Stats for Italian are completely off?
https://mozilla-pontoon-staging.herokuapp.com/it/

mathjazz:
https://mozilla-pontoon-staging.herokuapp.com/sl/ also shows 495 more Missing strings in the heading than the sum of Missing ProjectLocale strings in the table.

Are we including private projects?

eemeli:
> Are we including private projects?

No, but we aren't filtering out disabled projects in some cases where we should. I'll apply a fix later today.

pontoon/base/models/translated_resource.py
```diff
@@ -1,149 +1,79 @@
...
-            unreviewed=Sum("unreviewed_strings"),
+    def string_stats(
+        self, user: User | None = None, *, show_disabled: bool = False
+    ) -> dict[str, int]:
+        query = (
+            self
+            if show_disabled
+            else self.filter(
+                resource__project__disabled=False,
+                resource__project__system_project=False,
```
mathjazz, line 26: I find the name `show_disabled`, given it also impacts `system_projects`.

 ```suggestion
             if show_localizable
             else self.filter(
                 resource__project__disabled=False,
                 resource__project__system_project=False,
```

File: pontoon/base/views.py [hunk 2]
```diff
@@ -103,30 +103,51 @@ def locale_projects(request, locale):
...
+                "title",
+                "total",
+                "pretranslated",
+                "errors",
+                "warnings",
+                "unreviewed",
+                "approved",
+            )
+        )
+        all_res_stats = tr.string_stats(request.user, show_disabled=True)
```
mathjazz, line 142: Why? Same question for other places where we use `show_disabled=True`.

mathjazz: I think I know why. It's related to my question above? It's to show the right data for system projects, right?

pontoon/base/models/translated_resource.py
eemeli: Maybe we should _always_ filter out disabled, and have a toggle for system projects?

mathjazz (approved):
Looks good!

We tested on stage and not only do stats look fine, but we also didn't notice performance regressions.

We should check New Relic after deploying to prod and monitor any performance impact. We should also verify the Insights are still collected as expected.

PR #3537, opened 2025-01-22
Renamed status banner to user banner in css

### LINKED ISSUE #3434: Rename status banners to user banners
To be consistent with the [localizer documentation](https://mozilla-l10n.github.io/localizer-documentation/tools/pontoon/translation_workspace.html#user-banners), we should rename all instances of `status-banner` to just `user-banner` in the codebase.

It seems like `status-banner` only appears in `UserAvatar.css`.

harmitgoswami:
Fix #3434


mathjazz (commented):
Changes look good.

I also see instances of `user-status` and `userStatus` that I believe should be `user-banner` and `userBanner`.

translate/src/hooks/useUserStatus.test.js
mathjazz, line 1: We rename this file, too.

mathjazz (approved):

PR #3541, opened 2025-01-23
Send translation comment notifications to managers and translators

### LINKED ISSUE #3469: Send translation comment notifications to more users
Locale managers should always receive notifications for translation comments. That's particularly important when a comment is added to a translation that doesn't have an author.

Should we also send notifications to translators? @flodolo?

For reference, translation comment notifications are sent to significantly less users than team comment notifications:
https://github.com/mozilla/pontoon/blob/a6195e4fa6ddfb9ed32416abc3c41305980c554d/pontoon/base/views.py#L476-L502

harmitgoswami:
Fix #3469 

This PR factors out the logic for sending comment notifications for both team and translation comments to locale managers and translators.

**To reproduce results**: Upon starting the server, go to any locale and add yourself as a manager/translator within that locale (using the `/permissions/` page). Then, create a new superuser in a new terminal. Open a new container tab/browser session, and log in to the local session of Pontoon using the new superuser you created. Go to any string in the same locale that the first account is now a manager/translator for, and add a team comment and translation comment. The first account should receive a notification in both cases.

TODO:
- Update [notification documentation](https://mozilla-l10n.github.io/localizer-documentation/tools/pontoon/notifications.html#comments)

### DIFF
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -478,6 +478,8 @@ def get_team_comments(request):
 
 def _send_add_comment_notifications(user, comment, entity, locale, translation):
     # On translation comment, notify:
+    #   - project-locale translators or locale translators
+    #   - locale managers
     #   - authors of other translation comments in the thread
     #   - translation author
     #   - translation reviewers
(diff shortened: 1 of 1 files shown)


mathjazz (approved):
Nice work!

PR #3548, opened 2025-01-29
Use timeago format in Latest Start column of the Sync Log

mathjazz:
As discussed on Slack...

### DIFF
--- a/pontoon/sync/templates/sync/log_list.html
+++ b/pontoon/sync/templates/sync/log_list.html
@@ -30,8 +30,8 @@
         <thead>
           <tr>
             <th class="project asc">Project<i class="fas"></i></th>
-            <th class="last-status">Last Status<i class="fas"></i></th>
-            <th class="last-start">Last Start<i class="fas"></i></th>
+            <th class="latest-status">Latest Status<i class="fas"></i></th>
+            <th class="latest-start">Latest Start<i class="fas"></i></th>
             <th class="duration" data-sort="">Duration<i class="fas"></i></th>
           </tr>
         </thead>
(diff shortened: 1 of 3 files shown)


eemeli (commented):
A few concerns:
1. It's not clear what a "month" is? Is it exactly four weeks, or something else? It would probably be better to go from days to weeks to years.
2. Why is the change only applied to the listing of all projects' most recent sync events, and not the errors view or the project-specific views?
3. As discussed on Slack, it would probably be best to explicitly use and show (with a trailing `Z`) UTC timestamps for all non-relative times.

mathjazz:
> It's not clear what a "month" is? Is it exactly four weeks, or something else? It would probably be better to go from days to weeks to years.

Yeah, that's how Django works for this feature and it's used across the app already.

> Why is the change only applied to the listing of all projects' most recent sync events, and not the errors view or the project-specific views?

The answer is simple and straighforward: I forgot.

> As discussed on Slack, it would probably be best to explicitly use and show (with a trailing `Z`) UTC timestamps for all non-relative times.

I added UTC. I wasn't aware of Z(ulu).

pontoon/sync/templates/sync/log_list.html
```diff
@@ -47,13 +47,13 @@ <h4>
...
+              <td class="latest-status {{ format_sync_status_class(project.sync_status) }}">
                 {{ format_sync_status_label(project.sync_status) }}
               </td>
-              <td class="last-start">
-                {{ project.sync_start_time|date("Y-m-d H:i:s") or "―" }}
+              <td class="latest-start">
+                <time {% if project.sync_start_time %}datetime="{{ project.sync_start_time.isoformat() }}" title='{{ project.sync_start_time|date("Y-m-d H:i:s") }} UTC'{% endif %}>{{ project.sync_start_time|naturaltime or "―" }}</time>
               </td>
-              <td class="duration" title='{% if project.sync_end_time %}Last end: {{ project.sync_end_time|date("Y-m-d H:i:s") }}{% endif %}' data-sort="{{ format_sync_duration(project.sync_start_time, project.sync_end_time, ms=True) }}">
+              <td class="duration" title='{% if project.sync_end_time %}Latest End: {{ project.sync_end_time|date("Y-m-d H:i:s") }} UTC{% endif %}' data-sort="{{ format_sync_duration(project.sync_start_time, project.sync_end_time, ms=True) }}">
```
eemeli, line 56: The start time formatting (here and also in the other files) uses `.isoformat()`, but that's not included for end times. Should it be?

mathjazz: `.isoformat()` is only used in the `datetime` attribute, which is used for sorting.

eemeli (approved):

PR #3549, opened 2025-01-29
Change default theme to system

### LINKED ISSUE #3339: Pontoon should use light theme for visitors requesting light theme
On a new profile, set Firefox to request light color scheme and visit Pontoon. Pontoon keeps using the dark theme, while it should follow the user's preferences.

harmitgoswami:
Fix #3339 

This PR changes the default theme for all users to match their system preferences, regardless of if they're an authenticated user.

To reproduce results: After running the server, open `localhost:8000` in a new tab. Then, go into your computer's system settings and switch your preferred theme to the opposite of what you currently have selected. You should see Pontoon's theme also change at the same time you switch your preferred theme.

### DIFF
--- a/pontoon/base/models/user_profile.py
+++ b/pontoon/base/models/user_profile.py
@@ -35,7 +35,7 @@ class UserProfile(models.Model):
     theme = models.CharField(
         choices=Themes.choices,
         max_length=20,
-        default=Themes.DARK,
+        default=Themes.SYSTEM,
     )
 
     # External accounts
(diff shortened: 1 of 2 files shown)


mathjazz (commented):
Nice work!

I noticed two issues when using the light system theme:
* The Sign In button uses the light text color, which is not readable on the light background.
* When I click the Sign In button, the login page appears dark. See https://github.com/mozilla/pontoon/tree/main/pontoon/base/templates/django.

mathjazz (commented):
Nice work!

The only remaining issue is that if a system theme is set to light, users will experience FOUC when logged out. 

We must still be defaulting to the dark theme somewhere, and need to make sure we honour `prefers-color-scheme` before the page renders.

PR #3556, opened 2025-02-07
Replace deprecated runtime.txt with .python-version

mathjazz:
https://devcenter.heroku.com/articles/python-runtimes#selecting-a-runtime


.gitignore
```diff
@@ -1,4 +1,5 @@
 .*
+!.python-version
```
eemeli, line 2: This is unnecessary. `.gitignore` is only considered for files which aren't already in git, and a file matching a pattern here can be added with `git add --force`.

```suggestion
```

eemeli (approved):
This is fine, even though it does impact me a bit: I'll probably need to start using `uv` to run `./manage.py`, `pytest`, etc. I currently have a `pontoon` virtualenv set up, and with that as the contents of the `.python-version` file, pyenv automatically mounts it when I `cd` into Pontoon.

Also not quite sure if this will require some extra fiddling in VS Code to have it pick the right Python.

See inline for one small fix that ought to be applied before merging.

PR #3558, opened 2025-02-08
Add button to trigger sync in sync logs page

### LINKED ISSUE #3557: Allow requesting manual sync from the sync log page
It would be nice to be able to request a manual sync directly from the sync log.

flodolo:
Fixes #3557

### DIFF
--- a/pontoon/base/static/css/style.css
+++ b/pontoon/base/static/css/style.css
@@ -203,8 +203,7 @@ th {
 }
 
 .table-sort th i {
-  margin-left: 5px;
-  position: absolute;
+  pointer-events: none;
 }
 
 .table-sort th.asc i:after,
--- a/pontoon/base/static/css/style.css
+++ b/pontoon/base/static/css/style.css
@@ -197,7 +197,7 @@ th {
   color: var(--white-1);
 }
 
-.table-sort th {
+.table-sort th.no-sort {
   pointer-events: none;
 }
 
--- a/pontoon/base/static/css/style.css
+++ b/pontoon/base/static/css/style.css
@@ -197,13 +197,13 @@ th {
   color: var(--white-1);
 }
 
-.table-sort th.no-sort:hover {
-  color: var(--light-grey-7);
-  cursor: default;
+.table-sort th {
+  pointer-events: none;
 }
 
 .table-sort th i {
-  pointer-events: none;
+  margin-left: 5px;
+  position: absolute;
 }
 
 .table-sort th.asc i:after,
--- a/pontoon/base/static/css/style.css
+++ b/pontoon/base/static/css/style.css
@@ -197,6 +197,11 @@ th {
   color: var(--white-1);
 }
 
+.table-sort th.no-sort:hover {
+  color: var(--light-grey-7);
+  cursor: default;
+}
+
 .table-sort th i {
   margin-left: 5px;
   position: absolute;
--- a/pontoon/sync/static/js/sync_log.js
+++ b/pontoon/sync/static/js/sync_log.js
@@ -1,22 +1,10 @@
 $(function () {
-  // Remove any existing sync-status-* class
-  function removeSyncStatusClasses($element) {
-    $element.removeClass(function (index, className) {
-      return (className.match(/(^|\s)sync-status-\S+/g) || []).join(' ');
-    });
-  }
-
   // Request to sync a project
   $('.sync').click(function (e) {
     e.preventDefault();
 
     const projectSlug = $(this).attr('project-slug');
-    const $row = $(this).closest('tr');
-    const $statusCell = $row.find('td').eq(1);
-
-    removeSyncStatusClasses($statusCell);
-    $statusCell.addClass('sync-status-other').text('Requested');
-
+    Pontoon.endLoader('Sync in progress...');
     $.ajax({
       url: `/admin/projects/${projectSlug}/sync/`,
       success: function () {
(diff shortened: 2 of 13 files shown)


pontoon/base/static/css/style.css
```diff
@@ -197,6 +197,11 @@ th {
   color: var(--white-1);
 }
 
+.table-sort th.no-sort:hover {
+  color: var(--light-grey-7);
+  cursor: default;
```
mathjazz, line 202:
```suggestion
  pointer-events: none;
```

With that, you can revert the changes made to [table.js](https://github.com/mozilla/pontoon/pull/3558/files#diff-79b6a0398959ea93fb1e1c5c9028bdfbca94d1e7d2d884a9ab3e4024e89add95).

Seems like `color` is not needed?

This change should be moved to table.css though, because we might needed it elsewhere.

pontoon/sync/templates/sync/log_list.html
```diff
@@ -33,6 +33,9 @@
             <th class="latest-status">Latest Status<i class="fas"></i></th>
             <th class="latest-start relative-time">Latest Start<i class="fas"></i></th>
             <th class="duration" data-sort="">Duration<i class="fas"></i></th>
+            {% if settings.MANUAL_SYNC and is_manager %}
```
mathjazz, line 36: A general observation, since this is from the previous PR: `request` is available in the templates, so passing `is_manager` is not needed.

pontoon/sync/templates/sync/log_list.html
```diff
@@ -56,6 +59,11 @@ <h4>
               <td class="duration" title='{% if project.sync_end_time %}Latest End: {{ project.sync_end_time|date("Y-m-d H:i:s") }} UTC{% endif %}' data-sort="{{ format_sync_duration(project.sync_start_time, project.sync_end_time, ms=True) }}">
                 {{ format_sync_duration(project.sync_start_time, project.sync_end_time) }}
               </td>
+              {% if settings.MANUAL_SYNC and is_manager and not project.system_project %}
```
mathjazz, line 62: It's the DB projects (not system projects) for which we should not show the Sync button. 

We should also hide that Sync button for those in the Project Admin.

pontoon/sync/static/js/sync_log.js
```diff
@@ -0,0 +1,33 @@
...
+    });
+  }
+
+  // Request to sync a project
+  $('.sync').click(function (e) {
+    e.preventDefault();
+
+    const projectSlug = $(this).attr('project-slug');
+    const $row = $(this).closest('tr');
+    const $statusCell = $row.find('td').eq(1);
```
mathjazz, line 15:
```suggestion
    const $statusCell = $row.find('td.latest-status');
```

pontoon/sync/static/js/sync_log.js
```diff
@@ -0,0 +1,33 @@
...
+
+    removeSyncStatusClasses($statusCell);
+    $statusCell.addClass('sync-status-other').text('Requested');
```
mathjazz, line 18: Why not use the "In progress" status here?

Alternatively, you could avoid all the code that manages the statuses and their classes by using something like this:

```js
Pontoon.endLoader("Sync in progress...", "error")
```

And this in case of an error:

```js
Pontoon.endLoader("Unable to sync the project.", "error")
```

I think that's a fair compromise given the change will only be visible for a second or two.

mathjazz (commented):
Nice work!

Left a few suggestions. Before we merge, I might want to make a small CSS change to account for adding the ability to disable sync in the future.

pontoon/base/static/css/style.css
flodolo: Interesting, never seen `pointer-events` before. I tried it but it doesn't seem to work completely: 
* The pointer doesn't change (so I can replace both rules).
* The JS event is still triggered, so I still need the code in `table.js`.

> This change should be moved to table.css though, because we might needed it elsewhere.

That's one confusing bit I discovered: CSS seems to be in both files? `.table-sort th:hover` is literally the rule above this one. Will leave it here for now, especially if you plan to make other CSS changes.

pontoon/sync/static/js/sync_log.js
flodolo: > Why not use the "In progress" status here?

Because I cannot know if the sync actually is in progress or failed.

Going with your proposal, I assume the `error` in your first code is not needed.

pontoon/sync/templates/sync/log_list.html
flodolo: But then you would need to repeat the condition `request.user.has_perm("base.can_manage_project")` 3 times in this template (or enclose everything in a `with`), which doesn't seem more readable? Is there a better way to do that?

mathjazz: You could [declare a variable in the template](https://jinja.palletsprojects.com/en/stable/templates/#assignments):
```jinja
{% set is_manager = request.user.has_perm("base.can_manage_project") %}
```

pontoon/base/static/css/style.css
mathjazz: The `:hover` bit needs to be deleted from the CSS selector.

flodolo: I also edited the wrong part of the CSS 🤦🏼

mathjazz (approved):
Made a couple of minor changes. When and if we make more changes to the sync log, we'll adapt again if needed.

@flodolo Could you please take a look at the changes and test on stage?

flodolo:
Changes look good (both stage and local instance) 👍🏼

PR #3559, opened 2025-02-10
Apply ruff to root, not just pontoon/

eemeli:
Apply ruff linting & formatting to repo root.


mathjazz (commented):
We should also update the ruff command in other places, namely `Makefile`.

mathjazz (approved):

PR #3561, opened 2025-02-11
Reduce database queries on dashboards

### LINKED ISSUE #3545: N+1 query problem on dashboards
The new stats code (https://github.com/mozilla/pontoon/pull/3536) is making a separate DB query for each of the 369 locales in the /teams page to get theirs stats, which makes it noticeably slower than the old code.

We should change the way get_chart() works and only do 1 query for all locales.

eemeli:
Fixes #3545 

This should speed things up a bit. Two related changes are included to resolve the identified inefficiencies, and a couple of the `Locale` methods are dropped/simplified to match current usage. The `ProjectLocale.get_latest_activity()` function is split into two separate implementations on `Locale` and `Project`, as they are now rather different, and the `get_object_or_none()` utility is dropped as obsolete.

### DIFF
--- a/pontoon/base/models/project.py
+++ b/pontoon/base/models/project.py
@@ -13,7 +13,7 @@ from pontoon.base.models.locale import Locale
 
 
 if TYPE_CHECKING:
-    from pontoon.base.models import Resource
+    from pontoon.base.models import ProjectLocale, Resource
 
 
 class Priority(models.IntegerChoices):
@@ -124,6 +124,7 @@ class Project(models.Model, AggregatedStats):
     slug = models.SlugField(unique=True)
     locales = models.ManyToManyField(Locale, through="ProjectLocale")
 
+    project_locale: BaseManager["ProjectLocale"]
     resources: BaseManager["Resource"]
 
     class DataSource(models.TextChoices):
--- a/pontoon/base/models/project.py
+++ b/pontoon/base/models/project.py
@@ -284,9 +284,11 @@ class Project(models.Model, AggregatedStats):
                 else None
             )
 
+        pl_fetched = getattr(self, "fetched_project_locale", None)
         project_locale = (
-            getattr(self, "fetched_project_locale", None)
-            or ProjectLocale.objects.filter(project=self, locale=locale).first()
+            pl_fetched[0]
+            if pl_fetched
+            else ProjectLocale.objects.filter(project=self, locale=locale).first()
         )
         return (
             project_locale.latest_translation.latest_activity
--- a/pontoon/base/models/project.py
+++ b/pontoon/base/models/project.py
@@ -4,7 +4,7 @@ from typing import TYPE_CHECKING
 from django.conf import settings
 from django.contrib.auth.models import User
 from django.db import models
-from django.db.models import Prefetch, Sum
+from django.db.models import Sum
 from django.db.models.manager import BaseManager
 from django.utils import timezone
 
@@ -67,24 +67,6 @@ class ProjectQuerySet(models.QuerySet):
         Syncable projects are same as force-syncable, but must not have sync disabled.
         """
         return self.force_syncable().filter(sync_disabled=False)
-
-    def prefetch_project_locale(self, locale):
-        """
-        Prefetch ProjectLocale and latest translation data for given locale.
-        """
-        from pontoon.base.models.project_locale import ProjectLocale
-
-        return self.prefetch_related(
-            Prefetch(
-                "project_locale",
-                queryset=(
-                    ProjectLocale.objects.filter(locale=locale).prefetch_related(
-                        "latest_translation__user", "latest_translation__approved_user"
-                    )
-                ),
-                to_attr="fetched_project_locale",
-            )
-        )
 
     def stats_data(self, locale=None) -> dict[int, dict[str, int]]:
         """Mapping of project `id` to dict with counts."""
@@ -274,26 +256,9 @@ class Project(models.Model, AggregatedStats):
         """Path where this project's VCS checkouts are located."""
         return join(settings.MEDIA_ROOT, "projects", self.slug)
 
-    def get_latest_activity(self, locale=None):
-        from pontoon.base.models.project_locale import ProjectLocale
-
-        if locale is None:
-            return (
-                self.latest_translation.latest_activity
-                if self.latest_translation
-                else None
-            )
-
-        pl_fetched = getattr(self, "fetched_project_locale", None)
-        project_locale = (
-            pl_fetched[0]
-            if pl_fetched
-            else ProjectLocale.objects.filter(project=self, locale=locale).first()
-        )
+    def get_latest_activity(self):
         return (
-            project_locale.latest_translation.latest_activity
-            if project_locale and project_locale.latest_translation
-            else None
+            self.latest_translation.latest_activity if self.latest_translation else None
         )
 
     def resource_priority_map(self):
--- a/pontoon/base/models/project.py
+++ b/pontoon/base/models/project.py
@@ -278,10 +278,6 @@ class Project(models.Model, AggregatedStats):
 
         return ProjectLocale.get_latest_activity(self, locale)
 
-    @property
-    def avg_string_count(self):
-        return int(self.total_strings / self.enabled_locales)
-
     def resource_priority_map(self):
         """
         Returns a map of resource paths and highest priorities of resource tags.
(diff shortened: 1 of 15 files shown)


mathjazz:
I'm getting this error on the locale pages:

```
[server] [INFO:django.server] 2025-02-11 19:45:53,493 "GET /de/ HTTP/1.1" 200 3956
[server] [ERROR:django.request] 2025-02-11 19:45:54,292 Internal Server Error: /de/ajax/
[server] Traceback (most recent call last):
[server]   File "/usr/local/lib/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner
[server]     response = get_response(request)
[server]                ^^^^^^^^^^^^^^^^^^^^^
[server]   File "/usr/local/lib/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response
[server]     response = wrapped_callback(request, *callback_args, **callback_kwargs)
[server]                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[server]   File "/app/pontoon/base/utils.py", line 84, in wrap
[server]     return f(request, *args, **kwargs)
[server]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[server]   File "/app/pontoon/teams/views.py", line 135, in ajax_projects
[server]     return render(
[server]            ^^^^^^^
[server]   File "/usr/local/lib/python3.11/site-packages/django/shortcuts.py", line 24, in render
[server]     content = loader.render_to_string(template_name, context, request, using=using)
[server]               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[server]   File "/usr/local/lib/python3.11/site-packages/django/template/loader.py", line 62, in render_to_string
[server]     return template.render(context, request)
[server]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[server]   File "/usr/local/lib/python3.11/site-packages/django_jinja/backend.py", line 59, in render
[server]     return mark_safe(self._process_template(self.template.render, context, request))
[server]                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[server]   File "/usr/local/lib/python3.11/site-packages/django_jinja/backend.py", line 105, in _process_template
[server]     return handler(context)
[server]            ^^^^^^^^^^^^^^^^
[server]   File "/usr/local/lib/python3.11/site-packages/jinja2/environment.py", line 1304, in render
[server]     self.environment.handle_exception()
[server]   File "/usr/local/lib/python3.11/site-packages/jinja2/environment.py", line 939, in handle_exception
[server]     raise rewrite_traceback_stack(source=source)
[server]   File "/app/pontoon/teams/templates/teams/includes/projects.html", line 28, in top-level template code
[server]     {% set latest_activity = project.get_latest_activity(locale) %}
[server]     ^^^^^^^^^^^^^^^^^^^^^^^^^
[server]   File "/app/pontoon/base/models/project.py", line 293, in get_latest_activity
[server]     if project_locale and project_locale.latest_translation
[server]                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[server] AttributeError: 'list' object has no attribute 'latest_translation'
```

eemeli:
> I'm getting this error on the locale pages:

Should be fixed now, was a silly mistake.

pontoon/base/models/project.py
```diff
@@ -276,11 +277,24 @@ def checkout_path(self):
...
+            )
 
-    @property
-    def avg_string_count(self):
-        return int(self.total_strings / self.enabled_locales)
+        pl_fetched = getattr(self, "fetched_project_locale", None)
+        project_locale = (
+            pl_fetched[0]
+            if pl_fetched
+            else ProjectLocale.objects.filter(project=self, locale=locale).first()
```
mathjazz, line 291: I think this should be sufficient (and save us some more DB queries)? We only get here if the locale is not available for the project, in which case we don't have latest activity data anyways.

```suggestion
            else None
```

mathjazz: If the above logic "works", you might be able to simplify the code a little (didn't actually test it):

```
pl_fetched = getattr(self, "fetched_project_locale", None)
if pl_fetched:
    project_locale = pl_fetched[0]
    return getattr(project_locale.latest_translation, "latest_activity", None)
return None
```

pontoon/projects/views.py
```diff
@@ -81,18 +84,29 @@ def project(request, slug):
 def ajax_teams(request, slug):
     """Project Teams tab."""
     project = get_object_or_404(
-        Project.objects.visible_for(request.user).available(), slug=slug
+        cast(
+            BaseManager[Project], Project.objects.visible_for(request.user).available()
+        ),
+        slug=slug,
```
mathjazz, line 90: Is this really necessary?

pontoon/projects/views.py
```diff
@@ -81,18 +84,29 @@ def project(request, slug):
...
-    locales = locales.prefetch_project_locale(project).order_by("name")
+    latest_trans_ids = project.project_locale.values_list(
+        "latest_translation_id", flat=True
+    )
+    latest_activities = {
+        trans.locale_id: trans.latest_activity
+        for trans in Translation.objects.filter(id__in=latest_trans_ids).select_related(
+            "user", "approved_user"
+        )
+    }
```
mathjazz, line 109: Shouldn't we use the same pattern in `ajax_projects`? We're making 24 DB queries there vs 7 here and I wonder if that's the culprit.

mathjazz (commented):
Nice work! Tested locally and on stage and performance seems better. I also didn't spot any issues with the stats and latest activity data.

Left some comments inline.

pontoon/base/models/project.py
eemeli: Ah, true, because this is only called from `ajax_projects()`.

pontoon/projects/views.py
eemeli: The `cast()` is a no-op at runtime, but with it the type of `project` is correct within this function.

eemeli:
I got `ajax_projects()` down to 4 queries for anonymous users, and up to 7 if you're logged in with admin rights. That's probably the best that can be done here easily.

pontoon/projects/templates/projects/widgets/project_list.html
```diff
@@ -22,7 +22,7 @@
     <tbody>
 {% endmacro %}
 
-{% macro item(project, main_link, chart_link, latest_activity, chart, class='limited', request=False, link_parameter=False) %}
+{% macro item(project, main_link, chart_link, latest_activity, chart, class='limited', allStrings=False, request=False, link_parameter=False) %}
```
mathjazz, line 25: Nit:
```suggestion
{% macro item(project, main_link, chart_link, latest_activity, chart, class='limited', all_strings=False, request=False, link_parameter=False) %}
```

pontoon/projects/views.py
mathjazz: While I appreciate the intent behind improving type safety, our Python codebase does not use static typing consistently. Introducing it in just this one place creates an inconsistency that might confuse other developers who are not expecting type hints.

mathjazz (commented):
Deployed latest version to stage.

pontoon/projects/views.py
eemeli: On the other hand, type hints were how I was able to catch the bug fixed in #3562. Without them, working with Pontoon code is much more difficult and more error-prone, in particular for developers not previously familiar with it.

We should be working towards more type hinting being available in Pontoon.

pontoon/teams/templates/teams/includes/projects.html
```diff
@@ -20,32 +20,37 @@
...
     {{ ProjectList.item(project, main_link, chart_link, latest_activity, chart, class, request=True, link_parameter=True) }}
   {% endfor %}
 
+  {% for project in projects_to_request %}
+    {% set main_link = url('pontoon.projects.project', project.slug) %}
+    {% set chart_link = url('pontoon.translate', locale.code, project.slug, 'all-resources') %}
+    {% set latest_activity = latest_activities.get(project.id) %}
+    {% set chart = project_stats.get(project.id, {'total': 0}) %}
+
+    {{ ProjectList.item(project, main_link, chart_link, latest_activity, chart, class='', allStrings=True, request=True, link_parameter=True) }}
```
eemeli, line 47:
```suggestion
    {{ ProjectList.item(project, main_link, chart_link, latest_activity, chart, class='', all_strings=True, request=True, link_parameter=True) }}
```

mathjazz:
@flodolo Could you take this for a spin on stage?

flodolo:
Seems to work fine for me, couldn't find inconsistencies.

mathjazz (approved):

PR #3562, opened 2025-02-11
Move badge notification code out of update_db_translations()

eemeli:
Code that's specifically and only for manualy file imports should not be in the sync internals.

This PR also fixes a bug introduced in #3522, as the type hint for `update_db_translations()` was not updated, and the early return was also not updated:

https://github.com/mozilla/pontoon/blob/1278167d98ac9810eccaf98698718988a23f2038/pontoon/sync/core/translations_from_repo.py#L81-L83

https://github.com/mozilla/pontoon/blob/1278167d98ac9810eccaf98698718988a23f2038/pontoon/sync/core/translations_from_repo.py#L295-L302


harmitgoswami (approved):
Works perfectly on my end! Thanks :^)

PR #3567, opened 2025-02-12
Highlighting: provide fallback code for older browsers (iOS 15)

### LINKED ISSUE #3550: Translate UI breaks on iOS 15
@simon0302010 spotted an issue where translation interface breaks (only gray background shows up) in Firefox and Safari.

I am so far unable to reproduce the problem. 

@simon0302010 Could you please share the URL of the broken page (with the grey background)? Also, are you logged into Pontoon or not when the page breaks? Thanks!

flodolo:
Fixes #3550

### DIFF
--- a/translate/src/modules/placeable/components/Highlight.tsx
+++ b/translate/src/modules/placeable/components/Highlight.tsx
@@ -168,29 +168,14 @@ export function Highlight({
   marks.sort((a, b) => a.index - b.index || b.length - a.length);
 
   if (search) {
-    let searchTerms;
+    let regexp: RegExp;
     try {
-      // Older browsers will fail parsing the regular expression, so it needs to
-      // be defined as string
-      const regex = new RegExp(
-        String.raw`(?<!\\)"(?:\\"|[^"])+(?<!\\)"|\S+`,
-        'g',
-      );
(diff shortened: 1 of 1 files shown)


flodolo:
@eemeli 
Thoughts? I couldn't figure out a way to replicate the current regex, and my head is already hurting. Should we give up highlighting in older browsers?

Even declaring the regex will break iOS 15, so it needs to be defined as a string.

eemeli (changes requested):
We don't need to have the fallback work differently from the default case, so something like this would be more appropriate:

```js
let regexp: RegExp
try {
  regexp = new RegExp(String.raw`(?<!\\)"(?:\\"|[^"])+(?<!\\)"|\S+`, 'g');
} catch {
  // Fallback for older browsers (e.g. iOS 15) not supporting lookbehind.
  regexp = /"(?:\\"|[^"])+"|\S+/g;
}
const searchTerms = search.match(regexp);
```

eemeli:
That'll have the highlighting be a little bit off on iOS 15 for messages that include a `\"` escaped doublequote, but those are really rather rare.

eemeli (approved):

PR #3570, opened 2025-02-15
Fix #3505: Do not link zero-value metrics in monthly reports

### LINKED ISSUE #3505: Monthly report: don't link on "0" results
For example, in the latest monthly report I've received

----

Completion: 100.0% (vs. 100.0% in the previous month)

Strings with missing translations: [0](https://pontoon.mozilla.org/it/all-projects/all-resources/?status=missing) (vs. 0 in the previous month)

Unreviewed suggestions: [0](https://pontoon.mozilla.org/it/all-projects/all-resources/?status=unreviewed) (vs. 0 in the previous month)

Unreviewed pretranslations: [0](https://pontoon.mozilla.org/it/all-projects/all-resources/?status=pretranslated) (vs. 0 in the previous month)

----

We should probably skip the link if the number is 0.

0xShubhamSolanki:
Problem
The monthly activity report email displays links for metrics like "missing translations" even when their value is 0, leading to unnecessary/empty filtered views.
Example: Strings with missing translations: 0 (vs. 0) had a hyperlink despite no actionable items.

Solution
Conditionally render links only when the metric value is > 0 for:
!)  Missing translations
2) Unreviewed suggestions
3) Unreviewed pretranslations
4) Maintain plain text display for zero values.

Changes
Updated status_item macro in monthly_activity_report.html to include value checks.
Ensured consistency in both HTML and text email templates.

Result:
All metrics at 0 (links removed)
Mixed zero/non-zero values (links only for >0 metrics)
Confirmed URL parameters work correctly for non-zero cases.


pontoon/messaging/templates/messaging/emails/monthly_activity_summary.html
```diff
@@ -106,11 +106,11 @@ <h2 class="team-activity">Team activity for {{ locale.name }} ({{ locale.code }}
 <p class="item {{ cls }}">
     <span class="bullet {{ url_parameter }}"></span>
     {{ label }}:
-    {% if url_parameter %}
+    {% if url_parameter and (locale.month_stats[attribute] or 0) > 0 %}
```
theoniko, line 109: I think locale.month_stats.get(attribute, 0) would be more readable.

pontoon/messaging/templates/messaging/emails/monthly_activity_summary.html
```diff
@@ -53,7 +53,7 @@
 <p>{{ settings.EMAIL_MONTHLY_ACTIVITY_SUMMARY_INTRO }}</p>
 {% endif %}
 
-<h2>You personal activity</h2>
+<h2>Your personal activity</h2>
```
mathjazz, line 56: Nice catch!

pontoon/messaging/templates/messaging/emails/monthly_activity_summary.html
mathjazz, line 109: It should be sufficient to do this:
```suggestion
    {% if url_parameter and locale.month_stats[attribute] %}
```

Which is also what @allie5 suggested in https://github.com/mozilla/pontoon/discussions/3568.

mathjazz (commented):
Thanks for the patch! Left a suggestion for improvement inline.

Please note that the issue was assigned to [allie5](https://github.com/allie5), so I'll reference her as a co-author.

mathjazz (approved):
Thanks for the patch!

I went ahead and apply the fixes myself in order to get them to the users ASAP.

PR #3579, opened 2025-02-25
add the yaml keyvalue format

gedaiu:
Hey everyone, I want to translate my ember.js frontend, which uses https://github.com/ember-intl/ember-intl , where the translations are defined as yaml files. This PR adds support for this format, which might be useful to some React apps, too.


eemeli:
I'm currently working on deeply refactoring how we parse and serialize resource files in Pontoon, so adding support for a new format should probably be done via [moz.l10n](https://github.com/mozilla/moz-l10n/) instead of being directly added into Pontoon.

That being said, we'll want to evaluate which formats to support directly in moz.l10n, and probably allow for some extensibility of both moz.l10n and Pontoon, so the parser for this format might end up in your own code instead. In other words, it would probably be best to open a Pontoon issue instead of a PR in which we could discuss your request.

mathjazz:
👋 Hey @gedaiu, I've filed https://github.com/mozilla/pontoon/issues/3584. Let's continue there.

PR #3580, opened 2025-02-25
Replace hardcoded source locale en-US from the sync task

gedaiu:
Unfortunately, my reference translation files don't have the `en-US` locale and sync fails. This fixes my issue. 

I'm looking forward to your feedback!


eemeli (changes requested):
This can't be the only place where we've hard-coded an expectation of the source locale being English. There's also at least [this](https://github.com/mozilla/pontoon/blob/e08335ac5a384e2c582f07444f22dce514120198/pontoon/sync/core/paths.py#L38) earlier in the sync, and the places where we communicate with machine translation.And probably elsewhere as well.

If we're going to fix this, the fix ought to be proper.

gedaiu:
I was not aware of this. Only this change was necessary to make my project work. If you give me all the places where this needs to be fixed, I can spend another day or two fixing this issue.

eemeli:
I think identifying the places where we have a built-in expectation of the source locale being `en` or `en-US` is in fact a large part of the work required here, coupled with the need to then define ways of retaining and passing around the source locale information in a way that isn't done atm.

To be clear, this is functionality that we'd be happy to have in Pontoon, it's just that Mozilla's own use does not require it at least for now, and so our prioritisation for this is rather low.

gedaiu:
@eemeli, I see. Could you merge this as it is because it is enough to fix my issues and does not break any existing functionality? Otherwise, I can keep it on my fork and use it from there.

eemeli:
I discussed this a bit with @mathjazz, and I'm afraid we'd prefer not merging this partial patch. Doing so is far too likely to lead to issues later for anyone depending on it, as there are many places where the assumption is hard-coded.

mathjazz:
Hi @gedaiu,

Thanks for taking the time to submit this pull request — we really appreciate your effort and willingness to contribute!

As Eemeli mentioned, we've decided not to move forward with merging this PR. While it touches on the right area, it doesn't fully solve the underlying issue, and the proper solution would require a more comprehensive approach that goes beyond the current scope.

That said, your initiative is truly appreciated, and we hope you’ll continue contributing. If you're interested, we’d be happy to [point you](https://chat.mozilla.org/#/room/#pontoon:mozilla.org) to something more self-contained that could use attention.

Issue #3538, opened 2025-01-23
All message formats should be represented the same way


eemeli:
Currently:
- Plurals in Gettext .po files are stored using `string` and `string_plural` for the source entity, and with separate translations for each plural variant.
- Fluent messages with multiple patterns in its value and attributes are represented as a single entity, and with a single translation for the whole message.
- All other formats allow only for a single pattern for each message.

With this structure, adding plural support for formats like iOS and Android is difficult, and extending support to powerful new syntaxes like MF2 is difficult.

We should refactor the data models used within Pontoon to be able to represent all messages in all formats the same way, and to make future extensibility easier.

These changes are enabled by the sync refactor in #3312, and the ongoing work in [moz.l10n](https://github.com/mozilla/moz-l10n/) to define the new data model.

eemeli:
As a first step, I'm working on replacing at least the serialisers (and potentially the parsers) with their moz.l10n equivalents, without touching the database.

The likely next step after that will be introducing the new data model in parallel and as write-only, next to the current models. Once that's in place, it should be easier for us to verify and validate that the code behaves as expected, before switching over completely to the new data models.

mathjazz:
> The likely next step after that will be introducing the new data model in parallel and as write-only, next to the current models. Once that's in place, it should be easier for us to verify and validate that the code behaves as expected, before switching over completely to the new data models.

Could we also fix https://github.com/mozilla/pontoon/issues/2755 in that step? It could be helpful in the validating step.

mathjazz:
The remaining work here is tracked under https://github.com/mozilla/pontoon/issues/4210.

PR #3587, opened 2025-02-28
Use moz.l10n for serialization

eemeli:
For context see #3538

~Filing initially as a draft; will update once mozilla/moz-l10n#58 is merged and a new moz.l10n release is available.~
_Edit: Done._

This PR replaces the file serialisers with their moz.l10n equivalents. Serialization is now done by first parsing the corresponding source file, and then modifying it according to the state of the database. This means that any comments in target files are lost, including obsolete messages in gettext files. The oddities required by our various file formats are encapsulated within `set_translations()` in `pontoon.sync.core.translations_to_repo`.

Dropping all serialization concerns from `pontoon.sync.formats` allows for the remaining code to be significantly simplified, removing e.g. the need to parse the source file when parsing a target file. The `ParsedResource` class and its subclasses are dropped, and `VCSTranslation` is no longer subclassed for each format.


mathjazz:
Deployed to stage.

pontoon/sync/core/translations_to_repo.py
```diff
@@ -1,35 +1,69 @@
...
+# and changes to POT-Creation-Date cause unnecessary churn.
+# Pontoon itself is the appropriate reference.
+gettext_trim_headers = (
+    "Language-Team",
+    "Last-Translator",
+    "PO-Revision-Date",
+    "POT-Creation-Date",
+    "Report-Msgid-Bugs-To",
+    "X-Generator",
+)
```
mathjazz, line 44: Why is this handled in Pontoon, rather than in `moz.l10n`?

pontoon/sync/core/translations_to_repo.py
```diff
@@ -1,35 +1,69 @@
...
+# TODO: This should almost certainly be handled better
+ios_locale_map = {
+    "ga-IE": "ga",
+    "nb-NO": "nb",
+    "nn-NO": "nn",
+    "sat": "sat-Olck",
+    "sv-SE": "sv",
+    "tl": "fil",
+    "zgh": "tzm",
+}
```
mathjazz, line 57: Same question as above. As demonstrated by the links in the comment, this isn't Pontoon-specific behavior.

pontoon/sync/core/translations_to_repo.py
```diff
@@ -223,3 +240,151 @@ def update_changed_resources(
                 log.error(f"{lc_scope} Update failed: {error}")
                 continue
     return count, updated_locales, translators
+
+
+def set_translations(
```
mathjazz, line 245: Once the new data model is in place, I assume this function will be much simpler.

mathjazz (commented):
Looks good. A couple of hints / questions inline.

This still needs to be tested.

pontoon/sync/core/translations_to_repo.py
eemeli: Because it's so opinionated, and specific to how the data is processed by Pontoon.

mathjazz:
Made these 3 changes on stage (appended "1" to existing translations):
https://mozilla-pontoon-staging.herokuapp.com/sl/pontoon-test/aboutAccounts.properties/?string=313684
https://mozilla-pontoon-staging.herokuapp.com/sl/pontoon-test/amo.po/?string=314547
https://mozilla-pontoon-staging.herokuapp.com/sl/pontoon-test/activities.xliff/?string=314494

This was committed, left comments next to the lines that surprised me:
https://github.com/mozilla-l10n/pontoon-test/commit/2ce6e76ec871fc37b93027cf4a97be39fc67609e#r153265214

pontoon/sync/core/translations_to_repo.py
eemeli: It's not Pontoon-specific, but it's really hacky, and smells way too much like cargo cult programming to me. If we were to include this in moz.l10n, removing it would be a breaking change. Dropping it later from Pontoon would not be.

eemeli:
> [Firefox for Android](https://github.com/mozilla-l10n/android-l10n/commit/34899fd04242034ea9d7f2d44e931bba6e585e4b): diff is a lot larger than I expected, but the content looks the same in the scripts I used to extract strings for Transvision.

The diff size here is mostly due to moz.l10n normalizing comments with a slightly different style than is currently used, and using `&lt;` escaping for HTML, rather than wrapping the whole message with `<![CDATA[...]]>`. It's also exaggerated a bit on GitHub as even with `?w=1` the whitespace-only change lines are not all deduplicated.

> I don't have a gettext project set up locally, so I'd ask someone else to do a test with a live one (accounts or AMO).

I've tested locally extensively with AMO, SUMO, and mozilla-accounts, and each of these works well. The diff for each consists mostly of:
- AMO: Adding context comments
- SUMO: Dropping obsolete commented-out `# ~...` entries
- accounts: Dropping unnecessary `##` Fluent section headers.

mathjazz:
> I replied there. Short answer to all is that the oddities are all specific to the pontoon-test project, and do not show up in production, or in files produced by sensible tooling.

Thanks for verifying that!

mathjazz:
Tested a couple more PO files locally and don't see any issues.

mathjazz (approved):

PR #3599, opened 2025-03-13
Only PMs should be able to pin and unpin comments

### LINKED ISSUE #3597: Any authenticated user can pin or unpin comments
See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1953061

mathjazz:
Fix #3597.

While frontend prevents non-PMs to pin or unpin comments, the backend allows any authenticated user to do so.


flodolo (approved):
Pending tests, the logic looks good.

flodolo (approved):

PR #3600, opened 2025-03-13
Only PMs can unapprove, unreject and delete translations

### LINKED ISSUE #3598: Any authenticated user can unapprove, unreject or delete translations
See:
https://bugzilla.mozilla.org/show_bug.cgi?id=1951879

mathjazz:
Fix #3598.


flodolo (approved):
Pending tests.

Took a while for the brain to parse the 2nd and 3rd condition, but it makes sense. It only returns False when the string is approved and the user has permissions (`not True or not True`).

flodolo (approved):

PR #3604, opened 2025-03-18
Create missing folders in path before writing translated file

### LINKED ISSUE #3602: Sync failing if path does not exist (missing folders)
Polish has been picking up translations for mozilla.org lately. Turns out Pontoon is submitting translations, but not creating some new files (not sure if this is about project configs).

https://github.com/mozilla-l10n/www-l10n/commits/master/pl

For example, https://pontoon.mozilla.org/pl/mozillaorg/en/firefox/browsers/mobile/android.ftl/ has been translated for a while, but there's no trace in the repository.

flodolo:
Fixes #3602

### DIFF
--- a/pontoon/sync/core/translations_to_repo.py
+++ b/pontoon/sync/core/translations_to_repo.py
@@ -3,8 +3,8 @@ import logging
 from collections import defaultdict
 from datetime import datetime
 from itertools import groupby
-from os import remove
-from os.path import commonpath, isfile, join, normpath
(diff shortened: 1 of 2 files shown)


flodolo:
Thanks for the test, I was searching for calls to `update_changed_resources()` and assumed there were none.

Confirmed that it fails without the new line of code, and passes with it.

mathjazz (approved):

PR #3609, opened 2025-03-31
Calculate stats only on enabled projects by default, add `-all` option

### LINKED ISSUE #3606: manage.py calculate_stats should ignore disabled project
Is there a reason to include disabled project? Right now the command has to go through 115 projects in prod, instead of 28.

flodolo:
Fixes #3606

### DIFF
--- a/pontoon/base/management/commands/calculate_stats.py
+++ b/pontoon/base/management/commands/calculate_stats.py
@@ -26,14 +26,29 @@ class Command(BaseCommand):
         See bug 1470337 for more details.
         """
 
-    def handle(self, *args, **options):
-        # Start with enabled projects in ascending order of resource count
-        projects = Project.objects.annotate(resource_count=Count("resources")).order_by(
-            "disabled", "resource_count"
+    def add_arguments(self, parser):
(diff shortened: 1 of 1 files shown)


pontoon/base/management/commands/calculate_stats.py
```diff
@@ -26,14 +26,29 @@ class Command(BaseCommand):
...
+                .annotate(resource_count=Count("resources"))
+                .order_by("resource_count")
+            )
+        log.info(f"Calculating stats for {action_txt} ({len(projects)})...")
+
         for project in projects:
             update_stats(project)
 
-        log.info("Calculating stats complete for all projects.")
+        log.info("Calculating stats complete.")
```
flodolo, line 54: Technically I could reuse `action_txt` here, but don't see much of a point.

flodolo:
There's no documentation about commands besides what's scheduled. We could probably add [something here](https://mozilla-l10n.github.io/documentation/tools/pontoon/index.html), but I wonder if it's useful, given only a handful of people have the required access anyway (and they can look at code).

PR #3615, opened 2025-04-10
Create TranslatedResources for all DB (rather than VCS) locales

### LINKED ISSUE #3523: New locale in existing (Sicilian, Mozilla accounts) stuck on "Not synced yet"
Locale was added a few days ago (around Jan 10), but it's still showing up as "Not synced yet".

A manual sync doesn't show any error in the logs.  ~Even a forced sync didn't make the project show up as available.~

mathjazz:
Fix #3523.

### DIFF
--- a/pontoon/sync/core/checkout.py
+++ b/pontoon/sync/core/checkout.py
@@ -101,8 +101,8 @@ def checkout_repos(
     target: Checkout | None = None
 
     # If the project using configuration file has unsynced locales, perform forced sync
-    if project.configuration_file:
-        force = force or project.unsynced_locales
+    if not force and project.configuration_file and project.unsynced_locales:
+        force = True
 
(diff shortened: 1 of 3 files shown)


pontoon/sync/core/checkout.py
```diff
@@ -99,6 +99,11 @@ def checkout_repos(
     """
     source: Checkout | None = None
     target: Checkout | None = None
+
+    # If the project using configuration file has unsynced locales, perform forced sync
+    if project.configuration_file:
+        force = force or project.unsynced_locales
```
eemeli, line 105: Better to avoid the query if `force` is already true, and keep `force` as a boolean value rather than a queryset.
```suggestion
    if not force and project.configuration_file and project.unsynced_locales:
        force = True
```

eemeli (approved):

PR #3617, opened 2025-04-13
Feat account disable page

### LINKED ISSUE #3153: Add ability to provide motivation for disabling a user
In #3152, we added the ability to disable a user. 

We should also communicate the reason for it to the user and ways to contact us on the Account disabled page.

We could use `settings.DEFAULT_FROM_EMAIL` for that.

RafaelJohn9:
Fix #3153.

> This patch doesn't work as expected locally, because if the is_active field is False, the user won't be authenticated during the request lifecycle, and user.is_authenticated will always be False. See the [docs](https://docs.djangoproject.com/en/5.1/ref/contrib/auth/#django.contrib.auth.models.User.is_active) for more details.
> We use a different authentication backend on stage / prod, which already serves inactive accounts with the [account_inactive.html](https://docs.allauth.org/en/latest/account/views.html#login) page after they try to log in. Sorry for not noticing this earlier!
> I suggest we drop the middleware changes (incl. the test) as they are not necessary, and override the default template used by django-allauth (see https://github.com/mozilla/pontoon/tree/main/pontoon/base/templates/django). Note that these need to be Django, rather than jinja2 templates.

did this as per the guidelines above


RafaelJohn9:
Sorry I  wasn't able to test this change locally, but as per the comment given above, seems that it can be checked on a staging env.

mathjazz:
This is how the page looks:
https://mozilla-pontoon-staging.herokuapp.com/accounts/inactive/

PR #3618, opened 2025-04-14
Improve sync logging

eemeli:
During sync, identify changed/removed/added entities, and separately list translations rejected due to repo changes.

### DIFF
--- a/pontoon/sync/core/entities.py
+++ b/pontoon/sync/core/entities.py
@@ -128,13 +128,13 @@ def update_resources(
         f"[{project.slug}] Changed source files: {', '.join(res.path for res in changed_resources)}"
     )
 
-    prev_entities = {
+    prev_entities: dict[tuple[str, str], Entity] = {
         (e.resource.path, e.key or e.string): e
         for e in Entity.objects.filter(resource__in=changed_resources, obsolete=False)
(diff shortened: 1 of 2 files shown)


pontoon/sync/core/entities.py
```diff
@@ -128,32 +128,33 @@ def update_resources(
...
+    next_not_prev = next_entities.keys() - prev_entities.keys()
+
+    obsolete_entities: list[Entity] = []
+    log_rm: dict[str, list[str]] = defaultdict(list)
+    for key, prev_ent in prev_entities.items():
+        if key in prev_not_next:
+            prev_ent.obsolete = True
+            prev_ent.date_obsoleted = now
+            obsolete_entities.append(prev_ent)
+            log_rm[key[0]].append(key[1].replace("\n", "\\n"))
```
mathjazz, line 156: Something like this would be more readbale:

```suggestion
            log_rm[path].append(entity.replace("\n", "\\n"))
```

mathjazz (commented):
While we're at it, it'd be great to also store these changes in the `SyncLog`.

That would take us closer to fixing https://github.com/mozilla/pontoon/issues/2023.

mathjazz (approved):

Issue #3543, opened 2025-01-24
Add ability to toggle between enabled and disabled projects in /admin


mathjazz:
Changing settings for disabled projects doesn't happen often (almost never), so we should hide them from the default view on /admin.

This same UI pattern could be used as is already in use on the team dashboard to toggle between projects that are enabled and not enabled for the team:
https://pontoon.mozilla.org/sl/

![Image](https://github.com/user-attachments/assets/63d72acf-5b9f-483c-90ee-ff3831ea0266)

The button text should probably be just "Disabled projects >".

flodolo:
> This same UI pattern could be used as the toggle between enabled and not enabled projects for the team on the team dashboard:

What does "enabled/disabled" mean at a team level? Opted-in or not? If that's the case, I don't see practical benefits in doing that (you would also need to differentiate between project that can be opted-in).

mathjazz:
This is a reference to the existing UX patter we can use here. Rephrased the sentence above.

RafaelJohn9:
Would love to work on this

mathjazz:
Fixed in https://github.com/mozilla/pontoon/pull/3622

PR #3622, opened 2025-04-17
Feature: Toggle between enabled and disabled project

RafaelJohn9:
addresses #3543 

Sample video of how the toggle disable/enable projects in the admin page looks like:

https://github.com/user-attachments/assets/0afdb02f-a29a-469f-a680-c0f761c17504

### DIFF
--- a/pontoon/administration/static/js/admin.js
+++ b/pontoon/administration/static/js/admin.js
@@ -1,0 +1,20 @@
+$(function () {
+  console.log('Admin JS loaded');
+  $('#toggle-projects').click(function () {
+    console.log('Button is clicked');
+    const button = $(this);
+    const currentShowDisabled = button.data('show-disabled') === true;
+    const nextShowDisabled = !currentShowDisabled;
+
+    // Update button text and data attribute
+    button.data('show-disabled', nextShowDisabled);
+    button.text(
+      nextShowDisabled ? 'Enabled projects >' : 'Disabled projects >',
+    );
+
+    // Reload the page with the appropriate query param
+    const url = new URL(window.location.href);
+    url.searchParams.set('show_disabled', nextShowDisabled);
+    window.location.href = url.toString();
+  });
+});
--- a/pontoon/administration/static/js/admin.js
+++ b/pontoon/administration/static/js/admin.js
@@ -4,11 +4,13 @@ document.addEventListener('DOMContentLoaded', () => {
   const disabledSection = document.querySelector('.disabled-projects');
 
   const showingDisabled = toggleBtn.dataset.showDisabled === 'true';
+  const iconLeft = '<span class="fa fa-chevron-left"></span>';
+  const iconRight = '<span class="fa fa-chevron-right"></span>';
 
   // Set initial text based on data attribute
-  toggleBtn.textContent = showingDisabled
-    ? toggleBtn.dataset.textShowEnabled
-    : toggleBtn.dataset.textShowDisabled;
+  toggleBtn.innerHTML = showingDisabled
+    ? iconLeft + toggleBtn.dataset.textShowEnabled
+    : toggleBtn.dataset.textShowDisabled + iconRight;
 
   toggleBtn.addEventListener('click', () => {
     const showingDisabled = toggleBtn.dataset.showDisabled === 'true';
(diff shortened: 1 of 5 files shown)


mathjazz:
Thanks for the patch.

Ideally, we should make the toggle use AJAX. Please log in to Pontoon and open https://pontoon.mozilla.org/sl/. You should see Request more projects button, please check it out.

There's also a slight UI difference (the icon).

RafaelJohn9:
checking it out

pontoon/administration/templates/admin.html
```diff
@@ -18,22 +18,39 @@
...
-      {% set chart_link = url('pontoon.admin.project', project.slug) %}
-      {% set latest_activity = project.get_latest_activity() %}
-      {% set chart = project_stats.get(project.id, {'total': 0}) %}
+      <a class="add button small" href="{{ url('pontoon.admin.project.new') }}">Add new
+        project</a>
+      <button class="button small request-toggle"
+                    id="toggle-projects"
+                    data-show-disabled="{{ 'true' if show_disabled else 'false' }}"
+                    data-text-show-disabled="Disabled projects >"
+                    data-text-show-enabled="< Enabled projects">
```
mathjazz, line 27: Note that we use [icons](https://github.com/mozilla/pontoon/blob/main/pontoon/teams/static/js/request.js#L15-L16) in the rest of the UI rather than < and > characters.

mathjazz (commented):
Nice job, works very well!

Left some ideas for improvement.

RafaelJohn9:
a min fixing sth :handshake:  :+1:

RafaelJohn9:
I was trying to  avoid the `span` tag inside the `js`  file , however inorder have the kind of toggling we have currently,  it seems that  it maybe  a little less seamless.

https://github.com/user-attachments/assets/a5c8c903-7760-44cf-b36b-313ec9b5436f

### Where my thoughts are:

1. Fetch content from the `toggle-button` tag (would  contain two span tags for `<` and `>`.
2. if the `data-*` text is to be appended in the beginning:
          - Remove current span tags inside
          - Append  the text 
          - append the `span` tags: one having display turned to `None`
3. else:
         - Change display of the appropriate `span` tag to None
         -  Append the text

pontoon/administration/templates/admin.html
```diff
@@ -23,32 +23,28 @@
...
-                {% set chart_link = url('pontoon.admin.project', project.slug) %}
-                {% set latest_activity = project.get_latest_activity() %}
-                {% set chart = project_stats.get(project.id, {'total': 0}) %}
-                {{ ProjectList.item(project, main_link, chart_link, latest_activity, chart) }}
+            {% for project in projects %}
+          {% set main_link = url('pontoon.admin.project', project.slug) %}
+          {% set chart_link = url('pontoon.admin.project', project.slug) %}
+          {% set latest_activity = project.get_latest_activity() %}
+          {% set chart = project_stats.get(project.id, {'total': 0}) %}
+          {{ ProjectList.item(project, main_link, chart_link, latest_activity, chart) }}
```
mathjazz, line 38: Seems like we lost indent here, and now the loop body is dedented compared to loop start / end.

pontoon/administration/templates/admin.html
```diff
@@ -23,32 +23,28 @@
...
                     data-show-disabled="{{ 'true' if show_disabled else 'false' }}"
-                    data-text-show-disabled="Disabled projects >"
-                    data-text-show-enabled="< Enabled projects">
+                    data-text-show-disabled="Disabled projects "
+                    data-text-show-enabled=" Enabled projects">
       </button>
 
     </menu>
-        <div class="project-list enabled-projects">
+        {% macro render_project_list(projects, project_stats) %}
```
mathjazz, line 31: In Jinja2, macros are almost always defined at the top level of the file — not inside blocks, loops, or conditional statements.

pontoon/administration/static/js/admin.js
```diff
@@ -0,0 +1,27 @@
...
+    enabledSection.style.display = showingDisabled ? 'block' : 'none';
+    disabledSection.style.display = showingDisabled ? 'none' : 'block';
+
+    toggleBtn.innerHTML = showingDisabled
+      ? toggleBtn.dataset.textShowDisabled + iconRight
+      : iconLeft + toggleBtn.dataset.textShowEnabled;
+
+    toggleBtn.dataset.showDisabled = (!showingDisabled).toString();
+  });
+});
```
mathjazz, line 27: The code seems to work. I wonder however why don't we just use the same code that's already used for requesting projects / pretranslation [here](https://github.com/mozilla/pontoon/blob/main/pontoon/teams/static/js/request.js)?

I'm not saying it's the best implementation of the functionality, but having two implementations of virtually same functionality is IMHO not a step in the right direction.

Thoughts?

mathjazz (commented):
Nice job! Left some more comments.

pontoon/administration/templates/admin.html
RafaelJohn9: will fix that :+1:

pontoon/administration/static/js/admin.js
RafaelJohn9: yeah agreed, :handshake: , Will refactor my approach :+1:

mathjazz (approved):
Great work, thanks for the patch and for keeping up with the long review process! 😄

Issue #3625, opened 2025-04-21
Translations with warnings are submitted and show no warning in Pontoon


flodolo:
[This string](https://pontoon.mozilla.org/tl/firefox-for-android/all-resources/?search=reduce_cookie_banner_details_panel_title_unsupported_site_request_2&search_identifiers=true&string=279732)

Reference

```
%1$s can’t automatically reject cookie requests on this site. You can send a request to support this site in the future.
```

Was translated as 

```
Hindi kusang makakapagtanggi ng cookie request sa Firefox sa site na ito. Maaari kang magpadala ng request para suportahan ang site na ito sa hinaharap.
```

The translation was accepted, no warning, and submitted to the repository (this is me [rejecting the translation](https://github.com/mozilla-l10n/android-l10n/commit/826566eeaa8bf3e0da2378e70eec521156344043) this morning).

I double-checked compare-locales, and the string is flagged as a warning (`Formatter %1$s not found in translation`) (same error as `%s`, since it's normalized for checks).

Note that this is, like #3589, breaking translation for all locales on both Nightly and Beta, so it's likely a P1.

flodolo:
I tried on [stage](https://mozilla-pontoon-staging.herokuapp.com/tl/firefox-for-android/all-resources/?search=reduce_cookie_banner_details_panel_title_unsupported_site_request_2&search_identifiers=true&string=279732), and I get the warning when submitting a suggestion. Once submitted, the warning is nowhere to be seen.

mathjazz:
The following works as expected:

1. The warning is diplayed on sumission.
2. The warning is stored in the DB: https://pontoon.mozilla.org/a/checks/warning/30957/change/.
3. Translations with warnings *are* committed to the repository.

So this seems like a frontend bug, where warnings are not displayed when openning a translation after it has been submitted.

flodolo:
> So this seems like a frontend bug, where warnings are not displayed when openning a translation after it has been submitted.

That's correct.

RafaelJohn9:
Hello here,

would like to be assigned this issue

PR #3629, opened 2025-05-02
Feature: WarningFailedChecksComponent && update EditorMenu

RafaelJohn9:
Addresses #3625 ,

Still not complete due to the inability  to reflect the changes locally in a browser. I  restarted the docker container, the `dist/translate.js` updated appropriately though reloading the browser window no changes reflects.

### DIFF
--- a/translate/src/modules/editor/components/EditorMenu.tsx
+++ b/translate/src/modules/editor/components/EditorMenu.tsx
@@ -17,6 +17,7 @@ import { FailedChecks } from './FailedChecks';
 import { FtlSwitch } from './FtlSwitch';
 import { KeyboardShortcuts } from './KeyboardShortcuts';
 import { TranslationLength } from './TranslationLength';
+import { WarningFailedChecks } from './WarningFailedChecks';
 
 /**
  * Shows a menu bar used to control the Editor.
@@ -99,6 +100,9 @@ function MenuContent() {
           </button>
         </Localized>
         <EditorMainAction />
+      </div>
+      <div>
+        <WarningFailedChecks />
       </div>
     </>
   );
--- a/translate/src/modules/editor/components/WarningFailedChecks.tsx
+++ b/translate/src/modules/editor/components/WarningFailedChecks.tsx
@@ -1,0 +1,35 @@
+import { Localized } from '@fluent/react';
+import React, { useContext } from 'react';
+import { FailedChecksData } from '~/context/FailedChecksData';
+import './FailedChecks.css';
+
+/**
+ * Displays a list of failed checks (errors and warnings) without any actions.
+ */
+export function WarningFailedChecks(): null | React.ReactElement<'div'> {
+    const { errors, warnings } = useContext(FailedChecksData);
+
+    if (!errors.length && !warnings.length) {
+        return null;
+    }
+
+    return (
+        <div className='failed-checks'>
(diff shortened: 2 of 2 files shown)


mathjazz:
> Addresses #3625 ,

Please use Fix #3625 [instead](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests?utm_source=chatgpt.com).

RafaelJohn9:
# Current Solution



https://github.com/user-attachments/assets/e93b8d0c-1588-4e1d-bf7b-7e0a42421b23

translate/src/modules/editor/components/Editor.tsx
```diff
@@ -14,5 +16,8 @@ export const Editor = () => (
     <TranslationForm />
     <MachinerySourceIndicator />
     <EditorMenu />
+    <FailedChecksProvider>
+      <TranslationWarningsErrors />
```
mathjazz: We already have the [FailedChecks](https://github.com/mozilla/pontoon/blob/076d2804e89d772b7979d860e2d6820580d11859/translate/src/modules/editor/components/FailedChecks.tsx) component. It seems like we just need to make sure it actually pops up rather than create another component?

RafaelJohn9: ooh, missed that will  look into it 👍

translate/src/modules/editor/components/Editor.tsx
```diff
@@ -6,12 +6,17 @@ import './Editor.css';
...
 
 export const Editor = () => (
   <div className='editor'>
     <NewContributorTooltip />
     <TranslationForm />
     <MachinerySourceIndicator />
     <EditorMenu />
+    <FailedChecksProvider>
+      <FailedChecks />
+    </FailedChecksProvider>
```
mathjazz, line 20: Have you verified that this fixes the problem? I'd be surprised if it does, because we already load the `FailedChecks` component in the `EditorMenu` element.

mathjazz:
OK, that's great news! Please make sure however that there's only once that we reference the `FailedChecks` component in the editor - right now it's where you added it, and also in the EditorMenu component. We should be able to remove one of these.

RafaelJohn9:
Hey @mathjazz , 

That's true,  we can remove one of the `FailedChecks`, though I have encountered some issues if we introduce this feature (not sure if they are bugs.)

This feature merely shows the previous  Failed Checks on the previous submission. Though

What if a user chooses to submit another `translation` before removing the popup ?

### 1)  Having the two `FailedChecks`

With the two `FailedChecks` we have the following output for the issue above: 


https://github.com/user-attachments/assets/cfa6d48a-6b97-40ab-a619-42f6fc5c3956

> it might not be visible though I'm pressing `Enter` to submit a translation.

### 2) Having one `FailedChecks`


https://github.com/user-attachments/assets/25a2dd99-439f-4a6c-841b-781122259402

> it might not be visible though I'm pressing `Enter` to submit a translation.


- You can notice on `# 2` we do not get the  Warning before submission,



What do you suggest to be the best course of action ?

mathjazz:
Hi @RafaelJohn9, thanks again for your work on this! Are you still planning to finish the PR?
If you’ve run into any issues or need guidance, we’d be happy to help you wrap it up.

PR #3634, opened 2025-05-07
Update messageformat & @messageformat/fluent dependencies to latest versions

### LINKED ISSUE #2877: Keep order of variants in FTL messages unchanged
A common use case for term variants are declension of a given term:
https://pontoon.mozilla.org/sl/firefox-monitor-website/all-resources/?search=-brand-Mozilla&string=191858

It is surprising that the nominative case, which is the default, isn't listed first.

Resources:
https://en.wikipedia.org/wiki/Slovene_declension

eemeli:
Fixes #2877 

This is effectively an intermediate step towards migrating the front-end to use `@mozilla/l10n`, as its data model more closely matches the latest `messageformat` data model -- but we're not quite there yet in terms of required capabilities.

This update should remove any remaining Fluent variant reordering.

My local env is a bit too messed up atm to test this manually, so we should deploy this to stage first.


eemeli:
This is currently deployed on [stage](https://mozilla-pontoon-staging.herokuapp.com/), and looks like it works as intended.

mathjazz:
The masculine/feminine order in the translation editor doesn't match the order in the source string (it does on main). That's risky and mistakes will be hard to catch automatically.

Posnetek zaslona 2025-05-08 ob 15 16 10

eemeli:
The order should now be retained. Redeploying to stage.

mathjazz:
> The order should now be retained. Redeploying to stage.

Works fine now, thanks!

(I was optimistically hoping the fix would also fix #2877. 😄 But that's a different beast.)

mathjazz:
> This update should remove any remaining Fluent variant reordering.

Was that supposed to include #2877?

mathjazz:
> This fixes #2877 for new translations, but does not include a data migration for existing translations.

That's good to know! Worth noting in the bug.

translate/src/utils/message/requiresSourceView.test.js
```diff
@@ -201,16 +201,4 @@ my-entry =
...
-    const input = `
-my-entry =
-    { FOO() ->
-        [foo] FOO
-       *[other] BAR
-    }`;
-    const message = parseEntry(input);
-
-    expect(requiresSourceView(message)).toEqual(true);
-  });
```
mathjazz, line 215: I guess dropping this validation will not cause too many headaches.

mathjazz (approved):

translate/src/utils/message/requiresSourceView.test.js
eemeli: Yeah, in practice we have no unknown selectors, and really we ought to show the selector in the message context.

PR #3635, opened 2025-05-08
Use MF2 syntax for plural gettext messages

### LINKED ISSUE #2304: Consolidate plural forms used for the source string in the string list and the editor
## [Bug 1726610](https://bugzilla.mozilla.org/show_bug.cgi?id=1726610)
Bug Reporter: @peiying2
CC:   @Pike, @flodolo, @mathjazz

Created attachment 9237093
Screen Shot 2021-08-19 at 10.55.09 AM.png

Issue: 
This appears to happen to languages using the same translation for singular and plural terms: https://pontoon.mozilla.org/vi/amo/LC_MESSAGES/djangojs.po/?string=66616

Notice the left panel shows the source string with a variable in singular form. The Translation field in the middle shows the source string in plural form.

Random checked East Asian and Southeast Asian languages, they all appear to have the same issue.

eemeli:
Closes #2304

This is an intermediate step towards #3538, de-specializing gettext .po plural message handling.

Previous to this PR, each gettext plural message had its source plural form stored as `Entity.string_plural`, and for each target locale, each of its variants was separately translated, with `Translation.plural_form` indicating the variant identifier. These fields are not used by any other format, and they are removed from the database in this change.

To represent plural source and target messages, Unicode MessageFormat (MF2) syntax representation is now used internally for all gettext messages, but this is never revealed to users. Placeholders like `{0}` or `%(type)s` are not parsed as such, but treated as a literal part of the string.

The translator front-end is modified to use the "rich editor" also for gettext plurals; this was previously only used by Fluent messages with variants. The FTL source view is not enabled for gettext plurals.

### DIFF
--- a/pontoon/checks/libraries/pontoon_db.py
+++ b/pontoon/checks/libraries/pontoon_db.py
@@ -2,6 +2,8 @@ from collections import defaultdict
 
 from fluent.syntax import FluentParser, ast
 from fluent.syntax.visitor import Visitor
+from moz.l10n.formats.mf2 import mf2_parse_message
+from moz.l10n.model import PatternMessage
 
 
 parser = FluentParser()
@@ -47,6 +49,18 @@ def run_checks(entity, original, string):
     if resource_ext == "po":
         if original.endswith("\n") != string.endswith("\n"):
             checks["pErrors"].append("Ending newline mismatch")
+        if string != "":
+            try:
+                msg = mf2_parse_message(string)
+                patterns = (
+                    (msg.pattern,)
+                    if isinstance(msg, PatternMessage)
+                    else msg.variants.values()
+                )
+                if any(not pattern or pattern == [""] for pattern in patterns):
+                    checks["pErrors"].append("Empty translations are not allowed")
+            except ValueError as e:
+                checks["pErrors"].append(f"Parse error: {e}")
 
     # Prevent empty translation submissions if not supported
     if string == "" and not entity.resource.allows_empty_translations:
--- a/pontoon/checks/tests/test_pontoon_db.py
+++ b/pontoon/checks/tests/test_pontoon_db.py
@@ -45,9 +45,18 @@ def test_empty_translations(get_entity_mock):
     """
     Empty translations shouldn't be allowed for some extensions.
     """
-    assert run_checks(get_entity_mock("po"), "", "") == {
-        "pErrors": ["Empty translations are not allowed"]
-    }
+    po_entity = get_entity_mock("po")
+    empty_error = {"pErrors": ["Empty translations are not allowed"]}
+    assert run_checks(po_entity, "", "") == empty_error
+    assert run_checks(po_entity, "", "{{}}") == empty_error
+    assert (
+        run_checks(po_entity, "", ".input {$n :number} .match $n * {{}}") == empty_error
+    )
+    assert (
+        run_checks(po_entity, "", ".input {$n :number} .match $n 1 {{}} * {{other}}")
+        == empty_error
+    )
+    assert run_checks(po_entity, "", "{{{||}}}") == {}
 
     assert run_checks(
         get_entity_mock("ftl", string="key = value"), "", 'key = { "" }'


eemeli:
I think this is finally actually ready for review. I needed to account for some Fluent changes due to mozilla/moz-l10n#83, and clean up some of the Locale utility methods -- the indirection there had been hiding a minor bug in our handling of locales with empty `cldr_locales`.

I left some of my in-progress changes out of this PR as they're not required for the gettext changes, and will be filing a separate follow-up PR for them.

The `moz.l10n` dependency will need a bump to a non-prerelease version before this should be merged.

pontoon/base/models/translated_resource.py
```diff
@@ -76,10 +75,10 @@ class TranslatedResource(models.Model):
     Resource representation for a specific locale.
     """
 
-    resource = models.ForeignKey(
+    resource: models.ForeignKey[Resource] = models.ForeignKey(
         Resource, models.CASCADE, related_name="translatedresources"
```
mathjazz, line 79: 🤡

pontoon/pretranslation/pretranslate.py
```diff
@@ -67,27 +68,21 @@ def get_pretranslations(entity, locale, preserve_placeables=False):
         authors = [services[service] for service in pretranslate.services]
         author = max(set(authors), key=authors.count) if authors else services["tm"]
 
-        return [(pretranslation, None, author)]
+        return (pretranslation, author)
+
+    # TODO
+    # elif entity.resource.format == "po":
```
mathjazz: Please file a regression issue instead.

pontoon/sync/core/stats.py
```diff
@@ -33,30 +33,14 @@ def update_stats(project: Project) -> None:
...
-                    LEFT OUTER JOIN "base_locale" loc ON (tr.locale_id = loc.id)
-                    LEFT OUTER JOIN "base_entity" ent ON (tr.resource_id = ent.resource_id)
-                    WHERE NOT ent.obsolete AND res.project_id = %s
-                    GROUP BY tr.id, loc.cldr_plurals
-                ) AS agg
-                WHERE agg.id = tr.id
+                UPDATE "base_translatedresource" tr
+                SET total_strings = res.total_strings
+                FROM "base_resource" res
+                WHERE tr.resource_id = res.id AND res.project_id = %s
```
mathjazz, line 43: Should we recalculate stats?

mathjazz (commented):
I'm half-way throught with the code review. Left some minor comments inline.

Submitting empty plural variants should be prohibited:
https://mozilla-pontoon-staging.herokuapp.com/sl/all-projects/all-resources/?status=translated&string=66626

This needs massive testing on stage, possibly involving other PMs. At least the following needs to be tested:
- Gettext plurals
- String search
- String filters
- Active translations
- Pretranslation

pontoon/pretranslation/pretranslate.py
eemeli: This was intended as a reminder for me while I was working on the PR. If you're fine with a regression issue, I'll be happy to do that instead of refactoring the pretranslation to run on top the MF2 data model -- and file a PR for that separately.

pontoon/sync/core/stats.py
eemeli: As a one-time task when merging this PR, yes.

requirements/default.txt
```diff
@@ -1,5 +1,5 @@
 # This file was autogenerated by uv via the following command:
-#    uv pip compile --generate-hashes --no-strip-extras requirements/default.in -o requirements/default.txt
+#    uv pip compile --generate-hashes --no-strip-extras default.in -o default.txt
```
flodolo: While CI doesn't fail anymore, please recompile from outside the requirements folder.

eemeli: Of course; I'll do that after the non-prerelease moz.l10n 0.8.0 is also ready to include here.

pontoon/pretranslation/pretranslate.py
eemeli: Filed as #3694.

translate/src/utils/message/index.ts
```diff
@@ -16,9 +16,8 @@ export { buildMessageEntry } from './buildMessageEntry';
 export { extractAccessKeyCandidates } from './extractAccessKeyCandidates';
 export { getEmptyMessageEntry } from './getEmptyMessage';
 export { getPlainMessage } from './getPlainMessage';
-export { getSimplePreview } from './getSimplePreview';
 export { editMessageEntry, editSource } from './editMessageEntry';
 export { findPluralSelectors } from './findPluralSelectors';
-export { parseEntry } from './parseEntry';
+export { parseEntry as parseEntry } from './parseEntry';
```
mathjazz: What's the purpose of this change?

eemeli: None. Probably an artifact of some refactoring that I'd missed cleaning up.
```suggestion
export { parseEntry } from './parseEntry';
```

mathjazz:
Pushed to stage for final testing.

PR #3636, opened 2025-05-09
Update command to start just the postgresql container

mathjazz:
This PR replaces the invalid command to start postgresql container with a working one:
* `up` starts the service as defined in `docker-compose.yml`
* `-d` means "detached" mode (runs in the background)
* `postgresql` is the name of the service


functionzz (approved):
Looks good, thanks Matjaz!

Issue #3605, opened 2025-03-24
Display Notification using 'real' date instead of nnn days/weeks ago


Patriccollu:
Hello,

I'm Corsican translator for [several Mozilla products](https://pontoon.mozilla.org/co/) and I often use the Pontoon's Notifications window to display the events history:
- [_product name_] updated with _nnn_ strings,
- [_product name_] due in _nn_ days,
- [_userid_] has sent you a message.

On the right of each event, you can see "3 DAYS AGO", "19 DAYS AGO" or "nn MONTHS, y WEEKS AGO".

![Image](https://github.com/user-attachments/assets/0e9f30a4-f127-481e-bdb5-307238189ce0)

When you see "3 DAYS AGO", it's easy to understand the date when the event occurred. But in the case of "19 DAYS AGO", or "4 MONTHS, 2 WEEKS AGO", it's really difficult to 'calculate' or even 'estimate' the real date of the event.

I would prefer to see a 'real' date, such as "2025/08/15".
And what would be more than perfect is to have an option to customize the date format!

Thanks for taking my request into consideration.

Best regards,
Patriccollu.

Patriccollu:
@mathjazz 
Hi Matjaž,

Thanks for your answer. Modifying the current date format is more important than having an option to customize it.

Thanks,
Patriccollu.

Patriccollu:
@mathjazz
Hi Matjaž,

I noticed that the _Weekly notifications summary_ report - sent by email - uses the same format. Did you plan to modify it?

In all the cases, PR #3637 answers my request!

Thanks,
Patriccollu.

mathjazz:
> I noticed that the _Weekly notifications summary_ report - sent by email - uses the same format. Did you plan to modify it?
> 
> In all the cases, PR [#3637](https://github.com/mozilla/pontoon/pull/3637) answers my request!

Hi  @Patriccollu, indeed the Weekly/Daily notifications summary uses the same code as the Notifications popup and the /notifications page, so that is also getting fixed by #3637.

Patriccollu:
@mathjazz It's perfect!

![Image](https://github.com/user-attachments/assets/d585c3d2-114a-4172-ab54-34a2818a0257)

PR #3637, opened 2025-05-09
Displays recent Notifications with date format instead of "nnn days/weeks ago"

functionzz:
This feature addresses issue #3605.

It effectively changes each Notification in the notifications page to display in the format "Month, Day Year" instead of "n DAYS AGO" if the notification is older than 7 days.

For example:
image


mathjazz:
Thanks for submitting the first PR! ✨

> This feature addresses issue #3605.

Note: You can link the PR to an issue by using one of the [following keywords](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests).

pontoon/contributors/templates/contributors/widgets/notifications_menu.html
```diff
@@ -48,7 +48,15 @@
...
-          <time class="timeago" datetime="{{ notification.timestamp }}">{{ notification.timesince() }} ago</time>
+
+          {% if not notification.is_old %}
+            <time class="timeago" datetime="{{ notification.timestamp }}">{{ notification.timesince() }}</time>
+          {% else %}
+            <time class="older-timeago" datetime="{{ notification.timestamp }}">
+              {{ notification.timestamp.strftime('%B') }} {{ notification.timestamp.day }}, {{ notification.timestamp.year }}
+            </time>
+          
+          {% endif %}
```
mathjazz, line 59: You could declare a macro on top of the file (see `menu()` as an example), so you don't need to repeat the same block of code three times. That will also reduce the chances to make mistakes like not including the trailing ` ago` in line 53.

mathjazz: Nit: why is the class name `older-timeago`? We're not showing time ago here, so maybe change it to `old-notification` (to match `notification.is_old`)?

pontoon/contributors/templates/contributors/widgets/notifications_menu.html
```diff
@@ -48,7 +48,15 @@
         {% if description and (description.startswith("Unreviewed suggestions") or notification.verb == "has reviewed suggestions" or notification.verb == "ignore") %}
           <span class="description">{{ description|safe }}</span>
 
-          <time class="timeago" datetime="{{ notification.timestamp }}">{{ notification.timesince() }} ago</time>
+
+          {% if not notification.is_old %}
+            <time class="timeago" datetime="{{ notification.timestamp }}">{{ notification.timesince() }}</time>
+          {% else %}
+            <time class="older-timeago" datetime="{{ notification.timestamp }}">
+              {{ notification.timestamp.strftime('%B') }} {{ notification.timestamp.day }}, {{ notification.timestamp.year }}
```
mathjazz, line 56: There's a [custom](https://github.com/mozilla/pontoon/blob/main/pontoon/base/templatetags/helpers.py#L163-L176) Jinja2 [filter](https://jinja.palletsprojects.com/en/stable/templates/#filters) we should be able to leverage here:

```suggestion
              {{ notification.timestamp|format_datetime("date") }}
```

pontoon/contributors/views.py
```diff
@@ -459,6 +461,10 @@ def notifications(request):
                     "notifications": [notification.id],
                 }
 
+        # checks if notification is older than 7 days and displays
+        notification.is_old = notification.timestamp < timezone.now() - timedelta(
```
mathjazz, line 465: `is_old` property is only used for formatting, not logic, so it should go to a template.

I'd suggest to instead add a [custom Jinja2 filter](https://github.com/mozilla/pontoon/blob/main/pontoon/base/templatetags/helpers.py).

mathjazz (commented):
Nice work! The code works as expected.

I've left some ideas on how to structure the code in a way that will make it more maintainable. Please let me know if you have any questions.

Once the changes are made, we'll sadly have to replicate them in https://github.com/mozilla/pontoon/blob/main/translate/src/modules/user/components/UserNotification.tsx, because the translation view uses a separate frontend, so the notifications UI is implemented twice.

pontoon/contributors/templates/contributors/widgets/notifications_menu.html
functionzz: Interesting, I'll implement it this way

pontoon/contributors/templates/contributors/widgets/notifications_menu.html
functionzz: You're right, I'll just do a macro, and old-notification makes more sense.

functionzz: Side question: what does Nit mean?

pontoon/contributors/views.py
functionzz: Yeah I was wondering if the formatting should be implemented inside the template or the view, I'll make a custom filter instead

functionzz:
As a positive side effect of making the formatting go into a filter, the notification bell's menu dates have also changed to reflect the desired date behavior. Previously it wasn't working.

pontoon/contributors/templates/contributors/widgets/notifications_menu.html
mathjazz: Short for "nitpick". 😄 A very small, non-blocking comment — things that won’t break the code but could improve readability or consistency.

pontoon/base/templatetags/helpers.py
```diff
@@ -313,3 +313,10 @@ def default_if_empty(value, default=""):
 
     # Mark as safe to include HTML tags
     return value if value else mark_safe(default)
+
+
+@library.filter
+def is_old_notification(notification):
+    """Return the"""
```
mathjazz, line 320: It looks like the comment is cut off.

pontoon/contributors/templates/contributors/widgets/notifications_menu.html
```diff
@@ -23,6 +23,18 @@
...
 
+{% macro notification_time(notification) %}
+
+{% if notification|is_old_notification %}
+  <time class="old-notification" datetime="{{ notification.timestamp }}">
+    {{ notification.timestamp|format_datetime("date") }}
+  </time>
+{% else %}
+  <time class="timeago" datetime="{{ notification.timestamp }}">{{ notification.timesince() }}</time>
+
```
mathjazz, line 34: You could remove this line.

functionzz: Are you referring to the "time" element line?

pontoon/contributors/templates/contributors/widgets/notifications_menu.html
functionzz: Oh ok nice

pontoon/contributors/templates/contributors/widgets/notifications_menu.html
mathjazz: No, I meant the blank line. ;)

translate/src/modules/user/components/UserNotification.tsx
```diff
@@ -1,6 +1,11 @@
 import classNames from 'classnames';
 import parse from 'html-react-parser';
-import React, { useEffect, useRef, useState } from 'react';
+import React, {
+  BlockquoteHTMLAttributes,
```
mathjazz, line 4: I'm curious what is this change meant for?

translate/src/modules/user/components/UserNotification.tsx
```diff
@@ -11,6 +16,36 @@ type Props = {
...
+  date_iso: string;
+}
+
+const DateDisplay: React.FC<DateDisplayProps> = ({ date_iso }) => {
+  const date: Date = new Date(date_iso);
+  const formattedDate: string = date.toLocaleDateString('en-US', {
+    day: 'numeric',
+    month: 'long',
+    year: 'numeric',
+  });
```
mathjazz, line 29: We already have a `date` property everywhere we call `DateDisplay`, so we should just pass it here.

Then, we should use it in the title attribute also for old dates.

mathjazz (commented):
Works nicely! Just a couple of minor suggestions.

translate/src/modules/user/components/UserNotification.tsx
functionzz: Ah ok, lemme change that real quick

translate/src/modules/user/components/UserNotification.tsx
functionzz: I did not know I changed that my bad

translate/src/modules/user/components/UserNotification.tsx
```diff
@@ -11,6 +11,38 @@ type Props = {
...
+}
+
+const DateDisplay: React.FC<DateDisplayProps> = ({ date, date_iso }) => {
+  const formattedDate: string = new Date(date).toLocaleDateString('en-US', {
+    day: 'numeric',
+    month: 'long',
+    year: 'numeric',
+  });
+  const isOld: boolean =
+    new Date().getTime() - new Date(date).getTime() > 7 * 24 * 60 * 60 * 1000;
```
mathjazz, line 26: I haven't test it, but I find the python logic (`is_old_notification()`) easier to read. Thoughts?

```suggestion
    new Date(date).getTime() < new Date().getTime() - 7 * 24 * 60 * 60 * 1000;
```

mathjazz (commented):
Just a minor note, other than that it looks great!

translate/src/modules/user/components/UserNotification.tsx
functionzz: I'm fine either way, although I wrote it originally because I wanted to read it left-to-right like
Current Date minus Notification Date is greater than 7 days.

functionzz: Other than that, I'll commit the change cuz this way is perfectly understandable.

translate/src/modules/user/components/UserNotification.tsx
```diff
@@ -62,6 +84,8 @@ const Comment = ({
         }
       </Linkify>
     </div>
+
+    <DateDisplay date={date} date_iso={date_iso} />
```
mathjazz, line 88: This position of `DateDisplay` shoud remain where the position of `ReactTimeAgo` was, otherwise the timestamp could appear very low on notifications with custom messages for example.

translate/src/modules/user/components/UserNotification.tsx
```diff
@@ -104,6 +122,8 @@ const Other = ({
         dangerouslySetInnerHTML={{ __html: description.content }}
       />
     ) : null}
+
+    <DateDisplay date={date} date_iso={date_iso} />
```
mathjazz, line 126: Same as above.

mathjazz (commented):
Almost there!

(Sorry for not catching this earlier)

PR #3639, opened 2025-05-12
Feature/drop field permalink prefix

### LINKED ISSUE #3489: Drop permalink_prefix repo field as unused
Once #3312 lands, we will no longer have a use for `Repository.permalink_prefix`, and should remove it.

This task is left separate from the parent, to make it easier to roll back if any issues are discovered in production.

functionzz:
Attempts to resolve #3489.

It deletes the field "permalink_prefix" from the Repository model, conducts the relevant migration and removes all references to "permalink_prefix" in the codebase.

For example, the admin project page used to contain a field for "Download prefix or path to TOML file".
image

It is removed now:

image

Similarly, the same removal can be observed in the admin "Edit Project" section.


mathjazz (approved):
Excellent work!

Welcome to Pontoon, welcome to Mozilla! 😄

PR #3642, opened 2025-05-14
Fix #3497: Recipient count displays email recipients for email-only m…

### LINKED ISSUE #3497: Recipient count should only include email recipients when message sent as email only
When the message in the Messaging center is sent as an email only, the recipient count in the Send button and in the Sent message header should only include the count of email recipients. 

Currently includes the count of all users that match the filters, i.e. all users that would be recipients if the message was (also) sent as a notification.

functionzz:
…essages

This feature fixes #3497  the bug where if someone wished to send a message email only, the message center's review message section displayed the count of people who would recieve a message if it was both a notification and an email, not solely the email-only recipients.

Before:
image
image

After:
image

To be clear, solely selecting "Email" in the Message Type is the only way for this to work, which is in line with the logic described in the bug report.

### DIFF
--- a/pontoon/messaging/views.py
+++ b/pontoon/messaging/views.py
@@ -98,7 +98,7 @@ def get_recipients(form):
     recipients = User.objects.none()
 
     """
-    Filter recipients by user role, locale, project and email/notification opt:
+    Filter recipients by user role, locale, project and email opt-in status:
     - Contributors of selected Locales and Projects
     - Managers of selected Locales
     - Translators of selected Locales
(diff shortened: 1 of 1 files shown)


functionzz:
Small note, fetching the count is actually quite slow, maybe down the line this is something I can try to work on.

pontoon/messaging/views.py
```diff
@@ -98,10 +98,11 @@ def get_recipients(form):
     recipients = User.objects.none()
 
     """
-    Filter recipients by user role, locale and project:
+    Filter recipients by user role, locale, project and email/notification opt:
     - Contributors of selected Locales and Projects
     - Managers of selected Locales
     - Translators of selected Locales
+    - Email/Notification opt
```
mathjazz, line 105: Same as above

mathjazz (approved):
Well done!

Issue #3633, opened 2025-05-07
Outdated node version in pontoon documentation failing jest tests


functionzz:
Hello,

Recently, I was attempting to set up Pontoon with the Documentation where I ran into an issue where the test cases for the translate front-end application failed to pass:
![Image](https://github.com/user-attachments/assets/8d3a5387-fb7f-456a-a327-c805fcfff3de)

I did some digging and found out what could be the reason for the failed tests:
<img width="463" alt="Image" src="https://github.com/user-attachments/assets/df46c598-ae01-4b41-be17-660a7a010455" />

In this image, I believe the .at feature at Line 10 is not available in Node 14, which the documentation tells you to install:

<img width="657" alt="Image" src="https://github.com/user-attachments/assets/a60c03a5-1cb4-4b06-8765-c9ec3a6cfec9" />

This is located in [https://mozilla-pontoon.readthedocs.io/en/latest/dev/setup.html](url)

I fixed this by installing a later Node version, in my case Node v23.11.0.

Thank you!

PR #3643, opened 2025-05-15
Fix: Changed min node version from 14 to 16.6.0

RafaelJohn9:
Fix: #3633

According to https://nodejs.org/en/blog/release/v16.6.0, the `array.at` feature was introduced in `node v16.6.0`

Changed this in the docs.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -65,7 +65,7 @@ the ``URL`` will need to include both the PAT and username, e.g.
 JavaScript setup
 ================
 
-For working on the front-end, you need at least Node.js 16.6.0 and npm 7
+For working on the front-end, you need the latest versions of Node.js and npm
 (`installation instructions <https://docs.npmjs.com/downloading-and-installing-node-js-and-npm>`_).
 Parts of the front-end use `npm workspaces <https://docs.npmjs.com/cli/v7/using-npm/workspaces>`_,
 which are not supported by earlier npm versions.
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -65,7 +65,7 @@ the ``URL`` will need to include both the PAT and username, e.g.
 JavaScript setup
 ================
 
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.rst
```diff
@@ -65,7 +65,7 @@ the ``URL`` will need to include both the PAT and username, e.g.
 JavaScript setup
 ================
 
-For working on the front-end, you need at least Node.js 14 and npm 7
+For working on the front-end, you need at least Node.js 16.6.0 and npm 7
```
mathjazz, line 68: I'd drop the reference to the version number, otherwise we'll need to play maintenance catch-up every time it changes:

```suggestion
For working on the front-end, you need the latest versions of Node.js and npm
```

mathjazz (commented):
Thanks for the patch!

Please also change the virtualenv setup docs:
https://github.com/mozilla/pontoon/blob/2e3a60226f171d1ff19f5ce50b51cd8cebac442f/docs/dev/setup-virtualenv.rst

docs/dev/setup-virtualenv.rst
```diff
@@ -17,8 +17,7 @@ This guide assumes you have already installed and set up the following:
 1. `Git <https://git-scm.com>`__
 2. `Python 3.11 <https://www.python.org>`__
 3. `uv <https://docs.astral.sh/uv/getting-started/installation/#standalone-installer>`_
-4. `Node.js 18 <https://nodejs.org>`__ and `npm 9 <https://www.npmjs.com>`__ or
-   later
+4. `Latest versions of Node.js <https://nodejs.org>`__ and `npm <https://www.npmjs.com>`__
```
mathjazz, line 20: I'd drop the "Latest versions" here, since we also don't call out the versions for Git and uv.

```suggestion
4. `Node.js <https://nodejs.org>`__ and `npm <https://www.npmjs.com>`__
```

mathjazz (approved):
Nice job!

PR #3644, opened 2025-05-15
Use node 24 in Docker image

flodolo:


mathjazz (approved):
Nice work, much more elegant.

Tested and it works great.

PR #3650, opened 2025-05-21
Fix escaped quotes and newlines for Android strings

eemeli:
Fixes two related regressions from #3611, as literal `"` quotes are now showing up as `\\\"` double-escaped, and escaped newlines `\n` are showing up as `\\n`.

Also normalises the presentation of escaped newlines to always be followed by a literal newline, and drops any whitespace immediately after an escaped newline when serialising.


mathjazz (approved):
Tested and it looks good!

PR #3652, opened 2025-05-22
Use moz.l10n 0.7.4, with fixed Android string resource handling

eemeli:
Three main changes here for Android string handling:
1. The `\u` escapes use hexadecimal rather than decimal values.
2. Leading and trailing whitespace is trimmed.
3. HTML-ish content is serialised with a `<![CDATA[...]]>` wrapper.

Note that an `lxml` workaround for https://bugs.launchpad.net/lxml/+bug/2111509 requires capping its version range to max at its current version (5.4.0), as the workaround will need to be removed once the bug is fixed.

### DIFF
--- a/pontoon/base/migrations/0079_trim_android.py
+++ b/pontoon/base/migrations/0079_trim_android.py
@@ -1,0 +1,30 @@
+from django.db import migrations
+from django.db.models import Q
+
+
+def trim_android(apps, schema_editor):
+    Entity = apps.get_model("base", "Entity")
+    android_entities = Entity.objects.filter(resource__format="xml").filter(
+        Q(string__startswith=" ") | Q(string__endswith=" ")
+    )
+    for ent in android_entities:
+        ent.string = ent.string.strip()
+    Entity.objects.bulk_update(android_entities, ["string"])
+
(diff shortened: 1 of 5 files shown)


pontoon/base/migrations/0079_trim_android.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def trim_android(apps, schema_editor):
+    Entity = apps.get_model("base", "Entity")
+    android_entities = Entity.objects.filter(resource__format="xml").filter(
+        Q(string__startswith=" ") | Q(string__endswith=" ")
+    )
+    for ent in android_entities:
+        ent.string = ent.string.strip()
+    Entity.objects.bulk_update(android_entities, ["string"])
```
mathjazz, line 12: I assume you tested and there's no need for specifyig `batch_size` here.

mathjazz (approved):

flodolo:
This is the output on Italian, which looks good
https://github.com/mozilla-l10n/android-l10n/commit/9eafa78a364cc2ff8f22e6617c10ca5f631f2e6c

Issue #3521, opened 2025-01-10
Record changes to advanced search options in UXActionLog


flodolo:
Analyzing the log for #3503 is not particularly useful, because we have too much noise coming from bots.

As a first step, I think it would be enough to record an action when the user clicks "apply search options" in the dropdown, recording as metadata: locale, project, `search_*` parameters set to true.

This would give us a sense of which options are used, and how frequently. 

Measuring how frequently the menu is used compared to the standard search doesn't seem useful, since the options persist once set.

flodolo:
> Measuring how frequently the menu is used compared to the standard search doesn't seem useful, since the options persist once set.

Rethinking about this: maybe it makes sense to track how many searches are performed through the search field, and store as metadata the advanced search options, on top of the main filters (string statuses, extra filters). That would give us a better picture, but it would also be a lot more data.

As for the last point, maybe we need a retention policy for the UXActionLog (e.g. 2 years?).

mathjazz:
> As for the last point, maybe we need a retention policy for the UXActionLog (e.g. 2 years?).

We could do something like this:
https://mozilla-pontoon.readthedocs.io/en/latest/admin/deployment.html#sync-log-retention

OTOH, the `uxactionlog_uxactionlog` table size is small:

```
 public.insights_projectlocaleinsightssnapshot | 4197 MB
 public.base_translation                       | 4137 MB
 public.base_translationmemoryentry            | 2016 MB
 public.notifications_notification             | 1498 MB
 public.django_session                         | 1338 MB
 public.actionlog_actionlog                    | 884 MB
 public.insights_localeinsightssnapshot        | 227 MB
 public.base_entity                            | 90 MB
 public.uxactionlog_uxactionlog                | 54 MB
 public.sync_projectsynclog                    | 52 MB
 public.base_translatedresource                | 51 MB
...
```

mathjazz:
Closed in https://github.com/mozilla/pontoon/pull/3654.

PR #3654, opened 2025-05-23
Added UXActionLog to Entity Fetch

functionzz:
This commit addresses #3521 by adding a UXActionLog to the Entity fetch process when the translation module loads. It logs the search value, as well as advanced various search options that have been applied, with the intention to gather usage frequency statistics.

### DIFF
--- a/translate/src/modules/entities/actions.ts
+++ b/translate/src/modules/entities/actions.ts
@@ -5,6 +5,7 @@ import {
   fetchSiblingEntities,
 } from '~/api/entity';
 import { EntityTranslation } from '~/api/translation';
+import { logUXAction } from '~/api/uxaction';
 import { Location } from '~/context/Location';
 import { updateStats } from '~/modules/stats/actions';
 import type { AppDispatch } from '~/store';
@@ -67,6 +68,25 @@ export const getEntities =
     dispatch({ type: REQUEST_ENTITIES });
 
     const content = await fetchEntities(location, page);
+    const {
+      search,
+      search_exclude_source_strings,
+      search_identifiers,
+      search_match_case,
+      search_match_whole_word,
+      search_rejected_translations,
+    } = location;
+
+    if (search) {
+      logUXAction('Search Options Selected', 'Search Option Statistics', {
+        search,
+        search_exclude_source_strings: search_exclude_source_strings,
+        search_identifiers: search_identifiers,
+        search_match_case: search_match_case,
+        search_match_whole_word: search_match_whole_word,
+        search_rejected_translations: search_rejected_translations,
+      });
+    }
 
     if (content.entities) {
       dispatch({


translate/src/modules/entities/actions.ts
```diff
@@ -67,6 +68,25 @@ export const getEntities =
...
+      search,
+      search_exclude_source_strings,
+      search_identifiers,
+      search_match_case,
+      search_match_whole_word,
+      search_rejected_translations,
+    } = location;
+
+    if (search) {
+      logUXAction('Search Options Selected', 'Search Option Statistics', {
```
mathjazz, line 81: These would be silghtly more consistent with existing action and experiment names:

```suggestion
      logUXAction('Load: String list with search parameter', 'Search Options Statistics', {
```

mathjazz (commented):
Works nicely!

mathjazz:
@flodolo Please take it for a spin on stage.

mathjazz:
> We probably shouldn't record the search terms (it takes up space, it's not really useful data).

👍 

> The action type is not very accurate: `Search Options Selected`. We record any search, even without changing search options, so it's more like `Search in projects`?

We changed that to `Load: String list with search parameter`.

> This is recording also if I load a search URL directly, it doesn't need to go through the UI. Can we avoid that? If not, we still have the bot problem (and the size of DB will grow quickly). If we can't filter that out, can we record the data only if the user is authenticated?

The assumption is most bots (like web crawlers, scrapers, etc.) use command-line tools (like curl) or libraries in code (like Python’s requests) to make direct HTTP requests. These will not be recorded.

> Talking about DB size: do we have a way to clean up old UX action log data?

Yeah, but nothing automated. The table size is just 60 MB, though.

flodolo:
> The assumption is most bots (like web crawlers, scrapers, etc.) use command-line tools (like curl) or libraries in code (like Python’s requests) to make direct HTTP requests. These will not be recorded.

I'm confused… From my testing, they are recorded (e.g. load [this URL](https://mozilla-pontoon-staging.herokuapp.com/it/amo/all-resources/?search=test) directly, in a new tab), but they don't seem to be recorded in private mode (i.e. not logged in), so I'm guessing it's indeed excluding searches from non-logged in users?

flodolo:
With that said, this works for me in practice, although I don't know why ;-)

mathjazz:
> > The assumption is most bots (like web crawlers, scrapers, etc.) use command-line tools (like curl) or libraries in code (like Python’s requests) to make direct HTTP requests. These will not be recorded.
> 
> I'm confused… From my testing, they are recorded (e.g. load [this URL](https://mozilla-pontoon-staging.herokuapp.com/it/amo/all-resources/?search=test) directly, in a new tab)

That's not a command-line tool, nor a library. 😉

mathjazz:
> but they don't seem to be recorded in private mode (i.e. not logged in), so I'm guessing it's indeed excluding searches from non-logged in users?

That's correct:
https://github.com/mozilla/pontoon/blob/main/pontoon/uxactionlog/views.py#L10

If we want to keep it that way, we should at least get rid of the JS error when logged out.

mathjazz (approved):
Works like a charm, nice work!

PR #3662, opened 2025-06-07
Update docker compose command in Makefile

mathjazz:
The old docker-compose command (with a dash) is not installed by default in Docker Desktop versions starting from mid-2022.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,3 @@
-DC := $(shell which docker-compose)
 DOCKER := $(shell which docker)
 
 # *IMPORTANT*
@@ -54,7 +53,7 @@ build-translate: node_modules
 	npm run build -w translate
 
 build-server: server-env translate/dist
-	"${DC}" build --build-arg USER_ID=$(USER_ID) --build-arg GROUP_ID=$(GROUP_ID) server
+	"${DOCKER}" compose build --build-arg USER_ID=$(USER_ID) --build-arg GROUP_ID=$(GROUP_ID) server
 	touch .server-build
 
 server-env:
@@ -67,12 +66,12 @@ server-env:
 	fi
 
 setup: .server-build
-	"${DC}" run server //app/docker/server_setup.sh
+	"${DOCKER}" compose run server //app/docker/server_setup.sh
 
 run: translate/dist .server-build
-	"${DC}" up --detach
+	"${DOCKER}" compose up --detach
 	bash -c 'set -m; bash ./bin/watch.sh'
-	"${DC}" stop
+	"${DOCKER}" compose stop
 
 clean:
 	rm -rf translate/dist .server-build
@@ -81,7 +80,7 @@ clean:
 	@container=$$(${DOCKER} ps -q --filter ancestor=local/pontoon | head -n 1); \
 	if [ -z "$$container" ]; then \
   		echo "Trying to start the container" >&2; \
-		"${DC}" up --detach; \
+		"${DOCKER}" compose up --detach; \
   		container=$$(${DOCKER} ps -q --filter ancestor=local/pontoon | head -n 1); \
 		if [ -z "$$container" ]; then \
 			echo "Error: No container running based on local/pontoon. Try running 'make build'." >&2; \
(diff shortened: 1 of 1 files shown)


flodolo (approved):

Makefile
```diff
@@ -1,4 +1,3 @@
-DC := $(shell which docker-compose)
```
flodolo, line 1: On second thoughts, if we want to support `pacman` (and `pacman-compose`), we might be better off keeping DC?

Also, is `command -v` more portable/reliable than `which`?

Untested. We might want to be smart and search for pacman-compose if `DOCKER` is pacman, instead of doing the full fallback.

```makefile
DOCKER := $(shell command -v docker || command -v pacman || (echo "Error: neither docker nor pacman found" >&2 && exit 1))

DC := $(shell \
	if command -v docker > /dev/null && docker compose version > /dev/null 2>&1; then \
		echo "$$(command -v docker) compose"; \
	elif command -v pacman-compose > /dev/null; then \
		echo "$$(command -v pacman-compose)"; \
	elif command -v docker-compose > /dev/null; then \
		echo "$$(command -v docker-compose)"; \
	else \
		echo "Error: no suitable docker compose command found" >&2; \
		exit 1; \
	fi \
)
```

flodolo: Again, untested, possibly better logic, no fall back to `docker-composer`. Maybe we could we ask jiri to test it?

`watch.sh` changes based on @eemeli post on Slack.

```diff
diff --git a/Makefile b/Makefile
index 12bb05585..92cc78270 100644
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,15 @@
-DC := $(shell which docker-compose)
-DOCKER := $(shell which docker)
+DOCKER := $(shell command -v docker || command -v pacman || (echo "Error: neither docker nor pacman found" >&2 && exit 1))
+
+DC := $(shell \
+	if [ "$(DOCKER)" = "$$(command -v docker)" ] && docker compose version > /dev/null 2>&1; then \
+		echo "$$(command -v docker) compose"; \
+	elif [ "$(DOCKER)" = "$$(command -v pacman)" ] && command -v docker-pacman > /dev/null 2>&1; then \
+		echo "$$(command -v docker-pacman)"; \
+	else \
+		echo "Error: docker command not found" >&2; \
+		exit 1; \
+	fi \
+)
 
 # *IMPORTANT*
 # Don't use this instance in a production setting. More info at:
diff --git a/bin/watch.sh b/bin/watch.sh
index c739c9515..72b5e7816 100755
--- a/bin/watch.sh
+++ b/bin/watch.sh
@@ -7,7 +7,22 @@
 sigint_handler() { exit 0; }
 trap sigint_handler INT
 
-npx concurrently -n translate,server,pg -c cyan,magenta,green,red \
-  'npm start -w translate' \
-  'docker-compose logs --tail=0 --follow --no-log-prefix server' \
-  'docker-compose logs --tail=0 --follow --no-log-prefix postgresql' \
+# Detect backend
+if command -v docker &>/dev/null && docker compose version &>/dev/null; then
+  DC="docker compose"
+  LOG_SERVER="$DC logs --tail=0 --follow --no-log-prefix server"
+  LOG_PG="$DC logs --tail=0 --follow --no-log-prefix postgresql"
+elif command -v pacman &>/dev/null && command -v docker-pacman &>/dev/null; then
+  DC="docker-pacman"
+  LOG_SERVER="$DC logs --tail=0 --follow server"
+  LOG_PG="$DC logs --tail=0 --follow postgresql"
+else
+  echo "Error: required Docker backend not found or unsupported." >&2
+  exit 1
+fi
+
+# Run processes
+npx concurrently -n translate,server,pg -c cyan,magenta,green \
+  "npm start -w translate" \
+  "$LOG_SERVER" \
+  "$LOG_PG"

```

mathjazz: I'd keep these changes outside this PR, which only ensures compatibility with the recent Docker versions.

Also, Jiri made padman to work with docker commands by using symlinks.

mathjazz:
Sep 2021 is since when `docker compose` works, so I don't think we need to keep supporting `docker-compose`.

Makefile
```diff
@@ -1,4 +1,3 @@
...
 DOCKER := $(shell which docker)
```
eemeli: We're already relying on the environment path, so the `which` call is kinda pointless here.
```suggestion
DOCKER := "docker"
```

mathjazz:
Turns out `docker compose` has been the default since 2021, so no need for backwards compatibility.

eemeli (approved):

PR #3669, opened 2025-06-11
Add Project.date_modified, use it for force-syncing after changes

### LINKED ISSUE #3653: New locale never syncs for webextension project
Added a new locale to Mozilla VPN Extension, locale is stuck on never synced. Had to run a forced sync.

I thought that was fixed in #3523, given it's not a bilingual format?

eemeli:
Fixes #3653

Hardest part here was putting together a minimal test case which fails without the change made to `pontoon/sync/tasks.py`.

This uses the method I originally suggested in https://github.com/mozilla/pontoon/issues/3523#issuecomment-2596860114.

### DIFF
--- a/pontoon/base/models/project.py
+++ b/pontoon/base/models/project.py
@@ -7,7 +7,6 @@ from django.db import models
 from django.db.models import Sum
 from django.db.models.manager import BaseManager
 from django.utils import timezone
-from django.utils.functional import cached_property
 
 from pontoon.base.aggregated_stats import AggregatedStats
 from pontoon.base.models.locale import Locale
@@ -232,17 +231,6 @@ class Project(models.Model, AggregatedStats):
 
     def __str__(self):
         return self.name
-
-    @cached_property
-    def unsynced_locales(self):
-        """
-        Project Locales that haven't been synchronized yet.
-        """
-        return self.locales.exclude(
-            pk__in=Locale.objects.filter(
-                translatedresources__resource__project=self
-            ).values_list("pk", flat=True)
-        )
 
     def serialize(self):
         return {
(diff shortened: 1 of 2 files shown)


eemeli:
@mathjazz I'm not certain, but should we revert the changes made in #3615 once this lands?

pontoon/base/signals.py
```diff
@@ -103,6 +104,48 @@ def create_project_locale_permissions_groups(sender, **kwargs):
...
+    try:
+        prev = sender.objects.get(pk=instance.pk)
+    except sender.DoesNotExist:
+        return
+    for attr in [
+        "configuration_file",
+        "data_source",
+        "disabled",
+        "sync_disabled",
+        "system_project",
```
mathjazz: Not that we change it often (or at all), but I'm curious why is this field on the list?

pontoon/sync/tasks.py
```diff
@@ -27,6 +27,17 @@ def sync_project_task(
         log.error(f"[id={project_pk}] Sync aborted: Project not found.")
         raise
 
+    if not force:
+        try:
+            prev_sync = Sync.objects.filter(project=project).latest("start_time")
+            if project.date_modified > prev_sync.start_time:
```
mathjazz, line 33: So from now on, when a locale is added or enabled for pretranslation in any project, we force sync the project.

That calls for a followup issue to be filed to optimize that.

pontoon/base/signals.py
eemeli: Because I misremembered that it had some impact on sync.
```suggestion
```

pontoon/sync/tasks.py
eemeli: Yes. I'm presuming that that's rare enough that the cost of doing so is marginal.

mathjazz (approved):
Works great!

PR #3673, opened 2025-06-16
Generate target files for XLIFF if they are missing

### LINKED ISSUE #3573: XLIFF project: force-sync doesn't add new locale
Added a new locale on Firefox for iOS (`tg`), even a forced sync didn't add the locale (it's stuck on "Not synced yet").

I had to manually add the file to the repository in order to pick it up.

eemeli:
Fixes #3573

As discussed in the issue, the gettext logic of requiring that each target file is created separately from Pontoon should not be applied to XLIFF files.

I was not able to find any mention of the current behaviour in any of our documentation.

### DIFF
--- a/pontoon/sync/core/entities.py
+++ b/pontoon/sync/core/entities.py
@@ -19,8 +19,6 @@ from pontoon.sync.formats.common import ParseError, VCSTranslation
 
 log = logging.getLogger(__name__)
 
-BILINGUAL_FORMATS = {"po", "xliff"}
-
 
 def sync_entities_from_repo(
     project: Project,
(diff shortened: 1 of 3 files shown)


mathjazz:
Adding a locale still requires a force sync, which isn't great and in line with other file formats. As of https://github.com/mozilla/pontoon/issues/3523, it shouldn't be required.

mathjazz:
Maybe that problem goes away when https://github.com/mozilla/pontoon/pull/3669 lands?

mathjazz:
> Maybe that problem goes away when #3669 lands?

Confirming it does.

mathjazz (approved):

PR #3686, opened 2025-06-28
Make Django Debug Toolbar easier to turn on/off

mathjazz:
Testing performance of https://github.com/mozilla/pontoon/pull/3666 is easier with Django Debug Toolbar than `debug_sql()`. Since Django Debug Toolbar is installed and made available in the local development environment, we should make it easy to turn it on or off.

### DIFF
--- a/docs/admin/deployment.rst
+++ b/docs/admin/deployment.rst
@@ -120,6 +120,9 @@ you create:
    Controls ``DEBUG`` mode for the site. Should be set to `False` in
    production.
 
+``DJANGO_DEBUG_TOOLBAR``
+   Enables Django Debug Toolbar (default: ``False``) if ``DJANGO_DEV`` set to `True`.
+
 ``DJANGO_DEV``
    Signifies whether this is a development server or not. Should be `False` in
    production.
--- a/pontoon/settings/base.py
+++ b/pontoon/settings/base.py
@@ -34,6 +34,8 @@ DEV = os.environ.get("DJANGO_DEV", "False") != "False"
 
 DEBUG = os.environ.get("DJANGO_DEBUG", "False") != "False"
 
+DJANGO_DEBUG_TOOLBAR = os.environ.get("DJANGO_DEBUG_TOOLBAR", "False") != "False"
+
 HEROKU_DEMO = os.environ.get("HEROKU_DEMO", "False") != "False"
 
 LOGOUT_REDIRECT_URL = "/"
--- a/pontoon/settings/dev.py
+++ b/pontoon/settings/dev.py
@@ -42,3 +42,8 @@ TEMPLATES[0]["OPTIONS"]["match_regex"] = re.compile(
 )
 
 GRAPHENE = {"MIDDLEWARE": ["graphene_django.debug.DjangoDebugMiddleware"]}
+
+if base.DJANGO_DEBUG_TOOLBAR:
+    DEBUG_TOOLBAR_CONFIG = {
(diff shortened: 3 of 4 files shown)


docs/admin/deployment.rst
```diff
@@ -120,6 +120,9 @@ you create:
    Controls ``DEBUG`` mode for the site. Should be set to `False` in
    production.
 
+``DJANGO_DEBUG_TOOLBAR``
+   Enables Django Debug Toolbar (default: ``False``) if ``DJANGO_DEV`` set to `True`.
```
mathjazz:
```suggestion
   Enables Django Debug Toolbar (default: ``False``). ``DJANGO_DEV`` must be set to `True`.
```

functionzz:
So I accidentally committed my recommended changes in order to get this working, but I changed the `DJANGO_DEBUG_TOOLBAR` reference in base.py to `DJANGO_DEV`, as there is no `DJANGO_DEBUG_TOOLBAR` set in the env list.

pontoon/settings/base.py
```diff
@@ -34,6 +34,8 @@ def path(*args):
 
 DEBUG = os.environ.get("DJANGO_DEBUG", "False") != "False"
 
+DJANGO_DEBUG_TOOLBAR = os.environ.get("DJANGO_DEBUG_TOOLBAR", "False") != "False"
```
functionzz, line 37:
```suggestion
DJANGO_DEBUG_TOOLBAR = os.environ.get("DJANGO_DEV", "False") != "False"
```

functionzz: Pretty sure this needs to change to `DJANGO_DEV`, because you wanna check if `DJANGO_DEV` is False/True in order to set DJANGO_DEBUG_TOOLBAR.

functionzz (commented):
Another important thing: having the Debug Toolbar ON breaks tests. So I think that's why the Github actions were showing red when I tested it earlier

pontoon/settings/base.py
flodolo: This might be relevant?
https://django-debug-toolbar.readthedocs.io/en/latest/installation.html#disable-the-toolbar-when-running-tests-optional

mathjazz: > Pretty sure this needs to change to `DJANGO_DEV`, because you wanna check if `DJANGO_DEV` is False/True in order to set DJANGO_DEBUG_TOOLBAR.

These two settings should be separate. For example, we must keep the specifics of the DEV environment (e.g. no HTTPS redirects) available also when Django debug toolbar isn't enabled.

mathjazz:
> Another important thing: having the Debug Toolbar ON breaks tests. So I think that's why the Github actions were showing red when I tested it earlier.

Thanks, just fixed that.

Just a quick note for next time: while the PR is still under review, it’s best not to push commits directly to it unless we’ve agreed on that approach. It helps keep the review process clear and avoids unexpected changes while feedback is still being incorporated.

functionzz:
💯  I originally intended to do a code suggestion(as above), but I only recently figured out how to do that :..)

pontoon/settings/base.py
functionzz: So in a typical workflow, would a developer need to manually set the env variable `DJANGO_DEBUG_TOOLBAR=True` inside the docker/config/env file?

mathjazz: Yes. The toolbar will be disabled by default.

functionzz (approved):

PR #3689, opened 2025-06-30
Mostly fix badly normalised Android spaces

### LINKED ISSUE #3672: Polish: Non-breaking spaces converted to normal spaces when serializing Android files
https://github.com/mozilla-l10n/android-l10n/commit/4391c4c35544122fd21f378eca1320231be85005

I don't see anything in Pontoon showing a non-breaking space, but there is match in TM
https://pontoon.mozilla.org/pl/firefox-for-android/all-resources/?search=mozac_feature_addons_permissions_privacy_description&search_identifiers=true&string=206783

If I look at the previous version of the file, it has non-breaking spaces
https://github.com/mozilla-l10n/android-l10n/blob/75a7a5cfe8ee948aa788e673ab09b72a6a0c6caf/mozilla-mobile/android-components/components/feature/addons/src/main/res/values-pl/strings.xml

CC @eemeli @piotrdrag

eemeli:
Mostly fixes #3672

This migration fixes a bit over 1500 Android strings, which is at least three quarters of the strings for which the non-default spaces should have been taken into use after #3611. A little over 500 translations with differences between the TM and the translations table are left unchanged, as they involve other, more complicated whitespace fixes around inline HTML elements.

### DIFF
--- a/pontoon/base/migrations/0081_fix_android_spaces.py
+++ b/pontoon/base/migrations/0081_fix_android_spaces.py
@@ -1,0 +1,29 @@
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("base", "0080_project_date_modified"),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            r"""
+WITH tm AS (
+    SELECT
+        t.id,
+        regexp_replace(trim(BOTH FROM m.target), '\s\s+', ' ', 'g') AS string
+    FROM base_translationmemoryentry m
+    JOIN base_translation t ON t.id = m.translation_id
+    WHERE
+        m.target SIMILAR TO '%[^\S \n]%' AND
+        t.string != m.target
+)
+UPDATE base_translation AS t
+SET string = tm.string
+FROM tm
+WHERE t.id = tm.id AND
+    t.string = regexp_replace(tm.string, '[^\S \n]', ' ', 'g');
+"""
+        ),
+    ]


pontoon/base/migrations/0081_fix_android_spaces.py
```diff
@@ -0,0 +1,29 @@
...
+    WHERE
+        m.target SIMILAR TO '%[^\S \n]%' AND
+        t.string != m.target
+)
+UPDATE base_translation AS t
+SET string = tm.string
+FROM tm
+WHERE t.id = tm.id AND
+    t.string = regexp_replace(tm.string, '[^\S \n]', ' ', 'g');
+"""
```
mathjazz, line 27: Nit: Factoring that out and adding it as a parameter to `migrations.RunSQL` would be a bit more readable and consistent with the rest of the codebase.

pontoon/base/migrations/0081_fix_android_spaces.py
```diff
@@ -0,0 +1,29 @@
...
+            r"""
+WITH tm AS (
+    SELECT
+        t.id,
+        regexp_replace(trim(BOTH FROM m.target), '\s\s+', ' ', 'g') AS string
+    FROM base_translationmemoryentry m
+    JOIN base_translation t ON t.id = m.translation_id
+    WHERE
+        m.target SIMILAR TO '%[^\S \n]%' AND
+        t.string != m.target
```
mathjazz, line 20: Should we explicitly filter only Android format?

eemeli: There aren't any non-Android strings for which these restrictions hold, so doing so wouldn't have any effect.

mathjazz: What about 3rd party installs?

pontoon/base/migrations/0081_fix_android_spaces.py
eemeli: Eh, with the indent level difference I think it's plenty readable for the very few times when this'll need to be read (almost all of which are already now in the past).

pontoon/base/migrations/0081_fix_android_spaces.py
eemeli: We've historically not mutated the `base_translations` table really at all, so such a hit would be highly unlikely. And if some such does occur, it's almost certainly also a bug that this would fix.

PR #3691, opened 2025-06-30
Added stats_data_as_dict and changed stats_data

functionzz:
stats_data now returns a LocaleQuerySet, which is required for our REST API implementation, while stats_data_as_dict retains the old behavior of returning a dictionary, which is used in the /teams page as well as an ajax render.


eemeli (changes requested):
Won't we want to have similar changes applied to `ProjectQuerySet`? If so, that change ought to be included here as well.

Also, if/as we're now calculating the `missing` value, that should be included in the dict, and `translate/src/modules/stats/actions.ts` ought to not recalculate it.

PR #3692, opened 2025-06-30
Only conditionally include Django Debug Toolbar URLs

mathjazz:
Without this fix, deploying Pontoon to prod fails, because `debug_toolbar` isn't installed there.

### DIFF
--- a/pontoon/urls.py
+++ b/pontoon/urls.py
@@ -1,5 +1,3 @@
-import debug_toolbar
-
 from django.contrib import admin
 from django.contrib.auth import logout
 from django.urls import include, path, register_converter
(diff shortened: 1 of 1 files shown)


pontoon/urls.py
```diff
@@ -74,8 +72,17 @@ class LocaleConverter(StringConverter):
...
-    # Django Debug Toolbar
-    path("__debug__/", include(debug_toolbar.urls)),
-    # Team page: Must be at the end
+]
+
+# Conditionally include Django Debug Toolbar
+try:
+    import debug_toolbar
+
+    urlpatterns = [path("__debug__/", include(debug_toolbar.urls))] + urlpatterns
```
flodolo:
```suggestion
    urlpatterns.insert(0 ,path("__debug__/", include(debug_toolbar.urls)))
```

Is this better (and `.append()`  for the teams page)?

mathjazz: I don't think so, but I'll keep debug toolbar URLs appended (rather than prepended).

PR #3696, opened 2025-07-03
Install DEV requirements on stage if DJANGO_DEV is True

mathjazz:
I think there's still value in being able to debug performance on stage with DDT and such, so I'm keeping the ability to install dev requirements through the same variable (`DJANGO_DEV`) that is also used to enable them in settings.

### DIFF
--- a/bin/post_compile
+++ b/bin/post_compile
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Install requirements/dev.txt on staging
+# Install development requirements in stage / production environment if enabled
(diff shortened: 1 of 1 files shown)


bin/post_compile
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 # Install requirements/dev.txt on staging
```
eemeli: Shouldn't this comment be updated?

eemeli (approved):

PR #3697, opened 2025-07-03
Fix stats_data() call in Admin

mathjazz:
Regression from:
https://github.com/mozilla/pontoon/commit/2e7d8db594ecd1d662818b13b58a5fb0e41bc7f4


eemeli (approved):
Missing this was my bad.

flodolo (approved):
Thanks, this works for me locally.

PR #3702, opened 2025-07-08
Added locale string information to individual project endpoints

### LINKED ISSUE #3701: API v2: list localization stats within /api/v2/projects/[slug]
In line with the functionality of the GraphQL endpoint, which exposes locale information within individual project endpoints, we should also try to emulate similar functionality in the REST API.

Old GraphQL query:
```
{
  project(slug: "firefox") {
    localizations {
      locale {
        code
      }
      approvedStrings
      stringsWithWarnings
      missingStrings
      pretranslatedStrings
      totalStrings
    }
  }
}
```

functionzz:
This resolves #3701.

It changes the api/v2/projects/[slug] endpoint to be in line with the GraphQL functionality.
This is what will now be returned for api/v2/projects/firefox/ for example:
```
{
    "approved_strings": 1290933,
    ...
    "tags": [
        ...
    ],
    "project_locale": [
        {
            "locale": "si",
            "total_strings": 14302,
            "approved_strings": 9571,
            "pretranslated_strings": 0,
            "strings_with_errors": 0,
            "strings_with_warnings": 1,
            "missing_strings": 4730,
            "unreviewed_strings": 190
        },
        {
            "locale": "ilo",
            "total_strings": 14302,
            "approved_strings": 1918,
            "pretranslated_strings": 0,
            "strings_with_errors": 0,
            "strings_with_warnings": 3,
            "missing_strings": 12381,
            "unreviewed_strings": 269
        },
        ...
]
}
```

Optimizations for api/v2/projects/[slug] as well as api/v2/[code]/[slug] are included. Some changes were included for the project_locale serializer in order to make the code more extensible.


pontoon/api/serializers.py
```diff
@@ -140,6 +176,61 @@ def get_unreviewed_strings(self, obj):
         return obj.unreviewed
 
 
+class NestedIndividualProjectSerializer(ProjectSerializer):
+    tags = TagSerializer(many=True, read_only=True)
+    project_locale = serializers.SerializerMethodField()
```
mathjazz: Why not `locale`, which was used until now and is used in /projects/?

```suggestion
    locale = serializers.SerializerMethodField()
```

functionzz: Maybe the word `localizations` makes more sense? I believe that's how it looks in the GraphQL

functionzz:
We can definitely include `complete` as a field, I can add it in so we know what it looks like.

pontoon/api/serializers.py
mathjazz: Let's go with that, @flodolo agrees. And let's use the plural form, right.

flodolo:
Tested with [one script](https://github.com/mozilla-l10n/pontoon-scripts/commit/65a22bdd30579157bebb2c28a6b96635a2f1eead), and the output is the same.

flodolo (approved):

flodolo:
Actually, just realized that we lost the list of `locales`. Can we keep that, so you don't have to browse the whole list of `localizations`?

flodolo:
> Hm, not sure that's a good idea. We already have this data available in the `localizations` list.

I would agree if this was structured as a dictionary, with the locale as key. Getting the list of locales is trivial.
But now the locale is a field in an array of records, so you need to traverse the array.

mathjazz:
Should we restructure it then?
```json
  "localizations": {
    "it": {
      "total_strings": 14302,
      "approved_strings": 9571,
      "pretranslated_strings": 0,
      "strings_with_errors": 0,
      "strings_with_warnings": 1,
      "missing_strings": 4730,
      "unreviewed_strings": 190
    },
  }
```

I'd try to avoid redundancy (and extra footprint), as well as simplify the API contract (having both, locales and localizations raises the need to explain the relationship between them).

flodolo:
> Should we restructure it then?

That would work for me. I can get the list of locales with a simple `.keys()`.

pontoon/api/serializers.py
```diff
@@ -115,32 +158,44 @@ class Meta(ProjectSerializer.Meta):
...
-        return obj.warnings
+        slug = self.context["view"].kwargs.get("slug")
+        project = get_object_or_404(
+            Project,
+            slug=slug,
+        )
 
-    def get_total_strings(self, obj):
-        return obj.total
+        project_locales = obj.project_locale.stats_data(project=project)
```
mathjazz: That way you don't need to re-fetch `project` above:

```suggestion
        project_locales = obj.project_locale.stats_data(project=obj)
```

pontoon/api/serializers.py
```diff
@@ -27,25 +55,25 @@ class Meta:
...
-            "plural_rule",
-            "population",
-            "pretranslated_strings",
-            "script",
-            "strings_with_errors",
-            "strings_with_warnings",
             "systran_translate_code",
             "team_description",
             "total_strings",
+            "pretranslated_strings",
```
mathjazz: Let's use the same order as in the dashboards (and `ProjectLocaleSerializer`): `approved_strings` before `approved_strings`.

pontoon/api/serializers.py
```diff
@@ -82,21 +110,36 @@ def get_contact(self, obj):
         return None
 
 
-class ProjectLocaleSerializer(serializers.ModelSerializer):
+class ProjectLocaleSerializer(serializers.ModelSerializer, TranslationStatsMixin):
+    locale = serializers.SerializerMethodField()
```
mathjazz: This should work and then you can drop `get_locale()` below:

```suggestion
    locale = serializers.SlugRelatedField(read_only=True, slug_field="code")
```

pontoon/api/serializers.py
```diff
@@ -82,21 +110,36 @@ def get_contact(self, obj):
...
+    approved_strings = serializers.SerializerMethodField()
+    missing_strings = serializers.SerializerMethodField()
+    pretranslated_strings = serializers.SerializerMethodField()
+    strings_with_errors = serializers.SerializerMethodField()
+    strings_with_warnings = serializers.SerializerMethodField()
+    total_strings = serializers.SerializerMethodField()
+    unreviewed_strings = serializers.SerializerMethodField()
+    complete = serializers.SerializerMethodField()
```
mathjazz: You are declaring these `SerializerMethodFields` in every serializer that uses the mixin.
Better solution: Define the fields once in the mixin.

mathjazz (commented):
A few more improvements and we're good!

pontoon/api/serializers.py
functionzz: Do you mean `approved_strings` before `total_strings`? Can specify the order you are referring to out of these fields:
```
"total_strings",
"approved_strings",
"pretranslated_strings",
"strings_with_errors",
"strings_with_warnings",
"missing_strings",
"unreviewed_strings",
"complete",
```

pontoon/api/serializers.py
functionzz: Yes, currently attempting to make this work

pontoon/api/serializers.py
mathjazz: Sorry. :) `approved_strings` before `pretranslated_strings`:

```
"total_strings",
"approved_strings",
"pretranslated_strings",
"strings_with_warnings",
"strings_with_errors",
"missing_strings",
"unreviewed_strings",
"complete",
```

pontoon/api/serializers.py
```diff
@@ -12,6 +12,41 @@
 )
 
 
+class TranslationStatsMixin(metaclass=serializers.SerializerMetaclass):
```
mathjazz, line 28: This should be safe to remove?

```suggestion
class TranslationStatsMixin:
```

pontoon/api/serializers.py
```diff
@@ -12,6 +12,41 @@
...
+    approved_strings = serializers.SerializerMethodField()
+    complete = serializers.SerializerMethodField()
```
mathjazz: Nit: It would be nice to keep the order of the fields and the methods below in line with the order of fields specified in the serializers.

pontoon/api/serializers.py
```diff
@@ -82,106 +117,65 @@ def get_contact(self, obj):
...
 
 
-class ProjectLocaleSerializer(serializers.ModelSerializer):
+class ProjectLocaleSerializer(TranslationStatsMixin, serializers.ModelSerializer):
+    locale = serializers.SlugRelatedField(read_only=True, slug_field="code")
+
     class Meta:
         model = ProjectLocale
         fields = [
+            "locale",
```
mathjazz, line 123: If you can remove `project`, but it still appear in the response, why are you adding `locale`?

flodolo:
> I noticed performance regressions in the following endpoints:

Regression compared to graphQL or previous version of this patch?

pontoon/api/serializers.py
functionzz: So apparently this is necessary for Mixins to operate in drf serializers: [https://stackoverflow.com/questions/28747487/mixin-common-fields-between-serializers-in-django-rest-framework](url)

pontoon/api/serializers.py
functionzz: So the NestedIndividualProjectSerializer still needs to have `locale` in order to display it as a key. If you remove it, there will be a `KeyError` indeed

mathjazz:
It turns out it's not this PR, it's this:
https://github.com/mozilla/pontoon/commit/9072824b33866cd0faf940bb7a709e9448bb36c4

pontoon/api/views.py
```diff
@@ -159,34 +161,46 @@ def get_queryset(self):
             filters |= Q(system_project=True)
         if filters:
             queryset = queryset | Project.objects.filter(filters).distinct()
-        return queryset.stats_data()
+        return queryset.stats_data().order_by("slug")
```
mathjazz, line 164: We order locales by `name` and then `code`:
https://github.com/mozilla/pontoon/blob/0c24ae10172738bab6eb00324fd52c139a57eed4/pontoon/base/models/locale.py#L313

More consistent project ordering would be by `name`. Alternatively we can also order locales by `code`, which might be more useful for API users.

pontoon/api/serializers.py
mathjazz: Is there something that breaks or doesn't work if you don't specify the `metaclass`?

pontoon/api/serializers.py
```diff
@@ -27,25 +62,25 @@ class Meta:
...
             "team_description",
             "total_strings",
+            "approved_strings",
+            "pretranslated_strings",
+            "strings_with_warnings",
+            "strings_with_errors",
+            "missing_strings",
             "unreviewed_strings",
+            "complete",
         ]
```
mathjazz: Have you thought about folloing the DRY principle here by factoring out `TRANSLATION_STATS_FIELDS`?

```python
TRANSLATION_STATS_FIELDS = [
    "total_strings",
    "approved_strings",
    "pretranslated_strings",
    "strings_with_warnings",
    "strings_with_errors",
    "missing_strings",
    "unreviewed_strings",
    "complete",
]

fields = [
  # fields
] + TRANSLATION_STATS_FIELDS
```

mathjazz (commented):
Just a couple more comments.

Not related to this PR, but since we're refactoring, I think

```python
for translated_term in obj.translations.all():
    if translated_term.locale.code == locale:
        return translated_term.text
```

Can be replaced with:

```python
term = obj.translations.filter(locale__code=locale).first()
return term.text if term else None
```

functionzz:
> Is there something that breaks or doesn't work if you don't specify the metaclass?

Yes. When you remove it, the content inside the mixin has no effect, and everything reverts to unoptimized queries

mathjazz:
> Yes. When you remove it, the content inside the mixin has no effect, and everything reverts to unoptimized queries

Good call, thanks! Please add a comment?

pontoon/api/views.py
functionzz: for the locales endpoint, should I order it by `name` then `code`?

functionzz: then do the same for the projects endpoint, instead having `name`, then `slug`?

mathjazz: I think @flodolo (who uses the API most) would prefer if you keep the projects ordered by `slug` and order locales by `code`.

pontoon/api/serializers.py
```diff
@@ -12,6 +12,54 @@
...
+        return obj.missing
+
+    def get_unreviewed_strings(self, obj):
+        return obj.unreviewed
+
+    def get_complete(self, obj):
+        return obj.is_complete
+
+
+TRANSLATION_STATS_FIELDS = [
```
mathjazz: I'd move that above all classes, right after the imports.

mathjazz (approved):
One last nit and we're good to go!

mathjazz (approved):
Thanks for keeping up with the long review process.

I believe our DRF API is in a much better state now!

mathjazz (approved):

PR #3703, opened 2025-07-08
REST API authentication spec

functionzz:
This PR includes the Authentication specification for the REST API implementation.

### DIFF
--- a/specs/0122-rest-api-authentication.md
+++ b/specs/0122-rest-api-authentication.md
@@ -1,0 +1,59 @@
+- Feature Name: API Authentication
+- Created: 2025-06-20
+- Associated Issue: #N/A
+
+# Summary
+
+Introduce a personal access token (PAT) implementation for accessing certain public-facing REST API endpoints in Pontoon.
+
+# Motivation
+
+With the rollout of our new series of REST API endpoints built with Django REST Framework, there is an increasing need to protect or restrict access to specific endpoints using authentication.
+
+This feature sets a standard for third parties and localizers by providing a more secure method of accessing Pontoon's external REST API. It also lays the groundwork for future enhancements, such as permission levels based on token access scope.
+
+# Feature Explanation
+
+## Features of a Personal Access Token (PAT)
+
+The PAT implementation includes the following features:
+
+- Tokens are created from the user settings page  
+- Tokens have an expiration date  
+- Users can hold up to 10 tokens simultaneously  
+- Tokens can be deleted, but their values cannot be viewed again after the initial creation  
+- Admins can monitor token usage and revoke tokens if necessary  
+- A user’s Pontoon permissions determine what their token can access
+
+## Creating, Viewing, and Deleting a Personal Access Token
+
+A new **Personal Access Tokens** section will appear in the Settings page. A **Generate a New Access Token** button will allow users to create a new token, which is appended to a list (maximum of 10). Users can set an expiration date for each token.
+
+After generation, the token is displayed once for the user to copy. Refreshing the page or navigating away will hide the token permanently.
+
+Users can delete any of their existing tokens at any time.
+
+## Using a Personal Access Token
+
+To access protected external REST API endpoints at `https://pontoon.com/api/v2/`, users must attach a valid token to the `Authorization` header.
+
+For example, to access the `/api/v2/locales/` endpoint:
+
+```bash
+curl \
+  -H "Authorization: Bearer <YOUR-TOKEN>" \
+  https://example.com/api/v2/locales/
+```
+
+
+### Incorrect Usage
+
+If a user tries to access a protected endpoint without a valid token or without sufficient permissions, a `403 Unauthorized` response is returned.
+
+Multiple consecutive failed attempts may result in the user's IP address being temporarily blocked.
+
+## Admin Usage
+
+Admins will have access to a dashboard to view token usage and revoke individual tokens as needed.
+
+This dashboard includes functionality for composing messages with a subject and body, sending in-app notifications and/or emails, and targeting recipients based on common criteria.


specs/0122-rest-api-authentication.md
```diff
@@ -0,0 +1,59 @@
+- Feature Name: API Authentication
+- Created: 2025-06-20
+- Associated Issue: #N/A
```
mathjazz:
```suggestion
- Associated Issue: #2502
```

specs/0122-rest-api-authentication.md
```diff
@@ -0,0 +1,59 @@
...
+
+# Summary
+
+Introduce a personal access token (PAT) implementation for accessing certain public-facing REST API endpoints in Pontoon.
+
+# Motivation
+
+With the rollout of our new series of REST API endpoints built with Django REST Framework, there is an increasing need to protect or restrict access to specific endpoints using authentication.
```
mathjazz: This feature is actually independent of the API implementation:
```suggestion
There is an increasing need to protect or restrict access to specific endpoints using authentication.
```

specs/0122-rest-api-authentication.md
```diff
@@ -0,0 +1,59 @@
...
+  -H "Authorization: Bearer <YOUR-TOKEN>" \
+  https://example.com/api/v2/locales/
+```
+
+
+### Incorrect Usage
+
+If a user tries to access a protected endpoint without a valid token or without sufficient permissions, a `403 Unauthorized` response is returned.
+
+Multiple consecutive failed attempts may result in the user's IP address being temporarily blocked.
```
mathjazz: _May_ is not the best word to be used in the spec. We either need to implement the blocking mechanism or file an issue to do that at a later stage.

specs/0122-rest-api-authentication.md
```diff
@@ -0,0 +1,59 @@
...
+- Admins can monitor token usage and revoke tokens if necessary  
+- A user’s Pontoon permissions determine what their token can access
+
+## Creating, Viewing, and Deleting a Personal Access Token
+
+A new **Personal Access Tokens** section will appear in the Settings page. A **Generate a New Access Token** button will allow users to create a new token, which is appended to a list (maximum of 10). Users can set an expiration date for each token.
+
+After generation, the token is displayed once for the user to copy. Refreshing the page or navigating away will hide the token permanently.
+
+Users can delete any of their existing tokens at any time.
```
mathjazz, line 34: How is the token displayed after it's been hidden? We still need a way to identify it in order to be able to delete it.

specs/0122-rest-api-authentication.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Using a Personal Access Token
+
+To access protected external REST API endpoints at `https://pontoon.com/api/v2/`, users must attach a valid token to the `Authorization` header.
+
+For example, to access the `/api/v2/locales/` endpoint:
```
mathjazz: That's not a good example, because it's a public endpoint. Since we don't have any endpoints that require authentication, maybe use something generic, e.g. `/api/v2/resource/`.

specs/0122-rest-api-authentication.md
```diff
@@ -0,0 +1,59 @@
...
+
+If a user tries to access a protected endpoint without a valid token or without sufficient permissions, a `403 Unauthorized` response is returned.
+
+Multiple consecutive failed attempts may result in the user's IP address being temporarily blocked.
+
+## Admin Usage
+
+Admins will have access to a dashboard to view token usage and revoke individual tokens as needed.
+
+This dashboard includes functionality for composing messages with a subject and body, sending in-app notifications and/or emails, and targeting recipients based on common criteria.
```
mathjazz: That almost sounds like an overkill? Should we just have a table in the Django admin with:
* token ID
* user
* usage count
* button to revoke it

And send notification to the user once the token has been revoked?

mathjazz (commented):
Nice work! Left some high-level comments.

Any reason why is this a Draft?

specs/0122-rest-api-authentication.md
```diff
@@ -0,0 +1,59 @@
+- Feature Name: API Authentication
+- Created: 2025-06-20
```
flodolo: Odd date given the timing of this PR?

specs/0122-rest-api-authentication.md
functionzz: makes sense

specs/0122-rest-api-authentication.md
functionzz: So in Github you have the option to name the token something, but other than usage and expiry date, no other info is deleted

image

functionzz: Should we do something like this?(Have the user name their token)

mathjazz: That makes sense. And then we can show the name, creation date and expiration date when listing tokens.

specs/0122-rest-api-authentication.md
```diff
@@ -0,0 +1,59 @@
...
+
+This feature sets a standard for third parties and localizers by providing a more secure method of accessing Pontoon's external REST API. It also lays the groundwork for future enhancements, such as permission levels based on token access scope.
+
+# Feature Explanation
+
+## Features of a Personal Access Token (PAT)
+
+The PAT implementation includes the following features:
+
+- Tokens are created from the user settings page
```
flodolo:
```suggestion
- Tokens are created from the user settings page
```

Nitpicking (also in the other list items)

specs/0122-rest-api-authentication.md
```diff
@@ -0,0 +1,59 @@
...
+
+### Incorrect Usage
+
+If a user tries to access a protected endpoint without a valid token or without sufficient permissions, a `403 Unauthorized` response is returned.
+
+Multiple consecutive failed attempts will result in the user's IP address being temporarily blocked.
+
+## Admin Usage
+
+Admins have access to the [`Django Admin Dashboard`](https://docs.djangoproject.com/en/5.2/ref/contrib/admin/), where they can view token usage, usernames, and revoke tokens as needed.
```
flodolo: How does revocation look like? It is removed, is it flagged as revoked? I'm guessing the latter, because we should also exclude that the same token exists twice over time.

flodolo (commented):
What's the token length and what's the algorithm to generate it?

specs/0122-rest-api-authentication.md
functionzz: Yeah, I created the file a long time ago :)

specs/0122-rest-api-authentication.md
functionzz: it is revoked, ideally we have a flag in the token model which specifies whether it is revoked or not.

specs/0122-rest-api-authentication.md
```diff
@@ -0,0 +1,59 @@
...
+## Features of a Personal Access Token (PAT)
+
+The PAT implementation includes the following features:
+
+- Tokens are created from the user settings page
+- Tokens have an expiration date  
+- Users can hold up to 10 tokens simultaneously  
+- Tokens can be deleted, but their values cannot be viewed again after the initial creation  
+- Admins can monitor token usage and revoke tokens if necessary  
+- A user’s Pontoon permissions determine what their token can access
```
eemeli, line 26: Does this mean that actions taken with the token will be indistinguishable from the user's other actions? If so, this should be clarified.

specs/0122-rest-api-authentication.md
```diff
@@ -0,0 +1,59 @@
+- Feature Name: API Authentication
+- Created: 2025-07-14
+- Associated Issue: #2502
```
eemeli:
```suggestion
- Associated Issue: [#2502](https://github.com/mozilla/pontoon/issues/2502)
```

specs/0122-rest-api-authentication.md
```diff
@@ -0,0 +1,59 @@
...
+This feature sets a standard for third parties and localizers by providing a more secure method of accessing Pontoon's external REST API. It also lays the groundwork for future enhancements, such as permission levels based on token access scope.
+
+# Feature Explanation
+
+## Features of a Personal Access Token (PAT)
+
+The PAT implementation includes the following features:
+
+- Tokens are created from the user settings page
+- Tokens have an expiration date
```
eemeli, line 22: Have, or may have?

On line 30, having one is expressed as a possibility.

specs/0122-rest-api-authentication.md
```diff
@@ -0,0 +1,59 @@
...
+
+### Incorrect Usage
+
+If a user tries to access a protected endpoint without a valid token or without sufficient permissions, a `403 Unauthorized` response is returned.
+
+Multiple consecutive failed attempts will result in the user's IP address being temporarily blocked.
+
+## Admin Usage
+
+Admins have access to the [`Django Admin Dashboard`](https://docs.djangoproject.com/en/5.2/ref/contrib/admin/), where they can view token usage, usernames, and revoke tokens as needed.
```
eemeli: Should mention that admins ought not have the ability to see the secret part of the token, only its hash.

eemeli (commented):
This is clearly being reviewed, so ought not be marked as "Draft". Overall, looks pretty good.

specs/0122-rest-api-authentication.md
functionzz: We should probably make it a requirement to have an expiration date.

specs/0122-rest-api-authentication.md
functionzz: can you clarify what you mean by that? I'm not sure I understand...

eemeli: I mean that the actions taken with the token will not be tracked as separate from the user's other actions; it won't be handled (at least for now) as as separate Pontoon user.

functionzz: that is correct.

specs/0122-rest-api-authentication.md
```diff
@@ -0,0 +1,60 @@
...
+
+The PAT implementation includes the following features:
+
+- Tokens are created from the user settings page
+- Tokens have an expiration date  
+- Users can hold up to 10 tokens simultaneously  
+- Tokens can be deleted, but their values cannot be viewed again after the initial creation  
+- Admins can monitor token usage and revoke tokens if necessary  
+- A user’s Pontoon permissions determine what their token can access
+- Actions performed with a user's PAT correspond directly to their account.
```
mathjazz: Could you clarify what you mean by that?

specs/0122-rest-api-authentication.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Creating, Viewing, and Deleting a Personal Access Token
+
+A new **Personal Access Tokens** section will appear in the Settings page. A **Generate a New Access Token** button will allow users to create a new token, which is appended to a list (maximum of 10). Users must select an expiration date and optionally include a note to each token to help identify its purpose.
```
mathjazz: Nit: 
```suggestion
A new **Personal Access Tokens** section will appear in the Settings page. A **Generate a New Access Token** button will allow users to create a new token, which is appended to a list (maximum of 10). Users must select an expiration date and optionally add a note to each token to help identify its purpose.
```

specs/0122-rest-api-authentication.md
```diff
@@ -0,0 +1,60 @@
...
+
+For example, to access the `/api/v2/resource/` endpoint:
+
+```bash
+curl \
+  -H "Authorization: Bearer <YOUR-TOKEN>" \
+  https://example.com/api/v2/resource/
+```
+
+
```
mathjazz:
```suggestion

```

specs/0122-rest-api-authentication.md
```diff
@@ -0,0 +1,60 @@
...
+### Incorrect Usage
+
+If a user tries to access a protected endpoint without a valid token or without sufficient permissions, a `403 Unauthorized` response is returned.
+
+Multiple consecutive failed attempts will result in the user's IP address being temporarily blocked.
+
+## Admin Usage
+
+Admins have access to the [`Django Admin Dashboard`](https://docs.djangoproject.com/en/5.2/ref/contrib/admin/), where they can view token usage, usernames, and revoke tokens as needed. For security reasons, only the hashed version of each token is visible—admins cannot view the original token value.
```
mathjazz:
```suggestion
Admins have access to the [`Django Admin`](https://docs.djangoproject.com/en/5.2/ref/contrib/admin/), where they can view token usage, usernames, and revoke tokens as needed. For security reasons, only the hashed version of each token is visible—admins cannot view the original token value.
```

eemeli (approved):
Looks good, pending resolution of Matjaž's comments above.

specs/0122-rest-api-authentication.md
functionzz: "Actions performed with a user's PAT correspond directly to their account." meaning each action a user takes is not tracked separately from their other actions in Pontoon. Personally, I'm not sure if adding this note is actually necessary.

mathjazz: As in Actions performed with a user's PAT are logged (and we know who's the author through the ForeignKey)? 

If yes, I don't know if we need this in general. I'd rather remove it and implement it for each endpoint the requires authentication as needed.

specs/0122-rest-api-authentication.md
```diff
@@ -0,0 +1,58 @@
...
+To access protected external REST API endpoints at `https://pontoon.com/api/v2/`, users must attach a valid token to the `Authorization` header.
+
+For example, to access the `/api/v2/resource/` endpoint:
+
+```bash
+curl \
+  -H "Authorization: Bearer <YOUR-TOKEN>" \
+  https://example.com/api/v2/resource/
+```
+### Incorrect Usage
```
mathjazz, line 48: Previously we had 2 empty lines, now we have 0. The right number is 1. :)

specs/0122-rest-api-authentication.md
functionzz: OK, yeah wasn't too sure, I'll just remove it.

Issue #3676, opened 2025-06-18
Deprecate GraphQL


functionzz:
As a precursor to migrating the public-facing API endpoints from GraphQL to DRF, we also need to remove GraphQL from the codebase.

flodolo:
> As a precursor to migrating the public-facing API endpoints from GraphQL to DRF, we also need to remove GraphQL from the codebase.

This doesn't sound right: we need to deprecate first, then remove GraphQL code (after the deprecation period), not the other way around?

functionzz:
right.

PR #3704, opened 2025-07-08
Api/V2/: User-actions Endpoint

### LINKED ISSUE #3680: Migrate get_user_actions to DRF
In line with our adoption of the Django REST Framework, we should look into migrating the get_user_actions function in the api/urls.py file, where the current DRF implementation resides.

functionzz:
This PR resolves #3680.

It creates a new endpoint at `/api/v2/user-actions/<str:date>/project/<slug:slug>/`, which lists out user actions on the specified day related to a specified project.

It mirrors the behavior of `/api/v1/user-actions/<str:date>/project/<slug:slug>/`. It requires the user to be signed in in order to be able to access the data.

Future action is needed to clean up the old endpoint `/api/v2/user-actions/<str:date>/project/<slug:slug>/`.

### DIFF
--- a/pontoon/api/urls.py
+++ b/pontoon/api/urls.py
@@ -51,6 +51,12 @@ api_v2_patterns = [
         views.ProjectLocaleIndividualView.as_view(),
         name="project-locale-individual",
     ),
+    path(
+        # User Actions
+        "user-actions/<str:date>/project/<slug:slug>/",
+        views.UserActionsView.as_view(),
+        name="user-actions",
+    ),
 ]
 urlpatterns = [
     # GraphQL endpoint. Serves the GraphiQL IDE if accessed with Accept: text/html.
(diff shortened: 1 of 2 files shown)


mathjazz:
> Future action is needed to clean up the old endpoint `/api/v2/user-actions/<str:date>/project/<slug:slug>/`.

We should file a bug for that or add a comment to https://github.com/mozilla/pontoon/issues/3676 to also include this endpoint.

mathjazz:
Please note that the new endpoint should also require authentication.

functionzz:
In this case, should we wait until after API authentication is implemented to merge this? We would have to rewrite the auth for this anyways right?

mathjazz:
No need to, but we should at least check that the user is authenticated in Pontoon in order to make the request.

mathjazz (approved):
\o/

PR #3710, opened 2025-07-09
Refactor pretranslation to support gettext plurals

### LINKED ISSUE #3694: Re-enable pretranslation for multi-variant gettext messages
In #3635, the representation of gettext plural messages changes to use a single message with MF2 syntax. This needs to be accounted for also for pretranslation, which currently special-cases Fluent messages with multiple variants and/or attributes.

The cleanest way of applying the change would be to also refactor the Fluent processing to use the `moz.l10n` message representation rather than the Fluent AST.

eemeli:
Fixes #3694

As it doesn't really make sense to maintain two parallel implementations for different formats, the preceding Fluent AST transformer is replaced with a moz.l10n based approach. This also fixes a few corner-case bugs that I noticed during the work; will comment inline on those, and other differences.


pontoon/pretranslation/pretranslate.py
```diff
@@ -1,130 +1,257 @@
...
 
-    :arg Entity entity: the Entity object
-    :arg Locale locale: the Locale object
-    :arg boolean preserve_placeables
-
-    :returns: None, or a tuple consisting of:
+    :returns: A tuple consisting of:
         - a pretranslation of the entity
-        - a user (representing TM or GT service)
+        - a pretranslation service identifier, either "gt" or "tm"
```
eemeli, line 43: Previously we were re-fetching the user data for every single pretranslation. We only need to get that info once, in the calling code.

pontoon/pretranslation/pretranslate.py
```diff
@@ -1,130 +1,257 @@
...
+
+    def pattern(self, pattern: Pattern) -> Pattern:
+        # First try to get a 100% match from Translation Memory
+        tm_source = (
+            "".join(
+                el.value
+                if isinstance(el, FTL.TextElement)
+                else serialize_expression(el)
+                for el in fluent_astify_message(PatternMessage(pattern)).elements
+            )
```
eemeli, line 144: This smells like a use case for `moz.l10n.message.serialize_message` to support Fluent pattern serialization, which it doesn't yet.

pontoon/pretranslation/pretranslate.py
```diff
@@ -1,130 +1,257 @@
...
-    # Fetch from Google Translate
-    elif locale.google_translate_code:
-        gt_response = get_google_translate_data(
-            text=source, locale=locale, preserve_placeables=preserve_placeables
+        if not tm_source or tm_source.isspace():
+            return pattern
+        tm_q100 = list(
+            TranslationMemoryEntry.objects.filter(
+                locale=self.locale, source=tm_source
+            ).values_list("target", flat=True)
```
eemeli, line 153: As we're only interested in 100% matches, this is much more straightforward than looking for strings that have an edit distance of 0.

pontoon/pretranslation/pretranslate.py
```diff
@@ -1,130 +1,257 @@
...
         )
-        if gt_response["status"]:
-            return gt_response["translation"], "gt"
+        if tm_q100:
+            tm_best = max(set(tm_q100), key=tm_q100.count)
+            self.services.append("tm")
+            if self.format == Format.fluent:
+                te = fluent_parse_entry(f"key = {tm_best}\n")
+                assert isinstance(te.value, PatternMessage)
+                return te.value.pattern
```
eemeli, line 161: This feels a bit hacky as well, and probably should be supported by moz.l10n.

pontoon/pretranslation/tasks.py
```diff
@@ -126,34 +128,37 @@ def pretranslate(project: Project, paths: set[str] | None):
...
-                    pretranslated=True,
-                    active=True,
-                )
+                try:
+                    pretranslation = get_pretranslation(
+                        entity, locale, preserve_placeables=True
+                    )
+                except ValueError as e:
+                    log.info(f"Pretranslation error: {e}")
+                    continue
```
eemeli, line 137: Previously, we weren't accounting for `get_translations` failing when it's called with `preserve_placeables=True`.

pontoon/pretranslation/tests/test_pretranslate.py
```diff
@@ -108,32 +106,31 @@ def test_get_pretranslations_fluent_simple(
...
     }
 
-    fluent_string = 'hello-world = { "" }'
+    fluent_string = 'hello-world = { "" }\n'
     fluent_entity = EntityFactory(resource=fluent_resource, string=fluent_string)
-    pretranslated_string = "hello-world = gt_translation\n"
 
     response = get_pretranslation(fluent_entity, google_translate_locale)
-    assert response == (pretranslated_string, gt_user)
+    assert response == (fluent_string, "tm")
```
eemeli, line 127: The source `{ "" }` is now correctly parsed as not containing any literal text, and therefore isn't passed to Google Translate.

pontoon/pretranslation/tests/test_pretranslate.py
```diff
@@ -154,21 +151,21 @@ def test_get_pretranslations_fluent_accesskeys_no_attribute_source(
         """
         title =
             .foo = gt_translation
-            .accesskey = gt_translation
+            .accesskey = B
```
eemeli, line 154: Accesskeys should never go through machine translation, and so this is left untouched as the message does not have a value or attribute from which we can look up what this is an accesskey of.

pontoon/pretranslation/tests/test_pretranslate.py
```diff
@@ -377,33 +374,33 @@ def test_get_pretranslations_fluent_accesskeys_prefixed_label_attribute(
...
-        "translation": "{ brand } gt_translation",
+        "translation": "{ $0 } gt_translation",
     }
 
     expected = dedent(
         """
-        title = { brand } gt_translation
-            .label = { brand } gt_translation
+        title = {$0} gt_translation
+            .label = { $brand } gt_translation
```
eemeli, line 408: The `title` value here is hacky, but demonstrates that a Google Translate response that includes an unexpected `{$0}` does not cause a failure, just odd output. For the `.label` it's expected, and therefore replaced with the correct placeholder.

pontoon/pretranslation/tests/test_pretranslate.py
```diff
@@ -622,8 +617,8 @@ def test_get_pretranslations_fluent_multiline(
     # 100% TM match exists
     tm = TranslationMemoryFactory.create(
         entity=entity_b,
-        source="Multi Line Message",
-        target="TM: Multi Line Message",
+        source="Multi\nLine\nMessage",
+        target="TM: Multi\nLine\nMessage",
```
eemeli, line 626: In actual use, we do have multiline values in the translation memory, and that's why we need separate `tm_source` and `gt_source` values in `Pretranslate.pattern()`.

Previously, values with newlines were never matched to translation memory entries.

pontoon/pretranslation/tests/test_pretranslate.py
```diff
@@ -735,38 +730,77 @@ def test_get_pretranslations_fluent_complex(
...
     response = get_pretranslation(fluent_entity, google_translate_locale)
-    assert response == (pretranslated_string, tm_user)
+    assert response == (pretranslated_string, "tm")
+
+
+@patch("pontoon.pretranslation.pretranslate.get_google_translate_data")
+@pytest.mark.django_db
+def test_get_pretranslations_fluent_placeholders(
+    gt_mock, fluent_resource, google_translate_locale
+):
```
eemeli, line 745: This test is the only one of the `test_transformer.py` tests that doesn't replicate other tests already in this file.

pontoon/pretranslation/tests/test_pretranslate.py
```diff
@@ -735,38 +730,77 @@ def test_get_pretranslations_fluent_complex(
...
     # - Move any text outside selectors into selector variants.
     # - Keep leading/trailing whitespace outside selector variants.
     TranslationMemoryFactory.create(
         entity=entity_a,
-        source="{ $key_count } key",
-        target="TM: { $key_count } key",
+        source="{ $key_count } key and { $lock_count } lock",
+        target="TM: { $key_count } key and { $lock_count } lock",
         locale=google_translate_locale,
     )
```
eemeli, line 795: The previous pretranslation behaviour did not match what's actually in the translation memory; switching to complete patterns will allow some of them to actually match.

pontoon/pretranslation/tests/test_pretranslate.py
mathjazz: > Accesskeys should never go through machine translation

You mean in this particular case (when the message has no value or required attributes), right?

pontoon/pretranslation/tests/test_pretranslate.py
```diff
@@ -377,33 +374,33 @@ def test_get_pretranslations_fluent_accesskeys_prefixed_label_attribute(
...
+    assert response == (pretranslated_string, "gt")
 
 
 @patch("pontoon.pretranslation.pretranslate.get_google_translate_data")
 @pytest.mark.django_db
 def test_get_pretranslations_fluent_accesskeys_ignore_placeables(
-    gt_mock, fluent_resource, google_translate_locale, gt_user
+    gt_mock, fluent_resource, google_translate_locale
 ):
     # Ignore placeables whene generating accesskeys
```
mathjazz:
```suggestion
    # Ignore placeables when generating accesskeys
```

pontoon/pretranslation/tests/test_pretranslate.py
mathjazz: Add that as a test comment, please?

pontoon/pretranslation/tests/test_pretranslate.py
mathjazz: Please add a short comment like we do for other tests.

pontoon/pretranslation/pretranslate.py
mathjazz: File a boog?

pontoon/pretranslation/pretranslate.py
mathjazz: Should file a bug, too.

eemeli: https://github.com/mozilla/moz-l10n/issues/92

pontoon/pretranslation/pretranslate.py
eemeli: https://github.com/mozilla/moz-l10n/issues/92

pontoon/pretranslation/tests/test_pretranslate.py
eemeli: No; in general. The accesskey is effectively a derived property of the label it's for, and machine-translating a single letter will not take that into account at all.

pontoon/pretranslation/tests/test_pretranslate.py
```diff
@@ -377,33 +374,33 @@ def test_get_pretranslations_fluent_accesskeys_prefixed_label_attribute(
...
     )
     fluent_entity = EntityFactory(resource=fluent_resource, string=input_string)
 
     gt_mock.return_value = {
         "status": True,
-        "translation": "{ brand } gt_translation",
+        "translation": "{ $0 } gt_translation",
     }
 
     expected = dedent(
```
eemeli, line 405:
```suggestion
    # The `title` value here is hacky, but demonstrates that
    # a Google Translate response that includes an unexpected `{$0}`
    # does not cause a failure, just odd output.
    # For the `.label` it's expected,
    # and therefore replaced with the correct placeholder.
    expected = dedent(
```

pontoon/pretranslation/tests/test_pretranslate.py
eemeli:
```suggestion
def test_get_pretranslations_fluent_placeholders(
    gt_mock, fluent_resource, google_translate_locale
):
    # Tests for various Fluent placeholders.
    # - Placeholders are replaced by {$0}, {$1}, ... when passing to Google Translate.
    # - Original placeholders are retained in final result.
```

pontoon/pretranslation/tests/test_pretranslate.py
mathjazz: Ah, sorry - pretranslation / machine translation confusion on my end.

mathjazz (approved):
Thanks - tested again on stage and it looks fine.

PR #3711, opened 2025-07-09
Drop .inc format support

### LINKED ISSUE #3588: Support for .inc should be dropped
Resolving SeaMonkey [bug 1951305](https://bugzilla.mozilla.org/show_bug.cgi?id=1951305) will hopefully be able to remove the one final remaining `.inc` file that's used for localization on pontoon.mozilla.org.

As I hope this file format has no external users (it relies on a [custom preprocessor](https://searchfox.org/mozilla-central/source/python/mozbuild/mozbuild/preprocessor.py)), this should allow us to remove support for the format entirely from Pontoon.

eemeli:
Fixes #3588

As we [no longer](https://bugzilla.mozilla.org/show_bug.cgi?id=1951305#c5) have any .inc files, we no longer need to support them. 🎉 

This also removes the `MOZ_LANGPACK_CONTRIBUTORS` hack.


mathjazz (approved):
LGTM

PR #3713, opened 2025-07-09
Remove `Entity.word_count` as unused

eemeli:
Tests for this PR will likely fail until #3711 is merged and this is rebased on that. Even though there are no file conflicts, the data migration here is indexed on top of the migration in #3711.

The `Entity.word_count` field is unused, so we should drop it, along with the patches that kept it updated.


flodolo:
While this is not displayed, I think this is used to extract info when PM need it?

mathjazz:
We have a couple of scripts that rely on `Entity.word_count`:
https://github.com/mozilla-l10n/pontoon-scripts/tree/main/stats

mathjazz (approved):
I think the proper solution here would be to refactor word count to also support Fluent / MF2 strings. But no strong opinions.

eemeli:
> I think the proper solution here would be to refactor word count to also support Fluent / MF2 strings. But no strong opinions.

@mathjazz Could you file a moz.l10n issue asking for that? It seems like potentially the right place for the functionality, unless we want to make postgres do the math.

PR #3716, opened 2025-07-11
Add Section model, move `group_comment` and `resource_comment` out of Entity

eemeli:
See #3538 for context.

This brings Pontoon's data model a significant step closer to the moz.l10n data model, by inserting Section between Entity and Resource. For now, the only observable part of the change is moving `Entity.group_comment` values to `Section.comment`, and `Entity.resource_comment` to `Resource.comment`, but under the hood the moz.l10n entry/section/resource metadata is also sync'd into the database -- a next step will start using that in serialization.

Getting the section data required some format parsing refactoring, as well as using moz.l10n for Fluent parsing. Other next steps unlocked by this PR will allow us to simplify and eventually remove the per-format customizations.

The migration needed to be split into two files to make it fully reversible.

### DIFF
--- a/pontoon/sync/formats/ftl.py
+++ b/pontoon/sync/formats/ftl.py
@@ -16,9 +16,7 @@ serializer = FluentSerializer()
 
 def _string(entry: Entry[Message]):
     # Do not store comments in the string field
-    entry.comment = ""
-    entry.meta = []
-    return serializer.serialize_entry(fluent_astify_entry(entry))
+    return serializer.serialize_entry(fluent_astify_entry(entry, lambda _: ""))
 
 
 def ftl_as_translation(entry: Entry[Message]):


pontoon/sync/formats/ftl.py
```diff
@@ -1,56 +1,39 @@
...
-
-        elif isinstance(obj, ast.GroupComment):
-            group_comment = [obj.content] if obj.content else []
-
-        elif isinstance(obj, ast.ResourceComment) and obj.content:
-            resource_comment.append(obj.content)
-
-    return translations
+def _string(entry: Entry[Message]):
+    # Do not store comments in the string column
```
mathjazz: Nit (use a more Django lingo):

```suggestion
    # Do not store comments in the string field
```

pontoon/base/models/entity.py
```diff
@@ -598,6 +611,7 @@ def map_entities(
             try:
                 readonly = entity.resource.project.projectlocale[0].readonly
             except IndexError:
+                print(entity, "IndexError")
```
mathjazz: Should be removed (or use log instead).

pontoon/base/models/entity.py
```diff
@@ -632,8 +646,8 @@ def map_entities(
                     "project": entity.resource.project.serialize(),
                     "format": entity.resource.format,
                     "comment": entity.comment,
-                    "group_comment": entity.group_comment,
-                    "resource_comment": entity.resource_comment,
+                    "group_comment": entity.section_comment or "",
```
mathjazz, line 648: Why do we annotate section comments, rather than prefetch sections? It's surprising we use different approach for section and resource comments.

pontoon/base/migrations/0087_move_comments_to_section_and_resource.py
```diff
@@ -0,0 +1,69 @@
...
+    sections = {}
+    for entity in es:
+        s_key = (entity.resource_id, entity.group_comment)
+        section = sections.get(s_key, None)
+        if section is None:
+            section = Section(
+                resource_id=entity.resource_id, key=[], comment=entity.group_comment
+            )
+            sections[s_key] = section
+        entity.section = section
```
mathjazz: Have you tested if that works? At this point `section.pk` hasn't been populated, but the code might work in the migration context.

pontoon/base/migrations/0087_move_comments_to_section_and_resource.py
```diff
@@ -0,0 +1,69 @@
...
+
+    er = (
+        Entity.objects.exclude(resource_comment="")
+        .order_by("resource", "-date_created")
+        .distinct("resource")
+        .select_related("resource")
+    )
+    for entity in er:
+        entity.resource.comment = entity.resource_comment
+    Resource.objects.bulk_update((e.resource for e in er), ["comment"])
```
mathjazz: Using a list feels safer than a generator:

```suggestion
    Resource.objects.bulk_update(list({e.resource for e in er}), ["comment"])
```

But maybe it's just a personal preference. :)

pontoon/sync/tests/formats/test_xliff.py
```diff
@@ -35,58 +39,44 @@ def test_xliff(self):
...
-            # missing translation
-            assert t3.comments == []
-            assert t3.key == "filename\x04Missing Translation Key"
-            assert t3.string is None
-            assert t3.source_string == "Missing Translation"
-            assert t3.order == 3
+        assert t1.key == "filename\x04Multiple Comments Key"
+        assert t1.string == "Translated Multiple Comments"
 
-    def test_invalid_xliff(self):
```
mathjazz, line 73: Does removing this test mean it's covered by `moz.l10n`?

mathjazz (commented):
Great work. See comments inline.

Should we test on stage?

pontoon/base/migrations/0087_move_comments_to_section_and_resource.py
eemeli: Yes, I've tested this. This works because the `Section.objects.bulk_create()` will set the `pk` value on the section objects before it's needed by the `Entity.objects.bulk_update()`.

pontoon/base/migrations/0087_move_comments_to_section_and_resource.py
eemeli: Django will do that internally, so there's no need to do it in our code: https://github.com/django/django/blob/a698dc223be0e245c8e9cf347defc1892ae5e3ea/django/db/models/query.py#L851

pontoon/base/models/entity.py
eemeli: Oops, didn't mean to leave this in.
```suggestion
```

pontoon/base/models/entity.py
eemeli: Because we only need the comment field from sections, and I didn't want to add yet another query to this API call -- we're already at 7. Resource kinda makes sense as a prefetch, given that we do use multiple values from it.

pontoon/sync/tests/formats/test_xliff.py
eemeli: Yes: https://github.com/mozilla/moz-l10n/blob/bd9450721815ae906ea6e5aa420abe733074eb1a/python/tests/formats/test_xliff1.py#L141-L143

eemeli:
> Should we test on stage?

Yes.

pontoon/base/models/entity.py
mathjazz: Fair.

mathjazz:
> > Should we test on stage?
> 
> Yes.

This PR is now on stage.

pontoon/sync/formats/ftl.py
```diff
@@ -1,56 +1,37 @@
...
-
-        elif isinstance(obj, ast.GroupComment):
-            group_comment = [obj.content] if obj.content else []
-
-        elif isinstance(obj, ast.ResourceComment) and obj.content:
-            resource_comment.append(obj.content)
-
-    return translations
+def _string(entry: Entry[Message]):
+    # Do not store comments in the string field
```
mathjazz, line 18: This comment now seems obsolete.

mathjazz (approved):
Tested a bit more and it seems to work.

pontoon/sync/formats/ftl.py
eemeli: It's still valid, as the second argument is blocking the serialisation of comments.

mathjazz: Yeah, I just thought it wasn't necessary. But maybe it's even more necessary now that this isn't explicit.

Issue #3656, opened 2025-05-27
Project and ProjectLocale Insights pages don't have data since January


mathjazz:
![Image](https://github.com/user-attachments/assets/f5d00742-2c03-4b65-8601-7974860f379d)

mathjazz:
The missing data (e.g. `ProjectLocaleInsightsSnapshot.human_translations`) was last collected on January 23. Later that day we deployed [stats refactor](https://github.com/mozilla/pontoon/commit/948e6cf02b1e58e60ddd599c8e40c27a3f432fdc#diff-0ed2f4a5a57cb48d7b7cf0bb58deeaf1a6a28d8b6206c78e86cdb0fe61c7db1e), which also touched the `pontoon.insights` module and since then, `ProjectLocaleInsightsSnapshot` data is partially missing.

PR #3720, opened 2025-07-15
Replace custom `group_dict_by()` with `itertools.groupby()`

eemeli:
Noticed this while trawling through the code looking for the string "key": We have a custom function that's only used in one related set of functions that duplicates functionality that's also provided by `itertools.groupby()`, so let's get rid of it.

### DIFF
--- a/pontoon/base/utils.py
+++ b/pontoon/base/utils.py
@@ -2,7 +2,6 @@ import functools
 import re
 import time
 
-from collections import defaultdict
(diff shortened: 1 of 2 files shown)


mathjazz (commented):
Could you add tests here? We risk introducing hard-to-discover issues like #3656 wihout gaing that much.

Also, we are likely to drop this new code as part of #2938.

PR #3721, opened 2025-07-15
Convert Entity.key into an ArrayField of TextElement

eemeli:
~This builds on top of #3716, and will need a merge or rebase after that lands.~ Edit: Done.

This PR replaces the `Entity.key` and `Entity.context` text fields with an `Entity.key` that's an array of text, to match the moz.l10n `tuple[str, ...]` identifier type. Because the new field name is the same as the old one, there's an intermediate step where it's called `Entity.new_key`.


translate/src/modules/entitydetails/components/Metadata.tsx
```diff
@@ -184,12 +184,15 @@ const EntityContext = ({
...
     <Property title='CONTEXT' className='context'>
-      {context ? (
-        <>
-          {context}
-          <span className='divider'>&bull;</span>
-        </>
-      ) : null}
+      {key
+        .filter((_, i) => i !== 0 || format !== 'po')
+        .toReversed()
```
eemeli, line 189: `Array.p.toReversed()` is relatively recent, but supported in Firefox 115 (oldest ESR) and later. Adding it did require updating `typescript` as well as moving the CI actions to run on the Node v22, its current LTS version.

mathjazz (commented):
Looks good, pushing to stage to test.

mathjazz (approved):
Looks good. Let's ship!

PR #3722, opened 2025-07-15
Do not import empty gettext plural translations & reject any current ones

eemeli:
This is fixing a regresssion from #3635.

### DIFF
--- a/pontoon/base/migrations/0086_reject_empty_gettext_plurals.py
+++ b/pontoon/base/migrations/0086_reject_empty_gettext_plurals.py
@@ -1,0 +1,46 @@
+from django.db import migrations
+from django.utils import timezone
+
+
+def reject_empty_gettext_plurals(apps, schema_editor):
+    ActionLog = apps.get_model("actionlog", "ActionLog")
+    User = apps.get_model("auth", "User")
+    Translation = apps.get_model("base", "Translation")
+
+    now = timezone.now()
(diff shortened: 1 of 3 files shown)


pontoon/base/migrations/0086_delete_empty_gettext_plurals.py
```diff
@@ -0,0 +1,38 @@
...
+    mf2_empty_plural = ".input {$n :number}\n.match $n\none {{}}\n* {{}}"
+    translations = Translation.objects.filter(string=mf2_empty_plural)
+    ActionLog.objects.bulk_create(
+        [
+            ActionLog(
+                action_type="translation:deleted",
+                created_at=now,
+                performed_by=sync_user,
+                entity_id=t.entity_id,
+                locale_id=t.locale_id,
```
eemeli, line 22: With the `.bulk_create()`, the custom `ActionLog.save()` validation is skipped; it would fail for these, but using the `_id` values actually works fine for this use.

mathjazz:
Adding this important bit of information from Slack for reference:

> nl localizer reported [an empty imported string](https://pontoon.mozilla.org/de/amo-frontend/LC_MESSAGES/amo.po/?string=321404), another one with plural forms. It shows as a translated string when it's not.

mathjazz (approved):

PR #3725, opened 2025-07-16
Exclude disabled and system projects in API

### LINKED ISSUE #3723: API v2 /locales is including disabled projects
Check "af" for example.
https://pontoon.mozilla.org/api/v2/locales/

This was reported and fixed as #3706, regressed in #3702. I think there were some performance issues, but I can't see anything explaining the change.

Is it possible to have tests to avoid this happening again in the future?

mathjazz:
Fix #3723.

### DIFF
--- a/pontoon/api/serializers.py
+++ b/pontoon/api/serializers.py
@@ -132,7 +132,7 @@ class NestedProjectSerializer(TranslationStatsMixin, ProjectSerializer):
         fields = ProjectSerializer.Meta.fields + ["tags", "locales"]
 
     def get_locales(self, obj):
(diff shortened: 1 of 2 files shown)


functionzz:
For this change, didn't it previously cause some regressions? Is there a new reason we are adding it back?

mathjazz:
Oh, you're right. This is regressing DB performance. Let me work on that here.

functionzz:
Sorry! The results in `/api/v2/locales` and `/api/v2/projects` looks good.

functionzz (approved):
Looks good to me.

PR #3727, opened 2025-07-17
Add tests to REST API views

### LINKED ISSUE #3724: Add tests for DRF API
The code powering API endpoints based on Django REST Framework doesn't have any unit tests. Let's fix that.

functionzz:
This PR intends to resolve #3724.

### DIFF
--- a/pontoon/api/tests/test_views.py
+++ b/pontoon/api/tests/test_views.py
@@ -235,65 +235,7 @@ def test_project(django_assert_num_queries):
 
 
 @pytest.mark.django_db
-def test_projects(
-    django_assert_num_queries,
-):
-    ProjectFactory.create_batch(3, disabled=True)
-    ProjectFactory.create_batch(3, system_project=True)
-    with django_assert_num_queries(4):
-        response = APIClient().get("/api/v2/projects/")
-
-    assert response.status_code == 200
-
-    expected_results = [
-        {
-            "slug": p.slug,
-            "name": p.name,
-            "priority": p.priority,
-            "deadline": p.deadline,
-            "visibility": p.visibility,
-            "contact": p.contact,
-            "info": p.info,
-            "system_project": p.system_project,
-            "disabled": p.disabled,
-            "sync_disabled": p.sync_disabled,
-            "pretranslation_enabled": p.pretranslation_enabled,
-            "total_strings": p.total_strings,
-            "approved_strings": p.approved_strings,
-            "pretranslated_strings": p.pretranslated_strings,
-            "strings_with_warnings": p.strings_with_warnings,
-            "strings_with_errors": p.strings_with_errors,
-            "missing_strings": p.missing_strings,
-            "unreviewed_strings": p.unreviewed_strings,
-            "complete": p.complete,
-            "tags": [
-                tag.name for tag in sorted(p.tags.all(), key=lambda tag: tag.name)
-            ],
-            "locales": [
-                locale.code
-                for locale in sorted(p.locales.all(), key=lambda locale: locale.code)
-            ],
-        }
-        for p in sorted(
-            Project.objects.filter(disabled=False, system_project=False),
-            key=lambda p: p.pk,
-        )
-    ]
-
-    for project in response.data["results"]:
-        project["locales"] = sorted(project["locales"])
-        project["tags"] = sorted(project["tags"])
-
-    results = sorted(response.data["results"], key=lambda p: p["slug"])
-    expected_results = sorted(expected_results, key=lambda p: p["slug"])
-
-    # includes Terminology project
-    assert response.data["count"] == 1
-    assert results == expected_results
-
-
-@pytest.mark.django_db
-def test_all_projects_stats(django_assert_num_queries):
+def test_projects(django_assert_num_queries):
     locale_a = LocaleFactory(
         code="kg",
         name="Klingon",


pontoon/api/tests/test_views.py
```diff
@@ -0,0 +1,165 @@
...
+
+from rest_framework.test import APIClient
+
+from pontoon.base.models.project import Project
+from pontoon.base.models.translation_memory import TranslationMemoryEntry
+from pontoon.terminology.models import Term, TermTranslation
+from pontoon.test.factories import LocaleFactory, ProjectFactory
+
+
+@pytest.fixture
```
mathjazz: As discussed at the call, let's drop the fixtures and move the code to the tests where they are actually used, even if that means the code will be repeated.

pontoon/api/tests/test_views.py
```diff
@@ -0,0 +1,165 @@
...
+def test_locale(locale_a):
+    response = APIClient().get(
+        f"/api/v2/locales/{locale_a.code}/", HTTP_ACCEPT="application/json"
+    )
+    assert response.status_code == 200
+    assert isinstance(response.data, dict)
+
+
+@pytest.mark.django_db
+def test_locales_optimization(django_assert_num_queries, locales):
```
mathjazz: I would drop `_optimization` from the name, since we're also testing for a successful response.

```suggestion
def test_locales(django_assert_num_queries, locales):
```

pontoon/api/tests/test_views.py
```diff
@@ -0,0 +1,165 @@
...
+    with django_assert_num_queries(3):
```
mathjazz: Nice! There's also `django_assert_max_num_queries()`, which would require less changes to the `num` parameter going forward. But let's start with the fixed value and see how it scales.

pontoon/api/tests/test_views.py
```diff
@@ -0,0 +1,165 @@
...
+    regular_projects,
+    disabled_projects,
+    system_projects,
+):
+    with django_assert_num_queries(4):
+        response = APIClient().get("/api/v2/projects/")
+
+    assert isinstance(response.data, dict)
+    assert response.status_code == 200
+    # assert len(response["data"]["result"]) == 3
```
mathjazz: What's the purpose of comment this line out, but leaving it in?

pontoon/api/tests/test_views.py
```diff
@@ -0,0 +1,165 @@
...
+
+@pytest.mark.django_db
+def test_projects_flags_optimization(
+    django_assert_num_queries,
+    regular_projects,
+    disabled_projects,
+    system_projects,
+):
+    with django_assert_num_queries(4):
+        response = APIClient().get("/api/v2/projects/?includeSystem&includeDisabled")
```
mathjazz:
```suggestion
        response = APIClient().get("/api/v2/projects/?include_system&include_disabled")
```

pontoon/api/tests/test_views.py
```diff
@@ -0,0 +1,165 @@
...
+
+    assert isinstance(response.data, dict)
+    assert response.status_code == 200
+
+
+# NON FUNCTIONAL
```
mathjazz: What does that mean?

mathjazz (commented):
Nice work! For each of the tests I'd also test the full response JSON.

pontoon/api/tests/test_views.py
functionzz: not supposed to be there, let me remove that

pontoon/api/tests/test_views.py
functionzz: not supposed to be there!

flodolo:
Is it possible to test also the number of translated/reviewed strings, to make sure that they match (right now they're all 0) and we don't include excluded projects in stats? That's one of the issues we've seen when porting queries to API v2

pontoon/api/tests/test_views.py
```diff
@@ -0,0 +1,580 @@
...
+def test_locale():
+    locale_a = factories.LocaleFactory(
+        code="kg",
+        name="Klingon",
+    )
+    response = APIClient().get(
+        f"/api/v2/locales/{locale_a.code}/", HTTP_ACCEPT="application/json"
+    )
+    assert response.status_code == 200
+    assert isinstance(response.data, dict)
```
mathjazz: That's unnecessary given the assertion of the entire `response.data` content below.

pontoon/api/tests/test_views.py
```diff
@@ -0,0 +1,580 @@
...
+        "total_strings": 0,
+        "approved_strings": 0,
+        "pretranslated_strings": 0,
+        "strings_with_warnings": 0,
+        "strings_with_errors": 0,
+        "missing_strings": 0,
+        "unreviewed_strings": 0,
+        "complete": True,
+        "tags": [],
+        "locales": [
```
mathjazz: The entire `locales` list looks like an overkill.

pontoon/api/tests/test_views.py
```diff
@@ -0,0 +1,580 @@
...
+    assert project_terminology in response.data["results"]
+
+
+@pytest.mark.django_db
+def test_projects_flags(
+    django_assert_num_queries,
+):
+    ProjectFactory.create_batch(3, visibility=Project.Visibility.PUBLIC) + list(
+        Project.objects.filter(slug__in=["terminology"])
+    )
```
mathjazz: I don't understand what are you trying to accomplish with `+ list()`.

pontoon/api/tests/test_views.py
```diff
@@ -0,0 +1,580 @@
...
+    ProjectFactory.create_batch(3, system_project=True) + list(
+        Project.objects.filter(slug__in=["tutorial"])
+    )
+    with django_assert_num_queries(4):
+        response = APIClient().get("/api/v2/projects/?include_system&include_disabled")
+
+    assert isinstance(response.data, dict)
+    assert response.status_code == 200
+
+    project_terminology = {
```
mathjazz: The Terminology project is neither system nor disabled, so you're not actually testing flags here.

pontoon/api/tests/test_views.py
functionzz: i think that's a relic of me copy pasting code from test_schema. Removed

pontoon/api/tests/test_views.py
```diff
@@ -0,0 +1,615 @@
...
+    expected_results = sorted(expected_results, key=lambda p: p["slug"])
+
+    # includes Terminology and 3 other disabled projects
+    assert response.data["count"] == 4
+    assert results == expected_results
+
+
+@pytest.mark.django_db
+def test_project_translation_stats():
+    print(Project.objects.all())
```
mathjazz: Please remove?

pontoon/api/tests/test_views.py
```diff
@@ -0,0 +1,615 @@
...
+        "strings_with_errors": 0,
+        "missing_strings": 0,
+        "unreviewed_strings": 0,
+        "complete": True,
+        "projects": ["terminology"],
+    }
+
+
+@pytest.mark.django_db
+def test_locale_stats():
```
mathjazz: Wouldn't it be simpler to join this test with `test_locale()`? 
And same for `test_project_stats()` and `test_project()`?

Also, why don't we check the number of DB queries in all tests?

pontoon/api/tests/test_views.py
```diff
@@ -0,0 +1,615 @@
...
+    assert results == expected_results
+
+
+@pytest.mark.django_db
+def test_project_translation_stats():
+    print(Project.objects.all())
+
+
+@pytest.mark.django_db
+def test_project_locale():
```
mathjazz: I'd create and tests some stats here, too. And also verify the number of DB queries.

pontoon/api/tests/test_views.py
```diff
@@ -0,0 +1,615 @@
...
+    results = sorted(response.data["results"], key=lambda p: p["slug"])
+    expected_results = sorted(expected_results, key=lambda p: p["slug"])
+
+    # includes Terminology project
+    assert response.data["count"] == 1
+    assert results == expected_results
+
+
+@pytest.mark.django_db
+def test_all_projects_stats():
```
mathjazz: Why tests stats for `/projects/`, but not for `/locales/`?

mathjazz (commented):
Nice progress! Left some comments and questions.

pontoon/api/tests/test_views.py
```diff
@@ -0,0 +1,675 @@
...
+        "missing_strings": 5,
+        "unreviewed_strings": 5,
+        "complete": False,
+        "tags": [],
+        "localizations": {},
+    }
+
+
+@pytest.mark.django_db
+def test_regular_projects(
```
mathjazz:
```suggestion
def test_projects(
```

pontoon/api/tests/test_views.py
```diff
@@ -0,0 +1,675 @@
...
+    results = sorted(response.data["results"], key=lambda p: p["slug"])
+    expected_results = sorted(expected_results, key=lambda p: p["slug"])
+
+    # includes Terminology project
+    assert response.data["count"] == 1
+    assert results == expected_results
+
+
+@pytest.mark.django_db
+def test_all_projects_stats(django_assert_num_queries):
```
mathjazz: This and the `test_projects()` above should be just one test case, same as we joined test cases for locales.

functionzz: So test_regular_projects(test_projects now) tests the endpoint `/api/v2/projects/` while test_all_projects_stats tests `/api/v2/projects/?include_system&include_disabled`. Should I remove one of these endpoints in order to merge these 2 test together?

mathjazz: Oh, good point. I think you can remove `test_all_projects_stats()` and move stats test to `test_projects`, because we already separatelly test system and disabled projects.

PR #3729, opened 2025-07-21
Personal Access Tokens for REST API authentication

### LINKED ISSUE #2502: API authentication for automated interaction
At the end of 2022, Chrome browser will deprecate extensions manifest v2, which will also mean quite limited ability for Pontoon Add-on to communicate with Pontoon server using the session cookie as authorization. As author of Pontoon Add-on, I would like Pontoon to support better authentication mechanism for its API.

There are probably several ways how this could be implemented, so I will just try to list them here. I understand Pontoon does not have coherent API per-se, so it may be an overkill to go for a full-blow separation of the authN and authZ from frontend cookies to some tokens etc. Some easier but less reusable methods I can see:
- basic/digest auth header (but Pontoon does not always have a way to validate user credentials, e.g. with FxA, right?)
- session cookie or ID accepted also via another header
- some form of API client token, which Pontoon Add-on could obtain (understand _steal_) via automated website interaction, or via the messaging API which is now used for its detection, or the user would need to generate and provide for Pontoon Add-on in a semi-assisted way

functionzz:
This PR will hopefully resolve #2502, the end goal being that users of the Pontoon external REST API will be able to use Personal Access Token(PAT) authentication in order to access endpoints.

TODO:
- Errors for max tokens reached + form errors need to be displayed & functional [COMPLETE]
- General css, jQuery(if needed), copy pasting token [COMPLETE]
- Perhaps a revamp of the settings page to put PATs in a place more accessible [COMPLETE]
- blocking people who spam the api [COMPLETE]
- visual warning for revoked token [COMPLETE]


mathjazz:
We've made the following decisions on the Pontoon call yesterday:

> * Max expiration date?

We'll just set the expiration date to 1 year for now for all tokens, so the users will not have the ability to manually set the expiration date. Users will need to delete tokens in order to make them expire earlier.

> * blocking people who spam the api

We can leverage the [`ThrottleIpMiddleware`](https://github.com/mozilla/pontoon/blob/main/pontoon/base/middleware.py#L84C7-L84C27) for this purpose.

pontoon/api/authentication.py
```diff
@@ -0,0 +1,43 @@
...
+        except PersonalAccessToken.DoesNotExist:
+            raise AuthenticationFailed({"detail": "Invalid token."})
+
+        if not check_password(unhashed_token, pat.token_hash):
+            raise AuthenticationFailed({"detail": "Invalid token."})
+
+        if pat.revoked:
+            raise AuthenticationFailed({"detail": "Token has been revoked."})
+
+        if pat.expires_at and pat.expires_at.astimezone() < timezone.now():
```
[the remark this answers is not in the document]
functionzz: This is true, although I will need help resolving this one to have a time that exactly fits the user's timezone.

pontoon/api/urls.py
```diff
@@ -58,6 +58,8 @@
         name="user-actions",
     ),
 ]
+
+
```
mathjazz, line 62: :-)

pontoon/api/migrations/0007_alter_personalaccesstoken_name.py
```diff
@@ -0,0 +1,17 @@
...
+        ("api", "0006_remove_personalaccesstoken_description"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="personalaccesstoken",
+            name="name",
+            field=models.CharField(max_length=32),
+        ),
+    ]
```
mathjazz: It would be nice to replace these migrations with just one migration before we merge the PR.

pontoon/api/views.py
```diff
@@ -283,6 +279,7 @@ def get_queryset(self):
 
 
 class ProjectLocaleIndividualView(generics.RetrieveAPIView):
+    queryset = ProjectLocale.objects.all().prefetch_related("project", "locale")
```
mathjazz: What is the purpose of this change?

pontoon/api/views.py
```diff
@@ -128,8 +126,6 @@ def get_user_actions(request, date, slug):
 
 
 class UserActionsView(APIView):
-    permission_classes = [IsAuthenticated]
```
mathjazz: This requirement should remain - it's needed for performance reasons.

mathjazz (commented):
Some high-level notes.

pontoon/api/migrations/0007_alter_personalaccesstoken_name.py
functionzz: yeah, it was just me flip flopping back and forth. Should I squash?

pontoon/api/views.py
functionzz: The projectLocale endpoint was not optimized, and we didn't realize why because it just called 1 ProjectLocale object. Now it should be optimized with this queryset.

pontoon/api/admin.py
```diff
@@ -0,0 +1,29 @@
...
+        "user__username",
+        "user__email",
+    )
+    list_display = (
+        "name",
+        "user",
+        "expires_at",
+        "revoked",
+        "created_at",
+        "last_used",
```
mathjazz:
```suggestion
        "created_at",
        "expires_at",
        "last_used",
        "revoked",
```

pontoon/api/models.py
```diff
@@ -0,0 +1,15 @@
...
+class PersonalAccessToken(models.Model):
+    token_hash = models.CharField(max_length=128)  # SHA-256
+    user = models.ForeignKey(
+        User, on_delete=models.CASCADE, related_name="personal_access_tokens"
+    )
+    name = models.CharField(max_length=32)
+    expires_at = models.DateTimeField()
+    revoked = models.BooleanField(default=False)
+    created_at = models.DateTimeField(auto_now_add=True)
+    last_used = models.DateTimeField(null=True, blank=True)
```
mathjazz:
```suggestion
    created_at = models.DateTimeField(auto_now_add=True)
    expires_at = models.DateTimeField()
    last_used = models.DateTimeField(null=True, blank=True)
    revoked = models.BooleanField(default=False)
```

pontoon/base/forms.py
```diff
@@ -397,3 +398,32 @@ class AddCommentForm(forms.Form):
...
+        model = PersonalAccessToken
+        fields = ("name",)
+
+    def __init__(self, *args, **kwargs):
+        self.user = kwargs.pop("user", None)
+        super().__init__(*args, **kwargs)
+
+    def clean_name(self):
+        if not self.user:
+            raise ValidationError("User must be provided to create a token.")
```
mathjazz: Not sure there's a need to validate that on the form level - data model already takes care of it.

pontoon/base/forms.py
```diff
@@ -397,3 +398,32 @@ class AddCommentForm(forms.Form):
...
+    """
+    Form for creating Personal Access Tokens.
+    """
+
+    name = forms.RegexField(
+        regex=r"^[a-zA-Z0-9-]+$",
+        error_messages={
+            "invalid": "Name may only contain letters, numbers, and hyphens."
+        },
+    )
```
mathjazz: Why do we use this limitation? Why not just allow any character the model field allows? I think GitHub does the same.

pontoon/contributors/static/js/settings.js
```diff
@@ -47,6 +47,123 @@ $(function () {
...
+                    <span>Expires on:</span> 
+                    <span class="date">${response.data['new_token_expires_at']}</span>
+                  </div>
+                  <div class="token-info">
+                    <span class="icon fas fa-calendar-alt"></span>
+                    <span>Last used:</span> 
+                    <span class="date">
+                      ${response.data['new_token_last_used'] ? response.data['new_token_last_used'] : 'Never'}
+                    </span>
+                  </div>
```
mathjazz: No need to show the Last used info for the newly created token. It will always be "Never".

pontoon/contributors/views.py
```diff
@@ -411,13 +423,122 @@ def settings(request):
...
+            secrets.choice(string.ascii_letters + string.digits) for _ in range(43)
+        )
+
+    try:
+        name = request.POST["name"]
+    except KeyError:
+        return JsonResponse(
+            {"status": "error", "message": "Token name is required."}, status=400
+        )
+    if request.method == "POST":
```
mathjazz: Note that there's a `require_POST` decorator for this purpose, and it also saves us a level of indentation.

pontoon/contributors/views.py
```diff
@@ -411,13 +423,122 @@ def settings(request):
...
+        pat_count = PersonalAccessToken.objects.filter(user=request.user).count()
+
+        if pat_count < PERSONAL_ACCESS_TOKEN_MAX_COUNT:
```
mathjazz: You can save a level of indentation by flipping the condition. If the count is 10 or more, you return the error message. If not, you just continue after the `if` block, no need for an `else`.

pontoon/contributors/views.py
```diff
@@ -411,13 +423,122 @@ def settings(request):
...
+            create_token_form = forms.CreateTokenForm(request.POST, user=request.user)
+
+            if create_token_form.is_valid():
```
mathjazz: Same as above: flip the condition, return when invalid, continue on the same level as `if` when valid.

pontoon/contributors/views.py
```diff
@@ -411,13 +423,122 @@ def settings(request):
...
+                date_midnight = timezone.now().date() + timedelta(days=365)
+                token.name = name
+                token.token_hash = make_password(token_unhashed)
+                token.expires_at = timezone.make_aware(
+                    datetime.datetime.combine(date_midnight, time.min)
+                )
+                token.user = request.user
+                token.save()
+                token_id = token.id
+                token_secret = f"{token_id}_{token_unhashed}"
```
mathjazz: You should be able to drop `token_id`.

```suggestion
                token_secret = f"{token.id}_{token_unhashed}"
```

pontoon/contributors/views.py
```diff
@@ -411,13 +423,122 @@ def settings(request):
...
+
+                return JsonResponse(
+                    {
+                        "status": "success",
+                        "message": "Form submitted successfully!",
+                        "data": {
+                            "new_token_id": token.id,
+                            "new_token_name": token.name,
+                            "new_token_last_used": token.last_used.strftime(
+                                "%B, %d, %Y"
```
mathjazz:
```suggestion
                                "%B %d, %Y"
```

pontoon/contributors/views.py
```diff
@@ -411,13 +423,122 @@ def settings(request):
...
+                            )
+                            if token.last_used
+                            else None,
+                            "new_token_expires_at": token.expires_at.strftime(
+                                "%B, %d, %Y"
```
mathjazz:
```suggestion
                                "%B %d, %Y"
```

pontoon/contributors/views.py
mathjazz: Actually, you can probably drop `new_token_last_used` as mentioned above.

pontoon/contributors/views.py
```diff
@@ -411,13 +423,122 @@ def settings(request):
...
+            )
+    return JsonResponse(
+        {"status": "error", "message": "Invalid request method."}, status=405
+    )
+
+
+@login_required(redirect_field_name="", login_url="/403")
+@require_AJAX
+def delete_token(request, token_id):
+    if request.method == "POST":
```
mathjazz: Use either `require_POST` or the `DELETE` method and flip the condition to reduce indentation.

pontoon/api/urls.py
mathjazz: Oh wait, that was intentional!

mathjazz (commented):
This works really well, good job! Left a few comments, nothing major.

Please add some tests for `authentication.py` and the `generate_token()` and `delete_token()` views.

Screenshot 2025-08-07 at 19 12 49

I'd also ask to make the following design changes in the new token section:
* Use full-width for the token field. That way the message underneath will fit into one line. Either move "Copied!" message above the field or use `Pontoon.endLoader()`.
* Align "Expires on" horizontally with the token name.
* Expand the token name field to fit the entire section width and style it like other text fields on the page. Then the error message can also occpuy the entire width and is less likely to break into multiple lines.
* Style the Generate token button same as the Save button (`font-weight`).

pontoon/contributors/static/js/settings.js
```diff
@@ -47,6 +47,148 @@ $(function () {
...
+  $(document).on('click', '.copy-btn', function (e) {
+    e.preventDefault();
+  });
+
+  $('.token-name-input').on('input', function () {
+    if ($(this).val().trim() !== '') {
+      $('.generate-token-btn').prop('disabled', false); // Enable the button
+    } else {
+      $('.generate-token-btn').prop('disabled', true); // Disable the button
+    }
```
mathjazz: That should be the same:
```suggestion
    $('.generate-token-btn').prop('disabled', $(this).val().trim() === '');
```

pontoon/contributors/static/js/settings.js
```diff
@@ -47,6 +47,148 @@ $(function () {
...
+              .attr('tabindex', '-1')
+              .attr('data-clipboard-text', response.data['new_token_secret']),
+          );
+
+          tokenCard.append(tokenHeader, tokenInfoContainer, tokenDetails);
+          tokenCard.append(
+            $('<p>')
+              .addClass('copy-message')
+              .text(
+                'Make sure to copy your personal access token now as you will not be able to see this again.',
```
mathjazz:
```suggestion
                'Make sure to copy your personal access token now as you will not be able to see it again.',
```

pontoon/contributors/static/js/settings.js
```diff
@@ -47,6 +47,148 @@ $(function () {
...
+
+    $.ajax({
+      url: '/generate-token/',
+      type: 'POST',
+      data: {
+        name: $('.token-name-input').val(),
+        csrfmiddlewaretoken: csrfmiddlewaretoken,
+      },
+      success: function (response) {
+        if (response.status === 'success') {
```
mathjazz: You can drop the indentation by inverting the condition and quitting early.

pontoon/api/tests/test_views.py
```diff
@@ -615,3 +620,134 @@ def test_tm_search(django_assert_num_queries):
             }
         ],
     }
+
+
+@pytest.mark.django_db
+def test_authenticate_valid_token(member):
```
mathjazz: Please move `authentication.py` tests to `tests/test_authentication.py`.

pontoon/contributors/static/js/settings.js
```diff
@@ -47,6 +47,144 @@ $(function () {
...
+    $.ajax({
+      url: '/generate-token/',
+      type: 'POST',
+      data: {
+        name: $('.token-name-input').val(),
+        csrfmiddlewaretoken: csrfmiddlewaretoken,
+      },
+      success: function (response) {
+        if (response.status !== 'success') {
+          return;
```
mathjazz, line 65:
```suggestion
          Pontoon.endLoader('Oops, something went wrong.', 'error');
          return;
```

pontoon/contributors/static/js/settings.js
```diff
@@ -47,6 +47,144 @@ $(function () {
...
+        }
+
+        const li = $('<li>').attr(
+          'data-token-id',
+          response.data['new_token_id'],
+        );
```
mathjazz: I missed that you've dropped HTML string literals. Building the DOM manually in jQuery is fine for quick hacks, but brittle for long-term maintenance. What was the motivation for the change?

functionzz: I realized that using html made it possible to inject javascript into the code to break the application, the 2 options were validate the user input and prevent things like `><script>Alert!<script><` or bypass the issue with jQuery. I chose the second option, can rollback if we want to use the another option.

pontoon/contributors/static/js/settings.js
```diff
@@ -47,6 +47,115 @@ $(function () {
...
+        if (response.status !== 'success') {
+          Pontoon.endLoader('Oops, something went wrong.', 'error');
+          return;
+        }
+
+        const newTokenHTML = `
+        <li data-token-id="${response.data['new_token_id']}">
+            <div class="token-card created controls clearfix">
+            <div class="token-header">
+              <span class="token-name">${escapeHTML(response.data['new_token_name'])}</span>
```
mathjazz: At this point we already have the token name stored in the DB, so it better be safe. If we want to sanitize the input, we should do so on the backend before saving into the DB (perhaps using `bleach`, which we already use).

Also, we use Dot Notation instead of Bracket Notation.

```suggestion
              <span class="token-name">${response.data.new_token_name}</span>
```

mathjazz (commented):
Nice work!

One step to go. :)

I'd also like to make a few small CSS changes before merging.

pontoon/contributors/static/js/settings.js
functionzz: By using bleach the input is successfully sanitized, by removing any inline HTML. Problem is, it automatically escapes characters, which increases the length of input name, which can violate its max. length of 32 in the db.

For example, ">>>>>>>>>>>>>" crashes the server.

I can try to reconvert the strings back into symbols but at this point I'm questioning if that's worth it. In this case, can't I just go back to form validation and limit the input to numbers,characters etc.?

I'm not sure how github does it, they have a max token length which is respected by whatever is being typed, it is escaped. Maybe they store the un-escaped name directly into the db?

image

mathjazz: Can you actually trigger alerts if you input the same (or similar) into the input box with `escapeHTML` removed?

`<script>` tags added dynamically are usually treated as inert (non-executable) by modern browsers.

pontoon/base/forms.py
```diff
@@ -397,3 +398,27 @@ class AddCommentForm(forms.Form):
...
+    class Meta:
+        model = PersonalAccessToken
+        fields = ("name",)
+
+    def __init__(self, *args, **kwargs):
+        self.user = kwargs.pop("user", None)
+        super().__init__(*args, **kwargs)
+
+    def clean_name(self):
+        invalid_symbols = ["&", "<", ">", '"', "'"]
```
mathjazz: Do we need to block anything else then < and >?

```suggestion
        invalid_symbols = ["<", ">"]
```

functionzz: was proposing to keep the quotation marks just in case for potential SQL injection or other XSS attacks. Also the & can be used to write in escaped characters I suppose.

pontoon/base/forms.py
```diff
@@ -397,3 +398,27 @@ class AddCommentForm(forms.Form):
...
+    def __init__(self, *args, **kwargs):
+        self.user = kwargs.pop("user", None)
+        super().__init__(*args, **kwargs)
+
+    def clean_name(self):
+        invalid_symbols = ["<", ">"]
+
+        name = self.cleaned_data.get("name")
+        if any(symbol in name for symbol in invalid_symbols):
+            raise ValidationError("Name cannot contain &, <, >, \", or '.")
```
mathjazz: This message should match the symbols in `invalid_symbols`.

mathjazz:
@functionzz Flagging you for a review here, since you openned the PR and I can't request the review from you formally.

pontoon/api/views.py
mathjazz: This comment was marked as resolved, but the change is still in. Could you elaborate on this?

pontoon/api/views.py
```diff
@@ -269,7 +265,7 @@ def get_queryset(self):
 
 
 class ProjectIndividualView(generics.RetrieveAPIView):
-    serializer_class = NestedIndividualProjectSerializer
+    serializer_class = NestedProjectSerializer
```
mathjazz: What is the purpose of this change?

functionzz: I think I changed it to accommodate [issue 3761](https://github.com/mozilla/pontoon/pull/3761). I'll remove it as its unnecessary.

pontoon/api/views.py
mathjazz: Sorry, it seems like you just resolved the comment without addressing it for the 2nd time. 🤣 
Am I missing something?

functionzz:
I would add a happy crying emoji if there was one. I hope it doesn't turn into a sad crying emoji from regressions :..)

Issue #3735, opened 2025-07-23
Sign in button text is barely visible in the light theme


mathjazz:
<img width="1260" height="358" alt="Image" src="https://github.com/user-attachments/assets/afc45fd7-8610-4016-bc6b-4ab62712273c" />

mathjazz:
Assigned. :)

PR #3736, opened 2025-07-23
Fix: updated Sign In button styles

skools-here:
Fixes: #3735

Updated SignIn.css because SignIn button was not visible before .So made it bit dark because bg was white, also changes the color when user hovers.

### DIFF
--- a/translate/src/modules/user/components/SignIn.css
+++ b/translate/src/modules/user/components/SignIn.css
@@ -6,7 +6,7 @@
   background: transparent;
   border: 1.5px solid var(--status-translated);
   border-radius: 2px;
-  color: var(--dark-grey-1);
+  color: var(--white-1);
   font-size: 14px;
   font-weight: 300;
   height: 40px;


translate/src/modules/user/components/SignIn.css
```diff
@@ -6,7 +6,7 @@
   background: transparent;
   border: 1.5px solid var(--status-translated);
   border-radius: 2px;
-  color: white;
+  color: #000;
```
mathjazz: Please note that Pontoon uses themes, so rather than hardcoding color values, we use variables. 

See https://github.com/mozilla/pontoon/blob/main/pontoon/base/static/css/style.css#L665-L681 for the values that are used in all views but the translate view.

translate/src/modules/user/components/SignIn.css
```diff
@@ -16,5 +16,5 @@
 }
 
 .user-signin button:hover {
-  background-color: var(--status-translated);
+  background-color: #333;
```
mathjazz: That value should remain unchanged.

mathjazz (commented):
Thanks for your patch! See comments inline.

translate/src/modules/user/components/SignIn.css
```diff
@@ -6,7 +6,7 @@
   background: transparent;
   border: 1.5px solid var(--status-translated);
   border-radius: 2px;
-  color: white;
+  color: var(--dark-grey-1);
```
mathjazz: Why not use the same value as in the other frontend, i.e. `var(--white-1)`?

translate/src/modules/user/components/SignIn.css
```diff
@@ -16,5 +16,5 @@
 }
 
 .user-signin button:hover {
-  background-color: var(--status-translated);
+   background-color: var(--status-translated);
```
mathjazz: Mind the extra space.

mathjazz (commented):
Thanks for the addressing the comments quickly!

mathjazz (approved):
Nice work!

Welcome to Pontoon! :)

Issue #3738, opened 2025-07-24
Enabling/disabling sync triggers a forced sync


flodolo:
Because of #3737, I disabled Sync for Firefox. I then started a manual sync, and was surprised to see

```
Jul 24 04:05:54 Using forced sync due to project config change on 2025-07-24 03:48:50.654017+00:00
Jul 24 04:05:54 [firefox] Sync start
Jul 24 04:05:54 [firefox] Repo at ffcb71fca
Jul 24 04:05:54 [firefox] Considering all files as changed
Jul 24 04:05:55 [firefox] Repo updated from 3ce3d8ac5b7 to df0c0a6ca14
Jul 24 04:05:55 [firefox] Considering all files as changed
Jul 24 04:06:18 [firefox] Syncing entities from repo...
```

The only change I made was enabling/disabling sync in the project 🤔

flodolo:
I understand this is an assumption on my side, but I can't think of another reason — maybe looking into logs might give one.

PR #3739, opened 2025-07-24
Include section id in .key for ini files

### LINKED ISSUE #3737: Existing .ini files showing up as empty in Pontoon after last update
@bcolsson @eemeli 
Not sure what happened, but new strings were exposed 4 hours ago, and all .ini files are marked as empty (while they were translated).

Not sure if related, but the notification says that Firefox has 153 new strings, but I can only see 46 missings, all from .ini files.

eemeli:
Fixes #3737, a regression from #3721.

To un-reject the lost translations:
1. Disable sync for the affected project.
2. Ensure that its repo includes the .ini files in their previous state.
3. Deploy with this PR's changes.
4. Re-enable sync, which due to #3738 will trigger a force-sync.


mathjazz (approved):
Looks good. I also tested locally and on stage.

PR #3751, opened 2025-08-04
Run translate-toolkit checks separately for each gettext plural pattern, rather than the message as a whole

### LINKED ISSUE #3750: Translate Toolkit checks show unrelated warnings when submitting gettext plural translations
I assume they're examining the MF2 syntax?

Image

eemeli:
Fixes #3750 

Rather than skipping all of the translate-toolkit checks as suggested in https://github.com/mozilla/pontoon/issues/3750#issuecomment-3151228293, let's run them separately for each of the patterns, as was done previously.

This uses the plural pattern of the source for all variants except for `one`.

### DIFF
--- a/pontoon/checks/libraries/__init__.py
+++ b/pontoon/checks/libraries/__init__.py
@@ -1,4 +1,12 @@
+from moz.l10n.formats.mf2 import mf2_parse_message, mf2_serialize_pattern
+from moz.l10n.model import CatchallKey, Pattern, PatternMessage, SelectMessage
+
 from . import compare_locales, pontoon_db, pontoon_non_db, translate_toolkit
+
+
+def as_gettext(pattern: Pattern) -> str:
+    s = "".join(mf2_serialize_pattern(pattern))
+    return s.replace(r"\{", "{").replace(r"\}", "}")
 
 
 def run_checks(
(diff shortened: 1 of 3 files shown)


pontoon/checks/tests/test_translate_toolkit.py
```diff
@@ -1,42 +1,34 @@
...
-def test_tt_invalid_translation(mock_locale):
+def test_tt_invalid_translation():
     """
     Check if translate toolkit returns errors if chek
     """
     assert run_checks(
         "Original string",
         "Translation \\q",
-        mock_locale,
+        "en-US",
```
mathjazz: This change (and all other changes to this file) are out of scope of https://github.com/mozilla/pontoon/issues/3750, which this PR is addressing.

pontoon/checks/tests/test_libraries.py
```diff
@@ -181,3 +181,20 @@ def test_tt_disabled_checks(
...
     )
+
+
+def test_tt_gettext_checks():
+    entity = MagicMock()
+    entity.resource.path = "file.po"
+    entity.resource.format = Resource.Format.PO
+    entity.resource.all.return_value = []
+    entity.string = (
+        ".input {$n :number}\n.match $n\none {{One user}}\n* {{%(count)s Users}}"
```
mathjazz: Can we make this syntax more readable using `dedent()` or something? The bit that's particularly hard to digest is `none`.

pontoon/checks/libraries/__init__.py
```diff
@@ -72,13 +73,35 @@ def run_checks(
...
+                        if keys == ("one",):
+                            src = as_gettext(src_msg.variants[keys])
+                        else:
+                            src = s0
+                        tt_patterns.append((src, as_gettext(pattern)))
+                else:
+                    tt_patterns.append((s0, as_gettext(tgt_msg.pattern)))
+            elif isinstance(tgt_msg, PatternMessage):
+                tt_patterns.append(
+                    (as_gettext(src_msg.pattern), as_gettext(tgt_msg.pattern))
```
mathjazz, line 93: Even for singular strings we need to use `as_gettext()`?

mathjazz (commented):
I tested locally and didn't spot any issues.

Since this is a fix for a regression, I'd refrain from making any changed other than what's necessary to fix the regression.

pontoon/checks/libraries/__init__.py
eemeli: Yes, because they can include escaped curly braces `\{` `\}`.

mathjazz (approved):

PR #3753, opened 2025-08-07
Add Entity.value & Entity.properties, update them during sync

eemeli:
This adds the data model representation of entities to the database and maintains their values during sync, but doesn't yet do anything with them.

The next step from here will likely be to add the value+properties to the `/get-entities/` response, and to ensure that the frontend can reconstruct the Entity.string value from them, so that it can be dropped from the `/get-entities/` response.


pontoon/base/migrations/0096_add_entity_value_and_properties.py
```diff
@@ -0,0 +1,74 @@
...
+    for entity in fluent_entities:
+        try:
+            fe = fluent_parse_entry(entity.string, with_linepos=False)
+            entity.value = message_to_json(fe.value)
+            entity.properties = {
+                name: message_to_json(msg) for name, msg in fe.properties.items()
+            } or None
+        except Exception:
+            if entity.obsolete and entity.resource.project.disabled:
+                entity.value = [entity.string]
```
eemeli: This is needed as a workaround for some obsolete strings in the disabled test-pilot-website project on pontoon.mozilla.org, which use an out-of-date (2017) Fluent syntax representation.

pontoon/sync/utils.py
```diff
@@ -118,3 +126,42 @@ def import_uploaded_file(
...
+
+
+def parse_pontoon_message(
+    db_format: str, string: str, meta: list[list[str]]
+) -> Message:
+    """
+    Parse an Entity.string or a Translation.string as a moz.l10n Message.
+
+    This function should be removed once the string message representations are removed.
+    """
```
eemeli: This function is currently only used by the base/0096 migration, but it'll be needed later for translations as well.

pontoon/base/migrations/0096_add_entity_value_and_properties.py
flodolo: Let's remove them from the DB instead 👍🏼

pontoon/sync/utils.py
mathjazz: Nit: Add TODO to the comment, so it'll be easier not to forget about it. Or file an issue?

mathjazz (commented):
Looks good. Data migrations and schema migrations should be split in separate migrations. In the past we've hit issue when this wasn't the case when squashing migrations.

Other than that, LGTM. I haven't tested the changeset due to its low potential to break things.

pontoon/sync/utils.py
eemeli:
```suggestion
def parse_pontoon_message(
    db_format: str, string: str, meta: list[list[str]]
) -> Message:
    """
    Parse an Entity.string or a Translation.string as a moz.l10n Message.

    TODO: This function should be removed once the string message representations are removed.
    """
```

PR #3757, opened 2025-08-12
Switch to gpt-4.1, refine prompt and update OpenAI package

flodolo:


pontoon/machinery/openai_service.py
```diff
@@ -28,6 +28,7 @@ def get_translation(
             - Clarity is Key: Ensure the translation conveys the original message in the clearest possible manner, without ambiguity or unnecessary complexity.
             - Consistent Simplicity: Maintain a consistent level of simplicity throughout the translation.
             - Preserve All HTML Tags: Ensure that all HTML tags present in the original text are retained in the translation.
+            - Avoid adding closing punctuation if it's not present in the original text.
```
flodolo: Made this change because I've noticed over time that it tends to add punctuation at the end of strings that don't have it in the source.

flodolo: I'm still the issue. I'll try to iterate more with the prompt, since updating the package is not urgent.

PR #3760, opened 2025-08-18
Fix incorrect display of recipient count when selecting contributor, translator or manager

### LINKED ISSUE #3758: Translators and Team Managers included when filtering by only Contributors within Messaging Center
When attempting to filter for users who are not Translators or Team Managers by selecting only "Contributor" for the User Role filter, the number of recipients that would be sent to is the same as if all three roles were selected. 

Image

Image

It appears to limit the number of recipients when selecting Translator and Team Managers only:

Image

functionzz:
This issue resolves #3758. It addresses 3 issues:

- Before, if no Locale/Project were toggled, there would be 0 Managers/Translators in the recipient count, while Contributors had recipients, which was misleading.
Now, if no Locale/Project is toggled, all locales and projects are automatically selected for User Role recipient counting.

- Before, selecting the Manager, Translator & Contributor user roles vs. selecting just the Contributor user role equated the same count of recipients.
Now, each selected role has its own respective recipient counts, meaning the Contributor role does not equal all recipients

- Before, deselecting Locale toggle or Project toggle did not actually deselect the Locales or Projects chosen, which made the recipient count incorrect.
Now, deselecting the Locale/Project toggles also deselects the now hidden Locales or Projects previously chosen, ensuring accurate user experience.

### DIFF
--- a/pontoon/messaging/views.py
+++ b/pontoon/messaging/views.py
@@ -125,10 +125,9 @@ def get_recipients(form):
 
     translations = Translation.objects.all()
     if form.cleaned_data.get("contributors"):
-        non_contributors = User.objects.filter(
-            pk__in=manager_ids
-        ) | User.objects.filter(pk__in=translator_ids)
-        recipients = User.objects.exclude(Q(pk__in=non_contributors) | Q(pk=-1))
+        recipients = User.objects.exclude(
+            Q(pk__in=manager_ids) | Q(pk__in=translator_ids) | Q(pk=-1)
+        )
 
     if form.cleaned_data.get("managers"):
         recipients = recipients | User.objects.filter(pk__in=manager_ids)


pontoon/messaging/views.py
```diff
@@ -120,9 +124,11 @@ def get_recipients(form):
     )
 
     translations = Translation.objects.all()
-
     if form.cleaned_data.get("contributors"):
-        recipients = User.objects.exclude(pk=-1)
+        non_contributors = User.objects.filter(
+            pk__in=manager_ids
+        ) | User.objects.filter(pk__in=translator_ids)
+        recipients = User.objects.exclude(Q(pk__in=non_contributors) | Q(pk=-1))
```
mathjazz: This is simpler, hits the DB only once and should result in the same query:

```suggestion
        recipients = User.objects.exclude(
            Q(pk__in=manager_ids) | Q(pk__in=translator_ids) | Q(pk=-1)
        )
```

PR #3761, opened 2025-08-19
Remove Graphql API usage from Pontoon Translate module

functionzz:
This PR rewrites the usage of GraphQL queries to REST URLs in line with our plan to [deprecate](https://github.com/mozilla/pontoon/issues/3676) our Pontoon GraphQL endpoint.

TODO:
- [x] Update README to mark REST API as beta and add a note the structure is not stable yet.
- [x] @mathjazz: Add same info to https://github.com/mozilla/pontoon/discussions/3745
- [x] @mathjazz: Add same info to https://github.com/common-voice/common-voice/issues/5010
- [x] @mathjazz: Add same info to https://github.com/mozilla/addons/issues/15756


functionzz:
I do have concerns about efficiency, as the fetchProject function has the potential to call the localization endpoint `/api/v2/{code}/{slug}` many times.

translate/src/context/Locale.ts
```diff
@@ -49,35 +49,35 @@ export async function updateLocale(locale: Locale, code: string) {
...
-      }
-    }
-  }`;
+  let res = await GET(`/api/v2/locales/${code}`);
 
-  const search = new URLSearchParams({ query });
-  const res = await GET('/graphql', search);
+  let localizations = [];
+  for (const project of res.projects) {
+    const localization = await GET(`/api/v2/${code}/${project}`);
```
mathjazz: Why don't we instead expand the API similarly as we did for the [projects](https://github.com/mozilla/pontoon/issues/3701)?

functionzz: let's do that, this will make it easier for others in the future

functionzz: OK, I'm currently attempting to expand the `/locales` endpoint to include localizations as well. There are some issues I'd like to address. 

The first one is, the current query includes a project`name` field, which is not present in the REST implementation. Should I also include that within the localization extension of the `/locales` endpoint? I'm hesitant to do so because it's not really consistent with the `/projects` localization implementation which does not include a "name" of the Locale.

The second one is, the current format of returning a localization with `/projects` is this:

```
"firefox": {
    "locale": "en-CA",
    "total_strings": 14302,
    "approved_strings": 14038,
    "pretranslated_strings": 0,
    "strings_with_warnings": 0,
    "strings_with_errors": 0,
    "missing_strings": 264,
    "unreviewed_strings": 0,
    "complete": false
}
```
I realized that working with this un-normalized output (with firefox as the key and an object as a value) is kind of tricky in typescript, and I was wondering if it would be better to have it normalized, like this:

```
{
    "slug": "firefox",
    "locale": "en-CA",
    "total_strings": 14302,
    "approved_strings": 14038,
    "pretranslated_strings": 0,
    "strings_with_warnings": 0,
    "strings_with_errors": 0,
    "missing_strings": 264,
    "unreviewed_strings": 0,
    "complete": false
}
```
If we do this however, it would be inconsistent with the `/projects` localization section, of which I don't know if we can change it? Maybe someone has already started working with the un-normalized version.

mathjazz: > The first one is, the current query includes a project`name` field, which is not present in the REST implementation. Should I also include that within the localization extension of the `/locales` endpoint? I'm hesitant to do so because it's not really consistent with the `/projects` localization implementation which does not include a "name" of the Locale.

How about we add `localizations.LOCALE.name` to the `/projects` endpoint?
And `localizations.PROJECT.name` to the `/locales` endpoint?

> I realized that working with this un-normalized output (with firefox as the key and an object as a value) is kind of tricky in typescript?

Why is it tricky? I think we want to use the dict instead of list, because it makes searching for a specific entry easier.

translate/src/context/Locale.ts
```diff
@@ -49,35 +49,40 @@ export async function updateLocale(locale: Locale, code: string) {
...
-    }
-  }`;
+  let res = await GET(`/api/v2/locales/${code}`);
 
-  const search = new URLSearchParams({ query });
-  const res = await GET('/graphql', search);
+  let localizations = [];
+  for (const localization of res.localizations) {
+    // Each item is an object with one key
+    const [slug, object] = Object.entries(localization)[0] as [
```
functionzz: @mathjazz, I want your opinion on this line, do you think this is OK or more like a trick?

pontoon/api/tests/test_views.py
```diff
@@ -74,6 +74,22 @@ def test_locale(django_assert_num_queries):
         "unreviewed_strings": 5,
         "complete": False,
         "projects": ["terminology"],
+        "localizations": [
+            {
+                "terminology": {
+                    "approved_strings": 0,
+                    "complete": True,
+                    "locale": "kg",
```
mathjazz: We already have locale in the parent dict, why repeat it in the `localizations` dict?

pontoon/base/models/project_locale.py
```diff
@@ -29,11 +29,20 @@ def visible(self):
...
+            query = self.filter(
+                locale__translatedresources__resource__project=project
+            ).prefetch_related("locale")
+            tr = "locale__translatedresources"
+        elif locale:
+            query = self.filter(
+                project__resources__translatedresources__locale=locale,
+                project__resources__translatedresources__resource__project__disabled=False,
+                project__resources__translatedresources__resource__project__system_project=False,
+                project__resources__translatedresources__resource__project__visibility="public",
```
mathjazz: Is that not working OK?

```suggestion
                project__disabled=False,
                project__system_project=False,
                project__visibility="public",
```

translate/src/context/Locale.ts
mathjazz: Hmmmm. Ideally we should be using the API response as is, without post-processing. It's sub-optimal if the API response is formatted such that it doesn't event support our own use cases.

mathjazz (commented):
We should expand the `test_project` test case to cover the new functionality. You might be able to use the `terminology` project or create a new one in the test.

translate/src/context/Locale.ts
functionzz: Yep, 100%. the problem is this: project: { slug: slug, name: object.project_name },, which is a remnant of the GraphQL implementation. I can try to add that into the localization section.

translate/src/context/Locale.ts
```diff
@@ -49,41 +50,16 @@ export async function updateLocale(locale: Locale, code: string) {
...
+  res = keysToCamelCase(res);
 
-  const next = res.data.locale as Omit<Locale, 'cldrPlurals'> & {
+  const next = res as Omit<Locale, 'cldrPlurals'> & {
     cldrPlurals: string;
   };
   const cldrPlurals = next.cldrPlurals
     .split(',')
     .map((i: string) => parseInt(i, 10));
-  const direction = next.direction.toLowerCase();
```
mathjazz, line 86: I'm curious why is this no longer needed.

functionzz: The GraphQL returns direction as all caps. The REST api only returns it as "ltr" or "rtl", which gets the same result essentially

mathjazz: Interesting. Not sure why or how GraphQL does that, but thanks for clarifying!

functionzz:
> > Note: this most recent implementatio overfetches data from the API.
> 
> Could you elaborate on this?

I mean that the data fetched is more than needed for the feature to work i.e the field `population` is fetched and included as part of the `/locales` api response, but never actually used within the `Locale.ts` file.

We could not do anything about this or I could fetch the data in this manner, only specifying the fields needed:


```
res = {
    direction: res.direction,
    teamDescription: res.team_description,
    ...etc.
}
```

mathjazz:
> We could not do anything about this or I could fetch the data in this manner, only specifying the fields needed:

No, I think it's OK like that - the data still needs to go through the network, which is the slowest part.

functionzz:
am not sure if including all locales within the test is something you want to see

mathjazz (approved):
Great work again! Good job!

PR #3765, opened 2025-09-01
Use moz.l10n for simple preview generation, adding gettext support

### LINKED ISSUE #3741: Gettext TM entries contain `\` escapes
See this string and it's TM entry:
https://pontoon.mozilla.org/pl/all-projects/all-resources/?string=210094

> The MF2 syntax that's currently temporarily used for gettext serialization in the database is leaking through. We're not parsing the syntax from gettext messages, so e.g. `{url}` is being treated as a literal sequence of five characters. But because the curly braces are special in the MF2 syntax, they need to be `\{ and \}` escaped in the form we're storing them in the database.
> 
> It's a bug that the `\` escapes are ending up in the TM values.

### LINKED ISSUE #3742: Gettext translations in the Latest translation popup contain \ escapes
Image

> The MF2 syntax that's currently temporarily used for gettext serialization in the database is leaking through. We're not parsing the syntax from gettext messages, so e.g. `{url}` is being treated as a literal sequence of five characters. But because the curly braces are special in the MF2 syntax, they need to be `\{ and \}` escaped in the form we're storing them in the database.
> 
> It's a bug that they're not being stripped in the last-translation popup.

eemeli:
Fixes #3741
Fixes #3742

### DIFF
--- a/pontoon/base/migrations/0098_fix_gettext_translationmemory.py
+++ b/pontoon/base/migrations/0098_fix_gettext_translationmemory.py
@@ -1,0 +1,43 @@
+from moz.l10n.formats.mf2 import mf2_parse_message
+from moz.l10n.message import serialize_message
+from moz.l10n.model import CatchallKey, PatternMessage, SelectMessage
+
+from django.db import migrations, models
+
+
+def fix_string(string: str) -> str:
+    # Matches implementation in pontoon/base/simple_preview.py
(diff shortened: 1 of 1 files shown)


mathjazz (approved):
Changes look good.

I tested on stage:
- the migration runs fast
- the bugs are fixed

👍

PR #3769, opened 2025-09-07
[IMP] Switching from jest to vitest

### LINKED ISSUE #3768: Replace Jest with Vitest as test runner
While working on #3766 I was struck again by the clunkiness of Jest, and the difficulties of getting it to work with ES modules.

We would probably be better off switching instead to [Vitest](https://vitest.dev/) as our frontend test runner, esp. as our use of Jest isn't very complex.

I'm marking this as a "good first issue", given that this transition ought to be relatively straightforward.

Raj-G07:
## Description

- Migrated some unit tests from **Enzyme + Jest** to **React Testing Library + Vitest**.
- Verified that the converted tests are passing with the new setup.
- To check test use `npx vitest xyz.test.jsx/tsx`
- Some tests are still using **Enzyme + Sinon + Jest** — those will be migrated in follow-up PRs to keep this change small and easier to review.

## Notes
- Ensures partial migration progress without blocking ongoing development.
- Next PRs will complete the full migration.
- Fixes #3768


Raj-G07:
@eemeli Have a look at this PR, after your review. I’ll continue with the next batch of migrations if the approach looks good.

translate/src/App.tsx
```diff
@@ -5,7 +5,7 @@ import React, { useContext, useEffect, useState } from 'react';
 import './App.css';
 
 import { EntitiesList as EntitiesListContext } from './context/EntitiesList';
-import { EntityViewProvider } from '~/context/EntityView';
+import { EntityViewProvider } from '../src/context/EntityView';
```
eemeli: Dropping the ~ prefix from local paths ought not be necessary, see for example [this StackOverflow answer](https://stackoverflow.com/questions/72468249/vitest-src-folder-alias-not-resolved-in-test-files) for their configuration in Vitest.

Raj-G07:
@eemeli **Vitest** wasn’t able to resolve the `~` alias for local paths in this setup, so I switched to using relative paths `../` to make sure imports work correctly.

eemeli (changes requested):
This is mixing together a few too many things, which makes it a bit hard to review (or to eventually approve). It would probably be best to split the work into discrete chunks, such as:
1. Renaming test files as .jsx where appropriate.
2. Porting tests from Enzyme to testing-library.
3. Switching from Jest to Vitest.

Each of those ought to be a separate PR, and the test-porting can itself be split into multiple separate PRs, as we do have quite a few tests.

Non-test code should not need to change as a result of any of this.

Also, it might be a good idea to configure your editor to use Prettier to format Pontoon files when you save them? One of the CI tasks will complain if it notices code that doesn't match the expected formatting.

PR #3775, opened 2025-09-10
Add entity and translation search API endpoints

functionzz:
Introduces serializers and views for listing and retrieving entities, as well as searching for entity translations. Updates API URLs to include new endpoints for entities and translation search.


pontoon/api/views.py
```diff
@@ -361,3 +384,47 @@ def get_queryset(self):
...
+
+        data = {}
+
+        locale = get_object_or_404(Locale, code=locale_code)
+
+        # maybe we should check this, although if left empty/missing
+        # it doesn't matter inside for_project_locale
+        data.update({"search": search})
+
+        # locale is required
```
mathjazz: This comment seems out of place and unnecessary.

```suggestion
```

pontoon/api/views.py
```diff
@@ -361,3 +384,47 @@ def get_queryset(self):
...
+        search_identifiers = query_params.get("search_identifiers")
+        search_match_case = query_params.get("search_match_case")
+        search_match_whole_word = query_params.get("search_match_whole_word")
+
+        data = {}
+
+        locale = get_object_or_404(Locale, code=locale_code)
+
+        # maybe we should check this, although if left empty/missing
+        # it doesn't matter inside for_project_locale
```
mathjazz: Please use GitHub comments for this type of remarks.

```suggestion
```

pontoon/api/views.py
```diff
@@ -361,3 +384,47 @@ def get_queryset(self):
...
         queryset = super().filter_queryset(queryset)
         return queryset.prefetch_related("project", "locale")
+
+
+class TranslationSearchListView(generics.ListAPIView):
+    serializer_class = EntitySerializer
+
+    def get_queryset(self):
+        query_params = self.request.query_params
+
```
mathjazz, line 434: Can we [use `GetEntitiesForm`](https://github.com/mozilla/pontoon/blob/39f1a92b5c61fef074614a01ceb7b967da89cac9/pontoon/base/views.py#L237-L247) for validation?

pontoon/api/serializers.py
```diff
@@ -261,3 +267,92 @@ def get_project(self, obj):
...
+        for repository in repositories:
+            if repository.source_repo:
+                return repository.url
+        return repositories.first().url
+
+
+class EntitySerializer(serializers.ModelSerializer):
+    project = serializers.SerializerMethodField()
+    resource = MiniResourceSerializer(read_only=True)
+    translations = MiniTranslationSerializer(
```
mathjazz: Why do we use the prefix `Mini` here? I'd understand it if `ResourceSerializer` and `TranslationSerializer` were already taken.

pontoon/api/serializers.py
```diff
@@ -261,3 +267,92 @@ def get_project(self, obj):
...
+
+        return {
+            "code": obj.locale.code,
+            "name": obj.locale.name,
+        }
+
+    def get_editor_url(self, obj):
+        entity = Entity.objects.get(pk=obj.entity.pk)
+
+        return f"{os.environ.get('SITE_URL')}/{obj.locale.code}/{entity.resource.project.slug}/{entity.resource.path}/?string={entity.pk}"
```
mathjazz: Rather than hardcoding the path, we should use [`reverse()`](https://docs.djangoproject.com/en/5.2/ref/urlresolvers/#reverse).

pontoon/api/serializers.py
```diff
@@ -261,3 +267,92 @@ def get_project(self, obj):
...
+    def get_repository_url(self, obj):
+        repositories = Repository.objects.filter(project=obj.project)
+
+        if not repositories:
+            return None
+
+        for repository in repositories:
+            if repository.source_repo:
+                return repository.url
+        return repositories.first().url
```
mathjazz: The idea is to construct the full URL to the resource in the repository.

pontoon/api/serializers.py
functionzz: Yeah I changed it to Compact, because Mini is unique. I'm also making some changes so these might not exist anymore.

pontoon/api/views.py
functionzz: How does API access work without a form? I thought we were considering making it accessible directly as a public endpoint,?

mathjazz: You can use a Django form to validate GET parameters in a Django Rest Framework (DRF) endpoint without an HTML form. Django forms are versatile and can validate any data, including query parameters from a GET request.

functionzz: Oh cool I'll use that

pontoon/api/serializers.py
functionzz: Question: I don't think the base url (os.environ.get('SITE_URL')) works with reverse() as the base_url isn't considered as a view? The path `http://localhost:8000/ab/common-voice/web/locales/common-voice/en/pages/common.ftl/?string=312550` for example doesn't include any url patterns

pontoon/api/serializers.py
functionzz: Are you referring to the presence of things such as `git@github.com:common-voice/common-voice.git`?

pontoon/api/serializers.py
functionzz: fixed

mathjazz: It's still the same?

functionzz: it should be fixed now - check files changed

mathjazz:
Is this ready for another look? Is there are reason why is it marked as Draft?

pontoon/api/serializers.py
functionzz: this should be resolved now

mathjazz (commented):
Great work!

1. Why did we drop the resource path?

```python
{
    "resource": {
        "path": PATH,
    }
}
```

2. Please add tests for the new views (also check the DB query count).

3. Please make the `project` parameter optional in translation search and use it instead of the `all-projects` value.

4. Please use `text` parameter instead of `search` in translation search.

5. I wonder if we should move `id`, `string` and `key` keys in the translation search response under an `entity` key?

pontoon/api/views.py
```diff
@@ -361,3 +387,51 @@ def get_queryset(self):
...
+
+class TranslationSearchListView(generics.ListAPIView):
+    serializer_class = EntitySearchSerializer
+
+    def get_queryset(self):
+        query_params = self.request.query_params.copy()
+        if "text" in query_params:
+            query_params["search"] = query_params.get("text")
+            query_params.pop("text")
```
mathjazz:
```suggestion
            query_params["search"] = query_params.pop("text")
```

pontoon/api/views.py
```diff
@@ -361,3 +387,51 @@ def get_queryset(self):
...
+
+        query_params["project"] = query_params.get("project") or "all-projects"
```
mathjazz:
```suggestion
        query_params["project"] = query_params.get("project", "all-projects")
```

pontoon/api/serializers.py
```diff
@@ -261,3 +265,122 @@ def get_project(self, obj):
         if obj.project:
             return obj.project.slug
         return None
+
+
+class CompactTranslationSerializer(serializers.ModelSerializer):
```
mathjazz: Why not simply `TranslationSerializer`?

pontoon/api/serializers.py
```diff
@@ -261,3 +265,122 @@ def get_project(self, obj):
...
+                    "project": obj.entity.resource.project.slug,
+                    "resource": obj.entity.resource.path,
+                },
+                request=request,
+            )
+            + f"?string={obj.entity.pk}"
+        )
+
+
+class CompactResourceSerializer(serializers.ModelSerializer):
```
mathjazz: Same as above: Why not simply `ResourceSerializer`?

pontoon/api/tests/test_views.py
```diff
@@ -782,3 +902,304 @@ def test_tm_search(django_assert_num_queries):
             }
         ],
     }
+
+
+@pytest.mark.django_db
+def test_translation_search(django_assert_num_queries):
+    # this is gonna be long
```
mathjazz: Please remove.

pontoon/api/tests/test_views.py
```diff
@@ -782,3 +902,304 @@ def test_tm_search(django_assert_num_queries):
...
+    TranslationFactory.create(
+        entity=entities["entity_k"],
+        locale=locale_b,
+        string=f"translation_{locale_b.name}",
+    )
+    TranslationFactory.create(
+        entity=entities["entity_l"],
+        locale=locale_b,
+        string=f"translation_{locale_b.name}",
+    )
```
mathjazz: This could be a simple loop_:

```python
for entity in (entities["entity_c"], entities["entity_g"], entities["entity_h"]...)
...
```

pontoon/api/views.py
functionzz: in this case, `query_params.pop("text")` actually returns a list element and `query_params.get("text")` returns a string, which is kinda weird and why I placed it like that

mathjazz: Then why not use `query_params["search"] = query_params.pop("text")[0]`?

pontoon/api/views.py
```diff
@@ -361,3 +397,65 @@ def get_queryset(self):
...
         queryset = super().filter_queryset(queryset)
         return queryset.prefetch_related("project", "locale")
+
+
+class TranslationSearchListView(generics.ListAPIView):
+    serializer_class = EntitySearchSerializer
+
+    def get_queryset(self):
+        query_params = self.request.query_params.copy()
+        if "text" in query_params:
```
mathjazz: The text parameter should be required. The same is the case with terminology and tm search.

pontoon/api/tests/test_views.py
```diff
@@ -782,3 +902,485 @@ def test_tm_search(django_assert_num_queries):
...
+        response = APIClient().get(
+            "/api/v2/search/translations/?text=test",
+            HTTP_ACCEPT="application/json",
+        )
+
+    assert response.status_code == 400
+
+    assert response.data == {"locale": ["This field is required."]}
+
+    # text test, geonosian, match whole words
```
mathjazz: These commens are hard to comprehend. They should instead clarify what we're testing with each request. For example here, a helpful comment would be:

```suggestion
    # Test search_match_whole_words parameter
```

mathjazz:
> > We don't need to do that, but we should also test search without any search parameters and test each parameter separately.
> 
> I don't see this being addressed.

Still not addressed. We should test the minimum set of parameters, e.g.:
`/api/v2/search/translations/?locale=sl&text=open`

mathjazz:
> I guess its a good idea to decide whether we want to keep this swagger docs or not for the long term. What do you think @mathjazz?

Yeah, let's factor this comment out into a new issue where we can discuss this.

pontoon/api/views.py
```diff
@@ -361,3 +397,65 @@ def get_queryset(self):
...
         return queryset.prefetch_related("project", "locale")
+
+
+class TranslationSearchListView(generics.ListAPIView):
+    serializer_class = EntitySearchSerializer
+
+    def get_queryset(self):
+        query_params = self.request.query_params.copy()
+        if "text" not in query_params:
+            raise ValidationError({"text": ["This field is required."]})
```
mathjazz: Nice work! Could you please use the same `ValidationError` output in `TermSearchListView` and `TranslationMemorySearchListView`? Something like this:

```python
errors = {}
if not text:
    errors["text"] = ["This field is required."]
if not locale:
    errors["locale"] = ["This field is required."]
if errors:
    raise ValidationError(errors)
```

pontoon/api/tests/test_views.py
```diff
@@ -782,3 +902,547 @@ def test_tm_search(django_assert_num_queries):
...
+    ):
+        TranslationFactory.create(
+            entity=entity,
+            locale=locale_b,
+            string=f"translation_{locale_b.name}",
+        )
+
+    # Search parameters:
+    # - Locale: None
+    # - Text: None
```
mathjazz:
```suggestion
    # Test search without any parameters
```

pontoon/api/tests/test_views.py
```diff
@@ -782,3 +902,547 @@ def test_tm_search(django_assert_num_queries):
...
+            HTTP_ACCEPT="application/json",
+        )
+
+    assert response.status_code == 400
+
+    assert response.data == {"text": ["This field is required."]}
+
+    # Search parameters:
+    # - Locale: None
+    # - Text: test
```
mathjazz:
```suggestion
    # Test search without locale parameter
```

pontoon/api/tests/test_views.py
```diff
@@ -782,3 +902,547 @@ def test_tm_search(django_assert_num_queries):
...
+        )
+
+    assert response.status_code == 400
+
+    assert response.data == {"locale": ["This field is required."]}
+
+    # Search parameters:
+    # - Locale: Geonosian
+    # - Text: Flibbertigibbetelle
+    # - Projects: All Projects
```
mathjazz:
```suggestion
    # Test search with required parameters only
```

pontoon/api/tests/test_views.py
```diff
@@ -782,3 +902,547 @@ def test_tm_search(django_assert_num_queries):
...
+                "editor_url": f"http://testserver/gs/project-b/resource_project-b_3.ftl/?string={entities['entity_l'].id}",
+            },
+        }
+    ]
+
+    # Search parameters:
+    # - Locale: Geonosian
+    # - Text: Flibbertigibbet
+    # - Projects: All Projects
+    # - Match whole words: True
```
mathjazz:
```suggestion
    # Test search_match_whole_word parameter
```

pontoon/api/tests/test_views.py
```diff
@@ -782,3 +902,547 @@ def test_tm_search(django_assert_num_queries):
...
+                "editor_url": f"http://testserver/gs/project-b/resource_project-b_3.ftl/?string={entities['entity_k'].id}",
+            },
+        },
+    ]
+
+    # Search parameters:
+    # - Locale: Geonosian
+    # - Text: Dinglehopper
+    # - Projects: All Projects
+    # - Match case: True
```
mathjazz:
```suggestion
    # Test search_match_case parameter
```

pontoon/api/tests/test_views.py
```diff
@@ -782,3 +902,547 @@ def test_tm_search(django_assert_num_queries):
...
+                "editor_url": f"http://testserver/gs/project-b/resource_project-b_3.ftl/?string={entities['entity_l'].id}",
+            },
+        }
+    ]
+
+    # Search parameters:
+    # - Locale: Geonosian
+    # - Text: Dinglehopper
+    # - Projects: All Projects
+    # - Search identifiers: True
```
mathjazz:
```suggestion
    # Test search_identifiers parameter
```

pontoon/api/tests/test_views.py
```diff
@@ -782,3 +902,547 @@ def test_tm_search(django_assert_num_queries):
...
+            },
+        },
+    ]
+
+    # Search parameters:
+    # - Locale: Geonosian
+    # - Text: the%20Test
+    # - Projects: Project A
+    # - Match whole words: True
+    # - Match case: True
```
mathjazz:
```suggestion
    # Search multiple parameters
```

pontoon/api/tests/test_views.py
```diff
@@ -782,3 +902,547 @@ def test_tm_search(django_assert_num_queries):
...
+                "editor_url": f"http://testserver/gs/project-a/resource_project-a_2.ini/?string={entities['entity_f'].id}",
+            },
+        },
+    ]
+
+    # Search parameters:
+    # - Locale: Geonosian
+    # - Text: quibb%20Key_C
+    # - Projects: Project A
+    # - Search identifiers: True
```
mathjazz: What are we testing here that is not already tested with previous test cases?

pontoon/api/tests/test_views.py
```diff
@@ -782,3 +902,547 @@ def test_tm_search(django_assert_num_queries):
...
+                "editor_url": f"http://testserver/gs/project-a/resource_project-a_1.po/?string={entities['entity_c'].id}",
+            },
+        },
+    ]
+
+    # Search parameters:
+    # - Locale: Klingon
+    # - Text: "Test Flibbertigibbet"
+    # - Projects: All
+    # - Match case: True
```
mathjazz: What are we testing here that is not already tested with previous test cases?

pontoon/api/tests/test_views.py
```diff
@@ -782,3 +902,547 @@ def test_tm_search(django_assert_num_queries):
...
+                "editor_url": f"http://testserver/kg/project-b/resource_project-b_3.ftl/?string={entities['entity_l'].id}",
+            },
+        },
+    ]
+
+    # Search parameters:
+    # - Locale: Klingon
+    # - Text: dinglehopper
+    # - Search identifiers: True
+    # - Match whole word: True
```
mathjazz: What are we testing here that is not already tested with previous test cases?

mathjazz (commented):
Looks like we had a misunderstanding in how the comments should be changed. :)

pontoon/api/tests/test_views.py
functionzz: we are testing the "each word" default functionality with search identifiers. So a word that exists inside the key and another word that exists inside the string, is that an entity that is displayed from this endpoint for example.

pontoon/api/tests/test_views.py
functionzz: We are testing the quotation mark "exact" word search so "Test Flibbertigibbet", not Test%20Flibbertigibbet

pontoon/api/tests/test_views.py
functionzz: This is another random multiple query parameter test. I guess it can be removed

pontoon/api/views.py
```diff
@@ -350,14 +388,86 @@ def get_queryset(self):
...
         return queryset.prefetch_related("project", "locale")
+
+
+class TranslationSearchListView(generics.ListAPIView):
+    serializer_class = EntitySearchSerializer
+
+    def get_queryset(self):
+        query_params = self.request.query_params.copy()
+        text = query_params.get("text")
+        code = query_params.get("locale")
```
mathjazz: Why is this variable not called `locale`, like everywhere else?

pontoon/api/tests/test_views.py
```diff
@@ -782,3 +902,526 @@ def test_tm_search(django_assert_num_queries):
...
+            "resource": {"path": "resource_project-b_3.ftl"},
+            "translation": {
+                "locale": {"code": "gs", "name": "Geonosian"},
+                "string": "translation_Geonosian",
+                "editor_url": f"http://testserver/gs/project-b/resource_project-b_3.ftl/?string={entities['entity_l'].id}",
+            },
+        },
+    ]
+
+    # Search multiple parameters
```
mathjazz:
```suggestion
    # Test search with multiple parameters
```

pontoon/api/tests/test_views.py
mathjazz: That means we are essentially testing `Entity.for_project_locale()`, which is not in scope of this PR and should not be tested in the `pontoon.api` module. I'd drop this test.

pontoon/api/tests/test_views.py
mathjazz: That means we are essentially testing `Entity.for_project_locale()`, which is not in scope of this PR and should not be tested in the `pontoon.api` module. I'd drop this test.

pontoon/api/tests/test_views.py
mathjazz: Yes, let's remove it.

pontoon/api/tests/test_views.py
```diff
@@ -782,3 +902,394 @@ def test_tm_search(django_assert_num_queries):
...
+        entities["entity_k"],
+        entities["entity_l"],
+    ):
+        TranslationFactory.create(
+            entity=entity,
+            locale=locale_b,
+            string=f"translation_{locale_b.name}",
+        )
+
+    # Test translation search without any parameters
```
mathjazz: Nit: 

```suggestion
    # Test search without any parameters
```

mathjazz (commented):
Looks great! Just a tiny nit (my bad!) and we can deploy to stage.

pontoon/api/views.py
```diff
@@ -350,14 +388,86 @@ def get_queryset(self):
...
+        form_data = {
+            k: form.cleaned_data[k] for k in restrict_to_keys if k in form.cleaned_data
+        }
+
+        try:
+            entities = (
+                Entity.for_project_locale(
+                    self.request.user, project, locale, **form_data
+                )
+                .filter(translation__locale=locale)
```
mathjazz: Why is that filter necessary? `Entity.for_project_locale` already takes `locale` as an argument and should only return translations of the given locale.

mathjazz (commented):
I just realized that https://mozilla-pontoon-staging.herokuapp.com/api/v2/entities/66617/?include_translations also returns suggestions. It should only return approved translations.

Similarly, https://mozilla-pontoon-staging.herokuapp.com/api/v2/search/translations/?locale=sl&project=common-voice&text=dataset also returns suggestions, if they are active translations (e.g. the only submitted translation for the entity-locale combo). We should only return approved translations.

pontoon/api/views.py
functionzz: you are right

mathjazz:
Just wanted to stress that we should only return *approved* translations. Not *active*.

pontoon/api/tests/test_views.py
```diff
@@ -590,6 +592,134 @@ def test_disabled_projects(
...
+    ]
+
+    resource = ResourceFactory.create(
+        project=project_a, path=f"resource_{project_a.slug}.po", format="po"
+    )
+
+    entity = EntityFactory.create(string="Test String", resource=resource)
+
+    for locale in locales:
+        (
```
mathjazz, line 644: Why the bracket?

pontoon/api/views.py
```diff
@@ -291,6 +297,45 @@ def get_queryset(self):
...
+            Prefetch(
+                "translation_set",
+                queryset=Translation.objects.filter(approved=True).select_related(
+                    "locale"
+                ),
+                to_attr="filtered_translations",
+            ),
+            "resource",
+            "resource__project",
+            "translation_set__locale",
```
mathjazz: If I remove these, the number of DB queries stays the same (actually goes down slightly).

functionzz: is that so? Lemme check

pontoon/api/views.py
```diff
@@ -319,9 +319,9 @@ def get_queryset(self):
                 ),
                 to_attr="filtered_translations",
             ),
-            "resource",
-            "resource__project",
-            "translation_set__locale",
+            # "resource",
+            # "resource__project",
+            # "translation_set__locale",
```
mathjazz: Why comment out, rather than remove?

mathjazz (commented):
The first comment was marked as resolved, but it's not:
https://github.com/mozilla/pontoon/pull/3775#discussion_r2380213134

flodolo:
This is not what I expect from this API

```
                "string": "  | <p>In the spirit of experimentation, we're using a forum service\n  | that requires an account different from the one you\n  | created for Test Pilot. We are working hard to change this\n  | and make the forum experience smoother.</p>\n  | <p>If you don't want to create another account, you can\n  | always leave feedback through Test Pilot.\n  | <br>\n  | (We really do read this stuff)</p>\n",

                "string": "  [html/title] Active Users of this Experiment\n",

                "string": "Get the add-on to get going!\n",
```

I expect to get what's in the file, or displayed in the Pontoon editor, not the internal representation.

Also, shouldn't we exclude disabled projects? These are all from Test Pilot, which is not enabled.

flodolo:
The ID and newline are just a part of the problem. Look at the first two strings: I doubt `[html/title]` or `|` are part of the string (I can't confirm since the project is disabled).

> I don't understand what you mean by "This feels very long.". The API response (because editor_url is repeated so many times)?

Yes. I would also limit the locale data to the code (no name), but I feel like that might complicate things unnecessarily.

mathjazz:
> The ID and newline are just a part of the problem. Look at the first two strings: I doubt `[html/title]` or `|` are part of the string (I can't confirm since the project is disabled).

They were part of the old syntax:
https://mozilla-pontoon-staging.herokuapp.com/sl/firefox/all-resources/?string=153338
https://mozilla-pontoon-staging.herokuapp.com/sl/firefox/all-resources/?string=153335

> Yes. I would also limit the locale data to the code (no name), but I feel like that might complicate things unnecessarily.

Should we drop `editor_url` entirely for now? I only suggested we add it, because I saw it's included in the Transvision UI, but we can generate it from the rest of the data in the API response.

flodolo:
> They were part of the old syntax:

Wow, completely removed that from my brain.

> Should we drop editor_url entirely for now?

Works for me.

mathjazz:
OK, so for now we have at least the follwing action items:
- drop `editor_url` (sorry, @functionzz!)
- remove trailing newlines from `Entity.string` and `Translation.string`

flodolo:
> nstead, we should be providing the `.value` and `.properties` of each Entity and Translation, with JSON-ified moz.l10n message data model values.

Can you provide examples of what you mean?

eemeli:
The format-specific test cases show some examples of these values:
- https://github.com/mozilla/pontoon/blob/main/pontoon/sync/tests/formats/test_ftl.py
- https://github.com/mozilla/pontoon/blob/main/pontoon/sync/tests/formats/test_gettext.py
- and others next to the above

The Django admin view will also show these values for each Entity. The JSON format itself is described by https://github.com/mozilla/moz-l10n/blob/main/schemas/message.json.

eemeli:
> We need the API to provide text that can be used directly without the user having to know how to deal with the internal representation, i.e. `{ $promotionName } discount will be applied`.

Used directly for what? Presentation to a human as an indication of the message's actual contents, or for some purpose that needs an accurate representation of the message?

For a specific example of what we need to account for, consider e.g. [this translation](https://github.com/mozilla-l10n/sumo-l10n/blob/ed91b45b91afdb94ed6289d386af475bfbce3fb0/sl/LC_MESSAGES/djangojs.po#L762-L768) of a Gettext message:
```po
#, python-format
msgid "1 person has this problem"
msgid_plural "%s people have this problem"
msgstr[0] "1 oseba ima to težavo"
msgstr[1] "%s osebi imata to težavo"
msgstr[2] "%s ljudje imajo to težavo"
msgstr[3] "%s ljudi ima to težavo"
```

How should we represent that when it's requested via the API? I think our options are:
1. The above .po serialised representation.
2. A simplified single-pattern "preview": `'%s ljudi ima to težavo'`
3. The moz.l10n data model representation:
   ```js
   {
     decl: { n: { '$': 'n', fn: 'number' } },
     sel: [ 'n' ],
     alt: [
       { keys: [ 'one' ], pat: [ '1 oseba ima to težavo' ] },
       { keys: [ 'two' ], pat: [ '%s osebi imata to težavo' ] },
       { keys: [ 'few' ], pat: [ '%s ljudje imajo to težavo' ] },
       { keys: [{ '*': 'other' }], pat: [ '%s ljudi ima to težavo' ] }
   }
   ```
4. Some other representation, which is custom-built just for this API.

Note that we can pick multiple choices here, so we could e.g. include both (2.) a simplified preview as well as (3.) the actual message value in the response.

flodolo:
> One reason I'm pushing back here is that this approach is adding a new development burden (the plural-supporting entry parsers & serialisers) that becomes a blocker for Pontoon's Android plural support, and for which I'm not aware of us having a client yet, or even an explicit use case.

I'm not sure why that would be the case. For now we don't need to find a solution for Android or iOS plurals, since we don't have any. But we should find a solution for Fluent complex messages and Gettext plurals, if we want the API to be usable.

> Would our current use cases be satisfied by a single-pattern "simple preview" representation of each entity and translation? If so, that would allow for a more complete message representation to be added in a later step, once we can define a use case for it.

Going forward, could you provide practical examples? What does simple "simple preview" mean for strings across different formats?

mathjazz:
Looks good to me to. The only thing that remains is the alternative URL for the entity endpoint as discussed on Slack.

pontoon/api/views.py
```diff
@@ -291,6 +297,74 @@ def get_queryset(self):
...
+            if not entity:
+                errors["entity"] = ["This field is required."]
+            if errors:
+                raise ValidationError(errors)
+
+        queryset = queryset.filter(
+            resource__project__slug=project,
+            resource__path=resource,
+            key__overlap=[entity],
+        )
```
mathjazz: Rather than validate, I would ignore query parameters if not all of them are given.

Something like that should work in this case:

```suggestion
        if (project and resource and entity):
            queryset = queryset.filter(
                resource__project__slug=project,
                resource__path=resource,
                key__overlap=[entity],
            )

        return queryset.prefetch_related(
            "resource",
            "resource__project",
        )
```

functionzz: I disagree with this because if you for example misspell the project, resource or entity flags you are given 100,000 entities, which may confuse the user, given its already the default behavior.

pontoon/api/views.py
```diff
@@ -291,6 +297,65 @@ def get_queryset(self):
...
+            )
+
+        return queryset.prefetch_related(
+            Prefetch(
+                "translation_set",
+                queryset=Translation.objects.filter(approved=True).select_related(
+                    "locale"
+                ),
+                to_attr="filtered_translations",
+            ),
```
mathjazz: We should only be prefetching translations when relevant, i.e. when `include_translations` param is given. And as per the `NestedEntityListSerializer` definition, also project, resource and entity params need to be set.

flodolo:
Could you create a list of translations affected by the change in escaping? I expect those to be mistakes and not done on purpose, and that would help understanding the need for communications.

pontoon/api/views.py
functionzz: > Can we instead just handle the lookup in EntityListView and then redirect to EntityIndividualView?

The main problem with this which I am finding out is that ListAPIView returns a list format, while RetrieveAPIView, of which EntityIndividualView returns an object. Unless I override methods which display the data in the default format, I don't see how this can work out.

Why can't we just have 2 urls `/entities/{id}` and `/entities/{project}/{resource}/{entity}/`? We can use the same EntityIndividualView just with conditions, and don't have to worry about converting objects into lists for display?

mathjazz: > Why can't we just have 2 urls `/entities/{id}` and `/entities/{project}/{resource}/{entity}/`? We can use the same EntityIndividualView just with conditions, and don't have to worry about converting objects into lists for display?

Good call. Let's go ahead with this.

pontoon/api/urls.py
```diff
@@ -41,7 +41,12 @@
     path(
         "entities/<int:pk>/",
         views.EntityIndividualView.as_view(),
-        name="entity-individual",
+        name="entity-individual-id",
+    ),
+    path(
+        "entities/<slug:project_slug>/<path:resource>/<str:key>/",
```
mathjazz: For consistency, this should be either:
* `entities/<slug:project>/<path:resource>/<str:entity>/`
* `entities/<slug:slug>/<path:path>/<str:key>/`
* `entities/<slug:project_slug>/<path:resource_path>/<str:entity_key>/`

mathjazz (approved):
Excellent work! Let's go!

functionzz:
Hooray!

PR #3778, opened 2025-09-11
add permission check to token deletion

functionzz:
This PR addresses [https://bugzilla.mozilla.org/show_bug.cgi?id=1987028](url).

### DIFF
--- a/pontoon/contributors/views.py
+++ b/pontoon/contributors/views.py
@@ -508,6 +508,16 @@ def generate_token(request):
 def delete_token(request, token_id):
     try:
         token = PersonalAccessToken.objects.get(id=token_id)
+
+        if token.user != request.user:
+            return JsonResponse(
+                {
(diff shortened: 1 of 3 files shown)


pontoon/contributors/views.py
```diff
@@ -519,7 +529,7 @@ def delete_token(request, token_id):
 
     except PersonalAccessToken.DoesNotExist:
         return JsonResponse(
-            {"status": "error", "message": "Token not found."}, status=404
+            {"status": "error", "message": "Unable to delete the token."}, status=404
```
mathjazz, line 532: What's the purpose of this change?

functionzz: I thought it would be better since "Token not found." is more specific and if I left the message vague as in "Unable to delete the token.", a malicious user may not be able to tell which ids of tokens actually exist

PR #3781, opened 2025-09-15
Find plural categories based on language tag only

### LINKED ISSUE #3343: Special case plural handling should be on the primary subtag
Because of how #2780 got fixed, if someone makes a new language like `es-419`, it will again add plural for many and its the same for any other language.
I don't really understand why this implementation of the fix checks for specific language tags instead of checking if the primary language is the expected one if this is a bug. 
I also cant find the original reason as they are kinda correct plural rules, what happens if I use fluent.js? Does it have somewhere hardcoded that `es` is just `['one', 'other']`?

eemeli:
~There might be an issue filed about this, but I can't find it.~

Fix #3343.

With this change, adding a new variant of a language with customization in `getPluralCategories()` won't need to have that function updated to account for it.

### DIFF
--- a/translate/src/utils/message/getEmptyMessage.test.js
+++ b/translate/src/utils/message/getEmptyMessage.test.js
@@ -168,7 +168,31 @@ describe('getEmptyMessage', () => {
       `);
   });
 
-  it('handles plural messages for special locales', () => {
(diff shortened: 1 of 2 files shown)


mathjazz:
Added the related issue above.

PR #3782, opened 2025-09-15
Ensure all Fluent strings end in a newline

### LINKED ISSUE #2216: Fluent translations stored in database without the trailing newline

eemeli:
Fixes #2216

### DIFF
--- a/pontoon/base/migrations/0099_add_trailing_fluent_newlines.py
+++ b/pontoon/base/migrations/0099_add_trailing_fluent_newlines.py
@@ -1,0 +1,32 @@
+from django.db import migrations
+
+
+def add_trailing_fluent_newlines(apps, schema_editor):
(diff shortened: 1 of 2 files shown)


mathjazz (approved):
Tested alongside #3782 and #3783 on stage and it seems fine.

PR #3783, opened 2025-09-15
Enable context-sensitive autocompletion for placeholders

eemeli:
The original string is scanned with the same rules as we already use for its highlighting. If any highlights are detected, appropriate autocompletions are enabled; these will pop up automatically, or can be manually shown with Ctrl+Esc. A selected completion is accepted with Enter or Tab when the popup is visible, and it's dismissed with Esc or by focus change.

Automated tests are challenging for the editor, so this should be deployed to stage for manual testing when appropriate.

### DIFF
--- a/translate/src/modules/translationform/utils/editFieldExtensions.ts
+++ b/translate/src/modules/translationform/utils/editFieldExtensions.ts
@@ -1,4 +1,10 @@
-import { closeBrackets, closeBracketsKeymap } from '@codemirror/autocomplete';
+import {
+  acceptCompletion,
+  autocompletion,
+  closeBrackets,
+  closeBracketsKeymap,
+  type CompletionSource,
+} from '@codemirror/autocomplete';
 import {
   history,
   historyKeymap,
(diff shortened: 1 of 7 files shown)


mathjazz:
Nice work, I played with it a little and I'm starting to like it. :)

Here's a couple of suggestions:
* When autocompleting tags, first suggestions is almost always wrong - it's a closing tag. I assume we order in alphabetic order. We should be a tad smarter about that.
* Margin on the left side of the autocompletion popup should be removed.
* We should make the popup styling (colors) configurable via CSS themes.

Bonus:
* We should highlight syntax in the popup.

Screenshot 2025-09-16 at 00 20 51

translate/src/modules/translationform/utils/editFieldExtensions.ts
```diff
@@ -112,3 +124,77 @@ export const getExtensions = (
...
+function completePlaceholder(
+  orig: string,
+  start: number,
+  ends: number[],
+): CompletionSource {
+  const esc = (chars: string) => chars.replace(/[^\w\s]/g, '\\$&');
+  const rest = new Set(orig.substring(start + 1, ends[0]));
+  const restChars = Array.from(rest).join('');
+  const match = new RegExp(`[${esc(orig[start])}][${esc(restChars)}]*$`);
+  const boost = orig.substring(start, start + 2) === '</' ? -1 : 0;
```
mathjazz, line 196: It would be nice to add a comment explaining why we're doing this.

mathjazz (approved):
This looks good to me now. Please let me know (or make changes) if you disagree with the CSS modifications, otherwise feel free to merge.

translate/src/modules/translationform/utils/editFieldExtensions.ts
eemeli:
```suggestion
  // Deprioritize closing tags, as they otherwise show up first in the list.
  const boost = orig.substring(start, start + 2) === '</' ? -1 : 0;
```

mathjazz:
Sorry, but I'd keep the autocomplete menu dark in the dark theme, because all popups are dark.

Going forward, I think we should make the editor also dark. Potentially with an option to change the editor theme independently from the rest of the UI.

Screenshot 2025-09-16 at 21 39 22

mathjazz (approved):
Looks good!

PR #3784, opened 2025-09-16
Note on localization

mathjazz:
Added a readme note as discussed.

### DIFF
--- a/translate/README.md
+++ b/translate/README.md
@@ -233,7 +233,7 @@ You can read [more about pseudo-localization on Wikipedia](https://en.wikipedia.
 
 ### Localizing Pontoon
 
-At this stage, only the translation workspace is localizable. Until the full application becomes localizable, Pontoon will not be available for localization on [pontoon.mozilla.org](https://pontoon.mozilla.org) and we will not be accepting translations for it.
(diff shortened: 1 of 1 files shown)


translate/README.md
```diff
@@ -231,6 +231,10 @@ Pseudo-localization turns every supported string into a different version of its
 
 You can read [more about pseudo-localization on Wikipedia](https://en.wikipedia.org/wiki/Pseudolocalization).
 
+### Localizing Pontoon
+
+At this stage, only the translation workspace is localizable. Until the full application becomes localizable, Pontoon will not be available for localization on [pontoon.mozilla.org](https://pontoon.mozilla.org) and we will not be accepting translations for it.
```
flodolo:
```suggestion
At this stage, only the translation workspace is localizable. Until the full application becomes localizable and we can provide an option for users to select their language, Pontoon will not be available for localization on [pontoon.mozilla.org](https://pontoon.mozilla.org) and we will not be accepting translations updates through pull requests.
```

flodolo (changes requested):
Can you put a README also in https://github.com/mozilla/pontoon/tree/main/translate/public/locale? I expect most people to look there.

translate/README.md
mathjazz:
```suggestion
At this stage, only the translation workspace is localizable. Until the full application becomes localizable and users can select their language, Pontoon will not be available for localization on [pontoon.mozilla.org](https://pontoon.mozilla.org) and we will not be accepting translation updates through pull requests.
```

PR #3788, opened 2025-09-19
Fix messaging center Icon in search field

### LINKED ISSUE #3762: Messaging Center: search field displayed incorrectly (icon out of place, text overlapping)
The icon is off before starting to type

Image

It gets worse when you start typing

Image

functionzz:
This fixes #3762.

### DIFF
--- a/pontoon/base/static/css/check-box.css
+++ b/pontoon/base/static/css/check-box.css
@@ -33,9 +33,4 @@
   .check-box-wrapper:hover {
     color: var(--white-1);
   }
-
-  .fas {
-    margin-left: 23px;
-    margin-right: 0;
-  }
 }


mathjazz (commented):
Thanks for the patch. It fixes the problem reported in #3762, but breaks the check-box position e.g. in the /settings page.

Screenshot 2025-09-19 at 21 50 28

functionzz:
ohhhhhhh

pontoon/base/static/css/check-box.css
```diff
@@ -34,7 +34,7 @@
     color: var(--white-1);
   }
 
-  .fas {
+  .check-box .fas {
```
mathjazz: Thanks to [CSS nesting](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_nesting) we use more and more and in particular in this file, you can move this rule under `.check-box` in line 7.

mathjazz (commented):
Works well!

PR #3789, opened 2025-09-23
Refactor & fix insights data collection

### LINKED ISSUE #3656: Project and ProjectLocale Insights pages don't have data since January

eemeli:
Fixes #3656

The attached migration is quite slow for pontoon.mozilla.org data (as in, multiple minutes of runtime) due to needing to recalculate activity stats for about 1.4 million projectlocales.

### DIFF
--- a/pontoon/insights/tasks.py
+++ b/pontoon/insights/tasks.py
@@ -80,7 +80,7 @@ def count_activities(dt_max: datetime):
     `projectlocale_id -> Activity`
 
     Fetch and prepare activity data.
-    
+
     Note that this function is also called from
     pontoon.insights.migrations.0017_fix_projectlocale_insights_again,
     which may need a local copy if the behaviour here is modified.


pontoon/insights/migrations/0017_fix_projectlocale_insights_again.py
```diff
@@ -0,0 +1,91 @@
...
+    while start_date in nonempty_dates:
+        nonempty_dates.remove(start_date)
+        start_date += timedelta(days=1)
+    end_date = min(nonempty_dates) if nonempty_dates else date.today()
+
+    prev_date = None
+    for days in range((end_date - start_date).days):
+        d = start_date + timedelta(days=days)
+        dt = datetime.combine(d, time(), tzinfo=timezone.utc)
+        activities = count_activities(dt)
```
mathjazz, line 42: In Django data migrations, calling a function from the main codebase (which itself calls other functions) is risky because migrations are meant to be stable and reproducible over time. 

If the function in the main codebase changes in the future, it could break the migration or produce inconsistent results when re-running it on a fresh database or during rollbacks.

pontoon/insights/tasks.py
```diff
@@ -30,653 +27,585 @@
...
+        )
+        .annotate(
+            locale=F("projectlocale__locale"),
+            total=Sum("total_strings", default=0),
+            approved=Sum("approved_strings", default=0),
+            pretranslated=Sum("pretranslated_strings", default=0),
+            errors=Sum("strings_with_errors", default=0),
+            warnings=Sum("strings_with_warnings", default=0),
+            unreviewed=Sum("unreviewed_strings", default=0),
+        )
```
mathjazz, line 279: Why do we need the `FilteredRelation` here?

pontoon/insights/migrations/0017_fix_projectlocale_insights_again.py
eemeli: I agree in general. In this case that would mean copying about 200 lines of code, though, which seems like work that might be best done if/when `count_activities()` is changed in a way that would break this usage.

pontoon/insights/tasks.py
eemeli: Because just `F('resource__project__project_locale')` would include projectlocales for all locales that are enabled for the project.

pontoon/insights/migrations/0017_fix_projectlocale_insights_again.py
mathjazz: Then we should at least add a comment to `count_activities()` and the functions it calls that they are also used in the data migration.

pontoon/insights/tasks.py
mathjazz: Unless you include `resource__project__project_locale__locale=F("locale")` in `filter()`?

pontoon/insights/tasks.py
```diff
@@ -30,653 +27,585 @@
...
 
 
-def collect_project_locale_insights(start_of_today, activities, entities):
+def count_activities(dt_max: datetime):
     """
-    Collect insights for each available ProjectLocale.
+    `projectlocale_id -> Activity`
+
+    Fetch and prepare activity data.
     """
```
eemeli, line 87:
```suggestion
    
    Note that this function is also called from
    pontoon.insights.migrations.0017_fix_projectlocale_insights_again,
    which may need a local copy if the behaviour here is modified.
    """
```

pontoon/insights/migrations/0017_fix_projectlocale_insights_again.py
eemeli: Done: https://github.com/mozilla/pontoon/pull/3789#discussion_r2378009905.

mathjazz (commented):
Forgot to submit my other inline comment.

pontoon/insights/tasks.py
eemeli: Sure; that arguably does simplify the queries a bit.

mathjazz (approved):
Looks good, great work!

Issue #3759, opened 2025-08-18
Support Android strings plurals


eemeli:
The primary user-visibile benefits of the moz.l10n data model adoption (#3538) are that it:
1. enables plural support in all available formats,
2. allows linting/validation to be refactored, and
3. makes introducing new formats much easier.

Our initial focus is the first of these, and Android Strings plural support in particular. That requires the following related changes:

- [x] Parse and store plural Android messages as such.
  - [x] The capability is ready in moz.l10n, though not tested with real data.
  - [x] Pontoon needs to store Android strings in a way that supports plurals. This is partly done, as they're fine getting written in `Entity.value`, but we still have `Entity.string` hanging around. So, let's use the MF2 syntax as the `Entity.string` for Android like we already do for Gettext.
- [x] Support multi-variant translations for Android.
  - [ ] Add the `.value` and `.properties` fields to `Translation`, as they now are only on `Entity`.
  - [x] Switch `Translation.string` to use MF2 syntax for Android strings.
  - [ ] Either serialize the `.value` as `.string`, or parse the `.string` into the `.value`, and do this either in the frontend or the backend.
- [x] Use the parsed `.value` data to build the frontend's multi-field "rich" editor for plurals.
  - [ ] Add `.value` and `.properties` to the relevant API endpoints.
  - [x] The editor was already extended to support MF2 syntax for Gettext, and it now parses both it and Fluent into the moz.l10n data model, so reading the parsed data model values should be relatively straighforward.
- [x] Serialize plural translations
  - [x] The capability is ready in moz.l10n, though not tested with real Android data.
  - [x] To start, use the sync logic already used for Gettext for Android as well.
  - [ ] Longer-term, serialization ought to rely on `Translation.value` and `Translation.properties`, rather than `Translation.string`.

It's pretty likely that the work here will ultimately get split a bit differently than it's presented above.

One potentially tempting next step would be to change the Android string serializations in the database to use MF2 syntax. This would remove the hard dependency on the data model migration from blocking the nominaly feature development, and it would ensure that a plural message showing up would mostly "just work". However, this would also increase the prominence of the following regressions that probably should be addressed first:
- #3741
- #3742

[...]

PR #3790, opened 2025-09-24
Use MF2 syntax for Android

### LINKED ISSUE #3448: Android xml resource (strings.xml) containing tags aren't parsed correctly
[As described here](https://developer.android.com/guide/topics/resources/localization?hl=en) Android resources can contain xliff tags. For example :
```
<string name="shortcuts_additional_settings">
  <xliff:g id="vendor" example="Xiaomi">%1$s</xliff:g> additional settings
</string>
```

Unfortunately, when they do, the synchronized resource is not parsed correctly. With the example above, the text (appearing in the left column) will contain the whole string tag, and not only its content.

Capture d'écran 2024-11-19 004407

A workaround for a correct parsing would be to include content in CDATA, but this would then lead to Android studio not replacing %1$s with example data, as it does with the initial syntax.

Is there any way to make the xml parser not fail and output the whole tag instead of only its content as it normally should ?

eemeli:
This is an enabler for #3759; most of the support structures required for Android plurals are added here, but their functionality is not validated -- the focus instead is on making sure that existing features continue to work as intended.

Probably also fixes #3448, but that ought to be validated by e.g. @spocky, if they're available.

I've included some refactoring of "Pontoon checks", as their handling in two separate runs no longer really makes sense.


mathjazz (commented):
Thanks for this!

I've deployed the patch to stage and set up Pontoon Test again:
https://mozilla-pontoon-staging.herokuapp.com/sl/pontoon-test-2/

So far I haven't found any issues. It'd be good if other folks take a look, too.

I wouldn't make all these changes to the checks, because we'll be getting rid of this code anyways, but it is what it is now.

Are you planning to add more checks to cover these changes:
https://mozilla.slack.com/archives/C06KGJ6PEDV/p1757949007724179

eemeli:
> I wouldn't make all these changes to the checks, because we'll be getting rid of this code anyways, but it is what it is now.
> 
> Are you planning to add more checks to cover these changes: https://mozilla.slack.com/archives/C06KGJ6PEDV/p1757949007724179

Inlining for context:
> The changes I'm working on will include some changes to how translators work with Android strings:
> - Translations should not include any Android escapes of any sort; that's all taken care of internally.
> - Placeholders in translations must match the source exactly, i.e. if there's a `<b>` in the source, having `<B>` in the target is problematic.

Both of the above appear to be already caught at least as warnings. In practice, at least the specific `<b>`/`<B>` should not cause a runtime problem as both of those get parsed equivalently by the Android formatter. I am not currently planning on adding further checks; as you mention, the plan is to replace all of the current ones with a new implementation.

mathjazz:
> Both of the above appear to be already caught at least as warnings.

They don't seem to be caught by anything else but Translate Toolkit checks, which can be disabled. 590 users have these checks disabled.

mathjazz:
> They don't seem to be caught by anything else but Translate Toolkit checks, which can be disabled. 590 users have these checks disabled.

@flodolo Do you think that's sufficient for the time being?

flodolo:
> > Both of the above appear to be already caught at least as warnings.
> 
> They don't seem to be caught by anything else but Translate Toolkit checks, which can be disabled. 590 users have these checks disabled.

> The changes I'm working on will include some changes to how translators work with Android strings

What's the timeline for these changes? What happens to broken translations submitted in the meantime? I think [moz-l10n-linter](https://github.com/mozilla-l10n/mozl10n-linter) should already catch these.

eemeli:
> > The changes I'm working on will include some changes to how translators work with Android strings
> 
> What's the timeline for these changes? What happens to broken translations submitted in the meantime? I think [moz-l10n-linter](https://github.com/mozilla-l10n/mozl10n-linter) should already catch these.

These changes refer effectively to this PR. The linter/validator fixes are one of our planned next steps of the data model refactor, so they're likely to land within some number of months.

The nominal danger here is that a translation is accepted with e.g. an explicit `" foo"` escaping a leading or trailing space, which would then get stored and then rendered with the quotes included in the output. For HTML elements and placeholders, the handling of broken messages will stay the same as it currently is.

flodolo:
Could you create a list of translations affected by the change in escaping? I expect those to be mistakes and not done on purpose, and that would help understanding the need for communications.

eemeli:
> > Both of the above appear to be already caught at least as warnings.
> 
> They don't seem to be caught by anything else but Translate Toolkit checks, which can be disabled. 590 users have these checks disabled.

@mathjazz With the most recent commit, mismatches in printf-style formatters and HTML tags should now be caught as errors by the custom checks, not just translate-toolkit.

mathjazz:
In https://mozilla-pontoon-staging.herokuapp.com/sl/focus-for-android/all-resources/?string=224023, I replaced the first `<p>` with `<pj>`, which triggerred the following server error when I tried to save:

```
Message: XMLSyntaxError: Opening and ending tag mismatch: pj line 1 and p, line 1, column 46 (<string>, line 1)

Traceback:

File "/app/.heroku/python/lib/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response

File "/app/.heroku/python/lib/python3.11/site-packages/newrelic/hooks/framework_django.py", line 529, in wrapper
                  return wrapped(*args, **kwargs)

File "/app/.heroku/python/lib/python3.11/site-packages/django/views/decorators/http.py", line 43, in inner
              return func(request, *args, **kwargs)

File "/app/pontoon/base/utils.py", line 70, in wrap
          return f(request, *args, **kwargs)

File "/app/.heroku/python/lib/python3.11/site-packages/django/contrib/auth/decorators.py", line 23, in _wrapper_view
                  return view_func(request, *args, **kwargs)

File "/app/.heroku/python/lib/python3.11/contextlib.py", line 81, in inner
                  return func(*args, **kwds)

File "/app/pontoon/translations/views.py", line 95, in create_translation
          failed_checks = run_checks(

File "/app/pontoon/checks/libraries/__init__.py", line 31, in run_checks
      checks.update(run_custom_checks(entity, original, string))

File "/app/pontoon/checks/libraries/custom.py", line 99, in run_custom_checks
                  android_msg = android_parse_message(android_simple_preview(pattern))

File "/app/.heroku/python/lib/python3.11/site-packages/moz/l10n/formats/android/parse.py", line 241, in android_parse_message
              raise err

File "/app/.heroku/python/lib/python3.11/site-packages/moz/l10n/formats/android/parse.py", line 232, in android_parse_message
              el = etree.fromstring(f"{doctype}{source}", parser)

File "src/lxml/etree.pyx", line 3306, in lxml.etree.fromstring
  null
File "src/lxml/parser.pxi", line 1995, in lxml.etree._parseMemoryDocument
  null
File "src/lxml/parser.pxi", line 1875, in lxml.etree._parseDoc
  null
File "src/lxml/parser.pxi", line 1105, in lxml.etree._BaseParser._parseUnicodeDoc
  null
File "src/lxml/parser.pxi", line 633, in lxml.etree._ParserContext._handleParseResultDoc
  null
File "src/lxml/parser.pxi", line 743, in lxml.etree._handleParseResult
  null
File "src/lxml/parser.pxi", line 672, in lxml.etree._raiseParseError
  null
```

mathjazz:
In https://mozilla-pontoon-staging.herokuapp.com/sl/focus-for-android/all-resources/?string=223878, I thought I wouldn't be able to submit the translation with \n, because _Translations should not include any Android escapes of any sort; that's all taken care of internally_.

eemeli:
> In https://mozilla-pontoon-staging.herokuapp.com/sl/focus-for-android/all-resources/?string=224023, I replaced the first `<p>` with `<pj>`, which triggerred the following server error when I tried to save:

Should work now; I was missing an HTML `escape()` wrapper.

> In https://mozilla-pontoon-staging.herokuapp.com/sl/focus-for-android/all-resources/?string=223878, I thought I wouldn't be able to submit the translation with \n, because _Translations should not include any Android escapes of any sort; that's all taken care of internally_.

Should we add a new, custom rule forbidding the character sequence `\n` from showing up in Android translations if it's not in the original?

mathjazz:
> Should we add a new, custom rule forbidding the character sequence `\n` from showing up in Android translations if it's not in the original?

If it's breaking build or causing issues on runtime, yes.
If it's just a cosmetic mistake when added (like in other file formats), no.

eemeli:
> > Should we add a new, custom rule forbidding the character sequence `\n` from showing up in Android translations if it's not in the original?
> 
> If it's breaking build or causing issues on runtime, yes.
> If it's just a cosmetic mistake when added (like in other file formats), no.

Going forward, a literal `\n` in an Android translation will show up as those two characters in the formatted result. It won't break the build or cause errors at runtime.

PR #3795, opened 2025-10-01
Ensure LocaleInsightsSnapshot.unreviewed_suggestions_lifespan is never null

### LINKED ISSUE #3794: Monthly activity summary for September is incorrect
```
Team activity for Italian (it)

(this part is correct)

Completion: 0% (vs. 99.87% in the previous month)
Strings with missing translations: 0 (vs. 76 in the previous month)
Unreviewed suggestions: 0 (vs. 111 in the previous month)
Unreviewed pretranslations: 0 (vs. 0 in the previous month)
```

The summary has everything at 0. I'm not sure about the last 3, but completion is definitely not 0 for Italian.

eemeli:
Fixes #3794

The attached migration took about 90 seconds to run for me locally. The `unreviewed_suggestions_lifespan` values that it sets are for the time when the migration is run, as generating the historical data would be quite cumbersome.

### DIFF
--- a/pontoon/insights/migrations/0018_fix_locale_insights.py
+++ b/pontoon/insights/migrations/0018_fix_locale_insights.py
@@ -1,0 +1,48 @@
+from datetime import date, datetime, time, timedelta, timezone
+
+from django.db import migrations
+
+from pontoon.insights.tasks import (
+    count_activities,
+    count_created_entities,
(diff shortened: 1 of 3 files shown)


mathjazz (approved):
Looks good! We should monitor performance of the daily task once this lands.

PR #3799, opened 2025-10-07
Add Entity and Search pages to Pontoon

functionzz:


pontoon/search/views.py
```diff
@@ -0,0 +1,44 @@
+import requests
+
+from django.http import Http404
+from django.shortcuts import get_object_or_404, redirect, render
+from django.urls import reverse
+
+from pontoon.base.models.entity import Entity
+
+
+BASE_URL = "http://localhost:8000/api/v2"
```
functionzz: Is there anywhere we have a macro that includes the BASE_URL, irrespective of production or local environment?

pontoon/search/static/css/translations.css
mathjazz, line 1: Where is this file loaded?

pontoon/search/views.py
mathjazz: We have [SITE_URL](https://github.com/mozilla/pontoon/blob/main/pontoon/settings/base.py#L83), but I'm not sure you mean that. Irrespective or respective?

mathjazz: If the DRF endpoint is part of the same Django project, you don’t need to make an actual HTTP request — you can just invoke the view logic directly, which should be faster and doesn’t require the network stack.

pontoon/search/static/css/translations.css
functionzz: come again?

functionzz:
> We have [SITE_URL](https://github.com/mozilla/pontoon/blob/main/pontoon/settings/base.py#L83), but I'm not sure you mean that. Irrespective or respective?
Exactly what I mean, thank you.

> If the DRF endpoint is part of the same Django project, you don’t need to make an actual HTTP request — you can just invoke the view logic directly, which should be faster and doesn’t require the network stack.

You are 100% right

pontoon/search/static/css/translations.css
mathjazz: I can't see is this file is actually being used anywhere.

pontoon/search/views.py
```diff
@@ -0,0 +1,77 @@
...
+
+    # Build the API URL or perform further logic
+    return render(request, "search/results.html", {"query_params": query_params})
+
+
+def entity(request, pk):
+    """Get corresponding entity given entity."""
+
+    # Modify the request object to include translations
+    drf_request = Request(request)
```
functionzz: @mathjazz, can you give your 2 cents on this section? In response to not using an api url to do return backend responses, I needed to modify the request query parameters before it was sent to the EntityIndividualView instantiation. Case in point, the "included_translation" flag is manually included.

pontoon/search/views.py
functionzz: I might need to make this happen - for some endpoints like `/new-search/?search=fi&locale=en-CA`, the fetching is quite slow(around 5 seconds), which could signify a problem with the implementation, but will also be sped up with this.

pontoon/search/static/css/translations.css
functionzz: Will remove these useless styles after review.

pontoon/search/views.py
mathjazz: This sounds more like a DB rather than a network latency issue.

pontoon/api/tests/test_views.py
```diff
@@ -612,7 +612,9 @@ def test_entity(django_assert_num_queries):
     assert response.status_code == 200
 
     assert response.data == {
-        "entity": {"id": entity.pk, "key": [], "string": "Test String"},
+        "id": entity.pk,
+        "key": [],
+        "string": "Test String",
```
mathjazz, line 720: What's the motivation for changing this data structure? I know we're currently still in beta, but we might ship this patch after GraphQL API is already deprecated.

pontoon/api/views.py
```diff
@@ -425,6 +425,13 @@ def get_queryset(self):
         text = query_params.get("text")
         locale = query_params.get("locale")
 
+        if query_params.get("search_identifiers") != "true":
+            query_params.pop("search_identifiers", None)
+        if query_params.get("search_match_case") != "true":
+            query_params.pop("search_match_case", None)
+        if query_params.get("search_match_whole_word") != "true":
+            query_params.pop("search_match_whole_word", None)
+
```
mathjazz, line 422: Should we also treat `include_disabled` and `include_system` in `ProjectListView` the same way?

pontoon/search/static/css/translations.css
mathjazz: Thanks, that's be great!

pontoon/search/static/js/search.js
```diff
@@ -0,0 +1,143 @@
...
+  $('.search-button').click(function () {
+    const searchOptions = {};
+
+    searchOptions['search'] = $('.search-input').val();
+    if (!searchOptions['search']) {
+      const errorMessage = 'Search term cannot be empty.';
+      $('.errors').html($('<p>').text(errorMessage));
+      return;
+    } else if (searchOptions['search'].length < 2) {
+      const errorMessage = 'Search term should be at least 2 characters long.';
```
mathjazz: Why 2? Why not 1?

pontoon/search/templates/search/widgets/entity_list.html
```diff
@@ -0,0 +1,31 @@
+
+{% macro list_entities(entities, entities_url, has_more) %}
+<ul class="entity-list" data-has-more="{{ 'true' if has_more else 'false' }}">
+  {% for entity in entities %}
+    <li class="entity-item">
+        <span>{{ entity.string }}</span>
```
mathjazz: We should add some semantic class names to this elements and the elements below.

```suggestion
        <span class="source-string">{{ entity.string }}</span>
```

pontoon/search/templates/search/widgets/entity_list.html
```diff
@@ -0,0 +1,31 @@
...
+        <span>{{ entity.key }}</span>
```
mathjazz: We should not render entity keys where they aren't meaningful. We should drop them at least for PO entities, because they are essentially the same as the source string.

pontoon/teams/templates/teams/widgets/team_selector.html
```diff
@@ -31,3 +31,34 @@
   </div>
 </div>
 {% endmacro %}
+
+<!-- Project selector -->
```
mathjazz: This macro should live under the `pontoon.projects` module.

pontoon/search/static/js/search.js
functionzz: Transvision has a < 2 character rule, so I figured to copy that.

functionzz: I will say that even at 2 characters, the current implementation is pretty slow in some cases.

pontoon/api/views.py
functionzz: I am not sure what implications that might have for people already using the API. I think its a good idea though

pontoon/api/tests/test_views.py
functionzz: this is to prevent the problem of calling `entity.entity.string`. Previously in our slack correspondence, you said to change it to `data.entity.string`. I thought that to be a bit confusing, especially for frontend developers, of which sometimes the response is also called `data`. So it might look like `data.data.entity.string`, which is not ideal. Which is why I made it like `entity.string`, `entity.key` etc.

pontoon/api/views.py
mathjazz: We might also already have users of the `search_*` parameters.

pontoon/search/static/js/search.js
mathjazz: Is the performance the same as in the translate view? https://pontoon.mozilla.org/sl/all-projects/all-resources/

pontoon/api/tests/test_views.py
functionzz: this is to prevent the problem of calling entity.entity.string. Previously in our slack correspondence, you said to change it to data.entity.string. I thought that to be a bit confusing, especially for frontend developers, of which sometimes the response is also called data. So it might look like data.data.entity.string, which is not ideal. Which is why I made it like entity.string, entity.key etc.

pontoon/api/views.py
functionzz: You mean the /search/translate view? If that's the case, is changing still advisable?

pontoon/search/static/js/search.js
functionzz: its a great deal slower. This is instant in comparison :(

mathjazz: OK, then we need to fix this first.

pontoon/api/views.py
mathjazz: Yeah, I think this change makes sense, so it should be applied everywhere. Also for the `include_translations` flag, which I forgot to mention above.

pontoon/api/tests/test_views.py
mathjazz: The first `entity` in `entity.entity.string` is not part of the API response, so I suggested we use a different variable name in the code. [Swagger](https://pontoon.mozilla.org/api/v2/entities/) uses `results`, it could be called `data`, `output` or something like that.

I do agree though that `for entity in entities` is easier to reason about than e.g. `for entity_data in entities`, and that dropping the `entity` dict is more consistent with the way `locale` and `project` endpoints work, so let's keep this change. 👍

pontoon/search/static/js/search.js
functionzz: Removed 2 character rule

pontoon/search/templates/search/widgets/entity_list.html
mathjazz: Looks like you dropped entity keys for all entities? I don't think that's a good idea, because they are often meaningful. Not showing them will be quite painful when using the "Search Identifiers" parameter.

pontoon/search/templates/search/widgets/entity_list.html
mathjazz: Thanks for adding some class names. Some elements still don't have them, though.

We should also use semantic elements. `<span>` is an inline element and we are unlikely to render source string and entity URL in one line.

mathjazz (commented):
Thanks for the update.

**A general note:** It's hard to follow the changes if not all comments are addressed. We either need to keep track of the exact list of things to do, or (preferable) address all review comments before requesting a new review.

**We should:**
- [ ] Investigate why is the initial page load of the /search page rather slow
- [x] Require explicit `true` value for all boolean query parameters
- [ ] Use semantic elements and class names in markup
- [x] Keep entity.key in /search and /entities unless it's just a reformated string (e.g. in PO files)
- [ ] Add keyboard shortcut to trigger search with `Enter`
- [ ] Add ability to copy search results to clipboard

pontoon/search/templates/search/widgets/entity_list.html
```diff
@@ -4,7 +4,11 @@
   {% for entity in entities %}
     <li class="entity-item">
         <span class="source-string">{{ entity.string }}</span>
-        <span class="entity-keys">{{ ", ".join(entity.key) }}</span>
+        {% if entity.resource.format != "gettext" %}
+          <span class="entity-keys">{{ ", ".join(entity.key) }}</span>
+        {% else %}
+          <span class="entity-keys">N/A</span>
```
mathjazz: I'd just skip this block for Gettext entities, as we do in the translate view.

pontoon/projects/templates/projects/widgets/project_selector.html
```diff
@@ -0,0 +1,30 @@
+<!-- Project selector -->
+{% macro project(projects_list, project_selected) %}
+<div class="locale-selector">
```
mathjazz: locale -> project (also in a few places below)

pontoon/base/templates/widgets/checkbox.html
```diff
@@ -1,8 +1,13 @@
-{% macro checkbox(label, class, attribute, is_enabled=False, title=False, help=False, form_field=None) %}
+{% macro checkbox(label, class, attribute, is_enabled=False, title=False, help=False, form_field=None, icon_first=False) %}
 <div class="check-box {{ class }}{% if is_enabled %} enabled{% endif %}" data-attribute="{{ attribute }}"{% if title %} title="{{title}}"{% endif %}>
     <div class="check-box-wrapper">
+        {% if icon_first %}
```
mathjazz: Can't we do that with CSS?

pontoon/base/static/css/style.css
```diff
@@ -229,6 +229,10 @@ table.striped tr:nth-child(even) {
   background: var(--dark-grey-1);
 }
 
+ul.striped li:nth-child(even) {
+  background: var(--dark-grey-1);
+}
```
mathjazz: This is probably no longer needed?

pontoon/search/urls.py
```diff
@@ -0,0 +1,24 @@
...
+)
+
+
+urlpatterns = [
+    path(
+        "search/",
+        entity_search,
+        name="pontoon.search",
+    ),
+    path("ajax/more-translations/", more_entities, name="pontoon.search.more"),
```
mathjazz: Can we make this behave like the TM search page, i.e. use query params for additional pages that will allow users to share links to pages with more results?

mathjazz: Worst case scenario, the URL like `search/more/` would be more approriate IMHO.

pontoon/search/templates/search/widgets/search_results.html
```diff
@@ -0,0 +1,37 @@
...
+      <div class="source-string-container">
+        <span class="source-string">{{ entity.string|advanced_highlight_matches(search, match_case_enabled, match_whole_word_enabled) }}</span>
+      </div>
+
+      <div class="translation-string-container">
+        <span class="translation-string">{{ entity.translation.string|advanced_highlight_matches(search, match_case_enabled, match_whole_word_enabled) }}</span>
+      </div>
+
+      <div class="entity-info-container">
+        <a class="project" href="/projects/{{ entity.project.slug }}">{{ entity.project.name }}</a>
```
mathjazz: Rather than hardocoding the URLs, we should use the `url()` helper in the templates. 

For example, see:
https://github.com/mozilla/pontoon/blob/main/pontoon/teams/templates/teams/includes/permissions.html#L12C6

pontoon/base/templatetags/helpers.py
```diff
@@ -307,6 +307,35 @@ def highlight_matches(text, search_query):
     return mark_safe(highlighted_text)
 
 
+@library.filter
+def advanced_highlight_matches(
```
mathjazz: Have you considered expanding `highlight_matches` rather than adding a separate filter?

pontoon/search/urls.py
functionzz: Should we take another look at pagination? Or is this scroll functionality something we want to keep using?

pontoon/base/templatetags/helpers.py
functionzz: I have. The default case for `highlight_matches` and `advanced_highlight_matches` is different, I believe the options are to have 2 seperate implementations or have a flag like `advanced=True` to denote the difference between the two.

`advanced_highlight_matches` supports individual word matching by splitting the input, while `highlight_matches` takes the whole input as one word. These 2 default cases conflict.

pontoon/search/templates/search/search.html
```diff
@@ -0,0 +1,64 @@
...
+        </div>
+      </div>
+    <div class="check-list noselect">
+      {{ Checkbox.checkbox('Search in string Identifiers', class='field search-identifiers-enabled', attribute='search-identifiers-enabled',
+          is_enabled=search_identifiers_enabled) }}
+      {{ Checkbox.checkbox('Case Sensitive', class='field match-case-enabled', attribute='match-case-enabled',
+          is_enabled=match_case_enabled) }}
+      {{ Checkbox.checkbox('Entire Words', class='field match-whole-word-enabled', attribute='match-whole-word-enabled',
+          is_enabled=match_whole_word_enabled) }}
+    </div>
```
mathjazz: Nit: Please indent this block.

pontoon/search/templates/search/search.html
```diff
@@ -0,0 +1,64 @@
...
+  <section id="search-results">
+      {% if entities and search %}
+        <div id="entity-list-header">
+          <span class="source-header">Source</span>
+          <span class="translation-header">Translation</span>
+        </div>
+        <ul id="entity-list" data-has-more="{{ 'true' if has_more else 'false' }}">
+          {% include 'search/widgets/search_results.html' %}
+        </ul>
+      {% elif not entities and search %}
```
mathjazz: Would that not work here?

```suggestion
      {% else %}
```

pontoon/search/templates/search/widgets/entity_translations.html
```diff
@@ -0,0 +1,60 @@
+{# Widget to display list of translations. #}
+{% macro list_translations(entity) %}
+    <div id="initial-section-container" class="translation-container">
+      <div class="source-string-container">
+        <span class="source-string">{{ entity.string }}</span>
+      </div>
+
+      <div class="translation-info-container">
+        <div class="entity-info-container">
+          <a class="project" href="/projects/{{ entity.project.slug }}">{{ entity.project.name }}</a>
```
mathjazz: The URLs in this file are still hardcoded.

pontoon/search/templates/search/widgets/entity_translations.html
```diff
@@ -0,0 +1,60 @@
+{# Widget to display list of translations. #}
+{% macro list_translations(entity) %}
+    <div id="initial-section-container" class="translation-container">
```
mathjazz: Please dedent this block (my bad!).

pontoon/search/templates/search/widgets/entity_translations.html
```diff
@@ -0,0 +1,60 @@
...
+            Copy
+          </button>
+        </div>
+      </div>
+    </div>
+
+<ul id="translation-list">
+    {% for translation in entity.translations %}
+      <li class="translation-container">
+        <div class="source-string-container">
```
mathjazz:
```suggestion
        <div class="translation-string-container">
```

pontoon/search/urls.py
mathjazz: I would make it work exactly as on https://pontoon.mozilla.org/sl/translation-memory/?pages=2 - so load more data on scroll, but make sure we also update the URL.

pontoon/base/templatetags/helpers.py
mathjazz: Yup, a flag is what I've been thinking about.

pontoon/search/static/js/search.js
```diff
@@ -0,0 +1,129 @@
+$(function () {
+  $('.search-btn').click(function () {
+    $('.search-input').trigger('enterKey');
+  });
```
mathjazz: Drop?

pontoon/search/static/js/search.js
```diff
@@ -0,0 +1,129 @@
...
+      if (e.which === 13) {
+        $(this).trigger('enterKey');
+        return false;
+      }
+    });
+
+  $('.search-input').on('enterKey', function () {
+    const searchOptions = {};
+
+    searchOptions['search'] = $('.search-input').val();
```
mathjazz: I'd normalize the input:

```suggestion
    searchOptions['search'] = $('.search-input').val()?.trim();
```

pontoon/search/static/js/search.js
```diff
@@ -0,0 +1,129 @@
...
+    .unbind('keydown.pontoon')
+    .bind('keydown.pontoon', function (e) {
+      if (e.which === 13) {
+        $(this).trigger('enterKey');
+        return false;
+      }
+    });
+
+  $('.search-input').on('enterKey', function () {
+    const searchOptions = {};
```
mathjazz: Why do you use `searchOptions`? I think something like the code below would be much simpler.

```js
$('.search-input').on('enterKey', function () {
  const $input = $('.search-input');
  const search = $input.val()?.trim();
  if (!search) return;

  const url = new URL(window.location.href);
  url.search = '';

  const params = url.searchParams;
  params.set('search', search);

  const locale = $('.locale .selector .language').data('code');
  params.set('locale', locale);

  const project = $('.project .selector .selected-project').data('slug');
  if (project && project !== 'all-projects') {
    params.set('project', project);
  }

  const checkboxParamMap = {
    'search-identifiers-enabled': 'search_identifiers',
    'match-case-enabled': 'search_match_case',
    'match-whole-word-enabled': 'search_match_whole_word',
  };

  $('.check-box.enabled').each(function () {
    const attr = $(this).data('attribute');
    const param = checkboxParamMap[attr];
    if (param) params.set(param, 'true');
  });

  window.location.href = url.toString();
});
```

pontoon/search/templates/search/search.html
```diff
@@ -0,0 +1,64 @@
...
+          is_enabled=search_identifiers_enabled) }}
+      {{ Checkbox.checkbox('Case Sensitive', class='field match-case-enabled', attribute='match-case-enabled',
+          is_enabled=match_case_enabled) }}
+      {{ Checkbox.checkbox('Entire Words', class='field match-whole-word-enabled', attribute='match-whole-word-enabled',
+          is_enabled=match_whole_word_enabled) }}
+    </div>
+    </menu>
+
+  <section id="search-results">
+      {% if entities and search %}
```
mathjazz: Whould that not ne sufficient here?

```suggestion
      {% if entities %}
```

pontoon/search/views.py
```diff
@@ -0,0 +1,228 @@
...
+        entity = response.json()
+        return render(
+            request,
+            "search/entity.html",
+            {
+                "entity": entity,
+            },
+        )
+    else:
+        raise Http404
```
mathjazz: I think this would be similar:

```suggestion
        data = response.json()
    except (RequestException, ValueError):
        raise Http404

    return render(request, "search/entity.html", {"entity": data})
```

pontoon/search/views.py
```diff
@@ -0,0 +1,228 @@
...
+    return f"{SITE_URL}/api/v2/search/translations/?{urlencode(query_params)}"
+
+
+def entity_search(request):
+    """Get corresponding entity given entity."""
+
+    search = request.GET.get("search")
+    locale = request.GET.get("locale")
+    project = request.GET.get("project")
+    search_identifiers = request.GET.get("search_identifiers")
```
mathjazz: Something like this would be more reliable and save as from all the ` == "true"`'s below:

```suggestion
    search_identifiers = parse_bool(request.GET.get("search_identifiers"))
```

```python
# pontoon/base/utils.py
def parse_bool(value) -> bool:
    return str(value).lower() in ("1", "true", "yes", "on")

File: pontoon/search/views.py [hunk 27]
```diff
@@ -0,0 +1,228 @@
...
+    )
+
+    default_project = Project(name="All Projects", slug="all-projects")
+    projects.insert(0, default_project)
+
+    if not project or not Project.objects.filter(slug=project).exists():
+        preferred_project = default_project
+        project = None
+    else:
+        preferred_project = Project.objects.get(slug=project)
```
mathjazz: This would only hit the DB once:

```suggestion
    preferred_project = Project.objects.filter(slug=project_slug).first()
    if not preferred_project:
        preferred_project = default_project
        project = None
```

pontoon/search/views.py
```diff
@@ -0,0 +1,228 @@
...
+
+    try:
+        response = requests.get(api_url)
+        response.raise_for_status()
+    except RequestException:
+        raise Http404
+
+    if response.status_code == 200:
+        entities = response.json()["results"]
+        has_more = response.json()["next"]
```
mathjazz:
```suggestion
        data = response.json()
        entities = data["results"]
        has_more = data["next"]
```

pontoon/search/views.py
```diff
@@ -0,0 +1,228 @@
...
+
+    try:
+        response = requests.get(api_url)
+        response.raise_for_status()
+    except RequestException:
+        raise Http404
+
+    if response.status_code == 200:
+        entities = response.json()["results"]
+        has_more = response.json()["next"] is not None
```
mathjazz:
```suggestion
        data = response.json()
        entities = data["results"]
        has_more = data["next"] is not None
```

pontoon/search/views.py
```diff
@@ -0,0 +1,228 @@
...
+        search_match_whole_word,
+    )
+
+    try:
+        response = requests.get(api_url)
+        response.raise_for_status()
+    except RequestException:
+        raise Http404
+
+    if response.status_code == 200:
```
mathjazz: If you call `raise_for_status()` and it didn’t raise, you’re already in the 2xx/3xx land, so no need for this check.

pontoon/search/views.py
```diff
@@ -0,0 +1,228 @@
...
+        page=page,
+    )
+
+    try:
+        response = requests.get(api_url)
+        response.raise_for_status()
+    except RequestException:
+        raise Http404
+
+    if response.status_code == 200:
```
mathjazz: Same as above.

pontoon/search/static/css/search.css
```diff
@@ -0,0 +1,248 @@
...
+  .container {
+    menu {
+      margin-bottom: 40px;
+
+      #search-menu {
+        margin: 0 auto;
+        display: flex;
+        gap: 8px;
+
+        .search-wrapper {
```
mathjazz: Fixes the magnifier icon position in the dropdown menus pointed out by @flodolo:
```suggestion
        >  .search-wrapper {
```

pontoon/search/templates/search/search.html
```diff
@@ -0,0 +1,64 @@
...
+        </div>
+        <div id="project-selector">
+          {{ project_selector.project(projects, preferred_project) }}
+        </div>
+        <div id="locale-selector">
+          {{ team_selector.locale(locales, preferred_locale) }}
+        </div>
+      </div>
+    <div class="check-list noselect">
+      {{ Checkbox.checkbox('Search in string Identifiers', class='field search-identifiers-enabled', attribute='search-identifiers-enabled',
```
mathjazz: As suggested by @flodolo:

```suggestion
      {{ Checkbox.checkbox('Search in string identifiers', class='field search-identifiers-enabled', attribute='search-identifiers-enabled',
```

pontoon/search/templates/search/search.html
```diff
@@ -0,0 +1,64 @@
...
+          is_enabled=search_identifiers_enabled) }}
+      {{ Checkbox.checkbox('Case Sensitive', class='field match-case-enabled', attribute='match-case-enabled',
```
mathjazz: As suggested by @flodolo:

```suggestion
      {{ Checkbox.checkbox('Case sensitive', class='field match-case-enabled', attribute='match-case-enabled',
```

pontoon/search/templates/search/search.html
```diff
@@ -0,0 +1,64 @@
...
+          is_enabled=match_case_enabled) }}
+      {{ Checkbox.checkbox('Entire Words', class='field match-whole-word-enabled', attribute='match-whole-word-enabled',
```
mathjazz: As suggested by @flodolo:

```suggestion
      {{ Checkbox.checkbox('Entire words', class='field match-whole-word-enabled', attribute='match-whole-word-enabled',
```

mathjazz (commented):
Thanks for the updates!

Left some more suggestions, including the ones pointed out by Flod.

pontoon/search/templates/search/search.html
functionzz: No because we want the initial render of the page to not include `No Results` :)

mathjazz: 🤦

pontoon/search/static/js/search.js
functionzz: I think my motivation for using a map was to keep everything consistent, but I like this ways simplicity.

pontoon/base/templatetags/helpers.py
```diff
@@ -287,20 +287,36 @@ def set_attrs(attrs, new=False):
...
+):
     """Highlight all occurrences of the search query in the text."""
     if not search_query:
         return text
 
-    # First, escape the text to prevent HTML rendering
     escaped_text = escape(text)
+    flags = 0 if match_case_enabled else re.IGNORECASE
+
+    if advanced:
```
mathjazz, line 309: A comment explaining what the `advanced` param does would be nice.

pontoon/base/templatetags/helpers.py
```diff
@@ -287,20 +287,36 @@ def set_attrs(attrs, new=False):
...
+        terms = list(set(search_query.split() + [search_query]))
+        escaped_terms = [escape(term) for term in terms]
+        pattern = "|".join(rf"{re.escape(term)}" for term in escaped_terms)
+
+        if match_whole_word_enabled:
+            pattern = "|".join(rf"\b{re.escape(term)}\b" for term in escaped_terms)
```
mathjazz: The following would be following the DRY principle:

```suggestion
        boundary = r"\b" if match_whole_word_enabled else ""
        pattern = "|".join(
            rf"{boundary}{re.escape(term)}{boundary}"
            for term in escaped_terms
        )
```

pontoon/search/templates/search/search.html
```diff
@@ -0,0 +1,64 @@
...
+        {{ Checkbox.checkbox('Search in string identifiers', class='field search-identifiers-enabled', attribute='search-identifiers-enabled',
+            is_enabled=search_identifiers_enabled) }}
+        {{ Checkbox.checkbox('Case sensitive', class='field match-case-enabled', attribute='match-case-enabled',
+            is_enabled=match_case_enabled) }}
+        {{ Checkbox.checkbox('Entire words', class='field match-whole-word-enabled', attribute='match-whole-word-enabled',
+            is_enabled=match_whole_word_enabled) }}
+      </div>
+    </menu>
+
+  <section id="search-results">
```
mathjazz: Nit: This line and line 53 need to be indented.

Also, we should introduce a formatter for HTML: https://github.com/mozilla/pontoon/issues/3941.

pontoon/search/templates/search/widgets/entity_translations.html
```diff
@@ -0,0 +1,60 @@
+{# Widget to display list of translations. #}
+{% macro list_translations(entity) %}
+<div id="initial-section-container" class="translation-container">
+  <div class="translation-string-container">
```
mathjazz:
```suggestion
<div id="initial-section-container">
  <div class="source-string-container">
```

pontoon/search/templates/search/search.html
```diff
@@ -0,0 +1,64 @@
...
+<section id="main">
+  <div class="container">
+    <menu>
+      <div id="search-menu" class="controls">
+        <div class="search-wrapper big clearfix">
+          <div class="icon fas fa-search"></div>
+          <input class="search-input" type="search" autocomplete="off" autofocus placeholder="Type and press Enter to search" value="{{ search }}">
+        </div>
+        <div id="project-selector">
+          {{ project_selector.project(projects, preferred_project) }}
```
mathjazz: I suggest we rename all `preferred_project` and `preferred_locale` instances to `project` and `locale` respectivelly. I suspect the `preferred_` prefix is a bad copy-paste from the `settings.html` file, where it actually makes sense to use it.

```suggestion
          {{ project_selector.project(projects, locale) }}
```

pontoon/search/views.py
```diff
@@ -0,0 +1,221 @@
...
+        if value is not None and value != "" and value is not False
+    }
+
+    if page:
+        query_params["page"] = page
+
+    return f"{SITE_URL}/api/v2/search/translations/?{urlencode(query_params)}"
+
+
+def entity_search(request):
```
mathjazz:
```suggestion
def search(request):
```

mathjazz: I would just load the empty search page with this view. And the view below (`more_entities` -> `search_results`) would load the search results, both on initial page load if search parameters are given and on scroll. That would be more DRY and consistent with the TM search page.

mathjazz (commented):
Excellent work! Left a few more minor suggestions and one request for a slightly bigger /search page logic change.

pontoon/base/templatetags/helpers.py
```diff
@@ -287,20 +287,42 @@ def set_attrs(attrs, new=False):
...
+def highlight_matches(
+    text,
+    search_query,
+    advanced=False,
+    match_case_enabled=False,
+    match_whole_word_enabled=False,
+):
+    """Highlight all occurrences of the search query in the text.
+
+    Advanced usage is relegated for Translation Search, where more
```
mathjazz: This is a general-purpose filter, we shouldn't call out out the only use case we have for it ATM, because there might be more in the future.

pontoon/base/templatetags/helpers.py
```diff
@@ -287,20 +287,42 @@ def set_attrs(attrs, new=False):
...
+
+    inclusive whole word checks can be used.
```
mathjazz: It's not clear from the description what exactly is the difference the `advanced` parameter makes.

pontoon/search/urls.py
```diff
@@ -0,0 +1,24 @@
...
+)
+
+
+urlpatterns = [
+    path(
+        "search/",
+        search,
+        name="pontoon.search",
+    ),
+    path("ajax/more-translations/", search_results, name="pontoon.search.more"),
```
mathjazz:
```suggestion
    path("search/results/", search_results, name="pontoon.search.results"),
```

pontoon/search/templates/search/search.html
```diff
@@ -38,19 +38,19 @@
...
+          <div id="entity-list-header">
+            <span class="source-header">Source</span>
+            <span class="translation-header">Translation</span>
+          </div>
+          <ul id="entity-list" data-has-more="{{ 'true' if has_more else 'false' }}">
+            {% include 'search/widgets/search_results.html' %}
+          </ul>
+        {% elif not entities and search %}
+          <p id="no-results">No results found.</p>
+        {% endif %}
```
mathjazz: (This block should not be indented - the indent is 2 spaces, not 4. Just the `<section>` openning and closing tag should be. But let's leave it now.)

pontoon/search/views.py
mathjazz: This still remains to be done.

pontoon/search/views.py
```diff
@@ -0,0 +1,176 @@
...
+
+    default_project = Project(name="All Projects", slug="all-projects")
+    projects.insert(0, default_project)
+
+    project = (
+        Project.objects.filter(slug=project_slug).first() if project_slug else None
+    )
+    if not project:
+        project = default_project
+        project_slug = None
```
mathjazz: What do you think about that?

```python
project = (
    Project.objects.filter(slug=project_slug).first()
    if project_slug
    else default_project
)
```

We don't seem to need `project_slug` past this point.

pontoon/search/static/js/search.js
```diff
@@ -0,0 +1,142 @@
...
+
+    $('.check-box.enabled').each(function () {
+      const attr = $(this).data('attribute');
+      const param = checkboxParamMap[attr];
+      if (param) {
+        params.set(param, 'true');
+      }
+    });
+
+    window.location.href = `${url.pathname}?${params.toString()}`;
```
mathjazz: Why don't we call `loadMoreEntries()` here instead?

pontoon/search/static/js/search.js
```diff
@@ -0,0 +1,142 @@
...
+      },
+    });
+  }
+
+  $(window).on('load', async function () {
+    const pages = parseInt(params.get('pages')) || 1;
+
+    if (params.get('search')) {
+      Pontoon.NProgressUnbind();
+      NProgress.start();
```
mathjazz: Why do we disable the progress bar (`Pontoon.NProgressUnbind`) and the enable it again (`NProgress.start();`)? Shouldn't we just remove these lines and the 3 related lines below?

functionzz: Since I used promises to load the for loop of responses, the progress bar is manually configured. That is because if you remove the configuration, the progress bar shows progress multiple times i.e it will show a green loading pattern 5 times for 5 responses. So thus I overrode the default behaviour.

pontoon/search/static/js/search.js
functionzz: I figured since setting `window.location.href` reloads the page and calls `$(window).on('load'...`, which is functionally similar to directly calling loadMoreEntries().

pontoon/search/static/js/search.js
mathjazz: Oh, this is suboptimal:
https://github.com/mozilla/pontoon/pull/3799/changes#diff-f953a22dba5bc95355bc0bfe715ee5668611212f4229f068464c391e6d858794R62-R68

There should only be one request. See how it's implemented in Translation Memory tab.

pontoon/search/static/js/search.js
functionzz: Removed `window.location.href` in favor for ajax call.

pontoon/search/static/js/search.js
functionzz: Implemented single requests over multiple looped requests.

pontoon/search/templates/search/search.html
```diff
@@ -0,0 +1,60 @@
...
+    </menu>
+
+    <section id="search-results">
+          <div id="entity-list-header">
+            <span class="source-header">Source</span>
+            <span class="translation-header">Translation</span>
+          </div>
+          <ul id="entity-list">
+          </ul>
+          <p id="no-results">No results found.</p>
```
mathjazz:
```suggestion
          <p id="no-results">No matches found.</p>
```

pontoon/search/views.py
```diff
@@ -0,0 +1,215 @@
...
```
mathjazz:
```suggestion
    project = (
        Project.objects.filter(slug=project_slug).first()
        if project_slug
        else default_project
    )
```

PR #3802, opened 2025-10-09
Add Translation.value and .properties, with data model contents

eemeli:
Adds the JSON fields `.value` (required) and `.properties` (optional, default None) to Translation, and sets them by parsing the `.string` at all places where translation objects are created.

These fields are not (yet!) read anywhere, which also matches the corresponding Entity fields. The intent is to submit a follow-up PR that changes the REST API to pass these fields rather than `.string` for all entities and translations, in both requests and replies.

The data migration included here is... heavy. It needs to read, parse, and modify _all_ Translation objects. That'll take a while even on a high-end machine.


mathjazz (commented):
Thanks for the patch. The code looks good.

We need an explicit deployment plan, including:
* expected runtime on prod-ish data
* rollback/failure plan
* whether this should be split into schema migration + offline/backfill command instead of a blocking Django migration

mathjazz (approved):

PR #3820, opened 2025-11-03
Exclude private projects from TM search endpoint

mathjazz:
Translations from private projects should not be returned by the TranslationMemory search endpoint.

I've also simplified the code a little.

### DIFF
--- a/pontoon/api/tests/test_views.py
+++ b/pontoon/api/tests/test_views.py
@@ -972,6 +972,20 @@ def test_tm_search(django_assert_num_queries):
         string="Entity B",
         resource=resource_b,
     )
+    project_private = ProjectFactory(
+        slug="project_private",
+        name="Project Private",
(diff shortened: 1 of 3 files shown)


pontoon/api/tests/test_views.py
```diff
@@ -993,8 +1007,15 @@ def test_tm_search(django_assert_num_queries):
...
+    TranslationMemoryEntry.objects.create(
+        source="Hello",
+        target="Hola",
+        locale=locale_a,
+        project=project_private,
+        entity=entity_private,
+    )
 
-    with django_assert_num_queries(4):
+    with django_assert_num_queries(2):
```
functionzz, line 1018: Just curious, how did this change decrease the amount of queries?

functionzz (dismissed):
LGTM.

PR #3827, opened 2025-11-07
API Optimizations

### LINKED ISSUE #3824: Generalize selective field exclusion to all API endpoint views
Recently, through [this PR](https://github.com/mozilla/pontoon/pull/3823), we added the feature of selectively excluding fields to the `/locale` endpoint view. Extending this capability to all API views would improve efficiency for API calls that do not require all fields, reducing response sizes and bandwidth usage.

functionzz:
## Description
Applies the API optimizations found in #3823 to every API view.
- Conditionally prefetches fields based on the `fields` query parameter
- Includes a N+1 query problem fix for `TermSearchListView`
- Applies `DynamicFieldsModelSerializer` to `ProjectLocaleIndividualView`
- Adjusts `TermListView` tests

Fixes #3824.

## Additional Notes
- ~Probably needs more tests for the `TermListView`~
- ~Needs more thorough testing for individual fields to ensure no N+1 query problems are introduced into the code~

### DIFF
--- a/pontoon/api/serializers.py
+++ b/pontoon/api/serializers.py
@@ -249,7 +249,7 @@ class TermSerializer(DynamicFieldsModelSerializer):
         if hasattr(obj, "filtered_translations") and (ft := obj.filtered_translations):
             return ft[0].text
 
-        return translation
+        return None
 
 
 class TranslationMemorySerializer(DynamicFieldsModelSerializer):


pontoon/api/serializers.py
```diff
@@ -255,13 +255,17 @@ class Meta:
 
     def get_translation_text(self, obj):
         request = self.context.get("request")
-        locale = request.query_params.get("locale") if request else None
 
-        if not locale:
+        if not request:
             return None
```
eemeli: Why is this check included? `request` isn't used later in the function.

pontoon/api/views.py
```diff
@@ -153,14 +154,28 @@ class LocaleIndividualView(generics.RetrieveAPIView):
...
-                "project_locale",
-                queryset=ProjectLocale.objects.visible().select_related("project"),
-                to_attr="fetched_project_locales",
+        qs = Locale.objects.visible()
+
+        fields_param = self.request.query_params.get("fields", "")
+        requested = set(f.strip() for f in fields_param.split(",") if f.strip())
+
+        # Only prefetch project data when requested
+        needs_projects = not requested or requested & {"projects"}
```
eemeli: Wat? Is this not the same as:
```suggestion
        needs_projects = not requested or "projects" in requested
```

pontoon/api/views.py
```diff
@@ -153,14 +154,28 @@ class LocaleIndividualView(generics.RetrieveAPIView):
...
     def get_queryset(self):
-        queryset = Locale.objects.prefetch_related(
-            Prefetch(
-                "project_locale",
-                queryset=ProjectLocale.objects.visible().select_related("project"),
-                to_attr="fetched_project_locales",
+        qs = Locale.objects.visible()
+
+        fields_param = self.request.query_params.get("fields", "")
+        requested = set(f.strip() for f in fields_param.split(",") if f.strip())
```
eemeli:
```suggestion
        requested = set(fs for f in fields_param.split(",") if (fs := f.strip()))
```

pontoon/api/views.py
```diff
@@ -171,81 +186,122 @@ def get_queryset(self):
...
+                )
             )
-        ).distinct()
 
-        return queryset.stats_data()
+        needs_contact = not requested or requested & {"contact"}
+        if needs_contact:
+            qs = qs.prefetch_related("contact")
+
+        needs_tags = not requested or requested & {"tags"}
```
eemeli: Same wat as above. There's more below that need fixing the same way.

pontoon/api/views.py
```diff
@@ -171,81 +186,122 @@ def get_queryset(self):
...
-                "resource",
-                "resource__project",
-            )
-            .order_by("id")
+        qs = Entity.objects.filter(resource__project__disabled=False).prefetch_related(
+            "resource"
         )
 
+        fields_param = self.request.query_params.get("fields", "")
+        requested = set(f.strip() for f in fields_param.split(",") if f.strip())
```
eemeli: Same fix as above, also applies below.

pontoon/api/serializers.py
```diff
@@ -254,14 +246,13 @@ class Meta:
...
-        request = self.context.get("request")
-        locale = request.query_params.get("locale") if request else None
-
-        if not locale:
-            return None
+        translation = (
+            obj.filtered_translations[0].text
+            if hasattr(obj, "filtered_translations") and obj.filtered_translations
+            else None
+        )
```
eemeli: (nitpick, style) Split in two as GitHub can't handle them in one:
```suggestion
        if hasattr(obj, "filtered_translations") and (ft := obj.filtered_translations):
            return ft[0].text
```

pontoon/api/serializers.py
```diff
@@ -254,14 +246,13 @@ class Meta:
...
 
-        term = obj.translations.filter(locale__code=locale).first()
-        return term.text if term else None
+        return translation
```
eemeli:
```suggestion
        return None
```

pontoon/api/views.py
```diff
@@ -153,14 +154,28 @@ class LocaleIndividualView(generics.RetrieveAPIView):
...
-                queryset=ProjectLocale.objects.visible().select_related("project"),
-                to_attr="fetched_project_locales",
+        qs = Locale.objects.visible()
+
+        fields_param = self.request.query_params.get("fields", "")
+        requested = set(fs for f in fields_param.split(",") if (fs := f.strip()))
+
+        # Only prefetch project data when requested
+        needs_projects = not requested or "projects" in requested
+        if needs_projects:
```
eemeli: Not really convinced that these `needs_*` variables are helping; I'd leave them out, but wouldn't complain if you think they're helping. Same also applies to all the other similar cases.

```suggestion
        if not requested or "projects" in requested:
```

pontoon/api/views.py
```diff
@@ -153,14 +154,28 @@ class LocaleIndividualView(generics.RetrieveAPIView):
...
     def get_queryset(self):
-        queryset = Locale.objects.prefetch_related(
-            Prefetch(
-                "project_locale",
-                queryset=ProjectLocale.objects.visible().select_related("project"),
-                to_attr="fetched_project_locales",
+        qs = Locale.objects.visible()
+
+        fields_param = self.request.query_params.get("fields", "")
+        requested = set(fs for f in fields_param.split(",") if (fs := f.strip()))
```
eemeli: You're repeating this same operation so many times you should extract it into a utility function, so that you could here do something like:
```suggestion
        requested = self.request_fields()
```

eemeli (changes requested):
More nitpicking.

PR #3829, opened 2025-11-12
Fixes db migration with PostgreSQL 17

flozz:
An error occurs when running the database migration 81 on PostgreSQL 17 (I have not tested on PostgreSQL 16, and it works with PostgreSQL 15):

    django.db.utils.DataError: invalid regular expression: invalid escape \ sequence

This PR just replaces a `\S` escape sequence by `[^[:space:]]` that is equivalent¹ but that works fine on PG17. :)

¹ Doc: https://www.postgresql.org/docs/current/functions-matching.html#:~:text=matches%20any%20non-whitespace%20character

### DIFF
--- a/pontoon/base/migrations/0081_fix_android_spaces.py
+++ b/pontoon/base/migrations/0081_fix_android_spaces.py
@@ -16,7 +16,7 @@ WITH tm AS (
     FROM base_translationmemoryentry m
     JOIN base_translation t ON t.id = m.translation_id
     WHERE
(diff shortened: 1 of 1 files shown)


eemeli:
This is rather surprising, as according to the docs linked to above, `\S` should work also in Postgres 17. I'm going to see if I can replicate this issue, and get a better handle for what's going on here.

mathjazz:
@flozz Could you share the exact steps to reproduce the problem this PR is trying to fix?

PR #3833, opened 2025-11-17
Add ability to configure USE_X_FORWARDED_HOST

mathjazz:
We need to set `USE_X_FORWARDED_HOST` to `True` or `redirect_uri` used by Mozilla Accounts on Sign in will point to https://mozilla-pontoon-staging.herokuapp.com/ even when on https://pontoon.allizom.org/.

See https://docs.allauth.org/en/dev/socialaccount/providers/saml.html#guidelines.

Also: fix a typo and order in the docs.


functionzz (approved):
LGTM

PR #3834, opened 2025-11-17
Pretranslation API

### LINKED ISSUE #3830: Implement Pretranslation API
## Feature Summary
Implement an authenticated API endpoint which returns a Pretranslation for a given Entity and locale. 
Located at `/api/v2/pretranslation/code/entity_id`.

## Motivation / Use Case
We currently have existing Pretranslation functionality in Pontoon. Extending this to an API endpoint can provide easier access for external users of Pretranslation functionality.

functionzz:
## Description
This PR implements a Pretranslation API at `api/v2/pretranslate/?format_string=FORMAT&locale=CODE`. It utilizes existing pretranslation functionality in Pontoon and extends it as an authenticated service for those who need a pretranslation from either Translation Memory or AutoML.

Fixes #3830.

## Additional Notes
~Authentication for this endpoint is temporarily disabled for easier testing~
~Requires some testing on `android` and `gettext` resource formats.~

### DIFF
--- a/pontoon/api/views.py
+++ b/pontoon/api/views.py
@@ -459,7 +459,7 @@ class TranslationSearchListView(RequestFieldsMixin, generics.ListAPIView):
         except ValueError as error:
             raise ValueError(error)
 
-        return entities
+        return qs
 
 
 class PretranslationView(APIView):


pontoon/api/views.py
```diff
@@ -459,3 +464,59 @@ def get_queryset(self):
...
+MAX_TEXT_CHARS = 2048
+MAX_TEXT_BYTES = MAX_TEXT_CHARS * 4
+
+
+class PretranslationView(APIView):
+    permission_classes = [IsAuthenticated]
+    # authentication_classes = [PersonalAccessTokenAuthentication]
+
+    def post(self, request):
+        string_format = request.query_params.get("string_format")
```
mathjazz: Can we just call the parameter `format`? It's not a string format, it's a resource format, but I think simply format makes it clear enough.

pontoon/api/views.py
```diff
@@ -459,3 +464,59 @@ def get_queryset(self):
...
+        locale = request.query_params.get("locale")
+
+        if not request.body:
+            raise ValidationError({"text": ["This field is required."]})
+
+        if len(request.body) > MAX_TEXT_BYTES:
+            raise ValidationError({"text": ["Payload too large."]})
```
mathjazz: Why do we need that limit? Do we have anything like that in the `pontoon.pretranslate` module? If not, why impose it here? If yes, why not just leave it to the pretranslation to handle errors?

pontoon/api/views.py
```diff
@@ -459,3 +464,59 @@ def get_queryset(self):
...
+        try:
+            text = request.body.decode("utf-8", errors="replace")
+        except Exception:
+            raise ValidationError({"text": ["Unable to decode request body as UTF-8."]})
+
+        errors = {}
+        if not text.strip():
+            errors["text"] = ["This field is required."]
+        elif len(text) > MAX_TEXT_CHARS:
+            errors["text"] = ["Text exceeds maximum length of 2048 characters."]
```
mathjazz: Same as above, why do we need to impose this limit?

pontoon/api/views.py
```diff
@@ -459,3 +464,59 @@ def get_queryset(self):
...
+        if not locale:
+            errors["locale"] = ["This field is required."]
+        if string_format and string_format not in {
+            Resource.Format.FLUENT,
+            Resource.Format.ANDROID,
+            Resource.Format.GETTEXT,
+        }:
+            errors["format"] = ["Choose a correct format."]
+        if errors:
+            raise ValidationError(errors)
```
mathjazz, line 493: Why are some errors raised immediatelly and why are some collected and then raised all together?

pontoon/api/views.py
```diff
@@ -459,3 +464,59 @@ def get_queryset(self):
...
+        if not text.strip():
+            errors["text"] = ["This field is required."]
+        elif len(text) > MAX_TEXT_CHARS:
+            errors["text"] = ["Text exceeds maximum length of 2048 characters."]
+        if not locale:
+            errors["locale"] = ["This field is required."]
+        if string_format and string_format not in {
+            Resource.Format.FLUENT,
+            Resource.Format.ANDROID,
+            Resource.Format.GETTEXT,
```
mathjazz: Let's allow all supported formats here.

pontoon/api/views.py
functionzz: Unfortunately, `format` already exists as a default parameter for input formats (json, text, etc.). I can change it to `resource_format` or something that sounds better.

pontoon/api/views.py
functionzz: I was thinking that we probably don't want the user to be able to input 1GB worth of plain text. Does pretranslation handle such cases?

pontoon/api/views.py
functionzz: Do the parsers check for max size?

pontoon/api/views.py
functionzz: This is a remnant of previous implementations of input validation for the API. I can collect the errors or file them immediately.

functionzz: I changed it such that all errors are collected.

pontoon/api/views.py
functionzz: I changed the parameter to `resource_format`

mathjazz: > Unfortunately, `format` already exists as a default parameter for input formats (json, text, etc.). I can change it to `resource_format` or something that sounds better.

Could you elaborate on this? I meant that the URL parameter should be called `format` and as far as I can tell we don't have any other URL parameter here (much less with the same name).

pontoon/api/views.py
```diff
@@ -459,3 +463,68 @@ def get_queryset(self):
...
+    Resource.Format.DTD,
+    Resource.Format.FLUENT,
+    Resource.Format.GETTEXT,
+    Resource.Format.INI,
+    Resource.Format.PLAIN_JSON,
+    Resource.Format.PROPERTIES,
+    Resource.Format.WEBEXT,
+    Resource.Format.XCODE,
+    Resource.Format.XLIFF,
+}
```
mathjazz: Let's avoid hardcoding the formats here (and updating them every time in two places when a new format is introduced):

```suggestion
VALID_FORMATS = set(Resource.Format)
```

pontoon/api/views.py
```diff
@@ -459,3 +463,68 @@ def get_queryset(self):
             raise ValueError(error)
 
         return entities
+
+
+MAX_TEXT_CHARS = 2048
```
mathjazz: We should make this a per-instance configurable setting, so we don't need to deploy a new version to change it and have it hardcoded for all instances.

pontoon/api/views.py
```diff
@@ -459,3 +463,68 @@ def get_queryset(self):
...
+            errors["text"] = ["Unable to decode request body as UTF-8."]
+            text = None
+
+        if text is not None:
+            if len(request.body) > MAX_TEXT_BYTES:
+                errors["text"] = ["Payload too large."]
+            elif len(text) > MAX_TEXT_CHARS:
+                errors["text"] = [
+                    f"Text exceeds maximum length of {MAX_TEXT_CHARS} characters."
+                ]
```
mathjazz: Why don't we just check the length of either `request.body` or `text`? Why check each of them?

pontoon/api/views.py
functionzz: By default, DRF has the `format` query parameter taken. For example, doing `format=json` will indicate json is being sent to the API. I can't use it because it already has included functionality.

pontoon/api/views.py
functionzz: It would be possible that the user inputs some sort of value whose character is less than 4 bytes like "A", "B" or "C" more times than the stated limit, which is 2048. For example, "A" is 1 byte large. "A" multiplied by 8000 is 8000 characters and 8000 bytes, less than the hardcap of `MAX_BYTES`, which is 2048 * 4 = 8192. This check is just to keep consistency with checking exactly how many chars we allow within `MAX_CHARS`

pontoon/api/views.py
mathjazz: Ohhhhhh, I see. Nevermind then.

pontoon/api/views.py
mathjazz: I'd keep it simple for now and only cap on the character length.

mathjazz:
The code looks good! 

Why is the PR maked as Draft? Is there any functionality still missing?

pontoon/api/views.py
```diff
@@ -459,3 +464,55 @@ def get_queryset(self):
             raise ValueError(error)
 
         return entities
+
+
+VALID_FORMATS = set(Resource.Format)
```
mathjazz: I don't think that we need to declare a constant anymore, I'd just use the set directly in the condition.

pontoon/api/views.py
```diff
@@ -459,3 +464,55 @@ def get_queryset(self):
...
+        if text is not None:
+            if len(text) > PRETRANSLATION_API_MAX_CHARS:
+                errors["text"] = [
+                    f"Text exceeds maximum length of {PRETRANSLATION_API_MAX_CHARS} characters."
+                ]
+            elif not text.strip():
+                errors["text"] = ["This field is required."]
+
+        if not locale:
+            errors["locale"] = ["This field is required."]
```
mathjazz, line 489: I wonder if we should also check if `Locale.google_automl_model` is set. That way Pretranslation API will only work for locales for which Pretranslation in Pontoon works, i.e. for the ones that have custom MT models.

@flodolo Thoughts?

flodolo:
> You can test the PR on [mozilla-pontoon-staging.herokuapp.com](https://mozilla-pontoon-staging.herokuapp.com):
> 
> ```python
> import requests
> 
> response = requests.post(
>     "https://mozilla-pontoon-staging.herokuapp.com/api/v2/pretranslate/?locale=sl",
>     data="Open",
>     headers={
>         "Content-Type": "text/plain",
>         "Accept": "application/json",
>         "Authorization": f"Bearer {TOKEN}",
>     },
> )
> ```

Noting that this works on Heroku, but not on GCP (CSRF validation fails).

pontoon/base/migrations/0102_add_pretranslators_group.py
flodolo, line 1: This file should have a better name. Also the PR needs a rebase at this point.

Issue #3835, opened 2025-11-18
Containerize Pontoon for MozCloud Deployment


bqbn:
To run applications in MozCloud, they must be containerized and their Docker images published to Google Artifact Registry (GAR).

This generally involves two steps:

1. Creating a Dockerfile to build the application’s Docker image.
2. Adding a GitHub workflow to build and push the image to GAR.

This ticket requests that the Pontoon application be containerized and prepared for deployment to MozCloud.

bqbn:
For pushing Docker images to GAR, you can use the actions available at: https://github.com/mozilla-it/deploy-actions/tree/main/docker-push or implement your own workflow logic.

We can provide the required values for `workload_identity_pool_project_number`, `project_id`, and etc. Typically, these values are stored as repository-level environment variables.

PR #3838, opened 2025-11-20
Containerize the app and push the image to GAR

bqbn:
Fix https://github.com/mozilla/pontoon/issues/3835

### DIFF
--- a/docker/Dockerfile-mozcloud
+++ b/docker/Dockerfile-mozcloud
@@ -34,7 +34,7 @@ WORKDIR /app
 # Copy root package.json for workspace configuration
 COPY package.json package-lock.json /app/
 
-# Install pontoon Node.js dependencies (terser, yuglify for compression)
+# Copy pontoon Node.js dependencies (terser, yuglify for compression)
 COPY pontoon/package.json /app/pontoon/package.json
 
 # Install translate frontend dependencies
@@ -51,8 +51,12 @@ RUN npm run build:prod
 # Stage 3: Static files builder
 FROM python:3.11-slim-bookworm AS static-builder
 
+# SECRET_KEY and DATABASE_URL are set only to prevent collectstatic from
+# failing—their actual values don’t matter. collectstatic loads settings/base.py,
+# which requires SECRET_KEY and DATABASE_URL to be set.
 ARG SECRET_KEY=foo
 ARG DATABASE_URL=postgres://pontoon:asdf@postgresql/pontoon
+
 ARG GIT_REVISION=unknown
 
 ENV PYTHONUNBUFFERED=1


docker/Dockerfile-mozcloud
```diff
@@ -0,0 +1,154 @@
...
+
+# Copy translate source code and build production frontend
+COPY translate/ /app/translate/
+RUN npm run build:prod
+
+
+# Stage 3: Static files builder
+FROM python:3.11-slim-bookworm AS static-builder
+
+ARG SECRET_KEY=foo
```
flodolo, line 57: How should authentication data be managed in these cases?

.gitignore
```diff
@@ -1,4 +1,5 @@
 .*
+!.github
```
flodolo, line 2: 🤦🏼

docker/Dockerfile-mozcloud
bqbn: This is the key Django uses to protect its sessions. It isn’t used during the Docker build process, but it is required by the `collectstatic` command.

Are you asking how Docker authenticates to GCP GAR? That authentication is handled through the GCP Workload Identity Pool mechanism, as described in this action: https://github.com/mozilla-it/deploy-actions/tree/main/docker-push

.gitignore
bqbn: I’m not entirely sure why `.*` was included originally, so I didn’t remove it.

However, I do think it’s generally good practice to track changes in the `.github` directory.

docker/Dockerfile-mozcloud
flodolo: Thanks, I didn't know what this was used for. My comment was actually referring to the whole group (the next 2 lines are for postgres authentication).

.gitignore
flodolo: Agreed. We had a bunch of hidden folders to exclude, but went too far in https://github.com/mozilla/pontoon/pull/3401

docker/Dockerfile-mozcloud
bqbn: Both environment variables (`SECRET_KEY` and `DATABASE_URL`) are set only to ensure that the `collectstatic` command doesn’t fail—their actual values don’t matter. 

This is because `collectstatic` loads `settings/base.py`, where `SECRET_KEY` is required (otherwise [line 30](https://github.com/mozilla/pontoon/blob/main/pontoon/settings/base.py#L30) fails), and `DATABASE_URL` is needed to avoid an error in `dj_database_url.config` ([line 55](https://github.com/mozilla/pontoon/blob/main/pontoon/settings/base.py#L55)).

docker/Dockerfile-mozcloud
```diff
@@ -0,0 +1,154 @@
...
+
+# Stage 2: Node.js frontend builder
+FROM node:24-bookworm-slim AS node-builder
+
+WORKDIR /app
+
+# Copy root package.json for workspace configuration
+COPY package.json package-lock.json /app/
+
+# Install pontoon Node.js dependencies (terser, yuglify for compression)
```
mathjazz: Nit (same for translate deps below):

```suggestion
# Copy pontoon Node.js dependencies (terser, yuglify for compression)
```

docker/Dockerfile-mozcloud
mathjazz: A comment would be helpful to avoid confusing other people (or a self-explanatory key value).

docker/Dockerfile-mozcloud
```diff
@@ -0,0 +1,154 @@
...
+STOPSIGNAL SIGINT
+CMD ["gunicorn", "pontoon.wsgi:application",    \
+     "--bind", "0.0.0.0:8000",                  \
+     "--workers", "${GUNICORN_WORKERS:-4}",     \
+     "--worker-class", "sync",                  \
+     "--worker-tmp-dir", "/dev/shm",            \
+     "--log-file", "-",                         \
+     "--access-logfile", "-",                   \
+     "--error-logfile", "-",                    \
+     "--log-level", "info"]
```
mathjazz: Please add a blank line.

docker/Dockerfile-mozcloud
bqbn: Yep, and added a comment.

docker/Dockerfile-mozcloud
bqbn: Done.

docker/Dockerfile-mozcloud
bqbn: Fixed.

PR #3845, opened 2025-11-25
Configured vitest similar to jest

nishitmistry:
Addresses https://github.com/mozilla/pontoon/issues/3768

### Change Summary:
- Added vitest, @vitest/coverage-v8 coverage provider packages
- Added temporary vitest command  and vitest ci job which would continue on error 
- Configured vitest along with coverage options
- renamed the file which required jsx extension

From the project root directory  run this command to check vitest `cd translate && npm run vitest`


package.json
```diff
@@ -55,7 +55,7 @@
     "@types/linkify-it": "^3.0.1",
     "@types/lodash.escaperegexp": "^4.1.6",
     "@types/lodash.isempty": "^4.4.6",
-    "@types/node": "^15.3.0",
+    "@types/node": "^24.10.1",
```
eemeli: Why is this included here? It odesn't seem to have anything to do with the rest of the PR, and it's not aligned e.g. with our CI tests, which are currently on Node.js 22. This change should be left out.

translate/package.json
```diff
@@ -46,6 +46,7 @@
     "build": "rollup -c",
     "build:prod": "rollup -c --environment BUILD:production",
     "test": "jest",
+    "vitest": "vitest run",
```
eemeli: If this command is temporary and isn't run by CI, it should be left out -- it's just as easy to type
```
npx vitest run
```
as it is to type
```
npm run vitest
```

translate/vitest.config.ts
```diff
@@ -0,0 +1,28 @@
+import path from 'path';
+import { defineConfig } from 'vitest/config';
+import react from '@vitejs/plugin-react';
+
+export default defineConfig({
+  plugins: [react()],
+  test: {
+    globals: true,
+    coverage: {
+      enabled: true,
```
eemeli: Let's default to disabling coverage. I know that's different from Jest, but it makes the output unnecessarily verbose and it's easy to enable with `--coverage`.

translate/vitest.config.ts
```diff
@@ -0,0 +1,28 @@
...
+      reportOnFailure: true,
+      include: ['src/**/*.{js,jsx,ts,tsx}'],
+      exclude: ['src/**/*.d.ts', 'node_modules/**'],
```
eemeli: We don't have any `.d.ts` files in `src/`, and `node_modules/` isn't in `src/`.

translate/vitest.config.ts
```diff
@@ -0,0 +1,28 @@
...
+    mockReset: true,
+    testTimeout: 10000,
+  },
+  resolve: {
+    alias: [
+      { find: /^~(.*)$/, replacement: path.resolve(__dirname, 'src/$1') },
+      {
+        find: /\.svg$/,
+        replacement: path.resolve(__dirname, '__mocks__/svg.js'),
+      },
```
eemeli: This is an ESM file, so these should be using `import.meta.dirname` instead.

translate/vitest.config.ts
nishitmistry: i'll remove both

translate/vitest.config.ts
nishitmistry: changed to `import.meta.dirname`

package.json
nishitmistry: i had to change it since vitejs react plugin had some peer dependency issues with type/node ^15.3.0.
i have change the version to ^22.19.1 matching our ci node js version

translate/package.json
nishitmistry: i have added vitest workflow, can you please review it

.github/workflows/frontend.yml
```diff
@@ -68,3 +68,18 @@ jobs:
...
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+
+  vitest:
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v5
+        - uses: actions/setup-node@v5
+          with: { node-version: '22' }
+        - name: Install globals
+          run: npm install --global npm@8
```
eemeli: Why are you downgrading npm? Node.js 22 comes with npm 10, why not use that?

```suggestion
```

.github/workflows/frontend.yml
```diff
@@ -68,3 +68,18 @@ jobs:
...
+        - name: Install dependencies
+          run: npm ci
+        - name: vitest
+          run: npm run vitest -- --coverage
```
eemeli: This command will eventually be run by `npm test`, and so we don't need a separate script for it.

```suggestion
          run: npx vitest run --coverage
```

translate/package.json
eemeli: As I suggest in the workflow file, we don't need a temporary npm script for this.

```suggestion
```

.github/workflows/frontend.yml
nishitmistry: thanks for mentioning this, i missed removing it

.github/workflows/frontend.yml
nishitmistry: yes

translate/package.json
nishitmistry: removed the unnecessary command

PR #3847, opened 2025-11-26
Do not assume that Android plural source will always have a "one" variant

eemeli:


mathjazz (approved):
Changes look good! I tested locally and it works fine.

PR #3848, opened 2025-11-26
Use MF2 syntax for webext strings

eemeli:
This is effectively a prerequisite for #3802, as it despecializes our handling of webext strings. It follows closely the patterns used in #3790, expanding them to apply to webext in addition to Android strings.

Some additional validation is added, allowing us to catch typos in webext placeholders the same way as we do for Android. The editor highlighting of webext messages is also improved.


pontoon/checks/tests/test_custom.py
```diff
@@ -258,3 +260,69 @@ def test_android_bad_html():
...
+    translation = "Translation with a $FOO$"
+    entity = mock_entity("webext", string=original)
+    assert run_custom_checks(entity, translation) == {}
+
+
+def test_webext_extra_index_placeholder():
+    original = "Source string"
+    translation = "Translation with a $1"
+    entity = mock_entity("webext", string=original)
+    # This should probably also be caught
```
mathjazz, line 306: What is the reason it is not?

eemeli: We have no strings using indexed rather than named variables in production, and this will get refactored with the upcoming wider changes to validation/linting, so adding the complexity needed for this isn't worthwhile atm.

Issue #3148, opened 2024-03-18
String displayed even if not matching filters


flodolo:
I assume it's a fall-out of #3128 

Take for example this: https://pontoon.mozilla.org/it/mozillaorg/all-resources/?status=missing,pretranslated&string=301402
The string in question is neither missing nor pretranslated, but it's still displayed.

eemeli:
Here's an amusing way to see what this enables: If you go to https://pontoon.mozilla.org/fr/firefox/all-resources/?string=50 and then click on the first entry in the context, you end up with the UI in this state:

![Image](https://github.com/user-attachments/assets/92297de0-7050-4fef-a407-ffbd7087a37e)

Note that navigating to the corresponding URL does not work.

PR #3851, opened 2025-11-27
Show obsolete entities as such

eemeli:
It's currently possible to end up looking at an obsolete entity in Pontoon when using its entity identifier directly as a `string` query parameter. When this happens, we should highlight for the user that the string they're looking at is not in active use, and ensure that it's treated as readonly.

Some cleanup of the previous/next entity hooks is included.


mathjazz:
> It's currently possible to end up looking at an obsolete entity in Pontoon when using its entity identifier directly as a `string` query parameter.

Should we keep allowing that, or prevent it completely as we'll do when https://github.com/mozilla/pontoon/issues/3148 is fixed?

PR #3853, opened 2025-11-30
Settings page: Remove save button, blur input and change APIs

### LINKED ISSUE #3746: Drop the Save button from the Settings page
Most of the widgets in the /settings page save changes immediately. Text fields and the Preferred locales selector require clicking the Save button. That's conunter-intuitive. Let's drop the Save button and save everything immediately (with a slight delay in the case of text fields) after the changes are made.

### LINKED ISSUE #3844: Migrate api/v1/user/<username:username>/theme/ and api/v1/user/<username:username>/ to REST API
The endpoints `api/v1/user/<username:username>/theme/ ` and `api/v1/user/<username:username>/`, located [here](https://github.com/mozilla/pontoon/blob/8f71af1ead04cec9eb3f13532a6c4bf147eb0d0a/pontoon/contributors/urls.py#L65) need to be deprecated and replaced with an `api/v2` equivalent.

functionzz:
## Description
This PR removes the html form functionality of the settings page and replaces it with auto-save with debouncing. It also changes APIv1 endpoints to reflect similar AJAX calls, and modifies `multiple_locale_selector` behaviour.

Fixes #3746 
Fixes #3844 

## TODO:
- [x] ~Discuss email field debouncing issue~
- [x] ~Locale Selector auto-save is not finished~ 
- [x] ~Remove Save button~
- [x] ~Test cases need to be rewritten due to behavior change~

## Additional Notes
~I noticed that the `multiple_locale_selector` and the `preferred_locale_selector` et. al still includes locales like `bn_BD` and `bn-IN`.~ See #3854.

~Also, one concern I have for this implementation lies with the email field. If a user wants, they can spam random emails with verification mail by changing a letter of the input every 2 seconds. Also, if a user doesn't finish writing out said email within 2 seconds, it will submit due to the debouncing, accidentally sending an email to someone else.~

### DIFF
--- a/pontoon/contributors/static/js/settings.js
+++ b/pontoon/contributors/static/js/settings.js
@@ -31,6 +31,7 @@ $(function () {
       return;
     }
 
+    this.dataset.originalValue = value;
     $.ajax({
       url: '/user/attributes/field/',
       type: 'POST',
@@ -38,9 +39,7 @@ $(function () {
         csrfmiddlewaretoken: $('body').data('csrf'),
         [attribute]: value,
       },
-      success: () => {
-        this.dataset.originalValue = value;
-
+      success() {
         // contact_email special case
         if (attribute === 'contact_email') {
           if (value !== '') {
@@ -56,7 +55,8 @@ $(function () {
         const message = 'Settings saved.';
         Pontoon.endLoader(message);
       },
-      error(response) {
+      error: (response) => {
+        this.dataset.originalValue = originalValue;
         // contact_email special case
         if (attribute === 'contact_email') {
           self.parents('.field').find('.help').addClass('hide');


pontoon/contributors/templates/contributors/settings.html
```diff
@@ -61,7 +84,7 @@ <h3>Email</h3>
                     communications, please do so here. Note, this email will also appear under your <a
                         href="{{ url('pontoon.contributors.profile') }}">Profile page</a>.</p>
                 {% endif %}
-            </div>
+            </div> -->
```
mathjazz: Why is the old field commented out, rather than removed?

pontoon/base/templates/widgets/field.html
```diff
@@ -0,0 +1,25 @@
...
+    {{ field.label_tag(label_suffix='') }}
+    <textarea class="input" 
+           data-attribute="{{ attribute }}" 
+           data-original-value="{{ value or '' }}"
+           title="{{ title }}"
+           name="{{ field.name }}" 
+           id="{{ field.auto_id }}"
+           {% if field.field.max_length %}maxlength="{{ field.field.max_length }}"{% endif %}>{{ value or '' }}</textarea>
+    <div class="errorlist"></div>
+{% endmacro %}
```
mathjazz, line 25: It's not clear to me what is the benefit of using a custom widget here compared to what we had before. Other than that it looks good!

functionzz: The Field.text and Field.textarea define 2 separate behaviours for the fields. As far as I know, I need to redefine the field functionality because its not entirely the same as a normal form field. But most things from the original form field implementation are the same as the Field macro, so if the functionality was written with <input> or <textarea> it would need to look something like this each time:

```
<input class="input"
       value="{{ value|default_if_none:'' }}"
       data-attribute="{{ attribute }}"
       data-original-value="{{ value|default_if_none:'' }}"
       title="{{ title }}"
       name="{{ field.name }}"
       id="{{ field.auto_id }}"
       type="text"
       {% if field.field.max_length %}maxlength="{{ field.field.max_length }}"{% endif %}>
<div class="errorlist"></div>
```

So in short I made the macro because I needed to copy the original form field functionality, but it wasn't entirely the same.

pontoon/contributors/static/js/settings.js
```diff
@@ -1,4 +1,118 @@
...
+      }
+    });
+
+  $('.field .input').on('blur', function (e) {
+    e.preventDefault();
+
+    const self = $(this);
+    const value = self.val().trim();
+    const attribute = self.data('attribute');
+    const originalValue = self.data('original-value') || '';
```
eemeli: What's the reason for using `self.data` rather than `this.dataset`? Values in the latter are automatically reflected in the element attributes, and so wouldn't need the duplicate setters that you have below.

In general, where the functionality overlaps, we should prefer DOM over jQuery.

pontoon/contributors/static/js/settings.js
```diff
@@ -1,4 +1,118 @@
...
+    $.ajax({
+      url:
+        '/user/' +
+        $('#profile input[name="username"]').val() +
+        '/attributes/field/',
+      type: 'POST',
+      data: {
+        csrfmiddlewaretoken: $('body').data('csrf'),
+        attribute: attribute,
+        [attribute]: value || '',
```
eemeli: `value` is set from a `.trim()` result, so if it's falsy, it must be `''`.
```suggestion
        [attribute]: value,
```

pontoon/contributors/static/js/settings.js
```diff
@@ -1,4 +1,118 @@
...
+        }
+
+        const errors = response.responseJSON.errors;
+
+        const errorContainer = self.parents('.field').find('.errorlist');
+        errorContainer.empty();
+        for (const error in errors) {
+          const errorMessages = errors[error].join(', ');
+          errorContainer.append($('<p>').text(errorMessages));
+        }
```
eemeli: It's almost never the right choice to use `for...in` iteration.

```suggestion
        const errorContainer = self.parents('.field').find('.errorlist');
        errorContainer.empty();
        for (const error of response.responseJSON.errors) {
          errorContainer.append($('<p>').text(error.join(', ')));
        }
```

pontoon/contributors/static/js/settings.js
```diff
@@ -1,4 +1,118 @@
...
+    'click',
+    '.multiple-team-selector .locale.select li',
+    function () {
+      saveLocalesOrder();
+    },
+  );
+
+  $('body').on('click', '.multiple-team-selector .move-all', function () {
+    saveLocalesOrder();
+  });
```
eemeli: Remember that JS has first-class functions.

```suggestion
  $('body').on(
    'click',
    '.multiple-team-selector .locale.select li',
    saveLocalesOrder,
  );

  $('body').on('click', '.multiple-team-selector .move-all', saveLocalesOrder);
```

pontoon/contributors/static/js/settings.js
```diff
@@ -1,4 +1,118 @@
...
+        '/user/' +
+        $('#profile input[name="username"]').val() +
+        '/attributes/field/',
+      type: 'POST',
+      data: {
+        csrfmiddlewaretoken: $('body').data('csrf'),
+        attribute: attribute,
+        [attribute]: value || '',
+      },
+      success: function () {
```
eemeli: Stylistic nit, this also applies to multiple other cases further down.

```suggestion
      success() {
```

pontoon/contributors/static/js/settings.js
```diff
@@ -1,4 +1,118 @@
...
+
+    if (value === originalValue) {
+      return;
+    }
+
+    $.ajax({
+      url:
+        '/user/' +
+        $('#profile input[name="username"]').val() +
+        '/attributes/field/',
```
eemeli: This is duplicated often enough that it'd be a good idea to add a utility function for it, something like:
```js
function userAttributeUrl(action) {
  const username = $('#profile input[name="username"]').val();
  return `/user/${username}/attributes/${action}/`;
}
```

pontoon/contributors/static/js/settings.js
```diff
@@ -102,11 +219,12 @@ $(function () {
...
 
-        $('.error-message').html('');
+        const errorContainer = $('#pat-settings').find('.error-message');
+        errorContainer.empty();
 
         for (const error in errors) {
           const errorMessages = errors[error].join(', ');
-          $(`.error-message`).append($('<p>').text(errorMessages));
+          errorContainer.append($('<p>').text(errorMessages));
         }
```
eemeli: See comment above about `for...in`. This code style may be from a time when `for...of` wasn't generally available.

pontoon/contributors/urls.py
```diff
@@ -62,18 +62,30 @@ class UsernameConverter(StringConverter):
         views.mark_all_notifications_as_read,
         name="pontoon.contributors.notifications.mark.all.as.read",
     ),
-    # API: Toggle user theme preference
+    # AJAX: Toggle user theme preference
```
eemeli: None of the new API endpoints are AJAX, even if they're accessed via `$.ajax()`; recall (or research) what the last X stands for.

pontoon/contributors/views.py
```diff
@@ -180,6 +180,90 @@ def update_contribution_timeline(request):
...
+    value = request.POST.get(attribute, None)
+
+    if attribute not in text_attributes:
+        return JsonResponse(
+            {"status": False, "message": "Forbidden: Attribute not allowed"},
+            status=403,
+        )
+
+    profile = user.profile
+    if attribute in text_attributes:
```
eemeli: This is never false.

pontoon/contributors/views.py
```diff
@@ -180,6 +180,90 @@ def update_contribution_timeline(request):
...
+    ]
+
+    attribute = request.POST.get("attribute", None)
+    value = request.POST.get(attribute, None)
+
+    if attribute not in text_attributes:
+        return JsonResponse(
+            {"status": False, "message": "Forbidden: Attribute not allowed"},
+            status=403,
+        )
```
eemeli: The request's `attribute` field is unnecessary, and needlessly limits the API to only setting one field at a time.

We could/should instead do something like:
```python
for attribute, value in request.POST.items():
    match attribute:
        case "first_name":
            ...
        case "contact_email":
            ...
        case "username" | "bio" | "chat" | "github" | "bugzilla":
            ...
        case _:
            ...
```

pontoon/contributors/views.py
```diff
@@ -250,6 +334,40 @@ def toggle_user_profile_attribute(request, username):
...
+                "status": False,
+                "message": "Forbidden: You don't have permission to edit this user",
+            },
+            status=403,
+        )
+    profile = user.profile
+    locales_form = forms.UserLocalesOrderForm(
+        request.POST,
+        instance=profile,
+    )
```
eemeli:
```suggestion
    locales_form = forms.UserLocalesOrderForm(
        request.POST,
        instance=user.profile,
    )
```

pontoon/contributors/static/js/settings.js
functionzz: I changed it to `for (const error of Object.values(response.responseJSON.errors))` as `response.responseJSON.errors` is a dictionary.

pontoon/contributors/views.py
functionzz: To be clear, the behaviour will still in practice be that the API is called once every time the field is saving (i.e when the blur happens).

pontoon/contributors/urls.py
functionzz: I wanted to follow the existing comment pattern in pontoon/contributors/urls.py. Should I change all of them to something like "API: ______"?

pontoon/contributors/views.py
functionzz: OK, I applied the changes in that I removed the redundant attribute field and replaced the view logic with a match statement that functionally has a similar behavior to the old code.

pontoon/contributors/urls.py
eemeli: Yeah, that sounds like a good idea. Though could also drop some of these, as a comment like "Save custom homepage" on an endpoint `save-custom-homepage/` that's mapped to a view `save_custom_homepage` and a name `pontoon.contributors.save_custom_homepage` is rather completely useless.

pontoon/contributors/static/js/settings.js
```diff
@@ -13,27 +18,23 @@ $(function () {
...
-    const attribute = self.data('attribute');
-    const originalValue = self.data('original-value') || '';
+    const attribute = this.dataset.attribute;
+    const originalValue = this.dataset.originalValue || '';
 
     if (value === originalValue) {
       return;
     }
 
+    this.dataset.originalValue = value;
```
eemeli: Isn't this presuming success? If you get an error response from the POST, shouldn't the original value be retained?

pontoon/contributors/views.py
```diff
@@ -194,71 +194,65 @@ def edit_user_profile_fields(request, username):
...
-                        "status": "error",
-                        "message": "Form validation failed.",
-                        "errors": user_form.errors,
-                    },
-                    status=400,
+    for attribute, value in request.POST.items():
+        if attribute == "csrfmiddlewaretoken":
+            continue
+
+        match attribute:
```
eemeli:
```suggestion
        match attribute:
            case "csrfmiddlewaretoken":
                continue
```

pontoon/contributors/static/js/settings.js
functionzz: The originalValue check is mainly to stop users from resaving the the same value after editing the field. For example, for the username if you type in `hello`, then presses ENTER multiple times, it should fire AJAX once.

Indeed, if you get an error response from the POST, originalValue saves as the malformed input, but for all intents and purposes that does not change the functionality of the field.

eemeli: That makes sense as a way to stop multiple submissions before a request has been handled, but on error, the value should be reset to `originalValue`.

pontoon/contributors/views.py
```diff
@@ -180,6 +180,83 @@ def update_contribution_timeline(request):
     )
 
 
+@login_required(redirect_field_name="", login_url="/403")
+@require_POST
+@transaction.atomic
+def edit_user_profile_fields(request, username):
+    user = get_object_or_404(User, username=username)
+    if user != request.user:
```
eemeli: Hang on, why do we include the username as a query parameter? Why not just use `request.user`?

functionzz: It seems I copid the existing pattern for `/user/<username>/theme`. I'll remove both cases.

pontoon/contributors/static/js/settings.js
```diff
@@ -1,7 +1,6 @@
 $(function () {
   function userAttributeURL(action) {
-    const username = $('#profile input[name="username"]').val();
-    return `/user/${username}/attributes/${action}/`;
+    return `/user/attributes/${action}/`;
   }
```
eemeli: With this simplification, this function should be dropped and plain strings used as `url` values.

pontoon/contributors/static/js/settings.js
```diff
@@ -22,20 +21,29 @@ $(function () {
...
+        } else {
+          self.parents('.field').find('.help').removeClass('hide');
+          self.parents('.field').find('.verify').addClass('hide');
+        }
+      }
       return;
     }
 
-    this.dataset.originalValue = value;
-
```
eemeli: You had the right idea here, this should still happen before the API call is made. Remember that it may take time for the request to go to the server and come back, which is when the `success()` handler will be called. During that time, having the `dataset.originalValue` set will allow duplicate requests to be avoided.

The desired change from the previous is that in the `error()` handler you should include something like:
```js
this.dataset.originalValue = originalValue;
```

pontoon/contributors/views.py
```diff
@@ -183,16 +183,8 @@ def update_contribution_timeline(request):
...
-    if user != request.user:
-        return JsonResponse(
-            {
-                "status": False,
-                "message": "Forbidden: You don't have permission to edit this user",
-            },
-            status=403,
-        )
+def edit_user_profile_fields(request):
+    user = request.user
```
eemeli, line 187: There are a few places below that should refer to `user` rather than `request.user`.

pontoon/contributors/views.py
```diff
@@ -260,17 +252,8 @@ def edit_user_profile_fields(request, username):
...
-        return JsonResponse(
-            {
-                "status": False,
-                "message": "Forbidden: You don't have permission to edit this user",
-            },
-            status=403,
-        )
-
+def toggle_user_profile_attribute(request):
+    user = request.user
```
eemeli: This is unnecessary, the only usage is on line 302 below, which should be updated.

pontoon/contributors/views.py
```diff
@@ -330,16 +313,8 @@ def toggle_user_profile_attribute(request, username):
...
-    if user != request.user:
-        return JsonResponse(
-            {
-                "status": False,
-                "message": "Forbidden: You don't have permission to edit this user",
-            },
-            status=403,
-        )
+def edit_user_profile_locale_selector(request):
+    user = request.user
```
eemeli: This is unnecessary, the only usage is on line 320 below, which should be updated.

pontoon/contributors/views.py
```diff
@@ -363,17 +338,8 @@ def edit_user_profile_locale_selector(request, username):
...
-        return JsonResponse(
-            {
-                "status": False,
-                "message": "Forbidden: You don't have permission to edit this user",
-            },
-            status=403,
-        )
-
+def toggle_theme(request):
+    user = request.user
```
eemeli: This is unnecessary, the only usage is on line 346 below, which should be updated.

translate/src/api/user.ts
```diff
@@ -98,7 +98,7 @@ export function updateUserSetting(
     csrfmiddlewaretoken: csrfToken,
   });
   const headers = new Headers({ 'X-CSRFToken': csrfToken });
-  return POST(`/api/v1/user/${username}/`, payload, { headers });
+  return POST(`/user/attributes/toggle/`, payload, { headers });
```
eemeli:
```suggestion
  return POST('/user/attributes/toggle/', payload, { headers });
```

translate/src/api/user.ts
```diff
@@ -111,7 +111,7 @@ export function updateUserTheme(
     csrfmiddlewaretoken: csrfToken,
   });
   const headers = new Headers({ 'X-CSRFToken': csrfToken });
-  return POST(`/api/v1/user/${username}/theme/`, payload, { headers });
+  return POST(`/user/theme/`, payload, { headers });
```
eemeli:
```suggestion
  return POST('/user/theme/', payload, { headers });
```

eemeli (changes requested):

PR #3854, opened 2025-12-01
Change locale visibility

### LINKED ISSUE #3700: Hide teams not enabled for any non-system and non-disabled project
https://pontoon.mozilla.org/bn-BD/ and https://pontoon.mozilla.org/bn-IN/ are both shown on the /teams page and their respective team dashboards, despite not being enabled for any any non-system and non-disabled project.

functionzz:
## Description
This PR changes the behaviour of Pontoon to exclude locales that have at least one `TranslatedResource` defined but that `TranslatedResource` belongs to a system project. This primarily affects locales `bn-BD` and `bn-IN`, which are currently displayed in multiple areas, despite having no activity. These locations include, `/settings`, `/messaging`, `/teams`, `/bn-BD`, `/bn-IN`, `/machinery`, `/translate`(machinery), `/api/v2/locales`.

Fixes #3700.

### DIFF
--- a/pontoon/administration/views.py
+++ b/pontoon/administration/views.py
@@ -273,13 +273,11 @@ def manage_project(request, slug=None, template="admin_project.html"):
 
     projects = sorted([p.name for p in Project.objects.all()])
 
-    locales_available = (
-        Locale.objects.visible()
-        .exclude(pk__in=locales_readonly)
-        .exclude(pk__in=locales_selected)
+    locales_available = Locale.objects.exclude(pk__in=locales_readonly).exclude(
+        pk__in=locales_selected
     )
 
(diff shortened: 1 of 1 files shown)


pontoon/administration/forms.py
eemeli, line 1: Why also filter the admin view locales?

functionzz: I guess that doesn't make sense. Reverting change.

pontoon/administration/views.py
eemeli, line 1: Shouldn't changes here also be reverted?

functionzz: Yes indeed.

PR #3855, opened 2025-12-01
Use MF2 syntax for xcode & xliff strings

eemeli:
~Depends on #3848 and mozilla/moz-l10n#116; tests will fail until the latter is merged and a new moz.l10n version included here~.

The preview code is generalised here for use in all MF2-ish formats, not just Android. Otherwise the code changes follow the same shape as used for previous MF2 migrations.

Technically, this PR adds XCode plural support to Pontoon, but that has not been validated yet, and we'll want to add [string catalog](https://developer.apple.com/documentation/xcode/localizing-and-varying-text-with-a-string-catalog) support before advertising the functionality.


eemeli:
Huh, looks like I managed to leave out the dependencies on mozilla/moz-l10n#116 from this PR, so that update can and should be handled separately. This is still blocking on #3848, though.

pontoon/base/migrations/0102_xliff_as_mf2.py
```diff
@@ -0,0 +1,82 @@
...
+    translations = Translation.objects.filter(
+        entity__resource__format__in=["xliff", "xcode"]
+    )
+    trans_fixed = [t for t in translations if xliff_string_changed(t)]
+    n = Translation.objects.bulk_update(trans_fixed, ["string"], batch_size=10_000)
+    print(f" ({n} translations)", end="", flush=True)
+
+
+class Migration(migrations.Migration):
+    dependencies = [("base", "0101_webext_as_mf2")]
```
mathjazz:
```suggestion
    dependencies = [("base", "0102_add_pretranslators_group")]
```

We'll also need to update the filename.

mathjazz (approved):
Looks good! I tested and didn't bump into any issues.

PR #3869, opened 2025-12-17
Fix misleading CSRF_TRUSTED_ORIGINS variable description

cachiusa:
https://github.com/mozilla/pontoon/blob/22aa0af2748375a88e8ad517fbf427dfeaee2440/pontoon/settings/base.py#L832

The code expects a string like `"a,b,c”`, not a list `['a','b','c']`

### DIFF
--- a/docs/admin/deployment.rst
+++ b/docs/admin/deployment.rst
@@ -290,10 +290,12 @@ you create:
    Learn more in the `Django documentation <https://docs.djangoproject.com/en/5.1/ref/settings/#secure-ssl-redirect>`_.
 
 ``CSRF_TRUSTED_ORIGINS``
-   Optional. A list of trusted origins for unsafe requests. It should contain the domains
-   where the app is available, seperated by comma. The setting also supports subdomains, so you could
-   add `https://*.example.com`, for example, to allow access from all
-   subdomains of `example.com`. Default value is `""`.
+   Optional. A comma-separated list of trusted origins for unsafe requests.
+   It should contain the domains where the app is available.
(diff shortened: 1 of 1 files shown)


docs/admin/deployment.rst
```diff
@@ -291,9 +291,9 @@ you create:
 
 ``CSRF_TRUSTED_ORIGINS``
    Optional. A list of trusted origins for unsafe requests. It should contain the domains
```
eemeli: Splitting up this suggestion because GitHub can't deal with it at once.

We should match the other list descriptions.
```suggestion
   Optional. A comma-separated list of trusted origins for unsafe requests.
```

docs/admin/deployment.rst
```diff
@@ -291,9 +291,9 @@ you create:
...
-   where the app is available. The setting also supports subdomains, so you could
+   where the app is available, seperated by comma. The setting also supports subdomains, so you could
    add `https://*.example.com`, for example, to allow access from all
```
eemeli:
```suggestion
   It should contain the domains where the app is available.
   The setting also supports subdomains,
   so you could add ``https://*.example.com``, for example,
```

docs/admin/deployment.rst
```diff
@@ -291,9 +291,9 @@ you create:
...
-   subdomains of `example.com`. Default value is `[]`.
+   subdomains of `example.com`. Default value is `""`.
```
eemeli:
```suggestion
   to allow access from all subdomains of ``example.com``.
   Default value is an empty string.
```

Issue #2340, opened 2021-10-21
Exclude disabled projects from the notifications page


mathjazz:
User notifications page (/notifications) lists all notifications the user has received. 

We should only list notifications from projects that are still enabled and remove the disabled projects from the filter in the sidebar.

PR #3871, opened 2025-12-18
Notifications: prevent translation notifications from disabled projects

### LINKED ISSUE #3714: Should not send notifications for disabled projects
Looks like the gettext update created a bunch of changes, which was a bit confusing.

Image

Even more confusing is that one of projects is disabled (MDN), so the link is broken.

functionzz:
## Description
This fix prevents translation updates (approvals, rejections etc.) from disabled projects to showing up in the notification feed in the translate module (bell) and the notifications page.

Fixes #3714 and #2340.


eemeli (approved):
Code looks good. Did not test manually.

PR #3875, opened 2025-12-23
Ported editor module test to vitest

nishitmistry:
In continuation of the previous Mr https://github.com/mozilla/pontoon/pull/3845 for task   https://github.com/mozilla/pontoon/issues/3768

- Added continue on error in frontend ci since jest test case for the migrated test will be failing
- Added JsDom environment similar to the previous jest config 
- updated broken tests under modules/editor directory to vitest implementation

### DIFF
--- a/translate/src/modules/editor/components/Editor.test.jsx
+++ b/translate/src/modules/editor/components/Editor.test.jsx
@@ -155,7 +155,9 @@ describe('<Editor>', () => {
   });
 
   it('renders the source form when passing a broken string', () => {
+    const spy = vi.spyOn(console, 'warn').mockImplementation(() => {});
     const [wrapper] = mountEditor(6);
+    expect(spy).toHaveBeenCalledWith(expect.stringContaining('E0028'));
 
     const input = wrapper.find(EditField);
     expect(input).toHaveLength(1);


.github/workflows/frontend.yml
```diff
@@ -63,6 +63,7 @@ jobs:
       - name: Test
         run: npm test --coverage
         working-directory: translate
+        continue-on-error: true
```
eemeli: With this change, our CI will stop failing for broken tests. That's not acceptable. When migrating tests from Jest to Vitest, the relevant configs should be updated so that Jest no longer runs the affected tests, but Vitest does, and fails in CI if those tests are broken.

translate/src/modules/editor/components/Editor.test.jsx
```diff
@@ -211,7 +212,9 @@ describe('<Editor>', () => {
   });
 
   it('passes a reconstructed translation to sendTranslation', async () => {
-    sinon.stub(TranslationAPI, 'createTranslation').returns({});
```
eemeli, line 214: If `sinon` is no longer used here, its import needs to be dropped.

translate/src/modules/editor/components/EditorMainAction.test.jsx
```diff
@@ -1,6 +1,5 @@
-import Fluent from '@fluent/react';
 import { mount } from 'enzyme';
-import React from 'react';
```
eemeli, line 3: We're still stuck on React v16 and our [Babel config](https://github.com/mozilla/pontoon/blob/main/babel.config.json) doesn't specify a runtime, so this import should not be dropped in a .jsx file.

.github/workflows/frontend.yml
nishitmistry: i have remove continue on error from both the test and added a temp config file which will follow migrated test paths such that jest ignore this test and vitest picks them up, never breaking the feedback provided by test. please let me know if there's any issue in this approach.

translate/src/modules/editor/components/Editor.test.jsx
nishitmistry: removed

translate/src/modules/editor/components/EditorMainAction.test.jsx
nishitmistry: added React import

.github/workflows/frontend.yml
nishitmistry: @eemeli i had a concern that ignoring test will also decrease the total coverage percentage, is that fine or should i try to merge coverage of both the test runner ?

eemeli: We can and should ignore coverage during the transition.

nishitmistry: ohk thank,  then i think we may continue with the current implementation

translate/src/modules/editor/components/Editor.test.jsx
eemeli, line 1: The `renders the source form when passing a broken string` test here (at line 157) is a bit noisy, let's also fix that:

```diff
   it('renders the source form when passing a broken string', () => {
+    const spy = vi.spyOn(console, 'warn').mockImplementation(() => {});
     const [wrapper] = mountEditor(6);
+    expect(spy).toHaveBeenCalledWith(expect.stringContaining('E0028'));
```

nishitmistry: added

PR #3877, opened 2025-12-24
Port hooks test and test with minor change to vitest

nishitmistry:
In continuation of the previous Mr https://github.com/mozilla/pontoon/pull/3875 for task https://github.com/mozilla/pontoon/issues/3768

- ported hooks test to be compatible with vitest 
- ported other test requiring simple changes in there implementation to be compatible with vitest
- added jsdom url for [EntityDetail](https://github.com/mozilla/pontoon/blob/main/translate/src/modules/entitydetails/components/EntityDetails.tsx) and [EntityNavigation](https://github.com/mozilla/pontoon/blob/main/translate/src/modules/entitydetails/components/EntityNavigation.test.jsx)

PR #3880, opened 2025-12-30
Port remaining test in module and context dir to vitest

nishitmistry:
In continuation of the previous Mr https://github.com/mozilla/pontoon/pull/3877 for task https://github.com/mozilla/pontoon/issues/3768

- This Mr includes porting of broken test to vitest under context and modules directory 
- moved useActiveTranslation test to separate file as the mock of react affected all the imports in EntityView.test.jsx, and i found no workaround to keep both test in same file.

- Changed the test-ownership file to just ignore two test files in vitest and to be ran by jest  
'src/utils/message/getEmptyMessage.test.js',
and 
'src/modules/entitieslist/components/EntitiesList.test.js',

- The getEmptyMessage test has been failing since before the transition. I’m currently looking into what needs to be done here. @eemeli  , your guidance would be helpful.

- EntitiesList uses react-intersection-observer which has some ESM issue with vitest, and newer version which might solve this  requires react^17.0.0 || ^18.0.0 || ^19.0.0"

### DIFF
--- a/translate/src/modules/notification/components/NotificationPanel.test.jsx
+++ b/translate/src/modules/notification/components/NotificationPanel.test.jsx
@@ -29,7 +29,7 @@ describe('<NotificationPanel>', () => {
       <ComponentWithProvider initialMessage={null} />,
     );
 
-    expect(container.firstChild.childNodes.length).toBe(1);
+    expect(container.firstChild.childNodes).toHaveLength(1);
     expect(container.querySelector('span').textContent).toBe('');
   });
 
--- a/translate/test-ownership.js
+++ b/translate/test-ownership.js
@@ -7,8 +7,8 @@
  * This file is temporary and will be deleted after the
  * Vitest migration is complete.
  */
-export const ingoreFromVitest = [
+export const ignoreFromVitest = [
   'src/modules/entitieslist/components/EntitiesList.test.js',
   'src/utils/message/getEmptyMessage.test.js',
 ];
-export const includeInJest = ingoreFromVitest.map((x) => `<rootDir>/${x}`);
+export const includeInJest = ignoreFromVitest.map((x) => `<rootDir>/${x}`);
--- a/translate/vitest.config.ts
+++ b/translate/vitest.config.ts
@@ -1,7 +1,7 @@
 import path from 'path';
 import { defineConfig } from 'vitest/config';
 import react from '@vitejs/plugin-react';
-import { ingoreFromVitest } from './test-ownership';
+import { ignoreFromVitest } from './test-ownership';
 
 export default defineConfig({
   plugins: [react()],
@@ -14,7 +14,7 @@ export default defineConfig({
     setupFiles: ['./src/setupTests.ts'],
     mockReset: true,
     testTimeout: 10000,
-    exclude: ingoreFromVitest,
+    exclude: ignoreFromVitest,
     environment: 'jsdom',
     environmentOptions: {
       jsdom: {


translate/test-ownership.js
```diff
@@ -7,20 +7,8 @@
...
-  './src/modules/entities',
-  './src/modules/entitydetails',
-  './src/modules/history',
-  './src/modules/machinery',
-  './src/modules/otherlocales',
-  './src/modules/resourceprogress',
-  './src/modules/teamcomments',
-  './src/modules/unsavedchanges',
-  './src/modules/user',
+export const ingoreFromVitest = [
```
eemeli: Typo also needs fixing where this is referenced.
```suggestion
export const ignoreFromVitest = [
```

translate/src/modules/notification/components/NotificationPanel.test.jsx
```diff
@@ -1,67 +1,71 @@
...
 
   it('returns an empty element when there is no notification', () => {
-    const wrapper = mountNotifications(null);
+    const { container } = render(
+      <ComponentWithProvider initialMessage={null} />,
+    );
 
-    expect(wrapper.children()).toHaveLength(1);
-    expect(wrapper.find('span').text()).toEqual('');
+    expect(container.firstChild.childNodes.length).toBe(1);
```
eemeli:
```suggestion
    expect(container.firstChild.childNodes).toHaveLength(1);
```

eemeli (changes requested):
Looks like we're nearly done with this! See inline for a few tweaks.

> The getEmptyMessage test has been failing since before the transition. I’m currently looking into what needs to be done here.

This test suite works for me locally with both Jest and Vitest as-is. How is it failing for you, and what's your OS & Node.js version?

> EntitiesList uses react-intersection-observer which has some ESM issue with vitest, and newer version which might solve this  requires react^17.0.0 || ^18.0.0 || ^19.0.0"

Let's disable this test suite for now, and add a GitHub issue about it rather than trying to hack together a solution. Because of the enzyme dependency, we're stuck on React 16, which doesn't provide a package.json exports mapping that Vitest expects to resolve the import in the version of react-intersection-observer we're stuck on because, again, we're stuck on React 16.

I think the next PR in this sequence should let us drop Jest, make `npm test` run Vitest, and drop the `test-ownership.js` file, yes?

translate/src/modules/notification/components/NotificationPanel.test.jsx
nishitmistry: fixed

translate/test-ownership.js
nishitmistry: fixed

eemeli (approved):
It looks like the getEmptyMessage test failure is a regression from #3874, as Node.js 22 and earlier don't sort the plural keys like I'd presumed. I'll submit a separate PR fixing my mistake there.

I am quite confused though by why this isn't getting classified as a CI failure, which it ought to do. But that's a separate matter from the changes here, which look good!

> Yes, next Pr will totally drop the jest, test-ownership.js file and will configure vitest as our only test runner. i will create a github issue for EntitiesList with the jest removal PR such that the issue get documented.

Excellent!

> Are we planning to drop sinon, currently the it's not causing any trouble but i think it would be great if we rely just on vitest which provides its mock tools and methods?

Yeah, it'd be good to drop Sinon once we can. Vitest mocking ought to be enough for us.

PR #3891, opened 2026-01-09
Send logs to stdout instead of stderr

### LINKED ISSUE #3888: log.info() messages showing as errors in Cloud Logging
Resources:
https://stackoverflow.com/questions/75361466/google-cloud-logging-assigns-error-severity-to-all-python-logging-info-calls
https://docs.djangoproject.com/en/6.0/topics/logging/#configuring-logging

flodolo:
I'm not sure if this is going to fix #3888 completely, but also not sure how to test it more.

My assumption is that duplicated messages are coming from `log.debug()` ([example](https://github.com/mozilla/pontoon/blob/879bbabcd931598e242c2edde10f6e189f0cc720/pontoon/sync/core/translations_from_repo.py#L139)):
* Those are not displayed in prod.
* On DEV, debug messages are also printed, and anything printed is picked up as an error by GCP Logs.

### DIFF
--- a/pontoon/settings/base.py
+++ b/pontoon/settings/base.py
@@ -4,6 +4,7 @@ import logging
 import os
 import re
 import socket
+import sys
 
 from datetime import datetime
 from ipaddress import ip_address, ip_network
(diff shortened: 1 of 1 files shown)


mathjazz (approved):
Changes look good and make sense in Django deployment via Docker / Kubernetes.

(Duplicated log messages are not specific to the new deployment and they only seem to happen on background tasks. Perhaps Celery has something to do with that.)

PR #3903, opened 2026-01-15
Add management command to be used for AutoML warmup

### LINKED ISSUE #3902: Add management command to be used for AutoML warmup
The only reason we had a separate AutoML dyno on Heroku was that Heroku Scheduler (Heroku's cron equivalent) did not support running jobs at a minute interval. 

That design decision propagated to GCP, but we should just add a management command and run it in cron.

mathjazz:
Fix #3902.

### DIFF
--- a/pontoon/machinery/management/commands/warmup_automl.py
+++ b/pontoon/machinery/management/commands/warmup_automl.py
@@ -5,7 +5,7 @@ from pontoon.machinery.utils import get_google_automl_translation
 
 
 class Command(BaseCommand):
-    help = "Collect data needed for the Insights tab"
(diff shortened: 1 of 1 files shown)


pontoon/machinery/management/commands/warmup_automl.py
```diff
@@ -0,0 +1,24 @@
+from django.core.management.base import BaseCommand
+
+from pontoon.base.models import Locale
+from pontoon.machinery.utils import get_google_automl_translation
+
+
+class Command(BaseCommand):
+    help = "Collect data needed for the Insights tab"
```
flodolo: This needs to be updated.

Issue #3767, opened 2025-09-04
Move remaining enzyme tests to @testing-library/react


eemeli:
As (belated) follow-up to #2760, we should complete the test migration and get rid of `enzyme`, which has not been updated sinze 2019, and which is blocking our overall update from React 16 onwards.

Some examples of tests using `@testing-library/react` are available in [EditField.test.tsx](https://github.com/mozilla/pontoon/blob/main/translate/src/modules/translationform/components/EditField.test.tsx) and [Terms.test.js](https://github.com/mozilla/pontoon/blob/main/translate/src/modules/terms/components/Terms.test.js).

Raj-G07:
@eemeli I am also working on this issue with #3768

PR #3909, opened 2026-01-17
Shift mocked store enzyme test to React Testing Library

nishitmistry:
- Starting the migration of enzyme test to @testing-library/react under task https://github.com/mozilla/pontoon/issues/3767
- Added @testing-library/jest-dom package for some handy assertion methods 
- Changed the mountComponentWithStore to use @testing-library/react package's render method.
- Fixed the tests which broke with the change made to mountComponentWithStore method.

### DIFF
--- a/translate/src/modules/comments/components/CommentsList.test.js
+++ b/translate/src/modules/comments/components/CommentsList.test.js
@@ -20,18 +20,21 @@ describe('<CommentsList>', () => {
 
   it('shows the correct number of comments', () => {
     const store = createReduxStore();
-    const wrapper = mountComponentWithStore(CommentsList, store, {
+    const comments = [
+      { id: 1, content: '11', userBanner: '' },
+      { id: 2, content: '22', userBanner: '' },
+      { id: 3, content: '33', userBanner: '' },
+    ];
+    const { getByText } = mountComponentWithStore(CommentsList, store, {
       translation: {
         ...DEFAULT_TRANSLATION,
-        comments: [
-          { id: 1, content: '11', userBanner: '' },
-          { id: 2, content: '22', userBanner: '' },
-          { id: 3, content: '33', userBanner: '' },
-        ],
+        comments,
       },
       user: DEFAULT_USER,
     });
 
-    expect(wrapper.find('ul > *')).toHaveLength(3);
+    for (const comment of comments) {
+      getByText(comment.content);
+    }
   });
 });
--- a/translate/src/modules/editor/components/FailedChecks.test.jsx
+++ b/translate/src/modules/editor/components/FailedChecks.test.jsx
@@ -10,7 +10,7 @@ import {
 } from '~/test/store';
 
 import { FailedChecks } from './FailedChecks';
-import { vi } from 'vitest';
+import { vi, expect } from 'vitest';
 
 function mountFailedChecks(failedChecks, user) {
   const store = createReduxStore({ project: { slug: 'firefox', tags: [] } });
@@ -30,56 +30,66 @@ describe('<FailedChecks>', () => {
   it('does not render if no errors or warnings present', () => {
     const wrapper = mountFailedChecks({ errors: [], warnings: [] });
 
-    expect(wrapper.find('.failed-checks')).toHaveLength(0);
+    expect(
+      wrapper.container.querySelector('.failed-checks'),
+    ).not.toBeInTheDocument();
   });
 
   it('renders popup with errors and warnings', () => {
-    const wrapper = mountFailedChecks({
-      errors: ['one error'],
-      warnings: ['a warning', 'two warnings'],
+    const errors = ['one error'];
+    const warnings = ['a warning', 'two warnings'];
+    const { container, getByRole, getByText } = mountFailedChecks({
+      errors,
+      warnings,
     });
 
-    expect(wrapper.find('.failed-checks')).toHaveLength(1);
-    expect(wrapper.find('#editor-FailedChecks--close')).toHaveLength(1);
-    expect(wrapper.find('#editor-FailedChecks--title')).toHaveLength(1);
-    expect(wrapper.find('.error')).toHaveLength(1);
-    expect(wrapper.find('.warning')).toHaveLength(2);
+    expect(container.querySelector('.failed-checks')).toBeInTheDocument();
+    expect(getByRole('button', { name: /Close/i })).toHaveTextContent('×');
+    getByText('THE FOLLOWING CHECKS HAVE FAILED');
+
+    for (const error of errors) {
+      getByText(error);
+    }
+
+    for (const warning of warnings) {
+      getByText(warning);
+    }
   });
 
   it('renders save anyway button if translation with warnings submitted', () => {
-    const wrapper = mountFailedChecks(
+    const { getByRole } = mountFailedChecks(
       { errors: [], warnings: ['a warning'], source: 'submitted' },
       { settings: { force_suggestions: false } },
     );
 
-    expect(wrapper.find('.save.anyway')).toHaveLength(1);
+    getByRole('button', { name: /save anyway/i });
   });
 
   it('renders suggest anyway button if translation with warnings suggested', () => {
-    const wrapper = mountFailedChecks(
+    const { getByRole } = mountFailedChecks(
       { errors: [], warnings: ['a warning'], source: 'submitted' },
       { settings: { force_suggestions: true } },
     );
 
-    expect(wrapper.find('.suggest.anyway')).toHaveLength(1);
+    getByRole('button', { name: /suggest anyway/i });
   });
 
   it('renders suggest anyway button if user does not have sufficient permissions', () => {
-    const wrapper = mountFailedChecks(
+    const { getByRole } = mountFailedChecks(
       { errors: [], warnings: ['a warning'], source: 'submitted' },
       { can_manage_locales: [] },
     );
 
-    expect(wrapper.find('.suggest.anyway')).toHaveLength(1);
+    getByRole('button', { name: /suggest anyway/i });
   });
 
   it('renders approve anyway button if translation with warnings approved', () => {
-    const wrapper = mountFailedChecks({
+    const { getByRole } = mountFailedChecks({
       errors: [],
       warnings: ['a warning'],
       source: 42,
     });
 
-    expect(wrapper.find('.approve.anyway')).toHaveLength(1);
+    getByRole('button', { name: /approve anyway/i });
   });
 });
--- a/translate/src/modules/editor/components/FailedChecks.test.jsx
+++ b/translate/src/modules/editor/components/FailedChecks.test.jsx
@@ -28,11 +28,9 @@ function mountFailedChecks(failedChecks, user) {
 
 describe('<FailedChecks>', () => {
   it('does not render if no errors or warnings present', () => {
-    const wrapper = mountFailedChecks({ errors: [], warnings: [] });
+    const { container } = mountFailedChecks({ errors: [], warnings: [] });
 
-    expect(
-      wrapper.container.querySelector('.failed-checks'),
-    ).not.toBeInTheDocument();
+    expect(container.querySelector('.failed-checks')).not.toBeInTheDocument();
   });
 
   it('renders popup with errors and warnings', () => {
--- a/translate/src/modules/entitieslist/components/EntitiesList.test.js
+++ b/translate/src/modules/entitieslist/components/EntitiesList.test.js
@@ -13,6 +13,7 @@ import {
 
 import { EntitiesList } from './EntitiesList';
 import { expect, vi } from 'vitest';
+import { fireEvent } from '@testing-library/react';
 
 // Entities shared between tests
 const ENTITIES = [
@@ -58,8 +59,8 @@ describe('<EntitiesList>', () => {
       entities: ENTITIES,
       hasMore: true,
     });
-    const wrapper = mountComponentWithStore(EntitiesList, store);
-    expect(wrapper.find('SkeletonLoader')).toHaveLength(1);
+    const { getByTestId } = mountComponentWithStore(EntitiesList, store);
+    getByTestId('skeleton-loader');
   });
 
   it("doesn't display a loading animation when there aren't entities to load", () => {
@@ -69,18 +70,18 @@ describe('<EntitiesList>', () => {
       entities: ENTITIES,
       hasMore: false,
     });
-    const wrapper = mountComponentWithStore(EntitiesList, store);
+    const { queryByTestId } = mountComponentWithStore(EntitiesList, store);
 
-    expect(wrapper.find('SkeletonLoader')).toHaveLength(0);
+    expect(queryByTestId('skeleton-loader')).not.toBeInTheDocument();
   });
 
   // FIXME: https://github.com/mozilla/pontoon/issues/3883
   it.skip('shows a loading animation when entities are being fetched from the server', () => {
     const store = createReduxStore();
     store.dispatch({ type: EntitiesActions.REQUEST_ENTITIES });
-    const wrapper = mountComponentWithStore(EntitiesList, store);
+    const { getByTestId } = mountComponentWithStore(EntitiesList, store);
 
-    expect(wrapper.find('SkeletonLoader')).toHaveLength(1);
+    getByTestId('skeleton-loader');
   });
 
   it('shows the correct number of entities', () => {
@@ -94,9 +95,14 @@ describe('<EntitiesList>', () => {
       entities: ENTITIES,
       hasMore: false,
     });
-    const wrapper = mountComponentWithStore(EntitiesList, store, {}, history);
+    const { getAllByRole } = mountComponentWithStore(
+      EntitiesList,
+      store,
+      {},
+      history,
+    );
 
-    expect(wrapper.find('Entity')).toHaveLength(2);
+    expect(getAllByRole('listitem')).toHaveLength(2);
   });
 
   // FIXME: https://github.com/mozilla/pontoon/issues/3883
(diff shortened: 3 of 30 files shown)


nishitmistry:
Hey @mathjazz, I am done with the migrations of store related tests, you can go ahead with the review.

translate/src/modules/editor/components/FailedChecks.test.jsx
```diff
@@ -11,6 +11,7 @@ import {
 
 import { FailedChecks } from './FailedChecks';
 import { vi } from 'vitest';
+import { expect } from 'vitest';
```
mathjazz: Why not put everything in one line?

```js
import { vi, expect } from 'vitest';
```

translate/src/modules/comments/components/CommentsList.test.js
```diff
@@ -32,6 +32,6 @@ describe('<CommentsList>', () => {
       user: DEFAULT_USER,
     });
 
-    expect(wrapper.find('ul > *')).toHaveLength(3);
+    expect(container.querySelectorAll('ul > *')).toHaveLength(3);
```
mathjazz: Could you please help me understand what's the motivation for this change (and many other similar changes below)? Is `find()` coming from `enzyme`?

translate/src/modules/entitieslist/components/Entity.tsx
```diff
@@ -113,7 +113,7 @@ export function Entity({
   );
 
   return (
-    <li className={cn} onClick={handleSelectEntity}>
+    <li className={cn} onClick={handleSelectEntity} data-testid='entity'>
```
mathjazz: Not a big fan of adding `data-*` attributes just for the sake of testing. Is it really necessary?

mathjazz (commented):
Nice work! 

Added some high-level questions and comments. Please also rebase against `main`.

Thanks for your contributions to Pontoon!

translate/src/modules/comments/components/CommentsList.test.js
nishitmistry: yes, `find()` is returned  by the `mount()` method of enzyme, which locates a node based on any selector. There's no such method in React Testing Library to select a node with html selector, so we use the container's `querySelector()` to select the nodes.

translate/src/modules/editor/components/FailedChecks.test.jsx
nishitmistry: fixed

translate/src/modules/entitieslist/components/Entity.tsx
nishitmistry: We can't query based on the component as we do with Enzyme ( `wrapper.find(Entity);` is not possible in React Testing Library). So if we wish to know whether a particular component exists or not,  adding data-testid='entity' and then querying it would be the way afaik. 
React Testing Library recommends an accessibility-based testing approach [RTL queries priority guide](https://testing-library.com/docs/queries/about/#priority), but the task is limited to just the migration.

translate/src/modules/editor/components/FailedChecks.test.jsx
```diff
@@ -30,7 +30,9 @@ describe('<FailedChecks>', () => {
   it('does not render if no errors or warnings present', () => {
     const wrapper = mountFailedChecks({ errors: [], warnings: [] });
```
eemeli: Let's go with the destructuring style you use elsewhere in this PR.
```suggestion
    const { container } = mountFailedChecks({ errors: [], warnings: [] });
```

This same suggestion also applies to other places where the `render()` result is used.

translate/src/modules/comments/components/CommentsList.test.js
eemeli: Using `querySelector()` is a bit of a code smell with testing-library, tbh. It's ok to do so initially here, but it would be good to leave `// TODO` code comments in the vicinity of each such test to replace them with `getBy*` and similar testing-library-ish approaches.

translate/src/modules/history/components/History.tsx
```diff
@@ -26,7 +26,10 @@ export function History(): React.ReactElement<'section'> | null {
 
   if (!translations.length) {
     return fetching ? null : (
-      <section className='history'>
+      <section
+        className='history'
+        data-testid='history-History--no-translations'
```
eemeli: You've taken this testid from the localization id. Rather than doing so, this (and potentially other similar cases) seem like a perfect use case for `getByText()`.

translate/src/modules/entitieslist/components/EntitiesList.test.js
```diff
@@ -52,7 +53,7 @@ describe('<EntitiesList>', () => {
       hasMore: true,
     });
     const wrapper = mountComponentWithStore(EntitiesList, store);
```
eemeli: Breaking this suggestion in two to deal with GitHub's limitations; similar also applies to other tests -- the `getBy*` functions should be preferred where they're an option, and they don't need an `expect()` wrapper for an N=1 expected result.

```suggestion
    const { getByTestId } = mountComponentWithStore(EntitiesList, store);
```

translate/src/modules/entitieslist/components/EntitiesList.test.js
```diff
@@ -52,7 +53,7 @@ describe('<EntitiesList>', () => {
...
-    expect(wrapper.find('SkeletonLoader')).toHaveLength(1);
+    expect(wrapper.queryByTestId('skeleton-loader')).toBeInTheDocument();
```
eemeli:
```suggestion
    getByTestId('skeleton-loader');
```

translate/src/modules/entitieslist/components/Entity.tsx
eemeli: The testid attributes do have their place, but as the page you link to itself documents, they should only be used when no other option works.

translate/src/modules/navbar/components/Navigation.test.jsx
```diff
@@ -35,7 +35,7 @@ describe('<Navigation>', () => {
       HISTORY,
     );
 
-    expect(wrapper.text()).toContain('Klingon');
-    expect(wrapper.text()).toContain('kg');
+    expect(wrapper.container).toHaveTextContent('Klingon');
+    expect(wrapper.container).toHaveTextContent('kg');
```
eemeli: All of the tests (not just in this file, but also elsewhere) that use `.toHaveTextContent()` and other replacements for Enzyme's `.text()` might be best replaced with `getByText()`, which also accepts a regexp argument.

translate/src/modules/otherlocales/components/OtherLocaleTranslation.test.jsx
```diff
@@ -61,32 +61,28 @@ describe('<OtherLocaleTranslationComponent>', () => {
   });
 
   it('renders a plain translation correctly', () => {
-    const wrapper = createTranslation('', PLAIN_TRANSLATION);
+    createTranslation('', PLAIN_TRANSLATION);
```
eemeli:
```suggestion
    const { getByText } = createTranslation('', PLAIN_TRANSLATION);
```

translate/src/modules/otherlocales/components/OtherLocaleTranslation.test.jsx
```diff
@@ -61,32 +61,28 @@ describe('<OtherLocaleTranslationComponent>', () => {
...
 
-    const gt = wrapper.find('GenericTranslation');
-    expect(gt.props().content).toMatch(/^Un cheval/);
+    expect(screen.getByText(/^Un cheval/)).toBeInTheDocument();
```
eemeli:
```suggestion
    getByText(/^Un cheval/);
```

translate/src/modules/translationform/components/TranslationForm-multiple.test.jsx
```diff
@@ -55,10 +56,10 @@ function mountForm(string) {
     ),
     store,
   );
-
-  const views = Array.from(
-    wrapper.find('.translationform').instance().querySelectorAll('.cm-content'),
-  ).map((el) => el.cmView.view);
+  const form = wrapper.container.querySelector('.translationform');
```
eemeli, line 60: Wouldn't `getByRole('form')` work here?

translate/src/modules/translationform/components/TranslationForm-multiple.test.jsx
```diff
@@ -195,53 +212,74 @@ describe('<TranslationForm> with multiple fields', () => {
...
+    ).toHaveTextContent('i');
+    expect(
+      container.querySelectorAll('.accesskeys button')[5],
+    ).toHaveTextContent('t');
+    expect(
+      container.querySelectorAll('.accesskeys button')[6],
+    ).toHaveTextContent('e');
+    expect(
+      container.querySelectorAll('.accesskeys button')[7],
+    ).toHaveTextContent('s');
```
eemeli: Something like this would be more legible:
```suggestion
    const buttonTexts = Array.from(
      container.querySelectorAll('.accesskeys button'),
      button => button.textContent
    );
    expect(buttonTexts).toMatchObject(['C', 'a', 'n', 'd', 'i', 't', 'e', 's']);
```

eemeli (changes requested):
Looks pretty decent. A few too many test ids and on occasion a bit too verbose, but decent. More comments inline.

mathjazz:
Hey @nishitmistry - just checking if you still want to finish this PR. Thanks!

translate/src/modules/history/components/History.tsx
nishitmistry: fixed

translate/src/modules/entitieslist/components/EntitiesList.test.js
nishitmistry: fixed

translate/src/modules/editor/components/FailedChecks.test.jsx
nishitmistry: fixed

translate/src/modules/comments/components/CommentsList.test.js
nishitmistry: Hey @eemeli, i have replaced most of the queryselector where testing-library-ish approach was possible. some queryselector still exist given there were no clear approach asserting there presence.

translate/src/modules/entitieslist/components/Entity.tsx
nishitmistry: i have removed most of the testid assertion and have updated them with the equivalent testing library approach

translate/src/modules/navbar/components/Navigation.test.jsx
nishitmistry: i have updated all toHaveTextContent assertion with getByText approach, except where the order of text might matter or text is broken into multiple element.

translate/src/modules/otherlocales/components/OtherLocaleTranslation.test.jsx
nishitmistry: fixed

translate/src/modules/otherlocales/components/OtherLocaleTranslation.test.jsx
nishitmistry: fixed

translate/src/modules/translationform/components/TranslationForm-multiple.test.jsx
nishitmistry: no, because in the TranslationForm.tsx we are not using the form tag
i have added a todo to maybe update it later

translate/src/modules/translationform/components/TranslationForm-multiple.test.jsx
nishitmistry: fixed

flodolo:
Note that you need to rebase and resolve conflicts.

translate/src/modules/editor/components/FailedChecks.test.jsx
eemeli: This does not appear to be fixed?

translate/src/modules/interactivetour/components/InteractiveTour.test.jsx
```diff
@@ -19,28 +21,28 @@ describe('<InteractiveTour>', () => {
       project: { slug: 'tutorial' },
       user: { isAuthenticated: false },
     });
-    const wrapper = mountComponentWithStore(InteractiveTour, store);
+    const { queryByTestId } = mountComponentWithStore(InteractiveTour, store);
 
-    expect(wrapper.find(Tour)).toHaveLength(1);
+    expect(queryByTestId('mock-tour')).toBeInTheDocument();
```
eemeli:
```suggestion
    const { getByTestId } = mountComponentWithStore(InteractiveTour, store);

    getByTestId('mock-tour');
```

translate/src/modules/projectinfo/components/ProjectInfo.tsx
```diff
@@ -39,7 +39,7 @@ export function ProjectInfo(): React.ReactElement<'div'> | null {
 
   return fetching || !info ? null : (
     <div className='project-info'>
-      <div className='button' onClick={toggleVisible}>
+      <div className='button' role='button' onClick={toggleVisible}>
```
eemeli, line 42: This kind of fix is part of the reason why we're migrating to testing-library. :)

translate/src/modules/resourceprogress/components/ResourceProgress.test.js
```diff
@@ -28,21 +29,25 @@ describe('<ResourceProgress>', () => {
...
+      ResourceProgress,
+      store,
+      {
+        parameters: PARAMETERS,
+      },
+    );
 
-    expect(wrapper.find('.selector').exists()).toBeTruthy();
-    expect(wrapper.find('ResourceProgressDialog').exists()).toBeFalsy();
+    expect(getByRole('button')).toBeInTheDocument();
```
eemeli:
```suggestion
    getByRole('button');
```

translate/src/modules/projectinfo/components/ProjectInfo.test.js
```diff
@@ -1,46 +1,49 @@
...
     const store = createReduxStore({ project: PROJECT });
-    const wrapper = mountComponentWithStore(ProjectInfo, store);
+    const { getByRole, queryByRole } = mountComponentWithStore(
+      ProjectInfo,
+      store,
+    );
 
-    expect(wrapper.find('.button').exists()).toBeTruthy();
-    expect(wrapper.find('aside').exists()).toBeFalsy();
+    expect(getByRole('button')).toBeInTheDocument();
```
eemeli:
```suggestion
    getByRole('button');
```

translate/src/modules/projectinfo/components/ProjectInfo.test.js
```diff
@@ -1,46 +1,49 @@
...
 
   it('shows the info panel after a click', () => {
     const store = createReduxStore({ project: PROJECT });
-    const wrapper = mountComponentWithStore(ProjectInfo, store);
-    wrapper.find('.button').simulate('click');
+    const { getByRole } = mountComponentWithStore(ProjectInfo, store);
+    fireEvent.click(getByRole('button'));
 
-    expect(wrapper.find('aside.panel').exists()).toBeTruthy();
+    expect(getByRole('complementary')).toBeInTheDocument();
```
eemeli:
```suggestion
    getByRole('complementary');
```

translate/src/modules/editor/components/FailedChecks.test.jsx
nishitmistry: oh right, i missed this one, i have pushed the change.

eemeli (changes requested):
This looks nearly ready, though see inline for a few remaining suggestions (which you should be able to merge directly from the website).

Overall, I'm not completely convinced that
```js
expect(query...()).not.toBeInTheDocument();
```
is more readable than
```js
expect(query...()).toBeNull();
```
but I'm fine with leaving these as-is.

> Hey @eemeli, the AddComment.test.jsx test is failing, the issue seems to be jsdom not having support for contenteditable component like slate, so the editor's value is not updated if type event is fire from test and clicking the submit get blocked because of the trimmed value check. Should i skip the test for now and create a issue for it ?

Let's just drop the `'calls submitComment function'` test altogether. The only reason it's not failing at the moment is that the final `.toBeTruthy` check is missing `()` to actually call it. To properly test this we'd need to expose the internal `editor` value just for testing, and that seems a bit pointless for what this test is actually testing.

translate/src/modules/interactivetour/components/InteractiveTour.test.jsx
nishitmistry: fixed

translate/src/modules/resourceprogress/components/ResourceProgress.test.js
nishitmistry: fixed

translate/src/modules/projectinfo/components/ProjectInfo.test.js
nishitmistry: fixed

translate/src/modules/projectinfo/components/ProjectInfo.test.js
nishitmistry: fixed

translate/src/modules/projectinfo/components/ProjectInfo.tsx
nishitmistry: yeah,  it makes the website more accessible and also this type of testing approach seems more promising. 
if we are targeting for WCAG compliance , this would be our first step.

eemeli (approved):

PR #3917, opened 2026-01-23
Fix error with timezone when accessing /insights

flodolo:


pontoon/insights/utils.py
```diff
@@ -1,6 +1,6 @@
 import json
 
-from datetime import timedelta
+from datetime import datetime, timedelta
```
mathjazz:
```suggestion
from datetime import timedelta
```

pontoon/insights/utils.py
```diff
@@ -22,8 +22,10 @@ def get_insight_start_date():
     """Include at most the last year of data in insights."""
     now = timezone.now()
     if now.month == 12:
-        return timezone.datetime(now.year, 1, 1)
-    return timezone.datetime(now.year - 1, now.month + 1, 1)
+        naive_dt = datetime(now.year, 1, 1)
+    else:
+        naive_dt = datetime(now.year - 1, now.month + 1, 1)
```
mathjazz:
```suggestion
        naive_dt = timezone.datetime(now.year, 1, 1)
    else:
        naive_dt = timezone.datetime(now.year - 1, now.month + 1, 1)
```

pontoon/insights/utils.py
```diff
@@ -22,8 +22,10 @@ def get_insight_start_date():
...
+    return timezone.make_aware(naive_dt, timezone.get_default_timezone())
```
mathjazz: I believe you can leave that 2nd argument out?

```suggestion
    return timezone.make_aware(naive_dt)
```

PR #3918, opened 2026-01-23
Reorder Django middlewares

flodolo:
From https://docs.djangoproject.com/en/6.0/ref/middleware/#middleware-ordering

```
SessionMiddleware

Before any middleware that may raise an exception to trigger an error view (such as PermissionDenied) if you’re using CSRF_USE_SESSIONS.
```

Is the current order incorrect? Not sure if that's what is causing errors in the log on `django/middleware/csrf.py`

### DIFF
--- a/pontoon/settings/base.py
+++ b/pontoon/settings/base.py
@@ -333,18 +333,18 @@ MIDDLEWARE = (
     "corsheaders.middleware.CorsMiddleware",
     "whitenoise.middleware.WhiteNoiseMiddleware",
     "django.middleware.gzip.GZipMiddleware",
+    "pontoon.base.middleware.RaygunExceptionMiddleware",
     "pontoon.base.middleware.BlockedIpMiddleware",
     "django.contrib.sessions.middleware.SessionMiddleware",
     "django.middleware.common.CommonMiddleware",
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "allauth.account.middleware.AccountMiddleware",
+    "pontoon.base.middleware.ThrottleIpMiddleware",
     "django.middleware.csrf.CsrfViewMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
     "csp.middleware.CSPMiddleware",
-    "pontoon.base.middleware.ThrottleIpMiddleware",
     "pontoon.base.middleware.EmailConsentMiddleware",
-    "pontoon.base.middleware.RaygunExceptionMiddleware",
 )
 
 CONTEXT_PROCESSORS = (


pontoon/settings/base.py
```diff
@@ -330,21 +330,21 @@ def _default_from_email():
 
 MIDDLEWARE = (
     "django.middleware.security.SecurityMiddleware",
+    "corsheaders.middleware.CorsMiddleware",
```
flodolo, line 333: https://pypi.org/project/django-cors-headers/

> CorsMiddleware should be placed as high as possible, especially before any middleware that can generate responses such as Django’s CommonMiddleware or Whitenoise’s WhiteNoiseMiddleware. If it is not before, it will not be able to add the CORS headers to these responses.

pontoon/settings/base.py
```diff
@@ -330,21 +330,21 @@ def _default_from_email():
...
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "allauth.account.middleware.AccountMiddleware",
-    "pontoon.base.middleware.ThrottleIpMiddleware",
     "django.middleware.csrf.CsrfViewMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
     "csp.middleware.CSPMiddleware",
+    "pontoon.base.middleware.ThrottleIpMiddleware",
     "pontoon.base.middleware.EmailConsentMiddleware",
+    "pontoon.base.middleware.RaygunExceptionMiddleware",
```
flodolo: Having this at the end should help catching all errors?

mathjazz: Since migrating to GCP, we no longer use Raygun. That said, there might be instances of Pontoon that are hosted on Heroku and rely on the Raygun Add-On.

Still, I'd say we should remove Raygun from the codebase, because it'll be difficult for us to main it without testing it in real life. We should also consider dropping other Heroku-specific bits from the codebase.

flodolo: I filed #3919 for that.

pontoon/settings/base.py
```diff
@@ -330,21 +330,21 @@ def _default_from_email():
...
     "django.contrib.sessions.middleware.SessionMiddleware",
+    "django.middleware.common.CommonMiddleware",
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "allauth.account.middleware.AccountMiddleware",
-    "pontoon.base.middleware.ThrottleIpMiddleware",
     "django.middleware.csrf.CsrfViewMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
     "csp.middleware.CSPMiddleware",
+    "pontoon.base.middleware.ThrottleIpMiddleware",
```
mathjazz: What's the motivation for moving `ThrottleIpMiddleware` further down?

pontoon/settings/base.py
mathjazz: Should we then restore the position of `RaygunExceptionMiddleware` until #3919 is addressed and communicated?

flodolo: Yes, it makes sense.

pontoon/settings/base.py
flodolo: I moved it closer to the Pontoon custom middlewares, but there is no real reason (the only strict one is putting in behind AccountMiddleware, since we need the user). I'll move it back to minimize the diff 👍🏼

PR #3939, opened 2026-01-31
Community Health: change completion and missing strings calculation

functionzz:
## Summary
This PR addresses Part 2 of the Community Health OKR where we change completion metrics to exclude pretranslations, which will provide a more accurate picture of if a project/locale is finished being localized.

The new completion formula is as follows:
`total == approved + warnings` instead of `total == approved + pretranslated + warnings`.

Changes to the `missing_strings` formula are also noted.
`missing_strings == total - approved - pretranslated` instead of
`missing_strings == total - approved - pretranslated - warnings - errors`.

### DIFF
--- a/pontoon/api/tests/test_views.py
+++ b/pontoon/api/tests/test_views.py
@@ -114,7 +114,7 @@ def test_locale(django_assert_num_queries):
     translated_resource.pretranslated_strings = 0
     translated_resource.strings_with_errors = 3
     translated_resource.strings_with_warnings = 2
-    translated_resource.missing_strings = 5
+    translated_resource.missing_strings = 10
     translated_resource.unreviewed_strings = 5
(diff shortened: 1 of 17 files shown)


mathjazz:
The current formula is: `total = translated + pretranslated + warnings + errors + missing`. Each of the 5 translation statuses that add up to `total` are represented with their own color and plotted on the "pie" chart. I believe this should stay the same.

KR2 says "Stop counting pretranslations toward completion", but completion is a different thing. It's currently calculated as `completion = (translated + pretranslated + warnings) / total` and we want to change that to `completion = (translated + warnings) / total`. That's what this PR should address.

Also, given that it's practically impossible to submit a string with errors... should we drop the Errors status at some point?

functionzz:
Closing due to this being quite messy :/

PR #3946, opened 2026-02-05
Translate view: don't try to log UX action for unauthenticated users

### LINKED ISSUE #3900: `Forbidden: /log-ux-action/` errors in logs
There are several errors in the logs, but UX actions are logged in the meantime, so it seems to only affects a subset of POST.

I've tried to look around the error in logs, but can't figure out what's the cause.

flodolo:
Fixes #3900 

Not sure if there's a better way to do it.

### DIFF
--- a/translate/src/modules/entitieslist/components/EntitiesList.test.js
+++ b/translate/src/modules/entitieslist/components/EntitiesList.test.js
@@ -164,7 +164,13 @@ describe('<EntitiesList>', () => {
     expect(BatchActions.toggleSelection).toHaveBeenCalledTimes(1);
   });
 
-  it('renders correctly for unauthenticated user', () => {
+  it('does not log UX action when unauthenticated user loads page with search parameter', () => {
+    const history = createMemoryHistory({
+      initialEntries: [
+        '/kg/firefox/all-resources/?string=1&search=test&search_identifiers=true',
+      ],
+    });
+
     const store = createReduxStore();
     store.dispatch({
       type: EntitiesActions.RECEIVE_ENTITIES,
@@ -174,13 +180,19 @@ describe('<EntitiesList>', () => {
 
     createDefaultUser(store, { is_authenticated: false });
 
-    const wrapper = mountComponentWithStore(EntitiesList, store);
+    mountComponentWithStore(EntitiesList, store, {}, history);
 
-    expect(wrapper.find('.entities')).toHaveLength(1);
+    // Verify that logUXAction was NOT called for unauthenticated user
     expect(mockLogUXAction).not.toHaveBeenCalled();
   });
 
-  it('renders correctly for authenticated user', () => {
+  it('logs UX action when authenticated user loads page with search parameter', () => {
+    const history = createMemoryHistory({
+      initialEntries: [
+        '/kg/firefox/all-resources/?string=1&search=test&search_identifiers=true',
+      ],
+    });
+
     const store = createReduxStore();
     store.dispatch({
       type: EntitiesActions.RECEIVE_ENTITIES,
@@ -190,9 +202,19 @@ describe('<EntitiesList>', () => {
 
     createDefaultUser(store, { is_authenticated: true });
 
-    const wrapper = mountComponentWithStore(EntitiesList, store);
+    mountComponentWithStore(EntitiesList, store, {}, history);
 
-    expect(wrapper.find('.entities')).toHaveLength(1);
-    expect(mockLogUXAction).not.toHaveBeenCalled();
+    // Verify that logUXAction was called when component mounted with search parameters
+    expect(mockLogUXAction).toHaveBeenCalledWith(
+      'Load: String list with search parameter',
+      'Search Options Statistics',
+      {
+        search_exclude_source_strings: false,
+        search_identifiers: true,
+        search_match_case: false,
+        search_match_whole_word: false,
+        search_rejected_translations: false,
+      },
+    );
   });
 });
--- a/translate/src/modules/entitieslist/components/EntitiesList.tsx
+++ b/translate/src/modules/entitieslist/components/EntitiesList.tsx
@@ -154,6 +154,31 @@ export function EntitiesList(): React.ReactElement<'div'> {
     }
   });
 
+  // Log UX action when authenticated user loads page with search parameters
+  useEffect(() => {
+    if (isAuthUser && location.search) {
+      logUXAction(
+        'Load: String list with search parameter',
+        'Search Options Statistics',
+        {
+          search_exclude_source_strings: location.search_exclude_source_strings,
+          search_identifiers: location.search_identifiers,
+          search_match_case: location.search_match_case,
+          search_match_whole_word: location.search_match_whole_word,
+          search_rejected_translations: location.search_rejected_translations,
+        },
+      );
+    }
+  }, [
+    isAuthUser,
+    location.search,
+    location.search_exclude_source_strings,
+    location.search_identifiers,
+    location.search_match_case,
+    location.search_match_whole_word,
+    location.search_rejected_translations,
+  ]);
+
   // Whenever the route changes, we want to verify that the user didn't
   // change locale, project, resource... If they did, then we'll have
   // to reset the current list of entities, in order to start a fresh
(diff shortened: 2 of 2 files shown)


translate/public/translate.html
```diff
@@ -30,6 +30,7 @@
     <div
         id="root"
         data-csrf-token="{{ csrf_token }}"
+        data-is-authenticated="{{ 'true' if request.user.is_authenticated else 'false' }}"
```
mathjazz:
```suggestion
        data-is-authenticated="{{ request.user.is_authenticated | to_json() }}"
```

mathjazz: @eemeli Is there an elegant way to reuse authentication data that's already sent with the `/user-data/` request rather than sending it one more time here?

flodolo: I've committed this change in one commit, then gave it a try with a JS only approach. It works and tests still pass, but I don't know if that's the best way to write JS code.

flodolo:
I tried to use the data already available in the frontend instead of adding info to the Django template.

Just in case, I kept the changes in a separate commit, so it's easy to revert https://github.com/mozilla/pontoon/pull/3946/changes/9f6564bb5786f5bd12bc7b8192f86f5e09c04b6a

Worth noting that I'm very much hand waving when writing JavaScript, and even more so when writing tests (asked ChatGPT to write one based on test cases).

translate/src/api/uxaction.ts
```diff
@@ -1,11 +1,27 @@
+import { fetchUserData } from './user';
 import { POST } from './utils/base';
 import { getCSRFToken } from './utils/csrfToken';
 
+let userDataPromise: ReturnType<typeof fetchUserData> | null = null;
+
+async function isAuthenticated(): Promise<boolean> {
+  if (!userDataPromise) {
+    userDataPromise = fetchUserData();
```
eemeli: Re-fetching user data for logging like this is not the right solution; the info is available at least via `useAppSelector((state) => state[USER].isAuthenticated)`. Accesssing that might require renaming some functions to be hooks, or to pass the data along in arguments.

flodolo:
I think I'm in React/Redux hell. I tried moving the authentication state as a parameter of the existing function, but that required adding it to all props, update calls, tests, etc.

This is how far I manage to go. It works, I'm not sure how correct it is in terms of model (e.g. separation).

translate/src/modules/machinery/components/MachineryTranslation.tsx
```diff
@@ -53,13 +54,21 @@ export function MachineryTranslationComponent({
...
       setEditorFromHelpers(content, sources, true);
       if (llmTranslation) {
-        logUXAction('LLM Translation Copied', 'LLM Feature Adoption', {
+        log('LLM Translation Copied', 'LLM Feature Adoption', {
           action: 'Copy LLM Translation',
           localeCode: locale.code,
         });
       }
     }
-  }, [index, setEditorFromHelpers, translation, llmTranslation]);
```
flodolo, line 62: Was this wrong before?

translate/src/modules/entities/actions.ts
```diff
@@ -62,7 +63,8 @@ export const updateEntityTranslation = (
 
 /** Fetch entities and their translation.  */
 export const getEntities =
-  (location: Location, page: number) => async (dispatch: AppDispatch) => {
+  (location: Location, page: number) =>
+  async (dispatch: AppDispatch, getState: () => RootState) => {
```
flodolo: Is `RootState` the correct type?

translate/src/modules/user/components/UserNotificationsMenu.tsx
```diff
@@ -71,30 +71,32 @@ export function UserNotificationsMenu({
...
 
   useEffect(() => {
     if (user.isAuthenticated && unread) {
-      logUXAction('Render: Unread notifications icon', 'Notifications 1.0', {
+      log('Render: Unread notifications icon', 'Notifications 1.0', {
         pathname: window.location.pathname,
       });
     }
-  }, []);
+  }, [user.isAuthenticated, unread, log]);
```
flodolo: Same here: is this unnecessary?

flodolo:
I seem to have introduced a bug for "unread notifications", where this action is sent multiple times.

translate/src/modules/user/components/UserNotificationsMenu.tsx
flodolo: Again, trying to wrap my head about React, and failing 😵‍💫 

I think this was OK before, because the log was sent once at startup (could we potentially log the wrong number of unread notifications though?). But now it depends on authentication status.

The "multiple notifications" was caused by the log changing each render, so the solution is stabilize that across re-renders?

translate/src/modules/user/components/UserNotificationsMenu.tsx
```diff
@@ -71,30 +71,32 @@ export function UserNotificationsMenu({
...
   const handleDiscard = useCallback(() => setVisible(false), []);
 
+  const log = useLogUXAction();
+
   const unread = user?.notifications.has_unread;
 
   useEffect(() => {
     if (user.isAuthenticated && unread) {
-      logUXAction('Render: Unread notifications icon', 'Notifications 1.0', {
+      log('Render: Unread notifications icon', 'Notifications 1.0', {
```
eemeli: This is already inside a check for `user.isAuthenticated`, so could/should call `logUXAction()` directly, as it was doing before.

translate/src/modules/entities/actions.ts
eemeli: Introducing a state dependency to a dispatch like this is a bit messy, and we should avoid it. We can clean this up by instead by moving the `logUXAction()` call to the only call site for `getEntities()` in the EntitiesList component, where `isAuthUser` is already determined from the state.

translate/src/modules/machinery/components/source/GoogleTranslation.tsx
```diff
@@ -44,7 +45,7 @@ export function GoogleTranslation({
       restoreOriginal(translation);
     } else {
       await transformLLMTranslation(translation, characteristic, locale.name);
-      logUXAction('LLM Dropdown Select', 'LLM Feature Adoption', {
+      log('LLM Dropdown Select', 'LLM Feature Adoption', {
```
eemeli: LLM translations are only shown for authenticated users, so this doesn't need the isAuthenticated check.

translate/src/modules/machinery/components/MachineryTranslation.tsx
```diff
@@ -53,13 +54,13 @@ export function MachineryTranslationComponent({
         : translation.sources;
       setEditorFromHelpers(content, sources, true);
       if (llmTranslation) {
-        logUXAction('LLM Translation Copied', 'LLM Feature Adoption', {
+        log('LLM Translation Copied', 'LLM Feature Adoption', {
```
eemeli: LLM translations are only shown for authenticated users, so this doesn't need the isAuthenticated check.

translate/src/modules/translationform/utils/editFieldShortcuts.ts
```diff
@@ -126,14 +127,10 @@ export function useHandleCtrlShiftArrow(): (
...
       if (llmState.llmTranslation) {
-        logUXAction(
-          'LLM Translation Copied via Shortcut',
-          'LLM Feature Adoption',
-          {
-            action: 'Copy LLM Translation via Shortcut',
-            localeCode: locale.code,
-          },
-        );
+        log('LLM Translation Copied via Shortcut', 'LLM Feature Adoption', {
```
eemeli: LLM translations are only shown for authenticated users, so this doesn't need the isAuthenticated check.

translate/src/modules/entitieslist/components/EntitiesList.tsx
```diff
@@ -246,8 +247,22 @@ export function EntitiesList(): React.ReactElement<'div'> {
     if (!fetching) {
       // Currently shown entities should be excluded from the next results.
       dispatch(getEntities(location, page));
+      if (isAuthUser) {
```
eemeli: To keep the same behaviour as before, this should only trigger if search is non-empty.
```suggestion
      if (isAuthUser && location.search) {
```

translate/src/modules/entitieslist/components/EntitiesList.test.js
```diff
@@ -156,4 +163,36 @@ describe('<EntitiesList>', () => {
...
+      entities: ENTITIES,
+      hasMore: false,
+    });
+
+    createDefaultUser(store, { is_authenticated: true });
+
+    const wrapper = mountComponentWithStore(EntitiesList, store);
+
+    expect(wrapper.find('.entities')).toHaveLength(1);
+    expect(mockLogUXAction).not.toHaveBeenCalled();
```
eemeli: It's hard to say if this test is doing anything if we've no test where the logging _is_ being called.

flodolo: Meh, this was dumb. I'll see if I can add a better test and actually check the parameters, otherwise I'll drop it, since we were not testing this anyway (although that's what got us in this situation…)

flodolo:
Sorry for the churn with this. While trying to update the tests, I realized that we log on `getEntities()`, which means we log not only when the user searches, but also on scrolling.

I believe that behavior is incorrect: we want to measure the options used when searching, it's not useful to know if the user scrolls through the same results.

I moved the log to the loading of the component, and updated tests (hopefully correctly this time 🤦🏼).

flodolo:
Side note: while the tests were bad, I've been testing the code authenticated and not authenticated in my local docker installation.

eemeli (approved):
This looks pretty good, and likely fixes the scrolling issue as well.

mathjazz (approved):

PR #3948, opened 2026-02-05
Fix #3417: Add data migration to create project_managers group

### LINKED ISSUE #3417: project_managers group is not created by default
This feels like a bug, since the group is then used to set up point of contact in projects?

first-afk:
Fix #3417.

Creates a data migration file that checks if project_managers group exists; if not, it creates it. Checks if can_mange_project permission exists; if not, the migration file creates it. Finally, grant project_managers can_mange_project permission.

### DIFF
--- a/pontoon/base/migrations/0106_create_project_managers_group.py
+++ b/pontoon/base/migrations/0106_create_project_managers_group.py
@@ -25,9 +25,10 @@ def create_pm_group(apps, schema_editor):
         model="project",
     )
 
-    permission = Permission.objects.get(
+    permission, _ = Permission.objects.get_or_create(
         codename="can_manage_project",
         content_type=project_content_type,
+        defaults={"name": "Can manage project"},
     )
 
     group.permissions.add(permission)


pontoon/base/migrations/0106_create_project_managers_group.py
```diff
@@ -0,0 +1,51 @@
...
+    # 1. Get the models we need from the historical state
+    Group = apps.get_model("auth", "Group")
+    Permission = apps.get_model("auth", "Permission")
+    ContentType = apps.get_model("contenttypes", "ContentType")
+    Project = apps.get_model("base", "Project")
+
+    # 2. Get or Create the 'project_managers' group
+    group, created = Group.objects.get_or_create(name="project_managers")
+    if created:
+        print("Created 'project_managers' group.")
```
mathjazz: Please use `log.info()` instead of `print()`.

pontoon/base/migrations/0106_create_project_managers_group.py
```diff
@@ -0,0 +1,51 @@
...
+
+    # 3. Ensure the 'can_manage_project' permission exists
+    # We find the ContentType ID for the Project model
+    project_content_type = ContentType.objects.get_for_model(Project)
+
+    permission, perm_created = Permission.objects.get_or_create(
+        codename="can_manage_project",
+        content_type=project_content_type,
+        defaults={"name": "Can manage project"},
+    )
```
mathjazz: Using `ContentType.objects.get_for_model()` in Django migrations is not inherently "safe", because it relies on the live application state rather than the historical migration state.

Also, we should not create the permission in the migration, since it's already created in the model.

```suggestion
    # 3. Fetch the permission declared on Project.Meta.permissions
    project_content_type = ContentType.objects.get(app_label="base", model="project")

    permission = Permission.objects.get(
        codename="can_manage_project",
        content_type=project_content_type,
    )
```

pontoon/base/migrations/0106_create_project_managers_group.py
```diff
@@ -0,0 +1,51 @@
...
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("base", "0105_fix_mf2_translations"),
+        ("auth", "0001_initial"),
+        ("contenttypes", "0001_initial"),
+    ]
+
+    operations = [
+        migrations.RunPython(create_pm_group, remove_pm_group),
```
mathjazz: I would drop `remove_pm_group()`, because it seems risky. It deletes the whole group, even if it existed before this migration. 

```suggestion
        migrations.RunPython(create_pm_group, reverse_code=migrations.RunPython.noop),
```

pontoon/base/migrations/0106_create_project_managers_group.py
mathjazz: Please use the comment and `permission` definition as suggested above.

mathjazz (commented):
Thanks for the update! Please check again suggested changes in the unresolved comment.

Please also make sure tests don't fail.

pontoon/base/migrations/0106_create_project_managers_group.py
```diff
@@ -0,0 +1,46 @@
...
+
+log = logging.getLogger(__name__)
+
+
+def create_pm_group(apps, schema_editor):
+    Group = apps.get_model("auth", "Group")
+    Permission = apps.get_model("auth", "Permission")
+    ContentType = apps.get_model("contenttypes", "ContentType")
+
+    emit_post_migrate_signal(2, False, "default")
```
flodolo: Can this be the cause of the failures?

mathjazz: What is the purpose of this line? It should be removed.

pontoon/base/migrations/0106_create_project_managers_group.py
```diff
@@ -0,0 +1,46 @@
...
+
+    group, created = Group.objects.get_or_create(name="project_managers")
+    if created:
+        log.info("Created 'project_managers' group.")
+    else:
+        log.info("'project_managers' group already exists.")
+
```
mathjazz, line 22: If the group already exists, we should quit early (without adding the permission).

```suggestion
    if not created:
        log.info("'project_managers' group already exists.")
        return

    log.info("Created 'project_managers' group.")
```

pontoon/base/migrations/0106_create_project_managers_group.py
```diff
@@ -0,0 +1,46 @@
...
+    group.permissions.add(permission)
+    log.info(f"Added 'can_manage_project' permission to {group.name}")
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("base", "0105_fix_mf2_translations"),
+        ("auth", "0001_initial"),
+        ("contenttypes", "0001_initial"),
+        ("sites", "0001_initial"),
```
mathjazz: I don't think these are needed?

```suggestion
```

pontoon/base/migrations/0106_create_project_managers_group.py
```diff
@@ -0,0 +1,46 @@
...
+
+    group.permissions.add(permission)
+    log.info(f"Added 'can_manage_project' permission to {group.name}")
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("base", "0105_fix_mf2_translations"),
+        ("auth", "0001_initial"),
+        ("contenttypes", "0001_initial"),
```
mathjazz:
```suggestion
```

pontoon/base/migrations/0106_create_project_managers_group.py
```diff
@@ -0,0 +1,46 @@
...
+    log.info("Created 'project_managers' group.")
+
+    project_content_type = ContentType.objects.get_or_create(
+        app_label="base", model="project"
+    )
+
+    permission, _ = Permission.objects.get_or_create(
+        codename="can_manage_project",
+        content_type=project_content_type,
+        defaults={"name": "can manage project"},
```
mathjazz:
```suggestion
```

pontoon/base/migrations/0106_create_project_managers_group.py
```diff
@@ -0,0 +1,43 @@
...
+    ContentType = apps.get_model("contenttypes", "ContentType")
+
+    group, created = Group.objects.get_or_create(name="project_managers")
+    if not created:
+        log.info("'project_managers' group already exists.")
+        return
+
+    log.info("Created 'project_managers' group.")
+
+    project_content_type = ContentType.objects.get_or_create(
```
mathjazz:
```suggestion
    project_content_type = ContentType.objects.get(
```

pontoon/base/migrations/0106_create_project_managers_group.py
```diff
@@ -0,0 +1,43 @@
...
+        app_label="base", model="project"
+    )
+
+    permission, _ = Permission.objects.get_or_create(
```
mathjazz:
```suggestion
    permission = Permission.objects.get(
```

pontoon/base/migrations/0106_create_project_managers_group.py
```diff
@@ -0,0 +1,43 @@
...
+    group, created = Group.objects.get_or_create(name="project_managers")
+    if not created:
+        log.info("'project_managers' group already exists.")
+        return
+
+    log.info("Created 'project_managers' group.")
+
+    project_content_type = ContentType.objects.get(
+        app_label="base", model="project"
+    )
```
mathjazz:
```suggestion
    project_content_type = ContentType.objects.get(app_label="base", model="project")
```

pontoon/base/migrations/0106_create_project_managers_group.py
```diff
@@ -0,0 +1,41 @@
...
+    ContentType = apps.get_model("contenttypes", "ContentType")
+
+    group, created = Group.objects.get_or_create(name="project_managers")
+    if not created:
+        log.info("'project_managers' group already exists.")
+        return
+
+    log.info("Created 'project_managers' group.")
+
+    project_content_type = ContentType.objects.get(app_label="base", model="project")
```
mathjazz:
```suggestion
    project_content_type, _ = ContentType.objects.get_or_create(
        app_label="base",
        model="project",
    )
```

pontoon/base/migrations/0106_create_project_managers_group.py
```diff
@@ -0,0 +1,44 @@
...
+
+    project_content_type, _ = ContentType.objects.get_or_create(
+        app_label="base",
+        model="project",
+    )
+
+    permission = Permission.objects.get(
+        codename="can_manage_project",
+        content_type=project_content_type,
+    )
```
mathjazz:
```suggestion
    permission, _ = Permission.objects.get_or_create(
        codename="can_manage_project",
        content_type=project_content_type,
        defaults={"name": "Can manage project"},
    )
```

PR #3964, opened 2026-02-13
Guard against deleted actor in notifications

flodolo:
For example, review notifications don't have an actor defined.

### DIFF
--- a/pontoon/contributors/templates/contributors/widgets/notifications_menu.html
+++ b/pontoon/contributors/templates/contributors/widgets/notifications_menu.html
@@ -56,7 +56,7 @@
           {% set actor_anchor = notification.actor.name_or_email|nospam %}
           {% set actor_url = full_url('pontoon.contributors.contributor.username', notification.actor.username) %}
         {% else %}
-          {% set actor_anchor = "[Deleted User]" %}
+          {% set actor_anchor = "Deleted User" %}
           {% set has_actor = false %}
         {% endif %}
 
--- a/translate/src/modules/user/components/UserNotification.tsx
+++ b/translate/src/modules/user/components/UserNotification.tsx
@@ -68,7 +68,7 @@ const Comment = ({
     {actor ? (
       <span className='actor'>{actor.anchor}</span>
     ) : (
-      <span className='actor deleted'>[Deleted User]</span>
+      <span className='actor deleted'>Deleted User</span>
     )}
 
     <span className='verb'>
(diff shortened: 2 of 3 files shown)


pontoon/contributors/templates/contributors/widgets/notifications_menu.html
```diff
@@ -54,6 +54,8 @@
         {% elif notification.actor.email %}
           {% set actor_anchor = notification.actor.name_or_email|nospam %}
           {% set actor_url = full_url('pontoon.contributors.contributor.username', notification.actor.username) %}
+        {% else %}
+          {% set actor_anchor = "(Deleted user)" %}
```
flodolo: This is showing empty on DEV, but locally it shows `{{ actor_anchor }}`.

translate/src/modules/user/components/UserNotification.tsx
```diff
@@ -65,13 +65,13 @@ const Comment = ({
   verb,
 }: Props['notification']) => (
   <div className='item-content'>
-    <span className='actor'>{actor.anchor}</span>
+    <span className='actor'>{actor ? actor.anchor : '(Deleted user)'}</span>
```
mathjazz: For [consistency](https://github.com/mozilla/pontoon/blob/main/pontoon/base/admin.py#L75), let's use Deleted User. I'd also add a `deleted` class and set `color: var(--light-grey-2)`, or at least use square brackets.

```suggestion
    <span className='actor'>{actor ? actor.anchor : '(Deleted User)'}</span>
```

pontoon/contributors/templates/contributors/widgets/notifications_menu.html
mathjazz: Hm, I see `(Deleted user)` locally, which is expected.

translate/src/modules/user/components/UserNotification.tsx
```diff
@@ -65,13 +65,13 @@ const Comment = ({
...
 
     <span className='verb'>
-      <a href={target.url}>{verb}</a>
+      {target ? <a href={target.url}>{verb}</a> : verb}
```
mathjazz: Is this change (and the one below) related to the PR?

pontoon/contributors/templates/contributors/widgets/notifications_menu.html
flodolo: Without this patch and with the snippet I provided to generate the notification?

Image

translate/src/modules/user/components/UserNotification.tsx
flodolo: No, they sneaked in after the initial commit (I was testing another issue, and missed I committed them).

PR #3965, opened 2026-02-13
Community Health: Obsolete resources and corresponding entities

### LINKED ISSUE #2133: Obsolete Resources instead of deleting them

functionzz:
## Summary
This PR addresses the data loss incurred when someone removes strings from projects like Firefox. Instead of obsoleting the strings, we currently delete the data which causes contributior attribution to be lost.

Fixes #2133.


pontoon/base/models/entity.py
```diff
@@ -251,6 +251,14 @@ def prefetch_entities_data(self, locale: Locale, preferred_source_locale: str):
 
         return entities
 
+    def obsolete(self, now):
+        entities = list(self)
+        for entity in entities:
+            entity.obsolete = True
+            entity.date_obsoleted = now
+            entity.section = None
+        Entity.objects.bulk_update(entities, ["obsolete", "date_obsoleted", "section"])
```
mathjazz: Why not simply this?

```suggestion
        self.update(
            obsolete=True,
            date_obsoleted=now,
            section=None,
        )
```

mathjazz (commented):
Nice work!

Please also make sure that everywhere in the code where we references **all** resources (and count on the fact that obsolete resources are currently not present), we now reference **active** resources only.

pontoon/sync/core/entities.py
```diff
@@ -340,12 +342,13 @@ def update_translated_resources(
 ) -> None:
     prev_tr_keys: set[tuple[int, int]] = set(
         (tr["resource_id"], tr["locale_id"])
-        for tr in TranslatedResource.objects.filter(resource__project=project)
```
functionzz, line 345: For sync related items, should I be changing the `TranslatedResource` objects or should they be the same?

pontoon/sync/core/translations_from_repo.py
```diff
@@ -115,9 +115,9 @@ def delete_removed_gettext_resources(
             Translation.objects.filter(entity__resource__project=project).filter(
                 rm_t
             ).delete()
-            TranslatedResource.objects.filter(resource__project=project).filter(
-                rm_tr
-            ).delete()
+            TranslatedResource.objects.current().filter(
```
functionzz: Same question for this part.

pontoon/sync/core/translations_from_repo.py
```diff
@@ -178,9 +178,9 @@ def find_db_updates(
 
     resources: dict[str, Resource] = {
         res.path: res
-        for res in Resource.objects.filter(
-            project=project, path__in=resource_paths
-        ).iterator()
+        for res in Resource.objects.current()
```
functionzz, line 182: Same question for this part.

pontoon/base/models/resource.py
```diff
@@ -3,6 +3,17 @@
 from django.utils import timezone
 
 
+class ResourceQuerySet(models.QuerySet):
+    def obsolete(self, now):
```
mathjazz: I'd call it `mark_as_obsolete()`, otherwise somebody might thing we are returning obsolete Resources only.

pontoon/base/models/translation.py
```diff
@@ -310,8 +312,10 @@ def save(self, failed_checks=None, *args, **kwargs):
             self.entity.reset_term_translation(self.locale)
 
         # We use get_or_create() instead of just get() to make it easier to test.
-        translatedresource, created = TranslatedResource.objects.get_or_create(
-            resource=self.entity.resource, locale=self.locale
+        translatedresource, created = (
+            TranslatedResource.objects.current().get_or_create(
```
mathjazz: That looks like a mistake - we shouldn't make this change.

pontoon/terminology/models.py
```diff
@@ -126,7 +126,7 @@ def create_entity(self):
         - Entity.comment contains joint content of several fields:
           Term.part_of_speech. Term.definition. E.g.: Term.usage.
         """
-        resource = Resource.objects.get(project__slug="terminology")
+        resource = Resource.objects.current().get(project__slug="terminology")
```
mathjazz: No need for `current()` here.

mathjazz (commented):
Thanks for the update!

This will need extensive testing before we ship it.

pontoon/base/models/translation.py
functionzz: Are you saying in this case adding `.current()` to `TranslatedResource` is a mistake, or all `TranslatedResource` shouldn't have `.current()`? For context, I also added `.current()` to `TranslatedResource` so cases that do not directly reference `Resources` can also exclude `obsolete` Resources.

mathjazz: There is no need for `.current()` in this particular case.

mathjazz:
> I guess the next natural step is writing tests for the various cases listed here if everything looks fine.

That would be great!

pontoon/administration/tests/test_views.py
```diff
@@ -380,6 +380,32 @@ def test_manage_project_strings_download_csv(client_superuser):
     assert "Mächt’ge".encode() in response.content
 
 
+@pytest.mark.django_db
+def test_manage_project_translate_link_excludes_obsolete_resources(client_superuser):
+    """Test that translate_locale is only set when non-obsolete resources exist."""
+    locale_kl = LocaleFactory.create(code="kl", name="Klingon")
```
eemeli: While I'm aware that there's existing test code that uses `kl` as "Klingon", let's not expand this mistake.

The actual locale code for Klingon is `tlh`, for [tlhIngan Hol](https://en.wikipedia.org/wiki/Klingon_language). `kl` is the locale for [Kalaallisut](https://en.wikipedia.org/wiki/Greenlandic_language), i.e. Greenlandic.

pontoon/base/models/entity.py
```diff
@@ -251,6 +251,13 @@ def prefetch_entities_data(self, locale: Locale, preferred_source_locale: str):
 
         return entities
 
+    def obsolete(self, now):
```
eemeli: It's not clear that this is a setter rather than a getter, so let's rename it.

The second argument should also be optional, to be set from `timezone.now()` if it's `None`.

OTOH, if/as this is only ever called from `ResourceQuerySet.mark_as_obsolete()` then it doesn't necessarily make sense for this to exist at all, and for the `EntityQuerySet.update()` call to be inlined there.

```suggestion
    def set_obsolete(self, now = None):
```

pontoon/base/models/resource.py
```diff
@@ -3,6 +3,17 @@
 from django.utils import timezone
 
 
+class ResourceQuerySet(models.QuerySet):
+    def mark_as_obsolete(self, now):
```
eemeli: This name also works for the EntityQuerySet method; pick one, and use it for both.

Also, same thing about the second argument being optional.

pontoon/base/tests/managers/test_user.py
```diff
@@ -322,3 +325,69 @@ def test_mgr_user_query_args_filtering(
...
+    entities = EntityFactory.create_batch(size=12, resource=resource_a)
+
+    batch_kwargs = sum(
+        [
+            [dict(approved=True)] * 7,
+            [dict(approved=False, fuzzy=False, rejected=True)] * 3,
+            [dict(fuzzy=True)] * 2,
+        ],
+        [],
+    )
```
eemeli: Using `sum()` here is weird.
```suggestion
    batch_kwargs = (
        [dict(approved=True)] * 7 +
        [dict(approved=False, fuzzy=False, rejected=True)] * 3 +
        [dict(fuzzy=True)] * 2
    )
```

pontoon/sync/tests/test_entities.py
```diff
@@ -54,13 +55,17 @@ def test_remove_resource():
...
+            entity=entity_c, locale=locale, string="Bonjour"
+        )
 
         # Filesystem setup
         makedirs(repo.checkout_path)
         build_file_tree(
             repo.checkout_path,
             {
-                "en-US": {"a.ftl": "", "b.pot": ""},
+                "en-US": {"a.ftl": "", "b.po": ""},
```
eemeli: Why this change? `.pot` is the appropriate extension for gettext template files.

pontoon/sync/tests/test_entities.py
```diff
@@ -79,9 +84,18 @@ def test_remove_resource():
...
-            TranslatedResource.objects.get(resource=res_c)
+        assert {res.path for res in project.resources.all()} == {
+            "a.ftl",
+            "b.po",
+            "c.ftl",
+        }
+        assert project.resources.count() == 3
+        assert project.resources.get(path="c.ftl").obsolete is True
+        assert project.resources.get(path="a.ftl").obsolete is False
+        assert project.resources.get(path="b.po").obsolete is False
```
eemeli:
```suggestion
        assert {res.path: res.obsolete for res in project.resources.all()} == {
            "a.ftl": False,
            "b.po": False,
            "c.ftl": True,
        }
        assert {res.path for res in project.resources.current()} == {
            "a.ftl",
            "b.po",
        }
```

eemeli (changes requested):
Looks good; see inline for some relatively minor things.

> I would like some guidance on how to test these `.current()` properties. Since all previous tests pass. I'm finding it difficult to come up with tests without just testing how `.current()` works over and over again.

Eh, What's here is already pretty decent. I would like to see something exercising `TranslatedResource.objects.current()` after `ResourceQuerySet.mark_as_obsolete()` is called, but that's about it.

pontoon/sync/tests/test_entities.py
functionzz: I thought this was indeed a typo for `.po`, oops.

pontoon/base/models/entity.py
functionzz: I decided to remove the method from `Entity` and called `.update` directly.

pontoon/base/tests/managers/test_user.py
functionzz: sorry :) mindless copy paste from above function.

eemeli (approved):

eemeli (changes requested):
When eventually re-requesting a review from me on this, please include a link to a diff from my previous review to the updated state, so I don't need to re-review everything.

For future reference, rebasing and force-pushing a reviewed PR is not well handled by GitHub's web UI; to account for changes made elsewhere during the lifetime of a PR, merging from master to the PR's branch works _much_ better.

eemeli (approved):

mathjazz:
Thanks for the hard work, Jamie!

I created the Pontoon Test project locally on your branch, synced, removed a file in https://github.com/mozilla-l10n/pontoon-test/commit/9df781730910516d6d83a95e7f55f08fed98d331 and synced again and the file was removed from the DB, the stats were updated and everything seemed to be working fine.

Then I added the file back in https://github.com/mozilla-l10n/pontoon-test/commit/101dcfb8d02a80c250bf616df4fa2f4beb6e2d33 and synced again, but the file didn't get imported. Seems like we don't de-obsolete obsoleted Resources?

functionzz:
> Then I added the file back in pontoon-test and synced again, but the file didn't get imported. Seems like we don't de-obsolete obsoleted Resources?
 
I see, regardless of whether we de-obsolete resources or create a new one each time, we should be importing the resource in so this is definitely a major issue.

I think what I'm going to do for the case of a resource added back in is this:

1. De-obsolete the resource
2. Conduct a union on the obsoleted entities and new entities - de-obsolete the entities within the new entities set and add new entities (which are obviously untranslated). Also, removed entities in the new entites would NOT be deleted, and will remain obsoleted.

Let me know what you think.

pontoon/sync/core/entities.py
```diff
@@ -279,11 +281,46 @@ def add_resources(
...
+
+    if not valid_updates:
+        return 0, set()
+
+    # Resource de-obsoletion for existing Resources that are added back in
+    # TODO Entity de-obsoletion needs to accompany Resource de-obsoletion
+    if deobsoletion_paths:
+        Resource.objects.filter(project=project, path__in=deobsoletion_paths).update(
+            obsolete=False
+        )
```
eemeli: Why is this work happening here in `add_resources()`, and not in `update_resources()`? That seems like a better place where to update resources.

eemeli (changes requested):

pontoon/sync/core/entities.py
functionzz: I've thought about that, and the reason is while the above code snippet is definitely an update, the `add_resources` function needs to be responsible in detecting resources that exist that are added back in (i.e de-obsoletion). If I put the de-obsoletion logic in `update_resources()`, I would need to duplicate the `valid_updates` logic into `add_resources()` anyway to filter out the de-obsoleted resources.

eemeli: If the de-obsoletion happens in `update_resources()`, won't the de-obsoleted resources be included in `changed_paths`? I still don't see why any of the code in `add_resources()` needs to change.

eemeli (approved):

mathjazz (approved):
> Then I added the file back in https://github.com/mozilla-l10n/pontoon-test/commit/101dcfb8d02a80c250bf616df4fa2f4beb6e2d33 and synced again, but the file didn't get imported. Seems like we don't de-obsolete obsoleted Resources?

I tested this again and it works fine now. Great work, this is a massive improvement! 👍

PR #3966, opened 2026-02-13
Fix search functionality when list parameter is present

### LINKED ISSUE #3878: Search broken when the `list` query string parameter is used
From a notification entry I landed on a URL such as `https://pontoon.mozilla.org/eu/firefox-for-android/all-resources/?list=327882%2C327894%2C327892%2C327886%2C327923%2C327924%2C324461%2C324004%2C326177%2C327105%2C327363%2C327395%2C327912`, which contains a comma-separated list of strings to go through.

I started reviewing the strings, and at one point I wanted to perform a search. The search did nothing but refresh the list of strings and the editor, leaving everything in the same state as before.

I would have expected the search to behave the same as in other filtered contexts, where I get results scoped to the constrained context, or "No results" when nothing is found.

Check the following recording for reference:
<video src="https://github.com/user-attachments/assets/94f9d98c-e024-461c-add2-658f15578e26" />

MundiaNderi:
Fixes #3878 
Fix search functionality when list parameter is present

Previously, when users navigated to a URL with a list parameter (e.g., from notifications), the search functionality would refresh the page without applying any filters.

Changes:
- Modified parse() in Location.tsx to allow for search and list to coexist
- Modified stringify() to include search params along list
- Modified buildFetchPayload() to send search with list
- Added search filtering to _get_entities_list() backend function
- Updated selectEntity() to preserve search and list in URL

To test:
- Navigate to this URL: http://localhost:8000/sl/pontoon-test/all-resources/?list=1,2,3,4,5
- Conduct a search.  The below should be your views.

Screenshot 2026-02-13 at 20 14 24
Screenshot 2026-02-13 at 20 14 54

### DIFF
--- a/translate/src/modules/entitieslist/components/EntitiesList.test.js
+++ b/translate/src/modules/entitieslist/components/EntitiesList.test.js
@@ -12,7 +12,7 @@ import {
 } from '~/test/store';
 
 import { EntitiesList } from './EntitiesList';
-import { vi } from 'vitest';
+import { expect, vi } from 'vitest';
 
 // Entities shared between tests
 const ENTITIES = [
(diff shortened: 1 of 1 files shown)


mathjazz (approved):
Well done!

PR #3968, opened 2026-02-14
 Community Health: change completion formula

functionzz:
This PR addresses Part 2 of the Community Health OKR where we change completion metrics to exclude pretranslations, which will provide a more accurate picture of if a project/locale is finished being localized.

### DIFF
--- a/pontoon/base/static/js/progress-chart.js
+++ b/pontoon/base/static/js/progress-chart.js
@@ -21,17 +21,13 @@ $(function () {
 
     const fraction = {
         translated: stats.all ? stats.translated / stats.all : 0,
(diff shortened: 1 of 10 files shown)


pontoon/base/models/locale.py
```diff
@@ -86,7 +86,7 @@ def stats_data(self, project=None):
             - F("warnings"),
             is_complete=Case(
```
flodolo, line 87: I can't help wondering if we should put this definition in one place and import it, instead of defining it in each model (risk of crossing imports?).

flodolo:
Note that you need to rebase and resolve conflicts.

mathjazz (approved):

PR #3977, opened 2026-02-18
Translate: fix unauthenticated AJAX calls

functionzz:
Addresses [this bugzilla report](https://bugzilla.mozilla.org/show_bug.cgi?id=2013660).

Places AJAX calls behind `Project.objects.visible()` filter, preventing unauthorized access of data relating to private projects. Chose `Project.objects.visible_for(user)`.

UNRELATED: also discovered redundant filter in `EntityListView` and `EntityIndividualView`. Changed in this PR for expediency.

### DIFF
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -424,18 +424,12 @@ def get_translation_history(request):
             status=400,
         )
 
-    entity = get_object_or_404(Entity, pk=entity)
+    visible_projects = Project.objects.visible().visible_for(request.user)
+    entities = Entity.objects.filter(resource__project__in=visible_projects)
+
+    entity = get_object_or_404(entities, pk=entity)
     locale = get_object_or_404(Locale, code=locale)
     project = entity.resource.project
-
-    if not Project.objects.filter(pk=project.pk).visible_for(request.user).exists():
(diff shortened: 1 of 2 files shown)


pontoon/base/views.py
```diff
@@ -333,7 +333,12 @@ def get_translations_from_other_locales(request):
             status=400,
         )
 
-    entity = get_object_or_404(Entity, pk=entity)
+    visible_projects = Project.objects.visible().visible_for(request.user)
+    entities = Entity.objects.filter(
+        resource__project__in=visible_projects
+    ).prefetch_related("resource")
```
mathjazz: Why do you need prefetch here and below?

functionzz: I think that was a remnant of a copy paste from EntityListView - oops

mathjazz:
> I was considering using this code, but figured it would be more consistent if 404 Not Found was returned instead of 403 Unauthorized. I can change it to this version however.

That works, too. Let's be consisent!

functionzz:
I actually like the if statement way, so I just changed up the error code and the message.

mathjazz (approved):
Nice work!

PR #3978, opened 2026-02-18
API: change ProjectListView filter logic

functionzz:
This PR prevents regular users from viewing private projects in the `api/v2/projects/?include_disabled=true&include_system=true` endpoint, and allows admin/superuser viewing only.

See:
https://bugzilla.mozilla.org/show_bug.cgi?id=2009196#c13

### DIFF
--- a/pontoon/api/tests/test_views.py
+++ b/pontoon/api/tests/test_views.py
@@ -710,6 +710,7 @@ def test_project_visibility(django_assert_num_queries, client_superuser):
 
         assert response.status_code == 200
 
+    # 8 projects includes Tutorial & Terminology
     assert response.data["count"] == 8
     for res in response.data["results"]:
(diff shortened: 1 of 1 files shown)


pontoon/api/tests/test_views.py
```diff
@@ -695,6 +695,34 @@ def test_disabled_projects(
...
+
+    url = "/api/v2/projects/?fields=slug,visibility&include_disabled=true&include_system=true"
+
+    # public API usage
+    with django_assert_num_queries(2):
+        response = APIClient().get(url)
+
+        assert response.status_code == 200
+
+    assert response.data["count"] == 8
```
mathjazz, line 714: We should add a comment why the number is 8 (rather than 6), because it's not obvious.

PR #3980, opened 2026-02-19
Exclude system projects from suggestion notifications

### LINKED ISSUE #3949: Don't send notifications for the tutorial project
Currently, we're sending locale managers for notifications when there are new suggestions, which is not particularly useful.

MundiaNderi:
Fixes #3949 

Locale managers were receiving notifications about unreviewed suggestions in system projects (like Tutorial), which is not useful since these are not real translation projects.

Change:
Added `entity__resource__project__system_project=False `to the `get_suggestions()` query in send_suggestion_notifications.py, which excludes translations belonging to system projects from the notification pipeline.

Testing:
I added a unit test that fails before the change and passes after the change in  `pontoon/messaging/tests/test_send_notifications.py`

### DIFF
--- a/pontoon/messaging/tests/test_send_notifications.py
+++ b/pontoon/messaging/tests/test_send_notifications.py
@@ -8,8 +8,6 @@ from pontoon.test.factories import (
     TranslationFactory,
 )
 
-# For bug #3949: https://github.com/mozilla/pontoon/issues/3949
-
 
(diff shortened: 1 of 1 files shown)


functionzz (approved):
This looks good, 
tested locally - works well, 
written test makes sense.

Nice :)

PR #3982, opened 2026-02-19
API: fix swagger ui

### LINKED ISSUE #3979: Swagger UI broken
See: https://pontoon.mozilla.org/api/v2/

I broke that in https://github.com/mozilla/pontoon/pull/3972.

Ideally, we should not allow everything from `https://cdn.jsdelivr.net`, just this specific script.

functionzz:
This PR reinstates the Swagger docs UI, which was not loading due to the deletion of CSP in #3972. It uses `drf-spectacular[sidecar]` to avoid dealing with CSP.
The documentation for using `drf-spectacular[sidecar]` to avoid this issue can be found [here](https://drf-spectacular.readthedocs.io/en/latest/faq.html#my-swagger-ui-and-or-redoc-page-is-blank).

Fixes #3979.

### DIFF
--- a/pontoon/settings/base.py
+++ b/pontoon/settings/base.py
@@ -283,6 +283,7 @@ INSTALLED_APPS = (
     "django_ace",
     "rest_framework",
     "drf_spectacular",
+    "drf_spectacular_sidecar",
 )
 
 # A list of IP addresses or IP ranges to be blocked from accessing the app,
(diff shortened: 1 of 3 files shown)


mathjazz:
> No but it was a close call.

Sorry about that - my bad completely! I'll share more suggestions on Matrix.

PR #3987, opened 2026-02-20
Search: default non-visible locales in locale selector to en-GB

### LINKED ISSUE #3984: Pontoon search: some users default to en-US locale, which doesn't return results
Pontoon auto-detects a users locale, but for users who are determined to be in the en-US locale, searching for anything without changing the locale results in no matches found.

Image

functionzz:
This PR defaults locales like `en-US` to `en-GB`. This happens when Pontoon auto-detects locales based on the browser, which can return locale codes that are unsupported by the `/api/v2/locales/` API endpoint.

Fixes #3984.

### DIFF
--- a/pontoon/search/views.py
+++ b/pontoon/search/views.py
@@ -16,6 +16,14 @@ from pontoon.base.utils import get_project_locale_from_request, parse_bool, requ
 from pontoon.settings.base import SITE_URL
 
 
+def get_valid_locale_code(request, locale_code):
+    """Return a valid locale code, falling back to project locale or en-GB."""
+    visible_locales = Locale.objects.visible()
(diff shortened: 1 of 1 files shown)


pontoon/search/views.py
```diff
@@ -113,9 +117,13 @@ def search_results(request):
...
 
-    if not locale_code or not Locale.objects.filter(code=locale_code).exists():
+    if (
+        not locale_code
+        or not Locale.objects.visible().filter(code=locale_code).exists()
+    ):
         locale_code = (
-            get_project_locale_from_request(request, Locale.objects) or "en-GB"
+            get_project_locale_from_request(request, Locale.objects.visible())
+            or "en-GB"
```
mathjazz: This block of code is repeated twice. Can you factor it out?

PR #3993, opened 2026-02-25
Include last day of month in monthly locale stats

### LINKED ISSUE #3819: Monthly report: completion data and missing strings count are off
I got a report that Italian has 6 missing strings. Italian is at 100%, and has been for over 24h.

Firefox for iOS got 6 new strings at `2025-10-30 22:03:15.315327+00:00`. If I have to guess, Pontoon is looking at the data from beginning of day in the last month, instead of end of day.

Image

flodolo:
Fixes #3819

### DIFF
--- a/pontoon/messaging/emails.py
+++ b/pontoon/messaging/emails.py
@@ -78,14 +78,16 @@ def _get_monthly_locale_stats(months_ago):
     # use the snapshot from the first day of the current month.
     next_month = month_date + relativedelta(months=1)
 
-    snapshots = LocaleInsightsSnapshot.objects.filter(
-        created_at__day=1,
-        created_at__month=next_month.month,
-        created_at__year=next_month.year,
+    snapshots = list(
+        LocaleInsightsSnapshot.objects.filter(
+            created_at__day=1,
+            created_at__month=next_month.month,
+            created_at__year=next_month.year,
+        )
     )
 
-    if not snapshots.exists():
-        raise ValueError(
+    if not snapshots:
+        raise LookupError(
             f"No LocaleInsightsSnapshot found for "
             f"{next_month.year}-{next_month.month:02d}-01."
         )


flodolo:
The interval between taking the snapshot for the 1st of the month and sending the email is quite narrow. I've added a fallback in a separate commit. If that's not useful — the current code doesn't have any, it's just sending out `{}`— it's easy to revert.

pontoon/messaging/emails.py
```diff
@@ -72,14 +72,31 @@ def _get_monthly_locale_actions(months_ago):
...
+        # last day of the target month and log a warning.
+        fallback_date = (next_month - datetime.timedelta(days=1)).date()
+        log.warning(
+            f"No LocaleInsightsSnapshot found for "
+            f"{next_month.year}-{next_month.month:02d}-01. "
+            f"Falling back to {fallback_date}."
+        )
+        snapshots = LocaleInsightsSnapshot.objects.filter(
+            created_at=fallback_date,
+        )
```
mathjazz: That seems like a fix for a bad configuration. If the snapshots are not available, we should abort.

pontoon/messaging/emails.py
```diff
@@ -72,14 +72,31 @@ def _get_monthly_locale_actions(months_ago):
 
 def _get_monthly_locale_stats(months_ago):
     month_date = timezone.now() - relativedelta(months=months_ago)
-    last_day = calendar.monthrange(month_date.year, month_date.month)[1]
+
+    # collect_insights() takes data snapshots at midnight. To get the
+    # end-of-month state, use the snapshot from the first day of the following
+    # month.
+    next_month = month_date + relativedelta(months=1)
```
mathjazz: 01:00 AM actually:
https://github.com/mozilla/webservices-infra/blob/22aca1f6379a27d1cdf5a3e450ec7c38ffc07ad5/pontoon/k8s/pontoon/values-prod.yaml#L11

Since the exact time is configured outside the codebase, how about this (inspired by [this](https://github.com/flodolo/pontoon/blob/0a901715d2efc79dcd04d081b721293c01726a58/pontoon/insights/management/commands/collect_insights.py)):

```suggestion
    # collect_insights() takes data snapshots at the beginning of the day,
    # every day. To represent the state of the past month, we should
    # use the snapshot from the first day of the current month.
    next_month = month_date + relativedelta(months=1)
```

pontoon/messaging/emails.py
flodolo: Makes sense. I'll back out the 2nd commit.

flodolo: Actually, we want the log, so I'll just clean up.

flodolo:
(the force push is to rebase against main, kept the change in a separate commit)

pontoon/messaging/emails.py
```diff
@@ -72,14 +72,26 @@ def _get_monthly_locale_actions(months_ago):
...
-        created_at__month=month_date.month,
-        created_at__year=month_date.year,
+        created_at__day=1,
+        created_at__month=next_month.month,
+        created_at__year=next_month.year,
     )
 
+    if not snapshots.exists():
+        # Log a warning if 1st-of-month snapshot is not available.
+        log.warning(
```
flodolo: Error or warning?

mathjazz: I was thinking we should raise here, i.e. log an error and abort sending email.

flodolo: Are we sending an email in any other cases for errors?

flodolo: Misread your message: we should not send the email, not "abort and send an email about the failure".

Will fix in a minute.

pontoon/messaging/emails.py
```diff
@@ -72,14 +72,24 @@ def _get_monthly_locale_actions(months_ago):
...
-        created_at__day=last_day,
-        created_at__month=month_date.month,
-        created_at__year=month_date.year,
+        created_at__day=1,
+        created_at__month=next_month.month,
+        created_at__year=next_month.year,
     )
 
+    if not snapshots.exists():
+        raise ValueError(
```
mathjazz: I'd use the `LookupError` her or create a custom type.

pontoon/messaging/emails.py
```diff
@@ -72,14 +72,24 @@ def _get_monthly_locale_actions(months_ago):
...
     snapshots = LocaleInsightsSnapshot.objects.filter(
-        created_at__day=last_day,
-        created_at__month=month_date.month,
-        created_at__year=month_date.year,
+        created_at__day=1,
+        created_at__month=next_month.month,
+        created_at__year=next_month.year,
     )
 
+    if not snapshots.exists():
```
mathjazz: Nit - this would query the DB only once:

```python
snapshots = list(
    LocaleInsightsSnapshot.objects.filter(...)
)

if not snapshots:
    raise LookupError(...)

return {...}
```

PR #3994, opened 2026-02-25
Insights: store days without pretranslation reviews as NULL instead of 0

### LINKED ISSUE #3860: Insights: pretranslation acceptance rate is stable, but chrf++ score plummeted from September
I checked a few locales (it, fr, de) and they all show the same pattern, e.g. https://pontoon.mozilla.org/it/insights/

Image

This feels like a problem with the calculation?

flodolo:
Fixes #3860

### DIFF
--- a/pontoon/insights/tasks.py
+++ b/pontoon/insights/tasks.py
@@ -154,7 +154,7 @@ def count_activities(dt_max: datetime):
                         data.times_to_review_pretranslations.append(review_time)
                 if user in pretranslation_users:
                     data.pretranslations_rejected.add(translation)
-                    score = get_chrf_score(action, approved_translations)
+                    score = calculate_chrf_score(action, approved_translations)
                     if score is not None:
                         data.pretranslations_chrf_scores.append(score)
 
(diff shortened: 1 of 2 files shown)


pontoon/insights/tasks.py
```diff
@@ -155,7 +155,7 @@ def count_activities(dt_max: datetime):
                 if user in pretranslation_users:
                     data.pretranslations_rejected.add(translation)
                     score = get_chrf_score(action, approved_translations)
```
eemeli: We should rename one of the `get_chrf_score()` functions to something else. Atm we've two different ones, one in this file, and the other in `pontoon/insights/utils.py`.

pontoon/insights/tests/test_utils.py
eemeli, line 1: This is indeed unnecessarily detailed, as the function isn't used outside its own module.

flodolo: Do you suggest to drop it completely?

pontoon/insights/tasks.py
flodolo: Rename the one in tasks to `calculate_chrf_score()`, since that's what it does (the other gets it from the insights).

PR #3995, opened 2026-02-26
Fix #3614 Add ability to toggle between dashboard progress bar and detailed view

### LINKED ISSUE #3614: Add ability to toggle between progress bar and detailed stats on the dashboard
Pontoon dashboards show the progress bar for each project, locale, resource or tag, and reveal detailed stats on hover. Sometimes it would be more convenient to show the detailed stats for all entries in the table at the same time, e.g. when trying to find all localizations that contain warnings.

Let's add a toggle to switch between progress bar and detailed stats for all entries in the table.

dannycolin:
This adds a toggle button to always show the progress detailed view on the dashboard instead of only showing it on hover.

It currently doesn't implement a way to remember the user preference. This could be added in this patch or in a follow up patch. Any thought @mathjazz. 

image

Fixes #3614

### DIFF
--- a/pontoon/base/static/css/table.css
+++ b/pontoon/base/static/css/table.css
@@ -69,6 +69,12 @@ table.table.project-list.hidden {
   width: 16px;
 }
 
+.table th.progress button {
+  float: right;
+  margin: -3px -10px -3px 0;
+  padding: 3px 12px;
+}
+
 .table-sort th.unreviewed-status i {
   margin: -13px 0 0 13px;
 }
@@ -314,6 +320,14 @@ table.table.project-list.hidden {
 }
 
 .table tr:hover .progress .legend {
+  display: block;
+}
+
+.table .progress.show-detailed-view .chart-wrapper {
+  display: none;
+}
+
+.table .progress.show-detailed-view .legend {
   display: block;
 }
 
--- a/pontoon/base/static/js/table.js
+++ b/pontoon/base/static/js/table.js
@@ -116,6 +116,31 @@ var Pontoon = (function (my) {
       })(),
 
       /*
+       * showDetailedView
+       *
+       * Toggle button to show/hide the detailed view in the progress column of tables.
+       */
+      showDetailedView: (function () {
+        $('body').on(
+          'click',
+          'table th.progress button#detailed-view-toggle',
+          function (e) {
+            // Do not reorder the list when clicking on the toggle button
+            e.stopPropagation();
+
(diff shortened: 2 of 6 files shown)


pontoon/base/static/js/table.js
```diff
@@ -115,6 +115,34 @@ var Pontoon = (function (my) {
...
+              .forEach((el) => el.classList.toggle('show-detailed-view'));
+
+            const detailedViewToggle = document.getElementById(
+              'detailed-view-toggle',
+            );
+            detailedViewToggle.textContent =
+              detailedViewToggle.textContent.trim() === 'Show detailed view'
+                ? 'Hide detailed view'
+                : 'Show detailed view';
+          },
```
mathjazz, line 139: No strong preference, but since we use jQuery, I'd "jqueryfy" this code.

dannycolin: Even if it'd pain me, I can jquerify it.

pontoon/base/static/css/table.css
```diff
@@ -69,6 +69,14 @@ table.table.project-list.hidden {
   width: 16px;
 }
 
+.table th.progress button {
+  float: right;
+  margin-block-start: -3px;
+  margin-inline-end: -10px;
+  padding-block: 3px;
+  padding-inline: 12px;
```
mathjazz: This change will keep the table header bottom border position unchanged:

```suggestion
  padding: 3px 12px;
  margin: -3px -10px -3px 0;
```

pontoon/teams/templates/teams/widgets/team_list.html
```diff
@@ -9,7 +9,13 @@
...
         <th class="population">Speakers<i class="fas"></i></th>
         <th class="latest-activity">Latest Activity<i class="fas"></i></th>
-        <th class="progress">Progress<i class="fas"></i></th>
+        <th class="progress controls">
+          Progress
+          <i class="fas"></i>
+          <button class="button" id="detailed-view-toggle" title="Show details">
+            Show details
+          </button>
+        </th>
```
mathjazz: We are now repeating this block of code 3 times. It would be nice if it were factored out (like for example `widgets/latest_activity.html` or `widgets/progress_chart.html`).

dannycolin: I had the same idea but I wasn't sure if it was worth it. Done!

mathjazz (approved):
Well done! \o/

PR #3997, opened 2026-02-26
Sort @-mention results by locale and project activity

### LINKED ISSUE #2548: @user search result in Comment field limited to 5 results only

MundiaNderi:
Fixes #2548

Adds locale and project parameters to the endpoint and use them to annotate each user with their translation activity. Results are then ordered by that activity.

The frontend is updated to pass the locale and project when calling the endpoint
 How to review?
- Check the backend changes in `pontoon/base/views.py`
- Check the frontend changes in `translate/src/api/user.ts` and
  `translate/src/context/MentionUsers.tsx`
- Check the new tests in
  `pontoon/base/tests/views/test_get_users.py`

Testing
- Added tests verifying that users active in the current locale
  appear before users active in other locales
- Verified locally via DevTools Network tab that the request to
  `/get-users/` includes `locale` and `project` query parameters

Steps to reproduce
1. Open the translate view for any locale and project
2. Open the comment box on any string
3. Type `@` followed by a common name
4. Verify the 5 results shown are contributors active in the
   current locale and project

Screenshot 2026-02-26 at 21 14 50

### DIFF
--- a/translate/src/context/MentionUsers.tsx
+++ b/translate/src/context/MentionUsers.tsx
@@ -28,7 +28,7 @@ export function MentionUsersProvider({
 
   const initMentions = useMemo(
     () => (locale: string, projectId: number) => {
-      if (state?.projectId === projectId && state.mentionUsers.length > 0) {
+      if (state?.projectId === projectId) {
         return;
       }
       fetchUsersList(locale, projectId).then((list) => {


pontoon/base/tests/views/test_get_users.py
eemeli, line 1: I'd also like to see a test that shows that project or locale activity is prioritised over alphabetic order, for users that are both active in the project or locale.

translate/src/context/MentionUsers.tsx
```diff
@@ -21,10 +22,11 @@ export function MentionUsersProvider({
 }: {
   children: React.ReactElement;
 }) {
+  const { locale, project } = useContext(Location);
```
eemeli: Is there a reason to rely on the `project` from the Location rather than the Entity? The former could be `'all-projects'`, while the latter is always related to some actual project -- and from there you could get the project id, using which would simplify the database lookup a little.

translate/src/context/MentionUsers.tsx
```diff
@@ -21,10 +22,11 @@ export function MentionUsersProvider({
...
   const [mentionUsers, setUsers] = useState<MentionUser[]>([]);
   const [initMentions, setInit] = useState(() => () => {
     setInit(() => () => {});
-    fetchUsersList().then((list) => {
+    fetchUsersList(locale, project).then((list) => {
       if (Array.isArray(list)) {
         setUsers(list);
```
eemeli: Note that we're effectively memoising the list here, and so if you apply the change above you'll need to change the data structure here to account for the project id potentially changing when we navigate between different entities.

eemeli (changes requested):
Looks pretty promising; see inline for comments. Also, sorry for not getting to this before now.

translate/src/context/MentionUsers.tsx
```diff
@@ -21,18 +21,33 @@ export function MentionUsersProvider({
...
+      });
+    },
+    [],
+  );
+
   const value = useMemo(
-    () => ({ initMentions, mentionUsers }),
-    [initMentions, mentionUsers],
+    () => ({ initMentions, mentionUsers: state.mentionUsers }),
+    [initMentions, state.mentionUsers],
```
eemeli:
```suggestion
    [state],
```

mathjazz: @MundiaNderi I'll let you make this change and then merge. Nice work!

translate/src/context/MentionUsers.tsx
```diff
@@ -21,18 +21,28 @@ export function MentionUsersProvider({
...
-      if (Array.isArray(list)) {
-        setUsers(list);
+  const [state, setState] = useState<{
+    projectId: number;
+    mentionUsers: MentionUser[];
+  } | null>(null);
+
+  const initMentions = useMemo(
+    () => (locale: string, projectId: number) => {
+      if (state?.projectId === projectId && state.mentionUsers.length > 0) {
```
eemeli:
```suggestion
      if (state?.projectId === projectId) {
```

PR #3999, opened 2026-02-28
Use the right docker compose command based on the user setup

dannycolin:
Currently, `/bin/watch.sh` uses `docker-compose` command. The script fails to execute properly if the newer `docker compose` subcommand is installed instead of the legacy `docker-compose` command.

To ensure the script doesn't fail, this patch detects which command is available before passing it to `concurrently`.

### DIFF
--- a/bin/watch.sh
+++ b/bin/watch.sh
@@ -9,5 +9,5 @@ trap sigint_handler INT
 
 npx concurrently -n translate,server,pg -c cyan,magenta,green,red \
   'npm start -w translate' \
-  'docker-compose logs --tail=0 --follow --no-log-prefix server' \
-  'docker-compose logs --tail=0 --follow --no-log-prefix postgresql' \
+  'docker compose logs --tail=0 --follow --no-log-prefix server' \
+  'docker compose logs --tail=0 --follow --no-log-prefix postgresql'


bin/watch.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+# Detect which compose command is installed
+docker compose --help >/dev/null 2>&1 && DOCKER="docker compose"
```
flodolo: I think if/else would be more readable. Also, just running the command without a parameter should be enough.

```bash
if docker compose version >/dev/null 2>&1; then
    DOCKER="docker compose"
elif command -v docker-compose >/dev/null 2>&1; then
    DOCKER="docker-compose"
else
    echo "Error: neither 'docker compose' nor 'docker-compose' is available" >&2
    exit 1
fi
```

bin/watch.sh
```diff
@@ -9,5 +13,5 @@ trap sigint_handler INT
 
 npx concurrently -n translate,server,pg -c cyan,magenta,green,red \
   'npm start -w translate' \
-  'docker-compose logs --tail=0 --follow --no-log-prefix server' \
-  'docker-compose logs --tail=0 --follow --no-log-prefix postgresql' \
+  "sh -c '${DOCKER} logs --tail=0 --follow --no-log-prefix server'" \
```
flodolo:
```suggestion
  "${DOCKER} logs --tail=0 --follow --no-log-prefix server" \
```

Is the `sh -c` necessary? `I would expect `${DOCKER}` to work.

bin/watch.sh
flodolo: On second thoughts, me already made a call in the makefile, maybe we should stick to it and only use `docker compose`.
https://github.com/mozilla/pontoon/commit/31800a5d64319bde39a8d491ffd9c7de3454ce48

dannycolin: It was also my initial thought but I didn't know if there was a reason to use the legacy `docker-compose` command. So if you think it's fine to go with `docker compose`, I can update my patch to reflect that decision.

bin/watch.sh
dannycolin: It is necessary. Without it, it is interperted as two separated strings after Bash expansion ('docker compose' 'logs ...') instead of one which `concurrently` is expecting.

bin/watch.sh
mathjazz: Yeah, let's be consistent and use `docker compose` everywhere.

bin/watch.sh
dannycolin: Not needed with the decision to use `docker compose` only.

PR #4005, opened 2026-03-03
implement security guidelines for github actions

fkiriakos07:


fkiriakos07:
we can wait to apply my changes after the other PR is merged, there is no rush, do you mean pinning to hashes would break dependabot? dependabot can be configured to update github actions, [instructions](https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/secure-your-dependencies/keeping-your-actions-up-to-date-with-dependabot)

PR #4009, opened 2026-03-04
On the profile page only expose actions visible to the viewer

functionzz:
This PR fixes [https://bugzilla.mozilla.org/show_bug.cgi?id=1815203](https://bugzilla.mozilla.org/show_bug.cgi?id=1815203).

It creates a new method on `ActionLogQuerySet` called `.visible_for`, which mirrors the `.visible_for` method found in `ProjectQuerySet`. It prevents regular users from accessing contributions to private projects via the contribution graph/timeline, while maintaining behavior for super users.

### DIFF
--- a/pontoon/contributors/utils.py
+++ b/pontoon/contributors/utils.py
@@ -279,14 +279,14 @@ def get_approvals_charts_data(user):
     }
 
 
-def get_contributions_map(user, contribution_period=None, actions=None):
+def get_contributions_map(contributor, viewer, contribution_period=None):
     """
     Return a map of contribution types and corresponding QuerySets of contributions.
 
     :param django.db.models.Q contribution_period: ActionLog time interval.
     """
-    if actions is None:
-        actions = ActionLog.objects.all()
+
+    actions = ActionLog.objects.visible_for(viewer)
 
     if contribution_period is not None:
         actions = actions.filter(contribution_period)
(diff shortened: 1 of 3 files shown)


pontoon/actionlog/models.py
```diff
@@ -1,8 +1,29 @@
 from django.core.exceptions import ValidationError
 from django.db import models
+from django.db.models import Q
 from django.utils import timezone
 
 
+class ActionLogQuerySet(models.QuerySet):
+    def visible_for(self, user):
+        """
+        The visiblity of actionlogs is determined by the role of the user:
```
flodolo:
```suggestion
        The visibility of actionlogs is determined by the role of the user:
```

flodolo (commented):
(passing by comment)

pontoon/contributors/utils.py
```diff
@@ -279,13 +279,14 @@ def get_approvals_charts_data(user):
     }
 
 
-def get_contributions_map(user, contribution_period=None):
+def get_contributions_map(user, contribution_period=None, actions=None):
```
mathjazz: Why do you pass actions all the way from the view, instead of just the user (you could refer to it as `viewer` and the user for which you are showing the profile page as `contributor`)? And then you would query actions here, like you query them for all the other criteria.

PR #4012, opened 2026-03-05
Add advanced search options to user profile

### LINKED ISSUE #4008: Add advanced search options to user profile
Most of the time I end up including string IDs in searches. 

While I still think that most users don't care about including them, so default to false is better, it would be nice to be able to set a different default in settings.

In my profile, I should be able to decide if I want to include search identifiers and rejected strings, since these 2 expand the pool of results (the others are limiting it).

flodolo:
Fixes #4008

The Python part feels straightforward, the front-end not so much (likely because of my lack of love for JS and React).

One surprise was the amount of places where we hardcoded a default false value for these settings. I tried to define them in one place, so we can change them more easily in the future (well, one for Python and one for JS).

Profile settings are also used on the new /search page.

Screenshot 2026-03-05 alle 06 27 06

### DIFF
--- a/translate/src/modules/search/components/SearchPanel.css
+++ b/translate/src/modules/search/components/SearchPanel.css
@@ -56,6 +56,7 @@
       cursor: pointer;
       font-weight: 300;
       padding: 2px 4px;
+      line-height: 22px;
 
       &:hover {
         color: var(--white-1);


pontoon/contributors/templates/contributors/settings.html
```diff
@@ -358,6 +358,69 @@ <h3>Editor</h3>
...
+            Customize your search experience by choosing which
+            <a
+              href="https://mozilla-l10n.github.io/localizer-documentation/tools/pontoon/translation_workspace.html#search-options"
+              target="_blank"
+              >options</a
+            >
+            to enable by default when searching for strings. It’s always
+            possible to change them in the search bar of the translation
+            workspace to override your default settings for specific searches.
+          </p>
```
mathjazz: I would drop the help text. I don't think it adds much value here, so I'd rather use it in the docs. It's also not used consistenly in all sections, so I'd even remove it from the two that use it.

flodolo: I personally like a bit of explanation (I would add one for the editor settings). It also makes the documentation more discoverable.

pontoon/base/views.py
```diff
@@ -290,6 +291,18 @@ def entities(request):
...
+    # value from the model field (avoid specifying the default in two places).
+    for name in (
+        "search_identifiers",
+        "search_exclude_source_strings",
+        "search_rejected_translations",
+        "search_match_case",
+        "search_match_whole_word",
+    ):
+        if name not in request.POST:
+            form_data[name] = UserProfile._meta.get_field(name).default
```
mathjazz:
```suggestion
            form_data[name] = UserProfile._meta.get_field(name).get_default()
```

See: https://stackoverflow.com/questions/1308332/how-do-i-get-the-default-value-for-a-field-in-a-django-model

pontoon/contributors/templates/contributors/settings.html
mathjazz: Sorry, I missed your comment earlier. It's surprising that only one section links to the docs. (And the docs link is likely to change soon.) Changes like this should be done across the board, in a separate PR.

I would drop the description and change the title Search > Default search options.

BTW, the editor settings should be remove from /settings when Translate Toolkit will be deprecated.

translate/src/modules/search/components/SearchPanel.css
```diff
@@ -55,14 +55,44 @@
       color: var(--light-grey-7);
       cursor: pointer;
       font-weight: 300;
-      padding: 4px 2px;
+      padding: 2px 4px;
```
mathjazz, line 58:
```suggestion
      padding: 2px 4px;
      line-height: 22px;
```

mathjazz: Please also apply this one.

PR #4020, opened 2026-03-14
Cache calls to external API services for one week

### LINKED ISSUE #2664: Cache API requests to external translation services
Caching this type of request would provide performance benefits, and potentially some cost reduction.

In case of the custom Google engine, the answer is not going to change until we re-train the model, so we could even cache for a long time (the downside would be the cache size).

Even for other engines — Microsoft, standard Google Translate — we can assume the answer won't change frequently given a certain input.

flodolo:
Fixes #2664

### DIFF
--- a/pontoon/machinery/utils.py
+++ b/pontoon/machinery/utils.py
@@ -12,6 +12,7 @@ import google.auth
 import google.auth.transport.requests
 import requests
 
+from google.auth.exceptions import DefaultCredentialsError
 from google.cloud import translate
 from google.oauth2 import service_account
 from rapidfuzz.distance.Indel import normalized_distance
@@ -19,6 +20,7 @@ from rapidfuzz.distance.Indel import normalized_distance
 from django.conf import settings
 from django.contrib.postgres.aggregates import ArrayAgg
 from django.core.cache import cache
+from django.core.exceptions import ImproperlyConfigured
 from django.db.models import Q
 
 import pontoon.base as base
@@ -62,7 +64,7 @@ def get_google_generic_translation(text, locale_code, format="text"):
     api_key = settings.GOOGLE_TRANSLATE_API_KEY
 
     if not api_key:
-        raise ValueError("GOOGLE_TRANSLATE_API_KEY not set")
+        raise ImproperlyConfigured("GOOGLE_TRANSLATE_API_KEY not set")
 
     url = "https://translation.googleapis.com/language/translate/v2"
 
@@ -100,12 +102,17 @@ def get_google_automl_translation(
     if cached is not None:
         return cached
 
-    client = translate.TranslationServiceClient()
+    try:
+        client = translate.TranslationServiceClient()
+    except DefaultCredentialsError as e:
+        raise ImproperlyConfigured(
+            f"Google AutoML credentials incorrectly configured: {e}"
+        )
 
     project_id = settings.GOOGLE_AUTOML_PROJECT_ID
 
     if not project_id:
-        raise ValueError("GOOGLE_AUTOML_PROJECT_ID not set")
+        raise ImproperlyConfigured("GOOGLE_AUTOML_PROJECT_ID not set")
 
     # Google AutoML Translation requires location "us-central1"
     location = "us-central1"
@@ -152,7 +159,7 @@ def get_microsoft_translator_data(text, locale_code):
     api_key = settings.MICROSOFT_TRANSLATOR_API_KEY
 
     if not api_key:
-        raise ValueError("MICROSOFT_TRANSLATOR_API_KEY not set")
+        raise ImproperlyConfigured("MICROSOFT_TRANSLATOR_API_KEY not set")
 
     url = "https://api.cognitive.microsofttranslator.com/translate"
     headers = {"Ocp-Apim-Subscription-Key": api_key, "Content-Type": "application/json"}
--- a/pontoon/machinery/views.py
+++ b/pontoon/machinery/views.py
@@ -162,7 +162,7 @@ def gpt_transform(request):
         )
 
     except Exception as e:
-        return JsonResponse({"status": False, "message": str(e)}, status=400)
+        return _machinery_error_response("GPT Transform", e)
 
 
 @login_required(redirect_field_name="", login_url="/403")


pontoon/machinery/utils.py
```diff
@@ -29,6 +31,15 @@
 MAX_RESULTS = 5
 
 
+def get_ext_api_cache_key(service, *parts):
+    digest = hashlib.md5(":".join(str(p) for p in parts).encode()).hexdigest()
+    return f"ext_api:{service}:{digest}"
+
+
+def ext_api_cache_set(key, value):
```
mathjazz: Same as above:

```suggestion
def set_machinery_service_cache_key(key, value):
```

pontoon/machinery/utils.py
```diff
@@ -29,6 +31,15 @@
 MAX_RESULTS = 5
 
 
+def get_ext_api_cache_key(service, *parts):
+    digest = hashlib.md5(":".join(str(p) for p in parts).encode()).hexdigest()
+    return f"ext_api:{service}:{digest}"
```
mathjazz: The naming here takes a bit of an effort to decypher.

Perhaps something like this is clearer?

```suggestion
def get_machinery_service_cache_key(service, *parts):
    digest = hashlib.md5(":".join(str(p) for p in parts).encode()).hexdigest()
    return f"machinery_service:{service}:{digest}"
```

pontoon/machinery/utils.py
```diff
@@ -73,10 +89,12 @@ def get_google_generic_translation(text, locale_code, format="text"):
                 "message": f"Bad Request: {root}",
             }
 
-        return {
+        result = {
             "status": True,
             "translation": root["data"]["translations"][0]["translatedText"],
         }
+        ext_api_cache_set(cache_key, result)
```
mathjazz: Why cache the entire dict, not just the result?

pontoon/settings/base.py
```diff
@@ -788,6 +788,9 @@ def _default_from_email():
 # Default timeout for the per-view cache, in seconds.
 VIEW_CACHE_TIMEOUT = 60 * 60 * 24  # 1 day
 
+# Timeout for external API cache entries, in seconds.
+EXT_API_CACHE_TIMEOUT = 60 * 60 * 24 * 7  # 1 week
```
mathjazz:
```suggestion
# Timeout for external Machiney service cache, in seconds.
MACHINERY_SERVICE_CACHE_TIMEOUT = 60 * 60 * 24 * 7  # 1 week
```

pontoon/machinery/utils.py
```diff
@@ -89,6 +107,18 @@ def get_google_generic_translation(text, locale_code, format="text"):
 def get_google_automl_translation(
     text, locale, format="text", preserve_placeables=False
 ):
+    cache_key = get_ext_api_cache_key(
+        "google_automl",
+        text,
+        locale.google_automl_model,
+        locale.google_translate_code,
```
mathjazz: Should be sufficient? 

```suggestion
        locale.code,
```

pontoon/machinery/utils.py
flodolo: Good point, no benefit.

pontoon/machinery/utils.py
flodolo: Also going to rename the setting in a similar way.

pontoon/machinery/utils.py
flodolo: The diff is a bit larger, because I don't think I can get away without having the same structure in the response, while caching only the result part.

pontoon/machinery/utils.py
```diff
@@ -89,6 +104,18 @@ def get_google_generic_translation(text, locale_code, format="text"):
 def get_google_automl_translation(
     text, locale, format="text", preserve_placeables=False
 ):
+    cache_key = get_machinery_service_cache_key(
+        "google_automl",
+        text,
+        locale.google_automl_model,
```
flodolo: Need to remove this.

pontoon/machinery/views.py
```diff
@@ -126,7 +134,9 @@ def microsoft_translator(request):
                 status=400,
             )
 
-        return JsonResponse({"translation": root[0]["translations"][0]["text"]})
+        result = {"translation": root[0]["translations"][0]["text"]}
+        set_machinery_service_cache_key(cache_key, result)
```
flodolo: Should this also cache the result instead of the dictionary?

pontoon/machinery/utils.py
```diff
@@ -89,6 +104,17 @@ def get_google_generic_translation(text, locale_code, format="text"):
...
+    cache_key = get_machinery_service_cache_key(
+        "google_automl",
+        text,
+        locale.code,
+        format,
+        preserve_placeables,
+    )
+    cached = cache.get(cache_key)
+    if cached is not None:
+        return {"status": True, "translation": cached}
```
flodolo: This doesn't seem particularly readable. Is it worth the space saving?

pontoon/machinery/utils.py
```diff
@@ -42,6 +53,11 @@ def get_google_translate_data(text, locale, format="text", preserve_placeables=F
 
 
 def get_google_generic_translation(text, locale_code, format="text"):
+    cache_key = get_machinery_service_cache_key("google", text, locale_code, format)
```
mathjazz:
```suggestion
    cache_key = get_machinery_service_cache_key("google_generic", text, locale_code, format)
```

pontoon/machinery/views.py
```diff
@@ -102,6 +105,11 @@ def microsoft_translator(request):
             status=400,
         )
 
+    cache_key = get_machinery_service_cache_key("microsoft", text, locale_code)
```
mathjazz:
```suggestion
    cache_key = get_machinery_service_cache_key("microsoft_translator", text, locale_code)
```

pontoon/machinery/openai_service.py
```diff
@@ -2,19 +2,36 @@
...
-        if not OPENAI_API_KEY:
+        if not settings.OPENAI_API_KEY:
             raise ValueError("Missing OpenAI API key")
         self.client = OpenAI()
 
     def get_translation(
         self, english_text, translated_text, characteristic, target_language_name
     ):
+        cache_key = get_machinery_service_cache_key(
+            "openai",
```
mathjazz:
```suggestion
            "openai_chatgpt",
```

pontoon/machinery/utils.py
mathjazz: We should use the same pattern everywhrere. I'll leave it to you whether we optimize for code readability or cache space.

Another argument for caching only the variable bits is that you can change the response structure in the future more easily.

pontoon/machinery/utils.py
```diff
@@ -115,25 +98,14 @@ def get_google_automl_translation(
...
 
-    try:
-        client = translate.TranslationServiceClient()
-    except DefaultCredentialsError as e:
-        log.error(f"Google AutoML Translation error: {e}")
-        return {
-            "status": False,
-            "message": f"{e}",
-        }
+    client = translate.TranslationServiceClient()
```
mathjazz: Why do we no longer cache the error here?

pontoon/machinery/utils.py
```diff
@@ -29,27 +30,39 @@
...
+
     api_key = settings.GOOGLE_TRANSLATE_API_KEY
 
     if not api_key:
-        log.error("GOOGLE_TRANSLATE_API_KEY not set")
-        return {
-            "status": False,
-            "message": "Bad Request: Missing api key.",
-        }
+        raise ValueError("GOOGLE_TRANSLATE_API_KEY not set")
```
mathjazz: Nit: `ImproperlyConfigured` would be more appropriate error type here and in a couple of places below.

pontoon/machinery/views.py
```diff
@@ -171,12 +153,14 @@ def gpt_transform(request):
...
+                "translation": service.get_translation(
+                    english_text, translated_text, characteristic, target_language_name
+                )
+            }
         )
 
-        return JsonResponse({"translation": transformed_text})
-
     except Exception as e:
         return JsonResponse({"status": False, "message": str(e)}, status=400)
```
mathjazz: You might also want to leverage `_machinery_error_response()` here.

pontoon/machinery/utils.py
flodolo: Caching the error = wrapping in try/except?

Before we were catching `DefaultCredentialsError`. This would propagate to `_machinery_error_response()` and still be logged with a generic error.

I'm going to wrap it back and use a clearer error message.

Let me know if that makes sense.

mathjazz: Ugh, I meant "catch". :) Thanks.

PR #4021, opened 2026-03-14
Fix warnings in Python back-end tests

### LINKED ISSUE #3955: Investigate error `UnorderedObjectListWarning` in logs
```
/app/pontoon/teams/views.py:356: UnorderedObjectListWarning: Pagination may yield inconsistent results with an unordered object_list: <class 'pontoon.base.models.translation_memory.TranslationMemoryEntry'> TranslationMemoryEntryQuerySet.
```

There are only a handful per day, so it might be an edge case.

https://github.com/mozilla/pontoon/blob/main/pontoon/teams/views.py#L356C1-L356C56

flodolo:
Noticed them while working on the caching. Also fixes #3955 (didn't realize it was showing up in tests as well).

### DIFF
--- a/pontoon/messaging/tests/test_emails.py
+++ b/pontoon/messaging/tests/test_emails.py
@@ -1,10 +1,9 @@
-from datetime import date, timezone as dt_timezone
+from datetime import date, datetime, timezone
 from unittest.mock import patch
 
 import pytest
 
 from django.test.client import RequestFactory
-from django.utils import timezone
 
 from pontoon.insights.models import LocaleInsightsSnapshot
 from pontoon.messaging.emails import (
@@ -54,9 +53,7 @@ def test_get_monthly_locale_stats_uses_end_of_month_snapshot():
     )
 
     with patch("pontoon.messaging.emails.timezone") as mock_tz:
-        mock_tz.now.return_value = timezone.datetime(
-            2025, 11, 1, 6, 30, 0, tzinfo=dt_timezone.utc
-        )
+        mock_tz.now.return_value = datetime(2025, 11, 1, 6, 30, 0, tzinfo=timezone.utc)
         result = _get_monthly_locale_stats(months_ago=1)
 
     assert locale.pk in result


pontoon/api/views.py
```diff
@@ -382,7 +382,7 @@ class TermSearchListView(RequestFieldsMixin, generics.ListAPIView):
     def get_queryset(self):
         locale = self.request.query_params.get("locale")
 
-        qs = Term.objects.all()
+        qs = Term.objects.order_by("id")
```
flodolo, line 385: Do we want something different?

pytest.ini
```diff
@@ -1,5 +1,9 @@
 [pytest]
-addopts = --doctest-glob='*.tst' --reuse-db
```
flodolo, line 2: We don't have .tst files.

pytest.ini
```diff
@@ -1,5 +1,9 @@
...
+addopts = --reuse-db
 norecursedirs = .git _build commands media node_modules requirements tmp*
 DJANGO_SETTINGS_MODULE = pontoon.settings
+filterwarnings =
+    ignore::DeprecationWarning:notifications
+    ignore::DeprecationWarning:distutils
+    ignore::django.utils.deprecation.RemovedInDjango51Warning:notifications
```
flodolo: Most of the warnings come from `django-notifications-hg`, we should clean up these when we update django in a couple of months.

flodolo: This doesn't seem to work 🤔

flodolo: Couldn't figure out a sensible way to make these go away, so dropping this part.

pontoon/api/views.py
mathjazz: Are we getting warnings without this change?

pontoon/settings/base.py
```diff
@@ -1190,7 +1191,7 @@ def account_username(user):
 badges_start_date = os.environ.get("BADGES_START_DATE", "1970-01-01")
 try:
     BADGES_START_DATE = timezone.make_aware(
-        datetime.strptime(badges_start_date, "%Y-%m-%d"), timezone=timezone.utc
+        datetime.strptime(badges_start_date, "%Y-%m-%d"), timezone=dt.timezone.utc
     )
```
mathjazz: This should be the equivalent code, but simpler and with less import complexity.

```python
    from datetime import datetime, timezone

    BADGES_START_DATE = datetime.strptime(
        badges_start_date,
        "%Y-%m-%d"
    ).replace(tzinfo=timezone.utc)
```

pontoon/api/views.py
flodolo: Yes
https://github.com/mozilla/pontoon/actions/runs/23161736932/job/67291074934

```
pontoon/api/tests/test_views.py::test_terminology_search
  /opt/hostedtoolcache/Python/3.11.15/x64/lib/python3.11/site-packages/rest_framework/pagination.py:207: UnorderedObjectListWarning: Pagination may yield inconsistent results with an unordered object_list: <class 'pontoon.terminology.models.Term'> TermQuerySet.
    paginator = self.django_paginator_class(queryset, page_size)

pontoon/api/tests/test_views.py::test_tm_search
  /opt/hostedtoolcache/Python/3.11.15/x64/lib/python3.11/site-packages/rest_framework/pagination.py:207: UnorderedObjectListWarning: Pagination may yield inconsistent results with an unordered object_list: <class 'pontoon.base.models.translation_memory.TranslationMemoryEntry'> TranslationMemoryEntryQuerySet.
    paginator = self.django_paginator_class(queryset, page_size)
```

PR #4024, opened 2026-03-17
Allow comment authors to edit and delete their own comments

### LINKED ISSUE #2383: ability to edit and remove a comment.
for now we have pontoon.base.views.add_comment view for creating new comments, if the comments contains a typo, commenter needs to create another one.

MundiaNderi:
Fixes #2383 

This PR
Adds edit and delete comment views, corresponding URLs, changes the comment components and adds tests for both views and component changes

To test/how I tested:
- Test the changes locally to see you can edit & delete comments.
- Check if the corresponding tests pass / are sufficient

### DIFF
--- a/pontoon/base/tests/views/test_comment.py
+++ b/pontoon/base/tests/views/test_comment.py
@@ -4,6 +4,7 @@ from django.contrib.auth.models import Permission
 from django.urls import reverse
 
 from pontoon.base.models import Comment
+from pontoon.test.factories import TranslationCommentFactory
 
 
 @pytest.mark.django_db
@@ -118,3 +119,17 @@ def test_delete_comment(member, client, comment_a):
 
     assert response.status_code == 200
     assert not Comment.objects.filter(pk=comment_a.pk).exists()
+
+    # the author can delete their comment
+    member.user.user_permissions.remove(permission)
+    member.user.refresh_from_db()
+
+    own_comment = TranslationCommentFactory(author=member.user)
+
+    response = member.client.post(
+        url,
+        {"comment_id": own_comment.pk},
+        HTTP_X_REQUESTED_WITH="XMLHttpRequest",
+    )
+    assert response.status_code == 200
+    assert not Comment.objects.filter(pk=own_comment.pk).exists()
--- a/translate/src/modules/comments/components/Comment.css
+++ b/translate/src/modules/comments/components/Comment.css
@@ -83,10 +83,14 @@
 .comments-list .comment .info .delete-button {
   background-color: transparent;
   border: none;
-  color: var(--status-error);
+  color: var(--light-grey-7);
   font-size: 11px;
   font-weight: 300;
   padding: 0;
+}
+
+.comments-list .comment .info .delete-button:hover {
+  color: var(--status-error);
 }
 
 .comments-list .comment .info .pin-button:hover {


pontoon/base/views.py
```diff
@@ -751,6 +751,61 @@ def unpin_comment(request):
...
+def edit_comment(request):
+    """Edit a comment"""
+    comment_id = request.POST.get("comment_id", None)
+
+    if not comment_id:
+        return JsonResponse({"status": False, "message": "Bad Request"}, status=400)
+
+    comment = get_object_or_404(Comment, id=comment_id)
+
+    if request.user != comment.author:
```
flodolo, line 766: Should we allow admins to delete and edit comments?

MundiaNderi: Good question , I think allowing users with can_manage_project permission to edit and delete any comment makes sense for moderation purposes. I'll update the permission check to allow both the comment author and project managers to delete comments.

I think admins should be allowed to delete comments, I'm not sure if they should be allowed to edit comments.

translate/public/locale/en-US/translate.ftl
```diff
@@ -101,7 +101,9 @@ comments-Comment--unpin-button = UNPIN
 comments-Comment--pinned = PINNED
 comments-CommentsList--pinned-comments = PINNED COMMENTS
 comments-CommentsList--all-comments = ALL COMMENTS
-
+comments-Comment--edit-button = EDIT
+comments-Comment--delete-button = DELETE
+comments-Comment--cancel-edit = CANCEL EDIT
```
mathjazz: Nit:

```suggestion
comments-Comment--cancel-edit-button = CANCEL EDIT
```

pontoon/base/urls.py
```diff
@@ -83,6 +83,8 @@
         views.unpin_comment,
         name="pontoon.unpin_comment",
     ),
+    path("edit-comment/", views.edit_comment, name="pontoon.edit_comment"),
+    path("delete-comment/", views.delete_comment, name="pontoon.delete_comment"),
```
mathjazz: Nit: it would be nice to Format these two lines same as the lines above, i.e. split each parameter into a new line.

pontoon/base/views.py
mathjazz: Agreed.

mathjazz (commented):
Great work!

The feature works fine and the code looks good on the surface.

My only complaint is that the alignment of the comment editor in Translation History panel and the alignment of buttons underneath it, is off:

Image

BTW, this feature will also need docs update (just added a label to the issue):
https://mozilla-l10n.github.io/localizer-documentation/tools/pontoon/glossary.html#comment

pontoon/base/tests/views/test_comment.py
```diff
@@ -60,3 +62,59 @@ def test_unpin_comment(member, client, team_comment_a):
...
+    # a user cannot delete someone elses comment
+    response = member.client.post(
+        url,
+        {"comment_id": comment_a.pk, "content": "deleted"},
+        HTTP_X_REQUESTED_WITH="XMLHttpRequest",
+    )
+
+    assert response.status_code == 403
+
+    # the author can delete their comment
```
mathjazz: Please also test `base.can_manage_project` permission here.

translate/public/locale/en-US/translate.ftl
```diff
@@ -101,7 +101,9 @@ comments-Comment--unpin-button = UNPIN
 comments-Comment--pinned = PINNED
 comments-CommentsList--pinned-comments = PINNED COMMENTS
 comments-CommentsList--all-comments = ALL COMMENTS
-
+comments-Comment--edit-button = EDIT
+comments-Comment--delete-button = DELETE
+comments-Comment--cancel-edit-button = CANCEL EDIT
```
mathjazz: Actually, do you think just CANCEL would be sufficient here? It would save some space.

```suggestion
comments-Comment--cancel-button = CANCEL
```

translate/src/modules/comments/components/CommentsList.css
```diff
@@ -3,6 +3,11 @@
   padding: 8px 0;
 }
 
+.comments-list .comment.is-editing .info {
+  margin-left: 44px;
+  margin-top: -16px; /* Offset editor height(-2px normal mode gap +  -14px edit mode gap) */
```
mathjazz: Nit:

```suggestion
  margin-top: -16px; /* Offset editor height(-2px normal mode gap + -14px edit mode gap) */
```

pontoon/base/tests/views/test_comment.py
```diff
@@ -60,3 +62,59 @@ def test_unpin_comment(member, client, team_comment_a):
...
+    # a user cannot delete someone elses comment
+    response = member.client.post(
+        url,
+        {"comment_id": comment_a.pk, "content": "deleted"},
+        HTTP_X_REQUESTED_WITH="XMLHttpRequest",
+    )
+
+    assert response.status_code == 403
+
+    # A user with can_manage_project permission can delete someone else's comment
```
mathjazz, line 109: We should leave the code to test that the author can delete their own comments in.

translate/src/modules/comments/components/Comment.css
```diff
@@ -73,6 +74,21 @@
...
+.comments-list .comment .info .delete-button:before {
+  content: '•';
+  padding: 0 3px 0 3px;
+  color: var(--light-grey-7);
+}
+
+.comments-list .comment .info .delete-button {
+  background-color: transparent;
+  border: none;
+  color: var(--status-error);
```
mathjazz: We should only use this color on hover.

PR #4037, opened 2026-03-24
Consider missing Android placeholders as errors if translation has extra `%` signs

### LINKED ISSUE #4035: Android linter failed to catch variable mispelled as %1 (instead of %1$s)
https://pontoon.mozilla.org/om/firefox-for-android/mozilla-mobile/android-components/components/feature/downloads/src/main/res/values/strings.xml/?string=316489

Fixed in https://github.com/mozilla-l10n/android-l10n/commit/51d55f15c06583f7a9f6f11da6e36ff275088c90

The English string `Download file again? (%1$s)` was translated as `Irradeebiin faayila buufadhaa? (%1)`.

CC @eemeli 
I don't think this is fixed by https://github.com/mozilla/moz-l10n/pull/128?

If that's the case, we clearly need something more reliable (this was caught by the [external linter](https://github.com/mozilla-l10n/mozl10n-linter/actions/runs/23417592018)).

eemeli:
Fixes #4035
See also [bug 2025733](https://bugzilla.mozilla.org/show_bug.cgi?id=2025733)

This should cover the cases where a placeholder is mistyped so that it doesn't parse as an Android placeholder, but still at least starts with a `%` character.

### DIFF
--- a/pontoon/checks/libraries/custom.py
+++ b/pontoon/checks/libraries/custom.py
@@ -72,17 +72,17 @@ def run_custom_checks(entity: Entity, string: str) -> dict[str, list[str]]:
                 errors.append(f"Parse error: {e}")
 
             orig_pct_count = 0
-            orig_ps: set[str] = set()
+            orig_ph_strings: set[str] = set()
             try:
                 orig_msg = mf2_parse_message(entity.string)
                 for pattern in get_patterns(orig_msg):
-                    ppc = 0
+                    pattern_pct_count = 0
                     for el in pattern:
                         if isinstance(el, str):
-                            ppc += el.count("%")
+                            pattern_pct_count += el.count("%")
                         else:
-                            orig_ps.add(preview_placeholder(el))
-                    orig_pct_count = max(orig_pct_count, ppc)
+                            orig_ph_strings.add(preview_placeholder(el))
+                    orig_pct_count = max(orig_pct_count, pattern_pct_count)
             except ValueError:
                 orig_msg = None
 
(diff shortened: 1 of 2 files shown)


pontoon/checks/libraries/custom.py
```diff
@@ -70,18 +70,21 @@ def run_custom_checks(entity: Entity, string: str) -> dict[str, list[str]]:
                 msg = None
                 patterns = ()
                 errors.append(f"Parse error: {e}")
+
+            orig_pct_count = 0
+            orig_ps: set[str] = set()
```
flodolo: Could we use a clearer variable name? I assume it stands for `orig_placeholders`?

pontoon/checks/libraries/custom.py
```diff
@@ -70,18 +70,21 @@ def run_custom_checks(entity: Entity, string: str) -> dict[str, list[str]]:
...
                 orig_msg = mf2_parse_message(entity.string)
-                orig_ph_iter = (
-                    el
-                    for pattern in get_patterns(orig_msg)
-                    for el in pattern
-                    if not isinstance(el, str)
-                )
-                orig_ps = {preview_placeholder(ph) for ph in orig_ph_iter}
+                for pattern in get_patterns(orig_msg):
+                    ppc = 0
```
flodolo: Same here, `placeholder_pct_count` maybe?

pontoon/checks/tests/test_custom.py
flodolo, line 1: Maybe we should have a test with plurals? I have no clue if the syntax is correct.

```python
def test_android_plural_placeholders():
    # Source has placeholder only in 'other'
    original = (
        ".input {$n :number} .match $n"
        " one {{One item}}"
        " * {{{$arg1 :number @source=|%1$d|} items}}"
    )
    # Translation has placeholder in all variants (one, many, other)
    translation = (
        ".input {$n :number} .match $n"
        " one {{{$arg1 :number @source=|%1$d|} item}}"
        " many {{{$arg1 :number @source=|%1$d|} items}}"
        " * {{{$arg1 :number @source=|%1$d|} items}}"
    )
    entity = mock_entity("android", string=original)
    assert run_custom_checks(entity, translation) == {}
```

PR #4040, opened 2026-03-26
Use additional context in LLM prompt, update OpenAI dep

### LINKED ISSUE #4030: Improve LLM prompt by providing clearer instructions and additional context
I'm currently experimenting with a variant of the existing prompt. 

Besides a general rewrite, for example to try and solve a problem with missing closing periods, I think there's value in passing the comment and string ID to the LLM to provide additional context.

Other things that might be worth investigating in the future, but not as part of this iteration:
* Test gpt5-mini instead of gpt4.1, across a few languages.
* Store more context, e.g. some info about the project (additional field in the model), and some structured info about files. That could be interesting for Firefox, where specific files (e.g. asrouter.ftl) are more marketing oriented, and might benefit from a looser translation.

flodolo:
Fixes #4030

- Update OpenAI package to latest version (2.29.0)
- Improve prompt formulation
- Pass string ID, comment, terminology matches when available
- Move OpenAI GPT version to settings

### DIFF
--- a/pontoon/machinery/openai_service.py
+++ b/pontoon/machinery/openai_service.py
@@ -23,7 +23,7 @@ class OpenAIService:
         translated_text,
         characteristic,
         locale,
-        entity_id=None,
+        entity_key=None,
         entity_comment=None,
         group_comment=None,
         resource_comment=None,
@@ -42,7 +42,7 @@ class OpenAIService:
             translated_text,
             characteristic,
             locale.code,
-            entity_id or "",
+            entity_key or "",
             entity_comment or "",
             group_comment or "",
             resource_comment or "",
@@ -66,8 +66,8 @@ class OpenAIService:
         # Separate the instruction from the data.
         # It makes it hard for injected text to masquerade as instructions.
         context_parts = []
-        if entity_id:
-            context_parts.append(f"STRING ID:\n{entity_id}")
+        if entity_key:
+            context_parts.append(f"STRING ID:\n{entity_key}")
         if resource_comment:
             context_parts.append(f"RESOURCE COMMENT:\n{resource_comment}")
         if group_comment:
@@ -107,7 +107,7 @@ class OpenAIService:
         )
 
         context_instructions = []
-        if entity_id:
+        if entity_key:
             context_instructions.append(
                 "STRING ID: use it to infer the UI context (e.g., button, menu item, page title, tooltip) and adapt length and phrasing accordingly."
             )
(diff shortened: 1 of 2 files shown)


translate/src/modules/machinery/getEntityStringId.ts
flodolo, line 1: There might be a better way to do this?

pontoon/machinery/openai_service.py
```diff
@@ -39,70 +61,122 @@ def get_translation(
...
+        system_message = system_header + context_block + system_rules
 
-        # Separate the instruction from the data.
-        # It makes it hard for injected text to masquerade as instructions.
-        user_prompt = (
-            f"{intro_text}\n\n"
-            f"ENGLISH SOURCE:\n{english_text}\n\n"
-            f"MACHINE TRANSLATION TO REFINE:\n{translated_text}"
-        )
+        # TODO: remove before merge.
```
flodolo: This is here to help with testing, needs to be removed before merge.

pontoon/machinery/management/commands/refine_translation.py
```diff
@@ -4,7 +4,7 @@
 
 
 class Command(BaseCommand):
-    help = "Refines machine translations using OpenAI's GPT-4 API with specified characteristics"
+    help = "Refines machine translations using OpenAI's GPT API with specified characteristics"
```
flodolo, line 7: At some point we're going to move away from GPT-4, not sure there is a benefit in having references to GTP-4 in the code.

pontoon/machinery/openai_service.py
```diff
@@ -19,14 +19,36 @@ def __init__(self):
...
 
     def get_translation(
-        self, english_text, translated_text, characteristic, target_language_name
+        self,
+        english_text,
+        translated_text,
+        characteristic,
+        target_language_name,
+        string_id=None,
+        string_comment=None,
```
mathjazz:
```suggestion
        entity_key=None,
        entity_comment=None,
```

pontoon/machinery/openai_service.py
```diff
@@ -19,90 +18,157 @@ def __init__(self):
         self.client = OpenAI()
 
     def get_translation(
-        self, english_text, translated_text, characteristic, target_language_name
+        self,
+        english_text,
+        translated_text,
+        characteristic,
+        locale,
+        entity_id=None,
```
mathjazz: Why not `entity_key`?

flodolo: No reason, in my head it's always ID 🤷🏼 

I'll update and remove the debug print.

PR #4055, opened 2026-04-01
Improve behavior of translation search page with browser's history

### LINKED ISSUE #3942: Implement search history within translation search page
Upon finishing the translation search page in #3799, we should make it easier for users by allowing them to seamlessly navigate their search history.

functionzz:
This patch allows users of [https://pontoon.mozilla.org/search/](https://pontoon.mozilla.org/search/) to be able to click and the "back" and "forward" buttons on their browser to access historical searches, which is inline with expected search engine behaviour.

Fixes #3942.

### DIFF
--- a/pontoon/search/static/js/search.js
+++ b/pontoon/search/static/js/search.js
@@ -144,22 +144,7 @@ $(function () {
   });
 
   $(window).on('popstate', function () {
-    url.href = window.location.href;
-    const search = url.searchParams.get('search');
-    currentPage = 1;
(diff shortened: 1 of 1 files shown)


mathjazz (commented):
Thanks for the patch!

If I search and click the Back button, I get back to the initial search page state, except:
1. Dropdowns are not reset to their initial state.
1. Checkboxes are not reset to their initial state.
1. The table header remains visible.

PR #4070, opened 2026-04-08
Deduplicate XCode translations

### LINKED ISSUE #3904: Strings rejected without visible difference after import (iOS, Android)
This happened for Android a while ago
https://pontoon.mozilla.org/it/firefox-for-android/all-resources/?list=327892,327894,327896,327886,327882,327912,327923&string=327923

But it happened again for iOS
https://pontoon.mozilla.org/it/firefox-for-ios/all-resources/?list=328415,328418&string=328415

Clearly there is something wrong happening with the MF2 vs native representation and strings with variables.

eemeli:
Fixes #3904


mathjazz (approved):
The code looks good. Please test on DEV before we push to PROD.

PR #4072, opened 2026-04-08
Search: compact entity view

### LINKED ISSUE #4069: Provide a more compact "All translations" view, link to API
I find myself not sharing the Pontoon's link and still using Transvision because the latter has a much more compact view. The goal of this view is to be able to quickly scan all translations, and that's not possible with so much space and scrolling.

Example: https://pontoon.mozilla.org/entities/302772/ vs [Transvision](https://transvision.flod.org/string/?entity=android_l10n/mozilla-mobile/android-components/components/browser/menu/src/main/res/values/strings.xml:mozac_browser_menu_extensions_content_description&repo=android_l10n).

I see two options: make it more compact by default, or have a "Compact view" button (passed also via GET parameter). I'd be fine with the latter.

Another thing that I'm missing is having a quick link to the corresponding API.

functionzz:
This PR compacts the new-ish Entity view with small css style changes.

Fixes #4069.

### DIFF
--- a/pontoon/search/static/css/search.css
+++ b/pontoon/search/static/css/search.css
@@ -164,7 +164,7 @@
   color: var(--translation-secondary-color);
   font-style: italic;
   line-height: 22px;
-  margin-top: 20px;
+  margin-top: 8px;
 }
 
 .entity-info-container a,
@@ -189,7 +189,6 @@
   align-items: flex-end;
   gap: 8px;
   margin-bottom: 0;
-  margin-top: 20px;
   font-size: 13px;
 }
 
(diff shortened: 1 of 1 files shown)


functionzz:
@flodolo 
Let me know if these changes fix the issue.

flodolo (changes requested):
It doesn't. Reducing padding is not enough, we should have all data in one row. It also doesn't address the API link part.

The issue is marked as "not ready". Aren't we supposed to work only on issues marked as "ready"? I think this requires a conversation on the design, not diving into code.

functionzz:
If that's the case, I'll close this for further discussion.

PR #4078, opened 2026-04-10
Translate: link project names in concordance search to corresponding views

### LINKED ISSUE #2276: Link project names in Concordance search results to translate view
## [Bug 1697651](https://bugzilla.mozilla.org/show_bug.cgi?id=1697651)
Bug Reporter: @mathjazz
CC:   @gaby2300

*Filing on behalf of :gaby2300:*

*Using Concordance Search and typing "Extensions" I get a lot of results in Firefox and Android L10n, translated some times as "Extensiones" other times as "Complementos". I'd like to change all "Extensions" for "Complementos". It would be nice if clicking FIREFOX or ANDROID L10N at the top would browse to the string where I could then change the translation.*

Note that the feature depends on either:
- the ability to perform "exact search" on source strings, or
- the ability to filter by string IDs - #2049

functionzz:
This PR allows localizers to select certain translation memories through the Concordance Search tab via the project names list and change them via redirect. Since a project can have multiple entities with the same translation, clicking on the project name redirects the user to `/code/project/all-resources/?list=entity_1,entity_2,entity_3...&string=entity_1`.

Fixes #2276.

NOTE: currently private, disabled and system projects are excluded from the Concordance Search results, which is NOT the original behaviour of Concordance Search. A quick discussion is needed on whether to include said results and whether they count as a "security breach". For example currently, projects like ENGAGEMENT and MDN are included.

### DIFF
--- a/pontoon/machinery/utils.py
+++ b/pontoon/machinery/utils.py
@@ -21,12 +21,13 @@ from django.conf import settings
 from django.contrib.postgres.aggregates import JSONBAgg
 from django.core.cache import cache
 from django.core.exceptions import ImproperlyConfigured
-from django.db.models import Q
+from django.db.models import BooleanField, Case, Q, When
 from django.db.models.functions import JSONObject
 
 import pontoon.base as base
 
 from pontoon.base.models.project import Project
+from pontoon.base.models.project_locale import ProjectLocale
 from pontoon.base.models.translation_memory import TranslationMemoryEntry
 from pontoon.base.placeables import get_placeables
 
@@ -280,6 +281,9 @@ def get_concordance_search_data(user, text, locale):
     search_query = reduce(operator.and_, search_filters)
 
     projects = Project.objects.visible_for(user)
+    pl_project_ids = ProjectLocale.objects.filter(locale=locale).values_list(
+        "project_id", flat=True
+    )
 
     search_results = (
         TranslationMemoryEntry.objects.filter(search_query, project__in=projects)
@@ -290,6 +294,11 @@ def get_concordance_search_data(user, text, locale):
                     project_name="project__name",
                     project_slug="project__slug",
                     project_disabled="project__disabled",
+                    project_locale_exists=Case(
+                        When(project__id__in=pl_project_ids, then=True),
+                        default=False,
+                        output_field=BooleanField(),
+                    ),
                     entity="entity",
                 ),
                 distinct=True,
@@ -305,6 +314,7 @@ def get_concordance_search_data(user, text, locale):
                 entry["project_name"],
                 entry["project_slug"],
                 entry["project_disabled"],
+                entry["project_locale_exists"],
             )
             if entry["project_slug"] and entry["entity"]:
                 grouped[key].append(entry["entity"])
@@ -316,6 +326,7 @@ def get_concordance_search_data(user, text, locale):
                 "project_name": k[0],
                 "project_slug": k[1],
                 "project_disabled": k[2],
+                "project_locale_exists": k[3],
                 "entities": entities,
             }
             for k, entities in grouped.items()
--- a/pontoon/machinery/utils.py
+++ b/pontoon/machinery/utils.py
@@ -18,13 +18,15 @@ from google.oauth2 import service_account
 from rapidfuzz.distance.Indel import normalized_distance
 
 from django.conf import settings
-from django.contrib.postgres.aggregates import ArrayAgg
+from django.contrib.postgres.aggregates import JSONBAgg
 from django.core.cache import cache
 from django.core.exceptions import ImproperlyConfigured
 from django.db.models import Q
+from django.db.models.functions import JSONObject
 
 import pontoon.base as base
 
+from pontoon.base.models.translation_memory import TranslationMemoryEntry
 from pontoon.base.placeables import get_placeables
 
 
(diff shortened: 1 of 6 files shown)


pontoon/machinery/utils.py
```diff
@@ -264,7 +267,7 @@ def store_new_terms(url, headers, new_terms):
             log.error(f"Adding new glossary terms failed: {r.content}")
 
 
-def get_concordance_search_data(text, locale):
+def get_concordance_search_data(request, text, locale):
```
mathjazz: I'd avoid passing `request`, otherwise this function looks like it should be a Django view:

```suggestion
def get_concordance_search_data(user, text, locale):
```

pontoon/machinery/utils.py
```diff
@@ -276,13 +279,49 @@ def get_concordance_search_data(text, locale):
     )
     search_query = reduce(operator.and_, search_filters)
 
+    projects = Project.objects.visible_for(request.user)
+
     search_results = (
-        base.models.TranslationMemoryEntry.objects.filter(search_query)
+        TranslationMemoryEntry.objects.filter(search_query, project__in=projects)
```
mathjazz, line 289: There's another place where we use `base.models.TranslationMemoryEntry`, and we should just use `TranslationMemoryEntry` if it's no longer causing a circular import.

Please also check if we can import `base.models.Locale` and `base.utils.get_search_phrases` directly.

pontoon/machinery/utils.py
```diff
@@ -276,13 +279,49 @@ def get_concordance_search_data(text, locale):
...
+        # TranslationMemoryEntry.objects.filter(search_query)
```
mathjazz: Should this line be removed?

pontoon/machinery/utils.py
```diff
@@ -276,13 +279,49 @@ def get_concordance_search_data(text, locale):
...
+            if entry["project_slug"] and entry["entity"]:
+                grouped[key].append(entry["entity"])
+            elif entry["project_slug"]:
+                grouped.setdefault(key, [])
+
+        result["tmEntries"] = [
+            {
+                "projectName": k[0],
+                "projectSlug": k[1],
+                "projectDisabled": k[2],
```
mathjazz: We use the snake_case in Python.

```suggestion
        result["tm_entries"] = [
            {
                "project_name": k[0],
                "project_slug": k[1],
                "project_disabled": k[2],
```

translate/src/modules/machinery/components/ConcordanceSearch.tsx
```diff
@@ -11,22 +11,41 @@ type Props = {
...
-          </span>
-        </li>
-      ))}
+      {tmEntries.map((tmEntry) =>
+        tmEntry.projectSlug ? (
+          <li key={tmEntry.projectName}>
+            {tmEntry.entities.length > 0 && !tmEntry.projectDisabled ? (
+              <a
+                className='translation-source'
+                href={`/${code}/${tmEntry.projectSlug}/all-resources/?list=${tmEntry.entities.join(',')}&string=${tmEntry.entities[0]}`}
```
mathjazz: Why do we need to set the `?string=` parameter here?

pontoon/machinery/views.py
```diff
@@ -77,22 +77,15 @@ def concordance_search(request):
...
 
     try:
         data = paginator.page(page)
     except EmptyPage:
         return JsonResponse({"results": [], "has_next": False})
 
-    # ArrayAgg (used in get_concordance_search_data()) does not support using
-    # distinct=True in combination with ordering, so we need to do one of them
-    # manually - after pagination, to reduce the number of rows processed.
-    projects = Project.objects.order_by("disabled", "-priority", "name").values_list(
```
mathjazz, line 90: We should preserve this order or projects.

mathjazz (commented):
Thanks for the patch, nice work!

Please see my comments in line.

Could you please use `cursor:default` on project names that aren't linked?

mathjazz (commented):
Thanks for the patch, it works as expected. Nicely done!

The code looks good on the surface. Before I dive a bit deeper, here's an idea I got while testing on the DEV server:

What if we made the entire project list link to /LOCALE/all-projects/all-resources/?list=X,Y,Z rather than specific projects? That way the user will be able to access entities of all projects, not just the ones that are visible (and not cut-off).

Thoughts?

Also, on Concordance search, we also show results from Google Translate. That should not be the case.

functionzz:
> Also, on Concordance search, we also show results from Google Translate. That should not be the case.

Are you saying that Google Translate is not supposed to be there, or that we should remove it as a feature in Concordance Search? Because it currently exists in prod

mathjazz:
> Are you saying that Google Translate is not supposed to be there, or that we should remove it as a feature in Concordance Search? Because it currently exists in prod

Yeah, we should drop it. Didn't realize we had that "feature".

translate/src/modules/machinery/components/MachineryTranslation.css
```diff
@@ -39,9 +39,14 @@
   pointer-events: auto;
 }
 
-.machinery .translation > header > ul > li:not(:first-child)::before {
+.sources.projects > li:not(:first-child)::before {
   content: '•';
   padding-right: 3px;
+  padding-left: 3px;
```
mathjazz: What is the purpose of this change? It seems to break the middot alignment: 

Image

pontoon/machinery/utils.py
```diff
@@ -276,11 +280,32 @@ def get_concordance_search_data(text, locale):
...
-        base.models.TranslationMemoryEntry.objects.filter(search_query)
+        TranslationMemoryEntry.objects.filter(search_query, project__in=projects)
         .values("source", "target")
-        .annotate(project_names=ArrayAgg("project__name", distinct=True))
-        .distinct()
+        .annotate(
+            projects=JSONBAgg(
+                JSONObject(
+                    project_name="project__name",
+                    project_slug="project__slug",
```
mathjazz: Why use the prefix `project_` here? Also in the rest of the code, including frontend.

```suggestion
                    name="project__name",
                    slug="project__slug",
```

mathjazz (commented):
Works very nicely, great job!

pontoon/machinery/utils.py
functionzz: I believe I did this due to previous changes (which have been reverted) that included fields from other models, making model differentiation necessary - I'll change these to just `name` and `slug`.

translate/src/modules/machinery/components/MachineryTranslation.css
functionzz: Could've sworn this _aligned_ it for me, but now its unaligned... will revert

pontoon/machinery/views.py
```diff
@@ -17,7 +17,8 @@
 from django.utils.html import strip_tags
 from django.views.decorators.http import require_POST
 
-from pontoon.base.models import Comment, Entity, Locale, Project, Translation
+from pontoon.base.models import Comment, Entity, Locale, Translation
+from pontoon.base.models.project import Project
```
mathjazz: What's the purpose of this change?

pontoon/machinery/utils.py
```diff
@@ -18,13 +18,17 @@
...
 from django.core.exceptions import ImproperlyConfigured
 from django.db.models import Q
+from django.db.models.functions import JSONObject
 
 import pontoon.base as base
 
+from pontoon.base.models.project import Project
+from pontoon.base.models.project_locale import ProjectLocale
+from pontoon.base.models.translation_memory import TranslationMemoryEntry
 from pontoon.base.placeables import get_placeables
```
mathjazz: Can we simplify this?

```suggestion
from pontoon.base.models import Locale, Project, ProjectLocale, TranslationMemoryEntry
from pontoon.base.placeables import get_placeables
from pontoon.base.utils import get_search_phrases
```

mathjazz (commented):
Just a coupe of suggestions to simplify imports, other than that this is good to go!

mathjazz (approved):
Great job!

PR #4086, opened 2026-04-15
Added skeleton loader on slow machinery requests

### LINKED ISSUE #4074: Loading state of Google Translate request should be clearer
Even with caching and warm-up, I often find the Google Translate API unresponsive, and need to wait 15–20 seconds to get a response.

That is not evident from the UI. Would it make more sense to display something, e.g. "Loading machine translation results…"?

wassafshahzad:
## Description
Added a skeleton loader and fetch state to render skeleton loader on slow machinery requests.

## Linked Issue
Closes #4074 

## Screenshot
skeleton structure

### Skeleton Machinery Loader
SkeletonStructure2

### Skeleton Machinery Loader Updated UI
MachinerySkeletonLoader3

### DIFF
--- a/translate/src/modules/loaders/components/MachinerySkeletonLoader.tsx
+++ b/translate/src/modules/loaders/components/MachinerySkeletonLoader.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import '~/modules/loaders/components/SkeletonLoader.css';
+import './MachinerySkeletonLoader.css';
 
 export function MachinerySkeletonLoader({
   items,
@@ -15,7 +15,7 @@ export function MachinerySkeletonLoader({
 
   return (
     <ul
-      className='skeleton-loader machinery-skeleton entities'
+      className={`machinery-skeleton-loader${firstLoad ? '' : ' scroll'}`}
       data-testid='machinery-skeleton-loader'
       ref={sentryRef}
     >
--- a/translate/src/modules/loaders/components/MachinerySkeletonLoader.tsx
+++ b/translate/src/modules/loaders/components/MachinerySkeletonLoader.tsx
@@ -3,8 +3,8 @@ import React from 'react';
 import './MachinerySkeletonLoader.css';
 
 export function MachinerySkeletonLoader({
-  items,
   sentryRef,
+  items,
 }: {
   items: unknown[];
   sentryRef?: React.Ref<HTMLUListElement>;
@@ -15,7 +15,7 @@ export function MachinerySkeletonLoader({
 
   return (
     <ul
-      className={`machinery-skeleton-loader${firstLoad ? '' : ' scroll'}`}
+      className={`machinery-skeleton-loader ${firstLoad ? null : 'scroll'}`}
       data-testid='machinery-skeleton-loader'
       ref={sentryRef}
     >
--- a/translate/src/modules/loaders/components/MachinerySkeletonLoader.tsx
+++ b/translate/src/modules/loaders/components/MachinerySkeletonLoader.tsx
@@ -6,8 +6,8 @@ export function MachinerySkeletonLoader({
   sentryRef,
   items,
 }: {
-  items: unknown[];
   sentryRef?: React.Ref<HTMLUListElement>;
+  items: unknown[];
 }) {
   const firstLoad = items.length === 0;
   const itemCount = firstLoad ? 30 : 2;
--- a/translate/src/modules/machinery/components/Machinery.tsx
+++ b/translate/src/modules/machinery/components/Machinery.tsx
@@ -81,9 +81,7 @@ export function Machinery(): React.ReactElement<'section'> {
             />
           ))}
         </ul>
-        {machineryFetching && translations.length === 0 && (
-          <MachinerySkeletonLoader items={[]} />
-        )}
+        {machineryFetching && <MachinerySkeletonLoader items={[]} />}
         <ul>
           {results.map((result, index) => (
             <MachineryTranslationComponent
--- a/translate/src/modules/machinery/components/Machinery.tsx
+++ b/translate/src/modules/machinery/components/Machinery.tsx
@@ -4,7 +4,7 @@ import useInfiniteScroll from 'react-infinite-scroll-hook';
 import { MachineryTranslations } from '~/context/MachineryTranslations';
 import { SearchData } from '~/context/SearchData';
 
-import { MachinerySkeletonLoader } from './MachinerySkeletonLoader';
+import { MachinerySkeletonLoader } from '../../loaders/components/MachinerySkeletonLoader';
 
 import './Machinery.css';
 import { MachineryTranslationComponent } from './MachineryTranslation';
--- a/translate/src/modules/machinery/components/Machinery.tsx
+++ b/translate/src/modules/machinery/components/Machinery.tsx
@@ -4,7 +4,7 @@ import useInfiniteScroll from 'react-infinite-scroll-hook';
 import { MachineryTranslations } from '~/context/MachineryTranslations';
 import { SearchData } from '~/context/SearchData';
 
-import { SkeletonLoader } from '~/modules/loaders';
+import { MachinerySkeletonLoader } from './MachinerySkeletonLoader';
 
 import './Machinery.css';
 import { MachineryTranslationComponent } from './MachineryTranslation';
@@ -83,7 +83,7 @@ export function Machinery(): React.ReactElement<'section'> {
           ))}
         </ul>
         {machineryFetching && translations.length === 0 && (
-          <SkeletonLoader items={[]} />
+          <MachinerySkeletonLoader items={[]} />
         )}
         <ul>
           {results.map((result, index) => (
@@ -96,7 +96,7 @@ export function Machinery(): React.ReactElement<'section'> {
           ))}
         </ul>
         {(fetching || hasMore) && (
-          <SkeletonLoader items={results} sentryRef={sentryRef} />
+          <MachinerySkeletonLoader items={results} sentryRef={sentryRef} />
         )}
       </div>
     </section>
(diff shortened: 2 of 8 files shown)


wassafshahzad:
The missing lines in test, are in MachineryTranslations.tsx around the async fetch logic. That code was already there before my changes, I just wired fetching state into it.

Testing it properly would need mocking 5 API calls, Redux, and multiple React contexts which felt out of scope for a loading indicator fix. Added two tests in Machinery.test.jsx covering the actual UI change instead.

Happy to add provider tests in a follow-up if needed.

translate/src/modules/machinery/components/Machinery.tsx
```diff
@@ -78,6 +82,9 @@ export function Machinery(): React.ReactElement<'section'> {
             />
           ))}
         </ul>
+        {machineryFetching && translations.length === 0 && (
+          <SkeletonLoader items={[]} />
```
mathjazz: This is a good opportunity to introduce a new skeleton loader, that matches the layout of Machinery results. The one we use in this file is actually imitating the string list results.

mathjazz (commented):
Thanks for the patch!

Please see my comment below for the SkeletonLoader layout.

translate/src/modules/machinery/components/Machinery.tsx
wassafshahzad: Is this UI fine ?
SkeletonStructure2

wassafshahzad: Done

mathjazz: Thanks for the update! Could you please separate the lines for source and target text, and make the full-width and same height as they were in the `SkeletonLoader`?

wassafshahzad: Sure will do

wassafshahzad: @mathjazz Updated loader UI

translate/src/modules/machinery/components/MachinerySkeletonLoader.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+
+import '~/modules/loaders/components/SkeletonLoader.css';
```
mathjazz: This file should be moved to translate/src/modules/loaders/components and have it's own MachinerySkeletonLoader.tsx file.

mathjazz (commented):
Nice work - it works really well!

Please see my comment inline for improved code maintainability.

translate/src/modules/machinery/components/MachinerySkeletonLoader.tsx
wassafshahzad: I am reusing SkeletonLoader.css for the machinery skeleton too and added the machinery styles there, that way both loaders stay consistent. I've also moved the tsx to the right location. Let me know if you want me to extract the machinery styles to a separate css file!

mathjazz: Sorry for the confusion - I mean this file ([translate/src/modules/machinery/components/MachinerySkeletonLoader.tsx](https://github.com/mozilla/pontoon/pull/4086/files/9cc014bc446dbe0e44c7b4936de6e2a85bca9c50#diff-168e69aac3b5dc15bebf06055dfaeb3956887cab4640bd757ba02e8c1c053aea)) should be moved.

I'd prefer each component to have separate CSS styles for consistent code pattern with the rest of the codebase.

wassafshahzad: I have completed the following
- Moved MachinerySkeletonLoader.tsx to the correct path
- Extracted shared + machinery skeleton styles into a separate MachinerySkeletonLoader.css
- This does cause some duplication with SkeletonLoader.css
I have these changes locally, let me know if this approach works and I'll push!

mathjazz: Yeah, that sounds good. You are welcome to push the changes!

wassafshahzad: Just pushed my changes, I will keep you feedback in mind for my next contributions

translate/src/modules/loaders/components/MachinerySkeletonLoader.tsx
```diff
@@ -0,0 +1,33 @@
...
+  items: unknown[];
+  sentryRef?: React.Ref<HTMLUListElement>;
+}) {
+  const firstLoad = items.length === 0;
+  const itemCount = firstLoad ? 30 : 2;
+  const list = Array.from(Array(itemCount).keys());
+
+  return (
+    <ul
+      className={`machinery-skeleton-loader${firstLoad ? '' : ' scroll'}`}
```
mathjazz: I prefer the syntax used here:
https://github.com/wassafshahzad/pontoon/blob/4203cd7deef3c02953cfc0f0399dd58fb3e06e13/translate/src/modules/loaders/components/SkeletonLoader.tsx#L19

translate/src/modules/loaders/components/MachinerySkeletonLoader.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+
+import './MachinerySkeletonLoader.css';
+
+export function MachinerySkeletonLoader({
+  items,
+  sentryRef,
```
mathjazz, line 6: Nit: Is there a reason why the order of these arguments (also below) is flipped compared to the `SkeletonLoader` component?

translate/src/modules/loaders/components/MachinerySkeletonLoader.tsx
mathjazz: I prefer the syntax used here:
https://github.com/wassafshahzad/pontoon/blob/4203cd7deef3c02953cfc0f0399dd58fb3e06e13/translate/src/modules/loaders/components/SkeletonLoader.tsx#L19

translate/src/modules/machinery/components/Machinery.tsx
```diff
@@ -4,7 +4,7 @@ import useInfiniteScroll from 'react-infinite-scroll-hook';
 import { MachineryTranslations } from '~/context/MachineryTranslations';
 import { SearchData } from '~/context/SearchData';
 
-import { SkeletonLoader } from '~/modules/loaders';
+import { MachinerySkeletonLoader } from '../../loaders/components/MachinerySkeletonLoader';
```
mathjazz: Why not this?
```suggestion
import { MachinerySkeletonLoader } from '~/modules/loaders';
```

mathjazz (commented):
Thanks for the update! A few more suggestions.

translate/src/modules/loaders/components/MachinerySkeletonLoader.tsx
wassafshahzad: Implemented

translate/src/modules/loaders/components/MachinerySkeletonLoader.tsx
wassafshahzad: Yes, they have a increasing char count. I have fixed it as per your requests. It is consistant with SkeletonLoader now

translate/src/modules/loaders/components/MachinerySkeletonLoader.tsx
wassafshahzad: Done

translate/src/modules/machinery/components/Machinery.tsx
wassafshahzad: Fixed. Unfortunately its a force of habit

translate/src/modules/machinery/components/Machinery.tsx
```diff
@@ -4,7 +4,7 @@ import useInfiniteScroll from 'react-infinite-scroll-hook';
 import { MachineryTranslations } from '~/context/MachineryTranslations';
 import { SearchData } from '~/context/SearchData';
 
-import { SkeletonLoader } from '~/modules/loaders';
+import { MachinerySkeletonLoader } from '~/modules/loaders/components/MachinerySkeletonLoader';
```
mathjazz: I meant this:

```suggestion
import { MachinerySkeletonLoader } from '~/modules/loaders/';
```

Note that you'll need to export the `MachinerySkeletonLoader` from the `loaders` module.

mathjazz (commented):
Thanks - one last thing and we're done!

translate/src/modules/machinery/components/Machinery.tsx
wassafshahzad: Oh sorry, I miss understood the feedback

mathjazz (commented):
Thanks, the code looks great now.

I’ve deployed the patch to the DEV server for more realistic testing:
https://pontoon.allizom.org/sl/firefox/all-resources/?string=315567

And only now I realized we only show the loader until the first result appears:
```js
machineryFetching && translations.length === 0
```

As you can see on the DEV server, the problem we are fixing in https://github.com/mozilla/pontoon/issues/4074 is that some Machinery results appear almost immediatelly, while the others sometimes take seconds. We'd like to make the users aware of the upcoming Machinery results by showing the loader.

So please change the condition such that the Machinery loader will be visible until all results are returned. Sorry for not catching this earlier!

wassafshahzad:
> Thanks, the code looks great now.
> 
> I’ve deployed the patch to the DEV server for more realistic testing: https://pontoon.allizom.org/sl/firefox/all-resources/?string=315567
> 
> And only now I realized we only show the loader until the first result appears:
> 
> ```js
> machineryFetching && translations.length === 0
> ```
> 
> As you can see on the DEV server, the problem we are fixing in #4074 is that some Machinery results appear almost immediatelly, while the others sometimes take seconds. We'd like to make the users aware of the upcoming Machinery results by showing the loader.
> 
> So please change the condition such that the Machinery loader will be visible until all results are returned. Sorry for not catching this earlier!

No Problem, This is on me, I should have doubled checked the success criteria. I have updated the code accordingly

mathjazz:
Thanks for the updatre!

@flodolo Could you please take this for a spin on https://pontoon.allizom.org/ and provide your thoughts about the UX? Do we need an always visible loading indicator near the tab title or is skeleton loader sufficient?

flodolo:
Looks fine, and the skeleton loader is enough. This is mostly for unexperienced users, who might not realize Pontoon provides machine translation services.

mathjazz (approved):

PR #4090, opened 2026-04-16
Sanitize comment content on edit

mathjazz:
Comment content is stored without sanitization when a comment is edited. Here we are fixing this by applying the same sanitization on edit that we already apply on comment creation.

@eemeli See https://bugzilla.mozilla.org/show_bug.cgi?id=2032281#c1 for more details
(@MundiaNderi doesn't have access sadly)

### DIFF
--- a/pontoon/base/views.py
+++ b/pontoon/base/views.py
@@ -245,9 +245,7 @@ def entities(request):
         return JsonResponse(
             {
                 "status": False,
-                "message": "{error}".format(
-                    error=form.errors.as_json(escape_html=True)
-                ),
+                "message": form.errors,
             },
             status=400,
         )
@@ -666,9 +664,7 @@ def add_comment(request):
         return JsonResponse(
             {
                 "status": False,
-                "message": "{error}".format(
-                    error=form.errors.as_json(escape_html=True)
-                ),
+                "message": form.errors,
             },
             status=400,
         )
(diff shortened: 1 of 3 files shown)


MundiaNderi:
I approve of this change. The current implementation relies on server-side sanitisation, and in https://github.com/mozilla/pontoon/blob/main/pontoon/base/tests/views/test_comment.py we currently only cover the happy paths.
To prevent regressions, I suggest adding dedicated sanitisation tests case(s).

pontoon/base/views.py
```diff
@@ -753,33 +753,31 @@ def unpin_comment(request):
...
-        return JsonResponse({"status": False, "message": "Bad Request"}, status=400)
+    """Edit a comment."""
+    form = forms.EditCommentForm(request.POST)
+    if not form.is_valid():
+        return JsonResponse(
+            {
+                "status": False,
+                "message": "{error}".format(
+                    error=form.errors.as_json(escape_html=True)
+                ),
```
eemeli: This is silly. You could just wrap the value with `str()`, but why are you using a JSON representation in the first place? I don't understand the intent here.

eemeli (approved):
See inline for one small fix, otherwise looks fine.

pontoon/base/views.py
mathjazz: Copy - paste. I fixed it in all places.

MundiaNderi (approved):

PR #4092, opened 2026-04-19
Fix false pending changes warning

### LINKED ISSUE #4083: Editor warns about unsaved changes also with no changes
I was navigating through some untranslated entities in the _Firefox for Android_ project and I got warned with the _"Unsaved Changes"_ semi-modal, even though I had made no changes to such entities whatsoever.

I tested the same in another project (e.g. _Common Voice_) and couldn't reproduce the issue there.

It looks as though the entities that I was navigating through, even though empty, triggered something that made the unsaved changes context check to return true.

See recording below for reference:
<video src="https://github.com/user-attachments/assets/ccda1b75-ec6a-4fed-b661-0568dd0e6fbf" />

flodolo:
Fix #4083 

`buildPattern()` returned `[]` on an empty string with a non-empty placeholders map. `getEmptyMessageEntry()` returns `['']`, so the comparison fails.

### DIFF
--- a/translate/src/context/Editor.test.jsx
+++ b/translate/src/context/Editor.test.jsx
@@ -15,6 +15,7 @@ import {
 import { EntityView, EntityViewProvider } from './EntityView';
 import { Locale } from './Locale';
 import { Location, LocationProvider } from './Location';
+import { UnsavedChanges, UnsavedChangesProvider } from './UnsavedChanges';
 
 function mountSpy(Spy, format, translation, original) {
   const history = createMemoryHistory({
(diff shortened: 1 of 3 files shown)


mathjazz (commented):
Thanks!

Could you please add a test case?

flodolo:
> Could you please add a test case?

This file was completely uncovered, I wasn't sure if we wanted to add a new file in these cases.

flodolo:
> This file was completely uncovered, I wasn't sure if we wanted to add a new file in these cases.

Not sure if this was the testing you were thinking of.

mathjazz:
> Not sure if this was the testing you were thinking of.

That's exactly what I meant - a test that fails with the old code and passes with the new. It's even better, because you've tested other stuff, too.

flodolo:
Added also a test also on the editor part. Both fails with the old code.

mathjazz (approved):

PR #4093, opened 2026-04-20
Represent empty patterns as `[]` rather than `['']`

eemeli:
PR #4092 was merged a bit too eagerly, as it was applying the wrong fix -- we should prefer `[]` rather than `['']` as the canonical representation of an empty message.

### DIFF
--- a/translate/src/utils/message/buildMessageEntry.test.js
+++ b/translate/src/utils/message/buildMessageEntry.test.js
@@ -3,7 +3,6 @@ import { getEmptyMessageEntry } from './getEmptyMessage';
 import { getPlaceholderMap } from './placeholders';
 import { parseEntry } from './parseEntry';
 import { serializeEntry } from './serializeEntry';
-import { pojoEquals } from '../pojo';
 
 const LOCALE = { code: 'en-US' };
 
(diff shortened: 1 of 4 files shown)


translate/src/utils/message/buildMessageEntry.test.js
```diff
@@ -21,9 +20,7 @@ describe('buildMessageEntry', () => {
     ]);
     const empty = getEmptyMessageEntry(base, LOCALE);
 
-    // serializeEntry() returns the same result for [] and [''], so using
-    // pojoEquals() in this test.
-    expect(pojoEquals(result, empty)).toBe(true);
```
eemeli, line 26: There's no need for `pojoEquals()` in tests, where `expect().toEqual()` does the same thing.

translate/src/utils/message/buildMessageEntry.tsx
```diff
@@ -79,5 +79,5 @@ function buildPattern(
     pos = end;
   }
   if (pos < str.length) pattern.push(str.substring(pos));
-  return pattern.length > 0 ? pattern : [str];
```
eemeli, line 82: This was confusing, as I needed to trace the whole function a few times to make sure that the alternative condition here is only reached when `str === ''`. So if we were to prefer code like this, it would be much clearer if the returned value was an explicit `['']`.

But returning an empty array is indeed the right thing to do here.

mathjazz (approved):
LGTM

flodolo (approved):

PR #4099, opened 2026-04-21
Fix incorrect warning on unchanged empty string

### LINKED ISSUE #4083: Editor warns about unsaved changes also with no changes

flodolo:
Fixed #4083, again 🤞🏼

### DIFF
--- a/translate/src/context/Editor.tsx
+++ b/translate/src/context/Editor.tsx
@@ -147,7 +147,7 @@ const createSimpleMessageEntry = (
 ): MessageEntry => ({
   format: getMessageEntryFormat(format),
   id: key[0] ?? '',
-  value: [value],
+  value: value ? [value] : [],
 });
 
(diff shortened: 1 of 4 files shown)


flodolo:
@eemeli 
Feel free to close and throw away if you have a better approach.

eemeli (approved):
These fixes look good.

PR #4106, opened 2026-04-23
Admin: sandbox jinja2 templates in emails

### LINKED ISSUE #3985: Use sandboxed Jinja2 environment in EmailContent rendering paths
See also:
https://bugzilla.mozilla.org/show_bug.cgi?id=2018156

functionzz:
This PR removes the ability for users with admin access to conduct RCE via jinja2 templates within Pontoon onboarding and user inactivity emails. It uses `SandboxedEnvironment` to prevent access to dangerous commands.

Fixes #3985.

Note: access to `settings` has been restricted to `INACTIVE_CONTRIBUTOR_PERIOD`, `INACTIVE_TRANSLATOR_PERIOD` and `INACTIVE_MANAGER_PERIOD`. Therefore, we must subsequently edit the onboarding and inactivity emails in the PROD and DEV dbs to reflect that, or else we risk crashing Pontoon.

Note 2: The above note expanded: all template tags in PROD and DEV must be of the format `{placeholder}`. `full_url` is no longer used within email templates, and jinja templating brackets (`{{}}`) are replaced with single brackets (`{}`).

### DIFF
--- a/pontoon/administration/tests/test_views.py
+++ b/pontoon/administration/tests/test_views.py
@@ -382,7 +382,7 @@ def test_manage_project_strings_download_csv(client_superuser):
 
 @pytest.mark.django_db
 def test_manage_project_translate_link_excludes_obsolete_resources(client_superuser):
-    """Test that translate_locale is only set when non-obsolete resources exist."""
+    """Test that Translate link is only shown when non-obsolete resources exist."""
     locale_kl = LocaleFactory.create(code="tlh", name="Klingon")
     project = ProjectFactory.create(
         data_source=Project.DataSource.DATABASE,
@@ -390,20 +390,24 @@ def test_manage_project_translate_link_excludes_obsolete_resources(client_superu
         repositories=[],
     )
 
+    url = reverse("pontoon.admin.project", args=(project.slug,))
+    translate_url = reverse(
+        "pontoon.localizations.localization", args=(locale_kl.code, project.slug)
+    )
+
     # add obsolete resource
     ResourceFactory.create(project=project, obsolete=True)
 
-    url = reverse("pontoon.admin.project", args=(project.slug,))
     response = client_superuser.get(url)
     assert response.status_code == 200
-    assert "translate_locale" not in response.context
+    assert translate_url.encode() not in response.content
 
     # add non-obsolete resource
     ResourceFactory.create(project=project, obsolete=False)
 
     response = client_superuser.get(url)
     assert response.status_code == 200
-    assert response.context["translate_locale"] == "tlh"
+    assert translate_url.encode() in response.content
 
 
 @pytest.mark.django_db
--- a/pontoon/administration/tests/test_views.py
+++ b/pontoon/administration/tests/test_views.py
@@ -396,14 +396,14 @@ def test_manage_project_translate_link_excludes_obsolete_resources(client_superu
     url = reverse("pontoon.admin.project", args=(project.slug,))
     response = client_superuser.get(url)
     assert response.status_code == 200
-    assert "translate_locale" not in response.context
+    assert "Translate" not in response.content.decode()
 
     # add non-obsolete resource
     ResourceFactory.create(project=project, obsolete=False)
 
     response = client_superuser.get(url)
     assert response.status_code == 200
-    assert response.context["translate_locale"] == "tlh"
+    assert "Translate" in response.content.decode()
 
 
 @pytest.mark.django_db
--- a/pontoon/administration/tests/test_views.py
+++ b/pontoon/administration/tests/test_views.py
@@ -396,14 +396,14 @@ def test_manage_project_translate_link_excludes_obsolete_resources(client_superu
     url = reverse("pontoon.admin.project", args=(project.slug,))
     response = client_superuser.get(url)
     assert response.status_code == 200
-    assert "Translate" not in response.content.decode()
+    assert "translate_locale" not in response.context
 
     # add non-obsolete resource
     ResourceFactory.create(project=project, obsolete=False)
 
     response = client_superuser.get(url)
     assert response.status_code == 200
-    assert "Translate" in response.content.decode()
+    assert response.context["translate_locale"] == "tlh"
 
 
 @pytest.mark.django_db
--- a/pontoon/messaging/emails.py
+++ b/pontoon/messaging/emails.py
@@ -6,7 +6,7 @@ from collections import defaultdict
 
 from celery import shared_task
 from dateutil.relativedelta import relativedelta
-from django_jinja.backend import Jinja2
+from jinja2.sandbox import SandboxedEnvironment
 from notifications.models import Notification
 
 from django.conf import settings
@@ -18,13 +18,21 @@ from django.utils import timezone
 
 from pontoon.actionlog.models import ActionLog
 from pontoon.base.models import Locale, UserProfile
+from pontoon.base.templatetags.helpers import full_url
 from pontoon.insights.models import LocaleInsightsSnapshot
 from pontoon.messaging.models import EmailContent
 from pontoon.messaging.utils import html_to_plain_text_with_links
+from pontoon.settings.base import (
+    INACTIVE_CONTRIBUTOR_PERIOD,
+    INACTIVE_MANAGER_PERIOD,
+    INACTIVE_TRANSLATOR_PERIOD,
+)
 
 
-jinja_env = Jinja2.get_default().env
+jinja_env = SandboxedEnvironment()
 log = logging.getLogger(__name__)
+
+jinja_env.globals["full_url"] = full_url
 
 
 def _get_monthly_user_actions(users, months_ago):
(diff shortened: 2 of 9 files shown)


mathjazz (commented):
I'm afraid this is not the right fix for #3985, which sadly has a misleading title.

The sandbox reduces risk but doesn't eliminate it, and it does so at the cost of ongoing complexity and trust in a security boundary that has already failed before. Eliminating the execution model entirely is a fundamentally stronger position.

Considering we still want editors to write template-like content in the DB, we should use a logic-less engine that has no concept of code execution.

We should probably be good with using [`str.format_map()`](https://docs.python.org/3/library/stdtypes.html#str.format_map):

```python
body = template_str.format_map({
    "INACTIVE_CONTRIBUTOR_PERIOD": settings.INACTIVE_CONTRIBUTOR_PERIOD,
    "homepage_url": request.build_absolute_uri(reverse("pontoon.homepage")),,
})
```

Alternatively, we could look into something more advanced like [Chevron](https://github.com/noahmorrison/chevron).

functionzz:
I see what you mean - we want to eliminate `SandboxedEnvironment` entirely?

mathjazz:
Yeah.

functionzz:
Just spent the last couple of hours debugging this - apparently the removal of jinja2 `SandboxedEnvironment` initialization in messaging somehow affects the way `request.context` was used for a specific test I wrote for #2133, which in turn was a flawed test because using `request.context` in such a way does not sit well with Jinja2, but was frailly working until the messaging related Jinja2 environment was removed.

I'm including the test change in this PR as its small and makes more sense to explain it here rather than merge seperately.

pontoon/messaging/emails.py
```diff
@@ -490,7 +513,13 @@ def send_inactive_translator_emails(users, translator_map):
             log.error(f"User {user} is not a translator of any locale.")
             continue
 
-        content = jinja_env.from_string(email_content.body).render({"locale": locale})
+        content = email_content.body.format_map(
+            {
+                "locale": locale,
```
mathjazz: Passing `locale` means templates can do things like:

```
{locale.__class__}
{locale._state}
{locale.__dict__}
```

But why do you even need to pass the locale in first place, now that the URLs are resolved in the view?

pontoon/administration/tests/test_views.py
```diff
@@ -396,14 +396,14 @@ def test_manage_project_translate_link_excludes_obsolete_resources(client_superu
     url = reverse("pontoon.admin.project", args=(project.slug,))
     response = client_superuser.get(url)
     assert response.status_code == 200
-    assert "translate_locale" not in response.context
+    assert "Translate" not in response.content.decode()
```
mathjazz: Are these changes related to the PR?

pontoon/messaging/emails.py
```diff
@@ -334,7 +333,16 @@ def send_onboarding_email_1(user):
...
+    content = email_content.body.format_map(
+        {
+            "INACTIVE_CONTRIBUTOR_PERIOD": settings.INACTIVE_CONTRIBUTOR_PERIOD,
+            "translate_url": full_url(
+                "pontoon.translate", "projects", "tutorial", "playground"
+            ),
+            "settings_url": full_url("pontoon.contributors.settings"),
+            "teams_url": full_url("pontoon.teams"),
+        }
+    )
```
mathjazz, line 351: I'd use this pattern to avoid `KeyError` on missing keys and still allow the use of curly braces:

```python
class SafeDict(dict):
    def __missing__(self, key):
        return "{" + key + "}"

content = email_content.body.format_map(SafeDict(context))
```

mathjazz (commented):
This moves in the right direction by removing Jinja execution for DB email content. 👏 

Note that `format_map()` still allows attribute access and currently receives a Django model instance. Please replace it as suggested.

We should also update the template files accordingly.

First in the repo:
https://github.com/functionzz/pontoon/tree/eb266b7b8a870461e8e7baecdafba7932ce6720a/pontoon/messaging/templates/messaging/emails/content

Then in the DB.

pontoon/administration/tests/test_views.py
functionzz: Yes, but in an indirect way. This test was failing due to reasons I posted about above, I'm including it in this PR so the test passes and there is a paper trail.

pontoon/messaging/emails.py
functionzz: correct, forgot to remove in the most recent commit.

pontoon/messaging/emails.py
functionzz: I was thinking maybe the `KeyError` would help us catch errors but realized that it would straight up crash the app lol.

pontoon/messaging/emails.py
```diff
@@ -406,7 +428,14 @@ def send_onboarding_emails_3(users):
     log.info("Start sending 3rd onboarding emails.")
 
     email_content = EmailContent.objects.get(email="onboarding_3")
-    content = jinja_env.from_string(email_content.body).render()
+    content = email_content.body.format_map(
+        SafeDict(
+            {
+                "docs_url": full_url("pontoon.docs", "localizer"),
```
mathjazz: This will raise an error.

pontoon/administration/tests/test_views.py
mathjazz: I just tested locally without this change and the tests are passing.

pontoon/messaging/emails.py
mathjazz: Which calls for unit tests.

functionzz:
The remote tests will not pass if the change is not applied. At least that's what I tested recently.

pontoon/messaging/emails.py
functionzz: Wait, so you want the SafeDict to raise an error? I'm assuming you mean that you would like some unit tests to cover this functionality.

mathjazz: I'm saying `full_url("pontoon.docs", "localizer")` raises:
```
django.urls.exceptions.NoReverseMatch: Reverse for 'pontoon.docs' with arguments '('localizer',)' not found. 1 pattern(s) tried: ['docs/\\Z']
```

And that we should increase tests coverage of `test_emails.py`, because right now send_onboarding_* functions for example are completely untested.

functionzz: I've increased test coverage, but I believe the problem is that the docs/localizer exists on prod but not on local or remote testing environments.

mathjazz: The problem is that there is no `pontoon.docs` URL pattern defined which expects arguments:
https://github.com/mozilla/pontoon/blob/main/pontoon/urls.py#L70C37-L70C49

pontoon/administration/tests/test_views.py
functionzz: > The remote tests will not pass if the change is not applied. At least that's what I tested recently.

Just quoting myself because I didn't put this comment in the right place - If I push the old code it fails the pytest in CI.

pontoon/messaging/emails.py
functionzz: I see, so way that `/docs/localizer` was formatted in the old template was completely intentional. Changed.

pontoon/messaging/templates/messaging/emails/content/onboarding_1.html
```diff
@@ -10,22 +10,19 @@
 <ol>
   <li>
     Take a
-    <a
-      href="{{ full_url('pontoon.translate', 'projects', 'tutorial', 'playground') }}"
-      >tour</a
-    >
+    <a href="{translate_url}">tour</a>
```
mathjazz: This would be slightly better:
```suggestion
    <a href="{tutorial_url}">tour</a>
```

mathjazz:
> The remote tests will not pass if the change is not applied. At least that's what I tested recently.

Could you please push? I'd like to look into it.

pontoon/messaging/tests/test_emails.py
```diff
@@ -60,3 +71,105 @@ def test_get_monthly_locale_stats_uses_end_of_month_snapshot():
...
+    try:
+        send_onboarding_email_1(user_a)
+    except EmailContent.DoesNotExist:
+        pytest.fail("EmailContent for 'onboarding_1' is missing from the DB.")
+    except NoReverseMatch as e:
+        pytest.fail(f"URL resolution failed: check URL config: {e}")
+    except TemplateSyntaxError as e:
+        pytest.fail(f"Template is broken: {e}")
+    except Exception as e:
+        pytest.fail(f"An unexpected error occurred: {e}")
```
mathjazz, line 88: These pass even if the function silently does nothing.

I’d prefer using Django’s mail outbox:

```python
from django.core import mail

send_onboarding_email_1(user_a)

assert len(mail.outbox) == 1
assert mail.outbox[0].to == [user_a.email]
```

functionzz: I added mail outbox as a check below the try-except

mathjazz:
Could you please rebase against upstream/main?

pontoon/messaging/tests/test_emails.py
mathjazz: I'd still drop at least `except Exception` -  we shouldn't be catching all exceptions.

flodolo:
Like @mathjazz, this PR passes for me locally without any changes 🤔

flodolo:
Correction after more ☕ : it fails locally as well, I was only running the emails test, not admin 😳 

Looks like the removal of `jinja_env = Jinja2.get_default().env` is what triggered the failure in another place, so Jamie's original fix was probably the correct one?

mathjazz:
Forgot to update this yesterday: The PR changes when the default Jinja backend is initialized, which seems to change what Django’s test client captures. so the fix here is to assert rendered behavior instead of Jinja internals.

mathjazz (approved):

PR #4112, opened 2026-04-26
Add ban audit trail for disabled users

### LINKED ISSUE #4016: Provide better information for admins about disabled users
It's now easy for admins to disable users, but we should have better information in case we remove the data that led to the ban (e.g. comments, translations).

- When the user was disabled
- By whom
- A field text with an explanation for the ban

All the info should be visible for admins in the user's profile page.

wassafshahzad:
## Description
Added UserBanLog model to create an audit trail for user enable/disable actions. Updated the contributor view and toggle_active_user_status to persist these actions to the database.

## Linked Issue
Closes #4016

### Disabled User
Profile Page  of disabled user

### Enabled User
enabled user

## What needs to be resolved before merging
- Is the UI for the profile page fine ?
- Right now, The ban reason is optional, should it be required ?
- The latest ban activity only shows up when the user is banned since I figured for an enabled user it might be noise. I that correct ? 
- I show the last ban activity and who perfumed it and the reason if it exists, Is that a valid approach or should we show last {x} ban activity with the reasons plus the enabled reasons ?


pontoon/contributors/templates/contributors/profile.html
```diff
@@ -123,6 +123,15 @@ <h2 class="display-name">{{ contributor.first_name }}</h2>
               method="POST"
             >
               {% csrf_token %}
+              <h4 class="subtitle account-status-heading">
+                {{ 'Disable' if contributor.is_active else 'Enable' }} account
+              </h4>
+              <textarea
+                class="text-area"
```
mathjazz:
```suggestion
                class="disable-reason"
```

pontoon/contributors/templates/contributors/profile.html
```diff
@@ -197,6 +206,26 @@ <h4 class="subtitle">Member since</h4>
               <span>{{ contributor.date_joined|format_datetime("date") }}</span>
             </div>
           </div>
+          {% if ban_activity and not contributor.is_active and is_admin %}
+            <div class="block">
+              <h4 class="subtitle">Latest ban</h4>
```
mathjazz:
```suggestion
              <h4 class="subtitle">Disabled on</h4>
```

pontoon/contributors/templates/contributors/profile.html
```diff
@@ -197,6 +206,26 @@ <h4 class="subtitle">Member since</h4>
...
+                >
+                {% if ban_activity.performed_by %}
+                  <span
+                    >by
+                    {{ ban_activity.performed_by.first_name or ban_activity.performed_by.username }}</span
+                  >
+                {% endif %}
+                {% if ban_activity.action_reason %}
+                  <span>— {{ ban_activity.action_reason }}</span>
+                {% endif %}
```
mathjazz: These are partially cut-off in the UI, so I would show this information on hover only. I would also show more details about the ban when you hover the "Account disabled by an administrator" text.

pontoon/contributors/views.py
```diff
@@ -135,6 +140,20 @@ def contributor(request, user):
...
+        UserBanLog.ActionType.UNBANNED
+        if target.is_active
+        else UserBanLog.ActionType.BANNED
+    )
+    UserBanLog.objects.create(
+        performed_by=actor,
+        performed_on=target,
+        action_reason=reason,
+        action_type=action_type,
+    )
```
mathjazz: Since this is not a view, I'd rather move it to the model or the `utils.py` file in this module.

pontoon/contributors/views.py
```diff
@@ -126,6 +126,11 @@ def contributor(request, user):
         "contribution_timeline": {
             "contributions": timeline_data,
         },
+        "ban_activity": user.ban_log.filter(
```
mathjazz, line 129:
```suggestion
        "latest_ban": user.ban_log.filter(
```

pontoon/contributors/templates/contributors/profile.html
```diff
@@ -123,6 +123,15 @@ <h2 class="display-name">{{ contributor.first_name }}</h2>
               method="POST"
             >
               {% csrf_token %}
+              <h4 class="subtitle account-status-heading">
+                {{ 'Disable' if contributor.is_active else 'Enable' }} account
+              </h4>
+              <textarea
+                class="text-area"
+                name="action_reason"
+                placeholder="Reason (optional)"
```
mathjazz:
```suggestion
                placeholder="Reason for disabling (optional)"
```

pontoon/contributors/templates/contributors/profile.html
```diff
@@ -123,6 +123,15 @@ <h2 class="display-name">{{ contributor.first_name }}</h2>
               method="POST"
             >
               {% csrf_token %}
+              <h4 class="subtitle account-status-heading">
```
mathjazz: I would drop this title. I would only make the `<textarea>` visible after the user clicks the Disable button for the first time. Then the placeholder reveals the purpose of it. And when you click the Disable button the 2nd time, the user gets disabled.

pontoon/contributors/templates/contributors/profile.html
```diff
@@ -197,6 +206,26 @@ <h4 class="subtitle">Member since</h4>
               <span>{{ contributor.date_joined|format_datetime("date") }}</span>
             </div>
           </div>
+          {% if ban_activity and not contributor.is_active and is_admin %}
+            <div class="block">
+              <h4 class="subtitle">Latest ban</h4>
+              <div class="item-with-icon">
+                <span class="icon fas fa-ban"></span>
```
mathjazz, line 220: We should also use this icon in the Disable button.

mathjazz (commented):
Great work! Left some suggestions inline.

We should also update the Deactivating users section in that docs, now that this can be done in the Profile page and even the reason can be provided:
https://pontoon.mozilla.org/docs/admin/managing-users/#deactivating-users

flodolo:
Note the [comment](https://github.com/mozilla/pontoon/issues/4016#issuecomment-4330281424) left in the issue that might get lost

> When writing tests I discovered the following pieces of code:
> 
> assert response.status_code, 200 assert user_a.is_active is True assert response.status_code == 200, User.objects.get(pk=user_a).is_active is False assert response.status_code == 200, User.objects.get(pk=user_a).is_active is True
> 
> In these asserts, is the intent to verify is_active and status_code? If so, these are incorrect, the second argument to assert is just the error message, not a second condition being checked. They should probably be split into separate assert statement

wassafshahzad:
@mathjazz Addressed feedback

pontoon/contributors/templates/contributors/profile.html
```diff
@@ -31,9 +31,13 @@
             <div class="desc">Update profile picture</div>
           {% endif %}
           {% if profile_is_disabled %}
+            {% set disable_by = ("Disabled by " + (ban_activity.performed_by.first_name or ban_activity.performed_by.username)) if (ban_activity and ban_activity.performed_by) else "" %}
+            {% set disable_reason = (": " + ban_activity.action_reason) if (ban_activity and ban_activity.action_reason) else "" %}
```
mathjazz: I would factor this out and put it after the `set` block that ends in line 24:

```jinja2
{% if ban_activity %}
  {% set disabled_title =
    "Disabled by " ~ ban_activity.performed_by.display_name ~
    (": " ~ ban_activity.action_reason if ban_activity.action_reason else "")
  %}
{% endif %}
```

mathjazz: And then just use this below:

```jinja2
<span title="{{ disabled_title }}">
```

mathjazz (commented):
Nice work! One last improvement for code maintainability and we're good to go.

wassafshahzad:
> Nice work! One last improvement for code maintainability and we're good to go.

Ok will do and sorry for the delay

wassafshahzad:
@mathjazz Addressed feedback

pontoon/contributors/templates/contributors/profile.html
```diff
@@ -33,7 +39,9 @@
           {% if profile_is_disabled %}
             <div class="status">
               <i class="fas fa-exclamation-circle fa-sm"></i>
-              <span>Account disabled by an administrator</span>
+              <span title="{{ disabled_title }}"
```
mathjazz: This will likely fail for old disabled users for which we don't have `ban_activity` on record (so `disabled_title` will not be set). So please use something like this:

```suggestion
              <span{% if disabled_title %} title="{{ disabled_title }}"{% endif %}>
```

pontoon/contributors/templates/contributors/profile.html
```diff
@@ -197,6 +211,23 @@ <h4 class="subtitle">Member since</h4>
...
+              <h4 class="subtitle">Disabled on</h4>
+              <div class="item-with-icon">
+                <span class="icon fas fa-ban"></span>
+                <span
+                  title="{% if ban_activity.performed_by %}
+                    Disabled by
+                    {{ ban_activity.performed_by.first_name or ban_activity.performed_by.username }}
+                  {% endif %}{% if ban_activity.action_reason %}
+                    : {{ ban_activity.action_reason }}
+                  {% endif %}"
```
mathjazz: Why don't use use `disabled_title` here?

pontoon/contributors/templates/contributors/profile.html
wassafshahzad: Done

pontoon/contributors/templates/contributors/profile.html
wassafshahzad: This part escaped my notice, I have implemented the fix

mathjazz (approved):
Great work! Thanks for the patch!

PR #4118, opened 2026-04-28
Translate: remove google translate from concordance search results

### LINKED ISSUE #4117: Translate: remove Google Translate as an option when accessing Concordance Search
In the Translate module, the MACHINERY tab includes a Google Translate section which is useful for viewing an alternative translation to Translation Memories. The issue is, this behaviour leaks into Concordance Search, where a user can search for something within Concordance Search and also get a Google Translation, which is not the intent of the search results. We need to remove Google Translate as an option within Concordance Search to more accurately reflect the intent of the user.

## Screenshots

Image

functionzz:
When searching via Concordance Search, Google Translate shows up as one of the first results, which is not the intended behaviour of Concordance Search. This PR simply removes this functionality.

Fixes #4117.

### DIFF
--- a/translate/src/modules/machinery/components/Machinery.test.jsx
+++ b/translate/src/modules/machinery/components/Machinery.test.jsx
@@ -62,15 +62,12 @@ describe('<Machinery>', () => {
         { original: '3', sources: [] },
       ],
       {
-        results: [
-          { original: '4', sources: [] },
-          { original: '5', sources: [] },
-        ],
+        results: [],
       },
     );
 
-    expect(getAllByRole('listitem')).toHaveLength(5);
-    expect(getAllByTitle(copyListItemTitle)).toHaveLength(5);
+    expect(getAllByRole('listitem')).toHaveLength(3);
+    expect(getAllByTitle(copyListItemTitle)).toHaveLength(3);
   });
 
   it('renders a reset button if a search query is present', () => {
--- a/translate/src/modules/machinery/components/Machinery.test.jsx
+++ b/translate/src/modules/machinery/components/Machinery.test.jsx
@@ -54,7 +54,7 @@ describe('<Machinery>', () => {
     getByRole('searchbox');
   });
 
-  it('shows the correct number of translations', () => {
+  it('shows machinery translations when concordance search is not active', () => {
     const { getAllByRole, getAllByTitle } = mountMachinery(
       [
         { original: '1', sources: [] },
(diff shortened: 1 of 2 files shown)


translate/src/modules/machinery/components/Machinery.tsx
```diff
@@ -71,27 +71,30 @@ export function Machinery(): React.ReactElement<'section'> {
...
-          {translations.map((translation, index) => (
-            <MachineryTranslationComponent
-              index={index}
-              sourceString={source}
-              translation={translation}
-              key={index}
-            />
-          ))}
-        </ul>
+        {results.length === 0 ? (
```
mathjazz: I'm not sure about this condition. If I search for something that doesn't return any Concordance search results, will I again see Machine Translation results?

translate/src/modules/machinery/components/Machinery.test.jsx
```diff
@@ -62,15 +62,12 @@ describe('<Machinery>', () => {
...
-          { original: '5', sources: [] },
-        ],
+        results: [],
       },
     );
 
-    expect(getAllByRole('listitem')).toHaveLength(5);
-    expect(getAllByTitle(copyListItemTitle)).toHaveLength(5);
+    expect(getAllByRole('listitem')).toHaveLength(3);
+    expect(getAllByTitle(copyListItemTitle)).toHaveLength(3);
```
mathjazz, line 77: We should explicity test that only Concordance search results are shown when using Concordance search,

translate/src/modules/machinery/components/Machinery.tsx
functionzz: good point.

translate/src/modules/machinery/components/Machinery.tsx
```diff
@@ -71,27 +72,30 @@ export function Machinery(): React.ReactElement<'section'> {
...
-          ))}
-        </ul>
+        {isConcordanceSearchActive ? (
+          <ul>
+            {results.map((result, index) => (
+              <MachineryTranslationComponent
+                index={index + translations.length}
+                sourceString={query}
+                translation={result}
+                key={index + translations.length}
```
mathjazz: I’d change those to `index={index}` and `key={index}`. 

The old offset only made sense when both lists rendered together.

functionzz: ah yes, forgot about this

mathjazz (approved):

PR #4124, opened 2026-05-04
Fix missing space in 'message new' class in sent.html

### LINKED ISSUE #4113: Messaging center: "Sent" layout looks broken right after sending a message
Looks like it gets an unstyled class "messagenew". Clicking on Sent again didn't change things, had to force-reload the page to fix.

Image

Kasturijadhav14:
Fixes #4113

## Problem
After sending a message, the "Sent" panel displayed the newly sent 
message with an unstyled class "messagenew" instead of "message new". 
This caused the highlight styling to not apply to the new message. 
Clicking "Sent" again didn't fix it — a force-reload was required.

## Root Cause
In `sent.html`, the `new` class was missing a space before it:

```html
<!-- Before (buggy) -->
<li class="message{% if message.is_new() %}new{% endif %}">

<!-- After (fixed) -->
<li class="message{% if message.is_new() %} new{% endif %}">
```

## Fix
Added a single space before `new` inside the `{% if %}` block so the 
two CSS classes are properly separated.

### DIFF
--- a/pontoon/messaging/templates/messaging/includes/sent.html
+++ b/pontoon/messaging/templates/messaging/includes/sent.html
@@ -7,7 +7,7 @@
   {% endif %}
   <ul>
     {% for message in sent_messages %}
-      <li class="message{% if message.is_new() %}new{% endif %}">
+      <li class="message {% if message.is_new() %}new{% endif %}">
         <div class="user-avatar">
           <a
             href="{{ url('pontoon.contributors.contributor.username', message.sender.username) }}"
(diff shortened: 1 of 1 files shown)


flodolo:
You have undone your change with your second commit, there are currently no changes compared to `main`.

Kasturijadhav14:
Hi @flodolo, thank you for the feedback! I have pushed the fix again with the correct space added. The third commit restores the change. Sorry for the confusion caused by the second commit.

flodolo (approved):
Thanks, this works as expected now.

Congratulation on your first contribution to Pontoon 🎉

PR #4126, opened 2026-05-05
Update obsolete links in docs and READMEs

flodolo:


documentation/README.md
```diff
@@ -6,6 +6,6 @@ It consolidates Pontoon documentation from three previously separate sites:
 
 | Audience | Previous location |
 |---|---|
-| Developers | [docs/](https://mozilla-pontoon.readthedocs.io/en/latest/) |
+| Developers | [docs/](https://pontoon.mozilla.org/docs/dev/) |
```
mathjazz: Not sure about this one. Maybe we should drop the link to Previous location for this Audience?

flodolo: I don't think I understand. What are you suggesting to do?

mathjazz: The text above says: "It [unified Pontoon documentation site] consolidates Pontoon documentation from three previously separate sites:"

And then the first bullet point links to the "unified Pontoon documentation site".

It should link to the old documentation site, like the other two bullet points. But it can't, because the old Developer docs are redirected to the new docs.

Maybe... we should just leave the first line ("This folder contains") and drop the rest? Or drop the entire README file?

flodolo: Thanks, clearer now. I think having the file is useful for people landing in the folder, I'll try adding a link to "contributing" with instructions on how to build the docs.

mathjazz (approved):

Issue #2168, opened 2019-05-27
Use language-sensitive formatting in asLocaleString


bugzilla-to-github:
*This issue was created automatically by a [script](https://github.com/mathjazz/bugzilla2github/).*
## [Bug 1554940](https://bugzilla.mozilla.org/show_bug.cgi?id=1554940)
Bug Reporter: @mathjazz

The asLocaleString() method in core/utils hardcodes the en-GB locale to make number formatting consistent across Pontoon.

Once we have support for more locales in Pontoon, we should make this internationalized.

ewerybody:
I'd like to work on this! Forked and prepared a branch already. PR asap. :)

mathjazz:
👋 Assigned. :)

PR #4127, opened 2026-05-05
#2168 make asLocaleString use locale code

ewerybody:
Now `asLocaleString` additionally gets locale code passed from context. No more hardcoded en-GB.
Force latin digits via `numberingSystem: 'latn'` to keep UI stats readable regardless of locale script (e.g. sat-Olck, fa).
Added tests for Latin digit and en-GB formatting.

See issue #2168

### DIFF
--- a/translate/src/utils/asLocaleString.test.js
+++ b/translate/src/utils/asLocaleString.test.js
@@ -1,0 +1,19 @@
+import { asLocaleString } from './asLocaleString';
+
+// French (and other locales) use Unicode Character 202F "NARROW NO-BREAK SPACE"
+// (U+202F) as a thousands separator. See:
+// https://github.com/unicode-org/cldr/blob/24f92afa0b467992a503c17aec466cdabebab282/common/main/fr.xml#L7165
+const NNBSP = '\u202f';
+
+describe('asLocaleString', () => {
+  it('formats numbers with language-sensitive separators', () => {
+    expect(asLocaleString(1234567, 'en-GB')).toEqual('1,234,567');
+    expect(asLocaleString(1234567, 'de')).toEqual('1.234.567');
+    expect(asLocaleString(1234567, 'fr')).toEqual(`1${NNBSP}234${NNBSP}567`);
+  });
+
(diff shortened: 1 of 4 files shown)


translate/src/utils/asLocaleString.test.js
```diff
@@ -0,0 +1,19 @@
...
+  it('formats numbers with language-sensitive separators', () => {
+    expect(asLocaleString(1234567, 'en-GB')).toEqual('1,234,567');
+    expect(asLocaleString(1234567, 'de')).toEqual('1.234.567');
+    expect(asLocaleString(1234567, 'fr')).toEqual(`1${NNBSP}234${NNBSP}567`);
+  });
+
+  it('always uses Latin digits regardless of locale', () => {
+    expect(asLocaleString(1234567, 'fa')).toMatch(/^[0-9,.\s${NNBSP}]+$/);
+    expect(asLocaleString(1234567, 'sat-Olck')).toMatch(/^[0-9,.\s${NNBSP}]+$/);
+  });
```
eemeli: This is overfitting to the current serializations. We should (in general) avoid testing against exact localized output, as that may change in any CLDR/ICU update. The function being tested is also rather simple.

It should be enough to test that the output contains digits, and that e.g. French and English outputs differ.

eemeli (changes requested):
Looks pretty good; see inline for a required fix to the tests.

We should either modify the parent issue or file a new issue to cover the [other places](https://github.com/search?q=repo%3Amozilla%2Fpontoon%20Intl.NumberFormat&type=code) where we're formatting numbers with a fixed `'en'` locale.

translate/src/utils/asLocaleString.test.js
ewerybody: Ah OK. Yeah can do! Thanks for reviewing! :)

ewerybody:
Pardon me I was a little too quick with the review request button. Now it should work.

And I'd open up a new issue for the other ones if that's OK?

eemeli (approved):

